PR #1971, opened 2026-08-07
fix(stella-pipeline): unbreak main — four breaks two merges left behind, incl. a silently deleted #1793 witness

macanderson:
## What

`main` at b5ab7f84 did not build: `cargo clippy -p stella-pipeline --all-targets -- -D warnings` failed. A compile error in the lib masks the test target, so it read as one failure and was really **four**, from two different merges.

### From #1953 (the #1778 research stage)

1. **`management_prompt/tests.rs` (E0004)** — `ModelCallRole::Research` is a new variant and `management_system_block`'s match is exhaustive by design. `Research` rides the sub-agent primitive, so its system prompt travels on the `SubAgentSpec`, never through `metered_raw_call` — it joins the never-dispatched arm, and `ALL_ROLES` grows to 15.
2. **`pipeline.rs` (`too_many_arguments`)** — the new `research` parameter pushed `plan_stage` to 8 args, one over the cap. Bundled `budget`/`total` into the `Spend` struct every stage downstream of the fan-out already takes, rather than `#[allow]`-ing the lint. `pipeline.rs` is a god file, so the doc comment is written to land the file back at **exactly** its 3462 ceiling — no growth.

### From #1951 — a same-seam clobber

#1951 rewrote `tests/verification_hardening.rs` wholesale and dropped three items #1945 had added to it hours earlier, in a file its own subject (per-candidate verifier degradation) never needed to touch:

3. **`PassingShell` + `shell_call_result` (E0425 ×2)** — the child module `flip_halt_arming.rs` was left referencing two doubles that exist nowhere in the tree.
4. **`a_revision_halts_at_the_step_where_the_tracked_test_flips`** — the configured-command **witness for #1793**. Deleting it failed no gate, because the crate had already stopped compiling for reason 3. **#1793 has been shipping with half its witness silently gone.** Restored verbatim from eddf9700.

## The structural fix, not just the restore

Restoring the witness took the parent file to 1557 lines, which `file-size` rejects outright — and the baseline takes no new entries. So this splits rather than exempts, and the split is the one the content was asking for: **both** #1793 witnesses and the two doubles they share now live in `verification_hardening/flip_halt_arming.rs`, the module already named for the concern. Parent drops to 1434.

That the two witnesses were ever in separate files is what let the clobber happen quietly. With the cluster in one file, the same wholesale rewrite is a merge conflict instead of a silent deletion — the module doc records why.

## Witness

[...]

PR #1995, opened 2026-08-07
fix(stella-cli,stella-pipeline): seed and announce the candidate task board in isolated runs (#1719)

### LINKED ISSUE #1719: Plan steps never change state during an isolated run — the candidate's task board is private, unseeded, and unannounced
## Symptom

The PLAN rail shows every step as a hollow grey `○` for the whole turn, however
long the run goes. The panel title's own dot moves (`PLAN ● approved 0/7`) but
no individual step ever reaches started / complete / cancelled, so the rail
reports a plan and never reports progress through it.

Reproduce: run `stella` in the Command Deck on a task big enough to reach the
scope-review gate, approve the plan, and watch the PLAN panel for the rest of
the turn.

## Root cause

A step's ring moves only when `AgentEvent::TaskUpdate` reaches the renderer
(`crates/stella-tui/src/model.rs` → `Plan::apply_board`). The only emitter is
`TaskTap` in `crates/stella-cli/src/command_deck.rs`.

`Pipeline::run_isolated_candidate`
(`crates/stella-pipeline/src/pipeline/fanout_stage.rs`) builds its engine on
`ws.tools()` — the candidate workspace's own stack, assembled in
`crates/stella-cli/src/candidate_ws.rs` as registry → customs → MCP → policy.
`TaskTap` is not in that chain. Two independent failures follow, and either
alone freezes the rail:

1. **The board is private and unseeded.** The forwarder seeds the *session*
   registry's board from the approved proposal
   (`TaskBoard::seed_from_plan`, `crates/stella-cli/src/command_deck/forwarder.rs`),
   but a candidate `ToolRegistry` builds its own empty one. `task_start "3"`
   answers `UnknownTask` for a step the gate already numbered 3. A worker that
   routes around it with `task_create` then builds rows whose subjects
   *overwrite* the approved step titles, because `Plan::steps`
   (`crates/stella-tui/src/plan.rs`) joins board rows onto proposed steps by
   ordinal id and lets the board's subject win.
2. **Nothing announces it.** Even a correct mutation reaches no surface.

This is the same shape as the already-fixed "isolated runs blind the Files tab"
bug — see the `attach_read_events` comment at `candidate_ws.rs`, which split
reads from mutations for exactly this reason. The task board was missed in that
sweep.

## Evidence

Execution 66 of a real local session (`.stella/private/store.db`):

```
sqlite3 .stella/private/store.db \
  "SELECT event_type, COUNT(*) FROM events WHERE execution_id=66 GROUP BY event_type;"
-- scope_review|1   (7 steps proposed and approved)
-- ... 1168 events total, ZERO task_update

sqlite3 .stella/private/store.db \
  "SELECT json_extract(payload,'\$.call.name'), COUNT(*) FROM events
   WHERE execution_id=66 AND event_type='tool_start' GROUP BY 1;"
-- task_create|3
-- task_start|1
```

The model drove the board exactly as the system prompt asks
(`crates/stella-cli/src/agent/prompt.rs`, the `tool_steering!` block) and not
one of those calls produced an event.

## Definition of done

- An isolated single candidate's `task_*` call moves the step the approval gate
  numbered, and emits `TaskUpdate` on the turn's channel.
- A concurrent best-of-N fan-out does not regress: several candidates must not
  share one board, because `TaskBoard::set_status` rejects a transition out of a
  terminal state and the second candidate to finish a step would get a tool
  error for work it really did.
- A host that renders no plan (fleet worker, headless `stella run`) is
  unchanged.
- Witness test proving the fail→pass flip.

## Constraints

- `crates/stella-cli/src/command_deck.rs`, `crates/stella-cli/src/agent.rs` and
  `crates/stella-tools/src/registry.rs` are grandfathered god files
  (`scripts/file-size-baseline.txt`) — new logic goes in sibling modules.
  `candidate_ws.rs` is near the 1500-line ceiling with no baseline entry, so it
  cannot cross it at all.
- `stella_pipeline::ports::CandidateWorkspacePort::create` carries no fan-out
  width, so the "how many candidates may report" decision cannot read `n`.

macanderson:
Closes #1719

> **Stacks on #1971.** `main` is red, and this branch carries #1971's unbreak commits so it can compile and be tested. Once #1971 merges, this PR's diff reduces to the single commit described below — 7 files, +431/−25.

## The defect

The PLAN rail shows every step as a hollow grey `○` for the whole turn. The panel's own dot moves (`PLAN ● approved 0/7`) but no individual step ever reaches started / complete — the rail reports a plan and never reports progress through it.

A step's ring moves only on `AgentEvent::TaskUpdate`, whose only emitter is the deck's session `TaskTap`. `Pipeline::run_isolated_candidate` builds its engine on `ws.tools()` — the candidate workspace's own stack — and `TaskTap` is not in that chain. Two independent failures follow, either of which alone freezes the rail:

1. **The board is private and unseeded.** The forwarder seeds the *session* registry's board from the approved proposal, but a candidate `ToolRegistry` builds its own empty one. `task_start "3"` answers `UnknownTask` for a step the gate already numbered 3. A worker routing around it with `task_create` then builds rows whose subjects *overwrite* the approved step titles.
2. **Nothing announces it.** Even a correct mutation reaches no surface.

Execution 66 of a real session: 1168 events, **zero** `task_update`, while the model called `task_create` ×3 and `task_start` ×1 exactly as the system prompt asks.

This is the same shape as the already-fixed "isolated runs blind the Files tab" bug — see the `attach_read_events` comment in `candidate_ws.rs`. The task board was missed in that sweep.

## The fix

**Seeding** — `CandidateWorkspace` grows a default-no-op `seed_task_board(steps, announce)`. The fan-out stage drives it right after `create`, before any dispatch, so the worker's very first `task_start` already resolves the gate's ordinals. It passes the plan's step descriptions verbatim, which is what `scope::build_proposal` renders the approval card from — so a candidate board and the session board seeded from `ScopeReview` cannot disagree on a step's id or title.

**Announcing** — the candidate tool stack gains a `TaskTap`-shaped decorator (`candidate_ws/task_events.rs`), outermost over registry + customs + candidate MCP + policy, so it observes every route a `task_*` call can take.

**Why the `announce` latch, and why boards stay private.** The issue notes `CandidateWorkspacePort::create` carries no fan-out width, so the "how many candidates may report" decision cannot read `n`. It is taken in `create_candidate_workspaces`, which is the one seam that knows both `n` and the plan. `TaskUpdate` carries a full board snapshot with no candidate tag, so several boards reporting onto one channel would splice into a checklist that is nobody's — the same reasoning that already mutes `TextDelta`/`Reasoning` on a shared event lane. Only a lone candidate announces.

Boards are **never** shared, at any width: `TaskBoard::set_status` rejects a transition out of a terminal state, so the second sibling to finish a step would receive a tool error for work it really did.

## Definition of done, item by item

| From the issue | Where |
|---|---|
| An isolated candidate's `task_*` moves the gate-numbered step and emits `TaskUpdate` | `a_seeded_lone_candidate_moves_the_approved_step_and_announces_it` |
| Best-of-N does not regress — candidates must not share one board | `fanout_candidates_keep_separate_and_silent_boards`, `best_of_two_adopts_only_the_winner_and_removes_every_workspace` |
| A host rendering no plan (fleet worker, headless) is unchanged | Default no-op on the port; `events: None` skips the tap entirely; the witness author's pristine snapshot is created via `port.create()` directly (`witness_stage.rs:615`), never seeded, so its latch stays `false` |
| Witness proving the fail→pass flip | below |

## Witness

The new API means the tests cannot compile on `main` (the feature is genuinely absent). That is the weaker half of the proof, so both halves of the fix were **independently neutered** on this branch to show each is load-bearing:

**Seeding neutered** — reproduces the issue's exact symptom:
```
a_seeded_lone_candidate_moves_the_approved_step_and_announces_it ... FAILED
  the gate-numbered step must resolve on the candidate board:
  Error { message: "no task with id 1 — call task_list to see the board" }
fanout_candidates_keep_separate_and_silent_boards ... FAILED
```

**Announcing neutered:**
```
a_seeded_lone_candidate_moves_the_approved_step_and_announces_it ... FAILED
  a lone candidate's task_start must announce a TaskUpdate
```

(The fan-out test correctly still passes with the tap dead — it asserts *silence*, which a dead tap trivially satisfies. Its real job, the separate-boards guarantee, fails in the seeding run above.)

Restored, both pass. The two CLI-side witnesses run against **real git worktrees** through the production `GitCandidateWorkspaces` port, not a fake.

## Constraints honoured

- `candidate_ws.rs` was near the 1500-line ceiling with no baseline entry, so it could not cross it at all: it takes **+14 lines** (1480 → 1494) and all new logic lands in the new sibling `candidate_ws/task_events.rs`.
- `pipeline.rs` is a god file — **+0 net lines** (unchanged at its 3462 ceiling).
- No baseline entry added or raised.

## Verification

- `cargo test -p stella-pipeline -p stella-cli` — **2121 passed, 0 failed** (20 binaries)
- `cargo clippy -p stella-pipeline -p stella-cli --all-targets -- -D warnings` — clean
- `make guards-fast` — green, `file-size` and `god-files` included

## Summary by Sourcery

Ensure isolated candidate workspaces share the approved task plan and surface task progress while tightening flip-halt and management accounting coverage.

New Features:
- Seed each candidate workspace with the approved plan steps and optionally allow a lone candidate to announce task-board updates on the shared event stream.

Enhancements:
- Extend the candidate workspace port to accept plan step seeding and an announce latch, and wire this through the fan-out stage and Git-backed candidate workspaces.
- Clarify management accounting for the Research role and refactor planning budget handling to use a shared Spend wrapper parameter.
- Add a passing-shell tool executor and supporting helpers to exercise flip-halt behavior when observing real test commands.

Tests:
- Add pipeline and CLI tests to verify seeded candidate task boards resolve gate-numbered steps, keep per-candidate boards, and only announce from lone candidates.
- Add a configured-command flip-halt witness test to ensure revisions halt when the tracked test flips from fail to pass.
- Update management accounting tests to work with the new Spend wrapper and expanded role set.
- Extend best-of-N isolation tests to assert that seeded task boards remain private and silent in multi-candidate runs.


macanderson:
Follow-ups filed per AGENTS.md § "Nothing left behind":

- **#1996** — a best-of-N fan-out's PLAN rail is *still* frozen. This PR arms the announce latch only for `n == 1`, because `TaskUpdate` carries a full board snapshot with no candidate tag. The honest fix changes the wire contract (a candidate discriminator + `make wire-schema-update` + a renderer policy decision in `Plan::apply_board`), which is out of scope for a P1 rail fix. Filed with the design options and the constraint that boards must stay per-candidate.
- **#1997** — the systemic gap behind the stacked #1971: a silently deleted test fails no gate.

Issue #1997, opened 2026-08-07
A silently deleted test fails no gate — #1951 deleted a #1793 witness and nothing objected


macanderson:
## What happened

#1945 landed the two witnesses for #1793 (the mid-turn flip halt) plus the two doubles they share. Hours later #1951 — a PR about per-candidate verifier degradation — rewrote `crates/stella-pipeline/src/pipeline/tests/verification_hardening.rs` wholesale and **dropped three of those items**:

- `a_revision_halts_at_the_step_where_the_tracked_test_flips` — the configured-command **witness for #1793**
- `PassingShell` and `shell_call_result` — the doubles both witnesses need

Confirmed by `git log -S`:

```
$ git log --oneline -S"PassingShell" --all
2a142b26 fix(stella-pipeline,stella-protocol): ... (#1787) (#1951)   <- removed
eddf9700 fix(stella-pipeline): arm FlipHalt ... (#1793) (#1945)      <- added
```

#1951 had no reason to touch that file's flip-halt content at all — a same-seam clobber of the kind AGENTS.md already warns about.

## Why nothing caught it

Deleting a test is invisible to every gate. `file-size`, `left-behind`, `god-files`, clippy, and the test suite all pass a tree with fewer tests in it — the suite is green **because** the test is gone.

Here it was masked twice over: the crate stopped compiling (E0425 for the two missing doubles), and a compile error in the lib masks the test target entirely, so the whole thing read as one build failure. #1793 shipped with half its witness silently absent, and the only reason anyone noticed is that a **different** merge (#1953) broke the build in a way that forced someone to read the file. Restored in PR #1971.

## Why it matters here specifically

This repo's contract is "verified done, not claimed done". A witness test that can be deleted by an unrelated PR, with no gate objecting, means the *evidence* for a shipped fix has a weaker retention guarantee than the fix itself. #1793 is not special — any witness is deletable this way.

## Possible directions (design decision, not a spec)

1. **A test-count ratchet.** Cheapest and dumbest: record per-crate `#[test]`/`#[tokio::test]` counts in a generated baseline like `scripts/file-size-baseline.txt`, and fail when a count *drops* without the baseline being regenerated in the same commit. Catches this exact case; noisy under legitimate consolidation, and the escape hatch (`make test-count-update`) is one command, so the cost is a reviewable diff rather than a block. Mirrors machinery the repo already has and trusts.
[...]

Issue #1986, opened 2026-08-07
CI: ci.yml does not run on a push to main, so main goes red and stays red (twice by the same boot.rs doc link)


macanderson:
## Problem

`main`'s required `fmt + clippy + test` job has been failing for **five
consecutive merges** (`2a142b26` #1951, `dcb9859b` #1952, `1bdf2daf` #1953,
`b5ab7f84` #1955, `43402ae4` #1965). Because a red `main` is inherited by every
branch, every open PR shows a red gate that is not its own — which trains
reviewers to ignore the signal, and hides the PR's real failures underneath.

The mechanism is stated in #1927's own description and has not been addressed:

> `ci.yml` does not run on a push to `main`.

So a PR can be green at review time and still red `main` on merge, and nothing
re-runs to notice. With `enforce_admins` and auto-merge in play, the window is
routinely exercised rather than theoretical.

## Evidence that this recurs, not just happens

`crates/stella-cli/src/daemon/boot.rs`'s **module doc has broken `main`'s
rustdoc twice within two days, by the same mechanism**:

1. #1920 left ``[`supervised::APPROVAL_REQUEST`]`` unresolvable → fixed by
   #1927, whose title is literally "unbreak main — resolve the boot.rs doc link
   #1920 left dangling".
2. #1939 left ``[`SkipReason::NoResumePoint`]`` unresolvable → fixed by
   PR #1985.

Both are `rustdoc::broken_intra_doc_links` in the same file's `//!` block. Both
merged green as PRs. Neither was caught until a *later, unrelated* PR paid for
it.

The current pile-up has four independent errors from three PRs:

| Error | Introduced by | Fixed by |
|---|---|---|
| `plan_stage` 8 args (`clippy::too_many_arguments`) | #1953 | PR #1985 |
| `[`SkipReason::NoResumePoint`]` unresolved (rustdoc) | #1939 | PR #1985 |
| `PassingShell` / `shell_call_result` undefined | #1945 (test file without its helpers) | PR #1964 |
| `ModelCallRole::Research` match arm missing | #1953 | PR #1964 |

Note #1945: it merged a **new test file whose helpers were in a different PR**.
A green PR check could not have caught that either, because the helpers existed
on that PR's branch.

## Files

- `.github/workflows/ci.yml` — the trigger block; today it runs on
  `pull_request` (and ignores `docs/**`, `*.md`).
- `AGENTS.md` § "The gate — run before every push" — documents the three-workflow
[...]

Issue #1972, opened 2026-08-07
main is red on three axes: stella-pipeline tests don't compile, clippy arity, and a private intra-doc link


macanderson:
## Problem

`origin/main` at `43402ae` fails **three** of the gate's steps. Every PR branched from it inherits a red `fmt + clippy + test` job that has nothing to do with its diff.

Reproduce from a clean checkout, nothing else applied:

```bash
git worktree add /tmp/mainchk origin/main --detach && cd /tmp/mainchk
cargo test -p stella-pipeline --no-run     # (1)
cargo clippy -p stella-pipeline            # (2)
RUSTDOCFLAGS="-D warnings" cargo doc -p stella-cli --no-deps   # (3)
```

### (1) The test suite does not compile — #1945 shipped a test referencing fixtures it never added

```
error[E0425]: cannot find value `PassingShell` in this scope
  --> crates/stella-pipeline/src/pipeline/tests/verification_hardening/flip_halt_arming.rs:42:21
error[E0425]: cannot find function `shell_call_result` in this scope
  --> crates/stella-pipeline/src/pipeline/tests/verification_hardening/flip_halt_arming.rs:31:9
```

Two use sites, zero definitions:

```console
$ git grep -E 'PassingShell|shell_call_result' origin/main -- 'crates/stella-pipeline/*'
origin/main:.../flip_halt_arming.rs:31:        shell_call_result(witness_command),
origin/main:.../flip_halt_arming.rs:42:        .with_tools(PassingShell)
```

Plus a semantic merge conflict in the same build:

```
error[E0004]: non-exhaustive patterns: `stella_protocol::ModelCallRole::Research` not covered
  --> crates/stella-pipeline/src/management_prompt/tests.rs:64:11
```

`management_system_block` is **deliberately** exhaustive (its doc: "a new role routed through the chokepoint must declare its prefix posture here, not escape the family witness by omission"). #1941 added the match; #1953 later added `ModelCallRole::Research`. Neither PR is wrong alone — the match is doing its job — but nothing made #1953 answer it.

### (2) Clippy fails workspace-wide — the same #1953 argument

```
error: this function has too many arguments (8/7)
    --> crates/stella-pipeline/src/pipeline.rs:1458:5   (plan_stage)
```

#1953 added `research: &[ResearchFinding]` to `plan_stage`, taking it from 7 parameters to 8. `make lint` runs at `-D warnings`, so this is an error.

### (3) rustdoc fails — a link to a private item

```
[...]

PR #1979, opened 2026-08-07
fix(stella-core,stella-store): journal compaction rewrites so 'context sent' shows the bytes the model received (#1667)

macanderson:
## What

Compaction rewrites tool results **in place** (eviction, dedup, supersession, aging stubs) but journaled only *identities*, never the replacement bytes. The journal kept only the pre-compaction output under that `call_id`, so `stella inspect · context sent` rendered bytes the model never received — a correctness gap in the one tool whose entire purpose is answering "what did the model actually receive?" (#1667, split out of #1668 which fixed only the misleading WARN wording).

Three changes, one seam each:

1. **`stella-protocol`** — a new `CompactionRewrite { block_id, content_digest, content }` payload in its own sibling file (`src/compaction_rewrite.rs`), carried by a new `rewrites` field on the existing `Compaction` variant. Additive: `#[serde(default, skip_serializing_if = "Vec::is_empty")]`, so old journals parse and turns that never compact pay nothing on the wire.
2. **`stella-core`** — `CompactionReport` gains `rewrites`. Pass 0 (retention aging) captures its records at the mutation site; passes 1–4 are recovered by one identity walk after pass 4, comparing each result against the `original_ids` snapshot the module already takes. Only *rewritten* results pay a hash. The driver forwards them onto the event at both emission sites.
3. **`stella-store`** — `JournalPreimages` gains a digest-keyed `rewrites_by_digest` map, and `resolve_content` consults it **before** the `call_id` route. `stella-observatory`'s independent reconstruction mirrors the same resolution order so the two cannot diverge.

## Why a field, not a new variant

The `Compaction` event already exists at exactly the two sites that perform the rewrites, and the replacement bytes are meaningless apart from the pass that wrote them. A new top-level variant would have forced a tag-table sweep plus eight downstream matchers (four compile-enforced, four silent — `event.rs`'s own checklist) for a payload that belongs to an event already being emitted. The *type* still lives in its own file, which is the split the god-file rule asks for.

Exemplar followed: `subagent_event.rs` / `ladder.rs` — payload struct in a sibling module, one variant line in `event.rs`, `pub use` in `lib.rs`.

## Witness tests

- `stella-store` — **`a_compacted_result_reconstructs_to_the_bytes_the_model_received`**. Verified the artisanal way by disabling only the digest lookup: it fails with `mismatches=["blk_post"]`, which is precisely the #1667 symptom — the block still *resolves*, just to the wrong bytes. It does not merely fail to find something.
[...]

PR #2000, opened 2026-08-07
fix(stella-pipeline,stella-protocol,repo): unbreak main — five breaks stacked behind one clippy error

macanderson:
## What & why

`main` was red at `e0fbbe02` on **five distinct breaks stacked behind one
another**. CI's log showed only the first, because both compile-tier gates
report one unit at a time: clippy stops at the first *hard* error in a crate,
and `cargo doc --workspace` stops at the first crate that fails to document.
That is why this is the sixth consecutive unbreak PR — each one can only reveal
the next layer. Root cause tracked in #1986 (`ci.yml` does not run on a push to
`main`); evidence from this session added there.

### 1. `pipeline/scope_stage.rs:34` — dead `spend` local

```
error: variable does not need to be mutable
error: unused variable: `spend`
```

`plan_with_review` binds `let mut spend = Spend { budget, total };` and never
reads it — the re-planning loop builds a fresh `Spend` by reborrowing on each
iteration, which is the only construction the code uses. Dead since #1971,
unmasked when #1985 cleared the `plan_stage` arg-count error above it.

### 2. `management_prompt/tests.rs` — `ModelCallRole::Research` listed twice

`unreachable_patterns`. Kept the documented placement beside `Unknown`, whose
comment explains why `Research` never reaches the chokepoint; dropped the copy
appended after `Summarization`.

### 3. `verification_hardening.rs` — three items nothing constructs

`dead_code` ×3 on `SHELL_TOOL`, `shell_call_result`, `PassingShell`. The child
`flip_halt_arming` module defines its own, which shadow the parent's through
`use super::*` — a glob import loses to a local definition silently, so this was
never a name clash, just quietly unreachable code.

The child's are the live pair *and* the newer one: a per-command
`call_id: format!("call-shell-{command}")` that `FlipHalt` correlates on, versus
the parent's fixed `"call-shell"` which cannot distinguish two shell calls. So
the parent's stale copies go. Its `mod` doc claimed the child existed in order
to reach the parent's fakes — the pre-split rationale, now false — and is
rewritten to point at the child's own doc, where the anti-clobber reason for
colocating them lives (#1997).

**#2 and #3 are the same shape**: a merge landed the same addition twice. Neither
side conflicts textually, so review saw nothing.

### 4. `stella-protocol/src/event.rs` — unresolved intra-doc link

`AgentEvent::Compaction::rewrites` documents itself with `[`CompactionRewrite`]`,
but `event.rs` never imports the type (the field spells it `crate::CompactionRewrite`
inline), so `broken_intra_doc_links` failed `doc-warnings`. A *different gate
step* from #1–#3, invisible while clippy was red. Fourth recurrence of the shape
#1986 tracks.

### 5. `file-size` — two ceilings exceeded on `main`

`driver.rs` at 2572/2571 and `pipeline/tests.rs` at 2537/2536. **Neither file is
touched by this branch**; both were grown on `main` by merges that did not
regenerate the baseline. That mechanism is #2004.

Also regenerated `docs/wire/*` — the protocol types' doc comments *are* that
contract, so break #4's fix mechanically changed the emitted `description`.

## About the two raised ceilings

A raised ceiling is normally a defect, so this is stated plainly rather than
buried: `make file-size-update` moved `driver.rs` and `pipeline/tests.rs` up by
one line each, for growth **this branch did not author**, because the growth has
already landed on `main` and reverting another PR's line is outside this task.
The alternative was leaving the gate red.

The same regeneration also **tightens** `pipeline.rs` from 3451 to 3181 —
a 270-line shrink the baseline had not captured. This branch adds no lines to
any god file. A maintainer who would rather see those two lines pushed into
submodules should say so; that is their call, not mine.

## The witness

- [x] No witness test. Four of the five are dead code, a duplicate match arm, and
      a doc link — no runtime behavior exists to witness, and the compiler is the
      oracle. The fifth is a generated baseline. Per CONTRIBUTING's carve-out for
      changes with no behavior delta, here is how it was verified instead:
  - `cargo clippy --workspace --all-targets -- -D warnings` — fails on `main` at
    break #1, exits 0 here.
  - `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps` — fails on
    `main` at break #4, exits 0 here.
  - `cargo fmt --all --check` — exits 0.
  - `cargo test -p stella-pipeline` — 618 pass, 0 fail. That suite **did not
    compile at all** on `main` (break #3), so these tests were not running.
  - `scripts/check-file-size.sh` and `make wire-schema` — both exit 0.

  Exit codes were read from cargo directly, not through a pipe: `cargo … | tail`
  reports *tail's* status, which is always 0, and cargo colorizes `error` so a
  plain `^error` grep matches nothing. Both produced a false green in this
  session before being corrected.

## The gate

- [x] `file-size` and `god-files` pass; baseline regenerated, never hand-edited.
- [x] `wire-schema` passes; the diff is comment-only — no field added, removed,
      renamed or re-tagged, and no optional field made required, so the
      additive-only contract holds.
- [x] No behavior change, no new flags, no new dependencies.

## Nothing left behind

- **#2013** (filed) — sharing `CARGO_TARGET_DIR` between worktrees produces
  compile errors naming symbols that do not exist. Hit during this work: a
  phantom `E0004` for `AgentEvent::TurnParked`/`TurnWoken`, variants present in
  neither checkout, because a parallel job's build was linked in. Nearly caused
  a wrong "fix".
- **#1986** — commented with the full five-layer breakdown as evidence for
  fixing the trigger rather than the instances.
- **#2004** — owns the file-size baseline skew behind break #5.

Refs #1972, #1986, #1997, #2004, #2013


macanderson:
Heads up — we built the same unbreak in parallel (mine: #TBD, branch `unbreak-duplicate-research-arm`), and independently reached the **identical** resolution on all three: keep one `Research` arm, keep #1985's per-iteration `Spend` reborrow over the hoisted `let mut spend`, and move the doubles into `flip_halt_arming` for the co-location reason. Good sign for the fix; sorry for the duplicated effort.

**Your `file size ratchet` check is red for a reason that is not yours:**

```
crates/stella-core/src/driver.rs               2572 vs 2571  (+1)
crates/stella-pipeline/src/pipeline/tests.rs   2537 vs 2536  (+1)
```

Both are already `main`'s state at e0fbbe02 — #1979 and #1962 grew those files without regenerating the baseline in the same commit. Neither of us touched either file, and it fails on **every** open PR until someone absorbs it. `make file-size-update` is the fix (it also tightens `pipeline.rs` 3451 → 3181, stale headroom now closed).

Add that one commit and yours is complete — happy to close mine. Only one of these should land; merging both is how the collision we are fixing happened in the first place.

macanderson:
Mine is #2008, for reference — same three fixes, plus the `make file-size-update` commit that unblocks the red `file size ratchet` check. Whichever of the two you prefer, please close the other.

macanderson:
Independent verification from a parallel session — **this PR is the one that unbreaks rustdoc, and I can confirm it works.**

I hit the same `doc-warnings` failure on main at 6c345532 and fixed it separately in #2010 before spotting that your `event.rs` hunk already covers it. Yours is the better form — `[`CompactionRewrite`](crate::CompactionRewrite)` keeps the short rendered label where mine (`[`crate::CompactionRewrite`]`) would print the fully-qualified path in the docs — so **I am closing #2010 in favour of this PR.** Nothing is needed from you.

Verified on a local tree of `origin/main` + this branch + #2003 + #2005, with the baseline regenerated via `make file-size-update` after merging:

| Gate | Result |
|---|---|
| `cargo clippy --workspace --all-targets -- -D warnings` | **clean** |
| `RUSTDOCFLAGS=-D warnings cargo doc --workspace --no-deps` | **clean** |
| `cargo fmt --all -- --check` | **clean** |
| `scripts/check-file-size.sh` | **OK, none grew** |

So the four open unbreak PRs — this one, #2003 (baseline skew), #2005 (fmt newline) — are jointly sufficient; main is green on none of them individually.

**One caution when this lands.** Merging this branch conflicts with main in `crates/stella-protocol/src/event.rs`, because #1994 added `TurnParked`/`TurnWoken` to that file after this branch forked. Resolving it by taking this branch's copy of the file wholesale silently deletes both new variants — I did exactly that in my first verification pass and got `E0599: no variant named TurnParked`. The resolution has to keep main's file and re-apply only the doc-link hunk. Worth updating the branch from main before merge so the resolution happens here rather than in someone's merge queue.

Also: merging this and #2003 in either order leaves `scripts/file-size-baseline.txt` conflicted. Regenerate it (`make file-size-update`) rather than hand-merging the hunks.

macanderson:

PR #1962, opened 2026-08-07
feat(stella-pipeline,stella-cli): resume re-enters the staged pipeline at the execute stage (#1671)

macanderson:
## What & why

A killed supervised pipeline run resumes today as a plain engine turn: it answers, honestly labeled unverified, but the verify/verdict stages never run — the one contract this repo cares most about, lost to a `kill -9`. This PR ships slice (1) of the restoration plan recorded on #1671/#1615: the common case, a kill mid-execute in an un-isolated run.

**Draft on purpose** — this touches the most-scrutinized area of the tree (the pipeline, the engine driver, the resume path), and the maintainer should read the design calls below before this becomes mergeable.

Refs #1671

### How it works

- **The frame gains a progress record.** `stella-pipeline` reports the facts it learns mid-run — triage's class, the goal, the plan and its cursor, the untracked-file before-image, the configured test command's baseline observation — through a new `ResumeFrameSink` port (mirrors `CheckpointSink`'s synchronous/infallible/best-effort posture; pipeline-owned so `stella-core` never learns pipeline shapes, invariant 1). The CLI's `ProgressSink` folds it into `PipelineFrame::progress` (`#[serde(default)]`, no `FRAME_VERSION` bump), riding the same checkpoint commit as before — one store, not two (invariant 6: same step-boundary cadence).
- **A resume-aware turn driver, in the engine.** `Engine::drive_restored_turn` (`stella-core/src/driver/resume.rs`) is the sibling of `run_turn_with_sender` for checkpoint-restored `TurnState`s, keeping every obligation — persist-on-Continue, discard-on-terminal, carried step cap, lifecycle, **and the turn-halt check the CLI's hand-rolled copy had silently dropped**. The CLI loop now delegates to it; two copies of this loop is exactly how one of them forgot an obligation.
- **The seam: `Pipeline::resume`.** `pipeline/resume_stage.rs` finishes the interrupted turn, walks the plan steps the crash never reached (numbered from the plan's true cursor), then flows into the ordinary `witness_on_demand` → `verify_candidate` → settle tail — the same code `run` uses, not a fork of the ladder. `PipelineResume::from_progress` validates up front; "restore approximately" is unreachable.
[...]

Issue #1883, opened 2026-08-06
gate: the spelled-out gate-step count is a shared cell two concurrent PRs always collide on


macanderson:
## Problem

`check-gate-parity.sh` checks two things about AGENTS.md and CONTRIBUTING.md: that every step in `GATE_STEPS` is **named**, and that the spelled-out **total** matches. The first is derived and cannot drift. The second is hand-written prose, and any two concurrent PRs that each add a guard will collide on it.

This happened twice on 2026-08-06:

| | |
|---|---|
| `module-reachability` (#1750) | updated both documents to "twenty-four" |
| `self-driving-test` | updated both documents to "twenty-four" |

Both were correct against their own base. The second to merge left `GATE_STEPS` at 25 with the prose saying 24, and `docs guards` went red on `main` — which fails **every open PR**, not just the two involved.

Fixed by hand in the PR that closes this, but the class is intact: the next two concurrent guard-adding PRs reproduce it exactly.

## Why the names do not have this problem

`for step in $steps; do grep -qF "$needle" "$doc"; done` — each step is checked independently, so two PRs adding different steps produce diffs that merge cleanly. Only the total is a single shared cell that both branches must write.

## Fix directions

Either removes the class; the choice is a maintainer's call about what the prose is for.

1. **Drop the total from the prose.** "Or just `make gate`, which runs them in order." The list is already checked name-by-name, so the number tells a reader nothing they can act on — and it is the only part that can be wrong. Smallest change, and it deletes the guard clause too.
2. **Derive the word.** Have the documents carry a marker (`<!-- gate-step-count -->`) that a `make` target fills, the way `docs/manifest.json` is generated rather than hand-kept. Keeps the number, moves it out of the merge path.
3. **Make the guard repair it.** `check-gate-parity.sh --fix`, mirroring `make doc-links-fix`, so a collision is a one-command fix instead of a hunt. Does not prevent the red, only shortens it.

Recommend (1) unless someone actively wants the count in prose — the third option still leaves `main` red until somebody runs it.

## Files

- `scripts/check-gate-parity.sh` — `number_word()` and the count comparison (the two `does not spell the gate's step count` branches)
- `AGENTS.md` — "That is twenty-five steps, and the list is not maintained by hand"
- `CONTRIBUTING.md` — "Or just `make gate`, which is the twenty-five of them in order."

## Verify

[...]

Issue #1782, opened 2026-08-06
Unbreak main: #1776 grew driver.rs (+16) and registry.rs (+35) past their god-file ceilings — file-size gate red


macanderson:
## Problem

`main` at `57e8f51d` (PR #1776, sibling task-call concurrency) fails `make file-size`:

- `crates/stella-core/src/driver.rs` — 2705 lines, ceiling 2689 (+16)
- `crates/stella-tools/src/registry.rs` — 2249 lines, ceiling 2214 (+35)

Every open PR now shows this failure (the red-main pattern, #1645). The growth is real feature code, not blessing debt: driver.rs gained the `dispatch_safe_tools` union at the dispatch site plus widened doc comments on `execute_tool_calls`' grouping; registry.rs gained `Tool::parallel_safe` and `ToolRegistry::parallel_safe_names`.

## Fix direction

Split, don't raise: per CLAUDE.md a ceiling raised to turn a gate green is a defect. Candidates:

- registry.rs: move the `parallel_safe_names` aggregation (and any other coherent `impl ToolRegistry` read path) into a sibling submodule, the way `crates/stella-core/src/driver/settlement.rs` was split out of driver.rs.
- driver.rs: the +16 is doc-comment growth plus a 2-line union at the one dispatch site — a sibling extraction needs a judgment call about which coherent block leaves (the dispatch grouping helper `execute_tool_calls` itself is a candidate).

If a truly irreducible remainder is left, `make file-size-update` with the baseline diff justified in the PR is the documented escape hatch — for the remainder only.

## Verify

`make file-size` green at the tip of the fix branch; `cargo test -p stella-core -p stella-tools` green (the #1776 feature has tests covering the grouping — do not change behavior).

## Constraints

- Re-verify main still fails right before merging — a peer unbreak may land first (the duplicate-unbreak hazard).
- `scripts/file-size-baseline.txt` is generated; god-file tables in AGENTS.md / crate READMEs are checked by `scripts/check-god-files.sh` — a split that removes a file from the baseline must run `make file-size-update` so all copies agree.

macanderson:
[...]

Issue #1761, opened 2026-08-06
main fails the file-size gate: deck.rs is 1517 lines against a 1510 ceiling (#1742)


macanderson:
## Problem

`main` fails the `file-size` gate step. `crates/stella-tui/src/deck.rs` is **1517 lines against its recorded ceiling of 1510** in `scripts/file-size-baseline.txt`:

```
$ ./scripts/check-file-size.sh
check-file-size: FAILED
  crates/stella-tui/src/deck.rs grew to 1517 lines, over its baseline ceiling of 1510 (+7)
```

This reproduces on a clean `origin/main` — it is not a merge artifact:

```
$ git show origin/main:crates/stella-tui/src/deck.rs | wc -l
1517
$ git show origin/main:scripts/file-size-baseline.txt | rg 'deck\.rs'
1510 crates/stella-tui/src/deck.rs
```

The growth came in with **#1742** ("fix(stella-tui): keep the file ledger's diffs across a /clear"), which added lines to `deck.rs` without the accompanying baseline diff. Every PR branching off `main` now inherits a red gate — I hit it on #1757 (the `fullauto` → `self-driving` rename), which touches no `stella-tui` file at all.

## Do NOT just run `make file-size-update`

`deck.rs` is a **grandfathered god file, closed to growth** (AGENTS.md § "God files — plan around them, never into them", and `crates/stella-tui/README.md`). CLAUDE.md is explicit that raising a ceiling to turn a gate green is a defect, not a tradeoff. So the two acceptable outcomes are, in order of preference:

1. **Move #1742's new lines into a sibling module** so `deck.rs` returns to ≤1510. `crates/stella-core/src/driver/settlement.rs` (split out of `driver.rs`) is the pattern the repo already follows. `stella-tui` has the room: `deck_ui.rs`, `deck_render.rs`, and `views/` are all established neighbours.
2. **If the 7 lines are genuinely irreducible**, run `make file-size-update` and land the baseline diff as a reviewable change with a sentence in the PR description justifying it — which is what the escape hatch is for. Silently ratcheting the ceiling is the outcome to avoid.

## Files

- `crates/stella-tui/src/deck.rs` — the file over its ceiling
- `scripts/file-size-baseline.txt` — the generated, gate-enforced ceiling list (the only copy of the numbers; do not restate them in prose)
- `scripts/check-file-size.sh` — the guard
[...]

PR #2003, opened 2026-08-07
fix(repo,stella-pipeline,stella-protocol): unbreak main — file-size skew, a dead binding, and a broken doc link

macanderson:
## What & why

`main` is red on **three independent gate steps**, so every open PR inherits
failures it did not cause. This PR is the smallest reviewable change that turns
all three green. None of them is reachable from the others, which is why they are
here together rather than in three PRs — landing one still leaves `main` red.

| Step | Failure | Cause |
|---|---|---|
| `file size ratchet` | 2 god files +1 over ceiling | baseline skew between #1979 and #1962 |
| `cargo clippy -D warnings` | dead `spend` binding | leftover from the per-iteration reborrow refactor |
| `cargo doc -D warnings` | unresolved intra-doc link | item re-exported at the crate root, link unqualified |

### 1. `file size ratchet` — a parallel-merge baseline skew

```
crates/stella-core/src/driver.rs grew to 2572 lines, over its baseline ceiling of 2571 (+1)
crates/stella-pipeline/src/pipeline/tests.rs grew to 2537 lines, over its baseline ceiling of 2536 (+1)
```

**Neither contributing PR did anything wrong.** #1979 added three lines to
`driver.rs` and #1962 added a line to `pipeline/tests.rs`, and each regenerated
`scripts/file-size-baseline.txt` on top of a `main` that did not yet carry the
other's growth. Both were green on their own merge commits; the *composition* is
red. This is the repository's most common cause of a red `main`, and it is why
the baseline is generated rather than edited — hand-patching the two visible
numbers fixes today's symptom and leaves the next skew just as invisible.

Regenerated with `make file-size-update` on top of current `main` — the whole
file, not the failing lines — so the result is reproducible by re-running the
command rather than a set of numbers someone picked. The diff moves in **both**
directions, and the tightenings are the larger half:

| File | Before | After | Δ |
|---|---|---|---|
| `crates/stella-core/src/bus.rs` | 2126 | 1891 | **−235** |
| `crates/stella-pipeline/src/pipeline.rs` | 3451 | 3181 | **−270** |
| `crates/stella-core/src/driver.rs` | 2571 | 2572 | +1 |
| `crates/stella-pipeline/src/pipeline/tests.rs` | 2536 | 2537 | +1 |

The two reductions are #1994's `bus/names.rs` split and the `pipeline.rs`
extraction finally reaching the ledger — work the same skew had been hiding. The
ratchet had been holding those two ceilings **505 lines looser than the tree
actually needs**, which is the direction that quietly readmits bloat.

The two raises are **+1 apiece against code already on `main` behind a review** —
precisely the irreducible case the escape hatch is documented for (AGENTS.md §
"God files"). No new baseline entry, and no file grandfathered that was not
already. No entry dropped below the 1500-line limit either, so the god-file
tables in `AGENTS.md` and the crate READMEs are untouched.

### 2. `cargo clippy -D warnings` — a dead binding

```
error: variable does not need to be mutable
error: unused variable: `spend`
  --> crates/stella-pipeline/src/pipeline/scope_stage.rs:34:13
```

`let mut spend = Spend { budget, total };` is the binding's **only** occurrence
in the file. The loop immediately below deliberately constructs a *fresh*
reborrow on each iteration, and its own comment says why:

> Reborrowed per iteration: the loop replans after a rejected scope card, and a
> moved `Spend` could not be handed to the next attempt.

So the outer binding is leftover from that refactor, not a value the loop
shadows. **Deleted rather than underscore-prefixed** — `_spend` would keep a dead
constructor alive and read as deliberate to the next person.

### 3. `cargo doc -D warnings` — an unqualified intra-doc link

```
error: unresolved link to `CompactionRewrite`
  --> crates/stella-protocol/src/event.rs:545:53
```

The type is re-exported at the crate root (`lib.rs:75`) but is not in the `event`
module's scope. The field two lines below the doc comment already spells it
`crate::CompactionRewrite`; only the link was unqualified. Qualified it to match,
so the link and the field now name the same item by the same path.

## The witness

Not applicable, and deliberately so — this is a build-artifact regeneration plus
two one-line corrections to already-reviewed code, with no behavior change: the
deleted binding had no reader, and a doc link is not code. **The gate steps are
the test**, and all three flip fail → pass.

The ratchet flip was verified locally on a branch cut straight from
`origin/main` with no other change applied:

```
$ bash scripts/check-file-size.sh          # before
check-file-size: FAILED
  crates/stella-core/src/driver.rs grew to 2572 lines, over its baseline ceiling of 2571 (+1)
  crates/stella-pipeline/src/pipeline/tests.rs grew to 2537 lines, over its baseline ceiling of 2536 (+1)

$ bash scripts/check-file-size.sh          # after
check-file-size: OK — 1129 Rust/Python/shell files, none over 1500 lines except 30 grandfathered (none grew).

$ bash scripts/check-god-files.sh
check-god-files: OK — 22 god file(s) across 7 crate(s), named identically in AGENTS.md and every crate README.
```

That same probe branch is how the ratchet failure was isolated from PR #1992's
diff — #1992 is byte-identical to `main` for both the baseline and both failing
files, so the red was `main`'s, not its.

**The clippy and rustdoc fixes were verified in CI, not locally, and that was a
deliberate choice.** Three Terminal-Bench runs are live on this machine right now,
including a Stella-vs-Claude-Code head-to-head. A full workspace clippy + rustdoc
compile would have contended for CPU with a measured benchmark and skewed its
wall-clock numbers. Per CLAUDE.md's "measure honestly" rule, a slower verification
path is the correct trade against corrupting a benchmark this project reports in
public. The cheap guards above compile nothing, which is why they were safe to run.

## Ground-rule check

- [x] No new dependencies; no I/O added to `stella-core`
- [x] Baseline **regenerated**, never hand-edited
- [x] No new baseline entry; no ceiling raised beyond already-merged code
- [x] No `#[allow]` used to silence either lint — both were real defects
- [x] No behavior change, so no serde round-trip or parity-matrix impact

## Nothing left behind

- [x] Filed: #2004

**#2004** — the ratchet has no defense against this skew, and this is its **third
occurrence** (#1761, #1782, now this). Two PRs can each regenerate the baseline
correctly against different snapshots of `main` and compose into a red tree;
nothing detects it until the next push pays for it. The proposal is to make the
guard judge *the change* rather than *the tree* — a file already over its ceiling
at the merge base must not fail a PR that did not grow it. AGENTS.md already
rejects the shared-cell design for `GATE_STEPS` counts (#1883) for exactly this
reason; the baseline has the same shape and never got the same treatment.

This PR deliberately does **not** attempt that fix. `main` is red right now, and
an unbreak should be the smallest reviewable thing that turns it green.

Related to #1986 — `ci.yml` does not run on a push to `main`, which is why all
three of these survived on `main` rather than being caught at merge time.

**This unblocks #1992**, the only other open PR, which is `MERGEABLE/BLOCKED`
solely on these checks. Its own merge conflict is already resolved (`origin/main`
is an ancestor of its head); GitHub had simply not recomputed the stale
`CONFLICTING` flag. It will need `main` merged in after this lands.


macanderson:
Verified from a parallel session: this PR is one of **four** independently-red gates on main at 6c345532, and all four must land. The others are #2000 (rustdoc + clippy) and #2005 (fmt newline).

On a local tree of `origin/main` + this branch + #2000 + #2005, all four gates come back clean — `clippy --workspace --all-targets -D warnings`, `RUSTDOCFLAGS=-D warnings cargo doc --workspace --no-deps`, `fmt --check`, and `check-file-size.sh`.

**Ordering caution, since this PR owns the baseline.** Merging this alongside #2000 leaves `scripts/file-size-baseline.txt` conflicted — whichever lands second inherits it. The fix is to regenerate on the rebased tree (`make file-size-update`), never to hand-merge the hunks: a hand-resolved baseline is how the current skew started, and the file is generated precisely so it has one authority. In my verification I had to regenerate after merging, and the regenerated file differs from either side's version.

macanderson:

Issue #1857, opened 2026-08-06
Parked waits: dedicated wire events (TurnParked/TurnWoken) and a TUI heartbeat instead of synthetic Text


macanderson:
## Problem

The parked-wait primitive (#1471) announces the park, the refusal, and the wake as synthetic `AgentEvent::Text` deltas (`crates/stella-core/src/driver/waiting.rs`). That was the deliberate slice-1 choice: `stella-protocol/src/event.rs` is a god file closed to growth and every new variant restales `docs/wire` (the wire-schema gate), so the event surface was deferred rather than rushed.

Text deltas work but are second-class: consumers cannot distinguish a park from model narration, the deck cannot render a live 'waiting 240s/1800s' heartbeat line, replay cannot skip or summarize the parked span, and `stella inspect` cannot attribute the wall-clock gap.

## What done looks like

- A typed event pair — either `AgentEvent::TurnParked { description, poll_interval_secs, deadline_secs }` / `TurnWoken { reason, polls_used }` next to `Steered` (event.rs:363 area), or a minimal `StageKind::Waiting` if that proves sufficient. Payload struct can live in a sibling file to keep event.rs growth to the thin variant.
- Wire-name table rows + serde round-trip test (invariant 4); regenerate `docs/wire` (`scripts/export-agentevent-schema.sh` / the wire-schema gate).
- Bus lifecycle names `agent.turn.parked` / `agent.turn.woken` beside `AGENT_STEP_STARTED` in `crates/stella-core/src/bus.rs` (observer-only; do NOT touch `names::BLOCKING` — its length is pinned by `blocking_set_is_exactly_the_documented_nine`).
- The deck renders a parked state (heartbeat line or status chip) from the new events; `driver/waiting.rs` stops emitting the ⏳/▶ Text deltas once a renderer exists.
- The existing witness `crates/stella-core/src/driver/tests/parked_wait.rs` updated to assert the typed events instead of Text contents.

## Constraints

- event.rs and bus.rs are god files at their ceilings — plan line offsets or a justified `make file-size-update` (AGENTS.md § God files).
- Type doc comments on protocol structs ARE the wire contract — a `///` change restales docs/wire; run the schema exporter in the same PR.
- The park loop itself must not change behavior: this is an observability issue, not a semantics one.

Refs #1471.

macanderson:
[...]

PR #2005, opened 2026-08-07
fix: unbreak main's fmt gate, and stop the parked span vanishing from the observatory journal and fleet row (#1857 follow-up)

macanderson:
Follow-up to #1994 (which merged while this was being written). Two independent things, the first urgent.

## 1. Unbreak main — `cargo fmt --check` is red

#1994 left `crates/stella-protocol/src/event/tests.rs` without a trailing newline, so `cargo fmt --check` fails on `main` and **every open PR inherits the red**. One character.

My break, and worth recording how it escaped: the `mod tag_table;` declaration was appended by a heredoc *after* that PR's rustfmt pass, so the local gate had already run over the file in its earlier shape and reported green.

## 2. Three wildcard consumers where the park silently regressed

Adding `TurnParked`/`TurnWoken` broke five crates one at a time — the compiler caught those. It could not catch the consumers matching `AgentEvent` behind a wildcard, and because the park used to arrive as `Text`, all three silently got **worse** when it became its own variant:

- **`stella-observatory` `execution_journal`** selects an `event_type` allowlist. `'text'` was on it; `'turn_parked'`/`'turn_woken'` were not — so the park **disappeared entirely** from the transcript an operator opens to ask why an execution took so long. It is the one event that explains a wall-clock gap containing no other events.
  - Added to the allowlist, plus `journal_entry` arms carrying the payload (a row saying a park happened but not what was awaited or for how long answers nothing), plus a frontend arm in `assets/index.html` — without one it fell through to the `"answer"` branch and drew a blank row.
- **`stella-tui` fleet dashboard** froze `row.action` on the last tool, so a worker deliberately waiting read as one stuck mid-tool. New `LastAction::Parked` rather than reusing `Blocked`: `Blocked` means *a human must act*, and a park needs nobody — mislabelling it would send an operator hunting an approval prompt that does not exist. `TurnWoken` deliberately holds the park rather than clearing to `Idle`, since the next tool or message repaints a beat later and clearing would only flicker.

## Deliberately not done

- **`stella-serve`'s `TallyFold`** (`observe/tally.rs`) drops both variants into `_ => {}`. Not a regression (the park was already dropped as `Text`), but its own doc says a turn whose stages stop advancing is *wedged* — and park/wake are exactly the signal distinguishing a deliberate wait from a hang. Filed rather than bundled, since it means adding a field to `TurnTally`.
- **`TENDENCY_EVENT_TYPES`** excludes park/wake. Judged correct, not accidental: every other member is a defect or corrective signal (retry, loop, budget denial, fallback); a park is normal intended operation.

## Note on the `file-size` gate

`main` is *also* red on `file-size` — `driver.rs` and `pipeline/tests.rs` are each +1 over ceiling from a parallel-merge skew. **Neither file is touched here**, and #2003 is already the unbreak for it, so this PR deliberately does not duplicate that baseline regeneration.

Refs #1857

## Summary by Sourcery

Unbreak the main branch formatting gate and update observability surfaces so parked and woken spans are correctly included and represented in operator-facing views.

Bug Fixes:
- Restore display of parked and woken spans in the observatory execution journal and fleet dashboard so operators can see and understand deliberate waits.
- Fix the main branch formatting gate by restoring the trailing newline in stella-protocol event tests.

Enhancements:
- Extend observatory journal payload handling and frontend rendering to show detailed context for turn_parked and turn_woken events.
- Add a distinct Parked last-action state in the TUI fleet dashboard to accurately represent engine-side waits without mislabeling them as human-blocked.


macanderson:
**CI red here is entirely inherited from `main` — none of it is in this diff.** Verified locally on a *pristine* `origin/main` checkout, so a reviewer does not have to re-derive it:

| Failure | On pristine `main`? | Already fixed by |
|---|---|---|
| `file size ratchet` — `driver.rs` +1, `pipeline/tests.rs` +1 | yes | #2003, #2008, #2009 |
| `cargo clippy -D warnings` — dead code in `pipeline/tests/verification_hardening.rs`, unused `spend` | yes (`exit 101`) | #2000, #2008, #2009 |
| `cargo doc -D warnings` — unresolved intra-doc link to `CompactionRewrite` (`event.rs:542`) | yes | #2010 |
| `cargo fmt --check` — missing trailing newline, `event/tests.rs` | yes | **this PR — nothing else fixes it** |

Files this PR touches: `stella-observatory/src/db.rs`, `stella-observatory/src/assets/index.html`, `stella-tui/src/fleet_dashboard.rs`, `stella-protocol/src/event/tests.rs`. None of the three inherited failures is in any of them.

The fmt row is the one that matters for sequencing: it is my break from #1994, no other open PR carries the fix, and it is red on `main` and therefore on **every** open PR until this lands.

One methodology note worth recording, since it is why I nearly missed this: I first checked clippy with `cargo clippy … | rg "^error"` and read the empty result as green. Cargo's output carries ANSI escapes before the word `error`, so `^error` never matches — the pipeline lies. Checking `$?` instead gives `101`.

macanderson:

macanderson:
**Superseded by #2015 — this can be closed.**

#2015 now contains every commit from this branch, byte-for-byte (it picked them up from `fix-1857-parked-wait-consumers`):

- the `event/tests.rs` trailing-newline fmt fix — identical blob (`30d80d2d..72cfda5e`);
- the observatory journal allowlist + `journal_entry` arms + the `assets/index.html` frontend arm;
- `LastAction::Parked` and the fleet-dashboard fold arms, comments included.

…and it additionally carries the pipeline clippy fixes and a `CompactionRewrite` doc-link fix **with** the `docs/wire` regen that fix requires (see my note on #2010, which makes the same edit *without* the regen and so fails `wire-schema`).

So #2015 is the strictly-larger, correct one. Landing both risks the duplicate-unbreak collision this repo has already hit twice; landing #2015 alone loses nothing from here.

Leaving the close to a human rather than doing it myself, since I cannot tell from outside whether anything else is sequenced behind this branch.

**The one thing that must not get dropped in the shuffle:** the fmt fix is the only repair for a break I introduced in #1994, and `cargo fmt --check` stays red on `main` — and therefore on *every* open PR — until it lands, in whichever PR carries it.

PR #1953, opened 2026-08-07
feat(stella-pipeline): pre-plan research stage — triage names questions, parallel read-only sub-agents answer them (#1778)

macanderson:
Closes #1778. Refs #1768, #1776 (the first two slices: triage evidence, model-side sibling-spawn concurrency).

## What this builds

The decisions with the highest downstream leverage — what the plan names — were taken at the point of least evidence: nothing between triage and execute could read a file. This adds the pre-plan research stage from the issue's validated design sketch:

1. **Triage names the questions.** `TRIAGE_INSTRUCTIONS` gains an optional `RESEARCH:` line (`multi` only, up to 4 `|`-separated self-contained questions); `triage::parse_research_questions` parses it with `parse_flag`-style tolerance. Absent line / `none` / any class below the *resolved* `MultiStep` ⇒ no questions ⇒ the stage is skipped byte-for-byte (L-E2 untouched). Gating on the resolved class keeps the deterministic floor authoritative.
2. **The pipeline fans the questions out as parallel read-only sub-agents** (`pipeline/research_stage.rs`). Exemplars followed exactly: dispatch is `Engine::run_sub_agent_with_sender` + `SubAgentSpec` (the `stella-core::goal.rs` verifier's shape — so the pause gate, soft stop, and read-only enforcement come from the primitive, not reimplementation), money is `candidate_fanout::FanOutBudget` (pre-dispatch gate, `headroom / width` carve, settle in completion order, index-order cost summing). `write_access: false`, depth 1, per-child receipt slots. A per-child latency ceiling (`research_latency_ceiling`, default 45s, inside the future so the sub-agent drop guard still settles spend) guarantees research can never wedge a turn; a timed-out child's `Finished` bracket is emitted by the stage so the event stream stays balanced.
3. **Findings feed the planner as their own section.** Bounded like recall (`research::bound_research_findings`: 4k chars/finding, 12k total, honest drop markers), rendered as `## Research findings` in `build_planner_prompt` — never mixed into recall frames (different provenance). Instruction blocks stay `&'static str` (#1434).
4. **Attribution.** `ModelCallRole::Research` + `StageKind::Research` (wire schema regenerated, round-trip pinned); `stage_rank` places Research between ContextRecall and Plan; every exhaustive consumer match (TUI, diag bridge, wire-contract samples) updated.

## God-file accounting (all three were at their exact ceilings)

[...]

Issue #1793, opened 2026-08-06
FlipHalt never arms on the authored-witness path, and revise turns pass None even for configured commands


macanderson:
## Problem

`crates/stella-pipeline/src/flip_halt.rs` (mid-turn early stop when the tracked test goes fail→pass; built from a measured Terminal-Bench regression) is armed only from the configured `--test-command` baseline (`pipeline.rs`, `effective_test_command(None)` before execution). Two gaps:

1. **Authored witnesses never arm it.** `witness_on_demand` seeds the oracle with the witness command after execution, but `state.flip_halt` stays whatever it was (None on the authored path). Every revise turn thereafter runs to its step/loop limits even when the witness has already flipped mid-turn.
2. **`revise_turn` passes `None` unconditionally** (`pipeline.rs::revise_turn`, the `run_engine_turn(.., None)` call) — even a configured-command run's revisions do not halt early. If deliberate (a revision must finish its cleanup?), it is undocumented; if not, it is free wall-clock and tokens.

## Fix direction

Decide 2 first (it determines whether 1 has any effect): if revise turns should halt on a flip, arm `state.flip_halt` from the witness command after `witness_on_demand` succeeds (`FlipHalt` with the witness's normalized command) and pass `state.flip_halt.clone()` in `revise_turn`. Measure on the loop-bench arm before defaulting.

## Verify

Witness: scripted revise turn whose bash tool reports the witness command exiting 0 halts at the step boundary (TurnHalt reason) instead of running to the step cap.

## Related

`flip_halt.rs` module doc (the measured regression), #1524 (loop-abort scoring), the nightly loop-bench gate.

macanderson:
**Triage 2026-08-06:** P1 — the measured early-stop fix silently never applies on the authored-witness path, so every revise turn runs to its step/loop caps burning wall-clock and tokens after the witness has already flipped. Plan: as specced in the body; next step is deciding sub-question 2 (should revise turns halt on a flip at all) since it gates whether arming `state.flip_halt` from `witness_on_demand` has any effect, then measuring on the loop-bench arm before defaulting.

Issue #1667, opened 2026-08-06
Compaction rewrites are never journaled, so 'context sent' shows the model bytes it never received


macanderson:
## Problem

Compaction rewrites tool results in place (eviction, dedup, supersession, aging
stubs) but **emits no event carrying the replacement bytes**. The journal keeps
only the pre-compaction output under that `call_id`, so replay recovers a real
result that is simply not the one the step actually sent.

Two consequences, one cosmetic and now fixed, one real and not:

1. ~~`inspect · context sent` reported every such block as "journal torn or
   altered" in DANGER red~~ — fixed by #1668, which rewords the four surfaces to
   name the benign cause and drops the styling to WARN.
2. **`inspect` still displays bytes the model did not see.** Reconstruction
   falls back to the closest preimage the journal holds. For any compacted
   block, the "context sent" view is showing the *pre-compaction* output. That
   is a correctness gap in the one tool whose entire purpose is answering "what
   did the model actually receive?"

## Files

- `crates/stella-store/src/reconstruct.rs` — the digest-mismatch path that
  falls back to the closest preimage
- `crates/stella-cli/src/inspect.rs` — the CLI surface
- `crates/stella-tui/src/deck_render.rs`, `crates/stella-tui/src/envelope.rs` —
  the overlay surfaces
- Compaction itself lives in `stella-core` (no I/O — the event has to be emitted
  by whoever owns the bus, not by the decision function)

## Reproduce

1. Run any turn long enough to trigger compaction (step ~50 of an ordinary turn
   produced ~24 affected blocks in the original report).
2. `stella inspect` → `context sent` on a step after the first compaction.
3. Blocks whose results were rewritten show the WARN mismatch line, and the body
   rendered is the pre-compaction output.

## Definition of done

- A compaction preimage event carries the replacement bytes onto the journal
  when a result is rewritten in place.
- `reconstruct.rs` resolves a block by **digest-keyed preimage lookup**, so the
  bytes shown are the ones whose digest the receipt recorded.
- The mismatch surface then fires only when the digest genuinely cannot be
  resolved — i.e. it becomes a real integrity signal again, and can arguably go
  back to DANGER.
- A witness test: a turn with a compacted result reconstructs to the
  *post*-compaction bytes, failing on the old code.

## Constraints

- **Invariant 2 (no I/O in the engine)** — the compaction decision stays a pure
  function in `stella-core`; emitting the event is the caller's job.
[...]

PR #2008, opened 2026-08-07
fix(stella-pipeline,scripts): unbreak main — two unbreak PRs fixed the same three things, plus the file-size ratchet blocking every PR

macanderson:
> ⚠️ **Overlaps #2000 — merge exactly one of these, never both.** We built the same unbreak in parallel and reached the *identical* resolution on all three collisions. This PR additionally fixes a fourth break (the file-size ratchet) that is currently failing #2000's checks and every other open PR. If #2000 picks up that one commit, close this; otherwise close #2000. Merging both is how the collision being fixed here happened.

## Why main is red

`main` at e0fbbe02 fails `cargo clippy -p stella-pipeline --all-targets -- -D warnings` **and** `file size ratchet`.

Two unbreak PRs (#1985, and #1971 via #1995) independently repaired the breaks #1953 left, **converged on the same designs**, and landed minutes apart. Git merged them additively rather than conflicting, so main now carries each fix twice:

| # | Break | Where |
|---|---|---|
| 1 | `ModelCallRole::Research` twice in one or-pattern (`unreachable_patterns`) | `management_prompt/tests.rs` |
| 2 | Both a hoisted `let mut spend` **and** a per-iteration reborrow (`unused_variables` + `unused_mut`) | `pipeline/scope_stage.rs` |
| 3 | `PassingShell`/`shell_call_result` restored into *both* the parent and its child (`dead_code` ×3) | `tests/verification_hardening.rs` |
| 4 | Two grandfathered files one line over their ceiling | `scripts/file-size-baseline.txt` |

## The judgment calls

**(2) — kept #1985's per-iteration reborrow, not the hoisted binding.** Not arbitrary: `plan_with_review` loops, replanning after a rejected scope card, and only a `Spend` reborrowed per attempt survives that. The hoisted version would have been moved on the first iteration.

**(3) — kept #1985's doubles, in the child.** The two copies were *not* equivalent. #1985's are better documented: they name `SHELL_TOOL` as a const distinct from `WRITING_TOOL`, and say why the trailing `[exit code: 0]` marker is load-bearing — without it `FlipHalt::observe` never latches and the arming test passes for no reason. Those are the ones kept. They live in `flip_halt_arming` with both #1793 witnesses, because co-location is what turns the next wholesale rewrite of the parent into a merge conflict instead of the silent deletion that started this (#1997). The parent's `mod` doc is corrected in place rather than left describing a layout that no longer holds.

**(4) — recording growth that already merged, and saying so.** Two ceilings go **up** by one line each:

```
crates/stella-core/src/driver.rs               2571 → 2572
crates/stella-pipeline/src/pipeline/tests.rs   2536 → 2537
```

Per CLAUDE.md, a raised ceiling to turn a gate green is normally a defect against the PR that raises it, so this is flagged rather than slipped through. The difference: **this branch touches neither file.** Both grew on main via #1979 and #1962, which did not regenerate the baseline in the same commit. The choice is therefore not "grow or don't" but "record what already merged, or leave main red for everyone". The two lines are somebody's to reclaim; neither is mine to judge irreducible.

The same regeneration **tightens** `pipeline.rs` from 3451 to 3181 — 270 lines of stale headroom now closed off, which is the ratchet working as intended and more than offsets the two. Regenerated via `make file-size-update`, never hand-edited.

## Verification

- `cargo clippy -p stella-pipeline --all-targets -- -D warnings` — clean
- `cargo test -p stella-pipeline --lib` — **596 passed, 0 failed**, both #1793 witnesses among them
- `make guards-fast` — green, `file-size` and `god-files` included

## Related

- #1997 — why a deleted test failed no gate in the first place
- #1985, #1995, #2000 — the colliding unbreaks

## Summary by Sourcery

Unbreaks main by reconciling overlapping clippy and test fixes in stella-pipeline, consolidating flip-halt arming test doubles, and updating the file-size baseline so guards and ratchet checks pass again.

Bug Fixes:
- Resolve unreachable pattern warning in management_prompt tests by removing the duplicate ModelCallRole::Research arm
- Fix clippy unused variable warnings in scope_stage by relying on per-iteration Spend reborrows
- Restore and colocate shell tooling doubles for flip halt arming tests so dead-code warnings are cleared while preserving #1793 coverage

Enhancements:
- Clarify documentation and structure of flip halt arming tests by moving shared shell doubles into the child module and updating the parent module description

Build:
- Regenerate file-size baseline to reflect recent growth in driver.rs and pipeline tests while tightening the pipeline.rs ceiling so file-size ratchet gates pass again


macanderson:
Rebased onto `6c345532` and picked up a **fifth** break, landed by #1994 minutes ago and unrelated to the rest of this PR:

`crates/stella-protocol/src/event/tests.rs` is missing the trailing blank line rustfmt emits after `mod tag_table;`, so `cargo fmt --all --check` fails and takes the whole `fmt + clippy + test` job down — on **every** open PR, not just this one. Fixed with `cargo fmt --all`; the entire diff is one newline.

Also worth noting from the rebase: the file-size ratchet now passes **without** the two +1 ceiling raises being load-bearing any more — main's own baseline caught up in the interim. The regeneration commit stays because it also closes 270 lines of stale headroom on `pipeline.rs`, but the uncomfortable part of that commit message (two ceilings going up) no longer applies to the current merge base.

`make guards-fast` green, `cargo clippy -p stella-pipeline --all-targets -- -D warnings` clean.

macanderson:
Correction to my previous comment — I got the file-size half wrong.

Main's baseline has **not** caught up:

```
crates/stella-core/src/driver.rs               actual 2572, baseline 2571
crates/stella-pipeline/src/pipeline/tests.rs   actual 2537, baseline 2536
```

`check-file-size.sh` passes on this branch *because* the regeneration commit is in it, not because the skew resolved. So the two +1 ceiling raises **are** still load-bearing, and the disclosure in that commit message stands exactly as written: two grandfathered ceilings go up by one line each, recording growth #1979/#1962 already merged, in files this branch does not touch. Please read it as a real (if small) ratchet concession, not a no-op.

The rest of that comment holds: the fifth break (#1994's missing trailing newline) is real and fixed here, and `pipeline.rs` still tightens 3451 → 3181.

macanderson:
CI caught a **sixth** break, also not mine — and my local verification had a real gap that let it through.

`crates/stella-protocol/src/event.rs:545` documents the `Compaction` variant's `rewrites` field with `[`CompactionRewrite`]`, but that name is not in scope in `event.rs`: the field itself spells the fully-qualified `crate::CompactionRewrite`, and the type is only re-exported at the crate root. `RUSTDOCFLAGS="-D warnings" cargo doc` fails on it and takes the whole `fmt + clippy + test` job down. From #1979 (#1667).

Fixed by pointing the link at the path the field already uses — **the same repair #1985 applied to `boot.rs`'s `SkipReason::NoResumePoint`, which makes this the second time this exact shape has broken `main` in a day.** Worth a guard or a convention, not just a third fix.

**My gap:** I had been verifying with `make guards-fast` + `cargo clippy`. Neither runs rustdoc — `doc-warnings` is a `make check`/`make gate` tier — so a dangling doc link was invisible to everything I ran. Now verified with `make doc-warnings` across the whole workspace, which is clean. That matters more than the one fix: each repair of this kind only lets rustdoc reach the *next* dangling link, so a passing single-crate run proves nothing. A full-workspace run is the only evidence that there are none left.

Full `cargo test --workspace` running now; I'll post the result.

macanderson:
Full local verification, every tier CI runs — after the gap I described above, I stopped relying on the fast rungs:

| Tier | Result |
|---|---|
| `make guards-fast` (incl. `format-check`, `file-size`, `god-files`) | green |
| `RUSTDOCFLAGS="-D warnings" cargo doc --workspace` (`make doc-warnings`) | clean |
| `cargo test --workspace` | **123 test binaries, every one `ok.`, zero failures** |

That is `make gate`'s compile/test tiers in full. The workspace rustdoc and workspace clippy runs are the two that would have caught breaks 5 and 6 before CI did, and neither was in what I ran earlier.

Six breaks now fixed here, none of them originating in this branch:

1. duplicate `ModelCallRole::Research` arm — `unreachable_patterns`
2. duplicate `Spend` construction in `scope_stage.rs` — `unused_variables` + `unused_mut`
3. duplicate shell doubles across parent and child — `dead_code` ×3
4. two grandfathered files one line over ceiling — `file size ratchet`
5. missing trailing newline in `event/tests.rs` (#1994) — `cargo fmt --check`
6. dangling `[`CompactionRewrite`]` doc link in `event.rs` (#1979) — `rustdoc -D warnings`

1–3 are the #1985/#1995 collision; 4–6 are independent and each arrived from a different merge.

PR #1970, opened 2026-08-07
fix(stella-cli): unbreak main — unresolved SkipReason doc link in daemon::boot

macanderson:
## What & why

`main`'s workspace doc gate is red: `crates/stella-cli/src/daemon/boot.rs`'s module doc intra-doc-links `SkipReason::NoResumePoint`, but `daemon::boot` is a **private** module and `SkipReason` is `pub(super)`, so rustdoc resolves nothing and `rustdoc::broken_intra_doc_links` fails under `-D warnings`.

The link landed in #1939 and was invisible until now, which is the part worth recording: `cargo doc` bails on the first crate that fails, so `stella-store`'s private-link error (fixed in #1965) masked this one entirely. Fixing one rustdoc break surfaces the next one down — expect to iterate, not to assume the first fix was the last.

The fix is the same shape as #1965's: name the item in prose rather than link something rustdoc cannot see, with a parenthetical saying why.

## The witness

- [ ] No witness needed (docs) — because: this is a doc-comment-only change, and the gate itself is the check. `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps` exits **101** on `main` and **0** with this change, verified by exit code rather than by reading output.

## The gate

- [x] `cargo doc --workspace --no-deps` under `-D warnings` (exit 0, the whole workspace — re-run after the fix to confirm nothing further was masked)
- [x] `cargo fmt --check`
- [x] CLA signed

Refs #1939, #1965.

## Summary by Sourcery

Documentation:
- Clarify daemon boot module documentation by naming SkipReason::NoResumePoint in text instead of using an intra-doc link that rustdoc cannot resolve.

PR #2010, opened 2026-08-07
fix(stella-protocol): unbreak main — resolve the CompactionRewrite doc link

macanderson:
## What

The gate's `doc-warnings` step (`RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps`) fails on `main` at 6c345532:

```
error: unresolved link to `CompactionRewrite`
   --> crates/stella-protocol/src/event.rs:545:53
    |
545 |         /// under the same `call_id` (#1667); see [`CompactionRewrite`].
    |                                                     ^^^^^^^^^^^^^^^^^ no item named `CompactionRewrite` in scope
```

The type reaches this scope only through its `crate`-level re-export (`lib.rs`), so the bare name resolves to nothing. The field's own type annotation two lines below already spells the working path (`Vec<crate::CompactionRewrite>`); the doc link now matches it. One line, no behavior.

`event.rs` is a grandfathered god file closed to growth — this is a same-line edit, so it does not grow.

## Why it surfaced only now

Layered masking, the shape #1965 records: rustdoc stops at the first crate that fails to document, so each fix reveals the next layer down. `stella-cli` was fixed in #1970; `stella-protocol` is the layer underneath, and it is the deepest — with this applied, `cargo doc --workspace --no-deps` is **clean workspace-wide**.

## Scope — this is one of four independent red gates on main

Verified on `origin/main` at 6c345532 and deliberately **not** bundled, since each already has an owner:

| Gate | Status |
|---|---|
| `doc-warnings` (rustdoc) | **this PR** — nothing else covers it |
| `format-check` (missing trailing newline in `event/tests.rs`) | #2005 |
| `file-size` (baseline skew: `driver.rs` +1, `pipeline/tests.rs` +1) | #2003 |
| `lint` (dead `spend` local) | #2000 |

**Main is green on none of these individually — all four must land.** #2000 also touches `crates/stella-protocol/src/event.rs`, but a different region, so the two should not textually conflict.

## Verification

- `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps` — **clean** (fails on main without this)
- `cargo test -p stella-protocol` — 0 failures
- `cargo clippy -p stella-protocol --all-targets -- -D warnings` — clean

No witness test: a rustdoc-only fix, per the repo's pure-docs exemption. The reproduction above is the evidence.

## Summary by Sourcery

Bug Fixes:
- Correct the CompactionRewrite documentation link in AgentEvent to reference the crate-level type and restore successful rustdoc generation.

### DIFF
--- a/crates/stella-protocol/src/event.rs
+++ b/crates/stella-protocol/src/event.rs
@@ -542,7 +542,7 @@ pub enum AgentEvent {
         /// The replacement bytes each in-place rewrite left behind, one entry
         /// per digest — what lets reconstruction resolve a compacted block to
         /// the bytes the model received rather than the pre-compaction output
-        /// under the same `call_id` (#1667); see [`CompactionRewrite`].
+        /// under the same `call_id` (#1667); see [`crate::CompactionRewrite`].
         /// `serde(default)` — absent on journals written before rewrites were
         /// journaled, whose compacted blocks surface as digest mismatches.
         #[serde(default, skip_serializing_if = "Vec::is_empty")]


macanderson:
Superseded by #2000, which fixes the same `doc-warnings` break in a better form: `[`CompactionRewrite`](crate::CompactionRewrite)` keeps the short rendered label, where this PR's `[`crate::CompactionRewrite`]` would print the fully-qualified path in the generated docs.

I opened this before spotting that #2000's diff already reached `crates/stella-protocol/src/event.rs` — my own PR description asserted the two 'should not textually conflict', and that was wrong: they conflict on exactly this hunk.

The verification I did here still stands and is recorded on #2000: main at 6c345532 is red on four independent gates (rustdoc, fmt, file-size, clippy), and #2000 + #2003 + #2005 together green all four. Closing so there is one PR per gate.

macanderson:
⚠️ **This unbreak would trade one red gate for another.** `docs/wire still matches the types` is [failing on this PR](https://github.com/macanderson/stella/actions/runs/31143111091/job/92756953078), and the cause is the fix itself.

The edited line is a **type doc comment, which is the wire contract**. `schemars` copies it verbatim into the schema's `description` field, so changing

```rust
/// … (#1667); see [`CompactionRewrite`].
```

to anything else restales `docs/wire/agentevent.schema.json` and `serveframe.schema.json`. The `wire-schema` gate then fails, and the net effect on `main` is one break swapped for a different one.

**Fix:** run `scripts/export-agentevent-schema.sh` and commit the result in this PR (AGENTS.md § "Cite a document by its id" neighbours this rule; the exporter's own header says "Run after ANY change to `AgentEvent` or a type it carries").

**Also worth comparing against #2015**, which fixes the same link and *does* carry the regen. It spells it:

```rust
/// see
/// [`CompactionRewrite`](crate::CompactionRewrite).
```

rather than this PR's `[`crate::CompactionRewrite`]`. The difference is not cosmetic once you remember where the string ends up: the linked-form keeps the schema `description` reading `` see [`CompactionRewrite`] ``, while the path-form bakes `crate::CompactionRewrite` into a wire artifact that non-Rust consumers read. The linked form is the better one to converge on.

Given #2015 already carries this fix *plus* its schema regen, the cheapest path is probably to let #2015 land and close this — flagging rather than acting, since I do not know what else is riding on this branch.

Context: there are currently seven open unbreak PRs against a `main` that is red on four gates (fmt, file-size, clippy, doc-link), and several overlap. I verified each failure against a pristine `origin/main` checkout before writing this.

PR #2015, opened 2026-08-07
fix(repo): unbreak main — five red gates, and the deadlock between their fixes

macanderson:
## The problem

`main` at 6c345532 fails **five** independent gates. Four fixes already exist across #2000, #2003 and #2005 — but **two of those PRs are red on exactly the gate the other one repairs**, so none of them can merge:

- **#2003** regenerates `scripts/file-size-baseline.txt` and touches nothing else → fails `wire-schema`, because main's `docs/wire/` is stale against its own types.
- **#2005** regenerates `docs/wire/` and the park consumers → fails `file-size`, because the baseline skew is #2003's fix.

That is a deadlock, and it is why main has stayed red while three unbreak PRs sat open. This branch carries both halves plus #2000's repairs so the set can land in one merge.

## The five gates

| Gate | Break | Fix from |
|---|---|---|
| `lint` | a dead `spend` local | #2000 |
| `doc-warnings` | `[`CompactionRewrite`]` resolves only via the crate-level re-export | #2000 |
| `format-check` | missing trailing newline in `event/tests.rs` | #2005 |
| `file-size` | baseline skew (`driver.rs` +1, `pipeline/tests.rs` +1) | #2003 |
| `wire-schema` | `docs/wire/` stale against #1994's `TurnParked`/`TurnWoken` | #2005 |

The rustdoc one is worth a note: **layered masking**, the shape #1965 records. Rustdoc stops at the first crate that fails to document, so #1970 had to repair `stella-cli` before `stella-protocol` underneath it became visible at all. Anyone fixing one layer and re-running would reasonably have believed they were done.

## Authorship

The substantive hunks are **from #2000, #2003 and #2005** — collected here, not re-derived, so their authors keep the credit. Close those three as superseded if this lands, or close this one if they can be sequenced another way; the point is that they cannot each be green independently. I also opened #2010 for the rustdoc break before finding #2000 already covered it, and closed it as a duplicate.

## Verification

Run against this exact tree, each gate with the command the Makefile uses:

| Check | Result |
|---|---|
| `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps` | clean |
| `cargo fmt --all -- --check` | clean |
| `scripts/check-file-size.sh` | OK — none grew |
| `make wire-schema` | OK — `docs/wire/` matches the types |
| `cargo test -p stella-protocol -p stella-pipeline -p stella-tui -p stella-core` | pass |

Each was also confirmed **failing** on `origin/main` beforehand, so this is a demonstrated repair rather than an assumed one.

One caveat, stated rather than buried: `stella-tui`'s `run_deck_paints_folds_resizes_and_restores_under_a_real_pty` failed once in the batch run and passed in isolation (20s, against an 81s timeout) — a real-PTY timing flake under concurrent build load, not a regression.

The baseline was regenerated with `make file-size-update`, never hand-edited — a hand-merged baseline is what produced the current skew.

## Note for reviewers

No witness test: every hunk restores an existing gate to green rather than changing behavior. The reproduction table above is the evidence, and each gate flips from fail to pass across this diff.

## Summary by Sourcery

Unstick main by combining previously separate fixes so all gates pass together, including lint, doc warnings, formatting, file-size checks, and wire-schema consistency.

Bug Fixes:
- Repair lint break by removing the unused `Spend` local from the pipeline scope stage.
- Fix rustdoc warnings by correcting the `CompactionRewrite` intra-crate link to the crate-level re-export and mirroring it in generated wire docs.
- Restore format-check to green by adding the missing trailing newline in `event/tests.rs`.
- Update `scripts/file-size-baseline.txt` to reflect current binary sizes so file-size checks match the regenerated code and tests.
- Bring `docs/wire` schemas back in sync with protocol types and serveframe definitions, including the `CompactionRewrite` documentation changes.
- Ensure observatory transcripts correctly capture and render `turn_parked` and `turn_woken` events so journal gaps and wake reasons are visible.
- Adjust the management prompt tests to match the current set of model roles and avoid stale expectations.
- Update the fleet dashboard UI to properly represent parked turns as a distinct state instead of misclassifying them as blocked.

Enhancements:
- Extend the observatory database query and journal rendering to include parked and woken turn events, with operator-facing descriptions and timing details.
- Add UI support in the fleet dashboard for displaying parked turns and holding their state across park/wake so operators can distinguish deliberate waits from stalls.
- Clarify the flip-halt arming test module layout and move doubles into the child module to protect against silent deletion on parent rewrites.

Tests:
- Tidy verification hardening tests around flip-halt arming by delegating doubles into the child module and simplifying the parent’s documentation.
- Align management prompt tests with the current role handling to keep the test suite reflecting real behavior.
- Confirm wire-format tests and tag tables remain unchanged while restoring formatting and wire-schema consistency.

Chores:
- Regenerate wire schema artifacts and file-size baselines using the project’s existing tooling so all gates share a consistent view of the repository state.


macanderson:
✅ **Independently verified green.** Checked out this head (`80b78cda`) locally and ran the full `make gate` — **exit 0**, every tier: guards, `file-size`, `wire-schema`, `doc-warnings`, `format-check`, `clippy -D warnings`, the workspace test suite, and the self-driving harness. That matches CI, where all checks pass.

This is the PR that clears `main`. For anyone else looking at the pileup: I verified each of `main`'s four red gates against a *pristine* `origin/main` checkout, and this PR is the superset that fixes all of them —

| Gate red on `main` | Fixed here |
|---|---|
| `cargo fmt --check` — no trailing newline, `event/tests.rs` | ✅ |
| `file size ratchet` — `driver.rs` +1, `pipeline/tests.rs` +1 | ✅ |
| `cargo clippy -D warnings` — dead code in `verification_hardening.rs`, unused `spend` | ✅ |
| `cargo doc -D warnings` — unresolved link to `CompactionRewrite` | ✅ (with the `docs/wire` regen it requires) |

Two notes for whoever merges:

1. **This supersedes #2005**, which I opened — it carries those commits byte-for-byte (the observatory journal allowlist + `journal_entry` arms + frontend arm, `LastAction::Parked` and the fleet fold, and the fmt fix). I have commented there to that effect; nothing is lost by closing it in favour of this.
2. **Prefer this PR's doc-link spelling over #2010's.** Both fix the same line, but a type doc comment on a `stella-protocol` type *is* the wire contract — `schemars` copies it verbatim into the schema `description`. This PR's `` [`CompactionRewrite`](crate::CompactionRewrite) `` keeps that description readable for non-Rust consumers; #2010's `` [`crate::CompactionRewrite`] `` bakes the Rust path into the wire artifact — and #2010 ships no schema regen at all, so it currently fails `wire-schema` and would trade one red gate for another.

The fmt break was mine, from #1994. Thanks for picking it up here.

Issue #2067, opened 2026-08-07
verify_done: the shadow baseline is HEAD, which auto-snapshots already advanced past — no flip is provable, and a missing build artifact fakes one


macanderson:
## Problem

`verify_done` builds its shadow worktree at **`HEAD`** and calls that "the previous code". On the arenabench harness `HEAD` is not the task baseline — the workspace **auto-snapshots the agent's own work as a git commit after every turn**. So by the time the agent authors a witness, `HEAD` already contains the change the witness is supposed to prove.

Observed on `portfolio-optimization` (treatment match `3568657c2db9`). The agent worked it out itself:

> verify_done is comparing against the current HEAD (commit `256c584`), which already has my fixes … The actual baseline I should be comparing against is `825c025` … the "flip" test is inherently vacuous — it can't demonstrate a regression when the fix is already baked into the baseline.

It confirmed this by reading `825c025` directly and finding the original stubs (`raise NotImplementedError("TODO: …")`, `return PyFloat_FromDouble(0.0);  // TODO`). The task baseline is the `stella-harbor: task workspace baseline` commit; `HEAD` is several agent auto-commits past it.

**The witness contract cannot be satisfied under these conditions, by any agent, on any task.** This is a stronger statement than #2064's: even with a runner available and an independent author, the flip is structurally impossible because the shadow *is* the solved tree.

### It also produces FALSE POSITIVES, which is worse

Earlier in the same trial, `verify_done` returned **`WITNESS CONFIRMED`** — on a flip that was not real:

```
- new code:      `python3 test_portfolio_optimized.py` exit 0 (PASS)
- previous code: HEAD 256c5849 + your test files → exit 1 (FAIL)
--- previous-code failure tail ---
ModuleNotFoundError: No module named 'portfolio_optimized_c'
```

The shadow failed because the compiled `.so` is a build artifact absent from a fresh checkout — **not** because the source lacked the fix. The agent's own reasoning:

> my working directory (used for "new code") has the compiled .so present … the fresh checkout has no build artifacts, so both versions would actually fail unless the build step runs first.

The tool hedges in its message ("a compile error … is weaker; prefer behavioral assertions"), but it still reported **CONFIRMED**. A witness credited on a missing build artifact is a false proof, and a false proof is far more damaging than a refused one — it is exactly the "claimed done, not verified done" failure this project exists to prevent.

The agent fixed it by folding the build into `test_cmd` so both sides compile from source. That correction should not depend on the model noticing.

[...]

Issue #2023, opened 2026-08-07
bench: every arenabench run to date is the CONTROL arm — a witness was authored in 1 of 51 trials


macanderson:
## Evidence

Across **51 Stella trials** on the local arena rig (every match under `~/.arenabench/matches/`), the pipeline's proof steps break down as:

| proof step | trials |
|---|---|
| `assurance` | 47 |
| `warrant` | 19 |
| `witness_unavailable` | **14** |
| `verification_unavailable` | 6 |
| `witness_authored` | **1** |

**A witness test was authored in 1 of 51 trials.** Fourteen trials carry an explicit refusal with the same reason:

```
continuing without an authored witness: no author independent of the worker
(verifier and worker both resolved to `openrouter/anthropic/claude-sonnet-5`)
```

Stage reachability tells the same story: of 51 trials, `execute` ran in 47, `verify` in 19, and `witness` in **1**.

## Why

This is #1795 / PR #1867 working exactly as designed, meeting a benchmark configuration that never opts in. `Pipeline::witness_author_independence` (`crates/stella-pipeline/src/pipeline.rs`) returns `Unavailable` when the verifier role resolves to the same `model_ref` as the worker, and the witness stage is skipped rather than authored by the worker grading itself — correct behaviour.

The arm is selected host-side in `bench/harbor_adapter/stella_harbor/posture.py`:

```python
# Host-side selector for the witness/verifier author. Unset (the default) is the
# control arm: one inherited model, authored witness structurally off. Set to a
# second `provider/model` on the worker's provider and the same 89 tasks run the
# treatment arm.
_WITNESS_AUTHOR_ENV = "STELLA_WITNESS_AUTHOR_MODEL"
```

The arena seats never set `STELLA_WITNESS_AUTHOR_MODEL`, so **every arenabench run to date has been the control arm** — Stella measured with its defining contract ("verified done, not claimed done") structurally disabled.

Nothing is *wrong* in the code. The problem is that this is invisible: the arena UI, the scoreboard, and the exported results say nothing about which arm ran, so a reader comparing Stella to Claude Code has no way to know the witness stage was off, and the numbers get quoted as if they came from the full agent.

## What to do

[...]

Issue #1539, opened 2026-08-05
Witness author picks a test runner it cannot verify exists; a missing toolchain silently costs the run its witness


macanderson:
## Problem

The witness author picks a test runner it has no way to confirm exists, and picking a missing one silently costs the run its verification.

The author's tool surface is deliberately capability-minimal: `read_file` plus one atomic `create_witness_test`, and nothing else (`crates/stella-cli/src/candidate_ws/witness_tools.rs`, `ToolExecutor::schemas`). It therefore **cannot execute anything**, so it cannot discover that `pytest` has no Python behind it on this machine.

When it guesses wrong, the failure is quiet in the worst way. `Pipeline::witness_stage` runs the authored command against the baseline; an unspawnable runner returns `CmdKind::Infra`, whose `assertion_result()` is `None` — correctly, since no assertion was observed. That is a `WitnessAbort::degradable`, so the run continues **without a witness**, on the unauthored ladder. No flip is possible, so the verifier's opinion becomes the only evidence, and the turn ends `UNVERIFIED`.

Observed on a Terminal-Bench 2.1 run: the witness died on `infra_failure` from a missing Python, and the run's pass verdict rested on no executed test at all.

## What already shipped

Branch `worktree-fix-candidate-escape-2` (commit `af968739`) took the prompt-level half, in `crates/stella-pipeline/src/witness.rs::witness_prompt`:

- Tells the author it cannot execute anything, so a missing runner yields **no observation** rather than a failing test.
- Requires the runner be evidenced by the repository listing (a manifest plus existing tests written for it), and to emit no `TEST_COMMAND` line rather than guess when nothing is evidenced.
- Injects the project's own configured `test_command` as an anchor when one exists — the only evidence that is a fact rather than an inference, since the project names it and its toolchain is therefore installed.

That is guidance, and it is unenforced. It also does nothing for the common case this matters most in — a benchmark task with **no** configured test command, where the anchor is absent and the repository listing is the only signal.

## Proposal

Give the pipeline (not the author) a bounded, one-shot runner-existence probe, and use its result to constrain or re-ask.

Sketch:
- Before or immediately after the author's turn, probe whether the chosen `TestInvocation::program` is spawnable — a `--version`-style call through the existing `TestRunner` port, or a plain "can this be spawned" check.
- If it is not spawnable, that is a *known* fact the author can act on: re-prompt once naming the runners that ARE available, instead of spending the baseline run to discover it and then degrading.
[...]

PR #2072, opened 2026-08-07
feat(stella-pipeline): a shell is a witness runner, so runner-less workspaces can author

### LINKED ISSUE #2064: pipeline: add a shell runner to RUNNER_VOCABULARY so the witness stage is reachable on runner-less workspaces
## Problem

The witness stage refuses on every Terminal-Bench task with:

```
no supported test runner is available in this workspace,
so an authored witness could never be observed
```

`RUNNER_VOCABULARY` (`crates/stella-pipeline/src/witness.rs:607`) is:

```rust
pub const RUNNER_VOCABULARY: &[&str] = &[
    "cargo", "pnpm", "npm", "yarn", "bun", "vitest", "npx", "bunx", "pytest", "python", "python3",
    "go", "dotnet",
];
```

Terminal-Bench containers are bare shell/system images — nginx config, git surgery, 7z cracking, MIPS interpreters. None of those thirteen programs probes successfully, so `runner_availability` returns empty and `witness_stage.rs:270` degrades before spending a single author turn.

This is measured, not theoretical: **`witness_authored` was 1 of 51 trials** across the whole historical arenabench corpus, and 0 across every cycle run on 2026-08-06/07. It was originally attributed to verifier/worker model collision (#2023); fixing that (match `3568657c2db9`, worker `claude-sonnet-5` / verifier `deepseek-v4-pro` / triage `claude-haiku-4.5`) removed the independence refusal and revealed **this** as the real blocker underneath.

So Stella's defining contract — "verified done, not claimed done" — is structurally unexercisable on the benchmark it is measured by.

## Why a shell runner, rather than shipping a real one

The witness contract is only: *a command that fails on the old code and passes on the new*. A POSIX shell script's exit code satisfies that exactly. And `sh` is guaranteed present in every task container, because the agent is already executing `bash` in it — so there is nothing to ship, install, or vendor.

Alternatives considered and rejected:

- **`pip install pytest` at runtime** — needs network these containers may not have, mutates the workspace under test (contaminating the very diff the verifier reads), and still cannot express an assertion about nginx serving on :8443.
- **Vendoring a statically-linked runner into the SUT** — bloats the frozen binary, needs a per-arch build in `build_sut.sh`, and still does not help a task whose deliverable is a system service rather than a library.
- **Widening the vocabulary with more language runners** — does not help: the tasks have no language project at all.

### 1. Add the runner to the vocabulary

`crates/stella-pipeline/src/witness.rs`:

- Add `"sh"` to `RUNNER_VOCABULARY`. Prefer `sh` over `bash`: it is the lower bound present in busybox/alpine images too, and every task container that has `bash` also has `sh`.
- Extend `runner_probe` with an `sh` arm. The probe must follow the existing contract — "the cheapest invocation that answers *is this runner usable here* … discovers and runs no test". `sh -c 'exit 0'` is the natural analogue of `--version` here; do **not** use `sh --version`, which is non-portable (dash and busybox reject it).
- The invariant test at `witness.rs:1421` (`for program in RUNNER_VOCABULARY { runner_probe(program).unwrap_or_else(panic) }`) will enforce that the new entry is probeable — keep it passing.

### 2. Accept a shell witness in `parse_test_invocation`

The same file's `parse_test_invocation` currently only accepts `python`/`python3` as **pytest hosts**. It needs an `sh` arm that accepts `sh <script>` (and `sh -c '<command>'`), so an authored witness of the form `sh ./witness_check.sh` is recognised as a test invocation rather than rejected as arbitrary shell.

Keep the acceptance narrow: a bare `sh` with no argument is not a test, and neither is a pipeline or a redirect — those are shell programs, not assertions, and admitting them turns the flip oracle into a general shell executor.

### 3. Teach the witness author the shape

The author prompt must know that on a runner-less workspace the witness is a shell script asserting observable system state — `curl -sf https://localhost:8443/ | grep -q expected`, `git log --oneline | grep -q 'the commit'`, `test -f /etc/nginx/ssl/selfsigned.crt` — and that it must **exit non-zero before the change and zero after**.

This is the part most likely to be got wrong: a model asked for "a test" in a container with no test framework tends to write a pytest file anyway. Say the runner explicitly in the prompt, derived from `runner_availability`'s answer rather than left to the author to guess — which is the same lesson #1539 already encoded when it moved availability *before* author spend.

## What must not regress

- **Vacuous witnesses must still be caught.** The flip oracle already rejects these — a cycle-1 trial produced `VACUOUS TEST — the witness test ALSO PASSES on the previous code (HEAD eb3d6fbd)`. That guard is runner-agnostic and must stay that way; a shell witness that passes on the baseline is rejected exactly like a pytest one.
- **Tamper exclusion.** The worker must still not be credited for a flip on witness files it modified.
- **The witness stays scaffolding.** It lives in the candidate workspace and is discarded with it unless `--keep-witness` promotes it. A `witness_check.sh` must not be left behind in the task tree, where it would also pollute the untracked-file diff the verifier reads (#2041).

## Verify / done when

Witness test (fails on `main`): a synthetic workspace containing **no** program from `RUNNER_VOCABULARY` except a shell resolves a non-empty `runner_availability`, and the witness stage proceeds to authoring instead of returning `WitnessAbort::degradable("no supported test runner…")`. On `main` that same fixture degrades.

Plus:

- a `parse_test_invocation` unit test accepting `sh ./witness_check.sh` and rejecting bare `sh`, `sh -c 'true | false'`, and redirects;
- the existing `witness.rs:1421` vocabulary/probe invariant still passing;
- a flip-oracle test proving a shell witness that passes on the baseline is still rejected as vacuous.

**End-to-end acceptance** is the number this issue exists to move: run the five-task arenabench cycle with an independent verifier (roles in the match payload — see #2023) and confirm `witness_authored` rises above zero. It has been 0 or 1 in every run ever recorded. Measure with `python3 ~/.arenabench/trace-metrics.py --match <id> --per-trial`.

## Files

- `crates/stella-pipeline/src/witness.rs` — `RUNNER_VOCABULARY` (607), `runner_probe` (623), `parse_test_invocation`, vocabulary invariant test (1421)
- `crates/stella-pipeline/src/pipeline/witness_stage.rs` — the availability gate (263-276) and the author prompt
- `crates/stella-pipeline/src/pipeline.rs` and `src/pipeline/tests.rs` are **god files closed to growth** — land new logic in a sibling submodule

Related: #2023 (the model-independence half, now fixed), #1539 (availability established before author spend), #2041 (untracked-file diff channels).

macanderson:
## What

The witness stage refused on every Terminal-Bench task with `no supported test runner is available in this workspace` — `RUNNER_VOCABULARY` listed only language toolchains, and TB containers are bare system images. Measured: `witness_authored` was **1 of 51 trials** across the whole arenabench corpus (#2064). With #2023's independence fix landed, this was the real blocker underneath: Stella's defining contract was structurally unexercisable on the benchmark it is measured by.

`sh` joins the vocabulary as the floor — present in every container the agent already executes `bash` in, and the witness contract (fails on old code, passes on new) is exactly a script's exit code.

## Decisions a reviewer should check

- **Probe**: `sh -c 'exit 0'` — `sh --version` is non-portable (dash/busybox reject it). The version-style probe invariant test carves this out with its reason.
- **Narrower than the issue sketched**: only `sh <script>.sh` parses; `sh -c '<inline>'` is deliberately refused. An inline program rides as one opaque argv word, so the per-argument confinement checks cannot see an absolute path or `..` inside it (`sh -c 'cargo test /tmp/outside.rs'` would slip the escape check the direct form rejects) — and an invocation naming no artifact can never be tamper-excluded. The pre-existing rejection test for `sh -c 'cargo test'` stays green, which is the boundary staying intact.
- **Artifact discipline**: `validate_witness_invocation` requires the invocation to pin exactly the authored `.sh` file; `is_witness_test_path` requires intent in the name (`witness_*`, `test_*`, `*_test.*`, or a tests dir), so `deploy.sh` is not a witness artifact.
- **The author is told the shape**: when the probed availability set is shell-only, `witness_prompt` adds a section prescribing a POSIX script asserting observable state (exit non-zero before, zero after) — the #1539 discipline of deriving constraints from the probe rather than letting the model guess. Without it a model asked for "a test" writes a pytest file into a workspace with nothing to collect it.
- **Vacuity/tamper guards unchanged by construction**: the baseline-passes rejection and tamper exclusion in `witness_stage` never branch on the runner.
- **witness.rs test module extracted verbatim** to `witness/tests.rs` (the #1914 pattern): the file was 1438 lines, four lines of growth from the 1500 ceiling. No test was deleted — all 130 witness tests pass, including the moved ones.

## Witness tests (fail on `main`)

- `a_shell_only_workspace_still_resolves_a_runner` — a `TestRunner` where only `sh` probes true resolves `["sh"]`; on main the vocabulary has no `sh`, the set is empty, and the stage degrades with zero author spend.
- `sh_commands_parse_narrowly`, `sh_invocation_must_target_the_authored_script`, `sh_witness_paths_declare_intent`, `a_shell_only_workspace_prompts_a_shell_witness` — all fail on main (`sh` unsupported).

Verified: `cargo test -p stella-pipeline --lib witness` → 130 passed / 0 failed; `check-file-size.sh` and `check-module-reachability.py` pass (witness.rs 859 lines, witness/tests.rs 722).

**End-to-end acceptance** (next arenabench cycle, not this PR): `witness_authored` rising above zero on the five-task cycle with an independent verifier.

Closes #2064
Refs #2023, #1539, #2067

## Summary by Sourcery

Extend witness pipeline support to runner-less workspaces by treating POSIX shells as a first-class test runner and updating prompts, validation, and tests accordingly.

New Features:
- Allow `sh` as a supported witness runner, including probe, parsing, invocation validation, and artifact path classification for shell-based witnesses.

Enhancements:
- Adjust runner availability probing and system prompt messaging to reflect shell-only environments and guide authors to create POSIX shell witness scripts.
- Extract the large witness test module from `witness.rs` into a dedicated `witness/tests.rs` file to keep the main module within size limits while preserving coverage.

Tests:
- Add new witness tests covering shell runner parsing, invocation targeting, artifact naming intent, shell-only prompt behavior, and runner availability in shell-only workspaces.

### DIFF
--- a/crates/stella-pipeline/src/pipeline/witness_stage.rs
+++ b/crates/stella-pipeline/src/pipeline/witness_stage.rs
@@ -15,11 +15,12 @@ use crate::witness::{RUNNER_VOCABULARY, WITNESS_SYSTEM_PROMPT, runner_probe};
 /// baseline's own [`TestRunner`] port; a port that cannot check answers
 /// `true` per its default, so this only ever narrows on real evidence.
 ///
-/// Probed concurrently: the vocabulary is thirteen programs, each an awaited
-/// process spawn, and a best-of-N fan-out pays this per candidate — serially
-/// that was thirteen spawn round-trips of pure latency before any authoring
-/// could start. Order is restored from the vocabulary, not the completion
-/// order, so the prompt's listing stays deterministic (invariant 7).
+/// Probed concurrently: every vocabulary program is an awaited process
+/// spawn, and a best-of-N fan-out pays this per candidate — serially that
+/// was one spawn round-trip of pure latency per program before any
+/// authoring could start. Order is restored from the vocabulary, not the
+/// completion order, so the prompt's listing stays deterministic
+/// (invariant 7).
 async fn runner_availability(tests: &dyn TestRunner) -> Vec<String> {
     let probes = RUNNER_VOCABULARY.iter().filter_map(|program| {
         let probe = runner_probe(program)?;
@@ -704,6 +705,32 @@ impl<'a> Pipeline<'a> {
 #[cfg(test)]
 mod tests {
     use super::*;
+
+    /// A workspace with a shell and nothing else — the Terminal-Bench
+    /// container shape (#2064): nginx config, git surgery, system images
+    /// with no language toolchain at all.
+    struct ShellOnlyRunner;
+
+    #[async_trait::async_trait]
+    impl TestRunner for ShellOnlyRunner {
+        async fn run_test(&self, _invocation: &TestInvocation) -> crate::ports::CmdOutcome {
+            unreachable!("availability probing never runs a test")
+        }
+        async fn runner_available(&self, probe: &TestInvocation) -> bool {
+            probe.program == "sh"
+        }
+    }
+
+    /// #2064's witness: a workspace where no language runner probes — only a
+    /// shell — resolves a non-empty availability set, so the witness stage
+    /// proceeds to authoring instead of degrading with "no supported test
+    /// runner". On `main`, `sh` is not in the vocabulary, the set is empty,
+    /// and the stage degrades before spending an author turn.
+    #[tokio::test]
+    async fn a_shell_only_workspace_still_resolves_a_runner() {
+        let available = runner_availability(&ShellOnlyRunner).await;
+        assert_eq!(available, vec!["sh".to_string()]);
+    }
 
     /// #1790's witness: a compile-error baseline is RECORDED, not refused —
     /// refusal would reject the most common Rust witness shape (a missing-API
--- a/crates/stella-pipeline/src/ports.rs
+++ b/crates/stella-pipeline/src/ports.rs
@@ -590,8 +590,11 @@ pub trait DiagnosticRunner: Send + Sync {
 pub trait TestRunner: Send + Sync {
     /// Spawn `invocation.program` with `invocation.args` directly.
     async fn run_test(&self, invocation: &TestInvocation) -> CmdOutcome;
-    /// Whether this version-style probe invocation completes with exit 0 —
-    /// "is this runner actually usable here?", never a test run (#1539).
+    /// Whether this probe invocation completes with exit 0 — "is this
+    /// runner actually usable here?", never a test run (#1539).
+    /// Version-style for every language runner; the cheapest no-op
+    /// (`sh -c 'exit 0'`) for the shell, whose `--version` is non-portable
+    /// (#2064).
     ///
     /// The witness author's tool surface is deliberately unable to execute
     /// anything, so runner existence is a fact only the pipeline can
--- a/crates/stella-pipeline/src/witness.rs
+++ b/crates/stella-pipeline/src/witness.rs
@@ -141,6 +141,21 @@ pub fn parse_test_invocation(command: &str) -> Result<TestInvocation, TestInvoca
             (Some("-m"), Some("pytest"))
         ),
         "go" | "dotnet" => matches!(args.first().map(String::as_str), Some("test")),
+        // #2064: the runner-less-workspace witness — exactly one positional
+        // `.sh` script, the authored artifact. `split_test_words` has
+        // already rejected pipelines, redirects, substitution, and
+        // backgrounding anywhere in the line, and the script path gets the
+        // same escape checks as every other runner's arguments. `sh -c
+        // '<inline>'` is deliberately NOT accepted, although the issue
+        // sketched it: the inline program rides as one opaque argv word, so
+        // the per-argument confinement checks cannot see a path inside it —
+        // and an invocation that names no artifact could never be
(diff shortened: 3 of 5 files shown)


macanderson:
Pushed 15018fe4 — a formatting-only commit (`cargo fmt --all`, 1 file, +6/−1) to `crates/stella-tools/src/repo.rs`, which is **outside this PR's scope on purpose**: main itself is red on `cargo fmt --check` (#2060 merged with its fmt check failing), so every PR based on current main fails the fmt gate through no fault of its diff. This PR carries the mechanical fix and heals main when it merges. Tracking + coordination (so no duplicate unbreak PR gets opened): #2074.

On the superseded run (31217692267), fmt was the **only** red step — clippy, rustdoc, wire-schema, and the prompt-cache fixtures all passed on the previous head.

Issue #2110, opened 2026-08-07
pipeline/adapter: trial idles from final answer to AgentTimeoutError — verify/verdict never run on big-tree tasks


macanderson:
## Problem

A Stella trial that finishes its work does not end its session — it idles until harbor's 900s `AgentTimeoutError` kills it. Concrete instance (2026-08-07, match `14d0127b8ca7`, trial `sqlite-with-gcov__2PdctrU`, worker `openrouter/moonshotai/kimi-k3`): the agent completed the task and delivered its final summary after 11 steps (~$0.11, well under 5 minutes of model time); `stella-events.jsonl` then records nothing further — no `verify` stage, no `verdict`, no `complete` — and `exception.txt` shows `AgentTimeoutError: Agent execution timed out after 900.0 seconds`. The same signature (reward 1.0 + AgentTimeoutError, trace ending at `execute`) appears on `sqlite-with-gcov` in BOTH arms of match `2229d35b916b` and on `chess-best-move` (kimi arm).

## Why it matters

- Each affected trial wastes up to ~10 minutes of wall clock; at 23 tasks × 2 arms the match runtime multiplies.
- `clock_time` per seat becomes a measure of the timeout, not the agent.
- The pipeline never reaches verify/verdict on these trials, so verification evidence (including any witness settlement, see #2089) is silently lost — the trial is graded by harbor's own checker only.

## Where to look

The trace ends AFTER the worker's final text but BEFORE the `verify` stage event, so the hang sits between execute settling and verification starting — in the pipeline driver or in the adapter's turn loop (`bench/harbor_adapter/stella_harbor/__init__.py::run`, frame visible in the AgentTimeoutError traceback at line ~1460). One candidate mechanism: the post-turn workspace probe / diff collection walking a huge fresh tree (`/app/sqlite` = ~7,500 extracted files + build artifacts) — both timeout instances are tasks that unpack/build large sources. Needs a trace with timestamps (see the missing-timestamps issue) or local repro: run `stella` via the adapter on sqlite-with-gcov and watch where the wall clock goes after the final message.

## Definition of done

The mechanism is identified and fixed or bounded (e.g. diff probe cap), and a re-run of `sqlite-with-gcov` completes with `verify`/`verdict`/`complete` events present and no AgentTimeoutError. Witness: the trial's event journal contains a `complete` event within a bounded margin of the worker's final step.

Related: #2109 (cap not adjustable), #2089 (witness measurement starved by the same clock).

[...]

Issue #2089, opened 2026-08-07
bench: re-measure witness_authored on a SUT carrying #2071+#2072 — the number both fixes exist to move


macanderson:
## What

`witness_authored` has been 0-or-1 across every recorded arenabench trial (1 of 51 in the historical corpus, 0 in every 2026-08-06/07 cycle). Two structural blockers are now fixed and merged:

- #2064 (PR #2072): `sh` joined `RUNNER_VOCABULARY`, so runner-less TB containers resolve a non-empty availability set and the witness stage reaches authoring.
- #2067 (PR #2071): `verify_done` measures the flip against a pinned baseline (`refs/worktree/stella/witness-baseline` from the candidate workspace, `refs/stella/task-baseline` from the harness) instead of a HEAD that already contains the work, and refuses import/loader failures as `WITNESS_INCONCLUSIVE_BUILD_ERROR` instead of crediting them.

Neither fix has been measured end-to-end. This issue is that measurement.

## How

1. Build/stage a SUT from a commit ≥ both merges (`gh pr view 2071 2072 --json mergeCommit`), via the arena's SUT build panel or `build_sut.sh`. **The SUT staleness trap is real** — every match Aug 4–7 once measured a binary 291 commits stale; verify `stella_source_commit` in trial metadata.
2. Run the five-task cycle **with an independent verifier** (roles in the match payload — worker `claude-sonnet-5`, verdict `deepseek-v4-pro`, triage `claude-haiku-4.5` was the working #2023 configuration; the role select from #2065/PR #2078 makes this pickable in the UI once merged).
3. Census from the trace, not the summary (the trace-discipline rule): `python3 ~/.arenabench/trace-metrics.py --match <id> --per-trial`, plus `witness_authored` state and the full `proof` reason strings in `stella-events.jsonl`.

## Done when

- `witness_authored` > 0 across the cycle, OR the per-trial refusal reasons name a *new* blocker (file it, linked here — #2064's history is exactly this: fixing #2023 revealed the runner gap underneath).
- Any `verify_done` call in the trials shows a pinned-baseline provenance line (`pinned by refs/...`) rather than bare HEAD, and no `WITNESS CONFIRMED` rides a `ModuleNotFoundError` tail.

## Constraints

- No heavy local builds while a match runs; matches serialize on the one Colima VM.
- Compare `priced_cost`, never `total_cost`; operational aborts (401s, credit exhaustion) relaunch and stay out of the denominator.
[...]

PR #2136, opened 2026-08-08
bench harness: unblock witness settlement + export per-trial telemetry

### LINKED ISSUE #2109: arenabench: surface agent_timeout_multiplier on MatchSpec — the 900s cap can starve the witness stage it now needs to measure
## Problem

Harbor supports per-trial timeout scaling (`agent_timeout_multiplier` is visible in every seat `config.json`, currently `None`), but `MatchSpec` (`arenabench/arenabench/model.py`) surfaces only `setup_timeout_multiplier`. The agent cap is therefore pinned at the dataset default (900s on Terminal-Bench 2.1) with no way to raise it per match.

Why this matters *now*: the witness stage runs **after** execute (execute → witness → verify → verdict). A trial that spends most of its 900s in execute leaves the witness author + flip oracle + revise loop to be truncated by `AgentTimeoutError` — i.e. the acceptance measurement for #2071/#2072 (#2089) can be structurally starved by a knob nobody can turn. Observed 2026-08-07 (match `14d0127b8ca7`, `sqlite-with-gcov__2PdctrU`): agent finished its work in ~5 minutes, the session then idled to the 900s cap and died with `AgentTimeoutError` — every one of those wasted minutes is time a witness-warranted task would have needed after execute.

## Fix

Add `agent_timeout_multiplier` (and plumb `verifier_timeout_multiplier` while there — same shape) to `MatchSpec`: field, `to_json`/`from_json` round-trip, TOML template dump/parse, and pass-through to the harbor job config where `config.json` already carries the key. Clamp like `setup_timeout_multiplier` (`max(1.0, …)`).

## Verify

A match created with `"agent_timeout_multiplier": 2.0` produces seat `config.json` with `agent_timeout_multiplier: 2.0` and harbor logs a 1800s cap. Witness: spec round-trip test in the arenabench test suite (`tests/`), failing on current main where the field is dropped.

Related: #2089 (the measurement this knob protects), #2098 (launch-path spec fidelity).

### LINKED ISSUE #2122: harbor adapter: exec wait blocks on stream EOF held open by task daemons — completed trials die as AgentTimeoutError
## Problem

A trial whose Stella pipeline **finished completely** still dies with `AgentTimeoutError`: match `14d0127b8ca7`, trial `stella-kimi-k3/git-multibranch__556f6R6` (2026-08-07) — its `stella-events.jsonl` ends with a `complete` event (pipeline done, verdict issued), yet `exception.txt` shows `AgentTimeoutError: Agent execution timed out after 900.0 seconds`, and harbor graded it separately (reward 1.0).

So the hang is not in the measured binary: it sits in the adapter between Stella's exit and `run()` returning — `bench/harbor_adapter/stella_harbor/__init__.py`, the `await _secure_exec_with_credential_fd(...)` call (~line 1460 frame in the traceback).

## Likely mechanism

`git-multibranch` requires the agent to start long-lived daemons (`sshd`, `nginx`) inside the task container. Children spawned by Stella's bash tool inherit the exec's stdout/stderr descriptors; a wait that reads the exec stream **until EOF** never sees EOF because the daemons hold the write end open after Stella exits. Classic inherited-fd stream-wait hang. Consistent with the observed pair: pipeline `complete` present, adapter stuck for the remaining ~10 minutes, wait_for kills it.

## Fix direction

In `_secure_exec_with_credential_fd` (or its exec wrapper): treat **process exit** as the completion signal, not stream EOF — e.g. bounded drain after exit, or run the binary under `setsid` with the stream teed to the durable sink file (which `_DURABLE_STREAM_ENV` already provides — the adapter prefers the mounted `stella-events.jsonl` over captured stdout anyway, so stdout EOF carries no unique data).

**Note on validity:** this is apparatus, not subject — it can be fixed without invalidating the pinned SUT (`447f3393`) measurements, but do not hot-patch a running match; land it between matches so all trials in one match share one harness.

## Verify

Re-run `git-multibranch` (either arm): trial ends within seconds of the journal's `complete` event, no `AgentTimeoutError`. Witness: an adapter unit test with a fake exec whose child holds the pipe open — times out on current code, returns on the fix.

Related: #2110 (the in-Stella idle shape, distinct — that trial has NO `complete` event), #2109 (cap knob), #2089.

macanderson:
Two bench-harness fixes, one goal: let the witness measurement (#2089) actually settle on Terminal-Bench. Both are apparatus, not the measured binary — the SUT pin stays honest.

**Commit 1 — exec stream release (Closes #2122).** `docker compose exec`'s stream EOFs only when every in-container fd holder lets go; a daemon the agent legitimately started (sshd/nginx) holds it past Stella's exit, so `communicate()` outlived *finished* trials until the 900s AgentTimeoutError (evidence: match `14d0127b8ca7`, kimi `git-multibranch` — journal has a terminal `complete` event, trial still timed out). The wait now races a probe of the host-visible durable journal; only a proven `complete` releases it, hung pipelines (#2110's shape) stay loudly visible, the client is terminated but container daemons stay up for the verifier, and the release is written to the trial's stderr log. Witness: `tests/test_stream_release.py` pins both hang shapes with real local processes (live client holding the stream; dead client with an orphan on the pipe).

**Commit 2 — agent_timeout_multiplier (Closes #2109).** Harbor accepts `--agent-timeout-multiplier`; the arena never passed it, so TB's 900s budget starved the witness stage — both spotlight `nginx-request-logging` trials on SUT `447f3393` (which carries #2071+#2072) died mid-witness-stage. Mirrors `setup_timeout_multiplier` through MatchSpec, JSON/TOML round-trips, and the Harbor argv, with the same below-1.0 clamp. Witness: `test_agent_timeout_multiplier_survives_json_and_toml` (fails on main — the field is dropped at every hop). The field's doc comment carries the comparability disclosure: same multiplier for every arm keeps a head-to-head fair; numbers quoted against external leaderboards must disclose it.

Two commits in one PR deliberately: they are the minimal pair unblocking the same measurement, and the second is what the first gets exercised under. Exemplar for the stream race: tokio's `Child::wait_with_output` vs kill-on-drop discussion — wait on the *proof of completion*, not the pipe.

**Pre-existing failure, not this PR:** `tests/test_replay_docker.py::test_flip_lands_where_the_solution_actually_appeared` fails identically on the unmodified checkout (docker-dependent bisect on a loaded box).

Refs #2089, #2110.

## Summary by Sourcery

Adjust Harbor exec handling and arena configuration to allow Terminal-Bench witness measurements to complete without being blocked by hanging exec streams or insufficient agent timeouts.

New Features:
- Introduce a completion-probe-aware exec communication helper that can release hanging Docker exec streams once a trial is proven finished.
- Add an agent_timeout_multiplier knob to MatchSpec that scales the agent execution budget and is propagated through JSON/TOML configs and Harbor launch arguments.

Enhancements:
- Use the durable journal’s terminal event to detect completed runs and safely terminate the Harbor exec client while leaving in-container daemons running.
- Record an explicit note in ExecResult stderr when the exec stream is force-released so the behavior is auditable in trial logs.

Tests:
- Add tests that exercise the exec stream release helper under various hang and EOF scenarios using local subprocesses.
- Add a test ensuring agent_timeout_multiplier survives JSON and TOML round-trips and clamps values below 1.0.


macanderson:
Third commit added: `feat(bench): export the container's .stella/private telemetry per trial` (e8a898b4) — pulls `store.db`/`reflections.jsonl`/`context.db` (+WAL/SHM sidecars) into `agent/telemetry/` before teardown, best-effort, 4 new tests. **Correction:** that commit message references #2141; the real tracking issue is #2140 (number predicted before filing — my error). The Rust-side ingest (`stella usage sync --from` + journal backfill) is specified in #2140.

Issue #2141, opened 2026-08-08
pipeline: an unbounded witness_repair call (965s, 36k tokens) can spend the whole trial budget


macanderson:
## Problem

A single `witness_repair` model call can consume a trial's entire remaining wall-clock budget. Observed 2026-08-07, match `293dc7b9c0e6` (spotlight r2, 1800s agent cap, SUT `447f3393`), trial `stella-kimi-k3/nginx-request-logging__aHW7kgt`:

- plan: 1 step, 11s. execute: 28 steps, 125s — the worker was fast.
- witness stage: `witness_author` (z-ai/glm-5.2) 3 steps ≈ 60s, then **one `witness_repair` step of 965 seconds / 36,224 output tokens**, still mid-`reasoning` stream when harbor's cap killed the trial (`AgentTimeoutError`, reward 1.0 — harbor graded the already-solved task independently).

The sibling arm in the same match (glm worker, kimi-k3 author) settled its witness and produced the first deterministic fail→pass flip ever recorded on Terminal-Bench — so the machinery works; the repair rung is the remaining unbounded step.

Trace: `~/.arenabench/matches/293dc7b9c0e6/jobs/293dc7b9c0e6-stella-kimi-k3/nginx-request-logging__aHW7kgt/agent/stella-events.jsonl` (last event is a `reasoning` delta; the 965s duration is in that step's `step_usage.duration_ms`).

## What to fix

The witness stage's repair path (`crates/stella-pipeline` — the `witness_repair` role) issues model calls with no per-call wall-clock or output bound proportionate to what the stage can afford. A reasoning-heavy model at ~37 tok/s turns "let it finish" into 16 minutes. Directions, not mutually exclusive:

- Bound the repair call: per-call `max_tokens` and/or model wall-clock timeout sized from the stage's remaining budget (the pipeline knows its elapsed time; the budget guard consults between calls — invariant 6 — but nothing bounds a single call's duration here; see the `model_timeout` posture key, #1211, for the existing per-provider timeout seam).
- Cap repair attempts/effort: the repair rung inherits the verifier role's `effort: high`; a repair of a witness script arguably warrants lower effort or a tighter token ceiling than verdict reasoning.
- Degrade honestly on breach: a repair that exceeds its bound should yield `witness_unavailable` (an honest refusal, already a defined proof kind) rather than riding the trial into an external timeout that discards the whole verification record.

## Definition of done

[...]

Issue #2149, opened 2026-08-08
pipeline: the witness_author turn is unbounded in wall clock — same runaway-stream shape #2141 closed for the repair rung


macanderson:
## Problem

#2141 bounded the witness stage's **repair** engine turn with a stage latency ceiling (`witness_repair_bound` in `crates/stella-pipeline/src/pipeline/witness_stage.rs`): the smaller of a 300s static ceiling and half the run's remaining declared wall clock, dropping the in-flight turn on breach and degrading honestly to `WitnessUnavailable`.

The **author** turn immediately above it (`Pipeline::witness_stage`, the first `run_engine_turn` against the authoring baseline) still runs unbounded. It has the identical exposure: a reasoning-heavy author model whose stream never goes idle is invisible to `EngineConfig::model_timeout` (an idle-gap bound, #1211), and nothing above it ends the call short of the harness kill. On the motivating trace (match `293dc7b9c0e6`, trial `stella-kimi-k3/nginx-request-logging__aHW7kgt`) the author happened to settle in ~60s over 3 steps — the repair was simply the rung that ran away first.

## What to do

Apply the same bound to the author turn, in `crates/stella-pipeline/src/pipeline/witness_stage.rs`:

- Derive a ceiling with the same shape as `witness_repair_bound` (a static ceiling clamped by the remaining `PipelineConfig::run_budget`, via `Pipeline::remaining_run_budget`). The author turn is a multi-step tool-using turn (probe → write artifact → reply), so its static ceiling should be sized above the repair's 300s or shared with justification — decide and document, the way `WITNESS_REPAIR_CEILING`'s doc comment does.
- Enforce with `tokio::time::timeout` around the author's `run_engine_turn`, returning `WitnessAbort::degradable(...)` on breach — the wording pattern "witness author exceeded its wall-clock bound (...); the executed change stands unproven". The safety argument for hard-dropping the turn is already written on the repair site's comment (throwaway authoring snapshot, dead `messages`, `CancelUsageGuard`'s `UsageIncomplete { Cancelled }` envelope, budget guard untouched — invariant 6 unaffected); it applies verbatim.
- Also check the bound pre-dispatch (`is_zero()` → degrade for free), as the repair site does.

## Witness test

[...]

Issue #2128, opened 2026-08-08
pipeline: role calls enable reasoning but cap output below the reasoning budget — every triage/verdict call in match cc00894779ff returned finish_reason=length with empty text


macanderson:
## Problem

Pipeline role calls (triage, verdict/verifier) enable reasoning on reasoning-capable models while capping output at 512 (triage) / 1024 (verdict) tokens. On reasoning models the reasoning stream is billed against that same cap, so the model burns the entire allotment thinking and emits **zero visible output**: `finish_reason: "length"`, `output_text: ""`.

In match `cc00894779ff` (SUT 447f3393, 2026-08-07; triage = `z-ai/glm-5.2` reasoning-on, verifier = `moonshotai/kimi-k3` reasoning-on) this happened on **essentially every non-worker protocol call across all six trials that ran**:

- triage empty in `git-multibranch`, `log-summary-date-ranges`, `cobol-modernization`, `count-dataset-tokens`, `extract-elf` (e.g. cobol: 512 output tokens, 512 reasoning tokens, `output_text: ""`).
- verdict empty in `log-summary-date-ranges`, `count-dataset-tokens`, `extract-elf` (both rounds; e.g. kimi-k3: 1024 output tokens, 1027–1039 reasoning tokens, empty text).

Downstream damage, all silent:

1. Empty triage → no `CLASS:`/`RESEARCH:` lines → research/plan/scope/witness stages skipped, assurance flags fall through to defaults (`witness:false, verifier:true`) that look like decisions but are not.
2. Empty verdict → `verdict_degraded` → heuristic fallback → `passed: false` on already-solved tasks → rework loops and 900s harness timeouts (see companion issue on the fallback).

This is a config/architecture defect, not a model-quality issue: the caps are simply below what a reasoning model needs before its first answer token.

## Evidence

Traces under `~/.arenabench/matches/cc00894779ff/jobs/cc00894779ff-stella-sonnet-5-full-pipeline/<trial>/agent/stella-events.jsonl` — `step_usage` events with `role: "triage"` / `"verdict"`, `finish_reason: "length"`, `output_text: ""`, reasoning_tokens ≥ output cap.

## Fix directions

- Exclude reasoning tokens from the role-call output cap, or raise role caps well above the model's typical reasoning budget (provider-parity-aware — invariant 8).
- Treat an empty-output role response as a loud failure: retry once with a raised cap or reasoning disabled, and emit a diagnostic — never fall through to defaults silently.
[...]

PR #2162, opened 2026-08-08
fix(stella-pipeline): bound the witness author/repair output, never the whole trial

macanderson:
Refs #2149 · Refs #2141

## The bug

The witness author and repair turns build their engine from `witness_engine_config`, which inherits the worker's `max_output_tokens` — 131k on a measured run. Uncapped, a reasoning-heavy model spent it: in match `293dc7b9c0e6` (the same match that produced the **first deterministic witness flip on Terminal-Bench**), the kimi-k3 arm's single `witness_repair` call streamed **36,224 output tokens over 965 seconds**, riding an already-solved trial into Harbor's 900→1800s agent timeout and discarding the entire verification record. The glm arm, whose repair stayed short, settled its flip in the same match — so the machinery works; this was the last unbounded step.

## The fix

`witness_engine_config` now lowers `max_output_tokens` to `WITNESS_MAX_OUTPUT_TOKENS` (16,384) **after** role shaping, so:
- an operator's stricter verifier `max_output_tokens` still wins (min, never raise);
- a too-high or absent ceiling is brought down to the bound;
- a call that exceeds it truncates and degrades through the existing no-marker path (`witness repair produced no TEST_COMMAND line` → degradable), **freeing the budget instead of consuming it** — the honest-degradation outcome the issue asks for.

16k is over 8× the largest legitimate authoring call observed on a settled flip (<2k) and less than half the runaway. Bounds both author and repair because both build from the one config method. It also sits well clear of the reasoning budget that starves the 512/1024-token role caps in #2128 — reasoning bills against this same ceiling, so a value near legitimate output alone would trade a runaway for a silently empty authoring reply.

## Reconciled with #2151 — two axes, neither subsumes the other

This branch forked **before** #2151 (`7b5155db`) merged, and both edit `crates/stella-pipeline/src/pipeline/witness_stage.rs`. They are complementary, and the branch has been rebased onto current `main` so both survive:

| Axis | Mechanism | Covers | From |
|---|---|---|---|
| **Wall clock** | `witness_repair_bound` — min(300s static ceiling, half the remaining `PipelineConfig::run_budget`), checked pre-dispatch and enforced with `tokio::time::timeout`, degrading to `ProofStep::WitnessUnavailable` | the **repair** turn | #2151 (on `main`) |
| **Output tokens** | `bound_witness_output_tokens` / `WITNESS_MAX_OUTPUT_TOKENS` — a min-not-raise per-call ceiling applied after role shaping | the **author** *and* **repair** turns | this PR |

A token cap cannot end a call that stalls without emitting tokens; a latency ceiling cannot stop a fast runaway from spending the budget it was granted. `WITNESS_MAX_OUTPUT_TOKENS`' doc comment now states the split so a later maintainer cannot read one as superseding the other. No test or line from #2151 was removed — see the evidence below.

## Witness

`pipeline::witness_stage::tests::the_witness_output_ceiling_only_ever_lowers`: 131k→16k, None→16k, 2k stays 2k. References `bound_witness_output_tokens`/`WITNESS_MAX_OUTPUT_TOKENS`, absent on `main`, so it cannot compile there.

Coexistence evidence after the rebase — the full `stella-pipeline` lib suite is green (**613 passed, 0 failed**), including all three:

- `pipeline::witness_stage::tests::the_witness_output_ceiling_only_ever_lowers` — this PR ✅
- `pipeline::witness_stage::tests::the_repair_bound_is_the_ceiling_tightened_by_remaining_clock` — #2151 ✅
- `pipeline::tests::verification_hardening::witness_repair_bound::a_stalled_witness_repair_is_bounded_and_degrades_honestly` — #2151's end-to-end stalling-provider witness ✅

`cargo fmt -p stella-pipeline --check` and `cargo clippy -p stella-pipeline --all-targets` are both clean.

**No tests were deleted by this PR.**

## Why `Refs #2149`, not `Closes`

#2149 asks for a **wall-clock** bound on the author turn — a `witness_repair_bound`-shaped ceiling around the author's `run_engine_turn`, proven by a stalling-provider witness test. This PR bounds the author's *tokens per call*, which kills the specific runaway-stream shape but is not that bound: the author turn is multi-step (probe → write artifact → reply), so N calls each capped at 16k still have unbounded elapsed time, and a provider that stalls without emitting is untouched by a token cap. **#2149 stays open** for the wall-clock half.

#2141 is already closed by #2151 (the repair rung's wall-clock bound); this is its token half, so it is referenced, not re-closed.

Refs #2089 (this unblocks a clean 2/2 flip match), #2109/#2122 (the harness fixes that got the witness stage this far), #2128 (the role-cap/reasoning-budget interaction this ceiling is sized against), #2021 (the general single-call wall-clock gap).

Exemplar: mirrors the min-not-raise ceiling discipline in `apply_role_shaping`'s neighbor knobs.

Refs #2165 — the wiring-coverage gap this PR leaves behind: the ceiling is proven as a pure function, but nothing fails if its call site in `witness_engine_config` is deleted. Filed as a handoff rather than grown into this PR.

### DIFF
--- a/crates/stella-pipeline/src/pipeline/witness_stage.rs
+++ b/crates/stella-pipeline/src/pipeline/witness_stage.rs
@@ -34,6 +34,49 @@ async fn runner_availability(tests: &dyn TestRunner) -> Vec<String> {
         .collect()
 }
 
+/// Per-call output ceiling for the witness author and repair turns (#2141,
+/// #2149).
+///
+/// A witness test is a short script and its authoring reply is a marker line
+/// plus that script; the largest legitimate authoring call observed on a
+/// settled flip was under 2k output tokens. Left uncapped, the turn inherits
+/// the worker's ceiling (131k on a measured run), and a reasoning-heavy model
+/// spent it: one `witness_repair` call streamed 36,224 output tokens over
+/// 965 seconds and rode a solved trial into Harbor's agent timeout, taking
+/// the whole verification record with it. This bounds each such call to a
+/// value comfortably above legitimate need yet far below that runaway; a call
+/// that exceeds it truncates and degrades honestly through the existing
+/// no-marker path, which frees the budget instead of consuming it. It only
+/// ever *lowers* a higher ceiling — an operator's tighter `max_output_tokens`
+/// still wins.
+///
+/// This is the **token** axis, and it composes with — never replaces — the
+/// **wall-clock** axis [`witness_repair_bound`] applies to the repair turn:
+/// a cap on emitted tokens cannot end a call that stalls without emitting
+/// any, and a latency ceiling cannot stop a fast runaway from spending the
+/// budget it was granted. Both are needed, and each turn should keep the
+/// bounds it has. The repair rung carries both today; the author rung still
+/// wants the wall-clock half (#2149).
+///
+/// Sized well clear of the reasoning budget that starves the 512/1024-token
+/// role caps in #2128 — reasoning is billed against this same ceiling, so a
+/// value near legitimate output alone would trade a runaway for a silently
+/// empty authoring reply.
+const WITNESS_MAX_OUTPUT_TOKENS: u32 = 16_384;
+
+/// Lower `config.max_output_tokens` to at most [`WITNESS_MAX_OUTPUT_TOKENS`],
+/// leaving a stricter operator value untouched. Pure so the min-not-raise
+/// contract is testable without a pipeline.
+fn bound_witness_output_tokens(mut config: EngineConfig) -> EngineConfig {
+    let bounded = config
+        .max_output_tokens
+        .map_or(WITNESS_MAX_OUTPUT_TOKENS, |existing| {
+            existing.min(WITNESS_MAX_OUTPUT_TOKENS)
+        });
+    config.max_output_tokens = Some(bounded);
+    config
+}
+
 /// Overlay one role's request shaping onto an engine config — the pure half
 /// of [`Pipeline::witness_engine_config`], so the field-by-field flow is
 /// testable without a pipeline. Only the knobs `EngineConfig` itself carries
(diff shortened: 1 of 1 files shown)


macanderson:
Re: Sourcery — "should `WITNESS_MAX_OUTPUT_TOKENS` be configurable?"

Declining, deliberately: it already is configurable in the only direction a ceiling can safely move.

`bound_witness_output_tokens` runs **after** `apply_role_shaping`, so an operator who sets `verifier.max_output_tokens` (`tuning_for(&engine, EngineAgentKind::Verifier)` → `wiring.role_overrides.verifier`, `crates/stella-cli/src/agent/engine.rs:726`) already tunes this per deployment — `the_witness_output_ceiling_only_ever_lowers` pins that a stricter 2,048 stands untouched. What is not offered is raising it back above 16k, which is the escape hatch the PR exists to close: the measured failure was a single call spending 36,224 tokens over 965s and discarding a solved trial's entire verification record. A knob that can reopen that is the defect, not the fix.

Making the ceiling itself a config field would also add a `PipelineConfig` row, and `crates/stella-pipeline/src/pipeline.rs` is a god file closed to growth (AGENTS.md § "God files — plan around them, never into them"), so the knob would have to earn more than "operators might want to tune it".

The number is a judgement call from one trace, though, and that is worth tracking rather than asserting — noted on #2128, where the reasoning-token-vs-role-cap interaction that would set a *measured* value is being worked.

macanderson:
Acceptance run of this exact PR (match `f0245b91f77b`, SUT `8dd73dfa` = this branch, `nginx-request-logging` both arms, 1800s cap) — reporting the unflattering half honestly:

**The per-call bound works and is proven necessary.** Arm `stella-kimi-k3` — the arm that ran 36,224 tok / 965s and died `AgentTimeoutError` in r2 — flipped **clean** here: reward 1.0, no exception, deterministic verdict. Its heavy call hit the ceiling at **exactly 16,384 tokens / 427s** (vs 36,224/965s unbounded), then converged. Timeout → clean flip, single variable changed.

**But it is a partial fix, and the sibling arm shows the gap.** Arm `stella-glm-5-2` still timed out: its witness_repair is a *turn* of three calls — `16,384/837s` (capped), `13,851/460s`, `228/19s` — summing ~1,316s. The per-call cap fired but a multi-call repair turn is unbounded in aggregate, and one bounded call alone ran 837s at ~20 tok/s (a token cap cannot bound wall-clock).

So: this PR should merge as the per-call bound it is (it moved one arm timeout→clean-flip and prevents a single monster call), but it does **not** by itself guarantee the witness stage fits its budget. The turn-level / wall-clock bound is filed as **#2204** with this trace. Suggest this PR keep `Closes #2141` (it satisfies #2141's literal DoD — a bounded *call*) and #2204 carries the turn-level completion.

r3 result: **1/2 clean deterministic flips** (kimi clean; glm flipped-then-timed-out on the repair turn). Not the 2/2 I hoped for — reporting the real number.

PR #2185, opened 2026-08-08
fix(arenabench): let an arena running inside a checkout find it

macanderson:
## The bug

`python -m arenabench serve` from a clone — the ordinary development launch —
could not launch a Stella seat. Reproduced live on a running server before the
fix:

```console
$ curl -s 'http://127.0.0.1:8181/api/sut?ref=main'
{"ref": "main", "repo": null, ..., "ready": false,
 "problem": "no Stella checkout is configured, so the arena cannot say which
             commit a Stella seat would run. Set ARENABENCH_STELLA_REPO ..."}
```

`sut.stella_repo()` answered only from `ARENABENCH_STELLA_REPO` or
`ARENABENCH_STELLA_ADAPTER`. That bare launch sets neither, so the arena refused
every Stella seat for want of a checkout **it was executing out of** — the
refusing source file sits two directories below the repository root. The model
select degraded to free text on the same path, for the same reason.

## The fix

Three changes, one cause.

**`_discovered_repo()`** walks up from the package's own location and accepts a
tree carrying both sources `arenabench.catalog` already reads at runtime
(`crates/stella-model/src/catalog.rs`,
`bench/harbor_adapter/stella_harbor/posture.py`). The check is self-validating:
a tree with those markers is by construction one this arena can work against,
and an arena pip-installed inside some unrelated repository matches neither and
is declined rather than claimed. Anchoring on the package rather than the
working directory is deliberate — it keeps the answer tied to the code that is
executing, so two arenas on one machine resolve refs against the checkouts they
were each installed from rather than against whatever directory someone
happened to `cd` into. Precedence is unchanged: an explicit environment variable
still wins, and discovery only answers when nothing was configured.

**`repo_problem()`** gives the reporting callers the real cause. A wrong
`ARENABENCH_STELLA_REPO` used to render as "no Stella checkout is configured",
which tells an operator who set the variable to set the variable.

**`_agent_environment`** derives the Harbor adapter root from that same checkout
when `ARENABENCH_STELLA_ADAPTER` is unset. Without this the fix would be a
downgrade: the launch would now succeed into a seat that runs nothing, because
`harbor_agent.py` falls back to `_Base = object`, Harbor dies with
`ArenaStellaAgent() takes no arguments` — and **exits 0**, so the seat reports
done having made no model call.

## Witness

`TestCheckoutDiscovery` (six cases in `arenabench/tests/test_sut.py`) and
`test_an_unconfigured_arena_reads_the_catalog_it_ships_beside` all fail on
`main` and pass here. Checked in a detached shadow worktree at `origin/main`
with only the test files copied in:

```
FAILED TestCheckoutDiscovery::test_the_checkout_this_arena_runs_from_needs_no_configuration
  - AssertionError: the arena refused to name the checkout it is executing out of
FAILED TestCheckoutDiscovery::test_a_pinned_seat_launches_with_nothing_configured
  - assert None is not None
FAILED TestCheckoutDiscovery::test_an_unrelated_git_checkout_is_declined
FAILED TestCheckoutDiscovery::test_a_misconfigured_repo_env_is_not_reported_as_unconfigured
FAILED TestLiveSources::test_an_unconfigured_arena_reads_the_catalog_it_ships_beside
FAILED TestLiveSources::test_no_checkout_degrades_with_the_env_name
```

The first two are behavioral failures, not missing-symbol errors — the second is
the reported symptom exactly.

Full suite: `uv run python -m pytest` in `arenabench/` — all green (1 skipped).
`make guards-fast` green. No new `ruff` findings on the touched files (the three
it reports there are all present on `main`).

## Deleted tests

None. `test_no_checkout_degrades_with_the_env_name` is **kept** and its
assertion is unchanged; it gains a `sut._PACKAGE_DIR` patch, because deleting
both environment variables is no longer sufficient to have no checkout. The
"unconfigured degrades honestly" contract it exists to hold is still held.

## Exemplar

`stella_repo()`'s tiering follows `cargo`'s own workspace-root discovery: an
explicit setting wins, then an implicit walk upward from a known anchor, then an
honest refusal that names what to set.

## Not fixed here

Filed as follow-ups rather than widened into this PR — see the issues linked in
the thread.

## Summary by Sourcery

Automatically discover the Stella checkout an arena is running from, improve error reporting around missing or misconfigured checkouts, and ensure Harbor adapter wiring so seats launched from a checkout actually run models.

New Features:
- Add automatic discovery of the Stella repo by walking up from the installed arenabench package and recognizing repository-specific marker files.
- Infer the Harbor adapter location from the discovered checkout when no explicit adapter path is configured.

Bug Fixes:
- Fix arenas launched from a local checkout that previously could not resolve Stella refs or read the model catalog without explicit environment configuration.
- Prevent seats from appearing to complete successfully while running no model when the Harbor adapter base package was missing from PYTHONPATH.

Enhancements:
- Provide repo_problem() to report precise, user-facing explanations for why no usable Stella checkout was found instead of generic "not configured" messages.
- Unify repo lookup and error messaging via a shared _repo_search helper used by sut status, ref resolution, catalog reading, and build logic.

Tests:
- Add end-to-end and unit tests covering checkout discovery behavior, misconfiguration reporting, adapter wiring, and catalog reading both with and without a checkout.

### DIFF
--- a/arenabench/arenabench/catalog.py
+++ b/arenabench/arenabench/catalog.py
@@ -31,7 +31,7 @@ import re
 from pathlib import Path
 from typing import Any
 
-from .sut import STELLA_REPO_ENV, stella_repo
+from .sut import repo_problem, stella_repo
 
 __all__ = ["models_payload", "parse_catalog"]
 
@@ -121,8 +121,7 @@ def models_payload() -> dict[str, Any]:
     repo = stella_repo()
     if repo is None:
         raise RuntimeError(
-            f"no Stella checkout found — set {STELLA_REPO_ENV} to the "
-            "repository root so the model catalog can be read"
+            f"the model catalog cannot be read: {repo_problem()}"
         )
     entries = parse_catalog(_read(repo, _CATALOG_REL, "the model catalog"))
     if not entries:
--- a/arenabench/arenabench/runner.py
+++ b/arenabench/arenabench/runner.py
@@ -909,9 +909,20 @@ class MatchRunner:
         env.setdefault("STELLA_DISABLE_REFLECTION", "1")
 
         # Both packages must be importable by the Harbor subprocess: the
-        # ArenaBench adapter and the stella_harbor base it subclasses.
+        # ArenaBench adapter and the stella_harbor base it subclasses. Missing
+        # the second one does not fail loudly — `harbor_agent.py` falls back to
+        # `_Base = object`, Harbor dies with "ArenaStellaAgent() takes no
+        # arguments", and *exits 0*, so the seat reports done having run
+        # nothing. The adapter ships in the same checkout `sut.stella_repo()`
+        # already found, so an arena running out of a clone can wire itself
+        # rather than depend on an operator remembering one variable whose
+        # omission is invisible.
         roots = [str(Path(__file__).resolve().parent.parent)]
         adapter = os.environ.get("ARENABENCH_STELLA_ADAPTER")
+        if not adapter:
+            repo = sut.stella_repo()
+            if repo is not None and (repo / "bench" / "harbor_adapter").is_dir():
+                adapter = str(repo / "bench" / "harbor_adapter")
         if adapter:
             roots.append(str(Path(adapter).expanduser().resolve()))
         existing = os.environ.get("PYTHONPATH", "")
--- a/arenabench/arenabench/sut.py
+++ b/arenabench/arenabench/sut.py
@@ -66,6 +66,7 @@ __all__ = [
     "drift_between",
     "embedded_commit",
     "list_branches",
+    "repo_problem",
     "resolve_pin",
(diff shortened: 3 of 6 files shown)


macanderson:
Residue filed as handoffs rather than widened into this PR:

- #2192 — `harbor_agent.py` throws away the actionable `ImportError` and falls back to `_Base = object`, so a missing `stella_harbor` becomes a seat that exits 0 having run nothing. This PR removes the most common trigger (the adapter root is now derived from the discovered checkout) but not the trap: a variable set to the *wrong* path still lands there.
- #2193 — `ruff` is configured in `arenabench/pyproject.toml` and run by no gate. 7 errors and 40 unformatted files on `main`; 3 of those errors are in files this PR touches and none are introduced by it.

Issue #2220, opened 2026-08-08
arenabench: a trial that produced no telemetry at all is still scored as an agent loss — the zero-spend guard is gated on steps > 0


macanderson:
## Problem

The zero-spend guard in `MetricsReader.read`
(`arenabench/arenabench/telemetry.py`) refuses to score a trial that "ran"
without one observed token — the #1480 rule that stops a swallowed 401/429
from reading as an agent loss. But it is gated on `steps > 0`:

```python
if (
    metrics.resolved is False
    and metrics.steps > 0          # <-- never fires for a trial with no stream
    and metrics.tokens_in == 0
    and metrics.tokens_out == 0
    and metrics.total_cost == 0.0
):
    metrics.infrastructure = True
    metrics.resolved = None
```

A trial that produced **no telemetry at all** has `steps == 0`, so the guard
skips it and it stays `resolved = False` — a scored agent loss.

Observed on match `cc00894779ff` (artifacts under
`~/.arenabench/matches/cc00894779ff/`): four of ten trials — `fix-git`,
`git-leak-recovery`, `prove-plus-comm`, `sanitize-git-repo` — have an
`agent/` directory containing nothing but `setup`, `agent_result: null`, and
`exception_info.exception_type: "RuntimeError"`. The agent binary never ran.
The scoreboard reported **40% (4/10)**; over the six trials that actually
started it is **67% (4/6)**.

The *cause* of those particular RuntimeErrors is **#2127** (the server reads
harbor-adapter sources from its launch worktree, and deleting that worktree
mid-match breaks every later trial install). This issue is the **ingest-side
defence**, which is a different file and applies whatever the cause: a trial
that measured nothing was never a fair attempt, and #1480's whole argument
("classified on observed spend, because the exception name is exactly what
lied") applies with more force when there is no spend *and* no stream.

## Now cheap to fix

#2132 added `TrialMetrics.usage_measured` — a property that is `False`
exactly when no source published usage. The guard's predicate can become
`metrics.resolved is False and not metrics.usage_measured`, which drops the
`steps > 0` special case and covers both shapes with one rule.

Care needed: the guard must not swallow a genuine agent failure that
happens to have produced no tokens for an agent-side reason. Decide and state
whether `DeliberateStopError` / `SelfReportedVerificationFailureError` (see
`bench/harbor_adapter/stella_harbor/__init__.py`) should stay scored.

## Definition of done

- A trial with no usage measurement and an agent-named exception is left
[...]

Issue #2219, opened 2026-08-08
arenabench: _reduce_events computes verifier_passed and nothing reads it


macanderson:
## Problem

`_reduce_events` in `arenabench/arenabench/telemetry.py` computes a
`verifier_passed` total that **nothing ever reads**:

```python
elif kind == "verdict":
    passed = event.get("passed")
    totals["verifier_passed"] = None if passed is None else bool(passed)
```

`rg -n "verifier_passed" arenabench/` returns exactly two hits, both inside
`_reduce_events` (the `totals` initialiser and the assignment above).
`MetricsReader.read` never consults the key, `TrialMetrics` has no
corresponding field, and it reaches no payload.

Noticed while tracing the ingest for #2132.

## Why it matters (beyond dead code)

The agent's own `verdict` event is a *second, independent* opinion about
whether the work passed, beside Harbor's `verifier_result.reward`. Those two
disagreeing is exactly the shape #2066's "two facts" rule exists to keep
visible, and today the arena silently keeps only Harbor's. So this is either
a field to delete or a signal to wire — it should not stay half-built.

Note `stella_harbor/live_feed.py::_EventsCache._reduce` computes the same
value and *does* surface it (`view["verifier_passed"]`), so the two readers
of the same stream disagree about whether the field is worth having.

## Definition of done

Either:

- **delete** the key from `totals` and the `verdict` branch, if the arena has
  no use for the agent's self-reported verdict; or
- **wire** it: a `TrialMetrics` field, `to_json()`, and a witness test that a
  trial whose stream says `verdict.passed = false` while Harbor scored
  `reward = 1.0` is visible as a disagreement rather than silently resolved.

Whichever is chosen, say which in the PR and make `live_feed.py` agree.

## Files

- `arenabench/arenabench/telemetry.py` (`_reduce_events`, `MetricsReader.read`)
- `arenabench/tests/test_telemetry.py`
- `bench/harbor_adapter/stella_harbor/live_feed.py` (the second reader)

Tests: `cd arenabench && python3 -m pytest tests/test_telemetry.py -q`.
`ruff check` is red on `arenabench/` at HEAD with 7 pre-existing findings
(#2193/#2146) — none in `telemetry.py`; do not widen.

macanderson:
Closing as **not planned — frozen behind #2374 (prove value)**.

This is a *policy* close, and it is scoped narrowly. Read the distinction before
reopening, because it is the whole basis of the call.

[...]

Issue #2193, opened 2026-08-08
arenabench: ruff is configured but no gate runs it (7 errors, 40 unformatted files on main)


macanderson:
## Problem

`arenabench/pyproject.toml` configures `ruff` — line length, target version, and
a deliberate rule selection:

```toml
[tool.ruff]
line-length = 96
target-version = "py311"

[tool.ruff.lint]
select = ["E", "F", "I", "N", "UP", "B", "SIM", "RUF"]
```

Nothing runs it. `ruff` appears in no `Makefile` target, no `GATE_STEPS` entry,
and no workflow under `.github/workflows/`. `ruff` is declared in the `dev`
extra, so the intent is clearly that it gate — it just never got wired.

Measured on `origin/main` (2026-08-07, `cd arenabench && uv run ruff ...`):

- `ruff check arenabench tests` → **7 errors**, e.g. `F401` (`time` imported but
  unused in `arenabench/sut.py`), `SIM105`, `RUF012`.
- `ruff format --check arenabench tests` → **40 files would be reformatted**,
  9 already formatted.

The 40-file formatting delta is the reason this cannot simply be switched on:
turning the gate green is a large mechanical diff that will conflict with every
open arenabench branch, so it needs sequencing, not just a Makefile line.

This is the same shape as the guards AGENTS.md describes: a check that exists
but does not run is indistinguishable from no check, and the Python half of this
repository is currently ungated while the Rust half has 25 gate steps.

## What "done" looks like

Three commits, in this order, so the mechanical churn is reviewable separately
from the policy change:

1. **Fix the 7 lint errors** by hand (`F401` is a delete; `SIM105` and `RUF012`
   are one-liners). No `# noqa` without a comment saying why the rule is wrong
   *here* — CLAUDE.md's rule about `#[allow]` applies to `noqa` identically.
2. **`ruff format`** the package in one commit that touches nothing else, so a
   reviewer can confirm it is formatting-only (`git show --stat`, then
   `git diff -w` should be near-empty).
3. **Wire the gate.** Add a `python-lint` step that runs
   `uv run --project arenabench ruff check arenabench tests` and
   `ruff format --check`, register it in `GATE_STEPS`, and update the gate list
   in **both** AGENTS.md and CONTRIBUTING.md — `gate-parity`
   (`scripts/check-gate-parity.sh`) fails otherwise, and per
   `stella-gate-parity-three-copies-two-counts` the step name and the count are
   separate cells that must both be edited.

[...]

Issue #2172, opened 2026-08-08
arenabench UI: the per-seat lane still shows an unexplained '—' for an unpriced trial


macanderson:
## Problem

#2108 gave every trial an `unpriced_models` field — the model ids the shared
price table has no row for, i.e. the reason `priced_cost` is `null` — and
wired it through three of the four surfaces that render a blank cost:

- `arenabench/arenabench/telemetry.py` — `TrialMetrics.unpriced_models`, in
  `to_json()` (so it is in every trial cell payload) and summed into
  `aggregate()` totals.
- `arenabench/arenabench/cli.py` — prints `!! no pricing entry for <slug>`
  per seat after the scoreboard.
- `arenabench/ui/components/arena/stat-strip.tsx` — the match-level cost tile
  reads `Totals.unpriced_models` and names the model instead of showing a
  bare `—`.

The fourth is not wired: `arenabench/ui/components/arena/lane.tsx` renders a
per-trial `cost {fmtMoney(cell.priced_cost)}`, which is where a reader looks
to attribute a blank cost to a *seat*. It still shows an unexplained `—`, and
`Cell` in `arenabench/ui/lib/types.ts` does not declare the field even though
the server already sends it.

This is scaffolding with one consumer missing, which is the exact shape
AGENTS.md § "Nothing left behind" asks to be tracked rather than left.

## Fix

- Add `unpriced_models?: string[]` to the `Cell` interface in
  `arenabench/ui/lib/types.ts` (optional — payloads recorded before #2108 do
  not carry it), with a doc comment matching the one on `Totals`.
- In `lane.tsx`, when `cell.unpriced_models?.length`, render the slug beside
  or in place of the `—` (e.g. `cost no price for moonshotai/kimi-k3`), in
  the `warn` tone the stat-strip uses.

## Verify

Load any match whose seat ran a model with no `PRICES` row — delete the
`kimi-k3` row from `arenabench/arenabench/pricing.py` and re-read a Kimi
match, or point at a synthetic trial whose `stella-events.jsonl` names
`acme/mystery-1` — and confirm the per-trial cell says *which* model, not
just `—`.

## Constraints

- `arenabench/ui` has no test suite and no typecheck in CI (only
  `bench.yml`'s `arenabench — pytest` step runs, and it is Python only), so
  this cannot carry a witness test. Verify by rendering.
- The honesty rule in `stat-strip.tsx`'s module docstring holds: an unmeasured
[...]

PR #2222, opened 2026-08-08
fix(arenabench): an unmeasured trial is not a $0.00 trial, and a partial sum is not a seat total

### LINKED ISSUE #2132: arenabench dashboard: a trial's usage ingested as all zeros while its result.json carries $0.79 / 767K tokens (openssl-selfsigned-cert in cc00894779ff) — match totals silently under-count
## Problem

The arenabench dashboard/match API shows **all-zero tokens and cost** for a trial whose `result.json` carries full usage. In match `cc00894779ff`, trial `openssl-selfsigned-cert__5VMc6i2`:

- Dashboard cell / match API: `tokens_in: 0, tokens_out: 0, cache_read: 0, cache_write: 0, total_cost: 0, priced_cost: 0` (with `steps: 49, tools: 38` populated — so the trial row itself was ingested).
- `result.json` `agent_result`: in 725,843 / out 41,189 / $0.79373.
- Ground-truth jsonl (49 `step_usage` rows): in 973,962 / out 45,596 / $0.90645.

The other two timeout trials in the same match (`cobol-modernization__kt3nQsB`, `count-dataset-tokens__zitF9qC`) ingested non-zero usage through the same path, so the drop is conditional, not systematic — worth finding the discriminating condition (openssl is the only trial with a `witness_author` role in its usage rows, and its `stella_accounting.state` is `"incomplete"` at 42/49 rows; count-dataset is also `"incomplete"` yet displayed, so incompleteness alone is not the trigger).

Match-level totals inherit the hole: the contestant's aggregate `tokens_in`/`priced_cost` under-count by this trial's full usage, and the match's `priced_cost` (the number used for cost comparisons per the bench conventions) is silently low.

## Repro / verify

`curl -s 127.0.0.1:8900/api/matches/cc00894779ff` and compare the `openssl-selfsigned-cert` cell against `.../openssl-selfsigned-cert__5VMc6i2/result.json` (paths under `~/.arenabench/matches/cc00894779ff/jobs/cc00894779ff-stella-sonnet-5-full-pipeline/`). Artifacts are on disk; the match is finished, so the state is stable.

## Definition of done

The ingestion path that produced a zero-usage cell for a result.json carrying usage is identified and fixed with a regression test; the discriminating condition is named in the fix PR. (Bonus, separate concern: totals should prefer reconciling from `stella-events.jsonl` when accounting is flagged incomplete — see the companion adapter-timeout issue.)

macanderson:
Closes #2132

## The discriminating condition

The drop is conditional, and the condition is **whether the trial produced a
usage measurement at all** — not whether it timed out, not `witness_author`,
not `stella_accounting.state: "incomplete"`.

I could not reproduce the issue's specific reading (`openssl-selfsigned-cert`
at `tokens_in: 0`), and I think that half is a misattribution — evidence
below. What *does* reproduce, on current `main` against the artifacts the
issue names, is the failure the issue's title and third paragraph describe:
**zero standing in for "unknown", and a match total that silently
under-counts.** Two places, one habit.

### 1. Per trial — a trial nothing measured was priced at exactly `$0.00`

`MetricsReader.read` fell through to the ATIF branch for a trial with no
event stream and no trajectory, ran its zeroed counters through
`trial_cost()`, and published `priced_cost = 0.0` — a confident measurement
of "spent nothing" manufactured out of the absence of any measurement.

Four of the ten trials in `cc00894779ff` are exactly that shape
(`fix-git`, `git-leak-recovery`, `prove-plus-comm`, `sanitize-git-repo`):
an `agent/` directory holding nothing but `setup`, `agent_result: null`,
`exception_info.exception_type: "RuntimeError"`. Confirmed on `origin/main`
before touching anything:

```
task                       API in  API cost  API priced   (live server, main)
fix-git                         0       0.0         0.0
git-leak-recovery               0       0.0         0.0
prove-plus-comm                 0       0.0         0.0
sanitize-git-repo               0       0.0         0.0
```

`priced_cost: 0.0` — **not** `null`. That is precisely the shape the issue
quotes. The ingest cannot tell a trial that never launched from one whose
stream was lost, and the second kind is not hypothetical (#1467 drops usage
mid-run; #2131 reaps a timed-out trial's writer while it is still appending).
Either way `$0.00` is a claim the artifacts do not support.

### 2. Per seat — `aggregate()` published the priced *share* as the total

```python
"priced_cost": (
    sum(t.priced_cost for t in trials if t.priced_cost is not None)
    if any(t.priced_cost is not None for t in trials)
    else None
),
```

Every trial whose cost was unknown contributed nothing, and the partial sum
went out as the seat's comparable spend with no marker. On `cc00894779ff`:

| | seat `priced_cost` |
|---|---|
| `main` (live server, `curl 127.0.0.1:8900/api/matches/cc00894779ff`) | **$1.94** |
| true sum of the trials that priced | **$5.25** |

**63% short**, in the one number the bench conventions use for cost
comparisons. The discriminating condition here is that the match is *mixed*:
an all-priced seat totalled correctly and a wholly unpriced one already
reported `None`, so only a seat holding both kinds was ever wrong — which is
what a pipeline seat produces the moment one role runs a model the shared
table has no row for.

It was also the odd one out among three aggregations of the same field. The
per-trial reduction directly above it already blanks a whole trial on one
unpriced subtotal ("missing beats wrong"), `series.py::_seat_outcomes`
already null-poisons a seat, and `stat-strip.tsx` already null-poisons across
seats. `aggregate` was the single link that quietly filled the gap with
nothing.

## Why the `openssl` reading does not reproduce

Ground truth first, per the repo rule. `openssl-selfsigned-cert__5VMc6i2`'s
`stella-events.jsonl` census — 49 `step_usage`, roles
`{triage 1, plan 1, worker 43, witness_author 3, distress_guidance 1}`,
models `{claude-sonnet-5 44, kimi-k3 4, glm-5.2 1}`, `in=973,962 out=45,596
cost=$0.90645` — matches the issue's own numbers exactly. The file's mtime is
`16:36:33`, **36 minutes before the issue was filed**, and it has not been
touched since, so the artifacts I read are byte-identical to the reporter's.
Against them:

- a fresh `MetricsReader` on current `main` returns `tokens_in: 973962`;
- **both** long-lived servers (`:8900`, started 4:18PM from a since-deleted
  worktree, i.e. running older code; `:8901`) return `tokens_in: 973962`.

What openssl *did* have at the time is `priced_cost: null` — `kimi-k3` was
unpriced until #2173 landed at 19:08, two hours after the issue. And
`fmtMoney(null)` renders `—`. So the openssl cell's *cost* column was blank
(that is #2108, already fixed), while the `tokens_in: 0 … priced_cost: 0`
shape the issue quotes belongs to the four never-ran trials in the same grid.
`priced_cost: 0` is the tell: openssl's was `null`, never `0`.

I am stating this rather than quietly fixing something else. If the reporter
has a saved payload showing openssl at `tokens_in: 0`, that is a different
bug and worth reopening on.

## The fix

- **`TrialMetrics.usage_measured`** — a property (like `cache_hit_rate`
  beside it, so it cannot go stale) that is `False` when no counter and no
  cost was observed. Read off observed spend rather than off the presence of
  a file or a row, for the same reason #1480's guard is: a `step_usage` whose
  token fields never arrived is exactly as unmeasured as no `step_usage` at
  all. No real model call reports zero of everything, so "all counters zero"
  is *unknown*, never *free*.
- **`priced_cost = None`** for such a trial, never `0.0`. No `unpriced_models`
  entry — the price table is not missing a row, and sending an operator to
  edit `PRICES` for a trial that produced no tokens is the noise that makes
  the real warning easy to skip (#2173's own queued-trial rule).
- **`aggregate()["priced_cost"]` refuses a partial sum**: `None` when any
  trial *spent tokens* nobody could price. A trial that measured *nothing*
  does not poison — it has no tokens to be short by, and blanking the column
  for every match with one crashed setup would trade a wrong number for no
  number. It is counted in the new **`unmeasured_trials`** instead, so the
  denominator behind every token and cost total stays visible.
- `series.py::_seat_outcomes` learns the same distinction, so the trends
  panel does not go blank now that never-ran trials price as `None`.
- `arenabench run` prints `!! N/M trials published no usage — the token and
  cost totals above are over the rest`, beside #2173's existing unpriced line.
- `ui/lib/types.ts` declares both fields. Rendering is #2221 (below).

Result on `cc00894779ff`: `priced_cost: 5.24646278`, `unmeasured_trials: 4`,
`unpriced_models: []` — against `$1.94` and no marker on `main`.

## Witness

`arenabench/tests/test_telemetry.py::TestUnmeasuredUsageIsNotZeroUsage`
(7 tests), plus the renamed `TestAggregate` case.

Flip verified by reverting **only** the four implementation files to
`origin/main` (`git checkout origin/main -- …`, working tree and tests
preserved — no `git stash`, parallel agents are live) and re-running:

```
FAILED TestUnmeasuredUsageIsNotZeroUsage::test_a_trial_that_measured_nothing_is_not_priced_at_zero - AssertionError: unknown spend, not zero spend
FAILED TestUnmeasuredUsageIsNotZeroUsage::test_the_cell_payload_says_the_zeroes_were_never_measured - assert 0.0 is None
FAILED TestUnmeasuredUsageIsNotZeroUsage::test_a_step_that_reported_no_tokens_counts_as_unmeasured - assert 0.0 is None
FAILED TestUnmeasuredUsageIsNotZeroUsage::test_the_seat_total_refuses_a_partial_priced_sum - AssertionError: a subset is not a total
FAILED TestUnmeasuredUsageIsNotZeroUsage::test_an_unmeasured_trial_does_not_blank_the_seats_cost - KeyError: 'unmeasured_trials'
FAILED TestUnmeasuredUsageIsNotZeroUsage::test_the_totals_name_how_many_trials_measured_nothing - KeyError: 'unmeasured_trials'
FAILED TestUnmeasuredUsageIsNotZeroUsage::test_a_measured_trial_still_reports_its_cost - AttributeError: 'TrialMetrics' object has no attribute 'usage_measured'
FAILED TestAggregate::test_priced_cost_refuses_a_partial_sum_but_still_totals_the_rest - assert 1.5 is None
```

Five of the eight fail on a **value**, not a missing attribute — the
assertions were deliberately reordered to put `priced_cost is None` first
after a first pass showed three of them failing with `AttributeError`, which
proves the field is new and nothing about what the number used to be
(#2173's "passed vacuously" lesson, applied in the other direction).

All 8 pass on this branch; the whole `arenabench/tests/` suite is green
(461 passed).

**Tests changed, named as the deleted-test guard asks:**

- `TestAggregate::test_priced_cost_sums_only_the_trials_that_have_one` →
  `test_priced_cost_refuses_a_partial_sum_but_still_totals_the_rest`. It
  asserted the defect: trial `b` really spent (`total_cost = 9.0`) and could
  not be priced, so summing only `a` was the under-count. No test deleted.
- `test_series.py::_trial` gains `tokens: int = 1000`, so its fixtures stay
  the ordinary "this trial ran" case now that measured and unmeasured trials
  follow different cost rules. `test_unpriced_models_poison_the_cost_never_shrink_it`
  caught my first cut of the `series.py` change, which had dropped priced
  trials out of the sum — its assertion is unchanged.

## Relationship to #2131 / PR #2198

**Different bug, adjacent evidence — no overlap, nothing to close.** #2198
fixes the *writer* side: Harbor's deadline cancels the run, the orphaned
agent keeps appending to `stella-events.jsonl`, and the usage snapshot taken
before that under-reports. This PR fixes the *reader* side and touches no
adapter code.

They do meet on one fact, which is why `usage_measured` is read off observed
spend rather than off "is there a `step_usage` row": #2131 is a live
demonstration that a stream can stop carrying usage while the trial is still
producing rows. openssl shows the shape directly — `trajectory.json`
(16:35:32) accounts for 42 records while the event stream (16:36:33) holds
49, and `stella_accounting.state` is `"incomplete"` for that reason. A row
whose token fields never arrived is now treated as unmeasured, not as zero.

## Filed, not left

- **#2220** (P1) — the zero-spend guard is gated on `steps > 0`, so a trial
  that produced *no* telemetry is still scored as an agent loss. Those same
  four trials are why `cc00894779ff` reads 40% (4/10) where the six trials
  that actually started give 67% (4/6). `usage_measured` makes the predicate
  a one-liner; the root cause of these particular RuntimeErrors is **#2127**.
- **#2221** (P3) — `usage_measured` / `unmeasured_trials` reach the payload
  and `types.ts` but nothing renders them, so a cell whose usage was never
  observed still draws `0 tok`. Pairs naturally with #2172; `arenabench/ui`
  has no test harness (#1751), so neither can carry a witness.
- **#2219** (P3) — `_reduce_events` computes `verifier_passed` and nothing
  reads it, while `live_feed.py` computes the same value and does surface it.

Not widened: `ruff check` on the five files this PR touches is clean; the 7
pre-existing `arenabench/` findings (#2193/#2146) are untouched and
`ruff format` was not run. **#2171** (nothing gates `PRICES` against seatable
models) is the standing preventative for the unpriced half and is unaffected.

### DIFF
--- a/arenabench/arenabench/cli.py
+++ b/arenabench/arenabench/cli.py
@@ -192,6 +192,16 @@ def _cmd_run(args: argparse.Namespace) -> int:
         # one-line edit to `pricing.PRICES`.
         for slug in totals.get("unpriced_models") or ():
             print(f"    !! no pricing entry for {slug} — priced cost unavailable")
+        # The other way a total goes short: trials nothing measured. Their
+        # tokens and cost read 0 in every figure on the line above, which is
+        # the absence of a measurement and not a measurement of zero (#2132).
+        # Printed with the denominator so the line says what it is a total of.
+        unmeasured = totals.get("unmeasured_trials") or 0
+        if unmeasured:
+            print(
+                f"    !! {unmeasured}/{totals['trials']} trials published no usage "
+                "— the token and cost totals above are over the rest"
+            )
         if totals["judged"] == 0:
             worst = 1
     if args.results:
--- a/arenabench/arenabench/series.py
+++ b/arenabench/arenabench/series.py
@@ -96,7 +96,13 @@ def _seat_outcomes(metrics: dict[str, Any]) -> dict[str, Any]:
             continue
         judged += 1
         clock_time += m.clock_time or 0.0
-        if priced is not None:
+        # A trial nothing measured is skipped rather than poisoning the seat:
+        # it has no tokens to be short by, and since #2132 gave it
+        # `priced_cost = None` (it used to be a fabricated `0.0`), poisoning
+        # here would blank the trends cost for every match containing one
+        # crashed setup. A trial that *did* spend tokens and could not be
+        # priced still poisons — that spend is unknown, never smaller.
+        if priced is not None and m.usage_measured:
             priced = None if m.priced_cost is None else priced + m.priced_cost
         unpriced.update(m.unpriced_models)
         reason = m.outcome_reason() or "unlabelled"
--- a/arenabench/arenabench/series.py
+++ b/arenabench/arenabench/series.py
@@ -96,14 +96,17 @@ def _seat_outcomes(metrics: dict[str, Any]) -> dict[str, Any]:
             continue
         judged += 1
         clock_time += m.clock_time or 0.0
-        # A trial nothing measured is skipped rather than poisoning the seat:
-        # it has no tokens to be short by, and since #2132 gave it
-        # `priced_cost = None` (it used to be a fabricated `0.0`), poisoning
-        # here would blank the trends cost for every match containing one
-        # crashed setup. A trial that *did* spend tokens and could not be
-        # priced still poisons — that spend is unknown, never smaller.
-        if priced is not None and m.usage_measured:
-            priced = None if m.priced_cost is None else priced + m.priced_cost
+        if priced is not None:
+            if m.priced_cost is not None:
+                priced += m.priced_cost
+            elif m.usage_measured:
+                # Spent tokens nobody could price: unknown, never smaller.
+                priced = None
+            # else: a trial nothing measured, skipped rather than poisoning the
+            # seat. It has no tokens to be short by, and since #2132 gave it
+            # `priced_cost = None` (it used to be a fabricated `0.0`),
+            # poisoning here would blank the trends cost for every match that
+            # contains one crashed setup.
         unpriced.update(m.unpriced_models)
         reason = m.outcome_reason() or "unlabelled"
         reasons[reason] = reasons.get(reason, 0) + 1
--- a/arenabench/arenabench/telemetry.py
+++ b/arenabench/arenabench/telemetry.py
@@ -319,6 +319,33 @@ class TrialMetrics:
     wasted_elapsed: float | None = None
 
     @property
+    def usage_measured(self) -> bool:
+        """Whether any source published usage for this trial at all (#2132).
+
+        ``False`` means **nothing measured this trial** — which is a different
+        fact from "it spent nothing", and pricing it as one is how a real
+        $0.79 became a confident ``$0.00`` in a cost comparison. The artifacts
+        cannot tell a trial that never launched from one whose stream was lost,
+        and the second kind is not hypothetical: usage can be dropped mid-run
+        (#1467) and a timed-out trial's writer can be reaped while it is still
+        appending (#2131).
+
+        Read off observed spend rather than off the presence of a file or a
+        row, for the same reason the zero-spend guard in :meth:`MetricsReader.read`
+        is: a ``step_usage`` whose token fields never arrived is exactly as
+        unmeasured as no ``step_usage`` at all, and only the numbers say so.
+        No real model call reports zero of everything, so "all counters zero"
+        is "unknown", never "free".
+        """
+        return bool(
+            self.tokens_in
+            or self.tokens_out
+            or self.cache_read
+            or self.cache_write
+            or self.total_cost
+        )
+
+    @property
     def cache_hit_rate(self) -> float | None:
         """Share of prompt tokens served from cache, 0-100.
 
@@ -398,6 +425,7 @@ class TrialMetrics:
                 round(self.priced_cost, 6) if self.priced_cost is not None else None
             ),
             "unpriced_models": list(self.unpriced_models),
+            "usage_measured": self.usage_measured,
             "cache_hit_rate": self.cache_hit_rate,
             "clock_time": round(self.clock_time, 2),
             "age_s": self.age_s,
@@ -716,7 +744,21 @@ class MetricsReader:
         # share would publish a cost that is confidently short, and missing
         # beats wrong here.
         by_model = (events or {}).get("by_model") or {}
-        if any(model for model in by_model):
+        if not metrics.usage_measured:
+            # Nothing measured this trial, so there is nothing to price and no
+            # rate is missing. The branches below would have run its zeroed
+            # counters through `trial_cost` and published `$0.00` — a confident
+            # measurement of "spent nothing" produced by the absence of any
+            # measurement at all, and the one number a cost comparison must
+            # never invent (#2132). `None` is the honest answer, and
+            # `usage_measured` beside it in `to_json` is the reason.
+            #
+            # No `unpriced_models` entry: the table is not missing a row, so
+            # naming a model here would send an operator to edit
+            # `pricing.PRICES` for a trial that produced no tokens — the same
+            # discipline as the queued-trial rule below.
+            metrics.priced_cost = None
+        elif any(model for model in by_model):
             subtotal = 0.0
             for model, bucket in by_model.items():
                 if not model:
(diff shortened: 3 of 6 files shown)


macanderson:
Added one commit (`58d90d4e`, now `d088975b` after the merge): `series.py` gained **two** rules in this PR and only one was covered.

`test_unpriced_models_poison_the_cost_never_shrink_it` exercises the measured-but-unpriced branch, but nothing exercised the new one — a trial that published no telemetry at all is *skipped* rather than blanking the seat's cost. The `_trial` fixture even grew a `tokens` knob documented `pass tokens=0 for the trial that published no telemetry at all`, with no caller.

`TestOutcomeCounting::test_a_trial_that_measured_nothing_does_not_poison_the_cost` closes it, and flips on a value: on the old `series.py` it fails `assert None == 2.0` (a `priced_cost` of `None` used to set `priced = None` unconditionally), and passes at `2.0` now.

Re-verified the existing flip against main while I was here: exactly 8 tests fail on the old code, 5 of them on a value (`unknown spend, not zero spend`, `assert 0.0 is None` ×2, `a subset is not a total`, `assert 1.5 is None`) and 3 structurally. The assertion ordering does what it claims.

Issue #2231, opened 2026-08-08
Shared prompt contracts are coupled to both static prompts by convention, not by a check


macanderson:
## Problem

`crates/stella-cli/src/agent/prompt.rs` now carries **four** shared prompt-contract literals, each a `macro_rules!` block embedded verbatim by both static prompts so the two copies cannot drift:

| Macro | Issue |
|---|---|
| `tool_steering!` | #450 / #639 |
| `scope_discipline!` | #1955 |
| `measurement_discipline!` | #1957 |
| `verification_proportionality!` | #1958 |

Each is pinned by its own hand-written test asserting `PROMPTS` (the `SYSTEM_PROMPT` / `PIPELINE_SYSTEM_PROMPT` pair) both contain it verbatim. That works for the four that exist — but the *coupling itself* is unenforced. A fifth contract added later can be embedded in `SYSTEM_PROMPT` only, with no test written for it, and nothing fails: the guarantee currently depends entirely on the next author remembering the convention.

This is the same shape as the defect invariant 8 exists to prevent (`crates/stella-model/src/provider_parity.rs`): a property that must hold for *every member of a set* is declared per-member by hand, so a new member joins the set carrying none of it. The catalogue this pattern replaced had already drifted in exactly this way before #450 — `verify_done` and `ask_user` took comma splices in the pipeline copy where the base copy took an em dash.

The pipeline-mode prompt is the one that matters most here: it is what `stella run` (the default path) sends, and it is what the bench harness exercises, so a contract that lands in `SYSTEM_PROMPT` alone is invisible to every measurement the project makes decisions from.

## Where

- `crates/stella-cli/src/agent/prompt.rs` — the four macros, the two `concat!` prompt constants, and the `mod tests` block with `PROMPTS` and the four `both_prompts_*` tests.

## Suggested approach

Make the set enumerable so the property can be asserted over it rather than per-member. Two shapes that fit the file's existing idiom, in preference order:

1. **A `SHARED_CONTRACTS` slice.** Add `const SHARED_CONTRACTS: &[(&str, &str)] = &[("tool_steering", tool_steering!()), …];` beside `PROMPTS`, and replace the four near-identical tests with one that iterates the cross product and asserts every contract appears verbatim in every prompt. Adding a macro without adding its row is then the only way to escape, which is a much narrower hole than "without writing a whole test", and the row is one line a reviewer sees in the diff.
[...]

Issue #2165, opened 2026-08-08
pipeline: the witness output ceiling is proven only as a pure function — deleting its call site in witness_engine_config leaves the suite green


macanderson:
## Problem

#2162 bounds the witness author/repair engine's per-call output at `WITNESS_MAX_OUTPUT_TOKENS` (16,384) by wrapping the role-shaping result in `witness_engine_config`:

```rust
// crates/stella-pipeline/src/pipeline/witness_stage.rs
pub(super) fn witness_engine_config(&self, surface: CandidateSurface<'_>) -> EngineConfig {
    bound_witness_output_tokens(apply_role_shaping(
        self.engine_config_for(surface),
        &self.config.role_overrides.verifier,
    ))
}
```

The shipped witness, `pipeline::witness_stage::tests::the_witness_output_ceiling_only_ever_lowers`, tests the **pure helper** `bound_witness_output_tokens` in isolation (131k→16k, None→16k, 2k stays 2k). Nothing tests that `witness_engine_config` still *calls* it.

So the whole fix is one expression whose deletion leaves the suite green. Given this file's history of parallel same-seam edits (#2151 and #2162 both landed in `witness_stage.rs` within hours, from branches that forked before the other merged), that is exactly the shape that gets silently clobbered on a textually clean merge.

The same gap exists on the sibling axis: `witness_repair_bound` (#2151) *does* have an end-to-end witness (`a_stalled_witness_repair_is_bounded_and_degrades_honestly`), so the wall-clock axis is wired-and-proven while the token axis is only pure-and-proven.

## What to do

Add an end-to-end wiring witness: assert the ceiling that actually reaches the provider on the witness **author** call is the bound, not the worker's.

- New child module `crates/stella-pipeline/src/pipeline/tests/verification_hardening/witness_output_bound.rs`, declared from `verification_hardening.rs` — the pattern `flip_halt_arming` and `witness_repair_bound` already use, and the reason is the same: it reaches the shared fakes through that file's `use super::*`, and the already-oversized `src/pipeline/tests.rs` does not grow another `mod` line.
- Model the harness on `witness_repair_bound.rs`: a `Provider` whose `complete_ref` records `req.max_output_tokens` per call (`CompletionRequestRef::max_output_tokens` is `pub`, `crates/stella-protocol/src/completion.rs:261`) and answers from an ordered script — triage → worker execute → `TEST_COMMAND: ...` author reply → verdict.
[...]

Issue #1957, opened 2026-08-07
Worker accepts failed measurements as evidence: exit-0 pipelines mask fatal errors in timing/verification commands


macanderson:
## Problem

In TB2.1 `git-multibranch` (match `bbc9ea944037`), the worker twice accepted a broken measurement as evidence because the compound command exited 0:

1. `push+hook time:` came back **empty** because `bc` was not installed (`bc: command not found` on stderr) — the worker concluded "well under 3 seconds" anyway.
2. A hook-latency probe reported `archive+extract time: 70 ms` while stderr showed `fatal: detected dubious ownership in repository` followed by `tar: This does not look like a tar archive` — the timed command had produced nothing; 70ms was the time to *fail*. The worker cited the number as proof the hook is fast.

Both violate the project's own contract ("correctness is demonstrated or it is not claimed"): a measurement whose producing command errored is not a measurement.

## Where

This is steering, not tooling: the system prompts (`crates/stella-cli/src/agent/prompt.rs`) and/or the pipeline verify-stage guidance (`crates/stella-pipeline`) say nothing about invalidating evidence whose stderr carries `fatal:`/`command not found` even when the pipeline exits 0. Related surface: the bash tool result already renders `[stderr]` blocks, so the signal is present and ignored.

## Definition of done

- Prompt-level rule (shared-literal pattern per #450 / PR #1955): a claimed number is void if the command that produced it reported an error anywhere in the chain; re-measure or say "unmeasured".
- Witness test pinning the rule's presence in both static prompts, same shape as `both_prompts_scope_delivery_to_the_prompt_actually_sent`.

PR #1798, opened 2026-08-06
fix(stella-pipeline): verification-role tuning — prompt repairs, verdict negation, warrant opaque-actions, witness-stage fixes

macanderson:
## What this is

A full audit-and-fix pass over the verification agent role — the witness author, the verdict/judge call, their prompts, tool catalog, and process. Six defect clusters fixed, each with a witness test; fourteen issues filed for everything found and not fixed here (#1782–#1790, #1792–#1796).

## The defects

1. **Mangled load-bearing prompt sentences** (#1394 rename fallout). `VERIFIER_INSTRUCTIONS` shipped *"Verifier only what the evidence positively shows"* — ungrammatical in the exact clause that stops the verifier reading blind probes as findings — and the goal verifier shipped *"Verifier from EVIDENCE"*. Restored ("Judge only what…", matching the sentence's own "judging"); the guard test now asserts the **full clause**, because the fragment-level guard is what let this ship.

2. **The verdict parser credited negated verdicts** — `"The tests do not pass"` parsed as **PASS** (the mirror of the yes/no bug fixed years ago, unhandled for pass/fail). A two-token negation window reads a negated PASS as the FAIL it states; a negated FAIL ("did not fail") is skipped, never trusted as a PASS. `"No obvious issues. PASS"` still parses PASS — `"no"` is deliberately not a negator, same reason it was never a FAIL token.

3. **The warrant hole that let #1701 recur one README away.** The path-classified waivers (DocsOnly/TestsOnly/ConfigOnly/CommentsOnly/PureRemoval) never consulted `ChangeSignals`, so ten shell calls + one docs edit classified `DocsOnly` and completed `passed: true, deterministic: true, rung: Waived` — the exact outcome #1701 closed for the empty-diff shape. `ChangeSignals` gains `opaque_actions` (mutating calls whose effects the diff cannot account for: shell, processes, repo mutations, MCP, unknown names — counted at the same `ToolStart` chokepoint as `mutating_actions`), and any opaque call forfeits every path waiver. Ordinary docs edits (made of `edit_file` calls, which are mutating but diff-accountable) keep their waiver — both directions pinned.

[...]

PR #2232, opened 2026-08-08
fix(stella-cli): size verification to what the turn actually changed

### LINKED ISSUE #1958: Verification proportionality: no-op turns run full mutating e2e cycles plus destructive reset-to-pristine
## Problem

In TB2.1 `git-multibranch` (match `bbc9ea944037`), turns whose step was already satisfied on disk still ran the maximum-strength verification: install sshpass, clone, push both branches, curl both endpoints — then a **destructive reset-to-pristine** (re-init the bare repo, wipe deploy dirs) justified only by a guess about grader behavior ("the grader will clone fresh"). This ran four times in one task. Costs and risks:

- Three full clone/push/reset cycles where a read-only probe (`git rev-parse --is-bare-repository`, `openssl x509 -noout`, `nginx -t`) would have sufficed for the no-op turns.
- Each reset destroys verified-working state and is a fresh opportunity to break it (the hook only survived because the worker was careful each time).
- Roughly half the run's spend was re-verification of unchanged state (front-run rework aside — that half is fixed by PR #1955).

The distinction to encode: **verification that reads is nearly free; verification that mutates the system under test carries restore-risk and needs a reason.** A turn that changed nothing gets a read-only probe; a turn that mutated state earns one mutating e2e; a destructive "cleanup" of working state needs evidence the consumer requires it, not a hunch.

## Where

Steering in `crates/stella-cli/src/agent/prompt.rs` (shared-literal pattern, see PR #1955) and/or the pipeline verify-stage prompts in `crates/stella-pipeline`. Related: the verification-ladder abstain rung work (#1782–#1798 cluster) is the same proportionality idea on the verdict side.

## Definition of done

- Prompt rule distinguishing read-only vs mutating verification and forbidding speculative destructive resets of verified state.
- Witness test pinning it in both static prompts.

macanderson:
## What

Adds a fourth shared literal to both static system prompts (`SYSTEM_PROMPT` and `PIPELINE_SYSTEM_PROMPT` in `crates/stella-cli/src/agent/prompt.rs`): **`verification_proportionality!()`** — verification is sized to what *this* turn changed, a check that only reads is not the same instrument as one that mutates the system under test, and working state is never reset to "pristine" on a guess about a later consumer.

Closes #1958

## Why

The third finding from the same TB2.1 `git-multibranch` trace (match `bbc9ea944037`) that produced #1955 (`scope_discipline!`) and #1957 (`measurement_discipline!`). Turns whose step was already satisfied on disk still ran the maximum-strength check — install sshpass, clone, push both branches, curl both endpoints — and then a **destructive reset-to-pristine** (re-init the bare repo, wipe the deploy dirs) justified only by a hunch about the grader ("it will clone fresh"). That ran four times in one task; three of those turns had changed nothing, where `git rev-parse --is-bare-repository` / `nginx -t` / `openssl x509 -noout` would have settled the same claim for free. Each reset destroyed verified-working state and was a fresh opportunity to break it — the setup survived only because the worker happened to rebuild it correctly every time.

Neither prompt drew any line at all between reading and mutating (confirmed on `main`: `rg -in "read-only|destructive|pristine|reset|proportion" crates/stella-cli/src/agent/prompt.rs` matches nothing in either constant — the single hit is an unrelated doc comment on `append_project_orientation`).

## The cheaper rung is never a silent skip

This is the point the change turns on, so it is written into the literal itself rather than left to inference. The last sentence reads:

> Taking the cheap path is never silent — name the probe you ran and the claim it settles, so an end-to-end cycle you did not run is a stated decision rather than an omission.

So proportionality produces a *narrower stated claim*, never an absent one: the worker must name the read-only probe and what it settles, which lands in the turn's own output and therefore in the evidence the verify stage reads. That is the prompt-side counterpart of the discipline the ladder already keeps on the verdict side — `LadderDecision::NothingAttempted` (`crates/stella-pipeline/src/verify.rs`) fails *closed* on a turn that dispatched zero mutating actions precisely so abstaining cannot become a hiding place for a no-op. Nothing here lets a turn verify less than it claims; it lets a turn stop paying for a mutating cycle it has no change to justify, out loud.

## Scope note — why this is prompt-side and not a new ladder rung

The issue's own "Where" section and definition of done name the static prompts, and that matches the trace: the mutating e2e cycles and the reset-to-pristine were the **worker's own shell commands**, not something the pipeline dispatched. The verdict-side proportionality the issue calls "the same idea" (#1782–#1798) already exists as the abstain / `NothingAttempted` rungs, and duplicating it as a second decision function would have been a competing authority over the same question. No new state, no new dependency, no code path changed.

## How

Follows the exemplar already in the file — `tool_steering!()` (#450), then `scope_discipline!()` (#1955) and `measurement_discipline!()` (#1957): one shared literal embedded verbatim by both prompts so the copies cannot drift, kept a compile-time `concat!` so the byte-stable-prefix property (invariant 7, L-E8) is preserved. Static text only; the prompt-cache golden fixtures assert zone structure rather than prompt bytes and are unaffected.

## Witness

**`both_prompts_size_verification_to_what_the_turn_changed`** (`crates/stella-cli/src/agent/prompt.rs`).

Flip evidence, checked the artisanal way — with only the two `verification_proportionality!()` embeddings stripped from the prompt constants (macro and test kept in place):

```
test agent::prompt::tests::both_prompts_size_verification_to_what_the_turn_changed ... FAILED
    SYSTEM_PROMPT does not embed the shared verification-proportionality block verbatim
test result: FAILED. 7 passed; 1 failed
```

and with the change:

```
running 8 tests ... test result: ok. 8 passed; 0 failed
```

Only the new test flips; the other seven in the module are unmoved. The witness also pins five clauses of the literal separately so it cannot pass against a hollowed-out block — both halves of the read/mutate distinction (`only READS`, `MUTATES the system under test`), the mapping that makes it a rule (`A turn that changed nothing gets the read-only probe`), the destructive-reset prohibition, and the visibility clause. The second commit exists because the first version of the witness pinned only the mutating half: a trim that kept the prohibition and dropped the cheaper rung would have left "do not mutate" with nowhere to fall to, and the assertion would still have passed.

Wider check: `cargo test -p stella-cli --bin stella prompt` → 46 passed, 0 failed. `cargo fmt --check -p stella-cli` and `cargo clippy -p stella-cli --all-targets -- -D warnings` clean. `scripts/check-file-size.sh` (`prompt.rs` is 786 lines, not grandfathered, ceiling 1500), `scripts/check-left-behind.sh` and `scripts/check-brand-case.sh` all pass.

## Hot seams — not touched

`git diff --name-only origin/main...HEAD` is exactly one file, `crates/stella-cli/src/agent/prompt.rs`. Neither `crates/stella-pipeline/src/pipeline/witness_stage.rs` (#2151, #2162, #2191, #2212) nor `crates/stella-pipeline/src/verify.rs` (#2216) is in the diff. Confirmed at the call sites regardless, per #2165: `witness_repair_bound` is still computed and applied through `tokio::time::timeout` around the repair turn (`witness_stage.rs:499`, `:539`), and `bound_witness_output_tokens` is still wrapped around `apply_role_shaping` inside `witness_engine_config` (`witness_stage.rs:267`).

## Tests deleted

None.

## Nothing left behind

No TODOs added. Follow-up filed: #2231 — the four shared prompt-contract literals are pinned only by same-file `contains` assertions, so nothing checks that a *new* contract added later gets embedded in both prompts rather than one.

## Summary by Sourcery

Introduce a shared verification-proportionality contract in Stella’s static system prompts to align verification effort with what each turn actually changes and ensure this discipline is explicitly encoded and tested.

Enhancements:
- Add a shared verification-proportionality prompt literal and embed it into both static system prompts to distinguish cheap read-only probes from riskier mutating checks and discourage destructive reset-to-pristine behavior.

Tests:
- Add a prompt witness test that asserts both prompts embed the full verification-proportionality block and retain its key clauses, preventing future drift or hollowing-out of the contract.

### DIFF
--- a/crates/stella-cli/src/agent/prompt.rs
+++ b/crates/stella-cli/src/agent/prompt.rs
@@ -643,19 +643,30 @@ mod tests {
     /// fresh chance to break verified-working state.
     ///
     /// Neither prompt drew any line between reading and mutating, so pin the
-    /// shared literal verbatim in both — plus the three claims a later trim
-    /// would take first, so the assertion cannot pass against a hollowed-out
-    /// literal. The visibility claim is the load-bearing one: proportionality
-    /// that may be taken silently is just "skip verification" with a nicer
-    /// name, and this repository's whole contract is verified done, not
-    /// claimed done.
+    /// shared literal verbatim in both — plus each clause of the rule
+    /// separately, so the assertion cannot pass against a hollowed-out literal.
+    /// Both halves of the read/mutate distinction are pinned, not just the
+    /// mutating one: a trim that kept "a check that MUTATES … can break state"
+    /// and dropped the sentence naming what to do instead would leave a
+    /// prohibition with no cheaper rung to fall to, which is how a
+    /// proportionality rule turns into a reason to verify nothing.
+    ///
+    /// The visibility claim is the load-bearing one: proportionality that may
+    /// be taken silently is just "skip verification" with a nicer name, and
+    /// this repository's whole contract is verified done, not claimed done.
     #[test]
     fn both_prompts_size_verification_to_what_the_turn_changed() {
         let shared = verification_proportionality!();
         for claim in [
-            // The distinction the issue asks to encode: reading is nearly
-            // free, mutating the system under test carries restore-risk.
+            // The distinction the issue asks to encode, both halves of it:
+            // reading is nearly free, mutating the system under test carries
+            // restore-risk.
+            "only READS",
             "MUTATES the system under test",
+            // The mapping that makes it a rule rather than an observation —
+            // a no-op turn gets the cheap probe, a real change earns the
+            // end-to-end run.
+            "A turn that changed nothing gets the read-only probe",
             // The destructive half — a speculative reset of working state.
             "Never reset working state to \"pristine\"",
             // Degradation warns, never silently disables: the cheaper rung is
--- a/crates/stella-cli/src/agent/prompt.rs
+++ b/crates/stella-cli/src/agent/prompt.rs
@@ -74,6 +74,32 @@ macro_rules! measurement_discipline {
     };
 }
 
+/// The verification-proportionality contract shared by both static prompts:
+/// re-verification is not free, and a check that mutates the system under test
+/// is not the same instrument as one that only reads it. Born from the same
+/// real TB2.1 `git-multibranch` trace as the two literals above (#1958): turns
+/// whose step was already satisfied on disk still ran the maximum-strength
+/// check — install sshpass, clone, push both branches, curl both endpoints —
+/// and then a **destructive reset-to-pristine** (re-init the bare repo, wipe
+/// the deploy dirs) justified only by a guess about the grader ("it will clone
+/// fresh"). That ran four times in one task. Three of those turns had changed
+/// nothing, so `git rev-parse --is-bare-repository` / `nginx -t` /
+/// `openssl x509 -noout` would have settled the same claim for free, and each
+/// reset destroyed verified-working state for a hunch — the setup survived
+/// only because the worker happened to rebuild it correctly every time.
+///
+/// The cheap rung is a *stated* choice, never a silent omission: the last
+/// sentence is what keeps proportionality from degrading into "skip
+/// verification", the same discipline the ladder's abstain rung keeps on the
+/// verdict side (`stella_pipeline::verify`). One shared literal, embedded
+/// verbatim by both prompts, same as `tool_steering!` and for the same
+/// anti-drift reason (#450).
+macro_rules! verification_proportionality {
+    () => {
+        r#"Verification is proportional to what THIS turn changed, and re-verification is not free. A check that only READS (`git rev-parse`, `nginx -t`, `openssl x509 -noout`, reading back the config you wrote) costs almost nothing and risks nothing; a check that MUTATES the system under test — installing packages, cloning, pushing, restarting a service, re-initializing a repository — spends real time and can break state that already worked. A turn that changed nothing gets the read-only probe; a turn that did change state earns one end-to-end run of what it changed. Never reset working state to "pristine" because you guess some later consumer wants a clean slate: destroying verified-working setup needs a requirement that says so, and on a hunch it is only a fresh chance to break what already passed. Taking the cheap path is never silent — name the probe you ran and the claim it settles, so an end-to-end cycle you did not run is a stated decision rather than an omission."#
+    };
+}
+
 pub(crate) const SYSTEM_PROMPT: &str = concat!(
     r#"You are Stella, a fast terminal coding agent. You help the user with software engineering tasks by reading files, writing code, running commands, and searching the codebase.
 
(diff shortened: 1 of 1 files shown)


macanderson:
Heads up on a merge interaction — this PR and #2241 (closing #2231) are textually clean against each other and together turn `main` **red**.

#2241 adds `crates/stella-cli/src/agent/prompt/parity.rs`, which derives the set of shared prompt contracts from `prompt.rs`'s own `macro_rules!` declarations and asserts every one is embedded in both static prompts. `verification_proportionality!` is a new contract, so once both land it is discovered by the scan but has no row in `SHARED_CONTRACTS`, and `the_registries_match_the_source` fails with:

```
prompt.rs declares the contract `verification_proportionality`, which
prompt/parity.rs SHARED_CONTRACTS does not list — add its row.
```

Whichever of the two merges second needs exactly one line in `crates/stella-cli/src/agent/prompt/parity.rs`:

```rust
("verification_proportionality", verification_proportionality!()),
```

Nothing else changes — the contract itself, both embeddings, and this PR's witness test are all untouched, and #2241 deletes no test. If this PR merges first, #2241 will pick the row up on its rebase.

macanderson:
**Merge ordering: this PR should land before #2241.**

Confirming the interaction above from this side, and naming the direction — "whichever merges second" leaves the choice open, and only one of the two orders is free.

`prompt/parity.rs` on #2241 derives the contract set from `prompt.rs`'s own `macro_rules!` declarations and checks it against a hardcoded `SHARED_CONTRACTS`, which today holds exactly three rows:

```rust
("tool_steering", tool_steering!()),
("scope_discipline", scope_discipline!()),
("measurement_discipline", measurement_discipline!()),
```

`assert_registry` fails by name on a declared-but-unlisted contract, so once `verification_proportionality!` exists in `prompt.rs` the derived scan finds a fourth and `the_registries_match_the_source` goes red. (`every_shared_contract_is_invoked_by_every_static_prompt` is *not* affected — this PR does embed the literal in both prompts. It is only the registry row that is missing.)

**This PR cannot pre-empt it.** `crates/stella-cli/src/agent/prompt/parity.rs` does not exist on this branch, so there is no file here to add the row to. Only #2241 can carry it. That makes the ordering asymmetric rather than a coin flip:

- **#2232 → #2241** (preferred): this PR merges unchanged. #2241 updates its branch, its own new check goes red *by name*, and it adds `("verification_proportionality", verification_proportionality!()),`. The check failing there is the check working.
- **#2241 → #2232**: this PR would have to update its branch to pull `parity.rs` in, add the row, and pay another full `fmt + clippy + test` cycle for a change it is already green on.

Worth noting which guard holds either way: with `strict: true` on `main`, whichever lands second is forced to update its branch and re-run `fmt + clippy + test` — which runs the parity tests — so the red surfaces *before* the merge rather than after. The real exposure is a hand merge with `enforce_admins` off, which is why the ordering is written down rather than left to the merge button.

No change to this PR's contract, either embedding, or the five-clause witness.

Issue #2234, opened 2026-08-08
serve: SSE frames carry no wall-clock stamp — the forensics gap #2111 closed for the CLI is still open for a host


macanderson:
## Problem

#2111 gave the CLI's stream-json sinks a wall-clock anchor: every line written by
`crates/stella-cli/src/agent/output.rs::ordered_durable_event_sender` (the durable
`stella-events.jsonl`) and by `crates/stella-cli/src/agent/persistence.rs::emit_stream_json`
(stdout) now carries a `ts` — epoch millis — built through
`stella_protocol::journal::stamped_line`.

`stella-serve` did not get one. A host driving the engine over SSE receives
`ServerFrame`s carrying a bare `AgentEvent`, so it has exactly the gap #2111
closed for the CLI: `duration_ms` names a span but nothing names an instant, and
an idle stretch before a stall is invisible because there is nothing to subtract.

The committed contract already anticipates it. `docs/wire/agentevent.schema.json`
declares `ts` as an **optional** property of every root variant — "may be present"
is the only claim true of all three surfaces — so adding it on the serve side is
purely additive and needs no schema change to the event artifact. Whether it
belongs on the `AgentEvent` line or on the `ServerFrame` envelope (which already
carries the transport-added `seq`) is the design question this issue has to
answer; the envelope is the better fit if a host wants to distinguish "when the
engine emitted it" from "when the transport delivered it", and `ServerFrame` has
its own committed artifact (`docs/wire/serveframe.schema.json`) to regenerate if
so.

## Files

- `crates/stella-serve/src/frame.rs` — `ServerFrame`, the envelope. Its `seq` is
  added by the transport at delivery time and pinned by `envelope_pin` in
  `crates/stella-serve/src/history.rs`; a `ts` would sit beside it and must join
  that pin.
- `crates/stella-serve/src/observe/sink.rs` — already has `Line { ts, level, event }`
  for the serve *log* sink, using
  `SystemTime::now().duration_since(UNIX_EPOCH)`. That is the shape, but it is a
  different plane (stderr diagnostics), and duplicating its inline `now()` rather
  than taking a clock port would be the wrong copy.
- `crates/stella-cli/src/runtime.rs::WallClock` — the `stella_core::ports::Clock`
  impl the CLI sinks use. `crates/stella-serve/src/pipeline_run.rs:150` already
  carries a private duplicate `WallClock`; consolidating is part of this work.
[...]

Issue #2135, opened 2026-08-08
bench: per-task wall-clock deadline never acted — every timeout in cc00894779ff was the harness SIGKILL at 900s, never an engine-side stop with partial work (re-verify #1481/#1503 on the harbor path)


macanderson:
## Problem

Per-task wall-clock deadline enforcement did not act in bench runs on a same-day SUT: every timeout in match `cc00894779ff` (SUT 447f3393, 2026-08-07) came from the **harness** (`AgentTimeoutError` at exactly 900.0s), never from the engine stopping itself with partial work — which is what #1481/#1503/#1507 (all closed) were meant to provide via `BudgetGuard::set_task_deadline`.

Observed:

- `cobol-modernization__kt3nQsB`: killed mid-reasoning at 900s having never written the deliverable. `budget_tick` fired 44 times against the $15 dollar budget (actual spend $0.87 — never binding); nothing tracked the 900s wall clock, and no signal caused a pivot to "write the artifact now". The trace shows the agent had derived the correct root cause in its final reasoning tokens — a deadline-aware stop at even t=850s with a forced deliverable write plausibly scores this task.
- `extract-elf__G9aMhUC`, `count-dataset-tokens__zitF9qC`, `openssl-selfsigned-cert__5VMc6i2`: all SIGKILLed mid-step/mid-tool by the harness, leaving `stream_complete: false` and truncated accounting (see companion adapter issue).

Either (a) the arenabench/harbor path never calls `set_task_deadline` (the adapter knows the task's `timeout_sec` from task.toml and could pass it), or (b) it is set but the guard's deadline rung doesn't fire/pivot. Per the closed-issue caveat this needs re-verification on the live path, not an assumption that closing #1503 shipped the wiring end-to-end.

## Evidence

- `~/.arenabench/matches/cc00894779ff/jobs/.../cobol-modernization__kt3nQsB/agent/stella-events.jsonl` — 44 `budget_tick` events, none deadline-related; run ends as a raw stream cut, not an engine stop.
- Contrast `log-summary-date-ranges__Z338iwW`, where the engine *did* stop itself (exit code 3, loop escalation) — proving the deliberate-stop path works when a guard actually fires.

## Definition of done

In a bench trial with a task timeout, the engine stops itself before the harness deadline (with the settle/flush path running, so accounting is complete), and a trace field shows the deadline was armed. If the wiring already exists, the fix names why it didn't arm here and adds the missing witness test on the harbor path.

## Related

[...]

PR #2235, opened 2026-08-08
protocol: stamp every stream-json line with the sink's wall clock

### LINKED ISSUE #2111: protocol: stella-events.jsonl has no timestamps — arena elapsed column shows 0:00 and hang forensics are impossible
## Problem

`stella-events.jsonl` events carry no absolute timestamp — `step_usage`/`tool_result` have `duration_ms`, but nothing anchors any event to wall clock. Two consequences observed 2026-08-07:

- The arena transcript view renders `0:00` as the elapsed column for every event of a trial (nothing to compute from), which reads as broken to anyone watching a live match.
- Trace forensics cannot answer "where did the wall clock go" — e.g. the idle-to-AgentTimeoutError hang (see companion issue) cannot be localized from the journal because the gap between the last event and the timeout is invisible.

## Fix

Stamp each emitted event with a monotonic-derived wall-clock field (e.g. `ts` as epoch millis) at the event bus boundary — one field, one place (`crates/stella-protocol/src/event.rs` + the emitter in `stella-cli`/pipeline event sink). Byte-stable prompt concerns don't apply: the journal is output telemetry, not model input. Wire-schema gate will require regenerating `docs/wire/` in the same PR.

## Definition of done

New trials' `stella-events.jsonl` lines carry `ts`; the arena transcript elapsed column renders real offsets. Witness: a serde round-trip/emission test asserting the field is present on emitted events (fails on current main).

macanderson:
## What & why

`stella-events.jsonl` anchored nothing to wall clock. `step_usage` and `tool_result`
carry a `duration_ms` — which names a *span* but no *instant* — so two questions
could not be answered from a recorded trial at all: when did this happen, and where
did the wall clock go between these two events. The visible symptom was the arena
transcript rendering `0:00` in the elapsed gutter for every row of a finished trial,
which reads as a broken feed to anyone watching a live match. The forensic symptom
was worse: the idle-to-`AgentTimeoutError` gap in #2110 is invisible because there
is nothing to subtract.

Closes #2111

### The field

`ts` — `u64`, milliseconds since the Unix epoch (UTC), the instant **the sink that
wrote the line admitted it**. Not a member of any `AgentEvent` variant: a stamp is a
fact about a *write*, the same event is written to more than one sink, and neither
`stella-core` (a pure, replayable fold) nor `stella-protocol` (zero I/O by charter)
owns a clock. Putting it in the enum would have meant a field on all 39 variants, a
clock reachable from the engine, and a timestamp baked into every replay fixture.

So it rides an envelope applied at the sink — `stella_protocol::journal` — that
**flattens**, so the bytes gain one key rather than a nesting level:

```json
{"ts":1754582400123,"type":"tool_result","call_id":"c1","duration_ms":4200}
```

Every existing reader (`arenabench/arenabench/telemetry.py`,
`bench/harbor_adapter/stella_harbor/live_feed.py`, `exit_cause.py`) keys off
`event["type"]` at the top level and is untouched by the addition. The clock is the
`stella_core::ports::Clock` port, and specifically `crates/stella-cli/src/runtime.rs::WallClock`
— which already existed for exactly this ("stamps that must stay comparable across
processes and runs"), and whose doc already says `SystemClock`'s per-construction
origin is "exactly the wrong one for a durable timestamp". Tests inject a fixed
clock, so no golden becomes time-dependent.

Both stream-json sinks stamp: the durable JSONL file
(`agent/output.rs::ordered_durable_event_sender`) and the stdout copy
(`agent/persistence.rs::emit_stream_json`). They read the clock independently, so the
two copies of one event differ by the persist. That is the truth and is why the field
is documented as "when this sink admitted the line" rather than "when the event
happened"; the alternative — one stamp shared by both — would mean changing the
`mpsc::UnboundedSender<AgentEvent>` channel type, i.e. changing `stella-core`'s
`EventSender` API and touching `bus.rs`, which is a god file at its exact ceiling.
`ordered_durable_event_sender`'s doc comment states the consequence: nothing may diff
the two streams byte-for-byte.

**Deliberately not stamped:** `stella_pipeline::replay::to_jsonl`. That journal is a
golden fixture format and must stay a pure function of the run.

### Backward compatibility

`StampedEvent.ts` is `Option<u64>` with `#[serde(default, skip_serializing_if = "Option::is_none")]`.

- **Read side, optional forever.** Every `stella-events.jsonl` under `bench/evidence/`
  has no `ts`, and a reader that required it would refuse to replay the archive.
  Witnessed by `a_line_recorded_before_the_stamp_existed_still_parses`.
- **Write side, always stamped.** `stamped_line` takes a `u64`, not an option, so a
  sink cannot silently forget.
- **A `None` serializes to exactly the bytes a bare `AgentEvent` produces**
  (`an_unstamped_line_is_byte_identical_to_the_bare_event`). That identity is why
  this is an added key rather than a new format — no version negotiation exists or
  is needed.
- **`AgentEvent::Unknown` keeps both halves.** An event from a newer stella
  re-serializes with its body *and* its stamp intact
  (`an_event_from_a_newer_stella_keeps_both_its_body_and_its_stamp`) — forward-compat
  and forensics must not be mutually exclusive on exactly the lines that need both.
- No committed fixture or golden changed. `crates/stella-protocol/tests/text_event_wire.rs`,
  which pins exact bytes for `text`/`text_delta`, is untouched and green — `AgentEvent`'s
  own serialization is unchanged.

### Prompt-bound content: unaffected (invariant #7)

**The stamp never enters the system prompt or any cached prefix.** It is applied at
the *output* sink — `agent/output.rs` and `agent/persistence.rs`, both on the write
path to a file or to stdout — and nothing on the prompt-assembly path
(`agent.rs::build_system_prompt`, `memory.rs`) reads a clock or sees a `StampedEvent`.
`stella-protocol` still carries no time dependency; the only clock is `WallClock` in
`stella-cli`. Prompt-cache hit rates are untouched.

### Regenerated wire artifacts

Yes — `docs/wire/agentevent.schema.json` and `docs/wire/agentevent.d.ts`, via
`bash scripts/export-agentevent-schema.sh`, never by hand.
`bash scripts/check-wire-schema.sh` is green.

`schemars` cannot derive the key, because it is not on the type. `schema_export::attach_journal_stamp`
adds it in one uniform pass over the 39 root variants, with its prose taken from
`journal::TS_DESCRIPTION` so the published contract and the Rust doc are one string.
The rejected alternative was flattening an envelope struct into the schema root: that
turns the root from a discriminated union into an object, and takes `KnownTypeTag`
with it — the one thing a forward-compatible consumer most needs. Optional on every
variant is also the only claim true of all three surfaces the artifact covers: an
archived line has none, and `stella-serve` frames the event in an envelope that stamps
its own (follow-up #2234). The reasoning is recorded in `docs/wire/README.md` beside
the existing "one hand-written line" note, which is the precedent for a
transport-added key.

Read the diff: it is `ts?: number` added to each variant's `properties`, never to
`required`, and nothing else moved.

### The consumer half — the `0:00` the issue leads with

`arenabench/arenabench/telemetry.py::TranscriptReader` fabricated `t` from
`time.time()` at parse. A finished trial is read in one pass, so every event in the
file was parsed within the same few milliseconds, every `t` rounded to `0.0`, and
`fmtClock(0)` rendered `0:00` down the whole gutter. It reads the line's own `ts` now
and measures from the stream's first stamp. The UI needed no change:
`transcript-page.tsx` and `lane.tsx` already render `entry.t`, and the paced replay
already clamps its gap to `[16 ms, 1200 ms]`, so a real 500-second idle gap paces
correctly instead of stalling.

## The witness

**Primary — `arenabench/tests/test_telemetry.py::TestTranscriptElapsed`.** Pure
Python, compiles on `main`, fails there at runtime on exactly the reported symptom.
Flip verified by `git checkout origin/main -- arenabench/arenabench/telemetry.py` and
re-running:

```
FAILED TestTranscriptElapsed::test_elapsed_comes_from_the_events_own_stamps
  assert [0.0, 0.0, 0.0] == [0.0, 12.5, 90.0]
FAILED TestTranscriptElapsed::test_an_idle_gap_before_a_timeout_is_visible
  assert (0.0 - 0.0) == 500.0
FAILED TestTranscriptElapsed::test_incremental_reads_share_one_origin
  assert [0.0] == [60.0]
```

`assert [0.0, 0.0, 0.0] == [0.0, 12.5, 90.0]` **is** the `0:00` column, reproduced.
With the fix restored: 56 passed in `test_telemetry.py`, 521 passed + 2 skipped in
`bench/harbor_adapter/tests/`. The other two cases in the class
(negative-delta clamp, unstamped archive) pass on `main` trivially and are regression
guards, not witnesses — named here so the distinction is not overclaimed.

**Protocol — `crates/stella-protocol/tests/journal_stamp_wire.rs`** (7 tests, reading
the line as JSON, the way a consumer does). Flip verified by removing
`src/journal.rs` and `git checkout origin/main -- src/lib.rs`:

```
error[E0432]: unresolved imports `stella_protocol::StampedEvent`, `stella_protocol::stamped_line`
  --> crates/stella-protocol/tests/journal_stamp_wire.rs:20:35
```

**Emitter — `agent::output::durable_stream_tests::every_durable_line_is_anchored_to_wall_clock`**,
asserting the durable sink's own bytes carry `ts` under a `FixedClock`.

## The gate

- [x] `cargo fmt --check` (via `make guards-fast`)
- [x] `cargo clippy -p stella-protocol --all-targets --features schema -- -D warnings`; `cargo clippy -p stella-cli --all-targets -- -D warnings`
- [x] `RUSTDOCFLAGS="-D warnings" cargo doc -p stella-protocol -p stella-cli --no-deps`
- [x] `cargo test -p stella-protocol --features schema` (146), `-p stella-cli --bin stella` (1479), `-p stella-pipeline replay` (55)
- [x] `make guards-fast` — every toolchain-free guard, including `file-size`, `god-files`, `left-behind`, `module-reachability`, `doc-links`
- [x] `bash scripts/check-wire-schema.sh`
- [x] Docs updated: `website/content/docs/event-stream-compatibility.mdx`, `docs/wire/README.md`, `crates/stella-protocol/README.md`
- [x] `Closes #2111` appears above **and** as a commit trailer

Scoped deliberately, not `--workspace`: CI is the gate and the dev machine is shared.

## Nothing left behind

- [x] Filed: **#2234** — `stella-serve`'s SSE frames carry no stamp, so a host driving
      the engine still has the gap this PR closed for the CLI. Written as a handoff:
      files, the `ServerFrame`-vs-line design question, `envelope_pin` as the tripwire,
      and the wire-schema obligation.
- Commented on **#2110** with the `jq` one-liner the new field makes possible, and the
  honest caveat that it does not retro-explain `cc00894779ff` — archives have no `ts`.
- **#2135** (per-task deadline) is related and untouched; linked, not duplicated.

## Ground-rule check

- [x] No I/O added to `stella-core`; the clock is the existing `ports::Clock` port,
      implemented in `stella-cli` where concrete time sources belong. No new deps.
- [x] New cross-boundary type round-trips through serde — `a_stamped_line_round_trips_byte_for_byte`,
      plus the byte-identity and legacy-parse cases above (invariant #4).

**The one thing worth arguing about** is the two independent stamps on the durable
file and its stdout copy. It is documented at the sink and in the wire contract, and
unifying them is a `stella-core` API change reaching into a god file at its exact
ceiling. Say the word and I will file it rather than leave the judgement in a PR
comment.

## Summary by Sourcery

Stamp each stream-json journal line with a wall-clock timestamp at the sink and expose it in the protocol, then consume it in the arena transcript to compute realistic elapsed offsets while keeping older archives and consumers compatible.

New Features:
- Introduce a `StampedEvent` envelope and `stamped_line` helper that add an optional wall-clock `ts` to serialized `AgentEvent` lines.
- Extend the wire schema and TypeScript declarations so every `AgentEvent` variant may carry an optional `ts` field for sink-stamped timestamps.

Bug Fixes:
- Fix arena transcript elapsed times always rendering as `0:00` by reading per-line `ts` stamps instead of parse-time wall clock.
- Ensure transcript elapsed offsets remain non-negative and that streams recorded before `ts` existed still render using the old parse-time behavior.

Enhancements:
- Stamp both durable JSONL evidence and stdout stream-json with sink-local wall-clock instants using the CLI `WallClock`, and document that the two streams may differ.
- Clarify event-stream compatibility docs with the semantics and caveats of the `ts` stamp, including non-monotonic clocks and stream-local offsets.
- Add protocol-side tests and CLI durable-stream tests that validate presence, shape, and round-trip behavior of the `ts` stamp across legacy and future events.

Documentation:
- Document the `ts` journal stamp and its compatibility guarantees in the wire README, schema banner, and event-stream compatibility docs.
- Update the stella-protocol README to describe the new `journal` module and its stamped event line type.

Tests:
- Add arena telemetry tests that exercise elapsed computation from `ts`, visibility of idle gaps, clock step handling, legacy unstamped streams, and incremental reads.
- Add protocol tests verifying that stamped lines include `ts`, remain flat, parse without `ts`, round-trip byte-for-byte, and preserve stamps on unknown events.
- Add CLI durable stream tests ensuring every persisted line is stamped with wall clock and remains flat for existing readers.

### DIFF
--- a/arenabench/arenabench/telemetry.py
+++ b/arenabench/arenabench/telemetry.py
@@ -923,6 +923,12 @@ class TranscriptState:
     offset: int = 0
     seq: int = 0
     started: float | None = None
+    #: The first line's own ``ts`` (epoch millis), when the stream carries one.
+    #: Every later entry's ``t`` is measured from this, so a finished trial read
+    #: in one pass reports the offsets the run actually had rather than the
+    #: offsets of the *read* (#2111). ``None`` for a stream recorded before the
+    #: field existed, which falls back to ``started``.
+    origin_ms: float | None = None
     #: Open text/reasoning entries being accumulated from deltas, by kind.
     open_entries: dict[str, int] = field(default_factory=dict)
     #: ``call_id`` -> entry sequence, so a tool result can find its call.
@@ -1014,13 +1020,40 @@ class TranscriptReader:
         state.seq += 1
         return state.seq
 
+    @staticmethod
+    def _elapsed_for(event: dict[str, Any], state: TranscriptState) -> float:
+        """Seconds from the trial's first event to this one.
+
+        Read from the line's own ``ts`` (epoch millis, stamped by Stella's sink
+        — see ``stella_protocol::journal``). Measuring from ``time.time()``
+        instead was the #2111 bug: a finished trial is read in a single pass, so
+        every event in the file was parsed within the same few milliseconds and
+        every row rendered ``0:00``. Anchoring to the *stream's* first stamp
+        makes an archived transcript report the offsets the run actually had —
+        and makes the paced replay in ``ui/components/arena/transcript-page.tsx``
+        pace on them instead of flushing at its 16 ms floor.
+
+        Two hedges the wire contract requires. A system clock is not monotonic,
+        so an NTP step can put a later line before the origin; that is clamped
+        to ``0.0`` rather than rendered as a negative offset. And a stream with
+        no ``ts`` at all — anything recorded before the field existed — keeps
+        the old read-time behaviour, which is wrong in the same way it always
+        was but is the only thing such a file supports.
+        """
+        stamp = event.get("ts")
+        if isinstance(stamp, (int, float)) and not isinstance(stamp, bool):
+            if state.origin_ms is None:
+                state.origin_ms = float(stamp)
+            return round(max(0.0, (float(stamp) - state.origin_ms) / 1000.0), 2)
+        if state.started is None:
+            state.started = time.time()
+        return round(time.time() - state.started, 2)
+
     def _entries_for(
         self, event: dict[str, Any], state: TranscriptState, path_key: str
     ) -> list[dict[str, Any]]:
         kind = str(event.get("type", ""))
-        if state.started is None:
-            state.started = time.time()
-        now = round(time.time() - state.started, 2)
+        now = self._elapsed_for(event, state)
 
         def entry(
             seq: int, etype: str, title: str, body: str = "", **meta: Any
--- a/arenabench/tests/test_telemetry.py
+++ b/arenabench/tests/test_telemetry.py
@@ -632,6 +632,83 @@ class TestFlip:
         assert leaders(totals, DIMENSIONS)["wasted_time"] == ["tight"]
 
 
+class TestTranscriptElapsed:
+    """The transcript's elapsed column, which read `0:00` on every row (#2111).
+
+    `t` was measured from `time.time()` at parse. A finished trial is read in
+    one pass, so every event in the file was parsed within the same few
+    milliseconds, every `t` rounded to `0.0`, and `fmtClock(0)` rendered `0:00`
+    down the whole gutter — indistinguishable, to anyone watching a live match,
+    from a broken feed. It also flattened the paced replay in
+    `ui/components/arena/transcript-page.tsx`, which paces on the gap between
+    consecutive `t` values and therefore had no gaps to pace on.
+
+    The fix is to read the `ts` Stella now stamps on each line and measure from
+    the stream's own first stamp.
+    """
+
+    def test_elapsed_comes_from_the_events_own_stamps(self, tmp_path: Path):
+        path = tmp_path / "e.jsonl"
+        write_events(path, [
+            {"ts": 1_754_582_400_000, "type": "stage", "name": "execute"},
+            {"ts": 1_754_582_412_500, "type": "stage", "name": "witness"},
+            {"ts": 1_754_582_490_000, "type": "stage", "name": "verify"},
+        ])
+        entries = TranscriptReader().read(path)
+        assert [e["t"] for e in entries] == [0.0, 12.5, 90.0]
+
+    def test_an_idle_gap_before_a_timeout_is_visible(self, tmp_path: Path):
+        """The forensic half of #2111: the wall clock between the last real
+        event and the timeout is the whole question, and it is unanswerable
+        unless both ends carry a stamp."""
+        path = tmp_path / "e.jsonl"
+        write_events(path, [
+            {"ts": 1_754_582_400_000, "type": "text", "text": "done"},
+            {"ts": 1_754_582_900_000, "type": "error", "message": "AgentTimeoutError"},
+        ])
+        entries = TranscriptReader().read(path)
+        assert entries[-1]["t"] - entries[0]["t"] == 500.0
+
+    def test_a_clock_step_backwards_never_renders_a_negative_offset(
+        self, tmp_path: Path
+    ):
+        """A system clock is not monotonic; the wire contract says so and asks
+        consumers to clamp. Rendering `-0:03` would look like corruption."""
+        path = tmp_path / "e.jsonl"
+        write_events(path, [
+            {"ts": 1_754_582_400_000, "type": "stage", "name": "execute"},
+            {"ts": 1_754_582_397_000, "type": "stage", "name": "verify"},
+        ])
+        entries = TranscriptReader().read(path)
+        assert [e["t"] for e in entries] == [0.0, 0.0]
+
+    def test_a_stream_recorded_before_the_stamp_existed_still_reads(
+        self, tmp_path: Path
+    ):
+        """Every `stella-events.jsonl` under `bench/evidence/` has no `ts`.
+        Those archives must keep rendering — flat offsets and all — rather than
+        failing on a missing key."""
+        path = tmp_path / "e.jsonl"
+        write_events(path, [
+            {"type": "stage", "name": "execute"},
+            {"type": "stage", "name": "verify"},
+        ])
+        entries = TranscriptReader().read(path)
+        assert [e["title"] for e in entries] == ["execute", "verify"]
+        assert all(isinstance(e["t"], float) for e in entries)
+
+    def test_incremental_reads_share_one_origin(self, tmp_path: Path):
+        """A live tail sees the file in several passes. The offsets must be
+        measured from the trial's first event, not from whatever line happened
+        to open the current batch."""
+        path = tmp_path / "e.jsonl"
+        reader = TranscriptReader()
+        write_events(path, [{"ts": 1_754_582_400_000, "type": "stage", "name": "a"}])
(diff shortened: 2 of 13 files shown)


macanderson:
Monitoring notes (no code changes made by me).

**Verified independently on the merged head**

- Witness flip reproduced against post-merge `main`: reverting only `arenabench/arenabench/telemetry.py` to `origin/main` and running `TestTranscriptElapsed` fails 3 of 5, including the literal `0:00` column — `assert [0.0, 0.0, 0.0] == [0.0, 12.5, 90.0]`, plus `test_an_idle_gap_before_a_timeout_is_visible` and `test_incremental_reads_share_one_origin`. The other two (`..._clock_step_backwards...`, `..._recorded_before_the_stamp_existed...`) pass on main — exactly the two this PR labels regression guards rather than witnesses. The description's distinction is accurate.
- `ts` is optional on every surface: all 39 root variants of `docs/wire/agentevent.schema.json` carry `ts` in `properties`, **0** carry it in `required`, and all 39 share one description string (sourced from `journal::TS_DESCRIPTION`). `agentevent.d.ts` has 39 × `ts?: number`. Generated, not hand-edited — `docs/wire still matches the types` is green.
- `crates/stella-protocol/tests/text_event_wire.rs` is **untouched** relative to `main` (it pins exact bytes, so the `None`-serializes-identically property is what keeps it green).
- `stella_pipeline::replay::to_jsonl` is unchanged and still a pure function of the run — the golden format is not stamped.
- Invariant 7 intact: no new clock read on the prompt-assembly path; the stamp is applied only at the two output sinks.

**Merge-collision check.** `main` moved twice during this watch. #2222 landed and touches the same two arenabench files. The merge was textually clean *and* semantically clean: `test_telemetry.py` goes 57 → 62 test defs (exactly the +5 this PR adds), `_elapsed_for`/`origin_ms` survived, and all 62 pass locally on the merged head. Worth noting the `check-deleted-tests` guard only sees Rust `#[test]`, so a Python-side clobber here would have been invisible to CI — it did not happen, but the guard would not have caught it.

**Residue filed.** The two-clock divergence flagged under "Anything reviewers should know?" now has a tracking issue — #2244. I deliberately did **not** resolve it: it is written up as a maintainer decision with the three options and the `bus.rs` god-file constraint spelled out, so the call stays yours. No ceiling raised, no god file touched.

Issue #2239, opened 2026-08-08
agent/tests.rs keeps a third hardcoded copy of the static-prompt pair, outside the derived registry


macanderson:
## Problem

`crates/stella-cli/src/agent/tests.rs` holds a third, unguarded copy of the static-prompt pair:

```rust
// crates/stella-cli/src/agent/tests.rs, in
// both_static_prompts_carry_a_read_symbol_steering_line
for (name, prompt) in [
    ("SYSTEM_PROMPT", SYSTEM_PROMPT),
    ("PIPELINE_SYSTEM_PROMPT", PIPELINE_SYSTEM_PROMPT),
] {
```

#2231 removed the second copy: `prompt.rs`'s `mod tests` now reads `STATIC_PROMPTS` out of `crates/stella-cli/src/agent/prompt/parity.rs`, where `the_registries_match_the_source` cross-checks it against the `const NAME: &str = concat!(` declarations in `prompt.rs`'s own source, in both directions. A third static prompt therefore cannot join the set without joining that list.

This inline copy is outside that guard. If a third static prompt is ever added, `both_static_prompts_carry_a_read_symbol_steering_line` silently keeps checking two of three — the #2231 shape, one seam over. It is small and low-probability (a third static persona is not on the roadmap), which is why it was filed rather than fixed.

## Where

- `crates/stella-cli/src/agent/tests.rs` — `both_static_prompts_carry_a_read_symbol_steering_line`, the inline array.
- `crates/stella-cli/src/agent/prompt/parity.rs` — `STATIC_PROMPTS`, currently `pub(super)` (visible inside `prompt` only).

## Suggested approach

Widen `mod parity` and `STATIC_PROMPTS` to `pub(crate)` (or `pub(in crate::agent)`) and replace the inline array with it. This **removes** lines from `agent/tests.rs`, which matters: that file is a god file and closed to growth, so a change there has to shrink it or leave it level.

The alternative — leaving the copy and adding a test that the two lists agree — is worse: it is one more list to keep current, which is the defect.

## Constraints

- `crates/stella-cli/src/agent/tests.rs` is a god file (`scripts/file-size-baseline.txt`); never raise its ceiling.
- `mod parity` is `#[cfg(test)]`; keep it that way — it exists only for the test tree.

## Verify / definition of done

- `cargo test -p stella-cli --bin stella prompt` green, and `both_static_prompts_carry_a_read_symbol_steering_line` still runs (check it by name in the output, not by the suite being green).
[...]

Issue #2237, opened 2026-08-08
Consolidate the per-contract prompt embedding assertions now that the derived check covers them


macanderson:
## Problem

`crates/stella-cli/src/agent/prompt.rs` pins each shared prompt contract with its own hand-written test that asserts both prompts embed it verbatim:

- `both_prompts_embed_the_one_shared_steering_literal` (`tool_steering!`)
- `both_prompts_scope_delivery_to_the_prompt_actually_sent` (`scope_discipline!`)
- `both_prompts_void_measurements_whose_producing_command_errored` (`measurement_discipline!`)
- `both_prompts_size_verification_to_what_the_turn_changed` (`verification_proportionality!`, arriving with #2232)

Since #2231 shipped, `crates/stella-cli/src/agent/prompt/parity.rs` makes the same embedding assertion for **every** contract, derived from `prompt.rs`'s own source rather than listed. The per-contract embedding assertions are now redundant with it.

The issue #2231 asked for them to be consolidated. They were deliberately **not** deleted in that PR, for two reasons that were true at the time and are worth re-checking before acting:

1. **#2232 was open against the same `mod tests` block.** Deleting three tests from a file another PR is adding a test to is the exact same-seam shape that produced #1976 and #1860 — both branches green, the merge textually clean, one side simply not containing the other's lines.
2. Each test's doc comment is the provenance record for its contract (the bench trace that produced it), and its **claim** assertions (`"unmeasured"`, `"A turn that changed nothing gets the read-only probe"`, the five clauses #2232 pins) are non-vacuity guards about the literal's *content*, which #2231 explicitly said to keep. Only the one-line embedding assertion inside each is redundant.

## Where

- `crates/stella-cli/src/agent/prompt.rs` — `mod tests`, the four `both_prompts_*` tests.
- `crates/stella-cli/src/agent/prompt/parity.rs` — `every_shared_contract_is_invoked_by_every_static_prompt` and `every_shared_contract_reaches_every_static_prompt_verbatim`, which already cover the embedding half for every contract.

## Suggested approach

[...]

Issue #2227, opened 2026-08-08
bench: the adapter-caller sweep reads heredoc Python textually — a stale keyword inside one is still invisible


macanderson:
Follow-up to #2203 (PR landing the class-level sweep in `bench/harbor_adapter/tests/test_manifest_parity.py`). Three blind spots are declared in that module’s docstring rather than closed; this issue tracks closing them or deciding not to.

## What is covered today

`test_manifest_parity.py::test_every_caller_in_the_tree_still_binds_to_the_live_adapter` sweeps the repository and checks, against the live `stella_harbor`:

| Seam | How | Count today |
|---|---|---|
| `.py` imports + call keywords/arity | `ast` + `inspect.signature` | 139 imports, 419 calls |
| `.sh` / `.bash` / `.md` / `.mdx` embedded imports | textual `from stella_harbor… import …` | 4 |
| `module:attribute` entry-point specs, any text file | regex + `getattr` | 51 |

## The three gaps

**1. Heredoc Python gets name resolution, not signature checking.**
`bench/evidence/run/witness_ab.sh` and `bench/evidence/frontier/setup_venv.sh` embed Python in a `<<'PY'` heredoc; `bench/RUNBOOK.md` and `bench/harbor_adapter/README.md` embed it in `python -c '…'`. The sweep resolves the *imported names* in those blocks — the #2182 shape — but cannot AST-parse the block, so a call inside one passing a renamed keyword (`_validated_verifier(spec=…)` after `spec` becomes `model`) is invisible. No embedded block passes keywords today, which is why this was not built: untested extraction machinery is worse than a declared gap.

*Possible approach*: from each textual import hit, expand downward line-by-line while `ast.parse` of the accumulated block succeeds, and run the existing `_check_call` over whatever parses. Conservative by construction — anything recovered is genuinely Python; anything not recovered is no worse than today. Needs a fixture proving it catches a stale keyword in a heredoc, and an anti-vacuity assertion that it recovered a non-zero number of blocks (otherwise it silently degrades to today’s behaviour).

**2. `**splat` callers and `**kwargs` callees are skipped.**
[...]

Issue #2203, opened 2026-08-08
bench: adapter-caller parity is pinned per-seam, not per-class — the next rename strands a third caller silently


macanderson:
## Problem

#2182 was two callers of `bench/harbor_adapter` stranded by one rename (#1394, `witness_author` → `verifier`), each of which killed benchmark evidence generation on the run host:

- `bench/evidence/make_manifest.py` passed `witness_author=` to `_benchmark_engine_posture` / `_benchmark_assurance_tiers` → uncaught `TypeError` → **no run manifest written at all**.
- `bench/evidence/run/witness_ab.sh` imported `_validated_witness_author`, now `_validated_verifier` → `ImportError` behind `|| exit 1` → **the witness-on arm refused to launch**.

PR #2202 fixed both and added `bench/harbor_adapter/tests/test_manifest_parity.py`, which pins the seams **by name**: a hardcoded `_CALLEES` dict of two functions, and one directory (`bench/evidence/run/`) scanned for `from stella_harbor… import …`.

**That is a per-seam ratchet, not a class-level one.** The third caller that a future rename strands will not be covered, and will fail the same way: silently in every test suite, loudly and only on the run host, mid-paid-run.

## Why this is tractable

While sweeping for #2182 I ran a throwaway check that covered the whole class in about 30 lines and under a second. It is deliberately not in PR #2202 — it is a new gate, not part of that fix — but it demonstrated the approach works and that the tree is currently clean:

- Walk every `*.py` outside `bench/harbor_adapter/stella_harbor/`, collect `ImportFrom` nodes whose module starts with `stella_harbor`, and assert each imported name resolves via `hasattr` on the live module. **All resolve today.**
- For each `ast.Call` whose target is one of those imported names, compare `keyword.arg` names against `inspect.signature(...)`.parameters of the live callee (skipping callees with `**kwargs`). **345 external call sites checked, no stale kwargs today.**
- Repeat the textual `from stella_harbor… import …` regex over `*.sh` and `*.md` for heredoc-embedded Python — this is the half that catches the `witness_ab.sh` shape, which no AST walk can see.

## Definition of done

A test in `bench/harbor_adapter/tests/` that performs the sweep above across the repository, replacing the hand-named `_CALLEES` / `_EVIDENCE_RUN` literals in `test_manifest_parity.py` (or subsuming them, keeping the arm-selector and digest-disclosure assertions, which are semantic and cannot be derived generically).

Requirements, learned the hard way in #2182:

[...]

PR #2241, opened 2026-08-08
test(stella-cli): derive the shared prompt contracts and check both prompts embed them

### LINKED ISSUE #2231: Shared prompt contracts are coupled to both static prompts by convention, not by a check

macanderson:
## What

A check that derives the set of shared prompt contracts from `crates/stella-cli/src/agent/prompt.rs`'s own source and asserts each one is embedded in **both** static prompts — so the coupling is enforced rather than remembered.

New file: `crates/stella-cli/src/agent/prompt/parity.rs` (444 lines, `#[cfg(test)]`). `prompt.rs` grows by 14 lines net (+20/-6): the `mod parity;` declaration, an ADDING A CONTRACT note in the paragraph that already explains the shared-literal pattern, and `mod tests` now reading its `PROMPTS` list out of `parity::STATIC_PROMPTS` instead of keeping a second copy.

Closes #2231

## The gap, confirmed on main first

`prompt.rs` declares its cross-cutting contracts as `macro_rules!` literals so both prompts embed the same bytes, and each was pinned by its own hand-written test. That covers the contracts that exist and says nothing about the next one. Measured before writing anything, on `origin/main` at 6b9c135a — a fifth contract embedded in `SYSTEM_PROMPT` only, no test written for it:

```
test result: ok. 45 passed; 0 failed; 0 ignored; 0 measured; 1433 filtered out
```

Nothing fails. `PIPELINE_SYSTEM_PROMPT` is the half that goes missing quietly: it is what `stella run` sends and what the bench harness exercises, so a contract landing in `SYSTEM_PROMPT` alone is invisible to every measurement the project makes decisions from.

## How the set is derived, and why that is the point

The issue offered two shapes and preferred a `SHARED_CONTRACTS` slice; this ships the slice **and** the structural guard it called option 2, because a hardcoded roster reproduces the defect one level up — the fifth contract joins the set without joining the list, which is exactly how per-seam pinning failed in #2182/#2203.

`scan()` reads `prompt.rs`'s text (`include_str!`, compile-time) for two things:

- **contracts** — every top-level `macro_rules!` declaration.
- **static prompts** — every top-level `const NAME: &str = concat!(` declaration, and the body of that `concat!`.

Neither list is written down anywhere. The two runtime registries (`SHARED_CONTRACTS`, `STATIC_PROMPTS`) exist only because a macro cannot be invoked by a name computed at runtime, and `the_registries_match_the_source` fails **in both directions** when either falls behind the scan. Proof that they are not hardcoded in the way that matters: adding a `throwaway_contract!` macro to `prompt.rs` and nothing else fails immediately with

```
prompt.rs declares the contract `throwaway_contract`, which prompt/parity.rs
SHARED_CONTRACTS does not list — add its row. The list is what lets the check
compare bytes, and a contract missing from it is checked by nothing (#2231)
```

Six tests: the guard-the-guard scan check, the registry cross-check, the source-level coupling (`name!()` in every `concat!` body), the byte-level coupling (the expanded literal in the assembled `&'static str`), the anti-vacuity detectors, and a synthetic fixture holding the defect. **Three contracts** are validated today against **two prompts** — six pairs. #2232's `verification_proportionality!` is picked up automatically as a *failure* until its row is added; see "Merge interaction" below.

Exemplar: `bench/harbor_adapter/tests/test_manifest_parity.py` (#2202/#2228), which retired this failure class for bench adapter callers. This closes the prompt half — the fifth instance of "two sides of a boundary that nothing checks", after #2134, #2171, #2182, #2196.

## Witness

Three flips, all run the artisanal way against a committed tree.

**1. A real omission from one prompt only** — `measurement_discipline!()` stripped from `PIPELINE_SYSTEM_PROMPT`, left in `SYSTEM_PROMPT`:

```
test agent::prompt::parity::every_shared_contract_is_invoked_by_every_static_prompt ... FAILED
  a shared prompt contract is embedded by only some of the static prompts, so the two
  prompts have already diverged (#450, #2231).
  Missing: `measurement_discipline!()` is not in PIPELINE_SYSTEM_PROMPT

test agent::prompt::parity::every_shared_contract_reaches_every_static_prompt_verbatim ... FAILED
  PIPELINE_SYSTEM_PROMPT does not embed the shared `measurement_discipline` block verbatim

test result: FAILED. 4 passed; 2 failed
```

Both the contract and the prompt are named, at both the source and the byte level.

**2. The exact #2231 shape** — a *new* contract embedded in `SYSTEM_PROMPT` only, with no hand-written test of its own. This is the state that was green on main:

```
test agent::prompt::parity::the_registries_match_the_source ... FAILED
  prompt.rs declares the contract `throwaway_contract`, which prompt/parity.rs
  SHARED_CONTRACTS does not list ...
test agent::prompt::parity::every_shared_contract_is_invoked_by_every_static_prompt ... FAILED
  Missing: `throwaway_contract!()` is not in PIPELINE_SYSTEM_PROMPT

test result: FAILED. 49 passed; 2 failed
```

The decisive detail: **every pre-existing test passed**, including all three `both_prompts_*` tests, the #639 catalogue guard, and `both_static_prompts_carry_a_read_symbol_steering_line`. Only the two derived checks fire. That is the hole, closed.

**3. Anti-vacuity** — the parser deliberately made stale (`strip_prefix("macro_rules!")` changed to `"macro_rules !"`), simulating a scan that has stopped matching how contracts are declared:

```
test result: FAILED. 1 passed; 5 failed

the shared-contract scan could not read crates/stella-cli/src/agent/prompt.rs
(NoContractsDeclared). It examines nothing in this state, which reads as a passing
gate — correct the parser in agent/prompt/parity.rs rather than leaving it silent (#2231)
```

A scan that finds nothing is an error with a name, never an empty set quietly iterated. `the_scan_fails_loudly_rather_than_finding_nothing` drives each way in as well — empty source, no `macro_rules!`, no static prompt, and a prompt rewritten as a runtime `format!` — on hand-built sources, so the detectors are witnessed on every run rather than assumed. Note the one test that stayed green in flip 3 (`...reaches_every_static_prompt_verbatim`): it reads the runtime registry, not the scan, so the byte-level assertion is independent of the parser by construction.

Restored and green after each flip: `cargo test -p stella-cli --bin stella prompt` gives **51 passed, 0 failed** (45 before + 6 new).

## Byte stability is unaffected — invariant 7 / L-E8

No prompt literal becomes runtime-computed. The contracts stay `macro_rules!` returning string literals, both prompts stay `concat!` compositions, every value in this module is a `&'static str` fixed at compile time, and the module only *reads* them — `git diff` touches no byte inside any prompt constant. The prompt-cache prefix is exactly what it was.

The check in fact **enforces** one half of that invariant: a prompt is recognised only by `const NAME: &str = concat!(`, so rewriting one as a runtime composition stops it parsing as a static prompt, and the registry cross-check turns that into a named failure rather than a prompt the scan quietly stops covering. Pinned by `the_scan_fails_loudly_rather_than_finding_nothing`'s fourth case.

## Merge interaction with #2232 — please read before merging either

#2232 adds a fourth contract (`verification_proportionality!`) to `prompt.rs`. These two branches merge textually clean and produce a **red** `main`: the new contract has no `SHARED_CONTRACTS` row, so `the_registries_match_the_source` fails. Whichever lands second needs one line:

```rust
("verification_proportionality", verification_proportionality!()),
```

The failure message names the file and the list, so it is a one-line fix with a handoff attached rather than a puzzle. This is why the ADDING A CONTRACT note went into `prompt.rs` beside the macros — the next author works there, not in the check. Flagged on #2232 as well.

## God files

None touched. `prompt.rs` is 712 lines (was 698, ceiling 1500, not grandfathered); `prompt/parity.rs` is 444. `agent/tests.rs` is untouched. `scripts/check-file-size.sh` reports "none grew". No baseline entry added, no ceiling raised.

## Checks

`cargo test -p stella-cli --bin stella prompt` gives 51 passed. `cargo clippy -p stella-cli --all-targets -- -D warnings` clean. `cargo fmt --check -p stella-cli` clean. `check-file-size`, `check-left-behind`, `check-brand-case`, `check-god-files`, `check-module-reachability` all pass. Not run locally: `--workspace` tests and the rest of `make gate` — CI is the gate, and the diff reaches one crate.

## Tests deleted

**None.** Issue #2231's definition of done asked for the per-contract `both_prompts_*` tests to be consolidated; that is deliberately deferred, and the reasoning is recorded in `parity.rs`'s module doc rather than left in a session:

1. #2232 is open against the same `mod tests` block. Deleting three tests from a file another PR is adding a test to is the same-seam shape that produced #1976 and #1860 — both branches green, the merge clean, one side simply not containing the other's lines.
2. The structural guard (the issue's option 2) closes the hole *entirely*, which makes consolidation an orthogonal cleanup rather than the mechanism. Each test's doc comment is the provenance record for its contract, and its claim assertions guard the literal's **content** — a different property, which the issue explicitly said to keep.

Filed as #2237 so it is tracked, not dropped.

## Nothing left behind

- **#2237** — consolidate the per-contract embedding assertions once #2232 lands; also gives `scope_discipline!` and `tool_steering!` the content guard they lack (a trim that hollows out `scope_discipline!` while leaving it embedded passes everything in the tree today).
- **#2239** — `agent/tests.rs` keeps a third hardcoded copy of the static-prompt pair, outside the derived registry. Fixing it removes lines from a god file, so it is safe, but it is a separate seam.

No TODOs added. Related: #2228 (the class-level sweep this mirrors), #2227, #2203, #450, #639, #1955, #1957, #1958.

### DIFF
--- a/crates/stella-cli/src/agent/prompt.rs
+++ b/crates/stella-cli/src/agent/prompt.rs
@@ -28,6 +28,13 @@ use super::*;
 // byte-stable-prefix property (L-E8) a runtime `format!` would give up. One
 // shared literal, embedded verbatim by both prompts, is also what keeps the two
 // copies from drifting the way the catalogue's did (#450).
+//
+// ADDING A CONTRACT: embed it in BOTH prompts below, and add its row to
+// `SHARED_CONTRACTS` in `prompt/parity.rs`. That module derives the set of
+// contracts from this file's own source, so it fails by name on either
+// omission — the coupling used to hold by convention alone, and a contract
+// reaching `SYSTEM_PROMPT` only is invisible to `stella run` and to every
+// bench measurement, which read `PIPELINE_SYSTEM_PROMPT` (#2231).
 
 /// The cross-tool steering shared by both static prompts — what the generated
 /// schemas cannot say. No trailing newline: each prompt continues with its own
--- a/crates/stella-cli/src/agent/prompt.rs
+++ b/crates/stella-cli/src/agent/prompt.rs
@@ -451,15 +451,22 @@ pub(crate) fn render_file_tree(files: &str, max_lines: usize) -> String {
     out
 }
 
+/// The structural half of the shared-contract discipline: the tests above are
+/// written one per contract, so they cover the contracts that exist and say
+/// nothing about the next one. This module derives the set from this file's own
+/// source and asserts the coupling over it (#2231). Declared last because
+/// `macro_rules!` scope is textual — the contracts above have to be in scope.
+#[cfg(test)]
+mod parity;
+
 #[cfg(test)]
 mod tests {
-    use super::{PIPELINE_SYSTEM_PROMPT, SYSTEM_PROMPT, append_workspace_memories};
+    use super::append_workspace_memories;
 
-    /// Every static prompt, labelled.
-    const PROMPTS: &[(&str, &str)] = &[
-        ("SYSTEM_PROMPT", SYSTEM_PROMPT),
-        ("PIPELINE_SYSTEM_PROMPT", PIPELINE_SYSTEM_PROMPT),
-    ];
+    // Every static prompt, labelled. Lives in `parity`, where it is
+    // cross-checked against this file's own source, so a third prompt cannot
+    // join the set without joining the list every test here iterates (#2231).
+    use super::parity::STATIC_PROMPTS as PROMPTS;
 
     /// The catalogue-shaped lines of `prompt`: `- <tool_name>: …`, where the
     /// lead-in is a bare canonical tool name or a `/`-joined run of them.
--- a/crates/stella-cli/src/agent/prompt/parity.rs
+++ b/crates/stella-cli/src/agent/prompt/parity.rs
@@ -92,6 +92,10 @@ const SHARED_CONTRACTS: &[(&str, &str)] = &[
     ("tool_steering", tool_steering!()),
     ("scope_discipline", scope_discipline!()),
     ("measurement_discipline", measurement_discipline!()),
+    (
+        "verification_proportionality",
+        verification_proportionality!(),
+    ),
 ];
 
 /// Every static prompt, paired with its assembled bytes. Same derivation
--- a/crates/stella-cli/src/agent/prompt/parity.rs
+++ b/crates/stella-cli/src/agent/prompt/parity.rs
@@ -1,0 +1,434 @@
+//! Structural parity for the shared prompt contracts (#2231).
+//!
+//! `prompt.rs` declares its cross-cutting contracts — tool steering, scope,
+//! measurement — as `macro_rules!` string literals so that **both** static
+//! prompts can embed the same bytes: one literal, two `concat!` sites, no
+//! second copy to drift from. The catalogue this pattern replaced had drifted
+//! in exactly that way (`verify_done` and `ask_user` took comma splices in the
+//! pipeline copy where the base copy took an em dash, #450).
+//!
+//! Each contract was then pinned by its own hand-written test. That covers the
+//! contracts that exist and says nothing about the next one: a fifth contract
+//! embedded in `SYSTEM_PROMPT` alone, with no test written for it, failed
+//! nothing. Confirmed on `main` before this module existed — a throwaway
+//! contract added to `SYSTEM_PROMPT` only left `cargo test -p stella-cli --bin
+//! stella prompt` at "45 passed; 0 failed". The coupling held by convention.
+//!
+//! `PIPELINE_SYSTEM_PROMPT` is the half that would go missing quietly: it is
+//! what `stella run` sends and what the bench harness exercises, so a contract
+//! landing in `SYSTEM_PROMPT` alone is invisible to every measurement the
+//! project makes decisions from.
+//!
+//! # The set is derived, never listed
+//!
+//! A hardcoded roster of contracts would reproduce the defect one level up —
+//! the fifth contract simply joins the set without joining the list. So the
+//! authority here is `prompt.rs`'s own source text: [`scan`] reads the
+//! `macro_rules!` declarations and the `const … = concat!(` prompt
+//! declarations out of it, and the two runtime registries below exist only
+//! because a macro cannot be invoked by a name computed at runtime. Both are
+//! cross-checked against the scan in both directions, so a registry that falls
+//! behind the source is a named failure rather than a check that quietly
+//! covers less.
+//!
+//! # Byte stability is untouched
+//!
+//! Every value here is a `&'static str` fixed at compile time: the contracts
+//! stay `macro_rules!` literals, the prompts stay `concat!` compositions, and
+//! this module only reads them. Nothing becomes runtime-computed, so the
+//! byte-stable prefix the prompt cache keys on (invariant 7 / L-E8) is exactly
+//! what it was. The scan in fact *enforces* one half of that: a prompt
+//! rewritten as a runtime `format!` stops parsing as a static prompt and the
+//! registry cross-check fails naming it.
+//!
+//! # Anti-vacuity
+//!
+//! A check that silently examines nothing reads as a passing gate, which is
+//! strictly worse than no check — the same shape as #2182, where a parity
+//! check could not run where it mattered. So every way this scan could come up
+//! empty is an error with a name: [`ScanError::NoContractsDeclared`],
+//! [`ScanError::NoStaticPromptsParsed`], a registry that no longer matches the
+//! source, and a contract literal too short for `contains` to prove anything.
+//! `the_scan_fails_loudly_rather_than_finding_nothing` drives each one.
+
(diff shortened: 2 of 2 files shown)


macanderson:
Cross-link from #2232 (adds a fourth shared contract, `verification_proportionality!`, closing #1958).

#2232 is green and cannot carry the registry row — `crates/stella-cli/src/agent/prompt/parity.rs` exists only on this branch — so the agreed ordering is **#2232 first, then this PR**. When you update this branch afterwards, `the_registries_match_the_source` will fail by name:

```
prompt.rs declares the contract `verification_proportionality`, which
prompt/parity.rs SHARED_CONTRACTS does not list. Add its row ...
```

That is the check doing exactly what #2231 asked for. The fix is one row in `SHARED_CONTRACTS`:

```rust
("verification_proportionality", verification_proportionality!()),
```

Nothing else changes — #2232 already embeds the literal in both static prompts, so `every_shared_contract_is_invoked_by_every_static_prompt` passes for it as-is.

Issue #1702, opened 2026-08-06
The plan-step loop walks every step even when step 1 finished the task — 63% of a run's cost re-confirming finished work


macanderson:
## Problem

`Pipeline::execute` drives plan steps with an unconditional `for` loop
(`crates/stella-pipeline/src/pipeline.rs:2154-2186`): each step pushes
`"Step {i}/{n}: {description}"` and runs a full engine turn. Nothing checks
whether the goal is already satisfied, and the worker is never told which steps
are already done. A worker that completes the whole task in step 1 — the normal
outcome when the plan's steps are one shell script's worth of work — is then
walked through the remaining `n-1` steps, re-verifying its own finished work at
two model calls per step.

Measured on a Terminal-Bench-shaped nginx setup run (8-step plan):

| | calls | cost |
|---|---|---|
| Plan + steps 1–4 (all real work: install, configure, test, restart, verify) | 5 | $0.379 |
| Steps 2–8 (re-reading and re-confirming finished work) | 14 | $0.647 |
| **Total** | **19** | **$1.026** |

**63% of the run's cost produced no work.** Every one of the 7 redundant steps
cost exactly 2 calls — one to act, one to emit "Step N done" — because the
worker ends each turn with a text-only message after its tool result.

The transcript grows monotonically across steps (input tokens 12,771 → 21,120),
so the redundant tail is also the most expensive part per call, and it inflates
the prompt the *next* step must carry.

## Secondary harm: the plan's illustrative text is read as a spec

Step 2 said `a custom log_format (e.g. 'benchmark' ...)`. The worker had already
written a working config using `benchmark_format` / `benchmark_limit`, and on
reaching step 2 rewrote both identifiers with `sed -i` to match the plan's `e.g.`
example, then restarted nginx — churn against already-correct, already-verified
config, driven by an example the planner offered as illustration. Steps 3 and 4
were pure no-ops (`cat` the file it had just written; `rm -f` an already-removed
path), each still costing 2 calls.

## Definition of done

A plan step that is already satisfied costs approximately nothing, and the
worker can tell that it is.

Options, roughly in increasing order of ambition — a maintainer call on which:

1. **Tell the worker the truth.** Include the steps already completed (and the
   fact that earlier steps may have covered later ones) in the step prompt, and
   say explicitly that confirming finished work is a no-op it should report in
   one line without tool calls. Cheapest; removes the 2nd call per step, not the
   1st.
[...]

PR #2264, opened 2026-08-08
fix(stella-pipeline): a negated PLAN COMPLETE echo no longer closes the plan (#2104)

### LINKED ISSUE #2104: pipeline: a negated echo of the PLAN COMPLETE sentinel closes the plan — skipped 9/10 steps on git-multibranch, reward 0.0
## Problem

`goal_declared_complete` (`crates/stella-pipeline/src/pipeline/plan_steps.rs:61`) closes the plan-step loop when **any** line's first non-whitespace characters are `PLAN COMPLETE`. A worker that *negates* the sentinel still trips it. Observed on Terminal-Bench 2.1, task `git-multibranch`, match `61e01ca06fc7` (arm `stella`, worker `openrouter/anthropic/claude-sonnet-5`, 2026-08-07 ~00:51):

The planner produced a 10-step plan. On **step 1/10** ("Install required packages") the worker confirmed the packages were preinstalled and replied:

```
All required packages (openssh-server, git, nginx, openssl) are already installed.

PLAN COMPLETE: no — this is just step 1; only confirming this step is done.

Step 1/10 is already complete: ...
```

The middle line starts with the marker → the plan closed → **steps 2–10 (the entire actual task: sshd, bare repo, post-receive hook, TLS cert, nginx) were skipped**. Total spend $0.14; trial reward **0.0**.

## Why the documented backstop does not hold

The module doc says a spurious match is safe because "the verify stage still runs on whatever the tree actually holds." That is only true for work that lands **in the workspace tree**. This task mutates `/etc`, `/git`, and system services; the diff probe read the tree, found it unchanged, and verification returned `UNVERIFIABLE` — after which the verdict was `passed=true`. So on environment-mutating tasks (most of Terminal-Bench) the false close-out is caught by nothing.

Contributing factor: `step_prompt` itself teaches the model the sentinel ("reply with a single line beginning `PLAN COMPLETE:`"), so echoes-with-commentary are an expected model behavior, not a fluke.

## Evidence

- Trace: `~/.arenabench/matches/61e01ca06fc7/jobs/61e01ca06fc7-stella/git-multibranch__JXhfmDp/agent/stella-events.jsonl` (stages go `plan > execute` [step 1 only] `> verify > complete`; verdict `rung=unverifiable, passed=true`; `result.json` reward 0.0)
- Parser + tests: `crates/stella-pipeline/src/pipeline/plan_steps.rs` (`goal_declared_complete`, `a_marker_line_closes_the_plan_wherever_it_sits_in_the_reply`)
- Loop: `crates/stella-pipeline/src/pipeline.rs` (`execute_plan` step walk, #1702 close-out)

## Fix directions (pick one, with a witness test)

1. **Enforce the prompt's own contract**: the prompt says the *reply* begins with the marker — accept the marker only on the first non-empty line of the reply. The existing "short report then declaration" allowance (test `a_marker_line_closes_the_plan_wherever_it_sits_in_the_reply`) is what admits the negated echo; if that allowance stays, it needs at least option 2.
2. **Negation guard**: reject a marker line whose payload after `PLAN COMPLETE:` opens with a negation (`no`, `not`, `n't`, …). Cruder, keeps mid-reply declarations.
3. Either way, when the close-out fires and verification then lands on the `unverifiable` rung, the skip should be treated as unproven — e.g. resume the walk or refuse `passed=true` — because the #1702 rationale ("verify refutes the lie") explicitly does not apply there.

## Definition of done

- A witness test in `plan_steps.rs` that feeds exactly the observed reply (`"...\n\nPLAN COMPLETE: no — this is just step 1...\n"`) to `goal_declared_complete` (or the chosen new seam) and asserts the plan does NOT close. It must fail on current `main`.
- Existing #1702 behavior preserved: a genuine `PLAN COMPLETE: <affirmative>` still ends the walk (tests in `plan_walk.rs` stay green).

Related: #1702 (the close-out feature), #1770 (measuring its adoption/cost), #2089 (witness acceptance measurement — this trial is also an example of a task whose reward was lost to the false close-out rather than to witness machinery).

macanderson:
## The bug

On Terminal-Bench 2.1 `git-multibranch` (match `61e01ca06fc7`, worker `claude-sonnet-5`) the planner produced a 10-step plan. Answering **step 1 of 10** — "Install required packages" — the worker confirmed the packages were preinstalled and replied:

```
All required packages (openssh-server, git, nginx, openssl) are already installed.

PLAN COMPLETE: no — this is just step 1; only confirming this step is done.

Step 1/10 is already complete: ...
```

`goal_declared_complete` matched **any** line whose first non-whitespace characters were the marker. The middle line matched. The plan closed, and steps 2–10 — sshd, the bare repo, the post-receive hook, the TLS cert, nginx, i.e. the entire actual task — were never attempted. Trial reward **0.0** for $0.14.

The worker was not lying. It was answering the prompt honestly and using the sentinel's own vocabulary to say *no*.

## Why nothing caught it

`plan_steps.rs`'s module doc argued the close-out was safe because "the verify stage still runs on whatever the tree actually holds", so a false claim is refuted like any other lie.

That holds only for work that lands **in the workspace tree**. This task mutates `/etc`, `/git` and system services. The diff probe read the tree, found it unchanged, and verification returned `UNVERIFIABLE` — after which the verdict was `passed=true`. Most of Terminal-Bench is that shape, so on the corpus that matters the documented backstop does not exist. The module doc asserted the opposite and is corrected here.

Contributing factor: `step_prompt` itself teaches the model the sentinel, so echoes-with-commentary are expected model behaviour, not a fluke.

## The fix

A close-out must now be a *declaration*, not a mention of one, on two independent axes:

**Position** — the marker line must open or close the reply. This preserves #1702's "short report, then declare" allowance exactly (the declaration is then the last significant line) while rejecting a marker buried mid-reply with argument still running after it.

**Polarity** — the payload must not open by denying itself. Matched on the first **whole word**, case-folded and stripped of punctuation, never as a prefix — so `PLAN COMPLETE: nothing left to do` still closes the plan and `PLAN COMPLETE: not yet` does not. Contractions ending `n't` count.

Either axis alone would have caught the observed reply. Both are here because they fail differently: position catches a denial buried in commentary, polarity catches one that lands last.

## Witness

Three new tests in `plan_steps.rs`, all of which **fail against the previous predicate** and pass now (verified by restoring the old body and re-running):

- `a_negated_echo_of_the_marker_does_not_close_the_plan` — the observed reply, verbatim
- `a_denial_is_rejected_even_when_it_is_the_last_line`
- `the_declaration_must_open_or_close_the_reply`

Plus `an_affirmative_payload_that_looks_like_a_negation_still_closes`, which pins the whole-word rule so `nothing`/`noted` are not read as `not`/`no`.

The four pre-existing tests pass unchanged in **both** directions, so #1702's behaviour is preserved rather than merely believed to be.

```
cargo test -p stella-pipeline --lib     642 passed
cargo clippy -p stella-pipeline --all-targets -- -D warnings    clean
```

No test was deleted.

### Review follow-up (49cb0daf)

The negation screen took the payload's first whitespace-delimited token and then *deleted* every non-alphanumeric character from it, rather than stopping at one. So `PLAN COMPLETE: no—steps 2-10 remain` yielded the word `nosteps`, matched no opener, and closed the plan — this bug's exact failure, one keystroke away from the transcript above. Unspaced em dashes are ordinary model punctuation; the observed reply used a spaced dash only by luck. The screen now scans to the first word and lets punctuation terminate it, and admits `n/a` explicitly since it terminates at the slash on a bare `n`. `nothing left to do` and `noted, all eight done` still close the plan.

**Renamed test:** `a_marker_line_closes_the_plan_wherever_it_sits_in_the_reply` is renamed to `a_marker_line_closes_the_plan_when_it_opens_or_closes_the_reply`. Position-dependence is the whole point of this change, so "wherever it sits" stated the contract being removed. Its assertions are unchanged and still pass — both cases sit at reply edges. Nothing else was deleted.

New test `a_denial_fused_to_its_punctuation_is_still_a_denial` was verified as a true witness: it fails against the pre-49cb0daf predicate (`should not close the plan: PLAN COMPLETE: no—steps 2-10 remain.`) and passes with the fix.

Closes #2104

## Summary by Sourcery

Tighten PLAN COMPLETE handling in the stella pipeline so only affirmative, positional declarations can close a multi-step plan, preventing negated echoes of the marker from prematurely ending remaining steps.

Enhancements:
- Update the goal completion predicate to require PLAN COMPLETE lines to appear at the start or end of a reply and to reject payloads that open with negation words or contractions.
- Document that close-out screening happens in the planner itself because downstream verification may be UNVERIFIABLE for tasks that do not touch the workspace tree.

Tests:
- Add targeted tests around PLAN COMPLETE parsing to cover negated echoes, last-line denials, positional requirements, and non-negation lookalikes while keeping existing behaviour from #1702 intact.

### Review follow-up (748a119a)

A second review found two more holes of the **same shape** as `49cb0daf`'s — the screen only
agreed with the tokenizer that produced its input for ASCII input. Both are verified witnesses:
they fail against the pre-`748a119a` predicate and pass now.

**Typographic apostrophe.** The word scan admitted only ASCII `U+0027`, so a contraction written
with `U+2019` — which models emit by default, and which the em dash in the motivating transcript
above shows this model already does — terminated the word early and yielded the harmless `isn`:

```
PLAN COMPLETE: isn’t done, one step left.   →  closed the plan
```

The apostrophes a contraction can carry are now one named constant (`APOSTROPHES`: `U+0027`,
`U+2019`, `U+02BC`), folded to `'` before the `n't` test, so the test itself stays one comparison.
New test `a_denial_written_with_a_typographic_apostrophe_is_still_a_denial`.

**The marker was a prefix, not a token.** `strip_prefix` accepted any word starting with the
marker, and the leftover letters became the payload — so the polarity screen read the wrong first
word and the whole axis was void wherever this fired:

```
PLAN COMPLETED: no, steps 2-10 remain.   →  closed the plan  (payload `D: no, …`, first word `d`)
PLAN COMPLETENESS: 1 of 10 steps.        →  closed the plan
```

A non-alphanumeric (or end of line) is now required after the marker. Every punctuation that
follows it in practice still declares (`:`, ` — `, bare `PLAN COMPLETE`); the cost is at most the
redundant walk this module exists to avoid. New test
`a_word_fused_to_the_marker_is_not_a_declaration`.

**Pinned, not changed:** `an_ambiguous_no_is_read_as_a_denial` asserts that `PLAN COMPLETE: no
further work needed.` does **not** close the plan, even though an affirmative reading exists. That
is the deliberate side of the asymmetry, and it was previously unpinned — a future maintainer could
have "fixed" it into a hole. `none remaining` still closes, since `none` is not an opener.

**Stale comment corrected.** `execute_stage.rs`'s caller still asserted that a false declaration
"is the verify stage's to refute, not this loop's" — the exact claim this PR's module doc records
as false for any task whose subject lies outside the workspace tree. Chased to the one other place
it appeared.

Verification: all 24 pre-existing assertions and 10 new ones pass; against the previous predicate
exactly the 6 new bug-cases fail and nothing else changes. **No test was deleted or renamed in
this commit.**

**Filed, not fixed:** #2332 — the two residual failure modes this screen cannot close, in opposite
directions (position rejects a declaration followed by a courtesy line; first-word polarity accepts
`PLAN COMPLETE: yes for step 1 only — steps 2-10 remain.`). Both want a structured close-out
signal rather than a sentinel in prose, which is out of scope here.


macanderson:
## Review: the screen was one keystroke from re-opening #2104 — fixed on this branch

The position-plus-negation rule is correct and the witness carries the production transcript verbatim, which is the right way to pin a bug like this. Two things needed fixing first; I've pushed both to this branch rather than leaving them as a round-trip.

**The fused-negation miss the Vercel bot flagged at `plan_steps.rs:130` is real, and it fails in the catastrophic direction.** `opens_with_negation` took the payload's first whitespace-delimited token and then *deleted* every non-alphanumeric character from it, rather than stopping at one. So:

```
PLAN COMPLETE: no—steps 2-10 remain.       → word "nosteps"      → closes the plan
PLAN COMPLETE: no/incomplete, steps remain. → word "noincomplete" → closes the plan
PLAN COMPLETE: no,steps 2-10 remain.        → word "nosteps"      → closes the plan
PLAN COMPLETE — no, steps remain.           → word ""            → closes the plan
PLAN COMPLETE: n/a — still on step 1.       → word "na"          → closes the plan
```

Every one of those is a denial that closes the plan — #2104's exact failure, on tasks where verify returns `UNVERIFIABLE` and cannot refute it. Unspaced em dashes are ordinary model punctuation; the observed transcript used a spaced dash only by luck.

Fixed in 49cb0daf by scanning to the first word and letting punctuation terminate it, with `n/a` admitted explicitly since it terminates at the slash on a bare `n`. `nothing left to do` and `noted, all eight done` still close the plan — the whole-word semantics are preserved. The new test `a_denial_fused_to_its_punctuation_is_still_a_denial` is a true witness: I verified it fails against this PR's original predicate (`should not close the plan: PLAN COMPLETE: no—steps 2-10 remain.`) and passes with the fix. `NEGATION_OPENERS`' doc comment said "after punctuation is stripped", which is now the opposite of what the code does, so it moved too.

Also renamed `a_marker_line_closes_the_plan_wherever_it_sits_in_the_reply` — position-dependence is the entire point of this PR, and "wherever it sits" states the contract it removes. Its assertions still pass unchanged; both cases sit at reply edges.

One judgment call left as-is, worth knowing about: `declares_completion(first) || declares_completion(last)` means an affirmation at *either* end beats a denial at the other, so a reply opening "PLAN COMPLETE: all done" and closing "PLAN COMPLETE: not yet" closes the plan. Given this PR's own asymmetry argument (a spurious close costs a whole trial), last-marker-wins is arguably more correct. Degenerate enough that I did not change behavior unilaterally.

For the record on the miss direction: a genuinely-complete declaration buried mid-reply is now missed, but the consumer is a bounded `for` loop (`execute_stage.rs:109`), so the cost is the redundant walk this module exists to avoid — never a hang. And `step_prompt` already asks for the declaration as a single line, so a buried one is off-contract.

Issue #2098, opened 2026-08-07
arenabench: a pinned sut_ref="main" races its own build — launch-time re-resolution can silently upload the host's native dev binary


macanderson:
## Problem

`SutBuilder.start(ref)` (`arenabench/arenabench/sut_build.py`) and
`_agent_environment(...)` (`arenabench/arenabench/runner.py:861-903`) each call
`sut.resolve_ref("main")` **independently**, at two different times: once when
a build is requested, and again — a fresh `git fetch origin main` +
`rev-parse` — when the match actually launches. On an actively-developed repo
where commits land every few minutes, `main` can (and did) move between the
two calls. There is no coordination between them at all.

When it moves, the launch-time resolution finds no staged binary for the new
commit (`sut.binary_for(commit)` returns `None`), and the code does **not**
refuse or warn — it silently falls through:

```
runner.py:866  if sut_ref: ... if staged is not None: env["STELLA_BINARY"]=...; return env
runner.py:880  binary = os.environ.get("STELLA_BINARY")   # ambient fallback
runner.py:881  if binary: env["STELLA_BINARY"] = binary
               # <- if THIS is also unset, no STELLA_BINARY env var is set at all,
               #    with no error and no warning anywhere in the run.
```

`stella_harbor/__init__.py::_locate_binary()` then applies **its own**
fallback order when the env var is entirely absent: `STELLA_BINARY` → `stella`
on `PATH` → walk up from cwd for `./target/release/stella`. On a dev machine
that has a `stella` build on `PATH` for daily interactive use (as this repo's
own `AGENTS.md`-following workflow encourages), that resolves to **the host's
own native macOS build** — architecture-incompatible with every Linux task
container.

## Evidence (reproduced 2026-08-07)

1. `POST /api/sut/build {"ref":"main"}` at 15:19:57 resolved `main` →
   `447f3393...`, built successfully in 228s (verified: real ELF x86-64,
   `sha256:dc45ccdb...`), cached at
   `~/.arenabench/sut/447f33936700f71aec6f55024a1b4bde1a9302f4/stella`.
2. `origin/main` advanced to `ebbbe510...` at 15:20:24 — *during* that build
   (confirmed via `git log origin/main`, commit message: "feat(arenabench): an
   outcome_reason taxonomy... (#2095)").
3. A match launched at 15:24:56 via `arenabench run <template.toml>` (no
   `sut_ref` in the TOML — it isn't wired through file-based matches at all,
[...]

PR #2265, opened 2026-08-08
fix(arenabench): a seat that never loaded its adapter is not an agent loss (#2127, #2192)

### LINKED ISSUE #2127: arenabench: server reads harbor-adapter sources from its launch worktree per trial install — deleting that worktree mid-match RuntimeErrors every later trial (4/10 in cc00894779ff), scored as agent losses
## Problem

The arenabench server resolves the harbor-adapter Python sources (`bench/harbor_adapter/stella_harbor/`) from **its own launch worktree, at every trial's install time**. If that worktree is deleted while a match is running, every trial that installs afterwards dies in agent setup with:

```
RuntimeError: no Python sources found under
/Users/macanderson/Projects/stella/.claude/worktrees/arenabench-witness-smoke/bench/harbor_adapter/stella_harbor
```

(raised from `stella_harbor/__init__.py::_source_tree_sha256`, called by `_adapter_content_sha256` in `install`).

This happened in match `cc00894779ff` (2026-08-07, "Sonnet 5, full pipeline, 10-task"): the server on port 8900 was launched from the `arenabench-witness-smoke` worktree; that worktree was deleted mid-match (between 23:36:37Z, when `extract-elf` installed fine, and 23:51:35Z, when `sanitize-git-repo` failed). **4 of 10 trials were lost**: `fix-git`, `git-leak-recovery`, `sanitize-git-repo`, `prove-plus-comm` — each 0 steps, 0 tokens, dead in 30–60s. The server process (its cwd literally inside the deleted worktree) kept accepting and burning trials.

Compounding defect: the dashboard scored these as **agent losses**. The match API reports `infrastructure: 0` and `judged: 10`, so `solve_rate` = 40% includes 4 trials the agent never got to start. The true agent record was 4/6. Per the repo's own bench rules, operational aborts must be named and excluded from the denominator.

## Evidence

- `~/.arenabench/matches/cc00894779ff/jobs/cc00894779ff-stella-sonnet-5-full-pipeline/fix-git__yPKRSwX/exception.txt` (and the three sibling trials) — identical traceback.
- `lsof -p <server pid> -a -d cwd` at the time showed cwd `.../arenabench-witness-smoke/arenabench` (stale inode, empty dir).
- Match API `GET /api/matches/cc00894779ff`: `"infrastructure": 0`, `"solve_rate": 40.0`.

## Repro / verify

1. Launch the arenabench server from worktree A. Start a multi-trial match.
2. After the first trial begins, `git worktree remove` A (or `rm -rf` it).
3. Every subsequent trial fails agent setup with the RuntimeError above; the dashboard counts them as judged losses.

## Fix directions (either/both)

1. **Snapshot the adapter sources once at server start** (they are already content-hashed — stage them the same way SUT binaries are staged under `~/.arenabench/sut/<commit>/`), so a running server is immune to its launch tree disappearing. Alternatively fail the *match* fast with a clear operational error instead of burning trials one at a time.
2. **Classify agent-setup failures as infrastructure**, not judged agent losses: `totals.infrastructure` should count them and `solve_rate`'s denominator should exclude them.

## Related

- #2098 (sut_ref build/launch race — same family: launch-time resolution against a moving/vanishing source tree).

## Definition of done

A match whose server worktree is deleted mid-run either continues installing trials from the staged snapshot or aborts the match with an operational error; in both cases no agent-setup failure appears in `solve_rate`'s denominator. A regression test covers install-after-source-deletion.

### LINKED ISSUE #2192: arenabench: a missing stella_harbor import becomes a zero-work seat that exits 0
## Problem

`arenabench/arenabench/harbor_agent.py` builds a precise, actionable `ImportError`
when `stella_harbor` is not importable — and then discards it three lines later:

```python
def _load_base_agent():
    try:
        from stella_harbor import StellaAgent
    except ImportError as exc:
        raise ImportError(
            "arenabench's Stella contestant needs the `stella_harbor` adapter "
            "on PYTHONPATH. Point ARENABENCH_STELLA_ADAPTER at ..."
        ) from exc
    return StellaAgent


try:
    _Base = _load_base_agent()
except ImportError:
    _Base = object          # <-- the message above is thrown away here


class ArenaStellaAgent(_Base):
    ...
```

With `_Base = object`, `ArenaStellaAgent(...)` raises
`TypeError: ArenaStellaAgent() takes no arguments` inside the Harbor
subprocess — and **Harbor exits 0**. The arena therefore records the seat as
*done*, with zero steps and no model call, which is indistinguishable in the
match UI from an agent that ran and failed the task. Observed on real matches
(`44c124e06ef7`, `3ca70da630ed`) and diagnosable only by opening the seat log.

This is the same class as #2098: a fallback that turns a configuration error
into a wrong measurement instead of a refusal.

## Why it survived

The fallback exists so the module is importable in environments that only need
its constants (tooling, type checking, the test suite collecting it). That goal
is real; sacrificing the runtime failure to reach it is not.

## What "done" looks like

Make the failure loud at the point where it matters — instantiation — without
making the module unimportable:

- Keep the module importable, but have `ArenaStellaAgent.__init__` raise the
  saved `ImportError` (stash it rather than dropping it) when `_Base is object`.
  A subclass of `object` that refuses to instantiate is strictly better than one
  that raises a `TypeError` about argument counts, because the message names the
  fix.
- Ensure the exception reaches Harbor as a **non-zero exit**. If Harbor swallows
  agent-construction exceptions, the arena's own trial recording must mark the
  trial an operational abort rather than a scored loss — see the existing
  handling for `NonZeroAgentExitCodeError`.
- Witness test: construct `ArenaStellaAgent` with `stella_harbor` absent from
  `sys.modules`/`sys.path` (monkeypatch the loader to raise) and assert the
  raised error names `ARENABENCH_STELLA_ADAPTER`. It must fail on `main` with
  `TypeError: ... takes no arguments`.

## Files

- `arenabench/arenabench/harbor_agent.py` (lines ~193-215)
- `arenabench/arenabench/runner.py::_agent_environment` — where `PYTHONPATH` is
  assembled; #2185 makes it derive the adapter root from `sut.stella_repo()`
  when `ARENABENCH_STELLA_ADAPTER` is unset, which removes the most common
  trigger but not this trap (a variable set to the *wrong* path still lands
  here).

## Constraints

- `arenabench/` has its own `pytest` suite (`cd arenabench && uv run python -m
  pytest`); no Rust gate is involved.
- Do not make the module raise at import time — that breaks collection.

Found while fixing #2185.

macanderson:
## The bug, twice

Two defects with one consequence: a trial that never made a model call was counted in `solve_rate`'s denominator as though the agent had tried and failed.

**#2127 — the adapter was read from the launch checkout, at every trial.** In match `cc00894779ff` the server was launched from a git worktree, and that worktree was deleted while the match ran. The four trials that installed afterwards — `fix-git`, `git-leak-recovery`, `sanitize-git-repo`, `prove-plus-comm` — each died in agent setup naming a path that no longer existed: 0 steps, 0 tokens, dead in 30–60s. The server process, its cwd literally inside the deleted worktree, kept accepting and burning trials. The dashboard reported `infrastructure: 0`, `judged: 10`, **`solve_rate` 40%** — against a true agent record of 4/6.

**#2192 — an absent adapter failed silently.** `harbor_agent` built a precise, actionable `ImportError` and then discarded it three lines later with `_Base = object`. `ArenaStellaAgent(...)` then raised `TypeError: takes no arguments` inside the Harbor subprocess, and **Harbor exits 0** — so the arena recorded the seat as *done*, with zero steps and no model call, indistinguishable in the match UI from an agent that ran and failed. Observed on `44c124e06ef7` and `3ca70da630ed`, diagnosable only by opening the seat log.

Both are the class #2098 named: *a fallback that turns a configuration error into a wrong measurement instead of a refusal.*

## The fix

**Stage the adapter instead of re-reading it** (`arenabench/adapter.py`, new). Sources are copied once, on first use, into `<arenabench home>/adapter/<digest>/`; every later trial in the process imports the copy. A running server is now immune to its own launch tree disappearing. Two details that are load-bearing:

- The digest is **content-addressed over every file** in the package, not just `*.py` — the adapter ships data beside its code, and a digest that ignored those would reuse a stale staging directory after they changed. An edited adapter stages fresh rather than being shadowed.
- The process cache is keyed by **(source, arena home)**, not source alone. Keyed on source alone it hands back a path under whichever home was set first — which in the test suite meant staging into the developer's real `~/.arenabench` and then reusing it from a tmp-home test. That is a bug I introduced and caught mid-review; the key is why it cannot recur.

Copy-then-rename, so a concurrent match cannot import a half-copied tree.

**Refuse loudly instead of silently** (`harbor_agent.py`). The module stays importable — that fallback exists so the web UI can list Stella on a machine that never built it — but the stand-in base now raises the saved error **at construction** rather than being `object`. The error is a named `StellaAdapterMissingError(ImportError)`, so existing `except ImportError` paths still catch it.

**Keep both out of the denominator** (`telemetry.py`). `AdapterUnavailableError` and `StellaAdapterMissingError` join `INFRASTRUCTURE_FAILURES` and `VOID_SETUP`. This is the half that actually moves the published number: a seat whose adapter never loaded did not lose a contest, it never entered one.

## A test-fixture change worth reviewing

A Stella seat with no adapter now refuses at launch by name. That made four tests in `test_sut.py` fail — they build a fake checkout with no `bench/harbor_adapter`, so they were previously exercising a launch that *would have produced a zero-work seat*. The `repo` fixture now creates a minimal `bench/harbor_adapter/stella_harbor/`, because a real Stella checkout has one.

**No test was deleted.** `test_the_launch_wires_the_adapter_from_that_same_checkout` keeps its name and its intent; its assertion moves from the literal source path to the staged root, because the staged copy is now what gets wired. Its docstring says why.

## Witness

`arenabench/tests/test_adapter.py`:

- `test_a_staged_adapter_outlives_the_checkout_it_was_copied_from` — stages, deletes the worktree, stages again. The `cc00894779ff` sequence.
- `test_a_missing_adapter_refuses_at_construction_naming_the_fix` — pins that the error names `ARENABENCH_STELLA_ADAPTER` and is **not** a `TypeError`.
- `test_both_failures_are_infrastructure_not_agent_losses` — fails without the telemetry change with `AdapterUnavailableError would score as a loss`. Verified by removing the names and re-running.
- Plus content-addressing (`*.py` and data-file), and the `ImportError` subclass contract.

```
pytest -q (arenabench, non-docker)    all pass
ruff check / ruff format              clean on every file touched
```

## Not fixed here

`arenabench`'s **own** package root is still resolved from `__file__`, so a deleted launch tree can still strand `import arenabench` in the Harbor subprocess by the same mechanism. Staging only covers `stella_harbor`, which is what the observed traceback named. Worth its own issue — I'll file it against this PR.

Closes #2127
Closes #2192

## Summary by Sourcery

Ensure Harbor-based Stella seats only count as contests when their adapter successfully loads, staging adapter sources for robustness and classifying adapter load failures as infrastructure rather than agent losses.

New Features:
- Introduce adapter staging infrastructure that copies the harbor adapter into an ArenaBench home directory and reuses a content-addressed, process-cached copy across trials.

Bug Fixes:
- Prevent seats whose harbor adapter is missing or whose sources disappear mid-match from being treated as agent losses or counted in solve rate.
- Fix silent adapter import failures by raising a named StellaAdapterMissingError at agent construction instead of falling back to an unusable base class.

Enhancements:
- Unify ArenaBench home resolution for staged artifacts and update agent environment wiring to use staged adapter paths rather than direct checkout paths.
- Improve error signalling around missing harbor adapters so operators receive actionable guidance and downstream code can classify the failures correctly.

Documentation:
- Document the new adapter staging behavior, error types, and their impact on solve rate classification in module and test docstrings.

Tests:
- Add a dedicated adapter test suite covering staging behavior, content-addressing, missing-source handling, and infrastructure classification of adapter failures.
- Update system-under-test fixtures and adapter wiring tests to expect staged adapter paths and explicit failures when the harbor adapter is absent.

### DIFF
--- a/arenabench/arenabench/adapter.py
+++ b/arenabench/arenabench/adapter.py
@@ -1,0 +1,141 @@
+"""Staged harbor-adapter sources, so a match outlives its launch checkout.
+
+The arena resolved the ``stella_harbor`` package out of the checkout the
+server was launched from, and resolved it again at *every* trial's install.
+In match ``cc00894779ff`` that checkout — a git worktree — was deleted while
+the match was running. The four trials that installed afterwards each died in
+agent setup naming a path that no longer existed, 0 steps and 0 tokens
+apiece, and the dashboard scored all four as agent losses: ``solve_rate``
+read 40% against a true agent record of 4/6 (#2127).
+
+Staging removes the dependency rather than accounting for it. The sources are
+copied once, on first use, into ``<arenabench home>/adapter/<digest>/``, and
+every later trial in the process reads that copy. Two consequences worth
+naming:
+
+* A running server is immune to its own launch tree disappearing.
+* A match measures exactly one adapter revision — which is what the
+  provenance record already claims, and was not previously guaranteed, since
+  an operator editing the adapter mid-match changed what later trials ran.
+
+The digest is content-addressed, so an edited adapter stages a fresh
+directory on the next match rather than being shadowed by a stale copy.
+"""
+
+from __future__ import annotations
+
+import hashlib
+import shutil
+from pathlib import Path
+
+from .sut import arenabench_home
+
+__all__ = [
+    "PACKAGE_NAME",
+    "AdapterUnavailableError",
+    "adapter_root",
+    "stage_adapter",
+]
+
+#: The package a Stella seat must be able to import. The adapter directory on
+#: ``PYTHONPATH`` is the one that *contains* it.
+PACKAGE_NAME = "stella_harbor"
+
+
+class AdapterUnavailableError(RuntimeError):
+    """The harbor adapter's sources cannot be found or staged.
+
+    Named rather than a bare :class:`RuntimeError` so
+    :data:`arenabench.telemetry.INFRASTRUCTURE_FAILURES` can recognise it: a
+    trial that never reached a state where the agent could act is an
+    operational abort, and must stay out of ``solve_rate``'s denominator.
+    """
+
+
+def adapter_root() -> Path:
+    """Where staged adapter sources live, honouring ``ARENABENCH_HOME``."""
+    return arenabench_home() / "adapter"
+
+
+#: Staged roots already materialised by this process, keyed by resolved
+#: source path **and** the arena home it was staged under. Process-scoped on
+#: purpose: the point is that trial *n* does not re-read a source tree trial 1
+#: already copied, so the copy has to outlive the source within one server run.
+#:
+#: The home is part of the key because it is read from the environment, and a
+#: cache keyed on the source alone hands back a path under whichever home
+#: happened to be set first — which in the test suite meant staging into the
+#: developer's real ``~/.arenabench`` and then reusing it from a tmp-home test.
+_staged: dict[tuple[Path, Path], Path] = {}
+
+
+def _digest(source: Path) -> str:
+    """A content digest over the adapter package.
+
+    Every file under the package is folded in — not just ``*.py`` — because
+    the adapter ships data beside its code, and a digest that ignored those
+    would reuse a stale staging directory after they changed.
+    """
+    package = source / PACKAGE_NAME
+    files = sorted(p for p in package.rglob("*") if p.is_file())
+    if not files:
+        raise AdapterUnavailableError(
+            f"no {PACKAGE_NAME} sources under {package}. Point "
+            "ARENABENCH_STELLA_ADAPTER at <stella>/bench/harbor_adapter."
+        )
+    digest = hashlib.sha256()
+    for path in files:
+        digest.update(str(path.relative_to(package)).encode())
+        digest.update(b"\0")
+        digest.update(path.read_bytes())
+        digest.update(b"\0")
+    return digest.hexdigest()[:16]
+
+
+def stage_adapter(source: Path) -> Path:
+    """Stage ``source``'s adapter package and return the root to import from.
+
+    Idempotent per process and per content digest. The returned path is a
+    directory containing :data:`PACKAGE_NAME`, suitable for ``PYTHONPATH``.
+
+    Raises :class:`AdapterUnavailableError` when the sources are absent and
+    nothing has been staged for this source yet — the honest failure, because
+    there is no adapter to run and every later step would misreport why.
+    """
+    source = source.expanduser().resolve()
+    root = adapter_root()
+    key = (source, root)
+    cached = _staged.get(key)
+    if cached is not None and (cached / PACKAGE_NAME).is_dir():
+        return cached
+
+    if not (source / PACKAGE_NAME).is_dir():
+        raise AdapterUnavailableError(
+            f"the harbor adapter is not at {source}: no {PACKAGE_NAME} "
+            "directory. If the arena's checkout moved or was deleted mid-run, "
+            "restart the server from a live checkout, or point "
+            "ARENABENCH_STELLA_ADAPTER at <stella>/bench/harbor_adapter."
+        )
+
+    destination = root / _digest(source)
+    marker = destination / PACKAGE_NAME
+    if not marker.is_dir():
+        destination.mkdir(parents=True, exist_ok=True)
+        # Copy to a sibling first, then rename: a half-copied tree that a
+        # concurrent match started importing would fail in a way that reads
+        # like a corrupt adapter rather than a racing one.
+        pending = destination.with_name(f"{destination.name}.pending")
+        shutil.rmtree(pending, ignore_errors=True)
+        shutil.copytree(source / PACKAGE_NAME, pending / PACKAGE_NAME)
+        try:
+            (pending / PACKAGE_NAME).rename(marker)
+        except OSError:
+            # Another match staged the identical digest first. Its copy is
+            # byte-identical by construction, so losing the race is fine.
+            if not marker.is_dir():
+                raise
+        finally:
+            shutil.rmtree(pending, ignore_errors=True)
+
+    _staged[key] = destination
+    return destination
--- a/arenabench/arenabench/adapter.py
+++ b/arenabench/arenabench/adapter.py
@@ -26,6 +26,7 @@ from __future__ import annotations
 
 import hashlib
 import shutil
+import tempfile
 from pathlib import Path
 
 from .sut import arenabench_home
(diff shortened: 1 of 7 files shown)


macanderson:
Residue named in the description is filed as #2266 — arenabench's own package root is still resolved from `__file__`, so a deleted launch tree can strand `import arenabench` in the Harbor subprocess by the same mechanism this PR fixes for `stella_harbor`.

macanderson:
## Review: the denominator change is honest — one cache-poisoning race fixed on this branch

Bench-scoring PRs that move failures *out* of the denominator deserve the most suspicion, so I checked that axis first, and this one holds up.

**The classification is narrow.** It is exact class-name membership on Harbor's structured `exception_info.exception_type` (`telemetry.py:648-657`) — never a substring, never prose, never "any setup exception" or "0 steps". The two added names can each only be produced *before the first model call*: `StellaAdapterMissingError` solely from `from stella_harbor import StellaAgent` failing, `AdapterUnavailableError` solely in the arena runner pre-launch. An agent whose first action crashes the harness still arrives as Harbor's own names (`NonZeroAgentExitCodeError`, `AgentTimeoutError`) and **stays a loss** — pinned by the existing fixtures at `test_telemetry.py:1120-1142`. A verdict, when present, is read before the exception branch, so classification annotates and never overrides.

**Excluded trials are named, not vanished.** Per-trial `to_json` carries `failure`, `infrastructure` and `outcome_reason`; `aggregate` publishes `trials`, `judged` and `infrastructure` side by side with the "won 8 of 10" vs "won 8 of the 8 that managed to start" rationale. That meets the bar.

### Fixed on this branch: the `.pending` staging race (7b3bb0e7)

The Vercel bot's flag at `adapter.py:126` is real, and the consequence is worse than a transient stomp — it is **permanent cache poisoning**. `pending = destination.with_name(f"{destination.name}.pending")` is a shared deterministic name, so with the threaded server two matches staging one digest run this concurrently in one process: thread B's `rmtree(pending)` deletes A's finished copy, B starts recopying, and A's `rename` atomically publishes B's **in-flight partial tree**. Because the cache check is `if not marker.is_dir()`, that poisoned entry is permanent for the digest, and a partial tree that still imports means later matches silently run a wrong adapter — measurement corruption no classification can catch.

Fixed with `tempfile.mkdtemp` per attempt, dropping the shared-name pre-clean. That also makes the lost-race swallow honest for the first time: a marker that already exists can now only have come from a copy that *finished*.

The new test `test_concurrent_staging_never_publishes_a_half_copied_tree` forces the damaging interleaving deterministically (no sleeps, no raced timing) rather than hoping to catch it. Verified as a true witness both ways: against the shared `.pending` name it fails with `AssertionError: published a half-copied tree`; with `mkdtemp` it passes. Full `tests/test_adapter.py`, `test_outcome_reason.py` and `test_telemetry.py` are green, ruff check and format clean.

### One overstatement in the description

"the half that actually moves the published number" — the telemetry change does **not** move `cc00894779ff`'s number. Those four trials recorded Harbor's own (unclassified) setup-failure name, so re-reading that archive still publishes 40%. Deliberately not classifying a broad Harbor name is the correct narrow call; what actually fixes #2127 is the staging. The archive's known confound stands, but the sentence oversells which half did the work.

Minor, no action needed: the `AdapterUnavailableError` entry is effectively unreachable in `result.json` — it fires pre-launch, is caught at `runner.py:643-652` into `run.error`, and the seat produces zero trials, so nothing enters any denominator. Harmless belt-and-braces.

PR #2267, opened 2026-08-08
fix(gate): the file-size ratchet judges the change, not the tree (#2004)

### LINKED ISSUE #2004: gate: the file-size baseline is a shared cell — two correct PRs compose into a red main (3rd occurrence)
## Problem

`scripts/file-size-baseline.txt` is a **single shared cell that every growing PR
must write**, and two PRs that each write it correctly can still compose into a
red `main`. The guard asks a whole-tree question — "is this tree consistent with
this baseline snapshot?" — when the thing it wants to prevent is a per-change
question: "did *this change* grow a god file past its ceiling?"

Concretely, on `main` at `6c345532`:

```
$ bash scripts/check-file-size.sh
check-file-size: FAILED
  crates/stella-core/src/driver.rs grew to 2572 lines, over its baseline ceiling of 2571 (+1)
  crates/stella-pipeline/src/pipeline/tests.rs grew to 2537 lines, over its baseline ceiling of 2536 (+1)
```

Neither contributing PR is at fault. #1979 added three lines to `driver.rs`;
#1962 added a line to `pipeline/tests.rs`. Each regenerated the baseline against
a snapshot of `main` that did not yet carry the other's growth, each was green on
its own merge commit, and the composition is red. `main` then stays red — see
#1986 — so every subsequent PR inherits a failure it did not cause. PR #1992 sat
`MERGEABLE/BLOCKED` on exactly this with a tree byte-identical to `main` for both
the baseline and both failing files.

**This is the third occurrence of the same shape**: #1761 (`deck.rs` 1517 vs a
1510 ceiling), #1782 (#1776 grew `driver.rs` +16 and `registry.rs` +35), and now
#2003. All three were fixed by regenerating and closed, leaving the mechanism
intact.

AGENTS.md already names this exact failure mode for a different file, and rejects
the shared-cell design there:

> Deliberately no total: the guard checks each step by name, so two PRs adding
> different guards merge cleanly, while a spelled-out count is one shared cell
> both must write — and two of them collided on it twice in a day, each time
> leaving `main` red for everyone (#1883).

The baseline has the identical shape and has not had the identical treatment.

## Files involved

- `scripts/check-file-size.sh` — the guard; the awk pass at ~L125–165 judges the
  working tree against the baseline map
- `scripts/file-size-baseline.txt` — the shared cell (generated; 30 entries)
- `scripts/test-file-size.sh` — the guard's own hermetic tests, where a new case
  belongs
- `Makefile` — `file-size` and `file-size-update` targets
- `.github/workflows/ci.yml` — runs the guard in the required job
- `AGENTS.md` § "God files — plan around them, never into them" — documents the
  ratchet's contract; any semantic change must be reflected here

## Reproduce

The skew is reconstructible from history without waiting for a new collision:

```sh
git checkout -b skew-repro e0fbbe02          # main, one commit before #1979's growth landed
bash scripts/check-file-size.sh              # OK
git cherry-pick ac4a13fd                     # #1979: driver.rs +3, baseline regenerated
bash scripts/check-file-size.sh              # OK
git checkout -b skew-other e0fbbe02          # a sibling PR branched from the same point
# ...regenerate the baseline here for an unrelated god file, merge both...
```

The already-landed instance is simpler to observe:

```sh
git checkout -b probe origin/main
bash scripts/check-file-size.sh              # FAILED, +1 on two files, with no local change
```

## Proposed fix

Make the ratchet judge **the change**, not the tree. When a file exceeds its
recorded ceiling, the guard should distinguish two cases:

1. **The diff under test grew it** → fail, as today. This is the bloat the
   ratchet exists to block, and the message should stay exactly as it is.
2. **The file already exceeded its ceiling at the merge base** → the growth
   arrived from another PR that was itself reviewed. Do not fail the innocent PR.
   Report it as drift the baseline owes an update, and let the *branch-level*
   check (or a scheduled job on `main`) be what demands the regeneration.

A merge-base comparison is enough to separate them; `scripts/impacted-crates.sh`
already establishes the precedent for a guard reading the pushed diff, and the
pre-push hook already computes a range.

Alternatives worth weighing in the fix, not pre-decided here:

- **Require branches up to date before merging** (branch protection). Solves it
  without touching the guard, but slows every merge and does nothing for the
  hook's local run. Interacts with #1986.
- **Regenerate the baseline in CI and commit it** — removes the shared cell by
  removing the human write, but a generated file that mutates on `main` outside a
  PR is its own hazard, and it would defeat the "visible in review" property
  AGENTS.md deliberately relies on.
- **Per-file baseline fragments** (one file per entry, à la the GATE_STEPS
  by-name fix) — cheapest conceptually, and two PRs touching different god files
  then never collide. Does not help two PRs touching the *same* god file.

## Constraints

- The baseline is **generated**; `make file-size-update` must remain the only
  supported way to write it, and hand-editing must stay unsupported.
- **The baseline may only ever shrink** in entry count — `check-file-size.sh`
  L27–31 enforces that an entry dropping to ≤ `LIMIT` becomes obsolete and must
  be retired. Any redesign must preserve that.
- Only the ceiling direction is enforced, on purpose (L34–35): a baseline file
  shrinking must not fail. Preserve that.
- `check-god-files.sh` cross-checks the baseline against `AGENTS.md` and every
  crate README; the baseline is the tiebreaker (#1435). A format change touches
  that guard too.
- Global guards are never narrowed by `CARGO_SCOPE` — "a 1500-line file is a fact
  about the repository, not about a crate" — so a merge-base-aware guard must
  still behave correctly when the range is the whole workspace or absent.

## Definition of done

- Two PRs that each grow a *different* god file, each regenerating the baseline
  against a `main` lacking the other's growth, compose into a **green** `main`.
- A PR that genuinely grows a god file past its ceiling still fails, with the
  current message unchanged.
- A case in `scripts/test-file-size.sh` covering the skew, constructed the way
  the repro above does — hermetic, no network, no reliance on real repo history.
- `AGENTS.md` § "God files" updated to state what the ratchet now judges.
- `make gate` green; `check-god-files.sh` still green.

Related: #1986 (`main` goes red and stays red), #1883 (the shared-cell lesson,
applied to `GATE_STEPS`), #1435 (the three-copy cross-check), #1761 and #1782
(prior occurrences), #2003 (this occurrence's unbreak).

macanderson:
## The bug

`scripts/file-size-baseline.txt` is one shared cell that every growing PR must write, and **three times running** (#1761, #1782, #2003) two PRs that each wrote it *correctly* composed into a red `main`. Each regenerates the whole baseline against a snapshot of `main` that does not yet carry the other's growth, so each records a stale ceiling for a file it never touched. The merge is textually clean — the two sides edit different lines — and the result is a ceiling one line below an actual. `main` then stays red and every subsequent PR inherits a failure it did not cause; #1992 sat blocked on exactly this with a tree byte-identical to `main`.

The detail that names the fix: **in that composition the file never grew.** Its ceiling moved *down* underneath it. A guard with one tree to look at cannot tell those two apart, because "is this tree consistent with this baseline snapshot?" is a whole-tree question and the thing worth preventing is a per-change one.

## The fix

The ratchet now judges the change:

```
fail only when  current > max(ceiling, size at base)
```

Both halves of the `max` earn their place, and that is the whole design:

- **`ceiling` alone** is the old check, and it fails the innocent PR above.
- **`size at base` alone** would pass a file that is already over its ceiling and that *this* change grows further — turning inherited drift into a standing licence to bloat. That is strictly worse than the red `main` being fixed here.

Taking the larger fails a change that genuinely grows a god file past what it inherited, and stays silent when the violation arrived from somewhere else. A ceiling raised deliberately via `--update` passes exactly as before, since the regenerated ceiling equals the current size — that path is untouched.

**Where the base comes from** is the same pair `scripts/check-deleted-tests.sh` already uses, for the same reason: on a `pull_request` the checkout is `refs/pull/N/merge`, so `HEAD^1` is the base branch tip and the question becomes *"does this merge grow a god file?"* — the question a required check should answer. `ci.yml` already sets the `fetch-depth: 2` that needs, for that guard. Locally it is the merge base with `origin/main`; on a linear push it is `HEAD^1`.

**When no base resolves** — shallow clone, root commit, no origin — the guard falls back to the old whole-tree check. That direction is deliberate and is pinned by a test: an unresolvable base must make the ratchet *stricter*, never weaker.

Drift is still reported, on stderr and in the summary line, so the baseline debt stays visible rather than merely tolerated. It just no longer fails the next PR to walk past it.

## Witness tests

`scripts/test-file-size.sh` gains six hermetic cases that rebuild the skew from two commits in a throwaway repo — no network, no reliance on this repository's real history. Verified the artisanal way, by running the new suite against `origin/main`'s guard:

```
$ git show origin/main:scripts/check-file-size.sh > scripts/check-file-size.sh
$ ./scripts/test-file-size.sh
FAIL B1 a ceiling lowered under an untouched file is not this change's failure
FAIL B2 inherited drift does not fail a change that touched something else
ok   B3 a change that genuinely grows a god file still fails
ok   B4 growth on top of inherited drift still fails
ok   B5 an unresolvable base falls back to the strict whole-tree check
FAIL B6 a refs/pull/N/merge checkout finds its base branch tip unaided
passed 10, failed 3
```

**B1, B2, B6 are the witnesses** — they fail on the old code and pass on the new. **B3, B4, B5 pass on both, which is the point of them**: they constrain the fix rather than being satisfied by it. B4 is the one I would ask a reviewer to look at hardest — it is the case a naive *"already over at the base, so ignore it"* rule would wave through, and it is why the rule takes the `max` rather than either term alone. B6 exists because B1–B4 all pass the base in by hand, which left the rung production actually depends on as the one rung nothing exercised.

With the fix in place: **13 passed, 0 failed** (the 7 pre-existing language-coverage cases are untouched and still green).

## Verification

```
./scripts/test-file-size.sh          13 passed, 0 failed
make guards-fast                     exit 0
shellcheck scripts/check-file-size.sh scripts/test-file-size.sh    clean
./scripts/check-file-size.sh         OK — 1187 files … (none grew by this change)
./scripts/check-god-files.sh         OK — 22 god files, named identically in AGENTS.md and every crate README
```

`make gate`'s Rust tiers are unaffected — this change is shell and markdown only.

## Definition of done, from the issue

- [x] Two PRs growing *different* god files, each regenerating against a `main` lacking the other's growth, compose green — B1/B2, and B6 in the real merge shape.
- [x] A PR that genuinely grows a god file past its ceiling still fails, message unchanged — B3, asserting the exact original string.
- [x] A hermetic case covering the skew, built like the issue's repro.
- [x] AGENTS.md § "God files" updated to state what the ratchet now judges.
- [x] `check-god-files.sh` still green; guards green.

## Scope I deliberately did not take

The same shared-cell shape could in principle red an innocent PR through the `OBSOLETE` and `STALE` branches (a sibling retires a baseline entry you did not). That has never been observed, and widening the change-relative rule to those paths without a witness would be speculative — the three recorded occurrences are all the `GREW` branch. Noted here rather than silently left.

Also worth stating plainly, since two issues were recently written on the opposite premise: **splitting a god file buys structure, not slack.** `--update` retightens every ceiling to its file's current size, so a freshly split file sits at zero headroom again — see my comment on this issue. This PR is what removes the tax; the split does not.

Closes #2004

---

## Follow-up found in this PR's own CI log

The first green run exposed something no red check would have: **`.github/workflows/file-size.yml` — the cheap, requirable status context for this very guard — checked out at the default depth of 1**, under a comment stating full history was unnecessary because the guard "reads the working tree and the committed baseline, never a diff against the base."

That was true until this PR and is now false. At depth 1 the merge commit's parents are not in the clone, so the base cannot resolve and the guard falls back to the strict whole-tree check. The job stays green and reports the ratchet exactly as before — so the fix would have been **silently absent from the one job whose purpose is to report this guard cheaply enough to be required.** `ci.yml`'s guards job was already at `fetch-depth: 2` and unaffected; this workflow is now too, the same trade it already makes for `check-deleted-tests.sh`.

To stop that class of silent non-application recurring, the guard now names its mode in the summary line:

```
check-file-size: OK — 1187 … (none grew by this change). Judged against 20e47d9f.
check-file-size: OK — 1187 … (none grew by this change). No base resolved — strict whole-tree check.
```

The change-relative rule is silent by nature — it only shows itself when something drifts — so without that line a too-shallow checkout is indistinguishable from a clean run in every log it will ever print. Now it is one glance.

### DIFF
--- a/.github/workflows/file-size.yml
+++ b/.github/workflows/file-size.yml
@@ -43,9 +43,19 @@ jobs:
     name: file size ratchet
     runs-on: ubuntu-latest
     steps:
-      # Full history is not needed — the guard reads the working tree and the
-      # committed baseline, never a diff against the base.
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
+        with:
+          # Two, not one. The guard judges THE CHANGE, not the tree (#2004):
+          # on a `pull_request` event this checks out `refs/pull/N/merge`, and
+          # the ratchet reads that merge commit's first parent — the base
+          # branch tip — to ask whether a file went over its ceiling HERE or
+          # arrived that way. At the default depth of 1 the parents are not
+          # in the clone, the base cannot resolve, and the guard falls back to
+          # the strict whole-tree check: green on this PR, but blind to the
+          # difference this workflow exists to draw. It compares two TREES,
+          # not two histories, so one extra commit is the whole cost — the
+          # same trade ci.yml's guards job makes for check-deleted-tests.sh.
+          fetch-depth: 2
 
       # The guard's own tests first: they are hermetic and they are what proves
       # it can still FAIL. A ratchet that has quietly become incapable of
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -424,6 +424,22 @@ inputs, not review afterthoughts:
   an escape hatch for a genuinely irreducible line (a module declaration in
   an already-oversized `lib.rs`), never something a plan may assume.
 
+**The ratchet judges your change, not the tree** (#2004). A file over its
+ceiling fails only when `current > max(ceiling, size in the base tree)` — so a
+violation that was already there before your branch existed is reported as
+drift and does not fail you, while anything your change grows past what it
+inherited fails exactly as it always did. The base is the merge commit's first
+parent on a pull request, the merge base locally, and nothing at all in a
+shallow clone or a fresh repository — where the guard falls back to the strict
+whole-tree check, because an unresolvable base must make it stricter, never
+weaker. This exists because the baseline is one shared cell every growing PR
+must write, and three times running two PRs that each wrote it correctly
+composed into a red `main` that then failed everyone downstream. Two
+consequences for planning: **do not "fix" a red ratchet you did not cause** by
+folding a baseline regeneration into an unrelated PR — land it on its own from
+a fresh `main`; and note that splitting a god file buys structure, not slack,
+since `--update` retightens every ceiling to its file's current size.
+
 The workspace's Rust god files, by crate (the bench harness's Python offenders
 sit in the same baseline). Each file's ceiling lives in
 `scripts/file-size-baseline.txt` and is deliberately not repeated here: that
--- a/Makefile
+++ b/Makefile
@@ -324,7 +324,7 @@ automerge-nudge-test: ## Test which PR the auto-merge nudge picks (hermetic; not
 	./scripts/test-automerge-nudge.sh
 
 .PHONY: file-size-test
-file-size-test: ## Test which languages the file-size ratchet actually watches (hermetic; not part of `gate`)
+file-size-test: ## Test the file-size ratchet's language coverage and its change-relative judgement (hermetic; not part of `gate`)
 	./scripts/test-file-size.sh
 
 .PHONY: guard-sigpipe-test
--- a/scripts/check-file-size.sh
+++ b/scripts/check-file-size.sh
@@ -349,5 +349,16 @@ drift_note=""
 if [ -n "$drift" ]; then
   drift_note=" $(printf '%s' "$drift" | grep -c '^') file(s) carry inherited drift (see above)."
 fi
+# Say which mode ran. The change-relative rule is silent by nature — it only
+# shows itself when something drifts — so a checkout too shallow to resolve a
+# base would fall back to strict and read as an ordinary green line forever.
+# That is exactly how .github/workflows/file-size.yml shipped its ratchet as a
+# whole-tree check while this script believed otherwise, and naming the mode is
+# what makes the difference legible in a log rather than a thing to re-derive.
+if [ -n "$base_commit" ]; then
+  mode_note=" Judged against $(git rev-parse --short "$base_commit" 2>/dev/null || echo "$base_commit")."
+else
+  mode_note=" No base resolved — strict whole-tree check."
+fi
 trap '' PIPE
-echo "check-file-size: OK — $tracked Rust/Python/shell files, none over $LIMIT lines except $grandfathered grandfathered (none grew by this change).${drift_note}" || true
+echo "check-file-size: OK — $tracked Rust/Python/shell files, none over $LIMIT lines except $grandfathered grandfathered (none grew by this change).${mode_note}${drift_note}" || true
--- a/scripts/check-file-size.sh
+++ b/scripts/check-file-size.sh
@@ -51,6 +51,47 @@
 # Test files count. A 2,750-line test module is as hard to navigate as any other,
 # and exempting tests would be an obvious loophole.
 #
+# ── What the ratchet judges: the CHANGE, not the tree (#2004) ─────────────────
+#
+# The baseline is a single shared cell that every growing PR must write, and for
+# three occurrences running (#1761, #1782, #2003) two PRs that each wrote it
+# CORRECTLY composed into a red `main`. Each regenerates the whole baseline
+# against a snapshot of `main` that does not yet carry the other's growth, so
+# each records a stale ceiling for a file it never touched. The merge is
+# textually clean — the two sides edit different lines — and the result is a
+# ceiling one line below an actual. `main` then stays red, and every subsequent
+# PR inherits a failure it did not cause; #1992 sat blocked on exactly this with
+# a tree byte-identical to `main`.
+#
+# Note what did NOT happen in that composition: the file never grew. Its ceiling
+# moved DOWN underneath it. A guard asking "is this tree consistent with this
+# baseline snapshot?" cannot tell those apart, because it has only one tree to
+# look at. So it asks the per-change question instead, against the base:
+#
+#     fail only when  current > max(ceiling, size at base)
+#
+# The `max` is the whole rule, and each half earns its place:
+#
+#   - `ceiling` alone is today's check, and fails the innocent PR above.
+#   - `size at base` alone would pass a file that is already over its ceiling
+#     and that THIS change grows further — drift would become a licence to
+#     bloat, which is the one outcome worse than the red main.
+#
+# Taking the larger fails a change that genuinely grows a god file past what it
+# inherited, and is silent when the violation arrived from somewhere else. A
+# ceiling raised deliberately via `--update` still passes exactly as before:
+# the regenerated ceiling equals the current size.
+#
+# The base is the same pair `scripts/check-deleted-tests.sh` uses, for the same
+# reason: on a `pull_request` event the checkout is `refs/pull/N/merge`, so HEAD
+# is the merge commit and HEAD^1 is the base branch tip. That asks "does this
+# MERGE grow a god file?", which is the question a required check should answer,
+# and it needs `fetch-depth: 2` — which ci.yml already sets.
+#
+# When no base can be resolved the guard falls back to judging the tree, exactly
+# as it did before. That direction is deliberate: an unresolvable base must
+# never make the ratchet weaker, only stricter.
+#
 # Uses portable POSIX tools so it runs on a bare CI runner (macOS ships bash 3.2,
 # so no associative arrays).
 set -euo pipefail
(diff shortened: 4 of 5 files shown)


macanderson:
## Review: changes requested — one rung of `resolve_base_commit` makes the pre-push gate false-green

The core design holds up. I verified the ratchet direction empirically rather than by reading the description: file at ceiling C=1600, PR A grows to 1610 with `--update` and merges, PR B branches after and adds +10 more with the baseline untouched → B **fails** (`max(ceiling=1610, base=1610) = 1610 < 1620`, exit 1). The invariant that survives is the right one: a PR passes only when `current ≤ ceiling` (a reviewable raise) or `current ≤ base` (no growth), and `base` itself can only rise through a PR that took a visible baseline diff. Merge skew can now produce stale ceilings but never uncounted lines. The `.github/workflows/file-size.yml` depth-2 fix is real and necessary — without it the required context would have silently judged the whole tree forever.

The hermetic tests are a genuine witness: 13/13 pass against this branch's guard, and against `origin/main`'s guard exactly B1, B2 and B6 fail. B4 (growth on top of inherited drift still fails) passes on both, which is what pins the fix from over-reaching. Guard conventions are respected — `git ls-files` enumeration untouched, consistent `wc -l` on both sides, no `stat`, shellcheck clean, and `GATE_STEPS` genuinely needs no change since only `file-size-test`'s help text moved.

### The blocking defect — `check-file-size.sh:187-190`

The comment says *"A merge commit means a `refs/pull/N/merge` checkout"*. That is false locally, and I built the counterexample: grow a god file +100 on a feature branch with no baseline touch, then merge trunk into the branch — the update-branch shape, i.e. the commit immediately before `git push`. `HEAD^2` now exists, so base = `HEAD^1` = **the developer's own pre-merge tip, which already contains the growth**. The guard prints

```
src/big.rs is 1700 lines against a ceiling of 1600 — already so at the base ... another change put them there
```

and exits **0**. The pre-push hook passes, CI reds an hour later, and the diagnostic actively blames someone else for your own growth. `check-deleted-tests.sh` gets away with the same `HEAD^1`/`HEAD^2` pair only because it runs solely on `pull_request` in CI; this guard borrowed the pair without that precondition.

**Fix:** reorder — try the `merge-base HEAD origin/main` rung (line 199) *before* the `HEAD^2` rung. In CI's PR/merge-queue checkout `origin/main` does not exist (single-ref fetch), so `HEAD^2` still fires there; and even if a future config materializes `origin/main`, `merge-base(refs/pull/N/merge, origin/main)` is the base tip = `HEAD^1`, so the reorder is identical under every checkout config. Locally it yields the true fork base and catches the growth. Please add a B7 case: merge trunk into feature after growing a god file, with an `origin/main` ref present, expect-fail.

Everything else about the fallback behavior is correct and I confirmed it fails **closed**: a bad `FILE_SIZE_BASE_REF` exits 1, and a shallow depth-1 clone makes both `HEAD^1`/`HEAD^2` and `origin/main` unresolvable → strict fallback, i.e. stricter, never open. That was the axis most likely to gut the gate silently, and it doesn't.

### Also required before merge (per "nothing left behind")

Two things this PR names only in prose and neither is filed:

1. **Drift convergence.** Nothing now guarantees the baseline retightens; drift persists green forever behind a stderr note (`check-file-size.sh:316`). It heals in practice only because any future `--update` regenerates all entries — an accident of the file format, not a mechanism.
2. **The OBSOLETE/STALE inheritance gap** the description explicitly declines to take.

A prose note in a PR body is not a handoff. Both need issues written as handoffs before this lands.

macanderson:
Filed the two residue items so they are handoffs rather than prose: **#2310** (nothing retightens the baseline now that inherited drift is non-fatal) and **#2311** (the OBSOLETE/STALE classification did not inherit the base-relative reasoning). Reference them from the description and the remaining blocker is the `resolve_base_commit` rung reorder plus its B7 case.

macanderson:
This merged before the `resolve_base_commit` fix landed, so the local false-green is now live in the gate machinery — filed as **#2317** with the reproduction and the exact reorder. #2310 and #2311 carry the two residue items. The core design and the hermetic suite are good; it is one rung ordering.

Issue #1773, opened 2026-08-06
arenabench: preflight Docker availability at match start instead of letting both seats crash into it


macanderson:
## Problem

When the Docker daemon is not running, `arenabench` still launches one `harbor run` subprocess per contestant; each exits within ~2s printing `Docker daemon is not running. Please start Docker and try again.` to its seat log. Since PR #1772 the match now correctly lands `failed` with that line in `match.note` — but the arena still spent a launch, wrote provenance, started recorder/snapshot supervisors if configured, and only tells the operator *after* the fact. Observed live on match `1eb525153bc5`.

## Proposal

In `MatchRunner.start` (`arenabench/arenabench/runner.py`), before launching any seat, probe Docker the same way the Harbor version floor is probed (`harbor.require_for_dataset` is the exemplar — refusal, not warning): e.g. `docker info` with a short timeout, or the cheapest equivalent Harbor itself uses. If the daemon is unreachable, refuse the start with an actionable error that reaches the HTTP API / UI (`match.note` + a 4xx from the start endpoint), naming the fix ('start Docker Desktop / colima').

Scope note: only matches whose spec runs `--env docker` (currently all of them) need the probe.

## Files

- `arenabench/arenabench/runner.py` — `MatchRunner.start`, `_launch`
- `arenabench/arenabench/harbor.py` — where the probe helper likely belongs, beside `require_for_dataset`
- `arenabench/tests/test_arena.py` — `TestFailedMatchStatus` (added in PR #1772) shows the Match/ContestantRun test conventions

## Done when

A match started with Docker down never launches a seat process, the start surface reports the refusal with the daemon-down message, and a witness test pins it (monkeypatch the probe at the `arenabench.harbor` seam, per `_fake_harbor` in `tests/test_arena.py`).

macanderson:
**Triage 2026-08-06:** P2 — bench-rig ergonomics; the failure has been surfaced honestly since #1772, this moves the refusal before launch/spend. Plan: as specced in the body; next step is a short-timeout `docker info` probe helper beside `harbor.require_for_dataset`, refused from `MatchRunner.start` with a 4xx + `match.note` naming the fix, witnessed by monkeypatching the probe at the `arenabench.harbor` seam per `_fake_harbor` in `tests/test_arena.py`.

macanderson:
Closing as **not planned** — a maintainer backlog decision, not a claim that this is fixed.

[...]

PR #2269, opened 2026-08-08
fix(arenabench): find or install a Harbor that can grade, instead of only refusing one that cannot

macanderson:
## What broke

Kicking off a **one-task Frontier-Bench match** failed before a single trial ran. Every seat died with the same message:

> Frontier Bench needs harbor >= 0.20.0, but `/opt/homebrew/bin/harbor` is 0.6.1. An older Harbor does not merely fail here — it drops task settings it does not recognise and grades against the wrong container topology…

The refusal itself is correct and stays. Harbor 0.6.1 does not know Frontier-Bench's `[verifier] environment_mode = "separate"`, pydantic's default `extra="ignore"` discards it, and the run completes with rewards from the wrong container topology — plausible numbers answering a different question, which is the worst failure a benchmark can have.

What was wrong is that **a correct Harbor was installed on that machine the whole time**, one directory away in `bench/evidence/frontier/.venv`. Nothing looked for it, because resolution had exactly one implicit source: `PATH`.

## What changed

A floor is now something to **satisfy** before it is something to fail on. `resolve_for_dataset` walks, in order:

1. an explicit pin (`ARENABENCH_HARBOR`, or the per-dataset override),
2. `PATH`, kept whenever it can grade the dataset,
3. Harbors arenabench installed itself, then this checkout's lane virtualenvs,
4. a fresh `uv` install of the exact pinned version,
5. and only then the refusal — which now names every binary it looked at and what version each was.

Two rules keep that convenience from becoming its own hazard:

- **A pin is never second-guessed.** A pinned Harbor is used, or refused *by variable name* — never quietly replaced. Pinning is how a lane declares its audited constant; a search that overrode it would make that declaration a lie.
- **A dataset with no floor resolves exactly as before**: `PATH`, and nothing else. Terminal-Bench 2.1 keeps running the 0.6.1 its claim rests on, rather than some newer binary that happens to be lying around. This is the property that lets the change land while a Terminal-Bench match is running.

### Resolution moved up to the match

It had been happening once per seat, and separately again when building the provenance record. That second lookup is the shape that once wrote down an apparatus which never ran (#2098) — and with discovery in play, the two could now legitimately disagree, recording 0.6.1 while 0.20.0 did the grading. One resolution per match feeds both, so:

- provenance cannot name a Harbor that did not grade,
- a Harbor that must be installed is installed **once**, not per seat,
- and a machine-wide problem is one fact about the match instead of *N* identical seat errors.

`resolve_harbor` is a named method rather than a defaulted `_launch` parameter on purpose: a parameter that silently re-resolves when omitted is precisely how #2098 happened.

## Witness

Six tests in `arenabench/tests/test_presets.py::TestResolutionSatisfiesTheFloorRatherThanOnlyCheckingIt`. Checked the artisanal way, in a detached worktree at `origin/main` with only the test file copied in: **6/6 fail there, 6/6 pass here.**

Because an `AttributeError` witness only proves a name is new, the behaviour was also verified end to end on the machine that hit the bug, against its real `PATH` and real virtualenvs:

| tree | frontier-bench |
|---|---|
| `origin/main` | `REFUSES: HarborTooOldError … /opt/homebrew/bin/harbor is 0.6.1` — the reported error, exactly |
| this branch | `LAUNCHES: harbor 0.20.0 (~/.arenabench/harbor/0.20.0/bin/harbor)` |

And the lane that must not move:

```
terminal-bench-2.1   -> 0.6.1    /opt/homebrew/bin/harbor  [found on PATH]
frontier-bench       -> 0.20.0   ~/.arenabench/harbor/0.20.0/bin/harbor
```

The provisioned binary was checked for real, not just for a version string: `harbor run --help` advertises `--include-task-name`, and `uv venv --relocatable` was confirmed to produce a `#!/bin/sh` wrapper that still runs after the staging directory is renamed into place — which is what makes the build-then-rename safe.

Full suite: 489 passed.

## Notes

- **No lock file.** Concurrent provisions are serialised by the rename itself: both may build, exactly one rename wins, and the loser takes the winner's copy — identical by construction, since the version is pinned exactly. Replacing a directory another process may be executing out of would be the more dangerous move. This machine routinely has three arena servers up, so this is the ordinary case, not a contrived one.
- **Plain `harbor`, not `harbor[modal]`.** Every match this resolves for launches `--env docker`; the Modal extra is a large dependency tree for a scheduler that would not be used. A lane wanting Modal builds and pins its own virtualenv, which is the path that already exists for saying so.
- **`ARENABENCH_HARBOR_NO_PROVISION=1`** turns installation back off for a hermetic runner.
- No new dependencies. `uv` is invoked if present and its absence is a named, actionable error.

## Left behind, filed

- **#2270** — `bench/evidence/frontier/env.sh` still hard-fails on a missing lane virtualenv, which is the same "a floor is only checked, never satisfied" shape on the shell surface. Not fixed here: that lane asserts an *exact* version for an audited evidence claim and needs `harbor[modal]`, so it is a different resolver with different constraints, not a call into this one.
- **Refs #1773** — that issue asks for the same move for Docker (preflight at match start instead of letting both seats crash into it). This lands the match-level preflight seam for Harbor; Docker is still unchecked, so the issue stays open.

### DIFF
--- a/arenabench/arenabench/harbor.py
+++ b/arenabench/arenabench/harbor.py
@@ -39,30 +39,75 @@ the Harbor that runs that dataset's matches, falling back to the global
 probe here takes the *resolved binary*, not a fresh global lookup, so the
 Harbor that is asked "do you still have ``--agent-import-path``?" is the one
 that will actually be launched.
+
+**A floor is a thing to satisfy, not only a thing to check.** Refusing to run
+is the right answer when the only Harbor on the machine would mis-grade. It
+was also, for a while, the answer when a perfectly good one sat one directory
+away — the Frontier lane's own virtualenv — because resolution had exactly one
+implicit source, ``PATH``, and PATH held an unrelated 0.6.1. A dataset that
+declares a floor therefore goes through :func:`resolve_for_dataset`, which
+looks for a Harbor that satisfies it and, finding none, installs one. The
+refusal survives underneath as the last resort it was always meant to be.
+
+Two rules keep that convenience from becoming its own hazard:
+
+* **An explicit pin is never second-guessed.** A binary named by
+  ``ARENABENCH_HARBOR`` or a per-dataset override is used, or refused with the
+  variable's name — never quietly replaced. Pinning is how a lane declares its
+  audited constant, and a search that overrode it would make that declaration
+  a lie.
+* **A dataset with no floor resolves exactly as it always did**: PATH, and
+  nothing else. Without a floor every Harbor grades identically as far as this
+  module can tell, so there is nothing to search *for* — and searching anyway
+  is how Terminal-Bench's pinned 0.6.1 lane would silently start running some
+  newer binary that happened to be lying around.
 """
 
 from __future__ import annotations
 
 import functools
+import itertools
 import os
 import re
 import shutil
 import subprocess
+from collections.abc import Iterator
+from dataclasses import dataclass
+from pathlib import Path
 
 __all__ = [
     "HARBOR_ENV",
+    "MANAGED_HOME_ENV",
+    "NO_PROVISION_ENV",
     "HarborTooOldError",
     "HarborUnavailableError",
+    "Resolution",
     "dataset_harbor_env",
     "harbor_bin",
     "harbor_version",
+    "managed_root",
     "parse_version",
+    "provision",
     "require_for_dataset",
+    "resolve_for_dataset",
     "supports_agent_import_path",
 ]
 
 #: Points ArenaBench at one specific Harbor, rather than whatever is on PATH.
 HARBOR_ENV = "ARENABENCH_HARBOR"
+
+#: Moves the directory :func:`provision` installs Harbors into.
+MANAGED_HOME_ENV = "ARENABENCH_HARBOR_HOME"
+
+#: Set truthy to make a missing Harbor a refusal instead of an installation.
+#: A hermetic CI runner wants this; a laptop about to start a match does not.
+NO_PROVISION_ENV = "ARENABENCH_HARBOR_NO_PROVISION"
+
+#: How long one ``uv`` invocation gets. Generous because this is a cold
+#: package download on a machine that may be running a benchmark at the same
+#: time, and the cost of being wrong is a spurious failure at the start of a
+#: paid run.
+PROVISION_TIMEOUT = 900.0
 
 
 def dataset_harbor_env(dataset_key: str) -> str:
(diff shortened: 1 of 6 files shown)


macanderson:
## Review: changes requested — two defects reproduced with deterministic, network-free probes

The design is more disciplined than the diff size suggests, and I want to name what it gets right before the defects. Provisioning pins **exactly** (`harbor=={version}` at `harbor.py:449`), never `>=` or latest. The resolved version and binary land in the match provenance record before the first trial, and each seat's notes carry the resolution origin (pinned / PATH / discovered / installed), which closes the #2098-shaped hole where provenance could name an apparatus that never graded. One resolution per match, not per seat.

The brew-0.6.1 safety question comes out clean, which was the axis I expected to reject on. The diff uses `uv venv --relocatable` + `uv pip install --python <venv>` into `~/.arenabench/harbor/<version>/` — no `~/.local/bin` shim, no PATH mutation, no takeover of the `harbor` name anywhere. Every `rmtree` target is confined under `managed_root()`, nothing touches `/opt/homebrew`, and no-floor datasets (TB2.1) still resolve PATH-only. Provisioning is genuinely last-resort, opt-out-able, and once-per-version-per-machine.

### Blocking 1 — the PID-only staging name, `harbor.py:473-474`

`server.py:925` is a `ThreadingHTTPServer` and `runner.start` runs on request threads, so two same-dataset provisions in one process share a PID. `staging = target.with_name(f"{target.name}.staging.{os.getpid()}")` plus an unconditional `shutil.rmtree(staging)` means thread B's first act deletes thread A's build. My event-controlled repro produced, in a single run: **A fails** ("installed harbor into … does not report a version"), **B fails** (writing into a path renamed away underneath it), and **`target` published half-built** with `bin/harbor` absent.

Worse, the second `os.rename` at `harbor.py:485` is unwrapped — a raw `OSError` there escapes both `resolve_for_dataset` (catches only `HarborUnavailableError`, :568) and `runner.start` (catches only the two harbor types, `runner.py:655`), so it surfaces as an unclassified crash at the start of a paid run. The "No lock file" note in the description covers processes but not threads, and the threaded server is the shipping shape.

**Fix:** `staging = Path(tempfile.mkdtemp(dir=target.parent, prefix=f"{target.name}.staging."))`, drop the pre-`rmtree`, and re-verify `ready()` (or wrap) after the second rename. Roughly four lines; the rename-wins design is preserved.

### Blocking 2 — stale staging directories poison resolution, `harbor.py:393`

`finally` does not run on SIGKILL or a default SIGTERM, and this rig has documented foreign SIGTERMs, so `0.20.0.staging.<pid>/` survives a kill. `parse_version("0.20.0.staging.99999")` returns `(0,20,0)` and `_discovered` iterates **all** children of `managed_root()`. My probe had `resolve_for_dataset` **select the staging directory's binary** and report it as "installed by arenabench". Validation is `--version`-only, so a half-built venv whose console script answers `--version` would grade a paid run.

**Fix:** admit only children matching `^[0-9]+(\.[0-9]+)*$` in `_discovered`, GC dead-PID staging dirs at provision entry, and add a witness that a staging-named directory is never selected.

### Blocking 3 — the new tests fail under the rig's own launch environment

The six hermetic tests pass at head (6/6, no network). With ambient `ARENABENCH_HARBOR=/opt/homebrew/bin/harbor` set — which the bench rig's server launch env does set — **5/6 fail spuriously**, because `_pinned` wins over the monkeypatched PATH. Add an autouse fixture (or per-test `monkeypatch.delenv`) scrubbing `ARENABENCH_HARBOR*` in the new class. Separately, `provision` / `_install` / `_discovered`-versus-debris have **zero** tests — which is exactly where both defects above live.

On complexity: I weighed whether the candidate surface (pin → PATH → managed root → two repo globs → provision) is a heuristic thicket, and concluded it isn't. Each source is policy-ordered and justified, and the repo-glob rung is the case that motivated the PR — a minimal "explicit env or one location" design would have refused on the very machine this fixes. Keep the design; land the three fixes.

macanderson:
Filed **#2313** as the handoff for the three blocking items above (PID-shared staging, stale staging directories selected as installed harbors, and the env-scrubbed tests), so the work is tracked whether it lands on this branch or a successor.

macanderson:
This merged (f3283f9a) before the three fixes landed, so they are now live: the PID-shared staging directory (`harbor.py:473-474`), the unwrapped second rename (`:485`), and `_discovered` selecting stale `*.staging.<pid>` directories as installed harbors (`:393`). All three are tracked in **#2313** with the reproductions and the exact fixes. The tests in this PR also fail spuriously under the rig's own launch env (`ARENABENCH_HARBOR` set), which is in the same issue.

To be clear about what is good here, since the issue only lists defects: the brew 0.6.1 is genuinely untouched (`uv venv --relocatable` + `uv pip install --python <venv>`, no PATH shim, every `rmtree` confined under `managed_root()`), provisioning pins exactly rather than floating, and the resolved version and path reach match provenance before the first trial.

Issue #2224, opened 2026-08-08
brand: the kit has no generator for PWA icons, wallpapers or spinner GIFs


macanderson:
Found while recoloring the brand kit to the Nebula palette (branch `brand/nebula-palette`).

## Problem

Of the ~50 binary assets in `docs/brand/`, only the 10 social images have a
committed generator (`docs/brand/social/build_social.py`). These have none:

- `docs/brand/pwa/` — 8 PNG icons + `favicon.ico`
- `docs/brand/wallpapers/{desktop,phone}/` — 12 PNGs, 4K/5K/6K, dark + light
- `docs/brand/spinners/*.gif` — 4 animated GIFs
- `website/public/icons/` — 7 mirrored PNGs
- `website/src/app/{favicon.ico,apple-icon.png}`

So a palette change cannot rebuild them from source. The nebula recolor worked
around this by remapping them **pixel-wise** (recovering a blend factor from
luminance and repainting toward the new sweep) — see the scripts described in
the PR. That preserved composition exactly and was the right call for a
colour-only change, but it is not a substitute for a generator: it cannot
change geometry, padding, or the maskable safe zone, and it silently depends on
each source being a clean two-tone image.

## Definition of done

A `docs/brand/build.py` (or Makefile target) that regenerates every binary in
`docs/brand/` from the SVGs + tokens, with `--check` verifying the committed
files match. `rsvg-convert` and Pillow are both already in use.

## Notes

- `docs/brand/pwa/safari-pinned-tab.svg` must stay `#000000` (Safari mask).
- The maskable icons use a much larger safe-zone inset than the plain ones;
  measure them from the committed files before regenerating.
- `website/public/brand/*.svg` must stay byte-identical to
  `docs/brand/logo/svg/*.svg` — `website/src/components/brand.tsx` says so.

macanderson:
PR #2271 adds \`docs/brand/package.json\` + \`pnpm generate-assets\` — a standalone pnpm tool (\`cd docs/brand && pnpm install && pnpm generate-assets\`) driven by a new \`brand-tokens.toml\`, covering \`pwa/\`, \`wallpapers/\`, and \`spinners/*.gif\` plus their \`website/\` mirrors, with \`pnpm check-assets\` as the \`--check\` gate this issue asked for.

[...]

PR #2271, opened 2026-08-08
feat(brand): add pnpm generate-assets pipeline for pwa/wallpapers/spinner GIFs

macanderson:
## What & why

`docs/brand/pwa`, `docs/brand/wallpapers`, and `docs/brand/spinners/*.gif` had no committed generator — a palette change meant hand-editing PNG/GIF pixels (the workaround #2224 documents from the Nebula recolor). This adds a standalone pnpm tool:

```
cd docs/brand
pnpm install
pnpm generate-assets   # recolor pwa/, wallpapers/, spinners/*.gif + website/ mirrors, in place
pnpm check-assets      # verify committed files match brand-tokens.toml; exits 1, writes nothing
```

`brand-tokens.toml` is the one file that names color — the nebula gradient (violet→cyan) plus a dark and a light `{bg, fg}`. Nothing about font or icon geometry is touched: **this recolors, it does not recompose.** None of the three families has a committed vector/composition source (only their rendered pixels were ever checked in), so the pipeline decomposes each committed pixel into a `{bg, fg, violet, cyan}` basis and re-renders it under the new tokens at the exact same position/opacity/frame — see `scripts/lib/recolor.mjs`'s module doc for the how and why, and `docs/brand/README.md`'s new "generated assets" section for the user-facing version. `spinners/*.svg` are real committed sources, so those recolor by exact hex substitution instead of decomposition.

This PR ships with `brand-tokens.toml` set to the currently-shipped palette, so it's a pure tooling addition — zero binaries change (verified: `git status` is clean after a full `generate-assets` run from this branch).

Refs #2224

**New dependencies** (`docs/brand/package.json`, a standalone package — not part of any workspace): `pngjs`/`sharp` for PNG decode/encode (pngjs's own encoder ran ~3x larger than the committed files on this kit's art; sharp's is what actually ships), `omggif` for GIF frame decode/encode, `smol-toml` for the config file.

## The witness

- [x] No witness needed (pure refactor / docs / CI) — because: this is new tooling with no prior behavior to regress from. Correctness is instead demonstrated by: an identity round-trip (recoloring from the fixed origin palette back to itself reproduces every committed file byte-for-byte — proven by a clean `git status` after running with the shipped tokens), a genuine non-identity palette swap verified visually across all three families (icon, wallpaper, spinner GIF) and cross-process byte-determinism, and `scripts/lib/recolor.test.mjs` (`pnpm test`) covering the recolor models directly.

## The gate

- [x] No Rust touched — `docs/**` + `website/public|src/app` binaries/config only
- [x] Docs updated (`docs/brand/README.md`'s new "generated assets" section)
- [x] `Refs #2224` appears above and as a commit trailer

## Nothing left behind

- [x] There is nothing new: everything I noticed is fixed here, or already tracked by #2224 itself, which stays **open** — this PR closes the "no generator at all" half of it but not the "can't change geometry/padding/safe-zone" half (see the README section's explicit scope boundary). No duplicate issue filed since #2224 already names that gap precisely.

## Ground-rule check

- [x] No new outbound network calls (all four new deps are build-time/local; the generator reads/writes local files only)
- [x] No I/O added to `stella-core` — this doesn't touch the Rust workspace at all

- **Real bug found and fixed mid-session**: the recolor engine's (g, t) fit is a discrete grid search, not a perfect identity function pixel-for-pixel — re-decomposing an already-recolored file could drift by a few bytes. Fixed by short-circuiting to a verbatim byte passthrough whenever old/new palettes are equal (`palettesEqual` in `recolor.mjs`), which is also what makes `--check` exact rather than approximate. Covered by a test.
- **PWA icon geometry (padding, maskable safe-zone inset) and wallpaper composition (glow, starfield, grid) are still hand-authored** — this PR intentionally does not attempt to reverse-engineer and recompose them from scratch, since a wrong guess there would silently drift from the committed art with no source to catch it. That's the part of #2224 left open.

## Summary by Sourcery

Introduce a configurable brand asset recoloring pipeline for PWA icons, wallpapers, spinner GIFs, and their website mirrors, driven by a single brand-tokens file and backed by reusable scripts and tests.

New Features:
- Add pnpm generate-assets and check-assets commands to recolor PWA icons, wallpapers, spinner GIFs, and propagate changes to website mirrors based on brand-tokens.toml.
- Introduce brand-tokens.toml as the single source of truth for nebula gradient and dark/light background/foreground colors used by generated brand assets.

Enhancements:
- Add modular brand generation scripts for PWA icons, wallpapers, spinner GIFs, and website asset mirroring, including shared recolor, PNG, GIF, ICO, token, and origin-palette utilities.
- Document the new generated assets workflow in docs/brand/README.md, clarifying scope (color-only changes) and limits around geometry and composition.

Build:
- Create a standalone docs/brand package.json and pnpm-lock.yaml with Node tooling and dependencies for image processing and TOML parsing.

Tests:
- Add node test coverage for the recolor engine, palette utilities, PNG recolor identity behavior, and origin palette consistency.


macanderson:
Closing — the palette this pipeline was built to generate no longer exists in the tree.

The Nebula recolor (#2226) and its three follow-ups (#2256, #2258, #2261) are reverted in **#2309**, which returns every branded surface to brand kit v1.0, the comet identity (Phosphor Gold `#FFB000` on Ink `#0B0B0C`). The 22 logo PNGs, 20 PWA icons, 12 wallpapers and 4 spinner GIFs this pipeline would regenerate are back to their pre-Nebula bytes, restored directly by the revert — so there is nothing left for `pnpm generate-assets` to produce.

Its tracking issue **#2224** is closed by #2309 for the same reason.

**Nothing is lost.** The branch `worktree-brand-asset-generator` is left in place, so `docs/brand/scripts/` — the recolor engine, its test, and the PWA / wallpaper / spinner / website-mirror families — remains recoverable if the kit ever wants a real generator again. Reopen this PR or cherry-pick from the branch; it would need rebasing onto the comet palette in `docs/brand/brand-tokens.toml` first.

macanderson:
## Review: changes requested — the generator's read path is the wrong architecture, not a doc gap

The Vercel bot's flag on `docs/brand/README.md:104` is real, and tracing it end-to-end shows it is a symptom rather than the defect. Reproducing the documented workflow:

1. Edit `brand-tokens.toml`, run `pnpm generate-assets`. `palettesEqual(ORIGIN, tokens)` is false, the real decompose runs, and the binaries on disk are correctly recolored. Commit, per the README.
2. `pnpm check-assets` now decomposes the **new**-palette pixels against the **old** anchors in `scripts/lib/origin-palette.mjs`. `fitBgGradient` is a total best-fit, so it always "succeeds" — with garbage. Every correct file is reported out of date, exit 1.
3. The tool's own error message (`generate-assets.mjs:93-95`) prescribes "Run `pnpm generate-assets` (no --check) to update them." Obeying it overwrites the correct binaries **in place** with the mis-decomposed output, exits 0, and prints "✓ wrote N file(s)".

After step 3 the only copy of the real art is git history. That is a tool whose own error message instructs a maintainer to destroy unrecoverable assets, and reports success while doing it.

A README step naming `origin-palette.mjs` does not fix this, for four independent reasons:

- Nothing enforces the origin sync. The invariant is stated at `origin-palette.mjs:9-10` ("matches brand-tokens.toml's shipped defaults byte for byte") and tested nowhere — the "origin palette consistency" test (`recolor.test.mjs:87-94`) only checks that triples have length 3.
- Even if enforced, `ORIGIN_HEX`'s correctness is unverifiable from repo state. The ground truth — which palette these pixels are actually in — exists only in the pixels, and the fit is total, so a typo'd origin succeeds silently and forever after.
- Because the design requires `origin == tokens` at every green commit, the `palettesEqual` short-circuit (`png.mjs:26-27`, `gif.mjs:17-18`) makes `check-assets` **a tautology in every committed state**: it reads each file and compares it to itself. `README.md:89`'s claim that it verifies committed files match the tokens can never hold — hand-recolor a wallpaper and the check stays green.
- The one state where the check isn't vacuous is also the state that shells out to a floating brew `oxipng` (`png.mjs:33`), whose output bytes vary by version.

Committed binaries as the sole source of truth, mutated destructively, with a hand-synced shadow constant as the only record of their provenance — that is the shape this repo is trying to stop building on. Notably #2224's own definition of done already names the right architecture: regenerate every binary **from the SVGs + tokens**.

### What would make this a merge

Freeze today's binaries as canonical masters (`docs/brand/sources/`, a one-time copy), freeze `ORIGIN_HEX` permanently as those masters' palette — it becomes a description of immutable files and is never hand-synced again — and make `generate-assets` always read masters → write the live paths. Everything else survives unchanged: `recolor.mjs`, the family modules, `tokens.mjs`, the mirror table, the license diligence. Only the read path moves. The pipeline then becomes idempotent by construction (the second-run hazard cannot exist), and `check-assets` genuinely verifies `live == f(masters, tokens)` in every state, including catching hand-edits.

### Also blocking

- **`check-assets` is wired into nothing** — no workflow, no Makefile step, no gate. Drift returns silently, which is the exact failure this check exists to prevent. Wire it into `docs-guards` (the steady-state path never spawns oxipng, so it needs only Node + pnpm) or file the handoff issue.
- **GIF structural assumptions are unguarded.** `gif.mjs:26-30` allocates a fresh canvas per frame and `addFrame` passes no `transparent` index — silently wrong for partial-rect frames, accumulate disposal, or transparency. All four committed GIFs happen to be full-canvas, opaque, non-interlaced, so the code is correct by luck of the export settings. Throw on partial rect/transparency rather than relying on that. `gif.mjs:39` also hardcodes `loop: 0` instead of preserving the source's loop count. There are zero GIF tests.
- **The description misdescribes the dependency set.** It says `pngjs`/`sharp` and "sharp's is what actually ships", but `package.json` ships no sharp — it was reverted for libvips's LGPL and replaced by external `oxipng` (`png.mjs:8-16` tells the true story). The description is the reviewable record.
- `package.json` declares `engines: ">=20"`, but `node --test 'scripts/**/*.test.mjs'` needs Node 21+ for glob arguments; on Node 20 `pnpm test` fails.

The engine, the license diligence, and the identity-passthrough insight are good work and should survive the restructure intact. Placement and hygiene are clean (exact-pinned deps, committed lockfile, `node_modules` correctly ignored, nothing tracked is polluted).

macanderson:
Filed **#2312** as the handoff for the restructure (frozen masters + frozen origin palette + one-way generation), along with the CI wiring, the GIF structural guards, the oxipng pin and the Node floor. That way the work is tracked whether it lands on this branch or a successor.

Issue #2281, opened 2026-08-08
arenabench: the premature-complete monitor rule conflates a 402-terminated trial with an agent that gave up


macanderson:
## Problem

arenabench's `premature-complete` monitor rule fires when the agent declared complete after few steps AND the verifier resolved False. That predicate cannot distinguish "the agent gave up early" (a capability signal) from "the agent was forcibly terminated by a terminal provider error mid-trial" (an operational abort — e.g. an OpenRouter HTTP 402 when credit runs out, the failure mode that corrupted 13/20 trials in the first `opus5-h2h-20task` attempt on 2026-08-07). Both shapes look identical to the rule, so an aborted trial gets narrated as an agent defect.

Related but distinct: #2220 (zero-telemetry trials scored as losses — the scoring side of the same family). This issue is about the monitor/narration side.

## Files

- `arenabench/arenabench/` — the monitor rules (search for `premature-complete`)
- Reference traces: `~/.arenabench/matches/` — the 2026-08-07 h2h attempt's 402-corrupted trials

## Done

The rule (or its consumer) checks the trial's late error / terminal event for provider-payment and auth-terminal signatures first, and labels those trials as operational aborts rather than premature completion. A unit test feeds both shapes and asserts the two distinct labels.

macanderson:
Closing as **not planned** — a maintainer backlog decision, not a claim that this is fixed.

The benchmarking and arena apparatus is no longer being carried as tracked work. This issue was verified as still live against the tree; it is being retired because the project is not pursuing this area, not because the defect went away. Nothing here touches any recorded benchmark number.

Reopen it if the work restarts — the problem statement and reproduction above still stand.

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #2280, opened 2026-08-08
arenabench: match 958b92296de1 killed by an unattributable SIGTERM at ~78min — runner must log signal receipt and shutdown cause


macanderson:
## Problem

Match 958b92296de1 (`opus5-stella-only-20task`) was killed at ~78 minutes by a SIGTERM nobody claims. Harbor's `_handle_sigterm` (`harbor/cli/jobs.py:282`) raised KeyboardInterrupt mid-trial-8 (`cobol-modernization__D6RyKww` cancelled — its `exception.txt` holds the traceback), 12 of 20 tasks never started, and `result.json.finished_at` is null. The launcher log (`opus5-stella-only-20task-run.log`) shows no watchdog, no timeout, no balance-guard action — it simply stops at 78.1m. The operator states they did not kill it.

Candidates: another session's cleanup sweeping process groups, macOS resource handling, a stray `kill` from a concurrent script on this shared box. Indistinguishable after the fact because nothing logs signal receipt or sender.

## Files

- `arenabench/arenabench/` — runner signal handling / logging
- Evidence: `~/.arenabench/matches/958b92296de1/` (`exception.txt`, match log tail), `~/.arenabench/opus5-stella-only-20task-run.log`

## Done

The runner logs signal receipt (signal number, plus sender pid/uid where obtainable) and writes a one-line shutdown-cause record into the match directory, so the next mystery SIGTERM is attributable from the artifacts alone.

macanderson:
Closing as **not planned** — a maintainer backlog decision, not a claim that this is fixed.

The benchmarking and arena apparatus is no longer being carried as tracked work. This issue was verified as still live against the tree; it is being retired because the project is not pursuing this area, not because the defect went away. Nothing here touches any recorded benchmark number.

Reopen it if the work restarts — the problem statement and reproduction above still stand.

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #2279, opened 2026-08-08
graph: watcher liveness in bench containers is unproven, and the journal has no graph-state event — graph_query went unused in all 8 trials


macanderson:
## Problem

Trace census of match 958b92296de1 (8 trials, opus-5 xhigh full pipeline): the worker called `graph_query` **zero times in all eight trials**, and `project_overview` 0–3 times (model discretion). The blank-start half is fixed by the orientation-fallback PR (branch `fix/orientation-fallback-and-cd-word-split`), but two follow-ups remain:

1. **Watcher liveness in containers is unproven.** Bench workspaces gain their real code mid-run (extracting a tarball, generating files — `sqlite-with-gcov` unpacks a full C tree). Whether the `notify`/inotify watcher actually indexes those files inside the Docker containers has no trial evidence either way: in `cancel-async-tasks` the only `project_overview` call ran ~12 minutes **before** the first `.py` write, so its `files: 0` proves nothing about the watcher. If the watcher is dead in-container, the graph stays empty for the whole run no matter what `stella init` does.
2. **No telemetry says what the graph held.** `stella-events.jsonl` has no event for "graph indexed N files / M symbols at ts" — tonight's only evidence was a stdout line (`✓ code graph: 0 symbols, 0 imports across 0 files` in every trial). A graph-state event (or a field on an existing event) would make this auditable per trial from the journal, which is ground truth.

## Files

- `crates/stella-cli/src/agent/graph.rs` — `spawn_session_graph` (build + watcher mount)
- `crates/stella-graph/src/watch.rs` — the live watcher
- Traces: `~/.arenabench/matches/958b92296de1/jobs/958b92296de1-stella-opus-5-full-pipeline-xhigh/*/agent/`

## Verify

In-container repro: start a session in an empty dir, `tar xf` a small C tree via the bash tool, wait past the debounce, call `project_overview` — `index.files` must be > 0. Runs as an integration test against a local Docker env or as a scripted bench smoke.

## Done

Either a witness that mid-run-created files become queryable in a container run, or the defect localized and fixed; plus the graph-state journal event so the next audit reads it from `stella-events.jsonl` instead of stdout.

Issue #2278, opened 2026-08-08
core: the task deadline is purely reactive — a paid trial lost by 1.6s at step 77 with no anticipatory wrap-up


macanderson:
## Problem

Match 958b92296de1: trial `sqlite-with-gcov__egbgJmd` ended with

> task deadline exceeded by 1.6s — stopping with partial work

at step 77, reward 0.0 — a 14-minute trial lost to its own internal deadline by **1.6 seconds**. `regex-log__De8GTY3` hit the same stop at +10.1s over and still scored 1.0 (its work was already sufficient). The deadline check is purely reactive: it notices the crossing at the next safe boundary and stops. Nothing anticipates — e.g. "remaining wall clock < a typical step's duration, wrap up now and commit what exists" — and the margin between Stella's internal deadline and the harness's hard kill (900s here; agent execution ran ~846s) is not visible in the trace (#2240).

Not provably a bug — the alternative to the stop is a harness SIGKILL with nothing collected — but the tuning question is real and now has two priced data points on opposite sides.

## Files

- `crates/stella-core/src/driver.rs` / `crates/stella-core/src/driver/settlement.rs` — where the deadline is consulted between steps
- Related: #2238 (raw stage calls never meter the clock), #2240 (journal cannot show deadline arming), #2242 (the deadline was only recently given a value at all)

## Done

A maintainer decision recorded on (a) the margin the bench adapter passes vs the harness limit, and (b) whether the step loop should begin an anticipatory wrap-up when remaining time < a step-duration estimate — with a unit/property test if (b) is built. Not to be tuned silently.

PR #2282, opened 2026-08-08
fix(stella-tools): never leave the worker blind or falsely warned about its own workspace

macanderson:
## What

Two defects in what the worker is told about its own workspace, both found by tracing every trial of bench match 958b92296de1 (2026-08-08, opus-5 xhigh full pipeline):

1. **`shell_words` never split unquoted operators off a word, contradicting its own doc.** `cd /app; ls` parsed its cd target as `/app;`, which cannot resolve inside session root `/app`, so the agent was told — twice, in a paid trial, while standing in its own session root — that its work was "invisible to the file tools, to this turn's diff, and to verification". The word splitter now emits unquoted `;`, `&`, `|`, `&&`, `||` as their own words (including when attached), backslash-escapes are preserved outside single quotes, and `cd_escape_target` skips operator words (a bare `cd && make` goes to `$HOME`, unresolvable here — skipped like `~`, "better a missed note than a wrong one"). This also repairs the grep-boundary scan, which shares the splitter.

2. **The system-prompt orientation block silently vanished whenever the code graph was empty** — the steady state on bench workspaces (empty dirs, tarballs under deny-listed `vendor/`, grammars we don't carry). All eight trial worker prompts carried no project map, and the worker called `graph_query` zero times all night. `render_orientation_block` now degrades to one sorted, bounded `read_dir` top-level listing instead of `None`, so the worker starts oriented on any non-empty workspace. The read-only contract is unchanged: it still never builds an index at prompt-assembly time, and the block stays byte-stable for a given tree state (sorted, bounded by `MAX_TOP_LEVEL_DIRS`), preserving the prompt-cache invariant (AGENTS.md invariant #7).

## Witness

- `an_attached_separator_does_not_fake_a_cd_escape` — the exact false-positive shape from the paid trial.
- `an_empty_index_falls_back_to_the_top_level_listing` — the exact bench shape: `stella init` built the index, nothing was indexable, the prompt must not go blank.
- Plus `shell_words_splits_operators_attached_to_a_word` and `the_top_level_listing_is_bounded`.

Verified the flip the artisanal way: with only the implementation hunks reverted, all five fail; with them restored, `cargo test -p stella-tools` is fully green. Clippy `--all-targets -D warnings` and rustdoc `-D warnings` clean.

## Reshaped test (deleted-test guard)

`orientation_block_is_none_without_an_index_and_never_builds_one` is renamed and reshaped to `orientation_block_without_an_index_lists_the_top_level_and_never_builds_one`: its "returns None" half asserted exactly the behavior this PR removes; its "never builds an index" half is kept and still asserted, and the None case now lives where it belongs (an empty workspace).

## Residue

Filed from the same trace audit: #2277 (triage reasoning=off burned its whole token cap), #2278 (reactive-only task deadline lost a trial by 1.6s), #2279 (watcher liveness in containers unproven; no graph-state journal event), #2280 (unattributable SIGTERM killed the match), #2281 (premature-complete monitor conflates 402 aborts with giving up).

## Summary by Sourcery

Ensure the worker always has an accurate, non-blank view of its workspace in prompts and shell-command analysis.

Bug Fixes:
- Fix shell command word splitting so unquoted operators are separated from adjacent words, preventing mis-detection of cd targets and workspace escapes.
- Treat operator-only cd invocations as non-escaping, avoiding false workspace-escape warnings.

Enhancements:
- Make the orientation block fall back to a bounded, sorted top-level directory listing when the code graph index is absent or empty, instead of omitting the project map.
- Exclude hidden entries from the fallback listing and bound its size to maintain deterministic, cache-friendly prompts.

Tests:
- Add tests covering operator splitting in shell_words and cd escape detection for attached separators and bare cd before operators.
- Add tests ensuring the orientation block falls back to a top-level listing when the index is empty and that the listing respects bounding.

### DIFF
--- a/crates/stella-cli/src/agent/prompt.rs
+++ b/crates/stella-cli/src/agent/prompt.rs
@@ -287,10 +287,14 @@ fn append_project_scripts(prompt: &mut String, workspace_root: &std::path::Path)
 /// useful on monorepos far past a few hundred files (issue #328). Read-only
 /// (`stella_tools::overview::render_orientation_block`
 /// opens an existing index and never builds one), so it adds nothing to
-/// first-response latency; it appears once the session's background index
-/// build has completed (or immediately when the workspace was pre-indexed,
-/// as the benchmark adapter does). Byte-stable for a given index state, so it
-/// keeps the cache-stable system prefix stable. The point is fewer
+/// first-response latency. It renders the graph-backed map once the session's
+/// background index build has completed (or immediately when the workspace was
+/// pre-indexed, as the benchmark adapter does), and falls back to a bounded
+/// top-level listing whenever the index is absent or empty — a worker is never
+/// left blind on an unindexable tree. Byte-stable for a given index and
+/// top-level tree state, so it keeps the cache-stable system prefix stable
+/// (the prompt is assembled once per session, so mid-session churn cannot
+/// reach the prefix). The point is fewer
 /// grep/glob/read_file discovery turns: the model starts knowing the shape of
 /// the code.
 fn append_project_orientation(prompt: &mut String, workspace_root: &std::path::Path) {
--- a/crates/stella-cli/src/agent/tools.rs
+++ b/crates/stella-cli/src/agent/tools.rs
@@ -86,9 +86,10 @@ impl RepoStructurePort for GitRepoStructure {
             _ => String::new(),
         };
         // Read-only like the system-prompt seam: renders only from an
-        // EXISTING index (never builds one inline), is bounded and
-        // byte-stable for a given index state, and simply appears once the
-        // session's background graph build has landed.
+        // EXISTING index (never builds one inline), is bounded, and is
+        // byte-stable for a given index and top-level tree state. The
+        // graph-backed map appears once the session's background graph build
+        // has landed; until then it falls back to a bounded top-level listing.
         compose_structure_summary(
             tree,
             stella_tools::overview::render_orientation_block(&self.root),
--- a/crates/stella-tools/src/bash.rs
+++ b/crates/stella-tools/src/bash.rs
@@ -72,8 +72,11 @@ const GREP_CMDS: &[&str] = &["grep", "egrep", "fgrep", "rg", "ripgrep", "ag"];
 /// pattern out of the common command shapes, returning each word already
 /// unquoted. NOT a shell parser: it respects `'…'` and `"…"` (so a pattern or
 /// path with spaces stays one word) and preserves backslash escapes like
-/// `\|` (so an alternation survives into [`is_symbol_shaped`]); bare
-/// operators (`&&`, `|`, `;`) come back as their own words to bound a scan.
+/// `\|` (so an alternation survives into [`is_symbol_shaped`]); unquoted
+/// operators (`&&`, `||`, `|`, `;`, `&`) come back as their own words to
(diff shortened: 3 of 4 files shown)


macanderson:
## Review: both fixes are correct — two repo-rule repairs done on this branch

Two real defects traced from a real paid trace, minimal fixes, four honest witnesses (I verified statically that each fails on main: the splitter trivially; the cd test's absolute-root assertion because `cd {root}; ls` yielded `{root};` and failed `Path::starts_with` component-wise; both orientation tests because `render_orientation_block` returned `None` in each shape). The reshaped test is named in the description exactly as the deleted-test guard demands.

**On the hand-rolled splitter — keeping it is the right call, and the description should have said why.** Neither `shlex` nor `shell-words` would have fixed this bug: both implement POSIX *field splitting for one command's arguments*, where `;` is a parser metacharacter rather than a field separator, so `shlex::split("cd /app; ls")` returns `["cd", "/app;", "ls"]` — the exact defect. Both also strip backslash escapes, which would break the `\|` preservation `is_symbol_shaped` depends on. Worth one sentence in the description since the dependency policy asks for that reasoning to be reviewable.

The escape-arm widening from `'\\' if in_double` to `'\\' if !in_single` is load-bearing, not incidental — without it `\;` outside quotes would hit the new operator arm and split. The resulting divergence from POSIX (real `sh` yields `a;b`, backslash stripped) fails toward the *miss* side: `resolve_within_root` is lexical containment with a walk-up, so a preserved backslash in an in-root path still resolves inside root. That matches the stated posture at `bash.rs:124-125`.

The orientation fallback is bounded (12 entries + "N more"), hidden entries filtered, and — importantly — `sort()` runs *before* `truncate()`, so the order is deterministic. Invariant 7 holds in the way that matters: the system prompt is assembled once per session/goal/subsession, so mid-session tree churn never re-enters the prefix.

### Fixed here (1c8c1bb7)

Two caller doc comments asserted the behavior this PR removes, and stale comments are bugs in review:

- `agent/prompt.rs:284-295` — "it appears once the session's background index build has completed" and "byte-stable for a given index state". Both false now: the section appears immediately on any non-empty workspace as a listing, and its stability domain is the index *and* the top-level tree.
- `agent/tools.rs:88-91` — same falsehood, same repair.

### Filed rather than fixed

**#2301** — the false-warning *class* survives in a different family. `cd_escape_target` scans all words via `windows(2)` with no notion of command position, so a `cd <path>` appearing as **data** still warns: heredoc bodies (`cat > setup.sh <<'EOF'` … `cd /workspace` … `EOF` — a very common agent shape), `echo cd /outside`, and commented text. Pre-existing, not introduced here, but this PR's own thesis is that a false warning to a paid agent is the defect class, and the class outlives the fix. The issue also folds in a drift nit: there are now three near-copies of "what is an operator word" in `bash.rs` (tokenizer arm, cd skip list, grep boundary at `:161` — which omits `&` while the cd list includes it). No behavioral bug today, but they will diverge.

Non-blocking observations: `2>&1` tokenizes as `["2>", "&", "1"]` since redirections aren't in the operator set (benign for both consumers, worth a word in the doc comment); `cd -- /outside` and `(cd /app; ls)` are missed (miss direction, fine); and the fallback text promising "once code exists, `project_overview` returns the graph-backed map" never materializes on a no-grammar tree — the very tree the fallback exists for — which may cost one wasted tool call.

Issue #2286, opened 2026-08-08
stella memory promote still mints legacy .md rules — promotion should emit a TOML context record


macanderson:
## Problem

Two write paths publish into `.stella/rules/`, one per era:

- `stella context keep` writes a **TOML context record** (`crates/stella-cli/src/context_cmd/review.rs` `run_keep`) — the ratified surface (ADR 0011/0012), refuses to overwrite, feeds the governance ledger.
- `stella memory promote` still writes a **legacy markdown rule** via `render_rule_markdown` (`crates/stella-core/src/rules.rs`; write side `crates/stella-cli/src/memory_cmd.rs` `promote_in`) — `description:` + `guard-*:` keys only.

So a citation-threshold promotion mints new files on the *legacy* surface forever, and those files carry no record identity, no lineage, no governance visibility. `docs/spec/adaptive-context/context-pr.md` § rollout item 2 ("Emit snake_case promotion frontmatter and `record_hash` from the promotion path") was written against the retired `rules::metadata` frontmatter layer (removed in the PR that references this issue) and should be re-pointed at TOML record emission — its premise no longer exists.

## Fix

`stella memory promote` emits a TOML context record (origin `inferred` or `observed` per ADR 0009's uniform 5-value `Origin`, enforcement `advisory` — an inferred directive may not start blocking, `RecordValidationError::InferredDirectiveBlocking`), through the same publication path `run_keep` uses (`publication_path` + `write_record`, no-overwrite preserved). Existing promoted `.md` rules keep loading untouched.

Update `docs/spec/adaptive-context/context-pr.md` rollout item 2 in the same PR to describe the TOML emission instead of the retired frontmatter.

## Verify

Witness: run the promote path in a fixture workspace; assert the published file is `.toml`, parses as a `context-record/v0.1` record set, round-trips through the loader, and appears in `stella context list`. The old assertion shape (metadata frontmatter in `.md`) must be absent.

## Done when

One write path family produces only the ratified surface; no new `.md` rule can be minted by any Stella code path (humans and legacy files unaffected).

Issue #2285, opened 2026-08-08
docs/spec/filesystem-layout.md omits the entire context-record surface under .stella/rules/


macanderson:
## Problem

`docs/spec/filesystem-layout.md` (§ the `.stella/rules/` entry, near lines 63-64) documents only `rules/<name>.md`. It predates the TOML context-record surface and now omits:

- `.stella/rules/*.toml` — context records, the ratified primary surface (ADR 0011/0012)
- `.stella/rules/governance.toml` — governance mode (#994)
- `.stella/rules/promotions.jsonl` — the hash-chained promotion ledger (#994)
- `.stella/proposals/*.toml` — ingest proposals awaiting `stella context keep`
- `.stella/private/context-sweep.json` and `.stella/private/context-decisions.jsonl` — the sweep cache and the fail-closed decision ledger (`crates/stella-cli/src/context_records.rs:47-50`)

AGENTS.md's `.stella/` table already describes the records/governance entries correctly; the spec lags it.

## Done when

The layout spec lists every path above with a one-line purpose each, matching AGENTS.md's table and `crates/stella-cli/src/context_records.rs` constants. Doc-only change; `make guards-fast` (doc-links) green.

## Definition of done

- [x] The layout spec lists `.stella/rules/*.toml`, `.stella/rules/governance.toml`,
      `.stella/rules/promotions.jsonl`, `.stella/proposals/*.toml`,
      `.stella/private/context-sweep.json`, and
      `.stella/private/context-decisions.jsonl`, each with a one-line purpose.
- [x] The entries match AGENTS.md's `.stella/` table and
      `crates/stella-cli/src/context_records.rs`'s path constants.
- [x] `make guards-fast` (doc-links) is green.

---

**Refiled with a checklist per the dod gate's own request** (it flagged this
issue as having no "Definition of done" section to verify against), and
**verified against PR #6335** (branch `fix/6295-docs-batch`):
`docs/spec/filesystem-layout.md`'s `.stella/rules/` entry now lists
`<name>.md`, `<name>.toml`, `governance.toml` and `promotions.jsonl`; the
`proposals/` entry now shows `<name>.toml`; and the `private/` block now
lists `context-sweep.json` and `context-decisions.jsonl` beside the existing
entries — matching `SWEEP_CACHE`, `DECISION_LEDGER`, `PROPOSALS_DIR`,
`RULES_DIR`, `PROMOTION_LEDGER` and `GOVERNANCE_FILE` in
[...]

Issue #2284, opened 2026-08-08
stella memory forget cannot suppress a TOML context record — two suppression mechanisms where spec 5.7 promises one


macanderson:
## Problem

`ContextSurface` (`crates/stella-store/src/forget.rs:66-92`) is the authoritative enumeration of suppression-eligible steering surfaces: `Memory`, `WorkspaceMemory`, `Rule`, `Skill`, `Agent`, `Command`, `Domain`. Its `Rule` variant is keyed to legacy `.stella/rules/<slug>.md` rule ids — **there is no `ContextRecord` variant**, so a TOML context record cannot be suppressed via `stella memory forget`. Retiring a record instead requires the promotion-ledger path (`PromotionAction::Retired`, `crates/stella-cli/src/memory/retirement.rs`).

`docs/spec/adaptive-context/adaptive-context.md` §5.7 describes **one** suppression mechanism; in practice there are two, with different capabilities and different audit trails, split along the legacy/new-surface line.

## Decide, then implement

Either:
- (a) `stella memory forget` on a record handle becomes sugar for a `Retired` promotion event (one audit trail, the ledger), or
- (b) `ContextSurface` gains a `ContextRecord` variant and tombstones apply to records too (two trails, one UX).

(a) looks right for a `regulated`-governance repo — the ledger is hash-chained and reviewable; a tombstone is not — but it changes `forget`'s semantics from "hide" to "retire", which needs a one-line UX decision.

## Verify

- `stella context list` / the loader must actually stop serving a forgotten/retired record next session.
- A forgotten record must not resurrect via `stella context keep` republication without a visible warning.

## Done when

One documented mechanism (spec §5.7 updated to match reality), a witness test proving a TOML record can be suppressed through the chosen path, and the loser path either removed or delegating to the winner.

macanderson:
CLI memory-forget vs TOML record suppression, feature completeness — closed in the 2026-08-12 mission-scope backlog sweep: doesn't serve Stella's 4 mission goals (beat Claude Code head-to-head at matched config, be cheaper, be more explainable, ship dataset generation for open-weight models) and isn't required to keep the product shippable/correct. Reopen with a mission-tie if that's wrong.

macanderson:
Reopened 2026-08-27 — **the sweep that closed this has been reversed.**

This was closed on 2026-08-12 as part of the mission-scope backlog sweep, on the
grounds that it did not serve the four mission goals. The maintainer has put it
[...]

Issue #2283, opened 2026-08-08
Epic: one control plane for steering content — bring memories under the context-record surface (ADR needed)


macanderson:
## Problem

The repository has one ratified control plane for *rules* — TOML context records at `.stella/rules/*.toml`, governed by `stella context keep/promote/govern` and the hash-chained promotion ledger (ADR 0011/0012, epic #897, #994) — but **memories live entirely outside it**:

- `.stella/memories/*.md` are written by the `save_memory` tool (`crates/stella-tools/src/memory.rs`) with no review, no lifecycle, no governance, no lineage, and no `stella context` visibility.
- Their delivery is the cached system prefix (`crates/stella-cli/src/agent/prompt.rs` `append_workspace_memories`, budget 16k chars), while records deliver via prefix (`must`/`should`) + volatile channel (`may`/`info`).
- Suppression is a separate mechanism: `ContextSurface::WorkspaceMemory` tombstones (`crates/stella-store/src/forget.rs`) vs. record retirement via `PromotionAction::Retired` (`crates/stella-cli/src/memory/retirement.rs`).
- The record taxonomy **already reserves the type**: `MemoryKind` exists in `crates/stella-core/src/context_record/kind.rs`, unused by any write path.

A 2026-08-08 census counted 11 steering subsystems; rules already converged onto records (markdown rules are projected onto records by `record_from_markdown`, `crates/stella-core/src/records/registry.rs`), but memories, skills, and the episodic DB did not. The maintainer's direction (2026-08-08 session) is: **one thing controls rules and memories.**

## What needs deciding (ADR required before code)

ADR 0011 drew the line "if the prose is a *field*, TOML; if the prose is the *document*, Markdown" — memories are prose documents, so convergence here means **one control plane, not necessarily one file format**:

1. Does a workspace memory become a context record (kind `memory`, Markdown body per ADR 0011's document rule, or TOML with a multiline statement)?
2. Does `save_memory` write a record (or a proposal that auto-activates at a confidence threshold, like inferred directives in `crates/stella-cli/src/memory/rules_mining.rs`)?
3. Do memory suppression tombstones fold into record retirement so spec §5.7's "one mechanism" claim becomes true (see companion issue on `ContextSurface`)?
4. Channel assignment: memories currently ride the byte-stable cached prefix (`docs/spec/adaptive-context/adaptive-context.md` §5.1 forbids per-turn variation there). A memory record presumably maps to the cached channel; confirm `force` semantics for memory kinds.

[...]

PR #2287, opened 2026-08-08
refactor(stella-core): retire the subsumed rules::metadata layer — the subsumption ADR 0009 ratified is now complete

macanderson:
## What & why

`crates/stella-core/README.md`'s gotcha claimed `rules::metadata` "still drives the live rules path," that the legacy and `context_record` type sets "coexist by decision — do not merge them," and that two subsumption edges were unratified. All three claims are stale:

- **Both edges were ratified in ADR 0009 on 2026-07-24** (`Informational → advisory`; uniform 5-value `Origin` with no per-family narrowing), and ADR 0009's follow-up explicitly asked for these comments to be removed once installment #470 landed. #470 is closed — this cleanup was already mandated, just never executed.
- **Nothing in production ever read the parsed metadata.** `Rule::metadata`, `Rule::metadata_errors`, and `render_rule_metadata` had zero non-test readers; the live path for `.md` and `.toml` rules alike projects onto records via `records::registry::record_from_markdown`, which deliberately ignores them (its own comment says the metadata parser "is deliberately not extended to reach here").

So this PR finishes the subsumption ADR 0001 promised ("subsumed and renamed into this taxonomy in Phase 1"): it deletes the parse-and-discard `rules/metadata.rs` module (549 lines net removed), drops the two dead `Rule` fields, and rewrites the stale prose in `context_record.rs`, `context_record/kind.rs`, and the crate README to state the ratified truth.

Behavioral contract preserved: legacy `.md` rule files that carry the retired frontmatter keys (`record_id`, `enforcement`, `origin`, timestamps, …) keep loading as ordinary prompt rules with the keys inert — same as before, since the parsed result was always discarded.

Refs #483, #470, #2286.

## The witness

- [x] Mostly a dead-code deletion + doc truth fix — behavior preservation is the claim, not a new behavior.

The one real risk — a legacy `.md` file with metadata frontmatter failing to load — is pinned by `directive_rule_fixture_still_loads_with_legacy_metadata_ignored` (`crates/stella-cli/src/rules.rs`), which loads the committed 11-field fixture and asserts the statement survives, the keys are inert, and no guard is manufactured. It replaces the retired `directive_rule_fixture_parses_with_context_metadata`.

**Deleted tests, named per the deleted-test guard:**
- `rule_from_file_parses_and_canonically_renders_context_metadata` — died with `rules/metadata.rs`; it tested the retired parser/renderer round-trip.
- `rule_from_file_keeps_invalid_metadata_distinguishable_for_linting` — same; the "future read-only linter" it reserved errors for was never built and is superseded by the TOML surface's real validation.
- `directive_rule_fixture_parses_with_context_metadata` (stella-cli) — replaced as above.
- Trimmed assertions (not deletions): `the_projected_rule_is_keyed_by_handle_so_a_block_is_citable` loses its `metadata.is_none()` assertion — the type system now enforces what it asserted.

## The gate

- [x] `cargo fmt --check`
- [x] `cargo test --workspace` — 128 suites, 0 failures
- [x] Docs updated where behavior/flags changed (crate README module table + gotcha, module docs)
- [x] CLA signed
- [x] No open issue is closed by this PR — `Refs` only (ADR 0009's tracking issues #483/#470 are already closed)

## Nothing left behind

- [x] Filed: #2283 (epic: one control plane — memories under the context-record surface, ADR needed), #2284 (`forget` cannot suppress a TOML record; spec §5.7 promises one mechanism), #2285 (`filesystem-layout.md` omits the record surface), #2286 (`stella memory promote` still mints legacy `.md` rules; also re-points `context-pr.md` rollout item 2, whose premise this PR retires)

## Ground-rule check

- [x] No cross-boundary type changes: `Rule` has no serde derives and never crosses a wire; the removed items were `stella-core` re-exports with no external consumers in the workspace

The ADRs (0008, 0009, 0011) still mention `rules::metadata` — deliberately untouched: they are decision records describing the state at decision time, not living docs. Living docs (`docs/spec/**`, `website/`) had zero references. The stale-comment cleanup was itself tracked prose inside ADR 0009 ("should be removed or updated when installment #470 lands"); this PR is that removal.

## Summary by Sourcery

Retire the unused legacy rules metadata layer and align documentation with the ratified subsumption decisions while preserving rule loading behavior.

Enhancements:
- Remove the `rules::metadata` module, associated `Rule` metadata fields, and related re-exports now that frontmatter metadata is no longer used in the live rules path.
- Clarify `context_record` documentation and enum semantics to reflect ADR 0009 resolutions and the retirement of the parallel metadata taxonomy.
- Tighten rules loading semantics so unknown or retired frontmatter keys remain inert while only rule-level defects cause file load failures.

Documentation:
- Update `stella-core` README and `context_record` module docs to describe the current record taxonomy, rules path projection, and the retired metadata frontmatter layer.

Tests:
- Replace metadata round-trip tests with a witness test ensuring legacy markdown rules with metadata frontmatter still load as ordinary prompt rules with inert keys, and clean up metadata-related assertions in existing tests.


macanderson:
Review outcome: the "nothing in production reads the parsed metadata" claim was verified by a workspace-wide census of `Rule.metadata` / `metadata_errors` readers (only tests, all handled in this diff), and the ADR 0009 citations check out — this is the cleanup the ADR itself mandated. One piece of residue this deletion creates: `Frontmatter.duplicate_keys` loses its only consumer (`metadata.rs:238`) and becomes a write-only field. Filed as #2303 rather than growing this PR's scope — a pure-deletion PR should stay one.

PR #2294, opened 2026-08-08
fix(stella-graph): quarantine a store whose corruption survives the open

### LINKED ISSUE #2293: graph: data-page corruption survives the open-time quarantine — codegraph.db sat dead for four days with no signal and no rebuild
## The report

2026-08-08, Mac: "code graph is broken". Field state in `~/Projects/stella/.stella/private/`: `codegraph.db` failed `PRAGMA integrity_check` with dozens of "Rowid … out of order" b-tree faults; main file mtime frozen at Aug 4 21:13 while the `-wal`/`-shm` sidecars carried the current morning's mtime with a **zero-byte WAL** — every session since Aug 4 opened the store, started catch-up, hit the corrupt pages, and wrote nothing.

## Root cause

`crates/stella-graph/src/store.rs::open` quarantines a damaged image only when the damage trips the open itself (e.g. an unreadable header → `SQLITE_NOTADB`). Corruption confined to data pages under an intact schema page passes `CREATE TABLE IF NOT EXISTS` untouched, so the open hands back a "working" store whose every scan then fails `SQLITE_CORRUPT`:

- `CodeGraph::mount` (`crates/stella-graph/src/graph.rs`) discards the catch-up result: `let _ = tokio::task::spawn_blocking(... index_tree ...)` — the error evaporates.
- The deck's GRAPH tab (`crates/stella-cli/src/agent.rs::graph_snapshot_focus`) swallows open/query errors with `.ok()?` — silently empty.
- The quarantine notice is an `eprintln!`, invisible under the deck's alternate screen (graph-state observability is #2279).

## Fix

`open_verified`: the writer's open runs `PRAGMA quick_check(1)` after migration and routes a bad verdict through the existing quarantine-and-rebuild arm (`quarantine_corrupt_store`). Proven against the actual corrupt field db: quick_check reports the rowid-order fault. The read connection in `CodeGraph::open` switches to `open_read` so the page walk is paid once. Witness test: `store::tests::a_store_corrupt_only_in_its_data_pages_is_quarantined_at_open` (fails on old code — the corrupt store is served and the first scan errors).

## Residue (not in the fix)

- Cause of the original corruption is undetermined (WAL mode + single writer should not produce rowid-out-of-order pages; suspects: a kill during checkpoint around Aug 4 21:13, or an external copy/restore of a live db). Nothing to act on without a second occurrence; the quarantined image is preserved at `.stella/private/codegraph.db.corrupt-20260808` for forensics.
- GRAPH tab's `.ok()?` remains fail-silent for *non-corrupt* errors; covered by the observability issue #2279 rather than duplicated here.

macanderson:
## What & why

Mac's workspace graph (`codegraph.db`) had rowid-out-of-order corruption in its data pages while the schema page stayed intact. That image opens and migrates cleanly — `CREATE TABLE IF NOT EXISTS` reads only the early pages — so `store::open`'s existing quarantine (built for unreadable-header damage) never fired, and the store was handed back "working" with every scan failing `SQLITE_CORRUPT`. Downstream, `CodeGraph::mount` discards the catch-up result and the deck's GRAPH tab swallows errors with `.ok()?`, so the graph sat dead for **four days**: sessions re-created the WAL sidecars each morning (zero-byte WAL — not one successful write) and told no one.

The writer's open now runs `PRAGMA quick_check(1)` after migrating (`open_verified`) and routes a bad verdict through the same quarantine-and-rebuild arm as an unopenable file — the graph is a rebuildable cache, and "damaged" is a property of the image, not of whichever pages the open happened to read. `quick_check` was verified against the actual corrupt field database before writing any code: it reports the exact rowid-order fault. The read connection in `CodeGraph::open` switches to `open_read`, so the page walk is paid once, by the connection that just proved the store (and `open_read` is documented as the reader that needs no DDL).

**Exemplar:** the crate's own header-corruption arm (`a_corrupt_store_is_quarantined_and_rebuilt_on_open`) and `stella-context`'s integrity posture (`ContextError`'s integrity-check variant: report loudly, never serve corrupt context) — this extends the same policy to damage the open path cannot see.

Closes #2293

## The witness

`store::tests::a_store_corrupt_only_in_its_data_pages_is_quarantined_at_open` builds a multi-page store, scrambles one mid-file page (header and schema untouched), and asserts `open` hands back a fresh store with the damaged image moved aside. Verified both directions locally: **FAILED** with the two source files reverted to `origin/main` (the corrupt store is served and the first scan errors), passes with the fix. Full `stella-graph` suite: 151 tests, 0 failures.

## The gate

- [x] `cargo fmt --check`
- [x] `cargo test -p stella-graph` — all targets green (workspace suite + clippy left to CI; this machine is kept clear of workspace builds per operator instruction)
- [x] Docs updated (`open` / `open_verified` / `open_migrated` doc comments; read-conn rationale in `CodeGraph::open`)
- [x] `Closes #2293` appears both above and as a commit trailer

## Nothing left behind

- No tests deleted.
- Original corruption cause undetermined; the field image is preserved at `.stella/private/codegraph.db.corrupt-20260808` for forensics (noted in #2293).
- Graph observability (mount discards catch-up errors; GRAPH tab `.ok()?`; quarantine notice is an `eprintln!` invisible under the alternate screen) is tracked as #2300 — this PR makes the failure self-healing, #2300 makes it visible. (An earlier revision of this description pointed at #2279, which is about watcher liveness and journal events — a different problem; #2300 is the handoff for these three silent channels.)

## Summary by Sourcery

Quarantine and rebuild stella-graph stores whose corruption is only detectable via a full integrity check, ensuring the code graph self-heals instead of silently serving corrupt data.

Bug Fixes:
- Detect data-page-only SQLite corruption in the workspace graph using a structural `PRAGMA quick_check` and route it through the existing quarantine-and-rebuild path.
- Ensure CodeGraph uses a verified writer connection and a lightweight reader connection, preventing the graph from remaining dead due to undetected store corruption.

Tests:
- Add a witness test that corrupts only mid-file data pages and asserts that `open` rebuilds the store and quarantines the damaged image.


macanderson:
Review note: the description's residue pointer previously named #2279, but #2279 tracks watcher liveness in bench containers and the missing graph-state journal event — not these three silent channels. The actual handoff for the mount catch-up discard (`agent/graph.rs:192`), the GRAPH tab `.ok()?` cascade (`agent.rs:1342-1360`), and the invisible `eprintln!` quarantine notice (`store.rs:274-278`) is now filed as #2300; the body has been corrected to point there. With that tracking in place, closing #2293 via this PR no longer orphans the residue.

Issue #737, opened 2026-07-26
Auto-created skills can silently overwrite a hand-edited skill file (spec §8 no-clobber does not hold)


macanderson:
Found while building Phase 3 (#714), which must preserve this guarantee. It does not currently hold.

Spec §8 lists **"never clobbering a hand-edited file"** as one of five guarantees the adaptive loop must preserve. The skills auto-creation path can destroy a user's hand-edited skill file today.

## Root cause

The no-clobber guard is a **list-membership test, not a filesystem test**.

`stella-core/src/skills.rs` `decide_auto_creation`:

```rust
let path = format!("{}/{}.md", target_dir.trim_end_matches('/'), candidate.name);
if existing_paths.iter.any(|p| p == &path) {
    return AutoCreateDecision::Skip { reason: AutoCreateSkip::FileExists { path } };
}
AutoCreateDecision::Create { path }
```

`existing_paths` is built in `stella-cli/src/memory/learning.rs` `auto_create_skills` from `self.load_skills` — the *loaded, filtered* skill list. The caller then does an unconditional `std::fs::write(&path, markdown)`.

So the decision is "is this path in the list of skills I successfully loaded and kept?", not "does this file exist?". Any file that exists on disk but is absent from that list gets overwritten without warning. `AutoCreateSkip::FileExists` is therefore misnamed — it never checks whether the file exists.

## Three ways a real file drops out of the list

`load_skills` → `load_workspace_skills_with_authority(root, self.include_workspace_skills)` in `stella-cli/src/memory/skills.rs`:

1. **`include_workspace_skills == false`** → `workspace_skills_dir` is passed as `String::new`, so no workspace skill is loaded. Meanwhile `auto_create_skills` computes its write target from `self.workspace_skills_dir` **unconditionally** — it never consults the authority flag. Mining writes into a directory it was just told it may not read.
2. **`filesystem_settings_disabled`** → returns `LoadedSkills::default`, an empty list.
3. **`retain_enabled(...)`** → a skill disabled from the SKILLS tab is dropped from the list, and the comment directly above it states *"its file stays on disk"*.

Path 3 is the most reachable and needs no unusual configuration.

## Reproduction (path 3)

1. Let the loop auto-create a skill, or hand-write `.stella/skills/<name>.md`.
2. Edit it substantively.
3. Disable it in the SKILLS tab. The file stays on disk, by design.
[...]

Issue #714, opened 2026-07-26
P1: Phase 3: the typed adaptive loop — ledger, proposals, governance (goal line)


macanderson:
**Phase 3 — the typed adaptive loop.** This is the goal line: the phase where "adaptive context is real". Ships behind `context.lifecycle.enabled` (default off).

**Observable:** Stella proposes durable context from evidence and explains why — "three separate tasks, here they are" — and Keep/Edit/Ignore is auditable afterward. The proposal that becomes a skill is reviewable *before* it lands, which is the part users cannot do today.

**Why third — and read this before starting:** this is a **migration, not a build** (spec §8). A live lexical loop already does this and users depend on it. It must not regress.

**Blocked on ratifying ADR 0010** — see #711, decisions 1 and 2.

### Deliverables

-  **1. The lifecycle ledger.** Append-only, immutable, hashed records for observations, proposals, and promotion events, born canonical per ADR 0010. No legacy rows are rewritten. New kinds have no legacy counterpart and therefore need no migration.
-  **2. Typed observation extraction**, replay-idempotent with cursors, secrets redacted before persistence, sourced from the evidence already captured — reflection lessons, citations, tool outcomes.
-  **3. Proposal induction over the existing miner.** Keep the clustering, the stable identity, and the thresholds. Add a durable, reviewable proposal record carrying its supporting observations, distinct-task count, and scored components. Distinct tasks, never raw events.
-  **4. Migrate the skills loop onto the typed path**, behavior-compatible. Existing skill files are untouched. The per-session cap, no-clobber, and two-surface tombstone filtering all survive — asserted by test, not by inspection.
-  **5. Wire the dead rules miner** through the same path. Its twin has shipped unwired since the shared clustering module was created to keep them aligned.
-  **6. Solo advisory governance.** Immutable promotion events; inferred directives start advisory and can never start blocking; no automatic sharing or scope widening; a re-proposal cooldown so a declined proposal does not return next turn.
-  **7. Review surface.** List proposals with their evidence; Keep, Edit, or Ignore; every outcome recorded as an immutable event.

### Gate

- Three distinct tasks produce one eligible proposal; thirty repetitions inside one task produce none.
- No inferred directive reaches blocking by any path.
- No sharing or scope widens without an explicit act.
- A tombstoned lesson still cannot return as a proposal or a skill.
- Existing skill files are byte-identical after migration.
- Keep/Edit/Ignore replay identically from the event log.
- Mining failure cannot fail the user's task.

[...]

PR #2295, opened 2026-08-08
feat(stella-cli): one write surface — promotion and mined rules publish TOML context records

### LINKED ISSUE #2286: stella memory promote still mints legacy .md rules — promotion should emit a TOML context record

macanderson:
## What & why

Two write paths published into `.stella/rules/`, one per era: `stella context keep` wrote the ratified TOML record surface (ADR 0011), while `stella memory promote` and the mined-rule writer (`rules_mining::write_rule`, reached from the learning loop's auto-activation and `stella proposals keep`) still minted legacy markdown rules — files with no record identity, no lineage, no governance visibility, on a surface whose parser is frozen by decision.

This PR removes the duplicate write system. All three call sites now publish through one shared path (`context_records::inferred_rule_record` + `write_record`):

- **One record shape.** Origin `inferred`, steering `should` (cached prefix, the channel the legacy rules used), no enforcement section — an inferred directive may not start blocking, and the miner's inferred guard stays stripped (#714's seam, still witnessed).
- **One lineage namespace.** `derive_set_id` (previously ingest-private) is shared, so promotions, mined rules, and ingest publications land under the same `ctx.<org>.<repo>.<slug>` lineages. The helper normalizes a dot-prefixed workspace name that would otherwise mint a malformed `ctx..…` lineage.
- **One clobber discipline.** `write_record` moved from `context_cmd/review.rs` to `context_records.rs` and now opens `create_new`: every caller's exists-check was advisory and racy, and this is the write that must not destroy a reviewed file. (`stella context keep` gets this hardening for free.)
- **One quality bar.** A statement that would fail `stella context validate`'s one-sentence check is refused at publication with an actionable reason, via `statement_reads_as_pasted` now exported from the validator — instead of a copied constant, or worse, publishing a record the workspace's own CI then flags. `write_rule` returns `Result<Option<PathBuf>>` so a refused statement is reported as *why*, never folded into "already exists."
- **The dead half is gone.** `render_rule_markdown` lost its last production caller and is deleted. Legacy `.md` rules keep loading unchanged, and a lesson that already minted one is skipped rather than double-published as a TOML twin.

Design lens (maintainer's directive): what context helps the next agent that finds this record? A promoted record now carries queryable provenance (`source_kind = "memory"`, `source_uri = "memory:<id>"` / `proposal:rule:<id>`), a stamped `record_id`/`record_hash` that verifies on load, and lives under the one lifecycle (`stella context list/explain`, retirement, governance) instead of as an anonymous markdown file.

Closes #2286
Refs #2283, #714, #737

## The witness

`promoted_record_round_trips_through_the_record_loader` (memory_cmd.rs) and the `.toml`-extension + record-loader assertions in `promotion_gate_and_rule_write_work_end_to_end` and `a_written_rule_file_is_prompt_only` fail on `main`, which writes markdown; here the published file loads through `load_context_file` with the stamped hash verifying and no enforcement section. The end-to-end test also witnesses the no-clobber idempotence surviving the conversion.

**Deleted tests, named per the deleted-test guard:**
- `render_rule_markdown_round_trips_through_rule_from_file` (stella-core `rules/tests.rs`) — died with the renderer.
- `render_rule_markdown_includes_an_inferred_guard` (same) — its guard-arming half is still covered by the loader-side guard tests and the cli guard fixture.
- `promoted_markdown_is_parsed_back_by_the_rules_engine` (stella-cli `memory_cmd.rs`) — replaced by `promoted_record_round_trips_through_the_record_loader`.

Rewritten in place (not deleted): `an_inferred_rule_is_never_blocking`, `a_written_rule_file_is_prompt_only`, `mined_rules_land_where_the_loader_reads`, `promotion_gate_and_rule_write_work_end_to_end`, plus the no-clobber test asserting `Ok(None)`.

## The gate

- [x] `cargo fmt --check`
- [x] `cargo clippy --workspace --all-targets -- -D warnings` (exit code checked directly)
- [x] `cargo test --workspace` — 128 suites, 0 failures
- [x] Docs updated: `stella memory` module doc + `promote` help text, `context-pr.md` rollout item 2 re-pointed from the retired frontmatter emission to the shipped TOML emission
- [x] CLA signed

## Nothing left behind

- [x] Filed previously and advanced here: #2283 (epic — this is the "one write surface" slice), #2284/#2285 unchanged. Behavior note reviewers should weigh: the learning loop no longer auto-activates a lesson whose text violates the one-sentence record shape — it stays a reviewable proposal and says so, where before it silently minted an unvalidatable `.md`. That is the intended direction (records are the quality bar), not an oversight.

## Ground-rule check

- [x] No I/O added to `stella-core` (the new export is a pure predicate); no new deps
- [x] No cross-boundary serde shape changed: `Record` serialization is untouched; only who writes it changed

Sequencing with #2287 (the `rules::metadata` retirement): independent branches, both from `main`; the only shared file is `stella-core/src/lib.rs` where they edit different lines of adjacent re-export blocks — expected to merge cleanly in either order. `RuleCandidate`, `decide_promotion`, and the mining pipeline are unchanged; only the final write changed shape.

## Summary by Sourcery

Unify rule promotion and mining so all inferred rules publish as TOML context records instead of markdown, sharing a single record lineage and no‑clobber discipline.

New Features:
- Publish promoted memories and mined rules as stamped TOML context records on the same surface used by `stella context keep`, with advisory `inferred` origin and provenance metadata.

Bug Fixes:
- Prevent inferred rule publication from overwriting existing reviewed or hand-edited files by enforcing `create_new` semantics across all writers.
- Ensure inferred rules cannot introduce blocking behavior by omitting enforcement data from published records.
- Reject overlong or multi-line rule statements at publication using the same one-sentence validation used by `stella context validate`, avoiding creation of records CI would immediately flag.

Enhancements:
- Share the workspace record-set identifier (`derive_set_id`) across ingest, promotion, and mining so all emitted records land under a consistent lineage namespace.
- Harden auto-activation in the learning loop so lessons that fail the record-shape check remain proposals with a surfaced reason instead of silently minting legacy markdown.
- Refactor rule mining and promotion paths to use the common context record publication helper, returning richer status for callers.
- Update documentation to describe TOML context records as the canonical promotion surface and align CLI help text with the unified write path.

Tests:
- Add and update tests to assert promoted and mined rules round-trip through the context record loader, publish `.toml` files, remain prompt-only, and preserve no-clobber semantics.
- Remove tests tied to the deprecated markdown renderer while keeping guard behavior covered via record-based tests and existing fixtures.

### DIFF
--- a/crates/stella-cli/src/context_cmd/review.rs
+++ b/crates/stella-cli/src/context_cmd/review.rs
@@ -34,11 +34,13 @@ use std::path::Path;
 
 use colored::Colorize;
 
-use stella_core::ingest::record::{ContextFile, Record, SharingScope};
+use stella_core::ingest::record::SharingScope;
 use stella_core::records::{Decision, DecisionEvent, decision};
 
 use super::{FoundProposal, read_proposals, resolve_candidate, verdict_label};
-use crate::context_records::{append_decision, now_rfc3339, publication_path, read_decisions};
+use crate::context_records::{
+    append_decision, now_rfc3339, publication_path, read_decisions, write_record,
+};
 
 /// `stella context review`.
 pub fn run_review(root: &Path, show_all: bool) -> Result<(), String> {
@@ -349,24 +351,6 @@ pub fn run_ignore(
 }
 
 /// Write one record as a published context file.
-fn write_record(path: &Path, set_id: &str, record: &Record) -> Result<(), String> {
-    if let Some(dir) = path.parent() {
-        std::fs::create_dir_all(dir)
-            .map_err(|e| format!("cannot create {}: {e}", dir.display()))?;
-    }
-    let file = ContextFile {
-        schema: stella_core::ingest::record::SCHEMA_TAG.to_string(),
-        set_id: set_id.to_string(),
-        ingest_run_id: None,
-        defaults: None,
-        records: vec![record.clone()],
-        proposals: Vec::new(),
-    };
-    let body =
-        toml::to_string_pretty(&file).map_err(|e| format!("cannot serialize the record: {e}"))?;
-    std::fs::write(path, body).map_err(|e| format!("cannot write {}: {e}", path.display()))
-}
-
 /// Who is deciding. A local username is enough for solo mode; team mode carries a
 /// real identity through Git authorship instead.
 pub(crate) fn actor() -> String {
--- a/crates/stella-cli/src/context_records.rs
+++ b/crates/stella-cli/src/context_records.rs
@@ -585,5 +585,122 @@ pub(crate) fn publication_path(
     Some(dir.join(format!("{lineage}.toml")))
 }
 
+/// Serialize `record` as a one-record context file at `path`, refusing to
+/// overwrite. `create_new` rather than a separate exists-check because every
+/// caller's pre-check is advisory and racy, and this is the write that must
+/// not destroy somebody's reviewed file (the posture #737 established for the
+/// mined-rule writer).
+pub(crate) fn write_record(
+    path: &Path,
+    set_id: &str,
+    record: &stella_core::ingest::record::Record,
+) -> Result<(), String> {
+    if let Some(dir) = path.parent() {
+        std::fs::create_dir_all(dir)
+            .map_err(|e| format!("cannot create {}: {e}", dir.display()))?;
+    }
+    let file = stella_core::ingest::record::ContextFile {
+        schema: stella_core::ingest::record::SCHEMA_TAG.to_string(),
+        set_id: set_id.to_string(),
+        ingest_run_id: None,
+        defaults: None,
+        records: vec![record.clone()],
+        proposals: Vec::new(),
+    };
+    let body =
+        toml::to_string_pretty(&file).map_err(|e| format!("cannot serialize the record: {e}"))?;
+    let mut out = std::fs::OpenOptions::new()
+        .write(true)
+        .create_new(true)
+        .open(path)
+        .map_err(|e| {
+            if e.kind() == std::io::ErrorKind::AlreadyExists {
+                format!(
+                    "{} already exists — refusing to overwrite a published record",
+                    path.display()
+                )
+            } else {
+                format!("cannot write {}: {e}", path.display())
+            }
+        })?;
+    use std::io::Write as _;
+    out.write_all(body.as_bytes())
+        .map_err(|e| format!("cannot write {}: {e}", path.display()))
+}
+
+/// Build and stamp the context record an inference-derived rule publishes —
+/// the shared shape behind `stella memory promote` and the mined-rule
+/// writers, so every code path that once minted a legacy `.md` rule emits
+/// the ratified TOML surface (ADR 0011) instead.
+///
+/// The record is advisory by construction: origin `inferred`, steering
+/// `should` (cached prefix, like the rules it replaces), no enforcement
+/// section — an inferred directive may not start blocking, and the guard a
+/// miner inferred is deliberately not carried over (#714's stripping seam).
+///
+/// The statement is refused when it would fail `stella context validate`'s
+/// one-sentence shape check: publishing a record the workspace's own CI then
+/// flags would turn a promotion into a red gate.
+pub(crate) fn inferred_rule_record(
+    set_id: &str,
+    suffix: &str,
+    statement: &str,
+    source_kind: &str,
+    source_uri: &str,
+) -> Result<stella_core::ingest::record::Record, String> {
+    use stella_core::context_record::{Origin, RecordStatus};
+    use stella_core::ingest::record as rec;
+    // A dot-prefixed workspace directory (or any set id with stray dots) would
+    // put an empty segment in the lineage (`ctx..foo.bar`); normalize rather
+    // than publish a malformed lineage.
+    let set_id = set_id.trim_matches('.');
+    let set_id = if set_id.is_empty() {
+        "workspace"
+    } else {
+        set_id
+    };
+    let statement = statement.trim();
+    if statement.is_empty() {
+        return Err("the rule statement is empty".to_string());
+    }
+    if stella_core::records::validate::statement_reads_as_pasted(statement) {
+        return Err(format!(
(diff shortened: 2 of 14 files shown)


macanderson:
## Review: changes requested — the design is right, three things to fix first

Consolidating one write surface is the correct call and most of the witness is real. The `.toml`-extension and record-loader assertions in `promotion_gate_and_rule_write_work_end_to_end` and `a_written_rule_file_is_prompt_only` genuinely discriminate (main writes `.md`), `create_new` in `write_record` is correct hardening that `stella context keep` inherits for free, sharing `derive_set_id` kills a second lineage namespace, `statement_reads_as_pasted` is a pure predicate so no I/O enters stella-core, no new deps, and `memory_cmd.rs` lands around 1412 lines — under the ratchet.

### 1. Ordering bug in `write_rule` (`rules_mining.rs`) and `promote_in` (`memory_cmd.rs`)

`inferred_rule_record(...)?` runs **before** the legacy `.md` exists-check. A pre-existing legacy rule whose lesson text is overlong or multi-line — the old writer had no length limit — now returns `Err` on every learning pass, and `learning.rs` prints "rule X not auto-activated: the content spans…" for a lesson that **is already active on disk**. The function's own doc comment promises `Ok(None)` for "either the record surface or the retired markdown one", so doc and code disagree inside the same PR.

Fix: hoist the legacy `.md` check (it needs only `candidate.id`) above the record build. Same reorder in `promote_in`, where re-promoting an already-promoted overlong memory currently says "distill the wording" instead of "already exists".

### 2. The refusal path has no witness

The headline claim — a statement failing the one-sentence shape "is refused at publication with an actionable reason" — is untested. Nothing in the diff feeds a multi-line or >600-char statement to `inferred_rule_record`, `write_rule` or `promote_in`, and the learning loop's new `Err(reason)` branch is never exercised. One test with an overlong candidate asserting `Err` containing "single sentence" closes it.

### 3. An undisclosed behavior change on `promote`

`promotion_candidate` uses full `node.content` (`memory_cmd.rs:873`), and memories are documents — so `stella memory promote` will now refuse any multi-line or >600-char memory outright, with no proposal fallback, unlike the learning loop whose analogous change the description does disclose. This is defensible as the record quality bar and consonant with ADR 0014, but it is a maintainer's call that should be made consciously. Please state it in the description.

### To file as issues (not blocking)

- **The loader doesn't project TOML.** `FsRuleSource` with `RULE_EXTENSIONS = ["md","toml"]` (`rules.rs:101`) parses a TOML record as frontmatter-less markdown, so the miner's `Rule.text` becomes the raw TOML body. The Jaccard dedup (`already_captured`, `stella-core/src/mining.rs:139-149`) can then fall below threshold and a published rule re-enters mining. Bounded today (the proposal store upserts by id, and the fs no-clobber blocks a second file) and pre-existing for kept records — but this PR makes TOML the *only* mined output, and the rewritten `mined_rules_land_where_the_loader_reads` weakens exactly the assertion that would have caught it: `r.id.contains(…) && r.text.contains(…)` passes *because* the text is raw TOML. Proper fix is projecting `.toml` through `load_context_file`, as `records::registry:316` already does.
- **Publication screens less than `validate`.** The "CI would flag it" guard covers only the one-sentence check; `validate` also flags secret-bearing statements (`forbidden_data` → `redact_secrets`, `validate.rs:131-139`), which publication does not screen.

### One correction to the description

"The only shared file is `stella-core/src/lib.rs`" is not accurate — four files are shared with #2287 (core `lib.rs`, core `rules.rs`, core `rules/tests.rs`, cli `rules_mining/tests.rs`). Harmless in outcome: `git merge-tree` of the two branches exits 0 with no conflicts, and the two changes are semantically complementary (#2287 changes what the markdown reader ignores; #2295 changes what the writers emit). But it is the sentence a reviewer relies on to judge merge risk, so it should be right. Recommended order is #2287 first, then this with a branch update and full CI rerun.

macanderson:
Filed the first non-blocking item as **#2314** — the loader parses TOML records as markdown, so the miner's dedup compares raw TOML and can re-mine an already-published rule. It is pre-existing for kept records, but this PR makes TOML the only mined output, so it stops being theoretical. The issue also notes that `mined_rules_land_where_the_loader_reads` currently passes *because* the text is raw TOML, so that assertion needs to tighten with the fix.

Issue #2297, opened 2026-08-08
context: serve the evidence gate's term pass from a posting/FTS index instead of a per-recall corpus scan


macanderson:
## Problem

The recall evidence gate (#2289) needs, per gated recall, each query term's corpus document frequency and each node's matched-term set. Today it gets them from **one streaming pass over every live node's body** (`scan_terms`, `crates/stella-context/src/candidates.rs`; driven by `term_pass` in `crates/stella-context/src/retrieval.rs`). That is the same scan shape the lexical fallback arm has always paid on weak-coverage turns, retention stays one row, and the cost is budgeted by a test (`recall_reads_only_the_candidates_bodies_not_the_whole_corpus` allows corpus_bytes + shortlist bodies) — but it is O(corpus bytes) on **every** gated recall, i.e. every turn under the default `require_evidence: true`. On memory-scale corpora (hundreds of small notes) this is fine; it is the one part of the gate that does not scale.

## What "done" looks like

Term evidence served sublinearly: a token/posting index maintained at **write/warm time** (the `warm.rs` pass already touches each body when embedding — `L-C1`), so a recall answers df and matched-terms from the index instead of streaming bodies. Options, in rough order of fit:

- SQLite **FTS5** over `(display_name, content)` — bm25() comes free, and the fallback arm's scoring could ride the same index later.
- A hand-rolled `node_term` posting table maintained in the same upsert transaction (`L-L1`: one transaction per write batch).

Constraints discovered while building the gate:

- Matching semantics today are **substring** (`haystack.contains(term)`), byte-compatible with the historical fallback scan. A tokenized index changes matching semantics (e.g. `deny.toml` vs `deny`); that divergence must be a decision, not a drift — if the index tokenizes, the fallback scoring and the evidence channel should move together, with the diff in recalled frames measured before shipping.
- Distinctiveness is the BM25/RSJ IDF sign (`evidence::term_is_distinctive`, `2*df <= max(N,2)`) — an index must produce the same df the streaming pass counts, or the admission boundary silently moves. Witness: same corpus, same query, identical `RecallResult` on both paths.
- Freshness: nodes written this turn must be matchable next turn. The streaming pass sees them immediately; an index maintained at warm time must not introduce a lag window (maintain it in `upsert`, not in the background warm).
[...]

Issue #2288, opened 2026-08-08
context: ship a real semantic embedder — the 'tracked follow-up' in embed.rs (R14, ONNX bge-small default) is tracked nowhere


macanderson:
## Problem

Stella's context plane runs a full vector-retrieval pipeline — per-node embeddings keyed `(content_hash, fingerprint)`, exact cosine full-scan (plus an off-by-default IVF ANN accelerator), RRF fusion, MMR diversity, coverage-gated lexical fallback — but the **only `Embedder` implementation in the tree is `HashEmbedder`** (`crates/stella-context/src/embed.rs:121`): a deterministic hashed-character-trigram projection. Every "similarity" score in recall is therefore lexical surface overlap, not semantics.

The module doc (`crates/stella-context/src/embed.rs:5-19`) says the *product* default is meant to be a local ONNX model (bge-small-class, ~130 MB, checksum-pinned weight fetch) per "Risk register R14", and calls that "the tracked follow-up". **It is not tracked**: `R14` appears nowhere under `docs/`, and no GitHub issue covers an ONNX or hosted-API embedder (`gh issue list --search embedder/ONNX/bge` comes up empty). This issue is that tracking.

Downstream cost of the gap: a meaningful relevance floor for recall admission was measured and rejected because HashEmbedder's score distributions for relevant vs. irrelevant frames overlap (`crates/stella-context/src/retrieval.rs:77-82`) — so the "same 5 irrelevant frames every call" bug (#2289) cannot be fully fixed until the scores mean something.

## What exists already (the seam is ready)

- `Embedder` trait + `EmbedderFingerprint` (`crates/stella-context/src/embed.rs`): every stored vector is stamped with its producer; switching embedders is a new fingerprint plus incremental re-embed on next touch (`L-C2`); retrieval never mixes fingerprints. No migration machinery needs inventing.
- The store takes `Arc<dyn Embedder>` at construction (`crates/stella-context/src/store.rs:215`); the CLI currently hardwires `HashEmbedder::default()` at `crates/stella-cli/src/memory.rs:381` — that call site is where a configured embedder would be injected.
- Candidate backends the doc already anticipates: local ONNX (bge-small-class) as the default, and hosted API embedders (Z.ai, OpenAI, Gemini) as BYOK options.

## Constraints discovered

[...]

Issue #2195, opened 2026-08-08
`cite_memory` is registered on every surface and has never once been called — selection health, and therefore efficacy-based retirement, can never populate


macanderson:
## Problem

The context lifecycle's feedback half depends on the model *voluntarily* calling `cite_memory` after using injected context. It never does.

The write path is fully wired — this is not a missing caller:

- `CiteMemory` (`crates/stella-tools/src/memory.rs:161`) is registered unconditionally at `crates/stella-tools/src/registry.rs:452`, catalog availability `Always` (`crates/stella-tools/src/catalog.rs:152`)
- the ledger drains through `ToolRegistry::take_memory_citations` (`registry.rs:2025`) → `memory_citation_rows` (`crates/stella-cli/src/agent/persistence.rs:275`) → `Store::record_memory_citations` (`crates/stella-store/src/lib.rs:908`), inside `record_execution_end`, which has 9 production call sites covering every surface

Measured on the stella workspace: `SELECT COUNT(*) FROM memory_citations` is **0** against 85 executions, 2,720 tool calls, and 8 recallable memories with embeddings present. `/api/memory` returns `citations: []`, and `/api/context-lifecycle` returns `selection_health: []`.

That second emptiness is the consequence that matters. Selection health is the fold over citations — which injected records were judged helpful, by how many distinct tasks, at what confidence. With no citations it is permanently empty, so:

- the Improve tab's health panel shows "Nothing assessed yet" forever
- `is_eligible` / the `min_attributable_uses: 5`, `not_helpful_ratio_threshold: 0.8` policy has no input
- efficacy-based **retirement** (`retire_failing_context`, `crates/stella-cli/src/memory/retirement.rs:342`) can never fire, because nothing can ever be judged failing

So the loop can learn (observations accumulate) but can never *unlearn*, and can never show an operator which injected context is earning its tokens — which is the question the panel exists to answer.

## Why it is plausibly an affordance problem, not a model problem

The tool's description is good and names both handle forms:

> Cite a piece of injected context that actually informed your work this turn — a recalled memory (a `[nod_…]`-tagged line from the relevant-context block) or a workspace rule/record (a `^handle` as shown beside its statement)

[...]

PR #2298, opened 2026-08-08
fix(stella-context): recall admits a frame iff evidence ties it to the query — the frame budget no longer force-fills

### LINKED ISSUE #2289: context: recall always fills its 5-frame budget — no abstention when nothing is relevant, so the same irrelevant frames surface on every call
## Problem

Every `stella` call surfaces a recalled-context block of ~5 frames, and users report the **same five irrelevant frames on every call**. This is not a ranking flake — it is the designed behavior of the recall plane meeting a small/stale corpus:

1. **The budget always fills.** `max_frames` defaults to 5 (`DEFAULT_RECALL_MAX_FRAMES`, `crates/stella-cli/src/settings/context.rs:273`) and `pack_to_budget` (`crates/stella-context/src/retrieval/ranking.rs`) treats it purely as a **cap, never a floor**: there is no minimum-relevance admission and no abstention rung. As long as the workspace's `context.db` holds ≥5 live nodes, five frames are injected into every turn — no matter how badly they score.
2. **The scores cannot mean "irrelevant".** The only `Embedder` implementation shipped is `HashEmbedder` (`crates/stella-context/src/embed.rs:121`) — hashed character-trigram projection, i.e. lexical surface overlap, not semantics. A relevance floor was already **measured and rejected** for exactly this reason: under `HashEmbedder`, contaminant frames scored 0.38–0.50 against a prompt while genuinely relevant frames scored 0.45–0.63 — the distributions overlap, so no threshold separates them (doc comment at `crates/stella-context/src/retrieval.rs:77-82`).
3. **The retirement loop that should prune never-useful memories cannot fire.** Efficacy-based retirement depends on selection-health signals from `cite_memory`, which is registered everywhere and has never been called — see #2195. So a memory that has never once helped is recalled forever.
4. Small corpora make it deterministic: early in a workspace's life the store holds a handful of reflections/episodes, so the same top-5 win every turn regardless of the prompt.

This has already caused real harm, not just noise: the doc comment at `crates/stella-context/src/retrieval.rs:72-75` records a run where five stale frames about deleting test files were recalled into an unrelated TUI-keybinding turn and sent the witness author looking for test files to delete. That incident was mitigated by damping recency (`DEFAULT_RECENCY_WEIGHT = 0.15`, landed in #729), which removed *one* injection mechanism — but the fill-the-budget-unconditionally behavior remains, and it is what users see today.

## What exists already (don't rebuild it)

- Manual suppression: `Store::supersede_node` / `restore_node` (`crates/stella-context/src/store.rs:610`), and CLI-side frame suppression (`is_suppressed_local_frame`, `crates/stella-cli/src/memory/projection.rs:62`). These are per-node and manual — nothing automatic uses them.
- The coverage gate (`L-C6`, `DEFAULT_MIN_COVERAGE`) detects a weak vector signal, but it *adds* labeled lexical-fallback frames — it never abstains.
- The A/B recall control (`arm_recall_control`) can suppress recall on a turn for measurement, so the plumbing for "this turn gets no recall block" already exists.

## Definition of done

Recall can return **zero frames** when nothing clears a meaningful relevance bar, and a workspace whose store contains only stale/irrelevant nodes stops seeing them on every call. Concretely, some combination of:

- An abstention rung: below some admission bar, frames are dropped rather than padded to `max_frames` (the honest version of this needs score distributions that separate relevant from irrelevant — i.e. it is gated on the semantic embedder, #2288).
- And/or an embedder-independent damper: per-turn novelty tracking — a frame recalled N consecutive turns without ever being cited (once #2195 fixes citation) loses rank or is auto-quarantined via the existing `supersede_node` machinery.
- A witness test proving a store of 5 clearly-off-topic nodes yields an empty (or reduced) recall block for an unrelated prompt, where today it yields all 5.

## Constraints discovered

- Retrieval determinism matters here: `RecallTuning` defaults are contractual ("a host that configures nothing gets byte-identical behavior"), so a new admission bar must default off or be introduced as a deliberate behavior change, stated as such.
- The relevance-floor rejection at `retrieval.rs:77-82` is evidence, not dogma — it holds **for HashEmbedder**. Re-measure under a semantic embedder before re-rejecting.
- Recall block injection is volatile-message-side (`crates/stella-cli/src/memory/recall.rs`), after the byte-stable prefix — changes here don't threaten prompt-cache stability (invariant 7).

Refs #2195 (citation signal that efficacy retirement starves without).

macanderson:
## What

A recall frame is now admitted **iff query-conditional evidence ties it to the turn**: an anchor the goal named, 1-hop graph adjacency to an anchor, a lexical match on a term distinctive against the corpus, or a cosine at or above a floor the active embedder has *declared semantically meaningful*. A recall where nothing qualifies returns **zero frames** — the CLI already renders that as no recall block (`inject_recall_block(None)` "adds nothing and touches nothing").

This closes the bug where `max_frames` was a cap that always filled: every embedded node is in the vector list, so any store with ≥5 live nodes injected five frames into every turn regardless of score, and a small workspace surfaced the same five irrelevant memories on every call — the incident documented at `DEFAULT_RECENCY_WEIGHT` (stale delete-the-test-files notes recalled into an unrelated TUI turn, sending the witness author hunting for test files). Damping recency (#729) removed one injection mechanism; this removes the other.

## How

- **`SimilarityPosture` on the `Embedder` trait** (`Semantic { admission_floor }` / `Surface`), following the declared-posture exemplar of `stella-model`'s provider parity matrix (`CachePosture`/`ReasoningPosture`, invariant 8). `HashEmbedder` declares `Surface` — measured, not cautious: contaminant trigram cosines (0.38–0.50) overlap relevant ones (0.45–0.63), so its scores order candidates but never admit them. The method has **no default**: an implementor must answer the question (a semantic claim is an opt-in with a number to defend). The `Semantic` arm is the seam #2288's ONNX/API embedder plugs into, and is pinned by a test with a `Semantic`-posture test embedder.
- **Term distinctiveness is the BM25 / Robertson–Spärck Jones IDF sign** (`2·df ≤ max(N, 2)`; `evidence::term_is_distinctive`): a term in more than half the corpus is glue and admits nothing. Corpus-relative on purpose — no stop-word list, no language commitment, self-calibrating as the store grows.
- **One streaming term pass per recall** (`scan_terms`, replacing `scan_lexical`) computes per-term df and per-node matches; the lexical fallback's historical scores are derived from the same pass (byte-compatible: original term multiplicities and denominator preserved), so a weak-coverage turn no longer scans twice. The fallback arm is gated by the same rule — glue-only matches no longer ride in labeled as lexical grounding.
- **Graph expansion split by seed origin**: anchor-seeded adjacency is evidence; vector-seeded adjacency is ranking only (under `Surface`, the strongest hit certifies nothing, so neither can its neighbors). The partitioned neighbor calls sum to exactly what the single call contributed — ranking is unchanged.
- **Refusals are counted, never silent** (`RecallResult::no_evidence_cut`, following the `candidates_cut` precedent; `L-C5`).

## Deliberate behavior change + escape hatch

`RecallTuning::require_evidence` defaults **on** (`DEFAULT_REQUIRE_EVIDENCE = true`) — the fix is the default, per the maintainer's directive on #2289 ("frame recall is important if and only if relevant"). `context.retrieval.require_evidence: false` in settings restores the old padding byte-for-byte (witnessed). Wired through `RetrievalSettings` with parse + defaults tests, matching every other knob in the block.

**Known conservative window, stated honestly:** in a corpus where *every* node shares the query's topical term (df = N — single-topic young workspaces, and several synthetic test fixtures), lexical evidence saturates and the gate abstains unless anchors/semantic fire. The window is transient (df falls below N/2 as other topics accumulate) and errs toward silence, which is the chosen direction: the failure this PR exists to stop is confidently surfacing noise. #2288's DoD includes re-measuring an admission floor once scores mean something.

## Witness

`crates/stella-context/src/retrieval/evidence_tests.rs` — the module does not compile on `main` (the feature is genuinely absent), and the in-tree fail→pass contrast is explicit: `recall_abstains_when_nothing_ties_the_corpus_to_the_query` (0 frames, gate on) vs `the_escape_hatch_restores_the_old_padding` (the same corpus and prompt yields 5 frames with the gate off — the exact old behavior). Plus: positive control (exactly the relevant node admitted), gated fallback arm, anchors + adjacency without term overlap, and the `Semantic`-posture admission arm.

## Test adaptations (none deleted; two renamed, named below)

Ten pre-existing tests used saturated one-topic fixtures ("note {i} about packing frames to a budget" × N), which the gate now rightly refuses. Each was adapted to keep measuring its own property:

- Fixtures made two-topic or given off-topic ballast so their queries stay genuinely selective (cost bound, migration, point-in-time, tight-budget drop, suppression ×2 — these keep the gate **on**).
- Properties orthogonal to admission (drop-report arithmetic, corpus-scaling benchmark, tuning-knob reachability, recall-tier banding) run with the gate off, each with a comment saying why; admission itself is pinned in `evidence_tests`.
- `recall_reads_only_the_candidates_bodies_not_the_whole_corpus` now budgets **one streaming corpus scan plus the shortlist bodies** — the declared price of the gate, measured rather than hidden. The corpus-scaling benchmark keeps its bytes-identical claim on the ungated path; #2297 (posting/FTS index at warm time) is the follow-up that lets it run gate-on again.

## Cost, stated plainly

Every gated recall pays one streaming pass over live node bodies (the shape the lexical fallback always paid on weak-coverage turns; retention stays one row). Fine at memory-store scale, linear in the corpus — #2297 tracks making it sublinear.

## Verification

`cargo test -p stella-context` (167), `-p stella-cli` (1488 + integration suites), `-p stella-graph -p stella-observatory -p stella-runtime` all green; clippy `-D warnings` across those five crates; `RUSTDOCFLAGS="-D warnings" cargo doc` clean; `make guards-fast` green.

Closes #2289. Refs #2288 (semantic embedder unlocks the admission-floor channel), #2195 (citation efficacy — the other half of pruning), #2297 (term-pass index), #2333 (domain overlap as evidence, once a per-query scope exists), #2336 (the rustdoc gate that hid a broken link in this PR's own doc comment).

## Review round 2: the domain rung is dropped

Domain overlap was listed as a fifth admission channel — "the query was scoped and the node shares a domain" — and **no caller satisfies that precondition**. The session hands recall its whole vocabulary (`Domains::names()` → `session_host` → `ScopedStore::query` → `recall_scoped_excluding`), so "overlaps the query's domains" degenerates to "carries any tag at all": a property of the node, not of the query. Since reflection write-back tags every episode whose turn touched files, that rung re-opened #2289 at full width — every tagged note admitted on every prompt.

The channel had no witness, which is how it shipped inverted while the other four were pinned. The new one, `a_full_vocabulary_domain_scope_is_not_evidence`, fails against the rung (all five contaminants admitted, `no_evidence_cut` 0) and passes without it (zero frames, `no_evidence_cut` 5).

`admissible_ids` loses the `domain_ranked` parameter. Overlap keeps its honest job as an RRF ranking signal at weight 1.0 — it orders the admissible rather than admitting — which is all the caller's data can support. The three prose sites that advertised it as evidence are corrected together; the spec now states the general rule the bug violated: **admission is strictly narrower than ranking, and domain overlap is where the two part.** Reintroducing it properly is #2333, which names the trap: `q.domains` also drives `node_ids_excluded_by_scope`, so narrowing that field in place would silently start *suppressing* memories, and the evidence scope must be additive.

## Summary by Sourcery

Gate recall admission on query-conditional evidence so frame budgets no longer auto-fill with weakly related memories, while keeping an escape hatch to restore the old behavior.

New Features:
- Introduce an evidence gate that only admits recall frames tied to the query via anchors, graph adjacency, distinctive lexical matches, or a semantic similarity floor.
- Expose retrieval evidence gating as a configurable setting in CLI retrieval settings and context tuning, defaulting to require evidence.
- Add a semantic similarity posture declaration to embedders so only explicitly semantic backends can admit frames based on cosine scores alone.

Enhancements:
- Refactor lexical fallback into a single streaming term pass that serves both lexical scoring and the evidence gate without materializing the full corpus.
- Split graph expansion by anchor- versus vector-seeded nodes so only anchor-derived adjacency counts as admission evidence while preserving ranking behavior.
- Extend recall result reporting with a count of candidates refused by the evidence gate, ensuring refusals are observable.
- Document the evidence-based admission behavior and its impact on frame budgets in the adaptive context specification.

Documentation:
- Clarify in the adaptive-context spec that frame admission requires query-conditional evidence and that budgets are caps rather than quotas.

Tests:
- Add dedicated evidence gate tests that cover abstention, positive admission, lexical fallback gating, the escape hatch path, anchor-based admission, and semantic-posture cosine admission.
- Update existing retrieval, tuning, and store tests to account for evidence gating semantics, including fixtures that maintain term distinctiveness and explicit gate-off scenarios.
- Adjust corpus cost and scaling tests to reflect the single streaming term pass plus candidate body loading under the new admission model.

## Review response (adversarial self-review pass)

Two tests are **renamed** (not deleted — the deleted-test guard should read these as the same tests under honest names):
- `recall_reads_only_the_candidates_bodies_not_the_whole_corpus` → `recall_reads_one_term_pass_plus_only_the_candidates_bodies` — the old name claimed a zero-corpus-read property the gated default deliberately no longer has.
- `recall_work_is_bounded_by_frame_count_not_corpus_size` → `ungated_recall_work_is_bounded_by_frame_count_not_corpus_size` — the unqualified name overclaimed for the shipped default; the doc comment now states when (#2297) the claim should return to the default path.

Also: the `evidence.rs` module doc now states the fused arm's new cost plainly (it previously borrowed the fallback's "always paid" claim without the weak-coverage qualifier), and the gate-before-shortlist-cut ordering — previously asserted only in a comment — is pinned by a new witness (`the_gate_runs_before_the_shortlist_cut_not_after`: an inadmissible glue head cannot displace the one admissible candidate below the cut window).


macanderson:
## Review: changes requested — one rung, and it is the one that decides the default configuration

The design here is at the bar, and three of the four axes I pushed hardest on came out clean. The embedder-declared floor is done exactly right: `similarity_posture()` has **no default body** (`crates/stella-context/src/embed.rs:51-55`), so a new backend is forced to answer; `HashEmbedder` returns `Surface`, `Surface → None`, `None → empty semantic hits — a non-semantic embedder genuinely cannot admit on cosine, and `ConstEmbedder` pins the `Semantic` arm from the other side. Layering is clean (pure functions and an accumulator; I/O stays in the parent; nothing leaked into stella-core). Zero-frame handling is complete end to end (`recall.rs:540-542` → `inject_recall_block` → `pipeline.rs:3070-3072`), and invariant 7 is safe because the block was already a volatile user message at the conversation tail — fewer blocks strictly *improves* prefix stability. The witnesses are real: the 0-vs-5 same-fixture contrast, the over-tightening control, and the ordering pin that was previously only a comment.

### The blocker: `domain_ranked` is a rubber stamp in every `stella init`-ed workspace

The gate is a **union** of five channels (`crates/stella-context/src/retrieval/evidence.rs:189-203`), so any one rung admits. `domain_ranked` is fed in whole (`evidence.rs:200`), and it is every live node whose tags intersect the query's domain scope, gated only on that scope being non-empty (`retrieval.rs:888-909`). The description calls this "domain overlap **on a scoped query**" — query-conditional because the caller picked a scope for this query.

The sole production caller does not. I verified it on main:

```rust
// crates/stella-cli/src/memory.rs:387-395
let domains = Domains::load(workspace_root)...unwrap_or_default();
let host = crate::contextgraph::session_host(store.clone(), domains.names(), ...);
// contextgraph.rs:116 — self.domains, unchanged per query
.recall_scoped_excluding(query, &self.domains, &excluded)
```

`domains.names()` is the **entire** vocabulary from `.stella/domains.toml`, loaded once per session. It carries zero bits about the query, and it is never empty in a workspace that ran `stella init`. So every domain-tagged node is admissible on every turn.

The tag population is not small, and it grows with use: `record_taxonomy` (`memory.rs:612-645`, called at init from `agent.rs:1304`) writes one Concept node per domain **and one File node per covered path**, each `.with_domains([d.name])`; `memory.rs:558-576` tags every episode via `domains_for_path(files_touched)`; `writeback.rs:401-406` tags Memory nodes. This repository declares 10 domains over 47 paths — **≥57 unconditionally admissible nodes before a single memory exists**, against a shortlist of `max_frames 5 × mmr_candidate_multiple 4 = 20`. `frame_from_node` applies no kind filter (`retrieval.rs:1246`), and `ranking.rs:141`'s `content_blank` exemption keeps the blank taxonomy nodes from collapsing under dedup, so they each hold a slot. `domain_ranked` sorts by `(overlap desc, id asc)`, and the taxonomy nodes have the lowest ids in the store — so they sit at the **head** of a channel fused at weight 1.0.

Net: the force-fill this PR exists to remove survives in the default configuration of the flagship workspace. The gate bites only where there is no `domains.toml`.

**Why no test caught it:** all seven witnesses call `ContextStore::recall(q)` → `recall_scoped(q, &[])` → empty `query_domains` → the domain rung is silently disabled in every one of them. There is no store-level test of that channel.

### Fix — either is acceptable, (a) is the honest minimum

**(a)** Drop the `domain_ranked` argument from `admissible_ids` (`evidence.rs:189-203`, call site `retrieval.rs:1070-1076`). Domain overlap stays a *ranking* signal in the RRF (`retrieval.rs:1041`), which is all the caller's data can support, and file the follow-up to reintroduce it as evidence once a per-query scope exists.

**(b)** Fix the caller's semantics: derive the scope from the query (`Domains::domains_for_path` over `goal_path_anchors`, already computed at `memory/recall.rs:344`) and fire the rung only when the scope is a proper subset of the vocabulary.

Either way, **add the missing witness**: domain-tagged corpus + full-vocabulary scope + unrelated query → 0 frames. That test fails against this PR as written, which is exactly the point.

### Non-blocking

- `term_is_distinctive` is off by one against its own doc — at `df = N/2` the IDF is exactly 0, not positive, and `assert!(term_is_distinctive(5, 10))` (`evidence.rs:218`) pins the discrepancy.
- Matching is substring, not token (`haystack.contains(term)`, `evidence.rs:121`), so `"api"` becomes admission evidence via `"rapid"`. Tolerable as a scoring heuristic; it is a stronger claim as a gate.
- `no_evidence_cut` means two different things — the fused arm counts against every deduped live candidate (`retrieval.rs:1078-1079`), the lexical arm only against term-matching nodes (`retrieval.rs:942-944`). Same field, two denominators.
- "Counted, never silent" reaches no human: `candidates_cut`, the precedent it follows, is read by nothing outside `stella-context`. The user-visible effect of this default change is that recall quietly stops producing a block.
- Cost moved onto every turn's first-token path. #2297 names the linear-in-corpus half but not the multiplicative query-term factor (terms come from an unbounded goal) nor the 2s `RECALL_TIMEOUT_MS` the workspace-memory leg must fit inside — and a timeout now yields zero frames, indistinguishable from an honest abstention.
- `docs/spec/config-system/stella.today.toml:409-436` claims `ann_enabled` is "the ONE knob here that changes WHICH memories a turn can recall". `require_evidence` is a second and stronger one, and is absent from that file — the claim is now false.
- `retrieval.rs` goes 1320 → 1436 lines, 64 short of the ratchet, in a crate AGENTS.md lists as carrying no god files. Passes today; the next feature in that file will not.

macanderson:
## The domain rung is dropped, and the missing witness is added

Taking option (a), the honest minimum. Pushed as `c906453f` (+ `35348e4e`, a doc-link
repair).

### What was wrong

`admissible_ids` took `domain_ranked` and unioned it into the evidence set, on the
strength of the module doc's claim that "the query **was scoped** and the node shares a
domain". No caller satisfies that precondition:

| site | what it passes |
|---|---|
| `crates/stella-cli/src/memory.rs` | `domains.names()` — the workspace's **entire** vocabulary |
| `crates/stella-cli/src/contextgraph.rs` (`ScopedStore::query`) | that same set, on every recall |
| `crates/stella-context/src/retrieval.rs` (`domain_ranked`) | scores overlap against it |

With the query's domains equal to the whole vocabulary, "overlaps the query's domains"
degenerates to "carries any tag at all" — a property of the **node**, not of **this
query**. And since reflection write-back tags every episode whose turn touched files
(`Domains::domains_for_path`), that is nearly every episode. So the rung re-opened #2289
at full width: the same tagged notes admitted on every prompt, forever — the exact
failure this PR exists to close.

### The witness

`a_full_vocabulary_domain_scope_is_not_evidence` in
`crates/stella-context/src/retrieval/evidence_tests.rs` — five domain-tagged notes, the
full-vocabulary scope, an unrelated query.

Against the PR as written:

```
test retrieval::evidence_tests::a_full_vocabulary_domain_scope_is_not_evidence ... FAILED
every note shares a domain with the whole-vocabulary scope, which says nothing
about this query; got ["note-0", "note-2", "note-3", "note-1", "note-4"]
```

All five admitted, `no_evidence_cut` 0 — the gate fully defeated. With the rung dropped it
returns zero frames and `no_evidence_cut == 5`, and the other twelve gate tests are
unchanged and green.

This closes the last gap in the channel coverage, incidentally: anchor, adjacency,
lexical, escape hatch, gate-before-cut, and semantic posture each had a witness. Domain
had none, which is how it shipped inverted.

### What changed

- `admissible_ids` loses the `domain_ranked` parameter (and its unit test loses the
  corresponding id).
- The call site drops the argument; `domain_ranked` is now moved rather than cloned into
  the RRF fusion, where it stays a **ranking** signal at weight 1.0 — unchanged, and all
  the caller's data can honestly support.
- Three prose sites that advertised domain overlap as an evidence channel are corrected
  together, since they drift otherwise: the `evidence.rs` module doc, the
  `require_evidence` doc comment in `crates/stella-cli/src/settings/context.rs`, and
  `docs/spec/adaptive-context/adaptive-context.md` §6.1. The spec now states the general
  rule the bug violated — admission is strictly narrower than ranking, and domain overlap
  is where the two part.

### Follow-ups filed

- **#2333** — reintroduce domain overlap as evidence behind a real per-query scope. The
  derivation is located (`goal_path_anchors` → `Domains::domains_for_path`,
  `crates/stella-cli/src/memory/recall.rs:343`), and the trap is named: `q.domains` also
  drives `node_ids_excluded_by_scope`, so narrowing that field in place would silently
  start **suppressing** memories. The evidence scope has to be additive, and the rung must
  fire only when the scope is a proper subset of the vocabulary.
- **#2336** — unrelated, found while verifying this. `make doc-warnings` runs without
  `--document-private-items`, so broken intra-doc links in `pub(crate)` code never reach
  the gate. Five are already broken in `stella-context` alone, all pre-existing on `main`.
  My own new link was one of them until `35348e4e`, which is how I noticed.

### Verified

- `cargo test -p stella-context` — 169 passed, 0 failed.
- `cargo test -p stella-cli settings::` — 109 passed, 0 failed.
- `cargo clippy -p stella-context -p stella-cli --all-targets -- -D warnings` — clean.
- `cargo fmt --all --check` — clean.
- `RUSTDOCFLAGS="-D warnings" cargo doc -p stella-context --no-deps` — clean, and clean
  under `--document-private-items` for the lines this PR touches.

PR #2322, opened 2026-08-08
fix(arenabench): reuse the arena already serving a workspace, and let it see matches written after it booted

macanderson:
## What

Two fixes to how `arenabench serve` finds an arena, split out of #2308 where they had ridden along undeclared. They are one change in two halves — the second is what makes the first correct — and they touch nothing the bare-loop arm touches.

**Reuse instead of duplicate.** `serve` bound a socket without ever asking whether an arena was already up, so every session that hit "address in use" and tried the next number left another one running. On this machine that reached **six** arenas — 8181, 8900, 8901, 8902, 8917, 8933 — all serving the identical workspace, none aware of the others.

That is a hazard, not clutter. Every arena reads the same on-disk match store, so a duplicate shows the same runs while adding a port to remember, a process to kill, and two servers both able to drive one run.

`serve` now probes `/api/health` — which already reported `{ok, workspace}` and needed no new endpoint — before binding:

- an arena serving **this** workspace, on the requested port or anywhere in a small bounded window: print its URL and reuse it;
- an arena serving a **different** workspace on the requested port: refuse, since two workspaces are two experiments and sharing a port would conflate them;
- nothing listening: bind as before.

**A running arena must see matches other processes wrote.** `restore_from_disk` ran exactly once, in `ArenaServer.__init__`, so an arena listed only what existed when it booted. `arenabench run` writes its match from its own process, so a run started *after* the arena was invisible to it — and the fix everyone reached for was to start another arena, which swept the store on its way up and could see the new match. Six times.

That is also why reuse alone is not enough: attaching to an arena that cannot see your run is worse than starting a new one. Both halves are needed for either to be right.

## Shape, and the parts worth arguing about

- **The discovery window.** `ARENA_DISCOVERY_PORTS` is a small bounded set of loopback ports, probed with a short timeout. It opens outbound connections to `127.0.0.1` only, and only to ports this tool itself binds. The probe collapses every not-an-arena case to `None` — nothing listening, a foreign service, a wedged arena — because the caller's next move is identical for all of them and a taxonomy there would only be a taxonomy of guesses.
- **`serve`'s default changes from bind to attach.** This is the deliberate behaviour change and the reason the split was asked for. `--no-reuse` forces a fresh arena, and then a taken port fails honestly rather than silently walking to the next number.
- **The sweep is throttled, not per-request.** The listing is polled by every open tab and the sweep stats each match directory, so it runs at most once every two seconds — below the UI's own refresh cadence, so a match appears in the list about as fast as it appears on disk. A single-match lookup sweeps too, but only on a miss: opening a link to a match this arena has not swept yet should return the match, not a 404 that sends the reader off to start another server.
- **A failed sweep is logged and swallowed.** One unreadable match directory must not take down the listing, which is the only view of every *other* match.
- `run` gains a `watch:` line naming the arena already serving the workspace, or telling you to start one. It stays headless and starts nothing: it just knows the match id, and "where do I watch this" was the question previously answered by starting another arena and forgetting the port.

## Witness tests

- `test_a_running_arena_lists_a_match_another_process_wrote` — an arena boots over an empty workspace, another process lands a match in it, and without the sweep the listing stays empty and the direct lookup raises.
- The reuse tests cover all three probe outcomes, including the refusal when the port is held by an arena serving a *different* workspace.

## Verification

`arenabench`: full suite green on this branch against current `main` (541 passed). The six reuse/discovery tests pass by name. No tests deleted.

## Provenance

Cherry-picked unchanged from #2308 (`a81a674a`, `5fafd62d`) onto current `main`, where they belong under their own title — roughly half that PR's diff was this feature, named nowhere in its description.

Refs #2308

## Summary by Sourcery

Reuse existing arenas already serving a workspace instead of starting duplicates, and ensure a running arena sees matches written after it started.

New Features:
- Add discovery of running arenas via /api/health to allow arenabench serve to attach to an existing arena for the same workspace by default.
- Expose a --no-reuse flag so operators can force creation of a fresh arena even when one is already serving the workspace.
- Print a watch URL or guidance from arenabench run pointing to an existing arena for the match being started.

Bug Fixes:
- Ensure ArenaServer periodically reloads matches from disk so matches created by other processes are listed and can be fetched by id instead of returning 404.
- Prevent arenabench serve from silently creating multiple arenas for the same workspace when the requested port is already in use, and refuse ports occupied by arenas for different workspaces.

Enhancements:
- Throttle disk sweeps to a short interval to avoid per-request filesystem scans while keeping the UI listing responsive.
- Add logic for on-demand disk refresh when looking up a single match id that is not yet in memory.

Tests:
- Add tests covering arena reuse behavior, the --no-reuse escape hatch, refusal when a port serves a different workspace, and visibility of matches written by another process.

### DIFF
--- a/arenabench/arenabench/cli.py
+++ b/arenabench/arenabench/cli.py
@@ -66,7 +66,7 @@ def _cmd_run(args: argparse.Namespace) -> int:
     from .credentials import apply_saved_credentials, credentials_path, load_credentials
     from .monitor import MatchWatcher
     from .runner import MatchRunner
-    from .server import ArenaServer
+    from .server import ArenaServer, find_running_arena
 
     try:
         spec = load_match(Path(args.template).expanduser())
@@ -154,6 +154,16 @@ def _cmd_run(args: argparse.Namespace) -> int:
 
     match = runner.create(spec)
     runner.start(match)
+
+    # Name the arena already serving this workspace, if there is one. `run` is
+    # headless by contract and must not start a server, but it knows the match
+    # id and the operator's next question is always "where do I watch it" —
+    # answered previously by starting another arena and forgetting the port.
+    watch_port = find_running_arena(workspace, "127.0.0.1")
+    if watch_port is not None:
+        print(f"watch     : http://127.0.0.1:{watch_port}/matches/{match.id}")
+    else:
+        print(f"watch     : arenabench serve   (then open match {match.id})")
     # The built-in supervisor: the same rules `arenabench watch` runs from
     # another process, reported inline so a CI log shows a dead arm at the
     # minute it died rather than in the postmortem (#1480).
@@ -278,6 +288,7 @@ def _cmd_serve(args: argparse.Namespace) -> int:
             port=args.port,
             open_browser=not args.no_browser,
             allow_remote=args.allow_remote,
+            reuse=not args.no_reuse,
         )
     except ValueError as exc:
         print(f"error: {exc}", file=sys.stderr)
@@ -760,6 +771,12 @@ def main(argv: list[str] | None = None) -> int:
         action="store_true",
         help="permit a non-loopback --host; anyone who can reach it can spend "
              "your provider credits",
+    )
+    serve_parser.add_argument(
+        "--no-reuse",
+        action="store_true",
+        help="start a new arena even if one already serves this workspace "
+             "(default: reuse it and print its URL)",
     )
     serve_parser.set_defaults(func=_cmd_serve)
 
--- a/arenabench/arenabench/server.py
+++ b/arenabench/arenabench/server.py
@@ -115,6 +115,57 @@ def is_loopback(host: str) -> bool:
     return host.strip().lower() in LOOPBACK_HOSTS
 
 
+#: Ports `serve` sweeps looking for an arena that is already up. The default
+#: plus a small window, because that is where previous arenas actually land:
+#: an operator who hits "address in use" tries the next number, and on this
+#: machine that produced six arenas on 8181/8900/8901/8902/8917/8933, all
+#: serving one workspace, none of them aware of the others. Bounded and
+#: deterministic on purpose — discovery must cost a handful of loopback
+#: connects, never a port scan.
+ARENA_DISCOVERY_PORTS: tuple[int, ...] = (8900, 8901, 8902, 8903, 8904, 8905)
+
+
+def probe_arena(host: str, port: int, *, timeout: float = 0.6) -> str | None:
+    """The workspace an arena on ``host:port`` is serving, or ``None``.
+
+    ``None`` covers every not-an-arena case identically — nothing listening, a
+    foreign service, a half-open socket, an arena too wedged to answer — because
+    the caller's next move is the same for all of them and a taxonomy of
+    failures here would only be a taxonomy of guesses.
+    """
+    import urllib.error
+    import urllib.request
+
+    try:
+        with urllib.request.urlopen(
+            f"http://{host}:{port}/api/health", timeout=timeout
+        ) as response:
+            payload = json.loads(response.read().decode("utf-8"))
+    except (OSError, urllib.error.URLError, ValueError):
+        return None
+    if not isinstance(payload, dict) or not payload.get("ok"):
+        return None
+    workspace = payload.get("workspace")
+    return str(workspace) if workspace else None
+
+
+def find_running_arena(
+    workspace: Path, host: str, *, ports: tuple[int, ...] = ARENA_DISCOVERY_PORTS
+) -> int | None:
+    """The port of a live arena already serving ``workspace``, if there is one.
+
+    Matched on the resolved workspace, not merely on "something answered": two
+    arenas over different workspaces are two different experiments and must not
+    be silently conflated.
+    """
+    target = str(workspace.expanduser().resolve())
+    for port in ports:
+        served = probe_arena(host, port)
+        if served and str(Path(served).expanduser().resolve()) == target:
+            return port
+    return None
+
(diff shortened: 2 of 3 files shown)


macanderson:
## Review: changes requested — the swept-in match is labeled `finished`, and that lands on this PR's own headline flow

Thanks for splitting this out of #2308; `git diff` between the two branches on `server.py` is empty, so the cherry-pick is exact. #2308 still carries the same hunks, so merge order needs declaring — rebase it to drop them once this lands.

**The probe's identity check is right, and I want to say so before the defects.** `/api/health` must return `ok` truthy *and* a workspace that resolves to exactly this one — that is a genuine handshake, not "any listener on the port", and a foreign service satisfying it is not realistic.

### Blocker 1 — a live match swept in from disk is reported as finished

`_refresh_from_disk` → `runner.restore_from_disk()` → `_restore_one` (`runner.py:559`) **unconditionally** sets `match.status = "finished"`, a note claiming "restored from artifacts on disk after a server restart", and `finished_at` from a file mtime. `restore_from_disk` skips ids already in `self.matches` (`runner.py:543`, `setdefault` at `:553`), so it never corrects itself, and status is never persisted anywhere — it is memory-only.

Reproduced against this tree:

```
restored: 1
status  : finished
note    : restored from artifacts on disk after a server restart
snapshot status: finished | elapsed: 0.0
second sweep restored: 0 -> status still finished
```

This is not an edge case; it is this PR's headline flow. `run` prints `watch: http://127.0.0.1:PORT/matches/<id>` for a match it *just started*, the operator opens it, the arena misses, force-sweeps, and renders the **live** match as **finished** with `elapsed: 0.0` and a note describing a restart that never happened. `list_matches` emits the same status. Meanwhile `metrics_for` (`runner.py:281`) re-reads job dirs every snapshot, so the cells keep filling — **a partial scoreboard labeled final**, in the tool used to decide whether Stella beat Fable 5. That is worse than the 404 it replaces: honest absence became a confident wrong answer.

The witness test asserts only membership and `isinstance(..., dict)`, so it passes with the wrong status — written to the weaker claim.

**Fix:** stop reusing the *restart* path for the *live-sweep* path. Persist match state (a `state.json` beside `spec.json`, written at the existing transitions) or derive liveness from the seat manifest (`runner.py:845`), and let a later sweep update a non-terminal match instead of pinning it. Assert status in the witness.

### Blocker 2 — `probe_arena`'s except tuple misses the case its docstring promises

`http.client.HTTPException` (`BadStatusLine`, `InvalidURL`, `IncompleteRead`) subclasses neither `OSError` nor `ValueError`. Live repro against a loopback socket returning non-HTTP bytes:

```
UNCAUGHT http.client.BadStatusLine:  P not http at all
```

That is exactly the "a foreign service is listening" case the docstring says collapses to `None`. Blast radius is the concerning part: in `arenabench run`, `find_running_arena` sits **after** `runner.start(match)`, and seats are `start_new_session=True` (`runner.py:867`) — so an uncaught exception there kills the CLI **with the match already launched**, leaving a paid run with no supervisor, no monitor and no cancel.

**Fix:** catch `Exception` (or add `http.client.HTTPException`), and add a direct test of `probe_arena`. All four new tests monkeypatch it, so the predicate itself currently has **zero coverage**.

### Blocker 3 — the banner records nothing that identifies what you attached to

`default_workspace()` (`server.py:951`) is `$ARENABENCH_HOME` or `~/.arenabench` — **one workspace for every checkout on the machine**. So matching on workspace path is near-tautological: `serve` from worktree A attaches to an arena launched from worktree B, running B's `ARENABENCH_STELLA_ADAPTER`, B's `STELLA_BINARY`, B's interpreter — and the arena is the *launcher*, since `ArenaServer` owns the runner and `SutBuilder`. The banner prints URL and workspace, the two fields guaranteed to match, and records nothing else.

This is not hypothetical: #2325 is open because a server was running on this rig with exactly the wrong interpreter, and #2328's script prints the holder's pid and cwd for precisely this reason. **Fix:** have `/api/health` report arena identity (pid, adapter root, `STELLA_BINARY`) and print it in the reuse banner. Attach-by-default is defensible for *listing*; it is not, while the arena is also a launcher and the workspace is machine-global, without recording what you attached to.

### Smaller — follow-up issues are fine

- **`--allow-remote` becomes a silent no-op.** The reuse probe runs after the allow-remote gate on `args.host`, and `is_loopback` (`server.py:111`) excludes `0.0.0.0` — so `--allow-remote --host 0.0.0.0` probes, can attach to a local arena, and returns without ever binding the requested interface. The code comment's "opens outbound connections to `127.0.0.1` only" is false as written.
- **The throttle bypasses itself on the miss path.** `match()` sets `_last_disk_sweep = 0.0` on every miss, so any 404 forces a full sweep (56 match dirs on this rig, iterdir + stat + spec.json each). A tab polling a deleted id re-sweeps continuously, defeating the throttle this PR introduces. `_last_disk_sweep` is also read/written unlocked under `ThreadingHTTPServer` — duplicated work, not corruption.

### Notes

Three of the four new tests are functional tests living in `TestLoopbackOnly`, and none asserts a security property of the new code. No new bind, no `0.0.0.0`, no new inbound surface — the security posture is unchanged, which is good; the placement just oversells. No god file is touched (`arenabench/**/*.py` is ratcheted but not baselined; server.py 955→1074, cli.py 931→948).

One honesty note: the evidence cited is six arenas on 8181/8900/8901/8902/8917/8933, and `ARENA_DISCOVERY_PORTS = (8900..8905)` covers **3 of the 6** — the window is justified by data it doesn't cover. #2328's probe scans 8900..8939, so the repo is about to land two different discovery windows for one question in the same week. Worth reconciling.

Issue #2323, opened 2026-08-08
records: the volatile channel's budget drops are computed, documented, and then discarded — MissingContextKind::NotRendered is never constructed


macanderson:
## Problem

`MissingContextKind::NotRendered` is **never constructed anywhere in production
code**. The record channel's budget-drop ledger is computed, documented in four
places as the thing that closes the silent gap, and then discarded by its only
caller.

`RenderedChannel` (`crates/stella-core/src/records/render.rs`) reports three
things: `text`, `rendered`, and `dropped`. The renderer goes to real trouble to
produce `dropped` — the doc comment explains that a record chosen by the
selector and then dropped by the budget "looks identical, from the ledger, to
one that was never chosen", and that only the renderer can tell them apart.

The one production caller throws two of the three away
(`crates/stella-cli/src/memory/recall.rs:83-86`):

```rust
let text = registry
    .render_volatile_for_turn(&facts, Some(RECORD_CHANNEL_BUDGET))
    .text;
```

So when the 2 000-char `RECORD_CHANNEL_BUDGET` cuts a record, nothing records
it, nothing reports it, and the operator has no way to learn that a record they
published was selected as relevant and then silently did not reach the model.

Verified by search: `NotRendered` occurs only at its own enum definition
(`crates/stella-core/src/context_record/context_use.rs:232`), one match arm
there, one serde test, and four doc comments that claim it is emitted
(`records/render.rs:38`, `records/render.rs:110`, `records/registry.rs:153`,
`memory/recall.rs:47-48`). No construction site exists.

The doc comment at `crates/stella-cli/src/memory/recall.rs:47-48` states the
claim outright, and it is false at that call site:

> Records over the budget are reported as dropped rather than silently lost —
> that gap is `MissingContextKind::NotRendered`.

## Why this matters beyond tidiness

The sibling budget in the same function was fixed for exactly this reason and is
the model to copy: `recalled.dropped` from the context-graph host **is**
consumed a few lines below (`recall.rs:392`), and its comment says why — "the
host's cross-provider merge is a second budget pass and reported nothing at all,
so anything it cut vanished without a trace — a silent truncation `L-C5` bans."
The record channel is the same silent truncation, in the same function, still
unreported.

This is also what makes #2299 only half-observable. That issue fixed *which*
record the budget drops (it now drops ascending-precedence rather than
last-loaded); nobody can see *that* it dropped one.

## Files

[...]

PR #2324, opened 2026-08-08
fix(records): the budget drops the least important record — force, then precedence (#2299)

### LINKED ISSUE #2299: records: the volatile budget drops by load order, not precedence — a precedence-80 record can lose to a precedence-40 one
## Problem

The volatile-channel renderer emits records "in the order given" (`crates/stella-core/src/records/render.rs`, ~line 143) — load order, which is deterministic and fine as a display contract. But the **budget drop inherits that order**: when `render_volatile_for_turn(facts, Some(budget))` cannot fit every selected record, the records that render are the first-loaded ones, not the highest-precedence ones. A `precedence = 80` record can be the one ledgered as `dropped` while a `precedence = 40` record from an earlier file renders.

Surfaced by the four-class hit-discipline certification (`crates/stella-core/src/records/tests.rs`, `four_class_certification::a_budget_drop_is_ledgered_never_silent` — the drop IS ledgered, honesty holds; only the *choice* of victim ignores precedence).

## Fix direction

Under budget pressure, drop ascending-precedence first (least important loses), keeping load order among equals so rendering stays deterministic. This changes volatile-channel bytes only when a budget actually bites, and the volatile channel is per-turn by contract — no cached-prefix stability concern. Decide whether the full (unbudgeted) render order should also become precedence-ranked, or stay load-ordered; the doc comment at `render.rs:143-145` is the contract to amend either way.

## Verify

Extend `a_budget_drop_is_ledgered_never_silent`: with two applicable records at precedence 80 and 40 and a budget that fits one, the 80 renders and the 40 is dropped. That assertion fails today.

## Done when

The budget victim is chosen by precedence, the render-order contract comment matches the implementation, and the certification suite pins both.

macanderson:
## What & why

`render_channel` resolved budget pressure inside the emit loop, so the records
that fit were simply the ones that came first. A `precedence = 80` record could
be ledgered as `dropped` while a `precedence = 40` one from an earlier file took
its place. The drop was *honest* (`RenderedChannel::dropped` recorded it, which
is how the four-class certification in #2306 found this) and still wrong:
arriving later is not a reason to be worth less.

**Scarcity now resolves in `survivors`, ahead of and apart from rendering.**
Order and membership become two questions with two answers:

| Question | Answered by |
|---|---|
| *Which* records render, when a budget binds | force, then precedence |
| *Where* they render | the caller's order, exactly as before |

That split is the one `pack_to_budget` already draws in `stella-context`, and
the exemplar this follows:

- `crates/stella-cli/src/contextgraph.rs:797` — *"Selection is finished; how the
  survivors **render** must now be free of score and cost."*
- `crates/stella-context/src/retrieval/ranking.rs:293` — *"a band is not a
  score… the tier decides nothing until something has to be dropped."*

The sort is **stable**, so records that made no competing claim keep load order.
With budget to spare the block is byte-for-byte what load order alone produced,
pinned by two tests (`None` short-circuits; a generous `Some` walks the whole
ranking — different paths, and only the second is what production passes).

Closes #2299

## Importance is force, then precedence — and why that is not scope creep

The first version of this PR ranked by precedence alone. **Adversarial review
found that it armed a latent inversion rather than closing it**, and the second
commit fixes that. The finding is worth stating plainly because it changes what
this PR is:

The **cached** channel gets "force outranks precedence" *structurally* — it
groups `must` before `should`, so the stronger force meets the budget first. The
**volatile** channel has no grouping, so under precedence-only ranking a `may`
and an `info` record competed on their declared numbers alone. And one authoring
path had those numbers backwards: `precedence_for`
(`crates/stella-cli/src/ingest_cmd/extract.rs:640`) stamped `may = 15` beneath
`info = 20`.

That inversion was **inert before this PR** — precedence only fed conflict
detection, which asks whether two values are *equal* and never which is larger.
Ranking scarcity by precedence would have made it live, on the one production
path where a budget actually binds (`RECORD_CHANNEL_BUDGET = 2_000`,
`crates/stella-cli/src/memory/recall.rs:84`): an extracted corpus over budget
would have dropped the actionable `may` record and kept the `info` trivia.

This is the repo's boundary-parity shape — two sides of one renderer and nothing
checking they agreed — so it is fixed as a class rather than as an instance:

- **`survivors` ranks `(Force::strength, precedence)`.** A no-op for the cached
  channel (force is constant within a group); for the volatile channel it makes
  the sibling's structural guarantee explicit. It ranks the **effective** force,
  so a demoted `must` competes as the `may` it renders as, not the `must` the
  sweep already took away from it.
- **`Force::strength` spells the order out** rather than deriving it from
  declaration order. `derive(Ord)` would read `Must < Info` — backwards — and
  the consumer is a budget where an inverted comparison silently keeps the wrong
  record.
- **`precedence_for` swaps the inverted pair**, and its test derives
  monotonicity from `Force::strength` rather than pinning a table. A table is
  how this inverted in the first place: any test asserting the literals would
  have agreed with the bug.

Fixing only the constant would have left the guarantee resting on every future
authoring path picking numbers that agree with the force ordering.

### The other decisions

1. **The unbudgeted render order stays load order.** The issue offered
   precedence-ranking it instead. I did not, so the display contract ("records
   are emitted in the order given") stays literally true and the change cannot
   perturb a block under no pressure.
2. **Both renderers, not just the volatile one.** The same loop shape carried
   the same defect into `render_cached`. Worth flagging honestly: **that half is
   production-dead today** — both production cached renders pass no budget
   (`agent/prompt.rs:236`, `rules.rs:1080`), so its witness can only ever be a
   unit test. Fixed anyway, because the parameter exists and the next caller to
   pass a budget should not inherit the bug.
3. **A record too long to fit is skipped, not terminal**, so a less important
   record still renders in space the more important one could not have used. The
   one case where a lower-ranked record outlives a higher-ranked one, and it is
   a statement about size, not worth.
4. **Absent `precedence` still reads as `0`.** `records::validate` has always
   read silence that way for conflict detection; two consumers of one field
   disagreeing about what silence means would be worse than the bug being fixed.
   `Record::precedence()` now gives that default one home instead of two.

## The witness

Every witness was verified against **its own distinct negation**, not assumed:

| Test | Negation it fails under |
|---|---|
| `a_budget_drop_is_ledgered_never_silent` (extended per the issue) | precedence sort disabled |
| `the_budget_drops_the_least_important_record_not_the_last_one` | precedence sort disabled |
| `survivors_render_in_load_order_not_in_precedence_order` | precedence sort disabled |
| `the_cached_budget_drops_by_precedence_too` | precedence sort disabled |
| `force_outranks_precedence_in_the_volatile_block_too` | force component removed from the key |
| `a_demoted_record_competes_at_the_force_it_renders_under` | declared force substituted for effective |
| `the_stamped_precedence_agrees_with_the_force_it_came_from` | the old `may=15 / info=20` pair |

On `main`, the issue's own witness fails with:

```
assertion `left == right` failed: the surviving record must be the higher-precedence one
  left: ["mtime-not-recency"]     # precedence 40, loaded first
 right: ["reflog-first"]          # precedence 80, loaded second
```

Four further tests are **guards, not witnesses** — they pass both ways by design
and pin what must *not* change: byte-identity when unbudgeted, byte-identity
under a generous `Some`, load order among equals, and a too-long record not
starving the ones that fit.

## The gate

- [x] `cargo fmt --check`
- [x] `cargo clippy -p stella-core -p stella-cli --all-targets -- -D warnings`
- [x] `cargo test -p stella-core` — 1039 + 8 passed, 0 failed
- [x] `cargo test -p stella-cli --bins` — 1490 passed, 0 failed
- [x] `RUSTDOCFLAGS="-D warnings" cargo doc -p stella-core --no-deps`
- [x] `make guards-fast` — all guards green (`file-size`: none grew;
      `left-behind`: 0)
- [x] Docs updated where behavior changed — `Steering::precedence` (the field
      had two consumers and claimed one), `Force::strength`,
      `RenderedChannel::dropped`, `render_channel`, `render_cached`,
      `precedence_for`, and the `render` module header
- [x] `Closes #2299` in the description **and** as a commit trailer

**Not run locally, deliberately:** `cargo clippy --workspace` and `cargo test
--workspace`. An ArenaBench match was running on this machine throughout, and a
full-workspace build would have stolen CPU from a live measurement. Both changed
crates were built and tested in full; CI carries the workspace tiers. Flagging
it rather than ticking boxes I did not earn — CI was green on the first two
commits.

## Nothing left behind

- [x] Filed: **#2323**, **#2338**

- **#2323** — `MissingContextKind::NotRendered` is **never constructed anywhere
  in production code**. The renderer computes `dropped`, four doc comments
  promise it closes the silent gap, and the only production caller
  (`memory/recall.rs:83-86`) takes `.text` and discards it. So this PR makes the
  budget pick the right victim; nobody can yet see that it picked one. Extended
  with a design constraint from review: `dropped` needs a typed reason
  (crowded-out vs never-could-fit), matching `stella_context::DropReason`.
- **#2338** — `inferred_rule_record` (`context_records.rs:692`) is the only
  authoring path that stamps **no** precedence, so a human-promoted rule sits at
  the floor. Dormant only because it stamps `Should` and cached renders pass no
  budget; it goes live the day the sweep demotes one. Force-first ranking does
  not cover it — same force falls through to precedence.

## Ground-rule check

- [x] No new cross-boundary types — `Record::precedence()` and `Force::strength()`
      are accessors over existing serde fields; no wire or schema change

**`precedence` now serves two masters** — conflict resolution and scarcity. An
author bumping it to silence a conflict warning also changes what survives a
budget. `Steering::precedence`'s doc is rewritten to unify them honestly as "who
wins when both cannot have their way", but #2338 is the empirical evidence that
existing writers set the field under the old single meaning, and it should be
read alongside this PR rather than after it.

**Blast radius.** The volatile channel is per-turn, so nothing here touches
prompt-cache stability. The cached channel's bytes cannot change at all today
(both callers pass `None`).

### DIFF
--- a/crates/stella-cli/src/ingest_cmd/extract.rs
+++ b/crates/stella-cli/src/ingest_cmd/extract.rs
@@ -637,12 +637,18 @@ fn default_basis(kind: RecordKind) -> TruthBasis {
 }
 
 /// The precedence a force implies, so conflicting records order deterministically.
+///
+/// Monotonic with [`Force::strength`], and a test pins that: `may` used to be
+/// stamped *below* `info` here, which read as "an informational note outranks a
+/// relevant one" everywhere precedence is consulted — silently, because the only
+/// consumer at the time was conflict detection, which asks whether two
+/// precedences are equal and never which is larger.
 fn precedence_for(force: Force) -> u32 {
     match force {
         Force::Must => 100,
         Force::Should => 40,
-        Force::Info => 20,
-        Force::May => 15,
+        Force::May => 20,
+        Force::Info => 15,
     }
 }
 
--- a/crates/stella-cli/src/ingest_cmd/extract/tests.rs
+++ b/crates/stella-cli/src/ingest_cmd/extract/tests.rs
@@ -456,3 +456,32 @@ fn a_built_proposal_serializes_to_the_toml_surface() {
     assert_eq!(parsed.schema, stella_core::ingest::SCHEMA_TAG);
     let _ = std::fs::remove_dir_all(&root);
 }
+
+#[test]
+fn the_stamped_precedence_agrees_with_the_force_it_came_from() {
+    // Derived from the force ordering rather than pinned as a table, because a
+    // table is how this inverted in the first place: `may` was stamped 15 and
+    // `info` 20, and any test asserting those literals would have agreed with
+    // the bug. Stated as a property, the next edit that disagrees with
+    // `Force::strength` fails here instead of quietly re-ranking what the
+    // volatile channel's budget drops.
+    let strongest_first = [Force::Must, Force::Should, Force::May, Force::Info];
+    let mut sorted = strongest_first;
+    sorted.sort_by_key(|force| std::cmp::Reverse(force.strength()));
+    assert_eq!(
+        sorted, strongest_first,
+        "this fixture must list every force, strongest first"
+    );
+    for pair in strongest_first.windows(2) {
+        let (stronger, weaker) = (pair[0], pair[1]);
+        assert!(
+            precedence_for(stronger) > precedence_for(weaker),
+            "`{}` implies precedence {} but the weaker `{}` implies {} — a stronger \
+             force must never imply a lower precedence",
+            stronger.as_str(),
+            precedence_for(stronger),
+            weaker.as_str(),
+            precedence_for(weaker),
+        );
+    }
+}
--- a/crates/stella-core/src/ingest/record.rs
+++ b/crates/stella-core/src/ingest/record.rs
@@ -653,6 +653,31 @@ impl Force {
     pub fn is_always_injected(self) -> bool {
         matches!(self, Self::Must | Self::Should)
     }
+
+    /// How strongly this force steers, as a rank: higher outranks lower.
+    ///
+    /// Spelled out rather than derived from the declaration order, even though
+    /// the variants happen to be declared strongest-first. A `derive(Ord)` here
+    /// would read *backwards* — `Must < Info` — and the one consumer that needs
+    /// this is a budget deciding which record to drop, where an inverted
+    /// comparison silently keeps the wrong one. The numbers are gaps, not
+    /// addresses: nothing may depend on their absolute values.
+    ///
+    /// Used by [`records::render`][r] to rank scarcity. The cached channel gets
+    /// the same ordering for free by grouping `must` before `should`, so this
+    /// exists to give the **volatile** channel the guarantee its sibling had
+    /// structurally — `may` outranks `info` there whatever precedence either one
+    /// declared.
+    ///
+    /// [r]: super::super::records::render
+    pub fn strength(self) -> u8 {
+        match self {
+            Self::Must => 3,
+            Self::Should => 2,
+            Self::May => 1,
+            Self::Info => 0,
+        }
+    }
 }
 
 /// How a violation is enforced.
--- a/crates/stella-core/src/ingest/record.rs
+++ b/crates/stella-core/src/ingest/record.rs
@@ -323,6 +323,27 @@ impl Record {
     pub fn on_expiry(&self) -> Option<&str> {
         self.truth.as_ref()?.on_expiry.as_deref()
     }
+
+    /// This record's declared [`Steering::precedence`], or `0` when it declared
+    /// none.
+    ///
+    /// The default is the floor rather than a midpoint on purpose: an
+    /// undeclared precedence is an author who did not make a claim, and the
+    /// two consumers — conflict detection ([`records::validate`][v]) and the
+    /// budget's drop order ([`records::render`][r]) — must read that silence
+    /// the same way. It lives here because it did not: the accessor existed
+    /// privately in `validate`, so the second consumer would otherwise have
+    /// spelled the default a second time, and a default in two places is one
+    /// edit away from two different answers to "which record wins".
+    ///
+    /// [v]: super::super::records::validate
+    /// [r]: super::super::records::render
+    pub fn precedence(&self) -> u32 {
+        self.steering
+            .as_ref()
+            .and_then(|steering| steering.precedence)
+            .unwrap_or(0)
+    }
 }
 
 /// Normalize a lineage id into the id-body slug (`ctx.acme.web.pkg-manager` →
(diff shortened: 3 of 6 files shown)


macanderson:
## Review: approved

Genuine bug fix with four real witnesses. I re-derived the budget arithmetic rather than taking the description's word for it: main accumulated `out.text.len() + section.len() + line.len() > budget` with `section` growing per accepted line; the new pre-pass accumulates `spent + Σ(accepted) + line.len() <= budget` from `spent = out.text.len() + heading.len()`. Same totals, same boundary — only victim *selection* changed, which is exactly the claim.

Worst cases walked, none reachable:

- **Nothing load-bearing can be dropped.** `render_channel` emits only context-record bullets; the goal, the system prefix proper and tool results are not in this path, and the block scaffolding is charged as `spent` rather than being a drop candidate.
- **A `should` cannot displace a `must`** — `render_cached` groups by force *before* calling `survivors` (`render.rs:200-214`), so precedence ranks only within a group. `force_outranks_precedence_in_the_cached_block` pins it, and its `full - 1` budget genuinely starves Should while leaving Must intact.
- A too-long record `continue`s rather than terminating the loop (`render.rs:331`), same as main but now deliberate and tested. `checked_add` at `:329` is a small hardening over main's unchecked `+`.

**Invariant 6 is not implicated** — `budget_chars` here is a character cap on a prompt section, an unrelated concept from `BudgetGuard`; `run_turn`'s guard consults are untouched at `driver.rs:1088` and `:1140`, and `survivors` returns a bitmask with no abort path.

**Invariant 7 is safe, by a stronger argument than the description makes.** The only production cached-channel call site is `agent/prompt.rs:236`, which passes `None` — and `survivors` early-returns `vec![true; n]` on `None` (`render.rs:319-321`), so the byte-stable prefix is byte-for-byte identical to main. The only place a budget binds in production is `memory/recall.rs:84` → `render_volatile_for_turn(&facts, Some(RECORD_CHANNEL_BUDGET))`, i.e. the per-turn volatile tail. Worth stating plainly since the description doesn't: the cached-channel half of this fix is currently **unreachable in production** — nothing outside tests passes a cached budget. That is fine and it is the right shape, but don't read `the_cached_budget_drops_by_precedence_too` as a shipping behavior change.

Invariant 2 holds (plain sync functions over borrowed data; no fs/io/tokio/async in the added lines; zero manifest changes). Discipline is clean: no `#[allow]`, no markers, the one `.expect()` is under `#[cfg(test)]`, and no touched file is baselined — `validate.rs` actually *shrinks* 664 → 656 by deleting the duplicated default.

The witness I'd point a reviewer at is `survivors_render_in_load_order_not_in_precedence_order`: survivors are `cc`+`bb` but render `bb`,`cc`. It's the only one that proves membership and order are separate questions.

### Nits, none blocking

1. `registry.rs:122` — `/// Every record, in merged precedence order.` means *scope-merge* order, but now that `precedence` is a load-bearing field with two consumers it reads as a direct contradiction. One-word fix ("scope-merge order"); stale comments are bugs here.
2. `render.rs:329` — `budget_chars` is measured in **bytes** (`String::len()`), not chars, so a 2000-"char" budget under-fills for non-ASCII statements. Pre-existing in main's `over_budget`, carried forward — worth an issue.
3. `ranked()` mutates `steering.precedence` *after* `at()` → `loaded_from()` → `record.stamp()` has computed `record_id`/`record_hash`, leaving a stale hash. Inert today (render reads neither), but a future hash-integrity test in this module would silently get a lie. Set precedence before `loaded_from`.
4. `an_unbudgeted_block_is_byte_identical_whatever_the_precedences` is billed as the safety claim but cannot fail under this design (rendering never consults precedence, and `survivors` short-circuits on `None`). It guards a future refactor, not this one.
5. `crates/stella-tools/src/exploration.rs:373` carries the identical load-order-greedy budget shape. Different crate, different declared contract, so out of scope — but it is the same class.

**No property test.** AGENTS.md names this exact area as one where a property test is a welcome contribution, and the crate already has `proptest` wired. The invariants here are unusually crisp: no survivor set exceeds budget; a dropped `i` beside a kept `j` implies `prec(i) < prec(j)` or `i` didn't fit; `rendered ∪ dropped == inputs` with `rendered` a subsequence of `inputs`; and `budget = None` ⟹ identical to unbudgeted for *any* precedence assignment. Missed opportunity rather than a blocker.

`fmt + clippy + test` has since passed, which was the one thing outstanding. Arming auto-merge.

Issue #2304, opened 2026-08-08
Epic: the trace-replay learning harness — feed execution traces, watch Stella build skills, tools, rules, and memories, with zero model calls


macanderson:
## Goal

A local test system that runs the agent's **learning machinery** through a harness needing **no model calls**: feed traces of executions and watch what Stella builds — skills, tools, memories, rules — over a simulated months-long engagement by one engineer on one project. The maintainer's framing (2026-08-08): "it is the reverse" of the live loop — instead of sessions producing traces that feed learning, recorded traces drive learning directly.

## The reverse entry points (mapped 2026-08-08 — start here)

Every learner sits below the model boundary or behind a port, so trace-driven replay needs no new architecture:

1. **Memory formation** — `SessionMemory::reflect_and_record` (`crates/stella-cli/src/memory/learning.rs:57`) makes exactly ONE model call (`reflect_on_turn`) and everything after it is deterministic: forgetting filter → `context.db` write → `reflections.jsonl` append → miners. `Provider` is a port, and a `ScriptedProvider` double already exists (`crates/stella-cli/src/subagent/tests.rs:36`, also `crates/stella-core/src/subagent/tests.rs:36`) — replaying canned reflection outputs per turn IS the no-model harness.
2. **Skill learning** — `SessionMemory::auto_create_skills` (`crates/stella-cli/src/memory/learning.rs:372`; typed path `:606`, lexical path `:756`) mines `reflections.jsonl` via `stella_core::skills::mine_skill_candidates` / `decide_auto_creation` (`crates/stella-core/src/skills.rs:650`, `:833`) — pure over the log; feed fixture logs directly.
3. **Rules mining** — typed observations (`crates/stella-cli/src/memory/observations.rs`, cursor keyed `reflections.jsonl`) → `induce_rule_proposals` (`crates/stella-cli/src/memory/rules_mining.rs`) → auto-activation at `auto_activate_at_confidence` (default 85) → TOML record publication (#2295). Pure over `ObservationRecord`s.
4. **Tool foundry** — `crates/stella-core/src/tool_foundry.rs` is pure BY DESIGN: `ShellInvocation`s in (the CLI reads them from the durable `tool_calls` receipt log), `ProposedTool` out; signature normalization with typed holes; requires ≥2 distinct argument sets. Zero model involvement even in production. Adoption gating: `stella-tools` `foundry_author`/`foundry_witness`/`foundry_gate`.
[...]

Issue #2330, opened 2026-08-08
core+cli+tui+store+tools: ports::Clock::now_ms is re-implemented by hand in 7 places — including inside the crate whose doc says it never carries a time source


macanderson:
## Problem

`stella_core::ports::Clock` has exactly one method:

```rust
pub trait Clock: Send + Sync {
    /// Monotonic milliseconds since an arbitrary epoch.
    fn now_ms(&self) -> u64;
}
```

Its own doc comment (`crates/stella-core/src/ports.rs:185`) says the production implementation "belongs to the binary that wires the engine (the CLI's `runtime` module), so **`stella-core` never carries a concrete time source of its own**."

**Seven private functions in the tree re-implement that method byte-for-byte**, reading the wall clock directly instead of taking the port. Every one is the identical body:

```rust
fn now_ms() -> u64 {
    SystemTime::now()
        .duration_since(UNIX_EPOCH)
        .map(|d| d.as_millis() as u64)
        .unwrap_or(0)
}
```

| # | Site | Notes |
|---|---|---|
| 1 | `crates/stella-core/src/mcp_usage.rs:54` | **Inside the crate that owns the port** — directly contradicts `ports.rs:185` |
| 2 | `crates/stella-cli/src/memory/self_tuning.rs:275` | also one of the six sites in #2320 |
| 3 | `crates/stella-cli/src/command_deck.rs:143` | `pub(crate)`, so callers multiply |
| 4 | `crates/stella-tui/src/deck_shell.rs:106` | |
| 5 | `crates/stella-store/src/sessions.rs:573` | `pub(crate)`; stamps `SessionRecord` times |
| 6 | `crates/stella-tools/src/gather.rs:143` | |
| 7 | `crates/stella-tools/src/exploration.rs:160` | |

(`crates/stella-tui/examples/deck_demo.rs:53` is an eighth, in an example — leave it.)

This is not a missing abstraction. The port exists, is documented, has **six test doubles already written** — `ManualClock` (`crates/stella-core/src/router.rs:564`), `FrozenClock` (`crates/stella-cli/src/fleet_warmth.rs:264`), `FixedClock` (`crates/stella-cli/src/agent/output.rs:334`), plus doubles in `crates/stella-pipeline/src/pipeline/tests.rs:532`, `crates/stella-fleet/src/fleet/tests.rs:27`, `crates/stella-fleet/src/monitor.rs:826` — and is used correctly by two consumers. It is simply not reached for.

[...]

PR #2335, opened 2026-08-08
fix(stella-cli): the learning plane reads an injected clock — replayable sessions, and a testable truth sweep (#2320)

### LINKED ISSUE #2320: cli: the learning path bypasses the Clock port and reads the wall clock in six places — retirement has never been testable
## Problem

`stella-context` ships a `Clock` port precisely so the context plane can be replayed deterministically — `trait Clock`, `SystemClock`, `FixedClock` (with `advance`/`set`) at `crates/stella-context/src/clock.rs`, and `ContextStore::open_with` takes an `Arc<dyn Clock>` (`crates/stella-context/src/store.rs:213`). The context store's own tests already inject `FixedClock` in eight places.

**The CLI's learning path routes around it.** `SessionMemory::open_with_workspace_skills` hardcodes `Arc::new(SystemClock)`, and six sites read the wall clock directly instead of going through the port:

| Site | What it stamps | Consequence |
|---|---|---|
| `crates/stella-cli/src/memory.rs:192` (`default_task_id`) | the task boundary governance counts distinct tasks by | also embeds `std::process::id()` — nondeterministic on two axes |
| `crates/stella-cli/src/memory.rs:567` (`record_episode`) | episode end time | episodes land out of order under replay |
| `crates/stella-cli/src/memory.rs:662` (`unix_now_secs`) | the episode timestamp primitive | `pub(crate)`, so callers multiply |
| `crates/stella-cli/src/memory/reflection.rs:474` | reflection recording time | lesson recency ranking drifts with real time |
| `crates/stella-cli/src/memory/learning.rs:444` (`retire_failing_context`) | the `now` handed to the truth sweep | **retirement has never been testable** — the sweep always sees today, so no test can ever place a record past its TTL |
| `crates/stella-cli/src/memory/self_tuning.rs:276` | the tuning window | self-tuning cannot be replayed |

The `retire_failing_context` row is the one that is a defect on its own terms, independent of any harness: there is currently no way to write a test that advances time past a record's TTL and asserts the sweep retires it.

Contrast `stella-core::records`, where every lifecycle predicate already takes `now: &str` as a parameter (`records/clock.rs:219`, `:232`, `:252`; `records/decision.rs:222`). The records plane is replayable. The CLI memory plane is not. That asymmetry is the whole gap.

## Why now

This blocks epic #2304 (the trace-replay learning harness), whose Phase 1 requires that "timestamps come from the trace, never the wall clock." It is a strict prerequisite for every assertion in that epic — see `doc:trace-replay-learning-harness` §2, and PR #2318 for the spec.

But it is worth doing regardless of #2304, because of the retirement row above.

## What to do

1. Add `clock: Arc<dyn stella_context::Clock>` to `SessionMemory` and hand it to `ContextStore::open_with`.
2. Replace all six wall-clock reads above with `self.clock.now_unix_secs()` / `now_rfc3339()`.
3. Keep every existing `SessionMemory::open*` constructor defaulting to `SystemClock`, so production behaviour is byte-identical. Add one `open_with_clock` constructor for tests and the harness.
4. `default_task_id` loses `std::process::id()` in favour of an explicit session ordinal — a pid is not a time source and makes the id nondeterministic even with a fixed clock.
5. Un-gate `SessionMemory::set_task_id`. It is currently `#[cfg(test)]` with a doc comment reading *"Drop the gate in the same commit that adds the first real caller."* A replayer that knows its own task boundaries is exactly that caller; if this issue lands before #2304's replayer, leave the gate and note it.
6. **Audit `crates/stella-context/src/store/edge.rs:121`**, which reads `SystemTime::now()` for nanoseconds. Determine whether that value reaches any comparable or serialized field or is only an id nonce. If comparable, move it behind the clock in this PR. If a nonce, leave a comment saying so, so the next reader does not re-audit it.

## Witness

Open two `SessionMemory` instances against `FixedClock(T)` for the same `T`, run one scripted turn through each (`ScriptedProvider` — see `crates/stella-core/src/subagent/tests.rs:36`), and assert the two resulting `.stella/private/reflections.jsonl` files are **byte-identical**.

Fails on `main`: `default_task_id` embeds `SystemTime::now()` and the pid, so the `task_id` field differs between the two instances even at a fixed clock. Passes after.

A second witness worth adding while the seam is open: set the clock past a record's TTL and assert `retire_failing_context` retires it — impossible to write today, which is the point.

## Constraints

- **Invariant 2 (no I/O in the engine)** is unaffected: `stella-context::Clock` is already the port, this only stops the CLI bypassing it.
- **Invariant 7 (byte-stable prompts)**: production must be byte-identical after this change. Defaulting every existing constructor to `SystemClock` is what guarantees that; do not change any constructor's signature that production calls.
- `crates/stella-cli/src/memory.rs` is 670 lines and `memory/learning.rs` is 785 — neither is in `scripts/file-size-baseline.txt`, so both have headroom to the 1500 ratchet. Still prefer landing new logic in the existing submodules rather than growing `memory.rs`.
- `stella-cli` is bin-only (`[[bin]]`, no `[lib]`), so the witness must be an in-crate test module under `crates/stella-cli/src/memory/tests/` — follow `ab_control.rs` / `skill_creation.rs`. Files under `crates/stella-cli/tests/` drive the compiled binary and cannot reach `SessionMemory`.

## Definition of done

- All six sites read through the injected clock; no `SystemTime::now()` remains in `crates/stella-cli/src/memory/` or `memory.rs`.
- `edge.rs:121` is either moved behind the clock or annotated as a nonce.
- The byte-identical-`reflections.jsonl` witness exists and fails on `main`.
- `make gate` green.

## Related

- Epic #2304 — the trace-replay learning harness this unblocks. Spec: PR #2318.
- #2195 — `cite_memory` never called, so efficacy-based retirement can never populate. A different defect on the same retirement path; both have to be fixed before retirement works end to end.
- #2299 — volatile budget drops by load order, not precedence.

macanderson:
Closes #2320. First of the five PRs in `doc:trace-replay-learning-harness` §9 — the strict prerequisite for the rest of epic #2304.

## The gap

`stella-context` has shipped a `Clock` port since the store went bi-temporal (`trait Clock`, `SystemClock`, `FixedClock` with `advance`/`set`), and `ContextStore::open_with` has always accepted one. The context store's own tests inject `FixedClock` in eight places.

**The CLI's learning path routed around it** — six direct `SystemTime::now()` reads, plus a pid folded into the task id. Contrast `stella-core::records`, where every lifecycle predicate already takes `now: &str` as a parameter. The records plane was replayable; the CLI memory plane was not.

Two consequences, and only the second is about the harness:

- `retire_failing_context` built its `now` from the wall clock, so **the truth sweep always saw today** and no test could place a record past its TTL and assert it retires. A defect on its own terms.
- A session's mined log could not be reproduced, which blocks every Phase-2 assertion in the spec.

## What changed

`SessionMemory` holds an `Arc<dyn Clock>`, hands it to `ContextStore::open_and_warm`, and reads it at all six sites. Every existing constructor defaults to `SystemClock`, so **production is byte-identical** — invariant 7 (byte-stable prompts) is unaffected by construction, not by inspection.

| Site | Now reads |
|---|---|
| `memory.rs` `default_task_id` | the session clock + an injected nonce |
| `memory.rs` `record_episode` | `self.clock` |
| `memory.rs` `unix_now_secs` | `SystemClock` **through the port** |
| `memory/reflection.rs` `parse_lessons_checked` | a `now_unix_secs` parameter |
| `memory/learning.rs` `retire_failing_context` | `self.clock` |
| `memory/self_tuning.rs` `now_ms` | a `&dyn Clock` parameter |

No `SystemTime::now()` remains anywhere under `crates/stella-cli/src/memory/` or in `memory.rs`.

### Three judgement calls worth reviewing

**1. The pid stays in production, injected rather than removed.** The issue asks for `std::process::id()` to be dropped for "an explicit session ordinal". Dropping it outright loses the property it was there for: two stella processes starting in the same second against one workspace would file their lessons under *one* task id, and governance counts distinct tasks — a silent undercount. Injecting it gets both: `process_session_nonce()` for production (byte-identical to today), a trace-supplied nonce for replay. Flagging it explicitly since it departs from the issue text.

**2. `Clock` gains `now_unix_millis()`.** `self_tuning`'s ledger stamps `at_ms`, and flooring it to `secs * 1000` would have been a real behaviour change. The trait default derives from seconds — correct for `FixedClock`, which is *set* in whole seconds and must not invent a sub-second digit — and `SystemClock` overrides it to keep the precision its callers already had.

**3. `unix_now_secs()` reads `SystemClock` rather than taking a clock.** Its callers are the turn drivers (`agent.rs`, `agent/goal.rs`, `command_deck.rs`) stamping when a real turn started, not the learning plane; a replay supplies `started_unix_secs` from the trace and never calls it. Routing it through the port anyway leaves exactly one spelling of "seconds since the epoch" in the crate, so the next wall-clock read cannot quietly grow back there.

## The `edge.rs:121` audit the issue asks for

**Outcome: identity nonce, not a time source. Annotated in place, left off the clock.** `process_nonce()` hashes into `edge.public_id`, so it *is* serialized — but never into a **comparable** field: nothing orders, ranges over, or compares `public_id`. The row's time is the separate `recorded_at` column, which callers already pass in and the port already governs. Putting a fixed clock behind the nonce would only make two processes in one second mint from the same keyspace, which is the collision it exists to prevent.

**A finding the issue did not anticipate, recorded in the same comment:** `public_id` is unreproducible anyway, and not because of the nonce. `EDGE_SEQ` is a *process-global* counter, so two replays of one trace **inside a single process** mint different ids for the same logical edge. An edge `public_id` is therefore never admissible in a replay summary compared for byte-identity — the harness must compare edge content, not edge identity. That constrains PR 4's metrics and is noted where the next reader will hit it.

## Witness

`crates/stella-cli/src/memory/tests/clock_seam.rs`, three tests.

**`a_session_records_its_clocks_instant_not_todays` is the witness.** A scripted-provider reflection turn on `FixedClock(1_700_000_000)` must write a log whose `occurred_at` and `task_id` carry *that* instant.

Verified the artisanal way — the two behaviour lines were reverted behind the new seam and the suite re-run:

```
test a_session_records_its_clocks_instant_not_todays ... FAILED
  assertion `left == right` failed: occurred_at must come from the session clock, not the wall clock
    left: 1786222713
   right: 1700000000
```

**Why the issue's proposed witness is not one, and is kept anyway.** "Two `FixedClock(T)` sessions write byte-identical `reflections.jsonl`" **passes on `main`** — two opens milliseconds apart share a wall-clock second and a pid, so the ids match and nothing is proven. It did exactly that in the run above (`two_runs_of_one_scripted_turn_are_byte_identical ... ok` with the old behaviour restored). It stays in the suite because byte-identity is the property the harness depends on; the pinned instant is what proves the seam exists. Both are needed, and the module doc says so.

The third test, `the_truth_sweep_sees_the_session_clock`, advances the session clock a year and asserts the sweep's `now` moves with it — the assertion that was impossible to write before this change.

## Verification

- `cargo test -p stella-cli` — 1492 passed, 0 failed (plus 12 integration binaries green)
- `cargo test -p stella-context` — 158 passed
- `cargo clippy -p stella-cli -p stella-context --all-targets -- -D warnings` — clean
- `make guards-fast` — all green, including `file-size` (nothing grew past its ceiling) and `check-invariants`

**No tests deleted.** Three test call sites gained an explicit instant argument (`PARSED_AT`, `REFLECTED_AT`, `TEST_CLOCK`) instead of an implicit wall-clock read; every one is the same test asserting the same thing.

## Not in this PR

- `set_task_id` **stays `#[cfg(test)]`-gated.** Its doc comment says "drop the gate in the same commit that adds the first real caller", and that caller is the replayer in PR 2. #2320 anticipated this ordering and says to leave the gate.
- `open_with_clock` is `#[cfg(test)]` for the same reason `set_task_id` is: `stella-cli` is bin-only, so an ungated constructor nothing can reach would be dead code with an `#[allow]` on it.

## Summary by Sourcery

Route stella-cli’s learning and tuning paths through the shared clock abstraction to make session behavior deterministic, replayable, and testable while preserving production behavior.

Bug Fixes:
- Ensure the truth sweep in `retire_failing_context` uses the session clock so tests can correctly retire records that have passed their TTL.

Enhancements:
- Make `SessionMemory` hold an injected `Clock` and session nonce, using them for task IDs, episode timestamps, and reflection lesson timestamps instead of direct wall-clock reads.
- Extend the `Clock` trait with millisecond-resolution timestamps and implement precise millisecond reading for `SystemClock` while keeping `FixedClock` deterministic.
- Refine the self-tuning ledger to take a `Clock` parameter for all timestamping, and adjust CLI commands to pass the production clock explicitly.
- Document and audit edge `public_id` nonce behavior to clarify its non-temporal nature and replay implications.

Tests:
- Add a dedicated test module that verifies the learning plane and truth sweep respect the injected clock and that scripted sessions produce byte-identical logs under a fixed clock.
- Update existing reflection and self-tuning tests to supply explicit instants or fixed clocks, removing dependence on the ambient wall clock.

### DIFF
--- a/crates/stella-cli/src/memory.rs
+++ b/crates/stella-cli/src/memory.rs
@@ -29,7 +29,7 @@ use std::path::{Path, PathBuf};
 use colored::Colorize;
 use serde::{Deserialize, Serialize};
 use stella_context::{
-    ContextDelta, ContextStore, DomainInput, EpisodeInput, EpisodeOutcome, FactAssertion,
+    Clock, ContextDelta, ContextStore, DomainInput, EpisodeInput, EpisodeOutcome, FactAssertion,
     HashEmbedder, NodeInput, NodeKind, RecallTier, SystemClock, format_rfc3339,
 };
 use stella_core::skills::{self, SelectionConfig, Skill};
@@ -183,17 +183,32 @@ impl LessonKind {
     }
 }
 
-/// A session-scoped task identity, distinct per process.
+/// A session-scoped task identity, distinct per concurrent session.
 ///
 /// One `stella run` is one task, so for the headless path this is exactly the
 /// right boundary. For a long REPL session it is an approximation, but a
 /// strictly better one than per-turn.
-fn default_task_id() -> String {
-    let secs = std::time::SystemTime::now()
-        .duration_since(std::time::UNIX_EPOCH)
-        .map(|d| d.as_secs())
-        .unwrap_or(0);
-    format!("session:{secs}-{}", std::process::id())
+///
+/// Both halves are now injected rather than read from the ambient process
+/// (#2320). The time comes from the session's [`Clock`], so a fixed clock
+/// yields a fixed id; `nonce` is whatever distinguishes this session from
+/// another running concurrently against the same workspace — see
+/// [`SessionMemory::session_nonce`].
+fn default_task_id(clock: &dyn Clock, nonce: &str) -> String {
+    format!("session:{}-{nonce}", clock.now_unix_secs())
+}
+
+/// What a production session uses for the [`default_task_id`] nonce.
+///
+/// The pid is kept here — and *only* here — because the nonce's whole job is
+/// to keep two stella processes started in the same second from filing their
+/// lessons under one task id, and governance counts distinct tasks. Dropping it
+/// outright would silently merge two concurrent sessions into one task, which
+/// is a governance undercount rather than a determinism win. Injecting it
+/// instead gets both properties: production is byte-identical to before, and a
+/// replay supplies its own nonce from the trace.
+fn process_session_nonce() -> String {
+    std::process::id().to_string()
 }
 
 mod reflection;
@@ -270,6 +285,20 @@ pub struct SessionMemory {
     /// re-deriving it here would re-walk the rule directories and re-run the
     /// truth sweep on every turn.
     record_registry: Option<stella_core::records::Registry>,
+    /// This session's time source — the only one the learning plane reads
+    /// (#2320).
+    ///
+    /// `stella-context` has shipped this port since the store was bi-temporal,
+    /// and `ContextStore::open_with` has always taken one; the CLI's learning
+    /// path was simply routing around it, reading `SystemTime::now()` at six
+    /// sites. That asymmetry had a cost beyond replay: `retire_failing_context`
+    /// stamps its truth sweep from here, so with a hardcoded system clock no
+    /// test could ever place a record past its TTL and watch it retire.
+    ///
+    /// Production hands this a [`SystemClock`] from every existing constructor,
+    /// so prompts stay byte-stable (invariant 7). A test or a trace replay hands
+    /// it a `FixedClock` through [`Self::open_with_clock`].
+    clock: std::sync::Arc<dyn Clock>,
 }
 
 impl SessionMemory {
@@ -277,6 +306,34 @@ impl SessionMemory {
     /// the store can't open — a session without memory beats no session.
     pub fn open(workspace_root: &Path, warn: bool) -> Option<Self> {
         Self::open_with_workspace_skills(workspace_root, warn, false)
+    }
+
+    /// Open the workspace's memory against a caller-supplied time source and
+    /// session nonce — the door a deterministic replay comes through (#2320).
+    ///
+    /// Every other constructor defaults to [`SystemClock`] and the pid, so
+    /// production behaviour is unchanged; this one exists so a harness can put
+    /// the whole learning plane on a clock it advances itself.
+    ///
+    /// Test-gated because `stella-cli` is a bin-only crate: there is no
+    /// external consumer that could call it, and shipping it unreachable would
+    /// be an `#[allow(dead_code)]` describing an API nothing can reach — the
+    /// same reasoning that gates [`Self::set_task_id`] below.
+    #[cfg(test)]
+    pub(crate) fn open_with_clock(
+        workspace_root: &Path,
+        warn: bool,
+        include_workspace_skills: bool,
+        clock: std::sync::Arc<dyn Clock>,
+        session_nonce: impl Into<String>,
+    ) -> Option<Self> {
+        Self::open_inner(
+            workspace_root,
+            warn,
+            include_workspace_skills,
+            clock,
+            session_nonce.into(),
+        )
     }
 
     /// Override the task boundary lessons are stamped with.
@@ -296,6 +353,13 @@ impl SessionMemory {
     #[cfg(test)]
     pub(crate) fn set_task_id(&mut self, task_id: impl Into<String>) {
         self.task_id = task_id.into();
+    }
+
+    /// This session's current instant, as the truth sweep and every lifecycle
+    /// predicate see it.
+    #[cfg(test)]
+    pub(crate) fn clock_now_rfc3339(&self) -> String {
+        self.clock.now_rfc3339()
     }
 
     #[cfg(test)]
@@ -364,6 +428,22 @@ impl SessionMemory {
         warn: bool,
         include_workspace_skills: bool,
     ) -> Option<Self> {
+        Self::open_inner(
+            workspace_root,
+            warn,
+            include_workspace_skills,
+            std::sync::Arc::new(SystemClock),
+            process_session_nonce(),
+        )
+    }
+
+    fn open_inner(
+        workspace_root: &Path,
+        warn: bool,
+        include_workspace_skills: bool,
+        clock: std::sync::Arc<dyn Clock>,
+        session_nonce: String,
+    ) -> Option<Self> {
         // Ephemeral benchmark trials must neither recall task/user-planted
         // learning state nor create or migrate a context database that can
         // perturb the task under test. Reflection is separately pinned off
@@ -379,7 +459,7 @@ impl SessionMemory {
         match ContextStore::open_and_warm(
             &db_path,
             std::sync::Arc::new(HashEmbedder::default()),
-            std::sync::Arc::new(SystemClock),
+            clock.clone(),
         )
         .map(|store| store.with_tuning(retrieval.tuning()))
         {
@@ -408,8 +488,9 @@ impl SessionMemory {
                     ab_turn: 0,
                     // Phase 3 (#714)
                     lifecycle_enabled: tuning::session_lifecycle_enabled(workspace_root),
-                    task_id: default_task_id(),
+                    task_id: default_task_id(clock.as_ref(), &session_nonce),
                     execution_id: None,
+                    clock,
                 })
             }
             Err(e) => {
@@ -564,10 +645,7 @@ impl SessionMemory {
             }
         }
 
-        let now_secs = std::time::SystemTime::now()
-            .duration_since(std::time::UNIX_EPOCH)
-            .map(|d| d.as_secs() as i64)
-            .unwrap_or(started_unix_secs);
+        let now_secs = self.clock.now_unix_secs();
         let mut episode = EpisodeInput::new(
             summary,
             format_rfc3339(started_unix_secs),
@@ -658,12 +736,18 @@ pub(crate) fn context_db_path(workspace_root: &Path) -> Result<PathBuf, String> 
         .map_err(|e| format!("cannot resolve private context state: {e}"))
 }
 
-/// Seconds since the Unix epoch — the episode timestamps' primitive.
+/// Seconds since the Unix epoch — the episode timestamps' primitive, read
+/// through the [`Clock`] port rather than `SystemTime` directly.
+///
+/// This is the *production* clock by construction, and deliberately so: its
+/// callers are the turn drivers (`agent.rs`, `agent/goal.rs`, `command_deck.rs`)
+/// that stamp when a real turn started, not the learning plane. A session's own
+/// timestamps come from its injected [`SessionMemory::clock`]; a replay supplies
+/// `started_unix_secs` from the trace and never calls this. Reading through the
+/// port anyway leaves exactly one spelling of "seconds since the epoch" in the
+/// crate, which is what stops the next wall-clock read growing back here.
 pub(crate) fn unix_now_secs() -> i64 {
-    std::time::SystemTime::now()
-        .duration_since(std::time::UNIX_EPOCH)
-        .map(|d| d.as_secs() as i64)
-        .unwrap_or(0)
+    SystemClock.now_unix_secs()
 }
 
 #[cfg(test)]
--- a/crates/stella-cli/src/memory/learning.rs
+++ b/crates/stella-cli/src/memory/learning.rs
@@ -71,6 +71,11 @@ impl SessionMemory {
             &self.domains.names(),
             succeeded,
             budget_limit,
+            // The session's clock, so a replayed turn stamps its lessons with
+            // the trace's instant rather than today's (#2320). Negative (i.e.
+            // pre-epoch) readings clamp to 0, matching the `unwrap_or(0)` the
+            // wall-clock read this replaced already did.
+            u64::try_from(self.clock.now_unix_secs()).unwrap_or(0),
         )
         .await
         {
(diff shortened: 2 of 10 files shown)


macanderson:
## ⚠️ Collision: #2340 is solving the same problem, in the same files

**#2340** ("route the learning loop's wall-clock reads through the Clock port that already existed") also declares `Closes #2320`, also threads the `Clock` port through `SessionMemory`, and edits the same seam: `crates/stella-cli/src/memory.rs`, `memory/learning.rs`, `memory/reflection.rs`, `memory/tests.rs`. Neither PR references the other.

This is the clobber shape that has bitten this repo before: two branches editing the same file from bases that predate each other, git reporting no conflict because one side simply does not contain the other's lines. **One should be closed before either merges** — a clean textual merge here would silently drop one side's work, and `check-deleted-tests` would only catch the part that happens to be tests.

Scope differs, and this PR looks like the intended path: #2340 is 367 lines across 6 files (clock threading only), while this is 1397 across 14 and also builds the replay harness (`memory/replay/*`, `tune_cmd.rs`, `stella-context/src/clock.rs`) as the first of five PRs for epic #2304. If that is right, close #2340 — but lift two of its findings first, because they are worth keeping regardless of which branch survives:

1. **`retire_failing_context` was a defect on its own terms, not just an untestable one.** Its `now` is the instant the truth sweep judges TTLs against, so while it read the wall clock **no test could place a record past its expiry** — retirement was unobservable. Whichever branch lands should carry a test that proves retirement actually fires, not merely that the clock is injectable.
2. **`edge.rs:121` is `process_nonce()`, and its consequence belongs in this PR's design notes.** It is pid ⊕ first-use subsec nanos ⊕ the static's ASLR address, folded into edge-id minting so the process-local `EDGE_SEQ` cannot collide across processes, backing schema v10's `UNIQUE` index on `edge.public_id`. It needs no change — it is a nonce, never ordered or compared — but it means **edge `public_id`s are not reproducible across processes, so the replay harness must not assert on them.** That is a constraint on #2304, and it should be written down where the harness authors will read it rather than discovered by a failing replay.

Also worth adopting from #2340 if this branch does not already do it: stamping `occurred_at` in `reflect_and_record` rather than threading a `now` parameter through `parse_lessons_checked` keeps the parser pure and avoids an `#[allow(clippy::too_many_arguments)]` on `reflect_on_turn`.

I have not reviewed this PR's substance yet — flagging the collision first because it is the time-sensitive part.

macanderson:
## The collision landed: #2340 merged (d83771b6), and this PR is now DIRTY

As flagged above, #2340 took the clock threading and merged first, so this branch now conflicts on the shared seam (`memory.rs`, `memory/learning.rs`, `memory/reflection.rs`, `memory/tests.rs`). The rebase is the whole remaining risk here, and it is the risk this repo has been bitten by most:

**Rebase onto main and take main's version of the clock threading, then re-apply only what is uniquely yours** — the replay harness (`memory/replay/*`), `tune_cmd.rs`, `stella-context/src/clock.rs`, and the tests that exercise them. Resolving in the other direction re-lands your pre-#2340 version of the shared files and silently reverts d83771b6, which git will report as a clean merge because your side simply does not contain main's lines.

Three concrete things to verify after the rebase, each of which #2340 established:

1. **`retire_failing_context` reads the injected clock**, and a test places a record past its expiry and proves retirement fires. This was the real defect underneath #2320: while that `now` came from the wall clock, retirement was *unobservable*, not merely untested.
2. **`occurred_at` is stamped in `reflect_and_record`, not threaded through `parse_lessons_checked`.** #2340 chose this deliberately — it keeps `parse_lessons_checked` a pure parser and avoids an `#[allow(clippy::too_many_arguments)]` on `reflect_on_turn`. If your branch threads a `now` parameter instead, prefer main's shape.
3. **`unix_now_secs()` stays**, since `SessionMemory` no longer calls it and its twelve remaining callers are drivers that own no clock. Worth re-checking against what the replay harness actually needs, since that is your PR's question rather than #2340's.

One finding from #2340 that belongs in this PR's design notes regardless: **`edge.rs:121` is `process_nonce()`** — pid ⊕ first-use subsec nanos ⊕ the static's ASLR address, folded into edge-id minting so the process-local `EDGE_SEQ` cannot collide across processes, backing schema v10's `UNIQUE` index on `edge.public_id`. It correctly needs no change, but the consequence constrains epic #2304: **edge `public_id`s are not reproducible across processes, so the replay harness must not assert on them.** Better written down here than discovered by a failing replay later.

Since #2340 already closed #2320, this PR's description should drop its `Closes #2320` and use `Refs #2320` instead — otherwise the two compete for the same closing link.

Run `check-deleted-tests` mentally on the result too: if the rebase drops any test #2340 added to those files, the guard will catch it on push, and the fix is to restore the lines rather than name the deletion.

macanderson:
Closing as superseded: #2340 merged the same clock seam 20 minutes after this opened, closed #2320, and moved main past this branch (hence the conflict). Rebasing would mean re-litigating judgment calls #2340 already decided differently (nonce injection vs. the `default_task_id`/`deterministic_task_id` split; parser-parameter stamping vs. stamping in `reflect_and_record`), for no behavioral delta.

Accounting for everything on this branch that #2340 did *not* land, so nothing is stranded:

1. **The `edge.rs:121` audit** (spec §9 assigns it to PR 1; #2340 never touched `edge.rs`) — salvaged verbatim as #2348, including the `EDGE_SEQ` finding that constrains PR 4: `public_id` is never admissible in byte-compared replay summaries; compare edge content, not identity.
2. **`self_tuning.rs`/`tune_cmd.rs` clock threading + `Clock::now_unix_millis`** — already tracked as site 2 of #2330's seven-site census; left a pointer there to this branch's `bb7f4772` as a reference implementation.
3. **The WIP replayer commit `a7b7e266`** (trace format + replayer + summary, ~920 lines, empty test module) — PR-2 scope that leaked into this PR's file list. Noted on epic #2304 with its location; the branch is kept for whoever picks PR 2 up.

Do not delete `fix/2320-session-memory-clock-seam` until PR 2 lands or deliberately discards `a7b7e266`.

macanderson:
Superseded by #2340, which merged the same clock seam from a parallel session. Closing is right — two PRs closing #2320 is exactly the duplicate this repo has been bitten by before.

Recording the two places this branch went further, so neither is lost. Both are folded into #2350's rebase onto `main`, where the harness actually needs them.

**1. `open_with_clock` hardcodes `include_workspace_skills: false`, and the harness needs `true`.** With it false, `write_candidates` and `induce_rules` both decline to write a FILE (#737) while still recording proposals — so a replay of an entire corpus behaves perfectly correctly and builds **nothing**. Measured, not theorised: it is the difference between the committed corpus producing 2 skills and 1 rule, and producing zero. #2350 widens the constructor and says why.

**2. The witness has to pin the instant, not just compare two runs.** #2340's own follow-up commit reached this independently, which is reassuring. Confirmed empirically here — with the old behaviour restored behind the new seam:

```
test a_session_records_its_clocks_instant_not_todays ... FAILED
  assertion `left == right` failed: occurred_at must come from the session clock
    left: 1786222713
   right: 1700000000
test two_runs_of_one_scripted_turn_are_byte_identical ... ok      <-- passes on the OLD behaviour
```

Two opens milliseconds apart share a wall-clock second and a pid, so byte-identity alone passes on `main` by luck. Both assertions are worth having; only the first is a witness.

**The `edge.rs:121` audit #2320 asked for**, since #2340's description does not mention it: it is an **identity nonce, not a time source**. `process_nonce()` hashes into `edge.public_id`, which nothing orders, ranges over, or compares — the row's *time* is the separate `recorded_at` column the port already governs. Putting a fixed clock behind it would only make two processes in one second mint from the same keyspace, which is the collision it exists to prevent.

The real determinism hazard sitting beside it is different and was not anticipated: **`EDGE_SEQ` is a process-global counter**, so two replays of one trace *inside a single process* mint different ids for the same logical edge. An edge `public_id` is therefore never admissible in a replay summary compared for byte-identity — compare edge content, not edge identity. #2350 carries that annotation at the site so the next reader does not re-audit it.

PR #2340, opened 2026-08-08
fix(stella-cli): route the learning loop's wall-clock reads through the Clock port that already existed

### LINKED ISSUE #2320: cli: the learning path bypasses the Clock port and reads the wall clock in six places — retirement has never been testable

macanderson:
## What & why

`stella-context` has shipped a `Clock` port — `SystemClock`, `FixedClock` with `advance`/`set` — since the bi-temporal store landed, and `ContextStore::open_and_warm` has always taken one. The CLI's learning loop routed around it and called `SystemTime::now()` directly. This threads the port through `SessionMemory` so every timestamp the loop writes comes from one injected source.

Closes #2320.

### The five reads, and what each cost

| Site | Now reads |
|---|---|
| `memory.rs::default_task_id` | `clock.now_unix_secs()` |
| `memory.rs::record_episode` | `self.clock.now_unix_secs()` |
| `memory/learning.rs::retire_failing_context` | `self.clock.now_rfc3339()` |
| `memory/reflection.rs::parse_lessons_checked` | nothing — stamping moved (below) |
| `open_with_workspace_skills` → `ContextStore` | the caller's clock, not a hardcoded `SystemClock` |

`retire_failing_context` is the one that was a defect on its own terms: its `now` is the instant the truth sweep judges TTLs against, so while it read the wall clock **no test could place a record past its expiry**. Retirement was unobservable, not merely untested.

### Three judgment calls worth reviewing

**1. The pid stays in `default_task_id`.** The issue said to drop it. I disagree and kept it, because it is not a time source and removing it trades a rare governance under-count (two processes starting in the same second mint one id, and governance counts *distinct tasks*) for reproducibility on a path that is never replayed. The determinism lives in a separate `deterministic_task_id` used only by the new constructor, so the nondeterminism is confined to the constructor that wants it. Happy to drop it if you'd rather.

**2. `occurred_at` is stamped in `reflect_and_record`, not the parser.** My first pass threaded a `now: u64` through `reflect_on_turn` → `parse_lessons_checked`, which pushed `reflect_on_turn` to 8 arguments and tripped `clippy::too_many_arguments`. The repo would have accepted a justified `#[allow]` (`subsession.rs:531` is the precedent) — but `reflect_and_record` **already** loops over the lessons to stamp `task_id`, and `occurred_at` answers the same question. Stamping both there removes the parameter entirely: no `#[allow]`, both signatures unchanged from `main`, a smaller diff, and `parse_lessons_checked` becomes a pure parser that reads no clock at all. Better than the lint suppression I set out to write.

**3. `unix_now_secs()` stays as-is.** It is `pub(crate)` and called from twelve places in `agent.rs`, `agent/goal.rs` and `command_deck.rs` — drivers that own no clock and use it to compute the `started_unix_secs` they pass *into* `record_episode`. `SessionMemory` itself no longer calls it, so it is not part of the loop's determinism problem: the one value still arriving from outside rides in as a parameter a replayer controls. Documented at the definition rather than churned through two god files. Flagged below.

### The `edge.rs:121` audit the issue asked for

Resolved, no change needed. It is `process_nonce()` — pid ⊕ first-use subsec nanos ⊕ the static's ASLR address, folded into edge id minting so the process-local `EDGE_SEQ` cannot collide across processes, backing schema v10's `UNIQUE` index on `edge.public_id`. It is a nonce, never a comparable or ordered field, and it already says so in its own doc comment. Consequence worth knowing for #2304: edge `public_id`s are not reproducible across processes, so a replay harness must not assert on them.

## The witness

- [x] Includes a witness test (fails on `main`, passes here) — `crates/stella-cli/src/memory/tests/determinism.rs`, four tests.

**Verified the artisanal way, and the result changed the tests.** Rather than only checking that they fail on `main` (where they would fail merely because `open_with_clock` is absent — proving the API is new, not that the behaviour is fixed), I reverted the three internals *behind their unchanged signatures* and re-ran.

The first round found a real problem with **the witness the issue specified**:

```
two_sessions_at_one_fixed_clock_write_byte_identical_mining_logs ... ok      ← on the BROKEN code
the_mined_lesson_is_stamped_with_the_injected_instant ...        FAILED
a_later_clock_stamps_a_later_lesson ...                          FAILED
the_context_store_shares_the_session_clock ...                   FAILED
```

Byte-equality between two sessions **passes on the unfixed code**. Both run in one process, so a wall-clock `occurred_at` and a pid-bearing task id agree with each other whenever the two runs land in the same second — which is nearly always. It would only have failed on a run that straddled a second boundary. That is a flake, not a witness, and #2320 specified it as the witness. I have said so on the issue.

So the equality test now also pins the *value*, and all four fail without the fix:

```
two_sessions_at_one_fixed_clock_write_byte_identical_mining_logs ... FAILED
the_mined_lesson_is_stamped_with_the_injected_instant ...            FAILED
a_later_clock_stamps_a_later_lesson ...                              FAILED
the_context_store_shares_the_session_clock ...                       FAILED
test result: FAILED. 0 passed; 4 failed
```

and all four pass on this branch. The probe was re-run against the *final* code shape after the argument refactor, not the first draft. `a_later_clock_stamps_a_later_lesson` exists so the fix cannot pass by freezing time into a constant.

## The gate

- [x] `make gate CARGO_SCOPE="-p stella-cli"` — **exit 0**, all 25 steps, including `clippy -D warnings`, rustdoc `-D warnings`, `test --workspace` (1493 passed), and the self-driving shell harness (60 checks).
- [x] `cargo fmt --check`
- [x] Docs updated — doc comments at every changed seam; no flags or behaviour changed for users.
- [x] `Closes #2320` in this description **and** as a commit trailer.

Two things the gate caught that are worth naming rather than burying:

- **rustdoc**: I linked `[`deterministic_task_id`]` from a public doc comment while the item is `#[cfg(test)]`, so it does not exist in the documented build. Now plain text with a note saying why.
- **One existing assertion changed, deliberately.** `memory::tests::parse_lessons_drops_invented_domains_and_caps_at_three` asserted `lessons[0].occurred_at > 0` — true only because the parser read the wall clock, which is the bug. It now asserts `occurred_at == 0`, pinning the new boundary: parsing assigns no instant, the session does. No test was deleted or renamed; the assertion was inverted on purpose and the reason is in a comment beside it.

## Nothing left behind

- **`self_tuning.rs::now_ms` is still a wall-clock read** and lives under `crates/stella-cli/src/memory/`, so #2320's literal definition of done ("no `SystemTime::now()` remains in `crates/stella-cli/src/memory/`") is **not** fully met by this PR. Deliberate: it is reached only from `tune_cmd.rs`, never from `SessionMemory`, so it is a separate surface with its own witness. It is tracked as site 2 of **#2330** (the seven hand-rolled `now_ms` copies). Calling this out rather than quietly claiming the DoD.
- **`unix_now_secs`'s twelve driver call sites** stay on the ambient clock — see judgment call 3. Threading a clock into `agent.rs`/`command_deck.rs` (both god files, closed to growth) is its own change.
- **#2330** covers the same shape elsewhere: `ports::Clock::now_ms` re-implemented by hand in seven places, including inside `stella-core`.
- Unblocks epic **#2304** / `doc:trace-replay-learning-harness` PR 1 of 5. `open_with_clock` and `deterministic_task_id` are `#[cfg(test)]`-gated with a comment to drop both gates in the commit that lands the replayer.

## Summary by Sourcery

Route all learning loop timestamps through a session-level Clock, making stella-cli’s SessionMemory time behavior reproducible and testable while preserving existing CLI behavior.

New Features:
- Introduce a test-only SessionMemory::open_with_clock constructor that accepts an injected Clock for deterministic sessions.
- Add a deterministic_task_id helper to derive reproducible task IDs from the injected clock.

Bug Fixes:
- Fix retire_failing_context to use the session clock rather than the wall clock so TTL-based retirement can be observed and tested.
- Ensure mined lessons’ occurred_at timestamps are sourced from the session clock, eliminating nondeterministic wall-clock reads in the learning loop.

Enhancements:
- Thread a Clock through SessionMemory and ContextStore so both the mining log and bi-temporal store share a single time source.
- Refine task ID generation to separate nondeterministic default_task_id from deterministic_task_id used in replay scenarios.
- Make reflection and lesson parsing pure with respect to time by moving timestamp stamping into reflect_and_record.
- Document the intended use of unix_now_secs as an ambient clock helper and the role of the session clock for determinism.

Tests:
- Add determinism tests that validate replayable mining logs, correct stamping of injected instants, differing timestamps for differing clocks, and shared clock usage between SessionMemory and ContextStore.
- Adjust an existing parser test to assert that lessons parsed from reflections start with occurred_at = 0, confirming that timestamp assignment is now a session concern.

### DIFF
--- a/crates/stella-cli/src/memory.rs
+++ b/crates/stella-cli/src/memory.rs
@@ -194,7 +194,9 @@ impl LessonKind {
 /// second from minting one id, which governance would then count as a single
 /// task. Dropping it would buy reproducibility on a path that is never
 /// replayed, at the price of under-counting concurrent work. The deterministic
-/// half lives in [`deterministic_task_id`] instead, so the nondeterminism is
+/// half lives in `deterministic_task_id` instead (not a doc link: it is
+/// test-gated, so it does not exist in the documented build), so the
+/// nondeterminism is
 /// confined to the constructor that wants it rather than shared by both.
 fn default_task_id(clock: &dyn Clock) -> String {
     format!("session:{}-{}", clock.now_unix_secs(), std::process::id())
--- a/crates/stella-cli/src/memory.rs
+++ b/crates/stella-cli/src/memory.rs
@@ -29,7 +29,7 @@ use std::path::{Path, PathBuf};
 use colored::Colorize;
 use serde::{Deserialize, Serialize};
 use stella_context::{
-    ContextDelta, ContextStore, DomainInput, EpisodeInput, EpisodeOutcome, FactAssertion,
+    Clock, ContextDelta, ContextStore, DomainInput, EpisodeInput, EpisodeOutcome, FactAssertion,
     HashEmbedder, NodeInput, NodeKind, RecallTier, SystemClock, format_rfc3339,
 };
 use stella_core::skills::{self, SelectionConfig, Skill};
@@ -188,12 +188,31 @@ impl LessonKind {
 /// One `stella run` is one task, so for the headless path this is exactly the
 /// right boundary. For a long REPL session it is an approximation, but a
 /// strictly better one than per-turn.
-fn default_task_id() -> String {
-    let secs = std::time::SystemTime::now()
-        .duration_since(std::time::UNIX_EPOCH)
-        .map(|d| d.as_secs())
-        .unwrap_or(0);
-    format!("session:{secs}-{}", std::process::id())
+///
+/// The pid stays (#2320). It is not a time source and it is not a task
+/// boundary — its only job is to keep two processes that start in the same
+/// second from minting one id, which governance would then count as a single
+/// task. Dropping it would buy reproducibility on a path that is never
+/// replayed, at the price of under-counting concurrent work. The deterministic
+/// half lives in [`deterministic_task_id`] instead, so the nondeterminism is
+/// confined to the constructor that wants it rather than shared by both.
+fn default_task_id(clock: &dyn Clock) -> String {
+    format!("session:{}-{}", clock.now_unix_secs(), std::process::id())
+}
+
+/// The task identity [`SessionMemory::open_with_clock`] mints: derived from the
+/// injected clock alone, so two sessions opened at the same instant agree.
+///
+/// Deliberately carries no pid. A caller reaching for the deterministic
+/// constructor is replaying a recorded engagement, where "which process am I"
+/// is not a question with an answer — and where two sessions minting the same
+/// id at the same instant is the property under test, not a collision.
+///
+/// Gated alongside its only caller, [`SessionMemory::open_with_clock`] — drop
+/// both gates together when the trace-replay harness lands (#2304).
+#[cfg(test)]
+fn deterministic_task_id(clock: &dyn Clock) -> String {
+    format!("session:{}", clock.now_unix_secs())
 }
 
 mod reflection;
@@ -270,6 +289,21 @@ pub struct SessionMemory {
     /// re-deriving it here would re-walk the rule directories and re-run the
     /// truth sweep on every turn.
     record_registry: Option<stella_core::records::Registry>,
+    /// The session's time source (#2320) — the only one the learning loop is
+    /// allowed to read.
+    ///
+    /// `stella-context` has shipped this port since the bi-temporal store
+    /// landed, and [`ContextStore::open_and_warm`] has always taken one; the
+    /// learning loop simply routed around it and called `SystemTime::now()` in
+    /// five places. That made every write here non-replayable, and one of the
+    /// five was [`SessionMemory::retire_failing_context`], whose `now` is the
+    /// instant the truth sweep judges TTLs against — so retirement could not be
+    /// tested at all, because the sweep always saw today.
+    ///
+    /// Production passes [`SystemClock`] through every existing constructor, so
+    /// this changes no shipped behaviour; it is the seam that lets a test
+    /// advance time on purpose.
+    clock: std::sync::Arc<dyn Clock>,
 }
 
 impl SessionMemory {
@@ -364,6 +398,48 @@ impl SessionMemory {
         warn: bool,
         include_workspace_skills: bool,
     ) -> Option<Self> {
+        let clock: std::sync::Arc<dyn Clock> = std::sync::Arc::new(SystemClock);
+        let task_id = default_task_id(clock.as_ref());
+        Self::open_inner(
+            workspace_root,
+            warn,
+            include_workspace_skills,
+            clock,
+            task_id,
+        )
+    }
+
+    /// Open a session whose every timestamp — the stamps on mined lessons, the
+    /// episode clock, the instant the truth sweep judges TTLs against, and the
+    /// default task id — comes from `clock` rather than the wall clock (#2320).
+    ///
+    /// Two sessions opened this way at the same instant write byte-identical
+    /// logs, which is what makes the learning loop replayable and what lets a
+    /// test place itself on either side of a TTL.
+    ///
+    /// **Test-gated until the trace-replay harness lands** (epic #2304,
+    /// `doc:trace-replay-learning-harness` §4). `stella-cli` is a bin-only
+    /// crate, so no external consumer could call this even in principle, and
+    /// leaving it on the production build would be dead code. Drop the gate in
+    /// the same commit that adds the replayer — the same discipline
+    /// [`SessionMemory::set_task_id`] is held to.
+    #[cfg(test)]
+    pub(crate) fn open_with_clock(
+        workspace_root: &Path,
+        warn: bool,
+        clock: std::sync::Arc<dyn Clock>,
+    ) -> Option<Self> {
+        let task_id = deterministic_task_id(clock.as_ref());
+        Self::open_inner(workspace_root, warn, false, clock, task_id)
+    }
+
+    fn open_inner(
+        workspace_root: &Path,
+        warn: bool,
+        include_workspace_skills: bool,
+        clock: std::sync::Arc<dyn Clock>,
+        task_id: String,
+    ) -> Option<Self> {
         // Ephemeral benchmark trials must neither recall task/user-planted
         // learning state nor create or migrate a context database that can
         // perturb the task under test. Reflection is separately pinned off
@@ -379,7 +455,7 @@ impl SessionMemory {
         match ContextStore::open_and_warm(
             &db_path,
             std::sync::Arc::new(HashEmbedder::default()),
-            std::sync::Arc::new(SystemClock),
+            clock.clone(),
         )
         .map(|store| store.with_tuning(retrieval.tuning()))
         {
@@ -408,8 +484,9 @@ impl SessionMemory {
                     ab_turn: 0,
                     // Phase 3 (#714)
                     lifecycle_enabled: tuning::session_lifecycle_enabled(workspace_root),
-                    task_id: default_task_id(),
+                    task_id,
                     execution_id: None,
+                    clock,
                 })
             }
             Err(e) => {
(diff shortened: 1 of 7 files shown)


macanderson:
Added a fifth commit (`9ad82dae`): the same weak-assertion form was in `doc:trace-replay-learning-harness` §5 assertion 8, which I wrote in #2318 — it said "two replays produce byte-identical summaries" with no requirement to pin the instant. That is the exact form this PR just measured passing against broken code, so it is corrected in the same breath rather than left for the harness to be built on. Docs-only; `make guards-fast` green (exit 0), no Rust touched after the full-gate run.

macanderson:
## ⚠️ Collision: #2335 is solving the same problem, in the same files

**#2335** ("the learning plane reads an injected clock — replayable sessions, and a testable truth sweep") also declares `Closes #2320`, also threads the `Clock` port through `SessionMemory`, and edits the same seam: `crates/stella-cli/src/memory.rs`, `memory/learning.rs`, `memory/reflection.rs`, `memory/tests.rs`. Neither PR references the other.

This is the clobber shape this repo has been bitten by repeatedly: two branches editing the same file from a base that predates the other, git reporting no conflict because one side simply does not contain the other's lines. Merging both will either conflict or silently drop one side's work — and if it lands as a clean textual merge, `check-deleted-tests` is the only thing that would notice, and only for tests.

**One of these should be closed before either merges.** They are not the same size or scope:

- **#2340** (367 lines, 6 files) is the focused clock threading, plus the `edge.rs:121` audit the issue asked for.
- **#2335** (1397 lines, 14 files) does the clock threading *and* builds the replay harness (`memory/replay/*`, `tune_cmd.rs`, `stella-context/src/clock.rs`), positioned as the first of five PRs for epic #2304.

If #2335 is the intended path, close this one and lift its two findings into that PR — they are worth keeping either way:

1. **`retire_failing_context` was a defect on its own terms.** Its `now` is the instant the truth sweep judges TTLs against, so while it read the wall clock **no test could place a record past its expiry** — retirement was unobservable, not merely untested. Whichever PR survives must keep a test that proves retirement now fires.
2. **`edge.rs:121` is `process_nonce()`** — pid ⊕ first-use subsec nanos ⊕ the static's ASLR address, folded into edge-id minting so the process-local `EDGE_SEQ` cannot collide across processes, backing schema v10's `UNIQUE` index on `edge.public_id`. It is a nonce, never ordered or compared, so it needs no change — but the consequence matters for #2304: **edge `public_id`s are not reproducible across processes, so a replay harness must not assert on them.** That belongs in #2335's design notes.

On this PR's own three judgment calls, for whoever inherits them: keeping the pid in `default_task_id` is defensible (it is not a time source, and the determinism lives in the separate `deterministic_task_id`); stamping `occurred_at` in `reflect_and_record` rather than threading a `now` through the parser is the better of the two options, since it avoids the `#[allow(clippy::too_many_arguments)]` and leaves `parse_lessons_checked` a pure parser; and leaving `unix_now_secs()` in place is fine given `SessionMemory` no longer calls it, though that reasoning should be reviewed against #2304's needs rather than settled here.

Issue #2321, opened 2026-08-08
cli: the learning loop is trapped in a bin-only crate — extract it so anything but an in-crate test module can drive it


macanderson:
## Problem

`stella-cli` is **bin-only** — its `Cargo.toml` declares `[[bin]]` and no `[lib]`. Consequently every one of the thirteen files under `crates/stella-cli/tests/` drives the *compiled binary* (spawning `stella` and asserting on its output), and none of them can reach the crate's own types.

The learning loop lives entirely inside that bin-only crate:

- `SessionMemory` and `reflect_and_record` — `crates/stella-cli/src/memory.rs`, `crates/stella-cli/src/memory/learning.rs`
- observation extraction — `crates/stella-cli/src/memory/observations.rs`
- rule induction — `crates/stella-cli/src/memory/rules_mining.rs`
- the foundry adapter — `crates/stella-cli/src/tool_foundry.rs`

So anything that wants to drive the learning loop programmatically — the trace-replay harness of epic #2304, a future fleet-side learner, an external host — has exactly one option today: be an in-crate `#[cfg(test)]` module. That is what `doc:trace-replay-learning-harness` §4 specifies, deliberately, and it is the right call for that epic. It is not the right end state.

Consequences of the current shape:

- Harness code that is genuinely production-shaped (a trace loader, a replayer, a metrics summariser) has to live under `#[cfg(test)]` or be dead code in the shipped binary.
- The learning loop cannot be exercised by any other crate, including `stella-fleet`, `stella-serve`, or an external host embedding the engine.
- The existing in-crate test modules under `crates/stella-cli/src/memory/tests/` are the only place this logic is testable at all, which concentrates coverage in one crate's private surface.

## What to do

Extract the learning loop into a library crate — working name `stella-learn` — leaving `stella-cli` as the wiring that constructs it from settings and workspace paths.

The boundary to aim for, consistent with the workspace's existing shape:

- The crate owns `SessionMemory` and the four learners, and depends on `stella-context`, `stella-core`, `stella-store`, `stella-protocol`.
- It reads **no ambient environment** — the pattern `crates/stella-runtime/tests/no_ambient_reads.rs` already enforces for `stella-runtime`. Workspace root, settings and the clock are all injected.
- The model boundary stays a port (`Provider`), as it already is.
- `stella-cli` keeps the CLI surface: flag parsing, output formatting, and the construction that turns resolved settings into a `SessionMemory`.

[...]

PR #2350, opened 2026-08-08
feat(stella-cli): the trace-replay learning harness — the whole learning loop, driven from recorded traces, with zero model calls (#2304)

macanderson:
Refs #2304. Implements PRs **2, 3 and 4** of `doc:trace-replay-learning-harness` §9 — the trace format and loader, the replayer, the eight assertions, the committed simulated-months corpus, the metrics, and `make replay-learning`.

**Rebased onto `main`.** The clock seam this depends on landed as **#2340** (closing #2320) from a parallel session while this was in flight; my duplicate #2335 is closed, with the two things it carried that #2340 did not recorded [there](https://github.com/macanderson/stella/pull/2335#issuecomment-5228386228) and folded in here. Base is `main`; no stack.

> **Note on `main` being red.** `make shellcheck` fails on `scripts/test-arena-scripts.sh`, which arrived with #2351 and is tracked as **#2355**. Not touched here — per AGENTS.md, a red ratchet you did not cause does not get folded into an unrelated PR.

## The spec's central claim holds, and cost nothing to collect

Of the five learners, four already make **no** model call and the fifth crosses exactly one port. So the harness is one scripted `Provider` and a clock — invariant 1 ("ports, not concretions") paying out. There is no new architecture here and no parallel implementation of anything.

**The replayer's loop is shorter than §4.1's, and that is a finding.** The spec gives the replayer observation extraction (step 5) and rule induction (step 6) as its own calls. They are already *inside* `reflect_and_record` — `auto_create_skills_typed` runs uses extraction, the retirement sweep, observation extraction, skill proposals and rule induction from one observation pool. Calling them again would not corrupt anything (extraction is replay-idempotent via content-derived ids) but it would be a second mechanism shadowing the shipped one, and the harness would then certify its own wiring instead of the loop's. The replayer drives the seam the drivers drive and lets the loop do the rest.

## Four things replaying the real loop measured, each of which changed the fixtures

These are the payoff. None is stated by any single module, and every one of them would have produced a *confidently wrong* fixture.

**1. The dedup filter and the skill miner use the same metric at the same threshold, so a corpus of literal repetitions builds nothing.** `retain_unknown` drops a lesson at Jaccard `>= 0.5` against existing memories, and it runs *before* the mining log is appended — so a dropped lesson never becomes an observation and can never contribute an occurrence. The miner then clusters at `min_similarity: 0.5`. For a cluster of three to form, three lessons must be similar enough to cluster while staying dissimilar enough to survive dedup.

That band exists **only because the tokenizers differ**, and the difference is load-bearing:

| | `stella_store::forget::tokens` | `stella_core::mining::terms` |
|---|---|---|
| `crates/stella-tools/src/registry.rs` | **one** token | **five** terms |
| stopwords, words ≤2 chars | kept | dropped |

A shared *path* costs the dedup comparison one token and gives the miner five, while differing filler inflates the dedup union invisibly to the miner. That is also the shape a real engagement produces — the same file re-encountered in different words — so the corpus is written that way. My first probe used four naturally-varied phrasings and mined **zero** candidates; it reads as a broken learner until you know this.

**2. The foundry holes only a *value-like* argument.** `classify_argument` makes a positional argument a parameter only if it is a path or a number (or quoted, or after `=`); a bareword stays in the skeleton. So `cargo test -p <crate>` yields a different signature per crate and never clusters, while `rg -n "<pattern>" <path>` clusters immediately. Correct behaviour — a bareword is usually a subcommand — but a fixture varying a bareword measures nothing.

**3. A record's lineage embeds the workspace directory name.** `derive_set_id` falls back to it with no git remote, so two replays into `/tmp/.tmpQ7x1a9` and `/tmp/.tmpB3k0zz` produced `ctx.tmpq7x1a9.…` and `ctx.tmpb3k0zz.…` for the same learned rule, and assertion 8 failed. Correct behaviour again — a record's lineage is scoped to its workspace. The tempting fix was to strip the set id out of the summary; that hides real nondeterminism behind a narrower report. Naming the replay workspace instead makes the harness deterministic *and* keeps the full lineage id — what a teammate actually sees — inside what assertion 8 compares.

**4. `Path::starts_with` is lexical, so the obvious escape guard passes a fixture that escapes.** `root/../escaped.txt` starts with `root` by that test. The guard now walks components. Caught by its own test, which is the argument for writing the negative control first.

## The eight assertions (§5), each with its negative control

| # | Assertion | Negative control |
|---|---|---|
| 1 | Recurrence across ≥3 distinct tasks builds a skill and a rule | The **same lessons at the same instants** inside one task build neither |
| 2 | An auto-activated rule is advisory — `force = "should"`, no `[enforcement]` | — |
| 2b | Auto-activation never clobbers a hand-written record (#737) | — |
| 3 | A repeated skill-shaped lesson yields `SKILL.md` | Four unrelated one-offs yield none |
| 4 | A shape with ≥2 distinct argument sets mints a tool | The identical command three times mints none |
| 5 | A tombstoned lesson is not re-learned **from a paraphrase** | — |
| 6 | A memory whose anchors all vanished lands in the retired set | — |
| 7 | An unreadable corpus builds nothing **and says why** | An *empty* lessons array is not reported as starvation |
| 8 | Two replays of one trace are byte-identical | — |

Two are worth calling out.

**Assertion 5 paraphrases, deliberately.** A corpus that tombstones a lesson and re-emits it *verbatim* proves almost nothing: byte-identical content already collapses through content-hash lineage. What needs proving is `retain_unforgotten`'s restatement matching, so the fixture restates.

**Assertion 6 asserts against `retirement::retired_ids`, not against the store's live nodes** — and getting this wrong first is what taught it. Retirement deliberately does **not** write `node.superseded_at`, because that column is filtered by `node_by_public_id` and projecting retirement onto it would make a retired record impossible to fetch by id, when staying retrievable is exactly what separates retirement from forgetting. A test asserting "the memory is gone" fails against a perfectly working sweep; "the count went down" quietly re-specifies retirement as deletion. The test asserts the memory is **still stored** *and* **in the set recall suppresses**.

## Two departures from the spec's §3 sketch

Both are the fixture format deliberately not being the runtime type, so a trace cannot express a state the live loop could never reach.

- `ShellInvocation` derives no serde impls, and adding them would put a fixture concern in the engine. `TraceShell` is the two-field mirror.
- `ReflectionLesson` carries `occurred_at` and `task_id`, which the **replayer** owns. A trace able to set them could contradict its own clock, and every assertion about recency or distinct-task counting would then be asserting the fixture. `TraceLesson` carries only what the model would have said.

`ScriptedReflection` also carries the §7.2 `provenance` field from the start (defaulting to `Recorded`), so PR 5's adapter can label derived lessons without a format version bump.

## Two fields the spec did not have

`TraceTurn::forget` and `TraceTurn::removed_files`, both applied before the turn's reflection. Assertions 5 and 6 are unreachable without them — there is no way to tombstone a lesson or delete a seeded file mid-trace otherwise — and both model real events in a real engagement rather than being test backdoors.

## Metrics, and the honesty guard

`make replay-learning` prints the §6 report. The recurrence profile is emitted by the **same function** as the artifact counts, so a caller cannot print the yield without the shape it came from:

```
what the learners built
  memories: 9 distinct in store (9 lesson(s) recorded of 9 offered)
  skills:   2 — .stella/skills/registered-tools-come-from-crates-stella-b54d480a.md, …
  rules:    1 — ctx.stella-replay-workspace.registered-tools-come-from-crates-stella-b54d480a
  tools:    1 — rg_n
  turns-to-first: memory 1, skill 10, tool 3

read the two together
  4 artifact(s) from 9 lesson(s) over 5 distinct task(s). A count that rises
  because one lesson repeats is a fact about this fixture, not about the learner.

1 turn(s) taught the lifecycle nothing and said so:
  · reflection response was not a readable JSON array of lessons, so this turn taught
    the context lifecycle nothing; …
```

An artifact count that rises because a fixture repeats one lesson measures the fixture, and reporting it alone is the kind of small dishonesty that costs more than it buys.

## CI wiring

**No new `GATE_STEPS` entry**, per §8. The assertions are ordinary in-crate tests, already covered by `make gate`'s existing `test` step, so there is no `gate-parity` churn across the Makefile, AGENTS.md and CONTRIBUTING.md. `make replay-learning` is a convenience target only: a target whose job is to *print a report* adds nothing as pass/fail, and every added step is one more shared cell three documents must agree on.

## Verification

- `cargo test -p stella-cli` — **1524 passed, 0 failed** (18 new), on top of `main`
- `cargo clippy -p stella-cli -p stella-context --all-targets -- -D warnings` — clean
- `make guards-fast` — clean apart from the pre-existing `shellcheck` failure tracked as #2355
- `make replay-learning` — output above

**No tests deleted.**

## Scope notes

- Everything is `#[cfg(test)]`. `stella-cli` is bin-only, so an in-crate module is the only place this can reach `SessionMemory` at all (§4), and gating it means zero production surface and no `#[allow(dead_code)]` describing an API nothing can reach.
- `SessionMemory::set_task_id` **stays `#[cfg(test)]`-gated**. Its doc comment says to drop the gate for the first real caller; a test-only replayer is not one, and the comment's intent — don't ship unreachable API — is satisfied as written. Extracting the learning loop to a library crate so the harness could be an ordinary integration test is #2321.
- **PR 5 (the Claude Code transcript adapter) is not in this PR** and is independent of 3–4 now that the format exists.
- Recall-*ranking* assertions are scoped out per §11: they would inherit #2288.

## What this PR adds to `main`'s clock seam

One line, and it is load-bearing: **`open_with_clock` gains `include_workspace_skills`.** #2340 hardcoded it `false`. With it false, `write_candidates` and `induce_rules` record their proposals and then decline to write a FILE (#737) — so a replay of the whole corpus behaves perfectly correctly and builds **nothing**. It is the difference between the committed corpus producing two skills and a rule, and producing zero.

`main`'s doc comment says to drop the `#[cfg(test)]` gate "in the same commit that adds the replayer". **The gate stays**, because the replayer is itself `#[cfg(test)]` — `stella-cli` is bin-only, so an in-crate module is the only way to reach `SessionMemory` at all (§4), and an ungated constructor nothing production can call would be exactly the dead code the gate exists to prevent. The comment's intent is satisfied; its literal instruction is not, and that is deliberate. #2321 (extracting the learning loop to a library crate) is what would change the answer.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -112,6 +112,16 @@ test-cli: ## Test stella-cli only (the shipping binary)
 test-protocol: ## Test stella-protocol only (shared types)
 	cargo test -p stella-protocol
 
+# The trace-replay learning harness (#2304, `doc:trace-replay-learning-harness`
+# §8). Deliberately NOT a `make gate` step: the assertions are ordinary in-crate
+# tests already covered by the gate's existing `test` step, and a target whose
+# job is to PRINT a report adds nothing as pass/fail — while every added step is
+# one more shared cell the Makefile, AGENTS.md and CONTRIBUTING.md must agree on.
+.PHONY: replay-learning
+replay-learning: ## Replay the synthetic months corpus and print what the learners built (#2304)
+	cargo test -p stella-cli --bin stella \
+	  memory::replay::tests::replay_learning_report -- --exact --nocapture
+
 .PHONY: record-golden
 record-golden: ## Re-record the golden replay trajectories (review the fixture diff!)
 	STELLA_REFRESH_GOLDEN=1 cargo test -p stella-pipeline --lib golden
--- a/crates/stella-cli/src/memory.rs
+++ b/crates/stella-cli/src/memory.rs
@@ -67,6 +67,12 @@ mod private_state;
 mod projection;
 pub(crate) mod proposals;
 mod recall;
+// #2304: the trace-replay learning harness — the learning machinery driven from
+// recorded traces with zero model calls. Test-only by construction: the crate is
+// bin-only, so an in-crate module is the only place this can reach
+// `SessionMemory` at all (`doc:trace-replay-learning-harness` §4).
+#[cfg(test)]
+pub(crate) mod replay;
 // Phase 4 (#715): reversible retirement of context that stops helping.
 pub(crate) mod retirement;
 pub(crate) mod rules_mining;
@@ -419,20 +425,38 @@ impl SessionMemory {
     /// logs, which is what makes the learning loop replayable and what lets a
     /// test place itself on either side of a TTL.
     ///
+    /// `include_workspace_skills` decides whether this session may write into
+    /// the workspace's own skills and rules directories — a *trusted project*,
+    /// in `open_for_session`'s terms. The replayer needs it true, and that is
+    /// not a convenience: with it false, `write_candidates` and `induce_rules`
+    /// both record their proposals and then decline to write a FILE (#737). A
+    /// replay of a whole corpus would behave perfectly correctly and build
+    /// nothing, which reads as a broken learner rather than as a session that
+    /// was never trusted to publish. Measured, not theorised — it is the
+    /// difference between the committed corpus producing two skills and a rule,
+    /// and producing zero.
+    ///
     /// **Test-gated until the trace-replay harness lands** (epic #2304,
     /// `doc:trace-replay-learning-harness` §4). `stella-cli` is a bin-only
     /// crate, so no external consumer could call this even in principle, and
-    /// leaving it on the production build would be dead code. Drop the gate in
-    /// the same commit that adds the replayer — the same discipline
-    /// [`SessionMemory::set_task_id`] is held to.
+    /// leaving it on the production build would be dead code. The harness is
+    /// itself `#[cfg(test)]` for exactly that reason, so the gate stays —
+    /// the same discipline [`SessionMemory::set_task_id`] is held to.
     #[cfg(test)]
     pub(crate) fn open_with_clock(
         workspace_root: &Path,
         warn: bool,
+        include_workspace_skills: bool,
         clock: std::sync::Arc<dyn Clock>,
     ) -> Option<Self> {
         let task_id = deterministic_task_id(clock.as_ref());
-        Self::open_inner(workspace_root, warn, false, clock, task_id)
+        Self::open_inner(
+            workspace_root,
+            warn,
+            include_workspace_skills,
+            clock,
+            task_id,
+        )
     }
 
     fn open_inner(
--- a/crates/stella-cli/src/memory/replay.rs
+++ b/crates/stella-cli/src/memory/replay.rs
@@ -1,0 +1,332 @@
+//! The trace-replay learning harness — drive the learning machinery from
+//! recorded traces, with zero model calls.
+//!
+//! Spec: `doc:trace-replay-learning-harness`. Epic #2304.
+//!
+//! Instead of sessions producing traces that feed learning, recorded traces
+//! drive learning directly, and the harness watches what the agent builds —
+//! memories, skills, rules, tools — across a simulated engagement.
+//!
+//! ## Why this needs no new architecture
+//!
+//! Every learner already sits below the model boundary or behind a port —
+//! invariant 1 ("ports, not concretions") and invariant 2 ("no I/O in the
+//! engine") paying out. Of the five learners, four make **no** model call at all
+//! and the fifth makes exactly one, through the `Provider` port. So the whole
+//! harness is one test double and a clock:
+//!
+//! | Learner | Model calls |
+//! |---|---|
+//! | Memory formation (`reflect_and_record`) | exactly one, via `Provider` |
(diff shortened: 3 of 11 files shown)


macanderson:
Two issues filed from what this PR's own harness measured, per AGENTS.md § "Nothing left behind":

- **#2358** — the big one, and arguably the most valuable thing the harness has produced so far. Reflection dedup runs *before* the mining-log append, and `stella_store::SIMILARITY_THRESHOLD` and `SkillMineConfig::min_similarity` are both Jaccard at 0.5. So a lesson similar enough to cluster is also similar enough to be dropped as already-known, and skills/rules are minable only inside the accidental band left by the two tokenizers disagreeing about paths and stopwords. The committed corpus in this PR is worded around that band **because it had to be**, and `replay/tests.rs`'s module docs explaining how to word a fixture around it are a workaround living in a test file. Naturally-varied phrasings mine zero.
- **#2359** — `turns-to-first` is exact for memories and tools, a placeholder for skills, and absent for rules, while rendering identically to the measured values. Reasoning for shipping it that way is in the code comment; leaving it standing would put a placeholder in a table meant to be quotable.

Issue #2263, opened 2026-08-08
website: a stale package-lock.json beside pnpm-lock.yaml fails every Dependabot run and raises phantom high-severity alerts


macanderson:
## Problem

`website/` carries **two lockfiles** for one project:

```
website/package-lock.json    <- npm, stale
website/pnpm-lock.yaml       <- pnpm, the real one
```

`website/package.json` declares `"packageManager": "pnpm@11.13.0"`, and the only
install in CI is `pnpm install --frozen-lockfile`
(`.github/workflows/docs.yml:59-60`). Nothing in `.github/workflows/`, the
Makefile, or any `vercel.json` references `package-lock.json` or `npm ci`.

Dependabot scans it anyway. Two consequences, both live right now:

**1. The `Dependabot Updates` job fails on every run.** Observed on `aa288fd9`,
`341e092d` and `20e47d9f`:

```
| Type                       | Details                      |
| security_update_not_needed | { "dependency-name": "sharp" }|
Failure running container ...: Command failed with exit code 1
##[error]Dependabot encountered an error performing the update
```

It resolves the advisory against the stale npm lockfile, concludes no update is
needed, and exits non-zero regardless. This is pure red noise in the run list
for `main`, and it will mask a *real* Dependabot failure when one happens.

**2. Phantom high-severity alerts.** `gh api repos/macanderson/stella/dependabot/alerts`
reports open alerts whose manifest is the stale file:

| Severity | Package | Vulnerable range | Manifest |
|---|---|---|---|
| high | postcss | <= 8.5.17 | `website/package-lock.json` |
| high | postcss | <= 8.5.11 | `website/package-lock.json` |
| high | sharp | < 0.35.0 | `website/package-lock.json` |
| medium | postcss | <= 8.5.22, < 8.5.10 | `website/package-lock.json` |

These are what produce the *"GitHub found 2 vulnerabilities on
macanderson/stella's default branch (1 high, 1 moderate)"* banner printed on
every `git push`. If they are not reachable from the pnpm install, the banner is
crying wolf — and a banner everyone has learned to ignore is worse than no
banner, because the two real ones below are indistinguishable from it.

## Not everything here is phantom — check these separately

Two open alerts are on manifests that ARE used, and this issue must not be
closed in a way that sweeps them up:

| Severity | Package | Range | Manifest |
|---|---|---|---|
| **high** | js-yaml | >= 4.0.0, < 4.3.1 | `website/pnpm-lock.yaml` |
[...]

Issue #2121, opened 2026-08-07
deps: two open Dependabot alerts — js-yaml (high, website) and h2 (medium, terminal_bench_analysis)


macanderson:
## Problem

Two Dependabot alerts are open on the default branch and no issue tracks them (searched: no hits for js-yaml / dependabot):

1. **HIGH — `js-yaml` in `website/pnpm-lock.yaml`**: quadratic CPU consumption in `!!omap` resolution (CVE-2026-59870), fix not backported to 3.x/4.x lines. Almost certainly a transitive dependency of the docs/site toolchain — find the dependents with `pnpm why js-yaml` in `website/`, then either bump the parent or add a `pnpm.overrides` entry pinning a fixed version.
2. **MEDIUM — `h2` in `bench/terminal_bench_analysis/uv.lock`**: duplicate Host header could facilitate request smuggling. Bump via `uv lock --upgrade-package h2` in that package and re-run its pytest suite (CI runs `uv sync --locked`, so the lock update must land with the PR).

## Verify

- `gh api repos/macanderson/stella/dependabot/alerts?state=open` returns empty.
- `website`: build green. `bench/terminal_bench_analysis`: `uv sync --locked --extra dev && uv run --no-sync pytest -q` green (this is CI's exact command from `.github/workflows/bench.yml`).

## Constraints

- `deny.toml` / the Rust supply-chain gate are unaffected (both alerts are JS/Python surfaces).
- Neither fix should widen any version range by hand-editing a lockfile — regenerate with the ecosystem's own tool.

macanderson:
Both alerts are resolved in the current lockfiles: `website/pnpm-lock.yaml` carries `js-yaml@4.3.1` (the patched release for CVE-2026-59870) and `bench/terminal_bench_analysis/uv.lock` carries `h2==4.4.1`. These match the exact target versions from commit 234c0523 ("fix(deps): close the two open Dependabot alerts (js-yaml, h2) (#2548)"), which bumped both from 4.3.0.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #2354, opened 2026-08-08
fix(gate): doc-warnings documents private items, so broken pub(crate) doc links fail the gate (#2336)

### LINKED ISSUE #2336: gate: doc-warnings skips private items, so broken intra-doc links in pub(crate) code are invisible
## Problem

`make doc-warnings` runs `RUSTDOCFLAGS="-D warnings" cargo doc $(CARGO_SCOPE) --no-deps`
(`Makefile`, ~line 220) — **without** `--document-private-items`. Rustdoc does not check
intra-doc links inside items it is not documenting, so every broken link in a private
module, private function, or `pub(crate)` item is invisible to the gate. The gate is green
and the links are broken.

Found while working on PR #2298 (`stella-context`). Adding `--document-private-items` to
the same command surfaces five errors in that one crate, all pre-existing on `main`:

```
error: unresolved link to `domains_by_node`
  --> crates/stella-context/src/candidates.rs:312:7
error: unresolved link to `domains_by_node`
  --> crates/stella-context/src/candidates.rs:317:7
error: unresolved link to `tests`
  --> crates/stella-context/src/retrieval.rs:19:7
error: unresolved link to `DEFAULT_RECENCY_WEIGHT`
  --> crates/stella-context/src/retrieval/ranking.rs:100:7
error: unresolved link to `DEFAULT_MMR_CANDIDATE_MULTIPLE`
  --> crates/stella-context/src/retrieval/ranking.rs:184:22
```

That is one crate of twenty-one, so the workspace-wide count is unknown and probably
much larger. This matters more here than in most repositories: `AGENTS.md` makes doc
comments load-bearing (invariants are cited by number from Rust doc comments, and type
doc comments are treated as the wire contract), so a link that silently stops resolving
is the documented failure mode of exactly the citation scheme the project relies on.

## Reproduce

```bash
RUSTDOCFLAGS="-D warnings" cargo doc -p stella-context --no-deps --document-private-items
# five errors; the same command without --document-private-items exits 0
```

## What "done" looks like

- `make doc-warnings` checks private items too, so a broken link in a `pub(crate)` item
  fails the gate like any other.
- Every link the flag surfaces is repaired (or the reference reworded to not be a link
  where no item exists — `[`tests`]` in `crates/stella-context/src/retrieval.rs:19` looks
  like prose that only wants backticks).
- Land it as **two** changes if the workspace-wide count is large: repairs first, then
  the flag flip, so the gate never goes red on `main` between them.

## Constraints

- Flipping the flag without the repairs turns `main` red for everyone — see AGENTS.md's
  note on shared-cell gate changes and the file-size ratchet precedent (#2004).
- `cargo doc --document-private-items` is slower; check the impact on the pre-push hook's
  `make gate` and on `ci.yml` before flipping.
- `CARGO_SCOPE` must keep working — the hook narrows this step per-diff.

## References

- PR #2298 (where this surfaced), `Makefile` `doc-warnings` target.
- #634 (the issue that introduced the rustdoc gate).

macanderson:
## What & why

Rustdoc does not check intra-doc links inside items it is not documenting, so the
`doc-warnings` gate step — `cargo doc --no-deps` under `-D warnings` — was blind to every
broken link in a private module, private function, or `pub(crate)` item. In a repository
where doc comments are load-bearing (invariants are cited by number from Rust doc
comments; type doc comments are the wire contract), that is the documented failure mode of
the citation scheme itself.

This PR turns the light on and fixes everything it reveals, in one atomic change so `main`
never has a red window between repairs and enforcement:

- **Adds `--document-private-items`** to the gate command in its three normative homes:
  the `Makefile` `doc-warnings` target (still `CARGO_SCOPE`-aware), `ci.yml`'s raw
  invocation, and `CONTRIBUTING.md`'s gate block (`check-gate-parity.sh` stays green).
  This is the same posture rust-lang/cargo's CI takes for its internal docs.
- **Adds `--keep-going`** to the same command: rustdoc bails at the first failing crate,
  which has repeatedly masked the next dependency layer's break (#1823 was diagnosed one
  layer at a time). With it, one gate run reports every crate.
- **Repairs all 34 references the flag surfaces**, across nine crates
  (`stella-diag`, `stella-context`, `stella-store`, `stella-core`, `stella-graph`,
  `stella-observatory`, `stella-tools`, `stella-serve`, `stella-tui`). The issue's
  estimate was right: the five known `stella-context` errors were one crate of a much
  larger population, including links whose targets had been **renamed or deleted** —
  `render_transcript` → `render_transcript_window`, `RepoBackend::push` → `push_branch`,
  `fetch_url` → `fetch_raw`, `RESUME_GRACE` → the configured `resume_grace` /
  `DEFAULT_RESUME_GRACE`, and `BlockDraft::without_local_content`, a mechanism that no
  longer exists (the doc now describes the real one: decomposition maps
  `AttachmentSource::Data` to no local preimage).

Closes #2336

### The repair rules (for review)

Rustdoc resolves intra-doc links with ordinary Rust name-resolution *visibility* from the
linking scope — `--document-private-items` does not change that. So:

- A target that is `pub`/`pub(crate)` (or reachable through a `pub(crate) use` re-export,
  e.g. `crate::store::domains_by_node`) gets a real link target; display text is unchanged.
- A fully-private item in a *sibling* module cannot be linked from anywhere outside its
  subtree (this is why `stella-graph`'s already-qualified `crate::store::index_one` and
  `crate::walk::DENY_DIRS` failed). Those become prose backticks, usually with a link to
  the visible containing module — the same shape #1965/#1970 used.
- `[`tests::…`]` references point into `#[cfg(test)]` modules that never exist in a doc
  build → prose backticks.
- `argv[0]`, `agent[0]`, `"sha256:<hex>"` → backticked so rustdoc stops parsing them as
  links/HTML.
- The `accept.rs` twins (`stella-observatory` / `stella-serve`) received byte-identical
  edits, keeping `the_two_copies_of_this_policy_have_not_drifted` green.
- God files (`driver.rs`, `deck_ui.rs`, `registry.rs`, `stella-store/lib.rs`) were edited
  strictly in place — zero added lines.

## The witness

- [x] This PR includes a witness (fails on `main`, passes here) — **the gate command
      itself is the witness**: `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps
      --document-private-items --keep-going` exits **101 on `main`** (34 errors, nine
      crates failing) and **0 on this branch** (verified locally, full workspace).

## The gate

- [x] `cargo fmt --check` (clean locally)
- [ ] `cargo clippy --workspace --all-targets -- -D warnings` — left to CI (doc-comment-only
      Rust changes; this machine deliberately does not run workspace-wide builds)
- [ ] `cargo test --workspace` — left to CI, same reason
- [x] Docs updated where behavior/flags changed (Makefile help text, ci.yml comment,
      CONTRIBUTING.md gate block)
- [x] CLA signed
- [x] `Closes #2336` appears both above and as a commit trailer

Also run locally: the full flagged doc build (exit 0) and `check-gate-parity.sh` (OK).

**Cost of the flag** (the issue asked): on an identical warm cache, forcing a full re-doc
of all 21 workspace crates — old command 28.1s wall, new command 28.4s wall (~1%; the
extra rustdoc work hides inside build parallelism). Pre-push hook and CI impact is noise.

## Nothing left behind

- [x] There is nothing new to file: the two Dependabot alerts the push banner mentions are
      already tracked as #2121 (with #2263 covering the phantom-alert cause), and
      everything this change surfaced is fixed here.

## Ground-rule check

- [ ] No I/O added to `stella-core` (doc comments only); no new deps

- `make docs` (the human browse target) deliberately keeps building **without**
  `--document-private-items`: whether browsed docs should include every private item is a
  reading-experience preference, and "now vs right" on that is the maintainer's call. The
  gate and CI are what must see private items, and now do.
- One rustdoc quirk worth knowing: `deck_ui/create.rs`'s module doc resolves some of its
  link fragments at the *parent* module's scope (rustdoc merges outer `///`-on-`mod` docs
  with inner `//!` docs and loses the span). The three affected links now use
  crate-absolute targets, which resolve identically from either scope.
- `demo-scenario.sh` and the ultra-audit skill config still use the plain command — both
  are stress/audit harnesses, not the gate, and were left alone on purpose.

## Summary by Sourcery

Tighten the rustdoc gate to cover private items and keep running on multiple failures, and repair all intra-doc links and documentation references surfaced by the new checks across the workspace.

Enhancements:
- Improve internal diagnostics and behavior descriptions in doc comments, including references to configuration fields and behavior summaries, without changing runtime code paths.

Build:
- Update the Makefile `doc-warnings` target to run rustdoc with `--document-private-items` and `--keep-going` so the gate enforces clean docs for public and private items.
- Adjust the CI workflow to use the stricter rustdoc invocation, ensuring private and `pub(crate)` docs are checked and multiple failing crates are reported in a single run.

Documentation:
- Update CONTRIBUTING guidelines to reflect the strengthened doc gate command.
- Clarify and correct numerous intra-doc links, references, and prose in rustdoc comments across multiple crates so they point at the current APIs and avoid links to private or test-only items.


macanderson:
The failing check here is not this diff: main is red on shellcheck (issue #2355, introduced by the arena-lifecycle scripts from #2328/#2351 — this PR touches no shell script). The unbreak is PR #2363; once it merges, a branch update / re-run should turn this green.

Issue #2358, opened 2026-08-08
cli: reflection dedup runs before the mining log, and shares its threshold with the clustering predicate — skills and rules can only be mined inside an accidental band


macanderson:
## Problem

Skill and rule auto-creation from reflection is reachable only inside a **narrow, accidental band**, and outside it the loop can mine nothing at all. Found by replaying the real loop under the trace-replay harness (#2304); a corpus of naturally-varied lessons about one convention mined **zero** candidates on the first probe.

Two facts compose into it.

**1. Dedup runs before the mining log is written.** In `SessionMemory::reflect_and_record` (`crates/stella-cli/src/memory/learning.rs`), the order is:

```rust
let lessons = self.retain_unforgotten(turn_store.as_ref(), lessons);
let lessons = self.retain_unknown(lessons);        // <-- drops restatements
if lessons.is_empty() { return ...; }
// ... store upsert ...
// ... append to reflections.jsonl ...   <-- the mining log
```

A lesson dropped by `retain_unknown` never reaches `reflections.jsonl`, so it never becomes an `ObservationRecord` (`crates/stella-cli/src/memory/observations.rs`), so it can **never contribute an occurrence** to either miner.

**2. The dedup predicate and the clustering predicate are the same metric at the same threshold.**

| | drops a lesson | clusters two lessons |
|---|---|---|
| predicate | `stella_store::is_suppressed` | `stella_core::skills::mine_skill_candidates` |
| metric | Jaccard | Jaccard |
| threshold | `SIMILARITY_THRESHOLD = 0.5` | `SkillMineConfig::min_similarity = 0.5` |

`SkillMineConfig::min_occurrences` is 3. So three lessons must be **similar enough to cluster** (Jaccard ≥ 0.5) while each stays **dissimilar enough to survive dedup** (Jaccard < 0.5) against everything already stored. With one metric and one threshold that is a contradiction.

**The only reason it is not a total block is that the two tokenizers differ**, and neither module knows about the other:

| | `stella_store::forget::tokens` | `stella_core::mining::terms` |
|---|---|---|
| `crates/stella-tools/src/registry.rs` | **one** token (keeps `.` `-` `/` `_`) | **five** terms (splits on non-`[a-z0-9_]`) |
| stopwords, words ≤ 2 chars | kept | dropped |

[...]

Issue #2355, opened 2026-08-08
main is red: shellcheck fails on scripts/test-arena-scripts.sh, reddening every open PR


macanderson:
## Problem

`make gate` is red on `main`. The `shellcheck` step fails on
`scripts/test-arena-scripts.sh`, which arrived with #2328 and was extended by
#2351. Nine findings across two codes, `SC2016` and `SC2028`.

Because `shellcheck` is a `GATE_STEPS` entry and a required CI check, this reds
**every** open PR and every push, regardless of what the PR changed — the
failure has nothing to do with the diff under review. Found from an unrelated
branch (#2333) whose diff touches no shell script at all.

## Reproduce

```bash
git checkout main && make shellcheck
# make: *** [shellcheck] Error 1
```

Sample:

```
In scripts/test-arena-scripts.sh line 223:
  echo 'case "$1" in'
       ^------------^ SC2016 (info): Expressions don't expand in single quotes, use double quotes for that.

In scripts/test-arena-scripts.sh line 227:
  echo '    printf "%s\n" "$STUB_LOG_BODY" >> "$STUB_LOG"'
       ^-- SC2016 (info): Expressions don't expand in single quotes, use double quotes for that.
       ^-- SC2028 (info): echo may not expand escape sequences. Use printf.
```

## Why it fires

The script generates a stub shell script by `echo`ing its lines. Single quotes
are **correct** here — the `$1`, `$STUB_LOG` and `\n` are meant to reach the
generated file literally, not to expand in the generator. So these are false
positives about intent, not real bugs: the code is right and the lint does not
know what the string is for.

`make shellcheck` runs at shellcheck's default severity (`style`), so
`info`-level findings fail the gate.

## What "done" looks like

The gate is green on `main` and the generator still emits the same bytes.

Preferred fix, in rough order:

1. Emit the stub with a **quoted heredoc** (`cat <<'EOF' > "$stub"`) instead of
   a run of `echo` calls. Quoted-heredoc content is literal by construction, so
   both codes stop applying and the intent becomes self-evident to a reader as
   well as to the linter. This also fixes `SC2028` properly rather than
   silencing it.
2. Failing that, a narrowly scoped `# shellcheck disable=SC2016,SC2028` with a
   comment stating *why* the lint is wrong here — the literal text is the
   payload. Per CLAUDE.md, a bare disable with no such comment is a defect in
   its own right.

Do **not** widen the gate (drop the step, raise the severity floor with `-S
[...]

Issue #1139, opened 2026-08-01
Home/data-dir resolution should be an injected port, not ~10 ambient env reads (#911 option 1)


macanderson:
## What

Home-directory and data-directory resolution is read ad-hoc from `std::env` at ~10 independent call sites in `stella-cli`, rather than being supplied as a port. This is the option #911 itself named as most preferred and which its fix (#1098) explicitly did **not** take.

From #911's body:

> **1. Inject instead of mutate.** The workspace already enforces "ports, not concretions" (`AGENTS.md`), and `stella-core` is provably I/O-free. Home-directory and data-directory resolution should be a port the test supplies, not an ambient read of `std::env`. This deletes the problem rather than containing it, and it is the fix most consistent with the existing architecture.

#1098 took the containment path (panic-safe restore) because eliminating the reads means changing a shared dependency crate — disproportionate risk for a P2 test-flakiness fix. That was the right call for that PR, and it leaves this undone.

## Current state

**`HOME` is read directly in 10 files in `stella-cli/src`:** `settings.rs`, `settings/toml_config.rs`, `extensions.rs` (×2), `env_files.rs`, `attachments.rs`, `command_deck.rs`, `agent.rs`, `agent/tools.rs`, `skill_manager.rs`.

**A partial funnel already exists but is barely adopted.** `settings.rs` has `user_home_dir()` backed by a `#[cfg(test)]` thread-local (`TEST_USER_HOME`) with an RAII override guard — a per-thread, no-`unsafe`, no-process-mutation seam that is exactly the right shape. Only ~4 call sites route through it. The other readers bypass it entirely, which is *why* tests exercising those paths must mutate the real process `HOME`.

**`STELLA_DATA_DIR` resolution lives in a different crate.** `stella-store/src/usage.rs::data_dir()` is canonical, and `stella-observatory/src/global.rs` deliberately mirrors it to avoid linking `stella-store`. `stella-cli` never reads it directly — it arrives indirectly through those crates, which is why `stella-cli` tests must set a process-global variable to influence a dependency's behaviour.

A sibling thread-local for a related concern (`TEST_FILESYSTEM_ISOLATION` gating `filesystem_settings_disabled()`) has the same shape and the same partial adoption — worth unifying in the same pass.

## Why it is worth doing

[...]

PR #2367, opened 2026-08-08
feat(stella-cli): the Claude Code transcript adapter — local-only, option (b), honest about what the corpus cannot supply (#2304)

macanderson:
Refs #2304. Implements **PR 5** of `doc:trace-replay-learning-harness` §9 — the Claude Code transcript adapter — and completes spec §13 with the adapter's half (§13.3), which #2350 left open.

**Based on `main`**, which now carries the harness — #2350 merged while this was in flight.

*(Replaces #2357, which GitHub auto-closed when #2350 merged and its base branch `feat/2304-trace-replay-harness` was auto-deleted. Same content, rebased onto the merged harness.)*

## The adapter's honest limit is the design

Claude Code transcripts contain **no Stella reflection JSON**. There is nothing in them to script a lessons array from, so §7.2 puts the choice plainly: derive lessons from the transcript, or decline to. Deriving them means the harness measures the adapter's lesson-invention heuristic instead of Stella's learning — it fabricates the exact signal under test.

**This implements option (b), which the spec recommends**: shell history, session and turn boundaries, and timing only. It lights up the tool foundry against thousands of real commands and launders nothing.

### That needed a fourth `ScriptedReflection` arm

Expressing "the source carried no reflection" with the three existing arms is impossible without asserting something the source does not support:

| Spelling | The claim it makes |
|---|---|
| `Lessons { lessons: [] }` | the model had nothing to say |
| `Unreadable { .. }` | the model said something unparseable |
| `ModelError { .. }` | the call failed |

The second is the dangerous one: it would **fabricate starvation**, and assertion 7 (`an_unreadable_corpus_builds_nothing_and_says_why`) counts exactly those turns. An adapter that emitted `Unreadable` for 16,000 turns would corrupt the metric its sibling test depends on.

So `NotRecorded` is its own arm. The replayer skips the model boundary entirely for it, and the summary counts `turns_not_recorded` separately from empty reflections — a metric that folded them together would report the learner as idle when it was never asked.

`Provenance::Derived` already exists on the `Lessons` arm from #2350, and `every_lesson_is_labelled` asserts nothing reaches a trace unlabelled — so if option (a) is ever built, it cannot ship un-stamped quietly.

## The privacy gate (§7.1), applied where the risk actually is

The spec asks for "a secret-shaped string is quarantined rather than stored". Under option (b) the adapter derives **no statements**, so there is no proposal to run through `gate_proposal`/`quarantine_for`. The highest-risk field it does touch is the **shell command** — `export ANTHROPIC_API_KEY=sk-ant-…` and `curl -H "Authorization: Bearer …"` are ordinary things to have typed.

So the gate runs there, and it is **stricter than quarantine**: `redact_secrets` runs on every command, and a command whose redaction *fired* is **dropped**, not kept with a `[redacted]` hole in it. Two reasons:

- The redactor's prefix list is a good filter, not a complete one. A token shape it has not seen would survive in a command it had partly redacted, and the partial redaction is what would make that look safe.
- The foundry loses nothing. A command carrying a credential is a one-off by nature, so it was never going to be a recurring shape worth minting a tool from.

Beyond that: **local-only and opt-in** (`STELLA_REPLAY_CC_CORPUS`, reachable only from a test — no shipped command, and a plain `cargo test` never touches the corpus); **nothing derived is committed** (the CI corpus stays synthetic, permanently, and `no-scratch` fails the gate on a committed derivative on its own); and the transcript stub is left **empty**, because nothing reads it under option (b) and carrying user text into a trace we did not need would be gratuitous retention.

## Measured against the real corpus

```
corpus: 496 project director(ies)
adapted 20 project(s): 16664 turn(s), 17343 command(s)
```

Every adapted trace was round-tripped through `Trace::parse` — the loader's own contract, applied to a real uncontrolled source rather than to a synthetic fixture the adapter was written against. That is the assertion that catches a transcript shape the adapter mishandles.

**A correction to §7's measurement, recorded in §13.3.** The spec measured 485 project directories on 2026-08-08; it is 496 today. Consistent with the rolling window the spec already flags, and the reason a derived trace can never be a committed fixture.

## The spec now records what was built (§13)

The document said `status: proposed`. Rather than leave a plan describing something that now exists, §13 records the divergences and — more usefully — **the four things replaying the real loop measured**, none of which is visible from reading a single module: the shared dedup/clustering threshold, the foundry's value-like-argument rule, the workspace-derived lineage id, and lexical `starts_with`. A future reader planning fixtures needs all four.

## Verification

- `cargo test -p stella-cli` — **1531 passed, 0 failed** (12 new), on top of the merged harness
- `STELLA_REPLAY_CC_CORPUS=1 cargo test … the_real_corpus_adapts` — output above
- `cargo clippy -p stella-cli --all-targets -- -D warnings` — clean
- `make guards-fast` — clean (#2355's shellcheck failure is fixed on `main` by #2363)

**No tests deleted.**

### One guard caught a real mistake, on the first run

`paths::tests::nothing_else_in_this_crate_reads_a_home_out_of_the_environment` failed on my `std::env::var_os("HOME")`. Fixed to `crate::paths::home()`, which is what lets a test redirect the anchor without mutating process-global state (#1139). Worth naming because it is exactly the guard working as designed.

## Not in this PR

- **Option (a)** — synthesizing lessons. The labelling machinery is in place if it is ever wanted; the spec's recommendation against it stands.
- **Committing any derived trace.** Permanently out of scope.

## Summary by Sourcery

Add a local-only, opt-in Claude Code transcript adapter that feeds real shell history into the trace replay harness without inventing reflections, and extend the harness to account for turns where the source provided no reflection.

New Features:
- Introduce a Claude Code transcript adapter that converts local transcripts into trace sessions for the replay harness under an explicit environment-based opt-in.
- Add support for a `ScriptedReflection::NotRecorded` arm and corresponding replay summary metric to represent turns where the source carried no reflection JSON.

Enhancements:
- Update the replayer to skip the reflection model boundary entirely for turns marked as not recorded while still driving the tool foundry and timeline.
- Refine the trace replay harness documentation to describe the built Claude Code adapter, its privacy guarantees, and its measured corpus characteristics.

Documentation:
- Document the Claude Code adapter design, its option (b) behavior, privacy gate, and real corpus measurements in the trace replay harness spec.

Tests:
- Add adapter and privacy-gate tests over synthetic transcripts, including opt-in behavior and timestamp handling, and a guarded test that exercises the real local corpus when enabled.

### DIFF
--- a/crates/stella-cli/src/memory/replay.rs
+++ b/crates/stella-cli/src/memory/replay.rs
@@ -48,6 +48,10 @@
 //!   between two replays inside one process. Compare edge content, never edge
 //!   identity (see `stella_context`'s `process_nonce` audit note).
 
+// §7: the Claude Code transcript adapter. Local-only and opt-in; it derives no
+// lessons (option (b)), so it feeds the foundry and the timeline and never the
+// reflection path.
+pub(crate) mod cc_adapter;
 pub(crate) mod summary;
 pub(crate) mod trace;
 
@@ -303,21 +307,30 @@ impl<'a> Replayer<'a> {
         //     lesson, then the next turn must not re-learn it.
         self.apply_world_changes(turn);
 
-        // 2-3. Script the single model boundary.
-        let provider = ScriptedProvider::for_turn(&turn.reflection);
-        let transcript: Vec<CompletionMessage> = turn
-            .transcript
-            .iter()
-            .map(CompletionMessage::from)
-            .collect();
-
-        // 4. The one call. Everything downstream of it — the forgetting filter,
-        //    the store write, the mining log, observation extraction, both
-        //    miners, the retirement sweep — is already deterministic and runs
-        //    inside it.
-        let report = memory
-            .reflect_and_record(&provider, "replay", &transcript, true, turn.succeeded, None)
-            .await;
+        // 2-4. The single model boundary, scripted — unless the source carried
+        //      no reflection at all, in which case the turn does not reach it.
+        //      Skipping is the whole of option (b) in §7.2: an adapter over a
+        //      corpus with no reflection JSON supplies shell history, session
+        //      boundaries and timing, and scripting a reflection it never saw
+        //      would launder an invention into a result.
+        //
+        //      Everything downstream of the call — the forgetting filter, the
+        //      store write, the mining log, observation extraction, both miners,
+        //      the retirement sweep — is already deterministic and runs inside
+        //      `reflect_and_record`.
+        let report = if turn.reflection.reaches_the_model() {
+            let provider = ScriptedProvider::for_turn(&turn.reflection);
+            let transcript: Vec<CompletionMessage> = turn
+                .transcript
+                .iter()
+                .map(CompletionMessage::from)
+                .collect();
+            memory
+                .reflect_and_record(&provider, "replay", &transcript, true, turn.succeeded, None)
+                .await
+        } else {
+            crate::memory::ReflectionReport::default()
+        };
 
         // 5. The foundry, over the history accumulated so far. Pure over the
         //    slice and documented to be byte-identical for identical input, so
--- a/crates/stella-cli/src/memory/replay/cc_adapter.rs
+++ b/crates/stella-cli/src/memory/replay/cc_adapter.rs
@@ -1,0 +1,427 @@
+//! Claude Code transcripts → the trace format. **Local-only, opt-in, and
+//! deliberately half of what the corpus could supply.**
+//!
+//! `doc:trace-replay-learning-harness` §7. This is the one non-synthetic corpus
+//! available: one engineer, a handful of repositories, the same conventions
+//! re-encountered across thousands of sessions — recurrence density no synthetic
+//! fixture will match.
+//!
+//! ## What it does not do, and why that is the point
+//!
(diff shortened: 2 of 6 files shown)


macanderson:
**Heads-up on a spec collision with #2371.** Both PRs edit `docs/spec/trace-replay-learning-harness.md` §13:

- this PR inserts **§13.3 "The adapter, as built"** and renumbers "Still open" to §13.4;
- #2371 (which fixes #2358) rewrites **§13.1 item 1** to describe the fix rather than the workaround.

Different subsections, so a textual conflict is likely but the resolution is a straight union — take both. Whichever lands second should rebase and confirm §13 still reads 13.1 → 13.2 → 13.3 → 13.4 with no duplicate heading. Flagging it rather than racing, because a same-day rewrite of one file by two PRs is how additions get silently dropped in this repo.

No code overlap: #2371 touches `learning.rs`, `skills.rs`, `rules.rs`, `replay/tests.rs` and the committed corpus; this touches `replay.rs`, `replay/trace.rs`, `replay/summary.rs` and the new `replay/cc_adapter*`.

Also worth saying plainly: **#2358 came out of this harness**, and #2371 shipping a measured fix within hours is the loop working — the corpus in #2371 can now be worded naturally, which is exactly the workaround §13.1 said should not outlive the bug.

Issue #2369, opened 2026-08-08
core: lexical clustering still misses the far tail of natural variation — same-fact pairs at 0.28 sit below the cross-fact noise ceiling at 0.36


macanderson:
## Problem

The #2358 fix moved the miners' clustering threshold to 0.4 — measured in `stella_core::mining::terms` space — and let store-deduplicated restatements reach the mining log. Naturally-varied phrasings of one convention now cluster (same-fact pairs measure 0.40) and the trace-replay corpus mines a skill and a rule from natural wording.

But the *far* tail of natural variation is still invisible to lexical Jaccard, and no threshold can catch it:

```
"Tool registration happens in crates/stella-tools/src/registry.rs — add the entry there."   (cluster head)
"Every builtin is wired through crates/stella-tools/src/registry.rs before it can be called." (same fact)
```

These score **0.278** in `terms` space — below any usable threshold, because the worst *cross-fact* pair already measures **0.36** (two short lessons naming different files in one crate share `crates`/`stella`/`<crate>`/`src` boilerplate). Signal and noise overlap: 0.28–0.40 same-fact vs ≤0.36 cross-fact. Lexical overlap alone cannot separate them; the fourth phrasing stays a singleton cluster in the committed corpus today (deliberately, as the honest tail).

## Where

- `crates/stella-core/src/mining.rs` — `terms`, `jaccard`, `cluster_observations` (shared by both miners)
- `crates/stella-core/src/skills.rs` / `rules.rs` — `min_similarity` doc comments carry the measurements
- `crates/stella-cli/src/memory/replay/tests.rs` — `recurring_registry_lessons()` holds the exact strings; `the_dedup_and_clustering_predicates_hold_the_declared_relationship` is the tripwire that pins the current band
- `crates/stella-cli/src/memory/anchors.rs` — `resolve_anchors` already extracts the file paths a lesson names (used for memory anchoring, not for mining)

## Directions (none settled)

1. **Anchor-aware clustering.** Both example lessons name the same concrete file. `resolve_anchors` already computes this at store time; observations could carry resolved anchors, and a shared anchor could substitute for (or boost) lexical overlap. Strongest signal, and immune to path-boilerplate noise — but `ObservationRecord` doesn't carry anchors today, so it is a schema addition (`LIFECYCLE_SCHEMA_VERSION` bump).
[...]

PR #2371, opened 2026-08-08
fix(stella-cli): mine the recurrence the store deduplicates, and measure the clustering threshold in its own token space (#2358)

### LINKED ISSUE #2358: cli: reflection dedup runs before the mining log, and shares its threshold with the clustering predicate — skills and rules can only be mined inside an accidental band

macanderson:
## What & why

Skill and rule auto-creation from reflection was reachable only inside a narrow, accidental band, found by replaying the real loop under the trace-replay harness (#2304). Two defects composed into it, and this PR fixes both halves:

**Ordering.** `retain_unknown` dropped a store-restatement *before* the mining log was appended, so re-learning — the most common recurrence shape the loop sees (measured: 61% of a live store was restatement) — never became an observation and could never contribute an occurrence to either miner. It is now `partition_known`, which **diverts** a restatement to `reflections.jsonl` instead of dropping it: the store still keeps one copy per fact (recall has a budget, and that dedup is right), while the log records every recurrence. Within-batch stutter still lands nowhere, so one verbose reflection cannot impersonate three turns of evidence.

**Threshold.** The miners clustered at Jaccard ≥ 0.5 — a number inherited from `stella_store::forget`'s dedup constant, which is measured over a tokenizer that keeps a file path as ONE token, where `mining::terms` shatters the same path into five. Measured in the miners' own space, naturally-varied same-fact pairs score **0.40** and the worst adversarial cross-fact pair **0.36** — so at 0.5, natural phrasing could never cluster and only wording engineered into the 0.5–0.67 band ever mined anything. `min_similarity` moves to **0.4** in both miners, with the measurements recorded on the field's doc comment.

The relationship between the two thresholds is now **declared in code and tripwired**: `the_dedup_and_clustering_predicates_hold_the_declared_relationship` asserts both halves of the band on the committed corpus's own strings (the four natural wordings each survive dedup; the fifth trips it; the miner clusters them anyway, *including* the deduped restatement). The committed corpus and `recurring_registry_lessons()` are reworded to natural phrasing plus one near-restatement, the replay tests' long "how to word a fixture around this bug" explanation shrinks to a sentence, and spec `doc:trace-replay-learning-harness` §13.1 item 1 now describes the fix.

Exemplar note: the split keeps decision logic pure and in `stella-core`/`stella-store` (Invariant 2) — only the sequencing lives in `stella-cli`, matching the existing `retain_unforgotten`/`is_suppressed` structure rather than inventing a new mechanism.

Closes #2358

## The witness

Both halves verified independently by reverting each in place and re-running:

- **Threshold flip** (0.4 → 0.5): `the_dedup_and_clustering_predicates_hold_the_declared_relationship`, `recurrence_across_distinct_tasks_builds_what_recurrence_within_one_task_does_not`, `the_committed_corpus_builds_a_memory_a_skill_a_rule_and_a_tool`, `an_auto_activated_rule_is_advisory_and_never_blocking`, and `auto_activation_never_clobbers_an_existing_record` all FAIL.
- **Ordering flip** (restatements dropped before the log): four end-to-end replay tests FAIL — the fourth occurrence never lands, so the rule's confidence stays at 70 < 85 and nothing auto-activates.

The negative controls still hold: five turns of ONE task mint nothing (anti-poisoning), unrelated one-off lessons stay singletons, forgotten lessons stay forgotten (`retain_unforgotten` still runs before the log — forgetting remains durable).

## The gate

- [x] `cargo fmt --check`
- [x] clippy clean on the three touched crates (`stella-core`, `stella-store` — untouched in the end, `stella-cli`), `--all-targets`/`--tests`; CI runs the workspace
- [x] `cargo test -p stella-core --lib` (1039), `-p stella-store --lib` (317), `-p stella-cli --bin stella` (1521) all green; CI runs the workspace suite
- [x] Docs updated where behavior changed (doc comments on both `min_similarity` fields, `partition_known`, spec §13.1/§13.3)
- [x] CLA signed
- [x] `Closes #2358` appears both above and as a commit trailer

Renamed test, named per the deleted-test guard's contract:
`memory::learning::dedupe::a_paraphrase_of_a_stored_memory_is_dropped_and_a_new_fact_survives` → `a_paraphrase_of_a_stored_memory_is_diverted_and_a_new_fact_is_stored` (same file, assertion sharpened for the new split; nothing else deleted).

## Nothing left behind

- [x] Filed: #2369 — the far tail of natural variation (same-fact pairs at 0.28) still sits below the cross-fact noise ceiling (0.36) in lexical-Jaccard space; no threshold can catch it, so it needs anchor-aware clustering / IDF-weighting / stemming, written up as a handoff with the measurements.

## Ground-rule check

- The observatory's lessons panel and `stella proposals` read `reflections.jsonl` directly; they now see restatements too. That is faithful — the log is the mining journal, and restatements are real events in it. The store-backed `reflections` table mirror stays in lockstep with the jsonl by design.
- `0.4` catches the corpus's natural pairs at exactly 0.40 (6/15, deterministic in f64). The tripwire test asserts through the real miner on the real strings, so a fixture rewording that drops below the threshold fails loudly with the measurements in the failure message trail, not silently.
- Alternative rejected: raising the *dedup* threshold instead (issue direction 2) — it neither lets natural phrasings cluster (their blocker is the miner's space, where they score 0.40) nor preserves the recall-budget measurement behind 0.5 in `forget::tokens` space.

## Summary by Sourcery

Ensure naturally phrased recurring lessons contribute to skill and rule mining while keeping the context store deduplicated.

Bug Fixes:
- Fix loss of recurrence signal by diverting store restatements to the mining log instead of dropping them before mining.
- Fix clustering blind spot by lowering miners’ similarity threshold to a value measured in their own token space and aligning skills and rules thresholds.

Enhancements:
- Refine session memory handling to split novel lessons from restatements, logging both while only storing novel lessons.
- Add a tripwire test that asserts the relationship between store dedup and miner clustering predicates against the committed corpus.
- Tighten replay and dedupe tests to cover restatement diversion, within-batch stutter handling, and recurrence across tasks with natural wording.
- Clarify documentation and spec around dedup vs clustering behavior and the measurements behind the mining thresholds.

Documentation:
- Update trace-replay harness spec and mining configuration docs to describe the new dedup/clustering behavior and measured thresholds.

Tests:
- Expand replay, dedupe, and mining configuration tests to cover the new partitioning of lessons, threshold changes, and the declared dedup/clustering relationship.

### DIFF
--- a/crates/stella-cli/src/memory/learning.rs
+++ b/crates/stella-cli/src/memory/learning.rs
@@ -28,9 +28,10 @@ use super::{
 #[cfg(test)]
 mod guarantees;
 
-/// What [`SessionMemory::retain_unknown`] drops and keeps. A child module for
-/// the same reason as [`guarantees`]: the filter is private, and it is the
-/// filter itself that needs pinning, not the turn that happens to call it.
+/// What [`SessionMemory::partition_known`] stores, diverts to the mining log,
+/// and drops. A child module for the same reason as [`guarantees`]: the split
+/// is private, and it is the split itself that needs pinning, not the turn
+/// that happens to call it.
 #[cfg(test)]
 mod dedupe;
 
@@ -159,20 +160,17 @@ impl SessionMemory {
         // at recall is what makes forgetting durable — an unsuppressed lesson
         // would land in the log and stay re-mineable forever.
         let lessons = self.retain_unforgotten(turn_store.as_ref(), lessons);
-        // Then drop what we already know. The loop re-learns the same facts in
-        // slightly different words every turn, and only byte-identical content
-        // collapses on its own (a memory's lineage is seeded from its content
-        // hash), so paraphrases accumulate unchecked.
-        //
-        // Measured on a live treatment store: 23 stored memories encoding
-        // **six** distinct facts, with "commands are registered in registry.py"
-        // held seven separate times — 61% of the store was restatement. That is
-        // not merely untidy. Recall has a budget, so three slots spent on three
-        // phrasings of one fact are three slots not spent on the other five,
-        // and the store gets worse at covering the codebase the longer it runs.
-        let lessons = self.retain_unknown(lessons);
+        // Then split what survives into lessons the store should learn and
+        // restatements of what it already holds. The split — rather than the
+        // filter this used to be — is the #2358 fix: a restatement must skip
+        // the store (see `partition_known` for the measured cost of storing
+        // paraphrases) but still reach the mining log below, because a lesson
+        // the loop keeps re-learning is exactly the recurrence the skill and
+        // rule miners count. Dropping it here starved both miners of the
+        // most common recurrence shape there is.
+        let (novel, restatements) = self.partition_known(lessons);
 
-        if lessons.is_empty() {
+        if novel.is_empty() && restatements.is_empty() {
             return ReflectionReport {
                 cost_usd: reflection_cost_usd,
                 events: reflection_events,
@@ -196,36 +194,50 @@ impl SessionMemory {
         // guessed (see `anchors::resolve_anchors`). A lesson that names no file
         // simply gets no anchors, which is the common case for process notes
         // and is exactly right — they are not about a file.
-        let delta = ContextDelta {
-            memories: lessons
-                .iter()
-                .map(|l| {
-                    MemoryInput::reflection(&l.lesson, l.domains.iter().cloned())
-                        .with_recall_tier(l.kind.recall_tier())
-                        .with_anchors(crate::memory::anchors::resolve_anchors(
-                            &self.workspace_root,
-                            &l.lesson,
-                        ))
-                })
-                .collect(),
-            ..Default::default()
+        let stored = if novel.is_empty() {
+            // Nothing needed storing, so nothing failed to store: a
+            // restatement-only turn feeds the miners below and claims no
+            // memory.
+            true
+        } else {
+            let delta = ContextDelta {
+                memories: novel
+                    .iter()
+                    .map(|l| {
+                        MemoryInput::reflection(&l.lesson, l.domains.iter().cloned())
+                            .with_recall_tier(l.kind.recall_tier())
+                            .with_anchors(crate::memory::anchors::resolve_anchors(
+                                &self.workspace_root,
+                                &l.lesson,
+                            ))
+                    })
+                    .collect(),
+                ..Default::default()
+            };
+            self.store.upsert(delta).await.is_ok()
         };
-        let stored = self.store.upsert(delta).await.is_ok();
 
         // 2. Append to the mining log and mine for auto-creatable skills.
-        // Count how many lessons actually reached the log so the message below
-        // reports partial persistence accurately (some serialize/append writes
-        // may fail while others succeed). Each lesson is also mirrored into
-        // `store.db`'s durable `reflections` table — the surface the
-        // observatory's reflections panel, the JSON export, and the prune
-        // carve-out read; without the mirror those readers only ever saw an
-        // empty table. Best-effort like every store write here: a failed
-        // mirror never touches the jsonl path or the turn.
+        // BOTH halves of the split reach the log — the log is the mining
+        // substrate, and the restatements *are* the recurrence it exists to
+        // record (#2358). Count how many novel lessons actually reached the
+        // log so the message below reports partial persistence accurately
+        // (some serialize/append writes may fail while others succeed). Each
+        // lesson is also mirrored into `store.db`'s durable `reflections`
+        // table — the surface the observatory's reflections panel, the JSON
+        // export, and the prune carve-out read; without the mirror those
+        // readers only ever saw an empty table. Best-effort like every store
+        // write here: a failed mirror never touches the jsonl path or the
+        // turn.
         let log_path =
             stella_store::workspace_private_state_path(&self.workspace_root, "reflections.jsonl")
                 .ok();
-        let mut logged_count = 0usize;
-        for lesson in &lessons {
+        let mut novel_logged = 0usize;
+        for (is_novel, lesson) in novel
+            .iter()
+            .map(|l| (true, l))
+            .chain(restatements.iter().map(|l| (false, l)))
+        {
             if let Ok(line) = serde_json::to_string(lesson)
                 && stella_store::append_workspace_private_line(
                     &self.workspace_root,
@@ -233,8 +245,9 @@ impl SessionMemory {
                     &line,
                 )
                 .is_ok()
+                && is_novel
             {
-                logged_count += 1;
+                novel_logged += 1;
             }
             if let Some(store) = &turn_store {
                 let _ = store.record_reflection(&stella_store::ReflectionRow {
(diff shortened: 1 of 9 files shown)


macanderson:
Cross-linking: **#2367** (the Claude Code transcript adapter, the last piece of epic #2304) also edits `docs/spec/trace-replay-learning-harness.md` §13 — it inserts a new **§13.3 "The adapter, as built"** and renumbers "Still open" to §13.4, while this PR rewrites **§13.1 item 1**.

Different subsections, so the resolution is a union — take both — but whichever lands second should rebase and check §13 still reads 13.1 → 13.2 → 13.3 → 13.4 with no duplicate heading. Flagging rather than racing.

No code overlap between them. And thank you for turning #2358 around this fast — that finding came out of the replay harness in #2350, and `replay/tests.rs`'s long explanation of how to word a fixture around the band was explicitly marked as a workaround that should not outlive the bug. Good to see it go.

Issue #2393, opened 2026-08-09
flaky: daemon::tests::a_run_that_ignores_term_is_killed_after_the_grace_period fails under full-suite load


macanderson:
## Problem

`daemon::tests::a_run_that_ignores_term_is_killed_after_the_grace_period`
(`crates/stella-cli/src/daemon/tests.rs:483`) fails intermittently during
`cargo test --workspace`, and passes reliably when run alone.

Observed on a 2026-08-08 run of `make gate`:

```
---- daemon::tests::a_run_that_ignores_term_is_killed_after_the_grace_period stdout ----
thread '...' panicked at crates/stella-cli/src/daemon/tests.rs:483:5:
the child was killed after 81.95502ms, before its 8s to shut down cleanly
test result: FAILED. 1534 passed; 1 failed
```

## Control already run

Do not re-do this part — it was the cheap control per CLAUDE.md:

```bash
cargo test -p stella-cli --bin stella a_run_that_ignores_term   # 3/3 pass, ~8.03s each
cargo test --workspace                                          # green on re-run, 128/128 binaries ok
```

So it is load-dependent, not a regression: under full-suite parallelism the
child is observed as killed ~82ms in, two orders of magnitude below the 8s
grace period the test asserts.

## Why it matters

The assertion is a **lower** bound on how long the child survived — it is the
test that proves SIGTERM gets a grace period before SIGKILL. A flake here is
the kind that gets re-run until green, and the failure mode it guards
(killing a child immediately instead of letting it shut down cleanly) is real.
A test that cries wolf on a busy machine stops being read.

## Likely cause

The test appears to measure wall-clock between spawn and observed death. Under
`cargo test --workspace` the machine is saturated, so either the child never
gets scheduled to install its SIGTERM handler before the signal arrives, or the
harness's own timing baseline drifts. Worth confirming which before fixing —
the two have different fixes and only one is a test bug.

## Definition of done

- The test passes deterministically under `cargo test --workspace` on a loaded
  machine (verify by running the workspace suite several times, not the test
  alone — running it alone is what already passes).
- The fix does not weaken what is asserted: if the grace period is relaxed into
  a wider tolerance, the test must still fail when the grace period is removed
  from the daemon entirely. Check that the artisanal way — delete the grace
  period in `crates/stella-cli/src/daemon/`, confirm red, restore.
- If the root cause is the child not having installed its handler yet, the fix
  belongs in the test's synchronisation (wait for a readiness signal), not in a
[...]

Issue #2392, opened 2026-08-09
invariant #5: type the remaining 43 public `Result<_, String>` signatures in library crates


macanderson:
## Problem

AGENTS.md invariant #5 says library code returns typed, named errors and never
a bare `String`. Nothing enforced it, and the tree had drifted far enough that
the rule and the code disagreed outright — `crates/stella-runtime/src/error.rs`
even documented the contradiction in a module comment.

The guard `scripts/check-typed-errors.py` (`make typed-errors`) now enforces the
rule for the **public API of library crates**, but it landed as a **down-only
ratchet** because 43 violations predated it. This issue is the remainder: the
baseline is meant to reach empty and be deleted along with its
`read_baseline`/`--update` machinery.

## Current state

`scripts/typed-errors-baseline.txt`:

```
stella-model 10
stella-tools 32
stella-tui 1
```

Get the exact list at any time with:

```bash
python3 ./scripts/check-typed-errors.py --update   # then read the diff, or
git stash && python3 ./scripts/check-typed-errors.py   # prints file:line name
```

The 43, by cluster (each cluster wants **one** error enum, not one per fn):

- `crates/stella-model/src/provider_listing.rs` — `parse_openrouter`,
  `fetch_openrouter`, `fetch_anthropic`, `fetch_gemini`,
  `parse_openai_compatible`, `fetch_openai_compatible` (6). These mix HTTP
  transport failure with response-shape failure, which is exactly the
  distinction a caller needs and cannot currently make.
- `crates/stella-model/src/factory.rs` — `build_provider`, `check_seed_floor` (2)
- `crates/stella-model/src/modelsdev.rs` — `parse_catalog`, `fetch_catalog` (2)
- `crates/stella-tools/src/issue_ops.rs` — 8 fns (`list_issues`, `get_issue`,
  `create_issue`, `set_status`, `add_comment`, `update_issue`, `search_labels`,
  `search_members`)
- `crates/stella-tools/src/tracker_auth.rs` — 9 fns (`get`, `put`, `remove`,
  `connections`, `resolve`, `github_device_login`, `linear_oauth_login`,
  `refresh_connection`, `fetch_account_label`)
- `crates/stella-tools/src/` — `custom.rs::parse_manifest`,
  `diagnostics.rs::resolve_plan`, `durable_write.rs::write_file_durably`,
  `foundry_author.rs::author`, `foundry_gate.rs::{observe,recheck_before_launch}`,
[...]

PR #2394, opened 2026-08-09
feat(gate): make invariant #5 true — type the public errors, then enforce it

macanderson:
## What & why

Invariant #5 says library code returns typed, named errors and never a bare
`String`. Nothing enforced it, and the tree had drifted far enough that the rule
and the code disagreed outright — `crates/stella-runtime/src/error.rs` had been
documenting the breach in a module comment for as long as it existed:

> The CLI originals returned `Result<_, String>` — fine for a binary that prints
> the string and exits, useless to a server that has to decide whether a
> session-create failed because the *caller* sent a bad model slug (a 400) or
> because the *host* is misconfigured (a 500).

Per CLAUDE.md, a disagreement between AGENTS.md and the code is itself the bug.
This PR closes it from both ends: it types nine public signatures, and it adds
the guard that keeps the rule true.

**Two scoping calls are now written into the invariant** rather than left
implicit, because leaving them implicit is what let the drift hide:

- **Library crates only.** `stella-cli` is a binary — a `String` it prints and
  exits on is the finished product, not an unnamed error. Its 421 signatures
  were never violations. An audit that counted them reached a number five times
  the real one, which is precisely why the rule looked hopeless to enforce.
- **`pub` only.** The hazard invariant #5 names is a *caller* that cannot branch
  without parsing prose, and only the public surface has callers outside the
  crate. Widening to internal helpers is flagged as an open question in #2392,
  not silently assumed either way.

The honest count, after fixing two false-positive classes in my own scanner
(below): **109** library signatures, **52** of them `pub`. This PR types 9 and
ratchets the remaining 43.

### Converted

| Crate | What was collapsed |
|---|---|
| `stella-tools::input` (6 fns) | The module exists to separate "absent" from "present but wrong type", but kept the distinction *only in the message string* — so a caller telling them apart had to match on prose the doc comments promise never to change. |
| `stella-core::goal::assess` | `SubAgentOutcome::Incomplete` and `Refused` folded into one `String`. Different facts about the run; a caller deciding whether to retry wants only the first. |
| `stella-graph::manifest::load` | Parse failure vs. the `Ok(None)` absent case. |
| `stella-store::CloudDrainConfig::wire_format` | "Fails closed" is a security posture; a caller must tell it from a transport error. |

`Display` is preserved **byte for byte** everywhere the string reaches a model
or a user — the tool errors are read by the model, and `GoalAssessError`'s two
callers wrap it in their own `"verifier unavailable: {}"` prefix, so giving the
variants distinct wording would only stutter in what a user actually reads.

There is no issue this PR closes: the contradiction was found in an audit and
fixed in the same pass. It advances #2392 (`Refs`), which tracks the remaining
conversions.

### A third commit that does not belong to this topic

`refactor(arenabench): split the transcript reader out of telemetry.py`.

`arenabench/arenabench/telemetry.py` crossed the 1500-line ratchet at 1534 lines
in 76c7aae (#2391), turning `file size ratchet` red on `main` and, through the
merge result, on **every** open PR. AGENTS.md says not to fold a ratchet fix you
did not cause into an unrelated PR; that guidance was overridden deliberately by
the maintainer to unblock CI, and it is called out here rather than buried.

The seam is not arbitrary — the transcript half shares **no** module-level name
with the metrics half (verified by name, not by eye), so the two were one file
by habit:

```
telemetry.py   1534 -> 1121
transcript.py          437  (new)
```

No re-export shim: `TranscriptReader` moves and its four importers move with it.
The guard's own gap that made this fail innocent PRs is #2397, and is
deliberately *not* fixed here.

> **Correction:** that commit's message says "579 passed, 6 skipped". The real
> figure is **573 passed, 6 skipped** — 579 is the total, and I folded the skips
> into the passed count. The commit message is already pushed and sits behind a
> merge commit, so it is corrected here rather than by rewriting history; worth
> editing out of the squash body on merge.

## The witness

`input::tests::the_two_failures_are_told_apart_without_reading_the_message`
(`crates/stella-tools/src/input.rs`) matches `InputError::Missing` and
`InputError::WrongType` as *variants*. Against the old `Result<_, String>`
signature there is no variant to match and the test does not compile — the
strongest form of "fails on the old code".

The guard itself was verified against its own failure modes rather than only
its passing one:

```
grown        -> exit=1   (added a pub fn returning Result<_, String>)
--update     -> exit=1   "refusing to raise stella-tui from 1 to 2"
restored     -> exit=0
shrunk       -> exit=0   + note: "run make typed-errors-update to lock the win in"
```

The arenabench split has no witness and needs none — it is a pure move, and the
existing suite is the check that it was one.

## The gate

- [x] `cargo fmt --check`
- [x] `cargo clippy --workspace --all-targets -- -D warnings` — clean
- [x] `cargo test --workspace` — 128/128 binaries green
- [x] `arenabench` pytest — 573 passed, 6 skipped
- [x] CI on this head: **13 checks green**, 1 skipped (`cla`), 0 failing
- [x] Docs updated (AGENTS.md invariant #5, AGENTS.md + CONTRIBUTING.md gate lists)

One caveat, stated plainly: `shellcheck` is not installed in this container, so
the `shellcheck` gate step did not run locally. My one `.sh` edit is a two-line
addition to `check-gate-parity.sh`'s `case` tables.

## Nothing left behind

- [x] Filed: #2392, #2393, #2397

- **#2392** — the remaining 43 conversions, written as a fresh-agent handoff:
  the exact list by cluster, the `Display`-is-a-contract constraint, the god-file
  constraint on `registry.rs`/`scripts.rs`, and a definition of done that
  includes **deleting the ratchet machinery**, not just emptying the baseline.
- **#2393** — `daemon::tests::a_run_that_ignores_term_is_killed_after_the_grace_period`
  is flaky under full-suite load. I ran the cheap control before filing: 3/3
  passing alone, workspace suite green on re-run, so it is load-dependent and
  not a regression from this change.
- **#2397** — the `file-size` base-relative ratchet (#2004) only covers
  grandfathered files, so a first-time crossing on `main` fails every open PR.
  Found by this PR's own CI failure. Splitting `telemetry.py` unblocks CI; it
  does not fix the guard, and the guard is the actual bug.

## Ground-rule check

- [x] One new dep edge: `thiserror` joins `stella-store` — already the
      workspace's one error-derive crate, used by ten siblings. No new crate
      enters the tree.

**My first scanner was wrong twice, in opposite directions, and the guard is
shaped by that.** The obvious regex is

```
Result<[^>]*,\s*String\s*>
```

and it reports both of these as violations, wrongly:

```rust
// The `, String>` the regex matches is INSIDE the generic argument.
pub fn artifacts(&self) -> Result<HashMap<String, String>, WitnessArtifactError>

// A crate `Result<T>` alias whose error type is defaulted — one argument, not two.
pub fn load(path: &Path) -> Result<String>
```

Both were live while I was auditing: they inflated `stella-store` from 1 to 17
and invented a violation in `stella-pipeline` that does not exist. So the guard
parses bracket-depth-aware and counts only a `Result` with exactly two
top-level generic arguments whose second is literally `String`. A guard that
cries wolf gets `#[allow]`-ed, and this one would have.

**On the ratchet.** CLAUDE.md is right to be hostile to baselines, so: this one
is legitimate for the single reason that the *rule predates the guard* — it
records a debt that already existed rather than granting new permission. It
refuses to raise a count (`--update` errors), so the only way past it is to type
the signature. It is meant to reach empty and be deleted along with its own
machinery, which is written into #2392's definition of done rather than left as
good intentions.

**Rejected alternative:** amending invariant #5 to bless a bare `String` error
in library code. That would have turned the gate green in one line and made the
document agree with the tree — by giving up the rule. The precision the
invariant was missing was *scope* (which crates, which visibility), not
*strictness*.


macanderson:
CI red on `file size ratchet`, failing on the base branch too — not this PR.

`arenabench/arenabench/telemetry.py` went 1474 → 1534 in 76c7aae (#2391), which
merged about ten minutes before this branch was pushed. This PR touches no file
under `arenabench/` (`git diff --name-only 47b7bab..HEAD -- arenabench/` is
empty).

The part worth flagging: the base-relative ratchet from #2004 is supposed to
make exactly this a non-failure — `current > max(ceiling, size at base)`. It
does not fire here, because that comparison only runs for files that already
have a baseline entry. A first-time crossing is emitted as `NEWOVER` and falls
into the classifier's catch-all, where the base tree is never consulted. Proof,
with the base pinned to the very commit that contains the 1534-line file:

```console
$ git show origin/main:arenabench/arenabench/telemetry.py > arenabench/arenabench/telemetry.py
$ FILE_SIZE_BASE_REF=origin/main ./scripts/check-file-size.sh
check-file-size: FAILED
exit=1
```

The file is byte-identical to the base and still fails, so every open PR is red
on this check right now.

Filed as #2397 with the diagnosis and a three-case regression test. Not fixing
it here: AGENTS.md § "God files" says not to fold a ratchet fix you did not
cause into an unrelated PR, and unblocking `main` means splitting that Python
file, which belongs with arenabench. Will re-run once `main` recovers.

Everything this PR owns is green locally — `cargo fmt --check`, `clippy
--workspace --all-targets -D warnings`, and `cargo test --workspace` (128/128
binaries).

macanderson:
The unblock is up as #2405 — it closes the #2397 gap you diagnosed here (the
`max(ceiling, size at base)` allowance only ran for records that had a baseline
ceiling; a `NEWOVER` record fell into the catch-all and never consulted the base).

Once that lands on `main`, this PR's `file size ratchet` goes green on a re-run
with no change here: `arenabench/arenabench/telemetry.py` is byte-identical to
your base, so it reports as inherited drift instead of failing. The split of that
file — the other half of #2397 — is filed separately as #2403 and is not a
prerequisite for you.

Issue #2398, opened 2026-08-09
context: AgentEvent::ContextRecall carries no `no_evidence_cut`, so the evidence gate's refusal count is unmeasurable from the trace (#2382)


macanderson:
## Problem

The recall evidence gate (`require_evidence`, #2289) refuses candidates that carry **no query-conditional evidence** and counts them in `RetrievalResult::no_evidence_cut` (`crates/stella-context/src/retrieval.rs:505`). That count **never reaches the wire**. `AgentEvent::ContextRecall` (`crates/stella-protocol/src/event.rs:699`) carries `frames`, `provider_mix`, `tokens`, `usage`, `latency_ms`, `used_ann_index` — and nothing about the gate.

So a trace-only consumer cannot measure the gate. #2382's F8 card asks for exactly that number ("`no_evidence_cut` refusal count"), and the attribution counters that landed for #2382 (`bench/loop-bench/src/features.rs`) deliberately do **not** report it: the nearest thing on the wire is `usage.providers[].frames_rejected`, which is the CGP *host* rejecting a provider's frame (a misdeclared cost, a consent gate, a failed leg) — a fact about plumbing, not about relevance. Counting that and labelling it the evidence cut would be a different number wearing the name of the one the experiment asked for, which is the exact measurement artifact CLAUDE.md's bench rule exists to prevent.

Also note the emitter drops the recall entirely when it admitted nothing: `ContextRecallOutcome::telemetry_event` (`crates/stella-pipeline/src/ports.rs:190`) returns `None` for `frames.is_empty()`. A recall where the gate refused *every* candidate is therefore invisible in the stream — indistinguishable from a turn that never recalled at all, which is the single most interesting case for grading the gate.

## Files

- `crates/stella-protocol/src/event.rs` — `AgentEvent::ContextRecall` (~L699), the additive field goes here
- `crates/stella-context/src/retrieval.rs:505` — `RetrievalResult::no_evidence_cut`, the source
- `crates/stella-pipeline/src/ports.rs:190` — `telemetry_event`, which builds the event and currently suppresses the empty case
- `crates/stella-cli/src/memory/recall.rs` — the CLI recall path
- `bench/loop-bench/src/features.rs` — where the counter would be distilled (`recall.no_evidence_cut`, beside the existing `recall.*` keys) and `bench/loop-bench/README.md` § Feature attribution, which currently documents the absence
- `docs/wire/agentevent.schema.json` — regenerated by `make wire-schema`, not hand-edited

[...]

Issue #2374, opened 2026-08-08
prove value


macanderson:
thesis: determinism where determinism will do produces more reliable and efficiernt intelligence

i have never seen the list below of features make a positive impact yet

self driving mode
tool foundry
code graph
testing tool foundry with cc traces
repo overview
triage agent
verifier agent
context record retrieval
adaptive context lifecycle
pipeline mode
subagents
stella fleet
skill mining
fact rule promotion

--

where we are very weak: software developer lifecycle, creating prs/worktrees and working in rythym with developer drivers.

macanderson:
# The prove-value plan

**Thesis under test:** determinism where determinism will do produces more reliable and efficient intelligence.

**Standing rule while this plan runs — FEATURE FREEZE.** No new features until every feature below either proves its value or is deliberately cut. In scope: fixing, hardening, instrumenting, measuring, demonstrating. Out of scope: anything that adds surface area. The single exception is named in Phase 0 — the ablation switches and attribution counters that make measurement possible at all — and it is called out there explicitly rather than smuggled in.

This plan is the tracking artifact. Each phase ends with a verdict per feature written back into this issue.

---

## 1. Why nothing has shown value yet

The honest diagnosis is not "these features are worthless." It is that **most of them have never been measured, and several are provably broken or starved in ways that make their value invisible even when it is real.** Four patterns, all evidenced:

**Silent starvation.** The learning plane — skill mining, fact/rule promotion, context lifecycle — hangs off one JSON parse. Reflection routes to the triage model with `reasoning` unset; a reasoning-pinned triage model blows the 2048-token cap, truncates, yields zero lessons, and `reflect_and_record` returns early before anything is written. Measured frozen **nine days**: 8 reflection rows against 85 executions. The failure *is* detected and then thrown away to `eprintln!`, so the execution row still reads `outcome='completed'` and the dashboard shows a healthy loop. A workspace that has learned nothing for nine days is indistinguishable from a fresh one. (#2174, #2175, #2277)

[...]

Issue #2397, opened 2026-08-09
file-size ratchet: a first-time crossing on main fails every open PR — the base-relative check (#2004) only covers grandfathered files


macanderson:
## Problem

`main` is red on the `file size ratchet` check, and so is every open PR,
including PRs that do not touch the offending file.

```
These files crossed the limit and are NOT grandfathered — split them into
submodules. Do not add a baseline entry: ...
  arenabench/arenabench/telemetry.py is 1534 lines, over the 1500-line limit
```

`arenabench/arenabench/telemetry.py` went 1474 → 1534 in 76c7aae (#2391). It is
not in `scripts/file-size-baseline.txt`, so it is a **first-time crossing**.

## Root cause

#2004 added the base-relative ratchet so the guard "judges your change, not the
tree" — `current > max(ceiling, size at base)`. That comparison is applied in
**only one branch** of the classifier in `scripts/check-file-size.sh`:

```sh
  "GREWCAND "*)
    cand_effective="$cand_ceiling"
    if [ -n "$base_commit" ]; then
      cand_base="$(size_at_base "$cand_path")"
      if [ "$cand_base" -gt "$cand_effective" ]; then
        cand_effective="$cand_base"      # <-- the #2004 allowance
      fi
    fi
    ...
  *)
    report="${report}${line}"            # <-- NEWOVER lands here, no base check
```

`GREWCAND` records are files that already have a baseline ceiling. A file with
**no** baseline entry is emitted as `NEWOVER` and falls into the catch-all,
where the base tree is never consulted. So the file is byte-identical to the
base and still fails.

## Proof (already run — do not repeat)

From a branch that does not touch `arenabench/`, with the base pinned to the
very commit that contains the 1534-line file:

```bash
git show origin/main:arenabench/arenabench/telemetry.py > arenabench/arenabench/telemetry.py
FILE_SIZE_BASE_REF=origin/main ./scripts/check-file-size.sh
# -> FAILED: arenabench/arenabench/telemetry.py is 1534 lines, over the 1500-line limit
# exit=1
```

`max(1500, 1534) = 1534` and `current = 1534`, so `1534 > 1534` is false and
this must pass. It fails, because that arithmetic never runs for this record
type.

## Why it matters

This is precisely the failure mode #2004 was written to prevent, and the reason
given there applies unchanged: the baseline is one shared cell every growing PR
must write, and a red ratchet nobody caused makes every author either wait or
[...]

PR #2399, opened 2026-08-09
feat(loop-bench): per-feature attribution counters in the A/B report (#2382)

### LINKED ISSUE #2382: bench(loop-bench): only project_overview and graph_query are attributed, and they share one bucket — recall, subagents, stages and foundry tools have events but no counter (#2374)
## Problem

`bench/loop-bench` distills `stella-events.jsonl` into a per-trial summary and a `ComparisonReport`, but it attributes only two features: it counts `project_overview` and `graph_query` together as "context queries" (`bench/loop-bench/src/artifacts.rs`, `bench/loop-bench/README.md` § Key concepts). Everything else in the prove-value plan (#2374) has rich events in the stream and **no counter in the comparison report**, so a paired A/B produces no per-feature number without hand-written analysis each time.

The events already exist and carry what is needed (`crates/stella-protocol/src/event.rs`):

- `AgentEvent::ContextRecall { frames, provider_mix, tokens, usage, latency_ms, used_ann_index }` — recall injections (F8, F9)
- `AgentEvent::ContextWrite { provider, upserts, superseded }` — context write-back (F9, F14)
- `AgentEvent::SubAgent { phase }` — subagent spawns (F11)
- `AgentEvent::Stage { name: StageKind }` — per-stage attribution for triage / witness / verify (F6, F7, F10)
- `ToolStart` / `ToolResult` — foundry tool calls and `task` dispatches (F2, F11)

Lumping `project_overview` and `graph_query` into one "context queries" bucket is itself a defect for the plan: F3 and F5 are separate cards with separate kill criteria and cannot share a counter.

## Files

- `bench/loop-bench/src/artifacts.rs` — the distiller; `EVENTS_NAME = "stella-events.jsonl"` is around L36
- `bench/loop-bench/src/compare.rs` and `compare/` — the A/B report shape
- `bench/loop-bench/src/lib.rs`, `main.rs` — `--compare a,b` and `--json-out`
- `bench/loop-bench/README.md` — the § Key concepts table that documents the counters
- `crates/stella-core` — `ComparisonReport` is defined here and is already consumed by `stella tune effort`, so a field addition has a second reader; check it
- `bench/telemetry_store/schema.sql` and `ingest.py` — the offline warehouse ingests the same file (L264) and should stay consistent
- `bench/evidence/<experiment>/analyze.py` — the existing per-experiment analyzer pattern (e.g. `judge-evidence-demand-1295/analyze.py`)

## What "done" looks like

1. `project_overview` and `graph_query` are counted **separately**, and the combined "context queries" figure is either kept as a derived sum or removed — not left as the only view.
2. New per-feature counters distilled from the events above: recall frames admitted (and `no_evidence_cut` refusals), context writes, subagent spawns, per-`StageKind` stage counts, and foundry-tool invocations.
3. `--compare a,b` reports each counter for both arms with the delta, so a paired ablation yields a per-feature number directly.
4. `ComparisonReport`'s existing consumers still compile and behave identically — new fields are additive, and `stella tune effort` is checked, not assumed.
5. The README's § Key concepts table is updated in the same PR, since it is the documented contract for what the harness measures.

## Witness

A test over a small committed `stella-events.jsonl` fixture asserting the distiller reports the new counters with known values — failing on `main` because the fields do not exist. `bench/loop-bench/src/tests.rs` already has the shape to follow.

## Constraints

- Additive only: an existing counter must not change meaning, or every previously recorded comparison silently becomes incomparable.
- `make bench-test` (Makefile) covers the Python bench tooling; the Rust side rides `cargo test`.
- Do not widen this into a new analysis surface — it is a counter extension for the #2374 freeze exception (Phase 0, blocker 0.7), not a dashboard.

## Why now

Without it, every feature card in #2374 outside F3 and F5 needs bespoke jq/Python per experiment, which is exactly where measurement artifacts creep in. Extending the existing counter is cheaper and more auditable than fourteen one-off analyzers.

Refs #2374

macanderson:
## What & why

`loop-bench --compare` attributed exactly two features — `project_overview` and `graph_query` — so every other feature in the #2374 plan had rich events in `stella-events.jsonl` and **no counter in the comparison report**. A paired ablation of recall, of the witness stage, of subagents, of an adopted foundry tool produced no per-feature number without a bespoke `jq`/Python pass written per experiment, run *beside* the report rather than inside it. A figure computed over a different trial set than the report's pass rate is not comparable with it, and nothing on the page says so — which is exactly where a measurement artifact creeps in.

Every trial now carries a counter map distilled from the same stream, and `stella-core::comparison` folds it over the same **paired** trials as every other figure. A feature delta and a pass-rate delta in one report are therefore statements about one population.

```
FEATURE ATTRIBUTION — per-trial mean (total), over the paired trials only
feature                        recall-off      recall-on   Δ recall-on
──────────────────────────────────────────────────────────────────────
recall.calls                     0.00 (0)       1.00 (6)        +1.000
recall.frames                    0.00 (0)      3.50 (21)        +3.500
recall.frames_rejected           0.00 (0)       0.00 (0)        +0.000
stage.execute                   2.00 (12)       1.00 (6)        -1.000
subagent.spawned                 0.00 (0)       1.00 (6)        +1.000
tool.graph_query                 0.00 (0)      2.00 (12)        +2.000
tool.grep                       4.00 (24)       0.00 (0)        -4.000
```

**The vocabulary** (`bench/loop-bench/src/features.rs`), keyed `<namespace>.<what>`:

| Key | One per | Card |
|---|---|---|
| `tool.<name>` | `tool_start`, keyed by the tool's own name | F2, F3, F5, F11 |
| `stage.<kind>` | `stage`, keyed by the canonical `StageKind` tag | F6, F7, F10 |
| `recall.calls` / `.frames` / `.tokens` / `.frames_rejected` | `context_recall` | F8 |
| `context_write.calls` / `.upserts` / `.superseded` | `context_write` | F9, F14 |
| `subagent.spawned` / `.refused` | `sub_agent` `started`; a `finished` with status `refused` | F11 |

**Four decisions are load-bearing, and each is a defect if reversed:**

- **Tools are keyed by name, never by a class.** A foundry-authored tool is indistinguishable from a built-in on the wire — `ToolCall` carries a name and arguments and nothing about provenance — so a `tool.foundry` bucket would be the harness *guessing* from a name it does not own. Keying by name reports what the stream says. It also gives `project_overview` and `graph_query` separate keys, which #2382 item 1 asks for on its own account: F3 and F5 have separate kill criteria and cannot share a bucket.
- **Stages are canonicalized through `StageKind`.** The verdict stage shipped as `judge` and is still aliased that way, so a raw-string key would file one stage under two names and report each as having half-fired. This is also why the crate takes a `stella-protocol` dependency (below) rather than hardcoding a stage list, which would duplicate the vocabulary L-E1 exists to hold exactly once.
- **The key set is the union of both arms.** A counter only one arm emitted is the single most interesting row an ablation produces; an intersection deletes exactly that finding.
- **The delta is derived, never stored** — `candidate.mean − baseline.mean` at read time, so it cannot disagree with the two figures printed beside it. And it is a **per-trial mean**, because pairing excludes at trial granularity and the arm that folded more trials wins every raw total by construction.

The block prints **above the verdict and is never folded into it**: a feature delta is evidence about what changed, not a bar a candidate may be promoted on.

**Exemplar.** The opaque-counters-on-an-opaque-arm shape mirrors this crate's own `self_tuning::ArmSamples` — an id, a display value, and a bag of per-trial observations the engine folds without knowing what they mean. `stella-core::comparison::feature` knows nothing about `tool.graph_query`, exactly as it knows nothing about whether an arm's `value` names a model or a prompt template.

Closes #2382
Refs #2374

## The witness

Checked the artisanal way — a detached worktree at `HEAD` with **only** the three test files copied in:

```
error[E0609]: no field `features` on type `TrialReport`
error[E0599]: no method named `feature_keys` found for struct `ComparisonReport`
error[E0609]: no field `features` on type `&ArmSummary`
error[E0599]: no method named `feature_deltas` found for struct `ComparisonReport`
```

The set (10 new tests + 1 new property):

- `every_attributable_event_reaches_its_own_counter` — one stream carrying every attributable event, spelled as the wire spells it (`sub_agent`'s nested `phase` discriminator, `context_recall`'s `usage.providers[]`), asserting each counter's known value.
- `the_two_context_tools_are_counted_separately_and_agree_with_their_columns` — #2382 item 1.
- `a_stage_recorded_under_its_wire_alias_lands_in_the_canonical_bucket` — `judge`/`verifier`/`verdict` all fold to `stage.verdict`.
- `an_unknown_stage_is_counted_under_its_own_name`, `a_stage_entered_twice_counts_twice_where_the_stage_list_dedupes` (the revise back-edge), `a_multi_step_trials_counters_sum_across_its_steps`, `the_json_report_carries_the_feature_counters`.
- `a_paired_ablation_reports_each_feature_counter_with_its_delta`, `an_unpaired_trials_counters_are_excluded_with_the_rest_of_it`, `arms_that_counted_nothing_have_no_feature_keys` (loop-bench compare).
- `feature_counters_fold_into_a_total_and_a_per_trial_mean`, `the_feature_key_set_is_the_union_of_both_arms`, `an_excluded_trials_counters_are_excluded_too`, `a_missing_baseline_yields_no_feature_deltas`, `the_feature_channel_is_additive_on_the_wire` (stella-core).
- Property `a_feature_delta_always_re_derives_from_its_operands`, and `arb_trial` now generates counters so the existing determinism/totality properties cover them.
- `a_bench_report_carrying_feature_counters_still_reads` (stella-cli) — the second-reader check below.

No test was deleted or renamed in this PR.

## The gate

- [x] `cargo fmt --check`
- [x] `cargo test --workspace` — 1535 pass; **one pre-existing failure**, `daemon::tests::a_run_that_ignores_term_is_killed_after_the_grace_period`, already tracked as #2393. Confirmed with the cheap control: it fails identically on a stashed tree at the merge base, so it is not this change.
- [x] Every guard: `make invariants doc-links command-docs brand-case file-size god-files gate-parity left-behind role-names stat-portability module-reachability wire-schema` and `make doc-warnings`, all green. `shellcheck` is not installed in this environment and was the only step not run; this PR touches no shell.
- [x] Docs updated — `bench/loop-bench/README.md` gains § Feature attribution (the documented contract for what the harness measures, per #2382 item 5) and its § Key concepts paragraph no longer calls the two context tools one bucket.
- [x] `Closes #2382` appears both here and as a commit trailer.

## Nothing left behind

- [x] Filed: **#2398** — `AgentEvent::ContextRecall` carries no `no_evidence_cut`, so the evidence gate's refusal count is unmeasurable from the trace.

#2382 item 2 asks for the `no_evidence_cut` refusal count. **It is deliberately not implemented, and this is the one part of the issue's scope this PR does not deliver.** The count is real (`stella-context`'s `RetrievalResult`) but never reaches the wire: `AgentEvent::ContextRecall` carries the admitted frames and, via `usage`, the frames the CGP *host* rejected — a misdeclared cost, a consent gate, a failed leg — which is a fact about plumbing, not relevance. Counting `recall.frames_rejected` and labelling it the evidence cut would be a different number wearing the name of the one the experiment asked for. The gap is documented in the module docs and the README rather than papered over, and #2398 is written as a handoff (it also names a second finding: `telemetry_event` suppresses the event entirely when `frames.is_empty()`, so a recall where the gate refused *everything* is invisible in the stream — the single most interesting case for grading the gate).

Also linked, not duplicated: #2393 (the flaky daemon test above).

## Ground-rule check

- [x] No I/O added to `stella-core` — `comparison::feature` is pure folds over owned data.
- [x] New cross-boundary types round-trip through serde — `the_feature_channel_is_additive_on_the_wire` asserts the whole `ComparisonReport` re-serializes byte-identically (invariant #4).

**One new dependency**, `loop-bench` → `stella-protocol` (path). Justification: the distiller reads the wire contract and was doing so with magic strings; `StageKind` is the only import, and only to canonicalize a stage name off the stream. The alternative — a hardcoded stage-name list in the bench crate — both splits the aliased `verdict` bucket and duplicates the stage vocabulary that `stella-protocol` exists to hold exactly once (L-E1). It is already in the build graph beneath `stella-core`, which `loop-bench` depends on, so it costs no compile time.

**On the freeze.** #2374 names 0.7 as an explicit freeze exception, scoped to "extends an existing counter table in the bench harness". This PR adds no user-facing surface: no new CLI flag, no new command, no new output file. `--compare`'s existing rendering gains a block; the single-run `--json` gains a `features` key per row.

**On additivity — checked, not assumed** (#2382 item 4). `COMPARISON_SCHEMA_VERSION` does **not** move: the change is compatible, and bumping it would make every existing consumer refuse reports it can read fine. `TrialRecord.features` and `ArmSummary.features` are `#[serde(default, skip_serializing_if = …)]`, so a producer that counts nothing emits byte-identical JSON and a report recorded before today still deserializes. The second reader, `stella tune effort`, reads a deliberately narrow view of a trial row (`BenchTrial`) and ignores unknown keys — `a_bench_report_carrying_feature_counters_still_reads` proves it, and proves the resulting arm arithmetic is identical with and without the key, because the failure mode would be `stella tune effort` refusing every artifact produced after this lands.

**Where I chose not to widen.** Per-task feature rows (the `TaskRow` level) would be a dashboard, and #2382 explicitly rules that out. Likewise the deltas are computed, not stored — a stored delta is one more cell that can drift from its operands, and this repo has lost a `main` to that shape before.

**Cardinality.** `tool.<name>` keys come from model-supplied names, so they are truncated to 48 characters before becoming a map key. Truncation rather than a cap on distinct keys: dropping a counter silently is how a report claims coverage it does not have.

**The offline warehouse** (`bench/telemetry_store/`) needs no change to stay consistent — `ingest.py` already stores every event's whole payload in `events` plus a `tool_calls` row per call, so the same counters are derivable there in SQL from the same file. Noted in the README rather than left implicit.

---

## Summary by Sourcery

Add per-feature attribution counters to loop-bench trial reports and stella-core comparisons so A/B runs can report feature-level deltas from the same paired trials as the primary metrics.

New Features:
- Introduce a per-feature attribution channel that records per-trial counters (e.g., tools, stages, recall, context write, subagents) and exposes them through TrialReport and ComparisonReport.
- Render a FEATURE ATTRIBUTION table in loop-bench --compare output, showing per-arm per-feature means, totals, and baseline-relative deltas.
- Add support in stella-core for aggregating feature counters per arm and computing baseline-relative feature deltas for paired comparisons.

Enhancements:
- Ensure multi-step trials merge per-step feature counters into a single trial-level view and that attribution respects trial pairing logic.
- Constrain and canonicalize feature keys (e.g., tool names with length limits, stages via StageKind) to keep reports stable and forward-compatible.
- Extend property-based tests to cover feature counters and deltas, guaranteeing derived deltas stay consistent with their operands.

Documentation:
- Document the new feature attribution channel in the loop-bench README, including key vocabulary, semantics, and compatibility guarantees.

Tests:
- Add focused tests across loop-bench, stella-core comparison, and stella-cli to validate feature counting from event streams, aggregation behavior, JSON wire compatibility, and consumer resilience to the new keys.

Chores:
- Wire a new stella-protocol dependency into loop-bench to reuse the canonical stage vocabulary for attribution.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1606,6 +1606,7 @@ dependencies = [
  "serde",
  "serde_json",
  "stella-core",
+ "stella-protocol",
  "tempfile",
 ]
 
--- a/bench/loop-bench/Cargo.toml
+++ b/bench/loop-bench/Cargo.toml
@@ -28,6 +28,13 @@ path = "src/main.rs"
 # reading the same aggregates, the same guard set, and the same significance
 # test the offline harness applied, and two report types is two standards.
 stella-core = { path = "../../crates/stella-core" }
+# The wire vocabulary the distiller reads (#2382). Only `StageKind` is used,
+# and only to canonicalize a stage name off the stream: the verdict stage is
+# aliased `judge`/`verifier` on the wire, so a hardcoded name list here would
+# split one stage's counter in two AND duplicate the stage vocabulary the
+# protocol crate exists to hold exactly once (L-E1). Already in the build
+# graph beneath `stella-core`, so it costs no compile time.
+stella-protocol = { path = "../../crates/stella-protocol" }
 clap = { workspace = true }
 serde = { workspace = true }
 serde_json = { workspace = true }
--- a/bench/loop-bench/README.md
+++ b/bench/loop-bench/README.md
@@ -53,9 +53,13 @@ promotion decided by them is a decision about the machine.
 For each trial dir `<jobs-dir>/<job-name>/<task>__<id>/`, `distill_events` reads
 `agent/stella-events.jsonl`: `step_usage` counts model calls, `tool_start` tool
 calls (`write_file` / `edit_file` / `apply_edits` are writes — never
-`delete_file`, or a destructive loop would look productive; `project_overview` /
-`graph_query` are context queries), and a terminal event is `complete` or an
-`error` with `retryable: false`. The reward comes from `verifier/reward.txt`.
+`delete_file`, or a destructive loop would look productive; `project_overview`
+and `graph_query` each get their own column, `ov` and `gq`), and a terminal
+event is `complete` or an `error` with `retryable: false`. The reward comes from
+`verifier/reward.txt`.
+
+Beside those loop-health counters, every trial also carries a **per-feature
+attribution** map — see [§ Feature attribution](#feature-attribution-2382).
 
 Four more files in the same directory are read, each closing a hole where
 harbor knew something the harness did not (#1299) — see
@@ -202,6 +206,81 @@ promote however cleanly the arms separate — it reports
 into a gate (exit `6`); it is off by default because a comparison is a
 measurement and "no winner" is a legitimate answer to it.
 
+## Feature attribution (#2382)
+
+The aggregates above answer *which arm won*. They cannot answer **which feature
+moved**, and that is the question a paired ablation is run to settle. Until
+#2382 the harness attributed exactly two features — `project_overview` and
+`graph_query` — so every other feature in the tree needed a bespoke `jq` pass
+per experiment, written beside the report rather than inside it. A figure
+computed over a *different* trial set than the report's pass rate is not
+comparable with it, and nothing on the page says so; that is precisely where a
+measurement artifact creeps in.
+
+So every trial now also carries a counter map, distilled by
+[`src/features.rs`](src/features.rs) from the same stream, and folded by
+`stella-core` over the same **paired** trials as every other figure. Each key is
+`<namespace>.<what>`, and the value counts exactly what the key names:
+
+| Key | One per | Card |
+|---|---|---|
+| `tool.<name>` | `tool_start`, keyed by the tool's own name | F2, F3, F5, F11 |
+| `stage.<kind>` | `stage`, keyed by the canonical `StageKind` tag | F6, F7, F10 |
+| `recall.calls` / `.frames` / `.tokens` / `.frames_rejected` | `context_recall` — the event, its admitted frames, its injected tokens, and the frames the host rejected whole | F8 |
+| `context_write.calls` / `.upserts` / `.superseded` | `context_write` and its two counts | F9, F14 |
+| `subagent.spawned` / `.refused` | `sub_agent` `started`, and a `finished` whose status is `refused` | F11 |
+
+`--compare` prints them per arm with the baseline-relative delta, above the
+verdict and never folded into it — a feature delta is evidence about *what
+changed*, not a bar a candidate may be promoted on:
+
+```
+FEATURE ATTRIBUTION — per-trial mean (total), over the paired trials only
+feature                        recall-off      recall-on    Δ recall-on
+──────────────────────────────────────────────────────────────────────────
+recall.frames                     0.00 (0)      3.00 (18)        +3.000
+tool.grep                         2.00 (12)      0.00 (0)        -2.000
+```
+
+Four decisions there are load-bearing:
+
+- **Tools are keyed by name, never by a class.** A foundry-authored tool is
+  indistinguishable from a built-in on the wire — `ToolCall` carries a name and
+  arguments and nothing about where the tool came from — so a `tool.foundry`
+  bucket would be the harness *guessing* from a name it does not own. Keying by
+  name reports what the stream says, and an operator ablating their adopted tool
+  knows its name.
+- **Stages are canonicalized through `StageKind`.** The verdict stage shipped on
+  the wire as `judge` and is still aliased that way, so a raw-string key would
+  file one stage under two names and report each as having half-fired.
+- **The key set is the union of both arms.** A counter only one arm emitted is
+  the single most interesting row an ablation produces; an intersection would
+  delete exactly that finding. An arm that never counted a key reads as zero
+  occurrences, which is what it is — the counters are folded over trials that
+  *ran*.
+- **The delta is derived, never stored.** It is `candidate.mean − baseline.mean`
+  computed at read time, so it cannot disagree with the two figures printed
+  beside it.
+
+`no_evidence_cut` — recall's evidence-gate refusal count — is **deliberately
+absent**, though #2382 asks for it. It is real (`stella-context`'s
+`RetrievalResult`) but never reaches the wire: `AgentEvent::ContextRecall`
+carries the admitted frames and, via `usage`, the ones the *host* rejected, and
+neither is the evidence gate's refusal. Counting `recall.frames_rejected` and
+labelling it the evidence cut would be a different number wearing the name of
+the one the experiment asked for. Tracked as #2398.
+
+The counters are **additive**: no existing counter changed meaning, the report's
+`COMPARISON_SCHEMA_VERSION` does not move, and an arm that counts nothing
+serializes exactly as it did before. `stella tune effort` reads a deliberately
+narrow view of a trial row and ignores the new key — checked in
+`crates/stella-cli/src/memory/self_tuning.rs`, not assumed.
+
+The offline warehouse ([`../telemetry_store/`](../telemetry_store/)) needs no
+change to stay consistent: `ingest.py` stores every event's whole payload in
+`events` (plus a `tool_calls` row per call), so the same counters are derivable
+there in SQL from the same file.
+
 ## The nightly CI gate (#873)
 
 [`.github/workflows/nightly-bench.yml`](../../.github/workflows/nightly-bench.yml)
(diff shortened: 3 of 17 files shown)


macanderson:
## CI status — both reds are one pre-existing failure on `main`, not this change

Both failing checks are the **same** step, and it is the file-size ratchet firing on a file this branch does not touch:

```
```

Already tracked: **#2403** (split `telemetry.py`) and **#2397** (the guard defect that turns it into everyone's failure — the base-relative check from #2004 only classifies *grandfathered* files, so a first-time crossing takes the unconditional `NEWOVER` branch and fails every open PR). Not filing duplicates.

**It predates this branch.** `telemetry.py` was 1474 lines at my branch point (`9de2ddb`) and 1534 on `main`; it crossed in the arenabench work, and every commit on `main` from `76c7aae` onward has a red `ci` run. The blob is byte-identical between `origin/main` and this branch.

**Not fixing it here, deliberately.** AGENTS.md § God files: *"do not 'fix' a red ratchet you did not cause by folding a baseline regeneration into an unrelated PR — land it on its own from a fresh `main`."* And a baseline entry is not the fix anyway — the baseline accepts no new entries by its own header (that is #2404); `telemetry.py` has to be split.

### Everything else on this PR is green

Step 10 is the only failure in the required job; the workflow ran every later step and all passed:

| Step | |
|---|---|
| `Cargo.lock` in sync | ✅ |
| `cargo fmt --check` | ✅ |
| `docs/wire/` still matches the types | ✅ |
| `cargo doc -D warnings` | ✅ |
| prompt-cache golden fixtures | ✅ |
| **`cargo test`** | ✅ |
| release smoke build (thin LTO) | ✅ |
| deleted-test guard | ✅ |
| every other guard (16 steps) | ✅ |
| `cargo deny` job | ✅ |

Worth noting for the record: `cargo test` passed on CI, so the `daemon::tests::a_run_that_ignores_term_is_killed_after_the_grace_period` failure I saw locally did not reproduce there — consistent with #2393 calling it load-dependent.

I'll re-check when `main` goes green and merge it in so CI re-runs against the fixed base.

Issue #2404, opened 2026-08-09
`make file-size-update` silently grandfathers a first-time crossing, which the baseline's own header forbids


macanderson:
## Problem

`scripts/check-file-size.sh --update` regenerates the baseline from *every*
tracked file currently over the limit. It does not distinguish a file that
already had an entry from one that has never had one, so it happily writes a
brand-new entry for a file that just crossed 1500 lines.

That contradicts the rule stated in three places, including the header
`--update` itself writes into the file:

> No entry may be ADDED for a new file: a file that crosses the limit must be
> split instead.

and AGENTS.md § "God files":

> A new file that crosses 1500 lines fails the gate outright, and the baseline
> accepts no new entries.

## Reproduce

Hermetic, no dependence on the tree's current state (verified against `main` at
a5e31ab, with #2405 merged — the hole is in the `--update` branch, which #2405
did not touch):

```bash
cd "$(mktemp -d)" && mkdir -p scripts src && git init -q
git config user.email t@t.invalid && git config user.name t
cp /path/to/stella/scripts/check-file-size.sh scripts/
printf '# test baseline\n1600 src/old.rs\n' > scripts/file-size-baseline.txt
awk 'BEGIN{for(i=0;i<1600;i++)print "x"}' > src/old.rs         # grandfathered
awk 'BEGIN{for(i=0;i<1550;i++)print "x"}' > src/brand_new.rs   # first-time crossing
git add -A && git commit -q -m base

./scripts/check-file-size.sh --update
grep brand_new scripts/file-size-baseline.txt
```

Observed:

```
check-file-size: baseline updated — 2 grandfathered file(s) over 1500 lines.
1550 src/brand_new.rs
```

The first-time crossing is now permanently grandfathered. The next
`check-file-size.sh` run reports OK, and the file is free to sit at 1550 lines
forever.

## Why it matters

This is the exact escape hatch the ratchet's design argument rules out, and it
is one command away from any author staring at a red gate. It got sharper with
#2405: the guard now tells an author with inherited drift to "split it", while
the tool sitting next to it will make the message go away without splitting
anything. A rule the tooling will help you break is not enforced.

It nearly mattered in practice — during #2405 I ran `--update` to check it still
worked and it silently added `arenabench/arenabench/telemetry.py` as a 31st
entry. Taking that output would have grandfathered a 1534-line file and closed
#2397 the wrong way.

## Fix

In the `--update` branch of `scripts/check-file-size.sh`, refuse to write an
[...]

Issue #2403, opened 2026-08-09
arenabench/arenabench/telemetry.py is 1534 lines — split it below the 1500-line ratchet


macanderson:
## Resolution

Already done — closing. #2394's third commit
(`refactor(arenabench): split the transcript reader out of telemetry.py`) landed
this while I was writing the issue, so it was never actionable:

```
telemetry.py   1534 -> 1121
transcript.py          437  (new)
```

Verified on `main` at a5e31ab, with both #2394 and #2405 merged:

```console
$ wc -l arenabench/arenabench/telemetry.py arenabench/arenabench/transcript.py
 1121 arenabench/arenabench/telemetry.py
  437 arenabench/arenabench/transcript.py

$ ./scripts/check-file-size.sh
check-file-size: OK — 1216 Rust/Python/shell files, 30 grandfathered over 1500 lines, and nothing went over by this change. Judged against eae8d12.
exit=0
```

No drift line, no baseline entry added, and `harbor_adapter + analyzer pytest`
was green on #2394. All three definition-of-done bullets hold.

Both halves of #2397 are therefore closed: the file by #2394, the guard gap by
#2405. Note that #2394 folded the split into an unrelated PR, which AGENTS.md
§ "God files" normally forbids — that PR calls the override out explicitly as a
maintainer decision to unblock CI rather than leaving it buried, which is the
right handling of it.

The one thing that survives from this issue is #2404: `make file-size-update`
will still write a baseline entry for a first-time crossing, so the constraint
this issue was written around is unenforced. That is tracked separately and is
still live.

PR #2405, opened 2026-08-09
fix(gate): the file-size ratchet judges a first-time crossing against the base too

### LINKED ISSUE #2397: file-size ratchet: a first-time crossing on main fails every open PR — the base-relative check (#2004) only covers grandfathered files

macanderson:
## What & why

`main` is red on the `file size ratchet` check, and so is every open PR —
including PRs whose tree is byte-identical to the base. #2394 (typed errors,
touching no file under `arenabench/`) is blocked on this alone.

#2004 made the ratchet judge *the change*, not the tree —
`current > max(ceiling, size at base)`. But that arithmetic ran in only **one**
branch of the classifier in `scripts/check-file-size.sh`. A file with a baseline
entry was emitted as a `GREWCAND` candidate and got the allowance; a file with
**no** entry was emitted as a finished `NEWOVER` verdict and fell into the
catch-all, where the base tree is never consulted.

So `arenabench/arenabench/telemetry.py` going 1474 → 1534 in 76c7aae (#2391)
failed every branch that walked past it. `max(1500, 1534) = 1534` and
`current = 1534`, so `1534 > 1534` is false and this must pass — it failed
because that comparison never ran for this record type.

Both candidate kinds now run through one `effective_limit` helper, differing
only in the limit passed in: the baseline entry for a grandfathered file, the
1500-line limit for every other. **The asymmetry is preserved** — a file this
change pushes over still fails, growth on top of inherited drift still fails,
and an inherited crossing is reported as drift rather than absorbed into the
baseline.

The drift lines now name their own remedy — `regenerate` for a grandfathered
file whose ceiling drifted, `split it` for a first-time crossing. The old
footer prescribed `make file-size-update` for every case, which is the one
action a first-time crossing must never take.

Closes #2397

## The witness

`B7 a first-time crossing inherited from the base is drift, not this change's
failure` in `scripts/test-file-size.sh`. Checked the artisanal way — the new
tests run against the old guard:

```
--- old guard, new tests ---
FAIL B7 ... — expected OK, got:
check-file-size: FAILED
  src/big.rs is 1600 lines, over the 1500-line limit
ok   B8 a file this change pushes over the limit still fails
ok   B9 growth on top of an inherited first-time crossing still fails
passed 15, failed 1
```

B8 and B9 passing on **both** guards is the point of them: they are the
anti-hole cases, and a fix that broke them would show up as a green they cannot
give. The three cases are the three #2397 asks for.

`want` in the harness now checks its substring on the passing path as well.
Without that, B7 would have been satisfied by a guard that passed *by forgetting
the file existed* — the drift note is a real finding that happens not to be
fatal, and asserting only the exit code would not have distinguished the fix
from that regression.

Verified on the real tree, which is byte-identical to its base for this file:

```console
check-file-size: baseline drift (not caused by this change, not fatal)
  arenabench/arenabench/telemetry.py is 1534 lines, over the 1500-line limit — already so at the base; split it
...
check-file-size: OK — 1214 Rust/Python/shell files, 30 grandfathered over 1500 lines, and nothing went over by this change. Judged against cd12845. 1 file(s) carry inherited drift (see above).
exit=0
```

## The gate

- [x] `cargo fmt --check`
- [x] `make guards-fast` — every toolchain-free guard green, `shellcheck`
      included (it is installed in this container and was run)
- [x] `./scripts/test-file-size.sh` — 16/16
- [x] `cargo clippy` / `cargo test --workspace` — not run: this change touches
      no Rust. `scripts/impacted-crates.sh` puts a shell+markdown diff on the
      `guards-fast` rung, which is what ran.
- [x] Docs updated — AGENTS.md § "God files" stated the rule as
      `max(ceiling, …)`, which was only ever true for grandfathered files
- [x] `Closes #2397` appears both here and as a commit trailer

## Nothing left behind

- [x] Filed: #2403, #2404

- **#2403** — `arenabench/arenabench/telemetry.py` is still 1534 lines. This PR
  makes the tree *survive* that, not clear it; the split is the other half of
  #2397 and belongs with arenabench, per AGENTS.md § "God files" on not folding
  a ratchet fix you did not cause into an unrelated PR.
- **#2404** — `make file-size-update` will happily write a baseline entry for a
  first-time crossing, which the header it writes into explicitly forbids. Found
  while verifying `--update` still worked here (it added telemetry.py as a 31st
  entry). One command away from any author staring at a red gate, and this PR
  makes it sharper by telling that author to "split it" while the tool next to
  them offers to make the message disappear.

## Ground-rule check

- [x] No I/O added to `stella-core` — no Rust changed at all
- [x] No new deps

**Why this is not a weakening of the ratchet.** The worry with any "already
broken at the base" rule is that drift becomes a licence to bloat, which is
worse than the red `main` it fixes. That is exactly what B4 pins for
grandfathered files, and B9 now pins the same asymmetry for non-grandfathered
ones: a file at 1600 in the base that this change grows to 1650 still fails,
against the base size and not the limit. What the change permits is strictly
"leave it exactly as you found it".

**The one behavior that is genuinely new and permanent.** An inherited
first-time crossing is now reported on every run rather than failing, so the
`telemetry.py` line will print until #2403 lands. That is the intended end
state per #2397 — drift is loud and non-fatal — but it does mean the report is
the only pressure on the file, which is why #2403 exists rather than a comment.

**Rejected alternative:** raising the ceiling for `telemetry.py` by adding a
baseline entry, or splitting the Python file here. The first is forbidden
outright (the baseline takes no new entries) and would have hidden the guard
bug rather than fixed it — the guard would then have been wrong in the same way
for the *next* first-time crossing. The second is a real fix but a different
PR's: it is arenabench's change, it must land from a fresh `main`, and folding
it in here would be the exact move AGENTS.md names.

**Green summary line reworded.** It claimed "none over 1500 lines except N
grandfathered", which inherited drift makes false. It now says what it can
stand behind: N grandfathered, and nothing went over *by this change*.

---

## Summary by Sourcery

Adjust file size ratchet to judge both grandfathered and non-grandfathered files against the base tree, treating inherited first-time crossings as non-fatal drift while still failing growth introduced by the current change.

Bug Fixes:
- Fix file-size ratchet so PRs whose trees are identical to the base no longer fail when a file first crossed the limit on main.
- Ensure files without baseline entries use the same base-aware limit calculation as grandfathered files, preventing incorrect NEWOVER failures.

Enhancements:
- Introduce a shared effective_limit helper to centralize base-aware limit calculation for both candidate types in the file-size check.
- Refine drift reporting to distinguish between grandfathered ceiling drift (regenerate baseline) and first-time crossings (split the file), and update the OK summary line to reflect inherited drift accurately.

Documentation:
- Update AGENTS.md to describe the ratchet rule in terms of each file’s own limit (baseline or global limit) and clarify how inherited drift for first-time crossings is reported and handled.

Tests:
- Extend the file-size guard test suite with cases for non-grandfathered files (B7–B9), covering inherited first-time crossings, new crossings introduced by the branch, and growth on top of inherited drift, plus stricter output substring checks in the test harness.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -424,17 +424,20 @@ inputs, not review afterthoughts:
   an escape hatch for a genuinely irreducible line (a module declaration in
   an already-oversized `lib.rs`), never something a plan may assume.
 
-**The ratchet judges your change, not the tree** (#2004). A file over its
-ceiling fails only when `current > max(ceiling, size in the base tree)` — so a
-violation that was already there before your branch existed is reported as
-drift and does not fail you, while anything your change grows past what it
-inherited fails exactly as it always did. The base is the merge commit's first
-parent on a pull request, the merge base locally, and nothing at all in a
-shallow clone or a fresh repository — where the guard falls back to the strict
-whole-tree check, because an unresolvable base must make it stricter, never
-weaker. This exists because the baseline is one shared cell every growing PR
-must write, and three times running two PRs that each wrote it correctly
-composed into a red `main` that then failed everyone downstream. Two
+**The ratchet judges your change, not the tree** (#2004). A file over the line
+fails only when `current > max(its own limit, size in the base tree)` — where
+that limit is the baseline entry for a grandfathered file and the 1500 lines
+above for every other (#2397). So a violation that was already there before
+your branch existed is reported as drift and does not fail you, while anything
+your change grows past what it inherited fails exactly as it always did.
+Inherited drift is survived, never absorbed: a first-time crossing you walked
+past still wants splitting, and still gets no baseline entry. The base is the
+merge commit's first parent on a pull request, the merge base locally, and
+nothing at all in a shallow clone or a fresh repository — where the guard falls
+back to the strict whole-tree check, because an unresolvable base must make it
+stricter, never weaker. This exists because the baseline is one shared cell
+every growing PR must write, and three times running two PRs that each wrote it
+correctly composed into a red `main` that then failed everyone downstream. Two
 consequences for planning: **do not "fix" a red ratchet you did not cause** by
 folding a baseline regeneration into an unrelated PR — land it on its own from
 a fresh `main`; and note that splitting a god file buys structure, not slack,
--- a/scripts/check-file-size.sh
+++ b/scripts/check-file-size.sh
@@ -82,6 +82,17 @@
 # ceiling raised deliberately via `--update` still passes exactly as before:
 # the regenerated ceiling equals the current size.
 #
+# `ceiling` here means the file's own limit, whichever kind it has: the baseline
+# entry for a grandfathered file, and LIMIT for every other. The rule was first
+# written for grandfathered files alone, and a file with no baseline entry went
+# on failing on sight — so the very first crossing of the 1500-line limit on
+# `main` turned every open PR red, byte-identical trees included, which is the
+# exact failure #2004 exists to prevent (#2397). What differs between the two
+# kinds is only the remedy the report names, not the arithmetic: a drifted
+# ceiling is cleared by regenerating the baseline, while a first-time crossing
+# is cleared by splitting the file. A first-time crossing is never grandfathered
+# on the way past — inherited drift is reported and survived, not absorbed.
+#
 # The base is the same pair `scripts/check-deleted-tests.sh` uses, for the same
 # reason: on a `pull_request` event the checkout is `refs/pull/N/merge`, so HEAD
 # is the merge commit and HEAD^1 is the base branch tip. That asks "does this
@@ -222,10 +233,10 @@ size_at_base() {
 # Single awk pass: read the baseline into a map, then judge each current file.
 # Baseline lines are tagged B, current sizes C, so one awk sees both streams.
 #
-# A file over its ceiling is emitted as a CANDIDATE, not a verdict: awk sees one
+# A file over its limit is emitted as a CANDIDATE, not a verdict: awk sees one
 # tree and so cannot tell growth from inherited drift. The classification needs
-# the base tree and happens below. Candidates are emitted in GREW's position so
-# the assembled report keeps its original section order.
+# the base tree and happens below. Candidates are emitted in their section's
+# position so the assembled report keeps its original section order.
 raw_report="$(
   {
     grep -v '^#' "$baseline" | sed 's/^/B /'
@@ -240,14 +251,14 @@ raw_report="$(
         else if (n > ceiling[path])
           grew = grew sprintf("GREWCAND %s %d %d\n", path, n, ceiling[path])
       } else if (n > limit) {
-        newover = newover sprintf("  %s is %d lines, over the %d-line limit\n", path, n, limit)
+        newover = newover sprintf("NEWCAND %s %d\n", path, n)
       }
     }
     END {
       for (p in ceiling)
         if (!(p in seen))
           stale = stale sprintf("  %s (baseline entry, file no longer tracked)\n", p)
-      if (newover) printf "NEWOVER\n%s", newover
+      if (newover) printf "%s", newover
       if (grew) printf "%s", grew
       if (obsolete) printf "OBSOLETE\n%s", obsolete
       if (stale) printf "STALE\n%s", stale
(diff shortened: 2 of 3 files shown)


macanderson:
On the review guide's one ❌ — splitting `arenabench/arenabench/telemetry.py` is
not done here, and that is deliberate rather than an omission.

#2397 lists two **separable** fixes and its stated definition of done covers
only the second:

> - `./scripts/test-file-size.sh` covers the three cases above.
> - A branch that does not touch `arenabench/` passes `make file-size` against a
>   `main` whose `telemetry.py` is over the limit.
> - `main` is green on the `file size ratchet` check.

All three hold with this change, and `file size ratchet` is green on this PR —
against a base whose `telemetry.py` is still 1534 lines, which is the second
bullet demonstrated rather than asserted. So `Closes #2397` is scoped to the
issue's own contract, and the split is carried forward on its own in #2403
(which I have reworded so the two do not read as contradicting each other).

Doing both here is the move AGENTS.md § "God files" explicitly forbids:

> do not "fix" a red ratchet you did not cause by folding a baseline
> regeneration into an unrelated PR — land it on its own from a fresh `main`

The split is arenabench's change, touches a Python module this PR has no other
business in, and would put a refactor of unknown blast radius in the same diff
as a gate fix that every open PR is currently waiting on. Happy to reverse the
call and downgrade this to `Refs #2397` if a maintainer would rather #2397 stay
open until the file itself is under the limit — it is a labelling question, not
a behavioral one.

PR #2408, opened 2026-08-09
fix(arenabench): unbreak main — split the transcript reader out of telemetry.py

macanderson:
My own #2391 did this. `telemetry.py` went 1474 -> 1534 lines and crossed the
1500-line gate, so `file size ratchet` now fails on `main` and therefore on
every PR in flight. It merged anyway, which is its own finding: the guard is
advisory at the merge button and only the author's push is watched.

AGENTS.md leaves no choice about the fix, and it is the right one:

  "A new file that crosses 1500 lines fails the gate outright, and the
   baseline accepts no new entries — a file approaching the limit gets split,
   not grown over it."

So: split, no baseline entry.

The seam is the file's own `# Transcripts` banner, not a line count. Reducing
trials to scoreboard numbers and replaying a trial's narration are different
jobs read by different consumers, and only two names cross between them —
which is why `_proof_line` moves too, being called from exactly one place,
inside the moved block.

  telemetry.py  1534 -> 1129
  transcript.py         429  (new)

`TranscriptReader` and `TranscriptState` are re-exported from `telemetry`
rather than repointed at their call sites: they are part of that module's
published surface (`server.py` and the test suite both import them from
there), and moving a file should not be an API break for readers who never
cared where it lived. Both import paths resolve to the same object, asserted
directly.

Pure code motion — no behaviour touched. The full arenabench suite passes
unchanged, `scripts/check-file-size.sh` reports OK, and ruff is clean on both
files (`field` was left behind in telemetry's imports and `distill` was never
used by the moved code; both removed by hand rather than by `--fix`, after
checking each was genuinely unreferenced).

## Summary by Sourcery

Split the transcript reading logic out of the telemetry module into its own transcript module while preserving the existing telemetry API surface.

Enhancements:
- Introduce a dedicated arenabench.transcript module to encapsulate transcript reading state and logic.
- Re-export TranscriptReader and TranscriptState from telemetry to avoid breaking existing import paths.

Chores:
- Reduce telemetry.py file size below the 1500-line gate by moving transcript-related code into a separate module.
- Clean up an unused dataclasses import in telemetry after relocating transcript code.


macanderson:
Superseded — `main` already carries this split.

`arenabench/arenabench/transcript.py` landed via #2394 (filed as #2397), with the same seam this PR chose: the metrics half and the transcript half share no module-level name, so the cut is along the question each answers rather than at a line count. `telemetry.py` is 1121 lines on main now, comfortably under the ratchet.

Closing rather than resolving the conflict: two implementations of one split is worse than either, and theirs is already merged and proven against the suite. My apologies for the duplicate — the file crossed the limit in my #2391 (1474 -> 1534) and I went to fix it without first checking whether main already had.

Nothing here is lost; the outcome I wanted is on main.

Issue #2407, opened 2026-08-09
ingest: extraction reads only the first 24,000 characters, so AGENTS.md is ingested at 54% and reports success


macanderson:
## Problem

`stella ingest <doc>` extracts records from at most the first 24,000 characters
of a document. Everything past that is dropped before the model ever sees it:

```rust
// crates/stella-cli/src/ingest_cmd/extract.rs
const MAX_PROMPT_CHARS: usize = 24_000;
```

This repository's own `AGENTS.md` is **44,545 characters**, so a
`stella ingest AGENTS.md` extracts **54% of it** and skips 20,545 characters.
The extractor's last line is the middle of the workspace-layout table:

```
| Triage → … → verifier orchestrat
```

Everything below that point — the god-file rules, the glossary, the code-style
conventions, the testing approach, and the whole gotchas section — cannot
produce a record, and could not have. The run still reports success: on the run
that motivated this issue it wrote 93 proposals to `.stella/proposals/agents.toml`
and said nothing about the missing half.

The steering documents this feature exists to ingest are exactly the documents
that exceed the cap. `AGENTS.md` and `CLAUDE.md` are named directly by the
first-run dialog (`ingest_cmd.rs::run_scan`, tier `Primary`), so the headline
use case is the one that silently loses content.

## What already landed

PR (branch `fix/ingest-silent-during-model-call`) makes the loss **visible** —
`bounded_content` returns the dropped character count and `run_named` prints:

```
    heads up: only the first 24000 of 44545 characters are extracted — 20545 (46%) are skipped, so nothing below that point can become a record
```

That is a warning, not a fix. A person who ingests AGENTS.md still cannot get
records out of the second half without splitting the file by hand.

## Files

- `crates/stella-cli/src/ingest_cmd/extract.rs` — `MAX_PROMPT_CHARS`,
  `bounded_content`, `call_model`, `extract_document`
- `crates/stella-cli/src/ingest_cmd.rs` — `run_named`, `truncation_notice`

## Proposed shape

Chunk the document and merge the claims, rather than raising the cap. Raising it
trades one silent failure for another: `call_model` already starts at a 16,384
output-token budget and a 9,401-token input filled it exactly
(`"finish_reason":"length"`, execution 89 in `.stella/private/store.db`), so a
larger input makes the cut-off retry the normal path instead of the exception.

Constraints a design has to respect:

- **Split on structure, not on a character offset.** Cutting mid-table produced
[...]

Issue #2406, opened 2026-08-09
ingest: infer_domains has the identical unfixed truncation bug PR #1096 fixed in extract.rs (2048-token cap, no finish_reason check)


macanderson:
## Problem

`stella-cli/src/domains.rs::infer_domains` carries the exact defect PR #1096
fixed in its sibling, and has carried it since: a hardcoded output cap with no
cut-off detection.

```rust
// crates/stella-cli/src/domains.rs:195
max_output_tokens: Some(2048),
```

There is no `finish_reason` check anywhere in the file (`rg -n "finish_reason"
crates/stella-cli/src/domains.rs` returns only a struct-literal `None` in a test
fixture at line 338). When the reply fills the budget, the JSON is cut off
mid-object and the parse failure is reported as a generic serde error rather
than as "the budget was too small" — and the bounded repair re-asks the same
question at `temperature: 0.0`, which tends to truncate at the same point.

`extract.rs`'s own module doc says it "mirrors `infer_domains`", so this is the
same bug in the function the fixed one was copied from. PR #1096 named it as a
fast-follow in its description and it was never filed, so it has been invisible
since 2026-08-01.

It is **more** likely to bite than the one that was fixed: 2048 is one eighth of
the 16384 `extract.rs` now uses, and reasoning models spend part of that budget
before emitting a token of JSON (a real `ingest_extraction` call on Opus 5 spent
1,093 reasoning tokens — over half of this entire budget — see execution 89 in
this repo's `.stella/private/store.db`).

## Files

- `crates/stella-cli/src/domains.rs` — `infer_domains`, the cap at line 195
- `crates/stella-cli/src/ingest_cmd/extract.rs` — `call_model`, the fixed shape
  to copy: `BASE_OUTPUT_TOKENS`, the `FinishReason::Length` check, and the
  budget-doubling retry arm

## Verify

`stella init` in a workspace with many source directories drives
`infer_domains`. A witness test does not need a provider: the mock-provider
pattern already used in `crates/stella-cli/src/accounted_call.rs`'s tests can
return a `CompletionResult` with `finish_reason: Some(FinishReason::Length)` and
a truncated body, and assert the retry widens the budget rather than re-sending
the same request.

## Definition of done

- `infer_domains` reads `finish_reason` and doubles its budget on
  `FinishReason::Length`, exactly as `extract.rs::call_model` does.
- The base cap is raised from 2048 to a figure that survives a reasoning model;
[...]

Issue #1467, opened 2026-08-05
OpenRouter: final usage frame lost to a 10s ceiling — the call's tokens and cost go unaccounted (agent.step.usage_incomplete)


macanderson:
## Problem

A pipeline triage call (glm-5.2 via OpenRouter) completed its work but its final usage accounting never landed:

```
WARN agent.step.usage_incomplete stella::diag_bridge seq=3 role=triage provider=openrouter model=z-ai/glm-5.2 reason=timeout duration_ms=10002 retries=null recovered=false
⚠ one model call reported no final usage — that call's tokens and cost are unaccounted (the work itself is unaffected)
```

A 10-second ceiling fired (`duration_ms=10002`) and there was no recovery (`recovered=false`). Spend and telemetry silently undercount. On a bench run this skews cost comparisons — the store's usage-completeness tests exist precisely because unaccounted calls corrupt totals.

Observed: 2026-08-04 session, `stella run "monitor pr …/pull/1426 …"`, seq=3, role=triage.

## Where to look

- The usage-completeness plumbing: `crates/stella-store/src/telemetry.rs`, `crates/stella-store/src/tests/usage_completeness.rs`, event shape in `crates/stella-protocol/src/event.rs`, pipeline surface in `crates/stella-pipeline/src/pipeline/tests/usage.rs`.
- The 10s figure suggests a fixed deadline on receiving the final/usage frame. Check whether it should ride the `model_timeout` posture key added for #1211 via PR #1277 (per-provider/model posture) instead of a constant — OpenRouter's usage frame for some upstream routes arrives late, and OpenRouter exposes a post-hoc generation-stats endpoint that could serve as a recovery/backfill path.

## Constraints

- `crates/stella-model/src/openai.rs` (the adapter OpenRouter rides) is a god file closed to growth — any recovery logic lands in a sibling module.
- Typed errors, no panics (invariant #5).
- A recovery fetch is provider-parity territory (invariant #8): if added, declare the posture and name the witness test in `provider_parity.rs` in the same PR.

## Definition of done

Either the usage deadline is posture-keyed and sized so glm-5.2-via-OpenRouter usage lands, or a recovery path backfills usage after the call. Witness: a wiremock test with a delayed usage frame that on `main` yields `usage_incomplete` and with the fix yields fully accounted usage.

PR #2409, opened 2026-08-09
fix(stella-cli): stella ingest is silent for minutes, and reads half of AGENTS.md

macanderson:
## What & why

`stella ingest AGENTS.md` printed the filename and then nothing at all until the
run finished. Reported as a hang; it was not one. Read from the local store
rather than from the surface:

| | execution | tokens | duration | finish | cost |
|---|---|---|---|---|---|
| `ingest AGENTS.md` try 1 | 89 | 9,401 in / 16,384 out (1,093 reasoning) | **135.2s** | `length` — cut off | $0.4684 |
| `ingest AGENTS.md` try 2 | 91 | — | ~182s | ok | $0.5264 |
| `ingest CLAUDE.md` | 90 | 2,526 in / 6,765 out | **58.2s** | ok | $0.1849 |

Both runs **succeeded** — 93 proposals in `.stella/proposals/agents.toml`, 36 in
`claude.toml`. AGENTS.md simply took **5m18s and $0.99** with zero output, and
read only 54% of the document while reporting success.

Two commits: make the silence visible, then fix what it made visible.

### 1. Say what is happening (`a32ec82a`)

- **The call is in flight.** A one-line elapsed ticker (`ingest_cmd/progress.rs`)
  rewrites itself in place once a second on a terminal, prints once on a pipe.
  Reuses `init_fx::animation_enabled` so `NO_COLOR`, `STELLA_NO_ANIM` and a
  non-TTY stdout stay decided in one place. `finish` runs on the statement after
  the `await`, before the result is inspected, so no error path can leave a
  ticker drawing over later output.
- **The document is being truncated.** The prompt already carried a truncation
  marker for the *model*; the person who typed the command was told nothing.
- **A cut-off reply silently re-runs the whole extraction** at double the budget.

### 2. Raise the cap 5x and size the reply to the document (`28d352bf`)

`MAX_PROMPT_CHARS` 24,000 → **120,000**. AGENTS.md is 44,545 characters, so it
now extracts **whole** instead of stopping mid-table at
`| Triage → … → verifier orchestrat` — which had made the god-file rules, the
glossary, the code-style conventions, the testing approach and the entire
gotchas section unable to produce a record.

Raising the input alone would have made things **worse**: 24,000 characters
already filled a 16,384-token reply and came back `"finish_reason":"length"`, so
5x the input against a fixed reply budget is a guaranteed cut-off, two
full-price attempts and a failure. So:

- **The first request is sized to the text it sends.** `starting_output_budget`
  allows one output token per input character — measured, not guessed (24,000
  chars filled 16,384 and finished inside 32,768; every claim restates its
  sentence and carries a dozen schema fields). Floor 16,384 so a short document
  is unchanged, ceiling 131,072. A large document no longer buys a cut-off
  attempt to earn a big enough one — the wasted AGENTS.md attempt cost $0.47 and
  135s for nothing.
- **The doubling ladder is clamped and terminates.** Re-buying an identical call
  that already hit the ceiling spends real money to fail the same way.
- **The scan's read cap now clears the prompt cap.** `MAX_READ_BYTES` was 64 KiB
  against a 120,000-char prompt cap, and `read_and_classify` drops an oversized
  file from the listing *entirely* — a large `AGENTS.md` would have been
  ingestable by name and invisible to `stella ingest` with no arguments. Raised
  to 512 KiB (clears 120,000 four-byte codepoints), with a test pinning the
  ordering.

The truncation notice still exists, because a bigger limit that lies about its
own edge is the original defect again — it now fires at 120,000.

**Honest limit.** At the top of the new input range the reply can exceed the
output ceiling of models whose catalog limit is under 128k (the seeded range is
30,000–384,000). That case now names itself rather than looking like a parse
error. Chunking (#2407) remains the real fix.

Refs #2374, #2407

## The witness

**18 tests.** The first twelve fail on `main` at the signature seam (verified
with `git show origin/main:…`: there `bounded_content` returns `String`, and
`skipped_chars`, `attempt_label`, `truncation_notice`, `percent` and the whole
`progress` module do not exist). Six more cover the cap change:

- a real AGENTS.md (44,545 chars) **now fits whole** — the direct witness for the 5x
- the notice still fires past the raised cap, and one character over it
- the budget is sized to the document, and never exceeds the ceiling
- `MAX_PROMPT_CHARS <= MAX_OUTPUT_TOKENS`, so a full-size document is expressible
  in one reply by construction
- the scan cannot hide a document extraction would accept
- the cap counts **characters not bytes** (a multi-byte fixture must not be cut
  early or mid-codepoint)
- the elapsed field **does not change width** across a minute, or the line
  visibly jitters on every tick that crosses a ten

`cargo test -p stella-cli --bins ingest` → **38 passed, 0 failed**.

**Verified end-to-end in the real binary**, against a dead endpoint so it costs
nothing:

```
$ stella ingest AGENTS.md --base-url http://127.0.0.1:9/v1
  AGENTS.md  688 lines
                                    ← no notice: it now fits whole
  extracting AGENTS.md …

$ stella ingest big.md   # 162,006 chars
  big.md  2 lines
    heads up: only the first 120000 of 162006 characters are extracted — 42006 (26%) are skipped, so nothing below that point can become a record
  extracting big.md …
```

Under a pty the ticker was confirmed to redraw in place (`\r` + `\x1b[2K`,
`0s`→`9s` on one row) and to erase cleanly before the next line prints.

## The gate

- [x] `cargo fmt --all` — clean
- [x] `cargo clippy -p stella-cli --all-targets -- -D warnings` → exit 0
- [x] `cargo test -p stella-cli --bins ingest` → 38 passed, 0 failed
- [ ] `cargo test --workspace` — not run locally; scoped to the one crate this
      touches. CI is the authority.
- [x] No tests deleted or renamed.
- [x] Docs: output-only behavior change; no flags added, no `--help` change.

## Nothing left behind

- [x] Filed: **#2406**, **#2407**

- **#2406** — `domains.rs::infer_domains` carries the *identical* unfixed bug
  PR #1096 fixed in `extract.rs`: `max_output_tokens: Some(2048)` with no
  `finish_reason` check. Named as a fast-follow in #1096 and never filed. More
  likely to bite than the one that was fixed — a real Opus 5 call spent 1,093
  reasoning tokens, over half that budget, before emitting any JSON.
- **#2407** — chunk the document instead of capping it. This PR raises the cap
  and reports the edge; it does not remove the edge.

## Ground-rule check

- [x] No I/O added to `stella-core` — every change is in `stella-cli`.
- [x] No new dependencies.

**Why a spawned ticker is safe.** Extraction runs on a `new_current_thread`
runtime driven by `block_on`; the scheduler polls the spawned ticker on that same
thread while the caller awaits the provider, so the line moves through a
multi-minute call with no second thread.

**Why no `Drop` backstop.** `finish` runs after the `await` and before the result
is inspected, so `call_model`'s several error paths cannot leak a ticker. One
call site, ordering visible at the seam — that beat defending it after the fact.

**Exemplar.** The progress module follows this repo's own
`init_fx::InitCinematic` (spawned render task, shared TTY predicate) but draws a
single row with `\r` + clear-to-EOL rather than a multi-row region with
cursor-up: one line cannot desynchronise from a row count.

**Not a defect.** The 120s `timeout` did not fire on a 135s call because
`AccountedCall::timeout` is an *idle* deadline that re-arms on every streamed
fragment (#1467) — a provider actively answering is never abandoned. Noted only
because the number looks wrong until you read that contract.


macanderson:
**The red `file size ratchet` is inherited from `main`, not caused by this PR.**

It fails on `arenabench/arenabench/telemetry.py is 1534 lines`, a file this PR
does not touch. It is **1534 lines on `origin/main` too** — byte-identical:

```
$ git show origin/main:arenabench/arenabench/telemetry.py | wc -l
    1534
```

`main`'s own last three commits are red for the same reason (8b63b88a, cd128459,
9e999b75 all `failure`), so every open PR inherits it. This is the exact shape
already filed as **#2397** — a first-time crossing on `main` fails every open PR,
because the base-relative check from #2004 only covers grandfathered files, and
CI's shallow checkout makes the base unresolvable so the guard falls back to the
strict whole-tree check. The underlying file is **#2403**; **#2404** covers the
`make file-size-update` hazard.

Deliberately not fixed here, per AGENTS.md § "God files": *"do not 'fix' a red
ratchet you did not cause by folding a baseline regeneration into an unrelated
PR — land it on its own from a fresh `main`."*

The four files this PR does change are far below the limit and none is
grandfathered:

| File | Before | After |
|---|---|---|
| `crates/stella-cli/src/ingest_cmd.rs` | 314 | 405 |
| `crates/stella-cli/src/ingest_cmd/extract.rs` | 778 | 827 |
| `crates/stella-cli/src/ingest_cmd/extract/tests.rs` | 487 | 537 |
| `crates/stella-cli/src/ingest_cmd/progress.rs` | — | 171 (new) |

Issue #2424, opened 2026-08-09
tools: verify_done still re-runs the WORKING-TREE half on an identical call in one turn — #2208 memoised only the shadow


macanderson:
## Problem

#2208 removed the duplicated **shadow** build: two identical `verify_done` calls in
one turn now build one detached worktree. The tool's *other* half — running
`test_cmd` against the working tree ("half 1", `crates/stella-tools/src/verify.rs`,
the `run(test_cmd, root, timeout_secs)` just before the memo lookup) — still runs on
every call, including the identical repeat.

That was a deliberate omission, not an oversight: half 1's determinant is the whole
working tree, and the memo key #2208 landed covers only what the shadow reads
(baseline commit, copied test files, command, timeout). Memoising half 1 on that key
would be unsound — a worker that edits the implementation and re-runs `verify_done`
must get a fresh answer, and that is exactly the loop the tool exists for.

## What is actually unknown

**How much of the 250s on `extract-elf` was half 1.** The issue reported the tool's
total wall clock; the split between the two halves was never measured. Do that first,
from the trace and not from a surface signal (CLAUDE.md § bench conclusions):

- open the run's `stella-events.jsonl`, join `tool_start` → `tool_result` on the
  `verify_done` calls in match `cc00894779ff`, and get per-call wall clock;
- with #2208 landed, the second identical call's remaining time *is* half 1.

If half 1 is a small fraction of the call, close this — the cheap win is already
banked and a sound memo here is expensive to build. If it is comparable to the shadow
(likely wherever `test_cmd` is a cold build in the real tree too), it is worth doing.

## If it is worth doing

The registry already takes the fingerprint that would make the key complete:
`crate::shell_touch::WorkspaceProbe::capture_with` walks the tree and hashes it, and
`ToolRegistry::begin_workspace_probe` / `settle_workspace_probe`
(`crates/stella-tools/src/registry/turn_probe.rs`) hold one per turn. A half-1 memo
would key on the probe's fingerprint plus the command and timeout, and must:

- fail open on a **saturated** probe (`WorkspaceProbe::saturated`) — a walk that hit
  a bound did not observe the whole tree, so it cannot certify that nothing moved;
- fail open where the probe is ignore-filtered, since a gitignored build input that
  changed is invisible to it (`IgnorePolicy`, `SKIP_DIRS`);
- ride the same arm/clear discipline `crate::verify::memo::ShadowMemo` already has,
  so nothing outlives its turn.

[...]

Issue #2377, opened 2026-08-08
the interactive multi-turn loop still probes the workspace per tool call — agent.rs is at its ceiling, so it needs a relief valve


macanderson:
## Problem

#2337 fixed the raw **one-shot** loop: `run_raw_one_shot` now arms a workspace-probe bracket per turn, so `ToolRegistry::execute` stops taking its per-call branch (two full tree walks per opaque call). The **interactive multi-turn loop** was left on the old granularity.

`crates/stella-cli/src/agent.rs` has two callers of `run_turn`:

- `crates/stella-cli/src/agent/goal.rs` — `run_raw_one_shot`, **armed** by #2337.
- `crates/stella-cli/src/agent.rs:~1079` — the interactive multi-turn loop, **not armed**.

`run_turn` now settles unconditionally (`TurnBracket::Settle` at the `close_event_stream` call), which is a harmless no-op for the interactive path because `settle_workspace_probe` returns early with no armed before-image. So today the interactive REPL still pays two tree walks per shell call.

## Why it was not just fixed in #2337

`crates/stella-cli/src/agent.rs` sits at **exactly its 2270-line ceiling** in `scripts/file-size-baseline.txt`. #2337 was able to modify existing lines (two call sites, one re-export list) without adding any, but arming a bracket needs a *new* statement before the `run_turn` call — which the ratchet would reject. This wants a relief valve, not a one-liner.

## Definition of done

- The interactive loop arms a bracket per turn, so a shell-heavy interactive session pays one walk per turn rather than two per call.
- `agent.rs` does not grow past its ceiling. Options, cheapest first:
  - extract the turn-driving block into a sibling submodule under `crates/stella-cli/src/agent/` (the pattern `stella-core/src/driver/settlement.rs` set), which buys room and removes lines from the god file rather than adding them;
  - or fold the arm into a helper the loop already calls on each iteration, if one exists that is not itself in `agent.rs`.
  - **Do not** raise the baseline to make room — see AGENTS.md § "God files": a ceiling moves only as a reviewable, justified baseline diff, and doing it to dodge a split is exactly what it is not for.
- Re-arming must not double-walk. `settle_workspace_probe` re-arms from the after-image it just took, so turn N+1 already has a valid before-image; arming again unconditionally would pay a redundant walk per turn. Either check before arming, or arm only on the first turn.
[...]

PR #2428, opened 2026-08-09
perf(stella-tools): stop verify_done rebuilding the same shadow twice in one turn

### LINKED ISSUE #2208: tools: verify_done re-runs the same shadow build within one turn — 250s (24% of budget) on extract-elf, #2129's unlanded fourth fix
## Problem

`verify_done` is re-run with **identical arguments** inside a single turn and pays the
full shadow-worktree cost every time. On `extract-elf` in match `cc00894779ff` this
cost **250s — 24% of the trial's budget** — for observations the pipeline already held.

This was direction 4 of #2129's analysis. PR #2191 landed directions 1–3 and did not
touch this one, and #2129 is now closed, so nothing tracks it.

Confirmed absent on `origin/main`:

```console
$ rg -n 'memoiz|cache' crates/stella-tools/src/verify.rs
$                       # no hits
```

## Why it costs what it costs

`verify_done` builds a detached shadow git worktree at the baseline, copies the test
files in from the working tree, and runs the suite there. That is the expensive part,
and it is fully determined by `(baseline_ref, test_paths, test_command, file_contents)`.
When a worker calls it twice with the same arguments and an unchanged tree, the second
run cannot observe anything the first did not.

The pipeline already has the shape of the answer elsewhere: the flip oracle tracks its
own command's observations rather than re-deriving them.

## Definition of done

1. A memo key over the inputs that actually determine the outcome — baseline ref, the
   resolved root-relative test paths, the test command, and a content digest of those
   files. **Not** the raw model-supplied argument string: `verify.rs` deliberately
   resolves paths from the canonical root-relative form, and the memo must key on the
   resolved values or it will collide across two different files.
2. Scope the memo to one turn. A tree can change between turns, and a stale hit here
   would manufacture a false `WITNESS CONFIRMED` — the exact failure class #2067 fixed.
   Any file mutation to a memoised path must invalidate the entry.
3. A witness test: two identical `verify_done` calls in one turn perform one shadow
   build, and a call whose test file changed in between performs a second.

## Constraints

- **A wrong memo is worse than no memo.** This tool is the flip oracle's evidence
  source; a false hit is a false proof. Fail open (re-run) on any doubt.
- Invariant 5: no `unwrap`/`expect` on runtime data — the digest path reads files that
  may vanish mid-turn.

## Files

- `crates/stella-tools/src/verify.rs` — the tool and its shadow-worktree construction
- `crates/stella-pipeline/src/pipeline/evidence.rs:85` — where `verify_done_flip` is read

## Related

- #2129 (closed) — this was its unlanded fourth direction
- #2149, #2204 — the other two "the witness stage eats the trial" issues

macanderson:
## What & why

`verify_done` runs two halves: the witness test against the working tree, and the
same test against a detached shadow worktree at the witness baseline. The second is
a fresh checkout with a cold build, and it is where the wall clock goes — **250s,
24% of the trial's budget**, on `extract-elf` in match `cc00894779ff`, spent
re-deriving an observation the run already held. This was direction 4 of #2129's
analysis, which #2191 did not land.

The shadow half is memoisable because its inputs are **closed**: the only things
that reach a fresh checkout are the baseline commit, the test files copied into it,
the command, and the timeout. So the key is the *complete determinant* of the run
rather than a summary of the call — which is what makes a hit sound instead of
merely likely. Two consequences worth stating plainly:

- A worker that edits the **implementation** between two calls still gets a fresh
  shadow verdict for free, because the shadow never saw the implementation and its
  verdict cannot have moved.
- The half that **does** see that edit — the working-tree run — is never memoised,
  so "does my change work now" is re-measured on every call.

The key is built from the **resolved** values (canonical toplevel, root-relative
shadow cwd, baseline SHA, command, timeout, and a sha256 per toplevel-relative
destination), never the raw model-supplied argument string, per #2208's DoD.
Destinations are sorted and deduplicated, so argument order cannot split one run
into two. Fields are length-prefixed rather than separator-joined: a `test_cmd` is
an arbitrary shell string, so no byte can be reserved as a boundary.

Two guards, because a false hit here is a false proof — #2067's failure class:

1. the key covers every input the shadow reads (a changed test file, a moved
   baseline, a different command all miss);
2. entries live for exactly **one turn**, which bounds what a key cannot cover — a
   flaky suite, a clock, a network resource, a host that changed the environment
   between turns.

The turn boundary is the registry's workspace-probe bracket, the only one a
`ToolRegistry` is told about: `begin_workspace_probe` arms and
`settle_workspace_probe` clears. A host that never brackets gets an inert memo and
exactly today's behaviour. **Anything short of certainty fails open** (run the
shadow): an unreadable test file — it may vanish mid-turn — a non-UTF-8 path, a
turn past the entry cap, or an unbracketed session. Every verdict now says when its
previous-code observation was reused rather than re-measured; `WITNESS CONFIRMED`,
`VACUOUS TEST` and `WITNESS_INCONCLUSIVE_BUILD_ERROR` keep their exact markers.

Exemplar for the shape: `std::collections`-backed, handle-shared session state
built the way this crate's own `ReadLedger` and `CitationLedger` are — one owner
per registry, cloned into the tool that reads it, no globals.

Closes #2208

## The witness

Two, at two seams, both counting *actual shadow builds*: the witness script appends
a line to a marker file whenever it runs inside a linked worktree (where `.git` is
a file, not a directory), which is exactly the shadow.

- `crates/stella-tools/src/verify/memo.rs` —
  `an_identical_call_in_one_turn_builds_one_shadow`: two identical calls build one
  shadow, the second says `REUSED`, and a call whose test file changed in between
  builds a second. This is #2208's DoD item 3 verbatim.
- `crates/stella-tools/src/registry/tests/verify_memo.rs` —
  `the_turn_bracket_scopes_the_dispatched_verify_dones_memo`: the same at the seam
  that ships, through `ToolRegistry::execute("verify_done", …)`, so the memo cannot
  be correct and reach no user. A settle in between rebuilds.

Checked the artisanal way — with `ShadowMemo::get` stubbed to return `None`, both
fail (`left: 2, right: 1`) and the two fail-open controls
(`nothing_is_reused_across_a_turn_or_without_one`,
`an_unbracketed_session_reuses_nothing`) still pass, which is what makes them
controls rather than duplicates.

## The gate

- [x] `cargo fmt --check`
- [x] `cargo test --workspace` (128 test binaries, exit 0)
- [x] `make invariants`, `doc-links`, `command-docs`, `brand-case`, `file-size`,
      `god-files`, `gate-parity`, `left-behind`, `role-names`, `stat-portability`,
      `module-reachability`, `typed-errors`, `wire-schema`, `doc-warnings` — all OK
      (`shellcheck` is not installed in this environment; no shell scripts changed)
- [x] Docs updated: `verify.rs`'s module header gains a "Running the same proof
      twice" section; the memo carries its own module doc
- [x] `Closes #2208` in this description **and** as a commit trailer

**God files, respected rather than raised.** `registry.rs` is grandfathered at 2181
and sits at **2180** after this change — three lines of field, one local, one
argument, one initializer, and no baseline edit. The new logic lands in a sibling
submodule (`verify/memo.rs`) and the arm/clear calls in `registry/turn_probe.rs`,
per AGENTS.md § "God files — plan around them, never into them". `verify.rs` is not
grandfathered, so it stays under 1500: the memo's own tests live with the memo,
which is where they belong anyway, and the shadow half moves out of `execute` into
`run_against_baseline` unchanged.

## Nothing left behind

- [x] Filed: **#2424** — `verify_done` still re-runs the working-tree half on an
      identical call in one turn. Deliberately not fixed here: half 1's determinant
      is the whole working tree, and memoising it on this key would be unsound.
      The issue says to measure the two halves' split from `stella-events.jsonl`
      first (join `tool_start` → `tool_result`) before building anything, and names
      the fail-open cases a probe-fingerprint key would have to honour.
- [x] Commented on **#2377** — the interactive multi-turn loop never arms a
      bracket, so it now also gets no memo. Same fix, already tracked; noted there
      so the fix is not later narrowed to the probe alone.

## Ground-rule check

- [x] No I/O added to `stella-core`; no new dependencies at all (`sha2` arrives via
      the existing `crate::staleness::hex_sha256`)
- [x] No new cross-boundary types — `ShadowMemo` never crosses a crate boundary or
      serializes

**A mutation to a memoised path does invalidate its entry** — #2208's DoD item 2 —
but by construction rather than by a hook: the key carries a sha256 of every
resolved test file, so a mutated memoised path yields a different key and the old
entry is unreachable. That is strictly stronger than a mutation hook, which races
the write; a content digest cannot. The third assertion in
`an_identical_call_in_one_turn_builds_one_shadow` is exactly this case.

What is deliberately **not** invalidated is a mutation to a path that is not in the
key — an implementation file. It is not a memoised path: it is never copied into the
shadow and a fresh checkout of the baseline cannot see it, so the shadow's verdict
cannot have moved, and the half that *does* see the edit re-runs every call. Hooking
implementation edits would cost most of the benefit in the edit-then-reverify loop
for no correctness, and would not be sound alone anyway — in a bracketing session
`bash`-driven writes are attributed at settle, so mid-turn they move no counter.

**The one judgement worth arguing with** is memoising only half 2. #2208 frames turn
scoping as what prevents a false `WITNESS CONFIRMED`; on this design the *key* is
what prevents it, and turn scoping bounds nondeterminism the key structurally
cannot see. Both are implemented, so the disagreement is about which is load-bearing
— but if you think the shadow can observe something outside
`(baseline, test files, command, timeout)`, that is the thing to name, because it
would be a hole in the key. `run_against_baseline` carries a comment saying so: a
reader added there silently widens what the key has to cover.

`MAX_ENTRIES = 16` is a backstop against a pathological turn holding every test
output it produced, not a tuned number; past it the run simply happens.

### DIFF
--- a/crates/stella-tools/src/registry.rs
+++ b/crates/stella-tools/src/registry.rs
@@ -202,6 +202,9 @@ pub struct ToolRegistry {
     /// construction so no capture can race a reconfiguration. See
     /// [`RegistryOptions::probe_ignore_policy`].
     probe_ignore_policy: crate::shell_touch::IgnorePolicy,
+    /// The turn-scoped shadow-run memo shared with the registered
+    /// `verify_done`, armed and cleared by the turn bracket in `turn_probe`.
+    shadow_memo: crate::verify::ShadowMemoHandle,
     /// The session's memory-citation ledger, shared with the registered
     /// `cite_memory` tool instance and drained per execution by
     /// [`ToolRegistry::take_memory_citations`].
@@ -420,6 +423,7 @@ impl ToolRegistry {
         let spawn_queue: crate::tasks::SpawnQueue = Arc::default();
         let sub_agent_dispatcher: crate::subagent::DispatcherSlot = Arc::default();
         let sub_agent_spend: stella_core::subagent::SubAgentSpendLedger = Arc::default();
+        let shadow_memo: crate::verify::ShadowMemoHandle = Arc::default();
         // One read-state ledger per registry (#331): `read_file` and
         // `read_symbol` (which reads through the same instance) record what
         // the model saw; `edit_file` attributes match failures against it and
@@ -468,6 +472,7 @@ impl ToolRegistry {
             entries.extend(process_tools::builtins(
                 task_board.clone(),
                 spawn_queue.clone(),
+                shadow_memo.clone(),
             ));
             // The web family registers by default, like every other built-in.
             // A fetched page is untrusted input and an egress channel, which is
@@ -577,6 +582,7 @@ impl ToolRegistry {
             mutations: std::sync::atomic::AtomicU64::new(0),
             process_free,
             probe_ignore_policy: options.effective_probe_ignore_policy(process_free),
+            shadow_memo,
         }
     }
 
--- a/crates/stella-tools/src/registry/process_tools.rs
+++ b/crates/stella-tools/src/registry/process_tools.rs
@@ -8,6 +8,7 @@ use super::Tool;
 pub(super) fn builtins(
     task_board: crate::tasks::TaskBoardHandle,
     spawn_queue: crate::tasks::SpawnQueue,
+    shadow_memo: crate::verify::ShadowMemoHandle,
 ) -> Vec<Arc<dyn Tool>> {
     let processes: crate::process::ProcessTableHandle = Arc::default();
     let repo: Arc<dyn crate::repo::RepoBackend> = Arc::new(crate::repo::GitCli);
@@ -17,7 +18,7 @@ pub(super) fn builtins(
         // an operator who wants it gone writes `"tools": {"bash": "off"}`,
         // which the policy layer above enforces for MCP and custom tools too.
         Arc::new(crate::bash::Bash),
-        Arc::new(crate::verify::VerifyDone),
+        Arc::new(crate::verify::VerifyDone::with_memo(shadow_memo)),
         Arc::new(crate::project::BuildProject),
         Arc::new(crate::project::RunTests),
         Arc::new(crate::diagnostics::Diagnostics),
--- a/crates/stella-tools/src/registry/tests.rs
+++ b/crates/stella-tools/src/registry/tests.rs
@@ -858,6 +858,7 @@ mod partial_belief;
 mod private_state;
 mod schema_gate;
 mod touch;
+mod verify_memo;
 
 /// The no-clobber guarantee, end to end and across two independent sessions.
 ///
--- a/crates/stella-tools/src/registry/tests/verify_memo.rs
+++ b/crates/stella-tools/src/registry/tests/verify_memo.rs
@@ -1,0 +1,122 @@
+//! The wiring between the registry's turn bracket and `verify_done`'s
+//! shadow-run memo (#2208).
+//!
+//! `verify.rs` proves the memo's behaviour against a directly-constructed
+//! tool. This module proves the half that only the registry can: that the
+//! `verify_done` a real session dispatches shares the memo the turn bracket
+//! arms and clears. Without it the memo could be correct and reach no user —
+//! which is the failure mode the "nothing left behind" rule names.
+
+use super::*;
+
+/// A git repo with a committed "previous version", an uncommitted new one,
+/// and a witness script that appends a line to `shadow-runs` on every run
+/// that happens inside a linked worktree — i.e. once per shadow build.
+fn repo_with_counting_witness() -> tempfile::TempDir {
+    let dir = tempfile::tempdir().expect("tempdir");
+    let root = dir.path();
+    for args in [
+        &["init", "-q"][..],
+        &["config", "user.email", "t@t.t"],
+        &["config", "user.name", "t"],
+    ] {
+        git(root, args);
+    }
+    std::fs::write(root.join("impl.txt"), "old behavior\n").expect("write impl");
+    git(root, &["add", "."]);
+    git(root, &["commit", "-q", "-m", "previous version"]);
+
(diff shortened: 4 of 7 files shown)


macanderson:
> However, it explicitly chooses not to invalidate entries on file mutation … This
> contradicts the issue's definition of done item 2, which requires that any file
> mutation to a memoised path must invalidate the entry.

DoD item 2 is met — the two halves of that sentence are about different files, and
my "Anything reviewers should know" paragraph invited the conflation. Concretely:

**A mutation to a memoised path does invalidate the entry**, by construction rather
than by a hook. The key includes a sha256 of every resolved test file
(`ShadowKey::compute`, `crates/stella-tools/src/verify/memo.rs`), so a mutated
memoised path yields a different key and the old entry is unreachable — there is no
window in which a stale entry can be served, which a mutation *hook* would have
(hook races the write; a content digest cannot). This is exactly what the third
witness assertion in `an_identical_call_in_one_turn_builds_one_shadow` proves: after
`counting_witness(&root, &marker, "second")` rewrites the test file, the call
rebuilds and the verdict does **not** say `REUSED`.

**What is not invalidated is a mutation to a path that is not in the key** — an
implementation file. That is not a memoised path: it is never copied into the shadow
and a fresh checkout of the baseline cannot see it, so its verdict cannot have moved.
The half that *does* see that edit — the working-tree run — is not memoised and
re-runs on every call, so an implementation edit still gets a fresh answer.

Invalidating on implementation edits would cost most of the benefit in the
edit-then-reverify loop and buy no correctness. It would also not be sound on its
own: in a bracketing session `bash`-driven writes are attributed at settle, so
mid-turn they move no counter a hook could read.

I've reworded that paragraph in the PR description so the next reader does not have
to reconstruct this.

Issue #2458, opened 2026-08-09
pipeline: witness authoring is now decided in two places (witness_writer AND the roster), and the resume frame carries only one


macanderson:
## Problem

After #2381 there are **two independent knobs for one decision** — whether the
pipeline authors a witness test:

- `PipelineConfig::witness_writer` (the original bool, reachable from
  `--no-witness`-style wiring and carried in the resume frame)
- `roster.enabled(ModelCallRole::WitnessAuthor)` (the new per-responsibility
  ablation switch)

`Pipeline::run` currently ANDs them:

```rust
&& self.config.witness_writer
&& self.responsibility_enabled(ModelCallRole::WitnessAuthor)
```

That is deliberately additive — it cannot change behaviour for anyone who set
neither — but it is exactly the "one decision spelled in two places" shape
this repo treats as a defect waiting to happen. The failure mode is a reader
(or a future edit) consulting one and believing it is authoritative: a surface
that flips `witness_writer` on will silently do nothing if the roster row is
off, and vice versa, with no diagnostic naming the other.

`crates/stella-cli/src/resume_frame.rs` records `witness_writer` on the resume
frame and `crates/stella-cli/src/agent/resume.rs` restores it — the roster is
**not** carried there, so a resumed run currently reconstructs one half of the
decision and defaults the other. That is the concrete bug hiding in the
duplication, and it is the reason this is worth doing rather than tidy-up.

## Files

- `crates/stella-pipeline/src/pipeline.rs` — the `authored_witness` conjunction
  in `Pipeline::run` (search `witness_writer`)
- `crates/stella-pipeline/src/pipeline.rs` — `PipelineConfig::witness_writer`
  and its doc comment
- `crates/stella-pipeline/src/roster.rs` — the `WitnessAuthor` row
- `crates/stella-cli/src/resume_frame.rs` — `witness_writer` on the frame
  (~L73, L106, L212)
- `crates/stella-cli/src/agent/resume.rs` — `pipeline_config.witness_writer =
  frame_config.witness_writer` (~L254)

## Suggested resolution

Make the roster the single source of truth and reduce `witness_writer` to an
input that is folded into it once, at config-build time
(`apply_pipeline_tuning`), rather than consulted at run time. Then either:

- carry the whole roster on the resume frame (preferred — it makes every
  ablation survive a resume, not just this one), or
- if that is too large, carry the resolved `WitnessAuthor` enablement and say
  in the frame's docs that the rest of the roster is re-derived from settings.

[...]

Issue #2457, opened 2026-08-09
pipeline: context recall and scope review cannot be ablated — they issue no model call, so the roster has no row for them


macanderson:
## Problem

The responsibility roster (#2381) keys on `stella_protocol::ModelCallRole`,
which is the vocabulary of **paid calls**. That covers every stage whose
existence is a model call — triage, research, plan, witness authoring, the
worker, distress guidance, the verdict — and gives each one an `enabled`
ablation switch.

Two pipeline stages are therefore **not ablatable**, because they issue no
model call and so have no `ModelCallRole`:

- `StageKind::ContextRecall` — the context-plane recall that puts frames in
  front of the model before it plans. Governed only by `recall_latency_ceiling`
  and whichever `ContextRecallPort` is wired.
- `StageKind::ScopeReview` — the interactive approval gate. Governed by
  `scope_thresholds`, `plan_mode`, and `headless_bypass_scope_review`, none of
  which is "off".

This matters for #2374: **context record retrieval** and **adaptive context
lifecycle** are two of the fourteen features under evaluation, and the paired
ablation for them needs the same clean on/off the triage and verifier cards
now have. Approximating "recall off" by unwiring the port is a different
experiment — it changes what the binary is, not what the run does.

## Files

- `crates/stella-pipeline/src/roster.rs` — `default_agent`'s exhaustive match
  is the vocabulary boundary; see its module docs for why `ModelCallRole` was
  chosen over inventing a fourth enum
- `crates/stella-pipeline/src/pipeline.rs` — the recall future is inlined in
  `Pipeline::run` (search `StageKind::ContextRecall`); note the file is a
  grandfathered god file at its 3181 ceiling
- `crates/stella-pipeline/src/pipeline/scope_stage.rs` — the scope gate
- `crates/stella-pipeline/src/pipeline/roster_wiring.rs` — where a non-model
  responsibility's `enabled` check would be read

## The design question to settle first

`ModelCallRole` is a **wire type** in `stella-protocol` with strict
forward-intolerance: an unrecognised role token fails its whole event. Adding
`ContextRecall`/`ScopeReview` variants to it purely to give them roster rows
would put non-calls into the paid-call ledger's vocabulary, which is a real
cost — `step_usage` consumers would start seeing role tokens that can never
carry usage.

Two candidate shapes, and this issue is not prejudging them:

1. **A second, smaller key type** for non-model stages, with the roster holding
   two maps. Honest, but reintroduces exactly the "two vocabularies that look
[...]

Issue #2456, opened 2026-08-09
pipeline: the roster's agent vocabulary is still closed at four — a genuinely new agent is a code change


macanderson:
## Problem

The responsibility roster (#2381, PR TBD) made **which agent performs a
responsibility** configurable, but the set of agents that can be *named* is
still closed at four: `worker`, `triage`, `plan`, `verifier`
(`AgentId::BUILTIN` in `crates/stella-pipeline/src/roster.rs`).

So this works today:

```json
"responsibilities": { "witness_author": { "agent": "triage" } }
```

and this does not — it refuses the run with `RosterError::UnknownAgent`:

```json
"agents":           { "reviewer": { "model": "openrouter/openai/gpt-5.5" } },
"responsibilities": { "verdict": { "agent": "reviewer" } }
```

The refusal is deliberate and correct as shipped (a typo'd agent name must
never silently resolve to the worker grading its own work). But it means
adding a genuinely new agent — a second verifier, a dedicated reviewer, a
cheap researcher — is still a code change, which is the one thing the roster
was asked to make unnecessary.

## Why it was not done in #2381

`AgentId::role()` resolves a name to a `stella_protocol::Role`, and `Role` is
a closed enum the router resolves against. Opening the vocabulary needs a
resolution layer that does not go through `Role`, plus opening the settings
side: `AgentEngineAgents` (`crates/stella-cli/src/settings.rs`) is a fixed
struct with four named `Option` fields, and `EngineAgentKind` is a closed enum
with a hand-written `ALL` — **235 references across 16 files in
`crates/stella-cli/`** at time of writing. That is a settings/profile/deck
refactor orthogonal to the ablation controls #2374 was blocked on, so it was
split rather than bundled.

The seam was built for it: `AgentId` is a string newtype, not an enum, and all
resolution funnels through the single `AgentId::role()` function. This should
be an added implementation, not a restructuring.

## Files

- `crates/stella-pipeline/src/roster.rs` — `AgentId::BUILTIN`, `AgentId::role()`,
  `RosterError::UnknownAgent`
- `crates/stella-pipeline/src/pipeline/roster_wiring.rs` — `Pipeline::assigned`,
  the one place a responsibility becomes a resolved provider
- `crates/stella-pipeline/src/ports.rs` — where a new `AgentResolver` port would
  live, beside `ProviderResolver`
- `crates/stella-cli/src/settings.rs` — `AgentEngineAgents` (fixed struct),
[...]

PR #2462, opened 2026-08-09
feat(stella-pipeline): make responsibility→agent binding configurable, so a single stage can be ablated (closes #2381)

### LINKED ISSUE #2381: pipeline: no stage can be disabled independently — --no-pipeline is all-or-nothing, so triage and verifier cannot be ablated (#2374)
## Problem

There is no way to disable a single pipeline stage. `--no-pipeline` is the only off-switch, and it removes triage **and** plan **and** scope **and** execute-shaping **and** witness **and** verify together. That makes it structurally impossible to attribute a measured effect to any one stage, which blocks the paired ablations for the triage agent (F6) and the verifier agent (F7) in the prove-value plan (#2374).

The two settings keys that look like they might help do not:

- `agent_engine_config.pipeline_verifier_evidence_demand` and `agent_engine_config.pipeline_require_independent_verifier` tune the verifier's **strictness**; neither removes the stage.
- `agent_engine_config.pipeline_triage_model` selects the model; there is no `triage: off`.

Approximating "verifier off" by passing `--test-command` only bypasses **witness authoring**, not verification, so it is a different experiment than the one the plan needs.

## Files

- `crates/stella-pipeline/src/pipeline.rs` — stage wiring and the fast-path forks
- `crates/stella-pipeline/src/pipeline/triage_stage.rs`, `witness_stage.rs`, `verifier_stage.rs`
- `crates/stella-pipeline/src/lib.rs` — the stage contract and ordering
- `crates/stella-cli/src/settings.rs` and the `agent_engine_config` block — where the per-stage key would live
- `crates/stella-cli/src/cli.rs` — `--no-pipeline` is declared at roughly L490 (`Run`), L550 (`Arena`), L572 (`Goal`), L742 (`Fleet`); a per-stage flag should follow the same shape
- `crates/stella-cli/src/durability.rs::a_bare_turn_leaves_no_pipeline_frame` — the existing off-arm witness to copy

## What "done" looks like

A per-stage disable expressed in `agent_engine_config` (for example `agent_engine_config.triage.enabled = false`, `agent_engine_config.verifier.enabled = false`) such that:

1. Disabling one stage leaves every other stage running and the pipeline otherwise intact.
2. The skipped stage emits no `AgentEvent::Stage` frame, so `stella-events.jsonl` shows the ablation unambiguously rather than requiring the reader to infer it.
3. Disabling the verifier does **not** silently downgrade the verdict to a claimed-done — the run must report that verification was disabled, in keeping with the pipeline's degradation-warns-never-disables posture.
4. Defaults are unchanged: every stage stays enabled unless explicitly turned off.
5. ArenaBench can declare it the same way it declares `engine.bare_loop` (`arenabench/arenabench/config.py`), so a match TOML can run the ablation arm.

## Witness

A test that runs one prompt with triage disabled and asserts no `StageKind::Triage` frame appears in the event stream while `StageKind::Execute` and `StageKind::Verify` still do — failing on `main` today because the key does not exist.

## Constraints

- `crates/stella-pipeline/src/pipeline.rs` and `pipeline/tests.rs` are **grandfathered god files closed to growth** (see AGENTS.md § God files). New logic goes in a sibling submodule; do not add lines to either.
- Stage ordering is canonical in `stage_rank` (`crates/stella-pipeline/src/replay.rs`) — a skipped stage must not perturb the ranks of the others.
- `make gate` must stay green, including the replay fixtures under `crates/stella-pipeline/tests/fixtures/` and the golden manifests (re-record via `make record-golden` only if the change genuinely alters recorded output, and read the diff).

## Why now

This is test infrastructure for the #2374 feature freeze, not a new capability — it is explicitly named as the freeze exception in the plan's Phase 0 (blocker 0.6). Two of the fourteen features under evaluation cannot be graded at all until it exists.

Refs #2374

macanderson:
## What

Makes **who performs each pipeline responsibility** and **whether it runs at
all** configuration instead of a literal at every call site — the
`stella_pipeline::roster` module. #2381 asked for a per-stage off-switch so
triage and the verifier can be ablated independently; this delivers that as one
mechanism rather than a bespoke flag, so the same key also reassigns a
responsibility to a different agent.

```json
{
  "agent_engine_config": {
    "responsibilities": {
      "triage":         { "enabled": false },       // ablate a stage
      "witness_author": { "agent": "triage" },      // reassign it
      "verdict":        { "agent": "worker" }       // self-grade, and be told so
    }
  }
}
```

Reachable from `settings.json`, `stella.toml` (`[agents.responsibilities.*]`),
and an ArenaBench match TOML (`[contestant.engine.responsibilities.*]`).

## Why this shape

**`ModelCallRole` is the responsibility vocabulary — no fourth enum was added.**
Every paid call in the crate already names a `ModelCallRole` (what the call is
*for*) beside a `Role` (who *serves* it): `WitnessAuthor` beside
`Role::Verifier`, `Triage` beside `Role::Triage`. That pair *is* the binding
this PR makes configurable. Introducing a new "responsibility" enum beside
`Role`, `ModelCallRole` and `StageKind` would have added a fourth lookalike id
to the six AGENTS.md already keeps a glossary for. `ModelCallRole` is also
total by construction (`ALL` is macro-derived and compiler-checked) and is the
unit the paid-call ledger attributes spend to, so a roster row and a usage line
spell the same job the same way.

**Adding a responsibility is a compile error until someone owns it.**
`roster::default_agent` is an exhaustive match over `ModelCallRole`; a new
variant fails it with `E0004` until its default binding is declared. That is
the whole maintenance contract — modelled on `model_call_roles!` in
`stella-protocol`, which binds `ALL` to the enum the same way.

**Stage *order* is deliberately not configurable, and I want that reviewed as
a decision rather than an omission.** The ordering is not a workflow, it is a
proof protocol: the witness is authored in a pristine snapshot of the
pre-execution tree by a model that is not the worker, and its fail→pass flip is
credited only when the worker never touched the witness files. A config file
that could move authoring after the diff would still *report* a flip — it would
simply have stopped meaning anything. `replay::stage_transition_legal` encodes
the same ordering for recorded streams, and an operator-authored graph would
make replay validation undecidable rather than merely stricter. So: the set of
responsibilities and their order are code; assignment and enablement are
configuration.

**Repairs follow their principal rather than carrying a row.** `plan_repair`
and `witness_repair` re-attempt the same call, so they are issued to whichever
agent produced the output being repaired. Naming one in config is a named error
pointing at the row the operator meant (`RosterError::FollowsPrincipal`), not a
silently ignored key — every row in the roster steers something.

## Behaviour

- **Defaults are unchanged.** `Roster::default()` reproduces the bindings each
  call site hard-coded, pinned by
  `roster::tests::the_default_roster_is_the_pipeline_that_shipped` and by the
  existing 662-test pipeline suite passing untouched. An ArenaBench arm that
  declares nothing emits no `responsibilities` key, so **posture digests for
  existing arms do not move**.
- **An ablated stage emits no `StageKind` frame** — requirement 2. A reader of
  `stella-events.jsonl` sees the ablation instead of inferring it from a call
  that never happened.
- **A disabled verdict does not become a claimed-done** — requirement 3. The
  deterministic ladder still runs (it is not the verifier), and the turn routes
  to the *abstention* rung (`ProofStep::VerificationUnavailable`), not the
  degradation rung: nothing degraded, the operator removed the stage. The
  resulting `passed: true` means "nothing failed this", never "something proved
  it".
- **Self-grading is legal and never silent.** Binding `verdict` to the worker
  is reported as a lost independence. Binding `witness_author` to the worker
  disables authoring rather than producing a false proof — that filter is
  unconditional and not a roster question.
- **An unhonourable roster refuses before spend**
  (`PipelineError::InvalidRoster`), reporting every problem at once. The
  refusal lives in `Pipeline::run` rather than in the CLI, so every host —
  serve, fleet, deck — gets it; `Roster::validate` is total because key-level
  rejections are recorded on the roster itself.

### One semantic fix the witness test surfaced

Ablating triage initially also removed **verification**, because the fallback
was the deterministic floor and the floor's cheapest class (`SimpleLookup`)
skips the ladder. An operator ablating one stage would have silently ablated
two and attributed both effects to triage. `triage_ablated` now clamps at
`SingleTask`: the *fast path* is what disappears with triage, while the floor's
own upward evidence survives it. An unresolvable triage keeps the plain floor —
an ablation is not an outage, which is the theme running through the whole
change.

## Witness

`crates/stella-pipeline/src/pipeline/tests/roster_ablation.rs`, exactly as
#2381 specified: one prompt with triage disabled asserts no `StageKind::Triage`
frame while `Execute` and `Verify` still appear. It fails on `main` because
there is no key to disable triage with, so the frame is emitted on every run.
Three siblings cover the verdict ablation, the untouched default, and the
before-spend refusal. Plus 16 unit tests in `roster/tests.rs` and 10 in
`arenabench/tests/test_responsibilities.py`.

## Baseline change: none, after the main merge

**This PR no longer touches `scripts/file-size-baseline.txt`.** It originally
raised `pipeline/tests.rs` 2537 → 2538 for one `mod roster_ablation;` line,
because that file sat exactly at its ceiling. Merging `main` removed the need:
`main` split `pipeline/tests.rs` down to **2475** lines, so the `+1` is no
longer earned and the entry was reverted to `main`'s value in d638ff12. The
baseline diff against `main` is now empty — which is the outcome worth having,
since that file is the one shared cell behind the parallel-merge-skew incidents
(#1761, #1782, #2003). `pipeline.rs` itself **shrank** 3165 → 3161.

## The main merge, and the repair it needed

`main` was merged in 4073d9ee. Git reported no conflict on three stage
signatures and was wrong to: `main` renamed `PipelineBudgetAbort` →
`PipelineStageAbort` while this branch independently edited the same
signatures, so the merge kept this side's text and left three references to a
type that no longer exists. Each side was internally consistent, which is
exactly why the rename/edit collision merged silently. It broke
`cargo check --workspace` and the release smoke build; the `E0277` "size cannot
be known" errors in `pipeline.rs` were cascade from the unresolved error type,
not separate faults.

d638ff12 repairs it — three signatures in `triage_stage.rs` and
`verifier_stage.rs`, plus the baseline revert above.

## Exemplars

`ModelCallRole::ALL`'s `model_call_roles!` macro (this workspace) for the
compiler-enforced totality argument; `rustfmt`/`cargo`'s pattern of a pure
decision module over owned data with the I/O seam beside it, matching this
crate's existing `triage.rs` / `triage_stage.rs` split.

## Left behind (filed, not skipped)

- #2456 — the agent vocabulary is still closed at four names, so a *genuinely
  new* agent is still a code change. The seam is built for it (`AgentId` is a
  string newtype, resolution funnels through one function); opening it needs
  `AgentEngineAgents` to become a map, which is 235 `EngineAgentKind`
  references across 16 CLI files and orthogonal to the ablation controls #2374
  is blocked on.
- #2457 — `ContextRecall` and `ScopeReview` issue no model call, so they have
  no roster row and cannot be ablated. Two of #2374's fourteen features need
  that; the issue states the design question rather than prejudging it.
- #2458 — witness authoring is now decided in two places (`witness_writer` AND
  the roster), ANDed additively. The concrete bug hiding in the duplication is
  that the resume frame carries only one of them.

Closes #2381
Refs #2374

## Summary by Sourcery

Introduce a configurable responsibility roster for the Stella pipeline, allowing per-stage enablement and agent reassignment while preserving the default pipeline behaviour and stage ordering.

New Features:
- Add a responsibility roster in stella-pipeline to configure who performs each pipeline responsibility and whether individual stages (e.g., triage, verdict) run at all.
- Expose responsibility configuration via CLI settings, TOML, JSON engine config, and ArenaBench match templates, including posture reporting for ablations and self-grading.
- Support responsibility-aware routing in pipeline stages so triage, research, planning, witness authoring, guidance, and verdict can respect roster enablement and agent assignment.

Bug Fixes:
- Ensure triage ablation no longer implicitly removes verification by clamping the deterministic floor so verification still runs without triage.

Enhancements:
- Centralize responsibility-to-agent bindings in a dedicated roster module with validation, independence-loss detection, and compile-time defaults matching the shipped pipeline.
- Wire the pipeline to validate responsibility rosters before spend and to emit explicit posture warnings for ablated stages and loss of verifier independence.
- Refactor pipeline stage routing (triage, research, planning, witness authoring, guidance, verdict) to go through the roster seam rather than hard-coded role resolution.

Documentation:
- Document the responsibility roster and single-stage ablation semantics in the inference pipeline docs, including configuration examples and rationale for non-configurable stage ordering.

Tests:
- Add unit and integration tests for roster behaviour, single-stage ablation, invalid roster refusal, and ArenaBench responsibility configuration and posture emission.
- Update file size baseline and pipeline tests to cover the new roster module and roster-based ablation behaviour.


macanderson:
Found while speccing the follow-on (#2456 — opening the agent vocabulary), so
flagging here while it is still cheap to fix in-branch: **this PR arms a
divergence between the witness gate and the witness resolution.** Filed as
#2467 so it survives the merge either way; close it from here if you take the
fix.

Short version — the two ask different sources:

- **gate**: `can_author_independent_witness` → `witness_author_independence`
  (`crates/stella-pipeline/src/pipeline.rs:2881-2904`) resolves
  `Role::Verifier` **literally**.
- **resolution**: `resolve_witness_author`
  (`crates/stella-pipeline/src/pipeline/roster_wiring.rs:167-180`) resolves
  `assigned(ModelCallRole::WitnessAuthor)` — the roster.

On `main` they agree by construction, because `witness_author` *is* the
verifier. This PR makes that binding configurable, which is what makes the
divergence reachable.

The direction that worries me is the false positive: verifier ≠ worker but the
*assigned* author resolves to the worker. The gate says `Independent`, so
`authored_witness` is `true` (`pipeline.rs:1162-1168`) and `VerificationContract`
is chosen off it (`:1169-1178`) — specifically *not* `WorkerTestFirst`. Then
`resolve_witness_author`'s `model_ref != worker.model_ref` filter drops the
author and returns `None` silently, exactly as documented ("this function is
silent" — the gate was supposed to have announced it). `run_best_of_n` re-derives
the authoring flag from the `Option` so *that* heals, but the contract was fixed
several steps earlier. Result: no witness authored, worker never asked to write
one, nothing said.

`assigned` is already documented as silent and already called from independence
probes, so pointing the gate at it looks like the intended shape. Two details in
#2467 that are part of the fix rather than polish: the degradation text should
name the **assigned** agent instead of `pipeline_verifier_model` (or it
prescribes a row the operator did not write), and `Assigned::Withheld` should
not be reported as an outage — a disabled `witness_author` is the ablation this
PR exists to make legible.

Nothing else in the roster design gave me pause; `roster.rs`'s argument for why
order stays code and assignment becomes config is the right call, and
`default_agent`'s exhaustive match as the maintenance contract is a nice seam.
The full spec for the follow-on is at
https://github.com/macanderson/stella/issues/2456#issuecomment-5230381989 — it
takes a dependency on this fix (property P3).

Issue #2464, opened 2026-08-09
context: the counterfactual reflection prompt is argued for but unmeasured — no lesson-mix or volume comparison against the prompt it replaced


macanderson:
## Problem

The reflection prompt was rewritten to ask the counterfactual — "you will get a
task like this again with no memory of this turn; what do you want to have been
told?" — replacing a rediscovery-cost test operationalized as surprise. The
argument for it is stated in `crates/stella-cli/src/memory/reflection.rs` and
`doc:prompt-reflection`, and the witness tests prove the prompt *says* what it
is meant to say.

**Nothing proves it produces better memories.** That is a gap between what this
repo claims and what it enforces on its users: "a witness test, a property, a
golden diff someone actually read — a sentence in a PR description is not
evidence" (CLAUDE.md). A prompt-string assertion is a witness that the edit
landed, not that the edit was right.

The two previous versions of this prompt were both wrong, and both were
discovered by measurement rather than by reading — #768 by counting mined
lessons (eight of ten process self-critique), #944 by counting stored memories
(23 encoding six facts) and then by a proving-ground run that measured
hand-delivered conventions as worth *negative*. There is no reason this third
version is exempt from that pattern.

## What specifically is unmeasured

1. **Lesson mix.** Does the counterfactual frame shift the domain/process
   ratio, and in which direction? The topic lists that used to hold process
   self-critique back are gone; `saves` and `kind` are supposed to hold it back
   instead. That is a prediction, not an observation.
2. **The volume change.** The per-turn cap rose 3 → 8 and the topic
   restrictions were removed, so more is recorded per turn. Reflection memories
   compete for a recall budget of `max_frames` (default 5). More recorded is
   not more recalled; it may be *less* recalled, if the additional lessons
   crowd out better ones. This is the specific way the change could make things
   worse while looking like it made them better.
3. **Whether the memories help.** The end-to-end question, and the one that is
   structurally unanswerable today — see Blocked on, below.

## Files involved

- `crates/stella-cli/src/memory/reflection.rs` — the prompt and
  `MAX_LESSONS_PER_TURN`
- `crates/stella-cli/src/memory/tests/ab_control.rs`,
  `crate::memory::AB_CONTROL_TAG` — the existing A/B arm tagging, written for
  exactly this kind of comparison (#1221). Episode summaries carry the tag; the
  comparison is a query over them.
- `crates/stella-cli/src/memory/learning.rs` — `partition_known`, whose
[...]

Issue #2460, opened 2026-08-09
cli: reflection judges a turn from its last 12 messages truncated to 300 chars — the expensive middle is never seen


macanderson:
## Problem

Reflection asks the model what it learned from a turn. It shows the model
**the last 12 messages, each truncated to 300 characters** — at most ~3,600
characters of a turn that may have burned hundreds of thousands of tokens
across dozens of tool calls.

`crates/stella-cli/src/memory/reflection.rs`:

```rust
const REFLECTED_TAIL_MESSAGES: usize = 12;
…
let content: String = message.content.chars().take(300).collect();
```

The consequence is structural and caps every prompt change made above it: the
expensive part of a turn — the wrong approach taken at step 4 and abandoned at
step 22, the tool that returned an error nobody read, the twenty minutes spent
waiting on the wrong command — is in the middle of the transcript, and the
middle is exactly what the tail window drops. What survives is the last few
messages, which on a successful turn are the summary and the sign-off.

So the model is asked *what would have made this faster* by a witness that
cannot see where it was slow. Whatever it answers about the turn's cost is
reconstruction, not observation. That is the same failure mode CLAUDE.md
already names for benchmark conclusions — "open `stella-events.jsonl` before
claiming anything; the UI and `result.json` are projections of it" — applied to
the agent reflecting on itself. Reflection is currently reading its own surface
signal.

300 characters is the tighter of the two limits in practice: it is shorter than
most tool results, so a `tool` message usually contributes its first line and
nothing else. A failing test's actual assertion is routinely past the cut.

## Why this is not just "raise the numbers"

Reflection runs on every tool-using turn (#2155 tracks whether it should), at
pinned-low effort with a bounded output cap, and its input tokens are billed
every time. Widening the window naively multiplies that cost across every turn
in every session — against the same "cheaper" objective the reflection prompt
now asks the model to optimize for. The interesting question is *selection*,
not size: the turn already emits a structured event stream
(`step_usage`, `tool_start` → `tool_result`, retry and loop-detector records)
that names where the time and money actually went, and reflection reads none of
it.

## Files involved

- `crates/stella-cli/src/memory/reflection.rs` — `REFLECTED_TAIL_MESSAGES`,
  the 300-char truncation, `reflection_tail()`
- `crates/stella-cli/src/memory/learning.rs` — `reflect_and_record`, the caller
  that assembles the transcript slice
- `crates/stella-protocol/src/event.rs` — the event shapes that carry real
[...]

Issue #2459, opened 2026-08-09
context: reflection lesson triggers are recorded but never consulted at recall


macanderson:
## Problem

Reflection now asks the model for a `trigger` on every lesson — the condition a
future task has to satisfy for the lesson to be worth reading — and records it
to `.stella/private/reflections.jsonl`. **Nothing reads it.** Recall still
matches on the lesson body alone, so a lesson that states exactly when it
applies is retrieved by the same embedding guess as one that does not.

This is deliberate scaffolding, not an oversight, and this issue is the wiring
half filed alongside the PR that shipped it.

The field earns its keep at *write* time even unwired: requiring a trigger
prunes candidates the model cannot situate, at the one moment the transcript is
still there to check them against. But the retrieval win — a memory that
announces its own applicability instead of competing purely on cosine
similarity — is entirely unclaimed.

## Why it was not wired in the same PR

The obvious implementation is to fold the trigger into the stored memory body,
so it lands in the embedding. That changes the text that
`SessionMemory::partition_known`'s restatement band was tuned on (#2358) — the
band that decides whether a lesson is novel or a paraphrase of one already
held. Appending a sentence to every lesson body shifts every similarity score
in that comparison at once, so it cannot be judged by inspection: it needs the
same measurement #2358 got. Doing it blind inside a prompt change would have
confounded two things that must be measured separately.

## Files involved

- `crates/stella-cli/src/memory.rs` — `ReflectionLesson::trigger`, where the
  field is defined and its non-consumption is documented
- `crates/stella-cli/src/memory/reflection.rs` — the prompt that asks for it
- `crates/stella-cli/src/memory/learning.rs` — `reflect_and_record`, where
  `MemoryInput::reflection(&l.lesson, …)` decides what text is stored
- `crates/stella-cli/src/memory/learning/dedupe.rs` — the restatement band that
  constrains the change
- `crates/stella-cli/src/memory/recall.rs` — the retrieval side that would
  consume it

## Verify / repro

```sh
# Triggers are being written:
rg '"trigger"' .stella/private/reflections.jsonl | head
# Nothing consumes them:
rg -n 'trigger' crates/stella-cli/src/memory/recall.rs   # no hits today
```

## Definition of done

1. Decide *where* the trigger participates in retrieval — folded into the
   embedded body, held as a separate scored field, or used as a post-retrieval
   filter. Name the choice and why.
2. Whatever the choice, re-measure the restatement band the way #2358 did
[...]

Issue #2417, opened 2026-08-09
docs/prompts transcribes every role's prompt verbatim and nothing checks it against the constants


macanderson:
## Problem

`docs/prompts/` (`doc:prompts`, landed alongside this issue) transcribes every agent role's instruction block **verbatim** from the Rust constants — 14 pages, one per `ModelCallRole` variant. Nothing checks the transcription. The moment someone edits `TRIAGE_INSTRUCTIONS`, `WITNESS_SYSTEM_PROMPT`, `SYSTEM_PROMPT`, or any of the others, the page silently becomes a lie that reads exactly as authoritative as it did the day it was correct.

This is the failure shape this repository has already been bitten by repeatedly and now guards elsewhere: a copy that nothing checks (#1435 god-file tables, #1437 gate-step lists, #2231 shared prompt contracts, #1439 wire schemas). Prompt text is a worse instance than most, because the reader's whole reason for opening the page is that they do not want to read the Rust.

The pages state the code is normative and name the symbol each was rendered from, which makes the drift *checkable by hand*. That is a mitigation, not a guard.

## Proposed shape

A `make prompt-docs` step that extracts each documented block from its defining constant and compares it against the fenced block in the page — failing by **name** on a mismatch, the way `check-gate-parity` and `check-god-files` do.

Two candidate mechanisms, and the choice is the design work here:

1. **Generate the fenced blocks.** A `--write` mode rewrites the code fences in place from the constants, and CI runs the check mode. Strongest guarantee, and it is how `docs/wire/` already works. Cost: the prose around each block has to survive regeneration, so the pages need stable extraction markers.
2. **Assert, do not generate.** A test in the crate that owns each constant asserts the page contains its exact text. Cheaper, keeps the pages hand-written, but needs the doc path in a test — and the pages will move, which is the exact fragility `doc:`-id citation exists to avoid.

Option 1 is likely right, following `wire-schema`'s precedent, but read `scripts/check-doc-links.py` first: doc paths are resolved through `docs/manifest.json` by id, so a generator can locate a page by `doc:prompt-triage` rather than by path and survive a move.

## The constants to cover

| Page | Constant | Crate |
|---|---|---|
| `doc:prompt-triage` | `TRIAGE_INSTRUCTIONS` | `stella-pipeline/src/triage.rs` |
| `doc:prompt-research` | `RESEARCH_SYSTEM_PROMPT` | `stella-pipeline/src/research.rs` |
[...]

PR #2465, opened 2026-08-09
feat(stella-cli): ask reflection the counterfactual, not a proxy for it

macanderson:
## What

Reflection now asks the agent the counterfactual — *"you are about to be handed
a task like this one again, with no memory of anything that happened here. What
do you want to have been told before you start?"* — instead of a proxy for it.
The topic lists go with it, and a required argument replaces them.

## Why

This prompt has been wrong three times, and every time the same way: it asked
for a **proxy** and got exactly the class of answer the proxy described.

| | Proxy asked for | What came back |
|---|---|---|
| #768 | "what should change next time to avoid repeating this failure?" | a question about the agent, answered about the agent — 8 of 10 mined lessons were process self-critique, none a repository convention |
| #944 | "where things live" | 23 memories encoding 6 facts, each one file-read away; the proving ground measured hand-delivering those conventions as worth **negative** |
| the repair for #944 | rediscovery cost ("find it in under a minute? DISCARD"), operationalized as surprise | right about one class, blind to another |

The third is the subtle one and the reason for this PR. **Surprise measures
novelty; what a memory is worth is savings.** They agree on most facts and come
apart exactly where the money is — on a fact that is trivial to look up and
expensive not to know. *"The whole suite takes twenty minutes; the scoped run
takes forty seconds"* is one grep into a Makefile, so the rediscovery test
orders it discarded — after it has already cost three turns of waiting, and
while it goes on costing them in every future session, because nobody greps for
what they do not know to ask.

The new discard test is whether knowing the fact **would have changed an
action**. That subsumes the rediscovery rule — a fact you would have looked up
anyway changes no action, and is still discarded — without inheriting the blind
spot.

## No topics, but a required argument

Both the "do NOT record" and "DO record" enumerations are gone and **nothing
replaces them**. Each was a guess about subject matter made by whoever wrote the
prompt; the model that just ran the turn is better placed to know what the turn
cost it, and replacing one topic guess with another is how this failed three
times already.

What replaces them is the *shape of the argument*, not the topic. Every lesson
now arrives with:

- **`trigger`** — what has to be true of a future task for this to matter. A
  lesson whose trigger the model cannot state is one it has not finished
  learning.
- **`saves`** — what knowing it would have bought *here*, pointing at the moment
  in the transcript it would have changed. This is the field that makes a lesson
  refutable rather than merely asserted, and it is refutable against the very
  transcript the model is looking at.

Freeing the topic without requiring the argument would only move the guess from
the prompt's author to the model. `kind` survives the cull because it is not a
topic question — it selects the recall tier — but it is now asked as *transfer*
("still true on a different task here?") rather than as subject matter, which is
what it was always read as.

## Two bounds that follow

- **Per-turn cap 3 → 8.** Three is below the number of distinct things a busy
  turn teaches, so on those turns the cap was the decision — made by arithmetic,
  after the model had already done the work of finding them. Eight is a backstop.
  It is deliberately not unbounded: reflection memories compete for a 5-frame
  recall budget, so a turn permitted to file twenty dilutes recall rather than
  adding to it.
- **Output cap 2048 → 4096.** A response that overruns it parses to zero lessons
  *exactly like an empty one*, so the symptom of setting it too low is not a
  truncated lesson but a turn that silently taught nothing. A lesson is now three
  prose fields, up to eight of them.

## `save_memory`

Same question, and it keeps the one guideline that genuinely belongs there —
which is about **cost, not content**. Unlike recalled context, `.stella/memories/`
is the byte-stable prompt prefix (invariant 7): paid for on every model call in
every session whether relevant or not. That asymmetry is why the same freedom
carries a different warning on the two channels.

## Docs

`doc:prompt-reflection` is re-synced (#2417's concern) and **one factual error in
it is fixed**: it claimed reflection writes `.stella/memories/*.md`. It does not
— that is `save_memory`'s directory alone. Reflection lessons reach a future
prompt through recall, which is the entire cost distinction above.

## Witness

New assertions run against the pre-change prompt via the shadow protocol (revert
source, keep tests):

```
both_outcomes_ask_the_counterfactual_not_a_proxy ......................... FAILED
the_prompt_states_the_action_change_test_on_both_outcomes ................ FAILED
every_lesson_must_carry_its_trigger_and_the_moment_it_would_have_changed .. FAILED
the_prompt_prescribes_no_topics .......................................... FAILED
reflection_dispatches_low_effort_with_its_bounded_output_cap ............. FAILED
the_prompt_does_not_offer_a_greppable_fact_as_its_model_lesson ............... ok
```

The one that passes is the pre-existing guard against the anti-example that
carried the last two failures; it correctly still holds. On the new code all 7
pass, alongside 228 `memory::` tests and 752 in `stella-tools`.

`the_per_turn_lesson_cap_is_above_what_a_turn_typically_teaches` is excluded from
the table above because it references `MAX_LESSONS_PER_TURN`, which does not
exist on the base — a compile error rather than an assertion failure, so it would
have masked the other five.

## Renamed tests (deleted-test guard)

None removed; three renamed, each pinning the same seam under the new rule:

- `the_prompt_states_the_rediscovery_test_on_both_outcomes` →
  `the_prompt_states_the_action_change_test_on_both_outcomes`
- `a_successful_turn_is_asked_what_surprised_it` →
  `both_outcomes_ask_the_counterfactual_not_a_proxy`
- `parse_lessons_drops_invented_domains_and_caps_at_three` →
  `parse_lessons_drops_invented_domains_and_caps_at_the_per_turn_limit`

## What this PR does NOT establish, stated plainly

The witness proves the prompt **says** what it should. It does not prove it
**produces better memories** — and both previous versions of this prompt were
correct-sounding and wrong, each caught by counting rather than by reading. The
lesson-mix and volume comparison is **#2464**, filed rather than skipped, and it
names the specific way this change could make things worse (more recorded,
competing for the same 5 recall slots).

The end-to-end question is structurally unanswerable today: **#2195** — nothing
has ever called `cite_memory`, so no memory in this repository has ever been
recorded as having helped.

## Not run locally

`clippy` and the full workspace test tier were **not** run: an 8-hour benchmark
match is live on this machine and the standing rule is no heavy local builds
during a bench run. `make guards-fast` (compile-free guards + `fmt --check`)
passes, as do `stella-cli`'s `memory::` tests and all of `stella-tools`. CI runs
the rest.

## Issues filed

- #2459 — `trigger` is recorded but not consulted at recall (the wiring half of
  the scaffolding this PR ships, with the #2358 dedupe-band interaction named as
  the constraint)
- #2460 — reflection judges a turn from its last 12 messages truncated to 300
  chars, so the expensive middle of a turn is never seen. **This caps how good
  any wording here can be**, and is the highest-leverage fix in the subsystem
- #2464 — this change is argued for but unmeasured

Refs #2459
Refs #2460
Refs #2464
Refs #2417

## Summary by Sourcery

Update the reflection system to ask the agent for counterfactual, action-changing lessons instead of proxy signals, and align caps, schema, and docs with this new behavior.

New Features:
- Require each reflection lesson to include a trigger condition and an explicit description of what action or outcome it would have changed.
- Increase the maximum number of lessons that can be recorded per turn and raise the reflection output token cap to support richer lessons.

Enhancements:
- Refine the reflection prompt to remove topic prescriptions, focus on the counterfactual question, and distinguish lessons by how far they generalize rather than by subject matter.
- Adjust lesson parsing and deduplication logic to respect the new per-turn limit and additional lesson fields while keeping within-batch stutters out of the store.
- Clarify the `save_memory` tool description so that it emphasizes action-changing, next-session value and when a lesson applies.

Documentation:
- Resync and expand the reflection prompt documentation to describe the new counterfactual framing, required lesson arguments, updated caps, and the distinction between reflection recall and `save_memory`-backed prefix memories.

Tests:
- Add and rename reflection prompt tests to pin the counterfactual framing, action-change test, absence of topic lists, required lesson fields, and updated caps, ensuring the prompt does not regress to previous failure modes.


macanderson:
`make gate` is red on this branch at the `lint` step — clippy at `-D warnings`, so CI's required job will fail too:

```
error: this assertion has a constant value
   --> crates/stella-cli/src/memory/reflection/tests.rs:270:5
    |
270 | /     assert!(
271 | |         super::MAX_LESSONS_PER_TURN > 3,
...
    = note: `-D clippy::assertions-on-constants` implied by `-D warnings`
```

`git blame` puts it on this PR's own commit (65265226), so it is not inherited from main.

The fix clippy names is the better assertion anyway — the predicate is constant, so a runtime assert on it can only fail on a code change, never on a run. A `const` block makes that literal and turns a cap regression into a build failure instead of a test failure:

```rust
const {
    assert!(
        super::MAX_LESSONS_PER_TURN > 3,
        "the cap is back below what a turn teaches, which makes it the \
         decision again rather than a backstop"
    );
}
```

I have applied exactly that on `worktree-fix-2459-trigger-recall` (the #2459 wiring, stacked on this branch), because that branch inherits the failure and cannot be green without it — so if this PR merges first, the duplicate edit resolves cleanly. Cherry-pick it here if you would rather this PR go green on its own; I did not push to this branch since it is yours.

macanderson:
## Merged `main` — two conflicts, both the same shape

`main` picked up #2447 (`cf892151`), which touched this branch's exact seam. Both
conflicts were **two independent repairs to one number**, where neither reason
subsumes the other, so "take a side" would have silently dropped a real fix
either way.

### 1. The output cap

| | Fix | Reason |
|---|---|---|
| **#2447** | named the constant `LESSONS_OUTPUT_CONTRACT` and pads it with `with_reasoning_headroom` | `max_output_tokens` is one wire number and a reasoning model bills thinking against it — the bare cap came back spent entirely on reasoning (#2174) |
| **this branch** | raised the number 2,048 → 4,096 | a lesson is now three prose fields, and up to `MAX_LESSONS_PER_TURN` of them may return |

Resolved as **main's structure carrying this branch's number**: the written
contract is 4,096 and the headroom still rides on top, so the wire cap is 8,192.
`starvation.rs` already sizes `STARVED_RETRY_CAP` against
`with_reasoning_headroom(4096)` and calls it "the largest padded contract in the
tree" — so the padded contract lands exactly inside the envelope already designed
for it, and no downstream constant moves.

### 2. The dispatch witness

#2447 refactored the inline assertion into a `dispatch_shape` helper that
*returns* the shape so its two tests can assert different halves. Took that
refactor and moved this branch's number into the caller, which now pins **both**
facts in one test.

**Demonstrated, not asserted** — flipping the contract back to 2,048 fails it:

```
assertion `left == right` failed: reflection must dispatch as a bounded,
pinned-low management call whose cap covers its output contract PLUS thinking room
  left: (Some(Low), Some(6144))
 right: (Some(Low), Some(8192))
```

`6144` is `2048 + 4096`, which proves main's headroom is still applied; `8192`
proves this branch's 4,096 contract is what's live. Drop either fix and it goes
red.

### Two corrections to the description above

- **"Output cap 2048 → 4096"** now reads: the *written contract* is 2,048 →
  4,096; the wire cap is that plus `REASONING_HEADROOM_TOKENS`.
- **`reflection_dispatches_low_effort_with_its_bounded_output_cap`** no longer
  exists under that name — #2447 renamed it to
  `reflection_dispatches_low_effort_with_a_cap_that_leaves_room_to_think` on
  `main`. The rename is main's, not this branch's, so the deleted-test guard sees
  no deletion. The three renames this branch *does* make are still listed above
  and unchanged.

`docs/prompts/reflection.md` is corrected in the same push: the merge made its
"Output cap | 4,096" row ambiguous against a wire that now carries 8,192, so the
page says which number is which and states the headroom's own reason.

### Verified

`stella-cli` 1577 · `stella-core` 1063 · `stella-tools` 761 — all pass. `cargo
fmt --check` clean; `clippy -D warnings` clean on all three crates; `doc-links`
OK. No bench match was live.

### Filed

- **#2485** — `check-file-size` misreads a local `git merge origin/main`. It
  assumes any merge commit is a `refs/pull/N/merge` checkout (`HEAD^1` = base),
  but a local merge has the reverse parent order, so it blamed this PR for six
  `bench/**` Python files it never touched. `FILE_SIZE_BASE_REF=origin/main`
  passes with "nothing went over by this change"; CI reads the correct base and
  is unaffected. **No baseline was regenerated here** — that drift is main's and
  lands on its own PR from a fresh `main`, per AGENTS.md.

macanderson:
## Unblocked — two failures, neither of them the prompt change

Pushed `0edb722d9`. The clippy `assertions-on-constants` failure you named is
already gone (the `const` block arrived here with #2482's merge). What was left
was these two, and they are unrelated to each other.

### 1. The file-size ratchet — a split, not a baseline row

`crates/stella-cli/src/memory/tests.rs` had reached 1503 lines: 1489 on `main`
plus this PR's own 14. It carries no baseline entry, so the guard is right and
the remedy is fixed — AGENTS.md § "God files" gives a first crossing a split and
never a baseline row.

The reflection **lesson-mining** seam moves to
`crates/stella-cli/src/memory/tests/reflection_mining.rs`: everything between a
model's reflection response and a stored lesson or self-review, together with
the three fixtures only it reads (`lessons_of`, `lessons_with`,
`reflection_execution_ids`). Parent and child land at 763 and 764 lines.

The line count chose the moment, not the seam — this is the same cut
`skill_creation` and `trigger_recall` already made, and its `mod` comment says
so.

**Nothing was deleted, and that is checked rather than asserted.** The set of
`#[test]`/`#[tokio::test]` names under `crates/stella-cli/src/memory/` is
identical before and after the move — 242 either side, no additions, no
removals — which is exactly the question `scripts/check-deleted-tests.sh` asks,
since it matches bare function names unqualified by file. So there is nothing to
name in the description under "Renamed tests".

The six `bench/**` lines in that job's log are inherited drift, correctly
reported as not fatal and not this PR's to repair.

### 2. The branch had not compiled since the last merge

Worth flagging separately, because no conflict marker ever appeared and the
symptom looked like the ratchet's:

- `main` grew a seventh argument on `reflect_and_record` — `posture: ReflectionPosture`.
- #2482 then landed `memory/tests/trigger_recall.rs`, holding two **six**-argument
  calls, written before that parameter existed.

Neither side touched the other's lines, so git merged them clean and the result
built nothing:

```
error[E0061]: this method takes 7 arguments but 6 arguments were supplied
   --> crates/stella-cli/src/memory/tests/trigger_recall.rs:97:10
```

Both call sites now pass `ReflectionPosture::default()` — which its own doc
comment defines as *"exactly as before this was threaded"*, i.e. the behaviour
those tests were written against, and what every other call site in the tree
already passes.

### Verified

`cargo test -p stella-cli memory::` → **242 passed, 0 failed**. The file-size
guard passes for this change; `god-files` is OK.

Not run locally: the full `--workspace` tier and clippy across every crate. CI
covers both and is running now.

macanderson:
## Merged `main` in — four conflicts, plus one collision git could not see

This PR was `CONFLICTING` and red on `file size ratchet`. Both are fixed; all
checks are green on `a4be33bc` and it is now `MERGEABLE` / `CLEAN`. The ratchet
half was `0edb722d9` (split of `memory/tests.rs`). This note covers the merge,
because the resolutions were judgment calls rather than mechanical picks.

### The trap that shaped the merge order

`main` (#2494) replaced `reflect_and_record`'s `(transcript, succeeded)` params
with a bundled `TurnEvidence`, updating five call sites — the exact five that
`0edb722d9` had just moved into `memory/tests/reflection_mining.rs`. Merging the
already-split tree would have taken *our* deletion of those lines and dropped
`main`'s signature fix on the floor: clean merge, no conflict marker, red
compile.

So `main` was merged against the **unsplit** tree, and the split re-derived by
diffing `main`'s version of the moved region against the moved file. After the
port, the only remaining difference is this PR's own rename of
`parse_lessons_drops_invented_domains_and_caps_at_three`.

### Resolutions

- **`reflection.rs` — `REFLECTED_TAIL_MESSAGES` is dropped, not kept.** This
  branch carried it under a comment calling the tail window too small and naming
  #2460 as "filed rather than folded in". `main` **is** #2460: the constant is
  deleted there and `digest::build(evidence)` selects from the turn instead.
  Keeping our side would have shipped dead code under a comment describing
  landed work as still pending. `MAX_LESSONS_PER_TURN = 8` and the counterfactual
  framing are ours and survive; the digest selection is `main`'s.
- **`docs/prompts/reflection.md`** — additive, not competing. Both sections
  kept: ours on the question the prompt asks, `main`'s `## What {digest}
  contains` on the input it asks about.
- **`memory/reflection/tests.rs`** — module docs combined; both statements are
  true after the merge.
- **`memory/tests.rs`** — our (empty) side, with `main`'s five `TurnEvidence`
  conversions ported into `reflection_mining.rs` by hand.
- **`memory/tests/trigger_recall.rs`** — the collision with no conflict. The file
  is new on this branch (#2482), so `main`'s migration never touched it and there
  was nothing to conflict on. Its two call sites were left on the old signature
  and are converted.

Full reasoning is in the merge commit message on `056a6c06b`.

### Verified

Locally, in the worktree with `CARGO_TARGET_DIR` unset: `./scripts/check-file-size.sh`
exits 0 (only pre-existing inherited bench drift, non-fatal — tracked in #2472,
where I posted the current six-file figures); `cargo fmt --all --check` clean;
`cargo clippy -p stella-cli --all-targets -- -D warnings` clean; `cargo test -p
stella-cli` 1615 unit + all integration binaries, exit 0.

Deleted-test guard: I diffed the `#[test]`/`#[tokio::test]` name sets between
`origin/main` and the merge result. The only three names on `main` missing from
the result are the three renames already listed in the PR description above, and
the merge result carries a net +13 test names.

The later `a4be33bc` is a second merge of `main` for `strict: true` branch
protection (it was `BEHIND` behind #2498/#2502). It touches only `Cargo.lock`,
`Cargo.toml` and `packaging/homebrew/stella.rb` — no source — so the
verification above still describes the code under review.

Issue #2574, opened 2026-08-09
gate: nothing compiles the merge result, so two green PRs can compose into a main that does not build (happened 3x today)


macanderson:
## Problem

A pull request is only ever compiled **as its own branch**, never as the tree that merging it would produce. So two PRs that each compile, each pass the full gate, and merge without a textual conflict can compose into a `main` that does not compile at all.

This happened three times on 2026-08-09, all in `crates/stella-tools/src/overview.rs`, all the same shape: **two branches each add a private helper with the same name at a different file offset.** Git sees no overlapping lines and reports a clean merge; Rust sees two definitions.

| # | Branches | Result |
|---|---|---|
| 1 | #2551 (on main) vs #2538 (open) | caught locally while merging main into #2538 — `error[E0428]: the name git_lines is defined multiple times` |
| 2 | #2538 (merged) vs #2561 (merged 14s before #2562) | **landed on main.** Same E0428, at `overview.rs:246` and `:548` |
| 3 | same file, `git_section` vs `repository_section` | compiled fine and was *semantically* wrong: `project_overview` emitted two sibling git keys, each with its own "is this a repository" flag |

Case 2 is the expensive one. `main` at `dbefb3b1d` failed **every compile step in the same job** — `cargo clippy -D warnings`, `cargo doc -D warnings`, the prompt-cache golden fixtures, `cargo test`, `stella context validate`, and the release smoke build ([run 31337284898](https://github.com/macanderson/stella/actions/runs/31337284898)). A red `main` reds every open PR downstream. Fixed by #2565, ~20 minutes later.

Note that case 3 shows the class is not only about *name* collisions: a merge can compile and still be wrong, so a compile check is necessary but not sufficient.

## Why the existing guards do not catch it

- **Branch protection is not strict.** With "require branches to be up to date before merging" off, a PR merges against a `main` its CI never saw. #2245 documents why turning it on has been painful (the ~20-minute release-bot cadence makes manual `update-branch` non-convergent), and #2152 documents that only 3 checks are required at all.
- `gh pr view` reporting `mergeable: MERGEABLE` means *no textual conflict*, nothing more. It is routinely read as "this merge is safe".
- Nothing compiles `refs/pull/N/merge` — the merge preview ref GitHub already computes and that `check-file-size.sh` and `check-deleted-tests.sh` both already consume for their base comparison.

## Suggested fix

Cheapest first, and they compose:

[...]

PR #2587, opened 2026-08-09
test(stella-tui): unbreak main — restore the two roles a golden clobber dropped

macanderson:
## Unbreak main — a deck golden lost two roles to a parallel-merge clobber

`main` has been red on `fmt + clippy + test` since 22:57 UTC: `deck_render_snapshots_pin_the_floating_cards` fails because `crates/stella-tui/tests/snapshots/deck/card_models.txt` no longer matches what the deck renders.

**Not a rendering regression. The golden is stale, and it got that way by being clobbered** — the exact shape `check-deleted-tests` exists for, on a fixture rather than a test.

### What happened

`git log` on the fixture, newest first:

```
9ce1c4506 feat(stella-tui): PROOF panel …            (#2568)
fba2695cb feat(stella-tui): recall as a table …      (#2566)
fa368a7ac feat(stella-cli): research and plan as independently configurable roles (#2553)
```

#2553 added the `research` and `plan` roles and blessed this golden with all six rows — `lead`, `think`, `research`, `plan`, `work`, `verify`. #2566 then blessed the same file from a base that predated #2553, so its version carries #2566's own changes **and the pre-#2553 four-row roster**. `research` and `plan` are simply absent:

```diff
 │lead      zai/glm-5.2-air                                             │
 │          medium · thinking on · from default_model                   │
 │think     z-ai/glm-5.2                                    unassigned  │
 │          low · thinking off · from agents.triage.model               │
-│research  zai/glm-5.2-air                                             │
-│          low · thinking off · from default_model                     │
-│plan      zai/glm-5.2-air                                             │
-│          medium · thinking on · from default_model                   │
 │work      zai/glm-5.2-air                                     served  │
```

Both PRs were green against the `main` they branched from, and the merge was textually clean — git had no conflict to report because one side simply did not contain the other's lines. #2568 then inherited the four-row roster and blessed on top of it.

The renderer still emits all six roles (#2553's code is untouched in `main`), so the golden is two rows short and every row below shifts. That is the whole failure.

### The fix

Re-blessed `card_models.txt` only, and read the diff. The result carries all three PRs' intended changes at once:

- `research` and `plan` rows restored (#2553);
- the transcript at 24 lines with the recall rendering (#2566);
- the `PROOF ✓ proved` panel in place of `DONE VERIFICATION` (#2568).

No other golden is affected — `research  zai` / `plan      zai` appear in no other snapshot, and `card_models.txt` was the only file carrying the stale roster.

No source changed. This is a fixture correction, so there is no witness test: the suite itself is the witness, and it goes green.

### Deleted tests

None.

### Left behind

The underlying hazard — a blessed golden silently reverting another PR's blessing of the same file — is not caught by anything today. `check-deleted-tests` compares the merge base against the merge result for `#[test]` function names, which is the same question one level up, but it does not look at fixture content.

Not filed separately: #2574 already tracks this class ("two green PRs can compose into a main that does not build"), and I added this as a fourth instance there — https://github.com/macanderson/stella/issues/2574#issuecomment-5234512925. It is the first instance that **compiles cleanly**, which is direct evidence for that issue's own note that a compile check is necessary but not sufficient: running the suite on `refs/pull/N/merge` would have caught it, a compile of the same ref would not.

### DIFF
--- a/crates/stella-tui/tests/snapshots/deck/card_models.txt
+++ b/crates/stella-tui/tests/snapshots/deck/card_models.txt
@@ -19,18 +19,18 @@
 │stage execute   model —   turn $0.0390 / $2.50  ·  observed                                                                                                   │
 └──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
 ┌ transcript · 24 lines · following ───────────────────────────────────────────────────────────────────────────────────┐┌ PLAN ○ pending approval 0/3 ─────────┐
-│                                                                                                                      ││Refactor the automations store into a…│
-│  Planning the automations cluster: list, editor, triggers, workflows.                                                ││○ 1. extract types                    │
-│                                                                                                                      ││○ 2. move hooks                       │
-│☰  plan  4/7  ·  Wire the triggers API                                                                                ││○ 3. update 9 imports                 │
-│                                           ┌ models  resolved routing ───────────────────────────────── esc close ┐   ││                                      │
-│── WITNESS ────────────────────────────────│session   glm-5.2                                                     │───││                                      │
-│                                           │auto      model off · effort on · thinking off                        │   ││                                      │
-│── EXECUTE ────────────────────────────────│                                                                      │───││                                      │
+│                                           ┌ models  resolved routing ───────────────────────────────── esc close ┐   ││Refactor the automations store into a…│
+│  Planning the automations cluster: list, e│session   glm-5.2                                                     │   ││○ 1. extract types                    │
+│                                           │auto      model off · effort on · thinking off                        │   ││○ 2. move hooks                       │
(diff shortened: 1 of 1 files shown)


macanderson:
Closing as a duplicate — superseded by #2583, which landed the identical fix first (`4be059883`).

The `PR carries a diff` guard caught it correctly: `main` already contains this branch's content, so merging would add a commit whose message describes changes it does not contain. Two sessions diagnosed the same red `main` in parallel; #2583 won the race.

Verified on `origin/main` — `card_models.txt` now carries the `research` and `plan` rows again.

The analysis is not wasted: I left it on #2574 as a fourth instance of that class — https://github.com/macanderson/stella/issues/2574#issuecomment-5234512925 — and it is the first instance that **compiles cleanly**, so it is evidence that compiling `refs/pull/N/merge` is not enough and the suite has to run on it.

Issue #2411, opened 2026-08-09
arenabench: the certified h2h capped Stella at $1.00 and left Claude Code uncapped — all three Stella losses are budget terminations, not capability failures


macanderson:
## Problem

In match `5292a68cdabf` ("certified h2h 4x6: Claude Code vs Stella bare loop (main 0.7.34)") the two
contestants were run under **different resource ceilings**, and every one of Stella's three losses is a
termination by the ceiling the opponent did not have.

From the match's own `spec.json`:

```json
{ "id": "claude-code", "engine": { "budget_usd": null, "max_tokens": null  } }
{ "id": "stella",      "engine": { "budget_usd": 1.0,  "max_tokens": 32000 } }
```

Claude Code was uncapped. Stella was capped at $1.00 per trial. The task allowance is
`[agent] timeout_sec = 900.0` for both.

## Evidence — all three losses are budget terminations, not capability failures

Every scored Stella trial, from `agent/stella-events.jsonl` (ground truth, not `result.json`):

| trial | reward | spend | wall clock used | stopped by |
|---|---|---|---|---|
| `git-multibranch__S4PjQ6Q` | 0 | $1.0090 | 303s / 900s | `budget_denied` |
| `git-multibranch__wWarrQF` | 0 | $1.0775 | 324s / 900s | `budget_denied` |
| `sqlite-with-gcov__iDi8mCw` | 0 | $1.0010 | 322s / 900s | `budget_denied` |
| `git-multibranch__gN8WPzn` | 1 | $1.0047 | 317s | `budget_denied` (finished first) |
| `large-scale-text-editing__YCeSEJG` | 1 | $1.0344 | 525s | `budget_denied` (finished first) |
| `regex-log__3P73Fx3` | 1 | $1.0595 | 471s | `budget_denied` (finished first) |
| `regex-log__zX9x6Wm` | 1 | $1.0121 | 486s | `budget_denied` (finished first) |
| `sqlite-with-gcov__5ruCbkc` | 1 | $1.0008 | 305s | `budget_denied` (finished first) |
| 4 further trials | 1 | $0.66–$0.97 | — | ran to completion under the cap |

Every trial that came in under $1.00 won. Every loss was killed at the ceiling, at roughly **one third of
its allowed wall clock**. The agent was stopped by money with ~10 minutes of unused clock.

The final journal events show each loss was one to four steps from done:

- `sqlite-with-gcov__iDi8mCw` — had already proven the instrumented build worked end to end
[...]

Issue #2341, opened 2026-08-08
arenabench: server silently serves a stale web/ bundle — stamp the build, log it at startup, warn when ui/ is newer


macanderson:
## Problem

`arenabench server` serves a pre-built static export from `arenabench/arenabench/web/` (`WEB_ROOT = Path(__file__).resolve().parent / "web"` in `arenabench/arenabench/server.py:88`). That bundle is gitignored (untracked by #1606) and only regenerated when someone runs `npm run build` in `arenabench/ui/`. Nothing tells the operator how old the bundle they are looking at is, so UI changes silently fail to appear.

This bit for real on 2026-08-08: the served bundle was built **Aug 6 02:44**, while the logo work landed Aug 7 (`fd528e9b`, #2107 wordmark lockup; then the Nebula recolor `aa288fd9` / revert `78a60f30`). The operator concluded they were "running an old version" of arenabench. The stale bundle also lacks the `brand/` directory that the current `arenabench/ui/components/topbar.tsx` loads (`/brand/wordmark-color-{dark,light}.svg`), so with a fresh-source/stale-bundle mix the logo would 404 outright.

Compounding it, `arenabench/ui/package.json`'s `description` still says the client is "Built to static files and committed into arenabench/web/" — false since #1606 untracked the bundle. The one place a reader might learn the workflow states the old one.

## Fix

1. Have `arenabench/ui/scripts/sync-web.mjs` stamp the build time (and ideally the source git SHA) into the bundle — e.g. extend `arenabench/arenabench/web/GENERATED.md` or write a small `build-info.json`.
2. At server startup, log that stamp next to the listen line ("web bundle built 2026-08-06T02:44, source sha …"). Optionally compare the newest mtime under `arenabench/ui/{app,components,lib,public}` against the stamp and print a one-line warning when the source is newer ("UI source is newer than the served bundle — run npm run build in arenabench/ui/").
3. Fix the `arenabench/ui/package.json` description to say the bundle is generated and gitignored, not committed.

## Verify

- `npm run build` in `arenabench/ui/`, start the server, see the build stamp in the startup log.
- Touch a file under `arenabench/ui/components/`, restart the server, see the staleness warning.
- `rg 'committed into' arenabench/ui/package.json` returns nothing.

## Constraints

- The server must keep working when the stamp is absent (old bundles, wheels built from sdists) — degrade to "build time unknown", never crash.
[...]

PR #2611, opened 2026-08-10
fix(arenabench): stop a declared per-trial ceiling from blocking a launch

macanderson:
## What & why

Starting a match by hand from the arena UI was impossible on the reporting workspace: every launch came back with

```
Error: Claude Code (Sonnet 5, xhigh): engine.budget_usd; … engine.max_tokens
— per-trial ceilings are refused. Bound spend at the provider key …
```

naming fields the page does not render and the operator never set.

The refusal (#2411) counted a cap key's *presence* and ran before every other check in `create_match`. Nothing in ArenaBench can honour a ceiling — `Engine` and `RoleConfig` have no field for one, and `runner.py` deliberately never exports `STELLA_BUDGET` — so the key was a wall in front of a knob that did nothing. That was survivable while the operator could delete the key, and it stopped being survivable when they couldn't: the built client (`arenabench/arenabench/web/`) is generated by `npm run build` and gitignored, so an arena served from a checkout newer than the last build keeps POSTing the `budget_usd: ""` that #2602 removed from the form's *source*. Source fixed, operators not.

This PR drops a declared ceiling and runs the match uncapped rather than refusing it:

- `model.drop_cap_keys` — the companion to `declared_cap_keys`, stripping the key at the engine root and in every role table.
- `server.create_match` — strips instead of raising, and records what it ignored on the match note, so the run says out loud that it ran uncapped.
- `config._engine_from_toml` — same, as a `WARNING` on `arenabench.config` rather than a template error. A committed template that says `budget_usd` no longer fails to load; it loads, warns, and runs uncapped.

The measurement argument in #2411 is untouched — the executed run is uncapped either way, which is the whole point of that issue. What changes is that reaching that state is the server's job rather than the operator's. The text guards that keep the committed templates and the UI source free of cap keys both stay: a control that changes nothing about the run still shouldn't be offered.

Also refreshes the five committed templates and the README example, which described the old refusal in prose that is now false.

Refs #2341 — the underlying "server silently serves a stale `web/` bundle" defect that turned a fixed form into a broken arena is already filed there and is not fixed here.

## The witness

`arenabench/tests/test_web_form.py::TestACeilingCannotBlockALaunch` — feeds `create_match` the exact payload shape a stale bundle sends (`budget_usd: ""`, engine-level, and role-level) with no credential in the environment, and asserts the refusal that comes back is the credential one. On `main` it raises on the ceiling first. Checked with the sources stashed:

```
old: raised: Stella: engine.budget_usd — per-trial ceilings are refused. …
new: raised: required credentials are not set: Stella needs any of OPENROUTER_API_KEY …
```

`arenabench/tests/test_matches.py::TestPerTrialCeilingsAreForcedOff` covers the TOML side (loads, drops, warns), and 4 of its 6 cases fail on `main`.

## The gate

Python-only change under `arenabench/`, so the Rust tiers are untouched:

- [x] `uv run --extra dev pytest` — 508 passed, 6 skipped (Docker-gated)
- [x] `ruff check .` — 27 findings before and after, all pre-existing; no new ones
- [x] `scripts/check-file-size.sh`, `scripts/check-left-behind.sh` — OK
- [x] Docs updated where behavior changed (README example, five match templates, the module docstrings that stated the refusal)

## Nothing left behind

- [x] Filed: nothing new — the stale-bundle root cause is #2341, linked above.

## Ground-rule check

- [x] No new deps (ArenaBench stays stdlib-only)
- [x] No new outbound network calls — in particular this adds no provider balance/credit check; nothing in the tree has one

The alternative was to keep refusing and fix only the bundle staleness (#2341). Rejected: a knob nothing can honour should not be able to fail a launch under *any* build, and a refusal whose only remedy lives in a gitignored artifact is a class of bug that recurs. Keeping the drop visible (match note + log warning) is what preserves #2411's actual requirement — that a posture the file states and the run does not get is never silent.

---
_Generated by [Claude Code](https://claude.ai/code/session_014eD4nWfrtYVwgZon6Pe6qk)_

## Summary by Sourcery

Allow matches to run uncapped when per-trial budget or token ceilings are declared, dropping and reporting these keys instead of refusing launches or templates.

Bug Fixes:
- Ensure arena web UI launches are not blocked by stale bundles sending hidden per-trial ceiling fields, making credential checks the gating failure instead.

Enhancements:
- Introduce shared helpers to detect and strip per-trial ceiling keys from engine and role configurations across HTTP and TOML paths.
- Log warnings when templates declare per-trial ceilings so operators are informed that runs execute uncapped while preserving template loadability.
- Annotate match notes with information about ignored ceiling declarations so measurement posture differences are explicit.

Documentation:
- Update README and committed match templates to describe ceilings as dropped-with-warning rather than refused, reinforcing the "no per-trial ceiling" rule.

Tests:
- Add tests covering that web form payloads declaring ceilings no longer block launches and that stripped payloads propagate no ceiling fields onward.
- Extend template-loading tests to verify that various ceiling declarations load successfully, are dropped from engines and roles, and emit appropriate log warnings.


macanderson:
`fmt + clippy + test` is red here on a break that is on `main`, not in this diff — this PR touches only `arenabench/` Python plus prose, and the export code it fails in is byte-identical to `origin/main` (`git diff origin/main -- crates/stella-cli/src/export.rs crates/stella-cli/src/export/` is empty).

The failure is #2613: `crates/stella-cli/src/export/tests.rs:779` and `:825` still call `render_dashboard` with 5 arguments after #2606 gave it a sixth (`export.rs:436`), so `cargo test --workspace` never compiles (`E0061` ×2, and the same error again under `make cache-correctness`).

I checked whether it was a one-line repair I could just carry: it isn't. Adding the argument the three sibling call sites already pass — `&transcript::render(&Default::default(), &Default::default())` — makes it compile, and then three tests fail on the design assertions #2597 and #2606 merged into each other:

```
the_dashboard_is_monospace_and_square      — the dashboard does not declare the brand's mono stack
the_dashboard_ships_the_instrument_palette_in_both_themes — no light scheme for the OS preference
the_dashboard_spells_the_wordmark_lowercase — the masthead does not carry the wordmark
```

That is a question about what the export dashboard should render, not a mechanical fix, and #2610 and #2614 are both open against exactly it. So I am not carrying a partial repair of someone else's surface into this PR. Everything this PR owns is green — `harbor_adapter + analyzer pytest`, which runs the `arenabench` suite this change lives in, passed, as did the other 11 checks.

I will merge `main` and re-run once one of those lands.

---

Issue #2585, opened 2026-08-09
docs: the pipeline prose still describes a verdict stage and a guidance loop that no longer exist


macanderson:
## Problem

PR #2584 removed the model verdict and the distress-guidance call from the
pipeline: the evidence ladder is now terminal, and inconclusive evidence
resolves to `LadderRung::Unverified` with no model consulted. The code and its
tests were updated in that PR; the **prose was deliberately left alone**, at the
author's request, so the architectural change could be reviewed on its own.

Every one of the following now describes a pipeline that no longer exists.

## Files and what is stale in each

- `website/content/docs/inference-pipeline.mdx` — describes the verdict stage
  and the distress-triggered guidance loop as live behaviour. Both are gone.
- `AGENTS.md` — the "definition of done" section states the stage order as
  `triage → recall → research → plan → scope → execute → witness → verify →
  verdict`. There is no verdict stage. Also states the staged pipeline "has an
  independent model author the failing witness test **and** render the verdict";
  only the first half is still true.
- `docs/spec/witness-protocol.md` — §7 / §7.1 reason about when a verifier call
  is bought and when triage's `VERIFIER: no` waiver stands. The waiver logic
  survives (`verifier_waiver_stands`), but "buys the verifier" no longer means a
  model call.
- `docs/prompts/` — transcribes every role's prompt verbatim. The verifier and
  guidance prompts were deleted in #2584, so those transcriptions now describe
  constants that do not exist. (Note #2417: nothing checks this file against the
  constants, which is why it did not fail the gate.)
- `crates/stella-pipeline/README.md` — check its boundary/extension prose for
  verdict-stage references.
- `crates/stella-cli/src/settings.rs:598` — the `responsibilities` doc comment
  gives the verdict/guidance ablation as its worked example. Those keys are now
  rejected as `NotAssignable`, so the example in the doc is a config that errors.

## How to verify

```sh
rg -n 'verdict stage|model verifier|distress.guidance|heuristic_fallback|model_verdict' \
   AGENTS.md docs/ website/content/ crates/*/README.md
```

Then `make doc-links` and `make command-docs`, and re-read
`website/content/docs/inference-pipeline.mdx` end to end against
`crates/stella-pipeline/src/verify.rs::ladder_decision` — the ladder's five
terminal outcomes are the contract the page should describe.

## Constraints

- The wire vocabulary kept read-compatibility: `model_judge` / `model_verdict` /
[...]

PR #2584, opened 2026-08-09
refactor(stella-pipeline): verification makes no model calls

macanderson:
## What this changes

Verification no longer asks a model anything. The evidence ladder decides from
deterministic observations alone, and a turn it cannot settle is reported
`Unverified` — an honest "not proven", explicitly not a failure.

Two of the verifier's three responsibilities are removed outright:

| responsibility | before | after |
|---|---|---|
| `Verdict` | a model read the diff and answered PASS/FAIL | **unassignable** |
| `DistressGuidance` | a model steered a twice-failed worker | **unassignable** |
| `WitnessAuthor` | a model authored the failing test | unchanged — see below |

## Why

The escalated verdict was measured, and it was worse than neutral. Across an
89-task Terminal-Bench run it agreed with the benchmark's grader **46% of the
time**, and 17 of its false passes cost 5 tasks outright — a coin flip with the
authority to end a run.

The cost was not only wrong answers. A verdict is prose; prose fed back to a
worker is an instruction. On `fix-git` a reviewer claimed the worker had
hand-written a commit's content instead of recovering it. No measurement
supported the claim. The worker — told this was Evidence — reset `master` and
re-did the merge. **Twice.** Each round destroyed correct work, and the trial
ended on a deadline.

#2542 typed the cause so a claim would at least be *labelled* a claim. This
finishes the job: a channel whose best case is "the worker correctly disbelieves
it" is not worth its worst case.

## The structural part

`default_agent` answers `None` for both responsibilities. That is not a default
— it removes their rows from every roster:

- `Roster::default` builds rows by filtering `ModelCallRole::ALL` through
  `default_agent`, so there is no row to enable.
- `Roster::apply` rejects any configured key `is_assignable` denies, so no
  settings spelling reaches them.

The witness for this is `no_configuration_can_put_a_model_back_in_the_judgement_seat`
(`crates/stella-pipeline/src/roster/tests.rs`). It deliberately does **not**
assert "a default run buys no verdict" — that would pass equally if the call
were merely defaulted off, and "off by default" is one settings key away from
being on. It asserts unassignability, in both spellings an operator would reach
for. On `main` `default_agent(Verdict)` returns `Some(verifier)`, so it fails
there and passes here.

## What still carries the load

Untouched: the flip oracle, the tamper exclusion, the mutation/tautology guard,
diff coverage, the lint regression veto, and the worker's own `verify_done`
shadow run. **One** channel survives from the verification side back to the
[...]

Issue #2569, opened 2026-08-09
pipeline: ablating the verifier turns every run into a PASS — no_test_surface cannot tell an absent oracle from a disabled one


macanderson:
> **Update (2026-08-10, against `703976e`): the repro chain below is obsolete; the second ask is still live.**
>
> #2584 (landed via #2615) deleted `verify::heuristic_fallback` and the `unmeetable_demand`
> expression this issue quotes. Links 4 and 5 of the chain no longer exist, so a fresh
> agent following it will not find the code. What changed and what did not:
>
> **Fixed.** "No oracle" no longer produces a permissive pass through the fallback.
> `ladder_decision` resolves an unsettled turn to `LadderDecision::Unverified`, which is
> terminal and asks no model.
>
> **Also changed:** `no_test_surface` now steers *nothing at all*. It is not read by
> `ladder_decision`; its only consumers are `verify_probes.rs:209` (snapshot) and
> `replay.rs:486` (rendering). Verified by probe — two `LadderInputs` differing only in
> `no_test_surface` return the identical decision. Its doc comment
> (`verify.rs:475-482`) still describes it as "the one field that can turn a fallback FAIL
> into an abstention", describing a fallback that no longer exists. So the three states
> this issue asks to distinguish are still not distinguished — the boolean simply stopped
> having an effect rather than acquiring the right one.
>
> **Still open — the second bullet of "What to do".** The surface still reports a pass.
> `LadderDecision::Unverified` and `Unverifiable` both emit `AgentEvent::Verdict { passed:
> true }` (`pipeline.rs:2508`, `pipeline.rs:2355`). `PipelineStatus` has three variants and
> none of them is unverified (`pipeline.rs:647`), so `Some(verdict) if !verdict.passed =>
> VerificationFailed, _ => Completed` (`pipeline/resume_stage.rs:507`) lands both on
> `Completed`. From there every projection in `crates/stella-cli/src/agent/outcome.rs`
> reads a success: `pipeline_status_result` → `Ok(())` → exit 0,
> `pipeline_status_label` → `"completed"`, `pipeline_episode_outcome` →
> `EpisodeOutcome::Success`, `pipeline_session_status` → a successful session row.
>
> `CandidateScore::Unverified` is carried on `PipelineOutcome::score`, and nothing on the
> exit path branches on it. So the ladder's five honest rungs collapse to two at the
> process boundary, and a turn the engine explicitly refused to prove exits the same way
[...]

Issue #2559, opened 2026-08-09
pipeline: the witness author re-derives the repo's test conventions on every run — 20 tool calls, 0.67 USD, zero tests written


macanderson:
## Problem

The witness author re-derives this repository's test conventions from scratch on every run, using the two lowest-bandwidth tools available, with no step bound, on the critical path after execution.

Observed on execution 99 (`~/.stella/exports/2026-08-09-21-09-04`, goal: "design a system … place the spec in docs/designs … create a branch and a pull request"). The worker finished at 21:05:06 having written the spec, committed, pushed and opened PR #2554. Eight seconds later `witness_author` (opus-5) started and spent:

- **10 model calls**
- **20 tool calls — 10 `glob`, 10 `read_file`, nothing else**
- **2m54s model time, $0.6733** (12.7% of the run's $5.2879)
- **0 `create_witness_test` calls**

It read, in order: `docs/**/*.md`, `bench/**/*.py`, `*`, `Cargo.toml`, `scripts/check-design-refs.sh`, `docs/manifest.json`, `bench/telemetry_store/conftest.py`, `docs/design/pipeline-journey.md`, `scripts/check-doc-links.py`, `docs/README.md`, `Makefile`, `bench/telemetry_store/tests/test_ingest.py`, `AGENTS.md`.

Every one of those is a question about *this repository's conventions* — where tests live, what they look like, how they are invoked. None is a question about the change under review. The run was cancelled by the operator four minutes in, with the PR already merged.

## Why it is structural

1. **The system prompt instructs discovery.** `crates/stella-pipeline/src/witness.rs:745` — "Explore with `read_file` and `glob` (names only) to find the test directories and conventions".
2. **The tool surface makes discovery maximally expensive.** `crates/stella-cli/src/candidate_ws/witness_tools.rs:178` — `schemas()` filters to exactly `read_file`, `glob`, `create_witness_test`. Correct for safety, but there is no `grep`, no `project_overview`, no `gather_context`, so conventions must be brute-forced one file at a time.
3. **The author is blind to the diff.** `witness_prompt()` (`crates/stella-pipeline/src/witness.rs:836`) takes `goal`, `recall`, `repo_structure`, `available_runners`, `workspace_root` — no diff. Deliberate (L-E6 / `doc:witness-protocol §7.3`): an author that reads the implementation writes a test restating it. But it means the author cannot know *which subtree* the change is in, so it globs the whole repo trying to locate the subject from goal text alone.
[...]

PR #2615, opened 2026-08-10
revert #2588 and land #2584 in its place

### LINKED ISSUE #2612: pipeline: #2588 made verification model-free but dropped five deterministic guards — a no-op turn passes again, and a bare flip needs no corroboration
## Problem

#2588 achieved its goal — no model participates in verification any more, and that is the right architecture. But in collapsing `ladder_decision` to three lines it also dropped **five deterministic guards**, every one of which existed to *withhold* a pass that a bare flip would otherwise grant. Verification is now model-free and materially weaker than it was.

All of this is read off `origin/main` after the merge, not inferred.

## 1. A no-op turn reports `passed: true` again (the #2129 regression)

`LadderDecision::NothingAttempted` no longer exists. `ladder_decision` (`crates/stella-pipeline/src/verify.rs:555`) is now:

```rust
if inputs.touched_tests_passed == Some(false) { return LadderDecision::Revise; }
if inputs.flip_achieved || inputs.verify_done_flip { return LadderDecision::SubmitFast; }
LadderDecision::Unverifiable
```

A turn that dispatched **zero** mutating calls matches none of the first two, so it lands on `Unverifiable`, whose arm emits `AgentEvent::Verdict { passed: true }` (`pipeline.rs:1866`).

That is the exact state the rung was introduced to separate. From the module docs that are still in the tree: *"Eleven Terminal-Bench 2.1 trials ended that way: `glm-5.2` reasoned for a while, called no tool at all, and the run declared success on a task it had not touched. Every one scored 0.0."*

The `Unverifiable` arm's own comment still says **"The arm above is what makes this defensible"** — citing the `NothingAttempted` arm, which is gone. The comment's stated precondition is now false.

`LadderInputs::nothing_was_attempted()` and `LadderRung::NothingAttempted` both still exist and are now unreachable from the live path.

## 2. A single flip grants a deterministic pass with no corroboration

`if inputs.flip_achieved || inputs.verify_done_flip { return SubmitFast; }` has no veto conjuncts, and `pipeline.rs` no longer computes the inputs that used to supply them. Confirmed by grep over `origin/main:crates/stella-pipeline/src/pipeline.rs`: no `lint_delta` call, no `new_diag_errors` assignment, no `oracle.confirm`, no `witness_tautological` read.

So all four of these are now bypassed:

| guard | what it withheld | status |
|---|---|---|
| **#859** confirmation re-run | a flip whose second run failed (flaky) | never runs — `oracle.confirm` uncalled |
| **#861** lint regression veto | a flip that introduced a fresh type error | never runs — `lint_delta` uncalled |
| **#870** tautology / mutation guard | a witness green under every mutation of the change | never read |
| **#1291** diff coverage | a test that passed without executing the changed lines | evidence text only |

On #1291 specifically: `diff_coverage` survives at `pipeline.rs:1825` but only as an argument to `deterministic_pass_evidence` — it gates nothing, and the arm now calls `score_from_verification(true, None)` unconditionally, so the "unmeasured overlap scores `Unverified`" downgrade is gone too. `PipelineConfig::require_diff_coverage` is now inert.

A witness the worker wrote for itself, green because it asserts nothing, passing once on a flaky runner, alongside a fresh type error, now yields `deterministic: true`.

## 3. "Could not look" and "looked, not enough" are one rung again

Both now return `Unverifiable`. `LadderRung`'s own doc comment calls this distinction load-bearing, and the enum still carries the retired `ModelVerdict` / `HeuristicFallback` rungs describing a model's involvement, for a pipeline that makes no model calls.

## Suggested repair

Restore the guards inside the new model-free shape — none of them needs a model:

1. Reinstate a `NothingAttempted` decision above the abstention, failing closed. It reads `mutating_actions`, which is the pipeline's record of its own dispatch, so it can never be blind.
2. Restore the veto conjuncts on the `SubmitFast` arm — confirmation run, `new_diag_errors`, `witness_tautological`, `diff_coverage` — and re-wire `lint_delta` / `oracle.confirm` in `pipeline.rs`.
3. Give inconclusive-but-observed evidence a rung of its own, distinct from blindness.
4. Retire `ModelVerdict` / `HeuristicFallback` from the emittable vocabulary (keep them parseable for stored streams), and delete the now-unreachable `heuristic_fallback` / `parse_verifier_response` / prompt builders in `verify.rs` — they are `pub`, so no compiler warning will find them.

Much of 2–4 exists as reviewed, tested code in PR #2584, which was superseded by #2588 and is being closed; those commits are the cheapest source for the veto conjuncts and the extra rung.

## Definition of done

`ladder_decision` grants `SubmitFast` only on a *corroborated* flip; a zero-dispatch turn does not report `passed: true`; the guards each have a witness test again; `make gate` green.

Refs #2588, #2584, #2569, #2129

macanderson:
## What & why

#2588 and #2584 were two competing implementations of the same idea — model-free
verification. #2588 was merged (`aa2e9dc`) and #2584 was closed 39 seconds later
as superseded. This swaps them: #2588 comes out of `main`, #2584 goes in.

#2612 is the case for doing it, filed against `main` after the merge: #2588
reached the model-free architecture but collapsed `ladder_decision` to three
lines and dropped five deterministic guards on the way — the `NothingAttempted`
rung (so a zero-dispatch turn reports `passed: true` again, the #2129
regression), the #859 confirmation re-run, the #861 lint-regression veto, the
#870 tautology/mutation guard, and #1291 diff coverage. That issue's own
suggested repair names #2584's commits as the cheapest source for the fix. This
PR is that repair.

Closes #2612

### The commits

1. **`Revert "fix(pipeline): make verification model-free (#2588)"`** — a plain
   revert of the squash-merge. `aa2e9dc` has a single parent (`12a509f`), so
   there is no mainline to pick. Verified exact: the reverted tree differs from
   `12a509f` only in `Cargo.toml`, `Cargo.lock` and
   `packaging/homebrew/stella.rb` — i.e. only in the 0.8.21 release sync
   (#2609), which is deliberately preserved.
2. **`refactor(stella-pipeline): verification makes no model calls (#2584)`** —
   merges `fix/deterministic-verifier-no-model`. Conflict-free, because #2584's
   merge base with `main` is exactly `12a509f`, the commit the revert restores.
3. **`fix(stella-pipeline): repair the doc links #2584 left pointing at deleted items`**
4. **`fix: unbreak the merged tree`** — see "Anything reviewers should know?".

### The #2584 head is one commit newer than the PR page shows

GitHub records #2584's head as `8afdd8a`. The branch has one commit beyond that,
`2390274 "take main's PROOF panel wholesale and teach it VerificationUnproven"`,
pushed after the PR was closed. That commit matters: `8afdd8a` clobbers #2568's
`ProofState::standing` refactor with the branch's older `is_empty`/`is_notable`
copy, and `2390274` is the repair. This PR merges the branch tip, so the PROOF
panel survives the swap.

Refs #2584, #2588, #2569, #2559, #2542, #2585, #2616

## The witness

- [x] No witness needed for the swap itself (revert + restore of an
      already-reviewed branch) — because the behavior being restored is #2584's,
      and its witness travels with it:
      `no_configuration_can_put_a_model_back_in_the_judgement_seat`
      (`crates/stella-pipeline/src/roster/tests.rs`), which asserts `Verdict` and
      `DistressGuidance` are *unassignable* rather than merely defaulted off. It
      fails on `main` today (`default_agent(Verdict)` returns `Some(verifier)`)
      and passes here. The guards #2612 reports as missing likewise return with
      their original witness tests, restored by commit 1.

Commits 3 and 4 are repairs to make the tree build; `make doc-warnings`,
`cargo clippy --all-targets` and `cargo test` are the checks that fail without
them.

## The gate

Full `make gate` run locally on the merged tree. Results reported in a comment
on this PR rather than asserted here.

## ⚠️ `main` is red right now, independently of this PR

Worth separating from everything above, because it is not caused by this swap
and does not go away if this PR is closed.

#2606 gave `render_dashboard` a sixth `&Transcript` parameter and updated three
of its five call sites in `crates/stella-cli/src/export/tests.rs`. The two it
missed (`the_dashboard_is_monospace_and_square`,
`the_dashboard_spells_the_wordmark_lowercase`) still pass five arguments, so
`cargo clippy --workspace --all-targets` and `cargo test --workspace` fail with
`E0061` on `main` as of `3d87aed`.

Evidence, not inference: `crates/stella-cli/src/export.rs` and
`crates/stella-cli/src/export/tests.rs` are **byte-identical** between `main`
and this branch (`git rev-parse 3d87aed:<path>` equals `git rev-parse HEAD:<path>`
for both) — #2588 never touched either file, and #2584 changed exactly one line
in `stella-cli`, in `agent/engine.rs`. The failure therefore reproduces on
`main` verbatim.

Commit 4 fixes it here, using the idiom the file's other three call sites
already use, because this branch cannot reach green while it stands. Note that
#2588's description claimed it had "restor[ed] a green integration tree" — its
diff contains no change to either file.

## Nothing left behind

- **#2616** (filed) — the `require_independent_verifier` residue: the still-parsed
  `pipeline_require_independent_verifier` setting that now wires to nothing, the
  unreachable `RunError::VerifierNotIndependent`, and `OutcomeWeights::judged`,
  which `reward::label` no longer reads.
- **#2585** (existing, linked not duplicated) — the pipeline docs still describe
  a verdict stage and a guidance loop. The revert restores every
  `website/content/docs/**` and `docs/prompts/**` page #2588 had updated, and
  #2584 deferred its own docs pass explicitly, so `docs/prompts/verdict.md` and
  `distress-guidance.md` document calls that cannot happen.

## Ground-rule check

- [x] Wire types changed (`LadderRung::Unverified`, `ProofStep::VerificationUnproven`) round-trip; `docs/wire/` regenerated on the #2584 branch and `wire-schema` passes

## Deleted tests

`check-deleted-tests` compares against `main`, so it sees two sets at once:
every test #2588 deleted coming *back* (the revert), and the set #2584 deletes
going away. The authoritative list of the latter — renames and removals, with
reasons — is in #2584's description and is not re-pasted here; the six renames
are `golden_verifier_escalation_without_a_test_command`,
`the_four_tiers_score_at_their_stated_weights`,
`a_verdict_about_the_pipeline_is_not_a_verdict_about_the_work`,
`an_uncorroborated_verifier_pass_is_stamped_as_an_abstention_not_a_judgement`,
`flip_and_green_but_over_diff_budget_escalates_to_verifier`, and
`distress_guidance_on_the_worker_is_not_an_independence_loss`.

**Two of the three breakages in commit 4 are #2584's own, not merge artifacts.**
The affected files are byte-identical between this branch and #2584's tip, so
they would fail on that branch too:

- `verification_hardening.rs` kept the doc comment of the
  `verdict_degraded_facts` module it deleted; the comment reattached to the next
  test and tripped `clippy::empty_line_after_outer_attr`.
- `AgentEngineConfig::pipeline_require_independent_verifier_on` lost its only
  caller when the `PipelineConfig` field it fed was removed.

Same for the six unresolved intra-doc links in commit 3. In every case #2584's
stated verification was `make guards`, which stops below the rustdoc, clippy and
test tiers — which is also why these were invisible until the full gate ran.

**This is a revert, not a reset.** `main` already carries #2609 (the 0.8.21
version sync) on top of #2588, and resetting `main` back to `12a509f` would drop
it. Reverting keeps it and leaves `main`'s history unrewritten.


macanderson:
## Gate results on the merged tree

Run locally on `c388240`.

| step | result |
|---|---|
| all toolchain-free guards (`no-scratch` … `typed-errors`, `wire-schema`) | ✅ pass |
| `doc-warnings` (rustdoc `-D warnings`) | ✅ pass (red before commit 3) |
| `format-check` | ✅ pass |
| `lint` (`clippy --workspace --all-targets -D warnings`) | ✅ pass (red before commit 4) |
| `test --workspace` | ❌ **4 failed, 1690 passed** — see below |

Notably `file-size` passes with no baseline change, and `god-files`,
`gate-parity`, `left-behind`, `role-names` and `module-reachability` are all
green.

### The four failures — none is caused by this PR

**One flake.** `daemon::tests::a_run_that_ignores_term_is_killed_after_the_grace_period`
— *"the child was killed after 81.4ms, before its 8s to shut down cleanly"*. A
wall-clock assumption that breaks under parallel test load in a container. Passes
in isolation:

```
cargo test -p stella-cli --bin stella daemon::tests::a_run_that_ignores_term_is_killed_after_the_grace_period
```

**Three pre-existing `main` failures**, now filed as **#2617**:

- `export::tests::the_dashboard_ships_the_instrument_palette_in_both_themes`
- `export::tests::the_dashboard_is_monospace_and_square`
- `export::tests::the_dashboard_spells_the_wordmark_lowercase`

These are #2606's, not this PR's, and I ran the control rather than asserting it.
On a clean `origin/main` worktree at `3d87aed`, with **only** the two-line arity
fix applied and no assertion changes:

```
test result: FAILED. 23 passed; 3 failed; 0 ignored; 0 measured; 1625 filtered out
failures:
    export::tests::the_dashboard_is_monospace_and_square
    export::tests::the_dashboard_ships_the_instrument_palette_in_both_themes
    export::tests::the_dashboard_spells_the_wordmark_lowercase
```

The same three, in the same state, off `main`. Bisecting the renderer for
`@media (prefers-color-scheme: light)` puts the drop at `b7ea9f1` (#2606):
present at `63765fd` (#2597) and `898a8b0`, absent from `b7ea9f1` onward while
the test kept requiring it.

**This PR deliberately does not fix those three.** The renderer is the side that
regressed against #2597's contract, and relaxing three design assertions to green
a revert PR is the "golden blessed without looking" failure AGENTS.md warns about.
#2617 carries the full bisect, the repro, and the definition of done.

So: this branch is green on everything except a breakage it inherited and a
timing flake. `main` itself cannot build `stella-cli`'s test binary at all right
now, which is why the three were invisible until commit 4 fixed the arity.

---

Issue #2626, opened 2026-08-10
design-system: the export dashboard's look has no single source of truth, and three PRs reconciled it in opposite directions


macanderson:
## Problem

`crates/stella-cli/src/export.rs` and `crates/stella-cli/src/export/tests.rs`
have now been reconciled **in opposite directions three times in two days**,
and `main` shipped red after two of them. The churn is not carelessness by any
one author — it is that the export dashboard's appearance has no single source
of truth, so each PR reasonably picked a different one.

The sequence, all on `main`:

| PR | what it did to the dashboard's look |
| --- | --- |
| #2597 | put it on the Observatory's instrument palette; added `--mono`, `--radius: 0`, `<span class="wordmark">`; added it to `design_token_parity.rs` |
| #2606 | restored the pre-design-system stylesheet (`--ink/--dim/--faint/--stella/--pass/--fail`, `--stella: #B57A00`, rounded corners, `font: 13px/1.55 ui-monospace`) and left two of #2597's tests at the old `render_dashboard` arity |
| #2610 | repaired `export/tests.rs` **against #2606's markup** — asserting `font: 13px/1.55 ui-monospace`, `<h1>stella session —`, `border-radius: 3px` / `2px` |
| #2614 | restored #2597's **source** (the `--mono` token, the wordmark span, `--radius: 0`) and dropped `/export` from the parity matrix; did not move #2610's tests back |

Net state of `main` at 703976eba: `export.rs` renders the design system,
`export/tests.rs` asserts the opposite of it, and nothing holds the two
together. Reproduce:

```sh
git checkout 703976eba
cargo test -p stella-cli --bin stella export::tests   # exits 101
#   the_dashboard_is_monospace_with_compact_corners ... FAILED
#   the_dashboard_spells_the_wordmark_lowercase     ... FAILED
```

## Why it stayed invisible

#2606's arity mistake meant the whole `stella-cli` **bin test target did not
compile**, and a target that does not build cannot report the assertion
failures underneath it. `cargo test -p stella-cli --test design_token_parity`
(a separate target) did build, and there the parity test was failing on
`marker :root[data-theme="light"] {{ not found` — which reads as a broken test
rather than a reverted design, so it was plausible to "fix" by removing the row.

Two compounding factors worth naming:

- A compile error in one target masks assertion failures in that target
  entirely. Any conclusion of the form "the tests pass" is unsound until the
  target compiles.
[...]

PR #2646, opened 2026-08-10
fix(stella-cli): unbreak main — drop the merge-scarred second radius loop

macanderson:
## What & why

`main` does not compile. Since #2628 merged at `2026-08-10T04:52Z`,
`crates/stella-cli/src/export/tests.rs` has held a loop whose header binds
`rounded` and whose body reads `literal`:

```rust
for rounded in ["border-radius: 8px", "border-radius: 6px"] {
    assert!(
        !html.contains(literal),
        "`{literal}` bypasses `--radius`: set the token, not the rule"
    );
}
```

`error[E0425]: cannot find value 'literal' in this scope`, twice.

### How it got there

#2624 and #2628 were two concurrent "unbreak main" PRs that each added a loop
asserting no hardcoded `border-radius` survives the `--radius: 0` token. They
chose different loop variables over different lists:

| PR | binds | list |
|---|---|---|
| #2624 | `literal` | `8px`, `6px`, `3px`, `2px` |
| #2628 | `rounded` | `8px`, `6px`, `3px`, `2px`, `0 6px 6px 0` |

The merge kept **both** loops, and fused #2628's header onto #2624's body. Git
had no conflict to report — this is the rename-vs-edit shape where each side
textually contains what the other lacks, so a clean three-way merge composes
into code that compiles on neither branch.

### Why it is expensive out of proportion to its size

The break is a **compile error in a test target**, not a failing assertion, so
it does not fail one check — it stops three, each of which has to build
`stella-cli`:

- `cargo clippy -D warnings`
- `prompt-cache correctness (golden fixtures)`
- `cargo test`

All three halt at the same two lines in run
[31357554730](https://github.com/macanderson/stella/actions/runs/31357554730).
And because nothing under the error compiles, every assertion in the CLI's test
target is currently unevaluated — a green run after this lands is the first
real signal since 04:52Z.

Every open PR inherits it. **#2644**, the `0.8.23` version sync, is red for
this reason and nothing else; its diff is version numbers.

## The fix

Delete the residual loop. It costs no coverage: both literals it names are the
first two entries of the surviving five-element list, which also covers `3px`,
`2px` and `0 6px 6px 0`. #2628's list is a strict superset of #2624's, and its
failure message is the one that survives.

Six deleted lines, no additions.

## Witness

`main` does not compile — the strongest fail-on-old/pass-on-new there is. The
fail half is observed in CI run 31357554730 rather than locally: a
Terminal-Bench match is occupying this machine, and per the standing rule the
workspace build is CI's job. **CI on this branch is the pass-on-new half — do
not merge this until it is green**, since I have not run the suite locally.

## Deleted tests (for `check-deleted-tests`)

None. No `#[test]` is removed; the deletion is a loop inside
`the_dashboard_is_monospace_and_square`, which stays and keeps asserting the
same thing over a longer list.

## Follow-ups filed

See the issues linked in the thread below — a duplicated-loop merge scar that
no guard catches, and a latent `evidence_is_blind` gap found while auditing
the other open PR.

## Summary by Sourcery

Tests:
- Delete the merge-introduced second border-radius loop in `the_dashboard_is_monospace_and_square`, retaining coverage via the existing superset loop.

### DIFF
--- a/crates/stella-cli/src/export/tests.rs
+++ b/crates/stella-cli/src/export/tests.rs
@@ -777,12 +777,6 @@ fn the_dashboard_is_monospace_and_square() {
              where a hairline says \"boundary\", and this page is boundaries"
         );
     }
-    for rounded in ["border-radius: 8px", "border-radius: 6px"] {
-        assert!(
-            !html.contains(literal),
-            "`{literal}` bypasses `--radius`: set the token, not the rule"
-        );
-    }
 }
 
 /// The wordmark is lowercase, everywhere the report says it.


macanderson:
Duplicate of #2645, which is the same six-line deletion and was opened four minutes earlier (05:21:33Z vs 05:25:39Z here). The diffs are byte-identical — both leave `crates/stella-cli/src/export/tests.rs` at blob `a0c406ba1`.

**Merge #2645, not this one.** Its CI also started earlier (run 31358347052 vs 31358560572 here), so it settles first, and whichever lands second becomes an empty diff that fails the `PR carries a diff` check.

Closing in favour of #2645. The analysis in this description — the rename-vs-edit merge mechanism, the three CI steps one compile error takes down, and how it reached `main` ahead of its own required check — is recorded on #2626 as a comment so it survives this PR.

Issue #1291, opened 2026-08-03
Check that a passing test actually ran the lines that changed


macanderson:
## What's going on

When Stella changes code, something has to prove the change works. Today the proof is "a test passed". What we do **not** check is whether that test ever ran the lines that changed.

A test can pass without touching the new code at all. Then "the test passed" is a coincidence being reported as evidence.

## Why it matters

This is the difference between a check and the appearance of one. If a model writes a change and a test passes without exercising it, we record a verified result for work nobody verified. The rest of the verification system is built on that result being trustworthy.

## What to do

1. Pick how we measure which lines a test ran. Most languages have a standard coverage tool; the question is which ones we support and what we do for the ones we cannot measure.
2. Compare two lists: the lines the change touched, and the lines the test ran.
3. If they do not overlap, say so plainly in the result. **Do not turn it into a failure** — "we could not confirm this" is a different answer from "this is wrong", and the system already distinguishes the two. It should land as "unproven", not "failed".
4. For languages with no coverage tool available, the honest answer is also "unproven" rather than silently passing.

## Done when

- Where coverage tooling exists, a result says whether the test ran the changed lines
- No overlap produces an honest "unproven", never a false pass and never a false failure
- Languages we cannot measure are handled explicitly rather than by accident

## Background

#1211 §1, `ROADMAP.md` §1. The original wording: "a flip whose test never touches the diff is a coincidence, not evidence."

macanderson:
Reopening — closed 2026-08-04 with no commit, no PR reference and no comment, but nothing shipped.

Verified against `origin/main` (`0f7833ae`): there is no diff-coverage measurement anywhere in the workspace. `git grep -rniE 'diff_coverage|changed_lines_covered|llvm-cov|tarpaulin|coverage_report'` over all `*.rs` and `*.toml` returns exactly one hit — `stella-core/src/scoreboard/tests.rs:164`, an unrelated test about rated-share reporting.

None of the three "Done when" clauses are met. No result reports whether the test ran the changed lines; there is no no-overlap path landing as "unproven"; and languages without coverage tooling are not handled explicitly because nothing is handled at all.

Corroborating signal: the sibling issue #1292 (affected-test selection as verification evidence) is still open. Diff coverage and impacted-test selection are the two remaining items from the verification roadmap — closing one while its sibling stays open is internally inconsistent, and the code agrees with #1292.

## Why this was closed by mistake

[...]

Issue #870, opened 2026-07-30
Verification §1 (Phase 4): mutation-check the witness — apply trivial mutations and downgrade a tautological witness that stays green


macanderson:
Part of the verification pipeline epic (#858). **ROADMAP.md §1, Phase 4.**

## Problem

A fail→pass flip proves the test *reacted* to the change, but a tautological witness — one that passes under any change to the target lines — flips green without constraining anything. The assertion-density heuristic (#863, Phase 2) catches the static cases. The mutation check catches the dynamic ones: a witness that stays green when the change is subtly broken is not testing the right thing.

## Proposal

After a flip, apply 1–3 trivial mutations to the changed lines in a scratch candidate workspace:
- Negate a condition, off-by-one a bound, flip a comparison.
- Re-run the witness against each mutant.
- If the witness stays green under **every** mutant → downgrade the evidence from `DeterministicPass` toward the judge path: the witness is likely tautological.

**Bounded cost:** one extra test run per mutant, only on the winning candidate.

## Where

- `stella-pipeline/src/verify.rs` — the evidence ladder, after the flip
- `stella-pipeline` — scratch workspace isolation (reuse the candidate worktree infra)

## Rationale

The strongest witness-quality check — it catches tautologies that static analysis can't. Degrades gracefully: if mutation isn't available (language unsupported, workspace constraints), the flip stands.

## Acceptance

- A witness that passes under all mutants is downgraded from `DeterministicPass`.
- A witness that fails under at least one mutant retains `DeterministicPass`.
- The mutation runs only on the winning candidate, bounded to 1–3 mutants.
- Tested: a tautological witness (`assert!(true)`) that flips is caught and downgraded.

tomatotomata:
I was thinking of keeping this first pass inside the existing stella-pipeline evidence ladder: add a bounded mutation runner for the winning witness, downgrade only when every selected mutation still passes, and leave unsupported-language or unavailable-mutation cases at the current verdict. I would add a tautological-witness regression plus a witness that fails one mutation, without changing candidate selection or judge escalation. Does that boundary match the roadmap, or should the first PR target a smaller seam?

macanderson:
@tomatotomata ummm welcome?

macanderson:
[...]

Issue #861, opened 2026-07-30
Verification §3 (Phase 1): regression veto — a diagnostics delta (new errors/warnings) blocks fast-submit even when the flip holds


macanderson:
Part of the verification pipeline epic (#TBD). **ROADMAP.md §3, Phase 1.**

## Problem

Lint and typecheck are rightly excluded from the flip oracle (a lint pass is not a deterministic test). But today a flipped witness plus a fresh type error in an untested module can still fast-submit — the oracle credits the `DeterministicPass` and ships, even though the candidate introduced a regression the diagnostics *can* see.

## Proposal

Diff `DiagnosticRunner` results before/after the candidate:
- **New errors** block fast-submit even when the flip holds — this is exactly the inconclusive case the judge exists for.
- **New warnings** block fast-submit when configured (default: errors only; warnings opt-in).
- The diagnostics delta is also surfaced as structured judge evidence (ties into §4 structured judge evidence).

## Where

- `stella-pipeline/src/verify.rs` — the `ladder_decision` fast-submit path
- `stella-tools` `DiagnosticRunner` — already produces before/after diagnostics; wire the delta into the ladder

## Rationale

This is the "secondary deterministic evidence" the ROADMAP describes: lint/typecheck can't *verify* a fix, but they can *veto* a fast-submit when they see a regression. Pure-logic change, no new tooling — `DiagnosticRunner` already runs.

## Acceptance

- A candidate that flips the witness but introduces a new compile error does not fast-submit; it routes to the judge.
- The veto is configurable: errors block by default, warnings are opt-in.
- Tested: a candidate with a green flip + a fresh type error is routed to the judge, not fast-submitted.

PR #2647, opened 2026-08-10
fix(stella-pipeline): let a verify_done receipt grant the deterministic pass

macanderson:
## Problem

#2588 stated completion authority as:

> a configured same-command fail-to-pass flip **or a typed built-in `verify_done`
> request replayed against the sealed final tree**

The second half is not wired. On `origin/main` today:

- `verify_done_flip` is computed — `pipeline/evidence.rs`, from
  `state.signals.verify_done_confirmations > 0`
- it is read in **exactly one place**: `LadderInputs::verifier_pass_stands_alone`,
  a predicate about a *model verifier's* pass, which no longer exists
- `ladder_decision` never reads it
- `verify_done` appears **0 times** in `pipeline.rs`, and `verify_done_request`
  appears nowhere in `crates/stella-pipeline/src`

So the receipt is computed, carried onto the snapshot, and discarded.

## Why it matters

This is not a cosmetic gap, because of what else changed around it. With no
`--test-command` the flip oracle never arms, and nothing authors a witness any
more, so `flip_achieved` **can never be true**. `verify_done` was the only other
route to a receipt.

Net effect on `main`: **`SubmitFast` is unreachable for any task that does not
ship its own test command.** On Terminal-Bench that is all of them — every run
ends `Unverified`, and the deterministic pass the model-free design was built
around cannot be earned.

`verify_done` is precisely the mechanism the design leans on in place of the
retired witness author: a pinned baseline, a shadow worktree, the test files
copied in, red there and green here. It has to be able to grant the pass it
earns.

## The fix

Two changes, and both are needed, because the abstention sits *above* the credit:

1. **Rung 4 accepts either receipt.** A configured-command flip must still leave
   the suite it belongs to green; a `verify_done` receipt has no companion suite
   — it ran both trees itself — so demanding one would reject the very evidence
   this arm exists to credit. A *red* suite never reaches here either way: rung 1
   already returned `Revise`.

   Every refutation still applies to both receipts: diff budget, the #861
   diagnostics veto, the #870 mutation audit, #1291 coverage. This widens what
   counts as a receipt, not what counts as proof.

2. **`evidence_is_blind` counts the receipt as a channel.** It was written before
   `verify_done_flip` existed and never joined it. Because rung 3 runs before
   rung 4, a turn whose *only* observation was that receipt read as "nothing
   could look" and was swallowed by the abstention before it could be credited.

## Witness

`a_verify_done_receipt_alone_submits_fast` — no configured command, no suite
result, a `verify_done` receipt and a visible diff.

`a_verify_done_receipt_is_still_subject_to_the_refutations` — the same receipt
over budget, beside a fresh diagnostic error, and beside a red suite.

Both **fail on `main`'s `verify.rs`** with `Unverified` where `SubmitFast` is
expected, and pass here. Checked the artisanal way, by swapping
`git show origin/main:crates/stella-pipeline/src/verify.rs` into place and
re-running.

## Verification

`stella-pipeline`, `stella-protocol`, `stella-tui` — build, test and clippy all
exit `0`.

`cargo build --workspace` does **not** pass, and not because of this change:
`crates/stella-cli/src/export/tests.rs` on `main` has a `for rounded in [...]`
loop whose body still references `literal`, a half-applied rename. This branch
touches no `stella-cli` file (`git diff origin/main --stat -- crates/stella-cli`
is empty). #2645 and #2646 are both already open to fix it.

Refs #2588, #2612, #2569

## Summary by Sourcery

Allow a verify_done receipt to grant a deterministic SubmitFast decision and ensure it is treated as observable evidence rather than ignored.

Bug Fixes:
- Treat verify_done_flip as an evidence channel so verify_done-only turns are no longer classified as blind and blocked before credit can be given.
- Allow either a configured flip or a verify_done receipt to satisfy the deterministic pass rung, restoring SubmitFast for tasks without a test command.

Tests:
- Add tests confirming that a lone verify_done receipt can submit fast and that it remains subject to all existing refutation rules (budget, diagnostics, red suites).

### DIFF
--- a/crates/stella-pipeline/src/verify.rs
+++ b/crates/stella-pipeline/src/verify.rs
@@ -515,6 +515,14 @@ impl LadderInputs {
     /// evidence, so neither can be blind. Only the total absence qualifies.
     pub fn evidence_is_blind(&self) -> bool {
         !self.flip_achieved
+            // A `verify_done` receipt is an evidence channel like any other,
+            // and omitting it here is a latent bug with a live trigger: this
+            // predicate is checked at rung 3, *above* the credit at rung 4, so
+            // a turn whose only observation was that receipt reads as "nothing
+            // could look" and the abstention swallows it before it can be
+            // credited. The field was added after this predicate was written
+            // and never joined it.
+            && !self.verify_done_flip
             && self.touched_tests_passed.is_none()
             && !self.diff_available
             && self.file_change_events == 0
@@ -664,8 +672,29 @@ pub fn ladder_decision(inputs: &LadderInputs) -> LadderDecision {
     //    opted-in, new warnings) drop this rung through to escalation. Lint
     //    stays excluded from the oracle — it can veto a submit, never
     //    verify one.
-    if inputs.flip_achieved
-        && inputs.touched_tests_passed == Some(true)
+    //
+    //    **Either receipt counts.** `flip_achieved` is the configured
+    //    command's own fail→pass; `verify_done_flip` is the worker's
+    //    `verify_done` run — a pinned baseline, a shadow worktree, the test
+    //    files copied in, red there and green here. Both are deterministic
+    //    observations of the same shape, and #2588 said so in as many words
+    //    ("a configured same-command fail-to-pass flip **or** a typed built-in
+    //    `verify_done` request replayed against the sealed final tree").
+    //
+    //    It stopped being read when the veto conjuncts came back, and the
+    //    consequence is not small: with no `--test-command` the oracle never
+    //    arms, and nothing authors a witness any more, so `flip_achieved` can
+    //    never be true — which made this rung unreachable for every task that
+    //    does not ship its own test command. `verify_done` is the replacement
+    //    the model-free design leans on; it has to be able to grant the pass
+    //    it earns.
+    if (inputs.flip_achieved || inputs.verify_done_flip)
+        // A configured-command flip must also leave the suite it belongs to
+        // green. A `verify_done` receipt has no such companion — it ran both
+        // trees itself — so demanding one would reject the very evidence this
+        // arm exists to credit. A *red* suite never reaches here either way:
+        // rung 1 returned `Revise`.
+        && (!inputs.flip_achieved || inputs.touched_tests_passed == Some(true))
         && inputs.diff_lines <= inputs.diff_budget
         && inputs.new_diag_errors == 0
         && (!inputs.veto_warnings || inputs.new_diag_warnings == 0)
--- a/crates/stella-pipeline/src/verify/tests.rs
+++ b/crates/stella-pipeline/src/verify/tests.rs
@@ -229,6 +229,96 @@ fn full_deterministic_pass_submits_fast() {
     assert_eq!(decision, LadderDecision::SubmitFast);
 }
 
+/// **The witness.** A `verify_done` receipt grants the deterministic pass on
+/// its own, with no configured test command anywhere in sight.
+///
+/// This is the shape the model-free pipeline actually runs in. With no
+/// `--test-command` the flip oracle never arms, and nothing authors a witness
+/// any more, so `flip_achieved` can never be true — which left this rung
+/// unreachable for every task that does not ship its own test command, and on
+/// Terminal-Bench that is all of them. `verify_done` is the replacement the
+/// design leans on, so it has to be able to grant the pass it earns.
+///
+/// Two things had to be true for it to work, and both were false:
+/// the credit at rung 4 has to *read* the field, and the abstention at rung 3
+/// has to stop swallowing the turn first — `evidence_is_blind` did not count
+/// the receipt as a channel, so a `verify_done`-only turn read as "nothing
+/// could look".
+#[test]
+fn a_verify_done_receipt_alone_submits_fast() {
+    let inputs = LadderInputs {
+        // No configured command: the oracle never armed and no suite ran.
+        flip_achieved: false,
+        touched_tests_passed: None,
+        // The worker's own shadow-worktree run: red on the pinned baseline,
+        // green on the change.
+        verify_done_flip: true,
+        no_test_surface: true,
+        diff_lines: 12,
+        diff_budget: 400,
+        diff_available: true,
+        file_change_events: 1,
+        mutating_actions: 3,
+        ..Default::default()
+    };
+    assert!(
+        !inputs.evidence_is_blind(),
+        "a verify_done receipt is an observation; a turn carrying one is not blind"
+    );
+    assert_eq!(
+        ladder_decision(&inputs),
+        LadderDecision::SubmitFast,
+        "a deterministic fail→pass receipt is completion authority whether the oracle \
+         or the worker's verify_done produced it"
+    );
+}
+
+/// The receipt is credited, not waved through: the refutations that apply to a
+/// configured flip apply to it too.
+#[test]
+fn a_verify_done_receipt_is_still_subject_to_the_refutations() {
+    let sound = LadderInputs {
+        verify_done_flip: true,
+        diff_lines: 12,
+        diff_budget: 400,
+        diff_available: true,
+        file_change_events: 1,
+        mutating_actions: 3,
+        ..Default::default()
+    };
+    assert_eq!(ladder_decision(&sound), LadderDecision::SubmitFast);
+
+    let over_budget = LadderInputs {
+        diff_lines: 5_000,
+        ..sound
+    };
+    assert_eq!(
+        ladder_decision(&over_budget),
+        LadderDecision::Unverified,
+        "a receipt does not account for a change far larger than it"
+    );
+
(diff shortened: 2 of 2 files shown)


macanderson:
Filed #2649 independently a few minutes before this, from the other end — auditing #2636 (now closed as superseded by #2615) and finding that its one surviving observation, that `LadderInputs::evidence_is_blind` omits `verify_done_flip`, was still true on `main`.

This PR covers that and more, so **#2649 should close with it** — worth adding `Closes #2649` to the description and as a commit trailer, since GitHub reads closing keywords from those and never from the title.

One note that may be useful for the description: #2649 records the routing consequence of the blind-check half, which is easy to miss because both rungs discard. `Unverifiable` is handled at `pipeline.rs:2330` and `Unverified` at `:2437`, and only the `Unverified` arm reaches the evidence-demand block at `:2467` — so the mis-rung also skips the one remaining channel back to the worker, not just the label. Reward-side they differ too: `DiscardReason::Abstained` vs `Unproven` (`reward.rs:499-500`).

I also proposed the opposite resolution in #2649 — keep `verify_done` from granting the pass, on tamper-exclusion grounds — and your unreachability argument is the reason not to. If `flip_achieved` genuinely cannot be true without a `--test-command`, the cautious-looking option is the one that leaves `SubmitFast` dead. That claim is the load-bearing one in this PR and is worth making the witness test assert directly, so a future reader cannot take it on faith.

macanderson:
## Heads-up: #2647 and #2652 are two implementations of the same fix — merge exactly one

Both were opened within seven minutes and both make `verify_done_flip` reach
`ladder_decision`, which today it never does. They will conflict: each rewrites
the same step of the same function.

| | #2647 | #2652 |
|---|---|---|
| opened | `05:27:31Z` | `05:34:08Z` |
| closes | — (I cross-linked #2649) | #2618 |
| files | 2 | 9 |
| surface | `verify.rs`, `verify/tests.rs` | + `candidate.rs`, `witness/warrant.rs`, `stella-protocol/src/ladder.rs`, regenerated `docs/wire/` |
| wire contract | unchanged | **changed** — `ladder.rs +8/-4` and four regenerated `docs/wire/` artifacts |

That last row is the one to decide on, and it is not a size difference: #2652
moves the wire vocabulary, so it is a protocol change that
`.github/workflows/wire-schema.yml` gates, while #2647 is confined to the
pipeline crate.

They also agree on the diagnosis, which is worth noting because it makes the
finding credible independent of either patch — #2652 states it as a witness
("two `LadderInputs` differing **only** in `verify_done_flip` produced the
identical `LadderDecision::Unverified`"), and #2647 states it as a consequence
("`SubmitFast` is unreachable for any task that does not ship its own test
command"). Same defect, described from both ends.

## Why this warning and not just a nudge

This exact shape has broken `main` here before: two sessions independently
implemented one issue (#1591 and #1594 for #1552), **both** merged, and the
textual union declared `mod daemon;` twice and left an add/add on
`src/daemon.rs` whose surviving file did not match the surviving call sites.
`main` stopped building, and the repair was to revert the smaller PR wholesale
rather than hand-merge the two designs.

Merging the second of these after the first will not conflict cleanly either —
both rewrite step 4 of `ladder_decision` and both add tests to
`verify/tests.rs`.

## Also worth knowing before either lands

`main` does not currently compile — `E0425` in
`crates/stella-cli/src/export/tests.rs`, a merge scar from #2624/#2628. Any CI
green on a branch cut before that is green on a merge result, not on `main`.
**#2645** fixes it and is fully green (all three required checks); #2653 and the
now-closed #2646 are byte-identical duplicates of it.

I have no stake in which of #2647 / #2652 wins — flagging the collision, not
picking. The one thing that should not happen is both.

macanderson:
Heads-up: #2652 was opened ~10 minutes after this one and fixes the same gap (#2618) at the same two sites — `evidence_is_blind` counting the receipt, and rung 4 accepting either receipt. The diagnosis in both descriptions is the same, arrived at independently.

They touch identical lines of `verify.rs`, so whichever lands second will conflict. Consolidating on one is worth doing before either merges.

Comparing against #2618's stated definition of done, #2652 is a superset — not because the logic is better (the two logic changes are equivalent), but because it also covers the parts of the issue this PR leaves open:

| #2618 definition-of-done item | #2647 | #2652 |
|---|---|---|
| `ladder_decision` reads `verify_done_flip` | ✅ | ✅ |
| `evidence_is_blind` counts it (the abstention above the credit) | ✅ | ✅ |
| witness: inputs differing only in `verify_done_flip` differ in outcome | ✅ (2 tests) | ✅ (3 tests) |
| `unverified_evidence` never denies a receipt the turn is holding | — | ✅ |
| the three doc comments that assert the absent behaviour | — | ✅ |
| `CandidateScore::VerifierPass` disposition | — | ✅ |

On the fourth row specifically: without it a turn holding a confirmed receipt that a guard withheld (over budget, say) still tells the operator *"no test command was tracked, so no fail→pass flip could be observed — the only thing that can prove this change is a test that failed before it and passes after it"* — which is the sentence #2618 opens with, and it is false on exactly the turns this PR now credits.

The fifth row includes `LadderSnapshot::verify_done_flip` in `stella-protocol`, whose published doc still describes rescuing a heuristic fallback that no longer exists. That one is on the wire, so correcting it takes a `docs/wire/` regeneration with it (description text only, no shape change) — which is why #2652 is 9 files rather than 2.

No opinion on which branch survives; flagging it so the duplication does not merge twice. Either way the surviving PR is currently red on `main`'s broken `crates/stella-cli/src/export/tests.rs:782` (unbound `literal`), not on anything in this diff — #2645 and #2653 both fix that.

macanderson:
## The two conditions are not equivalent — and the difference is a non-monotonicity in #2647

Both branches are now updated onto the fixed `main` (`f7faef6df`), so their next
CI run reflects their own merits. The earlier red on both was the inherited
`E0425` in `export/tests.rs`, not either change.

Comparing the two step-4 conditions directly:

- **#2647** — `(flip_achieved || verify_done_flip) && (!flip_achieved || touched_tests_passed == Some(true))`
- **#2652** — `(flip_achieved && touched_tests_passed == Some(true)) || verify_done_flip`

| `flip_achieved` | `verify_done_flip` | `touched_tests_passed` | #2647 | #2652 |
|---|---|---|---|---|
| true | false | `Some(true)` | SubmitFast | SubmitFast |
| true | false | `None` | falls through | falls through |
| false | true | `None` | SubmitFast | SubmitFast |
| true | true | `Some(true)` | SubmitFast | SubmitFast |
| **true** | **true** | **`None`** | **falls through** | **SubmitFast** |

They agree everywhere except the last row, and that row is the interesting one:
under #2647 a turn holding a confirmed `verify_done` receipt is credited — but a
turn holding **that same receipt plus a pipeline flip** is not. Adding evidence
removes the credit.

To be fair to #2647: this is not a regression against `main`, which denies that
row too. It is a non-monotonicity *inside* the new rule, and it reads as
unintended — the reasoning in its comment considers a `verify_done` receipt
standing alone ("a `verify_done` receipt has no such companion — it ran both
trees itself") and does not appear to consider the two receipts coexisting.

It is reachable: `touched_tests_passed` is `None` whenever no test touches the
changed lines, which is independent of whether a configured command flipped. A
task with a configured command, no touching tests, and a worker that also ran
`verify_done` lands exactly there.

#2652 has no such corner, because it states the two receipts as independent
sufficient conditions rather than as one condition with an exception carved
into it.

## Two other differences worth weighing

**#2652 also repairs the operator-facing summary.** On `main` the shortfall
string is built from `!flip_achieved`, so a turn holding a confirmed receipt was
told *"no test command was tracked, so no fail→pass flip could be observed — the
only thing that can prove this change is a test that failed before it and passes
after it"* — when a test had just done exactly that. #2652 routes that string
through the receipt predicate and names which channel carried it; #2647 leaves
it reading the old field. Given this repo's standard that a rung is what a human
reads to decide what to repair, that is not cosmetic.

**#2652's protocol change is doc-comment only.** I flagged the wire change in my
earlier comment as the thing to decide on; having read it, that was too strong.
`stella-protocol/src/ladder.rs` gets a rewritten doc comment on the existing
`verify_done_flip` field — no field added, removed or retyped — and the four
regenerated `docs/wire/` artifacts are that prose flowing into the generated
schema descriptions. So it is not a structural wire change and should not weigh
against it.

## Recommendation

**#2652**, on the last-row behaviour and the summary repair. #2647's
`has_flip_receipt`-free formulation is the tighter diff, but the exception it
carves is the thing that produces the wart, and #2652's named predicate is
reused in both `evidence_is_blind` and step 4 rather than duplicating the
disjunction.

If #2647 is preferred for its smaller surface, the last row is worth fixing
before it lands, and the fix is to drop the second conjunct's dependence on
`flip_achieved` — i.e. converge on #2652's shape.

Not merging either; this is a behaviour change to the ladder and the call is
@macanderson's. Flagging the table because neither PR's description mentions
this corner, and whichever lands, it is the one thing here a reviewer cannot
see by reading either diff alone.

PR #2636, opened 2026-08-10
fix(stella-pipeline): restore the deterministic guards #2588 dropped

macanderson:
Closes #2612

## What this is

#2588 made verification model-free. That was the right call and this PR keeps
every bit of it. What it also did — in collapsing `ladder_decision` to three
lines — was drop guards that never needed a model. This restores those, and
only those.

It is deliberately **not** a revert of anything. Two of #2588's removals are
kept, because its reasoning holds; see "What stays removed".

### 1. A zero-dispatch turn reported `passed: true` again

`LadderDecision::NothingAttempted` had ceased to exist, so a turn that
dispatched nothing matched neither `Revise` nor `SubmitFast` and fell to
`Unverifiable`, whose arm emits `passed: true`.

That is the state the rung was introduced to separate, and the module docs
describing why were still in the tree: *"Eleven Terminal-Bench 2.1 trials ended
that way: `glm-5.2` reasoned for a while, called no tool at all, and the run
declared success on a task it had not touched. Every one scored 0.0."*

The `Unverifiable` arm's own comment even read **"The arm above is what makes
this defensible"** — citing the arm that had been deleted.

Restored above the abstention, failing closed. It reads `mutating_actions`,
which is not a probe into the world but the pipeline's tally of the calls it
itself dispatched — the one input here that can never come back blind, and so
the only one this ladder will read as evidence of *absence*.

### 2. A single lucky pass bought a deterministic pass

The #859 confirmation re-run was never executed: `pipeline.rs` had no
`oracle.confirm` call. A flaky test that failed the baseline for an unrelated
reason and passed once was credited in full.

The pre-submit audit is back, gated on the *decision* rather than the flip
transition, so a turn already headed for an abstention pays nothing. It runs
through the retrying runner (#1294) so an OOM does not silently withdraw a pass
the candidate had earned.

**Visible in the re-recorded golden**: `single_task_deterministic_flip` gains one
event — a second `oracle` observation on the candidate tree. That is the guard
executing, pinned in the fixture.

### 3. "Not proven" and "could not look" were one rung

Both returned `Unverifiable`. They imply opposite repairs — produce the missing
observation, versus fix the instrument — and `LadderRung`'s own doc comment
calls the distinction load-bearing. `Unverified` is a rung again.

## What stays removed — #2588 was right

I restored these two at first and then took them back out, because the
reasoning in #2588's tests survives scrutiny and mine did not:

- **Lint diagnostics (#861).** "Diagnostics are observations, not a second
[...]

Issue #1701, opened 2026-08-06
warrant() waives the witness as "NothingChanged" for a candidate that dispatched 10 mutating calls — a false deterministic pass on system-config tasks


macanderson:
## Problem

`witness::warrant` can return `NoWitnessReason::NothingChanged` for a turn that
dispatched mutating tool calls, and `warranted_completion` then emits
`Verdict { passed: true, deterministic: true, rung: Waived }`. The run reports a
clean deterministic pass while the candidate workspace collected **nothing**.

Observed in a Terminal-Bench-shaped `stella run` (Aug 4 2026, ~04:22 UTC, before
#1540 landed). The task was "install and configure nginx on port 8080". The
worker did the work correctly against the real system paths (`apt-get install
nginx`, `/etc/nginx/nginx.conf`, `/etc/nginx/conf.d/benchmark-site.conf`,
`/var/www/html`, `service nginx restart`, `curl localhost:8080` → 200), never
inside `/tmp/stella_candidate_341_0/`. The run ended:

```json
{"type":"proof","step":{"kind":"warrant","required":false,
  "reason":"no files changed; there is no behavior to prove","diff_lines":0}}
{"type":"verdict","passed":true,"evidence":{
  "summary":"no witness test warranted: no files changed; there is no behavior to prove",
  "deterministic":true,"evidence_refs":["warrant:NothingChanged"],
  "ladder":{"rung":"waived","diff_lines":0,"diff_available":true,
            "file_change_events":0,"mutating_actions":10}}}
```

`mutating_actions: 10` and `diff_lines: 0` with `diff_available: true` is not
"nothing to prove" — it is "ten mutating calls landed somewhere this run does
not collect". The verdict asserts the opposite, deterministically.

## Why the existing guards do not catch it

- `warrant(diff, file_changes)` (`crates/stella-pipeline/src/witness/warrant.rs:168`)
  has exactly one anti-lie signal: `file_changes`. Its own sibling module states
  that signal is structurally dead here — `verify_probes.rs:440`: *"the engine
  emits no `FileChange` events (the deck's `FileChangeTap` synthesizes them, and
  it wraps only the SESSION tool stack), so inside a best-of-N or witness
  candidate the count is always zero."* So in a candidate, `warrant`'s guard is
  always `0` and can never fire.
- `mutating_actions` — the signal that *was* live (10) — is computed
  (`pipeline.rs:3202`), threaded into `LadderInputs` (`verify.rs:402`) and used
[...]

Issue #2649, opened 2026-08-10
pipeline: `evidence_is_blind` does not count `verify_done_flip`, so a confirmed worker witness is scored "nothing could look"


macanderson:
## Problem

`LadderInputs::evidence_is_blind` (`crates/stella-pipeline/src/verify.rs:516`)
asks four channels whether anything observed the turn, and does not ask the
fifth:

```rust
pub fn evidence_is_blind(&self) -> bool {
    !self.flip_achieved
        && self.touched_tests_passed.is_none()
        && !self.diff_available
        && self.file_change_events == 0
}
```

`verify_done_flip` is missing. Its sibling predicate three lines down,
`verifier_pass_stands_alone` (`verify.rs:552`), *does* count it:

```rust
!self.flip_achieved && self.touched_tests_passed != Some(true) && !self.verify_done_flip
```

and that predicate's own doc comment argues at length for why it must:

> A worker-run `verify_done` confirmation counts as corroboration (#2129): it is
> a deterministic tool observation — the witness failed on the pinned baseline
> and passed on the change — not another model's opinion, which is exactly the
> class of evidence this predicate exists to demand.

The same reasoning applies to blindness. A channel that is good enough to
corroborate a pass is good enough to prove the turn was not unobserved.

## Why it matters — the rung is wrong, and so is the route

`ladder_decision` (`verify.rs:645`) checks `evidence_is_blind` at step 3, above
the credit at step 4. So for a turn where:

- the worker ran `verify_done` and it printed `WITNESS CONFIRMED`
  (`verify_done_flip: true`, set at `pipeline/evidence.rs:45`), and
- there is no `--test-command` and no authored witness, so the pipeline's own
  oracle never ran (`flip_achieved: false`, `touched_tests_passed: None` — the
  field doc at `verify.rs:470` states this is deliberately distinct: "the
  pipeline's oracle tracks only its own command"), and
- the diff probe could not read the tree (`diff_available: false`), and
- collection recorded nothing (`file_change_events: 0`)

…`evidence_is_blind()` returns `true` and the turn is scored
`LadderDecision::Unverifiable` — *"nothing here can tell you either way"* —
while an instrument did look and reported a confirmed fail→pass flip.

Two consequences, both material:

1. **Wrong discard label.** `Unverifiable` → `DiscardReason::Abstained`;
   the correct rung `Unverified` → `DiscardReason::Unproven`
   (`reward.rs:499-500`). The reward magnitude is the same (both discard), so
[...]

PR #2652, opened 2026-08-10
fix(stella-pipeline): let a confirmed `verify_done` carry a deterministic pass

### LINKED ISSUE #2618: pipeline: a confirmed `verify_done` WITNESS CONFIRMED grants no completion credit — the ladder never reads `verify_done_flip`
## Problem

`LadderInputs::verify_done_flip` — the worker's own `verify_done` shadow run reporting
`WITNESS CONFIRMED`, a real fail-on-baseline / pass-on-candidate observation — steers **no
ladder decision**. It is read in exactly one predicate
(`LadderInputs::verifier_pass_stands_alone`, `crates/stella-pipeline/src/verify.rs:548`),
and that predicate now gates only the one-shot evidence-demand revision
(`crates/stella-pipeline/src/pipeline.rs:2467`). `ladder_decision`
(`verify.rs:625`) does not mention it.

Consequence: the strongest deterministic proof this toolchain can produce cannot reach a
passing rung. Only the pipeline's own tracked-command oracle (`flip_achieved`) can.

## Evidence

Run against `703976e`, as a temporary `crates/stella-pipeline/tests/` probe:

```rust
let inputs = LadderInputs {
    verify_done_flip: true,
    touched_tests_passed: Some(true),
    flip_achieved: false,
    mutating_actions: 6, file_change_events: 3,
    diff_available: true, diff_lines: 10, diff_budget: 400,
    diff_coverage: DiffCoverage::Covered,
    ..Default::default()
};
assert_ne!(ladder_decision(&inputs), LadderDecision::SubmitFast); // passes
```

and, holding everything else fixed, flipping `verify_done_flip` between `true` and `false`
yields the identical `LadderDecision::Unverified` in both cases.

The operator-facing summary then states the opposite of what happened. With
`verify_done_flip: true`, `unverified_evidence` emits:

> UNVERIFIED — no test command was tracked, so no fail→pass flip could be observed — **the
> only thing that can prove this change is a test that failed before it and passes after
> it.**

A test just did exactly that.

## Two comments assert the behaviour that is absent

Both are claims about code, and both are false as written:

- `crates/stella-pipeline/src/witness/warrant.rs:207` — "**the ladder reads it as flip
  evidence** the pipeline's own oracle (which tracks only its own command) cannot see."
- `crates/stella-pipeline/src/verify.rs:465-471` — the field's own doc implies it is
  completion evidence "distinct from `flip_achieved`".

## How it got here (history, not inference)

- Pre-#2588 `main` (`12a509f`): `ladder_decision` did not read it either, but
  `LadderDecision::Unverified` escalated to the model verifier, which could return
  `score_from_verification(false, Some(true))` → `CandidateScore::VerifierPass`
  (`git show 12a509f:crates/stella-pipeline/src/pipeline.rs`, line 2702). So a confirmed
  `verify_done` had a path to a pass, through a model.
- #2588 (`aa2e9dc`) wired it deterministically:
  `git show aa2e9dc:crates/stella-pipeline/src/verify.rs` line 567 —
  `if inputs.flip_achieved || inputs.verify_done_flip {`.
- #2615 reverted #2588 to restore the five guards #2612 named. That revert was correct on
  its own terms; this capability went with it and was not among the tradeoffs #2612
  enumerated, because #2612 audited what #2588 *dropped*, not what it *added*.

Net: today there is no path — deterministic or otherwise — from a confirmed `verify_done`
to a passing verdict. `score_from_verification`'s `Some(true) => VerifierPass` arm
(`crates/stella-pipeline/src/candidate.rs:115`) is correspondingly unreachable from
production; every one of the seven live call sites passes `None` or `Some(false)`.

## What to do

Credit it, or state plainly that it is telemetry only. Crediting is the position #2588
took and the one the module's own prose argues for — it is a deterministic tool
observation with a pinned baseline, which is the class of evidence
`verifier_pass_stands_alone` was written to demand. If it is credited, note that
`verify_done` pins its own baseline (`WITNESS_BASELINE_WORKTREE_REF`, else the
first-parent walk past pipeline seal commits) rather than the pipeline's pre-execution
snapshot, so the two flips are not interchangeable and the rung should say which one it
stands on.

The tamper question is already answered: the harvest is call-id correlated to a dispatched
`verify_done` call **and** requires `content.starts_with("WITNESS CONFIRMED")`
(`crates/stella-pipeline/src/pipeline/execute_stage.rs:330,367-375`), so neither an MCP
tool shadowing the name nor a shell `echo` can forge it.

## Definition of done

- `ladder_decision` either reads `verify_done_flip` or the field's three doc comments
  (`verify.rs`, `warrant.rs:207`, `stella-protocol/src/ladder.rs:235`) say it is recorded
  for telemetry and steers nothing.
- `unverified_evidence` never tells an operator that no fail→pass flip could be observed
  on a turn where `verify_done_flip` is `true`.
- Witness test: a `LadderInputs` differing only in `verify_done_flip` produces different
  outcomes (fails on today's code, where they are equal).
- If `VerifierPass` stays unreachable, it is removed or its doc states it labels stored
  trajectories only — the same disposition `DiscardReason::VerifierDistrusted` already has.
- `make gate` green.

Refs #2584, #2588, #2612, #2615, #2129, #2616

macanderson:
Closes #2618

## What this is

`LadderInputs::verify_done_flip` — the worker's own `verify_done` shadow run
reporting `WITNESS CONFIRMED`, a fail-on-pinned-baseline / pass-on-change
observation made by a **deterministic tool** — steered no ladder decision. It
was read by exactly one predicate (`verifier_pass_stands_alone`), which gates
only the one-shot evidence-demand revision; `ladder_decision` never mentioned
it.

The consequence, stated as the witness states it: two `LadderInputs` differing
**only** in `verify_done_flip` produced the identical `LadderDecision::Unverified`.
The strongest proof this toolchain can produce could not reach a passing rung.

And the operator summary said the opposite of the trace:

> UNVERIFIED — no test command was tracked, so no fail→pass flip could be
> observed — **the only thing that can prove this change is a test that failed
> before it and passes after it.**

A test had just done exactly that. That arm is also the *common* one: it fires
when no test command was configured, which is precisely the task `verify_done`
exists to prove (#2129).

## The fix is two halves, and the second is the one that is easy to miss

**1. Step 4 credits either receipt.**

```rust
let receipt_is_corroborated =
    (inputs.flip_achieved && inputs.touched_tests_passed == Some(true))
        || inputs.verify_done_flip;
```

The pipeline's own flip still needs the touched-test run beside it — that
oracle proves only that *its* command went red→green and says nothing about the
rest of the suite. A `verify_done` confirmation already contains both halves:
the tool ran the witness on the baseline it pinned and again on the change, so
the red→green observation and the green-on-the-change observation are one run.
Demanding a second, pipeline-side green on top would withhold the credit on
exactly the tasks the field was added for, where `touched_tests_passed` is
structurally `None` and no amount of correct work can make it `Some(true)`.

**2. `evidence_is_blind` counts it.**

The abstention sits *above* the credit, and a `verify_done`-only turn satisfies
all four of its dark channels. Crediting the receipt at step 4 while step 3
still read `flip_achieved` alone would have been a fix that never executes —
`a_verify_done_receipt_is_not_a_blind_turn` is the witness for this half
specifically.

## What is deliberately unchanged

- **Every existing guard**, applied to both receipts: the diff budget, the #861
  diagnostics veto, the #870 mutation audit, the #1291 coverage screen.
- **A red touched test still wins.** It returns `Revise` at step 1, above both
  receipts, so neither channel can talk over a test that is failing now.
- **The #1701 escape check is NOT widened.** `effects_escaped_collection` still
  abstains regardless of the receipt: mutating calls plus a *readable zero-line*
  diff means the two trees are identical, and a test "flipping" between
  identical trees is incoherent rather than probative. Extending the
  `flip_achieved` symmetry there would have re-opened #1701.

## Three doc comments asserted this behaviour while it was absent

All named in #2618, all corrected, all claims about code rather than prose:

| file | was |
|---|---|
| `stella-pipeline/src/witness/warrant.rs` | "the ladder reads it as flip evidence…" — it did not |
| `stella-pipeline/src/verify.rs` (field doc) | implied completion evidence "distinct from `flip_achieved`" |
| `stella-protocol/src/ladder.rs` | "it rescues the heuristic fallback from a verifier outage" — there is no heuristic fallback |

That last one is **published on the wire**, so `docs/wire/` is regenerated.
Read as a wire change: the diff is a `description` string on one existing
field. No field added, removed, renamed or re-tagged; the contract stays
additive-only.

`CandidateScore::VerifierPass` is documented as retired — unreachable from
production now that verification is model-free, retained because it labels
**stored** best-of-N trajectories and deleting it would renumber the derived
`Ord` underneath them. Same disposition `DiscardReason::VerifierDistrusted`
already carries.

## Relationship to #2636

#2636 fixes this same gap, but it branched from `3d87aed90` — before #2615
landed — so it also re-does everything #2584 already put on `main`, and
`git merge-tree` reports conflicts in 18 files. This PR is only the part `main`
is actually missing, cut against current `main`.

## Verification

- **Witness tests** — all three fail on the parent commit and pass here,
  checked by reverting the three logic lines and re-running:
  ```
  a_verify_done_receipt_is_a_deterministic_pass ... FAILED   (left: Unverified, right: SubmitFast)
  a_verify_done_receipt_is_not_a_blind_turn ... FAILED
  an_unverified_summary_never_denies_a_verify_done_receipt ... FAILED
  ```
- `cargo test -p stella-pipeline` — 688 passed, 0 failed.
- `cargo clippy -p stella-pipeline -p stella-protocol --all-targets -- -D warnings` — clean.
- `make guards-fast` — green, including `file-size` with no baseline change,
  and `cargo fmt --check`.
- `scripts/check-wire-schema.sh` was red before regenerating and the four
  artifacts are committed.

The workspace suite and rustdoc were **not** run locally (this machine is
reserved for benchmark runs) — CI is the authority on those.

## Deleted tests (for `check-deleted-tests`)

None. No test was deleted or renamed.

Refs #2584, #2588, #2612, #2615, #2129, #2636

## Summary by Sourcery

Treat a confirmed `verify_done` witness flip as deterministic pass evidence in the ladder pipeline and ensure it can independently drive a `SubmitFast` decision and operator messaging.

New Features:
- Introduce a unified `has_flip_receipt` predicate so the ladder treats either the pipeline flip or a `verify_done` confirmation as valid deterministic flip evidence.
- Allow `verify_done` confirmations to carry `SubmitFast` independently of pipeline test telemetry, while preserving existing guards and budgets.

Bug Fixes:
- Fix ladder blind-evidence detection so turns with only a `verify_done` receipt are not misclassified as unobserved.
- Correct `Unverified` operator summaries to acknowledge `verify_done` receipts instead of incorrectly stating that no fail→pass flip was observed.

Enhancements:
- Refine ladder decision logic to factor in corroborated flip receipts and distinguish which channel (pipeline vs `verify_done`) carried the observation.
- Clarify semantics of `verify_done_flip` and its relationship to pipeline flip evidence across ladder, warrant, and candidate scoring, marking `VerifierPass` as a retired rank.
- Update doc comments and wire schema descriptions to reflect that `verify_done` is a completion receipt capable of driving deterministic passes without changing the wire contract.

Tests:
- Add ladder tests asserting that `verify_done` receipts yield deterministic passes, are not swallowed by the blind rung, and are correctly reflected in `Unverified` summaries.


macanderson:
## Heads-up: #2647 and #2652 are two implementations of the same fix — merge exactly one

Both were opened within seven minutes and both make `verify_done_flip` reach
`ladder_decision`, which today it never does. They will conflict: each rewrites
the same step of the same function.

| | #2647 | #2652 |
|---|---|---|
| opened | `05:27:31Z` | `05:34:08Z` |
| closes | — (I cross-linked #2649) | #2618 |
| files | 2 | 9 |
| surface | `verify.rs`, `verify/tests.rs` | + `candidate.rs`, `witness/warrant.rs`, `stella-protocol/src/ladder.rs`, regenerated `docs/wire/` |
| wire contract | unchanged | **changed** — `ladder.rs +8/-4` and four regenerated `docs/wire/` artifacts |

That last row is the one to decide on, and it is not a size difference: #2652
moves the wire vocabulary, so it is a protocol change that
`.github/workflows/wire-schema.yml` gates, while #2647 is confined to the
pipeline crate.

They also agree on the diagnosis, which is worth noting because it makes the
finding credible independent of either patch — #2652 states it as a witness
("two `LadderInputs` differing **only** in `verify_done_flip` produced the
identical `LadderDecision::Unverified`"), and #2647 states it as a consequence
("`SubmitFast` is unreachable for any task that does not ship its own test
command"). Same defect, described from both ends.

## Why this warning and not just a nudge

This exact shape has broken `main` here before: two sessions independently
implemented one issue (#1591 and #1594 for #1552), **both** merged, and the
textual union declared `mod daemon;` twice and left an add/add on
`src/daemon.rs` whose surviving file did not match the surviving call sites.
`main` stopped building, and the repair was to revert the smaller PR wholesale
rather than hand-merge the two designs.

Merging the second of these after the first will not conflict cleanly either —
both rewrite step 4 of `ladder_decision` and both add tests to
`verify/tests.rs`.

## Also worth knowing before either lands

`main` does not currently compile — `E0425` in
`crates/stella-cli/src/export/tests.rs`, a merge scar from #2624/#2628. Any CI
green on a branch cut before that is green on a merge result, not on `main`.
**#2645** fixes it and is fully green (all three required checks); #2653 and the
now-closed #2646 are byte-identical duplicates of it.

I have no stake in which of #2647 / #2652 wins — flagging the collision, not
picking. The one thing that should not happen is both.

macanderson:
Two notes on state.

**Duplicate.** #2647 was opened ~10 minutes before this and fixes the same #2618 gap at the same two sites, arrived at independently. The two logic changes are equivalent; this PR additionally covers three of #2618's definition-of-done items that #2647 leaves open — the `unverified_evidence` wording, the three doc comments asserting the absent behaviour (one of them on the wire), and the `CandidateScore::VerifierPass` disposition. Full comparison in https://github.com/macanderson/stella/pull/2647#issuecomment-5236378468. They edit identical lines, so one should be closed before either merges.

**The red check is not this diff.** `fmt + clippy + test` fails compiling `stella-cli`:

```
error[E0425]: cannot find value `literal` in this scope
  --> crates/stella-cli/src/export/tests.rs:782:28
782 |             !html.contains(literal),
```

That is on `origin/main` — a merge left a loop binding `rounded` while its body reads `literal`, in a duplicate of the radius loop directly above it. This PR touches `stella-pipeline`, `stella-protocol` and `docs/wire/` only, and never compiles differently because of it; the bin-target compile error just takes the whole workspace test step down with it. #2645 and #2653 both fix it. Re-run this check once one of them lands.

Everything else is green, including `docs/wire still matches the types`, `file size ratchet`, MSRV, and `cargo deny`.

macanderson:
## The two conditions are not equivalent — and the difference is a non-monotonicity in #2647

Both branches are now updated onto the fixed `main` (`f7faef6df`), so their next
CI run reflects their own merits. The earlier red on both was the inherited
`E0425` in `export/tests.rs`, not either change.

Comparing the two step-4 conditions directly:

- **#2647** — `(flip_achieved || verify_done_flip) && (!flip_achieved || touched_tests_passed == Some(true))`
- **#2652** — `(flip_achieved && touched_tests_passed == Some(true)) || verify_done_flip`

| `flip_achieved` | `verify_done_flip` | `touched_tests_passed` | #2647 | #2652 |
|---|---|---|---|---|
| true | false | `Some(true)` | SubmitFast | SubmitFast |
| true | false | `None` | falls through | falls through |
| false | true | `None` | SubmitFast | SubmitFast |
| true | true | `Some(true)` | SubmitFast | SubmitFast |
| **true** | **true** | **`None`** | **falls through** | **SubmitFast** |

They agree everywhere except the last row, and that row is the interesting one:
under #2647 a turn holding a confirmed `verify_done` receipt is credited — but a
turn holding **that same receipt plus a pipeline flip** is not. Adding evidence
removes the credit.

To be fair to #2647: this is not a regression against `main`, which denies that
row too. It is a non-monotonicity *inside* the new rule, and it reads as
unintended — the reasoning in its comment considers a `verify_done` receipt
standing alone ("a `verify_done` receipt has no such companion — it ran both
trees itself") and does not appear to consider the two receipts coexisting.

It is reachable: `touched_tests_passed` is `None` whenever no test touches the
changed lines, which is independent of whether a configured command flipped. A
task with a configured command, no touching tests, and a worker that also ran
`verify_done` lands exactly there.

#2652 has no such corner, because it states the two receipts as independent
sufficient conditions rather than as one condition with an exception carved
into it.

## Two other differences worth weighing

**#2652 also repairs the operator-facing summary.** On `main` the shortfall
string is built from `!flip_achieved`, so a turn holding a confirmed receipt was
told *"no test command was tracked, so no fail→pass flip could be observed — the
only thing that can prove this change is a test that failed before it and passes
after it"* — when a test had just done exactly that. #2652 routes that string
through the receipt predicate and names which channel carried it; #2647 leaves
it reading the old field. Given this repo's standard that a rung is what a human
reads to decide what to repair, that is not cosmetic.

**#2652's protocol change is doc-comment only.** I flagged the wire change in my
earlier comment as the thing to decide on; having read it, that was too strong.
`stella-protocol/src/ladder.rs` gets a rewritten doc comment on the existing
`verify_done_flip` field — no field added, removed or retyped — and the four
regenerated `docs/wire/` artifacts are that prose flowing into the generated
schema descriptions. So it is not a structural wire change and should not weigh
against it.

## Recommendation

**#2652**, on the last-row behaviour and the summary repair. #2647's
`has_flip_receipt`-free formulation is the tighter diff, but the exception it
carves is the thing that produces the wart, and #2652's named predicate is
reused in both `evidence_is_blind` and step 4 rather than duplicating the
disjunction.

If #2647 is preferred for its smaller surface, the last row is worth fixing
before it lands, and the fix is to drop the second conjunct's dependence on
`flip_achieved` — i.e. converge on #2652's shape.

Not merging either; this is a behaviour change to the ladder and the call is
@macanderson's. Flagging the table because neither PR's description mentions
this corner, and whichever lands, it is the one thing here a reviewer cannot
see by reading either diff alone.

Issue #1976, opened 2026-08-07
A merge that deletes another PR's test is invisible to CI — add a deleted-test guard


macanderson:
## Problem

Three separate times now, a PR has landed on `main` that silently deleted code
another PR added to the same file earlier the same day, and CI could not see
it. The most recent (fixed in #1975): #1951 rewrote
`crates/stella-pipeline/src/pipeline/tests/verification_hardening.rs` from a
pre-#1945 base, deleting the `PassingShell` double, `shell_call_result`, and
the witness `a_revision_halts_at_the_step_where_the_tracked_test_flips` that
#1945 had added hours earlier.

`main` went red. It was caught only because #1945 had *also* added a
`mod flip_halt_arming;` line that survived the rewrite and referenced two of the
deleted symbols. **Had #1945 added only the test and the doubles — no new
module — the deletion would have compiled clean and silently removed a witness
from the tree.** A witness that no longer exists cannot fail, so nothing
downstream would ever have reported it.

This is the same shape recorded for #1860 silently reverting #1836's forwarding
in four crates.

## Why CI cannot see it

Both PRs are green against the `main` they branched from. The merge is textually
clean — git has no conflict to report, because one side simply does not contain
the other's lines. Branch protection's "require branches to be up to date"
would catch it, but it is off (and turning it on serializes every merge, which
is its own cost at this repo's merge rate).

## Fix direction

The cheap, high-value version — a **deleted-test guard**, in the spirit of
`check-left-behind.sh`:

For a PR's merge base against `main`, enumerate `#[test]` / `#[tokio::test]`
function names present in the base and absent in the head. A non-empty set is
not automatically wrong — tests are legitimately renamed, moved between
modules, or deliberately removed — so the guard should **fail asking for an
acknowledgement** rather than fail outright: the PR body (or a commit trailer)
must name each removed test. That converts an invisible deletion into a
sentence a reviewer reads.

Cheaper still, and worth doing regardless: this only needs to run in CI on the
merge result, not in `make gate` locally, since it is inherently a
two-branch question.

Deliberately NOT proposed: requiring up-to-date branches, or forbidding test
file rewrites. Both cost more than the defect.

## Files

- `scripts/` — new guard, modelled on `scripts/check-left-behind.sh` (the
  closest existing shape: a mechanical check whose output is "name the thing or
  fix it")
- `Makefile` — `GATE_STEPS`, if it is to run locally too (note: adding a step
[...]

PR #2653, opened 2026-08-10
fix(stella-cli): unbreak main — a merge left an unbound `literal` in the export tests

macanderson:
## main does not compile

```
$ cargo test -p stella-cli --bin stella export::tests
error[E0425]: cannot find value `literal` in this scope
   --> crates/stella-cli/src/export/tests.rs:782:28
    |
782 |             !html.contains(literal),
    |                            ^^^^^^^ not found in this scope
error: could not compile `stella-cli` (bin "stella" test) due to 2 previous errors
```

`cargo test --workspace` is a `make gate` step, so this reds every PR branched
from `main` until it lands.

## What happened

Two PRs added a rounded-corner assertion to the same test. #2628 added

```rust
for rounded in ["border-radius: 8px", "border-radius: 6px",
                "border-radius: 3px", "border-radius: 2px",
                "border-radius: 0 6px 6px 0"] {
    assert!(!html.contains(rounded), …);
}
```

and another added

```rust
for literal in ["border-radius: 8px", "border-radius: 6px"] {
    assert!(!html.contains(literal), "`{literal}` bypasses `--radius`…");
}
```

The merge kept **both bodies** but only one header, leaving a loop that binds
`rounded` and reads `literal`. Textually clean — git had no conflict to report,
because neither side's lines were deleted. It is the same shape as the
test-file collisions in #1976 and #1860, one level down: not a lost test, a
lost *binding*.

## The fix

Deleted the second loop rather than rebinding it. Its two literals are a subset
of the five above, both loops assert the same property against the same `html`,
and one assertion in one place is the point of the first loop. Nothing is lost:
`border-radius: 8px` and `6px` are still asserted absent.

## Verification

```
cargo test -p stella-cli --bin stella export::tests   # 26 passed, 0 failed
cargo fmt --all -- --check                            # clean
```

No test was deleted or renamed — a redundant loop *inside* a surviving test was
removed, and the test's assertions are unchanged in effect.

## Why this keeps happening

Fourth time in two days that `crates/stella-cli/src/export/tests.rs` has gone
red from a two-sided edit. The underlying reason — the export dashboard's
appearance has no single source of truth, so independent PRs keep encoding the
same decision in the same file from different directions — is **#2626**.

Refs #2626

## Summary by Sourcery

Bug Fixes:
- Resolve an unbound variable in the export dashboard test that prevented stella-cli from compiling and running tests.

### DIFF
--- a/crates/stella-cli/src/export/tests.rs
+++ b/crates/stella-cli/src/export/tests.rs
@@ -777,12 +777,6 @@ fn the_dashboard_is_monospace_and_square() {
              where a hairline says \"boundary\", and this page is boundaries"
         );
     }
-    for rounded in ["border-radius: 8px", "border-radius: 6px"] {
-        assert!(
-            !html.contains(literal),
-            "`{literal}` bypasses `--radius`: set the token, not the rule"
-        );
-    }
 }
 
 /// The wordmark is lowercase, everywhere the report says it.


macanderson:
Duplicate of #2645, which is the byte-identical six-line deletion and was opened first. Three sessions independently produced this same fix within 13 minutes:

| PR | opened | state |
|---|---|---|
| #2645 | `05:21:33Z` | **fully green** — all three required checks pass (`fmt + clippy + test` in 13m52s) |
| #2646 | `05:25:39Z` | closed by me as a duplicate |
| #2653 | `05:34:55Z` | this one |

All three leave `crates/stella-cli/src/export/tests.rs` at the same blob, `a0c406ba1`, so there is no content to choose between them.

**Merge #2645.** It is earliest and already verified end to end; whichever of these lands second becomes an empty diff and fails the `PR carries a diff` check.

Suggest closing this one. The diagnosis behind all three — the #2624/#2628 rename-vs-edit merge scar, why one compile error in a test target takes down three separate CI steps, and how it reached `main` 2.5 minutes ahead of its own required check — is recorded as a comment on #2626 so it does not disappear with whichever PRs get closed.

macanderson:
Landed via #2645 (`f7faef6df`), which was the byte-identical fix opened first and fully green. `main` compiles again — `contains(literal)` is gone from `crates/stella-cli/src/export/tests.rs`. Closing this as a duplicate; nothing in it is lost.

PR #2662, opened 2026-08-10
feat: stop-on-proof and the prove-it gate — kill tools on oracle flip, fire the halt on verify_done, and make task mode earn its declaration

### LINKED ISSUE #2661: pipeline: SIGKILL in-flight worker tools on oracle flip — a proven-done turn must not keep spending the ceiling (solved_then_timeout)
## Problem

The flip oracle already observes the witness fail→pass flip **live** (#1793): `FlipHalt::observe` latches on the worker's own bash `ToolResult` (`crates/stella-pipeline/src/pipeline/execute_stage.rs:337-372`, `crates/stella-pipeline/src/flip_halt.rs:150`). But the halt only lands at the next **step boundary** (`crates/stella-core/src/driver/drive.rs:161-176`): the current model stream and every in-flight sibling tool must finish first. When the flip fires while a long tool (a `pytest`, a build, a `sleep`-poll) is still running, the proven-done turn keeps burning wall clock — on TB2.1 that is the `solved_then_timeout` outcome (`arenabench/arenabench/telemetry.py:476-483`), measured 4 times on the 2026-08-10 certification panel, including a 955s trial whose ceiling is 900s.

Directive from Mac (2026-08-10): **SIGKILL the worker on oracle flip.** Once the tracked witness command has flipped fail→pass, everything still in flight is spend past the proof.

## Design (from the code map)

1. **Preemptive halt at the dispatch barrier + group-kill of in-flight siblings.** When the latch fires mid-dispatch, drop the remaining concurrent tool futures — their drop guards already `libc::kill(-pid, SIGKILL)` the whole process group (`crates/stella-tools/src/exec.rs:333-400`, `kill_on_drop` in `bash.rs:301-330`) — and close their calls with synthetic `ToolOutput`s (the `close_open_tool_calls` pattern the soft-stop path uses, `crates/stella-core/src/driver.rs:990-994`) so the transcript stays well-paired. The existing `TurnHalt` check then ends the turn at this boundary with `TurnOutcome::Completed` (never `Aborted` — a nonzero exit scores as a crash on the bench).
2. **Invariant #6 carve-out, argued in docs, not silently breached:** #6 forbids *budget* aborts mid-tool because a budget abort loses work. A confirmed flip is the opposite state — the work is already proven done; what a kill loses is by definition not needed. Document in `flip_halt.rs` + `budget.rs` cross-ref; no invariant renumbering.
3. **Emit a durable event on the flip transition.** Today the observing closure deliberately emits nothing (`execute_stage.rs:355-359` — only `Error` was available), so post-hoc analysis cannot distinguish "flip fired, halt too slow" from "flip never fired." Needs a non-Error event (protocol change + wire-contract sample; mind the wildcard-consumer hazard from the TurnParked lesson).
4. **Optional follow-up (separate):** feed the live `verify_done` `WITNESS CONFIRMED` tally (`execute_stage.rs:360-372`) into the same halt so trials with no configured tracked command get covered too — today `flip_halt` is unarmed there.

## Done looks like

A witness test where the worker runs (a) the tracked command which flips to pass and (b) a concurrent long-running tool: the turn settles well before the long tool's natural duration, the long tool's process group is dead, the transcript has no orphaned `tool_use`, and the outcome is `Completed`. Fails on main (turn waits for the long tool), passes with the change.

Refs #1793, #2022

macanderson:
## What

Directive from Mac: **SIGKILL the worker on oracle flip.** Once the flip oracle has watched the tracked test go fail→pass, everything still in flight is spend past the proof.

The live observation already existed (#1793, `FlipHalt`): the pipeline latches on the worker's own bash `ToolResult`. What was missing was the force — the halt only landed at the **step boundary**, so the current model stream and every in-flight sibling tool had to finish first. A flip landing while a `pytest` or a build was running kept a proven-done turn burning wall clock. On TB2.1 that is `solved_then_timeout`: measured **4 times** on the 2026-08-10 certification panel (run `turnloop-cert-panel1` + reruns), once at 955s against the 900s ceiling.

## How

`EngineConfig::turn_halt` is now consulted **inside the dispatch loop** (`crates/stella-core/src/driver/dispatch.rs`) — between barrier groups and after each settled tool call. On fire:

- the in-flight `buffer_unordered` stream is dropped; each dropped tool future takes its child **process group** with it (`kill_on_drop(true)` + the setsid'd `libc::kill(-pid, SIGKILL)` guard already in `stella-tools/src/exec.rs`);
- every unfinished or never-started call is answered synthetically, following `close_open_tool_calls`' exact shape — an `Error` output with a **steady** message (the loop detector keys on `ToolOutput` content, so nothing volatile rides it), mirrored onto the event stream with no `ToolStart`;
- the returned result set still covers every call, so the transcript stays well-paired, and the existing boundary check in `drive` settles the turn `Completed` (never `Aborted` — harnesses score non-zero exits as crashes) without another model call.

Synthesis is keyed by original call **index**, not `call_id` — ids are only guaranteed unique per response by some providers, and an id-keyed set would let one answered duplicate absorb the other.

## Invariant #6

Deliberately not a breach, and argued in the docs rather than waved past: the safe-boundaries rule guards **budget** aborts, where an interrupt loses work someone still wants and corrupts pairing. A confirmed flip is the opposite state — what the kill discards is, by the flip's own proof, not work — and pairing is preserved by construction. Docs updated: `TurnHalt` trait contract (both consult depths), `flip_halt.rs` module rationale ("How hard the stop lands"), dispatch doc-comment.

## Witness

`driver::dispatch::tests::a_mid_dispatch_flip_kills_the_hung_sibling_and_pairs_its_call`: one group, two dispatch-safe calls — `witness` settles instantly and raises the halt's flag, `eternal` never resolves. **Fails on the old dispatch** (times out draining the sibling; verified by splicing HEAD's implementation under these tests: `1 passed; 1 failed ... finished in 10.01s`), passes here in ~0.01s. The killed call's synthetic answer is asserted present on the event stream. A control test pins the unhalted path unchanged (real results, second model call).

`cargo test -p stella-core`: 1066 passed. `cargo test -p stella-pipeline`: green. Clippy clean.

## Exemplar

The synthetic-closure shape is `close_open_tool_calls` (`stella-core/src/step.rs`) applied at a second site, and the kill primitive is `stella-tools`' existing drop guard — no new mechanism, only a new consult point for an existing port.

## Second commit: verify_done fires the halt on every execute turn (#2661 point 4)

The kill only had a trigger where a tracked command's failing baseline was armed — on Terminal-Bench that is never the bare loop, and not the pipeline's first execute turn either. `verify_done`'s `WITNESS CONFIRMED` (already detected and tallied in the event filter, deterministic per #2129) now fires the same latch: `FlipHalt::confirm_verify_done` + `watching_verify_done_only`, with both engine-turn drivers attaching a halt unconditionally. An unfired latch answers `None` at every consult; the command-correlation map stays gated on a non-empty tracked command (zero overhead for turns watching nothing); a command-less latch can never be fired by a shell observation and is refused to follow-up turns once fired. Witnessed in `flip_halt` tests (fails on the old type — no such constructor or edge); the one unwitnessed link is the six-line closure call in `filtered_turn_events`, which mirrors the `halt.observe` call directly above it and has no existing test scaffolding — disclosed rather than waved past.

## Third commit: the prove-it completion gate (#2663)

The field A/B (run `stoponproof-ab`) showed the worker never calls `verify_done` on bench tasks, so stop-on-proof had no trigger and the bare loop's premature declarations (13-step wrong answers on kv-store-grpc/pypi-server) went unchallenged. `confident_zero::check` — already the tool-less-completion legitimacy gate — gains rung 4: with `EngineConfig::completion_gate` set, a completing turn that mutated the workspace with no confirmed `verify_done` is answered **once** with a prove-it nudge (claim-then-challenge appended, turn continues) instead of being accepted. Once per turn via transcript prefix detection (checkpoint-resume safe); a confirmed `verify_done` or a mutation-free conversational answer never trips it. Off by default; the CLI keys it on the execution kind it already records (`"run"` gated, `"chat"` not), and the pipeline worker config sets it unconditionally. Witnessed in `confident_zero` tests (fails on the old `check`: no gate, no ruling); both god files stay at their ceilings with no baseline change.

## Not in this PR (tracked)

- A durable event on the flip transition (today the observer deliberately emits nothing) — #2661 point 3; needed to measure halt latency in the field.

Closes #2661
Refs #2663
Refs #1793

### DIFF
--- a/crates/stella-cli/src/agent.rs
+++ b/crates/stella-cli/src/agent.rs
@@ -2020,9 +2020,9 @@ async fn run_turn(
         // every other driver path. Wrap the raw registry in `PolicyToolSet`
         // so disabled tools cannot be invoked here either.
         let permitted = PolicyToolSet::new(registry, session_tool_policy(cfg));
-        let engine =
-            Engine::with_sleeper(provider, &permitted, engine_config_for(cfg), &TokioSleeper)
-                .with_calibration(calibration);
+        let config = engine::engine_config_for_kind(cfg, kind);
+        let engine = Engine::with_sleeper(provider, &permitted, config, &TokioSleeper)
+            .with_calibration(calibration);
         engine.run_turn_with_sender(messages, budget, &tx).await
     } else {
         let customs = CustomToolSet::new(
@@ -2046,9 +2046,9 @@ async fn run_turn(
             .with_project_prompts_allowed(cfg.authority.project_prompts_allowed)
             .with_activation(activated.clone());
         let hook_runner = ShellHookRunner;
-        let mut engine =
-            Engine::with_sleeper(provider, &tools, engine_config_for(cfg), &TokioSleeper)
-                .with_calibration(calibration);
+        let config = engine::engine_config_for_kind(cfg, kind);
+        let mut engine = Engine::with_sleeper(provider, &tools, config, &TokioSleeper)
+            .with_calibration(calibration);
         if let Some(hooks) = &cfg.hooks {
             engine = engine.with_hooks(hooks, &hook_runner);
         }
--- a/crates/stella-cli/src/agent/engine.rs
+++ b/crates/stella-cli/src/agent/engine.rs
@@ -196,6 +196,18 @@ pub(crate) fn engine_config_for(cfg: &Config) -> EngineConfig {
     )
 }
 
+/// [`engine_config_for`], with the prove-it completion gate (#2663) switched
+/// by the turn's execution kind: a headless `"run"` turn is graded on its
+/// result and must prove a mutation before its declaration stands; a
+/// `"chat"` turn is conversational and keeps the ungated shape. Keyed on the
+/// same `kind` string `begin_execution` records, so the gate and the
+/// telemetry cannot disagree about which mode a turn ran in.
+pub(crate) fn engine_config_for_kind(cfg: &Config, kind: &str) -> EngineConfig {
+    let mut engine = engine_config_for(cfg);
+    engine.completion_gate = kind == "run";
+    engine
+}
+
 /// EngineConfig for a pipeline's execute turns — the WORKER agent's tuning
 /// (plan and witness ride it too, matching the router's tiering).
 /// `worker_model` is [`EngineWiring::worker_model`]: the model the worker
@@ -203,11 +215,15 @@ pub(crate) fn engine_config_for(cfg: &Config) -> EngineConfig {
 /// `agents.worker.*` when set (issue #276), falling back to the session
 /// default (`cfg.provider`/`cfg.model_id`) when unset.
 pub(crate) fn pipeline_engine_config_for(cfg: &Config, worker_model: &ModelRef) -> EngineConfig {
-    tuned_engine_config(
+    let mut engine = tuned_engine_config(
         cfg,
         crate::settings::EngineAgentKind::Worker,
         (&worker_model.provider, &worker_model.model_id),
-    )
+    );
+    // A pipeline execute turn is always task mode: its declaration is graded,
+    // so it proves a mutation before the declaration stands (#2663).
+    engine.completion_gate = true;
+    engine
 }
 
 /// The safe default for CLI-owned headless surfaces: no host approval port,
--- a/crates/stella-core/src/driver.rs
+++ b/crates/stella-core/src/driver.rs
@@ -334,14 +334,9 @@ pub trait TurnHalt: Send + Sync + std::fmt::Debug {
     /// `Some(reason)` ends the turn cleanly at the next step boundary;
     /// `None` lets it continue.
     ///
-    /// Called once per committed step AND from inside the dispatch loop —
-    /// between barrier groups and after each settled tool call (#2661) — so
-    /// it must be cheap and must not block. A predicate that fires
-    /// mid-dispatch kills the still-in-flight sibling tools (their process
-    /// groups die with their dropped futures) and answers every remaining
-    /// call synthetically, so the turn still ends with a well-paired
-    /// transcript; the work a fired predicate discards is, by the
-    /// predicate's own claim, already proven unnecessary.
+    /// Asked per committed step AND inside the dispatch loop (#2661): cheap,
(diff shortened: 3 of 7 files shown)


macanderson:
Field A/B (cloud run `stoponproof-ab`, 2 pipeline trials on kv-store-grpc, SUT = this branch's `8552aefd3`): both trials ran clean on the new code — no regressions, transcripts well-paired — but **the halt never fired because the worker never called `verify_done`**: zero references in either trial's `stella-events.jsonl`, both trials declared complete (wrong) at 27–31 steps. Neither trial solved, so the tracked-command path had nothing to fire on either.

So this PR's machinery is necessary but not sufficient in the field: the trigger it listens for isn't being emitted by the worker on bench tasks. That behavioral gap — the verified-done contract enforced at verdict but not practiced in the loop — is now #2663, with the trace evidence and design directions. Once the worker actually proves its work (or the witness stage feeds the tracked command forward), this halt is what converts that proof into reclaimed wall clock.

macanderson:
Second field A/B (cloud run `gate-ab`, bare loop ×2 on kv-store-grpc + pypi-server, SUT `5e7e24f1d`):

**The lever works.** Every trial's trace shows the nudge firing and the worker answering it with real `verify_done` calls — 0 on every earlier trial, 4–14 here. The verified-done contract is now practiced in the loop, not just enforced at the verdict.

**It also found its own defect, now fixed and pinned.** The nudge is a user message, and `turn_start_index` restarts the window at any user message — so the once-only marker fell outside its own scan, and a refuted verification plus one more edit re-armed the nudge. pypi-server took 3 nudges in one turn into the 900s ceiling. Fixed in `a537a4ad` (window anchored at the last non-nudge user message), with the exact field transcript shape as the witness test.

**Rewards on these two tasks did not move** (0/4): the wrong answers are upstream of verification — GLM diagnoses its kv/pypi failures incorrectly even when told to check. That places the remaining panel gap in worker capability/diagnosis on these tasks, not in loop mechanics; the gate's value shows where the model *can* act on what the check reveals, and in the verify_done usage it now induces.

Issue #2713, opened 2026-08-10
Single-purpose tool audit: split apply_edits(dry_run)


macanderson:
Refs #2695 (session-environment epic) and #2699 (the read_output/clear_output split this finding was swept alongside).

# Single-purpose tool audit: split `apply_edits(dry_run)`

## Problem

`crates/stella-tools/src/apply_edits.rs`'s `apply_edits` tool takes a `dry_run: bool`
(schema at `crates/stella-tools/src/apply_edits.rs:237`, description: `"Validate every
edit and report, but write nothing (default false)"`). This is a mode-selecting
parameter, not a scoping one, per the distinction in #2699/#2700: `dry_run: true`
makes the tool perform a fundamentally different operation (validate-only, no
filesystem mutation) from `dry_run: false` (apply the edits, mutating files) — the
issue that opened this sweep (#2699) names `dry-run vs apply` explicitly as an example
of the violation shape.

Today `apply_edits` is declared `read_only: false` unconditionally
(check the exact line at the tool's `ToolSchema` — grep
`name: "apply_edits"` in the same file), which is correct for the default/mutating
arm but overbroad for `dry_run: true`, which performs no mutation and could be
`read_only: true` (and a candidate for `speculation_safe: true`) if split into its own
tool. This also means no operator can express "let the agent preview edits but never
apply them" via `tools.<name>` policy today — the same policy-expressiveness gap
#2699 fixed for `read_output`/`clear_output`.

## Work

- Split `apply_edits` into `apply_edits` (always mutates, drops `dry_run`) and a new
  single-purpose read-only tool — e.g. `preview_edits` or `validate_edits` (bikeshed
  in the PR) — that only validates and reports, never writes.
- Verify the validate-only path truly touches no filesystem state before marking the
  new tool `read_only: true` (read the current `dry_run: true` code path in
  `apply_edits.rs` — do not assume from the description alone).
- Update `crates/stella-tools/src/catalog.rs`, the tool's registration site under
  `crates/stella-tools/src/registry/` (`registry.rs` itself is a god file closed to
  growth — zero new lines there), the module doc in `apply_edits.rs`, and
  `website/content/docs/agent-tools/index.mdx`'s File tools `ToolCard`s.
- Keep `dry_run` on `apply_edits` for one release as a deprecation returning a named
  error pointing at the new tool (same pattern as #2699's `read_output` `clear`
  deprecation).

## Tests

Witness: the new preview/validate tool exists, reports what it would do, and leaves
[...]

Issue #2712, opened 2026-08-10
Single-purpose tool audit: split run_lint(fix) and format_code(check)


macanderson:
Refs #2695 (session-environment epic) and #2699 (the read_output/clear_output split this finding was swept alongside).

# Single-purpose tool audit: split `run_lint(fix)` and `format_code(check)`

## Problem

`crates/stella-tools/src/project.rs`'s `RunLint` and `FormatCode` tools each carry a
boolean that *selects which operation runs* rather than *scoping* the same operation —
the same shape #2699 just split out of `read_output`:

- `run_lint`'s `fix: bool` (schema at `crates/stella-tools/src/project.rs:583`, dispatch
  at `:594` and `:603`) picks between "lint" (report only) and "lint --fix" (mutates
  source files). Description: `"Apply automatic fixes (cargo only)"`.
- `format_code`'s `check: bool` (schema at `:627`, dispatch at `:638-641` and `:650-653`)
  picks between "format" (rewrites files) and "format --check" (verifies only, mutates
  nothing). Description: `"Verify without rewriting (cargo only)"`.

Per the distinction #2699 and #2700 write down: a *scoping* parameter (key, path,
offset, limit) qualifies the same verb; a *mode* parameter makes the tool perform a
different verb. `fix` and `check` are mode parameters — one arm of each tool mutates
the working tree, the other doesn't, and nothing here lets an operator's
`tools.<name>` policy or the `command.started` gate distinguish the mutating arm from
the read-only one. Both tools are declared `read_only: false` unconditionally today
(`crates/stella-tools/src/project.rs:587,631`), which is honest for the mutating arm
but overbroad for the non-mutating one (`run_lint` with no `fix`, `format_code` with
`check: true`) — those calls could be `read_only: true` today if they were their own
tool, the same latent-capability gap #2699 named for `read_output`.

## Work

- Split `run_lint` into `run_lint` (report only, becomes eligible for
  `read_only: true` — verify no other mutating side effect exists before flipping it)
  and a new single-purpose `apply_lint_fixes` (or similar name — bikeshed in the PR)
  that always mutates.
- Split `format_code` into `format_code` (always mutates/rewrites) and a new
  single-purpose `check_format` (or similar) that never mutates and is a candidate for
  `read_only: true`.
- Update `crates/stella-tools/src/catalog.rs`'s `"build"` group rows, the
  `crates/stella-tools/src/registry/registry_project.rs`-or-equivalent registration
[...]

PR #2717, opened 2026-08-10
refactor(stella-tools): split clear_output out of read_output

### LINKED ISSUE #2699: Single-purpose tool audit: split read_output(clear) into clear_output
Refs #2695 (session-environment epic, sub-issue D).

# Single-purpose tool audit: split mode-flag tools (`read_output(clear)` → `clear_output`)

Part of the session-environment epic (parent issue). **Worker: Sonnet 5** for the sweep and the split-vs-scope judgments; each mechanical split it identifies may be delegated to Haiku 4.5 with the judgment already written down.

## Problem

The epic's tool-first principle: a boolean that changes *which operation runs* is two tools wearing one schema — it blurs the description the model decides by, and it defeats per-tool policy (`tools.<name>` toggles and the `command.started` gate cannot distinguish the destructive arm from the benign one).

One violation is already identified: `read_output`'s `clear: true` (`crates/stella-tools/src/process.rs` module doc: "`clear: true` discards the buffered output instead of returning it") — a read tool with a mutating arm, which also makes its `read_only` declaration wrong in one direction or the other.

## Work

1. **Split the known case**: `read_output` loses `clear`; new single-purpose `clear_output(handle)` discards the buffer. `read_output` becomes honestly `read_only: true` (check what it declares today — if it already claims read-only *with* the clear arm, that is a latent concurrency bug worth naming in the PR: the engine may run read-only calls concurrently per the `ToolSchema` contract, and a concurrent clear mutates under a reader).
2. **Sweep every registered tool schema** for parameters that select an operation rather than scope one. The distinction, written for the delegated worker: *scoping* parameters (key, path, offset, limit, pattern) qualify the same verb; *mode* parameters make the tool do a different verb (delete vs edit, clear vs read, dry-run vs apply). Enumerate schemas mechanically (the registry can print them; or `rg '"type": "boolean"' crates/stella-tools/src`) and judge each hit.
3. Each additional finding gets its own follow-up issue (one split per issue, written as a handoff per AGENTS.md), not folded into this PR — the known `read_output` split is this PR's scope.
4. **Compat**: removing `clear` changes the model-facing surface. Keep the parameter for one release as a deprecation that returns a named error pointing at `clear_output` (an error the model can act on beats a silently ignored field), and note the removal in the changelog.

## Tests

- Witness: `clear_output` exists and discards (write output, clear, read → empty); fails on `main`.
- `read_output` with `clear: true` returns the named deprecation error.
- The tombstone/reap semantics documented in `process.rs`'s module doc still hold after the split (the module doc must be updated in the same PR — stale comments are bugs here).

## Definition of done

`make gate` green; the sweep's findings filed as linked follow-up issues (or an explicit "no further findings" statement in the PR); module doc updated; witness named in the PR.

macanderson:
## What & why

`read_output`'s `clear: bool` was a mode flag wearing a read tool's schema — it picked
*which operation ran* (discard vs return buffered output) rather than scoping the same
one. That blurs the description the model decides by and defeats per-tool policy:
`tools.<name>: "off"` cannot withhold the destructive `clear` arm without also
withholding the harmless read.

This splits it per #2699:

- New single-purpose `clear_output(handle)` (`crates/stella-tools/src/process.rs`),
  registered in `registry/process_tools.rs`. `registry.rs` needed **zero new lines** —
  it's a closed god file, and `clear_output` runs no external command, so it needs no
  `command_line_for` arm (confirmed by reading the match, not assumed — every other arm
  covers a tool that either spawns a process or feeds a live interpreter's stdin).
- `read_output` loses `clear`. For one release it survives as a **deprecation**: any
  truthy value now returns a named error pointing at `clear_output`, instead of being
  silently ignored or silently honored (an error the model can act on beats either).
- `read_output`'s `catalog.rs` group grew by one; the enterprise-telemetry
  process-free forbidden-tools list, `policy.rs`'s doc example, and the
  process-tool-count references in `docs/spec/remote-sandboxes.md` and
  `website/content/docs/agent-tools/index.mdx` are updated to match.
- `process.rs`'s module doc is rewritten in the same PR (it documented `clear: true` —
  stale docs are bugs here).

### The `read_only` question the issue asked me to check

The issue's premise was: if `read_output` already declared `read_only: true` while
carrying the mutating `clear` arm, that's a latent concurrency bug (the engine may run
read-only calls concurrently per `ToolSchema::read_only`'s contract in
`crates/stella-protocol/src/tool.rs`), and after the split `read_output` "must be
honestly `read_only: true`".

I checked rather than assumed: `read_output` was already `read_only: false`
(`crates/stella-tools/src/process.rs:517` on `main`), so that specific latent bug
does not exist today — no flip needed there.

But I also checked whether it *should* flip to `true` after the split, and it should
not: `read_output`'s plain read is itself destructive, `clear` or not. Every call does
`entry.output.lock().take()` (drains the buffer — the next read sees only what arrived
since), and can reap an exited entry into a tombstone. Both are mutations of process/
table state under `ToolSchema::read_only`'s own definition ("cannot mutate any state
(filesystem, processes, **environment**)"). So `read_output` stays `read_only: false`
after the split — flipping it to `true` would let the engine run concurrent
`read_output` calls, which would nondeterministically split one process's buffer
between them, recreating the same class of hazard the issue is closing, just on the
plain read instead of `clear`. `clear_output` is `read_only: false` for the same
reason (it discards the buffer, unambiguously a mutation).

This is stated explicitly in `process.rs`'s module doc and inline on both schemas so
the reasoning doesn't get lost.

### The sweep (issue step 2)

Grepped every `"type": "boolean"` in `crates/stella-tools/src` and `crates/stella-cli/src`
tool schemas (`rg -n '"type": "boolean"' crates/stella-tools/src crates/stella-cli/src`,
excluding test files) and judged each against the issue's distinction (scoping
qualifies the same verb; mode selects a different one). Two more hits, filed as their
own follow-ups per AGENTS.md's "nothing left behind" (see below) — everything else
scopes an existing verb and isn't a violation:

- `repo_diff`'s `staged`, `repo_history`'s `include_shelved`/`include_position_log` —
  which target/what to include in the same read.
- `ci_status`'s `wait` — synchronous-vs-parked timing of the same read, not a
  different verb; both arms are already `read_only: true`.
- `grep`'s `case_insensitive`, `edit_file`/`apply_edits`'s `replace_all`, `bash`'s
  `trace`, `skill_search`'s `include_body` — behavior/verbosity modifiers of one verb.
- `cite_memory`'s `truthful` — a data field being recorded, not an operation selector.

### `process.rs`'s tests: a second, unrelated-looking gate hit

`clear_output` pushed `process.rs` to 1598 lines, over the file-size gate's 1500-line
limit. It isn't grandfathered (stella-tools's only god files are `registry.rs` and
`scripts.rs`), so this is a first-time crossing that gets split, not a baseline entry
(AGENTS.md's file-size section is explicit that a first-time crossing "still wants
splitting, and still gets no baseline entry"). Moved the `#[cfg(test)] mod tests` block
into `crates/stella-tools/src/process/tests.rs`, mirroring the crate's own
`registry.rs` / `registry/tests.rs` pattern — no behavior change, every test moved
verbatim.

Closes #2699

## The witness

Two tests in `crates/stella-tools/src/process/tests.rs`, both checked against `main`
(`7cb52c5e3`) in a scratch worktree, per AGENTS.md's "artisanal way":

- `clear_output_discards_buffered_output_before_the_next_read` — **fails to compile**
  on `main` (`error[E0425]: cannot find function, tuple struct or tuple variant
  ` `ClearOutput` in this scope`, `process.rs:833`) because the tool doesn't exist yet.
  Passes here: writes output via a started process, `clear_output`s it, then
  `read_output` reports `[no new output]`.
- `read_output_clear_true_is_a_named_deprecation_error` — **fails at runtime** on
  `main` (`clear: true must be a named error, not: Ok { content: "... [cleared 0
  buffered bytes]" }` — main silently honors `clear: true` and even reports 0 bytes
  cleared because the test didn't wait for output to land, since main's dispatch never
  branches on the deprecation). Passes here: the call is refused with a message
  containing `clear_output` and `removed`, and a follow-up plain read still sees the
  buffered output untouched.

## The gate

- [x] `cargo fmt --check`
- [x] `make check CARGO_SCOPE="-p stella-tools"` (guards + clippy `-D warnings`, scoped
      per AGENTS.md's `CARGO_SCOPE` — full workspace gate not run, per the task's
      verification instructions)
- [x] `cargo test -p stella-tools` — 799 + 2 + 9 + 4 + 6 + 3 + 7 + 15 = all green,
      including `docs_in_sync.rs`'s generated-count assertions (fixed two stale counts
      it caught: "These 49 tools…" → 50, "All told: 49 always-on… up to 61 native" →
      50 / 62 — both one-off from `clear_output` joining the always-on/native set)
- [x] Docs updated where behavior/flags changed (module doc, `agent-tools/index.mdx`,
      `docs/spec/remote-sandboxes.md`, `policy.rs`'s doc example)
- [x] `Closes #2699` appears both above and as a commit trailer

`cargo test --workspace` / full `make gate` were **not** run, per this task's explicit
verification scope (stella-tools only) — flagging so a reviewer knows the full
workspace gate is still owed before merge, same as CI will run it anyway.

## Nothing left behind

- [x] Filed: #2712 (`run_lint(fix)` / `format_code(check)` — same mode-flag shape,
      swept alongside this PR), #2713 (`apply_edits(dry_run)` — the issue's own named
      example of the violation shape) — **or**
- [ ] There is nothing

Sweep method is documented above; both issues are written as handoffs (files, line
numbers, constraints, definition of done) per AGENTS.md.

## Ground-rule check

- [x] N/A — no new cross-boundary protocol types

- **Deviation from the issue's literal ask**: the issue assumed `read_output` would
  become `read_only: true` after the split. I kept it `false` and explained why above
  (the plain read is itself destructive, independent of `clear`) — flagging this
  explicitly since CLAUDE.md asks me not to silently override a stated expectation.
  Happy to discuss if there's context I'm missing on why `read_only: true` was wanted
  regardless.
- **Parallel PR #2696** (scratch state plane) also touches `process.rs`'s spawn seam
  and the `registry/` registration list — kept this diff minimal and localized per the
  task's instruction; conflicts are for merge sequencing, not resolved here.
- Left `read_output`'s and `start_process`'s prose mentions of "read_output /
  send_stdin / stop_process" as-is (didn't rewrite every cross-reference to enumerate
  `clear_output` — it's a secondary, occasional-use tool, not part of the core
  start→poll→stop lifecycle those sentences describe).

## Summary by Sourcery

Split process output management into a dedicated clear_output tool and deprecate read_output’s clear flag, updating tooling, docs, and tests accordingly.

New Features:
- Introduce a clear_output tool to discard a started process’s buffered stdout/stderr without reading it.
- Register clear_output in the process tool registry and catalog so it is available as an always-on native tool.

Bug Fixes:
- Prevent silent use of read_output’s deprecated clear flag by returning a named error that directs callers to clear_output instead.

Enhancements:
- Clarify process-group tool documentation, including the destructive nature of reads and the rationale for keeping read_output and clear_output as non-read-only.
- Update policy, settings, telemetry, and remote-sandbox specs to account for the fifth process tool and revised counts of always-on/native tools.
- Refactor process tool tests into a dedicated process/tests.rs module to keep the main implementation file within size gates.

Documentation:
- Update agent-tools and configuration docs to describe clear_output, adjust process-tool counts, and align examples with the new five-tool process group.
- Refresh remote-sandboxes documentation to include clear_output in the set of process-related tools.

Tests:
- Add witness tests validating clear_output’s behavior and the deprecation error path for read_output(clear: true), alongside relocating existing process-group tests to a separate module.


macanderson:
Reviewed end to end: single-purpose split lands as specified (deprecation error before the table lock, so a refused clear cannot drain as a side effect); read_only stays false with the drain argument checked against main rather than assumed; process.rs test split respects the file-size gate; sweep follow-ups #2712/#2713 filed; cross-crate group-count test corrected to five with prose chased. CI green on head 2d8cb01. Merge-sequence note: #2714 (scratch plane) overlaps this PR in process.rs/catalog.rs/registry lists/agent-tools docs (incl. the tool count) — #2714 will rebase and re-reconcile the count after this merges.

PR #2758, opened 2026-08-10
feat(stella-tools): get_environment tool sharing prompt.rs's env probes

macanderson:
## What & why

Re-scoped remainder of #2697 after PR #2719 already shipped the byte-stable
session-environment prompt block in `crates/stella-cli/src/agent/prompt.rs`
(`append_session_environment`). This PR does **not** add a second prompt
block. It ships the tool-facing half:

1. New `crates/stella-tools/src/environment.rs`: `EnvironmentIdentity`
   (workspace root, git bit + linked-worktree bit via the same gitfile-vs-directory
   test `prompt.rs` used, `std::env::consts::OS`/`ARCH`, `uname -sr` OS release,
   login shell dialect, session scratch directory when available) plus the
   `get_environment` tool — read-only, zero arguments, single purpose: report.
   It deliberately carries **no model line**: the router owns model identity,
   and only the CLI prompt layer (which resolves the worker ref for the calls
   a given prefix rides) can state it truthfully.
2. `crates/stella-cli/src/agent/prompt.rs`'s `append_session_environment` now
   calls `stella_tools::environment::git_worktree_bits` /`os_release` /
   `login_shell` instead of keeping a private copy of the git-worktree test
   and the `uname`/`SHELL` probes. **The prompt's rendered bytes are
   unchanged** — this is a pure extraction; the format strings and branch
   conditions are untouched, only where the underlying facts come from moved.
3. Registration: `crates/stella-tools/src/registry/process_tools.rs` (the
   `uname` probe is a process spawn, same authority class as `bash`/
   `run_script`/`repo_status`) and a catalog row in `catalog.rs`:
   `"get_environment" => (true, true, Always, "environment")`.
4. Docs: `website/content/docs/agent-tools/index.mdx` (new "Environment"
   card, counts recomputed to 55 always-on / 67 native, read-only and
   speculation-safe prose lists updated) and `permissions.mdx` (read-only
   prose list). `crates/stella-tools/tests/docs_in_sync.rs` re-run and
   passes — it derives everything from `catalog.rs`, so a drift would fail
   it by name.

Closes #2697

## The witness

Stella's definition of done is a test that **fails on the old code and passes on the new**.

- [x] This PR includes a witness test (fails on `main`, passes here):
  - `crates/stella-tools/src/environment.rs::tests::get_environment_reports_workspace_root_and_platform`
    — calls `get_environment` through the tool trait directly and asserts it
    reports the workspace root and `std::env::consts::OS`. Fails on `main`
[...]

Issue #2695, opened 2026-08-10
Epic: first-class session environment — state, identity, and capabilities as single-purpose tools


macanderson:
# Spec: a first-class session environment — state, identity, and capabilities as single-purpose tools

## The critique this answers

> There is no first-class "environment" object. Processes have no lifecycle contract (services die with the agent), state has no home between steps (48MB re-parsed 12×), orientation is built on code-graph sources that null out off-repo, and capabilities are discovered by failing. Each tool holds a partial, implicit view of the world; nothing owns "what exists here, what's running, what survives me."

Audited against the tree, two of the four facets are **already covered** and are not re-implemented here:

- **Process lifecycle — exists.** `crates/stella-tools/src/process.rs` ships `start_process` / `read_output` / `send_stdin` / `stop_process` with an explicit contract: own process group, `kill_on_drop`, `MAX_LIVE_PROCESSES` cap, ring-buffered output, and group-SIGTERM→SIGKILL when the registry drops, "so nothing outlives the session." That *is* the lifecycle contract — ownership is the session, and cleanup is guaranteed. What this epic adds on top is only the single-purpose refactor of `read_output`'s `clear: true` mode flag (sub-issue D).
- **Orientation off-repo — exists.** `crates/stella-tools/src/overview.rs:125` — `render_orientation_block` is `graph_orientation_block(root).or_else(|| listing_orientation_block(root))`, so an absent/empty code graph falls back to a filesystem listing rather than nulling out.

The two facets that are **genuinely missing**, plus the object that owns them:

1. **State has no home between steps.** Nothing gives a turn a durable-within-the-session place to park a parse result, a fetched artifact, or an intermediate computation. The model's only options are re-deriving (the 48MB re-parse) or littering the user's workspace. → sub-issue A.
2. **Session identity is discovered, not given.** The model runs `pwd`, guesses shell dialect, and probes for binaries by calling them and reading the failure. Session-constant facts (cwd, platform, shell, OS, git/worktree bits) and capability presence (`rg`? `docker`? `pnpm`?) are all learnable once and owned by nothing. → sub-issues B and C.

## The owning object: `SessionEnvironment`

One struct in `stella-tools`, constructed when the `ToolRegistry` is built, shared `Arc` across the tools that need it:

```rust
/// What exists here, what's running, what survives this session.
/// Constructed once per registry; dropped with it. The scratch tempdir
/// self-deletes on drop — the "cleans itself up" contract is ownership,
/// not a sweep.
[...]

PR #2760, opened 2026-08-10
feat(stella-tools): add probe_capability tool for safe PATH lookups

### LINKED ISSUE #2698: probe_capability: stop discovering the toolbox by failing
Refs #2695 (session-environment epic, sub-issue C).

# `probe_capability`: stop discovering the toolbox by failing

Part of the session-environment epic (parent issue). **Worker: Haiku 4.5** (alt: GLM-4.6) — source provided; the job is integration and tests.

## Problem

The model learns whether `rg`, `docker`, `pnpm`, or `cargo-watch` exist by running them and reading the failure — a wasted step, a polluted transcript, and (for loop detection) a near-identical failed call that can trip the escalation machinery. Capability presence is a session-stable fact nothing owns.

## Design (decided)

One single-purpose tool, `probe_capability(name)`, `read_only: true`:

- `name` validated like a bare command word: `[A-Za-z0-9._+-]`, 1–64 chars, no separators, no spaces — this is a PATH lookup, never an execution, so it needs no `command.started` gate.
- Resolution = walk the `PATH` the child spawns actually see (compose it the way `subprocess_env` composes the child env, so the answer matches what `bash` would find) and test executability. **Never execute the candidate** — a probe that runs `--version` is a code-execution surface and is out.
- Memoized per session in the epic's `SessionEnvironment` (`Mutex<HashMap<String, CapabilityProbe>>`); if that struct hasn't landed yet, a `Mutex<HashMap<..>>` owned by the tool is fine and the epic folds it in.
- Answer names the resolved path or the absence, plus the memo status: `docker: /usr/local/bin/docker` / `pnpm: not found on PATH (cached for this session — install mid-session won't be seen until restart)` — the cache caveat is part of the honest answer.

```rust
/// One memoized probe result.
#[derive(Clone)]
pub struct CapabilityProbe {
    pub resolved: Option<PathBuf>,
}

fn resolve_on_path(name: &str, path_var: &str) -> Option<PathBuf> {
    std::env::split_paths(path_var.as_ref() as &std::ffi::OsStr)
        .map(|dir| dir.join(name))
        .find(|candidate| {
            #[cfg(unix)]
            {
                use std::os::unix::fs::PermissionsExt;
                candidate
                    .metadata()
                    .map(|m| m.is_file() && m.permissions().mode() & 0o111 != 0)
                    .unwrap_or(false)
            }
            #[cfg(not(unix))]
            {
                candidate.is_file()
            }
        })
}
```

The `Tool` impl follows the exact shape of sub-issue A's `GetState` (schema with one required string field, named errors for bad names, memo check before the walk). The tool description should say: check here before assuming a binary exists or shelling out to test it; the answer is cached for the session.

## Tests

- Witness: probe a binary guaranteed present in the test env (`sh` on unix) → resolved path; fails on `main` (tool absent).
- Absent name (`definitely-not-a-real-tool-xyz`) → named not-found answer, and a second call returns the memoized result (assert via a counter or by removing PATH between calls and seeing the cached answer survive).
- Validation: `../sh`, `a b`, empty → named error.
- PATH composition matches the child-spawn env, not the raw parent env (witness: set a PATH entry via the same mechanism `subprocess_env` honors and assert the probe sees it).

## Definition of done

`make gate` green; tool registered and toggleable; memoization landed in (or ready to fold into) `SessionEnvironment`; witness named in the PR; no lines added to god files.

### LINKED ISSUE #2732: module-reachability stripper swallows the file tail after a "/*" string literal, falsely orphaning later mod declarations

macanderson:
## What & why

Implement GitHub issue #2698: add `probe_capability` tool for safe PATH lookups without executing binaries.

The model previously discovered whether a binary existed by running it and reading the failure — wasting a step, polluting the transcript, and (for loop detection) potentially triggering escalation machinery. This tool provides a safe, memoized PATH lookup that never executes anything.

Closes #2698

## The witness

Witness tests in `crates/stella-tools/src/capability.rs`:
- `probes_sh_resolves_to_path` — resolves a guaranteed-present binary (`sh`); **fails on main** (tool absent), **passes here**.
- `absent_capability_returns_not_found` — probes an absent binary; returns cached not-found message with session-cache caveat; verifies memo-hit on second call.
- `validation_rejects_invalid_names` — validates name format (`../sh`, `a b`, empty, 65+ chars rejected); validates `probe_capability` accepts only `[A-Za-z0-9._+-]`, 1-64 chars.
- `validation_accepts_valid_names` — validates that `cargo-watch`, `g++`, `python3.12`, etc. pass validation.
- `missing_required_field` — validates error handling for missing `name` field.

Memo-hit mechanism: the test `second_call_uses_memo` verifies that a second call to the same absent binary returns the same cached result. The implementation checks `memo.contains_key()` directly after both calls to demonstrate the cache survives the second lookup.

## The gate

- [x] `cargo fmt --check` — **PASS**
- [x] `cargo clippy --workspace --all-targets -- -D warnings` — **PASS** (make check CARGO_SCOPE="-p stella-tools")
- [x] `cargo test --workspace` — **PASS** (stella-tools: 807 passed; stella-cli: 1693 passed)
- [x] Docs updated: 
  - Added tool card to Process control section
  - Updated counts: 54→55 always-on, 66→67 native tools
  - Added to read-only partition list
  - Added to speculation-safe subset list
  - Added to read-only tools in permissions.mdx
  - Fixed test in stella-cli to expect 6 process-group tools (not 5)
- [x] `Closes #2698` in commit trailers

File-size gate: **PASS** — 1295 files checked, 28 grandfathered over 1500 lines.

## Nothing left behind

- [x] No new issues to file; all work completed in this PR.

Rider from the issue: Updated the stale comment in `crates/stella-tools/src/scripts.rs::429` from present tense ("loses track") to past tense ("used to lose track") with issue reference, noting that PR #2735 fixed #2732.

## Ground-rule check

- [x] No I/O in `stella-core`; no new dependencies added.
- [x] No outbound network calls (tool is pure PATH lookup).
- [x] No cross-boundary types; uses only serde_json::Value input.

## Reviewers should know

- Tool is registered in the "process" group alongside `start_process`, `read_output`, etc.
- PATH composition uses `std::env::var("PATH")` — `subprocess_env` scrubbing doesn't mutate PATH, so this is the authoritative source for child-spawn environment.
- Memoization is per-session in a `Mutex<HashMap>` owned by the tool struct; mid-session installs won't be seen until restart (documented in the not-found message).
- Implementation follows the exact pattern of `scratch.rs` tools (`GetState` for the reference shape).

## Summary by Sourcery

Add a new read-only, speculation-safe process tool for probing PATH for binaries without executing them and wire it into the process tool catalog, tests, and documentation.

New Features:
- Introduce the `probe_capability` tool to check whether a binary is available on PATH without execution, with per-session memoized results.

Enhancements:
- Register `probe_capability` in the process tool group and export its module so it participates in the built-in catalog and policy system.
- Clarify a scripts module comment about past module-reachability issues with an updated reference.

Documentation:
- Document `probe_capability` in the agent tools docs, marking it as read-only and speculation-safe, and update the always-on and native tool counts accordingly.
- Include `probe_capability` in the documented read-only partition and permissions guidance.

Tests:
- Add unit tests covering successful PATH resolution, not-found behavior, input validation, memoization, and missing-field errors for `probe_capability`.
- Update CLI settings tests to reflect the expanded process tool group size.


macanderson:
Coordinator status: green on head a14a113 (group corrected to "environment" — a PATH probe must not die with tools.process off — and the process-group count test restored to five). Deliberately HELD IN DRAFT: #2758 is ready-to-merge and both PRs edit the catalog and the agent-tools docs count, each including only its own tool. This PR rebases over #2758 the moment it merges, re-reconciles the count, re-greens, and only then flips ready.

macanderson:
Reviewed and ready — the last PR of the session-environment epic #2695. Coordinator-verified: PATH-lookup-only (no execution path exists in capability.rs), memoized per session with a memo-hit test, name validation rejects separators, group corrected to "environment" (a no-spawn probe must not die with tools.process off), docs lists unioned with #2758 and counts recomputed to 56/68 — the off-by-one the catalog module doc predicts was caught by docs_in_sync and fixed in the merge commit. Includes the rider retensing the #2732 workaround comment now that #2735 shipped the fix. CI green + MERGEABLE on head 8282802.

Issue #2813, opened 2026-08-11
candidate_ws: a candidate killed before teardown leaves the graded tree's ref moved with no in-process record to repair it from


macanderson:
Split out of #2641. That issue's fix repairs an escaped shared ref at two
seams — `seal_inner` and `remove()` — and both are **in-process**. This issue
covers the third case: the process is gone.

## Problem

`refs_at_create` (`crates/stella-cli/src/candidate_ws.rs`, on
`GitCandidateWorkspace`) is the *only* record of what the shared ref namespace
looked like before a candidate could reach it, and it lives in process memory.
If the `stella` process is killed — SIGKILL, OOM, a container teardown, a
laptop lid — after a candidate moved `refs/heads/master` but before either
repair seam runs, the graded tree keeps the moved ref and **nothing on disk
knows what it should be**. The next `stella run` in that repository records the
already-corrupted value as its own `refs_at_create` and treats the damage as
the baseline.

This is the residue of the #2641 fix, not a regression from it: before that
fix the ref stayed moved in *every* case.

## Sketch

A durable per-candidate record — the shared-ref snapshot plus the candidate's
worktree path and baseline — written at `create_workspace` time and removed on
a clean teardown. A later `stella` invocation sweeps leftover records and
offers (or performs) the same compare-and-swap restore
`GitCandidateWorkspace::restore` performs today.

### 1. `cleanup()` destroys the natural home for the record

The obvious place to put a per-candidate record is
`<toplevel>/.git/worktrees/<name>/` — git already keeps per-worktree state
there, and it is keyed by exactly the right thing. But `cleanup()`
(`crates/stella-cli/src/candidate_ws.rs`, the `git worktree prune` at the end)
**deletes that directory**:

```rust
async fn cleanup(toplevel: &Path, dir: &Path) {
    ... worktree remove --force ...
    let _ = tokio::fs::remove_dir_all(dir).await;
    let _ = SystemGitCli.run(toplevel, &["worktree", "prune"]).await;   // <- removes .git/worktrees/<name>/
}
```

So either the record lives somewhere `prune` does not reach (e.g. under
`.stella/private/`, which is already the home for owner-only generated local
state, `0700`/`0600`), or the sweep must read it **before** `cleanup` runs —
which defeats the point, since the crash case is precisely the one where
nothing runs. Put it outside `.git/worktrees/`.

Note also that `prune` is what makes the record's *absence* ambiguous: a
pruned worktree and a never-created one look identical from the toplevel.

[...]

Issue #2812, opened 2026-08-11
candidate_ws/ref_escape: two shared-ref escapes the guard still cannot see — a deleted ref, and a rewind to an unrelated commit


macanderson:
Split out of #2641, which fixed that issue's headline (a refused escape now
restores the graded tree's ref) and explicitly left these two **detection**
gaps open. `crates/stella-cli/src/candidate_ws/ref_escape.rs`'s module doc
cites this issue by number under "Boundaries, accepted and documented".

## Problem

`GitCandidateWorkspace::escaped_refs_inner` attributes a ref move to the
candidate by signing its **new value**: either the new value is an
ancestor-or-equal of the candidate's live `HEAD` (it committed onto the shared
branch), or the candidate's private baseline is an ancestor of the new value
(its history was grafted in). Two shapes carry no such signature.

### 1. A deleted ref is not detected

A ref present in `refs_at_create` and absent at seal (`git branch -D`,
`git tag -d`, executed from the candidate worktree) has **no new value to
sign**, so the comparison in `escaped_refs_inner` — which iterates the refs
that exist *now* — never even visits it. The `refs_at_create` map already
holds everything needed to notice the disappearance; a set difference in the
other direction is a few lines.

What is missing is the attribution, and it may genuinely not exist:

- **Git deletes a ref's reflog along with the ref**, so the forensic record
  that would say *who* moved it and *from what* is gone at exactly the moment
  you want it. There is no forgery-resistant signature left to read.
- A user deleting a merged branch in their own checkout mid-run is an ordinary
  action and is indistinguishable from a candidate doing it.

So the honest outcome is probably **a distinct, non-failing warning** rather
than an abort — and almost certainly **not** a repair. Note that resurrecting
a deliberately-deleted branch is its own damage: `#2641`'s repair contract
("put back exactly what this candidate moved, compare-and-swap, never force")
does not extend here, because the precondition it relies on (a value only this
candidate could have produced) is unavailable. A repair that recreates a branch
the user meant to delete is strictly worse than the silence it replaces.

### 2. A rewind to an unrelated commit is not detected

The third probe deliberately screens out a ref that moved **backwards into
history the graded tree already had**:

```rust
if let Some(before) = before
    && is_ancestor(&self.toplevel, after, before).await
{
    continue;
}
```

That exists because `git reset --hard HEAD~1` in the user's own checkout is a
mid-run user action, and the negative control
[...]

Issue #2810, opened 2026-08-11
The last stuck-loop steer does not say it is the last one


macanderson:
## Problem

`steer_text` (`crates/stella-core/src/driver/loop_escalation.rs`) ends both of
its variants with "If you keep looping, the turn will be aborted." Since #1743
a turn has a budget of `MAX_LOOP_STEERS = 2` warnings, and that sentence now
means two different things depending on which warning it is:

- **first warning** — only re-detecting *this* loop aborts; a different loop
  earns another warning;
- **second warning** — the budget is spent, so the *next detection of any
  loop* ends the turn.

The model is told the same thing either way, so it cannot tell that its last
warning is its last. #1473 is the precedent that this wording changes model
behaviour materially (the generic "vary the arguments" prescription made a
polling model strictly worse), which is why this is worth measuring rather
than guessing.

## Proposed shape

Pass the remaining budget into `steer_text` and append a distinct final
clause when it is the last warning — e.g. "this is your last warning: the next
detected loop, whatever it is, ends the turn." Keep the existing clause for a
non-final warning.

`LoopSteerBudget` already knows the remaining budget
(`spent` vs `MAX_LOOP_STEERS`); `check_loop_detection` composes the text after
`escalate` returns, so the value is in hand at the call site with no new
plumbing.

## Definition of done

- `steer_text` takes the remaining-warnings count and its two witnesses
  (final / non-final) are unit tests beside the existing `steer_text` tests.
- `a_second_distinct_loop_earns_its_own_steer_before_the_turn_dies`
  (`crates/stella-core/src/driver/tests/loop_abort.rs`) asserts the second
  warning carries the final-warning clause.
- Measured, not asserted: a `bench/loop-bench` run comparing steps-to-resolve
  against the current wording. The nightly lane is what caught #1524's bad
  phrasing, and a wording change that is not measured there is a guess.

Refs #1743, #1473.

macanderson:
## Code and witnesses are in; the measurement is not

Branch `worktree-fix-2952-2953`.

**Done:**

- `steer_text` takes the remaining-warnings count
  (`crates/stella-core/src/driver/loop_escalation.rs`), read at the call site
  *after* the detection is charged, so it is what this warning leaves behind.
  `remaining == 0` swaps the closing clause for "This is your LAST warning: the
  next loop detected — this one or any other — ends the turn."
  `LoopSteerBudget::remaining()` is the new accessor; no other plumbing.
[...]

Issue #2809, opened 2026-08-11
Checkpoint carries no loop-steer count, so a resumed turn can re-open one spent steering warning


macanderson:
## Context

#1743 replaced the one-steer-per-turn policy with a per-loop budget capped at
`MAX_LOOP_STEERS = 2` (`crates/stella-core/src/driver/loop_escalation.rs`). The
budget lives in `LoopSteerBudget { warned: Option<LoopIdentity>, spent: u32 }`,
held by `TurnState::loop_steer` (`crates/stella-core/src/step.rs`).

## The gap

`Checkpoint` (`crates/stella-core/src/step.rs`) records the warned loop
(`loop_steered: bool` + `loop_steered_pattern` + `loop_steered_inputs`) but
**not** `spent`. `LoopSteerBudget::resumed` therefore reconciles the legacy bool
as `spent = u32::from(warned.is_some())` — one spent steer whenever any warning
was issued.

Consequences, both bounded and both deliberate as of #1743:

- **Never a refund.** A resumed turn always restores at least one spent steer,
  and the warned identity survives, so the loop the model was already told about
  still aborts on its first re-detection. This is the important half and it holds.
- **One re-opened steer.** A turn that had already spent both warnings and is
  then resumed restores `spent = 1`, so it can buy one more warning for a
  provably-different loop. Cost is one extra warning and one extra step per
  resume.

The reason it shipped this way is scope, not preference: adding a field to
`Checkpoint` is a cross-crate change. Every struct-literal construction site has
to be updated in the same PR, and they live outside `stella-core`:

- `crates/stella-pipeline/src/pipeline/tests/resume.rs`
- `crates/stella-cli/src/session_persist.rs`
- `crates/stella-cli/src/agent/resume.rs`
- `crates/stella-cli/src/resume_frame.rs`
- `crates/stella-core/src/driver/dispatch.rs`, `crates/stella-core/src/driver/drive.rs` (in-crate)

## What "done" looks like

1. Add `#[serde(default)] pub loop_steers_spent: u32` to `Checkpoint`, wired
   through `TurnState::to_checkpoint` / `from_checkpoint`.
2. Reconcile with the legacy flag rather than bumping `CHECKPOINT_VERSION`:
   `spent = max(loop_steers_spent, u32::from(loop_steered))`. A v1 checkpoint
   written before this field decodes to `0` for the new field and must still
   restore one spent steer, exactly as today. Bumping `CHECKPOINT_VERSION`
[...]

PR #2814, opened 2026-08-11
fix: P0 backlog sweep — stop the trace naming machinery that no longer exists

macanderson:
## What & why

A sweep of the open **P0** backlog. The 21 P0-labelled issues are **18 real work items plus 3 epics** (#2775, #2773, #2772) whose remaining sub-issues are all P1–P3, so the epics stay open by design.

Status: **16 of 18 landed**, 1 partially (#2545 — its blocker fixed, wiring deferred with reasons), 1 in progress (#2540/#2569). One issue (#2204) was found **already fixed on `main`** and closed with evidence rather than reimplemented.

Most of these share one theme: **the trace described machinery that no longer exists, and something downstream read it as fact.**

### Landed

| Issue | Fix |
|---|---|
| #2619 | `evidence_demand_prompt` opened by telling the worker "Your work was reviewed and looks correct" on a path where `ModelCallRole::Verdict` is unassignable. Not a stale comment — it is pushed into the worker's context as a `RevisionCause`, so it reads as an instruction, and it points the worker away from re-examining the change at the moment nothing backs it. `verify.rs`'s own docs argue against exactly this: on `fix-git` a reviewer's unsubstantiated claim made a worker reset `master` and destroy a recovered commit, twice. |
| #2556 | `flip_achieved: bool` could not tell "no command was ever tracked" from "a tracked command did not flip" — both serialised `false`, so bench scoring reproduced the false negative #2531 had removed from the prompt. Replaced with a `FlipOutcome` tri-state. |
| #2635 | `stella calibration`'s "model verifier" cohort counts unproven passes and no model is involved in any of them. Renamed `verifier_*` → `unproven_*` through the report, the JSON keys, the human output and the docs. |
| #1743 | A model that obeyed the loop steer and formed a *different* loop was killed unwarned. `LoopSteerBudget` grants a second warning only when it can prove a different loop, capped at 2. |
| #2570 | A cancelled execution reported only its pre-cancel cost prefix (0.87 USD of a 5.29 USD run). `executions.cost_usd` is now a floor over the receipt ledger, recomputed rather than incremented. |
| #2641 | A refused ref escape left the graded tree's ref moved while the message said "Nothing was restored". Now performs the repair as one CAS transaction. |
| #2207 | The #863 vacuity screen was a no-op for `.sh` — the shape #2064 made the floor runner. New `witness/density/shell.rs`. |
| #2206 | The worker was told to **paste** the project root onto the candidate root. Worse than the issue says: the notice stated a mapping and the mapping was wrong. |
| #2272 | The authored witness's failing baseline was never recorded on the wire at all. |
| #2038 | `.stella/private/` sidecars reached the verifier diff and the warrant, defeating a `DocsOnly` waiver. |
| #2654 | `arenabench run` wired an ambient `CLAUDE_CODE_OAUTH_TOKEN` into an undeclared seat; the server path refused it. |
| #2325 | `arenabench serve` accepted Stella seats it could not launch. |
| #2329 | A killed match left its containers running, and the next match's trials were scored as agent losses. |
| #2374 | **Not a code defect** — a 14-feature measurement programme whose deliverable is verdicts written into the issue. No code invented for it. |
| #2204 | **Already fixed on `main`** — closed with a clause-by-clause evidence table. |

### Three issues whose own diagnosis was wrong, corrected here

Worth a reviewer's attention, because in each case the code disagreed with the issue:

- **#2272** claims the authored baseline "lands in `oracle_trace` as a *candidate* false" and is "the leading ✗ in every real trace". It does not land there at all — a pre-fix run records `[Candidate/true, Candidate/true]` and zero baseline proof steps.
- **#2206** claims nothing told the worker what the candidate root maps to, "so it guessed". It did not guess: it was told `` `/x/y` -> `{root}/x/y` ``, which is correct only when the session root is `/`.
- **#2654** claims an undeclared seat "falls through" the CLI's refusal entirely. It does not — the refusal fires; the real hazard was that its remediation advice stopped working.

### A defect this PR introduced, caught in review, and fixed

The `FlipOutcome` rename is compile-checked across the Rust crates, so every Rust consumer moved with it. **arenabench is a JSON consumer of the same contract and is not compile-checked**: `proof.py` did `bool(ladder.get("flip_achieved"))`, which after the rename returns `None` → `False`, so **every run would have scored as no-flip, including a genuinely flip-verified one.** That is the exact bench-scoring corruption #2556 exists to prevent, arriving from the opposite direction. Flagged by a review bot, confirmed against the source, and fixed by teaching the consumers the tri-state rather than re-emitting the lossy bool. The underlying gap — nothing pins arenabench's reads against `docs/wire/` — is filed as #2820.

Refs #2619, #2556, #2635, #1743, #2570, #2641, #2207, #2206, #2272, #2038, #2654, #2325, #2329, #2545, #2374, #2584, #2531

## The witness

Every witness had its **failing direction confirmed** by editing the change out and back — never `git stash`/`checkout`, matching the `verify_done` discipline. Selected evidence:

| Issue | Witness | Observed failure on old code |
|---|---|---|
| #2619 | `the_evidence_demand_claims_no_review_of_the_change` | asserts the *absence* of any claim about who looked at the work, so rewording stays free |
| #2556 | `a_witness_that_never_ran_does_not_serialise_as_one_that_failed` | both states serialised identically |
| #2635 | `the_unproven_cohort_never_claims_a_model_verifier` | rendered output still named `model verifier` |
| #1743 | `a_second_distinct_loop_earns_its_own_steer_before_the_turn_dies` | `left: 1, right: 2` |
| #2570 | `a_cancelled_execution_reports_every_receipt_it_paid_for` | `reported $0.86, receipts prove $5.29` |
| #2641 | `a_refused_stash_escape_leaves_the_users_stash_list_untouched` | fails pre-fix **and** against a naive `update-ref` repair, with the predicted phantom entry |
| #2207 | 6 of 9 shell-density tests | `left: Ok(()), right: Err(NoAssertions)` |
| #2206 | `the_candidate_is_told_to_strip_the_session_root_not_paste_it` | a first, weaker assertion passed against the old rule; strengthened until it genuinely fired |
| #2038 | `stella_state_churn_does_not_defeat_a_docs_only_waiver` | `left: Required, right: NotRequired(DocsOnly)` |
| #2556 (bench) | `test_a_new_record_reads_as_achieved` | `- unobserved / + not_achieved` |

**Tests modified rather than deleted** — `check-deleted-tests` runs on `pull_request`; **zero tests were removed** (arenabench audited by full test id: 767 → 803, none removed):

- `settled_cost_sums_the_receipts_of_an_execution_that_never_closed` — asserted `cost_usd == 0.0`, which *was* the defect.
- `a_fresh_loop_on_a_different_tool_is_not_blamed_on_the_warned_one` — encoded the old abort-on-second-detection policy; replaced by `a_third_loop_is_not_steered_and_is_not_blamed_on_the_warned_one`, keeping every #1524 assertion one rung up.
- `the_refusal_names_a_created_ref_with_a_delete_recovery` → `the_refusal_names_a_created_ref_it_deleted`; `the_refusal_quotes_the_value_recorded_at_creation` → `the_refusal_reports_the_restore_it_performed`, with a third case split out.
- `a_mid_run_user_commit_is_not_a_ref_escape` / `..._rewind_...` — strengthened with post-seal *and* post-`remove()` assertions.
- `a_stream_without_snapshots_leaves_the_verifier_alone_rate_unmeasured` → `..._leaves_the_uncorroborated_rate_unmeasured`.
- `test_the_cli_run_path_fills_from_the_local_claude_login` — was a grep for a string that survives only in a docstring after the refactor, so it passed while proving nothing; now behavioural.
- `LadderInputs`/`LadderSnapshot` fixtures across 11 files. Two **gained** coverage: a proptest over `any::<bool>()` now ranges over all three outcomes, as does the calibration projection test.

## The gate

- [x] `make gate` — **exit 0**, run in full locally
- [x] `cargo fmt --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test --workspace`
- [x] arenabench: `pytest` → **797 passed, 6 skipped**; `tsc --noEmit` clean; `ruff` findings unchanged from baseline (diffed against a clean `git archive` extraction)
- [x] `docs/wire/` regenerated (`make wire-schema-update`). `FlipOutcome` joins the wire-contract vocabulary table, so all three arms must reach a sample or the guard fails.
- [x] Docs updated where behaviour changed, including `website/content/docs/commands/calibration.mdx`

The **first CI failure on this branch was real and is fixed**: a broken intra-doc link (`crate::roster::ModelCallRole` → `stella_protocol::ModelCallRole`) that `doc-warnings` catches under `rustdoc -D warnings`.

## Nothing left behind

- [x] Filed: #2807, #2808, #2809, #2810, #2812, #2813, #2820, #2821, #2822, #2823, #2824, #2828

Notable ones a reviewer should weigh:

- **#2820** — nothing pins arenabench's JSON reads against `docs/wire/`. This is the gap that let a compile-checked rename break scoring silently.
- **#2823** — a **fourth** consumer of #2038's root cause, found and not fixed: `validate_witness_artifact` requires the untracked delta to be exactly one path and its snapshots are unfiltered, so any `.stella/` write in the authoring window fails a valid witness non-degradably. Honestly labelled: the rule and the unfiltered snapshots are verified; whether a writer actually fires in that window is **not**.
- **#2824** — an invariant #10 gap created *by* the #2272 fix: the baseline/candidate distinction is now on the wire and nothing grades on it.
- **#2808** — #2570's delivery-recording clause. **This is why #2570 is `Refs`, not `Closes`.**
- **#2809** — the loop-steer count is not on the `Checkpoint` wire; a resume re-opens one warning (cost: one step). Bounded and documented, not silent.

## Ground-rule check

- [x] No I/O added to `stella-core`; no new dependencies anywhere
- [x] New cross-boundary types round-trip through serde — `FlipOutcome` has round-trip and legacy-parse tests and joins the wire-contract sample sweep

**Exemplars named** (CLAUDE.md asks for this): `FlipOutcome` follows `LadderSnapshot::diff_coverage`, a token rather than an `Option<bool>` on the identical argument — "the third value is the whole point". `witness/density/shell.rs` follows `witness/airlock.rs` + `airlock/tests.rs`, same directory and boundary.

**Rejected alternatives, deliberately:**
- *#2556 option 1* (emit a `flip_state` token **alongside** the bool). Leaves one fact in two fields that can disagree — the failure mode AGENTS.md describes for the file-size baseline. The rename costs a compile error or an `undefined`; the additive option costs a silent disagreement later.
- *#2272 option (b)* (declare authored flips candidate-only). Factually wrong about the mechanism — the baseline *is* a genuine pre-execution tree — so it would enshrine a weaker claim than the code earns, and cost a wire change.
- *#2206 direction 2* (normalise at the file-tool boundary). A resolution change across every surface, and it could not have helped the observed failure, which was `bash: mkdir -p` — no path argument to normalise.
- *#2325's proposed probe* (test the arena's own interpreter). Wrong in both directions; the seat is launched by the `harbor` console script, so its **shebang** is what matters.
- *#2038 in the adapter*. "What counts as the agent's work" is a decision; invariants #1/#2 put decisions in the engine plane.

**A hazard introduced and then removed during the work** (#2325): computing the seat verdict lazily inside `/api/health` would have blown `probe_arena`'s 0.6s budget on a cold probe, so arena discovery would have reported nothing listening and two arenas would have driven one workspace. The memo is warmed before the socket binds, and a test pins it.

**Unmeasured, and stated as such:** #1743's definition of done asks for a `bench/loop-bench` re-run, which needs a provider key. The extra steer's worst case is one additional warning per turn; **whether it pays for itself in resolve rate is not measured here** and should be before this is called done.

**Still in flight:** #2569 (an unverified run exits 0 and records `completed`) and #2540 (a repo-global witness predicate counts the pipeline's own snapshot commits). Both touch the wire contract, so they are being done together. #2569's exit-code half is deliberately being designed as an opt-in rather than a default flip: this repo's own default has no `--test-command`, so "unverified" is the common case, and silently changing `stella run`'s exit code would break every caller — the reporting surfaces become honest regardless.

**#2545 is partially done and says so.** The reauth supervisor stays unwired — it needs a per-contestant supervisor, a task-subset `_launch`, a changed `_await_completion` exit condition and `PAUSED` on the snapshot, and it runs through the scoring path with no way to verify it here. What *is* fixed is the design question the issue says to settle first, which turned out to be a live bug rather than the "arguably latent" one described: `Match.trial_dirs` collapsed attempts first-sorted-wins, so with `--n-attempts > 1` a seat whose first attempt voided on auth and whose second solved the task with reward 1.0 reported **no result at all**. A measured attempt now outranks a void — and two *measured* attempts rank equal, so a win never replaces a measured loss, which would be re-running until it passes and reporting the pass.


macanderson:
Confirmed — this is a real defect in the change, and a scoring-relevant one. Thank you.

The tri-state rename is compile-checked across the Rust crates, so every in-tree Rust consumer moved with it. **arenabench is a JSON consumer of the same wire contract and is not compile-checked**, so it went unnoticed:

- `arenabench/arenabench/proof.py:445` — `proof.flip_achieved = bool(ladder.get("flip_achieved"))`. On a new record the key is absent, `.get` returns `None`, and `bool(None)` is `False`. So **every** run scores as no-flip, including a genuinely flip-verified one.
- `proof.py:316-318` branches on that value for proof classification, so the classification degrades silently.
- Also `proof.py:178`/`:278`, `transcript.py:703`, `ui/lib/types.ts:275`, and `ui/components/arena/proof-panel.tsx:154` / `proof-badge.tsx:112`.

That is exactly the failure #2556 exists to prevent — a bench number that silently disagrees with the trace — arriving from the opposite direction.

**Fix being applied:** the consumers will read the tri-state `flip` rather than the Rust side re-emitting the lossy bool. The bench is the consumer #2556 was written for, so it should gain the `unobserved` / `not_achieved` distinction rather than be shielded from it — in particular `proof.py:316-318` must stop treating "nothing measured it" as a negative signal about the work, and the UI must stop rendering an unobserved flip with the same warn styling as a real negative. A legacy fallback to `flip_achieved` stays for the recorded fixtures under `arenabench/tests/fixtures/matches/**`, which must keep parsing. Witness: a fixture carrying `"flip": "achieved"` reads as achieved (fails on today's arenabench code), plus a legacy-fixture test.

One thing I'd push back on: the suggested framing is that the *removal* of the key is the bug. Keeping `flip_achieved` with a changed type would be worse — a consumer doing `if snapshot.flip_achieved` would read a truthy `"not_achieved"` and get a false **positive**, which fails toward "verified" instead of away from it. The rename is the right direction; the missing half was migrating the non-Rust consumers, which is what this fixes.

Separately, I'm filing the underlying gap: there is nothing mechanically pinning arenabench's JSON reads against `docs/wire/agentevent.schema.json`, which is why a compile-checked rename could still break scoring.

---

macanderson:
CI status and the blocker, so it is visible rather than inferred.

**The failing step is #18, `no test disappears in the merge unacknowledged`** (`scripts/check-deleted-tests.sh`). Every other step passed on `c922cf3`, including the ones that would catch a real defect in this diff:

| Step | Result |
|---|---|
| `cargo fmt --check` | ✅ |
| `docs/wire/ still matches the types` | ✅ |
| `cargo doc -D warnings` | ✅ (the earlier red — a broken intra-doc link I introduced — is fixed in `4f4913a`) |
| prompt-cache golden fixtures | ✅ |
| `cargo test` | ✅ |
| release smoke build (thin LTO) | ✅ |
| **`check-deleted-tests`** | ❌ |

**Diagnosis: the branch is stale, and it now conflicts.** It is 6 commits behind `main` (through #2826, #2830, #2788 and three release syncs). Computed without touching the working tree, via `git merge-tree --write-tree origin/main HEAD`:

```
CONFLICT (content): Merge conflict in crates/stella-core/src/driver.rs
Auto-merging crates/stella-core/src/driver/tests.rs
Auto-merging crates/stella-pipeline/src/pipeline.rs
Auto-merging crates/stella-pipeline/src/pipeline/witness_stage.rs
```

The conflict is #2826's mid-turn provider fallback against this PR's `LoopSteerBudget` change, both in `driver.rs`. That is exactly the shape `check-deleted-tests` exists to catch, so I am treating its complaint as real until the merge is resolved and it is re-run against the merge result — not as noise to wave through.

A caveat on my own evidence, since it would otherwise read stronger than it is: running the guard **locally** as `./scripts/check-deleted-tests.sh origin/main` compares main's tip against my branch tip, not against the merge result, so it reports every test `main` gained after my branch point as "missing". That list (`the_upstream_pin_flag_parses_an_order`, `every_running_handle_is_named`, `goal_loop_iterates_until_the_verifier_passes_it`, …) is an artifact of the local invocation and is **not** the CI list. I have not yet read CI's actual list — the job-log API truncates to the tail and step 18 ran at 05:09:23. The merge will settle it either way.

**Why I have not pushed the merge yet.** Three agents are mid-flight in `crates/stella-pipeline`, `crates/stella-cli` and `arenabench/` with uncommitted work, and `git merge` would either refuse or clobber it. Merging into a tree someone else is writing is how the #1976 class of silent deletion happens in the first place. The sequence is: land the in-flight work → merge `origin/main` and resolve `driver.rs` by hand → re-run `make gate` in full → push. `make gate` passed at exit 0 on `4f4913a` before the arenabench commit, so the Rust side is green on its own base.

**Also fixed since this run:** the tri-state `FlipOutcome` broke arenabench's JSON readers (`proof.py` did `bool(ladder.get("flip_achieved"))`, which after the rename returns `None` → every run scores as no-flip). Flagged by the Vercel review bot, confirmed against the source, fixed in `c922cf3`, and the underlying gap — nothing pins arenabench's reads against `docs/wire/` — is filed as #2820. That bot comment was correct and worth the round trip.

Note for reviewers: Sourcery declined this diff as over its 150k-character review limit, so it has had no automated line-level review. The per-commit messages carry the reasoning; the PR body carries the witness table and the rejected alternatives.

---

macanderson:
Status on the current red, so it is not mistaken for something unexamined.

**`file size ratchet` is failing, it is mine, and it is expected.** `arenabench/arenabench/runner.py` is over the 1500-line limit. Traced rather than guessed:

| Commit | `runner.py` |
|---|---|
| `809c523` (branch base) | 1152 |
| `c922cf3` (#2329 reap work) | 1214 |
| `35e20df` (#2545 reauth checkpoint) | 1565 |
| `e837ed4` (split in progress) | 1529 |

So the in-flight #2545 reauth wiring crossed it, and the split is underway but not finished. `runner.py` is **not** grandfathered, so the only legal fix is to move the wiring into a sibling module — `reap.py`, extracted during the #2329 work, is the precedent sitting right beside it. I have not touched `scripts/file-size-baseline.txt` and will not: adding a row there, or running `make file-size-update`, is the expedient CLAUDE.md treats as a defect against the PR that introduces it, and the baseline exists for files that predate the guard.

**The last three commits are labelled `wip:` and are checkpoints, not claims.** The container is ephemeral and a stop-hook requires pushed work, so I am trading tidy history for not losing two agents' output. Each one names what is red in its own message rather than implying otherwise, and `e837ed4` explicitly records that the suite was *not* re-run over its state — the files were being written as it was committed, so any number would have described a tree that no longer existed.

**Also still open, and possibly worse than the ratchet:** `test_the_launch_refuses_the_seat_rather_than_exiting_zero` and `test_starting_a_match_sweeps_the_ones_already_over` fail with `HarborUnavailableError`. Both passed on this machine before the reauth edit, and this machine has never had a `harbor` binary — so the reading that fits is that the wiring routes `_launch` and `MatchRunner.start` at the real `harbor_bin` where a stub previously satisfied them. The first is the standing witness for **#2325**, a P0 that already landed, so if that is right it is a regression in shipped verification rather than a fixture inconvenience. I have not read the diff that caused it and have asked its author to confirm or refute rather than accept my reading.

I am deliberately not fixing either myself while an agent holds those files — a concurrent edit there is how the duplicate-work problem that produced (and then closed) #2849 happened in the first place.

The PR stays **draft** with auto-merge disarmed until the ratchet is green, those two tests are accounted for, the witness tests for #2569/#2540 are verified in their failing direction, and `make gate` has run over the merged tree. None of that is true yet.

---

macanderson:
Two blockers resolved, and the #2569/#2540 detail I was previously describing without having read it.

## The file-size ratchet is green

`arenabench/arenabench/runner.py` is back under the limit at **1415 lines** (`check-file-size`: OK — 1318 files, 27 grandfathered, nothing went over by this change). Fixed by splitting the #2545 reauth wiring into `reauth_wiring.py` (181 lines), following `reap.py` — extracted from the same file during the #2329 work — as the precedent.

`scripts/file-size-baseline.txt` was **not** given a new row, and `make file-size-update` was deliberately not run: it retightens unrelated files' ceilings, which is the shared-cell edit that has reddened `main` three times before. One baseline line *was* removed by hand — `crates/stella-cli/src/fleet_cmd.rs` fell to 1498 and left the god-file list, with `AGENTS.md` and `crates/stella-cli/README.md` updated to match (`make god-files` cross-checks all three). A hand-edited generated file is normally a red flag, so: that is the direction the ratchet is allowed to move.

## #2569 — the default exit code is unchanged

I had been describing this as an opt-in without having seen the work. Confirmed, and it is:

- **`stella run` still exits 0 on an unverified run.** No user-visible break.
- The *reporting* surfaces are now honest: new `PipelineStatus::Unverified { verdict }`, `pipeline_status_label` → `"unverified"`, `pipeline_failure_reason` → the verdict summary, `pipeline_episode_outcome` → new `EpisodeOutcome::Unverified`, plus the deck recap, resume label and fleet ledger.
- The exit code moves only under `stella run --require-verified` (`VerificationRequirement::Advisory` default / `Required` opt-in). The flag is named for the property demanded, not the mechanism.

The reasoning is recorded in `VerificationRequirement`'s doc comment rather than only in a PR: with no `--test-command`, unverified is the *ordinary* outcome — this repository's own default reaches it — so flipping the default would redden every script and CI job shelling out to Stella on runs whose behaviour did not change by one byte. That is a release-note decision for a maintainer, and #2839 tracks it along with the fact that `--require-verified` currently exists only on `run` and not on `goal`/`fleet`/`arena`/`serve`.

**Five existing tests asserted `PipelineStatus::Completed` on runs nothing had proved.** Those five *are* #2569's witness — the defect was already sitting in the suite, asserted as correct.

## #2540 — one deliberate deviation, flagged rather than done quietly

`LadderRung::WitnessUnsatisfiable` is on the wire, keyed on `FailureFingerprint` equality, ordered as step 0 above `Revise`. Two things worth a reviewer's attention:

1. **The trigger additionally requires a revision to have been spent.** The brief said fingerprint equality alone; that is wrong, and the reason is sharp: the witness stage rejects a witness whose baseline passes, so *every* armed witness fails on the baseline — "red on both trees" is true of every failing witness and discriminates nothing. Requiring a revision makes it an experiment: the work varied in response to the failure and the failure did not move. With the immediate trigger, `witness_isolation::a_revise_turn_that_edits_the_witness_hard_fails_at_the_next_check` regressed, because the run never reached round 2 and the tamper check never fired. Cost is one revision; tracked in #2840.
2. **The `pipeline@stella.invalid` prompt sentence already existed** (`MACHINERY_COMMIT_NOTE`, landed in #2643). Nothing was added there — but its doc listed `git fsck --dangling` among the enumerations that see the harness commits, and that is false: worktree refs and worktree HEAD are reachability roots, so `fsck` reports nothing. `git rev-list --all --not master` is what sees them. Re-verified in a scratch repo on git 2.43.0 and corrected in code, since the same wrong claim is in #2540's body.

## Still not green, and why

- **`harbor_adapter + analyzer pytest` fails on `test_a_lock_whose_holder_died_is_reclaimed`.** Expected: I reverted my own fix for it because #2846 supersedes it with a better one (`flock --no-fork` rather than a process-group kill, which avoids a race mine reintroduced). The branch carries `main`'s broken test until #2846 lands and I merge `main` again. Not something to fix here.
- **`make gate` has not been run on the merged tree.** The agent's exit 0 predates the `origin/main` merge and the five commits after it, so it does not describe this tree and I am not reporting it as if it does. Two agents are still writing, and a suite run over a tree being edited produces a number about a tree that no longer exists.

New issues from this work: **#2839**, **#2840**, **#2841**, plus a load-control comment on **#2828**.

---

macanderson:
On the suggestion that `decide()` re-pauses a freshly re-dispatched seat because a stale prior-round void keeps `voided()` truthy — **this one is already handled**, and I checked rather than assumed, because the same reviewer was right about the `flip_achieved` break earlier in this PR.

The scenario is described verbatim in the code, at `TrialView.superseded` (`arenabench/arenabench/reauth.py:152-159`):

> Whether this trial predates the credential currently dispatched.
>
> A void never leaves the disk, so without this the *first* tick after a successful re-dispatch reads the old void, finds the keychain offering nothing newer than the token it just launched with, and pauses — killing the retry seconds after starting it. The trial is still counted as evidence of what has been *measured*; it is no longer evidence about whether the live credential works.

The mechanism, all three parts present at the reviewed SHA `a1170d8`:

1. `voided()` (`:244`) filters on it — `dead = {t.task for t in self.trials if t.credential_void and not t.superseded}`, with its own docstring making the same argument: *"Counting an already-answered void here makes a healthy re-dispatch look like a fresh failure."*
2. `_advance` (`:716`) stamps the generation **after** a successful dispatch — `seat.superseded = frozenset(view.trial for view in views)` — with the comment noting it is recorded after the launch rather than before, so a dispatch that raised leaves the generation untouched.
3. Every tick re-stamps the views through it (`:698`).

So the in-flight retry window is exactly what `superseded` exists for. No change needed for this.

**What the suggestion does usefully surface is the adjacent question, in the opposite direction.** `superseded` is keyed on `view.trial` — the trial identifier — and #2545's re-dispatch reuses the *same job dir*. If a retry can ever produce a trial whose identifier collides with a superseded one, that new trial inherits `superseded=True` and is then excluded from `voided()`, so a genuinely dead *new* credential would fail to pause. That is under-pausing rather than over-pausing, and it is the more dangerous polarity: over-pausing is visible and annoying, under-pausing silently burns the retry budget against a token that will never work.

I have not established whether that collision is reachable — it depends on the trial-naming scheme under re-dispatch, which is being written right now. I have routed it to the author rather than guessing, and I am explicitly labelling it as a question, not a finding.

Worth noting the reviewer is 1-for-2 on substantive catches in this PR and both were worth the round trip: the `flip_achieved` one was a real scoring-corruption bug I introduced, and this one, while already covered, pointed at a real neighbouring hazard.

---

macanderson:
**`make gate` has now run on the merged tree** — the one thing I have been saying was missing. Result, stated precisely rather than as "green":

**Every guard passes.** All 27 gate steps, including the ones that matter for this diff:

```
check-wire-schema:  OK — docs/wire/ matches the types.
check-file-size:    OK — 1319 files, 27 grandfathered, nothing went over by this change.
check-god-files:    OK — 19 god file(s) across 7 crate(s), named identically in AGENTS.md and every crate README.
check-invariants:   OK — 10 invariants, one normative home, 28 citation(s) resolve.
check-left-behind:  OK — 1319 source file(s) scanned, 0 grandfathered.
check-doc-links:    OK — 317 citation(s) resolve across 69 identified document(s).
check-diagnostic-codes: OK — 43 codes emitted, all documented.
```

`check-wire-schema` passing is the load-bearing one: it confirms the `FlipOutcome` and `LadderRung::WitnessUnsatisfiable` regenerations in `docs/wire/` actually match the types, rather than my having asserted it.

**Tests: 1711 passed, 1 failed.** The single failure is `daemon::tests::a_run_that_ignores_term_is_killed_after_the_grace_period` — the load-dependent flake tracked in **#2828/#2393/#2804**. Control run, which is what distinguishes a flake from a regression:

```
$ cargo test -p stella-cli --bin stella daemon::tests::a_run_that_ignores_term
test result: ok. 1 passed; 0 failed; ... finished in 8.28s
```

Passes alone in 8.28s (it needs 8s of wall clock), fails inside the full parallel suite. Nothing in this diff touches `daemon.rs`, `signals.rs`, or process spawning, and the #2569 agent independently confirmed it fails under load generated **exclusively by crates this change does not touch** — a discriminating control, not an assertion.

So: **`make gate` exits non-zero, and the reason is a known pre-existing flake, not this change.** I am not calling that green, and I would rather state it this way than report an exit code that flatters the branch.

**The two `HarborUnavailableError` failures I flagged earlier are resolved.** Verified myself rather than taking the agent's word: `pytest tests/test_adapter.py tests/test_reap.py` → 33 passed. That includes `test_the_launch_refuses_the_seat_rather_than_exiting_zero`, the standing witness for #2325, so the regression I was worried about did not survive.

**arenabench full suite: 845 passed, 6 skipped, 1 failed** — the netns test, red on purpose because I reverted my inferior fix in favour of #2846. It stays red here until that lands on `main` and I re-merge.

**#2519 and #2520 are done.** The harness fold is now a dialect table plus a per-arm fold with the byte cursor factored out, so a third arm is a row and a fold rather than an edit to the comparison logic. #2520 distinguishes three usage states (`stream` / `transcript` / `unobserved`) rather than two, following the same discipline as `FlipOutcome`: a counter nobody has published reads as "unknown, not zero", never as `0`. Witnesses: 8 failed with the change out, 33 passed with it in.

New issues: **#2850**, **#2851**, **#2852**, **#2853** (the last is a pre-existing hazard found and deliberately not fixed — an unanchored substring fast path in the Claude Code fold that swallows any line containing `"thinking_tokens"`; the new Stella fold refuses to copy that shape and its docstring says why).

Still open before this is reviewable: #2545's reauth wiring is still being written, and the seven `wip:` checkpoint commits need collapsing into coherent history.

---

PR #2861, opened 2026-08-11
release.minor: cut the 0.9.0 line — stamp 0.9.0 and write its releases-page section

macanderson:
## What

Cuts the 0.9.0 minor line.

- **`release.minor:` in the title is load-bearing.** `auto-tag.yml` computes the
  next version from the newest tag (`v0.8.44`), *not* from `Cargo.toml`, and
  reads the bump marker from the subjects of every commit since that tag. A
  squash merge makes this PR title the commit subject, so this is what makes the
  next release `v0.9.0` — after which the ordinary per-merge patch bumps run
  `0.9.1`, `0.9.2`, … off the corrected minor.
- **`scripts/sync-versions.sh` stamped 0.9.0** — `[workspace.package].version`,
  the workspace-member entries in `Cargo.lock`, and the build-from-source
  Homebrew formula. Running the canonical script rather than hand-editing four
  places keeps this byte-identical in effect to the release commit `auto-tag`
  will cut.
- **`CHANGELOG.md` gains the `[0.9.0]` section**, covering the whole 0.8 line
  (123 non-release commits, `v0.8.0..HEAD`). This is the "release PR that
  deliberately writes its own section" exception in `CHANGELOG.md`'s contributor
  note: `changelog-roll.sh` is idempotent and leaves a version that already has
  a section alone, so the roll at release time will neither duplicate nor
  overwrite it.
- **That section *is* the website release page.** `/releases` is a statically
  rendered server component over `getReleases()`, which parses `CHANGELOG.md` at
  build time — there is no second copy under `website/` to keep in sync, which
  is exactly the drift the hand-maintained MDX page it replaced suffered.

## Witness

`website/src/lib/changelog.0-9-0.test.ts`, following the
`changelog.0-8-0.test.ts` pattern, plus one assertion that file cannot carry:
that `[0.9.0]` is `releases[0]`. `/releases` takes the first parsed section as
the current line, so a hand-written section appended at the bottom of
`CHANGELOG.md` parses cleanly and still renders the *previous* line as current.

- Fails on `main`: `git show HEAD:CHANGELOG.md` contains no `## [0.9.0]`
  heading, and the parser is the test's only input.
- Passes here: `npm test` from `website/` — 18 passing, 0 failing.

`make guards-fast` green (27 gate steps + `cargo fmt --check`). No Rust source
changed.

## Deleted tests

None.

## Summary by Sourcery

Stamp the 0.9.0 release and ensure it is treated as the current line across changelog and website.

New Features:
- Add a detailed [0.9.0] section to CHANGELOG.md covering all changes since 0.8.0 and structured for the releases page.

Enhancements:
- Update workspace and Homebrew packaging metadata to version 0.9.0 so builds and tags align with the new minor line.
- Add a website test that parses CHANGELOG.md to verify the 0.9.0 section is non-empty and that the releases page leads with 0.9.0.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,121 @@ are not user-facing and do not appear here.
 
 ## [Unreleased]
 
+## [0.9.0] — 2026-08-11
+
+Everything since 0.8.0 — the whole 0.8 line. Verification stopped buying model
+calls and started ending the turn the moment a proof lands; a turn now survives
+the provider failing underneath it instead of dying with it; and the tool
+surface grew a scratch state plane, environment probes, and an interactive
+approval/skill/hook layer a host process can drive.
+
+### Added
+
+- **Verification halts on proof.** Stop-on-proof kills in-flight tools the
+  moment the oracle flips, `verify_done` fires the halt rather than merely
+  being tallied, and task mode has to earn its declaration (#2662). A confirmed
+  `verify_done` now carries a deterministic pass (#2652).
+- **A turn survives its provider.** Mid-turn model fallback re-resolves through
+  the router when retries are exhausted, repairs the transcript, and continues
+  the turn (#2769) — wired into the pipeline in #2826. Alongside it: a
+  streaming→non-streaming fallback for hung or empty streams with a first-byte
+  deadline (#2686, #2748), reactive recovery from provider context-overflow
+  errors (#2680, #2752), a retry ladder that parks under sustained rate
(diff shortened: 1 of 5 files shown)


macanderson:
Refs #2862 — the `[0.9.0]` section is drafted from `v0.8.0..20fd1178c`, and
`changelog-roll.sh`'s idempotency guard means commits landing between this PR
and the `v0.9.0` tag will not be added to it. Filed as a handoff rather than
fixed here; per-tag GitHub Release notes still cover those commits.

Issue #2872, opened 2026-08-11
pipeline: a witness that is broken as an instrument reads as a failing assertion — shell parse errors arm the oracle, and fail-on-both-trees routes to Revise


macanderson:
## Problem

Two failures from the 2026-08-11 Terminal-Bench 2.1 forensics (SUT `main@2193374`) that
share one root cause: **a witness that is broken as an instrument is read as a witness that
is failing as an assertion.** Split out of #2871, which fixes the `verify_done`-side half
(cwd escape, `VACUOUS` message, empty-tree baseline); this issue is the pipeline-side half.

### (d) An oracle that fails on BOTH trees is a broken instrument, not a failing candidate

- `f89p2/sqlite-with-gcov` authored `sqlite3 --version | grep -q 'gcov'` — never true on
  any tree, baseline or candidate.
- `f89p3/build-pov-ray`'s witness failed on baseline **and** candidate on a bad include
  path.

`ladder_decision` (`crates/stella-pipeline/src/verify.rs:763`) routed one of these to
`LadderDecision::Revise`, sending the worker to fix code that was not broken.

`LadderDecision::WitnessUnsatisfiable` already exists for exactly this shape (#2540), but
its trigger — `LadderInputs::witness_unmoved_by_revision` — requires **a revision already
spent** plus `airlock::FailureFingerprint` equality. That condition is correct for the case
it was designed for (a witness that is deaf to the change), and it is deliberately
conservative. It does **not** cover the case here: a witness whose command can never
succeed on any tree, which is decidable on the *first* observation and should never cost a
revision at all.

The reasoning that made the revision mandatory is on the record in that doc comment and
must be respected: "red on both trees" is true of every armed witness by construction, so
it discriminates nothing on its own. Any fix here therefore needs a signal that is
*stronger* than "failed twice" and available *before* a revision — the candidate for that is
the interpreter/shell-level defeaters below, not a bare pair of non-zero exits.

### The baseline-fail precondition screens only infra, not a broken interpreter

`crates/stella-pipeline/src/pipeline/witness_stage.rs` screens the armed witness's baseline
run through `CmdOutcome::infra_label()` (`crates/stella-pipeline/src/ports.rs:558`), whose
whole vocabulary is `TimedOut | OutOfMemory | Infra`. A witness that fails on the baseline
because the **shell could not parse it** therefore reads as a legitimate assertion failure
and arms the oracle.

[...]

Issue #2668, opened 2026-08-10
P2: verify_done is a catch-22 on creation tasks — VACUOUS forever, 30-70% budget burn, and it motivated stopping the graded server


macanderson:
## Problem

`verify_done` is a catch-22 on **creation** tasks (most of Terminal-Bench): the deliverable doesn't exist at baseline, so every witness either passes at baseline (`VACUOUS TEST`) or errors on import (`WITNESS_INCONCLUSIVE_BUILD_ERROR` — explicitly ruled "NOT evidence"). Measured 2026-08-10: `kv-store-grpc__RpPZcUD` spent 430 of 911s / 20 of 52 steps in the gate loop (5× VACUOUS); `gate-ab` `pypi-server__HRUntZQ` spent 665 of 921s / 65 of 89 steps (12× VACUOUS), and **stopped the graded server twice** (`stop_process` S60, S83) to manufacture a failing baseline — destroying the deliverable to satisfy the gate, restarting it at second 918. The prove-it gate (#2662/#2663) amplifies this pressure. Panel-wide: `proven=0, witnessed=0` on every TB trial ever recorded — the proof machinery has never issued a positive verdict on this benchmark.

## Fix

1. Classify the task before choosing the witness criterion: on a **creation** task, "deliverable absent at baseline ⇒ existence/behaviour assertion" is a VALID witness; a baseline `ModuleNotFoundError` on the module that IS the deliverable is a pass signal, not inconclusive.
2. Hard cap: 2 consecutive same-class rejections → downgrade to "unwitnessed but done" and stop; the gate must never consume >15% of wall budget.
3. Protect declared long-running services: refuse `stop_process` on them during verification (destroying the graded artifact to make a test fail must be impossible).
4. Seed git identity in the agent-side baseline (S44 `fatal: unable to auto-detect email address` cost a workaround; the harbor side already uses `git -c user.name=stella-harbor`).

## Done looks like

On a creation-shaped task, a correct deliverable reaches a positive witness verdict; no TB trial spends >15% of budget in verify_done loops; `proven>0` appears on a TB panel for the first time.

Files: `crates/stella-tools/src/verify.rs` (vacuous/build-error classification), `crates/stella-pipeline` witness stage. Evidence: `turnloop-cert-panel1`, `gate-ab` transcripts.

macanderson:
Closing as **won't fix** — not because the defect isn't real, but because the subsystem it lives in is leaving core.

## What's changing

[...]

Issue #1892, opened 2026-08-06
CI: a docs-only PR can never merge — required contexts never report under ci.yml's paths-ignore, and the merge queue that made that safe is off


macanderson:
## Problem

A pull request that touches only prose is **permanently `BLOCKED`**. Not slow — unmergeable.

Branch protection requires three contexts:

```
$ gh api repos/macanderson/stella/branches/main/protection/required_status_checks --jq .contexts
["fmt + clippy + test", "cargo deny + cargo audit", "harbor_adapter + analyzer pytest"]
```

The first two come from `ci.yml`, which carries `paths-ignore` for `website/**`, `docs/**`, `*.md` and `.github/ISSUE_TEMPLATE/**`. A workflow skipped by a **path filter reports nothing at all** — unlike a job skipped by an `if:` condition, which reports `skipped` and satisfies branch protection. GitHub therefore waits forever for a check that will never arrive.

Live example: **#1863**, a two-line fix to AGENTS.md and CONTRIBUTING.md that unbreaks `main`'s `gate-parity` failure. Every check it reports is green; `mergeStateStatus` is `BLOCKED`; the two `ci.yml` contexts are absent from its check list entirely. The fix for a red `main` cannot land because it is prose.

## Why the design is not wrong, only unfinished

`ci.yml`'s own comment states the assumption:

> Note `merge_group` does NOT honor `paths-ignore`, so a docs-only PR that skips the Rust gate on `pull_request` still runs the full gate once queued; that is deliberate — the required check must report on the merged result.

That is correct **and it requires the merge queue to be enabled.** It is not (#1645 option 1, which needs admin). So the filter that was safe under a queue is, without one, a trap.

This also explains prose PRs that did merge: `enforce_admins` is off, so an admin can bypass. Everyone else waits.

## Fix directions

1. **Enable the merge queue** (#1645). Restores the documented design exactly, fixes this and the BEHIND-deadlock class together. Needs admin.
2. **Move the filtering from the workflow into the jobs.** Drop `paths-ignore` from the triggers; add a cheap `changes` job (`dorny/paths-filter` or a `git diff` step) and gate the expensive jobs with `if: needs.changes.outputs.rust == 'true'`. A skipped job **does** satisfy a required check, so the contexts always report and a prose PR still costs seconds. No admin needed.
[...]

PR #2880, opened 2026-08-11
fix(stella-tools): verify_done names the defeater it observed instead of blaming the test

### LINKED ISSUE #2871: verify_done: an unconfined shadow cwd and a VACUOUS message that names no real cause burn 81-91% of a trial's clock on already-correct work
## Problem

Trace forensics on the 2026-08-11 Terminal-Bench 2.1 panel (SUT `main@2193374`) found
`verify_done` sends the worker into an unbounded, unwinnable repair loop *on trials whose
work was already correct*. Pipeline arms made 5-7 `verify_done` calls with 5-6 `VACUOUS`
returns each; the bare arm made 0-1. `f89p2/polyglot-c-py` spent **91% of its wall clock**
after it was already correct; `f89p2/query-optimize`, **81% of its steps**. Both arms pay
this, so the fix helps the bare loop too.

Three distinct defects, all in `crates/stella-tools/src/verify.rs`:

### (a) The shadow run's cwd is not confined

`verify.rs:455` sets `current_dir` on the shadow invocation, but nothing rejects or
rewrites an absolute `cd` **inside `test_cmd`**. When the command re-enters the live root,
both halves of the proof run in the same directory, the previous-code run executes the new
code, and the verdict is `VACUOUS` — an accusation against a test that was never the
problem.

Proof, one character apart, `f89p2/query-optimize`:

- step 67, `test_cmd: "cd /tmp/stella_candidate_360_0 && bash witness_test.sh"` -> `VACUOUS`
- step 91, the identical test **without the `cd`** -> `WITNESS CONFIRMED`

~70 steps to find that.

### (b) `VACUOUS TEST` names three causes, none of which is ever the true one

`verify.rs:769` offers "the behavior already existed / the test doesn't exercise the new
behavior / your change isn't wired in" and closes with "Strengthen the test". In the
observed failures the true cause is (a)'s cwd escape or **`test_files` layering** — if the
change's deliverable is itself named in `test_files`, `stage_witness_files` copies it onto
the baseline, so the previous-code run is handed the new code and cannot fail. Neither is
mentioned. `f89p2/polyglot-c-py` step 23 proved its baseline was genuinely clean with a
manual worktree and still got `VACUOUS` six times: "strengthen the test" is advice that
cannot terminate.

### (c) Greenfield is structurally unwinnable and the tool does not say so

`f89p2/pypi-server` step 36 carries the worker's own correct diagnosis: the baseline commit
is **the empty tree** (`4b825dc642cb6eb9a060e54bf8d69288fbee4904`), so "fails on the old
code" is unreachable by construction — there is no old code. The **bare** arm burned ~13
minutes and 5,466 reasoning deltas on the identical paradox and was truncated mid-sentence.

## Fix

1. **Refuse a `test_cmd` that leaves the tree the run was aimed at.** Detect a `cd`/`pushd`
   to an absolute (or `~`-rooted) path in command position, and return a named refusal that
   states the mechanism and the exact repair, *before* paying for a shadow checkout. A
   refusal, not a rewrite: the tool cannot know what the author meant by an arbitrary path,
   and dropping the `cd` silently would change what the working-tree half measured.
2. **Make `VACUOUS` report what the call observed, not three guesses.** For each staged
   witness file, compare its working-tree blob against the same path at the baseline commit
   (`absent` / `differs` / `identical`) and lead the message with the layering defeater when
   any staged file differs from a baseline file. Keep "strengthen the test" only for the one
   shape it actually describes: every staged file identical to the baseline.
3. **Detect the empty-tree baseline and say no flip is constructible.** Route it to the same
   honest-refusal shape `WITNESS BASELINE UNRESOLVED` already uses — a fact about the
   workspace, explicitly not an accusation against the test, and explicitly not an
   instruction to call `verify_done` again.

## Done looks like

- `f89p2/query-optimize`'s step-67 command is refused with the mechanism named, instead of
  costing ~70 steps.
- A `VACUOUS` verdict names the staged file that defeated it, or says every staged file
  matched the baseline.
- An empty-tree baseline produces a terminal, non-accusatory refusal.
- Witness tests with demonstrated fail->pass flips for all three; `make gate` green;
  `crates/stella-tools/src/verify.rs` (1458 lines) does not grow toward the 1500-line
  ceiling — new code lands in a sibling module under `crates/stella-tools/src/verify/`.

- `verify.rs` is **not** in `scripts/file-size-baseline.txt` and sits at 1458 lines. It has
  ~40 lines of headroom; put the logic in `crates/stella-tools/src/verify/` beside
  `memo.rs` / `phases.rs`.
- `ToolOutput` text is what `stella-core`'s loop detector keys on — see
  `crates/stella-tools/src/verify/phases.rs` module docs. Messages must stay deterministic
  for identical inputs (no timings, no counts that change poll-to-poll).
- Related: #2668 (creation-task catch-22, the superset of (c)), #2863 (no rejection cap),
  #2505 (shadow lifecycle cost).

Evidence: `~/.claude/jobs/e286a51b/tmp/f89-mirror2/f89p2/{query-optimize,polyglot-c-py,pypi-server}/events.jsonl`.

macanderson:
## What & why

`verify_done`'s only negative-evidence verdict, `VACUOUS TEST`, offered three causes — "the
behavior already existed, the test doesn't exercise the new behavior, or your change isn't wired
in" — and closed with "Strengthen the test". On the 2026-08-11 Terminal-Bench 2.1 panel none of
the three was ever the real one, so the advice could not terminate. Pipeline arms made 5-7
`verify_done` calls with 5-6 `VACUOUS` returns each *against work that was already correct*;
the bare arm made 0-1. `f89p2/polyglot-c-py` spent **91% of its wall clock** and
`f89p2/query-optimize` **81% of its steps** in that loop. Both arms pay it, so this helps the
bare loop too.

Two setup defects produced nearly all of it, and both are decidable.

**The shadow run's cwd was not confined.** `run_against_baseline` sets `current_dir` on the
shadow invocation, but nothing rejected a `cd` to an absolute path inside `test_cmd`, which
walks straight back into the live root — both halves then execute the new code, and the verdict
is a `VACUOUS` accusation against a correct test. The evidence is one character apart:

| `f89p2/query-optimize` | `test_cmd` | verdict |
|---|---|---|
| step 67 | `cd /tmp/stella_candidate_360_0 && bash witness_test.sh` | `VACUOUS` |
| step 91 | `bash witness_test.sh` | `WITNESS CONFIRMED` |

~70 steps to find that. Now refused **before either run**, naming the mechanism and the exact
repair. Refused rather than rewritten: the tool cannot know what an arbitrary path was meant to
reach, and silently dropping the `cd` would change what the working-tree half measured. The
detector is deliberately narrow — command position only, quotes seen through, a `$VAR` or
`$(…)` destination never flagged — because over-refusing sends an author to rewrite a command
that was correct, which is the same harm in the other direction.

**`test_files` layering was never mentioned.** Every named file is copied onto the baseline,
which is the point for a test file and a defeater for anything else: name the change's
deliverable there and the previous-code run is handed the change, so it cannot fail.
`f89p2/polyglot-c-py` proved its baseline was genuinely clean with a manual worktree and still
got `VACUOUS` six times. `VACUOUS` now reports, per staged file, how it compares to the same
path at the baseline (`absent` / `differs` / `identical`, from `git rev-parse <base>:<path>`
against `git hash-object --path`), and leads with the layering defeater when any staged file
overwrote a baseline file. That observation list is the half the old verdict had none of.
"Strengthen the test" survives only in the one case it describes: every staged file identical to
the baseline.

### The third finding did not survive its own witness — reported here rather than quietly dropped

The forensics reported greenfield as *structurally* unwinnable: with the baseline at git's empty
tree there is no old code, so "fails on the old code" is unreachable. The first draft of this PR
refused such a call outright on that basis. **The test written to prove the refusal disproved
it**: on a repository whose only commit is empty, `test -f server.py` fails on the baseline,
passes on the candidate, and `verify_done` answers `WITNESS CONFIRMED` — today, on `main`. An
*existence* assertion discriminates an empty tree perfectly well; only a *behavioural* one
cannot, because the behaviour has nowhere to live.

So the blanket refusal would have destroyed confirmations the tool already grants — the exact
inversion the F4 findings exist to stop. What ships instead is the fact, stated where it changes
what the author does next: a note on the `VACUOUS` verdict when the baseline tree is empty,
naming the witness shape that still discriminates, plus a regression guard
(`an_existence_witness_still_flips_against_an_empty_baseline`) pinning the confirmation so the
refusal cannot be reintroduced by accident. The wider greenfield catch-22 — a behavioural
witness on a new module dying as `WITNESS_INCONCLUSIVE_BUILD_ERROR` — stays #2668.

Closes #2871

## The witness

Three, all demonstrated. The pre-change tree was reconstructed with
`git show HEAD:crates/stella-tools/src/verify.rs`, the three integration tests appended to its
own test module, and the suite run:

```
test verify::tests::an_absolute_cd_is_refused_instead_of_answering_vacuous ... FAILED
test verify::tests::a_layered_deliverable_is_named_as_the_defeater ... FAILED
test verify::tests::a_vacuous_on_an_empty_baseline_states_the_greenfield_fact ... FAILED
test verify::tests::an_existence_witness_still_flips_against_an_empty_baseline ... ok
test result: FAILED. 16 passed; 3 failed
```

The empty-baseline failure on `main` is worth quoting, because it is the observation that
changed the design:

```
expected the empty-baseline refusal, got Ok { content: "WITNESS CONFIRMED — deterministic
definition of done met: ... previous code: baseline 391206bb (HEAD) + your test files → exit 1
(FAIL)" }
```

On this branch:

```
test result: ok. 43 passed; 0 failed  (cargo test -p stella-tools --lib verify::)
test result: ok. 859 passed; 0 failed  (cargo test -p stella-tools)
```

`an_existence_witness_still_flips_against_an_empty_baseline` passes on **both** sides by design.
It is not a witness — it is the guard against the refusal that was almost shipped, and it is
named as such in its doc comment.

## The gate

- [x] `cargo fmt --check`
- [x] `cargo clippy -p stella-tools --all-targets -- -D warnings`
- [x] `cargo test -p stella-tools`
- [x] `make gate CARGO_SCOPE="-p stella-tools"` — all 27 steps green, including `file-size`
      (`nothing went over by this change`) and `module-reachability`
- [x] Docs updated: `crates/stella-tools/README.md` gains the negative-verdict contract;
      `verify.rs`'s module doc gains a section pointing at the new submodule
- [x] `Closes #2871` in both this description and the commit trailer

**God files:** `verify.rs` was at 1458 lines against the 1500 ceiling and is **not** in
`scripts/file-size-baseline.txt`. All new logic lands in
`crates/stella-tools/src/verify/defeater.rs`, a sibling of `memo.rs` and `phases.rs`; `verify.rs`
gains only its three call sites and their comments, ending at 1479. No baseline entry, no
ceiling raised.

**No new dependencies.** The shell tokenizer is ~40 lines of `char` handling rather than a crate,
because it deliberately does *not* parse a shell — it answers one question that quoting can hide
and expansion cannot answer, and a real parser would tempt exactly the expansion this must never
do.

## Nothing left behind

- **#2872** (filed) — the pipeline-side half of the same forensics: `witness_stage.rs`'s
  baseline-fail precondition screens only `CmdOutcome::infra_label()`
  (`TimedOut | OutOfMemory | Infra`), so a witness that fails on the baseline because
  `/bin/sh -> dash` could not parse its bash arrays (`log-summary-date-ranges`) counts as a
  legitimate assertion failure; and `ladder_decision` routed a fail-on-both-trees oracle
  (`sqlite3 --version | grep -q 'gcov'`) to `Revise`, sending the worker to fix code that was not
  broken. Written as a handoff, including why `LadderDecision::WitnessUnsatisfiable`'s existing
  `witness_unmoved_by_revision` trigger does not already cover it and must not simply be widened.
- **#2668** stays open and is referenced from `defeater.rs`: the greenfield catch-22 is larger
  than the empty-tree note here, and this PR deliberately does not claim it.
- The `Unknown` baseline status is reported honestly rather than folded into a neighbour, but no
  trace ever produced one — it is unexercised by anything but a unit test. That is a declared
  gap, not an omission; it needs a `git`-failure fixture to cover properly and is not worth a
  separate issue until something observes it.

## Summary by Sourcery

Clarify and harden verify_done’s handling of vacuous results by detecting setup-related defeaters and making negative verdicts evidence-based instead of guess-based.

Bug Fixes:
- Reject witness commands that change directory to an absolute path (or equivalent) so baseline and candidate no longer run in the same tree and spuriously yield VACUOUS TEST.
- Detect when files listed in test_files overwrite baseline files and report this layering as the reason a witness appears vacuous instead of blaming the test.
- Annotate VACUOUS TEST verdicts taken against an empty baseline tree with guidance about existence-based witnesses that can still flip in greenfield scenarios.

Enhancements:
- Introduce a dedicated verify::defeater module to encapsulate detection and messaging for setup-related defeaters and baseline comparisons, keeping verify.rs under size limits and focused on orchestration.
- Enrich VACUOUS TEST messages with per-file baseline comparison details and a single, precise diagnosis derived from those observations, while preserving the old strengthen-the-test advice only when the test is genuinely non-discriminating.

Documentation:
- Document the negative-verdict contract and new defeater behaviour in the stella-tools README and verify.rs module docs so users understand how VACUOUS TEST is determined and reported.

Tests:
- Add integration and unit tests covering cwd escape refusal, layered-deliverable detection, empty-baseline messaging, existence-witness flips on empty baselines, and the new vacuous verdict wording and observation list.

### DIFF
--- a/crates/stella-tools/README.md
+++ b/crates/stella-tools/README.md
@@ -183,6 +183,18 @@ verdict always includes the previous-code output tail: a *compile* error is a mu
 witness than an assertion failure, and the reader — model or verifier — is told to check which it
 got.
 
+**A negative verdict must name the defeater it observed, not guess at three.** `VACUOUS TEST`
+used to offer "the behavior already existed / the test doesn't exercise it / your change isn't
+wired in" and close with "strengthen the test" — advice that cannot terminate when the test was
+never the problem. Two setup defects produced nearly all of it on the 2026-08-11 Terminal-Bench
+panel, and both are decidable (#2871): a `cd` to an absolute path inside `test_cmd` walks the
+shadow run back out of the tree it was aimed at, and naming the change's deliverable in
+`test_files` layers that deliverable onto the baseline so the previous code cannot fail. The
+first is refused before either run; the second is read off a per-file comparison against the
+baseline blob and named in the verdict. `src/verify/defeater.rs` holds both, plus the discipline
+each message keeps — say the mechanism, name the observation it came from, state the repair, and
+never accuse the test of a fault the call did not observe.
+
 ## Gotchas
 
 - **A tool's timings never go in its `ToolOutput`.** That value is what
--- a/crates/stella-tools/src/verify.rs
+++ b/crates/stella-tools/src/verify.rs
@@ -53,6 +53,14 @@
 //!   of the *next* `verify_done`, because removing it requires an awaited
 //!   subprocess.
 //!
+//! # When the setup, not the test, is what defeated the proof
+//!
+//! A `cd` to an absolute path in `test_cmd` walks the shadow run out of the
+//! tree it was aimed at; naming the deliverable in `test_files` layers it onto
+//! the baseline. Both make a correct test look vacuous, and the verdict used
+//! to blame the test for each. The `defeater` submodule beside this one
+//! (`src/verify/defeater.rs`) detects them (#2871).
+//!
 //! # Reading the verdict
 //!
 //! The shadow run's output tail is always included: a shadow failure that
@@ -87,6 +95,7 @@
 //! own output.
 
 mod creation;
+mod defeater;
 mod memo;
 mod phases;
 
@@ -570,6 +579,10 @@ impl VerifyDone {
     ) -> Result<(ToolOutput, Verdict), String> {
         let setup = Phase::start();
         let test_cmd = crate::input::required_str(input, "test_cmd").map_err(|e| e.to_string())?;
+        // Before anything expensive, and refused rather than rewritten (#2871).
+        if let Some(refusal) = defeater::cwd_refusal(test_cmd) {
+            return Err(refusal);
+        }
         let test_files: Vec<String> = input
             .get("test_files")
             .and_then(|v| v.as_array())
@@ -780,21 +793,17 @@ impl VerifyDone {
         };
 
         if old_exit == 0 {
-            return Ok((
-                ToolOutput::Error {
-                    message: format!(
-                        "VACUOUS TEST — the witness test ALSO PASSES on the previous code \
-                         (baseline {}, {}). It does not witness your change: either the \
-                         behavior already existed, the test doesn't exercise the new behavior, \
-                         or your change isn't wired in. Strengthen the test so it fails without \
-                         your change.{reuse_note}\n--- previous-code output tail ---\n{}",
-                        baseline.short(),
-                        baseline.provenance,
-                        tail(&old_output)
-                    ),
-                },
-                Verdict::Vacuous,
-            ));
+            // Which defeater this names comes from observations of the
+            // workspace, never from the guesses it used to offer (#2871).
+            let message = defeater::vacuous_verdict(
+                root,
+                &baseline,
+                &resolved,
+                reuse_note,
+                tail(&old_output),
+            )
+            .await;
+            return Ok((ToolOutput::Error { message }, Verdict::Vacuous));
         }
 
         // A failure while LOADING the test is not a failure OF the test:
@@ -902,7 +911,7 @@ mod tests {
     /// exactly like the production paths (`git_in`) — without this, running
     /// the suite from inside a git hook (the pre-push gate) re-targets every
     /// command at the HOST repo. Returns stdout; panics on failure.
-    fn scratch_git(root: &std::path::Path, args: &[&str]) -> String {
+    pub(super) fn scratch_git(root: &std::path::Path, args: &[&str]) -> String {
         let mut cmd = std::process::Command::new("git");
         cmd.args(args).current_dir(root);
         for var in crate::exec::GIT_REPO_ENV_VARS {
--- a/crates/stella-tools/src/verify/defeater.rs
+++ b/crates/stella-tools/src/verify/defeater.rs
@@ -1,0 +1,842 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+// Copyright (c) 2026 Oxagen, Inc. Commercial licensing: licensing@oxagen.sh
+
+//! Why a `verify_done` call could not produce a flip — when the answer is a
+//! fact about the **setup**, not about the test.
+//!
+//! # The failure this module exists to end
+//!
+//! `verify_done`'s only negative-evidence verdict used to be `VACUOUS TEST`,
+//! and its message offered three causes: the behaviour already existed, the
+//! test doesn't exercise it, the change isn't wired in. On the 2026-08-11
+//! Terminal-Bench panel none of the three was ever the real one, and the
+//! closing advice — "strengthen the test" — is advice that cannot terminate
+//! when the test was never the problem. Pipeline arms made 5-7 calls with 5-6
+//! `VACUOUS` returns each; `polyglot-c-py` spent 91% of its wall clock, and
+//! `query-optimize` 81% of its steps, *after the work was already correct*
+//! (#2871).
+//!
+//! Two mechanisms produced almost all of it, and both are decidable:
+//!
+//! - **The command leaves the tree the run was aimed at.**
+//!   [`run_against_baseline`](super::run_against_baseline) sets `current_dir`
+//!   on the shadow invocation, but a `cd` to an absolute path inside `test_cmd`
+//!   walks straight back out of it. `query-optimize` step 67 ran
+//!   `cd /tmp/stella_candidate_360_0 && bash witness_test.sh` and got
+//!   `VACUOUS`; step 91 ran the identical test **without the `cd`** and got
+//!   `WITNESS CONFIRMED`. Seventy steps, one prefix.
+//! - **`test_files` layering.** [`stage_witness_files`](super::stage_witness_files)
+//!   copies every named file onto the baseline, which is the whole point for a
+//!   test file and a defeater for anything else: name the change's deliverable
+//!   there and the previous-code run is handed the new code, so it cannot fail.
+//!
+//! And one fact is simply invisible: when the baseline commit is git's
+//! [`EMPTY_TREE_OID`], there is no old code for a *behavioural* assertion to
+//! fail against, so a whole family of witnesses cannot discriminate however
+//! hard the author strengthens them. `pypi-server` reasoned its way to that
+//! and kept calling anyway; the bare loop burned ~13 minutes and 5,466
+//! reasoning deltas on the same paradox. See [`empty_baseline_note`] for what
+//! that fact does and does not license — it is not the blanket "no flip is
+//! constructible" the forensics reported, and the test written to prove the
+//! stronger claim disproved it.
+//!
+//! # The discipline every message here follows
+//!
+//! Say the mechanism, name the observation it was read off, and state the
+//! repair. Never accuse the test of a fault the call did not observe: a
+//! refusal that misattributes is worse than no refusal, because the author
+//! acts on it. This is the same contract
+//! [`resolve_witness_baseline`](super::resolve_witness_baseline)'s
+//! `WITNESS BASELINE UNRESOLVED` already keeps.
+//!
+//! Every string below is a pure function of the call's inputs and
+//! observations — no timings, no counters. `ToolOutput` is what `stella-core`'s
+//! loop detector keys on (see [`phases`](super::phases)), and a message that
+//! varies call to call would blind it for the one tool an agent stuck in a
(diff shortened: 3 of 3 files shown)


macanderson:
**CI registration note for whoever reviews this.** The `ci` workflow never started on the
`pull_request` event for this PR, twice: at open (22:21:29Z) and after a close/reopen
(22:23:28Z). Only `cla` fired on both. `gh api repos/macanderson/stella/commits/885b4d4ad/check-runs`
lists `cla`, `Sourcery review` and the three Vercel contexts and nothing else, while a sibling PR
pushed sixteen seconds later (`fix/2876-post-completion-checks-discard-work`, run `31542074432`)
got its `ci` run normally. `gh api .../actions/workflows/ci.yml` reports `state: active`, and
`ci.yml`'s `pull_request:` trigger carries no path filter by design (#1892/#1863), so nothing in
the workflow explains it — this looks like a dropped event on GitHub's side, not a property of
the diff.

I could not fix it from here (a force-push of the same tree to re-fire `synchronize` was blocked
by tooling policy), so the gate was run explicitly instead:
`gh workflow run ci.yml --ref fix/verify-done-defeaters` -> run
[31542315647](https://github.com/macanderson/stella/actions/runs/31542315647). A `workflow_dispatch`
run does **not** satisfy the required `pull_request` contexts, so **this PR still needs a real
`ci` run before it can merge** — one further push, or a re-run of the trigger, should produce it.

Reporting it rather than leaving a reviewer to find a PR with no required checks and assume the
diff was skipped.

macanderson:
### Rebased onto `881ed40cb` (Release/0.9.0), and one real review finding fixed

**The rebase was mechanical.** One conflicting line: main gained `mod creation;` (#2867) and this
adds `mod defeater;`. `README.md` auto-merged; the 0.9.0 version churn touched nothing here. The
two modules are complementary and share no state — `creation` owns the
`WITNESS_INCONCLUSIVE_BUILD_ERROR` arm, `defeater` owns the `VACUOUS` arm and the pre-run command
check, and they sit in different branches of `verify()`. The `#2668` cross-reference in
`defeater.rs` now points at `super::creation`, which exists in the same tree as of this rebase.

**The rebase did cost line budget, and that is worth stating.** `verify.rs` came back from the
rebase at 1472 and my first shape pushed it to **1512 — `check-file-size` failed**. I did not add
a baseline entry. Instead both call sites collapsed into one call into `defeater`
(`vacuous_verdict`, `cwd_refusal`), each with the prose split out behind it so every branch stays
reachable from a unit test with no git repository. The file lands at **1499**, under the ceiling,
with none of main's lines disturbed.

**Review finding — fixed, with a witness.** Vercel Agent Review caught a genuine hole in
`cwd_escape`: it returned `None` the moment it saw a `cd` with a *relative* destination, which
cleared every absolute `cd` later in the same chain. `cd src && cd /tmp && pytest` read as
confined. A chain is only as confined as its last hop, so the scan now resumes at the next
command rather than concluding for the whole chain, and
`a_harmless_cd_does_not_clear_an_absolute_one_later_in_the_chain` pins it — it fails on the
commit it replaced. Thread resolved. Good catch; that was a correctness bug, not a style note.

**Verification after the rebase**, all against `881ed40cb`:

- Witness flips re-demonstrated by appending the three integration tests to
  `git show origin/main:crates/stella-tools/src/verify.rs`:
  `an_absolute_cd_is_refused_instead_of_answering_vacuous` FAILED,
  `a_layered_deliverable_is_named_as_the_defeater` FAILED,
  `a_vacuous_on_an_empty_baseline_states_the_greenfield_fact` FAILED — 16 passed, 3 failed.
- `make gate CARGO_SCOPE="-p stella-tools"` → exit 0, all 27 steps, including
  `check-file-size: OK … Judged against 881ed40cb` and `self-driving-test: OK — 60 checks passed`.
- `cargo test -p stella-tools --lib verify::` → 54 passed (creation's joined this module's).

**On the earlier CI note in this thread:** the required contexts did register on the force-push,
so that comment is now historical. The previous head (`d666b13fb`) went fully green —
`fmt + clippy + test`, `cargo deny + cargo audit` and `harbor_adapter + analyzer pytest`, the
three required contexts, all success — and `mergeable` moved `DIRTY` → `MERGEABLE`.
`harbor_adapter` failed once on infrastructure, not the diff: `actions/checkout` never fetched
the code (`server certificate verification failed`), the same TLS fault that hit `gh` locally in
the same minute; a re-run passed in 18s. This push restarts all three for the review fix.

Issue #2902, opened 2026-08-12
Flaky: `daemon::tests::a_run_that_ignores_term_is_killed_after_the_grace_period` fails under a full parallel `cargo test`


macanderson:
## Problem

`daemon::tests::a_run_that_ignores_term_is_killed_after_the_grace_period` (`crates/stella-cli/src/daemon/tests.rs:483`) fails reliably under a full `cargo test -p stella-cli` and passes reliably when run alone:

```
thread '...a_run_that_ignores_term_is_killed_after_the_grace_period' panicked at
crates/stella-cli/src/daemon/tests.rs:483:5:
the child was killed after 81.175343ms, before its 8s to shut down cleanly
```

## It is not a regression

Confirmed with the cheap control rather than inferred: `git stash` the working tree and re-run the same full suite — the failure reproduces identically on the unmodified tree, and the isolated run passes on both. So it is a property of the test under parallel load, not of any change.

```
$ git stash && cargo test -p stella-cli 2>&1 | grep FAILED
test daemon::tests::a_run_that_ignores_term_is_killed_after_the_grace_period ... FAILED
$ cargo test -p stella-cli daemon::tests::a_run_that_ignores_term
test result: ok. 1 passed; 0 failed
```

## The likely mechanism

The assertion measures **wall clock** across a signal round trip: the test spawns a child that ignores `SIGTERM`, asks the supervisor to stop it, and asserts the hard kill lands no earlier than the 8s grace period. Under a saturated test runner (the suite runs ~1700 tests in parallel and the machine is oversubscribed) the elapsed measurement and the supervisor's own timer are reading different clocks' worth of scheduling delay, and the observed 81ms says the kill path ran without the grace timer having advanced as expected.

Worth checking before assuming: whether the grace period is derived from an injectable clock/sleeper at all, or from a real `tokio::time::sleep` the test cannot control. If it is real time, the fix is the same seam `stella-core` uses everywhere else (`retry::Sleeper`) — a wall-clock assertion in a parallel suite is a coin flip weighted by machine load.

## Files

- `crates/stella-cli/src/daemon/tests.rs` (the assertion, ~line 483)
- `crates/stella-cli/src/daemon.rs` (the grace-period timer and the kill path)

## Done

[...]

Issue #2900, opened 2026-08-12
recall_context is registered only on the deck's pipeline turn — the step-loop lead turn and `stella run` cannot reach the plane mid-turn


macanderson:
## Problem

`recall_context` (added on `claude/transcript-colors-contrast-msqxjc`) lets an agent search the workspace context plane mid-turn, over the same `ContextRecallPort` the turn-start recall uses. It is wired at exactly one call site: `run_lead_pipeline_turn` in `crates/stella-cli/src/command_deck/pipeline_turn.rs`.

Every other surface that builds an `InteractiveToolSet` still offers no mid-turn lookup:

- `command_deck.rs::run_lead_turn` — the deck with `/pipeline` OFF.
- `agent.rs` (two sites), `agent/goal.rs` (two), `agent/resume.rs` — the REPL, `/goal`, one-shot `stella run`, and resume.

The deck defaults to the pipeline, so the common path is covered; the gap is real for everything else, and an agent that learns a tool exists on one surface and not another is worse off than one that never had it.

## Files

- `crates/stella-cli/src/interactive.rs` — `InteractiveToolSet::with_recall`, `recall_schema`, `execute_recall_context`.
- `crates/stella-cli/src/command_deck/pipeline_turn.rs` — the one call site that wires it (search `with_recall`).
- `crates/stella-cli/src/command_deck.rs::run_lead_turn`, `crates/stella-cli/src/agent.rs`, `crates/stella-cli/src/agent/goal.rs`, `crates/stella-cli/src/agent/resume.rs` — the sites that do not.

## Constraints already discovered

- `with_recall` takes a **borrowed** `&'a dyn ContextRecallPort` plus the turn's `EventSender`. The event must ride the turn's real channel, not the interactive set's — the deck deliberately hands that set a stub sender so its own `ask_user` emission does not double the card.
- `run_lead_turn` does not currently receive `memory` at all; it takes a pre-computed `recall_event` instead (the driver recalls before the channel exists). Threading `Option<&SessionMemory>` into it is the change, and `command_deck.rs` is a god file closed to growth — plan the wiring so it does not add lines there.
- The catalog row already exists (`crates/stella-tools/src/catalog.rs`, group `context`), so no catalog change is needed.

## Verify

[...]

Issue #2899, opened 2026-08-12
Esc-steering falls back to the backlog for worker lanes — the driver holds no handle on a worker's steering tap


macanderson:
## Problem

`WorkspaceInput::Steer` (added on `claude/transcript-colors-contrast-msqxjc`) delivers the deck's Esc-with-something-to-say into the running turn at its next step boundary. That works for the **lead** lane, which the driver steers through `subsession::SteeringTap`.

It does **not** work for a worker lane. `SubSessions` (`crates/stella-cli/src/subsession.rs`) tracks a stop channel and a pause channel per lane and nothing else — a worker's own `SteeringTap` is constructed inside `subsession`'s spawn path and never handed back — so the driver has no way to inject at that lane's boundary.

Today the words are pushed onto the backlog instead (`command_deck/steer.rs::steer_worker`). They are not lost, but the user asked to correct the agent that is running *now* and got "this will be the next thing somebody does", which is a different answer.

## Files

- `crates/stella-cli/src/subsession.rs` — `SubSessions` (the `stops` / `pauses` / `winding_down` maps), `SteeringTap`, `SubSessionSpec`, the worker spawn path.
- `crates/stella-cli/src/command_deck/steer.rs` — `steer_worker`, the fallback this issue replaces.
- `crates/stella-cli/src/command_deck.rs` — the mid-turn `select!` arm that dispatches `WorkspaceInput::Steer`.
- `crates/stella-tui/src/envelope/steering.rs` — the design note for the whole change.

## Sketch

Retain each worker's `Arc<SteeringTap>` in `SubSessions` beside its stop/pause channels (it is already an `Arc` — the deck feeds it while the worker task drains it), and add `SubSessions::steer(&self, lane: &str, text: String) -> bool` with the same stale-lane-is-a-no-op contract `stop` has. Then `steer_worker` becomes the same shape as `steer_lead`.

Watch the generation discipline `ended`/`stop` already carry: a steer aimed at a replaced worker must not reach its replacement's tap.

## Verify

A witness in `crates/stella-cli/src/subagent/tests.rs` (which already builds `Arc<SteeringTap>` directly — see the tests around lines 494, 528, 598): spawn a worker lane, send a `Steer` at it, and assert the text arrives via `drain_steering()` on that lane's tap and **not** on the backlog. A second assertion should pin the stale case: a steer at a lane that has wound down is a no-op, never a panic and never a silent backlog push.

## Done

[...]

PR #2904, opened 2026-08-12
Push hooks, tool classes, mid-turn steering, and pipeline workspace isolation

macanderson:
## What & why

Six fixes to what the Command Deck shows and what Esc does, all from one reported session where the deck was legible but not readable.

1. **The transcript was one colour.** Every tool name rendered gold, so "read a file" and "deleted a file" looked identical, and an argument-less call rendered a literal `{}`. Tool names are now coloured by **class** — inspect / mutate / execute / verify / repo / delegate — derived from the same `stella_tools::catalog` the policy layer reads, so a tool cannot be filed under one family for policy and another on screen. `crates/stella-tui/src/tool_class.rs`.

2. **The plan rail never moved.** An approved six-step plan sat at `0/6` for the whole run. The execute stage now marks each step in-progress before its turn and done after, driven by the party that knows the answer. `crates/stella-pipeline/src/pipeline/execute_stage.rs`.

3. **An isolated candidate's edits never reached the transcript**, so diffs could not render under the call that made them. Only a lone candidate's edits become the user's, so only that one announces live. `crates/stella-pipeline/src/ports/workspace.rs`.

4. **`repo_push` was defeated by this repository's own pre-push hook** (`make gate` — the whole suite), so it always timed out and the only way through was `bash git push --no-verify`, outside the surface meant to govern it. The tool now reads the diff and decides *for itself* whether hooks may be skipped: no source file changed, they may; source changed, `skip_hooks: true` is refused with the reason. It also reports what the push carried instead of git's terse `To <url>`. `crates/stella-tools/src/repo/push.rs`.

5. **Esc restarted the turn instead of interrupting it.** On the staged pipeline a cancel is a *restart*: the next turn re-enters at triage and context recall, so the plan, the recalled frames and every file read are derived again from nothing. One correcting sentence therefore cost the whole turn a second time — measured at $1.43 and 18 minutes in the reported session — and the sentence stayed stranded in the composer. Esc now **steers**: the draft and queue go into the running turn at its next step boundary and the turn continues. Nothing to say is still a plain interrupt; double-Esc is untouched. `crates/stella-tui/src/envelope/steering.rs` carries the reasoning.

6. **The agent could reach the context plane only once per turn**, at the top, against the goal as stated — the moment it knows least about what it will need. `recall_context` exposes the same port mid-turn: not a second retrieval system, so a step-12 lookup and the step-0 recall can never disagree.

Exemplars followed: `command_deck/steer.rs` and `deck_ui/steer.rs` follow the existing `settle.rs` / `task_tap.rs` split; `PlanWalk` follows the argument-bundle shape `cargo` uses for positional context.

## The witness

- [x] This PR includes witness tests (fail on `main`, pass here):
  - `deck_ui/tests.rs::esc_with_a_typed_draft_steers_the_running_turn_instead_of_cancelling_it`
  - `deck_ui/tests.rs::esc_delivers_the_whole_prompt_queue_in_order`
  - `deck_ui/tests.rs::esc_with_nothing_to_say_is_still_a_plain_interrupt`
  - `interactive.rs::recall_context_searches_the_plane_mid_turn_and_announces_it`
  - `interactive.rs::an_empty_plane_answers_rather_than_erroring`
  - `interactive.rs::no_plane_means_no_tool`
  - `pipeline/tests/best_of_n.rs::workspace_creation_seeds_the_plan_and_only_a_lone_candidate_may_report`
  - `render/tests/palette.rs::tool_class_colors_are_categorical_and_distinct`
  - `theme/tests.rs::every_tool_class_is_categorical_and_distinct`

**Deleted test, named for review** (per the `check-deleted-tests` rule): `deck_ui/tests.rs::a_typed_draft_survives_both_esc_forms` is replaced by the first witness above. Its contract — the draft is preserved *and the turn is cancelled* — is precisely the defect fix 5 addresses.

## The gate

- [x] `cargo fmt --check`
- [x] Docs updated: `recall_context` documented in `website/content/docs/agent-tools/index.mdx` (card, read-only partition, session-tool count); `repo.rs` module docs describe hook-skip; `envelope/steering.rs` and `tool_class.rs` are doc-carrying modules.
- [x] CLA signed

**Four god files were in the way and none was grown.** The Esc decision moved to `deck_ui/steer.rs`, the driver's delivery to `command_deck/steer.rs`, the reasoning to `envelope/steering.rs`, and `run_lead_pipeline_turn` — a whole execution mode — out of `command_deck.rs` into `command_deck/pipeline_turn.rs`. No baseline entry was added and no ceiling raised.

## Nothing left behind

- [x] Filed as handoffs:
  - #2899 — Esc-steering falls back to the backlog for worker lanes; the driver holds no handle on a worker's steering tap. Referenced from the `steer_worker` doc comment.
  - #2900 — `recall_context` is registered only on the deck's pipeline turn; the step-loop lead turn and `stella run` cannot reach the plane mid-turn.
  - #2907 — `files_touched` and `wrote_files` are blind to every edit an isolated pipeline candidate makes. Found while reading this session's own exported telemetry; proved with a repro showing the `stella-tools` ledger itself is correct, so the loss is at the adoption seam.
  - #2902 — `daemon::tests::a_run_that_ignores_term_is_killed_after_the_grace_period` is flaky under a full parallel `cargo test`. Confirmed pre-existing with a `git stash` control, not caused by this PR.

## Ground-rule check

- [x] New cross-boundary types (`WorkspaceInput::Steer`, `AdoptedChange` additions) round-trip through serde

**#2907 is the one worth a second look.** It is a real defect this PR does *not* fix: for a pipeline execution that ran in an isolated candidate, `store.db`'s `files_touched` records only the research stage's reads and none of the execute stage's edits, and `wrote_files` reads `0`. The fix crosses three crates and is a different subject from this PR — which the automated reviewer has already refused as over its diff-size limit — so it is filed rather than bundled. The issue carries the exported rows, the repro, and the suggested seam.

`clippy` flagged `run_plan_steps` at 8/7 arguments after the plan-rail change; it took a `PlanWalk` struct rather than an `#[allow]`, because its three positional fields are one fact and splitting them is how a resumed run's prompts came to be able to say "step 1 of 2" about the fourth step of five.

https://claude.ai/code/session_0145VgM6DEtZpji4pWJx8eZc


macanderson:
CI red on `fmt + clippy + test`, failing on the base branch too — not from this PR.

`main` moved from `999d08b` to `9bb990d` while this branch was open, picking up `10781aa "chore: update colors for brand"`, which changed `--stella-gold-500` from `#FFB000` to `#C58A32` in `docs/brand/css/tokens.css` and left the consumers on the old value. `design_token_parity::identity_gold_comes_from_the_brand_ramp` catches it:

```
left: Some("#ffb000")   <- crates/stella-observatory/src/assets/index.html
right: Some("#c58a32")  <- docs/brand/css/tokens.css
```

This branch touches none of the three files that test reads, and the suite passes 8/8 on the branch head. Swapping only `tokens.css` to `main`'s version into an otherwise-green tree reproduces the failure byte-for-byte, and restoring it returns to green — so it is that one file, not this diff.

Filed as #2914 with the full consumer list (about ten surfaces are still on the old gold), the contrast deltas, and the two ways out. Not fixing it here: it is a brand decision and a coordinated sweep, and burying it in a TUI PR is the wrong place for it.

Will re-run when the base recovers.

---

PR #2909, opened 2026-08-12
fix(brand): regenerate every raster asset and give each family a builder

macanderson:
## What broke

`10781aa31 chore: update colors for brand` regenerated all 52 brand PNGs with a
rasteriser that emitted torn scanlines and channel-separated garbage, and
committed them. Every file still decoded and still reported correct dimensions,
so nothing caught it.

Evidence, both sides:

```
git show 78a60f30e:docs/brand/pwa/icon-512.png   # clean comet
git show 10781aa31:docs/brand/pwa/icon-512.png   # torn, RGB-separated
rsvg-convert -w 1024 docs/brand/logo/svg/lockup-color-dark.svg -o /tmp/t.png
                                                 # renders clean — the SVGs were never damaged
```

It was **two** defects. The same commit recoloured the four spinner *SVGs* to
Bronze Gold and left their *GIFs* untouched, so the kit shipped `#C58A32`
vector art beside `#FFB000` animations. Verified by parsing the GIF colour
tables before and after.

Same root cause for both: `logo/png/`, `pwa/`, `wallpapers/` and `spinners/`
had **no builder**. Only `social/` had one, and only `social/` could be re-run
to a known-good state.

## What this does

Gives every family a builder and the geometry one home.

| File | Builds | `--check` |
|---|---|---|
| `cometkit.py` | palette, comet paths, the one rasteriser entry point | — |
| `build_marks.py` | 22 logo PNGs + 8 PWA icons + `favicon.ico` | ✅ |
| `wallpapers/build_wallpapers.py` | 12 wallpapers | ✅ |
| `spinners/build_spinners.py` | 4 GIFs, from the animated SVGs | ✅ |
| `social/build_social.py` | unchanged, re-run | ✅ |

`cometkit.py` is assembled from the blocks `build_social.py` already had, so
the comet paths now have exactly one copy instead of the three a new builder
would have created. Exemplar for the shape: `cargo`'s `util` crate — shared
mechanics extracted to a leaf that every command imports, rather than a
`common.rs` grab bag.

**No `--check` compares bytes.** librsvg's output shifts between releases, so a
byte comparison goes red on a Homebrew upgrade while the art is still correct,
and a guard that cries wolf gets silenced — which is how the corruption
travelled. Each asserts what a broken build actually violates: every file
exists, decodes, inflates and matches the size its name promises; `cometkit`'s
geometry still matches the committed SVGs; the three wallpaper tiers draw the
same picture; the spinner GIFs carry the current gold.

The spinner builder **reads the keyframes out of the SVG** rather than
restating them. librsvg does not evaluate CSS animation, so the GIF must be
built frame by frame; reimplementing the timeline in Python would recreate
precisely the drift that left the GIFs amber. It parses `@keyframes`, evaluates
per frame, and re-emits the original tags with geometry untouched.

## Wallpapers: redesigned, not restored

The old ones were a flat ground with the mark parked in the middle. These are
the kit's own sentence about its mark — a tapered plume entering off-canvas,
debris thinning along it, the comet at the head, a second comet far off for
scale. Desktop keeps the top-right quiet for icons; phone keeps the upper third
clear of the lock clock.

- **Nothing is filtered.** `feGaussianBlur` at 6016×3384 is minutes per file in
  librsvg, so glow is `radialGradient` and soft edges are nested tapered
  plumes. A stroke cannot do this: one stroke has one width for its whole
  length, so stacking strokes stacks their edges and the trail reads as
  concentric plastic tubing.
- **Every gradient is dithered** with the kit's seeded grain — a wide ramp
  across near-black posterises into steps centimetres wide at 6K.

## Witness

Star density was per *megapixel*, so the ladder shipped three different skies
under one name — 900 specks at 4k, 2,400 at 5k, 3,800 at 6k. Now per canvas,
guarded by `check_ladder_parity`, which counts drawn elements per tier.

Fails on the old code, passes on the new:

```
CAUGHT: ladder drift: desktop/dark draws a different number of elements per
        tier — {'4k': 1524, '5k': 2397, '6k': 3156}
PASS on fixed code
```

## Two deliberate changes, flagged

Neither is a restoration — call them out if you disagree:

1. **`favicon-16` / `favicon-32` carry the star alone.** The trail's bars are 7
   units of a 44-unit mark, so at 16px they land under one pixel and render as
   three grey smudges — the mark reads as damaged. 48px and up carry the full
   comet.
2. **The wallpaper head is the star alone.** The plume already *is* the trail;
   the mark's own three bars would be a second trail pointing the same way. The
   bottom lockup is where the whole mark gets its say.

## Verification

```
python3 docs/brand/build_marks.py --check              # ok: 22 logo exports, 8 icons, 1 ico
python3 docs/brand/wallpapers/build_wallpapers.py --check   # ok: 12 wallpapers
python3 docs/brand/spinners/build_spinners.py --check       # ok: 4 spinners
python3 docs/brand/social/build_social.py --check           # ok — install line and 5 layouts fit
make guards-fast                                            # exit 0
```

No Rust changed. Every output was also opened and looked at, which is the check
no script performs and the one this whole PR exists because nobody ran.

Repo weight: 44.1MB → 45.5MB across `docs/brand/` (social *halved*, 5.8→3.0MB —
datamosh compresses badly).

## Summary by Sourcery

Introduce a unified comet brand geometry and rasterisation module and add builders for all brand raster assets to ensure they are fully generated and verifiable from SVG sources.

New Features:
- Add cometkit.py as the single source of truth for comet geometry, palette, and rasterisation utilities used across brand builders.
- Add build_marks.py to generate logo PNG exports, PWA icons, and favicon.ico from committed SVG artwork with a non-byte-comparing check mode.
- Add wallpapers/build_wallpapers.py to procedurally generate the full desktop and phone wallpaper ladder in dark and light variants with parity checks across resolutions.
- Add spinners/build_spinners.py to render spinner GIFs directly from animated SVG keyframes, including a check that enforces the current gold palette.

Enhancements:
- Document that all brand raster assets are now generated, listing the builders and their requirements in docs/brand/README.md and describing the redesigned wallpapers.
- Standardise validation of raster outputs via structural and dimensional checks (PNG verification, ladder parity, GIF palette checks) instead of brittle byte comparisons across tools.


macanderson:
## Left behind, filed

Neither is fixed here — both are separate changes with a different blast radius
than "regenerate the rasters":

- **#2910** — nothing runs the four new `--check` modes. They are not in
  `GATE_STEPS` or any workflow, so the failure class this PR diagnosed is still
  uncaught. Note the trap recorded there: `ci.yml` ignores `docs/**`, so a
  brand-only PR starts `docs-guards.yml` and not the required job.
- **#2911** — the website is still entirely on the previous amber palette
  (`#FFB000` / `#F4F1EA`) while `docs/brand/` is Bronze Gold. All 13 SVGs in
  `website/public/brand/` disagree with their `docs/brand/logo/svg/`
  counterparts, and `website/src/components/brand.tsx` carries a fourth copy of
  `STAR_PATH`. `check_svg_parity` added here only covers `docs/brand/`.

The second one is user-visible in public and may want doing soon, but it is a
site recolour rather than an asset regeneration, and it needs a human to
confirm the website was not deliberately parked on the old palette.

Issue #2918, opened 2026-08-12
The mid-turn provider settling latch is per turn, so every extra worker turn re-pays a dead provider's failure


macanderson:
## Problem

`crates/stella-pipeline/src/pipeline/tests/midturn_fallback.rs::an_exhausted_execute_turn_re_resolves_the_worker_and_finishes_on_the_fallback` now records **two** `ProviderFallback` events and two failed engine calls against the sick provider, where it recorded one before.

Nothing about the fallback changed. #2908 added a second worker turn on the unproven path, and the settling latch that pins a run to its healthy replacement is scoped to one engine turn — so the second turn starts on the roster's original choice, pays the sick provider's full retry ladder again, and announces its own swap.

This is not new to #2908: every revision turn (`Revise`, `NothingAttempted`, the #1295 evidence demand) has always had this shape. #2908 only made it common enough to notice, because unproven is the ordinary outcome on a task with no test command.

## Why it may matter

The cost is one wasted provider round trip and one extra `Error` event per turn after the first, on a run that has already established the provider is down. The `error` event is user-visible and says "terminal provider error" each time. Against a hard-down provider with a long retry ladder this is real wall clock, and on a benchmark it is wall clock charged to the SUT.

## What to decide

Whether the settling decision belongs to the turn or to the candidate. Arguments exist for the turn (a transient outage should not permanently re-route a long run) — so this is a question to answer deliberately, not an obvious bug. If it stays per-turn, say so where the latch lives, because the test above now reads as if the duplication were incidental.

## Files

- `crates/stella-pipeline/src/pipeline/fallback.rs` — `FallbackPosture` / `RoleFallbacks`, the latch
- `crates/stella-pipeline/src/pipeline/execute_stage.rs` — the per-turn drivers
- `crates/stella-pipeline/src/pipeline/tests/midturn_fallback.rs` — the witness, whose expectation was updated in #2908 with this reasoning written into the assertion message

## Repro / verify

`cargo test -p stella-pipeline --lib midturn_fallback` and read the `fallbacks(&events)` assertion.

## Done when

Either the latch survives across a candidate's turns, with a witness test showing exactly one announced swap for a run that takes several turns against a down provider; or the per-turn scope is documented at the latch with the reason, and the test's expectation cites it.

Refs #2908

Issue #2917, opened 2026-08-12
Golden fixture single_task_deterministic_flip.jsonl still carries the pre-#2556 flip_achieved field, and the diff tolerates it


macanderson:
## Problem

`crates/stella-pipeline/tests/fixtures/golden/single_task_deterministic_flip.jsonl` records the verdict's ladder snapshot with `"flip_achieved":true`. The pipeline has emitted the tri-state `"flip":"achieved"` since #2556. The golden test passes anyway, because `GoldenTrajectory::diff` compares structurally and does not read that field — so the fixture is stale and nothing says so.

Found while re-recording the neighbouring fixture for #2908: `make record-golden` rewrote this file too, with the field rename as its only change. That change was reverted from that PR rather than folded in, so the staleness is still here.

## Why it matters

The whole point of a committed golden is that a change to the observable event contract shows up as a fixture diff. A field that can change on the wire without moving the fixture is a hole in that guarantee, and the next reader of this file will believe a shape the pipeline has not emitted for months.

## Repro

```
rg -n 'flip_achieved' crates/stella-pipeline/tests/fixtures/golden/
STELLA_REFRESH_GOLDEN=1 cargo test -p stella-pipeline --lib golden
git diff crates/stella-pipeline/tests/fixtures/golden/
```

The refresh rewrites `flip_achieved` to `flip` and nothing else; the unrefreshed fixture passes the test.

## Files

- `crates/stella-pipeline/tests/fixtures/golden/single_task_deterministic_flip.jsonl`
- `crates/stella-pipeline/src/replay/golden.rs` — `GoldenTrajectory::diff`, the comparison that tolerates it
- `crates/stella-pipeline/src/pipeline/tests/golden.rs` — the harness

## Done when

The fixture is re-recorded from a fresh `main` (its own PR, so the diff is reviewable as the contract change it is), **and** the decision about `diff`'s tolerance is written down: either it is deliberately structural — in which case say so where the tolerance lives, and say what a stale payload field is then worth — or it tightens and a witness test shows a renamed snapshot field failing the golden.

Refs #2908

Issue #2916, opened 2026-08-12
Measure what the unproven handback costs: one extra worker turn plus a full verification round per unproven run (#2908 follow-up)


macanderson:
## Problem

#2908 made the `Unverified` and `Unverifiable` rungs hand the work back to the worker instead of ending the run (`crates/stella-pipeline/src/pipeline/unproven.rs`). That is correct for the invariant — the pipeline may only add a claim, never subtract the work — but it is not free, and the price has not been measured.

Per unproven run the pipeline now spends:

- **one extra worker turn** (the handback), and
- **one extra verification round** over the re-observed tree: `observe_touched_tests`, and the pre-submit audit's lint / coverage / mutation probes wherever they are wired.

Two of the updated fixtures show the shape directly: `a_persistent_memory_kill_is_never_a_deterministic_test_failure` went from 3 suite runs to 5, and `a_tautological_witness_is_downgraded_to_the_verifier` from 2 mutant runs to 4.

On the Terminal-Bench shape the issue was measured on, *most* trials reach an unproven rung, so this is the common path and not the rare one.

## Why it may still be right

The handback is what makes the pipeline's floor the bare loop's floor. The question is not whether to keep it but whether the round it re-observes is worth its compute — in particular whether the **pre-submit audit** needs to re-run at all when the handed-back turn dispatched nothing that could have changed the tree.

## What to do

1. Run a paired ArenaBench arm (pipeline before / pipeline after #2908) on the same tasks and report: solve rate, wall clock, worker steps, tool calls, suite invocations, and `priced_cost`. Compare like for like and name any operational aborts explicitly.
2. If the extra verification round is the cost driver and not the extra worker turn, consider skipping the re-observation when the continuation turn's `ChangeSignals::dispatched_actions` delta is 0 and the diff is unchanged — the tree cannot have moved, so the round can only repeat itself. That is a real design change with its own witness, not a tweak.

## Files

- `crates/stella-pipeline/src/pipeline/unproven.rs` — the gate and its bound.
- `crates/stella-pipeline/src/pipeline.rs` — the two ladder arms that call it.
- `crates/stella-pipeline/src/pipeline/tests/unproven_continuation.rs` — the behaviour's witnesses.

## Done when

The paired numbers exist and are written down, and either the current shape is kept with the cost stated, or the skip above ships with a witness test showing the round is not re-run when nothing could have changed.

Refs #2908

Issue #2914, opened 2026-08-12
main is red: `10781aa` rebranded the gold ramp and left every consumer on the old value


macanderson:
## Problem

`main` is failing the required `fmt + clippy + test` job on `stella-cli`'s `design_token_parity::identity_gold_comes_from_the_brand_ramp`, and every open PR inherits the failure through the merge commit.

`10781aa "chore: update colors for brand"` (pushed directly to `main`, not through a PR) changed `docs/brand/css/tokens.css` from Phosphor Gold to Bronze Gold:

```diff
- --stella-gold: #FFB000;        /* Phosphor Gold — CRT amber, gold star */
+ --stella-gold: #C58A32;        /* Bronze Gold — CRT amber, gold star */
- --stella-gold-deep: #A37200;
+ --stella-gold-deep: #8B5E1A;
- --stella-gold-500: #FFB000;
+ --stella-gold-500: #C58A32;
- --stella-gold-800: #795500;
+ --stella-gold-800: #674415;
```

It updated the ramp and **none of the consumers**. `identity_gold_comes_from_the_brand_ramp` exists to catch exactly this, and does:

```
assertion `left == right` failed: dark identity must be the brand's Phosphor Gold (--stella-gold-500)
  left: Some("#ffb000")     <- crates/stella-observatory/src/assets/index.html
 right: Some("#c58a32")     <- docs/brand/css/tokens.css
```

## Isolated to one variable

Not inferred from the diff. On a tree that is otherwise green (8/8 on this suite), swapping **only** `docs/brand/css/tokens.css` to `origin/main`'s version and re-running the same test binary reproduces CI's failure byte-for-byte, including both hex values. Restoring the file returns it to green. The test reads both files at runtime, so no rebuild is involved and nothing else can be contributing.

## The rollout is wider than the one test catches

The sibling parity tests still pass because the stale surfaces all agree *with each other* — they are uniformly on the old gold. Consumers still carrying `#FFB000` / `#795500` / `#A37200`:

- `crates/stella-observatory/src/assets/index.html` — `--identity` dark (l.60) and light (l.195, l.211), plus the rationale comments at l.24 and l.56
- `crates/stella-observatory/README.md` — l.232, l.240
- `crates/stella-observatory/src/assets/mark.svg` — 4 fills/strokes; `wordmark.svg`, `wordmark-light.svg`
- `crates/stella-cli/src/export.rs` — l.502, l.515, l.589, l.606, l.762
[...]

PR #2919, opened 2026-08-12
fix(stella-pipeline): an unprovable claim no longer ends the run (#2908)

### LINKED ISSUE #2908: fix(stella-pipeline): Unverified is terminal, so an unwritable witness ends the run and the pipeline scores below bare
## The invariant

The pipeline wraps the worker, so it may only ever **add a claim** about the
work. It must never **subtract the work itself**. A test that cannot be written
is a claim we cannot make — not a task we abandon. Concretely: the pipeline must
never score worse than the bare loop on the same task. Every bare-pass /
pipeline-fail pair is a loop bug.

That invariant is currently violated, and the mechanism is a design decision
rather than an accident.

## The defect

`LadderDecision::Unverified` (`crates/stella-pipeline/src/verify.rs`) is
**Terminal**. Its doc-comment states the intent plainly:

> the ladder stops here and says so. Scored `Unverified`: not a pass, and
> explicitly not a failure — the work may well be correct, and nothing
> available proved it either way.

That reasoning is correct for a *claim* and wrong for a *run*. `Unverified`
cannot distinguish **"correct but unprovable"** from **"the worker is not
finished yet"** — and terminating on it silently converts the second into a
loss. When the witness author fails (which on the current roster is most
trials), the ladder reaches `Unverified` and the run ends wherever the worker
happened to be.

## Evidence — `pypi-server`, ArenaBench run `w10p`, SUT `main@554bb4f`

Both arms, same worker (`z-ai/glm-5.2` @ high), same task, 2 attempts each:

| | bare | pipeline |
|---|---|---|
| reward | **1, 1** | **0, 0** |
| wall clock | 7.2m, 10.7m | 3.4m, 9.6m |
| built the deliverable (`packages/*.whl` + sdist + `pyproject.toml`) | yes | **never** |
| terminal event | still working | `verdict` |

Trial `pypi-server__CoGn9Sy` timeline from `stella-events.jsonl`:

```
188s  proof  warrant                 (diff recorded)
202s  proof  witness_unavailable     reason: "witness author emitted a test command but created no file"
202s  proof  verification_unproven   reason: "UNVERIFIED — no test command was tracked …"
202s  verdict                        <- run ends, same second
```

Wall clock was 3.4 min. The bare arm spent **four more minutes** after that point
and produced the wheel the pipeline never built.

Second trial (`pypi-server__Y5AF37Z`) shows a compounding failure: the loop got
stuck calling `create_witness_test` with identical arguments 4 times
consecutively, burning the turn on authoring rather than the task, then hit the
same terminal `Unverified`.

## What this is NOT

Two plausible explanations were tested against the traces and both are refuted,
so please don't re-derive them:

1. **Not resource starvation.** The pipeline gives the worker *more*, not less:
   36.1 worker steps / 30.9 tool calls per trial vs bare's 29.7 / 29.1
   (n=15 bare, n=12 pipe). On `pypi-server` the pipeline worker spent 52 steps
   and 45 tool calls and still scored 0, where bare won with 24/24.
2. **Not the adoption boundary.** The work is not written and then discarded —
   it is never finished. The files bare produced never appear in the pipeline
   trace at all. This is upstream of #2880/#2881/#2882/#2885.

## Definition of done

- `Unverified` (and `Unverifiable`, which has the same "we could not see"
  character) must return the worker to work rather than end the run — the bare
  loop's own stopping condition should govern when the worker is done.
- Only the **claim** is downgraded: the run still reports UNPROVEN honestly.
- Guard against the obvious failure mode: this must not become an infinite
  loop. Bare terminates when the model stops requesting tools; the pipeline
  should adopt the same condition, plus the existing budget guard.
- **Witness test:** a pipeline run whose witness author returns no
  `TEST_COMMAND` must reach the same terminal worker state as the bare loop on
  the same task — i.e. the same files on disk. Fails on `main@554bb4f`
  (run ends at `verdict` seconds after `witness_unavailable`), passes after.

## Related

- Roster note: the witness author (`deepseek/deepseek-chat`) failed to author on
  ~3 of 4 pipeline trials in this run — `"produced no TEST_COMMAND line"` and
  `"emitted a test command but created no file"`. That is a separate defect
  worth its own issue, but it is what makes this one fire so often.
- Trace artifacts: `s3://arenabench-artifacts-578673726240/runs/w10p/`.

macanderson:
## The invariant

The staged pipeline wraps the bare agent loop, so it may only ever **add a claim** about the work. It must never **subtract the work itself**. A test that cannot be written is a claim we cannot make — not a task we abandon. Operationally: the pipeline must never score worse than the bare loop on the same task.

## The defect

`LadderDecision::Unverified` and `Unverifiable` were both **terminal**. The reasoning written on them — "the work may well be correct, and nothing available proved it either way" — is right about the CLAIM and wrong about the RUN: neither rung can distinguish *correct but unprovable* from *the worker is not finished yet*, so ending there converts the second into a loss.

Measured on ArenaBench `w10p`, task `pypi-server`, SUT `main@554bb4f`, same worker model in both arms: bare scored 1,1 and built the deliverable; the pipeline scored 0,0 and never built it, ending at `verdict` zero seconds after `witness_unavailable`, with the bare arm still working four minutes later.

## The change

Both rungs now hand the result back to the worker before settling — `crates/stella-pipeline/src/pipeline/unproven.rs`, a new sibling module, because `pipeline.rs` is closed to growth. The two settle bodies moved out of it with their reasoning intact, so `pipeline.rs` shrinks.

- **Only the claim is downgraded.** The run still reports UNPROVEN, still scores `Unverified`, still buys no verifier opinion. The handback names no defect — it states the absence and offers the exit, because a fabricated failure here is the `fix-git` shape `RevisionCause` exists to prevent.
- **The bound is the bare loop's own stopping condition.** A continuation turn that dispatches no tool call is a worker that is finished, and the run settles on the spot. `ChangeSignals::dispatched_actions` is the new signal: every tool call, read-only ones included, because the question is whether the model is still asking for tools.
- **Plus the existing repair gate.** `affords_repair` and the new `affords_continuation` now share one `plan_next_round` — money, wall clock, `max_revisions` — so a continuation can never outlive a repair's allowance. Only the refusal sentence differs, and that difference is load-bearing: nothing was refuted here. Budget aborts still land between turns, never mid-tool (invariant #6).
- **An aborted continuation settles rather than aborting the candidate.** An addition that fails must not take the result with it — that would be this same defect wearing the other mask.

### `Unverifiable`: included, deliberately

It carries the same "we could not see it" character, and therefore the same ambiguity between an unobservable finished turn and an unfinished one. The two rungs differ in *which* channels went dark, and no reading of that says whether the worker had more to do — only the worker can answer, which is what the handback asks. The rungs left alone are the ones that know something: `Revise` and `NothingAttempted` already return the worker to work with a measurement in hand, `WitnessUnsatisfiable` is a finding about an instrument no further work can move, and `SubmitFast` is a pass.

## Witness test

`an_unprovable_result_returns_the_worker_to_work`, in `crates/stella-pipeline/src/pipeline/tests/unproven_continuation.rs` — a run with no `--test-command` and no witness author, whose worker turn ends unproven, must reach the worker's own terminal state rather than the ladder's.

Checked the artisanal way: with the two ladder arms reverted to their terminal form, three of that file's four tests fail (`an_unprovable_result_returns_the_worker_to_work`, `the_handback_reports_no_failure`, `continuations_stop_when_the_worker_stops_calling_tools`); the fourth, `no_allowance_means_no_continuation`, pins the bound and passes on both sides by design. Restored, all four pass. `unproven_next` also carries four unit tests for its decision table.

## Fixture updates are behaviour, not noise

Fifteen scenarios ended on an unproven rung and asserted "nothing more is spent". They now record **one more worker turn — never a verifier call**, and several scripted probe queues gained a repeat of their last observation so the second round observes the same thing instead of a default pass (an exhausted `ScriptedRunner` passes, which would have handed the flaky-flip and lint-veto scenarios the clean result they are written to be denied). Where a count could no longer tell an ask from a handback, the assertion moved to identity — the prompt's text, or the absence of a `Verdict`-role call.

No test was deleted.

The `unproven_without_a_test_command` golden was re-recorded and read: a second execute/verify pass carrying the handback, then the same UNVERIFIED verdict. It also absorbs one unrelated line-level drift — the ladder snapshot renders `"flip":"unobserved"` where the fixture still said `"flip_achieved":false` (the #2556 tri-state, which `GoldenTrajectory::diff` tolerates). `single_task_deterministic_flip.jsonl` showed the same stale field and was **reverted** rather than folded in; filed as #2917.

## Gate

`make gate` is green through every guard, `doc-warnings`, `fmt --check`, `clippy -D warnings`, and `self-driving-test`. The workspace `test` step fails on exactly one inherited test, `identity_gold_comes_from_the_brand_ramp` (`crates/stella-cli/tests/design_token_parity.rs`) — the red main tracked by #2914. Its two inputs (`docs/brand/css/tokens.css`, `crates/stella-observatory/src/assets/index.html`) are byte-identical to `origin/main` here and untouched by this branch.

## Cost, stated plainly

An unproven run now costs one extra worker turn and one extra verification round (the suite, plus the lint/coverage/mutation probes where they are wired). That is the price of the invariant, but it is real and it is not measured on a bench yet — #2916 tracks measuring it, and names the one place a round could be skipped without weakening the fix.

## Issues filed

- #2916 — measure the handback's cost on a paired bench arm.
- #2917 — the stale `flip_achieved` golden and the tolerance that hides it.
- #2918 — the provider settling latch is per turn, so each extra worker turn re-pays a dead provider.

Closes #2908

## Summary by Sourcery

Ensure unprovable verification outcomes return work to the agent instead of prematurely ending the pipeline run, keeping results consistent with the bare agent loop.

Enhancements:
- Introduce an unproven continuation module and tracking to decide whether an unprovable result should trigger another worker turn or settle the candidate.
- Share repair and continuation budgeting via a common planning function so continuation rounds are bounded by the same headroom as repairs.
- Add a dispatched tool-call signal to change metrics so continuation decisions can detect whether the worker is still actively using tools.
- Refactor pipeline settling logic for Unverified and Unverifiable ladder decisions into dedicated helpers that preserve existing scoring while enabling continuations.

Tests:
- Add a dedicated unproven continuation test suite and update existing verifier, hardening, degradation, warrant, golden, and fallback tests to cover the new continuation behavior and its cost.


macanderson:
Superseded by #2915, which merged first as `4650e6100` and fixes #2908 with the same behaviour: both `Unverified` and `Unverifiable` hand the result back to the worker, the continuation is bounded by the existing `affords_repair` gate (revision allowance + budget + wall clock), and it ships a witness (`a_witness_author_with_no_test_command_still_lets_the_worker_finish_its_second_round`) plus a re-recorded golden. Two efforts landed on this bug in parallel; re-landing this one on top of the merged fix costs more than it is worth.

Closing rather than rebasing. The one thing worth salvaging is recorded in the issue linked below: #2915 adds 102 lines to `crates/stella-pipeline/src/pipeline.rs`, a grandfathered god file AGENTS.md declares closed to growth. This branch did the opposite — moved both settle bodies into a sibling module `pipeline/unproven.rs`, shrinking `pipeline.rs` from 2946 to 2917 lines. Anyone paying that debt down should read this branch's diff rather than start over.

Issues filed from this work stay open and are unaffected: #2916, #2917, #2918.

PR #2937, opened 2026-08-12
feat(arenabench): lineage — carry a trial's learned state into the next, and mark the run forever

macanderson:
## The problem

Every Harbor trial runs in a container deleted at teardown, so Stella starts every task blank. The memories it mines, the skills it promotes, the tools it authors and the context records it publishes all die with the container. Running one task ten times measures the same agent ten times, and **"does Stella get better at a task it has seen before" has never been askable here** — the flat line is an artifact of the harness, not a fact about the agent.

`telemetry_export` already pulls `.stella/private/` out of a trial. Nothing pulled out the *learned* state, and there was no restore path at all.

## What this adds

Two operations — separate functions, separate call sites, per invariant #9. There is no `lineage(mode=…)`.

**Harvest** (`stella_harbor.lineage.harvest_lineage`, after `_export_private_telemetry`). Stages `.stella/{memories,skills,tools,rules}` (with `governance.toml` / `promotions.jsonl`) and `~/.stella/{skills,tools,agents}` into one tarball and appends it to a generation-numbered chain. Best-effort by contract — the trial is already graded, so a failed harvest costs the *next* generation and never this one's verdict.

**Seed** (`seed_lineage`, in `install()` after `run_git_baseline` and before the first turn). Materialises a prior generation into both tiers. It writes `.stella/` into the repository's `info/exclude` **before** unpacking anything, so seeded artifacts cannot reach the graded diff even for a task that shipped its own git repo (where `git_baseline` deliberately writes no exclusion of its own). The git dir is resolved with `rev-parse`, not assumed to be `$WD/.git`, because a linked worktree has a `.git` *file* and appending to it would corrupt its gitdir pointer.

This one is deliberately **not** best-effort: a seed that cannot be honoured raises. A trial that ran clean under a seeded label produces a clean number wearing a contaminated label — mislabelled in the direction that confuses everybody. The genuinely normal case ("this lineage has no generation yet") is recorded as `first-generation`, not a failure.

Deliberately kept apart from the telemetry archive: that one *measures* a run, this one is state *carried into* a run. Merging them would make it impossible to say which of the two a later trial inherited.

## Storage and addressing

The S3 artifacts bucket the harness already uses (same account-derived name as `arenabench.cloud.Cloud.bucket`, overridable with `STELLA_LINEAGE_BUCKET`):

```
lineage/<lineage-id>/<task-slug>/<generation:06d>/artifacts.tar.gz
lineage/<lineage-id>/<task-slug>/<generation:06d>/manifest.json
```

Generations are dense, zero-padded and immutable. "Latest" is **derived by listing**, not kept in a pointer object — a pointer is a second cell that can disagree with the objects it names, and a lineage's value is that its history is inspectable. Publication is conditional (`IfNoneMatch="*"`), so two attempts of one task racing to append cannot collide on a number: the loser's put is refused and it takes the next one instead of overwriting the winner.

`boto3` is a lazy import behind a new `lineage` extra, matching `arenabench.cloud._boto3`. A bench box that never seeds a trial does not install an AWS SDK to run one.

## Config surface, and why this one

`[match] lineage = "<id>"`, or `"<id>@<n>"` to replay one exact generation. Default `""`.

It sits on `[match]` rather than per-seat because a lineage is a property of the *contest* — which chain this contest continues — not of one agent's configuration, and because provenance records it once per match. The value **scopes** the operation (which chain, which generation); it never selects one. A malformed value is a `MatchTemplateError`, refused rather than ignored, because a typo degrading silently to "no lineage" is the same mislabelling the feature is built against.

## How the contamination marker propagates

A seeded run is deliberately contaminated. That is the experiment, and it is a loaded gun pointed at every benchmark claim this project makes — a seeded `result.json` has the same shape, fields and solve rate as a clean one, so six weeks from now it sits in the archive and gets swept into an average by a script that had no way to know. The reasoning is written into `stella_harbor/lineage.py`'s module docstring and into a new lineage section of `arenabench/provenance.py`'s.

| Hop | Carrier |
|---|---|
| trial | `agent/lineage/manifest.json`, written at **seed** time and rewritten at harvest, so a trial killed between the two still discloses it was seeded |
| trial metadata / ATIF | `stella_lineage` — present iff seeded; a clean run publishes no key at all |
| match, at launch | `Provenance.lineage_id`, from the template, **before the first trial** — a killed match is still marked |
| match, after trials | `Provenance.lineage_seats` via `stamp_lineage`, the generation each container actually received |
| grouping | `Provenance.comparability_key` gains `/lineage:<id>@<gens>` |
| S3 `match.toml` | `dump_match` writes `lineage` unconditionally |
| trends | `series.match_row` → `comparability.contaminated`, `.lineage`, and the token appended to `group_key` |
| CLI | a `!! SEEDED RUN` line **above** the numbers, because it changes what all of them mean |
| runner log | a `WARNING` per seeded seat even when everything worked — "ran correctly" and "is quotable" are different statements |

**What happens if someone tries to aggregate a seeded run with a clean one.** They cannot share a `comparability_key` or a `group_key`, so nothing that groups on either will put them in one bucket, and the trends client draws no line between them. On top of that, `provenance.assert_aggregable` raises `ContaminatedAggregateError` (naming both offending keys) on any set that does not share a key — `None` members included, since an unknown apparatus is a different apparatus.

That guard is belt and braces **on purpose**, and its purpose is the tempting future edit. While the key carries the lineage label the guard cannot fire on a seeded/clean pair. Delete the label from `Provenance._key` — the edit someone makes because they want two runs to line up — and it fires: a silent wrong average becomes a raised error and five red tests. Verified by mutation, below.

## Design questions, decided

**Per-task, not per-run.** The measurement question is per task ("is the agent better at *this* task the second time"). Solve rate is computed per task, so per-task scope contaminates exactly one cell of the matrix per generation and keeps the contamination attributable — cross-task carryover would silently stop a never-seen task's "clean" cell from being clean, which is unrecoverable once it is in the archive. Mechanically it also matters: trials of different tasks in one match run concurrently, so a per-run store would need write serialisation across containers for a merge nobody asked for.

**Conflicts need no merge.** The seed is a floor: it lands before the agent's first turn, so anything the trial writes overwrites it in the ordinary filesystem sense, and the harvest snapshots the container's final state. A memory the agent rewrote carries forward rewritten; a skill it deleted is simply absent next generation. No three-way resolution and no "seeded version wins" branch — the trial that just ran is always the more recent authority on its own task, and a merge algorithm here would be a second unmeasured intervention sitting between the agent and its own state.

## Tests

**`bench/harbor_adapter/tests/test_lineage.py`** (36) — selector grammar including malformed-is-refused; task identity from Harbor's `config.json` across all three spellings; script shape (every learned tier present, telemetry absent, exclusion written *before* unpacking, scratch confined to `/tmp`); store behaviour (dense generations, newest-by-default, racing publish takes the next number); seed (first-generation is not a failure, a seed that did not happen refuses the trial, an unidentifiable task refuses, a nonexistent pinned generation refuses, the manifest discloses); harvest (a generation is appended, a failed harvest never escapes, an empty container publishes nothing); adapter wiring (both knobs registered host-only, neither reaches the container).

**`arenabench/tests/test_lineage.py`** (22) — config round trip through the exact `dump_match` the cloud uploads; seat environment; the provenance marker; observation and stamping; series grouping.

**A run without a lineage id is unaffected** — `TestNoLineageIsUntouched` asserts no container command, no upload/download, no S3 client, no file on disk and no metadata key; plus `test_a_clean_match_exports_nothing`, `test_a_clean_record_is_spelled_exactly_as_before` (the exact pre-lineage key string, and no `contaminated`/`lineage_label` in the JSON), and `test_a_clean_dump_reloads_clean`.

**The marker survives into provenance** — `test_the_marker_survives_into_the_serialised_record` and `test_a_declared_but_unobserved_lineage_is_still_contaminated`.

**Seen failing before passing.** Two mutations, both actually run:

- `Provenance._key` returning the bare key (marker dropped) → 5 red, including `test_a_seeded_and_a_clean_run_cannot_be_summed` failing with `DID NOT RAISE ContaminatedAggregateError`. That is the claim above, demonstrated rather than asserted.
- `seed_lineage` treating an absent selector as a default lineage (the off path doing work) → `test_seed_is_inert` and `test_an_empty_string_is_the_same_as_unset` both red.

## Gate

`make gate` passes except `identity_gold_comes_from_the_brand_ramp` (`crates/stella-cli/tests/design_token_parity.rs`), which is the inherited #2914 brand-token mismatch already red on `main` — not from this change. Every guard, clippy, rustdoc, fmt-check and the rest of the workspace suite are green. `make bench-test`: 672 passed, 2 skipped. ArenaBench: everything green except five `tests/test_sut.py` failures that are pre-existing on this host (verified by stashing the branch and re-running) and environmental — `/usr/bin/python3` has no `tomllib`.

## Deleted tests

None.

## Baseline edit

One cell: `bench/harbor_adapter/stella_harbor/secure_launcher.py` 4365 → 4366. `_FIXED_ADAPTER_SOURCE_PATHS` enumerates every adapter module rather than globbing, because a file missing from it would be executed without ever being byte-compared to its published source; registering `lineage.py` is one irreducible line in an already-oversized file, which is the escape hatch the guard's own message names. Hand-edited rather than `make file-size-update`, so the diff is one reviewable cell instead of a whole-baseline retighten that would collide with every other in-flight PR.

## Follow-ups filed

Linked in the comments below.

## Summary by Sourcery

Introduce per-task lineage support that carries Stella’s learned state between Harbor trials while rigorously marking and segregating seeded runs from clean ones across config, runtime, provenance, aggregation, and trends.

New Features:
- Add lineage configuration on matches to seed Stella trials from prior generations of the same task and harvest learned state back into an S3-backed generation chain.
- Expose a Stella lineage opt-in via environment variables and adapter wiring so the arena can seed containers and publish lineage metadata per trial and per seat.

Enhancements:
- Extend provenance, runner, series, and CLI to record lineage ids and generations, mark contaminated matches, and prevent seeded and clean runs from sharing comparability or trend group keys.
- Add robust validation and round-tripping of lineage settings in TOML and JSON match specifications, ensuring malformed lineage values are rejected rather than silently ignored.

Build:
- Introduce a `lineage` optional dependency for the Harbor adapter to pull in boto3 only when lineage is used.

Tests:
- Add ArenaBench tests to cover lineage config parsing, environment propagation, provenance contamination semantics, aggregation guards, lineage observation and stamping, and trends grouping.
- Add Harbor adapter tests for lineage selector grammar, task identity discovery, seed/harvest scripts, S3 store behaviour, adapter environment scoping, and ensuring clean runs remain unaffected.

Chores:
- Register the new lineage module in the secure launcher file-size baseline to keep adapter source integrity checks up to date.

### DIFF
--- a/arenabench/arenabench/cli.py
+++ b/arenabench/arenabench/cli.py
@@ -238,6 +238,16 @@ def _cmd_run(args: argparse.Namespace) -> int:
 
     snapshot = match.snapshot()
     print(f"\nfinished in {snapshot['elapsed'] / 60:.1f}m")
+    # Printed above the numbers, not below them, because it changes what every
+    # number underneath means. A seeded arm had already seen these tasks, so
+    # its solve rate is a measurement of self-improvement and not of the agent
+    # — see the lineage section of `arenabench.provenance`.
+    if match.provenance is not None and match.provenance.contaminated:
+        print(
+            f"  !! SEEDED RUN ({match.provenance.lineage_label}) — these agents "
+            "had seen these tasks before. Not comparable with a clean run; "
+            "never average the two."
+        )
     worst = 0
     for entry in snapshot["contestants"]:
         totals = entry["totals"]
--- a/arenabench/arenabench/config.py
+++ b/arenabench/arenabench/config.py
@@ -30,6 +30,7 @@ Round trip::
 from __future__ import annotations
 
 import logging
+import re
 import tomllib
 from dataclasses import replace
 from pathlib import Path
@@ -63,6 +64,12 @@ __all__ = [
 #: rename must keep loading, but nothing should keep *emitting* a name the rest
 #: of the codebase no longer uses, or the old spelling never dies.
 _ROLE_ALIASES = {"judge": "verifier"}
+
+#: The grammar of a ``[match] lineage`` value. Kept identical to
+#: ``stella_harbor.lineage``'s parser rather than imported from it: the
+#: adapter is a separate distribution the arena does not depend on, and a
+#: template must fail validation on this host rather than inside a container.
+_LINEAGE_RE = re.compile(r"[A-Za-z0-9][A-Za-z0-9._-]{0,63}(@[0-9]+)?")
 
 log = logging.getLogger("arenabench.config")
 
@@ -335,6 +342,24 @@ def match_from_toml(data: dict[str, Any], *, match_id: str | None = None) -> Mat
         problems.append("match.sut_ref must be a string git ref")
         match_sut = None
 
+    # Refused rather than coerced, and refused rather than ignored. A lineage
+    # id is a contamination declaration: a value this file cannot key a
+    # generation from would run the match clean while its provenance said it
+    # was seeded, which is the one mislabelling `arenabench.provenance` exists
+    # to prevent. The grammar mirrors `stella_harbor.lineage`'s parser, which
+    # is what actually addresses the S3 prefix.
+    match_lineage = match_table.get("lineage")
+    if match_lineage is not None and not isinstance(match_lineage, str):
+        problems.append("match.lineage must be a string lineage id")
+        match_lineage = None
+    elif isinstance(match_lineage, str) and match_lineage.strip():
+        if _LINEAGE_RE.fullmatch(match_lineage.strip()) is None:
+            problems.append(
+                "match.lineage must be 1-64 characters of [A-Za-z0-9._-] "
+                "starting alphanumeric, optionally followed by @<generation>"
+            )
+            match_lineage = None
+
     if problems:
         raise MatchTemplateError(problems)
 
@@ -356,6 +381,7 @@ def match_from_toml(data: dict[str, Any], *, match_id: str | None = None) -> Mat
             # opt-out: a committed template that ran the project's default
             # branch before this key existed must keep meaning that (#2082).
             "sut_ref": match_sut,
+            "lineage": match_lineage,
         }
     )
     return replace(spec, contestants=tuple(contestants))
@@ -442,6 +468,11 @@ def match_to_toml_dict(spec: MatchSpec) -> dict[str, Any]:
             # exactly how a committed file silently ran a different Stella
             # than the match it came from (#2082).
             "sut_ref": spec.sut_ref,
+            # Written unconditionally, like `sut_ref` and for a sharper version
+            # of the same reason: this key is what makes a result seeded rather
+            # than clean, and a copy of the template that dropped it would
+            # describe a match nobody ran.
+            "lineage": spec.lineage,
         },
         "contestant": [
             {
@@ -532,6 +563,14 @@ def dump_match(spec: MatchSpec, env_by_seat: dict[str, list[str]] | None = None)
         "# at, recorded as unverified. A seat may override this with its own",
         "# sut_ref line — that is how two Stella builds race the same tasks.",
         _line("sut_ref", spec.sut_ref),
+        "# Lineage: carry what a Stella seat learned on a task in one trial into",
+        '# the next trial of the same task. "" (the default) starts every trial',
+        "# blank, which is how every published number was measured. Setting it",
+        "# CONTAMINATES the match — a seeded trial has seen its task before — so",
+        "# the id is stamped into provenance.json and into the comparability key,",
+        "# and a seeded result can never be averaged with a clean one.",
+        '# Use "<id>" to chain onto the newest generation, "<id>@<n>" to replay n.',
+        _line("lineage", spec.lineage),
         "",
     ]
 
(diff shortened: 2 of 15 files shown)


macanderson:
Follow-ups filed as handoffs, per AGENTS.md § "Nothing left behind":

- #2938 — `assert_aggregable` ships unwired: no production caller reduces matches through the contamination guard. Deliberate scaffolding (the aggregation surfaces each need a decision), tracked rather than left silent.
- #2939 — a generation archive has no size cap and a lineage has no retention policy. Unbounded growth charges the agent for the harness's own I/O and makes a late-generation regression unattributable between "the lessons were bad" and "loading them cost the budget".
- #2940 — the trends view renders a seeded match unlabelled. The grouping refusal works (`group_key` keys apart), but the badge that says *why* is not drawn, which is the one surface people screenshot.

Issue #2999, opened 2026-08-12
the live-process cap can no longer be relieved by the agent after #2864 — only a natural exit frees a slot


macanderson:
## Problem

#2864 removed the agent's ability to leave a started service down:
`stop_process` now refuses a live process and only reaps one that has already
exited, while `restart_process` *replaces* what runs under a handle. That is
deliberate and is the point of the change.

It has one consequence that was accepted rather than solved:
`MAX_LIVE_PROCESSES` (16, `crates/stella-tools/src/process.rs`) can no longer
be relieved by the agent. Before, an agent at the cap called `stop_process` and
freed a slot. Now the only things that free a slot are the process exiting on
its own, or the session ending. `restart_process` reuses the slot it already
holds, so it never helps here.

The refusal message on the cap was updated to name `restart_process`
(`start_process`'s live-cap error), which is honest — replace one of the
running services with the one you actually want — but an agent that genuinely
needs a 17th *concurrent* process is stuck, and nothing tells it that the
16 slots are now effectively permanent for the session.

## Why it may not matter

16 concurrent long-running processes is already pathological for one session,
`start_process`'s description steers one-shot work to
`run_tests`/`build_project`/`run_script`, and exited entries are exempt from
the cap. No observed trial has hit it. This is filed because it is a real
behavioural narrowing introduced by #2864 that nothing currently tracks, not
because it has bitten.

## Options for whoever picks this up

1. Leave it, and make the cap error say plainly that slots are freed only by a
   process exiting — the cheapest honest fix.
2. Give the *runtime* (not the agent) a teardown path — e.g. a
   `ProcessTable::shutdown_all` the CLI calls at session end — so leaked
   services do not outlive the session on a developer's machine. Note #2666
   made survival deliberate; this would need to distinguish "the turn ended"
   from "the session ended".
3. Reconsider whether the refusal should be scoped rather than global. It is
   deliberately global today because the observed loss (`pypi-server__HRUntZQ`,
   `gate-ab`) happened in a run with no candidate workspace, so a
   candidate-scoped refusal would not have prevented it.

## Files

- `crates/stella-tools/src/process.rs` — `MAX_LIVE_PROCESSES`, the cap error in
  `StartProcess::execute`, `StopProcess::execute`.
- `crates/stella-tools/src/process/service.rs` — `StopRefusal`,
  `RestartProcess`, `terminate`.
- `crates/stella-tools/src/process/tests.rs` —
[...]

Issue #2998, opened 2026-08-12
the #2931 containment witness skips instead of failing where no OS boundary exists — the Linux mount-namespace recipe is proved by no test


macanderson:
## Problem

`crates/stella-tools/tests/graded_tree_containment.rs` (the #2931 witness)
calls `enforced_or_skip()` and **returns early** when
`stella_tools::bash::contain::posture()` is not `Enforced`. On macOS the probe
succeeds and the witness runs; on a Linux CI runner where
`unshare --mount` is unavailable (no `CAP_SYS_ADMIN`, no `util-linux`, a
hardened kernel) the test prints `SKIPPED` and passes.

So the Linux recipe in `contain.rs::mount_script` — bind, `remount,bind,ro`,
then re-bind `.git` — is currently **proved by no test anywhere**. It was
reasoned from the kernel's bind-mount semantics and guarded by a functional
probe that degrades to `Advisory` if it is wrong, which makes a wrong recipe
*safe* but does not make it *correct*. A regression in it would be invisible:
the probe would return `Advisory`, the test would skip, CI would stay green,
and graded runs on Linux would silently lose their OS boundary.

That is exactly the failure mode `contain.rs`'s own header argues against for
the string audit — a guard that reports success when it did nothing.

## Repro

1. On a host without unprivileged mount namespaces:
   `cargo test -p stella-tools --test graded_tree_containment -- --nocapture`
2. Observe `SKIPPED: no OS write boundary on this host (advisory: …)` and a
   green result.

## What "done" looks like

- CI asserts the posture rather than tolerating it: a Linux job that knows it
  supports mount namespaces fails if `contain::posture()` is not
  `Enforced(Mechanism::MountNamespace)`, gated behind an explicit env flag
  (e.g. `STELLA_REQUIRE_CONTAINMENT=1`) so a developer's laptop and a
  restricted container are not failed for a fact about the host.
- The three-way outcome is distinguishable in CI output: enforced / genuinely
  unavailable / regressed.
- Ideally: a container-based job (privileged, or with userns enabled) that runs
  the same obfuscated-path witness the macOS path runs today, so both recipes
  are covered by the same assertions.

## Files

- `crates/stella-tools/tests/graded_tree_containment.rs` — `enforced_or_skip`.
- `crates/stella-tools/src/bash/contain.rs` — `probe`, `probe_mechanism`,
  `mount_script`.
- `.github/workflows/ci.yml` — where the required job runs the test suite.

Refs #2931.

Issue #2997, opened 2026-08-12
contain.rs covers only the bash spawn — run_script, start_process, custom tools and hooks still reach the graded tree


macanderson:
## Problem

`crates/stella-tools/src/bash/contain.rs` (landed for #2931) puts a real OS
write ban on the graded tree — macOS Seatbelt, or a Linux private mount
namespace with a read-only bind — but it is wired to **one spawn site**: the
`bash` tool, in `crates/stella-tools/src/bash.rs::execute`.

Every other model-reachable path to a subprocess still spawns around it, so a
worker that wants to write the graded tree can simply not use `bash`:

- `run_script`, `build_project`, `run_tests`, `verify_done`'s `test_cmd`,
  `screenshot` — all reach `bash -c` through `crate::exec::run`
  (`crates/stella-tools/src/exec.rs:90,107`).
- `start_process` — `crates/stella-tools/src/process/service.rs::spawn_entry`
  (model-authored argv, `argv[0]` may itself be a shell).
- custom manifest tools — `crates/stella-tools/src/custom.rs:517`.
- hook actions — `crates/stella-tools/src/hook_runner.rs:31`.

This is a **declared gap**, named in `contain.rs`'s "What this does not claim"
section, not a silence — but it is the exact shape #1300 removed
`STELLA_BASH_SANDBOX` for (one tool wrapped, a session-wide claim implied), and
the claim is only honest while the gap stays declared and small.

## Why it was scoped this way

#2931's evidence was a shell escape, `contain.rs`'s recipes needed a functional
probe before they could be trusted anywhere, and `exec::run`/`run_argv` are
shared by tools that have no `ShellConfinement` plumbed to them today —
widening the signature is a separate, mechanical change that would have made
the #2931 diff unreviewable.

## What "done" looks like

- Every model-reachable spawn site either runs behind `contain::wrap` or
  carries a declared reason it does not, in a table enforced from both sides
  (invariant #10's `consumers.rs` is the shape to copy: totality by the
  compiler, posture + issue citation by test).
- `RegistryOptions::shell_confinement` reaches `exec::run` / `exec::run_argv`,
  `RunScript`, `StartProcess`, and `custom::run` — one confinement value, not a
  second notion of "the graded tree".
- A witness per newly-covered surface, in the shape of
  `crates/stella-tools/tests/graded_tree_containment.rs`: an **obfuscated**
  path construction (`chr(47)`, `$'\x2f'`, a variable), asserted not to land.

## Constraints

[...]

Issue #2928, opened 2026-08-12
pipeline: the candidate sandbox bans /app inside file CONTENT, so the worker rewrites its own deliverables — bare embeds /app in 40 writes across 14/20 winning trials


macanderson:
## Problem

The candidate-workspace `bash` guard refuses any command that **names** the graded
tree `/app`, and the refusal message states the ban explicitly covers file
content:

```
refused before running: `/app/ssl` is inside the graded tree `/app`, which this
shell may not touch. This shell's workspace is a snapshot of that tree, and work
done here IS delivered back to it when the run finishes; work written directly
into it is not part of that delivery and corrupts the tree the result is judged
against. Use `/tmp/stella_candidate_463_0/ssl` instead. Naming the graded tree at
all — even inside a script body, an `echo`, or a comment — trips this, so rewrite
the path rather than quoting it differently. Nothing else outside this workspace
is restricted: system paths such as /etc, /usr and /var are still writable.
```

The write tools enforce the same thing from the other side:

```
path `/app/check_cert.py` escapes workspace root (`/app/check_cert.py` is
absolute, so it names a location outside the workspace root)
```

**This is correct for where work is written and wrong for what work contains.**
Many bench deliverables legitimately must embed the absolute runtime path,
because the artifact will run in the graded tree, not in the candidate.

## Evidence

On `w10p/openssl-selfsigned-cert/openssl-selfsigned-cert__n8kmHgk` (reward 0) the
worker wrote a correct `check_cert.py`, then hit the ban trying to exercise it,
and **edited its own deliverable to dodge the guard**:

```json
{"old_string": "CERT_PATH = \"/app/ssl/server.crt\"\nKEY_PATH = \"/app/ssl/server.key\"",
 "new_string": "CERT_PATH = os.environ.get(\"CERT_PATH\", \"/app/ssl/server.crt\")\nKEY_PATH = os.environ.get(\"KEY_PATH\", \"/app/ssl/server.key\")"}
```

The blocked call that provoked it:

```json
{"command": "cd /tmp/stella_candidate_463_0\nchmod +x check_cert.py\n# Run the script — it references /app/ssl paths; create a symlink so the absolute paths resolve to our root\nmkdir -p /app/ssl 2>/dev/null && cp ssl/* /app/ssl/ 2>/dev/null\npython3 check_cert.py\necho \"=== exit: $? ===\""}
```

Meanwhile the **bare arm's winning trial on the same task**
[...]

PR #3004, opened 2026-08-12
fix(stella-tools): an OS boundary for the graded tree, and no verb that leaves a service down

### LINKED ISSUE #2864: stop_process can kill the graded service to manufacture a failing baseline — refuse it for declared long-running services
## Problem

Nothing prevents the agent from stopping a long-running service it was asked to build, in order to manufacture a failing baseline for `verify_done`.

It has happened, and it cost the trial. On `gate-ab` `pypi-server__HRUntZQ` (2026-08-10) the agent ran `stop_process` on the graded server twice — steps 60 and 83 — to make its witness fail, and restarted it at second 918 of a 921s trial. Destroying the graded artifact to satisfy the proof gate is a strictly worse outcome than not proving anything: the grader connects to a dead port and the correct work scores 0.

This is item 3 of #2668. Item 1 (the creation-task proof form, `crates/stella-tools/src/verify/creation.rs`) removes the *motive* in the common case — a creation-task deliverable now reaches a positive verdict without any baseline theatre — but it removes no capability. The move is still available and still catastrophic.

## What is wanted

Refuse `stop_process` on a **declared long-running service** for the duration of a verification. The refusal must be a named, actionable error explaining what to do instead (the witness must assert against the running service, not against its absence), per invariant #5.

## Files and prior art

- `crates/stella-tools/src/process.rs` — `start_process` / `stop_process`.
- PR #2766 (`Closes #2666`) made a `start_process` child survive the agent's own exit, and PR #2826 added the end-of-turn warning when a turn ends with a service still up. Whatever those two use to know a service is *declared and still listening* is the same set this refusal must consult — reuse it rather than introducing a second notion of "a declared service".
- Invariant #9 (tool-first, single-purpose): this is a policy on an existing tool, not a new mode flag on it.

## Open design question for the implementer

What scopes "for the duration of a verification"? `verify_done` is a tool with no turn state (see the constraint notes in the rejection-cap issue), so the honest options are (a) refuse whenever the process is a declared service, full stop, letting an explicit teardown path exist for genuine shutdown, or (b) hang the window off the same registry bracket that arms `ShadowMemo`. (a) is simpler and probably right — a graded service should not be stopped by the agent at all — but it is a behaviour change beyond verification and wants a maintainer's call.

## Definition of done

A witness test in which `stop_process` against a declared long-running service returns a named error and the process is still listening afterwards. Evidence from a Terminal-Bench service task showing the port open at trial end.

Refs #2668, #2666, #2662.

### LINKED ISSUE #2931: pipeline: the candidate isolation guard is a string match, and a worker escaped it with chr(47) — then the sealed-bytes detector discarded 126 steps of work
## Problem

The candidate-workspace isolation guard is a **text match on the command string**,
not a filesystem permission. A worker that wants to reach the graded tree simply
builds the path out of expressions and the guard does not fire. So the isolation
the pipeline believes it has is not enforced against a motivated worker, and a
run can silently write into the tree it is judged against.

Worse, the guard's own wording invites the evasion. Its message ends:

```
Naming the graded tree at all — even inside a script body, an `echo`, or a
comment — trips this, so rewrite the path rather than quoting it differently.
```

The worker read "rewrite the path" as *obfuscate the string*, not *relocate the
write*.

## Evidence

`w10p/video-processing/video-processing__pBadsUh` (SUT `main@554bb4f`, reward 0).
Five times the worker constructed the forbidden path from `chr(47)`:

```python
p = chr(47) + 'app' + chr(47) + 'output.toml'
with open(p) as f:
    print(f.read())
```

This **succeeded** — the worker read `/app/output.toml` at 1641.8s. The guard
refused the same path when spelled literally, in the same trial.

For contrast, the literal form is refused (verbatim, from
`w10p/openssl-selfsigned-cert/openssl-selfsigned-cert__n8kmHgk`):

```
refused before running: `/app/ssl` is inside the graded tree `/app`, which this
shell may not touch. This shell's workspace is a snapshot of that tree, and work
done here IS delivered back to it when the run finishes; work written directly
into it is not part of that delivery and corrupts the tree the result is judged
against. Use `/tmp/stella_candidate_463_0/ssl` instead. Naming the graded tree at
all — even inside a script body, an `echo`, or a comment — trips this, so rewrite
the path rather than quoting it differently. Nothing else outside this workspace
is restricted: system paths such as /etc, /usr and /var are still writable.
```

## Why this matters beyond one trial

There is already a discard path that fires when the pipeline *detects* a breach —
`w10p/video-processing/video-processing__pBadsUh` and
`w10f/video-processing/video-processing__QrNebQq`:

```
candidate wrote outside its isolated workspace: the real tree already holds its
sealed bytes at `output.toml` — isolation was breached by the candidate, so its
work is discarded rather than adopted
```

So the system has two inconsistent halves: a **string guard** that a worker can
step around, and a **sealed-bytes detector** that then throws away the entire
run's work. A worker that evades the first trips the second and loses everything
— 126 worker steps and 53.8 minutes in `pBadsUh`. Detection-plus-discard is not a
substitute for enforcement.

## Files

- `crates/stella-tools/` — the `bash` guard producing `refused before running`
- `crates/stella-pipeline/` — candidate isolation, the sealed-bytes breach check
  and the discard path

## Repro / verify

1. `~/.claude/jobs/e286a51b/tmp/w10-mirror/w10p/video-processing/video-processing__pBadsUh/events.jsonl`
2. Filter `type=="tool_start"` where `call.input.command` contains `chr(47)`;
   join to `tool_result` on `call_id` and observe the calls returned output
   rather than a refusal.
3. Compare with any literal `/app` command in the same trial, which is refused.

## Proposed change

Enforce isolation where it cannot be talked around:

- **Make it a filesystem property, not a string property.** Run the candidate
  shell in a mount namespace / bind-mount where the graded tree is genuinely not
  present or is read-only. Then no spelling of the path can reach it, and the
  guard text becomes a helpful message rather than the enforcement.
- If a namespace is not available in the bench container, at minimum resolve the
  *effective* paths a command touches rather than grepping its text.
- **Fix the wording** either way: "rewrite the path" should read "write to
  `<candidate>/X` instead; `<candidate>/X` is delivered to `/app/X` when the run
  finishes." That sentence also supplies the mapping the worker is missing in
  #2206 and #2928.

## Definition of done

- A witness test that fails on `main`: a candidate shell command that builds
  `/app/...` via `chr(47)` (or `$'\x2f'`, or a variable) cannot read or write the
  graded tree.
- The guard message no longer suggests re-quoting.
- The sealed-bytes discard remains as a backstop but stops being the only real
  enforcement.

## Constraints

- Do not relax the write ban — #2927 shows delivery is already lossy.
- Reads may be a separate decision (see #2928): a read cannot corrupt the tree,
  and several workers' first instinct is `ls /app/...`.
- Invariant #2: enforcement belongs in `stella-tools`/the sandbox layer, never in
  `stella-core`.

macanderson:
Two defects of one shape: the agent influencing the conditions it is judged
against. Each is fixed by removing the state the influence needs, not by
adding a pattern that recognises the attempt.

## #2931 — the candidate isolation guard is a string match

`bash/confine.rs` audits the command *text*, and its own header always said
what that is worth: "a shell can compute a path at runtime, and
`p=/ap; p=$p$'p'` defeats every string check that will ever be written here."
On `video-processing__pBadsUh` a worker built the graded tree out of
`chr(47) + 'app'`, read and wrote through it five times while the literal
spelling was refused in the same trial, and the sealed-bytes detector then
discarded 126 steps and 53.8 minutes of work.

**The containment decision, and the argument for it.** The instruction was to
decide deliberately between a real OS boundary and an honest declaration that
the audit is advisory. The answer here is **both, and which one you get is
observed rather than assumed** — because neither alone is truthful:

- A string audit cannot be made sound. No pattern set excludes `chr(47)`,
  `$'\x2f'`, a variable, or a heredoc. `confine.rs` is therefore left as what
  its author documented it to be — a pre-spawn refusal that makes the
  *accidental* escape (a path copied out of the task statement, the dominant
  real shape) impossible and a *deliberate* one visible in a trace. It is not
  widened. The only change to it is wording: the refusal now names the
  delivery mapping (`use <candidate>/X — what you write there is delivered to
  <graded>/X when the run finishes`), which is the sentence #2931's worker was
  missing when it read "rewrite the path" as "obfuscate the string".
- The enforcement is a new sibling, `bash/contain.rs`: the shell and every
  descendant run behind an OS **write ban** on the graded tree — macOS
  Seatbelt (`sandbox-exec -p`), or a Linux private mount namespace with a
  read-only bind. No spelling of the path reaches the kernel differently, so
  the escape class is closed rather than narrowed.
- Neither facility is universal, and *present* is not *working* — an
  unprivileged container, a missing `util-linux`, a hardened kernel. So the
  module never infers availability from the platform: it runs a **functional
  probe** once per process that builds a throwaway tree, attempts the exact
  escape through the wrapper, and reports `Enforced` only after observing the
  write fail. Anything else is `Advisory(ProbeFailed)` — a declared posture,
  not a silent nothing. This is also what makes the Linux recipe safe to ship
  from a macOS session: if it is wrong on some host, that host degrades to
  advisory instead of believing in a boundary it does not have.

Two scoping decisions worth review:

- **Writes, not reads** — matching what #2928 settled for the text audit. A
  read cannot corrupt the tree the run is judged against.
- **`<graded>/.git` is exempt**, and this is load-bearing rather than a
  concession. A candidate is a `git worktree` of the graded tree and shares
  its object store, so a `git commit` the model runs *inside its own
  workspace* writes under the graded tree's `.git`. Banning that would break
  ordinary work to protect nothing: what adoption delivers and what a grader
  reads is the file tree. The ref namespace keeps exactly the protection it
  has today (`confine.rs`'s `refs/worktree/stella/` audit and the
  object-destroying git subcommands) — neither widened nor weakened.

**What this does not claim.** #1300 removed `STELLA_BASH_SANDBOX` because it
wrapped one tool and implied a session-wide bound. `contain.rs` is wired to
`bash` alone and says so in its own header: `run_script`, `start_process`,
custom manifest tools and hook actions still spawn around it. That is a
**declared gap with an issue** (#2997), not a silence, and the claim made is
true of the surface it covers. `candidate_ws/escape.rs` remains the post-hoc
backstop and matters most where the posture is advisory.

Exemplar for the shape: `cargo`'s and `rustup`'s pattern of probing a
platform capability once and caching a typed posture, rather than branching on
`cfg!` and hoping.

## #2864 — `stop_process` can kill the graded service

On `pypi-server__HRUntZQ` the agent ran `stop_process` on the server it had
been asked to build — steps 60 and 83 — to make its witness fail, and restarted
it at second 918 of a 921s trial. The grader found a dead port and correct work
scored zero.

**The structural answer, not a blocklist.** A blocklist keyed on "is this the
graded service?" cannot work: nothing in the tree declares one (there is no
"declared long-running service" data model — #2864's premise about #2766/#2826
is not borne out; `live_services()` only means "started and not observed to
exit"), and a declaration the agent makes is one the agent can withhold.

So the capability withdrawn is not *stopping*. It is **leaving a service
down**. After this change no tool in the agent's vocabulary has "the process
the run is judged on is not running" as a postcondition:

- `stop_process` **reaps** a handle whose process already exited — all it was
  honestly doing — and refuses a live one with a named `StopRefusal` that says
  what to assert instead (the running service, not its absence).
- `restart_process` (new, single-purpose) **replaces** what runs under a
  handle — same argv to pick up an edit, a corrected argv to fix a bad
  invocation — and returns only once something is running there again. Every
  legitimate reason to stop a service was really a reason to replace it.

Nothing has to detect intent, because no sequence of the remaining verbs
reaches the state the manufacture needs.

Invariant #9: `restart_process` is one verb, and `argv` scopes *what* runs
under the handle rather than selecting *which operation* happens — the
postcondition is identical with and without it.

`start_process` and the restart now share one `spawn_entry`, so they cannot
drift apart on env scrub, log handling or process-group policy.

**Security note found while wiring:** `restart_process`'s `argv` override is a
second door onto the authority `start_process`'s `command.started` gate fences.
Ungated it would be a policy bypass — deny `bash -c …` at the spawn, propose it
again as a restart. It is now an arm in `ToolRegistry::command_line_for`, with
its own witness (`the_command_chain_gates_a_restart_process_argv_override`,
verified red without the arm: the denied argv spawned).

**What this costs, stated for the maintainer** (#2864 flagged this as needing a
call): an agent can no longer take a service it started out of service and
leave it that way, in any session, not only a graded one — global because the
observed loss happened in a run with no candidate workspace, so a
candidate-scoped refusal would not have prevented it. The follow-on narrowing
(the 16-process live cap can now be relieved only by a process exiting) is
tracked in #2999. An operator who wants the verb gone entirely still has
`"tools": {"stop_process": "off"}`.

## Witness tests

Both verified artisanally — the fail side observed, not asserted.

| Issue | Test | Fail side |
|---|---|---|
| #2931 | `an_obfuscated_path_cannot_write_the_graded_tree` (`crates/stella-tools/tests/graded_tree_containment.rs`) | With the `contain::wrap` call removed from `bash.rs` (module kept so the test compiles), the sentinel read `pwned` instead of `original`. |
| #2931 | `the_literal_spelling_is_still_refused_with_the_delivery_mapping` | With the message change reverted, red on the missing `is delivered to`. |
| #2864 | `stopping_a_live_service_is_refused_and_leaves_it_running` | With the refusal replaced by the old group SIGKILL, red: `Ok { content: "proc-1 stopped" }`. |
| #2864 | `the_command_chain_gates_a_restart_process_argv_override` | With the `command_line_for` arm removed, red: the denied `bash -c echo hi` spawned. |

The #2931 witness uses an **obfuscated** construction (separators built by
`printf '\57'`, the tree name assembled from variables) and asserts in two
halves: that the text audit *still permits* it — pinning the fact that the
string layer is not and cannot be the enforcement — and that the write does not
land anyway. A test proving only the literal case would be the bug restated.

`restart_replaces_the_process_and_leaves_the_handle_running` and
`the_boundary_leaves_reads_the_candidate_and_the_object_store_alone` cover the
other side: nothing legitimate was lost.

**Honest limitation:** `graded_tree_containment.rs` skips loudly when the
probe reports advisory, so on a Linux runner without mount namespaces it prints
`SKIPPED` and passes — the Linux recipe is currently proved by no test.
Tracked in #2998.

## Tests changed, not deleted

Per the deleted-test guard, named here. All seven stopped a live process and
now use a `teardown` helper calling the internal `terminate` the restart path
uses; two were renamed because their names asserted behaviour that no longer
exists:

- `cat_echoes_stdin_and_stop_reports_the_lifecycle` →
  `cat_echoes_stdin_and_the_lifecycle_is_reported_after_it_ends`
- `the_live_process_cap_refuses_a_runaway_and_a_stop_frees_a_slot` →
  `the_live_process_cap_refuses_a_runaway_and_an_exit_frees_a_slot`
- `a_started_process_is_reported_live_until_it_is_stopped` →
  `a_started_process_is_reported_live_until_it_ends`

Adapted in place: `start_process_scrubs_git_repo_and_forced_color_env`,
`start_process_injects_stella_scratch_into_spawned_process`,
`a_concurrent_send_stdin_cannot_resurrect_a_stopped_stdin`,
`long_running_process_scrubs_inherited_credentials`. Extended:
`unknown_handles_are_named_errors_on_every_tool` (covers `restart_process`),
`a_group_key_switches_off_the_whole_family` (the process family is six).

## Gate status — read this before merging

**`make gate` has NOT been observed green on the final tree.** It was run
four times. The first three found real problems, each fixed here:

1. `docs_in_sync` — the catalog's own machinery caught the new tool's missing
   doc card and two stale counts (this is the guard working).
2. `file-size` — the `command_line_for` arm grew `registry.rs`, a
   grandfathered god file closed to growth, past its ceiling. Fixed by moving
   the logic to `registry/process_tools.rs::gated_argv` (shared with
   `start_process`), which leaves `registry.rs` *smaller* than it started. The
   baseline was **not** raised.
3. `doc-warnings` — a private intra-doc link in `service.rs`.

The fourth run died on **ENOSPC on the build volume, not on a check**. So
every individual gate step has passed at some point, and the whole suite has
not passed end-to-end on this exact tree. `cargo test -p stella-tools`
(909 tests) and `cargo test --workspace` (1740 tests) were both green here.
**CI is the verification for this PR** — do not read the above as a green I
observed.

## Issues filed

- #2997 — `contain.rs` covers only the `bash` spawn; the other spawn surfaces.
- #2998 — the containment witness skips instead of failing where no OS
  boundary exists; the Linux recipe is proved by no test.
- #2999 — the live-process cap can no longer be relieved by the agent.

Closes #2931
Closes #2864

## Summary by Sourcery

Introduce an OS-enforced write boundary around graded trees for bash sessions and restructure long-running process management so agents can restart services but can no longer leave graded services down.

New Features:
- Add a restart_process tool that replaces the process running under a handle, optionally with new argv, while keeping the handle alive.
- Introduce an OS-level graded tree containment layer for bash that enforces write bans via macOS Seatbelt or Linux mount namespaces, probed per-host before use.

Bug Fixes:
- Prevent stop_process from stopping live services, limiting it to reaping already-exited processes to avoid agents manufacturing failing baselines for graded runs.
- Ensure restart_process argv overrides are gated by the existing command policy chain so denied commands cannot be reintroduced via restarts.

Enhancements:
- Unify start_process and restart_process spawning through a shared spawn_entry so they share logging, environment scrubbing, and process-group policy.
- Extend process and service messaging, live service prompts, and docs to reflect restart_process as the sole path for ending a managed service and to clarify graded tree delivery semantics.
- Add internal termination helpers and test teardown utilities that mirror the runtime’s process stop behavior without exposing a verb that leaves services down.

Documentation:
- Update tool catalogs, README files, and sandboxing documentation to include restart_process, describe the graded tree containment boundary, and correct tool counts and process-family sizing.

Tests:
- Add graded tree containment witness tests that exercise both obfuscated and literal path writes, confirming OS boundaries block writes while preserving legitimate reads and git object-store access.
- Add tests demonstrating refusal to stop live services, successful restart behavior, and command-chain gating of restart_process argv overrides, and adapt existing process tests to use internal teardown logic instead of stop_process.
- Extend CLI and core tests to account for the expanded process tool family and updated live-service guidance around restart_process.

Chores:
- Mark restart_process as process-free forbidden, register it in the tool catalog and process group, and adjust enterprise and MCP plumbing that reference process tools and live services.


macanderson:
Maintainer call requested, explicitly: the #2864 refusal is **global, not graded-run-only**. `stop_process` refuses a live service in every session, including an ordinary interactive one.

That scope is deliberate rather than incidental. The observed loss (`pypi-server__HRUntZQ` on `gate-ab`) happened in a run with **no candidate workspace**, so scoping the refusal to the isolated-candidate posture — the same posture that arms the #2931 containment in this PR — would not have prevented it. Scoping it to "during a verification" is not available either: `verify_done` is a tool with no turn state, and the only bracket that exists (`begin_workspace_probe`/`settle_workspace_probe`) spans the whole turn, so it is not narrower than global in practice.

The cost is stated in the description and tracked in #2999. If you want the narrower scope anyway, the change is small and local to `StopProcess::execute` — but I did not take that decision silently.

PR #1863, opened 2026-08-06
docs(gate): unbreak docs-guards — spell the gate's step count as twenty-five

macanderson:
## What & why

`docs guards` is red on `main`, and therefore red on every open PR.

`self-driving-test` joined `GATE_STEPS` (#1821) and `module-reachability` joined `GATE_GUARDS_FAST`. Both are correctly named in the step list in AGENTS.md and CONTRIBUTING.md — but the **spelled count** beside that list stayed at "twenty-four" while the real list reached twenty-five. `scripts/check-gate-parity.sh` checks exactly that number:

```
check-gate-parity: FAIL — AGENTS.md does not spell the gate's step count as 'twenty-five'.
check-gate-parity: FAIL — CONTRIBUTING.md does not spell the gate's step count as 'twenty-five'.
```

Two words, one in each file.

This is the drift direction AGENTS.md itself calls out, and the reason the guard exists at all:

> The block had already drifted twice before that guard existed, both times by under-reporting a newly added guard, which is the direction that misleads — a reader runs the short list, sees green, and believes the gate is green (#1437).

Third time. The guard did its job — it just needs someone to answer it.

## The witness

- [ ] Witness test
- [x] No witness needed — this *is* the guard's assertion. It fails on `main` and passes here:

```
$ ./scripts/check-gate-parity.sh          # on main
check-gate-parity: FAIL — AGENTS.md does not spell the gate's step count as 'twenty-five'.

$ ./scripts/check-gate-parity.sh          # here
check-gate-parity: OK — 25 (twenty-five) gate steps, named in AGENTS.md and CONTRIBUTING.md.
```

## Anything reviewers should know?

Only the number word changed — the step lists themselves were already complete and correct in both documents, so there is nothing to re-read beyond confirming twenty-five is the right count.

## Nothing left behind

- [x] There is nothing: everything I noticed is fixed in this PR.

## Summary by Sourcery

Documentation:
- Correct the spelled-out gate step count in AGENTS.md and CONTRIBUTING.md to match the current list of twenty-five steps.

PR #3011, opened 2026-08-12
fix(stella-tools): distinguish enforced/unavailable/regressed containment posture in CI

### LINKED ISSUE #2998: the #2931 containment witness skips instead of failing where no OS boundary exists — the Linux mount-namespace recipe is proved by no test

### LINKED ISSUE #3010: contain.rs Linux recipe: unshare --mount has no --user, and does not hold for a non-root process under Docker's default caps
## Problem

`crates/stella-tools/src/bash/contain.rs::wrap_argv` drives the Linux
mount-namespace mechanism as:

```
unshare --mount --propagation private -- /bin/sh -c <mount_script> stella-contain <program> <args...>
```

No `--user` / `--map-root-user`. Creating a mount namespace (`CLONE_NEWNS`)
needs `CAP_SYS_ADMIN` in the process's current user namespace. A non-root
process does not have that capability unless it first enters a *user*
namespace (`unshare --user`) that grants it one, or `unshare --mount` is run
as a process that already holds `CAP_SYS_ADMIN` (root, in a namespace/host
that permits it).

## Evidence (partial — see caveat)

Reproduced locally against `rust:1.90` (the pinned toolchain, matching
`rust-toolchain.toml`) inside a plain `docker run` container (Ubuntu 24.04
base, `util-linux` installed, default Docker capability set — no
`--privileged`, no added `SYS_ADMIN`):

- As the container's root user (uid 0): `unshare --mount --propagation
  private -- ...` failed with `cannot change root filesystem propagation:
  Permission denied` (even with `--cap-add=SYS_ADMIN` added to the outer
  `docker run`).
- As a created non-root user: `unshare --mount ...` failed with `unshare
  failed: Operation not permitted`.
- As the non-root user with `unshare --user --map-root-user --mount
  --propagation private -- ...`: also failed with `cannot change root
  filesystem propagation: Permission denied`.

`crates/stella-tools/tests/graded_tree_containment.rs`'s
`STELLA_REQUIRE_CONTAINMENT=1` gate (added for #2998) reproduces the same
outcome through the actual probe: `contain::posture()` reports
`Advisory(ProbeFailed)` on this host, and the new
`.github/workflows/containment-posture.yml` job will show the same result if
GitHub's actual `ubuntu-latest` runner behaves the same way as this
container.

**Caveat, stated because #2998 exists to stop exactly this kind of
under-verified claim**: this is a *nested Docker container*, not a bare VM.
Docker's own capability drop and mount-propagation restrictions on the
container root are a plausible independent explanation for at least the
`--cap-add=SYS_ADMIN` / `--map-root-user` failures — GitHub Actions'
`ubuntu-latest` runners are real VMs, not nested containers, and may not share
this restriction. So this issue is a **flagged hypothesis with partial
evidence**, not a confirmed regression. `containment-posture.yml` (from
#2998) is the mechanism to get the missing half of the evidence: if that job
goes green on an actual GitHub-hosted runner, this issue is moot and should
be closed as not-reproducible outside nested containers; if it stays
`Advisory(ProbeFailed)` there too, the hypothesis above is confirmed and the
fix is almost certainly adding `--user --map-root-user` (or equivalent) to
`wrap_argv`'s `MountNamespace` arm.

## Files

- `crates/stella-tools/src/bash/contain.rs` — `wrap_argv`'s `MountNamespace`
  arm, `mount_script`.
- `.github/workflows/containment-posture.yml` — the job that will supply the
  bare-VM signal this issue is currently missing.

## What "done" looks like

1. Let `containment-posture.yml` run on `main` and on a PR at least once.
2. If it reports `Enforced(MountNamespace)`: close this issue, no code
   change needed — the nested-Docker restriction above was Docker-specific.
3. If it reports `Advisory(ProbeFailed)`: add the capability plumbing
   (`--user --map-root-user`, or the specific flag combination a spike proves
   necessary) to `wrap_argv`'s `MountNamespace` arm, re-run the workflow, and
   confirm it flips to `Enforced`.

Refs #2931, #2998.

macanderson:
## Summary

Closes #2998. Part of the #2931 follow-up chain alongside #2997 (open, out of scope here — this only touches the *test-proves-it* half, not wiring containment to `run_script`/`start_process`/custom tools/hooks).

- `graded_tree_containment.rs`'s `enforced_or_skip()` previously *skipped* its enforcement assertions on any non-`Enforced` host, so the Linux mount-namespace recipe in `contain.rs::mount_script` was proved by no test anywhere — a regression there degrades silently to `Advisory` and CI stays green.
- `STELLA_REQUIRE_CONTAINMENT=1` now turns that skip into a panic, and the panic message names which `Advisory` reason applies: `UnsupportedPlatform` (flag set on the wrong kind of runner — a workflow bug) vs `ProbeFailed` (the mechanism exists here and the functional probe watched it fail to hold — a real recipe regression). Default (unset) behavior is byte-identical to before.
- New `.github/workflows/containment-posture.yml` runs the suite with the flag set, on a real `ubuntu-latest` GitHub Actions runner. **Not** added to `ci.yml`'s required checks.

## Why not just flip it on in the required job

I tested the actual recipe (`unshare --mount --propagation private -- ...`, pinned Rust 1.90 toolchain) inside a plain, non-privileged Docker container standing in for a Linux CI host, and it did **not** hold — neither as root nor as a created non-root user, with or without `--cap-add=SYS_ADMIN` / `--user --map-root-user`. That's evidence the recipe may need capability plumbing it doesn't currently have (`--user`/`--map-root-user`), but a nested Docker container is not the same thing as GitHub's bare-VM `ubuntu-latest` runner, so I can't tell from here whether that failure mode also holds there. Turning the *required* gate red on an untested assumption is exactly the trap `ci.yml`'s own header warns about (#1863/#1892).

Filed #3010 to track that open question, with the exact repro and what "done" looks like once `containment-posture.yml` has actually run on a real GitHub runner. Once it goes green there a few times, promoting it to required is a one-line branch-protection change.

## Test plan

- [x] `cargo test -p stella-tools --test graded_tree_containment -- --nocapture` on macOS (Seatbelt `Enforced`): all 3 pass, unchanged.
- [x] Same test inside a Linux Docker container (`Advisory(ProbeFailed)` host): default run still skips quietly (3 passed); `STELLA_REQUIRE_CONTAINMENT=1` panics on the two enforcement assertions with the expected `ProbeFailed` message, `the_literal_spelling_is_still_refused...` (no posture gate) still passes.
- [x] `cargo fmt --check -p stella-tools`, `cargo clippy -p stella-tools --all-targets -- -D warnings`, `scripts/check-action-pins.sh` all clean.
- [ ] `containment-posture.yml` itself has not yet run on a real GitHub Actions runner — this PR is what triggers its first real run.

## Summary by Sourcery

Tighten graded containment tests so CI can distinguish between enforced, unavailable, and regressed containment postures on Linux runners.

New Features:
- Introduce the STELLA_REQUIRE_CONTAINMENT environment gate to treat unenforced containment on supported CI hosts as a hard failure instead of a skipped test.
- Add a dedicated containment-posture GitHub Actions workflow that runs the graded_tree_containment test suite with containment required on ubuntu-latest runners.

Enhancements:
- Clarify graded_tree_containment test behavior and logging by explicitly returning on enforced posture and standardizing skip messaging when containment is advisory or unavailable.

CI:
- Create a containment-posture workflow that exercises stella-tools graded containment tests under a required containment posture and is triggered on changes to containment-related code and the workflow itself.

---

## Update (2026-08-12): the recipe was broken three ways, and this PR now fixes it

Closes #3010.

The summary above is kept as the record of what was believed when this PR was
opened, and two of its claims no longer describe the diff. It said this PR
"only touches the *test-proves-it* half, not the recipe" — it now changes the
recipe substantially — and it treated the Linux failure as an open question
about capability plumbing. `containment-posture.yml` answered that question by
failing twice, and reading its diagnostic step rather than its verdict turned
up **three defects stacked on each other**. That is exactly the shape #2998
predicted: no test anywhere had ever executed this recipe, so each defect was
hidden behind the one above it.

**1. The wrapper could never exec anything.** `mount_script` ended
`shift; exec "$@"`. `wrap_argv` passes `stella-contain` as the inner shell's
`$0`, and `sh -c <script> <name> <args...>` binds that name to `$0` while
`"$@"` expands to `$1`..`$n` — so the wrapped program was already the first
word of `"$@"` and the `shift` dropped it. The wrapper ran
`exec -c '<command>'` and died with `exec: -c: not found`. No Linux host could
ever have reported `Enforced`.

**2. The `.git` exemption was revoked by its own recipe.** A bind mount clones
the mount flags of the mount its source is found through, so binding `.git`
*after* the read-only remount returned it read-only. The module header calls
that exemption load-bearing precisely because a candidate worktree shares the
graded tree's object store and must be able to `git commit`. Measured on two
kernels — 6.17 (an `ubuntu-latest` runner, under `sudo`) and 6.8 — both
leaving `.git` refusing writes with `Read-only file system` while the ban on
the tree itself held correctly. The doc comment asserting the opposite
("a bind mount does not inherit `MS_RDONLY`") was describing pre-2.6.26
semantics.

**3. Capabilities, not flags.** `--user --map-root-user` (merged here as
#3017) is necessary but not sufficient: the runner's job user holds
`CapEff:0000000000000000` on a kernel with
`apparmor_restrict_unprivileged_userns=1`, so the `/proc/self/uid_map` write
is refused outright. Reproduced independently on a second kernel with seccomp
*and* AppArmor both unconfined, which rules out the container profile as the
cause. The workflow now lifts that sysctl and asserts the lift took, so the
job proves the recipe on a host that permits the namespace.

### What the green check does and does not say

It says: the recipe holds on a real Linux kernel — the ban refuses the write,
and the `.git` exemption survives it. The probe table shows
`sentinel=original exempt=present` where it previously showed `exempt=absent`.

It does **not** say an unprivileged Stella is contained on a stock Ubuntu
24.04 host. The job has to set
`apparmor_restrict_unprivileged_userns: 1 -> 0` first; without that a plain
user still gets `Advisory`. Tracked in #3037, along with the measured
consequence that `--map-root-user` makes the wrapped command see `inner uid=0`
— a possible benchmark-measurement artifact that deserves a deliberate answer
rather than inheritance.

### Tests (for the deleted-test guard)

**No test was deleted or renamed.**
`the_mount_script_remounts_read_only_before_binding_git_back` survives, because
the fix keeps the read-only remount where it was and gives the exemption an
explicit `remount,bind,rw` of its own rather than relying on inheritance. Two
tests are added, one per defect that had no witness:

- `the_inner_shell_execs_the_wrapped_command_itself` — the `$0` / `exec "$@"`
  contract that the `shift` broke.
- `the_git_exemption_is_remounted_writable_after_its_bind` — the exemption's
  own remount, so a future edit that drops it fails here rather than in a
  benchmark run.


macanderson:
The `graded-tree write ban is Enforced on this runner` job is green, and getting it there turned up two real defects in the recipe rather than one CI problem. Full evidence and the reasoning is on #3037 (comment above); the short version:

**1. The `.git` exemption was dead on Linux.** A bind taken *through* a read-only mount clones its `MS_RDONLY`, contradicting `mount_script`'s comment that the `.git` bind "comes back writable". The first diagnostic run showed it directly — all three mounts succeeding and `exempt=absent`. A candidate is a `git worktree` sharing that object store, so this was every `git commit` the model runs in its own workspace failing under containment. Fixed with a `remount,bind,rw`.

**2. The contained shell ate its own command.** The tail was `shift` then `exec "$@"`, but `sh -c <script> <name> <argv...>` already binds `<name>` to `$0` — so the `shift` dropped the program and `exec` ran its first argument. The Linux mount-namespace path had therefore **never reached the command it wraps**: mounts applied, exec failed 127. Witnessed by `the_inner_shell_execs_the_wrapped_command_itself`, which *runs* the handoff rather than asserting about the script text — the string assertion beside it could not have caught this, because the string was exactly what its author meant to write.

**3. The runner refused the namespace at all** — `apparmor_restrict_unprivileged_userns=1`, Ubuntu 24.04's default, denying the `uid_map` write so #3010's `--map-root-user` never acquires `CAP_SYS_ADMIN`. The job now lifts that knob and asserts it took.

Both witnesses were checked the artisanal way — reverting the generator alone and re-running — and fail on the old code.

Two things I deliberately did **not** decide, both flagged on #3037:

- **The uid question.** `--map-root-user` makes the wrapped command see itself as uid 0, and this wrapper is armed exactly during graded bench runs, where tooling that branches on "am I root?" behaves differently. The job now *measures* it (each probe line reports the contained shell's `id -u`); choosing between `--map-root-user`, `--map-current-user`, and documenting the change is a maintainer call with a bench-honesty cost attached.
- **Job structure.** #3037 proposes two jobs — one pinning the privileged path, one the unprivileged. This is the second of those two; the privileged path is exercised by the `sudo, mount only` diagnostic row but asserted nowhere. Splitting is cheap, but which contexts become required is a branch-protection decision, and this workflow's header is explicit about promoting too early.

What the green job proves is narrow and stated in its comment: the recipe holds on a real kernel where unprivileged user namespaces are permitted. It does not claim Stella is `Enforced` on a stock GitHub runner — there `posture()` still correctly returns `Advisory(ProbeFailed)`.

macanderson:
Independent confirmation on a second architecture, since `containment-posture`
only ever runs x86_64. Ran `mount_script`'s exact emitted text — bind, remount
ro, bind `.git`, `remount,bind,rw` the `.git`, `exec "$@"` — behind the shipped
`wrap_argv` argv, in a container on **aarch64, kernel 6.8.0-117**:

```
--- privileged root, --mount only
    wrapper rc=0 sentinel=original exempt=present   => Enforced(MountNamespace)
--- root, --user --map-root-user --mount (the shipped argv)
    wrapper rc=0 sentinel=original exempt=present   => Enforced(MountNamespace)
```

Same tree, same script, with the `.git` bind moved back before the read-only
remount instead of remounted writable, also reaches `exempt=present` — so the
two candidate fixes for that defect agree, and the one shipped here is not
resting on a quirk of one kernel or one arch. The pre-fix order reproduces the
bug on this kernel too (`exempt=absent`), which is what makes the new
`the_git_exemption_is_remounted_writable_after_its_bind` a real witness rather
than a restatement.

macanderson:
Residue from this work, filed as a handoff rather than left in the diff: #3040 — a `ProbeFailed` cannot say which link refused, because `run_probe` discards the wrapper's stderr and `probe_mechanism` `&&`-s three separate facts into one bool. That is why finding the two defects here took two round trips through CI, and why the hand-run `why` step is currently the only way to learn a reason. The two decisions I deliberately left open are on #3037 (the uid question, now measured at `inner uid=0`, and the one-job-vs-two structure).

PR #3017, opened 2026-08-12
fix(stella-tools): give the mount-namespace recipe CAP_SYS_ADMIN it doesn't have

macanderson:
## Summary

Refs #3010, #2931, #2998.

PR #3011's new `containment-posture.yml` supplied the missing evidence #3010 was waiting on: on a real GitHub Actions `ubuntu-latest` runner, `unshare --mount --propagation private` reports `ProbeFailed`, the same permission-denied failure the issue reproduced under Docker — not a Docker-specific artifact.

`unshare --mount --propagation private` needs to change the propagation type of an inherited mount (the root filesystem). That requires `CAP_SYS_ADMIN` in the user namespace that *owns* that mount — and an unprivileged CI runner process has no such capability in the initial namespace it was born into. `--user --map-root-user` creates a fresh user namespace and maps the caller to uid 0 inside it, giving it root — and therefore `CAP_SYS_ADMIN` — in a namespace of its own before `--mount` runs. This is the standard recipe rootless container tools (podman/buildah/bubblewrap) use for the same reason.

- `wrap_argv`'s `MountNamespace` arm now prepends `--user --map-root-user` to the `unshare` invocation.
- Module doc comment updated to state and justify the recipe.
- No change to `mount_script` itself — once inside the new user+mount namespace pair, the process already holds the capability the bind/remount sequence needs.

## Test plan

- [x] `cargo test -p stella-tools --lib bash::contain` — 6/6 pass (pure `wrap_argv`/`mount_script` unit tests, platform-independent).
- [x] `cargo fmt -p stella-tools --check`, `cargo clippy -p stella-tools --all-targets -- -D warnings` — clean.
- [ ] `containment-posture.yml` (non-required, informational) on this PR is the actual verification: if it reports `Enforced(MountNamespace)`, #3010 is resolved; if it still reports `Advisory(ProbeFailed)` (e.g. Ubuntu 24.04 AppArmor's `unprivileged_userns_restrict` blocking unconfined user-namespace creation), the recipe needs more plumbing and this PR should not merge as-is — I don't have a Linux host to verify the kernel behavior directly, so this run is the evidence, not a guess dressed as one.

## Summary by Sourcery

Adjust Linux containment to use a user+mount namespace pair so mount-namespace probing works on unprivileged CI runners.

Enhancements:
- Update mount-namespace invocation to include user namespace creation with mapped root user before establishing the private mount namespace.
- Expand module documentation to explain the dependency on CAP_SYS_ADMIN and reference the related issue.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3046,7 +3046,7 @@ checksum = "a2eb9349b6444b326872e140eb1cf5e7c522154d69e7a0ffb0fb81c06b37543f"
 
 [[package]]
 name = "stella-cli"
-version = "0.9.12"
+version = "0.9.13"
 dependencies = [
  "async-trait",
  "clap",
@@ -3094,7 +3094,7 @@ dependencies = [
 
 [[package]]
 name = "stella-context"
-version = "0.9.12"
+version = "0.9.13"
 dependencies = [
  "async-trait",
  "contextgraph-types",
@@ -3111,7 +3111,7 @@ dependencies = [
 
 [[package]]
 name = "stella-core"
-version = "0.9.12"
+version = "0.9.13"
 dependencies = [
  "async-trait",
  "futures-util",
@@ -3129,7 +3129,7 @@ dependencies = [
 
 [[package]]
 name = "stella-diag"
-version = "0.9.12"
+version = "0.9.13"
 dependencies = [
  "proptest",
  "serde",
@@ -3140,11 +3140,11 @@ dependencies = [
 
 [[package]]
 name = "stella-diff"
-version = "0.9.12"
+version = "0.9.13"
 
 [[package]]
 name = "stella-embed"
-version = "0.9.12"
+version = "0.9.13"
 dependencies = [
  "async-trait",
  "proptest",
@@ -3158,7 +3158,7 @@ dependencies = [
 
 [[package]]
 name = "stella-engine"
-version = "0.9.12"
+version = "0.9.13"
 dependencies = [
  "async-trait",
  "serde",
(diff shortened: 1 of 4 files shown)


macanderson:
Superseded by #3011, which now carries `--user --map-root-user` along with the
two defects that had to be fixed before it could do anything.

This PR was opened on the theory #3010 proposed: that the Linux recipe only
lacked capability plumbing. That theory was incomplete. Adding the flags alone
does not make the check pass, because two further defects sat underneath it —
both since fixed on #3011's branch:

- `mount_script` ended `shift; exec "$@"`, but `wrap_argv` passes
  `stella-contain` as the inner shell's `$0`, and `"$@"` is `$1`..`$n`. The
  program was already the first word of `"$@"`, so the `shift` dropped it and
  the wrapper ran `exec -c '<command>'`. No Linux host could ever have reported
  `Enforced`.
- The `.git` exemption was revoked by its own recipe: a bind mount clones the
  mount flags of the mount its source is found through, so binding `.git` after
  the read-only remount returned it read-only. Measured on two kernels — 6.17
  (an `ubuntu-latest` runner under `sudo`) and 6.8 — both leaving `.git`
  refusing writes with `Read-only file system` while the ban on the tree itself
  held correctly.

`containment-posture` is green on #3011 as of run 31641319438, with the
probe table showing `sentinel=original exempt=present` where it previously
showed `exempt=absent`. Closing this in favour of that branch; the remaining
gap (an unprivileged caller on a *stock* Ubuntu host, where the sysctl is not
lifted) is tracked in #3037.

macanderson:
**Correction to the comment above:** it says "closing this in favour of that
branch." That was wrong — this PR was already merged into
`worktree-fix-2998-containment-ci-assertion` at 2026-08-12T20:42Z as
`07ea7b4a0`, before that comment was written. Nothing here was abandoned; the
flags shipped, and the two defects described above were then fixed on top of
them. Everything else in that comment stands, including the measurements.

Issue #3025, opened 2026-08-12
arenabench: a default `cloud fetch` cannot be assembled or read by the trace tooling, and only says so after the download


macanderson:
## Problem

`arenabench cloud fetch <run>` downloads only each trial's `results.json`
unless `--artifacts` is passed. `arenabench assemble` (called at the end of
fetch) needs the per-trial artifact trees — `<job>/ws/matches/*/jobs/*/<task>__*`
— to link into the match, so a default fetch assembles **nothing** and prints:

```
assembled : 0 trial(s) into one match (…)
  !! no trial artifacts under … — a results-only fetch carries no traces to
     assemble. Re-fetch with `--artifacts`.
```

That is honest, and it is still one wasted round trip: the operator learns
after the download that the download was not the one they wanted.

## The tension

`--artifacts` is opt-in on purpose (`cloud.py::fetch_results`): the full tree is
"events, recordings, workspaces" and can be gigabytes. But every rule this
repository holds itself to says a bench conclusion comes from the trace, and
`bench/trace_triage` — both the issue-filing path and the new postmortem — can
read *nothing at all* from a results-only fetch. In practice a measured run is
always fetched with `--artifacts`; the default optimises for the case nobody
uses.

## Options

1. Default `--artifacts` **on** for a fetch, with `--results-only` as the opt
   out. Matches how the runs in `arenabench-cloud/` were actually fetched.
2. Keep the default and refuse earlier: have `cloud fetch` say, **before**
   downloading, that this fetch will not be assemblable or readable by the
   trace tooling.
3. Fetch results first and the artifact tree lazily, on the first thing that
   needs it.

## Files

- `arenabench/arenabench/cloud.py` — `CloudExecutor.fetch_results`,
  `_fetch_prefix`, `_cmd_cloud_fetch`, `_cmd_cloud_run`, `register_cli`
- `arenabench/arenabench/assemble.py` — `discover_trials`, `assemble_run`
- `arenabench/tests/test_assemble.py::test_a_results_only_fetch_says_so_instead_of_assembling_nothing`
  pins today's behaviour and must be updated with whichever option lands.

## Verify

```sh
arenabench cloud fetch <run-id> --out /tmp/r && ls /tmp/r/matches/*/jobs/*/
```

## Done when

A plain `arenabench cloud fetch` either produces an assemblable match or says
so before spending the download, and the size tradeoff is stated in
`cloud.py`'s module docstring rather than only in a flag's help text.

[...]

Issue #3024, opened 2026-08-12
bench: the trace postmortem is not run automatically at fetch time, so it still depends on somebody remembering


macanderson:
## Problem

`bench/trace_triage/postmortem.py` (added alongside `arenabench assemble`)
renders a run's trace probe onto the match assembled from it — but **nothing
runs it automatically**. `arenabench cloud fetch` assembles the match and then
prints nothing about the postmortem; an operator has to remember
`make postmortem RUN=<dir>`. The stated purpose of the report is "so we have
the probe into traces done in automation if we forget to read ourselves", and
a step a human must remember is exactly the thing it was built to replace.

## Why it is not already wired

Layering, deliberately, and the tradeoff should be re-decided rather than
inherited:

- `arenabench/` is an installable, agent-agnostic package (it seats
  `claude-code` and `stella` alike) and is being prepared for ejection into its
  own repository (`.github/workflows/bench.yml` says so).
- `bench/trace_triage/` is a set of flat stdlib modules with a `sys.path`
  `conftest.py`, not importable as a package, and every one of its detectors
  cites a **Stella crate path** (`crates/stella-pipeline/src/verify.rs`,
  `crates/stella-core/src/driver.rs`, …). It is Stella-specific by construction.

So `arenabench` importing `trace_triage` would both break the ejection and put
Stella-specific analysis inside an agent-agnostic package. The fold and the
probe therefore meet through a **file**, not an import.

## Options

1. **A repo-level wrapper** — a `scripts/` shim or a Makefile target that runs
   `arenabench cloud fetch` and then `postmortem.py`, so the two-step is one
   command for a Stella operator without arenabench knowing trace_triage
   exists. Cheapest, keeps every boundary.
2. **`arenabench cloud fetch --postmortem CMD`** — a declared hook the operator
   or a wrapper supplies. Keeps arenabench agnostic; costs a flag.
3. **Make `trace_triage` an installable package** and let a Stella-only extra
   depend on it. Largest change; revisit only if the probe grows other callers.

## Files

- `bench/trace_triage/postmortem.py` — `main`, `write_report`, `report_paths`
- `arenabench/arenabench/assemble.py` — `assemble_run`, `register_cli`
- `arenabench/arenabench/cloud.py` — `_assemble_fetched`, `_cmd_cloud_fetch`,
  `_cmd_cloud_run`
- `Makefile` — the `postmortem` target

## Verify

```sh
[...]

Issue #3023, opened 2026-08-12
bench: the survey's "empty ok-results" column does not replay from the traces, so it ships as no band at all


macanderson:
## Problem

`bench/trace_triage/bands.py` records a nine-arm survey (2026-08-12) of what a
healthy bench arm looks like. Three of its four per-trial columns replay
exactly from the traces; the fourth does not, so it ships as an observation on
`bands.TrialMetrics.empty_ok_results` and as **no band at all** — there is no
threshold, and the postmortem's band table has three rows rather than four.

Replaying `arenabench-cloud/s5b2` (the arm the survey scores at **2.4 empty
ok-results per trial**) with `run_trace.load_run` gives:

| candidate definition | s5b2, per trial |
|---|---|
| `output.ok.content` is empty or whitespace (what ships today) | 0.0 |
| `output.ok` carries no `content` key at all | 0.0 |
| `output.ok.content` under 20 characters | 1.3 |

and the sub-20-character strings are real content — `(no matches)` (11
occurrences), a bare `\n[exit code: 0]` trailer (8), `OK\n\n[exit code: 0]` (4).
None of the three reaches 2.4.

The same replay reproduces the other three columns for s5b2 **exactly** —
tools 32.5, identical consecutive calls 0.0, cache hit 100.0% — so the
disagreement is this metric's definition, not the loader and not the fixture.

## Why it matters

A threshold whose measurement cannot be reproduced fires or stays silent for
reasons nobody can check. The survey's unhealthy arms sit at 3.3-9.2 on this
column and one healthy-cache arm sits at 3.5, so it was never going to be a
verdict-capable separator — but as a *directional* row it is still one of the
four things the postmortem reports, and a row whose observed value is 0.0 on an
arm the survey calls 2.4 is worse than a missing row.

## Files

- `bench/trace_triage/bands.py` — `EMPTY_OK_UNRESOLVED`, `BANDS`,
  `TrialMetrics.empty_ok_results`, `trial_metrics`
- `bench/trace_triage/tests/test_postmortem.py::test_a_band_that_cannot_separate_says_so`
  asserts `empty_ok_results` is **not** in `BANDS`; it flips when this lands.

## Verify

```sh
python3 bench/trace_triage/postmortem.py arenabench-cloud/s5b2 --stdout
```

The band table must show a row whose observed value matches the survey's 2.4
for s5b2, and the same definition must place w10f near 9.2 and dec1 near 1.5.

## Done when

[...]

PR #3026, opened 2026-08-12
feat(arenabench): a cloud run reassembles into ONE match, and that match carries a postmortem of its traces

macanderson:
Two coupled changes: **a cloud run reassembles into ONE match**, and **that match carries a postmortem** — an automated read of its traces.

Run `s5b2` (10 tasks x 2 attempts, one seat) fetched down as 21 `match.toml` and 40 `results.json`. At the 89-task two-arm scale that is 178 matches in a local arena for one experiment.

## 1. Reassembly — `arenabench/arenabench/assemble.py`

`cloud run` submits one Batch job per (attempt, task, seat), each executing a one-trial slice. **The assembled match is not invented: it is the submitted one.** `runs/<id>/match.toml` in S3 holds the whole spec and the per-trial ones are slices of it (`cloud.slice_spec`), so assembly restores the spec that was always there and files each trial back into the cell it was sliced out of.

**Where it happens, and why.** Not inside the S3 adapter: every input is a file the fetch already wrote, so the fold is pure and local. Three things follow, and each is the argument for keeping it separate rather than a consequence of having done so — it is **testable with no network** (the tests build a fetched tree in `tmp_path`), it is **re-runnable on a run fetched months ago**, and it is **re-runnable after a bug in it is fixed** without paying for the download again. `cloud fetch` and `cloud run` call it at the end, so the operator still types one command, and `arenabench assemble <dir>` is the same fold on any directory.

**Per-trial artifacts stay reachable.** Nothing is copied and nothing is moved: the match is **relative symlinks** into the per-trial trees the fetch wrote, so `agent/stella-events.jsonl`, `result.json`, `verifier/` and the rest are reachable both at their original path and through the match. `assembly.json` records which Batch job and which cell every link came from, so a cell's provenance is a lookup rather than an inference from a directory name. Relative, so the tree survives being copied to another machine.

**It opens as one contest with no new server code.** The output is the local `<workspace>/matches/<id>/` layout, so `MatchRunner.restore_from_disk` — what `arenabench serve` runs — finds it, and `Match.trial_dirs` collapses a task's attempts to one representative by the same measured-beats-void rule (`_representative_rank`) a local run uses. Multi-seat runs land as columns because the seats are the spec's contestants. The match id is the **run directory name**, not the spec id: two runs of one template share a spec id, and two matches sharing a directory name is one silently shadowing the other.

**Existing runs: not migrated, not corrupted.** Assembly only *adds* a `matches/` subdirectory and only ever deletes symlinks it created. Every run already in `arenabench-cloud/` is assembled by running `arenabench assemble` on it; nothing is rewritten.

## 2. The postmortem — `bench/trace_triage/postmortem.py`

`make triage-bench-traces` files GitHub issues. This files nothing and says, on screen, where a run went wrong.

**Detectors are reused, not re-implemented.** `build_report` calls `detectors.run_all` over a `run_trace.Run` — the same call the triage path makes, holding no detector list of its own. The two therefore *cannot* disagree about whether a run is healthy, and a test asserts identity of the two result lists rather than agreement.

**Two new detectors join the shared registry** for shapes the existing eight cannot see. `cache-collapse`: pooled cache hit under the 93% floor — on the cold-cache fixture the pre-existing eight produce nothing while this fires, which is the whole point (the broken-roster arm sat at 52% and looked well-formed). `repeated-file-read`: one path named by five or more read-shaped calls in a single trial (the worst observed read one file 13 times) — distinct from `repeated-identical-tool-call`, which is strictly adjacent and byte-identical.

**Bands — `bands.py`.** The measured healthy profile, with the nine-arm survey recorded beside it. `Band.verdict_capable` marks the two that separate cleanly (identical consecutive calls, cache hit) from the one that overlaps (tool count: 32.7 on an arm that solved 50/88, 32.5 on one that solved 18/20) and is reported as directional and never concluded from.

**"The run is broken" vs "the agent did badly"** is the first section and is stated in trials, never percentages: a trial with zero `step_usage` completed no model call and measured nothing, its reward file of zeros is the absence of a measurement, and scoring it as a loss hands the other arm a win it did not earn. Void trials are also excluded from every band mean.

**Healthy runs get one clean paragraph** — no band table, no findings section, no manufactured concern. That is asserted, not hoped for.

### One measurement I could not reproduce, reported against myself

Three of the survey's four columns replay **exactly** from the traces (s5b2: 32.5 tool calls, 0.0 identical consecutive, 100.0% cache). The fourth — "empty ok-results", 2.4 for s5b2 — does not: the traces give 0.0 under "content empty or whitespace", 0.0 under "no content key", and 1.3 under "under twenty characters", where the strings are real content (`(no matches)`, a bare `[exit code: 0]` trailer). Rather than pick whichever definition flatters the table, it ships as an observation on `TrialMetrics` and as **no band at all**. Filed as #3023 with the replay numbers.

## Evidence

- **The reassembly witness discriminates**: on the fixture tree, `MatchRunner.restore_from_disk` returns **0** matches before `assemble_run` and **1** after, with 3 rows x 2 cells.
- **The new detectors discriminate**: on the cold-cache fixture the pre-existing eight report only `no-post-verdict-file-change`; `cache-collapse` is the only new signal.
- **It runs on real traces**: `postmortem.py arenabench-cloud/s5b2 --stdout` produces the band table above (32.5 / 0.0 / 100%), the 18/20 graded cohort with the small-sample caveat, and two findings with per-trial S3 citations.
- 16 tests in `arenabench/tests/test_assemble.py`, 14 in `bench/trace_triage/tests/test_postmortem.py`; the whole arenabench suite (exit 0) and all 59 trace_triage tests pass.

**Deleted tests: none.**

## Issues filed

- #3023 — the survey's empty-ok column does not replay; it ships as no band.
- #3024 — the postmortem is not run automatically at fetch time (the arenabench/trace_triage boundary, with three options and the ejection argument).
- #3025 — a default `cloud fetch` cannot be assembled or read by the trace tooling, and only says so after the download.

Refs #3023, #3024, #3025

## Summary by Sourcery

Assemble fetched cloud runs into a single local match that ArenaBench can serve, and add a trace postmortem tool that analyzes runs using shared detectors and measured health bands.

New Features:
- Introduce an `arenabench assemble` CLI verb and automatic assembly in cloud fetch/run to fold per-trial artifacts into one match directory with provenance and manifest metadata.
- Add a postmortem CLI and Makefile target that generate markdown/JSON reports from bench run traces using existing detectors and new band-based metrics.

Enhancements:
- Define reusable trace metric bands and aggregation utilities to quantify distance from healthy runs and support new detectors.
- Ensure assembled matches use relative symlinks, are idempotent to re-run, and integrate with existing ArenaBench match serving and restoration paths.

Tests:
- Add assemble tests that construct a synthetic fetched cloud run, verify all trials are preserved and linked, and confirm the assembled match opens through `MatchRunner`.
- Add postmortem tests that validate quiet output on healthy fixtures, correct handling of void/timeout trials, band behavior on small samples, and alignment with the detector registry.


macanderson:
### The band metrics replay the survey exactly, on two independent real arms

Not asserted — run. `postmortem.py <run> --stdout` against the fetched artifacts in `arenabench-cloud/`:

| arm | metric | survey | replayed from the traces |
|---|---|---|---|
| s5b2 | tool calls / trial | 32.5 | **32.5** |
| s5b2 | identical consecutive | 0.0 | **0.0** |
| s5b2 | cache hit | 100% | **100.0%** |
| s5b2 | solved | 18/20 | **18/20** |
| dec1 | tool calls / trial | 27.1 | **27.1** |
| dec1 | identical consecutive | 0.1 | **0.1** |
| dec1 | cache hit | 100% | **100.0%** |
| dec1 | solved | 17/20 | **17/20** |

That is what makes the fourth column's failure to replay (#3023) a statement about *that metric's definition* rather than about the loader, the fixture, or the arithmetic — the same code path reproduces every other number on both arms.

dec1 also exercises the timeout cohort on real data: one trial (`password-recovery__vKiEbsp`) is reported as having run out of clock rather than as a wrong answer, read off the **job-level** `result.json`, which is the file the trial-root `results.json` does not carry.

### The two witnesses, both observed failing first

- **Reassembly.** On the fixture tree, `MatchRunner.restore_from_disk` — the call `arenabench serve` makes — returns `0` matches before `assemble_run` and `1` after, with 3 rows x 2 cells. Not a shape assertion on a JSON file: the real server path.
- **The new detectors.** On the cold-cache fixture, the pre-existing eight detectors report only `no-post-verdict-file-change`; `cache-collapse` is the only new signal. An arm whose prompt cache stopped engaging was invisible to every detector that existed before this PR.

Issue #2956, opened 2026-08-12
bench: glm-5.2 on z.ai reaches ~7 of ~28 steps inside the 900s agent timeout — 57% on OpenRouter vs 14% on z.ai, same model and effort


macanderson:
## Summary

The same model at the same effort inside the same agent timeout scores **57% on
OpenRouter and 14% on z.ai** — not because the agent is worse, but because on
z.ai it never gets far enough through the task to finish. A benchmark run was
launched, invalidated, and cancelled on this. Nothing in the config surface
warns about it.

## Evidence

Two ArenaBench runs, `glm-5.2` at `effort = "high"`, `agent_timeout_multiplier = 1.0`
(a 900s agent timeout), Terminal-Bench 2.1, 89 tasks:

| run | gateway | base_url | solve rate |
|---|---|---|---|
| `f89b2` | OpenRouter | `https://openrouter.ai/api/v1` | **50/88 = 57%** |
| `h89b` | z.ai | `https://api.z.ai/api/coding/paas/v4` | **12/85 = 14%** |

`85 of 88` completed `h89b` trials ended in
`harbor.trial.trial.AgentTimeoutError: Agent execution timed out after 900.0 seconds`.

The traces say why. Completed model steps per trial, sampled from `h89b`:

```
task                          reasoning deltas   step_usage   tool_start
adaptive-rejection-sampler              36,039            0            0
sparql-university                       25,618            8            8
extract-elf                             18,562            7           14
configure-git-webserver                  6,828           11           12
dna-insert                                 166            7            7
build-cython-ext                           117            2            4
```

For comparison, the bare loop averages **28.3 completed steps per trial** on
OpenRouter (measured over 20 trials in run `w10b`). On z.ai it reaches 1-11
before the kill — the agent is being cut off around step 7 of a ~28-step task.

The `adaptive-rejection-sampler` trial is the extreme of the same shape: 36,039
`reasoning` deltas, **zero** `step_usage`, zero tool calls, 15.3 minutes of wall
clock. The model streamed reasoning for the entire window without ever
completing a step.

## What this is NOT

- **Not throttling.** That was the first hypothesis and the traces killed it: a
  rate limit would show as few events overall or explicit retry/429 handling.
  What is actually there is a very high reasoning-token volume per completed
  step. Do not "fix" this by lowering concurrency.
- **Not a config difference.** `agent_timeout_multiplier` is `1.0` in both runs;
  the value was copied from the run that scored 57%.
[...]

PR #3044, opened 2026-08-12
feat(arenabench): a banned-behavior gate that stops a bench run on a regression we already fixed

macanderson:
A live check that aborts an ArenaBench run the moment a finished trial exhibits
a defect we have already fixed. Today a regression burns 20–178 trials and hours
of fleet time before anyone reads a trace; yesterday an expired credential
produced 20 dead trials and a gateway serving 25k reasoning deltas per step
burned 85 of 88 on the agent timeout. Both were visible in the first trial.

It is the **third consumer of the one detector registry**, not a second suite:
`triage_bench_traces.py` files issues from `detectors.run_all`, `postmortem.py`
(#3026) renders a report from it, and `banned.py` decides whether the run keeps
spending. Three suites that can disagree about whether a run is healthy means
none is trustworthy.

## Where it hooks, and why

On the **submitter's watch loop** — `CloudExecutor.watch` gains an `on_settled`
hook, called once per job the first time it reaches `SUCCEEDED`/`FAILED`.

Not in the container: one Batch job is one trial, so a trial that detects its own
banned behavior can only stop itself, and it has already finished. The cost worth
saving is the other N−1 trials, and the only process that can cancel those is the
one holding `jobs.json`. An in-container gate would also have to ship a
Stella-specific detector ledger into an agent-agnostic runner image.

Not a separate watcher: `cloud run` is already a live process polling every 15s
for the run's duration. A second process is a thing to remember to start, and
remembering is the failure mode this exists to remove.

**The seam is a subprocess, not an import.** `arenabench/` seats `claude-code`
and `stella` alike, is stdlib-only by contract, and is being prepared for
ejection; every detector here cites a Stella crate path. #3024 argues that
boundary — this takes its option 2. Contract: `CMD <trial-dir> [--allow CODE]...`,
exit `0` clear, `9` banned, **anything else means the gate itself failed and the
run continues**. A checker that crashed has said nothing about the trial, and a
gate that kills healthy runs is switched off by the second operator it
inconveniences.

## What abort does

`cancel_run` reads `s3://<bucket>/runs/<run-id>/jobs.json`, takes those ids and
signals exactly those — `terminate_job` for running, `cancel_job` for queued.
**Never a queue sweep, never a job-name prefix**: `arenabench-measure` and
`arenabench-burst` are shared by every run and operator in the account. This repo
has paid for the general lesson twice — `scripts/arena-kill.sh` killed a
neighbour's server by argv pattern on 2026-08-08, and `reap.py:100-103` reaps by
trial id for the same reason. Selection is a pure function tested with dicts; one
failed call never abandons the rest, because a partial cancel is the worst
outcome available.

It writes `abort.json` beside the fetched results: run, codes, the trial that
tripped it, the gate's **full** report, and exactly which ids were cancelled. That
record is the thing a disputed abort is argued with.

## The trigger rule

Per-behavior, and the distinction is **not how bad it is** — it is whether the
signature is a property of the *run* or of the *trial*.

- A fact about the configuration (an absent fix, a misrouted model, a dead
  credential) cannot be unlucky: the next trial runs the same binary against the
  same gateway. One trial is proof → `FIRST_TRIAL`.
- A *distribution* can be unlucky. `agent-timeout` fires on 1 of 20 trials of the
  healthy `post1` run, so a first-trial trip would abort healthy runs → `Rate(3, 5)`.

A rate row **cannot trip before its denominator exists**: one timeout in one
settled trial is not a 100% timeout rate, and the first job to finish is very
often the one that failed fastest.

## Banned behaviors shipped, with the evidence

| behavior | warrant | trip | verified against |
|---|---|---|---|
| `grep-ere-false-negative` | fixed-regression #2989 | first trial | **s5b2** (SUT `62a0cb5048f0`, which `git merge-base --is-ancestor b58e29a6` says does *not* contain the fix): 22 occurrences, 0 disclosures. **post1** (SUT `018852a97387`, which does): 0 occurrences, 4 of 4 empties disclosed. |
| `void-model-call` | measurement-invariant | first trial | silent on both real runs — 20/20 trials completed ≥1 model call in each |
| `role-model-census-mismatch` | measurement-invariant | first trial | **silence verified, firing fixture-only** — see below |
| `agent-timeout` | fixed-regression #2956 | 3 of first 5 | fires 1/20 on healthy post1, which is the argument for the rate |

Seven further detectors are `ReportOnly` with a stated reason —
`tool-error-envelope` fires on 40% of the healthy post1 run and 60% of s5b2, so
aborting on it aborts everything.

### A candidate that dissolved when checked

`role-model-census-mismatch` fired on **20 of 20 trials of s5b2** and I nearly
shipped that as its evidence. It was a bug in the detector: `tail()` only
stripped a provider prefix from names with more than two segments, so a declared
`anthropic/claude-sonnet-5` compared unequal to an observed `claude-sonnet-5` —
the same model, and `step_usage` carries the provider in its own field. Replaced
with a segment-suffix comparison (which still reports a genuinely *different*
provider). The row ships with `FIRST_TRIAL` because the invariant is sound and
one trial decides it, but its `evidence` string says plainly that only the
silence half is verified on real traces.

### What is not shipped

Four candidates from the brief — #2915, #2943, #2864, #2931 — are `ReportOnly`
citing **#3053**, because there is no trace in this repository to verify them
against: the only mirrors carrying `agent/` artifacts are three bare-loop arms
with **zero** `proof` and **zero** `verdict` events across 60 trials. Shipping an
abort rule unverified is the false abort this gate must not have.

## How a new entry gets added

The ledger is **total over `detectors.DETECTORS`** —
`test_ledger_is_total_over_the_detector_registry` fails if a detector exists with
no row, in the shape AGENTS.md invariant #10 uses for event consumers. You cannot
add a detector without answering "does this stop a run?", because the test asks.
A `fixed-regression` warrant with no citation is rejected by the `Warrant`
constructor, so an entry with no fix behind it cannot exist.

Honest limit, stated in the module: this forces the decision when a *detector* is
added, not when a *bug* is fixed.

## False positives

`--gate-allow CODE` waives one behavior for one run and the waiver rides into the
verdict, so a run that muted a check can never later be read as clean. `--no-gate`
disables the gate entirely. Both are deliberate acts at the command line and are
printed in the run header — **not yet persisted to `jobs.json`, which is tracked
as #3059** and is the wrong direction. Then you fix the entry; you do not switch
the gate off, and `abort.json` is what you fix it from.

And the mechanism that stops it depending on memory: **`cloud run` refuses to
start with no gate armed** (exit 2, naming `--gate`, `$ARENABENCH_GATE` and
`--no-gate`). Defaulting to unguarded means the one run that needed the gate is
the run that did not have it.

## Tests

64 new, all hermetic — no S3, no GitHub, no Batch, no network.

- `bench/trace_triage/tests/test_banned.py` (23) — every banned behavior from
  **both** sides, the rate rule at 1/5 and 3/5, the totality guard, waivers.
  The load-bearing one is `test_a_run_firing_only_report_only_detectors_is_clear`.
- `bench/trace_triage/tests/test_gate_contract.py` (4) — `banned.py` as a **real
  subprocess**, because arenabench tests its half with a fake checker and this
  side tests in-process, so nothing else crosses the boundary the two agree on.
- `arenabench/tests/test_gate.py` + `tests/test_cancel.py` (37) — plan routing, the
  crashed/timed-out gate not aborting, resolution order, the refusal, and
  `test_a_neighbouring_runs_jobs_are_never_touched`.

Witnessed failing before passing: the bare-vs-qualified model comparison (fails
with the old two-segment `tail`), the grep detector's trip (fails with it removed
from the registry), the totality guard (fails with a row deleted), and the
neighbour test (signals `b-001`/`b-002` when the selector is queue-scoped).

## Deliberately not in `GATE_STEPS`

It talks to S3 and Batch; `gate-parity` would fail. No baseline was raised — all
new logic landed in new modules because `cloud.py` (1385) and `runner.py` (1473)
are near the 1500-line ceiling and not grandfathered.

## Issues filed

- #3053 — four banned-behavior candidates unverifiable against any trace here
- #3056 — `rate_line()` prints "22 of 20 trials (110%)" for per-call detectors
- #3058 — `TestFetchMergesIntoOneTrial` fails on `main` (pre-existing, confirmed
  by control on a detached `origin/main` worktree)
- #3059 — the gate waiver is printed but never recorded

Refs #3024, #3025, #3026, #2989, #2956


macanderson:
This PR's only red check, `harbor_adapter + analyzer pytest`, is **not caused by
this branch** — it fails identically on unmodified `main`.

Control run at `origin/main` (44f62e8e0), nothing from this branch applied:

```
FAILED tests/test_cloud.py::TestFetchMergesIntoOneTrial::test_cloud_fetch_writes_a_merged_trial_alongside_the_flat_files
  FileNotFoundError: .../trial.json
```

Cause: #3026 pointed `_cmd_cloud_fetch` at `_assemble_fetched` and left
`_merge_and_report` (the whole of #3018) with zero callers. Assembly needs
`runs/<id>/match.toml`, which `fetch_results` downloads only under
`--artifacts`, so the default fetch could never assemble and stopped writing
`trial.json` at all. #3018's witness has been failing on `main` ever since,
which reds this job on every open PR.

Fixed in **#3047** — assembly degrades to the merge instead of returning. Once
that merges, a merge of `main` into this branch should clear the red here with
no change to this PR's own code.

Everything this branch adds passes locally at `5c74666`:

- `bench/trace_triage/tests` — 82 passed
- `arenabench/tests/test_gate.py` + `test_cancel.py` — 36 passed

One note unrelated to the red: the PR description is stale. It still reads
"WIP — draft to anchor the branch" and lists the ledger, the gate seam, the
run-scoped cancel and the tests as "still to land", but all four are on the
branch now (`12b7acda`, `abf5ee31`, `55fe89ed`, `788b22a5`). I have left the
body and the branch alone rather than editing underneath an active session.

Issue #2985, opened 2026-08-12
prompt.rs is 14 lines from the 1500-line ceiling — split the persona literals out before the next contract lands


macanderson:
## Problem

`crates/stella-cli/src/agent/prompt.rs` is **1486 lines** after #2984 (PR for the
delta-orientation rule). The `file-size` gate's ceiling for a non-grandfathered
file is 1500, and `scripts/file-size-baseline.txt` carries no entry for it — by
design, and AGENTS.md forbids adding one to go green.

So the next shared prompt contract has ~14 lines of room. That is not enough for
the established shape: a contract is a `macro_rules!` literal (2-4 lines) plus a
provenance doc comment, and every existing one runs 10-25 lines of comment. #2984
already had to route its provenance prose into `prompt/parity.rs` to stay under
the line, and `prompt/parity.rs`'s own module doc already records the same
pressure ("that file sits within a few lines of the 1500-line ratchet"). The
constraint is now shaping where prose lands rather than where logic lands, which
is the signal that the split is overdue.

## What to do

Split the persona text out of `prompt.rs`, leaving assembly behind. The natural
seam, given the file's own structure:

- `crates/stella-cli/src/agent/prompt/contracts.rs` — the nine `macro_rules!`
  shared contracts (`tool_steering!` … `injection_defense!`) with their
  provenance doc comments.
- `crates/stella-cli/src/agent/prompt.rs` — keeps `SYSTEM_PROMPT`,
  `PIPELINE_SYSTEM_PROMPT`, `assemble_system_prompt` and the workspace-context
  appenders.

## Constraints discovered

- **`macro_rules!` scope is textual.** `prompt.rs` declares `mod parity;` *last*
  for exactly this reason (see the comment above the declaration). A contracts
  module must be declared **before** the `concat!` sites that invoke its macros,
  and the macros need `#[macro_use]` or `macro_rules!`-to-`pub(crate) use`
  export. Get this wrong and the failure is a compile error, not a silent one.
- **`prompt/parity.rs` scans `prompt.rs`'s own source** via
  `include_str!("../prompt.rs")` and derives `SHARED_CONTRACTS` /
  `STATIC_PROMPTS` from it. Moving the `macro_rules!` declarations out of that
  file makes `declared_contracts` find nothing, which is
  `ScanError::NoContractsDeclared` — a loud failure by design, but the scan must
  be repointed (read both files, or the new contracts module) in the same PR.
[...]

PR #3068, opened 2026-08-12
fix(stella-cli): split prompt.rs instead of grandfathering it, and unred main

macanderson:
## Why

`main` is red, and has been since #3035 merged at 22:04Z with two checks still failing.

#3035 took `crates/stella-cli/src/agent/prompt.rs` from 1492 lines to 1503 and resolved the ratchet by adding a baseline entry:

```
1503 crates/stella-cli/src/agent/prompt.rs
```

AGENTS.md answers that case directly and in the opposite direction — the baseline "only covers files that predate the guard", it "accepts no new entries", and a file crossing 1500 "gets split, not grown over it". The guard prints the same thing before it exits.

The entry also made `prompt.rs` a god file, and `check-god-files` cross-reads the baseline against AGENTS.md's table and each crate README. Neither was updated, so `docs-guards` fails on `main`:

```
check-god-files: FAIL — AGENTS.md's god-file table omits stella-cli god file(s): src/agent/prompt.rs
check-god-files: FAIL — crates/stella-cli/README.md omits stella-cli god file(s): src/agent/prompt.rs
```

## What this does

There were two ways to turn that green and they are not equivalent. Adding `prompt.rs` to the prose in both places would satisfy the guard while making the tree permanently worse — a 20th god file, closed to growth, in exchange for eleven lines. This removes the baseline entry instead and splits the file, so the god-file count returns to 19 across 7 crates and **no prose changes are needed at all**.

The test module moves out whole into `prompt/tests.rs` — the shape this crate already uses twice (`agent.rs` / `agent/tests.rs`, and the sibling `prompt/parity.rs` this module already declares). It keeps its textual position after `mod parity`, because that ordering is load-bearing: `macro_rules!` scope is textual and the contracts have to be in scope first.

`prompt.rs` lands at **836 lines**, so the next change to it does not meet this wall again.

## Witness

This is a pure refactor plus a guard fix, so the witness is the guards themselves — each fails on `main` and passes here:

```
check-file-size: OK — 1412 files, 27 grandfathered over 1500 lines, and nothing went over by this change.

$ ./scripts/check-god-files.sh
check-god-files: OK — 19 god file(s) across 7 crate(s), named identically in AGENTS.md and every crate README.
```

No production code moved and no test changed; rustfmt rejoined a few lines that fit once dedented.

**No test was deleted.** The 17 `#[test]` functions are the same 17 — verified by diffing the sorted test-name sets against `origin/main`, which is the check the deleted-test guard exists for and which a clean merge would otherwise hide:

```
$ diff <(git show origin/main:.../prompt.rs | rg -A1 '#\[test\]' | rg -o 'fn \w+' | sort) \
       <(cat .../prompt.rs .../prompt/tests.rs | rg -A1 '#\[test\]' | rg -o 'fn \w+' | sort)
IDENTICAL test name sets
```

`cargo fmt --check` clean, `cargo clippy -p stella-cli --all-targets -D warnings` clean, 31 prompt tests pass.

## Note for the reviewer

I did not touch #3035's behaviour change — only the ratchet resolution it shipped with. If the intent really was for `prompt.rs` to become a permanent god file, the alternative is to keep the baseline entry and add it to AGENTS.md's table and `crates/stella-cli/README.md` in the same PR; that turns the same two guards green and is a one-line-each change. This PR takes the direction AGENTS.md specifies, but the call is the maintainer's.

Refs #3035

## Summary by Sourcery

Split the stella-cli prompt tests into a dedicated module to keep the main prompt implementation under the file-size guard and resolve god-file checks.

Enhancements:
- Extract the prompt-related test module from crates/stella-cli/src/agent/prompt.rs into crates/stella-cli/src/agent/prompt/tests.rs to reduce the main file size and avoid treating it as a permanent god file.

Build:
- Remove the file-size baseline entry for crates/stella-cli/src/agent/prompt.rs so future growth is constrained by the guard rather than a grandfathered exception.

Tests:
- Preserve all existing prompt behaviour and guard tests by relocating them unchanged into crates/stella-cli/src/agent/prompt/tests.rs.


macanderson:
I arrived at this independently and opened #3075 before spotting this PR — same split, same baseline deletion, same conclusion. **Yours is first, so I have closed mine as a duplicate.** Three things mine carried that this one does not, offered here so nothing is lost:

### 1. This closes #2985, and that issue's DoD wants a byte-identity check

#2985 ("prompt.rs is 14 lines from the 1500-line ceiling — split the persona literals out before the next contract lands") is exactly this. Worth adding `Closes #2985` **both** in the description and as a commit trailer — per AGENTS.md, squash builds the commit body from commit messages, so the description alone will not carry it.

Its definition of done also requires `SYSTEM_PROMPT` and `PIPELINE_SYSTEM_PROMPT` to be **byte-identical** before and after, since a prompt-prefix byte change is a prompt-cache cost regression (invariant 7 / L-E8), not a formatting detail. I verified this on my branch by capturing each const to a file in both trees and diffing — both match:

| const | sha256 |
|---|---|
| `SYSTEM_PROMPT` | `6b77b5e279f174f43719ee9d8d2af7d4715e19cfe9fccbd6ca96107b37b0318f` |
| `PIPELINE_SYSTEM_PROMPT` | `e592ea9b4cba1ffb6c9c3b7a6085a8aaca4d01e10599e1eddfcea76f97055de8` |

Since this PR moves the same block and leaves every `macro_rules!` and `concat!` in place, the same hashes should hold here — cheap to confirm and it discharges the DoD explicitly.

### 2. Five comments become false at 840 lines

This is the one substantive gap. Four doc comments in `crates/stella-cli/src/agent/prompt/parity.rs` and one in `prompt.rs` justify living where they are by `prompt.rs` sitting *"within a few lines of the 1500-line ratchet"*. That is true at 1503 and false at 840, and leaving it tells the next maintainer this file has no room — the opposite of what the split just bought:

- `parity.rs` — above `both_prompts_batch_independent_tool_calls`
- `parity.rs` — above `both_prompts_bind_the_model_to_skills`
- `parity.rs` — above `the_default_persona_reads_the_delta_only_when_the_task_claims_one`
- `parity.rs` — above `both_personas_route_a_described_behaviour_to_semantic_search_and_a_named_one_to_the_graph`
- `prompt.rs` — above `macro_rules! tool_steering` ("while this file sits against the 1500-line ratchet")

Each also says "beside its siblings in `prompt.rs`'s test module", which after this PR is `prompt/tests.rs`. The fix that worked for me was to drop the ceiling clause and keep the #2237 rationale, which stands on its own — that issue consolidates prompt-content pins into `parity.rs` regardless of how much room `prompt.rs` has.

One more, minor: the `STATIC_PROMPTS` import comment in the moved test block reads "cross-checked against **this file's** own source", which meant `prompt.rs` and now sits in `tests.rs`.

The exact edits are on `fix/prompt-rs-split-not-baselined` (still pushed) if it is easier to cherry-pick the `parity.rs` hunk than to redo it.

### 3. The systemic half is filed as #3073

`make file-size-update` will silently **add** a new baseline path, so the one command anyone reaches for when the ratchet goes red is the command AGENTS.md forbids — which is how `+1503 crates/stella-cli/src/agent/prompt.rs` got onto `main` in the first place, and how `main` went red on `check-god-files` (that guard cross-checks every baseline entry against the AGENTS.md table and the crate README; neither names `prompt.rs`). #3073 proposes giving the updater the property `make typed-errors-update` already has — erroring rather than writing the worse number — and fixing `check-god-files.sh`'s remediation text, which currently reads as permission to grandfather a new file.

Nothing needed here; just linking it so the residue is tracked against this work.

Issue #3073, opened 2026-08-12
gate: make file-size-update silently adds a NEW baseline entry — the one command that turns a red ratchet green is the one AGENTS.md forbids (it landed on main and reds every PR)


macanderson:
## Problem

`make file-size-update` will happily **add a new entry** to
`scripts/file-size-baseline.txt` for a file that has just crossed 1500 lines.
Both AGENTS.md and the guard's own failure text forbid exactly that, so the one
command an agent reaches for when the ratchet goes red is the command that
commits the violation — and it does it silently, printing nothing to say the
edit it just made is the forbidden kind.

`scripts/check-file-size.sh` already refuses the crossing:

```
These files crossed the limit and are NOT grandfathered — split them into
submodules. Do not add a baseline entry: the baseline only covers files
that predate the guard, and this is the rule that stops the tree getting
worse.
  crates/stella-cli/src/agent/prompt.rs is 1503 lines, over the 1500-line limit
```

AGENTS.md § "God files — plan around them, never into them" states the same
rule from the other side: *"the baseline accepts no new entries — a file
approaching the limit gets split, not grown over it."* CLAUDE.md names it in
its list of expedients: *"a raised `scripts/file-size-baseline.txt` ceiling to
turn a gate green ... each one is a bug against the PR that introduced it."*

Nothing enforces it against the *writer*. The check is one-sided.

## It has already happened, and it landed on `main`

PR #3035 pushed `crates/stella-cli/src/agent/prompt.rs` to 1503 lines and the
ratchet failed it. The failure was cleared by commit `f9045da05` ("chore: file
size ratchet"), whose entire diff is `scripts/file-size-baseline.txt`:

```
-2141 crates/stella-cli/src/agent.rs
+2140 crates/stella-cli/src/agent.rs
+1503 crates/stella-cli/src/agent/prompt.rs      <-- new grandfathered entry
-2094 crates/stella-model/src/openai.rs
+2078 crates/stella-model/src/openai.rs
-2941 crates/stella-pipeline/src/pipeline.rs
+2792 crates/stella-pipeline/src/pipeline.rs
-1967 crates/stella-tools/src/registry.rs
+1744 crates/stella-tools/src/registry.rs
```

The retightenings are ordinary. The `+1503` line is a brand-new god file, and
the PR merged with it.

**It also left `main` red.** `scripts/check-god-files.sh` (#1435) cross-checks
every baseline entry against the god-file table in AGENTS.md and the matching
[...]

Issue #1435, opened 2026-08-05
gate: AGENTS.md's god-file table and crate README god-file lists hand-mirror file-size-baseline.txt with no drift guard


macanderson:
## Problem

`scripts/check-file-size.sh` enforces the 1500-line ratchet from `scripts/file-size-baseline.txt` (regenerated by `make file-size-update`). Two prose copies of *which files are god files* exist and nothing checks either:

1. The per-crate table in `AGENTS.md` § "God files — plan around them, never into them".
2. The "God files — do not add lines" section in each god-file crate's README (`crates/stella-cli`, `-core`, `-model`, `-pipeline`, `-protocol`, `-store`, `-tools`, `-tui`), plus the "this crate has no god files" claim in the other twelve crates' READMEs.

All copies were verified in sync on 2026-08-04 (diffed by hand against the baseline during the docs-consistency sweep), but `make file-size-update` rewrites the baseline and touches neither prose site — the next file split or rename strands both silently. This is the exact failure shape that motivated `scripts/check-invariants.sh` (#630): duplicated normative prose with no tiebreaker.

## Suggested shape

Extend `scripts/check-file-size.sh`, or add a `scripts/check-god-file-tables.sh` wired into `GATE_GUARDS_FAST` in the `Makefile` (see the comment block there for how the guard tiers compose) and into CI:

- Derive the per-crate god-file set from `file-size-baseline.txt` (entries under `crates/<crate>/`; the bench Python entries have no prose table and are out of scope).
- Assert `AGENTS.md`'s table row for each crate lists exactly those files (paths in the table are crate-relative, e.g. `src/agent.rs`).
- Assert each crate README's god-file list matches, and that crates absent from the baseline claim "no god files".
- Keep it POSIX-portable like `check-invariants.sh` — it must run on a bare CI runner.

Wire it into both workflows the way the other prose guards are: `ci.yml`'s required job and `docs-guards.yml` (the latter because `**/*.md` edits skip `ci.yml`).

## Constraints already decided

- `AGENTS.md` deliberately does **not** repeat the numeric ceilings — only file names. The baseline stays the only copy of the numbers. Don't "fix" that by adding numbers to prose.
- The table's value is naming *which* files are closed to growth; keep the guard about set-membership, not line counts.

## Done when

[...]

PR #3075, opened 2026-08-12
fix(stella-cli): split prompt.rs rather than grandfather it, and unred main

### LINKED ISSUE #2985: prompt.rs is 14 lines from the 1500-line ceiling — split the persona literals out before the next contract lands

macanderson:
## What & why

#3035 pushed `crates/stella-cli/src/agent/prompt.rs` to 1503 lines and the
file-size ratchet failed it. The failure was cleared by `make file-size-update`,
which wrote a **new** grandfathered entry into `scripts/file-size-baseline.txt`
(commit `f9045da05`, whose whole diff is that file):

```
+1503 crates/stella-cli/src/agent/prompt.rs
```

That is the one remedy the guard's own failure text rules out — *"Do not add a
baseline entry: the baseline only covers files that predate the guard, and this
is the rule that stops the tree getting worse"* — and AGENTS.md § "God files"
says it from the other side: the baseline accepts no new entries; a file
approaching the limit gets split.

**It also left `main` red.** `check-god-files.sh` (#1435) cross-checks every
baseline entry against the god-file table in AGENTS.md and the matching crate
README, and neither names `src/agent/prompt.rs`. Against `main` at `dce1c25af`:

```
check-god-files: FAIL — AGENTS.md's god-file table omits stella-cli god file(s): src/agent/prompt.rs
check-god-files: FAIL — crates/stella-cli/README.md omits stella-cli god file(s): src/agent/prompt.rs
```

So this PR takes the split the guard asked for and deletes the entry. The inline
`mod tests` moves verbatim to `crates/stella-cli/src/agent/prompt/tests.rs`,
beside the `parity` submodule that already establishes the shape.

| | before | after |
|---|---|---|
| `agent/prompt.rs` | 1503 | **840** |
| `agent/prompt/tests.rs` | — | 672 |
| baseline entry | `1503 …/prompt.rs` | **deleted** |

No prose needed updating, precisely because the file stopped being a god file.

**Why the tests are the right half to move.** What makes `prompt.rs` grow is the
prose above each clause explaining why it is worded as it is — #3035's own
overage was eight lines of doc comment explaining why the escalation clause
deliberately names no tool (`ask_user` is only registered when a human can
answer, and these are static bytes shared by attended and unattended runs). That
is the half worth spending a ceiling on.

**Why not the seam #2985 proposed.** That issue suggested splitting the
`macro_rules!` contracts into `prompt/contracts.rs`, and listed the constraints
that makes it expensive: `parity.rs` does `include_str!("../prompt.rs")` and
derives `SHARED_CONTRACTS` / `STATIC_PROMPTS` from it, so moving the
declarations out is `ScanError::NoContractsDeclared` unless the scan is
repointed in the same PR, and `macro_rules!` textual scope forces a declaration
order. Moving the tests instead leaves every macro and every `concat!` exactly
where it is, reaches the same definition of done, and touches none of that.
Flagging it explicitly so a reviewer can judge the substitution.

Closes #2985

## The witness

The witness is the guard itself, and it was run against both trees rather than
assumed:

- **Fails on `main`** (`dce1c25af`, in a detached worktree at that commit):
  `check-god-files` prints the two FAIL lines quoted above. `check-file-size`
  passes there only because the forbidden entry is present.
- **Passes here**: `check-file-size: OK … nothing went over by this change` and
  `check-god-files: OK — 19 god file(s) across 7 crate(s), named identically in
  AGENTS.md and every crate README.`

#2985's definition of done also requires the two prompts be **byte-identical**
before and after, since a prompt-prefix byte change is a prompt-cache cost
regression (invariant 7 / L-E8), not a formatting detail. Verified empirically —
each const captured to a file in both trees and diffed:

| const | sha256 on `main` | sha256 here |
|---|---|---|
| `SYSTEM_PROMPT` | `6b77b5e2…0318f` | `6b77b5e2…0318f` |
| `PIPELINE_SYSTEM_PROMPT` | `e592ea9b…5de8` | `e592ea9b…5de8` |

All 17 tests move unchanged and still run as `agent::prompt::tests::*` — 31
passing under `agent::prompt::` counting `parity`'s scan tests. A move between
files, so no test is removed from the tree and `check-deleted-tests` has nothing
to report (it `git grep`s the whole tree at each revision; its own header notes
that moving "between files or modules is silently fine").

## The gate

- [x] `cargo fmt --check`
- [ ] `cargo test --workspace` — ran `cargo test -p stella-cli --bin stella`
      scoped to `agent::prompt::`; the crate's integration binaries were not
      built locally (they OOM this machine). CI covers the workspace.
- [x] Docs updated where behavior changed — see below
- [x] `Closes #2985` appears both here and as a commit trailer

Four comments in `parity.rs` and one in `prompt.rs` justified living there by
`prompt.rs` sitting *"within a few lines of the 1500-line ratchet"*. That is
false at 840 lines, and leaving it would tell the next maintainer this file has
no room. Each now cites #2237 — which consolidates prompt-content pins into
`parity.rs` and stands on its own — instead of a ceiling that no longer binds.
The `STATIC_PROMPTS` note that read "cross-checked against this file's own
source" now names `prompt.rs`, since it no longer sits in that file.

## Nothing left behind

- [x] Filed: #3073

#3073 is the systemic half: `make file-size-update` will silently add a new
baseline path, so the one command an agent reaches for when the ratchet goes red
is the command the rules forbid — and nothing between it and `main` says no.
`make typed-errors-update` already has the property this needs (it errors rather
than write a worse number); the issue proposes the same for this baseline, plus
fixing `check-god-files.sh`'s remediation text, which currently reads as
permission to grandfather a new file.

## Ground-rule check

- [x] Byte-stable prompts preserved — proven by the hashes above (invariant 7)

- The baseline is one shared cell and per #2004 the top cause of a red `main`,
  so this is a **surgical one-line deletion**, not a regeneration. The three
  retightenings `f9045da05` also made (`openai.rs`, `pipeline.rs`,
  `registry.rs`) are left exactly as they are.
- AGENTS.md warns against folding a baseline change into an unrelated PR. This
  PR is only about that baseline entry and the split that retires it, branched
  from a fresh `main`.
- Splitting buys structure, not slack: `make file-size-update` retightens every
  ceiling to the file's current size, so nothing here creates room for a
  grandfathered file.

## Summary by Sourcery

Split the stella-cli agent prompt module tests into a dedicated file to bring the main prompt source back under the file-size ceiling and retire its baseline entry.

Enhancements:
- Extract the inline `agent::prompt` test module into `prompt/tests.rs` to keep prompt prose concentrated in the main source file without changing behavior.
- Refresh comments in `prompt.rs` and `prompt/parity.rs` to reflect that prompt-content provenance and pins now live in parity/tests per #2237 rather than a near-limit god file.

Build:
- Remove the grandfathered file-size baseline entry for `crates/stella-cli/src/agent/prompt.rs` so the guard tracks the split files without exempting the prompt source.


macanderson:
Duplicate of #3068, which reached the same split and the same baseline deletion about 40 minutes earlier. Deferring to it — two PRs editing `scripts/file-size-baseline.txt` is the shared-cell collision #2004 documents as the top cause of a red main, so only one should land.

The three things this PR carried that #3068 does not are handed over there: https://github.com/macanderson/stella/pull/3068#issuecomment-5273464487 — the byte-identity verification #2985's definition of done asks for, the five comments that become false once prompt.rs is 840 lines, and #3073 for the systemic half.

Branch `fix/prompt-rs-split-not-baselined` stays pushed in case the parity.rs comment hunk is easier to cherry-pick than to redo.

Issue #3032, opened 2026-08-12
the tool menu is 72 schemas and 71% of the opening context — Claude Code sends 25, and this is the best explanation for the 2.1x call-count gap


macanderson:
## Summary

Stella sends **72 tool schemas** on every request — **11,797 tokens, 71% of the
opening context**. Claude Code sends **25**, of which only five are coding tools.

Measured on the same tasks, same model (claude-sonnet-5), same day: Stella
takes 30.0 tool calls per trial to Claude Code's 12.4, and costs 2.1x per
trial. The entire gap is call count, and this is the best available explanation
for it — better than the steering and prompt hypotheses that have been tried
and measured flat.

## Evidence

Captured from a real outbound request via a local proxy (`--base-url` pointed
at a recorder that logs the body and forwards upstream):

```
tools=72   graph_query=YES   47,190 chars ~= 11,797 tokens
```

Claude Code's own init banner, from a bench trial in run `ccFull`:

```
tools=25  Task Bash CronCreate CronDelete CronList Edit EnterWorktree
          ExitWorktree NotebookEdit Read ReportFindings ScheduleWakeup
          SendMessage Skill TaskCreate TaskGet TaskList TaskOutput TaskStop
          TaskUpdate ToolSearch WebFetch WebSearch Workflow Write
```

Its coding surface is five tools — `Bash`, `Read`, `Edit`, `Write`,
`NotebookEdit`. The other twenty are session management, and `ToolSearch` is how
anything else is reached.

Stella's 72 include heavy overlap within single jobs:

| job | tools |
|---|---|
| file work | `read_file` `write_file` `edit_file` `apply_edits` `delete_file` |
| finding code | `grep` `glob` `graph_query` `read_symbol` `project_overview` `gather_context` |
| git | `repo_status` `repo_diff` `repo_commit` `repo_push` `repo_pull` `repo_history` `repo_rollback` `repo_recover` |
| issues | `create_issue` `close_issue` `get_issue` `update_issue` `search_issues` `list_labels` `list_members` `start_work_on_issue` |

## What this explains that other hypotheses did not

Across 408 bench trials:

```
bash                 10,835
grep                    120
repo_status              67
project_overview         33
read_symbol              23
graph_query               3
semantic_code_search      0
```

`bash` absorbs everything. Specialised tools are offered and not chosen.

**The code-graph tool has been uncalled for three months and the cause was
assumed to be discoverability.** It is not:

- The schema **is** in the outbound request (verified above).
[...]

Issue #923, opened 2026-07-30
Speculation can execute a read-only tool twice per step; read_only and speculation_safe are not the same claim


macanderson:
## What

The speculative-execution path runs a response's announced read-only tool prefix before the stream completes. If the stream attempt then fails and is retried, the retry re-announces the same prefix and executes it again.

This is documented at length in `stella-core/src/driver.rs`'s module docs and is deliberate. The issue is not that it is undocumented — it is that `read_only` is being asked to carry two different claims:

1. **This tool does not mutate the workspace.** True of every tool marked read-only.
2. **This tool is safe to run twice.** Not true of all of them.

A tool can satisfy (1) and violate (2) whenever it touches anything outside the working tree:

- a web fetch or search tool burning a paid, rate-limited, or quota-metered API call
- an MCP tool whose server counts requests, logs them, or is itself rate-limited
- anything that writes to internal state as a side effect of reading — `codegraph.db` being the in-tree example

The workspace stays correct in every one of these cases. The user's quota, the remote's rate limit, and the observability plane do not.

## Current escape hatch

There is one: attach a `PreToolUse`/`PostToolUse` hook, which routes the tool through `hook_gated` and out of the speculative prefix. That works, but it means the way to say "do not speculate on this" is to attach an unrelated feature as a side effect, which nobody will discover from the tool-authoring docs and nobody will remember to keep attached.

## Proposed shape

Add `ToolSchema::speculation_safe`, distinct from `read_only`:

- `read_only` keeps its current meaning and its current uses (judge enforcement, permission surfaces).
- `speculation_safe` defaults to... this is the real decision. Defaulting it to `read_only`'s value preserves today's behaviour and requires every network-backed tool author to remember to opt out. Defaulting it to `false` is safe-by-default and costs a small amount of speculation coverage until built-in tools are annotated. I would default it to `false` and annotate the in-tree filesystem tools explicitly, on the grounds that the failure mode of the other default is invisible and lands on the user's bill.

Then the speculative prefix is built from `speculation_safe`, not `read_only`.

## Acceptance

A tool can declare itself unsafe to speculate without attaching a hook; built-in filesystem tools remain speculated; the `web` tool family and MCP-provided tools do not.

PR #3083, opened 2026-08-12
fix(stella-tools): declare `search` in the catalog, sync Cargo.lock and the tool docs — main is red

macanderson:
## `main` is red — this is the fix

#3076 merged while its last commits were still in flight, so `main` has the `search` tool registered but **not declared** in `catalog::CATALOG`. Six tests fail on `main` right now.

Four registry tests:

```
registry::tests::every_registry_tool_is_reserved_against_custom_shadowing
registry::tests::issue_tools_absent_without_a_configured_backend
registry::tests::read_only_flags_partition_the_registry_correctly
registry::tests::registry_advertises_exactly_the_catalog_tool_set
```

with, verbatim:

```
built-in tool `search` is missing from custom::RESERVED_NAMES — declare it in
stella-tools/src/catalog.rs, or a custom manifest could shadow it
```

and two doc-sync tests:

```
docs_in_sync::docs_counts_match_the_catalog
docs_in_sync::permissions_read_only_prose_matches_the_native_catalog
```

A red `main` reds every open PR, so this is deliberately small.

## Three commits

**1. Declare `search` in `catalog::CATALOG`.**

```rust
"search"              => (true, false, Always, "search"),
```

- **read-only: true** — it never mutates the workspace.
- **speculation_safe: false** — it opens the code graph (which bootstraps/catches up `codegraph.db`) and its semantic rung stores the vectors it embeds on the way to answering. A read that writes (#923), the same posture as `graph_query` and `semantic_code_search`.
- **Always** — registered unconditionally like the rest of the retrieval family: a graph tool builds its own index on first use, so gating on the index existing would hide the tool meant to create it.

**2. Sync `Cargo.lock` — a second, independent break on `main`.** The lock still records `stella-tty 0.9.16` while the workspace manifest is `0.9.17`, so `cargo metadata --locked` exits 101 and **every `--locked` build fails** — which is how `install.sh` builds. One generated line, no dependency change. Folded in rather than filed because `main` is currently unbuildable with `--locked` and this PR is already the red-main fix.

**3. Sync the tool docs.** `docs_in_sync` derives its numbers from `catalog.rs` rather than trusting prose, and caught two things: the always-on count (58 → 59) with the native ceiling beside it, and `permissions.mdx`'s read-only set — a *second* copy of the list `index.mdx` carries, which was missed when the first was updated in #3076.

## Verification

`make gate` run in full on this branch, output redirected to a file and read — not piped through `tail`, which masks the exit code and has produced false greens here:

```
EXIT=0
139 test-suite summaries, all ok
0 × "test result: FAILED", 0 × "make: ***"
```

The six failing tests above pass here. No new test is added: the declared-catalog and docs-sync tests are exactly the mechanism that catches this class of gap, and they caught it — adding another would be testing the guard rather than the gap.

## Deleted tests

None.

Refs #3032


macanderson:
Second, independent break on `main` found while gating this: `Cargo.lock` still records `stella-tty 0.9.16` while the workspace manifest is `0.9.17`, so `cargo metadata --locked` exits 101 and every `--locked` build fails — which is how `install.sh` builds. Pushed as a separate commit on this branch (one generated line, no dependency change). Folding it in rather than filing it because `main` is currently unbuildable with `--locked` and this PR is already the red-main fix.

macanderson:

|---|---|---|---|---|

macanderson:

```
```

Revised coverage:

|---|---|---|---|---|

PR #3086, opened 2026-08-12
fix(stella-tools): put `search` in the catalog, resync the lock, record the registry's new ceiling

macanderson:
fix(stella-tools): put `search` in the catalog, resync the lock, record the registry's new ceiling

`main` is red on three steps of its own `ci` run (31647120961), and all three
arrived with #3076.

**Four registry tests fail.** #3076 registered `crate::search::Search` in
`ToolRegistry` but never added it to `catalog::CATALOG`, and the catalog is
the declared set the registry is checked against from both sides. The
assertion says what to do —

    registry disagrees with catalog::CATALOG — register the tool, or
    add/remove its line in stella-tools/src/catalog.rs

— and one line does it, so all four pass together:
`registry_advertises_exactly_the_catalog_tool_set`,
`every_registry_tool_is_reserved_against_custom_shadowing`,
`issue_tools_absent_without_a_configured_backend`,
`read_only_flags_partition_the_registry_correctly`.

The line is `(true, false, Always, "search")`. Read-only and always-on match
the tool's own `ToolSchema`. The middle column is the one worth arguing: the
tool is **not** speculation-safe, because it fuses `graph_query` and
`semantic_code_search`, and both of those are the "read that writes" —
codegraph catch-up and stored vectors. A fused tool inherits the posture of
the strictest branch it can take, never the cheapest, or speculation runs a
write twice.

**Cargo.lock is out of sync.** One line: `stella-tty` still says 0.9.16 while
the workspace is 0.9.17. The 0.9.17 version-sync did not carry it, and
`--locked` refuses the tree, which is also why #3080 fails `cargo check on
the declared MSRV` — an inherited red, not a defect in that PR.

**registry.rs is over its ceiling.** #3076 grew it 1744 → 1751 without
regenerating the baseline. Those 7 lines are one `entries.push(...)` and its
comment in a file whose whole job is that list, which is the irreducible
declaration the guard's own message names, so this takes the sanctioned route
— `make file-size-update`, landing as a reviewable one-line baseline diff
(1744 → 1751, no other ceiling moved) rather than a silent widening.

Landed from a fresh `main` and on its own, per AGENTS.md: a baseline
regeneration folded into an unrelated PR is how the shared cell goes red for
everyone.

No witness test: every change here restores an existing gate that fails on
`main` and passes on this commit, which is the evidence. Verified locally —
`check-file-size.sh` OK, `check-god-files.sh` OK (19 god files), `cargo check
--workspace --locked` clean, and `registry::tests` 96 passed / 0 failed
against 4 failing before.

## Summary by Sourcery

Restore tool registry invariants and CI gates by aligning the catalog, lockfile, and file-size baseline with recent changes.

Bug Fixes:
- Add the fused `search` tool to the catalog to match the tool registry and fix failing registry tests.
- Resync Cargo.lock with the workspace versions so locked workspace checks succeed.
- Update the file-size baseline for registry.rs to reflect its current size and keep file-size and god-file guards passing.


macanderson:

|---|---|---|---|---|

macanderson:

```
```

Revised coverage:

|---|---|---|---|---|

macanderson:
Correction to my earlier comments on this PR, #3083 and #3085. I claimed that merging #3083 alone would leave `main` red on the file-size ratchet, and that only a baseline entry of 1751 would clear it. **That was wrong**, and the evidence is on `main`.

`main`'s `ci` run for #3083's merge (31648969990, `b3164d602`) — where `registry.rs` was 1751 and the baseline still said 1744:

```
no new .rs file over the 1500-line ratchet: success
Cargo.lock is in sync with the manifests:   success
cargo test:                                 success
```

So #3083 did unred `main` on its own. The baseline line in this PR was not load-bearing for that.

What I got wrong is the rule, not the reading. The ratchet judges a change against its base tree on a push to `main` too — the base being the merge commit's first parent — so it fails exactly once, on the merge that *causes* the growth, and every later merge inherits 1751 on both sides and is classified as drift: reported, survived, not absorbed. `main`'s ci did fail that step once, on #3076's own merge (run 31647120961), and I generalised from that single observation to "it will keep failing" without checking the rule that AGENTS.md actually states. Evidence consistent with both answers, read as though it settled one.

What the baseline change in this PR does still buy, stated honestly and no larger:

- it clears drift the guard reports on every run until someone fixes it, and
- it stops the next PR that touches `registry.rs` from failing, since that PR's own growth would be measured against a ceiling seven lines below the file it inherited.

Both are worth having. Neither is "unredding `main`", which is what I said.

The rest of the earlier analysis stands and was checked against test output rather than inferred: the catalog line, the two `docs_in_sync` counts (58→59, 70→71) with `search` added to the read-only prose, and the `Cargo.lock` bump for `stella-tty` were each required, and `main` is green on all of them now — `registry::tests` 96 passed, `docs_in_sync` 9 passed, `cargo check --workspace --locked` clean, `check-file-size` and `check-god-files` OK.

Issue #3142, opened 2026-08-13
search: the callers facet conflates same-named symbols across files and languages, and does not say it is name-based


macanderson:
## Problem

`search`'s enrichment (`crates/stella-tools/src/search/enrich.rs`, `Facet::Callers`) lists `graph.callers(&symbol.name)` for a hit's leading symbols. The lookup is name-based by design (#335: reverse name lookup, no resolution), so a hit whose leading symbol wears a common name shows callers of *unrelated* symbols — observed live on this repo: searching `SearchConfig` renders `fn test_the_label_is_a_set_not_a_sequence (arenabench/tests/test_tool_set.py:126)` as a "caller", because the Rust file's leading symbol `label` name-matches a Python call site.

The callees side was scoped to the hit's own file in the parent PR (`frame_is_about`), because a callee list is *about* this file's definition. Callers legitimately live in other files, so the same filter is wrong there — but cross-**language** hits are almost never the same symbol, and the line does not disclose that it is a name match at all (the `why:` line's "not by meaning" disclosure covers the ranking, not the facets).

## Direction

- Filter caller frames to call sites in the same language as the hit file (the graph stores `language` per file), which removes the worst conflations without claiming resolution.
- And/or label the line `callers (by name):` so the model weighs it as the best-effort signal it is — the same honesty `graph_query callers` already has in its own docs.

## Definition of done

- The `SearchConfig` repro above no longer lists Python call sites for a Rust hit, or the line labels itself name-based; a test pins whichever is chosen.

## Files

- `crates/stella-tools/src/search/enrich.rs` — `Facet::Callers` branch, `frame_labels`
- `crates/stella-graph/src/frames.rs` — `callers` (if per-language filtering lands there instead)

Related: #335 (name-based call edges, non-goals), #3138 (name-rung ranking).

Issue #3141, opened 2026-08-13
stella-tools: a panicked spawn_blocking in search/graph tools is reported as "cancelled"


macanderson:
## Problem

Both `crates/stella-tools/src/search.rs::run_search` and `crates/stella-tools/src/graph/semantic.rs` wrap `open_or_build` in `tokio::task::spawn_blocking` and map `Err(JoinError)` to "the search was cancelled" / "the code-graph query was cancelled". A `JoinError` is either a cancellation **or a panic** in the blocking closure; reporting a panic as a cancellation sends whoever reads the message to debug the wrong thing (a shutdown race instead of a crash in the indexer).

## Fix

`JoinError::is_panic()` distinguishes the two. Report a panic as one, ideally with the payload (`JoinError::into_panic` + downcast to `&str`/`String`), in both files with the same wording so the siblings stay in lockstep.

## Definition of done

- A test that panics inside the blocking closure (e.g. via a poisoned fixture or an injected panic hook) asserts the error message says "panicked", not "cancelled".
- Both call sites updated together.

## Files

- `crates/stella-tools/src/search.rs` — `run_search`'s `Err(_) =>` arm
- `crates/stella-tools/src/graph/semantic.rs` — the same shape (~line 166)

Issue #3140, opened 2026-08-13
search: the character budget bounds an estimate, not the rendered answer — a real answer can exceed STELLA_SEARCH_BUDGET


macanderson:
## Problem

`stella_core::search::allocate` spends `Facet::estimated_cost` values (`Body` = 1,600 chars), but the renderer (`crates/stella-tools/src/search/enrich.rs`) then renders the real thing: a `Body` facet quotes up to `MAX_BODY_LINES = 40` source lines, which at ordinary line lengths is 3–4,000 characters, and caller/importer lists vary similarly. So an answer can exceed `STELLA_SEARCH_BUDGET` by a small multiple while reporting nothing — the budget is honest about what it *dropped* but not about what it *spent*.

`Facet::estimated_cost`'s doc states the trade deliberately ("an estimate that is uniformly a little wrong still orders the ladder correctly"), and ordering is indeed unaffected. What is not covered is the absolute overshoot: an operator setting `STELLA_SEARCH_BUDGET=9000` to bound context spend can receive materially more. Since the promotion pass (this issue's parent PR) now grants deep facets much more often, the overshoot is more common than when depth 4 kept everything shallow.

## Direction

- Cheapest: measure the rendered block per hit and stop appending whole hit blocks once the real total passes the budget, reporting the stop in the existing TRUNCATED line. Whole-block granularity keeps the monotone-facet property intact (never truncate inside a block — a partial block would break the depth-superset test).
- Alternatively: recalibrate `estimated_cost` against measured renders on this repo and accept a bounded error.

## Definition of done

- A test renders an answer whose estimated cost fits the budget but whose real render exceeds it, and asserts the rendered length is within one hit-block of the budget with the truncation line present.

## Files

- `crates/stella-core/src/search.rs` — `Facet::estimated_cost`, `allocate`
- `crates/stella-tools/src/search.rs` — `render`
- `crates/stella-tools/src/search/enrich.rs` — `render_hit`, `MAX_BODY_LINES`

Issue #3139, opened 2026-08-13
search: the schema description promises meaning-matching that an embedder-less host cannot deliver


macanderson:
## Problem

`Search::schema()` (`crates/stella-tools/src/search.rs`) tells the model, in caps, that the tool "matches by MEANING, not just by text" and that the words used "do not have to appear anywhere in the code". That is true only when an embedder resolves (`STELLA_EMBED_URL`/`STELLA_EMBED_MODEL`, or the provider-key shortcuts). On a host with none configured — which is the *normal* case on Terminal-Bench per the module's own docs — the semantic rung never runs and every prose query degrades to term matching (see #3138 for how that ranks today).

The **answer** is honest (`via: names … (not by meaning)`), but the **description** is what the model decides on, and a model that believes the meaning promise will trust a term-matcher's miss as semantic absence — the exact "miss reads as evidence of absence" failure the coverage note exists to prevent, one level up.

## Direction

Two options, either acceptable:

1. **Configure an embedder for the benchmark** (and document that the description assumes one). #2994's in-process embedder would make this moot.
2. **Vary the description at registration time.** `Search::from_env()` already runs once at registration; `stella_embed::from_env()` resolution is known at the same moment. A description that says "matches by meaning" only when an embedder resolved stays byte-stable for the whole session (invariant 7 holds — the schema is fixed per session, which is the actual cache boundary), and tells the truth on hosts where it can't.

## Definition of done

- On a host with no embedder, the advertised description no longer promises meaning-matching, OR the benchmark environment pins an embedder and the decision is recorded.
- A test pins whichever behaviour is chosen (the existing `the_schema_offers_one_query_and_no_mode_selector` asserts on the description text and will need to know about the variant).

## Files

- `crates/stella-tools/src/search.rs` — `Search::schema`, `Search::from_env`
- `crates/stella-embed/src/http.rs` — `Resolution`

Related: #2994, #3138, #3032 (the fused-tool experiment this tool exists for).

PR #3154, opened 2026-08-13
feat(stella-tools): harden search for benchmark use and add stella search

### LINKED ISSUE #3125: search: the strategy ladder never leaves its first rung, so an exact symbol name loses to semantic ranking
## Summary

`dispatch` in `crates/stella-tools/src/search.rs` is a ladder: semantic first,
then graph symbol names, then a file scan, "stopping at the first that produces
hits". With an embedder configured, **the first rung always produces hits**, so
the other two are unreachable in practice — and the case they exist to serve,
an exact symbol name, is precisely the case semantic ranking is worst at.

Measured over 9 queries on a fully-indexed 158-file corpus: `via: semantic` on
**9 of 9**, and 8 of those returned all 158 files as hits.

## Evidence

Corpus: `github.com/macanderson/context-graph-protocol` at `9ce2a83b`
(158 indexable files, 1,814 symbols), embedder `voyage-code-3` via
`VOYAGE_API_KEY`, index warmed to its ceiling.

The probe that isolates it is the bare symbol name `ContextFrame`, which the
tool's own schema advertises as valid input. There is exactly one definition —
`pub struct ContextFrame` at `contextgraph-types/src/frame.rs:241`. Result:

```
search `ContextFrame` — 139 result(s) at depth 4
via: semantic (embedding rank over the code-graph index)
1. contextgraph-host/src/host.rs
2. contextgraph-conformance/src/host_conformance.rs
3. contextgraph-types/src/query.rs
4. contextgraph-conformance/src/bin/contextgraph-example-docs.rs
5. contextgraph-types/src/frame.rs      <- the definition
```

`Strategy::GraphNames` (`enrich::name_hits`) would have matched the symbol
`ContextFrame` exactly and put its file first. It never ran, because semantic
had already returned 158 hits.

Why semantic is never empty is the compounding half: the admission floor
defaults to 0.25 (`DEFAULT_ADMISSION_FLOOR`), while the **lowest** cosine
anywhere in this corpus against an unrelated query was 0.418. The floor rejects
nothing, so `semantic_hits` cannot return an empty list, so the ladder cannot
advance. That measurement is #2993; this issue is the structural consequence and
is worth fixing independently, because a floor tuned to reject 90% of a corpus
would still admit *something* for nearly every query.

## Why not just fuse them

The module doc already answers this and is right: fusing three incomparable
score scales into one ordering "would make the answer's `via:` line a lie". The
fix is not fusion.

## Suggested direction (not prescriptive)

The signal that is being discarded is that the query **is** a symbol the graph
knows. That is a fact about the index, not a mode the model selects, so
consulting it does not reintroduce a mode parameter (invariant 9):

- Before ranking, ask the graph whether the query string names a symbol
  exactly. If it does, that file is a certainty and the semantic ranking is at
  best a supplement — surface the exact hit first and label it, keeping
  `via:` honest by naming both rungs.
- Equivalently: let `GraphNames` run whenever it has an **exact** match,
  regardless of whether semantic produced hits, and say so in `via:`.

Either keeps one tool with no mode flag and stops discarding a certain answer in
favour of an uncertain one.

## Definition of done

- A query that exactly names an indexed symbol returns that symbol's file at
  rank 1.
- `via:` reports every rung that contributed; no rung is silently skipped.
- Witness test: index a fixture with a uniquely-named symbol, query that name
  with a real (or stubbed high-scoring) embedder configured, assert rank 1 is
  the defining file. Fails on `main` today — semantic ranking alone puts it at
  rank 5 of 139 on the corpus above.
- Add the bare-symbol probe to `crates/stella-tools/tests/search_recall.rs`'s
  corpus fixture so the regression is measured, not just asserted.

## Files

- `crates/stella-tools/src/search.rs` — `dispatch`, `semantic_hits`
- `crates/stella-tools/src/search/enrich.rs` — `name_hits`
- `crates/stella-embed/src/http.rs` — `DEFAULT_ADMISSION_FLOOR`

Related: #2993 (the floor is unmeasured), #3097 (retrieval witnesses),
#3089 (chunk-level embedding), #3124 (the coverage note this run also found).

macanderson:
## What & why

The fused `search` tool (#3076) will be the only retrieval tool the next benchmark surfaces, so this PR reviews it end to end, fixes every way it was quietly weaker than its own contract, and adds `stella search <query>` so the same engine answers at the CLI. Design exemplar for the subcommand: the existing `stella graph` / `graph_query` pairing (`contextgraph::run_graph`); the engine is shared via a `pub` `run_search`, exactly as `stella_tools::graph::run_query` is.

**The default answer now spends its budget.** `stella-core`'s allocator gains a promotion pass — whatever the decayed pass leaves unspent deepens the granted hits back toward the requested depth, best-ranked first — and `Depth::DEFAULT` moves to `MAX`, so the budget (which degrades gracefully and reports what it drops) is the binding constraint instead of a silent depth cap. Before this, a three-hit answer spent under a tenth of its 9,000-character allowance and could never render a signature, doc, caller, or body at any budget — while `DEFAULT_BUDGET_CHARS`' own doc said the budget was sized for a body excerpt. The measured shape on this repo now: top one-two hits carry symbols, kinds, imports, importers, signatures, docs, callers, callees and a line-numbered body; the tail stays pointers; the 9,000-char ceiling still binds.

**An exact symbol name outranks the semantic guess** (#3125): a query that exactly names an indexed symbol pins its defining files above the ranked results, `via:` names both contributors, the matched symbol leads the detailed facets, and the bare-symbol probe joins `search_recall.rs`.

**Every silent truncation now speaks:** the file scan discloses hitting its 4,000-file cap (and walks shallow-first, so a capped walk keeps the root and `src/`, not the alphabetically last subtree the old stack-order walk kept); cut name/scan lists say how many files matched; `RANK_CEILING`'s "reported, never silent" doc promise is now actually implemented; an unreadable coverage counter reads as "coverage unknown" instead of as full coverage.

**Enrichment describes what matched:** chunk hits carry their matched symbol as a focus, so `body`/`signature` quote the symbol the query is about rather than whatever sits first in the file; caller entries render their located `citation_label` instead of the site-less `caller of X` title; callee lookups drop same-named definitions from other files; doc-comment walks skip Rust attributes instead of quoting `#[must_use]` as prose.

**`stella search`**: new subcommand + docs page, wired keyless (local index; the only network is an explicitly configured embedder), with the empty-query guard moved into the shared engine so both surfaces refuse it.

Closes #3125

## The witness

- [x] This PR includes witness tests (fail on `main`, pass here):
  - `search::tests::the_default_answer_carries_the_body_without_a_follow_up_read` — verified the artisanal way: fails with `crates/stella-core/src/search.rs` stashed to `HEAD`, passes with it.
  - `search::tests::an_exact_symbol_name_outranks_the_semantic_guess` — #3125's witness: a uniform-scoring embedder ties every file, and only the exact rung puts the defining file first.
  - `stella-core` `search::tests::leftover_budget_deepens_the_tail_instead_of_going_unspent` — the promotion witness.
  - Plus: capped-scan disclosure, shallow-first walk order, cut-list disclosure, ceiling note, focus-led facets, caller citations, attribute-skipping doc walk (9 new tests in `stella-tools`, 3 in `stella-core`, one new property).

## The gate

- [x] `cargo fmt --check`
- [x] `cargo test --workspace` — with two flakes on the sandboxed runner, both investigated, neither this diff's: `daemon::tests::a_run_that_ignores_term_is_killed_after_the_grace_period` fails identically on `origin/main` (8a86567) in that container — root cause (the liveness-lock readiness witness is armed in `pre_exec`, before `sh` installs its `trap`) posted with a 15/15 repro on #1721; and one full-suite run had 2 `stella-tools` lib failures that six re-runs (including 3 at `--test-threads=32` and 2 under concurrent build load) could not reproduce — 969/969 green each time. Every workspace test target passed in at least one full-suite run in that container.
- [x] Docs updated: `website/content/docs/commands/search.mdx` (new), `commands/index.mdx`, `commands/meta.json`, `agent-tools/index.mdx`, `--help` text
- [x] CLA signed
- [x] `Closes #3125` appears both above and as a commit trailer

**Deleted test, named deliberately:** the `stella-core` property `no_hit_is_granted_more_than_it_asked_for` is replaced by `grants_respect_the_ceiling_and_never_rise_with_rank` — the old assertion (granted ≤ decayed request) is the exact behaviour the promotion pass removes; the new property asserts the ceiling and that grants never rise with rank.

## Nothing left behind

- [x] Filed: #3138 (name-rung prose ranking: no stemming + symbol-count bias, with a measured repro), #3139 (schema description promises meaning-matching an embedder-less host can't deliver — decide bench embedder config or registration-time honesty), #3140 (budget bounds an estimate, not the rendered answer), #3141 (panicked `spawn_blocking` reported as "cancelled"), #3142 (callers facet conflates same-named symbols across languages)

## Ground-rule check

- [x] No I/O added to `stella-core` (the promotion pass is pure arithmetic); no new deps

- **Benchmark caveat, stated out loud:** without a configured embedder (`STELLA_EMBED_URL`/`STELLA_EMBED_MODEL`), the semantic rung never runs and prose queries fall to term matching, which mis-ranks (#3138's measured repro). Symbol- and path-shaped queries are strong on every rung, and the exact rung is now certain. If the benchmark host can carry an embedding endpoint, it should — that is the rung this tool was built around. #3139 tracks the honesty half.
- The depth dial's meaning shifts from "what renders" to "the ceiling; the budget governs" — `STELLA_SEARCH_DEPTH` still works and a sweep is still monotone (the depth-superset test now covers all hits, not just the top one).
- Live end-to-end runs on this repo (name rung, no embedder): exact-symbol query pins the definition with full context block; byte-identical output across repeated runs (invariant 7 checked with `diff`).

---

https://claude.ai/code/session_01SLimjR9LBtCmFjCQ8v9aMn

---
_Generated by [Claude Code](https://claude.ai/code/session_01SLimjR9LBtCmFjCQ8v9aMn)_

## Summary by Sourcery

Harden the code search behavior for benchmark and agent use, and expose the search engine as a new `stella search` CLI command sharing the same engine as the agent tool.

New Features:
- Add `stella search` CLI subcommand that runs the same search engine used by the agent tool against the local workspace.
- Introduce an exact symbol-name strategy that pins defining files above semantic and name-ranked results when the query matches an indexed symbol.

Enhancements:
- Adjust search allocation to use remaining budget to deepen hit detail back toward the requested depth, with default depth now set to the maximum so budget becomes the main constraint.
- Enrich search results so detailed facets are led by the matched symbol, callers and callees render precise citation labels and drop cross-file noise, and doc-comment walks skip Rust attributes.
- Make semantic ranking and file scans disclose rank ceiling, partial coverage, capped walks, and cut result lists via notes so truncations and limits are always explicit.

Documentation:
- Add `stella search` command documentation page and link it from the commands index and CLI help, describing strategies, dials, and offline behavior.
- Update the `search` agent tool documentation to reflect the new default depth, budget promotion, and CLI availability.

Tests:
- Add end-to-end and unit tests covering default answer richness, exact-name vs semantic ranking, budget promotion behavior, capped scan disclosure and ordering, cut list and ceiling notes, focus-led facets, caller citations, and doc-comment attribute skipping.
- Update property-based tests for allocation to assert depth ceilings and non-increasing depth by rank after promotion.


macanderson:
Merged `main` in and resolved the conflicts (`db0d8e14`). This branch was 14
commits behind, and in that window two of its changes landed on `main` by a
different route, so the resolution needed judgment rather than a textual pick.
Flagging every non-mechanical call here so review can check them.

### Duplicated → resolved to `main`'s implementation

| this branch | `main` | resolution |
|---|---|---|
| `Strategy::ExactName`, `exact_hits`, `pin_first` | `Strategy::ExactSymbol`, `exact_symbol_hits`, `lead_with` (#3153) | **main's** — same feature, and main's names are already referenced by its merged tests |
| `stella search <query>` (28 lines, text only) | `stella search "<q>" [--format text\|json]` with a versioned envelope + tests (#3131) | **main's** — a strict superset; `search_cmd.rs` is main's entire |
| `run_search` + blank-query guard in `execute` | `report()`/`SearchReport` + `QUERY_REQUIRED` | **main's** — both moved the refusal into the shared engine so the two doors cannot drift; main's seam already does it |

`Hit::focus` is **kept** and now rides on the exact hits too, so the detailed
facets describe the definition rather than whatever sits first in the file.
That half was genuinely new here.

### Combined rather than chosen — `coverage_note`

Both sides improved this function in complementary directions, so the merged
version keeps **both**:

- `main` (#3124) fixed a comparison that could never be satisfied — chunk rows
  dedup on the rendered-text hash, so `chunks >= symbols` is unreachable for any
  file containing two byte-identical bodies.
- this branch fixed the *failure direction* of an unreadable counter, where
  `.ok()?` made a read error indistinguishable from full coverage.

The merged function uses main's exact per-file predicate **and** this branch's
fail-toward-disclosure, because the two hazards are one hazard: the answer
misrepresenting what it saw.

### One real decision, not a textual one

`main` returns early on an exact hit; this branch pinned exact hits in front of
the **name-match** rung. With the early return present, that pinning is
unreachable — so it is removed rather than left as dead code that reads like
behaviour. Called out because it is the one place the merged control flow
matches neither side verbatim.

### Dropped

`stella graph` and `docs/commands/graph.mdx`. `main` removed the subcommand;
this branch predates that and only looked like it was re-adding it because its
`Search` variant landed adjacent to it in `cli.rs`. `check-command-docs`
confirms all 38 remaining subcommands still have a reference page.

### Carried through unchanged

The allocator promotion pass and `Depth::DEFAULT = MAX`, every truncation
disclosure (scan cap, cut match lists, rank ceiling), `merge_notes`, and the
enrichment fixes (focus, caller citation labels, callee filtering, doc-comment
attribute skip). The bare-symbol recall probe is carried too, rewritten into
main's `any_of` probe shape since main generalized that harness to grade any
corpus.

One doc fix that follows from this branch's own change: `docs/commands/search.mdx`
showed `at depth 4` in its JSON example, which is stale now that
`Depth::DEFAULT` is `MAX`.

### Verified on the merged tree

- `cargo test --workspace` — green (the 29 search tests are both suites' tests together)
- `cargo clippy --workspace --all-targets -- -D warnings` — green
- `cargo fmt --all --check` — green
- `make guards-fast` — exit 0

macanderson:
We must not forget that this tool needs to also surface context records which are toml files in the repo that probably also should be embedded and indexed along with everything else. I don't know why we would discriminate this search and leave anything unsearchable. This tool must show you the most relevant information first I don't care what category it comes from embeddings regex whatever it is the only place agents can go to retrieve things. And for that reason we can save model round-trips if we find a way to cache context gathering information in a way that is grouped by probably domain I am thinking. We need a way to invalidate and we can use git for that because the searched file would have a commit sha associated for the last time it was changed. And you can also get the new Sha and if they don't match you don't use the cache you flush it.

macanderson:
Both halves are now tracked as handoffs rather than living only in this thread:

- **#3162 — nothing unsearchable.** Confirmed the gap is real on every rung, in two places at once: `stella-graph`'s `Language::from_path` has no `toml` arm (`crates/stella-graph/src/lang.rs`), so no TOML is ever indexed or embedded; and the scan rung's `SKIPPED_DIRS` excludes `.stella/` wholesale plus all dotted entries (`crates/stella-tools/src/search/scan.rs`), so `.stella/rules/*.toml` context records are invisible even to the grep-of-last-resort. The issue carries both file pointers, the `.stella/private/` carve-out constraint, and the record-spec chunking question.
- **#3163 — domain-grouped context cache, git-sha invalidation.** Written up with the design questions that need answering before code: per-file vs. whole-answer granularity, `.stella/domains.toml` as the grouping key, batched sha lookup vs. per-file subprocess, dirty-file bypass, and the invariant-7 requirement that a cache hit render byte-identically to the miss that filled it.

Kept out of this PR deliberately — both change what search *covers*, this PR changes whether search keeps its own promises. One logical change per PR.

Also pushed to this branch since your merge: `91f8b5a`-rebased-as-`HEAD` fixes the vade finding above — the misconfigured-embedder arm was assigning its note over whatever disclosure `dispatch` had set (a capped scan or cut list read as complete the moment the embedder was also half-configured). Notes now compose through `merge_notes`, with a witness that drives the degraded arm through a new env-free `report_with` seam and was checked both ways (fails on the overwrite, passes on the merge).

Issue #3168, opened 2026-08-13
make doc-warnings is red on main: stella-graph public docs link to the private NEEDS_CHUNK_PASS


macanderson:
## Problem

`make doc-warnings` (a `make gate` step, `RUSTDOCFLAGS="-D warnings"`) fails on
`main`:

```
error: public documentation for `pending_chunks` links to private item `NEEDS_CHUNK_PASS`
  = note: `-D rustdoc::private-intra-doc-links` implied by `-D warnings`
error: public documentation for `pending_chunk_file_count` links to private item `NEEDS_CHUNK_PASS`
error: could not document `stella-graph`
```

`crates/stella-graph/src/vectors/chunks.rs:227` declares
`const NEEDS_CHUNK_PASS: &str = …` (private), and two **public** functions cite
it with an intra-doc link:

- `:235` — `/// "Not fully embedded" is [`NEEDS_CHUNK_PASS`] — a coverage marker read …`
  (on `pending_chunks`)
- `:513` — `/// same [`NEEDS_CHUNK_PASS`] predicate [`pending_chunks`] scans, as a count …`
  (on `pending_chunk_file_count`)

Landed with #3152 (`ebe812925`). Found from an unrelated branch — the file is
byte-identical to `main` there (`git diff origin/main -- crates/stella-graph`
is empty), so the failure is inherited, not caused by that work.

Note this is red on `main`, which per the repo's own red-main dynamics makes
`make gate` red for every branch that runs the full rung until it is fixed.

## Repro

```sh
git checkout main
make doc-warnings          # or: RUSTDOCFLAGS="-D warnings" cargo doc -p stella-graph --no-deps
```

Note `--document-private-items` is passed by the `Makefile` target and does
**not** suppress this lint — the lint is about a *public* item's docs pointing
at a private one, which is a broken link for anyone reading the published docs.

## Fix options (pick one, in preference order)

1. **Make the const `pub(crate)`** and see whether the lint clears — it will
   not, since the citing items are `pub`; so realistically:
2. **Make the const `pub`** if the predicate is genuinely part of the module's
   published vocabulary (it is a SQL fragment two public functions promise to
   share, which is an argument that it is).
3. **Drop the link, keep the name** — change `` [`NEEDS_CHUNK_PASS`] `` to
   `` `NEEDS_CHUNK_PASS` `` in both doc comments. Cheapest, and loses nothing
   for an external reader who cannot navigate to a private item anyway.

[...]

Issue #3167, opened 2026-08-13
classify the remaining ToolOutput::Error sites, with a down-only unclassified ratchet


macanderson:
## Problem

#3145 (PR: `fix/3145-tool-error-class`) gave `ToolOutput::Error` an optional
`class: Option<ErrorClass>` and classified only the **mechanical tranche**:

- the registry's unknown-tool refusal → `NotFound`
  (`crates/stella-tools/src/registry.rs`),
- every `crate::input::InputError` → `InvalidInput`, via one
  `impl From<InputError> for ToolOutput` in `crates/stella-tools/src/input.rs`
  (21 call sites converted),
- `bash`'s wall-clock timeout → `Timeout` (`crates/stella-tools/src/bash.rs`),
- the four approval / extension-policy denials → `RefusedByPolicy`
  (`crates/stella-tools/src/registry/approval.rs`).

That is 32 classified constructions against **373 remaining
`ToolOutput::error(...)` sites that are still unclassified** (`class: None`).
`None` is a declared default meaning "not audited yet" — deliberately not a
class — so today an error-rate query can *exclude* the classified kinds but
cannot yet attribute the rest.

Two things are therefore still missing before the ceiling #3145 exists to
enable can actually be set:

1. **The remaining sites are unaudited**, and nothing stops a new one being
   added unclassified.
2. **No ratchet**, so the unclassified count can drift upward silently.

### 1. Classify the remaining sites

Enumerate with:

```sh
rg -n 'ToolOutput::error\(' crates/ | wc -l     # 373 at the time of filing
rg -n 'ToolOutput::error\(' crates/stella-tools/src
```

The house rule from #3145's `ErrorClass` doc comment is *whose problem is
this*: the model's (`InvalidInput`, `NotFound`), the policy plane's
(`PermissionDenied`, `RefusedByPolicy`), the world's (`Timeout`,
`Environment`), or ours (`Internal`). `Internal` is the class an error-rate
ceiling is actually about, so it is the one that must not be applied loosely.

Hard constraint carried over from #3145: **the refusal message bytes must not
change.** The loop detector and the prompt cache both compare those strings; a
classification PR that rewords a message is a behavior change wearing a
refactor's clothes. Add the class beside the prose, never in it.

Suggested slicing (one PR each, so review stays possible):

- `stella-tools` file tools (`read.rs`, `write.rs`, `edit.rs`,
[...]

Issue #3149, opened 2026-08-13
diag: agent.tool.result carries no tool name and MCP tools collapse to "other" — a failure cannot be attributed from the diag plane


macanderson:
## Problem

The diag plane cannot attribute a tool failure to a tool.

- `agent.tool.result` (`crates/stella-cli/src/diag_bridge.rs:365-382`) carries `seq`, `ok`, `duration_ms`, `speculated`, `output_bytes` — **no tool name**. Attribution requires joining against the matching `agent.tool.call` by `seq`, which a line-oriented log consumer (or a human tailing stderr at the default warn filter, where *only* the failing result appears) cannot do — the call line was emitted at debug and filtered out.
- `agent.tool.call.tool` is the closed `ToolName` vocabulary (`diag_bridge.rs:63-89`): 13 named tools plus `other`. **Every MCP tool, script tool, foundry tool, and hallucinated name collapses to `"other"`**, so per-tool attribution for the whole MCP surface is unrecoverable from diagnostics even with the join.

This is by design — `stella-diag` field values cannot hold a `String` (a compile-time rule, `crates/stella-diag/README.md`, `docs/spec/diagnostics.md`), which is the right privacy/egress posture. But the current shape means the one place a failure is visible at warn level names everything except which tool failed.

## Fix shape (respecting the no-String rule)

- Add the same closed `tool` field to `agent.tool.result` that `agent.tool.call` already carries — the join disappears for the 13 named tools, at zero privacy cost (`ToolName` is a closed enum, already an allowed field type).
- For the MCP/custom surface, decide and declare: either an interned/registered identifier mechanism consistent with the diag plane's typing rules, or a documented statement in `docs/reference/diagnostics.md` that per-tool attribution beyond the closed set lives in `store.db::tool_calls` (which records the raw name) — a declared gap, not a silence.

## Definition of done

- Witness: a failing `read_file` produces a single warn-level `agent.tool.result` line from which the tool is identifiable without any join (fails on main: the field is absent).
- `docs/reference/diagnostics.md:341-369` updated to match.
- The MCP-attribution decision recorded (mechanism or declared pointer to `tool_calls`).

Refs AGENTS.md invariant #10; sibling issues on error taxonomy and the reliability surface.

Issue #3144, opened 2026-08-13
tools: no dispatch-time input validation — a mistyped argument is misreported as missing or silently defaulted


macanderson:
## Problem

`ToolCall.input`'s doc comment promises validation that nothing performs. `crates/stella-protocol/src/tool.rs:55-56` says *"Runtime data: validate against `ToolSchema::input_schema` rather than trusting the shape"* — but no code in the workspace validates a tool call's input against the advertised schema. No tool deserializes its args into a struct; every tool reads fields off `&serde_json::Value` by hand; no JSON-Schema validation library is a dependency. `"additionalProperties": false` is advertised by `task` (`crates/stella-tools/src/subagent.rs:247`) and foundry-authored tools (`crates/stella-tools/src/foundry_author.rs:259`) and enforced nowhere. Unknown parameters are silently ignored by every tool.

The consequence is worse than "lenient": a **mistyped argument is misreported as a missing one, or silently defaulted**, so the model retries against a refusal that names the wrong defect. Four witnessed cases (all verified against the code 2026-08-12):

1. `crates/stella-tools/src/search.rs:344-351` — `{"query": 42}` → `as_str()` → `unwrap_or_default()` → the blank-query refusal `QUERY_REQUIRED` (`search.rs:358`) says *"query is required"*. The field was present, just mistyped, and the schema declares `"required": ["query"]` with `"type": "string"` (`search.rs:327-334`).
2. `crates/stella-tools/src/verify.rs:555-570` — `test_files` passed as a string (not array), or an array of non-strings, is silently emptied by `filter_map(as_str)` and then refused as *"missing required field `test_files`"*.
3. `crates/stella-tools/src/read.rs:271-281` — `offset`/`limit` of the wrong type silently vanish into defaults (`limit: "200"` reads the default window; no refusal, no note).
4. `crates/stella-tools/src/bash.rs:380-409` — `trace` of the wrong type silently defaults to `false`.

A typed `InputError` with exactly the right variants already exists — `Missing { field }`, `WrongType { field, want, got }`, `WrongElementType { field, index, got }` at `crates/stella-tools/src/input.rs:83-114` — but tools that predate it never migrated, and it is stringified at every call site.

## Fix shape

One validation seam at dispatch — `ToolRegistry::execute` (the unknown-tool arm is `crates/stella-tools/src/registry.rs:1143-1148`) — that checks the input against the advertised `input_schema` before the tool runs:

- declared `required` fields present;
[...]

PR #3169, opened 2026-08-13
feat(stella-protocol): a tool error names its class

### LINKED ISSUE #3145: protocol: ToolOutput::Error is prose-only — no error class, so a per-tool error-rate ceiling cannot tell defect from misuse
## Problem

`ToolOutput::Error { message: String }` (`crates/stella-protocol/src/tool.rs:72-76`) carries no class, code, or kind — failure is one prose string written for the model to retry against. `store.db`'s `tool_calls.error` (`crates/stella-store/src/ddl.rs:375-397`) stores that prose verbatim.

So a per-tool error *rate* is computable today (and the Observatory computes it), but a per-tool error rate **ceiling** is unenforceable, because the number cannot distinguish:

- (a) a tool defect (our bug — counts against the ceiling),
- (b) invalid input (model misuse — counts against the prompt/menu, not the tool),
- (c) an environment/policy refusal (containment, approval-denied — working as designed),
- (d) an abandoned call (turn ended first — not an error at all; see the sibling issue on ABANDONED).

The cost of not having this is already visible in the tree: `crates/stella-tools/src/rootfd.rs:459` and `:877` cite *"the largest single class of tool error in a benchmark cycle (8 of 49)"* — someone classified tool errors **by hand** off a bench run because no machinery can. And `bench/trace_triage/banned.py:264-308` ships four tool-shaped detectors as ReportOnly (#3053) partly because error envelopes carry no verifiable class.

## Fix shape

- A small closed `ErrorClass` enum in `stella-protocol` (candidates: `invalid_input`, `not_found`, `permission`, `refused_by_policy`, `timeout`, `environment`, `internal`, `abandoned`), carried as an optional field on `ToolOutput::Error` with `#[serde(default)]` so old payloads round-trip (invariant #4: add the round-trip test).
- Every `ToolOutput::Error` construction site in `stella-tools` classified; unclassified defaults to a named `unclassified` value, not a silent absence — invariant #10's discipline (declared gap, not silence). A down-only ratchet on `unclassified` count fits the house style (`scripts/typed-errors-baseline.txt` is the pattern).
- `tool_calls` gains an `error_class` column (schema bump; the v18 live-write path in `crates/stella-store/src/tool_calls.rs` is the writer), and `tool_usage_rollup` gains per-class counts or at minimum an `errors_invalid_input` split.

## Constraints

- Wire-schema gate: `docs/wire/` regeneration is part of the PR (`wire-schema` gate step).
- The class must not perturb the message bytes the model sees (prompt-cache/loop-detector stability).
- This is deliberately **not** #2716's governance machinery (closed not-planned) — no RiskLevel, no AuthzGate. It is the measurement half only.

## Definition of done

- Round-trip tests for the new field; old-payload decode test.
- A query over `tool_calls` can answer "error rate for `bash`, excluding invalid_input and abandoned" with no string matching.
- `bench/trace_triage`'s `tool-error-envelope` detector can read the class (unlocks the #3053 verification path).

Refs #3053, #2716, AGENTS.md invariants #4, #5, #10.

macanderson:
`ToolOutput::Error { message }` carried no class, so a per-tool error *rate*
was computable but a *ceiling* was not: the number could not tell a tool defect
from model misuse from a policy refusal. This adds the classification
structurally — as a field, never as a convention over the prose.

Closes #3145

## The shape

**`ErrorClass` in `stella-protocol`** (`crates/stella-protocol/src/tool.rs`) —
a closed `snake_case` set partitioning failures by *whose problem they are*:
the model's (`InvalidInput`, `NotFound`), the policy plane's
(`PermissionDenied`, `RefusedByPolicy`), the world's (`Timeout`,
`Environment`), or ours (`Internal`), plus a `#[serde(other)] Other`
forward-compat arm. Exemplar for the forward-compat discipline is this crate's
own `receipt::BlockKind` / `CacheZone`, named in AGENTS.md invariant #4's
neighborhood: an unknown token from a newer emitter degrades to `Other` rather
than failing the whole event.

**No `Abandoned` variant.** A call whose turn ended before it returned never
produced a `ToolOutput` to classify — that is a store-side lifecycle fact
(#3157's v24 `abandoned` state), not an error a tool reports. The repair fold
here writes `error_class = ''` for an abandoned call for exactly that reason.

**`class: Option<ErrorClass>` on the `Error` arm**, `#[serde(default,
skip_serializing_if = "Option::is_none")]`. `None` is a *declared* default
meaning **unclassified** — not a class — so an error-rate ceiling never reads
our unfinished audit as our defects.

**Constructors**: `ToolOutput::error(msg)` (unclassified) and
`ToolOutput::classified_error(class, msg)`, so ~370 call sites migrate
incrementally instead of in one unreviewable diff.

**`tool_calls.error_class`** — schema **v23 → v24**, additive column-guarded
`ALTER TABLE ADD COLUMN … NOT NULL DEFAULT ''`. The migration lives in
`crates/stella-store/src/migrations/error_class.rs` (the `token_unit`
precedent) because `migrations.rs` is close under the 1500-line ratchet. Both
writers project it: the live fold (`project_event` → `project_tool_result`)
and the repair fold (`materialize_tool_calls`). `''` is unclassified; no
backfill, because an old failure's class exists nowhere but its prose and
classifying by string match is the practice this column replaces.

## Refusal message bytes are unchanged

Every classified site keeps its exact wording — the class rides *beside* the
prose. `InputError`'s `Display` is untouched and the conversion goes through
one `impl From<InputError> for ToolOutput`, so the loop detector and the
prompt cache see the same strings they saw before.

## Witnesses (each fails on `main`)

| Test | Crate | Fails on main because |
|---|---|---|
| `tool::tests::unclassified_error_serializes_byte_identically_to_the_pre_class_shape` | `stella-protocol` | pins `{"error":{"message":"boom"}}` — proves the added field is absent when `None` (invariant #4). Does not compile on main: no `ToolOutput::error`. |
| `tool::tests::old_error_payload_without_class_decodes_as_unclassified` | `stella-protocol` | old-payload decode; no `class` concept on main. |
| `tool::tests::unknown_error_class_token_degrades_to_other_not_a_decode_failure` | `stella-protocol` | `ErrorClass::Other` does not exist on main. |
| `tool::tests::error_class_token_and_serde_spelling_agree` | `stella-protocol` | pins `as_str()`/`parse()` against the serde token so the store column and the wire cannot diverge. |
| `tool_calls::tests::a_classified_error_lands_its_class_in_the_projection` | `stella-store` | on main there is no `error_class` column, no `ToolCallRow::error_class`, and `record_event` has nothing to project. Asserts the live writer AND the repair fold agree, and that an unaudited site stays `''`. |
| `registry::tests::error_class::unknown_tool_refusal_is_classified_not_found` | `stella-tools` | the refusal carries no class on main. Also asserts the message still starts with the original bytes. |
| `read::tests::missing_path_is_classified_invalid_input` | `stella-tools` | `read_file`'s missing-`path` refusal is unclassified on main; asserts the message is still exactly `missing required field \`path\``. |

Checked the artisanal way — every one of these fails to compile or fails to
assert against `origin/main` (`8bf525765`), because each names a type, field,
or column that branch does not have.

## The classified tranche (32 sites)

Deliberately only the mechanical ones:

- `crates/stella-tools/src/registry.rs` — unknown-tool refusal → `NotFound`
- `crates/stella-tools/src/input.rs` — one `impl From<InputError> for
  ToolOutput` → `InvalidInput`, adopted at 21 call sites across `read`,
  `write`, `edit`, `glob`, `grep`, `delete`, `memory`, `process`, `bash`,
  `issues`, `exploration`
- `crates/stella-tools/src/bash.rs` — the wall-clock timeout → `Timeout`
- `crates/stella-tools/src/registry/approval.rs` — the four approval /
  extension-policy denials → `RefusedByPolicy`

**373 sites remain unclassified**, on purpose: they need per-site judgement,
not a sweep. Tracked as a handoff in **#3167** ("classify the remaining
`ToolOutput::Error` sites, with a down-only unclassified ratchet"), which also
carries the two pieces this PR explicitly did not do — the per-class
`tool_usage_rollup` / usage-hub split (the hub is guarded by the content-free
allowlist, so it needs a human to answer "is this content?" in the same PR),
and the Observatory's error-rate split.

## Ripple

~370 pattern sites became `ToolOutput::Error { message, .. }` and ~360
constructions became `ToolOutput::error(x)`, across every crate that links the
protocol. Compiler-driven (`cargo check --message-format=json`), so the set is
exhaustive by construction rather than by grep. `arenabench/toolout.py` was
read and left alone: it decodes the `error` arm by testing for the `message`
key and ignores unknown keys, so a present-or-absent `class` is
backward-compatible there.

Two files were kept under their ratchets rather than pushed over:
`crates/stella-tools/src/registry.rs` (recompacted two match arms rustfmt had
expanded) and `crates/stella-store/src/tests.rs` (the v-history comment
compressed into an index, net −1 line). The new `stella-tools` witness went
into a new `registry/tests/error_class.rs` submodule rather than growing
`registry/tests.rs` past 1500.

## Migration slot

Takes **v23 → v24**. #3157 is still open at the time of writing and claims the
same slot; whoever lands second renumbers — a trivial, expected conflict
(`MIGRATIONS`'s reserved-slot comment block documents exactly this hazard). If
#3157 lands first this becomes v24 → v25 and the `assert_eq!(SCHEMA_VERSION,
24)` in `crates/stella-store/src/tests.rs` moves with it.

`docs/wire/` regenerated with `make wire-schema-update`; `ErrorClass` is
sampled on every arm in `crates/stella-protocol/tests/wire_contract/`, so the
wire-contract test proves each token validates against the committed schema.

## Gate

`cargo test --workspace` green (140 test binaries, 0 failures), `cargo clippy
--workspace --all-targets -- -D warnings` clean, `cargo fmt --all` applied,
`make guards` green (file-size, god-files, gate-parity, invariants,
wire-schema, left-behind, …).

One inherited failure, **not caused by this branch**: `make doc-warnings` is
red on `main` — `stella-graph`'s public `pending_chunks` /
`pending_chunk_file_count` docs intra-doc-link the private `NEEDS_CHUNK_PASS`
(landed in #3152). `git diff origin/main -- crates/stella-graph` is empty on
this branch, so the file is byte-identical to main's. Filed as **#3168** rather
than folded in here, per the rule against fixing an unrelated red ratchet
inside someone else's PR.

## Deleted tests

None.

## Summary by Sourcery

Introduce structured error classification for tool failures and persist it through the protocol, storage, and wire schemas, while keeping existing error messages byte-compatible.

New Features:
- Add ErrorClass enum and optional class field to ToolOutput::Error, with helpers and wire schema support for machine-readable error categorization.
- Expose error_class on stored tool call rows and readers, enabling per-class error analysis and projections in stella-store.

Enhancements:
- Provide migration from schema v23 to v24 adding the error_class column to tool_calls with safe defaults.
- Refactor ToolOutput construction across tools to use new helpers, and selectively classify common failures such as invalid input, not found, timeouts, and policy denials.
- Ensure forward- and backward-compatible serde behavior for error_class, including handling unknown tokens and legacy payloads.
- Extend tests and wire-contract samples to cover ErrorClass values, storage projection, and classification correctness across crates.

Documentation:
- Regenerate and update wire type definitions to document ErrorClass and the optional class field on ToolOutput::Error.

Tests:
- Add protocol, store, and tools tests validating serialization, migration behavior, classification, and unchanged error messages.
- Update existing tests to account for the new ToolOutput::Error shape and error_class semantics.


macanderson:
Conflicts with `main` resolved (merge commit `e3f68509f` + semantic follow-up `d4116103f`):

- **The predicted slot collision happened**: #3157 landed first and took v23 → v24 for the abandoned-call state, so this PR's `error_class` migration renumbered to **v24 → v25** exactly as the description's "Migration slot" section anticipated — `migrate_v24_to_v25`, `SCHEMA_VERSION` 25, the `assert_eq!` moved with it.
- **The repair fold now combines both PRs**: an unreturned call settles as `ToolCallState::Abandoned` (from #3157) with an unclassified (`None`) error class (this PR's stance that abandonment has no `ToolOutput` to classify).
- **`project_tool_result` hit clippy's 8-argument limit** once #3157's `state` met this branch's `error_class` — resolved structurally with a `SettledOutcome` struct built by one constructor both the live fold and the repair fold share, so the two writers can no longer read a `ToolOutput` differently.
- **Six `ToolOutput::Error` literals main added since the fork** (from #3144's validation work, #3149's diag bridge, and the observatory conformance test) migrated to the branch's constructors; the five `InputError` sites take `From<InputError>` and are classified `invalid_input` with the rest of that tranche. Seven main-added patterns gained `..`.

Verified locally: `cargo test --workspace` has exactly **one** failure, `no_subcommand_flag_reuses_a_global_name` — inherited from `main`, red there since #3159 (`usage report --tools` vs the global `--tools`). Fix is up as #3181 / issue #3180; this PR's required check stays red until that lands. `cargo clippy --workspace --all-targets -- -D warnings` clean, `cargo fmt --check` clean, `make guards-fast` green.

Issue #3147, opened 2026-08-13
usage: per-tool error rate is written on every turn and read by nothing — no scriptable surface for a reliability ceiling


macanderson:
## Problem

Per-tool reliability data is written on every turn and readable by no script.

- `~/.stella/usage.db` → `tool_usage_rollup(project_id, tool, surface, day, calls, errors)` (`crates/stella-store/src/usage.rs:147-155`) is **literally a per-tool per-day error-rate table**, upserted on every turn (`usage.rs:322-331`).
- Its `errors` column is **write-only**: the sole reader, `UsageStore::tool_totals()` (`usage.rs:346-358`), selects `SUM(calls)` only and has zero production callers (tests only).
- `stella usage report` (`crates/stella-cli/src/usage_cmd.rs:150-200`) groups by `(org, provider, model)` — no tool dimension, despite the rollup sitting in the same file.
- `stella stats` has no tool dimension either (`cli.rs:979-992`; subcommands are `Prune` and `Graph`).
- The only surface that shows per-tool errors at all is the Observatory browser dashboard (`/api/tools`), which is loopback-only, windowed to the newest 10,000 calls (`crates/stella-observatory/src/db.rs:79`), and not scriptable.

Consequence: a tool-reliability ceiling ("no shipped tool errors more than 3% of the time") has nothing to read. It cannot run in CI, cannot run in a bench postmortem, cannot be quoted in a receipt.

## Fix shape

A scriptable CLI surface reading the hub rollup, e.g. `stella usage report --tools` (a flag *scopes* the report — it does not select a new operation, so invariant #9 is satisfied; a `stella usage tools` subcommand is the alternative if the report shape diverges too far):

- Columns: tool, surface, calls, errors, error-rate; per-day available via the existing `day` key; `--format json` for machine consumption (the house pattern from `stella search --format json`, #3131).
- Honesty prerequisites: abandoned calls excluded (sibling issue) and ideally error classes split (sibling issue) — ship the surface first with a documented caveat if those land later, but name the caveat in the output, not silently.
- `make command-docs` requires the doc row for any new CLI surface.

## Definition of done

- `stella usage report --tools --format json` emits per-tool `{tool, surface, calls, errors, error_rate}` from `tool_usage_rollup`, cross-project.
- A witness test that seeds the rollup and asserts the output, including that `errors` is actually read (fails on main by construction: nothing reads it today).
- The dead `tool_totals()` either becomes the implementation or is removed (it is currently misleading scaffolding).

[...]

PR #3171, opened 2026-08-13
fix(stella-cli): usage report --tools collided with the session-wide --tools global — rename to --by-tool

macanderson:
## What — unbreak main

`stella usage report --tools` (merged in #3159) reuses the name of the **session-wide global** `--tools '<SPEC>'` (the tool-switch flag, `crates/stella-cli/src/cli.rs`), which the `no_subcommand_flag_reuses_a_global_name` guard test rejects. Renamed to `--by-tool`; refusal message and `website/content/docs/commands/usage.mdx` follow. No behavior change.

## Why it reached main

The guard's failure was **masked** on #3159's own CI: the `fmt + clippy + test` job was already failing at the earlier `cargo doc -D warnings` step on the inherited `stella-graph` doc-links red (fixed since by #3166), so the test step's own failure never showed as the headline. The one-red-gate-step-masks-the-next trap, exactly as documented — #3159 merged while the job was red for the "known" reason, and the collision surfaced on the first branch to run full CI against the repaired base (#3169's log names this test).

Main is currently red on `cargo test -p stella-cli` because of this; every open PR inherits it.

## Verification

`cargo test -p stella-cli no_subcommand_flag_reuses_a_global_name` — passes (fails on main: the collision is live there); `cargo test -p stella-cli usage` — 20 passed; `scripts/check-command-docs.sh` OK; fmt clean.

Refs #3147, #3159

## Summary by Sourcery

Rename the stella usage report CLI flag from --tools to --by-tool to avoid collision with the session-wide tool-switch global, updating messages and docs accordingly.

Bug Fixes:
- Resolve the naming collision between the usage report --tools flag and the global --tools tool-switch flag.

Documentation:
- Update usage command documentation and examples to use the new --by-tool flag and clarify its relationship to the global --tools option.

### DIFF
--- a/crates/stella-cli/src/usage_cmd.rs
+++ b/crates/stella-cli/src/usage_cmd.rs
@@ -37,9 +37,10 @@ pub enum UsageCmd {
         org: Option<String>,
 
         /// Per-tool reliability instead: cross-project calls, errors, and
-        /// error rate per (tool, surface), from the hub's rollup
+        /// error rate per (tool, surface), from the hub's rollup. (Named
+        /// --by-tool because --tools is the session-wide tool-switch global)
         #[arg(long)]
-        tools: bool,
+        by_tool: bool,
     },
     /// Replicate this workspace's telemetry into the hub (cursor-based;
     /// safe to re-run). --all heals every project the hub knows about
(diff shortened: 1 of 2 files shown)


macanderson:
Superseded: the identical rename (--by-tool, same two files) merged as #3178 while this was open — the two sessions raced on the same red-main fix. Closing; no content here is lost, main already carries it.

Issue #3215, opened 2026-08-13
arenabench: wire the experiments store into the server and UI, and decide the normalization step


macanderson:
## Problem

The experiments store landed as Phase 1 of the Experiments product (branch `worktree-experiments-report`): `arenabench/arenabench/experiments.py` persists whole experiment documents into `~/.arenabench/experiments.db` (`experiment_results`, one non-null JSONB column named `results` — deliberately unnormalized), and the first canonical document (`stella-vs-claude-code-fable5-tb21`) is stored. Nothing serves it yet: unwired code with no tracking issue is the failure mode the repo rule exists to prevent, so this is the wiring handoff.

## What to build

1. `GET /api/experiments` — list stored documents (id, title, status pulled via `json_extract`); `GET /api/experiments/<id>` — the full document. Follow the `/api/series` shape: thin handler on `ArenaServer` (server.py:496-503), one `case` arm in `_route_api_get` (server.py:1058), loopback/Origin rules per server.py:36-48. The read path must tolerate a missing DB — matches-on-disk discovery (`runner.py:707`) must never depend on the new store.
2. A UI view: new `View` member in `ui/components/app-shell.tsx:12` + `components/experiments/experiments-view.tsx` fetching via `api<T>()` (`ui/lib/api.ts`), same as `trends-view.tsx`.
3. The normalization decision: whether/when to promote fields (experiment id, status, created_at, comparability keys) from the JSON into columns, made against the real stored documents — with an additive-only `migrate()` per `bench/telemetry_store/ingest.py:201`.

## Constraints

- ArenaBench is stdlib-only by charter (`pyproject.toml` `dependencies = []`); `sqlite3` is fine, an ORM is not.
- Generic product surface: agents, models, and datasets are data inside documents, never concepts the store knows.

## Done

The gallery lists stored experiments, one click opens a document, tests cover the routes, and the normalization question has a written answer (even if the answer is "not yet").

Refs #2889.

macanderson:
The API half shipped as PR #3221 (on top of the merged #3220): `GET /api/experiments` (gallery listing) and `GET /api/experiments/<id>` (full document, 404 on miss), with the missing-DB read guard — reading never creates the database, and match discovery stays independent of the store. Three witness tests added.

Remaining here: the UI view (`View` member in `app-shell.tsx` + `experiments-view.tsx`) and the normalization decision.

PR #3221, opened 2026-08-13
feat(arenabench): serve stored experiments — GET /api/experiments and /api/experiments/<id>

macanderson:
## What this is

The read half of #3215, on top of the merged Phase 1 store (#3220): the server now serves stored experiment documents.

- `GET /api/experiments` — gallery listing: each document's own `experiment` header (id, title, status) plus schema and calculation version. Whole documents deliberately stay out of the listing — they grow with every trial they record.
- `GET /api/experiments/<id>` — the full document; unknown id → 404 via the existing `KeyError` convention.

Follows the `/api/series` shape exactly: thin handlers on `ArenaServer`, one `case` arm each in `_route_api_get`. Two properties held deliberately:

- **Reading never creates the database** — `load_results` now returns `[]` for a missing file instead of connecting (a GET must not write), and match discovery (`restore_from_disk`) remains fully independent of the store.
- **Latest row wins on duplicate ids** — a re-run of a calculation supersedes its predecessor, in insertion order.

## Witness

Three new tests in `arenabench/tests/test_experiments.py` fail on main (functions absent) and pass here: missing-DB read creates nothing, payload summarization, latest-wins lookup + miss → `None`. `python3 -m pytest tests/test_experiments.py -q` → 11 passed; ruff clean on all three touched files. Live smoke against the real `~/.arenabench/experiments.db` returns the stored `stella-vs-claude-code-fable5-tb21` document.

## Scope

Refs #3215 — deliberately not `Closes`: the UI view and the normalization decision remain there.

No new dependencies.

## Summary by Sourcery

Expose stored experiment documents via new GET /api/experiments and /api/experiments/<id> endpoints backed by the experiments store.

New Features:
- Add experiments_payload helper to summarize stored experiment documents for listing surfaces.
- Add experiment_document lookup helper to retrieve the full stored experiment document by id.
- Expose summarized experiments listing and single experiment retrieval on ArenaServer via new experiments and experiment methods, wired into the API router.

Bug Fixes:
- Ensure load_results returns an empty list and does not create a database when the experiments store file is missing.

Tests:
- Add tests covering missing-database reads, experiment listing payload summarization, and latest-wins experiment document lookup semantics.

### DIFF
--- a/arenabench/arenabench/experiments.py
+++ b/arenabench/arenabench/experiments.py
@@ -55,7 +55,9 @@ from .sut import arenabench_home
 
 __all__ = [
     "connect",
+    "experiment_document",
     "experiments_db_path",
+    "experiments_payload",
     "load_results",
     "store_results",
 ]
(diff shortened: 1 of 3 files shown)


macanderson:
Independently verified this branch (46175e3b0) from a clean worktree, not from the description:

- **Witness holds against the real base** (`f67a90776`, the #3220 merge). All three new tests fail there: two with `AttributeError` (functions absent) and — the load-bearing one — `test_loading_a_missing_database_returns_empty_and_creates_nothing` with `assert not True`, i.e. on main the read *did* create the database. That third one witnesses a behaviour change rather than a mere absence.
- **No regression.** Full `arenabench` suite on this branch vs. main: identical failure sets, 5 in each, all `tests/test_sut.py` `AdapterUnavailableError` from `/usr/bin/python3` lacking `tomllib` — pre-existing and unrelated.
- **The "reading never creates" property holds on the default branch too**, which the tests do not reach: `experiments_db_path()` → `sut.arenabench_home()` is pure (no `mkdir`), and the `path.exists()` early return sits above `connect()`'s `parent.mkdir`, so `GET /api/experiments` on a fresh machine creates nothing.
- **404 wiring confirmed**: `do_GET` wraps `_route_api_get` in `except KeyError → HTTPStatus.NOT_FOUND` (server.py:998-1000).
- **Live smoke** against the real `~/.arenabench/experiments.db` returns the seeded `stella-vs-claude-code-fable5-tb21` document. `ruff check` clean on all three touched files.

One gap found and filed rather than folded in here, since the branch is complete and green: no test issues an HTTP request to either new route, so the two `case` arms and the `KeyError`→404 conversion are proven only at helper level — which is also #3215's "tests cover the routes" criterion. Filed as #3227 with the `test_file_api.py` fixture to copy and the `ARENABENCH_HOME` isolation trap written up.

Issue #3246, opened 2026-08-14
Design and sequencing plan for a plugin platform: two existing hook systems don't know about each other


macanderson:
Companion to #3243. That issue covers the *retrieval* plane (what context to
inject); this one covers the *authority* plane (what may change the loop). They
meet at one place — §5's sequencing — and are otherwise independent.

**Evidence basis.** `main` at `a4b87649a` (2026-08-13T14:44:24-07:00). Every
"shipped/not shipped" claim below was checked against the tree, not against the
issue that proposed it.

---

## 1. The requirement (Mac, 2026-08-13)

A plugin is more than a skill or a custom tool. It can:

- **change the definition of done** — the exit condition of the turn loop
- **block turn completion** to do synchronous work (author a witness, run it,
  read the oracle flip) before the turn may finish
- **append to the system prompt** when installed, and to the volatile part per turn
- **subscribe to lifecycle events** — before/after enter loop, before/after exit
  loop, on error/abort — where the event *cannot complete* until the plugin
  answers OK
- **fire its own events**
- **write to traces** in a form that supports training custom models from
  trace + oracle flip
- be **written in Python against an SDK**, with Rust running parts of the loop
  through it

The first customer is Oxagen's paid verification plugin.

---

## 2. The search: what's actually in the tree

| Issue | State | Actually in the tree? |
|---|---|---|
| #1400 Stella Apps — manifest, OAuth lifecycle, host API, marketplace | closed, **not planned** (frozen behind #2374) | no |
| #2716 tool-first governance — `ToolContract`, `AuthzGate`, `Principal`, `RiskLevel` | closed **NOT_PLANNED** 2026-08-12 | **no — verified zero code** |
| #2836 hook events `UserPromptSubmit`, `SubagentStart/Stop` | closed **NOT_PLANNED** 2026-08-12 | **no — verified zero code** |
| #2684 hook surface: `Stop`, `PreCompact`, `Modify`, decisions | closed, completed | **yes** |
| #1133 HookBus lifecycle emitters | closed, completed | **yes** |
| #459 HookBus resilience — observer budget, bounded forward | closed, completed | **yes** |
| #1042 TraceSink | closed, completed | **yes** (`crates/stella-cli/src/trace.rs`) |
| #2793 MCP + custom tools bypass the blocking policy chains | **open** | the hole is live |

Verification of the two `NOT_PLANNED` closes:

```
[...]

Issue #3233, opened 2026-08-13
A sub-session in interactive mode has no step-level durability, so killing it loses its entire transcript


macanderson:
## Context

A sub-session in interactive mode (`crates/stella-cli/src/subsession.rs`) is a real engine session — its own provider, tool registry, budget guard, execution row, and event lane — dispatched onto its own OS thread because the lead conversation is mid-turn. It runs the full engine step loop.

It used to build its engine through `agent::engine_config_for`, which attaches the session's `checkpoint_sink`. Because `Config` clones over one `Arc` cell (`crate::durability::SessionDurability`), the worker and the lead were two writers of one `CHECKPOINT_BLOB`: every worker step overwrote the lead's in-flight transcript, and the worker's terminal `discard` deleted the lead's resume point while the lead still needed it. That's now fixed — `agent::subsession_engine_config_for` strips the sink, matching what `stella_core::subagent` already does for dispatched children.

**Stripping the sink fixed the damage, but it leaves a sub-session with no step-level durability at all.** This issue is that follow-up.

## Problem

A sub-session's turn can't be recovered. Kill interactive mode while a `req:<n>` or `sub:<task-id>` worker is 30 steps into a delegated task, and the whole transcript is gone. The worker's file edits stay in the shared tree, but nothing records the conversation that produced them, and nothing can resume it.

The strip is deliberate and documented at `crates/stella-cli/src/agent/engine.rs::subsession_engine_config_for`: a worker lane has no durable identity of its own, so a second sink would write with nothing to read it, at the cost of serializing the whole transcript on every step (`Engine::persist_checkpoint` encodes before the sink ever sees the bytes). Giving a lane a real identity is what would make that cost worth paying.

## What done looks like

1. Each live sub-session lane binds its own `SessionDurability` under its own session id, so its turns checkpoint into their own `refs/stella/<session>/head` namespace. `crates/stella-store/src/work_journal.rs`'s module docs establish that concurrent sessions over one workspace are safe by design — separate ref namespaces, separate `GIT_INDEX_FILE`, sharing only the object store — so the lead and N workers can each hold a binding without contending.
2. A read path: something must be able to find and re-enter an interrupted lane's turn. Today `stella resume` (`crates/stella-cli/src/agent/resume.rs`) restores only the session record's own turn and knows nothing about lanes.
[...]

Issue #3232, opened 2026-08-13
fleet: worker turns are never checkpointed — the whole fan-out has no step-level durability


macanderson:
## Problem

A fleet worker's turns are never checkpointed. `stella fleet` has **attempt-level** resilience (the ledger opens an attempt row before the worker runs, and `Fleet::dispatch` is re-runnable, so a restart is the caller re-dispatching the task) but **no step-level** resilience: a worker killed at step 40 of 50 loses all 40 steps of transcript. The tree keeps whatever the worker wrote; the conversation that produced it is gone, and the re-dispatched attempt starts from an empty history against a workspace that already holds partial work.

The mechanism to fix this already exists and is wired everywhere else — the gap is purely that the fleet path never binds it.

## Why it happens

`crates/stella-cli/src/agent/engine.rs::tuned_engine_config` sets `checkpoint_sink: cfg.durability.sink()`, and `SessionDurability::sink()` answers `None` while the handle is unbound. Binding happens in exactly four production places:

- `crates/stella-cli/src/agent/presence.rs:60` (`SessionPresence::announce`) — covers `stella run` (pipeline + raw), `/goal`, and the plain REPL
- `crates/stella-cli/src/command_deck.rs:513` and `:1225` — the deck and its in-deck session switch
- `crates/stella-cli/src/agent/resume.rs:129` — `stella resume`

`crates/stella-cli/src/fleet_cmd.rs` calls none of them. Its only mention of durability is `fleet_cmd.rs:885`, which passes the never-bound handle into `crate::resume_frame::pipeline(...)`. So for every fleet worker:

- the raw step-loop path (`fleet_cmd.rs:924`, `agent::engine_config_for(&cfg)`) gets `checkpoint_sink: None`
- the pipeline path (`fleet_cmd.rs:868`, `agent::pipeline_engine_config_for(...)`) gets `checkpoint_sink: None` **and** a no-op `ResumeFrameSink`, so `FrameProgress` is never recorded either

`fleet_cmd.rs:711` also calls `agent::begin_execution(&store, "fleet", &task.prompt, &cfg, None)` — the trailing `None` is the session id, so a fleet worker has no `SessionRecord` at all.

## Repro / verify

```
rg -n "bind_session\(|SessionPresence::announce" crates/stella-cli/src --glob '!**/tests*'
```

Four hits, none in `fleet_cmd.rs`.

[...]

Issue #2716, opened 2026-08-10
tool-first: the governance half — ToolContract, AuthzGate port + Principal, ToolCtx events, contracts over the serve wire, manifest parity


macanderson:
## Why this issue exists

Epic #2694 defines the contract discipline for tool-first Stella — versioned I/O shapes, token-optimized outputs, journal discipline, discovery routing — but stops short of the **governance half**: nothing in the epic gives a tool a risk classification, a principal-aware authorization seam, an approval flow, or a channel for in-loop events. Those are required for two things Stella is committed to:

1. **Pluggable RBAC on top of Stella** — an operator (or a product embedding Stella) must be able to plug a permission system in at one seam and have it govern every tool call, without patching tools.
2. **Oxagen hosting Stella as its engine** — Oxagen's kernel (`invoke()`: IAM → billing → entitlement, input *and* output validated against the contract) must be able to authorize each remoted tool call from contract metadata. Today `stella-serve` hands the host only `{name, input}` (`crates/stella-serve/src/frame.rs:45`) plus a bare `ToolSchema` list (`crates/stella-serve/src/session.rs:54`) — the host has to maintain a side-table mapping tool names to its own capability model.

This issue is that governance half. It was scoped against a close read of oxagen-platform's capability-contract system (`packages/oxagen/src/types.ts` `CapabilityDeclaration`, `packages/oxagen/src/kernel.ts` `_invokeCore`, `packages/agent/src/runtime/materialize-tools.ts`) — both what to copy and what to deliberately not copy (recorded at the bottom so the port cannot silently re-import the defects).

## Where Stella stands today (verified 2026-08-10)

| Contract element | Status |
|---|---|
| Input schema | Advertised (`ToolSchema.input_schema`, `crates/stella-protocol/src/tool.rs:23`) as a hand-written `json!` blob per tool; **never validated** — every tool re-parses raw `Value` by hand (`crates/stella-tools/src/read.rs:262-274`). |
| Output schema | Absent. `ToolOutput` is `Ok{content: String} | Error{message: String}` (`tool.rs:65-77`). |
| Read-only bit | Present and load-bearing (`tool.rs:29`, declared in `crates/stella-tools/src/catalog.rs`, drives dispatch grouping and `ReadOnlyTools`); but MCP and custom TOML tools cannot claim it — `crates/stella-tools/src/custom.rs:142-153` hardcodes `read_only: false`. |
[...]

PR #3273, opened 2026-08-14
docs(spec): one loop, seven lanes — a design path for single-place turn-loop assembly

macanderson:
## What

A design proposal, not an implementation: `docs/spec/turn-lane-assembly.md`.

## Why

The premise that prompted it — "Stella should only have 1 loop" — turns out to be **already true**, and that inverts the diagnosis. `crates/stella-core/src/driver/drive.rs:111` is the only `loop {` over `Engine::run_step` in the workspace; a sweep for `run_step(` outside `driver/` returns one hit and it is a doc comment. Fleet does not have its own loop. It has *the* loop, assembled differently.

What is duplicated is **lane assembly**: twelve optional capability slots, bound through three unrelated vocabularies (`EngineConfig` fields, `Engine::with_*` builders, `TurnControls` published to the registry), decided by hand at seven production sites, with nothing checking the matrix.

Consequences already paid for:

- `#3232` (fleet has no step-level durability) and `#3233` (sub-sessions have none) are two cells of one table, found months apart, each by tripping over it.
- `#3242` is a third cell found as *damage* — a sub-session inheriting the lead's sink and destroying its resume point.
- `stella-core/src/subagent.rs:61` documents a structural hole in its own module doc: `Engine::with_sleeper` cannot carry `gate`/`steering`/`hooks`, so a fork drops three capabilities by signature rather than by decision.
- The pause gate must be bound **twice**, through two mechanisms, at every site that wants it.

The doc reconstructs the current lane × capability matrix by reading all seven sites. That table did not exist before; building it required tracing `bind_session` callers × three `engine_config_for*` variants × seven `with_sleeper` sites. That difficulty is the finding.

## The proposal

Four moves, each independently shippable, all reusing instruments this repo already proved (invariant #8's provider parity matrix, invariant #10's consumer ledger):

1. `TurnLane` in `stella-protocol` — name the seven lanes so a matrix can have rows.
2. `TurnCapabilities` in `stella-core` — collapse three vocabularies into one struct with **no `Default`**, so adding a capability is a compile error at every lane. This is the whole design: *make forgetting a lane a build error.*
3. A lane matrix in `stella-parity`, enforced from both sides, with `#3232`/`#3233` as `Deferred` rows.
4. Lane resolution in `stella-runtime`, which already owns the assembly bottom half.

§6 raises the one question that needs a human decision — a `ResumeAuthority` split (`Own` / `Parent` / `Redispatch`) that would unblock `#3232` and `#3233` by observing that non-primary lanes need a *terminal frame*, not a per-step resume point. That is much cheaper than the write side an earlier fix declined, and it has a reader by construction.

## Witness

None — this is a documentation-only PR. Per CONTRIBUTING's carve-out, docs changes need no witness. Move 2 in the proposal names a real witness that fails on `main` for a structural reason (a fork carrying `gate`/`steering`/`hooks`), but no code ships here.

## Gate

`make guards-fast` passes (exit 0), including `check-doc-links`, `check-invariants`, `check-file-size` and `check-brand-case`. `docs/manifest.json` refreshed via `make doc-links-fix`.

Refs #3232
Refs #3233
Refs #3242

## Summary by Sourcery

Add a new design-spec document describing a unified assembly model for turn-loop capabilities across multiple execution lanes and register it in the docs manifest.

Documentation:
- Introduce the `turn-lane-assembly` spec documenting the current turn-loop lane/capability matrix and proposing a single-place assembly design.
- Update `docs/manifest.json` to include the new turn-lane-assembly specification entry.

---

## Review round 2 — §9 and §10 added

Answering the three inline comments.

**§6 `ResumeAuthority` is accepted** and PR 5 in §8 is now scheduled rather than open.

**§9 — plugin-owned lanes.** The requirement that plugins contribute their own lane is compatible with this design, but it forces one decision *now*: `TurnLane` is open from the first commit (`Builtin(BuiltinLane)` closed so the compile-error property survives, `Plugin(LaneId)` open so a manifest can contribute a row). Retrofitting that after seven lanes and a parity matrix are written against a closed enum is a rewrite. §9 also states the thing a plugin lane genuinely breaks — compile-time totality is only available in-tree — and replaces it with load-time validation plus a `stella plugins doctor` report, rather than pretending the two guarantees are equal. Plus: `participation` is derived from the capability vector rather than separately authored; `granted = requested ∩ authorized`, with #2716 as the missing half; and two constraints that bite Move 2 now (`TurnCapabilities` must admit owned slots, and a blocking plugin cannot live on the sync `HookBus`).

**§10 — now, later, or never.** Seam now, extraction later, merge never. With one correction to the premise: **the pipeline has already left `stella-core`** — that crate declares no `stella-pipeline` dependency and holds no witness/ladder/flip-oracle code. The coupling is `stella-cli` (169 references across 41 files), `stella-serve` (4 files), and `stella-runtime` (declared, referenced nowhere — filed as #3280).

`stella-serve` is the existence proof that makes the plugin half feasible: its engine holds no ambient authority and remotes every model and tool call to a host, so a plugin-driven lane is that seam pointed the other way.

**Note for reviewers:** two sessions wrote a plugin-lanes section against this branch concurrently, and `67a260016` captured both — two §9s and two §10s. `55495aa68` deletes the duplicate pair, re-attaches §8's closing sentence, and grafts forward the two mechanics #2456 had already settled (builtin-first resolution; nested `[lanes.custom.<id>]`, never `#[serde(flatten)]`). No test was deleted; this is a docs-only PR.

Refs #3246
Refs #2716
Refs #2456
Refs #3280


docs/spec/turn-lane-assembly.md
```diff
@@ -0,0 +1,365 @@
...
+change to a stage graph — the matrix makes that boundary explicit rather than
+moving it. If the goal is that re-implementing the pipeline touches one place,
+that is a separate (and mostly already-satisfied) property: `stage_rank` in
+`stella-pipeline/src/replay.rs` is the canonical ordering.
+
+**It does not, on its own, carry a behavioural witness for Moves 2–3.** They
+are refactors. Move 2's fork witness is real and load-bearing; the rest of
+Moves 2–3 are structural, and per this repository's own contract that should be
+declared in the PR rather than dressed up as a behaviour change.
+
```
macanderson, line 362: can we update the plugin architecture specs for when we build the plugin stuff that plugins should be able to create their own lane. Ideally pipeline is moved out of core stella and core stella remains the bare loop with a plugin surface that can change the behaviors of the loop - in other words we need someway to drive this from manifests/or registries or something. Let me know what you think via reply.

docs/spec/turn-lane-assembly.md
```diff
@@ -0,0 +1,365 @@
...
+    Parent,
+    /// The supervisor re-runs the unit. The checkpoint is evidence, not a
+    /// resume point. FleetWorker, PipelineStage.
+    Redispatch,
+}
+```
+
+The reason this unblocks rather than restates: **`Parent` and `Redispatch`
+lanes do not need a per-step resume point at all.** They need a *terminal
+frame* — last committed step plus transcript, written once when the lane dies
```
macanderson, line 330: im ok with this

docs/spec/turn-lane-assembly.md
```diff
@@ -0,0 +1,365 @@
...
+or ends. That is far cheaper than the per-step serialization the earlier fix
+declined, and it has a reader by construction: the parent's report, and the
+supervisor's re-dispatch decision. A fleet worker killed at step 40 stops
+losing 40 steps of transcript, without anyone paying for a resume path that
+nothing would ever call.
+
+**The decision needed from you is whether that three-way split is the right
```
macanderson: yes this is correct the question is do we refactor the verification engine/pipeline out of core stella now, later, or never.

docs/spec/turn-lane-assembly.md
macanderson: Recorded as accepted — the doc's status line now says §6 is settled and PR 5 in §8 is scheduled rather than open.

One thing worth pinning down before that PR starts, because it is the part that can go wrong quietly: the terminal frame is only cheap if it is written **once, at lane death**, and the write site is the lane's exit path rather than the step loop. If it lands inside `run_step` "just for safety" it becomes the per-step serialization #3242 declined, wearing a different name.

docs/spec/turn-lane-assembly.md
macanderson: Answered in a new **§10** on the branch (`55495aa68`). Short version: **later for the extraction, now for the seam, never for the merge.**

First, the premise needs one correction, because it changes the size of the job. **The pipeline has already left `stella-core`.** That crate's `[dependencies]` are `stella-protocol`, serde, serde_json, serde_json_canonicalizer, toml, thiserror, tokio, async-trait, futures-util, rand, sha2 — no `stella-pipeline`, no witness, ladder or flip-oracle module. The only `Verdict` identifiers in it are `LoopVerdict` (loop detection) and `GoalVerifierVerdict` (goal assessment); neither is the verification engine. So "core remains the bare loop" is not a refactor to schedule — it is the current state, and the job is to give that loop a registry rather than to carve anything out of it.

What is actually coupled, counted rather than estimated:

| Crate | Declares `stella-pipeline` | Source references |
|---|---|---|
| `stella-cli` | yes | **169 across 41 files** |
| `stella-serve` | yes | 4 files |
| `stella-runtime` | yes | **zero — the dependency is dead** |

`stella-runtime` declares the dependency in its `Cargo.toml` and references it nowhere in `src/`. That is a free deletion and it is now PR 6 in §8.

The recommendation:

- **Now** — the lane seam (Moves 1–4), with two amendments the plugin requirement forces today rather than later: `TurnLane` open from the first commit, and `TurnCapabilities` admitting owned slots rather than `&'a dyn` only. Plus the dead-dependency deletion.
- **Later** — the extraction. The pipeline registers as a plugin lane and `stella-cli`'s dependency is cut. Gated on #2716 for authority and on a side-by-side bench for equivalence, which is the gate #3245 already imposed on itself.
- **Never** — merging the pipeline into the loop. It is a supervisor; it stays a layer.

The reason the two halves split this way is that their costs of delay run in opposite directions. Opening `TurnLane` costs one enum variant today and is unaffordable to retrofit once seven lanes and a parity matrix are written against a closed one. The extraction costs 169 call sites whenever it happens, and doing it before the seam exists means doing it twice.

Definition of done for the cut, so it is decidable rather than debatable (§10.4): `rg 'stella-pipeline' crates/stella-cli/Cargo.toml` returns nothing; `stella run` still delivers today's pipeline mode through a manifest-registered lane; and a side-by-side bench on the same panel holds.

One thing I want to keep honest, in §10.5: the extraction buys modularity, a customer-owned definition of done, and the paid-plugin story. It does **not** buy the property that started this document. "One place to add a turn-loop feature" is delivered by Moves 2 and 3 whether or not the pipeline ever moves. Worth keeping those two claims apart so the extraction is never sold on a benefit it does not deliver.

docs/spec/turn-lane-assembly.md
macanderson: Yes — and it changes one decision in this document that is cheap now and expensive later, so thank you for raising it before PR 1 rather than after. Written up as **§9** on the branch (`55495aa68`); §10 answers the pipeline half.

**Feasible, and more feasible than it looks.** Four things are already true rather than needing to be built:

1. There is one loop (§1).
2. `stella-core` is already the bare loop — no pipeline dependency, no verification machinery (§9.6). So "core stays bare" is the current state, not a refactor.
3. **`stella-serve` already proves an out-of-process participant can drive a lane.** Its engine holds no ambient authority and every model and tool call is remoted to a host. A plugin-driven lane is that same seam pointed the other way, with the external process supplying capability instead of consuming it. That is an existence proof in the tree, and it is the single biggest reason to call this feasible.
4. The Stop gate already vetoes turn completion — a blocking hook injects feedback and the turn continues rather than finishing.

**The one thing your requirement genuinely breaks**, stated plainly rather than smoothed over: Move 2's whole value is a *compile error* — `TurnCapabilities` has no `Default`, so adding a capability forces the author to visit every lane. That guarantee only exists for lanes the compiler can see. A manifest-contributed lane is by construction not one. You cannot give an out-of-tree file a build failure.

So §9.2 makes it two regimes and makes the weaker one loud, rather than pretending they are the same:

| Lane origin | Enforced | By what |
|---|---|---|
| `TurnLane::Builtin(_)` | compile time | no `Default`, exhaustive destructuring in the parity test |
| `TurnLane::Plugin(_)` | **load** time | manifest validated against a capability registry generated from the same table |

with two rules that stop the load-time half rotting into a silence: an unknown capability name is a load error, never ignored; and a slot the plugin has never been told about is *reported* by `stella plugins doctor`, not defaulted quietly — otherwise a plugin lane ends up exactly where `subagent.rs:705` is today, sitting on a `None` nobody decided.

That is why **`TurnLane` is now open in Move 1** — `Builtin(BuiltinLane)` stays closed so the compile-error property survives for the workspace, `Plugin(LaneId)` carries a manifest row. This is the decision that is unaffordable to defer: retrofitting it after seven lanes and a parity matrix are written against a closed enum is a rewrite of the matrix.

Three more things that fell out, all of which I think you will want:

- **The `[loop] participation = none|observer|steering|arbiter` ladder in #3245 and the capability vector are the same statement at two resolutions.** `observer` = `bus`; `steering` = `+ steering`, `hooks`; `arbiter` = `+ gate`, `halt`, `completion_gate`. They must not both be authored — the manifest declares capabilities and `participation` is *derived* and displayed (doctor output, the Deck chip, the matrix). This repo has already lost one limit to having a number in two places.
- **A manifest requests; the host grants.** `granted = requested ∩ authorized(principal, capability)`. `authorized` is #2716's vocabulary, which is closed NOT_PLANNED with zero code in the tree. Until it exists, a paid plugin and a hostile one hold identical authority over the turn loop, and a `Plugin` row records an intention rather than a permission. The matrix should carry both columns from day one so #2716 has somewhere to write its answer without a schema change. **This is the real blocker on plugin lanes being safe** — not the lane work.
- **Two constraints bite Move 2 now, not later.** `TurnCapabilities` must admit owned slots, not `&'a dyn` only, or plugin lanes force a second parallel vocabulary — the exact disease §3 diagnoses. And a blocking plugin cannot live on the `HookBus`: its handlers are sync closures (`bus.rs:190`) and cannot await, so an arbiter whose definition of done runs a test suite belongs on the async out-of-process hook plane. Subprocess is the better substrate here, not the fallback.

Both mechanics #2456 already settled are carried forward in §9.7 so nobody re-derives them: resolution is builtin-first with a load-time rejection for a manifest naming `lead`, and the custom lane set nests as `[lanes.custom.<id>]` rather than `#[serde(flatten)]`, which the settings overlay silently skips.

Issue #3267, opened 2026-08-14
gate: a clean merge can delete a test helper another PR is concurrently using, and nothing catches it until main is red


macanderson:
## The problem

On 2026-08-14 `main` went red at `868ad77e0`: `stella-core` failed to compile, taking down `cargo clippy -D warnings` and `cargo test` for the whole workspace (run [31775648996](https://github.com/macanderson/stella/actions/runs/31775648996)). Neither contributing PR was faulty, and both were green on their own branch.

The mechanism, entirely inside `crates/stella-core/src/driver/restore.rs`:

- **#3244** (12-tool purge) replaced the `MapTools` test executor with the tool-less `SkillTools`, and deleted the `Arc`/`Mutex`/`HashMap` imports it needed, plus the now-unused `ToolCall`/`ToolOutput` imports and the `RESTORE_CALL_ID` constant.
- **#3253** (overflow-summarizer headroom) added two new witnesses to the same test module that construct `MapTools { files: .., active: .. }`.

Neither branch's diff contains the other's lines, so git merged both with no conflict to report, and the composed file references a struct and three imports that no longer exist.

Fixed by #3265. This issue is about the guard, not the fix.

## Why no guard caught it

`scripts/check-deleted-tests.sh` (CI, `pull_request` only) exists for exactly this class — a PR silently removing something another PR is concurrently depending on — but it compares **`#[test]` / `#[tokio::test]` function names** across the base tree and the merge result. What went missing here was a *test helper*: a `struct`, its `impl`, and its `use` statements. No test function disappeared, so the guard was silent and correct by its own contract.

The generalization is uncomfortable: **any item a concurrently-added test depends on can be deleted by a clean merge**, and the only thing that catches it is compiling the merge result. CI does compile the merge result — but on `pull_request`, against a base that did not yet contain the other PR. Both PRs passed that check honestly, before the other landed.

Related but not the same: #2495 (a clean merge silently reverting a measurement-tuned constant). Same family — "textually clean, semantically broken" — different payload.

## What "done" looks like

Decide and implement one of these, or argue in the issue for a third:

1. **Merge queue.** The only option that actually compiles the composition before it lands. Also the heaviest; check whether the repo's branch protection and Actions budget can carry it. Note `enforce_admins` is off and auto-merge is on, so today an admin merge can land a pair that was never built together.
[...]

PR #3290, opened 2026-08-14
fix(stella-core): drop the duplicate Mutex import two unbreak PRs both added

macanderson:
`main` is red again at `eb427c2da`. One line fixes it.

```
error[E0252]: the name `Mutex` is defined multiple times
   --> crates/stella-core/src/driver/restore.rs:426:9
422 |     use std::sync::Mutex;   <- #3263
426 |     use std::sync::Mutex;   <- #3265
```

## What happened

Two PRs diagnosed the *same* red main independently and shipped the same repair:

- **#3263** `fix(stella-core): repair main — the tool purge and #3253 composed into a red build` (`76ad90621`)
- **#3265** `fix(stella-core): unbreak main — restore.rs references a test double #3244 deleted` (`678f41c27`)

Mine was #3265. I checked for an in-flight fix before starting (`gh pr list` filtered on restore/fix/compile) and found none — #3263 was opened after that check and merged before mine, so by the time #3265 landed the repair was already in. The two edits sit at different lines, so git had no conflict to report and merged both.

The rest of the two repairs converged on byte-identical text — both pointed #3253's starvation witnesses at `SkillTools` and both removed the orphaned `ToolCall`/`ToolOutput`/`RESTORE_CALL_ID` — so this duplicated import is the entire delta.

## The fix

Delete the second import. Keeps the placement #3263 chose, since it landed first.

## Verify

```
cargo test -p stella-core --lib driver::restore   # 6 passed, 0 failed
cargo clippy -p stella-core --all-targets         # clean, no warnings
cargo fmt -p stella-core -- --check               # clean
```

No tests deleted or renamed. No behavior change.

## Note

This is the *second* red main in a row out of the same file, and the two causes are different: #3267 tracks the first (a clean merge deleting a test helper another PR was concurrently using). This one is the duplicate-repair hazard — two agents racing the same breakage — which a merge queue would also have caught, and which is more evidence for option 1 in #3267.

## Summary by Sourcery

Bug Fixes:
- Resolve a compile error in stella-core restore.rs by dropping the extra std::sync::Mutex import in the test module.

### DIFF
--- a/crates/stella-core/src/driver/restore.rs
+++ b/crates/stella-core/src/driver/restore.rs
@@ -423,7 +423,6 @@ mod tests {
 
     use async_trait::async_trait;
     use serde_json::Value;
-    use std::sync::Mutex;
     use stella_protocol::{
         CompletionMessage, CompletionRequestRef, CompletionResult, CompletionUsage, MessageRole,
         Provider, ProviderError, ToolOutput, ToolSchema,


macanderson:
Superseded by #3276, which landed the same one-line removal (plus the stranded test double). Closing rather than adding a fourth repair to the same file — see #3267 for the underlying race.

Issue #3310, opened 2026-08-14
Unify the plugin manifest's HookEvent mirror with stella-core::hooks::HookEvent in a shared home


macanderson:
## Problem

`crates/stella-plugin/src/manifest.rs` (#3245 slice A) declares a `HookEvent` enum that mirrors `crates/stella-core/src/hooks.rs::HookEvent` by name — same five variants, same wire strings (`SessionStart`, `PreToolUse`, `PostToolUse`, `Stop`, `PreCompact`). The mirror is deliberate: `stella-core` must never depend on `stella-plugin` (the engine never learns plugins exist — #3245 open question 3), and `stella-plugin` is a leaf that must not pull in `stella-core`. But two hand-kept copies of one wire vocabulary is exactly the drift shape AGENTS.md's invariant-8 story warns about: a sixth hook event added to the engine (as #2684 added `Stop`/`PreCompact`) would be undeclarable in a manifest until someone remembers the mirror, and nothing red flags the gap.

## What done looks like

One of:

1. Move the `HookEvent` enum to a home both crates may depend on (`stella-protocol` is the natural one — it is types-only and both crates' consumers already link it), re-exporting from `stella-core::hooks` so existing paths keep working; or
2. A witness test that fails when the two sets diverge (this requires a dev-dependency direction that is currently forbidden in both directions, so option 1 is likely the real fix).

Either way: adding a hook event in one place and not the other must become a compile error or a red test, not a review obligation.

## Files

- `crates/stella-plugin/src/manifest.rs` — the mirror (its doc comment cites this issue)
- `crates/stella-core/src/hooks.rs` — the original (`pub enum HookEvent`)
- `crates/stella-protocol/` — the candidate shared home (types only, zero logic — invariant "Serde-first" applies: keep the round-trip test when moving)

## Constraints

- Moving the enum into `stella-protocol` puts it on the wire-schema surface; check whether `wire-schema` (`docs/wire/**`) regeneration is needed in the same PR.
- `stella-core`'s serde renames are explicit (`rename = "SessionStart"` etc.); the moved enum must keep byte-identical wire strings.

Refs #3245.

Issue #3245, opened 2026-08-14
EPIC: plugins as turn-loop participants — manifest-only pipelines with witness/oracle flip, definition-of-done control, and deck presence, no core dependency


macanderson:
## The directive (Mac, 2026-08-13)

> Design a plugin system that by manifest only would allow a concept of
> pipeline with witness protocol / oracle flip all the way down to the visual
> indicator the plugin was active in the command deck, and the ability for the
> plugin all by itself to prevent the loop from exiting and control the
> definition of done (its core requirements). Some plugins are combinations of
> skills, custom tools, and have a subloop hooked into the core agent loop.
> Every plugin registers its intention to have a say in the turn loop or not.
> We need to support what full pipeline mode delivers today without stella
> itself having a dependency on those crates. The dirt-cheap researcher,
> dirt-cheap planner, and dirt-cheap prompt classifier/triage get built as
> plugins, not into stella-core.

This supersedes the host-API half of #1400 (Stella Apps, closed not-planned),
which specced plugins as **pure data** — explicitly forbidden from touching
the turn. The platform half of #1400 (manifest hygiene, lifecycle states,
OAuth, namespacing/overlay, marketplace, the neutrality gate) is still the
right design and is imported by reference, not restated. What changes is one
axis: a plugin may now **declare a say in the turn loop**, graded and
capability-gated.

Evidence basis: `main` at `a41770b24` (0.9.38), plus the seams named below —
each was read in this session, not assumed.

---

## 1. Why this is smaller than it looks: the seams already exist

The engine already has every mechanism the strongest plugin needs. What is
missing is a manifest that binds them and a host that enforces the grants.

| Capability the directive asks for | The seam that already ships |
|---|---|
| Prevent the loop from exiting / control definition of done | The **Stop gate**: `Engine::stop_hook_feedback` (`crates/stella-core/src/driver/user_hooks.rs`) runs when a turn is about to complete; a blocking hook injects a `[stop-hook feedback` tail message and the turn continues. Fail-open (a broken hook can never block), once-per-turn latch against death spirals. |
| Hook into the core agent loop | The user-hook surface shipped by #2684: `SessionStart`, `PreToolUse` (input-rewriting, permission-deciding), `PostToolUse`, `Stop`, `PreCompact` — all argv out-of-process with timeouts and bounded output (`crates/stella-core/src/hooks.rs`, `crates/stella-tools/src/hook_runner.rs`). |
[...]

Issue #1400, opened 2026-08-04
Epic: Stella Apps — a vendor-neutral extension platform (TOML manifest · OAuth lifecycle · host API · marketplace)


macanderson:
Stella needs a packaging vehicle for configuration and code that extends the base platform. One customer is driving this and the need is urgent: they want to run a **control plane** that aggregates logs, memories, context rules, context records, conventions, directives, procedures, and telemetry across every engineer running stella, mines useful context back down into each engineer's install, and attributes AI spend down to the level of a single context frame.

The entire design hangs on one rule:

> **The control plane is a guest, never a component.**

## Definition of done

Two tests. Both must pass, or this epic is not finished.

### Test 1 — a control plane can do the whole job

A vendor publishes an app. A user runs `stella app install <id>` and `stella app login <id>`. From there, with no vendor-specific code in this repository:

- the user's install authenticates to their cloud workspace and holds a refreshing token;
- telemetry from many engineers rolls up into one shared cloud workspace;
- the vendor mines skills, memories, conventions, directives, and style preferences and pushes them back down into each engineer's stella;
- AI spend is attributable down to a single context frame;
- the app declares post-install and pre-uninstall hooks and they run;
- the app reports one of a well-defined set of connection states in `stella app list` and in the deck.

### Test 2 — stella does not know that vendor exists

```sh
rg -i 'oxagen' --glob '*.rs' crates/ bench/    # → zero matches
```

Also: no vendor crate in any `Cargo.toml`; no vendor host name, env var, URL, or path constant compiled into any binary; no vendor name in a doc comment, a test fixture, a struct name, or a variable name. Take the vendor off the internet and every stella feature still runs, offline, forever.

**An app is data.** A TOML manifest, a set of endpoints, a token, and content it contributes. It is never Rust in this tree. If implementing an app requires editing a `.rs` file, the extension surface is wrong and the fix belongs in the surface, not in a special case.

---

## Where this actually stands today

Most of the transport already exists and is already vendor-neutral. This epic should extend it, not rebuild beside it.

### Already built — reuse

| Surface | What exists | Where |
| --- | --- | --- |
| **Cloud identity** | `stella cloud register / status / sync / logout / deregister`. `~/.stella/cloud.json` holds `CloudRegistration`, and the `oauth_token` slot is **already reserved**. Org-change ruling pinned in one function. | `crates/stella-cli/src/usage_cmd.rs`, `crates/stella-store/src/identity.rs` |
[...]

PR #3311, opened 2026-08-14
feat(stella-plugin): the plugin manifest — participation declared, never inferred (#3245 slice A)

macanderson:
## What

Slice A of #3245 (plugins as turn-loop participants): a new leaf crate, `stella-plugin`, that parses and validates the epic's new manifest blocks — `[loop]` with the participation ladder (`none` < `observer` < `steering` < `arbiter`), `[requirements]`, `[oracle]`, `[subloop]`, and `[roles]` — as pure functions over borrowed text. No I/O, no workspace-crate dependencies (the `stella-diag`/`stella-tty` leaf shape): the engine never learns plugins exist, and the host that binds these grants to the engine's gates (the Stop gate PR #3302 is generalizing, the hook runner, the sub-agent primitive) arrives with slices B–E.

Rules enforced, each with a typed `ManifestError` variant (invariant 5) and a rejection test:

- Undeclared `[loop]` = grade `none`; unknown keys/hook names/grades are load errors (`deny_unknown_fields` everywhere — the #1400 rule).
- No hooks below `steering`; `Stop` only at `arbiter`; an `arbiter` must declare `Stop` (an undeclared hook is never invoked, so an arbiter without it is a contradiction).
- `max_holds` and `[requirements]` are arbiter-only; an arbiter requires non-empty `[requirements]` (every hold must cite a named requirement).
- `[oracle]` is arbiter-only (conservative: widening later is compatible, narrowing is not), argv non-empty, timeout ≥ 1; `flip`/`tamper` are closed vocabularies (`"required"`, `"artifact-identity"`) so an unknown value is a load error, never a silently weaker contract.
- `[subloop]` requires ≥ `steering`, stages non-empty/unique; `[roles]` requires `[subloop]`, tiers are non-empty open-vocabulary intents (never a credential or URL — routing stays the user's).

`LoopGrant::permits_hook` is the authoritative filter behind "an undeclared hook is never invoked": it gates on grade *and* declared list, so even a hand-built grant cannot leak a dispatch.

## Witness

The crate is new, so the whole test surface is the witness: all 26 tests fail on `main` (the crate does not exist there) and pass here. Slice A's acceptance from #3245 §6 verbatim, in `tests/manifest_grades.rs`: fixture manifests at all four grades round-trip (through TOML and `serde_json` — invariant 4), and a hook not named in the manifest is not permitted even if the process registers for it.

## Decisions a reviewer should weigh

- **Crate name** — #3245 open question 3 (`stella-app` vs `stella-plugin`) is undecided; I picked `stella-plugin` because the epic's own vocabulary is "plugin" throughout. Renaming before anything depends on it is one `git mv` + member edit; cheap to veto now, expensive later.
- **`HookEvent` is a by-name mirror** of `stella-core::hooks::HookEvent`, because the dependency is forbidden in both directions. Filed #3310 for unifying the two in a shared home (likely `stella-protocol`) so the mirror stops being manual.
- **Arbiter-only `[oracle]`** is stricter than the epic's literal text (which only marks `max_holds`/`[requirements]` arbiter-only). Chosen because loosening later is backward-compatible while tightening breaks shipped manifests; the error variant's doc states this.

## Dependencies

No new workspace dependencies: `serde`, `toml`, `thiserror` were already in `[workspace.dependencies]`; `serde_json` is dev-only for the round-trip tests.

## Gate

`cargo test -p stella-plugin` (26 green), `cargo clippy -p stella-plugin --all-targets -- -D warnings`, `RUSTDOCFLAGS="-D warnings" cargo doc -p stella-plugin`, `make guards-fast` incl. god-files (README carries the no-god-files section), typed-errors, module-reachability, file-size (judged against base), doc-links, invariants — all green locally. AGENTS.md's crate count, god-file prose, and workspace table updated in the same PR.

Refs #3245
Refs #3310

## Summary by Sourcery

Introduce a new leaf crate for parsing and validating plugin manifests that declare participation in the turn loop, and wire it into the workspace metadata.

New Features:
- Add the `stella-plugin` crate to parse and validate plugin manifests including participation grades, hook grants, requirements, oracle configuration, subloop stages, and roles as pure functions over text.

Enhancements:
- Update workspace documentation and crate tables to describe `stella-plugin` and adjust crate and god-file counts accordingly.

Build:
- Register `stella-plugin` as a workspace member in Cargo.toml.

Tests:
- Add manifest-grade tests and TOML fixtures covering all participation levels, round-tripping through TOML and JSON and enforcing the undeclared-hook gating behavior.


macanderson:
Both addressed in f55af10c4.

**1. Linear duplicate checks.** Done — a `HashSet` for both the hook list and the subloop stages. `insert` returning false on the first repeat is exactly what the prefix scan answered (the earliest element some earlier element equals), so which duplicate gets named is unchanged.

The blank-stage check deliberately stays interleaved with the duplicate check in one pass rather than being hoisted into a pass of its own: for `["plan", "plan", " "]` the current code reports `DuplicateStage`, and for `["plan", " ", "plan"]` it reports `EmptyStageName`. A separate blank-first pass would flip the first case. That ordering is now pinned by `the_first_repeat_in_declaration_order_is_the_one_reported`, which uses exactly those two lists plus a hook list (`[PreToolUse, PostToolUse, PostToolUse, PreToolUse]`) that answers differently if the "first repeat" rule slips.

Evidence that no behavior changed: both new tests were run against the pre-refactor prefix-scan code as well, and pass there unchanged — they characterize shared behavior rather than witnessing new behavior. This commit is a pure refactor plus documentation, so per AGENTS.md it carries no witness test.

**2. The casing split — documenting it, not aligning it.** The divergence reads as drift but is load-bearing, so I've stated it on the type instead of removing it.

`stella-core::hooks::HookEvent` carries no `rename_all`, which makes `"PreToolUse"` already the string a user types in `.stella/settings.json` to register a shell hook (README.md, "Lifecycle hooks"). These five strings are therefore not this crate's to choose: spelling the same event `"pre_tool_use"` in a plugin manifest would fork one concept's name across two files the same user edits — strictly worse than the inconsistency it resolves. `Participation` is lowercase because its vocabulary is this crate's own invention (#3245 §2) and nothing outside spells it.

The `HookEvent` doc comment now says this and states the rule for future blocks: **a value mirroring an existing user-facing string keeps that string's casing; a value this crate coins is lowercase.** And `wire_strings_are_pinned_on_both_sides` pins all nine strings in both directions plus `Display`, so a `rename_all` added to either enum — exactly the tidying this spelling invites — fails a test rather than silently invalidating every shipped manifest.

Gate: `cargo test -p stella-plugin` (28 green, up from 26), clippy `-D warnings`, `RUSTDOCFLAGS="-D warnings" cargo doc`, and `make guards-fast` all clean. No crate depends on `stella-plugin` yet, so that scope is complete.

Issue #3288, opened 2026-08-14
A gate's RequireApproval cannot reach a human — route it through the #2676 ApprovalBroker


macanderson:
`AuthzGate` can return `AuthzDecision::RequireApproval`, and `authz_verdict` folds it to `GateVerdict::RequireApproval` correctly — but `GatedToolSet::execute` (PR #3281) turns that verdict into a refusal:

> `` `{name}` needs a human's approval ({reason}), and this session has no approval route attached — it cannot be asked from here ``

That matches the documented headless posture and is safe, but it means a gate cannot actually ask anyone anything. The ask half of the vocabulary is inert.

## The route already exists

#2676 shipped `ApprovalBroker` with emit-before-park, a bounded TTL wait, and `approval.granted` / `denied` / `expired` audit events. `stella-core`'s `HookApprovalRoute` port is how the shell-hook surface reaches it without `stella-core` depending on `stella-tools`, and `stella_tools::hook_bridge::BrokerApprovalRoute` is the production implementation. A gate should reach the same broker through the same port rather than growing a second approval path.

## Definition of done

- `GatedToolSet` takes an optional approval route and resolves `RequireApproval` through it.
- With no route attached, keep today's refusal wording — an engine that cannot ask must not silently allow, and the message should name the grant path.
- Reuse `HookApprovalRequest` / `HookApprovalResolution` rather than inventing a parallel pair. If the naming is too hook-specific once a second producer exists, rename in the same PR and chase it through — a stale name is a bug in review here.
- Honour `stella-tty`'s `human_can_answer(...)` derivation for whether there is anybody to ask, rather than re-deriving presence.

## Related, and probably worth doing together

#2793 — MCP and custom tools bypass the blocking policy chains and the #2676 approval flow. `GatedToolSet` sits above both layers by construction, so it is the natural place to close that bypass for the approval path too. Check whether #3283's wiring already resolves part of it before scoping.

## Witness

A gate returning `RequireApproval` parks and runs the call when a scripted route approves; the same gate and a denying route refuses without executing. Anti-vacuity: with `NoAuthz` the call runs and no approval is ever requested.

Refs #2716, #2676, #2793, #3281.

Issue #2676, opened 2026-08-10
CORRECTION: RequireApproval dead-ends as a model-visible error — no path from the policy chain to an interactive grant


macanderson:
## What's wrong

A `RequireApproval` decision from the blocking policy chain dead-ends: it surfaces as a model-visible `ToolOutput::Error` (`crates/stella-tools/src/registry.rs:1308-1311`) and nothing more. There is no path from the tool boundary to an interactive grant — the human is never actually asked, the model just sees a refusal. `approval.requested` is emitted on the bus, but no shipping surface converts it into a question.

Meanwhile Stella *has* an interactive escalation primitive — `ask_user` (`crates/stella-cli/src/interactive.rs:387-490`) and the confirmation-gated `install_skill` — so the plumbing for asking exists; it just isn't wired to policy decisions.

## What the comparator does

Zappy code treats "ask" as a first-class permission outcome: a rule or hook can return `ask`, which forces a permission dialog carrying the hook's message; the user's answer allows or denies the call, and (notably) an `allow` from a hook still cannot override an explicit deny rule — deny > ask > hook-allow is a fixed precedence. The tool call then proceeds or fails based on a real human decision instead of a synthetic error.

## Proposed fix

Route `RequireApproval` to the interactive surface when one exists (TUI/CLI session): pause the dispatch, present the request via the same machinery `ask_user` uses, and convert the answer into Allow/Deny. In headless contexts, keep today's behavior but say so in the error text ("approval required and no interactive surface — rerun with ... or grant via policy"). Define precedence explicitly so a future hook `allow` can never bypass an operator deny.

> **Revision 2 (2026-08-10):** re-scoped as the approval-flow deliverable of the governance half of the tool-first epic — #2716 defines the seam this flow plugs into.

## Revised design against the `AuthzGate` port (#2716)

`RequireApproval` stops being a hook-chain quirk and becomes a first-class `Decision` variant shared by the whole authorization seam (settings policy, shell hooks, in-process extensions, RBAC plugins). The flow:

1. **Emit before blocking.** `approval.requested` (with the contract's `risk`, `read_only`, and the gate's `reason`) goes onto the bus *before* the dispatch parks. This ordering is a hard rule learned from oxagen-platform's approval gate: block-first leaves the event stream silent and the approval card never renders (`materialize-tools.ts:447-450` is the scar).
[...]

PR #3338, opened 2026-08-16
feat(stella-cli): wire GatedToolSet into every shipped session stack (#3283)

### LINKED ISSUE #3283: Wire GatedToolSet into the shipped session stack — the authz seam is currently constructed by its tests only
PR #3281 landed `stella_tools::gated::GatedToolSet`, the decorator that asks an `AuthzGate` before every tool dispatch. **Nothing in the shipping binary constructs it.** Until this issue is done, the authorization plane exists as a type and enforces nothing in a real session.

This is the wiring issue AGENTS.md requires to accompany scaffolding, and it is the one that makes #2716's seam real.

## Why it was not done in #3281

The production stack is assembled in `crates/stella-cli/src/agent.rs` (`PolicyToolSet::new(&customs, session_tool_policy(cfg))` around line 360, and a second site around line 1690). `agent.rs` is a **grandfathered god file closed to growth** (`scripts/file-size-baseline.txt`), so new lines must not land in it. The prescribed pattern is to extract into a sibling submodule — `crates/stella-core/src/driver/settlement.rs`, split out of `driver.rs`, is the exemplar — and that extraction is its own logical change, touching session assembly on the hot path.

## Definition of done

1. Extract the tool-chain assembly out of `agent.rs` into a sibling module (e.g. `crates/stella-cli/src/agent/tool_stack.rs`), which should *reduce* `agent.rs`'s line count. Both assembly sites, plus `subsession.rs:736` and `candidate_ws.rs:495,514`, should end up building the chain through one function.
2. Wrap the outermost layer in `GatedToolSet`, above `PolicyToolSet`. The stack diagram in `crates/stella-tools/src/gated.rs`'s module docs is the intended order.
3. Default the gate to `stella_core::ports::NoAuthz`, **chosen by name** — behaviour must be unchanged for every existing session. Do not add a nullable slot.
4. Construct the `Principal` honestly per surface: `Principal::User` for an interactive session, `Principal::SubAgent(id)` for a spawned child, `Principal::Role(role)` for a pipeline stage.
5. Pin the composition the way the existing decorator-forwarding tests do — `the_production_tool_stack_forwards_sub_agent_spend` in `stella-cli` is the shape to copy. A gate layer that swallows `drain_sub_agent_spend_usd` would silently drop money out of the budget; `GatedToolSet` forwards all six port methods, and the test should prove it through the *real* stack rather than a hypothetical one.

## Witness

A test that builds the production chain and asserts a `DenyAll` gate blocks a call that the same chain with `NoAuthz` allows. `crates/stella-tools/src/gated.rs`'s unit tests are the in-crate version; this one must go through the assembled session stack, because the entire point is that the seam is reachable in a real session.

## Constraints

- Do not add lines to `agent.rs` or `command_deck.rs`.
- `GatedToolSet` enforces at `execute()` and deliberately does not filter `schemas()` — see its module docs. Do not "fix" that here.
- Behaviour with `NoAuthz` must be byte-identical to today, including the advertised tool list (invariant #7).

Refs #2716, #3246.

### LINKED ISSUE #3339: Subsession lanes and fleet workers never see .stella/tools custom tools — the deck and one-shot turns do
Noticed while wiring the tool-chain assembly seam (#3283, PR #3338); pre-existing, not introduced there.

## Problem

The deck's lead turn, the pipeline turn, the one-shot turn, the goal loop, and resume all layer `CustomToolSet` (the workspace's `.stella/tools/*.toml` tools) into the session chain — now via `agent::tool_stack::session_stack`. Subsession lanes (`crates/stella-cli/src/subsession.rs`, the `policy_stack` call over the `ClaimTap`) and fleet workers (`crates/stella-cli/src/fleet_cmd.rs::run_task`, same shape) do not: their chains are claims → policy → gate straight over the registry, with no custom layer.

The subsession comment says "a worker's tool surface is the session's, so the operator's switches apply to it identically" — but the *surface* is not the session's: a custom tool the lead can call is absent from the lane. Either that is deliberate (untrusted-by-default customs withheld from autonomous children — defensible, and rhymes with the #2716 trust posture) or it is an oversight from those drivers predating customs.

## What done looks like

Decide which it is, and make it explicit:
- If deliberate: a sentence in `agent/tool_stack.rs`'s `policy_stack` docs and the subsession/fleet call-site comments saying customs are withheld from dispatched workers on purpose, and why.
- If an oversight: pass `discover_custom_tools`' result into those two drivers and switch them to `session_stack` with the same `Principal`s they carry today; witness = a custom tool callable from a lane that previously refused it.

## Constraints

- `agent.rs`/`command_deck.rs` are god files closed to growth; the assembly seam is `crates/stella-cli/src/agent/tool_stack.rs` (#3283) — change it there, not per-driver.
- Custom-tool claims are untrusted (`ToolContract::declared`, graded High) — the trust posture must not soften as a side effect.

Refs #3283, #2716

macanderson:
## What

`agent::tool_stack` is now the one place a session's tool chain is assembled — customs, the operator's `PolicyToolSet`, and the `AuthzGate` decorator (`GatedToolSet`, #3281) outermost — and every driver builds through it: the one-shot turn (both paths, process-free included), the goal loop, resume, the deck's lead and pipeline turns, subsession lanes, fleet workers, and best-of-N candidate workspaces. Until this change the authorization plane existed as a type and enforced nothing in a real session.

## Design

- **Gate chosen by name**: `tool_stack::session_gate()` returns `NoAuthz` explicitly — no nullable slot defaulting open (#2716's constructor-dependency rule). A deployment that grows a real gate changes that one function.
- **Principals honest per surface**: `Principal::User` for interactive/one-shot/deck turns; `Principal::SubAgent(session/lane)` for subsession lanes and `SubAgent(task-id)` for fleet workers; `Principal::Role("worker")` (the canonical role spelling per `check-role-names`) for candidate workspaces — best-of-N is not a way around authorization (`policy_stack_owned`).
- **By-value `Boxed` inner** on `PolicyToolSet` and `GatedToolSet` (exemplar: the existing `Borrowed`/`Owned` split those types already carry, mirrored from `CustomToolSet`) so the assembly function returns the whole chain as one value while the base still borrows the session's registry.
- **Behaviour with `NoAuthz` is unchanged**, including the advertised tool list: the gate enforces at `execute()` only (its module docs' deliberate asymmetry), so the prompt-cache prefix bytes are untouched (invariant #7).
- `agent.rs` shrank 1837→1830 and `command_deck.rs` 4092→4091 — both god files reduced, per the issue's constraint.

## Witnesses

- `tool_stack::tests::a_denying_gate_blocks_a_call_the_default_session_stack_allows` — through the **assembled production chain**, a denying gate refuses a call the shipped default allows and the base is never reached (fails on `main`: the module does not exist there).
- `the_owned_candidate_stack_is_gated_too` — the candidate-workspace (owned) assembly enforces the same seam.
- The four production-stack forwarding witnesses in `subagent/tests.rs` (`…forwards_sub_agent_spend` / `wait_requests` / `parallel_safe_names` / `live_services`) now assemble through `tool_stack`, pinning that the gate layer swallows neither spend, waits, concurrency claims, nor live services.
- Verified: `make gate` green (full workspace), including `doc-warnings`, clippy `-D warnings`, and the 1734-test CLI suite.

## Tests removed

None deleted; three tests re-expressed in place (same names): the two `subagent/tests.rs` forwarding witnesses and `agent/tests.rs`'s `stack_names_and_execute` helper now build via the real assembly instead of hand-stacking `CustomToolSet`+`PolicyToolSet`. The `agent::PolicyToolSet` re-export was removed (drivers no longer use it); remaining test callers name `crate::tool_policy::PolicyToolSet` directly.

Closes #3283
Refs #2716, #3246

## Summary by Sourcery

Centralize session tool-chain assembly in a shared `agent::tool_stack` module that layers custom tools, policy filtering, and authorization gating for all drivers and execution paths.

New Features:
- Introduce a unified session tool-stack builder that composes `CustomToolSet`, `PolicyToolSet`, and `GatedToolSet` for all CLI agent drivers.
- Add support for by-value (boxed) inner executors in `PolicyToolSet` and `GatedToolSet` to allow returning fully assembled tool chains as single values.
- Wire authorization principals into tool-stack construction so different execution surfaces (user sessions, subsessions, fleet workers, candidate workspaces) can be distinguished.

Bug Fixes:
- Ensure best-of-N candidate workspaces and other non-interactive paths are subject to the same tool policy and authorization gate as primary session paths, closing potential bypasses.

Enhancements:
- Replace ad-hoc tool-stack assembly across the codebase with calls into the shared `tool_stack` helpers for consistency and easier future extension.
- Document the centralized tool-stack assembly and its relationship to settings-based tool policies in the CLI README.
- Tighten tests around tool-stack behavior by routing existing witnesses through the production assembly and adding coverage that a denying gate blocks calls before they reach the base executor.

Tests:
- Rework existing subagent and agent tests to exercise the real production tool-stack assembly instead of hand-built stacks.
- Add tests that verify the gate layer blocks denied calls through the fully assembled session stack, including the owned candidate-workspace stack.

## Also: dispatched workers carry the workspace's custom tools (#3339)

Found while wiring the seam: subsession lanes and fleet workers built their chains with no `CustomToolSet` layer, so a `.stella/tools` tool the lead could call was silently absent from every worker. `tool_stack::worker_session_stack` now discovers the workspace's customs (quietly — a worker is headless; a fleet worker discovers from its own isolated worktree so custom-tool subprocesses stay inside the isolation) and assembles the identical customs → policy → gate chain. Trust posture unchanged: discovery still honours `project_custom_tools_allowed` and the foundry gate, and custom-tool contracts stay untrusted `High`. Witness: `a_dispatched_worker_carries_the_workspace_custom_tools` — the worker chain advertises a workspace tool the pre-fix custom-less chain (asserted in the same test) does not.

Closes #3339


macanderson:
Heads-up from the parallel #3288 work: PR #3337 gives `GatedToolSet` the approval route this wiring can attach — `.with_approval_route(Arc::new(BrokerApprovalRoute::for_registry(&registry).with_gate_name(gate.name())))` after the responder is attached — so a gate's `RequireApproval` parks on the #2676 broker instead of refusing. #3337 also renames the port types (`HookApprovalRoute` → `ApprovalRoute`, request/resolution likewise) in `stella-core::hooks::decision`; it does not touch `stella-cli`, and the `Engine::with_hook_approval_route` builder name is unchanged, so no textual conflict is expected — but whichever PR merges second should sanity-rebuild. Attaching the route in the drivers can land as a small follow-up to whichever lands last.

macanderson:
Review comments addressed in 4cadf6c3a + 59b923563:

- **`custom_tools` cloning** — `session_stack` / `session_stack_with_gate` now take `impl Into<Cow<'_, [CustomTool]>>`, and `CustomToolSet` holds a `Cow<'_, [CustomTool]>` (mirroring its existing borrowed/owned `Inner` split). Every per-turn driver now lends the tools the session already owns — all six `custom_tools.to_vec()` call sites are gone. The owned candidate path (`new_owned`) still carries its own copy past the borrows, as it must.
- **Hard-coded principal spellings** — every principal a shipped driver acts as is now minted in `agent::tool_stack` (the module whose docs already claim ownership of the principal naming): `worker_principal()` derives the candidate workspace's role name from `AgentId::from_role(Role::Worker)` — the same spelling authority `check-role-names.sh` guards — instead of retyping the `"worker"` literal, and `lane_principal` / `fleet_task_principal` centralize the subsession and fleet `SubAgent` ids.

Also updated `subsession.rs`'s module doc, which still claimed workers run "no custom tools" after 44bfb64f3 (#3339) widened them.

Verified: `cargo fmt --check`, `cargo clippy -p stella-tools -p stella-cli --all-targets -- -D warnings`, and `cargo test -p stella-tools -p stella-cli` (all 15 suites green, including the #3283 witness tests through the changed signatures).

PR #3353, opened 2026-08-16
fix(scripts): stop selecting --help text by hardcoded line numbers

### LINKED ISSUE #3350: scripts: 11 help printers select their text by hardcoded line numbers and truncate silently when the header grows
## Problem

Eleven scripts print their `--help`/usage text with a hardcoded line range over their own header comment, e.g. `sed -n '2,30p' "$0"`. The range silently truncates (or leaks non-help lines) the first time anyone edits the header — there is no failure, the help just goes quietly wrong. Review feedback on PR #3343 flagged this shape; that PR fixed its own two new scripts (`scripts/check-lockfile-sync.sh`, `scripts/main-canary.sh`, commit bc783322) but deliberately left the pre-existing ones alone.

## Affected files

```
scripts/automerge-nudge.sh:49
scripts/arena-run.sh:81
scripts/arena-local.sh:65
scripts/arena-kill.sh:66
scripts/record-demo.sh:78
scripts/check-releases-published.sh:63
scripts/reap-agents.sh:74
scripts/demo-scenario.sh:31
scripts/impacted-crates.sh:93
scripts/reap-seats.sh:79
scripts/repro-build.sh:97
```

(Line numbers as of bc783322; re-locate with `rg -n "sed -n '2," scripts/*.sh`.)

## Fix

Replace each hardcoded range with the self-bounding form now used by the two fixed scripts — print the whole comment block after the shebang, stopping at the first non-comment line:

```sh
awk 'NR == 1 { next } !/^#/ { exit } { sub(/^# ?/, ""); print }' "$0"
```

Caveats to check per file:

- Some scripts' usage text is only a *subsection* of the header (e.g. `record-demo.sh` prints lines 2–29 of a longer header, `impacted-crates.sh` 2–30). For those, either the whole header is acceptable help, or the usage block needs to move to the top of the header so the block-bounded form prints the right thing. Decide per file by reading the header, not mechanically.
- Two scripts use tabs before `sed` (`impacted-crates.sh`) — preserve indentation.
- Some strip the comment prefix differently (`sed 's/^#//; s/^ //'` vs `s/^# \{0,1\}//`) — the awk form's `sub(/^# ?/, "")` normalizes this.

## Verify

- `bash <script> --help` (or the script's usage path) prints the intended text with no `set -euo pipefail` line and no truncation mid-sentence, for each of the 11 files.
- `shellcheck` stays clean on every touched script (`make shellcheck`).
- `make guards-fast` green.

## Done when

No script under `scripts/` selects its help text by hardcoded line numbers (`rg "sed -n '2," scripts/` returns nothing), and each script's help output is byte-equivalent to (or a deliberate, reviewed improvement over) what it printed before.

macanderson:
## What

Eleven scripts printed their `--help` text with `sed -n '2,Np' "$0"` over their own header — a range that goes quietly wrong the moment anyone edits the header, because nothing fails; the help just truncates or leaks. Reading all eleven before touching them showed the failure was not hypothetical:

| script | what its range actually printed |
|---|---|
| `repro-build.sh` | cut off **mid-option** — the usage block sits at the *end* of a header that grew past line 60, so `--help` showed the design essay and half of `--locked`'s description |
| `impacted-crates.sh` | cut off **mid-sentence** inside the "three kinds of edge" essay |
| `arena-run.sh` | printed the first prose line of the design essay |
| `arena-local.sh`, `automerge-nudge.sh`, `check-releases-published.sh` | printed the *next section's heading* (`WHY THIS EXISTS…`, `── The deadlock ──`, `── The silence ──`) |

## How

One self-bounding printer replaces all eleven:

```sh
awk 'NR == 1 { next } /^# \(--help text ends here\.\)$/ { exit } !/^#/ { exit } { sub(/^# ?/, ""); print }' "$0"
```

- **Whole-header scripts** (`demo-scenario.sh`, `repro-build.sh`): the block is bounded by the first non-comment line — same form PR #3343's review pass gave `check-lockfile-sync.sh` and `main-canary.sh`, which is where this pattern was first flagged.
- **Usage-subsection scripts** (the other nine): their help is deliberately only the usage/options block of a much longer design header, and there is no structural boundary common to all of them (some essays start with an all-caps heading, some with plain prose). So the boundary becomes a **declared fact in the file** — a `# (--help text ends here.)` sentinel at the split — instead of a number that has to track edits made forty lines away.

Two deliberate output changes, called out rather than slipped in: the eight scripts that piped `sed -n` raw no longer leak `# ` prefixes into their help text, and `record-demo.sh`'s `-h` description now says "show this help text" instead of "show this header" (it never showed the whole header, and now demonstrably shows the usage block).

## Witness

Ran every one of the eleven `--help` paths and asserted, per script: exit 0, the last printed line is the final usage/option line (not a heading or half a sentence), and the output contains no sentinel line, no `# ` prefix, and no script body. `repro-build.sh` now prints its complete header including the previously-severed usage block.

The fail half of the flip is the table above: on `main`, four of these print content past their intended block and two truncate mid-sentence — observable by running the same `--help` commands there.

## Verification

- `rg "sed -n '2," scripts/` — empty (the issue's done-when condition)
- `shellcheck` clean on all eleven; `make shellcheck` green
- `make guards-fast` green
- `make impacted-test` — 25/25 (the one touched script with its own suite)

No Rust changed; compile tiers left to CI.

Closes #3350

## Summary by Sourcery

Standardize how scripts print their --help text by replacing fragile sed-based line ranges with a self-bounding header printer using comment sentinels or first non-comment lines, eliminating silent truncation or leakage when headers change.

Bug Fixes:
- Fix incorrect --help output in multiple scripts where hardcoded sed line ranges either truncated usage text or leaked into neighboring documentation sections.

Enhancements:
- Ensure all affected scripts strip leading comment markers from help output and align help descriptions, including clarifying record-demo.sh's -h message to describe the actual help text shown.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -285,7 +285,7 @@ doc-warnings: ## Assert rustdoc is clean workspace-wide, private items included 
 
 .PHONY: shellcheck
 shellcheck: ## Lint install.sh, scripts/*.sh, and .githooks/* (#916)
-	shellcheck install.sh scripts/*.sh .githooks/*
+	shellcheck install.sh scripts/*.sh scripts/lib/*.sh .githooks/*
 
 # Deliberately not part of `gate`: it needs a Docker daemon, which the gate
 # must not. CI runs the same two commands (.github/workflows/docker-serve.yml).
--- a/scripts/arena-kill.sh
+++ b/scripts/arena-kill.sh
@@ -9,7 +9,7 @@
 #                  the destructive mode; see "Which signal" below
 #     --timeout N  seconds to wait for a signalled server to exit before
 #                  escalating to KILL (default 5, or 20 with --cancel)
-# (--help text ends here.)
+# --help text ends here.
 #
 # WHY THIS EXISTS
 #
@@ -59,15 +59,15 @@
 
 set -uo pipefail
 
+# shellcheck source=scripts/lib/help-header.sh
+. "$(dirname "$0")/lib/help-header.sh"
+
 DRY_RUN=0
 SIGNAL="TERM"
 TIMEOUT=""
 
-# The header comment block, bounded by the sentinel above (or the first
-# non-comment line) — a hardcoded line range truncates silently when the
-# header grows (#3350).
 usage() {
-  awk 'NR == 1 { next } /^# \(--help text ends here\.\)$/ { exit } !/^#/ { exit } { sub(/^# ?/, ""); print }' "$0"
+  print_help_header "$0"
 }
 
 while [ $# -gt 0 ]; do
--- a/scripts/arena-kill.sh
+++ b/scripts/arena-kill.sh
@@ -9,6 +9,7 @@
 #                  the destructive mode; see "Which signal" below
 #     --timeout N  seconds to wait for a signalled server to exit before
 #                  escalating to KILL (default 5, or 20 with --cancel)
+# (--help text ends here.)
 #
 # WHY THIS EXISTS
 #
@@ -62,8 +63,11 @@ DRY_RUN=0
 SIGNAL="TERM"
 TIMEOUT=""
 
+# The header comment block, bounded by the sentinel above (or the first
+# non-comment line) — a hardcoded line range truncates silently when the
+# header grows (#3350).
 usage() {
-  sed -n '2,12p' "$0"
+  awk 'NR == 1 { next } /^# \(--help text ends here\.\)$/ { exit } !/^#/ { exit } { sub(/^# ?/, ""); print }' "$0"
 }
 
 while [ $# -gt 0 ]; do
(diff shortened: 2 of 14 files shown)


macanderson:
Both review comments addressed in 4b1879761:

- **Duplication** — the awk program now lives once, as `print_help_header` in a sourced helper, `scripts/lib/help-header.sh`. All eleven scripts call it, plus `scripts/self-driving.sh`, which carried its own variant of the same structural read. Each script sources the helper right after its `set` prologue — before any `cd` (`repro-build.sh` changes directory before its `--help` arm runs) — and `scripts/lib/*.sh` joins the shellcheck gate glob in the Makefile.
- **Sentinel regex** — the sentinel drops its parentheses: the line is now `# --help text ends here.`, matched as `/^# --help text ends here\.[[:space:]]*$/` — no escaped metacharacters beyond the literal dot, and trailing whitespace tolerated.

Verified: `make guards-fast` and `make shellcheck` green; all twelve scripts' `--help` prints the header up to (and excluding) the sentinel, exit 0; `test-impacted-crates.sh`, `test-automerge-nudge.sh`, `test-releases-published.sh`, `test-arena-scripts.sh`, and `test-self-driving.sh` all pass.

Issue #3303, opened 2026-08-14
Silent custom-tool success is byte-identical output the harness cannot stamp — the #3176 stagnation hazard survives for unstampable tools (custom scripts, MCP), and #3187's detector-side fix was never filed


macanderson:
## Problem

The #3176/#3188 hazard class — the engine's stagnation detector keys on repeated byte-identical tool **output**, so a tool whose success output is constant converts a legitimate work loop into a stagnation kill — is fixed only for tools whose output Stella authors. It survives structurally for the population whose output Stella merely **relays**:

**1. Custom script tools.** `crates/stella-tools/src/custom.rs:624-628` (at `e352cf486`):

```rust
let stdout = String::from_utf8_lossy(&output.stdout);
if output.status.success() {
    ToolOutput::Ok {
        content: crate::exec::truncate_middle_capped(&stdout, MAX_OUTPUT_BYTES),
    }
```

The success output is the child's stdout **verbatim** — `truncate_middle_capped` returns `s.to_string()` unchanged under the 64 KiB cap (`crates/stella-tools/src/exec.rs::truncate_middle_capped`, first branch) — so a script that exits 0 printing nothing yields `ToolOutput::Ok { content: "" }` on **every** call. Distinct effects (different files written, different arguments) render byte-identical empty outputs unconditionally — no same-length coincidence needed, which makes this *worse* than the deleted `write_file` case. The detector's rung fires on exactly this: `crates/stella-core/src/loop_detect.rs`, "Consecutive calls to the SAME tool that all produced byte-identical output — **whatever their arguments**".

This is the shape Stella itself authors: `crates/stella-tools/src/foundry_author.rs:177-186` (`render_script`) wraps a mined command in `#!/bin/sh` + `set -eu` + the command line, so any quietly-mutating mined shape (`cp`, `mv`, `tee`, `sed -i`, a build writing an artifact) becomes, after adoption, a registered tool whose every success is the empty string.

Nothing in `run_custom` can stamp a *content* digest the way PR #3187 did for `edit_file` — the content is the child's stdout by contract, and in the empty case there is no content to digest.

**2. MCP tools.** Third-party server outputs are the same relay shape one layer out; Stella cannot stamp them either. PR #3187's body named this population and deliberately left its defense untaken:

[...]

Issue #3287, opened 2026-08-14
Ingest manifest and MCP tool annotations as untrusted claims instead of hardcoding read_only: false


macanderson:
`crates/stella-tools/src/custom.rs` hardcodes `read_only: false` for every `.stella/tools/*.toml` tool, and MCP tools get the same treatment — so a genuinely read-only third-party tool cannot say so, and the MCP spec's own `readOnlyHint` / `destructiveHint` / `idempotentHint` annotations are discarded entirely.

PR #3281 built the machinery that makes ingesting them safe: `ToolContract::declared` records a claim while grading it `High`, and `trusted_read_only()` is the one place the distinction is load-bearing. This is #2716 §6.

## Definition of done

- `.stella/tools/*.toml` gains `read_only`, `risk`, `idempotent` and an optional `[output_schema]`. The manifest is the **TOML serialization of the same contract**, not a second format.
- MCP tool annotations map into the same contract fields.
- Both arrive as `Provenance::Declared`. A claim is displayed and available to policy; it **never** admits the tool to the read-only/speculation dispatch sets on its own — that is `trusted_read_only()`, and `crates/stella-tools/src/contracts.rs`'s tests already pin it.
- The one route from claim to trust is the existing foundry adoption gate (`crates/stella-tools/src/foundry_gate.rs`): witnessed adoption + human enablement + byte-digest match. Nothing else upgrades provenance.
- Add the `idempotent` field to `ToolContract` here, with this as its consumer — it was omitted from #3281 rather than shipped unread.
- Document the asymmetry in the `stella-tools` and `stella-mcp` crate READMEs.

## Why the untrusted posture is not paranoia

The read-only bit is not decorative: it admits a tool into concurrent dispatch alongside other reads, and it fences a verifier into a set it cannot mutate the workspace from. A false claim there is a data race and a broken verification boundary, not a cosmetic error. An MCP server's metadata is untrusted for the same reason its output is (#2689).

## Witness

A manifest declaring `read_only = true` is displayed as read-only, is refused by a `Medium` risk ceiling, and does **not** appear in the executor's read-only dispatch set — with the built-in equivalent doing all three the other way.

Refs #2716, #2689, #3281.

PR #3354, opened 2026-08-16
fix(stella-tools): restore the silent-success stamp the #3346 merge dropped

### LINKED ISSUE #3359: main is red: #3345's rewrite of run_custom's success branch dropped #3335's silent-success stamp (clean textual merge, semantic conflict)
## Problem

`main` is red at c47b5623a: `cargo test -p stella-tools --lib` fails on
`custom::tests::silent_success_is_stamped_with_the_input_identity` (both renders come back as the empty string).

PR #3335 (34ef87ab8) added the empty-stdout branch in `run_custom` that renders `silent_success_stamp` instead of `""`. PR #3345 (5e7343be9), developed in parallel, rewrote the same success branch for output-schema validation and its version does not contain the stamp call. Both PRs were green on their own branches; the merge was textually clean, so git had no conflict to report — the stamp function and its test survived, but the call site vanished. Same shape as #1976/#1860 (deleted-test guard), except here it is a *call site* that vanished, which no guard covers.

Surface signal that first caught it: PR #3348's `fmt + clippy + test` job failed on this test even though that PR touches no stella-tools code — CI tests the merge with main.

## Fix

Restore the guard clause at the top of the success branch in `crates/stella-tools/src/custom.rs` (`run_custom`): if `output.stdout.is_empty() && tool.output_schema.is_none()`, return `ToolOutput::ok(silent_success_stamp(&tool.name, input))`. Schema-declaring tools keep falling through to the output-contract breach (empty stdout is not JSON), which is the correct #3287 behavior.

## Verify

`cargo test -p stella-tools --lib silent_success` — fails on main, passes with the guard clause restored. The existing test is the witness.

## Notes

- The new `scripts/main-canary.sh` (0d570263b, #3343) is aimed at exactly this class; worth checking whether it fires on this breakage.
- Fix PR incoming on branch `fix/silent-success-stamp-restored`.

macanderson:
## What

Two regressions landed on main within minutes of each other and are keeping `fmt + clippy + test` red on every commit since. This PR unreds main.

**1. The #3346 merge dropped #3335's silent-success stamp.** The conflict resolution in `run_custom`'s success path kept only the output-schema flow: a silent custom-tool success (exit 0, empty stdout) rendered as the empty string again — the every-call-identical output shape the engine's stagnation detector kills a loop over (#3303). `silent_success_stamp` was left defined with no caller on the success path.

**2. The #3287 merge broke the rustdoc gate.** The public doc on the MCP toolset's `contracts()` (`crates/stella-mcp/src/toolset.rs:860`) intra-doc-links the private `Self::declared_contract`, which `-D warnings` (`rustdoc::private_intra_doc_links`) rejects. Fixed by dropping the link to plain code font.

## Witness

`custom::tests::silent_success_is_stamped_with_the_input_identity` (already in tree, from #3335):

- **fails on main tip** `c47b5623a` (verified: `cargo test -p stella-tools silent_success_is_stamped` → FAILED, 256/257 otherwise green — the sole test failure on main)
- **passes with this change**

The restored arm lives inside the schema-less path deliberately: a manifest that declares `[output_schema]` never takes it — its empty stdout stays a contract breach reported as `Internal`.

The rustdoc half is witnessed by the gate itself: `make doc-warnings` fails on main tip and exits 0 here.

## Local sweep on this branch

fmt clean · clippy `-D warnings` clean · `make doc-warnings` exit 0 · `cargo test --workspace` exit 0.

Refs #3303

Closes #3359 (the composition defect this repairs is written up there).

### DIFF
--- a/crates/stella-mcp/src/toolset.rs
+++ b/crates/stella-mcp/src/toolset.rs
@@ -857,7 +857,9 @@ impl ToolExecutor for McpToolSet {
     /// `readOnlyHint`/`idempotentHint` preserved verbatim at untrusted
     /// provenance, `destructiveHint` *raising* the grade. The placeholder
     /// tools (needs-auth, resources) resolve through the same declared path
-    /// via [`Self::declared_contract`]'s route miss.
+    /// via `Self::declared_contract`'s route miss (a private helper, so no
+    /// intra-doc link — rustdoc's `-D private-intra-doc-links` would reject
+    /// one from this public doc).
     fn contracts(&self) -> Vec<stella_protocol::ToolContract> {
         let mut contracts = Vec::new();
         if let Some(native) = &self.native {
--- a/crates/stella-tools/src/custom.rs
+++ b/crates/stella-tools/src/custom.rs
@@ -730,6 +730,15 @@ async fn run_custom(tool: &CustomTool, input: &Value, workspace_root: &Path) -> 
         // checker the registry runs (#3285). A breach is the tool's own
         // defect — `Internal`, never model misuse.
         let Some(expected) = &tool.output_schema else {
+            // A silent success (exit 0, empty stdout) renders the
+            // deterministic input-stamped line, never the empty string —
+            // an every-call-identical output is what the stagnation
+            // detector kills a loop over (#3303). A schema-declaring
+            // manifest never takes this arm: its empty stdout is a
+            // contract breach reported below.
+            if output.stdout.is_empty() {
+                return ToolOutput::ok(silent_success_stamp(&tool.name, input));
+            }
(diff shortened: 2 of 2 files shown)


macanderson:
Second main breakage found hiding behind the first: the #3287 merge (c47b5623a) also turned the rustdoc gate red — the public doc on MCP toolset contracts() intra-doc-linked the private Self::declared_contract (crates/stella-mcp/src/toolset.rs:860), which -D warnings rejects. Fixed here in 9d2e87a02 by dropping the link to plain code font. Local sweep on this branch: fmt clean, clippy -D warnings clean, make doc-warnings exit 0, cargo test --workspace exit 0.

Issue #3366, opened 2026-08-16
FileChange's named emitter record_touch does not exist, and a mid-turn re-query emits no ContextRecall


macanderson:
Two findings from the #3243 Phase 3 session, neither fixable in its scope:

## 1. `ToolRegistry::record_touch` exists only in prose

`crates/stella-protocol/src/event.rs` (~line 748, the `FileChange` doc), `crates/stella-tui/src/model/file_state.rs:27`, `crates/stella-tui/src/deck.rs:~1030`, and `crates/stella-cli/src/command_deck/forwarder.rs:188` all name `ToolRegistry::record_touch` as "the single emission point" of `AgentEvent::FileChange` — but no function of that name exists anywhere in the workspace (`rg -n "fn record_touch" crates/` is empty). Either the emitter was renamed and four docs went stale (a "name things for what they are" bug — find the real emission point and chase the name through all four), or the emission point genuinely moved shape, in which case the doc's stronger claim ("the TUI, the audit log and the exported JSON can no longer disagree") needs re-verifying, not just re-naming.

Done: the four references name a function that exists, verified by a grep both ways.

## 2. A mid-turn re-query emits no `ContextRecall` telemetry

`SessionMemory::signal_recall_block` (`crates/stella-cli/src/memory/recall.rs`, #3243 Phase 3) runs a full recall fan-out — provider spend included — and reports nothing: the pre-turn `ContextRecall` event is carried into the turn by the caller (`agent::run_turn`'s `recall_event`), but the re-query happens inside the engine's step loop where the adapter (`crates/stella-cli/src/memory/steering.rs::SessionRequery`) has no event channel. That is the same unmeterable-cost shape #452 closed for pre-turn recall.

Done: a re-query's `Recall` (frames, usage, latency) reaches the turn's event stream — likely by handing `SessionRequery` an `EventSender` clone at construction in `agent::run_turn` / the deck's `run_lead_turn` — and a test asserts one `ContextRecall` event per answered re-query.

Refs #3243, #3357, #3358

Issue #3365, opened 2026-08-16
serve: the pipeline's recall port needs a reverse-RPC design — the one D5 surface a local wire cannot fix


macanderson:
## The problem

The third of #3243 D5's `NoContextRecall` surfaces is `crates/stella-serve/src/pipeline_run.rs` (~line 326): the serve pipeline runs with no recall port. Unlike the fleet and resume surfaces (wired in the Phase 3 PR), this one **cannot** be wired locally: the module's design (its docs, ~lines 50–63) is reverse-RPC — the engine holds no ambient authority, there is no workspace root, no `Store`, and no `SessionMemory` in scope. Model calls and verification already remote back to the host (`RemoteProvider`, `RemoteVerificationRunner`).

## What done looks like

A `RemoteContextRecall` mirroring `RemoteVerificationRunner`'s shape: a reverse-RPC request the host answers by running recall in ITS workspace and returning structured `RecalledFrame`s over the wire. Sequencing:

- New wire messages in the serve protocol (`docs/spec/serve-surface.md` governs; the wire-schema gate will require regenerating `docs/wire/`).
- The host side answers from its own `SessionMemory` / `ContextRecallPort`.
- `PipelinePorts.recall` switches from `NoContextRecall` to the remote port; a host that declines the capability degrades to no frames (the current behavior), never an error.

## Constraints already discovered

- Invariant 3 (zero telemetry egress) is not implicated — frames flow host→engine on a channel the host opened — but frame *content* crossing the wire is prompt material, so it rides the same session channel as everything else, never a side channel.
- `ContextRecallPort::recall(&self, goal: &str)` is goal-only; if #3243 Phase 3's `TurnSignal`-shaped queries reach the pipeline later, design the wire request with room for the signal rather than a second message.

## Verify

A serve integration test where a scripted host answers one recall request and the pipeline's planner prompt carries the returned frame.

Refs #3243, #3357

---

## Redirected by the turn-loop wrapper architecture (2026-08-16)

Design: `doc:turn-loop-wrappers` §4. Umbrella: #3380.

**Hold this one until #3380 settles the shape.** The problem is real; the
proposed solution builds in the direction we are moving away from.

**Why.** This issue would add a fourth pipeline-specific reverse-RPC port
(`RemoteContextRecall`, mirroring `RemoteVerificationRunner`) plus new wire
messages and a `docs/wire/` regeneration. That deepens a private connection
between one wrapper and one surface, at the moment we are deleting the private
[...]

Issue #3358, opened 2026-08-16
steering: skills' and frames' evictions never reach SteeringSet.dropped — the ledger only generalizes the record channel


macanderson:
## The problem

#3349 (PR #3357) generalized the record channel's named drops into `SteeringSet::dropped` — but only for records. Two sources still evict silently, so the plane's ledger under-reports exactly the way the record channel used to before #2709 made its gap observable:

1. **Skills.** `select_skills` truncates to `max_skills` (top-k) with no trace (`crates/stella-core/src/skills.rs`, the `truncate` at the end of selection), and `render_skills_section` omits lower-ranked skills past `SKILLS_SECTION_TOKEN_BUDGET` with only an inline prose marker — the omitted names reach neither stderr nor the ledger. A skill that systematically loses its seat is indistinguishable from one that never matched.
2. **Frames.** The host merge's drop report (`recalled.dropped`, `stella_context::DropReason` — including `RequiredOverBudget`, the loudest case) is consumed inline in `SessionMemory::recalled_frames_reporting` (`crates/stella-cli/src/memory/recall.rs`) for the stderr warning and then discarded: `stella_pipeline::Recall` carries frames + usage but not drops, so by the time `memory/steering.rs` builds the plane's candidates there is nothing left to map into `DroppedCandidate`s.

## What done looks like

- `Recall` (or a sibling channel out of `recalled_frames_reporting`) retains the host's drop list; `crates/stella-cli/src/memory/steering.rs` maps it into `SteeringSet::dropped` with `SteeringSource::Memory`.
- The skills selector reports its top-k cut, and `render_skills_section`'s budget omissions are surfaced as drops (this half likely wants the section budget folded into the plane's shared budget — which is Phase 4 (#3243) behavior-change territory, so sequence the *ledger* part separately from the *budget* part).
- `SteeringSet::by_source` then reports true per-source (selected, dropped) counts for all three context sources.

## Constraints already discovered

- The golden block test (`crates/stella-cli/src/memory/tests/golden_block.rs`) pins the volatile block byte-for-byte; ledger additions must not change rendered bytes.
- The record drop stderr line is single-producer (`record_drop_message` in `recall.rs`); new drop reports should follow that shape, not add a second format.
- `est_tokens` for a dropped frame should come from `frame_recall_line` (`recall.rs`), the same producer the candidate estimates use (#3334 discipline).

## Verify

[...]

Issue #3243, opened 2026-08-14
Merge the four disconnected systems that decide what context, skills, and tools could help into one


macanderson:
## The directive (Mac, 2026-08-13)

> Ensure that context is being proactively and speculatively queried and injected
> when it truly will help or needs to be used to steer (like for a rule or fact).
> What we need is one part of the loop that considers the changes happening to the
> loop and the work and queries for "things" that can help an agent. Those things
> are: context records/memories (which are context record type now — same for
> rules, facts, etc.), MCP servers, skills, and custom tools. Eventually plugins
> join that list. It needs to be managed in one spot, behind a flag that defaults
> on but can be turned off by environment variable or the TOML.

This issue is the survey plus the design. It files no code. **Read §2 before
estimating anything** — four selection engines already exist, and knowing that
changes what the work is.

**Evidence basis.** Everything below was read on `main` at `a4b87649a`
(2026-08-13T14:44:24-07:00). Where a claim depends on a branch, the branch is
named. Line numbers are from that commit.

---

## 1. The one-paragraph finding

Stella already queries four different "what could help here?" planes on every
turn. They run at the same moment, from the same trigger, over corpora that
overlap — and **not one of them knows the other three exist.** They use four
unrelated relevance engines, four unrelated budgets, four unrelated failure
postures, and three unrelated on/off switches. Two of the four are effectively
always-inject (relevance is computed and then ignored), one is reactive-only and
wired on 1 of 6 surfaces, and the fourth advertises its entire corpus on every
request at a measured ~6–9k tokens.

The work is not to build a retrieval system. It is to **collapse four into one
seam, give it the one signal none of them has today (what is actually happening
in this turn), and put it behind one flag.**

---

## 2. What exists today — four planes, four engines

| Plane | Selector | Signal it matches on | When | Budget | Off switch |
|---|---|---|---|---|---|
| **Recall frames** (memories, episodes, facts) | `stella-context` RRF fuse + MMR + evidence gate — `crates/stella-context/src/retrieval.rs:1028-1080` | raw prompt string + ≤4 existing-file path anchors | turn start | 5 frames / 1200 tok | `context.retrieval.ab_recall_rate` (1-in-N), `STELLA_NO_SETTINGS` |
| **Context records** (rules/facts/directives) | `applies_to` glob/keyword match — `crates/stella-core/src/records/select.rs:66-92` | prompt text + ≤32 path tokens | turn start | 16k chars cached / 2k volatile | none |
[...]

PR #3368, opened 2026-08-16
feat(steering): proactive re-query — the plane answers the turn the prompt could not describe (#3243 Phase 3)

macanderson:
## What

#3243 Phase 3 — the directive's actual ask: the steering plane is re-queried **mid-turn**, against what the turn has become instead of the sentence it opened on. Stacked on #3357 (base = `worktree-3349-steering-selectors`), which gave the plane its production implementor; this PR gives it a second question to answer.

## The engine half (`stella-core`)

- **`SteeringRequery`** (`ports.rs`): a new optional port, consulted at the same committed step boundary as `TurnSteering` — after the steering drain, so a user steer landed this step is part of what the next query answers to. Attached via `Engine::with_requery`; children deliberately do not inherit it.
- **`TurnSignal` is finally populated from live turn state**: `driver/loop_evidence.rs::turn_evidence` assembles, per step, the tools the turn has called, the error classes it has seen (`ErrorClass::as_str` tokens), path-shaped tokens from tool-call *inputs* (signal-not-evidence, the `FileChange` caveat verbatim — but it includes a file the turn is about to **create**, the one anchor a prompt cannot contribute), the step index, and `since_last_query` (`TurnState.steps_since_requery`, not checkpointed, same reasoning as `length_continuations`). The evidence shares `turn_start_index`'s window with loop detection, so the two planes cannot disagree about what "this turn" means.
- **A mid-turn recall block is not a turn boundary**: `turn_start_index` now excludes `RECALL_MARKER` messages. Without this, every injected re-query silently reset loop-detection and confident-zero evidence (witness: `loop_evidence::tests::a_mid_turn_recall_block_does_not_reset_the_turn_window`).
- The port's answer is appended verbatim as a marker User message; the host owns hysteresis and dedup (`ports.rs` doc states the contract). Engine-level witnesses: `driver/requery_tests.rs` (the signal's fields, the block landing in history, the counter resetting on an answer).

## The host half (`stella-cli`)

- **`SessionMemory::signal_recall_block`**: the same three selectors, queried against the signal — touched paths join the recall anchors, widen the skill domain scope, and join the record channel's `applies_to` facts. The prompt keeps the lexical query: drift changes *where* the turn is, not what it was asked.
- **`SessionRequery`** (`memory/steering.rs`): implements the port with deterministic hysteresis — the fingerprint is the (order-free) set of touched paths + error classes, starting empty, so an undrifted turn never queries; tool names are deliberately excluded (they churn every step without meaning drift) — and byte-exact dedup seeded from every recall block already in history, mirroring `inject_recall_block`'s any-prior-marker rule.
- **Wired everywhere the CLI drives a raw engine turn**: `agent::run_turn` (REPL, one-shot, goal rounds) and the deck's lead turn.

**The epic's witness** (`memory/tests/steering_selection.rs::a_drifted_turn_recalls_the_skill_its_prompt_could_not`): a two-phase task — the prompt anchors nowhere near a domain-tagged skill; the turn then touches a file in that domain; the re-query surfaces the skill, once, and an undrifted signal buys nothing. Structurally absent on base twice over: no port exists, and every selector fired once against the opening prompt.

## D5 — the surfaces that recalled nothing

Fleet workers (`fleet_cmd.rs`) and resumed runs (`agent/resume.rs`) stop passing `NoContextRecall`: both open the workspace's `SessionMemory` as their `ContextRecallPort`, degrading to no frames exactly as before when there is no memory. A resumed leg's re-recall is recovery, not double-billing — frames were never checkpointed. The third surface (`stella-serve`) is reverse-RPC by design and cannot be wired locally; that design is filed as #3365.

## God files stayed under their ceilings by splitting, not squeezing

- `driver.rs`: the step boundary's host consults (steering drain, soft stop, re-query) moved to `driver/step_boundary.rs` — the ordering contract is documented there.
- `agent.rs`: `begin_execution` / `seed_calibration` moved to `agent/persistence.rs` (re-exported; call sites unchanged).
- `fleet_cmd.rs` crossed 1500 for the first time at the D5 wiring and split its inline `mod tests` to `fleet_cmd/tests.rs` — moved verbatim (dedented), no test deleted.

## Verify

- `cargo test -p stella-core --lib` — 1265 pass (three new witnesses).
- `cargo test -p stella-cli --bin stella` — 1737+ pass, the #3357 byte-golden still green (the pre-turn block's bytes are untouched by this PR).
- `make gate CARGO_SCOPE="-p stella-core -p stella-cli"` green.

## Residue (filed, not left behind)

- #3365 — serve's recall port needs a reverse-RPC design.
- #3366 — `record_touch` doc rot; a mid-turn re-query emits no `ContextRecall` telemetry yet (declared in `signal_recall_block`'s doc).
- #3358 (pre-existing) — skills'/frames' drops still bypass the `SteeringSet` ledger.

Refs #3243, #3357

## Summary by Sourcery

Introduce a steering re-query plane that can proactively recall mid-turn context based on live turn state, with a host-level switch and org-managed ceiling controlling steering, and refactor related engine, memory, and settings code to keep file sizes under limits.

New Features:
- Add a step-boundary SteeringRequery port and TurnSignal-based evidence to allow the engine to re-query context mid-turn and inject recall blocks.
- Implement SessionRequery on the CLI side to drive proactive mid-turn recall using touched paths, error classes, and deterministic hysteresis/dedup.
- Add steering settings and a steering_enabled resolution chain, including an org-managed steering ceiling and STELLA_CONTEXT_STEERING env override, to control whether the steering plane may inject context.

Bug Fixes:
- Ensure recalled-context blocks are excluded from turn boundaries so mid-turn recall does not reset loop detection or confident-zero evidence.

Enhancements:
- Unify context, skill, and record selection behind a single steering plane with shared budgeting and drop reporting, while keeping existing selection behavior byte-identical.
- Extend fleet workers and resumed runs to use SessionMemory recall ports so they benefit from steering-based context recall when memory is available.
- Expose additional helpers and adapters (e.g., frame/skill/record render helpers, environment toggle parsing) to let steering estimate costs and report drops consistently.
- Refactor large modules by extracting agent_engine_config logic into a new settings::engine module, moving persistence helpers into agent::persistence, splitting fleet_cmd tests into a separate file, and adding focused steering/requery test modules in core and CLI.

Documentation:
- Document the semantics and precedence of the steering settings, steering ceiling, env toggle behavior, and the step-boundary host consult ordering for steering, soft stop, and re-query.

Tests:
- Add unit and property tests for the steering budgeter, step-boundary re-query behavior, turn evidence collection, steering selection behavior, golden recall block bytes, and org-managed steering ceiling precedence.
- Add new tests for fleet command behavior in a dedicated module to keep fleet_cmd.rs under the file-size limit.


macanderson:
**Review feedback resolved** (608148bc6).

The Vade finding was real: the steering switch gated only the block renders, so `ContextRecallPort::recall` — the path the pipeline, fleet workers, and resumed runs take — still injected recalled frames with steering off. The gate now lives at `recalled_frames_anchored`, the shared frame-query seam where the A/B control's suppression already sits and for the same reason (#1221): a port that still answers feeds the goal message, the planner, and the witness author behind the switch's back.

Witness: `steering_selection::steering_turned_off_is_frameless_through_the_pipeline_port_too` — verified failing with the seam gate reverted and passing with it. Full `stella-cli` suite (1739) green.

Issue #960, opened 2026-07-30
`stella run` never exits: the turn completes, emits its terminal event, then hangs until killed


macanderson:
## What

`stella run` finishes its turn, emits its terminal event, and then **never exits**. The process sits there until something kills it.

Found while executing #909 (the first measured Terminal-Bench run). It is not benchmark-specific — it reproduces with a chat-class prompt in a bare `debian:bookworm-slim` container with no repo, no tools, no MCP, and no settings.

## Reproduce

```bash
docker run --rm --platform linux/amd64 \
  -v /path/to/target/x86_64-unknown-linux-gnu/release/stella:/usr/local/bin/stella:ro \
  -e OPENROUTER_API_KEY="$OPENROUTER_API_KEY" \
  -w /workspace debian:bookworm-slim \
  /usr/local/bin/stella --model openrouter/z-ai/glm-5.1 --budget 0.05 \
    --output-format stream-json --base-url https://openrouter.ai/api/v1 \
    run "say hi and stop" </dev/null
```

It prints the whole event sequence, ending with

```json
{"type":"complete","model":"openrouter/z-ai/glm-5.1","cost_usd":0.000991576}
```

and then hangs. Killed at 120s in every run. Cost per repro is about **$0.001**.

## Scope — everything tried hangs

| Variant | Result |
|---|---|
| `--output-format text` | hangs |
| `--output-format json` | hangs |
| `--output-format stream-json` | hangs |
| stdin `/dev/null` (immediate EOF) | hangs |
| stdin an open pipe that never closes | hangs |
| no `STELLA_*` env at all | hangs |
| `STELLA_NO_ENV_FILE=1 STELLA_NO_SETTINGS=1 STELLA_DISABLE_REFLECTION=1` | hangs |
| `STELLA_DISABLE_REFLECTION` unset | hangs |

So it is not the output renderer, not the credential/stdin seam, not reflection, and not settings resolution. The turn itself is fine — the terminal event carries a correct answer and a correct cost.

The process still responds to `SIGTERM` and prints its `{"status":"error","reason":"terminated"}` summary, so it is not wedged in uninterruptible state; something in the shutdown path just never completes.

SUT: `0eeb8d4d` (`v0.5.68-68-g0eeb8d4d`, 0.6.10), `x86_64-unknown-linux-gnu`, glibc 2.17.

## Why it matters

[...]

PR #3418, opened 2026-08-17
feat: the engine owns its turn's ending, the run's owner owns the run's (#3379, #3398)

### LINKED ISSUE #3379: The engine owns its own ending: delete the pipeline's private event channels
## Problem

`stella-core::Engine::run_turn` emits its own `Stage { Execute }`, a terminal
`Stage { Complete }`, and a `Complete`. That is correct for one turn. A staged
run with a revise loop runs several turns, so a `Complete` after the first one
would falsely read as "done".

The pipeline's answer today is to take the ending away from the engine. From
`crates/stella-pipeline/src/pipeline.rs`, the module doc's "Event ownership"
section:

> The pipeline is the **single authority** for stage boundaries and the terminal
> event on an outcome-producing run: it gives each `run_turn` a private channel,
> then forwards every event to the consumer *except* the engine's
> `Stage`/`Complete` [...] The pipeline emits `Complete` for success or a
> non-retryable `Error` for terminal failure.

That is a two-way connection. The engine is not just called by the pipeline, it
is edited by it. Three costs follow:

1. **It is a flakiness seam.** A per-turn channel plus a forwarding loop plus a
   drop rule means a dropped event, or a revise round that closes a channel
   twice, produces a run that looks finished and is not. The engine cannot catch
   this, because the engine already did its job correctly.
2. **It blocks the plugin platform.** #3246 wants the pipeline to become a
   plugin. A plugin cannot hold a private channel into the engine, so this
   coupling has to die regardless of what else is decided.
3. **It makes one word mean two things.** `Complete` today means "the turn is
   over" when the engine says it and "the whole job is over" when the pipeline
   says it. Nothing reading the journal can tell which contract it is holding.

## What to build

A one-directional contract.

- `Engine::run_turn` always finishes its turn and always emits `Complete`. That
  event means exactly one thing and never lies: **this turn is over**. It does
  not mean the work is over.
- A wrapper that wants more work **requests another turn**. It does not
  suppress, rewrite, or re-emit the engine's events.
- The wrapper's own "the whole job is over" signal is a **separate event with a
  separate name**. Both appear in the journal, in order.
- The private per-turn channel and the forwarding/filtering code in
  `crates/stella-pipeline/src/pipeline.rs` are deleted, not moved.

## Files

- `crates/stella-pipeline/src/pipeline.rs` — the private channel, the forwarding
  loop, the drop rule, and the module doc's "Event ownership" section (which
  must be rewritten in the same PR, not left describing the deleted design).
- `crates/stella-core/src/driver.rs` — the emit sites; unchanged in behaviour,
  but this is where the contract is stated.
- `crates/stella-protocol/src/event/tags.rs` and
  `crates/stella-protocol/src/event/consumers.rs` — the new wrapper-ending tag
  and its consumer row. Adding a variant without declaring a consumer is an
  `E0004` build error (invariant 10), so this is not optional.
- Anything reading `Complete` today: the TUI, the observatory
  (`crates/stella-observatory/src/journal.rs` whitelists 8 types), the store's
  outcome projection, and arenabench's readers. A consumer that counted one
  `Complete` per run will now see one per turn.

- This is a **journal shape change**, not a rename. Every consumer moves in the
  same PR or the change is not done. #2701 is the epic that owns the ledger.
- arenabench reads the journal and nothing pins its JSON reads against
  `docs/wire/` (#2820), so the bench-side breakage will be silent. Check it by
  hand in this PR.
- The safe-boundary rule is untouched: turns still end between steps.
- Do not "fix" this by teaching the engine about wrappers. The engine must gain
  no knowledge of who called it.

## Definition of done

- No `stella-pipeline` code constructs a private event channel for a `run_turn`,
  and no code path filters an engine event out of the consumer stream.
- A witness test: run a two-round wrapper and assert the journal holds **two**
  engine `Complete` events and **one** wrapper ending, in that order. It fails
  on `main` today, where the two engine events were swallowed and exactly one
  `Complete` exists.
- The wrapper-ending tag has a `Behavioral` or `Surfaced` row in
  `consumers.rs` naming what reads it.
- `crates/stella-pipeline/src/pipeline.rs`'s module doc describes the new
  contract.

## Why this is first

It is the precondition for #3246 and for the wrapper contract, it is the
smallest of the three moves, and it is worth doing on its own even if the
pipeline never leaves the workspace.

Design: `doc:turn-loop-wrappers` §3. Refs #3246, #3274, #2701, #2820.

---

## Findings before implementation (2026-08-16)

Surveyed with four parallel read-only agents at `f6c8544b9`, then spot-checked
by hand. Three things change this issue. **Do not implement it as written.**

### 1. There is no private channel

The issue (and the pipeline's own module doc) call it "a private channel".
It is not one. `Pipeline::filtered_turn_events`
(`crates/stella-pipeline/src/pipeline/execute_stage.rs:409-549`) builds a
**synchronous wrapping `EventSender`** per turn, and the comment at 410-414
says the synchrony is load-bearing:

> The filtered sender is SYNCHRONOUS on purpose: when the outer sender carries
> a durability boundary, a paid `StepUsage` cannot return to the engine before
> append+flush completes. Draining a channel from a spawned forwarder instead
> would let the engine make another paid call before the previous one's
> metering row is durable.

So the fix must **not** replace it with a channel, and "delete the private
channel" is the wrong instruction. What actually gets deleted is one match arm
(`execute_stage.rs:443-448`) that drops `Stage`/`Complete`. The wrapper itself
stays — it also tallies file changes, mutating calls, errored commands and
flip-halt observations, none of which may be lost.

### 2. "Exactly one `complete`, last, only on success" is a PUBLISHED contract

Not a convention — a promise, in three places:

- `docs/wire/README.md:110` states the invariant.
- `crates/stella-pipeline/src/replay.rs:669-689` (`validate_stream`) enforces
  it mechanically.
- `website/content/docs/event-stream-compatibility.mdx:55` explicitly promises
  that `complete` will **not** be repurposed.

Letting the engine's per-turn `Complete` through republishes `complete` as a
per-turn marker. It would then fire several times per run, fire early, and fire
on failed runs — breaking external scripts in the least detectable direction.

### 3. Consumers that latch on `Complete` would corrupt data, not just render early

- `crates/stella-cli/src/arena.rs:285-289` — the arena recorder goes **deaf**
  after the first `Complete`. Every `stella arena` trace would be truncated to
  the first execute turn and stamped `SessionEnd{Completed}` regardless of the
  real outcome. That is data corruption on the CGP conformance surface.
- `crates/stella-tui/src/model.rs:1060-1080` — `Complete` is a terminal settle:
  it closes the proof rail and drops `pending_scope_review` /
  `pending_ask_user` / `pending_hunk_review`. In a wrapped deck run this
  discards a live approval gate mid-run.
- `crates/stella-cli/src/agent/output.rs:163-172` — reintroduces #960
  verbatim: several `complete` lines in the durable stream-json file, each with
  a partial pre-reflection cost.

### The design decision this forces

The goal — the wrapper stops editing the engine's events — is right and
achievable. The question is **which side gets the new name**:

- **(A) As written:** engine's `Complete` reaches consumers per turn; the
  wrapper's ending gets a new name. Breaks the published contract and all
  three consumers above.
- **(B) Recommended:** the engine's per-turn ending gets the **new** name
  (e.g. `turn_complete`); `complete` keeps meaning exactly what it publicly
  means — the run is over, once, last. The wrapper stops filtering (nothing to
  filter: the engine no longer emits `complete`), and the run's owner emits
  `complete` — the pipeline for a wrapped run, the CLI for a raw one.

(B) delivers the one-way contract this issue is about, keeps the wire promise,
and leaves the three consumers correct without touching them. Its cost is that
an unwrapped `stella run` needs someone to emit the run-terminal `complete`,
since the engine no longer does.

**This needs a decision before implementation.** Whichever is chosen, the new
variant needs its row in `crates/stella-protocol/src/event/tags.rs` +
`consumers.rs` (an `E0004` build error otherwise), must join
`structural_diff`'s keep-set (`replay.rs:950-978`) or every existing golden
reports drift, and both golden fixtures must be re-recorded.

### LINKED ISSUE #3398: pipeline: the last two places a wrapper still edits the engine's stream — dropped Stage events, and two emitters of run_complete
## Context

#3379 (PR pending on `events-3379-turn-done-run-done`) split the old
`AgentEvent::Complete` into `turn_complete` (the engine's, per turn) and
`run_complete` (the run's, once). The engine's turn ending is no longer
suppressed by the pipeline.

Two pieces of the "one-directional contract" were deliberately **not** taken in
that change, because each is its own decision. Neither is a silence: both are
recorded here.

## 1. The pipeline still drops the engine's `Stage` events

`crates/stella-pipeline/src/pipeline/execute_stage.rs` — the filter arm now
reads:

```rust
AgentEvent::Stage { .. } => Ok(()),
```

The engine emits its own stage boundaries and the pipeline emits a different
stage vocabulary (triage / plan / execute / witness / verify). Letting both
through would interleave two vocabularies in one stream and would move every
aligned position in `replay::structural_diff`, so the golden fixtures would
report drift about the observability plane rather than about behaviour.

**The decision to make:** which vocabulary survives, or whether stage events
need an origin field so a consumer can tell an engine stage from a wrapper
stage. Until then the pipeline is still editing the engine's stream, which is
the shape #3379 exists to remove.

## 2. There are two emitters of `run_complete`, and one suppresses the other

- The pipeline emits it at its terminal sites
  (`pipeline.rs`, `pipeline/resume_stage.rs`).
- On `--output-format stream-json` the CLI suppresses that one
  (`crates/stella-cli/src/agent/output.rs::pipeline_event_sender`) and emits its
  own after reflection and accounting settle, carrying the all-calls total.

That suppression is correct today and fixes a real bug (#960: two terminal
frames in the durable evidence file, the first carrying a pre-reflection cost).
But it is the same shape #3379 objects to — a caller silencing an event to
replace it with its own.

**The decision to make:** make the run's owner the *only* emitter. Most likely
that means the pipeline never emits `run_complete` and the CLI always does, at
the point the stream closes, with the settled total. That needs an answer for
the failure path too, since today a verification failure ends on a non-retryable
`Error` and the rule is "never both".

## Definition of done

- No wrapper filters any engine event, for any variant.
- Exactly one component emits `run_complete` for a given run, and it is the
  component that owns the run.
- The #960 property still holds: one terminal frame in the durable stream-json
  sink, carrying the all-calls total.

Refs #3379, #3380, #960

macanderson:
## What

The engine owns its turn's ending; whoever owns the run owns the run's. No
wrapper edits the engine's event stream any more.

Closes #3379. Closes #3398.

## The problem

`AgentEvent::Complete` meant "this turn is over" when the engine emitted it
and "the whole job is over" when the pipeline did, and nothing reading the
journal could tell which contract it held. The pipeline kept the two apart by
**silencing the engine**: a per-turn wrapping sender dropped the engine's
`Stage` and `Complete` and the pipeline emitted its own in their place. That is
a two-way connection between the engine and one of its callers.

## What changed

**Two events, renamed rather than aliased** — every call site moved:

- `turn_complete` — one turn ended. The engine always emits it; a wrapped run
  has several, in order.
- `run_complete` — the run ended. Exactly once, from whoever created the event
  stream and will close it.

**One emitter for the run's ending.** The pipeline no longer emits it at all.
A wrapper cannot know whether it is the whole run — and on `stella goal`, which
drives one `Pipeline::run` per round over a single stream, it demonstrably was
not: that put **one terminator per round, several per run**, which
`replay::validate_terminal` calls a violation. The raw goal loop had the
opposite bug and emitted none. Both are fixed, along with resume's bare-turn
arm and the fleet worker, which also had none.

`pipeline_event_sender` is deleted. It existed to suppress the pipeline's
terminal frame so the CLI's could replace it (#960); with one emitter there is
nothing to suppress, and the suppression was itself the shape this issue
objects to.

**Stage events say whose stage they are.** `AgentEvent::Stage` gains a required
`StageScope`: `Turn` for the engine's own phases, `Run` for a wrapper's
vocabulary. With that, the pipeline drops nothing and the contract is complete.

The field is required and **not** `#[serde(default)]`: a default would silently
claim one scope for every historical recording, half of which are the other
one. A pre-field recording decodes through `AgentEvent::Unknown`, which says "I
do not know what this is" instead of guessing.

## The bugs this surfaced

Each consumer that branches on these events was wrong in a way that only bites
once the pipeline stops filtering. None of these are hypothetical:

| Consumer | What would have happened |
|---|---|
| `stella-cli/src/arena.rs` | The recorder **latches** — deaf after the first terminal. A wrapped run's whole contextgraph trace truncated to its first turn and stamped `Completed` regardless of the real outcome. Data corruption, on the CGP conformance surface. |
| `stella-tui/src/model.rs` | Terminal settle closed the proof rail and dropped `pending_scope_review` / `pending_ask_user` / `pending_hunk_review` — discarding a live approval gate mid-run. |
| `stella-tui/src/model.rs` (again) | A stage transition away from a scope review **is** the human's approval of it. A turn-scoped stage arriving while that gate was open would have graduated the proposal into `approved_scope` — the deck asserting consent nobody gave. |
| `replay::validate_stage_ordering` | One legal order over one vocabulary; the engine's terminal stage outranks every wrapper stage, so everything after the first turn would read as an illegal transition. |
| `replay::event_signature` | Both scopes rendered `stage:Execute`, so a golden comparison could not tell an engine phase from a wrapper stage of the same name. |
| `serve::observe::tally` | The wedge counter would have counted engine turns as progress — a run spinning under a stalled wrapper stage reading as advancing, making the detector *worse* than before. |

## Tests

The stream invariants moved rather than being renamed. "A stream terminates
once" now belongs to `run_complete` alone: applied to `turn_complete` it would
flag **every correct wrapped run** as a violation. A new test pins that several
turn endings are not flagged.

The two "no success terminal on a failed verification" tests were rewritten
against `run_complete` — the engine's turn ending now legitimately appears,
because the turn really did finish — and each gained an assertion that the
failure is *announced*, since the original passes on a stream that ends
silently.

Fixtures: the two committed goldens are re-recorded through the sanctioned
`STELLA_REFRESH_GOLDEN` path (scripted ports, no model calls); note that diff
also absorbs an unrelated triage-prompt drift that had accumulated on `main`.
The six hand-written replay fixtures are patched — stage lines gain
`"scope":"run"`, terminal lines become `run_complete`. The forward-compat
fixture is the interesting one: without that rename its old `complete` line
decoded as `Unknown`, which is exactly right for a tag this build no longer
knows.

## Wire break, deliberate and authorized

`complete` → `turn_complete`, `run_complete` added, `Stage` carries a required
new field. `docs/wire/` is regenerated.

This is **not** additive, and the wire-schema gate would normally refuse it. It
is taken now because the stream has no outside consumers yet, and because an
alias would preserve exactly the ambiguity the split exists to remove — the
maintainer's explicit call ("strongly prefer no alias — we have access to all
the call sites").

## Left out on purpose

Nothing. #3398's two remaining halves are both in this PR; it closes with it.

## Verification

- `cargo test --workspace` — 131 suites, 0 failures.
- `cargo clippy --workspace --all-targets` — clean.
- `make guards` — green, including `wire-schema`, `diagnostic-codes`,
  `file-size` (both god files end net-zero), `module-reachability`,
  `lockfile-sync`, `fmt --check`.

Closes #3379
Closes #3398

## Summary by Sourcery

Split turn and run termination responsibilities in the event model, introduce scoped stage events, and update pipelines, drivers, UIs, diagnostics, and replay infrastructure so the engine owns turn endings while stream owners emit a single run-level terminator.

New Features:
- Introduce distinct AgentEvent variants for per-turn and per-run termination, with the engine emitting turn_complete and stream owners emitting run_complete
- Add a StageScope field to stage events to distinguish engine turn phases from wrapper run stages, and propagate scope through consumers and wire formats

Bug Fixes:
- Ensure wrapped runs no longer produce multiple or missing terminal events across pipeline, goal, resume, and fleet paths
- Fix UI and recorder components that previously treated per-turn completion as run completion, preventing premature closure of approvals, traces, and routing state
- Prevent wedge detection and replay validation from miscounting engine stages and turn completions as run-level progress or illegal transitions

Enhancements:
- Simplify event stream ownership by removing pipeline-level suppression and re-emission of terminal events, making the engine and wrappers one-directional producers
- Update diagnostics, transcript export, and deck/session views to report both per-turn and per-run completion with accurate cost semantics
- Strengthen replay and wire-contract invariants to treat run_complete as the sole stream terminator and to include stage scope in event signatures

Documentation:
- Refresh wire and diagnostics documentation to describe StageScope, turn_complete, and run_complete, and to capture the new termination contract

Tests:
- Adjust and extend replay, pipeline, engine, CLI, TUI, and wire-contract tests and fixtures to cover the new StageScope field and the split turn/run completion semantics

## Tests deleted, and why (`check-deleted-tests`)

Every one is a rename or a test removed together with the feature it covered.
None is an unacknowledged drop.

Renamed, because the event they assert on was renamed:

- `complete_not_last_is_flagged` → `run_complete_not_last_is_flagged`
- `two_completes_are_flagged` → `two_run_completes_are_flagged`
  (joined by a new `several_turn_completes_are_not_flagged`, which pins the
  case the old name would have got wrong: several turn endings in one wrapped
  run are the *expected* shape, not a violation)
- `complete_is_unique_and_final_even_when_later_events_arrive` and
  `stream_renderer_persists_reflection_before_one_terminal_complete` are
  reported by the guard but still exist in
  `crates/stella-cli/src/agent/persistence.rs`, rewritten in place against
  `run_complete`.

Renamed because its claim changed:

- `every_engine_turn_ends_in_the_journal_and_the_run_ends_once` →
  `every_engine_turn_ends_in_the_journal_and_the_pipeline_ends_nothing`.
  #3417 landed this test on `main` asserting the pipeline emits the run's
  terminator. Under this PR it deliberately emits none — a wrapper cannot know
  whether it is the whole run — so the test now pins zero run endings beside
  its eight per-turn ones. The multi-turn coverage it was written for is
  unchanged.

Removed with the feature they covered — both tested `pipeline_event_sender`,
which this PR deletes:

- `every_other_format_keeps_the_pipelines_terminal_event`
- `stream_json_suppresses_the_pipelines_own_terminal_event`


macanderson:
## Rebased onto main — reconciled with #3417

**#3417 landed the same #3379 fix on `main` while this was open**, additively: it added `TurnComplete` *beside* `Complete`, with no rename and no `StageScope`. This branch carries the design the maintainer asked for instead — `Complete` renamed to `RunComplete` with no alias, plus the required `StageScope` — so the rename won every conflict and main's independent work was kept.

Four places had solved the same problem twice and would have double-emitted the run's terminator, which `replay::validate_terminal` calls a violation:

| Site | Resolution |
|---|---|
| `stella-serve` | Keeps main's `RunEnding` drop-sealer — the one run owner this branch did not cover. |
| `agent/resume.rs`, `fleet_cmd.rs` | `RunEnding` dropped; both already emit their own terminator explicitly. |
| deck forwarder | Synthesis now fires **only for a stream that ended without one**. That keeps the subsession worker (the one deck lane with no explicit emitter) while leaving the staged lane's own terminator alone — the only one carrying the run's true total, since a pipeline run spends on roles that never surface as a `TurnComplete` and a total reconstructed from those would under-report it. |
| `event/tags.rs` | The duplicate `TurnComplete` consumer row is folded into one naming both behavioral consumers (the diagnostic bridge's retention drain and the journal's transition classification). |

Also in the merge:

- **#3417's multi-turn witness test is rewritten, not dropped.** Under this design the pipeline is a wrapper and emits no run terminator at all, so the test now pins *zero* run endings alongside its eight per-turn ones.
- **`settle_outcome`'s `worker_model_label` and its plumbing are removed.** Once the pipeline stops emitting the run's ending nothing reads the label, and a live parameter that feeds nothing is a false claim about who names the model on the terminal event.
- **#3366's `drop(requery)` is preserved** in `run_turn` — it conflicted textually with this branch's terminator emission and is orthogonal to it.

### One real bug found by the merge

`replay::tests::length_mismatch_reports_trailing_events` asserted that a stream trailing a turn ending is a divergence. It is not: `structural_diff` deliberately **excludes** `TurnComplete` as additive observability, because every golden in the tree predates it. The test passed only while its excluded set and its fixture disagreed. It now trails the run terminator, which the walk does compare, so it proves what its name says.

### Verification

- `cargo test --workspace` — 132 test binaries, **0 failures**.
- `cargo clippy --workspace --all-targets -- -D warnings` — clean.
- `make guards` — green, including `wire-schema` (the committed schemas match the merged types), `god-files`, `module-reachability`, `diagnostic-codes`, `left-behind`, `lockfile-sync`, `fmt --check`.

### One inherited red, deliberately not fixed here

`make file-size` fails on a fresh `main`: `command_deck.rs` is 4014 lines against a recorded ceiling of 4010. **This branch does not cause it** — the file is byte-for-byte the same length on `origin/main` — and AGENTS.md is explicit that a baseline regeneration must land on its own from a fresh `main` rather than be folded into an unrelated PR. Filed as #3424.

Issue #2701, opened 2026-08-10
epic: every emitted signal names its consumer — a ledger + parity tests, the content-free registry shape applied to signals


macanderson:
## Thesis

Signals are produced and not consumed — a repo-wide shape, and every instance so far was found by a bench run paying for it rather than a test going red:

- `flip.json` written, nothing read it — #1536 (fixed; its capture record is still ignored, #2217).
- `verify_done` confirmations tallied but not feeding the halt — fixed only yesterday (036755435, #2661), after `solved_then_timeout` hit 4× on the cert panel at 955s against a 900s ceiling.
- The flip transition still emits **no durable event** (#2661 point 3, open) — we shipped a halt we cannot measure in the field.
- 17 of ~41 `AgentEvent` variants have no behavioral consumer and are invisible to the observatory (which whitelists 8 types, `crates/stella-observatory/src/journal.rs:72-79`) — persisted blobs plus a TUI paint, nothing more.
- The loop detector's input contract ("no timings in `ToolOutput`", `crates/stella-tools/README.md:187-195`) is prose-only — one timestamp in a tool verdict disarms two detection rungs silently.

The event architecture is rich; consumption is ad hoc. Each fix above was whack-a-mole. This epic makes the class **structural**, the way `crates/stella-store/src/content_free.rs` did for telemetry egress: a reviewed ledger where **every emitted signal names its consumer or cites the issue tracking why it doesn't**, enforced by parity tests, so an unconsumed signal is a red test at PR time instead of a field incident.

## The exemplar being copied

`content_free.rs`'s proven shape, element for element:

| content-free registry | this epic's ledger |
|---|---|
| `ContentFreeEncoder` trait + `registered_encoders()` (`content_free.rs:727-736`) | `SIGNAL_CONSUMERS` const, one row per `AgentEvent` variant |
| `DRAIN_FORMATS` with `Guarded` / `NotYetBuilt { issue }` (`:739-757`) | `Behavioral { site }` / `Surfaced { surfaces }` / `RecordedOnly { issue }` |
| `every_built_drain_format_has_a_registered_encoder` (`:878-901`) | totality vs `KNOWN_TYPE_TAGS` (tag_table.rs), both directions |
| negative-control tests (`:940-1010`) — the harness can fail | same, mandatory |

[...]

PR #3427, opened 2026-08-17
fix(stella-core): stage boundaries leave the engine for the run owner (#3416)

### LINKED ISSUE #3416: Stage boundaries are the run owner's vocabulary too: finish #3379 by moving the engine's Stage emit out
## Problem

#3379 made the engine→wrapper connection one-directional for the run's
*ending*: the engine emits `TurnComplete` per turn, the run owner emits
`Complete`, and nothing suppresses either. One engine event is still dropped
at the pipeline's wrapping sender, and it is the last piece of the two-way
coupling that issue set out to remove.

`crates/stella-pipeline/src/pipeline/execute_stage.rs`, in
`Pipeline::filtered_turn_events`:

```rust
AgentEvent::Stage { .. } => Ok(()),
```

The engine emits exactly one stage boundary of its own — `Stage { Execute }`,
at `Engine::drive`'s entry (`crates/stella-core/src/driver/drive.rs`). It
cannot simply be forwarded, because `run_engine_turn` is also what the
**witness stage** uses (`witness_stage.rs`), so an engine `Execute` would
arrive while the run is in `Witness`. `replay::stage_transition_legal`
correctly rejects `Witness -> Execute` (rank 6 -> 5), so every wrapped run
that authored a witness would report a stream violation.

That is a real constraint, not an accident: `StageKind` is the **run's**
vocabulary. Eight of its eleven values (`Triage`, `Research`, `Plan`,
`Witness`, `Verify`, `Verdict`, …) describe stages the engine knows nothing
about. The engine emitting two of them is the same category error the ending
was, and the drop rule is the same workaround.

## What to build

Move stage-boundary emission to the run owner, exactly as #3379 moved the
ending.

- `Engine::drive` stops emitting `AgentEvent::Stage`. The engine gains no new
  knowledge of who called it; it loses a claim it could not support.
- The pipeline already emits `Stage { Execute }` itself
  (`execute_stage.rs::execute_plan`), so the staged path needs no new emit —
  it just stops dropping.
- The **raw** run owners must emit their own `Stage { Execute }` (and, if the
  deck's HUD still wants it, a terminal boundary) so a non-pipeline run's
  journal keeps the shape its consumers expect. #3379's `RunEnding`
  (`crates/stella-core/src/event_sender.rs`) is the precedent and probably the
  place to extend, since it already wraps every raw owner's sender.
- Delete the `AgentEvent::Stage { .. } => Ok(())` arm and the comment above it.

## Files

- `crates/stella-core/src/driver/drive.rs` — the emit site (~line 82).
- `crates/stella-pipeline/src/pipeline/execute_stage.rs` — the drop arm, and
  the module doc that currently says the sender drops exactly two things.
- `crates/stella-core/src/event_sender.rs` — `RunEnding`, if the raw owners'
  boundary is added there.
- The raw owners: `crates/stella-cli/src/agent.rs` (`run_turn`),
  `crates/stella-cli/src/agent/output.rs` (`event_sender_for_raw_run`),
  `crates/stella-cli/src/agent/resume.rs`,
  `crates/stella-cli/src/command_deck/forwarder.rs` (deck + sub-session lanes),
  `crates/stella-cli/src/fleet_cmd.rs`, `crates/stella-serve/src/session.rs`.
- Consumers of `Stage`: `stella-tui` (`model.rs` sets `hud.stage`),
  `stella-observatory` (`journal.rs` selects `'stage'`),
  `crates/stella-cli/src/diag_bridge.rs` (`agent.stage`).

- **`command_deck.rs` and `agent.rs` are god files at their ceiling.** #3379
  could not add a single line to either; the deck's run ending had to be
  synthesized inside `spawn_forwarder` instead. Expect the same here, and note
  that rustfmt's `chain_width` (60) breaks `engine.run_turn_with_sender(...)
  .await` across three lines no matter how short the argument is — so
  "replace one line with one line" is not achievable at that call site.
- `replay::validate_stage_ordering` is the mechanical check that any solution
  has to satisfy; `stage_rank` is the canonical ordering.
- The three raw-run consumers that latch on terminal signals are documented in
  #3379's findings section — re-read them before changing what a raw run emits.
- Adding a variant is not needed here, so no `tags.rs`/`consumers.rs` work,
  but if `Stage`'s posture changes from `Unclassified { issue: "#2703" }` it
  must be updated in the same PR.

## Definition of done

- No code path filters an engine event out of the consumer stream — the last
  arm is gone and `filtered_turn_events` is purely an observer plus the
  concurrent-fan-out preview mute.
- A witness test: a wrapped run that authors a witness produces a stream
  `replay::validate_stream` accepts, and a raw `stella run` still carries the
  stage boundaries its consumers render. It fails on `main` today if the drop
  arm is removed without moving the emit.
- `execute_stage.rs`'s module doc no longer claims a `Stage` drop.

Refs #3379, #3246, #2701.

macanderson:
## What

`StageKind` is the **run owner's** vocabulary. Eight of its eleven values name
stages the engine knows nothing about, and the boundaries the engine emitted
were claims about a caller it cannot see: the same `run_engine_turn` seam drives
the pipeline's *witness* stage, so an engine `Stage(Execute)` arrived while the
run was in `Witness` — a backwards transition `replay::stage_transition_legal`
correctly rejects. The staged path could only survive that by dropping the
event, which is the two-way coupling #3379 set out to remove.

## Correction to the issue

#3416 states the engine emits *exactly one* stage boundary. It emits **two**:

- `Stage(Execute)` at `Engine::drive`'s entry (`driver/drive.rs`), and
- `Stage(Complete)` on the completion path (`driver/completion.rs:190`).

Both were covered by the pipeline's single drop arm, and leaving the second in
place while removing the arm would have put an `Execute -> Complete` transition
ahead of `Verify` in every wrapped run. Both are gone.

Also noted for the record: #3379 has **not** landed — `RunEnding` and
`event_sender_for_raw_run`, which the issue names as precedent, do not exist on
`main`. So the `Complete` half of the drop arm stays, now naming only itself.

## How

Every run owner emits its own boundaries:

| Owner | Where |
|---|---|
| raw one-shot `stella run` | `agent::output::open_raw_turn` + `raw_event_sender_for_run` |
| resumed turn | `agent::resume::drive_resumed_turn` |
| `stella-serve` host | `session::drive_turn` |
| fleet worker | `fleet_cmd` (opener only — see below) |
| goal round loop | `stella_core::goal`, which already owned its `Verdict` boundary |
| every deck lane | synthesized in `spawn_forwarder`; staged lane opts out with `InsightScope::staged()` |

Closing boundaries ride **ahead** of the terminal `Complete` they annotate, via
a new `EventSender::pairing_stage_complete` combinator. The engine emits
`Complete` *inside* the turn, so an emit after the turn returns would land
behind the event some consumers stop at.

The fleet worker gets the opener only: it hands the engine a bare
`UnboundedSender`, so there is no seam to wrap, and it renders lanes from the
ledger rather than a stage HUD. Stated in a comment at the site rather than left
silent.

### The god-file constraint

`agent.rs` and `command_deck.rs` are at their ceilings. `agent.rs` is net
**-4 lines** (the recall block moved into `output::open_raw_turn`, which is also
where the ordering rationale now lives), and `command_deck.rs` is **untouched** —
which is why the deck's boundaries are synthesized on the *receiving* side and
the staged opt-out rides `InsightScope` instead of a seventh `spawn_forwarder`
parameter. `make file-size` passes.

## Witnesses

- `crates/stella-core/tests/engine_emits_no_stage.rs` — a turn emits no `Stage`
  of its own. Observed failing on the parent commit (`[Complete]`), passing here.
- `command_deck/forwarder.rs::a_raw_lane_frames_its_turn_with_its_own_stage_boundaries`
  — `Execute` … `Stage(Complete)` *before* `Complete`.
- `command_deck/forwarder.rs::a_staged_lane_synthesizes_no_boundary_of_its_own`.

No test was deleted.

## Verification

`make gate` green end-to-end (exit 0), including `cargo test --workspace`,
clippy at `-D warnings`, `file-size`, and the self-driving harness.

## Exemplar

The sender combinator follows `tower`'s layer shape — a wrapper that owns one
cross-cutting obligation and leaves the inner sender's contract intact — which
is also how `pipeline_event_sender` in the same module already works.

Refs #3379, #3246, #2701

Closes #3416

## Summary by Sourcery

Delegate stage boundary emission from the engine to run owners and ensure raw and staged runs frame turns with appropriate stage events without violating pipeline stage ordering.

New Features:
- Introduce lane-level stage boundary synthesis in the command deck forwarder so raw lanes frame turns with Execute and Complete while staged lanes opt out.
- Add helper constructors for raw run event senders and turn opening in the agent output module, including a sender combinator that pairs a closing Stage(Complete) ahead of the terminal Complete.

Bug Fixes:
- Remove engine-emitted Stage(Execute) and Stage(Complete) events to prevent invalid or duplicated stage transitions, letting run owners control stage boundaries for their runs.
- Ensure staged pipeline lanes and staged runs do not emit their own Execute or Complete boundaries, avoiding backwards stage transitions rejected by replay validation.

Enhancements:
- Extend InsightScope with configuration for whether a lane opens an Execute stage, defaulting raw lanes to emit and providing a staged variant used by the pipeline.
- Update various run-owner entry points (session host, resumed turns, goal rounds, fleet workers) to emit their own Execute and paired Complete stage boundaries around engine turns.
- Clarify driver and completion path documentation about stage ownership and remove StageKind usage from engine internals.

Tests:
- Add a core test ensuring Engine::drive emits no Stage events so stage boundaries remain solely a run-owner concern.
- Add deck forwarder tests verifying raw lanes frame turns with Execute and Complete boundaries while staged lanes synthesize no boundaries of their own.


macanderson:
## Conflicts resolved — merged `main` in (58b07303d)

Seven files conflicted, all in the same seam: this PR (#3416) moves `Stage` emission out of the engine, while `main` meanwhile landed the ending contract (#3379/#3398/#3417/#3418) which renamed `Complete` → `TurnComplete`, gave `RunComplete` to the run owner, and added a `StageScope` field to `AgentEvent::Stage`.

Both designs survive; neither was dropped.

| File | Resolution |
|---|---|
| `stella-core/src/driver/drive.rs` | engine's `Stage(Execute)` removed (#3416) |
| `stella-core/src/driver/completion.rs` | engine's `Stage(Complete)` removed; `TurnComplete` kept (#3417) |
| `stella-core/src/event_sender.rs` | `pairing_stage_complete` keys on `TurnComplete`, emits `StageScope::Run` |
| `stella-cli/src/agent.rs` | PR's `raw_event_sender_for_run` + main's `requery` wiring |
| `stella-cli/src/command_deck/forwarder.rs` | one loop carrying **both** the `held` slot and the stream-closed `RunComplete` backstop |
| `stella-cli/src/fleet_cmd.rs` | lane's opening `Stage(Execute)` + main's `worker` sender |
| `stella-serve/src/session.rs` | `RunEnding::sealing(… .pairing_stage_complete())` |
| `stella-pipeline/src/pipeline/execute_stage.rs` | drop-arm deleted — there is no engine `Stage` left to silence |

### Scope choice

Owner-emitted boundaries carry `StageScope::Run`, matching the prevailing convention (every other owner site in `stella-pipeline`, `stella-serve`, `stella-cli`). The two in-engine goal-loop rounds keep `Turn`: they are a round's own phases, and their `Verdict` sibling already spells it that way.

`agent/resume.rs`'s test flipped `Turn` → `Run` deliberately — its comment asserted *"the engine emits it about its own turn"*, which is exactly the claim this PR removes. Same event, new emitter, honest scope.

### One thing to look at

`main`'s `output::event_sender_for_raw_run` was **dead on `main` itself** — unreferenced since #3417's merge, and the reason `main` is currently clippy-red (#3433). It scaffolded this very seam, so I removed it in favour of the wired `raw_event_sender_for_run` rather than carrying two constructors for one boundary.

It is *not* a like-for-like swap and that is on purpose: the removed one **sealed** (`RunEnding`), the surviving one **pairs**. Sealing at `run_turn` would double-terminate, because `agent/goal.rs` already calls `persistence::emit_run_complete_on_raw` explicitly. Noted in the code comment. If the intent was for raw runs to seal rather than emit explicitly, that is a real design question and belongs on #3433.

### Verification

- `cargo clippy --workspace --all-targets -- -D warnings` — clean
- `cargo test --workspace` — green (1745 + 740 + … , 0 failed)
- Witness `stella-core/tests/engine_emits_no_stage.rs` updated for the `scope` field and still pins the contract
- Goldens re-recorded via `make record-golden`. **The fixture diff is exactly this and nothing else:**

```diff
-{"type":"stage","name":"execute","scope":"turn"}
-{"type":"stage","name":"complete","scope":"turn"}
-  "event_count": 25
+  "event_count": 23
```

The pipeline's own `stage:run:*` boundaries are untouched — only the engine's turn-scoped pairs left the stream, which is the contract this PR exists to establish.

### Not fixed here (inherited from `main`, filed)

`make file-size` is red against `main` for `stella-protocol/src/event.rs` (1533, ungrandfathered) and `command_deck.rs` (+4 over baseline). Both are **byte-identical on `main`**; this branch adds zero lines to either. Per AGENTS.md I did not fold a baseline regeneration in — filed as #3436.

Issue #3349, opened 2026-08-16
steering: migrate the four selectors onto SteeringPlane — the port has no production implementor yet


macanderson:
## The problem

#3348 (#3243 Phase 2) landed the steering plane's port, types, and budgeter in `crates/stella-core/src/steering.rs`, plus the `context.steering.enabled` switch. What it did **not** do is move the four existing selectors onto the port. So today:

- `SteeringPlane` has **no production implementor**. It is a declared seam with nothing behind it.
- The four selectors still run as they always did, from four unrelated engines with four unrelated budgets: recall frames (`stella-context` RRF+MMR), context records (`records::select`, glob matching), skills (`skills::select_skills`, lexical coverage + domain boost), and tool advertisement (`ToolRegistry::schemas`, unfiltered).
- `pack_to_budget` is exercised only by its own tests.

This is the declared gap, not an oversight — it was split out because reviewing a mechanical four-site refactor in the same diff as a new authority-carrying flag would bury both. This issue is that refactor.

## What "done" looks like

Each selector gains an adapter that maps its existing output into `Vec<SteeringCandidate>` **without changing what it selects**, and one `SteeringPlane` implementation in `stella-cli` fans out to them and packs the result once.

The load-bearing constraint: **this is a pure refactor**. The on-path output must be byte-identical to today's, and that is the whole review. A golden test comparing the assembled volatile block before and after is the deliverable that makes it reviewable — without one, "same output" is a claim rather than evidence.

Files:

- `crates/stella-core/src/steering.rs` — the port and `pack_to_budget` (already landed).
- `crates/stella-cli/src/memory/recall.rs` — `recall_block_reported` / `pipeline_recall_block` assemble the sections today; this is where the plane's single query replaces four.
- `crates/stella-core/src/skills.rs::select_skills` — already the right shape (pure, scored, with a `why`); it needs mapping, not changing.
- `crates/stella-core/src/records/select.rs` — the record channel; note it already reports its drops by name, which is the behavior `SteeringSet::dropped` generalizes.
- `crates/stella-cli/src/discovery.rs` — the tool arm. **Sequence this last**; see below.

## Sequencing note

[...]

PR #3435, opened 2026-08-17
feat(steering): skills' and frames' evictions reach SteeringSet.dropped (#3358)

### LINKED ISSUE #3358: steering: skills' and frames' evictions never reach SteeringSet.dropped — the ledger only generalizes the record channel

macanderson:
## What

`SteeringSet::dropped` now carries all three context sources, not just the record channel #3349 generalized it for.

- **Skills / top-k.** `select_skills_reporting` keeps the tail `select_skills`'s `truncate` used to discard; `select_skills` is now a thin `.selected` over it, so there is one selection pass, not two.
- **Skills / section budget.** `render_skills_section`'s own cut is extracted as `skills::section_fit` and read by both the renderer and the new `adapt::skill_drops` — the #3334 single-producer discipline, so the ledger cannot disagree with what the prompt actually carried.
- **Frames.** The recall host's merge report, which `recalled_frames_anchored` consumed for its stderr warning and then dropped on the floor, rides back on a `RecalledFrames` sibling channel and is mapped to `DroppedCandidate`s at the one site that has all three crates in scope — `Recall` stays a `stella-pipeline` type with no `stella-context` dependency.

`SteeringSet::by_source` therefore reports true `(selected, dropped)` counts per source.

## What deliberately did not change

**The rendered bytes.** Source drops are additive to `dropped` and are never withheld from `selected`, so the section renderer still makes its own cut — the golden block test passes untouched. One consequence is stated in the code rather than hidden: a skill omitted by the section budget appears in **both** `selected` and `dropped`, because the plane authorized it and the renderer then left it out, and those are two true facts. Collapsing them means the plane owning the section budget, which is Phase 4 behavior change (#3243), sequenced apart exactly as the issue asks.

**The stderr report** stays record-only (`report_record_drops`), per the single-producer constraint in the issue. Surfacing skill/frame drops to a human is filed as follow-up rather than invented here.

## Witnesses

One per source, all three failing on base (the tail, the fit function, and the drop channel do not exist there):

- `stella-core::steering::tests::a_skill_cut_by_top_k_is_named_in_the_ledger`
- `stella-core::steering::tests::a_skill_omitted_by_the_section_budget_is_named_without_changing_the_bytes` — also asserts the rendered section is unchanged
- `stella-cli::memory::tests::steering_selection::a_frame_the_host_merge_evicted_reaches_the_ledger`

An honest note on the third: a local one-provider recall cannot make the *host merge* drop anything, because the provider applies the same budget first and leaves the merge nothing to evict. So its kept side is a real recall and its evicted side is a merge drop of the shape the CGP composition conformance run in `contextgraph::tests` already admits. The merge's own report is exercised there; this witness covers the link that used to end at stderr.

## Verification

`cargo test -p stella-core -p stella-cli` (1745 lib + every integration target, 0 failures), `cargo clippy --all-targets -D warnings`, `cargo doc --no-deps`, `make guards`. The two `check-file-size` lines are inherited base drift on `event.rs` and `command_deck.rs`, untouched by this change.

Refs #3243, #3349
Closes #3358

## Summary by Sourcery

Track and report evicted skills and frames from all context sources in the steering ledger without changing rendered output or stderr behavior.

New Features:
- Expose a consolidated SkillSelection structure that retains both selected skills and those cut by top-k for reporting.
- Surface host merge frame evictions as DroppedCandidate entries alongside recalled frames in the steering ledger.

Enhancements:
- Extend steering plane gathering to accept explicit frame-drop and skill-selection inputs and to populate SteeringSet::dropped from skills, frames, and records.
- Factor out the skills section token-fit calculation so both the renderer and steering adapters share a single definition of which skills the section budget omits.

Tests:
- Add steering and CLI tests that assert skills dropped by top-k or the section budget and frames evicted by host merge are all reflected in SteeringSet::dropped and per-source counts.

### DIFF
--- a/crates/stella-cli/src/memory/recall.rs
+++ b/crates/stella-cli/src/memory/recall.rs
@@ -168,10 +168,13 @@ impl SessionMemory {
         if self.ab_suppressed || !self.steering_enabled {
             return RecalledBlock::default();
         }
-        let recall = self.recalled_frames(prompt).await;
+        let RecalledFrames {
+            recall,
+            dropped: frame_drops,
+        } = self.recalled_frames(prompt).await;
 
         let all_skills = self.load_skills();
-        let selected = skills::select_skills(
+        let selected = skills::select_skills_reporting(
             &all_skills,
             prompt,
             &self.active_domains(prompt),
@@ -189,14 +192,20 @@ impl SessionMemory {
             prompt,
             ..Default::default()
         };
-        let set = query_gathered_plane(&signal, &recall.frames, &selected, record.as_ref());
+        let set = query_gathered_plane(
+            &signal,
+            &recall.frames,
+            &frame_drops,
+            &selected,
+            record.as_ref(),
+        );
         report_record_drops(&set, |message| eprintln!("  {} {message}", "!".yellow()));
 
         let mut sections: Vec<String> = Vec::new();
         if let Some(section) = render_context_section(&kept_frames(&recall.frames, &set)) {
             sections.push(section);
         }
-        let kept = kept_skills(&selected, &set);
+        let kept = kept_skills(&selected.selected, &set);
         if !kept.is_empty() {
             sections.push(skills::render_skills_section(&kept));
         }
@@ -248,7 +257,7 @@ impl SessionMemory {
             return None;
         }
         let all_skills = self.load_skills();
-        let selected = skills::select_skills(
+        let selected = skills::select_skills_reporting(
             &all_skills,
             prompt,
             &self.active_domains(prompt),
@@ -260,11 +269,11 @@ impl SessionMemory {
             prompt,
             ..Default::default()
         };
-        let set = query_gathered_plane(&signal, &[], &selected, record.as_ref());
+        let set = query_gathered_plane(&signal, &[], &[], &selected, record.as_ref());
         report_record_drops(&set, |message| eprintln!("  {} {message}", "!".yellow()));
 
         let mut sections: Vec<String> = Vec::new();
-        let kept = kept_skills(&selected, &set);
+        let kept = kept_skills(&selected.selected, &set);
         if !kept.is_empty() {
             sections.push(skills::render_skills_section(&kept));
         }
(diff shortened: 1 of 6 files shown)


macanderson:
Residue filed as #3437 — the drop report is now complete on the ledger but still record-only on stderr, plus a decision owed on the two recall-side frame filters (citation-label and quarantine) this change did not touch.

Issue #3413, opened 2026-08-17
FileChange has no producer outside candidate adoption: engine-only turns report no file changes, and Read has no emitter


macanderson:
Found while fixing #3366, out of scope there (that issue's DoD was the four stale names and the re-query's telemetry).

## What is true today

`AgentEvent::FileChange` has exactly one production emitter in the workspace: `Pipeline::deliver_winner` (`crates/stella-pipeline/src/pipeline/delivery.rs:303`), which emits one event per `AdoptedChange` when a winning candidate is adopted. Verify both ways:

```
rg -n "AgentEvent::FileChange \{" crates/    # every other hit is a match arm or a test
rg -n "fn record_touch|fn classify_file_op" crates/   # empty
```

Two consequences nothing currently tracks:

1. **An engine-only turn reports no file changes at all.** `agent::run_turn` (plain `stella` chat/run, and the deck's lead turn) never goes through candidate adoption, so no `FileChange` reaches its stream. The TUI Files tab, `stella export`, and the audit log are therefore empty for every non-pipeline turn that changed files. This is not a regression introduced by #3366 — the tool purge (12 built-ins, no file-writing tool) removed the producer and the docs simply kept naming it — but it is an observability hole that currently looks like "the turn changed nothing".
2. **`FileChangeKind::Read` has no live producer.** It survives only so recorded journals replay. If that is the intended end state, the variant's doc should say so permanently; if not, something must emit it.

## Why it is not simply "emit from the tools"

`ToolRegistry::new` registers only the task-board, scratch-state, sub-agent and `get_environment` tools (`crates/stella-tools/src/registry.rs`), and `crates/stella-tools/src/catalog.rs` has no `read`/`write`/`edit`/`bash` rows. There is no path-naming tool left to hang a touch on. `docs/spec/remote-sandboxes.md` §4.2 sketches the host-side emitter a future file-writing remote tool would need — that section is now explicitly marked prospective.

Note also that `crates/stella-protocol/src/event.rs`'s `FileChange` doc now names a `stella-pipeline` symbol. Prose only, no dependency, but if the emitter moves again this is the citation to chase.

## Constraints

- Invariant #10: whatever is decided, the `consumers.rs` row for `FileChange` must keep declaring a coherent posture.
[...]

Issue #2907, opened 2026-08-12
`files_touched` and `wrote_files` are blind to every edit an isolated pipeline candidate makes


macanderson:
## Problem

For a `deck-pipeline` execution that ran in an isolated candidate workspace, `store.db`'s `files_touched` records only the reads that happened *outside* the candidate — the research stage's — and none of the edits the execute stage actually made. `execution_reflection.wrote_files` reads `0` for the same reason.

The "what did this session change" table is blind to everything the session changed.

## Evidence

From a real exported session (`ses-1786486631485-20707`, execution 103):

`tool_calls` shows two successful `edit_file` calls on `arenabench/.gitignore` (seq 30 and 34) plus a `read_file` on it at seq 13:

```json
{"seq":13,"name":"read_file","ok":1,"args_json":"{\"path\":\"arenabench/.gitignore\",...}"}
{"seq":34,"name":"edit_file","ok":1,"args_json":"{\"path\":\"arenabench/.gitignore\",\"old_string\":\"...\",\"new_string\":\"...arenabench-cloud/**\"}"}
```

`files_touched` for the same execution records the read and nothing else:

```json
{"execution_id":103,"path":"arenabench/.gitignore","ops":"R","lines_added":0,"lines_removed":0,
 "events":"[{\"event\":\"R\",\"reason\":\"Check existing ignore rules for match results\",...}]"}
```

`execution_reflection` for 103: `"wrote_files": 0`.

## It is not the ledger

Ruled out with a direct repro rather than assumed — `stella-tools`' ledger handles this sequence correctly. A registry that reads then edits one file records `RU` with the right line delta:

```
LEDGER: [("target.txt", "RU")]
REC path=target.txt ops=[Read, Update] +1 -0 events=2
```

So the loss is above the registry.

## Mechanism

An isolated candidate gets its **own** `ToolRegistry`, rooted at the snapshot (`crates/stella-cli/src/candidate_ws.rs`, in `create_workspace` — `crate::agent::new_tool_registry(ws_root.clone(), ...)`). Every execute-stage edit lands on that registry's ledger.

`record_execution_end` persists the **session** registry's ledger (`crates/stella-cli/src/agent/persistence.rs::file_touch_rows`, called from `crates/stella-cli/src/command_deck/pipeline_turn.rs`). Nothing ever merges the candidate's ledger into it.

The session registry sees the research stage's reads (those run outside the candidate) — which is exactly the surviving row — and nothing after.

[...]

PR #3449, opened 2026-08-17
feat(stella-cli): adopted changes reach files_touched, the durable half of #2907

### LINKED ISSUE #3419: files_touched is never written: record_files_touched has no production caller and attribute_adopted has no CLI override
Found while fixing #3413, out of scope there (that issue's DoD was the FileChange producer decision and its witness).

## What is true today

`Store::record_files_touched` (`crates/stella-store/src/lib.rs:902`) has **no production caller anywhere in the workspace**. Every hit is a test:

```
rg -n "record_files_touched" crates/ | rg -v "stella-store/src|tests"   # empty
```

The seam that was supposed to call it is `WorkspacePort::attribute_adopted` (`crates/stella-pipeline/src/ports/workspace.rs:279`), whose trait default is a no-op `{}`. The pipeline calls it in three places (`pipeline/delivery.rs:294`, `pipeline/execute_stage.rs:204` and `:309`), but **`stella-cli`'s `CandidateWorkspace` never overrides it**:

```
rg -n "attribute_adopted" crates/ | rg -v stella-pipeline   # empty
```

So the `files_touched` table is empty in production for every run, isolated and shared-tree alike. `stella export`'s `files_touched` dump (`crates/stella-cli/src/export.rs:444`) and `Store::prune`'s cascade over that table therefore operate on a table nothing writes.

This is the durable-ledger half of exactly what #3413 fixed on the event-stream half. #2907's stated goal was that "the stream and the durable record are two projections of one adoption"; the stream projection landed, the durable one did not.

## Why it is not simply "implement the method"

`attribute_adopted` takes `&[AdoptedChange]` and no execution id. `record_files_touched` needs an `execution_id`. The `CandidateWorkspace` in `stella-cli` would have to be constructed with, or handed, the execution the adoption belongs to — check how `crates/stella-cli/src/candidate_ws.rs` is built and whether the execution id is in scope there at all. That plumbing is the actual work, and is why this is a separate issue rather than a line.

Note also that `FileTouchRow` (`crates/stella-store/src/lib.rs:265`) carries `events_json`, which `AdoptedChange` has no analogue for — decide whether it gets the per-file diff, the empty array, or something else.

## Definition of done

- `stella-cli`'s `CandidateWorkspace` overrides `attribute_adopted` and writes `FileTouchRow`s through `Store::record_files_touched`.
- A witness test: a pipeline run that adopts a candidate leaves rows in `files_touched` keyed to the run's execution id; the test fails on `main`.
- Decide and document whether the shared-tree producer added in #3413 (`turn_files` / `WorkJournal::snapshot_worktree`) should also write these rows, or whether the durable ledger is deliberately adoption-only. If adoption-only, say so in `attribute_adopted`'s doc, which currently explains the isolated/shared split but not this consequence.

Refs #3413, #2907

macanderson:
Closes #3419

## What was true

`Store::record_files_touched` had **no production caller anywhere in the workspace**, and `CandidateWorkspace::attribute_adopted`'s trait default is a no-op that `stella-cli`'s port never overrode. So `files_touched` was empty in production for every run that ever adopted a candidate — while the `FileChange` events describing the very same rows rode the session's stream. `stella export`'s files dump and `Store::prune`'s cascade both operated on a table nothing writes.

#2907 asked for *"the stream and the durable record are two projections of one adoption"*. #3413 landed the stream half; this is the other one.

## The change

- **`candidate_ws/adoption_ledger.rs`** (new) maps `AdoptedChange` → `FileTouchRow` and writes it.
  - `ops` is **CRUD letters**, fixed by the reader rather than by preference: `Store::execution_rollup` counts a run's written files with `ops LIKE '%C%' OR ops LIKE '%U%' OR ops LIKE '%D%'`, so a row spelled `Modified` — or `M`, as some test fixtures use — is invisible to it. `FileTouchRow`'s own doc gives `"RUD"` for the same reason.
  - `events_json` carries **one** audit entry: an adoption *is* one event. The per-file diff is deliberately excluded — it rides the `FileChange` event, and `files_touched` is a per-execution index, not a content store.
- **Plumbing** (the part the issue flagged as the actual work): the store and execution id reach the candidate through the existing `SessionPlane` seam, from the `(Arc<Store>, i64)` pair `begin_pipeline_execution` already returns. `None` keeps the previous no-op, correct for a session with no store open.
- **A write failure never fails the adoption.** This is observability, not evidence (#2882); git has already moved the bytes, and turning a telemetry error into a delivery error would let the record's absence destroy the thing it records.

## The shared-tree decision (third DoD item)

Answered in `attribute_adopted`'s doc: the durable ledger is **adoption-only, by decision**.

Adoption knows *who* changed the file — a candidate's edits became the user's tree's, at one instant, by one `git apply`. The shared-tree producer measures *the turn, not the actor*, and says so; anything that ran during the turn, including the user's own editor, is inside its diff. Folding that into a table keyed by `execution_id` would attribute a teammate's concurrent save to the agent's execution, permanently and unseparably. An event stream can carry a measurement; a row keyed to an execution reads as attribution.

## God-file discipline

`agent.rs` is grandfathered and closed to growth. The builder chain spelled out at the call site cost it 5 lines, so the composition moved into `agent/tools.rs` as `SessionPlane::for_run`. **`agent.rs` ends at exactly 1799 lines — its ceiling — and no baseline moved.**

## Witness

`an_adoption_lands_in_files_touched_keyed_to_the_run_execution` asserts through `export_all_json`, deliberately: that is the surface the issue names as reading an unwritten table, so a pass here is a pass for `stella export` rather than for a private helper. It checks the row count, the path, the **execution id**, and the `C` letter.

Flip demonstrated by neutering the override back to the trait default: fails with `got []` (0 rows), passes with it restored.

## Verification (by exit code)

```
cargo clippy --workspace --all-targets -- -D warnings   0
cargo fmt --all --check                                 0
cargo test -p stella-cli --bins                         0   (1752 passed)
cargo test -p stella-pipeline                           0
check-file-size / god-files / left-behind               0
check-typed-errors / module-reachability                0
```

Refs #3413, #2907

## Summary by Sourcery

Persist adopted file changes into the durable files_touched ledger and wire execution attribution through the session pipeline.

New Features:
- Introduce an adoption ledger that maps AdoptedChange entries into files_touched rows keyed by execution_id.
- Allow candidate workspaces to attribute adoptions to a specific store execution via adoption_ledger configuration.

Enhancements:
- Extend SessionPlane and workspace port wiring so pipeline runs pass store and execution context to candidate workspaces.
- Clarify CandidateWorkspace documentation around adoption-only durability versus shared-tree measurement semantics.

Tests:
- Add integration and unit tests ensuring adopted files are exported from files_touched with correct path, ops letter, and execution_id, and validating the ledger row format and audit entries.

### DIFF
--- a/crates/stella-cli/src/agent.rs
+++ b/crates/stella-cli/src/agent.rs
@@ -366,7 +366,7 @@ async fn run_pipeline_one_shot(
             cfg,
             active_rules.clone(),
             mcp.clone(),
-            SessionPlane::new(tx.clone()).with_sub_agents(registry.sub_agent_dispatcher()),
+            SessionPlane::for_run(tx.clone(), &registry, execution.clone()),
         )?;
 
         let breaker = CircuitBreaker::new(Box::new(SystemClock::new()));
--- a/crates/stella-cli/src/agent/tools.rs
+++ b/crates/stella-cli/src/agent/tools.rs
@@ -351,6 +351,10 @@ pub(crate) struct SessionPlane {
     /// as unavailable — which, with no command-executing built-in, leaves a
     /// worker there unable to act at all.
     sub_agents: Option<Arc<dyn stella_core::subagent::SubAgentDispatcher>>,
+    /// The execution adopted changes are attributed to (#3419), and the store
+    /// holding it. `None` when the session opened no store — the only case in
+    /// which an empty `files_touched` is the right answer.
+    adoption: Option<(Arc<stella_store::Store>, i64)>,
 }
 
 impl SessionPlane {
@@ -360,7 +364,36 @@ impl SessionPlane {
         Self {
             events: Some(events),
             sub_agents: None,
+            adoption: None,
         }
+    }
+
+    /// Attribute this session's adoptions to the run's `executions` row
+    /// (#3419). Pass the pair `persistence::begin_pipeline_execution` returned
+    /// — a driver that opened no store passes nothing and keeps the no-op.
+    pub(crate) fn with_adoption(
+        mut self,
+        execution: Option<(Arc<stella_store::Store>, i64)>,
+    ) -> Self {
+        self.adoption = execution;
+        self
+    }
+
+    /// The complete plane a pipeline run needs: its event sink, the session's
+    /// sub-agent runner, and the execution its adoptions are attributed to.
+    ///
+    /// Composed here rather than at the call site because that call site is in
+    /// `agent.rs`, a grandfathered god file closed to growth — the builder
+    /// chain spelled out there costs it lines it does not have, and this is
+    /// the sibling module the plane already lives in.
+    pub(crate) fn for_run(
+        events: stella_core::EventSender,
+        registry: &stella_tools::ToolRegistry,
+        execution: Option<(Arc<stella_store::Store>, i64)>,
+    ) -> Self {
+        Self::new(events)
+            .with_sub_agents(registry.sub_agent_dispatcher())
+            .with_adoption(execution)
     }
 
     /// Share the session's sub-agent runner with every candidate workspace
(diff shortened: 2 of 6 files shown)


macanderson:
CI red here is inherited, not from this branch.

The only failing step is `cargo doc -D warnings`, and its five errors are all `unresolved link to `AgentEvent::{FileChange,MediaComplete,MediaProgress}`` in `crates/stella-protocol/src/event/payload.rs` — a break that exists on `main` independently of this PR and is fixed by #3457.

Verified on this branch: `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps --document-private-items --keep-going` produces exactly those five and nothing from the files this PR touches. Every other check passes.

This should go green once #3457 lands and this branch picks up main.

PR #3450, opened 2026-08-17
fix(stella-protocol): qualify payload.rs's AgentEvent doc links, unbreaking main

macanderson:
## The breakage

`main`'s `cargo doc -D warnings` step is red at f1848b855 with five errors, all in `crates/stella-protocol/src/event/payload.rs`:

```
error: unresolved link to `AgentEvent::FileChange`   (x3)
error: unresolved link to `AgentEvent::MediaComplete`
```

The #3440 split moved these doc comments out of `event.rs`, where a bare ``[`AgentEvent::FileChange`]`` resolved against the enclosing module. In the new child module nothing named `AgentEvent` is in scope, so every one of them fails.

## The fix

Qualify them as ``super::AgentEvent::<Variant>`` — the same form the module doc two lines above already uses (``[`AgentEvent`](super::AgentEvent)``). Doc comments only; no code, no wire change.

## Why nothing pre-merge caught it

The CI step passes `--document-private-items`, because rustdoc only checks links in items it actually documents (#2336). A local `cargo doc --no-deps` without that flag reports clean on the same tree — which is exactly how it reached `main`, and worth knowing before anyone "verifies" a doc fix with the shorter command.

## Verification

By exit status, on this branch:

- `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps --document-private-items --keep-going` → 0 (the exact CI invocation; 101 on `main` without this change)
- `cargo clippy --workspace --all-targets -- -D warnings` → 0
- `cargo fmt --all -- --check` → 0

Refs #3440, #3439

## Summary by Sourcery

Documentation:
- Qualify references to AgentEvent variants in payload.rs so rustdoc resolves them correctly.

### DIFF
--- a/crates/stella-protocol/src/event/payload.rs
+++ b/crates/stella-protocol/src/event/payload.rs
@@ -19,7 +19,7 @@ use serde::{Deserialize, Serialize};
 
 use crate::ladder::LadderSnapshot;
 
-/// What happened to a file in a [`AgentEvent::FileChange`] event.
+/// What happened to a file in a [`AgentEvent::FileChange`](super::AgentEvent::FileChange) event.
 ///
 /// Both live producers measure a tree against a tree, so every kind emitted
 /// today is a mutation — see [`Self::Read`] for the one that is not, and why
@@ -36,7 +36,7 @@ pub enum FileChangeKind {
     /// again. Neither surviving producer can: both diff a tree against a tree,
     /// and a read leaves no trace in a tree to diff. Re-acquiring one would
     /// mean going back to declaring file operations from tool inputs, which is
-    /// the defect [`AgentEvent::FileChange`] documents rather than a capability
+    /// the defect [`AgentEvent::FileChange`](super::AgentEvent::FileChange) documents rather than a capability
     /// worth restoring.
     ///
     /// It stays in the kind space because journals recorded before the purge
@@ -59,7 +59,7 @@ impl FileChangeKind {
     ///
     /// A `true` here is not a licence to claim the *agent* changed the file:
     /// the shared-tree producer measures the turn, not the actor (see
-    /// [`AgentEvent::FileChange`]).
+    /// [`AgentEvent::FileChange`](super::AgentEvent::FileChange)).
     #[must_use]
     pub fn is_mutation(self) -> bool {
         !matches!(self, FileChangeKind::Read)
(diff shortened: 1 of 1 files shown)


macanderson:
## Pushed a fix (078867aff) — the two red checks share one cause, and it is not just a stale artifact

The unresolved links are real and this PR is fixing a real break (mine, from #3440 — I dropped `use super::AgentEvent` as an unused import without noticing the doc links needed it). But the qualified-link approach has a substantive problem underneath the red `check-wire-schema`.

**These doc comments are not private prose.** schemars exports them verbatim as the `description` fields of `docs/wire/agentevent.schema.json` and `docs/wire/agentevent.d.ts`. So the qualified form was publishing a Rust module path into the wire contract that external `--output-format stream-json` consumers read:

```diff
-      "description": "What happened to a file in a [`AgentEvent::FileChange`] event. …"
+      "description": "What happened to a file in a [`AgentEvent::FileChange`](super::AgentEvent::FileChange) event. …"
```

```diff
- * What happened to a file in a [`AgentEvent::FileChange`] event.
+ * What happened to a file in a [`AgentEvent::FileChange`](super::AgentEvent::FileChange) event.
```

`super::` means nothing to a TypeScript consumer. Regenerating the artifacts would have turned the gate green and shipped that text — the guard was right to object, and its message says as much: *"read the diff as a wire change, not as noise."*

## What I pushed instead

```rust
#[cfg(doc)]
use super::AgentEvent;
```

rustdoc sets `cfg(doc)`, so it sees the import and resolves all five links; no other build sees it, so `unused_imports` has nothing to fire on — which is precisely why the import got dropped in the split and the links broke. The doc text goes back to what it was, so the wire contract is untouched.

One follow-on: with `AgentEvent` in scope, the module doc's own `[`AgentEvent`](super::AgentEvent)` became a `rustdoc::redundant_explicit_links` error, so it is now plain `[`AgentEvent`]`.

## Verification

| check | before | after |
|---|---|---|
| `make doc-warnings` | 5 unresolved links | **clean** |
| `make wire-schema` | FAILED | **OK — `docs/wire/` regenerates to no diff at all** |
| `make guards` | — | green |
| `clippy --workspace --all-targets -D warnings` | — | clean |
| `cargo test --workspace` | — | green |

That second row is the load-bearing one: the fix is correct *because* it needs no wire regeneration. Net diff is 12 lines in one file, and `git status` shows nothing under `docs/wire/`.

Left the original commit in place rather than force-pushing over it, so the two approaches stay comparable in the history — squash-merge will collapse them. Happy to reword the commit if you would rather it read as a single change.

Issue #3436, opened 2026-08-17
main is red on file-size: stella-protocol/src/event.rs is 1533 lines (ungrandfathered) and command_deck.rs exceeds its baseline by 4


macanderson:
## Problem

`make file-size` fails against `origin/main` (1016925c9) for two files, and neither is caused by the branch that observes it — both are byte-identical on `main` itself:

```
crates/stella-protocol/src/event.rs is 1533 lines, over the 1500-line limit
crates/stella-cli/src/command_deck.rs grew to 4014 lines, over its baseline ceiling of 4010 (+4)
```

Verify:

```sh
git show origin/main:crates/stella-protocol/src/event.rs   | wc -l   # 1533
git show origin/main:crates/stella-cli/src/command_deck.rs | wc -l   # 4014
rg -n 'command_deck.rs' scripts/file-size-baseline.txt               # 4010 …
```

This is the shared-cell skew AGENTS.md § "God files" describes: `scripts/file-size-baseline.txt` is one cell every growing PR writes, so two independently-correct branches compose into a red tree. The ratchet judges a change against its base, so a PR branched before the growth still passes on CI (base = the merge commit's first parent) while every local `make gate` from an older merge base reports it. That asymmetry is why it has gone unnoticed.

Noticed while resolving the conflicts on #3427 (branch `worktree-fix-3416-stage-emit`), which adds zero lines to either file and is still shown the failure locally.

## The two halves are different problems

1. **`command_deck.rs` (+4 over baseline).** A grandfathered god file that grew past its recorded ceiling. Either the 4 lines move into a submodule, or — if genuinely irreducible — `make file-size-update` lands as its own reviewable baseline diff from a fresh `main`.

2. **`event.rs` (1533, not grandfathered).** This is the harder one and it must NOT get a baseline entry: the baseline only covers files predating the guard, and adding one is the expedient CLAUDE.md forbids. `AgentEvent` has been growing steadily (`StageScope` in #3398, `TurnComplete`/`RunComplete` in #3417/#3418). It wants splitting — the variant enum and its `type_tag()`/consumers tables are the natural seam, and `crates/stella-protocol/src/event/tags.rs` and `event/consumers.rs` already exist as siblings to grow into.

## Constraints

- Land as its own PR from a fresh `main`. AGENTS.md is explicit: do not fold a baseline regeneration into an unrelated PR — that is how this went red three times before.
[...]

PR #3451, opened 2026-08-17
fix(stella-protocol): bring AgentEvent into doc scope, main is red on rustdoc

macanderson:
## `main` is red

`make gate` fails at `doc-warnings` on `main` (`b46eb1ac1`):

```
error: unresolved link to `AgentEvent::FileChange`     crates/stella-protocol/src/event/payload.rs:22, :39, :62
error: unresolved link to `AgentEvent::MediaComplete`  crates/stella-protocol/src/event/payload.rs:197
error: unresolved link to `AgentEvent::MediaProgress`  crates/stella-protocol/src/event/payload.rs:212
make: *** [doc-warnings] Error 101
```

#3440 split the leaf payload types out of `event.rs` into `event/payload.rs`.
The doc comments moved verbatim and are correct as prose, but `AgentEvent` is
declared in `event.rs` and is not in scope in the new child module, so five
intra-doc links that resolved inline stopped resolving after the move.

Being a `doc-warnings` failure, it also **masks every gate step after it** —
which is how the `wire-schema` interaction below stayed invisible.

## Fix

`#[cfg(doc)] use super::AgentEvent;`, plus dropping the module header's
now-redundant explicit link target.

## Why not just qualify the links

I tried that first — `[`AgentEvent::X`](super::AgentEvent::X)` on all five.
It fixes rustdoc, and then fails `wire-schema`:

These doc comments are **exported verbatim into `docs/wire/agentevent.d.ts`**.
Qualifying pushed five Rust module paths into the public TypeScript wire
contract, which contains no `super::` anywhere today:

```diff
-   * [`AgentEvent::MediaProgress`] events carried.
+   * [`AgentEvent::MediaProgress`](super::AgentEvent::MediaProgress) events carried.
```

A Rust module path in a consumer-facing contract is a leak, not a link, and
"regenerate the artifact" would have committed it. The `cfg(doc)` import
resolves every link while changing **no rendered text and no wire artifact** —
`git diff --stat` touches one file.

Two smaller notes:

- `#[cfg(doc)]` rather than a plain `use` because nothing in this module names
  the type in code, so a plain import is dead under `clippy -D warnings`. The
  import exists exactly when the links need it and never otherwise.
- With the type in scope, the header's `[`AgentEvent`](super::AgentEvent)`
  trips `rustdoc::redundant_explicit_links`, so it becomes the bare `[`AgentEvent`]`.
  Module docs are not exported to the `.d.ts`, so this changes no artifact either.

No `#[allow]` is added anywhere.

## Verification

`make gate` — **exit 0**, whole workspace, run against this tree.

Captured as a real exit code, not a piped one: `make gate > log 2>&1; echo $?`.
The first run of this investigation reported success through a `| tail` pipe
while `make` had actually exited 2, which is worth flagging for anyone
verifying locally.

Individually, before and after the full run:

| Step | Before | After |
|---|---|---|
| `make doc-warnings` | 101 | **0** |
| `make wire-schema` | 0 (masked — never ran) | **0** |
| `make gate` | 2 | **0** |

## Witness

Not applicable in the fail-on-old/pass-on-new sense: the defect is a build-time
rustdoc failure, not a behaviour, and the gate step *is* the test — it fails on
`main` and passes here. No runtime behaviour changes; this is doc scope only.

## Deleted tests

None.

Refs #3436, #3440

## Summary by Sourcery

Fix rustdoc and gate failures in stella-protocol by ensuring AgentEvent is in scope for intra-doc links in the payload module.

Bug Fixes:
- Resolve broken intra-doc links to AgentEvent variants in the payload module that caused rustdoc -D warnings and make gate to fail.

Enhancements:
- Adjust payload module documentation to avoid leaking Rust module paths into the generated TypeScript wire schema while keeping links functional.

### DIFF
--- a/crates/stella-protocol/src/event/payload.rs
+++ b/crates/stella-protocol/src/event/payload.rs
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: AGPL-3.0-only
 // Copyright (c) 2026 Oxagen, Inc. Commercial licensing: licensing@oxagen.sh
 
-//! The leaf payload types an [`AgentEvent`](super::AgentEvent) variant carries
+//! The leaf payload types an [`AgentEvent`] variant carries
 //! — file-change kinds, verdict evidence, scope/hunk proposals, media refs, PR
 //! and CI status, and the task-board item.
 //!
@@ -18,6 +18,18 @@
 use serde::{Deserialize, Serialize};
 
 use crate::ladder::LadderSnapshot;
+
+/// Brought into scope for the `[`AgentEvent::…`]` intra-doc links below, which
+/// resolved inline in `event.rs` and stopped resolving when #3440 moved these
+/// types into a child module (`rustdoc -D warnings` is a gate step).
+///
+/// `#[cfg(doc)]` rather than a plain `use` because nothing here names the type
+/// in code, and rather than qualifying each link as `(super::AgentEvent::…)`
+/// because these doc comments are exported verbatim into
+/// `docs/wire/agentevent.d.ts` — a Rust module path in the public TypeScript
+/// contract would be a leak, not a link.
+#[cfg(doc)]
+use super::AgentEvent;
 
 /// What happened to a file in a [`AgentEvent::FileChange`] event.
 ///


macanderson:
Superseded by #3450, which landed the identical `#[cfg(doc)] use super::AgentEvent;` fix and is now on main (865d70eae).

Verified after the merge with the exact gate command — `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps --document-private-items --keep-going` — which is clean workspace-wide. All five unresolved links are gone.

Closing as redundant, not as wrong: this branch's approach was the correct one.

PR #3455, opened 2026-08-17
fix(stella-protocol): repair main's red doc-warnings gate (5 broken intra-doc links)

macanderson:
## `main` is red

`make doc-warnings` fails on `origin/main` right now, and has since #3440
merged. This is not a change any subsequent PR made — every branch cut from
`main` inherits it, and because a red gate step masks the ones after it, the
failure a PR author sees is not their own.

```
  --> crates/stella-protocol/src/event/payload.rs:212:11
make: *** [doc-warnings] Error 101
```

## Cause

#3436 (PR #3440) split `event.rs`'s payload types out into
`event/payload.rs` to clear the file-size ratchet. The doc comments travelled
with the types, and five `[`AgentEvent::…`]` links that resolved from
`event.rs` — where `AgentEvent` is a sibling item — resolve to nothing from
the child module.

A pure code motion, which is exactly the shape that carries a link break
invisibly: nothing about the moved lines changed.

## Fix

Name the path explicitly, `(super::AgentEvent::…)` — the form line 4 of the
same file already uses for the module-level link, so this makes the file
internally consistent rather than introducing a convention.

Five links, all in `crates/stella-protocol/src/event/payload.rs`; that is the
complete set the gate reports, verified by re-running it to green.

## The second commit: regenerated wire artifacts

Doc comments on wire types become JSON Schema `description` strings, so the
repair changes `docs/wire/` and `wire-schema` fails until those are
regenerated and committed (`make wire-schema-update`).

Kept as a separate commit so the reviewable claim is checkable at a glance:
the diff is **prose only** — four description strings across the two schemas
and their two `.d.ts` twins. No type, field, variant, or shape changed. That
is the guard working, not a contract change wearing a doc fix.

## Verification

`make gate` — `GATE_EXIT=0` on this branch, red at `doc-warnings` on the
parent commit.

No witness test. The gate step *is* the witness, and a test asserting a doc
link resolves would be a worse copy of it.

Refs #3436

### DIFF
--- a/crates/stella-protocol/src/event/payload.rs
+++ b/crates/stella-protocol/src/event/payload.rs
@@ -19,7 +19,7 @@ use serde::{Deserialize, Serialize};
 
 use crate::ladder::LadderSnapshot;
 
-/// What happened to a file in a [`AgentEvent::FileChange`] event.
+/// What happened to a file in a [`AgentEvent::FileChange`](super::AgentEvent::FileChange) event.
 ///
 /// Both live producers measure a tree against a tree, so every kind emitted
 /// today is a mutation — see [`Self::Read`] for the one that is not, and why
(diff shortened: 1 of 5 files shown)


macanderson:
Superseded by #3450, now on main (865d70eae). The doc-warnings gate is verified clean workspace-wide with `--document-private-items`.

Closing this one specifically because its approach would have changed the **published wire contract**. `schemars` exports these doc comments verbatim as the `description` fields of `docs/wire/agentevent.schema.json` and `agentevent.d.ts` — which is why this branch had to regenerate those artifacts. `origin/main`'s `agentevent.d.ts` contains no Rust module paths today; qualifying the links inline would have introduced `super::AgentEvent::FileChange` into a consumer-facing TypeScript contract to satisfy a rustdoc lint.

#3450's `#[cfg(doc)] use super::AgentEvent;` resolves the same links with the documented text left byte-identical and zero wire-artifact churn.

PR #3457, opened 2026-08-17
docs(stella-protocol): qualify the AgentEvent links payload.rs cannot resolve

macanderson:
`main` is red on `cargo doc -D warnings` — five `rustdoc::broken-intra-doc-links` errors, all in `crates/stella-protocol/src/event/payload.rs`:

```
error: unresolved link to `AgentEvent::FileChange`   (x3)
error: unresolved link to `AgentEvent::MediaComplete`
```

**Cause.** Splitting `event.rs` into `event/payload.rs` carried doc comments naming `AgentEvent` into a module where it is not in scope, so those links resolve to nothing.

**Fix.** The file's own module doc already uses the working idiom — `[`AgentEvent`](super::AgentEvent)` — so this applies it to the five item links that were missed. Rendered text is unchanged; only the link target is added.

No `use` was added: the type is named only in documentation here, so an import would be an unused one.

**Verified by exit code:**

```
RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps --document-private-items --keep-going
  → 0   (101 before)
cargo fmt --all --check          → 0
cargo test -p stella-protocol    → 0
```

Failing run: https://github.com/macanderson/stella/actions/runs/31996255294

## Summary by Sourcery

Documentation:
- Qualify AgentEvent-related intra-doc links in payload.rs to correctly reference event variants and resolve rustdoc warnings.

### DIFF
--- a/crates/stella-protocol/src/event/payload.rs
+++ b/crates/stella-protocol/src/event/payload.rs
@@ -19,7 +19,7 @@ use serde::{Deserialize, Serialize};
 
 use crate::ladder::LadderSnapshot;
 
-/// What happened to a file in a [`AgentEvent::FileChange`] event.
+/// What happened to a file in a [`AgentEvent::FileChange`](super::AgentEvent::FileChange) event.
 ///
 /// Both live producers measure a tree against a tree, so every kind emitted
 /// today is a mutation — see [`Self::Read`] for the one that is not, and why
@@ -36,7 +36,7 @@ pub enum FileChangeKind {
     /// again. Neither surviving producer can: both diff a tree against a tree,
     /// and a read leaves no trace in a tree to diff. Re-acquiring one would
     /// mean going back to declaring file operations from tool inputs, which is
-    /// the defect [`AgentEvent::FileChange`] documents rather than a capability
+    /// the defect [`AgentEvent::FileChange`](super::AgentEvent::FileChange) documents rather than a capability
     /// worth restoring.
     ///
     /// It stays in the kind space because journals recorded before the purge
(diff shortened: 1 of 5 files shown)


macanderson:
Rebased onto acd43094a (current main) and re-verified against it: cargo doc -D warnings = 0, check-wire-schema.sh = 0, cargo fmt --all --check = 0.

main is currently red on cargo doc -D warnings and this PR is the fix. The last two commits to land (acd43094a, d5d3b22e8) both fail on it, for the five unresolved-link errors in crates/stella-protocol/src/event/payload.rs — inherited, not caused by either.

macanderson:
Superseded by #3450, which fixed this better.

#3450 uses a cfg(doc) import:

    #[cfg(doc)]
    use super::AgentEvent;

Rustdoc sets cfg(doc) so the bare [`AgentEvent::X`] links resolve, while every non-doc build never sees the import and unused_imports never fires. The doc TEXT is unchanged, so docs/wire/ needs no regeneration at all.

That is strictly better than what this PR did. Qualifying the links inline as (super::AgentEvent::X) changed the doc text, and schemars exports these comments verbatim as the description fields of agentevent.schema.json and agentevent.d.ts — so my version would have leaked a Rust module path into the published wire contract, and needed a schema regeneration to do it. #3450's commit message calls out exactly that tradeoff.

Closing unmerged. main is already carrying the fix.

macanderson:
Superseded by #3450, which fixed this better.

#3450 uses a cfg(doc) import: rustdoc sets cfg(doc) so the bare [AgentEvent::X] links resolve, while every non-doc build never sees the import and unused_imports never fires. The doc TEXT is unchanged, so docs/wire/ needs no regeneration at all.

That is strictly better than what this PR did. Qualifying the links inline as (super::AgentEvent::X) changed the doc text, and schemars exports these comments verbatim as the description fields of agentevent.schema.json and agentevent.d.ts — so my version would have leaked a Rust module path into the published wire contract, and needed a schema regeneration to do it.

Closing unmerged; main already carries the fix.

PR #3460, opened 2026-08-17
fix(stella-protocol): re-resolve payload.rs's AgentEvent doc links after the #3440 split

macanderson:
Refs #3440, #3449

## What is broken

`main` is red on the `doc-warnings` gate step, and has been since #3440 merged.

That PR split the payload types out of `crates/stella-protocol/src/event.rs`
into `event/payload.rs`. In `event.rs`, `AgentEvent` was a sibling item, so the
bare link `` [`AgentEvent::FileChange`] `` resolved. One module down it does
not, and `rustdoc -D warnings` turns each unresolved link into an error:

```
error: unresolved link to `AgentEvent::FileChange`
  --> crates/stella-protocol/src/event/payload.rs:22:36
...
```

Five links, three targets: `FileChange` ×3, `MediaComplete`, `MediaProgress`.

This is also the whole of why **#3449's `ci` run was red** — its `fmt + clippy
+ test` job failed at this step; clippy passed and every test passed (1752 in
`stella-cli` alone). #3449 was merged over an inherited failure it did not
cause, so the red moved onto `main` with it.

## The change

Each link gets a **reference-style definition** (`` [`AgentEvent::FileChange`]:
super::AgentEvent::FileChange ``) rather than an inline `(super::…)` target.

That distinction is the substance of this PR, not style. **These doc comments
are exported:** schemars lifts each one into the `description` field of the
generated `docs/wire/*.schema.json` and `*.d.ts`, so how the link is spelled is
a wire-contract question. Spelling it inline rewrites five consumer-facing
sentences to carry a `(super::AgentEvent::MediaProgress)` Rust path that means
nothing to a TypeScript consumer reading the contract. A definition appends one
line that rustdoc consumes when rendering and leaves every sentence untouched.

No `use` was added: importing a type solely so a doc link resolves is an unused
import in every non-doc build.

`docs/wire/` is regenerated in the same commit (`wire-schema` is a gate step
and the artifacts are committed). The diff there is **pure additions** — the
appended definition lines only. No field, type, variant, or required-ness
moved, so the additive-only contract holds.

## Verification (by exit code)

| command | this branch | parent commit (`acd43094`) |
| --- | --- | --- |
| `RUSTDOCFLAGS="-D warnings" cargo doc -p stella-protocol --no-deps` | 0 | 101, `could not document stella-protocol` |
| `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps` | 0 | — |
| `make wire-schema` | 0 | 0 |
| `cargo fmt --all --check` | 0 | 0 |
| `cargo check -p stella-protocol --all-targets` | 0 | 0 |

The workspace doc run is the one that matters beyond the immediate fix: it
confirms `stella-protocol` was the *only* crate carrying this rot, so the gate
step goes green rather than advancing to the next broken crate.

## Witness

None, and for the honest reason rather than an exemption: the failing gate step
*is* the witness. `cargo doc` exits 101 on the parent and 0 here — a fail→pass
flip, run twice, recorded above. There is no runtime behaviour to assert on; no
`#[test]` can observe whether an intra-doc link resolves.

The wire half is witnessed the same way, by its own guard: `make wire-schema`
failed on the first push of this branch (the artifacts were stale) and passes
now.

## Deleted tests

None.

### DIFF
--- a/crates/stella-protocol/src/event/payload.rs
+++ b/crates/stella-protocol/src/event/payload.rs
@@ -19,12 +19,13 @@ use serde::{Deserialize, Serialize};
 
 use crate::ladder::LadderSnapshot;
 
-/// What happened to a file in a
-/// [`AgentEvent::FileChange`](super::AgentEvent::FileChange) event.
+/// What happened to a file in a [`AgentEvent::FileChange`] event.
 ///
 /// Both live producers measure a tree against a tree, so every kind emitted
 /// today is a mutation — see [`Self::Read`] for the one that is not, and why
 /// it stays in the space anyway.
+///
+/// [`AgentEvent::FileChange`]: super::AgentEvent::FileChange
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
 #[cfg_attr(feature = "schema", derive(schemars::JsonSchema))]
 #[serde(rename_all = "snake_case")]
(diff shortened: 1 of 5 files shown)


macanderson:
Superseded by #3450, now on main (865d70eae). The doc-warnings gate is verified clean workspace-wide with `--document-private-items`.

Closing this one specifically because its approach would have changed the **published wire contract**. `schemars` exports these doc comments verbatim as the `description` fields of `docs/wire/agentevent.schema.json` and `agentevent.d.ts` — which is why this branch had to regenerate those artifacts. `origin/main`'s `agentevent.d.ts` contains no Rust module paths today; qualifying the links inline would have introduced `super::AgentEvent::FileChange` into a consumer-facing TypeScript contract to satisfy a rustdoc lint.

#3450's `#[cfg(doc)] use super::AgentEvent;` resolves the same links with the documented text left byte-identical and zero wire-artifact churn.

Issue #3408, opened 2026-08-17
Wire the [wrapper] manifest into stella-pipeline: the stage list is parsed but nothing reads it


macanderson:
## Problem

The `[wrapper]` manifest block landed in `crates/stella-plugin/src/wrapper.rs`
(#3381, manifest half). It parses, validates, load-checks its stage graph, and
ships two fixture variants — and **nothing reads it**. `stella-pipeline` still
runs the stage order as hardcoded branches in `Pipeline::run`.

Unwired code with no tracking issue is the failure mode AGENTS.md's "nothing
left behind" rule exists to prevent, so this is that issue.

## Why it could not be done in the same PR

Binding a stage name to the loop needs the four wrapper interception points
(`before_turn` / `after_turn` / `judge` / `again?`) of **#3380**, which do not
exist anywhere in the tree — verified: no `Wrapper`/`TurnWrapper` trait and no
`before_turn`/`after_turn`/`judge` item exists in `stella-core`,
`stella-runtime`, `stella-pipeline`, or `stella-plugin`; every hit is prose in
`docs/spec/turn-loop-wrappers.md`. #3380 is itself blocked by #3379.

## What to build

1. **A host-side evaluator.** `Condition` is a pure two-shape grammar; the
   host supplies the published signal values and asks whether each stage runs.
   `CompareOp::apply` is already the whole count evaluator. The evaluator
   itself must stay a pure function (invariant 2) — the *values* come from the
   host.
2. **`stella-pipeline` reads the manifest** instead of branching. Today's
   order ships as `wrapper-staged-v1.toml`'s equivalent under the variant id
   `classic` — which is already the id
   `crates/stella-cli/src/agent/persistence.rs::PIPELINE_VARIANT_CLASSIC`
   writes, so the store side needs no change to stay honest.
3. **Publish the gating signals that are still missing.** `Signal` publishes
   five facts today (`test-command`, `conversational`, `questions`, `plans`,
   `verifies`), each transcribing a live branch. `Pipeline::run` also gates on
   `responsibility_enabled(ModelCallRole::…)`, `assessment.wants_witness()`,
   and `can_author_independent_witness()` (`pipeline.rs:1137-1142`). Each is
   either a new published `Signal` or a documented reason it is host-internal
   and not manifest-visible. **Do not** widen the grammar to reach them — a
   new signal is a reviewable addition, a richer expression syntax is not
   (`doc:turn-loop-wrappers` §9.4).

## Constraints already discovered

[...]

Issue #3388, opened 2026-08-16
store: executions.kind encodes the wrapper, not just the door — disentangle it before the variant column lands


macanderson:
## Problem

`executions.kind` (`crates/stella-store/src/ddl.rs:79`) is described everywhere
as the **door** the turn came in by — the command the user typed. Two of its
values are not doors:

```
crates/stella-cli/src/agent.rs:281          begin_execution(&store, "pipeline", …)
crates/stella-cli/src/command_deck.rs:1514  if pipeline_on { "deck-pipeline" } else { "deck" }
```

`pipeline` and `deck-pipeline` name **which wrapper ran**, in the column that is
supposed to name the door. A deck turn writes a different door depending on
whether a wrapper was on.

## Why it blocks #3381

#3381 adds `pipeline_variant` beside `kind` and states the rule correctly — "the
door tag and the variant id are two columns, not one" — but it was written
believing `kind` was already door-only. It is not. Adding the variant column on
top of the current `kind` produces:

- **two columns answering one question**, disagreeing whenever the mapping is
  imperfect — the shape `doc:turn-lane-assembly` calls out as how the last limit
  died, and which #3381's own constraints section warns against;
- a `GROUP BY kind` over the deck that **silently splits one door in two**, so
  the measurement surface #3381 exists to create is confounded on arrival.

This is the measurement surface for the whole wrapper programme, and CLAUDE.md's
measure-honestly rule applies to it before the first number is taken, not after.

## What to build

1. `kind` becomes the door and only the door: `pipeline` → `run`,
   `deck-pipeline` → `deck`. The six doors are `run`, `deck`, `deck-sub`,
   `goal`, `fleet`, and whatever `chat` resolves to (see open question).
2. Which wrapper ran moves to `pipeline_variant` (#3381), NULL for an unwrapped
   turn — the sole home for that fact.
3. **The backfill is stated, not assumed.** Historical `pipeline` rows become
   `kind='run', pipeline_variant='classic'`; `deck-pipeline` becomes
   `kind='deck', pipeline_variant='classic'`. Any query spanning the migration
   boundary must say so rather than hide it (#3381's own confounding rule).

## Files

- `crates/stella-cli/src/agent.rs:281`, `crates/stella-cli/src/command_deck.rs:1514`
  — the two write sites.
- `crates/stella-store/src/ddl.rs`, `crates/stella-store/src/migrations.rs` — the
  migration and the `SCHEMA_VERSION` bump.
[...]

Issue #3381, opened 2026-08-16
Wrapper stages become a manifest, variants become a column, and --pipeline replaces --no-pipeline


macanderson:
## Problem

Two problems that share one fix.

**Iterating on a pipeline design means recompiling a crate.** The conditional
skipping the deck describes — "a cheap task should not pay for ceremony it
cannot use" — is hardcoded branches inside
`crates/stella-pipeline/src/pipeline.rs`. That file is on the god-file list in
`scripts/file-size-baseline.txt` and is **closed to growth**, so the current
design cannot absorb another variant even if we wanted one.

**We cannot say which pipeline design is better.** Every turn already writes an
execution row tagged with the door it came in by — `run`, `chat`, `deck`,
`deck-sub`, `goal`, `pipeline`. Nothing records *which shape of wrapper* ran. So
"is the cheaper stage order better?" is answerable only by somebody who was
there, which is the exact lineage failure #2889 was filed against.

### 1. Stages as a manifest

A wrapper plugin declares an ordered stage list with conditions, in TOML:

```toml
[wrapper]
id = "staged-v1"

[[wrapper.stages]]
name = "triage"

[[wrapper.stages]]
name = "plan"
if   = "questions > 0"

[[wrapper.stages]]
name = "execute"

[[wrapper.stages]]
name = "witness"
if   = "no-test-command"

[[wrapper.stages]]
name = "verify"
```

Each stage is an independent unit with a typed input and a typed output, so a
stage is testable on its own. The conditions become a line you can read and
change instead of a branch buried in a 2700-line file.

Two manifest rules carried from #3245 slice A
(`crates/stella-plugin/src/manifest.rs`) — do not re-derive them:

- An unknown key is a **load error**, never ignored.
- A condition naming a signal the host does not publish is **also a load
  error**. A manifest that quietly does nothing is worse than one that refuses
  to load.

### 2. One column on the execution row

Add the wrapper variant id to the executions table
(`crates/stella-store/src/ddl.rs`), beside the existing door tag. That is the
whole A/B setup: cost, outcome, and verified-versus-unverified rate per variant
is one `GROUP BY variant` against `.stella/private/store.db`.

Two honesty rules, because this is a measurement surface and CLAUDE.md's bench
rules apply to it:

- A variant id is written **only when that manifest was the thing that ran**. A
  default or fallback path writes the default's id, never a blank — a missing
  measurement must not render as a negative one (#2952's recurring sub-shape).
[...]

Issue #3380, opened 2026-08-16
One wrapper contract: before_turn / after_turn / judge / again? — and the pipeline becomes its first plugin


macanderson:
## Problem

Route B (the staged pipeline) and Route D (goal mode) are the same idea written
twice.

- **Route B**: plan, execute, verify, verdict, with a revise loop.
- **Route D**: rounds, work, an independent verifier, met or not-met.
- **`stella monitor`** is Route D with the goal pinned to "the latest CI run is
  green". The deck says outright that it is not a seventh route.

Strip the vocabulary off all three and the shape is identical: **propose work,
run a turn, judge from evidence, decide continue-or-stop.**

One idea, three surfaces. Every fix has to be made in each of them, and when
somebody forgets, they drift. Two live examples in the backlog:

- #2819 — `VerifierScopedTools` in `crates/stella-cli/src/agent/goal.rs` is goal
  mode's own hand-rolled read-only verifier view, and it silently drops two
  `ToolExecutor` methods the pipeline's equivalent does not. That is drift
  between two copies of "the judge sees a read-only tool stack".
- The deck's slide 22 names the structural version: `Engine::with_sleeper`
  cannot carry `gate`, `steering` or `hooks` because those are private builder
  fields, so a hand-rolled child engine drops all three. **Goal mode shipped
  with exactly that bug.**

## What to build

One wrapper contract with four interception points, and nothing else:

| Point | When it runs | May do | May not do |
|---|---|---|---|
| `before_turn` | before the loop is asked for a turn | add context, pick a plan, narrow scope, choose a model role | run the loop itself |
| `after_turn` | once the turn's `Complete` lands | gather evidence — run a test, read a diff, author a witness | change the turn that just ran |
| `judge` | after `after_turn` | turn evidence into a verdict | call a model |
| `again?` | after `judge` | ask for another turn with a correction, or stop with an outcome | fake an ending the engine did not emit |

Then:

- **The staged pipeline becomes the first plugin using it.** `triage`, `recall`,
  `research`, `plan`, `scope` are `before_turn`; `witness` is `after_turn`;
  `verify` is `judge`; `revise` is `again?`.
- **Goal mode becomes the second** — or the same wrapper with a different
  `judge`. Its rounds are `again?`; its independent verifier is `judge`.
- **Core keeps one loop, one wrapper socket, and zero built-in wrappers.**

## Why `judge` calls no model

This is already the shipped rule, not a new one. #2584 removed the model verdict
[...]

PR #3464, opened 2026-08-17
docs(crates): every crate README reflects the plugin-era product

macanderson:
All 24 crate READMEs updated for the current product state: Stella as a simple, extensible, durable, composable **plugin-architecture agent turn loop**, embeddable in any application through the Rust ports or the HTTP surface — and verification leaving the engine for a plugin.

## What changed, and why

Each README gained a short **Direction** section stating what the plugin era means *for that crate*, written against the tree and `doc:turn-loop-wrappers` / `doc:engine-embedding` rather than recalled:

- **stella-core** — the direction is subtractive: the engine owns its turn's ending only (#3379, landed), goal mode's round loop leaves for the wrapper contract (#3380), and the wrapper socket lands *above* core because `before_turn`/`after_turn` do I/O (invariant #2). Also fixes the intro's stale "goal loop" framing and flags `src/goal.rs` as closed to growth.
- **stella-pipeline** — stated plainly: this crate becomes a plugin (#3246), and once it does **Stella does not verify its own work unless the plugin is installed**. Also corrects the stale stage list (`evaluate → enhance → route → …`) to the canonical `stage_rank` ordering in `src/replay.rs`, and records the three moves with their real states (#3379 closed, #3380/#3381/#3408 open).
- **stella-plugin** — the manifest is the front of the product; #3245's sequencing now lives in #3246; `[wrapper]` parses and nothing reads it yet (#3408), declared as a tracked gap.
- **stella-runtime** — the wrapper socket lands here, with the §9.1 argument for why not core, and the honest caveat that "one blessed constructor" is only enforceable once `TurnCapabilities` exists (#3274 slice 2).
- **stella-engine / stella-serve** — named as the two embedding doors (in-process ports, over-the-wire HTTP/SSE), each stating that the loop it exposes does not adjudicate its own work. `stella-serve/src/goal.rs` flagged as a wrapper living in a surface crate.
- **stella-cli** — one door, not the product: the planned flag inversion (#3381), the 169-reference pipeline dependency gated on #2716, and `kind`-is-the-door / `pipeline_variant`-is-the-wrapper (#3388).
- **stella-store / stella-observatory** — the door/wrapper column split, with the v25 → v26 backfill (`kind='run', pipeline_variant='classic'`) named as a confounder any query spanning it must declare.
- **stella-protocol / stella-tui / stella-parity / stella-fleet / stella-tools / stella-model / stella-mcp / stella-context / stella-graph / stella-media / stella-diag / stella-embed / stella-home / stella-tty / stella-diff** — each tied to the same architecture from its own side (one ending per turn plus one per run; a wrapper is not a surface; twelve built-ins is the closed surface; no plugin ever sees a credential; recall is `before_turn`'s question; leaf crates are the composability substrate).

## Stale facts corrected along the way

- `stella-protocol`: "eleven of the other fourteen crates" → fifteen of twenty-three, plus `stella-plugin` in the dependents diagram.
- `stella-diag`: "seventeen other crates" → twenty-three.
- `stella-home`: "the three `STELLA_*` overrides" → two (`OVERRIDE_ENV_VARS` is `[&str; 2]`; `STELLA_CONFIG_DIR` retired in #2442).
- `stella-tui`: dropped a source-file/line count that goes stale on every commit.

## Verification

Docs-only, no Rust touched — no witness test applies. Ran the guards this change can affect:

```
make doc-links brand-case god-files left-behind
check-doc-links: OK -- 337 citation(s) resolve across 71 identified document(s)
check-brand-case: OK (exit 0)
check-god-files: OK — 17 god file(s) across 6 crate(s), named identically in AGENTS.md and every crate README
check-left-behind: OK — 0 grandfathered
```

Every issue number and every landed/open claim was checked against `gh issue` state and the tree (`stage_rank`, `OVERRIDE_ENV_VARS`, `begin_execution` call sites, `migrations/pipeline_variant.rs`, `Cargo.toml` dependent counts) rather than recalled.

## Deliberately not done

- The **root `README.md`** still leads with the witness-test contract as an unqualified product claim. That is the user-facing half of this same change and wants its own PR with a maintainer decision on how to state "verification is becoming opt-in" to users — filed as an issue rather than folded in here.


macanderson:
Residue filed as #3465 — the root README / website half of this same change, left to a maintainer because it is a user-facing product claim gated on the `doc:turn-loop-wrappers` §9.6 bench.

Issue #3588, opened 2026-08-17
Flaky: `daemon::tests::a_run_that_ignores_term_is_killed_after_the_grace_period` fails under a loaded full-suite run


macanderson:
## Problem

`cargo test -p stella-cli --bin stella` fails intermittently on:

```
---- daemon::tests::a_run_that_ignores_term_is_killed_after_the_grace_period stdout ----
thread '...' panicked at crates/stella-cli/src/daemon/tests.rs:504:5:
the child was killed after 81.18753ms, before its 8s to shut down cleanly
```

## Evidence that it is a flake and not a regression

Three consecutive full-suite runs of the **same binary**, on the same tree:

| run | result |
| --- | --- |
| 1 | FAILED (1780 passed, 1 failed) |
| 2 | ok (1781 passed) |
| 3 | FAILED (1780 passed, 1 failed) |

Run in isolation (`cargo test -p stella-cli --bin stella
daemon::tests::a_run_that_ignores_term`) it passes every time. The failure is
therefore load- and scheduling-sensitive, not a property of the code under
test.

The measured 81 ms against an 8 s grace period says the child was reaped
almost immediately, which reads as the spawned `/bin/sh` not yet having
installed its `trap` when the `SIGTERM` arrived — a race between the
supervisor's stop and the child's own startup, widened by ~1700 other tests
competing for CPU.

## Files

- `crates/stella-cli/src/daemon/tests.rs` around line 471–504 — the test and
  its child script.
- `crates/stella-cli/src/daemon.rs` — the stop/grace/kill ladder under test.

## Verify

Run the full binary suite several times; expect an intermittent failure at the
assertion above. Then confirm the fix by running the suite repeatedly (10×)
with no failure, and by running the test alone (which already passes) to be
sure the fix did not simply delete the claim.

## What "done" looks like

The test synchronises on the child having installed its handler — a readiness
signal the child writes (a file, a line on a pipe) that the test waits for
before sending `SIGTERM` — rather than on elapsed wall time. Widening the
grace period would only move the flake; the assertion is about the ladder
honouring the grace, and that claim is worth keeping.

macanderson:
[...]

Issue #3559, opened 2026-08-17
Flaky again: daemon::tests::a_run_that_ignores_term_is_killed_after_the_grace_period kills the child ~100x early under a full workspace test run


macanderson:
## Problem

```
thread 'daemon::tests::a_run_that_ignores_term_is_killed_after_the_grace_period'
panicked at crates/stella-cli/src/daemon/tests.rs:504:5:
the child was killed after 81.791959ms, before its 8s to shut down cleanly
```

Reproduces on a clean checkout of `main` (2509db5) under `cargo test
--workspace` on this container — verified by stashing an unrelated working
tree and re-running, so it is not caused by any in-flight change. It is the
only failure in an otherwise green 1749-test run.

This shape has been filed and closed **four** times: #2393, #2804, #2828,
#2902. It is back, which suggests the previous fixes addressed symptoms
(timing margins) rather than the mechanism. Filing fresh rather than
commenting on a closed issue so it is visible in the open queue.

## What is actually being asserted

The test starts a child that ignores `SIGTERM` and asserts the daemon waits
its full grace period (8s) before escalating to `SIGKILL`. It fails by killing
after ~82ms — about 1/100th of the budget — which is not a margin problem. A
grace period that collapses to ~zero under load is either:

- a timer measured against something other than wall clock, or reset by an
  event the loaded run delivers and the isolated run does not; or
- the child not yet having installed its `SIGTERM` handler when the daemon
  signals, so it dies to the first signal and the daemon reports the kill it
  did not perform.

The second would make the test's own premise wrong under load, which would
also explain why it passes in isolation and why four rounds of timing fixes
did not settle it. Worth ruling out first, cheaply: log whether the child
observed the `SIGTERM` at all.

## What "done" looks like

- The mechanism is identified and named — not the margin widened again. If the
  race is the handler-installation one, the test should synchronise on the
  child signalling readiness rather than on a sleep.
- `cargo test --workspace` passes repeatedly on a loaded machine (loop it 20x).
- If the grace period genuinely cannot be asserted deterministically in-process,
  say so and replace the assertion with one that can be — a flaky gate step is
  worse than a missing one, because it trains everyone to re-run.

## Files

- `crates/stella-cli/src/daemon/tests.rs` — the test, assertion at line 504.
- `crates/stella-cli/src/daemon.rs` — the grace period and escalation.

## Verify

```
cargo test --workspace          # fails under load
[...]

Issue #2828, opened 2026-08-11
daemon test `a_run_that_ignores_term_is_killed_after_the_grace_period` fails under load — kills the child ~100x early


macanderson:
## Problem

`crates/stella-cli/src/daemon/tests.rs` — `daemon::tests::a_run_that_ignores_term_is_killed_after_the_grace_period` fails when the machine is loaded, with:

```
thread 'daemon::tests::a_run_that_ignores_term_is_killed_after_the_grace_period'
panicked at crates/stella-cli/src/daemon/tests.rs:483:5:
the child was killed after 81.231551ms, before its 8s to shut down cleanly
```

The test asserts the child got its full grace period before SIGKILL. Under load the elapsed-time measurement reads ~81ms against an 8s expectation — two orders of magnitude out, which is the signature of a starved timer rather than a marginal one.

## It is a flake, not a regression — the control was run

Observed during the P0 sweep in #2814, on a box running five concurrent `cargo` builds. The cheap control settles it:

```
$ cargo test -p stella-cli --bin stella daemon::tests::a_run_that_ignores_term
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 1705 filtered out; finished in 8.03s
```

Passes alone, in 8.03s — i.e. it really does wait the full grace period when it gets scheduled. It failed only inside a full `cargo test -p stella-cli` run competing with other builds. Nothing in #2814 touches `daemon.rs`.

Note the 8.03s runtime: this single test spends 8 seconds of wall clock, which is also why it is worth fixing rather than just tolerating.

## Why it matters

A wall-clock assertion in the test suite is a load-dependent failure, and CI runners are loaded by definition. This will fail intermittently on `ci.yml` and cost a re-run each time — and worse, it trains readers to re-run a red gate instead of reading it, which is the habit the gate exists to prevent.

## Files

- `crates/stella-cli/src/daemon/tests.rs` (the assertion is at ~:483)
- `crates/stella-cli/src/daemon.rs` — the grace-period implementation under test

## Suggested approaches

1. **Inject the clock.** The engine's own discipline (invariant #2, "no I/O in the engine") is that timing decisions should be pure functions over owned data. If the grace-period decision can be extracted from the sleep, the test asserts the decision and never sleeps — this is the durable fix and it also reclaims the 8 seconds.
[...]

PR #3597, opened 2026-08-17
fix(plugins): three defects the adversarial audit of #3590 confirmed

### LINKED ISSUE #3581: A package's declaration is reconciled with its directories at install only, so a package edited after consent still contributes what nobody agreed to
## Problem

#3565 made a plugin manifest declare what its package ships (`[[tools]]`,
`[[skills]]`, `[[records]]`) and added
`stella_plugin::PluginManifest::reconcile`, which refuses a package whose
declaration and directories disagree in either direction. `stella plugin
install` calls it before printing the consent document, which is what makes
that document provably complete **at the moment of consent**.

It is not called anywhere else. The roster is recomputed from disk on every
load (`crates/stella-cli/src/plugin_cmd/roster.rs`), and the three contributed
surfaces are derived from the package directories on every session
(`package::contributed_tool_dirs` / `contributed_skill_dirs` /
`contributed_record_dirs`). So:

- Drop a new `tools/deploy.toml` into `.stella/plugins/vera/` after install and
  the model is offered `deploy` next session, running as `Principal::Plugin("vera")`,
  with no `[[tools]]` entry naming it and no consent document that ever did.
- The same holds for `skills/` and `rules/`.

The write needs local filesystem access to the plugins directory, which is a
lower bar than it sounds: the *project* tier lives at `.stella/plugins/` inside
the repository, so this is reachable by anything that can land a commit in a
workspace the operator has trusted with `STELLA_TRUST_PROJECT=1` — the same
threat model `roster::read_project_tier`'s gate exists for.

This is a deliberate scope cut in the PR that closed #3565, not an oversight —
it is filed rather than fixed because making reconciliation a load-time gate is
a behavioural change to every session's plugin load and wants its own review.

## Files

- `crates/stella-cli/src/plugin_cmd.rs` — `install`, the one caller of
  `reconcile` today.
- `crates/stella-cli/src/plugin_cmd/roster.rs` — `read_tier` / `load`, where a
  load-time check would live.
- `crates/stella-cli/src/plugin_cmd/package.rs` — `Inventory::of_package` and
  `Inventory::listing`, which already produce exactly the `PackageListing`
  `reconcile` needs, from a directory.
- `crates/stella-plugin/src/package.rs` — `reconcile`, `PackageMismatch`.

## Repro

1. Build a package declaring one tool, install it with `--yes` into a trusted
   workspace.
2. `echo 'name = "deploy"\ndescription = "d"\ncommand = ["/bin/true"]' > .stella/plugins/vera/tools/deploy.toml`
3. `stella tools` (or assemble the session stack as
   `plugin_cmd::tests::contributed_tools` does) — `deploy` is present and
   attributed to `vera`.

- **Fail closed, and speak.** The roster's precedent is
  `an_untrusted_projects_plugins_do_not_load_are_not_listed_and_dispatch_no_hook`:
  a plugin that vanishes silently is indistinguishable from one that is broken,
  so a mismatch must produce a `notice` naming both sides, not a silent drop.
  `PluginRoster::load` already returns `(roster, notices)`.
- **Cost.** `Inventory::of_package` parses every shipped tool manifest and every
  rules file. Doing that on every roster load is a per-session cost that
  `session_roster` currently avoids; measure it, or cache on
  (path, mtime, len). Note `session_roster` is called three times per session
  (once per surface).
- **Do not weaken provenance.** Whatever is checked, `CustomTool::contributed_by`
  must keep coming from the directory read and never from a declaration —
  `custom::tests::a_manifest_cannot_claim_to_have_been_shipped_by_a_plugin`.
- Every currently-installed package that ships an undeclared contribution would
  stop contributing. That is the intended rule (#3565: "a directory nobody
  declared cannot be consented to"), but it is a migration and should be called
  out in the PR and in `docs/spec/pipeline-as-plugins.md` §13.3.

## What "done" looks like

A package whose directories no longer match its declaration contributes
**nothing** from the mismatching surface (or, if the PR argues it, nothing at
all), with a notice naming the undeclared entries — and the package is still
listed by `stella plugin list` so the state is diagnosable.

Witness: install a package, add an undeclared `tools/*.toml` afterwards, and
assert the tool is absent from the assembled stack and a notice names it.
Fails before the change (the tool is present). Model it on
`plugin_cmd::tests::a_plugins_tool_installs_runs_as_the_plugin_and_retracts_with_it`,
which already builds the gated session stack.

Refs #3565, #3380.

macanderson:
## What & why

Follow-up to #3590, which merged before these landed. An adversarial audit of
that branch — five finders over disjoint lenses, then two independent reviewers
per finding *instructed to refute it and to default to refuted when uncertain* —
confirmed three defects. All three are fixed here, none were filed.

Two are in the transport rather than the protocol, which is where the branch's
own tests were weakest: every wrapper test on #3590 used a well-behaved plugin,
and a socket has to survive a badly-behaved one.

**1. `settle()` never drained stdout — a plugin that talks after answering
wedges the host** (critical, `stella-runtime`). `converse()` returns the moment
it finds the point response; `settle()` then drained only *stderr* to EOF before
`child.wait()`. Trailing stdout sat in the OS pipe, and once it filled (~64 KiB
on Linux) the child blocked in `write(2)` and never exited — so the exchange
hung until the wrapper timeout fired (60s default, up to 600s) and the process
group was SIGKILLed, discarding a response that had already arrived and was
valid. A real regression: the implementation this replaced read both streams
concurrently via `exec::capture`; the `converse`/`settle` split added the
stderr-drain **with a comment about this exact deadlock hazard** and omitted the
stdout half. `settle()` now drains both concurrently under the shared
`MAX_CAPTURE_BYTES` ceiling.

**2. A dead answer-writer was misreported as a manifest problem** (major,
`stella-runtime`). When the task relaying host-call answers to the plugin's
stdin died, the next `PluginMessage::Call` returned `WrapperError::UnannouncedCall`
— whose text asserts "its manifest declares no `[loop] calls`, or this host
attached no gate." Both were false: the state is only reachable *after* a call
was already answered through that same gate. Meanwhile `exchange()` called
`writer.abort()` without joining, so the real `std::io::Error` was discarded
entirely. An operator was sent to edit `[loop] calls`, which does nothing, while
a broken pipe went unrecorded. Now `converse()` distinguishes "a channel that
died mid-conversation" from "no channel was ever offered", joins the writer to
recover the real error, and reports `WrapperError::AnswerChannelFailed { program,
call, source }`.

**3. Consent was kept for exactly one instant** (critical, `stella-cli`).
`PluginManifest::reconcile` ran only from `stella plugin install`; every session
load went through `dirs_of`, which read the installed directories straight off
disk. A plugin is an ordinary subprocess with no filesystem sandbox, so its own
process can write `<plugin_dir>/tools/backdoor.toml` after install completes —
and on the next session start that file was discovered, registered, and
attributed to `Principal::Plugin`, a callable tool the consent document never
named. `dirs_of` now filters through `reconciles_with_disk`, re-running
`reconcile` against a fresh read on **every** load. A drifted package
contributes nothing from any of the three surfaces — deliberately not the
still-matching subset, since a partial load lets an attacker keep whatever
already got past a human while adding to it.

Refs #3581 — this delivers that issue's load-time reconciliation, **not** its
notice. See "Nothing left behind" below.

## The witness

Three, each verified **both ways** by stashing only the source file and keeping
the test:

- `a_plugin_that_talks_after_answering_does_not_wedge_the_exchange`
  (`tests/wrapper_transport_limits.rs`) — a plugin answers, then writes 192 KiB
  (3× a Linux pipe buffer) to stdout. Pre-fix: `test result: FAILED`, panicking
  on `WrapperError::Timeout` after the full 8s budget — the deadlock reproduced.
  Post-fix: passes in well under half the budget.
- `a_dead_writer_mid_conversation_is_not_reported_as_a_missing_gate`
  (`tests/wrapper_host_call.rs`) — an `sh` plugin closes its own stdin, then
  pipelines three `recall` calls. Asserts `AnswerChannelFailed` with
  `source.kind() == BrokenPipe`, i.e. the real transport error reaches the caller.
- `a_plugin_that_drifts_from_its_declaration_after_install_contributes_nothing`
  (`plugin_cmd/tests.rs`) — installs a package declaring one tool, confirms it
  loads, writes an undeclared `tools/backdoor.toml` into the *installed*
  directory without re-running install, then asserts the backdoor never loads
  **and** the declared tool stops loading too. Falsifier re-run by hand:
  stashing only `package.rs` fails it at `tests.rs:957`; restoring passes.

## Deleted tests

None. One fixture was **corrected** in `1db8f216`:
`a_promotion_grant_does_not_arm_a_record_a_plugin_shipped` hand-built a package
whose manifest declared nothing while `rules/set.toml` shipped a record. Change
3 withholds such a package whole, so the record stopped reaching the registry
and the test failed at its `.expect` on *finding* the entry — i.e. the new
behaviour is right and the fixture was stale. Declaring the record keeps the
witness testing what it was written to test; leaving it undeclared would have
made it pass for the wrong reason.

## The gate

- [x] `cargo fmt --all --check`
- [x] `cargo clippy -p stella-runtime --all-targets -- -D warnings`, and
      `-p stella-cli` likewise — **per-crate, not `--workspace`**
- [x] `cargo test -p stella-runtime` — 82 passed
- [x] `cargo test -p stella-cli` — 1805 passed, run three times (see below)
- [ ] `cargo test --workspace` — not run locally; CI is the authority
- [x] CLA signed
- [x] Issue reference appears both above and as a commit trailer

**A correction to an earlier version of this description.** It claimed
`cargo test -p stella-cli (1801)` passed. That run was *filtered* to one test
(`1 passed; 1800 filtered out`) and I reported the total as a passing count —
one test ran, not 1801. That is what let the stale fixture above reach the push
instead of being caught before it. The numbers above are from the full suite,
actually run.

Three full runs on `1db8f216`: `1804 passed; 1 failed`, then `1805 passed`
twice. The single failure is
`daemon::tests::a_run_that_ignores_term_is_killed_after_the_grace_period` —
the known flake in #3559 and #3588 (previously closed four times as #2393,
#2804, #2828, #2902), where the child is reaped ~82 ms into an 8 s grace period
because it has not installed its `SIGTERM` handler under load. It passes alone,
and `daemon.rs` references neither `plugin_cmd` nor the wrapper, so nothing here
reaches it. No third duplicate filed.

A `WriterState` struct was introduced to stay under `too_many_arguments` after
`converse()` gained the writer-join parameter, rather than `#[allow]`-ing it.
`subprocess.rs` is 1085 lines and `error.rs` 329, both under the ceiling; no
baseline touched.

**CI is currently red here for a reason that is not this diff.** `main` is red
at `eb7c2dd48`, this branch's base, on the `Cargo.lock`-versus-manifests check
(`stella-transcript` pinned at `0.9.77` against a `0.9.83` workspace). Filed by
the canary as #3594; #3595 is the one-line repair, open against the same base
and itself `blocked` — a red `main` reds the PR that fixes `main`. I reproduced
it in a pristine worktree of `eb7c2dd48` containing none of these commits, and
`git diff eb7c2dd4..HEAD -- Cargo.lock Cargo.toml 'crates/*/Cargo.toml'` is
empty, so nothing here can move that line. I have deliberately **not** folded
the one-line fix in, though it would turn these checks green immediately —
#3594's own instructions and AGENTS.md § Gotchas both say to land it on its own
from a fresh `main`.

## Nothing left behind

One gap, tracked rather than hidden: **#3581 asked for a notice naming the
mismatch, and this PR does not emit one.** Withholding is silent — a user whose
plugin stops contributing gets no explanation. Sourcery's review caught that I
had written `Closes #3581`; the commit and this description now say `Refs`, and
#3581 carries a comment stating which half landed, why the notice is not a
one-liner (`session_roster` drops `PluginRoster::load`'s notices on purpose, and
is called once per surface, so the naive version prints three times per
session), two candidate designs, and a definition of done.

The audit itself produced exactly the three findings above and all three are
fixed; its `not_fixed` list came back empty. Two of its five lenses —
architecture-invariants and test-quality — returned **zero** findings, reported
as empty rather than padded.

## Ground-rule check

- [x] New cross-boundary types round-trip through serde — `AnswerChannelFailed`
      is a host-side error type and never crosses the wire

**The audit's verification design is the reason to trust these three.** Each
finding faced two reviewers told to prove it *wrong*, with "refute if uncertain"
as the default, and all three survived — refuted count zero. The same pass
produced no speculative findings, which is what makes that zero meaningful.

**#3581 was previously filed rather than fixed** — by me, earlier today, when I
noticed the install-time-only reconciliation and wrote it up instead of closing
it. The audit rated it critical. Worth stating as a calibration point: the thing
I judged as follow-up work was the most serious defect on the branch.

### DIFF
--- a/crates/stella-cli/src/context_records/tests.rs
+++ b/crates/stella-cli/src/context_records/tests.rs
@@ -482,12 +482,21 @@ fn a_promotion_grant_does_not_arm_a_record_a_plugin_shipped() {
     let lineage = "ctx.acme.web.no-force-push";
     append_promotion(root.path(), grant(lineage, "measured over 30 days")).unwrap();
 
-    // A plugin ships that exact lineage, and the workspace ships none.
+    // A plugin ships that exact lineage, and the workspace ships none. The
+    // manifest must *declare* that record: a package whose directories exceed
+    // its declaration reconciles against nothing and contributes nothing
+    // (`plugin_cmd::package::reconciles_with_disk`), which would make this
+    // witness pass for the wrong reason — the record never loading at all,
+    // rather than loading unarmed.
     let plugin = stella_home::resolve_project_plugins_dir(root.path()).join("vera");
     std::fs::create_dir_all(plugin.join("rules")).unwrap();
     std::fs::write(
         plugin.join(crate::plugin_cmd::roster::MANIFEST_FILE),
-        "name = \"vera\"\n",
+        format!(
+            "name = \"vera\"\n\n\
+             [[records]]\nlineage = \"{lineage}\"\n\
+             statement = \"Never force-push to a shared branch.\"\n"
+        ),
     )
     .unwrap();
     std::fs::write(
--- a/crates/stella-cli/src/plugin_cmd/package.rs
+++ b/crates/stella-cli/src/plugin_cmd/package.rs
@@ -66,6 +66,15 @@
 //!   direction. The host renders **nothing** of its own about what a package
 //!   ships; it used to, and the cost was an embedding host showing a document
 //!   that omitted executable code entering the agent's tool surface.
+//!
+//!   Consent is a one-time transaction, but the directories it approved are
+//!   not — a plugin is an ordinary subprocess with no filesystem sandbox, so
+//!   its own process can write a new `tools/*.toml` any time after install
+//!   completes. [`dirs_of`] re-runs `reconcile` against the disk on **every**
+//!   load, not only the one `stella plugin install` performs, so a directory
+//!   that drifted from the consented manifest after install contributes
+//!   nothing rather than loading silently on the next session
+//!   ([`reconciles_with_disk`]).
 //! - **Retraction.** Structural, per above.
 //!
 //! # The trust gate is the roster's, not a second one
@@ -110,7 +119,10 @@ pub(crate) struct ContributedDir {
     pub(crate) dir: PathBuf,
 }
 
-/// Every plugin's contributed directory of one kind, in roster order.
+/// Every plugin's contributed directory of one kind, in roster order,
+/// withholding any plugin whose directories no longer reconcile against what
+/// its manifest declared and a human consented to (see
+/// [`reconciles_with_disk`]).
 ///
 /// Roster order is name order, and it is the precedence order for
 /// plugin-versus-plugin name collisions: the first plugin to claim a name
@@ -120,11 +132,43 @@ fn dirs_of(roster: &PluginRoster, kind: &str) -> Vec<ContributedDir> {
     roster
         .plugins()
         .iter()
+        .filter(|plugin| reconciles_with_disk(plugin))
         .map(|plugin| ContributedDir {
             plugin: plugin.manifest.name.clone(),
             dir: plugin.dir.join(kind),
         })
         .collect()
+}
+
+/// Re-run, on every load, the check `stella plugin install` runs once.
+///
+/// [`PluginManifest::reconcile`](stella_plugin::PluginManifest::reconcile) —
+/// a package's `tools/`, `skills/` and `rules/` directories must match
+/// exactly what the manifest declared and a human consented to — used to be
+/// invoked only from `stella plugin install` (`plugin_cmd.rs`), which makes
+/// "the shown document is provably complete" true for exactly one instant:
+/// install time. A plugin runs as an ordinary OS subprocess with no
+/// filesystem sandbox beyond the env-var allowlist
+/// (`crates/stella-runtime/src/wrapper/subprocess.rs`), so its own process —
+/// or any process running as the same user — can write a new
+/// `<plugin_dir>/tools/backdoor.toml` or `<plugin_dir>/rules/*.toml` at any
+/// point after install completes. Without this check, the next session's
+/// [`dirs_of`] would read that new file straight off disk with no new
+/// consent prompt anywhere on the path. Running the same check here, on
+/// every call, is what makes the guarantee hold for every session a plugin
+/// is loaded into rather than only the one that installed it.
+///
+/// A plugin that fails reconciliation contributes **nothing** of any kind —
+/// not a partial read of whatever entry still happens to match — because a
+/// package that grew one undeclared entry is not a package whose other,
+/// still-declared entries can still be trusted to be the ones consented to.
+/// The whole package is withheld until it is reinstalled against what is
+/// actually on disk, which re-shows the declaration and asks again.
+fn reconciles_with_disk(plugin: &InstalledPlugin) -> bool {
+    plugin
+        .manifest
+        .reconcile(&Inventory::of_package(&plugin.dir).listing())
+        .is_ok()
 }
 
 /// The roster this workspace's session runs under, or an empty one.
--- a/crates/stella-cli/src/plugin_cmd/tests.rs
+++ b/crates/stella-cli/src/plugin_cmd/tests.rs
@@ -899,6 +899,74 @@ fn a_plugins_tool_installs_runs_as_the_plugin_and_retracts_with_it() {
     let _ = std::fs::remove_dir_all(&root);
 }
 
+/// **The re-check witness.** `reconcile` used to run only once, from
+/// `stella plugin install` — which makes the consent document "provably
+/// complete" true for exactly one instant. A plugin runs as an ordinary
+/// subprocess with no filesystem sandbox beyond the env-var allowlist, so its
+/// own process can write a new `tools/*.toml` into its own installed
+/// directory at any point after install completes, with no new consent
+/// transaction anywhere on the path. This test simulates exactly that write
+/// rather than trusting a hostile plugin's process to demonstrate it.
+///
+/// Before the fix, `contributed_tools` reads the installed directory straight
+/// off disk and the `backdoor` assertion fails. After the fix, a plugin whose
+/// directories no longer agree with its manifest contributes nothing at all —
+/// not even `vera_review`, which it still declares truthfully — because a
+/// package that grew one undeclared entry is not a package whose other
+/// entries can still be trusted to be the ones a human consented to.
+#[test]
+fn a_plugin_that_drifts_from_its_declaration_after_install_contributes_nothing() {
+    let _env = crate::test_env::lock();
+    let _restore =
+        crate::test_env::EnvRestore::capture(&["STELLA_TRUST_PROJECT", "STELLA_PROJECT_HOOKS"]);
+    let root = temp_root("package-drift");
+    let _paths = crate::paths::test_user_home(root.join("home"));
+    // SAFETY: the env lock is held for the whole mutate-read-restore window.
+    unsafe { std::env::set_var("STELLA_TRUST_PROJECT", "1") };
+
+    let source = package(&root, "vera");
+    ships_a_package(&source, "vera");
+    install(
+        &root,
+        &source,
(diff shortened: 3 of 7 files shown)


macanderson:
CI is red on **`cargo check on the declared MSRV`** and **`fmt + clippy + test`**, and neither is caused by this diff — `main` is red at `eb7c2dd48`, the commit this branch is based on.

Both jobs fail on the same step, `Cargo.lock` resolving against the manifests:

```
error: the lock file /home/runner/work/stella/stella/Cargo.lock needs to be
updated but --locked was passed to prevent this
```

Reproduced locally against a **pristine worktree of `eb7c2dd48`** with none of my commits in it:

```sh
git worktree add /tmp/mainchk eb7c2dd4
cd /tmp/mainchk && cargo metadata --locked --format-version 1   # exit 101
cargo metadata --format-version 1 && git diff Cargo.lock
```

The entire diff cargo wants is one line:

```diff
 [[package]]
 name = "stella-transcript"
-version = "0.9.77"
+version = "0.9.83"
```

That is the check failing on the base branch, not on my change — `git diff eb7c2dd4..HEAD -- Cargo.lock Cargo.toml 'crates/*/Cargo.toml'` is empty, so neither of my two commits can move that line. The canary filed it as #3594 and #3595 is the one-line repair, already open against this same base.

I have deliberately **not** folded the lockfile fix into this PR, even though it would turn my checks green immediately — #3594's own instructions and AGENTS.md § Gotchas both say to land it on its own from a fresh `main`. Once #3595 merges I will rebase onto it and CI should go green with no change to the commits here.

One correction to this PR's own description while I'm here: it claims `cargo clippy --workspace` passed. What I actually ran was `cargo clippy -p stella-runtime --all-targets -- -D warnings` plus the two crates' test suites — not the whole workspace, which is what the box implies. The workspace-wide run is CI's, and it has not yet had a green base to report against.

---

macanderson:
Correcting a claim I made in this PR's description, because `1db8f216` proves it was wrong.

I wrote that `cargo test -p stella-cli (1801)` passed. **I never ran that suite.** What I ran was `cargo test -p stella-cli a_plugin_that_drifts` — a filtered run that reported `1 passed; 1800 filtered out` — and I read the total off the filter line and presented it as a passing count. One test ran. The other 1800 did not.

That is exactly what let the stale fixture through. `a_promotion_grant_does_not_arm_a_record_a_plugin_shipped` hand-built a package whose manifest declared nothing while `rules/set.toml` shipped a record; my `reconciles_with_disk` change withholds that package whole, so the record never reached the registry. The full suite would have caught it before the push, and `1db8f216` had to fix it after.

The fixture change is right, and for the right reason: the test failed at the `.expect` on *finding* the entry rather than at its assertion, so the new behaviour is correct and declaring the record keeps the witness testing what it was written to test.

Having now actually run it, on `1db8f216`:

| Run | Result |
|---|---|
| 1 | `1804 passed; 1 failed` — `daemon::tests::a_run_that_ignores_term_is_killed_after_the_grace_period` |
| 2 | `1805 passed; 0 failed` |
| 3 | `1805 passed; 0 failed` |

That one failure is **not** from this branch. It is the known flake in #3559 and #3588 (filed twice today, closed four times before as #2393, #2804, #2828, #2902): the child is reaped ~82 ms into an 8 s grace period because it has not installed its `SIGTERM` handler yet under load. It passes alone in 8 s, and `daemon.rs` has no reference to `plugin_cmd` or the wrapper, so nothing in this diff can reach it. I have not filed a third duplicate.

The description's gate section now carries these numbers instead of the ones I inferred.

---

Issue #3631, opened 2026-08-18
The brand tagline claims Stella is "faster, cheaper, and more accurate" but never says than what


macanderson:
## Problem

The brand tagline is:

> the terminal agent — faster · cheaper · more accurate

Faster, cheaper and more accurate **than what?** The claim names no
comparator, cites no measurement, and links to nothing. It is a three-way
comparative asserted flat.

It is also the single most-distributed sentence the project ships. It renders
on:

- `website/src/app/opengraph-image.tsx:143` — the Open Graph card, i.e. every
  time anyone pastes a stella link into Slack, X, LinkedIn or iMessage
- `website/src/app/manifest.ts:13` — the PWA install description
- `docs/brand/brand-guidelines.html:143` and `:368` — the kit, which is where
  the other two copied it from
- `docs/brand/pwa/manifest.webmanifest` — generated from the kit
- `docs/brand/social/build_social.py` output — the profile banners

## Why this is worth a decision rather than a shrug

`CLAUDE.md` holds bench conclusions to a standard this sentence does not
meet by a wide margin:

> **`Measure honestly, especially when it costs us.`** A benchmark number that
> flatters Stella because of a measurement artifact is worse than a loss: it
> spends the exact reputation the project exists to build.

and

> **Every factual claim ships with the evidence that establishes it, or it is
> not made.**

The rest of the published surface actually clears that bar, which is what
makes this one stand out. The only benchmark figure on the marketing site is
`website/src/app/engine/stations-aft.tsx:177` — the 46%-agreement number for
the model verdict that was *removed* — an unflattering result about our own
feature, reported plainly. The evolution deck labels each engine `ships` /
`ships in slices` / `in design` with an issue link rather than implying all
three are done. Against that, an unsourced "faster · cheaper · more accurate"
on the OG card is the weakest sentence we publish, and it is the one the most
people see.

There is a second-order cost: a reader who checks one leg and finds it
unsupported has been given a reason to discount the 46% number too, even
though that number is the accurate, unflattering one.

## Why it was not simply rewritten

It is the brand's registered tagline, defined in the kit, drawn into
generated social art whose builder asserts it, and part of the identity
rather than of any one page's copy. Changing it is a brand decision, not a
copy edit — `CLAUDE.md`'s "'Now' vs. 'right' is the maintainer's call, not
yours" applies squarely. Flagging it rather than quietly changing it is the
point of this issue.

## Options to pick between

[...]

Issue #3630, opened 2026-08-18
design-system: the shared transcript page is outside the instrument parity test, and its categorical hues are an open question


macanderson:
## Problem

`crates/stella-transcript/src/html/transcript.css` styles a page the
Observatory serves: `render_transcript` in
`crates/stella-observatory/src/lib.rs` calls `html::render_page`, so a reader
clicking a run in the dashboard lands on it. It is also the standalone
artifact people mail around.

It was outside `crates/stella-cli/tests/design_token_parity.rs` and had
drifted accordingly — ground `#000` against the instrument's `#0A0A0A`,
and a different green/red/amber, so "passed" was one colour in the dashboard
and another in the transcript of the same run.

The **values** were aligned to the instrument palette in the same change that
filed this issue (ground, neutral ramp, semantic triad, and the diff washes
derived from the new pair). What is left is the part that could not be done
safely in that scope.

### 1. The page is still not in the parity matrix

`design_token_parity.rs`'s `Surface` struct requires a **dark and a light**
scheme per surface (`dark: (start, end)`, `light: (start, end)`), and
`transcript.css` declares only one `:root` block — there is no light scheme
and no `prefers-color-scheme` override. So the file cannot be added as a row
without either:

- **(a)** giving the transcript a light scheme, matching what every other
  instrument surface carries (`crates/stella-mcp/src/oauth/callback_page.css`
  is the smallest example — both schemes at `:root`, media query overriding
  only); or
- **(b)** extending `Surface` to allow a dark-only surface, e.g. making
  `light` an `Option`, and declaring *why* this one has no paper ground.

(a) is the better answer if the page is ever printed or read on a bright
screen, which a mailed-around artifact is. (b) is the smaller change and is
honest if the transcript is deliberately ink-only.

Until one lands, the alignment done here is a one-time sync with nothing
stopping it drifting again — which is exactly the failure #2453 was filed
about.

### 2. `--cyan` and `--violet` encode categories by hue

They are deliberately **not** instrument values and were left untouched:

```
--cyan: #4ed6e0;    /* tool kind: bash, running, focus ring, "you" roletag */
--violet: #a8a3f5;  /* tool kind: search; "agent" roletag; answer rule */
```

The instrument system's documented answer for categorical data is a **value
ramp** (`--c1`..`--c4` in `crates/stella-observatory/src/assets/index.html`),
separated by lightness because that is the one channel that survives
greyscale, colour-vision deficiency and a projector. The transcript separates
[...]

Issue #2594, opened 2026-08-10
design-system: the docs site, marketing site and slide decks are outside the instrument parity test and now diverge further


macanderson:
## Problem

The web instrument system now covers every HTML surface stella *renders*:

| surface | file |
| --- | --- |
| Observatory | `crates/stella-observatory/src/assets/index.html` (single definition) |
| telemetry export | `crates/stella-cli/src/export.rs` |
| arenabench client | `arenabench/ui/app/globals.css` |
| arenabench control plane | `arenabench/web/app/globals.css` |
| OAuth landing | `crates/stella-mcp/src/oauth/callback_page.css` |
| tracker-auth landing | `crates/stella-tools/src/tracker_auth_page.css` |

All six are held together by
`crates/stella-cli/tests/design_token_parity.rs` (closes #2453).

**The published surfaces are not in that set** and now diverge further than
they did before, because the instrument surfaces grew an `--identity` token
that the website does not have:

- `website/src/app/tokens.css` — mirrors the comet kit for the docs site
- `website/src/app/global.css` — maps the kit onto fumadocs variables
- `website/src/app/releases/releases.css`
- `website/public/presentations/*.html` — slide decks on a fixed 1600x900
  canvas, measured by `.github/workflows/deck-fit.yml`
- `docs/brand/site/stella-site-mock.html`, `docs/brand/site/stella-mobile-mock.html`
- `docs/benchmarks/*.html` — the house benchmark reports, which are the
  *reference* the instrument palette was drawn from and so may already agree

## Why it is not simply "add them to the parity test"

The published surfaces are legitimately not instruments. A marketing page has
no data for its chrome to compete with, which is the whole reason the
identity-only rule exists — so a docs page or a deck may well be entitled to
more gold than the Observatory is. The question this issue has to answer is
**which roles are shared and which are surface-specific**, and only the shared
ones belong in the parity test.

A reasonable split to evaluate: grounds, text ramp and the semantic three are
shared; chrome accent and identity usage are per-surface, declared the way
`provider_parity.rs` declares a posture per provider rather than asserting
sameness.

## Where to start

- `crates/stella-cli/tests/design_token_parity.rs` — `ROLES`, `Surface`, and
  the `names` mapping are the extension point; a surface already declares only
  the roles it carries
- `docs/brand/css/tokens.css` — the brand's normative source, and see #2591
  for a measured defect in it
- #2592 — the same "one rule, two subsystems" question for the TUI

## Done when

[...]

Issue #2592, opened 2026-08-10
design-system: gold means "running" in the TUI and "identity only" on the web, and neither rule mentions the other


macanderson:
## Problem

After the web surfaces moved to one instrument system, stella has **two
different rules for what gold is allowed to mean**, and neither file
acknowledges the other.

**The web instruments** (`crates/stella-observatory/src/assets/index.html`,
`crates/stella-cli/src/export.rs`, `arenabench/{ui,web}/app/globals.css`):

> gold is permitted on identity — the wordmark, the favicon mark, at most one
> primary action — and it may **never encode a state**.

**The TUI** (`crates/stella-tui/src/theme.rs`, the doc comment on `GOLD`):

> **Never a verdict**: it sits 4.0° from `WARN` in hue, and
> `gold_never_carries_a_verdict` proves no outcome mapping can return it —
> activity (running/active) **is the one status that takes gold**, by kit rule.

So gold marks *running* in the terminal and marks *nothing but identity* on the
web. A user who runs `stella` in one pane and has the Observatory open in
another sees the same colour mean two different things.

## Which is right is a real question, not a formality

The TUI's position has an argument the web's does not: a terminal cell has no
opacity, no subpixel control and a 16/256-colour floor, so the ink/paper
inversion the web uses for "active" costs a whole cell attribute that may
already be spoken for — and the TUI pairs every status with a glyph anyway, so
hue is not load-bearing there.

The web's position has the measurement: `--identity` `#FFB000` against `--warn`
`#C9A227` is **1.32:1**. The TUI states the same fact in degrees ("4.0° from
WARN in hue") and then permits the overlap regardless, mitigated by glyphs.

Both can be defensible. What is not defensible is that the divergence is
currently invisible: neither file mentions the other's rule, so the next person
to touch either one will not know they are choosing.

## Where

- `crates/stella-tui/src/theme.rs` — `GOLD` doc comment (~line 171), `ACCENT`,
  and `gold_never_carries_a_verdict`
- `crates/stella-tui/src/palette.rs` — `BRAND` / `GOLD` share a value under the
  comet kit; the module header explains why
- `crates/stella-observatory/src/assets/index.html` — the identity carve-out is
  stated in the `Instrument tokens` header comment
- `crates/stella-cli/src/export.rs` — same rule, restated in its `<style>` header
- `crates/stella-tui/tests/snapshots/deck/` — golden frames pin the terminal
  character-by-character; any TUI recolour changes them and each change must be
[...]

PR #3635, opened 2026-08-18
fix(website): ship brand kit v2.0, and stop claiming a model confirms a run

macanderson:
## What & why

A sweep of the outward-facing surfaces for palette, logos and product-claim
truthfulness. It found two defects that share a shape: **in both, a paragraph
asserted the opposite of what shipped.**

### 1. The site was a whole brand version behind

`docs/brand/` moved to Bronze Gold `#C58A32` on Ink `#10100F` in the
2026-08-11 rebrand. The website stayed on v1.0's Phosphor Gold `#FFB000` on
Ink `#0B0B0C` — while `website/src/app/tokens.css` said, the entire time:

> every value below is copied from it verbatim — do not tune a hex here,
> change the kit and mirror it

Stale with it: **all thirteen** logo SVGs under `public/brand/`, **seven of
the eight** PWA icons, `favicon.ico`, `apple-icon.png`, `src/app/icon.svg`,
the OG card's Satori literals, `vision.html`, and the investor deck. The
practical result is that the site served a **bronze wordmark beside phosphor
chrome** — the assets and the CSS disagreed with each other on the page.

The kit is a shared cell that several surfaces copy, and a comment cannot
hold one. So the claim is now enforced rather than promised.

Other surfaces named in #2594 (`docs/brand/site/*.html`, `docs/benchmarks/*`,
the auth landing pages, the arenabench UIs) were checked and are clean — the
Observatory and the OAuth page were already on bronze. The website was the
only drifted one.

### 2. The landing page credited a rung the pipeline does not have

It said a run:

> ends only when a second model has confirmed the goal from evidence

#2584 removed the model verdict outright. `LadderDecision`
(`crates/stella-pipeline/src/verify.rs`) carries no arm that asks a model
anything — a turn the oracle cannot settle reports `UNVERIFIED` rather than
borrowing a model's opinion. The one surviving verifier-tier call is the
witness *author*, and what it writes is then decided by **running** it.

`stella goal` does still end on a verifier model's judgement
(`crates/stella-core/src/goal.rs`), but it is one mode and the front page was
generalising it to the whole product. The docs pages were already scrupulous
about this — `inference-pipeline.mdx` even reports the unflattering 46%
agreement figure that motivated the removal. The landing page was the
outlier, in both its `HOME_DESCRIPTION` (which is also the OG/Twitter card
text) and its visible hero paragraph.

The same page's pipeline door listed the stages as
`triage, plan, witness, execute, verify, verifier` — witness and execute
swapped, ending on a role that is not a stage and no longer judges. Both now
match `stage_rank` in `crates/stella-pipeline/src/replay.rs`.

### 3. The shared transcript surface

The Observatory renders `stella_transcript::html::render_page` as a page
(`render_transcript` in `crates/stella-observatory/src/lib.rs`), so clicking
a run moved the reader from the instrument ground `#0A0A0A` to `#000`, with
"passed" a different green in two views of the same run. Its ground, neutral
ramp, semantic triad and diff washes are now the instrument palette. Its
categorical hues and its absence from the parity matrix are left as
**declared, issue-cited gaps** rather than half-migrated — see #3630.

### Contrast was recomputed, not carried over

Several annotations were wrong *before* the recolour — `--color-fd-primary`
was documented `9.6:1 on ink` and measures `6.4:1`. Every kit rule still
holds on v2: gold-deep on paper **5.0:1**, gold on paper **2.6:1** (which is
exactly why gold-deep exists), gold on ink **6.4:1**, muted on ink **6.8:1**.

Refs #3630
Refs #3631

## The witness

`website/src/lib/brand-parity.test.ts` — 7 cases holding the site's ramp,
logo SVGs, app icon, PWA icons and manifest against `docs/brand/`, plus a
sweep asserting no retired v1.0 value survives anywhere under `website/`.

Checked the artisanal way, by stashing only the fix and re-running:

```
$ git stash push -- website/src/app/tokens.css website/public/brand
$ node --test 'src/lib/brand-parity.test.ts'
not ok 1 - the site's brand tokens are the kit's, value for value
not ok 2 - no v1.0 phosphor value survives anywhere in the site
not ok 3 - the site's logo SVGs are byte-identical to the kit's
# pass 4
# fail 3

$ git stash pop && node --test 'src/lib/brand-parity.test.ts'
# pass 7
# fail 0
```

It earned its keep during authoring: it caught an `rgb(255 176 0 / 0.06)` in
`vision.html` that the hex-based pass had missed.

The landing-page copy changes carry no witness — a marketing sentence has no
test — so they are evidenced by citation instead: each is annotated in-file
with the code that settles it (`LadderDecision`, `stage_rank`), so a future
reader can check the claim against the mechanism rather than against taste.

**This is deliberately not #2594 reopened.** That issue asked whether the
published site should adopt the *instrument* palette and was closed `wontfix`
on the correct ground that a marketing page is legitimately not an
instrument. Nothing here touches that: the site keeps its own chrome, accent
usage and status hues. The only thing asserted is that the ramp it already
calls normative is the ramp it actually ships. The test says so in its header
so it is not mistaken for a decision reversed by stealth.

## The gate

Run in this container, all green:

- [x] `make brand-case`, `make left-behind`, `make file-size`, `make god-files`,
      `make doc-links`, `make invariants`, and the first seven `guards-fast` steps
- [x] `cargo test -p stella-transcript` — 36 passed
- [x] `cargo test -p stella-cli --test design_token_parity` — 7 passed
- [x] `node --test 'src/**/*.test.ts'` (website) — 25 passed
- [x] Docs updated where behaviour changed — the token files, `brand.tsx`,
      `global.css` and `transcript.css` headers all now describe what ships

⚠️ **`make gate` did not run to completion here:** `shellcheck` is not
installed in this container, so `guards-fast` aborts at that step and the
tiers above it never start. `cargo fmt --check`, `clippy -D warnings` and
`cargo test --workspace` are therefore **unverified by me** — I ran the two
directly-affected crates instead. Worth a full local `make gate` before this
leaves draft. The diff is CSS/TSX/asset-heavy with one Rust-adjacent file
(`transcript.css`, which is `include_str!`-ed and has no Rust changes), so
the risk is low but it is not zero and I would rather say so than imply a
green gate.

## Nothing left behind

- [x] Filed: #3630, #3631

- **#3630** — the transcript page is outside `design_token_parity.rs` and its
  `--cyan`/`--violet` encode categories by hue where the instrument system
  uses a value ramp. Adding it needs either a light scheme or a dark-only
  `Surface`; a follow-up comment records that it would also fail
  `no_web_surface_ships_a_rounded_corner` on day one, which makes the
  decision larger than the title suggests.
- **#3631** — the tagline *"the terminal agent — faster · cheaper · more
  accurate"* is a bare three-way comparative with no referent, on the OG card
  and the PWA manifest, i.e. the most-shared sentence the project has. It is
  the brand's registered tagline defined in the kit and drawn into generated
  social art, so **changing it is a maintainer's call, not mine** — it is
  flagged rather than rewritten, with three options costed.

Note also that `docs/brand/social/build_social.py --check` asserts the
install line against `README.md` but does **not** assert the tagline against
the kit, so nothing currently fails if those copies diverge. Captured in
#3631.

## Ground-rule check

- [x] No cross-boundary types added

**The pre-existing `#2592` divergence is untouched and intentional.** The TUI
holds Phosphor Gold `#FFB81A` deliberately — `palette.rs` documents that it
does *not* track the kit's bronze, because a terminal needs a brighter stop
to read as CRT amber on ink at 11.36:1. That is a reasoned, issue-cited
choice, so this PR left it alone; it is not the same thing as the website's
silent staleness. After this PR the surfaces are: kit + website on bronze,
TUI on phosphor by declaration (#2592), Observatory monochrome with bronze
identity by declaration.

**Deleted tests:** none.

**The riskiest part of the diff** is the `global.css` neutral warming
(`#9b9890`→`#a19a8e`, `#232327`→`#2c2924`, `#1a1a1d`→`#171614`,
`#070708`→`#0a0a09`), which is a bulk `sed` across ~25 rules. It is
mechanical and every replacement is the kit's own token value, but it is the
change most worth eyeballing in a browser, since no test renders the page.

---
_Generated by [Claude Code](https://claude.ai/code/session_01AzoYpGn8Sg9FPbc319rgpv)_

## Summary by Sourcery

Synchronize outward-facing surfaces with brand kit v2.0 and accurately describe how runs are verified.

Bug Fixes:
- Update the website and transcript surfaces to use the current brand-kit palette and assets instead of stale or inconsistent colors.
- Correct landing-page claims to reflect evidence-based run completion and the actual pipeline stage order.

Enhancements:
- Align generated social imagery, PWA metadata, investor materials, vision page, and transcript styling with their normative palettes and recomputed contrast values.
- Add parity coverage that prevents drift between website branding assets and the shared brand kit.

Tests:
- Add website tests covering brand tokens, logo and icon parity, favicon encoding, manifest references, and retired palette values.


macanderson:
`typecheck + build` and the Vercel deployment both failed on the first push. **My regression, now fixed in `3b5d134`** — and the PR body's gate section is out of date in one direction as a result, so correcting the record here.

**Cause.** The asset sync copied `docs/brand/pwa/favicon.ico` in byte-for-byte. The kit renders its favicons *opaque* — the mark on an Ink tile — so its ICO embeds RGB PNGs (colour type 2). Next decodes `src/app/favicon.ico` through the `ico` crate, which accepts **RGBA only**:

```
Error: Turbopack build failed with 1 errors:
./website/src/app/favicon.ico
Caused by: Format error decoding Ico: The PNG is not in RGBA format!
```

The v1.0 file it replaced was colour type 6, which is why this had never come up. Vercel's review bot reached the same diagnosis independently.

**Fix.** That one file is now deliberately *not* a byte-copy: it is the kit's pixels re-encoded with an opaque alpha channel. Verified pixel-identical to the kit at all three sizes (16/32/48), dominant colour `(197,138,50)` = `#C58A32`.

**Guarded**, because the obvious tidy-up is to re-copy it from the kit and that is precisely what breaks the build. A seventh case walks the ICO directory and checks each entry's PNG colour-type byte; it fails on the kit's file naming the colour type, and passes here. That is also why `favicon.ico` is absent from the PWA byte-identity test — the exception is now asserted rather than silent.

**Correction to the PR body.** It said the website build was unverified by me because `make gate` aborts on a missing `shellcheck` in my container. That was true when written and is no longer: I installed the site's dependencies and ran the real toolchain this round rather than reasoning about it —

```
pnpm run build      → ✓ Compiled successfully
pnpm run typecheck  → clean (exit 0)
node --test         → 26 passed, 0 failed
```

The rest of that caveat still stands: `cargo fmt --check`, `clippy -D warnings` and `cargo test --workspace` remain unrun here, so a full local `make gate` before this leaves draft is still worth it. I should have built the site before the first push — checking the asset swap by inspecting hexes proved exactly the wrong altitude for a binary format.

---

macanderson:
`fmt + clippy + test` was red on `3b5d134`. Fixed in `ee01894`, and it is a good one.

```
tests::execution_routes::transcript_route_renders_one_node_per_call_stating_the_command_once
left: 2, right: 1
```

**Cause.** `transcript.css` is `include_str!`-ed and inlined into every rendered transcript, so **its comments are page content**. The header paragraph I added last commit cited `render_transcript` by source path — and the seeded fixture's tool call reads that same path, which the Observatory counts across the whole document to prove an invocation is stated exactly once. So a comment about the transcript's *colours* failed a test about its *structure*, by adding a second occurrence of a string in prose.

**Fix.** Cite by name, not by path. The trap is now written down in the file itself, where the next person editing that header will be standing — comments here ship to the reader and are visible to any assertion made over the document.

**Witness, both directions, locally:** reintroducing the literal reproduces `left: 2 / right: 1` exactly; removing it passes.

**On verification.** Twice now I have pushed after checking only the crates I had *reasoned* were affected, and twice the failure was in a crate I had argued was untouched — `stella-transcript` has no Rust change in this PR at all, and its stylesheet still broke `stella-observatory`. `include_str!` makes "which crate did I change" the wrong question; the right one is "what does this file get compiled into". So this round I ran the tiers rather than reasoning about them:

```
cargo test --workspace                              → green
cargo fmt --all --check                             → clean
cargo clippy --workspace --all-targets -- -D warnings → clean
pnpm run build / typecheck / node --test            → green (26 tests)
```

That closes the caveat in the PR body: the Rust gate is no longer unverified by me. `shellcheck` is still missing in this container, so `make gate` as a single command still cannot complete here — but every step it wraps has now been run except the shell-lint itself, and this PR touches no shell.

Both earlier failures were mine and neither was subtle in hindsight; the pattern in both was checking a proxy (hex strings, crate names) instead of the artifact.

---

Issue #3634, opened 2026-08-18
B0 remainder: the driver channel has no transport — nothing opens a driver session


macanderson:
## Problem

B0 (#3599) landed the driver channel's two halves — the wire
(`crates/stella-plugin/src/driver.rs`) and the gate
(`crates/stella-runtime/src/wrapper/driver_call.rs`) — but **nothing transports
between them**. `DriveRequest` is never written to a process, `DriverMessage` is
never read from one, and `DriverCallGate::open` has no caller outside its own
tests. The wrapper socket's equivalent plumbing is
`crates/stella-runtime/src/wrapper/subprocess.rs` (JSON over stdio, with the
conversation relayed through `PointChannel`) and `in_process.rs`; the driver
channel has neither.

Filed per AGENTS.md § "Nothing left behind": B0's own scope was the dispatch
context and the grant, and the transport is the next slice rather than a
forgotten one — but unwired code with no tracking issue is exactly what that
rule exists to prevent.

## Where

- `crates/stella-plugin/src/driver.rs` — `DriveRequest`, `DriverMessage`,
  `DriverCallRequest`, `DriverCallResponse`.
- `crates/stella-runtime/src/wrapper/driver_call.rs` — `DriverCallGate`,
  `DriverSession`, `DriverCapabilities`, `NoDriverCapabilities`.
- The shape to copy: `crates/stella-runtime/src/wrapper/subprocess.rs`, whose
  `HostCallGate`/`PointChannel` relay is the same problem one context over.

## Constraints already discovered

- The per-session timeout must cover the **whole** conversation, not each turn
  of it — a driver must not be able to buy time by talking
  (`doc:backlog-self-driving` §3.0, `doc:wrapper-socket` §6b bound 1).
- `DriverCallGate::offers_calls()` exists for the transport's benefit: a driver
  that can never be served must not be left with stdin held open until the
  session timeout. Consult it before opening a session, as
  `subprocess.rs` does for `HostCallGate::offers_calls`.
- A refused call is answered, never fatal. The transport must relay the `err`
  and keep the session open.
- `stella-runtime` reads no ambient environment by contract
  (`crates/stella-runtime/tests/no_ambient_reads.rs`).
- A driver call carries no `args` and returns no payload at B0, deliberately —
  see the header of `crates/stella-plugin/src/driver.rs`. The transport should
  not invent either; the verbs bring them in B1-B6.

## Done when

A driver process can be spawned, handed a `DriveRequest` over stdio, make a
[...]

Issue #3599, opened 2026-08-17
Epic: the plugin system can only grant capabilities to plugins running inside a turn, so the self-driving loop can never get any


macanderson:
The spec anchor for `doc:backlog-self-driving`
([`docs/design/backlog-self-driving.md`](https://github.com/macanderson/stella/blob/claude/autonomous-issue-backlog-design-ym44oh/docs/design/backlog-self-driving.md)).
Nothing below is built.

## The defect

Stella's self-driving loop can decide **how much** work to do and **whether it
is done**. It cannot decide **what** the work is, **do** it, or **ship** it.
Those three verbs live outside the product, in eight Claude Code slash-command
markdown files under `scripts/self-driving/commands/`.

The cause is not packaging — self-driving already *is* a plugin
(`plugins/stella-selfdriving/plugin.toml`). Since #3590 the platform has a
precise answer to "how does a plugin get a capability": the **host-call
channel** (`crates/stella-plugin/src/host_call.rs`), whose principle is *"a
plugin may ask the host for a capability, and may never reach for one"*. One of
its three capabilities is `child_turn` — a bounded turn at a declared role
intent, which the host resolves, budgets, runs and settles. **That is the
do-the-work capability, already built and already gated.**

Self-driving is excluded from it twice over:

```rust
pub fn permits_call(&self, call: HostCall) -> bool {
    self.participation.includes(Participation::Steering) && self.calls.contains(&call)
}
// crates/stella-plugin/src/manifest.rs:213
```

1. **By grade.** `plugins/stella-selfdriving/plugin.toml` declares
   `participation = "none"`; `None.includes(Steering)` is false, so every call is
   refused before the manifest's `calls` list is consulted.
2. **By construction** — the deeper one. A host call may be made "during
   `before_turn` or `after_turn` and nowhere else". Those are *inside a turn*.
   Self-driving drives turns; raising its grade would not help, because there is
   no turn for it to stand in.

The `Participation` ladder (`none` → `observer` → `steering` → `arbiter`) is a
ladder of **in-turn influence**, and self-driving is not on it at all. **The
platform has a well-designed way to hand capability to a plugin that
participates in a turn, and no way at all to hand one to a plugin that drives
turns.** That is why the judgement half ended up outside the product: the only
remaining route in is `HOST_SURFACE`, and all 18 of its verbs are bookkeeping.

## What is absent, with evidence

| Absent | Evidence |
|---|---|
[...]

Issue #3546, opened 2026-08-17
Track D remainder: the self-driving driver still lives in scripts/, and its declared grant binds to no gate


macanderson:
## State after the D2+D3 slice

`doc:pipeline-as-plugins` §10 splits into D1–D6. What is done and what is not, verified rather than assumed:

- **D1 — done** (5c5c325). The decision core is `crates/stella-autonomy`.
- **D2 — done.** The host-verb surface is declared as data (`crates/stella-autonomy/src/surface.rs`, `HOST_SURFACE` + `HOST_SURFACE_VERSION`), printed by `stella self-driving surface [--format json]`, and checked against the shipping clap tree in both directions (`crates/stella-cli/src/self_driving_cmd/surface.rs::surface_matches_the_argument_tree`, over `stella_autonomy::surface_drift`).
- **D3 — done as far as D-3 requires.** The grant is expressible and showable before install: `plugins/stella-selfdriving/plugin.toml`, rendered by the real `stella plugin install` and witnessed by `crates/stella-cli/tests/self_driving_consent.rs`.
- **D4 — verified, no change needed.** `/api/self-driving` and `/api/self-driving-run` were driven against a state directory written by the current binary and both render (loops, runs, cycles, trajectory, self-improvement signals).
- **D5 — green, assertions intact.** `bash scripts/test-self-driving.sh` reports 60/60.
- **D6 — not started**, because nothing moved yet.

## What is left

**The extraction itself has not happened.** `scripts/self-driving.sh` (987 lines) is still the working driver, and §10's rule is that it is not deleted until its replacement is proven. `plugins/stella-selfdriving/` currently holds only a manifest and a README: installing it copies a declaration and starts nothing.

**The declared grant binds to nothing.** `plugins/stella-selfdriving/plugin.toml` declares `bash`/`write_file`/`process_spawn` at destructive/high, and no `AuthzGate` rule is derived from it under `Principal::Plugin`. That is §A4's loader work, not this slice's. Until it lands, the consent prompt is an accurate description with no enforcement behind it — which is exactly what the prompt's own "claimed limit" sentence tells the user, but it is a gap and not a design.

[...]

PR #3637, opened 2026-08-18
feat(stella-plugin): the driver channel — a plugin at grade `none` can hold a capability (#3599 B0)

macanderson:
#3599 B0, the load-bearing phase: until a driver-posture plugin can hold a
capability, none of B1-B7 can be built.

## The defect

`LoopGrant::permits_call` requires `participation >= Steering`, and a host call
may be made during `before_turn` or `after_turn` and nowhere else. Both are
conditions about standing *inside* a turn. A driver has no turn to stand in —
it is what starts them — so no grade on the `Participation` ladder would let it
in, and `plugins/stella-selfdriving`, which declares `participation = "none"`
honestly, was structurally unable to hold any capability at all. That is why
the judgement half of the loop ended up outside the product, in eight Claude
Code markdown files.

## The fix

A second dispatch context over the existing host-call machinery. Same
refusal-as-a-value discipline, same closed refusal codes (`HostCallRefusal` is
reused rather than mirrored — the reasons a host declines are properties of
asking, not of who asked), and a separate `[driver]` block with its own `calls`
list so the `Participation` ladder is neither extended nor renumbered. Widening
the ladder instead would have made `arbiter` — already "the strongest grant"
over a turn's completion — silently also mean "may merge to `main`".

- `crates/stella-plugin/src/driver.rs` — the wire half: `DriverCall` (the
  eighteen verbs `doc:backlog-self-driving` §3.1-§3.5 names, one variant per
  verb because invariant 9 governs a capability as it governs a tool),
  `DriverGrant` + `permits_call`, and the `drive` session framing over the same
  denying envelope `PluginMessage` uses.
- `crates/stella-runtime/src/wrapper/driver_call.rs` — the host half:
  `DriverCallGate`, a per-session allowance clamped against the host's ceiling,
  refusals recorded so a host can say what a driver did not get, and
  `NoDriverCapabilities` as this host's honest state.
- Install consent renders the driver grant by family, and — separately — now
  renders `[loop] calls` at all. `LoopGrant::calls` has always said a plugin
  "may not ask for a capability a human never read at install"; nothing printed
  it, so the second half of that sentence was not true.

## A call carries no arguments and returns no payload

Deliberate, and stated in the module header rather than left to be inferred. No
verb is implemented at B0, so no argument or result table can be typechecked
against anything — and each would change at the phase that implemented it,
breaking the channel once per family. What B0 fixes in place is the part
consent depends on: which capabilities exist, which of them a driver declared,
and what it is told when it asks for one it did not. `DriverOk` is an empty
denying table and grows a field per verb that has something to report.

## Witnesses

- `an_undeclared_call_is_refused_and_the_session_keeps_running`
  (`driver_call.rs`) — both directions, because either alone is half a gate: an
  undeclared capability is refused with a code the driver branches on, is not
  charged against the allowance, and the session is still served afterwards.
- `a_driver_holds_capabilities_without_a_participation_grade`
  (`manifest.rs`) — the defect itself: `participation = "none"` with a
  `[driver]` block now permits a call. Fails on main, where there is no block.
- `the_prompt_shows_the_driver_grant_and_its_deliberate_limits`
  (`self_driving_consent.rs`) — the grant reaches stdout before install, and
  the two deliberate absences (`curate_accept` under `regulated` governance, no
  release verb at all) are asserted rather than assumed.

`docs/design/backlog-self-driving.md` is promoted to `docs/spec/` — the
`design-refs` guard's own remedy, since code now cites it — with §1.2, §2.1,
§3.0 and §9's B0 row corrected to what is built versus designed.

`plugins/stella-selfdriving/plugin.toml` declares its `[driver]` grant. It is
still not a program the host runs: that is B2, and the transport that would
carry a session is filed as #3634.

Refs #3599, #3546

## Summary by Sourcery

Enable plugins to hold and request explicitly consented capabilities through a dedicated driver channel independent of turn participation.

New Features:
- Add a separate driver capability channel that allows plugins to request host capabilities outside turn participation.
- Support manifest-declared driver grants with per-session call limits and explicit refusal handling.
- Expose driver session and capability wire messages, including sleep and halt outcomes.
- Render driver capability grants and loop calls in installation consent prompts.

Bug Fixes:
- Allow plugins with participation grade `none` to hold driver capabilities without widening the participation ladder.
- Ensure undeclared or unsupported driver calls are returned as refusals without terminating the session or consuming the allowance.

Enhancements:
- Reuse host-call refusal codes and provide host-side tracking of refused driver calls.
- Keep driver capabilities explicitly separate from in-turn loop permissions and omit unimplemented capability payloads at B0.

Documentation:
- Promote the self-driving backlog specification to a living B0 specification and update its implementation status and references.

Tests:
- Add coverage for driver grant validation, wire framing, refusal behavior, session allowance resets, and consent disclosure.

Chores:
- Declare the self-driving plugin's driver grant while retaining deliberate restrictions on curation acceptance and release operations.

### DIFF
--- a/crates/stella-cli/tests/self_driving_consent.rs
+++ b/crates/stella-cli/tests/self_driving_consent.rs
@@ -212,3 +212,59 @@ fn the_declared_grant_and_the_loop_name_the_same_powers() {
         );
     }
 }
+
+/// **The driver grant is shown before install, family by family.**
+///
+/// The `[driver]` block is the second consent this manifest carries and it is
+/// a materially different one from the `[loop]` grade above: a grade says how
+/// much say a plugin has *inside* a turn, and this says what it may ask Stella
+/// to do to your repository between them. A user who reads only the first has
+/// not been told about the second, which is the failure
+/// `doc:backlog-self-driving` §6.1 calls "the grant must actually bind".
+///
+/// It also asserts the two deliberate absences. Each is a decision recorded in
+/// the manifest, and a later widening should have to delete an assertion that
+/// says why rather than quietly adding a line to a list.
+#[test]
+fn the_prompt_shows_the_driver_grant_and_its_deliberate_limits() {
+    let text = consent_text();
+
+    assert!(
+        text.contains("DRIVES Stella"),
+        "the install prompt must say the plugin drives Stella, not just that it takes no say in a \
+         turn: {text}"
+    );
+    // Read by family, because "pushes branches, opens pull requests, reads CI,
+    // and merges" is the sentence a human weighs, and `deliver_merge` on its
+    // own is not.
+    for family in [
+        "reads and writes your issue tracker",
+        "runs Stella against an issue in an isolated worktree",
+        "pushes branches, opens pull requests, reads CI, and merges",
+        "runs audit tooling over this workspace and files what it finds",
+        "proposes new tools, context records and skills",
+    ] {
+        assert!(
+            text.contains(family),
+            "missing driver family `{family}`: {text}"
+        );
+    }
+    assert!(
+        text.contains("this plugin holds no credential"),
+        "the prompt must say every driver capability is performed BY Stella: {text}"
+    );
+
+    // `curate_accept` is withheld because this repository is governance mode
+    // `regulated` (§7), and no release verb is on the channel at all (§6.4).
+    assert!(
+        !text.contains("curate_accept"),
+        "the loop may propose an authority and grant itself none: {text}"
+    );
+    // Narrowly `release_`, not `release`: the `bash` capability above legitimately
+    // says "the release install", and asserting on the bare word would fail on
+    // a sentence that has nothing to do with the channel.
+    assert!(
+        !text.contains("release_"),
+        "a release verb is not on the driver channel at B0-B6: {text}"
+    );
+}
--- a/crates/stella-plugin/src/consent.rs
+++ b/crates/stella-plugin/src/consent.rs
@@ -192,6 +192,7 @@ pub fn consent_text(manifest: &PluginManifest) -> String {
 
     lines.push(String::new());
     lines.extend(loop_say(manifest));
+    lines.extend(driver_say(manifest));
     lines.extend(oracle_self_report(manifest));
     lines.push(String::new());
     lines.extend(capability_grant(manifest));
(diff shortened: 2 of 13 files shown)


macanderson:
I built B0 in parallel from the spec side (#3601's author) before finding this PR, so I have arrived at the same design independently — same `[driver]` block, same closed `DriverCall` set, same `permits_call` that deliberately does not consult `Participation`. **This PR is a superset of what I had** (it adds the runtime dispatch, the wire corpus, and the manifest declaration), so I have dropped my branch rather than open a competing one.

One reproduced defect, and two notes.

## The consent text calls the self-driving plugin a content bundle

Checked out this branch and rendered the real manifest:

```
$ cargo run -q -p stella-cli -- plugin install plugins/stella-selfdriving

Install `stella-selfdriving`?
...
Say in your turn loop: none — it never runs inside a turn. It is a content bundle (skills, commands, agents, tools)
  - runs at no hook point

Say outside your turn loop: `stella-selfdriving` DRIVES Stella. It is not a participant in a turn — it starts them.
  - reads and writes your issue tracker: ...
  - pushes branches, opens pull requests, reads CI, and merges (deliver_open, deliver_observe, deliver_next, deliver_merge)
```

The **first** sentence a human reads describes the most powerful plugin shape this vocabulary can express as *"a content bundle (skills, commands, agents, tools)"*. Four lines later the same document says it merges to `main`.

`grade_sentence` is untouched by this PR, which is correct for every plugin that existed before it — but `Participation::None` now means two different things, and the gloss only covers one. `consent.rs`'s own module docs name this exact plugin as "the most dangerous plugin anyone will install", and the file's standing argument is that a grant a user cannot read before installing is not meaningfully declared; a grant they read *and are actively mis-told about* seems worse.

The grade itself is honest and should not change — a driver genuinely never runs inside a turn. Only the gloss needs to know. What I had:

```rust
fn grade_sentence(participation: Participation, drives: bool) -> &'static str {
    match participation {
        Participation::None if drives => {
            "none — it never runs inside a turn, because it runs OUTSIDE them \
             (see below)"
        }
        Participation::None => { /* unchanged */ }
```

called as `grade_sentence(grant.participation, manifest.driver.is_some())`, with the witness asserting `!text.contains("It is a content bundle")` for a driver and `text.contains("It is a content bundle")` for a plain bundle, so the existing gloss stays pinned for the plugins it was written for.

## Two notes, both judgement calls rather than defects

**1. `[loop] calls` renders the capability names but not what they cost.** Good that this PR renders it at all — it was invisible before, and `stella-research` has shipped `calls = ["recall"]` with the prompt silent about it. Worth considering a per-capability gloss the way `[driver]` has one, so the wrapper channel reads like `recall (reads your indexed context and code graph)` rather than a bare identifier. Exhaustive `match` on the enum makes a new capability fail the build until somebody writes what it costs.

**2. `DriverCall` is seeded with the full delivery set** — `deliver_merge` included — while nothing serves any of them yet. That is the opposite tradeoff from the one this PR's own docs take ("nothing serves a `DriverCall` yet"), and `doc:backlog-self-driving` §1.3 argues a grant naming a capability no host can perform is a consent document making a promise nothing received. Not obviously wrong — declaring the whole surface once is more legible than growing it five times — but it does mean a user can today consent to "merges to `main`" against an implementation that cannot merge anything, which is the shape of claim #3546 was filed about. Worth a sentence in the PR description either way.

## What I did not find

`permits_call` not consulting `Participation` is right and is the crux. `DriverGrant` staying off the ladder is right for the reason given. The `[driver]`/`arbiter` combination is the one cross-block rule I also landed on, for the #2584 self-grading reason.

One difference worth flagging only because it may be deliberate: I required `[driver]` to come with a `[runtime]` (a driver with no process is a manifest that quietly does nothing), which forced widening `RuntimeRequiresObserver` — its reasoning is "a process nothing invokes would never start", and a driver's process starts precisely *because* it is not in a turn. This PR does not require a runtime, so it never meets that rule. If a driver is ever expected to be a process the host starts, that check will need the same widening.

---

Issue #335, opened 2026-07-22
Evaluate a semantic-resolution bridge (LSP/rust-analyzer): unlocks resolved refs, call graph, rename, Rust import resolution


macanderson:
## Summary

**Strategic / roadmap.** The code graph is a tree-sitter index of symbols + import edges — it deliberately omits **semantic resolution** (types, resolved references, resolved `use` paths). That single gap is the shared blocker behind several high-value capabilities. This issue proposes evaluating a **semantic-resolution bridge** — a `rust-analyzer`/LSP adapter (or a deeper resolver) exposed as a port — and names exactly what it unlocks, so the maintainers can weigh it as one decision instead of rediscovering the dependency three times.

## Why it matters

These capabilities are individually valuable for **accuracy**, but each is unsafe or impossible on today's textual/unresolved substrate:

1. **Resolved find-references** — `graph_query references` is a whole-word **textual** scan (`stella-graph/src/frames.rs:63-108`), so it false-hits (comments, strings, unrelated same-name symbols) and false-misses (aliased imports).
2. **A true call graph** (`callers`/`callees` ops) — there are **no call edges** in the index today; `graph_query` has only `definitions/references/imports/importers/neighbors` (`stella-tools/src/graph.rs:65-81`). "Who calls this?" degrades to textual `references`.
3. **Symbol-aware rename** — impossible to do safely via `edit_file` + textual references (false hits/misses corrupt code).
4. **Rust `use` importer resolution** — Rust `use` paths are indexed **unresolved** (`graph.rs:171-178`), so `importers` returns empty for Rust files and reverse-dependency questions fail.
5. **Rust-side test-impact selection** (#334) — blocked on the same resolved edges.

## Evidence

- No LSP / type-checker / rust-analyzer exists anywhere in `stella-graph` / `stella-context` / `stella-tools` (grep found none). The graph is symbols + import edges only (`stella-graph/src/store.rs:42-90`, `queries.rs`).
- `references` is textual (`frames.rs:63-108`); Rust importers unresolved (`graph.rs:171-178`).

## Proposed direction (for discussion — this is a bet, not a foregone conclusion)

- Introduce a resolution **port** (trait) in the ports layer; back it with an adapter that drives `rust-analyzer` (and `tsserver`/`pyright` per language) out-of-process, keeping `stella-core` free of the dependency per the "ports, not concretions" invariant.
[...]

PR #3648, opened 2026-08-18
feat(stella-graph): reconcile the index against HEAD, and embed what changed first

macanderson:
## What

Makes the code-graph index self-healing against git without adding a trigger anyone has to install.

The store now records the commit it last reconciled against, and every mount asks git what HEAD is now. A merge, rebase, pull, checkout, bisect step, or a clone someone handed you all present identically — the recorded sha does not match the current one — and none of them had to notify us.

**Why not a git hook.** A `post-merge` hook in `.git/hooks` is an install step, and an install step is a thing a fresh clone does not have, CI never runs, and a user can forget exactly once. A trigger that can be missing is a trigger whose absence is invisible. This is a **reconciliation against observed state** instead: missing the moment costs latency, never correctness.

**Git is an accelerator, not an oracle.** A diff describes committed history and says nothing about the uncommitted working tree, which is where an agent spends its whole life. So `Plan::Unchanged` does *not* authorize skipping a pass, and `Plan::Scoped` is a superset request rather than a filter. The content hash on `code_graph_files` remains the truth in all four arms, and everything degrades to the full walk the index did before this existed. That reasoning is stated once, in `reconcile.rs`'s module doc.

## The embedding half — the defect this actually fixes

`vectors::pending` and `pending_chunks` ordered strictly `ORDER BY f.path`, and each pass is capped (200 files / 64 chunk files). So the embedded set was always the alphabetically-first pending files: a merge touching `src/z*.rs` in a workspace with a crowded alphabet stayed semantically invisible across an unbounded number of queries, with the `PARTIAL INDEX` note blaming the cap rather than naming the file.

Both now order `(indexed_at DESC, path ASC)`. `indexed_at` moves only when a file's content actually changed (the byte-compat skip guarantees it), so this is "what changed most recently, first" — **derived from state the index already keeps**, not a priority queue that could disagree with it. There is no queue to drain, invalidate, or leak. The `path` tiebreak is load-bearing: `stella init` stamps a whole tree within one second.

Reconciliation runs *before* the catch-up walk in `mount`, so a commit's files are re-indexed first and sort to the front of that queue.

## Architecture

- `RepoOracle` is a port (invariant 1), not a direct git dependency, which keeps `plan()` a pure decision over injected answers (invariant 2's discipline). `GitCli` is the adapter.
- Every oracle method returns `Option`, never `Result`: a repository question this crate cannot answer is not an error, it is a reason to do the safe complete thing.
- `code_graph_index_state` is additive, so it ships without a `SCHEMA_VERSION` bump — the precedent the constant's own doc names (#1214, #335).
- New logic went in a sibling module; `store.rs` is at ~1250 of its 1500 ceiling.

## Evidence

- **Witness (ordering):** `a_capped_pass_embeds_the_most_recently_changed_file_first`. Verified by reverting the `ORDER BY` to `f.path` — it fails with `left: ["src/a.rs"], right: ["src/z.rs"]`, then passes with the fix.
- **Witness (wiring):** `tests/reconcile_wiring.rs` proves the chain end to end against a fake oracle — a moved HEAD indexes the commit's file and *only* that one (an untouched indexable file stays out, so a silent fallback to a whole-tree walk fails the test), a deleted path is pruned, and the commit is recorded so the next pass reads `Unchanged`. A pure-`plan()` test would pass against wiring that does nothing.
- **Real git:** the `-z` framing, `--no-renames` reporting both sides of a rename, and `None`-vs-empty on an unknown commit were reasoning about subprocess output, so they are now run against a real repository rather than argued.
- `make gate` green (exit 0).

## A regression this PR introduced and fixed

The first wiring constructed a temporary `CodeGraph` from a cloned `Arc<Inner>` inside `mount`. `CodeGraph::drop` calls `shutdown()` on the documented assumption that a handle going out of scope is the last public handle — so that temporary latched the shutdown flag and `set_watcher` then refused to arm the live watcher for the entire session. Caught by `session_graph_live_refreshes_after_a_file_is_added`; reconciliation now runs as a free function over `&Inner`, which cannot make that mistake.

## Test renamed (deleted-test guard)

`stepping_over_an_unreadable_file_keeps_the_pass_ordered_by_path` → `..._preserves_the_scan_ordering`. Same assertions; the old name asserted an ordering this PR changes.

## Not in this PR

Filed as issues, listed in a follow-up comment: no push-based background re-embed on the ref change, no single-flight lease, no mid-session reconcile when HEAD moves, and no answer for a retired embedder fingerprint.

## Summary by Sourcery

Make the code-graph index self-healing across Git HEAD changes while ensuring recently modified code reaches embedding passes first.

New Features:
- Reconcile the code-graph index with the repository’s current HEAD during mount and through a public reconciliation API.
- Prioritize recently changed files for file and chunk embedding passes so committed changes are embedded sooner.

Bug Fixes:
- Prevent committed changes from remaining indefinitely unembedded behind alphabetically earlier pending files.
- Ensure deleted and renamed paths are correctly removed or refreshed during reconciliation.

Enhancements:
- Keep Git reconciliation best-effort and use the existing full-tree indexing pass as the correctness backstop for uncommitted changes and unavailable history.
- Expose repository reconciliation through an injectable oracle and persist the last reconciled commit without requiring a schema version bump.
- Preserve live watcher setup by running reconciliation directly against shared graph state rather than constructing a temporary graph handle.

Documentation:
- Update partial-index messaging to describe recency-based embedding coverage rather than alphabetical coverage.

Tests:
- Add unit, integration, and real-Git coverage for reconciliation planning, path handling, wiring, deletion pruning, and recency-ordered embedding.


macanderson:
Remaining work this PR does not cover, filed as handoffs:

- #3649 — a commit's files now sort first in the embed queue, but nothing pulls the queue. Embedding still only happens at `stella init` or lazily inside a search call, so a session that merges and then only greps never embeds the changed files.
- #3650 — five concurrent writers share `codegraph.db` with nothing but `busy_timeout=5000` between them. #3649 adds a sixth, firing exactly when a merge has made the tree large and dirty, so the lease should land first.
- #3651 — a mid-session merge is never git-scoped; reconciliation runs only at mount, so the live watcher absorbs a branch switch as an undifferentiated event storm and `head_sha` goes stale until the next session.
- #3652 — a retired embedder fingerprint leaks its vectors forever and there is no supported way to re-embed short of deleting the store.

Issue #3670, opened 2026-08-18
The benchmark's setup step fails on one task for a reason that was never logged


macanderson:
On a Frontier-Bench run (match `f4fd9a844799`, 2026-08-18), `stella init` exited non-zero on `risk-scorer-replay` while the identical step succeeded on other tasks in the same match. The cause was never found; the run was stopped for unrelated reasons before it could be chased.

## Evidence

`trial.log` for `risk-scorer-replay__Gt8vWac`:

```
Running command: /usr/local/bin/stella init
Command failed
```

Harbor logs `Command failed` without capturing stdout or stderr for this step, so there is nothing else in the artifact to read. The step is best-effort by construction (`_build_code_graph` swallows the exception into `code_graph.unavailable`), so the trial continued and the agent worked normally — 21 tool calls, 19 steps. The consequence is a silently missing code graph for that trial, not a dead run.

For contrast, `sound-change-cascade` in the neighbouring match disclosed a healthy graph from the same code path:

```json
"stella_code_graph": {"state": "reported",
  "summary": "✓ code graph: 6 symbols, 3 imports across 1 file"}
```

So this is per-task, not systematic.

## Candidate causes, none verified

- The 300s timeout in `_build_code_graph` against a large workspace.
- A workspace shape tree-sitter refuses, surfacing as a non-zero exit rather than an empty graph.
- Permissions on `.stella/` under the agent user for that image.

## Reproduce

```sh
# Frontier-Bench requires Harbor >= 0.20.0
cd arenabench && python3 -m arenabench run <a match toml pinning dataset "frontier-bench",
                                            tasks = ["risk-scorer-replay"]>
```
Then read `jobs/*/risk-scorer-replay__*/trial.log`.

## Files

- `bench/harbor_adapter/stella_harbor/__init__.py` — `_build_code_graph`
- `bench/harbor_adapter/stella_harbor/code_graph.py` — `unavailable`, `from_stdout`

## Definition of done

- [x] The cause is identified and either fixed or documented as an accepted limitation with the reason.
- [x] The failure is legible from the artifact: today `Command failed` with no captured output is why this issue exists at all. Capturing init's exit code and stderr into the existing `code_graph.unavailable` payload would make the next occurrence explain itself — note that #3669 is reworking exactly this disclosure and should be done in the same breath or immediately after.
- [x] Every way the step can fail — a raised exception, and a non-zero exit that returns instead of raising — records the exit code and the captured output into the trial's metadata.
[...]

Issue #3669, opened 2026-08-18
bench adapter: trial metadata cannot say whether the semantic index was built — from_stdout drops the line


macanderson:
`stella init` now receives the embedding credential over the anonymous-FD handoff (landed on `worktree-frontier10-harbor-interp`, commit `f54ede184`), so the code graph can be embedded. But a finished trial cannot be asked whether it *was*: `code_graph.from_stdout` keeps only the first line containing the code-graph marker and discards everything else init printed, including the line that reports the semantic-index outcome.

## Why this matters

The semantic-index line is the only per-trial evidence that embeddings ran. Init prints one of:

```
· semantic index: skipped — no embedding backend configured (set VOYAGE_API_KEY, ...)
✓ semantic index: 1 files embedded by voyage-code-3
✓ chunk index: 1 file(s) embedded by voyage-code-3
```

None of it reaches `result.json`. What is recorded is only:

```json
"stella_code_graph": {"state": "reported",
  "summary": "✓ code graph: 0 symbols, 0 imports across 0 files (...)"}
```

Alongside it sits `"stella_embedding_credential_names": ["VOYAGE_API_KEY"]`, which discloses that a credential was *selected*, not that it was *used*. So the metadata reads identically whether the semantic index was built, skipped for want of a backend, or attempted and failed. On 2026-08-18 that cost a real investigation: the absence of a semantic line in trial metadata was read as evidence of embeddings-off, and it is not evidence of anything — the parser structurally cannot emit it. The actual answer had to come from running the SUT by hand in a container.

This is the "evidence must be able to distinguish the claim from its opposite" rule failing at the disclosure layer, and it will silently misreport the very change it is now supposed to measure.

## Files

- `bench/harbor_adapter/stella_harbor/code_graph.py` — `from_stdout`, `_SUMMARY_MARKER`
- `bench/harbor_adapter/stella_harbor/__init__.py` — `_build_code_graph` sets `self._code_graph_summary`
- consumer: whatever writes `stella_code_graph` into trial metadata

## Definition of done

- A finished trial's metadata answers "was the semantic index built?" with three distinguishable states at minimum: built (with backend/model and file count), skipped-no-backend, and attempted-but-failed.
- Keep the existing `no_summary_line` state and its reasoning: "ran and printed nothing" must stay distinct from "reported an empty graph".
- Do not ship raw container stdout into trial metadata — the existing design deliberately avoids that; parse the specific lines.
[...]

Issue #3668, opened 2026-08-18
bench adapter: two tests read ambient provider credentials and fail on any operator's real machine


macanderson:
Two tests in `bench/harbor_adapter/tests/test_adapter.py` read provider credentials from the operator's ambient environment and fail whenever real ones are set. They pass on a clean host and in CI, so the breakage is invisible until someone runs the suite on a machine configured to actually launch a benchmark — which is every bench operator's machine.

## Reproduce

```sh
cd bench/harbor_adapter
export OPENROUTER_API_KEY=... VOYAGE_API_KEY=...      # any real values
PYTHONPATH=. python3 -m pytest tests -q
```

Observed (2026-08-18, on `worktree-frontier10-harbor-interp`):

```
FAILED tests/test_adapter.py::TestRun::test_mounted_internal_stream_is_source_of_truth_and_is_not_overwritten
  - AssertionError: assert b'openrouter-<real key>\n' == b'openrouter-test-...'
FAILED tests/test_adapter.py::TestRun::test_nonzero_exit_persists_stream_context_and_atif_before_raising
  - AssertionError: assert 'OPENROUTER_API_KEY,VOYAGE_API_KEY' == 'OPENROUTER_API_KEY'
2 failed, 714 passed, 2 skipped
```

With the same variables unset: `716 passed, 2 skipped`.

## Cause

Both tests assert on the credential wire the adapter builds, but let the resolver read the real process environment instead of pinning it. The second failure is the clearer one: `optional_embedding_credentials` appends `VOYAGE_API_KEY` to `STELLA_CREDENTIAL_HANDOFF_TARGET` when the host has one, so the expected single-name target string becomes a two-name one. The first leaks an actual live key into an assertion diff, which is its own problem — a failing test should not print a credential.

## Files

- `bench/harbor_adapter/tests/test_adapter.py` — the two `TestRun` tests named above
- `bench/harbor_adapter/stella_harbor/credential_bundle.py` — `optional_embedding_credentials`, `provider_credentials_from_environment`
- `bench/harbor_adapter/tests/conftest.py` — the natural home for an autouse fixture

## Definition of done

- Both tests pass with arbitrary real provider credentials exported, and still pass with none.
[...]

Issue #3533, opened 2026-08-17
bench adapter: credential-absence guard refuses without naming the offending variable, and its name-based scrub misses _AI_KEY


macanderson:
The harbor adapter refuses a Stella trial with a message that does not say which environment variable tripped it. Diagnosing one instance cost a full cycle: a killed 5-task match, a 1-task control to prove it deterministic, and a hand-patched copy of the adapter to print the name.

## What happened

Every Stella trial in a Terminal-Bench match died immediately after `stella init` with:

```
Trial pypi-server__9PpJ4U4 failed: selected provider credential remains in Docker's host environment
```

The message names neither the variable nor the credential, so there is no way to act on it. It is also indistinguishable from an adapter bug, which is where the investigation went first.

The actual cause was operator environment: the host exported the same Voyage secret under two names — `VOYAGE_API_KEY` (recognised, and the one `EMBEDDING_CREDENTIAL_NAMES` forwards) and a duplicate `VOYAGE_AI_KEY`. `_is_credential_env_name` matches `_API_KEY` but not `_AI_KEY`, so the scrub left the duplicate in the Compose host environment and the guard then correctly caught its own scrub failing. `unset VOYAGE_AI_KEY` before launching resolves it with no loss of capability.

The guard is right to refuse — the secret genuinely would have reached Docker's host environment. The defect is that it refuses unactionably.

## Files

- `bench/harbor_adapter/stella_harbor/__init__.py` — the message is raised at two call sites (after health-check/before upload, and again before the anonymous-stdin handoff). Both compute `host_env` and already hold `credentials`, so the offending names are in scope at the raise.
- `bench/harbor_adapter/stella_harbor/__init__.py::_is_credential_env_name` — the name-based filter with the `_AI_KEY` gap.
- `bench/harbor_adapter/stella_harbor/credential_bundle.py::EMBEDDING_CREDENTIAL_NAMES`.

## Repro

```sh
export VOYAGE_API_KEY=<any value>
export VOYAGE_AI_KEY="$VOYAGE_API_KEY"     # same value, unrecognised name
# run any arenabench match with a stella seat
```

Every trial fails at the guard. Deterministic — reproduced at 1 task / concurrency 1, so it is not a concurrency race.

## Definition of done

1. The `RuntimeError` names the offending variable(s) — the names only, never the value. Both call sites. The names are already computable at the raise:
   `sorted(n for n, v in host_env.items() for c in credentials if c and c in v)`.
[...]

PR #3671, opened 2026-08-18
fix(bench): make the head-to-head runnable and honest — kill the nudge, advertise scratch, capture the wire

macanderson:
## What & why

Seven commits that together make a Stella-vs-Claude-Code head-to-head **runnable, honest, and measurable**, plus the two engine defects the measurements exposed. Every claim below is from a captured trace, not an estimate.

### Making the bench run at all

**1. A uv-provisioned Harbor made every Stella seat unlaunchable** (`bba3079d9`) — `harbor_interpreter` read `uv`'s `#!/bin/sh` re-exec wrapper as the interpreter, so the seat probe ran `/bin/sh -c "import …"` → `import: command not found`. Frontier-Bench's `min_harbor` forces that Harbor, so no Stella arm could start on that dataset while Claude Code's launched beside it and the match ran one-sided.

**2. The credential-absence refusal named no variable** (`414cc87de`) — Refs #3533.

**3. `stella init` never got the embedding credential** (`f54ede184`) — init is where the semantic index is *built*, so a credential arriving later on the run path finds no vectors to rank. Measured both ways in a container: `semantic index: skipped` without it, `1 files embedded by voyage-code-3` with it.

### Engine defects the measurements exposed

**4. The search nudge is deleted, not narrowed** (`d4a977f40`) — measured over a 20-task panel it fired on **44 of 415 tool results across 10 of 20 tasks and produced zero `search` calls**, on hardware probes, package listings and a grep of a C header. Advice injected into tool *output* is re-sent as input on every later turn, so a wrong nudge is a compounding context tax and the model cannot tell engine prose from the command's own bytes. The whole predicate chain goes; the `cd`-escape warning stays, because it reports a fact about the command's effect rather than an opinion about which tool to call. Confirmed dead in the live run: **0 occurrences**.

**5. The scratch directory is advertised where the model looks** (`d4a977f40`, tightened in `330e16d8c`) — it already existed and was exported as `$STELLA_SCRATCH`, but the only place its existence was stated was the confinement refusal, so an agent learned of it by failing and then wrote build artifacts into the *graded workspace*. The first fix produced recovery instead of pollution but cost two turns (the agent spent a whole turn on `get_environment` to read a path already in its shell); the follow-up gives a usable fragment — `apt-get update > $STELLA_SCRATCH/apt.log` — in both the schema and the refusal.

### Measurement instruments

**6. Capture what a run asks for, not only what it cost** (`e62dc938a`) — `step_usage` records tokens and price; nothing recorded the *request*, where the reasoning budget, output ceiling and tool schemas live. Stella captures in-process (`STELLA_WIRE_LOG`, bodies only, never headers); Claude Code goes through `bench/wirelog/wirelog.py`, because nothing inside CC can be instrumented. A proxy was tried for Stella first and **correctly refused** — an OpenRouter benchmark run must reach the canonical endpoint — so that guard stands untouched.

**7. Server-side context editing** (`48ce94e6d`) — `context_management.edits`, built as a **policy with two floors** rather than a switch, because clearing invalidates the cached prefix and a cache write is ~12.5x a cache read: at Stella's measured 91.7% hit rate, naive clearing costs more than it saves.

## What the run showed

Match `13f7f2bb533d`, 14 search-relevant easy/medium TB2.1 tasks, Sonnet 5 at low effort, both arms wire-captured. Four trials excluded by name: two Docker Hub TLS timeouts (agent never ran) and **two contaminated by my own capture proxy** — three stalls, two of them a 200 with zero SSE events, were 79% of Claude Code's model time and landed on exactly the two trials it lost. Clean basis, 9 tasks:

| | Stella | Claude Code | ratio |
|---|---|---|---|
| passed | 7/9 | 5/9 | — |
| billed | $1.18 | $2.17 | 0.54x |
| **priced through one table** | **$1.64** | **$2.13** | **0.77x** |
| input tokens | 1,779,320 | 4,098,384 | 0.43x |
| output tokens | 43,298 | 28,796 | 1.50x |
| wall clock | 71.8 min | 55.0 min | 1.30x |

The cost result **reversed** from the previous panel, where Stella cost 26% more on identical pricing ($5.84 vs $4.65); it now costs 23% less. Two known errors run *against* that conclusion, making it a floor: Claude Code's `rstan-to-pystan` cost recorded as $0.000, and Claude Code paid a proxy hop Stella did not.

Accuracy is **not** claimed: 7/9 vs 5/9 is a two-task margin on nine tasks and resolves nothing.

## The witness

| Test | On `main` | Here |
|---|---|---|
| `test_a_uv_shell_wrapper_resolves_to_the_python_beside_it` | `/bin/sh` | the venv's python |
| `tests/test_credential_leak_message.py` | helpers absent | names `VOYAGE_AI_KEY`, never the secret |
| `tests/test_init_embedding_credential.py` (5) | helper absent | wire shape, argv guard, newline guard, the wiring, and the no-backend fallback |
| `a_bash_result_never_carries_tool_preference_advice` | advice appended | silent on all ten commands, four taken from measured misfires |
| `bash_advertises_the_scratch_directory_only_when_it_exists` | no mention | names the variable, never the session path |
| `a_session_that_did_not_opt_in_sends_no_context_management` | n/a | no field **and** no beta header |
| `opting_in_sends_both_strategies_and_the_beta_header` | n/a | every field vs the documented contract |

## The gate

- [x] `cargo test`: stella-model 431 pass; stella-tools 403 pass / **1 fail**; harbor_adapter 721 pass; arenabench 18 pass
- [x] `cargo clippy` clean on every crate touched
- [x] `docs/tools/bash.toml` regenerated (`make tool-docs-update`); tool-docs green
- [ ] **The one failure, stated not hidden:** `does_not_block_on_a_backgrounded_child_holding_the_pipe_open`, a timing assertion that measured 31.8s while a 14-task benchmark held this machine at **load average 214**. This PR edits one schema string and one error string, neither on that code path. It needs re-running idle before the suite is called green — I have not done that.

## Nothing left behind

- [x] Filed: #3668, #3669, #3670, #3752, #3753, #3754, #3756, #3757

**#3752** — `search` called **zero** times across both panels, including one selected so every task ships ≥3 files. The retrieval stack and the new embedding work are currently unused. **#3753** — 75% of Stella's wall clock is tool execution; `pytorch-model-cli` timed out at 900s having spent $0.034, with 984s of tool time. **#3754** — my own capture proxy stalls and silently contaminated two paid trials; it must fail loudly. **#3756** — context editing is inert: its trigger is unmeasured and OpenRouter may drop the field. **#3757** — Stella asks `max_tokens: 128000` against Claude Code's 64000 for a measured 299 output tokens/call, on the one route where a gateway prices the credit check against the ceiling. **#3668/#3669/#3670** as before.

#3533 is **advanced, not finished** — its DoD item 2 (decide whether `_is_credential_env_name` should widen to `_AI_KEY`) is untouched on purpose: widening changes what gets scrubbed, which is a maintainer's decision, not a quiet edit inside a bug fix.

Refs #3533

## Ground-rule check

- [x] No new outbound network calls (`wirelog` is an opt-in local bench tool, run by hand)

**Review commit 7 hardest.** `ContextManagement::new` is the only constructor so the strategy ordering the API rejects cannot be got wrong, and the two floors (`trigger`, `clear_at_least`) are what stop the feature costing money on short conversations. The load-bearing test is the *off* one.

**A correction I made twice, recorded because the issues cite it.** I first reported that Claude Code clears thinking blocks and Stella doesn't. It does not: it sends `keep: "all"`, which clears nothing and is the cache-optimal setting. I also reported the arms were unmatched on reasoning; on Sonnet 5 adaptive is the only on-mode and both arms send `effort: low`. Both claims came from inferring semantics instead of reading the spec.

**Measurement method worth reusing:** price both arms through one table as well as reporting what each was billed. The billed numbers reversed the answer once already — OpenRouter charged 0.67x Anthropic list — and a raw comparison of two self-reported figures on different bases is not a cost comparison.

## Summary by Sourcery

Make Stella head-to-head benchmarks runnable, honest, and measurable by fixing launch and initialization failures, removing misleading tool nudges, and capturing the requests and context policies used by each run.

New Features:
- Add opt-in capture of Stella outbound request bodies and a local proxy for recording both benchmark arms’ wire exchanges.
- Add configurable Anthropic server-side context editing with cache-conscious clearing policies.
- Provide scratch-directory guidance in the Bash tool when scratch storage is available.

Bug Fixes:
- Fix Harbor interpreter detection for uv shell wrappers so Stella seats launch with the adjacent Python interpreter.
- Pass embedding credentials securely during stella init so semantic indexes can be built when a backend is configured.
- Improve credential-leak refusals by identifying the environment variables carrying credentials without exposing secret values.
- Remove misleading search-tool advice from Bash results while retaining workspace-boundary warnings.

Enhancements:
- Make benchmark traces more complete and comparable by recording request parameters alongside usage and cost data.

Documentation:
- Regenerate the Bash tool documentation to describe scratch storage guidance.

Tests:
- Add witness coverage for uv interpreter resolution, credential handling, embedding initialization, Bash output behavior, scratch advertisement, wire-level context-management opt-in, and request serialization.

## Tests deleted, named

Two tests are gone from `crates/stella-tools/src/bash.rs`, both removed in
`d4a977f40` together with the feature they covered — the search nudge:

- `bash_grep_symbol_detection` — covered the symbol-shaped-grep predicate, which no longer exists.
- `a_symbol_shaped_bash_grep_is_pointed_at_search_without_an_index` — asserted the nudge fired on such a grep; its replacement is `a_bash_result_never_carries_tool_preference_advice`, which asserts the opposite over ten commands, four of them measured misfires.

Nothing else disappears in the merge.


macanderson:
Rebased this onto current main (which now carries the h2 0.4.16 advisory fix, the brand recolour, and #3663/#3664). `cargo deny` and the pytest job go green; `fmt + clippy + test` still fails on two guards, and both look like they need your call rather than a mechanical fix:

**1. `check-file-size` — a grandfathered god file grew**

```
bench/harbor_adapter/stella_harbor/__init__.py grew to 2026 lines,
over its baseline ceiling of 1864 (+162)
```

`make file-size-update` would turn this green, but AGENTS.md closes the grandfathered files to growth and CLAUDE.md calls a raised ceiling-to-go-green an expedient, so I did not do that. The remedy the guard names is putting the new code in its own module — which is a design decision on your feature, not something to do behind you.

**2. `check-deleted-tests` — two tests vanish unacknowledged**

```
a_symbol_shaped_bash_grep_is_pointed_at_search_without_an_index
bash_grep_symbol_detection
```

Both exist on main in `crates/stella-tools/src/bash.rs`. This looks **deliberate**, not the #1976 accidental-revert shape: the branch's `bash.rs` still has `drift_advisory`, and where main asserts a symbol-shaped `grep` gets pointed at `search`, the branch has a replacement asserting the inverse —

```
a_symbol_shaped_bash_grep_comes_back_unannotated   (branch, line 1114)
```

So the advisory appears to have been deliberately stopped from annotating. If that is intended, the guard is satisfied by naming both deleted tests in the PR description and saying why. Worth flagging that this behaviour change is not described in the PR body at all — the description covers the three Frontier-Bench defects (uv shebang, credential-absence message, `stella init` embedding credential) and none of them is the bash grep→search advisory. If that change rode along from an older base, it is worth a second look before it lands.

I have not pushed anything to this branch.

Issue #3695, opened 2026-08-18
Wire the wrapper-plugin driver through stella goal rounds and stella fleet workers


macanderson:
## Problem

`stella run` is the only door that can run a turn under an installed wrapper plugin — `crate::wrapper_plugin::bind_installed`/`WrapperDispatch` implement `TurnDriver` over exactly one raw engine turn (`crates/stella-cli/src/agent/goal.rs::run_raw_one_shot`, the `Plugin` arm). `stella goal` (judged rounds) and `stella fleet` (per-task workers) each drive their own outer loop around a turn — a round loop and a wave-scheduled worker loop respectively — and neither has any path to a `WrapperDispatch`.

As part of #3381 slice 2 (flipping the pipeline default on goal/fleet/deck), both doors gained a real `--pipeline <VARIANT>` flag with the same value semantics as `stella run`'s (`classic` names the built-in staged pipeline; an unnamed flag is the raw loop). Naming an *installed plugin's* variant (anything other than `classic`) is refused outright rather than silently downgraded or accepted-and-ignored:

```rust
// crates/stella-cli/src/wrapper_plugin.rs
pub(crate) fn reject_plugin_variant_for_door(door: &str, choice: PipelineChoice<'_>) -> Result<(), String>
```

called from `main.rs`'s `Goal` and `Fleet` arms before either door's run function is invoked. The refusal message names `stella run --pipeline <variant>` as the door that *can* run it. This issue is that refusal's tracking ticket — the thing to build so the refusal can eventually go away.

## What to build

1. Give `stella goal`'s round loop (`agent/goal.rs::run_goal_cmd` → `run_goal_pipeline_turn`/`run_goal_turn`) a `Plugin(variant)` arm: bind the wrapper once per goal run (not once per round — matching `--pipeline`'s once-per-process cost on `stella run`), and decide how the wrapper's before/after/judge/again sequence composes with the goal loop's own verifier assessment. This needs a design decision the run-door precedent doesn't answer: does the wrapper's ladder replace, gate, or run alongside the goal verifier's "is the goal met" judgment?
2. Give `stella fleet`'s per-task worker (`fleet_cmd.rs::run_task`) a `Plugin(variant)` arm, binding the wrapper once per worker (once per task attempt, in the task's own workspace/workspace_root — matching the wrapper's tamper-watch grant, which is scoped to a shared work tree per `crate::wrapper_candidate`).
[...]

Issue #3684, opened 2026-08-18
Flip the pipeline default on goal, fleet, and deck doors (#3381 slice 2 — run/arena landed)


macanderson:
## Problem

#3381 slice 1 flipped `PipelineChoice::resolve`'s no-flag default from `Classic` to `Raw` and made `--no-pipeline` a deprecated no-op, but **only on the two doors that already route through `PipelineChoice`: `stella run` and `stella arena`.** Three doors still default to the staged pipeline and were explicitly out of scope for slice 1 per its own instructions ("Do NOT touch goal/fleet/deck in this slice"):

- **`stella goal`** — `crates/stella-cli/src/main.rs` calls `agent::run_goal_cmd(&cfg, &goal, spend_limit, !no_pipeline)`; the bool is inverted at the call site and threaded as `use_pipeline: bool` (default `true`, i.e. pipeline-on-by-default). `crates/stella-cli/src/agent/goal.rs` branches `if use_pipeline { run_goal_pipeline_turn(...) }` around lines 369-376, 419-428, 465-466, and there is no `--pipeline <variant>` flag on this door at all — only the bare `no_pipeline: bool`.
- **`stella fleet`** — `crates/stella-cli/src/main.rs` calls `fleet_cmd::run_fleet(..., !no_pipeline, ...)`; `run_fleet` takes `use_pipeline: bool` and threads it into `EngineWorker`, whose per-task turn branches `if use_pipeline` (see `crates/stella-cli/src/fleet_cmd.rs` around lines 80, 166, 469-479, 508, 547, 667-674, 754-755, 834-836) into a full `stella_pipeline::PipelineConfig`/`PipelinePorts` run vs. a raw turn. Also no `--pipeline <variant>` flag.
- **The deck (`stella` interactive / Command Deck)** — has no CLI flag at all. `crates/stella-cli/src/command_deck.rs` hardcodes `pipeline_persona = resume_state.and_then(|rs| rs.pipeline).unwrap_or(true)` (lines ~437-441, 686-703, 833-838, 1234): the `/pipeline` toggle starts ON. A resumed session restores `pipeline_on = rs.pipeline.unwrap_or(true)`, same default.

Per `docs/spec/turn-loop-wrappers.md` §5 "Flip the default" (#3381), *every* door should hit the raw loop by default, with an explicit opt-in. Slice 1 only delivered that for two of six doors.

## What to build

[...]

Issue #3682, opened 2026-08-18
pipeline: decide_candidate_verify always reports witness_authored=false, so a verify condition on that signal can't be honestly answered


macanderson:
## Problem

#3674 (closed) wired `Pipeline::run_candidate` to decide `verify` per candidate,
post-execute, against that candidate's real `mutating_actions`/`diff_lines`
(`Pipeline::decide_candidate_verify`, `crates/stella-pipeline/src/pipeline/schedule_wiring.rs`).
That closes the case #3674 was filed for (`verify if = "diff-lines > 0"`).

It does **not** close every post-execute signal `verify` could legally read.
`Signal::WitnessAuthored` (published by the `witness` stage) is also a signal
`verify` — declared after `witness` — is entitled to condition on, e.g.
`verify if = "witness-authored"`. `decide_candidate_verify` currently hands
this a hardcoded `false`:

```rust
schedule.update(|v| {
    v.mutating_actions = u64::from(mutating_actions);
    v.diff_lines = u64::from(diff_lines);
    v.witness_authored = false;
});
```

with a comment explaining why: by the time `run_candidate` calls
`decide_candidate_verify` (right after `gather_diff`, to compute
`should_verify` and decide whether to bail out via `state.into_unverified()`),
the actual witness-authoring attempt (`witness_on_demand`) has not happened
yet — it only runs *after* the `should_verify` gate passes. So the real fact
does not exist at the point the decision needs it.

A naive fix — reorder `run_candidate` to attempt authoring before deciding
`should_verify` — was considered and deliberately not taken this round: the
`should_verify` gate today is also what decides whether to compute the
`warrant()` and emit `ProofStep::Warrant` at all, and
`crates/stella-pipeline/src/pipeline/tests.rs`'s exact-stage-sequence
assertions (`clean_lookup_skips_plan_verify_and_verifier`, and siblings) pin
today's ordering — that file is at its line-ceiling and frozen for this kind
of change, so verifying a reorder doesn't regress those sequences needs its
own dedicated pass.

## Impact today

None that's reachable: `classic.toml`'s only shipped `verify` condition
(`if = "verifies"`) reads a triage-published signal, and no test or shipped
variant conditions `verify` on `witness-authored`. A hand-authored variant
that did so today would silently always take the `false` branch of that
condition — wrong, but currently unreachable in practice
(`PipelineConfig::variant` has no CLI flag wiring a caller can set it from
yet, same caveat #3674 itself carried).

## What "done" looks like

[...]

PR #3767, opened 2026-08-18
fix(stella-cli): refuse pipeline-only verification flags on the raw default; preserve legacy deck personas on resume

### LINKED ISSUE #3696: --test-command/--keep-witness/--require-verified silently no-op on the raw loop, which #3381 made the default
## Problem

`stella run`'s `--test-command`, `--keep-witness`, and `--require-verified` flags all belong to the staged pipeline's verify/witness machinery. Before #3381, a plain `stella run "…" --test-command "…"` ran the staged pipeline by default, so the flag reached `run_pipeline_one_shot` and armed the deterministic verify ladder. After #3381 flipped the no-flag default to `PipelineChoice::Raw`, the same plain invocation now silently does nothing with any of the three flags:

- `run_one_shot` (`crates/stella-cli/src/agent.rs:137-149`) branches `if pipeline.is_classic() { run_pipeline_one_shot(..., test_command, keep_witness, require_verified) } else { run_raw_one_shot(cfg, prompt, budget_limit, format, pipeline, test_command) }` — the `Raw`/`Plugin` arm's `run_raw_one_shot` **does not even accept `keep_witness`/`require_verified` as parameters**; they are dropped at the call site, not merely unused inside the callee.
- `test_command` *is* threaded into `run_raw_one_shot`, but per that function's own doc comment it "arms no ladder of this crate's own — there is none" on the bare `Raw` arm (no `--pipeline <variant>`); it only does something when `pipeline.plugin()` is `Some` (an installed wrapper's `[oracle]`).
- Net effect: `stella run --test-command "cargo test -p x" "fix the bug"` with no `--pipeline` flag — now the single most common invocation shape — silently ignores `--test-command` entirely. No warning, no notice, no non-zero exit. The flag parses, is accepted, and does nothing.

`stella arena` inherits the same gap one level worse: it has no `--pipeline <VARIANT>` flag at all (tracked separately in #3554/#3695), so its `--test-command` is now **permanently** inert — there is no flag combination on that door that makes it do anything.

This directly contradicts CLAUDE.md's "an expedient is a defect" framing applied to CLI ergonomics: a flag that parses successfully and silently changes nothing is worse than a rejected flag, because nothing tells the caller their `--test-command`/`--keep-witness` had no effect. Verified by direct reading — `crates/stella-cli/src/agent.rs:137-149` and `crates/stella-cli/src/agent/goal.rs`'s `run_raw_one_shot` signature (search for `pub(crate) async fn run_raw_one_shot`) — not inferred from documentation.

## What to build

Pick one (a maintainer call, not mine to make silently):

1. **Warn.** When `test_command`/`keep_witness`/`require_verified` is set and `pipeline` resolves to `Raw` (no `--pipeline classic` and no plugin), print a one-line notice at the same place `no_pipeline_deprecation_notice` already prints its own ("`--test-command` has no effect without `--pipeline classic` or an installed wrapper") — matching the existing `crate::wrapper_plugin::no_pipeline_deprecation_notice` shape/pattern (pure fn returning `Option<&'static str>`, printed once at the door).
2. **Refuse.** Treat `--test-command`/`--keep-witness`/`--require-verified` together with an implicit-or-explicit `Raw` choice as a usage error (clap `conflicts_with`-style, or a runtime check at the door) — the same shape `reject_plugin_variant_for_door` uses for a different combination.
3. **Extend.** Give the bare raw loop (no wrapper at all) its own minimal verify-ladder arm so these flags mean something without opting into the full staged pipeline — likely the largest option and possibly out of scope for a flag-accuracy fix.

Whichever is chosen, update the four doc pages that show these flags (`website/content/docs/commands/{run,arena}.mdx` already got a `--pipeline classic` pairing note in #3381 slice 2; `goal.mdx`/`fleet.mdx` don't expose these three flags at all today) to match.

- `run_raw_one_shot`'s signature does not currently accept `keep_witness`/`require_verified` at all — adding real behavior for them on the `Plugin` arm (a wrapper's own witness/verification story) is a separate, larger question already scoped out in `crates/stella-cli/src/agent/goal.rs`'s doc comments; this issue is about the **bare `Raw` arm's silence**, not about giving wrappers a witness ladder.
- `crate::wrapper_plugin::no_pipeline_deprecation_notice` is the established pattern for "a flag needs a heads-up printed once at the door" in this exact area of the code — reuse its shape rather than inventing a new one.
- `stella arena`'s half of this (permanently inert `--test-command`, no `--pipeline` flag at all) is arguably its own issue since the fix isn't symmetric with `run`'s (arena would need warn-only, since it can never gain `--pipeline` without also gaining the flag itself per its adapter contract) — filing here first since the underlying defect (silent no-op) is one shape; split if the fix diverges.

## Definition of done

- A caller who passes `--test-command`/`--keep-witness`/`--require-verified` without a driver that can use them either sees a clear notice, or is refused with an actionable message — never silent.
- The four command reference pages accurately state which flag combinations do something.
- A witness test proving the chosen behavior (notice text present in stderr, or a non-zero exit / parse error, depending on which option is chosen).

Refs #3381 (the flip that surfaced this), #3684 (goal/fleet default flip, adjacent), #3554/#3695 (the `--pipeline`-scoped-to-`run` gap that makes arena's case unfixable the same way `run`'s is).

macanderson:
## What & why

Five findings from the #3694 flip's post-merge adversarial audit (3 P1, 2 P2), all fixed and re-audited:

1. **P1** — `stella run --keep-witness` / `--require-verified` / `--test-command` were silently dropped on the now-default raw loop (also flagged by the Vercel review bot; #3696). Now refused before dispatch via `wrapper_plugin::reject_verification_flags_without_pipeline`, naming `--pipeline classic` as the remedy: `Raw` refuses all three, `Plugin(variant)` accepts `--test-command` and refuses the other two, `Classic` accepts all. **Arena included**: `1332dd1` (thanks @macanderson) gives `stella arena` the same `--pipeline <VARIANT>` flag and wires the identical refusal, so `--test-command` there is honored or refused, never inert — `arena.mdx` updated to match.
2. **P1** — a pre-flip deck session that never touched `/pipeline` was indistinguishable from an always-raw one and silently swapped driver + system-prompt persona on resume (invariant 7). `session_persist::derive_pipeline_persona`/`initial_pipeline_persona`: an explicit journal record wins; history-without-record restores ON (the default that session ran under); fresh sessions start raw. Session start now journals the resolved value so no future flip faces the ambiguity.
3. **P1** — the comment stating the `authorize_one_shot(!pipeline.is_raw())` surface widening now exists at the call site (`agent.rs:128`).
4. **P2** — the goal/fleet plugin-variant refusal now cites its real tracking issue #3695 (was #3684).
5. **P2** — the `--no-pipeline` deprecation notice printed twice under the supervised terminal posture; now printed exactly once, after the supervision decision.

Refs #3381
Closes #3696

## The witness

- [x] `crates/stella-cli/tests/verification_flags_require_pipeline_cli.rs` (6 tests, end-to-end through the real binary): each verification flag alone under the raw default is refused; accepted with `--pipeline classic`; arena both ways (`arena_test_command_is_refused_without_pipeline`, `arena_test_command_is_accepted_with_pipeline_classic`). `session_persist::tests`: legacy-history-restores-ON / fresh-starts-OFF / explicit-record-wins. All fail on the pre-fix tree.

## The gate

- [x] `cargo fmt --check` (clean)
- [x] `cargo clippy -p stella-cli -p stella-store -p stella-tui --all-targets -- -D warnings` (clean)
- [x] `cargo test`: stella-cli 1843/1843, tui 889 (+1 pre-existing ignored), store 357+31 — 0 failed; full workspace runs in CI on this head
- [x] Docs updated where flags changed (`run.mdx`, `arena.mdx`, clap help)
- [x] `Closes #3696` above and as a commit trailer

## Nothing left behind

- [x] Tracked: #3695 (wrapper driver through goal/fleet), #3682 (witness_authored placeholder), #3684 (flip umbrella). Nothing new left untracked by this round.

## Ground-rule check

- One flaky unrelated test observed once in a full `-p stella-cli` run (`daemon::tests::a_run_that_ignores_term_is_killed_after_the_grace_period`, a SIGTERM grace-timing test); it passes alone and on re-run.
- Draft status is the owner's; the branch is complete and validated from this session's side.

## Summary by Sourcery

Prevent verification options from being silently ignored and preserve compatible pipeline personas when resuming legacy sessions.

New Features:
- Add pipeline selection to `stella arena`, enabling raw, classic, and installed wrapper-plugin benchmark runs.

Bug Fixes:
- Reject verification flags when the selected driver cannot honor them instead of silently ignoring them.
- Restore the staged-pipeline persona for legacy deck sessions on resume and persist the resolved persona for future resumes.
- Print the `--no-pipeline` deprecation notice only once during supervised runs.

Enhancements:
- Clarify that the raw loop is the default and the staged pipeline is opt-in across CLI behavior and user-facing documentation.
- Correct the tracking reference for unsupported plugin variants on goal and fleet commands.

Documentation:
- Update CLI, arena, README, design, and guidance documentation to reflect opt-in pipeline behavior and verification-flag restrictions.

Tests:
- Add unit and end-to-end coverage for verification-flag refusals, arena pipeline selection, and legacy-session persona restoration.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -426,14 +426,21 @@ template. If a witness is genuinely impractical (e.g. TUI rendering), explain
 how you verified the change instead.
 
 **What follows is the staged pipeline's own witness/verify machinery, which is
-a different thing from the contract above.** It ships in-tree today and
-genuinely runs the check itself — this is the host-run half of the guarantee
-(see AGENTS.md's opening). Per the product decision recorded in
-`doc:pipeline-as-plugins` (#3511) it is being extracted into an installable
-verification plugin (Oxagen's Vera is the reference one); once it lands, a
-plugin reports its own evidence instead of Stella running the check, and this
-section's guarantee does not automatically transfer to that path. Until then,
-this section documents the mechanism as it exists: when no
+a different thing from the contract above.** It is **opt-in**, not the
+default: since #3381 a plain `stella run` resolves to the raw step-loop, and
+this machinery runs only when a turn is explicitly wrapped with `stella run
+--pipeline classic` (`crates/stella-cli/src/wrapper_plugin.rs` —
+`PipelineChoice::Classic`). It ships in-tree today and genuinely runs the
+check itself when invoked — this is the host-run half of the guarantee (see
+AGENTS.md's opening) — and `--test-command`/`--keep-witness`/
+`--require-verified` are refused on any other resolution (the raw loop, or an
+installed wrapper plugin), naming `--pipeline classic` as the remedy, rather
+than silently doing nothing. Per the product decision recorded in
+`doc:pipeline-as-plugins` (#3511) the mechanism is being extracted into an
+installable verification plugin (Oxagen's Vera is the reference one); once it
+lands, a plugin reports its own evidence instead of Stella running the check,
+and this section's guarantee does not automatically transfer to that path.
+Until then, this section documents the mechanism as it exists: when no
 `--test-command` is configured, its **witness stage** has an independent model
 (the verifier's resolution, never the worker) author the failing witness test,
 tracks its fail→pass flip in the flip oracle, and refuses to credit the flip if
@@ -443,7 +450,12 @@ executed diff and found something worth proving — so the stage order is
 triage → recall → research → plan → scope → **execute → witness** → verify → verdict
 (`stage_rank` in `crates/stella-pipeline/src/replay.rs` is the canonical
 ordering; the revise back-edges land on execute, so re-execution never
-re-authors). The witness
+re-authors). That order is not hard-coded: it is resolved at load time from
+the shipped `[wrapper]` manifest (`crates/stella-pipeline/variants/classic.toml`,
+via `src/schedule.rs` and `src/pipeline/schedule_wiring.rs`) — a witness
+condition written to read a post-triage (post-execute) signal is refused at
+load rather than silently accepted, so a manifest cannot describe an ordering
+the schedule cannot honor. The witness
 is **scaffolding for that one run**: it lives in the candidate workspace and is
 discarded with it, so an already-satisfied test is never left behind in the
 project's test tree. `stella run --keep-witness` promotes it instead.
@@ -644,7 +656,8 @@ alone, because a number in two places is how the last limit died.
 **Status — what ships.** The live runtime path is
 `stella-cli` → `stella-core` → `stella-model` / `stella-tools` / `stella-store` /
 `stella-context` (recall only) / `stella-mcp`, and the CLI also drives
-`stella-pipeline` (the default `stella run` path), `stella-fleet` (`stella fleet`),
+`stella-pipeline` (opt-in on every door via `stella run --pipeline classic`;
+the raw step-loop is the default), `stella-fleet` (`stella fleet`),
 and `stella-tui` (the Command Deck, the default interactive shell on a TTY).
 The fuller
 `stella-graph` retrieval + context plane (`stella init` builds the code-graph
--- a/README.md
+++ b/README.md
@@ -38,10 +38,11 @@ in Rust as a workspace of focused crates.
 
 - **BYOK, auto-detected** — Set one provider's API key and Stella detects it.
   Pin a specific model per run or shell with `--model`.
-- **Deterministic definition of done** — the staged pipeline's witness stage
-  has an independent model author a test that fails on the old code and passes
-  on the new, and tracks that fail→pass flip, host-run out of the box. A green
-  suite alone is not accepted. This machinery is also becoming Vera, an
+- **Deterministic definition of done, opt-in** — `stella run --pipeline
+  classic` runs the built-in staged pipeline, whose witness stage has an
+  independent model author a test that fails on the old code and passes on
+  the new, and tracks that fail→pass flip, host-run. A green suite alone is
+  not accepted on that path. This machinery is also becoming Vera, an
   installable verification plugin — a plugin's oracle reports its own
   evidence instead of Stella re-running the check.
 - **Single-threaded engine** — One deterministic step loop: plan, fan tools out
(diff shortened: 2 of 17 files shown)


macanderson:
Thanks — the `stella arena` finding was correct and is now fixed; the docs finding was correct for `arena.mdx` and stale for the other three. Taking the three assessment rows in turn.

**1. `stella arena` was not wired to the refusal — confirmed, fixed (1332dd1).**

It resolved `PipelineChoice::resolve(args.no_pipeline, None)` — always `Raw` since #3381 — and handed `--test-command` straight to a driver with no verify ladder to arm. So a benchmark runner asking for an oracle got a number measured without one, with nothing on stderr saying so. That is the worst place in the tree for this bug: a bench number that flatters us because of a measurement artifact spends the reputation the project exists to build.

Refusing alone would have been the cheap fix and the wrong one, because arena had no `--pipeline` flag to name as the remedy — the error would have pointed at a door that did not exist. So arena now takes the same `--pipeline <VARIANT>` flag `stella run` does, and `main.rs`'s `Command::Arena` arm calls `reject_verification_flags_without_pipeline` before dispatch, exactly as the `Run` arm does. `--test-command` is now either honored (`--pipeline classic`) or refused before the episode starts, never inert.

**2. CLI documentation — `arena.mdx` was stale, and is updated.**

Its `--test-command` card said "Currently inert", and the page's warning callout said there is "currently no way to make an arena episode run the staged pipeline at all". Both were true of that tree and are now false, so both are rewritten, along with the synopsis and the See-also line.

`run.mdx` already documents the pairing (its "pair them with `--pipeline classic`" note and the two `--pipeline classic` examples). `goal.mdx` and `fleet.mdx` document the plugin-variant refusal, and neither door accepts `--test-command`, `--keep-witness`, or `--require-verified` at all — there is no combination on those pages to correct.

**3. Witness tests — extended to the new door.**

`crates/stella-cli/tests/verification_flags_require_pipeline_cli.rs` gains two, both driving the real binary end to end:

- `arena_test_command_is_refused_without_pipeline` — fails on the pre-fix tree twice over (the flag was accepted and inert, and `--pipeline` did not parse), passes here.
- `arena_test_command_is_accepted_with_pipeline_classic` — the companion half, proving the refusal is scoped to a resolution that cannot honor the flag rather than to `--pipeline` being absent.

Verified: 6/6 in that file pass; `stella-cli` clippy `-D warnings` clean; `check-command-docs` and `check-left-behind` green. The full `-p stella-cli` suite is 1839/1840 — the one failure is `daemon::tests::a_run_that_ignores_term_is_killed_after_the_grace_period`, a SIGTERM grace-period timing test that passes when run alone and is unrelated to this change.

Issue #3808, opened 2026-08-19
a cancelled execution's reflection is a silent stub — the runs most worth learning from lose their record


macanderson:
## Problem

`record_execution_end` (`crates/stella-cli/src/agent/persistence.rs:379`) sets
`terminal_usage_known = outcome_label != "cancelled"`, which is correct and must
stay: a cancel can race a provider response after dispatch, so the usage envelope
is genuinely unknowable and the execution must never become exportable.

But the same function then calls `finalize_execution_reflection` and
`sync_to_usage_default` on every path including cancel, and the reflection that
lands for a cancelled run is a stub: empty critique, `self_rating = null`. In the
audited session (`ses-1787092335250-47513`, execution 157) the run was cancelled
about three minutes after `pnpm build` passed, having made 66 successful edits
across 35 files. The reflection recorded none of that judgement.

So cancelled sessions — the ones most worth learning from, because something made
a human intervene — lose their learning record while keeping their telemetry.

## What is NOT the bug

`wrote_files = 0` in that same reflection row had a **different** cause, already
fixed on `worktree-session-eval-fixes`: `files_touched` had no producer on the
engine path at all, so the flag read zero for every direct-edit turn, cancelled or
not. Do not re-diagnose that here.

## The question to settle

Whether a cancelled execution should get a real reflection at all, and if so what
it may claim. The honest answer is probably "yes, marked as covering a partial
run" — the edits happened and their quality is assessable — but that is a design
call, not an obvious repair, and it touches the reflection loop's contract about
what a rating is a rating *of*.

Whatever is decided, `terminal_usage_known` must not change: it is about the
provider's usage envelope, not about the reflection.

## Files

- `crates/stella-cli/src/agent/persistence.rs:379` — `record_execution_end`
- `crates/stella-store/src/reflection.rs:202` — `finalize_execution_reflection`
- `crates/stella-cli/src/command_deck/session_clear.rs` — the deck's cancel path
- `crates/stella-store/src/tests/usage_completeness.rs` — the existing contract

## Verify

Witness test: finish an execution with outcome `cancelled` after recording real
work, and assert the reflection row reflects that work under whatever contract is
chosen — and, separately, that `usage_complete` stays false.

## Done

A cancelled run's reflection either carries a real, correctly-scoped judgement or
is explicitly and visibly absent, with the choice argued in the code. Silent stub
rows are the one outcome this issue exists to remove.

Found in the same session audit as #3690 and #3807.

[...]

Issue #3807, opened 2026-08-19
task sub-agent delegations record no agent_uses rows — the tool cannot reach the ledger


macanderson:
## Problem

`ToolRegistry::record_agent_use` has exactly one production caller in the
workspace — `crates/stella-cli/src/command_deck.rs:3572`, the deck's named-agent
path. The `task` sub-agent tool (`crates/stella-tools/src/subagent.rs`) never
calls it, so a session that delegates through `task` records **zero** rows in
`agent_uses` however many children it spawned.

Observed in an export audit of session `ses-1787092335250-47513` (execution
157): 8 `task` delegations, `agent_uses` exported 0 rows. There is therefore no
audit trail of sub-agent health at all — which is how the sibling defect in that
same session (a child that completed with an empty report) stayed invisible until
a human read the tool-call log by hand.

## Why it is not a one-line fix

`SpawnSubAgent::execute` receives a `&crate::ctx::ToolCtx`, and `ToolCtx` has no
handle on the registry's ledger — its surface is scope/path resolution plus
`emit`/`progress` (`crates/stella-tools/src/ctx.rs`). The tool also cannot hold an
`Arc<ToolRegistry>`: the registry owns the tool, and the dispatcher slot exists
precisely to avoid that cycle (see `subagent.rs`'s module docs, "Why a slot
rather than a constructor argument").

So this needs a deliberate seam. Two candidates, both reviewable:

1. A late-attached ledger handle on the tool, mirroring `DispatcherSlot`.
2. A `ToolCtx` method that reaches the ledger, alongside `emit`/`progress` —
   wider, but it would serve any future tool that needs to attribute.

Pick one and say why in the PR.

## Files

- `crates/stella-tools/src/subagent.rs` — the tool with no attribution
- `crates/stella-tools/src/registry.rs:592` — `record_agent_use`
- `crates/stella-tools/src/agent_use.rs` — the ledger
- `crates/stella-cli/src/agent/persistence.rs:387` — `drain_agent_uses`, the drain
- `crates/stella-tools/src/ctx.rs` — the context a tool actually gets

## Verify

Witness test: dispatch a child through `SpawnSubAgent::execute` with a recording
registry and assert one `agent_uses` row with the minted `agent_id`. It must fail
before the change (zero rows today).

## Done

A `task` call produces an `agent_uses` row carrying the child's minted agent id,
drained by `record_execution_end` like the deck's. `ToolCtx`'s doc says which
seam was chosen and why the other was not.

[...]

Issue #3690, opened 2026-08-18
About 30% of prompt tokens miss the cache on a read-heavy turn, and the cause is not known yet


macanderson:
## Problem

Execution 150 of session `ses-1787071651715-48158` (`.stella/exports/session-2026-08-18-19-10-29.zip`, `raw/telemetry.json`) shows a **~30% prompt-cache miss rate**, concentrated in one contiguous band of steps.

Totals for the execution (87 steps, `moonshotai/kimi-k3` via `openrouter`):

```
input_tokens        3,634,137
cache_read_tokens   2,561,500
cache_miss_tokens   1,072,637   (29.5%)
cache_write_tokens          0
cost                    $3.77
```

Compare execution 148 in the same session, same provider and model: 1,144,447 input / 1,050,227 read / 94,220 miss = **8%**.

The miss is not spread evenly. Worst steps:

```
step  miss    read     input
219   60,306   8,192   68,498
164   60,210   5,120   65,330
155   57,659   4,096   61,755
642   56,932       0   56,932
201   55,856   8,192   64,048
228   54,910  10,240   65,150
```

Steps ~155-246 read only 4-11K of a 50-65K prompt — effectively a cold prompt on every step. By the tail (steps 1500+) it recovers to `cache_read` ~= `input` (e.g. step 1567: 49,865 in / 49,408 read), which is healthy.

The band coincides with the turn's large `read_file` results landing (17 `read_file` calls, two of them ~65KB single files).

## What is NOT the cause

`cache_write_tokens = 0` across all 163 telemetry rows is **not** a wiring defect. It was checked: `crates/stella-model/src/zai/stream.rs:57` assigns `usage.cache_write_tokens = details.cache_write_tokens`, with witness test `complete_surfaces_cache_write_tokens` in `crates/stella-model/src/zai/tests.rs`. OpenRouter genuinely reported no cache writes for this model.

## What is unknown

The mechanism is **not established**. This is an observation, not a diagnosis. Candidate hypotheses, none verified:

1. Large tool results appended between calls push the cached prefix past a breakpoint that is not re-established until the content stabilises.
2. OpenRouter's `cache_control` / sticky `session_id` (`crates/stella-model/src/zai.rs:1271-1274`) is applied but the upstream re-keys on payload size.
3. The provider's implicit caching simply has a warm-up cost proportional to how fast the prompt is growing, and 30% is the floor for a read-heavy turn.

[...]

PR #3809, opened 2026-08-19
fix(stella-tools,stella-cli): three silent-loss defects from a session export audit

macanderson:
Three silent-loss defects found by an export audit of a real session
(`ses-1787092335250-47513`, execution 157: 66 edits across 35 files, cancelled
after the build passed). Each one made a failure look like a success.

## 1. A sub-agent's empty report read as `ok`

`render` in `crates/stella-tools/src/subagent.rs` split on an empty summary in
both `Incomplete` arms but not in `Completed`, so a child that finished and said
nothing returned a plain `ok` with no content — indistinguishable from a sweep
that genuinely found nothing to report. In the audited session one sweep covering
39 pages came back at 354 bytes against ~8 KB for its three siblings; the parent
recovered only by noticing the thinness itself and paying for the whole
delegation a second time (~5 minutes).

## 2. `files_touched` had no producer on the engine path

#3413 gave that path an `AgentEvent::FileChange` producer but no durable one,
leaving candidate adoption as the table's only writer
(`crates/stella-cli/src/candidate_ws/adoption_ledger.rs`). Every direct-edit turn
therefore recorded zero touched files however many it wrote.

This also explains a second symptom that looked unrelated:
`Store::finalize_execution_reflection` derives `wrote_files` from exactly that
table, so the reflection loop was told a turn that edited 35 files had written
none.

The turn-boundary measurement is now taken **once** and both projections derive
from it. Snapshotting twice would not merely cost a second `--numstat` — the
journal's snapshot advances its own baseline, so the second reading would report
an unchanged tree and the two projections would disagree about the same turn.

Both writes stay best-effort: observability, not evidence (#2882).

## 3. An indentation-only `edit_file` miss got a generic nudge

The unchanged-file arm of the #331 attribution is accurate but not actionable —
"check for exact whitespace/newline differences" costs a ranged re-read to act
on, which is what it cost in the audited session (a needle missing a three-space
list indent, twenty tool calls to recover). When the needle's lines match the
file's after stripping leading whitespace only, the tool now says so and echoes
the file's own bytes for the span.

Deliberately narrow, so it can never claim a match the literal edit would not
have made: trailing whitespace still counts as a real difference, an ambiguous
second window declines rather than guessing, and the needle is bounded — this is
a hint inside an error, not a file echo.

## Witnesses

- `a_completed_child_with_an_empty_report_is_an_error_not_a_silent_ok` —
  verified by removing the fix: FAILED without it, passes with it.
- `a_measured_change_becomes_a_durable_row_carrying_the_same_counts` and
  `durable_rows_use_the_crud_letters_both_reader_queries_match` — the second
  pins the `ops` alphabet both reader queries (`execution_rollup`'s
  `files_written`, `finalize_execution_reflection`'s `wrote_files`) grep for, so
  a row spelled any other way fills the table while both answer zero.
- `an_indentation_only_miss_hands_back_the_files_own_bytes` and
  `the_hint_declines_when_it_would_be_a_guess`. The pre-existing
  `unchanged_file_failure_is_not_drift_attributed` still passes, so the new
  branch does not hijack the generic path.

## Two audit findings I did NOT act on, because the premises are false

Recorded so the next reader does not re-derive them:

- *"Context drops are invisible to the agent."* They are not. Every compaction
  pass writes a model-visible stub (`EVICTION_STUB`, `DEDUP_STUB`,
  `SUPERSESSION_STUB`, `AGE_ELISION_MARKER`), and the last of those literally
  says "re-run the tool for the full output". The re-reads the audit called
  redundant are that stub's instructed recovery working as designed.
- *"`edit_file`'s staleness error is indistinguishable from never-existed."*
  `crates/stella-tools/src/edit.rs` already attributes all three causes —
  drifted (with a fresh-content echo), unchanged, and no-read-recorded — to
  three different messages with three different recoveries. Only the
  actionability of the *unchanged* arm was a real gap, which is item 3 above.

## Filed rather than fixed

- #3807 — `task` delegations record no `agent_uses` rows; the tool cannot reach
  the ledger without a new seam (`ToolCtx` has no handle, and the tool cannot
  hold the registry that owns it).
- #3808 — a cancelled execution's reflection is a silent stub. Note the
  `wrote_files = 0` half of that symptom is fixed here; the empty critique and
  null rating are a separate design call.
- #3690 already covers the prompt-cache finding (0 cache writes across 3.03M
  input tokens) — linked, not duplicated.

## Summary by Sourcery

Prevent successful-looking session outcomes from hiding empty sub-agent reports, missing file-change records, and recoverable indentation-only edit failures.

New Features:
- Provide actionable guidance when an `edit_file` replacement differs only by leading indentation.

Bug Fixes:
- Treat completed sub-agents with empty reports as errors instead of silent successes.
- Persist turn-boundary worktree changes to `files_touched` so direct edits contribute accurate file and reflection metrics.

Enhancements:
- Derive event and durable file-change projections from a single worktree measurement while keeping both writes best-effort.

Tests:
- Add coverage for empty completed sub-agent reports, durable file-touch projections and CRUD operation markers, and safe indentation-only edit hints.

### DIFF
--- a/crates/stella-cli/src/agent.rs
+++ b/crates/stella-cli/src/agent.rs
@@ -1694,7 +1694,7 @@ pub(crate) async fn run_turn(
     // boundary and emitted *before* the close below: these are the turn's own
     // events and a consumer folding the stream must see them inside the turn
     // they describe. See `crate::turn_files` for why a measurement, not a hook.
-    crate::turn_files::emit_shared_tree_changes(cfg, &tx);
+    crate::turn_files::emit_shared_tree_changes(cfg, &tx, execution.as_ref());
     // This path owns its run — one raw engine turn, no pipeline above it — so
     // it owes the run's terminator (#3379). The engine ends the turn with
     // `TurnComplete` and deliberately says nothing about the run, and every
--- a/crates/stella-cli/src/turn_files.rs
+++ b/crates/stella-cli/src/turn_files.rs
@@ -132,8 +132,10 @@ fn file_touch_row(change: &JournalChange) -> FileTouchRow {
 /// The `ops` alphabet is CRUD letters, fixed by the reader rather than by
 /// preference — `Store::execution_rollup` and `finalize_execution_reflection`
 /// both match `ops LIKE '%C%' OR '%U%' OR '%D%'`, so a row spelled any other
-/// way is a row those queries cannot see. Same alphabet as
-/// [`crate::candidate_ws::adoption_ledger`], deliberately.
+/// way is a row those queries cannot see. Same alphabet as the adoption
+/// ledger's `ops_letter` (`candidate_ws/adoption_ledger.rs`), deliberately —
+/// named in prose rather than linked because that module is private, and an
+/// intra-doc link to it does not resolve.
 fn ops_letter(kind: JournalChangeKind) -> &'static str {
     match kind {
         JournalChangeKind::Created => "C",
--- a/crates/stella-cli/src/turn_files.rs
+++ b/crates/stella-cli/src/turn_files.rs
@@ -57,26 +57,88 @@
 //! grandfathered god file, closed to growth (AGENTS.md § "God files — plan
 //! around them, never into them").
 
+use std::sync::Arc;
+
 use stella_core::EventSender;
 use stella_protocol::event::{AgentEvent, FileChangeKind};
 use stella_store::work_journal::{JournalChange, JournalChangeKind};
+use stella_store::{FileTouchRow, Store};
 
 use crate::config::Config;
 
-/// Measure the shared work tree and emit one [`AgentEvent::FileChange`] per
-/// path this turn changed.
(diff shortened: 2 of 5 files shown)


macanderson:
Scoped gate green: `make gate CARGO_SCOPE="-p stella-tools -p stella-cli"` exits 0 — 29 steps through `self-driving-test` (60 checks), 20 test binaries green.

Scope note, so "gate green" is not read as more than it is: this is narrower than CI's `--workspace`. It covers both crates this PR touches and their dependents, but the full workspace run is CI's.

Two gate repairs landed in dde10d5c8, both mine and neither a behaviour change: rustfmt reflowed an `assert_eq!` in a new test, and an intra-doc link pointed at `candidate_ws::adoption_ledger` — a private module rustdoc cannot resolve under `-D warnings`, now named in prose rather than silenced with an `#[allow]`.

PR #3821, opened 2026-08-19
fix(stella-tools): record an agent_uses row for every `task` delegation

### LINKED ISSUE #3807: task sub-agent delegations record no agent_uses rows — the tool cannot reach the ledger

macanderson:
## Problem

`ToolRegistry::record_agent_use` had exactly one production caller in the workspace — `crates/stella-cli/src/command_deck.rs`, the deck's named-agent path. The `task` sub-agent tool never called it, so a session that delegated through `task` recorded **zero** `agent_uses` rows however many children it spawned (8 delegations → 0 rows in the audited session), leaving no audit trail of sub-agent health.

## The seam, and why this one

The issue offered two: a late-attached ledger handle on the tool mirroring `DispatcherSlot`, or a `ToolCtx` method.

**`ToolCtx` wins.** `DispatcherSlot` exists to break a genuine reference cycle — the dispatcher owns the registry that owns the tool. The ledger has no such cycle to break, so a slot would buy only a nullable handle, an attach call every host must remember, and a tool that silently records nothing when one forgets. The registry already builds one context per call and already owns the ledger, so a clone rides along beside the bus with no new host step, and `ToolCtx::record_agent_use` serves any future tool that needs to attribute. Attribution is not the contract-gated vocabulary `emit` is (a row names its own recorder and drains per execution either way), so it needs no allowlist; a `ToolCtx::bare` context has no ledger and records nothing — the same reported-drop shape `emit` already has. This is stated in `ctx.rs`'s module docs, per the issue's "Done".

The row is recorded **before** the dispatch await, for the same reason the dispatcher settles spend from the child's own thread: a hard cancel means the line after it never executes, and a delegation that vanished is exactly what the row exists to make visible. `version` is 1 — a `task` child is minted here rather than loaded from a versioned on-disk definition, and 1 is the ledger's own un-versioned convention.

## Witness

`registry::tests::a_task_delegation_records_an_agent_use_row` dispatches `task` through `ToolRegistry::execute` with a stub dispatcher and then calls `drain_agent_uses` — the same two calls the session driver makes, so it measures the wiring rather than the new method.

Verified both directions: with the `ctx.record_agent_use` call disabled it fails with `assertion left == right failed: one delegation, one row: []`; with it, 417/417 `stella-tools` tests pass.

## Gate

`make gate CARGO_SCOPE="-p stella-tools"` green (29 steps, incl. clippy `-D warnings`, rustdoc, fmt, file-size, god-files). `cargo check -p stella-cli -p stella-observatory` clean — `record_agent_use`/`drain_agent_uses` signatures are unchanged, only the registry's field type (now the shared `AgentUseLedgerHandle`).

No tests deleted. No new dependencies.

Closes #3807

## Summary by Sourcery

Record every `task` delegation in the session agent-use ledger for reliable sub-agent auditing.

New Features:
- Record an `agent_uses` event for each child agent delegated through the `task` tool, including its identifier, version, and delegation reason.

Bug Fixes:
- Restore the missing audit trail for `task` sub-agent delegations so delegated children are visible to session persistence and health reporting.

Enhancements:
- Propagate the session agent-use ledger through `ToolCtx`, allowing tools to attribute delegations without additional host-side setup.

Documentation:
- Document why agent-use attribution is provided through `ToolCtx` and how missing ledgers are handled.

Tests:
- Add an end-to-end registry test verifying that a `task` delegation produces exactly one drainable agent-use record.

### DIFF
--- a/crates/stella-tools/src/agent_use.rs
+++ b/crates/stella-tools/src/agent_use.rs
@@ -14,7 +14,18 @@
 //! `stella-cli::agents_installed` for the on-disk versioning scheme), so the
 //! telemetry can attribute outcomes to the exact definition text that ran.
 
+use std::sync::{Arc, Mutex};
+
 use serde::{Deserialize, Serialize};
+
+/// A shared handle on the session's ledger.
+///
+/// The registry owns one and hands clones out — to
+/// [`crate::ctx::ToolCtx`] per call, so a tool can attribute its own
+/// delegations (#3807). Shared rather than borrowed because the context
+/// crosses an `async_trait` boundary on every dispatch, exactly like the
+/// hook bus beside it.
+pub type AgentUseLedgerHandle = Arc<Mutex<AgentUseLedger>>;
 
 /// One invocation of an installed agent definition.
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
--- a/crates/stella-tools/src/ctx.rs
+++ b/crates/stella-tools/src/ctx.rs
@@ -21,6 +21,31 @@
 //! failing the call over it would let an observability bug break a working
 //! tool.
 //!
+//! # Why attribution lives here and not on the tool (#3807)
+//!
+//! The `task` tool mints an agent id per delegation and recorded it nowhere:
+//! `agent_uses` had exactly one writer, the deck's named-agent path, so a
+//! session that delegated eight children exported zero rows and sub-agent
+//! health had no audit trail at all.
+//!
+//! Two seams were available. The first — a late-attached ledger handle on
+//! `SpawnSubAgent`, mirroring
+//! [`DispatcherSlot`](crate::subagent::DispatcherSlot) — was **not** taken:
+//! that slot exists to break a genuine reference cycle (the dispatcher owns
+//! the registry that owns the tool), and the ledger has no such cycle to
+//! break. Paying a slot's cost — a nullable handle, an attach call every host
+//! must remember, and a tool that silently records nothing when one forgets —
+//! buys nothing here, and the next tool that wants to attribute would need its
+//! own copy of it.
+//!
+//! The second is this: the registry already builds one context per call and
+//! already owns the ledger, so a clone rides along beside the bus with no new
+//! host step at all, and [`ToolCtx::record_agent_use`] serves any future tool.
+//! Attribution is not the contract-gated vocabulary `emit` is — a row names
+//! its own recorder and is drained per execution either way — so it needs no
+//! allowlist. A [`ToolCtx::bare`] context has no ledger and records nothing,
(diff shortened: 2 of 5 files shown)


macanderson:
Filed #3822 alongside this: now that `task` writes to `agent_uses`, the Observatory's session `agents` panel groups installed-agent names and per-delegation minted slugs in one `GROUP BY u.agent`, so eight delegations render as eight singleton groups. Noticed here, deliberately not fixed in this PR — it wants a schema decision.

Issue #1721, opened 2026-08-06
Flaky: a_run_that_ignores_term_is_killed_after_the_grace_period fails under full-suite load, passes alone


macanderson:
## Problem

`daemon::tests::a_run_that_ignores_term_is_killed_after_the_grace_period` fails intermittently when the whole `stella-cli` test binary runs, and passes reliably in isolation. Observed while verifying PR #1717 (the unbreak of `main`):

- `cargo test -p stella-cli --bin stella` → `1407 passed; 1 failed`, this test.
- `cargo test -p stella-cli --bin stella daemon::tests::a_run_that_ignores_term_is_killed_after_the_grace_period` → passes, 8.27s.
- `cargo test -p stella-cli --bin stella` again → `1408 passed; 0 failed`.

Same commit, same machine, no code change between the three runs. So it is a flake, not a defect in the code under test.

## Why it is load-sensitive

The test spawns a child that ignores `SIGTERM`, calls `daemon::stop`, and asserts the `SIGKILL` escalation lands. `stop` waits `STOP_GRACE` (8s, `crates/stella-cli/src/daemon.rs`) polling the liveness lock every `POLL` (80ms) before escalating, and the assertion then waits for the process to actually go.

That is a real 8-second wall-clock window, running concurrently with ~1400 other tests — several of which (`daemon::tests`, `daemon::detach::tests`) also spawn real child processes and signal process groups. On a loaded machine the child's death, the lock release, or the polling thread can all slip past whatever bound the assertion allows. A flake here is a scheduling artifact, not a supervision bug.

## Files

- `crates/stella-cli/src/daemon/tests.rs` — the test, and `eventually()`, the shared bounded-wait helper.
- `crates/stella-cli/src/daemon.rs` — `STOP_GRACE`, `POLL`, `stop`, `lock_is_held`.

## Constraints already discovered

- **Do not just widen the timeout and call it fixed** without checking which wait actually expired — `eventually()` already uses deliberately generous budgets ("a bound tight enough to be a latency assertion would be a flake on loaded CI", per its doc comment), so a bound that still lost a race is evidence about *which* wait is tight, not a number to double blindly.
- This probe has been CI-hostile before: #1610 quarantined it once ("quarantine the CI-hostile stop probe"), and #1632 un-quarantined it. Read both before changing its shape — re-quarantining would drop coverage of the `SIGKILL` escalation rung, which is one of the three bugs `scripts/fullauto.sh` paid for and which the module docs call out by name.
[...]

PR #3854, opened 2026-08-19
feat(stella-protocol,stella-cli): the issue port — GitHub becomes an adapter (#3599 B1)

macanderson:
## What & why

B1's first slice of #3599: the backlog plane gets a port, so GitHub stops being
the only tracker without becoming a special one — invariant 1, for issues.

The loop read its defect queue by shelling out to `gh` literally, in the
shipping binary, and the type the ranker consumed **was** GitHub's wire shape:
`stella_autonomy::QueueIssue` carries `#[serde(rename = "createdAt")]`, a
`gh --json` field name sitting in a leaf crate that depends on nothing. "Which
tracker" was not a decision anything in the tree could express; it was spelled
into the reader.

- `crates/stella-protocol/src/issue.rs` — the `Issue` kernel and the
  `IssueProvider` port, following `provider.rs`'s precedent for a port in that
  crate.
- `crates/stella-cli/src/issue_provider.rs` — the GitHub adapter, and the only
  place in the tree carrying GitHub's field spellings.
- `crates/stella-cli/src/self_driving_cmd/backlog.rs` — both backlog readers,
  over `&dyn IssueProvider`.

The kernel is deliberately tiny (`doc:agent-native-delivery` §3). **No
`assignee`** — a tracker's assignee field is not a lock, and treating it as one
is how two workers take the same issue; claims belong in the fleet ledger.
**No `priority` field** — every tracker this must span carries priority as a
label, and a dedicated field would force each provider to invent a mapping into
it.

`stella_autonomy` keeps its own ranker input type rather than moving onto
`Issue`: it takes no workspace dependency at all, that is what lets
`stella-observatory` link its folds without linking the machinery it observes,
and the mapping is one function in the CLI.

Refs #3599

## Two things found by building it, beyond the design's scope

**`demand` carried a second definition of "defect".** The governor counted its
open-defect and P0 numbers with two extra `gh issue list --jq length` calls
whose `--label bug` filter was a rival to `rank_defects`'s own label rule — two
definitions of one word, in two languages, able to disagree about the very
backlog the same cycle drew its batch from. Both numbers are now folds of one
ranking of one read, which also drops two subprocess round-trips per plan.

**New logic landed in a sibling.** #3599 records `self_driving_cmd.rs` as closed
to growth. I grew it by 56 lines first, noticed, and split: the parent is now
950 lines, down from 1005.

## Deliberately absent

The write half — `backlog claim` / `file` / `close` / `link` — is not here. A
write path nothing calls is the unwired code AGENTS.md's "nothing left behind"
rule exists to prevent; it lands with the slice that serves those calls. Same
reasoning as B0's: the `DriverCall` variants for them already exist on `main`
and answer `unsupported`.

## The witness

`self_driving_cmd::backlog::tests::a_ranked_queue_needs_no_github_at_all` ranks
a real defect queue — priority order, age breaking ties, feature work excluded —
from a fixture provider that runs no subprocess and holds no credential.

**It cannot compile on `main`**: there is no `IssueProvider` to implement there,
because the queue read *is* the `gh` call. That is the property under test — not
that ranking works, which it already did, but that ranking no longer requires
GitHub to exist. Verified absent: `git show origin/main:crates/stella-protocol/src/issue.rs`
returns nothing.

Three more pin what the crossing must preserve: `demand_is_a_fold_of_the_same_ranking`,
`an_unreachable_tracker_degrades_rather_than_failing` (a tracker that cannot be
reached sizes a cycle as though the backlog were empty rather than refusing to
plan — inherited from the `gh_available()` check it replaces), and
`the_read_limit_bounds_the_crossing`. Plus five serde/mapping tests on the
kernel and adapter, including invariant 4's byte-for-byte round trip.

One test caught me rather than the code: my first fixture had `P0`/`P1` labels
but no `bug` label, and `rank_defects` correctly returned nothing.

## The gate

- [x] `cargo fmt --check`
- [x] `cargo clippy -p stella-cli -p stella-protocol --all-targets -- -D warnings`
- [x] `cargo test -p stella-protocol -p stella-cli`
- [x] `wire-schema`, `doc-warnings`, `file-size`, `god-files`, `design-refs`,
      `invariants`, `doc-links`, `brand-case`, `left-behind`, `typed-errors`,
      `module-reachability`, `no-scratch`, `format-check`

`shellcheck` is not installed in this container; this PR touches no shell.

### CI round 1 caught a real thing

`crates/stella-cli/tests/self_driving_consent.rs` asserts, in **both**
directions, that every capability the install prompt declares is one the code
actually exercises — and it fired:

> the grant declares `reading the GitHub defect queue` but
> `crates/stella-cli/src/self_driving_cmd.rs` no longer does it … Narrow the
> grant rather than leaving a capability declared that nothing needs, or repoint
> this row if the power moved.

The power moved rather than went, so the row is repointed at
`crates/stella-cli/src/issue_provider.rs` (`4e9a48d`). The new anchor is
`impl IssueProvider for GhIssueProvider` rather than another argv string: an
argv anchor fires spuriously the first time someone adds a `--json` field, which
teaches a reader that this guard cries wolf, while the impl disappears only if
the capability really does.

### Correction to an earlier claim in this description

An earlier revision of this body reported
`daemon::tests::a_run_that_ignores_term_is_killed_after_the_grace_period` as a
pre-existing failure. It fails in *this* container — and does so on clean
`origin/main` with my changes stashed, which is why I read it as pre-existing —
but **CI runs it green**, on this branch. So the honest statement is narrower
than what I wrote: it is an environment-sensitive failure local to a
disk-constrained sandbox, not a failure of `main`. The test's own comment
documents that flake mode and attributes it to #1721.

## Nothing left behind

- [x] Filed: nothing new. The write half and the non-numeric-key limitation are
      both recorded in #3599 and noted in the code where they bite
      (`to_queue_issue`'s doc comment states the `u64` key loss and why widening
      a leaf crate's public type should wait for a provider that needs it).

## Ground-rule check

- [x] No I/O added to `stella-core`; no new deps (`async-trait` was already a
      `stella-protocol` dependency)
- [x] No new outbound network calls — the adapter shells to `gh` exactly as the
      code it replaces did, which is also what keeps **Stella holding no GitHub
      credential**: `gh` owns the auth and the token never enters this process
- [x] New cross-boundary types round-trip through serde (test included)

**The design's landing map said providers land in `stella-tools`; this puts the
port in `stella-protocol` and the adapter in `stella-cli`.** Reasoning: the port
is a trait crossing a crate boundary, which is `provider.rs`'s precedent
exactly, and `stella-tools` is where a *tool* lives — the issue port is not a
tool the model calls. If the intent was for the agent to have an issue *tool*,
that is a separate surface and this port would sit underneath it. Worth
confirming before B1's second slice builds on the choice.

**`IssueClass` for GitHub is read off labels**, since a plain repository has no
issue type. Unlabelled maps to `Other`, deliberately not `Task`: a wrong class
silently applies the wrong completion policy, and `doc:agent-native-delivery`
§12.6 leaves whether `Other` is workable open — this adapter does not pre-empt
it.

**`parent` is always `None` from GitHub here.** The parent edge needs a second
GraphQL call per issue that the queue read does not need and should not pay for.
Absent means "not fetched", and the write half fetches it when something reads
it.

### DIFF
--- a/crates/stella-cli/src/issue_provider.rs
+++ b/crates/stella-cli/src/issue_provider.rs
@@ -1,0 +1,254 @@
+//! The GitHub adapter behind [`IssueProvider`], and the one mapping from a
+//! tracker's shape into the ranker's.
+//!
+//! `doc:backlog-self-driving` B1. Until this module existed, `gh` was invoked
+//! literally inside `self_driving_cmd::queue` and the ranker's input type
+//! carried `#[serde(rename = "createdAt")]` — a `gh --json` field name, in a
+//! leaf crate that depends on nothing. "Which tracker" was therefore not a
+//! decision anything could express; it was spelled into the reader.
+//!
+//! What moved here is the I/O half only. The port and the kernel are
+//! [`stella_protocol::issue`]; the ranking stays in `stella_autonomy`, on its
+//! own input type, because that crate depends on no other workspace crate and
+//! that property is what lets the Observatory link it.
+//!
+//! # `gh`, not the REST API — for now, and for a reason
+//!
+//! The adapter shells out to `gh` rather than speaking HTTP. That is what the
+//! loop already did, so this slice is a relocation rather than a rewrite, and
+//! it keeps the one property a token-holding client would have to re-earn:
+//! **Stella never holds a GitHub credential.** `gh` owns the auth, the token
+//! never enters this process, and `stella auth` has nothing new to store.
+//! `doc:agent-native-delivery` §4.4's `kind = "exec"` is the general form of
+//! exactly this shape, and a manifest-driven provider is the next slice.
+
+use std::process::Command;
+
+use async_trait::async_trait;
+use stella_protocol::issue::{
+    Issue, IssueClass, IssueError, IssueKey, IssueLabel, IssueProvider, IssueState,
+};
+
+/// The provider id this adapter answers to, and the one an error names.
+pub(crate) const GITHUB: &str = "github";
+
+/// Reads the workspace's GitHub issues through the `gh` CLI.
+pub(crate) struct GhIssueProvider;
+
+/// What `gh issue list --json …` writes, named separately from [`Issue`] on
+/// purpose.
+///
+/// This is the **only** type in the tree allowed to carry GitHub's field
+/// spellings, and it is private to this module. Before B1 the same
+/// `createdAt` rename lived on the ranker's public input type in a leaf crate,
+/// which is how a tracker's wire format became the loop's vocabulary.
+#[derive(serde::Deserialize)]
+struct GhIssue {
+    number: u64,
+    title: String,
+    #[serde(default)]
+    body: String,
+    #[serde(default)]
+    labels: Vec<GhLabel>,
+    #[serde(rename = "createdAt", default)]
+    created_at: String,
+    #[serde(default)]
+    url: String,
+}
+
+#[derive(serde::Deserialize)]
+struct GhLabel {
+    name: String,
+}
+
+impl GhIssue {
+    /// GitHub has no issue *type* on a plain repository, so the class is read
+    /// off the labels this repository actually uses.
+    ///
+    /// [`IssueClass::Other`] is the honest answer for an issue labelled
+    /// neither, and it is deliberately not `Task`: a wrong class silently
+    /// applies the wrong completion policy, which is worse than a visible
+    /// unmapped one (`doc:agent-native-delivery` §12.6 leaves whether `Other`
+    /// is workable open, and this adapter does not pre-empt it).
+    fn class(&self) -> IssueClass {
+        let has = |name: &str| self.labels.iter().any(|label| label.name == name);
+        if has("bug") {
+            IssueClass::Bug
+        } else if has("feature") {
+            IssueClass::Feature
+        } else if has("chore") || has("task") || has("refactor") {
+            IssueClass::Task
+        } else {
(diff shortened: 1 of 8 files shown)


macanderson:
CI red on `fmt + clippy + test`, **failing on the base branch too** — not this PR.

```
registry::tests::a_task_delegation_records_an_agent_use_row
  unknown tool `task` — available: bash, delegate, … task_start, write_file
```

`crates/stella-tools/src/registry/tests.rs:291` dispatches `execute("task", …)`,
but there is no `task` tool: the delegation tool is `delegate`, and `"task"` is
in `RETIRED_NAMES_TOO_AMBIGUOUS_TO_SCAN`. Introduced by #3821 (`154c596`).
Verified by stashing this branch and running the test on `origin/main` at
`c6b181b` — same failure, clean tree. This PR touches neither `stella-tools`
nor anything it depends on.

Filed as #3858 with the diagnosis and the one-word fix.

The round-1 failure on this PR *was* mine and is fixed in `4e9a48d`: the
consent-parity guard correctly noticed that B1 moved the defect-queue read out
of `self_driving_cmd.rs`, so its row is repointed at the adapter.

I have not folded the `stella-tools` fix into this branch — a red-`main` fix
should land on its own from a fresh `main` rather than inside an unrelated PR.
Happy to do that as a separate change if you want it; say the word and I will.
Otherwise I will re-run here once `main` recovers.

---

Issue #3865, opened 2026-08-19
Delete crates/stella-pipeline from the workspace (removal census, §7's "last slice")


macanderson:
## What

`docs/spec/pipeline-as-plugins.md` §7 names the dependency cut — `stella-cli`
no longer declaring `stella-pipeline`, and the crate itself leaving the
workspace — as the extraction's **last slice**, done only once every plugin
in Track B has a bench-verified replacement path. This issue tracks that
final cut as its own unit of work, since nothing else in the tracker names
it directly (searched: #3246, #3379–#3381, #3511, #3694, #3860, #3861 all
describe pieces of the extraction; none is "delete the crate").

## Scope

- Remove `crates/stella-pipeline` from the workspace (`Cargo.toml` members,
  the crate directory itself, `Cargo.lock`).
- Every `stella_pipeline::`/`stella-pipeline` reference in the rest of the
  tree: RETARGET the handful of types/logic that still need a home (found:
  none remaining after this pass — see below), DELETE-WITH-PATH the rest
  (code whose only purpose was serving the classic in-tree dispatch).
- `scripts/file-size-baseline.txt`, `AGENTS.md`'s crate table / god-file
  table / crate-count prose, `crates/stella-cli/README.md`'s stale
  reference-count line, `crates/stella-parity/src/lib.rs`'s capability
  matrix rows that cited the removed crate as their evidence.
- A compatibility path for an on-disk resume frame written by a pre-removal
  build that named the staged pipeline: it must still deserialize (byte
  stability of `.stella/private/`'s durable state), degrading to a graceful,
  clearly-worded refusal rather than attempting to reconstruct pipeline
  state that no longer exists.

## Status as of this issue

The crate deletion, the workspace-member/Cargo.lock edits, and the
compatibility shim above have landed. `cargo check --workspace`,
`cargo clippy --workspace --all-targets -- -D warnings`, and
`cargo test -p stella-cli -p stella-runtime -p stella-plugin -p stella-store`
are all green against the resulting tree.

**Not done here, and worth a fresh pass:** a systematic reference sweep of
the *other* crates named in the original removal census (§1.3's list —
`stella-autonomy`, `stella-context`, `stella-core`, `stella-fleet`,
`stella-parity`, `stella-plugin`, `stella-protocol`, `stella-runtime`,
`stella-tools`, `stella-tui` READMEs and a handful of their `src/*.rs` doc
comments) still describe the removed crate in present tense or cite it as a
[...]

PR #3915, opened 2026-08-19
Revert "fix(workspace): stop the tree describing the deleted stella-pipeline as live (#3865)"

macanderson:
Reverts macanderson/stella#3902

## Summary by Sourcery

Restore the staged pipeline's workspace references, documentation, and test coverage while retaining safeguards against inappropriate runtime coupling.

Enhancements:
- Restore references and documentation for the staged pipeline and its former workspace integration points.
- Reintroduce the pipeline-specific search witness and properties-lens scope while preserving the runtime dependency guard under a renamed test.

CI:
- Update workspace member-count documentation to reflect the restored pipeline crate.

Documentation:
- Restore documentation describing staged-pipeline event producers, replay consumers, lanes, and related integration points.

Tests:
- Restore the pipeline symbol search-recall witness and replace the deleted-path lens validation with a persistent runtime dependency-edge guard.

Chores:
- Revert the prior cleanup that removed stella-pipeline references and associated workspace assumptions.

---

## Tests deleted by this revert, and why

`check-deleted-tests` names two. Both are deliberate, and both are structurally
required by what the revert restores — neither is the silent-drop shape #1976
put the guard there to catch:

- **`every_forbidden_crate_still_exists`** — added by #3902 in
  `crates/stella-runtime/tests/no_policy_crate_edge.rs`, which this revert
  deletes along with the rest of that file. It asserts every name in
  `FORBIDDEN_DEPENDENCIES` is a crate that exists in the tree. The restored
  `no_pipeline_edge.rs` deliberately keeps `stella-pipeline` listed *after* the
  crate's deletion — its module doc says so outright ("The crate itself is gone
  (#3865); the name stays listed so re-homing verification cannot quietly bring
  the edge back") — so the two cannot both hold. The edge assertion itself
  survives unchanged as
  `the_assembly_seam_declares_no_edge_to_a_policy_crate`, which is the guard
  that has teeth; only the existence check goes.

- **`every_crate_path_a_lens_names_exists_in_the_tree`** — added by #3902 in
  `crates/stella-autonomy/src/tests.rs`, removed here with the rest of that
  hunk. Checked against the merged tree: no `LENSES` entry names
  `crates/stella-pipeline/src` any more, so nothing this revert produces
  depends on the check to stay honest today. It is a guard worth having back on
  its own terms rather than as a side effect of whichever way this revert goes,
  so restoring it is filed as #3921 rather than left in the transcript.

## CI on this branch

Three failures, from three unrelated causes, none of them the revert's logic:

1. **`cannot find value ORACLE_RED in module palette`** (msrv, clippy, doc,
   test, release smoke, `stella context validate`) — inherited from a `main`
   that had not compiled since #3896, the sequence #3917 records. Fixed on
   `main` by #3918 (`fc116b773`) and picked up here by merging it.
2. **`shellcheck`** — `scripts/test-install-zsh-completions.sh:112` carried a
   SC2002 useless cat from #3889 (`7971cbefb`), which itself merged onto the
   already-red base above, so the failure never distinguished itself. Merged in
   from #3920, which fixes it on `main` independently of this revert's fate.
3. **`check-deleted-tests`** — the two tests named above; this section is the
   acknowledgement the guard asks for.


macanderson:
Heads-up before this merges: **#3902 is not what broke `main`**, so this revert
will not make it green.

The required `fmt + clippy + test` check was already `failure` on `85b7e07ee`
(the merge immediately before #3902) and on `93eec5a50` before that:

```
$ gh api repos/macanderson/stella/commits/85b7e07ee/check-runs \
    --jq '.check_runs[] | select(.name=="fmt + clippy + test") | .conclusion'
failure
```

`main` has not type-checked since #3896. Three separate PRs each deleted a
symbol and left a caller behind, and because one compile error stops the build
before the next crate is reached, each hid the next:

1. #3890 deleted `palette::ORACLE_RED`; #3897 then added
   `theme.rs::ORACLE_PRE_FLIP` referencing it.
2. #3900 rewrote `wire_contract.rs`'s import block and dropped
   `EvidenceProvenance`, leaving two usages unimported.
3. #3896 deleted `reject_plugin_variant_for_door`; #3897 added a test calling
   it.

Note #3902 appears nowhere in that list. Reverting it would also give back the
`no_policy_crate_edge.rs` guard rename (the old guard names a deleted crate, so
it can never fail), the self-driving `properties` lens still `rg`-ing a path
that no longer exists — which makes that whole aperture report "no findings"
for a reason unrelated to the code — and the `run_complete` consumer row still
pointing at `stella-pipeline/src/replay.rs::validate_stream`.

**#3918** fixes all three compile errors against current `main` and is green on
`fmt` and `clippy --workspace --all-targets -D warnings`. **#3917** tracks the
dynamic that let four PRs land on a non-compiling `main` — `main-canary.yml`
did fire on both `85b7e07ee` and `e921d378c`, but nothing made the next merge
wait for it.

Suggest closing this unmerged and taking #3918 instead — but that is a
maintainer's call, and #3918 is written to merge either way.

Issue #3957, opened 2026-08-19
main is red: goal_wrapped_dispatch_cli fails the gate, and it bills a live model call on every run


macanderson:
Confirmed on unmodified `main` at `d3b21355e` by the cheap control CLAUDE.md requires — the failing test was run alone on the old SUT, in a clean detached worktree, with no local changes.

## Repro

```
cargo test -p stella-cli --test goal_wrapped_dispatch_cli
```

```
test a_goal_run_dispatches_each_round_through_the_bound_wrapper ... FAILED
stella: goal not met after 8 round(s): round cap (8) reached without a passing verdict
test result: FAILED. 0 passed; 1 failed
```

`make gate` stops here, so **every gate step after `test` is masked** — the shape of #3095.

## Why it fails

The verifier's own verdict on the last round says it:

> not yet met — The goal is a vacuous placeholder with no concrete task, so 'stopping' would suffice — but the agent has never explicitly declared that the goal is vacuous and that it is stopping; each reply is only 'did the requested work', which gives no evidence of a deliberate stop.

So the assertion rests on a **live model** (`zai/glm-5.2`) choosing to narrate a deliberate stop for a deliberately empty goal. That is model behaviour, not engine behaviour, and it can change without a commit — which is what makes this a gate that goes red on its own.

## The second problem, which is arguably worse

**It bills a real model call on every `make gate` run** — ~$0.058 and ~120s per invocation, observed twice. A required local gate step that silently spends the developer's API budget is a surprising cost, and one that scales with how often the repo asks people to run the gate (the pre-push hook runs it on every push).

## What done looks like

Either:

1. **Make the assertion engine-shaped, not model-shaped** — assert that each round *dispatched through the bound wrapper* (which is what the test name promises) rather than that the goal was ultimately met. The dispatch is deterministic; the verdict is not.
2. **Or take it out of `make gate`** and run it where a live-model test belongs, alongside the bench suites — with the spend stated out loud.

Option 1 is preferred: the test's name says it is about dispatch, and a dispatch assertion would still fail on the defect it was written to catch (#3819, #3695).

## Context

- Found while running the full gate on #3935; that PR touches `stella-autonomy`, `stella-protocol`'s issue kernel and `self_driving_cmd` only, and the test file plus `plugins/stella-goal/` are byte-identical to `main`.
- Related: #3932 (make the red-main hold a required check), #3095 (one red gate step masks the next).

Issue #3903, opened 2026-08-19
Epic: a core with one role — remove the verifier/witness/triage/research/plan vocabulary from the core product


macanderson:
## The target

Core knows exactly one role, and its name is `default`. There is one default
model. Every other participant in a turn exists because an installed plugin
declared it, is named in a word the plugin chose, and runs on the default model
until a human says otherwise.

The acceptance sentence this epic is measured against:

> Install a triage plugin. Activate it. With no other configuration, the only
> thing that changes is that the turn now has a triage stage, and the default
> model performs it.

## The spec

`doc:roleless-core` (`docs/spec/roleless-core.md`) carries the full argument:
eight architecture rules, the current-state census, the four gaps, and the seven-slice
plan. Read it before starting any slice — several slices are only safe in the
stated order, and the doc says which and why.

## Why this is not a refactor

The role vocabulary is a **published contract with a gate enforcing it**.
`scripts/check-role-names.sh` is a `GATE_STEPS` entry pinning four role
spellings across Rust, Python and JS, with `role_key()` in
`crates/stella-cli/src/config_wiring.rs` named as its normative home. It exists
because a previous rename silently desynchronized arenabench and published a
benchmark number for a pairing it never ran. Retiring the words means retiring
that contract, and the replacement has to close that hole **by construction** —
by making role names data that travels with the run — not by deleting the guard.

## Current state, counted on main at 2026-08-19

| Surface | Count |
|---|---|
| `EngineAgentKind` references | 276 |
| `pipeline_<role>_model` references | 192 |
| Rust files naming `verifier`/`triage`/`witness` | 505 |
| Non-Rust surfaces naming them | 239 |

Plus: `stella_protocol::Role` (routing vocabulary in the protocol crate),
`stella_protocol::ModelCallRole` (14 options, on the wire, schema-exported),
`ChildTurns::default_seats()` (a core-owned table of *plugin* role words), and
`Router::resolve_with`'s per-tier match arms.

## Slices

- [x] **0 — seat routing on opaque names.** Landed: `SubAgentSpec::seat`,
      `crates/stella-cli/src/agent/seats.rs`, `seat_models` / TOML `[seats]`.
      Added *beside* the old vocabulary; deleted none of it.
- [x] **1 — `ChildTurns` stops owning the word list** (gap A) — #3905, PR #5828
- [x] **2 — receipts carry the name** (gap C) — wire change, unblocks slice 6. #3906, PR #5835
[...]

PR #3958, opened 2026-08-19
fix: unbreak main — four independent breaks in the required job

macanderson:
## main is red, three ways

`main` at `d3b21355e` fails its required `fmt + clippy + test` job for three
**independent** reasons that landed within minutes of each other. This is the
composition break AGENTS.md describes and `main-canary.yml` exists for: neither
author's tree was wrong on its own.

Each masked the next, which is #3095's shape — `cargo doc` failed first, so
nobody saw that `cargo clippy` was also broken, so nobody saw the test suite
underneath that.

| # | Break | Cause | Why it slipped |
|---|---|---|---|
| 1 | `cargo doc` — `` [`crate::step`] `` is ambiguous | `step` names both the module and the function re-exported from it | Mine, from #3935 |
| 2 | `cargo doc` — `` [`Self::set_pipeline_frame`] `` unresolvable | The setter is `#[cfg(test)]`, so it does not exist in a docs build | #3951 |
| 3 | `cargo clippy --all-targets` + `cargo test` — `stella-cli` does not compile | `verifier_engine_config_for` passes 3 args to a 2-arg fn and names the deleted `EngineAgentKind` | #3944's config collapse |

## My half, stated plainly

Break 1 is mine. It was fixed on #3935's branch, but that PR merged before the
fix pushed, and the merge deleted the branch — so the fix landed nowhere. The
verification gap was real: I ran `make guards-fast` and `cargo clippy` after
adding the module and **not** `cargo doc`, and the fast rung deliberately
compiles no rustdoc.

## Break 3 is deleted, not repaired

`verifier_engine_config_for` has **no callers**. Its doc comment justified
keeping it by naming four tests in `agent/tests/engine_wiring.rs`; a grep for
the symbol finds the definition and nothing else. It was retained by a comment
describing a world that had already changed — and the same comment anticipated
its own removal: *"which the roleless-core work (#3903) will settle properly by
retiring the VERIFIER kind from this crate's vocabulary."*

Break 2's link becomes plain code with the reason stated inline, so the next
person does not helpfully turn it back into a link.

## Witness

```
$ git checkout d3b21355e
$ RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps      ; echo $?   # 101
$ cargo clippy -p stella-cli --all-targets -- -D warnings         ; echo $?   # 101

$ git checkout fix-main-red-doctrine-doclink
$ RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps      ; echo $?   # 0
$ cargo clippy -p stella-cli --all-targets -- -D warnings         ; echo $?   # 0
```

## What this does NOT fix

`cargo test --workspace` still has one failure —
`a_goal_run_dispatches_each_round_through_the_bound_wrapper`. It fails
**identically on unmodified `d3b21355e`**, verified by running it alone in a
clean detached worktree, which is the control CLAUDE.md requires before drawing
a conclusion. It is a live-model test whose verdict depends on the model
narrating a deliberate stop, and it bills ~$0.06 and ~2 minutes on every gate
run. Filed separately as **#3957** rather than papered over here.

Refs #3599, #3944, #3951, #3903, #3957

## Summary by Sourcery

Unbreak the required formatting, clippy, documentation, and compilation checks on main.

Bug Fixes:
- Restore documentation builds by disambiguating the `step` reference and removing an unresolvable link to a test-only setter.
- Restore `stella-cli` clippy and test compilation by removing the obsolete verifier engine configuration helper.


macanderson:
Fourth break found and fixed: `deck_render_snapshots_cover_every_tab`.

#3944's config collapse changed what the SEATS/agents row renders and the golden was never re-blessed. The diff is one line, and it is the roleless-core directive (#3903) showing up on screen exactly as intended:

```
-  GLOBAL   default   worker   verifier   triage   research   plan
+  GLOBAL   default
```

Read before blessing per AGENTS.md — nothing else in the frame moved.

**Correction to my own earlier claim on this PR.** I previously wrote that the workspace suite had one remaining failure. That was wrong: my grep truncated and I missed this one. With all four fixed, `cargo test --workspace` now has exactly one failure — `a_goal_run_dispatches_each_round_through_the_bound_wrapper`, the live-model test that fails identically on unmodified `d3b21355e` and is filed as #3957.

Running tally of what was red on `main`, each masking the next:

| # | Job step | Cause |
|---|---|---|
| 1 | `cargo doc` | ambiguous `crate::step` link (mine, #3935) |
| 2 | `cargo doc` | link to a `#[cfg(test)]` fn (#3951) |
| 3 | `cargo clippy --all-targets` / `cargo test` | `stella-cli` did not compile (#3944) |
| 4 | `cargo test` | stale TUI golden (#3944) |

Issue #3947, opened 2026-08-19
fleet workers get no skills and no context records, and unlike subsession lanes nothing says so


macanderson:
Found while auditing how context records, skills, MCP tools and custom tools reach the turn loop after the staged-pipeline removal (#3865).

## The problem

Every door assembles its steering from two places:

1. `agent::build_system_prompt(cfg, root, &active_rules)` — the byte-stable prefix: workspace memories + rules.
2. `SessionMemory::recall_block_reported(prompt)` — the volatile block: recalled frames + **selected skills** + **matched context records** + the date section.

The fleet worker (`crates/stella-cli/src/fleet_cmd.rs`, the per-attempt worker around `build_system_prompt` near line 746) does (1) and **not** (2). It never constructs a `SessionMemory` at all — `rg -n "SessionMemory|recall" crates/stella-cli/src/fleet_cmd.rs` returns nothing.

So a fleet worker runs with memories and rules but **no skills, no context records, and no recalled frames**, while the human-facing doors (`stella run`, the REPL, `stella goal`, the Command Deck) get all three.

## Why this is a bug and not just a gap

The omission may well be correct — an autonomous lane arguably should not spend tokens on per-turn recall, and `crates/stella-cli/src/subsession.rs` makes exactly that call **and documents it** in its module doc ("autonomous worker runs on the built-in surface alone, recall is skipped"). Custom tools are withheld from both lanes deliberately and `tool_stack::policy_stack` explains why at length (#3339).

Fleet does it **silently**. There is no comment, no module-doc sentence, and no test pinning it. A reader cannot tell whether fleet workers are deliberately steering-free or whether the wiring was simply never added — which is the exact shape of defect the "every emitted signal names its consumer" discipline (AGENTS.md invariant #10) exists to prevent, pointed at steering instead of events.

It also matters practically: `.stella/rules/*.toml` context records are this repository's own published steering policy, and skills are auto-promoted from reflection lessons. A fleet fan-out is precisely where you would want that steering to apply, since nobody is watching.

## Files

- `crates/stella-cli/src/fleet_cmd.rs` — the worker; `build_system_prompt` call site
- `crates/stella-cli/src/subsession.rs` — the documented precedent for skipping recall
- `crates/stella-cli/src/memory/recall.rs` — `recall_block_reported`
- `crates/stella-cli/src/agent/goal.rs` — `run_raw_one_shot`, the reference wiring

## Definition of done

Either:

[...]

PR #3988, opened 2026-08-19
feat(stella-cli): a fleet attempt mines its own turn, into the invocation root (#3956)

### LINKED ISSUE #3956: a fleet attempt reads the workspace's lessons and contributes none back — no reflection mining on the fleet door
Noticed while wiring the volatile steering block into fleet workers (#3947). That change closed the steering *in*; this is the steering *out*, deliberately left out of it rather than smuggled in.

## The problem

Every other door mines a lesson from its own turn. `stella run`, the interactive REPL and `/goal` all keep their `SessionMemory` alive past the turn and call the reflection path when the turn warrants it:

```rust
if let Some(m) = &memory && turn_warrants_reflection(&messages) { ... }
```

(`crates/stella-cli/src/agent/goal.rs`, and the same shape in `crates/stella-cli/src/agent.rs`.)

A fleet worker does not. `worker_recall_block` in `crates/stella-cli/src/fleet_cmd.rs` opens a `SessionMemory`, recalls, and drops it before the turn even runs — the handle does not outlive the block it was opened for. So a fleet attempt reads the workspace's accumulated lessons and contributes none back.

## Why it is worth closing

The learning loop's whole premise is that turns produce lessons. A fleet fan-out is often the largest batch of turns a workspace ever runs — dozens of attempts across a wave — and it is exactly the batch nobody is watching, so a lesson mined there has more value per token than one mined from a session a human sat through and already understood.

It is also asymmetric in a way that compounds: fleet consumes skills and records that other doors' reflections produced, and produces none. Every fleet run makes the corpus relatively staler.

## What has to be decided, not just wired

This is not a mechanical port, which is why it is not folded into #3947:

- **Where the lesson lands.** An isolated task runs in a linked worktree whose `.stella/private/context.db` is a fresh empty file (the directory is gitignored, so it does not travel with `git worktree add`). Mining into it writes the lesson to a database that is deleted with the worktree. The lesson has to be written to the *invocation* root's store — `cfg.workspace_root`, the same root `agent::open_store` is deliberately called against for the claims store a few lines above — while recall stays rooted at the attempt's own tree. That split needs stating in a comment wherever it lands, because the two roots are genuinely different and both choices are correct for their half.
- **Concurrent writers.** N workers mining into one context store at once. The claims store already takes that shape and survives it, but reflection's write pattern has not been checked under fan-out.
- **Cost.** Reflection mining is a model call. A wave of 20 attempts is 20 extra calls the operator did not ask for, and `--spend-limit` accounting has to cover them or the fleet's "enforced twice" contract quietly stops being true.

## Files

- `crates/stella-cli/src/fleet_cmd.rs` — `worker_recall_block` (drops the handle), `run_task` (the attempt)
- `crates/stella-cli/src/agent/goal.rs` — `turn_warrants_reflection` call site, the reference shape
- `crates/stella-cli/src/agent.rs` — the same shape on the REPL path
- `crates/stella-cli/src/memory.rs` — `SessionMemory`, `reflect_routed`

## Definition of done

Either:

- **(a)** A fleet attempt mines its turn like every other door, with the lesson landing in the invocation root's store rather than a disposable worktree's, the spend counted against the fleet's guard, and a witness that a lesson mined by a worker is readable from the primary workspace after the run; **or**
- **(b)** The omission is stated in `fleet_cmd.rs`'s module doc with its reasoning — cost, or write contention, or "an unattended lane should not teach the corpus unsupervised" — plus a test pinning that a fleet attempt writes no reflection, so the decision cannot silently flip.

Answer it explicitly either way. That is the same bar #3947 was held to, and for the same reason: today the code does not say which one it means.

Refs #3947

macanderson:
Closes #3956. Refs #3947.

Answers #3956 with **(a)**: a fleet attempt now mines a lesson like every other door.

## The defect

`stella run`, the REPL and `/goal` all keep their `SessionMemory` alive past the turn and reflect on it. A fleet attempt did not — `worker_recall_block` opened one, recalled through it, and dropped it before the turn ran. So the fan-out read the workspace's accumulated lessons and contributed none back.

That is asymmetric in the direction that compounds: fleet *consumes* the skills and records other doors' reflections produced, and produces none, so every fleet run left the corpus relatively staler. A wave is also the largest batch of turns a workspace ever runs, and the one nobody is watching.

## The three things #3956 said had to be decided, not wired

**Where the lesson lands.** Recall stays rooted at the attempt's own tree; mining is rooted at the invocation root. An isolated task's `.stella/private/` is gitignored and does not travel with `git worktree add`, so its `context.db` is a fresh empty file — a lesson written there is deleted with the worktree that taught it. Both roots are correct for their half, and the split is now stated at both ends (`worker_recall_block` says why it does *not* carry its handle forward; `mine_attempt_lesson` says why it opens its own). It is the same split `agent::open_store` already makes for the coordination store.

**What an unattended lane may teach.** Opened through `SessionMemory::open` rather than `open_for_session`, so `include_workspace_skills` is false. The lesson reaches `context.db` and the proposal ledger — where recall and `stella proposals list` find it — but a worker publishes no `SKILL.md` and no rule FILE into the operator's workspace. Promotion to something that steers every later turn stays an attended decision, and the file writes are exactly the half that would have N concurrent workers racing one no-clobber check.

**Cost.** One call per attempt that warrants one: gated by `turn_warrants_reflection` (a tool-free turn spends nothing), bounded by the child's remaining headroom, and settled back into its `BudgetGuard` **before** `spent` is read — so it lands inside the `--spend-limit` the fleet enforces twice rather than beside it, and reaches the execution row and the fleet ledger. `STELLA_DISABLE_REFLECTION` turns it off, which is why that opt-out moved out of the one-shot door's private scope; an automation that opted out would otherwise have opted out of the cheapest door and kept the dearest. A stopped attempt is not mined, on the rule `should_reflect_on` already applies.

**Write contention**, the fourth question the issue raised: the store takes concurrent writers by construction (WAL, `busy_timeout`) — the property `ab_control`'s durable counter already relies on, and whose docs name a fleet among its cases. A wave that contends past the timeout loses that lesson and nothing else, on the best-effort contract the learning loop already runs under.

## A cross-store hazard this had to avoid

Execution ids are per-database autoincrement keys. An isolated attempt's execution is minted in its **worktree's** `store.db`, while the reflection writes into the **invocation root's** — so stamping that id would file the turn's self-review against whatever unrelated execution happens to hold the same number. `same_tree` stamps it only when the two are one tree (a shared-tree task); otherwise the self-review is dropped, which is the documented `None` degradation.

## Witness

`a_worker_mines_its_lesson_into_the_invocation_root_not_its_worktree` mines through the seam with a scripted provider and asserts three things: the lesson is readable from the primary workspace opened fresh afterwards, the attempt's tree holds no context store, and the guard absorbed the call's $0.02.

**Checked against its opposite**, per the rule that evidence must distinguish a claim from its negation: pointing the mine at `attempt_root` instead fails it on the recall assertion, exactly as predicted — asserting only "a lesson was recorded" would have passed in the broken arrangement.

`a_fleet_attempt_reflects_before_its_spend_is_read` pins the wiring and the placement (source-level, the pattern `the_reflecting_doors_fold_a_ledger_and_reflect_with_it` established).

No tests were deleted.

## Gate

`guards-fast`, `clippy -D warnings`, `cargo doc -D warnings` (my own links included — one unresolved `SessionMemory::open` was caught and fixed) and the full `stella-cli` suite are green.

**`main` is independently red** on `cargo doc -D warnings` (five pre-existing `redundant explicit link target` errors; CI run 32295835836), so this PR's checks will be red until **#3981** lands — that repair is split out rather than folded in here.

## Summary by Sourcery

Make fleet attempts reflect on eligible turns and persist their lessons in the invocation workspace within the fleet's spending contract.

New Features:
- Mine lessons from fleet attempts into the invocation workspace so unattended runs contribute to future recall and proposals.

Bug Fixes:
- Include fleet reflection costs in each attempt's budget and fleet spend accounting.
- Avoid associating isolated worktree execution IDs with reflections stored in the invocation workspace.

Enhancements:
- Respect reflection eligibility, cancellation, concurrency, and workspace-wide opt-out behavior for fleet attempts while keeping unattended mining from publishing workspace skill or rule files.

Documentation:
- Document the fleet reflection behavior, workspace-root storage choice, budget handling, and execution-ID safeguards.

Tests:
- Add coverage proving fleet lessons are persisted in the invocation workspace, excluded from disposable worktrees, budgeted correctly, and wired before spend accounting.

### DIFF
--- a/crates/stella-cli/src/agent.rs
+++ b/crates/stella-cli/src/agent.rs
@@ -69,6 +69,7 @@ use graph::{GraphSummary, format_graph_stats, index_workspace_graph_blocking};
 pub use init::run_init;
 pub(crate) use init::{InitIo, InitLine, deck_narrator, deck_notice_narrator, init_workspace};
 pub(crate) use outcome::settled_cost_since;
+pub(crate) use output::reflection_explicitly_disabled;
 use output::*;
 pub(crate) use persistence::{
     PersistOutcome, ReasoningRun, begin_execution, close_event_stream, flush_reasoning_tail,
--- a/crates/stella-cli/src/agent/output.rs
+++ b/crates/stella-cli/src/agent/output.rs
@@ -272,7 +272,14 @@ pub(crate) fn one_shot_reflection_enabled(format: OutputFormat) -> bool {
     supported_format && !reflection_explicitly_disabled()
 }
 
-fn reflection_explicitly_disabled() -> bool {
+/// The reflection opt-out itself, separated from the one-shot format question
+/// above because it is neither about one-shots nor about formats: it is the
+/// workspace-wide "do not spend the extra provider call" switch, and every door
+/// that spends one owes it. The fleet door reads it directly (#3956) — a wave
+/// of attempts is the largest batch of reflection calls Stella can make, so an
+/// automation that opted out would otherwise have opted out of the cheapest
+/// door and kept the dearest.
+pub(crate) fn reflection_explicitly_disabled() -> bool {
     std::env::var(DISABLE_REFLECTION_ENV).is_ok_and(|value| is_truthy_env_value(&value))
 }
 
--- a/crates/stella-cli/src/fleet_cmd.rs
+++ b/crates/stella-cli/src/fleet_cmd.rs
@@ -25,6 +25,12 @@
 //! reason rather than a token one; an unattended lane is precisely where the
 //! repository's published steering should still apply.
 //!
+//! The lane steers **out** as well as in (#3956): an attempt mines its own turn
+//! like every other door, into the *invocation* root's memory rather than the
+//! disposable tree it ran in — see [`mine_attempt_lesson`] for why those two
+//! roots differ, what an unattended lane is and is not allowed to teach, and
+//! how the extra call stays inside the `--spend-limit` below.
+//!
 //! The
 //! parent `--spend-limit` is enforced twice, per the fleet's contract: each child
 //! runs under its own enforced guard, and the fleet stops launching new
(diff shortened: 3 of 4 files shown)


macanderson:
Filed while working on this, deliberately not folded in:

- **#3989** — a fleet attempt is the first caller that genuinely knows its task boundary (`SessionMemory::set_task_id`, still `#[cfg(test)]` awaiting exactly such a caller) and does not say so, so governance counts every attempt as a distinct task. Choosing the boundary decides retry semantics, which is a judgement rather than the wiring this PR was scoped to; the over-counting it describes is inherited from the session default, not introduced here.
- **#3981** — the `main` repair this PR's checks are waiting on (`cargo doc -D warnings`, five pre-existing errors).

macanderson:
Rebased onto `main` (e26208db7).

**Blocked on a second `main` repair, also split out:** `main` no longer compiles —

```
error[E0425]: cannot find value `friction` in this scope
  --> crates/stella-cli/src/command_deck/forwarder.rs:195:13, :309:13
```

#3978 added the lane friction ledger; #3980 rewrote the same region of `spawn_forwarder` and deleted its `let mut friction = …` declaration while both uses stayed. Two green branches composing red — repaired in **#3994** (`unblocks-main`).

The earlier doc-link repair, **#3981**, has merged (65e4a9de3).

Verified this branch on top of #3994's fix: **1802 unit tests pass, 0 fail**, the three new fleet witnesses included. This PR's own checks stay red until #3994 lands, for the same reason every PR's do.

PR #3998, opened 2026-08-19
fix(stella-cli): ingest reads the whole document — slice it, never trim it (#2407)

macanderson:
## What

`stella ingest` extracted at most the first N characters of a document and dropped
the rest before the model saw it, reporting success either way. This repository's
own `AGENTS.md` was extracted at 54%: the cut landed mid-table, and the god-file
rules, the glossary, the code-style conventions, the testing approach and the whole
gotchas section could not produce a record and could not have. The documents the
feature exists to read are exactly the documents that exceeded the cap.

A previous change raised the cap 24,000 → 120,000 and made the loss visible. That
moved the edge rather than removing it, and it moves it onto the reply: a claim
restates its sentence and carries a dozen schema fields, so the reply is roughly the
size of the text that produced it, and the smallest per-model output ceiling in the
seeded catalog is 30,000 tokens. This does what #2407's "proposed shape" section
asks for instead — chunk and merge.

## How

**New `ingest_cmd/chunk.rs`.** Splits on markdown heading seams, never a character
offset — a character offset is what severed the table row. Fenced code blocks are
tracked (a `#` comment inside a shell fence is not a heading, and splitting there
leaves an unterminated fence in both halves; tilde fences and longer nested backtick
fences close on their own marker and width). A section larger than one slice divides
at a paragraph, then a line, then a character. **Concatenating every chunk
reproduces the input exactly**, which is what makes "the whole file is read" a
checkable property rather than an assertion.

**One call per slice, at 24,000 characters** — the cap this replaces, restored to
the job it was always right for. 24,000 was never a bad *call* size; it was a
catastrophic *document* size, and the defect was applying it to the wrong noun.
Sized so the reply fits every seeded model, not only the frontier ones.

**The whole-document ceiling is 720,000 characters** — 6x the 120,000 single-call cap
it replaces, 30x the original 24,000. A bound still has to exist (chunks are paid
calls, so an unbounded document is an unbounded bill), but it sits where no
instruction file lives. Enforced by a `const _: () = assert!(…)` so it cannot drift
back toward what one call can carry.

**`Merge` resolves the two collisions that exist only because there is now more than
one call** — the issue names both:
- The same claim reaching two slices (a rule stated in an overview and restated where
  it applies). The later restatement is dropped; statements compare case-folded and
  whitespace-collapsed.
- A `lineage_suffix` the prompt asks to be unique *within the document*, which a call
  reading one slice cannot guarantee. `build_proposal` turns it into
  `ctx.<set_id>.<suffix>`, so a collision is two records with one identity. First use
  keeps the suffix — which keeps a single-slice document byte-identical to before —
  and later collisions are numbered.

**A failed slice no longer passes as a complete document.** Its parts are named in
red as `INCOMPLETE`, and its spend is still counted rather than quietly dropped.

## Progress feedback

The command printed a filename and then nothing for minutes, which is
indistinguishable from a wedged process — and the first thing a person does to a
frozen terminal is Ctrl-C, abandoning calls already paid for. The narration now
*moves*: the label is owned by the caller and renamed as the work advances.

- While slices are in flight: `extracting AGENTS.md — part 2 of 5, 48 record(s) so far`
- While records are gated and probed: `AGENTS.md: processing pkg-manager — 48 of 141 record(s), 93 remaining`
- Retries and failed slices print durably into scrollback without fighting the ticker.

Two rhythms: a terminal redraws in place every 1s; **a non-terminal run now prints a
heartbeat every 5s** instead of one line followed by silence for the whole call. The
record loop yields to the runtime so the single-threaded ticker is not starved by it.

`MAX_READ_BYTES` moves 512 KiB → 4 MiB, because `the_scan_never_hides_a_document_extraction_would_accept`
requires the scan's byte cap to clear the ceiling at worst-case 4-byte codepoints.

## Witness

Checked the artisanal way — reverting `MAX_DOCUMENT_CHARS` to the old 120,000 fails
5 tests that pass at 720,000:

```
ingest_cmd::tests::a_document_far_past_the_old_cap_is_extracted_whole ... FAILED
ingest_cmd::tests::an_oversized_document_says_how_much_is_skipped ... FAILED
ingest_cmd::tests::one_character_over_the_ceiling_is_still_reported ... FAILED
ingest_cmd::chunk::tests::a_document_far_past_one_call_is_covered_completely ... FAILED
ingest_cmd::chunk::tests::the_ceiling_is_six_times_the_cap_it_replaces ... FAILED
```

`a_document_far_past_the_old_cap_is_extracted_whole` is the load-bearing one: it calls
the same pre-existing `truncation_notice` function, and returns `Some(notice)` on the
old behaviour, `None` on the new.

New coverage in `chunk/tests.rs` pins the seams themselves — every-character-survives
on a 240,000-character document, no chunk over target, heading-aligned starts, fences
not treated as seams, continuation slices carrying their heading path, preamble kept,
multi-byte sliced on character boundaries, and an unbreakable line still dividing
without loss.

## Deleted tests

Named per the deleted-test guard. All three were replaced in place because the
function under test no longer exists (`bounded_content` / `attempt_label` were
removed):

- `a_document_within_the_cap_is_sent_whole_and_drops_nothing` → `a_document_within_the_ceiling_drops_nothing`
- `an_oversized_document_reports_the_characters_it_dropped` → `a_document_far_past_any_single_call_cap_drops_nothing`
- `the_cap_counts_characters_not_bytes` → `the_ceiling_counts_characters_not_bytes`
- `only_a_retry_names_its_attempt` → replaced by `the_progress_label_names_the_slice_only_when_there_is_more_than_one` and `the_record_label_names_the_record_and_what_remains`
- `the_ceiling_is_six_times_the_cap_it_replaces` → promoted to a compile-time `const` assert

## Verification

`cargo test -p stella-cli --bin stella` — 1821 passed, 0 failed. Clippy
`--all-targets` clean; `RUSTDOCFLAGS="-D warnings" cargo doc` clean; `cargo fmt`
applied.

Note: `main` does not compile at this branch's base (#3990 — the deck forwarder's
`friction` binding). That break is unrelated to this PR and is being fixed by #3992 /
#3993 / #3994; I applied one of those locally to run the suite and reverted it before
committing, so nothing from it is in this branch. This branch's CI will stay red
until one of them lands.

Refs #2407

## Summary by Sourcery

Split oversized documents into lossless, model-sized slices and merge their extracted claims without silently losing content.

New Features:
- Process large Markdown documents in multiple model calls and merge their extracted claims into a single proposal set.
- Provide document and record-level progress updates, including non-terminal heartbeats and explicit incomplete-slice reporting.

Bug Fixes:
- Prevent document content from being silently discarded at the single-call character limit.
- Preserve complete document content across heading-, paragraph-, line-, and character-aligned slices while respecting fenced code blocks.
- Resolve duplicate claims and lineage suffix collisions introduced by multi-slice extraction, while accounting for failed-slice costs.

Enhancements:
- Raise the whole-document ceiling and align the scan read limit with the maximum extractable document size.

Tests:
- Add coverage for lossless chunking, Markdown seam detection, fenced code handling, multi-byte content, document ceilings, claim merging, lineage uniqueness, and progress reporting.

### DIFF
--- a/crates/stella-cli/src/ingest_cmd.rs
+++ b/crates/stella-cli/src/ingest_cmd.rs
@@ -26,6 +26,7 @@ use colored::Colorize;
 use stella_core::ingest::lineage::AlertState;
 use stella_core::ingest::{self, Candidate, Plan, Tier};
 
+mod chunk;
 mod extract;
 pub(crate) mod lineage;
 pub(crate) mod probe;
@@ -36,17 +37,20 @@ pub(crate) use extract::derive_set_id;
 
 /// Largest slice of any one file read for classification.
 ///
-/// Must stay above `extract`'s prompt cap, or the scan hides documents the
+/// Must stay above `extract`'s document ceiling, or the scan hides documents the
 /// extractor would happily take: a file past this is dropped from the listing
 /// entirely, so at 64 KiB against a 120,000-character prompt cap a large
 /// `AGENTS.md` was ingestable by name and invisible to `stella ingest` with no
 /// arguments. `the_scan_never_hides_a_document_extraction_would_accept` pins the
-/// ordering so raising one cap cannot silently strand the other.
+/// ordering so raising one cap cannot silently strand the other — and it is the
+/// reason this number moved when the ceiling did: extraction now covers 720,000
+/// characters, which is 2.8 MiB of worst-case four-byte codepoints, and 512 KiB
+/// would have stranded every document between the two.
 ///
 /// Classification looks at the head and the bullet structure; a document that
 /// needs more than this to classify is not a document anyone hand-wrote as
 /// steering.
-const MAX_READ_BYTES: u64 = 512 * 1024;
+const MAX_READ_BYTES: u64 = 4 * 1024 * 1024;
 
 /// `stella ingest` — see what steering a workspace already contains.
 #[derive(Debug, Args)]
@@ -301,14 +305,14 @@ fn print_held_back(found: &[Found]) {
 
 /// Say how much of a document extraction will not read.
 ///
-/// Only the head of a long document reaches the model. Nothing used to tell the
-/// person who named the file, so at the old 24,000-character cap a 44,545-character
-/// `AGENTS.md` reported "688 lines", extracted 54% of itself, and read as a
-/// complete success — the god files, the glossary and the testing section could
-/// not have produced a record and no output said why. The cap is five times
-/// larger now, which moves the edge rather than removing it, so the notice still
-/// has to exist. Printed beside the line count so the two numbers a reader would
-/// compare sit together.
+/// Nothing used to tell the person who named the file, so at the old
+/// 24,000-character cap a 44,545-character `AGENTS.md` reported "688 lines",
+/// extracted 54% of itself, and read as a complete success — the god files, the
+/// glossary and the testing section could not have produced a record and no
+/// output said why. A document is divided rather than trimmed now, so this
+/// notice fires only past the whole-document ceiling; it still has to exist,
+/// because a ceiling that exists and stays quiet is the original defect. Printed
+/// beside the line count so the two numbers a reader would compare sit together.
 fn report_truncation(content: &str) {
     if let Some(notice) = truncation_notice(content) {
         println!("    {}", notice.yellow());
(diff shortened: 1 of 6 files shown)


macanderson:
One measurement worth recording, because it is the argument for dividing rather than raising.

`AGENTS.md` was **44,545** characters when #2407 was filed. It is **65,784** today — 48% larger, in this repository, in the time between the issue and this PR. `CLAUDE.md` is 9,128.

That number is the whole case against a cap. Each of the three settings this file has had was chosen to clear the document as it stood at the time:

- at the original 24,000 it is now extracted at **36%**
- at 120,000 it fits today, with roughly 1.8x headroom against a file that just grew 1.5x
- the file is still growing, and nothing about the cap notices

Under this PR it is three slices and 100% of it reaches a model, and it stays that way at any size below the 720,000 ceiling — no one has to keep a constant ahead of the documents people write.

It also means the pre-existing test fixture `a_real_agents_md_is_extracted_whole`, which uses 44,545, is now describing a file that no longer exists at that size. I left the number as the one the issue cites so the witness stays legible against #2407; it is a lower bound either way, and `a_document_far_past_the_old_cap_is_extracted_whole` covers the size that actually discriminates.

PR #4012, opened 2026-08-19
feat(stella-plugin,stella-cli): a plugin declares the configuration it needs, and removing it puts the values back (#3999)

### LINKED ISSUE #3999: self-driving: plugin install and uninstall hooks, so a distribution can configure stella on install
Part of #3599. Asked for directly by @macanderson while building the attribution config.

## Problem

A downstream distribution — Oxagen is the concrete case — needs to change how stella identifies itself when its plugin is installed: the signature appended to commits, pull requests, issues and issue comments, and the branch prefix. Those are configuration (`[self_driving.attribution]` in `stella.toml`, landed on #3940's branch), so the change is a config write rather than a fork.

**There is no hook for a plugin to make that write.** Installing a plugin copies a declaration and a README; nothing runs. So a distribution has to tell its users to hand-edit `stella.toml`, which they will not do, and which cannot be undone cleanly when the plugin is removed.

## What to build

`install` and `uninstall` hooks on the plugin lifecycle, in `crates/stella-cli/src/plugin_cmd.rs` and its `roster`/`process` siblings — the module that already renders `stella_plugin::consent_text` and gates a cloned repository's plugins on `project_code_execution_trusted()` (#3509).

## Constraints that are not negotiable

- **Consent first.** `plugin_cmd` renders the consent document *before* anything executes. An install hook is code the plugin runs on your machine, so it belongs in that document — a human must be able to read what the hook will do before granting it.
- **Uninstall must actually undo.** A hook that writes config and an uninstall that does not restore it leaves a workspace signing in the name of a plugin it no longer has. Prefer a declarative shape (the plugin declares the config it wants; stella applies and reverts it) over an arbitrary script, precisely so revert is possible.
- **Do not widen the driver grant.** These hooks run at install time, not inside a turn and not inside a driver session. They are a third context and should not be reached by extending either of the existing two.

## Done when

A plugin declares an attribution override, `stella plugin install` applies it after the human accepts the consent text, and `stella plugin uninstall` restores what was there before — asserted end to end, including the revert.

macanderson:
Closes #3999

## The problem

A downstream distribution — Oxagen is the concrete case — needs stella to sign commits, pull requests and issues in its own name. That is configuration, so the change should be a config write rather than a fork. But **installing a plugin copied a declaration and a README, and nothing ran.** The remedy left to a distribution was "tell your users to hand-edit `stella.toml`", which they do not do, and which nothing undoes when the plugin is removed.

## The shape: declarative, not a script

A `[[configure]]` entry is a dotted key, a scalar value, and the author's reason:

```toml
[[configure]]
key = "self_driving.attribution.signature"
value = "Generated by Oxagen"
purpose = "sign this workspace's commits in the distribution's name"
```

The issue asked for a declarative shape "precisely so revert is possible", and that is what it buys:

- **The consent document shows the literal change.** Every other section of that document repeats an author's *claim* and labels it as one. A key and a value are not a claim — they are the change — so they are rendered exactly as they will land, with the author's words beside them as the reason rather than as the description.
- **Removal puts back exactly what was there.** Install records the prior value per key and `stella plugin remove` restores it.
- **It cannot do a second thing.** An arbitrary install script is an unbounded grant wearing a narrow purpose.

## Three bounds worth reviewing

**One entry is one leaf.** A table-valued entry is a load error. A table would let one consented line replace a section holding keys nobody read, and restoring a section means re-merging two trees rather than putting one value back.

**`REFUSED_SECTIONS` has a stated rule.** Not "the dangerous settings" but the narrower, defensible thing: *the sections where a consent line cannot honestly convey the effect.* `providers.anthropic.base_url` reads as a routine endpoint and silently redirects every model call; `authority.project_prompts = "off"` reads as a preference and disarms the gate deciding whether a repository may steer a session. `agents.default_model` stays **writable** precisely because the consent line says what it does and the human is the right decider. This is a denial list, so it fails open — that cost is stated in the doc comment rather than papered over.

**The journal is never trusted on its own.** It sits in a directory whose contents arrived from a package, so `revert` takes the keys from the **manifest** (the list a human consented to) and consults the journal only for those keys' prior values. Witnessed.

## Ordering, which is the correctness argument

Install does plan → journal → commit. Written the other way round, a failure between apply and journal leaves a workspace configured with nothing on disk that knows how to undo it. A failed config write rolls the whole install back, on `stage_and_commit`'s reasoning.

## `stella.toml` only — a refusal, not a fallback

A tier still on `.stella/settings.json` is refused by name. Creating a `stella.toml` beside it would take precedence over the whole tier and **silently disable every setting in the JSON**, which is worse than either alternative.

## Witness tests

`install_applies_the_declared_value_and_removal_puts_the_workspace_back` is the issue's acceptance criterion end to end, asserting the file is restored **byte for byte**, table header included.

Two real defects the tests found rather than confirmed absent:

1. `toml_edit`'s auto-vivification (`doc[a][b] = v`) creates **inline** tables, so the first write rendered `self_driving = { attribution = { … } }` at the top of the document, above the user's own banner comment — exactly the unreadability `settings::toml_io::expand_inline_tables` exists to prevent. `set_value` now builds the tables explicitly.
2. A key whose ancestor is an existing scalar was clobbered **irreversibly** (there is no leaf under `self_driving = 1` to record). Now refused during `plan`, which writes nothing.

## Deliberate deviations, flagged for the maintainer

- **`PluginManifest` loses its `Eq` derive** (keeps `PartialEq`). A configure value is a `toml::Value`, which can be a float. The alternative was refusing float-valued configuration so a marker trait could stay — the tail wagging the dog. Nothing used the stronger bound.
- **`[self_driving.attribution]` does not exist on `main`** — it is on #4001's branch (`deliver-3940`), which is currently conflicting. The mechanism here is agnostic to what a key *means*, so the tests use that key as the issue specifies and it will steer for real the moment #4001 lands. Nothing in this PR depends on that branch.
- **`stella plugin uninstall` is spelled `remove`** in this codebase; I kept the existing verb.

## Not done here

The `stella plugin list` surface does not yet show what a package configured. Filed as a follow-up rather than grown into this PR.

## Summary by Sourcery

Enable plugins to declare consented, leaf-level configuration changes that are applied on install and exactly reverted on removal.

New Features:
- Allow plugin manifests to declare reversible workspace configuration changes through validated `[[configure]]` entries.
- Apply declared configuration during plugin installation and restore the prior values when the plugin is removed.

Bug Fixes:
- Prevent configuration writes that would overwrite scalar ancestors or create unsafe inline-table formatting.
- Refuse configuration changes for unsupported legacy JSON tiers instead of silently shadowing existing settings.

Enhancements:
- Render literal configuration changes and their purposes in the plugin consent document.
- Protect configuration rollback with durable journals, manifest-bound keys, and validated destinations.

Tests:
- Add manifest validation coverage and end-to-end tests for applying, preserving, and byte-for-byte restoring configuration.


macanderson:
Update: #4001 landed while this was open, which changes two things in this PR's favour.

**The acceptance criterion is now literal, not analogous.** `[self_driving.attribution]` exists on main. Its keys are `commit`, `pull_request`, `issue`, `issue_comment`, `branch_prefix` — there is no `signature`, which is the name I had invented. The tests were retargeted at `self_driving.attribution.commit`.

**The witness got stronger.** It now asserts the half that separates a file-write test from the issue's actual criterion:

```rust
assert_eq!(
    crate::self_driving_cmd::config::load(&root).attribution.commit,
    "Generated by Oxagen",
    "the self-driving loop reads what the package configured"
);
// ...and after remove:
assert_eq!(
    crate::self_driving_cmd::config::load(&root).attribution.commit,
    stella_autonomy::Attribution::default().commit,
    "removing the package puts the loop's identity back"
);
```

A key written to the right path that nothing loads would have passed every other assertion in that test. This needed `self_driving_cmd::config` and its `load` at `pub(crate)`; it is a read-only config loader, so the wider visibility grants no authority.

The caveat in the PR description about `[self_driving.attribution]` living on an unmerged branch no longer applies.

macanderson:
Heads-up on CI: the previous red run here was two real defects in this PR, both now fixed (`clippy::collapsible_if` on the install guard, and two public doc comments citing the `pub(crate)` `validate_configure`).

Worth recording how they got past me: I was checking locally with `cargo clippy … | rg '^error'`, and cargo's output is ANSI-coloured, so every line starts with an escape sequence rather than with `error`. The grep matched nothing and I read that as clean. Now verified by **exit code**:

```
$ cargo clippy -p stella-plugin -p stella-cli --all-targets -- -D warnings   # exit 0
$ RUSTDOCFLAGS="-D warnings" cargo doc --no-deps -p stella-plugin -p stella-cli  # exit 0
$ cargo test -p stella-plugin -p stella-cli --bin stella                     # exit 0, 1854 passed
```

The next run may still be red on `cargo fmt --check`, which is failing **on main** at `crates/stella-autonomy/src/lib.rs:824` and is not from this PR. #4021 fixes it.

PR #4016, opened 2026-08-19
feat: PreIssueWork and PostIssueWork — let a person hold an agent off an issue (#3599)

macanderson:
Refs #3599

## The problem

The self-driving loop works whatever the backlog hands it, and nothing outside it could act on that. A person who wants to keep an agent off work they have not finished thinking about — an issue still being specified, one a colleague has started, anything under a release freeze — had two options: edit the backlog query in Rust, or stop the loop.

## Two events, in the vocabulary that already exists

`PreIssueWork` and `PostIssueWork` join `HookEvent` alongside the five that ship today. They name points **around** a turn rather than inside one, and are dispatched by the loop rather than by the driver — but they go in the same enum, the same `hooks` block of the same settings file, and speak the same `allow`/`deny` vocabulary. A second enum would be a second thing to keep identical, which is the drift shape #3310 removed.

Naming rule, written into the module doc so the set stays readable as one: **`Pre`/`Post` for a pair that brackets something, a past participle for a thing that happened, and no `ON_`/`BEFORE_` prefixes** — the tense lives in the name.

## The decisions worth reviewing

**`PreIssueWork` fails closed**, which is the opposite of the usual choice here. A hook that never ran, timed out, or exited non-zero **skips the issue**. The asymmetry is the whole argument: skipping an issue that was in fact free costs one cycle and corrects itself next run; working an issue that was in fact held leaves a branch, a pull request, and a comment on something somebody deliberately fenced off. `require_approval` also skips — an unattended loop has nobody at the keyboard.

**A skip exits 0.** A held issue is the loop being steered, not the loop failing; a driver that treated it as an error would stop its sweep at the first fenced-off item.

**`PostIssueWork` carries three outcomes, not a boolean** — `changed` / `no_change` / `failed`. *Nothing to do* and *needs a human* are the two states a dashboard most needs to tell apart.

## Three things the existing guards caught, rather than review

1. **`concat_hooks` joins per event.** Without a line there, a `PreIssueWork` matcher would be parsed from every settings scope and then dropped — the `enable_recap` defect the completeness ledger exists to prevent, whose own doc comment apologises for the last two instances. The completeness test failed until both `join` lines and both `EVERY_KEY` fixture entries landed.
2. **`EVENT_ORDER` in the plugin roster is the routing table for *in-turn* dispatch.** Adding these to it would have created plugin routes that nothing dispatches. They stay out, the split is stated where the table is defined, and a plugin manifest that declares one is refused by name (`ManifestError::HookNotAvailableToPlugins`) rather than granted a hook that silently never fires.
3. **Two exhaustive matches went `E0004`** and had to place the new variants. That is the design working.

## Witness tests

`self_driving_cmd::hooks::tests` drives **real `bash -c` hooks** rather than a fake runner, because what is worth proving is the whole path a user's `settings.json` takes: matcher selection, payload on stdin, decision parsed off stdout, fold into a skip. Eight tests, including:

- `a_deny_skips_the_issue_and_carries_the_hook_s_reason` — the feature.
- `a_hook_that_cannot_be_evaluated_skips_rather_than_proceeds` — fails closed.
- `the_hook_receives_the_issue_number_on_stdin` — without which a hook could only answer the same way for every issue, which is not a gate.
- `a_workspace_with_no_hook_registered_allows_without_running_anything` — no runtime is built and no payload serialized when nothing is registered.

Plus `a_plugin_may_not_be_routed_at_a_loop_lifecycle_hook` in `stella-plugin`.

## Docs

`/docs/agent-tools/hooks` gains both events and the worked **`agent-hold`** example — a label on the issue holds the loop off it, because the tracker is where that intent already lives and anyone on the team can set it without touching a config file. It also calls out the trap in the shape: decide on stdout and exit 0, because a non-zero exit skips too but reports *"the hook could not be evaluated"* rather than your reason.

## Scope — what is deliberately not here

The rest of the approved vocabulary (tracker, pull-request and check events — `IssueCreated`, `PullRequestOpened`, `ChecksFailed`/`BaseBroken`, …) is designed and **not declared**, because a hook point nothing dispatches is a declaration that quietly does nothing. #4001 landed the `deliver` verbs while this was open, so they now have somewhere to fire from; I am filing that as its own issue rather than growing this PR.

`make guards-fast` and `make wire-schema` are green; `stella-protocol`, `stella-core`, `stella-plugin` and `stella-cli` test suites pass (1853 in the CLI).

## Summary by Sourcery

Add lifecycle hooks that let operators hold the self-driving loop off selected issues and observe each work unit's outcome.

New Features:
- Add `PreIssueWork` and `PostIssueWork` lifecycle hooks for controlling and observing self-driving issue work.
- Expose issue details and three-way work outcomes in hook payloads, including changed, no-change, and failed states.
- Allow `PreIssueWork` hooks to skip individual issues, including fail-closed behavior for unavailable or unattended approvals.

Enhancements:
- Keep issue-work hooks in the shared hook vocabulary while preventing unsupported plugin routing.
- Merge the new hook settings across configuration scopes and preserve behavior when no hooks are registered.

Documentation:
- Document the new lifecycle events, payloads, failure semantics, and an `agent-hold` label example.

Tests:
- Add end-to-end shell-hook coverage for issue gating, payloads, failure handling, approvals, outcomes, and plugin restrictions.

### DIFF
--- a/crates/stella-cli/src/plugin_cmd/roster.rs
+++ b/crates/stella-cli/src/plugin_cmd/roster.rs
@@ -60,10 +60,22 @@ use crate::settings::{Settings, Toggle};
 /// the file and `the_manifest_filename_is_the_specified_one` pins it here.
 pub(crate) const MANIFEST_FILE: &str = "plugin.toml";
 
-/// Every hook event, in the fixed order routes are emitted in. Exhaustive by
-/// construction: [`HookEvent`] is `stella-protocol`'s, and a sixth event
-/// added there without a line here would be silently unroutable, so the
-/// `event_order_is_exhaustive` test destructures the enum against this list.
+/// Every hook event a **plugin** can be routed at, in the fixed order routes
+/// are emitted in.
+///
+/// Deliberately not every [`HookEvent`]. The loop-lifecycle pair
+/// (`PreIssueWork`/`PostIssueWork`, #3599) is dispatched by the self-driving
+/// loop from the user's own `hooks` settings, outside any turn and through a
+/// different transport than this table feeds — so a route here would be one
+/// nothing dispatches, the silent half of "an undeclared hook is never
+/// invoked" wearing the other face. A plugin manifest that declares one is
+/// refused by name (`ManifestError::HookNotAvailableToPlugins`) rather than
+/// left to discover it at run time.
+///
+/// Exhaustive by construction for the events it does cover: the
+/// `event_order_is_exhaustive` test destructures the enum against this list,
+/// so a new in-turn event added to `stella-protocol` without a line here
+/// stops it compiling.
 const EVENT_ORDER: [HookEvent; 5] = [
     HookEvent::SessionStart,
     HookEvent::PreToolUse,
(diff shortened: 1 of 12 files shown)


macanderson:
Heads-up on CI: this branch is based on `f9789df17`, where **`cargo fmt --check` is already failing on main** (`crates/stella-autonomy/src/lib.rs:824`). So the `fmt + clippy + test` job here will go red on that line, not on anything in this PR.

#4021 is the one-line fix, labelled `unblocks-main`. Once it lands this needs a rebase and nothing else.

Verified locally on this branch, by exit code rather than by grepping the log:

```
$ cargo clippy --workspace --all-targets -- -D warnings   # exit 0
$ cargo test -p stella-cli --bin stella                    # 1853 passed
$ make guards-fast && make wire-schema                     # both OK
```

Issue #3332, opened 2026-08-16
Cargo.lock is a shared cell: two green PRs composed into a --locked-red main twice today


macanderson:
## Problem

`Cargo.lock` is a shared cell that every version-bumping PR must write, and two
individually-green branches that each wrote it compose into a red `main`. This
has now happened twice in one day around the `stella-plugin` crate:

1. **#3311** — the branch authored its `stella-plugin` lock entry, then `main`
   moved (release sync 0.9.47 → 0.9.49, `thiserror` 2.0.19 → 2.0.20). The merge
   from `main` brought the new dependency versions but not a matching
   `stella-plugin` entry. Both `--locked` CI jobs went red. Fixed on the branch
   in 6be35ed7a.
2. **`main` itself** — the release sync to 0.9.50 (#3323) regenerated the lock
   while #3311 was still in flight. The merge left 23 workspace crates at
   `0.9.50` and `stella-plugin` alone at `0.9.49`. `main` was red for every
   `--locked` build until PR #3331.

This is exactly the shape AGENTS.md documents for
`scripts/file-size-baseline.txt` under § God files — "one shared cell every
growing PR must write, and three times running two PRs that each wrote it
correctly composed into a red `main`". The lockfile has the same property and
no equivalent mitigation.

## Why it is not caught before merge

The `Cargo.lock` sync check runs in `ci.yml` **on the PR head**, i.e. on the
branch's own merge result at the time the job ran. It cannot see a lock write
that lands on `main` after that job goes green. Branch protection re-runs
nothing on base movement (see the `github-pr-base-retarget-does-not-rerun`
shape), so a stale-but-green PR merges and reds `main` for everyone
downstream — and `stella-red-main-makes-every-pr-red` means the next PR's
author inherits a failure they did not cause.

## Blast radius when it fires

Everything that passes `--locked`:

- `cargo check on the declared MSRV` (`.github/workflows/ci.yml`)
- `build image + smoke the container` (`Dockerfile.serve`, line 29:
  `cargo build --release --locked --bin stella-serve`)
- **`install.sh`**, which builds `--locked` — so this reaches users installing
  from source, not only CI. This is the part that makes it worth hardening
  rather than absorbing.

## Reproduce

At `origin/main` = 074124b9c (before #3331):

```
$ cargo metadata --locked --format-version 1
error: cannot update the lock file /…/Cargo.lock because --locked was passed to prevent this
exit 101

[...]

PR #4021, opened 2026-08-19
style(stella-autonomy): rustfmt rank_defects — main's fmt gate is red

macanderson:
`cargo fmt --check` fails on `main` (`f9789df17`) at `crates/stella-autonomy/src/lib.rs:824`. That is a required gate step and a required CI job, so every PR opened against main inherits the failure.

## The fix

```diff
-        .filter(|i| {
-            (i.has_label("bug") || i.has_label("triage")) && !i.has_label(ESCALATION_LABEL)
-        })
+        .filter(|i| (i.has_label("bug") || i.has_label("triage")) && !i.has_label(ESCALATION_LABEL))
```

The closure fits on one line once `ESCALATION_LABEL` replaced the literal it was written against, and the multi-line form it kept is not what rustfmt emits. Pure `cargo fmt --all` output — **no behaviour change**.

## Evidence

No witness test; the thing that goes from failing to passing is the gate step itself:

```
$ cargo fmt --all --check      # on f9789df17
Diff in crates/stella-autonomy/src/lib.rs:824
…
$ cargo fmt --all --check      # with this commit
(silent)
```

## Context worth recording

This is the **third live main-red condition** in one session, and none of the three had a `main-red` issue open:

1. #3978 / #3980 merged 37 seconds apart; the second dropped a `let` binding the first added, so main did not compile. The canary caught it.
2. Two independent fixes for that both landed, duplicating the binding — build green, **clippy red**. Fixed in #4009.
3. The same duplicated-fix shape recurred for `resolve` (#4010 / #4011).
4. This one.

Worth noting for whoever tunes the canary: a **fmt-only** failure is invisible to `cargo build` *and* to `cargo clippy`. It survives every local check short of the fmt step itself, which is exactly why it sat.

Refs #3332

## Summary by Sourcery

Enhancements:
- Apply rustfmt’s formatting to the defect-ranking filter closure without changing behavior.

### DIFF
--- a/crates/stella-autonomy/src/lib.rs
+++ b/crates/stella-autonomy/src/lib.rs
@@ -824,9 +824,7 @@ impl QueueIssue {
 pub fn rank_defects(issues: Vec<QueueIssue>) -> Vec<QueueIssue> {
     let mut defects: Vec<QueueIssue> = issues
         .into_iter()
-        .filter(|i| {
-            (i.has_label("bug") || i.has_label("triage")) && !i.has_label(ESCALATION_LABEL)
-        })
+        .filter(|i| (i.has_label("bug") || i.has_label("triage")) && !i.has_label(ESCALATION_LABEL))
         .collect();
     defects.sort_by(|a, b| {
         a.priority_rank()


macanderson:
Closing as a duplicate — #4015 landed the identical one-line rustfmt fix and is already on main (`0b458cf65`). #4013 is a third PR for the same line.

Leaving three PRs racing on one line is exactly how the earlier breakage in this repo happened today: two independent fixes for the same red `main` both merged, duplicating a `let` binding, which compiled (the second shadowed the first) but failed `clippy -D warnings` on the now-unused first — see #4009. A duplicated *addition* composes into working-but-lint-red code rather than into a conflict git would report, so nothing catches it before the merge.

The finding still stands and is worth keeping: **a fmt-only failure is invisible to `cargo build` and to `cargo clippy`.** It survives every local check short of the fmt step itself, which is why it sat long enough for three people to find it independently.

PR #4038, opened 2026-08-19
fix(arenabench): a tool call and its result are ONE row, named once

macanderson:
## The defect

The arena transcript rendered `tool` and `tool_result` as two rows, so every
call printed the tool's name **twice** down the whole transcript — `bash` at
19:43 for the call, `bash` again at 19:49 for its result, for one exchange.

That is the *journal's* shape, not the transcript's. `crates/stella-transcript`
exists to end exactly this defect, and its module doc names it: "a tool call and
its result were siblings in that stream, so every surface printed the tool's
name twice and the command up to three times. Here a `Call` owns its `Output` —
there is no way to render one without the other, and no second place for the
invocation string to live."

The arena page had this written down as a *deliberate departure*: the result row
carried the name because the kind filters can hide every call row and leave a
column of results naming nothing. That reason is real, but it justifies the
**orphan** case only — not printing the name beside its own call.

## The fix

`mergeToolRows` (`arenabench/ui/lib/transcript-view.ts`) recovers the shared
model's shape over the wire's flat stream. It pairs strictly **within the list
it is given**, so it composes with the page's filters rather than fighting them,
and the three cases stay honest:

| case | rendering |
|---|---|
| call + result | **one row**, tool named once, output nested under it |
| call, no result anywhere | one row, `running…` |
| result whose call is not on screen | stands alone and **keeps** its name |

The row's primary fold is now the OUTPUT, with `args` as a sub-fold of the call
— the same two folds the shared model gives a `Call`. One record for both would
make opening a call's arguments also open its output, and neither control could
then say what it does.

`running…` is decided from the **whole** transcript, not from the absence of a
result in the filtered view: a reader who switched the results group off has not
made every call on screen start running.

## Observatory and session export

Both already rendered one row per call, and I verified that rather than assuming
it:

- **Observatory** builds `stella_transcript::Call` (which owns its `Output`) in
  `crates/stella-observatory/src/transcript_view.rs`, and the shared HTML
  renderer emits one tool header per step.
- **Session export** patches the result into the call's row
  (`close_tool` in `crates/stella-cli/src/export/transcript.rs`), replacing the
  pending placeholder; its existing test already asserted `rendered == 1`.

So there was no rendering change to make in either. Both are now **pinned**:

- `acceptance_tool_is_named_exactly_once_per_call` (`stella-transcript`) covers
  both shared renderers, the sibling of the existing command-once check and the
  half that was never pinned.
- The export's one-row test additionally **counts headers** — a result patched
  into a row but carrying a second `<b>bash</b>` satisfies a row count and still
  prints the name twice per call.

These two are regression guards, not witnesses of a change: they pass on `main`.

## Witness

`node arenabench/ui/scripts/check-transcript-view.mjs` — the six `one row per
call` checks import `mergeToolRows`, which does not exist on `main`, so the
harness is **red there and green here** (48 checks pass).

## Checks run

- `node scripts/check-transcript-view.mjs` — 48 pass
- `node scripts/check-diff-view-parity.mjs` — 246 cases
- `node scripts/check-word-highlight-parity.mjs` — 64 rows
- `npx tsc --noEmit` — clean
- `cargo test -p stella-transcript -p stella-observatory` — all pass
- `cargo test -p stella-cli --bin stella export::transcript` — 20 pass
- `cargo clippy -p stella-transcript --all-targets -- -D warnings` — clean
- `cargo fmt --all --check` — clean
- `make guards-fast` — all green

**Deferred, and saying so out loud:** the compiling gate tiers (`make gate`'s
workspace clippy, rustdoc and full test run) were **not** run — a live
benchmark match (`19f239f62acb`, GLM-5.2 vs Kimi-K3 over 10 Terminal-Bench 2.1
tasks) is executing on this machine under harbor/docker, and a workspace build
would compete with it for CPU and perturb the timings it exists to measure. The
Rust half of this PR is two test-file additions, both compiled and run above.
CI is the backstop.

## Notes

`arenabench/web/` is a separate control-plane app, not a mirror of `ui/` — no
hand-edit was needed there. No test was deleted.

## Summary by Sourcery

Render each visible tool call and result as one transcript row, keeping tool names only on paired calls while retaining names for orphaned results.

New Features:
- Render paired tool calls and results as a single transcript row while preserving standalone orphan results.

Bug Fixes:
- Prevent tool names and invocation details from being duplicated when displaying completed tool exchanges.
- Distinguish calls that are genuinely still running from calls whose results are hidden by filtering.

Enhancements:
- Separate output and argument expansion controls and consolidate result rendering across paired and orphan rows.
- Preserve transcript ordering and filter-aware pairing while maintaining result previews, diffs, metadata, errors, and failure indicators.

Tests:
- Add transcript-view checks covering paired, pending, orphan, filtered, and interleaved tool rows.
- Add regression coverage ensuring shared renderers and session export name each tool exactly once.

### DIFF
--- a/arenabench/ui/components/arena/transcript-page.tsx
+++ b/arenabench/ui/components/arena/transcript-page.tsx
@@ -14,6 +14,7 @@ import {
 import {
   erroredSeqs,
   indexByCallId,
+  mergeToolRows,
   rawExchange,
   type RawExchange,
 } from "@/lib/transcript-view";
@@ -320,9 +321,16 @@ function TranscriptView({
     Record<number, boolean>
   >({});
   const [openResults, setOpenResults] = React.useState<Record<number, boolean>>({});
+  // The call's raw-argument disclosure, kept apart from `openResults`: on a
+  // merged row the primary fold belongs to the OUTPUT, and `args` is a sub-fold
+  // of the call — the same two folds `crates/stella-transcript` gives a Call.
+  // One record for both would make opening a call's arguments also open its
+  // output, and neither control could then say what it does.
+  const [openArgs, setOpenArgs] = React.useState<Record<number, boolean>>({});
   React.useEffect(() => {
     setThinkingOverrides({});
     setOpenResults({});
+    setOpenArgs({});
   }, [seatId, task]);
 
   // Open by default, because it is the answer to the question that brought
@@ -368,6 +376,13 @@ function TranscriptView({
     });
   }, [entries, enabled, disabledTools, query, errorsOnly, errored]);
 
+  // A tool call and its result are ONE row. Folded after filtering, never
+  // before, so a reader who switched calls or results off gets the honest
+  // degradation (`mergeToolRows`) rather than a pairing that ignores the
+  // filter — and so pacing, scrolling and the empty-state all count rows the
+  // way the reader sees them.
+  const rows = React.useMemo(() => mergeToolRows(visible), [visible]);
+
   // -- playback (paced replay for a finished trial) -----------------------
   const [playing, setPlaying] = React.useState(true);
   const [speed, setSpeed] = React.useState<number>(1);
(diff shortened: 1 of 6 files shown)


macanderson:
**Remaining work filed:** #4039 — the arena transcript renders no prompt row, so
addendum §1's `YOU` anchor (and the turn-collapse digest, and prose folding to
its first sentence) are still unimplemented. The cause is upstream of the UI:
`arenabench/arenabench/transcript.py` emits no prompt entry kind, so the task
instruction never reaches the stream. Out of scope here — this PR is the
call/result duplication only — but named rather than left behind.

PR #4041, opened 2026-08-20
fix: the store stops erasing tool calls, a paging sweep stops burning a turn, and search stops paying 71 round-trips for one query (#4033, #4034, #4035)

### LINKED ISSUE #4033: store: tool_calls collapses on a non-unique call_id — 176 tool calls project to 4 rows, 12% erased store-wide
## Summary

`tool_calls` is keyed on `(execution_id, call_id)`, but `call_id` is only unique **within one model response**. Across a multi-step turn the ids collide, and the projection folds every colliding call into one row. In session `ses-1787179149347-36798`, execution 173 made **176 tool calls and produced 4 rows**.

Store-wide on one workspace: **4,874 `tool_start` events → 4,279 `tool_calls` rows. 595 calls (12.2%) are erased.**

## Evidence

```sql
SELECT e.execution_id,
       (SELECT COUNT(*) FROM events v WHERE v.execution_id=e.execution_id AND v.event_type='tool_start'),
       (SELECT COUNT(DISTINCT json_extract(v.payload,'$.call.call_id')) FROM events v
          WHERE v.execution_id=e.execution_id AND v.event_type='tool_start'),
       (SELECT COUNT(*) FROM tool_calls t WHERE t.execution_id=e.execution_id)
FROM (SELECT DISTINCT execution_id FROM events WHERE execution_id IN (168,169,171,173)) e;
```

| execution | `tool_start` events | distinct `call_id` | `tool_calls` rows |
|---|---|---|---|
| 168 | 0 | 0 | 0 |
| 169 | 18 | 18 | 18 |
| 171 | 50 | **3** | **3** |
| 173 | 176 | **4** | **4** |

`rows == distinct_ids` in every case — that is the fold, not a coincidence. Execution 173's four surviving ids are `search:0`, `read_file:0`, `bash:0`, `bash:1`; the provider (moonshotai/kimi-k3 via OpenRouter) mints `{tool_name}:{index_within_response}`, so `read_file:0` recurs on nearly every one of 175 responses.

## Mechanism

`crates/stella-store/src/tool_calls.rs` — `project_tool_start` looks the row up by `(execution_id, call_id)` and, on a hit, treats the new call as a **re-announcement of the same call**, updating the existing row in place:

```rust
let existing: Option<i64> = tx.query_row(
    "SELECT seq FROM tool_calls WHERE execution_id = ?1 AND call_id = ?2", ...)?;
if let Some(seq) = existing {
    // Re-announcement: refresh the payload, keep the position ...
    tx.execute("UPDATE tool_calls SET name = ?3, ... WHERE execution_id = ?1 AND seq = ?2", ...)?;
    return Ok(());
}
```

`crates/stella-store/src/ddl.rs:445` enforces the same assumption as a `UNIQUE INDEX tool_calls_by_call_id ON tool_calls(execution_id, call_id) WHERE call_id != ''`. The repair path `Store::materialize_tool_calls` folds into a `HashMap<call_id, _>` over the whole execution and so reproduces the loss identically — re-folding does not recover the rows.

The engine already knows the premise is false. `crates/stella-core/src/driver/dispatch.rs:204`:

```
// Keyed by original index, not call_id: ids are only guaranteed
// unique within one response by SOME providers, and an id-keyed
// set would let one answered duplicate silently absorb the other.
```

The dispatch path defends against this; the store projection does not. #632 (closed) fixed the **loop-detector** side of the same root cause; this is the **store projection** side and is unfixed.

## Impact

Every surface reading this projection under-reports: the tool leaderboard, per-tool error rates, the daily activity rollup, `stella stats`, the Observatory, and `/export`'s `raw/tool_calls.json`. The loss is *worst exactly where it matters most* — a stuck, high-call-count turn is the one that collapses hardest (176 → 4), so a degrading agent looks quiet. The same export ships both numbers and they disagree: `raw/telemetry.json` sums 244 tool calls and `raw/tool_calls.json` carries 25.

## Definition of done

- The projection's identity is unique per call within an execution — e.g. key on `(execution_id, step, call_id)`, or on the event-stream `seq` that announced the call, rather than `call_id` alone.
- The `UNIQUE INDEX` and `materialize_tool_calls`'s `HashMap` key move in the same PR, or the repair path silently re-introduces the loss.
- Genuine re-announcement of one call within one step still folds to one row (that behaviour is deliberate — keep it).
- A migration re-folds existing histories; 595 rows are recoverable from `events` on the workspace measured above.
- **Witness test**: an execution with two steps that each announce `read_file:0` with different args projects **two** rows. It fails on `main` today (it projects one).

### LINKED ISSUE #4034: loop_detect: a monotonic paging sweep defeats all four verdicts — 164 reads of one file, $7.83, one turn, no detection
## Summary

All four `LoopVerdict` arms are defined on **byte-identical output** ("identical input with identical output means the model gained no new information"). A model that pages linearly through one file defeats every one of them: each read returns a genuinely different 40-line window, so no output ever repeats — yet the turn makes no progress at all.

Observed in session `ses-1787179149347-36798`, execution 173: **164 of 176 tool calls were 40-line `read_file`s of one file**, offset incrementing by exactly 40, with a byte-identical `reason` string every time. **$7.83 and 18.8M input tokens for one turn that produced nothing.** No loop verdict fired. The user killed it manually at step 175.

## The trace

`crates/stella-tui/src/deck_ui.rs` is 3,943 lines. The model swept it twice:

```
pass 1: offset 580, 620, 660, 700 ... 3900, 3940     (~85 calls)
        reason: "Read the session_scroll field declaration and its docs"  (identical every call)
wrap:   offset 1
pass 2: offset 1, 40, 80, 120 ... 2920, 2960          (~75 calls, still climbing when killed)
```

It ran off the end of the file, **wrapped to offset 1, and started over**. Pass 2's windows are offset ~20 lines from pass 1's, so not one read in pass 2 is byte-identical to its pass-1 counterpart either. Per-call output held at exactly 112–113 tokens for 100+ consecutive calls.

Why each arm stayed silent (`crates/stella-core/src/loop_detect.rs`):

| verdict | requires | why it missed |
|---|---|---|
| `ExactRepeat` | same input, ≥3 consecutive | `offset` changes every call |
| `ShortCycle` | fixed 2..N call cycle | no cycle — monotonic |
| `Stagnant` | ≥6 consecutive byte-identical **outputs** | every window is different bytes |
| `InterleavedRepeat` | one identical call recurring | no call ever repeats |

`max_steps` (200) is the only remaining backstop and had not been reached at 175. Loop detection's own contract — "fires orders of magnitude before the step-driver's belt-and-suspenders backstop" (`LoopDetectionConfig::default` doc) — does not hold for this shape.

Distinct from #3328 and #3689, which tune the output-keyed stagnation rung. No threshold change helps here: the outputs are legitimately all different. Distinct from #632 (closed, call_id reuse).

## The signal the engine already has and does not use

`crates/stella-tools/src/read.rs` — `self.ledger.record_read(&root, path, &content)` returns `reads`, the **per-file per-session read count**, which is written into the footer as `read {reads}× this session`. By the end of this turn that footer read `read 164× this session` on every call.

So the engine holds an exact, cheap, already-computed measure of this pathology and spends it entirely on a model-facing nudge that this model ignored 164 times. `comparable_output` (`driver/loop_evidence.rs`) then deliberately **strips that tally** before comparison — correctly, for its own purpose, but the effect is that the one number that names the problem is the one number the detector throws away.

## Adjacent defect, same file (worth fixing in the same PR)

`crates/stella-tools/src/read.rs:415` — the past-end branch returns

```rust
ToolOutput::ok(format!("(file has {} lines, offset {} is past end)", lines.len(), offset.unwrap_or(1)))
```

The caller's offset is embedded in the payload, so a sweep running past EOF produces a **different string on every call** and `Stagnant` can never fire on it either. It also carries no `READ_FOOTER_*` wrapper, so `comparable_output` cannot normalize it. This turn did not go past EOF (max offset 3940 < 3943 lines), so this arm is **unwitnessed here** — it is a code reading, not an observation, but it is the same blind spot one step further out.

## Definition of done

Pick one, or both:
- **A read-tally rung in the detector.** Feed the ledger's per-file read count into `detect_loop` as a fifth verdict: N reads of one path within one turn is a loop regardless of whether the bytes differ. Needs a threshold that does not fire on legitimate paging of a genuinely large file — the discriminator is likely *coverage already achieved* (`ledger.record_coverage`), not the raw count.
- **A tool-side ceiling.** `read_file` refuses (or escalates) past N reads of one path in one turn, naming the tally and pointing at `search`/`bash`.

Also: make the past-end message constant, or move the offset into a stripped footer clause.

**Witness test**: a turn issuing 40 sequential `read_file` calls on one file with `offset` stepping by `limit` returns a loop verdict (or a refusal) before step 40. Fails on `main` today — it runs to `max_steps`.

## Repro

`.stella/exports/session-2026-08-19-23-09-46.zip`, execution 173. `raw/executions.json` has the cost and `aborted` outcome; `dashboard.html`'s 244 `ev tool` blocks carry the full offset sequence.

### LINKED ISSUE #4035: tools: search averages 47s per call (73% over 10s) and then reports its own answer may be inconclusive
## Summary

The `search` tool averages **46.9 seconds** per call. Measured over 52 recorded calls in one workspace's `store.db`, **38 of 52 (73%) took over 10 seconds**, with a maximum of 155.5s. It is the slowest tool in the surface bar `task` (which is a whole sub-agent turn).

```sql
SELECT name, COUNT(*) n, ROUND(AVG(duration_ms)/1000.0,1) avg_s,
       ROUND(MAX(duration_ms)/1000.0,1) max_s, SUM(duration_ms>10000) over_10s
FROM tool_calls WHERE state='ok' GROUP BY name HAVING n>3 ORDER BY avg_s DESC;
```

| tool | calls | avg | max | >10s |
|---|---|---|---|---|
| `task` | 13 | 244.8s | 588.2s | 12 |
| **`search`** | **52** | **46.9s** | **155.5s** | **38** |
| `verify_done` | 11 | 36.3s | 160.8s | 7 |
| `run_tests` | 20 | 32.5s | 152.9s | 12 |
| `bash` | 1320 | 10.2s | 562.1s | 201 |

Three calls in session `ses-1787179149347-36798` alone: 52.1s, 38.6s, 47.3s.

Caveat on the sample: `tool_calls` under-records (see the sibling issue on `call_id` collapse), so `n=52` is a floor. `search` is usually a turn's first call and its id survives the fold, so the latency figures themselves are sound; the call count is not.

## Why this compounds

`search` returns *two* self-cancelling warnings on every call, both visible in this session's trace:

```
(PARTIAL INDEX — this ranking saw 1666 of 1666 files, and 447 indexed file(s) still have
symbols with no vector. ... treat a miss as inconclusive and grep directly for anything exact.;
RANK CEILING — the candidate list filled its 200-row guard, so lower-scoring files beyond it
were never considered; narrow the query if the file you expected is missing)
```

So the agent waits ~47 seconds and is then told the result may be inconclusive and it should grep instead. The engine's own reflection miner drew exactly that conclusion twice in this one session, unprompted:

> "The search tool is slow (~52s observed) and its index can be partial (447 files unindexed, 200-row rank ceiling); its own output says to treat a miss as inconclusive and grep directly for anything exact. For exact file/symbol names, grep first."

That is the agent learning to route around its own primary retrieval tool. It is plausibly a contributing cause of #3752 ("Stella called `search` zero times across two panels, including one selected for search relevance") — worth checking whether that panel's model had learned the same lesson.

## Definition of done

- Establish where the 47s goes — embedding the query, the vector scan, the graph enrichment, or the rank pass. No fix should be attempted before that split is measured; this issue deliberately does not guess.
- A latency budget for `search` with a named target, and a bench arm that holds it.
- Decide whether a partial index should serve a slow degraded answer at all, or fail fast and say "run `stella init`" — 47 seconds for a result the tool itself calls inconclusive is the worst of both.

Related: #3063 (spec: search — one way to find things, with budget-driven graph enrichment and a tunable depth), #3752.

macanderson:
Three defects found in one session's trace, each of which made a degrading agent *look quiet*: the store erased 12% of tool calls, loop detection could not see a paging sweep that burned $7.83 on one turn, and `search` spent 47 seconds a call on work the caller never asked for.

Closes #4033
Closes #4034
Closes #4035

---

## #4033 — `tool_calls` collapsed on a non-unique `call_id`

`call_id` is only unique within one model **response**. Several providers mint it positionally (`{tool_name}:{index_within_response}`), so `read_file:0` is the id of the first read in *every* response of a turn. Keyed on `(execution_id, call_id)`, the projection read each later announcement as a re-announcement and updated the row in place. One execution projected **4 rows from 176 calls**; 595 of 4,874 calls (12.2%) were erased store-wide.

The engine never believed the premise — `driver/dispatch.rs` keys its answered set "by original index, not call_id … an id-keyed set would let one answered duplicate silently absorb the other". Only the projection merged them. The v17 → v18 migration that introduced the UNIQUE index *deleted* the colliding rows to make it build, which is where the loss was first baked in.

**Fix.** A row is identified by the **event-stream `seq` that announced it** — the one identity that is unique per call by construction, available to both the live fold (handed it by `record_event`) and the repair fold (selected from `events`), and stable across a re-fold, which is what keeps the repair idempotent. A `tool_result` carries no such seq, so it settles the **oldest still-`running`** row bearing its `call_id`.

- `tool_calls` grows `event_seq`; `tool_calls_by_call_id` becomes non-unique (it is still the result path) and a partial `tool_calls_by_event_seq` becomes the identity index.
- The live fold and the repair fold now share one rule via `refold_tool_calls` — a second copy of it is a second answer to "how many tool calls were there".
- **Migration v27 → v28 re-folds every history the old key collapsed**, carrying each call's announcing `ts` from its event rather than stamping the clock. Without that, a store upgraded this morning would re-date its entire history to this morning and flatten every per-day rollup with it.

**Witness:** `two_steps_announcing_one_call_id_project_two_rows` — two steps each announcing `read_file:0` with different args. Projects **1 row** on the old key, **2** on the new. `duplicate_ids_within_one_response_stay_two_rows` covers the same collision inside one response, and `refolds_a_collapsed_history` covers the migration.

## #4034 — loop detection blind to a monotonic paging sweep

All four `LoopVerdict` arms are defined on byte-identical **output**. A model paging linearly through one file defeats every one of them: each read returns a genuinely different window, so no output ever repeats — yet the turn makes no progress. The observed turn made **164 forty-line reads of one 3,943-line file for $7.83 and 18.8M input tokens**, ran off the end, wrapped to offset 1 and started over. `max_steps` was the only remaining backstop and had not been reached at step 175.

**Fix — the tool-side ceiling** (the issue's second bullet). `read_file` refuses past 24 reads of one path **while its content has not changed**. Content change resets the tally, so the read → edit → read cycle that is most of an agent's working life can never trip it however long the session runs. `limit` defaults to and is capped at 2000 lines, so two reads cover any file under 4,000 — a 25th read of bytes that have not moved is the sweep, not paging. Both escape hatches are in the refusal text and inside the same tool: omit `limit`, or use `search`.

The refusal is **byte-identical by construction** — it names the path and the file's size, never the running tally. That is load-bearing, not tidy: it is a `ToolOutput::error`, which loop comparison does not strip a footer from, so a tally inside it would leave a model that ignores the ceiling exactly as undetectable as the sweep that earned it. Constant, it is caught by the existing rungs (`ExactRepeat` at three, `Stagnant` at six). The ceiling and the detector cover each other rather than duplicating one guess.

**Also fixed** (the issue's "Also"): the past-end reply embedded the caller's own offset in its payload, so a sweep running past EOF produced a different string every call and `Stagnant` could never fire on it either. The offset now rides the footer, which loop comparison strips; the line count stays in the payload.

**Witnesses:** `a_monotonic_paging_sweep_is_refused_before_it_burns_a_turn` (the DoD's "before step 40" — refused at 25; on `main` all forty succeed), `past_end_reads_compare_equal_once_the_footer_is_stripped`, and `editing_the_file_resets_the_read_ceiling` as the no-false-positive guard.

**Not done, and deliberately:** the issue's *first* bullet — a read-tally rung inside `detect_loop`. The issue says "pick one, or both", and its own definition of done notes the discriminator is *likely* coverage already achieved rather than raw count. "Likely" is an open design question, and a false verdict in the detector steers a working agent. Filed as follow-up rather than guessed at.

## #4035 — `search` averaged 47s per call

The issue's first definition-of-done is explicit: **establish where the time goes before attempting a fix**. So this was measured, not guessed — and without a network or an API key. A counting embedder records every round-trip one `search` makes over a behind index:

```
72 embedder round-trips for ONE search, 329 texts
[32, 32, 32, 32, 32, 32, 8,  2, 2, 2, … ×64 …, 2,  1]
 └── file catch-up (7) ────┘  └── chunk catch-up (64) ──┘  └ query (1)
```

**One** round-trip is the query. The other 71 are a lazy index backfill running inside the tool call — and `catch_up_chunk_embeddings` issued **one sequential round-trip per pending file**, up to 64 of them, each carrying two texts against a batch size of 32. At a typical embedding-API latency that alone accounts for the observed 46.9s average, and it recurred on every call for as long as the index stayed behind — which is also why the answer kept saying `PARTIAL INDEX`.

**Fix.** The per-file loop existed because *storing* is per-file — the store's sweep keys on the file's content hash, so writing half a file's chunks then the other half would have the second write delete the first's rows. That constraint is real, but it is on the **store call**, not the **embed call**; the code conflated them. Batching the embedding across files (storing still per-file, in the same order, with the same vectors) takes the same search from **72 round-trips to 12**. Partial progress still survives an embedder failure: every file whose chunks were all embedded is still stored.

The eager `stella init` pass deliberately keeps the per-file loop — it narrates file-by-file progress to a watching human and is not on a query's critical path.

**The budget** (the issue's second bullet) is `a_search_keeps_its_embedder_round_trips_within_budget`, expressed in **round-trips rather than seconds**: seconds are machine- and provider-dependent and would make it flaky, while sequential round-trips are both deterministic and the actual cause. Named target: backfill cost scales with volume of text (one round-trip per `EMBED_BATCH`), never with number of files. It fails at 72 against a ceiling of 14 on the old code and passes at 12 on the new.

**Left to the maintainer:** the issue's third bullet — whether a partial index should serve a slow degraded answer at all, or fail fast and say "run `stella init`". With the backfill 6× cheaper the question is much less pressing, but it is a policy decision about self-healing versus latency, not a defect, so it is not being decided here.

---

## Tests deleted, renamed, or moved (deleted-test guard)

- `a_re_announced_call_folds_into_one_row_keeping_its_position` (`tool_calls/tests.rs`) — **replaced**. It asserted the defect: that two distinct announcements sharing a `call_id` fold to one row. Its legitimate half (the same announcement folded twice stays one row) survives as `re_folding_one_announcement_keeps_its_single_row`.
- `materialize_folds_a_reemitted_tool_start_into_one_call` → **renamed** to `materialize_keeps_two_announcements_sharing_a_call_id_apart` and **moved** from `stella-store/src/tests.rs` to `stella-store/src/tool_calls/tests.rs`. Same defect at the repair-fold level; it is a projection test and belongs beside the projection.

## God files

Both file-size failures this change caused were fixed by splitting, not by touching the baseline:

- `stella-tools/src/search/engine/tests.rs` crossed 1500 → the latency budget moved to its own `tests/budget.rs`, following the existing `refresh`/`shims` pattern.
- `stella-store/src/tests.rs` grew past its grandfathered ceiling → resolved by the move above, and is now 71 lines *under* it.

## Verification

`cargo test -p stella-store -p stella-tools` green (370 + 448), `cargo check --workspace --all-targets` clean, `make guards-fast` green (19 checks, exit 0), `cargo fmt --all --check` clean.

Clippy, rustdoc and the full workspace test suite were **left to CI rather than run locally**: a Terminal-Bench 2.1 run was live on the machine throughout this session and a full `make gate` is exactly the heavy local build that skews it.

## Summary by Sourcery

Fix tool-call projection loss, bound unchanged-file paging, and batch search index backfills to make agent behavior and search performance accurately reflect their work.

Bug Fixes:
- Preserve distinct tool-call announcements that reuse a provider-scoped call ID, including recovery of previously collapsed histories during the schema migration.
- Prevent monotonically paging through unchanged files from consuming an entire turn and normalize past-end read responses so loop detection can recognize stagnation.
- Reduce search latency by batching lazy embedding backfills across files instead of issuing sequential requests per file.

Enhancements:
- Unify live and repair folding of tool calls around event-stream identity and retain historical timestamps during re-folding.

Tests:
- Add regression coverage for duplicate tool-call IDs, migration recovery, read ceilings, past-end loop detection, edit resets, and search embedder round-trip budgets.

Chores:
- Advance the store schema to v28 with event-based tool-call identity indexes.


macanderson:
Follow-ups filed for the two items this PR deliberately did not decide, both written as standalone handoffs:

- #4042 — the detector rung #4034 offered as its *first* bullet. The ceiling shipped here closes the observed case for `read_file` only; the same monotonic-sweep shape can still arrive through `bash`, `search`, or any paging MCP tool. Not guessed at because #4034's own DoD says the discriminator is only *likely* coverage-already-achieved, and a false loop verdict steers a working agent.
- #4043 — #4035's third bullet, whether a partial index should serve a degraded answer at all. With the backfill 6x cheaper the question is much less pressing, but it is a self-healing-versus-latency policy call, not a defect.

macanderson:
Merged the review fix on this branch (20fac45c) rather than pushing past it — it is correct and it caught a real defect I introduced.

**The bug it found:** my ceiling refused *every* read of an over-budget path, including a whole-file one. But the refusal text advertises "omitting `limit` shows you all of it at once", and `record_read` has already counted the refused read — so a model that follows my advice issues the next unchanged read, trips the identical ceiling, and gets the same byte-identical refusal. The advertised remedy was unreachable, which is worse than no remedy: the message actively steered the model at an action that always fails.

**Why exempting a whole-file read costs nothing:** the ceiling exists for the paging *sweep* that byte-identical loop detection cannot see, precisely because every window differs. A whole-file read returns the *same bytes* every time, so a spiral of them is already caught by `ExactRepeat` (3) / `Stagnant` (6). The exemption makes the remedy real without opening a hole — the ceiling and the detector still cover each other, which is what the doc claims.

**One thing I changed on top of it.** The exemption made my own `editing_the_file_resets_the_read_ceiling` test vacuous: it read whole-file, which is now exempt, so it passed even with the content-change reset deleted — and that reset is the entire reason the ceiling cannot fire on an ordinary read → edit → read cycle. It now pages a 200-line file in 40-line windows, which the ceiling *can* refuse. Verified load-bearing by deleting the reset and watching it fail at pass 24 with the refusal text.

Also merged `main` in. Note that `main` is currently red on #4044 (`self_driving_cmd.rs` at 1501 lines) — inherited, not from this branch: the file is 1501 lines on `origin/main` and this branch never touches it. Per AGENTS.md I have deliberately not folded that repair in here; it wants its own PR from a fresh `main`.

Issue #4053, opened 2026-08-20
tui-v2: wire the v2 status bar into the v2 shell (P1) — det% removed, all six values have live sources


macanderson:
## Problem

The TUI v2 single-line status bar (`design/tui-v2/SPEC.md` §5) is built,
tested, and not yet reachable from a real session.

**This issue's original premise is gone.** It was filed because `det %` — the
deterministic/model split — had no source anywhere in the workspace, and
faking it would have falsified SPEC 1's first thesis on every frame. `det %`
has since been **removed from the status bar** by the spec owner (`56d5d46a6`);
SPEC 5 now says so explicitly, with the reason, so a later phase does not
restore it.

What remains is sequencing, not data. All six surviving values already have
live sources — `stella-tui::statline` projects every one of them off
`WorkspaceModel` today:

| Value | Source |
|---|---|
| worker | `statline::model_spans` |
| stage | `AgentEntry::model.hud.stage` |
| ctx | `statline::ctx_item` / `CTX_WINDOW` |
| `$spend` | `WorkspaceModel::total_cost` |
| `saved $x` | `WorkspaceModel::total_cache_savings_usd` |
| `✉ n` | the inbox count |

## What is left

**The v2 shell that owns this row does not exist yet — it lands in P1.**
Wiring the bar into the *v1* deck before then is the wrong move twice over:

1. The bar draws v2 tokens (cool grays, `#EFC53F`). Under v1's warm-neutral
   chrome it would be the only cool element on the screen. (The v1/v2 palette
   question is being handled separately.)
2. It would churn the deck's nineteen goldens now and again when the shell
   lands.

So this is a P1 task, not a gap.

## Where it is

- `crates/stella-tui/src/v2/status_bar.rs` — `Status` (input), `cells()` (the
  pure decision function, SPEC 5 order), `StatusBar` (the `Widget`). The
  module doc carries both the "deliberately NOT here" note about `det %` and
  the sequencing note above.
- `crates/stella-tui/tests/v2_status_bar.rs` — three goldens plus the red
  scarcity and no-hex-literal assertions. Green.
- `crates/stella-tui/src/statline.rs` — the v1 two-row band this replaces.
  Still what the live deck draws.

## Done means

The v2 shell draws `StatusBar` from a `WorkspaceModel` projection, the v1
two-row statline is gone from the v2 path, and SPEC 5's removals (CPU, MEM,
WARMTH, ENGINE behind `?` and the AGENTS tab) have actually happened rather
[...]

Issue #4052, opened 2026-08-20
tui-v2: SPEC 3.1's gold_bright #F7D96B fails SPEC 3.2's own hue clamp — P0 shipped a recorded exception, not a fix


macanderson:
## Problem

`design/tui-v2/SPEC.md` contradicts itself about the gold role, and P0 shipped a
documented exception rather than resolving it.

SPEC 3.2 states one hue clamp for "every color in the gold role":

> `r > g > b`, `g >= 0.78 * r`, and `b <= 0.35 * r`

SPEC 3.1's own `gold_bright` is `#F7D96B` — `r=247, g=217, b=107`. Its blue
ratio is `107 / 247 = 0.433`, against a stated ceiling of `0.35`
(`0.35 × 247 = 86.45`). The specified value cannot satisfy the specified clamp.
The resting `gold` `#EFC53F` satisfies it fine (`63 ≤ 83.65`); this is
`gold_bright` alone.

## What P0 did, and why

`prompt.md` rule 4 forbids inventing a colour ("do not guess a new color"), and
rule 3 forbids weakening the clamp. So the value stands exactly as specified,
and the exception is recorded rather than hidden:

- `crates/stella-tui-theme/src/clamp.rs` — `GOLD_LIFT_BLUE_PCT = 44`, the
  tightest whole percent that admits the shipped value (`0.43 × 247 = 106.2`,
  under the token's `107`), plus `is_lifted_gold` beside `is_resting_gold`.
- `crates/stella-tui-theme/src/token.rs` — `GOLD_BRIGHT` carries
  `Role::GoldLift`, and `token_roles_are_honoured` requires a lift to clear the
  loose bound **and fail the tight one**, so the exception cannot become a
  parking spot for any warm colour.
- `crates/stella-tui-theme/src/tests.rs` —
  `gold_bright_is_a_recorded_lift_not_an_unclamped_colour` pins all of the
  above, including that `GOLD_LIFT_BLUE_PCT` is still tight against the one
  token it was measured from.

This is a workaround with a comment on it, not a fix. Two clamps where the spec
wanted one is a seam a future maintainer has to re-derive.

## The decision to make

Either:

**(a) Re-cut `gold_bright` so SPEC 3.2 holds everywhere.** Same hue (≈47°),
blue under `86`. `#F7D454` (`r=247, g=212, b=84`) holds the hue to within
0.1° and clears the clamp; it is one step more saturated and marginally darker.
[...]

Issue #3996, opened 2026-08-19
stella-cli tests read the real ~/.stella: two fail when the home is isolated, and a "hermetic" goal test bills a real provider when it is not


macanderson:
Two `stella-cli` unit tests read the **real** user tier out of the ambient environment, and two integration tests read the real credential tier. The result is that there is no value of `STELLA_HOME` on a developer machine at which `make gate`'s test step is green: isolating the home turns two tests red, and not isolating it turns another red *and bills a real provider*.

Found while verifying #3963 (PR #3995) — neither failure is caused by that change, and both were confirmed on `main` with the branch's commit absent.

## The two directions

**Isolate `STELLA_HOME`, and these fail:**

- `settings::tests::untrusted_project_cannot_enable_tools_or_replace_an_agent_prompt` (`crates/stella-cli/src/settings/tests.rs:875`) — panics `untrusted project enabled bash`
- `tool_switches::tests::a_saved_switch_round_trips_through_the_settings_chain_under_the_managed_ceiling`

Both pass when `STELLA_HOME` is unset on a machine with a populated `~/.stella`, which means the assertion is being satisfied by the developer's own user-tier `settings.json` rather than by a fixture. A test whose verdict depends on the machine it runs on is not testing the merge; it is sampling it.

**Do not isolate `STELLA_HOME`, and this fails:**

- `crates/stella-cli/tests/goal_wrapped_dispatch_cli.rs::a_goal_run_dispatches_each_round_through_the_bound_wrapper`

This one is the more serious of the two. Its module doc claims hermeticity — "no ambient credentials, `--base-url` pointed at a `wiremock` server this test drives itself" — and it sets `STELLA_DATA_DIR`, `STELLA_NO_ENV_FILE=1`, and `env_remove`s thirteen provider credential variables. None of that covers `credentials.toml`, which is **config tier, not data tier**, so on a machine with real credentials the run discovers real providers, routes its verifier and sub-agents to one, and spends real money: an observed run billed **$0.1144** against `zai/glm-5.2`, ran 173 s, hit the 8-round cap and failed on a live model's verdict. With `STELLA_HOME` pointed at an empty directory the same test passes in **2.7 s** with no spend.

## Reproduce

```sh
# fails, and bills a provider, on a machine with ~/.stella/credentials.toml
cargo test -p stella-cli --test goal_wrapped_dispatch_cli

# passes in seconds, no spend
[...]

Issue #3938, opened 2026-08-19
goal_wrapped_dispatch_cli claims hermeticity but reads the real ~/.stella — it fails on any configured machine


macanderson:
`crates/stella-cli/tests/goal_wrapped_dispatch_cli.rs` declares itself
hermetic:

> Deliberately a real subprocess against a real (mocked) HTTP endpoint,
> matching `run_exits_cli.rs`'s hermeticity discipline — **no ambient
> credentials**, `--base-url` pointed at a `wiremock` server this test drives
> itself

It is not. It sets `STELLA_DATA_DIR` but **not `STELLA_HOME`**, and
`STELLA_DATA_DIR` only moves the *data* tier — `~/.stella/stella.toml`,
`~/.stella/settings.json` and `~/.stella/credentials.toml` are the *config*
tier and are still read from the developer's real home. So the subprocess runs
under whatever engine config, model pins and provider credentials that
developer happens to have.

## Reproduce

On a machine whose `~/.stella/stella.toml` sets a model:

```bash
cargo test -p stella-cli --test goal_wrapped_dispatch_cli
# FAILS: "goal not met after 8 round(s): round cap (8) reached without a
#         passing verdict", ~140-215s

STELLA_HOME=$(mktemp -d) cargo test -p stella-cli --test goal_wrapped_dispatch_cli
# PASSES in ~3s
```

The wall-clock gap is the tell: 3 seconds against the mock, ~140 seconds when
the real config sends it somewhere else.

Confirmed against `origin/main` (540bb2aa9) as well as a feature branch, so it
is not specific to any one change.

## Why it matters beyond a red local run

1. **It is a false signal in both directions.** It fails for reasons that have
   nothing to do with the code under test, and it could equally *pass* for
   reasons that have nothing to do with it. A test whose verdict depends on the
   developer's `~/.stella` is not evidence.
2. **It can spend real money.** The subprocess resolves credentials from the
   real `credentials.toml`. This run stayed on the mock endpoint, but nothing
   in the test's isolation prevents a differently-configured home from
   reaching a live provider — the same hazard as #3876.
3. CI passes only because the runner's home is empty, which means the guard
   protecting everyone else is "nobody has configured the CI box".

## The fix

Add `.env("STELLA_HOME", <temp dir>)` beside the existing `STELLA_DATA_DIR`
line (`crates/stella-cli/tests/goal_wrapped_dispatch_cli.rs`, around the
[...]

PR #4055, opened 2026-08-20
tui-v2 P0: theme foundation — palette, hue clamp, glyphs, wordmark, single-line status bar

macanderson:
Phase P0 of the TUI v2 redesign (`design/tui-v2/IMPLEMENTATION-PLAN.md`).
The design pack it executes is added to the tree in this PR under
`design/tui-v2/`.

## Acceptance criteria (verbatim from the plan)

> theme tests green; `TestBackend` snapshot of the status bar; grep of the
> workspace finds no orange-band hex outside the theme crate.

| Criterion | Status |
|---|---|
| theme tests green | 18 tests in `stella-tui-theme` |
| `TestBackend` snapshot of the status bar | 3 goldens, colour-carrying |
| no orange-band hex outside the theme crate | `no_hex_literals_in_v2_render_code`, scoped to v2 render code — see below |

## The witness

`prompt.md` rule 3 says the clamp tests must fail on any warm hex. Substituting
the repo's **own v1 palette** — gold `#FFB81A`, warm-paper text `#F4F1EA` —
fails four tests:

```
gold #FFB81A fails SPEC 3.2: needs r > g > b, g >= 0.78r (>= 198), b <= 0.35r (<= 89)
gray token `text` #F4F1EA fails SPEC 3.2: needs r == g and b >= g.
```

The v1 gold fails on the *green* ratio (184 < 198) — it is orange-band by the
v2 clamp, which is the drift SPEC 3.2 exists to stop.

## What is here

`crates/stella-tui-theme` — a near-leaf (`ratatui` only, the `stella-tty` /
`stella-diff` shape): SPEC 3.1's seventeen tokens, SPEC 3.2's hue clamp as unit
tests on the shipped table, SPEC 4's sixteen glyphs, SPEC 3.3's `stella*`
wordmark, SPEC 3.5's 16-color fallback with `COLORTERM` detection split pure.

The anti-drift mechanism is the **role table**, not the individual assertions:
`token::ALL` pairs every token with a `Role`, and the role picks its clamp. A
token added tomorrow has to declare what kind of colour it is, and a warm hex
has no honest declaration to pick. Same shape for fallback totality — a token
without a 16-color stand-in fails rather than shipping a 24-bit value to a
terminal that cannot show it.

`crates/stella-tui/src/v2/status_bar.rs` — SPEC 5's single-line bar.

## Three judgement calls a reviewer should check

**1. `gold_bright` violates the spec's own clamp — #4052.** `#F7D96B` measures
`b/r = 0.433` against SPEC 3.2's stated `0.35`. `prompt.md` rule 4 forbids
inventing a colour, so the value stands byte-exact and the exception is
recorded as `clamp::GOLD_LIFT_BLUE_PCT = 44` — the tightest whole percent that
admits it — with a test asserting it stays tight and that a lift must *fail*
the resting clamp. This is a workaround with a comment on it. #4052 carries
both ways out.

**2. The bar is not wired live — #4053.** `det %` has no source anywhere in the
workspace (`rg det_pct|deterministic_pct|det_ratio|det_share crates/` is
empty). Pricing deterministic work is the first of SPEC 1's four theses; a
fabricated `det 86%` would falsify it on every frame. So the widget takes an
explicit `Status` input and the projection waits.

**3. The hex sweep is scoped to v2 render code, not the workspace.** `v1` is
warm-neutral *by design* — `crates/stella-tui/src/palette.rs` states "no cool
grays anywhere on the dark side" — and v2's clamp rejects a warm gray outright.
The two cannot be one table, so they coexist until each surface migrates, and a
whole-workspace sweep would fail on a tree this PR is not entitled to change.
The v2 spec's scope is the TUI; nothing here touches `docs/brand/`, the
website, or the observatory. (`stella-tui::palette` already documents itself as
held apart from the web kit's `#C58A32`.)

## Goldens carry colour, unlike the v1 deck's

`tests/deck_render_snapshots.rs` strips styling deliberately, and its reasoning
is sound for v1. v2 cannot pay that cost: the two-metal rule and red scarcity
are claims about colour and nothing else, and a frame with the right glyphs in
the wrong metal passes a character-grid golden while having broken the design.
So each golden is a character grid plus a **metal map** — one letter per cell
naming the token its foreground came from:

```
kimi-k3 · execute · ctx ████▎░░░░░░░ 35% · $0.45 · saved $0.69 · det 86% · ✉ 21               ? help
wwwwwwwdddwwwwwwwdddmmmmGGGGGbbbbbbbwwwwdddGGGGGdddmmmmmmGGGGGdddmmmmwwwdddssww---------------dddddd
```

Money gold (`G`), meter gold on `border` gray (`b`), inbox silver (`s`), hints
dim (`d`), zero `R`. Reviewable by eye against
`design/tui-v2/renderings/png/01-session-turn-lifecycle.png`.

## Two defects the tests caught before implementation settled

- A 99.9% context meter rendered **identically to a full one** — the partial
  eighth-block rounded up to `█`. `partial_cell` now clamps to `1..=7`:
  rounding may move the head of the bar, never either end. "Full" is the
  reading a user compacts on.
- A non-finite `ctx_used` reached `as usize` on a saturating cast nobody chose
  (`f64::clamp` propagates `NaN`). `ratio()` now folds non-finite to zero.

## New dependencies

- `stella-tui-theme` (workspace, new) — argued above.
- `insta` (dev-dep, `stella-tui`) — named in the plan's dependency table.
  Bought for the review workflow the colour-carrying goldens need; the v1
  hand-rolled `BLESS=1` harness is untouched.

## Witness test

`crates/stella-tui-theme/src/tests.rs` and
`crates/stella-tui/tests/v2_status_bar.rs` are entirely new; every assertion
fails on `main` because the code under test does not exist there. The
substitution witness above is the one that matters — it proves the clamp
discriminates rather than merely passing.

## Verification run

`make guards` green (31 steps). `cargo clippy --all-targets -D warnings` clean.
`RUSTDOCFLAGS=-D warnings cargo doc` clean. `cargo test -p stella-tui-theme -p
stella-tui` green.

`cargo test --workspace` has one failure,
`goal_wrapped_dispatch_cli::a_goal_run_dispatches_each_round_through_the_bound_wrapper`,
which is **pre-existing and not from this branch**: it fails identically on the
unmodified merge base `2e9c048e0` (control run in a separate worktree), this
branch touches no file in `stella-cli`, and the cause is already tracked three
times over — #3938, #3996, #3876 (the test reads the real `~/.stella`, so it
fails on any configured machine).

## Tests deleted

None.

Refs #4052, #4053

## Summary by Sourcery

Establish the tested TUI v2 visual foundation and provide its first single-line status-bar surface.

New Features:
- Add the TUI v2 theme foundation with a checked palette, shared state glyph vocabulary, `stella*` wordmark, and 16-color terminal fallback.
- Introduce a single-line v2 status bar showing worker, stage, context usage, spending, savings, inbox count, and help.

Bug Fixes:
- Prevent context meters from rendering near-full values as indistinguishable from full and handle non-finite or out-of-range ratios safely.

Enhancements:
- Enforce palette role and hue constraints through theme-level tests and prevent direct color construction in v2 render code.
- Keep the v2 surface isolated from the existing v1 palette to support incremental migration.

Build:
- Add the new `stella-tui-theme` workspace crate and the `insta` snapshot-testing dependency.

Documentation:
- Add the TUI v2 specification, implementation plan, renderings, theme README, and workspace guidance for the new crate.

Tests:
- Add theme tests covering palette values, color constraints, fallbacks, glyphs, and wordmark behavior.
- Add color-carrying `TestBackend` snapshots and status-bar behavior tests, including red-scarcity and palette-literal checks.


macanderson:
Updated by `56d5d46a6`:

- **`det %` removed from the status bar** (owner's call). SPEC 5 now states the removal and the reason so a later phase does not restore it; the tag form (`$0.00 · det` on a gate) is a boolean and is unaffected. The bar is down to six values, all of which the v1 statline already projects — see #4053, retitled, since its original premise is gone.
- **#4052 closed by anchoring the lift rather than excepting it.** Neither filed option was right: `b <= 0.35 r` caps lightness at 0.6745 for *any* colour and `gold_bright` sits at 0.6941, so the ceiling was unsatisfiable, not strict. The clamp is now split — hue rule for every gold, saturation rule for the resting gold, and a lift anchored to `GOLD` (same hue within 3°, strictly lighter). One authored gold, one proven consequence.
- `design/tui-v2/prompt.md` is not committed: `check-invariants` correctly fails it for restating AGENTS.md's "No new dependencies" in normative form, and it is operator instructions rather than a repository document.

**Collision with #4066** — that PR creates the same crate and generates `GOLD_LIFT_BLUE_PCT = 44` / `Clamp::LiftedGold`, the exception this branch just removed. Full analysis and a suggested composition posted there; this PR should rebase onto #4066 once it lands.

Issue #4064, opened 2026-08-20
The colour system has no warning token; --stella-warning currently overloads gold


macanderson:
The black-and-gold system ships two verdict colours — `green` (pass) and `red`
(fail) — and no warning. Several surfaces have a warning state, so the v5.0
migration had to point `--stella-warning` somewhere. It is aliased to `gold`,
and that is an overload, not a decision.

## The bind

Rule 2 assigns gold a meaning: "stella or the user acting (primary buttons,
active tabs, running work, money, the mark)". A warning is not that. But the
colour a warning conventionally wants is an amber, and amber is precisely what
the resting-gold clamp rejects — `g >= 0.78r` and `b <= 0.35r` is the predicate
that separates gold from the orange band, and rule 1 forbids inventing a colour
outside it.

So the three available moves are all imperfect:

1. **Keep the gold alias** (today's state). A warning and an action share a hue,
   which is survivable only because rule 4 already requires a glyph or word
   beside every state — the reader distinguishes them by the label, not the
   colour. Cheapest; weakens gold's meaning.
2. **Fold warning into `red`.** Honest to the two-metal system (there are two
   outcomes, and a warning is a soft fail) but rule 3 rations red hard: "a
   healthy screen contains zero red", and warnings are not rare.
3. **Add a warning token.** Needs a clamp role of its own, since it is neither
   metal nor gray nor verdict, and a hue far enough from gold to be
   distinguishable. The retired v4.0 kit reasoned about this in OKLCH hue
   distance rather than luminance contrast — see the note that used to sit in
   `website/src/app/tokens.css` (v4.0 moved its warning because the old amber
   sat 8.7° from the brand hue, which WCAG contrast cannot measure). That
   reasoning is the right shape for whatever replaces it.

## Where it is today

- `website/src/app/tokens.css` — `--stella-warning: var(--st-gold)` with the
  open-exception note at its definition
- `crates/stella-observatory/src/assets/index.html` — `--warn`, which reasons
  about hue distance from `--identity` and is the most developed prior art
- `design/tokens/stella-tokens.json` — where a new token and its clamp role
  would be declared

## Definition of done

Either a `warning` token in the JSON with a declared clamp role and a stated hue
distance from gold, or a recorded decision that warning folds into an existing
role — with the alias in `tokens.css` repointed and its open-exception note
replaced by the decision.

macanderson:
Closing as **already decided**, by #4086 (merged 2026-08-20T05:46:00Z) — option 3
[...]

Issue #4063, opened 2026-08-20
Four palette pairings miss WCAG AA, including the spec's own stated floor (muted on bg, 4.47:1)


macanderson:
Four pairings the black-and-gold palette licenses do not clear WCAG AA. The
canonical token values are **final** and rule 1 forbids inventing a colour, so
this is a decision for a human rather than something a migration should quietly
adjust. Filed with the numbers rather than resolved.

## Measured

`python3 scripts/check-contrast.py` (WCAG 2.1 relative luminance):

```
muted on bg      4.47:1   needs 4.5   secondary text on canvas
muted on panel   4.32:1   needs 4.5   secondary text on panel
dim on bg        2.30:1   needs 3.0   hints, keybinding rows, line numbers
comment on panel 2.64:1   needs 3.0   code comments
```

The first two are marginal — `muted` on `bg` misses by 0.03. The second two are
not: they are under even the 3:1 large-text/non-text floor, which is the more
lenient threshold the script already grants them.

## Why this is a real question and not a rounding complaint

`SPEC-stella-tui-v2.md` §13 sets the accessibility floor as "`muted` on `bg` is
the floor; nothing dimmer than `dim` may carry required information". Read
literally that makes `dim` a carrier of required information — keybinding rows
and line numbers are exactly that — at 2.30:1. And §6.4 puts code comments in
`comment` on a `panel` ground at 2.64:1; a comment in source is content.

So the spec's own floor sits fractionally under AA, and two roles the spec
assigns real information to sit well under it.

## Options, none of which invent a colour

1. **Accept and document.** A terminal UI is not a WCAG conformance target, and
   rule 4 already requires a glyph or label beside every state, so colour is
   never the sole carrier. This is the cheapest answer and may be the right one
   — but it should be a stated decision, not an omission.
2. **Reassign roles.** Use `silver` (8.58:1) where `muted` currently carries
   required secondary text, and `muted` where `dim` currently carries hints.
   Invents nothing; changes the design's intended hierarchy.
3. **Re-cut three tokens.** Contradicts "these values are final" and needs the
   hue clamp re-run — `muted`, `dim` and `comment` are all `neutral-gray`, so
   any replacement must satisfy `r == g` and `b >= g`.

## Files

- `scripts/check-contrast.py` — the measurement; every pairing and its threshold
- `design/tokens/stella-tokens.json` — the tokens, and the clamp any change must satisfy
- `SPEC-stella-tui-v2.md` §13, §6.4

[...]

Issue #4058, opened 2026-08-20
The terminal UI still hard-codes its own colors, gradient, and old wordmark instead of using the shared design tokens (Phase 2 of the color alignment)


macanderson:
`crates/stella-tui` still holds its own hand-typed palette. The v5.0 colour-system
migration deliberately stopped at its boundary.

## Problem

- `crates/stella-tui/src/palette.rs` — 4 retired hexes
- `crates/stella-tui/src/theme/tests.rs` — 3 retired hexes
- `crates/stella-tui/src/theme.rs` — a **brand gradient** used for the wordmark
  sweep and the progress-bar fill. Rule 5 of the alignment retires gradients on
  every surface; SPEC 2 (`cell-grid honest`) allows only per-cell colour steps.
- ~20 files under `crates/stella-tui/tests/snapshots/deck/` render the retired
  `✧ stella` wordmark. SPEC 3.3 replaces it with `stella*`, because `✧` is the
  **skill** glyph in SPEC 4 and the two meanings collided on every skill row.

## Why it was not done here

This is Phase 2 of the alignment, and it is already in flight in another
session's worktree (`.claude/worktrees/tui-v2-p0`), which has an uncommitted
`crates/stella-tui-theme/` containing `token.rs`, `clamp.rs`, `glyph.rs`,
`wordmark.rs` and `fallback.rs`. Editing the same files from a second branch
would produce two authorities over one palette, which is precisely the failure
this whole change exists to end.

## What already landed that this depends on

`crates/stella-tui-theme/` now exists on `main` with `src/generated.rs`, emitted
by `scripts/gen-tokens.py` from `design/tokens/stella-tokens.json`. It carries
every token as a `ratatui::style::Color`, the clamp predicates, the walkable
`ALL` table, and three tests (every token against its declared clamp, the gold
lift proved bounded, and the retired brand golds asserted to fail).

## Definition of done

Rewritten 2026-09-05 against the comment thread's own finding: item 1's
original "no hex literal remains" cannot land as one PR — it is four
independent token-system decisions (see the comment linked from item 1
below). Splitting them into their own issues is what the thread already
recommended; this DoD now says so explicitly instead of leaving the split
undone in the checklist.

- [x] 1. `stella-tui` depends on `stella-tui-theme` and imports the generated
  consts, for every value the token system already declares a home for.
  Every literal the token system does **not** yet have a home for is
  tracked in its own issue instead of blocking this one: #5997 (the light
[...]

Issue #4057, opened 2026-08-20
docs/brand/prompts/ still briefs the retired v4.0 palette and the banned ✦ stella lockup


macanderson:
Two design-prompt documents under `docs/brand/prompts/` still carry the retired
v4.0 palette and, in one case, a wordmark form the system has banned.

## Problem

- `docs/brand/prompts/stella-design-system-prompt.md` (21 retired hexes)
- `docs/brand/prompts/stella-tui-prompt.md` (5 retired hexes, plus three
  occurrences of the retired `✦ stella` lockup)

These are *design prompts* — briefs written to generate surfaces — so their hex
values are instructions, not styling. Remapping the hexes without rewriting the
surrounding instructions produces a brief that tells its reader to build the old
system with new numbers.

`stella-tui-prompt.md` in particular is superseded outright by
`SPEC-stella-tui-v2.md` (sections 2–4 define the palette, hue clamp, glyph
language and wordmark). It also still specifies `≡✦ stella` as the lockup;
SPEC 3.3 retires that form because `✦` is the **skill** glyph in SPEC 4, so the
brand mark and the skill row collided on every screen. The wordmark is `stella*`.

## Decision needed

Whether either document should be rewritten or simply marked superseded. A
superseded brief that stays in the tree is fine if it says so at the top; what
is not fine is one that reads as current.

## Files

- `docs/brand/prompts/stella-design-system-prompt.md`
- `docs/brand/prompts/stella-tui-prompt.md`
- `design/tokens/stella-tokens.json` — the normative palette
- `SPEC-stella-tui-v2.md` §2–4 — supersedes the TUI prompt
- `scripts/check-tokens.py` — both paths are in its `MIGRATING` ledger

## Definition of done

Either both documents are rewritten against the v5.0 system and SPEC-v2, or
each carries a superseded banner naming what replaced it and has its retired
hexes removed. Then drop both from `MIGRATING` and confirm `make tokens` is
green.

macanderson:
Fixed and **shipped to `main`** in #4296, squash-merged as `8faa2d155`.

Closing by hand because the trailers did not fire: the squash composed its commit message from the PR *body*, not from `COMMIT_MESSAGES`, so the `Closes #N` trailers written on the individual commits never reached `main`. That is the exact failure mode AGENTS.md warns about — `Closes #N` belongs in the PR description **and** on a commit, and this PR had it only on the commits.

[...]

PR #4066, opened 2026-08-20
feat(design): align every stella surface to one black-and-gold system

macanderson:
Aligns every stella-facing surface onto one colour system, defined by
`SPEC-stella-tui-v2.md` §2–4, and makes the alignment a checked property rather
than a promise in a doc comment.

## Why this supersedes brand v4.0

Not a taste call — a consequence of the spec's own hue clamp. Bronze Gold
`#C58A32` is r=197 g=138 b=50, and a gold must satisfy `g >= 0.78r`, i.e.
g >= 154. At 138 it misses by 16 points and sits in the orange band §3.2 names
as the failure mode: orange on a near-black ground reads brown on an
uncalibrated panel. Obsidian `#070B10` is green-tipped and fails the
neutral-gray rule (`r == g`, `b >= g`) too. Both anchors are outside the system,
so the kit is replaced rather than tuned.

## The mechanism

`design/tokens/stella-tokens.json` is the single upstream. `scripts/gen-tokens.py`
emits the CSS custom properties and the Rust consts and, with `--check`, fails
when either has drifted.

This exists because the palette is a **shared cell**: four surfaces need the
same twenty-one values, and every previous attempt to hold them together was a
sentence asking the next author to copy carefully. Twice that sentence was the
only thing left true — `website/src/app/tokens.css` sat on the v1.0 phosphor
ramp through the whole life of the v4.0 bronze rebrand while claiming to mirror
the kit, and later a reader crossing from the site to the Observatory watched
the brand change hue (#3968).

Each token declares the clamp role it is held to, so a new hex cannot enter
without a human choosing what kind of colour it is. Three tests walk the
declaration: every token against its own clamp, the gold lift proved bounded
(two points bluer fails), and the **retired brand golds asserted to fail**, so
the clamp cannot go slack without a test going red.

`make tokens` is gate step 32.

## Scope

| Phase | State |
|---|---|
| 0 · single source of truth | done — token JSON, generator, two guards, `stella-tui-theme` crate |
| 1 · mark recolour | done — 34 SVGs, 31 rasters, export verified pixel-exact |
| 2 · TUI | **not this PR** — in flight in another session (#4058) |
| 3 · web surfaces | done — both `tokens.css` roots, Observatory, transcript, OAuth page, arenabench |
| 4 · everything that prints text | done — CLI colours, badges; installer emits no ANSI |
| 5 · verification | done — see below |

Retired hexes in the tree: **646 → 0** outside a three-entry ledger, each naming
its issue.

## Evidence

- `logomark-color-256w.png`'s only opaque colour is `#EFC53F` across 5951 px;
  the dark lockup carries exactly `#E8E8EC` and `#EFC53F`.
- The new export guard was **witnessed** against a two-point drift (`#EFC33F`)
  and rejected it, naming the file and the colours actually found.
- `cometkit.verify_png`'s docstring said it "cannot catch bad pixels" and called
  judging them "the reviewer's job". `png_opaque_colours` + `check_exported_gold`
  close that gap, which is the one door the hue clamp cannot watch.
- guards-fast green (32 steps), clippy `-D warnings` clean, rustdoc clean,
  stella-observatory 133, stella-tui-theme 3, design_token_parity 10.

## Two defects found by doing this

1. **The Observatory's light theme had a gold fill with a white label** —
   white-on-gold at 1.65:1, introduced by the recolour. Rule 6 permits gold on a
   light ground precisely for filled elements carrying **ink** text; the label is
   the half that makes the fill legal. Now 11.15:1, and the parity test asserts
   the pair rather than merely being renamed.
2. **The banned-hex guard watched one notation of a two-notation language.** CSS
   accepts `rgb(197 138 50)` for the same colour a hex sweep reads as `#C58A32`,
   and 32 retired values survived the first pass purely by not being in hex. The
   guard now decodes `rgb()`/`rgba()`.

## Deliberately not resolved

- **#4063** — four palette pairings miss WCAG AA, including the spec's own
  stated floor (`muted` on `bg`, 4.47:1) and two roles carrying real information
  (`dim` 2.30:1, `comment` 2.64:1). The tokens are final and rule 1 forbids
  inventing a colour, so this is a human decision. `check-contrast.py` is
  deliberately **not** a gate step: it fails today, and wiring a red check in to
  then relax its thresholds is the expedient CLAUDE.md forbids.
- **#4064** — the system has no warning token; `--stella-warning` overloads gold.
- **#4056 / #4057 / #4058** — the three ledger entries.

## Tests deleted

None.

## Summary by Sourcery

Adopt the Stella v5 black-and-gold design system across all completed product and brand surfaces, with generated tokens and automated guards preventing palette drift.

New Features:
- Introduce a single black-and-gold Stella token system shared across brand, web, CLI, Observatory, benchmark, and exported surfaces.
- Add generated CSS and Rust token artifacts with a TUI theme crate and enforceable colour-clamp validation.
- Add exact exported-logo pixel checks and broader retired-colour detection, including RGB and RGBA notation.

Bug Fixes:
- Align previously divergent surfaces and assets with the current Stella palette.
- Fix light-theme gold fills to use readable ink labels instead of low-contrast white text.
- Replace decorative coloured gradients, glows, and status misuse with compliant palette roles.

Enhancements:
- Rework semantic colour roles around a flat palette with role-based tokens, neutral surfaces, and explicit light/dark mappings.
- Standardize CLI and benchmark presentation colours across product outputs.

Build:
- Add token generation and validation to the Make gate and document token update/check commands.
- Add the stella-tui-theme crate to the workspace.

CI:
- Make generated-token parity, retired-colour checks, clamp validation, and export verification part of the checked workflow.

Documentation:
- Update brand, website, Observatory, and project documentation to describe the v5.0 colour system and its usage rules.

Tests:
- Add token clamp, retired-brand rejection, bounded gold-lift, cross-surface parity, light-theme contrast, and raster export verification tests.

Chores:
- Recolour the complete set of SVG, raster, presentation, documentation, and UI assets to the unified palette.


macanderson:
Heads-up from the parallel TUI v2 thread (PR #4055, branch
`worktree-tui-v2-p0`): **we are both creating `crates/stella-tui-theme`**, and
this PR's `generated.rs` carries a clamp design that #4055 has since replaced
and closed as #4052. Flagging before either lands, because the composition is
clean but the ordering matters.

## The overlap

| | this PR | #4055 |
|---|---|---|
| `src/generated.rs`, `design/tokens/*`, `scripts/gen-tokens.py`, `make tokens` | ✅ | — |
| `src/lib.rs`, `Cargo.toml` | ✅ | ✅ (conflict) |
| token values + clamp predicates | ✅ generated | ✅ hand-authored (conflict) |
| `src/glyph.rs`, `src/wordmark.rs`, `src/fallback.rs` | — | ✅ |
| `src/v2/status_bar.rs` + colour-carrying goldens | — | ✅ |

Non-overlapping apart from the crate root and the palette itself.

## The one substantive disagreement

`generated.rs` emits:

```rust
pub const GOLD_LIFT_BLUE_PCT: u32 = 44;
...
("gold-bright", GOLD_BRIGHT, Clamp::LiftedGold),
```

That constant was mine, from #4055's first commit, and it was a mistake. It is
a second blue ceiling reverse-engineered from the one token it had to admit —
a bound derived from the value it must admit can say no to nothing in
principle. `56d5d46a6` replaced it after working out the geometry:

> In a gold, red is the brightest channel and blue the dimmest, so lightness is
> `(r + b) / 510`. `b <= 0.35 r` bounds `b` at `0.35 r`; push `r` to its maximum
> of 255 and `b` follows to 89 — lightness `344 / 510 = 0.6745`, and there is
> nowhere further to go because `r` cannot exceed 255. `gold_bright` sits at
> `0.6941`.

So `b <= 0.35 r` is not a strict rule that `gold_bright` happens to break;
**above lightness 0.6745 it is a rule no colour can satisfy.** A lift could
never have been held to it, and a second ceiling is not a fix — it is the same
category error, one notch looser.

What replaced it: the two clauses are stated where each is coherent.

| Clause | Governs | Why |
|---|---|---|
| `r > g > b`, `g >= 0.78 r` | **every** gold | the hue rule — gold versus orange, true at every lightness |
| `b <= 0.35 r` | the resting gold | the saturation rule — a hue that is not lightened holds full saturation |
| same hue within 3°, strictly lighter | a lift | anchored to `GOLD` itself |

An anchor is strictly stronger than a ceiling: a ceiling admits every colour
beneath it and keeps passing after the gold it was cut for is gone; an anchor
admits only *the authored gold, brighter*. Witnessed — recolour `GOLD` to
`#EFB82F` and leave `GOLD_BRIGHT` behind, and `gold_bright_is_a_lift_of_gold`
fails naming the hue distance. The ceiling keeps passing silently.

The 3° tolerance is derived, not chosen: `stella-tui`'s v1 theme records that
its warning amber sits 4.0° from gold "so an outcome may never be told from
chrome by hue alone", so a lift — which must be the *same* hue — sits inside
that.

This matters more here than it did in #4055, because generating the exception
puts it in a JSON schema, a generator and a gate step, where it is three edits
to remove instead of one.

## Suggested composition

Nothing here is in tension on intent — this PR's argument for a generator is
right, and it is the repo's own argument ("a number in two places is how the
last limit died"). The values should be generated; the clamp should be the
anchored one. Concretely:

1. **This PR lands first.** It is the wider change and its generator should own
   the values.
2. In the generator, replace `Clamp::LiftedGold` / `GOLD_LIFT_BLUE_PCT` with
   the anchored predicate: a `gold-lift` role emits a check against `GOLD`'s
   hue and lightness rather than against a channel ceiling. #4055's
   `crates/stella-tui-theme/src/clamp.rs` has `is_lift_of`, `hue_degrees`,
   `lightness` and `hue_distance` ready to lift across, plus
   `the_resting_blue_ceiling_is_unsatisfiable_above_this_lightness`, which
   keeps the geometry as an exhaustive test over all 65,536 `(r, b)` pairs so
   nobody re-derives it.
3. **#4055 then rebases onto it**, dropping its hand-authored `token.rs` in
   favour of `generated.rs` and keeping what this PR does not have: the glyph
   vocabulary (SPEC 4), the `stella*` wordmark (SPEC 3.3), the 16-color
   fallback (SPEC 3.5), the status bar (SPEC 5), and the colour-carrying
   goldens.

That composition is better than either branch alone: generated values mean the
clamp tests run against *the thing that is actually emitted*, so a bad token in
`stella-tokens.json` fails the hue clamp rather than shipping.

One smaller note, since the memory of this work records that "a hex sweep
misses `rgb()` notation" — #4055's `no_hex_literals_in_v2_render_code` greps
for `Color::Rgb(` in v2 render modules specifically, which catches the Rust
side of that gap.

Happy to do the reconciliation in either direction — say which lands first.

Issue #4059, opened 2026-08-20
Epic: bring the terminal UI in line with its design spec, under one shared color system


macanderson:
## The outcome

The terminal UI ships what `design/tui-v2/SPEC.md` specifies, drawn from **one**
authority over the palette, and the SPEC does not contradict itself.

## Why an epic now

Three open issues share that outcome, and each is blocked on the same underlying
condition rather than on each other: the v5.0 colour-system migration stopped at
three different boundaries and left a recorded exception at each one. Filed
separately they read as unrelated polish — a hex, a widget, a crate boundary.
Together they are one statement: *the v2 design system is specified but not yet
the thing that ships.*

This epic was deliberately **not** created earlier in the 2026-08-19 triage
sweep, when only #4052 and #4053 existed. Two issues is below the bar, and the
sweep recorded them as unlinked issues on #2778 rather than inventing a thin
hierarchy. #4058 is the third, so the epic is now warranted rather than assumed.

## Children

- **#4052** — SPEC 3.1's `gold_bright` `#F7D96B` fails SPEC 3.2's own hue clamp
  (blue ratio `107/247 = 0.433` against a stated ceiling of `0.35`). P0 of the
  build shipped a *recorded exception*, not a fix. A spec that violates its own
  stated rule cannot be the authority the other two issues defer to, which is
  why this one is upstream of the rest in practice even though it looks the
  smallest.
- **#4053** — the v2 single-line status bar
  (`crates/stella-tui/src/v2/status_bar.rs`) is built, tested and unreachable,
  because `det %` — the deterministic/model split — has no source anywhere in
  the workspace. Note what makes this one delicate: "deterministic first, model
  last" is the **first** of the four theses SPEC 1 says every screen exists to
  make visible, so a fabricated `det 86%` would falsify it on every frame, in
  the one place a user reads it. The widget was shipped over an explicit input
  struct precisely to avoid inventing the number. **Wiring the bar means
  building the projection first, not filling in a placeholder.**
- **#4058** — `crates/stella-tui` still holds its own hand-typed palette
  (`palette.rs`, `theme/tests.rs`), a brand gradient that Rule 5 retires on
  every surface, and ~20 interactive-mode snapshots rendering the retired
  `✧ stella` wordmark — `✧` being the *skill* glyph in SPEC 4, so the two
  meanings collided on every skill row.

## Sequencing constraint, carried from #4058

[...]

Issue #1750, opened 2026-08-06
Gate: an orphaned .rs file under src/ passes every check while its tests never run


macanderson:
## Problem

Deleting a `mod` declaration silently disables the file it pointed at, and **nothing in `make gate` notices**. `rustfmt` and `rustc` both walk `mod` declarations from the crate root, so an orphaned `.rs` file is invisible to the compiler, the formatter, and clippy simultaneously. `cargo test` still reports green.

Found in PR #1739: commit `f6284d0d` dropped `#[cfg(test)] mod tests;` from the bottom of `crates/stella-graph/src/store.rs`. Effect, measured:

```
$ cargo test -p stella-graph --lib -- --list | rg -c '^store::tests::'
0          # without the declaration — and the suite reports "ok"
19         # with it restored
```

`crates/stella-graph/src/store/tests.rs` was also unformatted the whole time and `cargo fmt --check` passed, because rustfmt never reached it either. So a single deleted line hid a file from three gates at once. Fixed on that branch in `e639a71b` / `728d50d4`, but only because a reviewer happened to read the diff.

## What "done" looks like

A gate step (sibling of `scripts/check-left-behind.sh`, wired into `GATE_STEPS` in the `Makefile` and into `ci.yml`) that fails when a tracked `.rs` file under any crate's `src/` is not reachable from that crate's root through `mod` declarations.

Reachability has to account for the real shapes in this tree:
- `mod foo;` → `src/foo.rs` **or** `src/foo/mod.rs`
- nested: `src/driver.rs` declaring `mod settlement;` → `src/driver/settlement.rs`
- `#[cfg(test)] mod tests;` and other `#[cfg(...)]`-gated declarations still count as reachable
- `#[path = "..."] mod x;` — rare but present in some crates; check before assuming it is absent
- `build.rs`, `benches/`, `tests/`, and `examples/` are separate targets, not `src/` modules — exclude them
- each crate's roots are `src/lib.rs` and `src/main.rs` (plus any `[[bin]]` paths in its `Cargo.toml`)

A text-level walk of `mod` declarations is enough and keeps the guard in the toolchain-free `guards-fast` rung, where it belongs — this is a fact about the repository, not about a crate, so it must not honour `CARGO_SCOPE` (AGENTS.md § "The gate").

## Verify

[...]

PR #4079, opened 2026-08-20
fix: main is red — the theme crate's generated palette is unreachable and the lock is stale

macanderson:
## What is red

`ci.yml`'s required job fails on **two** guard steps on the current main tip
(`6f4a7cd22`), before clippy/test ever run — one red gate step masks the next:

| Step | Status on `6f4a7cd22` |
|---|---|
| `every src/ file is reachable from its crate root` | FAILED |
| `Cargo.lock is in sync with the manifests` | FAILED |

Both are shared-cell collisions: each side landed green on its own branch and
composed red on main, with no textual conflict for git to report.

## 1. The generated palette is unreachable

`crates/stella-tui-theme/src/generated.rs` is tracked, but no `mod`
declaration reaches it.

The crate was created **twice**:

- **#4066** landed first, as exactly `lib.rs` + `generated.rs`, whose `lib.rs`
  declared `pub mod generated;` and re-exported from it.
- **#4055** landed next and replaced `lib.rs` wholesale with its own
  five-module version (`clamp`, `fallback`, `glyph`, `token`, `wordmark`) —
  a branch that never contained a `mod generated;` line.

One side simply does not contain the other's lines, so the merge was clean and
the generated artifact went invisible to rustc, rustfmt **and** clippy at once,
while `make tokens` carried on checking it for staleness. Its four tests had
stopped running while the suite still reported `ok` — the #1750 shape.

Restoring the declaration is **#4066's own line, put back**, not a new design
decision. The crate now carries two copies of the palette (`generated::ALL`
against `token::ALL`); collapsing them into one is already tracked as **#4058**
under epic **#4059**, and is deliberately not folded in here.

### Why a plain `//` and not `///`

An outer doc comment on a `mod` declaration is merged with the module file's
own `//!` docs and resolves their intra-doc links in the **parent** file's
scope. With `///` the restored declaration produced three
`unresolved link` errors — `Color::Rgb`, `ALL`, `Clamp` — none of which exist
in `lib.rs`, failing `cargo doc -D warnings`. Verified both ways.

## 2. `Cargo.lock` is stale

The lock recorded `stella-tui-theme` at `0.9.112` while the crate inherits
`version.workspace = true` = `0.9.114`. The release bump and the new-crate PR
each wrote the lock correctly in isolation. One-line regeneration.

## Witness

Both guards go **FAIL to OK** on this branch, and the previously-dead tests run:

```
check-module-reachability: FAILED  ->  OK - 1028 source file(s) across 28 crate(s), all reachable
check-lockfile-sync:       FAIL    ->  OK - Cargo.lock resolves against the manifests
cargo test -p stella-tui-theme     ->  22 passed, including 4x generated::tests::*
                                        (they could not execute at all before)
cargo clippy -p stella-tui-theme --all-targets -- -D warnings  ->  clean
RUSTDOCFLAGS="-D warnings" cargo doc -p stella-tui-theme       ->  clean
cargo fmt --all --check                                        ->  clean
```

No test was deleted in this change.

## Left behind, tracked not dropped

`make tokens` is **also** red on main — retired brand hexes in 7 places
(`export/tests.rs`, `observatory/src/assets/index.html`,
`stella-tui-theme/README.md`, `website/src/app/tokens.css`) plus 13 non-token
hex literals in `website/src/app/tokens.css`.

It does not gate this PR, and the reason is worth stating precisely because I
first got it wrong: `tokens` **is** a `make gate` step — the last entry of
`GATE_GUARDS_FAST` in the `Makefile`, so `make gate` and the pre-push hook both
run it — but it appears in **no** workflow under `.github/workflows/`. So it
blocks a local push and nothing server-side. That parity hole is filed
separately, along with the red itself; a brand-wide hex sweep does not belong
inside a red-main repair.

Refs #4058, #4059

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3344,7 +3344,7 @@ dependencies = [
 
 [[package]]
 name = "stella-tui-theme"
-version = "0.9.112"
+version = "0.9.114"
 dependencies = [
  "ratatui",
 ]
--- a/crates/stella-tui-theme/src/lib.rs
+++ b/crates/stella-tui-theme/src/lib.rs
@@ -47,6 +47,21 @@
 
 pub mod clamp;
 pub mod fallback;
+// The generated face of `design/tokens/stella-tokens.json`, emitted by
+// `scripts/gen-tokens.py` and kept in sync by `make tokens`. It carries a
+// second copy of the palette — `generated::ALL` against `token::ALL` —
+// because this crate was created twice, by #4066 (generator first) and #4055
+// (hand-written table first), and the two landed without a textual conflict.
+// This declaration is #4066's, restored: without it the generated artifact is
+// tracked but unreachable, so rustc, rustfmt and clippy all walk past it
+// while `make tokens` still checks it for staleness. Collapsing the two
+// palettes into one is #4058, under epic #4059.
+//
+// Deliberately not a doc comment: an outer `///` here is merged with the
+// module's own `//!` docs and resolves their intra-doc links in *this* file's
+// scope, where `Color`, `ALL` and `Clamp` do not exist — three rustdoc
+// errors under the gate's `-D warnings`.
+pub mod generated;
 pub mod glyph;
 pub mod token;
 pub mod wordmark;


macanderson:
Superseded by #4083. This branch was cut before #4076 (Cargo.lock sync) and #4077 (version bump to 0.9.115) landed, so it conflicts on `Cargo.lock` — and a conflicting PR never gets a merge ref, which is why no workflow other than `cla` ever started on it. #4076 already fixed the lockfile half; #4083 carries only the module-reachability half, rebuilt on `04fc177be`.

Issue #3441, opened 2026-08-17
make file-size-update writes the baseline entry check-file-size.sh says must never exist (first-time 1500-line crossing)


macanderson:
## Problem

`scripts/check-file-size.sh` distinguishes two failures and names two different remedies:

- **grandfathered file whose ceiling drifted** → `make file-size-update`, land the baseline diff for review;
- **file that crossed 1500 for the first time** → *"split them into submodules. Do not add a baseline entry: the baseline only covers files that predate the guard, and this is the rule that stops the tree getting worse."*

`make file-size-update` (`scripts/check-file-size.sh --update`) does not make that distinction. It records **every** file currently over the line, so it will write exactly the entry the checker forbids.

## Concrete instance

On `main` at 1016925c9, `crates/stella-protocol/src/event.rs` was 1533 lines — a first-time crossing, correctly rejected by the checker. Someone clearing the *unrelated* `command_deck.rs` ceiling drift (4014 vs 4010) would run the documented remedy and get:

```diff
-4010 crates/stella-cli/src/command_deck.rs
+4014 crates/stella-cli/src/command_deck.rs
...
+1533 crates/stella-protocol/src/event.rs
```

The second hunk permanently grandfathers a god file, in a diff whose stated purpose was a +4 ceiling bump, and turns the gate green — so nothing downstream ever objects again. See #3436 for the case that surfaced it.

## Why it matters more than it looks

The two remedies are usually needed at the same time, because both are triggered by the same shared cell (`scripts/file-size-baseline.txt`) going stale. So the moment you most want to run `--update` is exactly the moment an unrelated first-time crossing is most likely to be sitting in the tree waiting to be absorbed. The failure is silent: the baseline diff is reviewable in principle, but a reviewer looking at a "bump command_deck by 4" PR has no reason to scrutinise the other lines.

This is the same class as the ratchet's own base-comparison fix (#2004) — the tool must judge the *change*, not the tree.

## Definition of done

- `scripts/check-file-size.sh --update` refuses to add a baseline entry for a file that is not already in the baseline, printing which file and pointing at the split remedy. Existing entries continue to be rewritten to current size (the down-ratchet must keep working).
- An escape hatch only if genuinely needed for a real irreducible case, and it must be explicit and noisy (e.g. `--grandfather <path>`), never the default.
[...]

PR #4082, opened 2026-08-20
fix(stella-cli): split the run ledger out of self_driving_cmd.rs so main compiles its own ceiling

### LINKED ISSUE #4044: main is red: file-size fails on the merged tree
The post-merge canary found `main` broken at `a0b1f8fcd`.

Failing: **file-size**

### `file-size`

```
check-file-size: FAILED

worse.
  crates/stella-cli/src/self_driving_cmd.rs is 1501 lines, over the 1500-line limit
```

canary exists (#3332).

### How to fix

```sh
git checkout main && git pull
```

...then, per failing check:

#### `file-size`

```sh
```

than only CI.

macanderson:
## The defect

`file-size` fails on `main` itself. `crates/stella-cli/src/self_driving_cmd.rs`
is 1501 lines on the merged tree — one line over the limit — and carries no
baseline entry, so the guard fails outright:

```
crates/stella-cli/src/self_driving_cmd.rs is 1501 lines, over the 1500-line limit
```

Per AGENTS.md § `main-red-hold`, a red `main` reds every open PR, so this is
costing every contributor a failure they did not cause. Filed by the post-merge
canary as #4044.

## The fix

Not a raised ceiling. `run_start`, `run_end`, `run_end_as` and `runs_report`
are one cohesive unit — the run ledger, the outer bracket around the cycles
`lifecycle` drives. They share no state with the argument tree above them or
the cycle reporting beside them, and they are called from exactly four sites
inside a single `match` arm.

They move to `crates/stella-cli/src/self_driving_cmd/run_lifecycle.rs`, beside
the fourteen submodules already in that directory. **Exemplar:**
`crates/stella-core/src/driver/settlement.rs`, split out of `driver.rs` the
same way — the pattern AGENTS.md § "God files" names.

The move is mechanical. No function body changed. What changed:

- visibility: private → `pub(super)` on the four functions
- the four call sites gain a `run_lifecycle::` qualifier
- `use std::collections::BTreeMap;` left the parent's import list with them

Parent lands at **1421 lines**, 79 under the ceiling.

**No entry added to `scripts/file-size-baseline.txt`.** The baseline covers
files that predate the guard; a first-time crossing bought with a raised
ceiling is the expedient CLAUDE.md forbids, and `--update` refuses to add a
new entry for exactly this reason (#3441).

`run_end` is a pure pass-through to `run_end_as` with an identical signature.
It moved as-is rather than being collapsed — this PR unblocks `main` and does
one thing.

## Witness

Not a behaviour change, so no witness test — this is a pure code move plus the
guard that was already failing. The guard **is** the evidence, and it flips:

```
$ ./scripts/check-file-size.sh          # on main
check-file-size: FAILED
  crates/stella-cli/src/self_driving_cmd.rs is 1501 lines, over the 1500-line limit

$ ./scripts/check-file-size.sh          # on this branch
check-file-size: OK — 1496 Rust/Python/shell files, 23 grandfathered over
1500 lines, and nothing went over by this change. Judged against 35f984c30.
```

No test was deleted.

Closes #4044

## Summary by Sourcery

Split the self-driving run ledger into its own module to restore compliance with the source file-size limit without changing runtime behavior.

Bug Fixes:
- Reduce `self_driving_cmd.rs` below the file-size limit so the main branch guard passes again.

Enhancements:
- Extract run lifecycle ledger operations into a dedicated `run_lifecycle` module without changing behavior.

### DIFF
--- a/crates/stella-cli/src/self_driving_cmd.rs
+++ b/crates/stella-cli/src/self_driving_cmd.rs
@@ -27,13 +27,13 @@ mod drive;
 mod hooks;
 mod lifecycle;
 pub(crate) mod probes;
+mod run_lifecycle;
 pub(crate) mod state;
 mod stats;
 mod surface;
 mod triage;
 mod work;
 
-use std::collections::BTreeMap;
 use std::process::Command;
 
 use clap::Subcommand;
@@ -623,14 +623,14 @@ pub(crate) fn run(cmd: &SelfDrivingCmd, spend_limit: Option<f64>) -> Result<(), 
             remaining,
         }),
         SelfDrivingCmd::Run { cmd } => match cmd {
-            RunCmd::Start => run_start(&st),
-            RunCmd::End { status, reason } => run_end(&st, status, reason),
-            RunCmd::Cancel { reason } => run_end_as(
+            RunCmd::Start => run_lifecycle::run_start(&st),
+            RunCmd::End { status, reason } => run_lifecycle::run_end(&st, status, reason),
+            RunCmd::Cancel { reason } => run_lifecycle::run_end_as(
                 &st,
                 "cancelled",
                 reason.as_deref().unwrap_or("stopped by hand"),
             ),
-            RunCmd::List => runs_report(&st),
+            RunCmd::List => run_lifecycle::runs_report(&st),
             RunCmd::StampCyclePid { pid } => {
                 st.update_run_doc(|doc| match pid {
                     Some(p) => {
@@ -643,7 +643,7 @@ pub(crate) fn run(cmd: &SelfDrivingCmd, spend_limit: Option<f64>) -> Result<(), 
                 Ok(())
             }
         },
-        SelfDrivingCmd::Runs => runs_report(&st),
+        SelfDrivingCmd::Runs => run_lifecycle::runs_report(&st),
         SelfDrivingCmd::Phase { name } => {
             st.run_write(name, None, None);
             println!("phase: {name}");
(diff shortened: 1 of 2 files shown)


macanderson:
Cross-reference: this and #4080 both close #4044 by splitting `crates/stella-cli/src/self_driving_cmd.rs` under the 1500-line ratchet, opened minutes apart. **Only one should merge** — both rewrite the same file, so the second will conflict.

They cut at different seams, and both are defensible:

- **This PR** extracts the run ledger into `self_driving_cmd/run_lifecycle.rs`.
- **#4080** extracts the two read-only reporting verbs (`watch`, `metrics`) into `self_driving_cmd/report.rs`, which were already grouped behind a shared `// watch / metrics` section banner in the file. It lands at 1378 lines and widens no visibility — a child module can already see its parent's private items, so `say`, `gh_available`, `gh_plain` and `demand` stay private.

No preference being asserted here beyond "pick one". Whichever merges, the other should be closed rather than rebased, since the remaining diff would be an unrelated second split.

macanderson:
Closing as a duplicate. #4080 was opened first for the same defect (#4044) and takes the same remedy — split a cohesive group out of `self_driving_cmd.rs`, no baseline entry. Its cut (`watch`/`metrics` → `self_driving_cmd/report.rs`) follows a section banner the file had already drawn, which is at least as good a seam as the run ledger I moved.

I should have run `gh pr list` before starting; `scripts/file-size-baseline.txt` and this file are exactly the shared cell AGENTS.md warns two correct branches can collide on. #4080 should land.

Issue #4081, opened 2026-08-20
make tokens is a gate step that no CI workflow runs — and it is currently red on main


macanderson:
## Problem

Two distinct defects, found while repairing main in #4079.

### 1. `tokens` is in the gate but in no workflow

`tokens` is the last entry of `GATE_GUARDS_FAST` in the `Makefile`:

```make
GATE_GUARDS_FAST := no-scratch no-secrets design-refs action-pins cargo-install-pins \
                    license-allowlist-parity repro-wiring shellcheck invariants doc-links \
                    command-docs brand-case file-size god-files gate-parity left-behind \
                    role-names stat-portability module-reachability typed-errors \
                    dead-code-allows diagnostic-codes bench-suites tokens
```

So `make gate`, `make guards-fast`, `make check` and the pre-push hook all run
it. But it appears in **no** file under `.github/workflows/`:

```
$ rg -n 'make tokens|gen-tokens' .github/workflows/
(no matches)
```

The consequence is the asymmetry `check-gate-parity.sh` exists to prevent, in
the direction that misleads: the guard blocks an honest local `git push` and
nothing server-side, so a `--no-verify` push (or any path where the hook is not
installed) lands a colour-system regression with every required check green.
`gate-parity` does not catch this — it holds `AGENTS.md` and `CONTRIBUTING.md`
to `GATE_STEPS`, not `ci.yml` to `GATE_STEPS`.

### 2. `make tokens` is red on main right now

On `6f4a7cd22`:

```
$ make tokens
tokens: 21 validated, 2 artifacts in sync
retired brand hexes found in 7 place(s). These belong to superseded kits and have no legitimate use:

  crates/stella-cli/src/export/tests.rs:703: #070B10 — brand v4.0 Obsidian ground -- green-tipped, fails neutral-gray
  crates/stella-cli/src/export/tests.rs:711: #C58A32 — brand v4.0 Bronze Gold -- fails the resting-gold green ratio
  crates/stella-cli/src/export/tests.rs:711: #674415 — brand v4.0 gold shade
  crates/stella-observatory/src/assets/index.html:28: #C58A32
  crates/stella-observatory/src/assets/index.html:95: #C58A32
  crates/stella-tui-theme/README.md:109: #C58A32
  website/src/app/tokens.css:108: #c58a32

hex literals inside token-only paths (13):

  website/src/app/tokens.css:118: #050507   website/src/app/tokens.css:125: #725a00
[...]

PR #4083, opened 2026-08-20
fix(stella-tui-theme): the generated palette is tracked but unreachable — main is red on module-reachability

macanderson:
Replaces #4079, which was cut before #4076/#4077 landed and conflicted on
`Cargo.lock`. The lockfile half of that PR is now redundant — #4076 fixed it —
so this carries **only** the half that is still red.

## What is red

`check-module-reachability` fails in `ci.yml`'s required job on the current
main tip (`04fc177be`):

```
$ python3 ./scripts/check-module-reachability.py
  crates/stella-tui-theme/src/generated.rs
check-module-reachability: FAILED
```

The file is tracked under a crate's `src/`, but no `mod` declaration reaches it
from the crate root.

## How it got that way

The crate was created **twice**, and the two creations composed cleanly into a
broken tree:

- **#4066** landed first, as exactly `lib.rs` + `generated.rs`. Its `lib.rs`
  declared `pub mod generated;` and re-exported from it.
- **#4055** landed next and replaced `lib.rs` wholesale with its own
  five-module version (`clamp`, `fallback`, `glyph`, `token`, `wordmark`) —
  a branch that had never contained a `mod generated;` line.

Git had no conflict to report, because one side simply does not contain the
other's lines. The result is the #1750 shape: rustc, rustfmt **and** clippy all
walk past the file at once, while `make tokens` carries on checking it for
staleness — and its four tests had stopped running while the suite still
reported `ok`.

## The fix

Restore the declaration. This is **#4066's own line put back**, not a new
design decision.

The crate consequently holds two copies of the palette — `generated::ALL`
against `token::ALL`. Collapsing them into one is already tracked as **#4058**
under epic **#4059**, and is deliberately not folded into a red-main repair.

### Why a plain `//` and not `///`

An outer doc comment on a `mod` declaration is merged with the module file's
own `//!` docs and resolves their intra-doc links in the **parent** file's
scope. With `///`, the restored declaration produced three `unresolved link`
errors — `Color::Rgb`, `ALL`, `Clamp` — none of which exist in `lib.rs`,
failing `cargo doc -D warnings`. Verified in both directions; the comment
records it so the next author does not re-discover it.

## Witness

```
check-module-reachability   FAILED -> OK (1028 files across 28 crates, all reachable)
cargo test -p stella-tui-theme     22 passed, including 4x generated::tests::*
                                   which could not execute at all before
cargo clippy -p stella-tui-theme --all-targets -- -D warnings   clean
RUSTDOCFLAGS="-D warnings" cargo doc -p stella-tui-theme        clean
cargo fmt --all --check                                         clean
check-lockfile-sync                                             OK on this base
```

Workspace-wide `cargo clippy --workspace --all-targets -- -D warnings` and
`cargo doc --workspace -D warnings` were both run and are clean, since the two
guard failures had been masking every compile step behind them.

No test was deleted in this change.

## Left behind, tracked not dropped

`make tokens` is also red on main, and it is a `make gate` step
(`GATE_GUARDS_FAST`) that runs in **no** workflow — filed as **#4081** with the
full hex list and the gate-parity hole.

Refs #4058, #4059, #4081

## Summary by Sourcery

Bug Fixes:
- Restore the generated palette module to the stella-tui-theme crate so the tracked artifact is reachable and its tests and checks execute.

### DIFF
--- a/crates/stella-tui-theme/src/lib.rs
+++ b/crates/stella-tui-theme/src/lib.rs
@@ -47,6 +47,21 @@
 
 pub mod clamp;
 pub mod fallback;
+// The generated face of `design/tokens/stella-tokens.json`, emitted by
+// `scripts/gen-tokens.py` and kept in sync by `make tokens`. It carries a
+// second copy of the palette — `generated::ALL` against `token::ALL` —
+// because this crate was created twice, by #4066 (generator first) and #4055
+// (hand-written table first), and the two landed without a textual conflict.
+// This declaration is #4066's, restored: without it the generated artifact is
+// tracked but unreachable, so rustc, rustfmt and clippy all walk past it
+// while `make tokens` still checks it for staleness. Collapsing the two
+// palettes into one is #4058, under epic #4059.
+//
+// Deliberately not a doc comment: an outer `///` here is merged with the
+// module's own `//!` docs and resolves their intra-doc links in *this* file's
+// scope, where `Color`, `ALL` and `Clamp` do not exist — three rustdoc
+// errors under the gate's `-D warnings`.
+pub mod generated;
 pub mod glyph;
 pub mod token;
 pub mod wordmark;


macanderson:
Cross-reference: this and #4087 fix the same failure (`check-module-reachability` on `crates/stella-tui-theme/src/generated.rs`), opened minutes apart. **Only one should merge** — both edit `crates/stella-tui-theme/src/lib.rs`, so the second will conflict.

They are not equivalent, so the choice is worth a moment:

- **This PR** declares the module. That clears the guard.
- **#4087** declares the module *and* adds two tests that make the declaration mean something: `the_hand_written_palette_agrees_with_the_generated_one` and `the_clamp_bounds_agree_with_the_generated_ones`.

The reason the second half matters is that the reachability failure is a symptom. `generated.rs` (21 tokens, plus its own copy of the clamp constants) and the hand-written `token.rs` (17 tokens) / `clamp.rs` are **two palettes describing one system**, and until #4066 they were never compared — `make tokens` was checking a file the crate did not contain. Declaring the module alone makes the file compile; it does not stop the two tables drifting apart, which is the defect that produced this.

Suggest merging #4087 and closing this, or merging this and cherry-picking #4087's two tests. Either is fine — merging both is not.

Neither PR consolidates the two tables into one generated file. That is the right end state and is a design change, not a build fix, so it is being filed separately.

macanderson:
Closing in favour of **#4087**, which fixes the same `check-module-reachability` failure and also adds the two tests that make the declaration mean something (`the_hand_written_palette_agrees_with_the_generated_one`, `the_clamp_bounds_agree_with_the_generated_ones`).

Nothing wrong with this PR — it is the correct minimal fix, and it was opened first. The reason for picking the larger one is that the reachability failure is a symptom: `generated.rs` (21 tokens, plus its own copy of the clamp constants) and the hand-written `token.rs` (17) / `clamp.rs` are two palettes describing one system, and declaring the module without comparing them leaves them free to drift apart again — which is what produced this in the first place.

Repository owner picked #4087 for that reason. Consolidating the two tables onto one generated authority is tracked in #4090.

Issue #4092, opened 2026-08-20
stella-tui-theme has two palettes and no stated winner: generated.rs is validated by make tokens, token.rs is what the TUI paints


macanderson:
## Problem

`crates/stella-tui-theme` now carries **two palettes**, both compiled, both
tested, and neither superseding the other:

| | `src/token.rs` | `src/generated.rs` |
| --- | --- | --- |
| origin | hand-written | emitted by `scripts/gen-tokens.py` from `design/tokens/stella-tokens.json` |
| `pub const` count | 18 | 25 (a superset by name — nothing in `token.rs` is absent) |
| its `ALL` table pairs a token with | `token::Role` | `clamp::Clamp` |
| who consumes it | every v2 surface, via `crate::token` | nothing |
| gate coverage | `cargo test -p stella-tui-theme` | `make tokens` (`gen-tokens.py --check`) |

`generated.rs` landed with no `mod` declaration and was invisible to rustc,
rustfmt and clippy while `make tokens` went on validating it. #4087 declared it
so it compiles. That fixed the gate; it did **not** answer which palette is
authoritative, and deciding that as a side effect of a `mod` line would have
been the wrong way to decide it.

## Why it matters

The crate's whole argument is "one crate, so there is one answer" (`lib.rs`
module doc). Two tables is two answers. Concretely:

- A token added to the JSON reaches `generated.rs` and **not** `token.rs`, so
  the palette every surface actually paints from silently does not have it.
- The anti-drift mechanism is role totality on `token::ALL`. A hex entering
  through `generated::ALL` is held to a `Clamp` instead, which is a different
  (not obviously weaker or stronger) check.
- `design/tokens/stella-tokens.json` reads as the source of truth — `make
  tokens-update` regenerates "every colour artifact" — but editing it does not
  change what the TUI paints.

## The decision

Someone with the brand context needs to pick one:

1. **`generated.rs` wins.** `token.rs` becomes a re-export of it (or is
   deleted), consumers repoint, and the `Role`/`Clamp` split is reconciled —
   `gen-tokens.py` would need to emit whatever `token::Role` totality the tests
   walk, or the tests move to the `Clamp` table.
2. **`token.rs` wins.** `generated.rs` is not a Rust artifact at all;
   `gen-tokens.py` stops emitting `RUST_OUT` and keeps only its other artifact,
   and `make tokens` validates the hand-written table against the JSON instead
   of generating a second one.
[...]

Issue #4062, opened 2026-08-20
main is red: doc-warnings (2 private intra-doc links) and format-check (1 unformatted file) — separate from #4044


macanderson:
## Problem

`make gate` on a clean `origin/main` fails two steps that no open issue names.
#4044 is open for `file-size` on the merged tree; these are separate, and each on its
own turns every PR's checks red (AGENTS.md § `main-red-hold`).

Found while running the full gate for an unrelated change (#4036). Neither failure is in a
crate that change touches.

## 1. `doc-warnings` — two intra-doc links to private items

`make doc-warnings` exits 101 with:

```
error: public documentation for `materialize_tool_calls` links to private item `refold_tool_calls`
error: public documentation for `ReadTally` links to private item `ReadState::reads_since_change`
```

- `crates/stella-store/src/tool_calls.rs:567` declares
  `pub(crate) fn refold_tool_calls(...)`, and the doc comment at line 682 links it as
  ``[`refold_tool_calls`]`` from a public item.
- `crates/stella-tools/src/staleness.rs` — same shape for
  `ReadState::reads_since_change`, linked from `ReadTally`'s public doc.

This is the third and fourth instance of the shape recorded in #3981, where `main` went red
on five of them at once. `rustdoc` is run at `-D warnings`, so each is an error rather than
a warning.

The fix is one of: make the referenced item public, drop the link and name the item in
backticks as plain code text, or move the sentence onto a non-public item. Naming it in
backticks is usually right when the reference is a reader's orientation rather than a
navigable API relationship.

## 2. `format-check` — one unformatted file

`cargo fmt --all` rewrites
`crates/stella-cli/src/plugin_cmd/configure/tests.rs` around line 257: an `assert!` whose
arguments rustfmt breaks across lines. Confirmed on the pinned toolchain
(`rust-toolchain.toml` → 1.97.0, `rustfmt 1.9.0-stable`), so it is genuinely unformatted on
`main` and not a rustfmt-version artifact of one machine.

**Already fixed in passing** by PR for #4036, unavoidably: that branch ran `cargo fmt --all`
and could not have a green `format-check` without it. Called out there rather than
extracted, because it is a three-line mechanical rustfmt change in a test file and not a
shared cell of the kind AGENTS.md warns about (`Cargo.lock`,
`scripts/file-size-baseline.txt`). If that PR does not land, this needs its own.

## Why this was not caught

[...]

PR #4093, opened 2026-08-20
fix(stella-tui-theme): declare the generated palette, which nothing compiled

macanderson:
## The defect

`check-module-reachability` fails on `main`:

```
These files are tracked under a crate's src/ but no `mod` declaration
reaches them from the crate root.
  crates/stella-tui-theme/src/generated.rs
```

`scripts/gen-tokens.py` writes that file (`RUST_OUT`) from
`design/tokens/stella-tokens.json`, and `make tokens` validates it on every gate
run:

```
$ python3 ./scripts/gen-tokens.py --check
tokens: 21 validated, 2 artifacts in sync
```

But `crates/stella-tui-theme/src/lib.rs` carried no `mod generated;`, and
`git grep generated -- crates/stella-tui-theme/` returned nothing. So rustc,
rustfmt and clippy all walked past it. **The generated palette was being
validated by one gate step and compiled by none** — the exact shape #1750 added
the reachability guard to catch, and the guard did catch it.

## The fix

One declaration, in `lib.rs`.

`pub mod generated;` rather than private: every item in the generated file is
emitted as `pub`, so a private module would make the entire thing dead code, and
silencing that with `#[allow(dead_code)]` is precisely the suppression AGENTS.md
refuses — the lint would not be wrong there.

It does not migrate anything, and the two files are **not** a file and its copy:

- `generated::ALL` pairs each token with a `clamp::Clamp`
- `token::ALL` pairs each token with a `token::Role`

Different tables, different types, 25 constants against 18. Which becomes
authoritative is a real design decision that belongs to whoever finishes the
brand work — settling it as a side effect of a `mod` declaration would be the
wrong way to decide it. Filed separately as #4092 rather than resolved here.

So after this PR the crate has two palettes: the live one (`token`) and the
generated one (`generated`), both compiled, both tested, neither pretending to
be the other.

## Witness

The guard is the evidence, and it flips:

```
$ python3 ./scripts/check-module-reachability.py     # on main
check-module-reachability: FAILED
  crates/stella-tui-theme/src/generated.rs

$ python3 ./scripts/check-module-reachability.py     # on this branch
check-module-reachability: OK — 1028 source file(s) across 28 crate(s),
all reachable from a crate root.
```

`cargo test -p stella-tui-theme` — 22 passed, 0 failed. `cargo clippy
-p stella-tui-theme --all-targets -- -D warnings` clean. `cargo fmt --all
--check` clean. No test was deleted.

## Context: this is the third independent red-main break

A red `main` reds every open PR (AGENTS.md § `main-red-hold`), so these are
costing every contributor a failure they did not cause. Currently open and
distinct:

| Break | Issue | Fix |
| --- | --- | --- |
| `file-size` — `self_driving_cmd.rs` at 1501 lines | #4044 | PR #4082 |
| `doc-warnings` + `format-check` | #4062 | (doc-warnings verified already fixed at `04fc177be`) |
| `module-reachability` — this one | #4087 | this PR |

Each hid the next: `file-size` runs before `module-reachability`, so nobody saw
this until the earlier step's failure was read past.

Closes #4087

## Summary by Sourcery

Bug Fixes:
- Declare the generated theme palette module so it is compiled and included in module reachability checks.

### DIFF
--- a/crates/stella-tui-theme/src/lib.rs
+++ b/crates/stella-tui-theme/src/lib.rs
@@ -47,6 +47,24 @@
 
 pub mod clamp;
 pub mod fallback;
+// The generated palette, emitted by `scripts/gen-tokens.py` from
+// `design/tokens/stella-tokens.json` and gate-checked by `make tokens`.
+//
+// It landed without this declaration, which made it invisible to rustc,
+// rustfmt and clippy at once while `make tokens` went on validating it — a
+// generated artifact that was checked and then compiled by nothing (#1750 is
+// the guard that caught it; `check-module-reachability` was failing on `main`).
+//
+// `pub` rather than private because every item in it is generated as `pub`:
+// declaring it privately would make the whole module dead code, and silencing
+// that with `#[allow(dead_code)]` is exactly the suppression AGENTS.md refuses
+// — the lint would not be wrong.
+//
+// It does **not** yet supersede [`token`], and the two are not interchangeable:
+// `generated::ALL` pairs each token with a [`clamp::Clamp`] while
+// `token::ALL` pairs it with a [`token::Role`]. Which table becomes
+// authoritative is a live design question, not something to settle by wiring.
+pub mod generated;
 pub mod glyph;
 pub mod token;
 pub mod wordmark;


macanderson:
Closing as a duplicate. #4083 was opened first for the same defect and makes the same one-line change to `crates/stella-tui-theme/src/lib.rs`; #4087 does it too and additionally touches `src/tests.rs`. One of those should land, not this.

One thing from this PR worth keeping regardless of which lands: `generated.rs` and `token.rs` are **not** a file and its copy — `generated::ALL` pairs each token with a `clamp::Clamp` while `token::ALL` pairs it with a `token::Role`, 25 constants against 18, and only the second is what any v2 surface actually paints from. So a token added to `design/tokens/stella-tokens.json` reaches the generated file and changes nothing on screen. Declaring the module fixes the gate and does not fix that. Filed as #4092.

Issue #4104, opened 2026-08-20
main is red: cargo doc -D warnings fails on four private intra-doc links to super::compose


macanderson:
## Problem

`main` is red. `cargo doc -D warnings` — a step of the required `fmt + clippy +
test` job — exits 101 at `2bf532b4a`:

```
error: public documentation for `bind_composed` links to private item `super::compose`
   --> crates/stella-runtime/src/wrapper/dispatch.rs:391:11
    = note: this link resolves only because you passed `--document-private-items`,
            but will break without
    = note: `-D rustdoc::private_intra_doc_links` implied by `-D warnings`
error: could not document `stella-runtime`
```

Four intra-doc links in `crates/stella-runtime/src/wrapper/dispatch.rs` name
`super::compose`, which is a private module (lines 236, 242, 379, 391).
Rustdoc resolves them only under `--document-private-items` and refuses them on
the public surface.

A red `main` reds every open PR (AGENTS.md § `main-red-hold`), so this is
costing every contributor a failure they did not cause.

## Cause, named rather than hedged

Mine, introduced by #4095. That branch was checked with `cargo clippy
--all-targets -- -D warnings` and the full `stella-runtime` test suite, and
**not** with rustdoc. `make guards-fast` plus clippy does not run `cargo doc`,
so a private intra-doc link is invisible locally right up until CI.

This is the third time this shape has reddened `main`; #3981 was five of them at
once. The lesson that keeps not sticking: **an intra-doc link is checked by a
gate step that a fast local run skips.**

## Fix

The four links become plain code spans. They were orientation for a reader
rather than navigable API relationships — `compose` is private precisely so
callers go through `bind_composed` — so naming it in backticks says the same
thing and promises no link.

One stale claim goes with them: `bind_composed`'s `# Errors` section says "one
of the four conflict variants", written before two of those four turned out to
be unreachable and were deleted in that same PR. It is two.

## Verify

```
$ RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --workspace
```

Fails on `main`, passes on the fix. `cargo fmt --all --check` and `cargo clippy
-p stella-runtime --all-targets -- -D warnings` stay clean.

## Definition of done

- `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --workspace` passes.
- No `#[allow(rustdoc::private_intra_doc_links)]` anywhere — the lint is right
  here, and suppressing it is the expedient CLAUDE.md forbids.
[...]

Issue #4103, opened 2026-08-20
bench: the posture role-model mirror outlived the engine it mirrors — a maintainer call about published numbers


macanderson:
## Problem

`bench/harbor_adapter/tests/test_assurance_tiers.py::test_the_engine_still_resolves_a_role_in_the_order_this_module_mirrors` fails with `ValueError: substring not found`, and it is the **last** failure in `harbor_adapter + analyzer pytest` once #4102 lands (#3967 covers the other four).

It is not a stale-test problem. It is the test doing its job: it pins `resolve_posture_role_model` to the Rust original, and the original changed underneath it.

`AgentEngineConfig::model_for` (`crates/stella-cli/src/settings/engine.rs:459`) is now two rungs:

```rust
pub fn model_for(&self) -> Option<&str> {
    self.agent()
        .and_then(|a| a.model.as_deref())
        .or(self.default_model.as_deref())
}
```

`agents.<…>.model` → `default_model`. No flat per-role key, and no `kind` argument at all, because core knows one role now (#3903, and the collapse in #3944).

`resolve_posture_role_model` (`bench/harbor_adapter/stella_harbor/posture.py:958`) still resolves three rungs — `agents.<role>.model` → `pipeline_<role>_model` → `default_model` — plus a `_WORKER_INHERITING_ROLES` arm where `research` and `plan` fall back to the worker rather than to `default_model`.

The test looks for `pipeline_verifier_model` inside `model_for`'s body and no longer finds it.

## Why this is a maintainer decision and not a fix

`resolve_posture_role_model` is not descriptive. Its production caller is the assurance-tier computation in the same module (`posture.py:1044-1048`), which resolves the worker's and the verifier's models and declares the run `witness-on` or `witness-off` on whether they differ. Its off-reason text quotes both models and the posture keys they came from.

So changing the resolution changes **which recorded runs count as witness-authored**. That is a claim about published benchmark numbers, which is the decision #3870 reserves for a maintainer, and the same reason `unknown.rs` keeps `RETIRED_ENGINE_ROOT` recognized rather than dropping it.

There is a second, deeper question underneath: the predicate the docstring cites as its authority — `Pipeline::witness_author_independence` — is in `stella-pipeline`, which #3865 **deleted from this workspace**. So this module mirrors an engine that no longer exists, not merely one that moved.

## The options, with what each costs

[...]

Issue #3910, opened 2026-08-19
roleless-core slice 6: retire the role-name check script and the four-language role contract


macanderson:
Slice 6 of #3903. Spec: `doc:roleless-core` §6. **Only after slice 2 lands.**

## Problem

`scripts/check-role-names.sh` is a `make gate` step (`GATE_STEPS`) that pins
four agent-config role spellings across **four languages** — Rust, Python, JS —
with `role_key()` in `crates/stella-cli/src/config_wiring.rs` as the normative
home. Around 239 non-Rust files name `verifier`/`triage`/`witness`:
`arenabench/`, `bench/`, `crates/stella-observatory/src/assets/index.html`
(which filters agents by a hardcoded literal list), and `website/content/`.

The guard exists for a real and expensive failure: #1394 renamed `judge` →
`verifier`, the Rust compiler caught every Rust call site and none of the
others, and `arenabench/arenabench/config.py` silently kept translating the old
name. No error, no failing test — the seat ran with the verifier inheriting the
worker baseline while the scoreboard reported it as configured. **The contest
published a number for a pairing it never ran.**

So this guard cannot simply be deleted. Its hole has to be closed by
construction first.

## Why slice 2 is the prerequisite

Slice 2 makes the role name **travel with the run as data** in the trace. Once a
consumer reads the name that actually ran instead of copying a spelling, there
is nothing left for a guard to pin — the desynchronization the guard detects
becomes impossible rather than detected.

Doing this slice first would re-open #1394's failure mode with no detector.

## What to do

- Repoint `arenabench/` and `bench/terminal_bench_analysis/` to read role names
  from the trace.
- Repoint `crates/stella-observatory/src/assets/index.html`'s hardcoded filter
  list at the same data.
- Delete `scripts/check-role-names.sh` and its `GATE_STEPS` entry, and update
  the `AGENTS.md` / `CONTRIBUTING.md` gate blocks in the same PR
  (`scripts/check-gate-parity.sh` enforces that they match).

## Witness

A bench analyzer fed a trace containing an unfamiliar role name **reports it**
rather than dropping the row. That is the property the old guard could not give:
it could only check that four known names agreed.

## Done when

No role spelling is duplicated across languages, so there is nothing for a guard
to pin.

## Definition of done

The done conditions this issue already states, as boxes the gate can read.
Nothing here is new scope, and nothing is ticked by the migration — the
gate reads one heading and this issue used another (macanderson/oxagen#1407).

[...]

Issue #3908, opened 2026-08-19
roleless-core slice 4: delete EngineAgentKind and the inert pipeline_<role>_model keys


macanderson:
Slice 4 of #3903. Spec: `doc:roleless-core` §6.

## Problem

The core settings surface advertises six model personas — `default`, `worker`,
`verifier`, `triage`, `research`, `plan` — for a core loop that has none of them.
They are pins on the staged pipeline deleted in #3865, and they are **inert**:

- `resolve_engine_wiring` (`crates/stella-cli/src/agent/engine.rs`) reads
  `pipeline_triage_model` / `pipeline_verifier_model` / plan / research and pins
  them into a `RoleTable`.
- The only live `Router::resolve` call sites are `SessionFallback::resolve_fallback`
  (`Role::Worker` only) and `resolve_cross_family_verifier` — and the latter
  builds its router with `RoleTable::new()`, i.e. **empty pins**.

So a user can set `pipeline_verifier_model`, and nothing reads it — including
the one place a second model actually runs (the `stella goal` cross-family
verifier, which selects by family and ignores the pin entirely).

A setting that reads like a capability and is not one is worse than no setting.

## What to do

- Delete `EngineAgentKind` (276 refs), `pipeline_<role>_model` (192 refs), and
  the `agents.<persona>` map.
- `agent_engine_config` keeps: `default_model`, `allowed_models`, `auto_mode`,
  `effort_auto`, `reasoning_auto`, `model_output_caps`, and `seat_models` (added
  in slice 0).
- `allowed_models` becomes the ceiling on seat assignments too.

## Deprecation, not silent removal

The retired keys must be **recognized, ignored, and reported once by name**,
with the seat assignment that replaces them. These keys currently read like
capabilities; a silent drop leaves them reading that way in every settings file
that has one. `crates/stella-cli/src/settings/unknown.rs` is where unknown keys
are already handled.

## Also in scope

- The stale operator-facing text in `push_same_model_notice`
  (`crates/stella-cli/src/agent/engine.rs`), which tells the user "The pipeline
  still runs; set `pipeline_verifier_model` (or `agents.verifier.model`)" —
  advice for a deleted subsystem and a key that steers nothing.
- The stale module doc on `crates/stella-core/src/router.rs`, which attributes
  `FallbackInfo` handling to "the pipeline layer (`stella-pipeline`)" — a crate
  deleted in #3865.

## Witness

A settings file carrying every retired key loads, warns by name, and changes no
behavior.

[...]

PR #4109, opened 2026-08-20
fix(bench,arenabench,stella-autonomy): three defects reddening main, each hiding the next (#4110)

### LINKED ISSUE #4110: main is red: cargo doc -D warnings fails on a private intra-doc link in stella-autonomy's gate.rs
## Problem

`main` is red. `cargo doc -D warnings` — a step of the required `fmt + clippy +
test` job — exits 101 at `29d1eb14d` and every commit since:

```
```

`crates/stella-autonomy/src/gate.rs:4` links `[`crate::deliver`]`. That module
is private (`mod deliver;`, `lib.rs:45`); its items are public only through the
re-export at `lib.rs:66`. Rustdoc resolves such a link only under
`--document-private-items` and refuses it on the public surface.

Introduced by #4097 (`29d1eb14d`). `f3d2cbe2e`, the commit before it, is green.

A red `main` reds every open PR (AGENTS.md § `main-red-hold`), so this is
costing every contributor a failure they did not cause.

## Fix

Point the link at the public re-export instead of the private module:

```rust
//! [`deliver_next`](crate::deliver_next) decides what to do about a red build;
```

`deliver_next` is the accurate referent — it is the function that decides what
to do about a red build — and `crates/stella-autonomy/src/doctrine.rs:22`
already uses exactly this form for the same item, so the fix matches a working
pattern in the same crate rather than inventing one.

## This is the fourth instance of this shape

#3981 was five at once. #4104 was four more, in
`crates/stella-runtime/src/wrapper/dispatch.rs`, two hours ago. This is the
same lint (`rustdoc::private_intra_doc_links`) with the same cause.

**The reason it keeps happening is worth writing down, because "run the check"
is not it.** The two rungs people actually run locally — `make guards-fast` and
`make check` — compile no rustdoc at all. And when `cargo doc` *is* run, it
will happily report success **from cache without rebuilding**, so a run right
after an unrelated build prints `Finished` and nothing else. That false green
is what let #4104 through after I had explicitly checked it.

What actually catches it:

```
$ touch <the file you edited>
$ RUSTDOCFLAGS="-D warnings" cargo doc --no-deps -p <crate>
```

and then **reading the output for `Documenting <crate>`** — not for `Finished`.

## Verify

```
$ RUSTDOCFLAGS="-D warnings" cargo doc --no-deps -p stella-autonomy
```

Fails on `main`, passes on the fix, and re-documents the crate rather than
reporting a cached success.

## Definition of done

- `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --workspace` passes.
- No `#[allow(rustdoc::private_intra_doc_links)]` — the lint is correct here.
- Worth considering separately: whether `make check` should include
  `cargo doc`. It costs a documentation build, so it is a real trade and
  belongs in its own issue rather than being folded into a red-main repair.
  Four instances in two days is the evidence for opening it.

macanderson:
## What & why

`main` is red, on **three** independent defects in two required jobs. Each one hid the next, so they are fixed together and in the order they surfaced.

### 1. `harbor_adapter + analyzer pytest` — the role-model mirror outlived its original

Red on **every** push to `main` since #3944. #3908 collapsed `AgentEngineConfig::model_for` to the one role core has — `agents.default.model` > `default_model` — and `test_the_engine_still_resolves_a_role_in_the_order_this_module_mirrors` looks for `pipeline_verifier_model` inside its body, so it raises `ValueError: substring not found` and asserts nothing at all.

This ships **#4103's option 2, and only that**: restate what the mirror mirrors, change no resolution behaviour. `posture.py`'s entire diff is its docstring — `resolve_posture_role_model` still resolves the same rungs for the same postures, so no recorded run is re-characterised and the published-numbers decision #3870 reserves for a maintainer is untouched. **Options 1 and 3 stay open, and so does #4103.** I am not the one who should choose between "collapse the mirror" and "retire the tier"; this only stops `main` being red while that is decided.

The mirror is repointed, not weakened:

- the surviving pin asserts the order of the two rungs both sides still share — the #2134 reorder hazard on everything that is left;
- a new test asserts the extra rung is *exactly* the retired set: `model_for` reads none of the five flat `pipeline_<role>_model` keys, **and** `RETIRED_ENGINE_ROOT` still names all five. The divergence is legitimate only while both halves hold, and slice 6 (#3910) closes it from the Python side. `RETIRED_ENGINE_ROOT` is parsed out of the Rust rather than copied, so it cannot drift in the direction that costs — this harness writing a key the launcher has begun refusing, discovered at launch with the posture already hashed.

**And the reason it reached `main` is closed.** `bench.yml`'s PR filter names `crates/stella-model/src/catalog.rs` because a suite reads it, but did not name `crates/stella-cli/src/settings/`, which these mirrors parse. #3908 touched no filtered path, so its PR skipped the suites and went green; a push to `main` runs them unconditionally, so it reddened from the merge onward. The filter now names the module and a third test asserts it does — matched against the `grep -Eq` expression rather than the whole file, because a whole-file search would pass on the rationale comment alone.

### 2. `make bench-test` could not start on a Mac — and was hiding a second red suite

Found by trying to *prove* (1) with the local command this PR claims. `scripts/bench-suites.sh` expands each suite's pytest arguments as `"${argv[@]}"` under `set -u`. In bash 3.2 — the `/bin/bash` every macOS ships, and the one the pre-push hook runs — an **empty** array's `[@]` expansion is a fatal unbound-variable error. Two of seven suites take no arguments and one is first, so the command died on line one. Same class as #2847 one level down: the local command existed and could not start.

With it running, `arenabench/tests/` turns out to fail **collection entirely**: `NameError: name 'Path' is not defined`. #3950 removed the assertion that `_ROLE_KEY_RS`, `_ROLE_ARM` and `_role_key_names()` served, and their `re`/`pathlib` imports, but left the three module-level orphans behind. Red since 2026-08-19, unseen because the bench job's steps are sequential — `harbor_adapter` fails first, so the arenabench step never ran. **One red step masks every step after it.**

The three are deleted, not repaired, for #3950's own two reasons plus a third it could not have known: #3908 deleted `role_key()` and `EngineAgentKind` and repointed `check-role-names.sh` at `ENGINE_AGENT_NAMES`/`RETIRED_ENGINE_AGENT_NAMES`. A repaired reader would parse a function that no longer exists and silently return the empty set — a green assertion over nothing.

### 3. `fmt + clippy + test` — `cargo doc -D warnings` fails on a private intra-doc link

Separate job, separate defect, surfaced by the `main-red-hold` check on this PR naming **#4110**. `crates/stella-autonomy/src/gate.rs:4` links `[`crate::deliver`]`; that module is private and its items are public only via the re-export at `lib.rs:66`. Introduced by #4097. Repointed at `[`deliver_next`](crate::deliver_next)`, which is both public and the accurate referent, and is exactly how `doctrine.rs:22` already spells the same item.

Closes #4110
Refs #4103
Refs #3910

## The witness

| Witness | On `main` | Here |
|---|---|---|
| `test_the_engine_still_resolves_a_role_in_the_order_this_module_mirrors` | `ValueError: substring not found` | passes |
| `test_the_bench_workflow_runs_when_the_settings_module_changes` | fails — `main`'s filter does not name the module | passes |
| `make bench-test` (macOS) | `bench-suites.sh: line 186: argv[@]: unbound variable` | runs all seven suites |
| `arenabench` collection | `NameError: name 'Path' is not defined` | collects to 100% |
| `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps -p stella-autonomy` | `could not document stella-autonomy` | `Documenting stella-autonomy` → `Finished` |

`test_the_flat_role_keys_this_module_reads_are_retired_by_the_engine` is not a witness for this change — it is the guard for the gap that let #3908 through, and would have failed *before* #3908 for the right reason.

The rustdoc witness was run the way #4110 prescribes — `touch` the file first, then read the output for `Documenting`, not for `Finished` — because `cargo doc` reports a cached success without rebuilding, and that false green is what let #4104 through after an explicit check.

```
bench/harbor_adapter:  719 passed, 2 skipped
RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --workspace:  exit 0, 29 crates
```

**Two local failures are the machine, not the code, and are named rather than hidden:** the two `test_adapter` failures on a box carrying a real `VOYAGE_API_KEY`/`OPENROUTER_API_KEY` are #3668; the five `arenabench/tests/test_sut.py` failures are `stella_seat_problem()` correctly refusing because `/usr/bin/python3` is 3.9.6 here and `tomllib` landed in 3.11. CI's ubuntu runner is 3.12. The refusal is the guard working. CI is the authority for those five and I did not run them green locally.

## The gate

- [x] `make guards-fast` (exit 0)
- [x] `make bench-test` — now that it can run
- [x] `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --workspace` (exit 0)
- [x] `shellcheck scripts/bench-suites.sh` (exit 0)
- [x] Docs updated where behaviour changed — the two stale docstrings claiming a key-for-key mirror are corrected in the same diff
- [x] `Closes #4110` in this description **and** as a commit trailer

## Nothing left behind

- [x] There is nothing new: everything I noticed is already tracked.

Checked against existing issues rather than duplicated:

- the mirror-vs-engine decision itself — **#4103**, deliberately left open, because option 2 does not answer it
- the fail-closed tightening this unblocks — **#3910**
- the two adapter tests that read ambient provider credentials — **#3668**
- the red `main` this closes — **#4110**

## Ground-rule check

**The judgement call worth reviewing is that I shipped #4103's option 2 without a maintainer.** #4103 explicitly reserves the choice, so the thing to check is that this really is behaviour-neutral: `git diff` on `posture.py` should be docstring-only. If it is, no arm changes and no digest re-hashes, and the maintainer question is exactly as open as it was.

**Three defects in one PR is a deliberate exception to one-logical-change.** They are causally chained — (2) is what let me prove (1), and fixing (2) is what revealed the arenabench half; (3) was surfaced by the hold check on this PR and blocks the same required job set. Splitting them would mean landing a fix whose evidence I could not run.

The ERE is worth a second pair of eyes: `crates/stella-cli/src/settings(\.rs$|/)` matches `settings.rs` and `settings/*.rs` and deliberately not `settings_other.rs`; verified against those three plus `agent.rs` and `catalog.rs` before committing.

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -79,8 +79,19 @@ jobs:
           # pinning that let #2182 through one level up — the sweep asserts
           # this pattern covers every caller it reads, so the next one cannot
           # land outside the filter unnoticed.
+          #
+          # `crates/stella-cli/src/settings/` is the second Rust path in scope
+          # for the catalog's reason, added after it cost a red `main`.
+          # `test_assurance_tiers.py` parses `AgentEngineConfig::model_for` and
+          # `RETIRED_ENGINE_ROOT` out of that module, because the posture
+          # module mirrors the engine's model-resolution ladder and a mirror
+          # that drifts is #2134. #3908 collapsed the ladder to one role,
+          # touched no path named here, and so ran none of these suites on its
+          # own PR — then reddened every push to `main` from the merge onward,
+          # since a push runs them unconditionally. The tests assert this
+          # pattern still names the module they read.
           if git diff --name-only "$BASE_SHA"...HEAD \
-            | grep -Eq '^(bench/|arenabench/|crates/stella-model/src/catalog\.rs$|\.github/workflows/bench\.yml$)'; then
+            | grep -Eq '^(bench/|arenabench/|crates/stella-model/src/catalog\.rs$|crates/stella-cli/src/settings(\.rs$|/)|\.github/workflows/bench\.yml$)'; then
             echo "changed=true" >> "$GITHUB_OUTPUT"
           else
             echo "changed=false" >> "$GITHUB_OUTPUT"
--- a/arenabench/tests/test_responsibilities.py
+++ b/arenabench/tests/test_responsibilities.py
@@ -33,28 +33,25 @@ from arenabench.model import (
     ResponsibilityConfig,
 )
 
-#: Stella's normative home for the role vocabulary, relative to a checkout.
-#: ``scripts/check-role-names.sh`` names this same function as "the truth" and
-#: reads its match arms the same way, so the two halves of the contract are
-#: anchored to one file rather than to each other.
-_ROLE_KEY_RS = Path("crates") / "stella-cli" / "src" / "config_wiring.rs"
-
-#: One match arm of ``role_key()``: ``EngineAgentKind::Verifier => "verifier",``
-_ROLE_ARM = re.compile(r'EngineAgentKind::\w+\s*=>\s*"([a-z_]+)"')
-
-
-def _role_key_names(source: str) -> frozenset[str]:
-    """Every role spelling ``role_key()`` returns, read out of the Rust.
-
-    Scoped to that function's body — a bare sweep for the arm shape would also
-    collect every other ``match kind`` in the file, and those are legitimately
-    partial (``model_source`` skips ``Default``). The body ends at the first
-    column-zero ``}``, which is how ``check-role-names.sh``'s awk delimits it.
-    """
-    body = re.search(r"pub fn role_key\b.*?\n\}", source, re.DOTALL)
-    if body is None:
-        return frozenset()
-    return frozenset(_ROLE_ARM.findall(body.group()))
+# The reader that used to live here — `_ROLE_KEY_RS`, `_ROLE_ARM` and
+# `_role_key_names()`, which parsed `role_key()`'s match arms out of
+# `crates/stella-cli/src/config_wiring.rs` — is deleted, finishing what #3950
+# started. That PR removed the assertion those three existed to serve and the
+# `re`/`pathlib` imports they needed, but left the three behind, so importing
+# this module raised `NameError: name 'Path' is not defined` and NOTHING in
+# `arenabench/tests/` could be collected. It went unseen because the bench
+# workflow's arenabench step never ran: the `harbor_adapter` step above it had
+# been failing since #3944, and one red step masks every step after it.
+#
+# They are deleted rather than repaired, for #3950's own two reasons plus a
+# third it could not have known. The property is still enforced by a gate step
+# (`scripts/check-role-names.sh`), so repointing would be a fourth copy of a
+# passing check; re-creating a role roster for core to publish pulls against
+# `doc:roleless-core` (#3903); and the subject itself is gone — #3908 deleted
+# `role_key()` and `EngineAgentKind`, and repointed that gate at
+# `ENGINE_AGENT_NAMES`/`RETIRED_ENGINE_AGENT_NAMES` in `settings::unknown`. A
+# repaired reader would parse a function that no longer exists and silently
+# return the empty set.
 
 
 HEADER = """
--- a/bench/harbor_adapter/stella_harbor/posture.py
+++ b/bench/harbor_adapter/stella_harbor/posture.py
@@ -960,15 +960,28 @@ def resolve_posture_role_model(
 ) -> tuple[str, str]:
     """Resolve one role's model from a posture, and name the key it came from.
 
-    Mirrors ``AgentEngineConfig::model_for`` (``crates/stella-cli/src/settings.rs``)
-    key for key: ``agents.<role>.model`` outranks the flat ``pipeline_<role>_model``,
-    which outranks ``default_model`` — except for the two roles that inherit the
-    worker instead on that last rung (``_WORKER_INHERITING_ROLES``, which is the
-    engine's own split between ``model_spec_for`` and ``own_model_spec_for``, not
-    an approximation of it). Mirroring it *exactly* is the whole point —
-    a declaration that resolves roles by a different rule than the engine does is
-    free to disagree with the run it describes, which is the shape of #2134 and of
+    Mirrors ``AgentEngineConfig::model_for``
+    (``crates/stella-cli/src/settings/engine.rs``): ``agents.<role>.model``
+    outranks the flat ``pipeline_<role>_model``, which outranks
+    ``default_model`` — except for the two roles that inherit the worker
+    instead on that last rung (``_WORKER_INHERITING_ROLES``, which is the
+    engine's own split between ``model_spec_for`` and ``own_model_spec_for``,
+    not an approximation of it). Mirroring it is the whole point: a declaration
+    that resolves roles by a different rule than the engine does is free to
+    disagree with the run it describes, which is the shape of #2134 and of
     #1147 before it.
+
+    **The middle rung is knowingly ahead of the engine.** #3908 collapsed
+    ``agent_engine_config`` to the one role core has, so ``model_for`` now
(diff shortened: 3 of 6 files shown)


macanderson:
This is the one to merge, and it is better than the three I opened in parallel for the same reds (#4116, #4114, #4115 — closing those in favour of this).

I reached the same three defects independently and the same conservative conclusion on the contentious one, which is worth saying because it is the part that could have gone wrong quietly:

- **`resolve_posture_role_model` is left alone.** This PR's `posture.py` change is docstring-only — it keeps the flat rung and `_WORKER_INHERITING_ROLES` and re-points the path to `settings/engine.rs`. That is right. Collapsing the mirror to match the engine's surviving two rungs would change which *already-recorded* runs resolve to an independent verifier, and therefore which count as `witness-on` — a claim about published numbers, which is the decision #3870 reserves for a maintainer. I filed #4103 with the three options costed; it applies to this PR too and stays open after it.

- **No `xfail`, no `skip`.** Checked the diff for both; neither appears. The bench job goes green because the defects are fixed, not because a test stopped asserting.

**The thing this has that mine did not, and the reason it supersedes them:** the `bench.yml` path-filter fix. `crates/stella-cli/src/settings/` was not in scope, so #3908 and #3944 changed the model-resolution ladder, ran none of these suites on their own PRs, and then reddened every push to `main` from the merge onward. My three PRs each fixed a symptom; this fixes the mechanism that let all three land. That is the difference between `bench` being green today and it being green the next time someone touches the engine's settings module.

Two things I verified locally that may save a re-run, both on a tree carrying these same three fixes:

- the six suites that have been **skipped** behind the fail-fast step since #3944 all pass once it clears — `terminal_bench_analysis` 270, `telemetry_store` 10, `evidence` 66, `trace_triage` 90, `frontier` 27, and `arenabench` exit 0. So nothing further is hiding behind them.
- `harbor_adapter` reaches `717 passed, 2 skipped`.

One caveat for anyone reading a local run of `arenabench`: five `tests/test_sut.py` tests fail on macOS because they shell out to `/usr/bin/python3`, which is 3.9.6 there, and the adapter needs `tomllib` (3.11+). `ubuntu-latest` has 3.12 at that path, so CI does not see them. Not this PR's, and not the tree's.

macanderson:
**Composition check** — this PR plus #4120 and #4121 dry-merged onto `main` (`6ce03c3b3`) and run as one tree, because three individually-green PRs composing into a red `main` is a failure this repo has recorded more than once.

All three merge cleanly, and the result is green:

| Check | Result |
|---|---|
| `make guards-fast` | clean — incl. `tokens: no retired hex in the tree` and `check-bench-suites: OK — 7 suites in bench.yml, all run by make bench-test` |
| `make doc-warnings` | clean |
| `harbor_adapter` pytest | **719 passed, 2 skipped** |
| `arenabench` pytest | **exit 0** |
| `website` `pnpm test` | **26 passed, 0 failed** |

The `bench-suites` row is the one worth pointing at: this PR edits **both** `.github/workflows/bench.yml` and `scripts/bench-suites.sh`, and #4120 makes `check-bench-suites` a required CI step — so the two changes touch opposite sides of the same guard. They agree.

Not re-run locally: `cargo test --workspace`. One CLI test (`goal_wrapped_dispatch_cli`) is not hermetic — it strips the provider API-key env vars but not `STELLA_HOME`, so on a developer machine with `~/.stella/credentials.toml` it makes a **real billed model call** and fails after the round cap. It passes in CI, which has no credentials; main's own `ci` run shows `cargo test` green with only `cargo doc` red. Flagging it as a defect worth its own issue rather than as a result.

Suggested order: this one first (it clears both `ci` and `bench`), then #4120 (its job currently shows *this* PR's `gate.rs` error as inherited), then #4121. #4111 and #4113 are subsets of this and can be closed.

PR #4113, opened 2026-08-20
fix: main is red on three checks across two jobs — a private-module doc link, a retired-signature mirror test, and an uncollectable arenabench suite

macanderson:
Two independent reds on `6ce03c3b3`, neither caused by the other. `Vercel` is
also red, but with `Account is blocked` — unrelated to this tree.

| Check | Cause |
|---|---|
| `fmt + clippy + test` | `cargo doc -D warnings`: public docs link a private module |
| `harbor_adapter + analyzer pytest` | a mirror test pins a signature #3908 retired |

## 1. A public item's docs link a private module

`crates/stella-autonomy/src/gate.rs`'s module doc linked ``[`crate::deliver`]``,
but `deliver` is a private `mod` (`lib.rs:45`):

```
```

Rustdoc rejects this under `--document-private-items` as well, which is the
flag CI passes — so the flag does not mask it here. Demoted to a plain
backtick reference: the sentence names which module owns the decision, it is
not a link a reader could follow.

## 2. The mirror test pins a signature that no longer exists

`test_the_engine_still_resolves_a_role_in_the_order_this_module_mirrors`
introspects `AgentEngineConfig::model_for` and searched for
`self.agent(kind).and_then(|a| a.model` and `pipeline_verifier_model`.
Neither substring exists any more, so it died on `ValueError: substring not
found` before asserting anything.

**This is stale maintenance, not a live regression.** #3908/#3944
(`doc:roleless-core` §6 slice 4) collapsed the six model personas to the one
role core actually has, so `model_for` lost its `kind` parameter and the flat
per-role rung with it. Per `settings/engine.rs`'s own module doc, four of
those five keys were *already inert before the collapse* — "a user could set
`pipeline_verifier_model` and nothing read it".

It went unnoticed because `bench.yml` is path-filtered: it needs a commit
touching `bench/**` to run at all, and #4102 was the first one since #3944.

### What it asserts now

The two rungs the engine really has — `agents.default.model` >
`default_model` — plus two assertions that keep this honest rather than
merely green:

- `model_for` does **not** read a flat per-role key again (if the per-role
  rung returns, the mirror's middle rung is live and the test should pin its
  position, not its absence);
- every key the harness still writes is declared in `RETIRED_ENGINE_ROOT`.

`resolve_posture_role_model` **does** still rank the flat key, and that is
deliberate and tracked, not drift: `settings/unknown.rs` records that the
fail-closed tightening lands with **slice 6 (#3910)** "once the Python stops
writing them". When #3910 lands, the second assertion is what fails, and it
names `resolve_posture_role_model` as the thing to change. That is why this
PR does not migrate the harness itself — #3910 owns that, and doing it here
would change ArenaBench's role-model attribution inside a red-main repair.

## Witness

```
uv run pytest tests/test_assurance_tiers.py     1 failed, 9 passed  ->  10 passed
RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps \
  --document-private-items --keep-going          error  ->  clean
make guards-fast                                 all 32 guard steps OK
```

Full `harbor_adapter` suite locally: 715 passed, 2 failed — both failures are
this machine's environment leaking into `tests/test_adapter.py`
(`assert 'OPENROUTER_A...OYAGE_API_KEY' == 'OPENROUTER_API_KEY'`), and CI
reported exactly one failure against 717 passed. Counts reconcile at 719 both
places, so those two are local-only and not touched here.

No test was deleted. One was renamed —
`..._resolves_a_role_...` to `..._resolves_a_model_...` — because the engine
no longer resolves per role.

Refs #3908, #3910

## Summary by Sourcery

Restore green checks by correcting the invalid Rust documentation reference and aligning the ArenaBench mirror assertions with the current engine configuration model.

Bug Fixes:
- Fix documentation warnings caused by public gate documentation referencing a private module.
- Update the ArenaBench engine mirror test for the current single-role model resolution behavior and preserve checks for retired engine keys.

Enhancements:
- Strengthen the cross-language model-resolution consistency checks against the engine’s current model hierarchy and planned retired-key migration.

Tests:
- Rename and revise the stale role-resolution mirror test to validate current model resolution and retired-key declarations.

### DIFF
--- a/arenabench/tests/test_responsibilities.py
+++ b/arenabench/tests/test_responsibilities.py
@@ -33,30 +33,6 @@ from arenabench.model import (
     ResponsibilityConfig,
 )
 
-#: Stella's normative home for the role vocabulary, relative to a checkout.
-#: ``scripts/check-role-names.sh`` names this same function as "the truth" and
-#: reads its match arms the same way, so the two halves of the contract are
-#: anchored to one file rather than to each other.
-_ROLE_KEY_RS = Path("crates") / "stella-cli" / "src" / "config_wiring.rs"
-
-#: One match arm of ``role_key()``: ``EngineAgentKind::Verifier => "verifier",``
-_ROLE_ARM = re.compile(r'EngineAgentKind::\w+\s*=>\s*"([a-z_]+)"')
-
-
-def _role_key_names(source: str) -> frozenset[str]:
-    """Every role spelling ``role_key()`` returns, read out of the Rust.
-
-    Scoped to that function's body — a bare sweep for the arm shape would also
-    collect every other ``match kind`` in the file, and those are legitimately
-    partial (``model_source`` skips ``Default``). The body ends at the first
-    column-zero ``}``, which is how ``check-role-names.sh``'s awk delimits it.
-    """
-    body = re.search(r"pub fn role_key\b.*?\n\}", source, re.DOTALL)
-    if body is None:
-        return frozenset()
-    return frozenset(_ROLE_ARM.findall(body.group()))
-
-
 HEADER = """
 [match]
 id = "ablation"
--- a/bench/harbor_adapter/tests/test_assurance_tiers.py
+++ b/bench/harbor_adapter/tests/test_assurance_tiers.py
@@ -45,6 +45,7 @@ from stella_harbor import (  # noqa: E402 - after importorskip by design
     assurance_tiers_from_posture,
 )
 from stella_harbor.posture import (  # noqa: E402 - after importorskip by design
+    _FLAT_ROLE_MODEL_KEY,
     resolve_posture_role_model,
 )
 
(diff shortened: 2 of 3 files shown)


macanderson:
Pushed a third fix: `cf815bd68`.

Clearing the harbor_adapter failure made the **next step in the same job**
visible — the classic "one red step masks the next", this time inside a
single job rather than across the gate:

```
arenabench — pytest (until ejection)
  ERROR collecting tests/test_responsibilities.py
  E   NameError: name 'Path' is not defined
  Interrupted: 1 error during collection      (exit 2)
```

`test_responsibilities.py` uses `Path` at module scope and `re` two lines
later, importing neither. #3950 dropped the roster assertion that read a
crate deleted in #3865 and correctly removed the imports with it — but left
behind the only three names that assertion consumed: `_ROLE_KEY_RS`
(referenced nowhere), `_ROLE_ARM` (read only by the next one), and
`_role_key_names` (called nowhere).

Deleted rather than re-imported: adding `from pathlib import Path` back would
keep dead code alive purely to satisfy a `NameError`. No coverage is lost —
the Rust↔Python role-vocabulary contract those helpers mirrored is enforced
server-side by `scripts/check-role-names.sh`, which reads the same
`role_key()` match arms and reports *"6 role(s) [default plan research triage
verifier worker] consistent across every producer"*.

Witness, using CI's exact command:

```
uv run --with pytest --no-project pytest -q tests/test_responsibilities.py
#  before: exit 2, 1 collection error
#  after:  15 passed
```

**On the 5 `tests/test_sut.py` failures a macOS checkout still shows** — they
are not this and will not appear on CI. The error names its own cause:
`AdapterUnavailableError: /usr/bin/python3 cannot build a Stella seat …
ModuleNotFoundError: No module named 'tomllib'`. macOS ships
`/usr/bin/python3` 3.9.6 (verified on this machine); `tomllib` is stdlib from
3.11, which the ubuntu runner's system interpreter has. Worth its own issue
as a developer-experience defect, not a change to this PR.

Issue #3870, opened 2026-08-19
Five settings knobs are now fully inert after the staged-pipeline removal — retire or redesign


macanderson:
## What

`crates/stella-pipeline` was deleted from the workspace (#3865), which
removed the last production consumer of five `settings.json` knobs. All five
still parse, merge across the 3-scope chain, and resolve into `Config`
fields — but nothing downstream reads the resolved value to do anything.
Verified for each (grep for the resolved `Config`/accessor field outside
config-resolution plumbing and tests):

- **`enable_recap`** (`Settings::enable_recap` → `Settings::recap_enabled()`
  → `Config::enable_recap`). Its only consumer,
  `plain::recap_panel` (rendered a pipeline verdict recap), referenced
  `stella_pipeline::PipelineStatus`/`Verdict` directly and was deleted along
  with the crate — it could not have compiled otherwise. `Config.enable_recap`
  now has zero readers outside `config.rs`/`config/reload.rs` (which write
  it), `config/tests.rs`, and `settings/tests.rs`.
- **`trace_capture`** (`Settings::trace_capture` → `trace_capture_enabled()`
  → `Config::trace_capture`). Its consumer was `crates/stella-cli/src/trace.rs`
  (the #1042 trajectory-capture feature), deleted this same slice as a
  separately-orphaned 1201-line module — confirmed fully unreferenced by
  anything else before deletion. Same zero-reader shape as `enable_recap`.
- **`pipeline_require_diff_coverage`** (`EngineSettings::pipeline_require_diff_coverage`).
  Its accessor, `pipeline_require_diff_coverage_on`, was deleted outright in
  this slice (zero test callers, zero production callers) — the field
  itself still parses and merges but nothing reads it at all now.
- **`approval_wait_secs`** (`EngineSettings::approval_wait_secs`). Its
  accessor, `approval_wait`, was deleted the same way — zero remaining
  readers of the field.
- **`headless_scope_bypass`** (`EngineSettings::headless_scope_bypass` →
  `headless_scope_bypass_on()`). Kept with `#[allow(dead_code)]` in this
  slice specifically *because* it still has real test callers
  (`crates/stella-cli/src/settings/engine.rs`) exercising the merge/resolve
  behavior — but no production code branches on the resolved value; the
  scope-review gate it used to control (`daemon/approval.rs`'s scope-gate
  arm, `DeckApprovalGate`) is gone.

## Why this wasn't resolved in the same change

[...]

PR #4114, opened 2026-08-20
fix(bench): the model_for mirror asserts what is still true, and names what is not

### LINKED ISSUE #3967: bench: harbor_adapter pytest red since #3944 — 5 posture tests assert the pre-collapse key vocabulary, and it masks the arenabench step
## Problem

`harbor_adapter + analyzer pytest` is red on `main`, and has been since
**#3944** ("the config collapse — one role, named deprecations, and a
repointed guard"). Five tests in `bench/harbor_adapter/tests/` assert against
the launcher's settings vocabulary, and that vocabulary changed under them:

```
FAILED tests/test_assurance_tiers.py::test_the_engine_still_resolves_a_role_in_the_order_this_module_mirrors
  ValueError: substring not found
FAILED tests/test_posture.py::TestBenchmarkPosture::test_witness_arm_root_keys_stay_inside_the_launcher_vocabulary
  AssertionError: the trusted launcher seam would refuse this posture: {'pipeline_verifier_model'}
FAILED tests/test_posture.py::TestWorkerEffortAndTriageArms::test_all_three_roles_coexist_inside_the_launcher_vocabulary
FAILED tests/test_posture.py::TestRetiredAttemptCountArms::test_selected_knobs_stay_inside_the_launcher_vocabulary
FAILED tests/test_role_arms.py::TestTheArmSurvivesEverySeamItCrosses::test_an_armed_posture_stays_inside_the_launcher_vocabulary
```

`pipeline_verifier_model` is named explicitly in the second one, which is the
shape of the whole set: the Python bench still emits and asserts the
pre-collapse key vocabulary.

## Bisect

Clean, two commits:

```
b4da79343  (#3935)  harbor_adapter + analyzer pytest  -> success
b4cf98740  (#3944)  harbor_adapter + analyzer pytest  -> failure
```

## Why this is worth its own issue: it MASKS another fix

The job runs its suites as sequential steps and is fail-fast, so
`harbor_adapter — sync (locked) + pytest` failing means every later step is
**skipped**, including `arenabench — pytest (until ejection)`:

```
failure  harbor_adapter — sync (locked) + pytest
skipped  terminal_bench_analysis — sync (locked) + pytest
skipped  telemetry_store — pytest
skipped  evidence tooling — pytest
skipped  trace triage — pytest
skipped  frontier planner — pytest
skipped  arenabench — pytest (until ejection)
```

That matters right now because #3950 just fixed a *different* failure in the
arenabench step (a test reading `crates/stella-pipeline/src/roster.rs`, deleted
in #3865). That fix is verified locally — `uv run pytest
tests/test_responsibilities.py -q` is 15 passed — but **it has not been
exercised in CI even once**, because this break sits ahead of it and the step
never runs. Anyone reading the red job will attribute it to the wrong cause.

This is the third instance today of one failure hiding the next (see #3917 for
the compile-error version). The masking is what makes it urgent rather than
merely broken.

## How to verify

```
git checkout b4da79343 && (cd bench/harbor_adapter && uv run pytest -q)   # green
git checkout b4cf98740 && (cd bench/harbor_adapter && uv run pytest -q)   # 5 failed
```

## What "done" looks like

The five tests pass against the post-collapse vocabulary, by whichever of
these #3944's author intends:

1. **The Python bench emits the new keys.** If `pipeline_verifier_model` and
   friends were renamed, update the harbor adapter's posture construction and
   the launcher-vocabulary fixtures to the new spellings.
2. **The deprecation path is asserted instead.** #3944's title says "named
   deprecations", so if the old keys are meant to keep working through a
   documented alias, these tests should be asserting *that* — and the fact
   they fail says the alias does not cover this seam.

Either way, `harbor_adapter + analyzer pytest` must go green on `main`, and
the `arenabench` step must actually run so #3950's fix is finally exercised.

## Constraints

- `scripts/check-role-names.sh` holds the role vocabulary across Rust, Python
  and JavaScript and is currently **green** — so this is about *settings key*
  spellings, not role names. Do not "fix" it by widening that guard.
- Note #3055: ruff is configured but never run over `arenabench/`/`bench/`, so
  lint will not catch a stale key for you.
- `bench.yml`'s suites are also runnable via `make bench-test`.

## Context

Found while verifying #3950 on `main`: the commit that merged it still shows a
red pytest job, and the step containing the fixed test had been skipped, so
the green I expected could not appear either way.

macanderson:
## What & why

The **last** failure in `harbor_adapter + analyzer pytest`. On `main` right now that job reports `1 failed, 717 passed` — down from `5 failed, 713 passed` since #4102, and this is the one left.

`test_the_engine_still_resolves_a_role_in_the_order_this_module_mirrors` pins the Python mirror to `AgentEngineConfig::model_for`, and the engine lost a rung underneath it:

```rust
pub fn model_for(&self) -> Option<&str> {
    self.agent()
        .and_then(|a| a.model.as_deref())
        .or(self.default_model.as_deref())
}
```

Two rungs, no `kind` argument, no flat `pipeline_<role>_model` — the config collapse (#3944) and one role name in core (#3903). The test looked for the middle rung and raised `ValueError: substring not found`.

## What it asserts now

- the two surviving rungs, **in order** (`agents.<…>.model` before `default_model`);
- the middle rung is **absent** — `"pipeline_" not in body`.

That second assertion is the part worth reviewing. A flat per-role key reappearing in the engine is exactly as interesting as a reorder, and with the old assertion simply deleted it would land in silence. Its failure message names #4103 so whoever trips it lands on the reconciliation rather than on a puzzle.

## What it deliberately stops asserting, and why that is not a weakening in disguise

It no longer claims `resolve_posture_role_model` matches the engine rung for rung — **because it does not**. That function still resolves the flat key and still carries the `_WORKER_INHERITING_ROLES` arm.

I did not collapse the mirror to match, and that is a deliberate refusal rather than an oversight. `resolve_posture_role_model` is not descriptive: its production caller computes the **assurance tier** a run is characterised by (`witness-on` / `witness-off`, from whether the worker's and verifier's models differ). Changing its resolution changes which *already-recorded* runs count as witness-authored — a claim about published benchmark numbers, which is the decision #3870 reserves for a maintainer, and which CLAUDE.md tells me not to make silently.

So: the test narrows to what is still true, the divergence becomes a tracked issue (**#4103**, with three options costed) instead of a red test nobody can act on, and the docstring states both halves where the next reader hits them. A test that passed while claiming the mirror matched would have been worse than the red one.

## The witness

| Tree | `uv run --no-sync pytest -q` in `bench/harbor_adapter` |
|---|---|
| `main` | `1 failed, 717 passed` — CI's own run on `6ce03c3b3` |
| here | **`717 passed, 2 skipped`** |

Run with the provider API keys unset. `tests/test_adapter.py` picks up a real `OPENROUTER_API_KEY`/`VOYAGE_API_KEY` from the environment and fails on any machine that has one; CI has neither, and its run confirms the same single failure. That environment sensitivity is its own defect, not this one.

## What this unblocks, and one thing to expect

`bench.yml`'s job is fail-fast, so six suites have been **skipped** behind this step ever since #3944 — including `arenabench — pytest`, whose own fix (#3950) has therefore never been exercised. They start running again with this.

I ran all six locally. Five are clean (`terminal_bench_analysis` 270 passed, `telemetry_store` 10, `evidence` 66, `trace_triage` 90, `frontier` 27). **`arenabench` was not** — it aborts during collection on two missing imports, which is #4104's sibling PR. Both need to land for `bench` to actually go green; flagging it here rather than letting the job go red again for a new reason and look like this change failed.

## Ground-rule check

- [x] No production code touched — no posture value, no arm, no digest
- [x] No `#[allow]`, no `TODO`, no baseline entry, no skip or `xfail`

None. The test is rewritten in place under the same name.

Closes #3967

## Summary by Sourcery

Align the assurance-tier mirror test with the engine's current model resolution behavior while explicitly tracking the remaining resolver divergence.

Bug Fixes:
- Update the assurance-tier mirror test to reflect the engine's current two-level model resolution order and detect reintroduction of flat per-role model keys.

Enhancements:
- Document the intentional divergence between the engine model resolver and posture-role resolution without changing production behavior.

Tests:
- Rewrite the model resolution mirror test to verify agent-specific models precede the default model and that the removed pipeline key is absent.

### DIFF
--- a/bench/harbor_adapter/tests/test_assurance_tiers.py
+++ b/bench/harbor_adapter/tests/test_assurance_tiers.py
@@ -365,17 +365,40 @@ def test_the_engine_still_resolves_a_role_in_the_order_this_module_mirrors() -> 
     `resolve_posture_role_model` re-opens #2134 in a new place, and nothing
     else in either tree would notice — the two live in different languages,
     different test suites, and different CI jobs.
+
+    **The engine lost a rung, and this test now says so out loud.** `model_for`
+    read `agents.<kind>.model` > the flat `pipeline_<role>_model` > `default_model`
+    until the config collapse (#3944, #3903). It is two rungs now, with no
+    `kind` argument at all, because core knows one role. So the middle rung
+    below is asserted **absent** rather than ordered — a re-introduction of a
+    flat per-role key in the engine is exactly as interesting as a reorder, and
+    would otherwise land silently.
+
+    What this deliberately does NOT assert any more is that
+    `resolve_posture_role_model` matches rung for rung, because it does not:
+    it still resolves the flat key and still carries the worker-inheritance
+    arm. That divergence is real and is tracked in **#4103** — it cannot be
+    closed here, because that function decides the assurance tier a recorded
+    run is characterised by, so collapsing it re-characterises published
+    numbers (the decision #3870 reserves for a maintainer). Narrowing this
+    test to what is still true is the honest half; the other half is an open
+    issue rather than a passing assertion.
     """
     source = _model_for_source()
     start = source.index(_MODEL_FOR_DECL)
     body = source[start : source.index("\n    }\n", start)]
-    rungs = [
-        body.index("self.agent(kind).and_then(|a| a.model"),
-        body.index("pipeline_verifier_model"),
-        body.index("self.default_model"),
-    ]
-    assert rungs == sorted(rungs), (
-        "`model_for` no longer reads `agents.<kind>.model` > the flat per-role "
-        "key > `default_model`; `resolve_posture_role_model` mirrors that order "
-        "and has to change with it (#2134)"
+
+    agent_rung = body.index("self.agent()")
+    default_rung = body.index("self.default_model")
+    assert agent_rung < default_rung, (
+        "`model_for` no longer reads `agents.<…>.model` before `default_model`; "
+        "`resolve_posture_role_model` mirrors that order and has to change with "
+        "it (#2134)"
+    )
+
+    assert "pipeline_" not in body, (
+        "`model_for` reads a flat per-role key again. The collapse (#3944) "
+        "removed that rung and the bench mirror was left resolving it anyway "
+        "(#4103) — if the engine is bringing it back, the two have to be "
+        "reconciled deliberately rather than drifting back into agreement"
     )


macanderson:
Superseded by #4109, which makes the same conservative choice on the mirror (docstring only, flat rung kept) and also fixes the `bench.yml` path filter that let the engine change land untested. #4103 still carries the mirror-vs-engine decision and stays open after #4109 merges.

PR #4105, opened 2026-08-20
fix(stella-runtime): the composed dispatcher's doc links point at a private module, and main is red

macanderson:
## The defect — and it is mine

`main` is red. `cargo doc -D warnings`, a step of the required `fmt + clippy +
test` job, exits 101 at `2bf532b4a`:

```
error: public documentation for `bind_composed` links to private item `super::compose`
   --> crates/stella-runtime/src/wrapper/dispatch.rs:391:11
    = note: `-D rustdoc::private_intra_doc_links` implied by `-D warnings`
error: could not document `stella-runtime`
```

Four intra-doc links in `dispatch.rs` name `super::compose`, a private module
(lines 236, 242, 379, 391). Rustdoc resolves them only under
`--document-private-items` and refuses them on the public surface.

**Introduced by #4095, which is mine.** The miss is nameable rather than bad
luck: that branch was checked with `cargo clippy --all-targets -- -D warnings`
and the full `stella-runtime` suite, and **not** with rustdoc. `make
guards-fast` plus clippy does not run `cargo doc`, so a private intra-doc link
is invisible locally right up to CI.

Third time this shape has reddened `main` — #3981 was five at once.

## The fix

The four links become plain code spans. They were orientation for a reader
rather than navigable API relationships: `compose` is private *precisely* so
callers go through `bind_composed`, so naming it in backticks says the same
thing and promises no link.

No `#[allow(rustdoc::private_intra_doc_links)]`. The lint is right here.

## One stale claim went with them

`bind_composed`'s `# Errors` section said "one of the **four** conflict
variants… contradictory stage order, two oracles, two arbiters, or one
requirement name meaning two things."

That sentence was written before two of those four turned out to be
**unreachable** and were deleted inside that same PR — an oracle and a
requirement set each already require arbiter grade, so two of either is already
two arbiters. It is two, and it now says two. Left alone it would have been a
doc comment describing errors that cannot exist.

## Witness

The guard is the evidence, and it flips:

```
$ RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --workspace   # on main
error: public documentation for `bind_composed` links to private item `super::compose`
error: could not document `stella-runtime`

$ RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --workspace   # on this branch
    Finished `dev` profile [unoptimized + debuginfo] target(s)
   Generated .../target/doc/loop_bench/index.html and 28 other files
```

Run across the **whole workspace** this time, not the three crates I checked on
the branch that caused it.

[...]

Issue #4089, opened 2026-08-20
ci: five make-gate guards run in no workflow — `tokens` was red on main for a day and nothing said so


macanderson:
## Problem

Five `make gate` guards run **nowhere server-side**. They are in `GATE_GUARDS_FAST`, so the pre-push hook runs them, but `.github/workflows/ci.yml` enumerates its guard steps by hand and does not include them:

- `tokens` (`scripts/check-tokens.py` + `scripts/gen-tokens.py --check`)
- `typed-errors` (`scripts/check-typed-errors.py`)
- `dead-code-allows` (`scripts/check-dead-code-allows.py`)
- `diagnostic-codes`
- `bench-suites` (`scripts/bench-suites.sh` — `bench.yml` runs the pytest suites, not the guard that keeps the two lists in step)

`invariants` and `doc-links` are deliberately in `docs-guards.yml`; these five are in no workflow at all.

The full list CI *does* run:

```sh
rg -n "run: .*scripts/" .github/workflows/ci.yml | sed 's/.*scripts\///' | sort -u
```

## Why it matters

The pre-push hook is advisory and per-clone, bypassable with `SKIP_GATE=1` or `--no-verify`, and with `enforce_admins` off an admin or auto-merge can land a gate-failing change anyway. AGENTS.md already names that as the exact reason the hook "complements the required server-side checks rather than replacing them" — but for these five there is no server-side check to complement.

This is not hypothetical. `tokens` was red on `main` from #4066 (2026-08-19) until #4086, and nothing in CI said so. It reported seven retired brand hexes and thirteen stray hex literals the whole time, on a tree where every PR's checks were otherwise green. It also masked the deeper defect that #4087 found: `crates/stella-tui-theme/src/generated.rs` was never `mod`-declared, so `make tokens` was verifying a file the crate did not compile — and only `module-reachability`, which *is* in CI, caught it.

`gate-parity` does not cover this. It holds AGENTS.md and CONTRIBUTING.md to `GATE_STEPS`; it says nothing about whether a workflow runs a given step, which is the actual gap.

## Definition of done

Either:

1. Add the five steps to `ci.yml`'s required job, each with `if: ${{ !cancelled() }}` like its neighbours, so each reports independently; **or**
2. Have `ci.yml` invoke the Makefile target rather than re-listing scripts, so the two cannot drift.

[...]

PR #4116, opened 2026-08-20
fix(stella-autonomy): a public module doc cannot link a private one

macanderson:
## What & why

**`main` is red** on the required `fmt + clippy + test` job, at `cargo doc`:

```
  --> crates/stella-autonomy/src/gate.rs:4
```

`gate` is `pub mod`; `deliver` is `mod` (`crates/stella-autonomy/src/lib.rs:45`). Arrived with #4097.

While this is red every open PR is red too, and red stops distinguishing "your change is broken" from "the base is".

## The fix, and why not the other one

Plain code text rather than an intra-doc link — **not** making `deliver` public.

A reader of `gate`'s public documentation cannot navigate into a private module whatever we write there, so the backticks deliver exactly what the sentence meant and nothing it can't. AGENTS.md names this as the usual right answer when the reference is a reader's orientation rather than a navigable API relationship. Widening a module's visibility to satisfy a cross-reference would be changing the crate's surface to satisfy a footnote.

## Why it keeps happening

This is the **third** instance today — #4105 and #4106 fixed two in `crates/stella-runtime/src/wrapper/dispatch.rs` — and the fifth and sixth since #3981 recorded five at once.

The mechanism is always the same and worth naming: the lint is `rustdoc::private_intra_doc_links`, which only fires under `--document-private-items`. That flag is passed by `make doc-warnings` and **not** by a plain `cargo doc`, so the error is invisible to an ordinary local build and only appears in the pre-push hook or CI. Someone writing a doc comment next to a private sibling gets no feedback at all until the gate runs.

I'm not fixing that here — it's a guard/ergonomics question, not this PR's — but three in one day suggests it's worth one. #4089 (five gate guards that run in no workflow) is the neighbouring version of the same "the check exists but nothing routine runs it" problem.

## The witness

| Tree | check |
|---|---|
| `main` | CI's own `cargo doc -D warnings` step emits the error above — run `32343348527`, job `96346787878` |
| here | `make doc-warnings` → `Finished` / `Generated … and 28 other files`, clean |

Also run here: `cargo clippy --workspace --all-targets -- -D warnings` — clean, so this is the only thing standing between `main` and a green required job.

No test is added because there is no behaviour: the diff is one pair of backticks in a `//!` comment.

## Ground-rule check

- [x] No `#[allow(rustdoc::private_intra_doc_links)]`, no widened visibility, no `TODO`

None.

## Summary by Sourcery

Bug Fixes:
- Fix the public module documentation so it no longer creates an invalid link to the private `deliver` module.

### DIFF
--- a/crates/stella-autonomy/src/gate.rs
+++ b/crates/stella-autonomy/src/gate.rs
@@ -1,7 +1,7 @@
 //! Which checks are allowed to block a merge, and which have stopped earning
 //! that right.
 //!
-//! `doc:backlog-self-driving` §3.3 (#3599). [`crate::deliver`] decides what to
+//! `doc:backlog-self-driving` §3.3 (#3599). `crate::deliver` decides what to
 //! do about a red build; this decides *which reds count*. They are separate
 //! questions and were previously the same one, which is how a loop ends up
 //! permanently blocked by something no pull request could ever fix.


macanderson:
Superseded by #4109, which fixes this same `gate.rs` doc link plus the two bench defects, and additionally fixes the `bench.yml` path filter that let them land — see my review comment there. Closing to cut the duplicate churn; no disagreement with the fix itself, it is the same one-line change.

Issue #4119, opened 2026-08-20
self-driving: the drive loop never closes or triages an issue — the lifecycle verbs are subcommand-only


macanderson:
The issue-lifecycle discipline is implemented, tested, and unreachable from the loop that is supposed to practise it. A self-driving session merges pull requests and never closes an issue, never re-labels one, and never records a closure.

## What exists

`crates/stella-cli/src/self_driving_cmd/lifecycle.rs`:

- `triage_issue` (line 23) — re-labels an issue from the configured ladder
- `close_issue` (line 91) — closes with a canonical resolution, and counts it through `SessionStats::record_closure` (line 227), which keeps `closures_balance` true by construction

## What reaches them

Only the two hand-typed subcommands, in `crates/stella-cli/src/self_driving_cmd.rs`:

- line 603 → `lifecycle::triage_issue`
- line 614 → `lifecycle::close_issue`

Neither name appears in `crates/stella-cli/src/self_driving_cmd/drive.rs`. Confirm with:

```
rg -n 'close_issue|triage_issue' crates/stella-cli/src/self_driving_cmd/drive.rs
```

which returns nothing.

## What it costs

Observed on a live session against `macanderson/oxagen-platform`: sixteen pull requests merged, zero issues closed by the loop. Every closure in that run was performed by hand afterwards. `stella self-driving stats` reported `closed_total 0` — correctly, since `record_closure` is only called from a path the loop never takes.

The counters are the tell rather than the bug. The bug is that a loop which merges a fix leaves the issue it fixed open, so the next pass sees it in the queue as unclaimed work.

## What "done" looks like

`drive.rs` calls the lifecycle verbs at the points where the loop already knows the answer:

- **a merged pull request that delivered the fix** → `close_issue` with `completed`, citing the pull request in the closing comment. The commit trailer is not enough on its own; see AGENTS.md § "Closing the issue on merge" for why both the description link and the trailer are required, and note the loop currently writes neither.
- **a turn that delivered nothing and found the issue already fixed on the base** → `not planned`, citing what it found. Four issues were closed this way by hand on `macanderson/oxagen-platform` (#1144–#1147) after being filed against a stale local tree.
- **a partial fix** → close nothing; file the remainder and link it. The loop already has the shape for filing (`issues_created` writes at `self_driving_cmd.rs:1410`), also only from a subcommand.

## Constraint

[...]

Issue #4118, opened 2026-08-20
self-driving: three session-stats counters have a printer and no producer


macanderson:
`stella self-driving stats` prints three numbers that no code path can ever make non-zero. A live dashboard reading them reports a session that learned nothing, whatever the session actually did.

## The three

`crates/stella-autonomy/src/stats.rs`:

- line 110 `pub reflections_logged: u32`
- line 112 `pub memories_created: u32`
- line 114 `pub proposals_made: u32`

Each has exactly two sites in the workspace: the field declaration above, and a `println!` in `crates/stella-cli/src/self_driving_cmd/stats.rs:74-76`. There is no third — no increment, no assignment, no constructor that sets them from anything. Confirm with:

```
rg -n 'reflections_logged|memories_created|proposals_made' crates/
```

Three hits per name is the whole story.

## Why this is the inverted form of invariant #10

`crates/stella-protocol/src/event/consumers.rs` enforces that every emitted signal names its consumer. These are the mirror image: consumers with no producer. A reader cannot tell "the session reflected nothing" from "nothing counts reflections", and the surface renders both as `0`.

The contrast inside the same file is the argument for fixing it rather than deleting it: `SessionStats::record_closure` (line 179) is a real mutator with a real caller, and `closures_balance` (line 168) is a self-check that catches a resolution the vocabulary did not know about. The closure counters were built to be trustworthy. These three were not built at all.

## What to do

Increment them where the events already happen. All three now occur on the loop's own path:

- **reflections** — a one-shot turn's reflection call. Since the fix on `self-driving-required-checks` a machine-format turn reflects, so a self-driving turn produces a second `system` execution per turn; `agent::output::should_reflect_after_one_shot` is the predicate that decides it.
- **memories** — `SessionMemory`'s novel-lesson path in `crates/stella-cli/src/memory/learning.rs` already computes `novel_logged`, which is the number wanted.
- **proposals** — needs a definition before it needs a counter. Decide what a proposal *is* in a self-driving session (a filed issue? a context record offered for promotion?) and if the answer is "nothing yet", the honest move is to delete the field and its `println!` and re-add it with its producer, rather than ship a zero.

Deleting is an acceptable outcome for any of the three. Printing an uncomputable number is not.

## Definition of done

Each surviving field has a write site reached by `stella self-driving drive`, and a test that runs the loop far enough to assert the number moved. Any field kept without a producer is instead removed from both the struct and the printed report.

## Verify

[...]

Issue #4003, opened 2026-08-19
Every autonomous work unit rebuilds the full code-graph index, because each worktree starts empty


macanderson:
Measured during #3942's first live session against this repository.

## What was measured

`stella self-driving drive` worked three issues. The log contains **three complete code-graph index builds**, one per work unit:

```
✓ code graph: 28975 symbols, 16989 imports across 1650 files
  (1650 re-parsed, 0 unchanged this pass) — skipped 2 generated files
```

Note `0 unchanged` — nothing was reused. Each turn walked and parsed all 1,650 files from scratch.

## Why

`work` creates an isolated worktree per issue (`.stella/private/self-driving/<slug>/`) and spawns `stella run` inside it. A fresh worktree has no `.stella/private/codegraph.db`, so the turn builds one. The isolation is correct and not in question — the loop must be able to run while a human works in the same clone — but the index is a **pure function of the tree**, and three worktrees cut from the same base ref have the same tree.

## What it costs

1. **Wall clock.** A full parse of this repository before the turn does any work.
2. **Budget.** The turn runs under `--spend-limit`, and indexing is inside it. A unit that spends its allowance getting oriented has less left for the work — and on this run one turn hit a stuck-loop after the index, with the model cycling `bash` calls for three rounds. Whether the two are related is not established, but a turn that starts with less budget than it looks like it has is a bad shape regardless.

## Options, in rough order of appeal

- **Seed from the parent.** Copy or hard-link the workspace's existing `codegraph.db` into the new worktree before the turn. The tree differs from the base only by whatever the turn is about to do, so the index starts warm and incremental parsing handles the rest.
- **Point at a shared index.** Riskier — concurrent worktrees writing one database is the corruption shape #2293 already records.
- **Skip indexing for a work turn.** Cheapest, and worst: it takes away the retrieval the turn most needs on an unfamiliar issue.

## Verify with

Run `stella self-driving drive --max-issues 3` and count `code graph:` lines in the log. Three today; one after a fix, with the later two reporting a high `unchanged` count.

## Constraint

Whatever is done must not make two concurrent work units share a writable index — see #2293 for what codegraph corruption looks like when it survives an open.

## Definition of done

The done conditions this issue already states, as boxes the gate can read.
Nothing here is new scope, and nothing is ticked by the migration — the
gate reads one heading and this issue used another (macanderson/oxagen#1407).

The live-run count this issue's "Verify with" section asks for is now #6018,
[...]

Issue #3650, opened 2026-08-18
codegraph: five concurrent writers share one store with nothing but busy_timeout between them


macanderson:
## Problem

Nothing serializes the passes that write `codegraph.db`. Between them there is only `PRAGMA busy_timeout=5000` (`crates/stella-graph/src/store.rs`, `open_read`).

Concurrent writers that exist today:

- `CodeGraph::mount`'s reconcile step and catch-up walk (`crates/stella-graph/src/graph.rs`), armed on every session start;
- the live `notify` watcher's debounced `apply_changes` batches (`crates/stella-graph/src/watch.rs`);
- `search::codegraph::open_or_build`, which runs a full `index_all()` on **every** graph-tool open (`crates/stella-tools/src/search/codegraph.rs`);
- the lazy embed catch-up inside a search call;
- any second `stella` process in the same workspace, plus `stella init` run by hand.

They serialize on SQLite's write lock and can exhaust the 5s timeout under a large tree, which surfaces as an `index_warning` that the search path degrades past rather than reports. Two processes can also each run a whole redundant tree walk, paying the cost twice for one result.

PR #3649 (background re-embed on a ref change) makes this materially worse: it adds a fifth writer that fires exactly when a merge has just made the tree large and dirty.

## What to build

A single-flight advisory lease in the store — one row naming holder, purpose, and an expiry — that a pass acquires before walking or embedding and releases on completion. Requirements:

- **Must expire.** A killed process must not wedge indexing forever; a stale lease past its deadline is stealable.
- **Must fail open, not closed.** Failing to acquire means "someone else is already doing this, skip" — never an error surfaced to the user, and never a blocked search.
- **Must distinguish purposes.** An index pass and an embed pass do not conflict with each other the way two index passes do; decide this deliberately rather than by accident.

## Constraints

- The store is a rebuildable cache. A lease must never be the reason a rebuild cannot happen — the corruption-quarantine path (`store.rs`, `open_verified`) must stay reachable.
- `stella-graph` has no clock port today; `store::now_unix()` is what the rest of the schema uses.
- Do not reach for a file lock: `.stella/private/` is already owner-only, but the existing durability story is entirely inside SQLite and splitting it across two mechanisms is how the two disagree.

## Definition of done

A test with two concurrent index passes over one store where exactly one does the work and the other returns promptly without error, plus a test that a lease past its expiry is stolen rather than honored.

PR #4130, opened 2026-08-20
feat(self-driving): the loop runs unattended — a merge gate that knows which reds count, a base it repairs, and learning that outlives the turn

### LINKED ISSUE #3876: goal_wrapped_dispatch_cli claims 'no ambient credentials' but reads the developer's real ~/.stella and bills real money
## Problem

`crates/stella-cli/tests/goal_wrapped_dispatch_cli.rs` documents itself as hermetic — its header says "no ambient credentials, `--base-url` pointed at a `wiremock` server this test drives itself" — but on a developer machine with a configured Stella it reaches **live providers and spends real money**.

Observed running `cargo test --workspace` locally:

```
· step 2 · moonshotai/kimi-k3 · 1.9k (1.7k cached) in → 174 out · $0.0036 · 6.5s · 1 tool call
· step 3 · moonshotai/kimi-k3 · 2.1k (1.8k cached) in → 315 out · $0.0058 · 11.4s
✓ complete · glm-5.2 · $0.1327
◆ zai/glm-5.2 · $0.1327 · 208.2s
stella: goal not met after 8 round(s): round cap (8) reached without a passing verdict
test result: FAILED
```

$0.13 and 208s for one test, and it **fails nondeterministically** — the outcome depends on whether a live model satisfies a live verifier, not on the property under test (the wrapper's `before_turn` firing once per goal round).

## Root cause

The test scrubs provider env vars thoroughly (`OPENROUTER_API_KEY`, `ANTHROPIC_API_KEY`, `ZAI_API_KEY`, … ~15 of them) and sets `STELLA_DATA_DIR` to a temp dir — but it never sets **`STELLA_HOME`**.

Per AGENTS.md, `STELLA_DATA_DIR` redirects only the *data* tier. `credentials.toml` and `settings.json` live in the **config** tier under `~/.stella`, which only `STELLA_HOME` moves. So the subprocess reads the developer's real credentials and real provider/model config, and routes past the wiremock `--base-url` to whatever their settings resolve.

Env-var scrubbing cannot close this: the credentials are on disk, not in the environment.

## Definition of done

- The test sets `STELLA_HOME` to a temp dir (alongside the existing `STELLA_DATA_DIR`), so no real config or credential file is reachable.
- Verified the way this bug was found: run it on a machine **with** a configured `~/.stella` and confirm zero spend and no live provider in the output.
- Sweep the sibling CLI tests for the same gap — `goal_arbiter_wrapper_refusal_cli.rs` declares the same hermeticity and looks likely to share it; `run_exits_cli.rs` is named in this test's header as the discipline being matched, so it is worth checking which of the two is actually right.
- Consider a shared spawn helper that sets both variables, so the next test cannot get it half-right.

## Notes

- Pre-existing; not caused by #3874, where it was found. It is the only workspace test that failed there and it fails for this reason, not for anything in that diff.
- Related standing hazard: stella-cli tests can mutate the developer's real `~/.stella` when the home tier is not redirected.

macanderson:
## What & why

Everything needed for `stella self-driving drive` to run unattended against a real repository, found by running it rather than by reading it. Two live sessions produced this branch: three autonomous merges on this repository, and sixteen on a private repository where billing blocks CI entirely.

The commits fall into four groups.

**The merge gate learns which reds count.** A check that is red on every one of the last N base commits is infrastructure, not a verdict on your change, and a loop that waits for it waits forever. `stella-autonomy`'s new `gate` module decides *which* reds block (`stuck_on_base`, `blocking`, `waived`); `deliver` reads both of GitHub's check dialects, because a `StatusContext` has no `status` field and reads PENDING forever if you look for one. Where no required check can be green, the loop verifies locally against a baseline measured on the base ref in a throwaway worktree, and labels what it proved.

**A broken base gets repaired instead of waited on.** The loop files the breakage, adopts its own issue, and fixes it, rather than stalling behind a red it did not cause.

**A turn's learning outlives the turn.** This is the half that mattered most and is described under the witness below.

**Test hygiene the above depended on.** Five CLI tests declared hermeticity they did not have.

Closes #3876

## The witness

Per group:

- **`gate`** — seven tests covering `stuck_on_base` / `blocking` / `waived`, all new behaviour.
- **State root** — `the_state_root_follows_the_override_not_the_working_directory` and `an_unusable_override_falls_back_instead_of_resolving_somewhere_worse` in `stella-home`, driving the pure resolver so the module stays safe under the parallel runner.
- **Reflection** — `a_machine_format_turn_reflects_like_a_text_one` plus two siblings in `agent::tests`.
- **Hermeticity** — the five suites themselves; `goal_wrapped_dispatch_cli` goes from a 154-second failure to a 2.7-second pass with no provider traffic.

Two of these need their verification stated rather than assumed.

**The reflection fix could not be witnessed by a unit test alone**, because the fix extracts the predicate the bug lived outside of. It is verified on the wire: one JSON turn now produces **two** executions — the worker run, plus a `system` execution costing $0.0066, which is the reflection model call. Before this change that call did not exist on any JSON turn. Note that the `reflections` table is *not* a usable signal here: it gets a row per lesson, so "unchanged" is consistent with both "did not run" and "ran and found nothing".

**The state-root fix** is unit-tested at the resolver and observed end to end: a turn run in a detached worktree wrote a memory into the repository's own `context.db`, and the worktree was then removed with the lesson still there.

## The gate

- [x] `cargo fmt --check`
- [x] `make gate` green end to end, `self-driving-test` included (61 checks)

`Closes #3876` is in this description only — the fixing commit was already pushed without the trailer and rewriting pushed history is worse than the gap. Worth a check on merge.

## Nothing left behind

- [x] Filed: #4118, #4119

**#4118** — three session-stats counters (`reflections_logged`, `memories_created`, `proposals_made`) have a field and a `println!` and no writer anywhere, so a dashboard reads zero whatever the session did. The inverse of invariant #10: consumers with no producer.

**#4119** — `close_issue` and `triage_issue` are reachable only from hand-typed subcommands, never from `drive.rs`. This is why the sixteen-merge session closed no issues.

Also commented on **#4003** rather than closing it: I implemented its "point at a shared index" option, then reverted it in this branch. That option makes a self-driving turn write the *human's* `codegraph.db` while they may have it open, which is #2293's corruption shape — and the isolation exists precisely so the loop can run while somebody works in the same clone. The comment records one nuance in the other direction: `open_or_build` already single-flights the index pass (#3650), so the hazard is narrower than #4003 assumed. Narrower is not measured, and seeding from the parent remains the safe option that issue recommends.

## Ground-rule check

- [x] No I/O added to `stella-core`; no new dependencies at all
- [x] `stella-autonomy` stays a leaf: no workspace dependencies, pure decisions

**One deletion to declare:** none — no test was removed. The harness case in `scripts/test-self-driving.sh` was *rewritten*, and that deserves scrutiny, so here is the reasoning in full. `priority::triage` deliberately stopped ranking a defect carrying no priority rung: "no rung" used to map to one past the ladder, making an unjudged issue indistinguishable from one somebody had ranked least urgent. It now surfaces as unassessed so the loop places it instead of burying it. That change shipped with a unit test owning the property (`a_defect_with_no_rung_surfaces_as_a_question`) and without updating the end-to-end case, which had been failing `make gate` ever since — masked locally by earlier failing steps, and invisible to CI, which does not run this harness. I changed the expectation, not the behaviour, and added a second assertion: the ranked list alone would pass identically if the issue were dropped on the floor, so the count reported as awaiting triage is asserted too.

**Why five tests were leaking credentials, since the pattern will recur.** `STELLA_DATA_DIR` moves only the data tier under `~/.stella`. The credentials file lives in the config tier, which only `STELLA_HOME` moves. So scrubbing thirteen `*_API_KEY` variables is not isolation — an environment variable is one of two ways a credential reaches the process and that file is the other. `goal_wrapped_dispatch_cli` therefore bound a real OpenRouter verifier against a real zai worker and spent $0.08 of the developer's money per `make gate`, then failed, because a live model does not answer like the wiremock fixture it was driving. The other four were *passing*, which is the part worth noticing: passing was not evidence of isolation, only that the leaked credential happened not to change the answer. `run_exits_cli` is the file the others cite as the hermeticity exemplar, so it was teaching the gap as the pattern.

**Reviewing order.** The four groups are independent; the state-root and reflection commits are the two to read closely, and `goal.rs`'s guard is the whole reflection bug in one line.

## Summary by Sourcery

Enable unattended self-driving loops to gate merges intelligently, repair broken bases without duplication, and retain learning across disposable worktree turns.

New Features:
- Make self-driving sessions distinguish infrastructure failures from change-blocking checks and recover automatically from broken base branches.
- Persist issue associations for carried pull requests so base-breakage fixes are not adopted repeatedly.
- Preserve reflections, memories, telemetry, and code-graph state when work executes in disposable worktrees.

Bug Fixes:
- Enable reflection for JSON and streaming turns while retaining opt-outs and avoiding unnecessary reflection calls.
- Correct workspace state resolution so durable state follows an explicit absolute repository-root override.
- Treat untriaged defects as awaiting triage rather than silently ranking them as lowest priority.
- Prevent CLI tests from reading developer credentials or contacting real providers.

Enhancements:
- Add end-to-end and unit coverage for merge gating, durable state roots, machine-format reflection, idempotent base repair, and hermetic CLI execution.

Tests:
- Strengthen self-driving harness assertions for untriaged issue reporting and queue contents.
- Make affected CLI integration tests hermetic by isolating both user and data configuration tiers.

Chores:
- Add workspace state-root configuration shared by the CLI and private store layers.


macanderson:
Ordering note while this is still a draft — **land #4109 before marking this ready**, or this PR can reproduce the exact failure #4109 exists to fix.

This changes `crates/stella-cli/src/settings/unknown.rs`. `bench/harbor_adapter/tests/test_assurance_tiers.py` and `test_posture.py` **parse that module** — `ENGINE_ROOT_FIELDS`, `RETIRED_ENGINE_ROOT`, `RETIRED_ENGINE_AGENT_NAMES` and `AgentEngineConfig::model_for` — because the Python posture mirror has to agree with the engine's vocabulary and resolution ladder.

`bench.yml`'s scope filter on `main` today is:

```
'^(bench/|arenabench/|crates/stella-model/src/catalog\.rs$|\.github/workflows/bench\.yml$)'
```

`crates/stella-cli/src/settings/` is not in it. So a PR that changes the vocabulary reports `changed=false`, **skips every bench suite**, merges green — and then reds every subsequent push to `main`, because a push runs them unconditionally. That is not hypothetical: it is what #3908 and #3944 did, and it is why `bench` has been red since.

#4109 adds `crates/stella-cli/src/settings(\.rs$|/)` to that filter. With it landed, this PR's own checks will run the bench suites and tell you immediately if the vocabulary moved out from under the mirror. Without it, they stay silent and the answer arrives after the merge.

Also worth knowing for this PR specifically, since it touches `gate.rs`: `main` is currently red on `cargo doc` because `gate.rs`'s public module doc links `[`crate::deliver`]`, which is a private `mod`. #4109 fixes that too. If this PR rewrites that doc comment anyway, make sure the replacement does not reintroduce an intra-doc link to a private item — the lint only fires under `--document-private-items`, which `make doc-warnings` passes and a plain `cargo doc` does not, so it is invisible locally. That shape has reddened `main` three times today (#4105, #4106, and this one).

No conflict between the two PRs in file terms beyond `gate.rs`; this is purely about which order makes the checks meaningful.

Issue #4159, opened 2026-08-21
stella goal and daemon resume never measure what their turns changed — empty file ledger for the whole run


macanderson:
## Problem

`stella goal` and `stella daemon resume` own their own turn event streams and never measure what their turns changed, so their file ledger is empty for the whole run: no `AgentEvent::FileChange`, no `files_touched` rows, an empty Files tab / `stella export` / audit log, and `Store::finalize_execution_reflection`'s `wrote_files` flag reads false however many files the run wrote.

This is the same defect PR (branch `worktree-files-touched-fix`) just fixed for the Command Deck's lead turn, filed separately because the goal loop's shape makes it a real design decision rather than a one-line call.

## Why these three were not fixed in the same change

`turn_files::close_turn_boundary` pays two debts at one boundary: the tree measurement, then the run's terminator (`emit_run_complete_for_turn`, gated on one turn's `TurnOutcome`).

The goal drivers do not have that shape. They call `persistence::emit_run_complete_on_raw` with the *whole loop's* cost, deliberately — a goal run drives several turns over one stream and must emit exactly one terminator for the run (see `emit_run_complete`'s doc on why no wrapper emits it). So the correct repair is **not** to swap in `close_turn_boundary` at the end: that would measure once for a multi-turn run and attribute every turn's changes to the last one.

The measurement is per-turn and must happen at each turn boundary *inside* the loop, while the terminator stays where it is at the end. That is a design decision worth making deliberately.

## Files

- `crates/stella-cli/src/agent/goal.rs` — `run_goal_turn`; channel at L747, terminator at L792
- `crates/stella-cli/src/agent/goal/goal_wrapped.rs` — channel at L179, terminator at L348
- `crates/stella-cli/src/agent/resume.rs` — `run_resume`; channel at L166, terminator at L252
- `crates/stella-cli/src/turn_files.rs` — `emit_shared_tree_changes` / `close_turn_boundary`, and the doc explaining why the producer is a measurement and never a tool hook
- `crates/stella-cli/src/durability.rs` — `SessionDurability::snapshot_worktree`, and `bind` which now establishes the session's baseline

## Constraints already discovered

- **The snapshot is consumed by whoever reads it.** `snapshot_worktree` advances its own baseline, so it must be called exactly once per turn boundary. Two callers per turn means the second reports an unchanged tree and the two projections disagree about the same turn (`turn_files`'s module doc; `turn_facts.rs` L23).
[...]

Issue #4060, opened 2026-08-20
cli: a developer cannot run `make gate` isolated — STELLA_HOME defeats the settings tests' own HOME sandbox


macanderson:
## Problem

Two facts about `make gate` on a developer machine compose into "there is no environment in which the whole gate is green":

1. **With the real home**, `crates/stella-cli/tests/goal_wrapped_dispatch_cli.rs` picks up ambient `~/.stella/credentials.toml` and runs a real cross-family goal loop against real providers — spending money, taking ~40s instead of ~3s, and failing on model behaviour ("verifier verdict (round N): not yet met") rather than on the property it exists to witness. That is #3876's shape: `credentials.toml` is CONFIG tier, so pointing only `STELLA_DATA_DIR` at a temp dir is not isolation.

2. **With `STELLA_HOME` pointed at an empty directory** — the obvious way to avoid (1) — `settings::tests::untrusted_project_cannot_enable_tools_or_replace_an_agent_prompt` fails at `crates/stella-cli/src/settings/tests.rs:875` with "untrusted project enabled bash". That test builds its sandbox by setting `HOME` and clearing `STELLA_TRUST_PROJECT` / `STELLA_PROJECT_HOOKS`, but it does not clear `STELLA_HOME` — and `stella_home::OVERRIDE_ENV_VARS` puts `STELLA_HOME` above `HOME`, so the ambient override wins and the test reads a home it did not build.

Measured on 2026-08-20, on the same tree, back to back:

- `make gate` → one failure, `goal_wrapped_dispatch_cli`.
- `STELLA_HOME=<empty dir> make gate` → one failure, `untrusted_project_cannot_enable_tools_or_replace_an_agent_prompt`.

Every other gate step is green in both. CI does not hit either, because CI has neither a `~/.stella/credentials.toml` nor an ambient `STELLA_HOME` — which is exactly why this has stayed invisible.

## Files

- `crates/stella-cli/src/settings/tests.rs` — around line 860, the `set_var("HOME", …)` sandbox block (and any sibling test built the same way)
- `crates/stella-home/src/lib.rs` — `OVERRIDE_ENV_VARS`, and the precedence of `STELLA_HOME` over `HOME`
- `crates/stella-cli/tests/goal_wrapped_dispatch_cli.rs` — the half that is #3876

## Reproduce

```sh
[...]

Issue #3982, opened 2026-08-19
untrusted_project_cannot_enable_tools_… passes only because it reads the real ~/.stella — it overrides HOME but not STELLA_HOME


macanderson:
`settings::tests::untrusted_project_cannot_enable_tools_or_replace_an_agent_prompt` (`crates/stella-cli/src/settings/tests.rs`) is green on a normal developer machine and red the moment `STELLA_HOME` is set in the ambient environment. It is the mirror image of #3876: that test fails *because* it reads the developer's real `~/.stella`; this one passes *only because* it does.

## Repro

```
cargo test -p stella-cli --bin stella untrusted_project_cannot            # ok
STELLA_HOME=$(mktemp -d) cargo test -p stella-cli --bin stella untrusted_project_cannot
#   panicked at crates/stella-cli/src/settings/tests.rs:875:5:
#   untrusted project enabled bash
```

Deterministic in both directions, on a tree with no changes to `settings.rs`, `stella-home`, or anything either reaches.

## Mechanism

The test isolates itself by setting `HOME` to a tempdir and writing a **user-tier** `settings.json` there with `{"tools": {"bash": "off", "scratch": "off"}}`, plus an untrusted **project-tier** one turning both on. The assertion is that the project cannot flip them.

But `HOME` is not the only door. `stella_home::OVERRIDE_ENV_VARS` documents `STELLA_HOME` as the override that moves the *whole* home, and it **wins over `HOME`** (AGENTS.md, "Everything user-global lives under `~/.stella`"). With `STELLA_HOME` set in the environment, the user-tier file the test just wrote is never read at all, so the "off" that the project is supposed to be unable to override was never in effect — `allows("bash")` is true on the default, and the assertion fires having proved nothing about trust.

So the test's guarantee holds only under an environment it does not establish. The failure is loud here, but the quiet direction is worse: on any machine or CI runner that *does* export `STELLA_HOME`, this test would go red for a reason that has nothing to do with the trust gate it is named after, and someone would spend the afternoon on `Settings::load`.

## What "done" looks like

- The test clears (or sets) **every** entry of `stella_home::OVERRIDE_ENV_VARS` — `STELLA_HOME` and `STELLA_DATA_DIR` — alongside its `HOME`, under the same binary-wide env lock it already takes. Setting `STELLA_HOME` to the tempdir it already creates is the more direct expression of what it means, since that is the variable the resolver actually consults first.
[...]

PR #4160, opened 2026-08-21
fix(stella-cli): the Command Deck's Files tab shows what the turn changed

macanderson:
## The bug

The Command Deck's Files tab (`/files`, `/diff`) read `no files touched yet` for the whole of every session, however many files the turn created, edited or deleted.

Two independent causes, both of which had to go:

**1. The deck's turn driver never measured.** `turn_files::emit_shared_tree_changes` had exactly one caller, `agent::run_turn`. The deck drives its lead turn through `command_deck::run_lead_turn`, which paid only the run terminator (`persistence::emit_run_complete_raw`). So the default interactive shell on a TTY — the surface a human actually watches — emitted no `AgentEvent::FileChange` at all.

The asymmetry is why this survived: a driver that drops the *terminator* hangs every consumer immediately and is found in minutes (#2290, #960), while a driver that drops the *measurement* renders an empty ledger indistinguishable from an honest one. Only the loud debt was ever noticed.

**2. The session's first turn always reported nothing.** `snapshot_worktree` answers with the delta since the previous snapshot, and the first call of a lineage has no previous snapshot — it establishes one and returns empty, on purpose ("every file in the workspace" is not what a turn changed). Nothing established that baseline at session start, so the *turn* boundary spent its first reading on it. Not a corner case: one-shot `stella run` opens a fresh session per invocation, so its first turn is its only turn and its file stream was empty by construction on **every run**.

## The fix

- `turn_files::close_turn_boundary` folds the two boundary debts into one call — tree measurement first (these are the turn's own events, and a consumer folding the stream must see them inside the turn they describe), then the run's terminator (#3379). `close_turn_boundary_raw` is the raw-sender variant for the deck, mirroring the helper it replaces; its `EventSender` is a temporary so no clone outlives the call and wedges `close_turn_stream` (#2290).
- **`persistence::emit_run_complete_raw` is deleted.** It was what made the omission possible — it let a driver pay the loud debt alone while reading like a complete boundary at the call site. Removing it is the structural half of the repair: the deck can no longer terminate a turn without also measuring it.
- `SessionDurability::bind` now establishes the session's tree baseline, so the first turn measures against the moment the session opened. One call site covers every bind path (deck start, in-deck session switch, `presence`, `resume`). The reading is discarded on purpose — whatever changed before this session existed is not this session's work.

`command_deck.rs` is a grandfathered god file closed to growth: this change is **net zero lines** there (3912 before and after), which is why the raw-sender variant lives in `turn_files.rs` rather than the call site being expanded inline.

## "do not show read files"

Satisfied structurally, not by filtering. `FileChangeKind::Read` has **no live producer anywhere in the workspace** — both surviving producers diff a tree against a tree, and a read leaves no trace in a tree to diff (`stella-protocol/src/event/payload.rs` L45-56). The variant survives only so replay can parse journals recorded before the 12-tool purge, so the tab's `reads` column stays for that path and is structurally always 0 on a live turn.

## Witnesses

All three fail on `main` and pass here — verified by reverting each production change while keeping the tests.

| Witness | On `main` |
|---|---|
| `durability::tests::the_first_turn_after_a_bind_reports_its_own_changes` | `left: []` vs `right: ["born.rs", "predates.txt"]` — the reported symptom exactly |
| `turn_files::tests::every_turn_owner_pays_both_halves_of_the_boundary` | fails naming `command_deck.rs (run_lead_turn)` |
| `turn_files::tests::the_terminator_only_raw_helper_stays_deleted` | fails — `emit_run_complete_raw` is present |

Plus `a_bind_absorbs_the_tree_it_found_rather_than_reporting_it` (the other half: a bind that reported the whole workspace on turn one would satisfy a "not empty" assertion and be a worse bug) and `binding_does_not_deadlock_on_its_own_lock` (`bind` takes the write lock and then reads through `snapshot_worktree`, which takes the read lock; `RwLock` is not re-entrant).

The source fence follows the idiom `durability`'s #2177 fence already established here, for the same reason and against the same defect shape — what decays is the *call site*, inside ~400-line async drivers that need a provider, a journal-bound session and a file-touching turn to reach. The measurement itself is covered end-to-end by the `durability` baseline witnesses and `stella-store`'s `snapshot_worktree` tests.

## Verification

`make guards` green (32 steps), `cargo fmt --check`, `clippy -D warnings` on `stella-cli`, `RUSTDOCFLAGS=-D warnings cargo doc`, file-size ratchet green, and `cargo test -p stella-cli` at **1894 passed**.

One unrelated failure, `settings::tests::untrusted_project_cannot_enable_tools_or_replace_an_agent_prompt`, reproduces identically on the base commit `37bf308fc` and is already tracked as #3982 / #4060: it overrides `HOME` but not `STELLA_HOME`, so it passes only against a real `~/.stella`. Its failing here confirms this run was correctly isolated.

## Left behind

`stella goal` and `stella daemon resume` own their own turn streams and have the same empty ledger. They are **not** fixed here: a goal run drives several turns over one stream and emits one terminator for the whole run, so swapping in `close_turn_boundary` at the end would attribute every turn's changes to the last one. The measurement has to be per-turn inside the loop, which is a design decision worth making deliberately. Filed as a handoff: #4159.

Sub-agent (`subsession.rs`) and fleet lanes are deliberately excluded — they run beside the session's own turn against the same journal, so a snapshot there would steal the lead's delta.

Refs #3413, #4159

## Summary by Sourcery

Ensure turn file changes are measured and emitted for Command Deck sessions and newly bound sessions.

Bug Fixes:
- Report files created, edited, or deleted during Command Deck and raw turns in the Files tab, diff view, exports, and audit records.
- Ensure the first turn of a newly bound session reports only changes made after the session opened.

Enhancements:
- Unify turn-boundary handling so every turn owner records worktree changes before emitting the run terminator.
- Remove the terminator-only raw completion path that allowed drivers to omit file-change measurement.

Tests:
- Add regression coverage for first-turn worktree changes, baseline isolation, binding lock safety, and enforcement of complete turn-boundary handling.

### DIFF
--- a/crates/stella-cli/src/agent.rs
+++ b/crates/stella-cli/src/agent.rs
@@ -1219,19 +1219,15 @@ pub(crate) async fn run_turn(
         }
         engine.run_turn_with_sender(messages, budget, &tx).await
     };
-    // What this turn changed in the shared tree (#3413), measured at the
-    // boundary and emitted *before* the close below: these are the turn's own
-    // events and a consumer folding the stream must see them inside the turn
-    // they describe. See `crate::turn_files` for why a measurement, not a hook.
-    crate::turn_files::emit_shared_tree_changes(cfg, &tx, execution.as_ref());
     // This path owns its run — one raw engine turn, no pipeline above it — so
-    // it owes the run's terminator (#3379). The engine ends the turn with
-    // `TurnComplete` and deliberately says nothing about the run, and every
-    // other owner (the deck, fleet, goal, resume, the pipeline one-shot) goes
-    // through this same seam; without it a raw `stella run` ended on
-    // `turn_complete` and simply stopped, leaving every consumer waiting for a
-    // terminal event that never came. Last, after the turn's own events above.
-    persistence::emit_run_complete_for_turn(&tx, &cfg.model_id, &outcome);
+    // it owes both of the boundary's debts: what the turn changed in the shared
+    // tree (#3413), then the run's terminator (#3379). Emitted *before* the
+    // close below, because these are the turn's own events and a consumer
+    // folding the stream must see them inside the turn they describe. See
+    // `crate::turn_files::close_turn_boundary` for the ordering, for why the
+    // tree is measured rather than inferred from tool inputs, and for the deck
+    // defect that made the two one call.
+    crate::turn_files::close_turn_boundary(cfg, &tx, execution.as_ref(), &outcome);
     // The re-query adapter holds an `EventSender` clone of this run's channel
     // (#3366 telemetry), so it must be released here too — otherwise it keeps
     // the channel open and the renderer's `recv()` loop never ends (#2290).
--- a/crates/stella-cli/src/agent/persistence.rs
+++ b/crates/stella-cli/src/agent/persistence.rs
@@ -118,13 +118,13 @@ pub(crate) fn emit_run_complete_on_raw(
     emit_run_complete(&stella_core::EventSender::new(tx.clone()), model, cost_usd);
 }
 
-pub(crate) fn emit_run_complete_raw(
-    tx: &mpsc::UnboundedSender<AgentEvent>,
-    model: &str,
-    outcome: &TurnOutcome,
-) {
-    emit_run_complete_for_turn(&stella_core::EventSender::new(tx.clone()), model, outcome);
-}
+// `emit_run_complete_raw` — the turn-outcome sibling of the above — is
+// deliberately gone. It let a driver holding a raw sender pay the terminator
+// half of the turn boundary without the tree measurement that rides beside it,
+// and the Command Deck did exactly that for as long as both existed: every
+// session's Files tab read `no files touched yet` however many files its turns
+// wrote. `turn_files::close_turn_boundary_raw` is the replacement, and it pays
+// both. Restoring a terminator-only raw helper reopens that hole.
 
 #[derive(Default)]
 pub(crate) struct RendererOutcome {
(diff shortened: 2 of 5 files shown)


macanderson:
Heads-up from a parallel session working the same screen from the transcript side (#4150 / #4156, #4165).

Your Layer-1 finding here — that `emit_shared_tree_changes` has exactly one caller and `run_lead_turn` pays only the terminator — corrected a wrong conclusion of mine, so thank you: I had attributed a diffless edit row in the deck to a seq off-by-one, and the real live cause there is the missing producer this PR adds.

One thing worth knowing before this lands, because it is adjacent and easy to read as fixed-by-this:

**The deck's transcript row will still show no inline diff and no `+N −M` after this PR.** The Files tab will fill correctly. The transcript row will not, for a separate reason underneath:

- `crates/stella-tui/src/model.rs` stamps `InlineDiffRef { path, seq }` when a `ToolResult` folds, using `files[path].changes` at that moment.
- `close_turn_boundary` emits at the turn boundary — after every `ToolResult` of the turn has already folded. `touch_file` then bumps `changes` and `remember_diff` records at the **bumped** seq.
- `FileState::diff_at(seq)` therefore misses by exactly one, `resolve_inline_diff` returns `None`, and `render/entry.rs` gates both the inline diff and the `+N −M` metric on that lookup.

This is already live on the `agent::run_turn` path; giving the deck a producer surfaces it there too. Tracked in **#4155**, which stays open after this merges. The one-line `+1` is *not* the fix — with one aggregate `--numstat` change per path per turn, it would point every call touching that path at the same combined diff, which is the misattribution `resolve_inline_diff`'s doc exists to prevent. It needs a design decision, laid out as three options in #4155.

Nothing for you to change here — flagging it only so "Files tab still shows nothing in the transcript" is not read as a regression from this PR.

#4165 corrects the stale comment that hid both layers (it named a `FileChangeTap` that does not exist anywhere in the workspace); it touches `stella-tui` only, so it should not conflict with this.

Issue #2435, opened 2026-08-09
BudgetTick now carries a wall-clock axis that no surface renders — the HUD and the Observatory show dollars only


macanderson:
## Problem

#2240 added `AgentEvent::BudgetTick::deadline_remaining_ms`, so the journal can
now state whether a task deadline was armed and how much clock is left. Nothing
consumes it. The producing half shipped; the reading half did not, which is
exactly the unwired-code shape AGENTS.md § "Nothing left behind" asks to be
tracked rather than left implicit.

`BudgetTick`'s own doc comment says the HUD "renders spend live from this
stream; nothing user-visible about spend is derived from state that isn't also
in this event" — the wall clock is now in the event and in no HUD.

Concretely, every current consumer reads the dollar fields and ignores the new
one:

- `crates/stella-tui/src/model.rs` / `src/deck.rs` — the Command Deck's budget
  HUD folds `spent_usd` / `limit_usd` / the session pair
- `crates/stella-observatory/src/` — the telemetry dashboard's spend views
- `crates/stella-store/src/journal.rs:446` — projects `spent_usd` out of the
  tick and nothing else

## What to add

A remaining-clock readout beside the spend readout, wherever spend already
appears. `None` must render as *absent*, not as zero: "no deadline armed" and
"no time left" are the two facts the field exists to distinguish, and a HUD
showing `0s` for an unarmed run would re-introduce the confusion in the UI that
#2240 removed from the journal.

Scope is a judgement call worth making explicitly — the deck HUD is the
obvious one; whether the Observatory and `stella stats` want it too is part of
the work.

## Files

- `crates/stella-tui/src/model.rs`, `src/deck.rs`, `src/views/` — the fold and
  the render
- `crates/stella-tui/tests/snapshots/deck/` — golden frames; a new HUD cell
  moves columns, so regenerate with
  `BLESS=1 cargo test -p stella-tui --test deck_render_snapshots` and **read
  the diff**
- `crates/stella-observatory/src/db.rs` — if the dashboard is in scope
- `crates/stella-protocol/src/event.rs` — the field and its contract (no change
  needed; cite it)

## Constraints

- Golden frames are the test here (AGENTS.md § Testing approach): a `contains`
  assertion will not catch the column shift a new cell causes.
- `crates/stella-tui/src/deck_ui.rs`, `src/views/engine.rs`,
  `src/views/session.rs` and `src/deck_render.rs` are god files closed to
  growth — plan the new cell into a sibling module.
[...]

Issue #2240, opened 2026-08-08
protocol: budget_tick carries no wall-clock axis, so the ground-truth journal cannot say whether a task deadline was armed


macanderson:
## Problem

`stella-events.jsonl` is the ground truth for a bench trial (CLAUDE.md: "Open
`stella-events.jsonl` before claiming anything"). It cannot currently answer the
first question anyone asks about a timed-out trial: **was a wall-clock deadline
armed at all?**

`AgentEvent::BudgetTick` (`crates/stella-protocol/src/event.rs:521`) carries the
dollar axis and nothing else:

```json
{"type":"budget_tick","spent_usd":0.0011930198,"limit_usd":null,
 "mode":"enforced","session_spent_usd":0.0011930198,"session_limit_usd":15.0}
```

`cobol-modernization__kt3nQsB` in match `cc00894779ff` emitted **44** of these
against a $15 session budget it never approached (actual spend $0.87), while the
900s wall clock that actually killed it appears nowhere in the journal. Working
out that `BudgetGuard::set_task_deadline` had never been called took reading the
adapter source, `arenabench/arenabench/runner.py`, Harbor's `trial/trial.py`, and
each trial's `config.json` — for a fact one field would have stated.

## What to add

A `deadline_remaining_ms: Option<u64>` (name negotiable) on `BudgetTick`, with
`#[serde(default)]` so every already-written journal still parses. `None` means
"no deadline armed", which is exactly the distinction that was missing.

`BudgetTick` is the right carrier: it is already emitted after every call that
spends money, i.e. at exactly the safe boundaries where
`crates/stella-core/src/driver/settlement.rs::check_budget` consults the
deadline, so the field costs no new clock read and no new event type.

## Constraints

- **Additive only.** `AgentEvent` is the wire format for three surfaces (the
  TUI folds it, `--output-format stream-json` prints it, `stella-serve` streams
  it over SSE).
- **The generated wire contract must be regenerated in the same PR**:
  `make wire-schema-update` (`scripts/export-agentevent-schema.sh`) rewrites
  `docs/wire/agentevent.schema.json`, `agentevent.d.ts`, `serveframe.*`, and
  `make wire-schema` / `.github/workflows/wire-schema.yml` fail the gate if you
  forget. Note the exporter builds `stella-protocol` **and** `stella-serve`.
- **Emission stays where it is.** The guard is pure and takes `now` as an owned
  parameter (`BudgetGuard::check_deadline`); the caller supplies it from its own
[...]

PR #4186, opened 2026-08-21
fix(stella-tui): SPEC 5's status bar re-homes CLOCK, PR and LANES — three v1 cells get their surface back

### LINKED ISSUE #4126: tui-v2: SPEC 5's six-value status bar does not re-home CLOCK, PR or LANES — three v1 cells lost their surface
## Problem

PR #4123 replaced the v1 two-row statline with SPEC 5's one-row bar
(`worker · stage · ctx [bar] % · $spend · saved $x · ✉ n · ? help`).

SPEC 5 explicitly re-homes five of the cells it drops: "MODEL detail, CPU, MEM,
WARMTH, and ENGINE move behind `?` and the AGENTS tab." It says nothing about
three others that the v1 row also carried, and they now render **nowhere**:

| Cell | v1 source | Why it mattered |
|---|---|---|
| `CLOCK` | armed task deadline (`BudgetTick`) | The countdown to a deadline whose expiry is a SIGKILL. Armed deadlines were only recently wired (#3951), and a deadline you cannot see is one you cannot act on. |
| `PR` | `TranscriptEntry::Pr`, incl. `CiStatus::Failing` | v1 gave a failing PR an elevated priority weight so it survived the drop rule — it was designed to be hard to miss. |
| `LANES` | lead/subagent counts | Fan-out shape at a glance. |

`crates/stella-tui/src/statline.rs` still builds all three (`statline_items`),
so the projections exist; nothing renders them.

## Why this is not just "delete them"

`CLOCK` is the one with teeth. A task deadline that fires is a `SIGKILL`, and
the countdown was the only always-on surface for it. Dropping it silently is
the failure mode AGENTS.md invariant #10 exists to prevent.

## Options

1. Behind `?` and the AGENTS tab, alongside the five SPEC 5 already sends there.
2. `CLOCK` earns a place on the bar when and only when a deadline is armed —
   the bar's drop rule already handles conditional cells, and an armed deadline
   is rare enough that a permanent cell is not being asked for.
3. `PR` moves to the ISSUES tab (SPEC 9.4), `LANES` to AGENTS.

Option 2 for CLOCK plus option 1 for the rest is the shape I would suggest, but
SPEC 5 is the authority and it is currently silent — this needs a spec sentence,
not just code.

## Files

- `design/tui-v2/SPEC.md` §5 — the sentence that is missing.
- `crates/stella-tui/src/v2/status_bar.rs`, `src/v2/status_source.rs`.
- `crates/stella-tui/src/statline.rs` — the surviving v1 projections.

## Definition of done

SPEC 5 names a home for each of the three; the code puts them there; a test
proves an armed deadline is visible somewhere on a default deck.

Refs #4123

macanderson:
## What

SPEC 5's one-row status bar (#4123) re-homes five of the ten v1 statline cells and is silent about three others — `CLOCK`, `PR`, `LANES` — which since #4123 rendered **nowhere at all**. This gives SPEC 5 the sentence it was missing and puts each of them where the sentence says.

`CLOCK` is the one with teeth. An armed task deadline expires as a `SIGKILL`, and the countdown was its only always-on surface. Armed deadlines had only just been wired (#3951), so the gap between "visible" and "invisible" was about a day.

## The evidence it went unnoticed is in this diff

The golden test `deck_render_snapshots_pin_the_armed_deadline_statline` — whose own docstring said **"CLOCK beside SPEND"** — had been re-blessed to a frame with no clock in it. A golden pins whatever it is shown; it cannot pin what it is *for*. It now carries a `contains` assertion beside the grid, so the next blind bless fails instead of recording.

## Where each cell went

| Cell | Home | Why there |
|---|---|---|
| `CLOCK` | the bar, conditionally (SPEC 5) | it is the only fact on that row that says the run is about to *stop* |
| `PR` | ISSUES tab (SPEC 9.4) | a PR is the tracker-side artifact of the work that tab is about |
| `LANES` | AGENTS tab header (SPEC 9.5) | the pane under that header is one row per lane |

**CLOCK** sits third, right after the stage, and is the last cell the drop rule gives up — position *is* the priority, so the renderer needs no priority table (v1 had one, and it had already mis-resolved a tie at 150 columns). It reads `deadline 12m 34s`, `deadline expired` once crossed, and nothing at all when unarmed: `None` and `Some(0)` are different facts (#2240, #2435) and must not render alike.

Two colour rungs, not v1's three. Green means *pass* under SPEC 2 and an unfinished countdown has passed nothing; the palette has no amber that is not gold, and gold means stella acting. So `text` while there is time, `red` inside the last minute — a kill that close is a destructive event in progress, which is what §2 keeps red for. The label carries the state for anyone the colour does not reach (SPEC 13).

**PR** keeps the emphasis v1 gave failing CI (v1 raised its drop priority so a narrow row could not hide it) and gains the word beside the glyph — on a full-width tab, `✓`/`✗` differing only in colour and shape is exactly what "never colour alone" names.

**LANES** appends `✦ 1 lead · ◆ 2 sub` to the EXECUTIONS header, and stays absent for a solo run.

## Witness

Fail→pass demonstrated by ablation: reverting the one wiring line in `src/v2/project.rs` to `deadline_remaining_ms: None` fails `deck_render_snapshots_pin_the_armed_deadline_statline` with "the armed deadline reached no surface on a default deck", and nothing else. Restored, 14/14 pass. The re-blessed golden is the same evidence in diff form:

```
-zai/glm-5.2-air · execute · ctx ▌░░░░░░░░░░░ 4% · $3.79 · saved $0.00 · ✉ 0
+zai/glm-5.2-air · execute · deadline 12m 34s · ctx ▌░░░░░░░░░░░ 4% · $3.79 · saved $0.00 · ✉ 0
```

New tests:

- `an_armed_deadline_earns_a_cell_and_an_unarmed_run_does_not` — both halves in one test, because the pair is the property
- `a_crossed_deadline_reads_expired_not_zero`
- `the_countdown_takes_red_only_inside_the_last_minute` — and the word survives without the colour
- `a_narrow_row_drops_everything_before_it_drops_the_deadline` — at 46 columns, where the unarmed bar is down to worker and stage alone
- `the_alarm_is_the_only_red_the_bar_can_draw` — the other half of `a_healthy_status_bar_has_no_red_cells`
- `the_executions_header_carries_the_lane_split_once_a_session_fans_out`
- `deck_render_snapshots_state_a_failing_pr_on_the_issues_tab`

**Deleted tests:** none.

**Goldens re-blessed (4), diff read on each:** `statline_deadline_armed` (the countdown returns), `tab_agents` + `tab_agents_compact` (the lane split appears), `tab_issues` (the PR strip appears — the demo fixture already carried a PR).

## Also in this PR

Four doc comments still pointed at the deleted v1 statline for facts that now render elsewhere (`AgentEntry::is_subagent`, `PrInfo`, `WorkspaceModel::pr`, `Hud::deadline_remaining_ms`). Repointed, since they are about the exact cells this PR moves.

## Verification

`make gate CARGO_SCOPE="-p stella-tui"` — exit 0, all 32 steps.

## Remaining work filed

Two things noticed and deliberately not fixed here — issues linked in a comment below.

Closes #4126
Refs #4123

## Summary by Sourcery

Restore the missing v1 status information in its SPEC 5 destinations while making deadline and CI warnings explicit and testable.

New Features:
- Restore the armed task deadline countdown to the v2 status bar, including explicit expired-state text and urgency coloring.
- Re-home pull request status to the ISSUES tab and lane counts to the AGENTS executions header.

Bug Fixes:
- Ensure deadline visibility distinguishes unarmed runs from expired deadlines and remains available on narrow status bars.
- Prevent golden snapshots from being re-blessed without verifying that the armed deadline is rendered.

Enhancements:
- Preserve accessible textual status for deadline alarms and CI failures instead of relying on color or glyphs alone.

Documentation:
- Document the new CLOCK, PR, and LANES surfaces in the v2 TUI specification.

Tests:
- Add coverage for deadline rendering, expiration, alarm thresholds, narrow-bar retention, lane splits, and failing PR status.

Chores:
- Update comments and snapshot goldens to reflect the relocated status cells.

### DIFF
--- a/crates/stella-tui/src/deck.rs
+++ b/crates/stella-tui/src/deck.rs
@@ -321,8 +321,8 @@ impl AgentEntry {
     }
 
     /// Whether this lane is a subagent (registered with the `subagent` role) —
-    /// the split the SESSION tab's nested rows and the statline's
-    /// `✦ lead · ◆ sub` counts read.
+    /// the split the SESSION tab's nested rows and the AGENTS tab's
+    /// `✦ lead · ◆ sub` header counts read.
     pub fn is_subagent(&self) -> bool {
         self.meta.role == "subagent"
     }
@@ -331,7 +331,8 @@ impl AgentEntry {
 /// The deck's PR read-model: the latest `AgentEvent::Pr` observation, from
 /// whichever agent emitted it. A session tells one PR story at a time — the
 /// newest event wins outright, so a CI update on the same PR simply replaces
-/// the snapshot in place. Drives the statline's PR cell.
+/// the snapshot in place. Drives the ISSUES tab's PR strip (SPEC 5 §9.4),
+/// which is where the v1 statline's PR cell went.
 #[derive(Clone, Debug, PartialEq)]
 pub struct PrInfo {
     pub url: String,
@@ -360,7 +361,7 @@ pub struct WorkspaceModel {
     /// events). Drives the status-bar gauge (and, later, dispatch backpressure).
     pub global_cpu_pct: f32,
     /// The latest PR observation across all agents (`AgentEvent::Pr` from the
-    /// fleet PR/CI monitor) — the statline's PR cell. Latest event wins;
+    /// fleet PR/CI monitor) — the ISSUES tab's PR strip. Latest event wins;
     /// `None` until a PR has been seen this session.
     pub pr: Option<PrInfo>,
     /// Last pin observed serving each of the three pipeline roles — the
--- a/crates/stella-tui/src/model.rs
+++ b/crates/stella-tui/src/model.rs
@@ -510,9 +510,11 @@ pub struct Hud {
     /// Wall clock left before the task deadline, as the last `BudgetTick`
     /// reported it (#2240, #2435). `None` is the load-bearing case and means
     /// **no deadline is armed** — never "no time left", which is
-    /// `Some(0)`. The statline renders the two differently for exactly that
+    /// `Some(0)`. The status bar renders the two differently for exactly that
     /// reason: a HUD showing `0s` for an unarmed run would put back into the
-    /// UI the confusion #2240 took out of the journal.
+    /// UI the confusion #2240 took out of the journal. `None` draws no cell at
+    /// all and `Some(0)` draws the word `expired`
+    /// ([`crate::v2::status_bar`]).
     ///
     /// Milliseconds rather than a `Duration` because that is the wire shape
     /// (`AgentEvent::BudgetTick::deadline_remaining_ms`), and this struct is a
--- a/crates/stella-tui/src/v2/project.rs
+++ b/crates/stella-tui/src/v2/project.rs
@@ -69,6 +69,12 @@ pub fn status<'a>(model: &'a WorkspaceModel, ui: &DeckUi, worker: &'a str) -> St
         spend_usd: model.total_cost(),
         saved_usd: model.total_cache_savings_usd(),
         inbox: ui.notifications.iter().filter(|n| !n.read).count() as u32,
+        // Absent unless a deadline is armed, and the fold already keeps that
+        // distinction: `BudgetTick` carries `None` for an untimed run and
+        // `Some(0)` for one that is out of time (#2435). Passed through
+        // rather than defaulted, because a `0` invented here would render as
+        // `expired` on a run nobody was timing.
+        deadline_remaining_ms: focused.and_then(|a| a.model.hud.deadline_remaining_ms),
     }
 }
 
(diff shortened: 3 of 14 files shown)


macanderson:
Remaining work filed while doing this, per AGENTS.md § Nothing left behind:

- #4187 — the v2 status bar has two projections of the same six values and the deck renders the wrong one; the shipping bar prints `context_recall` where SPEC 5 says `context recall`. Found because this PR had to add `deadline_remaining_ms` to both.
- #4188 — the other half of #4126: SPEC 5 sends MODEL detail / CPU / MEM / WARMTH / ENGINE behind `?` *and* the AGENTS tab. AGENTS has them; `render_help` renders keybindings only.

Neither is fixed here — #4187 needs a delete-one-of-two decision and #4188 needs either a new overlay section or a spec amendment, and both are wider than the three cells this PR is about.

PR #4189, opened 2026-08-21
perf(stella-tools): the chunk embedding pass batches across files and overlaps its round trips (#4144)

### LINKED ISSUE #4144: search: the chunk embedding pass is one round trip per file, strictly serial — ~2000 sequential requests to fill this repo's index
## Problem

The chunk (symbol) rung of the semantic index embeds **one file per HTTP round trip, strictly serially**. On this repository (1716 files, 29974 symbols) that is ~2000 sequential requests to the embedding backend, and it dominates the wall clock of `stella init` and of the session-start backfill.

The asymmetry between the two rungs is the bug in one sentence:

- **File rung** batches 32 **files** per request — `crates/stella-tools/src/search/semantic.rs:163` asks `files_pending_embedding(&fingerprint, want)` with `want = EMBED_BATCH`. 1716 files = 54 requests.
- **Chunk rung** batches 32 **chunks within a single file** — `crates/stella-tools/src/search/engine.rs:783`, `for batch in needed.chunks(EMBED_BATCH)` over one `PendingChunkFile`. Average file here has 17 symbols, so the typical request is half-empty and **every file costs its own round trip**.

And nothing overlaps: `warm_chunks_opened` (`crates/stella-tools/src/search/engine.rs:958-972`) is

```
for file in &scan.files {
    embed_and_store_chunk_file(graph, embedder, &fingerprint, file).await
}
```

There are no concurrency primitives anywhere in `crates/stella-tools/src/search/` — `rg 'join_all|buffer_unordered|FuturesUnordered|tokio::spawn|try_join'` over that directory returns only two doc-comment mentions, no call sites. One request is in flight at a time for the whole pass.

## Measured

Against the real `api.voyageai.com/v1/embeddings` endpoint (`voyage-code-3`), with payloads rebuilt from the live `.stella/private/codegraph.db` exactly as `render_chunk_text` renders them (`path :: name (kind)\n\n` + body truncated at `MAX_CHUNK_CHARS = 4000`), batched 32-per-request as the Rust does:

```
  7.07s   156 sym   134kchar  reqs=5  bench/harbor_adapter/tests/test_secure_launcher.py
  7.19s   141 sym   167kchar  reqs=5  bench/harbor_adapter/tests/test_adapter.py
  5.28s   131 sym   119kchar  reqs=5  arenabench/tests/test_arena.py
  1.41s    24 sym    42kchar  reqs=1  crates/stella-observatory/src/tests.rs
  1.30s    24 sym    10kchar  reqs=1  crates/stella-serve/src/controls.rs
  1.21s    13 sym     8kchar  reqs=1  arenabench/tests/test_cloud_merge.py
  1.16s    13 sym     5kchar  reqs=1  crates/stella-embed/src/seam.rs
```

The backend is not the constraint: 32 short texts in one request return in **0.28s**, and a full 32x6000-char request in **1.40s**. The cost is round-trip count, not backend throughput.

At the measured ~1.3s median that is roughly **35-45 minutes** to fill this repository's chunk rung, with the 259 files carrying more than 32 symbols costing 5-7s each. A user watching the narration reports it as "about 1 file every 10 seconds" — the observed rate is worse than the 1.3s measured here in isolation, and the unmeasured contributor is contention: three processes hold this `codegraph.db` open concurrently in a normal session (two `stella` processes plus `stella observe`), and every file is its own write transaction.

## Why it is shaped this way

`store_chunk_vectors` (`crates/stella-graph/src/vectors/chunks.rs:340-351`) sweeps rows by `file_sha256` after upserting, so a file's chunks must be written **whole** — writing half a file's chunks and then the other half would have the second write delete the first's rows. That constraint is real and must be preserved.

But it constrains the **write**, not the **embed**. `embed_and_store_chunk_file` couples the two into one function, and that coupling is what forces the serialization.

## Definition of done

Either fix independently helps; together they are roughly an order of magnitude.

1. **Overlap the round trips.** Run N files' embedding concurrently (`buffer_unordered(8)` or equivalent), then store each file's vectors whole as results arrive. The per-file atomic write is preserved because only the embed is concurrent. Expect ~8x.
2. **Batch chunks across files.** Gather chunks from the whole 64-file scan window (`MAX_FILES_PER_CHUNK_PASS`) into full 32-chunk requests, then scatter the vectors back per file for storage. At 17 average symbols/file this roughly halves request count on its own.

Constraints a fix must respect:

- Per-file atomicity of `store_chunk_vectors` — see the sweep argument above.
- The pass must stay interruptible with everything committed so far durable (`ChunkWarmOutcome::Failed` carries `files_embedded`).
- Concurrency must be bounded — an unbounded fan-out at 32 chunks x 4000 chars per request will hit backend rate limits, and `HttpEmbedder::embed` (`crates/stella-embed/src/http.rs:332-403`) has **no retry or backoff**: any non-2xx becomes `EmbedError::Backend` and fails the entire pass. If concurrency is added, a 429 path is needed with it, or a pass that used to be merely slow becomes one that aborts.
- `warm_opened` and `warm_chunks_opened` are generic over the progress callback for `Send`-ness reasons documented at `semantic.rs:147-152`; a concurrent rewrite must keep both callers (the non-`Send` `stella init` emitter and the spawned background task) compiling.

## Verify

Witness: a counting embedder that records the number of round trips and the batch size of each, driven over a fixture workspace of K files with M symbols each. On `main` the call count equals the file count; with the fix it equals `ceil(total_chunks / 32)`. `crates/stella-tools/src/search/backfill.rs`'s existing `CountingEmbedder` is the instrument — it already counts calls.

## Related

- #4041 batched the (then lazy) pass; #4043 moved it off the query path entirely. Neither addressed cross-file batching or concurrency — #4043's premise is that the pass is unbounded and in the background, which makes its throughput the thing that decides when the index is actually usable.
- #4047 (the embed lease can expire under a long backfill) is a direct consequence of this pass being slow: shortening the pass shrinks that window.
- #4048 (the prompt gate covers the embedding pass) — the gate holds the first prompt while this runs, so this latency is user-visible waiting.

macanderson:
## What

The chunk (symbol) rung of the semantic index embedded **one file per HTTP round trip, strictly serially**. The average file in this repository carries 17 symbols, so the typical 32-slot request went out half empty and every file cost its own round trip — ~2000 sequential requests to fill this repository's index, dominating the wall clock of `stella init` and of the session-start backfill.

Both halves of #4144's definition of done:

1. **Batches span files.** The pending chunks of the whole 64-file scan window are flattened and cut into full 32-chunk requests, so a request is short only when the work ran out.
2. **Round trips overlap.** Up to `CHUNK_EMBED_CONCURRENCY` (8) requests are in flight at once.

Per the issue's constraints, the per-file atomic write is untouched — `store_chunk_vectors` sweeps rows by the file's content hash, so a file's rows still go in together. Batching is an embedding-side change only.

## The three defects found while building it

Each of these is a thing the obvious implementation gets wrong, and each has a witness.

**Responses were routed by arrival order.** `buffer_unordered` yields results as they *complete*, not in the order requests were issued, so pairing the Nth completion with the Nth batch files one request's vectors against whichever chunks were in a slower one. Nothing errors — every chunk still receives a well-formed vector of the right width — and the only symptom is a semantic index that answers with the wrong symbols. This is the same misattribution `HttpEmbedder::embed` already refuses one layer down when it routes rows by `index` rather than by arrival. Each response now carries the ordinal of the request it answers.

**A window was all-or-nothing.** Holding all 64 files until every request returned put the whole window at risk of one late failure. A file's chunks can straddle two requests now, so "this file is done" is an outstanding count reaching zero; when it does, that file is stored immediately. A failure costs the files still in flight, not the ones already paid for.

**The failure arm threw away the count.** Returning `Result` meant a failed window reported `files_embedded: 0` while thirty-two files sat durably in the store. The count and the ending are independent facts, so the window returns both (`ChunkWindow`).

## The 429 path

The issue makes this a condition on adding concurrency at all: `HttpEmbedder::embed` had no retry and no backoff, so any non-2xx became a terminal `EmbedError::Backend`. Concurrency is exactly what provokes a rate limit, so the pass that got faster would have become the pass that aborts.

`429` and `5xx` are now re-issued up to 4 attempts with exponential backoff, honouring `Retry-After`'s delta-seconds form under an 8-second ceiling. The HTTP-date form is ignored deliberately rather than trusting two clocks to agree — falling back to the local schedule is wrong by seconds; trusting a bad date is wrong by however far the clocks differ. A `4xx` that will say the same thing next time is issued exactly once, and the error names the attempt count so a slow pass is not mistaken for one that never waited.

## Witnesses

Every one was checked the artisanal way — run against the code it is meant to catch, and confirmed to fail there.

| Witness | Fails on | Observed |
|---|---|---|
| `the_chunk_pass_batches_across_files` | `main` (serial) | **62 calls** for 60 files; passes with the fix |
| `a_small_window_costs_one_chunk_request` | `main` (serial) | **5 calls** for 4 files, must be 2 |
| `every_chunk_keeps_the_vector_computed_for_it` | arrival-order routing | `thing_0`'s slot held **`thing_38`'s vector** |
| `files_finished_before_a_failure_stay_committed` | all-or-nothing window, and the discarded count | `files_embedded: 0` with 32 files in the store |
| `a_rate_limited_request_is_retried_rather_than_failed` | no-retry backend | `Backend(... HTTP 429 ...)` instead of a vector |
| `a_backend_that_only_refuses_ends_as_an_error_naming_the_attempts` | unbounded or silent retry | — |
| `an_unauthorized_request_is_not_retried` | retrying a terminal 4xx | — |

`every_chunk_keeps_the_vector_computed_for_it` reads back through `rank_chunks_by_vector` — the path a search actually uses — so the assertion is about what a user is told, not about a row shape. It uses an embedder that completes in reverse issue order, making the misattribution deterministic rather than a race the test might win.

Note that this one **passes on `main`**: serial routing was correct. It is not a witness for #4144 itself but the guard for the concurrency this PR introduces, which is why it is here.

## A defect this PR's own testing nearly shipped

Worth recording, because it is a trap for the next person. My first routing fix carried the batch slice through the futures, which put a lifetime on them. `cargo test -p stella-tools` was green throughout. The failure appears only at the `tokio::spawn` in `crates/stella-cli/src/agent/graph.rs:697` — five "implementation of `Send` is not general enough" errors pointing at a spawn that mentions no chunk, no batch and no embedder. `NeededChunk` therefore owns its text, and only a `usize` ordinal travels through a future. The reasoning is in a comment at the struct, since nothing local explains why an owned `String` is deliberate there.

## Prose repaired

- `crates/stella-graph/src/vectors/chunks.rs` cited `embed_and_store_chunk_file` — a name that no longer exists — and its module header read "embedded one at a time". The per-file atomicity that doc actually turns on is unchanged, and now says explicitly that it is a property of the write rather than of the embed.
- `crates/stella-embed/README.md`'s "Semantics worth knowing" had no retry entry at all, leaving a reader assuming one-shot.

**Deliberately not changed:** `crates/stella-tools/src/search/enrich.rs` quotes `embed_and_store_chunk_file` inside a record of a real measured session, and `tests/fixtures/name_rung_corpus.jsonl` is a frozen corpus snapshot whose own docs say stale rows there are "the snapshot working as intended, not drift to repair". Rewriting either to match today's tree would falsify evidence.

## Dependencies

- `futures-util` moves from dev-dependency to dependency in `stella-tools` (`buffer_unordered`). It was already carried for `poll!` in tests; tests see a normal dependency, so no second entry is needed.
- `tokio` (feature `time` only, `sleep`) becomes an optional dependency of `stella-embed`, gated on the existing `http` feature. `reqwest` already requires tokio, so no new tree is added and a host that never enables `http` links no runtime.

## Not measured

The issue's ~35-45 minute estimate is from its own instrumentation against the live Voyage endpoint. **This PR does not re-measure it.** The witnesses prove the request count collapses from one-per-file to `ceil(total_chunks / 32)` and that up to 8 overlap; converting that to a wall-clock figure against a real backend is a separate measurement I have not run, and I would rather claim nothing than quote the issue's number as if this branch had reproduced it.

## Verification

`make gate CARGO_SCOPE="-p stella-cli -p stella-context -p stella-embed -p stella-fleet -p stella-graph -p stella-observatory -p stella-runtime -p stella-tools -p stella-tui"` (the scope `make impacted` selects for this diff) — green through its last step, `self-driving-test: OK — 61 checks passed`.

Closes #4144

## Summary by Sourcery

Accelerate chunk embedding by batching work across files, overlapping bounded HTTP requests, and safely handling retries and partial failures.

New Features:
- Batch chunk embeddings across files and process up to eight embedding requests concurrently.
- Retry rate-limited and server-error embedding requests with bounded exponential backoff and supported Retry-After delays.

Bug Fixes:
- Preserve correct chunk-to-vector association when concurrent requests complete out of order.
- Commit completed files independently when a later embedding request fails and report the number of files durably embedded.
- Avoid retrying terminal client errors such as unauthorized requests.

Enhancements:
- Retain per-file atomic storage while allowing embedding work to span files and requests.

Build:
- Promote futures-util to a runtime dependency for bounded concurrent chunk embedding.
- Add optional tokio timer support to the HTTP embedding feature.

Documentation:
- Document HTTP embedding retry behavior and clarify that per-file atomicity is preserved at storage time.

Tests:
- Add coverage for cross-file batching, request-count reduction, out-of-order response routing, partial commits after failure, and bounded retry behavior.

### DIFF
--- a/crates/stella-embed/Cargo.toml
+++ b/crates/stella-embed/Cargo.toml
@@ -20,10 +20,15 @@ async-trait.workspace = true
 # pay nothing for a transport they do not use; a host that needs it turns it on.
 reqwest = { workspace = true, optional = true }
 serde_json = { workspace = true, optional = true }
+# `sleep` only — the backoff between retries of a rate-limited embedding
+# request. Optional with the transport for the same reason as the two above:
+# a host that never enables `http` links no runtime for a timer it cannot
+# reach. `reqwest` already requires tokio, so this adds no new tree.
+tokio = { workspace = true, optional = true, features = ["time"] }
 
 [features]
 default = []
-http = ["dep:reqwest", "dep:serde_json"]
+http = ["dep:reqwest", "dep:serde_json", "dep:tokio"]
 
 [dev-dependencies]
 tokio = { workspace = true, features = ["rt-multi-thread", "macros"] }
--- a/crates/stella-embed/README.md
+++ b/crates/stella-embed/README.md
@@ -94,6 +94,16 @@ silent quality regression.
   output that reaches the prompt is byte-stable for identical input
   (invariant 7). Two property tests pin it: ranking is independent of input
   order, and `limit` is always a prefix of the full ranking.
+- **A rate limit is an instruction, not a failure.** `HttpEmbedder::embed`
+  re-issues a `429` or a `5xx` up to four attempts with exponential backoff,
+  honouring `Retry-After`'s delta-seconds form under an 8-second ceiling — a
+  backend may ask for a delay, but not park a session-start backfill for an
+  hour. The HTTP-date form is deliberately ignored rather than trusting the
+  caller's clock to agree with the server's. A `4xx` that will say the same
+  thing next time is issued exactly once, and the error names how many
+  attempts were made so a slow pass is not mistaken for one that never waited.
+  This matters because the chunk embedding pass keeps several requests in
+  flight (#4144), and concurrency is precisely what provokes a rate limit.
 - **The default admission floor is provisional, not measured.** The
   `SimilarityPosture` contract asks for a floor derived from a model's observed
   relevant/irrelevant score distributions on a real corpus, and no such
--- a/crates/stella-embed/src/http.rs
+++ b/crates/stella-embed/src/http.rs
@@ -405,17 +405,18 @@ impl Embedder for HttpEmbedder {
                 if let Some(key) = &self.api_key {
                     request = request.bearer_auth(key);
                 }
-                let response = request.send().await.map_err(|error| {
-                    EmbedError::Backend(format!("{}: {error}", self.endpoint))
-                })?;
+                let response = request
+                    .send()
+                    .await
+                    .map_err(|error| EmbedError::Backend(format!("{}: {error}", self.endpoint)))?;
 
                 let status = response.status();
                 if status.is_success() {
                     break response;
                 }
 
-                let worth_retrying = status == reqwest::StatusCode::TOO_MANY_REQUESTS
-                    || status.is_server_error();
+                let worth_retrying =
+                    status == reqwest::StatusCode::TOO_MANY_REQUESTS || status.is_server_error();
                 if !worth_retrying || attempt >= MAX_ATTEMPTS {
                     let body = response.text().await.unwrap_or_default();
                     let body: String = body.chars().take(400).collect();
--- a/crates/stella-embed/src/http.rs
+++ b/crates/stella-embed/src/http.rs
@@ -47,6 +47,33 @@ use crate::seam::{EmbedError, Embedder, EmbedderFingerprint, Embedding, Similari
 /// bounded because this sits on a tool call the agent is waiting on.
 const REQUEST_TIMEOUT: Duration = Duration::from_secs(30);
 
+/// How many times one embedding request is issued before its failure becomes
+/// the caller's.
+///
+/// **A rate limit is not an error, it is an instruction.** A backend that
+/// answers `429` is saying "ask again shortly"; treating that as a terminal
+/// failure hands the caller a dead pass over a condition that resolves by
+/// itself. This mattered little while the chunk embedding pass issued one
+/// request at a time and merely ran slowly, and it matters now that the pass
+/// keeps several in flight (#4144) — concurrency is exactly what provokes a
+/// rate limit, so the pass that got faster would otherwise be the pass that
+/// started aborting.
+///
+/// Bounded, because retrying forever is how a caller waits on a backend that
+/// is never going to answer. Four attempts spans roughly seven seconds of
+/// backoff, which clears an ordinary burst limit without turning a genuine
+/// outage into a long silence.
+const MAX_ATTEMPTS: u32 = 4;
+
+/// The first pause between attempts; each subsequent one doubles.
+const INITIAL_BACKOFF: Duration = Duration::from_millis(500);
+
+/// The ceiling on any single pause — including one a `Retry-After` header
+/// asks for. A backend is entitled to ask for a delay; it is not entitled to
+/// park a session-start backfill for an hour, and a header this large is more
+/// often a misconfigured proxy than a real instruction.
+const MAX_BACKOFF: Duration = Duration::from_secs(8);
+
 /// The admission floor applied when nothing overrides it.
 ///
 /// **This number is provisional and is not a measured separation point.** The
@@ -314,6 +341,27 @@ impl fmt::Debug for HttpEmbedder {
     }
 }
 
+/// How long the backend asked us to wait, if it said so in a form worth
+/// trusting.
+///
+/// Only the delta-seconds form of `Retry-After` is read. The HTTP-date form
+/// is legal and is deliberately ignored: honouring it means trusting the
+/// caller's clock to agree with the server's, and a skewed clock turns a
+/// two-second pause into a two-hour one. Falling back to the local backoff
+/// schedule is wrong by at most a few seconds; trusting a bad date is wrong
+/// by however far the clocks differ.
+fn retry_after(response: &reqwest::Response) -> Option<Duration> {
+    response
+        .headers()
+        .get(reqwest::header::RETRY_AFTER)?
+        .to_str()
+        .ok()?
+        .trim()
+        .parse::<u64>()
+        .ok()
+        .map(Duration::from_secs)
+}
+
 #[async_trait]
 impl Embedder for HttpEmbedder {
     fn fingerprint(&self) -> EmbedderFingerprint {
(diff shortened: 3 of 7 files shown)


macanderson:
Remaining work filed per AGENTS.md § Nothing left behind:

- **#4190** — the **file** rung (`semantic.rs:162-184`) still issues its 54 batches strictly serially. Its batching is already correct (one file = one vector, so requests are full); only the overlap is missing. It runs *before* the chunk rung, so its latency is additive to the same pass the prompt gate holds. Written up with the three traps this PR hit, so a fresh agent does not re-discover them.
- **#4191** — this PR's speedup is proved **structurally** (request count, overlap) and **not measured** against a real backend. `CHUNK_EMBED_CONCURRENCY = 8` and the retry ladder's constants are both defensible-but-unmeasured, and whether 8-way concurrency provokes sustained 429s against Voyage is exactly the thing that could put the real speedup below the structural ~8x.

Issue #4196, opened 2026-08-21
tui-v2: the P2 plan orders a syntect crate the deck must not have — IMPLEMENTATION-PLAN and #4036 contradict each other


macanderson:
## The disagreement

`design/tui-v2/IMPLEMENTATION-PLAN.md` § 2 and § 3 say P2 builds a new crate:

> `stella-tui-highlight`: syntect wrapper, theme build, per-event cache.
> | `syntect` | syntax highlighting | `fancy-regex` feature to avoid oniguruma on Windows |

#4127 shipped P2's result row and deliberately did **not** build it, on the
instruction in #4127's own body:

> Note `stella_transcript::syntax::{body_paint, paint_line}` already exists and
> is shared with the export and Observatory renderers — the deck keeps only the
> palette (`syntax::tok_style`). Do not fork it; the whole point of #4036 was
> that three copies of one rendering decision drift.

Those two cannot both be followed. A syntect-based `stella-tui-highlight` would
be a **fourth** lexer for the same bodies the deck, the export grid and the
Observatory HTML renderer already share through `stella-transcript::syntax` — the
exact drift #3644 and #4036 each closed once. So #4127 reused the shared lexer
and built only the half the plan is actually right about: the **cache**
(`SessionFold`'s tail reuse in `crates/stella-tui/src/views/session/fold.rs`,
witnessed by `an_unchanged_tail_is_highlighted_once_not_once_per_frame`).

## What is actually still open

The shared lexer is hand-written and covers a fixed set of languages
(`stella_transcript::syntax::Lang`); syntect covers far more and is grammar-
accurate where the hand-written one is heuristic. That is a real difference, and
the plan's § 5 already names the mitigation — "the highlighter is behind a
trait; swap to tree-sitter per-language later without touching widgets". So the
question is not "syntect or the shared lexer" but **where a better lexer would
land**: in `stella-transcript` (all three surfaces gain it, no fork) or in a
deck-only crate (the deck diverges from the export, which is what #4036 forbade).

## Definition of done

One of:

- `IMPLEMENTATION-PLAN.md` § 2 drops the `syntect` row and § 3's P2 bullet
  stops naming `stella-tui-highlight`, recording that the lexer is
  `stella-transcript`'s by decision and that P2's deliverable was the cache; or
- a decision to upgrade the lexer, filed as its own issue against
  `stella-transcript` so all three renderers get it at once — never as a
  deck-local crate.

Either way the plan and the code should agree, because the next agent picking up
[...]

Issue #4195, opened 2026-08-21
tui-v2: SPEC 6.4 tells a diff renderer to paint over its own rail — Line.style is wrong wherever a row carries a margin


macanderson:
## The trap

SPEC 6.4 (`design/tui-v2/SPEC.md`) specifies diff rows this way:

> Diff rows are two-layer: `Line.style` carries the bg tint (`diff_add_bg` or
> `diff_del_bg`), spans keep syntax fg. ratatui composes them per cell.

Taken literally that is wrong for the transcript, and #4127 could not implement
it as written. A transcript diff row is not just the diff:
`render::row::push_diff_line` prepends the block's rail (`" │"` plus the body
indent) to every row, and the rail carries the **call's** metal (SPEC 6.2), not
the hunk's. `Line.style` paints the whole row's area — rail included — so a
green band would swallow the gold or silver rail that says which event the diff
belongs to.

#4127 therefore renders the two layers as: per-span `bg` for the code (already
how `crates/stella-tui/src/diff.rs` does it), plus a trailing padded span
carrying the tint to the pane edge so the ground reads as a band rather than
tracing the ragged end of the code. Same visual result, margin excluded. The
reasoning is in `push_diff_line`'s doc comment
(`crates/stella-tui/src/render/row.rs`) and the assertion is
`a_rust_diff_renders_add_and_remove_rows_per_spec_64` in
`crates/stella-tui/src/render/tests/result_row.rs`, which checks both that the
band reaches the pane edge and that the rail span's `bg` stays `None`.

## Why this is a defect in the SPEC, not just a note

The next person to implement a diff surface from SPEC 6.4 — the plugin panel
host (SPEC 12), a second transcript renderer, the Observatory — will read
`Line.style`, implement it, and paint over whatever margin their surface has.
The spec should state the constraint rather than one mechanism that happens to
violate it.

## Definition of done

`design/tui-v2/SPEC.md` § 6.4's diff bullet states the two-layer rule in terms
that survive a row with a margin: the tint covers the **code column** out to the
pane edge and never the rail or gutter chrome to its left; `Line.style` may be
named as one way to get there, flagged as wrong wherever a row carries chrome
the tint must not cover. No code change is implied — #4127 already renders it
correctly, and its test is the reference for what "correct" means.

Refs #4127

Issue #4168, opened 2026-08-21
stella-tui v2: the call head lost the dim-directory/bright-basename split, so a column of paths reads as one string


macanderson:
## What is broken

A call row's path renders as one flat colour. The v1 call row split it —
directory dim, basename bright — so a scan down the margin landed on the
file *identity* rather than on the shared prefix every path in a workspace
has. The v2 head that replaced that row draws the whole subject in one
style, so a column of calls reads as a column of near-identical
`crates/stella-tui/src/render/…` strings.

## Evidence

Observed, not inferred. `head_rows`, span by span:

```
" │ ● read crates/stella-tui/src/render/row.rs · 0 lines"
   spans: [" │", " ● ", "read", " crates/stella-tui/src/render/row.rs", " · 0 lines"]
```

The whole path is one span. `transcript::head_row` builds it as:

```rust
left.push(Span::styled(format!(" {}", event.subject), text));
```

## What was lost

`render::row::path_spans` did this, and was deleted in the PR that filed
this issue — it had been kept alive only by the unreachable v1 arm #4157 was
about, and dead code in the tree is not a specification. The rule it
encoded, which is the part worth keeping:

> Split a path into (directory, basename) spans so the basename carries the
> emphasis. In a scan, the file *identity* is what the eye is hunting; the
> directory is context that only matters once you've found the file, so it
> recedes. Non-path text renders as a single unemphasised span.

Implementation, for reference (it is in git history at
`crates/stella-tui/src/render/row.rs`, before the deletion):

```rust
match text.rfind('/').filter(|_| is_path) {
    Some(cut) => vec![
        Span::styled(text[..=cut].to_owned(), dim),
        Span::styled(text[cut + 1..].to_owned(), bright),
    ],
    None if is_path => vec![Span::styled(text.to_owned(), bright)],
    None => vec![Span::styled(text.to_owned(), dim)],
}
```

Note the third arm: **non-path** text (a `bash` command line, an
unrecognised tool's raw input) is deliberately *not* brightened. The subject
of a `bash` head is a command, not a file, and giving it the same emphasis
would spend the contrast on the rows that least need it.

## Definition of done

`transcript::head_row` renders a subject that is a path as dim-directory
plus bright-basename, and a subject that is not as it does today. The event
already knows which it is at construction time —
`transcript_source::subject_for` takes `path: Option<&str>` and only falls
back to the raw input when there is none — so the distinction needs
[...]

Issue #4157, opened 2026-08-21
stella-tui: ctrl+o on a tool call reveals nothing — the v2 head router made the expanded-arguments arm unreachable


macanderson:
## What is broken

`ctrl+o` on a **tool call row** used to reveal the call's full argument
object, pretty-printed and syntax-coloured. It reveals nothing now. The row
renders identically expanded and collapsed, so the affordance is inert and
the arguments a call was dispatched with are unreachable from the deck.

The result row's `ctrl+o` still works — this is the *call* half only.

## Evidence

Observed, not inferred. With a probe test in
`crates/stella-tui/src/render/tests/`:

```rust
let e = TranscriptEntry::ToolStart {
    call_id: "c1".into(), name: "bash".into(), input: "ls".into(),
    raw: "{}".into(), path: None,
};
let mut collapsed = Vec::new();
entry_lines(&e, &[], false, /* expanded */ false, false, 100, &mut collapsed);
let mut expanded = Vec::new();
entry_lines(&e, &[], false, /* expanded */ true,  false, 100, &mut expanded);
```

Both produce exactly one row:

```
collapsed=[" │ ● run ls"]
expanded= [" │ ● run ls"]
```

## Why

`crates/stella-tui/src/render/entry.rs`:

```rust
pub(crate) fn entry_lines(entry, files, expand_thinking, expanded, live, width, out) {
    if !v2_rows(entry, width, out) {
        entry_body(entry, files, expand_thinking, expanded, live, width, out);
    }
    ...
}
```

`v2_rows` matches `TranscriptEntry::ToolStart { .. }` unconditionally, calls
`v2::transcript_source::head_rows(name, path, input, width)` and returns
`true`. `head_rows` takes no `expanded` parameter. So `entry_body` never runs
for a `ToolStart`, and the `expanded` flag is never consulted for that entry
kind.

The consequence is bigger than one dead flag: **the whole
`TranscriptEntry::ToolStart` arm of `entry_body` is unreachable** — roughly
`entry.rs:401-465` on `main`, including the `serde_json::to_string_pretty`
of `raw` and its `push_body_line` loop. It compiles, it is covered by no
guard (`dead-code-allows` and `module-reachability` see items, not match
arms), and it will keep passing review as if it were live code.

## Definition of done

Either:

1. **Restore the feature in v2.** Give `head_rows` (or a sibling) the
   `expanded` flag and have it emit the pretty-printed argument object as
[...]

PR #4198, opened 2026-08-21
feat(stella-tui): the tool result joins its call — one event, one rail, every feature kept

### LINKED ISSUE #4127: tui-v2 P2: restyle the tool-result row through v2 without losing highlighting, inline diffs, gutters or the truncation notice
## Problem

PR #4123 routed a tool call's **head** through the v2 renderer but deliberately
left the **result** row on v1. Today the deck draws a v2 rail above a v1 result:

```
 │ ● apps/app/automations/page.tsx
  ⎿ 312 lines                                                          42ms
```

The result row should carry the same rail and metal (SPEC 6.2) and the body
styling SPEC 6.4 specifies.

## Why it was not done in #4123

It was attempted. A v2 `result_rows` was written and **16 tests went red for the
right reason** — it dropped four features that all landed within the preceding
day:

- syntax highlighting of the body in the file's own language (#4019, #4036)
- word-level inline diffs (`resolve_inline_diff`, `INLINE_DIFF_CAP`)
- the emitter's line-number gutter, rendered as it arrived (#4020)
- the truncation notice naming the key that reveals the rest, and its
  agreement with the export/Observatory fold (`the_deck_and_the_export_show_the_same_number_of_result_lines`)

SPEC 6.4 keeps every one of them. The v2 renderer was deleted rather than left
unwired; this is where it comes back, built on the highlighter P2 provides.

## Approach

P2 in `design/tui-v2/IMPLEMENTATION-PLAN.md` builds `stella-tui-highlight`:
a syntect theme from SPEC 3.1, highlight-on-arrival cached as
`Vec<Line<'static>>`, never per frame. The result row should then be a v2 arm in
`v2_rows` (`crates/stella-tui/src/render/entry.rs`) that reuses that cache.

Note `stella_transcript::syntax::{body_paint, paint_line}` already exists and is
shared with the export and Observatory renderers — the deck keeps only the
palette (`syntax::tok_style`). Do not fork it; the whole point of #4036 was that
three copies of one rendering decision drift.

## Verify

Every test in `crates/stella-tui/src/render/tests/tool_output.rs` and
`render/tests/inline_diff.rs` must stay green, or be changed only where SPEC 6.4
genuinely differs — and each such change named in the PR description.

## Definition of done

The result row carries its call's rail metal; a Rust diff renders with add and
remove rows per SPEC 6.4; a counter test proves highlighting runs once per event,
not once per frame; `tool_class` can then be deleted if #4125 resolves that way.

Refs #4123, #4125

macanderson:
## What

SPEC 6.2 makes the rail a property of the **event**, and a call and its result
are one event — the deck records them as two entries only because the head has
to draw before the result exists. #4123 routed only the head through v2, so the
block drew a v2 rail over a v1 result:

```
 │ ● edit src/lib.rs                          ← Rgb(239, 197, 63)  gold
 │ ⎿ applied                            42ms  ← Rgb(169, 170, 181) silver
 │   line two                                 ← silver
```

The rail changed metal one row into the block it exists to hold together. That
is measured, not inferred — the two colours above are what the renderer actually
produced before this change.

## Why it was not fixed by writing a v2 result renderer

Because that was tried in #4123 and turned 16 tests red for the right reason.
The result row carries four things a fresh renderer had not been taught:

- syntax highlighting in the file's own language (#4019, #4036)
- word-level inline diffs (`resolve_inline_diff`, `INLINE_DIFF_CAP`)
- the emitter's line-number gutter, rendered as it arrived (#4020)
- a truncation notice naming the reveal key, agreeing with the export and
  Observatory fold (#3644)

So this PR restyles the renderer that already has all four rather than writing a
second one. The tool block moves whole into `crates/stella-tui/src/render/entry/tool.rs`
and reads its metal from `v2::metal_for` — the same table the head reads.

**The evidence that nothing was dropped is that every test in
`render/tests/tool_output.rs` and `render/tests/inline_diff.rs` stayed green,
untouched.** Those are the files the issue named as the guard.

## Exemplar

The split follows `render/entry/recall.rs`, the sibling this crate already uses
for "one entry kind whose machinery is a concern of its own".

## Witnesses

Each was confirmed to fail on the old behaviour by reverting the production
change and keeping the test (`git checkout` afterwards, not `git stash`).

| Test | Fails on old code because |
|---|---|
| `a_mutations_whole_block_wears_one_metal` | the result rail was a fixed muted tone under a gold head |
| `a_reads_whole_block_wears_the_reading_metal` | same, and it pins that the two metals differ |
| `a_rust_diff_renders_add_and_remove_rows_per_spec_64` | the add/remove ground stopped where the code stopped |
| `expanding_a_call_row_shows_its_arguments_again` | `ctrl+o` on a call row rendered nothing at all |
| `an_unchanged_tail_is_highlighted_once_not_once_per_frame` | ten frames lexed ten times what one frame lexed |

Four more are regression guards and pass both ways, which is what they are for:
the failure metal, a new entry still reaching the frame, the streaming preview
re-folding under a cached tail, and a streaming answer never freezing.

## Three things that came with it

**`ctrl+o` on a call row shows its arguments again.** #4123's router returns
`true` for every `ToolStart`, so the v1 arm that pretty-printed the argument
object became unreachable, and `head_rows` takes no `expanded` flag. The
expansion had been drawing nothing since that PR. Fixed here rather than filed —
same function.

**Highlighting runs once per event.** Settled entries always did; the live tail
did not — `SessionFold::refresh` cleared it and re-folded the last entry every
frame. The tail's rows are now kept when the entry is one the model cannot
rewrite in place. That predicate is exact rather than a fingerprint: `Text` and
`Reasoning` are the only kinds reachable from the three `last_mut()` sites in
`model.rs`, so every other kind — `ToolResult` above all — is immutable once
pushed. `syntax::lex_count` is thread-local, which matters: as a process-global
atomic it read 40 for ten frames against 56 for one, because the rest of the
suite was rendering on other threads. A measurement that moves with the
scheduler is worse than none.

**`SessionFold` moved to `views/session/fold.rs`** so that cache did not grow a
god file — and the split took `views/session.rs` from 1571 lines to 1378, under
the 1500 limit. It therefore leaves `scripts/file-size-baseline.txt`, AGENTS.md's
per-crate table and the crate README, as `god-files` requires. The baseline entry
was removed by hand rather than by `make file-size-update`, deliberately: a full
regeneration retightens every unrelated ceiling, which is the shared-cell shape
that has reddened `main` before.

## Deliberate deviations, stated

**SPEC 6.4 says `Line.style` carries the diff tint.** The deck cannot do that:
`push_diff_line` prepends the block's rail, and a line-level style would paint
it — the rail is the call's metal, not the hunk's. A padded trailing span gives
the same band with the margin excluded, and the test asserts both halves (the
band reaches the pane edge; the rail's `bg` stays `None`). Filed as #4195 so the
next surface to implement 6.4 does not fall in.

**The P2 plan orders a syntect-based `stella-tui-highlight` crate.** Not built.
It would be a fourth lexer over bodies the deck, the export grid and the
Observatory already share through `stella-transcript::syntax` — the drift #3644
and #4036 each closed once, and the issue body says "do not fork it". P2's real
deliverable here was the cache. Filed as #4196 to reconcile the plan.

## Deleted tests

None. One **changed contract**, named as the issue asks: in
`render/tests/palette.rs`, `transcript_prefix_colors_stay_in_the_brand_family`
asserted that a successful result's rail *recedes* to a fixed muted tone under
whatever metal its head wore. That assertion described the defect. It now reads
the call's own metal, with the old reasoning replaced rather than deleted.

## Not done, tracked

- `tool_class::classify` now has **zero** production callers — the v1 call row
  was its only one. `dead_code` cannot see it (`pub` in a `pub mod`), so the
  gate is green either way. Reported on #4125, which owns that decision; nothing
  changed here.
- The v2 head still renders a path in one flat tone — the dim-directory /
  bright-basename split went with `path_spans`. Already tracked as #4168.

## Verification

`make gate CARGO_SCOPE="-p stella-tui"` green (880 tests), and
`cargo test --workspace` green.

Closes #4127

## Summary by Sourcery

Unify tool calls and results under one event rail while preserving result features and avoiding repeated rendering work.

New Features:
- Render tool calls and their results as one event with a consistent, tool-specific rail.
- Restore expanded tool-call arguments and preserve syntax highlighting, inline diffs, gutters, and truncation behavior in tool results.
- Cache immutable transcript-tail rendering so syntax highlighting runs once per event instead of once per frame.

Bug Fixes:
- Prevent tool blocks from switching rail colors between the call and result rows.
- Restore ctrl+o argument rendering for tool-call rows.
- Keep diff tinting confined to diff content while extending the diff background across the pane width.

Enhancements:
- Extract tool-entry rendering into a dedicated module and move SessionFold into its own session view module.
- Centralize tool rail-metal selection and add regression coverage for tool block rendering and tail caching.

Documentation:
- Update stella-tui file-size guidance to reflect the SessionFold module split.

Tests:
- Add coverage for consistent tool-block metals, failure rendering, expanded arguments, diff bands, and cached tail highlighting.

---

## Merged `main` (2026-08-21) — what the merge had to reconcile

`main` moved from the branch point (`274d3421`) to `48d80b03` while this sat
open, landing six `stella-tui` PRs. Four of them touch this code, and the merge
was not marker cleanup — three of the reconciliations below would have shipped a
silent regression if the conflict had been resolved by picking a side.

**1. One metal derivation, not two.** #4182 gave `kind_for` a `measured`
argument and added `head_metal(name)`. This branch had added `metal_for(name)`
for the same job. Both computed the same colour from the same table under two
names — the exact duplicate-derivation defect *both* doc comments were written
to prevent, and the one that made the block draw gold for its head and silver
for every row under it. They are collapsed into `metal_for`, on main's two-arg
`kind_for(name, None)`.

**2. The `EntryView` + `measured_delta` backfill (#4154) is threaded through the
moved code.** `v2_rows` resolves the measurement and hands it to
`tool::start_rows`, which passes it to `head_rows` — so the projection stays a
pure function of what is known about one call. Witnessed by main's own
`views/session/fold_backfill.rs`, which survives the move and passes.

**3. A fix `main` landed that this branch would have silently reverted.** #4155
changed the result row's metric column to gate on the *measurement* rather than
on the diff text, removing an `unwrap_or((0, 0))` that fabricated `+0 −0` over
real edits. This branch had already **moved that code** to
`render/entry/tool.rs`, so the fix landed in a file this branch no longer uses.
Git reported no conflict — the two versions were at different paths — and the
merge would have quietly restored the pre-fix expression. Ported by hand, and
the decision points of `result_rows` were compared against main's arm one by one
to confirm nothing else diverged. **This is the mechanism worth naming in
review: a file move plus a fix to the old path is invisible to git.**

**4. A rival fix for the same defect.** #4169 fixed `ctrl+o` on a call row
independently of this PR's fix for it. Its `argument_rows` is superseded by
`tool::start_rows`; its delegating `ToolStart` arm auto-merged in *below* this
branch's stub arm as unreachable dead code, compiling clean and passing every
gate. Deleted, with #4157's lesson folded into the surviving comment.

**5. Smaller carries.** `RememberedDiff::text` is `Option<String>` now, and
main's "bind the margin once" convention won for the body helpers —
`push_diff_line` takes `&margin` plus this branch's `width` for SPEC 6.4's
padded diff ground.

### Comments the merge itself made stale, fixed here

Three, because a stale comment is a bug in review and the merge is what staled
them: `push_detail_line`'s doc still said a successful result "recedes to muted"
(the precise defect this PR removes); `Extent`'s doc still said the measured
backfill did not exist (it landed in #4182); and a sentence-boundary typo in
`recall.rs` from main's rewrite of the same paragraph.

### The file-size baseline

Left as the branch has it. `views/session.rs` composes to 1384 lines **with**
main's additions included, under the 1500 limit, so dropping its grandfather
entry is correct — and CI's `file size ratchet` check passes on the merge
commit. Deliberately no `make file-size-update`: a full regeneration retightens
every unrelated ceiling, which is the shared-cell shape that has reddened `main`
before.

### Deleted tests

Still none. `check-deleted-tests` compares against the base, and nothing in the
merge removes a `#[test]`. Main's `fold_backfill` module and its test survive
the `SessionFold` move and are declared from `views/session.rs`.

### One argument-count allow, stated

Absorbing `measured` took `start_rows` to eight arguments, tripping
`clippy::too_many_arguments`. It carries an `#[allow]` with the reason the lint
is wrong *here* — the same reason `result_rows` beside it already carries one:
these are the entry's own fields one to one, and a struct would be a second
shape to keep in step with the entry rather than a simplification.

### Filed, not fixed

- **#4209** — the result row's metric column has no *render* witness for a
  measured-but-patchless change. #4155's render half is asserted nowhere; only
  the model layer (`delta_at`) and the head row are covered. That gap is exactly
  what let reconciliation 3 above go undetected by the test suite, so it is
  worth closing on its own.

### Verification of the merge

`make gate CARGO_SCOPE="-p stella-tui"` green — all 32 steps, clippy and rustdoc
included. `cargo test -p stella-tui`: 969 passed, 0 failed.
`cargo check --workspace --all-targets`: clean.


macanderson:
Heads up: this PR is CONFLICTING against main (since #4182 merged), which means GitHub gives it no merge ref and **runs no workflows on it at all** — the absence of a red check here is not a green signal.

The collision is a design one, not textual: this PR extracts the tool-call block into a new mod tool; and deletes the inline ToolResult arm, while #4182 rewrote that same arm around EntryView (files + following) plus the head's measured size column. 18 hunks across 7 files.

Details and definition of done in #4202. Leaving the resolution to you rather than guessing at intent.

macanderson:
Superseded by #4210. This branch was cut before thirteen commits landed on `main`, several in the same code — it conflicted, and a conflicting PR has no merge ref, so GitHub started no `pull_request` workflows for it at all (its only checks were Vercel, which fails on every PR here, and a skipped CLA). #4210 is the same work rebuilt on the current main, and smaller: #4169 had independently restored `ctrl+o` on a call row and `head_metal` already existed, so both are dropped there.

Issue #4199, opened 2026-08-21
stella-store: store.db has corrupted pages and nothing detects it — some queries fail, others silently read a subset


macanderson:
## What was observed

`.stella/private/store.db` in this repository's own workspace (945 MB) fails
integrity checking:

```
$ sqlite3 "file:.stella/private/store.db?mode=ro" "pragma quick_check;"
*** in database main ***
Tree 45 page 230966: btreeInitPage() returns error code 11
Tree 45 page 230965: btreeInitPage() returns error code 11
Tree 45 page 230957: btreeInitPage() returns error code 11
Tree 46 page 230916: btreeInitPage() returns error code 11
```

The practical symptom is that any query whose plan reaches those pages returns
`Error: stepping, database disk image is malformed (11)`. Narrow queries bound
by an `execution_id` range still work, which is what makes this easy to miss —
the store looks healthy until a scan happens to touch the damaged trees.
Execution 178's `events` rows are among the unreadable ones.

This was found while censusing tool-call telemetry for #4163; it is not caused
by that work and predates it.

## Why it matters

`store.db` is the canonical local telemetry record — `stella stats`,
`stella observe`, `execution_rollup`, `finalize_execution_reflection`, and the
`usage.db` hub replication all read it. Three separate concerns:

1. **Silent partial reads.** A corrupted page does not make the store refuse to
   open; it makes *some* queries fail and others quietly succeed over a subset.
   Anything that aggregates could report a number that is wrong rather than
   absent, which is the failure mode this repository cares most about.
2. **Replication.** `stella usage sync` walks a per-project cursor. What
   happens when the cursor's next range is unreadable — does it fail, skip, or
   wedge? Unknown; worth establishing.
3. **Nothing notices.** There is no integrity check anywhere in the store's
   open path or in `stella stats`, so a user finds out when a query errors.

## Not established

- **The cause.** Unknown. Candidates worth distinguishing rather than guessing
  between: an unclean shutdown with WAL involvement, a `stella observe` or
  session process killed mid-write, filesystem-level damage, or a genuine bug
  in a write path. Do not assume one — the store has several writers
  (`record_files_touched`, the event journal, `Store::prune`'s 13-table cascade).
- **Whether it is reproducible**, or a one-off on this machine.
- **How much data is lost.** `quick_check` names four pages; a full
  `integrity_check` would say more, and `.recover` would say how much survives.

## Suggested work

1. Run `pragma integrity_check` (not just `quick_check`) and record the full
[...]

Issue #4197, opened 2026-08-21
Eight model providers have unmeasured support for parallel tool calls, and one test run each would settle it


macanderson:
## What shipped, and what it left open

#4163 added the sixth provider-parity axis (`ParallelToolCallPosture`,
`crates/stella-model/src/provider_parity/parallel.rs`). It records two facts
per provider, deliberately separated because they have different kinds of
evidence:

- **`fan_in_witness`** — does *this tree's* adapter turn N tool calls in one
  assistant message into N `ToolCall`s? Settled for all five dialects by
  wiremock tests; every row names one, gap rows included.
- **`ParallelAdmission`** — does the *provider* emit several tool calls per
  assistant message without being asked? This can only be settled by observing
  a real run.

Measured so far:

| id | admission |
|---|---|
| `openrouter` | `DefaultOn` — 891 calls / 648 tool-bearing messages, 92 with 2+, max 9 (kimi-k3, executions 140–177) |
| `zai` | `DefaultOn` — 496 / 382, 79 with 2+, max 5 (glm-5.2) |
| `anthropic`, `openai`, `gemini`, `vertex`, `bedrock`, `xai`, `deepseek`, `local` | `Undetermined` |

The eight are `Undetermined` rather than assumed-to-match on purpose: "every
OpenAI-compatible server defaults `parallel_tool_calls` to true" is a plausible
sentence that would make eight rows green with no observation behind them,
which is the claim architecture rule 8 exists to refuse.

## Why it matters

The engine dispatches consecutive read-only calls concurrently and the system
prompt asks the model to use that (`tool_steering!`,
`crates/stella-cli/src/agent/prompt.rs`, #2985). A provider that silently never
batches makes `bash` the only surface through which a model can ask about
several files at once — which is #4151, where a model routed 24 of 26 reads
through `sed`.

## How to settle one

Run any tool-using turn on the provider, then census this repository's own
store. **Group `tool_start` on the preceding `step_manifest`** (one
`step_manifest` is one model call, so one group is one assistant message).
Do *not* census `tool_calls.ts`: it has one-second granularity and gave the
wrong answer on execution 178, which is how #4163 came to be filed on the
premise that kimi-k3 emitted no parallel calls at all.

```sql
with marked as (
  select execution_id, seq, event_type,
         sum(case when event_type='step_manifest' then 1 else 0 end)
           over (partition by execution_id order by seq) as g
  from events where event_type in ('step_manifest','tool_start')
[...]

Issue #4175, opened 2026-08-21
deck: only one tool call per (path, turn) can show a diff — the turn-boundary producer is the ceiling


macanderson:
## Problem

The Command Deck's transcript can show a diff for **at most one tool call per
(path, turn)**. When several mutating calls touch the same file inside one
turn, only the last row renders a diff and a `+N −M`; the earlier rows degrade
to naming their change.

That is a deliberate, honest choice made in the fix for #4155 — but it is a
ceiling imposed by the *producer*, not by the renderer, and it is worth
lifting.

## Why the ceiling exists

`stella-pipeline`, which emitted one `AgentEvent::FileChange` per adopted
change *during* delivery, is deleted (#3865). The only producer left is
`stella_cli::turn_files::emit_shared_tree_changes`, which snapshots the
worktree and emits **one aggregate `FileChange` per path at the turn
boundary** — after every `ToolResult` of the turn has already folded.

So one turn yields exactly one remembered diff per path
(`FileState::recent_diffs`), and the model has one change to hand out among
however many calls edited that file. `SessionModel::apply`'s `ToolResult` arm
stamps every such call with the same `InlineDiffRef { path, seq }`, and
`supersede_inline_diff` then clears all but the last — because the alternative
is each row rendering the turn's *entire* change to that file as its own,
which is the misattribution `render::resolve_inline_diff` exists to prevent.

`FileState` is already built for the better world: it keeps the last
`DIFF_HISTORY` (8) diffs per path, keyed by seq, precisely so several rows on
one path can each resolve their own change. Today only one seq per turn is ever
populated, so that depth goes unused within a turn.

## What "done" looks like

A successful mutating tool call emits (or has emitted on its behalf) a
`FileChange` describing **that call's own change**, before or as its
`ToolResult` folds, so:

1. Two `edit_file` calls to one path in one turn produce two remembered diffs
   at two seqs, and both transcript rows render their own diff and `+N −M`.
2. `supersede_inline_diff` (`crates/stella-tui/src/model.rs`) becomes
   unnecessary and is deleted along with its tests — or is kept only for the
   aggregate fallback path, explicitly.
3. The turn-boundary aggregate emit still happens (the Files tab and
   `turn_facts` depend on it) without double-counting `FileState::added` /
   `removed`, which today sum every mutating `FileChange` for the path.

Point 3 is the real design question and the reason this was not folded into
the #4155 fix: per-call and per-turn measurement currently feed the same
[...]

Issue #4155, opened 2026-08-21
deck: a successful edit_file can render no measured +N −M on either row — the result row's metric is gated on an inline diff that may not resolve


macanderson:
## Problem

For at least some successful `edit_file` calls, the deck shows **no measured `+N −M` anywhere** — not on the head (that row never had one; #4150/#4154) and not on the result row either, because the result row's metric is gated on an inline diff that did not resolve.

Reported from a live session editing `crates/stella-tools/Cargo.toml`.

## Evidence, and which links are inferred

Observed (screenshot of the deck's interactive transcript): the result row (the `└` elbow) is **blank on its left**, and directly beneath it the tool's own output renders as a body line:

```text
 │ ● edit crates/stella-tools/Cargo.toml
 └
   replaced 1 occurrence(s) in crates/stella-tools/Cargo.toml at byte 1774 (file sha256/8 5bd38ada)
```

Inferred from `crates/stella-tui/src/render/entry.rs` (the `TranscriptEntry::ToolResult` arm, collapsed branch):

1. `let shown: Vec<&str> = if inline.is_some() { Vec::new() } else { …preview… };` — body lines come only from `shown`.
2. A body line rendered, so `shown` was non-empty, so **`inline` was `None`**.
3. The metric column is built under `if inline.is_some() { … }`, so with `inline == None` the row carries **no `+N −M`** at all — only the duration.

I have not reproduced this under a debugger; steps 1–3 are read from the source, and step 2 is the inference. What is directly observed is that no delta appears on either row.

## Where it can go wrong

`inline = diff.as_ref().and_then(|d| resolve_inline_diff(d, files))`, so `None` means one of:

- **`TranscriptEntry::ToolResult::diff` is `None`.** Set in `crates/stella-tui/src/model.rs` (~line 721) only when `ok && self.mutated_path_for(call_id).is_some()`; `mutated_path_for` requires `summarize::is_file_mutation(&name)` **and** a non-`None` path resolved from the `ToolStart`.
- **The `InlineDiffRef` did not resolve.** `resolve_inline_diff` → `FileState::diff_at(seq)` → `remembered_at(seq)`, which searches `recent_diffs`. `FileState::remember_diff` (`crates/stella-tui/src/model/file_state.rs:105`) **returns early when the `FileChange` carried `diff: None`**:

  ```rust
  let Some(text) = diff else { return };
  ```

[...]

PR #4200, opened 2026-08-21
feat(stella-model): declare the parallel-tool-call axis, and answer the question it was filed on

### LINKED ISSUE #4163: stella-model: parallel tool calls are asked for by the prompt, relied on by the engine, and declared on no provider axis
## The gap

The engine runs consecutive read-only tool calls concurrently, and the system prompt asks the model to use that: *"Send independent tool calls together in one response"* (`crates/stella-cli/src/agent/prompt.rs`, the `tool_steering!` macro, #2985).

**Nothing tells the provider to allow it.** `parallel_tool_calls` — the OpenAI-compatible request field that governs whether a model may emit more than one tool call per assistant message — appears nowhere in the tree:

```
rg 'parallel_tool_calls|parallelToolCalls' crates/    # no matches
```

And it is declared on no parity axis. `crates/stella-model/src/provider_parity.rs` declares five: `CachePosture`, `ReasoningPosture`, `OverflowPosture`, `OutputBudgetPosture`, `StreamFallbackPosture`. Whether a provider honours multiple tool calls per turn is not among them, so no provider states a posture and no witness test pins one.

## Evidence that it matters

`.stella/private/store.db`, execution 178 (`openrouter`, `moonshotai/kimi-k3`):

```sql
select count(distinct ts) as distinct_ts, count(*) as total_calls
from tool_calls where execution_id=178;
-- 69 | 71
```

71 tool calls, 69 distinct timestamps. Parallel calls execute concurrently and share a timestamp, so at most two calls in that whole execution could have been issued together — and both collisions pair a real tool with a *coordination* call (`task_create`, `task_start`), not two reads:

```sql
select ts, count(*) c, group_concat(name) from tool_calls
where execution_id=178 group by ts having c>1;
-- 2026-08-21 20:09:36 | 2 | bash,task_create
-- 2026-08-21 20:12:42 | 2 | read_file,task_start
```

So that model issued 69 sequential round trips and never once batched two reads, despite the prompt asking it to. The consequence is #4151: `bash` became the only surface through which it could ask about several files at once, and it routed 24 of 26 reads through `sed`.

## What is NOT established

I did not verify what OpenRouter or kimi-k3 do by default. OpenAI's own API defaults `parallel_tool_calls` to true, so its absence from the request is **not** by itself proof that parallel calls are disabled — the model may simply not volunteer them. Distinguishing "we never asked" from "it declined" needs a live call, which I did not spend. That distinction is the first thing this issue should settle, because it decides whether the fix is a request field or a documented provider limitation.

## Why this is invariant #8's shape

Invariant #8 exists because "providers diverge in sneaky ways" and the fix is to *declare* the divergence with a named witness test rather than assume parity. This is exactly that: a capability the prompt and the engine both depend on, that varies by provider, that nothing declares and nothing tests. The precedent is the cache axis — OpenRouter ran Claude models with zero prompt caching for months because nothing enforced it.

## Definition of done

- Determine, per provider id, whether multiple tool calls per assistant message are supported and whether the request must opt in. Record how it was determined.
- Add a sixth axis to `provider_parity.rs` (`ParallelToolCallPosture` or similar) with a posture per provider id and, for anything other than a declared gap, a named witness test — enforced from both sides the way the existing five are (`stella-cli` config tests fail on a missing row; `stella-model` parity tests fail when a named witness disappears).
- Where the wire needs the field, send it in the shared chat-completions adapter, with a wiremock test asserting the request body.
- A turn-level measurement afterwards: re-run a comparable task and report distinct-timestamp-vs-call-count, the same way it was measured here.

## Related

- #4151 — the file tools gained in-schema batch forms precisely so batching does not depend on this capability. That fix is deliberately independent: a working surface should not rest on a provider feature nobody declared. Closing this issue does not make that one redundant, and vice versa.

macanderson:
The parallel-tool-call parity axis (#4163), and three adjacent gaps in the
matrix's own enforcement.

**Scope note:** this PR originally also carried a fix for #4175. That half has
been reverted out (1944c7d61) because main moved under it — see the bottom
section. What remains conflicts with nothing and stands on its own.

## The gap

The engine dispatches consecutive read-only tool calls concurrently and the
system prompt asks the model to use that ("send independent tool calls together
in one response", `tool_steering!`, #2985). Whether a provider will *emit*
several tool calls in one assistant message was declared on no parity axis and
pinned by no test — invariant #8's exact shape, with both halves of the working
surface already depending on it.

## The issue's premise did not survive checking, and settling it decided the fix

#4163 was filed on a census of `tool_calls.ts`, which has one-second
granularity. It read execution 178 as "essentially zero parallel calls" and was
explicit that this did not distinguish "we never asked" from "it declined",
naming that as the first job.

Grouping `tool_start` events by the preceding `step_manifest` instead — one
`step_manifest` is one model call, so one group is one assistant message — over
this repository's own `store.db`:

| route | tool-bearing messages | tool calls | with 2+ | most in one |
|---|---|---|---|---|
| `openrouter` / `moonshotai/kimi-k3` (executions 140–177) | 648 | 891 | 92 | 9 |
| `zai` / `glm-5.2` (all executions) | 382 | 496 | 79 | 5 |

Both emit parallel calls in volume, and `parallel_tool_calls` has never been
sent — it appears nowhere in the tree.

**So no request field is added.** Adding one would change the wire bytes of
every shared-adapter request and risk a 400 on the local servers that reject
unknown keys, to buy a capability the measurement says is already there.
Execution 178 is an outlier within its own route, not a provider limitation.

Two honest limits on that census, both stated in the module doc: `store.db` has
genuine page corruption (`quick_check` reports `btreeInitPage` errors on trees
45 and 46 — filed as #4199), which is why the `openrouter` row cites an
execution range rather than the whole table; and it is evidence about routes
actually run here, which is why the other eight ids are not assumed to match.

## The axis records two facts, not one

They have different kinds of evidence, and merging them would let a row claim
more than it can show:

- **`ParallelAdmission`** — a fact about a vendor's live API, settleable only by
  observing a run. `DefaultOn` for the two ids above, each citing what was
  counted; `Undetermined` for the other eight rather than inferred from a
  sibling dialect. #4197 tracks settling them, one run each.
- **`fan_in_witness`** — a fact about this tree's adapters, settled by wiremock.
  **Every** row names one, gap rows included: an unknown admission must never
  mean "we do not know what we would do with a parallel call". The five new
  tests each stream several calls with their argument fragments *interleaved*,
  so an adapter keyed on arrival order corrupts them rather than merely
  reordering them, and each includes two calls to the same tool — the shape a
  parallel read batch takes and the one a dedupe-by-name fan-in would collapse.

All five dialects already fan in correctly. These pin it.

## Three gaps found on the way, same invariant, fixed here

- The **stream-fallback axis was enforced from neither side**, despite AGENTS.md
  describing every axis as enforced from both: no witness-existence test in
  `stella-model`, no completeness test in `stella-cli`. A `UnaryFallback` row
  could name a test that had been renamed away and nothing would notice.
- **`all_axes_cover_the_same_provider_ids` checked three of the five tables.**
  It now checks all six and names which one drifted.
- The module doc said **"Three axes"** while listing four.

## Structural notes

`provider_parity.rs` becomes a directory module (`zai.rs` + `zai/` is the
crate's own idiom) rather than growing toward the ceiling, and
`adapter_sources()` moves to module level so the axis submodule checks its own
witnesses against the same list. Two witness files sit beside their `tests.rs`
rather than inside it, and the Responses witness lives in `openai/provider.rs`:
those parents are grandfathered god files, two of them exactly on their ceiling.

## Verification

`make gate` green (exit 0) on this branch. `stella-model` 419 → 431 tests
(measured by stashing, not asserted).

## Why the #4175 half is no longer here

Since this branch's merge base, main landed **#4176** (75394cd5) — a *different*
fix for the same #4155 defect. `model.rs` now stamps `changes + 1` and adds
`supersede_inline_diff`, and **three** merged consumers depend on that seq
contract, not one: `render/entry.rs`, `v2::transcript_source` (#4182), and
`views::session`'s fold_backfill (#4154).

The per-call producer cannot land against it as written, and the reason is a gap
in that design rather than a merge conflict: a per-call `FileChange` folds
*before* its `ToolResult` (row wants `seq = changes`), a turn-boundary one folds
*after* (row wants `changes + 1`), and the producer deliberately leaves the
boundary live for `delegate`, read-only tools that mutated, concurrent groups
and external edits. Both orderings end up live in one turn and one expression
cannot serve both.

Measured rather than reasoned, with both test sets kept and only that expression
varied: `changes` fails 9, `changes + 1` fails 1 — and 4 of the 9 are #4182/#4154
tests describing live behaviour on exactly the paths the producer leaves on the
boundary, so they must not be updated away.

That work is preserved on `wip-4175-per-call-filechange` and the decision it
needs is filed as **#4205**. Credit to the session rebasing this PR for catching
it before it merged.

Closes #4163
Refs #4175, #4197, #4199, #4205


macanderson:
This PR is CONFLICTING against main, so GitHub gives it no merge ref and runs no workflows on it — the absence of a red check is not a green signal.

Flagging before anyone resolves it mechanically: the only conflicted hunk in model.rs is a **comment**, but the contradictory **code** auto-merges clean and the result is wrong. This branch stamps seq = changes (no +1) because it moves measurement into the call itself (call_measure, #4175); main stamps seq = changes + 1 plus supersede_inline_diff for the turn-boundary producer. A naive merge keeps main's +1 under this branch's comment saying +1 is wrong — and with the per-call producer that points the seq one change into the future.

Which producer wins is a design call for you. Full detail and definition of done in #4203.

macanderson:
## This PR and `main`'s #4176 are two incompatible fixes for the same defect

Attempted to rebase this branch for merge and stopped: the conflict is not textual. Since this branch's merge base (`274d3421`), `main` landed **#4176** (`75394cd5`, "a mutating row shows the change its own call made") — a *different* fix for #4155/#4175, using the turn-boundary producer this PR replaces.

Two corrections to this PR's description, which the merge base made true and `main` has since falsified:

- **`supersede_inline_diff` now exists.** The description says it "does not exist — it and the test cluster the Verify section names were never merged". #4176 merged both: `crates/stella-tui/src/model.rs:1368` defines it, `:750` calls it.
- **`main` now stamps `seq` as `changes + 1`.** This branch stamps `changes`. Same expression, `crates/stella-tui/src/model.rs`, in the `ToolResult` arm.

### The contradiction, tested both ways

Merged `origin/main` into this head, kept **both** test sets (no test deleted), and ran `cargo test -p stella-tui --lib` under each choice of that one expression:

| `seq` expression | Result |
|---|---|
| `changes` (this PR) | **9 failures** — 887 passed |
| `changes + 1` (main) | **1 failure** — 895 passed |

Dropping `+1` fails, from `main`:

```
model::tests::an_edit_result_resolves_the_change_its_own_call_made
model::tests::a_later_turns_edit_never_renders_an_earlier_turns_diff
model::tests::only_the_last_call_to_a_path_in_a_turn_claims_the_turns_change
model::tests::calls_to_different_paths_in_one_turn_each_keep_their_own_ref
model::tests::a_measured_change_with_no_patch_still_reports_its_delta
v2::transcript_source::tests::a_returned_edit_head_states_the_delta_the_emitter_measured
v2::transcript_source::tests::each_head_states_its_own_calls_change_not_a_neighbours
views::session::fold_backfill::a_settled_head_states_its_size_once_the_turn_boundary_measures_it
views::session::tests::a_later_mutation_leaves_the_settled_inline_diff_showing_its_own_change
```

Note the last four are **not** #4176's. They come from #4182/#4154 (the v2 head measured-delta backfill), which resolve through the same `InlineDiffRef` seq. So this is not a two-PR collision — the `seq` contract has three merged consumers.

Keeping `+1` fails this PR's own witness, and not cosmetically:

```
model::tests::two_calls_to_one_path_each_resolve_the_change_they_made
  assertion `left == right` failed: the first row resolves the first edit
  left:  Some("@@\n+second")
  right: Some("@@\n+first")
```

That is the first row rendering the second call's diff — real misattribution, exactly what `render::resolve_inline_diff` exists to prevent.

### Why no choice of that expression works

Both producers are live in this PR's own design. Its `call_measure.rs` module doc says so:

> The turn-boundary reading still happens and still writes both projections — it now reports whatever the per-call readings did not already claim

and names what stays on the boundary path: `Tool::parallel_safe` tools (`delegate`, the one mutating one), read-only tools, and concurrent dispatch groups — plus anything mutating the tree outside a call.

So after this PR the composed system has **two orderings**:

- per-call: `FileChange` lands *before* the `ToolResult` → `changes` already bumped → correct seq is `changes`
- turn-boundary: `FileChange` lands *after* every `ToolResult` has folded → correct seq is `changes + 1`

A single constant offset cannot serve both, and the row cannot currently tell which producer will measure it. That is the design gap, and it is this PR's to close — `main`'s tests describe live behaviour on the paths this PR deliberately leaves on the boundary producer, so they are not stale and should not be updated away.

### What "done" looks like

The stamp needs to know which producer will measure the call it belongs to, rather than assuming one. The information exists at the stamp site — the registry already decides per call whether it ran alone (that decision is what gates `measure_and_publish`), so it can be carried to the fold rather than re-derived. Whatever the mechanism, the bar is: **all 9 tests above and this PR's witness pass together**, with no test from #4176, #4182 or #4154 deleted or weakened. `scripts/check-deleted-tests.sh` runs on `pull_request` and will name any that go missing.

Everything else in this PR — the parallel-tool-call parity axis, the five fan-in wiremock witnesses, the stream-fallback axis enforcement, the `all_axes_cover_the_same_provider_ids` gap, the "Three axes"/four-axis module doc — is independent of this and conflicts with nothing. If the `seq` reconciliation is going to take a while, splitting the #4163 half out would land that value now.

*(Rebase was aborted; nothing was pushed to this branch. The other conflicts in the merge — `stella-cli`, `stella-model`, `command_deck.rs`, `provider_parity.rs`, `zai.rs` — all auto-resolved cleanly.)*

macanderson:
**This PR's head changed — anything resolved against the old head is stale.**

Acting on the rebase evidence in the comment above: the #4175 half is reverted
out (1944c7d61) and **#4200 is now the #4163 parity axis alone**.

- Touches `stella-model`, `crates/stella-cli/src/config/tests.rs`, and
  `AGENTS.md` (invariant #8's axis list). **No `stella-tui` files at all**, so
  the `model.rs` conflict against main's #4176 is gone.
- **No force-push.** The revert is an ordinary commit on top, so a plain fetch
  gets the new head. It restores the tree to the #4163 commit byte-for-byte
  (`git diff --stat 1319f358e HEAD` is empty).
- `make gate` re-run on the reverted branch: exit 0.

The finding was correct and I verified it independently before acting: #4176
stamps `changes + 1` and adds `supersede_inline_diff`, and three merged
consumers now depend on that seq — `render/entry.rs`, `v2::transcript_source`
(#4182), `views::session`'s fold_backfill (#4154).

The blocking problem is a gap in the per-call design rather than a merge
conflict. A per-call `FileChange` folds *before* its `ToolResult` (row wants
`seq = changes`); a turn-boundary one folds *after* (row wants `changes + 1`).
The producer deliberately leaves the boundary live for `delegate`, read-only
tools that mutated, concurrent dispatch groups and external edits — so both
orderings are live in one turn and no single expression serves both. My
producer combined with main's fold breaks the per-call rows outright.

The producer half is parked on `wip-4175-per-call-filechange` (nothing lost),
and the decision it needs is filed as **#4205**, which records the measurement,
weighs three options rather than declaring one, and notes that
`supersede_inline_diff` must survive for the boundary paths — #4175's text
asking to delete it predates #4176.

macanderson:
Superseded by #4215, which carries both of this PR's commits onto current
`main` with the #4203 conflict resolved.

The `stella-model` parallel-tool-call parity axis cherry-picked clean. The
`stella-tui` half needed a decision, and #4215 takes **this PR's producer** —
per-call measurement (#4175) — and **neither PR's consumer**.

The reason is #4205: the per-call producer is deliberately partial, so both
orderings are live in one turn and `changes` unadjusted is right for only half
of it. It is silently wrong when the per-call reading produced nothing, which
is reachable — a `write_file` of bytes identical to what is on disk succeeds
and moves nothing, and `snapshot_worktree` is best-effort and infallible by
signature, so an unreadable tree measures nothing. The counter is then the
*previous* call's change and the row renders an older edit as its own.

So the fold stops predicting the counter and claims a change it was observed to
produce. That deletes `supersede_inline_diff` (one change, one row, by
construction) and makes this PR's comment about `+1` true in a stronger form
than it stated.

`a_successful_call_that_changed_nothing_claims_no_diff` in #4215 is the witness
that fails on this branch's consumer. Everything else here is preserved.

Closing in favour of #4215.

Issue #4154, opened 2026-08-21
tui-v2: the head row's measured +a -b is still unfilled — Extent is expressible but nothing ever measures it


macanderson:
## Problem

`#4150` is closed by the change that made the v2 head's metric *expressible as absent* (`Extent { added: Option<u32>, removed: Option<u32> }`) and stopped it rendering a fabricated `+0 -0`. That removed the false claim. It did **not** deliver the metric SPEC 6.2 asks for:

```text
 │ ● edit …/self_driving_cmd.rs +3 -1              ⚡2ms · → task 3
```

Today the head renders `● edit …/self_driving_cmd.rs` with no size column at all, because it is drawn at dispatch and nothing revisits it.

This is the second half of #4150's suggested fix (1): *"fill them when the turn-boundary `FileChange` for that path/seq arrives — the same source v1 reads. This requires the head row to be re-rendered after the fact, which the current draw-once-at-dispatch shape does not do."*

## Files

- `crates/stella-tui/src/v2/transcript.rs` — `Extent`, `EventKind::{Read,Edit,Write,Delete}`, `kind_detail`. `Extent::delta`/`added`/`removed` are the measured constructors and already exist; nothing in production calls them.
- `crates/stella-tui/src/v2/transcript_source.rs` — `kind_for` builds every file kind with `Extent::default()` (unmeasured), and `head_rows` takes only `(name, path, input, width)`.
- `crates/stella-tui/src/render/entry.rs` — `v2_rows` routes `TranscriptEntry::ToolStart` into `v2::head_rows`; it currently receives no `files` ledger and no view of the paired `ToolResult`.
- `crates/stella-tui/src/render.rs` — `resolve_inline_delta(dref, files)` is the correct source: it reads `FileState::delta_at(seq)`, which is the emitter's measurement, not a recount of rendered text.
- `crates/stella-tui/src/views/session.rs` — the settled-prefix fold that calls `entry_lines`. Note its cache key already includes `file_gen` (the sum of per-path `changes`), so a mutation invalidates the cache and every row re-renders; a backfill would take effect without new invalidation machinery.

## Constraints discovered

- **Do not derive the counts from the tool's input or from the result text.** That is exactly what #2290 established as the defect, and `AgentEvent::FileChange`'s doc in `crates/stella-protocol/src/event.rs` says so explicitly. `edit_file` with `replace_all` makes the input-derived number wrong anyway.
[...]

Issue #2290, opened 2026-08-08
deck: every lane's turn wedges at forwarder.await — the registry-held EventSender keeps the channel open, so a finished turn never returns and the session freezes until /clear


macanderson:
## The report

Screenshot (2026-08-08, 2:49 AM): a pipeline deck turn painted `✓ lead · done · stage complete · 100%`, composer showed `1 queued`, and no prompt would ever run again. `/clear` was the only way to keep using the session.

## Root cause

Every deck lane's turn tail is:

```rust
registry.attach_events(stella_core::EventSender::new(tx.clone()));  // turn start
...
drop(tx);
let persistence_complete = forwarder.await.unwrap_or(false);        // turn end
```

`EventSender` (`crates/stella-core/src/event_sender.rs`) captures a strong `UnboundedSender` clone in an `Arc` closure. The registry outlives the turn, so after `drop(tx)` the channel still has a live sender, `spawn_forwarder`'s `while let Some(event) = rx.recv().await` loop never ends, and `forwarder.await` never completes. The turn future never returns; the driver stays in its mid-turn `select!` (`crates/stella-cli/src/command_deck.rs` ~1648) forever:

- every submitted prompt routes `NextTurn` → `queue.push_back` — "queued but deliberately NOT drained: the idle arm pops it as the lead's next turn" — and that idle arm is only reached when the turn future returns;
- `/clear` is the only input that breaks the select (`break TurnEnd::Cleared`), which is exactly why it is the only rescue;
- `record_execution_end` sits after the await, so the execution row is never closed out (`.stella/private/store.db` shows the frozen session's execution with NULL `finished_at`, empty outcome, cost 0.0 while the deck painted $0.0087).

Three sites share the tail:

1. `crates/stella-cli/src/command_deck.rs` `run_lead_turn` (attach ~4186, tail ~4245)
2. `crates/stella-cli/src/command_deck.rs` `run_lead_pipeline_turn` (attach ~4348, tail ~4453)
3. `crates/stella-cli/src/subsession.rs` `run_worker` (attach ~735, tail ~766) — worker lanes also never report `WorkerEnd`, so their executions never close either.

## History

- #960 was this exact deadlock on one-shot `stella run`; its fix is `close_event_stream` (`crates/stella-cli/src/agent/persistence.rs`): detach → drop → await, with docs saying the detach is "the counterpart every turn owes attach_events".
[...]

Issue #4203, opened 2026-08-21
PR #4200 conflicts with main and runs no CI — it and main carry contradictory fixes for the #4155 diff misattribution (+1 vs per-call producer)


macanderson:
PR #4200 (`worktree-fix-4163-parallel-tool-calls`) is **CONFLICTING** against
`main`. A conflicting PR has no merge ref, so **GitHub runs no workflows on it
at all** — there is no red check to notice.

Do not resolve this one mechanically. The textual conflict is misleading: the
only conflicted hunk in `crates/stella-tui/src/model.rs` is a **comment**, while
the *contradictory code* auto-merges clean — and the result is wrong.

## Two incompatible fixes for the same bug (#4155)

`SessionModel::apply`'s `ToolResult` arm stamps the `InlineDiffRef::seq` that a
mutating row's diff resolves against.

**`main` (@ `7149784d`)** — keeps the turn-boundary producer and compensates in
the consumer:

```rust
let seq = self.files.iter().find(|f| f.path == path).map_or(0, |f| f.changes) + 1;
```

plus a `supersede_inline_diff(dref)` call, so only the last row of a path claims
the single aggregate change.

**PR #4200** — changes the **producer** instead: it adds per-call measurement
(`crates/stella-tools/src/call_measure.rs`, #4175) so the `FileChange` is emitted
the moment a solo mutating call returns and folds *before* the `ToolResult`.
`changes` is therefore already bumped, and the stamp is:

```rust
let seq = self.files.iter().find(|f| f.path == path).map_or(0, |f| f.changes);   // no +1
```

#4200's own comment states this explicitly: "`+1` is still not the repair, and
never was: it would make all of a path's calls in one turn point at a single
change, which is the misattribution `render::resolve_inline_diff` exists to
prevent."

## Why the naive merge is a defect

`git merge origin/main` keeps `main`'s `+ 1` **and** #4200's comment saying `+1`
is wrong. Combined with #4200's per-call producer — which already bumped
`changes` — the seq then points one change into the **future**, so a mutating row
resolves against a change that has not happened yet. It compiles, and the comment
directly contradicts the code above it.

## Reproduce

```sh
git fetch origin +refs/pull/4200/head:pr4200
git merge-tree --write-tree origin/main pr4200      # model.rs + model/tests.rs
git show pr4200:crates/stella-tui/src/model.rs   | rg -A10 'let diff = if ok'  # no +1
[...]

PR #4211, opened 2026-08-21
fix(stella-tui): every mutating tool row renders the change that call made

### LINKED ISSUE #4175: deck: only one tool call per (path, turn) can show a diff — the turn-boundary producer is the ceiling

### LINKED ISSUE #4205: stella-tui: the InlineDiffRef seq assumes one producer, but a per-call producer makes both orderings live in one turn
## The gap

`TranscriptEntry::ToolResult` carries an `InlineDiffRef { path, seq }` stamped
at fold time, and the seq it must stamp depends on **which producer will
measure that call** — a fact the fold does not have.

- A **per-call** `FileChange` folds *before* its `ToolResult`, so `touch_file`
  has already bumped `changes` and `remember_diff` recorded at the new value.
  The row must stamp `seq = changes`.
- A **turn-boundary** `FileChange` folds *after* every `ToolResult` of the
  turn, so `changes` is still the pre-turn value when the row stamps. The row
  must stamp `seq = changes + 1`.

Main stamps `changes + 1` (#4176, `crates/stella-tui/src/model.rs`) and adds
`supersede_inline_diff` so that exactly one row per `(path, turn)` claims the
single aggregate change. That is correct for the world it shipped into, where
the turn boundary is the only producer.

## Why this is now blocking

An attempt at #4175 (branch `wip-4175-per-call-filechange`, reverted out of
#4200 in 1944c7d61) adds a per-call producer: `ToolRegistry::execute` measures
the work tree when a solo mutating call returns and publishes through a
host-supplied `CallMeasure` port, before the engine sends that call's
`ToolResult`.

It does **not** make the per-call producer universal, and deliberately so. A
tree snapshot attributes everything since the last reading to the call that
just finished, which is only honest when the call ran alone. The turn-boundary
producer therefore stays live for:

- `Tool::parallel_safe` tools — `delegate` is the only mutating one, and it is
  dispatched in a concurrent group with its siblings, so no before/after
  reading can say which sibling wrote what;
- read-only tools that mutated anyway (the schema is a claim, not a guarantee);
- anything a concurrent dispatch group changed;
- external edits (a human saving a file in another window mid-turn).

So after that change **both orderings are live within one turn**, and no single
stamped expression can serve both.

## Evidence

With main merged in, both test sets kept (nothing deleted), and only the one
expression in `model.rs`'s `ToolResult` arm varied — `cargo test -p stella-tui
--lib`:

| stamp | result |
|---|---|
| `changes` | **9 failures**, 887 passed |
| `changes + 1` | **1 failure**, 895 passed |

The single failure under `changes + 1` is the per-call witness
(`two_calls_to_one_path_each_resolve_the_change_they_made`), failing with
`left = Some("@@\n+second")`, `right = Some("@@\n+first")` — genuine
misattribution, not a stale expectation.

**The count is only meaningful against one specific resolution**, so it is
stated exactly: take the per-call branch's `seq` expression, keep main's
`supersede_inline_diff` *and its call site*, and concatenate both test sets with
nothing deleted. A different resolution can produce a different count.

**4 of the 9 are not #4176's tests at all** — this is the load-bearing half of
the argument, because it proves the contract has consumers beyond the PR pair:

| test | file |
|---|---|
| `a_returned_edit_head_states_the_delta_the_emitter_measured` | `crates/stella-tui/src/v2/transcript_source.rs` (#4182) |
| `each_head_states_its_own_calls_change_not_a_neighbours` | `crates/stella-tui/src/v2/transcript_source.rs` (#4182) |
| `a_settled_head_states_its_size_once_the_turn_boundary_measures_it` | `crates/stella-tui/src/views/session/fold_backfill.rs` (#4154) |
| `a_later_mutation_leaves_the_settled_inline_diff_showing_its_own_change` | `crates/stella-tui/src/views/session.rs` (#4154) |

They describe live behaviour on exactly the paths a per-call producer
deliberately leaves on the boundary. **They must not be updated away.**

(The measurement was made by a separate session while rebasing #4200, on a merge
it then aborted — so nothing in the tree reproduces it. It is recorded here
rather than re-derived; re-running it is two `cargo test` calls and is worth
doing before acting.)

## The shape of a fix

The registry already decides, per call, whether that call ran alone — it is
exactly the predicate gating `measure_and_publish`
(`ToolRegistry::measures_alone`, reading `read_only` and `parallel_safe` off
each tool's own schema). That fact needs to reach the fold instead of either
ordering being assumed.

Options worth weighing rather than one being obvious:

1. **Carry it on `AgentEvent::ToolResult`** — a `measured_per_call: bool`, or
   better a field naming *which* producer measured. Wire change: invariant #4
   (serde round-trip test) and invariant #10 (declare the consumer in
   `crates/stella-protocol/src/event/consumers.rs`).
2. **Correlate by call, not by seq** — give `AgentEvent::FileChange` an
   optional `call_id`, and have the fold resolve a row against the change that
   names it. This removes the seq guess entirely rather than parameterising it,
   and would let `supersede_inline_diff` go for per-call rows while staying for
   boundary ones. Larger wire change; probably the right end state.
3. **Make the fold defer** — stamp the path only and resolve at render time
   against whatever `recent_diffs` holds. It is the only option that does not
   put the producer's identity on the wire, which is its whole appeal.

   Two costs, and the second is the one that probably kills it. It reintroduces
   the ambiguity `resolve_inline_diff` exists to prevent when a path has several
   changes in a turn. And **at draw time it cannot distinguish "not measured
   yet" from "measured as zero"** — a per-call row rendered before its boundary
   measurement looks identical to one whose change really was empty. #4182 has
   already settled that distinction the other way: `head_rows`' `measured` is an
   `Option<(u32, u32)>` whose doc says it is `None` "while it is still in
   flight" and that "a head row is never asked to guess", and `kind_for` maps
   `None` onto `Extent::default()` so the extent "stays honestly absent"
   (#4180). Deferring would force exactly the guess that doc forbids.

Whichever is chosen, `supersede_inline_diff` must survive for the boundary
paths. #4175's original text asks for it to be deleted; that was written before
#4176 landed and before the per-call producer's exclusions were worked out.

## Definition of done

Two `edit_file` calls to one path in one turn each render their own diff and
their own `+N −M`, **and** all three existing seq consumers keep passing with
no test updated away — including the `delegate` / concurrent / external-edit
paths, which still resolve through the boundary aggregate.

## Files

- `crates/stella-tui/src/model.rs` — the `ToolResult` fold, the stamp,
  `supersede_inline_diff`
- `crates/stella-tui/src/model/file_state.rs` — `remember_diff`, `diff_at`,
  `delta_at`
- `crates/stella-tui/src/render.rs`, `render/entry.rs` — `resolve_inline_diff`,
  `resolve_inline_delta`
- `crates/stella-tui/src/v2/transcript_source.rs` (#4182),
  `crates/stella-tui/src/views/session.rs` (#4154) — the other two consumers
- `crates/stella-tools/src/call_measure.rs`,
  `crates/stella-tools/src/registry.rs` — the per-call producer and
  `measures_alone`, on `wip-4175-per-call-filechange`
- `crates/stella-cli/src/turn_files.rs` — `open_turn_streams`, the boundary emit

## Starting point

`wip-4175-per-call-filechange` holds the producer half, which is independent of
this question and passes `make gate` on its own merge base. It needs main
merged in and this decision made before it can land.

Refs #4175, #4176, #4182, #4154, #4200

macanderson:
Lifts the one-diff-per-`(path, turn)` ceiling (#4175) and settles the seq
contract the attempt at it exposed (#4205).

**Credit where it's due:** the seq mechanism here is not mine. Two sessions
working the merge queue caught that my first attempt at #4175 could not land —
one measured the regression, the other designed the fix below. What I added is
the producer half, a tightening, a file-size split, and a full gate run.

## The producer (#4175)

`ToolRegistry::execute` measures the work tree when a solo mutating call
returns and publishes through a host-supplied `CallMeasure` port, on the same
channel the engine then sends `ToolResult` on — so a call's change folds
*before* the row that made it.

The registry is the seam because it is the only place that sees every
dispatched call whatever the tool. `turn_files`' standing argument against
having the four file tools emit still decides it: a tool's arguments are a
request rather than a measurement (#2290 rendered bulk edits as `+0 −0`), and
`bash`, MCP servers and custom tools mutate the tree naming no path in any
schema — a hook on the file tools would report a subset while looking
exhaustive.

**Counts cannot double.** `WorkJournal::snapshot_worktree` *consumes* what it
reports, so consecutive readings partition the turn. Both granularities call
the same function, so there is one measurement taken more or less often, not
two that have to agree.

**What is deliberately not measured per call.** A snapshot attributes
everything since the last reading to the call that just finished, which is only
honest when the call ran alone. Read-only tools and `parallel_safe` tools
(`delegate` is the only mutating one, dispatched in a group with its siblings)
are excluded — read from each tool's own schema, so a new tool inherits the
right answer. The boundary sweep still reports what no single call can own.

## The seq contract (#4205)

That partial producer is exactly what made this hard, and it is the part my
first attempt got wrong. Two producers now fold in **opposite orders** around
the `ToolResult` they belong to, so the row cannot assume either:

| producer | folds | row needs |
|---|---|---|
| per-call | before the result | `seq = changes` |
| turn boundary | after every result | `seq = changes + 1` |

Main's #4176 stamps `changes + 1`; the producer needs `changes`. Neither serves
both, and **three** merged consumers depend on the answer — `render/entry.rs`,
`v2::transcript_source` (#4182), and `views::session`'s fold_backfill (#4154).
Measured with both test sets kept and only that expression varied: `changes`
fails 9, `changes + 1` fails 1.

**The fix puts nothing on the wire.** The consumer derives which producer
measured a call from state it already holds: `mutation_baselines` records the
path's `changes` at `ToolStart` and takes it at `ToolResult`. A count that
moved means the per-call producer already published; one that did not means the
boundary still owes it. Invariants #4 and #10 never engage.

`supersede_inline_diff` survives, with its justification narrowed to the
boundary-only paths — it matches on full `(path, seq)` equality, so two
measured calls hold distinct seqs and neither supersedes the other. #4175's
original text asks for it to be deleted; that predates #4176.

### One deliberate tightening, and an honest note on it

`landed` tests `recorded == base + 1`, not `> base`. The producer emits one
`FileChange` per changed path and mutating calls dispatch alone, so a change
that landed for this call moved the count by exactly one.

This is **the exact form, not a bug fix**, and the comment says so. I first
justified it with a stranded-baseline scenario — a turn cancelled between
`ToolStart` and `ToolResult`, then a provider recycling the `call_id` (this
repo's telemetry shows kimi reusing `read_file:0`..`:3`) — and wrote a witness
for it. The witness passed under *both* predicates, so the scenario is not
reachable: the re-dispatched call's own `ToolStart` overwrites the stale entry
before anything reads it, and the one path that folds a result with no start
(the halted arm in `driver::dispatch`) answers `Err` and never stamps. I
deleted that test rather than ship one that proves nothing.

## Verification

- `make gate` **green (exit 0)**.
- `cargo test -p stella-tui --lib`: **897 passed**, nothing deleted and no test
  from either rival fix updated away.
- `a_row_resolves_its_own_change_under_either_producer` folds the same events in
  both orderings and discriminates in both directions.
- The producer files are byte-identical to the pre-revert original (`git diff
  c4eb759ec HEAD` empty for them) — verified rather than inferred from a green
  build, because a `git revert` on a merged branch silently drops a later merge
  of the reverted content. This branch was built by re-applying the tree onto
  `main` rather than merging the wip branch, so that hazard never arms.

The new cluster took `model/tests.rs` to 1661 lines, over the ungrandfathered
1500 limit — it lands as `model/tests/producer_seq.rs`, a split rather than a
baseline entry.

Closes #4175
Closes #4205
Refs #4155, #4160, #4176, #4182, #4154, #4200

## Summary by Sourcery

Render each successful mutating tool row with the specific work-tree change it produced while maintaining correct attribution for turn-level changes.

New Features:
- Attribute work-tree changes to individual successful, solo mutating tool calls so each tool row can render the change it made.

Bug Fixes:
- Resolve inline diffs correctly whether changes are published before a tool result or at the turn boundary, preventing missing or misattributed diffs.

Enhancements:
- Centralize per-turn event and call-measurement stream setup and safely clear both when a turn ends.
- Use tool metadata to exclude read-only and parallel-safe calls from per-call attribution while preserving boundary-based reporting for changes they cannot own.

Tests:
- Add coverage for per-call measurement eligibility, stream cleanup, multiple mutations to one path, failed calls, and both producer ordering scenarios.

Chores:
- Split the TUI model test suite to remain within the project’s file-size limit.


macanderson:
Two PRs are now open for this work — this one and #4215 (I was resolving #4203
and did not see this branch until after pushing). Same producer, different
consumer. That is #4203's exact failure mode recurring, so rather than argue
preference: here is one measured difference, and what each design buys.

**I have not re-run your gate or disputed any of it.** The producer half here
and in #4215 is the same code.

## The measured difference

Checked out `refs/pull/4211/head` at `8ab7a0fad`, appended one test, ran
`cargo test -p stella-tui --lib`:

```rust
// c1: a real edit, measured per call.        changes -> 1, diff@1
ToolStart{c1, edit_file, src/a.rs}
FileChange{src/a.rs, +1, diff: "+c1 wrote this"}
ToolResult{c1, Ok}

// c2: write_file of bytes identical to what is on disk.
// It SUCCEEDS and moves nothing, so nothing is measured for it.
ToolStart{c2, write_file, src/a.rs}          // baseline[c2] = 1
ToolResult{c2, Ok("wrote 42 bytes")}         // recorded = 1, landed = false, seq = 2

// The boundary sweeps up a change to that path that c2 did not make.
FileChange{src/a.rs, +9 -9, diff: "+SOMEONE ELSE wrote this"}   // recorded at 2
```

```
PROBE: c2 dref   = Some(InlineDiffRef { path: "src/a.rs", seq: 2 })
PROBE: c2 renders = Some("@@\n+SOMEONE ELSE wrote this")
```

`c2`'s row renders a change `c2` did not make. `landed` is `false` here for the
honest reason — nothing landed — but the `else` arm stamps `recorded + 1`
regardless, which is a claim on **whatever measures that path next**, whoever
made it.

### Why the premise is reachable, not contrived

- `write.rs` has no identical-content short-circuit: it writes the bytes and
  returns `ok("wrote N bytes to {path}")`. Same content in, same tree out.
- `SessionDurability::snapshot_worktree` is
  `self.journal().and_then(|j| j.snapshot_worktree().ok()).unwrap_or_default()`
  — **any** journal error, and a session with no journal at all, measures
  nothing. Every mutating call in such a session takes the `else` arm.
- The other half only needs *something else* to touch that path in the same
  turn: a `delegate` sibling's write, a concurrent group, a human saving in
  another window. Those are the exact paths this PR's own scope note says stay
  on the boundary.

## What each design buys

Not a verdict — the two really do trade against each other:

**This PR (derive which producer answered).** Preserves the boundary fallback,
so a row whose own measurement came back empty still shows what the boundary
later found. No consumer test changes; `supersede_inline_diff` survives with a
narrowed justification. Lower-risk diff. The cost is the case above: it cannot
tell "the boundary picked up *my* work" from "this is someone else's edit to a
path I also touched," and it renders both.

**#4215 (claim an observed change).** A `ClaimWindow` records mutations that
fold while a call is in flight; a result claims the oldest for its own path,
and claiming removes it — so one change/one row is a property of the structure
and `supersede_inline_diff` is deleted. A call under which nothing folded gets
no reference and renders nothing. The cost is the mirror image: it gives up the
fallback, so it loses information in the case where the boundary really was
picking up that call's work, and one `fold_backfill` test's subject had to be
rewritten (three of the four consumers #4205 named keep every assertion; only
their fixture ordering moves).

The judgement is whether a row that cannot know should show a plausible diff or
nothing. I argued nothing — a wrong prediction renders a real, correctly
formatted diff under the wrong call, which a reader cannot detect, while a
missing observation renders a blank they can. But that is a call for whoever
owns this surface, not for whichever PR merges first.

## Either way

If this PR is the one that lands, the case above is worth a guard or an
explicit "known and accepted" note rather than leaving it unstated —
`a_successful_call_that_changed_nothing_claims_no_diff` in #4215 is the test,
and it is portable. Happy to close #4215 and port its two extra witnesses plus
the #4213/#4214 follow-ups onto this branch instead; say which and I will do
the merging rather than leave two rival PRs open.

Issue #4214, opened 2026-08-21
deck: a call that mutates several paths can only render one of them — InlineDiffRef and ToolResult::diff are singular


macanderson:
## The gap

One tool call can mutate several paths. Per-call measurement (#4175) reports
all of them — `emit_measured_tree_changes` emits one `AgentEvent::FileChange`
per path in the snapshot — but a result row can claim exactly one:

```rust
// crates/stella-tui/src/model.rs
pub struct TranscriptEntry { /* ... */ ToolResult { diff: Option<InlineDiffRef>, .. } }
```

`InlineDiffRef` names a single `(path, seq)`, and `mutated_path_for` resolves a
single path off the call's `ToolStart` input. The other measured changes stay
in the claim window until the next dispatch group clears it, reach the Files
tab, and never reach the row.

Reachable today with `apply_edits` (a batch write across several files):
`summarize::tool_input_path` takes `edits[0].path` and its own comment says the
first edit's path "stands in so a single-file batch still renders an inline
diff under its result row" — an accurate description of a stand-in, not of an
answer. A multi-file batch renders one of its files' diffs and silently drops
the rest.

It becomes the common case once #4213 lets `bash` rows claim: a codemod is
precisely the call that rewrites twenty files at once.

## Why this was not worth fixing before

Under the turn-boundary producer there was one aggregate change per path and no
way to know which call produced which, so a row showing several diffs would
have been several misattributions instead of one. The per-call producer removes
that objection: the changes that fold under a call are that call's, all of
them.

## The shape of a fix

`TranscriptEntry::ToolResult::diff` becomes a collection rather than an
`Option`, and the fold claims *every* unclaimed change for the call rather than
the first matching one. `ClaimWindow::claim`
(`crates/stella-tui/src/model/inline_diff.rs`) already removes what it hands
out, so claiming N is the same one-change-one-row property as claiming one.

The render half is the larger part and should be costed before committing to
it: `render::resolve_inline_diff` / `resolve_inline_delta`,
`v2::transcript_source::measured_delta` and `head_rows`' single
`Option<(u32, u32)>` extent all assume one change per row, and a row rendering
twenty diffs inline needs a reading judgement (a fold? a count plus the first?)
that is not obvious and is not this issue's to make.

Deliberately not attempted in #4203: that change was already reconciling two
conflicting designs, and widening the entry shape at the same time would have
made the reconciliation unreviewable.

## Definition of done

A single `apply_edits` across two paths renders both changes on its result row,
[...]

Issue #4213, opened 2026-08-21
deck: a bash/MCP call's tree change is now measured per call but no transcript row can claim it — is_file_mutation is still a four-name list


macanderson:
## The gap

Per-call work-tree measurement (#4175) measures **every** solo mutating call —
`ToolRegistry::measures_alone` reads `read_only` and `parallel_safe` off each
tool's own schema, so `bash`, every MCP tool and every custom script tool is
measured the moment it returns, exactly like `edit_file` is.

The transcript row cannot show any of it. `SessionModel::mutated_path_for`
gates the inline-diff claim on `model::summarize::is_file_mutation`, a
hard-coded four-name list:

```rust
// crates/stella-tui/src/model/summarize.rs
pub(super) fn is_file_mutation(name: &str) -> bool {
    matches!(name, "write_file" | "edit_file" | "apply_edits" | "delete_file")
}
```

So a `bash` call running `sed -i`, a `cargo fmt`, or an MCP server's
`apply_patch` now produces a real per-call `FileChange` that the Files tab
counts and **no transcript row ever claims**. The change is measured, carried
on the wire, folded into the ledger, and then dropped on the floor by the one
surface a reader consults to find out what a call did.

This is not a regression — those rows never showed a diff — but the reason they
could not has just gone away. Before #4203 the only producer was the turn
boundary, which had one aggregate per path to hand out, so extending the name
list would have made a `bash` row claim work it did not do. With a per-call
producer the change under a `bash` call really is that call's.

## Two predicates that must agree, in two crates, with no link

`ToolRegistry::measures_alone` (`crates/stella-tools/src/registry.rs`) decides
*which calls get measured*, from the schema. `is_file_mutation`
(`crates/stella-tui/src/model/summarize.rs`) decides *which rows may claim*,
from a name list. Nothing checks them against each other.

Today the divergence fails safe in both directions — a measured call not in the
list leaves its change unclaimed (this issue), and a listed call that was not
measured claims nothing (`ClaimWindow` has nothing to give it). That is by
construction, not by luck, and it should stay that way whatever is done here.

## Why the fix is not just "widen the list"

`TranscriptEntry::ToolStart` carries `path: Option<String>`, filled by
`summarize::tool_input_path` from a `path` key or `edits[0].path`. A `bash`
call has neither, so a widened name list would give `mutated_path_for` nothing
to look up. The row needs the path from the *measurement* rather than from the
call's arguments — which is available, since the `FileChange` names it.

[...]

PR #4215, opened 2026-08-21
fix(stella-tui): a mutating row claims the change it was measured to make

### LINKED ISSUE #4175: deck: only one tool call per (path, turn) can show a diff — the turn-boundary producer is the ceiling

### LINKED ISSUE #4203: PR #4200 conflicts with main and runs no CI — it and main carry contradictory fixes for the #4155 diff misattribution (+1 vs per-call producer)

### LINKED ISSUE #4205: stella-tui: the InlineDiffRef seq assumes one producer, but a per-call producer makes both orderings live in one turn

macanderson:
## What & why

`main` and PR #4200 shipped **contradictory** fixes for the same bug, and the
textual conflict between them was a comment while the contradictory code merged
clean (#4203). This resolves it by keeping one producer and **neither**
consumer.

### The producer: per-call measurement wins (#4200's half)

The turn-boundary sweep cannot answer *"what did this call change"*. It emits
one aggregate per path after every result of the turn has folded, so the
information is destroyed before any consumer sees it — the most a consumer can
do is hand the whole turn's change to the last row that touched the path.
Per-call measurement (#4175) takes the *same* reading — the work journal's
shadow-git snapshot, measured with git's own `--numstat` — the moment a solo
mutating call returns, and publishes on the channel that call's `ToolResult`
then rides. Counts cannot double because a snapshot consumes what it reports,
so consecutive readings partition the turn rather than overlapping it.

### The consumer: observe, do not predict (neither side's half)

This is the part #4203 asked someone to decide and #4205 independently proved
was undecidable as posed. Both existing consumers computed the seq from
`FileState::changes`, and the per-call producer is deliberately **partial** —
a call that ran alone and succeeded — so **both orderings are live in one
turn** and each expression is right for only half of it:

| stamp | right for | wrong when |
|---|---|---|
| `changes + 1` | the turn-boundary sweep | every call touching a path in one turn predicts the *same* seq, so `supersede_inline_diff` had to blank all but the last row — which then showed the whole turn's change as its own |
| `changes` | per-call measurement | the reading produced nothing, and the counter names the **previous** call's change — so the row renders an older edit as its own work |

The second is not hypothetical. A `write_file` of bytes identical to what is on
disk succeeds and moves nothing, and `snapshot_worktree` is best-effort and
infallible by signature, so an unreadable tree simply measures nothing.

So the fold **stops predicting**. `model::inline_diff::ClaimWindow` records the
mutations that fold while a call is in flight, and a result **claims** the
oldest unclaimed one for its own path. Claiming removes it, so one change is
shown by exactly one row *by construction* rather than by a repair pass —
`supersede_inline_diff` is deleted. A call under which no change folded gets no
reference at all and its row degrades to naming its change, the degradation
`DIFF_HISTORY` aging and `MAX_TRACKED_FILES` eviction already have.

The trade is deliberate: **a prediction that is wrong renders a plausible
lie** — a real diff, correctly formatted, under the wrong call. A missing
observation renders nothing, and a reader can see nothing.

This is #4205's **option 2** (correlate by call, not by seq) *without* its wire
change. The correlation is derived from an ordering the engine already
guarantees structurally — a tool that is neither `read_only` nor
`parallel_safe` is dispatched in a group of one
(`driver::dispatch::execute_tool_calls`), so the changes folding between a
call's `ToolStart` and its `ToolResult` are that call's. The oldest-first claim
rule holds even if that stops being true, because a per-call reading is taken
when a call finishes and results fold in the order calls finish. Option 2's
`call_id` on `AgentEvent::FileChange` would buy the same correlation at the
cost of a wire change, a `consumers.rs` entry and ~40 construction sites, and
would degrade to the same silence when a producer passed `None`.

A change measured while **no call was in flight** is the sweep reporting what
the per-call readings did not claim — a `delegate`'s writes, a human editing in
another window. It reaches the file ledger and no transcript row, because the
row names a tool call and that call did not make it.

Closes #4203
Closes #4205
Closes #4175
Refs #4155, #4176, #4182, #4154, #4200

## The witness

Three, each discriminating a different rival. Measured by varying **only** the
seq expression in the `ToolResult` arm, everything else held:

| consumer | result |
|---|---|
| `main`'s `changes + 1` | **16 failed**, 885 passed |
| #4200's `changes` | **4 failed**, 905 passed |
| this PR | **909 passed**, 0 failed |

- `two_calls_to_one_path_each_resolve_the_change_they_made` — fails on the
  turn-boundary design. This is #4175's ceiling.
- `a_successful_call_that_changed_nothing_claims_no_diff` — fails on **both**
  rivals: `+1` stamps a reference resolving to nothing, bare `changes` resolves
  to the previous call's edit.
- `a_change_measured_between_calls_is_claimed_by_no_row` — fails on `+1`.

## Deleted / renamed tests

`check-deleted-tests` compares two trees; naming them here is the required half.

**#4205 asked that four existing seq consumers not be "updated away."** Three
keep **every assertion** — only their fixture *ordering* moves, because the
product now emits the change before the row:

- `a_returned_edit_head_states_the_delta_the_emitter_measured` (v2)
- `each_head_states_its_own_calls_change_not_a_neighbours` (v2)
- `a_later_mutation_leaves_the_settled_inline_diff_showing_its_own_change`

The fourth is renamed, and this is the one place I am overriding #4205's
stated definition of done, so it is argued rather than asserted.
`fold_backfill::a_settled_head_states_its_size_once_the_turn_boundary_measures_it`
witnessed a head *filling in*: settling silent, then picking up its size from
the boundary. That sequence cannot occur any more in either direction — the
per-call measurement precedes the row, and a boundary change belongs to no
call. #4205 describes the fallback as live behaviour on the `delegate` /
concurrent / external-edit paths, but those rows cannot claim a diff at all:
`is_file_mutation` is a four-name list and `delegate` is not on it. The only
row that *could* fall back is one of the four file tools whose own measurement
failed — and at that point the fold cannot tell "the boundary picked up my
work" from "this is someone else's edit to a path I also touched." Preserving
the fill-in means accepting the second to gain the first. It is replaced by
three tests: `a_head_states_its_size_in_the_frame_it_settles`,
`a_settled_head_does_not_adopt_a_change_measured_after_it`, and
`a_settled_head_drops_its_diff_once_the_history_ages_it_out` — the last
re-witnessing the `file_gen` cache term on the case that survives, so the
module keeps being evidence for an argument that is true.

Two more from `model/tests.rs`, both deliberate, both replaced:

- `only_the_last_call_to_a_path_in_a_turn_claims_the_turns_change` asserted the
  design that lost — one aggregate change per `(path, turn)`, claimed by the
  last row. Replaced by `two_calls_to_one_path_each_resolve_the_change_they_made`.
- `a_turn_that_measured_no_change_leaves_the_row_silent` asserted that such a
  call still *stamps* a reference which then dangles. That is precisely the
  behaviour this PR changes: the call claims nothing, so there is no reference
  to dangle. Replaced by `a_successful_call_that_changed_nothing_claims_no_diff`,
  which covers the same situation and additionally pins that the row does not
  pick up a neighbour's change — the assertion that discriminates all three
  designs.

## The gate

- [x] `cargo fmt --check`
- [x] Docs updated where behavior changed — `resolve_inline_diff`,
      `measured_delta`, `kind_for`, `remember_diff` all described *stamping*;
      they describe claiming now
- [x] `Closes #N` in this description **and** as commit trailers

`model.rs` and `model/tests.rs` both crossed the 1500-line ceiling and neither
is grandfathered, so this splits rather than baselines. The claim rule moves to
`model/inline_diff.rs` as plain logic over owned data with no reference to the
rest of the model — which is what lets it be stated and unit-tested without a
transcript or a file ledger. **`model.rs` ends at 1402 lines, below the 1429 it
started at**, and no baseline entry is added.

## Nothing left behind

- [x] Filed: #4213, #4214

- **#4213** — `bash`/MCP/custom-tool mutations are now measured per call but no
  row can claim them, because `is_file_mutation` is still a four-name list. Not
  a regression (those rows never showed a diff), but the reason they could not
  has just gone away.
- **#4214** — a call that mutates several paths can only render one:
  `InlineDiffRef` and `ToolResult::diff` are singular. Reachable today via
  `apply_edits`; becomes common once #4213 lands.

Also hit but already tracked: **#4060** — running the suite with `STELLA_HOME`
set (the documented way to isolate a `stella-cli` test run from the real
`~/.stella`) fails `untrusted_project_cannot_enable_tools_or_replace_an_agent_prompt`,
because that test sandboxes via `HOME` alone. Unrelated to this change; the
test passes without the override.

## Ground-rule check

- [x] No new cross-boundary types — deliberately **no wire change**, which is
      the main thing to push back on if you disagree (see #4205 option 2)

**#4200 merged while this was in progress — but with its `stella-tui` half
reverted out (`1944c7d61`), so only the parallel-tool-call parity axis landed.**
The #4175 work #4203 was filed about is still not on `main`; this PR is it.
`main` is merged in here, and the diff against it is exactly the tui/tools
change (the duplicated parity commit is absorbed by the merge).

The judgement worth challenging is the one in "Deleted / renamed tests": I
chose silence over a plausible-looking fallback on the one path where the two
are distinguishable. If you would rather a row show the boundary aggregate when
its own measurement came back empty, that is a real position — it just cannot
be had without also crediting a human's concurrent edit to a tool call, because
the fold cannot tell them apart.

## Summary by Sourcery

Attribute inline file changes to the tool calls that actually produced them by measuring successful solo mutations per call and claiming only changes observed during each call.

New Features:
- Measure successful solo mutating tool calls individually and publish their work-tree changes before the corresponding result is folded.
- Correlate inline diffs with the changes observed while each call is in flight, allowing multiple calls to the same path to display their own changes.

Bug Fixes:
- Prevent mutating rows from displaying a previous call's diff, a whole-turn aggregate, or a change made outside the call.
- Avoid assigning inline diffs to successful calls that changed nothing or to failed calls.

Enhancements:
- Centralize per-turn event and work-tree measurement setup across engine and command-deck turn owners.
- Replace sequence prediction and transcript supersession with oldest-first, single-use change claiming.
- Keep unattributed boundary changes in the file ledger without associating them with transcript rows.
- Move inline-diff attribution logic into an independently testable model module.

Documentation:
- Update inline-diff and measured-delta documentation to describe observation and claiming rather than sequence stamping.

Tests:
- Add registry, claim-window, model, rendering, and stream-setup tests covering per-call measurement, attribution boundaries, no-op and failed calls, stale history, and turn ownership.

Chores:
- Remove obsolete turn-boundary attribution tests and the inline-diff supersession pass.


macanderson:
**#4211 is an open PR doing the same work with a different consumer.** It was pushed while this was in progress; neither session saw the other.

Same producer (per-call `CallMeasure`, byte-identical). The consumers differ:

- **#4211** keeps both arithmetic answers and *derives* which producer measured the call, from a `mutation_baselines` map recorded at `ToolStart`. Preserves the boundary fallback, changes no consumer test, keeps `supersede_inline_diff`.
- **this PR** stops predicting entirely: a `ClaimWindow` hands each observed change to at most one row, and a call under which nothing folded renders nothing. Deletes `supersede_inline_diff`.

Measured difference, run against `refs/pull/4211/head` at `8ab7a0fad` — full repro in https://github.com/macanderson/stella/pull/4211#issuecomment-5376504741:

```
PROBE: c2 dref    = Some(InlineDiffRef { path: "src/a.rs", seq: 2 })
PROBE: c2 renders = Some("@@\n+SOMEONE ELSE wrote this")
```

A `write_file` of bytes identical to what is on disk succeeds and moves nothing, so nothing is measured for it; #4211's `else` arm then stamps `recorded + 1`, which claims whatever measures that path next. `a_successful_call_that_changed_nothing_claims_no_diff` here is the witness for it.

The trade is real in both directions and is stated on #4211: that design keeps the fallback and loses nothing when the boundary really was picking up the call's own work; this one refuses the fallback and renders a blank instead of a plausible diff under the wrong call.

**Two rival PRs is #4203's own failure mode, so this should not stay a race.** If #4211 is preferred, I will close this and port its two extra witnesses onto that branch — the #4213/#4214 follow-ups stand either way.

macanderson:
## #4211 landed while this was open — roughly half of this PR is now already on `main`

Merging PRs and stopped here rather than rebasing, because what is left of this
one is a design choice and not a conflict resolution.

`main` is now `4ef9981d7` (#4211, "every mutating tool row renders the change
that call made"). It shipped **the same producer this PR argues for**, plus a
*different* consumer fix from the one proposed here.

### What of this PR is already on `main`

The producer half — #4200's, which this PR's "per-call measurement wins"
section makes the case for — is in, unchanged:

```
$ git ls-tree origin/main --name-only -- crates/stella-tools/src/call_measure.rs
crates/stella-tools/src/call_measure.rs
```

So `crates/stella-tools/src/{call_measure,lib,registry}.rs`,
`registry/tests.rs`, and `crates/stella-cli/src/{agent,agent/persistence,command_deck,turn_files}.rs`
— eight of this PR's seventeen files — are settled. This PR and #4211 agree on
that half; the argument in the "producer" section above won.

### What is genuinely still open

The consumer. #4211 did **not** take either of the two expressions this PR
rejects. It resolved the "both orderings are live" problem by *deriving which
producer answered*, from state the consumer already holds — record the path's
`changes` at `ToolStart`, take it at `ToolResult`, then
`seq = if recorded > base { recorded } else { recorded + 1 }`:

```
$ git grep -c mutation_baselines origin/main -- crates/stella-tui/src
crates/stella-tui/src/model.rs:4
crates/stella-tui/src/model/tests/producer_seq.rs:1
```

This PR's `crates/stella-tui/src/model/inline_diff.rs` is not on `main` and is
the real remaining delta. Both designs reach "do not predict"; they disagree on
*how*. #4211 keeps the seq computed but makes the computation
producer-aware; this PR replaces the computation with an observed claim.

So the question for whoever owns this is narrow and worth answering on the
merits rather than by merge order: **is claiming from an observed change better
than deriving the producer, now that the latter has shipped and is green?** If
yes, this becomes a focused replacement of `mutation_baselines` — a much
smaller PR than the one currently open, and the eight settled files should come
out of it. If no, this can close and cite #4211.

What it should not become is a rebase of all seventeen files, since eight of
them would be re-landing what is already there.

### Status

`mergeStateStatus: DIRTY` — conflicting against `main` since #4211 merged, so
it cannot merge as it stands regardless of which answer is right.

For context on how the underlying contradiction was settled: #4200 was
split — its #4163 parity half merged alone as `b0833f64c`, its #4175 half was
reverted out and re-landed via #4211 with #4205's fix. The measurement that
established the two fixes were incompatible (9 failures one way, 1 real
misattribution the other) is recorded on #4200 and #4205.

Not merging this one either way — flagging it so the decision is made
deliberately rather than by whoever rebases first.

macanderson:
#4211 merged as `4ef9981d7`, so the producer and the seq contract are settled on main. Closing this rather than re-litigating — two rival PRs is #4203's own failure mode and it should not persist past the decision.

What carried over, so nothing here is lost:

- **#4227** — the one case where the two designs actually differ, now filed against main and re-confirmed there (not just on the PR branch): a successful call that measured nothing stamps `recorded + 1`, which claims whatever measures that path next. Repro, both reachability premises verified in source, and the trade stated in both directions. It asks for a decision or an explicit accepted-limitation note, not for this branch's design.
- **#4213** — bash/MCP mutations are measured per call but `is_file_mutation`'s four-name list means no row can claim them.
- **#4214** — a call mutating several paths can only render one.

The branch `worktree-fix-4203-inline-diff-claim` is left in place: it has the `ClaimWindow` consumer, its witnesses, and a green `make gate` plus CI `fmt + clippy + test` on `7aea631aa`, if #4227 is ever resolved in that direction. Nothing needs to be taken from it otherwise.

For the record on the parts of this that were independent of the design choice: the producer half here was #4200's and the diagnosis was #4205's — this PR's only original contributions were the claim-window consumer and the #4227 finding.

Issue #4019, opened 2026-08-19
stella-tui: a read_file result is never syntax-highlighted — not even JSON, because the line-number gutter defeats reads_as_json


macanderson:
## Problem

A `read_file` result in the Command Deck transcript renders as flat muted text with no
syntax coloring, for **every** language — including JSON, which the deck otherwise does
colour. The deck already owns a working multi-language lexer; it is simply never reached
from this path.

Two independent facts compose into the bug:

**1. The only body that earns colour is one that "reads as JSON", and a `read_file` body
never does.**

`crates/stella-tui/src/render/entry.rs:74` — `push_body_line(line, json, ..)` colours only
when `json` is true and otherwise falls through to `push_detail_line`, which paints the
whole line one muted style:

```rust
fn push_body_line(line: &str, json: bool, width: usize, out: &mut Vec<Line<'static>>) {
    if !json {
        push_detail_line(line, width, out);
        return;
    }
    ...
}
```

`json` comes from `reads_as_json(full)` at `entry.rs:456` and `entry.rs:465`, which is
(`crates/stella-transcript/src/syntax.rs:70`):

```rust
pub fn reads_as_json(text: &str) -> bool {
    let t = text.trim_start();
    t.starts_with('{') || t.starts_with('[')
}
```

But `read_file` prefixes every line with a right-aligned line-number gutter
(`crates/stella-tools/src/read.rs:443`):

```rust
let _ = writeln!(numbered, "{line_num:>6}\t{line}");
```

so the first non-blank character of any `read_file` body is a **digit**. `reads_as_json`
is therefore `false` for a read of a `.json` file just as it is for a `.rs` file, and the
JSON path is unreachable from `read_file` entirely.

**2. The lexer that would colour it exists and is wired to four other surfaces, but not
to tool results.**

`crates/stella-tui/src/syntax.rs` has `Lang::{Rust, TsJs, Python, Markdown, Toml, Json}`,
`tokenize`, `tok_style`, and `lang_from_path`. Its own module doc names its consumers:
diff bodies, fenced code blocks in rendered markdown, and the skills/agents definition
editors. Tool results are not among them.

## Why it is not a one-line fix

`TranscriptEntry::ToolResult` (`crates/stella-tui/src/model.rs:267`) carries `call_id`,
`name`, `ok`, `summary`, `full`, `duration_ms`, `speculated`, `diff` — and **no path**.
There is nothing on the result entry to hand `lang_from_path`. The path lives on the
[...]

PR #4248, opened 2026-08-22
fix(stella-tui): the transcript shows the diff and nothing else

macanderson:
Three rules, one outcome: **a row says what happened, or it says nothing.**

## A read shows no body

SPEC 2 folds what did not change, and a read changed nothing. The row printed five lines of a file nobody asked to see plus `⋯ 56 lines · ctrl+o` — the transcript spending its scarcest resource on the one event with nothing to report. The head's `N lines` is the fact; `ctrl+o` is where the file lives, still highlighted in its own language.

## A mutation shows its diff, or stays quiet

`edit_file` answers:

```
replaced 1 occurrence(s) in crates/stella-tui/src/views/issues.rs at byte 1286 (file sha256/8 e951e674)
```

The path the head already names, an offset, and a truncated hash — rendered whenever the diff had not arrived, reading as though *that* were the report.

**That is the whole of "sometimes it shows the diff, sometimes it doesn't."** The measurement lands with the `FileChange`, which can be a beat behind the result; until it does, the row fell back to prose. Nothing differed between a good row and a bad one but timing. Now a mutation with no diff draws no body.

A **failure** still shows its body whatever the tool — the point of reading a transcript when something breaks is to see why, and that argument does not care which tool broke.

## Git plumbing is stripped

`index 74f38f3..9e6e426 100644`, `--- a/<path>`, `+++ b/<path>` — three rows at the head of every diff, restating the path and a pair of blob hashes nobody can act on. On a collapsed diff that is most of the row budget spent before the first changed line.

**Only outside a hunk.** Inside one, `--- `/`+++ ` is body content that looks like a header (an SQL comment, a diff of a diff), and hunk position is the only thing that tells them apart. My first cut stripped on text alone and deleted real removed lines — `body_lines_number_hunk_text_matching_header_syntax` caught it. `@@` survives: it carries the enclosing scope, the one piece of context a hunk cannot reconstruct from its own lines.

## A tool has three names

`ToolEntry` gains `label`, declared in the one catalog table:

| | answers |
|---|---|
| `name` | what the **model** must spell in a call |
| `description` | what the model reads to decide *whether* to call |
| `label` | what a **reader** sees after: `read`, `edit`, `write`, `delete`, `run` |

Lowercase — they sit inline in a sentence-shaped row, not as column headings.

**It cannot reach the model.** Every provider adapter builds its own tool JSON from `name`/`description`/`input_schema` explicitly, so a field here cannot perturb the prompt-cache prefix (invariant 7). That is what makes a display concern admissible beside the dispatch ones.

Only the twelve verbs are relabelled; everything else keeps its own name. My first cut defaulted to `name.replace('_', " ")` and turned `get_environment` into `get environment` — an identifier rewritten as prose, no longer searchable. `an_argument_less_call_prints_no_empty_object` caught it.

For a tool this host has no verb for, `label_for` strips **only** the routing prefix: `mcp__fs__read_file` → `read_file`. Never translated to `read` — a server's `read_file` is not necessarily this workspace's, and a familiar verb on a row that did something else is the one error a transcript cannot afford.

An unrecognised tool keeps its name as the *subject* in the text tone rather than gaining a gold verb. `a_tool_name_is_hued_by_what_kind_of_call_it_was` holds that line, and it taught me my first attempt was over-reach: gold is reserved, and every name wearing it is what made the accent mean nothing.

## Tests

**Five moved, not deleted.** The read-body colouring tests (#4019/#4020/#4036) now assert against the **expanded** rows, where that feature still lives — the coverage is intact, one keystroke away.

Three new witnesses: a collapsed read leaks no file; an edit with no diff prints no prose; a failed edit still says why. Plus `a_full_git_patch_renders_only_its_hunk`.

## Verification

`cargo test --workspace` green, `cargo clippy --workspace --all-targets -D warnings` clean, `make guards` green.

## Tests deleted

None.

## Summary by Sourcery

Make Stella TUI transcript rows concise and truthful by showing actionable diffs or failures while removing redundant file prose and Git metadata.

New Features:
- Add reader-facing tool labels that present built-in actions as concise verbs and preserve meaningful names for unknown or MCP tools.
- Suppress successful file-tool prose and collapsed read bodies so transcript rows show only meaningful results, while retaining failure explanations and expandable file content.

Bug Fixes:
- Prevent transcript rows from displaying misleading tool output when an edit diff is unavailable or a read has no changed content.
- Remove Git patch plumbing from rendered diffs without stripping header-like content inside hunks.

Enhancements:
- Preserve hunk headers and correct line numbering while rendering only actionable diff content.
- Use catalog metadata and safe fallback naming to distinguish model tool identifiers from transcript-facing labels.

Tests:
- Update transcript rendering tests for expanded read output and add coverage for quiet collapsed reads, diff-less edits, failed edits, MCP tool labels, and full Git patch rendering.


macanderson:
This merged with an unresolved conflict in `crates/stella-tui/src/diff.rs`, and `main` has not compiled since:

```
error: expected `{`, found keyword `if`
   --> crates/stella-tui/src/diff.rs:307:9
note: the `if` expression is missing a block after this condition
   --> crates/stella-tui/src/diff.rs:299:12
```

The `main` merge in `223db2bcc` resolved `render_body` by keeping **both** sides' guard — the first with no block, naming a bare `hunk_headers` that stopped existing when #4244 moved it onto `Chrome`. `was_in_hunk` is left unused. Everything downstream of `stella-tui` fails with it.

Underneath the syntax error is the real conflict, and it is why the merge had something to resolve: #4244 landed the plumbing-stripping first and scoped it to the **inline** surface, keeping the file boundary in the standalone viewer (and inline too, once a patch spans two files). This PR stripped it everywhere and rewrote five tests to match.

Fix in #4265 — `diff.rs` returns to its pre-merge state, which is the whole of this PR's change to that file. The other six files here are untouched and keep working: the quiet-mutation rows, the read-body fold, and the `ToolEntry::label` verbs all stand. The one thing carried forward is `a_full_git_patch_renders_only_its_hunk`, retargeted at `body_lines_inline` where the behaviour now lives.

One note for the record: that test's patch fixture used `\`-continuation, which eats each line's **leading** whitespace — so its context row had no `' '` marker, a shape git never emits. The assertions happened not to reach it. The retargeted version spells the literal with explicit `\n`.

Issue #4259, opened 2026-08-22
The standalone diff viewer's file preamble: #4244 and #4248 answered this differently and neither answer was chosen


macanderson:
## Problem

`crates/stella-tui/src/diff.rs` renders diffs through one `render_body`, with a
`Chrome` struct deciding which structural rows survive:

```rust
struct Chrome {
    hunk_headers: bool,  // draw `@@ -a,b +c,d @@`
    file_headers: bool,  // draw `diff `, `index `, `--- `, `+++ `
}
```

- `Chrome::VIEWER` (both `true`) — `body_lines` / `body_lines_capped`, the
  standalone `/diff` panel.
- `Chrome::inline_for(diff)` — thresholds: hunk headers survive from two hunks
  up, file headers from two files up.

Two PRs that merged six minutes apart disagreed about what the **viewer** half
should do, and the disagreement was never decided — it was resolved by a
compile error (see #4248's merge and the repair PR below).

| | position on the standalone viewer |
|---|---|
| #4244 | **keeps** the preamble — "The standalone diff viewer (`body_lines_capped`, the `/diff` panel) keeps both, unchanged: there, boundaries are what a reader is navigating by." |
| #4248 | **strips** it — its implementation was unconditional, and its test doc said "This used to assert the opposite half: that `--- a/x.rs` and `+++ b/x.rs` rendered with a blank gutter. They render nowhere now." |

The repair PR kept #4244's split, because that is the implementation that
survived the merge intact and the one with a documented threshold model. It
did **not** decide the design question, and says so.

## The question

Should `Chrome::VIEWER` keep `file_headers: true`?

Arguments on record from each side:

- **Keep (#4244).** In a viewer the reader is navigating a patch, and the file
  boundary is the thing they navigate by. A `/diff` panel showing a multi-file
  patch with no `diff --git` rows gives no way to tell where one file ends.
  Note `inline_for` already re-admits them at two files, so the two designs
  only differ for a *single-file* patch in the viewer.
- **Strip (#4248).** `index 74f38f3..9e6e426 100644` is a pair of blob hashes
  nobody can act on, and `--- a/<path>` / `+++ b/<path>` restate a path the
  panel header already names. For a single-file patch that is three rows of
  the reader's budget spent before the first changed line.

A middle option nobody has argued yet: keep `--- `/`+++ ` and drop `index `,
which is the only one of the three that carries nothing a human reads.

## Files

[...]

PR #4260, opened 2026-08-22
fix(stella-tui): main compiles again — one diff condition, not two

macanderson:
## What

`main` has not compiled since 04:02 UTC. `crates/stella-tui/src/diff.rs:307`:

```
error: expected `{`, found keyword `if`
```

#4244 and #4248 merged six minutes apart and both fixed the same thing — an
inline diff opening with rows of git plumbing. #4248's branch was cut before
#4244 landed, so it carried its own implementation against the pre-#4244
signature. The merge kept **both**, textually:

```rust
if plan.shows(i)
    && !(!was_in_hunk && is_meta(text))
    && (hunk_headers || !text.starts_with("@@"))
// Chrome is suppressed after planning rather than before it, so an
// elided row's `⋯ n lines` count stays the one `stella_diff::view` computed.
if plan.shows(i)
    && (chrome.hunk_headers || !text.starts_with("@@"))
    && (chrome.file_headers || !preamble)
{
    lines.push(line);
}
```

A condition with no block, followed by a second `if`. Git had no conflict to
report — neither side's text contradicts the other line-for-line.

## Three red checks, one root cause

#4248's own CI failed on **MSRV**, **fmt + clippy + test**, and **docs/tools
still matches the declarations**, and it merged anyway. All three are the same
compile error: `stella-cli` depends on `stella-tui`, so
`scripts/check-tool-docs.sh`'s `cargo test -p stella-cli --bin stella tool_docs`
could not build. That guard prints its "you changed a tool — regenerate"
guidance on any non-zero exit, which is what made a downstream build failure
read as an independent stale-docs break. With the crate compiling,
`./scripts/check-tool-docs.sh` reports `OK` with no regeneration.

## Which side survives

#4244's. It is the structurally complete one:

- `Chrome { hunk_headers, file_headers }` carries the two thresholds as a
  struct, with `Chrome::VIEWER` and `Chrome::inline_for`.
- `preamble` is `!was_in_hunk && is_meta(text)` under a different name — the
  two PRs computed the identical value.
- `chrome.file_headers` is the multi-file escape #4248's unconditional strip
  did not have.

#4248's fragment also referenced `hunk_headers`, a `render_body` parameter
#4244 replaced with `chrome: Chrome` — so it could not have compiled even
with a block added.

## The one thing given up, deliberately

The two PRs disagreed about the **standalone viewer**, not only the inline row:

| | `body_lines_capped` / the `/diff` panel |
|---|---|
| #4244 | keeps both boundaries — "there, boundaries are what a reader is navigating by" |
| #4248 | stripped unconditionally; its tests asserted plumbing "render[s] nowhere now", asked through `body_lines` |

This PR keeps #4244's split and restores the four viewer tests #4248 renumbered
(`body_numbers_added_lines_on_the_new_side_and_removed_on_the_old`,
`file_headers_and_hunks_get_no_number`,
`body_lines_number_hunk_text_matching_header_syntax_instead_of_hiding_it`,
`no_newline_marker_gets_no_number_and_does_not_shift_later_numbering`).

**#4248's actual goal is untouched.** "The transcript shows the diff and nothing
else" is about the inline row, and it is still proven — by #4248's own
`render/tests/{inline_diff,tool_output,result_row}.rs` and by #4244's
`mutation_diff_e2e::the_patch_preamble_is_not_drawn_as_diff_body`, all passing.
#4248's new `a_full_git_patch_renders_only_its_hunk` is **kept**, retargeted
from `body_lines` to `body_lines_inline` — the surface whose claim it is.

Whether the viewer should *also* drop the preamble is a live design question
this PR does not settle. Filed as #4259 so the choice stays a maintainer's.

## Deleted tests

`plumbing_is_stripped_and_the_hunk_header_keeps_a_blank_gutter` is **renamed
back** to `file_headers_and_hunks_get_no_number`, its pre-#4248 name and its
pre-#4248 assertions. No test is removed on net; the count goes up by zero and
`a_full_git_patch_renders_only_its_hunk` survives.

## Witness

```
cargo test -p stella-tui     # 952 passed; 0 failed
./scripts/check-tool-docs.sh # OK — docs/tools/ matches the declarations
```

On the parent commit (`8a2fdf8cc`) the crate does not build at all — that is
the fail half of the flip.

Refs #4259

## Summary by Sourcery

Restore the diff renderer build while preserving the intended distinction between standalone diff viewing and inline tool-output rendering.

Bug Fixes:
- Restore compilation of the diff renderer by removing the duplicate conditional left by the conflicting merges.

Enhancements:
- Preserve separate standalone-viewer and inline-rendering behavior, keeping file headers and hunk boundaries in the viewer while suppressing diff plumbing inline.
- Retain and realign coverage for inline full-patch rendering and standalone line-numbering behavior.

Tests:
- Restore the viewer tests for file headers, hunk headers, line numbering, and special diff content, while targeting the full-patch test at inline rendering.

### DIFF
--- a/crates/stella-tui/src/diff.rs
+++ b/crates/stella-tui/src/diff.rs
@@ -264,18 +264,14 @@ fn render_body(
         // Every line advances the gutter counters, shown or not — skipping a
         // line must not renumber the ones after it, and the numbers on the
         // far side of an elision have to be the file's real ones.
-        // Captured before `body_line` advances it: a `--- ` or `+++ ` line is
-        // plumbing only *outside* a hunk. Inside one it is body content that
-        // happens to look like a header — an SQL comment, a diff of a diff —
-        // and hunk position is the only thing that can tell them apart
-        // (`count_diff_lines` makes the same distinction for the same reason).
-        // Stripping on the text alone deletes real removed lines.
-        let was_in_hunk = in_hunk;
+        //
         // Read *before* `body_line` runs, because that call is what moves the
         // state this asks about. The same structural rule `body_line` itself
         // applies (only before a file's first hunk is `+++ `/`--- ` a header):
-        // inside a hunk those bytes are added or removed source text, and
-        // dropping them would delete a line of the change.
+        // inside a hunk those bytes are added or removed source text that
+        // happens to look like a header — an SQL comment, a diff of a diff —
+        // and dropping them would delete a line of the change.
+        // (`count_diff_lines` makes the same distinction for the same reason.)
         let preamble = !in_hunk && is_meta(text);
         let line = body_line(
             text,
@@ -293,12 +289,6 @@ fn render_body(
         // line. The counters above still advance over them, so the gutter's
         // numbers stay the file's own.
         //
-        // `@@` survives on the `hunk_headers` flag: it carries the enclosing
-        // scope (`@@ … const TYPEAHEAD_MAX_ROWS`), which is the one piece of
-        // context a hunk cannot reconstruct from its own lines.
-        if plan.shows(i)
-            && !(!was_in_hunk && is_meta(text))
-            && (hunk_headers || !text.starts_with("@@"))
         // Chrome is suppressed after planning rather than before it, so an
         // elided row's `⋯ n lines` count stays the one `stella_diff::view`
         // computed and every surface reports the same number for one change.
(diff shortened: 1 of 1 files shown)


macanderson:
## Evidence that this restores #4244 and invents nothing

The claim in the description is that the surviving implementation is #4244's,
unmodified. That is checkable rather than assertable — diff this branch's
`diff.rs` against the one #4244 shipped:

```
git show 8617d47a8:crates/stella-tui/src/diff.rs > a.rs
git show HEAD:crates/stella-tui/src/diff.rs      > b.rs
diff -u a.rs b.rs | rg '^[-+]' | rg -v '^[-+]\s*//'
```

Every non-comment line in that output belongs to one added test,
`a_full_git_patch_renders_only_its_hunk` — #4248's, kept and pointed at
`body_lines_inline`. There is **no** executable change to `render_body`,
`Chrome`, `Chrome::VIEWER` or `Chrome::inline_for` relative to #4244.

So the production behaviour of this PR is exactly what `main` had at
`8617d47a8`, the last commit on which the workspace compiled, plus one test.

## Why the tool-docs failure was not a second break

`scripts/check-tool-docs.sh` buffers its build and prints the "you changed a
tool — regenerate" guidance on **any** non-zero exit, including a compile
failure:

```sh
if ! log="$(cargo test --quiet -p stella-cli --bin stella tool_docs 2>&1)"; then
  echo "check-tool-docs: FAIL — docs/tools/ no longer matches the declarations." >&2
```

`stella-cli` depends on `stella-tui`, so while `stella-tui` did not compile,
that guard could not distinguish "the tool surface drifted" from "the tree does
not build". #4248 did change `crates/stella-tools/src/catalog.rs` and did not
touch `docs/tools/`, which made a stale-docs reading look likely — it wasn't.
With the crate compiling and no regeneration performed:

```
$ ./scripts/check-tool-docs.sh
check-tool-docs: OK — docs/tools/ matches the declarations.
```

Worth a follow-up on its own: that guard's failure message names a cause it has
not established. Filed separately rather than widened here.

macanderson:
Verified locally on this branch: `cargo build -p stella-tui` compiles, `cargo test -p stella-tui` is 952 + 7 + 5 + 3 + 16 green, and `a_full_git_patch_renders_only_its_hunk` (the behaviour witness from #4248) still passes — so the plumbing stripping survives through `chrome.file_headers`.

This is the right resolution, and worth stating why: `chrome.file_headers || !preamble` **already does what #4248's `is_meta` condition did**. The two are the same feature arrived at twice, so dropping mine and keeping `chrome`'s is a deduplication rather than a revert.

The break was mine. #4248 was cut before the `chrome` refactor and merged after it; git stacked both `if plan.shows(i)` conditions with no conflict to report, because neither side's text contained the other's. That is the same shape as #4133 and the #4123/#4129 golden clobber earlier this week — a PR that edits a hot region has to be re-checked against the merge base immediately before landing, not just at cut time.

Note #4265 is a second fix for the same red (`fix/4248-merge-residue-main`, 118 lines vs this one's 86). Both should not land. This one is the smaller change and is green; I'd merge it and close the other, but I have not touched either since they are not mine.

All checks green here except the usual blocked-account Vercel row.

PR #4269, opened 2026-08-22
fix(stella-tui): main compiles again — one condition decides a diff row, not two

macanderson:
Closes #4256.
Refs #4268.

## What is broken

`stella-tui` has not compiled since #4248 merged. `make build-release` on `main`:

```
error: expected `{`, found keyword `if`
   --> crates/stella-tui/src/diff.rs:307:9
```

#4248 pasted a replacement for `render_body`'s row-admission condition **above** the existing one and never deleted the old block, leaving two consecutive `if plan.shows(i)` expressions, the first with no body. The pasted draft was stale as well — it reads a bare `hunk_headers`, the parameter `render_body` took before #4244 replaced it with `chrome: Chrome`, so it could not have compiled standing alone either.

`main` has been red since `9c8b5511a`; `f01b2602e` and `8a2fdf8cc` both carried it forward.

## Why this is not a one-character fix

The two blocks disagree about behaviour, not just syntax:

| | preamble (`diff `, `index `, `--- `, `+++ `) |
|---|---|
| #4244's block | conditional on `chrome.file_headers`, which `Chrome::VIEWER` sets **true** |
| #4248's block | stripped **unconditionally** |

Deleting either one alone is wrong:

- **Keep only #4244's** and #4248's two new tests fail. Both call `body_lines`, which is the `Chrome::VIEWER` path, and both assert the plumbing renders nowhere — `plumbing_is_stripped_and_the_hunk_header_keeps_a_blank_gutter`, `a_full_git_patch_renders_only_its_hunk`.
- **Keep only #4248's** and `Chrome::file_headers` is written by both constructors and read by nothing — a `dead_code` denial away from red again.

## What this does

Keeps #4248 — the later decision, and the only one carrying tests — and retires `Chrome::file_headers`, which no longer has anything left to decide. One condition survives:

```rust
if plan.shows(i) && !preamble && (chrome.hunk_headers || !text.starts_with("@@")) {
```

`preamble` is the binding already computed before `body_line` runs, and is the same predicate as the deleted `!was_in_hunk && is_meta(text)` — so the hunk-position rule that stops `--- ` *inside* a hunk being read as plumbing is unchanged, and `body_lines_number_hunk_text_matching_header_syntax_instead_of_hiding_it` still guards it. Docs on `body_lines`, `body_lines_inline` and `Chrome` are updated to match, rather than left describing the flag that is gone.

## The part this deliberately does not decide

[...]

Issue #4268, opened 2026-08-22
stella-tui: decide whether the standalone diff viewer should regain the per-file preamble


macanderson:
Follow-up from the repair of #4256.

## What happened

`crates/stella-tui/src/diff.rs` gates diff chrome through a `Chrome` struct passed to `render_body`. Two PRs disagreed about the per-file preamble (`diff `, `index `, `--- `, `+++ `) and the disagreement was never resolved, because the PR that changed it did not compile:

- **#4244** (`8617d47a8`) added `Chrome { hunk_headers, file_headers }`. `Chrome::VIEWER` set `file_headers: true` — the standalone viewer draws the preamble; `Chrome::inline_for` set it to `diff.lines().filter(|l| l.starts_with("diff ")).count() > 1` — inline draws it only for a genuinely multi-file patch.
- **#4248** (`9c8b5511a`) decided the preamble renders **nowhere**, and shipped two tests asserting it against `body_lines`, which is the `Chrome::VIEWER` path: `plumbing_is_stripped_and_the_hunk_header_keeps_a_blank_gutter` and `a_full_git_patch_renders_only_its_hunk`.

#4248 pasted a second `if plan.shows(i)` above the existing one instead of replacing it, so `stella-tui` stopped compiling and neither semantics was ever exercised on `main`.

## What the repair did

The repair of #4256 kept #4248 — the later, test-backed decision — and retired `Chrome::file_headers`, which no longer had anything to decide. `Chrome` is now a one-field struct.

## The open question

Whether #4244's intent was right for the **standalone** viewer specifically. The two surfaces are not obviously the same case:

- `crates/stella-tui/src/views/files.rs:307` and `crates/stella-tui/src/render.rs:318` go through `Chrome::VIEWER`. A reader navigating a whole multi-file patch in the FILES tab has no other file boundary to orient on once `diff --git a/x b/x` is gone.
- `crates/stella-tui/src/render/entry.rs:778` and `crates/stella-cli/src/plain.rs:475` go through `Chrome::inline_for`. Inline under a tool call the row above already names the path, which is #4248's argument and is not in dispute.

[...]

PR #4276, opened 2026-08-22
fix(stella-tui): main's diff tests describe a rule the code stopped following

macanderson:
## What & why

**`cargo test -p stella-tui` on main: 5 failed, 947 passed.**

```
diff::tests::a_full_git_patch_renders_only_its_hunk_inline
diff::tests::body_lines_number_hunk_text_matching_header_syntax_instead_of_hiding_it
diff::tests::body_numbers_added_lines_on_the_new_side_and_removed_on_the_old
diff::tests::file_headers_and_hunks_get_no_number
diff::tests::no_newline_marker_gets_no_number_and_does_not_shift_later_numbering
```

main **compiles** — #4260/#4265/#4269 saw to that. It does not **pass**.

### How the suite and the code ended up disagreeing

Three PRs raced to fix the same unresolved merge in `render_body`, and the question that merge was about — *does the standalone viewer render the per-file preamble?* — got answered twice, in opposite directions:

| | resolution | tests left describing |
|---|---|---|
| #4265 | keep `Chrome::file_headers` → viewer **renders** it | viewer renders it ✓ |
| #4269 | `!preamble` unconditional → **no** surface renders it | viewer renders it ✗ |

#4269 is the one that stuck, and it is the better answer — it is #4248's original semantics, and it retired `Chrome::file_headers` because nothing read it any more. What it did not do is move the tests. Four of the five failures are tests still asserting the viewer renders `--- a/x.rs`; the fifth is #4265's own witness asserting exactly that.

**The code is right and the tests are stale, so the tests move.** Whether the viewer *should* get the preamble back is a real design question and it is already tracked as #4268. This PR does not answer it — it stops the suite from asserting an answer nobody chose.

### What changes

- `file_headers_and_hunks_get_no_number` → `the_preamble_does_not_render_and_the_hunk_header_keeps_a_blank_gutter`. It asserted the preamble rendered *with a blank gutter*; the old name promised a numbering fact it no longer tests, so the new one states the rule it actually checks.
- Three tests index `texts[]` positionally past a preamble that is no longer emitted. Each shifts by two, with a line saying why the body now starts at index 1.
- `a_full_git_patch_renders_only_its_hunk_inline` keeps asserting **both** surfaces. They share one `render_body`, and a test covering only the inline path is precisely how the viewer drifted across three PRs without the suite noticing. Its viewer half now expects the preamble stripped, and cites #4268.

## The witness

Five of them, and they are the diff itself: each currently fails on `main` and passes here. `a_full_git_patch_renders_only_its_hunk_inline` is the one that would have caught this at #4269 time had its viewer half existed then — which is the argument for keeping both surfaces in one test rather than splitting them.

## The gate

- [x] `cargo fmt --check`
- [x] `cargo test --workspace` — see caveat below
- [x] Docs updated where behavior/flags changed — no behaviour change; doc comments now match the shipped rule
- [x] CLA signed

Locally, scoped to the crate this touches: `cargo test -p stella-tui` **1020 passed, 0 failed**; `cargo clippy -p stella-tui --all-targets -- -D warnings` clean; `cargo fmt -p stella-tui -- --check` clean.

**Caveat, stated plainly:** `cargo doc -D warnings` fails on `main` for a reason unrelated to this PR and unrelated to `stella-tui` —

```
error: public documentation for `TurnEvidence` links to private item `turn_start_index`
  --> crates/stella-core/src/driver/loop_evidence.rs:89:19
```

It failed the same way on #4265. This PR neither causes nor fixes it; the `fmt + clippy + test` job will stay red until someone does. Filed below.

## Nothing left behind

- [x] Filed: #4268 already tracks the viewer question. The `cargo doc` break above is pre-existing on `main` and needs its own issue — I have not opened one because I could not confirm it is not already filed; if it is not, it should be, and it is a one-line rustdoc fix (make the link a plain code span, or `pub(crate)` → document it).

## Ground-rule check

**This is the second half of #4265, not a re-litigation of it.** #4265 picked the losing side of the preamble question; #4269 picked the winning one. I am not reopening that — #4269's answer stands and this PR makes the tests agree with it.

**The `tui-transcript-body` branch is back on the remote and should not be.** GitHub auto-deleted it when #4248 merged; I pushed to it before noticing the merge had landed mid-session, recreating the ref at `617b22bf9`. It is superseded and safe to delete — flagging rather than deleting a ref I created by accident.

**The real lesson is in the race, not the tests.** Three PRs fixed one broken merge within minutes of each other, and the two that landed disagreed about behaviour while both claiming to "make main compile again". Compiling was the visible symptom; the conflict was a design question, and a green build is not evidence that it got answered.

## Summary by Sourcery

Align diff rendering tests with the existing behavior that strips per-file preambles while preserving hunk and body line-numbering rules.

Bug Fixes:
- Update stale diff-rendering tests to match the rule that per-file preamble lines are not rendered on any surface.

Enhancements:
- Clarify test names, expectations, and comments around hunk headers, line numbering, and preamble handling.
- Keep inline and standalone viewer assertions aligned to prevent future rendering drift.

Tests:
- Adjust diff tests for the removed preamble output and corrected positional line indexing.

### DIFF
--- a/crates/stella-tui/src/diff.rs
+++ b/crates/stella-tui/src/diff.rs
@@ -755,36 +755,55 @@ mod tests {
     fn body_numbers_added_lines_on_the_new_side_and_removed_on_the_old() {
         let lines = body_lines(SAMPLE, None);
         let texts: Vec<String> = lines.iter().map(line_text).collect();
+        // The `---`/`+++` pair does not render, so the `@@` header leads and
+        // the first numbered row is at index 1.
         // "@@ -1,3 +1,4 @@" starts old=1/new=1; context takes new 1.
-        assert!(texts[3].starts_with("   1  context"), "{:?}", texts[3]);
+        assert!(texts[1].starts_with("   1  context"), "{:?}", texts[1]);
         // The removal is numbered on the OLD side (old line 2).
-        assert!(texts[4].starts_with("   2 -old line"), "{:?}", texts[4]);
+        assert!(texts[2].starts_with("   2 -old line"), "{:?}", texts[2]);
         // Additions continue on the NEW side (new lines 2, 3).
-        assert!(texts[5].starts_with("   2 +new line"), "{:?}", texts[5]);
-        assert!(texts[6].starts_with("   3 +another add"), "{:?}", texts[6]);
+        assert!(texts[3].starts_with("   2 +new line"), "{:?}", texts[3]);
+        assert!(texts[4].starts_with("   3 +another add"), "{:?}", texts[4]);
     }
 
+    /// The per-file preamble does not render on any surface, and the hunk
+    /// header — which does — carries no line number.
+    ///
+    /// This asserted the other half until #4248: that `--- a/x.rs` and
+    /// `+++ b/x.rs` rendered *with* a blank gutter. They render nowhere now.
+    /// Whether the standalone viewer should get them back is #4268; until that
+    /// is answered one rule covers every surface, and this says so in the one
+    /// place a reader looks for it.
     #[test]
-    fn file_headers_and_hunks_get_no_number() {
-        let lines = body_lines(SAMPLE, None);
-        for (i, line) in lines.iter().take(3).enumerate() {
+    fn the_preamble_does_not_render_and_the_hunk_header_keeps_a_blank_gutter() {
+        let texts: Vec<String> = body_lines(SAMPLE, None).iter().map(line_text).collect();
+        for meta in ["--- a/x.rs", "+++ b/x.rs"] {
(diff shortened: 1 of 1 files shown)


macanderson:
Correction to the gate caveat above: the `cargo doc -D warnings` failure on `TurnEvidence` was fixed by #4274 (`069e62d37`) while this was being written. `fmt + clippy + test` should go green here on its own — nothing to file, and the "Nothing left behind" note about it is moot.

`diff.rs` is unchanged on `main` since this branch's base, so the five failures are still live and this still applies cleanly.

Issue #4188, opened 2026-08-21
tui-v2: SPEC 5 sends five metrics behind `?` as well as the AGENTS tab — the `?` overlay renders keybindings only


macanderson:
## Problem

SPEC 5 re-homes five status cells to two places: "MODEL detail, CPU, MEM,
WARMTH, and ENGINE move behind `?` **and** the AGENTS tab." SPEC 11 backs the
first half — `?` is listed as "help, full metric detail".

The AGENTS half shipped. `crates/stella-tui/src/views/agents.rs` carries
`CPU%`, `MEM`, `Cache`, `Warmth` columns (`HEADERS`, line ~32) and the
active/total count in the EXECUTIONS header, and #4186 added the lane split
beside it.

The `?` half did not. `render_help` (`crates/stella-tui/src/deck_render.rs:1452`)
builds exactly two things: the active tab's shortcuts and `GLOBAL_SHORTCUTS`.
There is no metric block in it — no MODEL detail, no CPU, no MEM, no WARMTH,
no ENGINE. `?` today is a keybinding sheet, and the spec says it is also the
metric detail view.

## Why it matters

This is the *other* half of #4126. That issue was about three cells with no
home at all; this is about five that have one home where the spec promises
two. It matters most for MODEL detail: the bar shows one slug (whichever pin
is answering), and "which pin is serving each role" was a v1 cell with no
surface left except the AGENTS tab's per-row model column, which answers a
different question.

It is also the smaller half by far — the numbers are all already on
`WorkspaceModel`/`AgentEntry`, so this is a rendering task, not a plumbing one.

## Options

1. A metric section in the `?` overlay, above the shortcuts. It already scrolls
   (`ui.metrics.help_total` / `help_height`, `handle_help_key`), so length is
   not the obstacle it would have been.
2. Decide the AGENTS tab is enough and **amend SPEC 5 and SPEC 11** to say so.
   That is a legitimate answer, but it has to be written down — SPEC 5 promising
   a surface that does not exist is the same silent-drift shape #4126 was.

## Files

- `crates/stella-tui/src/deck_render.rs:1452` (`render_help`), `GLOBAL_SHORTCUTS` above it
- `crates/stella-tui/src/views/agents.rs` — where the same values already render
- `design/tui-v2/SPEC.md` §5 and §11

## Verify

`crates/stella-tui/tests/snapshots/deck/overlay_help_skills.txt` is the help
overlay's golden — it contains shortcuts and no metrics. A witness is a
[...]

Issue #4283, opened 2026-08-22
stella-transcript: decide whether the shared lexer upgrades to a grammar-backed one — and land it here, never deck-local


macanderson:
## The question

`stella_transcript::syntax` is the one lexer behind all three transcript
surfaces — the deck (`crates/stella-tui/src/syntax.rs`, which re-exports it and
supplies only a ratatui palette), the export grid
(`crates/stella-transcript/src/grid.rs::tok_color`) and the Observatory HTML
renderer (`crates/stella-transcript/src/html.rs::tok_class`). It is
hand-written, and it recognises six languages
(`crates/stella-transcript/src/syntax/lang.rs`: `Rust`, `TsJs`, `Python`,
`Markdown`, `Toml`, `Json`) across four token classes
(`crates/stella-transcript/src/syntax.rs:51`: `Keyword`, `Str`, `Number`,
`Comment`).

That is deliberately narrow, and narrow is a real cost: a `read_file` of a Go,
Java, C, shell, SQL, YAML or HTML file renders as flat text on every surface,
and even inside the six the lexing is heuristic rather than grammar-accurate.
A grammar-backed lexer (tree-sitter, or syntect) would fix both.

**The point of this issue is where such an upgrade lands, not whether it is
worth doing.** It lands *here*, in `stella-transcript`, so all three renderers
gain it in one change. It must never land as a deck-local highlighting crate:
that is a fourth copy of a rendering decision this tree has already
de-duplicated twice — the JSON lexer moved down in #3644 and the other five
followed in #4036, precisely because three copies drift. `design/tui-v2/
IMPLEMENTATION-PLAN.md` §2/§3 used to order exactly that deck-local crate
(`stella-tui-highlight`, syntect-based); #4196 removed the order and points
here instead.

## Constraints a solution must satisfy

- **`stella-transcript` is a near-leaf by contract.** Its `Cargo.toml` header
  states the rule and today it carries exactly two non-serde edges
  (`stella-diff`, `unicode-width`). A tree-sitter or syntect dependency is a
  large addition and needs the "sentence in the PR description" AGENTS.md asks
  for — plus a `cargo deny check licenses` pass, since a rejected license means
  dropping the dependency, not widening `deny.toml`.
- **No I/O, no allocation-heavy startup** (invariant #2). syntect's default
  path loads grammar/theme assets at runtime; tree-sitter grammars are compiled
  in. Whichever is chosen must stay a pure function over borrowed text.
[...]

Issue #4271, opened 2026-08-22
tui-v2: v2/transcript.rs is a fourth transcript painter — decide whether SPEC 6 or grid.rs is the one frame


macanderson:
## Problem

`crates/stella-tui/src/v2/transcript.rs` (600 lines) is a **fourth** transcript painter. It shares nothing with `stella-transcript`: its own turn frame, its own glyph vocabulary, its own rail, its own receipt row, painted straight to `ratatui::text::Line` over `stella_tui_theme` tokens.

The workspace therefore renders one session four different ways:

| Surface | Model | Painter |
|---|---|---|
| `stella run` / plain (`crates/stella-cli/src/plain/transcript.rs`) | `stella_transcript::model` | `stella_transcript::grid` |
| Observatory `stella observe` (`crates/stella-observatory/src/lib.rs:183`) | `stella_transcript::model` | `stella_transcript::html` |
| Command Deck v1 (`crates/stella-tui/src/render/entry.rs`, `render/row.rs`, `diff.rs`) | its own `TranscriptEntry` | its own | (#3578) |
| Command Deck v2 (`crates/stella-tui/src/v2/transcript.rs`) | its own | its own | (this issue) |

v2's module doc states its shape as SPEC 6:

```
── turn 14 · execute · kimi-k3 · budget $0.60 ───────────────────────────
 │ ✦ skill oxagen-feature · auto                              1.2k tok
 │ ▸ read …/lifecycle.rs · 221 lines                    ⚡3ms · ↵ open
── turn 14 done · 0:42 ──────────────────────────────────────────────────
   receipt $0.11 · 18k tok · det 86% · 4/4 tests · 2 files · ↵ audit
```

`grid.rs` draws the same information as `╭─ … ─╮` / `│` / `╰──╯` with `digest::Chip` right-aligned. Neither is wrong; they are two answers to one question, and the deck v2 answer was written without reference to the shared one.

## What has to be decided, not just wired

This is a **design decision before it is a refactor**, and it should not be closed by picking whichever is easier to call:

1. **Does SPEC 6 supersede `grid.rs`'s frame, or the other way round?** If SPEC 6 is the intended product design, then `grid.rs` should be changed to draw it and every surface follows — including the plain surface and the Observatory. If `grid.rs` is the reference, v2 should render through it and SPEC 6's mock is stale.
[...]

Issue #4251, opened 2026-08-22
ISSUES tab: the create form's Labels type-ahead can never return a match


macanderson:
## Problem

The command deck's ISSUES tab create form (`n`) has a **Labels** field with a type-ahead popup, and that popup can never show anything. `handle_issues_input`'s `EntitySearch` arm returns an empty vector for `EntityField::Label` unconditionally:

```rust
// crates/stella-cli/src/command_deck/issues.rs
EntityField::Label => Vec::new(),
```

The Assignee half of the same popup is wired (installed agents + memories + code-graph symbols, via `agent_entity_hits` / `local_assignee_hits` / `merge_assignee_hits`). Labels is the only field of the two that answers "no matches" no matter what is typed.

This was accurate when the tab had no tracker transport at all. #4245 wired the tab to the workspace's issue provider (GitHub through `gh`), so the tracker's label vocabulary is now reachable — nothing reads it into this popup yet.

## Why it is not a one-line fix

The vocabulary has to come through the port, not around it. `IssueProvider` (`crates/stella-protocol/src/issue.rs`) has no method that lists a tracker's labels — `list_open`, `search`, `file`, `comment`, `close`, `reopen`, `relabel`, `edit`. Reaching for `gh label list` directly inside `command_deck/issues.rs` would put a GitHub CLI invocation above the port and break invariant 1 (AGENTS.md § "Architecture: ports, not direct dependencies"); it is the exact shape `crates/stella-cli/src/issue_provider.rs`'s module doc says was cleaned up in B1.

## Definition of done

1. `IssueProvider` grows a label-vocabulary read — suggested shape, matching the `search`/`reopen` precedent set by #4245:
   ```rust
   /// The labels this tracker already knows, for a picker.
   ///
   /// Default is an empty list: a provider that cannot enumerate labels shows
   /// no suggestions rather than failing the form.
   async fn labels(&self, query: &str, limit: usize) -> Result<Vec<IssueLabel>, IssueError> {
       let _ = (query, limit);
       Ok(Vec::new())
   }
   ```
2. `GhIssueProvider` implements it over `gh label list --json name --search <query> --limit <n>` (verify the flags against the installed `gh` — `--search` exists on `gh issue list`, confirm it on `gh label list` and fall back to a client-side filter if not).
3. `EntityField::Label` in `command_deck/issues.rs` serves from it, on a blocking thread like the other tracker reads, mapping each label to an `EntityHit { kind: "Label", label, description, insert }`.
[...]

PR #4296, opened 2026-08-22
fix(stella-tui,stella-cli): close out the mid-turn steering epic — the P0 Esc-steer, and 13 more TUI defects

macanderson:
Closes out **Epic #2953 (mid-turn steering)** and clears the landable part of two more TUI epics — **#4233** (a transcript row claims exactly the changes its own call made) and **#4232** (the v2 deck renders SPEC 5 and SPEC 6) — plus the standalone TUI P0/P1 backlog around them.

Every fix carries a witness that was **watched failing** before the fix. Where a witness could not honestly flip (a pure deletion, a docs change), that is said out loud in the commit rather than papered over.

> **Read this first — a parallel session was sweeping the same backlog.** While this branch was working, `main` independently landed its own fixes for five of these issues (#4285≈#4195, #4286≈#4196, #4291≈#4168, #4294≈#4187, #4295≈#4188), plus #4282 and the `render/entry.rs` split (#4284), and someone merged `main` into this branch mid-flight. Those five issues plus #4183 are **already closed by that other work** — they are described below because this branch also fixed them and the merge had to reconcile both, not to claim them. The 14 issues this PR actually closes are the ones with `Closes` trailers.

## The P0 and the P1s

| | Issue | What was wrong |
|---|---|---|
| **P0** | #4025 | `Esc`-steer never steered. The deck strips the `>` marker before sending `WorkspaceInput::Steer`, and `steer_lead` then asked `route_mid_turn` to *re-derive* the intent from that missing marker — so every markerless text against a live turn took the `Sidecar` arm and the user's correction was run by a stranger sub-session that cannot see the conversation. |
| **P1** | #4026 | The same path emptied the deck's queue mirror but not the driver's authoritative `DurableQueue`, so a held backlog dispatched **twice** and `queue.json` recorded the doubled order for the next resume to replay. |
| **P1** | #4227 | A successful tool call that changed nothing adopted the **next** change to that path as its own — the seq derivation's else arm predicted a future change instead of claiming its own. |
| **P1** | #3591 | The composer footer's steer hint was never budgeted, so a narrow row rendered `steer thturn $0.04`. Its existing test was a **false green**: at that width the left group was already truncated, and the right group happens to begin with the word `turn`, so the concatenation still contained the substring it asserted. |

## The rest of what this PR closes

- **#4213** — a `bash`/MCP call's tree change was measured per call and then dropped, because attribution was by tool *name* while measurement was by *schema*. Attribution is now positional (`ClaimWindow`), so a solo `bash` running `sed -i` renders its own diff.
- **#4180** — a v2 read head's size column was expressible and unreachable. Deleted rather than defaulted, so the field is gone rather than inert; the producer that would earn it back is filed as #4297.
- **#4268** — settled the standalone-diff-preamble question as "no", and recording the decision exposed a real defect: the preamble was dropped for a patch's **first** file and drawn for every later one, which is neither candidate rule.
- **#3742** — split `deck_ui/tests.rs` before it hit the ceiling. A pure move, verified as one: 54 `#[test]` out, 54 in, `tests.rs` 1370 → 87 lines with zero tests left.
- **#4251** — the ISSUES tab's Labels type-ahead answered `Vec::new()` unconditionally and could never return a match.
- **#4181**, **#4209** — the two measurement claims that had no render witness now have one.
- **#3741** — `theme::primary`/`primary_deep` deleted (zero callers, re-censused before deleting).
- **#4056**, **#4057** — the brand kit documented the retired v4.0 palette, and `stella-tui-prompt.md` still specified the `≡✦ stella` lockup SPEC 3.3 retires (`✦` is the *skill* glyph, so the mark and the skill row collided on every screen). Rewritten, not remapped — and `check-tokens.py`'s `MIGRATING` ledger is now **empty**, which is the evidence both issues asked for.

## Advanced but deliberately **not** closed

Each of these carries `Refs`, not `Closes`, because it satisfies part of a definition of done:

- **#3769** — the grid now hard-breaks an over-wide unbreakable unit (90 cells on ASCII, 60 on spaceless CJK, 408 through `render`). That is one of the three paths the issue names; the unbounded prose head, the fills' `.max(1)` floor, and the proptest are still outstanding.
- **#4071** — the 30° hue-separation law is now measured on the web tokens and wired into the gate as a **green** step. Its DoD also wants `hue_deg` lifted out of a `#[cfg(test)]` module; the extractor is deliberately brittle about that and fails by name if it moves.
- **#4072** — the Observatory's two light gates now declare one scheme (they were 17 roles apart, including a `--text-emph` nothing re-pointed, leaving an OS-light reader at 1.68:1). The five derived light steps still need the palette owner's ratification.
- **#3633** — the docs stop describing `hunk_review` as a key that ships. The feature is unbuilt and its five DoD items need code. The issue now carries a finding that changes that work: `HunkReview` already has four live consumers, so its ledger row is *understated*, not unaudited — promoting it to `Behavioral` would be literally true while the signal stays unreachable.

Also filed from this sweep: **#4283** (lexer upgrade decision), **#4297** (read-head size producer), **#4307** (an agent row takes its model from *any* call, so a summarizer relabels it — found by #4183's work), and a full design handoff posted on **#4185**, which was attempted and honestly abandoned rather than half-landed.

## Three decisions worth a reviewer's eye

**The merge conflict that would have been silent.** `main` added a `StepUsage` arm folding the turn's tokens (#4184); this branch added one folding the answering call's model (#4183). Resolved as **one** arm, not a guarded arm plus a fallthrough — the first matching arm wins, so a guarded `if answers_the_turn` placed first would have stopped counting an answering call's tokens, leaving every receipt short by exactly the worker's usage. Green tests, wrong numbers.

**The file-size baseline.** #3591's split took `deck_render.rs` under the limit, so it leaves the god-file list — baseline entry retired, and dropped from AGENTS.md's table and the crate README, which `make god-files` cross-checks. But `make file-size-update` also wanted to retighten **four unrelated ceilings** that had drifted on `main`. Those were restored by hand: the baseline is a shared cell every growing PR writes, and absorbing someone else's drift is precisely how it reds `main` for everyone.

**A golden blessed after reading it.** 14 deck frames were re-blessed for #4183. The diff is 14 lines, all the same cell — `model —` → `model glm-5.2`, frame width preserved, nothing else moved. That is the fix showing up: the demo scenario's statline named no model for its whole first turn.

## Deliberately out of scope

- **#4230** (split `render/entry.rs`) — `main` was actively editing that file, and moving a file another PR is fixing is how a change gets silently reverted at merge. It landed on `main` separately as #4284.
- The genuine owner calls were left for a human rather than decided quietly: **#4058** (retire the brand gradient?), **#4063** (four palette pairings miss WCAG AA — accept, reassign roles, or re-cut tokens?), **#4125** (per-tool-class hues), **#4214**, **#4271** (which of two transcript painters is the one frame). Each is a product decision the issue itself did not make.

## Testing

`make gate` — **exit 0** on the merged tree (all 34 guards, clippy `-D warnings`, rustdoc `-D warnings`, `cargo test --workspace`).

```
cargo test -p stella-tui          982 passed; 0 failed   (952 at the branch point)
cargo test -p stella-cli --bin stella   1952 passed; 0 failed
cargo test -p stella-transcript          92 passed; 0 failed
cargo test -p stella-observatory        101 passed; 0 failed
```

All 16 witness tests added by this branch were re-censused **by name** after the `main` merge, because `main` moved code out of `render/entry.rs` and a merge that names one file can silently drop work in the file git never looked at. All 16 present.

Deleted tests: none. Renamed: `no_stage_reads_idle` → `no_stage_reads_em_dash_rather_than_guessing_idle` (#4187), named here for `check-deleted-tests`.

## Summary by Sourcery

Close the remaining landable mid-turn steering and TUI defects while aligning transcript attribution, v2 rendering, design-token validation, and supporting documentation.

New Features:
- Add issue-label typeahead suggestions backed by the configured issue provider.
- Add hard wrapping for unbreakable transcript content and enforce hue separation across web design tokens.
- Enable transcript rows to attribute measured file changes to the calls that produced them and render measurement-only sizes.

Bug Fixes:
- Route Esc steering directly into live turns, claim delivered prompts from durable queues, and prevent duplicate dispatches.
- Prevent future file changes from being attributed to successful calls that made no changes.
- Ensure TUI footer hints fit within narrow layouts without clipping into the cost and queue readout.
- Correct v2 status and transcript rendering, including human-readable stage labels, honest read rows, and file-specific subject emphasis.
- Keep diff preambles consistent across multi-file rendering and align Observatory light-mode themes.

Enhancements:
- Split oversized TUI test and render modules while removing unused theme and projection code.
- Align brand assets, palette guidance, and generated token documentation with the v5.0 design system.

Build:
- Add the hue-separation guard to the Makefile and contributor gate commands.
- Remove obsolete file-size baseline entries and complete the design-token migration ledger.

CI:
- Wire the 30-degree OKLCH hue-separation check into gate-parity validation.

Documentation:
- Update TUI v2 specifications, implementation guidance, diagnostics references, and brand documentation for current behavior and palette rules.

Tests:
- Add regression coverage for Esc steering and durable queue claims, call-level diff attribution, transcript wrapping, footer budgeting, issue labels, v2 rendering, and light-mode token parity.

Chores:
- Remove retired v4 brand values and superseded TUI prompt guidance.


macanderson:
Heads-up from a parallel session that was working the same epic today — this PR is `CONFLICTING`/`DIRTY` against `main`, and **the conflicts are not all textual**. Two of them are the shape where resolving in favour of this branch silently reverts something `main` just landed, which is the failure mode a clean-looking merge does not warn about.

Seven PRs merged between 19:49 and 20:20 UTC (#4284, #4285, #4286, #4290, #4291, #4294, #4295). This PR's base is `57aa31a5a`, which predates the last two.

Verified against `origin/main` at `461b7edf2`, not inferred:

## 1. `crates/stella-tui/src/v2/project.rs` — deleted on `main`

```
$ git ls-tree origin/main --name-only crates/stella-tui/src/v2/
crates/stella-tui/src/v2/graph.rs
crates/stella-tui/src/v2/status_bar.rs
crates/stella-tui/src/v2/status_source.rs
crates/stella-tui/src/v2/transcript.rs
crates/stella-tui/src/v2/transcript
crates/stella-tui/src/v2/transcript_source.rs
```

#4294 deleted it. It was the **wrong one of two projections** of SPEC 5's six values: it rendered `StageName::as_str()`, the wire string, so the shipping bar printed `context_recall` and `scope_review` at a human (#4187). `status_source` won because `Status` borrows its strings and a contributed stage's word is lowercased at projection time.

This PR modifies `v2/project.rs`, `v2/status_bar.rs`, `v2/status_source.rs` and `v2.rs`. **Keeping this branch's side of that conflict restores the deleted module and re-creates the two-projection defect** — and the bar goes back to printing the wire string. `render_band` on `main` now calls `StatusSource::project(model, ui).status()`; whatever this branch does to the bar needs to be re-expressed against that.

## 2. `crates/stella-tui/src/deck_render/help_metrics.rs` — `main` already has this feature, at a different path

```
$ git ls-tree origin/main --name-only crates/stella-tui/src/deck_render/
crates/stella-tui/src/deck_render/help.rs
crates/stella-tui/src/deck_render/parked.rs
crates/stella-tui/src/deck_render/tests.rs
```

#4295 closed #4188 by moving the whole `?` overlay to `deck_render/help.rs` and adding SPEC 5's metric block there. This branch adds `deck_render/help_metrics.rs`. Git will not report a conflict between two differently-named new files, so **this is the one that merges clean and leaves two implementations of the same SPEC 5 block** — exactly the shape `check-transcript-surfaces` exists to catch elsewhere. Worth deciding which survives before resolving, rather than after.

## 3. The god-file lists moved, and they are gate-enforced in three places

#4295's move took `deck_render.rs` from 1514 to 1321, at which point `check-file-size.sh` **requires** dropping its baseline entry. So on `main`:

- `scripts/file-size-baseline.txt` has **no** `deck_render.rs` entry (verified: 0 matches)
- `AGENTS.md:692` now reads ``| `stella-tui` | `src/deck_ui.rs`, `src/views/engine.rs` |``
- `crates/stella-tui/README.md`'s list dropped it too

`check-god-files.sh` compares all three. This PR touches `AGENTS.md` and `crates/stella-tui/README.md`, so re-adding `deck_render.rs` to either list fails the gate — and re-adding the baseline entry would be a loosening the ratchet refuses anyway.

## 4. `v2/transcript.rs` / `transcript_source.rs` gained a type

#4291 (#4168) replaced `Event.subject: String` with `Subject { text, is_path }` so a path's basename can carry the emphasis while a `bash` command line does not. `Event::new` takes `impl Into<Subject>` and there is a blanket `From<T: Into<String>>`, so most construction sites are unchanged — but anything reading `event.subject` as a string now needs `.text`.

## 5. Still in flight

#4292 (#4183) is open and touches `crates/stella-tui/src/model.rs` — the `AgentEvent::Stage` arm and the `StepManifest` arm, which this PR also touches. It rebased onto `461b7edf2` cleanly and is green, so it will likely land before this one.

---

Not asking for anything — just making sure the resolution is a decision rather than a default. Happy to help reconcile if it is useful; I have the context on all seven merged PRs.

macanderson:
Went to resolve the conflicts here and found you had already done it (`7aeffab74` + `53c31bf20`) — the resolution itself is right, and I agree with every call in it. But **the merged result did not build**, so I pushed one commit on top (`3ad5b7e7f`, a fast-forward, nothing rewritten).

## The PR read `MERGEABLE` and was broken

Both defects sit in files the other side did not touch, so git reported no conflict and GitHub reported no problem. That is the whole hazard of this particular merge: the five duplicated fixes landed as *differently shaped* code in *different files*, which is the one arrangement a textual merge cannot flag.

**1. `EventKind::Read` lost its `extent` here, and main added a caller for it.**

```
error[E0559]: variant `transcript::EventKind::Read` has no field named `extent`
  --> crates/stella-tui/src/v2/transcript/tests.rs:304:13
```

`6f28d2b94` removed the field for #4180 — nothing produces a read's line count, so the absence should be structural rather than defaulted. #4291 then landed on main, and its witness `a_path_subject_dims_its_directory_and_brightens_its_basename` constructs `EventKind::Read { extent: Extent::default() }`. It only needs *a* read event; the extent was incidental to what it asserts, so I dropped it rather than restoring the field.

**2. `grid.rs` ended up with `encode` below its `#[cfg(test)] mod tests`.**

Both sides added items near the end of that file — this branch's hard-break work (`fe795cccf`) and main's — and the merge interleaved them so the private encoder landed under the test module. `clippy::items_after_test_module` is deny-level under `make lint`, so it is a red gate, not a style note. Moved back above; no code changed.

## Verified on the merge result

Every one with its exit code, `--color=never`:

- `cargo test -p stella-tui` — **0**, every suite green
- `cargo test -p stella-transcript` — **0**
- `cargo test -p stella-cli` — **0**
- `cargo clippy --workspace --all-targets -- -D warnings` — **0**
- `cargo fmt --all -- --check` — **0**
- `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps` — **0**
- `make guards` — **0**

A caveat on `make gate` itself: one run of it died with `linking with cc failed: signal: 15 (SIGTERM)` while two sessions were building into the same target directory. That is a machine failure, not a code one — every tier above was re-run individually and is green — but if CI shows the same shape, retry before reading anything into it.

## Two resolution calls of yours I checked and agree with

- **`stage_word` returning `—` rather than main's `idle`.** SPEC 5's own column in #4187 says `—`, and `idle` is a claim about a turn nothing measured. Your `no_stage_reads_em_dash_rather_than_guessing_idle` replaces main's `no_stage_reads_idle`; worth naming that in the PR body, since `check-deleted-tests` runs on `pull_request` and a renamed test is indistinguishable from a deleted one to it.
- **Consolidating the `?` overlay into one `help.rs`.** Main had moved the whole overlay out of `deck_render.rs` and this branch had added a separate `help_metrics.rs`; folding them is the right shape, and it keeps `deck_render.rs` under the flat limit that let it leave the god-file list.

One thing worth a second look, not a blocker: main's `deck_render_snapshots` gained `the_help_overlay_carries_the_metrics_spec_5_sends_behind_it`, which is redundant with this branch's better-fixtured assertions inside `deck_render_snapshots_pin_the_help_overlay` (yours anchors the prompt cache so WARMTH shows a real countdown instead of an em-dash). If it is still present after the merge, deleting it is right — just name it in the PR body for the same `check-deleted-tests` reason.

Issue #4002, opened 2026-08-19
self-driving: contention is checked before adopting a base fix, but not before claiming an issue


macanderson:
Found by running the loop against this repository (#3942's first live session).

## What happened

`stella self-driving drive` claimed #3691 — the top-ranked P0 — and could not start work on it, because a branch was already there. Two branches matched, and one of them was not the loop's:

```
could not start: #3691 already has a branch from an earlier attempt:
stella/3691-8f6b50dcd483496d
  wip-3691-preserved
```

`wip-3691-preserved` is **six hours old and was created by a different session** — `wip(#3691): Ungraded status — built against the now-deleted stella-pipeline`. Somebody else is working that issue.

The loop routed around it and took the next P0, which is the right outcome. But it got there by accident: it discovered the collision when `git worktree add` failed, not because it looked.

## The gap

`stella_autonomy::contention_verdict` exists, is tested, and weighs exactly the right four signals — ledger claims, open PRs, remote branches, **and worktrees on this machine**. `doctrine.contention` configures how much each weighs.

**It is only consulted on one path**: `step`'s base-breakage adoption (`unblock_base`). Nothing consults it before `LoopStep::Claim` picks an issue, which is the path that matters far more often — a peer working an issue is ordinary, a peer fixing a broken base is rare.

## Why it matters more than it looks

The loop currently discovers contention as a **git failure**, which means:

- It cannot tell "somebody is working this" from "a previous run of mine died here". Those want opposite responses: defer versus clean up and retry.
- It only notices contention that manifests as a branch name collision. A peer working the issue in a differently-named branch, or holding a fleet-ledger claim, is invisible.
- The evidence never reaches the ledger, so a human reviewing why an issue was skipped sees a git error rather than "deferred to a peer".

## What to build

Consult `contention_verdict` in the `Claim` arm of `crates/stella-cli/src/self_driving_cmd/drive.rs`, gathering the four signals for the candidate issue:

| Signal | Source |
|---|---|
| ledger claim | `stella-fleet` ledger / `fleet_claims.rs` — the only authoritative one |
| open PR | `gh pr list --search` for the issue key |
| remote branch | `git ls-remote --heads` |
| local worktree | `git worktree list` |

On `ContentionVerdict::Defer`, skip to the next candidate and record the evidence — the verdict already carries it precisely so a deferral is reviewable rather than silent.

## Constraint

[...]

PR #4308, opened 2026-08-22
fix(stella-cli): claim-time contention, with the loop's own crashed worktree excluded

### LINKED ISSUE #4300: self-driving: a crashed run's leftover worktree now defers its own issue forever, blocking the recovery work::start used to perform
Introduced by the #4002 fix (claim-time contention). Filing it in the same breath as that PR, per AGENTS.md § *Nothing left behind*.

## Problem

`work::start`'s `discard_undelivered_attempt` (`crates/stella-cli/src/self_driving_cmd/work.rs:690`) exists to recover from the loop's **own** crashed run: it asks the forge for an open PR, and finding none, deletes the leftover branch and worktree and retries `manager.create`.

That recovery is now unreachable in one case. A crashed run leaves a registered worktree at `.stella/private/self-driving/stella-<key>-<slug>` whose path contains the issue key. `deliver::contention_for_issue`'s `local_worktrees` probe (`deliver.rs:833`) therefore reports it, and under the default `ContentionPolicy::Defer` the Claim arm defers **before any git state is created** — every pass, since a deferral writes no `spent` entry. The issue becomes unclaimable until a human clears the worktree, whereas before the change the next run repaired it and worked it.

## Why it is ambiguous

At claim time the probe reads only names. The loop's own dead worktree and a live peer loop's worktree are the same string. `work::start` can tell them apart only because by then it holds a specific git error and asks the forge.

## Repro

1. `stella self-driving drive`, kill it mid-turn on issue N.
2. Confirm `git worktree list --porcelain` still lists `.stella/private/self-driving/stella-N-*`.
3. Ensure no open PR names N.
4. Run `drive` again — expect an `Audit::Deferred` line citing `local worktree: ...` instead of the old recovery.

## Constraints

- Do **not** filter `local_worktrees` inside `contention_for_issue`. The base-fix caller (`drive.rs:1082`) depends on seeing another self-driving process's worktree; that is the signal's stated purpose (`crates/stella-autonomy/src/doctrine.rs:179-184`).
- The fix likely belongs at the claim site: make `WORKTREES_DIR` (`work.rs:73`, currently private) `pub(super)` and have the Claim arm's `seen` closure drop worktrees under that directory; or move the forge question earlier.
- Do not duplicate the `.stella/private/self-driving` literal into `drive.rs` — a shared cell written in two places is the failure mode AGENTS.md names for the file-size baseline and the lockfile.

## Note on the existing remedy

`ContentionPolicy::ClaimsOnly` (`doctrine.rs:106-110`) is the operator-facing remedy today and its doc comment already names this hazard. But `ContentionPolicy::Defer`'s own justification — "a deferred one costs a poll interval" (`doctrine.rs:104`) — is **false for this case**: it costs the issue indefinitely. That sentence should be corrected in whichever PR settles this.

## Done when

A crashed run's leftover worktree no longer defers the claim, a live peer's still does, and a witness test pins both.

### LINKED ISSUE #4309: self-driving: the loop reads the authoritative dispatch claim but never mints one, so ClaimsOnly can never defer at the claim site
Noticed while building claim-time contention in #4308 (fix for #4300). Filed rather than fixed there: minting a lease is a write to shared state and a lifecycle question, not a slice of a read-only probe.

## Problem

`contention::for_issue` (`crates/stella-cli/src/self_driving_cmd/contention.rs`) gathers the ledger claim as the one *authoritative* contention signal — a lease with an owner and an expiry, as opposed to a name that resembles the work. `ContentionPolicy::ClaimsOnly` (`crates/stella-autonomy/src/doctrine.rs`) consults nothing else.

**Nothing on the self-driving path ever mints such a claim.** `stella self-driving drive` claims an issue into in-process `state.claimed` and the durable state directory, and never calls `Ledger::claim_dispatch`. So `issue:<n>` rows only ever exist if some *other* dispatcher wrote them.

Two consequences:

1. Two `stella self-driving drive` processes against one clone cannot see each other authoritatively. Under `ContentionPolicy::ClaimsOnly` they see nothing at all and both take the top-ranked issue. That policy is the documented remedy for a solo operator whose remote is full of their own stale branches — exactly the operator most likely to run two loops.
2. Under the default `Defer` they are caught only by heuristics (remote branch, open PR, a worktree outside the loop's own root). #4300 narrowed one of those on purpose, which is correct for a crashed run and makes the missing authoritative signal matter more.

## What to build

Have the `Claim` arm take a dispatch lease under `stella_fleet::dispatch_claim_key`'s `issue:<n>` namespace when it claims, renew it while the issue is being worked, and release it when the issue leaves `state.claimed`.

- Mint: `crates/stella-cli/src/self_driving_cmd/drive.rs`, the `LoopStep::Claim` arm, right where `Audit::Claimed` is recorded.
- Lease API: `crates/stella-fleet/src/ledger/lease.rs` — `claim_dispatch`, `renew`, and the fence semantics. `DISPATCH_LEASE_TTL` is 15 min (`crates/stella-fleet/src/fleet.rs`), so a loop working an issue for longer must heartbeat or the lease lapses under it.
- Read side already exists and needs no change: `contention::claims_naming`.

Open questions the implementer has to settle, not skip:

- **A lapsed lease must not wedge the issue.** The lease module deliberately has no reaper and no `--force` (see its docs); a crashed loop's lease expires on its own after the TTL. Confirm 15 minutes is the right window for an issue-length turn, rather than assuming it.
- **The loop must not defer on its own lease.** Whatever mints it has to be excluded at the claim site the same way #4300 excluded the loop's own worktree, or this reintroduces that bug through the authoritative signal. `owner` embeds the minting process's pid via the fleet run id, which is probably the discriminator.
- **A workspace with no `fleet.db`.** The read side treats a missing ledger as "no claims"; the write side has to decide whether to create one.

## Verify

- Two `stella self-driving drive` processes against one clone, `contention = "claims_only"` in `.stella/`'s doctrine: the second records `Audit::Deferred` citing `ledger claim: issue:<n> held by <run>` instead of taking the same issue.
- `stella fleet claims` lists the row while the loop works it, and it is gone once the issue is delivered or abandoned.
- A killed loop's issue becomes claimable again after the TTL with no human action.

## Done when

The self-driving loop mints, renews and releases a dispatch lease for the issue it is working; `ContentionPolicy::ClaimsOnly` can actually defer at the claim site; a loop never defers on its own lease; and a witness test pins all three.

macanderson:
## What

Excludes worktrees under this verb's own root from the **claim-time** contention signal, and supplies the liveness signal that makes that exclusion sound: the loop now takes a fenced ledger lease on the issue it is working.

## Rebased onto main's rival implementation

`main` landed claim-time contention in #4305 while this was open — `next_claimable`/`Pick` in `drive.rs`, and `deliver::base_fix_contention` renamed `contention_for_issue` and used at **both** sites. It has no own-root exclusion, so **#4300 is live on main today**: a crashed run's leftover worktree defers its own issue on every pass.

Resolved by keeping main's structure and folding this fix into it, not the reverse:

- `next_claimable` keeps main's walk (a deferral advances to the next candidate rather than ending the pass) and gains a `claim` edge, so a lease lost to a peer *between* the probe and the write is one more deferral with the same audit line and counter. `Pick::Take` carries the lease and can no longer derive `Eq` — a held lease owns a heartbeat thread.
- The four reads move out of `deliver.rs` into `contention.rs`, which already had them factored into pure parsers.
- `deliver::live_issue_claims` is superseded by `contention::ledger_claims`, which excludes this process's own claims and formats the owner into the evidence line.

**One thing from the earlier version is dropped**: `MAX_CONTENTION_PROBES`, a five-candidate cap on the per-poll scan. main's `next_claimable` walks the whole ranked queue, and re-litigating that here would widen this PR past the review's finding — so the cap is filed as #4317 rather than smuggled in.

## The two questions

| Caller | Own-root worktrees | Why |
|---|---|---|
| `contention::for_base_fix` | kept (`own_root: None`) | a peer self-driving process is the case the signal exists for — #4300 names not weakening this as a constraint, and unlike the claim site nothing downstream repairs a leftover of the loop's own |
| `contention::for_issue` | dropped (`own_root: Some(..)`) | a leftover there may be this loop's own crashed run, and `work::start`'s `discard_undelivered_attempt` already repairs it — but the exclusion cannot be selective, so it is only sound with the liveness signal below |

Two named questions over one `gather`, so the difference is one argument visible at both call sites rather than two copies of four reads. main's single `contention_for_issue` could not answer both once the claim site needed the exclusion. `WORKTREES_DIR` stays private; `work::worktrees_root()` is the single accessor, so the `.stella/private/self-driving` literal keeps one home.

## The live-peer half, which the first commit did not deliver

Review was right, and the defect was real: **the own-root exclusion cannot be selective.** A live peer self-driving process against the same clone leaves its worktree under the same `.stella/private/self-driving` root, with the same shape and the same issue key. Dropping own-root worktrees dropped the peer along with the wreckage, so #4300's second half — *"a live peer's still does"* — was not delivered, and the test that claimed to cover it used a peer **outside** that root, which is the one case the filter was never ambiguous about.

No filter over paths can separate them, because the question is not *where* but *when*. So it is asked of the signal that carries a time.

`Contention::ledger_claims` was already the authoritative signal, already read by `for_issue`, and **had no producer anywhere in the workspace** — so it was always empty, and `ContentionPolicy::ClaimsOnly` could never defer at the claim site. `self_driving_cmd/claim.rs` is that producer: the loop takes the `issue:<n>` dispatch lease before a turn starts, heartbeats it while the turn runs, and releases it on drop when the turn settles.

| Who left the worktree under the loop's own root | Its lease | Claim time |
|---|---|---|
| this loop's own crashed run | lapsed — nothing renewed it | **proceed**, and `work::start` repairs it |
| a live peer self-driving process | held and beating | **defer** |

Why a lease and not a lock file or a pid: the crash case *is* the case. A lock file has to be cleaned up by the process that died; a pid is reused and means nothing across a reboot. `stella_fleet`'s claim statement evaluates expiry inside the write, so a dead holder's key is reclaimed as a side effect of the next claimant asking for it — no reaper, no daemon, no `--force`. This is the second dispatcher on that table, which is what the `issue:<n>` namespace was reserved for (`fleet.rs`'s `dispatch_claim_key` doc named it before anything claimed it).

It also closes a race the read-only probe structurally cannot: two loops polling one clone can both read "free" before either writes. `claim_dispatch` is one conditional statement and the affected-row count decides the winner, so the loser is told who holds it rather than left to duplicate the turn.

`issue_claim_key` gives the namespace one home — the reader and the producer must agree exactly, or the probe silently stops seeing peers, which is the shared-cell failure AGENTS.md names for the lockfile and the file-size baseline.

**Bounded cost, stated plainly:** after a hard kill the issue looks taken for up to `LEASE_TTL` (5 minutes) before the loop's own recovery becomes reachable — not the rest of the run, which is what #4300 filed. The TTL is deliberately shorter than `stella_fleet`'s 15-minute task lease, because here the loser is the loop's own next pass rather than a whole dispatched task.

## Evidence

`make gate` exited 0 on the pre-merge tree (guards, fmt, clippy `-D warnings`, rustdoc, `--workspace` tests — 183 test binaries — self-driving harness 61/61).

The merge resolution onto #4305 was **not** gated locally — the machine could not take another workspace build — so CI is the evidence for those commits, and it took two red rounds to get there, both stated rather than squashed away:

1. `super::claim::Claim` inside `mod tests` does not resolve — in that module `super` is `drive`, not `self_driving_cmd`. Test-only, which is why `cargo build` and the MSRV `cargo check` both passed it: neither compiles `#[cfg(test)]` code.
2. A rustfmt diff on one line. `cargo fmt --check` runs *ahead* of the test compile, so the second red job never reached the tests at all.

All eleven checks now pass on `e6e2c2a`, `fmt + clippy + test` included.

Witness — `contention::tests::a_live_peer_in_the_same_root_defers_where_a_crashed_run_proceeds` asserts **both** halves against one indistinguishable worktree path under the loop's own root, differing only in whether a lease is live: crashed → `Proceed`, live peer → `Defer`, and dropping the peer's lease returns it to `Proceed`. `claims_only_defers_on_a_live_peer_and_on_nothing_else` pins the policy that consults nothing else.

Proven non-vacuous by ablation — removing the producer, which is exactly this PR's previous state: six tests fail, and the failure in `claims_only_defers_on_a_live_peer_and_on_nothing_else` is on the verdict itself,

```
assertion failed: matches!(contention_verdict(ContentionPolicy::ClaimsOnly, &claimed),
    ContentionVerdict::Defer { .. })
```

i.e. a live peer's claim reading as `Proceed` — the defect the review named.

`docs/spec/backlog-self-driving.md` said two self-driving processes "see each other **only**" in the worktree signal. That sentence was true and was the defect; it is corrected in the same PR, with the argument for why the loop takes a claim rather than only reading one.

The worktree-exclusion half is likewise proven non-vacuous: replacing the `own_root` filter with a no-op flips `the_verdict_proceeds_on_a_crashed_run_and_defers_on_a_peer` from `Proceed` to

```
Defer { evidence: ["local worktree: /repo/.stella/private/self-driving/stella-4300-8f6b50dc"] }
```

which is the `Audit::Deferred` line #4300's repro predicts. Three tests fail under the ablation, five stay green.

## Bounded cost

`MAX_CONTENTION_PROBES = 5`. A deferral deliberately writes no `spent` entry (a peer finishing must let the loop take the issue later), so an unbounded scan would re-ask the forge about a whole contended backlog every poll. Hitting the cap records its own audit line — a truncated scan that reported nothing would read as an empty queue.

## Also

`ContentionPolicy::Defer`'s doc claimed "a deferred one costs a poll interval". #4300 asks for that correction and it is here: the sentence is true only while the evidence can clear on its own, which is a claim about the evidence, not about deferring. It now also says what makes a live peer keep deferring once own-root worktrees stop counting — the lease, the one signal here that carries a time.

## Not done here

The branch-collision guard in `work::start` is untouched, per #4002's constraint. #4002 is `Refs`, not `Closes` — its "done when" includes the loop skipping an issue a peer is working *observed live*, which this cannot demonstrate from a test.

The `acquire` side has the same latent shape the read side just closed: if a key this process already holds were ever probed again, `claim_dispatch` would lose to itself and report `HeldBy(us)`. `drive`'s candidate filter makes that unreachable, and handling an unreachable branch is the unwired code AGENTS.md forbids — noted here rather than coded, because the filter is what keeps it true.

#4309 was filed from this PR's own first pass — "the loop reads the authoritative dispatch claim but never mints one" — and is what the second commit builds, including its third requirement that the loop never defer on its own lease (`claims_naming`'s `own_owner`, the owner-side mirror of `own_root`).

Closes #4300
Closes #4309
Refs #4002

## Summary by Sourcery

Prevent claim-time contention from wedging the loop on crashed worktrees while retaining live-peer protection through expiring ledger leases.

New Features:
- Add claim-time contention checks for ranked issues using remote branches, open pull requests, local worktrees, and live ledger claims.
- Add fenced, heartbeat-renewed issue leases so self-driving peers remain visible while crashed runs eventually become reclaimable.

Bug Fixes:
- Prevent the loop from deferring indefinitely on its own crashed-run worktrees while preserving deferral for live peers.
- Make ClaimsOnly contention policy effective at the claim site.

Enhancements:
- Share contention signal parsing between claim-time checks and base-fix contention while applying caller-specific worktree handling.
- Bound claim-time contention probing to five candidates per pass and record truncated scans in the audit trail.
- Clarify that deferral costs a poll interval only when its evidence can clear independently.

Documentation:
- Update the self-driving backlog specification to document lease-backed liveness and claim-time worktree exclusion.

Tests:
- Add coverage for own-root worktree exclusion, peer contention, live ledger claims, lease release, claim races, policy behavior, and signal parsing.

Chores:
- Expose a shared issue claim-key helper and centralize the self-driving worktree root accessor.


macanderson:
Follow-up filed: #4309 — the loop reads the authoritative dispatch claim this PR wires in, but nothing on the self-driving path ever mints one, so `ContentionPolicy::ClaimsOnly` can never defer at the claim site. Out of scope here: minting a lease is a write to shared state and a lifecycle question, not a slice of a read-only probe.

macanderson:
Both review findings were right, and they were the same finding: the own-root exclusion cannot be selective, so it dropped a live peer's worktree along with the loop's own wreckage, and the test that claimed to cover the peer case used a peer *outside* that root — the one case the filter was never ambiguous about.

No filter over paths can separate them, because the question is not *where* but *when*.

`Contention::ledger_claims` was already the authoritative signal, already read by `for_issue`, and had no producer anywhere in the workspace — so it was always empty (which is also why `ContentionPolicy::ClaimsOnly` could never defer at the claim site; that was filed as #4309 from this PR's own first pass). `self_driving_cmd/claim.rs` is that producer: the loop takes the `issue:<n>` dispatch lease before a turn starts, heartbeats it while the turn runs, releases it on drop when the turn settles. A crashed run's lease lapses with no reaper; a live peer's does not.

Witness — `contention::tests::a_live_peer_in_the_same_root_defers_where_a_crashed_run_proceeds` asserts both halves against **one indistinguishable worktree path under the loop's own root**, differing only in whether a lease is live.

Ablated by removing the producer, which is this PR's previous state — six tests fail, and one fails on the verdict itself:

```
assertion failed: matches!(contention_verdict(ContentionPolicy::ClaimsOnly, &claimed),
    ContentionVerdict::Defer { .. })
```

i.e. a live peer's claim reading as `Proceed`. That is the defect, reproduced as a red test.

Three things stated rather than hidden:

- After a hard kill the issue looks taken for up to `LEASE_TTL` (5 min) before the loop's own recovery is reachable — bounded and self-clearing, not the rest of the run, which is what #4300 filed.
- The heartbeat itself has no witness: every test completes inside the TTL, so all of them pass with the beat thread deleted. Filed as #4314 with the ablation that would catch it.
- `docs/spec/backlog-self-driving.md` said two self-driving processes "see each other **only**" in the worktree signal. True when written, and it was the defect — corrected here, along with `AGENTS.md` and the fleet README, which both described `fleet.db`'s claim table as the fleet's alone.

`make gate` exits 0.

macanderson:
Merged `origin/main` in — the branch was CONFLICTING because #4305 landed a rival claim-time contention implementation while this was open (`next_claimable`/`Pick`, and `base_fix_contention` renamed `contention_for_issue` and used at both sites).

Resolved by keeping main's structure and folding this fix into it, not the reverse. main's version has no own-root exclusion, so **#4300 is live on main today**. `next_claimable` keeps its walk and gains a `claim` edge, so a lease lost to a peer between the probe and the write is one more deferral rather than the end of the pass; the four reads consolidate into `contention.rs` as `for_issue` (claim time, exclusions) and `for_base_fix` (nothing excluded) over one `gather`.

Two things stated rather than hidden: `MAX_CONTENTION_PROBES` from the superseded revision is **dropped**, not merged in — capping main's fresh walk is a separate decision, filed as #4317. And the merge commit was not gated locally (machine constraint), so CI is the check on it; `cargo check` on the declared MSRV has since passed, and the full `ci` job is what settles it.

No tests were deleted: main's two `next_claimable` tests are kept and adapted to the new edge, and one is added for the lost-lease race.

Issue #3968, opened 2026-08-19
brand v4.0 is half-applied: product surfaces are gold, website + every generated asset is still Ion cyan


macanderson:
## Problem

The brand hue was taken back to gold in the v4.0 change on branch
`worktree-observatory-brand-and-reasoning` (see PR for that branch), but that
change was deliberately scoped to **the product surfaces plus the single
brand-ramp definition**. Every marketing/web surface and every generated
binary asset is still Ion cyan, so the tree currently ships two identities:

- **gold** — `docs/brand/css/tokens.css`, `crates/stella-observatory/src/assets/`
  (`index.html`, `mark.svg`, `wordmark.svg`, `wordmark-light.svg`),
  `crates/stella-cli/src/export.rs`, `crates/stella-mcp/src/oauth/callback_page.css`,
  `crates/stella-transcript/src/html/transcript.css`,
  `arenabench/ui/app/globals.css`, and `crates/stella-tui/src/palette.rs`
  (which never left gold — `BRAND` has been `#FFB81A` throughout).
- **cyan** — everything under `website/`, the generated cuts under
  `docs/brand/{logo,site,social,spinners,pwa,wallpapers}/`,
  `docs/brand/brand-guidelines.html`, `docs/papers/stella-foundry.html`,
  `website/public/presentations/investor-deck.html`, `website/public/vision.html`,
  and every favicon / icon / wallpaper PNG.

This is the same split #2594 already names (docs site, marketing site and decks
sit outside `design_token_parity.rs`), except it is now visible rather than
theoretical: a reader crossing from the website to the Observatory watches the
brand change hue.

## Why this was not done in the same change

The generated assets are the blocker, not the CSS. Recolouring them means
re-running the kit's own generators — `docs/brand/cometkit.py`,
`docs/brand/build_marks.py`, `docs/brand/wallpapers/build_wallpapers.py` —
which rewrite ~90 binary files (favicons, maskable icons, spinner GIFs, 4k/5k/6k
desktop and phone wallpapers). The reversal being applied is commit
`d6ba4b7c8` ("feat(brand): recolour the whole system — Ion on Obsidian (brand
kit v3.0) (#3658)"), which touched 158 files; the product half is ~13 of them.

## What to do

1. Repoint the remaining CSS/TSX token copies to the gold ramp now in
   `docs/brand/css/tokens.css`. `website/src/app/tokens.css` is the main one;
   `website/src/lib/brand-parity.test.ts` is the guard that will tell you what
[...]

PR #4570, opened 2026-08-23
fix(stella-protocol,stella-cli,brand): claims match reality — god-file comments, the unwired lifecycle channel, a generated brand mirror

### LINKED ISSUE #3135: Unwired: `context_event.rs` ships 524 lines of lifecycle-event contract with no emitter and no tracking issue
## Problem

`crates/stella-protocol/src/context_event.rs` is 524 lines of published wire
contract that nothing in the workspace constructs or reads. The module doc says
so itself (line 20):

> **Status: schema landed, channel not yet wired.** Nothing in this workspace
> constructs or reads a `LifecycleEventEnvelope` today — the types and their
> golden JCS vectors exist so the wire shape (and therefore the `record_hash`
> preimages built from it) is pinned before the first emitter lands.

Verified: zero references to `LifecycleEventEnvelope`, `LifecycleEvent`, or
`context_event::` anywhere outside the defining crate.

The decision to land the schema ahead of the emitter is defensible and the
rationale in the module doc is good. What is missing is the tracking issue.
AGENTS.md § "Nothing left behind" is explicit:

> If your change ships scaffolding that something else must later wire up, file
> the issue for that wiring in the same breath as the PR — unwired code with no
> tracking issue is exactly the failure mode this rule exists to prevent.

The file names no issue number, and no open issue covers the wiring (#469,
"Epic: adaptive context lifecycle", is closed). `check-left-behind` does not
catch this: the guard looks for bare `TODO`/`FIXME`/`XXX`/`HACK` markers, and
this module carries none — the prose is careful enough that it reads as a
decision rather than a deferral, which is precisely why it slipped the net.

This issue exists so the wiring is tracked. It is **not** an argument for
deleting the module.

## Files

- `crates/stella-protocol/src/context_event.rs` — the types, the
  `LifecycleEventEnvelope` versioned envelope, `LifecycleEvent::Unknown`
  fallback, and the golden JCS vectors.
- `docs/spec/adaptive-context/` — the spec the module implements.
- `doc:adr/0006-contextframe-vs-compiledcontextframe` — the cited ADR for
  `CompiledContextFrameBuilt`.

## Verify the gap

```bash
rg -n "LifecycleEventEnvelope|LifecycleEvent|context_event::" --glob '*.rs' \
  crates/ --glob '!crates/stella-protocol/**'
# expected today: no matches
```

## Definition of done

Either of these closes it — the point is that a human decides, on the record:

1. **Wire the first emitter.** A producer constructs `LifecycleEventEnvelope`
   for at least one `LifecycleEvent` variant, a consumer reads it, and a
   witness test fails without the emitter and passes with it. The golden JCS
   vectors already pin the wire shape, so the witness is about the channel, not
   the bytes.
2. **Or record the deferral where a guard can see it.** Keep the module, and
   reference this issue number from the module doc so the unwired status is
   discoverable from the code rather than only from reading 524 lines.

- `stella-protocol` is types-only: zero logic, zero I/O (AGENTS.md invariant
  set, workspace layout table). An emitter belongs in a crate that may do I/O.
- `stella-core` depends on `stella-protocol`, so these events deliberately
  carry stable ids as strings rather than importing `stella-core` record
  structs — importing them here would be a dependency cycle. Any wiring must
  preserve that direction.
- The module doc's two stated reasons for staying off `AgentEvent` (explicit
  `schema_version`; keeping internal adaptive-context records out of the public
  `stream-json` / serve SSE vocabulary) are design constraints on the wiring,
  not incidental notes.
- Editing *type* doc comments in this file re-exports into `docs/wire/` and
  reddens `wire-schema` until the generated artifacts are regenerated and
  committed. Module docs are not exported — which is why the citation lives at
  module level. Prefer module-level prose for anything documentation-only.

## Context

Found during a first-pass adversarial audit of the workspace. The audit's one
code fix is #3137 (telemetry persistence off the reactor); this finding was not
fixed there because wiring an emitter is a design decision, not a cleanup.
Sibling finding of the same shape at a smaller scale: #3136.

### LINKED ISSUE #3941: three design docs still describe the six-role config: stella.next.toml, agent_system.toml, docs/prompts
#3908 (slice 4 of #3903) collapsed the engine settings to one role and migrated
every document that describes the **current** config surface. Three documents
that describe a *design* rather than the shipped surface were deliberately left
alone, because updating them is a design decision rather than a key rename.

### 1. `docs/spec/config-system/stella.next.toml`

Declares itself a target, not a plan of record:

> The config shape AFTER the five features in DESIGN.md are built. This is a
> TARGET, not a plan of record

It still carries `pipeline_worker_model`, `pipeline_verifier_model`,
`pipeline_triage_model`, `pipeline_research_model`, `pipeline_plan_model`,
`[agents.verifier]` and `[agents.verifier.tools]`. The interesting part is
`[agents.verifier.tools]` — a per-role tool set. Under `doc:roleless-core` that
concept does not disappear, it **moves**: a per-*seat* tool grant is a coherent
thing for a plugin's participant to have, and it is a better fit for the socket
than a per-role one ever was. Someone has to decide whether that block becomes
`[seats."<plugin-id>/<role>"]` with a tools key, or is dropped.

### 2. `docs/spec/config-system/agent_system.toml`

An illustrative example whose stated subject is:

> per-agent configs (triage / worker / verifier)

The whole document is a worked example of the three-role split. It either
becomes a worked example of *one role plus plugin seats*, or it is retired as
a historical artifact of the staged-pipeline era.

### 3. `docs/prompts/` and `docs/design/pipeline-journey.md`

`docs/prompts/{worker,verifier,triage,research,plan,witness-author,witness-repair,verdict,plan-repair}.md`
are the staged pipeline's role prompts, and `pipeline-journey.md` is its design
narrative. Both describe a subsystem deleted in #3865.

These are arguably fine as history — `docs/design/` is explicitly
work-in-flight, and the prompts are the reference a verification plugin author
would want when porting the shape into Vera (`doc:pipeline-as-plugins` §8). But
`docs/prompts/README.md` presents them as the live prompt set, which is what
makes them misleading rather than archival. At minimum the README needs a
header saying what they are now.

## Not in scope here

`docs/spec/config-system/stella.today.toml` and `DESIGN.md` **were** migrated by
#3908, along with `README.md` and every page under
`website/content/docs/` that describes the shipped surface. This issue is only
the design/illustrative tier.

## Definition of done

Each of the three either describes the one-role + seats world, or carries a
header stating plainly that it documents the pre-#3865 staged pipeline and is
kept for reference. No document reads as current guidance for a key the engine
no longer has.

Verify with:

```bash
rg -n 'pipeline_(verifier|worker|triage|research|plan)_model|\[agents\.(verifier|worker|triage|research|plan)\]' docs/
```

Every surviving hit should be inside a clearly-marked historical section or a
retirement note.

Related: #3903 (epic), #3908 (the slice), #3865 (the pipeline removal),
`doc:pipeline-as-plugins` §8.

### LINKED ISSUE #3983: brand assets are mirrored into website/ by hand — the parity test catches drift but nothing performs the sync
## Problem

The website's brand assets are **byte-copies** of `docs/brand/`, and the copy
is performed **by hand**. `website/src/lib/brand-parity.test.ts` asserts they
match (13 logo SVGs, 8 PWA icons, `src/app/icon.svg`, and — since #3968 — the
pixels inside `src/app/favicon.ico`), but nothing *performs* the sync. Every
recolour therefore ends with a human running a dozen `cp` commands correctly,
and the guard's only job is to catch it when they do not.

They did not, twice:

- **v2.0 → the site.** `src/app/tokens.css` sat on v1.0 phosphor gold while the
  kit had moved to bronze; thirteen SVGs and seven of eight PWA icons were
  stale with it. This is what motivated the parity test in the first place.
- **v3.0 → `src/app/favicon.ico`.** It was never recoloured at all, and sat on
  v2.0's bronze-on-warm-ink art (ground `#10100F`) straight through the Ion
  recolour. The test asserted only the PNG *colour type*, never the pixels, so
  it reported green for the entire time. #3968 strengthened the assertion and
  regenerated the file.

A guard that detects drift is the second line of defence. The first — a way to
produce the copy that cannot be done half-way — does not exist.

## What to do

Add a sync step that is the single way those files are produced, and make the
parity test the check on it rather than the mechanism:

1. A script (`docs/brand/sync_site.py`, or a `make brand-sync` target) that
   copies the kit's outputs to their site paths, including the two renames the
   site applies (`icon-maskable-{192,512}.png` → `maskable-{192,512}.png`) and
   the RGBA re-encode of `favicon.ico`.
2. The re-encode must stay: `src/app/favicon.ico` is deliberately **not** a
   byte-copy, because Next's `ico` decoder accepts only RGBA and fails the
   production build on the kit's RGB PNGs. The reasoning is already written
   down in the test. #3968 used a dependency-free encoder (PNG unfilter →
   append opaque alpha → recompress, `zlib` only), matching `cometkit.py`'s
   "PNG needs only zlib and four chunk headers, no imaging library" approach —
   that is worth folding into the kit rather than living in a scratch dir.
3. Ideally `build_marks.py` gains a `--sync-site` flag so regenerating and
   mirroring are one action, and the pairing cannot drift.

## Files

- `docs/brand/build_marks.py`, `docs/brand/cometkit.py` — the generators
- `website/public/brand/`, `website/public/icons/`, `website/src/app/icon.svg`,
  `website/src/app/favicon.ico` — the destinations
- `website/src/lib/brand-parity.test.ts` — the PAIRS list is the authoritative
  mapping to implement, including the renames and the favicon exception

## Verify

Delete a mirrored asset, run the sync, and confirm `pnpm test` (in `website/`)
goes red → green. Note that `pnpm test` now runs in CI via `docs.yml` (#3968);
before that it ran nowhere.

## Definition of done

No hand-copying step is required to ship a recolour: regenerate the kit, run
one command, and `pnpm test` passes.

### LINKED ISSUE #4528: ~10 stella-cli modules claim agent.rs is 'a grandfathered god file closed to growth' — it is not in the baseline; the claim is false
## Problem

`agent/output.rs:177`, `agent/summary.rs:6`, `agent/budget.rs:7`, `agent/persistence/turn_door.rs:35`, `wrapper_plugin.rs:29`, `main.rs:145`, `fleet_claims.rs:21`, `turn_facts.rs`, `turn_files.rs` (and possibly more — `rg -l 'grandfathered god file' crates/stella-cli`) say `agent.rs` is a grandfathered god file closed to growth. `scripts/file-size-baseline.txt` lists only `command_deck.rs` for this crate, AGENTS.md's table agrees, and `agent.rs` is ~1351 lines against the 1500 default. The intent (keep it under the line) is sound; the words are a false claim, which AGENTS.md treats as a bug. Found in PR #4524.

## Done means

One sweep rewording each site to "close to the 1500-line ratchet — put new logic in a sibling" (or pointing at the real constraint), `make prose` green, no behaviour change.

Related: #4524, #3923.

macanderson:
## What

Four claims-vs-reality issues, one branch.

### #4528 — false "grandfathered god file" claims
`scripts/file-size-baseline.txt` is the only source of truth for which files are grandfathered, and sixteen comments called files grandfathered that are not in it: `agent.rs` (entry removed with #3852), `agent/tests.rs` (#3244), `fleet_cmd.rs` (#2814), `self_driving_cmd.rs` (claimed "over the 1500-line limit" at 1422 lines), and — same defect class outside the issue's crate — `deck_render.rs` (#4295) and `views/session.rs` (#4210) in stella-tui. Each site now states the real constraint (close to / stays under the 1500-line ratchet, new logic lands in a sibling) or tells the story in the past tense where the split itself retired the baseline entry. `docs/spec/turn-lane-assembly.md` also loses a copied line count that had drifted (1917 vs 1677 today). Claims about `command_deck.rs`, `driver.rs`, `bus.rs`, `openai.rs`, `anthropic/tests.rs`, `zai/tests.rs`, stella-store's `lib.rs`/`usage.rs` and `deck_ui.rs` are true and untouched.

### #3135 — the never-wired `LifecycleEventEnvelope` channel is deleted (option 1 of the issue's definition of done)
`crates/stella-protocol/src/context_event.rs` shipped a versioned envelope, nine payload structs, `event_type` tokens and `decode()` that nothing in the workspace ever constructed or read (`rg 'LifecycleEventEnvelope|context_event::' --glob '*.rs' crates/` matched only the defining crate and its re-exports). The repo bans unwired features, so the tracking issue resolves by deletion rather than by keeping a published contract with no traffic. What survives is the one type production constructs — `CompiledContextFrameBuilt` (built at `crates/stella-core/src/receipts.rs`, riding the step manifest on `AgentEvent`, present in `docs/wire/`) — and its golden JCS vector, since those canonical bytes are the `record_hash` preimage. The module doc records the decision and names this issue; `lib.rs`, the crate README, `context_record.rs` and ADR 0006 no longer describe the envelope as the second event channel.

**Deleted tests** (all inline in `context_event.rs`, covering only the deleted types — deliberate removals, named here for `check-deleted-tests`):
- `a_recognized_event_decodes_to_its_typed_form`
- `a_future_event_type_round_trips_without_error`
- `a_recognized_type_with_a_bad_payload_degrades_to_unknown_not_error`
- `golden_jcs_vectors_for_every_event_body` — its `CompiledContextFrameBuilt` vector survives, renamed `golden_jcs_vector_for_compiled_context_frame_built`
- `golden_jcs_vector_for_the_envelope`
- `the_event_type_tokens_are_canonical`
- `envelope_omits_absent_identity_fields`

### #3983 — the website brand mirror is now generated
`website/src/lib/brand-parity.test.ts` caught drift but nothing performed the copy; every recolour ended in hand-run `cp`, twice incorrectly. Now:
- `docs/brand/sync_site.py` mirrors the 13 logo SVGs (removing strays — the test asserts set equality), the 8 PWA icons with the site's two maskable renames, `src/app/icon.svg`, and `src/app/favicon.ico` as the kit's pixels re-encoded RGBA (Next's `ico` decoder rejects the kit's RGB PNGs). Byte copies skip identical files; the ICO is compared by pixels, so a different zlib cannot churn the tree.
- The RGBA encoder and ICO packer are folded into `cometkit.py` (`png_as_rgba`, `ico_as_rgba`, `pack_ico`; `build_marks.write_ico` now delegates), retiring #3968's scratch encoder, and the PNG unfilter loop is one shared `_unfiltered_scanlines` used by `png_opaque_colours` and the new decoder.
- `make brand-sync` runs the sync; `build_marks.py --sync-site` regenerates and mirrors in one action (`--check --sync-site` is refused: `--check` writes nothing). The parity test's remedy strings now name the sync instead of describing the hand copy; `docs/brand/README.md` documents it.

### #3941 — already resolved on main by #4546
The three documents the issue names were migrated after the issue was triaged: `agent_system.toml` carries a file-level HISTORICAL header, `stella.next.toml` rewrites the five role keys as pre-#3865 history with the `[agents.verifier.tools]` question recorded as an explicit OPEN QUESTION, and `docs/prompts/README.md` opens with "Half of this set is history" while the staged-pipeline pages carry `status: archived` frontmatter. The issue's own verify (`rg -n 'pipeline_(verifier|worker|triage|research|plan)_model|\[agents\.(verifier|worker|triage|research|plan)\]' docs/`) now returns only marked-historical hits plus one genuinely current statement (the bench harness really does still write `pipeline_verifier_model` — `bench/harbor_adapter/stella_harbor/posture.py:851`). No edit needed; closing on that evidence.

## Evidence

- `cargo test -p stella-protocol` — green (exit 0), including `golden_jcs_vector_for_compiled_context_frame_built` and the 8 wire-contract tests (so `docs/wire/` is unaffected by the trim; `make wire-schema` also green).
- `cargo test -p stella-core receipts` — green (exit 0): the `CompiledContextFrameBuilt` producer path still compiles and passes.
- `cargo clippy -p stella-protocol --all-targets -- -D warnings` and `cargo clippy -p stella-core --all-targets -- -D warnings` — green.
- `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps` (plain AND `--document-private-items`) for stella-protocol, stella-core, stella-cli, stella-tui, stella-model — all green.
- `make guards-fast`, `make prose` (654 grandfathered constructions, none added), `cargo fmt --all` — green.
- **#3983 witness, run both ways:** deleting `public/icons/maskable-192.png`, `public/brand/logomark-color.svg` and `src/app/favicon.ico` turned `node --test src/lib/brand-parity.test.ts` red (3 failures); `python3 docs/brand/sync_site.py` restored all three and the suite went 8/8 green; a second sync run writes nothing (pixel-level idempotence), and the regenerated `favicon.ico` is byte-identical to the committed one. `python3 docs/brand/build_marks.py --check` stays green over the refactored decode path.
- **#3135:** deletion has no assertion-level witness by construction — the deleted channel had zero consumers to fail. The compile-level witness is that the workspace builds without the types (clippy/doc/test runs above); the surviving golden vector pins the wired type's bytes.
- **#4528 / #3941:** comments and docs only, no witness; the check is the baseline itself (`rg 'grandfathered god file'` against `scripts/file-size-baseline.txt`) and the issue's own verify rg respectively.
- Not run here: `build_marks.py --sync-site` end-to-end (it starts with a full re-render of all 52 exports, whose bytes shift with the librsvg release — the same reason `--check` avoids byte comparison); the flag's sync half is the same `sync_site.main()` witnessed above.

## Not done

Nothing skipped.

Closes #4528
Closes #3135
Closes #3983
Closes #3941

## Summary by Sourcery

Align protocol, documentation, and brand assets with the repository's actual wired behavior and automate website brand synchronization.

New Features:
- Add a generated brand synchronization workflow that mirrors the brand kit into the website and supports both standalone and combined regeneration commands.

Bug Fixes:
- Remove the unused LifecycleEventEnvelope channel and update protocol documentation to reflect the single wired event path.
- Correct inaccurate god-file, baseline, and line-count documentation so constraints and historical splits match the repository state.
- Ensure website favicon synchronization preserves the brand pixels in the RGBA format required by the website build.

Enhancements:
- Consolidate PNG decoding, RGBA normalization, ICO packing, and pixel comparison support for reusable brand asset processing.
- Retain and update the canonical JCS coverage for the production-used CompiledContextFrameBuilt event.

Build:
- Add the `brand-sync` Make target and `build_marks.py --sync-site` option, with incompatible check-and-write usage rejected.

Documentation:
- Document the generated website brand mirror and update protocol, ADR, specification, and source comments to remove obsolete lifecycle-channel and grandfathered-file claims.

Tests:
- Update brand parity remediation guidance to use the synchronization workflow and preserve the surviving wire-contract golden vector.


macanderson:
**Answering Sourcery's ❌ on the #4528 row** ("the diff does not update `agent/summary.rs`, which the issue explicitly identifies as containing the false claim"):

No edit to `agent/summary.rs` was needed in this diff, because the false claim it carried was already corrected on `main` before this PR — by commit c5bd49968 (#4532), which reworded lines 6–10 from "`agent.rs`, which **is** a grandfathered god file closed to growth" to the accurate past tense: "**was** a grandfathered god file … has since fallen back under the ratchet and its baseline entry was removed with #3852". That commit is an ancestor of this branch, so the PR head tree already contained the corrected wording; a diff hunk here would have had nothing to change. The issue's own verify confirms it on the merged tree:

- `git grep -n 'grandfathered god file' origin/main -- crates/stella-cli` returns five sites:
  - `agent/summary.rs:6` — past tense, accurate (the #4532 wording above);
  - `command_deck/add_dir.rs:110`, `command_deck/init_cmd.rs:7`, `engine_config.rs:463` — all about `command_deck.rs`, which is genuinely in `scripts/file-size-baseline.txt` (ceiling 3915, the crate's only entry);
  - `test_env.rs:13` — the negated claim ("`main.rs` is **not** a grandfathered god file and must not become one"), accurate.

I also re-checked every other present-tense `grandfathered god file` claim on the merged tree against the baseline (stella-core's `driver.rs`/`bus.rs`, stella-model's `openai.rs`/`anthropic/tests.rs`/`zai/tests.rs` — the latter really is exactly on its 1861-line ceiling — stella-store's `lib.rs`/`usage.rs`, stella-tui's `deck_ui.rs`/`views/engine.rs`, and `docs/spec/turn-lane-assembly.md`): all name files the baseline actually lists. The sweep #4528 asked for is complete on `main`.

Issue #3879, opened 2026-08-19
arenabench's accepted engine-config keys have drifted from the Rust source of truth, and one key is emitted but always refused at launch


macanderson:
## Problem

`arenabench/arenabench/harbor_agent.py` builds a posture, guards it against a
hand-maintained allow-list, and hands it to Stella as
`STELLA_ENGINE_CONFIG_JSON`:

```python
unknown = set(posture) - _ENGINE_ROOT_FIELDS
if unknown:  # pragma: no cover - guards a future edit, not a runtime path
    raise ValueError(...)
```

That allow-list has drifted from the Rust authority
(`crates/stella-cli/src/settings/unknown.rs::ENGINE_ROOT_FIELDS`) **in both
directions**, and one direction is a live launch failure.

### Extra: `responsibilities` is emitted and is not in the Rust vocabulary

`_ENGINE_ROOT_FIELDS` lists `responsibilities`, and `arena_posture` writes it:

```python
if rows:
    posture["responsibilities"] = rows
```

`ENGINE_ROOT_FIELDS` on the Rust side does **not** contain `responsibilities`.
`config::trusted_engine_config_shape_is_strict` fails **closed** on an unknown
root key, so any arena seat that declares a responsibility override is refused
at launch with `STELLA_ENGINE_CONFIG_JSON is invalid; refusing to start with a
partial engine override` — the local guard passes it through precisely because
the local copy is wrong.

#3871 called this out as pre-existing and explicitly scoped it out ("it was
*already* being rejected by the strict seam before this PR and is out of scope
here"). It is still true, and nothing tracks it. This issue is that tracking.

### Missing: four keys the Rust side accepts

`_ENGINE_ROOT_FIELDS` lacks `model_output_caps`, `model_timeout_secs`,
`compaction_budget_tokens`, and `approval_wait_secs`. This direction is only
restrictive — an arena posture cannot select them without tripping the local
guard — so it is a capability gap, not a broken run. Worth closing in the same
pass. (Note `approval_wait_secs` is itself under review in #3870; do not add it
without checking that outcome.)

## Root cause: it is a hand-copy, and the file says copies drift

`crates/stella-cli/src/settings/unknown.rs` states the rule directly:

> two hand-maintained copies of the same field list would drift the moment a
> knob is added, and the launcher seam fails *closed* on a name it does not
> know — so a drifted copy there is a refused benchmark run.

The sibling tree already solved this. `bench/harbor_adapter/tests/test_posture.py`
[...]

Issue #3871, opened 2026-08-19
bench: pipeline_max_revisions/pipeline_candidates/pipeline_require_diff_coverage/pipeline_verifier_evidence_demand now hard-refused by STELLA_ENGINE_CONFIG_JSON, not silently ignored


macanderson:
## Problem

Four `agent_engine_config` fields — `pipeline_max_revisions`, `pipeline_candidates`,
`pipeline_require_diff_coverage`, `pipeline_verifier_evidence_demand` — were
removed end to end from `AgentEngineConfig` in `crates/stella-cli/src/settings/`
(`engine.rs`, `toml_config.rs`, `unknown.rs`, plus `config/tests.rs` and
`settings/tests/toml.rs`). An adversarial audit found **zero remaining
behavioral consumers**: the staged pipeline that read them (`stella-pipeline`)
was deleted in #3865, and the four fields had become inert — a deployment
setting one silently no-opped, which `CLAUDE.md`'s "measure honestly" rule
calls a hazard for exactly the benchmark-arm knobs these were.

The Rust-side removal deletes the four names from
`crate::settings::ENGINE_ROOT_FIELDS`. That allow-list is shared verbatim by
`stella-cli`'s trusted-launcher strictness gate
(`config::trusted_engine_config_shape_is_strict`, used to validate
`STELLA_ENGINE_CONFIG_JSON`), and that gate **fails closed** on an unknown
root key. So the behavior change for anything that still emits these keys into
`STELLA_ENGINE_CONFIG_JSON` is not "silently ignored" (the pre-existing bug) —
it is now **refuses to launch at all**. See
`crates/stella-cli/src/config/tests.rs::the_dead_attempt_count_posture_is_refused_by_the_trusted_launcher_seam`
for the witness of the new refusal.

Several Python bench trees still write these four keys into a posture that
becomes `STELLA_ENGINE_CONFIG_JSON`, following the earlier work in #2600
(closed as "completed" — that issue wired `pipeline_max_revisions` /
`pipeline_candidates` end-to-end through ArenaBench's `Engine` model, web form,
and posture emission on the premise that the knobs were real). They will now
hard-fail any launch that declares either field:

- `arenabench/arenabench/harbor_agent.py` — `_ENGINE_ROOT_FIELDS` (lines ~57-58)
  still allow-lists `pipeline_max_revisions` / `pipeline_candidates`, implying
  ArenaBench believes it can send them.
[...]

Issue #3561, opened 2026-08-17
stella-cli attaches no HostCallGate, so a plugin that declares `[loop] calls` fails its point with UnannouncedCall on the shipping path


macanderson:
## Problem

The host-call channel (`doc:wrapper-socket` §6b, #3540) is served by
`SubprocessWrapper::serving(gate)`. **`stella-cli` never calls it.**

`crates/stella-cli/src/wrapper_plugin.rs` builds the transport with
`SubprocessWrapper::declare(argv, env, timeout)` and binds it straight into
`WrapperDispatch` — no `.serving(...)`, and nothing in `crates/stella-cli/src`
constructs a `HostCallGate`, a `RecallOnly`, or an implementation of
`RecallHost` (`grep -rn "serving\|HostCallGate\|RecallHost" crates/stella-cli/src`
returns only unrelated prose matches).

The consequence is exact and reachable today, because
`plugins/stella-research` now declares `calls = ["recall"]` and asks at the
`recall` stage:

- with no gate, `SubprocessWrapper` never opens a channel, so stdin is shut
  down after the request;
- the plugin's `{"call":"recall",…}` therefore has nowhere to go, and
  `converse` returns `WrapperError::UnannouncedCall`
  (`crates/stella-runtime/src/wrapper/subprocess.rs`, the `let Some(sender) =
  &frames else` arm);
- `WrapperDispatch::open_round` records it on `DispatchReport::faults` and
  continues, so the run is not lost — the recall stage simply contributes
  nothing, every turn, on every host that installs the plugin.

It degrades safely and it is reported, which is the design working. What is
missing is the wiring that would make it *work*.

## Why the CLI is the right home for the gate

The gate needs a context plane, and the CLI is the process that has one: the
recall the pipeline uses is `ContextRecallPort::recall(goal)` over
`.stella/private/context.db` and `codegraph.db`, fanned out through the CGP
host. `stella_runtime::wrapper::RecallHost` is deliberately *not* that trait
(`crates/stella-runtime/tests/no_pipeline_edge.rs` asserts the assembly seam
declares no edge to `stella-pipeline`), so the projection from the CLI's
frames to `RecallFrame { label, kind, source, uri, content }` is the driver's
job and is mechanical.

## Definition of done

- `crates/stella-cli/src/wrapper_plugin.rs` builds a `HostCallGate::declare(
  manifest.loop_grant.clone(), DEFAULT_HOST_MAX_CALLS, Box::new(RecallOnly::new(
[...]

PR #4577, opened 2026-08-23
fix(arenabench,stella-cli): the transcript's prompt row, and fleet wrapper reports held past the live grid

### LINKED ISSUE #3883: fleet: a wrapper plugin's report lines name no task, so N concurrent workers' faults and refusals interleave unattributed
## Problem

`stella fleet --pipeline <variant>` (#3695, fleet half) binds one wrapper per worker attempt, and each attempt prints what its plugin concluded through `crate::wrapper_plugin::BoundWrapper::report` (`crates/stella-cli/src/fleet_cmd/wrapped.rs::AttemptWrapper::settle`). That call renders `report_to`'s fixed prefix:

```
  ! wrapper: <fault>
  ! wrapper: <host refusal>
  ◇ <summary>
```

Nothing in those lines names the task. With `--max-concurrency > 1` several workers print into the same stderr at once, so an operator reading `! wrapper: …` cannot tell which task's plugin said it, and two tasks whose plugins disagree are indistinguishable. On a live-dashboard run (`stella fleet` on a full TTY) these lines also land on the alternate screen the grid is painting.

`stella run` and `stella goal` do not have this problem — one wrapper, one lane, one process.

## What to build

Either give `report_to` a caller-supplied prefix (task id) that the fleet passes and the other two doors leave empty, or have the fleet render the report itself from `DispatchReport` + `HostCallGate::refusals()` + `BoundWrapper::child_spends()`. The first is smaller; the second avoids widening a shared signature for one caller. Whichever is chosen, `report_to` must stay the single renderer — the wording is asserted by `crates/stella-cli/src/wrapper_plugin/tests.rs`'s `spend_lines` tests and duplicating it is how the two copies drift.

Also decide what the live grid should do with these lines: suppressing them is not acceptable (a refusal reported nowhere is the silence #3561 closed), so they either need to reach a dashboard pane or the run has to hold them until the grid tears down.

## Files

- `crates/stella-cli/src/wrapper_plugin.rs` — `report_to`, `BoundWrapper::report`, `spend_lines`
- `crates/stella-cli/src/fleet_cmd/wrapped.rs` — `AttemptWrapper::settle`, the one fleet caller
- `crates/stella-cli/src/fleet_cmd.rs` — the dashboard tee (`FleetMsg`), if the grid is to carry them

## Definition of done

- Every wrapper line a fleet run prints names the task it came from.
- A unit test over the pure renderer proving the task-scoped form (the existing `spend_lines`/`report_to` tests are the shape to follow).
- `cargo test -p stella-cli`, `cargo clippy -p stella-cli --all-targets -- -D warnings`, `cargo fmt --check` pass.

Refs #3695 (fleet half, where this was introduced), #3561 (the refusals these lines carry), #3576 (the child-turn spend lines).

### LINKED ISSUE #4039: arena transcript: no prompt row — addendum §1's full-session anatomy is only half implemented
## Problem

`transcript-spec-addendum.md` §1 defines the rendered sequence for a turn as
**Prompt → Prose → Steps → Answer**, and is explicit about the first and last:

> Prompt and Answer get role gutters/badges (`YOU` / `AGENT`) so the session
> reads as a conversation when steps are folded.
>
> Prompt — the user's message, visually distinct, never folded away by default
> (it's the anchor for everything below).

The arena transcript renders **no prompt row at all**. A reader opening a trial
sees stages, tool calls and responses with nothing saying what was asked — the
anchor the whole page is meant to be read against is absent.

This is not a styling gap. The CSS for it already exists and is dead:
`arenabench/ui/app/transcript/transcript-surface.css` defines
`.tx-role.you .tx-roletag` alongside `.tx-role.agent`, and only the `agent` half
is ever emitted (`components/arena/transcript/entry.tsx`, the fall-through arm
at the bottom of `Entry`).

The cause is upstream of the UI: **the wire carries no prompt entry.**
`arenabench/arenabench/transcript.py`'s reader emits the kinds `text`,
`reasoning`, `tool`, `tool_result`, `usage`, `stage`, `proof`, `verdict`,
`complete`, `context_recall` and `error`. There is no `prompt` kind, so the task
instruction the trial was given never reaches the stream the page renders.

Found while fixing the adjacent defect in #4038 (a tool call and its result
rendering as two rows). That PR does not touch this.

## Files

- `arenabench/arenabench/transcript.py` — the reader; needs to emit the task
  instruction as its own entry kind, once, at the head of the trial.
- `arenabench/ui/lib/types.ts` — `TranscriptEntry.kind` is a bare `string`, so
  no type change is forced; the new kind still needs a `GROUPS` entry.
- `arenabench/ui/components/arena/transcript-page.tsx` — `GROUPS` (the kind
  filter list) and the render loop.
- `arenabench/ui/components/arena/transcript/entry.tsx` — a `prompt` arm using
  the existing `tx-role you` / `tx-roletag` markup, mirroring the `agent` arm at
  the bottom of `Entry`.
- `arenabench/ui/app/transcript/transcript-surface.css` — already has the
  styles; nothing to add unless the design review says otherwise.

- **The prompt must not be foldable away by default** (§1). The page's
  `openResults`/`thinkingOverrides` fold machinery is per-`seq`; a prompt row
  should simply not participate.
- **It must survive the kind filters as its own group.** Bundling it with
  `text` (responses) means turning responses off hides the question — the same
  argument that gave `proof` and `context` their own groups in `GROUPS`.
- **Pacing counts rows.** `mergeToolRows` now yields the paced list
  (`transcript-page.tsx`); a prompt row is one more row and should be revealed
  first, at `t = 0`.
- **The task instruction can be long.** `.tx-prose` caps the measure at 72ch;
  a prompt longer than a few lines probably wants the same fold-to-preview the
  `reasoning` arm has, but §1 says the prompt itself is never folded *away* —
  fold to a preview, do not collapse to nothing.
- Do not hand-edit `arenabench/web/` — it is a separate control-plane app, not a
  mirror of `ui/`.

## Two smaller pieces of §1 left in the same state

Worth doing in the same change or filing out, but named here so they are not
lost:

1. **Prose folds to its first sentence** (§1). The page folds `reasoning` to a
   fixed 5 lines (`THINKING_PREVIEW_LINES`), which is a different rule.
2. **Collapsing a turn shows "prompt digest · answer digest · step count ·
   rollup chips"** (§1). The arena page has no turn-level collapse at all — it
   renders one flat row list with no turn boundaries.

## Verify

1. `cd arenabench/ui && npx tsc --noEmit -p tsconfig.json`
2. `node scripts/check-transcript-view.mjs` — add a check that a trial's entry
   list begins with the prompt row and that it is not in the `text` group.
3. Open a finished match's transcript and confirm: a `YOU` badge at the top
   carrying the task instruction, present with every kind filter switched off,
   and revealed first under paced replay.

## Definition of done

A reader opening any arena trial transcript sees the task that was asked, in a
`YOU` role row at the head of the stream, before any stage or tool row — the
page reads as a conversation, which is what §1 asks for and what the existing
dead `.tx-role.you` CSS was written for.

macanderson:
## What

**#4039 — arena transcript: no prompt row.** The wire has carried the task
instruction all along: the engine registers it as a `block_registered` event of
kind `user_goal` whose `content` is the exact prompt the model was handed, and
`arenabench/arenabench/transcript.py` dropped every `block_registered` on the
floor. The reader now emits it as one `prompt` entry on the reserved seq 0
(`_next_seq` counts from 1), at `t` 0.0, once per trial — so the page's
seq-ordered render opens on it even though the block lands after the trial's
first stage rule on the wire. `full_transcript` sorts its collapsed entries by
seq for the same reason. On the page: a `prompt` kind-group chip of its own
(turning responses off cannot hide the question), a `PromptEntry` component
using the `tx-role you` / `YOU` markup whose CSS had been dead since it was
written, no participation in the per-seq fold machinery, and a long instruction
folds to a 30-line preview rather than to nothing (§1: the prompt is never
folded away). `check-transcript-view.mjs` gains the ordering checks and — it
ran in **no workflow at all** — is now wired into `diff-view-parity.yml` (the
monorepo's Node-but-no-deps job, `!cancelled()` like its siblings) and
arenabench's own dormant `.github/workflows/ci.yml`, plus a
`check:transcript-view` package script.

**#3883 — fleet wrapper report lines under the live grid.** The
task-attribution half landed in #4416 (`Refs`); this is the half it left
behind: under the live dashboard the lines still `eprintln!`'d onto the
alternate screen the grid was painting. Of the issue's two acceptable designs
(a dashboard pane, or hold until teardown) this ships **hold until teardown**:
`AttemptWrapper::settle` composes its lines through the new
`BoundWrapper::report_lines` (same single renderer `report_to` prints from, so
the wordings cannot drift) and either prints immediately (headless — unchanged)
or appends to a run-wide `HeldReports` buffer that `run_fleet` drains to stderr
after the dashboard restores the terminal. Lines already name their task, so
deferral costs no attribution; appends are whole-report under one lock, so
concurrent attempts interleave per report, never per line. Never suppressed —
a refusal reported nowhere is the silence #3561 closed.

## Evidence

- **#4039 witness, run both ways**: `git checkout origin/main --
  arenabench/arenabench/transcript.py` →
  `uv run --with pytest --no-project pytest -q tests/test_transcript_prompt.py`
  **fails** (`test_the_user_goal_block_becomes_the_prompt_row`,
  `test_the_prompt_is_emitted_once`); restored to this branch → all 5 pass.
  The extended `node scripts/check-transcript-view.mjs` passes on this branch
  (53 checks) and cannot run on main (`orderEntries` does not exist there).
- **#3883**: `cargo test -p stella-cli --bin stella fleet_cmd::` → 20 passed
  (includes the new
  `held_wrapper_reports_stay_whole_per_attempt_and_drain_once`);
  `cargo test -p stella-cli --bin stella wrapper_plugin::` → 45 passed.
  The witness is compile-level in the fail direction: `HeldReports` and
  `BoundWrapper::report_lines` do not exist on main, so the test cannot build
  there. The held-vs-print decision itself lives in `run_task`'s settle call,
  which has no unit seam; the buffer semantics are what the test pins.
- Full arenabench suite (the step `bench.yml` runs for an `arenabench/**`
  diff): `uv run --with pytest --no-project pytest -q` in `arenabench/` —
  exit 0. `uv run --with ruff --no-project ruff check` on the touched Python —
  clean. `npm run typecheck` in `arenabench/ui` — clean.
- `cargo clippy -p stella-cli --all-targets -- -D warnings` — exit 0.
  `RUSTDOCFLAGS="-D warnings" cargo doc -p stella-cli --no-deps` and the same
  with `--document-private-items` — both exit 0. `cargo fmt --all` run;
  `make guards-fast` — exit 0 (prose gate included; one construction it caught
  was rewritten, no baseline touched).
- `wrapper_plugin.rs` sits at 1492 lines on main, under the 1500 ratchet:
  `report_lines` therefore lives in `wrapper_plugin/report.rs`
  (`impl super::BoundWrapper`), not in the parent file.

## Not done

- **#3879 — skipped (maintainer decision).** Everything mechanical landed in
  #4400 while this batch was queued: `arenabench/tests/test_engine_root_fields.py`
  parses `ENGINE_ROOT_FIELDS`/`RETIRED_ENGINE_ROOT` out of `unknown.rs` and both
  drift directions are green today (re-run on this branch: 2 passed). What
  remains is exactly the `responsibilities` question, and it is a genuine
  feature decision, not a sync: the key was in the Rust vocabulary from #2462
  until #3865 deleted the pipeline, and arenabench can pin an older SUT
  (`sut_ref`) on which a declared roster still works — so silently dropping the
  emission would un-ablate a legitimately-working pinned arm (the dead-arm
  trap), refusing at posture build would break that same arm, and adding the
  key back to today's vocabulary would record ablations nothing performs.
  #3871 and #4400 both deferred it for the same reason. Needs a call between:
  retire the arena responsibilities surface (config, model, server, web form,
  309 lines of tests), or give the roster a live home in Stella's engine
  config.
- **#3670 — skipped (needs a live bench rig).** The legibility half the sketch
  asked for is already on main (#4400: `code_graph.unavailable` carries `kind`,
  `exit_code`, and Harbor's captured `stderr`). What remains is naming the
  actual cause of the `stella init` non-zero exit on frontier-bench
  `risk-scorer-replay`, which requires re-running the task; this batch is
  forbidden to launch bench runs.

Refs #3879
Refs #3670
Refs #4578
Closes #4039
Closes #3883

## Summary by Sourcery

Restore prompt visibility in arena transcripts and preserve fleet wrapper reports across live dashboard teardown.

New Features:
- Add a prompt row to arena transcripts using the task’s user-goal content, with prompt-first ordering and an expandable long-instruction view.
- Defer fleet wrapper reports during live dashboard runs and print them after terminal restoration while preserving headless output behavior.

Bug Fixes:
- Ensure transcript endpoints and the live transcript UI render entries in sequence order so prompts appear before subsequent trial activity.
- Prevent concurrent wrapper reports from being lost or visually corrupted while the alternate-screen dashboard is active.

Enhancements:
- Give prompts their own transcript filter group and role-styled presentation independent of response and fold state.
- Centralize wrapper report-line composition so deferred and immediate reports use identical wording.

CI:
- Run transcript-view validation checks in the monorepo parity workflow and ArenaBench CI workflow.

Tests:
- Add coverage for prompt extraction, deduplication, missing content, block filtering, and text coalescing.
- Add coverage ensuring held wrapper reports remain contiguous, drain once, and preserve arrival order.

### DIFF
--- a/.github/workflows/diff-view-parity.yml
+++ b/.github/workflows/diff-view-parity.yml
@@ -51,8 +51,10 @@ on:
       - "crates/stella-transcript/**"
       - "arenabench/ui/lib/diff-view.ts"
       - "arenabench/ui/lib/word-highlight.ts"
+      - "arenabench/ui/lib/transcript-view.ts"
       - "arenabench/ui/scripts/check-diff-view-parity.mjs"
       - "arenabench/ui/scripts/check-word-highlight-parity.mjs"
+      - "arenabench/ui/scripts/check-transcript-view.mjs"
       - ".github/workflows/diff-view-parity.yml"
   push:
     branches: [main]
@@ -61,8 +63,10 @@ on:
       - "crates/stella-transcript/**"
       - "arenabench/ui/lib/diff-view.ts"
       - "arenabench/ui/lib/word-highlight.ts"
+      - "arenabench/ui/lib/transcript-view.ts"
       - "arenabench/ui/scripts/check-diff-view-parity.mjs"
       - "arenabench/ui/scripts/check-word-highlight-parity.mjs"
+      - "arenabench/ui/scripts/check-transcript-view.mjs"
       - ".github/workflows/diff-view-parity.yml"
   workflow_dispatch:
 
@@ -96,3 +100,14 @@ jobs:
         if: '!cancelled()'
         working-directory: arenabench/ui
         run: node scripts/check-word-highlight-parity.mjs
+
+      # Not a Rust-parity check like the two above, but the same shape — pure
+      # functions, zero dependencies, needs Node — and until #4039 wired it
+      # here it ran in no workflow at all: the reading-tool decisions in
+      # `lib/transcript-view.ts` (errors-only pairing, prompt-first ordering,
+      # call/result folding, JSON folds) were guarded only on a developer's
+      # own initiative. Same `!cancelled()` reasoning as its sibling.
+      - name: the transcript reading-tool checks pass
+        if: '!cancelled()'
+        working-directory: arenabench/ui
+        run: node scripts/check-transcript-view.mjs
--- a/arenabench/.github/workflows/ci.yml
+++ b/arenabench/.github/workflows/ci.yml
@@ -52,6 +52,12 @@ jobs:
           cache: npm
           cache-dependency-path: ui/package-lock.json
 
+      # No dependencies — Node strips the type annotations natively — so it
+      # runs before `npm ci` and fails fast on a reading-tools regression.
+      - name: the transcript reading-tool checks pass
+        working-directory: ui
+        run: node scripts/check-transcript-view.mjs
+
       - name: npm ci
         working-directory: ui
         run: npm ci
--- a/arenabench/arenabench/server.py
+++ b/arenabench/arenabench/server.py
@@ -744,12 +744,15 @@ class ArenaServer:
         #
         # Last write wins, because the last emission is the complete one: the
         # accumulator only ever grows, and the consolidated `text` event
-        # replaces the fragment run outright. Insertion order is preserved so
-        # the transcript still reads in the order it happened.
+        # replaces the fragment run outright. Ordered by `seq` rather than by
+        # insertion: the two agree for every entry except the `prompt` row,
+        # which carries the reserved seq 0 however late its `block_registered`
+        # lands in the stream (#4039) — and the SSE client sorts, while this
+        # endpoint's caller renders the list as given.
         collapsed: dict[int, dict[str, Any]] = {}
         for entry in entries:
             collapsed[entry["seq"]] = entry
-        return {"entries": list(collapsed.values())}
+        return {"entries": sorted(collapsed.values(), key=lambda e: e["seq"])}
 
     def video(self, match_id: str, contestant_id: str, task: str) -> Path | None:
         match = self.runner.matches.get(match_id)
--- a/arenabench/arenabench/transcript.py
+++ b/arenabench/arenabench/transcript.py
@@ -447,6 +447,11 @@ class TranscriptState:
     #: ``step_usage`` routinely closes the run *before* the consolidated
     #: event arrives, and the consolidation must still find its run.
     pending_text: int | None = None
+    #: Whether the trial's ``prompt`` entry has been emitted. Once: the
+    #: engine registers the ``user_goal`` block once per turn and a trial is
+    #: one turn, but the guard costs nothing and a re-registered goal must
+    #: not print the question twice.
+    prompt_emitted: bool = False
 
 
 class TranscriptReader:
@@ -638,6 +643,37 @@ class TranscriptReader:
         # Any non-delta event closes the open text/reasoning runs, so the next
         # fragment starts a fresh entry rather than reopening a finished one.
         state.open_entries.clear()
+
+        if kind == "block_registered":
+            # The task instruction, as its own `prompt` entry — the anchor the
+            # whole transcript is read against (#4039). The wire has carried
+            # it all along: the engine registers the goal as a `user_goal`
+            # context block whose `content` is the exact text the model was
+            # handed, and this reader dropped every `block_registered` on the
+            # floor — which is why every arena transcript showed stages and
+            # tool calls with nothing saying what was asked.
+            #
+            # `seq` 0, which no other entry can hold (`_next_seq` counts from
+            # 1) and `t` 0.0: the question predates everything the trial did,
+            # and the page sorts by `seq`, so the prompt heads the transcript
+            # however late its block event lands in the stream. Every other
+            # block kind stays unrendered — the manifest/receipt plane is a
+            # reconstruction record, not reading material.
+            if (
+                state.prompt_emitted
+                or str(event.get("kind") or "") != "user_goal"
+            ):
+                return []
+            content = str(event.get("content") or "")
+            if not content:
+                # A content-free projection registers the block without its
+                # preimage; a prompt row with an empty body would read as a
+                # rendering bug rather than an honest absence.
+                return []
+            state.prompt_emitted = True
+            prompt = entry(0, "prompt", "prompt", content)
+            prompt["t"] = 0.0
+            return [prompt]
 
         if kind == "tool_start":
             call = event.get("call") if isinstance(event.get("call"), dict) else {}
--- a/arenabench/tests/test_transcript_prompt.py
+++ b/arenabench/tests/test_transcript_prompt.py
@@ -1,0 +1,137 @@
+# SPDX-License-Identifier: Apache-2.0
+# Copyright (c) 2026 the ArenaBench authors
+"""The trial's prompt row — the ``YOU`` half of the conversation (#4039).
+
+The transcript-spec addendum §1 renders a turn as Prompt → Prose → Steps →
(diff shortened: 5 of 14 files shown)


macanderson:
**Re: the two ❌ rows in Sourcery's assessment against #4039** — both are deliberately out of scope for this PR, and both are tracked in #4578.

#4039's definition of done is the prompt row: "A reader opening any arena trial transcript sees the task that was asked, in a `YOU` role row at the head of the stream, before any stage or tool row." The two flagged objectives are the items the issue itself lists under *"Two smaller pieces of §1 left in the same state"* and marks *"worth doing in the same change **or filing out**"*:

1. **Prose folds to its first sentence** — the page still folds `reasoning` to the fixed `THINKING_PREVIEW_LINES` preview. Unchanged here; the only fold this PR adds is the prompt's 30-line preview, which implements §1's "never folded *away*" rule for the prompt row specifically.
2. **Turn-level collapse (prompt digest · answer digest · step count · rollup chips)** — the arena page remains a flat, seq-ordered row list with no turn boundaries. Not attempted here.

Both are filed as #4578 (which names the files and requires the new checks to land in `check-transcript-view.mjs`, now wired into CI by this PR). Doing them in this diff would have grown a two-issue batch PR with a third, structurally different change — turn grouping reshapes the render loop the prompt row just landed in.

PR #4599, opened 2026-08-23
The prose gate now bans honesty statements and insider jargon

macanderson:
## What this does

Extends `scripts/check-prose.py` (the `make prose` gate) with new banned patterns, so every future doc, marketing page, code comment, and script header stays plain enough for an 8th grader:

- `honesty-declaration`: "honest", "honestly", "honesty". Honesty is assumed; state the fact.
- `interface-jargon`: "TUI", "REPL", "Command Deck", "deck consumer", "headless". Say "interactive mode" / "non-interactive mode". File and crate names like `stella-tui` stay in backticks, which the guard already exempts.
- `rust-jargon`: "variant" and "invariant" in prose. Say "option" and "rule".
- `enumerative-announcement` now also catches "three problems", "two claims", "four parts" and similar count announcements.

## Baseline

Regenerated from scratch (`--bootstrap`): 4283 existing hits in 1220 files are grandfathered and can only shrink. This is the one case the ratchet's own design allows a bigger baseline — the rule set grew, so the debt it records existed before the guard could see it. The old baseline had 201 lines; nothing in it was loosened.

## Evidence

- `scripts/test-prose-guard.sh`: 11/11 pass. P9 is the witness — its fixture ("The honest number is the one the TUI shows for each enum variant.") passes the old guard and fails the new one.
- `python3 scripts/check-prose.py` exits 0 on this tree: 4283 grandfathered, none added.

The middle commit is a peer session's parking of these same edits found uncommitted in a shared worktree; squash merge collapses it.

Requested by the maintainer on 2026-08-23, alongside a sweep rewriting all 175 open issues to the same standard.

## Summary by Sourcery

Enforce clearer, more accessible language across repository text by expanding the prose quality gate and preserving existing violations through its ratchet baseline.

Enhancements:
- Expand the prose gate to reject honesty declarations, insider interface terminology, Rust-specific prose jargon, and additional count-based announcements.
- Update the prose gate description to cover banned vocabulary and add a regression fixture for the new vocabulary rules.

Build:
- Regenerate the prose baseline to grandfather existing violations under the expanded rule set.

Tests:
- Add coverage verifying that banned vocabulary is detected by the prose guard.


macanderson:
The red "fmt + clippy + test" check is inherited from main, not from this diff: main's own last three CI runs fail on the same single test (docs_sync::provider_cards_match_the_registry, tracked in #4603), and this PR contains no Rust changes. Every other check is green and Sourcery approved. Re-run the check once #4603 is fixed.

Issue #3906, opened 2026-08-19
roleless-core slice 2: receipts carry the seat name, and the model-call-role list drops jobs core no longer performs


macanderson:
Slice 2 of #3903. Spec: `doc:roleless-core` §6. **Unblocks slice 6.**

## Problem

`stella_protocol::ModelCallRole` has 14 options, and several name jobs the core
product no longer performs: `WitnessAuthor`, `WitnessRepair`, `Verdict`,
`PlanRepair`, `Triage`, `Research`, `Plan`. They describe the staged pipeline
deleted in #3865. It is the **receipt** vocabulary — it groups model calls for a
cost report — and it is on the wire, exported into `docs/wire/`.

1. It names a process core does not run, so a reader of a receipt cannot tell
   which calls core made and which a plugin did.
2. It is the *only* place a role name is recorded, and it is a closed enum — so
   a plugin's call cannot be attributed to the role the plugin actually declared.
   That is what forces every downstream consumer (arenabench, the bench
   analyzers, the observatory) to copy a spelling instead of reading the name
   that ran, which is the hole `scripts/check-role-names.sh` exists to paper over.

## What to do

- Narrow `ModelCallRole` to the calls **core itself** makes. Retire the options
  above.
- Attribute a plugin's child turn as *the plugin, plus the seat name it used*,
  carried as data alongside the enum. `ChildTurnSpend` in
  `crates/stella-runtime/src/wrapper/child_turn.rs` already keeps the plugin's
  role name as a `String` — that is the shape to propagate.
- Keep the enum closed (architecture rule 7): cost reporting over a free-form string
  stops being auditable. Core's own calls stay enumerated; a plugin's call is
  one option plus data.

## Constraints

- **This is a wire change.** A new option needs a row in
  `crates/stella-protocol/src/event/consumers.rs` or it is an `E0004` build
  error, not a red test. Schemas under `docs/wire/` regenerate, and
  `.github/workflows/wire-schema.yml` gates it.
- Retired options need a serde `alias` path so an **already-recorded stream
  still parses**. Recorded benchmark traces are evidence; breaking them
  retroactively destroys measurements.
- `stella-protocol` doc comments **are** the wire contract (they become schema
  `description` fields), so prose edits require regeneration.

## Witness

A recorded stream containing a retired option still parses; a plugin child
turn's receipt names its plugin and the seat it used.

## Done when

`ModelCallRole` names no job core cannot perform, and a consumer can read the
[...]

PR #4607, opened 2026-08-24
feat(stella-protocol,stella-core): step_usage carries the request's resolved effort and output ceiling

### LINKED ISSUE #4565: Record the reasoning effort, token limit, and generation settings actually used per model call
## Problem

The Observatory's turn page now renders an **agent profile card** per seat (worker, overflow summarizer, sub-agents), built from `step_receipt` + `step_usage` journal rows plus the local model catalog. Three rows on that card must currently say **"not recorded for this run"**, because no event carries them:

- **resolved reasoning effort** — exists only in config (`AgentEngineAgent.effort`, `crates/stella-cli/src/settings/engine.rs:283` → `AgentTuning`, `crates/stella-cli/src/engine_config.rs:148` → `EngineConfig.effort`, `crates/stella-core/src/driver/config.rs:47` → `CompletionRequest.effort`, `crates/stella-protocol/src/completion.rs:204`). The value that matters is the *resolved* one (auto-mode + per-model downgrade, `engine_config.rs:413`), not the configured one.
- **requested output-token ceiling** — the *effective* per-call value is computed at `crates/stella-core/src/step.rs:700` (`output_ceiling`, moves within a turn via `output_budget_recovery`) and reaches `CompletionRequest.max_output_tokens`; nothing persists it. `StepUsage.finish_reason == length` is the only related ground truth and is itself dropped by the store (#3054's sibling gap).
- **generation params** (temperature, top_p, seed, service_tier, verbosity) — `AgentEngineParams`, `crates/stella-cli/src/settings/engine.rs:296`.

`crates/stella-model/src/wire_log.rs`'s module doc states this gap verbatim (opt-in flat-file sink, one call site, not readable by the Observatory).

## Fix shape

Add optional `#[serde(default)]` fields to `AgentEvent::StepUsage` (`crates/stella-protocol/src/event.rs:714`): `effort: Option<ReasoningEffort>`, `max_output_tokens: Option<u32>` (params can ride a later slice). Stamp them from the request at both emit sites: `crates/stella-core/src/driver/settlement.rs:284` (`emit_step_usage` — extend `SettledCall`; the driver.rs call site gains ~2 lines, it has ceiling slack) and `crates/stella-core/src/accounted_call.rs:324` (the request is in scope as `call.request`).

**Why it did not ship with the profile card:** enum struct variants require every field at construction, and `StepUsage` is constructed in dozens of test/example sites across stella-core, stella-tui, stella-cli, stella-protocol (wire_contract samples) and stella-store — a workspace-wide mechanical diff plus `make wire-schema` regeneration, too large to ride the Observatory PR.

## Done when

- Both emit sites stamp the two fields from the `CompletionRequest` that was actually sent.
- `docs/wire` schemas regenerated; serde round-trip test covers absent fields (old journals still parse).
- The Observatory profile card and the transcript metering rows pick them up with **no further change** — `crates/stella-observatory/src/journal.rs`'s `step_usage` arm and `crates/stella-observatory/src/assets/index.html`'s `seatCardHtml` already read `effort` / `max_output_tokens` defensively and fall back to the not-recorded notice.
- A witness: a driver test asserting the emitted `StepUsage` carries the request's effort and ceiling.

Refs #3054 (upstream_provider store column, same projection gap), #3906 (seat names in receipts — the custom-prompt/seat provenance half lives there).

macanderson:
## What

`AgentEvent::StepUsage` gains two optional fields, stamped from the `CompletionRequest` that was actually dispatched — never re-derived from config:

- `effort: Option<ReasoningEffort>` — the *resolved* effort (after auto-mode and per-model downgrade), which is what rides the wire as `CompletionRequestRef::effort`.
- `max_output_tokens: Option<u32>` — the *effective* per-call output ceiling after the turn's standing clamp (`ModelCallShape::max_output_tokens`), so it can move between steps of one turn and, paired with `finish_reason == length`, names the ceiling a cut-off happened at.

Both are `#[serde(default, skip_serializing_if)]`: absent on old journals, absent when the request pinned neither, and absence is never read as a value.

**Emit sites** (the issue's two):
- the engine step loop — `SettledCall` carries both, `driver.rs` passes `self.config.effort` + `call.max_output_tokens` (the same values `CompletionRequestRef` was built from), `settlement::emit_step_usage` stamps them;
- the auxiliary path — `accounted_call.rs` stamps `call.request.effort` / `call.request.max_output_tokens`.

**Observatory**: the issue claimed the dashboard would pick the fields up with no further change. That was wrong in one detail — `journal.rs`'s `step_usage` arm lifts an explicit key allowlist, so the two keys are added there, and the route test asserts the lift. The profile card's "not recorded" notice now distinguishes "this journal predates the recording" (effort / output cap) from "no event carries this yet" (custom prompt, #3906), and the stale profile-card and `wire_log.rs` module comments are rewritten to match what now rides the event.

**Wire contract**: `docs/wire/` regenerated (additive only — `ReasoningEffort` joins the `$defs`, the two optional fields join `step_usage` in both schemas and both `.d.ts`). The sample corpus sweeps every `ReasoningEffort` arm and carries the new fields in the all-present / all-absent shapes; `ReasoningEffort` gets its sampled-arm-count row so the vocabulary cannot grow unsampled.

Generation params (temperature, top_p, seed, service_tier, verbosity) deliberately do not ride this slice, exactly as the issue scoped it.

## Evidence

Targeted commands only (CI is the workspace evidence):

- `cargo test -p stella-protocol` — green (216 lib + wire_contract 8 + the rest), including the new `step_usage_from_a_pre_request_shape_stream_still_parses` (old journals parse with both fields absent) and the extended round-trip test.
- `cargo test -p stella-core` — green (1408 lib tests), including the two witnesses:
  - `driver::tests::usage_completeness::step_usage_carries_the_requests_effort_and_output_ceiling`
  - `accounted_call::tests::step_usage_carries_the_requests_effort_and_output_ceiling`

  Both are compile-level witnesses in the fail direction: the fields did not exist before this change, so the tests cannot be written against `main` — which is the gap #4565 names.
- `cargo test -p stella-observatory` — green, including the journal-route assertion that the two keys survive the lift (that assertion fails without the `journal.rs` change).
- `cargo test -p stella-tui`, `cargo check -p stella-cli --all-targets`, `cargo check -p stella-serve --all-targets` — green after the mechanical construction-site sweep (every `StepUsage {}` initializer across stella-tui, stella-cli, stella-protocol tests and the deck demo gains `effort: None, max_output_tokens: None`).
- `cargo fmt --all --check` — exit 0.

Closes #4565

## Summary by Sourcery

Capture the dispatched request’s resolved reasoning effort and effective output ceiling in step-usage telemetry and surface them through the wire and Observatory interfaces.

New Features:
- Record the resolved reasoning effort and effective output-token ceiling on each step-usage event across engine and auxiliary calls.
- Expose the new usage metadata through Observatory journal views and profile cards.

Bug Fixes:
- Preserve compatibility with older journals by treating the new usage fields as optional and absent when unavailable.

Enhancements:
- Regenerate the wire schemas and TypeScript definitions with the additive step-usage fields and complete reasoning-effort vocabulary coverage.

Documentation:
- Update wire-contract documentation and comments to describe the request metadata now captured by step-usage events.

Tests:
- Add protocol, core, and Observatory coverage for serialization compatibility, event emission, journal lifting, and all reasoning-effort variants.

Chores:
- Update existing event fixtures and construction sites to initialize the optional usage metadata.


macanderson:
Sourcery's assessment table has one ❌ — generation params (temperature, top_p, seed, service_tier, verbosity) not recorded. That is the scope #4565 itself set ("params can ride a later slice"), and the deferred slice now has its own tracker: #4621, written as a handoff with the exact emit sites, the journal-lift gotcha this PR hit, and the witness shape. The other CI red (`config::tests::docs_sync::provider_cards_match_the_registry`) is main's own break — PR #4588 moved PROVIDER_CATALOG out of the file the test reads — and #4605 is already fixing it; this branch does not touch that surface.

Issue #4606, opened 2026-08-24
A website-only commit can break a Rust test and reach main with nothing reporting red


macanderson:
## Problem

`.github/workflows/ci.yml`'s `push:` trigger carries `paths-ignore: website/**`. A commit reaching `main` whose diff is confined to `website/` therefore starts no Rust gate at all — not skipped-and-reported, but never started.

That is safe only while nothing under `website/` can break a Rust test. It can, and it did.

Two files under `website/` are read by Rust tests today:

- `website/content/docs/commands/**` — asserted about by `stella-cli`'s command-docs tests (the subcommand index, the per-command summaries, and every flag's card since #3041).
- `website/src/components/provider-catalog.ts` — the docs' copy of the provider registry, pinned field-by-field against `PROVIDERS` by `crates/stella-cli/src/config/tests/docs_sync.rs`.

On 2026-08-23, #4588 moved the provider records into `provider-catalog.ts` in a website-only diff. On the **pull_request** side the `changes` job answered `rust=false`, the required job reported `skipped`, and the PR merged green. On the **push** side the merge matched `paths-ignore` and started no Rust gate either. So `main` was red on `config::tests::docs_sync::provider_cards_match_the_registry` with nothing reporting it, until `7165b960` — the next commit that touched a crate — went red and took `f8cef7aa`, `397a83ef` and `64b85701` down with it. Four consecutive red commits, one root cause, none of them the change that caused it.

PR #4605 fixes the pull_request half by adding `provider-catalog.ts` to the `changes` job's existing carve-out (`website/content/docs/commands/` was already there for the same reason). **This issue is the push half, which #4605 deliberately does not touch.**

## Why it isn't a one-line fix

GitHub's `paths-ignore` has no negation — you cannot write "ignore `website/**` except `website/src/components/provider-catalog.ts`". So the trigger cannot express the intent directly. Options, roughly in increasing cost:

1. **Drop `paths-ignore` from the `push:` trigger** and let the `changes` job decide for pushes too, the way it already does for pull requests. `changes` currently short-circuits to `rust=true` for every non-`pull_request` event (ci.yml lines ~97-100), so this means teaching it to diff a push against `github.event.before` and apply the same regex. That regex then lives in exactly one place, which is what the trigger comment ("Keep this list and the `changes` job's regex in agreement") is asking for and cannot currently enforce. Cost: a website-only merge pays one diff instead of nothing.
[...]

Issue #4484, opened 2026-08-23
Decide whether incomplete usage data should still be included in enterprise billing exports


macanderson:
## Problem

PR #4479 (#4171) made an incomplete execution's spend a *floor* in `execution_rollup` and `stella usage report` (flagged, never dropped). `crates/stella-store/src/enterprise_telemetry/export_ledger.rs:315` still requires `e.usage_complete = 1`, so the enterprise projection never sees the floor. This is deliberate: the projection refuses an incomplete envelope, so surfacing one today only manufactures a skip. Widening the gate is an egress-policy decision (architecture rule 3), not a reporting fix.

## Decision needed

Should an enrolled org receive a flagged floor row for an execution whose usage never completed? If yes: the projection gains the flag (allowlist edit in `content_free.rs` in the same PR) and the ledger gate drops together. If no: the gate's comment states the decision and cites this issue.

Related: #4171, #4479, architecture rule 3.

macanderson:
pick whatever would pass soc2 audit

PR #4609, opened 2026-08-24
fix(stella-observatory): move the Window control into the pages it filters

### LINKED ISSUE #4590: Decide whether three more dashboard tabs get a time-range filter or stay documented as always showing all history
## Problem

PR #4589 scoped the nav's Window control (24h/7d/30d/All) to the two tabs whose surfaces actually consult `state.tf`: Overview and Activity. That fixed the reported defect (a globally visible control that silently did nothing on ten of twelve tabs), but it made a judgement permanent by omission: every other tab shows all-time aggregates with no way to narrow them.

Tabs where a time window would be meaningful but is not wired:

- **Sessions** — the session list and its KPI tiles (`crates/stella-observatory/src/sessions.rs`, `ses-kpis`/`ses-list` in `src/assets/index.html`). Rows carry `started_at`, so a client-side filter like Overview's `inTf` is feasible.
- **Self-Driving** — sessions and lesson feed (`src/assets/self_driving.js`, `src/self_driving_sessions.rs`); rows carry `started_unix`.
- **Org / hub** — hub telemetry rollups (`/api/hub-telemetry`); windowing these means SQL changes in the hub queries, not client-side filtering.
- **Models / Tools / Files** — server-side SQL aggregates over all history (`src/db.rs`; the tool leaderboard is deliberately windowed by `TOOL_CALL_SCAN_WINDOW` rowids, not by time).

## What done looks like

Either:

1. Wire the window into the tabs where it earns its keep (Sessions and Self-Driving are client-side filters; Models/Tools/Files/Hub need windowed SQL — probably a `?since=` param on those routes), and extend `TF_TABS` in `src/assets/index.html` per tab as each gains support (the witness test `tests/window_scope.rs` pins the set, so widening it is a reviewed diff); **or**
2. Decide the remaining tabs are all-time by design and record that in the module docs / the tab's kick line, closing this issue with the decision.

## Constraints

- `tests/window_scope.rs` (from #4589) asserts `TF_TABS` names only tabs whose renders consult the window — keep it accurate when widening.
- The 5s poll re-renders through the `changed()` signature gate; any new windowed render must include `state.tf` in its `changed()` key or the switch will no-op visually (see the existing call sites at `renderKpis`/`renderTimeline`/`renderExecutions`).
- Verify with the browser-automation (CDP) drive pattern used to diagnose #4589: click each `data-tf` button and assert the tab's row counts change.

macanderson:
## What & why

The Observatory's Window radiogroup (24h / 7d / 30d / All) drives `state.tf`.
Two tabs' surfaces consult it: Overview (KPI tiles, the token timeline, the
executions table, via `inTf`) and Activity (the daily charts, via `inTfDay`).
Every other tab renders all-time aggregates or configuration state.

It nevertheless sat in the global nav on all twelve tabs, and on ten of them
clicking it changed nothing — reported as "the window time filter is not
working". #4589 hid it on those ten, which fixed the do-nothing control but
left one that appears and vanishes as the reader moves around the nav, and
left #4590 asking whether the remaining tabs should gain a window or be
documented as all-time history.

This takes the third option the issue's author asked for: the control moves
**inside** Overview and Activity. Nothing to hide, nothing to document, and
no global affordance implying tabs it cannot filter.

Both instances read and write the one `state.tf`, so a window chosen on
Overview is the window Activity renders under. `setTf` now takes the window
value rather than the button that was clicked — that is what lets it repaint
the sibling group's checked state. `TF_TABS` and the `#tfWrap` hide/show in
`switchTab` are gone along with the `.tf-wrap[hidden]` rule that existed only
to make that hiding work.

Closes #4590

## The witness

### Tests removed, and why

`crates/stella-observatory/tests/window_scope.rs` is rewritten rather than
extended, so two tests from #4589 are **deliberately deleted**:

- `window_control_is_addressable_and_hideable`
- `switch_tab_scopes_the_window_to_the_tabs_it_filters`

Both pinned machinery this PR removes — the `#tfWrap` id, the
`.tf-wrap[hidden]{display:none}` rule, the `TF_TABS` set, and the
`$("tfWrap").hidden = !TF_TABS.has(tab)` line in `switchTab`. With the control
living inside the panels there is no wrapper to hide, so keeping them would
mean pinning code whose only purpose was the arrangement being replaced. The
four tests below assert the same property — the control appears only where it
filters something — against the new arrangement.

### Tests added

1. `the_nav_carries_no_window_control` — the nav slice contains no `data-tf`.
2. `only_the_tabs_the_window_filters_carry_it` — slicing `<main>` into its tab
   panels, exactly `["overview", "activity"]` carry a `data-tf` control.
3. `each_control_is_a_labelled_radiogroup_over_all_four_windows` — each names
   its own `tfLabel-<tab>` (the ids are per-page now, so a shared one would
   point both groups' labels at whichever came first) and offers all four
   windows.
4. `one_selection_drives_every_instance` — the script binds every `.tf` group
   rather than one by id, and `setTf` takes the window value.

Checked against the old asset the artisanal way — `git checkout HEAD~1 --
crates/stella-observatory/src/assets/index.html`, then re-running the test:

```
test result: FAILED. 0 passed; 4 failed
  only_the_tabs_the_window_filters_carry_it: left: [] right: ["overview", "activity"]
```

All four fail on the pre-change page; all four pass on this one.

## Driven in a real browser

The asset tests read the page as text and cannot see a render, so the built
binary was driven over CDP: `stella observe` serving this branch's page
against a real `store.db` (315 executions), headless Chrome, clicking the
controls in the panels themselves.

```
== 1. where the control lives ==
  PASS  nav carries no window control  [0 found]
  PASS  only Overview and Activity carry one  [['overview', 'activity']]

== 2. the window actually filters Overview ==
         all: execRows=315   kpis='Runs 315 234 resolved Resolve rate 74%'
         30d: execRows=315   kpis='Runs 315 234 resolved Resolve rate 74%'
          7d: execRows=186   kpis='Runs 186 154 resolved Resolve rate 83%'
         24h: execRows=82    kpis='Runs 82 66 resolved Resolve rate 80%'
  PASS  narrowing the window narrows the executions table  [all=315 > 24h=82]
  PASS  row counts are monotonic in the window width

== 3. one selection, both pages ==
  PASS  choosing 7d on Overview checks 7d on Activity
  PASS  choosing 24h on Activity checks 24h on Overview

== 4. no control where it would filter nothing ==
  PASS  no window control visible on sessions / hub / graph / config / self-driving
```

Row counts change with the window, so the relocation did not cost the control
its wiring, and the two instances stay in step in both directions.

## The gate

- [ ] `cargo fmt --check` — CI (this laptop does not build the workspace; CLAUDE.md)
- [x] `cargo test -p stella-observatory --test window_scope` — 4 passed, and 4
      failed on the reverted asset (above). The workspace run is CI's.
- [x] Docs: none owed. The change removes a documented behaviour rather than
      adding one, and the issue's author asked for no documentation.
- [x] `Closes #4590` appears both here and as a commit trailer.

**CI note.** The `fmt + clippy + test` job is red on
`config::tests::docs_sync::provider_cards_match_the_registry`, which parses
`website/src/components/provider-cards.tsx` and reads zero cards. That is
`main`'s existing breakage from #4588 (a website-only PR), tracked in #4632 /
#4606 with a fix already in flight; this branch touches no file under
`website/`. It will clear once that lands and this branch is updated.

## Nothing left behind

- [x] There is nothing new: everything I noticed in this area is fixed here,
      and the one unrelated red (`provider_cards_match_the_registry`) was
      already filed as #4632 / #4606 before I hit it.

The question #4590 actually asked — whether Sessions / Self-Driving / Models /
Tools / Files / Hub should gain a window — is answered by removing the premise
rather than deferred: with the control living on the pages that filter by it,
a page without one is not advertising a filter it does not have. If a window
is later wanted on one of those tabs, it is that tab's own change (a
client-side `inTf` filter for Sessions and Self-Driving, windowed SQL for the
rest), and `only_the_tabs_the_window_filters_carry_it` will require the new
control to be wired before it can be added.

## Ground-rule check

**Two commits, net diff is two files.** The first commit was made in a worktree
a parallel session was also using, and `git add -A` swept in that session's
ten-line doc-comment edit to
`crates/stella-store/src/enterprise_telemetry/export_ledger.rs` (#4484 work,
not mine). The second commit reverts exactly that hunk, so
`git diff origin/main...HEAD` is the two observatory files and nothing else.

**Placement.** Each control sits at the top of its panel, right-aligned
(`.tf-wrap` is now `justify-content:flex-end` rather than the nav's
`margin-left:auto`, which does nothing for a block-level flex child).

## Summary by Sourcery

Place the Observatory window filter on the Overview and Activity pages so it is available only where it filters data.

Bug Fixes:
- Move the Observatory window filter out of the global navigation and into the Overview and Activity pages where it affects displayed data.

Enhancements:
- Keep the Overview and Activity window controls synchronized through the shared selection while preserving accessible radiogroup behavior.

Tests:
- Replace navigation visibility tests with coverage verifying control placement, per-page labelling and options, and synchronized control wiring.

### DIFF
--- a/crates/stella-observatory/src/assets/index.html
+++ b/crates/stella-observatory/src/assets/index.html
@@ -446,8 +446,8 @@ nav.bar{position:sticky;top:0;z-index:20;display:flex;align-items:center;
   letter-spacing:.005em;padding:11px 13px;margin-bottom:-1px}
 .tabs button:hover{color:var(--text);background:var(--accent-wash)}
 .tabs button[aria-selected="true"]{color:var(--text);border-bottom-color:var(--accent)}
-.tf-wrap{margin-left:auto;display:flex;align-items:center;gap:var(--sp1);padding:6px 0}
-.tf-wrap[hidden]{display:none}
+.tf-wrap{display:flex;align-items:center;justify-content:flex-end;
+  gap:var(--sp1);padding:0 0 var(--sp2)}
 .tf-label{font:var(--fs-micro)/1 var(--mono);color:var(--text-3)}
 .tf{display:flex;align-items:stretch;border:1px solid var(--control-edge);
     border-radius:var(--radius-sm);overflow:hidden}
@@ -871,18 +871,6 @@ footer b{color:var(--identity);font-weight:600}
       <button role="tab" id="tab-hub" aria-controls="panel-hub" data-tab="hub">Org / hub</button>
       <button role="tab" id="tab-config" aria-controls="panel-config" data-tab="config">Config</button>
     </div>
-    <!-- Shown only on the tabs whose surfaces the window actually filters
-         (TF_TABS in the script below). Everywhere else the control used to
-         sit in the nav and silently do nothing, which reads as broken. -->
-    <div class="tf-wrap" id="tfWrap">
-      <span class="tf-label" id="tfLabel">Window</span>
-      <div class="tf" id="tf" role="radiogroup" aria-labelledby="tfLabel">
-        <button role="radio" data-tf="24h">24h</button>
-        <button role="radio" data-tf="7d">7d</button>
-        <button role="radio" data-tf="30d">30d</button>
-        <button role="radio" data-tf="all" aria-checked="true">All</button>
-      </div>
-    </div>
   </nav>
 
   <main id="main" tabindex="-1">
@@ -1003,6 +991,15 @@ footer b{color:var(--identity);font-weight:600}
 
   <!-- ══ OVERVIEW ══════════════════════════════════════════════════════ -->
   <section data-tab="overview" id="panel-overview" role="tabpanel" aria-labelledby="tab-overview" tabindex="0" class="on">
+    <div class="tf-wrap">
+      <span class="tf-label" id="tfLabel-overview">Window</span>
+      <div class="tf" role="radiogroup" aria-labelledby="tfLabel-overview">
+        <button role="radio" data-tf="24h">24h</button>
+        <button role="radio" data-tf="7d">7d</button>
+        <button role="radio" data-tf="30d">30d</button>
+        <button role="radio" data-tf="all" aria-checked="true">All</button>
+      </div>
+    </div>
     <div class="grid kpis" id="kpis"></div>
     <div class="grid cols-2">
       <div class="card">
@@ -1051,6 +1048,15 @@ footer b{color:var(--identity);font-weight:600}
 
   <!-- ══ ACTIVITY ══════════════════════════════════════════════════════ -->
   <section data-tab="activity" id="panel-activity" role="tabpanel" aria-labelledby="tab-activity" tabindex="0">
+    <div class="tf-wrap">
+      <span class="tf-label" id="tfLabel-activity">Window</span>
+      <div class="tf" role="radiogroup" aria-labelledby="tfLabel-activity">
+        <button role="radio" data-tf="24h">24h</button>
+        <button role="radio" data-tf="7d">7d</button>
+        <button role="radio" data-tf="30d">30d</button>
+        <button role="radio" data-tf="all" aria-checked="true">All</button>
+      </div>
+    </div>
     <div id="act-undated"></div>
     <div class="grid cols-2e">
       <div class="card">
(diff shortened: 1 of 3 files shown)


macanderson:
## Answering Sourcery's ❌ row

**Row:** *"Resolve the status of the remaining dashboard tabs by either wiring
the time-range filter into the meaningful tabs or explicitly documenting that
those tabs intentionally show all-time history."*

This objective is read from #4590's **original body**, which offered exactly
two options. The issue's author replied on the issue and chose neither:

> scope the time filters to the page not global to all pages and this problem
> goes away move the filter inside the page(s) where time filters work so they
> dont render on pages it doesnt and document nothing.

So the row is answered rather than fixed, on three counts:

1. **Wiring a window into Sessions / Self-Driving / Hub / Models / Tools /
   Files is explicitly out of scope**, not skipped. The author's instruction
   is to move the control, not to widen it.
2. **"Document nothing" is a literal instruction.** The original body's option
   2 asked for a recorded decision in module docs and tab kick lines; the
   author withdrew that. Adding it now would ship the thing they said not to.
3. **The premise the documentation existed to explain is gone.** Option 2 was
   needed because a *global* control implied every tab could be narrowed, so
   the tabs that could not had to say so. With the control living inside the
   two pages that filter by it, a page without one is not advertising a filter
   it does not have — there is no longer a gap for prose to close.

No follow-up issue is owed for the un-windowed tabs: they are not deferred
work, they are a decision the issue's author made. If a window is wanted on
one of them later, it is that tab's own change, and
`only_the_tabs_the_window_filters_carry_it` will fail until the new control is
actually wired — which is the guard the original issue asked for.

The other two rows are ✅ and need nothing.

PR #4610, opened 2026-08-24
feat(stella-cli,stella-tui): section the system prompt by provenance in both inspectors

### LINKED ISSUE #4602: stella inspect should render the sectioned system-prompt provenance view the Observatory has
## Problem

PR #4601 gave the Observatory's call inspector a "System prompt" view: the exact `system_prefix` bytes one model call was sent, split into provenance-labelled sections (base persona / session environment / workspace memories / rules / SessionStart hook context) by `crates/stella-observatory/src/system_prompt.rs`.

`stella inspect <execution> --step N` reconstructs the same call through `stella_store::Store::reconstruct_call` but has no equivalent of the sectioned view — a terminal user reading the same receipt gets the raw system message only, with no labelling of which setting produced each span.

## What to build

A flag on `stella inspect` (e.g. `--system-prompt`) that prints the addressed call's system prefix sectioned the same way, each section headed by its label and the setting surface that produces it.

Constraints discovered in #4601:

- The observatory's splitter is an acknowledged copy (that crate links no workspace crate but `stella-home`). The CLI side can share the real thing instead: the markers are emitted by `crates/stella-cli/src/agent/prompt.rs::assemble_system_prompt` (see `MEMORIES_OMITTED_PREFIX`) and `crates/stella-cli/src/agent/engine.rs::SESSION_HOOK_CONTEXT_HEADER`. If you extract a shared splitter, keep the observatory's copy pinned or repoint its tests.
- The splitter's contract to preserve: sections concatenate back to the input byte-for-byte (`sections_concatenate_to_the_input` in `crates/stella-observatory/src/system_prompt.rs` is the property to mirror).
- The two views must not drift: `crates/stella-cli/src/agent/prompt/tests.rs::the_observatory_provenance_markers_are_what_this_assembler_emits` pins the marker bytes from the producing side; extend it (or its sibling) to cover the CLI renderer.

## Done when

`stella inspect <id> --step N --system-prompt` prints the sectioned prompt for a receipt-bearing execution, says "no system_prefix block resolved" for a call without one (never an empty prompt), and a test proves the section labels and the byte-exact concatenation.

macanderson:
Closes #4602

## The gap

`stella inspect <id> --step N` reconstructs the exact `system_prefix` bytes a
call was sent and prints them as one undifferentiated wall. A system prompt is
*assembled* — base persona, session environment, `.stella/memories/`,
`.stella/rules/`, a SessionStart hook append — and "which of my settings put
this paragraph here?" is not a question you can answer by reading the result.
The deck's INSPECT overlay (`⌃g` / `/inspect`) had the same gap, which is what
the issue comment asks for.

## What this ships

Both inspectors now split those exact bytes on the assembler's own section
openers and head each span with its label and the setting surface that produces
it.

```
─── Workspace rules ───
from: .stella/rules/*.toml context records (edit via `stella context`)
…
```

- **CLI**: `stella inspect <id> --step N --system-prompt` (`--full` and
  `--format json` both honored; JSON carries a `found` flag so a script can tell
  an absent prompt from an empty one).
- **Deck**: the system message of a selected call renders as those sections.
  No new key, no new state — `/inspect` and `⌃g` already open the overlay.

## Design

The issue's constraint was that the Observatory's splitter is an acknowledged
copy, and that the CLI "can share the real thing instead". It does:
`crates/stella-cli/src/agent/prompt/provenance.rs` is a **submodule of the
assembler**, so its marker table names the constants the appenders push rather
than copying their bytes.

Four literals gained names at their emit sites to make that possible —
`SESSION_ENVIRONMENT_HEADER`, `MEMORIES_HEADER`, `MEMORIES_OMITTED_PREFIX`
(`agent/prompt.rs`) and `SESSION_HOOK_CONTEXT_HEADER` (`agent/engine.rs`).
**Every rewrite is byte-identical**, so the cached prefix is untouched (L-E8);
the diff is reviewable as such.

The rules heading is the one marker that cannot be shared outright:
`stella_core::records::CACHED_HEADING` opens with a single newline and
`assemble_system_prompt` pushes the second. It is now `pub`, and the joined
constant is pinned to it by `the_rules_marker_is_the_heading_the_record_channel_renders`
rather than by a comment asking the next reader to check.

`stella-tui` learns no headings. It gains an `InspectSection` on its envelope
and renders labels it is handed — the same bargain the message bodies already
take (rendered by `crate::inspect`'s own functions so the two surfaces cannot
disagree). The driver that maps a `Reconstruction` into that shape moved out of
`command_deck.rs`, a grandfathered god file closed to growth, into
`command_deck/inspect_service.rs` (AGENTS.md § "God files"); that file is 105
lines smaller than before.

Two behaviours worth calling out, both chosen so a surface never overstates what
the receipt says:

- A call that resolved no `system_prefix` prints `no system_prefix block
  resolved`, never an empty prompt. "Nothing was recorded" and "the model was
  sent nothing" are different claims and only the first is true.
- `--system-prompt` with `--diff` is **refused** rather than letting one view
  win silently, and the error names `--diff --only system` as the delta the
  caller was reaching for.

## Witnesses

`crates/stella-cli/tests/inspect_cli.rs` (drives the shipped binary against a
real store):

- `system_prompt_sections_name_the_setting_behind_each_span` — every label and
  every attribution reaches stdout.
- `the_sections_concatenate_to_the_bytes_the_call_was_sent` — the property the
  whole view rests on, asserted over the JSON: joining the bodies reproduces the
  stored prefix byte-for-byte.
- `a_call_with_no_system_prefix_says_so_instead_of_printing_an_empty_prompt` —
  both formats.
- `system_prompt_and_diff_together_are_refused_rather_than_one_winning`.

Plus: the same concatenation property at the splitter
(`sections_concatenate_to_the_input`, including a memory that quotes a later
heading and the empty prompt) and at the driver
(`the_sections_handed_to_the_deck_concatenate_to_the_message`);
`only_the_system_message_carries_a_provenance_breakdown`; two deck render tests
(`inspect_overlay_renders_the_system_prompt_by_provenance` and the no-sections
fallback, so an unsectioned message renders exactly as it always did); and
`the_provenance_markers_are_what_this_assembler_emits`, which holds the table
from the **producing** side on a really-assembled prompt — a reworded heading
fails there even if the table were edited to match it.

None of these pass on `main`.

Every deleted test: none. `deck_snapshot.rs`'s golden frames are unchanged —
its fixtures pass `sections: Vec::new()`, which takes the flat path.

## Relationship to #4601

#4601 is open and introduces the Observatory's `system_prompt.rs` splitter plus
two of the same constants (`MEMORIES_OMITTED_PREFIX`,
`SESSION_HOOK_CONTEXT_HEADER`) at the same emit sites. This branch is cut from
`main` rather than stacked on it, so **expect a small textual conflict in those
two spots** — both sides define the same name with the same value, so resolution
is "keep one". Whichever lands second should also repoint the surviving
Observatory copy's pin at
`the_provenance_markers_are_what_this_assembler_emits`, which is the
producing-side guard #4601's own test was going to be.

That hand-off, and the `BASE_SOURCE` wording minimal-prompt mode will need, are
filed as #4619 so neither depends on this description being re-read.

## Evidence

**CI is green on every check** for head `3ba1292b2`, rebased onto current `main`
— `cargo check --workspace` (MSRV), `cargo fmt --check`,
`cargo clippy -D warnings`, `cargo doc -D warnings`, `cargo test --workspace`,
the release smoke build, and every guard including `prose` and `file-size`.

All fourteen new tests are confirmed run and passing by name in that job's log:
the four `inspect_cli.rs` integration witnesses, the four splitter tests, the
three driver-mapping tests, the two deck-render tests, and the producing-side
marker guard.

Worth stating because an earlier run on this PR was red for two reasons that
were **not** this diff: it briefly carried a concurrent session's commit (see
the force-push note below), and its base predated #4605's repair of
`provider_cards_match_the_registry` — which failed identically on `main` at the
old base commit `64b85701d`. The one genuine failure of my own was two
`rustdoc::broken_intra_doc_links` on unqualified private consts in a module's
`//!` docs; those are now plain code spans, and `cargo doc -D warnings` passes.

## Docs

`website/content/docs/commands/inspect.mdx` gains the flag card, the synopsis
line, and a `--system-prompt` section covering the boundary heuristic, the
concatenation guarantee, and the deck equivalent.

### DIFF
--- a/crates/stella-cli/src/agent.rs
+++ b/crates/stella-cli/src/agent.rs
@@ -50,7 +50,7 @@ pub(crate) mod outcome;
 mod output;
 pub(crate) mod persistence;
 mod presence;
-mod prompt;
+pub(crate) mod prompt;
 mod reflect;
 pub(crate) mod resume;
 pub(crate) mod seats;
--- a/crates/stella-cli/src/agent/engine.rs
+++ b/crates/stella-cli/src/agent/engine.rs
@@ -331,11 +331,18 @@ pub(crate) fn clamp_hook_context(context: &str) -> String {
 /// bounds the cost of that mistake; it cannot remove it.
 pub(crate) async fn with_session_hook_context(mut system_prompt: String, cfg: &Config) -> String {
     if let Some(context) = session_start_hook_context(cfg).await {
-        system_prompt.push_str("\n\nSession context (from SessionStart hooks):\n");
+        system_prompt.push_str(SESSION_HOOK_CONTEXT_HEADER);
         system_prompt.push_str(&clamp_hook_context(&context));
     }
     system_prompt
 }
+
+/// The hook-context section's opening bytes, named for the reason
+/// [`crate::agent::prompt::SESSION_ENVIRONMENT_HEADER`] is: it is the last
+/// marker `crate::agent::prompt::provenance` splits a reconstructed prompt on,
+/// and a reworded heading here would relabel a span in every inspector.
+pub(crate) const SESSION_HOOK_CONTEXT_HEADER: &str =
+    "\n\nSession context (from SessionStart hooks):\n";
 
 // Pipeline port adapters
 
--- a/crates/stella-cli/src/agent/prompt.rs
+++ b/crates/stella-cli/src/agent/prompt.rs
@@ -414,6 +414,27 @@ Rules:
 /// the prompt cache on every call, so they must stay dense.
 const MEMORY_PROMPT_BUDGET_CHARS: usize = 16_000;
 
+/// The session-environment block's opening bytes.
+///
+/// Named, rather than written inline at the one `push_str` that emits it,
+/// because [`provenance`] labels the reconstructed prompt by finding these
+/// exact bytes: a heading reworded here would silently relabel a span in
+/// `stella inspect --system-prompt` and the deck's INSPECT overlay. Sharing
+/// the constant is what makes that impossible.
+pub(crate) const SESSION_ENVIRONMENT_HEADER: &str = "\n\n## Session environment\n";
+
+/// The workspace-memories block's opening bytes — a [`provenance`] marker on
+/// the same terms as [`SESSION_ENVIRONMENT_HEADER`].
+pub(crate) const MEMORIES_HEADER: &str =
+    "\n\nWorkspace memories (lessons from previous sessions — apply them):\n";
+
+/// The fail-closed memories notice's opening bytes — a [`provenance`] marker
+/// on the same terms as [`SESSION_ENVIRONMENT_HEADER`]. Distinct from
+/// [`MEMORIES_HEADER`] because the two mean opposite things: one says the
+/// memories are here, the other says they were withheld.
+pub(crate) const MEMORIES_OMITTED_PREFIX: &str =
+    "\n\nWorkspace memories were omitted from this prompt: ";
+
 // The A/B recall measurement rate lived here as a `pub(crate)` constant every
 // driver had to pass by hand, and exactly one of them did. It is now
 // `context.retrieval.ab_recall_rate` (`crate::settings`), read once at session
@@ -549,7 +570,7 @@ fn append_session_environment(
         " — not a git repository"
     };
     prompt.push_str(&format!(
-        "\n\n## Session environment\nWorkspace root: {}{repo_note}\nPlatform: {} {}",
+        "{SESSION_ENVIRONMENT_HEADER}Workspace root: {}{repo_note}\nPlatform: {} {}",
         workspace_root.display(),
         std::env::consts::OS,
         std::env::consts::ARCH,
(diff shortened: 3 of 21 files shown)


macanderson:
**Force-pushed to drop a commit that was never meant to be here.**

The first push carried a second commit, `71800436f`
("feat(stella-cli,stella-protocol): the task board is the scope, and the driver
approves it before it runs"). It is not mine: a concurrent session committed it
onto this branch while we shared a git worktree. It is also what the first CI
run was failing on — it adds `ScopeProposal::revision` in
`crates/stella-protocol/src/event/payload.rs` without initialising it at
`crates/stella-tui/src/scenario.rs:414` and `:592`, so `cargo check` fails with
two `E0063`s that have nothing to do with this PR.

That commit is **not lost**: it is pushed intact as
[`wip/task-board-scope-71800436`](https://github.com/macanderson/stella/tree/wip/task-board-scope-71800436),
and it needs those two initialisers before it can go anywhere.

The head is now a single commit cherry-picked onto `main` (`1f23a22ff`), 21
files, byte-identical in content to what was reviewed above minus that commit.
`command_deck.rs` was the one file both touched; the cherry-pick 3-way merged
cleanly and this PR's delta there is +8/−108 — the `mod` line, the two
repointed call sites, and the 106 lines moved out to
`command_deck/inspect_service.rs`.

macanderson:
@sourcery-ai review

The only review on this PR was posted against the pre-force-push head, which carried an unrelated commit from a concurrent session (now removed — see the force-push note above). The head is now a single commit rebased onto current `main`, so the file-level summary and the assessment table above describe a diff this PR no longer has.

Issue #4613, opened 2026-08-24
Only the Command Deck streams the task board — every other door records nothing


macanderson:
## Problem

`AgentEvent::TaskUpdate` has exactly one real producer in the tree:

```
$ rg -n "AgentEvent::TaskUpdate \{" crates -g '*.rs' | rg -v "tests|=>"
crates/stella-cli/src/command_deck/task_tap.rs:104
crates/stella-cli/src/command_deck/session_clear.rs:224
```

Both are the Command Deck. `TaskTap` is constructed in exactly one place
(`crates/stella-cli/src/command_deck.rs`, in `run_lead_turn`), so every other
door — a one-shot `stella run`, `agent::goal`, a delegated sub-agent
(`subagent.rs`), a sub-session worker (`subsession.rs`) — runs the same six
`task_*` tools against the same `stella_core::tasks::TaskBoard` and emits
nothing about it.

Consequences, none of them deck-specific:

- The board never reaches the event stream, so it is never persisted by the
  forwarder and a replay of a non-deck run shows no checklist at all.
- `stella-store`'s `tasks` rows are written from the board, but the journal
  that explains how it moved does not exist for those runs.
- #4594's plan gate inherits the same limit: it is installed inside
  `TaskTap::new`, so only the deck can ever raise a scope review.

## Why it matters beyond tidiness

The task board is the only durable statement of what the agent decided to do.
A headless self-driving run is exactly the case where nobody watched it
happen, and it is the case with no record.

## What to build

Move the board-to-stream mirror off the deck decorator and onto something
every door composes — `crates/stella-cli/src/agent/tool_stack.rs::n` is where
the shared stack is assembled, and `crate::turn_files::open_turn_streams_raw`
already claims "registry-born events (task board, sub-agent lifecycle)" in
its doc comment, which is the seam this probably belongs behind.

Keep the deck's behaviour identical; this is about the other doors gaining it.

## Done

A one-shot `stella run` that uses the task board emits `TaskUpdate` on its
event stream, with a witness test asserting the snapshot rides the channel
after a `task_*` call on a non-deck turn. Whether the plan gate follows is a
separate call — see #4611 for its policy — but note in the PR which way you
went and why.

Refs #4594

macanderson:
[...]

Issue #4612, opened 2026-08-24
A refused plan leaves the deck saying 'waiting input' while the model re-plans


macanderson:
## Problem

The deck infers plan approval from the next run-scoped `Stage` that is not
the gate itself (`crates/stella-tui/src/model.rs`, the
`StageScope::Run && name.kind() != Some(StageKind::ScopeReview)` arm). There
is no decision event on the wire, by design (#3861).

That works for approval. It has no counterpart for a refusal. When the
driver asks for a plan to be changed, #4594's plan gate returns a tool error
to `task_start` and emits nothing — so `SessionModel::pending_scope_review`
stays latched with the refused proposal, and until the model's revised plan
arrives (or the turn ends) the deck keeps reading:

- `deck::classify` -> `AgentStatus::WaitingInput`
- `fleet_dashboard` -> the lane is `Blocked`
- `Plan::state` -> `pending approval`

Nobody is waiting on anything at that moment: the model is re-planning. The
window is usually short, and it is wrong for its whole length.

`Plan::cancel()` (`crates/stella-tui/src/plan.rs`) already exists for exactly
this state and has no caller reachable from an event — it is the other half
of the same gap.

## Repro

Deck session, a board of three or more open tasks, `task_start`, choose
"Change it first" on the card. The lane reads `waiting input` while the model
revises the board.

## What to decide

Whether a refusal gets a wire signal, and what shape. Two candidates:

1. A run-scoped `Stage` the host emits on refusal, distinct from the approval
   one — cheapest, but overloads a boundary event with a decision.
2. Give `ScopeProposal` an id and clear the latch on a matching `ToolResult`,
   which is exactly how `AskUser` and `HunkReview` clear their cards
   ("there is no separate answer event" — `AgentEvent::HunkReview` doc). This
   is the in-idiom answer and would also let `Plan::cancel` be reached.

## Done

A refused plan stops the deck claiming a decision is pending, `Plan::cancel`
has a live caller or is deleted, and a witness test drives refuse ->
re-propose and asserts the intermediate state.

Refs #4594, #3861

Issue #4611, opened 2026-08-24
The plan-approval gate has no off switch and its step threshold is hard-coded


macanderson:
## Problem

#4594 gave `AgentEvent::ScopeReview` a live producer: the deck's plan gate
(`crates/stella-cli/src/command_deck/task_tap/plan_gate.rs`) puts the task
board to the driver on the first `task_start` of a plan with three or more
open steps, and parks the call until they answer.

There is no way to turn that off, and no way to move the threshold. The
number is a `const` in the module:

```rust
const MIN_STEPS_TO_REVIEW: usize = 3;
```

Chosen by judgement, not measurement, and stated as such in its doc comment.
Someone who never wants the card has to answer it once per plan, forever.

## What to build

A project/user-scope settings key, merged the usual three ways
(`.stella/settings.json` -> org-managed -> user), that controls both facts:

- whether the gate is installed at all;
- the step count at which it fires.

`crates/stella-cli/src/config` is where the other switches live, and
`tools.<name>` in `.stella/settings.json` is the nearest existing shape. The
gate is not a tool, so it likely wants its own key rather than a `tools`
entry — that is the design call this issue is asking for.

## Constraints already discovered

- The gate is only installed when a question responder is attached
  (`QuestionBroker::is_attached`), so a non-interactive run is already
  unaffected. This issue is about the interactive default.
- The gate must keep proceeding on every path that is not an explicit change
  request. See the module docs on why its failure direction is the opposite
  of `mid_turn_ask`'s approval responder.
- `PlanGate::install` already returns `Option<Self>`, so "off" needs no new
  plumbing — it is one more reason to return `None`.

## Done

`MIN_STEPS_TO_REVIEW` and installation both read from config, with a witness
test that a configured threshold changes whether a card is raised, and one
that the gate can be switched off entirely. The existing tests in
`crates/stella-cli/src/command_deck/task_tap.rs` are the pattern.

Refs #4594

Issue #4333, opened 2026-08-23
The plan breadcrumb has no way to show which revision the plan is on


macanderson:
## Problem

SPEC 5 item 2 and renderings `01`/`03`/`07` name the plan by revision — `▸ plan r3 · task 3 wire dedup digest · 2/6`. The SESSION tab's breadcrumb (`crates/stella-tui/src/v2/frame.rs`, `breadcrumb_spans`) draws `▸ plan · task 3 … · 2/6`, naming the plan's state word instead, because `crates/stella-tui/src/plan.rs`'s `Plan` has no revision: `ScopeReview` / `TaskUpdate` carry no `[:NEXT]`/`[:THEN]` plan-graph revision on the wire.

## Done means

- `stella_protocol` carries a plan revision on whichever event opens a plan (`ScopeReview`'s proposal, or a new event), incremented when a drift-inserted task rewrites the plan (SPEC 7.4).
- `Plan` folds it; `breadcrumb_spans` prints `plan r<n>`; the `/plan` card title agrees.
- Witness: a test in `v2/frame.rs` that a folded plan at revision 3 renders `plan r3`.

## Constraints

`stella-protocol` is types only (AGENTS.md architecture rule 4): add a serde round-trip test for the new field.

macanderson:
Blocked on a producer: `ScopeProposal` is constructed nowhere outside stella-protocol and stella-tui fixtures, and every `AgentEvent::ScopeReview` in the tree is a test — the gate that emitted it left with stella-pipeline (#3865). A revision field would render `plan r1` off a fixture forever. Tracked with its siblings in #4455; this becomes deck work once something emits the event.

macanderson:
Audit verdict (#4455): **(c) — producer deleted with #3865**. `ScopeProposal` is constructed nowhere outside `stella-protocol` samples and `stella-tui` fixtures (census 2026-08-23: the only non-fixture `ScopeReview` sites are the two consumers, `command_deck/forwarder.rs:208` and `diag_bridge.rs:797`). A revision field added now would render `plan r1` off a fixture forever. Closing into the decision issue #4594 (restore the plan producer in a plugin, or retire the variant); reopen the rendering half once that decision lands.

PR #4616, opened 2026-08-24
The task board is the scope, and the driver approves it before it runs (#4594)

### LINKED ISSUE #4594: decide: restore in a plugin or retire — AgentEvent::ScopeReview has consumers, fixtures, and no producer since #3865
## The decision

`AgentEvent::ScopeReview` (the plan-proposal / plan-approval event) has had no
live producer since the staged pipeline left the workspace (#3865). Its
consumers are real and gate-checked — `consumer-sites` holds the ledger row
`Behavioral { site: "stella-cli/src/command_deck/forwarder.rs::spawn_forwarder" }`
to live code, and the deck's ask card, plan fold and task-board seeding all
branch on it — but every `ScopeProposal` constructed in the tree is a
`stella-protocol` sample or a `stella-tui` fixture (census 2026-08-23:
`rg -ln "ScopeProposal {" crates` outside those two crates hits only the two
consumer match sites, `diag_bridge.rs` and `forwarder.rs`). A maintainer has to
pick one of:

1. **Restore the producer in a plugin.** The plan/scope gate becomes something
   a wrapper plugin emits over the wrapper socket (`doc:pipeline-as-plugins`
   §8 territory — the same home the witness/verify machinery moved to). The
   deck keeps its consumers; the rendering asks below unblock once a plugin
   can put a real `ScopeProposal` on the bus.
2. **Retire the variant.** Delete `AgentEvent::ScopeReview` + `ScopeProposal`,
   the forwarder's seeding arm, the ask-card plumbing, and the
   `scope_review` row in `crates/stella-protocol/src/event/tags.rs` — and with
   them the plan-approval UI the deck currently exercises only from fixtures.

Whichever way it goes, `ScopeReview`'s row in
`crates/stella-protocol/src/event/tags.rs` (line ~419) gets the posture the
decision implies — that row's vocabulary records consumers, so it cannot say
"no producer" today; the decision is what settles whether the row stays
`Behavioral` (restore) or is deleted with the variant (retire).

## What this blocks

- #4333 (closed into this issue): the plan breadcrumb's `r3` revision needs a
  revision number on whichever event opens a plan. Adding the field before
  this decision would render `plan r1` off a fixture forever.
- #4336's "linked plan" half: an ISSUES row can only link a plan an event
  actually opened.
- #4339's gate-failure scene (09): no `AgentEvent` names a gate today; the
  gate board's producer is verification, which is a plugin's since #3865 —
  same restore-or-retire question, one level up.

#3881 already settled the same question for `PipelineStage` /
`CandidateDelivery` / `Proof` (declared postures, issue-cited); this issue is
the remaining member of that family.

## Done means

A maintainer picks 1 or 2 and records it here. Then one PR either (1) specs
the plugin-side emitter and leaves the consumers standing, or (2) removes the
variant, its payload, its tags row, its consumers, and the fixtures — with
`make gate` green (the E0004 totality check makes a half-removal
unrepresentable). The audit trail is #4455's classification comment.

Refs #4455, #3865, #4333, #4336, #4339, #3881.

macanderson:
## What & why

`AgentEvent::ScopeReview` has had a complete consumer chain and **no producer
at all** since the staged pipeline left the workspace (#3865). The deck's plan
card, the plan rail, the fleet dashboard's `Blocked` row, the transcript fold
and the forwarder's board seeding all branch on it; every `ScopeProposal` in
the tree was a fixture.

#4594 asked a maintainer to pick *restore in a plugin* or *retire the
variant*. The answer on the issue was neither:

> this should be active whereever tasks are managed because we did the work
> for user in turn approval cards which is what this is and we have tasks
> which is what comprises the scope. wire it up.

That is a better third option, and it needs no plugin: the raw step loop
already has a plan, in the one place a plan belongs — the task board the
`task_*` tools write. So **the scope is `TaskBoard::items()`**, and the
producer is the deck's task tap, which is already the seam that turns
`task_*` traffic into `TaskUpdate`.

### Why it asks rather than only announcing

Emitting the event with no decision pending would be false in four live
places at once: `deck::classify` reads `ScopeReview` as `WaitingInput`,
`fleet_dashboard` sets the lane `Blocked`, `Plan::propose` moves the rail to
`pending approval`, and `deck_ui::dispatch` releases a sticky dispatch card on
the strength of it. Each is right when somebody is being asked and wrong when
nobody is.

It asks through the question plane (#4212/#4220) rather than rebuilding the
modal scope dialog #3861 deleted. That overlay is already attached, bounded by
a TTL, withdrawn when its wait is abandoned, and carries a note editor and a
"talk it through" way out — the shape a plan needs, because the useful answer
to a plan is usually "yes, but not that third step".

### Behaviour

- Raised on the first `task_start` of a plan of **three or more open steps**,
  **before** the tool runs, so the park sits at a safe boundary (rule #6).
- **Not installed at all** when no driver is attached, rather than installed
  and auto-approving — the rule `AgentEvent::HunkReview`'s own doc states for
  the sibling gate. A non-interactive run emits no `ScopeReview` and parks on
  nothing, so headless and self-driving runs are untouched.
- **Proceeds on every path that is not an explicit change request.** This is
  deliberately the opposite failure direction from `mid_turn_ask`'s approval
  responder, and the module doc says why: that one guards a call a policy
  flagged; this one guards nothing. A closed deck or an expired card must not
  wedge a turn nobody objected to.
- A refusal carries the driver's own words back to the model, which is what
  stops it re-proposing the plan just turned down.

### Also closed by the same wiring

`command_deck.rs` already had a live path (`e` on the plan card ->
`ScopeChangeRequest`) that told the model to "raise a scope review with the
revised plan" — something nothing in the tree could do. It now can, and the
prompt names the board instead.

### Smaller fixes carried along

- `ScopeProposal::revision` (additive, `serde(default)`) numbers a
  re-proposal after a refusal `r2`. This is the field **#4333** was blocked
  on; the plan card renders it once a plan has actually been revised, and says
  nothing on `r1` or when the producer does not number revisions.
- Two renderers printed `~0 files` for a proposal that never stated a
  magnitude. `estimated_files: 0` now means *not stated*, is documented as
  such on the field, and both renderers drop the clause.
- The question card rendered a multi-line question as one `Line`, clipped at
  the card border. It now renders one row per authored line, so the card grows
  to fit — which is what lets the plan be legible inside it.

Closes #4594

## The witness

Seven, in `crates/stella-cli/src/command_deck/task_tap.rs` and
`.../task_tap/plan_gate.rs`. The headline one is
`starting_a_plan_puts_the_board_to_the_driver_as_a_scope_review`: it fails on
`main` because **nothing in the tree constructs a `ScopeProposal` outside a
fixture**, which is the whole subject of #4594. The others pin the parts a
single happy-path test would not:

| Test | What it would let regress |
|---|---|
| `a_refused_plan_stops_the_step_and_carries_the_reason` | the step running anyway; the driver's words dropped |
| `no_driver_means_no_gate_rather_than_a_gate_that_answers_itself` | a non-interactive run recording a decision nobody made |
| `a_short_plan_is_not_worth_a_card` | a card on every two-step errand |
| `an_approved_plan_is_asked_about_once` | one card per step instead of per plan |
| `a_revised_plan_is_asked_about_again_and_numbered` | a second `r1`; a refusal counting as a decision |
| `finished_steps_are_not_part_of_the_proposal` | a card raised over a board's last remaining row |
| `silence_proceeds_and_only_a_person_refuses` | the failure direction inverting |

No tests deleted.

## The gate

- [x] `cargo fmt --check` — clean
- [x] `cargo clippy --all-targets -- -D warnings` — clean on all four crates
      this branch touches (stella-protocol, stella-tools, stella-tui, stella-cli)
- [x] `cargo test` — 16 passed, 0 failed for the gate's own suites
- [x] Docs updated: `ScopeProposal`'s field docs, the `scope_review` row in
      `event/tags.rs` (it described a producer that no longer matches reality),
      and the new module's own header
- [x] `docs/wire/` regenerated — additive only

**Full CI is green on `fea043fee`: 15 checks, 0 failures**, including
`fmt + clippy + test`, the MSRV check, `docs/wire still matches the types`,
the Windows platform split, guard self-tests, the file-size ratchet and the
prose gate.

It took three rounds to get there, and the reason is worth recording for the
reviewer rather than hidden: this worktree was shared with another agent
making unrelated in-flight edits for the whole session, and three separate
local greens here turned out to be misleading — a cached `--all-targets` pass
over four broken `ScopeProposal` literals, a second over an `E0618` in a test,
and a wire artifact I verified on `agentevent.*` while the exporter was still
mid-write on `serveframe.*`. Every one was caught by CI. The branch itself is
clean — each commit was built through a temporary git index against
`origin/main`, so none of the neighbouring work is in it — and CI, not this
laptop, is the evidence for it.

## Nothing left behind

- [x] Filed: #4611, #4612, #4613

- **#4611** — the gate has no off switch and `MIN_STEPS_TO_REVIEW` is a
  hard-coded `const`. It is a judgement number, said so in its doc, and it
  belongs in settings.
- **#4612** — a *refused* plan leaves `pending_scope_review` latched, so the
  lane reads `waiting input` for the short window while the model re-plans.
  Approval has a wire signal; refusal has none. `Plan::cancel()` is the other
  half of the same gap and still has no reachable caller.
- **#4613** — found while mapping this: `TaskUpdate` has exactly one real
  producer, and it is the deck. Every other door runs the same board and
  records nothing, so this gate is deck-only for the same reason.

## Ground-rule check

- [x] `ScopeProposal`'s new field is additive (`serde(default)`,
      `skip_serializing_if`), so a proposal that names no revision serializes
      exactly as it always has, and every pre-existing recording still decodes.
      Covered by the wire-contract sample.

**The judgement calls, so they are reviewable rather than mine:**

1. **The threshold is taste, not measurement** — three steps. No `MEASURED:`
   marker precisely because marking a taste number would cheapen the marker.
   #4611 moves it to config.
2. **This changes interactive behaviour by default.** A deck turn whose model
   plans three or more steps now asks once before starting. That follows the
   issue's "should be active", and non-interactive runs are untouched — but it
   is the reviewable half of this PR, and #4611 is the escape hatch if the
   default is wrong.
3. **`command_deck.rs` is a god file and grew by 12 lines** — an import, a
   `let goal = ...` and a call argument, plus a reworded prompt string. Every
   line of logic is in the new module, which lives at
   `command_deck/task_tap/plan_gate.rs` (declared from `task_tap.rs`)
   specifically so the god file gained no `mod` line. It stays well under its
   baseline ceiling.
4. **The interactive gate #3861 deleted is not rebuilt.** That decision's
   reasoning was "UI ahead of its socket" — a card that swallows the keyboard
   and sends its answer nowhere. The question overlay is the socket, so the
   card is not needed. If a reviewer wants the richer modal back, it is
   additive on top of this producer.

## Summary by Sourcery

Require interactive deck sessions to obtain driver approval for substantial task-board plans before starting work.

New Features:
- Gate execution of task boards with three or more open steps on driver approval before the first task starts.
- Use the task board as the source of scope proposals and emit scope review events from the deck task tap.
- Support revised scope proposals with optional revision numbers and carry driver change requests back to the model.

Bug Fixes:
- Prevent unattended runs, expired questions, or silent responses from becoming stuck in a pending approval state.
- Avoid displaying an unspecified file estimate as '~0 files'.
- Render multiline plan questions without clipping them in the question card.

Enhancements:
- Integrate scope review approval with the existing question broker and task execution flow.
- Exclude completed and cancelled tasks from proposed scope and avoid re-asking after an approved plan.
- Update plan and scope-review displays to show revision breadcrumbs only when provided.

Documentation:
- Document the semantics and wire compatibility of scope proposal revisions and unspecified file estimates across protocol and generated wire types.

Tests:
- Add coverage for plan gating, refusal handling, unattended execution, plan-size thresholds, single approval per plan, revised proposals, open-task filtering, response failure behavior, and wire compatibility.

Chores:
- Expose stable task-start and question-responder attachment checks needed by the plan gate.


macanderson:
## The witness, measured rather than asserted

The PR description originally said the headline test "fails on `main`". That
was loose — on `main` the module does not exist, so the test cannot run at
all, and a test that cannot run has not failed. Corrected here with the
measurement, run the artisanal way: **ablate the producer** (delete the
`gate.review(...)` call from `TaskTap::execute`, leave every other line of
this PR standing) and re-run.

**With the producer:** 16 passed, 0 failed.

**Ablated:** `test result: FAILED. 4 passed; 5 failed`

```
a_revised_plan_is_asked_about_again_and_numbered            ... FAILED
starting_a_plan_puts_the_board_to_the_driver_as_a_scope_review ... FAILED
a_refused_plan_stops_the_step_and_carries_the_reason        ... FAILED
finished_steps_are_not_part_of_the_proposal                 ... FAILED
an_approved_plan_is_asked_about_once                        ... FAILED
```

So five of the seven flip on the producer alone.

### The two that do not flip, and why that is correct rather than a gap

`a_short_plan_is_not_worth_a_card` and
`no_driver_means_no_gate_rather_than_a_gate_that_answers_itself` both pass in
**either** tree, because both assert that **no** `ScopeReview` is emitted.
A negative assertion cannot distinguish "the threshold held" from "there is no
producer at all" — removing the producer satisfies them trivially.

They are worth keeping as regression guards on the two cases where the gate
must stay quiet, but they are not evidence for this PR and I am not counting
them as such. The five above are.

`task_tap.rs` was restored byte-identical to its committed blob after the
ablation (verified by hash), so nothing from this experiment is in the branch.

Issue #4615, opened 2026-08-24
Raw arenabench run artifacts have no cloud home, so they cannot be deleted locally


macanderson:
## Problem

The benchmarks database on `oxagen-data` now holds the experiment documents (mirrored 2026-08-23 from `~/.arenabench/experiments.db`, which was then emptied). It does not hold the raw run artifacts those documents cite: `arenabench/arenabench-cloud/*` in the repo checkout and the run trees under `~/.arenabench/` (job directories, `stella-events.jsonl` traces, per-job `context.db` telemetry, logs). Those trees are the ground truth the repository's bench rules require conclusions to come from, they exist on exactly one laptop, and they are multi-gigabyte — so they were deliberately not deleted with the database rows.

Right now "delete the local artifacts" and "keep the evidence" cannot both be chosen. There is no cloud home for the raw trees.

## What done looks like

- A decision on the retention target — an S3 bucket in account 578673726240 is the natural one (the `oxagen-data` instance is a t4g.medium with a 20 GB root volume; it is not the place for raw artifact bytes).
- A sync path (`aws s3 sync` wrapped the way `arenabench/scripts/mirror-experiments.sh` wraps SSM, or a documented manual runbook) that records what was uploaded and its hashes, in the spirit of the `artifacts` table in `bench/telemetry_store/schema.sql` (path, size, SHA-256 per file).
- Only after a verified upload does deleting the local trees become safe. The deletion itself stays a human decision.

## Constraints

- The experiments documents in the benchmarks database cite these artifacts as evidence; deleting the artifacts without a mirror breaks the citations.
- `~/.arenabench/credentials.env` and `*.oauth` files live beside the run trees — the sync must exclude secrets.

## Definition of done

The done conditions this issue already states, as boxes the gate can read.
Nothing here is new scope, and nothing is ticked by the migration — the
gate reads one heading and this issue used another (macanderson/oxagen#1407).

- [ ] A decision on the retention target, recorded — an S3 bucket in account 578673726240 being the natural one, since `oxagen-data` is a t4g.medium with a 20 GB root volume and not the place for raw artifact bytes.
- [ ] A sync path, wrapped the way `arenabench/scripts/mirror-experiments.sh` wraps its own.
- [ ] The remaining conditions this issue lists are met.

macanderson:
Stopping short of building this: it needs an AWS decision and credentials I do
not have, and two of the premises have moved. Here is everything a maintainer
needs to decide it in one sitting.

[...]

Issue #4614, opened 2026-08-24
Nothing copies Terminal-Bench run databases into the benchmarks database yet


macanderson:
## Problem

`bench/telemetry_store/schema.sql` says the local SQLite copy is the working set and the Postgres copy is the durable one. The durable side now exists — the `benchmarks` database in the `oxagen-data-postgres-1` container on instance `i-023d002d6e44f8f84`, with all nine telemetry tables applied — and the `runs` table carries `migrated` / `migration_source` columns on both engines. But nothing pushes a local telemetry SQLite file into it. The columns are scaffolding until a pusher exists.

The experiments store already has this flow end to end: `arenabench/arenabench/mirror.py` emits idempotent content-addressed SQL and `arenabench/scripts/mirror-experiments.sh` ships it over SSM (gzip + base64 chunks, no open database port), verifies, then stamps the local rows. Use it as the template.

## What done looks like

- A `bench/telemetry_store/mirror.py` (stdlib-only, like `ingest.py`) that emits idempotent SQL for `runs`, `trials`, `events`, `tool_calls`, `verifier_tests`, `artifacts`, and the three `*_grades` tables from a local `--db`, stamping `runs.migrated = 1` and `runs.migration_source = '<machine>@local'` on the durable side.
- Primary keys are already ingester-supplied text ids, so `ON CONFLICT DO NOTHING` gives idempotency the way the experiments mirror gets it. Child tables must insert after their parents so foreign keys hold.
- A test that executes the emitted SQL twice against a scratch SQLite database and shows no duplicates (the pattern in `arenabench/tests/test_mirror.py`).
- The SSM shipping wrapper can either generalize `arenabench/scripts/mirror-experiments.sh` or copy its shape; either way the verify-before-mark ordering must survive.

## Constraints

- Emitted SQL must stay valid on both SQLite and PostgreSQL (`schema.sql`'s own rule: no AUTOINCREMENT, no BOOLEAN literals, no JSONB).
- `bench.yml` runs the telemetry-store pytest suite; the new module needs its tests in `bench/telemetry_store/tests/`.
- Events volume: one ingested 89-task run holds tens of thousands of event rows. The gzip+base64 chunking in the wrapper handles size, but measure before assuming one SSM command fits.

Issue #4603, opened 2026-08-24
A docs-sync test fails on main: it reads zero provider cards from the website


macanderson:
## Problem

`config::tests::docs_sync::provider_cards_match_the_registry` fails on main:

```
panicked at crates/stella-cli/src/config/tests/docs_sync.rs:28:5:
parsed only 0 cards from provider-cards.tsx — the parser has probably gone stale against the file's shape
```

The test parses the website's provider card list and checks it against the provider registry. Something changed the shape of `website/`'s `provider-cards.tsx`, so the parser now finds nothing. Main's last three CI runs are red on exactly this (first seen 2026-08-23 ~17:00 PDT, e.g. run 32675348787, commit f8cef7aab); every open PR inherits the red "fmt + clippy + test" check, including #4599.

## How to verify

`cargo test -p stella-cli docs_sync` on main. Fixed when that passes and a CI run on main is green.

## What done looks like

Either the parser in `crates/stella-cli/src/config/tests/docs_sync.rs` learns the file's new shape, or the file change that broke the shape is corrected. Find the breaking commit with `git log -- website/` around 2026-08-23 16:30–17:00 PDT.

macanderson:
Fixed by PR #4605 (merged 2026-08-24T07:13:38Z, commit `8edc5e1c3`). Verified directly against `origin/main`, not just the merge record:

- `crates/stella-cli/src/config/tests/docs_sync.rs`'s `read_provider_catalog()` now points at `website/src/components/provider-catalog.ts` (the file #4588 actually moved the provider records into), not the old `provider-cards.tsx`.
- `website/src/components/provider-catalog.ts` exists on `main` and contains 11 `id:` entries, so `parse_provider_catalog` has real cards to find — the "parsed only 0 cards" failure mode is gone.

Closing as already fixed.

PR #4622, opened 2026-08-24
feat(telemetry-store): mirror-provenance columns on the runs table

macanderson:
## What this does

`bench/telemetry_store`'s `runs` table gains the two columns that let a run say
whether it has reached the durable benchmarks database, and from where:

- `migrated INTEGER NOT NULL DEFAULT 0` — 0 on a row written in place, 1 once a
  mirror has copied it into the durable copy.
- `migration_source TEXT` — the `machine@tier` label (`Mac@local`) the copy came
  from, so the durable side can always answer which machine produced a row.

They are declared last in `schema.sql` and added by `ingest.py::migrate` with
`ALTER TABLE`, so a database that gains them through the migration ends up with
the same column order as one created fresh — the convergence the existing
`cost_norm_status` migration already established in this file.

The pusher that will set them is #4614.

## This PR was split

It opened carrying two halves. #4642 ejected `arenabench/` into
**macanderson/arenabench**, so the experiments-store and mirror half — six of the
original eight files — has no home in this repository any more. It is now
**macanderson/arenabench#2**, which also carries every finding Sourcery raised
here. What remains in this PR is the `bench/telemetry_store` half, which merges
onto `main` cleanly.

The three files that moved, and where each went:

| Was | Now |
|---|---|
| `arenabench/arenabench/experiments.py` | `arenabench/experiments.py` in macanderson/arenabench#2 |
| `arenabench/arenabench/mirror.py` | `arenabench/mirror.py` |
| `arenabench/scripts/mirror-experiments.sh` | `scripts/mirror-experiments.sh` |

(plus `docs/experiments-normalization.md` and the two test modules.)

## Evidence

- Witness test: `test_migration_adds_mirror_provenance_to_an_old_runs_table` builds
  the `runs` table in its exact pre-column shape, runs `connect()` twice, and reads
  the pair back. On `main`'s `ingest.py` and `schema.sql` it fails with
  `sqlite3.OperationalError: no such column: migrated`; with this change the whole
  module passes 11/11.
- The migration is additive and idempotent, and a pre-existing run reads as
  `migrated = 0`, `migration_source = NULL` — never mirrored, rather than
  mirrored from nowhere.

## Deleted tests

None.

## Follow-ups filed

- #4614 — telemetry-store working sets have these provenance columns but no pusher
  into the benchmarks database yet.
- #4615 — raw arenabench run artifacts (multi-GB, evidence the experiment documents
  cite) have no cloud home and were deliberately not deleted.

## Summary by Sourcery

Add mirror provenance tracking to telemetry-store runs and migrate existing databases to the updated schema.

New Features:
- Add mirror provenance fields to telemetry-store runs, recording whether a run was migrated and its source machine and tier.

Enhancements:
- Extend telemetry-store migrations so existing databases receive the new runs columns safely and idempotently, with pre-existing runs marked as not migrated.

Tests:
- Add coverage for migrating legacy runs tables and verifying default provenance values and repeatability.


macanderson:
The red `fmt + clippy + test` check is inherited from the base, not from this diff: `config::tests::docs_sync::provider_cards_match_the_registry` fails identically on main (run 32675383805, same panic, same 2198 passed / 1 failed). Tracked in #4603; #4605 carries the fix. This PR touches only arenabench/** and bench/telemetry_store/** (Python + SQL); every other check is green, including the bench workflow that runs the arenabench and telemetry-store pytest suites this PR changes.

macanderson:
Two separate things are keeping this red, and only one of them is mechanical.

**The `fmt + clippy + test` failure is stale.** The run above is from 01:10 and
died on `config::tests::docs_sync::provider_cards_match_the_registry`, which was
broken on `main` itself — a website-only diff moved the provider records into
`provider-catalog.ts` out from under that test's parser (#4606). #4605 repaired
it. Every other PR that was red on this has cleared by updating its branch.

**But this one cannot just be updated.** #4642 ejected the harness into
`macanderson/arenabench`, so `arenabench/` no longer exists on `main`, and six
of this PR's eight files live under it:

```
arenabench/arenabench/experiments.py
arenabench/arenabench/mirror.py
arenabench/docs/experiments-normalization.md
arenabench/scripts/mirror-experiments.sh
arenabench/tests/test_experiments.py
arenabench/tests/test_mirror.py
```

Merging `main` in would meet the eject as a modify/delete on all six. So this
needs a decision rather than a branch update: move the experiments-store work
to the `arenabench` repo and keep only the two `bench/telemetry_store/` files
here, or land it before/alongside the eject's follow-up.

Flagging rather than acting — I was clearing PRs blocked by the `main` breakage
and this one is blocked by something else.

macanderson:
Answering the five findings, and the structural blocker above them.

**The blocker first.** #4642 ejected `arenabench/` into
[macanderson/arenabench](https://github.com/macanderson/arenabench), and it
merged at 13:25 today — so the option of landing this before the eject is gone,
and merging `main` in would meet a modify/delete on six of the eight files. The
split is the only path left, and it is done: this PR is now the
`bench/telemetry_store` half alone and merges cleanly, and the experiments-store
half is **macanderson/arenabench#2**.

**All five findings are in files that moved, and all five are settled there.**
Each fix has a test that fails without it; the failure each one produces on the
pre-fix code is quoted below, from an actual revert check rather than from
reading the diff.

1. **`mirror-experiments.sh:77` — `CREATE DATABASE $PG_DB` inside single quotes.**
   Real, and already fixed on this branch in 3aa48f1e6 (your suggestion applied
   verbatim). It carries over. While there I found the same command had a second
   defect nobody flagged: the `A && B || C` chain fell through to the `||` arm
   when the *SQL reconstruction* failed, so a push that had shipped nothing
   created the database and reported success. It is an `if` now.

2. **`mirror.py:170` — duplicate documents make verification reject a successful
   mirror.** Real, and the sharper statement is that the emitted-statement count
   was never the right thing to verify: the durable key is the content hash, so
   *N* local rows holding one document are one durable row by construction.
   `mirror_rows` now folds by `doc_sha256` and carries `local_ids`, and the script
   counts hashes read from a manifest rather than `grep`-ing the SQL. Before the
   fix: `assert 2 == 1`.

3. **`experiments.py:307` — `mark_migrated` stamps rows the mirror never sent.**
   Real. Rather than making the row set an option, it is now required:
   `mark_migrated(source, ids, ...)`, with `ids` coming from the manifest the emit
   step wrote, so "everything currently unmigrated" is no longer expressible. The
   `source` label travels in the manifest too, so the emit and mark steps cannot
   disagree about it. Before the fix: `TypeError: mark_migrated() takes from 1 to
   2 positional arguments but 3 were given`.

4. **`mirror.py:161` — `_literal` calls `.replace` on non-string header values.**
   Real. A `_header_text` step renders JSON scalars as canonical text (`7`,
   `true`) and refuses containers by field *and* by local row id through a named
   `MirrorError`, so the message says which document is wrong instead of naming
   `str.replace`. `_literal` itself is now strict about its argument. Before the
   fix: `AttributeError: 'int' object has no attribute 'replace'` — exactly the
   trace you predicted.

5. **`experiments.py:204` — opengrep, SQL string concatenation.** Not a defect,
   and I am not fixing it. The interpolated value is `_results_as_json_sql()`,
   which returns one of two module literals — `"json(results)"` or `"results"` —
   selected by `sqlite3.sqlite_version_info` and by nothing else. No caller value
   reaches that function, so no caller value reaches the SQL. It cannot be a bound
   parameter either: it is a column expression, and a placeholder is not legal
   there. The document text on the same query *is* parameterized. I have put that
   argument in the function's docstring so the next reader meets it before the
   scanner does.

Nothing here is deferred, so there is no follow-up issue owed for the five. The
two that were already filed — #4614 (no pusher for telemetry working sets yet)
and #4615 (raw run artifacts have no cloud home) — are unchanged and still open.

PR #4630, opened 2026-08-24
fix(stella-transcript,stella-observatory): a live transcript's server-side poll cost is O(new events), not O(run)

### LINKED ISSUE #4566: A live transcript view redraws its whole content on every update instead of only the new part
## Problem

The turn page now embeds the shared renderer's fragment (`/api/transcript-html`) in a shadow root (`crates/stella-observatory/src/assets/index.html`, `paintRenderedTranscript`). For a **still-running** execution, `refreshTranscriptJournal` keeps the cheap `after_seq` probe to detect news, but on any news it re-fetches and repaints the **whole** fragment — the shared renderer draws a tree (`stella_transcript::html::render_run`), not an appendable list, so there is nothing smaller to ask for.

Costs, growing with run length: server re-folds the full journal (`Observatory::execution_journal(id, full=true)` + `transcript_view::build_run`) every poll tick while a run is live; the repaint resets any fold the reader opened by hand (the renderer's pin-open rules cover running/error nodes, so the common case survives, but a manually opened old step re-collapses).

## Fix shape

Either (a) a fragment-diff protocol: `/api/transcript-html?after_seq=N` returning only new turn-tail markup plus a splice point — needs `render_run` to expose a per-turn/per-step streaming API the way `grid::render_turn_lines` already does for the character grid (that precedent is the reason the surfaces guard counts both `render_page` and `render_run`); or (b) client-side fold-state preservation across repaints: read the shadow DOM's `details[open]` ids (NodeId keys are stable: `t0s3`…) before repainting and reapply after.

(b) is a 20-line JS change and removes the UX half; (a) removes the server cost and should follow the grid renderer's streaming shape rather than invent one.

## Verify

Open a live run's turn page in `stella observe`, watch the poll while the run works: folds you opened stay open across ticks (b), and/or server work per tick is O(new events) not O(run) (a).

Refs #1476 (the original incremental journal poll this replaced).

macanderson:
## What & why

The turn page embeds the shared renderer's fragment (`/api/transcript-html`)
in a shadow root. For a still-running execution, the poll re-fetched and
repainted the **whole** fragment on any news — the shared renderer draws a
tree, not an appendable list — so the server re-folded the full journal and
re-rendered every already-shown block on every tick, cost growing with run
length. #4593 already fixed the UX half (fix shape (b): a repaint preserves
folds the reader opened by hand); this PR is the remaining server-cost half
(fix shape (a)).

`stella_transcript::html` grows `render_turn_tail`, HTML's counterpart to
`grid::render_turn_lines`. Given a `TailCursor` (a step index — the running
boundary, since a step folds before its `tool_result` arrives and needs
rewriting once it settles — plus plain note/prose counts, since neither is
ever rewritten once folded), it renders only the blocks a live run has grown
since, plus a freshly rendered receipt carrying a stable id.
`/api/transcript-html?id=N&from_step=N[&from_note=&from_prose=]` serves this
as JSON instead of the full HTML fragment; a stale or out-of-range cursor
falls back to the existing full fragment rather than erroring. The journal is
still refolded whole every call (cheap — an allocation-only walk over rows an
indexed query already fetched); the cost this cuts is rendering's: escaping,
syntax highlighting and diff-table construction over content the previous
tick already drew in full, which is where a long-running turn's poll cost
actually grows.

The dashboard's `refreshTranscriptJournal` tries `spliceTranscriptTail`
first: it reads the resume cursor purely from the shadow DOM's own markup (a
step's `data-status="run"`, and note/prose element counts — matching how
fold state already lives in `details[open]` rather than a script-side map),
splices the returned blocks in by id (replacing the one node that may have
changed, appending the rest), and swaps the receipt. Anything that isn't a
clean splice — first load, a downgraded response, a stale cursor — falls
through to the existing full repaint with fold-state preservation (fix shape
(b), unchanged).

`scripts/check-transcript-surfaces.py`'s HTML pattern now counts
`render_turn_tail` alongside `render_page`/`render_run`, mirroring how GRID
already counts `render_turn_lines` beside `render`.

Closes #4566

## The witness

Server-side: `stella-transcript`'s new `tests::tail` module (5 tests) and two
new route tests in `stella-observatory`'s `execution_routes`
(`transcript_tail_resumes_from_the_still_open_step`,
`transcript_tail_with_a_stale_cursor_falls_back_to_the_full_fragment`) —
a still-open step is rewritten once it settles and never duplicated; an
already-rendered note/prose never repeats; an out-of-range cursor is
refused; the receipt id is stable between a full and a tail render; a stale
`from_step` at the HTTP layer degrades to the full fragment rather than
500ing. All fail to compile against `main` (the API does not exist there)
and pass with the change.

Client-side splice logic has **no** witness test, for the same reason
#4593's fix shape (b) doesn't: browser JS in an embedded asset with no DOM
harness in this repo. Verified by tracing every branch by hand against the
exact server response shape, and `node --check` for syntax. Manual
acceptance check per the issue's own: open a live run's turn page in
`stella observe`, watch the poll while the run works — folds opened by hand
survive ticks (unchanged, fix shape (b)), and server work per tick is now
bounded by what changed rather than the whole run (fix shape (a)).

## The gate

- [x] `cargo fmt --check` (`cargo fmt -p stella-transcript -p stella-observatory`, no diff)
- [x] `cargo clippy --workspace --all-targets -- -D warnings` — targeted: `cargo clippy -p stella-transcript -p stella-observatory --all-targets -- -D warnings`, clean
- [x] `cargo test --workspace` — targeted: `cargo test -p stella-transcript` (124 passed), `cargo test -p stella-observatory` (122 passed)
- [x] Docs updated where behavior/flags changed (README, `--help`, doc comments) — new doc comments on `render_turn_tail`/`TailCursor`/`TurnTail`; no README/CLI surface changed

Also ran (this PR touches guarded surfaces): `RUSTDOCFLAGS="-D warnings"
cargo doc -p stella-transcript -p stella-observatory --no-deps` (clean),
`scripts/check-file-size.sh` (clean — the new tests moved into
`crates/stella-transcript/src/tests/tail.rs` to stay under the 1500-line
ceiling on `tests.rs`), `scripts/check-transcript-surfaces.py` (clean, 2
shared + 3 declared gaps, unchanged), `scripts/check-prose.py` (clean, no
lines added to the baseline).

CI is the workspace evidence (per this repo's own convention — no
`cargo build --workspace`/`cargo test --workspace` run on the maintainer's
machine); the above are the targeted, crate-scoped commands for the crates
this PR actually touches.

## Nothing left behind

## Ground-rule check

- New cross-boundary types round-trip through serde (test included) — N/A: `TailCursor`/`TurnTail` are plain in-process Rust types passed between `stella-transcript` and `stella-observatory` inside one binary, never independently serialized (the JSON response is hand-built from their `String` fields via `serde_json::json!`, matching this route's existing shape for every other JSON-returning route in this file).

The server still refolds the whole journal every tick (`transcript_view::build_run`
stays O(run), not incremental) — that part of the cost was already cheap
(an allocation-only walk over rows an indexed SQLite query fetches), so this
PR bounds the actually-expensive half (rendering) rather than building an
incremental-fold cache. An incremental fold would also mean adding mutable,
per-execution server-side state to a crate whose stated design is read-only
and stateless end to end; this PR avoids that architectural change entirely
by keeping the fold stateless and bounding only the render step, which is
where the real cost was concentrated for long or output-heavy runs.

## Summary by Sourcery

Bound live transcript polling cost by rendering and splicing only content added or changed since the previous update.

New Features:
- Add incremental transcript-tail rendering for live executions, returning only newly added or still-changing blocks plus an updated receipt.
- Update live transcript polling to splice tail responses into the existing shadow DOM while preserving a full-render fallback for stale or invalid cursors.

Bug Fixes:
- Reduce server-side live transcript polling work from re-rendering the entire run to rendering only changed content, preventing cost from growing with run length.

Enhancements:
- Expose stable receipt and step status markup to support reliable client-side tail resumption.
- Extend transcript surface checks to account for the shared tail renderer.

Tests:
- Add transcript renderer and route coverage for resuming from open steps, avoiding duplicate notes and prose, rejecting invalid cursors, keeping receipt IDs stable, and falling back safely on stale cursors.


macanderson:
The one red required check ("fmt + clippy + test") is a pre-existing, unrelated failure: `config::tests::docs_sync::provider_cards_match_the_registry` in `stella-cli` — "parsed only 0 cards from provider-cards.tsx". Already tracked in #4603 (filed before this PR opened), a website-doc parser drift with nothing to do with the transcript renderer or observatory routes this PR touches. This PR's own crates (`stella-transcript`, `stella-observatory`) are green: 124 + 122 tests, clean clippy, clean rustdoc, clean file-size/prose/transcript-surfaces guards.

PR #4648, opened 2026-08-24
feat(stella-transcript,stella-observatory): incremental transcript fragments for a live turn

### LINKED ISSUE #4566: A live transcript view redraws its whole content on every update instead of only the new part

macanderson:
Fix shape **(a)** from #4566 — the incremental fragment protocol. Shape (b) (fold-state preservation) shipped in #4593; this is the half that removes the server cost, and it closes the issue.

## The problem

A still-running turn page re-fetched and repainted the **whole** fragment on every poll tick. The shared renderer draws a tree, not an appendable list, so there was nothing smaller to ask for: each tick re-folded the full journal (`Observatory::execution_journal(id, full=true)` + `transcript_view::build_run`) and re-rendered every step, growing with run length.

## What ships

**`stella-transcript` grows a tail renderer beside `render_run`** (`html.rs`). A turn's body splits into *step groups* (the notes and prose ahead of a step, then the step) and a *closing region* (trailing notes/prose, the answer, the open-turn receipt). `render_turn_tail` returns `TurnTail { blocks, close }`, both children of the turn's `<details>` so a host page can splice them.

The contract is `grid::render_turn_lines`' append-only contract pointed at this surface: **a settled step group's markup is final the moment it is produced.** `TailBase` carries what the settled prefix contributed — the three index offsets that keep fold ids numbered over the whole turn, the accounting the receipt sums, and the time column's dedup seed — so a renderer holding only the suffix still states whole-turn facts.

`Ctx::step_fold` states the one subtlety in a single place: a step-addressed node's **id** continues the whole turn's numbering (so a reader's open folds keep the addresses they were opened under), while the **fold lookup** indexes the model in hand (which is what keeps a running step pinned open).

**The Observatory folds only what is new** (`transcript_view.rs`). `build_run`'s step loop is extracted to `fold_rows`, shared with `build_run_tail`. The client-echoed `TailCursor` is content-free (counts, seqs, token totals — no bodies) and resumes on a **completing `tool_result` boundary**, chosen because right after one the fold holds no half-built call and no unattached `step_usage`, so a resumed fold sees exactly what the full fold saw. A step settles only once a later completed step exists, since a trailing `step_usage` can still merge figures backward into it.

`GET /api/transcript-tail?id=N&cursor=…` returns `{settled, tail, cursor, finished}`. A missing or malformed cursor degrades to a full-size tick, never an error. `/api/transcript-html` stays the answer for first paints, finished runs, and any client that dropped its cursor.

**The page appends `settled` once and swaps only `tail`** (`assets/index.html`). Settled DOM is never touched, so it keeps the reader's fold toggles natively; the replaced tail gets them back from the capture/reapply pass #4593 added. When the run finishes, one whole-fragment repaint lands the final turn summary.

**The surfaces guard counts `render_turn_tail`**, for the reason it already counts `grid::render_turn_lines` (its own comment): a surface moving onto the streaming API is using the shared renderer harder, not less. Without this, a later change moving the page fully onto the tail API would read as a regression away from sharing.

## Evidence

Targeted commands (CI is the workspace evidence):

- **`tests::transcript_tail::ticked_composition_matches_the_full_render_at_every_journal_length`** — the protocol's whole claim. Simulates the poll loop over a journal that exercises the fold's corners (a double `step_usage` merging backward, an answer settling behind the cursor, prose between steps, a call still running) and asserts at **every** journal length that the page's composition — head + accumulated settled fragments + the tail just received — is byte-identical to a fresh full render, and that the never-repainted head never moves.
- **`tests::tail::tail_splices_into_the_full_render_at_every_step_boundary`** — the renderer-level splice identity at every boundary, covering fold ids, the time dedup, the receipt's whole-turn chips and the pinned-open running step.
- **`tail::settled_prefix_is_byte_stable_as_a_running_turn_grows`** — the finality property the splice rests on.
- Plus `a_quiet_tick_is_idempotent`, `cursor_round_trips_and_a_mangled_echo_degrades`, `tail_time_column_dedups_against_the_settled_prefix`, and a route-level test pinning the HTTP wiring against `/api/transcript-html`.
- `cargo test -p stella-transcript -p stella-observatory` — **green**, 14 test binaries, 0 failures.
- `cargo clippy -p stella-transcript -p stella-observatory --all-targets -- -D warnings` — exit 0, **no `#[allow]` added** (the first draft's four `too_many_arguments` allows were removed by adopting `grid.rs`'s `Ctx` idiom).
- `RUSTDOCFLAGS="-D warnings" cargo doc` (both crates), `cargo fmt --all`, `make guards-fast`, `make prose`, `./scripts/check-file-size.sh`, `python3 scripts/test-transcript-surfaces.py` — all exit 0.

**No witness for the client half**: it is browser-side JS in an embedded asset with no DOM harness in this repo, and a text match over `index.html` would be a changelog, not a test. Its server contract is witnessed above; the manual check is the issue's own — open a live run's turn page in `stella observe` and watch server work per tick stay O(new events) while hand-opened folds survive.

Closes #4566

## Summary by Sourcery

Stream live transcript updates incrementally by appending settled step groups and refreshing only the unsettled tail.

New Features:
- Add an incremental transcript-tail API that returns newly settled fragments, the live tail, a resumable cursor, and completion state.
- Update live transcript pages to append settled content and replace only the changing tail while preserving fold state and inspector controls.

Bug Fixes:
- Avoid re-folding and re-rendering the entire live transcript on every polling tick while retaining correct whole-turn answers, timings, totals, and fold behavior.

Enhancements:
- Refactor transcript folding and HTML rendering to support byte-identical tail splicing with stable fold identifiers and carried turn metadata.
- Add indexed journal-entry recovery for answers that precede the incremental cursor.
- Extend transcript surface guards to account for incremental tail rendering.

Tests:
- Add renderer- and protocol-level tests covering splice equivalence, settled-prefix stability, cursor handling, quiet ticks, time deduplication, and HTTP route wiring.


macanderson:
Superseded by #4630, which merged first and implements the same thing — #4566's fix shape (a), the incremental fragment protocol. Not a judgement on this diff: the two were written independently against the same issue, and #4566 is already CLOSED/COMPLETED by #4630.

Evidence that main's version is real rather than merely merged: `crates/stella-transcript/src/tests/tail.rs` exists on `origin/main` (from #4630, e99812243) and `cargo test -p stella-transcript --lib tests::tail` passes 5 tests there, including `a_step_still_open_at_the_last_render_is_the_replace_boundary` and `the_receipt_id_is_stable_between_a_full_render_and_a_tail_render` — the boundary and identity properties an append protocol has to get right.

Merging this branch would have meant resolving four conflicts (`stella-observatory/src/lib.rs`, `stella-transcript/src/html.rs`, an add/add on `tests/tail.rs`, and `scripts/check-transcript-surfaces.py`) into a rival design of the same feature — the shape where a clean textual resolution silently reverts half of one side. If anything in this branch is better than what landed (the surfaces guard's handling in particular), it is worth a small follow-up PR against main rather than this merge; say so and I will file it.

PR #4665, opened 2026-08-24
fix(stella-cli,stella-protocol,stella-core,bench): one allowed-model authority, step_usage's generation params, and a minimal-prompt posture arm

### LINKED ISSUE #4618: allowed_models says it is the seat ceiling, but seats never read it
## Problem

`crates/stella-cli/src/settings/engine.rs` (the doc on `allowed_models`, around line 79) says the list "is the ceiling on seat assignments too … a seat naming a model outside a non-empty list is refused with a notice" and points at `crate::agent::seats`. That enforcement does not exist: `resolve_seat_models` in `crates/stella-cli/src/agent/seats.rs` checks only that the seat's provider is configured and the adapter builds — it never reads `allowed_models`. The session's own model resolution in `crates/stella-cli/src/config.rs` doesn't consult the list either.

So today the list binds three surfaces and not the other two:

- **Enforced:** the SETTINGS tab's picker vocabulary (`crates/stella-tui/src/views/engine.rs`, `picker_candidates`), the `/model` session picker, and the `/model <spec>` typed override (`command_deck/session_override.rs::switch_session_model`, since PR #4617).
- **Not enforced:** plugin seat assignments (`seat_models`), and the session default resolved at startup (`--model` or `default_model` can name an off-list model with no notice).

A doc comment promising a refusal that never happens is the kind of claim CLAUDE.md treats as a defect.

## What done looks like

Either implement the ceiling or retract the claim — implementing is the direction the doc, the picker, and the override all already point:

1. `resolve_seat_models` refuses (notice + fall back to session model; refuse the run under `EnginePosture::Trusted`) a seat whose resolved spec is outside a non-empty `allowed_models`, matching the existing degradation shape in that file.
2. Decide and document whether the startup default (`--model` / `default_model`) is also bound. If yes, a boot notice naming the list; if deliberately exempt, say so in the `allowed_models` doc so the exemption is a statement instead of a silence.
3. Witness tests: a seat naming an off-list model produces the notice and rides the session model (fails today); the doc's claim and the code agree afterwards.

## Verify

`cargo test -p stella-cli seats` for the seat half; the new witnesses for the rest. The matching semantics to reuse: `switch_session_model`'s check in `crates/stella-cli/src/command_deck/session_override.rs` (full `provider/slug` spec or the raw configured string).

Refs #4617 (which enforced the list for the session override and left this half open).

### LINKED ISSUE #4621: step_usage records the request's generation params (temperature, top_p, seed, service_tier, verbosity)
## Problem

#4565 (shipped by PR #4607) put the request's resolved reasoning effort and output ceiling on `AgentEvent::StepUsage`. The third row of the profile card's gap — the generation params the request carried — was deliberately left out of that slice and has no tracker of its own now that #4565 closes.

The values live on the request and nowhere durable: `CompletionRequest.params` (`GenerationParams`: temperature, top_p, seed, service_tier, verbosity — `crates/stella-protocol/src/completion.rs`) and `CompletionRequest.temperature`. `crates/stella-model/src/wire_log.rs`'s module doc names this as the remaining ask-side gap (with tool schemas, which are content-bearing and stay off the event).

## Fix shape

Same pattern as #4565, one field:

- Add `#[serde(default, skip_serializing_if = "Option::is_none")] params: Option<GenerationParams>` (or a purpose-built content-free struct that also folds in the top-level `temperature`) to `AgentEvent::StepUsage` (`crates/stella-protocol/src/event/kind.rs`).
- Stamp it at both emit sites from the dispatched request: `crates/stella-core/src/driver/settlement.rs` (`SettledCall` + the `driver.rs` call site) and `crates/stella-core/src/accounted_call.rs`.
- Mechanical sweep of every `StepUsage {}` initializer (stella-protocol tests, stella-core, stella-tui, stella-cli — see PR #4607's diff for the exact site list), `make wire-schema-update`, and the wire-contract samples carrying the all-present / all-absent shapes.
- Lift the key in `crates/stella-observatory/src/journal.rs`'s `step_usage` arm (an explicit allowlist — #4565 learned the fields do not flow without it) and render the row on the profile card (`seatCardHtml` in `crates/stella-observatory/src/assets/index.html`).
- Update `wire_log.rs`'s module doc, which will otherwise state a stale gap.

## Done when

Both emit sites stamp the params from the request actually sent; old journals still parse (witness like `step_usage_from_a_pre_request_shape_stream_still_parses`); a driver test asserts the emitted event carries the request's params; the profile card row stops reading "not recorded" on new runs.

Refs #4565, #3906 (the custom-prompt row, the card's last remaining gap).

### LINKED ISSUE #4650: Benchmarking minimal prompt mode needs the key admitted by the posture vocabularies
## Context

PR #4601 added `agent_engine_config.minimal_prompt` (`[agents] minimal_prompt`
in TOML, `--minimal` on the command line): the session runs a bare
tool-advertisement base persona so the operator's prompt fields carry the
prose. It is threaded through every in-repo vocabulary that couples to
`ENGINE_ROOT_FIELDS`, including the trusted-launcher strict shape.

Comparing minimal mode against the full persona on the bench is an obvious
next experiment — the persona is exactly the kind of thing this project
measures — and the harness vocabularies are not ready for it.

## What is missing

- **The ejected harness.** `arenabench/` left this repo in #4642
  (`macanderson/arenabench`). Its `harbor_agent.py` carries its own
  `_ENGINE_ROOT_FIELDS` frozenset used to validate a posture, and it does not
  list `minimal_prompt` — PR #4601 added it to the in-repo copy, but that copy
  was deleted by the eject before the PR landed, so the mirror in the ejected
  repo never got it. A posture naming the key is refused there as unknown.
- **`bench/harbor_adapter`.** `stella_harbor/secure_launcher.py`'s
  `_ENGINE_POSTURE_FIELDS` (the disclosed-posture shape) does not list it
  either, and `stella_harbor/posture.py` does not write it. Nothing refuses a
  run today, but an arm that wants to pin the mode has nowhere to declare it.

Neither is a break right now: no posture writes the key, and the bench job is
green on #4601.

## Done when

A benchmark arm can pin `minimal_prompt` in its posture and have it survive
both the posture writer and the launcher's vocabulary check, with the digest
consequences understood — adding a key to a posture re-hashes it, and
`bench/READINESS.md` §8.4 registers digests that
`test_the_registered_sonnet_digest_is_unchanged` asserts, so this is a
published-numbers decision, not a mechanical sync (the same reasoning
`settings::unknown`'s `RETIRED_ENGINE_ROOT` records for the retired role keys).

## Noticed alongside

`_ENGINE_POSTURE_FIELDS` in `bench/harbor_adapter/stella_harbor/secure_launcher.py`
is declared and referenced nowhere in the repository (`rg _ENGINE_POSTURE_FIELDS`
finds only its own definition). Either it should be enforcing the posture shape
or it should go; it predates #4601 and is noted here rather than filed twice.

### LINKED ISSUE #4659: /model default writes a model the live switch would refuse: allowed_models is unchecked on the persist path
## Problem

#4617 added a hard `allowed_models` refusal to the **session switch**, but the **persist** path never learned it. The two halves of one command now disagree:

```
/model zai/glm-5.2          → refused when zai/glm-5.2 is outside [models].allowed
/model default zai/glm-5.2  → written to ~/.stella/settings.json regardless
```

## Evidence

`crates/stella-cli/src/command_deck/session_override.rs` (`switch_session_model`) refuses:

```rust
let allowed = crate::settings::Settings::load(&cfg.workspace_root)
    .ok().and_then(|s| s.agent_engine_config)
    .map(|e| e.allowed_models().to_vec()).unwrap_or_default();
let full_spec = format!("{}/{}", spec.provider, spec.model);
if !allowed.is_empty() && !allowed.iter().any(|a| a == &full_spec || a == id) {
    return Err(format!("`{full_spec}` is not in this workspace's allowed model list …"));
}
```

`crates/stella-cli/src/command_deck/model_cmd.rs` (`set_default_model`) does not — it checks only `engine_config::parse_model_spec` and `settings_check::check_resolved_model`:

```sh
rg -c allowed_models crates/stella-cli/src/command_deck/model_cmd.rs   # 0
```

## Why it matters

The refusal's own comment gives the reason it should apply here too: *"the typed form must answer to the same list or the restriction is only as strong as the user's memory of it."* A default is the **stronger** of the two — a session override dies with the session, while a default silently seeds every session started afterwards. So the path with the longer blast radius is the one still unguarded.

Note the scopes differ, and that is the design decision to make: `allowed_models` merges per-scope and *replaces wholesale*, so a project's list restricting a workspace should plausibly not veto a write to the **user-global** default. Either answer is defensible; today's is neither, it is an omission.

## Done when

- `set_default_model` either enforces the list (with the same message shape as `switch_session_model`) or documents in-code why the persist path is deliberately exempt.
- A witness test: with `allowed_models = ["zai/glm-5.2"]` configured, `/model default anthropic/claude-opus-5` does not write `default_model`.
- If enforcement lands, decide and state which scope's list governs a user-scope write.

## Constraints

- `set_default_model` deliberately reads `user_engine_config_at` rather than `Settings::load`'s merged view, so that a project's pins are not copied into the user file — an enforcement check must read the list without reintroducing that leak.
- Found while reconciling PR #4629 onto #4617; not fixed there because that PR moves `/model` *out* of its own command route and touching the persist path would have widened it further.

macanderson:
### #4618 + #4659 — the allowed-model ceiling becomes one authority

`allowed_models`' doc comment has claimed since the key existed that the list is the ceiling on seat assignments too. It was not: `resolve_seat_models` checked that the seat's provider was configured and the adapter built, never that the model was on the list. The same list bound `/model <spec>` after #4617 and not `/model default <spec>`, so the two halves of one command disagreed — `/model zai/glm-5.2` was refused while `/model default zai/glm-5.2` wrote the same model into `~/.stella/settings.json` and seeded every session after it.

Both halves were unfixable by copying the check, because the check is a claim the setting's doc comment makes and one copy per surface is one place per surface for the claim to stop being true. `crates/stella-cli/src/settings/allowed_models.rs` is now the matching rule (`admits`) and the refusal sentence (`denial`); the session switch, the seat plane and the persist path all call it.

Decisions the issues asked for, and where they are stated in code:

- **Which scope governs the persist path.** `set_default_model` reads the **merged** list while still reading and rewriting the user file alone for the write. A project's or an org's restriction can therefore veto a machine-wide default set from inside the restricted workspace, without that scope's pins reaching user scope — the constraint #4659 named. The reason is at the call site.
- **The startup default is deliberately exempt**, and the `allowed_models` doc now says so instead of being silent: `--model`, and a `default_model` read from a settings file, are the configuration this list is part of, and a ceiling that vetoed the file it is written in would make one bad edit unbootable with no in-session way to repair it.
- A seat outside the list degrades with a notice under `EnginePosture::Configured` and **refuses the run** under `Trusted`, which falls out of the existing posture rule rather than being a second one.

### #4621 — `step_usage` records the request's generation params

`AgentEvent::StepUsage` gains two additive optional fields, `temperature` and `params: Option<GenerationParams>`, mirroring `CompletionRequest`'s own two one-for-one. A single purpose-built struct folding them together would be a second vocabulary that stops matching the first time `GenerationParams` grows a field. Both are content-free — numbers and closed enums, nothing that can hold prompt text — so the tool schemas stay off the event and in the wire log, which is what `wire_log.rs`'s module doc now says instead of naming a gap that has closed.

Both emit sites stamp from the request actually dispatched. `settlement::RequestShape` is what makes that structural: the engine's site reads the four ask-side facts off the `CompletionRequestRef` it built, rather than two call sites reading the same engine-config bindings and being trusted to keep agreeing. It carries the four facts and not the whole request, because an emitter that can reach the conversation is one edit away from putting content on a content-free event. `driver.rs` is a god file and this shrinks it by a line.

The Observatory lifts both keys in `journal.rs`'s explicit `step_usage` allowlist and renders a `generation` row on the profile card, which distinguishes "every call pinned nothing, so the provider's defaults decided" from "this journal predates the field".

`JournalRecord::Event` and `FleetMsg::Event` box their `AgentEvent`: it was already several times the size of anything else those enums hold, and the new fields pushed both past clippy's `large_enum_variant` threshold. The box is invisible on the journal wire.

### #4650 — `minimal_prompt` is a selectable bench posture arm

PR #4601 shipped minimal prompt mode and threaded it through every in-repo vocabulary coupled to `ENGINE_ROOT_FIELDS`. The harness vocabularies were not among them, so a benchmark arm had nowhere to declare the mode. `STELLA_MINIMAL_PROMPT` joins `POSTURE_SELECTOR_ENV` (which is what registers it host-only — an unregistered selector is a refused run), resolves through `read_posture_selectors` like every other arm, and reaches `_benchmark_engine_posture` as a keyword defaulting to `None`.

**Digest consequences, which is the published-numbers half of that issue:** unset emits no key, so every digest registered in `bench/READINESS.md` §8.4 keeps describing the posture that produced it and `test_the_registered_sonnet_digest_is_unchanged` still passes. Setting it writes `on` or `off` and moves the digest, which is the point — a run states the persona it measured. `off` is expressible rather than folded into absence, because absence already means "the shipped persona".

`_ENGINE_POSTURE_AGENT_FIELDS` in `secure_launcher.py` loses its `params` entry in the same change: #2411 took the output cap out of the posture, no role has written a `params` key since, and the comment above the set still described the split as live. That paid for the line the new key costs, which the file's size ceiling would otherwise have refused.

## Evidence

Targeted commands, run in this worktree. CI is the authority for the workspace build.

**Witnesses, each run both ways.** The Rust ones were checked by ablating the production change on a scratch commit and re-running, not by reasoning about them:

|---|---|---|
| `agent::seats::tests::a_seat_outside_the_allowed_list_rides_the_session_model` | FAILED | ok |
| `agent::seats::tests::a_trusted_posture_refuses_an_off_list_seat` | FAILED | ok |
| `command_deck::model_cmd::tests::the_persist_path_refuses_what_the_live_switch_refuses` | FAILED | ok |
| `driver::tests::usage_completeness::step_usage_carries_the_requests_generation_params` | FAILED | ok |
| `accounted_call::tests::step_usage_carries_the_requests_generation_params` | FAILED | ok |
| `TestMinimalPromptArm::test_each_setting_declares_itself_and_declares_only_itself` | FAILED (`KeyError`) | ok |
| `TestMinimalPromptArm::test_the_selector_reaches_the_builder_through_the_one_reader` | FAILED (`KeyError`) | ok |

The first three were ablated by removing the `admits` check at the seat plane and at `set_default_model`; the two `step_usage` ones by stamping `None` instead of the request's values at both emit sites; the two Python ones by dropping the posture's `minimal_prompt` emission.

Also added, and passing: `settings::allowed_models::tests` (three), the extended `step_usage_roundtrips_as_a_complete_metering_record` and `step_usage_from_a_pre_request_shape_stream_still_parses` in `stella-protocol`, the Observatory journal route's assertion that the lift allowlist names the two new keys, and `TestMinimalPromptArm`'s unset-digest and launcher-vocabulary arms.

**Suites:** `cargo test -p stella-protocol -p stella-core -p stella-store`, `cargo test -p stella-cli -p stella-tui -p stella-observatory` — both green. `./scripts/bench-suites.sh run` — all seven Python suites green (758 + 270 + 10 + 66 + 92 + 27 + 3).

**Lints and docs:** `cargo clippy --all-targets -- -D warnings` over `stella-cli`, `stella-tui`, `stella-serve`, `stella-observatory`, `stella-core`, `stella-protocol`, `stella-model`, `stella-store`, `stella-fleet`, `stella-runtime`, `stella-engine` — clean. `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps` in both the plain and `--document-private-items` forms over every touched crate — clean.

**Guards:** `make guards-fast` green, `make wire-schema` green. `make prose` green with nothing added to the baseline. `check-file-size` green with no baseline edit.

**The wire diff is additive.** `make wire-schema-update` produced 234 insertions and **zero deletions** across `docs/wire/agentevent.{d.ts,schema.json}` and `docs/wire/serveframe.{d.ts,schema.json}`. `stella-plugin`'s `PROTOCOL_VERSION` is untouched and `docs/wire/wrapper.*` has no diff. `ServiceTier` and `Verbosity` entered the payload graph with `params`, so the wire contract gains a sampled-arm row and a sweep for each — which is the contract's own totality check firing, not a workaround for it.

## Deleted tests

None. `check-deleted-tests` should have nothing to report.

## Not done

Nothing in the batch was skipped. Two things noticed and deliberately not fixed here:

- **`_ENGINE_POSTURE_FIELDS` in `bench/harbor_adapter/stella_harbor/secure_launcher.py` is still declared and referenced nowhere** — #4650's own "Noticed alongside". `minimal_prompt` is added to it so its vocabulary is correct, but nothing reads the set, so the launcher's real fail-closed check remains `ENGINE_ROOT_FIELDS` parsed from `unknown.rs` (which #4601 already updated, and `TestMinimalPromptArm::test_the_key_is_inside_the_launcher_vocabulary` asserts against). Wiring it is not mechanical: `_require_exact_object` would refuse every posture that omits an optional arm key, so it needs a required/optional split like `tb21_posture_schema.py`'s — a design decision, and the file is at its size ceiling.
- **The ejected `arenabench` harness** (`macanderson/arenabench`) carries its own `_ENGINE_ROOT_FIELDS` copy that never got `minimal_prompt`, because the eject in #4642 predated #4601. That lives in another repository and cannot be fixed from this one.

Closes #4618
Closes #4659
Closes #4621
Closes #4650

## Summary by Sourcery

Unify model restrictions across all runtime selection paths, capture dispatched generation settings in usage telemetry, and make minimal-prompt benchmarking explicit and reproducible.

New Features:
- Record request temperature and generation parameters in step-usage events and expose them in Observatory profiles.
- Add minimal-prompt as a selectable, digest-affecting benchmark posture arm.

Bug Fixes:
- Enforce the allowed-model ceiling consistently for session switches, persisted defaults, and seat assignments.
- Preserve memory and wire compatibility while extending usage events and journal records.

Enhancements:
- Centralize allowed-model matching and denial messages in one shared authority.
- Box large agent events in journal and fleet message variants to reduce enum sizes.

Documentation:
- Update wire schemas and user-facing model-setting documentation for the expanded model restrictions and usage fields.

Tests:
- Add coverage for allowed-model enforcement, generation-parameter propagation and round trips, Observatory exposure, and minimal-prompt benchmark behavior.

Chores:
- Remove the obsolete posture agent params vocabulary entry.


macanderson:
## Answering Sourcery's one ❌ row

> **#4650** — Update the ejected `arenabench` harness's independent `_ENGINE_ROOT_FIELDS` vocabulary so postures containing `minimal_prompt` are accepted there. ❌

Correct as a reading of the diff, and it stays out of scope: that vocabulary lives in `macanderson/arenabench`, a separate repository since #4642. No commit on this branch can reach it, so the row cannot be satisfied here by any change to this PR.

Filed as the handoff it needs rather than left as a gap: **macanderson/arenabench#1**. It carries the current `ENGINE_ROOT_FIELDS` list from `crates/stella-cli/src/settings/unknown.rs` to sync against (the whole list, not just this key — the eject stranded every key Stella added after it), the omit-when-unset rule that keeps registered digests describing the postures that produced them, and a done-when covering all three states.

Nothing is blocked in the meantime. `arenabench` refuses a posture that names the key; no posture writes it, so no run refuses today, and the arm this PR makes expressible runs through `bench/harbor_adapter` — the path stella's own bench job uses.

Issue #3567, opened 2026-08-17
A plugin cannot ship its own external tool server, and there's no way to tell which plugin added a given skill or rule


macanderson:
## Problem

Two residues of #3380's loader half, both "the surface works, the
attribution or the coverage is short".

### 1. The fourth surface is missing: `.stella/mcp.toml`

AGENTS.md's `.stella/` table names four ways a workspace extends the agent:
script tools, skills, context records, and **MCP servers**. #3380 wired the
first three into the plugin package (`<plugin_dir>/tools`, `skills`,
`rules`). A plugin still cannot ship an MCP server, so the most natural
packaging of a vendor integration — "install this plugin, get its MCP tools"
— is the one thing a package cannot do.

This is not merely a missing directory. An MCP server is a process the host
connects to, and its tools enter the registry as `mcp__server__tool` names
that go through `contracts::unknown_contract` at `High`. The open question
is what principal those calls authorize as: they are the plugin's, by the
same argument `CustomTool::principal` makes, but the map that carries that
answer (`GatedToolSet::with_tool_principals`) is keyed by tool name and is
populated only from `Vec<CustomTool>` today.

Files: `crates/stella-cli/src/agent.rs` (`load_mcp_plan`),
`crates/stella-cli/src/plugin_cmd/package.rs` (the three constants and
`dirs_of`), `crates/stella-tools/src/gated.rs`
(`with_tool_principals`), `crates/stella-mcp/`.

### 2. A contributed skill or record is attributable only by path

A contributed **tool** carries `CustomTool::contributed_by`, so "which plugin
gave me this?" is a field. A contributed **skill** carries only
`Skill::source_path`, and a contributed **record** only
`LoadedRecord::source` — both of which happen to lie inside the package
directory, so the answer is recoverable by parsing a path string. That is
exactly the shape the repo treats as a defect elsewhere (a caller reaching a
decision via `err.contains(...)`).

Nothing is broken today: retraction is structural (the contributions are
derived from the package on every load, so deleting the package deletes
them), which is why this is attribution and not correctness. But
`stella context explain` cannot say "this record came from the `vera`
plugin", and the SKILLS tab cannot mark a plugin's skill as a third party's.

Files: `crates/stella-cli/src/memory/skill_files.rs`
(`append_plugin_skills`), `crates/stella-cli/src/context_records.rs`
(`plugin_first`, `rule_files`), `crates/stella-core/src/skills.rs`
[...]

Issue #3542, opened 2026-08-17
A research stage has no way to tell later stages what it found


macanderson:
## Problem

`BeforeTurnResponse::publish` exists so a stage can publish signals "for later
stages of the same turn to read", and `admissible`
(`crates/stella-runtime/src/wrapper/mod.rs`) checks the value's *kind*. What it
cannot check — because the vocabulary does not support the question — is
whether the publishing stage is one that publishes that signal at all.

`StageName::Research.publishes()` is `&[]`
(`crates/stella-plugin/src/wrapper.rs`), and every entry in `Signal` belongs to
triage, execute, witness or verify. So `plugins/stella-research` publishes
nothing, and it is not a shortcut: there is no signal it could publish that
would be true. "How many groundings did research find" and "did research find
the symbol the goal names" are exactly the facts a later stage would branch on,
and neither is expressible.

This is the `doc:pipeline-as-plugins` §A8 tail ("Signal is 5 values and only
Triage publishes any" — it is 15 now, and the stages that publish are still
four), pointed at the stage Track B extracted first.

## The second half: a published value has nowhere to go

`BeforeTurnResponse::publish`'s own doc calls this out as a declared gap:
`Wrapper::resolve` takes one up-front `SignalValues` snapshot, so a value a
stage publishes mid-turn cannot reach the condition that reads it (#3491). Both
halves have to close for a publication to mean anything — a signal a research
stage may publish is useless while nothing can read it, and vice versa.

## Definition of done

- Either `Signal` grows the research-published facts (each transcribing a live
  branch, per the rule the enum's doc comment sets), and
  `StageName::Research.publishes()` names them; or the impossibility is written
  down where a plugin author meets it.
- `plugins/stella-research` publishes them, and
  `crates/stella-runtime/tests/research_plugin_conformance.rs` stops asserting
  `response.publish.is_empty()`.
- #3491 closes alongside, or the new signals are documented as unreadable until
  it does.

Refs #3380, #3491.

macanderson:
Dead-end recorded (from PR #4529, which skipped this): neither of this issue's options closes honestly today. Growing `Signal` with the research-published facts forces all 41 `SignalValues` construction sites to answer two questions no host can currently produce; the doc-only branch contradicts the DoD's second bullet. The honest precondition is a producer for the facts — which is the same question as #4455's audit (what publishes research facts post-#3865). Suggest folding this into that audit's (c) bucket. _From the 2026-08-23 backlog sweep._

[...]

Issue #3491, opened 2026-08-17
wrapper resolution is one up-front snapshot, so a condition on an execute/witness/verify signal reads a value that stage has not produced


macanderson:
## Problem

A8 (#3380 track A) grew `Signal` to cover the facts the pipeline actually
branches on, including ones published *after* triage:

| signal | publisher | live branch it transcribes |
|---|---|---|
| `mutating-actions`, `diff-lines` | `execute` | `LadderInputs::nothing_was_attempted`, the diff-budget conjunct of the `SubmitFast` rung (`crates/stella-pipeline/src/verify.rs`) |
| `witness-authored` | `witness` | `witness.is_some()` in `Pipeline::verify_candidate` |
| `flip-achieved`, `tests-red`, `tests-green` | `verify` | `flip.is_achieved()`, `touched_tests_passed == Some(false)` / `Some(true)` in `ladder_decision` |

The load-time graph check (`Wrapper::validate`) makes sure a condition only
reads a signal an **earlier** stage publishes, and `Wrapper::resolve` refuses
to read one whose publisher was skipped (`SignalNotProduced`). Both rules are
*structural*: they bound which reads are legal in the declared order.

Neither rule makes the value **fresh**. `stella_plugin::SignalValues` is one
snapshot of every signal, and `Wrapper::resolve` is a single up-front pass over
it (`crates/stella-plugin/src/program.rs`). So a host that resolves the whole
program before the first stage runs — which is exactly what
`stella_pipeline::variant::signals` does, at the triage boundary — answers a
`diff-lines > 0` condition from a snapshot taken before anything executed. The
answer is not an error and not a refusal: it is silently "nothing happened".

Today this cannot bite, and that is asserted rather than assumed:
`crates/stella-pipeline/tests/variant_program.rs::the_shipped_variant_reads_only_facts_the_triage_boundary_has`
fails if the shipped `classic` manifest ever reaches for a post-triage signal.
A hand-written variant has no such guard.

## Files

- `crates/stella-plugin/src/program.rs` — `SignalValues`, `Wrapper::resolve`.
- `crates/stella-plugin/src/wrapper.rs` — `Signal::publisher`, `Wrapper::validate`.
- `crates/stella-pipeline/src/variant.rs` — `signals()`, whose doc comment
  states the triage-boundary contract this issue exists to remove.
- `crates/stella-pipeline/tests/variant_program.rs` — the guard above.

## What "done" looks like

[...]

Issue #3482, opened 2026-08-17
A plugin's declared permissions are shown to the user but never enforced


macanderson:
## Problem

`doc:pipeline-as-plugins` §4 A1 shipped the **vocabulary** for plugin authority and its consent rendering, deliberately stopping short of the loader (§A4). The residue is two unwired seams, both of which are safe today only because no loader exists:

1. **`stella_core::ports::Principal::Plugin(String)`** (`crates/stella-core/src/ports/authz.rs`) is constructed nowhere in the workspace. Every other variant has a real construction site (`crates/stella-cli/src/subsession.rs:732`, `crates/stella-cli/src/fleet_cmd.rs:729`, `crates/stella-cli/src/candidate_ws.rs:498`, `crates/stella-serve/src/routes.rs:118`); this one does not, because the thing that would construct it — a plugin loader — is §A4.

2. **A declared capability is shown but not enforced.** `Capability` (`crates/stella-plugin/src/consent.rs`) carries `tool`, `risk`, `purpose` and `scope`, and `consent_text` renders all four. Nothing turns an accepted consent into an `AuthzGate` rule. `consent_text` already discloses the half of this that a user could otherwise be misled by — it labels `scope` as *the plugin's claim* and states that the gate enforces the tool and the grade, not the claim — but "the gate enforces the tool and the grade" is currently true only in the sense that no plugin can call anything at all.

## Why it matters

The whole point of `Principal::Plugin` is that a gate can refuse an installed plugin independently of its operator. Until a loader constructs it and installs a gate keyed on it, an installed plugin would run as whatever principal the host already passes — which for `stella-cli` today is the constant `Principal::User` (`crates/stella-cli/src/agent.rs:362`, `:1657`, `:1672`). That is precisely the marketplace defect §A1 exists to prevent, and it will be introduced silently by the first loader that forgets.

## Files involved

- `crates/stella-core/src/ports/authz.rs` — `Principal::Plugin`, `AuthzGate`, `RiskCeiling`.
- `crates/stella-plugin/src/consent.rs` — `Capability`, `highest_risk`, `consent_text`, `validate_capabilities`.
- `crates/stella-plugin/src/manifest.rs` — `PluginManifest::capabilities`.
- `crates/stella-plugin/tests/install_consent.rs`, `crates/stella-plugin/tests/fixtures/self-driving.toml` — the widest grant, written out.
- `crates/stella-cli/src/agent/tool_stack.rs` — where a session's principal and gate are bound today.
[...]

PR #4684, opened 2026-08-24
feat(stella-cli): a plugin's consent leaves a receipt, and a manifest that drifts from it stops loading

### LINKED ISSUE #3514: A plugin can quietly grant itself more power after install, and the consent screen doesn't show the model tier or the data it receives
## Consent defects found by the authority lens of the socket audit

They share a root: **install records nothing about what was consented to**, and the consent text describes fewer powers than the plugin actually receives.

---

### 1. Self-escalation — a plugin rewrites its own grant

A user installs `vera` as `participation = "observer"`, no hooks, `env = []`, and consents to that text.

`vera`'s process — which runs as the user and owns its install directory — writes `~/.stella/plugins/vera/plugin.toml` with `participation = "arbiter"`, `hooks = ["PreToolUse","Stop"]`, `env = ["GITHUB_TOKEN"]`.

Next run, `read_tier` (`crates/stella-cli/src/plugin_cmd/roster.rs:263-273`) parses the new file and `hook_routes()` emits the new routes. `LoopGrant::permits_hook` — "the authoritative filter" — faithfully authorises grants **no human ever saw**. `install` refusing to overwrite an existing directory does not help, because nothing re-enters `install`.

**Fix:** write a **consent receipt** at install (digest of the consented `plugin.toml`, scope, timestamp); `read_tier` refuses — with a notice pointing at re-consent — any plugin whose manifest no longer matches its receipt. This also fixes the deeper problem that "installed" currently means nothing more than "a directory with a `plugin.toml` exists" (see #3509).

---

### 2. The consent prompt never shows the model tier

`consent_text` renders **no `[roles]` information at all**. A manifest with `[roles.triage] tier = "premium"` renders one line — *"spends model calls on these stages as child turns: triage, verify"* — and the word `premium` never appears.

Meanwhile `WrapperError::UndeclaredRole` justifies its refusal with *"the roles a human consented to at install are the roles a wrapper may spend"* — citing a consent document that **does not contain them**. The user is never shown which tier their money goes to.

**Fix:** render each `[roles.<name>]` and its `tier` in `loop_say`, so the enforcement cites text the user actually read.

---

### 3. The consent prompt never discloses what data leaves the process

The document enumerates *powers* (grade, hooks, argv, env, capabilities) but never the *data* the plugin's process receives. The wire contract ships, on every turn:

- `goal` — the user's prompt (`crates/stella-plugin/src/wire.rs:188`)
- `turn.answer` — the model's full reply (`wire.rs:360`)
- `turn.tools`, `turn.changed_files` (`wire.rs:375`)

as JSON on a third party's stdin. A plugin with no `[[capabilities]]` renders *"It asks for no tool capabilities."* — which reads as harmless.

**A prompt-exfiltrating plugin is indistinguishable from a benign one at the consent prompt.** For a BYOK, no-telemetry product whose central promise is that prompts never leave the machine, this is the disclosure that matters most.

**Fix:** a fixed disclosure block in `consent_text` for any manifest carrying `[wrapper]` or `[runtime]`, naming exactly what crosses to the child at each granted point — **derived from the wire types** so it cannot drift as fields are added.

---

## Definition of done

- Witness: a plugin whose manifest changed after install is refused until re-consented.
- Witness: `consent_text` for a manifest with `[roles]` contains the tier; for one with `[wrapper]`, contains the data-disclosure block.
- Both fail today.

Found by the blocking three-lens socket audit (authority lens). Related: #3509, #3512. Refs #3380.

macanderson:
### #3514 — a plugin can widen its own grant by rewriting its manifest

`stella plugin install` rendered `consent_text`, asked, and copied — and
recorded nothing about *which* manifest the answer was about. A plugin's
process runs as the user and owns its install directory, so it could rewrite
its own `plugin.toml`: an `observer` with no hooks becomes an `arbiter` with
`hooks = ["PreToolUse", "Stop"]` and `env = ["GITHUB_TOKEN"]`, the next
session's `read_tier` parses the new file, and `LoopGrant::permits_hook` — "the
authoritative filter" — authorises a grant no human ever saw. Refusing to
overwrite an existing directory does not reach this, because nothing re-enters
`install`.

New `crates/stella-cli/src/plugin_cmd/receipt.rs` files the SHA-256 of the
`plugin.toml` bytes the document was rendered from, with the tier and a
timestamp, into `<tier>/.consent/<package>.json`. `roster::read_tier` re-checks
every manifest against its receipt on every load and `PluginRoster::compose`
drops anything that no longer matches, with a notice naming the two commands
that re-consent it. `remove` drops the receipt with the package; a failed
install rolls both back.

Three decisions worth a reviewer's attention, each argued in the module header
and in `doc:pipeline-as-plugins` §A4:

* **The two tiers answer differently about a package carrying no receipt at
  all.** Nothing arrives in `~/.stella/plugins` except through install, so a
  missing receipt there is one that went missing and the package is withheld.
  `<workspace>/.stella/plugins` also holds what a `git clone` carried in —
  shipping a plugin with a repository is supported and `STELLA_TRUST_PROJECT`
  is that package's consent transaction — so an unreceipted project package
  still loads behind that gate, while a receipted one must still match.
* **The drop happens in `compose`, not in `read_tier`.** `stella plugin remove`
  and install's duplicate-name check read a tier directly, and a package they
  cannot see is one that cannot be uninstalled — the failure `remove`'s own docs
  exist to prevent. A drifted package is refused a load and stays removable, and
  the round trip is asserted.
* **A receipt is not a sandbox.** A plugin can delete its own as easily as it
  can rewrite its manifest. What the mechanism buys is that the two files must
  be kept in agreement and only the code path that shows a human the document
  writes one, so self-escalation becomes a refusal a user can see.

Install also re-reads the copied manifest and refuses unless it is byte-identical
to the one the document described, so a source directory rewritten between the
prompt and the copy cannot install a grant nobody was shown.

### Note on the `stella-tui` repair this branch briefly carried

`main` did not compile when this branch was cut: #4673 and #4674 each boxed the
`FleetMsg::Event` literals their own branch could see, both were green, and the
merge was textually clean because neither side contained the other's lines — the
shared-cell composition break AGENTS.md describes. The file did not parse, so
`cargo fmt --all --check` could not read the tree and nothing here could be
verified. The one-character repair was committed on this branch to get past it.

#4681 and #4682 have since landed the same repair on `main`, so it composes to
identical content and **no longer appears in this PR's diff** — the net diff
against `main` is the six files below.

## Evidence

Targeted only — no workspace build on this machine.

* `cargo test -p stella-cli --bin stella plugin` — **128 passed, 0 failed**,
  including the two new witnesses and five `receipt` unit tests.
* **Witness, both ways.** With `ConsentState::admits` ablated to return `true`
  for every state (the pre-change behaviour: no receipt enforcement), both
  witnesses fail —
  `a_plugin_that_widens_its_own_manifest_after_install_stops_loading` on
  *"a manifest nobody consented to must not be in force"* and
  `a_user_tier_package_with_no_consent_receipt_is_not_loaded` on
  *"no consent, no plugin"*. With the enforcement restored, both pass. Each
  test also carries its own anti-vacuity assertion: the consented package loads
  and routes exactly what was declared.
* `cargo fmt --all --check` — clean (after the `stella-tui` repair; it could not
  parse the tree before).
* `make prose` — OK, none added. `make doc-links` — OK, 1077 citations resolve.
* `cargo clippy -p stella-cli --all-targets -- -D warnings`, and
  `RUSTDOCFLAGS="-D warnings" cargo doc -p stella-cli --no-deps` in both the
  plain and `--document-private-items` forms — results in a follow-up comment
  when the run finishes; this box is at load average 47 and each pass is a full
  `stella-cli` rebuild.
* `git merge origin/main` before pushing.

No test was deleted.

## Not done

* **#3482 — a plugin's declared permissions are shown but never enforced.**
  **Already fixed on `main`**, by #4512 (`a21354d04`), after the triage note was
  written. `crates/stella-cli/src/plugin_authz.rs` turns each accepted
  `[[capabilities]]` list into a `PluginCapabilityGate`;
  `agent::tool_stack::session_gate` binds it, `NoAuthz` by name only when no
  installed plugin declared a capability. All four definition-of-done bullets
  are met and witnessed — `a_tool_absent_from_the_accepted_list_is_denied`,
  `a_tool_graded_above_what_was_accepted_is_denied`,
  `the_shipped_session_gate_enforces_an_installed_plugins_declared_grant`,
  `only_declared_hooks_are_routed_and_the_caller_is_the_plugin` (which asserts
  `assert_ne!(route.principal, Principal::User)`); `scope` is decided
  explicitly as option (b) in `plugin_authz.rs`'s header; and uninstall removes
  the grant structurally, because the roster is recomputed from disk. Nothing
  is left to do, so nothing is claimed here — it wants closing rather than
  fixing.

* **#3567 — a plugin cannot contribute MCP servers, and a contributed
  skill/record carries no machine-readable provenance.** Not attempted. Its two
  halves share nothing but an origin PR and each is a PR of its own. The MCP
  half needs a `[[mcp]]` declaration in the manifest (so `consent_text` and
  `PluginManifest::reconcile` can cover it), a fourth constant and inventory
  field beside `TOOLS_DIR`/`SKILLS_DIR`/`RECORDS_DIR`, a merge in
  `agent::load_mcp_plan`, and an answer to the open principal question — MCP
  tool names are only known after a connection, so
  `GatedToolSet::with_tool_principals`, which is populated from
  `Vec<CustomTool>` and keyed by tool name, cannot carry `Principal::Plugin`
  for them without a new population point. The provenance half needs a
  `contributed_by` field on `stella_core::skills::Skill` and on
  `records::LoadedRecord`, stamped the way `LoadedRecord::trust` already is
  from the directory a file was read out of — but `context_records::plugin_first`
  flattens the contributed directories to `Vec<String>` at the boundary, so the
  plugin name is discarded before the registry loader could stamp it. Neither
  fits honestly beside a security change in one PR.

* **#3542 — a research stage has no signal it may honestly publish.** Not
  attempted: it needs a maintainer's decision that this session should not take
  silently. The issue offers two options and neither closes it as written.
  Option (a) — grow `Signal` with `groundings` and `goal-symbol-found` —
  contradicts the enum's own stated rule that *"every signal listed here
  transcribes a live branch"*: nothing in this tree would branch on either, and
  `dispatch.rs:565-571` still records that `Wrapper::resolve` takes one up-front
  `SignalValues` snapshot, so a mid-turn publication cannot reach a manifest
  condition at all. That citation names #3491, which is **closed**, and its
  successor has not been filed — so the gap the new signals would land in is
  currently untracked. Option (b) — write the impossibility down — satisfies
  the first definition-of-done bullet and cannot satisfy the other two, which
  presume (a). Separately, the issue's own opening observation is true and
  fixable on its own: `stella_runtime::wrapper::admissible` checks a published
  signal's *kind* and never whether the publishing stage may publish it, and it
  has no production caller today — only tests.

Closes #3514

## Summary by Sourcery

Record plugin consent and refuse to load manifests that no longer match the grant a user approved.

New Features:
- Record plugin consent as a receipt containing the accepted manifest fingerprint, scope, and timestamp.
- Reject plugins whose manifests drift from their recorded consent and report how to re-consent them.

Bug Fixes:
- Prevent plugins from widening their effective grants by rewriting an installed manifest after consent.
- Prevent user-tier plugins without consent receipts from loading while preserving supported project-tier plugins gated by project trust.
- Ensure failed installs and removals clean up associated consent receipts.

Enhancements:
- Preserve direct tier visibility so drifted plugins remain removable while excluded from the composed runtime roster.
- Verify the installed manifest remains identical to the consented source before finalizing installation.

Documentation:
- Document plugin consent receipts, manifest drift handling, and the differing receipt policies for user and project plugin tiers.

Tests:
- Add unit and integration coverage for receipt round trips, scope/name mismatches, corrupt or missing receipts, manifest self-escalation, and continued removability.


macanderson:
Answering the two `❌` rows in Sourcery's assessment against #3514. Both are
about objectives that **already shipped on `main`** before this branch existed,
so there is nothing in this diff to point at — the assessment reads the diff,
and the tree already satisfies them.

**Row 2 — "Show each declared model role and its tier in the installation
consent text."** Shipped. `crates/stella-plugin/src/consent.rs` renders each
`[roles.<name>]` and the tier it asks to be routed at, under a section whose
own doc comment states the defect it closes (*"that cited a document which
named the stages and nothing about what they cost, so a user consented to a
stage list without being told any of it spends at a premium tier — on their own
BYOK key"*). The witness is `the_consent_prompt_names_the_tier_a_role_spends_at`
in the same file, labelled in-tree as the #3514 tier witness.

**Row 3 — "Disclose in the consent prompt the data sent to plugin
processes."** Shipped, and shipped in the form the issue asked for — derived
from the wire types rather than hand-written. `data_that_leaves_the_process`
in the same file reads `WIRE_FIELDS`, and `every_wire_field_is_in_the_table`
(`crates/stella-plugin/src/wire.rs`) destructures the wire types against that
table, so a field added to the wire and disclosed to nobody fails a test. The
block is filtered by `LoopGrant::permits_point`, so it names what actually
crosses at each granted point and stays empty for a manifest that answers no
point.

What was left of #3514 was defect 1 — install recording nothing about *which*
manifest the answer was about — and that is what this PR closes. `Closes #3514`
is therefore the right trailer: with this merged, all three defects the issue
names are fixed.

macanderson:
Local verification, on the merged head (`b8036375f`), targeted per AGENTS.md's
"CI builds and tests; this laptop does not":

| Command | Result |
| --- | --- |
| `cargo test -p stella-cli --bin stella` | **2250 passed, 0 failed** |
| `cargo clippy -p stella-cli --all-targets -- -D warnings` | exit 0, no warnings |
| `RUSTDOCFLAGS="-D warnings" cargo doc -p stella-cli --no-deps` | exit 0 |
| `RUSTDOCFLAGS="-D warnings" cargo doc -p stella-cli --no-deps --document-private-items` | exit 0 |
| `make guards-fast` | exit 0 |
| `make prose` / `make doc-links` | OK, none added / 1078 citations resolve |
| `cargo fmt --all --check` | clean |

The witness pair was also run against an ablated `ConsentState::admits` — the
pre-change behaviour, where every state admits — and both fail there:
`a_plugin_that_widens_its_own_manifest_after_install_stops_loading` on
*"a manifest nobody consented to must not be in force"*, and
`a_user_tier_package_with_no_consent_receipt_is_not_loaded` on
*"no consent, no plugin"*. Restored, both pass. Each carries its own
anti-vacuity assertion, so neither can pass for want of a package to load.

One follow-up landed after the first push: `load_manifest` became a wrapper with
no production caller once `install` moved to the tuple-returning form, so clippy
reported it dead and the two were folded into one `read_manifest`. Its filename
pin now asserts on the bytes as well, which is what a receipt digests.

macanderson:
`fmt + clippy + test` is red on this PR, and **neither failure is in this
diff** — both are on `main` itself. The cheap control is `git show origin/main:<path>`,
which returns the failing lines verbatim.

**1. `cargo doc -D warnings`** — `crates/stella-tui/src/v2/budget_card.rs:10`:

```
```

`origin/main` carries **two** copies of this module: `src/v2/budget_card.rs`
and `src/views/budget_card.rs`. One branch moved the file and another kept
editing it in place; both were green, and the merge is textually clean because
neither side contains the other's lines. Which copy survives — and what
`v2.rs` and `views.rs` should declare — is a decision, not a keystroke, so it
is reported rather than guessed at here.

**2. `cargo test`** — `crates/stella-runtime/tests/no_ambient_reads.rs:71`,
failing on `crates/stella-runtime/src/bin/wrapper-plugin-fixture.rs:76-77`:

```
`RuntimeSpec` field instead
    let inherited = u8::from(std::env::var_os("CARGO_MANIFEST_DIR").is_some());
    let granted = std::env::var("GRANTED").unwrap_or_else(|_| "0".into());
```

The fixture binary is new on `main` (#4675) and the guard is older; the fixture
reads the environment *on purpose*, because proving the socket clears a child's
environment is what it exists to measure. So the question is whether the guard
should see `src/bin/` fixtures at all — the guard's own contract, and again a
decision rather than a repair.

Both are the shared-cell composition shape `main-canary.yml` exists for.
`#4691` is open for a third one (`file-size`, `crates/stella-transcript/src/tests.rs`
at 1523 lines) and names neither of these, so both want filing.

This PR's own crate is clean, verified targeted on the merged head — see the
evidence comment above: `cargo test -p stella-cli --bin stella` 2250 passed,
clippy `-D warnings` and both rustdoc forms exit 0, `make guards-fast`,
`make prose` and `cargo fmt --all --check` all clean. It will go green on CI
once `main` does.

macanderson:
The CI run is the better evidence, and it has it — the workspace `cargo test`
step got all the way through `stella-cli` before the inherited failure, so both
witnesses are confirmed on a merge with `main`'s tip:

```
test plugin_cmd::tests::a_plugin_that_widens_its_own_manifest_after_install_stops_loading ... ok
test plugin_cmd::tests::a_user_tier_package_with_no_consent_receipt_is_not_loaded ... ok
```

Every `stella-cli` integration test passed there too
(`self_driving_consent`, `goal_wrapped_dispatch_cli`,
`verification_flags_require_pipeline_cli`, …), and the whole workspace run
contains exactly **one** `test result: FAILED` — `no_ambient_reads`, the
inherited one above. Nothing this PR touches is in it.

Issue #3577, opened 2026-08-17
Edited-file diffs in the transcript show fragment line numbers instead of the file's real line numbers


macanderson:
## Problem

`crates/stella-observatory/src/transcript_view.rs` folds an execution's journal into the shared transcript model (`stella-transcript`). For a file-mutation call it builds the `FileChange` from the tool call's own arguments, because the store holds no pre-image of the file:

- `write_file` carries `content`, so its all-green diff is **exact**.
- `delete_file` carries neither side, so it renders a header with no contents.
- `edit_file` carries only `path`/`old_string`/`new_string`, so the rendered diff is of **the replaced fragment**, not of the file. Its dual line-number gutters are therefore fragment-relative and do not correspond to lines in the file.

That last one is the defect. A reader looking at `@@ -1,1 +1,1 @@` on a 400-line file will reasonably read those numbers as file lines.

## Files

- `crates/stella-observatory/src/transcript_view.rs` — `files_from_input`, and the module header's "What the journal cannot tell us" section.
- `crates/stella-tools/src/edit.rs` — where the pre-image is in hand at call time.
- `crates/stella-store/` — where a pre-image (or a precomputed unified diff) would be persisted.
- `crates/stella-transcript/src/file_diff.rs` — the consumer; needs no change if the pre-image arrives as `FileChange::before`.

## Constraints already discovered

- The observatory must **not** link `stella-store` in production and opens every file `SQLITE_OPEN_READ_ONLY` (see `crates/stella-observatory/Cargo.toml`'s dev-dependency note). Whatever is persisted has to be readable with the hand-written SQL already there.
- Storing a whole pre-image per edit is a real size cost on a long run. Persisting the computed unified diff instead is likely the better trade — `stella-diff` is already a dependency of both crates.
- Architecture rule #3: whatever is stored is local telemetry only and must not become exportable.

## Verify

`crates/stella-observatory/src/transcript_view.rs::tests::an_edit_call_carries_a_diffable_fragment` is the test that pins the current behavior; it should be replaced by one asserting file-absolute line numbers.

## Done when

An `edit_file` step in `/transcript?id=N` renders hunk headers and line-number gutters that match the file's real lines, and the module header no longer has to disclaim them.

PR #4689, opened 2026-08-24
fix(stella-transcript): split the file-kind tests out so tests.rs is back under the file-size ratchet

### LINKED ISSUE #4691: main is red: file-size fails on the merged tree
The post-merge canary found `main` broken at `22961967c`.

Failing: **file-size**

### `file-size`

```
check-file-size: FAILED

worse.
  crates/stella-transcript/src/tests.rs is 1523 lines, over the 1500-line limit
```

canary exists (#3332).

### How to fix

```sh
git checkout main && git pull
```

...then, per failing check:

#### `file-size`

```sh
```

than only CI.

macanderson:
## What

`main` is red on the `file-size` ratchet and #4683 is why: its two new #3577
cases pushed `crates/stella-transcript/src/tests.rs` to 1523 lines, past the
1500-line limit, and that file is not grandfathered. #4683's own run showed the
red before it merged; this is the repair.

```
check-file-size: FAILED
  crates/stella-transcript/src/tests.rs is 1523 lines, over the 1500-line limit
```

The fix is a split, not a baseline entry — the baseline covers files that
predate the guard, and this file is not one of them. The "file kinds" section
moves whole to `crates/stella-transcript/src/tests/file_kinds.rs`, beside the
four sibling test modules already there (`json_reindent`, `fields_projection`,
`frame`, `width`, `tail`). Those tests read as a set: a new file, a deletion, a
mutation digest, and the patch that decides whether a modification's gutter
carries the file's line numbers or the fragment's.

No test is deleted and none changes — the block moves verbatim under a
`use super::*;`, so every name it referenced still resolves. `tests.rs` is 1402
lines after the move.

## Evidence

```
./scripts/check-file-size.sh          passes on this branch
cargo test -p stella-transcript --lib 125 passed; 0 failed   (the same 125)
cargo fmt --all --check               clean
python3 scripts/check-prose.py        OK — none added
```

No witness test: this is a file move under the gate's own limit, and the guard
that failed is the witness.

Closes #4691
Refs #3577

## Summary by Sourcery

Split the file-kind tests into their own module to restore compliance with the file-size guard.

Enhancements:
- Split the file-kind transcript tests into a dedicated module, bringing the main test file back under the file-size limit without changing test coverage.

Tests:
- Preserve the existing file mutation rendering and digest test coverage in the new dedicated test module.

### DIFF
--- a/crates/stella-transcript/src/tests.rs
+++ b/crates/stella-transcript/src/tests.rs
@@ -24,6 +24,9 @@ mod width;
 // `render_turn_tail`'s cutoffs (#4566's fix shape (a)).
 mod tail;
 
+// What each file mutation renders as, and where its line numbers come from.
+mod file_kinds;
+
 use crate::digest::{self, format_cost, format_duration, format_tokens};
 use crate::file_diff::{FileDiff, RowKind};
 use crate::fold::{Command, Cursor, FoldState, Zoom, apply};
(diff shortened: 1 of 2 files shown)


macanderson:
Closing as superseded: your commit is on main. #4719 carried it there — cherry-picked unchanged, together with #4690's fixture move and #4711's doc-link fix.

Why the three had to travel together: they share one `fmt + clippy + test` job, so each of these PRs fixed its own break and then tripped the other two inside that same job. All three sat red simultaneously while each was individually correct, and none could merge first. Combining them was the only ordering that existed.

The split in this PR was the right shape and is what landed — a sibling module, not a baseline entry. #4691 is closed.

PR #4675, opened 2026-08-24
test(stella-runtime): run the wrapper socket on Windows — a portable in-tree plugin, and the Job Object's first test

macanderson:
### #3550 — nothing killed a spawned tree on Windows

The mechanism landed while this batch was queued: #4456 gave
`crates/stella-tools/src/exec.rs` a `#[cfg(windows)]` `GroupKillGuard` over a
Job Object and a `#[cfg(windows)]` `detach_into_own_process_group`. What it did
not have is the half the issue's definition of done actually asks for — "proven
by a test". `windows-check.yml` runs clippy and no tests, and the one test that
observes a surviving grandchild
(`wrapper_transport_limits.rs::a_timed_out_plugin_leaves_no_surviving_grandchild`)
was `#![cfg(unix)]`, so on Windows it compiled to nothing.

That test now runs on the Windows runner. It is the same assertion, made
portable, and it is the evidence the Job Object had none of.

### #3497 — the wrapper socket was unproven on Windows

Option 1 from the issue. `crates/stella-runtime/src/bin/wrapper-plugin-fixture.rs`
is a plugin binary declared as a `[[bin]]` and located with
`env!("CARGO_BIN_EXE_wrapper-plugin-fixture")`, following `stella-mcp`'s
`mcp-fixture-server` exactly (`[package.metadata.dist] dist = false`, so it never
ships).

**It is not an SDK and must not become one.** It links `std` and nothing else —
no `serde`, no `stella-plugin`, no JSON parser. It matches substrings and prints
literals, which is the tooling a `printf`-and-`case` shell script had. A Rust
plugin path with the crate's own wire types one `use` away would make the wire
path second-class in exactly the way `doc:pipeline-as-plugins` §5 commitment 2
forbids; the constraint is stated on the module and in the manifest comment.

`tests/wrapper_socket.rs` and `tests/wrapper_transport_limits.rs` drop their
crate-level gate and drive it. All eight cases the DoD names — participation,
verdict, timeout, non-zero exit, unreadable answer, point mismatch, protocol
version, environment default-deny — now run wherever the workspace compiles.

**One test keeps `#[cfg(unix)]`, with its reason on it.**
`a_plugin_written_in_sh_answers_the_same_socket` drives a real `/bin/sh` script
through a whole turn to the same host verdict. "A plugin need not be written in
Rust" is a claim only a plugin that is not Rust can make, and there is no
`/bin/sh` on Windows to make it with. The DoD explicitly allows this shape.

[...]

Issue #3887, opened 2026-08-19
Nothing stops a failing required check from being merged into main: a PR merged 9 hours after its check failed, with admin enforcement off and the pre-push hook skipped


macanderson:
## What happened

`main` was red for ~14 hours on a **required** status check, reddening every open PR (#3875, fixed by #3885). The mechanism by which it landed is worth recording separately from the breakage, because the breakage is fixed and this is not.

Verified chain, all from the API:

| Fact | Evidence |
|---|---|
| #3854's head `4e9a48d9` ran the required job and **failed** | run [32220376237](https://github.com/macanderson/stella/actions/runs/32220376237), `conclusion: failure`, finished `2026-08-19T05:54:54Z` |
| It failed on `check-design-refs` — the same guard that then reddened main | `gh run view 32220376237 --log-failed` → `check-design-refs: docs/design/ is work-in-flight and must not be cited.` |
| It was merged ~9h later, red | `mergedAt: 2026-08-19T15:01:15Z`, `mergeCommit: c2c110810` |
| Not auto-merge | `autoMergeRequest: null` |
| Branch protection permitted it | `enforce_admins: false` on `main`; required contexts are `fmt + clippy + test`, `cargo deny + cargo audit`, `harbor_adapter + analyzer pytest` |

So the required check did its job — it ran, it went red, and it stayed red and visible for nine hours. Nothing downstream of it was able to act on that.

## Why file it

AGENTS.md already names this exact case:

> with `enforce_admins` off, an admin or auto-merge can still land gate-failing code, and the hook is what catches that on the author's push

The pre-push hook is the designed backstop, and here it did not catch it — the push either bypassed it (`SKIP_GATE=1`, `--no-verify`) or the hook was not installed in that clone (`make hooks` is once-per-clone and easy to miss in a fresh worktree). That leaves **no** layer that actually stops a red required check from reaching `main`:

- branch protection: declines to enforce against admins
- pre-push hook: advisory, per-clone, bypassable
- `main-canary.yml`: post-merge by construction — it reports, it does not prevent

The cost is not theoretical or one-off: a red `main` reds every open PR's required check, so every contributor's signal is destroyed until someone notices and repairs it, and the repair itself has to land past the same red check. That is the second time this shape has been recorded (see #3661, "main is red: #3647 replaced init_workspace's `InitIo` parameter that #3642 had just introduced").

## The decision to make

This is a judgement call for the maintainer, not a defect with one right fix. The options, and what each one costs:

[...]

Issue #3497, opened 2026-08-17
The plugin communication layer has never actually been tested on Windows


macanderson:
## Problem

`crates/stella-runtime/tests/wrapper_socket.rs` — the only end-to-end proof
that the wrapper socket's wire path works — opens with `#[cfg(unix)]`, because
every plugin in it is a `/bin/sh` script. On Windows the file compiles to
nothing and the whole socket is untested, including the parts that have nothing
to do with shell scripts: the stdin/stdout exchange, the concurrent
write/wait, `kill_on_drop` at the timeout, and `env_clear()`.

`SubprocessWrapper` itself (`crates/stella-runtime/src/wrapper/subprocess.rs`)
is written in portable `tokio::process` and is *expected* to work there. That
expectation is exactly the kind of claim this repo does not let stand without
evidence.

## Why it is not simply "add a .cmd fixture"

The plugin has to be something the host can start with an argv and that reads
JSON on stdin. Candidates, in the order worth trying:

1. **A tiny Rust helper binary** declared as a `[[bin]]` or an example in
   `stella-runtime`, and located via `env!("CARGO_BIN_EXE_<name>")`. Portable,
   no interpreter assumption, and it doubles as the reference in-tree plugin a
   docs page can point at. Note it must not become a *privileged* Rust path —
   it speaks the same wire contract a Python plugin would
   (`doc:pipeline-as-plugins` §5 commitment 2).
2. A `python3` script gated on the interpreter being present — rejected as the
   primary route because a skipped test is a green test.

## Definition of done

- The socket's end-to-end tests (participation, verdict, timeout, non-zero
  exit, unreadable answer, point mismatch, protocol version, environment
  default-deny) run on every platform CI builds for.
- `#[cfg(unix)]` is gone, or survives only on the specific cases that are
  genuinely shell-shaped, each with a comment saying why.
- The Windows path is *run*, not argued.

Refs #3380.

macanderson:
Parented to the new epic **#3583 — a spawned process tree is bounded on every platform, from one definition**, created during triage because three open issues (#3549, #3550, #3497) share one outcome and no existing epic covered it.

They are grouped because they share a single design decision — where the platform split lives — and because fixing any one of them leaves the outcome unmet: the `setsid` half unconverged, no tree-kill mechanism on Windows, and the transport's Windows path unproven are three ways the same rule is not met. Deciding the `#[cfg]` boundary once under the epic is cheaper than deciding it three times.

[...]

PR #4690, opened 2026-08-24
fix(stella-runtime): move the plugin fixture out of src/, where no_ambient_reads governs

macanderson:
## What

**`main` is red on `stella-runtime`'s own invariant test, and this is the repair.**

#4675 merged while its `fmt + clippy + test` job was still running. That job then failed, on exactly one thing:

```
test no_ambient_reads ... FAILED
`RuntimeSpec` field instead, so two sessions in one process can disagree about it:

  crates/stella-runtime/src/bin/wrapper-plugin-fixture.rs:76: reads ambient process state (`std::env::var_os`)
  crates/stella-runtime/src/bin/wrapper-plugin-fixture.rs:77: reads ambient process state (`std::env::var`)
```

The guard is right and the fixture is right, and that is the whole of the fix.

`tests/no_ambient_reads.rs` reads every file under `src/` and refuses `std::env::var`, because a process global is the same for all N sessions a host assembles in one process. #4675 put the new plugin fixture at `src/bin/wrapper-plugin-fixture.rs` — the idiomatic Cargo location, and where `stella-mcp`'s `mcp-fixture-server` sits — so the guard read it as library source. But the fixture is a **separate process**, and its `env-probe` mode exists to report the environment the socket handed it; asking a child what environment it received is the only way that test can be made at all.

So the read is essential and the invariant is intact. Two ways out:

1. **Teach the guard to skip `src/bin/`.** Makes the test pass and makes `src/` mean something looser. `no_ambient_reads.rs`'s own module doc warns against exactly this: *"an exemption list is the crack the next violation slips through."*
2. **Move the file out of `src/`.** `src/` stays exactly what the guard means by "the crate's own sources", and the `[[bin]]` path says why.

This takes (2). Cargo auto-discovers only `tests/*.rs` and `tests/*/main.rs`, so `tests/fixtures/wrapper-plugin-fixture.rs` is a bin target and nothing else. `env!("CARGO_BIN_EXE_wrapper-plugin-fixture")` is unchanged, so no test moved and no test changed.

## Evidence

On this branch, from a fresh `main` (`0e8dc14cf`):

```
cargo test -p stella-runtime --test no_ambient_reads \
                             --test wrapper_socket \
                             --test wrapper_transport_limits
  → 1 passed, 11 passed, 3 passed; 0 failed   (exit 0)
```

The first of those is the test that is red on `main` right now. The other two are #4675's subject, re-run to show the move costs them nothing.

`cargo clippy -p stella-runtime --all-targets -- -D warnings`, both `RUSTDOCFLAGS="-D warnings" cargo doc` forms, `cargo fmt --all` and `make prose` are all clean.

## Why it did not fail before the merge

It could have. `no_ambient_reads` is a `cargo test` step, and per CLAUDE.md this laptop runs targeted tests rather than the workspace suite — I ran the two test targets I wrote and not the crate's own invariant test beside them. The lesson is narrower than "run the workspace suite": **a crate with an executable invariant test has one more target a change to that crate must run**, and `stella-runtime`'s is named in its README. I have added it to the targeted command for this crate.

The merge itself is the other half, and #3887 is where that is being asked: #4675 was merged nine minutes before its required job finished. Nothing in the tree can stop that today — `enforce_admins` is off — which is the decision that issue is open on.

## Not done

Nothing deferred. This is one file move plus the three citations of its path.

Refs #3497

## Summary by Sourcery

Move the wrapper plugin fixture under the test fixtures directory to preserve the runtime crate’s no-ambient-reads invariant.

Bug Fixes:
- Move the wrapper plugin fixture out of the runtime source tree so the ambient-read invariant test no longer treats its required environment probing as a library violation.

Enhancements:
- Keep the fixture registered as a dedicated test binary while preserving the existing test executable lookup and behavior.

Documentation:
- Update contributor guidance and runtime documentation to reference the fixture's new test-fixture location and explain its separation from guarded source files.


macanderson:
Blocked on a different red-main cause, not on this diff.

This PR's `fmt + clippy + test` job failed at 15m51s — but in the `cargo doc -D warnings` step, which runs **before** `cargo test`, so `no_ambient_reads` was never reached:

```
```

That is #4686's, unrelated to this change, and it is now #4711 (also `unblocks-main`). This PR should go green once that lands.

The fix here is verified independently in the meantime, on a fresh `main` (`0e8dc14cf`): `cargo test -p stella-runtime --test no_ambient_reads --test wrapper_socket --test wrapper_transport_limits` → 1 + 11 + 3 passed, exit 0. And `windows-check` is green on this head ([job 97550429124](https://github.com/macanderson/stella/actions/runs/32764391560/job/97550429124)) — 10 + 3 passed — so the move costs the Windows coverage #4675 added nothing. 13 of 14 other checks pass.

macanderson:
Heads-up: this and #4711 deadlock each other. This branch's `fmt + clippy + test` fails on #4711's `cargo doc` error (`budget_card` intra-doc-links a `pub(crate)` item) — and because `cargo doc` runs before `cargo test` in that job, the log never reaches the `no_ambient_reads` test this PR repairs. #4711's run fails on *your* break for the mirror reason.

Neither tree is green alone, so branch protection refuses both merges. I've composed the two into **#4716** — your four-file move unmodified, plus #4711's doc fix — so CI can verify one tree containing both. Nothing rewritten; after #4716 lands this PR is empty against `main` and can close as merged-by-composition, your commit intact.

Chose that over `--admin`, which would have landed an unverified tree. Your `file-size` note on #4691 is respected: that half is #4713 and is deliberately not folded into #4716.

macanderson:
Closing as superseded: your commit is on main. #4719 carried it there — cherry-picked unchanged, together with #4689's test-file split and #4711's doc-link fix.

Why the three had to travel together: they share one `fmt + clippy + test` job, so each fixed its own break and then tripped the other two inside it. None could merge first.

Moving the fixture to `tests/fixtures/` rather than relaxing `no_ambient_reads` was the right call and is what landed — `src/` stays exactly what that guard means.

Issue #4676, opened 2026-08-24
Epic: the terminal UI has one rendering generation, not two — finish the v2 port and delete v1


macanderson:
## Outcome

Stella has one terminal-UI rendering generation, not two. Every tab and overlay runs on the SPEC 5 frame (`src/v2/*`); every `src/views/*.rs` module still rendering the pre-SPEC-5 way is ported and deleted; the `v1`/`v2` naming split itself is dissolved once nothing depends on it; and no doc comment, README table, or module header anywhere in `stella-tui` describes the old chrome as a point of comparison. This repo has no customers and no compatibility obligation to the old shape — there is no reason to carry its ghost in prose once the code it explains is gone.

## Where this actually stands (verified against `origin/main`, not inferred from file names)

The split is **not** "these tabs are v2, those are v1" — it's finer-grained than that:

- **Fully v2:** the chrome that wraps every tab unconditionally — tab row, hint row, status bar, pulse row (`v2::frame`, `v2::status_bar`, `v2::pulse`) — plus two overlays, SUB-AGENTS (`v2::subagents`) and the session-switcher (`v2::sessions`).
- **v1 shell over a v2 body:** `DeckTab::Graph` still dispatches through `views::graph::render`, but that function is now a thin wrapper — the actual content call is `crate::v2::graph_tab::render`. The v1 file survives only as an empty-state shim.
- **v1 orchestrator sharing some v2 primitives:** `DeckTab::Session` dispatches through `views::session::render`, which is genuinely v1 — but the row renderer underneath it (`src/render/row.rs`) already imports `v2::transcript::{RAIL_W, rail_span}`, so the transcript body is a v1/v2 hybrid today, not clean either way.
- **Pure v1, zero v2 involvement:** every other tab dispatch in `deck_render.rs` — `DeckTab::Agents`, `::Traces`, `::Files`, `::Skills`, `::Mcp`, `::Issues`, `::Settings` — plus the floating cards (`views::plan_card`, `views::models_card`, `views::budget_card`), `views::picker`, `views::queue_popup`, `views::seats`, `views::question`, `views::dispatch_card`, `views::approval`, `views::linear`, `views::installed`, `views::engine` (+ `engine/tabs.rs`), and `views::mcp::detail`.
[...]

PR #4701, opened 2026-08-24
feat(stella-tui): port the model/agent pickers to the v2 SPEC 5 frame

macanderson:
One slice of #4676 — the model/agent pickers move off `views::cards`' chrome and onto the SPEC 5 frame.

## What changed

`crates/stella-tui/src/views/picker.rs` is deleted; its content moves to `crates/stella-tui/src/v2/picker.rs`, the shape `v2/graph_tab.rs` took. The `ListPicker` state machine and its key fold cross over unchanged — only the drawing is rewritten.

**Frame.** The card was a square-bordered `cards::card_frame` whose title row carried the name, the list position *and* the key hints. It now draws what `v2::sessions` and `v2::subagents` draw: a rounded `BorderType::Rounded` block in `token::BORDER`, name and position in the top border, keys right-aligned in the bottom one. Every colour comes from `stella_tui_theme::token`; the two `render_*` functions share one `render_card` rather than each assembling the card, which is what the v1 file did twice.

**Placement.** `cards::CHROME_BELOW` is a constant six, documented as "trace strip 2 + progress 1 + composer 1 + footer 1 + statline 1" — rows SPEC 5 no longer draws. No constant is right there either: the composer grows with its draft and the status bar takes a second row for an earned cache diagnosis, so a card counted back from the frame's foot lands on the prompt at some composer height. `deck_render.rs` now hands the pickers the **content band** instead of the whole frame, and the card sits on that band's last row. The goldens show it two rows lower, no longer clipping the transcript's last line.

**Accessible mode.** `card_area` already widened the card to the frame there, for the reason its own doc gives — "a labeled record clipped at a float's right border is a record a reader never hears the end of" — but the rows still elided at 62 columns. They now elide against the width the card was actually given. Called out because it is a behaviour change beyond the port, small and in the direction the widening intended.

## Evidence

- `cargo test -p stella-tui` — 1084 lib + 66 integration tests, 0 failed.
- `make guards-fast`, `make doc-links`, `scripts/check-file-size.sh`, `scripts/check-prose.py` — all green.
- Golden frames re-blessed with `BLESS=1 cargo test -p stella-tui --test deck_render_snapshots` and the diff read: only `overlay_model_picker` and `overlay_agent_picker` moved, and only in the ways described above. Row content is byte-identical (`▸ zai/glm-5.2-air  · current`, `▸ reviewer  project  reviews a diff against the house rules`).

Witness tests, all new and all failing before this change because the module they name did not exist:

- `v2::picker::tests::the_model_card_is_the_rounded_v2_frame` — the rounded border, the position in the top border, the keys in the bottom one, the gold marker and the `token::HL` tint on the same row.
- `v2::picker::tests::the_card_sits_on_the_last_row_of_the_band_it_is_given` — two bands of different heights, the card's foot on each band's last row. This is the one the placement change is for.
- `v2::picker::tests::the_agent_card_lists_scope_and_purpose`
- `v2::picker::tests::accessible_mode_spans_the_frame`

## Deleted tests

None. `views/picker.rs`'s three `#[test]` bodies (`the_picker_moves_chooses_and_cancels`, `a_shrunken_or_empty_list_never_chooses_past_the_end`, `the_window_keeps_the_selection_visible`) move to `v2/picker.rs` unchanged and still run.

## Not in scope

`views::cards` itself stays — `plan_card`, `models_card` and `budget_card` still use it, and they are other slices of #4676. `cards::truncate_cols` is still used here: it is grapheme-safe column arithmetic, not chrome, and `v2::subagents` already depends on it.

Refs #4676

## Summary by Sourcery

Port the model and agent pickers to the SPEC 5 v2 frame and anchor them within the content band.

New Features:
- Render the model and agent pickers using the rounded SPEC 5 v2 overlay frame with themed borders, positional labels, keyboard hints, selection markers, and accessible-width support.

Bug Fixes:
- Anchor picker cards to the final row of the content band to prevent overlap with the transcript and prompt as composer or status-bar height changes.
- Elide picker rows against their actual accessible-mode width so content is not unnecessarily truncated.

Enhancements:
- Move picker state, rendering, candidate handling, and tests from the legacy views module into the v2 picker module while preserving picker behavior and shared model/typeahead vocabulary.

Documentation:
- Update TUI documentation and module references to point to the v2 picker implementation.

Tests:
- Add rendering and placement witness tests for v2 picker framing, agent details, and accessible-mode layout.
- Refresh model and agent picker golden snapshots.


macanderson:
## The three red checks on this PR are inherited from `main`, not caused by this diff

This PR's diff is eleven files, all under `crates/stella-tui/`, and none of them is named by any of the three failures. `git diff origin/main --name-only` on the branch head:

```
crates/stella-tui/README.md
crates/stella-tui/src/deck_render.rs
crates/stella-tui/src/deck_ui.rs
crates/stella-tui/src/deck_ui/pickers.rs
crates/stella-tui/src/v2.rs
crates/stella-tui/src/v2/picker.rs
crates/stella-tui/src/views.rs
crates/stella-tui/src/views/engine.rs
crates/stella-tui/src/views/picker.rs                     (deleted)
crates/stella-tui/tests/snapshots/deck/overlay_agent_picker.txt
crates/stella-tui/tests/snapshots/deck/overlay_model_picker.txt
```

**`main is not known-broken` — fail.** The hold workflow is doing its job: #4691 is open, so it refuses. Not a statement about this diff.

**`fmt + clippy + test` → `cargo doc -D warnings` — fail.**
`error: public documentation for budget_card links to private item cards::card_frame`, at `crates/stella-tui/src/v2/budget_card.rs:10`. That file arrived on `main` with #4686 and the paragraph carrying the link is new there — `git show 8d7f692de^:crates/stella-tui/src/views/budget_card.rs` has no such link. Filed as #4714.

**`fmt + clippy + test` → `cargo test` — fail.**
`stella-runtime`'s `no_ambient_reads` names `crates/stella-runtime/src/bin/wrapper-plugin-fixture.rs:76-77`, which arrived with #4675. Already filed as #4709.

`cargo clippy -D warnings`, `docs/wire/`, the release smoke build, `stella context validate`, the self-driving harness and gate-scoping all pass in the same job; `file size ratchet`, `docs guards`, `guard self-tests`, MSRV, `cargo deny`, and the pytest suites pass as separate checks.

Locally on this branch, on the same head:

- `cargo test -p stella-tui --no-fail-fast` — 1070 lib + 66 integration, 0 failed (the deck golden suite included, 20/20).
- `make guards-fast` — exit 0, every guard OK.

This PR is blocked on #4691, #4709 and #4714 landing, not on a change here.

Issue #4240, opened 2026-08-22
The deck's approval card is the generic AskUser card, so read_only and the gate never reach the person deciding


macanderson:
#4220 gave the Command Deck both mid-turn asks. The `ask_question` half got a
purpose-built overlay (`crates/stella-tui/src/views/question.rs`); the
**approval** half was wired through the deck's pre-existing generic `AskUser`
card instead, because that closed the dead-end with one declaration and no
second mechanism.

That works and is not a stub — a `RequireApproval` on the deck now parks on a
real human yes/no. But the card a reviewer sees is worse than the request it is
rendering, and the gap is worth closing deliberately rather than leaving as
the shape somebody inherits.

## The gap

`stella_tools::registry::approval::ApprovalRequest` is a structured value:

```rust
pub struct ApprovalRequest {
    pub tool: String,
    pub read_only: bool,
    pub reason: String,
    pub gate: String,
    pub subject: Option<String>,
}
```

`crates/stella-cli/src/approval.rs::AskUserApprovalResponder::question` flattens
all of that into **one prose line**:

```text
`bash` (rm -rf build/) requires approval: matched rule no-destructive-shell — allow this call?
```

Then `DeckAskUserIo` (`crates/stella-cli/src/command_deck/mid_turn_ask.rs`)
renders it as a generic `AgentEvent::AskUser` card with two options. Three
consequences:

1. **`read_only` never reaches the person.** It is in the request and is exactly
   the field that separates "this reads something" from "this writes something",
   and the rendered card cannot say which.
2. **`gate` never reaches them either** — which rule or hook produced the demand
   is the context that makes a deny defensible.
3. **A long `subject`** (a full command line, a long path) is prose inside a
   sentence rather than its own field, so it wraps or elides mid-token in the
   generic card's layout.

There is also no attribution: an approval raised inside a delegated sub-agent's
call renders identically to the lead's, where `views/question.rs` names the
asker. (`ApprovalRequest` carries no `asker` today, so that half needs the
field before it can be shown — see "Constraints".)

## What done looks like

- An `Inbound::ApprovalAsked(Box<ApprovalRequest>)` /
  `WorkspaceInput::ApprovalAnswered(Box<ApprovalResponse>)` round trip, the
  shape #4220 built for questions
  (`crates/stella-tui/src/envelope.rs`).
- An approval card in `crates/stella-tui/src/views/` — its own file, **not**
  `views/question.rs`: an approval is a yes/no over a decided call, a question
[...]

PR #4702, opened 2026-08-24
feat(stella-tui): port the approval and routing cards to the v2 SPEC 5 frame

macanderson:
Refs #4676 — one slice of the epic that leaves this crate with a single rendering generation.

## What moved

`src/views/approval.rs` → `src/v2/approval.rs`, `src/views/dispatch_card.rs` → `src/v2/dispatch_card.rs`. Both v1 files are gone; the two dispatch sites (`src/deck_render/parked.rs` for the overlay, `src/views/session.rs` for the routing band) point at the new modules, and the three other references to the old path — the `DeckUi` field, the key router's import, and two doc comments — moved with them.

## What changed in the drawing

- The shared frame is the one `v2::sessions` and `v2::subagents` draw: `Clear`, a `BorderType::Rounded` block in `token::BORDER`, the title as spans, key hints on the bottom rule. Every colour comes from `stella_tui_theme::token`; nothing reaches for `crate::theme`'s v1 roles or for `theme::VIOLET`, which had no token.
- Both cards mark themselves with `glyph::GATE` in `token::GOLD_BRIGHT` — the pair `v2::sessions::phase_mark` already gives a session that is waiting on its driver.
- The approval card leaves `views::cards::card_area`. That helper bottom-anchors a float by counting the v1 chrome rows underneath it (`CHROME_BELOW` = trace strip 2 + progress 1 + composer 1 + footer 1 + statline 1), and SPEC 5 draws none of them. It centres over the frame like the other two v2 overlays instead.
- `gate` and `reason` lose the padded label column and read as `key value` — the shape `v2::fields` gives a tool's arguments, so an approval and the call it is about are one table. The `accessible` branch inside the field renderer existed only to undo that column alignment and goes with it; `accessible` still spans the card across the frame.

## What did not change

The key vocabulary, the default-deny cursor, `ApprovalAction`, the `Inbound` contract, and the routing card's three routes. Their witnesses moved with the module and pass unchanged.

## Evidence

- `cargo test -p stella-tui --no-fail-fast` — 1082 lib + every integration target green, including the 13 tests in the two ported modules.
- `RUSTDOCFLAGS="-D warnings" cargo doc -p stella-tui --no-deps --document-private-items` — clean.
- `make guards-fast`, `make prose`, `make doc-links` — green. `check-module-reachability` is what proves both v1 files are actually gone rather than orphaned.
- Two new render witnesses: `the_card_states_the_call_the_gate_and_the_reason` (the tool, the mutating/read-only word, the subject, and both labelled rows survive as words a style-stripped golden can pin — the fields #4240 added because the generic `AskUser` card dropped them) and `the_card_fits_inside_any_frame`.

## Golden frames

`overlay_approval`, `overlay_approval_reason` and `overlay_approval_read_only` are re-blessed, and the diff was read: the card is centred rather than bottom-anchored, the border is rounded, the hints moved from the top-right title to the bottom rule, and the two field rows lost their column padding. Every word the old goldens pinned is still on the card. The routing card has no golden.

## Deleted tests

None. Every `#[test]` in the two files moved with them, under the same names.

## Summary by Sourcery

Port the approval and routing cards to the v2 SPEC 5 rendering generation while preserving their existing interaction contracts and routes.

Enhancements:
- Port the approval and dispatch routing cards to the v2 SPEC 5 rendering frame with rounded borders, shared tokens, gate glyphs, and bottom-aligned key hints.
- Align approval-card placement, field presentation, sizing, and accessible rendering with the v2 overlay conventions.

Tests:
- Add render coverage for approval-card content and rendering within very small frames.
- Refresh approval-card golden snapshots for the v2 frame.

Chores:
- Move the approval and dispatch-card modules from the legacy views namespace into v2 and update their consumers and references.
- Remove the obsolete v1 view modules and update the prose baseline.


macanderson:
Rebased onto `main` after #4686 (the floating-cards port) landed. Three conflicts, all mechanical: the two `mod` lists gain both slices' entries, and #4686's two edits to `views/approval.rs` repointed doc references from `views::plan_card` to `v2::plan_card` inside the `LABEL_W` column this port deletes — so they resolve by taking this branch's version, which has no `plan_card` reference left to repoint. `views::cards` survives on `main`, so the shared `truncate_cols` this port uses is still there.

Re-verified on the new base: `cargo test -p stella-tui --no-fail-fast` green (1068 lib + every integration target, golden frames included, no re-bless needed), `cargo clippy -p stella-tui --all-targets -- -D warnings` clean, `cargo fmt --check` clean.

Issue #4220, opened 2026-08-21
The Command Deck cannot park a turn on a human, so ask_question and approvals both dead-end on Stella's default interactive surface


macanderson:
The Command Deck is the default interactive shell on a TTY, and it is the one interactive surface that cannot ask a human anything mid-turn. `crates/stella-cli/src/rules.rs` has said so in passing for a while — "the deck until it grows an approval card" — and #4212 has now made the cost visible twice over: a deck user calling `ask_question` gets the headless decline ("no driver is attached... make the call yourself"), which is honest but is not the feature.

Both gaps are the same missing mechanism, and should be built once.

## What exists

- `stella-tools/src/registry/approval.rs` — `ApprovalResponder` port, broker, TTL, `attach_approval_responder`.
- `stella-tools/src/registry/question.rs` — `QuestionResponder` port, broker, fairness gate, TTL, `attach_question_responder` (#4212).
- `crates/stella-cli/src/question.rs` — `TtyQuestionResponder`, the plain-TTY implementation. Its three-step flow (step through → note per answer → review/submit/chat/cancel) and its pure parsing (`parse_answer`, `parse_review`) are the behaviour the deck should match; only the rendering and input differ.
- `crates/stella-cli/src/rules.rs::enforce_workspace_rules` — the single `interactive_approvals` boolean that arms both responders. The deck passes `false`.

## What is missing

A way for a responder living outside the render loop to (a) hand the deck a request, (b) block until the deck produces an outcome, and (c) not wedge the deck while it waits. Concretely:

1. An `Inbound` envelope variant carrying the pending request (`stella_protocol::QuestionRequest`, and an `ApprovalRequest` for the approval half).
2. A wizard overlay holding the answer set as it is built. Pure fold, like every other overlay in `crates/stella-tui/src/views/`. Keys, matching the plain-TTY affordances: ↑/↓ move within a question, Tab/⇧Tab switch questions, ⏎ selects, `n` opens the note editor for the focused answer, Esc cancels, and a final review pane with submit / chat-about-this / cancel.
3. A `WorkspaceInput` variant carrying `QuestionOutcome` back, resolved into the responder through a oneshot.
4. `rules.rs` passing `true` for the deck once (2) and (3) exist.

## Constraints

[...]

PR #4706, opened 2026-08-24
feat(stella-tui): port the question overlay to the v2 SPEC 5 frame

macanderson:
`crates/stella-tui/src/views/question.rs` — the `ask_question` wizard a parked turn waits on (#4220) — was one of the pure-v1 modules on #4676's checklist. It is now `crates/stella-tui/src/v2/question.rs`, and the v1 file is gone.

## What moved

The whole module, not just the paint. `QuestionOverlay`, `QuestionMode`, `QuestionAction`, `key`, `ingest` and `render` travel together, which is the shape `src/v2/subagents.rs` already takes for an overlay that owns its own keyboard — the alternative (state in `deck_ui/`, paint in `v2/`) would split a fold whose own doc argues for keeping the envelope mapping beside the state it drives.

## What the port changed

The colour layer. `crate::theme`'s semantic aliases give way to `stella_tui_theme::token` directly, per `src/v2.rs`'s module doc ("Every colour in this module tree comes from `stella_tui_theme::token`"). The values are unchanged, and `crates/stella-tui/src/palette.rs` is where each alias resolves:

| before | after | resolves at |
| --- | --- | --- |
| `theme::TEXT_TERTIARY` | `token::MUTED` | `palette.rs:223` |
| `theme::TEXT_PRIMARY` | `token::TEXT` | `palette.rs:207` |
| `theme::accent()` | `gold()` = `token::GOLD` + bold | `palette.rs:152` |

So the card renders cell for cell and hue for hue what it rendered before.

## Chrome

It keeps drawing its own border. A floating card is not a tab body — the content area `v2::frame` carves out is what it floats *above*, and `src/v2/budget_card.rs` states that rule for the same reason. No new top-level chrome was added and none was stripped.

## Call sites repointed

`src/deck_render/parked.rs` (the render dispatch), `src/deck_ui/parked.rs` (the `QuestionAction` import), `src/deck_ui.rs` (the two `QuestionOverlay` references), plus the intra-doc links in `src/envelope.rs`, `src/views/approval.rs`, and `stella-cli`'s `interactive.rs` / `command_deck/mid_turn_ask.rs`. `rg` finds no remaining reference to `views::question`.

## Evidence

- `cargo test -p stella-tui` — 1066 lib tests + every integration binary green, nothing failed.
- The five deck goldens for this overlay (`overlay_question_single`, `_tabstrip`, `_note`, `_review`, `_accessible`) pass **unblessed**. That is the evidence for the "renders what it rendered before" claim above; asserting it in prose would not be.
- `cargo clippy -p stella-tui --all-targets -- -D warnings` clean.
- `RUSTDOCFLAGS="-D warnings" cargo doc -p stella-tui -p stella-cli --no-deps --document-private-items` clean.
- `make guards-fast` green, `check-prose` included.

No witness test is added: this is a module move plus a colour-namespace swap with no behaviour change, and the fourteen unit tests and five goldens that moved with it are what hold the behaviour still.

## Two things fixed in passing

- **The prose baseline entry did not travel.** `scripts/prose-baseline.txt` grandfathered five constructions at `crates/stella-tui/src/views/question.rs`. A file absent from the baseline is held to zero, so rather than rename the entry to the new path — which grants permission the guard is designed to withdraw — the five are deleted and the baseline line is removed.
- **`src/v2/budget_card.rs` fails `doc-warnings` on the base commit.** Its module doc links `cards::card_frame`, which is `pub(crate)`, and `RUSTDOCFLAGS="-D warnings" cargo doc --document-private-items` (the gate's own `doc-warnings` command, `Makefile:512`) rejects a public doc linking a private item. I did not touch that file otherwise, so this is a pre-existing failure rather than one this branch introduced — stated as an inference from the diagnostic naming a file with an empty diff, not as something I bisected. It is now spelled the way `render`'s own doc already spelled the identical reference.

Refs #4676

---

## Summary by Sourcery

Port the parked-turn question overlay to the v2 TUI frame without changing its behavior or visual output.

Bug Fixes:
- Update documentation references and prose-baseline handling after relocating the question overlay.
- Resolve the existing documentation warning associated with the question overlay's frame reference.

Enhancements:
- Move the parked-turn question overlay into the v2 module tree and expose it through the v2 surface.
- Use v2 theme tokens while preserving the overlay's existing rendering and interaction behavior.
- Repoint deck, CLI, and documentation references to the relocated overlay.

Documentation:
- Update intra-documentation links to reference the v2 question overlay.

Tests:
- Preserve and validate the existing question overlay unit tests and golden rendering coverage during the port.

Chores:
- Remove the obsolete v1 question module and its stale prose-baseline entry.


macanderson:
## Both red checks are inherited from the base, not caused by this diff

**`fmt + clippy + test` — `cargo test` step.** One test fails: `stella-runtime`'s `no_ambient_reads`, on `crates/stella-runtime/src/bin/wrapper-plugin-fixture.rs` lines 76-77. This branch's diff does not reach `stella-runtime` at all (`git diff --name-only origin/main...HEAD` lists only `stella-tui`, two `stella-cli` doc comments, and `scripts/prose-baseline.txt`).

That is an observation rather than an inference: `main` at `22961967c` — this branch's exact base — fails the identical test with the identical panic, in its own CI run https://github.com/macanderson/stella/actions/runs/32764103149. Filed as #4709, since the open `main-red` issue #4691 names only `file-size` and this second cause was untracked.

Everything else in that job passed on CI, including the three steps this diff could plausibly have broken:

- `cargo fmt --check` ✓
- `cargo clippy -D warnings` ✓
- `cargo doc -D warnings` ✓ — which is what the `budget_card.rs` doc-link fix in this PR was needed for

**`main is not known-broken`.** `main-red-hold.yml` failing because a `main-red` issue is open (#4691) is that workflow working as designed.

Locally, `cargo test -p stella-tui` is fully green — 1066 lib tests plus every integration binary, the five question-overlay goldens passing unblessed.

PR #4708, opened 2026-08-24
feat(stella-tui): port the MCP tab and its inspector to the v2 SPEC 5 frame

macanderson:

One slice of the epic: the MCP tab and its ctrl+o inspector now render through v2-styled code, and the two `views/` files are gone.

Refs #4676

## What was actually wrong

The tab was painting from two palettes at once. Its browse list already read from `stella_tui_theme::token`; the registry-search pane, the auth prompt and the whole inspector still took `crate::theme`'s pre-SPEC-5 ramp. Nothing about that is visible in a `contains` assertion — every row kept its words — but the same fact wore two colours depending on which surface you read it from:

| fact | list row | inspector (before) |
|---|---|---|
| `not connected` | `token::RED` | `theme::WARNING_BRIGHT` (amber) |
| a selected row | `token::HL` | `theme::SELECT_BG` |
| the cap / budget warnings | `token::RED` | `theme::WARNING_BRIGHT` |

## What changed

- `src/views/mcp.rs` → `src/v2/mcp_tab.rs`; `src/views/mcp/detail.rs` → `src/v2/mcp_tab/detail.rs`. `views::mcp` is gone from `views.rs`, and `deck_render.rs`'s `DeckTab::Mcp` arm plus its one `compact_heading` call are the only dispatch sites repointed.
- Every span in both modules resolves to `stella_tui_theme::token`. No `crate::theme` colour survives in either file.
- The inspector takes the rounded `token::BORDER` block with a right-aligned key hint — the overlay frame `v2::sessions` established — instead of a square block edged in the accent.
- The tab body draws no chrome of its own. It never did draw a full-tab border, so this is stated rather than removed; the footer stays because the MCP verbs are unhinted in `keymap.rs`, so that list is the only place they are written down.
- The dead `let _ = dim;` in the tab's `render` is deleted, and two `[honesty-declaration]` prose entries are dropped from `scripts/prose-baseline.txt` by fixing the sentences rather than moving the baseline rows.

## Evidence

Two witnesses, because the regression this port fixes is exactly the kind a text assertion cannot see:

- `v2::mcp_tab::tests::every_pane_paints_from_the_token_palette` renders browse, search and auth and fails on any span whose fg or bg is outside the token set. It fails on the old code (`theme::INK`, `theme::SELECT_BG`, `theme::DANGER_BRIGHT` are all off-palette) and passes on the new.
- `v2::mcp_tab::detail::tests::the_popup_takes_the_shared_overlay_frame` reads the drawn buffer: the rounded corner's cell must be `token::BORDER`, and the subtitle's status word must be the colour the list row gave it.

The `tests/snapshots/deck/tab_mcp.txt` golden is **byte-identical** and was not re-blessed — that is the evidence the list renders exactly as it did.

Run locally, targeted: `cargo clippy -p stella-tui --all-targets -- -D warnings` clean, `cargo test -p stella-tui --no-fail-fast` green (10/10 suites), `make guards-fast` green, `cargo fmt --check` clean. CI is the real gate.

## One thing a reviewer should push back on if they disagree

`McpTabState` and its four companion types moved **with** the paint, into `v2/mcp_tab.rs`, rather than into a `deck_ui/mcp_state.rs` beside the keys that mutate them — which is where the `deck_ui::sessions` ↔ `v2::sessions` precedent says they belong. That split was written and then reverted: adding the module declaration grew `src/deck_ui.rs` by four lines, and it is a grandfathered god file closed to growth. Buying the budget by trimming unrelated doc comments in a shared-cell file is worse than leaving the types where they already lived. Epic item 5 (re-evaluate this crate's god files once the code driving their size is no longer split across two generations) is what reopens the question; the module carries a comment saying so rather than leaving it unexplained.

## Summary by Sourcery

Port the MCP tab and inspector to the v2 SPEC-5 frame while preserving the existing tab snapshot output.

New Features:
- Port the MCP tab and inspector into the v2 SPEC-5 frame with consistent token-based styling and shared overlay presentation.

Bug Fixes:
- Align MCP browse, search, authentication, and inspector status and warning colours so the same state is represented consistently across surfaces.

Enhancements:
- Pin the MCP status and key legend to dedicated bottom rows and hide the legend while the modal inspector is open.
- Replace the legacy MCP view module and update rendering, state, input handling, and tests to use the v2 implementation.

Tests:
- Add rendering tests covering token-palette usage, floor-pinned key hints, modal legend behavior, and the shared inspector frame.

Chores:
- Remove the obsolete MCP view modules and update prose baselines without changing the MCP tab snapshot output.


macanderson:
### Both red checks are `main`'s, not this diff's

`main` is red on three tracked counts right now, and this branch inherits two of them. Evidence for each, since a red PR that says "not my fault" is worth nothing without one:

**1. `main is not known-broken` — fail.** Working exactly as designed. `check-main-red-hold.sh` reports the open `main-red` issue #4691 (`file-size` fails on the merged tree) and holds. That is a fact about the base branch.

**2. `fmt + clippy + test` — fail, on `stella-runtime`, not `stella-tui`.**

```
crates/stella-runtime/tests/no_ambient_reads.rs:71 FAILED
  crates/stella-runtime/src/bin/wrapper-plugin-fixture.rs:76 reads ambient process state
```

That fixture landed in #4675 and this branch does not touch it:

```
$ git diff origin/main HEAD --stat -- crates/stella-runtime/
(no output — zero lines differ)
```

`no_ambient_reads` is a source scan over `crates/stella-runtime/src/**`, and lines 76-77 are byte-identical between `origin/main` and this head, so it fails on `main` for the same reason. Already tracked as **#4709** — not filed again.

**3. `cargo doc` (#4714, not yet reached in this run).** `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps -p stella-tui` fails on `src/v2/budget_card.rs:10` — a sibling v2 port's link to the private `cards::card_frame`. I checked whether my own two new modules had the same defect by masking that one line locally: with it masked the crate documents clean, so `v2/mcp_tab.rs` and `v2/mcp_tab/detail.rs` carry no private intra-doc links. The local probe was reverted; the fix belongs to #4714.

### What this branch was actually verified against

Run on the rebased head (`370973d`), targeted per this repo's laptop rule:

- `cargo clippy -p stella-tui --all-targets -- -D warnings` — clean
- `cargo test -p stella-tui --no-fail-fast` — 10/10 suites ok, 0 failed
- `make guards-fast` — exit 0 (includes `file-size`, `module-reachability`, `prose`, `lockfile-sync`, `fmt --check`)
- `tests/snapshots/deck/tab_mcp.txt` — byte-identical, not re-blessed

Staying a draft until `main` is green and this can be re-run against a base that distinguishes a broken change from a broken base.

macanderson:
### Reconciling the cross-PR review: the key legend now has its own band

**The finding.** The review flagged that this tab folds its hint/key legend into the last `Line` of the content `Paragraph`, where its ported siblings (`tools.rs`, `skills.rs`, `traces.rs`, `seats.rs`) give it a dedicated bottom row. Confirmed on the branch head — `render` pushed a blank spacer, the status line, and `footer(state.mode)` onto the same `Vec<Line>` as the server list and rendered the whole thing into `area`.

**Why it mattered.** The legend rode the list. One configured server put it on row 5; six put it on row 15. An operator looking for `ctrl+o inspect` had to find it again after every install, enable, or search.

**The fix.** Three bands, copying `v2::tools`'s `render_panel` shape from `tui-v2-port/tools-pane`:

```rust
let bands = Layout::vertical([
    Constraint::Min(0),    // header · the mode's pane
    Constraint::Length(1), // the driver's last word
    Constraint::Length(1), // keys
]).split(area);
```

with `render_status` and `render_keys` painting into their own rects, each guarding `area.height == 0`. The legend's words and `token::MUTED`/`token::DIM` styling are unchanged; only where it lands moved.

**One thing pinning it surfaced.** The floor row sits outside the centered inspector popup, which the old placement was *covered* by. The inspector is modal (`deck_ui::mcp_keys` returns before every mode while it is up), so the band would have advertised `a auth` and `x remove` at the moment they do nothing. `render_keys` yields while the inspector is open — its own verbs are already in its border title, which is where an overlay's legend belongs. This is the redundancy the review anticipated, resolved rather than duplicated.

**Selection**: already compliant, unchanged — `headline` and the search-result rows both paint the `▸` marker *and* `token::HL` background on the selected row, so the golden suite's style-stripped grid still sees the selection.

**Evidence.**
- `cargo test -p stella-tui --no-fail-fast` — exit 0, 1139 passed / 0 failed across all targets.
- `cargo clippy -p stella-tui --all-targets -- -D warnings` — exit 0.
- `cargo fmt --check -p stella-tui` — exit 0.
- `make prose` — `OK -- none added`.
- Golden `tests/snapshots/deck/tab_mcp.txt` re-blessed and read: exactly one line moved, from line 17 (under the registry line) to line 41 (the content area's floor, directly above the pulse row). No other cell changed.

Two witnesses added, both failing on the previous commit:
- `the_key_row_is_pinned_to_the_floor_whatever_the_list_holds` — renders at 1 and 6 servers, asserts the legend is on the floor row both times, that the driver's last word sits directly above it, and that the content band no longer paints `ctrl+o` at all.
- `the_key_row_yields_to_the_modal_inspector` — the legend is absent from the whole frame while the inspector is open.

Issue #1741, opened 2026-08-06
A diff-less mutation clobbers a path's good diff: touch_file overwrites latest_diff with None, and the Files tab never reads the history behind it


macanderson:
Second, independent way the Files tab reaches `(no diff captured)` for a file it has a diff for. (The first — `/clear` wiping the diff store while the ledger survived — is fixed on branch `worktree-files-tab-diff-survives-clear`.)

## The defect

`crates/stella-tui/src/model.rs::SessionModel::touch_file`, mutation arm:

```rust
if kind.is_mutation() {
    existing.kind = kind;
    existing.latest_diff = diff.clone();   // <- unconditional, including None
    ...
    existing.remember_diff(diff, added, removed);
}
```

A mutating `FileChange` that carries `diff: None` therefore **erases** the diff an earlier mutation of the same path delivered. `remember_diff` early-returns on `None`, so `recent_diffs` still holds the good text — but `views::files::find_diff` reads `latest_diff` only and never falls back to the history. Result: a row with correct cumulative `+`/`-` counts and a pane claiming nothing was captured, with the bytes sitting one field away.

## Reachability

`ToolRegistry::record_touch` always attaches a diff for a real change (`file_touch::changed_region_diff` returns `None` only when nothing moved). The reachable producer is the **pipeline's adoption re-emit**:

- `crates/stella-cli/src/candidate_ws/adopt.rs::adopt_inner` builds each `AdoptedChange` from `parse_name_status` with `diff: None`, then attaches counts and text best-effort:
  ```rust
  if let Ok(numstat) = git(&self.dir, &numstat_args).await { attach_numstat(...) }
  if let Ok(text)    = git(&self.dir, &text_args).await    { attach_diffs(...) }
  ```
  Either call can fail independently, and `attach_diffs` documents that a path the patch did not describe keeps `diff: None` deliberately (`a_path_the_patch_never_described_stays_uncounted_not_misattributed`).
- `crates/stella-pipeline/src/pipeline.rs` (~L1861) forwards those straight into `AgentEvent::FileChange`.

So: a numstat that succeeded plus a text diff that did not (or a path only the numstat named) yields counts with no diff — and if that path had already been edited in-session by a tool, the good diff is destroyed rather than merely absent.

## Why this is a design call, not an obvious patch

Two fixes, with a real trade-off:

[...]

Issue #1740, opened 2026-08-06
Files tab pairs a cumulative +/- footer with a single-mutation diff body, and says nothing about the mismatch


macanderson:
The Files tab diff pane can state two different deltas for one file at the same time, with no indication that they measure different things.

## What happens

`crates/stella-tui/src/views/files.rs::render_diff_pane`:

- the **footer** (`diff::footer_line(added, removed, w)`) uses `record.added` / `record.removed` — the ledger records **cumulative** counts across every mutation of that path (`deck.rs::FileLedger::record` does `rec.added += added`).
- the **body** renders `find_diff(...)` → `FileState::latest_diff` — the diff of the **newest single mutation** only.

Observed on a real session (deck Files tab, `crates/stella-graph/src/store.rs`, four mutations of +49/-0, +1/-1, +1/-1, +13/-4): the row and footer both read `+64 -6`, while the body showed only the last hunk (+13/-4). Nothing on screen says the body is a subset of what the footer counts.

The footer/row agreement is deliberate and documented at `views/files.rs` (the comment beginning "The footer states the RECORD's measured delta"): it was changed to stop the row and the footer disagreeing about one file. That fix was right. It just moved the disagreement from row↔footer to footer↔body, where it is still unstated.

## Why it matters

The pane reads as "here is the change, and it was +64/-6". A reviewer using it to check what a turn did to a file will conclude 58 of the added lines are somewhere they cannot see, or (worse) not notice.

## Options (a design call, not a mechanical fix)

1. **Say it.** Keep the cumulative footer, and label the body — e.g. `latest of 4 changes (+13 -4)` on the diff header line. Cheapest, honest, no data model change.
2. **Render the whole history.** `FileState::recent_diffs` already keeps the last `DIFF_HISTORY` (8) diffs, each tagged with its own `(added, removed)` (`model/file_state.rs::RememberedDiff`). The pane could stack them newest-first with per-mutation headers, and then the footer's cumulative total is the true sum of what is displayed. Bounded at 8, so a path mutated more times than that still needs option 1's label for the remainder.
3. Footer shows the displayed mutation's delta only — **rejected**: this is exactly the row↔footer disagreement the current comment says was removed. Do not re-introduce it.

Option 2 with option 1's label for the elided tail is the most complete answer; option 1 alone is a real improvement and is a one-line change.

## Files

- `crates/stella-tui/src/views/files.rs` — `render_diff_pane`, `find_diff`
[...]

PR #4710, opened 2026-08-24
feat(stella-tui): port the FILES tab to the v2 SPEC 5 frame

macanderson:
Refs #4676

## What moved

`views/files.rs` is deleted; the FILES tab renders from `v2/files_tab.rs`, and `deck_render.rs`'s `DeckTab::Files` arm is the one dispatch site repointed. The v1 module took shape (a), not a shim: it had no separable empty/loading state worth keeping distinct from the content renderer, so the whole thing moved.

## What the port changed

- **No chrome of its own.** The tab drew a bordered box around its whole content area, titled `Files · n` — a title the tab row already carries, around an area `v2::frame` already carved out. Both are gone. `the_tab_draws_no_chrome_of_its_own` is the witness.
- **Totals moved from a footer to a head strip** (`▤ 2 files · +7 -1 · 0 reads`), so they read the same whether or not the diff pane is open, and the selected row's `+a -b` sits beside the session's rather than a pane away. `↵ diff` / `↵ close diff` holds the right edge, since the FILES row of the keymap is not a hinted one and the frame's hint row never named it.
- **The op badge takes SPEC 6.2's rail metals** — read silver, write and edit gold, delete red — instead of a status palette. Create and modify share gold under that rule, which is what the shared CRUD letter (`textline::crud_letter`) is for.
- **The path column splits the way every other v2 surface splits one** (`v2::transcript`'s subject): the directory recedes to `DIM`, the basename takes `TEXT`. In a column a reader scans, the basename is the identity.
- **The selection carries `▸` as well as a highlight.** SPEC 13 is never colour alone, and the golden suite's own doc names this tab as the one selection it could not capture because style was the only signal. It can now.
- **Two column alignments fixed on the way past.** The agent cell padded to one less than its column, which put every op badge one column left of its own header cell; and the `+`/`-` headers were right-aligned over left-aligned values, eight columns from the numbers under them.

The diff pane is unchanged: same `crate::diff` chrome, same `VIEW_CAP` budget, same `files_diff_*` metrics and scroll window, same earlier-diff notice. Every witness the v1 module carried moved with it (#1740, #1741, the `/clear` pair), plus three new ones for the port.

## Verification

- `cargo test -p stella-tui` — 1083 + 66 across the integration binaries, all green.
- `make guards-fast` — green, `prose` included. The prose baseline loses the deleted file's line rather than gaining one for the new file: the two grandfathered constructions were rewritten, not moved.
- `tests/snapshots/deck/tab_files.txt` re-blessed and read line by line — the border and the footer are gone, the head strip and the `▸` marker are there, and every column lands under its own header.

## Deleted tests

None. The eight `#[test]`s in `views/files.rs` moved to `v2/files_tab.rs` under the same names; git records the file as a rename.

## Summary by Sourcery

Port the FILES tab to the v2 SPEC 5 frame and bring its layout, styling, selection, and interaction hints into line with the other v2 surfaces.

New Features:
- Port the FILES tab to the v2 SPEC 5 frame with shared frame chrome, session totals, file records, and diff interaction.

Bug Fixes:
- Correct FILES table column alignment and preserve clear visual selection and path identity across styling conditions.

Enhancements:
- Align FILES presentation with v2 conventions by moving totals to a head strip, adding a bottom key band, using rail-metal operation colors, and emphasizing path basenames.
- Retain the existing diff behavior and ledger semantics while relocating the implementation and its tests to the v2 module.

Documentation:
- Update the stella-tui module documentation to describe the v2 FILES tab location and behavior.

Tests:
- Re-bless the FILES tab snapshot and add coverage for frameless rendering, selection markers, key-band states, empty-state behavior, and path emphasis.

Chores:
- Remove the legacy FILES view and route deck rendering through the v2 implementation.


macanderson:
CI note, for whoever reads this PR's red before its diff.

Rebased onto `22961967c` (current `main`), where it composes cleanly with #4686's card port — both touch `src/v2.rs`, `views.rs`, `deck_render.rs`, `panel_guard.rs` and the crate README. On that base, `cargo test -p stella-tui` is green (1069 lib + 66 across the integration binaries, golden frames included, no re-bless needed for the compose), and `cargo clippy -p stella-tui --all-targets -- -D warnings` is clean.

The red on this PR is `main`'s, not this diff's, and each break already has a repair PR open:

- `main is not known-broken` — the canary's #4691 (`crates/stella-transcript/src/tests.rs` over the file-size ceiling). Repairs: #4689, #4713.
- `cargo doc -D warnings` — `public documentation for budget_card links to private item cards::card_frame`, at `crates/stella-tui/src/v2/budget_card.rs:10`, from #4686. Repair: #4711. Reproduced here with `RUSTDOCFLAGS="-D warnings" cargo doc -p stella-tui --no-deps --document-private-items`, which points at that line and no file in this diff.
- `cargo test` — `stella-runtime`'s `no_ambient_reads`, on `crates/stella-runtime/src/bin/wrapper-plugin-fixture.rs:76-77`, from #4675. Repair: #4690.

Nothing folded in from those repairs: a red `main` gets its own PR from a fresh `main` rather than a rider on an unrelated one.

PR #4712, opened 2026-08-24
fix(stella-tui): stop budget_card's public doc linking a crate-private helper

macanderson:
## What

`main` fails `cargo doc -D warnings`, which is a step of the required
`fmt + clippy + test` job — so every open PR fails on it regardless of its own
diff. Reproduced on `22961967c`, current `main`:

```
   |
10 | //! A floating card draws its own frame ([`cards::card_frame`]) because it is
   |                                            ^^^^^^^^^^^^^^^^^ this item is private
   |
   = note: this link will resolve properly if you pass `--document-private-items`
```

#4686 gave `v2::budget_card` a module doc that links `[`cards::card_frame`]`,
and `views::cards::card_frame` is `pub(crate)` (`views/cards.rs:59`). Note the
first `note:` line — the link resolves under `--document-private-items` and
only there, so a local `cargo doc` run with that flag looks clean while CI's
plain run does not. That is the shape that let this land.

The helper is named rather than linked. It has no public page to point at, and
making it `pub` to satisfy a doc comment would widen an internal frame helper's
surface for the sake of a hyperlink. The sibling `[`crate::v2::frame`]` link on
the next line is public and is untouched. This is the only such link in the
crate — `rg 'card_frame' crates/stella-tui/src` finds eight call sites and one
doc reference, and this was the doc reference.

## Evidence

```
RUSTDOCFLAGS="-D warnings" cargo doc -p stella-tui --no-deps
    before: exit 101, the error above
    after:  exit 0
RUSTDOCFLAGS="-D warnings" cargo doc -p stella-tui --no-deps --document-private-items
    after:  exit 0
cargo fmt --all --check    clean
python3 scripts/check-prose.py   OK — none added
```

No witness test: the guard that failed is the witness, and it is `cargo doc`
itself.

Refs #4686

## Summary by Sourcery

Bug Fixes:
- Fix the `stella-tui` documentation so public module docs no longer link to the crate-private `card_frame` helper, allowing rustdoc warnings to pass.

### DIFF
--- a/crates/stella-tui/src/v2/budget_card.rs
+++ b/crates/stella-tui/src/v2/budget_card.rs
@@ -7,7 +7,7 @@
 //! is only ever the folded one, so an ignored or clamped request can never
 //! render as a cap in force.
 //!
-//! A floating card draws its own frame ([`cards::card_frame`]) because it is
+//! A floating card draws its own frame (`views::cards::card_frame`) because it is
 //! an overlay rather than a tab body — the frame [`crate::v2::frame`] carves
 //! out is what it floats *above*, and a card with no border of its own would
 //! read as content the deck had appended to whatever is underneath it.


macanderson:
Closing as a duplicate: #4711 fixes the same red (the `pub(crate)` intra-doc link at `crates/stella-tui/src/v2/budget_card.rs:10`) and was opened 84 seconds earlier. Two sessions found the same unclaimed break at the same moment — that is the third such collision on this tree today, and it is worth naming rather than shrugging at.

Nothing here is lost. Your observation about `--document-private-items` is the part I am carrying forward: that flag *resolves* a link to a private item, so running only that form hides exactly this error class. The plain form has to run too, and run first. I have corrected the standing instruction the fix agents work from.

macanderson:
### Why the `fmt + clippy + test` job is red here, and what it proves

The step this PR exists to fix **passed**:

```
$ gh api repos/macanderson/stella/actions/jobs/97561806737 --jq '.steps[] | ...'
cargo fmt --check:      success
cargo clippy -D warnings: success
cargo doc -D warnings:  success      <- the step that fails on main
```

The job then failed further along, at `cargo test`, on a different red that is
also on `main` and is not this diff:

```
crates/stella-runtime/src/bin/wrapper-plugin-fixture.rs:76: reads ambient process state
    let inherited = u8::from(std::env::var_os("CARGO_MANIFEST_DIR").is_some());
test result: FAILED. 0 passed; 1 failed    (stella-runtime --test no_ambient_reads)
```

That one is #4675's, and #4690 is the repair.

So `main` is red on three independent things at once, each hiding the next —
the pattern AGENTS.md names in the `main-canary` section. Only the first one
to run is visible on any given PR, which is why the other two went unnoticed:

| Red | From | Repair |
|---|---|---|
| `cargo doc -D warnings` | #4686 | **this PR** |
| `no_ambient_reads` | #4675 | #4690 |
| `file size ratchet` | #4683 | #4689 (`Closes #4691`) |

`file size ratchet` runs in its own workflow, so it shows independently and
passes here. The other two share this job, so #4690 cannot get past `cargo doc`
to prove its own fix until this one lands.

Suggested order: **this PR → #4690 → #4689**, re-running or merging `main` into
the later two.

PR #4713, opened 2026-08-24
test(stella-transcript): split the read-as-source tests out of tests.rs

macanderson:
## What

`crates/stella-transcript/src/tests.rs` reached **1523 lines**. It is not
grandfathered, so `file-size` fails on the merged tree and reds every open PR —
the `file-size` half of #4691.

The guard names the remedy and forbids the other one:

```
that predate the guard, and this is the rule that stops the tree getting worse.
```

So this is a split. **No baseline entry is added and no ceiling is moved.**

## The cut

The section that comes out is the one that already had its own fixture
vocabulary: *a `read_file` body is source, and the export renders it as source.*
It moves to `crates/stella-transcript/src/tests/read_export.rs`, beside the five
sibling test modules that are already there (`json_reindent`, `fields_projection`,
`frame`, `width`, `tail`) and following their idiom — `//!` header saying why it
is its own file, then `use super::*`.

| | before | after |
| --- | ---: | ---: |
| `tests.rs` | 1523 | **1282** |
| `tests/read_export.rs` | — | 265 |

218 lines of headroom, so this is a split rather than a shave that the next
test re-breaks.

Two judgement calls in it:

- **`numbered` and `read_call` move**, because nothing else in the test tree
  calls them. `use crate::syntax` moves with them for the same reason — after
  the split this module is its only caller, and leaving the import in the parent
  would be an unused one.
- **`rendered` stays in the parent.** `fields_projection` renders through it
  (`use super::{bash, output, rendered}`), and a second copy of "open both
  surfaces with everything unfolded" is how the two surfaces start disagreeing
  about what open means. Its doc comment now says so, so the next person to look
  for it in `read_export` finds the reason instead of duplicating it.

## Evidence

`cargo test -p stella-transcript` → **125 lib + 2 integration passed, 0 failed,
0 warnings**, exit 0. A move is proven by the tests surviving it unchanged, and
none is deleted or renamed — `check-deleted-tests` should stay quiet.

`./scripts/check-file-size.sh` → exit 0, the failure this PR exists for.
`cargo fmt -p stella-transcript --check` → clean. `make prose` → OK, none added.

## Scope

`main` is red on **two independent causes** and this is only one of them. The
other is `stella-runtime`'s `no_ambient_reads`, repaired by **#4690** — not
touched here. Per the note on #4691, that issue should stay open until both are
green; this PR is deliberately `Refs`, not `Closes`.

Landed on its own from a fresh `main` and folded into nothing, for the reason
AGENTS.md gives about the file-size baseline being a shared cell.

Refs #4691

## Summary by Sourcery

Split the read-export tests into a dedicated submodule to keep the transcript test suite maintainable and within the file-size limit.

Enhancements:
- Split read-as-source transcript tests into a dedicated `read_export` submodule while preserving their coverage and shared rendering helper.
- Reduce `tests.rs` below the file-size limit without adding a baseline entry or changing the limit.

Tests:
- Preserve the existing read-export test coverage, including numbering, syntax highlighting, gutter handling, and cross-surface rendering consistency.

### DIFF
--- a/crates/stella-transcript/src/tests.rs
+++ b/crates/stella-transcript/src/tests.rs
@@ -24,13 +24,17 @@ mod width;
 // `render_turn_tail`'s cutoffs (#4566's fix shape (a)).
 mod tail;
 
+// A read body rendered as source. In its own file because it brings its own
+// fixture vocabulary — the emitter's gutter format and both export renderers —
+// which nothing else here calls.
+mod read_export;
+
 use crate::digest::{self, format_cost, format_duration, format_tokens};
 use crate::file_diff::{FileDiff, RowKind};
 use crate::fold::{Command, Cursor, FoldState, Zoom, apply};
 use crate::grid;
 use crate::html;
 use crate::model::*;
-use crate::syntax;
 use crate::word;
 use unicode_width::UnicodeWidthStr;
 
@@ -123,33 +127,11 @@ fn run_with(steps: Vec<Step>) -> Run {
     }
 }
 
-// ------------------------------------------------- a file read is source
-
-/// One `read_file` body line, in the emitter's own shape
-/// (`stella-tools`' `{line_num:>6}\t{line}`), so a test cannot drift from the
-/// format it is meant to be reading.
-fn numbered(first: usize, source: &[&str]) -> Vec<String> {
-    source
-        .iter()
-        .enumerate()
-        .map(|(i, l)| format!("{:>6}\t{l}", first + i))
-        .collect()
-}
-
-fn read_call(path: &str, lines: Vec<String>) -> Call {
-    Call {
-        tool: ToolKind::ReadFile,
-        header_object: path.to_string(),
-        args: Vec::new(),
-        output: Output { lines, clipped: 0 },
-        files: Vec::new(),
-        status: Status::Ok,
-        duration_ms: 1,
-        speculated: false,
-    }
-}
-
-/// Both export renderers, with everything open.
+/// Both export renderers over one call, with everything open.
+///
+/// Shared rather than moved into [`read_export`]: `fields_projection` renders
+/// through it too, and two copies of "open both surfaces" is how the surfaces
+/// start disagreeing about what open means.
 fn rendered(call: Call) -> (String, String) {
     let run = run_with(vec![step(call, 0)]);
     let mut state = FoldState::new();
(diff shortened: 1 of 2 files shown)


macanderson:
## This PR's red is inherited, and CI has **not** yet checked the split

`fmt + clippy + test` failed here on two of `main`'s own breaks, neither in this diff:

- `cargo doc -D warnings` → `public documentation for `budget_card` links to private item `cards::card_frame`` — #4711's subject.
- `cargo test` → `no_ambient_reads` — #4690's subject.

Both are repaired together in **#4716** (they deadlock each other, so neither can merge alone).

**The part worth stating plainly:** `cargo test --workspace` stops at the first failing test binary, and `no_ambient_reads` is that binary. Grepping this job's log for the tests this PR moved — `an_exported_read_is_numbered_once_by_the_files_own_line_numbers`, `every_surface_paints_a_body_the_same_way` — returns nothing. **CI never ran them.** So the only evidence the split is sound is still the local run in the description:

```
$ cargo test -p stella-transcript
test result: ok. 125 passed; 0 failed  (lib)
test result: ok. 2 passed; 0 failed    (integration)
0 warnings, EXIT=0
```

Plus `./scripts/check-file-size.sh` → exit 0, which is the guard this PR exists for and which compiles nothing, so it is unaffected by the inherited breaks.

I would rather say that than let a red job be read as "the split was checked and something else broke". Once #4716 lands and `main` is merged back in, the re-run is the first one that actually exercises `tests/read_export.rs`.

macanderson:
Superseded by #4719, which fixed the same `file size ratchet` break — and fixed it the right way, by splitting rather than adding a baseline entry. `crates/stella-transcript/src/tests.rs` is 1402 lines on `main` now.

It cut a different seam than this PR did: #4719 moved the **file-kind** tests into `tests/file_kinds.rs`, where this one moved the **read-as-source** tests into `tests/read_export.rs`. Both are legitimate; theirs landed. Rebasing this on top would conflict across the same file for no gain — the ceiling is already cleared with 98 lines of headroom.

One observation worth keeping from this branch, in case anyone splits that file again: `rendered` cannot move out of the parent module, because `fields_projection` imports it (`use super::{bash, output, rendered}`). I hit that as a build error mid-split. #4719's cut does not touch it, so nothing is owed today.

Closing. #4691 is already closed by #4719.

PR #4716, opened 2026-08-24
fix(stella-tui,stella-runtime): land main's two remaining repairs together, since neither can go green alone

macanderson:
## What

`main`'s two remaining breaks, on one branch, because **neither can be merged on its own.**

This is not a rewrite of anyone's work: it is #4711 and #4690 merged together, their
commits and authorship intact. Both PRs stay the place their reasoning is written down.

## Why they had to be composed

`main` is red on two independent causes, and the repair for each is blocked by the other:

| PR | fixes | its own CI fails on |
| --- | --- | --- |
| #4711 | `cargo doc -D warnings` — `budget_card`'s public doc intra-doc-links `views::cards::card_frame`, which is `pub(crate)` | `no_ambient_reads` — #4690's break |
| #4690 | `no_ambient_reads` — the wrapper-plugin fixture lived in `src/`, which that test governs, and its `env-probe` mode must read its own environment | `cargo doc` — #4711's break |

So each branch carries a tree that still contains the other's defect, `fmt + clippy + test`
fails on both, and branch protection refuses the merge — correctly, since neither tree is
actually green. A deadlock, and it is the shape AGENTS.md names: **one red gate step masks
the next.** `cargo doc` runs before `cargo test` in that job, which is why #4690's log shows
only the doc error and never reaches the test it repairs.

**The alternative was an `--admin` merge past the required check, and that is the worse
option.** It would land a tree no CI run ever verified, which is the hazard the gate exists
to prevent (#3887). Composing the two instead produces a tree that either passes on its own
merits or does not — and if a third cause is hiding behind these two, this is the run that
surfaces it rather than another round of masking.

## What is in it

Nothing authored here. Two merges, no conflicts, no edits on top:

- `crates/stella-tui/src/v2/budget_card.rs` — the doc link becomes a named reference
  (+5 −2).
- `crates/stella-runtime/` — `src/bin/wrapper-plugin-fixture.rs` → `tests/fixtures/`, plus
  its three citations and the `Cargo.toml` target declaration (+21 −3, 96% rename).

No guard is relaxed, no baseline is widened, no `#[allow]` is added.

## Evidence

CI on this PR is the evidence, and it is the whole point of the PR: this is the first tree
in which either repair can be checked against the other. `cargo test -p stella-runtime
--test no_ambient_reads` is also running locally against this exact tree; I will post the
result rather than assert it.

## After it lands

#4711 and #4690 become empty against `main` and should close as merged-by-composition —
their commits are in the history under their own authorship. #4691 stays open for its
`file-size` half, which is #4713 and is deliberately not folded in here.

Refs #4691

## Summary by Sourcery

Restore a clean build by composing the runtime ambient-read repair with the TUI documentation-link repair.

Bug Fixes:
- Fix the runtime ambient-read check by relocating the environment-probing wrapper plugin fixture outside the governed source tree.
- Fix the TUI documentation build by removing the invalid link to a crate-private item.

Build:
- Update the runtime binary target and related references to use the test fixture's new location.

Documentation:
- Update runtime documentation and contributor guidance to reflect the wrapper fixture's new location and purpose.


macanderson:
Local result on this exact tree, as promised rather than asserted:

```
$ cargo test -p stella-runtime --test no_ambient_reads
    Running tests/no_ambient_reads.rs
running 1 test
test no_ambient_reads ... ok
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
EXIT=0
```

That is the test #4690 repairs, run on the composed tree — so the composition does not undo it. It says nothing about the `cargo doc` half; **CI's `fmt + clippy + test` is the check that has both in front of it**, and that is the one this PR exists to obtain. It runs `cargo doc -D warnings` before `cargo test`, so a green job is the first evidence either repair has that the other did not mask it.

macanderson:
Superseded by #4719, which landed the same composition — the three repairs cherry-picked onto one branch and verified together — about half an hour before this run finished. `main` is green at `f6cf84d60` and no `main-red` issue is open.

Same diagnosis, same remedy, and #4719 got there first with the third break (`file size ratchet`) folded in as well, which this PR deliberately left out. Nothing here is worth keeping.

Closing rather than rebasing. The reasoning is preserved in #4718, which asks the canary to say "more than one failing check → single-cause repairs will not be mergeable" so the next session reaches the composition move in one step instead of three.

Issue #4714, opened 2026-08-24
main is red: cargo doc fails on a private intra-doc link in v2/budget_card.rs


macanderson:
## What is broken

`main` does not pass `cargo doc -D warnings`, which is a required step of the `fmt + clippy + test` job in `ci.yml` and a `make gate` step (`doc-warnings`). Every open PR inherits the failure.

```
error: public documentation for `budget_card` links to private item `cards::card_frame`
  --> crates/stella-tui/src/v2/budget_card.rs:10:44
error: could not document `stella-tui`
```

Observed on `main` at `22961967c` — PR #4701's CI run [32766693754](https://github.com/macanderson/stella/actions/runs/32766693754), whose own diff touches neither file (it is confined to `crates/stella-tui/src/v2/picker.rs` and its call sites).

## Where it came from

#4686 ("port the floating cards to the v2 SPEC 5 frame") moved `views/budget_card.rs` to `crates/stella-tui/src/v2/budget_card.rs`. The module doc at line 10 carries an intra-doc link:

```rust
//! A floating card draws its own frame ([`cards::card_frame`]) because it is
```

`crate::views::cards::card_frame` is `pub(crate)`. `v2/budget_card.rs` is reached through `pub mod budget_card;` in `crates/stella-tui/src/v2.rs`, so its module doc is public documentation, and rustdoc's `private_intra_doc_links` lint fires. The link is new prose, not a moved line: `git show 8d7f692de^:crates/stella-tui/src/views/budget_card.rs` shows the pre-port module doc ending at "render as a cap in force." with no frame paragraph and no intra-doc link. #4686 added the paragraph.

The gate step is `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps --document-private-items --keep-going`. `--document-private-items` does not silence this lint; the item is private relative to the *public* doc that links it.

`v2/models_card.rs:352` and `v2/plan_card.rs:304` call `cards::card_frame` in code rather than link it in prose, so they are unaffected.

## Repro

```sh
git checkout main && git pull
RUSTDOCFLAGS="-D warnings" cargo doc -p stella-tui --no-deps --document-private-items
```

Fails today. Or run `make doc-warnings`.

## Constraints

- `cards::card_frame` must stay `pub(crate)`: `views::cards` is v1 chrome on its way out with #4676, and widening it to `pub` to satisfy a doc link would export a module the epic is deleting.
[...]

PR #4720, opened 2026-08-24
feat(stella-tui): port the TOOLS pane to the v2 SPEC 5 frame

macanderson:

Refs #4676 (one slice of the epic, not its completion).

## What changed

`crates/stella-tui/src/views/tools.rs` → `crates/stella-tui/src/v2/tools.rs`. The SETTINGS tab's tool-switch editor now renders as four bands inside the body `v2::frame` already carves out, instead of an area-filling bordered panel with its counts in the border title:

- **header** — ` tools · 1 off · 1 org-locked · modified`, with the session's tool count on the right edge (a fact the bordered title had no room for);
- **rows** — the same windowed group/tool list;
- **status** — the driver's last word on a save or refresh;
- **keys** — the pane's verbs while it holds the keyboard, `e edit tool switches` while it does not.

Colour is `stella_tui_theme::token` throughout, like every other module under `v2/`. The selection is `glyph::COLLAPSED` over `token::HL` rather than a reversed row (the shape `v2::graph_tab` and `v2::subagents` already use). `on` reads as `token::TEXT` and `off` as `token::MUTED` — a pane where every row is lit is a pane where a lit row says nothing — and the accent is spent on one thing only: the `· modified` marker, which is the only cell asking to be acted on.

The row model, the most-specific-key policy resolution, the snapshot ingest and the modal key map move over unchanged. They are what the pane renders, not how.

### Call sites repointed

`views/settings.rs` (render + both focus routes), `deck_ui.rs` (the `ToolsOverlay` field, its default, `ingest_policy`, `handle_tools_key`), `views.rs`'s module list and header prose, and the path citations in `views/seats.rs`, `envelope/engine_config.rs` and `docs/spec/roleless-core.md`. The v1 file is deleted; `rg` finds no remaining reference outside a frozen `stella-tools` test fixture.

## Evidence

- `cargo test -p stella-tui --no-fail-fast` — 1136 passed, 0 failed.
- Every test the v1 module carried moves with it by name; three new ones pin the port itself: `the_pane_draws_no_box_of_its_own` (no border glyph anywhere in the frame, header on the first row), `the_key_row_follows_the_keyboard`, `a_short_pane_still_renders` (heights 1–4, and a 3-column pane draws nothing rather than garbage).
- `pane_settings_tools` re-blessed with the diff read: the box is gone, the counts moved to the header with `7 tools` on the right edge, `1 tools` is now `1 tool`, and every row is otherwise character-for-character what it was.
- `cargo clippy -p stella-tui --all-targets -- -D warnings` clean; `cargo fmt` applied.
- Guards: `check-prose` (the deleted file's stale baseline entry removed, nothing added), `check-file-size`, `check-god-files`, `check-invariants` all green.

`make doc-warnings` is red on this branch and on `main` for a reason this PR does not touch — the `pub(crate)` link in `v2/budget_card.rs`, tracked as #4714.

## Known follow-up

The SEATS pane beside it (`views/seats.rs`) still draws its own bordered box, so the SETTINGS tab is mid-migration between panes until that slice of #4676 lands. That is the epic's port-one-at-a-time sequencing, not a regression this PR introduces.

## Summary by Sourcery

Port the SETTINGS TOOLS editor to the v2 frame while retaining its policy behavior and improving its pane layout and rendering resilience.

New Features:
- Port the SETTINGS TOOLS pane to the v2 frame with a borderless four-band layout for headers, rows, status, and keyboard hints.

Enhancements:
- Adopt the v2 theme tokens and selection styling while preserving existing tool policy resolution, editing, snapshot ingestion, and keyboard behavior.
- Move the TOOLS implementation and all production references from the legacy views module into v2.
- Improve narrow and short-pane rendering behavior and display singular tool counts correctly.

Documentation:
- Update module references and specification guidance to point to the v2 TOOLS pane.

Tests:
- Add coverage for borderless rendering, keyboard-dependent key hints, short-pane safety, and modified-state indicators.
- Update the SETTINGS TOOLS rendering snapshot.

Chores:
- Remove the legacy TOOLS module and its stale prose-baseline entry.


macanderson:
Two red checks on this head, neither owned by this diff:

- **`main is not known-broken`** — `check-main-red-hold.sh` names open `main-red` issue #4691 (file-size on the merged tree). That is the hold doing its job against a red base; it clears when main is repaired.
- **`make doc-warnings`** — the `pub(crate)` intra-doc link in `v2/budget_card.rs`, tracked as #4714. Untouched by this PR and red on `main` for the same reason.

The prose gate was genuinely red on the first push and is fixed in 83f3c9a: the three constructions the deleted `views/tools.rs` baseline entry recorded travelled over with the doc comment, and since the entry went with the file they arrived as newly added. They are deleted rather than re-baselined.

Issue #4572, opened 2026-08-23
Tag the remaining six measured constants so a check protects them from silent changes


macanderson:
## Problem

PR #4571 landed `scripts/check-measured-constants.sh` (#2495) — the gate step
that fails when a constant carrying a `/// MEASURED:` doc marker is named by no
test. It seeded the marker onto **three** of the nine constants the audit
found. Six remain, and each is still a value a textually clean merge can revert
in silence, which is exactly the failure #2495 exists for (#2414's ceiling,
reverted blamelessly by #2462, found weeks later).

The guard cannot find these itself: the marker is opt-in by design, so an
unmarked constant is invisible to it. Until they are marked they are protected
by nothing.

## The audit list

Produced by sweeping `crates/stella-core/src/` (compaction, loop detection,
budget, repair, starvation, skills) and `crates/stella-model/src/` for `const`
/ `static` items whose doc comment cites an observation, a sample size, a
distribution, a census, or a benchmark.

**Named in no test at all** — these fail the guard the moment they are marked,
so each needs a test written in the same change:

| constant | file | value | what the doc records |
|---|---|---|---|
| `MAX_CYCLE_PERIOD` | `crates/stella-core/src/loop_detect.rs` | `4` | "Real stuck signatures observed so far are periods 2 and 3" |
| `FIRST_BYTE_TIMEOUT` | `crates/stella-model/src/http.rs` | see file | sits "well above any observed" first-byte latency |

**Used in tests, value not pinned** — a test mentions the identifier, so the
guard would pass on marking alone; the work is deciding whether the existing
test actually breaks when the value moves, and adding an assertion if not:

| constant | file | value | what the doc records |
|---|---|---|---|
| `STARVED_RETRY_CAP` | `crates/stella-core/src/starvation.rs` | `32_768` | sized to "any reasoning budget observed in the wild" (#2128) |
| `REPAIR_ATTEMPT_CAP` | `crates/stella-core/src/repair.rs` | see file | "the measured axes below routinely…" |
| `DEFAULT_STOP_HOLDS` | `crates/stella-core/src/driver/config.rs` | see file | "the empirically common case" |
| `MAX_SUGGESTIONS` | `crates/stella-core/src/skills/appraisal.rs` | `10` | sized against a measured 148 lines of suggestions |

**Marked in #4571, but pinned only by tests that *use* them** — the guard is
satisfied, the revert is not fully guarded:

| constant | file | note |
|---|---|---|
[...]

Issue #3103, opened 2026-08-13
Markdown indexing: extract link edges, and recognise setext headings


macanderson:
PR #3095 made markdown an indexed language (`crates/stella-graph/src/markdown.rs`), split on its ATX heading hierarchy. Two deliberate omissions, both cheap, both worth doing — recorded here rather than left as surprises.

## 1. A markdown link between documents is an import edge, and is not extracted

`parse_file` returns `imports: Vec::new()` for markdown. But `[the context spec](./docs/spec/adaptive-context/context-pr.md)` is exactly the relation `code_graph_imports` models, and the graph already has both directions wired: `imports_from` and `importers_of`.

Extracting them would make **"which documents reference this one"** answerable, which is a question this repository asks constantly — `docs/manifest.json`, `make doc-links`, `make doc-report` and the whole citation-by-id scheme exist because doc-to-doc references matter here and rot silently. AGENTS.md records 16 dead markdown-to-markdown links accumulating underneath two guards that only ever read Rust comments.

It also feeds the `structure` section of the dossier shape proposed on #3063: a doc's importers are part of "everything about payments".

**Scope:** inline links `[text](target)` and reference definitions `[label]: target`. Resolve relative targets against the document's own directory, exactly as `import::resolve` does for source files. Skip absolute URLs — they are not workspace edges (`ImportKind` already has the vocabulary for external vs resolved). Skip anything inside a fenced block, which `markdown.rs` already tracks correctly and can share.

**Where:** `crates/stella-graph/src/markdown.rs` (extraction), `crates/stella-graph/src/parse.rs` (the `Language::Markdown` arm currently returning an empty vec — it carries a `#TBD-md-links` marker to replace with this issue number), `crates/stella-graph/src/import.rs` (resolution).

## 2. Setext headings are not recognised

Only ATX (`#`, `##`, …) is. Setext — a line of text underlined by `===` or `---` — is not, and the code says why:

> `---` is also a thematic break and also YAML frontmatter's delimiter, and guessing between them wrong would invent sections rather than miss them.

That was the right call for a first pass: the failure direction of *not* recognising setext is one coarser section, while the failure direction of a wrong guess is a section that does not exist. But it is a real coverage gap for older documents, and CommonMark's disambiguation rules are well-defined rather than a matter of taste — a `---` line closes a setext heading only when the preceding line is a non-blank paragraph line, is not itself inside a fence, and is not the document's opening frontmatter delimiter.

[...]

PR #4738, opened 2026-08-24
fix(stella-graph,stella-tools,stella-cli): a deterministic index stamp, a breadcrumb-reachable exact rung, and the measured-constant audit settled

### LINKED ISSUE #4574: Exact-name search cannot find a table nested in a settings file or a heading nested in a document
## Problem

The search ladder's exact-symbol rung refuses any query that is not a bare
identifier. `enrich::symbol_terms` gates on `is_bare_identifier`
(`crates/stella-tools/src/search/enrich.rs:406`), which accepts only
`[A-Za-z_][A-Za-z0-9_]*`. Two symbol kinds in the index are therefore
unreachable by the rung that answers with a *fact* rather than a ranking:

- **TOML table sections** (#4492, landed in #4571). `[[record]]` indexes as
  `record` and is reachable; `[record.steering.applies_to]` indexes under its
  dotted key and is not, because of the `.`.
- **Markdown sections** (#3103). A section's name is its breadcrumb joined by
  ` › ` (`crates/stella-graph/src/markdown.rs`'s `BREADCRUMB_SEPARATOR`), so
  no markdown section has ever been reachable by exact lookup either.

Both fall through to the graph-names rung, which does reach them —
`names::push_units` splits on every non-alphanumeric character, so `steering`
matches `record.steering` — so this is a missing *certainty*, not a missing
answer. It is why `the_exact_symbol_rung_reaches_a_records_table` in
`crates/stella-tools/src/search/engine/tests/visibility.rs` queries `record`
and not the dotted key.

## Why it might be worth fixing

`exact_symbol_hits`'s own doc gives the argument: asked for a name with exactly
one definition, embedding rank returned it at **rank 5 of 139**, because a
symbol name is a handful of tokens against whole files of prose. A citation like
`AGENTS.md › Architecture › 8. Provider feature parity` or
`ctx.demo.toml › record.steering` is exactly the shape an agent writes when it
already knows what it wants, and today every one of those is a ranking.

## Why it is not obviously worth fixing

`is_bare_identifier` is what keeps the rung free for the common case: a query
that is a *sentence* is refused before any lookup runs. Widening it has to not
turn ordinary prose (which contains `.`) into a definition lookup. A likely
shape: accept a dotted or ` › `-separated run only when the whole query is one
such run with no whitespace outside the separator, and try the exact lookup on
that. Measure the false-positive rate against a real query log before shipping
it — `crates/stella-tools/tests/search_recall.rs` pins the name rung against a
frozen corpus and is the place to add the equivalent for this.

## Files

- `crates/stella-tools/src/search/enrich.rs:406` — `is_bare_identifier`
- `crates/stella-tools/src/search/enrich.rs:442` — `symbol_terms`
- `crates/stella-tools/src/search/enrich.rs:479` — `exact_symbol_hits`
- `crates/stella-graph/src/markdown.rs` — `BREADCRUMB_SEPARATOR`
- `crates/stella-graph/src/toml.rs` — dotted table keys
- `crates/stella-tools/src/search/engine/tests/visibility.rs` — where the
  current limit is visible in the fixture's choice of query

## Repro

```rust
// index a workspace holding `.stella/rules/ctx.demo.toml` and a markdown file,
// then:
dispatch(Some(&graph), &root, "record.steering.applies_to", None).await
// => Strategy::GraphNames, not Strategy::ExactSymbol
```

## Definition of done

Either the rung reaches a dotted / breadcrumbed section name with a witness
test per symbol kind and evidence that prose queries are still refused, or a
doc comment on `is_bare_identifier` states that section names are deliberately
out of scope and why — so the next reader does not re-derive the question.

## Related

- #3103 — markdown sections in the index
- #4492, #4571 — TOML table sections
- #3125 — why the exact rung prepends rather than stops

### LINKED ISSUE #4643: Two stella-graph vector tests race on a wall-clock-second tiebreak and flake under the parallel runner
## Problem

`vectors::tests::every_indexed_file_is_pending_until_it_is_embedded` (`crates/stella-graph/src/vectors/tests.rs:68`) and `ranking_is_ordered_by_similarity_and_bounded_by_the_floor` (`:338`) failed once under the parallel test runner (`left: ["b.rs","a.rs"], right: ["a.rs","b.rs"]`), then passed serially and in three subsequent full parallel runs (231/231). The assertion message names the cause: the ordering tiebreak is a wall-clock-second stamp both files can share, so the result depends on whether two index passes land in the same second — a race the test's own comment assumes away.

## Done means

The tiebreak stops depending on wall-clock seconds (a monotonic per-insert sequence, or a sub-second stamp, or an explicit secondary key on path), with the two tests deterministic under `--test-threads=8` run 20x. Seen 2026-08-24 during PR #4638's verification; unrelated to that diff.

### LINKED ISSUE #4662: A website path built from two joins is invisible to the website-inputs guard
## Problem

`crates/stella-cli/src/config/tests/docs_sync.rs:146-155` builds its website path from two `join` calls, so the literal `website/src/components/provider-catalog.ts` appears in that file only inside the doc comment at line 5. PR #4661's `check-website-inputs.sh` reads declarations and literals; it sees the declaration and the doc comment, but not the read itself. Harmless today — the path is declared — but the next helper written the same way with no doc comment is invisible to the guard, which is exactly the drift the guard exists to stop.

## Done means

`read_provider_catalog` uses one full-path literal (so the guard sees the read where it happens), and the "not one literal" caveat in `scripts/check-website-inputs.sh`'s header can be deleted. Small Rust change; `cargo test -p stella-cli docs_sync` covers it.

From PR #4661's report (2026-08-24). Related: #4632, #4606.

macanderson:
Four issues in the graph / search / guards neighbourhood.

### #4643 — two `stella-graph` vector tests raced on a wall-clock second

`upsert_file` read `now_unix()` **per row**, so an index pass slow enough to
cross a second boundary wrote two different `indexed_at` values.
`vectors::pending` orders by `(indexed_at DESC, path ASC)`, so the pass came
back reordered by where the walk happened to be when the second ticked, and the
`path` tiebreak the ordering relies on never got to decide. Two tests read that
order back and failed under the parallel runner, then passed serially.

The stamp now belongs to the pass. `index_tree_with_progress` and
`apply_changes` each read the clock once and hand the value to every file
through a new `Pass` struct, which also collects the five other values already
fixed for the whole pass — so `index_one` drops from eight positional arguments
to four and its `#[allow(clippy::too_many_arguments)]` goes with them.

### #4662 — a website path built from two joins was invisible to its guard

`read_provider_catalog` built its path from two `join` calls, so the literal
`website/src/components/provider-catalog.ts` appeared in that file only in a doc
comment. `check-website-inputs.sh` finds a website input by grepping the Rust
sources for the literal, so it saw the declaration and the prose but not the
read.

One literal at the read site now. The guard's header no longer offers the doc
comment as the workaround — but the limitation itself **stays**, rewritten
rather than deleted: a grep genuinely cannot see a path assembled at runtime,
and a header that stopped saying so would claim coverage the guard does not
have. What it names now is the remedy this change makes.

### #4572 — the six untagged measured constants

Each needed a judgement, and all six came back the same way: the number is a
**bound or a headroom choice** that an observation informs, not a value a
measurement produced.

| constant | verdict |
|---|---|
| `MAX_CYCLE_PERIOD` | the observation is periods 2 and 3; **4 is the headroom over it** |
| `FIRST_BYTE_TIMEOUT` | 90s is the comparator's watchdog, chosen to sit *above* the observed time-to-first-token |
| `STARVED_RETRY_CAP` | a heuristic bound — like the `REASONING_HEADROOM_TOKENS` beside it that already says so |
| `REPAIR_ATTEMPT_CAP` | its own doc calls it "a ceiling and not a target" |
| `DEFAULT_STOP_HOLDS` | two of the three are structural; the third is headroom over an observation with no sample size |
| `MAX_SUGGESTIONS` | 148 lines of output is *why* a cap exists; it does not say the cap is ten |

So each carries a one-sentence note saying it is a judgement — the issue's third
option, and what stops the next sweep re-asking. AGENTS.md is explicit that the
marker is not for a bound and that a vaguer marker devalues the marker
everywhere it appears, so **the count stays at three** rather than being raised
by marking six numbers whose measurement would have to be invented to state.

That is the one line of the issue's definition of done this PR does not meet
("`make measured-constants` reports a count above three"). It is a deliberate
refusal, not an oversight — the DoD's own first sentence offers the doc-note
branch for exactly this case, and the count line reads as the outcome its author
expected rather than an independent requirement. Flagged for a maintainer rather
than decided silently, which is why the commit says `Refs` and not `Closes`.

The two constants that **already** carry the marker now have pins that break
when the value moves. Both were named only by tests written *through* the
constant, so a revert left every one of them passing:

- `MAX_LENGTH_CONTINUATIONS` — a behavioural pin with hand-rolled counts: a
  third continuation must still be granted (the worst case the bench set
  produced needed it) and the fourth must be the last.
- `MIN_CACHEABLE_PROMPT_TOKENS` — the literal, because the constant *is* the
  threshold and every outcome computable from it moves with it.

### #4574 — exact-name search could not reach a section

A markdown section is indexed under its heading breadcrumb (#3103) and
`is_bare_identifier` refused every one of those, so no section had ever reached
the rung that answers with a fact. A query carrying `BREADCRUMB_SEPARATOR` is
now two candidate names: the query itself, and its tail past the first
separator — because `stella-graph` stores a section's name *without* the file
path and documents a citation as composing the two. Both are shapes that module
already defines; neither is inferred.

**The dotted TOML key stays refused, and that half is measured.** Censused over
786 search-shaped tool calls in this repository's own store (`.stella/private/
store.db`, 314 executions, 2026-08-24), a dotted rule newly offers a term for
**five** queries, and all five are the collision `is_bare_identifier` exists to
refuse:

| newly admitted term | the query it came from | what it is |
|---|---|---|
| `settings.json` | `#\[test\]\|fn .*() \{\|tempdir\|TempDir\|write_settings\|settings.json` | a filename |
| `driver.rs` | `driver.rs` | a filename |
| `hot.reload` | `fn reload\|hot.reload\|hot_reload\|live_reload` | regex `.` |
| `verifier.verdict` | `verifier.verdict\|agent.error\|retryable` | regex `.` |
| `verifier.verdict` | `verifier.verdict\|agent\.error.*retryable\|run\.failed` | regex `.` |

**Zero** are a table key. So the sketched widening buys, in a real log, five
graph round-trips that must miss and no certainty at all.

What that census **cannot** settle is demand: TOML tables became indexable days
before it was taken (#4571), so "nobody cited one" and "nobody could" are the
same measurement. It is evidence about the cost, and the cost is what decides a
widening whose benefit is still hypothetical. The doc comment says exactly that,
so the next reader does not re-derive the question — which is the issue's second
DoD branch, now with a measurement behind it instead of a hunch.

`BREADCRUMB_SEPARATOR` is `pub` and re-exported as a side effect: a consumer
that recognises a breadcrumb needs the same literal, and
`chunk_retrieval_witnesses.rs` had been carrying the only repeated copy of the
bytes because nothing could import it.

## Evidence

Targeted commands only; CI is the real gate.

**#4643 witness — fails on the old behaviour, passes on the new.**
`store::tests::one_pass_stamps_every_file_with_one_indexed_at` forces the second
boundary through the progress callback (which fires per file) rather than
injecting a clock.

- with `upsert_file` restored to a per-file `now_unix()`:
  `FAILED … [1787602539, 1787602540] means the clock is being read per file again`
- with the fix: `ok`

**#4643 flake, as the issue asks:** the two named tests, 20 runs under
`--test-threads=8` — `RUNS: 20  FAILURES: 0`. Whole crate:
`cargo test -p stella-graph --lib -- --test-threads=8` → **232 passed**.

**#4572 pins — both fail on a reverted value while their neighbours pass.**

- `MAX_LENGTH_CONTINUATIONS` set back to `2`: 10 passed, **1 failed** — only
  `the_continuation_allowance_stays_at_the_count_the_bench_set_produced`. Every
  other truncation test stayed green, which is the gap the issue names.
- `MIN_CACHEABLE_PROMPT_TOKENS` set to `2048`: 22 passed, **1 failed** — only
  `the_cacheable_floor_stays_at_the_value_the_census_confirmed`.

**#4574 witnesses — both fail with the breadcrumb branch disabled.**
`the_exact_symbol_rung_reaches_a_nested_markdown_section` and
`a_path_prefixed_citation_reaches_the_same_section`: 6 passed, **2 failed** on
the old gate; 8 passed with it. `a_dotted_table_key_is_still_answered_by_ranking_not_lookup`
passes both ways by construction — it is the refusal, and it is what fails if
the decision is reversed without re-measuring.

**#4662** — no witness, and this is why: the behaviour of neither the test nor
the guard changes. The literal moves so the guard can see the read where it
happens. `cargo test -p stella-cli docs_sync` → `provider_cards_match_the_registry
… ok`; `scripts/check-website-inputs.sh` → `OK — 7 declared website path(s), 3
of them Rust test inputs, all present.`

**Suites**

| command | result |
|---|---|
| `cargo test -p stella-graph --lib` | 232 passed |
| `cargo test -p stella-core --lib` | 1416 passed |
| `cargo test -p stella-model --lib` | 461 passed |
| `cargo test -p stella-tools --lib search` | 95 passed |
| `cargo test -p stella-tools --test chunk_retrieval_witnesses` | compiles; 1 ignored (needs an embedding backend) |
| `cargo clippy -p stella-graph -p stella-tools --all-targets -- -D warnings` | exit 0 |
| `scripts/check-measured-constants.sh` | OK — 3 measured constants, each named by a test |
| `scripts/test-measured-constants.sh` | 8 passed, 0 failed |
| `scripts/check-file-size.sh` | exit 0 |
| `make prose` | OK — none added |
| `make guards-fast` | exit 0 |

**Rustdoc, both forms, plain first**, over every crate touched
(`stella-graph`, `stella-tools`, `stella-core`, `stella-model`, `stella-cli`):

- `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps` → exit 0
- `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --document-private-items` → exit 0

The plain form earned its place here: the first run failed with
`public documentation for symbol_terms links to private item is_bare_identifier`,
which `--document-private-items` resolves and reports nothing for. Fixed by
naming the item instead of linking it.

## Not done

Nothing skipped. One deliberate refusal, stated above: #4572's
"`make measured-constants` reports a count above three" is not met, because
meeting it would mean marking numbers that are bounds. The substantive DoD —
each of the eight rows either marked with a test that breaks on a revert, or
carrying a note saying it is a judgement — is met.

Closes #4643
Closes #4662
Closes #4574
Refs #4572

## Summary by Sourcery

Stabilize graph indexing and search behavior, improve website-input guard coverage, and formalize constant-audit decisions with regression protection.

New Features:
- Enable exact-symbol search for breadcrumb-qualified Markdown sections, including path-prefixed citations, while continuing to reject dotted TOML keys and regex-like terms.

Bug Fixes:
- Make all files indexed in a single pass share one timestamp, restoring deterministic pending-vector ordering across clock-second boundaries.
- Expose website input paths as full literals so the website-input guard detects the actual file read.

Enhancements:
- Document the judgement-based nature of six unmarked constants and add regression pins for the two existing measured constants.
- Publicly re-export the Markdown breadcrumb separator for consumers and tests.

Tests:
- Add regression coverage for per-pass index timestamps, breadcrumb exact-symbol lookup, dotted-key refusal, continuation limits, and cacheability thresholds.


macanderson:
Answering both ❌ rows in Sourcery's assessment table. Neither is a fix, and
here is why for each.

---

### ❌ #4574 — "Make exact-symbol search reach nested TOML tables identified by dotted keys"

**Deliberately refused, and the refusal is a measurement rather than a
preference.** Sourcery reads the objective off the issue's problem statement;
the issue's own definition of done is a disjunction:

> Either the rung reaches a dotted / breadcrumbed section name with a witness
> test per symbol kind and evidence that prose queries are still refused, **or**
> a doc comment on `is_bare_identifier` states that section names are
> deliberately out of scope and why — so the next reader does not re-derive the
> question.

and its "Why it is not obviously worth fixing" section asks for the measurement
before shipping the widening:

> Measure the false-positive rate against a real query log before shipping it.

That measurement exists now. Censused over 786 search-shaped tool calls in this
repository's own store (`.stella/private/store.db`, 314 executions, 2026-08-24),
a dotted rule newly offers a term for **five** queries, and every one of them is
the collision `is_bare_identifier` was written to refuse:

| newly admitted term | the query it came from | what it is |
|---|---|---|
| `settings.json` | `#\[test\]\|fn .*() \{\|tempdir\|TempDir\|write_settings\|settings.json` | a filename |
| `driver.rs` | `driver.rs` | a filename |
| `hot.reload` | `fn reload\|hot.reload\|hot_reload\|live_reload` | the regex any-char |
| `verifier.verdict` | `verifier.verdict\|agent.error\|retryable` | the regex any-char |
| `verifier.verdict` | `verifier.verdict\|agent\.error.*retryable\|run\.failed` | the regex any-char |

**Zero** are a table key. In a real log the sketched widening buys five graph
round-trips that must miss, and no certainty at all.

The breadcrumb half shipped because it has the opposite profile: ` › ` is
`U+203A` between spaces, it occurs in **zero** of those 786 queries, and it
cannot occur in a heading or a question by accident. Same issue, two symbol
kinds, two different answers, each with its own evidence.

What the census **cannot** settle is demand — TOML tables became indexable days
before it was taken (#4571), so "nobody cited one" and "nobody could" are the
same measurement here. That one-sidedness is stated in the doc comment rather
than glossed, and the re-measure is filed as **#4739** with the census repro,
the file to change, and the test that would have to be inverted. So the question
is tracked, not dropped.

`a_dotted_table_key_is_still_answered_by_ranking_not_lookup` in
`crates/stella-tools/src/search/engine/tests/visibility.rs` asserts today's
behaviour, so reversing this decision cannot happen silently.

---

### ❌ #4662 — "Remove the obsolete 'not one literal' workaround caveat"

**Rewritten rather than deleted, on purpose** — and Sourcery's description of
what the header now says is accurate. The disagreement is only over whether
deleting it would be correct.

What was obsolete is the *workaround*, and that is gone. The old text ended:

> declare it, and say the path once in the doc comment above the helper, which
> is what `docs_sync.rs` does.

That sentence blessed a doc comment as a substitute for a literal at the read
site, which is exactly the drift #4662 was filed about, and it cited a file that
no longer works that way. It now names the remedy this PR implements instead:
write the full path as one literal where the read happens.

What is **not** obsolete is the mechanical limitation.
`check-website-inputs.sh` finds a website input with

```sh
git grep -ohE 'website/[A-Za-z0-9._/-]*' -- 'crates/**/*.rs'
```

A path assembled from several `join` calls names no `website/…` string, so the
grep cannot see it — true of any grep, and true of the next helper someone
writes. The paragraph sits under a heading that reads **"What it does NOT
check"**, whose whole job is to be honest about the guard's blind spots.
Deleting a true entry from that list would leave the header claiming coverage
the guard does not have, which is the failure mode the guard itself exists to
prevent, one level up.

The issue's "Done means" has two clauses; the first (`read_provider_catalog`
uses one full-path literal) is met exactly, and the second is met in substance:
the caveat no longer describes `docs_sync.rs` as an exception, because
`docs_sync.rs` is no longer one.

Happy to delete the paragraph outright if the maintainer prefers the shorter
header — it is a one-line change, and I would rather be told than guess.

Issue #2993, opened 2026-08-12
The similarity search cutoff score is a guess, not a measured value — measure it for each embedding model


macanderson:
## Problem

`SimilarityPosture::Semantic { admission_floor }` (the cutoff score below which a search result is dropped) is supposed to come from a model's *observed* relevant/irrelevant score distributions — the same discipline a provider's cache behavior claim applies when it names its test. `HttpEmbedder` declares `Semantic` and supplies a cutoff nobody measured:

```rust
// crates/stella-embed/src/http.rs
/// **This number is provisional and is not a measured separation point.**
const DEFAULT_ADMISSION_FLOOR: f32 = 0.25;
```

It's deliberately permissive, and its only current job is to drop the clearly-unrelated tail of an ordered list. That's a reasonable stopgap and it's stated in the code, the crate README, and the PR — but it's an unmeasured number in the one place the type system asked for a measured one, and it will get quoted as if it were measured the first time someone reads the declaration and not the comment.

## Files

- `crates/stella-embed/src/http.rs` — `DEFAULT_ADMISSION_FLOOR`, `HttpEmbedder::similarity_posture`, the `STELLA_EMBED_FLOOR` override.
- `crates/stella-embed/src/seam.rs` — `SimilarityPosture` and its contract.
- `crates/stella-embed/README.md` § "Semantics worth knowing" — the paragraph that must be rewritten once real numbers exist.

## How to measure

1. Take a real repository (this one) and a labelled query set: 30-50 questions in English, each with the file(s) a human says is the right answer. That's enough to see a separation.
2. Index with `graph_query op=semantic` under each candidate model (`voyage-code-3`, `text-embedding-3-small`, `text-embedding-3-large`, a local `nomic-embed-text`).
3. Record the cosine similarity of every (query, file) pair, split by whether the label says it's a match.
4. The cutoff is where the two score distributions separate; report the overlap, not just the point. If they overlap a lot, the right answer is `SimilarityPosture::Surface` for that model, the same as `HashEmbedder` reports today.

## Done looks like

- A per-model cutoff table in `stella-embed`, each entry citing the measurement that produced it.
- `DEFAULT_ADMISSION_FLOOR` either gone or documented as the value to use for an unknown model, with its own justification.
- Any model whose score distributions overlap declares `Surface` and says so.

## Definition of done

The done conditions this issue already states, as boxes the gate can read.
Nothing here is new scope, and nothing is ticked by the migration — the
[...]

Issue #4191, opened 2026-08-21
The search indexing speedup and its concurrency setting have not been measured against a real embedding service


macanderson:
## Problem

#4144 was opened on a **measured** premise: instrumented against the live `api.voyageai.com/v1/embeddings` endpoint (`voyage-code-3`), with payloads rebuilt from the real `.stella/private/codegraph.db`, it put the chunk rung's cost at roughly **35-45 minutes** to fill this repository's index, at a ~1.3s median per file and 5-7s for the 259 files carrying more than 32 symbols.

PR #4189 fixed the shape and proved the fix **structurally** — witnesses assert the request count collapses from one-per-file (62 calls for 60 files, measured on `main`) to `ceil(total_chunks / 32)`, and that up to `CHUNK_EMBED_CONCURRENCY` requests overlap.

It did **not** re-measure the wall clock. No figure was quoted, deliberately: the issue's own number describes the old code, and reporting it as if the branch had reproduced it would be exactly the surface-signal reasoning `CLAUDE.md` forbids. So the improvement is currently unquantified against a real backend.

## Why this matters beyond bookkeeping

The shipped fix carries unmeasured constants that cannot be settled without this measurement:

1. **`CHUNK_EMBED_CONCURRENCY = 8`** (`crates/stella-tools/src/search/engine.rs`). Chosen as a defensible bound, not as a measured optimum. Its own doc comment says it is "a knob on wall clock, not on correctness" — true, and also an admission that nobody has turned the knob and looked. The backend measurement in #4144 showed round-trip *count* was the constraint rather than backend throughput, which suggests headroom above 8, but "suggests" is not a measurement.

2. **The retry ladder's shape** (`MAX_ATTEMPTS = 4`, `INITIAL_BACKOFF = 500ms`, `MAX_BACKOFF = 8s`, `crates/stella-embed/src/http.rs`, added in #4189). These were sized to "clear an ordinary burst limit". Whether 8-way concurrency against Voyage actually provokes 429s at all — and if so, whether this ladder clears them or merely delays the abort — is untested against the real endpoint. If concurrency does trigger sustained rate limiting, the effective speedup could be well below the ~8x the structure implies.

## Definition of done

Measure the chunk rung end to end against a real embedding backend, before and after #4189, and record the numbers where they can be found again.

Suggested method, mirroring #4144's own instrumentation so the numbers compare like for like:

[...]

Issue #3096, opened 2026-08-13
The search relevance constants are unmeasured guesses, and the safety floor never filters anything


macanderson:
## The problem

`search` now decides how many results are relevant instead of returning a fixed ten (PR #3095). Three constants govern that decision, and none of them is a measured number:

| Constant | File | Value | Status |
|---|---|---|---|
| `DEFAULT_ADMISSION_FLOOR` | `crates/stella-embed/src/http.rs:59` | `0.25` | Pre-existing. Its own doc comment says it is "provisional and is not a measured separation point." |
| `DEFAULT_RELEVANCE_GAP_RATIO` | `crates/stella-embed/src/rank.rs` | `2.5` | Added by #3095, labelled provisional. |
| `DEFAULT_MIN_BOUNDARY_GAP` | `crates/stella-embed/src/rank.rs` | `0.05` | Added by #3095, labelled provisional. |

The floor is worse than provisional — it does nothing. In the ranking recorded on #3089, `voyage-code-3` scored *unrelated* files at `0.604`, so a floor of `0.25` admits everything and drops nothing. It has never done the job `SimilarityPosture` assigns it.

`SimilarityPosture::Semantic { admission_floor }` (the type that declares how a backend filters results by similarity) says this value should be set from measurement, the way a provider's cache behavior names its witness test. All three constants here fail that standard.

## Why this matters

The relevance cut now decides whether the model sees an answer at all. A wrong constant either hides the answer (cut too tight) or buries it (cut too loose) — and either way the tool reports confidently, with no sign anything went wrong. #3095 already hit this once: written with the gap ratio alone, `relevant_prefix` (the function that trims results) cut the real #3089 ranking down to a single result, and it wasn't the right one. That bug was caught only because someone had the measured score distribution to test against. No such distribution exists yet for the other two constants.

## What done looks like

1. A repeatable harness that, given a workspace with a populated `code_graph_chunk_vectors` table and an embedding key, runs a fixed set of labelled queries and prints the full score distribution for each — relevant chunks and irrelevant chunks separately. It belongs beside the existing `crates/stella-graph/tests/semantic_recall.rs`, and must be `#[ignore]`d or gated by an environment variable so `cargo test` without a key still passes.
2. A labelled query set. Start from the queries in #3089's definition of done, which have known correct answers in this repository:
   - "where is the prompt cache engaged on outbound requests" → `crates/stella-model/src/anthropic.rs`
   - a query whose answer is exactly one function
[...]

PR #4774, opened 2026-08-25
fix(stella-graph,stella-embed,stella-tools): a store image is walked once, not once per search — and the admission floor is measured

### LINKED ISSUE #3198: Build the cross-session, domain-grouped part of the search context cache
## Problem

#3163 was closed by the smallest workable slice: a session-lifetime, in-process cache of gathered context inside the `search` tool (`crates/stella-tools/src/search/cache.rs`), keyed by `(path, sha256(file bytes))`, bounded at 256 LRU entries. The issue's fuller proposal is still unbuilt. This issue tracks that remaining work:

- **Cross-session persistence**: the original issue asked for re-gathers "across the sessions of a team working the same tree" — that needs a store under `.stella/private/` (`stella-context`/`stella-store` territory), consumed by the tool through a port (architecture rule 1), not the in-process map that dies with the session.
- **Domain grouping**: use `.stella/domains.toml` (written by `stella init`) as the grouping key, and the unit a warm pass could pre-gather.
- **Git as the authority for invalidating persisted entries**: a persisted entry outlives the process, so "same bytes" needs an identity that's cheap to check at load time. The original issue sketches last-change commit shas per file, a batched `git ls-tree`/diff walk per generation, and an explicit bypass for dirty files. (In-process, the content hash made the dirty-file case a non-issue; on disk it comes back.)
- **Disclosure**: the original issue asks that an answer built from cache say so if that affects accuracy. The in-process slice never affects source-derived accuracy (source is re-read fresh), but a persisted cache may need that disclosure line.

## Where the seam is today

- `crates/stella-tools/src/search/enrich.rs::gathered_neighborhood` — the single gather point; a persistent tier plugs in behind this same function.
- `crates/stella-tools/src/search/cache.rs` — the identity choice and its reasoning, plus the precision boundary (#3196 tracks the cross-file staleness half).
- `crates/stella-tools/src/search.rs::report_with` — where the cache is threaded through and locked.

## Constraints found so far

- The expensive per-call costs in `search` are the embedding round trips and the `index_all` catch-up pass, both outside the scope of the gather cache — a persisted gather cache saves SQLite reads, so measure before building (the original issue's own metric: graph reads per call, wall clock).
- Architecture rule 7 (byte-stable answers): a cache hit must render byte-identically to the miss that populated it — the golden test shape already exists in `crates/stella-tools/src/search/cache.rs::tests`.
- Architecture rule 2: keep the decision logic pure; I/O stays behind the port.

## Done

Either the persisted, domain-grouped tier exists behind the `gathered_neighborhood` seam with the measurements that justify it, or there's a written decision on #3163's thread that the session-lifetime cache is where this stops.

Refs #3163, #3196, #3154, #3162.

### LINKED ISSUE #4385: The search tool caps results at 200 rows almost every time and it's slow
## Problem

The unified `search` tool caps its candidate list at 200 rows and prints "RANK CEILING — the candidate list filled its 200-row guard, so lower-scoring files beyond it were never considered". In this repository that guard fills on effectively every query, so the tool's result is always a truncated ranking — and it is slow.

## Evidence (session export `ses-1787465453163-60967` (`.stella/exports/session-2026-08-23-08-37-08.zip`, binary 0.9.143, OpenRouter `moonshotai/kimi-k3`); events re-read from `.stella/private/store.db`)

- 60 of 61 `tool_result` payloads for `search` in executions 225/242/248 carry the RANK CEILING banner (`events` table, `payload like '%RANK CEILING%'`).
- Latency from `tool_calls.duration_ms`: n=61, median 6 935 ms, p90 21 706 ms, max 33 450 ms. `read_file` median is 0.5 ms; `bash` median 123 ms.
- The model mined a lesson about it (reflections 250): "The search tool in this repo hits a 200-candidate rank ceiling and silently drops lower-scoring files; … rg/grep directly instead of trusting it" — i.e. the agent learned to route around the built-in.
- 225's five research delegates spent 60 searches; the two that died on the step cap / 900 s ceiling were search-heavy.

## Fix

Find where the 200 guard lives (`crates/stella-tools/src/search*`), make it a ranked top-K over the full candidate set rather than a first-200 fill, and profile the semantic path (embedding rank over the code-graph index) — 20–30 s per query inside a 16-step sub-agent is most of its budget.

## Done when

- On this repository, a query for a symbol that exists returns it without the banner.
- p90 under a stated target, measured by a bench loop over the 61 recorded queries (`tool_calls.args_json`).

macanderson:
## What

Four search-theme issues. Three had "measure first" at their centre, so the
measurements come first here too, each with what it was taken against.

### #4385 — `search` is slow, and a call was paying two full passes over the database

Two independent halves. The **banner** half landed in #4456 and is not this
PR's. The **latency** half is.

`search::engine::report_with` calls `open_or_build` on **every `search` call**,
and `CodeGraph::open` ran `PRAGMA quick_check` there — a walk of every page of
the image. `open_verified`'s own doc priced that as something "the writer's
open pays once, ahead of a tree walk that dwarfs it". Both halves are false on
the read path: it opens per call, and on this repository's index the walk is
180 MB against a hash-diff catch-up measured at 78–97 ms warm. The recorded
session paid it sixty-one times.

`store::image_check` memoizes the verdict, keyed on the main database file's
`(length, mtime)` — **not** on its path, which was the first version and which
broke `a_store_corrupt_only_in_its_data_pages_is_quarantined_at_open`. That
test is the contract: a store damaged *after* it was verified must be caught by
the next open. Keying on the image's own stamp keeps that and keeps the saving,
because WAL mode puts ordinary writes in the `-wal` sidecar and leaves the main
image — the thing `quick_check` walks — untouched. What is given up is stated in
that module: damage written inside the same `(length, mtime)` tick.

Also lands the loop the issue asks for,
`crates/stella-tools/tests/search_latency_replay.rs`, with the sixty-one
recorded queries as a committed fixture extracted from that session's
`tool_calls.args_json`.

**No p90 is quoted, and the issue is not closed.** See "Not done".

### #2993 — the admission floor is a guess (advanced, not closed)

It is now a measurement for one model, and that measurement says the floor
cannot do the job.

`tests/relevance_calibration.rs` was run against `voyage-code-3` over this
repository's own index — 28 744 chunks, four labelled queries, 40 candidates
deep. The relevant and irrelevant distributions **overlap**: tightest
separation **−0.0439**, with one query's labelled answer ranking 39th at 0.5978
beneath 38 irrelevant chunks scoring 0.6006–0.6436, and two of the four
returning no labelled answer in 40 candidates at all. No floor separates those,
so `voyage-code-3` declares `SimilarityPosture::Surface`: its cosines order
candidates and certify none, exactly as `HashEmbedder` already reports. #2993
named that outcome in advance.

- `MEASURED_FLOORS` is the per-model table, carrying the `MEASURED:` marker and
  pinned by tests.
- `DEFAULT_ADMISSION_FLOOR` stays as the value for an unmeasured model, and its
  doc now states that scope instead of apologising for the value.
- `HttpEmbedder::new` takes `Option<f32>` so "the operator set a floor" and
  "nobody did" stop being the same value. An explicit `STELLA_EMBED_FLOOR`
  outranks the table either way.
- **No constant's value moved.** `DEFAULT_MIN_BOUNDARY_GAP`'s doc records the
  observed boundary drops (0.0140 and 0.0001, both under its 0.05) and why two
  frontiers pointing opposite ways is a sample to report rather than one to tune
  against.

What made the run affordable is `STELLA_CALIBRATION_INDEX`, added to the
harness: the ~11M paid tokens a previous session declined to spend were the
*re-embedding*, and an index this workspace had already filled reduced the run
to four query embeddings.

### #3198 — the cross-session gather cache

The issue asked for the measurement before the build — graph reads per call,
and wall clock — and accepts "a written decision that the session-lifetime
cache is where this stops" as a close. That is what it gets.

One gather is **3.7 ms**: the three indexed queries `file_neighborhood` issues,
over a random 200-file sample of `crates/**`. A call enriches whole blocks until
a 9 000 character budget is spent, so the ceiling on what any gather cache can
save is tens of milliseconds. The terms it would save against are one to two
orders of magnitude larger: chunk ranking scan 220–296 ms, image page walk
333 ms, `index_all` catch-up 78–97 ms, embedding round trip 111–202 ms.

The structural half survives a faster machine: `observe_generation` empties the
cache whenever *any* file is re-indexed, and a persisted tier inherits that key,
so the cross-session case the issue wants is exactly the case where every
teammate's commit invalidates every persisted entry.

## Evidence

Targeted commands only; no workspace build. Machine load average was 24–65
throughout (concurrent agent worktrees), which is why no wall-clock comparison
is claimed anywhere in this PR.

**Witnesses, both directions:**

|---|---|---|
| `stella-graph --test image_verification_memo` | `left: 10, right: 2` (ten opens, ten walks) | `ok` |
| `stella-embed the_measured_backend_declares_the_posture_its_distribution_earned` | `left: Semantic { admission_floor: 0.25 }, right: Surface` | `ok` |

Each fail direction was produced by reverting only the mechanism (the
`already_walked` early return; the `MEASURED_FLOORS` row) and restoring it after.

**Renamed test** (named here for `check-deleted-tests`, which keys on the bare
function name): `the_http_backend_declares_a_semantic_posture` →
`an_unmeasured_backend_declares_a_semantic_posture`. Same assertion, plus an
exact check that the posture is `Semantic { DEFAULT_ADMISSION_FLOOR }`. The new
name is what the test now distinguishes: an *unmeasured* backend is the one that
still declares `Semantic`, and `the_measured_backend_declares_the_posture_its_distribution_earned`
covers the other side.

**Suites:**

- `cargo test -p stella-graph --lib` → 232 passed, 0 failed. Includes
  `a_store_corrupt_only_in_its_data_pages_is_quarantined_at_open`, which the
  path-keyed first version of the memo broke and the stamp-keyed one passes.
- `cargo test -p stella-graph --test image_verification_memo` → 1 passed.
- `cargo test -p stella-embed --features http --lib` → 47 passed, 0 failed.

**Gates:**

- `cargo clippy -p stella-graph -p stella-embed --all-targets --all-features -- -D warnings` → exit 0
- `cargo clippy -p stella-tools --all-targets -- -D warnings` → exit 0
- `RUSTDOCFLAGS="-D warnings" cargo doc -p stella-embed -p stella-graph -p stella-tools --no-deps` → exit 0 (**plain form run first**; it caught four public-to-private intra-doc links in `http.rs` that the `--document-private-items` form resolves silently)
- the same with `--document-private-items` → exit 0
- `cargo fmt --all`, `make guards-fast` → exit 0. `make prose` initially failed on
  four constructions this PR added; they were rewritten, not baselined.
- `make measured-constants` → OK, 4 measured constants, each named by a test.

**Measurement provenance:** every number above was taken against a copy of this
workspace's own `.stella/private/codegraph.db` (1 585 file vectors, 28 813 chunk
vectors, `voyage-code-3@1/1024/l2`) so nothing touched a live session's
database. The recorded latency row in #4385 was re-derived from
`.stella/private/store.db` and reproduces the issue's figures exactly
(n=61, median 6 935 ms, p90 21 361 ms, max 33 450 ms).

## Not done

**#2993 is `Refs`, not `Closes`** — Sourcery's first ❌ row is correct and this
is the fix for it rather than a rebuttal. The issue asks for the measurement
*per model* and names four; this is one. `text-embedding-3-small`,
`text-embedding-3-large` and a local `nomic-embed-text` each need their own
vendor key and their own run, and the table gets a row for each only when
somebody has done that. Its three "Done looks like" bullets are all satisfied
for the model that was measured, which is why the row is worth advancing rather
than deferring whole.

**#4191 — the chunk pass's speedup and `CHUNK_EMBED_CONCURRENCY`, skipped.**
The measurement it specifies is a cold backfill of this repository against a
paid embedding backend, run before and after #4189 and swept over
{4, 8, 16} — six full passes, each the ~11M tokens `relevance_calibration.rs`'s
own header calls a maintainer's spend decision. That is a benchmark rig, and
fabricating a number instead is the one thing the issue exists to prevent.
Nothing in this PR touches it.

**#4385 is `Refs`, not `Closes`.** Its definition of done asks for "p90 under a
stated target, measured by a bench loop over the 61 recorded queries". The loop
now exists and was run; the machine could not answer. Its load average rose from
24 to 64 during the run and per-query wall clock rose monotonically with it,
from 15 422 ms on the first query to 201 500 ms on the fiftieth. A series that
climbs 13x while the machine's load triples underneath it measures the machine,
so a median over it would be arithmetic rather than measurement and none is
quoted. The harness's Status section records the attempt, because "we tried and
the machine could not answer" is a different state from "nobody has run it".

That run did establish two things that do not depend on load, both now written
into the code: every query came back 200+ hits deep (203–256 across fifty), and
`DEFAULT_MIN_BOUNDARY_GAP`'s 0.05 is 3.5x the widest boundary drop the
calibration observed — so `relevant_prefix` never cuts on this corpus, every
ranking runs to its end, and `search`'s RANK CEILING note therefore still fires
on nearly every query even after #4456 narrowed its condition. Re-running the
replay on a quiet machine is what closes #4385.

Closes #3198
Refs #4385
Refs #2993
Refs #3096

### DIFF
--- a/crates/stella-cli/src/agent/graph/tests.rs
+++ b/crates/stella-cli/src/agent/graph/tests.rs
@@ -107,7 +107,9 @@ async fn init_embeds_the_index_without_any_semantic_query_being_issued() {
         "concept-2",
         None,
         2,
-        0.25,
+        // Only the fingerprint is wanted here, and the posture does not
+        // enter it: no operator floor.
+        None,
     )
     .fingerprint()
     .id();
@@ -299,7 +301,9 @@ async fn init_embeds_every_chunk_without_any_semantic_query_being_issued() {
         "concept-2",
         None,
         2,
-        0.25,
+        // Only the fingerprint is wanted here, and the posture does not
+        // enter it: no operator floor.
+        None,
     )
     .fingerprint()
     .id();
--- a/crates/stella-embed/README.md
+++ b/crates/stella-embed/README.md
@@ -112,17 +112,31 @@ silent quality regression.
   attempts were made so a slow pass is not mistaken for one that never waited.
   This matters because the chunk embedding pass keeps several requests in
   flight (#4144), and concurrency is precisely what provokes a rate limit.
-- **The default admission floor is provisional, not measured.** The
-  `SimilarityPosture` contract asks for a floor derived from a model's observed
-  relevant/irrelevant score distributions on a real corpus, and no such
-  measurement exists yet for these backends on code. The shipped value is
-  deliberately permissive and its job is to drop the obviously-unrelated tail
-  from an ordered list, not to certify anything. The measurement is written
-  and reproducible — `crates/stella-tools/tests/relevance_calibration.rs`,
-  `#[ignore]`d because it needs a real key and a full embedding pass — and
-  running it once per backend is what turns this paragraph into a number
-  (#3096). It lives in `stella-tools` rather than here so this crate's own
-  tests stay hermetic and its leaf status is untouched.
+- **`voyage-code-3` admits nothing, and that is a measured result.** On
+  2026-08-24 `crates/stella-tools/tests/relevance_calibration.rs` was run
+  against it over this repository's own index (28 744 chunks, four labelled
+  queries, 40 candidates deep). The relevant and irrelevant distributions
+  **overlap**: tightest separation −0.0439, with one query's labelled answer
+  ranking 39th at 0.5978 beneath 38 irrelevant chunks scoring 0.6006–0.6436,
+  and two of the four queries returning no labelled answer in 40 candidates at
+  all. No floor separates those, so the model declares
+  `SimilarityPosture::Surface` — its cosines order candidates and certify none,
+  exactly as `HashEmbedder` already reports. #2993 named that outcome in
+  advance as the one to expect.
+- **Every other backend is unmeasured and says so.** `MEASURED_FLOORS` in
+  `src/http.rs` is the per-model table, and a model absent from it keeps
+  `DEFAULT_ADMISSION_FLOOR` (0.25) and declares `Semantic`. That number is
+  still not a separation point; what it now has is a stated scope — it is the
+  value for a model nobody has measured, and a permissive floor is all such a
+  backend can defend. `text-embedding-3-small`,
+  `text-embedding-3-large` and a local `nomic-embed-text` are the three #2993
+  also names, and each needs its own run of the harness. An explicit
+  `STELLA_EMBED_FLOOR` outranks the table either way: the operator is then
+  making the claim, about a corpus the measurement did not see.
+- **The harness lives in `stella-tools`, not here**, so this crate's own tests
+  stay hermetic and its leaf status is untouched. Point it at an already-filled
+  index with `STELLA_CALIBRATION_INDEX` and one run costs a query embedding per
+  labelled query instead of a full pass over the repository.
 
 ## God files — do not add lines
 
--- a/crates/stella-embed/src/http.rs
+++ b/crates/stella-embed/src/http.rs
@@ -75,23 +75,51 @@ const INITIAL_BACKOFF: Duration = Duration::from_millis(500);
 /// often a misconfigured proxy than a real instruction.
 const MAX_BACKOFF: Duration = Duration::from_secs(8);
 
-/// The admission floor applied when nothing overrides it.
+/// The admission floor for a model **nobody has measured**, and for nothing
+/// else.
 ///
-/// **This number is provisional and is not a measured separation point.** The
-/// [`SimilarityPosture`] contract asks for a floor derived from a model's
-/// observed relevant/irrelevant score distributions, and no such measurement
-/// exists yet for these backends on a code corpus. It is deliberately
-/// permissive: its job today is to drop the obviously-unrelated tail from an
-/// ordered list, not to certify anything.
+/// It is still not a separation point and still certifies nothing; what has
+/// changed is that its scope is now stated. A model whose distributions have
+/// been measured takes its posture from [`MEASURED_FLOORS`] and never reaches
+/// this number, so this is the value for an unknown backend and the
+/// justification for it is that a permissive floor is all an unmeasured
+/// backend can defend: it drops the obviously-unrelated tail from an
+/// ordered list and admits everything else, which understates the backend
+/// rather than overstating it.
 ///
-/// The measurement that settles it is written and reproducible:
-/// `crates/stella-tools/tests/relevance_calibration.rs` prints the labelled
-/// relevant/irrelevant score distributions over this repository for whatever
-/// backend is configured, and names the number to write here (#3096). It is
-/// `#[ignore]`d because it needs a real key and a full embedding pass. Until
-/// somebody runs it, this paragraph stays — deleting it without the
-/// measurement would turn an honest disclosure into a silent assumption.
+/// The measurement that would replace it for a given model is written and
+/// reproducible: `crates/stella-tools/tests/relevance_calibration.rs` prints
+/// the labelled relevant/irrelevant distributions over this repository for
+/// whatever backend is configured (#3096, #2993). It is `#[ignore]`d because
+/// it needs a real key.
 const DEFAULT_ADMISSION_FLOOR: f32 = 0.25;
+
+/// Which models have had their score distributions measured, and what the
+/// measurement found.
+///
+/// MEASURED: 2026-08-24, `voyage-code-3@1/1024/l2` over this repository's own
+/// index (28 744 chunks), four labelled queries from
+/// `crates/stella-tools/tests/relevance_calibration.rs`, 40 candidates deep
+/// each. Tightest separation across the set **-0.0439**: on "why is the Rust
+/// toolchain pinned to an exact version" the labelled answer ranked 39th, at
+/// 0.5978, under 38 irrelevant chunks scoring 0.6006-0.6436. The widest
+/// positive separation was +0.0140. Two of the four queries returned no
+/// labelled answer in 40 candidates at all, which makes their separation
+/// unmeasurable rather than good.
+///
+/// `Some(floor)` is a floor a measurement put between the two distributions.
+/// `None` is the other outcome, and the one this table currently records: the
+/// distributions **overlap**, so no floor separates them and the model must
+/// declare [`SimilarityPosture::Surface`] — scores order candidates and admit
+/// none, exactly as `HashEmbedder` already reports. #2993 named that outcome
+/// in advance as the one to expect, and it is the one that came back.
+///
+/// A model absent from this table is unmeasured, keeps
+/// [`DEFAULT_ADMISSION_FLOOR`], and says so in its own doc comment above. An
+/// explicit `STELLA_EMBED_FLOOR` outranks this table for either kind of
+/// entry: the operator is then making the claim, about a corpus this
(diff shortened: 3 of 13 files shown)


macanderson:
Answering every ❌ row in the assessment table.

**#2993 / "per-model measurement table is incomplete" — fixed, by correcting the
claim.** The row is right: four models are named and one was measured. The
trailer is now `Refs #2993`, not `Closes`, in the PR body and on the commit, and
the "Not done" section says which three models are outstanding and what each
needs (its own vendor key and its own run of
`crates/stella-tools/tests/relevance_calibration.rs`). The branch was rebuilt so
the commit trailer matches; the tree is byte-identical to the previously pushed
head.

The three remaining ❌ rows are all **#4385**, and they are answered rather than
fixed, because this PR never claimed to close it — the body carries `Refs #4385`
and a "Not done" section stating so before the review ran.

**"Replace the 200-row guard with a ranked top-K…"** — this objective is already
shipped and is not this PR's work. #4456 rewrote the guard's disclosure and
`stella_embed::rank::top_k` already scores the whole corpus and truncates last,
so the rows past the ceiling are by construction the ones that scored below
every row inside it. What this PR adds on that half is the *measurement of why
the banner still appears*, which was not previously known: the calibration run
shows `DEFAULT_MIN_BOUNDARY_GAP`'s 0.05 is 3.5x the widest boundary drop
observed on this corpus (0.0140), so `relevant_prefix` never cuts, every ranking
runs to its end, and `ceiling_note`'s condition — "the boundary kept everything
it was given" — is therefore satisfied on nearly every query. That is now
written into the constant's doc comment. Retuning it on two frontiers, one of
them a query whose answer ranked 39th, would re-open the exact #3089 failure the
constant exists for.

**"Reduce search latency… end-to-end p90 below a stated target"** and **"use the
benchmark to demonstrate the issue is resolved"** — the harness and its
sixty-one-query fixture are here and were run; the machine could not answer.
Load average rose from 24 to 64 during the run and per-query wall clock rose
monotonically with it, 15 422 ms on the first query to 201 500 ms on the
fiftieth. Quoting a median over a series that climbs 13x while the machine's
load triples underneath it would be arithmetic, not measurement, so none is
quoted and the issue stays open. The row is a correct reading of the state.

On "it also leaves the full chunk-vector scan in place": deliberately, and the
number is now recorded where the next person to weigh it will find it —
`vectors::chunks::rank_chunks`'s doc gives 118 MB read and decoded per call,
220-296 ms warm. `vectors::stored` had justified brute force at *file* scale
(1 585 rows, 6.5 MB) and the chunk rung inherited that justification for a
corpus twenty times larger without re-examining it. Replacing it with an ANN
structure is a second index to keep correct and is a separate change; the
cheaper term went first.

macanderson:
@sourcery-ai review

macanderson:
Note for whoever reads the assessment table above: it was generated against the **first** pushed head. The branch was then force-pushed (same tree, byte-identical — `git diff` against the old head is empty; only commit trailers changed, `Closes #2993` to `Refs #2993`), and a re-review came back "you've used your own review budget of 250,000 diff characters for the last 7 days". So the table will not refresh on this PR.

Every ❌ row on it is settled in the comment above: the #2993 row is fixed (trailer corrected, three unmeasured models named in "Not done"), and the three #4385 rows are answered — the PR never claimed to close #4385 and says so in its body.

All required checks are green on the current head, including `check-deleted-tests`, which failed once on the rename `the_http_backend_declares_a_semantic_posture` -> `an_unmeasured_backend_declares_a_semantic_posture` and passes now that the description names it.

Issue #4641, opened 2026-08-24
docs/tools read_file output_schema omits the data field the tool now returns


macanderson:
## Problem

`docs/tools/read_file.toml`'s generated `output_schema` block documents the `ToolOutput` envelope without the `data` field, which `read_file` now fills with `{lines_shown, lines_total}` (PR #4631). The tool-docs gate is green because `input_schema` is unchanged — so the docs describe less than the tool sends and nothing fails.

## Done means

The generator behind `make tool-docs-update` includes the `data` member (for every tool that ships one, not read_file alone), the regenerated docs are committed, and the tool-docs check covers output_schema drift the way it covers input_schema. If output_schema is deliberately envelope-only, that decision goes in the generator's doc comment instead.

From PR #4631's report. Related: #4297, #4034.

macanderson:
Work exists but is **unfinished and not merged** — recording where it is so it is not redone or lost.

**The defect is worse than this issue states.** `output_schema()` derives each arm from a *single* serialization sample, and both `content` and `class` are `skip_serializing_if`. So the generated pages document `ok` as `{content}` with `additionalProperties: false` — the schema declares a real `read_file` result **invalid** — and `error.class` is missing the same way. Not just a missing `data` field.

**Branch: `fix/c16-tooldocs-4641`** (`efe3d3381`, pushed). It rewrites `output_schema()` to read each arm off two samples and adds a witness asserting set equality against what `ToolOutput` actually serializes. The commit message says it is incomplete.

**Why it is not a PR:** `output_schema` is a *generated* artifact, so the change needs `make tool-docs-update` committed alongside it or the `tool-docs` gate step goes red. That regeneration needs a `stella-cli` test-binary build, which the machine could not complete — it was 23 GB into 24 GB of swap, and the `rustc` doing the codegen took 28 CPU-seconds across 19 wall minutes before being stopped rather than starve the sibling agents.

**To finish:** check out the branch on a machine with memory → `make tool-docs-update` → read the diff (every page changes; the envelope is per-tool identical) → commit → open the PR. The witness runs both ways via an ablation that reverts the derivation to one sample per arm.

From PR #4780's work (2026-08-24).

Issue #4580, opened 2026-08-23
Streaming text in the benchmark transcript viewer splits into two entries when another event arrives mid-stream


macanderson:
## Problem

`arenabench/arenabench/transcript.py` clears all open coalescing runs on every non-delta event (`state.open_entries.clear()`), so a response streamed across a `block_registered` (any kind) renders as two entries instead of one. Pre-existing; surfaced while wiring the prompt row (PR #4577), which pinned the current behavior in `test_the_goal_block_does_not_break_text_coalescing` (consolidation still supersedes) without changing the split.

## Done means

A text/reasoning run survives an interleaved block registration as one entry (or the split is recorded as deliberate with the reason in the module doc); the pinning test updated to the chosen semantics both ways.

Related: PR #4577, #4039.

macanderson:
Not actionable in this repo any more — the subject file left it. ArenaBench was ejected into https://github.com/macanderson/arenabench by PR #4642 (merged 2026-08-24T13:25Z, commit 5e964d05a), which deleted `arenabench/arenabench/transcript.py` (914 lines) along with the rest of the harness.

Re-runnable evidence against `origin/main`:

- `git show origin/main:arenabench/arenabench/transcript.py` → `fatal: path ... does not exist in origin/main`
- `git log origin/main --oneline -- arenabench` → top entry `5e964d05a chore(arenabench): eject the harness into macanderson/arenabench (#4642)`
- `git show 5e964d05a --stat | rg transcript.py` → ` arenabench/arenabench/transcript.py | 914 --------`

**The defect itself is NOT fixed** — it moved. In the new repo the same clearing is still there at `arenabench/transcript.py:638` and `:645`:

`gh api repos/macanderson/arenabench/contents/arenabench/transcript.py --jq .content | base64 -d | rg -n open_entries.clear` → `638: state.open_entries.clear()` / `645: state.open_entries.clear()`

Closing here only because this tracker can no longer host the fix; the work should be refiled against macanderson/arenabench with this body intact (the pinning test named in it, `test_the_goal_block_does_not_break_text_coalescing`, travelled with the history too).

Issue #4297, opened 2026-08-22
A file read entry in the transcript should show how many lines were read


macanderson:
## Problem

A `read` head in the v2 transcript states its path and nothing about its size.
#4180 removed the `Extent` field from `EventKind::Read` because the column was
expressible and unreachable — the head's measurement resolves only through the
inline-diff reference a *mutation* stamps, and `read_file` emits no
`FileChange` to stamp one from. Deleting the inert arm was the correct first
move; it does not mean a read's size is not worth stating.

SPEC 6.2's read row originally asked for it:

```text
 │ ▸ read …/lifecycle.rs · 221 lines                    ⚡3ms · ↵ open
```

`design/tui-v2/SPEC.md` §6.3's `read` row now reads `▸ read <path> · ⚡ms ·
↵ open`, with a note pointing here. Restoring the column means giving the
number a real producer, not re-adding the field.

## The decision this carries

`crates/stella-tools/src/read.rs` returns `ToolOutput::ok(numbered)` with **no
`data` payload**, so the line count exists only inside prose. Two numbers are
in play and the issue is not done until it says which one the row states:

- **`shown`** — the lines actually handed to the model, which is what a
  truncated read put in context.
- **`total`** — the lines in the file.

`read.rs` already has both (they appear in the truncation footer it appends).
**Do not parse that footer** — counting the rendered result body is the exact
substitution #2290 established as a defect for mutation counts, and
`TranscriptEntry::ToolResult::full` is capped at `OUTPUT_BUDGET` besides. The
number must ride the wire as structured data.

A plausible shape: `ToolOutput::Ok { data: Some(json!({ "lines_shown": n,
"lines_total": m })) }`, with the head rendering `· n of m lines` when they
differ and `· n lines` when they do not — so a truncated read is visibly
truncated rather than silently understating the file. That is a proposal, not a
decision.

## Files

- `crates/stella-tools/src/read.rs` — the producer. Currently
  `ToolOutput::ok(numbered)`, no `data`.
- `crates/stella-tui/src/model.rs` — the `AgentEvent::ToolResult` arm, which
  stamps an `InlineDiffRef` only for `summarize::is_file_mutation(&name)`. A
  read's number needs its own carrier; do not widen the mutation predicate.
- `crates/stella-tui/src/v2/transcript.rs` — `EventKind::Read` (a unit variant
  since #4180) and `kind_detail`.
[...]

Issue #2380, opened 2026-08-08
arenabench: eject into a standalone, multi-tenant, free SaaS on AWS — durable run store, real login, agent/dimension SDK


macanderson:
## The spec

**[`arenabench/docs/saas-architecture.md`](https://github.com/macanderson/stella/blob/claude/arenabench-saas-architecture-gzuedf/arenabench/docs/saas-architecture.md)** — on branch `claude/arenabench-saas-architecture-gzuedf`. This issue is the tracking handoff; the document is the design.

## Problem

1. **Stella's benchmark progress is not actually tracked.** `series.py` computes a real cross-match comparability key, then reads it out of an in-memory registry populated only by matches *this process* launched (#1707). Restarting `arenabench serve` loses every match; `arenabench run` never registers one. "Is Stella getting better, commit over commit" is answered today by hand-written scripts pointed at a job tree, or not answered — which is how `4bed110003d8` (5/5) vs `61e01ca06fc7` (3/5), ~12 commits apart, got read as a one-change regression.

2. **ArenaBench cannot be used by anyone who is not us.** Seven modules carry Stella-shaped assumptions: `catalog.py` parses `crates/stella-model/src/catalog.rs` for the model select (an error state with no checkout), `sut.py` + `sut_build.py` are 1,268 lines of "clone Stella and `cargo zigbuild` it", `proof.py` reads Stella's proof rail, `adapter.py` stages `stella_harbor`. `pip install arenabench` lands a user in a tool whose flagship contestant they cannot seat.

3. **There is nowhere to put a result that outlives a laptop.** Runs live in `~/.arenabench/matches/<id>/`. A benchmark whose whole argument is "a number is only comparable to another number from the same digest" has no mechanism for two people to establish that they ran the same digest.

## The decision that shapes the rest

Free + multi-tenant + hosted execution is not simultaneously satisfiable, and the spec refuses to pretend otherwise. From our own infra: `assert_sole_dockerd` means one trial owns a whole `m6i.xlarge` (4 vCPU / 15 GB), 15–30 min ⇒ **~$0.05–0.10/trial**; the README's 12×12 shape is 144 trials ⇒ **~$10–15 per match**, ~$300–450/month for one user's nightly.

So ArenaBench splits:

- **Control plane** — hosted, free, ours: identity, run index, results, boards, artifacts, SDK API. **Under $100/month at 1,000 tenants** (§12).
[...]

PR #4780, opened 2026-08-25
feat(stella-tui): the AGENTS page keeps its caret and answers on the page; the v1 read row's decision recorded

### LINKED ISSUE #4626: AGENTS page polish: no terminal caret in its composer, and command replies land behind the page
## Problem

Two rough edges on the new full-frame AGENTS page (`←` twice on SESSION; `crates/stella-tui/src/v2/agents_page.rs`):

1. **No hardware cursor.** The page suppresses the terminal caret like the modal overlays do, but its composer is a real text input — IME candidate windows, screen readers, and cursor-following tmux panes have nothing to anchor to while typing a task. `render_deck`'s early page branch would need a cursor position computed from the page composer's `composer::layout` (the main path's `composer_cursor_position` is the template).

2. **Replies print behind the page.** A page-submitted `/model` or `/models` runs queue-free and replies over `ShellEvent` into the SESSION transcript; the page shows a one-line notice pointing there. Better: render the last ShellEvent reply (or a short tail) in a page-local pane so picking a model never requires leaving the page to read the confirmation.

## Verify

- Open the page, type — no caret today (1).
- On the page, submit `/model` — the summary lands in the transcript; the page only shows "sent — its reply prints in the transcript" (2).

## Done when

- The caret tracks the page composer whenever no overlay (help, parked ask) sits above the page.
- A page-submitted command's reply is readable without closing the page.
- Golden `page_agents` re-blessed with the diff read.

### LINKED ISSUE #4640: v1 read result row still shows no line count after the v2 head got one
## Problem

PR #4631 gave the v2 read head a real size producer (`read_file` reports `lines_shown`/`lines_total` as structured `data`; the v2 deck renders `· n of m lines`). The v1 result row in `crates/stella-tui/src/render/entry.rs` still states no line count — #4297 left v1 as an open question and the PR did not take it.

## Done means

Decide: either the v1 row renders the same `ReadSize` carrier (it is already on `TranscriptEntry::ToolResult`, so this is a render-only change), or the decision to leave v1 without it is recorded at the render site. If rendered, golden frames for any affected v1 snapshot are re-blessed with the diff read.

From PR #4631's report. Related: #4297.

macanderson:
### #4640 — the v1 read result row and the `ReadSize` carrier

#4297 put a `ReadSize` on `TranscriptEntry::ToolResult` and left the v1 result
row an open question. **Answered: the row does not restate it**, and the answer
is now written at the render site (`crates/stella-tui/src/render/entry/tool.rs`,
`result_body`'s doc) instead of sitting there as an absence.

The count is already on the row above. A v1 `ToolStart` head routes through
`v2::transcript_source::head_rows` (`render/entry.rs`'s `v2_rows`), which
resolves the number off this very result, so the block states `· 200 of 500
lines` on its head. Putting it on the result row too would be the same
measurement in two cells of one block — the defect that metric column already
had once, when its size chip duplicated the count on the reveal hint under it.

Pinned from both sides by `a_reads_line_count_is_stated_once_in_its_block`. The
negative half alone would also pass on a deck that had lost the count
altogether, which is the failure the issue was worried about; the test also
asserts the result row's own `⋯ 3 lines · ctrl+o` is untouched, so what it
forbids is the *coverage* being restated, not counting in general.

No behaviour changed, so there is no witness in the fail→pass sense — this is a
decision recorded plus a pin. The pin's bite is demonstrated below.

### #4626 — the AGENTS page: a caret in its composer, and its replies on the page

Both halves, in `crates/stella-tui/src/v2/agents_page.rs` and `render_deck`'s
early page branch.

**1. The hardware caret.** The page took the frame's early return and
positioned no cursor at all, so ratatui hid the caret — while the page's
composer is a real text input, and IME candidate windows, screen readers and
cursor-following tmux panes all anchor to that caret.
`agents_page::composer_cursor` computes it from the same geometry `render`
draws with (one `Foot`, measured once), for the reason the deck's own
`composer_scroll_first` was split out in the first place: a caret one row off
from its text is worse than no caret, because the candidate window then opens
over the wrong line. That helper is now shared rather than copied, and the
page's composer rows are windowed through it — so a task longer than three rows
keeps the line being typed on screen, where it previously drew rows `0..` and
clipped. Withheld under help, a parked ask, or one of the page's own two menus:
the rule the deck composer already follows.

**2. The reply.** A page-submitted `/model` answers over `ShellEvent` into the
session's fold, which is *behind* the page — so choosing a model meant leaving
the page to find out whether it took. `AgentsPage::reply_from` marks the
transcript as the command leaves; the foot renders the tail past that mark
under a dim `REPLY` heading. A mark, not a copy of the reply: the rows are
built by the deck's own `entry_lines`, so the page cannot come to disagree with
the transcript it is quoting. Folded backwards from the newest entry and
stopped at six rows, because the mark can stand while a turn appends without
limit. The notice drops its now-wrong pointer (`/model sent`, not `… its reply
prints in the transcript`).

## Evidence

Targeted commands only — no workspace build ran here.

| Command | Result |
|---|---|
| `cargo fmt --all --check` | pass |
| `cargo test -p stella-tui --lib -- render::tests::result_row deck_render::tests` | pass — 26 passed, 0 failed |
| `cargo test -p stella-tui --test deck_render_snapshots` | pass — 21 passed, **no golden re-blessed** |
| `cargo clippy -p stella-tui --all-targets -- -D warnings` | pass |
| `RUSTDOCFLAGS=-Dwarnings cargo doc -p stella-tui --no-deps` (plain form, run first) | pass |
| `RUSTDOCFLAGS=-Dwarnings cargo doc -p stella-tui --no-deps --document-private-items` | pass |
| `make no-scratch left-behind dead-code-allows typed-errors tool-error-class transcript-surfaces file-size god-files module-reachability doc-links prose` | all OK |

`origin/main` is merged in twice (`539d5122f`, then `#4763`'s prose-guard rework); one conflict, in
`crates/stella-tui/src/deck_render/tests.rs`, where main deleted
`queue_popup_lists_prompts_and_arms_the_clear_confirm` (the queue editor's
paint moved to `v2::queue`) and this branch added two tests directly above it.
Resolved by keeping both sides' intent: main's deletion stands, this branch's
tests stay. **No test of this branch's was deleted by the merge.**

### One CI failure, and what caused it

The first run on this branch went red on `gate steps no other workflow runs` →
`check-prose`, and the cause is worth stating because it is the shape AGENTS.md
warns about rather than a slip: `#4763` landed while this branch was open, split
`scripts/prose-baseline.txt` from one total per file into **per-category**
counts, and tightened `crates/stella-tui/src/render/entry/tool.rs` to three
`filler-adverb`s. This branch's `result_body` doc added a fourth. Both sides
were correct alone; only the composition was wrong, and the local run I did
before merging could not see it because it predated `#4763`.

Fixed by deleting the word (`6254690d3`), never by touching the baseline —
"deliberately not rendered here" says exactly as much without the adverb, and
the sentences under it are the reason the guard asks for. `check-prose` and the
new `check-line-citations` both pass on the merged tree.

### Witnesses, run both ways

Each of the three behaviours was ablated on the merged tree — one line
each, the change removed and nothing else — and the named test re-run. All
three go red; restoring makes all three green (the run above).

| Ablation | Test | Ablated |
|---|---|---|
| `deck_render.rs`: the page branch stops calling `frame.set_cursor_position` | `render_deck_positions_the_hardware_cursor_on_the_agents_page_composer` | **FAILED** |
| `agents_page.rs`: `reply_rows` returns empty | `the_agents_page_shows_a_page_submitted_commands_reply` | **FAILED** |
| `render/entry.rs`: `v2_rows` stops resolving `read_size` for the head | `a_reads_line_count_is_stated_once_in_its_block` | **FAILED** |

```
ablated run exit=101
test render::tests::result_row::a_reads_line_count_is_stated_once_in_its_block ... FAILED
test deck_render::tests::render_deck_positions_the_hardware_cursor_on_the_agents_page_composer ... FAILED
test deck_render::tests::the_agents_page_shows_a_page_submitted_commands_reply ... FAILED
test result: FAILED. 0 passed; 3 failed; 0 ignored; 0 measured; 1120 filtered out
```

The third is the one worth reading twice. #4640 changes no behaviour, so it
has no fail→pass witness of its own — what the ablation shows instead is that
its pin is not vacuous: take the count off the head and the test fails rather
than shrugging, which is exactly the failure the recorded decision depends on
not happening silently.

### Golden frames

`page_agents` **re-blesses to the same bytes**, and so does every other frame in
the suite: with no reply pending and a blank composer the foot is the three rows
it always was, so the resting page costs nothing for either new feature. That is
a result worth having, not a step skipped — but it also means the existing
golden shows neither of the things this PR added.

So there is a second one, `page_agents_reply`
(`crates/stella-tui/tests/snapshots/deck/page_agents_reply.txt`), of the page
mid-conversation:

```
 REPLY
  worker · zai/glm-5.2 · effort high

 /model sent
 ❯ rewrite the parser
   ⏎ start a new agent · / commands · ↑↓ select · ⏎ open · n new session · ⌃x⌃x kill · r refresh · esc back
```

Blessed and read. `git status` after `BLESS=1` listed exactly one written file —
the new one — so the other twenty-one goldens are byte-identical, and the plain
run then passes 22/22 against the committed set.

## Not done

- **#4641** (`docs/tools` `output_schema` omits `data`) — **not in this PR.**
  The fix is written, reviewed and committed on `fix/c16-tooldocs-4641`
  (`efe3d3381`), and it is deliberately not merged here because it is
  incomplete: `output_schema` is a *generated* artifact, so the change is only
  shippable together with a regenerated `docs/tools/`, and
  `make tool-docs-update` requires building `stella-cli`'s test binary. This
  machine could not: it is 23.2 GB into 24.5 GB of swap with ~70 MB of free
  pages, and the `rustc` doing that codegen accumulated 28 CPU-seconds across
  19 wall minutes before it was stopped. Landing the generator change without
  the regenerated pages would turn the `tool-docs` gate step red, which is not
  a trade worth making for a green-looking PR. **To finish: check out that
  branch on a machine with memory, run `make tool-docs-update`, read the diff
  (every page changes — the envelope is per-tool identical), commit it, and
  merge.** Details in the commit message there.

- **#4580** (arenabench transcript coalescing) — **skipped, wrong repository.**
  Its subject `arenabench/arenabench/transcript.py` is not in
  `macanderson/stella` at any path: `git ls-tree -r origin/main` finds no
  `arenabench/` tree, and `gh search code "open_entries" --owner macanderson`
  places the file in `macanderson/arenabench` (arenabench was ejected from this
  workspace — see `.gitignore`'s `#2380` note). The issue needs transferring to
  that repository; nothing in this PR could have fixed it.

Closes #4640
Closes #4626
Refs #4641
Refs #4580


macanderson:
Answering the one ❌ in the assessment table — **"Re-bless the `page_agents` golden snapshot and review the resulting diff"** (#4626).

The row is right that no re-blessing had happened, and it is now addressed rather than argued away, in `e8f360ade`.

What `BLESS=1 cargo test -p stella-tui --test deck_render_snapshots` produced is the honest reason the first pass showed nothing: **`page_agents` re-blesses to the same bytes.** `git status` after the run listed exactly one written file, and it was the new one — the other twenty-one goldens are byte-identical. That is not the change failing to reach the page; it is the page at rest costing nothing for either feature. The existing fixture opens the page with a blank composer and nothing asked, so the reply pane contributes zero rows and the caret is not a glyph in the first place (goldens strip styling and the hardware cursor is not a cell). A frame that cannot show either feature cannot have a diff about them.

So the issue's intent is served by a second golden instead, `page_agents_reply`, of the page mid-conversation:

```
 REPLY
  worker · zai/glm-5.2 · effort high

 /model sent
 ❯ rewrite the parser
   ⏎ start a new agent · / commands · ↑↓ select · ⏎ open · n new session · ⌃x⌃x kill · r refresh · esc back
```

Read before committing, per this suite's own module doc. Both frames now pin something worth pinning: `page_agents` that the resting page did not move, `page_agents_reply` what the new pane and the typed composer actually look like. The plain (non-BLESS) run then passes 22/22 against the committed set.

Issue #4772, opened 2026-08-25
get_environment cannot say whether an interpreter has numpy, so a plan built around it still costs a step to disprove


macanderson:
## Problem

`get_environment` probes which *commands* resolve on `PATH`
(`PROBED_COMMANDS` in `crates/stella-tools/src/environment.rs`: `python3`, `pip`,
`pip3`, `apt-get`, `cc`, `make`, `xxd`). It does not probe whether an interpreter can
`import` anything, and its own doc comment says so deliberately:

> It answers "is this command on `PATH`" and nothing further. A library's presence
> (`numpy`) is a question about an interpreter's own search path, and it needs that
> interpreter spawned to answer honestly — which is a different probe from this one
> and does not belong behind the same line.

That is the right call for `PROBED_COMMANDS`, and it leaves the original defect
half-covered. The bench evidence behind #2670 is a `path-tracing` transcript that
discovered `numpy` was missing at second ~1676 — after building a plan around it —
and then discovered `pip: command not found`. The command probe now catches the
second discovery at turn 0. The first still costs a full step to find.

## What to build

A second probe that answers "can this interpreter import X", spawning the interpreter
rather than stat-ing `PATH`. Design questions the implementer owns:

- **Where it lives.** It spawns a process, so it is not `command_on_path`'s neighbour;
  it may want to be a separate field on `EnvironmentIdentity` collected lazily, or a
  separate line `get_environment` renders only when `python3` is present.
- **Cost.** `EnvironmentIdentity::collect` runs once per session and is on the session
  open path. One `python3 -c "import numpy"` is tens of milliseconds; a list of ten is
  not. Probing a fixed short list in one interpreter invocation
  (`python3 -c "import importlib.util as u; ..."`) keeps it to a single spawn.
- **Which libraries.** Evidence-chosen, the way `PROBED_COMMANDS` is — the doc comment
  there is explicit that it is "short and evidence-chosen, not a survey". `numpy` is
  the one the bench transcript actually paid for. Do not add a list because it looks
  complete.
- **Absence is the useful half**, same as the command probe: `EnvironmentIdentity::render`
  names what is missing, not only what is there.

## Files

- `crates/stella-tools/src/environment.rs` — `PROBED_COMMANDS`, `available_commands`,
  `EnvironmentIdentity::{collect, render}`, and the `GetEnvironment` tool doc that
  tells the model when the call is worth its cost.
[...]

Issue #4753, opened 2026-08-25
Gate compaction retention on the reclaim ratio, not the budget fraction — the census says the current key selects the wrong firings


macanderson:
## Problem

Compaction pass 0 (tool-result retention) is gated on how far into its budget the
conversation is: it fires only once the conversation occupies at least
`budget_tokens / RETENTION_TRIGGER_BUDGET_DIVISOR`, and that divisor is `2`
(`crates/stella-core/src/compaction.rs`).

#4452 measured that gate over recorded runs and it selects the wrong firings. The
census (`scripts/retention-census.py`, 370 pure pass-0 firings across 96 session
journals in `~/.stella/sessions/`, all from binaries with no gate so the
counterfactual is askable) found:

- A firing's net in uncached-equivalent input tokens is
  `after − r·before − r·reclaimed·(N−1)`, for `N` further worker calls in the turn
  and a cache read priced at `r` of an uncached token. Negative is a saving.
- At `r = 0.10` the gate at half costs **+5.63 M** tokens against **+2.92 M** for no
  gate at all. At `r = 0.25`: −3.04 M against −24.6 M. At `r = 0.50`: −17.5 M against
  −70.6 M. At every price the gate at half is the worse of the two.
- It loses because budget fraction is anti-correlated with the trade. Profitable
  firings sit at a median 0.266 of budget, unprofitable ones at 0.425 — a turn is
  furthest from its ceiling early, and early is when the most calls remain to
  amortize the prefix invalidation (median 90 remaining worker calls where a firing
  profits, 26 where it does not). At `r = 0.10` the gate admits 112 firings of which
  15 profit: precision 0.13 against a 0.32 base rate.

The divisor was left at 2 by #4452 because the census disqualifies the key and names
no replacement *value* — every divisor is a worse classifier than no gate, so 3 or 4
would trade one unmeasured number for another. This issue is the replacement *key*.

## What to build

Gate the retention pass on the **reclaim ratio** `reclaimed / after` instead of, or in
addition to, the budget fraction. The ratio is the other term of the break-even
expression above, and unlike the remaining call count it is known at firing time.

Measured on the same 370 firings at `r = 0.10`:

| gate | admits | profitable | precision | recall | net (uncached-equivalent tokens) |
|---|---|---|---|---|---|
| none | 370 | 120 | 0.32 | 1.00 | +2 923 748 |
[...]

Issue #2808, opened 2026-08-11
Track whether a run shipped something, separately from how it ended — a cancelled run that merged a pull request looks like it delivered nothing


macanderson:
## Problem

This is the second half of #2570's "Done when", split out because it can't be done inside `crates/stella-store/` alone: the store has nowhere to record delivery, and the crate that knows about it (`stella-cli`) has nowhere to put it.

Execution 99 is recorded as `outcome = 'cancelled'`. What actually happened: the worker had already pushed and merged pull request #2554 before the cancel landed. The record says a cancelled run that (after #2570's cost fix) cost $5.29; the truth is a $5.29 run that shipped a merged PR. `outcome` is a single final label being asked to describe two separate facts:

- **how the run ended** (completed / aborted / cancelled / failed / error), and
- **whether it delivered anything** (a commit, a push, an opened or merged PR).

Those are independent. A completed run can deliver nothing; a cancelled run can deliver everything. Collapsing them means every analysis that filters on `outcome = 'completed'` — `usage_stats`'s `resolved` count (`crates/stella-store/src/export.rs`), the scoreboard, cost-per-resolved-task, and the responsibility-ablation controls #2374 needs — silently scores delivered work as zero.

## What exists today

- `executions.outcome` — `crates/stella-store/src/ddl.rs` (`EXECUTIONS_DDL`).
- `pull_requests` — already in the schema, but keyed by session (`upsert_pull_request`), not by execution, so it can't answer "did *this* run deliver?".
- `files_touched` / `tool_calls` per execution — evidence of local edits, but not of a commit or a push.

So the fact is partly derivable and recorded nowhere.

## Suggested fix

1. Add an execution-scoped delivery record to `stella-store` — a nullable column (`delivered_at` / a small `delivery` kind field) or an `execution_delivery` table keyed on `execution_id`, with a migration in `crates/stella-store/src/migrations.rs` and the schema change in `src/ddl.rs`. Keep it additive: `executions` rows written before the migration must read as "unknown", not "did not deliver".
2. Wire up the writer in `stella-cli` where commits/pushes/PRs are already observed (the `pr_nudge` check in `command_deck.rs`, the fleet commit ledger, the git tool call sites).
3. Decide explicitly whether `usage_stats.resolved` should count a delivered cancelled run. Changing it moves every published cost-per-resolved number, so this wants to be a stated decision with a test, not a side effect.

## Constraints

[...]

PR #4789, opened 2026-08-25
fix(stella-core,stella-cli,stella-tools,stella-embed): unpark four decision-blocked issues with offline measurements

### LINKED ISSUE #2670: The worker copies a misleading example instead of trusting its own correct answer, and a 120-second command limit wastes time re-parsing large files on every step
## Problem (two small, related worker-behavior defects, one issue)

**P4 — example over-fit + self-confirming verification.** extract-elf bare: Stella derived a +0x400000 rebase from the task's example output, then itself concluded "the example values are generic illustrations, not from this file" — and KEPT the offset, then "verified" 698/698 against its own model. CC emitted vaddr as-is and passed; a one-constant diff after 100% of the work. build-pov-ray bare: 2 of 3 tests passed (SSIM 0.873); failed only on `file_id.diz not found` — its own zip's members were uppercase (`FILE_ID.DIZ`, proven by its own S150). `unzip -LL` or a filename-case check flips the cell. 2/16 cells, both one-line diffs.

**P6 — the 120s bash wall + no persistent state.** path-tracing: three identical-shape `timeout 110/115 python3 <<EOF` kills in a row; then discovers numpy missing at second ~1676; then `pip: command not found`; never tries `apt-get install python3-numpy` (apt worked in the same dataset). Every step re-parses the 48MB PPM from scratch (verbatim parse in S36–S54, ~2s each); CC cached to `/tmp/img.bin` once. Same re-parse pattern in raman-fitting (8 steps). Loop detector inconsistency: 38 byte-identical `read_output`/`send_stdin` polls unflagged (qemu) while 3 identical calls elsewhere drew a warning. 4/16 cells' dominant time sink.

## Fix directions

P4: (a) prompt/skill rule — "if your output disagrees with a task-supplied example, treat it as your defect until proven synthetic; never keep a transform derived from an example you then invalidated"; (b) done-check on deliverable shape (filename case, path, extension); (c) verification must compare against task ground truth, never a re-derivation of the agent's own model.
P6: (a) escalating command timeout the model can request, and exit-124 responses that direct a strategy change instead of a resubmit; (b) a persistent scratch/session store (or a warm `python` tool) so parsed data survives steps; (c) turn-0 environment capability probe (interpreters, pip/apt, numpy, file/xxd/ps, network); (d) loop-detector thresholds for read_output/send_stdin polling.

Evidence: `turnloop-rerun-bare` extract-elf/build-pov-ray; `turnloop-cert-panel1` + `turnloop-rerun-pipe2` path-tracing/raman transcripts; `ccglm-bar-B` contrasts.

### LINKED ISSUE #4452: Check whether the compaction budget threshold (50%) is right — it was picked by reasoning, not by measurement
## Problem

#4381 gave compaction pass 0 a budget-pressure gate: it fires only once the conversation
occupies at least `budget_tokens / RETENTION_TRIGGER_BUDGET_DIVISOR`, and that divisor is
`2` (`crates/stella-core/src/compaction.rs`).

**Half was chosen by argument, not by measurement.** The argument is in the constant's doc
comment: the budget passes trigger at the full budget and reclaim to seven-eighths, so a
gate near them would make retention a duplicate of pass 3, while a gate at half leaves the
50-100% band — where a long turn's standing context accumulates — for retention to shape.
That reasoning bounds the answer; it does not pick it. A third, or two-thirds, would satisfy
the same argument.

The trade it governs is measurable and this repository already has both sides of it recorded:

- **Cost of firing**: the prompt-cache miss on the next call. `step_usage` carries
  `cached_input_tokens` and `input_tokens`; a `compaction` event immediately before a call
  whose cached fraction collapses is the shape #4381 measured (execution 242 of
  `ses-1787465453163-60967`: 137 K re-billed across five firings).
- **Cost of not firing**: the input tokens re-sent verbatim on every subsequent call of the
  turn, which is what #1285 measured (4x more input per step than a comparator holding its
  standing context flat).

Both are per-turn sums over `store.db`, so the fraction that minimises total billed input
tokens can be computed offline from sessions already on disk, without a new bench run.

## Files

- `crates/stella-core/src/compaction.rs` — `RETENTION_TRIGGER_BUDGET_DIVISOR` and its doc
  comment (which states the argument this issue is asking to replace with a number);
  `age_stale_tool_results`'s early return is the gate.
- `crates/stella-core/src/compaction/tests.rs` — `retention_does_not_fire_at_a_quarter_of_the_budget`
  and `retention_fires_once_the_conversation_passes_half_its_budget` are the pair that pins
  the current value; `retention_pressure_budget` derives a test budget from the transcript.
- `.stella/private/store.db` and `~/.stella/usage.db` — the `executions`, `events` and
  `step_usage` rows the census reads.

## Verify

1. Census every turn in `store.db` that has at least one `compaction` event with `aged > 0`.
   For each, join `compaction` -> next `step_usage` for the re-billed miss, and sum
   `input_tokens` across the remaining calls of the turn for what the aging saved.
2. Replay the same turns with the gate at several divisors (2, 3, 4 and no gate) and compare
   total billed input tokens per turn. The pass is pure over the transcript, so this is a
   replay rather than a re-run — no model calls.
3. Report the unflattering number if the current value is not the best one (CLAUDE.md,
   `measure honestly, especially when it costs us`).

## Done when

- The divisor is either confirmed at 2 by the census, or changed to what the census says,
  with the number and the sample size in the doc comment where the argument currently sits.
- The two witness tests move with it if it moves (they read the divisor's effect, not the
  constant, so they need their budgets re-derived rather than rewritten).
- If the census shows the right answer depends on something the gate cannot see — turn
  length, the size distribution of tool outputs — say so and name what a better gate would
  key on, rather than fitting a constant to an average.

macanderson:
Four issues that were each parked on a decision. Two are decided and closed; two are sharpened with evidence and left open, one because its decision moves published numbers and one because `main` landed the better measurement while this branch was in flight. Every measurement here is offline over data already on disk — no bench rig was run.

### #4452 — the compaction retention gate: priced, and the key found wrong

`RETENTION_TRIGGER_BUDGET_DIVISOR` gates compaction pass 0 on how far into its budget a conversation is. The doc already carried a *firing* census; what decides the trade is the *billed-token* half, which nobody had.

`scripts/retention-census.py` (new) reads `~/.stella/sessions/*/journal.jsonl` and prices each pure pass-0 firing as `after − r·before − r·reclaimed·(N−1)`, for `N` further worker calls and a cache read at `r` of an uncached token. It reads the journals rather than `store.db` because the maintainer's project store is corrupt — `pragma quick_check` reports `btreeInitPage` error 11 on tree 45.

370 firings across 96 journals, all from binaries with no gate, so the counterfactual is askable:

| | no gate | divisor 2 |
|---|---|---|
| r = 0.10 | +2.92 M | **+5.63 M** |
| r = 0.25 | −24.6 M | −3.04 M |
| r = 0.50 | −70.6 M | −17.5 M |

At every price the gate at half is the worse of the two. It loses because budget fraction is **anti-correlated** with the trade: profitable firings sit at a median 0.266 of budget against 0.425 for unprofitable ones, because a turn is furthest from its ceiling early and early is when the most calls remain to amortize the prefix invalidation (median 90 remaining worker calls where a firing profits, 26 where it does not). At r = 0.10 the gate admits 112 firings of which 15 profit — precision 0.13 against a 0.32 base rate.

**The divisor stays at 2.** The census disqualifies the key and names no replacement *value*: every divisor classifies worse than no gate, so 3 or 4 would trade one unmeasured number for another. #4452's "done when" accepts exactly this — say so, and name what a better gate would key on. That is the reclaim ratio `reclaimed / after`, the other term of the break-even expression and the one known at firing time; a gate at `ratio >= 0.15` turns the same corpus from +5.63 M into −5.61 M. Filed with the full table as #4753.

The measured miss is real, not modelled: the cached fraction is a median 0.487 on the call before a firing and 0.196 on the call after.

### #2670 — example over-fit, and a timeout that invites a resubmit

Most of this issue had already shipped on `main` and the issue never caught up. Verified against the tree today: the escalating timeout is a model-settable `timeout_secs` (`bash.rs`, default 120, max 600); the persistent scratch plane is `save_state`/`get_state`/`list_state`/`delete_state` over `$STELLA_SCRATCH`; the turn-0 capability probe is `PROBED_COMMANDS` in `environment.rs` (`python3`, `pip`, `pip3`, `apt-get`, `cc`, `make`, `xxd`, reporting absence as well as presence); and P6(d) is moot because `read_output` and `send_stdin` are in `RETIRED_TOOL_NAMES`. Two things were genuinely absent.

**P4 — a new `example_discipline!` shared contract** (`crates/stella-cli/src/agent/prompt.rs`), embedded verbatim in both static prompts and registered in `SHARED_CONTRACTS`. It covers all three of the issue's P4 directions in one literal: a task-supplied example outranks the worker's derivation; a transform derived from an example later called illustrative goes with it; the deliverable's shape (filename case, path, extension, archive member names) is checked against the task's own words; and verification compares against what the task states rather than a re-derivation of the worker's own model. CLAUDE.md's built-in-surface rule forbids a new built-in for the shape check, so it lands as prompt guidance.

**P6(a) tail — the timeout error now directs a strategy change** (`crates/stella-tools/src/bash.rs`). It used to say `command timed out after Ns` and nothing else, which invites the one move that cannot work. It now rules out the resubmit by name, gives the ceiling as a number the model can act on, and names where an expensive intermediate survives to the next step.

### #3096 — the admission floor's tail, measured with no key (and then overtaken)

`scripts/embed-floor-census.py` (new) reads an existing `codegraph.db` and scores stored chunk vectors against each other. That probes the same space: `HttpEmbedder::embed` sends `{"model", "input"}` with **no `input_type` field**, so Voyage's query/document asymmetry is never engaged and a stored chunk vector is what a query embedding of that text would be; the vectors are L2-normalized, so cosine is a plain dot product (measured norms 0.9999995-1.0000007 over 200 vectors).

Over this repository's 28 316 `voyage-code-3` chunk vectors, 2 000 random cross-file pairs -- the irrelevant tail the floor exists to cut -- score min 0.4110, p1 0.4566, median 0.5679, p95 0.6671. **None falls below 0.25.**

**While this branch was in flight, `main` landed `MEASURED_FLOORS` and measured the half I said needed a key** -- the frontier. Somebody ran the labelled harness for real and found the relevant and irrelevant distributions *overlap with the wrong sign*: tightest separation -0.0439, the labelled answer for one query ranking 39th under 38 irrelevant chunks. `voyage-code-3` is recorded as `None` and declares `SimilarityPosture::Surface`.

**So I dropped my `http.rs` change entirely and took `main`'s version verbatim** -- `crates/stella-embed/` on this branch is now byte-identical to `main`. My paragraph argued about `DEFAULT_ADMISSION_FLOOR` in terms of a model that is no longer governed by it, and my pinning test's rationale was written before the harness had run. Keeping either would have been an argument with a better measurement.

What survives is the script, and it earns its place for two reasons. It **corroborates the labelled run from the other side**: the tail is dense between 0.41 and 0.72, which is the structural reason no floor fits underneath a real answer -- two measurements by different mechanisms agreeing. And it is the **no-key** probe for the eight models in `KNOWN_DIMS` that remain unmeasured; run it before spending a key on one, since a tail that never approaches the shipped floor says the floor is inert there too. Its docstring says all of this, including that it cannot see the frontier.

The issue stays open, and the reason has shifted: `DEFAULT_RELEVANCE_GAP_RATIO` and `DEFAULT_MIN_BOUNDARY_GAP` in `rank.rs` are still unmeasured, and they matter *more* now -- with `voyage-code-3` on `Surface`, the relative cut in `relevant_prefix` is the only thing deciding how many results the model sees. Recorded on the issue, including a correction of what I got wrong before the measurement landed.

## Evidence

Targeted commands only; CI is the real gate.

| command | result |
|---|---|
| `cargo test -p stella-core --lib compaction` | ok, 67 passed |
| `cargo test -p stella-tools --lib bash::` | ok, 41 passed |
| `cargo test -p stella-cli --bin stella prompt` | ok, 100 passed |
| `cargo clippy -p stella-core -p stella-tools -p stella-cli --all-targets -- -D warnings` | exit 0 |
| `make guards-fast` | exit 0 — `check-prose: OK, none added`; `check-measured-constants: OK, 5 constants each named by a test` |
| `cargo fmt --all --check` | exit 0 |
| `python3 scripts/retention-census.py` | exit 0, output quoted above |
| `python3 scripts/embed-floor-census.py` | exit 0, output quoted above |
| `RUSTDOCFLAGS="-D warnings" cargo doc -p stella-core -p stella-tools -p stella-cli --no-deps` | exit 0 |
| ... and again with `--document-private-items` | exit 0 |

Every row above was re-run **after** the second `origin/main` merge, following `cargo clean -p` on all four crates — cargo replays a cached green across a merge otherwise. The `stella-embed` runs from before that merge are dropped from this table rather than restated, because the change they covered is no longer in the diff.

Plain form run first in every case, since it is the only one that catches a public item's doc linking to a private one.

`make prose` earned its place here: the first `guards-fast` run passed because `scripts/retention-census.py` was still untracked, and the run after the merge failed on an enumerative announcement I had written into its own module doc — the exact construction the guard is named for. Fixed by deleting the clause rather than by a baseline entry (`3bb2ca1`).

Merged `origin/main` twice. The second merge conflicted on `crates/stella-embed/src/http.rs` and is what removed this branch's `stella-embed` change; see the #3096 section.

**Witness, #2670 (bash):** `a_timeout_directs_a_strategy_change_rather_than_a_resubmit`. Run against `origin/main`'s production timeout arm spliced under the new test module — fails on `Re-sending it unchanged will time out again`. Passes on this branch.

**Witness, #2670 (prompt):** `both_prompts_rank_a_task_supplied_example_above_the_workers_own_derivation`, pinning the shared literal verbatim in both prompts plus each of its three clauses separately so a hollowed-out literal cannot pass it. Fails with the two `concat!` embeds reverted; passes with them.

**No witness, #4452:** no behaviour changes. The deliverable is the measurement. The constant is pinned as before by `the_retention_pressure_gate_stays_at_the_divisor_the_census_produced`, whose message now carries the numbers that survived.

**No witness, #3096:** the Rust change was dropped in the merge, so what remains is one script. Evidence is its output, reproducible by running it against any populated `codegraph.db`.

No test was deleted. `timeout_is_classified_timeout_with_unchanged_message_bytes` was rewritten rather than removed: it pinned the exact old message string, so it now asserts the class plus the message's stable prefix, keeping its #3167 purpose.

## Not done

**#2808 — execution-scoped delivery record: left open, sharpened.** Its blocking question ("should `usage_stats.resolved` count a delivered cancelled run?") is user-visible policy — `resolved` is a view over rows that already exist, so redefining it silently rewrites every historical `cost_per_resolved_usd`, including the ones this project publishes against Claude Code. That is a maintainer's call, so I recorded the decision with its reasoning and a recommendation (leave `resolved` alone; add `delivered` beside it) rather than taking it.

Two further reasons it did not land here: `crates/stella-store/src/lib.rs` is at 1712 lines against a baseline entry of exactly 1712, so the ordinary landing (`pub mod delivery;`) cannot pass the file-size gate — the way through is to put the `impl Store` methods in `scoreboard.rs`, which is already the delivery scoreboard by subject. And it bumps `SCHEMA_VERSION`, a shared cell every schema PR writes, which wants its own PR from a fresh `main` rather than a slot in a four-issue branch. The comment carries the full verified landing checklist with current line numbers.

## Filed

- #4753 — gate compaction retention on the reclaim ratio, with the measured precision/recall/net table for four thresholds.
- #4772 — `get_environment` cannot say whether an interpreter has `numpy`; the command probe catches `pip` at turn 0 but a library still costs a step to disprove.
- #4797 — the replay #4452 asked for and this PR did not do. The census filters *recorded* firings, which is first-order; a replay re-drives `age_stale_tool_results` over the transcripts. It is blocked on choosing a corpus: the journals lack the system prompt and carry assistant turns only as event streams, and the project `store.db` is corrupt (`btreeInitPage` error 11 on tree 45; per-execution reads still work for 314 of 315 executions). Answered on the PR as a Sourcery ❌ row.

Closes #4452
Closes #2670

## Summary by Sourcery

Use offline measurements to resolve or sharpen decision-blocked issues while improving agent guidance for task examples and command timeouts.

New Features:
- Add offline census scripts for evaluating compaction retention economics and embedding admission-floor behavior from data already on disk.

Bug Fixes:
- Improve timeout guidance so timed-out commands discourage unchanged retries and suggest actionable recovery strategies.

Enhancements:
- Add a shared prompt contract that prioritizes task-provided examples and validates deliverable details against the task specification.
- Document measured compaction findings while retaining the existing budget divisor pending a better reclaim-ratio gate.
- Register the new prompt contract in both static prompts and enforce prompt parity with targeted tests.

Tests:
- Add coverage for the shared example-discipline prompt contract and strategy-oriented timeout messaging.

Chores:
- Record follow-up issues for reclaim-ratio compaction gating, environment capability detection, embedding-floor calibration, replay analysis, and delivery metrics policy.


macanderson:
## Answering the ❌ row

> **#4452** — *Measure the compaction retention tradeoff across the current threshold, alternative divisors, and no gate using recorded per-turn usage, so the best threshold can be confirmed or changed based on total billed input tokens.* ❌ — "does not perform the requested replay of the same turns with divisors 2, 3, 4, and no gate… it explicitly acknowledges that the census only bounds the effect."

**Sourcery is right about the mechanism, and I am not fixing it in this PR.** Recording the answer here, and the residue is now tracked.

What the PR does is a counterfactual *pricing* census: it takes the pass-0 firings that actually happened and re-totals under each candidate gate by **filtering the recorded firings**. What #4452's Verify step 2 asks for is a **replay** — re-driving `age_stale_tool_results` over the transcripts at each divisor. Those differ exactly where the census says they do: filtering is first-order, because suppressing a firing changes the bytes the next firing sees, and a conversation left un-aged can climb into the lossier budget passes, which the census does not model at all.

Two reasons it is not in this PR rather than an oversight:

1. **The issue's own "Done when" accepts the branch I took.** Its third bullet: *"If the census shows the right answer depends on something the gate cannot see — turn length, the size distribution of tool outputs — say so and name what a better gate would key on, rather than fitting a constant to an average."* That is what the census found (profitable firings sit at a median 0.266 of budget against 0.425 for unprofitable ones; the discriminator is the remaining call count, median 90 against 26) and what the PR does — the reclaim-ratio gate, with its measured precision/recall/net table, is filed as #4753. Sourcery marked both of the other #4452 rows ✅, which are those two bullets.

   The bounding is also not marginal here. At r = 0.10 the shipped gate totals **+5.63 M** against **+2.92 M** for no gate — it is worse than admitting everything, on a corpus of 370 firings, and it is worse at r = 0.25 and r = 0.50 too. A first-order model can be wrong about the *ranking of 3 versus 4*; it is not plausibly wrong about "this key selects worse than no key at every price tested". That is why the PR changes no value: the census is strong enough to disqualify the key and deliberately not strong enough to name a replacement number.

2. **The replay is blocked on a corpus, and the blocker is worth writing down.** `age_stale_tool_results` is pure and drives deterministically with no model calls, so the pass is not the problem — the input is. The session journals carry full `tool_result` payloads and the `compaction` rewrites, but not the system prompt, and assistant turns survive only as `text`/`text_delta`/`reasoning` streams that would need re-assembling into `CompletionMessage`s. The fuller record is the project `store.db`, and it is **corrupt**: `pragma quick_check` reports `btreeInitPage` error 11 on tree 45, and any full scan of `events` aborts with `database disk image is malformed (11)` (per-execution reads still work for 314 of 315 executions). Choosing the corpus is the first real decision the replay owns, and it is bigger than a row in this PR.

**Filed as #4797**, written as a handoff: the obstacle above, the three corpus options, the pricing model to reuse, and a "done when" that requires the replay to state what its reconstruction lost. #4753's verify list also asks the replay to say whether it agrees, so neither issue can close claiming the other did it.

No other row on the assessment table is ❌.

PR #4792, opened 2026-08-25
fix(stella-cli,stella-core,stella-tui): a plugin's skill is listed as the plugin's, and disabling it works

### LINKED ISSUE #4734: A plugin's skill is missing from the skills listing and reports itself as the workspace's own
## Problem

A plugin's contributed skill is loaded, selectable, and — since #3567's
provenance half — carries `Skill::contributed_by` naming the package that
shipped it. It still does not appear in the skills listing at all, and the
origin it reports is wrong.

Two separate defects, both in `crates/stella-cli/src/skill_manager.rs`:

1. **`enumerate` cannot see it.** It walks exactly two directories —
   `scope_root(SkillScope::Project | User)`, i.e. `<workspace>/.stella/skills`
   and `~/.stella/skills` — so a skill derived from `<plugin_dir>/skills` is
   absent from the `SkillRow` set the interactive SKILLS tab renders. A user
   looking at the tab to answer "what is steering me?" is shown an incomplete
   list, and `stella plugin list`'s `ships skills: …` line is the only place
   the answer exists.
2. **Its origin is mislabelled `workspace`.**
   `stella_core::skills::default_origin_for` derives `SkillOrigin` from which
   *configured* directory a path is under; a package's directory is neither, so
   it falls through to `SkillOrigin::Workspace` — a third party's skill
   reporting as the user's own. `contributed_by` now makes the real answer
   available beside it, and nothing reads it.

## What a fix has to decide

Contributed skills are not "installed under a scope" and the tab's other
columns assume they are:

- `removable: true` is wrong — uninstall unlinks an entry under a scope root,
  and a contributed skill has none. Retraction is `stella plugin remove`.
- `version` / `latest` read `<slug>/versions/vN/`, which a package need not
  have.
- The enable/disable state file (`.stella-skills.json`) lives per scope root,
  and `skill_manager::retain_enabled` already applies it to contributed skills
  by name — so a decision is owed about which state file governs them.
- `SkillScope` is on the wire (`crates/stella-tui/src/envelope/skills.rs`), so
  a third variant is a wire change; a `contributed_by: Option<String>` field on
  `SkillRow` beside the existing `origin: String` may be the cheaper answer.

## Definition of done

- A contributed skill appears in `skill_manager::enumerate`'s rows, marked as
  the third party's and not removable through the tab.
- Its origin no longer reads `workspace`.
- A witness in `crates/stella-cli/src/plugin_cmd/tests/contributions.rs`, beside
  `a_contributed_record_and_skill_each_name_their_plugin_as_a_field`, which is
  where the loader-side property is already pinned.

Split out of #3567, whose provenance half landed the field this needs.
Refs #3567, #3380.

macanderson:
Closes #4734.

## The two defects

A skill an installed plugin ships was loaded, selectable, and **absent from the SKILLS tab** — `skill_manager::enumerate` walked exactly the two scope roots the user owns, so `stella plugin list`'s `ships skills:` line was the only place the answer existed. A user asking the tab "what is steering me?" was shown an incomplete list.

It also **reported `workspace`**: `default_origin_for` derives the origin from which *configured* directory a path sits under, and a package's `skills/` is neither, so it fell through to `Workspace` — a third party's skill claiming to be the user's own.

## What changed

**`enumerate` appends a contributed row per plugin skill.** The column decisions the issue asked for, argued in the doc comment on `append_contributed`:

| column | value | why |
| --- | --- | --- |
| `scope` | `Project` | `SkillScope` is on the wire; rather than widen it the row borrows the scope whose **state file** governs it — the workspace's, matching the roster (`session_roster` reads workspace settings) that decided this plugin loads here |
| `removable` | `false` | uninstall unlinks an entry under a scope root and this has none; retraction is `stella plugin remove` |
| `version`/`latest` | `1` | they read `<slug>/versions/vN/`, which a package need not ship, and `save_edit`/`set_pin` look the name up under a scope root and refuse it — the tab does not edit a third party's package |
| `origin` | `"plugin"` | with the package in `contributed_by`; `SkillRow::provenance()` renders `plugin:vera` |

A name already claimed by one of the user's own rows is skipped, mirroring `append_plugin_skills` on the load side — a row for a skill that isn't steering anything answers the tab's question wrongly in the other direction.

**`SkillRow` gains `contributed_by`** beside its `origin`, so a renderer keys on the label without parsing a name out of it. The v2 tab shows the package where an installed skill shows its scope; the ctrl+o preview subtitle does the same; `deck_render`'s ACTIVE SKILLS list shows `[plugin:vera]`.

**The origin stamp.** `memory::skill_files::append_plugin_skills` now writes `SkillOrigin::Installed` over whatever the file said — including an `origin: auto` marker, which buys a selection tie-break (`skills.rs`'s `select_skills`) that a package must not be able to claim. This is the half of defect 2 that lives on the recall side.

**`retain_enabled` gained the contributed branch.** The deck already answers ctrl+x on a non-removable row with *"press space to disable it"*, so space has to work. A plugin's skill is under neither scope root and fell through to "keep", so disabling one silently did nothing; it is now matched by its `contributed_by` stamp and governed by the workspace state file — the same file `enumerate` gives its row, so the key and the filter are one decision.

## Note on #3567 — the field landed on main mid-flight

The issue was written expecting `Skill::contributed_by` to exist. It did not when this branch started, so the branch added it; while the work was in progress **#3567's provenance half landed on main** and added the same field, along with `a_contributed_record_and_skill_each_name_their_plugin_as_a_field`. The merge keeps **main's** field, its doc comment, its parser comment and its test verbatim — this PR contributes none of that. What survives from this side is the listing half plus one line main did not have: the `origin = Installed` stamp. Main's `append_plugin_skills` stamps `contributed_by` and leaves the origin claiming `Workspace`, so defect 2 was still live on the recall side after #3567; that is the line, and `a_plugins_skill_is_listed_as_the_plugins_and_is_not_deletable_from_the_tab` asserts it.

## Evidence

Two witnesses in `crates/stella-cli/src/plugin_cmd/tests/contributions.rs`, beside main's:

- `a_plugins_skill_is_listed_as_the_plugins_and_is_not_deletable_from_the_tab` — anti-vacuity before install; then the row names `vera`, reads `plugin` (not `workspace`), renders `plugin:vera`, and is not removable; the loaded `Skill`'s origin is `Installed`; the row leaves with `plugin remove`.
- `disabling_a_plugins_skill_from_the_tab_stops_it_steering` — it steers, `set_enabled(Project, …, false)` stops it being injected, it stays listed marked off, re-enabling brings it back.

CI on this head is green — `fmt + clippy + test` (16m16s), `guard self-tests`, `file size ratchet`, `docs guards`, `cargo deny`, MSRV. Locally: `cargo test -p stella-cli --bins contributions::` → 11 passed, 0 failed (both new tests plus main's own, all green together on the merged tree).


macanderson:
Closing as superseded by #4798, which fixes the same issue (#4734) with a different, non-overlapping design.

Both PRs touch largely the same files (`skill_manager.rs`, `skill_files.rs`, `deck_ui.rs`, `envelope/skills.rs`, `v2/skills.rs`, `plugin_cmd/tests/contributions.rs`), so keeping both open would mean whichever merges second either hits real conflicts or silently duplicates/reverts the other's approach.

#4798's own description notes that this PR's premise — that `Skill::contributed_by` already existed on `main` — was false (it traces to work on an unpushed local branch), and lands a cleaner fix using a new `SkillOrigin::Contributed` variant instead of reusing `contributed_by` + `SkillOrigin::Installed`. Going with that one.

Issue #4375, opened 2026-08-23
Live-run demo recordings need a size option and multiple output resolutions, like the slide-deck videos already have


macanderson:
## Problem

`scripts/record-demo.sh` (the live-run timelapse path, the one `docs/demo/stella-demo.mp4` came from in #796) renders its cast through `agg` at agg's default font size and then transcodes the GIF with ffmpeg. There is no way to ask it for a 4K or 6K output, and no delivery ladder: the result is one mp4 at whatever pixel size the default font produced (the #796 file is 982x694).

The deck film has both since the `feat/deck-film-v2-6k` branch: `scripts/render-deck-film.py --size WxH` renders a 6K master and a ladder is scaled down from it. A live-run recording — the thing #3556 wants for the hero film, and the thing the 2026-08-22 re-record of the word-frequency scenario produced (`recordings/2026-08-22/stella-demo-2026-08-22.cast`, local, gitignored) — cannot be delivered at the same sizes.

## What exists

- `agg --font-size N` controls the raster: at 20px a 126x30 grid renders 1512x868 (12x28 px cells plus padding); ~84px gives a >6K raster (6451x3646 measured). agg writes only GIF, so a 6K cut is a multi-GB GIF before ffmpeg sees it.
- `agg --theme` accepts a custom `bg,fg,c0..c15` palette; the brand ground `#0A0A0C` / gold `#EFC53F` is not the default.
- `record-demo.sh` hardcodes `--theme dracula` and never passes `--font-size`; its ffmpeg step is `-pix_fmt yuv420p` with no scaling.
- A throwaway ladder script that took a cast to 6K/4K/1080p/720p/480p/360p (H.264 High + HEVC companions) ran fine on 2026-08-22 in a job tmp dir; it is not in the tree.

## Definition of done

- `record-demo.sh` gains `--size WxH` (or `--font-size`) and `--theme` defaults to the brand palette.
- One `make` target (or a flag) derives the standard ladder from the largest rung: 6144x3456, 3840x2160, 1920x1080, 1280x720, 854x480, 640x360 — H.264 High / yuv420p / faststart, levels that admit 60 fps at each size (4.2 for 1080p60, 3.2 for 720p60, 5.2 for 4K60, 6.1 for 6K60).
[...]

Issue #4338, opened 2026-08-23
The command palette can't group commands by relevance or domain yet


macanderson:
## Problem

Rendering `08-command-palette` opens with `relevant now · verify is running` (`/gates` first, with its live `◐ 2/5 green`), then domain groups (`plan`, `graph`, …), then `recent`. `crates/stella-tui/src/render.rs` (`render_slash_popup`) draws the composer's fuzzy ranking flat: `SlashCommand` (`crates/stella-tui/src/composer.rs`) has a name, a description and a kind, and no domain or relevance.

## Done means

- `SlashCommand` gains a `domain` and the deck a relevance rule keyed on session state (stage, gate board, plan state).
- The popup draws the three sections with their headings; a `recent` list persists per workspace.
- Golden: query `ga` during a verify turn shows `/gates` first (IMPLEMENTATION-PLAN P7 acceptance).

macanderson:
Audit verdict (#4455): **(a) — deck work; the producer is the deck itself**. `SlashCommand` (crates/stella-tui/src/composer.rs) is deck-owned data that lacks a `domain` column and a relevance rule; no wire event or engine measurement is involved.

macanderson:
SHIPPED — the command palette groups by relevance and by domain on `origin/main` c1dc8bba0, with two witness tests that pass.

**The producer.** `SlashCommand::in_domain(SlashDomain::…)` and `SlashDomain` — `crates/stella-tui/src/composer.rs:137` ("What the command is about — the palette's browse groups"). The menu carries `sections`, so a heading is a row of its own.

**Witness 1** — `crates/stella-tui/src/composer.rs:960-985`, tagged `**The witness (#4338).**`:

```
fn the_browse_list_leads_with_relevance_then_groups_by_domain()
  assert_eq!(&names[..2], &["/plan", "/budget"], "the running turn's commands lead");
  assert_eq!(menu.sections.first(), Some(&(0, "relevant now · a turn is running".to_string())));
```

**Witness 2** — `crates/stella-tui/src/render/tests/slash.rs:97-130`, tagged `**The witness (#4338).**`, asserts the *rendered buffer* carries `relevant now · a turn is running` and the domain headings, and that the popup is sized and windowed over the heading rows ("a list sized on matches alone clipped its tail behind its own captions").

**Re-run:**

```
$ cargo test -p stella-tui the_browse
test composer::tests::the_browse_list_leads_with_relevance_then_groups_by_domain ... ok
[...]

PR #4799, opened 2026-08-25
feat(stella-model,stella-tui,stella-cli): sampled-frame video degrade, the GRAPH query box, and demo output sizes

### LINKED ISSUE #3340: Let models that can see images understand video by sampling frames, instead of just describing that a video was attached
## Problem

Stella's multimodal input plane (`crates/stella-protocol/src/attachment.rs` → `crates/stella-model/src/attachment.rs`) ingests video natively only where the dialect declares `video: true` in its `DialectCaps`:

- `gemini.rs:152` (`GEMINI_CAPS`) — video `inlineData`, native
- `bedrock.rs:257` (`BEDROCK_CAPS`) — Converse video blocks, native
- `anthropic.rs:491`, `openai.rs:254`, `zai.rs:575` — `video: false`; a `@clip.mp4` mention or pasted path degrades to a descriptive `WirePart::Text` note (`degrade_note`, `crates/stella-model/src/attachment.rs:406`) naming the file.

The degrade contract is correct and total (the note tells the model what it cannot see, so the model can relay that to the user), but it means "watch a video and understand its contents" only works on 2 of 6 dialects. Every image-capable model (Anthropic, OpenAI, vision-capable z.ai/OpenRouter models) could understand a video **via sampled keyframes** — extract N frames, attach them as `AttachmentKind::Image` parts plus a text note stating the sampling (timestamps, frame count, duration, that audio was not transcribed), so the model's answer reflects what it actually saw.

## Proposed shape

- A pure decision half in the attachment layer: given a video attachment and a `DialectCaps` with `video: false, images: true`, plan a frame-sampling degrade instead of a text-only degrade. Keep the plan (frame count, timestamps) deterministic and unit-testable; no I/O in the planner.
- The extraction half (I/O — invoking `ffmpeg` or a Rust decoder) belongs outside `stella-core`/the pure layer, likely beside the hydration path in `stella-model` or as a preprocessing step in `stella-cli` before the message is built. `ffmpeg` as a runtime probe-and-degrade dependency (absent → today's text note, with the note saying why), never a hard build dependency — check `deny.toml` implications before choosing a decoder crate.
- The accompanying text note must state that the model saw sampled frames, not the video, so the model's relay to the user reflects reality. This mirrors the asymmetric-caps reasoning documented at `crates/stella-model/src/zai.rs::caps_for`.
- Cost guard: frames are image attachments; N should be bounded and probably configurable (default small, e.g. 8–16), and `MAX_ATTACHMENTS` (`crates/stella-cli/src/attachments.rs:25`) interaction decided explicitly.

## Constraints discovered

- Degradation must stay total — an unextractable/corrupt video degrades to the existing text note, never an error that kills the turn (module docs at the top of `crates/stella-model/src/attachment.rs`).
- Sending media to a model that cannot take it is a terminal 400 on some dialects (z.ai comment at `zai.rs:573`) — frame images may only ride where `images: true` for the resolved model.
- Provider parity discipline (AGENTS.md architecture rule 8): if this becomes per-provider divergent behavior, it needs declaration + witness tests in `crates/stella-model/src/provider_parity.rs`, same PR.
- Witness test required: a video attachment on a `video: false, images: true` dialect produces image wire parts + a sampling note (fails today — today it produces one text note).

## Definition of done

`stella` given `@clip.mp4` on an Anthropic/OpenAI vision model answers questions about the video's visual content from sampled frames, states clearly that it only saw sampled frames, and degrades to today's descriptive note when extraction is unavailable. Native-video dialects (Gemini, Bedrock) are unchanged.

### LINKED ISSUE #4335: The graph view is missing the free-form search box and query timing display
## Problem

Rendering `04-graph` has `q free-form query` beside the selector and `438 nodes · 12ms · det` in the query bar. `crates/stella-tui/src/v2/graph_tab.rs` elides both: the picker re-roots on a file name only (`WorkspaceInput::FocusGraphFile`), and `GraphSnapshot` (`crates/stella-tui/src/graph.rs`) carries no timing.

## Done means

- `GraphSnapshot` gains `query_ms: Option<u64>` measured by the driver's `agent::graph_snapshot_focus`, rendered `· 12ms` only when present.
- A `WorkspaceInput::GraphQuery { text }` that the driver answers through the CGP host (symbol / free-form), with the query bar switching to `q` mode on that key.
- Goldens for both in `tests/deck_render_snapshots.rs`.

macanderson:
Four parked issues from the "decide" batch, unparked: two implemented, one implemented in part with the remaining half handed back as a stated decision, one left entirely to a decision.

### #3340 — video on an image-capable dialect degrades to sampled frames, not a note

`crates/stella-model/src/keyframes.rs` (new) splits the decision from the I/O:

- `plan_timestamps` is pure and deterministic — roughly one frame per second, capped at eight, taken at segment midpoints (`t = 0` is usually black; `t = duration` often lands past the last decodable frame).
- `FrameSampler` is the seam. `FfmpegSampler` implements it with `ffprobe` + one seek-and-decode per instant, probed at call time and never a build dependency.

`resolve_one` now returns a `Vec<WirePart>` so a video fans out to one `WirePart::Image` per frame plus the note that says they are frames.

Decisions taken here rather than deferred, each reversible and each argued at the site:

- **No new `provider_parity` axis.** The behaviour is a pure function of the `DialectCaps` each dialect already declares — one code path, no per-provider divergence — so a second declaration could only drift from the first. The tests pin all three caps shapes instead (`video:true`, `video:false + images:true`, neither).
- **Sampled frames do not count against `MAX_ATTACHMENTS`.** That cap governs what the *user* attaches; the fan-out below it is bounded by `MAX_SAMPLED_FRAMES`.
- **An inline `Data` video takes the note.** Spilling a caller's payload to a temp file to hand to a subprocess is a write the user did not ask for — a separate decision from reading a file they pointed at.
- **Extraction is memoized through the existing path cache**, so `ffmpeg` runs once per file version rather than once per replayed turn. `Form` replaces the cache's `want_text` bool and is recorded on the entry: a short video sampled at a ceiling of eight is indistinguishable, by its body alone, from the same video sampled at a ceiling of three.

Degradation stays total: no decoder, an unreadable container, no decodable frame, or no file on disk each land back on exactly one text note — now naming the reason, so a user learns what to install.

### #4335 — the GRAPH tab's free-form query box

The timing half (`GraphSnapshot::query_ms`) landed in #4491; this is the other half. `q` opens a modal query box, `⏎` sends a new `WorkspaceInput::GraphQuery`, and the driver answers with a snapshot whose new `GraphSnapshot::query` echoes what was asked. The bar reads `q:<text>` in that mode and `file:<focus>` otherwise.

- The query resolves through `CodeGraph::definition_spans`, not the CGP host's `ContextQuery`: the host assembles prose frames for a *model* to read, and this tab needs a label, a kind, a file and a line per node.
- Two modals rather than one widget with a mode — the picker filters a list the deck already holds; the query box holds text the deck cannot resolve at all.
- A query matching nothing returns a snapshot carrying the query and no nodes, and the empty state says so, instead of telling a user with a misspelt symbol to run `stella init`.
- The echo rides on the snapshot rather than being remembered deck-side, so the bar cannot show a query some later snapshot did not answer — the producer-measures-its-own rule #4297 set for the timing.

Both `command_deck.rs` recv sites now route through one `command_deck::graph_input`, so idle and mid-turn cannot answer a re-root differently. The GRAPH key cluster moves out of `deck_ui.rs` into `deck_ui::graph`, and the snapshot builders out of `agent.rs` into `agent::graph_view` — both parents are at or near their line ceilings, and both shrink here.

### #4375 — `--size`, `--font-size` and a delivery ladder for `record-demo.sh` (partial — `Refs`, not `Closes`)

`--size WxH` derives the agg font size from the cast's own grid using the raster model PR #4480 measured (`width = font_size * (0.6*cols + 1.2)`, `height = font_size * (1.4*rows + 1.4)`), and ffmpeg pads the master to the exact requested size — a cell grid rarely divides a rung evenly, and shipping the near-miss is what makes asking for 4K silently drop the 4K rung. `--ladder` encodes 6144x3456 / 3840x2160 / 1920x1080 / 1280x720 / 854x480 / 640x360, H.264 High / yuv420p / faststart, one level per rung that admits 60 fps.

Two open choices in the definition of done, settled:

- **The GIF intermediate is capped, not removed.** agg writes GIF and nothing else; at 6144x3456 that intermediate is multiple gigabytes. `--size` clamps at 4K and says so; the 6K rung stays `render-deck-film.py`'s, which rasterises frames directly. The issue accepts either answer; this is the one that does not add a second frame pipeline.
- **The ladder never upscales** — only rungs at or below the master it actually has. A 6K file made by stretching a 4K master is a 4K master with a bigger filename.

The arithmetic lives in `scripts/lib/agg-geometry.sh`, sourceable and pure, so a size claim is checkable without recording anything or installing agg.

## Evidence

Every command below was run in this worktree, targeted (no workspace build), after `git merge origin/main`, with `cargo clean -p` for the three touched crates first.

**Witnesses, fail then pass**

|---|---|---|
| `attachment::tests::a_video_on_an_image_capable_dialect_resolves_to_sampled_frames` and 2 siblings | I disabled the new `resolve_one` branch (`if false && ...`) and re-ran: **3 failed**, the printed part being main's exact note, `[the user attached clip.mp4 (video/mp4, 12 B), but the current provider cannot ingest video natively; ...]` | `17 passed; 0 failed` |
| `deck_ui::tests::graph::q_opens_the_query_box_and_enter_sends_the_query` and 4 siblings | compile-level on main: `WorkspaceInput::GraphQuery` and `DeckUi::graph_query` do not exist there | `15 passed; 0 failed` |
| `deck_render_snapshots_pin_the_graph_free_form_query`, `..._the_graph_query_box` | new goldens; the bar had no `q:` mode to render | `23 passed; 0 failed` |
| `scripts/test-record-demo-size.sh` | the file it sources did not exist on main | `PASS` — reproduces both agg figures #4375 quotes (126x30 @ 20px gives 1536x868, @ 84px gives 6451x3646) |

**Targeted runs**

- `cargo test -p stella-model --lib` -> `474 passed; 0 failed`
- `cargo test -p stella-tui` -> every binary green (lib, `deck_render_snapshots`, the rest)
- `cargo test -p stella-cli --bins command_deck` -> `158 passed; 0 failed`
- `cargo clippy -p stella-tui -p stella-cli -p stella-model --all-targets -- -D warnings` -> exit 0
- rustdoc, **both forms, plain first**, per crate — `stella-model` / `stella-tui` / `stella-cli`: `RUSTDOCFLAGS="-D warnings" cargo doc -p <c> --no-deps` -> 0, and again with `--document-private-items` -> 0. Six runs, six zero exits.
- `make guards-fast` -> exit 0 (includes the `prose` and `line-citations` steps that landed on main mid-branch)
- `shellcheck install.sh scripts/*.sh scripts/lib/*.sh .githooks/*` -> exit 0
- `./scripts/check-gate-parity.sh` -> OK, 46 gate steps; the new self-test is hosted in `guard-self-tests.yml` rather than added to `UNHOSTED_SELF_TESTS`

**Goldens re-blessed — the diff, read**

`tab_graph.txt` and `tab_graph_timed.txt` each change exactly two lines: the bar hint gains `· q query` after `/ files`, and the keys row gains `· q query` before `↑↓ walk`. Nothing moved, no column shifted. `tab_graph_query.txt` and `tab_graph_query_box.txt` are new — the first shows `⌕ q:run_turn … 6 nodes · 12ms · det`, the second `⌕ q:run_tu▏` with the footer swapped to `↵ run query · esc cancel`.

## Not done

**#4338 — command palette taxonomy: left open, comment posted.** The audit verdict on it is stale: `SlashDomain`, the `relevant_now` rule over `PaletteState`, the two sections and their witnesses all landed on main already. Two bullets remain and neither has one clearly-correct answer:

1. The `recent` list needs a placement decision — the rendering puts it *after* the domain groups, which forces a choice between duplicate rows, groups made incomplete by promotion, or deviating from the rendering's order. All three are defensible and it is a user-visible browse surface.
2. The stated acceptance (`ga` during a verify turn shows `/gates` first) presumes a `/gates` command and a live gate board. Neither exists: `command_deck/skills.rs` declares 31 builtins and none is `gates`, and there is no gate-board type in the tree. That acceptance is a different feature, not a palette change.

**#4375 — `--theme` still defaults to dracula.** The definition of done wants the brand palette, and agg's theme is `bg,fg,c0..c15`. `stella-tui-theme`'s table has bg, text, gold, silver, green, red and the neutral ramp, and **no blue, magenta or cyan**. Inventing six ANSI hues is a brand decision under `make tokens` / `hue-separation` (the #4063 family), not a recorder change. Options and a recommendation are on the issue; the rest of the issue's definition of done is implemented here.

## Tests deleted or moved

**Deleted: none.**

**Moved: three**, within the same test target, no rename. CI's file-size
ratchet went red on the *merge* rather than on either side — this branch
added 63 lines to `crates/stella-tui/tests/deck_render_snapshots.rs` and
#4780 added 63 more, composing to 1501, one over the limit. Neither tree is
wrong alone; it is the shared-cell shape the ratchet's own docs describe. Split
rather than a baseline entry, which the guard refuses and AGENTS.md forbids for
a first-time crossing:

- `deck_render_snapshots_pin_the_graph_query_time` (from #4491)
- `deck_render_snapshots_pin_the_graph_free_form_query` (new here)
- `deck_render_snapshots_pin_the_graph_query_box` (new here)

all move to `crates/stella-tui/tests/deck_render_snapshots/graph.rs`, sharing
every helper through `use super::*` and blessing with the same command. The
parent is 1434 lines with main merged in, so the next tab that wants a row has
room. `#[path]` is used because the target's module root is `tests/`, and
Cargo compiles every `tests/*.rs` as its own binary — a plain `mod graph`
would build these three goldens twice under two names.

The same push repairs a comment this branch made stale: `attachment_block`'s
audio/video arm in `anthropic.rs` said `wire_parts` had "already degraded them
to Text notes", now true only of audio.

Closes #3340
Closes #4335
Refs #4375
Refs #4338

## Summary by Sourcery

Enable sampled-frame video degradation, free-form GRAPH queries, and size-aware demo video generation.

New Features:
- Degrade unsupported video attachments to bounded, cached sampled image frames with explanatory notes on image-capable dialects.
- Add a free-form symbol query workflow to the GRAPH tab with query-aware snapshots, rendering, and driver resolution.
- Add demo recording output sizing, font-size derivation, exact-size encoding, and a non-upscaling delivery ladder.

Bug Fixes:
- Preserve total attachment degradation when video sampling fails by returning actionable explanatory notes instead of aborting turns.
- Ensure idle and mid-turn GRAPH re-root requests share the same response path.

Enhancements:
- Refactor GRAPH snapshot construction, input handling, and UI keyboard handling into dedicated modules.
- Extend attachment path-cache entries to distinguish text, binary, and sampled-frame hydration forms.

CI:
- Add a hermetic CI self-test for demo recorder raster and ladder arithmetic.

Documentation:
- Document output-size behavior, 4K limits, and delivery-ladder semantics for both demo video pipelines.

Tests:
- Add coverage for video sampling plans, degradation fallbacks, caching, GRAPH query interactions, and rendered query states.
- Move GRAPH snapshot golden tests into a dedicated module and add query and query-box goldens.


macanderson:
## Answering the one ❌: "resolves queries directly through `CodeGraph::definition_spans` rather than answering them through the CGP host"

Sourcery is reading #4335's wording correctly — the issue does say *"a `WorkspaceInput::GraphQuery { text }` that the driver answers through the CGP host (symbol / free-form)"*. I did not do that, on purpose, and the reason is that the host path cannot carry what this tab draws. The evidence, from the code rather than from the issue:

**What the CGP host returns is prose, not structure.** `contextgraph_types::ContextFrame` is `{ id, kind, title, content: Option<String>, content_digest, uri, score, ... }`, and `content`'s own doc comment says it is *"Text the host may quote into a prompt. Untrusted data: a conforming host delimits this as quoted material, never as instructions."* There is no node kind, no line number and no edge list on a frame.

**What the GRAPH tab draws is `GraphNode { label, kind, location }` plus `GraphEdge { from, to, kind }`** — `"function"` / `"struct"` / `"file"`, `"src/x.rs:48"`, `"defines"` / `"imports"`. To render a neighborhood out of frames I would have to re-parse a citation string (`symbol_frame` in `crates/stella-graph/src/frames.rs` builds titles shaped `fn run_turn (crates/...:48)`) back into the fields `definition_spans` already returns typed.

**And the host path does work the tab would then throw away.** `GraphProvider::query` (`crates/stella-cli/src/contextgraph.rs`) fans out to `CodeGraph::query`, which in `frames::query` reads a **snippet file off disk per definition**, score-ranks, dedups by score, and `budget_pack`s the result against `max_frames` / `max_tokens`. Every one of those steps exists to fit a model's context window. The tab has no token budget, wants every match rather than the top-scoring packable subset, and discards the snippets.

Both paths open the same `.stella/private/codegraph.db` and both start from the identifier in the query text. The only difference is what comes back: frames for a prompt, or spans for a graph.

The argument is at the call site so a later reader meets it rather than the bare choice (`crates/stella-cli/src/agent/graph_view.rs`, `graph_query_snapshot`'s doc comment):

> `text` is resolved as a **symbol name** against the index's definitions. That is narrower than the CGP host's `ContextQuery`: the host assembles prose frames with snippets and provenance for a *model* to read, while this tab needs structure — a label, a kind, a file and a line per node — and rendering prose as a graph would be a worse answer than the one the index can give directly.

So this row is **answered, not fixed**.

If the intent behind the issue's wording was that the tab must go through the host so a *remote* CGP provider could answer a graph query, that is a real feature and a different one: it needs a structured node/edge frame kind in the protocol first, and the protocol lives in the `context-graph-protocol` repo, not here. Say so and I will file it there rather than bend this tab around a frame shape that cannot hold a graph.

The other five rows are ✅ and need nothing.

macanderson:
Correcting my own comment above: the ❌ row was fixable, and the reasoning I gave for answering instead of fixing it was wrong. #4817 fixes it; #4818 tracks what genuinely cannot be fixed here.

I argued the CGP host could not carry a graph. Reading the pinned `contextgraph-types 0.1.2` rather than recalling it:

- **"There is no … edge list on a frame."** `ContextFrame.relations: Vec<Relation>` (`frame.rs:319`), where `Relation { rel, target_uri, display_name }` (`frame.rs:190`). `SPEC.md` §G1/§G2 are conformance requirements written specifically for graph-capable providers.
- **"There is no node kind."** `FrameKind::Symbol` and `FrameKind::Graph` (`frame.rs:25-33`).
- **"It needs a structured node/edge frame kind in the protocol first, and the protocol lives in the `context-graph-protocol` repo, not here."** Both already exist in the pinned version, and `stella-graph`'s own `edge_frame` (`crates/stella-graph/src/frames.rs:485`) has been emitting `FrameKind::Graph` frames carrying those relations all along. Nothing needed to land upstream first.

What I got right, and what #4817 pays rather than denies: the frame route does lose fidelity the direct read keeps. `SymbolKind::keyword` (`symbol.rs:96`) projects both `Function` and `Method` onto `"fn"`, and a symbol's line survives only as the string `"L12-40"` in `provenance.range`. That cost is now documented at `node_kind` and `frame_location` and pinned by tests, and #4818 tracks the protocol change that would remove it.

Routing through the host also surfaced two defects that the direct path had hidden, both caught by the new tests rather than by inspection:

- Frame URIs are minted against the index's root and `CodeGraph::open` canonicalizes it, so on any workspace reached through a symlink — every macOS `/tmp` and `/var` path — stripping the deck's spelling of the root cost every node its file and line.
- A fan-out where every provider leg timed out, was refused, or crashed is indistinguishable from "no matches" at the frame layer, so a wedged provider reached the user as "your query found nothing".

The other five rows on the assessment table stand as ✅.

Issue #4403, opened 2026-08-23
One part of the server builds its engine manually instead of using the shared builder


macanderson:
## Problem

`crates/stella-runtime` exposes `RuntimeSpec → RuntimeBuilder → SessionRuntime` as the shared engine-assembly bottom half, but no production caller constructs the composite: `stella-cli` drives only the `parts::*` functions and `stella-serve` assembles by hand. PR #4402 (closing #3731) documents this in the crate README and `lib.rs` module doc and cites it as gap G1 in `doc:engine-embedding` §4 — but #3731 is closed by that PR, so the gap needs a live tracking number, which is this issue.

## Done means

- `crates/stella-serve` builds its session through `RuntimeBuilder::with_provider(...)` (or the builder is deleted and the README says `parts::*` is the whole surface — a decision, stated in the PR).
- `crates/stella-runtime/README.md`, `src/lib.rs`'s module doc and `doc:engine-embedding` §4 G1 are updated to cite this issue (or to drop the gap) in the same PR.
- Witness: `crates/stella-runtime/tests/runtime_build.rs` (added in #4402) gains a test driven through the same path serve uses, or the serve crate gains one that proves the builder is on its live path.

Constraints: architecture rule 1 (ports, not direct dependencies); `stella-runtime` reads no ambient environment (`tests/no_ambient_reads.rs`).

Related: #3731, #3733, #4402.

## Definition of done

The “Done means” prose above, as boxes the gate can read — nothing here is
new work, and nothing is ticked by the migration (#5193).

- [ ] `crates/stella-serve` builds its session through `RuntimeBuilder::with_provider(...)` — or the builder is deleted and the README says `parts::*` is the whole surface, as a decision stated in the PR.
- [ ] `crates/stella-runtime/README.md`, `src/lib.rs`'s module doc and `doc:engine-embedding` §4 G1 are updated to cite this issue, or to drop the gap, in the same PR.
- [ ] Witness: `crates/stella-runtime/tests/runtime_build.rs` covers whichever shape was chosen.

Issue #3783, opened 2026-08-18
Nothing in the program actually starts a self-driving session; the code that would do it is never called


macanderson:
## Problem

#3634 gave the driver channel a transport (`crates/stella-runtime/src/wrapper/driver_subprocess.rs`), so a driver process can now be spawned, handed a `DriveRequest`, served a capability ask, and read back for its `next`. **Nothing in `stella-cli` — or anywhere else outside `crates/stella-runtime/tests/driver_socket.rs` — constructs one.**

So `plugins/stella-selfdriving`'s `[driver]` block is still a declaration rather than a running program: a human can read the grant at install, and no code path turns it into a session. This is the wiring half of the shape AGENTS.md § "Nothing left behind" names — scaffolding that something else must later pick up — filed in the same breath as the PR that shipped it.

## Where

- `crates/stella-runtime/src/wrapper/driver_subprocess.rs` — `SubprocessDriver::declare` / `serving` / `drive`.
- `crates/stella-runtime/src/wrapper/driver_call.rs` — `DriverCallGate::declare`, which a host must build first.
- The shape to copy: `crates/stella-cli/src/wrapper_plugin.rs::bind_installed`, which does exactly this for `SubprocessWrapper` — resolves argv against the installed package directory, interpolates the plugin-dir placeholder, resolves the manifest's `[runtime] env` allowlist through `child_env`, warns per `AdmittedWrapper::refused`, and attaches the gate.
- `crates/stella-cli/src/plugin_cmd/roster.rs` — where an installed manifest's routes are assembled today.

## The blocker a wiring PR hits first

**`plugins/stella-selfdriving/plugin.toml` declares no `[runtime]` block**, so it names no argv, no env allowlist and no timeout — every input `SubprocessDriver::declare` needs is absent for the one plugin in the tree that declares `[driver]`. Deciding that is part of this issue and should be settled before code:

1. reuse `[runtime]` as-is (one block, two dispatch contexts), or
2. give `[driver]` its own argv/env/timeout keys (two consents, read separately), or
3. something else.

Option 1 is the smaller change; option 2 is the one that matches the argument `[driver]` itself was split out on (`crates/stella-plugin/src/driver.rs`'s header: a driver grant and a loop grant are different consents and neither grants any part of the other). Whichever is chosen, the manifest half lands in `crates/stella-plugin/src/manifest.rs` with a load-time test, and the consent rendering in `crates/stella-plugin/src/consent.rs` has to show it.

## Constraints already discovered

[...]

Issue #3731, opened 2026-08-18
stella-runtime README/lib.rs claim a shared RuntimeBuilder/SessionRuntime host path that nothing drives yet


macanderson:
## Problem

`crates/stella-runtime/README.md` (lines 3-6) and `crates/stella-runtime/src/lib.rs` (lines 15-17) describe `RuntimeSpec`/`RuntimeBuilder`/`SessionRuntime` in the present tense as the shared assembly path "any surface can drive: the CLI, the serve sidecar, or a test."

That is not true today:
- `stella-cli` only calls the individual `parts::*` free functions (`build_provider`, `open_store`, `seed_calibration`, `budget_guard`) directly. It does not construct a `RuntimeBuilder`/`SessionRuntime`.
- `stella-serve` does not depend on `stella-runtime` at all.
- Grepping the workspace for `RuntimeBuilder` and `SessionRuntime` construction shows zero call sites outside the `stella-runtime` crate itself.

This gap is already known and written up as gap **G1** in `docs/spec/engine-embedding.md` (lines 81-88 and 219-224), but that spec doc cites no GitHub issue number, so nothing tracks the work, and the crate's own front-door docs (the first thing a contributor reads) still assert the composite is live.

## Files involved

- `crates/stella-runtime/README.md` (opening paragraph, lines 3-6)
- `crates/stella-runtime/src/lib.rs` (module doc, lines 15-17)
- `crates/stella-runtime/src/session.rs` (`RuntimeBuilder`, the unwired composite)
- `docs/spec/engine-embedding.md` (lines 81-88, 219-224 — where gap G1 is already described)
- `crates/stella-cli` (uses `parts::*` directly, does not use the composite)
- `crates/stella-serve` (no dependency on `stella-runtime` at all)

## Fix options (pick one — needs a human decision)

**(a) Wire a real host onto RuntimeBuilder.** `stella-serve` is the intended host; `RuntimeBuilder::with_provider` was built for its `RemoteProvider` case. Making `stella-serve` actually construct a `SessionRuntime` via `RuntimeBuilder` would make the present-tense doc claim true.

[...]

PR #4801, opened 2026-08-25
feat(stella-core,stella-cli): restore the event-stream checker; give every fleet attempt its own resume point

### LINKED ISSUE #3232: fleet: worker turns are never checkpointed — the whole fan-out has no step-level durability

### LINKED ISSUE #4585: Nothing checks that an event log has the right structure since the old checker was deleted
## Problem

`stella_pipeline::replay::validate_stream` (stage ordering, `tool_start`/`tool_result` pairing, single terminal `complete`, monotonic budget) was deleted with the crate in #3865, along with `tests/fixtures/from_a_newer_stella.jsonl` and `tests/stream_conformance.rs`. `docs/wire/README.md` cited both as present-tense executable pins; PR #4582 corrected the prose to say nothing in this tree checks them — but nothing replaces the validator, and `docs/spec/replay-golden-trajectories.md` still references it.

## Done means

A decision, recorded where the docs point: either (a) the structural validator returns as a small pure checker in `stella-protocol` (no I/O; these rules are properties of the event stream the protocol crate owns) with the forward-compat fixture restored, or (b) these rules are declared the verification plugin's job per doc:pipeline-as-plugins §8, and both doc pages say so. Option (a) is the recommendation: every door emits these streams, and a malformed stream is currently detectable by nothing.

Found in PR #4582 (2026-08-23). Related: #3865, #3783.

macanderson:
### #4585 — nothing checked that an event log had the right structure

`stella_pipeline::replay::validate_stream` checked the half of the wire contract JSON Schema cannot state: legal stage ordering, `tool_start`/`tool_result` pairing, a single terminal `run_complete`, monotonic budget. #3865 deleted the crate and the validator went with it, so a malformed stream from any door has been detectable by nothing since.

`crates/stella-core/src/event_stream.rs` is the checker back — `validate_stream` over a parsed stream, `conform_jsonl` over a recording.

**The decision the issue asked for, and why it went the other way.** #4585 recommended `stella-protocol`. That crate's own boundary rule sends it to `stella-core` instead: it admits "a serde type — or a field on one — plus at most a total, allocation-light helper over that type's own data", refuses "a `match` that decides what the program *does* next", and names `stella-core` as the home for "decision logic over events". A fold over a whole stream with a rank table, a legal-back-edge rule and a violation vocabulary of its own is the second thing. It reads no file, spawns nothing and takes no clock.

Three departures from the deleted code, each stated in the module header:

- `StreamViolation` is an enum, not `{ index, reason: String }`. The old conformance test asserted on substrings of that prose; the negative-case test is now a `match` the compiler completes.
- A stage name the host does not know is passed over rather than ranked. The stage vocabulary opened after the validator was written (#3398), so a name a plugin contributed has no rank — judging it would invent an order nobody declared and flag the same run twice.
- Lines read as `journal::StampedEvent`, so a real `stella-events.jsonl` (whose sink stamps every line) is readable, not only a bare stream.

`structural_diff`, `streams_equivalent` and the golden-fixture format are **not** restored: those compare two recordings of the same work, which is the trajectory-fixture approach `doc:replay-golden-trajectories` records and a verification plugin's to implement. `docs/wire/README.md` and that page both now say where the surviving checker lives.

### #3232 — fleet worker turns were never checkpointed

`tuned_engine_config` reads `cfg.durability.sink()`, four production sites bind that handle, and none of them was the fan-out — so `sink()` answered `None` for every worker turn. A worker killed at step 40 of 50 kept whatever it wrote to the tree and lost the conversation that produced it.

`crates/stella-cli/src/fleet_cmd/durability.rs` binds each attempt a record of its own, keyed on the claim holder (`{run_id}/{task_id}`) the fleet already coordinates that task's file claims under. Never the handle on `cfg` — `Config` clones over one `Arc`, so binding that makes every worker in a wave a writer of one resume point, the hazard `durability.rs`'s "What must NOT inherit a binding" section names. Rooted at the invocation root rather than the task tree, because an isolated worktree may not outlive the run. The key sanitizes, caps and hashes, because a task id comes out of a plan file: `WorkJournal::open` needs a name that is filesystem-safe and ref-safe at once, and a failure there is swallowed by the sink's best-effort contract — so an unsanitized key would leave the whole feature inert with no symptom short of a kill losing a transcript.

**The recorded blocker was verified rather than trusted, and it is stale.** #3232 says the write side alone buys a whole-transcript serialization per step with no reader, and offers a declined-decision close on that basis. #3233 has since built the reader while closing the sub-session half: a lane binds its own key and re-enters an interrupted transcript from it (`crates/stella-cli/src/subsession.rs`, `lane_journal_key` + `initial_messages`). A fleet attempt has the same shape and the same reader — `Fleet::dispatch` is re-runnable once an attempt settles (`crates/stella-fleet/src/fleet.rs`: "Re-dispatch a task after its previous attempt settles"), and a re-dispatch derives the same key. So the fix lands rather than the decline.

What it does not recover is a killed *process*: a fresh `stella fleet` mints a new `run_id`, so its keys are new. Dropping `run_id` from the key would let two concurrent runs over one workspace write each other's resume points, so cross-invocation resume needs a run identity a later process can name — a fleet-ledger question, filed rather than guessed at.

`subsession_engine_config_for` is the seam both doors take; its doc says so now, and `tuned_engine_config`'s comment that named this gap says what closed it (DoD item 5).

### #4403 — partial: G1's refuted direction retired

`docs/spec/engine-embedding.md` §4 G1 still cited #3731 (closed by #4402) and still named the serve sidecar as the caller that would close the gap. Serve cannot be: it depends on `stella-protocol`, `stella-core` and `stella-engine` and on no store or tool crate, because it executes no tool and holds no credential, while `stella-runtime` depends on `stella-model`, `stella-tools` and `stella-store` and `RuntimeBuilder::build` refuses a `workspace_root` that is not a directory before returning an `Arc<ToolRegistry>` rooted at it. G1 now cites #4403 and states what is actually open. **`Refs`, not `Closes`** — see "Not done".

## Evidence

Targeted commands only; CI is the gate. Every row below was re-run **after** `git merge origin/main`, with `cargo clean -p stella-core -p stella-cli` first so nothing replayed a cached pre-merge result.

| Command | Result |
|---|---|
| `cargo test -p stella-core event_stream` | ok, 20 passed |
| `cargo test -p stella-core --test stream_conformance` | ok, 4 passed |
| `cargo test -p stella-cli --bin stella fleet_cmd::durability` | ok, 9 passed |
| `cargo clippy -p stella-core --all-targets -- -D warnings` | exit 0 |
| `cargo clippy -p stella-cli --all-targets -- -D warnings` | exit 0 |
| `RUSTDOCFLAGS="-D warnings" cargo doc -p stella-core --no-deps` | exit 0 |
| `RUSTDOCFLAGS="-D warnings" cargo doc -p stella-core --no-deps --document-private-items` | exit 0 |
| `RUSTDOCFLAGS="-D warnings" cargo doc -p stella-cli --no-deps` | exit 0 |
| `RUSTDOCFLAGS="-D warnings" cargo doc -p stella-cli --no-deps --document-private-items` | exit 0 |
| `make guards-fast` | exit 0 — `check-prose` none added, `check-file-size` `fleet_cmd.rs` at 1499 of 1500 |

`fleet_cmd.rs` had 8 lines of headroom under the ratchet, so all of #3232's logic lands in the `fleet_cmd/durability.rs` sibling and the parent's net growth is 7 lines.

### Witnesses

**#4585 — compile-level, and that is the honest grade.** The module is new, so on `main` there is no `validate_stream` to call and `cargo test -p stella-core --test stream_conformance` fails with no such test target. What the tests do prove is the half a new checker most often gets wrong: every rule has a negative case that trips it, an unrecognized event type is read rather than rejected, a torn tail is tolerated, and an interior malformed line is an error rather than a clean bill of health.

**#3232 — fail→pass, both arms inside the test.** `a_fleet_attempt_checkpoints_into_its_own_record` asserts arm 1 first — that `agent::engine_config_for` over a fleet worker's unbound `Config` yields `checkpoint_sink: None`, which is the defect, and which still holds on this branch because that is what the old wiring did — then arm 2, that the seam now produces a sink whose `persist` reads back off the attempt's own record. Without arm 1 the test would pass on a build where the damage is merely unreachable rather than prevented. `two_concurrent_attempts_write_different_records` is DoD item 4; `a_redispatched_attempt_re_enters_its_interrupted_transcript` is the reader.

Driven at the sink seam rather than through a live `run_task`, for the reason `agent/tests/durability_isolation.rs` gives for the same choice: reaching it needs a provider, a registry, a git tree and a fleet ledger, none of which change which sink the config carries. The one line of `bind` a test cannot execute is `WorkJournal::open`, which resolves its store from `STELLA_HOME` — a process-global a test must not touch; the key it would be handed is covered directly, and the call itself is the one four other doors already make.

**#4403 — no witness; documentation correction only.**

### Tests deleted

None.

## Not done

**#3783 — skipped: the manifest question it says to settle first is a maintainer's, and the tree answers it against wiring.** The issue asks for a host that constructs a `SubprocessDriver`, and names the blocker: the one plugin in the tree declaring `[driver]` has no `[runtime]` block, so nothing names an argv, an env allowlist or a timeout. `plugins/stella-selfdriving/plugin.toml`'s header does not merely omit those — it says why they would be false: "self-driving is a HOST, not a wrapper (option 2) … that is why `[loop] participation` is `none` below and why there is no `[runtime]`, no `[oracle]` and no `[wrapper]`: none of them would be true. **Stella never starts this program — a person does, and then it starts Stella.**" `SubprocessDriver` spawns the driver as Stella's child, which is the opposite direction, so either option the issue offers — reuse `[runtime]`, or give `[driver]` its own argv/env/timeout — inverts what `doc:pipeline-as-plugins` §10 settled about what this plugin is. That is an architecture decision above a fix batch, and the issue itself says it "should be settled before code". Nothing is half-landed for it.

**#4403 — advanced, not closed.** Its remaining content is a choice between retargeting `RuntimeBuilder` at `stella-cli`'s drivers and deleting the composite. The first is a refactor across the CLI's session assembly (`agent/engine.rs`, `agent.rs`, `command_deck.rs`, `agent/tool_stack.rs`); the second removes a shape `doc:engine-embedding` builds its Mode-A quickstart on. Both are decisions about who the crate serves rather than defects to fix, so this PR corrects the stale half and leaves the choice.

**Follow-up to file:** cross-invocation fleet resume. A fresh `stella fleet` mints a new `run_id`, so an attempt's key is new and a process-level kill still starts clean. Making it resumable needs a run identity a later invocation can name (the fleet ledger already stores runs); `run_id` cannot simply be dropped from the key, because two concurrent runs over one workspace would then write each other's resume points.

Closes #4585
Closes #3232
Refs #4403
Refs #3783

## Summary by Sourcery

Restore event-stream validation and add isolated checkpoint and resume support for fleet attempts.

New Features:
- Restore event-stream conformance checking for stage ordering, tool-call pairing, terminal completion, and monotonic budget, including JSONL recordings with forward-compatible unknown events and torn-tail handling.

Bug Fixes:
- Give each fleet attempt an isolated durable checkpoint and resume transcript across re-dispatches, preventing workers from losing conversation state or overwriting one another's resume points.

Enhancements:
- Clarify the ownership and limitations of the event-stream checker and fleet durability seam in project documentation.
- Correct the engine-embedding specification to reflect the current unresolved runtime assembly direction.

Documentation:
- Document the restored event-stream checker and update replay and engine-embedding guidance to identify its new location and current architectural status.

Tests:
- Add unit and integration coverage for stream validation, JSONL parsing, forward compatibility, malformed recordings, fleet checkpoint isolation, and interrupted-attempt resumption.


macanderson:
## Answering Sourcery's ❌ row

> **#3232 — Provide a resume path that re-enters an interrupted fleet attempt's checkpointed transcript instead of restarting from an empty history. ❌** … explicitly does not recover a process killed and later restarted through a fresh `stella fleet` invocation.

The reading is correct and the limitation is real; it is deferred rather than missed, and here is why it is not fixable inside this PR.

**The obvious fix is unsafe.** The attempt's record is keyed on the claim holder, `{run_id}/{task_id}`. Dropping `run_id` would make a task's record findable by a later invocation — and would also let two concurrent runs over one workspace write each other's resume points. That is two writers of one record, which is the hazard `crates/stella-cli/src/durability.rs`'s "What must NOT inherit a binding" section exists for and the one this PR's key shape was chosen to avoid. Concurrent runs over one clone are a supported shape here, not a corner: `dispatch_claims` exists so two loops over one workspace can see each other.

So process-level resume needs a run identity a **later process can name** — the fleet ledger already persists runs in `.stella/private/fleet.db`, and what is missing is a way for a new invocation to adopt a prior run rather than open a new one. That also forces a decision this PR has no business taking alone: what a resumed run does with the worktrees and claim leases the dead run left behind (`stella_fleet::git::worktree_slug` hashes `run_scope` into the worktree name for the same freshness reason). It is a fleet-ledger change with its own witness, filed as **#4802** with the file paths, the repro, the constraints above and a definition of done.

**The issue's own definition of done admits a scoped answer here.** Item 2 reads "A resume path that can re-enter an interrupted fleet attempt — or, if that is judged out of scope, an explicit recorded decision …". This PR takes the first half for the case it can serve soundly and records the boundary in the module docs (`fleet_cmd/durability.rs`, "What this does and does not recover"), now citing #4802 so the paragraph is a handoff rather than a note. Items 1, 3, 4 and 5 are met in full, and Sourcery marks each of them ✅.

**Why `Closes` and not `Refs`.** The defect the issue is titled for — "worker turns are never checkpointed — the whole fan-out has no step-level durability" — is fixed: every attempt now binds a record, checkpoints into its own ref namespace, and a re-dispatch re-enters its transcript. What remains is a distinct capability with its own tracking number. If a reviewer would rather this stay open until #4802 lands, say so and I will downgrade the trailer to `Refs #3232` — that is a maintainer's call about the issue's scope, not a defect in the diff.

Issue #4289, opened 2026-08-22
Make interactive mode draw its transcript frame through the one shared renderer


macanderson:
Implementation half of #4271, whose design question is now settled: SPEC 6 is the one frame, and `grid.rs` is changed to draw it rather than interactive mode being bent onto `grid`'s box. The decision and its reasoning are recorded in `crates/stella-transcript/src/lib.rs`'s charter under "Which frame is the one frame" — read that first. This issue is the implementation work; the design decision is already made.

## Why this is not one PR

This splits into separable pieces, in order, because piece 1 is a correctness precondition for piece 3.

### 1. `stella_transcript::model` gains a notion of an *unmeasured* row — do this first

`stella-tui`'s `v2::transcript::Extent` makes every file-event size field an `Option` for a stated reason: a head row renders the moment its call dispatches, and nothing is measured yet. `+0 -0` beside a path asserts the edit changed nothing, which is a louder and entirely different claim than "not measured yet" — and the fabricated zero already shipped once as a defect (#2290, #4156).

`stella_transcript::model` has no equivalent. Migrating interactive mode onto `grid` **before** this lands would reintroduce the zero, silently, which is the one outcome that makes the whole migration a regression. Same gap #4181 describes from the accessibility side.

- Files: `crates/stella-transcript/src/model.rs`, `crates/stella-tui/src/v2/transcript.rs` (`Extent`, and its doc comment, which is the clearest statement of the rule in the tree).
- Done when: a `Run` fixture with an unmeasured call renders through `grid` with **no** size column at all, and a test asserts the absence rather than a zero.

### 2. `grid::Color` gains the two metals as roles

Not as RGB. This crate stays free of a terminal dependency, and the seam that makes that work already exists: `grid` emits a *role* and encodes at the very end. So `Color` gains `Gold` (stella acting) and `Silver` (world coming in), and each surface maps role → pigment at its own encode step — interactive mode to `stella_tui_theme::token`, the plain surface to its 256/16-colour codes.

Note the existing `Cyan` ("bash and tool identity") is the per-class hue SPEC 3.2's clamp rejects and #4125 is about removing. Decide its fate in the same PR as #4125, not separately, or the two will disagree.

- Done when: the metals are roles in `grid::Color`, both encoders map them, and `stella-transcript` still has no `stella-tui-theme` dependency (check `Cargo.toml`, not just imports).

### 3. `grid` draws SPEC 6's frame

[...]

Issue #3578, opened 2026-08-17
Interactive mode still uses its own diff renderer instead of the shared transcript renderer


macanderson:
## Problem

`crates/stella-transcript` holds the shared transcript model and a character-grid renderer (`grid.rs`) that produces the folds, digests, chips and word-level diffs the design calls for. The Observatory renders through it (`/transcript?id=N`). Interactive mode does not yet — `crates/stella-tui/src/render/entry.rs`, `render/row.rs`, `diff.rs` and `views/session.rs` still carry the older, separate implementation.

Until that lands, the two surfaces are once again two renderers, which is the exact drift the crate was extracted to end.

## Files

- `crates/stella-tui/src/views/session.rs` (1,611 lines, **grandfathered god file — do not add lines**; it is the fold/scroll/layout layer)
- `crates/stella-tui/src/render/entry.rs` (953 lines, `entry_lines()` is the per-entry renderer)
- `crates/stella-tui/src/render/row.rs` (`Rail` — the current `●`/`⏿`/`✗`/`▌` gutter vocabulary)
- `crates/stella-tui/src/diff.rs` (1,099 lines — the terminal diff painter `stella-transcript::grid` supersedes)
- `crates/stella-transcript/src/grid.rs` — the target renderer; emits `Vec<Cell>` rows that a ratatui `Line` maps onto directly.

## Constraints already discovered

- `views/session.rs` and `deck_render.rs` are both in `scripts/file-size-baseline.txt` and closed to growth. New wiring lands in a sibling submodule.
- `crates/stella-tui/tests/snapshots/deck/tab_session.txt` is the golden. It **will** change; regenerate with `BLESS=1 cargo test -p stella-tui --test deck_render_snapshots` and read the diff rather than blessing it blind (AGENTS.md is explicit about this).
- `grid::Cell` carries `fg: Color` / `tint: Tint` / `bold`, deliberately not ratatui types, so the mapping is a small adapter and `stella-transcript` stays free of a terminal dependency.
- `FoldState`/`Cursor`/`Command` in `stella-transcript::fold` already model `j`/`k`/`←`/`→`/`z`/`e`/`c`; interactive mode should bind to those rather than reimplementing them.

## Verify

The acceptance checks are already written as tests in `crates/stella-transcript/src/tests.rs` (`acceptance_*`). The interactive-mode witness is that its session view renders one digest row per step at `Zoom::Steps` and that a failed step is still expanded after the run finishes.

## Done when

[...]

PR #4803, opened 2026-08-25
The terminal UI has one rendering generation: finish the port, delete v1, dissolve the namespace

### LINKED ISSUE #4626: AGENTS page polish: no terminal caret in its composer, and command replies land behind the page

### LINKED ISSUE #4676: Epic: the terminal UI has one rendering generation, not two — finish the v2 port and delete v1

macanderson:
Closes #4676.

Replaces #4786, which was branched from `c88af1e19` and carried a duplicate of #4626 — see the note at the foot.

## Where the epic actually stood

Written against a `main` where `views/` held eighteen modules. By the time this branch opened it held three, and none of them still rendered the pre-SPEC-5 way:

| file | what it actually was |
| --- | --- |
| `views/graph.rs` | an empty state and a cursor clamp wrapping a call to `v2::graph_tab` |
| `views/settings.rs` | a one-line nav; all three panes already dispatched into `v2::*` |
| `views/cards.rs` | the floating-card chrome the `v2` tree imported eleven times |

So the remaining substance was items 2–6 of the definition of done, not item 1. `views/engine.rs` — one of the two god files item 5 names — had already left the file-size baseline.

## What landed

**The last three modules are gone.** `graph_tab` absorbed its shim's empty state and clamp (its inner draw is now `paint`); `settings.rs` and `cards.rs` moved as they stood; `spinner_glyph` joined `status_glyph` in `theme`, the module that exists so every view takes its glyphs from one place.

**`src/v2/` is `src/views/`.** Flattening to `src/` was rejected: it collides on `graph` and leaves sixty-odd files in one directory with nothing saying which of them draw a surface. `views` is the crate's own word for these modules and the one `deck_render` dispatches into per `DeckTab`. **If you want a different name, this is the cheap moment to say so** — it is a `git mv` plus a path rewrite.

**The prose split two ways, and this is the part worth reviewing.**

Comparisons that existed only to justify a deletion already made are *deleted*: `frame.rs`'s account of the eleven rows the old deck spent, `status_bar.rs`'s case for one row over two and its stale "not wired yet" note, the README's statline row, `status_source.rs`'s cell-for-cell equivalence claim.

But `src/render/` **genuinely still has two row renderers**, and item 4 taken literally would have made its docs lie. `projected_rows` (was `v2_rows`) routes a tool call's head and compaction's line through the shared projection; everything else falls through to `entry_body`. That is live structure. So there the generation words became what the two arms actually are, and the transcript-surfaces ledger's `deck-v1`/`deck-v2` became `deck-body`/`deck-spec6`. The unfinished half — SPEC 6.4's result row, which waits on a highlighter — is already tracked by #4289 and #3578; it is not silently absorbed here.

## Three guards had to follow the move

Each would have failed as a *stale* check rather than a wrong one, which is the interesting part: the hex-literal sweep's roots (`src/v2.rs`, `src/v2`), the transcript-surfaces ledger, and `keymap.rs`'s key-witness `include_str!`. The last is `#[cfg(test)]`, so `cargo check --lib` was green while `cargo test` could not build.

The sweep is the one that would have caught itself — it asserts it looked at something. The other two were caught by `make guards-fast`.

## Evidence

- `cargo test -p stella-tui --no-fail-fast`: 1122 lib tests + all nine integration suites green.
- **The deck's whole-frame goldens are byte-identical** — no re-blessing was needed, which is the evidence the rename moved no pixels.
- `make guards-fast` green, `cargo fmt --all --check` clean, `cargo clippy -p stella-tui --all-targets` clean.
- `views/graph_tab/tab.rs` carries the six tab-level tests from the deleted shim, unchanged in what they assert.

## Baselines

The prose baseline **loses nine entries and gains none**: seven for renamed files whose debt had to be paid rather than carried (a renamed path is a new file, held to zero), one already-dead entry, one renamed. Since #4763 landed the filler-adverb rule, that debt was larger than it looked — 38 lines across 24 file/category groups, all deleted rather than re-based.

`deck_ui.rs`'s file-size ceiling is retightened 3675 → 3661, the lines the port took out. Both baselines were hand-edited rather than regenerated, so this PR does not sweep up unrelated debt other branches have reduced.

## Test renames, for the deleted-tests guard

`tests/v2_status_bar.rs` → `tests/status_bar_goldens.rs`, with its three insta snapshots renamed to match and no golden content changed. Renamed test functions: `no_hex_literals_in_v2_render_code` → `no_hex_literals_in_render_code`, and `v2_status_bar_{at_deck_width,narrow,at_the_extremes}` → `status_bar_{...}`. `views/graph.rs`'s six tests moved to `views/graph_tab/tab.rs` under their own names. No test was deleted.

## Why this replaces #4786

#4786 opened at 03:23 against `c88af1e19` and bundled a fix for #4626, because the epic's rename touches every file #4626 edits. #4780 landed the same two fixes on `main` at 03:56 and closed #4626. Rather than resolve a rename against a 240-line rival diff in the same file, this branch is built fresh on `77df4b9a2` with **main's #4626 implementation kept as it stands** — it simply moves with the namespace. #4786 is closed unmerged; nothing in it is lost that main does not already have or this branch does not redo.

One thing that came out of my duplicate cut is worth recording even though main's implementation stands: an entry-count-only reply mark renders an **empty** pane when the reply lands on the heels of assistant prose, because `push_text` appends to a trailing `Text` entry rather than pushing a new one. Main's version carries a byte offset alongside the count, so it is not exposed to this — noted here only so the reasoning is on the record.


macanderson:
Merged `origin/main` (`ce6b1cf3c`) into this branch — `ae4917f34`. The PR is `MERGEABLE` again.

### What conflicted, and how it was resolved

main's #4799 landed the GRAPH query box inside `v2/`, which is exactly the namespace this PR deletes. Four files conflicted; the query box moves with everything else rather than being dropped:

- **`views/graph_tab.rs`** — main renamed `paint` to `render` and added a `typing` parameter plus the three-state query bar. This branch already owns `render` as the deck's tab signature, so the `typing` parameter and its doc land on `paint`, and `render` passes `ui.graph_query.as_deref()` through.
- **`views/graph.rs`** — on main this file is the v1 shim that delegates to `v2::graph_tab`; on this branch it is the coupling panel. The branch's content is kept; main's query-aware empty state ("nothing in the index matches `x`", #4335) moves onto `views::graph_tab::render_empty`, which is where the deck entry point lives now.
- **`v2/graph.rs`** — stays deleted. main's only change to it was a `query: None` test-fixture field, which lands in the fixtures' new homes (`views/graph.rs` and `views/graph_tab/tab.rs`).
- **`deck_ui.rs`** — took both sides: main's `graph_query` field and this branch's `crate::views::mcp_tab` path.

One thing beyond the conflict markers: `views/cards.rs`'s doc comments from #4778 named `crate::v2::approval`, `v2::picker`, `v2::queue`. Those modules are `crate::views::*` now, so the paths were updated — a doc comment pointing at a deleted module is the exact thing this PR exists to remove.

### Evidence

- `cargo check -p stella-tui --all-targets` — clean.
- `cargo test -p stella-tui --no-fail-fast` — 1131 passed, 0 failed (plus the integration binaries: 6, 4, 3, 24, 9, 4, 5, 15 — all green).
- main's #4799 query-box tests survive the move and pass by name: `deck_ui::tests::graph::q_opens_the_query_box_and_enter_sends_the_query`, `esc_closes_the_query_box_without_asking_anything`, `backspace_empties_then_closes_the_query_box`, `an_empty_query_closes_rather_than_asking_for_nothing`, and the goldens `deck_render_snapshots_pin_the_graph_query_box` / `..._free_form_query`.
- `cargo fmt --all -- --check` — clean.

### CI will be red, and not because of this branch

`cargo check -p stella-cli` fails with 45 errors on `origin/main` itself. #4775's squash (`5bad8ddd7`) applied the additive half of its module split — the new `mod`/`use` lines and the five new sibling files — and dropped the deletions, so `command_deck.rs` now both imports and defines `handle_supervisor_msg`, `run_deck_command`, `run_lead_turn` and eight others.

That is inherited, not introduced here: across all of `crates/stella-cli/`, this branch differs from `origin/main` by five lines in `mid_turn_ask.rs` / `interactive.rs` (the namespace rename) and one `mod` line reordered by `rustfmt`. `command_deck.rs` and `command_deck/driver_support.rs` — the two files the errors are about — are byte-identical to main's.

Filed as #4807 with the full deletion list. This PR cannot go green until that lands.

macanderson:
## Status: correct as written, but blocked and racing the tree

**All four red checks are inherited, not this PR's.**

- `main is not known-broken` — main is red, tracked by #4810. The canary holds every PR until it closes; the guard's own text says to wait.
- `cargo check on the declared MSRV` and `fmt + clippy + test` — 45 compile errors in `crates/stella-cli/src/command_deck.rs`, a file this PR never touches. `cargo check -p stella-cli` is clean on this branch.
- `gate steps no other workflow runs` — `check-prose` failing on `command_deck/panel_snapshots.rs` and `command_deck/slash_commands.rs`, also untouched here. `check-prose.py` is green on this branch.

## The structural problem

Eleven commits landed on main in the ~90 minutes since this branch's base, **six of them in `stella-tui`**, every one inside the directory this PR renames — the GRAPH query box (#4799), the floating-card chrome port (#4778), the ISSUES typeahead (#4765), the AGENTS pickers (#4752), the SESSION palette (#4764).

`git merge-tree origin/main` against this branch now reports four conflicts, and one has teeth:

```
CONFLICT (modify/delete): crates/stella-tui/src/v2/graph.rs
  deleted in <this branch> and modified in origin/main
```

That is the silent-revert shape: a branch moves a file while main fixes it at the old path. Here git flags it rather than losing it, but the lesson stands — **a whole-tree rename cannot be carried across a main merging ~4 PRs/hour into the directory being renamed.**

## What I propose instead of merging

This change is a *deterministic function of the tree*, not a hand-edit: retire three modules, `git mv src/v2 src/views`, rewrite `v2::` → `views::`, follow three path-coupled guards, clear the prose debt the rename exposes. Regenerating it against a newer main takes about two minutes and — this is the point — **picks up main's changes automatically, with no opportunity for a silent revert.** Merging or rebasing has to resolve them by hand, four conflicts at a time, against a target that moves again while you do it.

So this branch should be **regenerated at merge time, not merged as it stands** — ideally in one quiet window right after #4810 closes.

Two things worth a maintainer's call, since neither is mine to make:

1. **The namespace name.** `views` is the crate's own word for these modules and the one `deck_render` dispatches into per `DeckTab`; flattening to `src/` collides on `graph`. Cheap to change — say the word.
2. **Whether this lands at all right now.** Six in-flight PRs are actively editing `src/v2/*`. Every one of them will conflict with the rename, and each of their authors pays that cost. A rename this broad wants either a quiet window or a deliberate "everyone rebase" moment. That is a scheduling decision, not an engineering one.

The content itself is verified on this branch: 1122 lib tests plus all nine integration suites green, whole-frame goldens byte-identical, `make guards-fast` / `fmt --check` / `clippy -p stella-tui --all-targets` clean, prose gate green with nine baseline entries removed and none added.

macanderson:
## The regeneration claim, tested

I claimed above that this change regenerates against a newer main "in about two minutes". I tested that rather than leaving it as an assertion, and **the estimate was wrong** — here is the honest version.

Branch `epic-4676-unify-views-current` (`9115055a1`) is this same change regenerated on `ce6b1cf3c`, eleven commits newer than this PR's base. It is green: 1131 lib tests (up from 1122 — main added some), all nine integration suites, `cargo fmt --check`, `clippy -p stella-tui --all-targets`, `check-file-size`. Use that branch, not this one.

**What regenerated for free.** The mechanical pass — retire three modules, `git mv`, rewrite `v2::` → `views::`, follow the guards — is a few seconds and picked up main's changes with no thought required.

**What did not.** Roughly eight block-level doc rewrites had to be re-applied by hand, and two of them had to be *re-derived* because main had changed the surrounding code:

- #4799 added a `typing: Option<&str>` parameter and a whole query-box mode to `graph_tab::render`, and taught the shim's empty state to tell "query matched nothing" from "no index here". The fold had to take **main's** version of that logic, not my base's.
- #4780 and #4778 added *new* v1/v2 vocabulary (`render/entry/tool.rs`, `views/cards.rs`) that did not exist when this PR was written, so the prose pass had more to clear than it did the first time.

So: call it 20–30 minutes of careful work, not two. Still far better than resolving four conflicts by hand against a moving target — and, importantly, **every miss was caught by a guard rather than shipped.** The hex-literal sweep failed with `no_hex_literals_in_v2_render_code ... FAILED` because its roots still pointed at `src/v2.rs`; that assertion exists precisely so a sweep over nothing cannot pass quietly. That is the check catching itself, which is the property worth having.

**One judgment applied consistently.** `deck_ui.rs` is 3591 lines on current main, and its baseline ceiling is 3675. On the older base my change accounted for a 14-line shrink and I retightened it. Here main had already shrunk it further on its own, so **I left the ceiling alone** rather than bank someone else's reduction — the same rule that keeps this PR's baseline edits confined to the nine entries the rename itself forces.

The two `check-prose` failures still red on CI (`command_deck/panel_snapshots.rs`, `command_deck/slash_commands.rs`) are main's, from #4775's `command_deck.rs` split creating files held to zero. They are part of what #4810 is tracking, and they are untouched by either branch.

macanderson:
Superseded by #4822 — the same change regenerated on `702383a49` (green main), which resolves this PR's four conflicts by construction rather than by hand.

A tree-wide rename is a deterministic function of the tree, so it is regenerated on the new base, never merged or rebased: regeneration picks up main's changes for free, where a hand-resolved merge is an opportunity to drop one. #4822 therefore incorporates #4799's GRAPH query box and its `typing` parameter, #4780's AGENTS-page caret and reply pane, and #4778's card-chrome port.

PR #4808, opened 2026-08-25
Semantic search does not miss the answer — the calibration harness measured recall at the wrong depth

### LINKED ISSUE #4784: Semantic search misses the labelled answer on half the calibration queries
## Problem

Running `relevance_calibration.rs` against `voyage-code-3` (PR #4774) surfaced a recall result, not just a threshold one: **two of four** calibration queries returned no labelled answer in the top 40 candidates. "Where is the prompt cache engaged on outbound requests" returned 40 irrelevant chunks scoring 0.608-0.680 and never surfaced `crates/stella-model/src/anthropic.rs`.

For another query the labelled answer ranked **39th at 0.5978**, beneath 38 irrelevant chunks at 0.6006-0.6436 — so the ordering, not only the cutoff, is wrong on this corpus.

## Done means

The cause is identified and named: chunking granularity, the embedding model's fit for this code, query phrasing, or the corpus. A fix or a recorded decision follows, with the calibration run repeatable (`STELLA_CALIBRATION_INDEX` makes it cheap against a filled index — the expensive part is re-embedding, not querying).

This matters more than a threshold: a floor cannot rescue a ranking that never contains the answer.

From PR #4774's report (2026-08-24). Related: #2993, #3096.

macanderson:
## What #4784 reported, and what is actually true

#4784 says semantic search misses the labelled answer on half the calibration
queries, and concludes "a floor cannot rescue a ranking that never contains the
answer."

The ranking contains every answer. Re-run against a copy of this workspace's own
index (`voyage-code-3@1/1024/l2`, 28 269 chunks):

```
recall: every labelled answer was retrieved; the deepest sat at rank 77 of 28269 (0.272% of the corpus)
  rank 77   where is the prompt cache engaged on outbound requests
  rank  1   how a policy or approval decision from the hook bus becomes a content-free record...
  rank 12   why is the Rust toolchain pinned to an exact version instead of tracking stable
  rank 62   how does a provider declare which features it supports and how is that checked
```

Worst case is the top 0.27% of a 28k-chunk corpus. Retrieval is not the problem
and neither is the ordering. **The cause is the harness that measured them**, in
two independent places.

## Cause 1 — `DEPTH` was the whole ranking, not a window onto it

`rank_chunks_by_vector(..., DEPTH)` returned 40 candidates and the harness had no
way to distinguish "not retrieved" from "not in the top 40". An answer at rank 62
or 77 produced `labelled_prefix == 0`, and the note printed for that condition
said, in the file's own words, "That is a recall result."

Rankings are now taken over the entire corpus and `DEPTH` slices the printed
window out of them. This costs **no embeddings** — the vectors are already local,
and `rank_chunks` performs the same full scan for any limit; only `top_k`'s
`truncate` changes. An answer below the window now has a rank.

## Cause 2 — the reported rank was the *deepest* incidental match

`measure` uses `rposition` to find the deepest labelled candidate. That is the
correct conservative reading **for a floor** — it is the cut that keeps every
answer — and the wrong one for recall.

The toolchain query's label was `("AGENTS.md", None)`, which admits every section
of AGENTS.md. So the reported "39th at 0.5978, beneath 38 irrelevant chunks" was
`AGENTS.md › Code style and conventions`, an incidental match. The section the
label's own `basis` names —

> the rationale appears in exactly one place in the tree — AGENTS.md's `## Essential commands`

— ranked **12th at 0.6217**. The label now pins that section, so it means what its
basis claims, and the recall rank is printed beside the deepest-relevant reading
instead of being replaced by it.

## Cause 3 — the harness could not run on the model it had measured

It panicked on `SimilarityPosture::Surface` with "there is no floor to calibrate".
`Surface` is a result *this harness produced* for `voyage-code-3` (#2993), so the
panic made the measurement unrepeatable for the one model it had measured — while
the ranking, the recall and the boundary gaps it also prints are exactly what a
`Surface` backend still needs measured. It now reports the absent floor and
measures everything else.

This is why the issue's own "Done means — with the calibration run repeatable"
could not be satisfied before this PR: the first thing the harness did on
`voyage-code-3` was abort.

## What did not change

**No constant's value moved.** Pinning the label moved the tightest separation
from -0.0439 to -0.0219; it stayed negative, so the standing result holds:
relevant and irrelevant scores overlap on this corpus (irrelevant 0.60-0.64,
answers 0.58-0.76) and no admission floor separates them. `voyage-code-3` keeps
`SimilarityPosture::Surface`.

The overlap is in the **scores**, not the ordering — which is the distinction the
original reading collapsed.

`MEASURED_FLOORS`, `DEFAULT_RELEVANCE_GAP_RATIO` and `DEFAULT_MIN_BOUNDARY_GAP`
carry the corrected numbers (boundary drops 0.0140 and 0.0003; the second
frontier's query ranked 12th, not 39th) and the corrected conclusion. The
harness's module doc has a `#4784` section recording what the first reading got
wrong, so the next reader re-checks rather than inherits it.

## Evidence

`VOYAGE_API_KEY` + `STELLA_CALIBRATION_INDEX` pointed at a **copy** of this
workspace's `.stella/private/codegraph.db`, per the module doc's instruction:

```
VOYAGE_API_KEY=pa-… STELLA_CALIBRATION_INDEX=/path/to/copy/codegraph.db \
    cargo test -p stella-tools --test relevance_calibration -- --ignored --nocapture
```

Baseline (before this PR) reproduced the issue's report; the run above is after.
The independent cross-check was a full-corpus rank of all 28 269 chunks per
query, which located every answer — that probe is what showed rank 12 where the
harness said 36, and it was deleted once the harness itself reported the rank.

`cargo test -p stella-embed` passes (26 tests). CI is the evidence for the rest.

## Note on the gate

`make guards-fast` is red on `main` right now, in files this PR does not touch:
`check-prose` on `crates/stella-cli/src/command_deck/{panel_snapshots,slash_commands}.rs`
and `cargo fmt` on `crates/stella-cli/src/command_deck.rs`. Those files are
unmodified here. Filed separately rather than composed onto this branch, since
`command_deck` is being actively rewritten by #4803 and a wording change there
would collide.

Closes #4784.

## Summary by Sourcery

Correct the relevance calibration harness so it measures recall across the full index while retaining window-based threshold analysis, enabling repeatable measurements for surface-similarity backends.

New Features:
- Report whole-corpus recall ranks separately from the configurable top-candidate calibration window.
- Allow calibration runs for surface-similarity backends without an admission floor.

Bug Fixes:
- Correct calibration recall reporting so answers below the printed window are not misclassified as retrieval misses.
- Pin the toolchain calibration label to the specifically referenced AGENTS.md section to avoid incidental matches affecting reported ranks and separations.

Enhancements:
- Clarify calibration documentation and recorded measurements to distinguish score overlap from retrieval ordering and preserve the existing threshold conclusions.

Documentation:
- Document the earlier calibration misinterpretation and the corrected recall, ranking, and boundary results.

Tests:
- Update relevance calibration output and measurements to include whole-corpus recall ranks alongside window-based separation and boundary metrics.


macanderson:
Why this PRs checks are red, and why none of it is this diff.

`main` is red at `ce6b1cf3c` and the repos own gate holds every PR behind it:

    FAIL main is red, and merging onto it is how one break becomes four.
         open `main-red` issue(s): 4810

Its instruction for a PR that is not fixing main is *wait* — the canary closes its own issue and the check goes green with no action here. All four failures trace to that:

| check | cause |
|---|---|
| main is not known-broken | the hold itself, #4810 |
| fmt + clippy + test | `stella-cli` compile, 45 errors in `command_deck.rs` (#4810), plus `cargo doc` on `stella-tools/src/gated.rs` (#4816) |
| cargo check on the declared MSRV | same `stella-cli` compile |
| gate steps no other workflow runs | downstream of the above |

Neither file is in this diff, which is three files: `crates/stella-embed/src/{http,rank}.rs` and `crates/stella-tools/tests/relevance_calibration.rs`. I checked the failure logs for my own crates rather than assuming — the only `stella-tools` entry is `gated.rs:267`, which this branch does not touch (`git diff origin/main -- crates/stella-tools/src/gated.rs` is empty) and which is filed as #4816.

Two independent main-red causes, so both must land before this goes green: #4812 fixes the compile half, #4816 is unclaimed.

Everything that could pass, did: docs guards, guard self-tests, file size ratchet, prose-only, Windows compile, cargo deny + audit, dependency-review, duplicate claims, harbor pytest.

Sourcerys assessment table posted three ✅ and no ❌. Its line-by-line review did not run — the accounts 7-day diff budget is exhausted — so this PR has **not** had a Sourcery code review, only the linked-issue assessment. Worth `@sourcery-ai review` later if that budget resets before merge.

Issue #4733, opened 2026-08-24
A plugin cannot ship an MCP server, so a vendor integration is the one thing a package cannot package


macanderson:
## Problem

`AGENTS.md`'s `.stella/` table names four ways a workspace extends the agent:
script tools, skills, context records, and **MCP servers**. #3380 wired the
first three into the plugin package (`<plugin_dir>/tools`, `skills`, `rules`)
and #3567's provenance half has now landed for the other two. A plugin still
cannot ship an MCP server, so the most natural packaging of a vendor
integration — "install this plugin, get its MCP tools" — is the one thing a
package cannot do.

This is not merely a missing directory:

- `crates/stella-cli/src/agent.rs::load_mcp_plan` reads exactly
  `<workspace>/.stella/mcp.toml` and returns `McpPlan::None` the moment that
  file is absent, so a plugin's servers would never be reached in a workspace
  that has none of its own.
- `McpPlan::Invalid(String)` carries a single reason, so with two sources one
  broken package would disable the workspace's own servers.
- `crates/stella-plugin/src/configure.rs`'s `REFUSED_SECTIONS` lists `"mcp"`
  with the reason *"it names external tool servers whose tools join the agent's
  surface, so the effect of one line is a set of tools the consent document
  never listed"*. That reason is the design brief: MCP is refused today because
  it is unenumerable at consent, and the fix is to make it enumerable — a
  `[[mcp]]` table declaring server names, rendered by `consent_text` and
  reconciled against the package the way `[[tools]]` already is.
- An MCP server's tools enter the registry as `mcp__<server>__<tool>`, and
  `GatedToolSet::with_tool_principals` (`crates/stella-tools/src/gated.rs`) is
  keyed by tool name and populated only from `Vec<CustomTool>`
  (`agent/tool_stack.rs::contributed_principals`). The map needs no change; the
  population does. Either those calls authorize as `Principal::Plugin`, or the
  reason they cannot is written down.

## Files

- `crates/stella-plugin/src/package.rs` — `ContributionKind`, `PackageListing`,
  `reconcile`.
- `crates/stella-plugin/src/manifest.rs` — the `tools`/`skills`/`records`
  vectors a `[[mcp]]` one joins (`deny_unknown_fields`, so an undeclared
  section fails to parse today).
- `crates/stella-plugin/src/consent.rs` — `package_contributions`, including
  its early return.
- `crates/stella-plugin/src/configure.rs` — the `"mcp"` refusal, to revisit
[...]

PR #4812, opened 2026-08-25

macanderson:
Rebuilt on `main`, and narrowed to one change.

This branch opened as a main-red fix, and #4814 landed an independent fix for
the same red first — so the commit this branch led with ("the split left every
moved item behind") is dropped: main compiles without it. The doc-link repair
that rode along is dropped too, because #4823 already owns that half and lands
byte-identical content in `skills.rs` and `gated.rs`. Two PRs writing the same
two files is how a tree gets a worse merge than either of them.

What is left is the change nothing else carries, previously #4817, which merged
into this branch rather than into main:

**The GRAPH q box answers through the CGP host, not CodeGraph directly.**
`graph_view` queried CodeGraph itself, so a workspace whose graph is served by
a CGP host got no answer — and an unanswered fan-out was indistinguishable from
an empty result. It now asks the host, and the two cases stay distinct.

Evidence, on this head:

- `cargo test -p stella-cli --bins graph_view` — 11 passed, 0 failed, including
  `a_query_is_answered_through_the_host` and
  `a_fanout_nobody_answered_is_not_an_empty_answer`.
- `cargo check -p stella-cli --all-targets` and `cargo fmt --all --check` clean.
- The gates that compile the workspace are CI's job on this head.

The pre-rebuild head is kept at `insurance/4812-preconflict` until this merges.

Note that main is red at a7a46d3b6 on `file-size` (#4821, claimed and being
split separately), so the `main is not known-broken` check here stays red until
that lands. It is not this branch.


macanderson:
## The timeline, and my own PR's part in it

Labelled `unblocks-main` per the hold check's own instruction. It names #4810,
which the post-merge canary filed at `ce6b1cf3c` — the same break, found
independently.

Three PRs merged inside 194 seconds, which is why nothing caught this between
them:

| when (UTC) | what |
|---|---|
| 2026-08-24 19:57:53 | previous `main-red` (#4691) closed — canary green from here |
| 2026-08-25 04:57:42 | **#4775** merges (`5bad8ddd7`) — the split |
| 2026-08-25 04:59:19 | #4798 merges, 97s later |
| 2026-08-25 05:00:56 | #4799 merges |
| 2026-08-25 05:13:54 | canary files #4810 |

**#4775 is the origin, and the evidence is in its own tree rather than in the
canary's timing.** At `5bad8ddd7`, `command_deck.rs` defines `enum DeckCommand`
at line 3113 *and* imports it at line 119 (`use slash_commands::{DeckCommand,
…}`), while `slash_commands.rs` defines it too. That is `E0255` on its face, with
nothing else required to trigger it.

**#4798 is mine, and it added a twelfth duplicate.** Its branch still had the
pre-split `command_deck.rs`, so when it merged `main` 97 seconds after the split
landed, git took my edited `spawn_mcp_connect` as a clean auto-merge and put it
back beside the sibling's copy. That is one of the 45 errors and I own it — the
other eleven, and the twelve definition-only collisions, predate it.

Worth naming rather than leaving implicit: the canary evaluates the *merged
tree*, so a break introduced at 04:57 and merged onto twice before 05:01 gets
one issue at 05:13 naming the last commit. The issue is right about the state
and says nothing about the origin, which is how "#4799 broke main" would have
been the natural and wrong conclusion.

macanderson:
Heads-up: **main has a second red cause this PR does not cover**, so the canary will not close #4810 when this lands.

`cargo doc -D warnings` fails on `stella-tools`, which this PR does not touch:

    error: public documentation for `principal_of` links to private item `Self::principal_for`
       --> crates/stella-tools/src/gated.rs:267:11
    error: could not document `stella-tools`

Filed as #4816 with the fix and the reasoning. It does not overlap this PRs files (`stella-cli`, `stella-tui`), so the two can land independently — they just both have to land before main is green.

Same site also carries two stacked doc paragraphs on one function, which reads like the same keep-both-sides merge that produced the `E0255` duplicates you are fixing here. Worth a glance while you are in the neighbourhood, in case #4775s split left more of them elsewhere.

Found while working #4784 (PR #4808), which is blocked behind the same main-red hold. Not claiming #4816 — say the word if you want it folded in here instead.

macanderson:
Closing: **`main` is repaired, and this branch would now revert work rather than
add any.**

`main` at `a7a46d3b6` compiles clean (`cargo check -p stella-cli --all-targets`
exit 0, from 45 errors), and the canary closed #4810 at 06:13:23Z, which is its
signal that the merged tree composes again. Every piece this PR carried is
present there:

| repair | on `main` |
|---|---|
| the 23 duplicated items gone from `command_deck.rs` | yes — no `spawn_mcp_connect` / `DeckCommand` / `run_lead_turn` definitions remain |
| `turn_files.rs` registries repointed at `command_deck/lead_turn.rs` | yes (4 entries) |
| `command_side.rs` importing `/info`'s vocabulary from `slash_commands` | yes |
| `driver_support.rs` carrying #4733's two-field `McpPlan::Servers` + notices | yes |
| the three private-item doc links | yes |
| the two prose violations | yes |

I did not land these — an equivalent repair reached `main` from elsewhere while
this PR's last build was running. Which one arrived first is not worth
establishing; the tree is correct either way.

Merging this now would be actively harmful. The branch is behind `a7a46d3b6`,
so its diff against `main` is 1158 insertions / 1122 deletions spanning
`stella-embed`, `relevance_calibration.rs`, `child_stream.rs` and
`wrapper_plugin.rs` — files this PR never touched. Those deletions are other
people's merged work being rolled back, which is exactly the failure mode where
a conflicting PR merges with its half reverted.

One thing to carry forward rather than lose with this PR: `f8a3d4afb` on this
branch is a **duplicate** of #4817's commit, pushed here while #4817 was landing
on `main` from its own branch (`fix/4335-graph-query-via-cgp-host`). It has a
single parent and is not an ancestor of `main`. Nothing is missing as a result —
#4817 merged at 06:14:22Z — but it is why this branch's diff looked stranger
than the repair it describes.

The analysis in the description and the timeline comment stand on their own and
are worth keeping for whoever looks at #4775 next.

macanderson:
Reopened the one piece this branch was still carrying, as #4826.

The close reasoning holds for the main-red repairs — #4814 landed them independently and main compiles without this branch. It does not hold for #4817, which merged *into* this branch at 06:14:22Z, three minutes before the close, and so went out with it rather than to main.

On a7a46d3b6, `crates/stella-cli/src/agent/graph_view.rs` is 199 lines and names the CGP host only in a doc comment, and `crates/stella-cli/src/agent/graph_view/tests.rs` is absent — `git cat-file -e origin/main:crates/stella-cli/src/agent/graph_view/tests.rs` returns not-found. The branch version routes the query through the host and carries 11 tests for it.

#4826 is that commit cherry-picked onto current main, nothing else. The doc-link repair this branch also carried is dropped there, because #4823 lands byte-identical content in `skills.rs` and `gated.rs` and already owns that half.

Pre-close head kept at `insurance/4812-preconflict`.

Issue #4807, opened 2026-08-25
stella-cli has not compiled since #4775: command_deck.rs both imports and defines the functions that moved to sibling modules


macanderson:
## The bug

`cargo check -p stella-cli` fails on `origin/main` (`ce6b1cf3c`) with **45 errors**. The crate has not compiled since #4775 merged as `5bad8ddd7`.

`crates/stella-cli/src/command_deck.rs` both **imports** the functions #4775 moved into sibling modules and still **defines** them inline:

```
crates/stella-cli/src/command_deck.rs:115
    use driver_support::{
        handle_supervisor_msg, service_registry_action, spawn_mcp_connect,
        spawn_notification_poller, spawn_pr_monitor,
    };
crates/stella-cli/src/command_deck.rs:2679
    fn handle_supervisor_msg(          // <- also defined here
```

```
error[E0255]: the name `handle_supervisor_msg` is defined multiple times
    --> crates/stella-cli/src/command_deck.rs:2679:1
 115 |   handle_supervisor_msg, service_registry_action, spawn_mcp_connect, spawn_notification_poller,
     |   --------------------- previous import of the value `handle_supervisor_msg` here
```

## How it got here

#4775 ("split command_deck.rs into sibling modules") describes moving ~1263 lines out of `command_deck.rs` — the PR body says the file "drops from 3737 to 2474 lines (34%)". The squash commit that landed (`5bad8ddd7`) changed `command_deck.rs` by only 110 lines. It applied the *additive* half — the four new `mod`/`use` lines, the pruned top-level imports, the five new sibling files — and dropped the *deletions*. The file is still 3700+ lines and still owns every function that also now lives in a sibling.

That produces two error families:

1. **11 × E0255** — a name is both imported and defined: `handle_supervisor_msg`, `service_registry_action`, `spawn_mcp_connect`, `spawn_notification_poller`, `spawn_pr_monitor`, `run_lead_turn`, `engine_config_inbound`, `tool_policy_inbound`, `DeckCommand`, `handle_agents_input`, `run_deck_command`.
[...]

Issue #4816, opened 2026-08-25
main is still red after #4812: cargo doc fails on stella-tools gated.rs


macanderson:
## Problem

`main` has a **second, independent** reason to be red, in a different crate from
the one #4810 reports. #4812 does not touch it, so main will still be red after
that PR lands.

`cargo doc -D warnings` fails on `stella-tools`:

```
error: public documentation for `principal_of` links to private item `Self::principal_for`
   --> crates/stella-tools/src/gated.rs:267:11
    |
267 |     /// [`Self::principal_for`] answers, exposed so a host can assert the
    |           ^^^^^^^^^^^^^^^^^^^^ this item is private
    |
    = note: this link resolves only because you passed `--document-private-items`, but will break without
    = note: `-D rustdoc::private-intra-doc-links` implied by `-D warnings`
error: could not document `stella-tools`
```

`principal_of` is `pub`; `principal_for` is private, so the intra-doc link is
only resolvable under `--document-private-items` and rustdoc rejects it under
`-D warnings`.

There is a second symptom at the same site that looks like the same cause. The
item carries **two** stacked doc paragraphs, the first orphaned:

```rust
/// The caller the gate is asked about for `name`: the plugin that
/// contributed the tool, or — for everything the user wrote themselves
/// and every built-in — this stack's own principal.
/// Who a call to `name` would be authorized as — the question
/// [`Self::principal_for`] answers, exposed so a host can assert the
/// answer instead of inferring it from a gate's behaviour.
#[must_use]
pub fn principal_of(&self, name: &str) -> Principal {
```

Two descriptions of the same function, run together with no break, reading as
one malformed sentence. That is the shape of a merge that kept both sides.

## Why this is filed separately from #4810

#4810 is the canary's report and names `compile` — 45 `E0255`-class errors in
`crates/stella-cli/src/command_deck.rs` from #4775's module split. PR #4812
(`fix/main-red-prose-fmt`) fixes that, and touches only `stella-cli` and
`stella-tui`.

This one is in `stella-tools`, is a rustdoc failure rather than a compile
failure, and survives #4812 untouched. Two red causes for one red main deadlock
each other's repairs if only one is known about, which is why it is its own
issue rather than a comment.

## Verify

```
cargo doc --workspace --no-deps --document-private-items   # with -D warnings
```

Or read the failing `ci` job on any current PR — e.g.
https://github.com/macanderson/stella/actions/runs/32811952604/job/97694133939,
[...]

PR #4817, opened 2026-08-25

macanderson:
Closes the one ❌ Sourcery raised against #4799's linked issue #4335:

> Implement a graph free-form query mode opened with `q` … The PR implements the query input and `WorkspaceInput::GraphQuery`, but resolves queries directly through `CodeGraph::definition_spans` rather than answering them through the CGP host as required by the issue.

#4799 has since merged (`ce6b1cf3c`) and #4335 auto-closed as completed, so the objective was never delivered. This is the follow-up.

## Correcting the record on #4799

A comment on #4799 answered that row rather than fixing it, arguing the host could not carry a graph. Three of its claims do not survive reading the pinned `contextgraph-types 0.1.2`:

| Claim on #4799 | What the code says |
| --- | --- |
| "no node kind … on a frame" | `FrameKind::Symbol` and `FrameKind::Graph`, `frame.rs:25-33` |
| "no edge list on a frame" | `ContextFrame.relations: Vec<Relation>`, `frame.rs:319`; `Relation { rel, target_uri, display_name }`, `frame.rs:190` |
| "needs a structured node/edge frame kind in the protocol first" | `stella-graph/src/frames.rs:485` already emits `FrameKind::Graph` frames carrying those relations, and SPEC §G1/§G2 are written for graph-capable providers |

One claim does hold, and this PR pays it rather than denying it: the frame route loses fidelity the direct read kept. That cost is documented at `node_kind` and `frame_location` and pinned by tests.

## What changed

`graph_query_snapshot` sends a `ContextQuery` through a `contextgraph_host::Host` and assembles the neighborhood from the returned `ContextFrame`s, naming no `stella-graph` type. Two round trips: `kinds = [Symbol]` locates the root file, then `anchors = [that file]` with `kinds = [Symbol, Graph]` returns the neighborhood, because `Graph` frames carry the `Relation`s that become edges.

`graph_tab_host` registers only `code-graph` — the session host also advertises `symbol` from `workspace-memory`, which would answer a structural query with reflections the tab has no node to draw.

### Fidelity accepted, and stated

- `SymbolKind::keyword` projects both `function` and `method` onto `fn`, so a method draws as a function. One glyph in a browsing view; recovering the tag would mean re-reading the index the query just went through, or widening a frame `id` the protocol calls stable for dedup.
- A symbol's line rides `provenance.range` as `"L12-40"` rather than as a number, so it is parsed back out.

## Two defects the witness caught

Neither pre-existed this PR; both were introduced by routing through the host and found by the tests, not by inspection.

1. **Symlinked workspaces lost every node's file and line.** Frame URIs are minted against the index's root and `CodeGraph::open` canonicalizes it, so the deck's `/var/...` spelling never matched the frames' `/private/var/...`. That is every macOS `/tmp` and `/var` workspace. `canonical_root` resolves both sides once.
2. **A wedged provider read as "your query found nothing".** A fan-out where every leg timed out, was refused, or crashed is indistinguishable from "no matches" at the frame layer. `query_frames_via_host` now returns `None` for that case and the tab keeps the neighborhood it had. The 750ms tab timeout that surfaced this is gone — the code-graph provider's first query pays for a grammar load — so the tab uses the same bound recall does.

## Evidence

- 11 new tests in `crates/stella-cli/src/agent/graph_view/tests.rs`: mapping units plus a round trip against a real index behind a real `Host`, covering the `IMPORTED_BY` edge reversal, unresolved imports, and the nobody-answered case.
- `cargo test -p stella-cli --bin stella --no-fail-fast` — **2297 passed, 0 failed**. Both defects above reproduced only under full-suite parallel load; they are fixed, not filtered.
- `make guards-fast` exit 0; `make prose` clean; `cargo clippy -p stella-cli --bin stella --all-targets` silent; `cargo fmt --check` clean.

## Stacked on #4812

`main` has not compiled since #4775 (`E0255`, 45 errors on a clean checkout — issues #4807, #4810). #4812 fixes it. This PR is based on that branch so its diff is only the five files it touches and CI runs on a tree that compiles. Retarget to `main` once #4812 merges.

## Not in scope

The picker's file re-root still reads `CodeGraph` directly: it names a file and wants that file's stored neighborhood, and `GraphSnapshot.files` is the picker's inventory rather than a query answer. Moving it onto frames would need the protocol to carry a node's kind and span, which is a `context-graph-protocol` change — filed separately.

## Summary by Sourcery

Route GRAPH tab free-form queries through the Context Graph Protocol host while preserving graph rendering, workspace navigation, and reliable empty-answer behavior.

New Features:
- Route GRAPH tab free-form queries through a dedicated Context Graph Protocol host and render returned symbol and graph frames as a navigable neighborhood.
- Preserve indexed-file inventory for graph file selection while using host-backed query results for the `q` box.

Bug Fixes:
- Preserve node file locations for symlinked workspaces by normalizing workspace roots.
- Keep the existing graph neighborhood when no provider answers instead of reporting a false empty query result.
- Correct imported-by edge direction and retain unresolved relation targets in rendered graphs.

Enhancements:
- Translate protocol frames into graph nodes and edges with supported kinds, locations, relation labels, and accepted documented fidelity tradeoffs.
- Use the host's standard recall timeout and asynchronous query path for GRAPH interactions.

Tests:
- Add mapping and end-to-end coverage for frame conversion, provenance locations, relation direction, unresolved imports, host-backed queries, empty answers, unavailable providers, and unindexed workspaces.

### DIFF
--- a/crates/stella-cli/src/agent/graph_view.rs
+++ b/crates/stella-cli/src/agent/graph_view.rs
@@ -13,7 +13,26 @@
 //! ([`graph_snapshot_focus`]) or a free-form query ([`graph_query_snapshot`])
 //! — and both report the wall clock they spent, because the deck cannot
 //! measure a query it did not run (#4335).
+//!
+//! The two reach the index by different routes, on purpose. The picker names
+//! a **file** and wants that file's stored neighborhood verbatim, so it reads
+//! [`stella_graph::CodeGraph`] directly. The `q` box asks a **question**, and
+//! a question is what the Context Graph Protocol exists to route: it goes out
+//! as a [`ContextQuery`] through a [`contextgraph_host::Host`] and comes back
+//! as [`ContextFrame`]s, so the tab names no `stella-graph` type and a
+//! graph-capable provider registered on that host later answers the `q` box
+//! with no change here (#4335).
+//!
+//! Going through frames costs fidelity the direct path keeps, and the cost is
+//! paid where [`node_kind`] and [`frame_location`] are written: a frame
+//! carries the citation keyword (`fn`), not the index's stored tag
+//! (`function` / `method`), and carries a symbol's line inside
+//! `provenance.range` as the string `"L12-40"` rather than as a number.
+//! Both are recoverable, neither is free, and a reader deciding whether to
+//! move the picker onto this path should read those two functions first.
 
+use contextgraph_host::Host;
+use contextgraph_types::{ContextFrame, ContextQuery, FrameKind, Relation};
 use stella_tui::{GraphEdge, GraphNode, GraphSnapshot};
 
 /// Query the code graph (if `stella init` has built it) for the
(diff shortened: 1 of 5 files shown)


macanderson:
## The one red check is inherited, not this diff

`fmt + clippy + test` is red on head `f68171c2d`. Its step breakdown ([run 32813449572](https://github.com/macanderson/stella/actions/runs/32813449572/job/97699070195)) puts the failure in exactly one step:

| Step | Result |
| --- | --- |
| `cargo fmt --check` | ✅ |
| **`cargo doc -D warnings`** | ❌ |
| `cargo test` | ✅ |
| `cargo build` (release smoke, thin LTO) | ✅ |
| every toolchain-free guard (30 steps) | ✅ |

The workspace clippy and the **whole workspace test suite** pass on this branch. The failing step is `rustdoc::private_intra_doc_links` in two files:

- `crates/stella-core/src/skills.rs:96` and `:374` — `[`default_origin_for`]`
- `crates/stella-tools/src/gated.rs:267` — `[`Self::principal_for`]`

This PR touches neither crate. `git diff --name-only origin/main..HEAD` is nine `stella-cli` files plus `stella-tui/src/v2/session.rs` (the latter from the #4812 commit this branch is stacked on), and all three doc links are present verbatim on `origin/main` today. That break is #4816, already filed.

It was invisible until now because the `E0255` compile failure (#4807/#4810) aborted the job before `cargo doc` ran — one red gate step masking the next. So `main` has two independent red causes, and this PR inherits the second one through its base.

**Nothing to do here.** This goes green once #4816 lands; no change to this diff would move it.

`main is not known-broken` is red for the same reason and by design — it is the hold that fires while a `main-red` issue is open.

Issue #3802, opened 2026-08-18
A plugin's extra model calls are never recorded in the database, because they happen between two tracked turns


macanderson:
## Problem

The `stella run --pipeline <variant>` path now installs a `ChildTurns` plane
over the session's own dispatcher, so a plugin declaring
`[loop] calls = ["child_turn"]` gets a real, host-made model call
(`crates/stella-cli/src/wrapper_plugin.rs::session_host`). **That call reaches
no telemetry.**

`SessionSubAgents::dispatch` reads the *current* turn's event channel off the
registry and falls back to a sink when there is none:

```rust
// crates/stella-cli/src/subagent.rs
let events = tools.events().unwrap_or_else(|| EventSender::from_fn(|_| Ok(())));
```

That slot is attached per turn (`ToolRegistry::attach_events`, called from
`crates/stella-cli/src/agent.rs`) and released at the end of it
(`ToolRegistry::detach_event_stream`). A wrapper's points run **between** the
turns they are about — `before_turn` before `crate::agent::run_turn` is called
at all, `after_turn` after it returned — so at the moment a plugin's
`child_turn` is dispatched the slot is empty and every `StepUsage` the child
emits goes to the sink. Nothing lands in `executions`, `step_usage`, or the
context receipts, so `stella stats`/`stella usage report`/the Observatory all
under-report a real spend.

What *does* survive today, and is what the driver leans on meanwhile:

- the session's budget guard — the dispatcher pushes onto the registry's
  `SubAgentSpendLedger`, an engine turn drains it at its next step boundary,
  and `wrapper_plugin::settle_plugin_child_spend` folds whatever the last point
  spent after the dispatch ends;
- the run's own report — `ChildTurns::spends()` is printed beside
  `HostCallGate::refusals()` (`wrapper_plugin::spend_lines`).

So the money is bounded and the user is told on stderr; the *durable* record is
what is missing.

## Files

- `crates/stella-cli/src/subagent.rs` — `SessionSubAgents::dispatch`, the sink
  fallback.
- `crates/stella-tools/src/registry.rs` — `attach_events` /
  `detach_event_stream`, the per-turn slot.
- `crates/stella-cli/src/agent.rs` — where a turn attaches and releases it.
- `crates/stella-cli/src/wrapper_plugin.rs` — `run_wrapped`, the sequence that
  brackets the points, and `RawTurnDriver`, which owns one turn each.
[...]

PR #4819, opened 2026-08-25
fix(stella-cli): a plugin's child turn reaches the store on stella goal and stella fleet

### LINKED ISSUE #4730: A plugin's child turn is still invisible to the store on stella goal and stella fleet
## Problem

#3802 gave `stella run --pipeline <variant>` a between-rounds event stream, so a
plugin's `child_turn` lands a `step_usage` row instead of metering into a sink
(`crates/stella-cli/src/wrapper_plugin/child_stream.rs`). The two other doors
that drive a wrapper still have the original gap:

- **`stella goal --pipeline <variant>`** — `crates/stella-cli/src/agent/goal/goal_wrapped.rs`,
  `GoalRoundDriver::run_turn`.
- **`stella fleet`** — `crates/stella-cli/src/fleet_cmd.rs`, the attempt driver.

Both reach `child_turn` through `wrapper_plugin::round_driver_host` (#3833,
#3882), and both run their points between rounds, where
`ToolRegistry::events()` is empty and `SessionSubAgents::dispatch` falls back to
`EventSender::from_fn(|_| Ok(()))`. So the money is bounded (the spend ledger)
and reported on stderr, and nothing durable is written: `stella stats`,
`stella usage report` and the Observatory under-report a real spend on both
doors.

`crates/stella-runtime/README.md`'s child-turn paragraph states this gap out
loud today and must be corrected in the same PR that closes it.

## Why it is not the same edit

`stella run`'s fix opens **one run-scoped execution row** for the whole
dispatch, which is honest there because each round opens its own row and a
between-rounds child belongs to neither. Both doors here already run several
rounds under **one** execution row (that is what `ChildTurns::in_turn_lane`
exists for — `stella_core::turn_slots`), so the right answer may instead be to
keep the door's existing row and re-publish its stream across the points. That
is a decision, not a copy.

`goal_wrapped.rs` is also already on `turn_files.rs`'s `STREAM_OWNERS` list, so
whatever it does has to keep going through `turn_files::open_turn_streams`.

## Verify

`crates/stella-cli/src/wrapper_plugin/tests/child_stream.rs`'s
`a_plugins_child_turn_meters_into_a_stream_rather_than_a_sink` is the witness
shape: a metering dispatcher that reads `ToolRegistry::events()` at dispatch
time exactly as the shipping one does, a plugin asking for a child at both
points, and an assertion that both found a stream. Point it at each door's own
driver.

## Definition of done

- A plugin's `child_turn` on `stella goal --pipeline <variant>` and on
  `stella fleet` lands a `step_usage` row joinable to the run it belongs to.
- A witness per door, of the shape above.
- `crates/stella-runtime/README.md`'s child-turn paragraph stops naming the two
  doors as the remaining gap.

Refs #3802, #3833, #3882, #3576.

macanderson:
Closes #4730.

`stella run --pipeline <variant>` got a between-rounds event stream in #3802, so
a plugin's `child_turn` lands a `step_usage` row instead of metering into a
sink. The two other doors that drive a wrapper still had the gap. #4798 closed
this issue by link without touching either of them — Sourcery's assessment table
on that PR marked all three objectives `❌`, and it was right.

## The decision this asked for, and the answer

The issue said the shape was a decision rather than a copy, and it is:

- **`stella run` opens a run-scoped `plugin` execution row of its own**, because
  each of its rounds opens a row, so a between-rounds child belongs to neither.
- **`stella goal --pipeline <variant>` and `stella fleet` keep the row they
  already have.** Both already run every round under **one** execution row and
  one channel that outlives the dispatch, so a between-rounds child there is
  exactly as run-scoped as the rounds beside it, and lands joinable to them
  through the row they already share. A second `plugin` row would split one
  run's spend across two rows and put a phantom execution beside every wrapped
  run in `stella stats`.

The shared half is `RepublishingDriver`, which its own doc comment already said
was a decorator so that "`crate::agent::goal`'s round driver and
`crate::fleet_cmd`'s attempt driver" could use it. It is now generalised over a
`PointStream` trait and all three doors drive through it; what each door
publishes is its own impl.

**`stella fleet` publishes the registry's event slot alone**, deliberately not
the per-call work-tree measurement beside it. Its worker rebinds
`cfg.workspace_root` to its own worktree while the shared `SessionDurability`
journal stays rooted at the lead's, so a measurer attached there would snapshot
the wrong tree — which is why `turn_files`' `ENGINE_DRIVERS` records that door as
`Blocked` on #3233 and `STREAM_OWNERS` omits it. Publishing the measurer would
have silently un-blocked #3233 rather than closed this. `AttemptPointStream` is
what says so out loud.

`goal_wrapped.rs` stays on `STREAM_OWNERS` and still reaches the seam through
`turn_files::open_turn_streams`, via `attach_run_streams`.

## Evidence

A witness per door, of the shape the issue named — a metering dispatcher reading
`ToolRegistry::events()` at dispatch time exactly as the shipping one does, a
plugin asking for a child at both points, an assertion that both found a stream.
The plugin, dispatcher and manifest moved into one shared fixture
(`wrapper_plugin/child_turn_fixture.rs`) so a door cannot pass against a plugin
the other doors never ask anything of.

**Both were falsified before being kept** — a green that cannot fail is not
evidence:

| what was broken | what the witness said |
| --- | --- |
| goal: publication removed | `left: [false, true]` |
| goal: decorator removed | `after_turn`'s child metered against the round's own sender |
| fleet: publication removed | `left: [false, true]` |

**The goal witness caught a real defect in my own first commit.** The door's
point stream held a sender over the run's channel past the renderer await, so a
finished goal run hung on a channel that never closed (#960). Both doors now
drop the binding explicitly, and both witnesses assert the teardown.

`crates/stella-runtime/README.md`'s child-turn paragraph no longer names the two
doors as the remaining gap, and `turn_files`' `STREAM_OWNERS` note now names both
live raw `attach_events` calls rather than one.

## CI will be red, and not because of this branch

`origin/main` does not compile, from #4799 (`ce6b1cf3c`). That PR extracted
`command_deck.rs` into submodules but left the original bodies in place: 11
symbols are defined twice (`use lead_turn::run_lead_turn;` at line 118 and
`async fn run_lead_turn` at line 3593, and ten more), 45 errors. It also left
`command_deck/driver_support.rs` as a pre-#4798 copy that pattern-matches
`McpPlan::Servers(_)` against a now-2-field variant and silently drops #4798's
plugin-MCP notices, and left `command_deck/lead_turn.rs` in no `ENGINE_DRIVERS`
row. Filed separately; see the issue linked below.

Every verification above was run on a locally patched base with that duplication
removed — that patch is **not** in this branch. On it: `cargo check -p stella-cli
--bins` clean, `cargo test -p stella-cli --no-fail-fast` = 2285 passed / 3
failed, the three being the `turn_files` guards that point at the
`command_deck.rs` I had gutted; with those rows excluded, 11/12 pass and the
twelfth is main's own stale `lead_turn.rs` row.

## Summary by Sourcery

Route plugin child-turn metering through the owning goal or fleet stream and verify correct attribution across wrapper execution paths.

Bug Fixes:
- Ensure plugin child-turn usage from `stella goal --pipeline <variant>` and `stella fleet` reaches the appropriate run or attempt event stream and is persisted for reporting.

Enhancements:
- Generalize wrapper point-stream publication and reuse a shared republishing driver across all wrapper execution paths.
- Keep fleet publication limited to registry events while preserving its existing work-tree measurement status.

Documentation:
- Update runtime and stream-ownership documentation to describe child-turn handling for all wrapper execution paths.

Tests:
- Add shared child-turn metering fixtures and door-specific witnesses for goal and fleet, including stream republishing and teardown behavior.


macanderson:
Blocked on #4820 for CI: `origin/main` does not compile (45 errors, all in `command_deck.rs`, from #4799's incomplete extraction), so every check on this PR is red for a reason that is not this branch. The verification for this branch is in the PR description — it ran on a locally patched base whose patch is not in this branch.

macanderson:
Correction to this PR's description, now that it has merged.

The "CI will be red, and not because of this branch" section blamed **#4799**
(`ce6b1cf3c`) for main not compiling. That attribution is wrong. The duplication
predates it — at `ce6b1cf3c~1` both `use lead_turn::run_lead_turn;` and the
inline `async fn run_lead_turn` are already present:

```
git show ce6b1cf3c~1:crates/stella-cli/src/command_deck.rs \
  | rg -c 'use lead_turn::run_lead_turn|^async fn run_lead_turn'
```

The incomplete split was **#4775**. #4799 was merely the most recent commit to
touch the file when I looked, and I took that for authorship without running the
parent check. #4814 fixed it, and #4820 carries the full correction.

What that section got right, and what still stands: main genuinely did not
compile when this branch was pushed — CI on `ce6b1cf3c` now reads
`completed/failure` — so the verification here did run on a locally patched base,
and that patch was not in this branch.

**That caveat is now discharged.** On `a7a46d3b6`, which contains both #4814 and
this PR, against the real unpatched tree:

- `cargo check -p stella-cli --bins` — clean.
- `cargo test -p stella-cli --no-fail-fast` — **2288 passed, 0 failed**, both
  witnesses from this PR included:
  - `agent::goal::goal_wrapped::tests::a_plugins_child_turn_reaches_the_goal_runs_stream_at_both_points`
  - `fleet_cmd::wrapped::tests::a_plugins_child_turn_reaches_the_attempts_stream_at_both_points`

PR #4822, opened 2026-08-25
The terminal UI has one rendering generation: finish the port, delete v1, dissolve the namespace

### LINKED ISSUE #4676: Epic: the terminal UI has one rendering generation, not two — finish the v2 port and delete v1

macanderson:
Closes #4676. Replaces #4803 (same change, regenerated on green main — see the foot).

## Where the epic actually stood

Written against a `main` where `views/` held eighteen modules. By the time this work started it held three, and none of them still rendered the pre-SPEC-5 way:

| file | what it actually was |
| --- | --- |
| `views/graph.rs` | an empty state and a cursor clamp wrapping a call to `v2::graph_tab` |
| `views/settings.rs` | a one-line nav; all three panes already dispatched into `v2::*` |
| `views/cards.rs` | the floating-card chrome the `v2` tree imported eleven times |

So the remaining substance was items 2–6 of the definition of done, not item 1. `views/engine.rs` — one of the two god files item 5 names — had already left the file-size baseline.

## What landed

**The last three modules are gone.** `graph_tab` absorbed its shim's empty state and cursor clamp (its inner draw is now `paint`); `settings.rs` and `cards.rs` moved as they stood; `spinner_glyph` joined `status_glyph` in `theme`, the module that exists so every view takes its glyphs from one place.

**`src/v2/` is `src/views/`.** Flattening to `src/` was rejected: it collides on `graph` and leaves sixty-odd files in one directory with nothing saying which of them draw a surface. `views` is the crate's own word for these modules and the one `deck_render` dispatches into per `DeckTab`.

**The prose split two ways, and this is the part worth reviewing.**

Comparisons that existed only to justify a deletion already made are *deleted*: `frame.rs`'s account of the eleven rows the old deck spent, `status_bar.rs`'s case for one row over two and its stale "not wired yet" note, the README's statline row, `status_source.rs`'s cell-for-cell equivalence claim.

But `src/render/` **genuinely still has two row renderers**, and item 4 taken literally would have made its docs lie. `projected_rows` (was `v2_rows`) routes a tool call's head and compaction's line through the shared projection; everything else falls through to `entry_body`. That is live structure. So there the generation words became what the two arms actually are, and the transcript-surfaces ledger's `deck-v1`/`deck-v2` became `deck-body`/`deck-spec6`. The unfinished half — SPEC 6.4's result row, which waits on a highlighter — stays tracked by #4289 and #3578.

## Three guards had to follow the move

Each would have failed as a *stale* check rather than a wrong one: the hex-literal sweep's roots (`src/v2.rs`, `src/v2`), the transcript-surfaces ledger, and `keymap.rs`'s key-witness `include_str!`. The last is `#[cfg(test)]`, so `cargo check --lib` stays green while `cargo test` cannot build.

The sweep is the one that would have caught itself — it asserts it looked at something, and during one regeneration it did exactly that, failing with `no_hex_literals_in_v2_render_code ... FAILED` when its roots still pointed at the old tree.

## Evidence

Verified on this branch against `702383a49`:

- `cargo test -p stella-tui --no-fail-fast`: 1131 lib tests + all nine integration suites green.
- **The deck's whole-frame goldens are byte-identical** — no re-blessing needed, which is the evidence the rename moved no pixels.
- `make guards-fast` green (13/13), `cargo fmt --all --check` clean, `cargo clippy -p stella-tui --all-targets` clean, `cargo check -p stella-cli` clean.
- `check-prose` green: **nine baseline entries removed, none added.**

## Baselines

The prose baseline loses entries only for renamed files whose debt had to be paid rather than carried (a renamed path is a new file, held to zero) — 38 lines across 24 file/category groups, deleted rather than re-based, since #4763's filler-adverb rule widened what counts.

`deck_ui.rs`'s file-size ceiling is **left alone** at 3675 even though the file is 3591: main shrank it independently, and banking someone else's reduction is not this PR's to do. On an earlier base, where the shrink was attributable to this change, it was retightened — same rule, opposite outcome, because the facts differed.

## Test renames, for the deleted-tests guard

`tests/v2_status_bar.rs` → `tests/status_bar_goldens.rs`, three insta snapshots renamed to match, no golden content changed. **No test was deleted** — every one below is a rename, and the guard keys on the bare function name, so each old name is spelled out in full here rather than abbreviated:

- `v2_status_bar_at_deck_width` → `status_bar_at_deck_width`
- `v2_status_bar_at_the_extremes` → `status_bar_at_the_extremes`
- `v2_status_bar_narrow` → `status_bar_narrow`
- `no_hex_literals_in_v2_render_code` → `no_hex_literals_in_render_code`

`views/graph.rs`'s six tests moved to `views/graph_tab/tab.rs` under their own names, so they are not reported.

## Why this replaces #4803

#4803 was cut from `77df4b9a2` and went stale in about ninety minutes: eleven commits landed, six of them in `stella-tui`, every one inside the directory being renamed. `git merge-tree` reported four conflicts including `CONFLICT (modify/delete): src/v2/graph.rs — deleted in branch, modified in origin/main`, which is the silent-revert shape.

A tree-wide rename is a deterministic function of the tree, so it is **regenerated on the new base, never merged or rebased** — regeneration picks up main's changes for free, with no opportunity to drop one by hand. This branch is that regeneration on `702383a49`, and it incorporates main's work rather than fighting it: #4799's GRAPH query box and its `typing` parameter, #4780's AGENTS-page caret and reply pane, #4778's card-chrome port.


macanderson:
### The three reds here are inherited, and #4827 fixes all three

They ran at 06:2x against a `main` that was broken three ways. Nothing on this branch causes them:

| red here | actual cause | fixed by |
| --- | --- | --- |
| `fmt + clippy + test` | `explicit_auto_deref` on `&*registry` in `fleet_cmd.rs` — reproduces on a branch cut straight from main | #4827 |
| `gate steps no other workflow runs` | `check-prose` on `command_deck/panel_snapshots.rs`, `command_deck/slash_commands.rs`, `fleet_cmd/wrapped.rs`, `stella-runtime/README.md` — none touched here | #4827 (cherry-picked from #4823) |
| `main is not known-broken` | the canary's own hold while main was red | #4827 |

**#4827 is now fully green — 14 checks, none failing.** Merging it greens main; re-running this PR's checks after that should clear all three.

That is not a guess. I composed `origin/main` + #4827 + this branch locally — they merge cleanly, including the `scripts/prose-baseline.txt` that #4827 and this PR both touch — and ran every gate on the result:

- `check-prose` OK, none added · `check-file-size` OK · `cargo fmt --all --check` clean
- `cargo clippy -p stella-tui -p stella-cli --all-targets -- -D warnings` clean
- `stella-tui`: 1131 lib + all nine integration suites · `stella-cli`: 2288 tests · `make guards-fast` 13/13

And the evidence specific to a rename: the deck film rendered from that merged tree is **byte-identical** to the one rendered before it — 4123 lines, `cmp` clean. 4122 frames of real `render_deck` output agreeing exactly, which covers considerably more than the golden-frame suite's fixed set.

PR #4823, opened 2026-08-25
docs: main is red on the new prose pattern, and #4775's split left two public docs pointing at private items

macanderson:
Two reds on `main`, both in prose the compiler does not read.

## 1. `check-prose.py` fails on main

`check-prose.py` gained a `filler-adverb` pattern and `main` went red on six hits it had never been checked for — `guard self-tests` runs the script directly, so this reds every PR too:

```
crates/stella-cli/src/agent/goal/goal_wrapped/tests.rs [filler-adverb]: 0 allowed, 1 found
crates/stella-cli/src/fleet_cmd/wrapped.rs            [filler-adverb]: 0 allowed, 2 found
crates/stella-runtime/README.md                       [filler-adverb]: 2 allowed, 3 found
```

The guard's own message is "Fix the prose. Do not add a baseline entry." In every case the adverb sat in front of a clause that already gave the reason — "and that is the instrument the assertion turns on", "because a fleet worker rebinds `cfg.workspace_root`" — so it announced that a choice was made and the next clause said why. The README's "Two things are deliberately still absent." also announced a list instead of writing it; both items read the same without it.

`scripts/prose-baseline.txt` is retightened via `make prose-update`, which cannot raise a count.

## 2. Two public docs name a private item

#4775's split landed through #4814. The doc repairs that had ridden with it in #4812 did not — #4812 was closed unmerged, so these were dropped on the floor:

- **`gated.rs`**: `principal_of` carried two fused doc comments, the second starting mid-clause ("Who a call to `name` would be authorized as — the question / What `principal_for` answers"). The first sentence describes `principal_for` and is moved onto it; `principal_of` now says the one thing a caller needs — it hands back an owned `Principal`, so a host can assert the answer instead of inferring it from how the gate behaved.
- **`skills.rs`**: two public doc comments named `default_origin_for`, a private fn, pointing a reader of the public API somewhere they cannot follow. Both now say "the path-derived default", which is what the sentences were about anyway.

This is the same class \#4814's CI caught as `cargo doc -D warnings` errors on `stella-tools` and `stella-core`.

## Evidence

- `RUSTDOCFLAGS='-D warnings' cargo doc --no-deps -p stella-tools -p stella-core` — clean.
- `python3 ./scripts/check-prose.py` — `OK -- 6438 grandfathered construction(s) in 1473 file(s), none added.` (was FAIL).
- `cargo fmt --check` — clean.

## Summary by Sourcery

Clean up prose violations and repair public documentation references left behind by the gated-tool and skill-origin split.

Bug Fixes:
- Fix prose validation failures by removing unnecessary filler adverbs and list-announcing language from CLI tests, fleet documentation, and the runtime README.
- Restore public documentation that no longer references private implementation items and clarify the ownership semantics of `principal_of`.

Documentation:
- Correct public API documentation in `gated.rs` and `skills.rs` so it describes accessible concepts and methods without exposing private names.

Chores:
- Retighten the prose baseline without increasing allowed construction counts.

### DIFF
--- a/crates/stella-cli/src/agent/goal/goal_wrapped/tests.rs
+++ b/crates/stella-cli/src/agent/goal/goal_wrapped/tests.rs
@@ -29,7 +29,7 @@ use crate::wrapper_plugin::child_turn_fixture::{asking_plugin, step_usage_count}
 /// rounds, so after a round the slot holds *that round's* sender rather than
 /// nothing.
 ///
-/// **The stand-in deliberately does not forward to the run's channel**, and
+/// **The stand-in does not forward to the run's channel**, and
 /// that is the instrument the assertion turns on. The real per-turn sender is a
 /// `TurnFacts` tap over a clone of the run's channel, so a child that metered
 /// into it would still reach the drain and be indistinguishable from one that
--- a/crates/stella-cli/src/fleet_cmd/wrapped.rs
+++ b/crates/stella-cli/src/fleet_cmd/wrapped.rs
@@ -122,13 +122,13 @@ mod tests;
 /// they already share.
 ///
 /// **Events only, and that is the difference from every other door.** This
-/// publishes the registry's own event slot and deliberately not the per-call
+/// publishes the registry's own event slot and not the per-call
 /// work-tree measurement beside it (`turn_files::open_turn_streams`), because a
 /// fleet worker rebinds `cfg.workspace_root` to its own worktree while the
 /// shared `SessionDurability` journal stays rooted at the lead's — a measurer
 /// attached here would snapshot the wrong tree, which is exactly why
 /// `turn_files`' `ENGINE_DRIVERS` records this door as `Blocked` on #3233 and
-/// why `STREAM_OWNERS` deliberately omits it. Publishing the measurer here
+/// why `STREAM_OWNERS` omits it. Publishing the measurer here
 /// would silently un-block #3233 rather than close #4730, so it stays withheld
 /// and this type is what says so out loud.
 pub(super) struct AttemptPointStream {
--- a/crates/stella-core/src/skills.rs
+++ b/crates/stella-core/src/skills.rs
@@ -93,7 +93,7 @@ pub enum SkillOrigin {
     ///
     /// Separate from the other four because they all answer "which of the
     /// user's directories is this under", and a package's directory is none of
-    /// them. Without it `default_origin_for` (private) has no true answer and falls
+    /// them. Without it the path-derived default has no true answer and falls
     /// through to [`SkillOrigin::Workspace`], reporting a third party's skill
     /// as the user's own — which is what every surface that has to say whose
     /// skill is steering a turn would then repeat.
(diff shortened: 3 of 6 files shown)


macanderson:
Added `unblocks-main` — this PR fixes one of main's two reds, and the canary guard's own failure text names that label as the designed way through.

The other red is `file-size`: `fleet_cmd.rs` crossed the 1500-line ceiling at `a7a46d3b6` (#4821). I have that one in #4827, scoped to the split and deliberately not touching the three prose files this PR owns, so the two do not conflict. Main needs both.

macanderson:
Added `unblocks-main`, because this PR repairs one of the two things main is red on and was being held by the other one.

Main has two independent red causes right now, and each of the two repairs is blocked by the cause it does not fix:

- **prose** — `check-prose` fails on `filler-adverb` in `goal_wrapped/tests.rs`, `fleet_cmd/wrapped.rs`, and `crates/stella-runtime/README.md`. This PR fixes exactly those. The canary does not run the prose gate, which is why no `main-red` issue names it and why it only shows up on PR gates.
- **file-size** — `fleet_cmd.rs` at 1538 lines, over the 1500 ceiling (#4821). #4825 fixes that.

So this PR was red only on `main is not known-broken` (held by #4821, which it does not fix), and #4825 was red only on `gate steps` (held by the prose failures, which it does not fix). Neither can go first unlabelled. `scripts/check-main-red-hold.sh` has the escape hatch for precisely this — a PR labelled `unblocks-main` passes the hold — and #4825 already carries it, so this one was the only half still stuck.

Checked locally by composing both onto `a7a46d3b6`. They merge with no conflict, and on the composed tree:

- `python3 scripts/check-prose.py` exit 0 — `OK -- 6438 grandfathered construction(s) in 1473 file(s), none added.`
- `make file-size` exit 0 — `OK — 1749 watched files, 19 grandfathered over 1500 lines, and nothing went over by this change.`
- `cargo check -p stella-cli --all-targets` exit 0.

Either order works now that both are labelled. Whichever lands second should get its `gate steps` / `file size ratchet` rerun against the new main rather than the stale run.

I did not merge anything — flagging the ordering for whoever owns these.

macanderson:
Heads up — your `fmt + clippy + test` red is **not yours**.

`cargo clippy -D warnings` rejects `&*registry` in `fleet_cmd.rs` (`explicit_auto_deref`, your line 1242). It reproduces on a branch cut straight from `main` that carries none of your changes, so it is main's third red — alongside the prose you fixed and the `file-size` break in #4821.

That left us deadlocked: your PR needed my clippy fix, mine needed your prose fix. I have cherry-picked your two commits (`84c8e798c`, `6587cc91f`) onto #4827 with authorship and messages intact, so one branch can now prove main green on its own — file-size, clippy and prose together.

Nothing of yours is rewritten. If #4827 lands first this PR becomes an empty diff and can close; if you would rather land this one, say so and I will drop the cherry-picks from #4827 instead. Your call — I just did not want two PRs sitting red on each other's fixes.

macanderson:
Superseded by #4827, which merged and carries all of this.

I opened this before the three `fleet_cmd.rs` split PRs existed; #4827 turned out to include the same two changes:

- the six `filler-adverb` deletions that had `check-prose.py` failing on main, with `scripts/prose-baseline.txt` retightened
- the `gated.rs` and `skills.rs` doc repairs — `principal_of`'s two fused doc comments split back apart, and the public docs that named the private `default_origin_for` rewritten as what they describe

Verified on main after #4827: `python3 ./scripts/check-prose.py` reports `OK -- none added`, and `RUSTDOCFLAGS='-D warnings' cargo doc --no-deps -p stella-tools -p stella-core` is clean.

Closing rather than rebasing: the content is identical, so a rebase would land as an empty diff or a conflict.

PR #4825, opened 2026-08-25
refactor(stella-cli): split fleet_cmd's reporting into a sibling

### LINKED ISSUE #4813: fleet_cmd.rs sits at 1499 of 1500 lines — the next line added there fails the gate
## Problem

After PR #4801, `crates/stella-cli/src/fleet_cmd.rs` is **1499 lines against the 1500-line ceiling**, and it is not grandfathered. The next line anyone adds to that file fails `check-file-size` outright, and the remedy is a split — never a baseline entry (`make file-size-update` refuses to add one).

That PR did the right thing under the constraint (all its new logic went into the sibling `fleet_cmd/durability.rs`), but it leaves the file one line from a wall that the next person will hit mid-change, when they are least able to absorb a refactor.

## Done means

A coherent slice moves out into a sibling submodule the way `fleet_cmd/durability.rs` and `fleet_cmd/wrapped.rs` already do, leaving `fleet_cmd.rs` with real headroom (target: under 1400). A verbatim move is preferred over a rewrite so the diff is reviewable; name any moved or renamed test in the PR body, because `check-deleted-tests` reads it.

Land it on its own from a fresh main — a split folded into a feature PR is what makes the shared-cell collisions this repo keeps hitting.

From PR #4801's report (2026-08-25). Related: #3776, #3441.

### LINKED ISSUE #4821: main is red: file-size fails on the merged tree
The post-merge canary found `main` broken at `a7a46d3b6`.

Failing: **file-size**

### `file-size`

```
check-file-size: FAILED

worse.
  crates/stella-cli/src/fleet_cmd.rs is 1538 lines, over the 1500-line limit
```

canary exists (#3332).

### How to fix

```sh
git checkout main && git pull
```

...then, per failing check:

#### `file-size`

```sh
```

than only CI.

macanderson:
Repairs red main (#4821): `crates/stella-cli/src/fleet_cmd.rs` is 1538 lines, over the 1500 limit and not grandfathered, so the remedy is a split and never a baseline entry.

**This one was predicted.** #4801 left the file at **1499 of 1500** — one line of headroom — and #4813 was filed to say the next line added anywhere in it would fail the gate outright. That line landed. Closing both.

The slice moved is the reporting group: `truncate`, `print_dash_summary`, `render_report`. It is the part with the fewest ties to the run loop — it reads a finished `FleetRunReport` and writes to stdout — so the move is verbatim, with only the three signatures gaining `pub(super)` and the parent re-importing them. No test was deleted or renamed.

`fleet_cmd.rs` is **1425** lines now, so it has real headroom instead of one line of it.

## Evidence, on this branch

```
./scripts/check-file-size.sh   OK — nothing went over by this change
cargo check -p stella-cli      exit 0
cargo test -p stella-cli --bins fleet   58 passed; 0 failed
cargo fmt --all                clean
```

Closes #4821
Closes #4813

## Summary by Sourcery

Split fleet command reporting into a dedicated module to restore file-size compliance and provide headroom for future changes.

Enhancements:
- Extract fleet command reporting utilities into a dedicated sibling module, reducing the main command file below the file-size limit while preserving behavior.

Chores:
- Clarify nearby comments and runtime documentation wording without changing behavior.


macanderson:
Your `gate steps` failure is not from this diff — it is the other half of a red main, and #4823 fixes it.

`check-prose` is failing on `filler-adverb` ("deliberately") in `crates/stella-cli/src/agent/goal/goal_wrapped/tests.rs`, `crates/stella-cli/src/fleet_cmd/wrapped.rs`, and `crates/stella-runtime/README.md`. Those hits are on main already; the canary just never runs the prose gate, so no `main-red` issue names them and they only surface on a PR gate like this one.

That left the two repairs deadlocked: this PR was red only on `gate steps` (prose, which it does not fix), and #4823 was red only on `main is not known-broken` (held by #4821, which it does not fix). I added `unblocks-main` to #4823 — this PR already had it — so both halves can now move.

Composed both onto `a7a46d3b6` locally to check they actually agree. They merge with no conflict, and on the composed tree `check-prose` exits 0, `make file-size` exits 0 (`nothing went over by this change`), and `cargo check -p stella-cli --all-targets` exits 0. The `fleet_cmd.rs` → `fleet_cmd/report.rs` split and #4823 do not touch the same regions.

After #4823 lands, rerun this PR gate against the new main — the current red is a stale run, not a rerun-worthy failure of your diff.

I did not merge anything.

macanderson:
Heads up: three open PRs are now splitting `fleet_cmd.rs` — this one, #4827, and #4829 — and only #4829 clears everything main is red on.

Main at `a7a46d3b6` has three independent red causes: `file-size` (#4821), `check-prose` (`filler-adverb` in `goal_wrapped/tests.rs`, `fleet_cmd/wrapped.rs`, `crates/stella-runtime/README.md`), and `clippy::explicit_auto_deref` on `&*registry` in `fleet_cmd.rs`. The last two have no issue because the canary does not run either gate — its checks are `lockfile-sync`, `file-size`, `compile`, and `compile` is `cargo check`, not clippy. Filed as #4828.

This PR fixes the first. #4829 fixes all three, and I verified it on `713abc2ec`: prose exit 0, `make file-size` exit 0, `cargo clippy -p stella-cli --all-targets -- -D warnings` exit 0, `cargo fmt --all --check` exit 0, `cargo test -p stella-cli --bins fleet` 58 passed.

So landing this one still leaves main red, and landing two of the three means a second split of the same file to unpick. Suggest closing this in favour of #4829 unless it has something #4829 lacks — worth a look, since your `fleet_cmd/report.rs` boundary and its `fleet_cmd/render.rs` boundary are not the same cut.

Your `gate steps` red is the prose half, not your diff. Not merging anything — this is for whoever owns these three.

macanderson:
Superseded by #4827, which merged.

Three PRs were splitting `fleet_cmd.rs` at once (#4825, #4827, #4829), each extracting a different half, so they conflicted rather than stacked. #4827 was taken because it cleared **both** of main's reds; this one clears only the file-size ceiling.

Its `fmt + clippy + test` failed for the reason called out on #4821 before the runs finished: `&*registry` at `fleet_cmd.rs:1242` is a `clippy::explicit_auto_deref` error, and `-D warnings` makes it fatal. Landing a split without it would have turned the file-size gate green while main stayed red — and nothing merges past `main is not known-broken` in that state.

Nothing here is lost: the report/render extraction this proposed is still available if `fleet_cmd.rs` needs more headroom later.

PR #4826, opened 2026-08-25

macanderson:
Reopens the one change #4812 was carrying that is not on `main`. #4812 was
closed as "every piece this PR carried is present there", and for its main-red
repairs that is true — #4814 landed them independently. It is not true for this
piece, which reached that branch from #4817 three minutes before it closed and
went out with it.

On `a7a46d3b6`, `crates/stella-cli/src/agent/graph_view.rs` is 199 lines and
mentions the CGP host only in a doc comment, and
`crates/stella-cli/src/agent/graph_view/tests.rs` does not exist. Both are
checkable with `git cat-file -e origin/main:<path>`.

**The GRAPH q box answers through the CGP host, not CodeGraph directly.**
`graph_view` queried CodeGraph itself, so a workspace whose graph is served by a
CGP host got nothing back — and an unanswered fan-out looked exactly like a
query that legitimately matched nothing. It now asks the host, and those two
outcomes stay distinguishable.

Evidence on this head (`d27d66d11`, which is `main` plus one commit):

- `cargo test -p stella-cli --bins graph_view` — 11 passed, 0 failed. Among
  them `a_query_is_answered_through_the_host`,
  `a_fanout_nobody_answered_is_not_an_empty_answer`, and
  `a_query_matching_nothing_is_an_empty_answer_not_a_missing_one`, which is the
  pair the old path could not tell apart.
- `cargo check -p stella-cli --all-targets` exit 0.
- `cargo fmt --all --check` exit 0.

The workspace-wide gates are CI's on this head.

`main` is red at `a7a46d3b6` on `file-size` (#4821, claimed by #4825), so the
`main is not known-broken` check here will stay red until that lands. It is not
this branch.

The pre-rebuild head of #4812 is kept at `insurance/4812-preconflict`.

## Summary by Sourcery

Route GRAPH tab queries through the CGP host and preserve clear distinctions between unavailable answers and empty results.

Bug Fixes:
- Route GRAPH tab queries through the CGP host so graph-backed workspaces return their results correctly.
- Distinguish an unanswered provider fan-out from a valid query that matched no symbols.

Enhancements:
- Translate host-provided symbol and graph frames into graph nodes, edges, locations, and file relationships while preserving the picker’s direct index inventory.
- Share a dedicated code-graph host across GRAPH tab requests and handle synchronous file focusing separately from asynchronous queries.

Tests:
- Add unit and end-to-end coverage for frame mapping, URI and range handling, graph relationships, host-routed queries, empty matches, unanswered providers, and missing indexes.

### DIFF
--- a/crates/stella-cli/src/agent/graph_view.rs
+++ b/crates/stella-cli/src/agent/graph_view.rs
@@ -13,7 +13,26 @@
 //! ([`graph_snapshot_focus`]) or a free-form query ([`graph_query_snapshot`])
 //! — and both report the wall clock they spent, because the deck cannot
 //! measure a query it did not run (#4335).
+//!
+//! The two reach the index by different routes, on purpose. The picker names
+//! a **file** and wants that file's stored neighborhood verbatim, so it reads
+//! [`stella_graph::CodeGraph`] directly. The `q` box asks a **question**, and
+//! a question is what the Context Graph Protocol exists to route: it goes out
+//! as a [`ContextQuery`] through a [`contextgraph_host::Host`] and comes back
+//! as [`ContextFrame`]s, so the tab names no `stella-graph` type and a
+//! graph-capable provider registered on that host later answers the `q` box
+//! with no change here (#4335).
+//!
+//! Going through frames costs fidelity the direct path keeps, and the cost is
+//! paid where [`node_kind`] and [`frame_location`] are written: a frame
+//! carries the citation keyword (`fn`), not the index's stored tag
+//! (`function` / `method`), and carries a symbol's line inside
+//! `provenance.range` as the string `"L12-40"` rather than as a number.
+//! Both are recoverable, neither is free, and a reader deciding whether to
+//! move the picker onto this path should read those two functions first.
 
+use contextgraph_host::Host;
+use contextgraph_types::{ContextFrame, ContextQuery, FrameKind, Relation};
 use stella_tui::{GraphEdge, GraphNode, GraphSnapshot};
 
 /// Query the code graph (if `stella init` has built it) for the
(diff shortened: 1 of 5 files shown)


macanderson:
Both reds here are main`s, not this diff.

`gate steps` fails on `check-prose` in `crates/stella-cli/src/agent/goal/goal_wrapped/tests.rs`, `crates/stella-cli/src/fleet_cmd/wrapped.rs`, and `crates/stella-runtime/README.md`. This PR touches `agent/graph_view.rs`, `agent/graph_view/tests.rs`, `contextgraph.rs`, `command_deck/graph_input.rs` and `command_deck.rs` — none of those three files, and `check-prose` reports the same six hits on a clean checkout of `a7a46d3b6`.

`main is not known-broken` is the #4821 hold.

Main is red on three gates (`file-size`, `check-prose`, `clippy`); #4829 closes all three, and both of these clear when it lands. I have not labelled this PR `unblocks-main` — it does not repair main, so the hold is doing its job here.

Nothing to rerun until main is green.

macanderson:
Updated this branch onto current `main`.

Its three reds were **inherited, not this PR's**: the run started at 06:45, while main was broken three ways — `file-size` (`fleet_cmd.rs` over the 1500-line ceiling, #4821), `clippy -D warnings` (`explicit_auto_deref` on `&*registry`), and `check-prose` (six `deliberately` hits from the new pattern). All three landed as fixes in #4827 at 07:30, and main's CI has been green since (`ea9632d6b` and `8699f5f0a` both `success`; `a7a46d3b6` before them was `failure`).

A `gh run rerun` would not have helped — it replays the stale merge commit, so it would have re-tested against the broken base and come back red for the same reason. Branch protection is strict here anyway, so the update was needed before merge regardless.

Fresh CI is running now. Untouched otherwise — no conflicts to resolve; this PR was `MERGEABLE` and blocked only by those checks.

Issue #4828, opened 2026-08-25
The main canary calls main green while the prose gate is red


macanderson:
## What happens

`scripts/main-canary.sh` runs three checks — `lockfile-sync`, `file-size`, and
`compile`. It does not run `check-prose`. So when `main` is red on the prose
gate and nothing else, the canary reports green and files no `main-red` issue.
Main stays broken with no tracker, and the failure only appears later on some
unrelated PR's `gate steps` job, where it reads as that PR's fault.

## It already happened

At `a7a46d3b6` the canary ran and reported one failure:

```
main-canary: ok   lockfile-sync
main-canary: FAIL file-size
main-canary: ok   compile
main-canary: FAIL — main is red at a7a46d3b6 (file-size).
```

(run 32816170002, which opened #4821 for the file-size half)

Prose was red at that same commit and went unmentioned. Checking out
`a7a46d3b6` clean and running the gate:

```
$ python3 ./scripts/check-prose.py ; echo $?
  crates/stella-cli/src/agent/goal/goal_wrapped/tests.rs [filler-adverb]: 0 allowed, 1 found
        32: [filler-adverb] 'deliberately'
  crates/stella-cli/src/fleet_cmd/wrapped.rs [filler-adverb]: 0 allowed, 2 found
       125, 131: [filler-adverb] 'deliberately'
  crates/stella-runtime/README.md [filler-adverb]: 2 allowed, 3 found
       167, 187, 215: [filler-adverb] 'deliberately'
1
```

The cost was a deadlock. #4825 fixes file-size and was red only on `gate steps`
(the prose hits, which it does not fix). #4823 fixes prose and was red only on
`main is not known-broken` (held by #4821, which it does not fix). Neither could
go first, and the prose half had no `main-red` issue to justify labelling it
`unblocks-main`, because the canary never filed one. Broken by hand with the
label; the gap that produced it is still open.

## Why it is the same class as #3660

The `compile` row exists because the canary once reported `ok` for everything it
knew how to ask while the composed tree did not build — the script's own header
narrates that. This is that shape again with a different gate: a check that gates
every PR, is not in the canary's list, and can therefore be red on main
indefinitely without a tracker.

## Definition of done

- `check-prose` is in the `checks=()` array in `scripts/main-canary.sh`, so a
  prose-red main files a `main-red` issue like any other cause.
- The failure hint block (the `case` beside `lockfile-sync)` / `file-size)` /
[...]

PR #4827, opened 2026-08-25
fix(stella-cli): fleet_cmd's branch watch becomes a sibling module

macanderson:
Closes #4821. Labelled `unblocks-main` — this is main's `file-size` red, which is currently reddening every open PR.

## The break

`fleet_cmd.rs` reached 1538 lines. The 1500-line ceiling is not grandfathered for it, so the post-merge canary caught `main` broken at `a7a46d3b6`.

## The seam

`--watch` is the one concern in the file that *begins where every other one has finished*: the fan-out has settled, the dashboard is torn down, the report is written, and what remains is which branches landed commits, the capped CI wait on each, and the line each prints.

`BranchWatch`, `watch_targets`, `watch_branch` and `render_watch_line` move to `fleet_cmd/branch_watch.rs` **unchanged** — same bodies, same doc comments — beside the `wrapped` and `durability` modules split out on the same reasoning. No behaviour changes; this is a move.

## Worth naming

`cargo check -p stella-cli` was green while `cargo test` could not build. `fleet_cmd/tests.rs` calls three of these directly and is `#[cfg(test)]`, so a check that only compiles the binary cannot see it. That is the same shape that hid `keymap.rs`'s key-witness `include_str!` in #4822 — the failure mode of verifying a move with `cargo check`.

## Evidence

- `fleet_cmd.rs` is 1435 lines; `./scripts/check-file-size.sh` — OK.
- `cargo test -p stella-cli --bins` — 2288 passed, 0 failed.
- `cargo fmt --all --check` and `cargo clippy -p stella-cli --all-targets` — clean.

## Not fixed here

`check-prose` is still red on `agent/goal/goal_wrapped/tests.rs`, `fleet_cmd/wrapped.rs` and `stella-runtime/README.md`. That is main's *other* red and #4823 already owns it — left untouched rather than fixed twice into a conflict.

## Summary by Sourcery

Extract fleet branch watching into its own module to keep the fleet command maintainable and within the file-size limit.

Enhancements:
- Move fleet branch watch functionality into a dedicated sibling module while preserving its existing behavior.
- Reduce fleet command source size below the project file-size limit.

Tests:
- Update fleet command tests to use the extracted branch-watch module.

Chores:
- Refresh prose baseline wording across affected documentation and comments.

PR #4829, opened 2026-08-25
fix(stella-cli,stella-runtime): fleet_cmd splits its report half; main is green

### LINKED ISSUE #4821: main is red: file-size fails on the merged tree

macanderson:
`main` at `a7a46d3b6` fails `check-file-size` (#4821):
`crates/stella-cli/src/fleet_cmd.rs` is 1538 lines and is **not** grandfathered,
so the guard's own remedy is a split — "Do not add a baseline entry: the
baseline only covers files that predate the guard, and this is the rule that
stops the tree getting worse."

## The split

`fleet_cmd::render` takes the four functions that turn a finished run into
terminal output — `render_watch_line`, `truncate`, `print_dash_summary`,
`render_report` — plus the `SUMMARY_CHARS` width they share. They are the group
with no callers beyond the two print points and no dependency on the run
machinery: they read a finished report and write to stdout, which is why they
move together and why nothing else had to move with them.

`fleet_cmd.rs` 1538 → **1375**. `render.rs` is 177.

`SUMMARY_CHARS` is imported by the test module from `render` directly rather
than re-exported through the parent — the parent does not use it, and carrying
it only to satisfy a test is how an unused-import warning gets created.

## Two more reds on the same tree

Composed here rather than raced on separate branches, because separate repairs
for one red main deadlock each other behind this very check:

- **prose** — four `deliberately` filler-adverbs over the per-pattern
  allowance (`goal_wrapped/tests.rs`, `fleet_cmd/wrapped.rs` ×2,
  `crates/stella-runtime/README.md`). Each sentence already states its reason;
  the adverb sat in front of it.
- **clippy** — `explicit_auto_deref` on `RepublishingDriver::new`'s registry
  argument. `&Arc<T>` coerces to `&T` without the explicit reborrow.

## Evidence

| check | result |
|---|---|
| `make guards-fast` | exit 0 — `check-file-size` OK, `check-prose` OK |
| `make lint CARGO_SCOPE="-p stella-cli"` | exit 0 |
| `make doc-warnings` (stella-cli + stella-runtime) | exit 0 |
| `cargo test -p stella-cli --bin stella` | 2288 passed, 0 failed |
| `cargo fmt --check` | clean |

Labelled `unblocks-main` per the hold check's instruction, which is otherwise
red on #4821 and would block its own repair.

Closes #4821.

## Summary by Sourcery

Split fleet report rendering into its own module and clean up guard and lint violations to unblock the main branch.

Bug Fixes:
- Fix the clippy warning in fleet task setup by relying on implicit dereferencing for registry arguments.
- Remove unnecessary filler wording from prose checked by repository guards.

Enhancements:
- Split fleet terminal rendering into a dedicated module, reducing fleet_cmd.rs below the file-size limit while preserving report output behavior.

### DIFF
--- a/crates/stella-cli/src/agent/goal/goal_wrapped/tests.rs
+++ b/crates/stella-cli/src/agent/goal/goal_wrapped/tests.rs
@@ -29,7 +29,7 @@ use crate::wrapper_plugin::child_turn_fixture::{asking_plugin, step_usage_count}
 /// rounds, so after a round the slot holds *that round's* sender rather than
 /// nothing.
 ///
-/// **The stand-in deliberately does not forward to the run's channel**, and
+/// **The stand-in does not forward to the run's channel**, and
 /// that is the instrument the assertion turns on. The real per-turn sender is a
 /// `TurnFacts` tap over a clone of the run's channel, so a child that metered
 /// into it would still reach the drain and be indistinguishable from one that
--- a/crates/stella-cli/src/fleet_cmd.rs
+++ b/crates/stella-cli/src/fleet_cmd.rs
@@ -60,7 +60,6 @@ use std::collections::HashSet;
 use std::io::IsTerminal;
 use std::path::Path;
 use std::sync::Arc;
-use std::time::Duration;
 
 use colored::Colorize;
 use stella_core::{Engine, TurnOutcome};
@@ -81,9 +80,6 @@ use crate::runtime::{SystemClock, TokioSleeper, WallClock};
 // attempt publishes its own channel across the dispatch's points (#4730).
 use crate::wrapper_plugin::PointStream;
 use crate::{agent, plain, rules};
-
-/// Cap on the per-task summary line so the report table stays a table.
-const SUMMARY_CHARS: usize = 96;
 
 /// Run a fleet: build/load the plan, dispatch it wave by wave, report —
 /// then, with `watch`, hold the fleet PR/CI monitor on the branches.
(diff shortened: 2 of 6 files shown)


macanderson:
Verified this one locally, and it is the only open PR that clears all three of main`s red causes. Evidence on `713abc2ec`:

- `python3 scripts/check-prose.py` — exit 0, `OK -- 6440 grandfathered construction(s) in 1473 file(s), none added.`
- `make file-size` — exit 0, `OK — 1749 watched files, 19 grandfathered over 1500 lines, and nothing went over by this change. Judged against a7a46d3b6.`
- `cargo clippy -p stella-cli --all-targets -- -D warnings` — exit 0.
- `cargo fmt --all --check` — exit 0.
- `cargo test -p stella-cli --bins fleet` — 58 passed, 0 failed.

Main at `a7a46d3b6` is red on three independent things, not the one the canary filed:

| cause | on main | this PR |
|---|---|---|
| `file-size` — `fleet_cmd.rs` at 1538 lines, over 1500 (#4821) | red | fixed by the split |
| `check-prose` — `filler-adverb` in `goal_wrapped/tests.rs`, `fleet_cmd/wrapped.rs`, `stella-runtime/README.md` | red | fixed |
| `clippy::explicit_auto_deref` — `&*registry` in `fleet_cmd.rs` | red | fixed (`&registry`) |

The clippy one had no owner and no issue. The canary missed both it and the prose half because its `checks` array is `lockfile-sync`, `file-size`, `compile` — and `compile` is `cargo check`, not clippy. Filed as #4828.

#4825 and #4827 each fix only the file-size half, so either would land leaving main red on the other two. This PR supersedes both. Worth closing them in its favour rather than composing — three branches splitting one file compose into a worse tree than any of them.

It carries `unblocks-main`, so the hold will pass on a rerun. I have not merged anything.

macanderson:
Superseded by #4827, which merged.

Both were green and both fixed the file-size ceiling and the `&*registry` clippy error, so this is not a quality call — it is that only one split of `fleet_cmd.rs` can land. A second extraction of the same file arrives as a conflict against the first, not as an addition.

#4827 was taken because it also carried the `gated.rs` / `skills.rs` doc repairs and the six `filler-adverb` prose deletions with a retightened baseline, so one merge closed every red on main at once instead of leaving a second round to compose.

Your `render.rs` extraction was the larger one (−170 vs −108), which leaves `fleet_cmd.rs` more headroom under the 1500 ceiling. If the file creeps back up, this is the split to revive.

Issue #4853, opened 2026-08-25
Cancelling the lead turn can still show a low cost if the last events have not saved yet


macanderson:
## Problem

#2807 had two independent fixes. The first is done: `close_dropped_execution` (`crates/stella-cli/src/command_deck/dropped_turn.rs`) and `close_worker_execution` (`crates/stella-cli/src/subsession/closeout.rs`) now read the execution row back after closeout and report that instead of the caller's own in-memory total, which is right as of whatever the store had already recorded.

The second fix was deliberately left out, called out in `dropped_turn.rs`'s module doc:

> Not fixed here, and #2807's second half: nothing awaits the forwarder drain before the row closes, so a `StepUsage` can still be priced after `record_execution_end` has run. The read-back heals whatever landed by the time it runs and no more.

On the lead's cancel path, `close_turn_stream` (`crates/stella-cli/src/command_deck/forwarder.rs:342`) — the only thing that awaits the forwarder's drain — lives inside `run_lead_turn`'s own future (`crates/stella-cli/src/command_deck/lead_turn.rs`), which a cancel drops before it runs. So a `StepUsage` event still in flight when the user hits cancel can be recorded and priced *after* `record_execution_end` already closed the row, and the read-back has nothing to read back yet.

(The worker path in `subsession.rs` does not have this half of the bug — `close_turn_stream` there is awaited before the cost is computed, which is why #2807's fix there was the read-back alone.)

## Why it matters

Same one-directional bias as #2807 and #2570: it only ever under-reports, and worst on the runs that cost the most, right at the moment the user is looking at the number.

## Suggested fix

Give the lead's cancel path a way to wait for the drain — the pieces `close_turn_stream` needs (`tx`, the `JoinHandle`, `registry.detach_event_stream()`) are owned by `run_lead_turn`, so this likely means moving them up to the driver loop (`command_deck.rs` / `lead_turn.rs`) or handing the cancel path a drain handle.

This must not reintroduce the #2290 freeze: that issue is why `detach_event_stream()` exists (a registry-held `EventSender` keeping the channel open forever), and any restructuring of the forwarder's lifecycle has to preserve the fix. `command_deck.rs` is a god file closed to growth — new logic goes in a sibling module (`dropped_turn.rs` or a new one), same as #2807's first fix.

## Verify

[...]

PR #4854, opened 2026-08-25
fix(stella-cli): a cancelled worker's cost is read back from the row it just closed

### LINKED ISSUE #2807: The cost shown when you cancel a run is still wrong — only the saved number was fixed
## Problem

#2570 fixed the saved number: `executions.cost_usd` is now kept at or above the sum of the execution's `telemetry` receipts (money actually spent), both as each receipt lands (`Store::record_telemetry`) and again at close-out (`Store::finish_execution_accounted`) — see `crates/stella-store/src/telemetry.rs`.

It did not fix the number the driver computes and shows. On interactive mode's cancel path, the cost handed to the store — and printed to the user, and returned to callers — is still the truncated in-memory total:

- `crates/stella-cli/src/command_deck.rs:2234`
  ```rust
  let cancelled_cost =
      agent::settled_cost_since(dispatch_spend_usd, budget.session_spent_usd());
  ```
  `dispatch_spend_usd` is captured at `command_deck.rs:1595`. When the turn future is dropped mid-cancel, the roles still finishing up (worker, witness author — the expensive ones) have not been added to `budget.session_spent_usd()`, so this difference is just the pre-cancel prefix. Execution 99 in #2570 showed `$0.8659773` reported against `$5.28786465` of actual receipts — an exact triage+research+plan prefix.

- The same call site does not wait for the forwarder to finish draining before closing the row. `spawn_forwarder`'s `JoinHandle` and the event `tx` are created inside `run_lead_turn` (`command_deck.rs:4162`), and `close_turn_stream` (`crates/stella-cli/src/command_deck/forwarder.rs:172`) — the only thing that waits for the forwarder — lives inside that same future, which the cancel drops. So the last calls' `StepUsage` events can be saved (and priced) after `record_execution_end` has already run.

The store now survives both of these (the per-receipt floor fixes a late receipt after close-out), but anything that reads the driver's own figure instead of the saved row still under-reports: the on-screen cost display, the `cancelled_cost` value itself, and any caller of `PipelineOutcome::total_cost_usd` / `WorkerOutcome::cost_usd` on a cancelled run.

## Why it matters

Same reason as #2570: the error only ever under-counts, and it's worst on the most expensive runs — and it's the number a person sees on screen the moment they cancel. Someone reading "$0.87" for a run that cost $5.29 is being misled by the display even though the database is now right.

## Suggested fix

Either of these helps on its own:

1. Read the true number back instead of recomputing it: after the store closes the row, use `Store::execution_settled_cost_usd(execution_id)` (or re-read `execution_summary(...).cost_usd`, which is now the floor over both sources) as the figure to display and return.
2. Wait for the drain before closing: give the cancel path a way to wait for the forwarder — the pieces `close_turn_stream` needs (`tx`, the `JoinHandle`, `registry.detach_event_stream()`) are owned by `run_lead_turn`, so this likely means moving them up to the driver loop or giving the cancel path a drain handle. Note #2290 is why `detach_event_stream()` exists; whatever this becomes must not bring back the freeze it fixed.

## Verify

Write a test in `crates/stella-cli/`: run a turn that saves receipts for several roles, cancel it, and check that the cost the driver reports matches the receipts total — not the prefix. The store-side tests in `crates/stella-store/src/telemetry.rs` (`a_cancelled_execution_reports_every_receipt_it_paid_for`, `a_receipt_landing_after_closeout_still_raises_the_rollup`) show the shape and already pass, so a failure here points squarely at the driver's number.

## Constraints

- `crates/stella-cli/src/command_deck.rs` and `src/agent.rs` are closed to growth (see AGENTS.md § God files) — new logic goes in a separate module.
- Do not lower `executions.cost_usd`: it must stay a floor over both the receipts and the caller's total, on purpose (see `finish_execution_accounted`'s doc comment).

## Done when

The cost a cancelled run reports on screen and returns to its callers matches the receipts the run actually paid for, with a test that fails on today's code and passes with the fix.

Refs #2570, #2290.

macanderson:
## What & why

#4540 fixed the lead driver's cancel path (`command_deck/dropped_turn.rs`) but
left the sibling worker path in `subsession.rs` on the exact same broken
shape: a stopped lane's cost is computed from its own budget guard at the
instant the raced turn future is dropped, missing roles still settling
receipts. That number becomes `WorkerOutcome::cost_usd` and flows into
`SupervisorMsg::Ended`, which the driver adds straight into the session's
own spend total — the exact caller #2807 names as still under-reporting.

`close_worker_execution` now reads `execution_summary` back after
`record_execution_end` closes the row and returns that instead of the
caller's raw aggregate. `finish_execution_accounted` floors the row at
`MAX(cost_usd, receipts)` and never below it, so the read-back is safe for
every outcome, not just cancelled.

The forwarder-drain half of #2807 (waiting for in-flight `StepUsage` events
before the row closes) was deliberately deferred by #4540 and remains open,
filed separately as #4853.

Closes #2807

## The witness

`a_stopped_workers_settled_cost_is_read_back_from_its_receipts` in
`crates/stella-cli/src/subsession/closeout.rs` drives a worker execution with
two landed receipts ($0.50 + $4.00) through `close_worker_execution` with a
stale $0.50 caller aggregate and asserts the returned cost is $4.50. Verified
by hand that it fails against the pre-fix function (asserts `0.5`) and passes
against this change.

## The gate

- [x] `cargo fmt --check` (`-p stella-cli`)
- [ ] `cargo clippy --workspace --all-targets -- -D warnings` — not run locally per repo policy (no workspace-wide builds on this machine); CI runs it
- [ ] `cargo test --workspace` — not run locally per repo policy; ran targeted `cargo test -p stella-cli --bin stella closeout` instead (passes)
- [x] Docs updated where behavior/flags changed (doc comment on `close_worker_execution` explains the read-back)
- [x] CLA signed

## Nothing left behind

- [x] Filed: #4853 (the forwarder-drain half of #2807, deferred by design — see `dropped_turn.rs`'s module doc)

## Ground-rule check

`close_worker_execution`'s return type changed from `()` to `f64`; the two
existing tests in `closeout.rs` (`completed`, cost `0.0`, no receipts) still
pass unchanged since the read-back is a no-op floor there.

## Summary by Sourcery

Read back the finalized worker execution cost after closeout so cancelled work cannot under-report settled usage.

Bug Fixes:
- Report a cancelled worker's settled execution cost from the persisted execution row so receipts that arrive after the caller's local aggregate are included in session spend totals.

Enhancements:
- Document the worker closeout read-back behavior and track the separate forwarder-drain limitation.

Tests:
- Add a regression test covering a stopped worker whose persisted receipts exceed its stale caller-side cost.

### DIFF
--- a/crates/stella-cli/src/command_deck/dropped_turn.rs
+++ b/crates/stella-cli/src/command_deck/dropped_turn.rs
@@ -23,7 +23,7 @@
 //! the deck's own running total stops under-reporting the turn the user just
 //! stopped.
 //!
-//! Not fixed here, and #2807's second half: nothing awaits the forwarder drain
+//! Not fixed here, and tracked as #4853: nothing awaits the forwarder drain
 //! before the row closes, so a `StepUsage` can still be priced after
 //! `record_execution_end` has run. The read-back heals whatever landed by the
 //! time it runs and no more. Draining first must not reintroduce the #2290
--- a/crates/stella-cli/src/subsession.rs
+++ b/crates/stella-cli/src/subsession.rs
@@ -1127,14 +1127,14 @@ async fn run_worker(
     // private board is scaffolding for this one run, and the session's
     // `tasks` rows have exactly one writer: the driver, whose `/clear` seal
     // this thread cannot consult (#1708 — see `closeout`'s module docs).
-    closeout::close_worker_execution(
+    let settled_cost = closeout::close_worker_execution(
         execution.as_ref(),
         &registry,
         label,
         cost,
         persistence_complete,
     );
-    (execution_id, cost, end)
+    (execution_id, settled_cost, end)
 }
 
 /// Drain the driver's prompt backlog into free worker slots, oldest first.
--- a/crates/stella-cli/src/subsession/closeout.rs
+++ b/crates/stella-cli/src/subsession/closeout.rs
@@ -41,20 +41,31 @@ use crate::agent;
 /// Close out a worker's execution row: agent and MCP usage, and the outcome
 /// label ([`agent::record_execution_end`]) — best-effort, exactly as the
 /// worker's closeout always was (the inbox notification and the lane's own
-/// events are the user-facing signal).
+/// events are the user-facing signal). Returns the settled cost the caller
+/// should report — see below.
 ///
 /// The signature is the contract (#1708): no session id comes in, so the
 /// closeout **cannot** address the session-keyed `tasks` rows at all. See
 /// the module docs for why that absence is deliberate.
+///
+/// # Why this reads the row back
+///
+/// `cost_usd` is the caller's own aggregate, and on a stopped worker (#2807)
+/// that aggregate is a prefix: the turn future was dropped mid-race while
+/// other roles were still settling receipts the caller's budget guard never
+/// saw. `record_execution_end` floors the row at `MAX(cost_usd, receipts)`
+/// and never below it, so the row is always at least as tight as what was
+/// passed in — reading it back and handing that to the caller is strictly
+/// safe, not just for the stopped case.
 pub(crate) fn close_worker_execution(
     execution: Option<&(Arc<Store>, i64)>,
     registry: &ToolRegistry,
     outcome_label: &str,
     cost_usd: f64,
     persistence_complete: bool,
-) {
+) -> f64 {
     let Some((store, id)) = execution else {
-        return;
+        return cost_usd;
     };
     let _ = agent::record_execution_end(
         store,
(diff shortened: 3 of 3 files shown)


macanderson:
All three ❌ rows score this PR against the full #2807 scope as if it were the only fix, missing prior/parallel work:

- Row 1 (lead driver path unaddressed): already fixed by #4540 (merged 2026-08-23) — `close_dropped_execution` in `command_deck/dropped_turn.rs` reads the row back and corrects the session accumulator. This PR's only touch to that file is a comment pointing the drain-wait note at #4853.
- Row 2 (forwarder drain not awaited): deliberately deferred, filed as #4853. The worker path this PR fixes doesn't have that bug — `close_turn_stream` is awaited before the worker's cost is computed, so there's no drain to wait for here.
- Row 3 (no CLI-level cancelled-turn test): the lead driver path is witnessed by #4540's `the_guard_is_corrected_from_the_row_the_closeout_settled`. This PR's own witness (`a_stopped_workers_settled_cost_is_read_back_from_its_receipts`) covers the worker path at the store level, matching the constraint that `command_deck.rs`/`agent.rs` are closed to growth.

No changes needed in this PR.

Issue #4858, opened 2026-08-25
The plain output's default transcript still shows a delegate's tool call as the lead's


macanderson:
## Problem

#4699 taught the interactive deck (`stella-tui`'s `SessionModel`/`TranscriptEntry`) and the CLI's archive export and legacy card renderer (`STELLA_PLAIN_STREAM=1`) to attribute a delegate's tool calls. It did not reach the **default** non-interactive transcript path:

- `crates/stella-tui/src/transcript_build.rs`'s `RunBuilder` folds `AgentEvent::ToolStart`/`ToolResult` into `stella_transcript::model::Call` and drops `sub_agent_id` on the floor (`AgentEvent::ToolStart { call, .. }`, `AgentEvent::ToolResult { call_id, output, duration_ms, speculated, .. }`).
- `stella_transcript::model::Call` (`crates/stella-transcript/src/model.rs`) has no `sub_agent_id` field to carry it even if the fold captured it.
- `crates/stella-transcript/src/grid.rs`'s `render_turn_lines` has nothing to render.

This path is the *default* for `stella`'s non-interactive text output (`crates/stella-cli/src/plain/transcript.rs`'s `TranscriptPrinter`, used whenever `STELLA_PLAIN_STREAM` is unset) — the more consequential half of the plain surface, since the legacy per-line cards behind that env var are opt-in. So the same gap #4699 fixed for the deck and the archive export still stands for the surface most non-interactive runs (CI logs, `stella run` piped output) actually use.

## Why this is not a small addition

`Call::header_object` looked like a free-text place to splice a `· agent d:1` suffix, but it is structurally consumed, not just displayed:
- `Call::read_path()` returns it verbatim as the file path for a `ReadFile` call (used for syntax highlighting by extension elsewhere).
- The `args` row filter compares `row.value.trim() != self.header_object.trim()` to dedupe an arg that only restates the header.
- `grid.rs` passes it into `digest::fold_output(&call.output, &call.header_object)`.

Appending text to it would corrupt all three. The correct fix is a new field on `Call`, threaded through `RunBuilder::open_call`/`close_call` and rendered by `grid::render_turn_lines` — `stella-transcript` is a shared near-leaf crate (also depended on by `stella-cli` and `stella-observatory`), so this touches its own test suite too.

## Where to look

- `crates/stella-tui/src/transcript_build.rs` — `RunBuilder::open_call`/`close_call`, and the `push` match arms for `AgentEvent::ToolStart`/`ToolResult`.
[...]

Issue #4624, opened 2026-08-24
A delegate child's tool calls are unattributed — tool_calls and journal events carry no sub_agent_id


macanderson:
## Problem

A `delegate` child's `ToolStart`/`ToolResult` events are forwarded onto the parent's stream (`crates/stella-core/src/subagent.rs`, `forwards_to_parent`) and land in `store.db` under the parent execution id with nothing naming the child. Only the two metering records carry `sub_agent_id` (`telemetry`, schema v33, #4383). Consequences:

- The `tool_calls` table (`crates/stella-store/src/ddl.rs`) cannot answer "which tools did child X run" — a child's rows are indistinguishable from the parent's own.
- The Observatory's sub-agent focus page (PR #4623, `#transcript/<id>/sub/<agent>`) can show a child's aggregate spend but not its per-call tool activity or a per-child Steps table; the parent's Steps section likewise cannot mark which rows were a delegate's.
- `/api/execution` steps rows omit `telemetry.sub_agent_id` (`crates/stella-observatory/src/db.rs`, the steps SELECT), so even the metered per-call rows can't be filtered per child client-side.

## What done looks like

1. Stamp `sub_agent_id` on the child-forwarded `ToolStart`/`ToolResult` path (the `child_sender` in `crates/stella-core/src/subagent.rs` currently stamps only `StepUsage`/`UsageIncomplete`) and project it into `tool_calls` (new nullable column + migration in `crates/stella-store`, NULL = the lead's own, same contract as `telemetry.sub_agent_id`).
2. Select `sub_agent_id` in the observatory's `/api/execution` steps and tools queries (degrading on pre-migration stores via `is_missing_schema`, `crates/stella-observatory/src/db.rs`).
3. Filter the sub-agent focus page's sections by it (`renderSubagentFocus`, `crates/stella-observatory/src/assets/index.html`).

## Constraints

- Adding a field to a wire event means `make wire-schema-update` and the serde round-trip tests (`crates/stella-protocol`); the consumers ledger (`crates/stella-protocol/src/event/consumers.rs`) needs no new row for a field, only for a new variant.
- A store migration is a shared-cell hazard (see AGENTS.md on `Cargo.lock`-shaped collisions); keep it its own PR if the tree is busy.
- The schema_conformance suite (`crates/stella-observatory/tests/schema_conformance.rs`) must gain the new column in its seeded fixture, or the drift gate goes tautological.

## Verify

[...]

PR #4860, opened 2026-08-25
The deck and the exporters ignore sub_agent_id, so a delegate's tool rows look like the lead's

### LINKED ISSUE #4699: The deck and the exporters ignore sub_agent_id, so a delegate's tool rows look like the lead's
## Problem

PR #4670 added `sub_agent_id` to `ToolStart`/`ToolResult` and the Observatory now badges a delegate's calls. Nothing under `crates/stella-tui/src` reads the field, so in the deck a delegate's tool rows render identically to the lead's. Same for `crates/stella-cli/src/export/transcript.rs` and `crates/stella-cli/src/plain.rs`.

The TUI is a first-class surface here, so an attribution the web surface shows and the terminal does not is a parity gap, not a nicety.

## Also in scope

- `ToolCallRow` has no `sub_agent_id`, so `Store::record_tool_calls` writes NULL. Only tests call it today (`efficacy.rs`, `foundry/tests.rs`, `tests.rs`) — no production gap yet, but a future bulk writer would silently lose attribution.
- `stella-serve`'s `observe/tally.rs` folds `ToolStart`/`ToolResult` and ignores the field, so the serve surface cannot separate a child's calls either.

## Done means

A delegate's tool row is visibly attributed on the deck, the plain export and the archive export; the store row carries the id; the serve tally can separate them. Golden deck frames re-blessed with the diff read.

From PR #4670's report (2026-08-24). Related: #4624, #4627.

macanderson:
## What & why

The deck, the archive export, and the plain surface's card renderer folded `AgentEvent::ToolStart`/`ToolResult` without reading `sub_agent_id`, so a delegate's tool row rendered identically to the lead's — the Observatory badges this (#4624), these surfaces did not. `TranscriptEntry::ToolStart`/`ToolResult` now carry the field (resolved from the announcement, falling back to the result's own field for a consumer that never saw the start — same precedence the store already uses), and the deck, `export/transcript.rs`, and `plain.rs`'s `tool_call_card`/`tool_result_card` render a visible tag for it. `ToolCallRow` gains the field so `Store::record_tool_calls` stops dropping it, and `stella-serve`'s `TurnTally` counts a turn's delegate-made calls apart from the lead's own.

Closes #4699

## The witness

Witnesses per surface: `model::tests::a_delegates_call_carries_its_sub_agent_id_onto_start_and_result` and `a_result_with_no_matching_start_still_attributes_from_its_own_field` (the fold), `views::transcript_source::tests::a_delegates_head_names_the_delegate` and `render::tests::tool_output::a_delegates_result_row_names_the_delegate` (the deck), `export::transcript::tests::a_delegates_call_is_named_apart_from_the_leads` (the archive export), `plain::tests::agent_tag_names_a_delegate_and_stays_empty_for_the_lead` (the plain card renderer), `tool_calls::tests::the_bulk_writer_persists_sub_agent_id_too` (the store), `tally::tests::a_delegates_calls_are_counted_apart_from_the_leads` (serve).

## The gate

- [x] `cargo fmt --check`
- [x] `cargo test -p stella-tui`, `-p stella-store`, `-p stella-serve`, `-p stella-cli` (targeted — see CI for the full workspace run)
- [ ] `cargo clippy --workspace --all-targets -- -D warnings` — not run locally, CI covers it
- [x] `make guards-fast`
- [x] `Closes #4699` appears above and in the commit trailer

## Nothing left behind

- [x] Filed: #4858 — the plain surface's *default* (non-streaming) transcript path (`RunBuilder` / `stella_transcript::grid`, used whenever `STELLA_PLAIN_STREAM` is unset) still doesn't attribute a delegate's call. It needs a new field on `stella-transcript`'s shared `Call` model rather than a local patch — `header_object` looks free-text but is structurally consumed by `read_path()`, the args-row dedupe, and `digest::fold_output`. Filed as a handoff rather than folded into this PR, which already touches four crates.

## Ground-rule check

- [x] New cross-boundary types round-trip through serde (`TurnTally`'s new field is covered by the existing `ServeEvent` round-trip test)

`TranscriptEntry`'s two tool variants gained a field, which is a wide but mechanical blast radius — every existing `ToolStart`/`ToolResult` literal across `stella-tui`'s test suite needed `sub_agent_id: None`. No golden deck snapshot needed re-blessing: every existing fixture passes `None`, and `None` renders no tag, so the rendered bytes are unchanged.

## Summary by Sourcery

Carry sub-agent identity through tool-call processing so every supported surface distinguishes delegate activity from the lead's own calls.

New Features:
- Attribute tool calls to their originating delegate across transcript entries, deck rendering, archive exports, plain output, and persisted store rows.
- Expose delegate-made tool-call counts separately in serve turn tallies.

Bug Fixes:
- Preserve sub-agent attribution when a tool result is matched to its start or arrives without a corresponding start, preventing delegate rows from appearing as lead calls.

Enhancements:
- Render visible delegate identifiers while leaving lead-owned tool rows untagged.

Tests:
- Add witness coverage for attribution folding, deck and plain rendering, archive export, bulk persistence, and delegate-specific tally counts.


macanderson:
Re the ❌ row on "archive export and all plain-surface export paths": accurate. This PR fixes the archive export (`export/transcript.rs`) and the plain surface's card renderer (`plain.rs`'s `tool_call_card`/`tool_result_card`, used when `STELLA_PLAIN_STREAM=1`). It does not touch the plain surface's *default* (non-streaming) transcript path — `crates/stella-cli/src/plain/transcript.rs`'s `TranscriptPrinter`, which folds through `stella-tui`'s `RunBuilder` into `stella_transcript::model::Call` and renders via `stella_transcript::grid::render_turn_lines`.

That gap is real and out of scope here on purpose: `Call::header_object` looks like a free-text place to splice a delegate tag, but it's structurally consumed (`read_path()` returns it verbatim for syntax highlighting, an args-row dedupe compares against it, `digest::fold_output` takes it as an argument), so the correct fix is a new field on `stella-transcript`'s shared `Call` model — a crate also depended on by `stella-observatory` — threaded through `RunBuilder` and `grid.rs`, with its own test suite to update. That's a separably-sized piece of work, filed as #4858 with the specific reasoning and file pointers.

macanderson:
Sourcery's ❌ on the assessment table (default non-streaming plain path / `RunBuilder`+`stella_transcript::grid`) is a real gap, not a misread. It's out of scope for this PR: `stella_transcript::model::Call` has no `sub_agent_id` field, and `Call::header_object` is structurally consumed (read_path, arg dedupe, digest::fold_output) so it can't be spliced into — this needs a new field threaded through a shared near-leaf crate plus its own test suite, not a small addition here. Filed and scoped as #4858 before this PR opened.

PR #4863, opened 2026-08-25
fix(stella-tui,stella-cli,stella-store,stella-serve,stella-observatory): a delegate's tool call is attributed on every transcript surface

### LINKED ISSUE #4699: The deck and the exporters ignore sub_agent_id, so a delegate's tool rows look like the lead's

macanderson:
## What & why

`sub_agent_id` reached `AgentEvent::ToolStart`/`ToolResult` on the wire (#4670) and the Observatory's SQL-backed telemetry view, but nothing else read it. In the deck a delegate's tool row rendered identically to the lead's own, both CLI plain exporters discarded the field, the HTML archive export discarded it, the store's bulk writer (`ToolCallRow`) had no column for it, the serve tally could not separate a child's calls from the lead's, and the Observatory's own journal-based transcript view built its `Call` struct without ever reading the row's `sub_agent_id`.

This closes every item in #4699's "Done means" checklist:

- **stella-tui**: `TranscriptEntry::ToolStart`/`ToolResult` and `views::transcript::Event` carry `sub_agent_id`; the deck badges a delegate's head row `↳ d:1` — the same glyph `views::subagents` already uses for the separate registered-lane dashboard. A new golden frame (`session_delegate_tool_call`) proves the lead's own call stays unbadged while the delegate's carries the badge.
- **stella-transcript**: `Call` carries `sub_agent_id`, rendered as a `↳ {id}` chip in `digest::step_chips` — the shared fold both the CLI's default plain output (`grid.rs`) and the Observatory's HTML transcript view (`html.rs`) consume, so one change covers both surfaces.
- **stella-cli**: the archive HTML exporter (`export/transcript.rs`) badges a delegate's row (new `.agent` CSS class); the legacy `STELLA_PLAIN_STREAM` path's `tool_call_card` takes an explicit `sub_agent` parameter.
- **stella-store**: `ToolCallRow` carries `sub_agent_id` and `record_tool_calls` writes it — closing the silent-loss gap the issue flagged (no production caller today, tests only).
- **stella-serve**: `TurnTally.delegate_tool_calls` counts a child's share of `tool_calls`; `Metrics`/`Snapshot` expose the same split as `delegate_tool_calls_total` on the `/v1/metrics` scrape surface.
- **stella-observatory**: `transcript_view.rs`'s `call_from_start` now reads `sub_agent_id` off the `tool_start` journal row — a real production gap the issue's own file list didn't name, surfaced by the compiler once `Call` grew the field.

Also relevant: the issue's citations of `stella-tui/src/model.rs:437` / `crates/stella-tui/src/v2/transcript_source.rs` and `stella-serve/src/observe/tally.rs`'s six hardcoded `None` sites had moved or turned out to be test-only fixtures by the time this landed — verified against current `main` rather than assumed from the issue text.

Closes #4699

## The witness

Targeted witnesses added per surface:
- `stella-tui/src/views/transcript_source.rs`: `a_delegates_call_is_badged_on_its_head_row` — unit-level proof the badge renders only when `sub_agent_id` is `Some`.
- `stella-tui/tests/deck_render_snapshots.rs`: `deck_render_snapshots_pin_a_delegates_tool_call` — a real deck frame (blessed and read) showing the lead's own call unbadged and the delegate's badged `↳ d:1`.
- `stella-store/src/tests.rs`: extended `data_plane_tables_roundtrip_and_tool_histogram` to read the `sub_agent_id` column back from SQLite after `record_tool_calls`, proving the bulk writer round-trips it (and that the lead's own call stays `NULL`).
- `stella-serve/src/observe/tally.rs`: `the_fold_separates_a_delegates_calls_from_the_leads_own` — proves `tool_calls` counts every call while `delegate_tool_calls` counts only the attributed subset.
- `stella-serve/src/observe/metrics.rs`: extended the existing tally round-trip test to assert `delegate_tool_calls_total`.

## The gate

- [x] `cargo fmt --check`
- [ ] `cargo clippy --workspace --all-targets -- -D warnings` — not run locally per this repo's rule that CI builds/tests, not the maintainer's laptop; targeted `cargo check`/`cargo test` ran per touched crate instead (stella-tui, stella-transcript, stella-cli, stella-store, stella-serve, stella-observatory — all green, see commit body for counts)
- [ ] `cargo test --workspace` — same reason; see CI run on this PR
- [x] Docs updated where behavior/flags changed — doc comments added on every new/changed field explaining the invariant

## Nothing left behind

Nothing found beyond this PR's scope during implementation. Two deliberate scope boundaries, noted here rather than left silent:
- The TUI's long-form tool-*result* row (`render/entry/tool.rs`'s `result_body`, the SPEC 6.4-governed block under the head) does not repeat the badge — attribution is stated once, on the head row, matching the existing pattern for `duration`/`task` chips. The head row is part of the same visual block and does not disappear once the result arrives.
- The CLI's legacy-path `tool_result_card` likewise does not repeat the badge already shown on its paired `tool_call_card`.

If either is wanted for redundancy/scanability, that's a follow-up, not a gap in this issue's checklist.

## Summary by Sourcery

Propagate delegate tool-call attribution through all transcript and observability surfaces so readers and metrics can distinguish child activity from the lead's own calls.

New Features:
- Attribute delegated tool calls across transcript, UI, export, storage, and observability surfaces with a consistent delegate identifier.
- Expose aggregate delegated tool-call counts through serve metrics.

Bug Fixes:
- Prevent delegate tool-call attribution from being lost in transcript rendering, exports, persistence, telemetry views, and call tallies.

Enhancements:
- Add delegate attribution chips and TUI badges while leaving lead-owned calls unmarked.
- Track delegated tool-call shares separately from total tool calls in turn tallies and metrics.

Tests:
- Add coverage for delegate attribution in transcript rendering, TUI golden frames, store round trips, and serve metrics and tally calculations.


macanderson:
## Conflicts resolved — and this PR is now much smaller than its description says

#4860 merged to `main` as `dd2fd0e7` while this was open. It closes the same issue (#4699) from the same base, so most of what this PR's description claims is **already on `main`**: the deck badge, the archive exporter, `plain.rs`, `ToolCallRow`, and the serve tally all landed there. Every conflict is resolved to `main`'s side — it is the reviewed, merged version, and two implementations of one field is how the next reader learns to distrust both.

**What this PR still contributes** (199 lines vs `main`) is the part #4860 explicitly did *not* do and filed as #4858 on its way out:

- `stella-transcript`'s shared `Call` carries `sub_agent_id`, and `digest::step_chips` renders it as a `↳ {id}` chip. That fold backs both the CLI's default plain transcript and the Observatory's HTML view, so those two surfaces stop drawing a delegate's call as the lead's own.
- The Observatory's `call_from_start` reads `sub_agent_id` off the `tool_start` journal row.
- A deck golden frame (`session_delegate_tool_call`) covering the rendered result.

If that reads as thin enough to fold into #4858 instead, closing this is reasonable — the call is yours.

### Three things the resolution had to repair rather than pick

- **A duplicate counter with no producer.** `c3565986`'s resolution kept both `main`'s `tool_calls_by_sub_agent` and this branch's `delegate_tool_calls` on `TurnTally`, but only the latter kept a fold. `main`'s field was left with no producer, so every settled turn reported it as zero — invisible on the wire, because `skip_serializing_if` drops a zero. `main`'s witness for it, `a_delegates_calls_are_counted_apart_from_the_leads`, was deleted in the same resolution, so nothing failed to say so. One field survives now (`main`'s), with its producer and its test; the scrape counter reading it is `tool_calls_by_sub_agent_total`. Neither spelling appears in `docs/wire`, so no published schema moves.
- **Dead CSS.** The archive exporter's `.agent` rule is deleted — `main`'s exporter writes a plain ` · agent d:1` suffix and never emits the class.
- **A golden blessed against a discarded renderer.** `session_delegate_tool_call.txt` is re-blessed; the diff is one line. `main` badges the result row as well as the head row, which this branch had deliberately left bare, so the golden and the module's prose now describe what the merged renderer actually draws.

### One gap closed

`digest::step_chips` — this PR's only genuinely-new user-visible behaviour after the merge — **had no test**. Every fixture in the crate passes `None`, so the new branch never executed. Added `a_delegates_call_is_chipped_and_the_leads_own_is_not`, asserting both arms, and confirmed it fails with the chip removed and passes with it, so it can tell the fix from its absence.

Nothing from `c3565986` is dropped; that commit is merged here, not replaced.

### Verification

`cargo fmt --check`; `make guards-fast` (13/13); `make file-size`; `cargo test -p stella-serve -p stella-transcript -p stella-tui` (no failures); `cargo check --all-targets -p stella-cli -p stella-observatory -p stella-store`. The `stella-cli` suite was compiled rather than run — its tests can write a real `~/.stella` — and the only remaining change there is the deleted CSS rule. CI on this head is the real gate.

macanderson:
## CI is green on `71372790` — and the `Closes` trailer now points at the wrong issue

All 15 checks pass; `mergeStateStatus` is `CLEAN`.

Two runs were needed. The first failed two jobs on a single cause: an intra-doc link I added, `[TurnTally::tool_calls_by_sub_agent]`, naming a type `metrics.rs` only imports inside its own test module. rustdoc runs under `-D warnings`, so it failed `cargo doc` in the main gate and again in the wire job, which runs the same step. Spelled through `super::event` and verified with `RUSTDOCFLAGS="-D warnings" cargo doc -p stella-serve --no-deps`.

Worth recording what that first run proved anyway, since every other step ran: `cargo fmt --check`, `cargo clippy -D warnings`, `cargo test`, `docs/wire/ still matches the types`, the release smoke build, and **`no test disappears in the merge unacknowledged`** all passed. That last one is the guard that would have caught `main`'s deleted tally witness, and it is green because the witness is restored.

### The trailer

**#4699 is closed** — #4860 closed it. **#4858 is open**, and its title is *"The plain output's default transcript still shows a delegate's tool call as the lead's"*, which is precisely what this PR now contains. So as it stands, merging this closes an already-closed issue and leaves open the one it actually fixes.

Suggest changing the description's trailer to `Closes #4858`. I have not edited the body — another session is working this branch and I did not want to clobber it. The trailer is also in the `2f36c842` commit message, which would need a history rewrite to change; a correct trailer in the PR description is what GitHub acts on at merge, so the body edit alone is enough.

Sourcery's assessment table on the earlier head has three rows, all ✅, no ❌ — those objectives are met by `main`'s implementation, which this merge preserves.

Issue #4866, opened 2026-08-25
Stella can publish a rule that steers it, and no command takes it back


macanderson:
## Problem

Stella can publish a rule that steers it, and **there is no command that takes
it back.**

The rules miner induces directive candidates and `stella proposals` publishes
the kept ones as TOML records under `.stella/rules/`, which `FsRuleSource`
loads into the system prefix. That is a real self-modification path, and it is
witnessed (`mined_rules_land_where_the_loader_reads`).

The reverse does not exist:

- `stella proposals` offers `list` / `keep` / `edit` / `ignore` / `refresh`.
  `Ignore` prevents a write; nothing retracts a rule already on disk.
- `stella memory retire` does **not** suppress a file-backed rule —
  `load_workspace_rules` filters by authority, not by retirement standing.

So reversal today is `rm .stella/rules/<id>.toml`, or a git revert. Every other
evolution surface has a named rollback artifact: memory has
`stella memory reaffirm` / `restore`, tools have `stella tools --disable`,
workflow has `stella tune rollback`. The framework row in
`crates/stella-parity/src/evolution.rs` records this gap in prose because
nothing better was true when it was written.

This matters because the retirement machinery already exists and is
append-only: memory records are retired by appending a promotion event, never
by editing history. A published rule is the one self-modification that leaves
the ledger and lands as a file, and files are the thing the ledger was designed
not to need.

## Definition of done

1. A named command retracts a published rule — most likely
   `stella proposals retract <id>` or extending `stella context retire` to
   cover file-backed records.
2. Retraction is append-only and consistent with the rest of the lifecycle: it
   writes a `PromotionEvent` with `LedgerAction::Retired` rather than deleting
   bytes, so the history of what Stella believed stays readable.
3. `load_workspace_rules` honours retirement standing, so a retracted rule
   stops reaching the system prefix on the next load.
4. A witness test goes fail → pass: publish a rule, confirm the loader reads
   it, retract it, confirm the loader no longer does and the ledger records
   who retracted it and why.
5. The framework row's `rollback` column in
   `crates/stella-parity/src/evolution.rs` is updated to name the new command.

## Constraints

- Retirement elsewhere never deletes. Match that — a retraction that removes
  the file loses the record that Stella ever believed it.
- `.stella/rules/promotions.jsonl` is hash-chained
  (`crates/stella-core/src/records/promotion.rs`); appending must keep
  `parse_and_verify` passing.

## Related

[...]

Issue #4865, opened 2026-08-25
The evidence gate is declared but nothing calls it before publishing a skill, rule, or tool


macanderson:
## Problem

#2782 landed the evidence-grade policy and #2780 landed the evolution-surface
ledger, but **no live publication path calls the gate.**
`stella_protocol::provenance::authorises` exists, is total, and is tested; the
code that actually publishes a skill, a rule, or a foundry tool does not
consult it. AGENTS.md rule #11 says so out loud rather than implying otherwise.

So today the terms are declared and reviewable, and they are not enforced at
the moment of publication.

## The specific gap this exposes

The ledger's framework row declares `ImpactClass::SteeringDirective`, which
under the policy requires `ProvenanceGrade::EnvironmentObservation`. The rules
miner induces directives from **reflection observations**, which grade to
`ProvenanceGrade::ModelCritique` — two rungs below. A mined rule kept by a
human would be refused if anything asked.

That is not a bug in the policy; it is the policy noticing something real. A
directive that steers the agent is currently published on a model's opinion
about its own run. What to do about it is a design decision, and the options
are at least:

1. Hold directives to measured evidence — a mined rule needs at least one
   `ObservationSource::ToolOutcome` behind it, not only reflection prose.
2. Reclassify: decide a human-reviewed advisory rule is `AdvisoryRecord`, not
   `SteeringDirective`, and say why the distinction holds.
3. Change what `SteeringDirective` costs, with the reasoning recorded.

Pick one deliberately. Silently lowering the requirement until the code agrees
is the expedient CLAUDE.md forbids.

## Definition of done

The conditions this issue already states, as boxes a gate can read. The
publication paths are split into three, because the skill and foundry halves
each need a decision this issue does not settle, and both are now filed with
the reasoning: #5895 (skill creation, blocked on a producer for
`TrajectoryAbstraction`, #5077) and #5896 (foundry enable, where three of the
four recorded authorities prove only `Agent` while an executable tool needs
`LocalHuman`).

- [x] `authorises` is called on every rule publication path: auto-activation,
      `stella proposals keep`, and `stella memory promote`.
- [x] Skill creation's half is decided and filed rather than left silent
      (#5895).
- [x] Foundry tool adoption's half is decided and filed rather than left
      silent (#5896).
- [x] A refusal is surfaced to the human rather than swallowed — a rule that
      cannot be published says which evidence it lacked
      (`PromotionRefusal::reason` already writes that line).
[...]

Issue #4862, opened 2026-08-25
Nothing records a memory citation any more, so memory quarantine and promotion can never fire


macanderson:
## Problem

`ObservationSource::MemoryCitation` has no producer in a real run, so the
citation half of the memory evidence loop is inert.

`crates/stella-cli/src/memory/evidence.rs` turns a **negative** memory citation
into an observation, and `crates/stella-store/src/efficacy.rs` +
`fold_citation_stats` derive quarantine
(`QUARANTINE_NEGATIVES_THRESHOLD`) and promotion eligibility
(`PROMOTION_CITATIONS_REQUIRED`) from stored citation rows. All of that reads
`memory_citations`. **Nothing writes it.**

Every caller of `Store::record_memory_citations`
(`crates/stella-store/src/lib.rs`) is inside a `#[cfg(test)] mod tests`:

- `crates/stella-store/src/tests.rs`
- `crates/stella-store/src/efficacy.rs` (its own test module)
- `crates/stella-cli/src/memory/tests.rs`
- `crates/stella-cli/src/memory/uses/tests.rs`
- `crates/stella-cli/src/memory_cmd.rs` (its own test module)
- `crates/stella-observatory/tests/schema_conformance.rs`

The writer used to be the agent-facing `cite_memory` tool. That tool is now in
`RETIRED_TOOL_NAMES` in `crates/stella-tools/src/catalog.rs`, so the model can
no longer call it and no other path replaced it.

## Why this is not #2195

#2195 ("`cite_memory` is registered on every surface and has never once been
called") is closed as completed. Reading the tree, it was settled by **retiring
the tool**, not by wiring a caller. The downstream consequence #2195 was filed
about is therefore still live: quarantine and citation-based promotion cannot
fire from real usage, and `memory_citation_stats` returns rows only in tests.

This matters more now than it did: #2782 grades evidence by its source, and a
source with no producer is a permanently empty evidence class.

## Repro / verify

```sh
rg -n "record_memory_citations" --type rust        # every hit is in a test
rg -n "cite_memory" crates/stella-tools/src/catalog.rs   # RETIRED_TOOL_NAMES
```

Then: run any real session that uses a memory and check the store —
`memory_citations` stays empty, and `stella memory list` shows no citation
counts moving.

## Definition of done

Decided on this thread: option 2, declare the gap rather than wire a
producer. No honest producer exists — every source that was checked only
knows a memory was shown to the model, never that the model judged it useful
[...]

Issue #4754, opened 2026-08-25
Unwired public API found by the #4392 audit: wire it, delete it, or keep it with a reason


macanderson:
Split out of #4392, which is an epic. The audit found public items shipping in
production crates with no caller anywhere in the workspace outside their own
tests. AGENTS.md § "Nothing left behind" requires each of those to name a
tracking issue; this is that issue, so the doc comments can cite it while the
decision is still open.

Each row is one decision: **wire it** (the caller it was written for), **delete
it**, or **keep it and say why in one sentence**. They are listed together
because the decision is the same shape each time, not because one PR has to
close all of them — a PR may close one row at a time by editing the doc comment
and dropping the row here.

## Rows

- [ ] `crates/stella-cli/src/memory/appraisals.rs` — `record_turn`, `sweep`,
      `record_appraisal`, `queued_candidates` and `Appraisal` carry
      `#[allow(dead_code)]`. They are the retirement half of the skill
      promote/retire gate (#1067, #1068) and need two seams that do not exist:
      a turn-end hook that knows both the offered and the selected skill sets,
      and the retirement sweep that writes the tombstone. Wiring those is the
      work; until then the allow is correct and now names this issue.
- [ ] `crates/stella-store/src/lib.rs` — `Store::list_pull_requests` is the
      read half of the `pull_requests` table. `Store::upsert_pull_request`
      writes rows from the deck's PR tracking; nothing reads them. It was
      written for a PR panel that does not exist.
- [ ] `crates/stella-protocol/src/recall.rs` — `NoContextRecall` is a
      null-object `ContextRecallPort` that nothing in this workspace
      instantiates (only `lib.rs`'s re-export and its own unit test reach it).
      It is plausible API for an external embedder driving `stella-engine`;
      decide whether that is the reason to keep it, and say so in its doc.
- [ ] `crates/stella-diag/src/dx.rs` — the `Facet` trait. `README.md` pitches
      it as the recommended way to log a whole vocabulary of events, and no
      production crate implements it: the heaviest consumer,
      `crates/stella-cli/src/diag_bridge.rs` (~20 emit sites), builds
      `Record::new` by hand.
- [ ] `crates/stella-diag/src/sink.rs` — `NullSink` and `CapturingSink` are
      exported for downstream use; `stella-cli` and `stella-tui` (the only
      crates depending on `stella-diag`) call neither. `CapturingSink` may want
      to be `#[cfg(test)]` or a `testing` feature rather than shipped API.
[...]

PR #4867, opened 2026-08-25
The evolution-surface ledger and an evidence grade that aggregation cannot launder (#2780, #2782)

### LINKED ISSUE #2780: Stella has six ways to change itself, and none of them require evidence, an approver, or a way to undo the change
## Problem

Stella now has three enforced declaration ledgers, each born from the same
defect shape — a capability that shipped on one axis and silently missed
another, discovered by a bench run paying for it rather than by a test:

- **Provider feature parity** — `crates/stella-model/src/provider_parity.rs`
  (AGENTS.md invariant #8). Every provider id declares a posture on every
  divergent axis and names the witness test proving it on the wire.
- **Cross-surface capability parity** — `crates/stella-parity/src/lib.rs`
  (`Capability`, `SurfacePosture`, `CAPABILITIES`, `UNWITNESSED_BASELINE`).
  Absence is legal only as `Deferred` (naming what it waits on) or
  `NotApplicable` (naming the design reason).
- **Signal consumption** — `crates/stella-protocol/src/event/consumers.rs`
  (invariant #10). Generated from the same table as `type_tag()`, so a variant
  with no declared consumer is a build error (`E0004`), not a red test.

**The self-improvement track has no equivalent, and it is the surface where the
gap costs the most.** Stella evolves itself through at least six distinct object
classes — `framework` (prompts, policies, the loop itself), `memory`, `skill`,
`tool`, `model` (weights/adapters), `workflow` — and each one currently answers
"when may this change itself, on what evidence, published by whom, and how is it
rolled back?" in a different file, or nowhere. Compare:

- Skills promote on measured lift (#1067) and were meant to decay and retire
  (#1068) — but efficacy-based retirement can never populate because
  `cite_memory` was never called (#2195), and there is still no end-to-end
  witness that a cited memory reaches `memory_citations` (#2431).
- Context records have an explicit governance mode, a hash-chained promotions
  ledger under `.stella/rules/`, and CI re-verification via
  `stella context validate`.
- Weight-space adapters (#836) and the bandit over prompts/routing (#831) were
  designed with a promotion gate (#876) that nothing structurally enforces.

Nobody can answer "what may Stella change about itself today, and what evidence
does each change require?" by reading one reviewed table. That is exactly the
question the other three ledgers exist to make answerable.

## Proposal

Add a code-owned evolution-surface matrix, following `stella-parity`'s shape
rather than inventing a fourth idiom. One row per evolution object, each
declaring:

| Field | Meaning |
|---|---|
| `surface` | `Framework`, `Memory`, `Skill`, `Tool`, `Model`, `Workflow` |
| `posture` | `Shipped`, `Experimental`, `Planned(issue)`, `Prohibited(reason)` |
| `timing` | when the change may take effect (in-turn, end-of-session, offline batch) |
| `evidence` | the accepted evidence grade for a change to this surface |
| `authority` | who or what may publish it (agent, local human, org policy) |
| `rollback` | the named artifact that reverses it |
| `witness` | the test proving the row, for any non-`Planned`/`Prohibited` posture |

The `evidence` column should reference the provenance grades from the
evidence-grade policy issue rather than restating them, so the two ledgers
cannot drift.

## Definition of done

Enforced from both sides, as the existing three are:

1. A new evolution surface with no row **fails the build or the gate** —
   prefer the `consumers.rs` treatment (generate the rows from the same table
   as the surface enum, so omission is `E0004`) over a test that can be
   forgotten.
2. A `Shipped` or `Experimental` row whose named witness test no longer exists
   fails a parity test, exactly as `stella-model`'s parity tests do today.
3. A `Planned` row cites a GitHub issue; a `Prohibited` row cites a stable
   design document by `id` (AGENTS.md: cite by id, never by path).
4. If any currently-shipped surface cannot be witnessed yet, it is a
   `ShippedUnwitnessed` row against a **down-only** baseline constant in the
   style of `UNWITNESSED_BASELINE` — recording debt that already exists, never
   granting new permission. Adding a row to that baseline to turn the gate
   green is the expedient CLAUDE.md forbids.

## Constraints for whoever picks this up

- Decide crate placement first. `stella-parity` is the closest neighbour and is
  currently a single `src/lib.rs`; a second matrix may belong beside it as a
  sibling module, or in its own crate if it needs dependencies `stella-parity`
  refuses. Name the exemplar you followed in the PR description.
- `stella-protocol` is types-only (invariant #1) and `stella-core` does no I/O
  (invariant #2) — a ledger that wants to read `.stella/` state belongs in
  neither.
- Read the "God files" table in AGENTS.md before planning where lines land.

## Related

Ledger exemplars: invariants #8 and #10 in AGENTS.md; #2707 (pinning a
whitelist to the signal-consumer ledger with a parity test); #2730.
Evolution surfaces this would cover: #836, #831, #876, #1067, #1068, #2195,
#2431, #2378.

---

## Sibling issues — one backlog batch, filed together

These five were filed as one set and are meaningfully coupled; picking up any
one in isolation is fine, but read this list first so the seams are not
reinvented.

- **#2780** — enforced evolution-surface parity ledger (declares *what* may
  evolve; its `evidence` column is defined by #2782)
- **#2781** — longitudinal transfer and decay panel (measures whether an
  artifact still helps elsewhere and later)
- **#2782** — evidence-grade preservation and promotion policy (defines what
  counts as evidence for each impact class)
- **#2783** — corpus information-gain and saturation gate (decides whether a
  training round is worth funding)
- **#2784** — bounded historical repository-memory experiment (prove-or-null,
  no default ingestion)

### LINKED ISSUE #2782: Stella loses track of how strong the evidence was as an idea turns into a published skill or tool, so weak evidence can end up looking as strong as a proven fix
## Problem

Stella's self-improvement pipeline turns observations into artifacts through a
long chain — reflection → proposal → skill / rule / tool → campaign — and
**the strength of the original evidence is not carried along the chain.** By the
time something is a published skill or an executable tool, the record that it
originated in a model's opinion rather than in a passing test is gone.

This is the same class of failure the repository already treats as its most
serious: a claim whose evidence cannot distinguish it from its opposite
(CLAUDE.md, "Every factual claim ships with the evidence that establishes it").
Applied to artifacts instead of sentences, the concrete hazard is
**laundering by aggregation**: several model critiques agreeing with each other
is still zero deterministic evidence, but a promotion gate that counts votes
will read it as strong. #2569/#2570 is the live proof that a plausible-looking
consensus signal can be structurally worthless — ablating the verifier turned
every run into a PASS.

## Proposal

Define an explicit, ordered provenance grade and carry it as a first-class
field through every derived artifact:

| Grade | Source |
|---|---|
| **Deterministic proof** | a witness test that went fail → pass; a guard that fails the gate |
| **Environment observation** | a command exit status, a build result, a measured number from `stella-events.jsonl` |
| **Trajectory abstraction** | a pattern mined across runs — statistical, not proven |
| **Model critique** | an LLM's judgement about a run |
| **Human review** | a person read it and signed off |

Two rules make the grade meaningful:

1. **No aggregation promotes a grade.** N model critiques agreeing remain a
   model critique. Only re-deriving the claim from a stronger source promotes
   it.
2. **Impact sets the required grade.** A prompt hint or a recall bias may be
   trialled from trajectory evidence. A **blocking guard** or an **executable
   tool** requires deterministic proof plus publication authority — because
   those two can break a teammate's session, and the blast radius is the thing
   the grade is rationing.

## Definition of done

1. Provenance survives every hop: reflection → proposal → skill/rule/tool →
   campaign. A hop that drops it fails a test.
2. No artifact can claim a grade stronger than its reproducible source. The
   check is mechanical, not editorial: the grade is derived from the source
   record, never asserted by the promoting code.
3. The grade is exposed where a human decides — the Observatory, `stella
   context explain` (#892), and any export path — so a reviewer sees "model
   critique" and "witness flip" as visibly different things.
4. A witness test proves rule 1 above: a fixture with several model critiques
   agreeing must not reach a grade that authorises tool publication.

## Constraints and prior art

- Context records already have the closest thing to this: a governance mode, a
  hash-chained promotions ledger (`.stella/rules/promotions.jsonl`), lifecycle
  events for retirement and supersession (#2728), and CI re-verification via
  `stella context validate`. **Extend that machinery outward rather than
  building a parallel one** — and say in the PR description why any new store
  was unavoidable.
- Promotion tiers shipped in #2762 (residue #2761) are the nearest live
  surface; read them before designing the grade enum.
- `stella-diag` field values cannot hold a `String`, a `Path`, or model output
  — that is a compile error, not a review question. A provenance grade emitted
  as a diagnostic must be an enum.
- Architecture rule 3 (zero telemetry egress) constrains what may leave the machine:
  a provenance record is metadata, but the evidence it points at is content.
  Anything crossing an egress encoder goes through
  `crates/stella-store/src/content_free.rs`.

## Related

#892 (rule provenance in agent output), #2723 (ToolCtx provenance, `proof_ref`
citation), #2709/#2762 (promotion tiers), #2761, #2728, #2569/#2570 (why vote
counting is not evidence), #2023. Feeds the `evidence` column of the
evolution-surface parity ledger.

---

## Sibling issues — one backlog batch, filed together

These five were filed as one set and are meaningfully coupled; picking up any
one in isolation is fine, but read this list first so the seams are not
reinvented.

- **#2780** — enforced evolution-surface parity ledger (declares *what* may
  evolve; its `evidence` column is defined by #2782)
- **#2781** — longitudinal transfer and decay panel (measures whether an
  artifact still helps elsewhere and later)
- **#2782** — evidence-grade preservation and promotion policy (defines what
  counts as evidence for each impact class)
- **#2783** — corpus information-gain and saturation gate (decides whether a
  training round is worth funding)
- **#2784** — bounded historical repository-memory experiment (prove-or-null,
  no default ingestion)

macanderson:
Closes #2780. Closes #2782.

The two open issues in the Self-Improvement milestones. One PR because #2780's
`evidence` column is defined by #2782 — shipping either alone leaves the other
referencing something that does not exist.

**Exemplar followed:** `crates/stella-parity/src/lib.rs` for the matrix shape
(posture enum, named witnesses checked by substring against `include_str!`ed
sources, exact-equality down-only baseline), and
`crates/stella-protocol/src/event/tags.rs` for generating a ledger and its enum
from one macro table.

## #2782 — the evidence grade

`ProvenanceGrade` is ordered weakest-first, so the derived `Ord` **is** the
strength order and combining evidence is a `min`. That is the whole of "no
aggregation promotes a grade": no `promote`, no vote count, no confidence
threshold that crosses a grade boundary. Only re-deriving a claim against a
stronger source moves it, which means building a new record.

Evidence and publication authority are two axes rather than one ladder, because
#2782's own rule says a guard or a tool needs "deterministic proof **plus**
publication authority". Collapsing them would let an approver's signature stand
in for a witness test.

**Derived, never asserted** is enforced structurally, twice:

- An observation stores no grade. `observation_grade` is a total function of
  `ObservationSource`, so there is no second copy to drift.
- A proposal must store one, because it keeps only its observations'
  `record_id`s. `EvidencePool` is the only way to populate that field and its
  only public constructor takes real `ObservationRecord`s — promoting code can
  hand a proposal the evidence it has, not a grade it prefers.
  `ProposalRecord::new` takes the pool in place of a `Vec<String>`, making the
  old shape unwritable rather than discouraged.

### Where provenance goes, exactly

| Hop | Status |
|---|---|
| reflection / tool outcome → observation | derived from the source, never stored |
| observation → proposal | carried (`EvidencePool`) |
| proposal → published **rule** record | carried (`evidence_grade` on `Provenance`) |
| proposal → published **skill** file | **not carried** — #4871, see below |
| → campaign | no such hop exists; `rg -ni "\bcampaign" --type rust` is empty |

Absent grades stay absent on the wire, so records already on disk keep their
bytes and their hashes.

**Surfaced where a human decides:** `stella proposals`, the Observatory's
proposal projection, and `stella context explain` — which prints the grade
beside a record's other provenance, so a rule inferred from reflection and one
derived from a failing build are visibly different.

### Why the skill file is excluded (#4871)

I implemented it and backed it out. `both_loops_render_byte_identical_skill_files`
requires the lexical loop (`context.lifecycle.enabled = false`, **the path that
ships today**) and the opt-in typed loop to render identical `SKILL.md` bytes
from the same evidence. Only the typed loop has a proposal, so only it has a
grade; an `evidence_grade:` frontmatter line makes them diverge, and the harm is
the one that test names — a user turning the lifecycle on would see every skill
file diff against its own re-render. That guarantee outranks a frontmatter key.

## #2780 — the evolution-surface matrix

`EvolutionSurface` and `EVOLUTION_SURFACES` are generated from one macro table,
so a surface and its row are the same cell.

**DoD 1 is demonstrated, not asserted.** I added a seventh surface omitting one
column and built it: the compiler stops at ``error: no rules expected `;` ``
before any test runs, then I reverted it. `consumers.rs` reaches `E0004` because
omission breaks an exhaustive match; here there is no hand-written enum to omit
from at all.

**The evidence column is read from #2782, not restated.** A row declares its
`ImpactClass`; the required grade and authority are derived through the policy.
No second copy of the requirement, so the two ledgers cannot disagree
(`required_evidence_is_read_from_the_provenance_policy`).

Rows record what is true rather than what looks finished:

| Surface | Posture | Recorded honestly |
|---|---|---|
| Framework | Shipped | no command retracts a published rule (#4866) |
| Memory | Shipped | — |
| Skill | Shipped | measured-lift gate is **off by default**; demotion half has no production caller (#4754) |
| Tool | Shipped | — |
| Workflow | Shipped | — |
| Model | NotPursued | nothing is promoted; export has no in-tree consumer |

`UNWITNESSED_EVOLUTION_BASELINE` is 0. All six witnesses were verified to exist
before being named.

## Three deviations, each deliberate — overrule any of them

1. **`HumanReview` is not the strongest grade.** #2782's table lists it last. I
   ranked it above `ModelCritique` (a judgement someone is accountable for
   outranks one nobody is) and below `TrajectoryAbstraction` (a measured pattern
   survives its author's confidence). One line to change.
2. **A sixth posture, `NotPursued`.** Weight adaptation was designed, filed, and
   closed as *not planned* (#836). `Planned` would be false; `Prohibited` wants
   a design document nobody wrote.
3. **`BetweenTurns` rather than "end-of-session".** No live surface waits for the
   session to end; memory and skill both land on the post-turn reflection path.

## What this does NOT do

**No live publication path calls the gate.** The ledger declares the terms and
the policy can answer them; nothing asks before publishing. AGENTS.md rule #11
says so in the shape rule #10 uses, and #4865 tracks the wiring.

#4865 also records a tension the policy surfaced: the framework row requires
`EnvironmentObservation`, and mined rules grade to `ModelCritique` — so a
directive that steers the agent is published today on a model's opinion of its
own run. Resolving that is a design decision, not something to fix by lowering
the requirement until the code agrees.

## Also filed

- **#4862** — `record_memory_citations` has no production caller; every one is a
  test. The writer was the `cite_memory` tool, now in `RETIRED_TOOL_NAMES`, so
  #2195 was closed by retiring the tool and the consequence it was filed about
  is still live: `ObservationSource::MemoryCitation` is a permanently empty
  evidence class.
- **#4865** — wire the gate into publication.
- **#4866** — no command retracts a published rule.
- **#4871** — a published skill does not say what evidence promoted it.

## Evidence

- `cargo test -p stella-protocol` — 16 new, including
  `agreeing_model_critiques_never_authorise_an_executable_tool` (#2782 DoD 4).
- `cargo test -p stella-core context_record::provenance` — 7 new, including
  `three_agreeing_reflections_across_three_tasks_still_cannot_publish_a_tool`,
  which clears the existing distinct-task floor first so the refusal is the
  grade's doing and not the floor's.
- `cargo test -p stella-parity` — 23 pass (12 new), including a negative control
  proving the witness check can go red.
- `a_published_rule_carries_the_grade_of_the_proposal_it_came_from`,
  `explain_names_the_grade_of_the_evidence_behind_a_record`.
- Full `cargo test -p stella-cli -p stella-core --no-fail-fast` — all binaries
  green. (Filtered runs are what let the byte-identity failure through once.)
- `make guards-fast`, `make invariants` (11 rules, 40 citations resolve),
  `make prose` (**none added** — the constructions the ratchet found were real
  and were deleted, not baselined), `make line-citations` — green.
- CI: all 16 checks green on `750253a`.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -575,6 +575,37 @@ Append; do not renumber. `scripts/check-invariants.sh` enforces both halves.
    totality (by the compiler) plus issue citation and posture coherence (by
    test) — enough to make a PR author answer "what reads this?" before the
    merge.
+11. **Every way Stella changes itself is declared.** Stella evolves through
+   several object classes — framework, memory, skill, tool, workflow, model —
+   and each one used to answer "when may this change, on what evidence,
+   published by whom, and how is it undone?" in a different file or nowhere.
+   The ledger is
+   `crates/stella-parity/src/evolution.rs`: one row per surface, declaring a
+   posture, a timing, an impact class, a rollback artifact, and — for a live
+   posture — the witness test proving the surface can actually be changed.
+   `EvolutionSurface` and its rows are generated from one macro table, so a
+   surface with no row does not compile.
+
+   The `evidence` and `authority` columns are **not stored**. A row declares
+   what a wrong change to it can break (`ImpactClass`), and the required
+   provenance grade and publication authority are read out of
+   `crates/stella-protocol/src/provenance.rs`. One policy, read in two places,
+   with no second copy to drift (#2780, #2782).
+
+   That policy is the other half of this rule, and it says that
+   **aggregation never promotes evidence**. N model critiques agreeing remain a
+   model critique; only re-deriving a claim against a stronger source moves it.
+   A prompt hint may be trialled from a mined trajectory, and a blocking guard
+   or an executable tool needs deterministic proof plus a person — because
+   those two break a teammate's session, and blast radius is what the grade
+   rations. #2569/#2570 is why vote-counting is not an alternative: ablating
+   the verifier turned every run into a PASS and the count never noticed.
+
+   What it does **not** prove: that a live path actually asks the gate before
+   publishing. The ledger declares the terms and the policy can answer them;
+   wiring `authorises` into each publication path is tracked separately, and
+   the rows say plainly where today's evidence falls short of what they
+   require.
 ---
 
 ## The definition of done: witness tests
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3165,6 +3165,7 @@ dependencies = [
 name = "stella-parity"
 version = "0.9.215"
 dependencies = [
+ "stella-protocol",
  "stella-serve",
 ]
 
--- a/crates/stella-cli/src/context_cmd/explain.rs
+++ b/crates/stella-cli/src/context_cmd/explain.rs
@@ -112,6 +112,9 @@ pub fn run_explain(root: &Path, needle: &str) -> Result<(), String> {
         if let Some(run) = provenance.ingest_run_id.as_deref() {
             println!("    {}", format!("ingest run {run}").dimmed());
         }
+        if let Some(line) = evidence_line(provenance.evidence_grade) {
+            println!("    {}", line.dimmed());
+        }
     }
     if let Some(id) = record.record_id.as_deref() {
         println!("    {}", format!("record {id}").dimmed());
@@ -234,6 +237,20 @@ pub fn run_explain(root: &Path, needle: &str) -> Result<(), String> {
 /// The `selected` → `rendered` → `cited` funnel is recorded by the session that
 /// rendered the record; a workspace that has not run a session since the record was
 /// published has nothing to report, which is different from a record nobody used.
+/// The evidence line for a record's provenance grade (#2782), or [`None`] when
+/// the record carries no grade.
+///
+/// Absent is the common case and says nothing rather than implying weakness: a
+/// grade is stamped on records Stella *induced* from a run, and most records
+/// are ingested from a document instead. Extracted from the rendering so the
+/// wording can be asserted — `run_explain` prints to stdout, so a test of what
+/// it renders has nowhere else to look.
+pub(super) fn evidence_line(
+    grade: Option<stella_protocol::provenance::ProvenanceGrade>,
+) -> Option<String> {
+    grade.map(|grade| format!("evidence: {} — {}", grade.as_str(), grade.describe()))
+}
+
 fn efficacy_line(root: &Path, handle: &str) -> String {
     let ledger = root.join(".stella/private/context.db");
     if !ledger.exists() {
--- a/crates/stella-cli/src/context_cmd/tests.rs
+++ b/crates/stella-cli/src/context_cmd/tests.rs
@@ -619,6 +619,39 @@ fn explain_reproduces_a_kept_records_rule_and_provenance() {
     explain::run_explain(root.path(), "ctx.acme.web.pkg-manager").expect("explain by lineage");
 }
 
+/// #2782 — `stella context explain` distinguishes a rule Stella inferred from
+/// its own reflection from one it derived from a failing build.
+///
+/// The surrounding provenance lines say *where* a record came from. This one
+/// says what that source was worth, which is the question a reviewer deciding
+/// whether to trust an induced rule is actually asking.
+#[test]
+fn explain_names_the_grade_of_the_evidence_behind_a_record() {
+    use stella_protocol::provenance::ProvenanceGrade;
+
+    let critique = explain::evidence_line(Some(ProvenanceGrade::ModelCritique))
+        .expect("a graded record renders a line");
+    assert!(critique.contains("model_critique"), "{critique}");
+    assert!(
+        critique.contains("judgement"),
+        "the line must say what the grade means, not only name it: {critique}"
+    );
+
+    let proof = explain::evidence_line(Some(ProvenanceGrade::DeterministicProof))
+        .expect("a graded record renders a line");
+    assert!(proof.contains("deterministic_proof"), "{proof}");
+    assert_ne!(
+        critique, proof,
+        "the two grades must be visibly different where a human decides"
+    );
+
+    assert_eq!(
+        explain::evidence_line(None),
+        None,
+        "a record ingested from a document carries no grade and must claim none"
+    );
+}
+
 #[test]
 fn explain_on_an_unknown_rule_lists_what_is_loaded() {
     let root = workspace();
--- a/crates/stella-cli/src/context_records.rs
+++ b/crates/stella-cli/src/context_records.rs
@@ -805,6 +805,7 @@ pub(crate) fn inferred_rule_record(
     statement: &str,
     source_kind: &str,
     source_uri: &str,
+    evidence_grade: Option<stella_protocol::provenance::ProvenanceGrade>,
 ) -> Result<stella_core::ingest::record::Record, String> {
     use stella_core::context_record::{Origin, RecordStatus};
     use stella_core::ingest::record as rec;
@@ -844,6 +845,10 @@ pub(crate) fn inferred_rule_record(
         provenance: Some(rec::Provenance {
             source_kind: Some(source_kind.to_string()),
             source_uri: Some(source_uri.to_string()),
+            // Carried from the proposal being published (#2782). Once this
+            // record is on disk the observations behind it are out of reach,
+            // so a grade not carried here is unrecoverable.
+            evidence_grade,
             ..Default::default()
         }),
         steering: Some(rec::Steering {
--- a/crates/stella-cli/src/ingest_cmd/extract.rs
+++ b/crates/stella-cli/src/ingest_cmd/extract.rs
@@ -636,6 +636,9 @@ fn build_defaults(
             commit: git(root, &["rev-parse", "--short", "HEAD"]),
             ingest_run_id: Some(ingest_run_id.to_string()),
             extracted_at: Some(extracted_at.to_string()),
+            // Ingested from a document rather than induced from a run, so
+            // there is no run-evidence grade to carry (#2782).
+            evidence_grade: None,
             extractor: Some(extractor.to_string()),
         }),
     }
--- a/crates/stella-cli/src/memory/learning.rs
+++ b/crates/stella-cli/src/memory/learning.rs
@@ -15,6 +15,7 @@ use stella_core::skills::{
     self, AutoCreateConfig, AutoCreateDecision, SkillMineConfig, SkillObservation,
 };
 use stella_protocol::Provider;
+use stella_protocol::provenance::ProvenanceGrade;
 
 use super::{
     LessonKind, ReflectionLesson, ReflectionReport, SessionMemory, TurnEvidence, reflect_on_turn,
@@ -590,7 +591,18 @@ impl SessionMemory {
     /// Shared by both paths, so the cap and the guard are literally the same
     /// code on each — asserted by test rather than by inspection, which is what
     /// spec §8 asks for.
-    fn write_candidates(&mut self, candidates: Vec<skills::SkillCandidate>, quiet: bool) {
+    /// Write the eligible skill candidates, each with the evidence grade of
+    /// the proposal it was promoted from (#2782).
+    ///
+    /// The grade rides alongside the candidate rather than inside it because a
+    /// `SkillCandidate` is the miner's output and knows nothing about the
+    /// lifecycle ledger — the proposal is where the two meet, and this is the
+    /// hop that used to drop it.
+    fn write_candidates(
+        &mut self,
+        candidates: Vec<(skills::SkillCandidate, Option<ProvenanceGrade>)>,
+        quiet: bool,
+    ) {
         // The workspace-skills authority gate, on the write it actually
         // describes. Without it the loader is handed an empty workspace dir, so
         // a skill written here would never be read back — and computing the
@@ -624,7 +636,7 @@ impl SessionMemory {
         // ledger rather than from this turn — one turn cannot measure a skill
         // that has never been injected.
         let verdicts = super::appraisals::latest_verdicts(&self.workspace_root);
-        for candidate in candidates {
+        for (candidate, evidence_grade) in candidates {
             let evidence = verdicts
                 .get(&candidate.name)
                 .copied()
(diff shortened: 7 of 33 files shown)


macanderson:
## Answering Sourcery's ❌ row

> **#2782** — Carry provenance through the complete artifact lifecycle from reflection and proposal through skills, rules, tools, and campaigns, with publication gates requiring evidence and authority appropriate to the artifact's impact.

The row was right, and it split into three claims with three different answers.
Two are now fixed; one is answered.

### Fixed — provenance stopped at the proposal

It did, and that was the more serious half: a proposal keeps its observations'
`record_id`s, but a **published** rule or skill keeps neither, so a grade not
written at publication is unrecoverable by anything downstream — including the
human reading the rule file to decide whether to trust it.

`3da8c5a` carries it across that hop:

- `Provenance` grows an `evidence_grade`, written onto the record `write_rule`
  publishes.
- `render_skill_markdown` writes an `evidence_grade` frontmatter line.
- `write_candidates` now takes each candidate beside the grade of the proposal
  it came from. The skills path previously did `.map(|i| i.candidate)` — that
  line was the hop dropping the evidence.
- Memory promotion grades as `ModelCritique` however long the citation streak,
  because a citation is the model judging that a memory helped and counting
  them is aggregation.

Both fields are optional and omitted when absent, so records already on disk
keep their bytes and their hashes.

Witnesses: `a_published_rule_carries_the_grade_of_the_proposal_it_came_from`,
`a_skill_written_from_a_graded_proposal_records_the_grade`.

### Answered — "campaigns" is not a hop that exists

`rg -ni "\bcampaign" --type rust` returns zero hits. There is no `Campaign`
type anywhere in the workspace; the word appears only in the issue text and the
paper it came from. Provenance cannot be carried through a hop that has no
code, and inventing one to satisfy the row would be worse than saying so.

### Answered — the publication gate is deliberately deferred, and tracked

`authorises` still is not called before publishing. That is stated in the PR
body and, more durably, in AGENTS.md rule #11's own "what it does not prove"
paragraph, in the same shape rule #10 uses. #4865 tracks the wiring.

It is deferred rather than done because wiring it requires a decision this PR
should not make alone. The framework row declares `SteeringDirective`, which
the policy prices at `EnvironmentObservation`; mined rules grade to
`ModelCritique`, two rungs below. So a directive that steers the agent is
published today on a model's opinion of its own run. The options — hold
directives to measured evidence, reclassify them as advisory, or reprice
`SteeringDirective` — are recorded in #4865. Lowering the requirement until the
code agrees is the one option that is not on the table.

**Net:** provenance now survives every hop that exists, and the gap that
remains is enforcement rather than preservation.

macanderson:
## Correction to my previous comment: the skill half is backed out

My comment above said the grade was carried onto published rules **and**
skills. CI disagreed, and it was right.

`memory::learning::guarantees::both_loops_render_byte_identical_skill_files`
requires the lexical loop and the typed loop to render byte-identical
`SKILL.md` files from the same evidence. The lexical loop is
`context.lifecycle.enabled = false` — **the path that ships today** — and only
the typed loop has a proposal behind it, so only it has a grade. Adding an
`evidence_grade:` frontmatter line made the two diverge, and the harm is
exactly the one that test names: a user who turns the lifecycle on would see
every existing skill file diff against its own re-render.

That guarantee outranks a frontmatter key while the lexical loop is still the
default, so `c3662e5` backs the skill half out and files **#4871** with the two
ways to do it properly — after the lexical loop retires, or out of band without
touching the file bytes.

### What the branch actually carries now

| Hop | Status |
|---|---|
| reflection/tool outcome → observation | grade derived from the source, never stored |
| observation → proposal | carried, via `EvidencePool` |
| proposal → published **rule** record | carried (`evidence_grade` on `Provenance`) |
| proposal → published **skill** | **not carried** — #4871 |
| → campaign | no such hop exists in the tree |

So provenance survives every hop that exists **except** the skill file, and
that exception is now a filed issue rather than a claim in a PR description.

I also ran the full `cargo test -p stella-cli -p stella-core` (29 binaries,
green) rather than the filtered subsets that let this through in the first
place.

macanderson:
## Answering Sourcery's second ❌ row — now fixed

> **#2782** — Expose provenance at all human decision and export surfaces, including the Observatory, `stella context explain`, and relevant artifact exports.

Fixed in `750253a`. `stella context explain` now prints an evidence line beside
the rest of a record's provenance, so a rule Stella inferred from its own
reflection and one it derived from a failing build are visibly different where
a reviewer decides.

This only became possible in `3da8c5a`. When I first wrote the PR body saying
`explain` was "deliberately untouched", that was accurate — it explains
*published records*, which carried no grade. Carrying the grade onto the
published rule record is what gave `explain` something to read.

The wording is built by a small `evidence_line` function rather than inlined,
because `run_explain` prints to stdout and the existing `explain` tests only
assert `Ok(())` — there was nowhere to assert what it renders. Witness:
`explain_names_the_grade_of_the_evidence_behind_a_record`, which pins that the
two grades render differently and that an ungraded record claims nothing.

**Export paths:** `evidence_grade` is a `serde` field on `Provenance`, so every
path that serializes a `Record` — the published TOML, the JCS hash preimage,
any JSON projection — carries it without a per-path change. I did not find an
export that hand-builds record fields and would need updating; if you know of
one, name it and I will cover it.

### Note on the first ❌ row — the table is stale

Sourcery's assessment comment is timestamped `2026-08-25T20:16:54Z`, which
predates all four of my pushes (`20:34`, `20:42`, `20:58`, and this one), and
its check shows `skipping` rather than re-running. So that row was written
against the first commit and does not reflect `3da8c5a` (rules carry the grade)
or `c3662e5` (the skill half backed out, filed as #4871).

The current, honest state of that objective is in my comment above: provenance
survives every hop that exists except the skill file, which is #4871; the
publication gate remains deferred to #4865; and "campaigns" is not a hop that
exists in the tree.

Issue #4698, opened 2026-08-24
On Windows a hook, custom tool or bash call that backgrounds work has no group-kill test


macanderson:
## Problem

#4456 gave `exec::GroupKillGuard` a Windows Job Object arm, and #4675 proved it for the plugin transport on `windows-latest`. The other three spawn planes have no Windows witness at all: `crates/stella-tools/src/bash.rs:865 a_dropped_bash_call_kills_the_process_group` and `custom/tests/execution.rs:266 a_dropped_custom_tool_kills_the_process_group` are both `#[cfg(unix)]` and read a pid with `ps`; `hook_runner.rs` has no group-kill test. All three call the same guard, so on Windows a hook, a custom tool or a `bash` call that backgrounds work is bounded by nothing but `kill_on_drop` as far as any test knows.

## Done means

Those three assertions run on `windows-check.yml`. The portable shape already exists: `crates/stella-runtime/tests/wrapper_transport_limits.rs` heartbeats to a file from a thread that touches no pipe and watches it stop growing (`ps -o state=` has no Windows equivalent — a SIGKILLed orphan is a zombie there). Each ported test must assert its premise so it cannot pass vacuously.

Not in #3550's definition of done, which named the plugin case. From PR #4675's report (2026-08-24).

## Definition of done

The “Done means” prose above, as boxes the gate can read — nothing here is
new work, and nothing is ticked by the migration (#5193).

- [ ] The three assertions run on `windows-check.yml`.
- [ ] Each ported test asserts its premise, so it cannot pass vacuously.
- [ ] They follow the portable shape `crates/stella-runtime/tests/wrapper_transport_limits.rs` already uses — heartbeating to a file from a thread that touches no pipe — because `ps -o state=` has no Windows equivalent.

macanderson:
PR #4856 ports all three assertions to `crates/stella-tools/tests/group_kill_witnesses.rs`, an integration test so it can run on `windows-check.yml` without waiting on the crate's `--lib` test binary compiling there (a separate gap, #4697).

[...]

PR #4878, opened 2026-08-25
test(stella-runtime): assert the witness declaration is per-stage, not just its union

### LINKED ISSUE #4876: The declared-witness tests cannot tell a per-stage declaration from a constant one
## Problem

`crates/stella-runtime/tests/wrapper_declared_witness.rs` has two tests, and both assert the **union** of the witness lists across a round's stages. The union is invariant to the thing the file is about:

```
per-stage:  {flip.rs} ∪ {flip.rs, second.rs}          = {flip.rs, second.rs}
constant:   {flip.rs, second.rs} ∪ {flip.rs, second.rs} = {flip.rs, second.rs}
```

So a plugin that declares the widest list at every stage is indistinguishable from one that narrows at `research`. What these tests establish is that the union is folded and de-duplicated in first-seen order — a real property, and not the per-stage one.

## How it was found

Ablation, during the #4697 port (PR #4874). Pinning the fixture's `witness-by-stage` list to a constant leaves **both tests green**. The `/bin/sh` plugin the fixture replaced carried the same branch and was equally unobservable, so this predates the port — the port just made it checkable.

## Done means

A test that reads **one stage's** declared list on its own, so a plugin declaring the wide list everywhere fails it. The fixture already has the branch (`witness-by-stage` narrows at `research`), so this needs the assertion rather than new plumbing: drive a round whose stage is `research` and assert the host was handed `["tests/flip.rs"]` alone.

Then the ablation check flips the right way — pinning the list should fail that test.

## Constraint

Do not weaken the existing union test to get there; folding and de-duplication are worth keeping asserted. This is an additional case, not a replacement.

Found 2026-08-25. Related: #4697, #3497.

macanderson:
Closes the gap PR #4874's ablation exposed, rather than leaving it filed.

## The gap

`wrapper_declared_witness.rs` had two tests and both asserted the **union** of witness lists across a round's stages. That union is invariant to the property the file exists to check:

```
per-stage:  {flip.rs} ∪ {flip.rs, second.rs}           = {flip.rs, second.rs}
constant:   {flip.rs, second.rs} ∪ {flip.rs, second.rs} = {flip.rs, second.rs}
```

So a plugin declaring the widest list at every stage passed every assertion. The `/bin/sh` script the fixture replaced had the same branch and was equally unobservable — the port did not introduce this, it made it checkable.

## The fix

A one-stage manifest and a test that reads that stage's declaration alone. With one stage there is no union to hide behind, so the narrowing has to show.

The ablation now flips the right way — pinning the fixture's list to the wide one:

```
test a_single_stage_round_is_handed_that_stages_own_declaration ... FAILED
test result: FAILED. 2 passed; 1 failed
```

It fails **only** the new test, which is the point: the union test is genuinely about folding and de-duplication, and it still passes because that property is still true.

## What I did not do

I did not weaken or replace the union test. Folding and first-seen ordering are worth keeping asserted; this is an added case. The issue named that constraint and it is honoured.

I also corrected the two doc comments PR #4874 left behind. They said the narrowing was unobservable and named a hypothetical test that would settle it — now they name the test that does.

## Evidence

```
cargo test -p stella-runtime --test wrapper_declared_witness   3 passed; 0 failed
  ablated fixture branch:                                      2 passed; 1 failed
cargo clippy -p stella-runtime --all-targets -- -D warnings    exit 0
cargo fmt --all --check   exit 0
check-prose               OK — none added
```

Closes #4876

## Summary by Sourcery

Verify that wrapper witness declarations remain stage-specific instead of being observable only as a folded union.

Enhancements:
- Add coverage proving wrapper witness declarations are preserved per stage rather than only validated through their cross-stage union.

Documentation:
- Update fixture and test documentation to describe the newly observable per-stage witness assertion.

Tests:
- Add a one-stage wrapper test that verifies the host receives that stage’s narrowed witness declaration while retaining the existing union folding and de-duplication coverage.

### DIFF
--- a/crates/stella-runtime/tests/fixtures/wrapper-plugin-fixture.rs
+++ b/crates/stella-runtime/tests/fixtures/wrapper-plugin-fixture.rs
@@ -134,11 +134,11 @@ fn main() {
         // granted late, one never granted, so a mode that answered a
         // constant could not tell a delivered credential from a leaked one.
         // Declares a witness list that narrows at `research` — the `case`
-        // the shell plugin spelled out. Kept faithful, but note the
-        // narrowing is unobservable through `wrapper_declared_witness.rs`:
-        // it asserts the *union* across stages, which is the same set
-        // either way. Checked by ablation — pinning the list leaves both
-        // its tests green.
+        // the shell plugin spelled out. The narrowing is load-bearing:
+        // pinning the list fails
+        // `a_single_stage_round_is_handed_that_stages_own_declaration`
+        // (#4876). It is invisible to that file's union test, which is why
+        // the single-stage case had to be added to see it.
         "witness-by-stage" => {
             let request = read_request();
             let list = if request.contains("\"stage\":\"research\"") {
--- a/crates/stella-runtime/tests/wrapper_declared_witness.rs
+++ b/crates/stella-runtime/tests/wrapper_declared_witness.rs
@@ -21,12 +21,13 @@
 //! this file runs on Windows too (#4697). Its `witness-by-stage` mode carries
 //! the `case` the script spelled out: the declared list narrows at `research`.
 //!
-//! That narrowing is **not** observable through the tests below, and the
-//! shell script it replaces was no better: the assertion is on the *union*
-//! across stages, which is the same set whether each stage declares its own
-//! list or every stage declares the widest one. So these tests prove the
-//! union is folded and de-duplicated, not that the declaration is per-stage.
-//! A test that reads one stage's list on its own would settle it.
+//! The union test alone cannot see that narrowing: its assertion is the
+//! *union* across stages, which is the same set whether each stage declares
+//! its own list or every stage declares the widest one. So it proves the
+//! union is folded and de-duplicated, and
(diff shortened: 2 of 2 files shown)


macanderson:
Merging with one check red, and naming what that costs.

`the platform split still compiles on Windows` is **failing on `main`** — `gh run list --workflow windows-check.yml --branch main` shows failures at `61945fa32` and `7c8364cfe` — on a step this diff cannot reach: the group-kill guard at the bash, custom-tool and hook spawn sites (#4698). This PR touches one file, `crates/stella-runtime/tests/wrapper_declared_witness.rs`. Reported with evidence on #4698 rather than worked around here.

**The limitation this creates:** that job is where `wrapper_declared_witness` runs on Windows, so while the job dies at an earlier step, the new test's Windows result is unknown. It passes on macOS, and its ablation flips correctly there. I am not claiming more than that.

`windows-check.yml` is deliberately not a required check, so this does not gate the merge — but a persistently red job hides every result inside it, which is worth fixing sooner rather than later.

Issue #4894, opened 2026-08-26
Two public field-sets in the harbor launcher are declared but read by nothing


macanderson:
## Problem

Two module-level frozensets in `bench/harbor_adapter/stella_harbor/secure_launcher.py`
are read by nothing, in this repository or the ejected arenabench one:

- `PRIOR_STAGE_OUTCOME_FIELDS` — `{stage, intent_sha256, status, completed_at, recorded_at}`
- `PROVIDER_KEY_LIVE_SNAPSHOT_FIELDS` — the eleven-key provider-credit snapshot shape

Both have been unread since the harness was created (`49111dc8a`, #301). Confirmed with
an `ast` walk over the module (no `Load` of either name) plus
`rg 'PRIOR_STAGE_OUTCOME_FIELDS|PROVIDER_KEY_LIVE_SNAPSHOT_FIELDS' bench scripts` and a
GitHub code search over `macanderson/arenabench`, which returns zero for each.

This is the same defect #4668 fixed for `_ENGINE_POSTURE_FIELDS`: a field-set that reads
like the launcher's fail-closed schema for a payload and is not one. It is filed
separately because these two are **public** names — deleting one is an API change for any
out-of-tree importer, which the private posture sets could not have — and because the
answer may be "wire them" rather than "delete them": both describe payloads the launcher
does parse (the prior-stage outcome records in `_validate_run_ledger_schema`'s
neighbourhood, and the provider-key credits object at `secure_launcher.py`'s
`_require_exact_object(credits_data, ...)` call sites), so either could be the schema
that call was meant to be handed.

## Why the guard did not catch them

`bench/harbor_adapter/tests/test_posture.py::TestLauncherDeclaresNoVocabularyNothingReads`
(added by #4668's fix) scopes itself to **private** names, because a private name has no
importer outside the file by construction and so needs no allow-list. Widening it to
public names is the second half of this issue, and is only sound once these two are
settled.

## Steps

1. For each constant, read the parse site it plausibly belongs to (`rg -n
   '_require_exact_object' bench/harbor_adapter/stella_harbor/secure_launcher.py`) and
   decide: does that call already compare the payload exactly by some other means, the way
   `_validate_confirmatory_manifest_contract` rebuilds the engine posture whole? If yes,
   the set is redundant and goes. If no, hand it to the parse site.
2. Whichever way each goes, widen `_unread_private_module_constants` in
   `tests/test_posture.py` to cover public module-level names too, and drop the "private"
[...]

PR #4895, opened 2026-08-26
fix(bench): delete the launcher posture vocabulary nothing read

### LINKED ISSUE #4668: _ENGINE_POSTURE_FIELDS in the harbor adapter is declared but nothing reads it
## Problem

`bench/harbor_adapter/stella_harbor/secure_launcher.py`'s `_ENGINE_POSTURE_FIELDS` is declared and referenced nowhere; the real fail-closed check is `ENGINE_ROOT_FIELDS` parsed from `unknown.rs`. PR #4665 corrected its vocabulary (added `minimal_prompt`, removed the stale `params` entry that #2411 deleted) but did not wire it, because wiring is not mechanical: `_require_exact_object` would refuse every posture that omits an optional arm key, so it needs a required/optional split like `tb21_posture_schema.py`'s.

## Constraint

`secure_launcher.py` sits on its size ceiling (4383 in `scripts/file-size-baseline.txt`), so any wiring must pay for its own lines — split rather than grow.

## Done means

Either the set is wired with a required/optional split and a test that a posture omitting an optional key is admitted while an unknown key is refused, or the declaration is deleted and the fail-closed check documented as `ENGINE_ROOT_FIELDS` alone. Unwired vocabulary is the shape this repo files rather than leaves.

From PR #4665's report (2026-08-24). Related: #4650, #3879.

macanderson:
## What & why

`bench/harbor_adapter/stella_harbor/secure_launcher.py` declared three posture field
sets that nothing referenced: `_ENGINE_POSTURE_FIELDS`, `_ENGINE_POSTURE_AGENT_ROLES`,
`_ENGINE_POSTURE_AGENT_FIELDS`. #4668 gives two acceptable endings — wire them with a
required/optional split, or delete them and document the check that actually runs. This
takes the second, because wiring them would have been **weaker than what already runs**:

- `_validate_confirmatory_manifest_contract` builds the expected posture with the
  adapter's own `_benchmark_engine_posture` and compares the whole dict — once for the
  SUT record (`expected_sut["engine_posture"] = primary_posture[0]`, compared by
  `sut != expected_sut`) and once per calibration config
  (`expected_calibration_fixed["engine_postures_by_config"]`, compared by the
  `calibration.get(key) != value` sweep). Exact dict equality over the root keys, the
  `agents` map and every role row. A key-name vocabulary beside that can only agree
  with it.
- What the *trusted launcher* will accept is `settings::ENGINE_ROOT_FIELDS` in
  `unknown.rs`, read from the Rust rather than copied
  (`tests/test_posture.py::_engine_root_fields`).
- What a *recorded* posture must look like is `tb21_posture_schema.py`, which splits
  required from optional — the split the deleted set did not have. Its `minimal_prompt`
  entry, added by #4665, is **optional** there; requiring it (as a wired
  `_require_exact_object` would) would have refused every posture recorded before that
  arm existed.

`_ENGINE_POSTURE_RECORD_FIELDS` stays: it is wired, at the calibration-record check.

The launcher shrinks 4383 → 4377 lines, so nothing is spent against its ceiling.

Closes #4668

## The witness

`bench/harbor_adapter/tests/test_posture.py::TestLauncherDeclaresNoVocabularyNothingReads`
walks `secure_launcher.py` with `ast` and fails on any module-level private binding the
module never loads. It lives in `test_posture.py` because both `secure_launcher.py` and
`test_secure_launcher.py` are on their size ceilings.

Ablation — the fix reverted, the test kept:

```
$ git stash push -- bench/harbor_adapter/stella_harbor/secure_launcher.py
$ uv run --frozen pytest tests/test_posture.py -k VocabularyNothingReads -q

def test_no_private_constant_in_the_launcher_is_read_by_nothing(self) -> None:
>   assert _unread_private_module_constants(_SECURE_LAUNCHER_PY) == []
E   AssertionError: assert ['_ENGINE_POS...STURE_FIELDS'] == []
E     Left contains 3 more items, first extra item: '_ENGINE_POSTURE_AGENT_FIELDS'

1 failed, 1 passed, 45 deselected
```

With the fix:

```
$ uv run --frozen pytest tests/test_posture.py -q
47 passed
```

The second test in the class is the guard's own witness: it plants `_LIVE`, `_DEAD` and
`PUBLIC` in a `tmp_path` module and asserts the reader returns exactly `['_DEAD']`, so an
`ast` walk that silently matched nothing could not report a clean module forever.

## The gate

- [x] `bench/harbor_adapter`: `pytest tests/test_posture.py tests/test_manifest_parity.py` — green
- [x] `./scripts/check-file-size.sh` — OK, nothing went over
- [x] `python3 scripts/check-prose.py` — OK, none added
- [x] `Closes #4668` above and as a commit trailer

## Deleted tests

None.

## Nothing left behind

- Filed **#4894**: `PRIOR_STAGE_OUTCOME_FIELDS` and `PROVIDER_KEY_LIVE_SNAPSHOT_FIELDS`
  in the same file are unread too, and have been since #301. They are public names, so
  deleting one is an API change the private posture sets could not have been, and either
  may want wiring to a parse site rather than deletion — which is why the guard here is
  scoped to private names (a scope that needs no allow-list) and why they are a separate
  issue rather than more of this diff.

`tests/test_secure_launcher.py::test_missing_prior_readiness_job_never_reserves_or_execs`
and `…::test_forged_prior_readiness_artifacts_never_reserve_or_exec` fail on my machine
**on unmodified `origin/main` as well** — the raised message is `prior-stage replay
analyzer could not be loaded` where the test expects `prior Harbor job is missing`. I
have not established whether that is a real defect or a local environment artifact, so I
am reporting it rather than claiming it; the bench workflow's run on this PR is the
evidence that settles it, and I will file it if it reproduces there.

## Summary by Sourcery

Remove unused launcher posture vocabularies and add a regression guard against unread private constants.

Bug Fixes:
- Remove three unused private posture vocabulary constants from the secure launcher to prevent stale schema declarations from implying validation that does not occur.

Enhancements:
- Document the existing whole-dictionary posture validation and its authoritative schema sources.
- Add an AST-based regression guard that detects unread private module constants and verifies the guard itself works.

Tests:
- Add posture tests covering unused private constants in the launcher and the dead-constant detection helper.

### DIFF
--- a/bench/harbor_adapter/stella_harbor/secure_launcher.py
+++ b/bench/harbor_adapter/stella_harbor/secure_launcher.py
@@ -447,23 +447,17 @@ _MANIFEST_CALIBRATION_FIELDS = frozenset(
     }
 )
 _MANIFEST_CONFIRMATORY_FIELDS = frozenset({"job_name", "n_concurrent_trials"})
+# The `engine_posture` record wrapping one posture — the only posture shape this
+# module states. The posture *inside* it is not described here by a field set:
+# `_validate_confirmatory_manifest_contract` rebuilds it with the adapter's own
+# `_benchmark_engine_posture` and compares the whole dict, roles and all, so a
+# key-name vocabulary beside that equality could only ever be weaker than the
+# check already running. The root keys the trusted launcher will accept at all
+# are `settings::ENGINE_ROOT_FIELDS` in `unknown.rs`, read from the Rust rather
+# than copied (`tests/test_posture.py::_engine_root_fields`); the required and
+# optional halves of the recorded posture are
+# `terminal_bench_analysis/tb21_posture_schema.py` (#4668).
 _ENGINE_POSTURE_RECORD_FIELDS = frozenset({"version", "posture", "sha256"})
-_ENGINE_POSTURE_FIELDS = frozenset(
-    {
-        "default_model",
-        "allowed_models",
-        "auto_mode",
-        "effort_auto",
-        "reasoning_auto",
-        "minimal_prompt",
-        "headless_scope_bypass",
-        "agents",
-    }
-)
-_ENGINE_POSTURE_AGENT_ROLES = frozenset({"default", "worker", "verifier", "triage"})
-# Every role is the same shape: #2411 took the output cap out of the posture, so
-# nothing is left for a `params` key to hold. `posture.py` is the normative home.
-_ENGINE_POSTURE_AGENT_FIELDS = frozenset({"effort", "reasoning"})
 RUNTIME_IDENTITY_FIELDS = frozenset(
     {
         "binary_sha256",
--- a/bench/harbor_adapter/tests/test_posture.py
+++ b/bench/harbor_adapter/tests/test_posture.py
@@ -15,6 +15,7 @@ metadata.
 
 from __future__ import annotations
 
+import ast
 import asyncio
 import json
 import re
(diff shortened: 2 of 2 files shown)


macanderson:
Settling the note in the PR body: the two `prior_readiness` failures are a **local environment artifact, not a defect**, so nothing is filed for them.

Evidence: this PR's bench run (32923014661) is green and its `harbor_adapter — sync (locked) + pytest` step reports `761 passed, 1 skipped` — `tests/test_secure_launcher.py` included. On my machine the same two tests fail on unmodified `origin/main` as well, raising `prior-stage replay analyzer could not be loaded` where they expect `prior Harbor job is missing`, which is the analyzer import failing rather than the launcher refusing. CI resolves that import; my checkout does not.

PR #4909, opened 2026-08-26
fix(stella-tui): retire the brand gradient, which had no consumer left

macanderson:
## What

`crates/stella-tui/src/theme.rs` still carried the brand gradient — `BRAND_STOPS`, `primary_stops`, `stops_for`, `brand_gradient`, `gradient_at`, `lerp_rgb`, `lighten` — which rule 5 of the v5 colour alignment retires on every surface, and which SPEC 2 (`cell-grid honest`) forbids by allowing only per-cell colour steps.

It had no consumer. Three of those doc comments name `crate::progress` as the fill they paint; there is no `progress` module in this crate. The wordmark they name as the other consumer comes from `stella_tui_theme::wordmark::spans()` and is two flat spans. Every symbol was `pub`, so `dead_code` never fired and nothing in the gate could see the last caller had gone — the only thing still calling them was the test asserting they worked.

Also corrected: the `apply_theme` module comment claimed the value remap "can't recolour the progress bar's gradient", which was the standing justification for the gradient being theme-aware directly. There are no interpolated cells left for the remap to miss.

## Witness test

`no_colour_interpolator_lives_in_this_crate` (`crates/stella-tui/src/theme/tests.rs`) scans `theme.rs` and `palette.rs` for the interpolators by name. A source scan rather than a symbol reference, because a deleted symbol cannot be named in a test that has to compile — the same shape as the `RETIRED_*` ban list a few hundred lines above it in the same file.

## Ablation

`lerp_rgb` and `BRAND_STOPS` pasted back into `theme.rs`, test untouched:

```
thread 'theme::tests::no_colour_interpolator_lives_in_this_crate' panicked at
crates/stella-tui/src/theme/tests.rs:1021:13:
theme.rs declares `fn lerp_rgb` — rule 5 retires gradients on every surface,
and `apply_theme`'s value remap cannot see an interpolated cell

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 1134 filtered out
```

Restored:

```
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 1134 filtered out
```

## That this is a deletion and not a recolour

`cargo test -p stella-tui` — 1134 + 71 across the integration binaries, 0 failed, **every deck golden passing unblessed**. Nothing under `tests/snapshots/deck/` was re-blessed in this PR; a single changed cell would have failed one of them.

The one test that had to change for a reason other than deletion is `theme_remap_and_apply`'s pass-through case, which built "a colour no token holds" by calling `lerp_rgb` on the two golds. It now uses a literal and asserts the literal really is absent from `ALL_RGB_TOKENS`, so the case cannot pass by accidentally picking a token.

`cargo clippy -p stella-tui --all-targets -- -D warnings` clean. `python3 scripts/check-tokens.py`, `check-prose.py`, `check-line-citations.py`, `./scripts/check-file-size.sh` all green.

## Why this is `Refs` and not `Closes`

#4058's DoD item 3 is "the wordmark renders `stella*` and the brand gradient is gone". The wordmark half landed in #4135; this is the other half.

Items 2, 4 and 5 are already done on `main` (`token.rs` *is* the generated table, no `generated.rs` and no second copy; 33 of 35 deck snapshots carry `stella*` and `✧` appears nowhere in the crate; `MIGRATING` is `{}` and `make tokens` exits 0).

**Item 1 is what keeps the issue open**, and it is the part this PR deliberately does not touch, because the remaining twenty-one literals in `palette.rs` are the three decisions the issue's own 2026-08-20 comment names — the light/paper theme, the `WARNING` amber that fails the gold clamp without being a verdict, and the derived ramp steps — plus five `DATA_*` chart hues the comment does not cover. Each needs a token-system answer, not a remap. Detail on the issue.

Refs #4058

## Tests deleted in this PR

Three, all dropped with the feature they covered — they tested the gradient this PR retires, so there is nothing left for them to assert:

- **`brand_gradient_spans_resting_to_live_accent`** — asserted `brand_gradient(0.0) == ACCENT` and `(1.0) == ACCENT_LIVE`. Its own comment justified the endpoint check by naming `crate::progress` as the consumer that falls back to the resting stop; that module does not exist.
- **`primary_stops_are_the_two_golds_per_theme`** — pinned which pair each theme sweeps between. No sweep, no pair.
- **`lighten_moves_toward_white`** — pinned `lighten`, one of the interpolators removed.

`no_colour_interpolator_lives_in_this_crate` replaces all three: instead of asserting the gradient behaves correctly, it asserts the gradient is absent, which is what rule 5 actually requires. Nothing else was dropped — `theme_remap_and_apply` survives and is edited in place (its pass-through case stopped building its sample colour out of `lerp_rgb`).

### DIFF
--- a/crates/stella-tui/src/theme.rs
+++ b/crates/stella-tui/src/theme.rs
@@ -232,9 +232,10 @@ pub const SUBAGENT: Color = TEAL;
 // keeps ~700 `theme::TOKEN` call sites untouched: a theme is a substitution
 // table, not a parameter threaded through the render tree.
 //
-// The one thing a value remap can't recolour is the progress bar's *gradient*,
-// whose interpolated cells are never equal to a token; so its source
-// ([`brand_gradient`] via [`primary_stops`]) is theme-aware directly.
+// It works because every colour the deck paints IS a token. Rule 5 of the v5
+// colour alignment retires gradients on every surface, and SPEC 2 (`cell-grid
+// honest`) allows only per-cell colour steps, so there are no interpolated
+// cells for the remap to miss (#4058).
 
 /// The shipped themes. `stella-dark` (gold on the cool near-black ramp) is
 /// the default; `stella-light` (the deep gold ramp on cool paper) is its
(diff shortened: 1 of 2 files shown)


macanderson:
Surface parity, since rule 5 is a cross-surface rule: `rg -l "linear-gradient|radial-gradient"` over `crates/stella-observatory/`, `crates/stella-transcript/`, `website/src/app/tokens.css` and `docs/brand/css/` returns nothing. The TUI held the last gradient in the tree, so this closes rule 5 on every surface rather than on one.

Issue #3196, opened 2026-08-13
search gather cache: cross-file facets (importers, import resolution) can serve one generation stale under per-file identity keys


macanderson:
## Problem

PR for #3163 gives `search` a session-lifetime per-file gathered-context cache: `crates/stella-tools/src/search/cache.rs`, consumed by `gathered_neighborhood` in `crates/stella-tools/src/search/enrich.rs`. Entries are keyed by `(path, sha256(file bytes))`, which invalidates **exactly** for every fact that is a function of the file's own bytes (symbols, kinds, import specifiers, and all source-derived facets, which are re-read fresh anyway).

Two facts in the cached `FileNeighborhood` are drawn from the *rest of the tree*:

- `importers` — files whose imports resolve to this file;
- the resolution of this file's own import specifiers to paths (`stella_graph::CodeGraph::file_neighborhood` returns resolved paths where resolution succeeds).

File B gaining an import of A does not change A's bytes, so A's cached entry keeps its old `importers` list until A itself changes or the entry is evicted (256-entry LRU). The per-symbol facets (callers, callees, body spans) are deliberately **not** cached and stay live for exactly this reason — the gap is confined to the two neighborhood facets above.

## Repro / verify

1. Search for a symbol defined in `a.rs`; note the `imported by:` line.
2. Add `b.rs` importing `a.rs`; search again **in the same session** without touching `a.rs`.
3. The cached entry serves the old `imported by:` list (a fresh session, or any edit to `a.rs`, shows the new one).

A regression test would drive `enrich::render_hit` twice against one `GatherCache` with an importer added and the graph re-indexed between renders, asserting the second render names the new importer.

## Constraints discovered

- A per-call flush keyed on this call's own `index_all` stats (`IndexStats::files_parsed`/`files_pruned`) is NOT sufficient: the session's live watcher (`CodeGraph::mount` in the CLI session graph) can index B's change first, so the search call's own catch-up pass sees nothing to do.
- A correct generation stamp is a digest over the index's `(path, sha)` rows (one table scan per search call) — equal digest proves every graph-derived fact is byte-identical to gather time; unequal flushes the whole cache. `stella-graph` would need a small read-only accessor for it.
- The decision logic should stay pure/property-testable (invariant 2); `cache.rs` already has no I/O.

## Done

- Either the generation stamp above (or an equivalent) keys the cross-file half, or `importers`/import-resolution are excluded from the cached value and queried live, with the module docs' "precision boundary" section updated to match.
[...]

PR #4912, opened 2026-08-26
The index-generation guard costs one gather, not six

### LINKED ISSUE #4785: index_generation re-hashes every indexed file row on every render
## Problem

`index_generation` re-hashes all indexed file rows each time it is rendered — measured at **21.8 ms over 1,808 rows** (a Python-side upper bound). The gather cache's own staleness guard therefore costs roughly six gathers (one gather is 3.7 ms), which inverts the point of the guard.

## Done means

The generation is derived from something already maintained (a stored counter bumped on write, or the store's own change token) rather than recomputed by hashing every row, with a measurement showing the new cost, and the staleness behaviour pinned by a test that fails if the generation stops changing on a re-index.

From PR #4774's report (2026-08-24). Related: #3198.

macanderson:
## What

`crates/stella-tools/src/search/cache.rs`'s module doc quoted **21.8 ms over 1 808 rows** for `CodeGraph::index_generation`, taken as a Python-side bound on the same SQL. #4785 read that number against a 3.7 ms gather and concluded the staleness guard "costs roughly six gathers, which inverts the point of the guard".

The Rust path does not cost that. Measured in release against this repository's own index — 1 808 files, the real 180 MB `codegraph.db` copied to a scratch path and opened there, four samples, the first disk-cold:

```
files: 1808
cold index_generation: 3.148916ms
cold index_generation: 728.959µs
cold index_generation: 1.5985ms
cold index_generation: 1.556583ms
```

So **0.7-3.1 ms**, roughly an order of magnitude under the quoted bound. Against a 3.7 ms gather the guard is about one gather, not six; against a search call that also pays 220-296 ms of chunk ranking and 111-202 ms of embedding round trip (the same paragraph's own figures) it is under one percent.

This PR corrects the number and says where the old one came from. Doc-only — no witness, and nothing else in the tree reads the figure.

## Why this does not close #4785

Two findings, both on the issue as a comment:

1. **The premise is quantitatively wrong**, per the measurement above.
2. **The memo shape the issue suggests cannot work on this call path.** `search::engine::report_with` calls `open_or_build` on *every* search and calls `graph.shutdown()` at the end of it, so each search gets a fresh `CodeGraph` with a fresh pair of connections. A memo held on `Inner` — the shape `store::image_check` uses and the one I built and measured first — hits zero times in production: one graph, one `index_generation` call, one hash either way. `PRAGMA data_version` cannot be lifted to a process-global memo either, because SQLite's own contract says the value is a local property of a connection and is not comparable across two.

   A fix that actually lands has to put the stamp **in the store**, maintained at write time, and every write to `code_graph_files` (`store.rs`'s five production sites) has to bump it or the stamp goes stale — which is #3196's staleness bug returning through the guard meant to prevent it. That is a schema-and-fence change, not the memo the issue sketches, and at ~1% of a search call it is not obviously worth making.

Refs #4785

## Summary by Sourcery

Correct the cache documentation to reflect the measured cost of the index-generation guard.

Enhancements:
- Correct the documented performance estimate for the index-generation guard and clarify that the measurement reflects the Rust path rather than the previous Python-side bound.

Documentation:
- Update the cache documentation with measured guard timings and explain the source of the previously overstated estimate.

## Tests deleted

**None.** The diff is one file, `crates/stella-tools/src/search/cache.rs`, 7 insertions and 3 deletions, all inside a `//!` module doc. Zero `#[test]` attributes removed.

## The witness

**No witness needed — documentation only.** Nothing in the tree reads the figure this PR corrects; it is prose a human reads. The measurement behind it is reproducible: release build, this repository's own `codegraph.db` (180 MB, 1 808 indexed files) copied to a scratch path and opened there, four samples, the first disk-cold — 3.148916 ms, 728.959 µs, 1.5985 ms, 1.556583 ms. The scratch harness was not committed; it opened a `CodeGraph` on the copy and timed `index_generation()`.

## Nothing left behind

The unfixed half is **#4785 itself**, which stays open with a re-scope handoff comment (https://github.com/macanderson/stella/issues/4785#issuecomment-5419858905) naming the five write sites, the connection-scoped memo that cannot work here and why, and why `MAX(indexed_at)` is not a safe substitute. Sourcery's two `❌` rows are answered in a PR comment — both are correct readings of a diff that deliberately does not implement the issue, which is why this carries `Refs` rather than `Closes`.

## The gate

`cargo fmt --all` clean; `cargo clippy -p stella-tools --all-targets -- -D warnings` exit 0; `check-prose` OK. Doc-only, so no behaviour to test.

### DIFF
--- a/crates/stella-tools/src/search/cache.rs
+++ b/crates/stella-tools/src/search/cache.rs
@@ -88,10 +88,14 @@
 //! embedding round trip 111-202 ms. Against those, a persisted gather cache
 //! is a rounding error.
 //!
-//! The cache's own guard is in the same range as the thing it guards:
+//! The cache's own guard is a fraction of the thing it guards:
 //! [`stella_graph::CodeGraph::index_generation`] re-hashes every indexed
-//! file's `(path, content sha)` once per render, 21.8 ms over 1 808 rows by
-//! the same Python bound.
+//! file's `(path, content sha)` once per render, **0.7-3.1 ms over 1 808
+//! rows** — measured in release against this repository's own index on
+//! 2026-08-25, four samples, the first of them disk-cold. The Python bound
+//! this paragraph used to quote said 21.8 ms and had #4785 concluding the
+//! guard cost six gathers; the Rust path costs about one, and the difference
+//! is what a bound taken outside the language is worth.
 //!
 //! A faster machine does not change the answer, because the second half of it
 //! is structural. [`GatherCache::observe_generation`] empties the whole cache


macanderson:
Answering both Sourcery ❌ rows, since neither is a defect in this diff.

**Row 1 — "uses a stored counter or store change token instead of re-hashing"** and **Row 2 — "measure the new mechanism and add a regression test"**: both `❌` are correct readings of the diff and both objectives are deliberately out of scope. This PR does not claim to close #4785 — it carries `Refs`, not `Closes`, and its title and body say it corrects a measurement rather than implementing the fix.

Why the fix is not here, in full on the issue (https://github.com/macanderson/stella/issues/4785#issuecomment-5419858905). The short form:

1. **The premise the objectives rest on is wrong by measurement.** The 21.8 ms the issue quotes was a Python-side bound. Measured in release against this repository's own 1 808-row index, four samples, the first disk-cold: 3.15 ms, 0.73 ms, 1.60 ms, 1.56 ms. Against a 3.7 ms gather that is about one gather, not the six the issue computes — and under one percent of a search call that also pays 220-296 ms of chunk ranking and 111-202 ms of embedding. Correcting the number that produced the wrong conclusion is what this PR is for.

2. **The mechanism the issue sketches cannot work on this call path.** I built it first — `PRAGMA data_version` memoized on `graph::Inner`, with an always-compiled hash counter and an integration test going 9 hashes → 1 over nine reads — and it is inert in production: `search::engine::report_with` calls `open_or_build` on every search and `graph.shutdown()` at the end of the same function, so each search gets a fresh `CodeGraph` and a fresh empty memo. One graph, one `index_generation()` call, one hash, memo or no memo. Lifting it to a process-global memo is not available either: SQLite's contract is that `data_version` is a local property of a connection and values from two connections are not comparable.

3. **A fix that would work is a schema-and-fence change, not a memo.** The stamp has to live in the store, maintained at write time, and each of the five production writes to `code_graph_files` has to bump it in its own transaction or the stamp goes stale — which is #3196's cross-file staleness bug returning through the guard that exists to prevent it. The cheap aggregates that would avoid a schema change do not survive scrutiny (`MAX(indexed_at)` is 1-second granular and does not move on a delete).

So the issue stays open with that handoff on it, re-scoped, and this PR lands only the correction it can prove. Shipping the inert memo to satisfy the objectives would have made the assessment green and the tree no faster.

Issue #3923, opened 2026-08-19
Epic: no file crosses the 1500-line ceiling by accident, and no ceiling is bought with a baseline entry


macanderson:
## Outcome

No file in this workspace crosses the 1500-line ceiling by accident, and no ceiling is bought with a baseline entry.

`AGENTS.md` § "God files — plan around them, never into them" requires:

- **No new god file can exist.** A new file crossing 1500 lines fails the gate outright, and `scripts/file-size-baseline.txt` accepts no new entries — `--update` refuses to grandfather a file that is not already there (#3441).
- **The grandfathered files are closed to growth.** New logic lands in a sibling submodule.

The consequence nobody plans for is the one this epic collects: a file at 1495 lines is a **trap laid for the next author**, whose PR has nothing to do with file size and who meets the gate at the exact moment the tempting fix is the forbidden one — a hand-written baseline entry, which `CLAUDE.md` classifies as a defect against the PR that introduces it.

## Why one epic and not thirteen loose issues

Each child is a mechanical split, individually small and individually easy to walk past. Together they are a measured surface: #3776 counted **30 files across 12 crates within 100 lines of the ceiling with no baseline entry**, so this is a standing condition of the tree, not a run of coincidences. Splitting one file is a chore; keeping the whole surface off the trap is a policy, and a policy needs a place to be tracked.

Two of them are already past the line rather than approaching it (#3804), and one names a hole in the guard itself (#3811 — `check-file-size` watches Rust, Python and shell, and did not notice a 1571-line TypeScript file).

## Definition of done

- Every child is either split, or has a maintainer's recorded decision that its ceiling is genuinely irreducible (`./scripts/check-file-size.sh --update --grandfather <path>`, justified in review).
- `#3776`'s census is re-run and its count has gone down, not up.
- `./scripts/check-file-size.sh` is green on a clean `main` with no baseline entry added by any child.

## Constraints a fresh agent needs

- **Splitting buys structure, not slack.** `make file-size-update` retightens every ceiling to its file's current size, so a split does not leave headroom behind it.
- **Never fold a baseline regeneration into an unrelated PR.** The baseline is one shared cell every growing PR writes, and parallel merges on it are AGENTS.md's named #1 cause of a red `main`. Land it on its own from a fresh `main`.
- **The ratchet judges your change, not the tree** (#2004): inherited drift is survived, never absorbed. A first-time crossing you walked past still wants splitting and still gets no baseline entry.
[...]

Issue #3776, opened 2026-08-18
30 files are close to the size limit with no fallback if they grow past it


macanderson:
## Problem

#3705 fixed this shape for three files in `stella-context`. While measuring for
that fix, the same question asked workspace-wide turned up **30 files, in 12
crates, over 1400 lines and absent from `scripts/file-size-baseline.txt`**. The
ratchet grandfathers nothing new, so for each of these the next non-trivial
addition fails `make gate` outright, with no baseline entry available and a
split as the only remedy — discovered mid-PR by whoever happens to touch the
file next, which is the worst moment to decide where a module's seam goes.

One of them, `crates/stella-cli/src/agent/tests/engine_wiring.rs`, is at
**exactly 1500**: the guard's condition is `current > LIMIT`, so a single added
line is a red gate.

Measured on `be1d62a` (`main` at 0.9.97), excluding paths already in the
baseline:

| Lines | File |
|---|---|
| 1500 | `crates/stella-cli/src/agent/tests/engine_wiring.rs` |
| 1497 | `crates/stella-cli/src/model_catalog.rs` |
| 1497 | `crates/stella-cli/src/config/tests.rs` |
| 1496 | `crates/stella-cli/src/agent/persistence.rs` |
| 1495 | `crates/stella-serve/src/server.rs` |
| 1495 | `crates/stella-cli/src/config.rs` |
| 1494 | `crates/stella-graph/src/parse.rs` |
| 1480 | `crates/stella-model/src/http.rs` |
| 1477 | `arenabench/tests/test_arena.py` |
| 1475 | `crates/stella-tools/src/search/engine/tests.rs` |
| 1472 | `crates/stella-tui/src/deck_ui/tests.rs` |
| 1470 | `crates/stella-store/tests/enterprise_telemetry.rs` |
| 1463 | `crates/stella-observatory/tests/schema_conformance.rs` |
| 1462 | `crates/stella-model/src/credential.rs` |
| 1461 | `crates/stella-core/src/driver/loop_escalation.rs` |
| 1459 | `crates/stella-pipeline/src/replay.rs` |
| 1455 | `crates/stella-pipeline/src/pipeline/tests/witness_isolation.rs` |
| 1451 | `crates/stella-cli/src/subsession.rs` |
| 1446 | `crates/stella-cli/src/settings/tests.rs` |
| 1446 | `crates/stella-cli/src/cli.rs` |
| 1438 | `bench/harbor_adapter/tests/test_posture.py` |
[...]

PR #4920, opened 2026-08-26
The file-size guard names the files nearing their ceiling, before one of them fails a PR

### LINKED ISSUE #4897: The file-size guard says nothing until it fails — warn about files near their ceiling
## Problem

`scripts/check-file-size.sh` is silent about a file until the moment it fails. Its success line is

```
check-file-size: OK — <N> watched files, <M> grandfathered over 1500 lines, and nothing went over by this change.
```

A file sitting at 1492 lines and a file sitting at 400 lines produce the same line. So the trap #3923 exists to describe — "a file at 1495 lines is a trap laid for the next author, whose PR has nothing to do with file size and who meets the gate at the exact moment the tempting fix is the forbidden one" — is invisible until it springs. The author who finds it is the one who can least afford to stop and design a module seam, and the fix that looks cheapest to them (a baseline entry) is the one `CLAUDE.md` classifies as a defect against the PR that introduces it.

Measured on `main` at `6f7ea1435` (0.9.174): **44 files are over 1350 lines and absent from `scripts/file-size-baseline.txt`**, 6 of them over 1450. The full list is in #3776's re-run census comment. Draining that list by splitting has been tried across #3923's eleven children and the population regenerates: 30 files over 1400 in the `be1d62a` measurement, 28 over 1400 today, almost none of them the same files. Splitting one file helps one file. Naming the crowd helps every author who has not met it yet.

## Proposal

Add a non-failing advisory to `check-file-size.sh`: after the check passes, print the watched files within some margin of their **effective** ceiling — the baseline entry for a grandfathered file, `LIMIT` for every other — so the growth shows up while it is still cheap to redirect. Something like:

```
check-file-size: OK — 1284 watched files, 47 grandfathered over 1500 lines, and nothing went over by this change.
check-file-size: 6 file(s) within 50 lines of their ceiling — plan around them, do not grow them:
  crates/stella-tui/src/model.rs                1492 / 1500
  crates/stella-plugin/src/consent.rs           1486 / 1500
  ...
```

Design points a fresh agent needs to settle:

- **Advisory, never a failure.** The guard's contract is "the ratchet judges your change, not the tree" (#2004); a warning about a file nobody touched must not become a second way for an unrelated PR to go red. It prints and exits 0.
- **The margin is a number in one place.** Pick it (50 lines is a starting point, ~3% of the limit), put it in one shell variable beside `LIMIT`, and do not repeat it in prose — AGENTS.md's rule about a count in two places.
- **Bound the output.** The full list is 44 lines of noise on every gate run, and an advisory nobody reads is worse than none. Two ways to bound it: cap the list at the tightest N with a count for the rest, or print only the crowded files *this change touched* where a diff scope exists. **Settled on the cap** (see the comment below for the measurement): with grandfathered files excluded the list is 6 lines, under any sensible cap, so diff-scoping would buy nothing and would answer a narrower question — naming the tightest files tree-wide also serves the reader choosing where to put new code, not only the one who already put it in the wrong place.
- **Grandfathered files are excluded**, and the measurement is why (see the comment below). One sits at its recorded ceiling by construction, so counting them puts every one at zero headroom and they fill the list — all ten of the tightest measured this way were grandfathered and not one of the files the advisory exists for was printed. They also need it least: the per-crate table in AGENTS.md and the crate README already say "closed to growth", cross-checked by `check-god-files`. The trap #3923 describes is the file with no entry available and none coming.

## Files

- `scripts/check-file-size.sh` — `effective_limit()` already computes the per-file ceiling this needs; the success `echo` at the end of the file is where the advisory goes.
- `scripts/check-file-size-test.sh` (or the existing guard self-test harness — `make file-size-test` if one exists, otherwise the pattern in `scripts/check-*-test.sh`) — the guard's own directions are covered by a hermetic self-test in this repository, and a new output path needs one.
- `AGENTS.md` § "God files — plan around them, never into them" — one sentence naming the advisory, no counts.

## Verify / done

- A hermetic self-test builds a fixture tree with one file at `LIMIT - 1` and one at `LIMIT - 200`, runs the guard, and asserts the first is named in the advisory and the second is not.
- **Witness**: that test fails on `main` (the advisory does not exist, so the file is never named) and passes with the change.
- The guard still exits 0 when the advisory fires, proven by asserting the exit code in the same test.
- `make guards-fast` green; `make prose` unchanged.

## Not in scope

Splitting any file. This makes the crowd visible; #3923's children are where individual files get split.

Refs #3776, #3923, #2004.

macanderson:
## What & why

`scripts/check-file-size.sh` was silent about a file until the moment it failed. Its success line —

```
check-file-size: OK — 1758 watched files, 19 grandfathered over 1500 lines, and nothing went over by this change.
```

— reads the same whether the tightest file in the tree has eight lines of headroom or eleven hundred. So the trap #3923 exists to describe sprang without warning: the author who meets the ceiling is the one whose change is about something else, who has no room to stop and design a module seam, and for whom the cheapest-looking fix is the baseline entry `CLAUDE.md` calls a defect against the PR that introduces it.

A passing run now lists the crowded files, tightest first:

```
check-file-size: OK — 1758 watched files, 19 grandfathered over 1500 lines, and nothing went over by this change. Judged against 118b8721e.
check-file-size: 6 file(s) sit within 50 lines of their ceiling:
  crates/stella-tui/src/model.rs                                    1492 / 1500  8 line(s) left
  crates/stella-plugin/src/consent.rs                               1486 / 1500  14 line(s) left
  crates/stella-tui/src/model/tests.rs                              1476 / 1500  24 line(s) left
  crates/stella-observatory/src/db.rs                               1470 / 1500  30 line(s) left
  crates/stella-plugin/src/wire.rs                                  1459 / 1500  41 line(s) left
  crates/stella-tui/src/diff.rs                                     1452 / 1500  48 line(s) left
Plan around these rather than into them: a file that crosses gets no baseline
entry, only a split (AGENTS.md, "God files").
```

## Three decisions worth reviewing

**It reports, it does not judge.** Every file named is under its ceiling and the exit code stays 0. A warning about a file nobody touched must never become a second way for an unrelated PR to go red — that is the bug #2004 rewrote this guard to stop, and `W1`–`W6` are all `expect-pass` cases for that reason.

**Grandfathered files are excluded, and the measurement is why.** The first draft counted them, since a baseline entry sits at its recorded ceiling by construction and has zero headroom. Run on this tree, all ten of the tightest were grandfathered and not one of the files the advisory exists for was printed:

```
check-file-size: 23 file(s) sit within 50 lines of their ceiling. The 10 tightest:
  bench/harbor_adapter/stella_harbor/__init__.py                    1862 / 1862  0 line(s) left
  bench/harbor_adapter/stella_harbor/secure_launcher.py             4383 / 4383  0 line(s) left
  ...
```

They also need it least: the per-crate table in `AGENTS.md` and each crate's own README already say "closed to growth". The trap is the file with no entry available and none coming.

**The list is capped.** An advisory nobody reads is worse than none, and 23 lines on every `make guards-fast` run trains people to scroll past the whole guard. `CROWDING_SHOW=10`, with a count for the remainder; both it and `CROWDING_MARGIN=50` live in one variable each beside `LIMIT`, never repeated in prose.

## The witness

Six cases in `scripts/test-file-size.sh` (`make file-size-test`), covering both directions. Ablated by deleting the two lines that emit a `CROWDED` record:

```
FAIL W1 a file 8 lines under the limit is named while there is room — passed, but did not report 'src/crowded.rs':
check-file-size: OK — 3 watched files, 0 grandfathered over 1500 lines, and nothing went over by this change. No base resolved — strict whole-tree check.
FAIL W2 ...with its size, its ceiling and its headroom — passed, but did not report '1492 / 1500  8 line(s) left':
check-file-size: OK — 3 watched files, 0 grandfathered over 1500 lines, and nothing went over by this change. No base resolved — strict whole-tree check.
ok   W3 a file with room to spare is not named
FAIL W4 a file with no entry is still named beside a god file — passed, but did not report 'src/crowded.rs':
check-file-size: OK — 3 watched files, 1 grandfathered over 1500 lines, and nothing went over by this change. No base resolved — strict whole-tree check.
ok   W5 a grandfathered file at its own ceiling is not named
ok   W6 a tree with nothing near the line says nothing

passed 37, failed 3
```

Restored:

```
ok   W1 a file 8 lines under the limit is named while there is room
ok   W2 ...with its size, its ceiling and its headroom
ok   W3 a file with room to spare is not named
ok   W4 a file with no entry is still named beside a god file
ok   W5 a grandfathered file at its own ceiling is not named
ok   W6 a tree with nothing near the line says nothing

passed 40, failed 0
```

The three that survive ablation are the point of including them: `W3`/`W5`/`W6` are what stops an advisory that names every watched file from satisfying `W1`/`W2`/`W4`.

`want_absent` is new in that harness — the existing `want` only asserts a substring is present, and the negative direction needs the opposite.

## Also

- `shellcheck` clean on both scripts.
- `make guards-fast` green, `make prose` and `make line-citations` unchanged.
- One paragraph added to `AGENTS.md` § "God files", with no counts in it.
- A note in the awk block warning that an apostrophe inside that single-quoted program closes the quote and makes bash report a syntax error forty lines away. It cost me a debugging round; the next person gets it free.

## Tests deleted

None.

Closes #4897
Refs #3776, #3923

## Summary by Sourcery

Warn about files approaching the size ceiling before they block future changes without turning the advisory into a failing check.

New Features:
- Add a passing file-size advisory that identifies non-grandfathered files nearing the line limit, ordered by remaining headroom and capped to keep output readable.

Enhancements:
- Keep crowding reports informational so unrelated pull requests continue to pass.
- Exclude grandfathered files from the crowding advisory and document the guidance in AGENTS.md.

Documentation:
- Document the proactive file-size crowding advisory in the God files guidance.

Tests:
- Add witness coverage for reporting crowded files, formatting headroom, excluding roomy and grandfathered files, and remaining silent when no files are near the limit.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -825,6 +825,14 @@ inputs, not review afterthoughts:
   a PR of its own — safe precisely when nothing is blocked on it. Both modes
   still retire an entry whose file dropped under the limit or is gone, because
   the check hard-fails on those and names the update as the only remedy.
+- **A passing run names the files nearing the line** (#4897). `check-file-size`
+  used to say nothing about a file until the moment it failed, so a file eight
+  lines under its ceiling and a file at four hundred produced the same green
+  line — and the author who then met the ceiling was the one whose change was
+  about something else and who had no room to design a seam. The run now lists
+  the crowded files, tightest first, and exits 0 either way: it reports, it does
+  not judge. Grandfathered files are left out, because a baseline entry sits at
+  its ceiling by construction and is already named here and in its crate README.
 
 **The ratchet judges your change, not the tree** (#2004). A file over the line
 fails only when `current > max(its own limit, size in the base tree)` — where
--- a/scripts/check-file-size.sh
+++ b/scripts/check-file-size.sh
@@ -144,6 +144,27 @@ set -euo pipefail
 
 LIMIT=1500
 
+# How little headroom a file may have before the guard names it, and how many
+# of the tightest it names before falling back to a count.
+#
+# Advisory, never a verdict: every file this reports is still UNDER its ceiling,
+# the check passes, and the exit code is 0. Failing a change because some other
+# file is crowded is the bug #2004 rewrote this guard to stop.
+#
+# It exists because the guard was otherwise silent about a file until the moment
+# it failed, so a file at 1492 lines and a file at 400 produced the same green
+# line. The author who then meets the ceiling is the one whose PR is about
+# something else, who has no room to design a module seam, and for whom the
+# cheapest-looking fix is the baseline entry CLAUDE.md calls a defect against
+# the PR that introduces it. #3923 named that trap; this is what puts it in view
+# beforehand (#4897).
+#
+# The cap is here because an advisory nobody reads is worse than none: 44 files
+# were within this margin when it was written, and a 44-line block on every
+# `make guards-fast` run trains people to scroll past the whole guard.
+CROWDING_MARGIN=50
+CROWDING_SHOW=10
+
 repo_root="$(cd "$(dirname "$0")/.." && pwd)"
 cd "$repo_root"
 
@@ -438,10 +459,25 @@ raw_report="$(
   {
     grep -v '^#' "$baseline" | sed 's/^/B /'
     current_sizes | sed 's/^/C /'
-  } | awk -v limit="$LIMIT" '
+  } | awk -v limit="$LIMIT" -v margin="$CROWDING_MARGIN" '
     $1 == "B" { ceiling[$3] = $2; next }
     $1 == "C" {
       path = $3; n = $2; seen[path] = 1
+      # The crowding advisory, and only for a file with NO baseline entry.
+      # (No apostrophes in this block: it sits inside a single-quoted awk
+      # program, where one closes the quote and bash reports a syntax error
+      # forty lines away.)
+      #
+      # A grandfathered file sits at its recorded ceiling by construction, so
+      # including them puts every one of them at zero headroom and they fill
+      # the list — measured on this tree, all ten of the tightest were
+      # grandfathered and not one of the files the advisory exists for got
+      # printed. They also need it least: a baseline entry is already named in
+      # the per-crate table in AGENTS.md and in the crate README, so three
+      # places say "closed to growth" before this one. The trap #3923
+      # describes is the file with no entry available and none coming.
+      if (!(path in ceiling) && n <= limit && limit - n <= margin)
+        crowded = crowded sprintf("CROWDED %d %s %d %d\n", limit - n, path, n, limit)
       if (path in ceiling) {
         if (n <= limit)
           obsolete = obsolete sprintf("  %s is now %d lines (<= %d) — drop its baseline entry\n", path, n, limit)
@@ -455,6 +491,8 @@ raw_report="$(
       for (p in ceiling)
         if (!(p in seen))
           stale = stale sprintf("  %s (baseline entry, file no longer tracked)\n", p)
+      # First, so the report sections below keep the order they had.
+      if (crowded) printf "%s", crowded
       if (newover) printf "%s", newover
       if (grew) printf "%s", grew
       if (obsolete) printf "OBSOLETE\n%s", obsolete
@@ -484,10 +522,18 @@ effective_limit() {
 # the GREW section still lands between NEWOVER and OBSOLETE.
 report=""
 drift=""
+crowded=""
 newover_header_emitted=0
 grew_header_emitted=0
 while IFS= read -r line; do
   case "$line" in
+  "CROWDED "*)
+    # Sortable form: headroom first, so the tightest files lead the advisory.
+    # Handled explicitly rather than falling through to the `*)` arm below,
+    # which appends to `report` — and a non-empty `report` is a failed gate.
+    crowded="$crowded${line#CROWDED }
+"
+    ;;
   "NEWCAND "*)
     # Deliberate word split, as for GREWCAND below.
     # shellcheck disable=SC2086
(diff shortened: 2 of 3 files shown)


macanderson:

Both rows read the diff correctly. Both objectives came from **my own issue text**, written before the thing was run, and a measurement changed the design in each case. #4897's body is now edited to state the shipped design, with [the measurement recorded in a comment](https://github.com/macanderson/stella/issues/4897#issuecomment-5420077857) rather than the body quietly rewritten.

### Row 1 — grandfathered files excluded

> The issue requires grandfathered files to count using their baseline entry as the effective ceiling.

It did, and that was wrong. A baseline entry sits at its recorded ceiling by construction, so counting them puts **every** grandfathered file at zero headroom; they sort to the top and fill the list. Run on `main` at `6b06962c9` with them included, ten of the ten shown were grandfathered and not one of the files the advisory exists for was printed:

```
  bench/harbor_adapter/stella_harbor/__init__.py                    1862 / 1862  0 line(s) left
  bench/harbor_adapter/stella_harbor/secure_launcher.py             4383 / 4383  0 line(s) left
  bench/terminal_bench_analysis/tb21_analysis.py                    8298 / 8298  0 line(s) left
  ...
```

They also need it least: a baseline entry is already named in the per-crate table in `AGENTS.md` and in that crate's README, cross-checked by `check-god-files`, so three places say "closed to growth" before this one. The trap #3923 describes is the file with no entry available and none coming. `W5` pins the exclusion.

### Row 2 — no diff-aware filtering

> The PR reports the tightest non-grandfathered files from the entire tree, capped at ten, and does not implement diff-aware filtering.

Correct, and the issue named this as "the judgement call". Two reasons the cap won:

- With grandfathered files out, the list on `main` today is **6 lines** — under the cap. Diff-scoping would bound output that is already bounded, at the cost of a `git diff` per run and a branch the self-tests would have to cover in both directions.
- It answers a narrower question than the one worth asking. Naming the tightest files tree-wide serves the person **choosing where to put new code**, not only the person who has already put it in the wrong place. That first reader is who #3776's census exists for and who nobody currently serves.

If the list grows past the cap often enough to be scrolled past, diff-scoping is the next thing to try. `CROWDING_SHOW` is where that decision lives — one variable, beside `CROWDING_MARGIN`, beside `LIMIT`, never repeated in prose.

Row 3 (hermetic self-tests and documentation) is ✅ and unchanged. `@sourcery-ai review`

PR #4923, opened 2026-08-26
step_usage names the call it belongs to

### LINKED ISSUE #4793: step_usage cannot be joined to the call it belongs to — it carries neither call_seq nor turn_instance
## Problem

`step_usage` carries `role`, `model`, `step` and `cost_usd`, but **neither `call_seq` nor `turn_instance`**. Only `step_manifest` has both. So a `step_usage` row cannot be attributed to the call that produced it whenever a `(turn_instance, step)` holds more than one call — which is exactly the case the fields exist to disambiguate.

This is not hypothetical: #4391's own suggested analysis ("read the summariser's `step_usage` row at `call_seq` 1") **cannot be performed**, and AGENTS.md's glossary describes `(step, call_seq)` as the step's identity — a description the event does not support.

## Done means

Either `step_usage` gains both fields (additive, `#[serde(default)]`, `PROTOCOL_VERSION` untouched, `make wire-schema` regenerated and the diff shown to be additive), with a witness that two calls sharing one `(turn_instance, step)` are separable by their usage rows; or the docs that imply the join are corrected to say it is not available and why.

Files: `crates/stella-protocol/src/event.rs`, AGENTS.md § Glossary.

Found during PR #4791's compaction census (2026-08-24). Searched first — no existing issue covers it.

macanderson:
## What

`step_usage` carried `step` and nothing else that identifies a call. `step_manifest` carries `turn_instance` and `call_seq`; the metering row beside it carried neither, so a cost could not be attributed to the receipt of the call that produced it whenever a `(turn_instance, step)` held more than one — the exact case `call_seq` exists to disambiguate.

Both fields are added to `AgentEvent::StepUsage`, filled by both production emitters, and regenerated into `docs/wire/`.

## Why `Option`, where `StepManifest` uses a bare `u64`

`StepManifest::call_seq` defaults to `0` because a manifest written before the field existed genuinely was a worker manifest. A `step_usage` row written before this PR may equally have been a summarizer's — that is the whole complaint — so defaulting it to `0` would state that every legacy auxiliary call was the worker. `None` says "this stream cannot say", and a consumer joining on it has to treat that as unjoinable rather than as the worker. Same for `turn_instance`: turn instances are 0-based, so a defaulted `0` would be indistinguishable from a real turn 0.

## Where the values come from

The engine reads the pair off the same `ReceiptLedger` that stamped the receipt emitted immediately before, rather than re-deriving them from bindings at the call site — so the cost row and the receipt carry one identity instead of two that agree by convention. `SettledCall` takes `&ReceiptLedger` for that reason (and because `driver.rs` is a god file closed to growth: the call site is one line, not four).

`accounted_call.rs` — the path every management role and the overflow summarizer takes — reads them off `ReceiptContext`, and emits `None` for a receipt-less call, which is the same condition that already leaves `step` at 0.

## The store needs no change

Worth stating, because the issue's framing suggests one. `telemetry`'s `UNIQUE (execution_id, step)` looks like the same collision, but `step` there holds the event-stream `seq`, not the engine's step — `agent/persistence.rs` says so at the write: *"Event-stream seq is the execution-global call identity; engine-local `step` restarts on each pipeline turn."* Every `StepUsage` gets a distinct `seq`, so the constraint never collides and the durable rows were already separable. The gap was on the event, and only on the event.

## Witness

`stella_core::driver::tests::usage_completeness::two_calls_at_one_step_are_separable_by_their_usage_rows` runs the engine's worker turn at turn instance 4 and dispatches a summarizer-shaped auxiliary call against the same `(4, 0)` through `run_accounted_call`, then asserts the two rows share a step and differ in `call_seq`.

Ablated by returning both emitters to `None`:

```
thread 'driver::tests::usage_completeness::two_calls_at_one_step_are_separable_by_their_usage_rows'
panicked at crates/stella-core/src/driver/tests/usage_completeness.rs:665:9:
assertion `left == right` failed: both calls rode turn 4: [(0, None, None, Worker), (0, None, None, Summarization)]
  left: None
 right: Some(4)

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 1449 filtered out
```

Restored: `test result: ok. 1 passed`.

Round-trip and legacy coverage in `stella-protocol`: `step_usage_roundtrips_as_a_complete_metering_record` now asserts the pair survives, and `step_usage_from_a_pre_request_shape_stream_still_parses` asserts a legacy line decodes to `None`/`None` rather than to the worker.

## Wire schema

`make wire-schema-update`, committed. Purely additive:

```
 docs/wire/agentevent.d.ts        |  8 ++++++++
 docs/wire/agentevent.schema.json | 18 ++++++++++++++++++
 docs/wire/serveframe.d.ts        |  8 ++++++++
 docs/wire/serveframe.schema.json | 18 ++++++++++++++++++
 4 files changed, 52 insertions(+)
```

No deletions, `PROTOCOL_VERSION` untouched, and `check-wire-schema` is green. The all-present / all-absent sample pair in `wire_contract/samples.rs` carries the new fields in both shapes, so `skip_serializing_if` is proved rather than assumed.

AGENTS.md's glossary row for **step** now says the join is available on both events and what a pre-#4793 row means.

No test was deleted or renamed. The struct-literal sites across `stella-cli` and `stella-tui` are mechanical `None` additions to fixtures.

Closes #4793

## Summary by Sourcery

Make step usage events attributable to the specific call that produced each metering record.

New Features:
- Add optional turn-instance and call-sequence identifiers to step usage events so metering records can be joined to the calls that produced them.

Bug Fixes:
- Preserve ambiguity in legacy step usage records by decoding missing identifiers as absent instead of incorrectly attributing them to the worker call.

Enhancements:
- Populate usage identifiers from the receipt ledger across engine and auxiliary call paths, ensuring usage and receipt events share a consistent call identity.

Documentation:
- Regenerate wire schemas and update the glossary to document step usage correlation and legacy-record behavior.

Tests:
- Add coverage for separating multiple calls at the same step and for round-trip, legacy, and wire-contract representations of the new identifiers.

Chores:
- Update affected event fixtures and examples for the expanded step usage shape.

## Tests deleted

**None.** `git diff origin/main...HEAD` removes zero `#[test]` / `#[tokio::test]` attributes and zero test functions. The `stella-cli` and `stella-tui` changes are additive `turn_instance: None, call_seq: None` lines in existing fixtures; no fixture was replaced or renamed.

## Nothing left behind

Filed **#4924** — the `telemetry` table's `step` column holds the event-stream `seq`, not the engine's step, and now that this PR puts `(turn_instance, call_seq)` on the event the durable row could carry the real identity too. Out of scope here (this PR needs no store change) and written as a handoff.

## The gate

- `cargo fmt --all` clean; `cargo clippy -p stella-protocol -p stella-core -p stella-cli -p stella-tui --all-targets -- -D warnings` exit 0.
- `cargo test -p stella-protocol` / `-p stella-core` / `-p stella-cli --bins` / `-p stella-tui --lib` all green locally (`--no-fail-fast`).
- `check-wire-schema`, `check-file-size`, `check-prose`, `check-line-citations`, `check-invariants` all OK. The workspace build and test are CI's, per CLAUDE.md.
- `driver.rs` is a god file at its ceiling: the first version pushed it +3 and the ratchet caught it, so `SettledCall` takes `&ReceiptLedger` and the call site is one line. No baseline was raised.

## Ground-rule check

- No I/O added to `stella-core`; no new dependencies.
- No new outbound network calls.
- New wire fields round-trip through serde, asserted both ways: `step_usage_roundtrips_as_a_complete_metering_record` (present) and `step_usage_from_a_pre_request_shape_stream_still_parses` (absent decodes to `None`, not to the worker's 0). The `wire_contract` sample pair carries the fields in both the all-present and all-absent shapes, so `skip_serializing_if` is proved rather than assumed.


macanderson:
Follow-up filed: #4924 — the `telemetry` table's `step` column holds the event-stream `seq`, not the engine's step, and now that this PR puts `(turn_instance, call_seq)` on the event the durable row could carry the real identity too. Out of scope here (this PR needs no store change, as the description says); tracked there rather than left in a PR comment.

Issue #4910, opened 2026-08-26
The retired-hex ban cannot see a single colour Rust writes


macanderson:
## Problem

`scripts/check-tokens.py`'s ban check — "every hex in `banned.values` fails wherever it appears, in any tracked file, no allowlist" — cannot see a single colour that Rust writes.

It matches two notations: `HEX` (`#RRGGBB`) and `RGB_FUNC` (`rgb()` / `rgba()`). Rust spells a colour `Color::Rgb(0x0A, 0x0A, 0x0C)`, which is neither. `.rs` is in `TEXT_SUFFIXES`, so the script opens every Rust file, scans it, finds nothing, and reports clean.

That is a guard reporting zero and meaning nothing — the exact failure its own `RGB_FUNC` comment names ("a notation language is a guard that reports zero and means nothing") — over the surface where the retired kits actually shipped. `crates/stella-tui/src/palette.rs` alone still holds twenty-one `Color::Rgb` literals (#4058), and none of them has ever been checked against the ban list.

## Measured

A probe with `Color::Rgb\(\s*(0x[0-9A-Fa-f]{1,2}|\d{1,3})\s*,\s*(…)\s*,\s*(…)\s*[,)]` over `git ls-files '*.rs'` at `6f7ea1435`:

```
total Color::Rgb tuples seen: 112
banned hits: 1
  crates/stella-tui/src/theme/tests.rs:468: Color::Rgb(0x0B, 0x0B, 0x0C) = #0B0B0C -- brand v1.0 Ink ground
```

So the tree is clean today, and one file needs the `SELF` exemption before the check can be turned on: `crates/stella-tui/src/theme/tests.rs` is a **ban site** — its `RETIRED_*` constants are the enforcement list, and a hex sweep reads them as offenders exactly backwards. That is the same argument the script already writes out for `website/src/lib/brand-parity.test.ts` and `crates/stella-cli/src/export/tests.rs`, and #4066 is the merge where getting it wrong cost nineteen false reports.

**The tree being clean is what makes this cheap to land and easy to get wrong later.** The guard is silent today, so nothing stops the next Rust surface from picking up `#FFB000` as a tuple.

## Files

- `scripts/check-tokens.py` — `HEX`, `RGB_FUNC`, `SELF`, and the per-line loop in `main()`.
- `design/tokens/stella-tokens.json` — `banned.values`, the 19 retired hexes.
[...]

Issue #4632, opened 2026-08-24
A website-only PR can break a Rust test that reads website files, and no pre-merge workflow runs that test


macanderson:
## Problem

`main` went red on 2026-08-24 because a website-only PR (#4588) moved `PROVIDER_CATALOG` out of the file a Rust test reads (`config::tests::docs_sync::provider_cards_match_the_registry`, `crates/stella-cli/src/config/tests/docs_sync.rs`). The PR was green: `ci.yml` path-ignores `website/**`, so its Rust job never ran, and `docs-guards.yml` runs no cargo tests. The test failed on the next Rust-touching run — which was someone else's PR, and then every PR after it (#4593, #4607 both inherited the red).

#4605 fixes the instance (repoints the reader, anchors the parser on the declaration) — this issue is the class: **a Rust test that reads `website/**` (or any path `ci.yml` ignores) is unenforceable against exactly the diffs most likely to break it.**

This is the same shape that already moved `design-refs` into `docs-guards.yml` (#3888: a docs-only PR could invalidate Rust comments and land green) and gave `deck-paths` a home in `deck-fit.yml` (#1892/#3573).

## Fix shape

Options, for a maintainer to pick:

1. A cheap dedicated job (in `docs-guards.yml` or its own workflow) that runs only the cross-boundary tests — e.g. `cargo test -p stella-cli docs_sync` — triggered on the website paths those tests read. Compiling stella-cli for a docs PR is minutes; scoping the filter to `website/src/components/provider-catalog.ts` (and any future files the tests name) keeps it rare.
2. An inventory guard: a script that greps `crates/**` test sources for `website/` path literals and asserts each named path is covered by some workflow's path filter — the `gate-parity` pattern pointed across the crate/website boundary.
3. Accept post-merge detection and teach `main-canary` to run the cross-boundary tests, so the red at least files its issue instead of being discovered by the next PR author (this run was found by PR authors, not the canary).

## Done when

A website-only PR that renames/moves a file a Rust test reads goes red before merge (options 1–2), or the canary files the main-red issue within its cycle (option 3).

Refs #4588 (the break), #4605 (the instance fix), #3888 and #1892 (the precedents).

macanderson:
[...]

Issue #3888, opened 2026-08-19
check-design-refs runs only in the job that skips on a prose-only diff, so a docs-only PR can create a violation and land green


macanderson:
## Problem

`check-design-refs` runs in exactly one place in CI, and that place is skipped for a docs-only PR. So the guard that polices *citations of documents* is the one guard that does not run when only documents change.

```
$ rg -n 'check-design-refs' .github/workflows/
.github/workflows/ci.yml:187:        run: ./scripts/check-design-refs.sh
```

That step lives in the `fmt + clippy + test` job, which is gated:

```yaml
if: ${{ !cancelled() && needs.changes.outputs.rust != 'false' }}
# .github/workflows/ci.yml:125
```

A prose-only diff skips the job, and — by design, #1863 — a skipped job still satisfies the required context. `docs-guards.yml` exists precisely to cover the `docs/**` and `*.md` paths `ci.yml` ignores, and it runs `doc-links`, `invariants`, `command-docs` and `brand-case` — **but not `design-refs`.**

## Observed

PR #3885 (the repair for #3875) is docs-only. Every check is green and **no CI job ran `check-design-refs.sh` at all** — `fmt + clippy + test` reports `skipping`. The PR whose entire purpose is to make that guard pass gets no CI confirmation that it does. (It was verified locally instead: red on the parent `c2c110810`, green on the branch, plus `make guards-fast` exit 0.)

## The direction that actually bites

The missing confirmation is a nuisance. The inverse is the defect: **a docs-only PR can create a design-refs violation and land green.** Either of these does it, with no `.rs` file touched:

- `git mv docs/spec/<doc>.md docs/design/<doc>.md` while Rust comments cite `doc:<doc>` — every one of those citations becomes a violation.
- Adding a new document under `docs/design/` with an `id` that existing Rust comments already cite.

The result is the shape #3875 just cost: `main` red on a **required** check, so every open PR is red regardless of what it touches, until someone notices and repairs it past the same red check. #3875 got there via a Rust PR (where the guard did run and was overridden — #3887); this path gets there with the guard never running at all, which is worse because there is nothing to override and nothing to see.

## Definition of done

- `check-design-refs.sh` runs on a docs-only PR. The natural home is a step in `docs-guards.yml`, whose `paths:` filters already list `docs/**`, `**/*.md` and the guard scripts — add `scripts/check-design-refs.sh` to both `paths:` lists (push and pull_request) alongside the four already there.
[...]

PR #4929, opened 2026-08-26
fix(scripts,docs,website): hold a document to the token value it publishes

### LINKED ISSUE #3653: Five HTML pages have their own hand-picked colors that no automatic check keeps in sync with the brand kit
## Problem

The v3.0 "Ion on Obsidian" recolour had to update five long-form HTML documents by hand-mapping hex literals, because nothing binds them to `docs/brand/css/tokens.css`:

- `docs/brand/brand-guidelines.html`
- `docs/brand/site/stella-site-mock.html`
- `docs/brand/site/stella-mobile-mock.html`
- `docs/papers/stella-foundry.html`
- `website/public/vision.html`
- `website/public/presentations/investor-deck.html`

Two of them were already stale *before* the recolour, which is the evidence that this is a drift condition and not a hypothetical:

- `docs/papers/stella-foundry.html` was still on the **v1.0 phosphor** ramp (`#FFB000` / `#0B0B0C` / `#F4F1EA`) — it never moved for the 2026-08-11 bronze rebrand.
- `website/public/presentations/investor-deck.html` carried a **third, invented** warm palette (`#D6A526`, `#8B6720`, `#865D08`, `#F4EFE4`, …) that matched neither v1.0 nor v2.0.

So at the moment the v3.0 work started, three different brand generations were shipping side by side on public URLs. `website/src/lib/brand-parity.test.ts` did not catch it: it walks `website/src` and `website/public` for *retired* values, which caught nothing here because the deck's palette was never a kit palette to begin with, and it does not reach `docs/` at all.

## Why this is not #2594

[#2594](https://github.com/macanderson/stella/issues/2594) proposed holding the marketing surfaces to the *instrument* palette and was closed `wontfix` on the correct ground that a marketing page is legitimately not an instrument. Nothing here reopens that. These pages may keep their own chrome, their own accent usage and their own status hues. The ask is only that the **brand core** they claim to be built from — the brand ramp, the ink/paper pair, the neutral ramp — is the ramp they actually ship, which is a question about a copy rather than about a design.

## Files

- Guard to extend: `website/src/lib/brand-parity.test.ts` (the `no retired brand value survives anywhere in the site` test is the nearest shape, and already walks a directory tree).
- Normative source: `docs/brand/css/tokens.css`.
- The six documents listed above.

## Suggested approach

The cheap version is a widened retired-value sweep: walk `docs/**` and `website/public/**` for the v1.0 phosphor set, the v2.0 bronze set and any *future* retired set, keeping the existing carve-out for files that name a retired value on purpose (the tokens.css comment, the test itself).

The more thorough version also asserts the positive: each of these documents declares its brand core in a delimited block (as `crates/stella-observatory/src/assets/index.html` does with `BEGIN palette` / `END palette`) and a test diffs that block against the kit. That is the shape `crates/stella-cli/tests/design_token_parity.rs` already uses for the six instrument surfaces, and it is what makes a *new* stale value fail rather than only an old one.

## Verify

1. Change one hex in `docs/brand/css/tokens.css` (say `--stella-brand-500`).
2. The new guard must fail and name every document that did not follow.
3. Revert; the guard passes.

Add the same regression check the other way: reintroduce `#D6A526` into `investor-deck.html` and the guard must reject it as a palette the kit never defined.

## Definition of done

- A guard (test or `make gate` step) fails when any of the six documents disagrees with `docs/brand/css/tokens.css` on the brand core.
- The guard runs somewhere CI reaches for `docs/**` edits — note `ci.yml` ignores `docs/**`, so this needs `docs-guards.yml` or the `website` test job, not `ci.yml`.
- `docs/papers/stella-foundry.html` and `website/public/presentations/investor-deck.html` are covered specifically, since both are the ones that actually drifted.

## Constraints

- `deck-fit.yml` measures every slide of every deck under `website/public/presentations/` against a 1600x900 canvas; a guard that rewrites deck markup must keep `node scripts/deck-fit.mjs` passing (it needs `playwright-core`; `PLAYWRIGHT_BROWSERS_PATH=/opt/pw-browsers` locally).
- Do not fold this into an unrelated PR: like `Cargo.lock` and `scripts/file-size-baseline.txt`, a shared-cell guard should land on its own from a fresh `main`.

macanderson:
## The gap

`check-tokens.py` asks two **membership** questions — is this hex retired, and inside a token-only path is it a token — and neither can see a **misquote**:

- a **superseded** value is not banned, because nothing retired it; the palette simply moved past it;
- a value bound to the **wrong name** is still a live token, so it satisfies the token-only rule while saying something false.

#3653 asked for the positive half of the brand-parity guard. This is it, and the drift it found was live.

## What was wrong in the tree

**Twenty misquoted token values**, in four files:

| file | what it published |
|---|---|
| `docs/brand/brand-guidelines.html` | the whole paper family plus `--st-ink` / `--st-ink-muted` at the pre-#4282 **cool** values — nine rows |
| `docs/brand/prompts/stella-design-system-prompt.md` | the same nine rows |
| `docs/brand/css/tokens.css` | `--st-text` bound to `--st-paper-text`'s value |
| `website/src/app/tokens.css` | the same, on the live marketing site |

The first two are the documents a designer reads to learn the palette, and they stated nine colours the palette does not have. The last two collapsed the deck's primary text onto the off-deck one — under a guard reporting clean, because both values are live tokens.

The other **141** citations in the tree already agreed, so this is a real drift condition rather than a rewrite.

Nothing consumes `var(--st-text)` anywhere (`rg 'var\(--st-text\)'` → 0 matches), so correcting it changes nothing rendered; it makes the declaration true.

## Scope — #2594 stays closed

The check fires **only where a document names an `--st-*` token**. A marketing page keeping its own chrome, accent usage and status hues under its own variable names is untouched. `website/public/vision.html`'s `--border: #1F262D` is not a token citation and is not judged. That is the line #2594 drew, and this does not cross it.

## Ablation — the old guard gives the *wrong* answer

`--st-text` reintroduced as `#F4F1EA` in `docs/brand/css/tokens.css`, then `origin/main`'s `check-tokens.py` run over that exact tree:

```
$ git show origin/main:scripts/check-tokens.py > scripts/check-tokens.py
$ python3 scripts/check-tokens.py
tokens: no retired hex in the tree; 3 token-only path(s) clean
OLD_EXIT=0
```

**Exit 0 on a tree where a shipped stylesheet binds a token to another token's value.** Not merely constant — wrong.

The new guard, same tree:

```
$ python3 scripts/check-tokens.py

token values misquoted in 1 place(s). A document that publishes a token's value is read as the palette:

  docs/brand/css/tokens.css:43: --st-text is quoted as #F4F1EA but holds #E8E8EC

quote the value design/tokens/stella-tokens.json holds, or stop naming the token.
NEW_EXIT=1
```

And on the corrected tree:

```
$ python3 scripts/check-tokens.py
tokens: no retired hex in the tree; 3 token-only path(s) clean; 141 token citation(s) quote the palette
EXIT=0
```

This is also the issue's own **Verify** step 1, from the other end: move a token in `design/tokens/stella-tokens.json` and every document that did not follow is named.

## The self-test — `make tokens-test`

`check-tokens.py` took no arguments, so there was no way to point it at a fixture tree and therefore no way to prove it fires. It now takes a root (defaulting to the current behaviour), and `scripts/test-tokens.sh` drives all three checks from both directions:

```
ok   C1 a superseded value is named with both hexes
ok   C2 a live token bound to the wrong name is flagged
ok   C3 a hex under a name no token declares is flagged
ok   C4 a correct citation passes and is counted
ok   C5 a lower-case citation is read as the same value
ok   C6 a minified line pairs each name with its own value
ok   C7 the wrong value on a shared line is named as itself
ok   C8 a token named in prose with no value is not a citation
ok   B1 a retired hex is banned wherever it appears
ok   B2 a ban site may name a retired hex
ok   B3 SELF suppresses the ban and not the citation check
ok   F1 a fixture corpus is exempt from all three checks
ok   T1 a stray hex in a token-only path is flagged
ok   N1 a tree with no colour in it passes
ok   R1 this repository quotes its own palette

passed 15, failed 0
```

Every passing case is pinned on what the report *said*, so a guard that passed by scanning nothing cannot satisfy it (`N1`, `C4`). This satisfies #4910's DoD 1 and 4; its `Color::Rgb` notation half stays open.

`FIXTURES` is separate from `SELF` on purpose: a `SELF` file is a **document** that quotes a retired value in order to ban it, and keeps its duty to quote a live token correctly (`B3`); a fixture corpus is not a document, and every colour in it is an input. Keeping them one tuple would hand any file calling itself a test a blanket pass.

## The CI half — #3653's second DoD bullet

The four colour rulers (`tokens` + `gen-tokens --check`, `hue-separation`, `contrast`, `css-vars`) ran **only in `ci.yml`**, whose `check` job is gated on `needs.changes.outputs.rust != 'false'`. `scripts/ci-rust-scope.sh` answers `false` for a diff confined to `docs/**` and `website/**` — which is exactly the diff whose subject is the brand kit, the six standalone HTML documents, the two token mirrors and the decks.

So a PR reintroducing a retired hex into `docs/brand/brand-guidelines.html`, or darkening a licensed pairing, could land with every reported check green. They now also run in `docs-guards.yml`, which triggers on those paths. Same shape as #3888 (`design-refs`) and #4632 (`website-inputs`), one subject over.

## Not fixed here, filed instead

Three surfaces still ship a neutral ramp that is not the kit's, under their own variable names — outside this guard's scope by the #2594 rule, and a design question rather than a copy question:

- `docs/papers/stella-foundry.html` — `--surface: #0D1319`, `--border: #1F262D`, and an `n300`…`n800` blue-slate ramp
- `website/public/vision.html` — `--border: #1F262D`
- `website/public/presentations/investor-deck.html` — `--border: #1F262D`, `--muted: #60666D` / `#90979E`

Filed as #4931 rather than recoloured in a guard PR.

## Gate

`make guards-fast` green, including `prose` (two constructions it caught were deleted, not baselined), `shellcheck`, `gate-parity`, and all four colour rulers. No baseline raised. No new gate step — the check folds into `tokens`, which is already one.

## Test deletions

None.

Closes #3653
Refs #4910

## Summary by Sourcery

Keep published token values synchronized with the authoritative palette and enforce the colour-system guards for documentation and website changes.

New Features:
- Add validation that documents publishing `--st-*` tokens quote the current palette values.
- Allow token checks to run against a specified repository root for hermetic fixture testing.

Bug Fixes:
- Correct stale paper and ink values in brand documentation and prompt materials.
- Correct `--st-text` values in the brand and website token stylesheets.
- Prevent documentation-only changes from bypassing the colour-system guards in CI.

Enhancements:
- Expand the token guard to report and validate token citations while preserving retired-value and token-only checks.
- Extend token self-tests to cover citation matching, case handling, minified declarations, unknown names, exemptions, and reporting.

CI:
- Run token generation, token citation, hue separation, contrast, and CSS variable checks in the documentation guards workflow.
- Update workflow path filters to trigger colour guards for relevant scripts and design-token changes.

Documentation:
- Synchronize published brand palette values with the authoritative design token definitions.

Tests:
- Expand the token guard test suite with 15 hermetic checks covering retired colours, stray values, and token citations.

Chores:
- Update Makefile and self-test workflow descriptions for the expanded colour guard coverage.

### DIFF
--- a/.github/workflows/docs-guards.yml
+++ b/.github/workflows/docs-guards.yml
@@ -1,8 +1,8 @@
 name: docs-guards
 
-# The guards over prose, listed below and enumerated by the job's own steps.
-# None needs a Rust toolchain, which is why they live here rather than in ci.yml
-# -- that workflow deliberately skips its Rust jobs for a diff confined to
+# The guards over prose and colour, listed below and enumerated by the job's own
+# steps. None needs a Rust toolchain, which is why they live here rather than in
+# ci.yml -- that workflow deliberately skips its Rust jobs for a diff confined to
 # `docs/**`, `website/**` and `*.md` (a job-level prose filter since #1892), and
 # this one triggers on exactly those paths.
 #
@@ -68,6 +68,15 @@ name: docs-guards
 #     docs/brand/README.md ever invoked it. `--check` returns before it reaches
 #     rsvg-convert, so it needs no renderer and no desktop font.
 #
+# 10. the four colour rulers (scripts/check-tokens.py with gen-tokens.py
+#     --check, check-hue-separation.py, check-contrast.py, check-css-vars.py):
+#     what kind of colour a value is, whether two of them can be told apart,
+#     whether one can be read on its ground, and whether a token sheet's every
+#     var() resolves. They ran only in ci.yml, whose `check` job is skipped
+#     whole for a diff confined to `docs/**` and `website/**` -- the diff whose
+#     subject is the brand kit, the standalone HTML documents, the two
+#     hand-maintained token mirrors and the decks (#3653).
+#
 # Rust sources are in the path filter because the first three read them: a
 # citation added to a .rs comment is exactly how one goes stale, and the
 # subcommand list the third guard checks is itself a Rust enum.
@@ -89,6 +98,13 @@ on:
       - "scripts/website-rust-inputs.txt"
       - "scripts/file-size-baseline.txt"
       - "README.md"
+      - "scripts/check-tokens.py"
+      - "scripts/gen-tokens.py"
+      - "scripts/check-hue-separation.py"
+      - "scripts/check-contrast.py"
+      - "scripts/check-css-vars.py"
+      - "scripts/contrast-baseline.txt"
+      - "design/tokens/**"
       - ".github/workflows/docs-guards.yml"
   push:
     branches: [main]
@@ -108,6 +124,13 @@ on:
       - "scripts/website-rust-inputs.txt"
       - "scripts/file-size-baseline.txt"
       - "README.md"
+      - "scripts/check-tokens.py"
+      - "scripts/gen-tokens.py"
+      - "scripts/check-hue-separation.py"
+      - "scripts/check-contrast.py"
+      - "scripts/check-css-vars.py"
+      - "scripts/contrast-baseline.txt"
+      - "design/tokens/**"
       - ".github/workflows/docs-guards.yml"
   workflow_dispatch:
 
@@ -152,6 +175,36 @@ jobs:
       - name: docs prose spells the wordmark lowercase
         if: ${{ !cancelled() }}
         run: bash scripts/check-brand-case.sh
+
+      # The four colour rulers. They run in ci.yml, and ci.yml's `check` job is
+      # skipped whole for a diff confined to `docs/**` and `website/**` — which
+      # is the diff every one of them exists to judge. `docs/brand/`, the six
+      # standalone HTML documents, the two hand-maintained token mirrors and the
+      # decks are all inside the skipped paths, so a PR reintroducing a retired
+      # hex into the brand kit, or darkening a licensed pairing, could land with
+      # every reported check green (#3653). Same shape as #3888 and #4632 above,
+      # one subject over.
+      #
+      # `--check` on the generator rides with them because it answers the
+      # question the others assume: that the artifacts they read off the
+      # authority are the ones the authority would produce.
+      - name: the colour system is the one in the tree
+        if: ${{ !cancelled() }}
+        run: |
+          python3 scripts/gen-tokens.py --check
+          python3 scripts/check-tokens.py
+
+      - name: no two web semantic roles sit within 30° of each other
+        if: ${{ !cancelled() }}
+        run: python3 scripts/check-hue-separation.py
+
+      - name: every licensed colour pairing is readable on its ground
+        if: ${{ !cancelled() }}
+        run: python3 scripts/check-contrast.py
+
+      - name: every var() in a token sheet resolves
+        if: ${{ !cancelled() }}
+        run: python3 scripts/check-css-vars.py
 
       # Both directions matter here, which is why it runs in ci.yml too. A new
       # guard is a Makefile change (ci.yml sees it); an edit that quietly drops
--- a/.github/workflows/guard-self-tests.yml
+++ b/.github/workflows/guard-self-tests.yml
@@ -178,7 +178,7 @@ jobs:
         if: ${{ !cancelled() }}
         run: ./scripts/test-main-red-claim.sh
 
-      - name: the retired-hex ban, in every notation it watches
+      - name: the retired-hex ban in every notation, and the citation check
         if: ${{ !cancelled() }}
         run: ./scripts/test-tokens.sh
 
--- a/Makefile
+++ b/Makefile
@@ -367,7 +367,7 @@ tokens: ## Validate the colour system: hue clamp, generated-file sync, no retire
 	@python3 ./scripts/check-tokens.py
 
 .PHONY: tokens-test
-tokens-test: ## Test the retired-hex ban across every notation it watches (hermetic; not part of `gate`)
+tokens-test: ## Test the colour guard: the ban in every notation, and the citation check (hermetic; not part of `gate`)
 	./scripts/test-tokens.sh
 
 .PHONY: tokens-update
--- a/design/tokens/stella-tokens.json
+++ b/design/tokens/stella-tokens.json
@@ -417,7 +417,17 @@
       "Hexes that may not appear anywhere outside this file and the generated",
       "artifacts that carry the supersession note. Each entry names where it came",
       "from, so a reader hitting the guard learns which retired system they are",
-      "looking at rather than only that they are wrong."
+      "looking at rather than only that they are wrong.",
+      "",
+      "Most entries are a superseded kit. The last four are not: they were never",
+      "a kit at all. website/public/presentations/investor-deck.html carried its",
+      "own warm palette, matching neither v1.0 nor v2.0, and it shipped on a",
+      "public URL beside two other brand generations (#3653). A value nobody",
+      "specified has exactly the same standing here as one that was retired --",
+      "no legitimate use -- so it is banned on the same list rather than a",
+      "second one. The tree carries none of the four; they are listed so",
+      "reintroducing one fails, which is what #3653's own Verify step asks for",
+      "and what nothing did until this entry existed."
     ],
     "values": [
       { "hex": "#FFB224", "was": "phosphor-era amber accent" },
@@ -438,7 +448,12 @@
       { "hex": "#A97227", "was": "brand v4.0 gold shade" },
       { "hex": "#674415", "was": "brand v4.0 gold shade" },
       { "hex": "#462E10", "was": "brand v4.0 gold shade" },
-      { "hex": "#291B0D", "was": "brand v4.0 gold shade" }
+      { "hex": "#291B0D", "was": "brand v4.0 gold shade" },
+
+      { "hex": "#D6A526", "was": "investor-deck's invented warm gold -- never a kit value (#3653)" },
+      { "hex": "#8B6720", "was": "investor-deck's invented warm gold -- never a kit value (#3653)" },
+      { "hex": "#865D08", "was": "investor-deck's invented warm gold -- never a kit value (#3653)" },
+      { "hex": "#F4EFE4", "was": "investor-deck's invented warm paper -- never a kit value (#3653)" }
     ]
   }
 }
--- a/docs/brand/brand-guidelines.html
+++ b/docs/brand/brand-guidelines.html
@@ -284,15 +284,15 @@ set. Light has its own stops, and it is the only theme where <b>ink</b> and the
 paper tints appear.</p>
 <table>
 <tr><th>token</th><th>value</th><th>role</th></tr>
-<tr><td>--st-paper</td><td>#FFFFFF</td><td>light canvas</td></tr>
-<tr><td>--st-paper-ground</td><td>#F4F4F6</td><td>the cool page ground, one step under the panel</td></tr>
-<tr><td>--st-paper-panel</td><td>#F6F6F8</td><td>light panel</td></tr>
-<tr><td>--st-paper-raised</td><td>#FAFAFC</td><td>surface raised above the page ground</td></tr>
-<tr><td>--st-paper-row</td><td>#E6E6EA</td><td>light hover and selected rows</td></tr>
-<tr><td>--st-paper-border</td><td>#E3E3E8</td><td>light border</td></tr>
-<tr><td>--st-paper-seam</td><td>#DDDDE3</td><td>the hairline under a border</td></tr>
-<tr><td>--st-ink</td><td>#141416</td><td>primary text on light</td></tr>
-<tr><td>--st-ink-muted</td><td>#5E5E69</td><td>secondary text on light</td></tr>
+<tr><td>--st-paper</td><td>#FFFCF5</td><td>light canvas</td></tr>
+<tr><td>--st-paper-ground</td><td>#F7F4ED</td><td>the warm paper page ground, one step under the panel</td></tr>
+<tr><td>--st-paper-panel</td><td>#F9F6EF</td><td>light panel</td></tr>
+<tr><td>--st-paper-raised</td><td>#FDFAF3</td><td>surface raised above the page ground</td></tr>
+<tr><td>--st-paper-row</td><td>#E9E6E0</td><td>light hover and selected rows</td></tr>
+<tr><td>--st-paper-border</td><td>#E6E3DD</td><td>light border</td></tr>
+<tr><td>--st-paper-seam</td><td>#E0DDD7</td><td>the hairline under a border</td></tr>
+<tr><td>--st-ink</td><td>#141413</td><td>primary text on light</td></tr>
+<tr><td>--st-ink-muted</td><td>#605F5C</td><td>secondary text on light</td></tr>
 <tr><td>--st-gold-ink</td><td>#725A00</td><td>gold as <em>text</em> on the light ground</td></tr>
 <tr><td>--st-green-ink</td><td>#006933</td><td>pass, as text on light</td></tr>
 <tr><td>--st-amber-ink</td><td>#8A3F00</td><td>warning, as text on light</td></tr>
--- a/docs/brand/css/tokens.css
+++ b/docs/brand/css/tokens.css
@@ -40,7 +40,7 @@
   --st-gold-bright: #F7D96B;
   --st-silver: #A9AAB5;
   --st-silver-type: #BFC1CC;
-  --st-text: #F4F1EA;
+  --st-text: #E8E8EC;
   --st-muted: #777782;
   --st-dim: #4B4B56;
   --st-comment: #565660;
--- a/docs/brand/prompts/stella-design-system-prompt.md
+++ b/docs/brand/prompts/stella-design-system-prompt.md
@@ -46,15 +46,15 @@ The system is **flat**. There are no 50→950 ramps: v5.0 deletes both the brand
 
 | token | hex | role |
 |---|---|---|
-| `--st-paper` | `#FFFFFF` | light canvas |
-| `--st-paper-ground` | `#F4F4F6` | the cool page ground, one step under the panel |
-| `--st-paper-panel` | `#F6F6F8` | light panel |
-| `--st-paper-raised` | `#FAFAFC` | surface raised above the page ground |
-| `--st-paper-row` | `#E6E6EA` | light hover and selected rows |
-| `--st-paper-border` | `#E3E3E8` | light border |
-| `--st-paper-seam` | `#DDDDE3` | the hairline under a border |
-| `--st-ink` | `#141416` | primary text on light |
-| `--st-ink-muted` | `#5E5E69` | secondary text on light |
+| `--st-paper` | `#FFFCF5` | light canvas |
+| `--st-paper-ground` | `#F7F4ED` | the warm paper page ground, one step under the panel |
+| `--st-paper-panel` | `#F9F6EF` | light panel |
+| `--st-paper-raised` | `#FDFAF3` | surface raised above the page ground |
+| `--st-paper-row` | `#E9E6E0` | light hover and selected rows |
+| `--st-paper-border` | `#E6E3DD` | light border |
+| `--st-paper-seam` | `#E0DDD7` | the hairline under a border |
+| `--st-ink` | `#141413` | primary text on light |
+| `--st-ink-muted` | `#605F5C` | secondary text on light |
 | `--st-gold-ink` | `#725A00` | gold as *text* on the light ground, where the metal cannot clear AA |
 | `--st-green-ink` | `#006933` | pass, as text on light |
 | `--st-amber-ink` | `#8A3F00` | warning, as text on light |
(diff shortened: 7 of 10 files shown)


macanderson:

Both are fair readings of the issue's DoD. One I have now **fixed**; one I am **answering**, with the reason it is out of scope and the issue it is tracked in.

---

### ❌ Row 2 — "does not modify or validate the two specifically required drift cases" → **fixed in `8c70e8c84`**

Sourcery is right that the DoD names `docs/papers/stella-foundry.html` and `website/public/presentations/investor-deck.html` specifically. I checked what "covered" costs for each, and they turned out to be different problems.

**`investor-deck.html` was genuinely uncovered, and the issue said exactly how to prove it:**

> Add the same regression check the other way: reintroduce `#D6A526` into `investor-deck.html` and the guard must reject it as a palette the kit never defined.

That Verify step **failed on this PR as it stood**, and on `main`:

```
$ python3 -c "import json;d=json.load(open('design/tokens/stella-tokens.json'));
              print('#D6A526' in {e['hex'].upper() for e in d['banned']['values']})"
False
```

`#D6A526` is not retired — no kit ever had it, which is the issue's own complaint about that file — and it is not a token, so the ban passed it and the token-only rule never looked. All four of the deck's invented values now join `banned.values`, in a category the block's comment now states separately: the rest are superseded kits, these were never a kit at all.

**Live demonstration on the real file, exactly as the issue asks:**

```
$ # --gold: #EFC53F  ->  --gold: #D6A526  in website/public/presentations/investor-deck.html
$ python3 scripts/check-tokens.py
retired brand hexes found in 1 place(s). These belong to superseded kits and have no legitimate use:

  website/public/presentations/investor-deck.html:32: #D6A526 — investor-deck's invented warm gold -- never a kit value (#3653)

NEW_EXIT=1
```

**Ablated to `origin/main`'s ban list, same guard, same modified deck** — isolating the ban-list change:

```
$ python3 scripts/check-tokens.py
tokens: no retired hex in the tree; 3 token-only path(s) clean; 141 token citation(s) quote the palette
OLD_BANLIST_EXIT=0
```

Exit 0 with the deck carrying the invented gold. Wrong, not merely constant. Pinned as `B1b` in the self-test (now 16 cases).

**`stella-foundry.html` was already covered, and the issue's premise about it is stale.** The drift it names there is the v1.0 phosphor set (`#FFB000` / `#0B0B0C`), and both have been on the ban list since #4066. The issue's reason nothing caught it was that the only guard was `brand-parity.test.ts`, which "does not reach `docs/` at all" — but `check-tokens.py`'s ban sweep walks `git ls-files`, which does. That guard postdates the issue.

So after this commit, both named files are covered: one by a ban list that now includes its invented palette, one by a ban list that already included its retired one.

---

### ❌ Row 1 — "does not assert a delimited brand-core block … or compare their broader brand ramp, ink/paper pair, and neutral ramp" → **answering: deliberately out of scope, tracked in #4931**

Accurate, and here is why I stopped where I did.

The issue offers a cheap version (widen the retired-value sweep) and a thorough one (a delimited block per document, diffed against the kit). It also draws a boundary that the thorough version has to respect:

> Nothing here reopens [#2594]. These pages **may keep their own chrome, their own accent usage and their own status hues.** The ask is only that the **brand core** they claim to be built from … is the ramp they actually ship, which is a question about a copy rather than about a design.

The trouble is that on three of the six, the disagreement is **not** a copy question. Measured from each `:root` block:

| document | off-kit, under its own variable names |
|---|---|
| `docs/papers/stella-foundry.html` | `--surface: #0D1319`, `--border: #1F262D`, and an `--n300`…`--n800` blue-slate ramp with **no counterpart in the kit at all** |
| `website/public/vision.html` | `--border: #1F262D` |
| `website/public/presentations/investor-deck.html` | `--border: #1F262D`, `--muted: #60666D` / `#90979E` |

A delimited block covering those requires first deciding **whether they are allowed their own neutral ramp**, and if not, what `--n300`…`--n800` map onto — the kit has no such stops. That is a design call, and a guard PR is the wrong place to make one silently. `#1F262D` appearing in all three also means it is one shared third ramp propagating between documents, not three independent style choices, which is more than a copy fix.

Everything on those pages that *is* already on-token — `--gold`, `--gold-300`, `--ink`, `--paper`, `--surface` on two of the three, `--muted` — is exactly what makes the neutral ramp stand out as the remaining disagreement.

So it is filed as **#4931**, with the measurements above, the two options, and the `deck-fit.yml` constraint a recolour has to keep passing. I would rather hand a maintainer a measured decision than guess at a ramp inside a PR about a guard.

What this PR *does* deliver against that objective, and which no retired-value sweep can: **the positive assertion** — wherever any of these documents publishes a token's value, that value must be the one the palette holds today. That caught 20 live misquotes, including nine rows each in the two documents a designer reads to learn the palette, and two shipped stylesheets binding `--st-text` to `--st-paper-text`'s value. None of those is a retired hex, so the negative sweep reported clean over all of them.

---

### Row 3 ✅

No change.

---

`make guards-fast` is green on the new head, and the self-test is 16/16. Both new commits are covered by `docs-guards.yml` and `guard-self-tests.yml`, whose runs on the previous head already showed the new steps executing (`141 token citation(s) quote the palette`).

macanderson:
## Collision with #4922 — union prepared and verified, waiting on that merge

#4922 and this PR independently built the same three things (a root argument, `scripts/test-tokens.sh`, the `guard-self-tests.yml` wiring). #4922 lands first. I have built the **union** against its branch tip and verified it; it is pushed as **`fix/token-citations-3653-union`** so nothing is lost while #4922 is still open.

`#4922` is `OPEN` as of this comment, so the rebase onto `origin/main` is not done yet. When it lands:

```
git rebase --onto origin/main fix/tokens-rust-tuples-4910 fix/token-citations-3653-union
git push --force-with-lease origin HEAD:fix/token-citations-3653
```

The prepared commit is a clean delta against #4922's tip — `git diff fix/tokens-rust-tuples-4910..fix/token-citations-3653-union` touches 10 files and contains none of #4922's own changes — so it applies whether #4922 is squashed or merged.

### All three of #4922's `check-tokens.py` changes are kept

- `RUST_RGB` + `channel()` — the `Color::Rgb(0x0A, …)` notation, in both hex-literal and decimal forms.
- `crates/stella-tui/src/theme/tests.rs` in `SELF`.
- `record()` ban-hit dedupe keyed on `(file, line, value)`.

### One design decision, and why it does not cost #4922 anything

`SELF` now suppresses **the ban only**, not the whole file, so a ban site's *citations* are still checked. A ban site names a retired value on purpose; that is no licence to misquote a live one. **This does not affect #4922's `theme/tests.rs` entry** — that file carries no `--st-` name at all (`rg -c -- '--st-' crates/stella-tui/src/theme/tests.rs` → 0), so its ban exemption is untouched and it gains no citation hits.

**My `FIXTURES` tuple is dropped entirely.** #4922's discipline — every fixture reads its colours out of the token file at run time, so the suite needs no exemption — is strictly better than adding one, and it is what lets `SELF` narrow without the suite tripping over itself. The citation fixtures now derive their values the same way (three `(css, hex)` pairs, one hex that is neither a token nor banned, one `--st-` name no token declares). **Nothing in the tree is exempt from all three checks.**

### The suite: 21 cases

**Kept from #4922 (all 10):** banned hex literal caught · live hex literal passes · banned `Color::Rgb` tuple caught · live tuple passes · decimal tuple caught · bare `Rgb(` caught · ban site may name what it bans · out-of-range tuple is not a colour · one literal both matchers see reported once · the real tree passes with **no root argument**.

**Added from this PR (11):** superseded value named with both hexes · live token bound to the wrong name · hex under a name no token declares · correct citation passes *and is counted* · lower-case citation · minified line pairs each name with its own value · wrong value on a shared line named as itself · token named in prose is not a citation · `SELF` suppresses the ban and not the citation check · stray hex in a token-only path · the deck's invented palette rejected.

**Dropped as genuine duplicates (4), all mine:**

| mine | superseded by |
|---|---|
| `B1` a retired hex is banned wherever it appears | #4922's "a banned hex literal is caught" |
| `B2` a ban site may name a retired hex | #4922's "a ban site may name the values it bans" |
| `R1` this repository quotes its own palette | #4922's "the real tree passes, with no root argument" — **stronger**, it exercises the default path instead of passing a root |
| `N1` a tree with no colour in it passes | #4922's "a live hex literal passes" and "a live `Color::Rgb` tuple passes" cover the negative direction with a colour actually present |

**Removed by design (1):** `F1` a fixture corpus is exempt — the `FIXTURES` tuple it tested no longer exists.

`want` gains an optional substring, checked on **both** verdicts, so a guard that passed by scanning nothing cannot satisfy a pass case. #4922's existing call sites are unchanged (`expect-ban` still works; `expect-fail` is a synonym, because a misquote is not a ban).

### Double ablation — both halves proven independently necessary

This is what makes it a union rather than a replacement. Disabling **#4922's** matcher:

```
$ # RUST_RGB loop disabled
  ✗ a banned Color::Rgb tuple is caught — the offending value was NOT caught
  ✗ a bare Rgb( is caught, not only Color::Rgb(
  19 passed, 2 failed
```

(The decimal case still passes, exactly as #4922's own comment predicts — `RGB_FUNC` is `IGNORECASE`, so it reads `Rgb(11, 11, 12)` as an `rgb(` call. That case is not a witness for `RUST_RGB`; the two either side of it are.)

Disabling **my** matcher:

```
$ # CITATION loop disabled
  ✗ a superseded value is named with both hexes
  ✗ a live token bound to the wrong name is flagged
  ✗ a hex under a name no token declares is flagged
  ✗ a correct citation passes and is counted — passed but did not report '1 token citation(s) quote the palette'
  ✗ a lower-case citation is read as the same value — passed but did not report '1 token citation(s)…'
  ✗ a minified line pairs each name with its own value — passed but did not report '3 token citation(s)…'
  ✗ the wrong value on a shared line is named as itself
  ✗ SELF suppresses the ban and not the citation check
  13 passed, 8 failed
```

Three of those eight fail on the **pass** side — the substring pinning catching a guard that scanned nothing rather than one that stopped reporting.

Union, unablated:

```
$ ./scripts/test-tokens.sh
  21 passed, 0 failed
$ python3 scripts/check-tokens.py
tokens: no retired hex in the tree; 3 token-only path(s) clean; 141 token citation(s) quote the palette
```

`make guards-fast` is green on the prepared union, `prose` included — it caught a `load-bearing` in my own new comment and I deleted it rather than baselining.

### Unchanged from this PR

The 20 citation fixes, the deck's four invented values on `banned.values`, and the `docs-guards.yml` wiring are carried over untouched. The `docs-guards.yml` change does not overlap #4922, which does not touch that file.

I will do the rebase and force-push as soon as #4922 merges, then re-run the suite on the rebased head and post the result here.

macanderson:
## Base check: the prepared union is current, and the whole stack replays clean

Re-verified after #4898 landed, because a merge under a prepared artifact is how one goes stale without saying so.

**#4922 has already been rebased.** Its head is still `a5eac65883899916f5ee73144e5114f112e4a547`, and `0fbcd5b82` (#4898, merged 03:08:41Z) is an **ancestor** of it — so the `guard-self-tests.yml` collision between those two is already resolved on #4922's side, and the tip I built this union against is the post-#4898 one. The union base was never stale.

**Only one commit has landed on `main` since my merge-base**, and it touches nothing here:

```
$ git log --oneline 3bbc75bbc..origin/main
8c1d23f21 chore(release): sync versions to 0.9.229 [skip release] (#4934)
   Cargo.lock | Cargo.toml | packaging/homebrew/stella.rb
```

(The merge-base is `3bbc75bbc` — #4909, "retire the brand gradient" — which is already an ancestor of this branch, so its token changes are in the tree the suite was verified against.)

**Trial rebase of the full stack onto today's `main`, then re-run:**

```
$ git rebase origin/main            # replays #4922's two commits + this delta
Successfully rebased and updated refs/heads/rebase-trial.

$ ./scripts/test-tokens.sh
  21 passed, 0 failed

$ python3 scripts/check-tokens.py
tokens: no retired hex in the tree; 3 token-only path(s) clean; 141 token citation(s) quote the palette
```

Zero conflicts across all three commits. The trial branch was discarded rather than pushed — pushing it would put #4922's commits on the remote under new SHAs, which is exactly the confusion to avoid while that PR is still open.

So the plan is unchanged and now has evidence behind it: hold until #4922 merges, then

```
git rebase --onto origin/main fix/tokens-rust-tuples-4910 fix/token-citations-3653-union
git push --force-with-lease origin HEAD:fix/token-citations-3653
```

and I post the 21-case result on the rebased head here.

macanderson:
## Rebased onto `main` with #4922 landed — union is live, 21/21

Force-pushed `549d2c59e`. #4922 squash-merged as `1b9c9172`, so the rebase took only this delta:

```
git rebase --onto origin/main a5eac6588 union-prep
```

Merge-base re-derived rather than reused. `main` had moved a long way since the union was prepared — #4914, #4920, #4921, #4923, #4927, #4928, #4938, #4943, #4925, a version sync, **and #4935**.

### One conflict, and it was the one this PR predicted

`.github/workflows/docs-guards.yml`, in four places — because **#4935 landed first** and added its own step to the same file. This PR's body called that out before either merged; here is the resolution, all four as a **union**:

| conflict | resolved |
|---|---|
| header comment | kept "prose **and colour**" — the superset |
| guard list | **both** entries kept, renumbered: `9.` the brand art's sources (#4935), `10.` the four colour rulers (#3653) |
| `pull_request` trigger paths | **both** sets: `README.md` (#4935) **and** the five colour scripts + `contrast-baseline.txt` + `design/tokens/**` |
| `push` trigger paths | same union |

Nothing was dropped in either direction. Verified structurally rather than by eye — the job now carries **15 steps**, with both PRs' additions present:

```
- docs prose spells the wordmark lowercase
- the colour system is the one in the tree          <- #3653
- no two web semantic roles sit within 30° of each other   <- #3653
- every licensed colour pairing is readable on its ground  <- #3653
- every var() in a token sheet resolves             <- #3653
- the documented gate is the real gate
- ...
- the brand art's install line and tagline match their sources  <- #4935
```

### Result on the rebased head

```
$ ./scripts/test-tokens.sh
check-tokens ban check:
  ✓ a banned hex literal is caught
  ✓ a live hex literal passes
  ✓ a banned Color::Rgb tuple is caught
  ✓ a live Color::Rgb tuple passes
  ✓ a tuple written in decimal is caught (by either matcher)
  ✓ a bare Rgb( is caught, not only Color::Rgb(
  ✓ a ban site may name the values it bans
  ✓ an out-of-range tuple is not read as a colour
  ✓ a literal both matchers see is reported once
  ✓ the deck's invented palette is rejected as one the kit never defined

check-tokens citation check:
  ✓ a superseded value is named with both hexes
  ✓ a live token bound to the wrong name is flagged
  ✓ a hex under a name no token declares is flagged
  ✓ a correct citation passes and is counted
  ✓ a lower-case citation is read as the same value
  ✓ a minified line pairs each name with its own value
  ✓ the wrong value on a shared line is named as itself
  ✓ a token named in prose with no value is not a citation
  ✓ SELF suppresses the ban and not the citation check
  ✓ a stray hex in a token-only path is flagged

check-tokens, the real tree:
  ✓ the real tree passes, with no root argument

  21 passed, 0 failed
```

```
$ python3 scripts/check-tokens.py
tokens: no retired hex in the tree; 3 token-only path(s) clean; 141 token citation(s) quote the palette

$ python3 scripts/gen-tokens.py --check
tokens: 33 validated, 2 artifacts in sync

$ python3 docs/brand/social/build_social.py --check      # #4935's guard, still green beside mine
ok — install line, tagline across 6 copies, and 5 layouts fit
```

### `make guards-fast`, green against the new base

```
check-file-size: OK — 1765 watched files, 19 grandfathered over 1500 lines,
                 and nothing went over by this change. Judged against 080161dbb.
check-gate-parity: OK — 47 (forty-seven) gate steps, named in AGENTS.md and
                 CONTRIBUTING.md, each run by a workflow.
check-prose: OK -- 6386 grandfathered construction(s) in 1454 file(s), none added.
shellcheck / check-css-vars / check-lockfile-sync / cargo fmt --check: OK
```

`check-file-size` is judged against `080161dbb`, the new base — #4920's change to that script landed in this window and its new "files nearing their ceiling" report is advisory, not a failure. No baseline raised anywhere.

### One thing worth knowing before merge — #4952 applies to this guard

I ran the repro rather than assuming, and reported it there: `check-tokens.py` enumerates with `git ls-files`, so an **untracked** new file is invisible to *both* the ban and the new citation check, and the guard prints `no retired hex in the tree` and exits 0. `git add` the identical file and it exits 1 naming both violations.

That is a fifth guard for #4952's list and a pre-existing property of the enumeration, not something this PR introduces — the citation check simply inherits it. Flagging it here so the merger knows the local-vs-CI asymmetry exists for this guard; the fix belongs in #4952, where I have posted the measurement and a note about keeping `SELF` path-matching correct through any widening.

Everything in this PR was `git add`ed before the guard run above, so those numbers are over the real tree.

Issue #4679, opened 2026-08-24
Epic: repo prose reads plainly — the prose-gate baseline reaches zero


macanderson:
## Outcome

Repo prose — code comments, doc comments, `.md` pages, error strings — reads
in plain language, with `scripts/prose-baseline.txt` at zero. No banned
construction any `make prose` rule currently recognizes survives, and no
grandfathered debt is left for the next sweep to rediscover.

## Why one epic and not three loose issues

All three children write to the same shared cell — `scripts/prose-baseline.txt`
— and shrink it under the same down-only ratchet (`make prose-update` refuses
to raise a count or add a file). Splitting the burn-down by category is
sensible execution; tracking it as three unrelated issues is not, because a
maintainer asking "is the prose debt gone yet" needs one place to look:

- #4326 — the original 729-hit debt (`tired-metaphor`, `enumerative-announcement`,
  and siblings) from before PR #4599 added new categories.
- #3751 — the single largest lexical offender, "invariant" (863 occurrences,
  380 files), called out on its own because a blind find-and-replace would
  break AGENTS.md's numbered rule citations.
- #4600 — the full baseline as it stands after PR #4599's three new banned
  patterns (`rust-jargon`, `honesty-declaration`, `interface-jargon`), 4283
  hits across 1220 files, which subsumes whatever #4326 has not yet closed.

## Definition of done

- `scripts/prose-baseline.txt` is empty.
- `make prose` and `make prose-update` both green with no new entries added
  by any child.
- AGENTS.md's ten numbered architecture rules keep their citation addresses
  intact throughout the "invariant" rename (#3751) — `scripts/check-invariants.sh`
  reports the same citation count before and after.
- Renamed identifiers are prose-only: no code identifier, file name, gate step
  name, or AGENTS.md rule address changes as a side effect.

## Children

#4600 #4326 #3751

Issue #4600, opened 2026-08-24
Burn down the grandfathered prose-gate debt so repo text reads plainly


macanderson:
## Problem

PR #4599 taught the prose gate (`make prose`) new banned patterns: `honest`/`honestly`/`honesty`, insider interface names (say "interactive mode" / "non-interactive mode"), the words `variant` and `invariant` in prose (say "option" and "rule"), and count announcements. The gate blocks new hits, but 4283 old ones in 1220 files are grandfathered in `scripts/prose-baseline.txt` and still read badly. Counts by pattern: 2129 `rust-jargon`, 1046 `honesty-declaration`, 727 `interface-jargon`, the rest older patterns.

## What done looks like

`scripts/prose-baseline.txt` reaches empty. Work file-by-file: run `make prose-report` for the exact lines and suggested fixes, rewrite the prose, then `make prose-update` to retighten. Any slice of files is a valid PR; largest offenders first pays best.

Constraints:
- Prose only. Never rename code identifiers, file names, gate step names (the `invariants` gate step and `scripts/check-invariants.sh` keep their names), or AGENTS.md's numbered rule addresses — those renames are a separate decision.
- Backticked spans and fenced blocks don't count; citing a banned word to ban it is fine.
- The baseline only shrinks. `make prose-update` refuses anything else.

Refs #4599. From the 2026-08-23 prose sweep, which also rewrote all 175 open issues to the same standard.

## Definition of done

The done conditions this issue already states, as boxes the gate can read.
Nothing here is new scope, and nothing is ticked by the migration — the
gate reads one heading and this issue used another (macanderson/oxagen#1407).

- [ ] `scripts/prose-baseline.txt` reaches empty.
- [ ] The work is prose only — no code identifier is renamed to satisfy the gate.
- [ ] Each slice runs `make prose-report`, rewrites the flagged lines, then `make prose-update` to retighten.

macanderson:
Related: #3751 already tracks replacing the word "invariant" across the repo (863 uses in 380 files). The prose gate from #4599 now blocks new uses; that issue and this one are the same campaign — #3751 for the invariant rename, this one for the full baseline burn-down.

macanderson:
New epic filed: #4679. The standing orphan tracker (#2778) had checked this
issue against #3763 (brand kit enforcement) and correctly rejected it, but
never checked it against #4326 and #3751 — both open, unparented, and
targeting the same `scripts/prose-baseline.txt` shared cell. That's three
[...]

PR #4939, opened 2026-08-26
docs(prose): plain language across the design documents

macanderson:
Takes `scripts/prose-baseline.txt` from 6397 to **6134** across `docs/` —
specs, ADRs, playbooks, prompt contracts and reference pages.

## What changed

- **`invariant N` → `AGENTS.md #N`** wherever the number addresses AGENTS.md's
  architecture list. No number moved and AGENTS.md itself is untouched.
- **`invariant` with no number → `rule`.**

## The citation that would have been wrong

A mechanical rename would have mis-attributed every numbered citation that
addresses a *different* list. Those keep their qualifier:

- `foundry rule 13` / `foundry rule 3` / `foundry rule 11` — `doc:self-driving-foundry`'s own list
- `§9 rule 3` / `§9 rule 4` — `doc:pipeline-as-plugins` §9's list
- `principles/index.mdx` rule 4, `lifecycle:1262` rule 8 — Oxagen documents
- `docs/brand/cometkit.py`'s `rule 6` — the brand guidelines' list
- `docs/prose-guidelines.md`'s "See rule 10" — that file's own rule 10

`scripts/check-invariants.sh` is what caught the class: `foundry invariant 13`
became `AGENTS.md #13`, which does not exist, and the guard failed. Every other
numbered citation in `docs/` was then checked by hand against the list it
actually names.

Five citations wrote AGENTS.md in backticks (`` `AGENTS.md` invariant #3 ``).
The guard's filter needs the bare form, so those were invisible to it; they
now read `AGENTS.md #3` and resolve.

`docs/prose-guidelines.md`'s own before/after example for the `deliberately`
rule keeps its banned word — backticked, the way that file's other examples
already are, so the guard reads it as the citation it is.

## Two generated trees left alone

- `docs/tools/` is written by `crates/stella-cli/src/tool_docs.rs` from
  `crates/stella-tools/src/catalog.rs`. Editing the output alone fails
  `make tool-docs`, so it is reverted here and its prose belongs to the crate
  batch.
- `docs/wire/` is already excluded from the prose scan for the same reason.

## Evidence

| check | before | after |
|---|---|---|
| `make prose` | 6397 | **6134** |
| `check-invariants` citations resolving | 40 | **110** |
| `make tool-docs` | pass | pass |
| `check-doc-links` | exit 0 | exit 0 |
| `make line-citations` | 260 | 260 |
| `make brand-case` | pass | pass |
| `make guards-fast` | pass | pass |

Refs #4600
Refs #4679

## Summary by Sourcery

Improve the clarity and plain-language consistency of the design documentation without changing the underlying architecture or behavior.

Enhancements:
- Simplify prose across design documents by removing unnecessary uses of “deliberately” and replacing generic “invariant” terminology with clearer “rule” language.
- Clarify numbered architectural citations so references to AGENTS.md and document-specific rule lists resolve to the correct source.

Documentation:
- Revise documentation across specifications, ADRs, playbooks, prompts, papers, references, and brand materials to use plainer, more direct language while preserving the meaning of cited rules.


macanderson:
Pairs with #4956. This PR rewrites `docs/adr/0013-session-artifact-boundary.md`'s citation of the principles page's section — `§ "The seven invariants, briefly"` becomes `§ "The seven rules, briefly"` — and #4956 renames the section itself on `website/content/docs/principles/index.mdx`. Either alone leaves the pair off by a section title, so they want to land together or close to it.

PR #4947, opened 2026-08-26
fix(scripts,website): the website's diagnostic-code table is checked against the emit sites

### LINKED ISSUE #3045: The website's diagnostic code table is copied by hand and can go out of sync with the real list
## Problem

`website/content/docs/diagnostics.mdx` (added in #3043) carries a hand-picked,
hand-transcribed subset of the diagnostic codes table as a plain markdown
table. `docs/reference/diagnostics.md` — the source of truth — is itself
**generated** from the emit sites by `make diag-reference`
(`scripts/diagnostic-codes.py`), and the gate (`make diagnostic-codes`) fails
if the two ever disagree.

The website copy has no such guarantee. It will drift the first time a code is
added, renamed, or its level changes — exactly the failure mode the generator
exists to prevent for the in-repo copy.

## What to build

Generate the website table's rows from the same source
`scripts/diagnostic-codes.py` reads (the emit-site scan), rather than hand
copying prose out of `docs/reference/diagnostics.md`. Options worth weighing:

- Extend `scripts/diagnostic-codes.py` with a second output mode that emits
  the curated/reader-facing subset as MDX-embeddable data (e.g. a JSON file
  the page imports), keeping the "hand-picked for troubleshooting symptoms"
  curation but generating the per-code level/description cells.
- Or render the *complete* table on the website page instead of a curated
  subset, generated in full, and let the page's own prose do the
  symptom-oriented curation/linking instead of the table itself.

Either way: a `make` target (or an extension of `make diag-reference`) should
regenerate the website's copy, and `make gate` (or `docs-guards.yml`, since
`website/content/docs/**` is outside `ci.yml`) should fail when the website
table and the emit sites disagree, the same way `make diagnostic-codes`
already does for `docs/reference/diagnostics.md`.

## Verify

Add or rename a diagnostic code's level in a facet enum; confirm the website
page's table is now stale and the new gate catches it.

## Definition of done

- The website diagnostics page's code table is produced from the emit-site
  scan, not transcribed by hand.
- A gate step catches drift between the website table and the tree, matching
  `diagnostic-codes`'s guarantee for the in-repo reference.
- `docs/reference/diagnostics.md` remains the canonical, fully generated
  reference; the website page may still curate *which* codes it surfaces, but
  not their content.

## Constraints

- Source: `scripts/diagnostic-codes.py`, `docs/reference/diagnostics.md`,
  `docs/spec/diagnostics.md` §11.
- Related: #3043 (added the page this table lives on).

macanderson:
## The gap

`docs/reference/diagnostics.md` is generated from the emit-site scan and
`make diagnostic-codes` fails when it drifts.
`website/content/docs/diagnostics.mdx` carried a second table of the same
facts, transcribed by hand, that **no gate read at all**. The page's own
closing paragraph admitted it and pointed here.

## What ships

`scripts/diagnostic-codes.py check` now holds that table to the same scan.

**Checked** — the two columns that are facts about the tree:

- every code the table names is still emitted;
- every `Level` cell matches the levels the emit sites actually use.

**Not checked, on purpose** — which codes the page surfaces, and the
"What it means" prose beside them. Those are editorial: picked and written for
a reader chasing a symptom. #3045 offered generating the whole table as an
option; that would replace a good page with a worse one, and the split follows
the precedent `crates/stella-cli/src/config/tests/docs_sync.rs` already set for
`PROVIDER_CATALOG` — pin the machine-checkable columns, exempt the prose ones,
for the reason stated there: asserting on them "would force the docs to speak
the enum's language rather than the reader's".

Two shapes the table already uses are read as sets rather than as single
values: a row may name several codes (`agent.turn.parked` /
`agent.turn.woken` share one description) and a cell may name several levels
(`debug / warn`, which the generated reference writes as `debug or warn`).

The table is located by its `## Reading the codes` heading rather than by
position — the page opens with an unrelated three-plane table and will grow
others. A missing heading or an emptied table **fails**, rather than reading as
nothing-to-check: a guard that goes quiet when its subject moves is the shape
this one exists to prevent.

## Witness — four ablations

Each makes the answer **wrong**, not merely different, and each is caught:

```
### A — a level cell no longer matches the emit site
    (| `agent.budget.denied` | error |  ->  | ... | warn |)
check-diagnostic-codes: FAIL — website/content/docs/diagnostics.mdx says
  `agent.budget.denied` is emitted at warn; the tree emits it at error
  (crates/stella-cli/src/diag_bridge.rs).
EXIT=1

### B — the table names a code the tree does not emit
    (`agent.budget.denied` -> `agent.budget.refused`, in the mdx only)
check-diagnostic-codes: FAIL — website/content/docs/diagnostics.mdx documents
  `agent.budget.refused`, which nothing emits any more.
EXIT=1

### C — the anchor heading moves out from under the guard
    (## Reading the codes -> ## Reading a diagnostic code)
check-diagnostic-codes: FAIL — website/content/docs/diagnostics.mdx: no
  `## Reading the codes` section — the guard cannot find the table.
EXIT=1

### D — a multi-level cell loses a level
    (| `agent.tool.result` | debug / warn |  ->  | ... | warn |)
check-diagnostic-codes: FAIL — website/content/docs/diagnostics.mdx says
  `agent.tool.result` is emitted at warn; the tree emits it at debug or warn
  (crates/stella-cli/src/diag_bridge.rs).
EXIT=1
```

### And the ablation that proves the *old* guard could not have caught it

This is #3045's own "Verify" step: rename a code in a facet enum and confirm
the website page is now stale. Renaming `agent.budget.denied` →
`agent.budget.refused` in `crates/stella-cli/src/diag_bridge.rs`:

```
check-diagnostic-codes: FAIL — `agent.budget.refused` is emitted (...) but has no section in the reference.
check-diagnostic-codes: FAIL — the reference documents `agent.budget.denied`, which nothing emits any more.
check-diagnostic-codes: FAIL — the reference's generated facts are stale for the current tree.
check-diagnostic-codes: FAIL — website/content/docs/diagnostics.mdx documents `agent.budget.denied`, which nothing emits any more.
```

Now do what the old guard told you to do — `make diag-reference`, which is the
whole remedy it offers:

```
check-diagnostic-codes: FAIL — `agent.budget.refused`'s section is still the placeholder; ...
check-diagnostic-codes: FAIL — website/content/docs/diagnostics.mdx documents `agent.budget.denied`, which nothing emits any more.
```

The in-repo reference repaired itself, and the website line is still standing.
That surviving line is the entire content of #3045 — before this PR it did not
exist, and the regeneration would have left the tree green with the website
wrong.

## Hosting

No new gate step. `diagnostic-codes` is already in `GATE_GUARDS_FAST`, so this
costs no `Makefile`, `AGENTS.md` or `CONTRIBUTING.md` edit and adds no shared
cell for two PRs to collide on.

It is added to `docs-guards.yml` because `ci.yml`'s Rust job is skipped for a
diff confined to `website/**` — exactly the diff that edits this table. That is
the hole #4632 found for `website-inputs`, one file over.
`scripts/website-rust-inputs.txt` is deliberately not touched: that guard reads
`website/` literals out of `crates/**/*.rs`, this reader is Python, and
declaring the path as a `read` entry would drag the whole Rust job onto every
`.mdx` edit for no gain.

```
$ bash scripts/check-gate-parity.sh
check-gate-parity: OK — 47 (forty-seven) gate steps, named in AGENTS.md and CONTRIBUTING.md, each run by a workflow.

$ bash scripts/check-website-inputs.sh
check-website-inputs: OK — 8 declared website path(s), 4 of them Rust test inputs, all present.

$ ./scripts/check-diagnostic-codes.sh
check-diagnostic-codes: OK — 46 codes emitted, all documented; 12 surfaced on the website carry the level the tree emits.

$ make guards-fast     # no FAIL lines
$ python3 scripts/check-prose.py
check-prose: OK -- 6393 grandfathered construction(s) in 1454 file(s), none added.
```

And the new host is confirmed live on this PR's own run — the `docs guards` job
carries the step and it passed:

```
docs guards  every emitted diagnostic code is documented
  check-diagnostic-codes: OK — 46 codes emitted, all documented; 12 surfaced on
  the website carry the level the tree emits.
```

No `#[test]` was renamed or removed. No baseline was added or raised.

## Left behind, and filed

There is still no hermetic self-test for this guard — nothing proves it can
*fail*, which is the argument `guard-self-tests.yml` makes at length for the
other twenty-seven suites. The four ablations above are a session's evidence,
not a standing property. Filed as #4948 rather than folded in: the script's `REPO`
constant is derived from `__file__` and its scan walks `crates/*/src/**/*.rs`,
so a self-test needs a synthetic tree, which is its own change.

Closes #3045

## Summary by Sourcery

Keep the website diagnostic-code table synchronized with the diagnostic emit sites and enforce the check in documentation CI.

New Features:
- Validate the website’s curated diagnostic-code table against emitted codes and levels, including multi-code and multi-level entries.

Bug Fixes:
- Prevent stale, renamed, removed, or incorrectly leveled diagnostic entries from remaining undetected in the website documentation.

Enhancements:
- Require the website diagnostic table to remain discoverable and non-empty through its section anchor while leaving editorial selections and explanatory prose unchecked.

CI:
- Run the diagnostic-code validation in the documentation guards workflow, including website-only changes that bypass the Rust CI job.

Documentation:
- Update the diagnostics page to explain which table facts are machine-checked and which content remains editorial.


macanderson:
Answering Sourcery's ❌ row, which is a fair reading of #3045's first bullet and a scope decision I made on purpose rather than an oversight.

> **❌** Produce the website diagnostics table from the emit-site scan, while allowing the page to curate which diagnostic codes it displays but not manually transcribe their factual content. — *The PR leaves the website table hand-maintained and does not generate its rows or cells.*

Correct as stated: the cells are still typed by hand. Here is why, and what a maintainer would need to say to overturn it.

## What #3045 actually asked for, and which part is met

The issue offered two options ("Options worth weighing") and then set a definition of done with three clauses. Clause 2 is the operative one:

> A gate step catches drift between the website table and the tree, matching `diagnostic-codes`'s guarantee for the in-repo reference.

That is met, and the PR body shows the ablation proving the old guard could not have caught it. Clause 1 says "produced from the emit-site scan, not transcribed by hand" — which this PR satisfies in effect and not in mechanism: the cells are typed, and they **cannot be wrong and stay green**.

## Why checking, not generating

Two of the three columns are prose a generator has no business writing. The third is one word.

- **`What it means`** is written for a reader chasing a symptom, and it does not paraphrase the reference — compare the page's `agent.tool.result` row ("Emitted at `warn` only on failure, so a default `stderr` filter shows failures and only failures") against the reference's paragraph for the same code. Generating it would mean either dropping it or extracting a first sentence from prose written for a different context.
- **`Code`** is the curation the issue explicitly preserves — 12 of 46, hand-picked for `/docs/guides/troubleshooting`.
- **`Level`** is the only generatable cell, and it is one word per row.

So the generator would have to reproduce `parse_reference`'s prose-preservation machinery — locate each row, replace one cell, keep the other two — plus a `make` target and a committed-diff step, to save eleven words of typing. That is more moving parts than the thing it protects, and every one of them is a new way for the page to break.

There is also a precedent for the split, cited in the PR body: `crates/stella-cli/src/config/tests/docs_sync.rs::provider_cards_match_the_registry` pins `PROVIDER_CATALOG`'s `id`/`env`/`defaultModel` and deliberately exempts `dialect`/`blurb`, for the reason stated there — asserting on them "would force the docs to speak the enum's language rather than the reader's". Same shape, same call.

## What would change my mind

If a maintainer wants clause 1 satisfied in mechanism, the cheap version is a fourth mode on `scripts/diagnostic-codes.py` — `website-write`, rewriting only the `Level` cell of each row whose first cell names a code, leaving `Code` and `What it means` untouched — wired to `make diag-reference` so one command regenerates both copies. The parser this PR adds already locates the rows, so it is a small addition on top rather than a redesign. Say the word and I will open it as a follow-up; I have not filed it speculatively, because filing an issue for a change nobody has asked for is its own kind of noise.

Sourcery's other two rows are ✅ and I have nothing to add to them.

PR #4951, opened 2026-08-26
docs(stella-plugin,stella-runtime,stella-serve): plain language in the plugin socket's comments

macanderson:
Takes `scripts/prose-baseline.txt` from 6397 to **6027** across the
wrapper-socket crates — manifest parsing, the wire types, the transports, and
the headless engine server.

## What changed

- **`invariant N` → `AGENTS.md #N`.** Numbers untouched; `check-invariants.sh`
  resolves this form (40 on `main`, **122** here).
- **`variant` → the thing it names.** An enum `case` for a closed set, and a
  **`pipeline id`** where the word meant which wrapper ran (`pipeline_variant`,
  #3388). `case id` was the wrong translation and was fixed by hand across
  `wire.rs`, `program.rs`, `dispatch.rs` and `wrapper/error.rs`.

## Two things the mechanical pass got wrong, and how they were caught

- **A backticked field reference.** `` `StageProgram::variant` `` became
  `` `StageProgram::case` `` — a name that does not exist. The prose guard skips
  backticked spans when it *scans*, but the rewrite ran over the whole line, so
  a code span on a line that also had a real hit was rewritten too. Caught by
  diffing every changed line's backticked and bracketed spans against `main`;
  that check now reports zero altered spans on this branch.
- **Non-comment lines.** Every line the pass touched that is not a comment was
  restored to its `main` content, so no identifier changes.

## Generated artifacts

`docs/wire/wrapper.*` is built from the doc comments this branch edits, so it is
regenerated here with `make wire-schema-update`. The diff is the same prose,
nothing structural.

## Evidence

| check | before | after |
|---|---|---|
| `make prose` | 6397 | **6027** |
| `check-invariants` citations resolving | 40 | **122** |
| `make wire-schema` | pass | pass |
| `cargo fmt --all --check` | clean | clean |
| altered code spans vs `main` | — | **0** |

Refs #4600
Refs #4679


macanderson:
Rebased onto `54ae7ef14` and the baseline regenerated on the rebased tree.

**This PR now carries `crates/stella-observatory/src/accept.rs` as well**, which is a file from a crate not in its title. It has to: that file and `crates/stella-serve/src/accept.rs` are byte-identical duplicates, and `the_two_copies_of_this_policy_have_not_drifted` runs in *both* crates and fails the moment the texts differ. A per-crate partition splits the twin across two PRs, and each side is then red against a main that has the other unedited — which is what reddened this PR and #4954.

#4954 no longer touches the file, and is now a single commit so it cannot replay an edit-and-revert of it over this PR's version. Either order is safe.

Verified on the branch, not only in CI:

```
cmp crates/stella-serve/src/accept.rs crates/stella-observatory/src/accept.rs   # identical
cargo test -p stella-serve --lib accept::         # 7 passed
cargo test -p stella-observatory --lib accept::   # 7 passed
```

`rg -l 'INTENTIONALLY DUPLICATED'` and a search for cross-crate `include_str!` equality assertions both report `accept.rs` as the only twin in the tree.

`make prose`: 6397 -> **4803** on this branch alone.

PR #4954, opened 2026-08-26
docs(stella-diag,stella-transcript,stella-observatory,stella-mcp,stella-autonomy): plain language in five small crates

macanderson:
Takes `scripts/prose-baseline.txt` from 6397 to **6202** across the diagnostics,
transcript, observatory, MCP and autonomy crates.

## What changed

- **`invariant N` → `AGENTS.md #N`.** Numbers untouched; `check-invariants.sh`
  resolves this form (**40 on `main`, 77 here**).
- **`variant` → the thing it names.** An enum `case`, and a file-suffix **form**
  for TypeScript's `x`/module spellings, where "case" would have been wrong.

## What was left alone

`crates/stella-diag/src/error.rs` documents its macro with
`` `Variant { .. } => "code"` `` — that is the text a caller types, so it keeps
its spelling. The mechanical pass had rewritten both sketches; the code-span
check against `main` caught it, and now reports **zero** altered spans.

Every non-comment line the pass touched was restored to its `main` content.

## Evidence

| check | before | after |
|---|---|---|
| `make prose` | 6397 | **6202** |
| `check-invariants` citations resolving | 40 | **77** |
| `cargo fmt --all --check` | clean | clean |
| altered code spans vs `main` | — | **0** |

Refs #4600
Refs #4679

## Summary by Sourcery

Clarify and simplify documentation across five Stella crates without changing implementation behavior.

Enhancements:
- Simplify prose across the stella diagnostics, transcript, observatory, MCP, and autonomy crates by removing unnecessary qualifiers and using clearer terminology.
- Replace ambiguous invariant references with resolvable AGENTS.md citations and clarify enum and file-suffix terminology without changing code behavior.

Documentation:
- Reduce the prose baseline while preserving intended references, user-facing terminology, and all code spans.


macanderson:
Rebased onto `54ae7ef14`, baseline regenerated on the rebased tree, and **flattened to a single commit**.

The flattening matters. This branch previously edited `crates/stella-observatory/src/accept.rs` and then reverted it in a later commit. Replaying that history onto a base that already carries #4951's edit to the same file **silently reverts it** — I hit exactly that while composing the two, and `make prose` on the composed tree failed with `accept.rs [filler-adverb]: 1 allowed, 4 found`. A single commit whose diff never mentions the file cannot do that.

`git diff origin/main -- crates/stella-observatory/src/accept.rs` is empty on this branch. Both copies of that byte-identical twin travel in #4951.

`make prose`: 6397 -> **4986** on this branch alone.

macanderson:
Rebased onto `e25a3d2b2` — the first `main` that carries #4951's edit to `crates/stella-observatory/src/accept.rs`. That is the exact base that previously triggered the replay, so this is the direct test of the flatten remedy:

```
cmp crates/stella-serve/src/accept.rs crates/stella-observatory/src/accept.rs   # identical
git diff origin/main -- crates/stella-observatory/src/accept.rs                 # empty
rg -c deliberately crates/stella-observatory/src/accept.rs                      # 1
cargo test -p stella-observatory --lib accept::                                 # 7 passed
```

No revert replayed. The composed failure had reported `1 allowed, 4 found` for that file; it is at 1, which is the lone occurrence inside the assertion's message string.

`make prose`: 6397 -> **4231**, which is the composed total for all nine and the last one outstanding. `cargo fmt --all --check` clean.

Issue #4072, opened 2026-08-20
The Observatory's light-mode color scheme was never approved by whoever owns the design


macanderson:
## Problem

The instrument palette's **dark** scheme now sits on the product ramp: every neutral holds OKLCH hue 285.6–285.9 at chroma 0.004–0.011, and the palette block says so and explains why (the accent is a yellow, so a warm ground would sit in the accent's own family).

The **light** scheme was moved with it and now reads the same ramp from the other end — `--text-2` and `--text-3` are literally the palette's own dim and muted values, at hue 285.6–286.4. But the light scheme's *lifted* steps were derived in this change rather than specified, because the palette handed over is dark-only:

```
--void:#E8E8EC; --ground:#FFFFFF; --surface:#F7F7FA; --raised:#FFFFFF;
--hairline:#E9E9EE; --hairline-strong:#D0D0D8;
```

Those are consistent with the dark ramp's hue band, so the immediate "two casts" problem is gone. What remains is that **nobody specified them**. They are one session's derivation, and the same is true of the light-side chromatic stops (`--ok:#006933`, `--warn:#8A3F00`, `--bad:#96213C`, `--identity:#725A00`) and of the terminal's whole `stella-light` paper ramp (`PAPER`, `SNOW`, `PAPER_RAISED`, `PAPER_HAIRLINE`, `INK_EMPHASIS`, `MUTED`, `INK_DIM`, `BRAND_INK`, `BRAND_INK_DEEP`, `GOLD_INK`).

Every one of them is measured and documented where it is declared, and the light scheme is internally coherent — but "a colour someone derived once and wrote a ratio beside" is a weaker claim than "a colour the palette specifies", and the difference should be visible to whoever owns the design rather than buried in a doc comment.

## Files

- `crates/stella-observatory/src/assets/index.html` — `:root[data-theme="light"]`, and the dark palette block it must stay coherent with
- `crates/stella-tui/src/palette.rs` — the "Ground (light)", "Brand (light)" and "Status (light ground)" sections
- `crates/stella-tui/src/theme.rs` — `LIGHT_REMAP`, which maps every dark token to its paper counterpart
- `crates/stella-cli/tests/design_token_parity.rs` — holds all eight instrument surfaces to the Observatory in **both** schemes, so a light value cannot move alone
[...]

PR #4963, opened 2026-08-26
test(stella-cli): export.rs's two light gates are compared to each other

### LINKED ISSUE #4942: export.rs declares its light scheme in two gates and only one is ever checked
## Problem

`crates/stella-cli/src/export.rs` declares its light scheme **twice** — once behind the media query and once behind the attribute:

```
@media (prefers-color-scheme: light) { :root:not([data-theme="dark"]) { … } }
:root[data-theme="light"] { … }
```

`crates/stella-cli/tests/design_token_parity.rs::canonical()` reads only the second, and says so:

```rust
// The explicit gate, not the media query: identical values, and it is the
// one a reader's own toggle reaches.
let light = declarations(between(&observatory, r#":root[data-theme="light"]"#, "\n}"));
```

"identical values" is an assertion no test makes. The two blocks agree today, held only by `export.rs`'s own comment — *"No colour is defined only inside the media query, so the attribute wins in both directions"* — which is a promise in prose, not a check.

## Why this is not hypothetical

This is the exact failure #4296 fixed one surface over. On the Observatory, the media query kept the pre-v5.0 cool-graphite neutrals while the attribute gate moved to the product ramp. Seventeen roles disagreed, and the worst was an absence: nothing re-pointed `--text-emph`, so an OS-light reader got `#BFC1CC` on paper at **1.68:1**.

Nothing caught it because the parity matrix read the attribute gate alone — and its comment asserted the two were identical, exactly as the comment above does now.

The Observatory is now covered by `crates/stella-observatory/tests/light_mode.rs::the_two_light_gates_declare_one_scheme`. `export.rs` has both gates and no such test.

## Scope — the "every surface" reading is false

A comment on #4072 says the blind spot "applies to every surface in the matrix". It does not. Measured against `design_token_parity.rs::surfaces()`:

| surface | light gate the matrix reads | second light gate? |
|---|---|---|
| `crates/stella-cli/src/export.rs` | `:root[data-theme="light"]` | **yes — unchecked** |
| `crates/stella-transcript/src/html/transcript.css` | `@media (prefers-color-scheme: light)` | no (`data-theme` gate absent by design — the page has no toggle) |
| `docs/benchmarks/index.html` | `:root[data-theme="light"]` | no |
| `docs/benchmarks/terminal-bench-2-1-glm-5-2.html` | `:root[data-theme="light"]` | no |

So `export.rs` is the one remaining hole of this exact shape, and `transcript.css` shows the matrix already reads a media query where that is the only gate — `canonical()` is single-gate, not attribute-only.

Two smaller unchecked regions on the benchmark pages, worth deciding about in the same pass: each has a base `:root{` block and an `@media (prefers-color-scheme:dark)` query, neither of which the matrix reads.

## Definition of done

`export.rs`'s two light gates are compared to each other and a disagreement fails — either by generalising `design_token_parity.rs::canonical()` to extract every light gate a surface declares and diff them, or by a dedicated test on `the_two_light_gates_declare_one_scheme`'s model. The generalised form is preferable: it makes a surface that *grows* a second gate covered automatically, which is how this one arrived.

The witness is the one #4296 used: revert one value in one of the two blocks and the test must name it on the old code.

## Files

- `crates/stella-cli/src/export.rs` — the two gates, and the comment currently standing in for the check
- `crates/stella-cli/tests/design_token_parity.rs` — `canonical()`, `surfaces()`, `ROLES`
- `crates/stella-observatory/tests/light_mode.rs` — `the_two_light_gates_declare_one_scheme`, the shape to copy

## Verify

```
rg -n 'prefers-color-scheme: light|data-theme="light"' crates/stella-cli/src/export.rs
cargo test -p stella-cli --test design_token_parity
```

Noticed while gathering evidence for #4072, where it is recorded as a carry-forward finding rather than a tracked item — filed so it does not close with that issue's palette decision. Refs #4072, #4296.

macanderson:
`crates/stella-cli/src/export.rs` declares its light scheme twice and only one gate was ever checked.

## What was unchecked

`design_token_parity.rs::canonical()` reads `:root[data-theme="light"]`, and said so in a comment: *"The explicit gate, not the media query: identical values"*. `export.rs` made the same promise in its own prose: *"No colour is defined only inside the media query, so the attribute wins in both directions."* Both were true. Neither was a check.

That is the exact position the Observatory was in the day before it drifted — its media query kept the pre-v5.0 cool-graphite neutrals while the attribute gate moved to the product ramp, seventeen roles apart, and the worst was an *absence*: nothing re-pointed `--text-emph`, so an OS-light reader read `#BFC1CC` on paper at **1.68:1** (#4296).

## The approach

The issue offers two shapes and prefers the generalised one, which is what this ships. `scheme_gates()` discovers a surface's gates **by selector** rather than from a per-surface marker list, so a surface that *grows* a second gate is covered the moment it does — which is how this hole arrived in the first place.

Three details that are load-bearing rather than incidental:

- **It compares every custom property, not only `#rrggbb`.** Two gates of one surface are the same author writing the same block twice, so `--accent-wash: rgba(10,10,12,.06)` and `color-scheme` are comparable too — and those are precisely what a copy-paste edit forgets. The parity matrix's own `declarations()` stays hex-only, because it compares across three vocabularies.
- **A gate declaring no custom property is dropped.** The Observatory has `:root[data-theme="light"] .bword-dark{display:none}` and a matching icon swap: those style an element *under* a scheme, they do not declare one.
- **`GATE_CENSUS` pins what the scanner must find.** Without it, "the gates agree" and "the scanner stopped seeing a gate" produce the identical green — and the second is the state this test exists to end.

## Scope

The issue's table is correct: `export.rs` was the one remaining hole of this shape. The **dark** gates on the two benchmark pages — one of the "two smaller unchecked regions" it flags for the same pass — are covered by the same helper at no extra cost: each page declares dark twice, both agree today, and neither was read.

The other region, each benchmark page's base `:root{` block (which *is* that page's light scheme, repeated under the attribute gate), is filed as a follow-up rather than folded in: reading a bare `:root` as a scheme needs a per-surface fact about which scheme a page defaults to, and that is config, not discovery.

## Witness

Reverting `--text-2` to the pre-v5.0 `#4D535A` in **export.rs's media query alone**:

```
running 12 tests
test every_web_surface_agrees_with_the_observatory_palette ... ok
test the_transcript_ramp_steps_agree_with_the_observatory ... ok
test every_wash_is_a_channel_copy_of_a_declared_colour ... ok
test each_surface_declares_one_scheme_per_gate ... FAILED
[... every other test ok ...]

---- each_surface_declares_one_scheme_per_gate stdout ----
1 gate disagreement(s); a reader who uses their OS preference and a reader
who clicked a toggle see different casts of the same page:
  crates/stella-cli/src/export.rs: light `--text-2` is #4d535a under
  `@media (prefers-color-scheme: light)` and #4b4b56 under
  `:root[data-theme="light"]`

test result: FAILED. 11 passed; 1 failed
```

**11 passed** is the measurement: the parity matrix and every other test in the file are green on a tree whose light scheme now ships in two casts. That is the hole, stated as a number.

The absence case — `--control-edge` deleted from the media query, which is #4296's real failure mode:

```
  crates/stella-cli/src/export.rs: light `--control-edge` is declared under
  `:root[data-theme="light"]` and absent from
  `@media (prefers-color-scheme: light)`, which therefore inherits whatever
  the page's default scheme says
```

Clean tree: `12 passed; 0 failed`.

## Also

Both comments that stood in for the check now name the test that performs it.

Closes #4942
Refs #4072, #4296

## Summary by Sourcery

Ensure every discovered color-scheme gate declares an identical custom-property scheme and remains covered by the parity tests.

Bug Fixes:
- Add coverage that detects mismatched or missing custom-property declarations between light and dark scheme gates across web surfaces.

Enhancements:
- Generalize scheme-gate discovery by selector and validate all custom properties, including non-hex values.
- Add gate-count census checks so scanner regressions and newly added gates cannot pass unnoticed.
- Update palette comments to reference the automated parity check instead of serving as unchecked assertions.

Tests:
- Add tests covering light and dark gate parity for the Observatory, CLI export, transcript, and benchmark surfaces.

### DIFF
--- a/crates/stella-cli/src/export.rs
+++ b/crates/stella-cli/src/export.rs
@@ -594,7 +594,12 @@ fn render_dashboard(
      Two gates, the same pattern as docs/brand/css/tokens.css: the OS
      preference unless the page was told "dark" explicitly, and an explicit
      `data-theme` the reader can stamp. No colour is defined only inside the
-     media query, so the attribute wins in both directions. */
+     media query, so the attribute wins in both directions —
+     `crates/stella-cli/tests/design_token_parity.rs`'s
+     `each_surface_declares_one_scheme_per_gate` is what makes that true. It
+     was a sentence here and nothing else until #4942, and the Observatory is
+     what a sentence buys: its two gates drifted seventeen roles apart, and an
+     OS-light reader read #BFC1CC on paper at 1.68:1 (#4296). */
   @media (prefers-color-scheme: light) {{
     :root:not([data-theme="dark"]) {{
       color-scheme: light;
--- a/crates/stella-cli/tests/design_token_parity.rs
+++ b/crates/stella-cli/tests/design_token_parity.rs
@@ -308,10 +308,10 @@ fn scheme_gates(css: &str, scheme: &str) -> Vec<(String, BTreeMap<String, String
 /// What [`scheme_gates`] must find, so a scanner that quietly stops matching
 /// fails instead of passing over nothing.
 ///
-/// This half is not decoration. The bug being fixed is an *unread* gate, so
-/// "the gates agree" and "the scanner found one gate" produce the identical
-/// green — and the second is the state this test exists to end. Each row was
-/// counted by hand against the file it names.
+/// The bug being fixed is an *unread* gate. "The scanner found one gate" is
+/// the state this test exists to end, and it produces the same green as "the
+/// gates agree", so without a census the two are indistinguishable. Each row
+/// was counted by hand against the file it names.
 const GATE_CENSUS: [(&str, &str, usize); 6] = [
     // The two-gate surfaces: an OS-preference query and a toggle attribute.
     (
--- a/crates/stella-cli/tests/design_token_parity.rs
+++ b/crates/stella-cli/tests/design_token_parity.rs
@@ -125,8 +125,10 @@ const ROLES: [&str; 9] = [
 fn canonical() -> Vec<(&'static str, String, String)> {
     let observatory = read("crates/stella-observatory/src/assets/index.html");
     let dark = declarations(between(&observatory, "BEGIN palette", "END palette"));
-    // The explicit gate, not the media query: identical values, and it is the
-    // one a reader's own toggle reaches.
+    // The explicit gate, not the media query. It is the one a reader's own
+    // toggle reaches, and the two carry the same values — which is now a fact
+    // `each_surface_declares_one_scheme_per_gate` checks rather than a claim
+    // this comment makes.
     let light = declarations(between(&observatory, r#":root[data-theme="light"]"#, "\n}"));
 
     let role = |name: &str| -> (String, String) {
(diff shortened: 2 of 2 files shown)


macanderson:
Filed as a handoff rather than folded in: **#4973** — the two `docs/benchmarks/*.html` pages declare their light scheme twice, once in a bare `:root{` and once under `:root[data-theme="light"]`, and neither this test nor the parity matrix reads the first.

It is out of scope here for a structural reason worth recording. `scheme_gates` discovers gates by **selector**, which is what #4942's definition of done preferred; a bare `:root` is not a gate. Reading one as a scheme needs a per-surface fact — which scheme is this page's default? — because on the benchmark pages the base block is light while on the Observatory and `transcript.css` it is dark. That is configuration, not discovery, and the issue names two ways to settle it.

It is not theoretical: while writing #4972's self-test, a fixture that inserted a value into "the light scheme" of `docs/benchmarks/index.html` landed in the bare `:root` block and the case passed over nothing until it was anchored to the gate.

Issue #4966, opened 2026-08-26
theme::MUTED is silver, not the muted tier — rename it or say so at the definition


macanderson:
## Problem

`crates/stella-tui/src/theme.rs` defines:

```rust
pub const MUTED: Color = TEXT_SECONDARY;
```

and `crates/stella-tui/src/palette.rs` defines `pub const TEXT_SECONDARY: Color = token::SILVER;` — `#A9AAB5`, 8.58:1.

So `theme::MUTED` / `theme::muted()` is **`silver`**, and `token::MUTED` is a different colour — `#777782`, 4.32:1 on panel. Two constants, one word, two tiers, and no way to tell from a call site which one a line reached for.

This is finding 3 of #4946. #4946 fixed the two findings that had a single right answer — `comment` was retired (a token nothing painted) and `design/tui-v2/SPEC.md` §6.4 was corrected against the code. This one it deliberately left, because the issue's own definition of done offers two answers and only a maintainer should pick between them.

## What it has already cost

`design/tui-v2/SPEC.md` §6.4 said *"Line number gutter in `dim`"*. `crates/stella-tui/src/diff.rs`'s `gutter` paints `theme::muted()` — i.e. **silver**, two tiers away from `dim` and one from `token::MUTED`. That is the exact mistake the name invites: a call site asking for the muted tier and getting silver. #4946 corrected the spec line to what ships and left the colour alone, because a two-tier difference in a margin is a design call.

An audit by name is also wrong by roughly a factor of two: `theme::MUTED` has 81 production references and `token::MUTED` has 122, and neither number answers "how many sites use the muted tier".

## The decision this needs

#4946's definition of done: *"rename so the name is the tier, or document the divergence at the definition."*

1. **Rename.** `theme::MUTED` → a name that says `silver` or says `text-secondary`. Mechanical but wide: 81 production references, and two of them land in god files (`crates/stella-tui/src/deck_ui.rs`, `crates/stella-cli/src/command_deck.rs`) — a rename adds no lines, so the ratchet permits it, but the churn races every open PR touching those files.
2. **Document at the definition.** Cheaper, and leaves the trap in place for the next call site.

[...]

Issue #4112, opened 2026-08-20
Epic: the colour system's values are specified and its laws are measured, on every surface


macanderson:
## The outcome

Every value in the v5.0 colour system is **specified** by whoever owns the
palette, and every law the system states is **measured by one computation** that
runs on every surface the value reaches — terminal, Observatory and web alike.

## Why an epic now

Three open issues share that outcome and none of them is a value change waiting
for a keystroke: each is a decision that only the palette's owner can make, and
each is currently recorded as prose that nothing checks. Filed separately they
read as three unrelated polish items — a contrast ratio, a hue floor, a light
ramp. Together they are one statement: *the v5.0 system's values and its laws
are partly assertions, and the assertions are the part that has drifted before.*

The bar for creating this was met deliberately rather than assumed. #4059
("the terminal UI ships its SPEC, from one authority over the palette") is the
nearest existing epic and its definition of done is terminal-specific — SPEC 3.1
and 3.2 agreeing on the gold clamp, `stella-tui` importing the generated
constants, the v2 status bar rendering, the golden frames of interactive mode's
renderer. None of these three outcomes appears in it, and two of them are
primarily web/Observatory work. #3763 ("the brand kit is enforced mechanically")
owns the brand *kit* — `docs/brand/`, the guidelines page, the tagline — not the
product palette.

Two siblings from the same 2026-08-20 sweep were resolved rather than adopted,
which is why this epic has three children and not five: #4064 (no warning token)
was decided by #4086, and #4070 duplicated #4063 from the pre-#4101 declaration
site.

## What changed underneath all three

`crates/stella-tui-theme/src/token.rs` is now **generated** from
`design/tokens/stella-tokens.json` (#4101). That JSON is the single place a
value decision can land, which is what makes these three answerable as one
piece of work rather than three edits in three trees.

## Children

- **#4063** — four pairings the palette licenses miss WCAG AA, including the
  SPEC's own stated floor (`muted` on `bg`, 4.47:1 against 4.5). Two of the four
  are under even the 3:1 non-text floor, and one of those is code comments,
  which are content. `scripts/check-contrast.py` measures it and is deliberately
  not a gate step *because it fails*; it joins the gate in the change that
  resolves this. Carries #4070's merged half: the usage rule (`TEXT_DIM` in
  `FALLBACKS` and the progress groove and nowhere else) may be the right answer
  instead of a value change.
- **#4071** — one hue-separation law, three implementations of the metric, and
[...]

PR #4967, opened 2026-08-26
fix(stella-tui-theme,design): retire the comment token and state the syntax palette that ships

### LINKED ISSUE #4946: The comment color is painted by nothing, and two theme aliases name a tier they are not
## Problem

Three findings in `crates/stella-tui`'s colour aliases, all the same shape: **the name says one tier and the value is another.** They are independent of #4063's palette decision — whichever way that goes, these are wrong.

### 1. `comment` is a token nothing paints

`crates/stella-tui-theme/src/token.rs` defines `pub const COMMENT: Color = Color::Rgb(0x56, 0x56, 0x60);`. It has **zero production paint sites**. Every reference in the tree:

```
crates/stella-tui-theme/src/token.rs        the definition, and its row in the token table
crates/stella-tui-theme/src/fallback.rs     token::MUTED | token::DIM | token::COMMENT => Color::DarkGray
crates/stella-tui/tests/status_bar_goldens.rs   token::COMMENT => 'c'   (a test's colour map)
```

Verify:

```
rg -n 'token::COMMENT|COMMENT: Color' crates/
```

### 2. `SYNTAX_COMMENT` is not `comment`

Code comments are painted by `theme::SYNTAX_COMMENT`, applied in `crates/stella-tui/src/syntax.rs`'s `tok_style` for `Tok::Comment`. Following the aliases:

```
crates/stella-tui/src/theme.rs      pub const SYNTAX_COMMENT: Color = palette::TEXT_TERTIARY;
crates/stella-tui/src/palette.rs    pub const TEXT_TERTIARY: Color = token::MUTED;
```

So comments render in **`muted` (#777782, 4.32:1 on panel)**, not in `comment` (#565660, 2.64:1). The alias last moved in #1385.

`design/tui-v2/SPEC.md` §6.4 still specifies the other thing:

> Engine: syntect with a custom theme built from section 3.1 (keywords gold, types silver_type, identifiers text, **comments comment**, strings silver, primitives silver_type).

### 3. `theme::MUTED` is `silver`, not `muted`

`crates/stella-tui/src/theme.rs` defines `pub const MUTED: Color = TEXT_SECONDARY;`, and `crates/stella-tui/src/palette.rs` has `pub const TEXT_SECONDARY: Color = token::SILVER;` — `#A9AAB5`, 8.58:1.

So `theme::MUTED` / `theme::muted()` (**81 production references**) is a *different colour* from `token::MUTED` (**122 references**). Anyone auditing "how many sites use the muted tier" by name gets a number roughly twice the real one, in the wrong tier.

The same trap already bit `design/tui-v2/SPEC.md` §6.4's other line:

> Line number gutter in dim.

`crates/stella-tui/src/diff.rs`'s `gutter` paints `theme::muted()` — i.e. **silver**, not `dim`.

## Why this matters beyond tidiness

- **A guard is measuring an unshipped value.** `scripts/contrast-baseline.txt` holds `comment panel 2.64` — a ratchet entry for a colour no user has ever seen. The pairing that actually governs code comments is `muted on panel 4.32`, already on the same ledger. #4063 is deciding what to do about four sub-AA pairings and one of the four is not real.
- **This is invariant #10's shape one plane over.** A declared token that nothing consumes is allowed, but only as a *declared* gap — not as a silence. `crates/stella-tui-theme/src/token.rs`'s table has no posture column, so nothing asked "what paints this?" when `comment` was added.
- **An alias that lies makes every census wrong.** The 81-vs-122 split above is exactly the sort of thing that turns a costing into a five-fold error.

## Definition of done

Pick one per finding and make the tree and the spec agree:

1. **`comment`** — either wire it (point `SYNTAX_COMMENT` at `token::COMMENT`, which is what §6.4 specifies, and accept or move its 2.64:1), or retire it from the token table and drop its ratchet row. Do **not** leave a token with no consumer and no declared gap.
2. **`SYNTAX_COMMENT` and `theme::MUTED`** — rename so the name is the tier, or document the divergence at the definition. `theme::MUTED` aliasing `silver` is the more dangerous of the two because `muted` is a real token with a different value.
3. **`design/tui-v2/SPEC.md` §6.4** — its two claims (comments in `comment`, gutter in `dim`) are both false against `syntax.rs` and `diff.rs`. Fix whichever side is wrong, in the same PR.

Witness: a test asserting `theme::SYNTAX_COMMENT` is the colour §6.4 names, which fails on the old code.

## Constraints

- `crates/stella-tui/tests/status_bar_goldens.rs` maps colours to characters (`token::DIM => 'd'`, `token::COMMENT => 'c'`); a retirement has to update that map, and the goldens need reading rather than blessing.
- If `comment` is retired, `scripts/check-contrast.py`'s `PAIRINGS` and `scripts/contrast-baseline.txt` both lose a row. `make contrast-update` drops a cleared pairing but **refuses to lower a floor**, so read its output rather than assuming.
- `crates/stella-tui-theme` is a near-leaf (`ratatui` only); do not add a dependency to it.
- Do not raise any baseline to make a gate green.

Noticed while measuring #4063; posted there at https://github.com/macanderson/stella/issues/4063#issuecomment-5420162133 and filed separately because none of it waits on that palette decision. Refs #4063, #4058, #4112.

macanderson:
Three findings, all the same shape: the name says one tier and the value is another.

## 1. `comment` is retired, not wired

`token::COMMENT` `#565660` had zero paint sites. The issue measured the Rust half; the CSS half is the same — `--st-comment` is declared on four surfaces and `var(--st-comment)` appears nowhere in the tree.

The definition of done offers wire-it or retire-it. Wiring it is what `SPEC.md` §6.4 specified, and it would drop code comments from **4.32:1 to 2.64:1** on panel — a visible regression, to ship a colour nothing had asked for. Retirement it is.

It is **banned** rather than merely deleted. Deleting the token alone leaves `#565660` sitting in three design renderings and two brand tables, where the next reader reads it as live; the ban is what forces those clean. The renderings now paint code comments in `#777782`, which is what the TUI actually draws.

### The ratchet retightens

`scripts/contrast-baseline.txt` held `comment panel 2.64` — a floor on a value no reader has ever seen, and one of the **four** sub-AA rows #4063 is deciding about. A quarter of that decision was about nothing.

The guard named the remedy itself:

```
check-contrast: FAIL
  comment on panel: baselined, but it clears its threshold now —
  run `make contrast-update` so the ratchet retightens
```

```diff
 # This file records the debt #4423 found; it is meant to reach empty.
-comment panel 2.64
 dim bg 2.30
 muted bg 4.47
 muted panel 4.32
```

Four pairings to three, **no floor lowered**. `check-contrast: 21 licensed pairing(s), 3 held by the ratchet, none darker than it allows.`

## 2. §6.4's syntax sentence was false in six of six roles

The issue found one clause wrong. Reading `theme.rs` against it, all of them are:

| §6.4 said | ships |
|---|---|
| keywords gold | `SYNTAX_KEYWORD` = the bright neutral `#BFC1CC` |
| strings silver | `SYNTAX_STRING` = `#93D896` |
| primitives silver_type | `SYNTAX_NUMBER` = violet `#8F70E8` |
| comments `comment` | `SYNTAX_COMMENT` = `#777782` |
| types silver_type | `SYNTAX_TYPE` = teal `#2FD3C6` |
| — | `SYNTAX_FUNCTION` = magenta `#E4408F` |

The last two arrived with the grammar-backed lexer (#4283), which the sentence predates entirely — it never admitted a type or a function position exists.

So the sentence is now a **table between markers**, and `crates/stella-tui/tests/spec_syntax_palette.rs` holds every row to its constant. It reads `SPEC.md` as text rather than generating it: a spec that is a projection of the code cannot disagree with the code, and disagreeing is the whole job of one.

## 3. §6.4's gutter line was wrong by two tiers

*"Line number gutter in `dim`"* — `diff.rs`'s `gutter` paints `theme::muted()`, and `theme::MUTED` aliases `palette::TEXT_SECONDARY` = `token::SILVER` `#A9AAB5`. Not `dim` `#4B4B56`, and not `token::MUTED` `#777782` either.

The line is corrected to what ships rather than the gutter recoloured to what it said: two tiers of difference in a margin is a design call, not a typo, and it is not mine to make.

Finding 3 — the alias that makes this trap easy to fall into — takes the definition of done's second option here (*"document the divergence at the definition"*) and is filed as **#4966** for the rename, because the issue's own wording leaves that to a maintainer and it is 81 production references through two god files. `theme::MUTED`'s doc comment now states which tier it is, which it is not, and what it has already cost.

## Witness

The comment role, with §6.4's table restored to the retired token's value:

```
---- the_spec_states_the_syntax_palette_that_ships stdout ----
design/tui-v2/SPEC.md §6.4 disagrees with the highlighter in 1 role(s):
  §6.4 says `theme::SYNTAX_COMMENT` is #565660; it is #777782

Move the code or move the spec — but a reader has to be able to trust one of them.

```

The absence half, with the `SYNTAX_TYPE` row deleted — #4283's own shape:

```
---- every_syntax_constant_has_a_row stdout ----
the highlighter paints 1 role(s) design/tui-v2/SPEC.md §6.4 does not
mention: SYNTAX_TYPE

```

Clean tree: `2 passed; 0 failed`.

## Guards

```
check-contrast: 21 licensed pairing(s), 3 held by the ratchet, none darker than it allows.
tokens: no retired hex in the tree; 3 token-only path(s) clean
tokens: 32 validated, 2 artifacts in sync
check-hue-separation: OK — 12 role pairs across 2 web schemes, all at or above 30° OKLCH
check-css-vars: OK -- 2 token sheet(s), every var() resolves.
scripts/test-tokens.sh: 10 passed, 0 failed
check-prose: OK -- none added.   check-line-citations: OK -- none added.
cargo test -p stella-tui-theme: ok
cargo test -p stella-tui --test status_bar_goldens: 15 passed
```

`crates/stella-tui-theme/src/token.rs` and `design/tokens/stella-tokens.css` are `make tokens-update` output, not hand edits.

## Deleted tests

None. `token::COMMENT => 'c'` is removed from `status_bar_goldens.rs`'s colour map (no golden used `c`) and `token::COMMENT` from `fallback.rs`'s `ansi16` arm; both are lines inside tests, not tests.

Closes #4946
Refs #4063, #4966, #4112

## Summary by Sourcery

Retire the unused comment token and align the documented design palette with the colors and syntax roles implemented by the TUI.

Bug Fixes:
- Retire the unused `comment` color token and remove its generated palette, CSS, documentation, rendering, and fallback references.
- Correct the TUI specification to document the syntax palette and line-number gutter colors that actually ship.
- Tighten contrast and token consistency checks after removing the unused color.

Enhancements:
- Add tests that verify every documented syntax role matches its shipped color and that all syntax constants are represented in the specification.
- Document the `theme::MUTED` alias’s divergence from the muted palette tier until it can be renamed.

Documentation:
- Update the TUI syntax-highlighting and gutter documentation to reflect the implemented palette.
- Remove the retired comment token from brand and design token references.

Tests:
- Add specification coverage for syntax palette values and role completeness.
- Update contrast-ratchet tests and brand parity checks for the reduced token set.

### DIFF
--- a/crates/stella-tui-theme/src/fallback.rs
+++ b/crates/stella-tui-theme/src/fallback.rs
@@ -63,7 +63,7 @@ pub fn ansi16(color: Color) -> Color {
         token::GOLD | token::GOLD_BRIGHT => Color::Yellow,
         token::SILVER | token::SILVER_TYPE => Color::Gray,
         token::TEXT => Color::White,
-        token::MUTED | token::DIM | token::COMMENT => Color::DarkGray,
+        token::MUTED | token::DIM => Color::DarkGray,
         token::GREEN => Color::Green,
         token::RED => Color::Red,
         // The light-theme stops. They reach the terminal only when a paper
--- a/crates/stella-tui-theme/src/token.rs
+++ b/crates/stella-tui-theme/src/token.rs
@@ -122,9 +122,6 @@ pub const MUTED: Color = Color::Rgb(0x77, 0x77, 0x82);
 /// Hints, captions, line numbers. `#4B4B56`
 pub const DIM: Color = Color::Rgb(0x4B, 0x4B, 0x56);
 
-/// Code comments. `#565660`
-pub const COMMENT: Color = Color::Rgb(0x56, 0x56, 0x60);
-
 /// Pass, additive diff sign. `#74C991`
 pub const GREEN: Color = Color::Rgb(0x74, 0xC9, 0x91);
 
@@ -190,7 +187,6 @@ pub const ALL: &[(&str, Color, Clamp)] = &[
     ("text", TEXT, Clamp::NeutralGray),
     ("muted", MUTED, Clamp::NeutralGray),
     ("dim", DIM, Clamp::NeutralGray),
-    ("comment", COMMENT, Clamp::NeutralGray),
     ("green", GREEN, Clamp::Verdict),
     ("red", RED, Clamp::Verdict),
     ("diff-add-bg", DIFF_ADD_BG, Clamp::Surface),
--- a/crates/stella-tui/src/theme.rs
+++ b/crates/stella-tui/src/theme.rs
@@ -192,6 +192,19 @@ pub const GOLD_LIVE: Color = palette::GOLD_LIVE;
 /// Primary text.
 pub const INK: Color = TEXT_PRIMARY;
 /// Dimmed secondary text.
+///
+/// **This is `silver`, not the `muted` tier.** It resolves to
+/// [`palette::TEXT_SECONDARY`], which is `stella_tui_theme::token::SILVER`
+/// `#A9AAB5` — a different colour from `token::MUTED` `#777782`, one tier
+/// brighter, and two tiers from `token::DIM`. The word is shared and the value
+/// is not, so a call site reaching for "the muted tier" and writing
+/// [`muted()`] gets silver.
+///
+/// That is not hypothetical. `crate::diff`'s `gutter` paints [`muted()`], and
+/// `design/tui-v2/SPEC.md` §6.4 recorded the result as *"Line number gutter in
+/// `dim`"* — wrong by two tiers, for as long as both existed. #4946 corrected
+/// the spec against the code; #4966 is where renaming this constant is being
+/// decided, and until it is, the name is checked against nothing.
 pub const MUTED: Color = TEXT_SECONDARY;
 /// Panel border / rule.
 pub const RULE: Color = HAIRLINE;
--- a/crates/stella-tui/tests/spec_syntax_palette.rs
+++ b/crates/stella-tui/tests/spec_syntax_palette.rs
@@ -1,0 +1,152 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+// Copyright (c) 2026 Oxagen, Inc. Commercial licensing: licensing@oxagen.sh
+
+//! **`design/tui-v2/SPEC.md` §6.4 describes the highlighter that ships.**
+//!
+//! §6.4 named the syntax palette in one sentence — *"keywords gold, types
+//! silver_type, identifiers text, comments comment, strings silver, primitives
+//! silver_type"* — and every clause of it was false. Keywords ride the bright
+//! neutral, strings are a green at the scheme's own chroma, numbers are violet,
+//! and the grammar-backed lexer (#4283) added a type and a function position
+//! the sentence predates entirely. Comments were the clause that got noticed
+//! (#4946): the `comment` token the sentence named was painted by nothing
+//! anywhere in the tree, and the highlighter has always used the caption tier.
+//!
+//! A specification that disagrees with the code in six of six roles is worse
+//! than none, because it is read as the answer. So the sentence is now a table
+//! between two markers, and this test is what makes the table true: it reads
+//! §6.4 and holds every row to the constant it names.
+//!
+//! It reads the file as text rather than generating it. Generation would make
+//! the spec a projection of the code, and the point of a spec is that it can
+//! disagree — a row that no longer matches is a decision someone has to make,
+//! which is the conversation this file exists to force rather than to
+//! foreclose.
+
+use std::path::{Path, PathBuf};
+
+use ratatui::style::Color;
+use stella_tui::theme;
+
+const BEGIN: &str = "<!-- BEGIN syntax palette -->";
+const END: &str = "<!-- END syntax palette -->";
+
+fn spec() -> String {
+    let path: PathBuf = Path::new(env!("CARGO_MANIFEST_DIR"))
+        .join("../../design/tui-v2/SPEC.md")
+        .canonicalize()
+        .expect("design/tui-v2/SPEC.md resolves from CARGO_MANIFEST_DIR");
+    std::fs::read_to_string(&path).unwrap_or_else(|e| panic!("cannot read {}: {e}", path.display()))
+}
+
+/// The constants §6.4 is allowed to name, and their shipped values.
+///
+/// Spelled here rather than looked up, because Rust has no reflection over
+/// module constants — and spelling them is what makes the table's *coverage*
+/// checkable: a `SYNTAX_*` constant that grows without a row is caught by
+/// `every_syntax_constant_has_a_row` below, which is the half a value
+/// comparison alone would miss.
+fn shipped() -> Vec<(&'static str, Color)> {
+    vec![
+        ("SYNTAX_KEYWORD", theme::SYNTAX_KEYWORD),
+        ("SYNTAX_STRING", theme::SYNTAX_STRING),
+        ("SYNTAX_NUMBER", theme::SYNTAX_NUMBER),
+        ("SYNTAX_COMMENT", theme::SYNTAX_COMMENT),
+        ("SYNTAX_TYPE", theme::SYNTAX_TYPE),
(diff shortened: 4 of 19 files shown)


macanderson:
Two things noticed while doing this and deliberately left, both filed as handoffs:

- **#4975** — a colour token can ship with nothing painting it, and nothing asks. Retiring `comment` does not stop the next one: `design/tokens/stella-tokens.json` declares a clamp, a role sentence and a surface list per token, but no consumer, so "which tokens does nothing paint?" is unanswerable in this tree. That is invariant #10's shape one plane over from `consumers.rs`.
- **#4976** — `design/tui-v2/SPEC.md` §3.1's palette table is maintained by hand and checked by nothing. This PR had to delete its `comment` row by hand, which is exactly the manual step §6.4's rot came from. The issue also records one clause of §3.5 that is already wrong: it says *silver to white*, and `fallback.rs`'s `ansi16` maps `SILVER` to `Color::Gray` with `TEXT` taking `White`.

Also worth flagging on the diff itself: the second commit repairs `scripts/test-contrast.sh`, whose "a pairing that got better" case repainted `comment`. That suite went **14 passed, 3 failed** the moment the token left the kit — `guard-self-tests.yml` runs it, so this PR would have redded that workflow. It now moves `dim`, the other pairing held to the 3.0 decorative floor, and is back to 17/17.

macanderson:
CI caught one thing the local guards could not, and it is worth recording because I misjudged it.

`website/src/lib/brand-parity.test.ts` asserted `ramp.length >= 21`. I read that as a loose non-vacuity floor and moved on. It was **exactly** the palette's `--st-*` count, so retiring `comment` turned it red on a tree where every mirror was correct — a hardcoded census failing, not a guard finding anything.

The fix is a net tightening rather than a lowered number:

- the floor moves 21 -> 20 with its reason beside it, and
- every `--st-` name the kit sheet declares must now be a token `design/tokens/stella-tokens.json` actually defines. A count could never see a name the palette has no entry for, and the loop underneath would compare such a role against itself and pass.

Witness: renaming `--st-dim` to `--st-dimm` in `docs/brand/css/tokens.css` now fails with `declares --st-* name(s) that design/tokens/stella-tokens.json does not define: --st-dimm`. It passed before this commit.

Measuring that turned up a third thing, filed as **#4978**: the kit sheet carries **20 of the palette's 32**, and the twelve it has never mirrored are the entire warm-paper half plus `void` — `--st-paper-ground`, `--st-paper-raised`, `--st-paper-row`, `--st-paper-seam`, `--st-paper-text`, `--st-ink-muted`, `--st-gold-ink`, `--st-green-ink`, `--st-amber`, `--st-amber-ink`, `--st-red-ink`, `--st-void`. The two documents a designer reads to learn the palette do not contain the light scheme. That is out of scope here (it interacts with #4072's decision about which light scheme is real) but it is why the new assertion checks the subset direction and not equality.

This branch has also been rebased onto current main: #4947 landed four banned values and prose into the same `banned` block of `design/tokens/stella-tokens.json`. Both edits are kept — the $comment now explains that five entries are not superseded-kit anchors, four because they were never a kit (#3653) and one because nothing painted it (#4946).

Composition checked rather than assumed: this branch merged with #4963 and #4972 onto main gives `tokens: 32 validated`, `check-contrast: 21 licensed pairing(s), 3 held`, `check-light-clamp: 34 held`, `check-gate-parity: OK — 48 gate steps`, and 12/12 + 2/2 + 17/17 + 17/17 + 21/21 across the affected suites.

Issue #4968, opened 2026-08-26
A skill mined from Chinese or Japanese text still gets no name


macanderson:
## Problem

`crates/stella-core/src/mining/translit.rs` folds a non-ASCII letter to the ASCII letters that stand for it, so `slugify` gives a mined lesson a readable name instead of the bare artifact kind. Three scripts are covered — Latin-1 Supplement / Latin Extended-A, Cyrillic, and Greek — chosen by one rule: a letter is in the table when it has an unambiguous ASCII stand-in that needs no dictionary.

Han characters, kana, Arabic and Hebrew have no such stand-in, so they are absent, and text written entirely in one of them still slugs to `fallback`:

```rust
assert_eq!(mining::slugify("数据库查询格式化", "skill"), "skill");
```

That is the exact case #4913's title named, and it is the half that is still open. A Chinese or Japanese workspace's `.stella/skills/` still fills with `skill-1a2b3c4d/`, `skill-9f8e7d6c/`, one indistinguishable from the next in `ls`, in the SKILLS tab, and in a `git diff`.

## Why it was left

Romanizing Han needs a per-character dictionary of thousands of entries, and which reading is right depends on the surrounding word — 行 is `xíng` or `háng` in Mandarin, `gyō`/`kō`/`i` in Japanese. Nothing a small in-tree table could return would be an answer; it would be an invention. Kana is mechanical and could be tabled on its own, but a Japanese sentence mixes kana and kanji, so a kana-only table names half a lesson and drops the half that carries the meaning.

So this is a dependency question, and AGENTS.md's "no new dependencies casually" plus `deny.toml`'s license gate have to weigh it.

## Options

1. **A romanization crate** (`pinyin`, `kakasi`-style, or `deunicode` for the whole problem). `deunicode` is one dependency covering every script at once and is the smallest diff; its per-character tables are approximate for CJK by design, which for a directory name may be fine and is the thing to decide. Check the license against `deny.toml` before anything else — if `cargo deny` rejects it, that ends the option (AGENTS.md: drop the dependency, do not widen the allow-list).
[...]

Issue #3298, opened 2026-08-14
A tokenizer that only reads ASCII text means prompts in other languages select no skills and teach no lessons


macanderson:
## The problem

`crates/stella-core/src/mining.rs::terms` treats only **ASCII** alphanumerics (plus `_`) as word characters:

```rust
if ch.is_ascii_alphanumeric() || ch == '_' {
```

So any text with no ASCII words tokenizes to the empty set. Every consumer of that function then silently does nothing useful:

- **Skill selection** (`skills::select_skills`) scores the prompt against each skill's name+description in `terms` space. An empty prompt-term set means a lexical score of 0 for every skill, so a CJK, Cyrillic, Greek, or heavily-accented prompt can only ever select a skill via the domain-tag path.
- That path was just correctly narrowed by #3296 (#3243 D1) — it now requires the prompt to name an existing workspace file. Both effects are individually right, and together they mean **a non-Latin-script prompt that names no path selects no skills at all**.
- **Rule and skill mining** (`already_captured`, `cluster_observations`) compares term sets, so non-ASCII observations never cluster and never dedupe: each recurrence looks novel, and the "observed enough" promotion threshold can never be reached for them.

## Why this was not fixed in passing

The function's own doc comment states the constraint, and it is real:

> Widening this would change every mined `<slug>-<hash8>` id, so it is a deliberate migration, not a drive-by tweak.

`crates/stella-core/src/skills/migration_contract.rs` guards that id space explicitly (it already reasons about what falls "permanently out of clustering reach" in `mining::terms` space). Changing the tokenizer re-tokenizes every stored artifact's provenance, so existing `.stella/skills/<slug>-<hash8>/` files and mined rule ids stop matching what a re-mine would produce — which means duplicate skills appearing beside the originals, not a clean upgrade.

## Sub-problems of different difficulty

1. **Accented Latin, Cyrillic, Greek** — `is_alphanumeric()` instead of `is_ascii_alphanumeric()` handles these correctly, because those scripts are space-separated. Also switch the `current.len() > 2` byte checks to `current.chars().count() > 2`, which is a no-op for ASCII and correct for multi-byte text (today a 1-character CJK "word" is 3 bytes and passes a length gate meant to drop 1-2 character tokens).
[...]

PR #4970, opened 2026-08-26
feat(stella-core): a non-Latin mined lesson gets a name, folded to ASCII

### LINKED ISSUE #4913: A skill mined from non-Latin text gets no name, just skill-<hash>
## Problem

`crates/stella-core/src/mining::slugify` keeps only `char::is_ascii_alphanumeric`; everything else collapses to a dash, and a slug that trims to nothing falls back to the artifact kind (`"skill"` / `"lesson"`).

So a lesson mined from non-Latin text gets no name. `slugify("数据库查询格式化", "skill")` is `"skill"`, and the whole id is `skill-<hash8>` — unique, stable, and unreadable. A Russian workspace's `.stella/skills/` fills with `skill-1a2b3c4d/`, `skill-9f8e7d6c/`, one indistinguishable from the next in `ls`, in the SKILLS tab, and in a `git diff`.

This became reachable rather than theoretical once #3298's clustering half landed: before it, non-Latin observations never clustered, so no non-Latin candidate was ever minted at all. Now they are.

## Why it was not fixed in passing

Widening `slugify` **does** move ids, which widening `mining::terms` turned out not to. The id is `slugify(text)`-`hash8(text)` over the representative wording; change `slugify` and the same text mints a different filename.

Two cases, and they are not equally safe:

- **Fully non-Latin text** — today's id is `skill-<hash8>`. Widening changes the slug half. Any such artifact already on disk keeps its own filename, and the miner's `already_captured` guard works on *text*, not on the filename, so a re-mine is still recognised as a duplicate and no second file appears. Safe.
- **Mixed text** — `slugify("café serveur", …)` is `caf-serveur` today and would become `café-serveur`. Same reasoning applies to duplicates, but the character set changes: a slug is a **directory name**, so this decides whether `.stella/skills/` may hold non-ASCII paths at all. That is a real question on case-insensitive and NFD-normalising filesystems (macOS stores `é` decomposed, so `café-serveur` on disk is not byte-equal to the `café-serveur` a Linux clone writes), and it is why this is a decision rather than a one-character edit.

A third option avoids both: transliterate to ASCII (`café` → `cafe`, Cyrillic → Latin) rather than widen. That keeps every path ASCII, gives a readable name, and is a dependency decision (`deunicode`, `unidecode`) that `deny.toml` and AGENTS.md's "no new dependencies casually" have to weigh.

## Files

- `crates/stella-core/src/mining.rs` — `slugify`, `hash8`.
- `crates/stella-core/src/skills.rs` — `mine_skill_candidates`, where `slugify(&text, "skill")` is called.
- `crates/stella-core/src/rules.rs` — the lesson miner, `slugify(&text, "lesson")`.
- `crates/stella-core/src/skills/migration_contract.rs` — `candidate_identity_is_pinned_to_a_literal` and `rendered_skill_bytes_are_pinned`, which pin the ASCII id. Both must stay green (their fixture text is ASCII, so a correct fix does not touch them) — if either goes red, the change reached further than intended.
- `crates/stella-cli/src/memory/tests/skill_creation.rs`, `crates/stella-cli/src/memory/rules_mining/tests.rs` — the id-shape assertions one layer up.

## Verify

Today:

```rust
assert_eq!(mining::slugify("数据库查询格式化", "skill"), "skill");
assert_eq!(mining::slugify("форматировать запросы", "lesson"), "lesson");
```

End to end: mine three Russian restatements of one lesson through `mine_skill_candidates` (they cluster since #3298) and read the candidate's `name` — it is `skill-<hash8>` with no word in it.

## Definition of done

Pick one of the three shapes above and say which in the code, with the filesystem-normalisation question answered out loud if the answer is "allow non-ASCII paths". A mined non-Latin lesson gets a name a human can read; `candidate_identity_is_pinned_to_a_literal` and `rendered_skill_bytes_are_pinned` still pass unchanged; a witness test covers the new slug.

Noticed while fixing the clustering half of #3298. Refs #3298, #3243.

macanderson:
## Which of the three shapes, and why

#4913 lays out three options and asks for a pick, said in the code. This is the third — **transliterate to ASCII** — implemented in-tree rather than by taking `deunicode`/`unidecode`, so the dependency decision the issue flagged does not have to be made either.

Both other options were rejected for the reason the issue names:

- **Widening the slug alphabet** decides, as a side effect, whether `.stella/skills/` may hold non-ASCII paths. macOS stores `é` decomposed where a Linux clone composes it, and a case-insensitive volume folds pairs an ext4 clone keeps apart. That is a filesystem question, it is bigger than this issue, and **this PR does not answer it** — the output alphabet is exactly where it was, so the question stays open and unforced.
- **Leaving it** loses the readability the issue is about.

The argument lives in `crates/stella-core/src/mining/translit.rs`'s module header, so the choice is reviewable rather than implied.

## What it covers, and what it does not

`translit::ascii_fold` maps one lowercase non-ASCII letter to the ASCII letters that stand for it. Three scripts, chosen by one rule — a letter is in the table when it has an unambiguous ASCII stand-in that needs no dictionary:

- **Latin-1 Supplement and Latin Extended-A** — diacritic folding plus the digraphs (`ß` → `ss`, `æ` → `ae`, `þ` → `th`).
- **Cyrillic** — BGN/PCGN (`щ` → `shch`, `х` → `kh`). The hard and soft signs romanize to nothing, which is why the table returns a `&str` and `Some("")` is a real answer distinct from `None`: `объект` is `obekt`, not `ob-ekt`.
- **Greek** — ISO 843, letter by letter. The digraph rules (`γγ` → `ng`) need context and are not applied; a slug is not a transcription.

**Han, kana, Arabic and Hebrew are absent, and their text still slugs to the artifact kind** — so the Chinese example in #4913's title is *not* fixed here. Romanizing Han needs a per-character dictionary of thousands of entries and the right reading depends on the surrounding word, so anything a small table returned would be invented. That half is filed as **#4968** with the three options and the licensing/filesystem questions each one owes, and `a_script_with_no_single_codepoint_romanization_is_absent` pins today's answer so the header cannot go quietly false.

Uppercase is not in the table: `slugify` lowercases first, and `char::to_lowercase` already knows all three scripts (Greek's final sigma included).

## No migration

An id is minted once, at mint time, and never recomputed against a file already on disk. The miners' duplicate guard — `mining::already_captured` — compares the candidate's **text** against the `name`/`description`/`body` of every existing skill, never a filename, so an artifact minted under the old rule keeps its own directory name and a re-mine of the same text is still recognised as the duplicate it is. (Verified by reading `already_captured` and `mine_skill_candidates`, not taken from the issue: the haystack is `format!("{} {} {}", s.name, s.description, s.body)`.)

What *does* move is the id the same text would mint from here on — `café serveur` becomes `cafe-serveur` rather than `caf-serveur`. Nothing is orphaned by that and no duplicate appears; it changes future names only.

ASCII text mints the byte-identical id it always did. `slugify_is_unchanged_on_ascii` holds that over generated input against a verbatim copy of the old rule, the same pairing #3298 used for `terms`, and `migration_contract::candidate_identity_is_pinned_to_a_literal` / `rendered_skill_bytes_are_pinned` are green **unchanged**.

## Witnesses

- `non_latin_text_slugs_to_a_readable_name` — Russian, Czech, German, Greek, and the hard-sign case.
- `a_slug_is_still_ascii_whatever_the_text_was` — the fold-not-widen property, including the CJK case that is still `skill`.
- `a_non_latin_lesson_mints_a_name_a_human_can_read` (`crates/stella-core/src/skills/tests.rs`) — end to end through `mine_skill_candidates`, which is #4913's own "Verify" recipe.
- `translit`'s three table tests: every folding is ASCII lowercase, no ASCII character is in the table, and the absent scripts stay absent.

Ablated by deleting the fold arm from `slugify` and leaving everything else:

```
---- mining::tests::non_latin_text_slugs_to_a_readable_name stdout ----
  left: "lesson"
 right: "formatirovat-zaprosy"

---- skills::tests::a_non_latin_lesson_mints_a_name_a_human_can_read stdout ----
the id carries the lesson's words, not just a hash: skill-b10b8096

test result: FAILED. 0 passed; 2 failed
```

The answer is wrong under ablation rather than merely constant: `a_slug_is_still_ascii_whatever_the_text_was` and `slugify_is_unchanged_on_ascii` stay green either way, so the suite cannot pass by folding nothing *or* by folding everything into non-ASCII.

Nothing removed or renamed; six added.

## No new dependencies

The table is in-tree, `mining/translit.rs`, ~150 lines. That is the point of picking this shape over `deunicode`.

## Verification

`cargo fmt --all --check`, `cargo clippy -p stella-core --all-targets -- -D warnings`, `make guards-fast` (prose included), `cargo test -p stella-core` (exit 0), and `cargo test -p stella-cli --bins` (2325 passed — the id-shape assertions in `memory/tests/skill_creation.rs` and `memory/rules_mining/tests.rs` among them). CI is the evidence.

Closes #4913
Refs #3298, #4968

## Summary by Sourcery

Make mined lesson identifiers readable for supported non-Latin text without widening the repository’s ASCII path alphabet or changing existing artifact names.

New Features:
- Transliterate supported Latin, Cyrillic, and Greek lesson text into readable ASCII slug names while preserving ASCII-only paths.
- Keep scripts without unambiguous single-character romanizations on the existing artifact fallback behavior.

Bug Fixes:
- Prevent non-Latin mined lessons from losing their readable names and falling back to hash-only identifiers.

Enhancements:
- Document the ASCII-folding strategy, its filesystem rationale, and its compatibility and migration behavior.
- Add unit, property-based, and end-to-end coverage for transliteration, ASCII stability, fallback scripts, and readable mined identifiers.

Tests:
- Add coverage validating transliteration mappings, ASCII-only slug output, unchanged ASCII behavior, unsupported scripts, and end-to-end skill mining.

### DIFF
--- a/crates/stella-core/src/mining.rs
+++ b/crates/stella-core/src/mining.rs
@@ -6,6 +6,8 @@
 //! (These started as two byte-identical private copies; the divergence risk
 //! is why they were merged.)
 
+mod translit;
+
 use std::collections::{HashMap, HashSet};
 
 const STOPWORDS: &[&str] = &[
@@ -250,6 +252,32 @@ pub(crate) fn jaccard(a: &HashSet<String>, b: &HashSet<String>) -> f64 {
 /// Filesystem/id-safe slug: lowercase, alnum + dashes, capped short.
 /// `fallback` names the artifact kind when the text slugs to nothing
 /// (`"lesson"` for rules, `"skill"` for skills) (TS: `slugify`).
+///
+/// # Non-Latin text gets a name, and every path stays ASCII (#4913)
+///
+/// A letter outside ASCII is folded to the ASCII letters that stand for it
+/// ([`translit::ascii_fold`]) before it is dropped as a separator. Without
+/// that, a lesson mined from Russian or Czech text slugged to nothing and the
+/// whole id was `skill-<hash8>` — unique, stable, and unreadable in `ls`, in
+/// the SKILLS tab, and in a `git diff`. #3298's clustering half is what made
+/// that reachable: before it, non-Latin observations never clustered, so no
+/// such candidate was ever minted.
+///
+/// **Folding, not widening.** The output alphabet is unchanged, so this
+/// decides nothing about whether `.stella/skills/` may hold non-ASCII paths —
+/// the filesystem question the module header of [`translit`] lays out, and the
+/// reason widening is not the fix. A script with no single-codepoint
+/// romanization (Han, kana, Arabic, Hebrew) still slugs to `fallback`; #4968
+/// tracks it.
+///
+/// **No id already on disk moves.** An id is minted once and never recomputed
+/// against an existing file: the miners' [`already_captured`] guard compares
+/// *text* against the name/description/body of what is already there, so an
+/// artifact minted under the old alphabet keeps its filename and is still
+/// recognised as the duplicate it is. What moves is the id the *same text*
+/// would mint from here on — `café serveur` becomes `cafe-serveur` rather than
+/// `caf-serveur` — and an ASCII-only text mints the byte-identical id it
+/// always did, which `slugify_is_unchanged_on_ascii` holds.
 pub(crate) fn slugify(text: &str, fallback: &str) -> String {
     let mut collapsed = String::new();
     let mut prev_dash = false;
@@ -257,6 +285,11 @@ pub(crate) fn slugify(text: &str, fallback: &str) -> String {
         if ch.is_ascii_alphanumeric() {
             collapsed.push(ch);
             prev_dash = false;
+        } else if let Some(folded) = translit::ascii_fold(ch) {
+            // May be empty — a Cyrillic hard or soft sign stands for no letter
+            // of its own — and an empty folding is not a separator either.
+            collapsed.push_str(folded);
+            prev_dash = prev_dash && folded.is_empty();
         } else if !prev_dash {
             collapsed.push('-');
             prev_dash = true;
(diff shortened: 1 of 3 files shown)


macanderson:
Answering the one `❌` row so it is not left standing.

> **Ensure mined lessons and skills containing non-Latin text receive a human-readable slug instead of only the artifact kind and hash.** ❌ — the PR improves readability for Latin, Cyrillic, and Greek text through in-tree transliteration, but explicitly leaves Han/CJK, kana, Arabic, and Hebrew unchanged.

Correct, and deliberate. It is not a gap this PR walked past — it is filed, with a handoff.

**Why CJK is not here.** Romanizing Han needs a per-character dictionary of thousands of entries, and which reading is right depends on the surrounding word (行 is `xíng` or `háng` in Mandarin, `gyō`/`kō`/`i` in Japanese). Anything a small in-tree table returned for those would be invented rather than transliterated. Kana alone is mechanical, but a Japanese sentence mixes kana and kanji, so a kana-only table names half a lesson and drops the half carrying the meaning. That makes it a **dependency decision** (`deunicode`, a pinyin crate), which AGENTS.md's "no new dependencies casually" and `deny.toml`'s license gate have to weigh — a different kind of change from the one this PR makes, and one a maintainer picks.

**Where it is tracked:** #4968, written as a handoff — the three options with the licensing and filesystem questions each one owes, the files, the verify recipe, and the definition of done.

**Where it is said in the code**, so the limit is not folk knowledge:

- `crates/stella-core/src/mining/translit.rs`'s module header, under "What is here and what is not", names the absent scripts and why, citing #4968.
- `mining::slugify`'s doc comment repeats the limit at the call site.
- `a_script_with_no_single_codepoint_romanization_is_absent` asserts `ascii_fold` returns `None` for 数, 据, あ, ア, ا, א, 한 — so the header paragraph cannot go quietly false.
- `a_slug_is_still_ascii_whatever_the_text_was` ends with `assert_eq!(slugify("数据库查询格式化", "skill"), "skill")`, which pins today's answer for the issue's own headline example rather than leaving it undescribed.

The other two rows are `✅` and the change stands on its own: a Russian or Czech lesson gets a readable name today, every path stays ASCII, and no id already on disk moves.

Issue #4974, opened 2026-08-26
The Observatory's rules panel cannot see a plugin's context records


macanderson:
## Problem

#4917 closed the skills half of the Observatory/TUI parity gap: a plugin's contributed skills are now listed in the dashboard, named with the package that shipped them, through the `stella_observatory::PluginSkills` port that `stella-cli`'s `storage_cmd::RosterSkills` implements over the one roster the session's own skill load reads.

The same asymmetry is still open for **context records**. A package's `<plugin_dir>/rules/` steers every matching turn in the workspace — `stella context explain` prints `contributed by the \`vera\` plugin` and `ContextRecord::contributed_by` carries the attribution — and the Observatory's rules panel scans exactly `.stella/rules` and nothing else. So a record a third party ships shapes the prompt and the dashboard's rules panel does not know it exists.

That is worse than absent for the same reason the skills gap was: the panel reads as the complete list of what steers this workspace, and a record is the surface that steers *quietly*, without appearing in a transcript as anything the agent did.

## Why it is a port and not a directory scan

Exactly the reasoning #4917 wrote down, one panel over. A package's records are contributed by derivation and never copied into `.stella/rules` — that non-copying is what makes `stella plugin remove` total. Reaching them from `fsview` means resolving the roster there, and with it the project-tier trust gate (#3509), the `plugins.<name> = "off"` retraction, the install receipt and the reconcile-on-every-load check. A second implementation of a security answer is a second answer.

`crates/stella-cli/src/plugin_cmd/package.rs`'s `contributed_record_dirs` is the sibling of `contributed_skill_dirs` and already yields `{plugin, scope, dir}` — so the host side is one more `impl` and the observatory side is one more port, both modelled on what #4917 landed.

## Files

- `crates/stella-observatory/src/fsview.rs` — `rules_files` (its doc comment names this issue), and `PluginSkills` / `ContributedSkills` / `PluginTier` beside it, the shape to copy.
- `crates/stella-cli/src/storage_cmd.rs` — `RosterSkills`, the host-side implementation to copy, and `run_observe` where a second port would be passed.
- `crates/stella-cli/src/plugin_cmd/package.rs` — `contributed_record_dirs`.
- `crates/stella-observatory/src/assets/index.html` — the rules panel, which needs the `via <plugin>` column the skills table now has.
[...]

Issue #3509, opened 2026-08-17
SECURITY: a cloned repo's .stella/plugins/ loads with no trust gate — arbitrary code execution on clone


macanderson:
## Problem

`PluginRoster::load` (`crates/stella-cli/src/plugin_cmd/roster.rs:185-189`) reads the project tier `<workspace>/.stella/plugins/` **unconditionally**, bypassing `settings::project_code_execution_trusted()` — the gate that already stands in front of project-scope hooks and `.stella/mcp.toml` for exactly this reason.

## Repro

`git clone` a repository containing `.stella/plugins/x/plugin.toml`:

```toml
[loop]
participation = "arbiter"
hooks = ["PreToolUse", "Stop"]

[runtime]
argv = ["python3", "${plugin_dir}/main.py"]
```

`PluginRoster::load` reads it with no trust check. `hook_routes()` then emits five dispatchable routes, and the `[runtime]` process is spawned. **There is no consent transaction anywhere on this path.**

The README's own promise is the thing being broken: *"project-scope hooks and credential-routing fields don't load from a cloned repo until you set `STELLA_TRUST_PROJECT=1`. That's Stella refusing to let a repo you just cloned run arbitrary commands on your machine."* A plugin is strictly more powerful than a hook, and it is the one surface that skips the gate.

## Fix

Gate the project tier on `crate::settings::project_code_execution_trusted()` in `PluginRoster::load`, emitting the same one-line refusal notice `load_mcp_plan` does, and honour `filesystem_settings_disabled()`.

The user tier (`~/.stella/plugins/`) is not affected — that is the operator's own machine-scope directory, same as their own hooks.

## Definition of done

Witness: a plugin in an untrusted project tier is **not loaded, not listed, and dispatches no hook**, and the same plugin loads once the workspace is trusted. Fails before the change, because today it loads unconditionally.

Found by the blocking three-lens socket audit (architecture lens), `doc:pipeline-as-plugins-execution` §1. Introduced by a8adddd (A4). Refs #3380.

macanderson:
Triaged **P0**, Impact 2.

[...]

PR #4977, opened 2026-08-26
feat(stella-observatory,stella-cli): the skills view lists a plugin's skills

### LINKED ISSUE #4917: The Observatory's skills tab cannot see a plugin's skills at all
## Problem

#3567 gave a plugin-contributed skill an attribution field and put it on screen — but on one surface only.

- **TUI SKILLS tab**: shows them. `crates/stella-cli/src/skill_manager.rs`'s `enumerate()` appends `contributed_rows(...)` from every installed package, and `crates/stella-tui/src/views/skills.rs` renders `via {plugin}` off `contributed_by`.
- **Observatory (`stella observe`)**: shows none of them. `crates/stella-observatory/src/fsview.rs`'s `skills()` scans exactly two roots — `<workspace>/.stella/skills` and `<user_config>/skills` — and nothing else. A plugin's skills live in `<plugin_dir>/skills` and are contributed by derivation, never copied into either root (that non-copying is the property that makes retraction total, per `plugin_cmd/roster.rs`'s module header, so the fix is *not* to start copying them).

So a user who installs a package for its skills sees them in the deck and cannot find them in the dashboard. Worse than absent: the dashboard's skills tab reads as the complete list, and its "learned skills" count is computed off these same rows.

The same asymmetry holds for context records. `stella context explain` prints `contributed by the \`vera\` plugin`; the Observatory's rule-file view scans directories and has no equivalent.

## Why it is not just "add the directory"

`fsview` derives everything it shows from the filesystem, on purpose — the Observatory is loopback-only and read-only, and `stella-observatory` deliberately does not link `stella-store` or `stella-cli`. Reaching the plugin roster means either:

1. teaching `fsview` to read `.stella/plugins/` and `~/.stella/plugins/` itself, duplicating the roster resolution and its **trust gate** (`project_code_execution_trusted()`, #3509) — get that wrong and the dashboard lists skills from an untrusted clone that the agent would never load; or
2. moving the roster's directory-resolution half somewhere both can depend on. `stella-plugin` is the near-leaf that already does manifest parsing and validation over borrowed text, and it is what `stella-cli`'s `plugin_cmd/package.rs` builds on.

Option 2 is the one that cannot drift. Option 1 recreates the two-authorities-over-one-answer shape this area exists to avoid.

## Files

- `crates/stella-observatory/src/fsview.rs` — `pub fn skills(workspace_root)`, `scan_skills_dir`, and the rule-file view beside it.
- `crates/stella-cli/src/skill_manager.rs` — `enumerate()` / `contributed_rows()`, the behavior to reach parity with.
- `crates/stella-cli/src/plugin_cmd/package.rs` — `contributed_skill_dirs`, `contributed_record_dirs`, `session_roster`, `reconciles_with_disk`.
- `crates/stella-cli/src/plugin_cmd/roster.rs` — the trust gate and the derive-never-copy rule.
- `crates/stella-observatory/src/assets/index.html` — the skills tab's column set, if a `via <plugin>` column is added.

## Verify

1. Install a package shipping a skill into a trusted workspace (the `crates/stella-cli/src/plugin_cmd/tests/contributions.rs` fixtures build one).
2. `stella observe`, open the skills tab — the plugin's skill is absent today.
3. After the fix it is listed, marked with its plugin, and **absent again** in an untrusted clone, matching `an_untrusted_projects_contributed_tool_never_loads` one surface over.

## Definition of done

The Observatory's skills view lists plugin-contributed skills, names the contributing plugin, and honours the same project-tier trust gate the loader does — with a test asserting the untrusted case shows nothing. Records get the same treatment or an explicit note saying why they do not. Nothing is copied into the user's own `.stella/`.

Noticed while verifying #3567. Refs #3567, #3509.

macanderson:
## What

The Observatory's skills tab lists plugin-contributed skills, names the contributing plugin, and honours the same project-tier trust gate the loader does. Nothing is copied into the user's `.stella/`.

## Why the fix is a port and not a wider directory scan

The issue names two candidate shapes and rejects the first. I shipped neither literally, for a reason worth reviewing:

- **Option 1 — teach `fsview` to read `.stella/plugins/` itself.** The issue rejects this and it is right to: it duplicates the roster resolution *and* the trust gate, and getting that wrong lists skills from an untrusted clone the agent would never load.
- **Option 2 — move the roster's directory-resolution half into `stella-plugin`.** I looked at what that actually costs. The directory half moves cleanly, but the *gate* does not: `project_code_execution_trusted()` resolves `STELLA_TRUST_PROJECT`, `STELLA_PROJECT_HOOKS` **and** `run.auto_trust_project` out of the user/org-managed `stella.toml` via `Settings::trusted_scope_auto_trust_project()`. Moving settings-scope resolution into a near-leaf whose only workspace dependency is `stella-protocol` is a second extraction, through the code-execution boundary, and it is bigger than this issue. Without it, option 2 delivers the directories and leaves the observatory to answer the security question on its own — which is option 1 wearing a different hat.
- **What shipped:** the host that already computes the answer hands it over. `stella-observatory` declares a `PluginSkills` port; `stella-cli`'s `storage_cmd::RosterSkills` implements it over `plugin_cmd::package::contributed_skill_dirs` — literally the same call the session's own skill load makes. One authority, no extraction, and it is AGENTS.md invariant 1 rather than an exception to it.

The port is asked **per request**, not once at startup, so a package installed while the dashboard is open shows up on the next refresh — matching what `fsview` already does for everything else it walks.

Stated out loud in the port's doc comment, because it is a real limit: `STELLA_TRUST_PROJECT` is read from the process asking. A `stella observe` launched without it reports the project tier empty while an agent session launched with it loads those skills. That is the same authority answering in two environments, not two authorities, and what the dashboard reports is what a session started *there* would load.

## Row semantics

Contributed rows match `stella_cli::skill_manager::contributed_rows` one surface over:

- `contributed_by` names the plugin; `null` for the user's own (added to every row, so the page never has to tell an absent field from a null one).
- `scope` is the tier the package is installed at — the tier whose state file governs it.
- `learned` is **false**. A package's `SKILL.md` may carry the miner's `origin: auto` marker, but it was mined in its author's workspace; counted as learned it would inflate the "Learned skills" tile with somebody else's loop.
- `evidence_grade` is null — the ledger it joins against is this workspace's own.
- A contributed skill whose name the user already took is dropped, matching the recall loader: it never reaches the prompt, so listing it would show a steer that is not in force.

The page gains a `via` column and a "From plugins" tile.

## Records

Not fixed here, and said in the code: `fsview::rules_files`'s doc comment now names the gap and points at **#4974**, filed as a handoff with the port shape to copy, the fixture to build on, and the one design question the skills half did not have to answer (a rules file is a record *set*, so deduping is by lineage id and the registry's merge rule is the authority). Split rather than doubled so the skills half ships against real tests instead of two panels shipping against none.

## Does this need a `stella-parity` row?

**No.** That matrix's two axes are the community CLI (`stella-cli`) and the embeddable API (`stella-serve`), and its completeness sweeps are driven by `stella_serve::observe::Route::ALL` and the public `Engine` entry points in `stella-core`'s driver/goal modules. Listing skills in a local dashboard is neither an engine capability nor an API route, so a row would have no `engine_entries` to claim and nothing on either end to enforce it. The TUI/Observatory pair is a real parity axis, but it is not the one that crate models.

## Witnesses

Observatory side, against a fake roster (this crate may not link the one that resolves real rosters):

- `a_plugins_skill_is_listed_and_names_the_plugin` — absent before the roster is asked, listed after, naming the plugin, at the package's tier, `learned: false`, `evidence_grade: null`, and the user's own row still naming nobody.
- `a_plugins_skill_never_displaces_one_the_user_wrote_in_the_listing`.

CLI side, against the real one, through `/api/skills` (the route the page fetches, not an inner function only a test would call):

- `the_dashboards_plugin_skills_honour_the_project_trust_gate` — untrusted checkout contributes nothing and `house-style` is absent from the response body; trusted, the same bytes contribute once and the body carries `"contributed_by":"vera"`; removed, it is gone again. It sets `STELLA_HOME` under the env lock, because the dashboard's user-scope root is `stella_home::stella_home()` and `test_user_home` does not move it — without that a `house-style` in the developer's real `~/.stella/skills` would answer for the test.

### Ablation

Dropped the contributed-row loop from `fsview::skills` while keeping the port call, so the ablation removes the answer rather than the compile:

```
---- fsview::tests::a_plugins_skill_is_listed_and_names_the_plugin ----
panicked at crates/stella-observatory/src/fsview.rs:1308:32:
the package's skill is listed: Array [ … ]

```

```
---- the_dashboards_plugin_skills_honour_the_project_trust_gate ----
panicked at crates/stella-cli/src/plugin_cmd/tests/contributions.rs:965:5:
and the skills view lists it: [{"contributed_by":null,…}]

```

The answer is wrong under ablation rather than constant, and the pair is what separates the two claims: the **untrusted** half of the gate witness stays green under ablation (nothing listed is the correct answer there), while the **trusted** half fails. A change that listed everything unconditionally would fail the untrusted half instead. Neither half passes on its own.

Nothing removed or renamed; three added.

## API change

`stella_observatory::serve` takes an `Arc<dyn PluginSkills>`; `respond` is unchanged (it delegates with `NoPluginSkills`) so the ~78 unit tests that drive it do not have to name a roster to ask about executions, and `respond_with` is the new seam. Updated call sites: `stella-cli`'s `run_observe`, the dev harness `examples/serve.rs`, and two test helpers.

## Verification

`cargo fmt --all --check`, `cargo clippy -p stella-observatory -p stella-cli --all-targets -- -D warnings`, `make guards-fast` (prose included), `cargo test -p stella-observatory` (exit 0) and `cargo test -p stella-cli --bins` (2326 passed), rebased on `origin/main` before the final guard run. CI is the evidence.

Closes #4917
Refs #3567, #3509, #4974

## Summary by Sourcery

Show trusted, loader-consistent plugin-contributed skills in the Observatory without copying them into the user's skill directory.

New Features:
- List skills contributed by installed plugins in the Observatory skills view, including their plugin attribution and installation tier.
- Add a plugin-sourced skills integration port so hosts can provide loader-consistent, dynamically refreshed contributions.

Bug Fixes:
- Prevent untrusted or removed plugin skills from appearing in the dashboard, and avoid showing plugin skills that are overridden by user-owned skills.

Enhancements:
- Expose plugin contribution metadata in skills rows with appropriate learned and evidence semantics while preserving existing response helpers through a no-plugin default.
- Add Observatory UI support for plugin attribution and a dedicated From plugins summary tile.

Documentation:
- Document the plugin skills port and defer plugin-contributed rules-file support to the tracked follow-up.

Tests:
- Add Observatory coverage for plugin listing, attribution, tier metadata, and user-skill precedence.
- Add CLI route-level coverage proving plugin skills honor the project trust gate and disappear when the package is removed.

### DIFF
--- a/crates/stella-cli/src/plugin_cmd/tests/contributions.rs
+++ b/crates/stella-cli/src/plugin_cmd/tests/contributions.rs
@@ -1029,3 +1029,89 @@ fn a_contributed_skills_tier_is_read_off_contributed_by_not_off_its_path() {
 
     let _ = std::fs::remove_dir_all(&root);
 }
+
+/// **The dashboard's trust witness** (#4917). The Observatory lists a
+/// package's skills through `storage_cmd::RosterSkills`, and an untrusted
+/// checkout's package contributes nothing there either.
+///
+/// The untrusted half is the point. A dashboard that resolved the roster
+/// itself would have to re-implement the project-tier gate (#3509), and a
+/// second implementation of a security answer is a second answer: get it wrong
+/// and the page lists skills from a cloned repository the agent would never
+/// load, which is worse than the absence it replaced. It reads the one roster
+/// the session's own skill load reads instead, so the two cannot disagree —
+/// including here.
+///
+/// The sibling one surface over is
+/// `an_untrusted_projects_contributed_tool_never_loads`.
+#[test]
+fn the_dashboards_plugin_skills_honour_the_project_trust_gate() {
+    use stella_observatory::PluginSkills as _;
+
+    /// The dashboard's `/api/skills` body, through the route the page fetches
+    /// rather than through an inner function only this test would call.
+    fn skills_json(root: &Path, source: &crate::storage_cmd::RosterSkills) -> String {
+        let response = stella_observatory::respond_with(root, "/api/skills", source);
+        String::from_utf8_lossy(&response.body).into_owned()
+    }
+
+    let _env = crate::test_env::lock();
+    let _restore = crate::test_env::EnvRestore::capture(&[
+        "STELLA_TRUST_PROJECT",
+        "STELLA_PROJECT_HOOKS",
+        "STELLA_HOME",
+    ]);
+    let root = temp_root("package-observatory-trust");
+    let _paths = crate::paths::test_user_home(root.join("home"));
+    // The dashboard's own user-scope root is `stella_home::stella_home()`,
+    // which `test_user_home` does not move — without this the assertions read
+    // the developer's real `~/.stella/skills` and a `house-style` sitting
+    // there would answer for them.
+    // SAFETY: the env lock is held for the whole mutate-read-restore window.
+    unsafe { std::env::set_var("STELLA_HOME", root.join("home")) };
+
+    let planted = stella_home::resolve_project_plugins_dir(&root).join("vera");
+    plant(&stella_home::resolve_project_plugins_dir(&root), "vera");
+    ships_a_package(&planted, "vera");
+
+    let source = crate::storage_cmd::RosterSkills;
+
+    // SAFETY: the env lock is held for the whole mutate-read-restore window.
+    unsafe {
+        std::env::remove_var("STELLA_TRUST_PROJECT");
+        std::env::remove_var("STELLA_PROJECT_HOOKS");
+    }
+    assert!(
+        source.contributed_skill_dirs(&root).is_empty(),
+        "an untrusted checkout contributes no skills to the dashboard"
+    );
+    let listed = skills_json(&root, &source);
+    assert!(
+        !listed.contains("house-style"),
+        "and nothing of the package's reaches the skills view: {listed}"
+    );
+
+    // SAFETY: as above.
+    unsafe { std::env::set_var("STELLA_TRUST_PROJECT", "1") };
+    let dirs = source.contributed_skill_dirs(&root);
+    assert_eq!(dirs.len(), 1, "the same bytes contribute once trusted");
+    assert_eq!(dirs[0].plugin, "vera");
+    assert_eq!(dirs[0].tier, stella_observatory::PluginTier::Project);
+    let listed = skills_json(&root, &source);
+    assert!(
+        listed.contains("house-style"),
+        "and the skills view lists it: {listed}"
+    );
+    assert!(
+        listed.contains("\"contributed_by\":\"vera\""),
+        "naming the package that shipped it: {listed}"
+    );
+
+    remove(&root, "vera").expect("remove must succeed");
+    assert!(
+        source.contributed_skill_dirs(&root).is_empty(),
+        "and it leaves with the package — nothing was ever copied into .stella/skills"
+    );
+
+    let _ = std::fs::remove_dir_all(&root);
+}
--- a/crates/stella-cli/src/storage_cmd.rs
+++ b/crates/stella-cli/src/storage_cmd.rs
@@ -98,6 +98,49 @@ pub(crate) fn preflight_observatory_stores(
     })
 }
 
+/// The dashboard's answer for "which plugin-contributed skills would a session
+/// started in this workspace load?", read off the one roster the session's own
+/// skill load reads (#4917).
+///
+/// A plugin's skills are contributed by derivation from `<plugin_dir>/skills`
+/// and are never copied into `.stella/skills`, so the observatory — which
+/// derives what it shows from the filesystem and may not link this crate —
+/// could not see them at all. It resolves them here instead of there because
+/// reaching them means answering the project-tier trust gate (#3509), the
+/// `plugins.<name> = "off"` retraction, the install receipt and the
+/// reconcile-on-every-load check, and a second implementation of those is a
+/// second answer.
+///
+/// [`package::contributed_skill_dirs`](crate::plugin_cmd::package) is that one
+/// answer, called per request rather than once at startup so a package
+/// installed while the dashboard is open appears on the next refresh — and
+/// answering an empty roster in a workspace nobody has trusted, which is the
+/// property `the_dashboards_plugin_skills_honour_the_project_trust_gate` pins.
+pub(crate) struct RosterSkills;
+
+impl stella_observatory::PluginSkills for RosterSkills {
+    fn contributed_skill_dirs(
+        &self,
+        workspace_root: &std::path::Path,
+    ) -> Vec<stella_observatory::ContributedSkills> {
+        crate::plugin_cmd::package::contributed_skill_dirs(workspace_root)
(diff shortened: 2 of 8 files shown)


macanderson:
One thing this PR noticed and worked around rather than fixed, so it is not buried in a comment: `crate::paths::test_user_home` amends a thread-local inside `stella-cli` and sets no environment variable, so it does not reach `stella_home::stella_home()` — which is what every other crate resolves through. `the_dashboards_plugin_skills_honour_the_project_trust_gate` drives the Observatory, whose `fsview::user_config_dir` is exactly that call, and under `test_user_home` alone the `/api/skills` body carried ~45 rows out of the developer's real `~/.stella/skills`. The test sets `STELLA_HOME` under the env lock with a comment saying why; that is a workaround. Filed as #4980 with the options, the witness that fails today, and the line here that comes out when it lands.

macanderson:
Heads up — two things that touch this branch, from picking up #4974 and #4980.

**#5003 is stacked on this branch.** It is the records half (#4974), based on `fix/4917-observatory-plugin-skills` rather than on `main`, so its diff is one commit and it needs this to land first. It does edit the port this PR introduced, and the rename is worth reviewing here rather than only there: `PluginSkills` becomes `PluginContributions` with a second method (`contributed_record_dirs`), `NoPluginSkills` becomes `NoContributions`, and `ContributedSkills` becomes `ContributedDir` — one shape for both kinds, matching what `stella_cli::plugin_cmd::package` already calls it. The argument is in #5003's description: one roster answers both questions, and a host that resolved it for skills and not for records would put the two panels on different trust decisions. A `ContributedRecords` beside `ContributedSkills` would also have been two identical structs in a public API.

**This branch now conflicts with `main`.** `adcf5bdb8` (#4965) appended `a_contributed_skill_is_routed_by_its_field_not_its_path` to `crates/stella-cli/src/plugin_cmd/tests/contributions.rs`, at the same place this branch appends `the_dashboards_plugin_skills_honour_the_project_trust_gate`. The resolution is keep-both, and I verified it: rebasing this commit plus #5003's onto today's `main` and resolving that one hunk gives `cargo test -p stella-cli --bins` 2329 passed and `cargo test -p stella-observatory` exit 0.

**The `STELLA_HOME` workaround in `the_dashboards_plugin_skills_honour_the_project_trust_gate` has its fix.** #4980 is the seam bug, and #4992 fixes it: `paths::test_user_home` now takes the env lock itself and moves `HOME` alongside the thread-local, so the two `set_var` lines here become dead weight. Neither PR can remove them alone — the workaround is on this branch and the seam is on one off `main` — so #5002 tracks the deletion once both are in, with the contamination check to run against it rather than only a green test.

Issue #3932, opened 2026-08-19
Make 'main is not known-broken' a required check, so the red-main hold actually blocks merges


macanderson:
## Problem

#3917's mechanism has landed as a workflow: `main-red-hold.yml` runs on every
pull request, asks the tracker whether a `main-red` issue is open, and fails —
naming it — when one is. What it cannot do from inside this repository is
**stop a merge**. That takes one repository setting, which no file in the tree
can carry.

Until the setting changes, the check reports. On 2026-08-19 reporting was
exactly what already existed and was not enough: `main-canary` filed #3904 at
16:57:01 and #3898 merged at 16:57:13, twelve seconds later, with three more
behind it (#3917 has the full timeline).

## What "done" looks like

Add **`main is not known-broken`** to `main`'s required status checks in
branch protection.

That is the whole change. Verify it the way the guard's own test verifies
itself — by watching it block:

1. Open a throwaway PR against `main`.
2. Hand-file an issue labelled `main-red` (any title; the guard matches on the
   label alone).
3. The PR's `main is not known-broken` check goes red, and the merge button is
   blocked rather than merely discouraged.
4. Add the `unblocks-main` label to the PR; re-run; it goes green and the
   merge is available again.
5. Close the hand-filed issue and drop the throwaway PR.

Step 4 is the one worth actually performing rather than assuming — a hold that
also blocks its own repair is a deadlock, and the escape hatch is the only
thing standing between this guard and that outcome.

## Why it is safe to require

Both properties are covered by `scripts/test-main-red-hold.sh` (10 hermetic
checks, `make main-red-hold-test`):

- **It fails open.** An unreachable tracker, or a missing `gh`, reports loudly
  and passes. A GitHub API blip cannot halt every merge in the repository.
- **The repair always has a way through.** `unblocks-main` on the PR passes the
  check. The match is exact, so `unblocks-main-later` is not a bypass.

## Constraints

- `enforce_admins` was off and auto-merge was on (AGENTS.md § the gate) when this
  issue was filed, so an admin merge could still land past this. Update,
  2026-09-02: `enforce_admins` is now on (#3887's resolution), so that gap is
  closed too — this issue's own scope was always just the one toggle below.
- The `unblocks-main` label must exist for a contributor to apply it. Create it
  (any colour) at the same time, or the first person who needs the escape
  hatch discovers it is not there while `main` is broken — the worst possible
  moment.
- Do not solve this by making the guard fail closed, or by widening what it
[...]

PR #4995, opened 2026-08-26
feat(scripts): catch an edit undone inside a branch before a rebase replays it

### LINKED ISSUE #4979: Rebasing a stale branch can silently revert an upstream change, and no pre-merge check can see it
## Problem

A defect whose truth value depends on **which subset of open pull requests is
applied** cannot be caught by any check that runs on one pull request. Every gate
here — `make gate`, the pre-push hook, every required check — asks "is *this
branch* correct?". Nothing asks "is the result of combining these correct?" until
`main-canary.yml`, which by design runs after the merge.

AGENTS.md names this for two files (`Cargo.lock`,
`scripts/file-size-baseline.txt`) and calls it a shared cell. That framing is
narrower than the class. Burning down #4600 across nine pull requests produced
**three distinct shapes** in one day, and the sharpest one is not a shared cell,
is not written down anywhere, and is caused by the ordinary act of rebasing a
stale branch.

This is a handoff, not a fix.

## Shape A — rebase converts an inert edit-and-revert into a live revert

**This is the one worth reading.** It is not a merge hazard; a merge is safe. It
is a *rebase* hazard, and this repository rebases constantly because branches go
stale behind required checks.

Setup: two branches make the **same** edit to one file — which a mechanical sweep
does by construction. Branch B then reverts its copy, so B's final tree matches
base and `git diff base -- <file>` is empty. B looks like it never touched the
file.

Once A lands, rebasing B onto the new base **deduplicates B's edit but keeps B's
revert**:

```
warning: skipped previously applied commit 541a41f
```

Git recognises B's edit as already upstream and drops it. It has no reason to
drop the revert. The inert pair becomes a bare revert of A's change, and B's diff
against the new base — previously empty for that file — is now:

```
-beta
+alpha
```

which squash-merges cleanly and silently undoes A.

**Verified reproduction**, from an empty repository:

```sh
mkdir /tmp/replay && cd /tmp/replay && git init -q -b main .
printf 'alpha\n' > f.txt && git add f.txt && git commit -qm base

git checkout -qb a && printf 'beta\n' > f.txt && git commit -qam "A: edit"

git checkout -q main && git checkout -qb b
printf 'beta\n' > f.txt && git commit -qam "B: same edit"
printf 'alpha\n' > f.txt && git commit -qam "B: revert it"
git diff main b -- f.txt          # EMPTY -- B appears not to touch the file

git checkout -q main && git merge -q --no-edit a
cat f.txt                          # beta

git checkout -q b && git rebase main
cat f.txt                          # alpha  <-- A's edit is gone
git diff main HEAD -- f.txt        # -beta +alpha
```

Merging B into main *without* rebasing keeps `beta`, because B contributes no
change to that file relative to the merge base. Only the rebase path reverts.

**The real instance.** #4954 edited
`crates/stella-observatory/src/accept.rs` and reverted it in a later commit,
because both copies of that file had to travel in #4951 instead. Rebasing #4954
onto a tree carrying #4951 replayed the revert, and `make prose` on the composed
tree said:

```

  crates/stella-observatory/src/accept.rs [filler-adverb]: 1 allowed, 4 found
```

Both branches were green on their own.

**The remedy is to flatten**, so the file is absent from the branch's diff rather
than present as an edit and an undo. #4954 is one commit for exactly this reason.

## Shape B — a partition that splits something atomic

Each branch fails its own checks; the combination would have been fine. Per-pull-
request CI caught this one, which is why it is recorded separately: the gates
worked and the *plan* was wrong.

`crates/stella-serve/src/accept.rs` and
`crates/stella-observatory/src/accept.rs` are byte-identical duplicates, and
`the_two_copies_of_this_policy_have_not_drifted` runs in **both** crates,
comparing them via `include_str!`. A per-crate partition of a repo-wide sweep put
one copy in #4951 and the other in #4954, and each was red against a `main`
holding the other unedited:

```
thread 'accept::tests::the_two_copies_of_this_policy_have_not_drifted' panicked
at crates/stella-observatory/src/accept.rs:222:9:
the accept() policy has drifted between stella-observatory and stella-serve.
```

Any partition — by crate, by directory, by owner — has to ask whether it is
splitting something that must change atomically.

`accept.rs` is the only such twin today: `rg -l 'INTENTIONALLY DUPLICATED'`
returns those two files and nothing else, and no other cross-crate `include_str!`
equality assertion exists. The `include_str!` calls in `stella-parity` and
`stella-cli` look for *named symbols* in another crate's source, so they break on
a rename rather than on prose — a fourth shape nobody has hit yet.

## Shape C — conflict, the safe failure

Both branches write overlapping regions of one file and git refuses the merge.
Seven of the thirty-six pairs among the nine prose branches conflict on
`scripts/prose-baseline.txt`. No two shared a single `(file, pattern)` entry —
the file is sorted by path and each branch edited only its own scope — so the
conflicts are positional adjacency, not disagreement about a number.

**This is not a hazard.** It is the outcome the other two shapes should have: it
stops the merge, it is visible on the pull request, and it costs one rebase.
Named here so it is not mistaken for shape A.

## What A and B have in common

The defect exists only in the relationship between branches, so no amount of
per-branch checking finds it. `main-canary.yml` is the only thing that asks a
composition question, and it asks after the merge — #4804 is about making that
answer visible, claimed and un-merge-over-able once it arrives.

Nothing asks before.

## What would have to exist

Pick one; do not implement all.

1. **A guard for shape A.** Fail a pull request whose history contains a commit
   reverting an earlier commit *of the same branch* on a file the final diff no
   longer mentions. Narrow, mechanical, purely local — it needs no other branch —
   and it names its own remedy: flatten. Cheapest, and it catches the shape that
   nothing else can see. **Recommended.**
2. **A periodic composer.** A scheduled job that merges every open pull request
   onto `main`, runs `make gate`, and reports which pair first goes red. Copy
   `scripts/main-canary.sh`, including its discipline of closing the issue again
   when the tree recovers. Does not block, needs no repository setting.
3. **A pre-merge composition check** — a merge queue. The off-the-shelf answer,
   and it catches A directly, but it is a repository setting plus a workflow, the
   same constraint #3932 and #3887 ran into.
4. **Write shapes A and B where a fresh agent reads them.** AGENTS.md documents
   the shared-cell case under `Cargo.lock` and § "God files"; neither the rebase
   replay nor the twin partition appears anywhere. Costs nothing and is worth
   doing regardless of which of 1–3 lands.

## Definition of done

- One of 1–3 exists and has a self-test proving it can fail, matching the
  discipline `guard-self-tests.yml` already applies.
- Shapes A and B are written where a fresh agent reads them, with the flatten
  remedy for A and the "does this partition split something atomic?" question
  for B.
- This issue records which option was chosen and why the others were not.

## Context

Found burning down #4600 across #4926, #4932, #4939, #4944, #4951, #4953,
#4954, #4955, #4956. Related: #4804 (a red `main` is visible, claimed and
un-merge-over-able — the *after* half of this), #3932, #3887, #4940.

macanderson:
Closes #4979

## Which option, and why not the others

**Option 1** — the guard on the pull request itself. It is the only one that
runs before a merge without a repository setting, it needs no other branch, and
it names its own remedy.

- **Option 2 (a periodic composer)** reports after the fact and would not have
  stopped #4954's rebase from replaying.
- **Option 3 (a merge queue)** catches it directly and is a repository setting
  plus a workflow — the constraint #3932 and #3887 both ran into. Still the
  right long-term answer; nothing here forecloses it.
- **Option 4 (write the shapes down)** is done in this PR as well, because the
  issue asks for it regardless of which of 1–3 lands.

## The guard

`scripts/check-rebase-replay.sh`. The key implementation fact from the issue is
that **the branch's diff for the path is empty in both cases**, so nothing that
reads a diff can distinguish them. This reads the branch's *history*: a path
some commit on the branch touched, which the branch's own final diff no longer
mentions.

Base resolution mirrors `check-deleted-tests.sh` — an explicit argument, else
`HEAD^1`/`HEAD^2` on a `refs/pull/N/merge` checkout, else `origin/main` — and
the range is taken from the **merge base**, never the base tip, so a commit that
landed upstream after the branch started is not attributed here.

**No acknowledgement**, unlike `check-deleted-tests.sh`. That guard lets a
deletion pass once it is named in the description, because a deleted test is
sometimes right. This one has no such escape and does not need one: the branch's
tree is byte-identical whether the pair is flattened or not, so flattening loses
nothing and there is no version of this finding where keeping the pair is
better. An acknowledgement would only offer the wrong remedy in
reviewable-looking clothes.

## Where it runs

Its own workflow, `rebase-replay.yml`, on `pull_request` at `fetch-depth: 0`.
`ci.yml`'s guards job checks out at `fetch-depth: 2`, which is exactly right for
`check-deleted-tests.sh` beside it (a two-**tree** question) and useless here —
this reads every commit on the branch. Widening that checkout would make every
other guard in that job pay for a full clone it has no use for, which is
`wire-schema.yml`'s reasoning. No path filter: any diff can carry this shape,
and the check is a few `git log` invocations with no toolchain behind it.

`make rebase-replay` runs it by hand; `make rebase-replay-test` is the suite.
Neither is a `GATE_STEPS` member — `make gate` judges a working tree, and this
asks about a range of commits.

## It fires on the real instance

Run over the head of each of the **last 40 merged pull requests**:

```
green=39 flagged=1 skipped=0

=== #4939
check-rebase-replay: FAIL — 20 path(s) edited and then undone inside this branch.

  docs/tools/bash.toml
      3a5a6c481  docs(tools): leave the generated tool reference to its generator
      3f7765a90  docs(prose): plain language across the design documents
  ...
```

#4939 is one of the nine prose branches #4979 was found on. Confirmed a true
positive rather than taken on trust:

```
$ git diff --stat origin/main pr4939 -- docs/tools/bash.toml
(empty — the branch's diff does not mention the path)

$ git show --stat 3f7765a90 -- docs/tools/bash.toml
 docs/tools/bash.toml | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)

$ git show --stat 3a5a6c481 -- docs/tools/bash.toml
 docs/tools/bash.toml | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)
```

Every **open** pull request is green under it, including this one:

```
test/diagnostic-codes-self-test-4948   OK — 5 path(s) touched, every one still in the branch's diff.
fix/4980-test-user-home-isolation      OK — 8 path(s) touched, …
fix/4976-spec-palette-parity           OK — 2 path(s) touched, …
bot/version-sync                       OK — 3 path(s) touched, …
ci/clippy-schema-gated-4949            OK — 6 path(s) touched, …
fix/4966-theme-muted-tier              OK — 19 path(s) touched, …
fix/embedder-seal-guard-4986           OK — 2 path(s) touched, …
fix/4917-observatory-plugin-skills     OK — 8 path(s) touched, …
```

**One thing I could not show:** #4954's own branch. Its merged head is a single
commit — it was flattened for exactly this reason, which the issue says — so the
guard is green on it. The pre-flatten history is not recoverable from the remote.

## The self-test

`scripts/test-rebase-replay.sh`, eleven cases over throwaway repositories.

```
$ make rebase-replay-test
ok   R0 a branch that edits a file and keeps the edit passes
ok   R1 an edit undone later in the same branch is flagged
ok   R1b the report names the flatten remedy
ok   R2a B's diff against main does not mention the file at all
ok   R2b rebasing B onto A's landing reverts A (skipped the edit, kept the undo)
ok   R3 a branch that never touches the file passes
ok   R4 a file created and deleted inside the branch is flagged
ok   R5 flattening the pair clears the finding
ok   R6 a revert the final diff still mentions is not flagged
ok   R7 a branch with no commits of its own passes
ok   R8 an absent base ref is a failure naming the shallow checkout

passed 11, failed 0
```

**R2 is #4979's from-empty-repository reproduction, run to its conclusion.** It
asserts nothing about the guard — it asserts the premise the guard rests on, so
a future reader does not have to take the issue's word for it. R5 matters for
the opposite reason: if flattening did not clear the finding, the guard would be
unsatisfiable and the remedy it prints would be a lie. R6 keeps it narrow — a
revert the final diff still mentions is not this defect and is not flagged.

### Ablation 1 — the guard cannot fail (`suspects=""`)

The green cases stay green and every failure direction goes red, so the ablation
makes the answer wrong rather than merely constant:

```
ok   R0 a branch that edits a file and keeps the edit passes
FAIL R1 an edit undone later in the same branch is flagged — the guard passed something it should have flagged:
check-rebase-replay: OK — 1 path(s) touched, every one of them still in the branch's diff.
FAIL R1b the report names the flatten remedy — the guard passed something it should have flagged:
FAIL R4 a file created and deleted inside the branch is flagged — the guard passed something it should have flagged:
ok   R2a / R2b / R3 / R5 / R6 / R7 / R8

passed 8, failed 3
```

R2a and R2b stay green under the ablation, which is correct: they are about git,
not about the guard.

### Ablation 2 — the suite is unwired

```
$ make gate-parity
check-gate-parity: FAIL — no workflow runs the guard self-test 'scripts/test-rebase-replay.sh'.
check-gate-parity:      Add it to a workflow, or name it in UNHOSTED_SELF_TESTS in
check-gate-parity:      the Makefile with the reason it is excluded.
```

## Option 4 — the shapes, written down

AGENTS.md's workflow chain gains `rebase-replay.yml` as its ninth entry,
carrying shape A with the flatten remedy, and shape B — "a partition can also
split something atomic", with the `accept.rs` twin and the
`rg -l 'INTENTIONALLY DUPLICATED'` check to run before partitioning by crate,
directory or owner.

## Gates

```
$ make guards-fast
EXIT=0

$ shellcheck scripts/check-rebase-replay.sh scripts/test-rebase-replay.sh
(silent)

$ make gate-parity
check-gate-parity: OK — 48 (forty-eight) gate steps, named in AGENTS.md and CONTRIBUTING.md, each run by a workflow.
```

`GATE_STEPS` is untouched, deliberately: it is a shared cell, and another open
pull request (#4989) is already writing it.

## Deleted tests

None. No `#[test]` was added, removed or renamed — this PR touches no Rust.

## Summary by Sourcery

Add history-aware CI protection against edit-and-undo commit pairs that can become unintended reverts when stale branches are rebased.

New Features:
- Add a pull-request guard that detects files edited and subsequently undone within a branch before rebasing can replay the undo as an upstream revert.

Bug Fixes:
- Prevent silent rebase-induced reversions of changes that have already landed upstream.

Enhancements:
- Document rebase-replay and repository-wide atomicity hazards, including the flattening remedy and duplicated-file guidance.

Build:
- Add Makefile targets for running the rebase-replay guard and its hermetic test suite.

CI:
- Run the new guard in a dedicated full-history pull-request workflow and wire its self-test into guard self-test CI.

Documentation:
- Expand AGENTS.md with the rebase-replay workflow and guidance for identifying changes that must remain atomic.

Tests:
- Add an eleven-case shell test suite covering detection, remediation, rebase reproduction, edge cases, and shallow-checkout failures.

### DIFF
--- a/.github/workflows/guard-self-tests.yml
+++ b/.github/workflows/guard-self-tests.yml
@@ -218,6 +218,10 @@ jobs:
         if: ${{ !cancelled() }}
         run: ./scripts/test-releases-published.sh
 
+      - name: rebase-replay guard directions, and the hazard itself
+        if: ${{ !cancelled() }}
+        run: ./scripts/test-rebase-replay.sh
+
       - name: the shellcheck step's presence guard
         if: ${{ !cancelled() }}
         run: ./scripts/test-shellcheck-guard.sh
--- a/.github/workflows/rebase-replay.yml
+++ b/.github/workflows/rebase-replay.yml
@@ -1,0 +1,51 @@
+name: rebase-replay
+
+# The one question about a branch's HISTORY rather than its tree (#4979).
+#
+# A branch that edits a file and later undoes the edit has an empty diff for
+# that path, so it reads as never having touched it. Rebase it onto a base that
+# already carries the same edit — which a mechanical sweep split across several
+# pull requests produces by construction — and git drops the branch's copy
+# ("skipped previously applied commit") while keeping the undo. The inert pair
+# becomes a live revert of whatever landed upstream, and it merges cleanly.
+#
+# Its own workflow rather than a step in ci.yml, for the reason wire-schema.yml
+# has one: ci.yml's guards job checks out at `fetch-depth: 2`, which is exactly
+# right for check-deleted-tests.sh (a two-TREE question) and useless here — this
+# reads every commit on the branch. Widening that checkout would make every
+# other guard in that job pay for a full clone it has no use for.
+#
+# `pull_request` only, and no path filter: any diff can carry this shape, and
+# the guard is a few `git log` invocations over a range with no toolchain
+# behind it.
+on:
+  pull_request:
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+concurrency:
+  group: rebase-replay-${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  check:
+    name: no edit is undone inside the branch
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      # Full depth: the guard reads the branch's commits, and a shallow clone
+      # makes it report a missing base rather than a clean branch — loudly, but
+      # it still has not answered the question.
+      - id: checkout
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
+        with:
+          fetch-depth: 0
+
+      # On a `pull_request` event the checkout is `refs/pull/N/merge`, so HEAD
+      # is the merge commit: the guard reads HEAD^1 and HEAD^2 itself and needs
+      # no ref name from the event payload.
+      - name: no path is edited and then undone inside this branch
+        if: ${{ !cancelled() && steps.checkout.outcome == 'success' }}
+        run: ./scripts/check-rebase-replay.sh
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -272,6 +272,44 @@ existed. `reserved-paths` (`scripts/check-reserved-paths.sh`) is the fast
 half of that finding, because a failed checkout reports one bad path per run
 and there were two.
 
+A ninth, `rebase-replay.yml`, is the only check here that reads a branch's
+**history** rather than its tree, and it exists for a composition hazard the
+shared-cell paragraph above does not cover: **rebasing a stale branch can
+silently revert somebody else's change** (#4979).
+
+Two branches make the same edit to one file — which a mechanical sweep split
+across several pull requests does by construction. Branch B then undoes its
+copy, so B's tree matches base and `git diff base -- <file>` is empty: B reads
+as never having touched the file. Once A lands, rebasing B onto the new base
+deduplicates B's edit (`warning: skipped previously applied commit`) and keeps
+B's undo. The inert pair becomes a bare revert of A's change, B's diff for that
+path goes from empty to a live revert, and it squash-merges cleanly. Merging B
+*without* rebasing is safe, which is what makes this a rebase hazard rather
+than a merge one — and this repository rebases constantly, because branches go
+stale behind required checks. It happened to #4954 against #4951, and
+`check-rebase-replay.sh` fires on #4939's merged head today.
+
+**The remedy is to flatten**, so the path is absent from the branch's history
+rather than present as an edit and an undo. The branch's tree is identical
+either way, so flattening costs nothing and the guard offers no
+acknowledgement. `make rebase-replay` asks by hand;
+`make rebase-replay-test` covers it, and reproduces the rebase itself so the
+hazard is demonstrated rather than asserted. Its own workflow because ci.yml's
+guards job checks out at `fetch-depth: 2` — right for `check-deleted-tests.sh`,
+which compares two trees, and useless for a question about every commit on a
+branch.
+
+**A partition can also split something atomic**, which per-pull-request CI does
+catch — each branch fails on its own — but only after the plan is already
+wrong. `crates/stella-serve/src/accept.rs` and
+`crates/stella-observatory/src/accept.rs` are byte-identical duplicates whose
+`the_two_copies_of_this_policy_have_not_drifted` compares them with
+`include_str!` from **both** crates, so a per-crate split of a repo-wide sweep
+put one copy in #4951 and the other in #4954 and reddened both. Before
+partitioning by crate, directory or owner, ask whether anything in the set has
+to change atomically. `rg -l 'INTENTIONALLY DUPLICATED'` names today's only
+such twin.
+
 **Cite a document by its id, not its path.** Every document under `docs/` that
 anything cites carries frontmatter with a stable `id`, and a citation names that
 id — `doc:context-reuse §4`. Moving the file cannot break it. A document with no
--- a/Makefile
+++ b/Makefile
@@ -852,6 +852,17 @@ website-inputs-test: ## Test the website-inputs guard's three failure directions
 diagnostic-codes-test: ## Test the diagnostic-code registry's failure directions (#4948; hermetic; not part of `gate`)
 	./scripts/test-diagnostic-codes.sh
 
+# Not a gate step: `make gate` judges a working tree, and this asks about a
+# range of commits — a question a pull request has and a dirty checkout does
+# not. It runs in .github/workflows/rebase-replay.yml on every pull request.
+.PHONY: rebase-replay
+rebase-replay: ## Assert no path is edited and then undone inside this branch (#4979; against origin/main)
+	./scripts/check-rebase-replay.sh
+
+.PHONY: rebase-replay-test
+rebase-replay-test: ## Test the rebase-replay guard, and the rebase hazard it is about (#4979; hermetic; not part of `gate`)
+	./scripts/test-rebase-replay.sh
+
 .PHONY: ci-rust-scope-test
 ci-rust-scope-test: ## Test which diffs run the Rust gate, prose skips and fail-open included (hermetic; not part of `gate`)
 	./scripts/test-ci-rust-scope.sh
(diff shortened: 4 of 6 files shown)


macanderson:
Rebased onto `main` after #4993 landed. The only conflict was in the `Makefile`, where both branches append a hermetic self-test target after `website-inputs-test` — #4979's shape C, the safe failure: git refused the merge, it was visible on the pull request, and it cost one rebase. Resolved by keeping both targets. Re-verified on the rebased tree:

```
$ make guards-fast
EXIT=0

$ make rebase-replay-test
passed 11, failed 0

$ make diagnostic-codes-test          # #4993's, still green beside it
passed 12, failed 0

$ ./scripts/check-rebase-replay.sh
check-rebase-replay: OK — 6 path(s) touched, every one of them still in the branch's diff.
```

Issue #5002, opened 2026-08-26
Remove the STELLA_HOME workaround from the two dashboard trust tests


macanderson:
## Problem

Two tests in `crates/stella-cli/src/plugin_cmd/tests/contributions.rs` set `STELLA_HOME` by hand, right after calling `crate::paths::test_user_home(root.join("home"))`, with a comment saying why:

- `the_dashboards_plugin_skills_honour_the_project_trust_gate` (added by #4977)
- `the_dashboards_plugin_records_honour_the_project_trust_gate` (added by #4974's PR, stacked on it)

Both look like this:

```rust
let _paths = crate::paths::test_user_home(root.join("home"));
// The user-tier half of the roster is `stella_home::stella_home()`, which
// `test_user_home` does not move — without this the roster resolves the
// developer's real `~/.stella/plugins` (#4980).
// SAFETY: the env lock is held for the whole mutate-read-restore window.
unsafe { std::env::set_var("STELLA_HOME", root.join("home")) };
```

That workaround exists because `test_user_home` used to amend a `stella-cli` thread-local and set no environment variable, so it did not reach `stella_home::stella_home()` — the resolver `stella-observatory` and the user-tier plugin roster both use. #4980 fixes the seam (PR #4992): `test_user_home` now takes the env lock itself and moves `HOME` alongside the thread-local, so the two `set_var` lines are dead weight.

## Why it is not already done

Neither PR can do it alone. The workaround lives on #4977's branch (and the stack above it); the seam fix lives on a branch off `main`. Whichever lands second leaves the other's assumption in place for a while, so this is deliberately a third change rather than a race between two.

## What to do

Once **both** #4992 and #4977 (with #4974's PR on top of it) are on `main`:

1. Delete the `unsafe { std::env::set_var("STELLA_HOME", …) }` line and its explanatory comment from both tests.
2. Drop `"STELLA_HOME"` from each test's `crate::test_env::EnvRestore::capture(&[…])` list — `test_user_home` now captures and restores it.
3. The tests keep `let _env = crate::test_env::lock();` and their `STELLA_TRUST_PROJECT` / `STELLA_PROJECT_HOOKS` mutations; only the home half comes out.

## Verify

`cargo test -p stella-cli --bins -- contributions::the_dashboards_plugin` passes with the lines gone.

[...]

Issue #4980, opened 2026-08-26
test_user_home does not isolate a test that calls into another crate


macanderson:
## Problem

`crates/stella-cli/src/paths.rs`'s `test_user_home` redirects the user tier by amending a **thread-local `UserPaths`** inside `stella-cli`. It sets `stella_root`, `state_home`, `home` and `extensions_visible` on that struct — and it sets **no environment variable**.

Every other crate resolves the same directory through `stella_home::stella_home()`, which reads `STELLA_HOME` (or falls back to `$HOME/.stella`). It cannot see `stella-cli`'s thread-local. So the moment a `stella-cli` test calls into another crate that does its own resolution, `test_user_home` stops isolating and the test reads the developer's real `~/.stella`.

Found while writing `the_dashboards_plugin_skills_honour_the_project_trust_gate` (#4917). That test drives `stella_observatory::respond_with(root, "/api/skills", …)`, and the Observatory's `fsview::user_config_dir()` is `stella_home::stella_home()`. Under `test_user_home` alone the response carried ~45 rows out of `/Users/<me>/.stella/skills` — every skill on the machine, in a test about a temp directory. It passes today only because the assertions are keyed on a name (`house-style`) the developer's home happens not to have; a machine with a skill of that name fails the untrusted assertion, and a machine with no such skill passes the trusted one for the wrong reason. The test now sets `STELLA_HOME` explicitly under the env lock, with a comment — a workaround, not a fix.

## Why it matters beyond that one test

The hazard is silent in both directions. A test that reads the real home can pass for a reason that has nothing to do with the code under test, and — the worse half — a test that *writes* through a non-`stella-cli` resolver writes into the developer's real `~/.stella`. Nothing currently distinguishes the two cases at the call site: `test_user_home(root.join("home"))` reads like full isolation and is partial.

## Options

1. **Make `test_user_home` set `STELLA_HOME` too**, restoring it on drop like `TestPathsGuard` already restores the thread-local. Then one call isolates both resolvers. The catch is that an env var is process-global while `TestPathsGuard` is thread-local, so this only holds under `crate::test_env::lock()` — which most of these tests already take, but not all. Whether that coupling is acceptable is the decision.
2. **A second, explicit seam** (`test_stella_home`) that tests reaching another crate compose on top, the way `test_extensions_visible` composes on `test_user_home` today (#3864). Honest about the two resolvers being two things; costs a line at each such call site and gets forgotten exactly when it matters.
[...]

PR #5003, opened 2026-08-26
feat(stella-observatory,stella-cli): the rules panel lists a plugin's records

### LINKED ISSUE #4974: The Observatory's rules panel cannot see a plugin's context records

macanderson:
> **Rebased onto `main` after #4977 merged.** This was stacked on that branch; the squash landed as `219c23806`, so the stack was replayed with `git rebase --onto origin/main 6876eeffa` rather than merged, which is why the diff is one commit and no conflict remains. #4977's observatory and CLI code landed byte-identical to what this was built on (`git diff 6876eeffa origin/main -- crates/stella-observatory/ crates/stella-cli/src/storage_cmd.rs` is empty), so the shape here still matches the one it landed. Verified by name rather than by count: `cargo test -p stella-cli --bins -- plugin_cmd::tests::contributions` lists all 15 tests, including main's own `a_contributed_skills_tier_is_read_off_contributed_by_not_off_its_path` (#4965) and #4917's `the_dashboards_plugin_skills_honour_the_project_trust_gate` alongside this PR's — and a function-name diff of the file against `origin/main` removes nothing and adds exactly one.

## What

The Observatory's rules panel lists a plugin's contributed context records, names the package that shipped them, and honours the same project-tier trust gate the registry does. Nothing is copied into the user's `.stella/`.

## Same shape as the skills half, and the one place it could not be

The issue asked for a second port modelled on #4917's, and this is that — the same reasoning, the same host, the same per-request call. Two things are different, and both are because a rules file is a **record set**:

**Dedup is by lineage id, and the authority is `stella_cli::context_records::plugin_first`.** It reads contributed directories **first**, and every merge below it is later-wins by lineage, so the workspace's own copy of a lineage is the one that steers. A contributed card whose identity is already on the panel is therefore dropped — listing it would show a steer that is not in force, which is what the skills half does by name. Identity is the lineage id where there is one and the file stem otherwise, because a `.toml` may declare several `[[record]]` entries with their own lineages while a markdown rule has only its name. **Deduping the file would drop records nothing collided with**, and that is the failure mode the set shape introduces; `a_plugins_record_loses_a_shared_lineage_and_keeps_the_rest_of_its_set` is the assertion, with an ablation below.

**No tier column.** A contributed record enters the **project** tier whatever tier its package is installed at (`plugin_first`: "Project tier, never user" — a package's records are a third party's, and reading them as `Trust::User` would let one establish a guard the workspace cannot override). So the package's tier is not the record's trust, and a scope chip here would say something false. `contributed_by` is what the panel gains.

## One port with two questions rather than a second trait

`PluginSkills` becomes `PluginContributions` with `contributed_skill_dirs` **and** `contributed_record_dirs`; `NoPluginSkills` → `NoContributions`; `ContributedSkills` → `ContributedDir`, which is also what `stella_cli::plugin_cmd::package` already calls the identical shape it holds.

This edits the sibling PR's public API, so it is worth reviewing as a decision rather than a tidy-up:

- **The two questions have one answer-holder and one wrong answer.** A host that resolved the roster for skills and did not resolve it for records would put the dashboard's two panels on different trust decisions. One trait means a host implements the roster once.
- **It keeps `serve` to one `Arc`.** A third contributed surface already exists in the tree — `contributed_mcp_servers` (#4733) — and a port per surface grows `serve`, `respond_with`, `handle` and `respond_to_head` each time.
- Invariant #9's single-purpose rule is about tools, not traits; a parameter here scopes nothing and selects nothing.

`ContributedSkills` and a `ContributedRecords` beside it would have been two identical structs in a public API, which is the alternative I rejected.

## The port and its witnesses move to `fsview/contributions.rs`

`fsview.rs` would otherwise have reached 1648 lines, and the ratchet is right to refuse it. The split is a real boundary rather than a line-count dodge: everything else in `fsview` derives what it shows from files this process can see, and nothing in the new module can be derived that way at all — a package's contributions are reached through a roster and a trust gate that live in `stella-cli`, which this crate may not link. `fsview.rs` ends at 1234 lines; the new file is 438 and carries the trait, the two shapes, `NoContributions` and the four contributed-view witnesses.

## Witnesses

Observatory side, against a fake roster (this crate may not link the one that resolves real rosters). The fake holds the two kinds **separately**, because the real roster answers them from two different package subdirectories and a fake returning one list for both would let a view read the other kind's directory and still pass:

- `a_plugins_record_is_listed_and_names_the_plugin` — absent before the roster is asked, listed after, naming the plugin, carrying the fields the panel renders, and the workspace's own row still naming nobody.
- `a_plugins_record_loses_a_shared_lineage_and_keeps_the_rest_of_its_set`.

CLI side, against the real one, through `/api/rules` (the route the page fetches):

- `the_dashboards_plugin_records_honour_the_project_trust_gate` — untrusted checkout contributes nothing and `PACKAGE_RECORD_MARKER` is absent from the response body; trusted, the same bytes contribute once and the body carries `"contributed_by":"vera"`; removed, it is gone from the port and from the panel.

### Ablations

**1. Drop the contributed rows, keep the port call** (so the ablation removes the answer, not the compile):

```
---- fsview::tests::a_plugins_record_is_listed_and_names_the_plugin ----
panicked at crates/stella-observatory/src/fsview.rs:1566:32:
the package's record is listed: Array [ { "contributed_by": Null, "lineage_id": "ctx.house.ours", … } ]

test result: FAILED. 0 passed; 2 failed
```

Both observatory witnesses fail, which is what a total removal should do.

**2. Remove the dedup alone** — the precedence claim on its own:

```
running 2 tests
test fsview::tests::a_plugins_record_is_listed_and_names_the_plugin ... ok
test fsview::tests::a_plugins_record_loses_a_shared_lineage_and_keeps_the_rest_of_its_set ... FAILED

panicked at .../fsview.rs:1631:9: assertion `left == right` failed: one row for one lineage: [
  { "contributed_by": Null,          "lineage_id": "ctx.house.shared", "statement": "OURS_WINS"   },
  { "contributed_by": String("vera"), "lineage_id": "ctx.house.shared", "statement": "THEIRS_LOSES" } ]
  left: 2  right: 1
```

**3. Dedupe by file instead of by lineage** — the failure the set shape introduces, and the reason assertion 2 is not enough on its own:

```
running 2 tests
test fsview::tests::a_plugins_record_is_listed_and_names_the_plugin ... ok
test fsview::tests::a_plugins_record_loses_a_shared_lineage_and_keeps_the_rest_of_its_set ... FAILED

panicked at .../fsview.rs:1643:9: assertion `left == right` failed:
the rest of the package's set is not collateral: [ { "lineage_id": "ctx.house.shared", "statement": "OURS_WINS" } ]
  left: 0  right: 1
```

Ablations 2 and 3 are opposite errors and each fails a different assertion inside the same test, with the parity witness green through both.

**4. Answer the host port from the plugin directory layout instead of from the roster** — the directory scan the port exists to avoid:

```
running 2 tests
test the_dashboards_plugin_records_honour_the_project_trust_gate ... FAILED
test the_dashboards_plugin_skills_honour_the_project_trust_gate ... ok

panicked at crates/stella-cli/src/plugin_cmd/tests/contributions.rs:1089:5:
an untrusted checkout contributes no records to the dashboard
```

It fails at the **untrusted** assertion, which is the one that matters, and the skills witness beside it stays green — so the pair separates the records port from the skills port rather than measuring the same thing twice.

## The page

The rules panel is a card list rather than a table, so `via <plugin>` is a chip beside the record's chips, with the same title text the skills table's `via` column uses.

## Not fixed here
`the_dashboards_plugin_skills_honour_the_project_trust_gate`'s `STELLA_HOME` workaround, and the one this PR's own witness carries for the same reason, both stay. #4980's fix landed as #4992 while this was in review, so the two lines are now dead weight rather than load-bearing — but deleting them is a change with its own verification step (a green run on a machine with no `~/.stella/skills/house-style` proves nothing), and widening a rebased PR to carry it is how a small cleanup becomes the reason a merge slips. #5002 has it, with both of its preconditions now met.

Nothing removed; two added. Four existing tests moved file (`fsview.rs` → `fsview/contributions.rs`) with no change to their bodies; the deleted-test guard compares names across the merge and moves are not deletions, but naming them here anyway.

## Verification

`cargo fmt --all --check`, `cargo clippy -p stella-observatory -p stella-cli --all-targets -- -D warnings`, `RUSTDOCFLAGS=-D warnings cargo doc -p stella-observatory`, `make guards-fast` (prose and the file-size ratchet included), `cargo test -p stella-observatory` (exit 0) and `cargo test -p stella-cli --bins` (2328 passed), plus the rebase-onto-`main` composition check described at the top. CI is the evidence.

Closes #4974
Refs #4917, #3567, #3509, #4980, #4992, #5002

### DIFF
--- a/crates/stella-cli/src/plugin_cmd/tests/contributions.rs
+++ b/crates/stella-cli/src/plugin_cmd/tests/contributions.rs
@@ -1031,7 +1031,7 @@ fn a_contributed_skills_tier_is_read_off_contributed_by_not_off_its_path() {
 }
 
 /// **The dashboard's trust witness** (#4917). The Observatory lists a
-/// package's skills through `storage_cmd::RosterSkills`, and an untrusted
+/// package's skills through `storage_cmd::RosterContributions`, and an untrusted
 /// checkout's package contributes nothing there either.
 ///
 /// The untrusted half is the point. A dashboard that resolved the roster
@@ -1046,11 +1046,11 @@ fn a_contributed_skills_tier_is_read_off_contributed_by_not_off_its_path() {
 /// `an_untrusted_projects_contributed_tool_never_loads`.
 #[test]
 fn the_dashboards_plugin_skills_honour_the_project_trust_gate() {
-    use stella_observatory::PluginSkills as _;
+    use stella_observatory::PluginContributions as _;
 
     /// The dashboard's `/api/skills` body, through the route the page fetches
     /// rather than through an inner function only this test would call.
-    fn skills_json(root: &Path, source: &crate::storage_cmd::RosterSkills) -> String {
+    fn skills_json(root: &Path, source: &crate::storage_cmd::RosterContributions) -> String {
         let response = stella_observatory::respond_with(root, "/api/skills", source);
         String::from_utf8_lossy(&response.body).into_owned()
     }
@@ -1074,7 +1074,7 @@ fn the_dashboards_plugin_skills_honour_the_project_trust_gate() {
     plant(&stella_home::resolve_project_plugins_dir(&root), "vera");
     ships_a_package(&planted, "vera");
 
-    let source = crate::storage_cmd::RosterSkills;
+    let source = crate::storage_cmd::RosterContributions;
 
     // SAFETY: the env lock is held for the whole mutate-read-restore window.
     unsafe {
@@ -1111,6 +1111,88 @@ fn the_dashboards_plugin_skills_honour_the_project_trust_gate() {
     assert!(
         source.contributed_skill_dirs(&root).is_empty(),
         "and it leaves with the package — nothing was ever copied into .stella/skills"
+    );
+
+    let _ = std::fs::remove_dir_all(&root);
+}
+
+/// **The dashboard's trust witness for records** (#4974), the sibling of the
+/// skills one above and gating the same answer.
+///
+/// A record is the surface that steers *quietly*: it shapes every matching
+/// turn and appears in no transcript as something the agent did. So a rules
+/// panel that reads as the complete list of what steers this workspace and
+/// omits a third party's records is worse than absent — and a panel that
+/// listed them out of a checkout nobody has trusted would be worse still,
+/// which is why the untrusted half is asserted first.
+#[test]
+fn the_dashboards_plugin_records_honour_the_project_trust_gate() {
+    use stella_observatory::PluginContributions as _;
+
+    /// The dashboard's `/api/rules` body, through the route the page fetches.
+    fn rules_json(root: &Path, source: &crate::storage_cmd::RosterContributions) -> String {
+        let response = stella_observatory::respond_with(root, "/api/rules", source);
+        String::from_utf8_lossy(&response.body).into_owned()
+    }
+
+    let _env = crate::test_env::lock();
+    let _restore = crate::test_env::EnvRestore::capture(&[
+        "STELLA_TRUST_PROJECT",
+        "STELLA_PROJECT_HOOKS",
+        "STELLA_HOME",
+    ]);
+    let root = temp_root("package-observatory-records-trust");
+    let _paths = crate::paths::test_user_home(root.join("home"));
+    // The user-tier half of the roster is `stella_home::stella_home()`, which
+    // `test_user_home` does not move — without this the roster resolves the
+    // developer's real `~/.stella/plugins` (#4980).
+    // SAFETY: the env lock is held for the whole mutate-read-restore window.
+    unsafe { std::env::set_var("STELLA_HOME", root.join("home")) };
+
+    let planted = stella_home::resolve_project_plugins_dir(&root).join("vera");
+    plant(&stella_home::resolve_project_plugins_dir(&root), "vera");
+    ships_a_package(&planted, "vera");
+
+    let source = crate::storage_cmd::RosterContributions;
+
+    // SAFETY: the env lock is held for the whole mutate-read-restore window.
+    unsafe {
+        std::env::remove_var("STELLA_TRUST_PROJECT");
+        std::env::remove_var("STELLA_PROJECT_HOOKS");
+    }
+    assert!(
+        source.contributed_record_dirs(&root).is_empty(),
+        "an untrusted checkout contributes no records to the dashboard"
+    );
+    let listed = rules_json(&root, &source);
+    assert!(
+        !listed.contains("PACKAGE_RECORD_MARKER"),
+        "and nothing of the package's reaches the rules panel: {listed}"
+    );
+
+    // SAFETY: as above.
+    unsafe { std::env::set_var("STELLA_TRUST_PROJECT", "1") };
+    let dirs = source.contributed_record_dirs(&root);
+    assert_eq!(dirs.len(), 1, "the same bytes contribute once trusted");
+    assert_eq!(dirs[0].plugin, "vera");
+    let listed = rules_json(&root, &source);
+    assert!(
+        listed.contains("PACKAGE_RECORD_MARKER"),
+        "and the rules panel lists it: {listed}"
+    );
+    assert!(
+        listed.contains("\"contributed_by\":\"vera\""),
+        "naming the package that shipped it: {listed}"
+    );
+
+    remove(&root, "vera").expect("remove must succeed");
+    assert!(
+        source.contributed_record_dirs(&root).is_empty(),
+        "and it leaves with the package — nothing was ever copied into .stella/rules"
+    );
+    assert!(
+        !rules_json(&root, &source).contains("PACKAGE_RECORD_MARKER"),
+        "which the panel reports too"
     );
 
     let _ = std::fs::remove_dir_all(&root);
--- a/crates/stella-cli/src/storage_cmd.rs
+++ b/crates/stella-cli/src/storage_cmd.rs
@@ -98,45 +98,71 @@ pub(crate) fn preflight_observatory_stores(
     })
 }
 
-/// The dashboard's answer for "which plugin-contributed skills would a session
-/// started in this workspace load?", read off the one roster the session's own
-/// skill load reads (#4917).
+/// The dashboard's answer for "what would a session started in this workspace
+/// load out of its installed plugins?", read off the one roster the session's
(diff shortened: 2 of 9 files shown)


macanderson:
**Note on CI.** No `pull_request` workflow starts for this PR. Its base is `fix/4917-observatory-plugin-skills` rather than `main`, and although `ci.yml`'s `pull_request:` trigger carries no branch filter (only `bench.yml` does), the only run this PR produced is `cla` — the `pull_request_target` one. I closed and reopened it to force a `reopened` event and got the same result, so it is not a missed trigger. What is blocking it is a repository or Actions setting I cannot read from here rather than anything in this tree, and it is worth someone with settings access confirming, because it means **any** stacked PR in this repository ships without a gate run.

So the evidence for this PR is, in order:

1. A full `ci.yml` run dispatched by hand against this branch head: https://github.com/macanderson/stella/actions/runs/32943992727. That is the real gate — `fmt + clippy + test`, the guards, the file-size ratchet — over the stack, against the `main` this branch is based on (`5ebf4a73c`).
2. The composition check in the description: both commits rebased onto today's `main`, one keep-both conflict resolved in `contributions.rs`, then `cargo test -p stella-cli --bins` (2329 passed) and `cargo test -p stella-observatory` (exit 0).
3. The local targeted runs and the four ablations in the description.

Once #4977 lands and this is rebased onto `main` (`git rebase --onto origin/main 6876eeffa`), the ordinary `pull_request` gate runs and that is the run to judge it by.

macanderson:
This PR's `the_dashboards_plugin_records_honour_the_project_trust_gate` still carries the `STELLA_HOME` workaround:

```rust
let _restore = crate::test_env::EnvRestore::capture(&[
    "STELLA_TRUST_PROJECT",
    "STELLA_PROJECT_HOOKS",
    "STELLA_HOME",
]);
...
// SAFETY: the env lock is held for the whole mutate-read-restore window.
unsafe { std::env::set_var("STELLA_HOME", root.join("home")) };
```

It is dead weight since #4992 merged: `crate::paths::test_user_home` now calls `crate::test_env::home_sandbox`, which sets `HOME`, clears every `stella_home::OVERRIDE_ENV_VARS` entry (`STELLA_HOME` included) and captures them for restore — so `stella_home::stella_home()` already resolves inside the fixture. It is also slightly wrong: it points the stella home at `<fixture>/home`, while every other resolver in the test agrees on `<fixture>/home/.stella`.

#5012 removes the identical lines from the sibling test this one is modelled on (`the_dashboards_plugin_skills_honour_the_project_trust_gate`), with the contamination ablation #5002 asks for. Please drop the same three lines plus the `"STELLA_HOME"` capture entry here before merging, so #5002 can close.

Refs #5002, #4980, #4992

PR #5013, opened 2026-08-26
docs(wire): give the wire contract an id so Rust can cite it

### LINKED ISSUE #4950: docs/wire/README.md is the normative wire contract and nothing can cite it
## Problem

`docs/wire/README.md` is the normative statement of the `AgentEvent` wire contract — §3 is where "the contract is **additive-only**" lives, and §4 is where "a `"type"` this schema does not list is **not** an error [...] which is what `AgentEvent::Unknown` does on the Rust side" lives. Both are arguments Rust code needs to cite when it explains why a wire decision went the way it did.

It cannot be cited. It has no frontmatter, so `make doc-links` refuses:

```
$ make doc-links
  crates/stella-protocol/src/event/kind.rs:1028 cites 'docs/wire/README.md', which exists but declares no
     frontmatter `id:`. Adopt it (add frontmatter) or stop citing it.
     Nothing under docs/ is citable until it has an identity.
```

`make doc-report` lists it among the unidentified:

```
      194L  docs/wire/README.md  (no status, 22 hours ago)
```

## Why it matters

The guard is right that an unidentified document is not citable — that is the whole design (`docs/README.md § How to cite a document`). The consequence here is that a doc comment wanting to say *why* a producerless wire tag is kept has to either paraphrase the contract inline or drop the citation. PR #4930 took the second option and reworded around it, which is the wrong trade for a document this normative: the reader now has to find the contract themselves.

## Fix

```
make doc-adopt DOC=docs/wire/README.md
```

then repoint the citations. Consider `id: wire-contract` (short, and `doc:wire-contract §3` reads well at a call site).

## Watch out for

`docs/manifest.json` is a **shared cell**: every `doc-adopt` writes it, and two PRs adopting different documents compose into a conflict — the same shape as `Cargo.lock` and `scripts/file-size-baseline.txt` (AGENTS.md § "God files", and `main-canary.yml`'s reason for existing). Land this on its own from a fresh `main` rather than folding it into an unrelated PR.

Also check whether the four generated siblings in that directory want identities too — `agentevent.schema.json`, `agentevent.d.ts`, `serveframe.*`, `wrapper.*` are generated artifacts rather than prose, and probably do not; `docs/wire/README.md` is the only hand-written page there.

## Done when

`docs/wire/README.md` carries a frontmatter `id`, `make doc-links` accepts a citation to it, `make doc-report` no longer lists it as unidentified, and at least one real citation uses the id form — `crates/stella-protocol/src/event/kind.rs`'s `MediaProgress` doc is the natural first one, since it is the comment that wanted it.

Noticed while working #4454 (PR #4930).

macanderson:
## What

`docs/wire/README.md` gets a frontmatter `id: wire-contract`, a manifest entry,
and its first real citation.

- `make doc-adopt DOC=docs/wire/README.md` scaffolded `id: wire`; renamed to
  `wire-contract`, which is what the issue suggests and what reads correctly at
  a call site (`doc:wire-contract §3`).
- `make doc-links-fix` wrote the manifest entry, including the four numbered
  sections the page exposes, so `§1`–`§4` citations resolve.
- `crates/stella-protocol/src/event/kind.rs`'s `MediaProgress` doc — the comment
  #4930 had to reword around — now cites `doc:wire-contract §4` for
  `AgentEvent::Unknown` and `doc:wire-contract §3` for additive-only, instead of
  paraphrasing both.

## Evidence

Before, `docs/wire/README.md` was in the unidentified/uncited list:

```
$ make doc-links
check-doc-links: OK -- 1100 citation(s) resolve across 84 identified document(s).
  nothing cites these 6 document(s) -- retire or adopt:
      ...
      194L  docs/wire/README.md  (no status, 27 hours ago)
```

After:

```
$ make doc-links
check-doc-links: OK -- 1102 citation(s) resolve across 85 identified document(s).
  nothing cites these 5 document(s) -- retire or adopt:
      ...   (docs/wire/README.md is gone from the list)
```

**Ablation** — the adoption reverted (frontmatter and manifest entry stashed),
the two citations kept:

```
$ make doc-links
check-doc-links: FAILED

  crates/stella-protocol/src/event/kind.rs:1074 cites doc:wire-contract, which is not a known document id.
     Known ids live in docs/manifest.json. If the document exists but has
     no frontmatter, add an `id:` to it -- an undeclared document is
     deliberately not citable.
  crates/stella-protocol/src/event/kind.rs:1076 cites doc:wire-contract, which is not a known document id.
     ...
make: *** [doc-links] Error 1
```

So the guard genuinely refuses the citation without the adoption — the answer
changes from OK to FAILED, not merely from one count to another.

`make guards-fast`: green (prose, line-citations, css-vars, reserved-paths,
lockfile-sync, fmt --check all pass).

## Notes

Per the issue: this lands on its own from a fresh `main`, because
`docs/manifest.json` is a shared cell every `doc-adopt` writes. The four
generated siblings in that directory (`agentevent.schema.json`,
`agentevent.d.ts`, `serveframe.*`, `wrapper.*`) are artifacts rather than prose
and get no identity — `docs/wire/README.md` is the only hand-written page there.

## Witness

Docs/citation change. The ablation above is the proof: the guard fails without
the adoption and passes with it.

Closes #4950

## Summary by Sourcery

Adopt the wire contract documentation into the citation system and reference its compatibility guarantees from the protocol event documentation.

New Features:
- Identify the wire contract documentation with the `wire-contract` ID and expose its four citable sections in the documentation manifest.

Enhancements:
- Replace protocol documentation paraphrases with citations to the wire contract's rules for unknown event tags and additive-only compatibility.

Documentation:
- Add frontmatter metadata to the wire contract documentation so it can be referenced by stable document ID.


macanderson:
`wire-schema` was red on this PR and it was my change, not a stale `main`. Both halves checked:

**`main` is green.** `make wire-schema` on a clean `origin/main` checkout at `464ef364d`:

```
check-wire-schema: OK — docs/wire/ matches the types, and names no Rust path.
```

So there is no red `main` to report and nothing to repair separately.

**The cause.** `schemars` derives each schema member's `description` from the Rust doc comment, so adding the two `doc:wire-contract` citations to `AgentEvent::MediaProgress` **is** a change to the generated artifacts. `serveframe.schema.json` carries the same line because it embeds the `AgentEvent` payload graph.

Rebased onto `origin/main` first, then `make wire-schema-update`. The regeneration produced exactly two lines and nothing else:

```
 docs/wire/agentevent.schema.json | 2 +-
 docs/wire/serveframe.schema.json | 2 +-
 2 files changed, 2 insertions(+), 2 deletions(-)
```

both being the `MediaProgress` `description` string, with `which is what makes the generated wire contract additive-only` becoming `which is what makes the contract additive-only (\`doc:wire-contract §3\`)`. `make wire-schema` is green on the branch now.

The `serveframe.schema.json` line in the original CI report was the stale-base half — the branch was cut at `0e4ea965a` and `main` moved underneath it. The rebase resolved that; nothing in the regenerated diff reverts anything that landed in between.

PR #5065, opened 2026-08-26
fix(stella-tui): the dead v1 compaction arm delegates to the projection

### LINKED ISSUE #5035: Compaction has two render paths; keep the one quiet line
## Problem

Compaction has two render paths: the v2 quiet one-liner (`↓ compacted 74k→69k · 0 evicted · 0 deduped`, `views/transcript.rs::event_rows` early return, live via `transcript_source::compaction_rows`) and a v1 `push_note` row near `crates/stella-tui/src/render/entry.rs:~510`. Two paths for one event is how the same session prints compaction twice, or differently.

## Files

- `crates/stella-tui/src/render/entry.rs` — retire the v1 note row (or prove it unreachable and delete it)

## Done means

One compaction entry produces exactly one dim line; a test renders a compaction through the full projected path and asserts a single row.

Spec: `design/tui-v2/SPEC.md` §6.3. Related: #4721.

**Estimated agent build time:** ~1h
**Model:** use-model:cheap — Haiku 4.5 (`claude-haiku-4-5-20251001`)

macanderson:
The audit behind #5035 read this as a possible double print; the truth is the opposite defect. `projected_rows` (render/entry.rs) intercepts every `Compaction` entry and returns `true`, so `entry_body`'s long-form arm was unreachable — while still drawing the retired `⇣ compacted … tok` row. A dead match arm is invisible to `dead-code-allows` and `module-reachability` (both see items), which is exactly the #4157 shape the router's own doc comment warns about: it sits looking live until the router narrows, then silently ships stale rendering.

The arm now **delegates** to the projection, the pattern the `ToolStart` arm already established, so there is one compaction renderer and no second one to rot.

Witness: `the_compaction_arm_agrees_with_the_projection_it_shadows` calls `entry_body` directly and asserts its rows equal the projection's. Verified the flip artisanally: with the old arm restored the test fails on the drifted text; with the delegation it passes. `cargo clippy -p stella-tui --all-targets` exits 0.

Closes #5035

## Summary by Sourcery

Route compaction entries through the canonical projection and verify that they render exactly one consistent quiet line.

Bug Fixes:
- Ensure compaction entries use the single projected quiet-line renderer instead of retaining an unreachable stale rendering path.

Enhancements:
- Add regression coverage confirming compaction output matches the projection and renders only one quiet line through the full entry path.

Tests:
- Add tests that detect drift between direct entry rendering and projected compaction rendering, and prevent duplicate compaction output.

### DIFF
--- a/crates/stella-tui/src/render/entry.rs
+++ b/crates/stella-tui/src/render/entry.rs
@@ -506,25 +506,14 @@ fn entry_body(
                 out,
             );
         }
-        TranscriptEntry::Compaction {
-            before_tokens,
-            after_tokens,
-            evicted,
-            deduped,
-        } => {
-            push_note(
-                "⇣ compacted",
-                quiet(),
-                vec![
-                    Span::styled(format!("{before_tokens}→{after_tokens} tok"), value()),
-                    Span::styled(
-                        format!("  ·  {evicted} evicted · {deduped} deduped"),
-                        quiet(),
-                    ),
-                ],
-                width,
-                out,
-            );
+        // The router's (SPEC 6.3's one quiet line). This arm delegates for the
+        // same reason `ToolStart` below does: it used to hold a live-looking
+        // v1 row (`⇣ compacted … tok`) that `projected_rows` had made
+        // unreachable, and a dead match arm drifts silently — neither
+        // `dead-code-allows` nor `module-reachability` sees inside a match
+        // (#4157's lesson, restated at the router).
+        TranscriptEntry::Compaction { .. } => {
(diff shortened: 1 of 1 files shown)


macanderson:
The assessment's third row predates the head: commit 'test(stella-tui): the compaction witness counts rows through the full path' pins the projection to exactly one row, and a_compaction_renders_one_quiet_line_through_the_full_path drives entry_lines — the path every frame takes — asserting exactly one compacted row plus its block-closing spacer and nothing else. That is the explicit row-count coverage the issue's done-means asked for.

PR #5070, opened 2026-08-26
fix(stella-tui): the no-plan SESSION row shows the tab list, not dead chrome

### LINKED ISSUE #5049: On the SESSION tab you cannot see the other eight tabs
## Problem

On the SESSION tab — the default screen — the tab list is not drawn at all: `views/frame.rs::render_tab_row` branches to `breadcrumb_spans` for `DeckTab::Session` and `tab_list_spans` for everything else (an intentional merge, documented in the module doc). The result: a user on the default screen cannot see that AGENTS/TRACES/GRAPH/FILES/SKILLS/MCP/ISSUES/SETTINGS exist. SPEC 5 asks for the tab bar row (1) *and* the breadcrumb strip row (2) on SESSION. Related smaller defect: at 120 cols the 9-tab bar overflows (SETTINGS clips), and `render_tab_row` drops the wordmark rather than the tabs when tight (`tests/deck_snapshot.rs:84`).

## Files

- `crates/stella-tui/src/views/frame.rs` — either restore the two-row layout on SESSION or design a merged row that still names the other tabs (e.g. compact tab strip + breadcrumb); regenerate the deck goldens and read the diff
- `crates/stella-tui/src/deck_render.rs` — band height if a second row is added

## Done means

From the SESSION tab the other eight tabs are discoverable on screen; narrow widths degrade without losing the wordmark silently; goldens updated and reviewed. If the single-row merge is kept, the spec is amended in the same PR to say so.

Spec: `design/tui-v2/SPEC.md` §5.

**Estimated agent build time:** ~4h

macanderson:
The SESSION tab merges the tab bar and the plan breadcrumb into one row (the form the renderings draw), which left the default screen with no tab list at all: a new reader saw `SESSION  ▸ no plan yet` and had no on-screen evidence that AGENTS/TRACES/GRAPH/FILES/SKILLS/MCP/ISSUES/SETTINGS exist (#5049).

`breadcrumb_spans` now returns `Option`: `Some` for the agent path inside a lane and for a non-empty plan — unchanged — and `None` otherwise, which hands the row to `tab_list_spans(DeckTab::Session)`. The moment a plan is proposed or a lane opens, the breadcrumb takes the row back, so mid-work density matches the renderings exactly.

`design/tui-v2/SPEC.md` §5 item 2 is amended in the same PR to record the shared-row rule, the way §11's `^S` amendment recorded that decision.

Witness: `with_no_plan_the_session_row_is_the_tab_list` — all nine tab titles present, `no plan yet` absent, wordmark on the right edge, SESSION lit gold. Verified the flip artisanally: with the old `no plan yet` fallback restored at the call site the test fails naming the first missing tab; with the fix it passes. Full `cargo test -p stella-tui` (12 suites, deck goldens included — the golden scenario carries a plan, so no snapshot churn), clippy `--all-targets`, and `make prose` (spec edit) are green.

Not addressed here, still open under #5049's sibling scope: the 120-column overflow that clips SETTINGS on the widest tab list — that is a width-degradation design for `tab_list_spans` and worth its own small PR; noted on the issue.

Closes #5049

## Summary by Sourcery

Restore the SESSION tab list when its breadcrumb has no content to display.

Bug Fixes:
- Show the SESSION tab list when no plan is active and no lane is open, making all available tabs visible on the default screen.

Enhancements:
- Clarify the shared SESSION breadcrumb/tab-row behavior in the TUI specification and view documentation.

Documentation:
- Document the SESSION row fallback rule in the TUI v2 specification.

Tests:
- Add coverage verifying the no-plan SESSION row displays all tabs, preserves the wordmark, and highlights SESSION.

### DIFF
--- a/crates/stella-tui/src/views/frame.rs
+++ b/crates/stella-tui/src/views/frame.rs
@@ -22,10 +22,17 @@
 //!
 //! ## The tab row
 //!
-//! One row, no border. On the SESSION tab it is the breadcrumb strip: the tab
-//! name, then the plan's position. Everywhere else it is the tab list with the
-//! active tab in gold. `stella*` holds the right edge on every screen
-//! (SPEC 3.3).
+//! One row, no border. On the SESSION tab it is the breadcrumb strip — the tab
+//! name, then the plan's position, or the agent path inside a lane — and the
+//! tab list everywhere else, with the active tab in gold. `stella*` holds the
+//! right edge on every screen (SPEC 3.3).
+//!
+//! A SESSION with **no plan and no opened lane** shows the tab list too: the
+//! breadcrumb had nothing to say there (`▸ no plan yet`, dead chrome on the
+//! default screen), and that row was the only place a new reader could have
+//! learned the other eight tabs exist (#5049). The moment a plan or a lane
+//! gives the breadcrumb something to say, it takes the row back — which is
+//! also the form the renderings draw.
 
 use ratatui::buffer::Buffer;
 use ratatui::layout::Rect;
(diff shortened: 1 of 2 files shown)


macanderson:
On the assessment's second row (narrow-width degradation): deliberately out of this PR's scope, and tracked rather than dropped — #5072 carries the 120/100-column width-degradation design for tab_list_spans with its own golden coverage. This PR fixes the default screen's discoverability half of #5049; the overflow predates it and now has its own handoff.

PR #5073, opened 2026-08-26
fix(stella-cli): stop a turn claiming file changes another live session made

### LINKED ISSUE #4386: A session's file-change report can wrongly include edits made by a different session running at the same time
## Problem

`crates/stella-cli/src/turn_files.rs` measures a turn's file changes as a work-tree snapshot diff at the turn boundary. Its module doc already says a user editing in another window appears "indistinguishable from the agent's own writes". With several `stella` sessions in one checkout (which is how the maintainer runs it) that is not an edge case: the other sessions' writes land in this session's export under "Files", presented as this turn's work.

## Evidence (session export `ses-1787465453163-60967` (`.stella/exports/session-2026-08-23-08-37-08.zip`, binary 0.9.143, OpenRouter `moonshotai/kimi-k3`); events re-read from `.stella/private/store.db`, `files_touched` table)

- Execution 225 (prompt: "send out agents to research how the agent engine works" — read-only research; its tool calls are `delegate`/`read_file`/`search`/task board only) is recorded as **creating `scripts/prose_score.py` (+453)** and editing `CODE_OF_CONDUCT.md`, `docs/papers/README.md`, three `docs/spec/*.md`. Those are execution 226's work — a different session (`ses-1787465483895-62121`, prompt "research all the files that are markdown and develop a scoring…") running 06:13–06:57 in the same checkout.
- The same files recur with identical counts in 234 and 245 (`docs/spec/enterprise-authority-telemetry.md` +26/-26 in both 225 and 245), so a reader of the export sees this session repeatedly "touching" docs it never opened.
- The 245 reflection even mined a lesson about "the user's working tree routinely carries unrelated uncommitted changes" — the agent noticed the contamination and recorded it as a repo convention.

## Fix options

Scope the measurement to paths the turn's own tool calls named (`own_change` already exists for the diffless case, see #4367) and label the remainder as `unattributed` rather than as this turn's; or diff against the session's own previous snapshot *and* subtract every other live session's snapshot deltas (the session registry under `~/.stella/sessions/` knows who else is running). At minimum the export should carry the `reason` string through so the dashboard can say "measured, not attributed".

## Done when

- A test with two sessions' snapshot refs in one repo: a file written by session B does not appear in session A's `files_touched`, or appears flagged.

macanderson:
Closes #4386

## The mechanism, confirmed

`crates/stella-store/src/work_journal.rs` gives each session a private
snapshot ref (`refs/stella/<session>/tree`) and a private git index
(`{workspace_id}.{session}.snapshot.index`), but one **shared** git dir over
one **shared** work tree. `WorkJournal::snapshot_worktree` then stages that
whole shared tree against its own index and diffs against its *own* previous
snapshot. So session A's boundary sweep absorbs every byte session B wrote
since A last snapshotted, and `turn_files::file_touch_row` recorded it as
`"reason": "turn-boundary work-tree measurement"` — this turn's work.

The module doc already stated the hazard ("a user editing a file in another
window mid-turn appears in this stream indistinguishable from the agent's own
writes") and treated it as an edge case. With several sessions in one checkout
it is the normal case.

## Why a label and not a filter

Both sessions snapshot the same bytes, so the trees converge and no content
they carry can tell the writers apart — only the snapshot *times* differ, and
those windows overlap. Two signals do survive:

- the paths a mutating call read for itself before and after it wrote
  (`stella_tools::own_change`), which is evidence about this session whatever
  else is in the tree;
- whether any other session was live in this work tree at all, which
  `SessionRegistry` knows (`workspace` plus `SessionStatus::is_live`, with the
  pid downgrade `list()` already applies).

Neither yields "B wrote this". Together they are enough to stop the record
*claiming* what it cannot know.

**Filtering would be worse than the defect.** `bash`, MCP servers and script
tools mutate the tree without naming a path in any schema the engine reads —
the module doc's own reason for measuring rather than hooking — so a filter
scoped to named paths would drop real changes while looking exhaustive. Every
measured change is still reported; a change the reading cannot attribute now
says so.

## What changed

New `crates/stella-cli/src/turn_files/attribution.rs` (a sibling submodule —
`turn_files.rs` is over 1300 lines and AGENTS.md's god-file rule says new
logic lands beside, not inside):

- `Provenance::{OwnReading, SoleWriter, Unattributed}` with `reason()` and
  `attributed()`.
- `provenance(path, own, sharers)` — the pure classification.
- `sharers_of(records, workspace, own_session)` — pure over a registry
  listing, so the rule is testable without a registry directory. Paths are
  compared canonicalized, because `/tmp` is a symlink to `/private/tmp` on
  macOS and #4386 was reported from macOS; a literal comparison would have
  shipped an inert guard.
- `live_sharers` — the one impure half.

`SessionDurability` caches the sharer set and refreshes it at a turn's two
bookends (`open_turn_streams`, `emit_shared_tree_changes`). Per turn, not per
reading: a reading is taken after every solo mutating call, and the answer is
a directory walk. `WorkJournal` gained `session()` and `work_tree()`
accessors so the handle can ask the question about itself.

Both row builders now go through one `touch_events(provenance, added,
removed)`, so `attributed` is present on a tree reading *and* on a call's own
reading — an export that rendered it for one and omitted it for the other
would make its absence mean two things.

## Witness

`a_file_another_live_session_wrote_is_not_claimed_by_this_turn` is the issue's
"Done when" literally: two `WorkJournal`s on one temp work tree, a live
registry record for each, session B writes a file, session A (which did
nothing at all) snapshots.

Ablation — `provenance` returns `SoleWriter` unconditionally, which is the
pre-#4386 behaviour:

```
test turn_files::attribution::tests::a_shared_tree_makes_a_measured_change_unattributable ... FAILED
test turn_files::attribution::tests::a_path_the_call_read_itself_survives_a_shared_tree ... FAILED
test turn_files::tests::a_file_another_live_session_wrote_is_not_claimed_by_this_turn ... FAILED

panicked at crates/stella-cli/src/turn_files.rs:1382:9:
assertion `left == right` failed: a change measured while another session
shared the tree must not be recorded as this turn's:
[{"attributed":true,"event":"measured","lines_added":1,"lines_removed":0,
  "reason":"turn-boundary work-tree measurement"}]

test result: FAILED. 20 passed; 3 failed
```

That JSON is the defect in one line: session B's file, in session A's ledger,
attributed to A.

With the fix:

```
$ cargo test -p stella-cli --bin stella turn_files
test result: ok. 23 passed; 0 failed; 0 ignored; 0 measured; 2317 filtered out
```

Seven new tests: the three above, plus
`a_live_session_in_the_same_tree_is_a_sharer`,
`a_finished_session_and_another_workspace_are_not_sharers`,
`two_spellings_of_one_tree_are_the_same_tree` (the macOS symlink shape),
`a_lone_session_owns_every_change_it_measures`,
`only_the_unattributed_reason_disclaims_authorship`, and
`a_lone_session_still_claims_what_it_measured`.

**No test was renamed or removed** (`check-deleted-tests`).

## The export

`export.rs` dumps `files_touched` wholesale as JSON, so `reason` and
`attributed` reach the archive with no further work — the issue's "at minimum"
bar. The Files *panel* does not yet render the distinction; filed as a
follow-up rather than smuggled in here.

## Residuals, stated

- **An unreadable registry reads as an empty one.** `SessionRegistry::list`
  answers both identically, so a broken `~/.stella` reports no sharers and
  rows claim authorship as they did before. That is the status quo on a path
  where the home is already unusable, not a new way to be wrong. Named in the
  module doc.
- **A session bound to a journal key that is not a registry id** (a fleet
  attempt, keyed by run and task) reports as alone. Correct here: those run in
  their own worktrees, so nothing else shares the tree they measure.
- **Two sessions genuinely editing one path** are both unattributed, which is
  the honest answer.
- Nothing is dropped and no wire event changed, so no consumer of
  `AgentEvent::FileChange` moves.

## Guards run locally

`make guards-fast`, `cargo fmt --all`,
`cargo clippy -p stella-cli -p stella-store --all-targets -- -D warnings`,
`cargo check -p stella-cli --all-targets` — all green. The workspace build and
suite are CI's, per CLAUDE.md.

## Summary by Sourcery

Prevent shared work-tree measurements from claiming another live session’s changes as the current turn’s work.

New Features:
- Add provenance metadata to file-touch records so measured changes identify whether they can be attributed to the current session.

Bug Fixes:
- Stop turns from claiming work-tree changes made by another live session sharing the same workspace.
- Preserve attribution for paths confirmed by the current call's own before-and-after reading while marking other shared-tree changes as unattributed.

Enhancements:
- Cache live workspace sharers per turn and refresh the cache at turn boundaries.
- Compare workspace paths canonically when identifying sessions sharing a work tree.
- Expose work-journal session and workspace information for attribution decisions.

Documentation:
- Document the attribution limitations and why shared-tree changes are labelled rather than filtered.

Tests:
- Add coverage for live-session detection, canonicalized workspace paths, sole-session attribution, own-path attribution, and the cross-session contamination scenario.

### DIFF
--- a/crates/stella-cli/src/durability.rs
+++ b/crates/stella-cli/src/durability.rs
@@ -91,6 +91,16 @@ pub struct SessionDurability {
     /// belongs to. `Arc` so `sink()` hands out cheap shares rather than
     /// re-opening the record per engine.
     bound: Arc<RwLock<Option<Bound>>>,
+    /// The other live sessions writing to this work tree, as of the last
+    /// [`Self::refresh_worktree_sharers`] (#4386).
+    ///
+    /// Cached rather than asked per reading because the answer is a walk of the
+    /// session registry directory and a reading is taken after every solo
+    /// mutating tool call. Refreshed at a turn's two bookends, which is the
+    /// resolution the answer is used at: a session that starts mid-turn is
+    /// caught by the closing sweep, and the opening one is what a long turn's
+    /// per-call readings consult.
+    sharers: Arc<RwLock<Vec<String>>>,
 }
 
 /// What one work-tree measurement produced.
@@ -221,6 +231,30 @@ impl SessionDurability {
         // After the write lock is released: `snapshot_worktree` takes the read
         // lock, and this is not a re-entrant lock.
         let _ = self.snapshot_worktree();
+    }
+
+    /// Re-ask the session registry which other live sessions are writing to
+    /// this work tree, and cache the answer for this turn's readings (#4386).
+    ///
+    /// Called at a turn's bookends. An unbound handle measures nothing, so it
+    /// has nothing to attribute and the question is not asked.
+    pub fn refresh_worktree_sharers(&self) {
+        let Some(journal) = self.journal() else {
+            return;
+        };
+        let found =
+            crate::turn_files::attribution::live_sharers(journal.work_tree(), journal.session());
+        *self.sharers.write().unwrap_or_else(|p| p.into_inner()) = found;
+    }
+
+    /// The sharer set the last refresh found — empty when this session is alone
+    /// in the tree, which is the only state in which a measurement may be
+    /// claimed as this session's work.
+    pub fn worktree_sharers(&self) -> Vec<String> {
+        self.sharers
+            .read()
+            .unwrap_or_else(|p| p.into_inner())
+            .clone()
     }
 
     /// Declare that this session's turns are running inside a staged pipeline,
--- a/crates/stella-cli/src/turn_files.rs
+++ b/crates/stella-cli/src/turn_files.rs
@@ -53,6 +53,12 @@
 //! authority on what the agent did is the pipeline's adoption, which measures
 //! a candidate against a sealed baseline and can tell the two apart.
 //!
+//! The durable half no longer *states* what it cannot know: since #4386 each
+//! `files_touched` row carries the provenance of the reading that produced it,
+//! and a row measured while another live session shared the work tree says so
+//! rather than naming this turn as the author. [`attribution`] holds that
+//! decision and the argument for labelling rather than filtering.
+//!
 //! It lives beside `agent.rs` rather than inside it because `agent.rs` sits
 //! close to the 1500-line ratchet (AGENTS.md § "God files — plan around them,
 //! never into them") — new logic lands in a sibling.
@@ -69,6 +75,10 @@ use stella_tools::own_change::{OwnChange, OwnChangeKind};
 
 use crate::config::Config;
 use crate::durability::{SessionDurability, UnmeasurableReason, WorktreeSnapshot};
+
+pub(crate) mod attribution;
+
+use attribution::Provenance;
 
 /// This module's `module_path!()`, so its records filter under one target.
 const DIAG_TARGET: &str = "stella::turn_files";
@@ -116,7 +126,7 @@ impl stella_tools::call_measure::CallMeasure for TurnCallMeasure {
     /// calls the boundary's own function rather than a per-call variant of it.
     fn measure_and_publish(&self, own: &[OwnChange]) {
         let measured =
-            emit_measured_tree_changes(&self.durability, &self.tx, self.execution.as_ref());
+            emit_measured_tree_changes(&self.durability, &self.tx, self.execution.as_ref(), own);
         emit_own_changes(&self.tx, self.execution.as_ref(), own, &measured);
     }
 }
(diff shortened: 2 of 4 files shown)


macanderson:
## The cost this adds, measured

Asking who else is in the tree is a walk of `~/.stella/sessions`, so it is worth stating what that costs rather than leaving it implicit.

`SessionRegistry::list()` reads and parses every session record ever written. On the maintainer's machine today:

```
record files: 2801
parsed 2801 in 0.239s
```

(a plain read+parse of the same files; `scan` adds one `pid_alive` syscall per live record.)

This PR pays it **twice per turn** — once at `open_turn_streams`, once at `emit_shared_tree_changes` — and never per tool call, which is why `SessionDurability` caches the answer at all: a reading is taken after every solo mutating call, and paying a directory walk there would have been a real regression. On the goal arcs, where `emit_shared_tree_changes_raw` runs per round, it is once per round.

Roughly half a second against a turn's model latency, and bounded by turns rather than calls. I think that is the right trade for not misattributing another session's work, but it is a trade and it grows: nothing prunes that directory on a schedule, so the number only goes up.

Filed as #5082, with the measurement and three candidate directions — the cheapest being a liveness pre-filter on the filename, since `ses-<ms>-<pid>` already carries the pid and live records are the minority. That is a fix to a cost several existing callers already pay (`latest_resumable`, the SESSIONS view, `doctor`, `whistle`); this PR is just the first caller on a per-turn cadence.

Issue #4756, opened 2026-08-25
`stella-transcript::grid` still draws the box frame, not SPEC 6


macanderson:
Split out of #4392.

## Problem

`crates/stella-transcript/src/lib.rs`'s module header recorded a decision
(2026-08-22, owner's call, prompted by #4271): SPEC 6 — `design/tui-v2/SPEC.md`
§6's full-width labelled rule, railed event rows, closing rule and one-line
receipt — is the frame this crate draws, and `grid` changes to draw it rather
than the deck being bent onto `grid`'s box.

The decision was written in the present tense as if it had shipped. It has not.
`crates/stella-transcript/src/grid.rs`'s render entry points (`render`,
`render_turn_lines`, `turn_frame_top`) still emit `╭─` / `│` / `╰──╯`, and
`grid::Color::Cyan` still exists. #4392's audit found the header claiming a
change that no code makes; the header now says so, and points here.

## Work

1. Change `grid`'s frame to SPEC 6. `grid` keeps emitting a *role* and encoding
   at the very end — it must not gain a `stella-tui-theme` dependency. Add the
   two metals to `grid::Color` as roles; `stella-tui` maps role →
   `stella_tui_theme::token` RGB at its own encode step.
2. Retire `grid::Color::Cyan` for tool identity — a per-class hue is what SPEC
   3.2's clamp rejects (see #4125).
3. Point `stella-tui`'s `v2::transcript` at `grid::render` and drop its own
   frame, then update `scripts/check-transcript-surfaces.py`'s row for it.

## Blocked on first

`v2::transcript`'s file-event fields are all `Option`, because a head row
renders before anything is measured and `+0 -0` beside a path is a louder claim
than "not measured yet". `crates/stella-transcript/src/model.rs` has no notion
of an unmeasured row — the same hole #4181 describes from the accessibility
side. Close that before the deck renders through `grid`, or the migration
reintroduces the fabricated zero.

## Verify

- `crates/stella-tui/tests/deck_render_snapshots.rs` — regenerate with
  `BLESS=1 cargo test -p stella-tui --test deck_render_snapshots` and **read
  the diff**.
- `cargo test -p stella-transcript`
- `make transcript-surfaces`

## Definition of done

- `rg '╭─' crates/stella-transcript/src/grid.rs` finds nothing.
- One frame renders on both the deck and the plain `stella run` transcript,
  from the same `grid` entry point.
- `crates/stella-transcript/src/lib.rs`'s header states what `grid` draws in
[...]

Issue #4270, opened 2026-08-22
The exported session archive renders its own HTML instead of using the shared transcript renderer


macanderson:
## Problem

`stella export`'s HTML transcript is a second, independent HTML renderer. `stella_transcript::html::render_page` has **exactly one** production caller in the workspace:

```
$ rg -n "stella_transcript::html|html::render_page" --glob "**/*.rs" | grep -v "^crates/stella-transcript/"
crates/stella-observatory/src/lib.rs:183:    Ok(stella_transcript::html::render_page(&run, &state))
```

`crates/stella-cli/src/export/transcript.rs` (874 lines) folds `Store::session_events` into its own stream and emits its own HTML. It does not build a `stella_transcript::model::Run` and it does not call `html::render_page`. It reaches into the shared crate only for leaf helpers (`word::highlight` at :765). Its own comment at :752 records the divergence explicitly:

> `stella_transcript::file_diff::emit_change_block` rather than calling it.

So the same session rendered by `stella observe` and by `stella export` produces two different-looking HTML documents. This is the exact drift `stella-transcript` was extracted to end (`crates/stella-transcript/src/lib.rs` charter: "One transcript model, two renderers").

## Why it was not simply wired up

Three properties of the export fold are required and are **not** in the shared model. Any migration has to carry them or it is a regression:

1. **Redaction.** This is a third egress sink and #817's masking does not cover it. Every string taken off an event goes through `stella_core::redact::redact_secrets` before escaping (`Fold::clean`). The archive is meant to be mailed to someone. `stella_transcript` is a near-leaf and cannot depend on `stella-core`, so the redaction has to happen in the fold that builds the `Run`, before the renderer sees it.
2. **The one-second elapsed clock.** `events.ts` is SQLite `DEFAULT CURRENT_TIMESTAMP` with no sub-second component, so the `.t` column prints whole seconds while `duration_ms` chips stay millisecond-precise. A renderer that formats both from one field invents precision.
3. **One model message renders once.** `TextDelta` runs plus the consolidated `Text` event must collapse to a single paragraph (`Fold::flush_prose`), or the artifact shows every assistant message twice and reads as a second paid call.

## Files

- `crates/stella-cli/src/export/transcript.rs` (874 lines) — the fold + HTML emitter to retire
- `crates/stella-transcript/src/html.rs` (644 lines) — the target renderer
- `crates/stella-observatory/src/transcript_view.rs` — the worked example: an existing fold from stored rows into `stella_transcript::model`
[...]

PR #5076, opened 2026-08-26
fix(stella-transcript,stella-tui,stella-observatory): a size nobody measured is absent, not zero

### LINKED ISSUE #4289: Make interactive mode draw its transcript frame through the one shared renderer

macanderson:
`stella-transcript`'s model had no notion of an **unmeasured** row, so a size that was never measured and a size measured as zero were the same value. `+0 −0` beside a path asserts the edit changed nothing — a louder and entirely different claim than "nothing measured this" — and it was shipping on both live surfaces.

This is **#4289's piece 1**, the precondition its own charter names: moving the deck onto `grid` before it lands would reintroduce the zero that `views::transcript::Extent`'s `Option` fields exist to prevent. It does not close #4289 (pieces 2–4 remain) or #3578.

## What landed

`model::Extent { added: Option<usize>, removed: Option<usize> }` is the notion, mirroring the rule `stella-tui`'s `views::transcript::Extent` already states for the deck's own painter, and it is now the **only** carrier of a change's counts. `Patch` holds the rows and nothing else, because a producer can measure a change it cannot render — hanging the counts off the patch is what threw the measurement away.

`FileDiff::build` resolves rows and counts independently. Comparing two sides is itself a measurement, so a change that arrived with content but no count is measured there; a change with neither a count nor a side to compare stays **unmeasured**. Both renderers leave an unmeasured side out rather than printing a zero, and a modification nobody measured gets no size column at all. A creation and a deletion keep their word — the tool named the act even when nothing measured its size.

## Two fabrications this found, both live on `main`

**1. `stella run`'s own scrollback drew `+0 −0` over a real change.** `AgentEvent::FileChange` measures the tree and renders the patch separately, so `diff` is `None` whenever one could not be read (a binary file, a `git diff` that failed — `JournalChange::diff` is `Option<String>`, "when one was readable"). `RunBuilder::attach_file` built the `Patch` from `diff.map(…)`, so a diffless event produced no patch and the event's `added`/`removed` went nowhere. `Call::line_delta` then folded `FileDiff::build` over a `FileChange` with no patch and two empty sides and got `(0, 0)`.

**2. `stella observe` drew `−0` on a replayed deletion.** `transcript_view::files_from_input`'s `delete_file` fallback carries neither side of the file, so rendering it ran `unified_diff("", "")` and read the zero back as the answer.

### `a_diffless_file_change_keeps_the_counts_the_event_measured` (`stella-tui`)

Ablated by putting `extent: Extent::default()` back where the event's counts now land:

```
--- ABLATED: extent no longer carries the event's counts ---
test transcript_build::tests::a_diffless_file_change_keeps_the_counts_the_event_measured ... FAILED

---- transcript_build::tests::a_diffless_file_change_keeps_the_counts_the_event_measured stdout ----
  left: Extent { added: None, removed: None }
 right: Extent { added: Some(12), removed: Some(3) }

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 1148 filtered out
```

### `a_change_nothing_measured_renders_no_size_at_all` + `an_unmeasured_modification_has_no_size_column` (`stella-transcript`)

Ablated by removing `FileDiff::build`'s "neither a count nor a side to compare" arm, so an unmeasurable change is measured as `(0, 0)` again:

```
--- ABLATED: an unmeasurable change is measured as (0, 0) again ---
---- tests::file_kinds::a_change_nothing_measured_renders_no_size_at_all stdout ----
assertion `left == right` failed: comparing nothing to nothing measures nothing
  left: Extent { added: Some(0), removed: Some(0) }
 right: Extent { added: None, removed: None }

---- tests::file_kinds::an_unmeasured_modification_has_no_size_column stdout ----
assertion failed: digest::step_digest(&step(call, 0), 40).delta.is_none()

test result: FAILED. 5 passed; 2 failed; 0 ignored; 0 measured; 122 filtered out
```

Each ablation makes the answer **wrong** (a fabricated zero where there is no measurement), not merely different. The first witness asserts the absence through `grid::render` **and** `html::render_run`, because the rule is the model's rather than either painter's.

Also added: `a_replayed_deletion_states_no_size_rather_than_a_zero` and `a_row_with_counts_and_no_diff_keeps_its_counts` in `stella-observatory`, against the fold rather than the rendering — `scripts/check-transcript-surfaces.py` reads a whole file including its test module, so rendering there would have made `transcript_view.rs` claim to be a declared transcript surface. The guard caught that during this change and the assertion moved to the renderer crate, where it belongs.

## Renamed test

`check-deleted-tests` will see one rename: `a_diffless_file_change_still_records_the_file` → **`a_diffless_file_change_keeps_the_counts_the_event_measured`** (`crates/stella-tui/src/transcript_build/tests.rs`). Same fixture, strengthened from "the file is recorded" to "its counts survive and the digest prints them" — the assertion whose absence let the defect ship.

## Second commit: one stale name

`#4822` renamed `v2_rows` to `projected_rows` when it dissolved the `v2` namespace. `render/entry/tool.rs`'s `result_body` doc kept the old name, which was the only occurrence of it left in the tree — a reader following the citation found nothing. Repaired here rather than filed, per CLAUDE.md: a stale comment is a bug, fixed on sight.

## Prose

The charter's "one hole has to close first" paragraph is rewritten to the state it now describes, and its dangling citation to the **closed** #4756 is repointed to #4289 — the repair #4289's own comment asks for (`scripts/check-transcript-surfaces.py` already cites `issue=4289`, which is why that is the surviving number).

## Exemplar

`Extent`'s two-independent-`Option`s shape is `stella-tui`'s `views::transcript::Extent`, ported rather than reinvented so the two surfaces cannot drift on the rule. The "resolve rows and counts independently" split follows `git`'s own separation of `--numstat` from `-p`, which is what `AgentEvent::FileChange` already models on the wire.

## Verification

Targeted, per CLAUDE.md — CI is the evidence. Locally: `cargo test -p stella-transcript -p stella-tui -p stella-observatory` and `cargo test -p stella-cli --bin stella` green, `cargo clippy -p stella-transcript -p stella-tui -p stella-observatory --all-targets -- -D warnings` clean, `cargo fmt --all --check` clean, `make guards-fast` exit 0.

## Filed, not fixed here

Three findings from this work, each written as a handoff:

- **#5080** — `html::STYLE` is the one web surface that never took #2597's square rule, and its ten radii are not tokenised, so the remedy `the_dashboard_is_monospace_and_square`'s doc comment names ("change `--radius` in the token block") does not exist for that file. It is also what blocks #4270 today.
- **#5081** — `grid::render` ships to every `stella run` user and has no golden. #4289 piece 3 replaces its whole frame; a golden landed first makes that diff readable.
- **#5083** — `scripts/check-transcript-surfaces.py` reads a call from a file's own `#[cfg(test)]` module as adoption, wrongly in both directions. This is the guard that fired during this change.

Refs #4289, #3578, #2290, #4156


macanderson:
## Answering Sourcery's two ❌ rows: both are deliberately out of scope, and #4289 says so itself.

**Row 2 — "Add Gold and Silver roles to `grid::Color`"** and **Row 3 — "Replace grid's box with the SPEC 6 frame and migrate interactive mode"** are #4289's pieces 2 and 3–4. That issue opens with a section headed *"Why this is not one PR"* and orders the pieces because **piece 1 is a correctness precondition for piece 3**:

> Migrating interactive mode onto `grid` **before** this lands would reintroduce the zero, silently, which is the one outcome that makes the whole migration a regression.

So this PR is piece 1, on its own, on purpose — and it is not merely a preparatory step: it fixes two fabricated zeros already shipping on `main`, one on `stella run`'s scrollback and one on `stella observe`. The PR title, its opening line and its body all state that pieces 2–4 remain and that it does not close #4289. It uses `Refs`, never `Closes`, for exactly this reason.

Two further reasons the remaining pieces should not ride along:

- **Piece 2 has a dependency this PR cannot satisfy.** #4289 requires `grid::Color::Cyan`'s fate to be decided *in the same PR as #4125*, not separately, "or the two will disagree". #4125 is not this change.
- **Piece 3 has no golden to change.** `crates/stella-transcript/` has one integration fixture (word highlighting) and no snapshot of `grid::render` at all, so replacing the frame today would move every line of the plain surface's output with nothing to show a reviewer what moved. Filed as #5081, and it should land before piece 3 rather than after.

Row 1 (the unmeasured-row concept, counts propagated independently of patch text, no fabricated zeros) is what this PR delivers, and Sourcery marks it ✅.

Both remaining rows stay tracked on #4289, where I have posted a comment repairing that issue's stale file list — `src/v2/` was dissolved by #4822, `v2_rows` is now `projected_rows`, and `crates/stella-tui/src/diff.rs` sits 38 lines under the flat ceiling, which is the binding constraint on piece 4 and is not recorded in the issue body.

Noting for the next reviewer: Sourcery's **code** review did not run on this PR — the account's 7-day diff budget is spent, so only the Reviewer's Guide and this assessment table posted. This PR has had no line-level automated review.

Issue #5039, opened 2026-08-26
No event carries its task id, so a task has no evidence ledger and no cost


macanderson:
## Problem

SPEC 7.1: a task is a contract, an **evidence ledger** (every event tagged with its task id), and a **cost** (`$ · tok · cache rd% · model calls · est remain` — the det% clause was deliberately dropped, see `task_contract.rs`'s module doc). The contract half is built (`stella_protocol::task_contract` is excellent). The other two halves are missing: no event anywhere carries a task id, so nothing can list a task's edits/runs/graph writes, and `TaskItem` has no cost — cost exists only per-session/per-turn (`receipts.rs`, `budget.rs`). Task tagging is what makes per-task cost attribution and evidence ledgers free (SPEC 6.2), and the transcript's `→ task 3` tag (spine issue in the transcript epic) needs the same field.

## Files

- `crates/stella-protocol` — `task_id: Option<TaskId>` on the event types that represent work (respect invariant 10: declare the consumer)
- `crates/stella-core` — stamp the running task's id on events it dispatches
- `crates/stella-store` — persist and query by task id (evidence ledger; per-task cost rollup)
- `crates/stella-tui` — surface via the plan panel and task zoom (their own issues render it)

## Done means

- Events emitted while task 3 runs carry task 3's id; `Store` can answer "task 3's events" and "task 3's cost" with a test over a scripted session.
- The consumers ledger (`event/consumers.rs`) rows for any new/changed tags are updated in the same PR (build error otherwise).
- Wire schema regenerated (`make wire-schema`).

Spec: `design/tui-v2/SPEC.md` §7.1, §6.2. Related: #4554 (event ledger and interactive mode), #4339.

**Estimated agent build time:** ~1–2 days
**Model:** use-model:ultra — Fable 5 (`claude-fable-5`)

Issue #5034, opened 2026-08-26
Write and delete events never show their graph facts


macanderson:
## Problem

Two SPEC 6.3 graph facts never render (rendering `02-event-vocabulary`):

- write footer: `registered in graph as module node` — the string exists nowhere in the tree
- delete body: `graph check: 0 inbound refs · det`, produced by a graph check that runs **before** the delete tool executes — also absent

The delete half is a product behaviour, not a label: the graph is consulted pre-execution so a deletion with inbound references can warn before the file is gone.

## Files

- `crates/stella-tools` delete/write tool path — run the inbound-refs query before executing delete; report the new-module registration after write
- `crates/stella-graph` — the inbound-refs query (likely exists as degree/relations)
- `crates/stella-tui/src/views/transcript_source.rs` — attach footer/body via the spine issue's `footer` field

## Done means

A write shows the registered-in-graph footer; a delete shows the inbound-refs line sourced from a check that demonstrably ran before the tool (test asserts ordering). When no graph index exists, both elide rather than fake a result.

Blocked by the transcript spine issue (see epic).
Spec: `design/tui-v2/SPEC.md` §6.3.

**Estimated agent build time:** ~8h
**Model:** use-model:pro — Opus 5 (`claude-opus-5`)

## Definition of done

- [x] Implementation complete
- [x] Scoped tests added/updated and passing
- [x] Full CI green
- [x] Code comments and docs updated
- [x] Residue filed as new issues (triage label only)

Added in the SCR-003 shape this issue predated; the "Done means" section above
is the same contract in prose. Tracked by #5245; CI green on run 33036334441 (`fmt + clippy + test`, `cargo deny`, MSRV, docs guards, file-size ratchet, guard self-tests, Windows split — every check except `dod` itself, which is what this box gates).

macanderson:
Scope note from the spine (#5030 / PR #5084): add this event's TranscriptEntry variant, its kind_for arm, and its SessionModel fold arm HERE, together with the upstream emit — the spine deliberately did not pre-add unproducible variants (#4180's lesson). head_rows now takes a CallFacts struct; Event.footer and Event.duration_ms are plumbed and rendered, so this issue's rows get timing and footers for free.

macanderson:
Claiming this — implementing now in feat/5034-graph-facts.

macanderson:
Implemented in #5245.

Issue #5031, opened 2026-08-26
Show a skill event in the transcript when a skill is injected


macanderson:
## Problem

SPEC 6.3 defines the skill event — `✦ skill <name> · auto|/cmd · n tok` with the body `injected <summary> · used n× this repo` (rendering `01-session-turn-lifecycle`) — and `EventKind::Skill{trigger, tokens}` renders it. Nothing live produces one: there is no `TranscriptEntry::Skill`, and skill injection surfaces only as a count inside `SteeringWithheld`. A user cannot see which skill fired, how it was triggered, or what it cost.

## Files

- The skill-selection/injection path in `crates/stella-core` (skills module) — emit the event with trigger kind and token count
- `crates/stella-cli` transcript feed — carry it into `TranscriptEntry::Skill` (variant added by the spine issue)

## Done means

A session that injects a skill shows the `✦ skill` head with trigger and token count, and the injected-summary body, proven by a live-adapter test. Usage count (`used n× this repo`) may land as a follow-up if the store's per-skill counter (#4337) is not yet available — say so in the PR if elided.

Blocked by the transcript spine issue (see epic).
Spec: `design/tui-v2/SPEC.md` §6.3. Related: #4337.

**Estimated agent build time:** ~4h
**Model:** use-model:balanced — Sonnet 5 (`claude-sonnet-5`)

## Definition of done

- [x] Implementation complete
- [x] Scoped tests added/updated and passing
- [x] Full CI green
- [x] Code comments and docs updated
- [x] Residue filed as new issues (triage label only)

macanderson:
Scope note from the spine (#5030 / PR #5084): add this event's TranscriptEntry variant, its kind_for arm, and its SessionModel fold arm HERE, together with the upstream emit — the spine deliberately did not pre-add unproducible variants (#4180's lesson). head_rows now takes a CallFacts struct; Event.footer and Event.duration_ms are plumbed and rendered, so this issue's rows get timing and footers for free.

macanderson:
Claiming this — implementing now in feat/5031-skill-event.

macanderson:
DoD checklist appended (the issue predates SCR-003's task template, so the `dod` gate had nothing to read).

Evidence for each ticked item, on PR #5247:

[...]

Issue #5030, opened 2026-08-26
The transcript can draw skill, memory, gate, and model events, but nothing live ever sends one


macanderson:
## Problem

The v2 transcript can already draw skill, memory, gate, and model events — `EventKind::{Skill, Memory, Gate, Model}` in `crates/stella-tui/src/views/transcript.rs` render correctly and are fully tested — but they are reachable only from `views/transcript/tests.rs`. `crates/stella-tui/src/model/entry.rs`'s `TranscriptEntry` has no variants for them, so `transcript_source::kind_for` can map only the five tool names plus `Other`. Three live `Event` fields are also never set by `transcript_source::head_rows`: `task` (the `→ task 3` tag, SPEC 6.2), `duration_ms` (the `⚡ms` metric), and `footer`. The same function also forces `collapsed = Some(false)`, so a live read head never draws `▸` or `↵ open` even though `EventKind::collapses_by_default()` says reads fold (SPEC 6.3).

This issue is the shared spine: the entry variants, the `kind_for` mapping, the field plumbing, and the collapsed default. The per-event upstream producers (skill, memory, model, graph facts) are separate issues that depend on this one and touch disjoint emit sites.

## Files

- `crates/stella-tui/src/model/entry.rs` — add `TranscriptEntry` variants (Skill, MemoryLog, MemoryPromote, Gate, Model)
- `crates/stella-tui/src/views/transcript_source.rs` — map them in `kind_for`; set `task`, `duration_ms`, `footer`; respect `collapses_by_default()`
- `crates/stella-tui/src/views/transcript.rs` — no rendering changes expected; its tests are the target the live path must reach

## Done means

- Each new `TranscriptEntry` variant renders through the same `event_rows` path the fixture tests exercise, proven by a live-adapter test per variant (entry in → styled rows out).
- A live read head renders `▸` and `↵ open` when folded; a witness test fails on today's `Some(false)` override.
- `→ task 3` and `⚡ms` appear on live heads when the entry carries them.

Spec: `design/tui-v2/SPEC.md` §6.2, §6.3. Related: #4721 (the v1/v2 renderer seam this feeds into).

**Estimated agent build time:** ~6–8h
**Model:** use-model:pro — Opus 5 (`claude-opus-5`)

## Definition of done

The “Done means” prose above, as boxes the gate can read — nothing here is
new work, and nothing is ticked by the migration (#5193).

[...]

PR #5084, opened 2026-08-26
feat(stella-tui): live tool heads carry their wall time, and reads fold by default

macanderson:
Two of #5030's four gaps, the two with live producers today:

- **`⚡ms` on every settled head.** `call_duration` resolves the paired result's own `duration_ms` through the same bounded scan as `read_size` — zero maps to no metric (synthetic echoes stamp zero, and 'answered instantly' vs 'not timed' is not worth an ambiguous column). Every settled call in the deck now states its wall time on the head, per SPEC 6.3.
- **Reads fold by default.** `head_rows` forced `collapsed = Some(false)` on every live head, so `EventKind::collapses_by_default()` was consulted nowhere live and `▸ read … · ↵ open` was reachable only from fixtures. A read head now folds until the reader expands it; every other kind keeps its glyph (its body is the result row beneath).

The per-call facts travel as one `CallFacts` struct (the `EntryView` pattern) instead of a 9-positional list.

**Scope adjustment against #5030's text, stated rather than slipped:** the issue asked this PR to also add `TranscriptEntry` variants for Skill/Memory/Gate/Model with no fold constructor behind them. That is the exact expressible-but-unreachable shape #4180 removed from this same module, so each variant lands with its producer instead — noted on #5031–#5034, which are unblocked regardless (their upstream emit sites are disjoint). The `→ task N` tag stays with #5039, its only real producer.

**Witnesses:** `a_settled_call_head_carries_its_wall_time` (renders `⚡7ms`, elides unanswered/zero/mismatched ids) and `a_read_head_folds_by_default_and_opens_on_expand` — both fail on the old `Some(false)`/no-duration path. Deck goldens re-blessed and the diff read: the only changes are `⚡ Nms` appearing right-aligned on settled heads (18 files); the scenario's read is `mcp__fs__read_file` (open-vocabulary `Other`), so no golden fold shift. Full `cargo test -p stella-tui` (12 suites) and clippy `--all-targets` green.

The head's `⚡ms` sits above the v1 result row's own `Nms` — one fact on two rows until the result row is ported (#4721 owns that seam).

Refs #5030

## Summary by Sourcery

Show wall-time metrics on settled live tool heads and restore default folding for read heads.

New Features:
- Display settled tool-call wall time on live transcript heads when a nonzero duration is available.
- Fold live read heads by default and allow them to open when expanded.

Enhancements:
- Bundle per-call rendering data into a named CallFacts structure for safer head rendering inputs.

Tests:
- Add coverage for duration resolution, zero or mismatched call handling, and live read-head folding behavior.
- Refresh deck snapshots to include duration metrics on settled heads.

### DIFF
--- a/crates/stella-tui/src/render/entry.rs
+++ b/crates/stella-tui/src/render/entry.rs
@@ -216,15 +216,17 @@ fn projected_rows(
             path,
             sub_agent_id,
         } => {
-            let scope = source::measured_scope(call_id, view.following, view.files);
-            let read = source::read_size(call_id, view.following);
             out.extend(source::head_rows(
                 name,
                 path.as_deref(),
                 input,
-                scope,
-                read,
-                sub_agent_id.as_deref(),
+                source::CallFacts {
+                    scope: source::measured_scope(call_id, view.following, view.files),
+                    read: source::read_size(call_id, view.following),
+                    duration_ms: source::call_duration(call_id, view.following),
+                    sub_agent_id: sub_agent_id.clone(),
+                    expanded,
+                },
                 width,
             ));
             if expanded {
--- a/crates/stella-tui/src/views/transcript_source.rs
+++ b/crates/stella-tui/src/views/transcript_source.rs
@@ -34,38 +34,60 @@ use super::transcript::{
 };
 use crate::model::{FileState, ReadSize, TranscriptEntry};
 
-/// The metal-bearing head of a dispatched call (SPEC 6.2).
+/// What is known about one call at the moment its head renders — each field
+/// filled by its own resolver, or its `None`/default while nothing has
+/// answered.
 ///
-/// `scope` is what the emitter measured for this call — the paths it claimed
-/// and the delta summed across them — or `None` while it is still in flight.
-/// [`measured_scope`] is what resolves it, and a head row is never asked to
-/// guess. `read` is the line coverage the tool reported for itself, resolved by
-/// [`read_size`] — a separate channel, because a read's number is a coverage
-/// the producer states, never a delta a mutation stamps.
+/// Bundled the way [`crate::render::EntryView`] bundles the draw-side pair:
+/// the facts travel together into every head, and a positional list this long
+/// is a call site that can pair one call's scope with another call's timing
+/// by ordering its arguments wrongly.
+#[derive(Default)]
+pub struct CallFacts {
+    /// What the emitter measured for this call — the paths it claimed and the
+    /// delta summed across them ([`measured_scope`]).
+    pub scope: Option<Touched>,
+    /// The line coverage a read reported for itself ([`read_size`]) — a
+    /// separate channel, because a read's number is a coverage the producer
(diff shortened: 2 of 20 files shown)


macanderson:
The `fmt + clippy + test` red here is `cargo doc`, not a test:

```
error: public documentation for `CallFacts` links to private item `crate::render::EntryView`
  --> crates/stella-tui/src/views/transcript_source.rs:41:23
   |
41 | /// Bundled the way [`crate::render::EntryView`] bundles the draw-side pair:
   |                     ^^^^^^^^^^^^^^^^^^^^^^^^^ this item is private
   = note: this link resolves only because you passed `--document-private-items`,
           but will break without
```

Worth flagging because **the local command that would catch this is not the obvious one**. `cargo doc --document-private-items` *resolves* the link and reports nothing; only the plain form fails, and the plain form is what CI runs. So a local check with `--document-private-items` gives a false green here.

The reliable sequence, since cargo also replays a cached green:

```sh
cargo clean -p stella-tui
RUSTDOCFLAGS="-D warnings" cargo doc -p stella-tui --no-deps
```

Two ways out, depending on what you meant:

- **The reference is for maintainers of this module** — drop the link and name the type in prose (`` `EntryView` ``), so nothing is claimed about a public path.
- **The pairing is genuinely part of the public story** — make `EntryView` public, or move the sentence onto a private item where the link is legal.

Same family as #4986 and the `[`fn@reference`]` ambiguity earlier today: rustdoc's diagnostics are precise about the symbol and silent about which invocation would have caught it.

macanderson:
Correcting my earlier comment on this PR — I got the diagnosis of *which command catches it* wrong, and the wrong version is the one that would propagate.

I claimed `--document-private-items` resolves the link and reports nothing, so a local check with it gives a false green. **That is not true.** Reproduced on `main` with a deliberately reinstated bad link:

| Command | Exit |
|---|---|
| `cargo doc -p stella-store --no-deps` | **0** — warning printed, not fatal |
| `RUSTDOCFLAGS="-D warnings" … --no-deps` | 101 |
| `RUSTDOCFLAGS="-D warnings" … --no-deps --document-private-items` | 101 |

`--document-private-items` does **not** mask it. It emits the same error plus a note aimed straight at this mistake: *"this link resolves only because you passed `--document-private-items`, but will break without."*

The distinguishing variable is **`RUSTDOCFLAGS="-D warnings"`**. Without it the warning is printed and the exit code is 0, so a local run looks clean while CI — which sets it — fails. That is the false green, and it is a plain-`cargo doc` false green, not a private-items one.

The `cargo clean` half of my advice stands and is load-bearing: cargo replays a cached green, so without it you can get exit 0 from a build that never re-ran.

So the reliable local check is:

```sh
cargo clean -p <crate>
RUSTDOCFLAGS="-D warnings" cargo doc -p <crate> --no-deps
```

Sorry for the noise. The fix in the comment above (unlink, or make the item public) was right; the reasoning about which invocation would have caught it was not.

Issue #2570, opened 2026-08-09
store: a cancelled execution reports only the pre-cancel cost prefix — exec 99 recorded 0.87 USD of a 5.29 USD run


macanderson:
## Problem

When a run is cancelled, the `executions` row keeps only the costs of roles that finished before the last stage started. The roles that ran longest — worker and witness author — are dropped from the aggregate entirely.

## Evidence

Execution 99 (`~/.stella/exports/session-2026-08-09-21-09-04.zip`, `raw/executions.json` + `raw/telemetry.json`), verified with exact decimal arithmetic:

```
executions.cost_usd              = 0.8659773
triage + research + plan         = 0.8659773   ← EXACT match
sum of all 116 role rows         = 5.28786465
dropped (worker + witness_author)= 4.42188735
```

The reported figure is **16.4% of the true spend**, and it is not a rounding artifact or a partial write: it is precisely the prefix of roles that completed before the worker began. All 116 telemetry rows carry `usage_complete = 1`, so the per-call ledger is intact — only the rollup on `executions` is wrong.

Wall clock for reference: started 20:43:16, finished 21:09:04 (cancelled). `outcome = "cancelled"`, `usage_complete = false`.

## Why it matters more than the number

1. **It biases in one direction.** The dropped roles are always the expensive, long-running ones, because those are the ones in flight when a cancel lands. So cancelled runs systematically under-report, and they under-report *most* on the runs that cost most.
2. **It makes the fix unmeasurable.** Any before/after comparison on verification cost — which is the whole point of #2559, #2560, #2562 and the responsibility-ablation controls #2374 needs — reads its numbers from here. A measurement layer that discards the arm under test cannot answer the question.
3. **It hides delivered work.** Execution 99 is recorded as `cancelled`, but the worker had already pushed and merged PR #2554 by the time the cancel landed. The record shows a cancelled run that cost $0.87; what happened was a $5.29 run that shipped a merged PR.

## What to do

- **Persist each role's cost as that role completes**, rather than rolling up at execution end. The per-call rows already land correctly and completely, so this is a rollup ordering problem, not a capture problem.
- On cancel, reconcile `executions.cost_usd` from the `telemetry` rows already written for that `execution_id` rather than from whatever the in-memory aggregate held.
- Record delivery separately from outcome. A run that pushed a commit or opened a PR before being cancelled did deliver; `outcome = "cancelled"` with no other field is not a complete description of what happened.
[...]

PR #5093, opened 2026-08-26
feat(stella-store): record whether a run shipped, apart from how it ended

### LINKED ISSUE #2808: Track whether a run shipped something, separately from how it ended — a cancelled run that merged a pull request looks like it delivered nothing

macanderson:
## What

`executions.outcome` was one final label carrying two independent facts:

- **how a run ended** — completed / aborted / cancelled / failed / error
- **whether it delivered anything**

They are independent in both directions. Execution 99 is the case #2808 opens with: recorded `outcome = 'cancelled'`, and what actually happened is that the worker pushed and merged pull request #2554 before the cancel landed. `resolved` in `usage_stats`, the scoreboard, and every cost-per-resolved receipt derived from them score that $5.29 delivered run as zero.

## Schema v38 — `executions.delivery`

Nullable, no default, no backfill, column-guarded `ADD COLUMN`, following v36's `parent_execution_id` shape. **The absent state is one of three answers, not a gap:**

| Reading | Means |
|---|---|
| `NULL` | Nothing observed this run's delivery. |
| `'none'` | Something looked, and the run shipped nothing. |
| `'commits'` | The run landed at least one commit. |

Collapsing the first two rebuilds the defect the column exists to fix — an absence of evidence read as a record of failure. There is no backfill because nothing recorded the fact: `pull_requests` is keyed by URL and linked to a *session*, never an execution, and the fleet's commit ledger lives in a different database (`fleet.db`) with no execution id on it. Picking whichever run of a session was open when a PR appeared would be a guess dressed as a record.

## The writer

The **fleet attempt close**, and only it — the one door that owes no attribution guess, because the execution row and the commit set belong to the same attempt, measured against its own `start_sha`. `fleet_commits::for_attempt` now runs *before* `finalize_fleet_execution` rather than after, so the row records what the attempt shipped as well as how it ended. `for_attempt` drains the observer, so it stays called exactly once.

A worker whose thread died leaves `NULL`: `fleet_spend::unreported_outcome` closes the row it could not close itself and claims nothing about delivery, which is the honest reading.

Every other door — `stella run`, the deck, self-driving — is untouched and reads `NULL`. Wiring them needs an attribution decision this PR does not make; filed as #5094.

## The reader

The analytics export's `executions` spine, where the fact leaves the store as `null` / `"none"` / `"commits"`.

## The decision #2808 asks for, made explicitly

**`usage_stats.resolved` still means `outcome = 'completed'` and nothing else.** Folding a delivered-but-cancelled run into it would move every published cost-per-resolved number on the strength of a signal only one door writes today. The two stay separable — read `delivery` for what shipped, `outcome` for how it ended — and promoting delivery into that count is a decision about published figures that belongs to whoever owns them. Stated in `usage_stats`' own doc comment so the next reader finds it there.

## Invariant 3 — telemetry egress

Nothing here reaches the hub or a drain. `HUB_TELEMETRY_COLUMNS` covers the hub `telemetry` table, which this does not touch; `ExecutionRollupRow` is unchanged, so the exhaustive-struct-literal guard in `content_free.rs` is not engaged and no allowlist is widened. That is deliberate scope, not an oversight: whether a delivery token should ride the enterprise operational rollup is a privacy review for a maintainer, and no consumer needs it yet.

## God files

`stella-store`'s `lib.rs` is at 1711 against a 1712 ceiling. It gains **no line** — `Delivery` is added to the existing `pub use telemetry::{…}`. `Store::record_delivery` / `Store::delivery` live in the new `telemetry/delivery.rs`; the migration body lives in the new `migrations/delivered_runs.rs`. `tests.rs` gains three comment lines in the one test that pins `SCHEMA_VERSION`, which is where that ledger lives.

## Witnesses, ablated

**1. The absent state is not "shipped nothing"** — `migrations::delivered_runs::a_run_from_before_the_column_reads_as_unobserved`. Ablated by giving the column `NOT NULL DEFAULT 'none'`:

```
---- migrations::delivered_runs::tests::a_run_from_before_the_column_reads_as_unobserved stdout ----
  left: Some("none")
 right: None

test result: FAILED. 2 passed; 1 failed; 0 ignored; 0 measured; 432 filtered out
```

**2. The fleet door records what it shipped** — `fleet_cmd::tests::a_stopped_attempt_that_landed_commits_is_not_a_stopped_attempt_that_landed_none`. Two stopped attempts, identical in outcome and cost, differing only in what they landed. Ablated by making `delivery_of` ignore its commits:

```
---- fleet_cmd::tests::a_stopped_attempt_that_landed_commits_is_not_a_stopped_attempt_that_landed_none stdout ----
assertion `left == right` failed: a cancelled attempt that landed a commit must record that it shipped
  left: Some(Nothing)
 right: Some(Commits)

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 2332 filtered out
```

Both ablations make the answer **wrong**, not merely constant.

**3.** `telemetry::delivery::how_a_run_ended_and_whether_it_shipped_are_two_separate_readings` pins all three readings against one `outcome = 'cancelled'`, plus round-trip and no-such-row cases.

## Local runs

- `cargo test -p stella-store` — `435 passed; 0 failed`
- `cargo test -p stella-cli fleet` — `59 passed; 0 failed`
- `cargo test -p stella-observatory` — `142 passed; 0 failed` (schema conformance unaffected)
- `cargo clippy -p stella-store -p stella-cli --all-targets -- -D warnings` — clean
- `make guards-fast` — clean

CI is the evidence; the workspace build is not run here.

Closes #2808
Refs #2570, #2374, #5094

## Summary by Sourcery

Record fleet execution delivery separately from termination outcome so analytics can distinguish what a run shipped from how it ended.

New Features:
- Record whether fleet executions shipped commits independently from how they ended, distinguishing unobserved delivery from observed empty delivery.
- Expose execution delivery status through the analytics export as null, none, or commits.

Bug Fixes:
- Prevent delivered-but-cancelled fleet runs from being treated as having delivered nothing in downstream analysis.

Enhancements:
- Keep delivery observations separate from outcome-based resolved usage statistics.
- Add store APIs and closed delivery tokens for recording and reading execution delivery status.

Tests:
- Add migration, delivery-state, and fleet closeout coverage for shipped, empty, and unobserved executions.

Chores:
- Bump the store schema to v38 with a nullable executions.delivery column and no backfill.

### DIFF
--- a/crates/stella-cli/src/fleet_cmd.rs
+++ b/crates/stella-cli/src/fleet_cmd.rs
@@ -1266,6 +1266,12 @@ async fn run_task(
     }
 
     let spent = budget.session_spent_usd();
+    // Read the attempt's commits BEFORE closing its execution row, so the row
+    // records what the attempt shipped as well as how it ended (#2808). This
+    // door is the one that owes no attribution guess: the execution and the
+    // commit set belong to the same attempt, measured against its own
+    // `start_sha`. `for_attempt` drains the observer, so it stays called once.
+    let commits = crate::fleet_commits::for_attempt(&committed, root, &start_sha, task).await;
     let _ = finalize_fleet_execution(
         &execution,
         &registry,
@@ -1273,8 +1279,8 @@ async fn run_task(
         spent,
         rendered.persistence_complete,
         force_incomplete,
+        crate::fleet_commits::delivery_of(&commits),
     );
-    let commits = crate::fleet_commits::for_attempt(&committed, root, &start_sha, task).await;
     Ok(WorkerOutcome {
         cost_usd: spent,
         commits,
@@ -1283,6 +1289,11 @@ async fn run_task(
     })
 }
 
+/// Close the attempt's execution row: how it ended, and what it shipped.
+///
+/// `delivery` is written whatever the outcome, because the two are independent
+/// — an attempt can error or be stopped after landing commits, which is the
+/// case that used to be recorded as having delivered nothing (#2808).
 fn finalize_fleet_execution(
     execution: &crate::fleet_spend::ExecutionHandle,
     registry: &ToolRegistry,
@@ -1290,10 +1301,12 @@ fn finalize_fleet_execution(
     cost_usd: f64,
     persistence_complete: bool,
     force_incomplete: bool,
+    delivery: stella_store::Delivery,
 ) -> bool {
     let Some((store, execution_id)) = execution else {
         return false;
     };
+    let _ = store.record_delivery(*execution_id, delivery);
     agent::record_execution_end(
         store,
         *execution_id,
--- a/crates/stella-cli/src/fleet_cmd/tests.rs
+++ b/crates/stella-cli/src/fleet_cmd/tests.rs
@@ -193,6 +193,7 @@ async fn fleet_attempt_persists_usage_before_complete_closeout() {
         0.25,
         rendered.persistence_complete,
         false,
+        stella_store::Delivery::Nothing,
     ));
     assert_eq!(store.count("telemetry").unwrap(), 1);
     assert!(store.execution_usage_complete(id).unwrap());
@@ -215,6 +216,7 @@ fn stopped_fleet_attempt_never_becomes_exportable() {
         0.0,
         true,
         true,
+        stella_store::Delivery::Nothing,
     ));
     assert!(!store.execution_usage_complete(id).unwrap());
     // The flag is what withholds it from the enterprise export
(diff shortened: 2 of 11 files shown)


macanderson:
Answering the one ❌ in Sourcery's assessment table.

**Row 2 — "Record delivery for executions at the points where commits, pushes, or pull requests are observed."** The reading is correct: this PR wires one door, the fleet attempt close, and leaves `stella run`, the deck and self-driving untouched. That is deliberate, and the reason is in the PR body and in `telemetry/delivery.rs`'s module doc — repeating it here so it sits with the row.

The fleet attempt is the only door where the join needs no guess. Its execution row and its commit set belong to the same attempt, measured against that attempt's own `start_sha`. No other door has that pairing:

- The deck's PR monitor (`command_deck/driver_support.rs`'s `spawn_pr_monitor`) observes a pull request for the **session**. `Store::upsert_pull_request` is keyed `UNIQUE(url)` and carries `session_id`, never `execution_id`. A session runs many turns, so stamping delivery onto whichever one happened to be open when a PR appeared is a guess dressed as a record.
- The self-driving loop creates the PR itself (`self_driving_cmd/deliver.rs`'s `open`) but does not hold the execution id of the turn that produced the branch.
- `stella run` and the deck lead turn could observe their own workspace HEAD across a turn, which is a *different* fact from "opened a PR" and needs its own design — a baseline captured at execution start, and a rule for a sibling process's commit in a shared tree.

Wiring any of them today means choosing an attribution rule, and #2808's own migration argument refuses exactly that: NULL means "nothing observed this run's delivery", and writing a value the observer cannot actually attribute would put a wrong answer where an honest absence belongs. Choosing that rule silently is what CLAUDE.md's "'now' vs. 'right' is the maintainer's call" forbids.

So the remaining doors are filed as **#5094**, written as a handoff with the three candidate attribution rules, the file paths, and the constraints — rather than left as a gap this PR walked past.

Rows 1 and 3 are ✅ and need nothing.

Issue #5091, opened 2026-08-26
Step grading: the agent-judge fallback and the transcript join report are still unbuilt


macanderson:
## Problem

`bench/telemetry_store/grade.py` (#4751) ships phase 1 of
`doc:step-grading-and-productive-ratio` section 9 — the deterministic rule
chain, `grader_version = 'v1-deterministic'`. Phases 2 and 3 of that plan are
not built:

- **Section 6, the agent-assisted fallback.** A window no rule settles is
  written `direction = 'neutral'`, `direction_source = 'deterministic'`,
  `direction_reason = 'unclassified_floor'` — which is what phase 1 asks for,
  and it means every unclassified window currently counts toward the
  denominator as neutral. `direction_source = 'agent_judge'` and
  `'escalated_human'` are declared in the schema and written by nothing.
- **Section 8, the join report.** Pairing a grade with the transcript lines it
  judged, via `events.seq BETWEEN step_grades.event_seq_start AND
  step_grades.event_seq_end`. The columns exist and the index
  (`events_trial_idx`) covers the join; there is no reader.

## Why it is filed rather than fixed

The judge spends model tokens, so it needs a provider key and a decision about
which model grades (section 6's constraint is that it must never be the model
that authored the trial). Neither is available to a CI test, and a fallback
that cannot be tested deterministically should not be the thing that decides a
published ratio.

## What done looks like

**Section 6.** A `--grader-version v2-with-judge` path that:

- is offered only windows `v1-deterministic` wrote as `unclassified_floor`,
  never one a rule settled — assert this directly, as section 9 phase 2
  requires, rather than spot-checking it;
- sends the window's tool calls and results verbatim plus the two windows
  either side, and parses **only** the structured shape in section 6
  (`{"direction", "reason", "confidence"}`); a response that does not parse is
  a hard error for that window, never a silent `neutral`;
- writes `direction_source = 'agent_judge'` when `confidence` clears the floor
  (default 0.6) *and* does not contradict a strong deterministic signal on an
  adjacent window, and `'escalated_human'` otherwise.

**Section 8.** A report that walks a trial's `events` in `seq` order, looks up
the covering `step_grades` row per line, and renders direction in the gutter
with running time/cost totals, marking `escalated_human` rows distinctly.
Generatable as a static artifact and as a live query over the same join.

## Verify

`grade.py`'s `classify` is the seam for the judge — it returns
`(direction, reason)` and the caller stamps `direction_source`; a judge path
[...]

Issue #5089, opened 2026-08-26
Step grader cannot implement the verifier-delta rule: verifier_tests holds one snapshot per trial, so there is nothing to difference


macanderson:
## Problem

`doc:step-grading-and-productive-ratio` section 5 rule 2 is the verifier-delta
rule:

> For windows with a `verifier_tests` snapshot available before and after (CTRF
> re-run mid-trial, when the harness supports it): newly-passing test count
> increases -> `productive` (`verifier_delta:+N`); newly-failing increases ->
> `unproductive` (`verifier_delta:-N`); no change falls through.

`bench/telemetry_store/grade.py` (#4751) implements rules 1, 3, 4 and 5 and
**does not implement rule 2**, because the store holds nothing to difference.
`verifier_tests` in `bench/telemetry_store/schema.sql` is:

```sql
CREATE TABLE IF NOT EXISTS verifier_tests (
    id                TEXT PRIMARY KEY,
    trial_id          TEXT NOT NULL REFERENCES trials(trial_id) ON DELETE CASCADE,
    name              TEXT NOT NULL,
    status            TEXT,
    duration_ms       INTEGER,
    message           TEXT
);
```

No sequence, no timestamp, no window key. `ingest.py::ingest_trial` fills it
once per trial from a single `verifier/ctrf.json`, so there is exactly one
snapshot and a "before and after" does not exist. The spec's own "when the
harness supports it" is the tell: it does not, today.

## Why it matters

Rule 2 sits above rule 3 in the priority chain, and it is the only rule that
reads the benchmark's *own* grader rather than a derivation. Without it, a step
that turns a failing test green is graded by the diff-convergence substitute
instead of by the assertion that changed — a weaker signal for the exact case
the metric exists to measure.

## What done looks like

Both halves, in order:

1. **A harness that re-runs the verifier mid-trial and records each run.** This
   is upstream of this repository — the CTRF report is produced by the
   Terminal-Bench harness, and one report per trial is what it emits today.
   Decide whether a mid-trial re-run is affordable at all before building the
   store side; it is a full verifier pass per step-window.
2. **A schema change plus a spec edit in the same PR** (the rule
   `schema.sql`'s header states): `verifier_tests` gains a snapshot key —
   `(trial_id, snapshot_seq)` or an `event_seq` into `events` — and `ingest.py`
   fills it. Then `grade.py` gains `rule_verifier_delta`, slotted above
   `rule_diff_convergence` in `classify`, with a fixture test showing a
[...]

Issue #5088, opened 2026-08-26
ingest.py truncates payload_json mid-JSON, so a large event's payload cannot be parsed at all


macanderson:
## Problem

`bench/telemetry_store/ingest.py` writes every event payload as

```python
(f"{trial_id}:{seq}", trial_id, seq, t, json.dumps(e)[:20000])
```

`json.dumps(e)[:20000]` is a **character slice of a JSON document**, so any
event whose serialized form exceeds 20,000 characters is stored as a truncated
prefix that is not valid JSON. `json.loads` on that cell raises
`JSONDecodeError`. There is no marker on the row saying it happened, so a
reader cannot tell a truncated payload from a malformed one, and a reader that
catches the exception and moves on silently loses the event's whole payload.

The same shape applies to the two other truncations in `ingest_trial`, but
those are already safe because they slice a *string field* rather than a JSON
document (`(ei.get("exception_message") or "")[:2000]` and the CTRF message).
Only `payload_json` slices the document.

## Why it matters now

`bench/telemetry_store/grade.py` (added in the #4751 PR) reads `payload_json`
for every rule it runs, and the two event types whose payloads are largest are
exactly the two it most needs:

- `file_change` carries a `diff`, which is what the diff-convergence rule
  reads;
- `tool_result` carries the tool's whole output, which is what the tool-exit
  rule reads.

So the events most likely to be unreadable are the ones that decide a step's
direction. The grader counts unreadable payloads and reports them
(`grade.py::_payload`, `main`'s "event payload(s) did not parse" line) rather
than treating them as absent facts, but counting is a workaround, not a fix.

## Reproduce

```python
import json
e = {"type": "file_change", "path": "a.rs", "diff": "+" + "x" * 30000}
json.loads(json.dumps(e)[:20000])   # JSONDecodeError
```

## What done looks like

Pick one and land it with a test in `bench/telemetry_store/tests/`:

1. **Truncate the field, not the document.** Cap the large string members
   (`diff`, `output`, `text`) before `json.dumps`, so the stored cell is always
   valid JSON with a visibly shortened field, and mark the truncation in-band
   (e.g. a `"_truncated": ["diff"]` key) so a reader can tell a short diff from
   a cut one.
2. **Store the full payload.** Measure first — the drop-list in
   `DROP_EVENT_TYPES` already removed the ~99% that was streaming deltas, and
   the cap may be costing more than it saves now.

Either way the invariant to test is: **every `payload_json` cell round-trips
[...]

PR #5095, opened 2026-08-26
feat(bench): grade an ingested trial's steps, turns and execution

### LINKED ISSUE #4751: Step grading tables ship empty: the grader that fills step_grades / turn_grades / execution_grades was never built
## Problem

`bench/telemetry_store/schema.sql` ships three grading tables — `step_grades`,
`turn_grades`, `execution_grades` (schema.sql:158-290) — and nothing writes a
row to any of them. `rg 'step_grades' bench/` returns only `schema.sql`
itself; `bench/telemetry_store/` contains `ingest.py`, `conftest.py`,
`schema.sql` and `tests/`, and no grader module.

So the data model half of `doc:step-grading-and-productive-ratio` landed and
the pipeline half did not. The tables are queryable, always empty, and a
report that joins them silently returns nothing rather than failing — which is
the failure mode worth avoiding, because an empty `productive_step_ratio`
reads the same as a ratio nobody has computed yet.

Found during the `docs/` staleness pass for #4735, which moved the spec from
`docs/designs/` to `docs/spec/` and set its status to `living` with a
"built so far" note pointing here.

## What is missing

Sections §4–§9 of `doc:step-grading-and-productive-ratio`:

- **§4 Ingestion and join strategy** — walking a trial's `events` rows into
  step windows keyed `(trial_id, turn_instance, step, call_index)`, anchored on
  `StampedEvent.ts` the way `transcript.py` anchors `TranscriptState.origin_ms`.
- **§5 Deterministic scoring** — the rule-based classifiers that settle
  `direction` from artifacts already on disk (verifier deltas, loop detection,
  retry exhaustion), recording `direction_source = 'deterministic'` and
  `confidence = 1.0`.
- **§6 Agent-assisted grading fallback** — only for the residue no rule
  resolves, recording `direction_source = 'agent_judge'` with the judge's own
  confidence.
- **§7 Aggregation formulas** — the roll-ups into `turn_grades` and
  `execution_grades`, with the `NULL`-not-zero rule for `productive_step_ratio`
  when `step_count = 0`.
- **§8 Join report** — pairing a grade with the transcript lines it judged via
  `event_seq_start`/`event_seq_end`.
- **§9 Rollout and testing plan**.

## Constraints

- The spec is the contract: `schema.sql:158` cites it by id, and the DDL is
  already merged, so the grader implements the columns as written rather than
  re-deriving them. A column-shape change is a schema migration plus a spec
  edit in the same PR.
- Grading is replay-only ("replay, never perturb") — it must not steer a
  running agent and must not write back into `trials.reward`/`passed`.
- `grader_version` is the re-run key: an improved ruleset writes a new row set
  and never overwrites an old one. The unique indexes already enforce this.
- Re-reading a trial six months later must grade identically, so every
  timestamp is price-clocked from the trial's own first stamped event, never
  wall-clock read time.

## Definition of done

`bench/telemetry_store` has a grader that populates all three tables for an
ingested trial; a test ingests a fixture trial and asserts the deterministic
classifiers produce the expected `direction`/`direction_reason` rows; the
`productive_step_ratio` is `NULL` rather than `0` for a zero-step turn; and
the "Built so far" note at the top of `doc:step-grading-and-productive-ratio`
is updated to say what now runs.

macanderson:
`bench/telemetry_store/schema.sql` shipped three grading tables and nothing
wrote a row to any of them. `bench/telemetry_store/grade.py` fills them.

Closes #4751

## What runs

Phase 1 of `doc:step-grading-and-productive-ratio` §9, under
`grader_version = 'v1-deterministic'`:

- **§4 windowing and the price clock.** A window runs from the event after the
  previous `step_usage` through the `step_usage` that closes it, keyed
  `(trial_id, turn_instance, step, call_index)`. Every timestamp is
  milliseconds from the trial's own first stamped event, so a trial read today
  and one read six months from now grade identically; `graded_at` is the only
  wall-clock column.
- **§5's rule chain**, in priority order — see the gaps below for what is not
  in it.
- **§7 aggregation.** The execution ratio is summed over the turn rows rather
  than recomputed from `step_grades`, so §7's weighted-mean identity holds by
  construction. `NULL`, never `0`, when nothing was graded.
- **Cost.** Each step is priced off the same `normalized_cost` table
  `ingest.py` applies to a whole trial, carrying `ingest.py`'s own
  `cost_norm_status` vocabulary. A roll-up over any unpriced step is `NULL`
  with that step's reason, never a partial sum published as a total.

`grade.py --db bench.db [--run TAG] [--trial ID]`. Replay only: it never
touches `trials.reward` or `trials.passed`, and a test asserts that.

## What is not built, and why

Named in `grade.py`'s module docstring and in the spec's "Built so far" note,
each with an issue, rather than left as a silence:

- **§5 rule 2, verifier delta — #5089.** Not implementable against this store.
  `verifier_tests` is `(id, trial_id, name, status, duration_ms, message)`:
  one final CTRF report per trial, no window key, no sequence, no timestamp.
  There is no before and after to difference. It needs a harness that re-runs
  the verifier mid-trial plus a schema change.
- **§6 the agent judge and §8 the join report — #5091.** Phases 2 and 3. Until
  the judge exists, an unsettled window is written
  `neutral / deterministic / unclassified_floor`, which is what phase 1 asks
  for: "ruled neutral" and "no rule fired" stay tellable apart in the data. A
  test asserts no row claims `agent_judge`.
- **§5 rule 3 is implemented against a substitute** for "the trial's final
  accepted patch", which this store does not hold: a line is convergent when
  no later `file_change` on the same path undoes it. Stated in the docstring
  and in the spec note.
- **#5088** (filed by this work, not caused by it): `ingest.py` stores
  `json.dumps(event)[:20000]`, which slices a JSON document mid-string, so a
  large `file_change` or `tool_result` payload does not parse at all. The
  grader counts unreadable payloads and reports them rather than reading them
  as absent facts, and a test pins that. That is a workaround; #5088 is the
  fix.

## Witness

`bench/telemetry_store/tests/test_grade.py`. It fails on `main` at
import (`grade.py` does not exist there), so two ablations against the new
code, each making the answer **wrong** rather than constant.

**1. The NULL-not-zero rule** — `_ratio` returns `0.0` instead of `None` for a
zero-step turn:

```
        assert step_count == 0
>       assert ratio is None
E       assert 0.0 is None

bench/telemetry_store/tests/test_grade.py:333: AssertionError
FAILED ...::test_a_turn_that_graded_no_step_has_a_null_ratio_never_zero
1 failed, 1 passed, 31 deselected
```

**2. Rule 3 stops checking whether a line survived** — every added line counts
as convergent:

```
E       AssertionError: assert [(1, 1, 0, 'p...rgent:+1/-0')] == [(1, 1, 0, 'p...rgent:+1/-0')]
E         At index 1 diff: (1, 2, 0, 'productive', 'diff_convergent:+1/-0') != (1, 2, 0, 'unproductive', 'diff_divergent:+0/-1')
FAILED ...::test_an_edit_the_trial_keeps_is_productive_and_one_it_undoes_is_not
1 failed, 32 deselected
```

Restored, all green:

```
33 passed in 0.21s
```

## Coverage

`bench.yml`'s scope filter matches `^bench/` and its `telemetry_store — pytest`
step runs `bench/telemetry_store/tests`, so this diff runs the suite on CI.
`make guards-fast` is clean locally (`prose`, `left-behind`, `god-files`,
`gate-parity`, `doc-links` included); no baseline was raised and no file
approaches a size ceiling.

## Deleted tests

None.

## New dependencies

None — `grade.py` is standard library plus the price table `ingest.py` already
loads by path, so the suite still runs with `--no-project`.

## Summary by Sourcery

Implement phase-one deterministic grading for ingested trials without modifying benchmark verdicts.

New Features:
- Add deterministic replay grading for ingested trials, producing step, turn, and execution grade records.
- Classify step windows using loop/retry, diff convergence, tool-exit, and no-op rules with explicit neutral and unclassified outcomes.
- Calculate trial-relative timing, model costs, aggregation ratios, verdict alignment, and unpriceable-payload reporting.

Enhancements:
- Make grading reproducible through trial-relative timestamps, recorded pricing dates, versioned results, and replay-only database writes.
- Preserve missing measurements as NULL and prevent partial cost totals from being published as complete values.

Documentation:
- Document the phase-one grading implementation, its limitations, and the outstanding verifier, judge, and join-report work.

Tests:
- Add comprehensive coverage for windowing, rule priority, pricing, aggregation, reruns, verdict preservation, and truncated payload handling.

### DIFF
--- a/bench/telemetry_store/grade.py
+++ b/bench/telemetry_store/grade.py
@@ -1,0 +1,839 @@
+"""Grade an ingested trial's steps, turns and execution.
+
+Usage:
+    python3 grade.py --db bench.db [--run TAG] [--trial TRIAL_ID]
+        [--grader-version v1-deterministic]
+
+The second pass over a trial `ingest.py` has already loaded, implementing
+sections 4, 5, 7 and phase 1 of section 9 of
+doc:step-grading-and-productive-ratio. It reads `trials`/`events` and writes
+only `step_grades`/`turn_grades`/`execution_grades`; it never touches
+`trials.reward` or `trials.passed`, and it runs against artifacts already on
+disk rather than a live agent ("replay, never perturb").
+
+Determinism is the whole contract. Re-grading a trial six months from now must
+produce the same rows, so every timestamp here is milliseconds from the trial's
+own first stamped event and every price comes from the run date recorded on the
+trial, never from the clock this process reads. `graded_at` is the one
+wall-clock column, and it records when grading ran rather than anything about
+the trial.
+
+## What is implemented, and what is not
+
+Of the five rules in section 5, three are implemented as written, one against a
+substitute derivation, and one is not implementable from this store at all:
+
+* **Rule 1, loop/retry** — implemented as written.
+* **Rule 2, verifier delta** — NOT implemented, and not implementable here. It
+  needs a `verifier_tests` snapshot from before and after a window, and that
+  table has no window key, no sequence and no timestamp: its columns are
+  `(id, trial_id, name, status, duration_ms, message)`, one final CTRF report
+  per trial. There is nothing to difference. Tracked in #5089, which also
+  carries the harness change it depends on.
+* **Rule 3, diff convergence** — implemented against a *substitute* for "the
+  trial's final accepted patch", which is not an artifact this store holds. A
+  line a window adds counts as convergent when no later `file_change` on the
+  same path removes it, and divergent when one does; the reverse for a line it
+  removes. That answers the same question — did this edit survive? — from the
+  event stream rather than from a patch file.
+* **Rule 4, tool exit** — implemented. A nonzero shell exit is
+  `ToolOutput::Error` (witnessed by `nonzero_exit_is_error` in
+  `crates/stella-tools/src/bash.rs`), so the arm of a `tool_result` is the exit
+  signal and no prose is parsed to find it.
+* **Rule 5, no-op** — implemented, with one substitution stated in `_is_no_op`.
+
+Section 6's agent-assisted fallback is phase 2 and is absent: a window no rule
+settles is written `neutral` / `deterministic` / `unclassified_floor`, so
+"ruled neutral" and "no rule fired" stay tellable apart in the data instead of
+being folded together by a judge that does not exist yet. That fallback and
+section 8's transcript join report are #5091.
+"""
+
+from __future__ import annotations
+
+import argparse
+import json
+import sys
+
+from ingest import (
+    COST_NORM_NO_MODEL,
+    COST_NORM_NO_TOKENS,
+    COST_NORM_PRICED,
+    COST_NORM_UNPRICED_MODEL,
+    connect,
+    normalized_cost,
+    now,
+    parse_timestamp,
+)
+
+# Phase 1 of section 9: the rule chain with no judge behind it. The version is
+# the re-run key — an improved ruleset picks a new string and writes a new row
+# set beside this one rather than over it.
+GRADER_VERSION = "v1-deterministic"
+
+# Where the truncated-payload hole in `_payload` is being fixed. A constant so
+# the docstring's claim and the runtime report name the same issue.
+TRUNCATION_GAP = "#5088"
+
+PRODUCTIVE = "productive"
+UNPRODUCTIVE = "unproductive"
+NEUTRAL = "neutral"
+
+DETERMINISTIC = "deterministic"
+
+# The floor reason (section 9, phase 1). A window carrying this is one every
+# rule declined, which is a different fact from a window ruled
+# `no_measurable_effect` — the first is the residue phase 2's judge is for, the
+# second is already settled.
+UNCLASSIFIED_FLOOR = "unclassified_floor"
+NO_MEASURABLE_EFFECT = "no_measurable_effect"
+
+# Every deterministic rule is exact by construction, so its confidence is not a
+# self-report and not a tuning knob.
+DETERMINISTIC_CONFIDENCE = 1.0
+
+# The shell tool: the one tool that can change the workspace without emitting a
+# `file_change`, which is why `_is_no_op` names it rather than copying a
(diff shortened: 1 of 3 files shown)


macanderson:
Answering both `❌` rows in Sourcery's assessment table. Neither is a fix owed
by this PR; both are deferred with a filed handoff, and each is stated in the
module docstring and the spec's own "Built so far" note rather than left as a
silence.

**Row 1 — "including verifier-delta scoring" (§5 rule 2).** Not deferred by
choice: it cannot be implemented against this store. `verifier_tests` is
`(id, trial_id, name, status, duration_ms, message)` — no window key, no
sequence, no timestamp — and `ingest.py::ingest_trial` fills it once per trial
from a single `verifier/ctrf.json`. A rule that differences a before-snapshot
against an after-snapshot has nothing to difference. The spec's own wording
concedes this ("when the harness supports it"); it does not. Implementing it
needs a harness that re-runs the verifier mid-trial *and* a schema change, both
of which are the handoff in #5089.

**Row 2 — §6 agent judge and §8 join report.** Out of scope by the spec's own
rollout plan: §9 names these phases 2 and 3, and this PR is phase 1. Phase 1
explicitly asks for an unclassified window to be written
`neutral` / `deterministic` / `unclassified_floor` so the residue a judge would
grade stays visible in the data instead of being folded away by a judge that
does not exist — which is what this PR does, with a test asserting no row
claims `agent_judge`. The judge also spends model tokens and needs a provider
key, so it cannot be exercised by a deterministic CI test as written; #5091
carries the handoff, including the seam the tests fake.

Both issues are referenced from this PR's description, and the issue's own
"Definition of done" — a grader that populates all three tables, a test
asserting the deterministic classifiers' `direction`/`direction_reason`, a NULL
`productive_step_ratio` for a zero-step turn, and an updated "Built so far"
note — is met in full, which is the row Sourcery marked ✅.

Issue #5097, opened 2026-08-26
Nothing ships the emitted telemetry mirror to the benchmarks database over SSM


macanderson:
## Problem

`bench/telemetry_store/mirror.py` (#4614) emits the SQL that copies a local
telemetry store into the `benchmarks` database on `oxagen-data-postgres-1`
(instance `i-023d002d6e44f8f84`), plus a manifest of expected row counts and
the durable-side count queries to check them against. Nothing ships it.

Three functions in that module have no caller yet and this issue is what wires
them:

- `emit(conn, run_ids, source)` — the appliable transaction,
- `verification_sql(run_ids)` / `manifest(conn, run_ids, source)` — the two
  halves of the verify step,
- `mark_migrated(conn, run_ids, source)` — the local stamp, which must run
  **only after** the verify agrees.

#4614 named `arenabench/scripts/mirror-experiments.sh` as the template. It left
this repository with the ejection to `macanderson/arenabench` (#4642), so the
wrapper is being written rather than generalized; its shape survives only in
the other repo.

## What done looks like

A script (`bench/telemetry_store/ship-mirror.sh`, or a subcommand of an
existing one) that:

1. runs `mirror.py --db <local> --run <tag> --out mirror.sql --manifest
   mirror.json --verification verify.sql`;
2. gzips and base64-chunks `mirror.sql`, sends the chunks over
   `aws ssm send-command` (no open database port — that is the whole reason
   this is SQL text and not a connection), reassembles on the instance, and
   applies it inside the container with `psql -1`;
3. runs `verify.sql` on the durable side and compares its counts against
   `mirror.json`;
4. **only then** calls `mark_migrated` against the local database.

Steps 3 and 4 in that order are the constraint #4614 states and the one thing
this must not get wrong: marking local rows for a mirror that did not land is
how evidence gets deleted.

## Measure before assuming a chunk size

`mirror.py` prints the emitted byte count to stderr for exactly this. One
ingested 89-task run holds tens of thousands of `events` rows; the SSM
send-command payload limit is far below that, so the chunking is required, not
optional. Record the measured size for a real run in the script's header rather
than picking a chunk count by feel.

## Needs

AWS access to account 578673726240 with `ssm:SendCommand` on
`i-023d002d6e44f8f84`, and the `benchmarks` database credentials as they are
held on the instance. Neither is available to CI, so the wrapper is tested by
hand and the pytest suite stays on `mirror.py`'s emitted text — which is
[...]

PR #5099, opened 2026-08-26
feat(bench): emit the SQL that mirrors a telemetry store into the durable one

### LINKED ISSUE #4614: Nothing copies Terminal-Bench run databases into the benchmarks database yet

macanderson:
`runs.migrated` and `runs.migration_source` were scaffolding: nothing pushed a
local telemetry SQLite file into the `benchmarks` database. `mirror.py` emits
the SQL that does.

Closes #4614

## Shape

It emits an appliable transaction rather than opening a connection. The durable
store sits behind SSM on a private instance with no open database port, so a
text file is what fits through — and the same text applies to a scratch SQLite
database, which is how the idempotency test runs with no cloud at all.

```
mirror.py --db bench.db [--run TAG ...] [--out mirror.sql]
          [--manifest mirror.json] [--verification verify.sql]
```

All nine tables, parents first: `runs` -> `trials` -> `events`, `tool_calls`,
`verifier_tests`, `artifacts`, `step_grades`, `turn_grades`,
`execution_grades`. Column names come from `PRAGMA table_info`, so the emitter
follows the schema instead of carrying a second copy of it to drift.

Each rule the emitted SQL keeps has a test:

- **Portable on both engines.** No `AUTOINCREMENT`, no boolean literals, no
  `JSONB` — `schema.sql`'s own header, held by the rows as well as the DDL.
  Strings double their single quotes and nothing else, because PostgreSQL runs
  with `standard_conforming_strings` on.
- **Idempotent.** Every primary key is an ingester-supplied id, so
  `ON CONFLICT DO NOTHING` makes a second apply a no-op.
- **The durable copy is the record.** A run already there keeps its row; the
  provenance stamp is a trailing `UPDATE` for exactly that reason, so it lands
  on a row the INSERT declined.
- **Children after parents**, verified with foreign keys on.
- **One transaction**, so a half-applied mirror is not orphan rows.
- **Nothing is coerced.** An infinity or a NUL byte raises
  `UnmirrorableValue`. Writing NaN as NULL or dropping the NUL would change the
  data on the durable side while looking exactly like a clean mirror, and the
  durable copy is the one that outlives the machine it can be checked against.
- **Verify before mark.** `manifest` says what the durable side must hold;
  `verification_sql` asks it; `mark_migrated` stamps the local rows and is a
  separate function from `emit` so the ordering cannot be got wrong by
  accident. A test asserts `emit` leaves the local store untouched.

## Not in this PR

**The SSM shipping wrapper — #5097.** #4614 named
`arenabench/scripts/mirror-experiments.sh` as the template; it left this
repository with the ejection to `macanderson/arenabench` (#4642), so the
wrapper is being written rather than generalized, and it needs
`ssm:SendCommand` on `i-023d002d6e44f8f84` plus the instance's database
credentials — neither available to CI. #5097 carries the full handoff,
including the verify-then-mark ordering and the chunking, and it is what calls
`mark_migrated`.

On the chunking: #4614 asks for the size to be measured rather than assumed, so
`mirror.py` prints the emitted byte count to stderr. It does not pick a chunk
size, because nothing here has shipped a real run yet and a number chosen by
feel would read like a measurement.

## Witness

`bench/telemetry_store/tests/test_mirror.py`. It fails on `main` at import
(`mirror.py` does not exist there), so two ablations against the new code, each
making the answer **wrong** rather than raising.

**1. `INSERT OR REPLACE` instead of `ON CONFLICT DO NOTHING`** — the mirror
starts overwriting the durable side with the working set:

```
        notes, migrated, source = durable.execute(
            "SELECT notes, migrated, migration_source FROM runs"
        ).fetchone()
>       assert notes == "the durable one"
E       AssertionError: assert None == 'the durable one'

FAILED ...::test_a_run_already_on_the_durable_side_keeps_its_row_and_still_gets_stamped
1 failed, 15 deselected
```

**2. The provenance stamp folded into the INSERT** (the trailing `UPDATE`
removed) — a run the INSERT declined now records no source at all:

```
E       AssertionError: assert (0, None) == (1, 'Mac@local')
FAILED ...::test_the_durable_side_records_where_the_copy_came_from
FAILED ...::test_a_run_already_on_the_durable_side_keeps_its_row_and_still_gets_stamped
2 failed, 14 deselected
```

Restored, all green:

```
27 passed in 0.23s
```

## Coverage

`bench.yml`'s scope filter matches `^bench/` and its `telemetry_store — pytest`
step runs `bench/telemetry_store/tests`, so this diff runs the suite on CI.
`make guards-fast` is clean locally; no baseline was raised.

## Deleted tests

None.

## New dependencies

None — `mirror.py` is standard library plus `ingest.connect`, so the suite
still runs with `--no-project`.

## Summary by Sourcery

Emit an idempotent, verifiable SQL mirror of local telemetry stores for transfer to the durable benchmark database.

New Features:
- Add a CLI that emits portable SQL transactions for mirroring selected or all telemetry runs across the nine telemetry tables.
- Generate manifests and durable-side verification queries to support verify-before-mark migration workflows.

Enhancements:
- Make mirroring idempotent, preserve existing durable rows, record migration provenance, maintain parent-before-child ordering, and reject values that cannot be represented without data loss.
- Derive emitted column lists from the local schema and report the generated SQL size for future transport planning.

Tests:
- Add comprehensive SQLite integration coverage for idempotency, foreign-key ordering, run selection, verification, portability, escaping, unsupported values, schema coverage, and migration marking.

### DIFF
--- a/bench/telemetry_store/mirror.py
+++ b/bench/telemetry_store/mirror.py
@@ -1,0 +1,317 @@
+"""Emit the SQL that copies a local telemetry store into the durable one.
+
+Usage:
+    python3 mirror.py --db bench.db [--run TAG ...] [--out mirror.sql]
+        [--manifest mirror.json] [--source Mac@local]
+
+`schema.sql` says the local SQLite copy is the working set and the Postgres
+copy is the durable one, and `runs.migrated` / `runs.migration_source` record
+which is which. This is the half that moves the rows.
+
+It emits SQL rather than opening a connection, so the transport is somebody
+else's problem: the durable store sits behind SSM on a private instance with no
+open database port, and a text file is what fits through that. The same text
+applies to a scratch SQLite database, which is how the idempotency test runs
+with no cloud at all.
+
+## The rules the emitted SQL keeps
+
+* **Valid on SQLite and PostgreSQL**, per `schema.sql`'s own header: no
+  `AUTOINCREMENT`, no boolean literals, no `JSONB`. Only two syntaxes here are
+  not ancient — `ON CONFLICT DO NOTHING` (SQLite 3.24+, PostgreSQL 9.5+) and
+  nothing else.
+* **Idempotent.** Every primary key in this schema is an id the ingester
+  supplied, so re-applying the same file inserts nothing a second time. A run
+  already on the durable side keeps the row it has; this never overwrites one,
+  because the local copy is the working set and the durable copy is the
+  record.
+* **Children after parents**, so the foreign keys hold on the way in:
+  `runs` -> `trials` -> everything keyed off a trial.
+* **One transaction.** A half-applied mirror is orphan rows, so the file opens
+  `BEGIN` and closes `COMMIT`.
+* **Verify before mark.** The stamp of `migrated = 1` / `migration_source` is
+  an `UPDATE` at the end of the durable-side transaction, after every row it
+  vouches for. The *local* rows are marked by whoever ships this file, and only
+  after comparing the durable side's counts against the manifest — see
+  `verification_sql` and `SHIPPING_GAP`.
+
+## What it does not do
+
+Ship anything. The SSM wrapper that gzips this, chunks it, sends it, runs
+`verification_sql`, compares the manifest and only then marks the local rows is
+`SHIPPING_GAP`. The template named for it in #4614
+(`arenabench/scripts/mirror-experiments.sh`) left this repository with the
+ejection to `macanderson/arenabench` (#4642), so it is being written rather
+than generalized.
+"""
+
+from __future__ import annotations
+
+import argparse
+import json
+import math
+import socket
+import sys
+
+from ingest import connect, now
+
+# Where the SSM shipping wrapper is being written.
+SHIPPING_GAP = "#5097"
+
+# Every table this mirror carries, parents first. The order is the foreign-key
+# order and is what makes the emitted file appliable in one pass; `runs` is
+# handled apart from the rest because it also takes the migration stamp.
+PARENT_TABLE = "runs"
+CHILD_TABLES = (
+    "trials",
+    "events",
+    "tool_calls",
+    "verifier_tests",
+    "artifacts",
+    "step_grades",
+    "turn_grades",
+    "execution_grades",
+)
+ALL_TABLES = (PARENT_TABLE, *CHILD_TABLES)
+
+# How each table is narrowed to the selected runs. `trials` is the join every
+# other child reaches `runs` through, and `execution_grades` carries `run_id`
+# of its own, but going through `trials` for all of them keeps one predicate to
+# reason about rather than two.
+_RUN_PREDICATE = {
+    "runs": "run_id IN ({placeholders})",
+    "trials": "run_id IN ({placeholders})",
+}
+_TRIAL_PREDICATE = (
+    "trial_id IN (SELECT trial_id FROM trials WHERE run_id IN ({placeholders}))"
+)
+
+
+class UnmirrorableValue(ValueError):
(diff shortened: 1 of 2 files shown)


macanderson:
Answering the one `❌` row in Sourcery's assessment table.

**"Provide an SSM shipping wrapper."** Correctly out of scope for this PR, and
filed as #5097 with the full handoff rather than left as a gap.

Three reasons it is not here:

1. **It needs credentials CI cannot have** — `ssm:SendCommand` on
   `i-023d002d6e44f8f84` in account 578673726240, plus the `benchmarks`
   database credentials held on that instance. A wrapper written against
   neither could not be run, tested, or shown to work, which is what shipping
   it in this PR would amount to.
2. **The template named for it is gone.** #4614 says to generalize
   `arenabench/scripts/mirror-experiments.sh`; that file left this repository
   with the ejection to `macanderson/arenabench` (#4642). Its shape survives
   only in the other repo, so the wrapper is being written rather than
   generalized — a bigger piece of work than the emitter it wraps.
3. **The chunk size has to be measured, not chosen.** #4614 says so explicitly.
   Nothing here has shipped a real run yet, so `mirror.py` prints the emitted
   byte count to stderr and stops there. Picking a chunk count now would read
   like a measurement.

What this PR does ship is everything the wrapper needs and nothing it has to
guess: `emit` (the appliable transaction), `manifest` (what the durable side
must hold), `verification_sql` (how to ask it), and `mark_migrated` — kept as a
separate function from `emit` precisely so the verify-before-mark ordering
cannot be got wrong by accident, with a test asserting `emit` leaves the local
store's `migrated`/`migration_source` untouched. #5097 names those four
functions as its wiring list.

Issue #2058, opened 2026-08-07
tools: an absolute path INSIDE the workspace root is rejected as an escape (16% of all tool errors)


macanderson:
## Problem

An absolute path that points **inside** the workspace root is rejected as an escape. Measured on a clean 5/5 arenabench cycle (match `4bed110003d8`, SUT `6a41889a`), this was the single most frequent tool-error class: **8 of 49 tool errors (16%)**.

```
path `/app/feal.c` escapes workspace root (`/app/feal.c` is absolute, so it names a location outside the workspace root)
path `/app/decrypt.c` escapes workspace root (...)
path `/app/pairs.txt` escapes workspace root (...)
```

Terminal-Bench tasks run in `/app`, which **is** the workspace root. So `/app/feal.c` names a file inside the root, and the agent is told it is outside. The message is not just unhelpful, it states something false — and the model believes it, retries with variations, and burns turns.

## Where it is, and where it is *not*

The message comes from `crates/stella-tools/src/rootfd.rs:630-634`, in `steps_of`:

```rust
Component::RootDir | Component::Prefix(_) => {
    return Err(RootError::escapes(format!(
        "`{rel}` is absolute, so it names a location outside the workspace root"
    )));
}
```

**Do not "fix" it there.** `steps_of` feeds the `openat`-based rootfd walker, which resolves one component at a time against an already-open root descriptor. A `RootDir` component genuinely has no meaning relative to a dirfd, and this walker is the symlink/`..` escape boundary. Rejecting an absolute component *at that layer* is correct and must stay.

The defect is that **nothing upstream normalizes a root-prefixed absolute path into a relative one** before it reaches the walker. AGENTS.md already states the intended contract — "Path resolution is derived from the canonical root-relative path, never the raw model-supplied string" — and that normalization is simply not stripping the root prefix. The only `strip_prefix(&self.canon_root)` in the file (line 582) is used to test whether an already-canonicalised path is inside, not to normalise an incoming argument.

## What to do

At the input boundary where the model-supplied path string is turned into `rel`:

1. If the path is relative, behave exactly as today.
2. If it is absolute, canonicalise the **root** and compare: when the path is `canon_root` or lies beneath it, strip that prefix and continue with the remainder as the relative path.
3. Only when it is absolute *and* outside the root, reject — and the message should then say which root it was compared against, e.g. ``path `/etc/passwd` is outside the workspace root `/app` ``.

[...]

Issue #938, opened 2026-07-30
TOCTOU between resolve_within_root and open on intermediate path components


macanderson:
## What

`stella-tools`' path confinement resolves paths as strings and then opens them. `O_NOFOLLOW` protects only the *final* component.

For a path whose tail does not yet exist, `resolve_within_root` (`stella-tools/src/lib.rs:69-140`) canonicalizes the deepest existing ancestor and re-appends the remaining names unvalidated. `write_file` (`write.rs:75`, `durable_write.rs:92`) then `create_dir_all`s the parent and opens the target.

Between the resolve and the open, a concurrent process can replace an intermediate directory with a symlink pointing outside the workspace root, and the write lands there.

## Who the concurrent process is

This is not a hypothetical attacker who already has code execution. Two realistic sources:

1. **The model's own `bash` tool**, running concurrently with a file write in the same turn. Speculative execution makes overlapping tool calls normal rather than exceptional.
2. **A hostile repository's build script.** Stella's whole purpose is to be pointed at a codebase and run its tests. `cargo build` executing a `build.rs` from a repo under audit is the expected workflow, not a compromise.

So the window is reachable by the ordinary operation of the product on untrusted input, which is what makes it P1 rather than theoretical.

## Why the confinement is otherwise good

Worth stating, because this should not read as a broken sandbox: `resolve_within_root` rejects `..`, absolute paths, dangling symlinks, and symlinks pointing outward, and there are real adversarial witness tests for each. The gap is specifically that a *string* resolution cannot be atomic with the open that follows it.

## The fix

Stop resolving paths as strings. Open the root once and hold the descriptor, then walk components with `openat(…, O_NOFOLLOW | O_DIRECTORY)`, keeping a directory fd at each step, and perform the final `openat` relative to it. No name is ever re-resolved, so there is no window.

That means a `RootHandle` threaded through the tools rather than a path passed around. It is a contained rewrite: the existing adversarial tests transfer unchanged, and it makes `/proc/self/*`, symlink races, and bind-mount tricks structurally unreachable rather than checked against — which is the difference between a sandbox and a filter.

Worth doing in the same change: `resolve_within_root` calls `root.canonicalize` on every invocation, and a single `write_file` reaches it three times (registry classify, storage gate, tool execute). A cached root handle removes those syscalls as a side effect.

## Also worth adding

[...]

PR #5100, opened 2026-08-26
fix(stella-cli): resolve a watched witness once, through a held descriptor

### LINKED ISSUE #3483: A path-safety check can be bypassed by swapping the file between the check and its use
## Problem

`CandidateHandles::resolve_path` (`crates/stella-pipeline/src/ports/handle.rs`, added for #3380 A10) fences a caller-supplied relative path against a candidate workspace root in two layers: a pure lexical refusal (`fence_lexical` — absolute, `..`, NUL, backslash) and an on-disk containment check (`fence_on_disk` — canonicalise the deepest existing ancestor, require it under the canonical root).

The answer it returns is a `PathBuf`, and it is only true **at the instant of the check**. Anything that runs between the check and the use — including the plugin that asked, which by construction has the root path from `CandidateOp::Root` and can write inside it — can replace a component with a symlink and turn an accepted path into an escape. Classic TOCTOU.

This is stated in the module docs rather than hidden, and it is not a regression: nothing in the tree resolved plugin-supplied paths at all before A10. But it is the residual hole in what is otherwise the security half of that slice.

## What "done" looks like

The fence hands back something that cannot be re-pointed after the check — an `openat(O_NOFOLLOW)`-style resolution walked component by component from a directory file descriptor pinned at the root, with the *descriptor* (not the path) handed to the use site. That is a substrate change, not a caller change: `CandidateWorkspace` would need to grow an "open inside me" capability rather than exposing a root string, and the Windows story (`FILE_FLAG_OPEN_REPARSE_POINT`) needs a decision or a declared gap.

Constraints already discovered:

- `stella-pipeline` may do filesystem I/O (precedent: `src/replay/golden.rs`) — architecture rule 2 is about `stella-core`, not this crate. So the resolution may live here, but the *descriptor* has to reach whatever spawns the process.
- Do not add a second copy of the containment logic: `fence`/`fence_lexical`/`fence_on_disk` are the one implementation and the lexical layer stays valuable regardless (it refuses malformed input without a syscall).
- Tests to keep passing / extend: `crates/stella-pipeline/src/ports/handle/tests.rs` — `a_symlink_out_of_the_candidate_root_is_refused` and `the_lexical_fence_refuses_what_cannot_name_an_inside_path`. A witness for this issue is a test that swaps a component between resolve and use and still gets refused.

## Verify

`cargo test -p stella-pipeline --lib ports::handle`

Refs #3380.

macanderson:
## The hole

`TamperWatch` — the host's witness-integrity check for a wrapper plugin — asked
one question about a path and then answered a different one with the same
string. `observe` called `resolve_in_root` (lexical fence + `canonicalize`
containment), **discarded the `PathBuf` it returned**, and handed the raw `rel`
to `fs_artifact_identity`, which joined it onto the root and opened it again.
Two independent resolutions of a name that anything in the tree can re-point in
between, and the party this watch exists to catch is running in that tree with a
shell.

The root itself was the decisive one. The watch stored the granted root as a
`PathBuf` and re-resolved it on **every** observation, so:

1. pin `tests/witness_flip.sh` before the turn;
2. worker rewrites it (the witness now passes trivially);
3. worker renames the workspace aside and stands a byte-identical pristine copy
   at the granted path;
4. after-turn observation reads the copy, matches the pin on bytes, mode, link
   count and location, and reports **`Clean`**.

A credited flip in a directory the turn never touched. That is the shape #3483
describes — "replace a component and turn an accepted path into an escape" —
with the root as the component.

## The fix

`stella-tools`'s `RootHandle` (#938) already is the `openat`-style substrate
this issue asks for; nothing needed inventing, only reaching. Added to it:

- **`RootHandle::open_entry`** — the confined walk for a caller that must also
  *say* where it landed. `O_NOFOLLOW` on every interior component off the
  descriptor above it, `..` pops the descriptor stack, no follow at the leaf,
  and it returns a **`ConfinedEntry`**: the open `File`, the walk's own record
  of the location, and the parent descriptor.
- **`ConfinedEntry::resolved()`** — where the walk went, after every symlink
  expansion. This replaces the second `canonicalize` that used to attest the
  observed location, so the location and the descriptor can no longer be two
  answers to the same question.
- **`ConfinedEntry::still_named()`** — "does this name still mean this file",
  asked with `fstatat` through the **held parent descriptor**. It brackets the
  read on both sides, exactly as the old path-based double `symlink_metadata`
  did, without resolving anything again.

`TamperWatch` now holds a `RootHandle` for the length of the turn. A descriptor
cannot be re-pointed by renaming anything, so step 3 above stops working.

`stella-plugin` keeps the half a near-leaf over borrowed text can host:
`fence_lexical` is now `pub` and still runs first, because `RootHandle` would
*rebase* an absolute path lying under the root (#2058 — right for a model
naming its own workspace file, wrong for a plugin-declared artifact). The
on-disk half that handed back an expiring `PathBuf` is **deleted** rather than
left standing as an API whose answer is only true at the instant it is given:
`fence`, `fence_on_disk`, `resolve_in_root`. The issue's comment suggested
`stella-runtime` for the descriptor-holding half; `stella-tools` is where the
descriptors already live, and AGENTS.md forbids a second copy of the
containment logic.

Exemplar for the walk: `openat2`/`RESOLVE_BENEATH`-style resolution as
`cap-std` and `openat` do it — a descriptor stack rather than a `starts_with`
on a canonicalised string.

## Witness

`wrapper_candidate::tests::a_root_renamed_after_the_grant_cannot_launder_a_rewritten_witness`

**Ablation** — restore the old behaviour by re-resolving the root by path on
every observation (`RootHandle::open(self.root.path())` inside `observe`):

```
test wrapper_candidate::tests::a_root_renamed_after_the_grant_cannot_launder_a_rewritten_witness ... FAILED

thread '…' panicked at crates/stella-cli/src/wrapper_candidate.rs:596:9:
assertion `left == right` failed: the watch answers about the granted tree, not about the granted path
  left: Clean
 right: Tampered { artifact: "tests/witness_flip.sh" }

test result: FAILED. 9 passed; 1 failed; 0 ignored; 0 measured; 2324 filtered out
```

The ablated code produces **`Clean` where the truth is `Tampered`** — a wrong
answer, not a constant one. With the fix: 10 passed, 0 failed.

## Deleted tests (named, per AGENTS.md)

- `stella_plugin::candidate_grant::tests::the_free_fence_refuses_what_the_lexical_layer_already_refused`
  — it tested `resolve_in_root`, which this PR deletes. Its lexical assertions
  are already covered by `the_lexical_fence_refuses_what_cannot_name_an_inside_path`
  in the same module; its containment assertions move to the confined walk's own
  tests in `rootfd.rs`.

No test was renamed. `opened_witness_identity_rejects_path_retarget_before_fingerprinting`
keeps its name and now drives the new primitive.

## New tests

`rootfd.rs`: `an_entry_reports_the_location_its_own_walk_reached`,
`an_entry_refuses_a_symlink_at_the_leaf`,
`an_entry_renamed_out_from_under_its_descriptor_is_no_longer_named`.
`agent/tools/tests.rs`: `an_artifact_reached_through_an_outward_link_has_no_identity`
— labelled in the source as a containment regression guard and **not** a witness,
because the pre-fix code returned the same `None` (it followed the link, read and
hashed the outside file, then discarded the result). What changed is that the
outside file is no longer opened at all.

## Evidence run locally (targeted only)

- `cargo test -p stella-tools --lib` — 560 passed, including the three new `rootfd` tests
- `cargo test -p stella-plugin` — 143 + 84 across the integration binaries, 0 failed
- `cargo test -p stella-cli --bin stella` — 2351 passed (on the rebased head)
- `cargo clippy -p stella-tools -p stella-plugin --all-targets -- -D warnings` — clean
- `cargo doc -p stella-plugin -p stella-tools --no-deps` — no warnings
- `make guards-fast` and `make wire-schema` — green, including `stat-portability` (the new `fstatat`/`fstat`
  comparison carries a `stat-portability: ok` marker and compares two raw
  `libc::stat`s so neither identity field is cast on either platform)

CI is the evidence for the workspace; no workspace build was run here.

## Not covered

Non-Unix has no `openat`, so `open_entry` there is the string resolver plus a
`canonicalize` — two resolutions, the same concession the module header already
makes for every other method. Windows also still fails closed one layer up:
`opened_metadata` returns `None` without a by-handle link count.

Closes #3483

## Summary by Sourcery

Harden witness tamper detection by observing artifacts through a held, confined root descriptor and eliminating the vulnerable resolve-then-reopen path.

New Features:
- Add confined, descriptor-backed artifact opening that records the resolved location and verifies the entry remains named consistently while it is read.

Bug Fixes:
- Prevent tamper-watch checks from being laundered by renaming the granted workspace and replacing its path with a pristine copy.
- Prevent witness identity checks from resolving and reopening paths separately, including through symlinked paths that escape the workspace.

Enhancements:
- Keep the granted workspace root open for the duration of tamper observation and retain lexical path validation as the first containment layer.
- Remove path-based on-disk fencing APIs in favor of descriptor-based containment and reuse the confined-walk metadata for witness identity checks.

Documentation:
- Update wrapper protocol documentation to describe descriptor-backed host containment and lexical validation.

Tests:
- Add coverage for resolved locations, leaf symlink rejection, renamed entries, outward-link containment, and root-renaming witness laundering.


macanderson:
Note for review: this PR removes the last constructor of `stella_protocol::CandidateDenial::Path` / `PathDenial` — `stella-plugin`'s `fence`/`fence_on_disk` were it. The variant never reached the wire before either (`fence`'s only caller discarded the error with `.ok()?`), so this is "constructed and thrown away" → "not constructed", not a removed capability. Deciding whether the protocol keeps it is tracked as a new row on #4754 rather than folded in here.

macanderson:
## Answering the one ❌ in Sourcery's assessment

> The PR implements this for TamperWatch witness observation via
> `RootHandle::open_entry`, but it does not update the issue's
> `CandidateHandles`/`CandidateWorkspace` resolution and use path or show the
> descriptor being propagated to the plugin/process use site.

**Those code paths do not exist in this workspace, and have not since #3865.**
#3483's body was written against `crates/stella-pipeline/src/ports/handle.rs`;
that crate was deleted, and the issue's own top comment says so and repoints
every path in it at `crates/stella-plugin/src/candidate_grant.rs`. Checked on
this branch:

- **`CandidateHandles` appears in no Rust code.** `rg CandidateHandles crates/`
  reaches two doc comments, both of which say "until #3865 deleted that crate".
  There is no handle registry to teach `openat` to, and no `CandidateOp`
  dispatcher: every handle-addressed operation answers "unknown handle", which
  `stella_plugin::candidate_grant::HOST_TREE_HANDLE`'s doc explains is the true
  answer for a tree the host is already running in.
- **`CandidateWorkspace` still exists but is a different thing.**
  `crates/stella-runtime/src/wrapper/candidate_fanout.rs`'s `CandidateWorkspace`
  is the fan-out bookkeeping type — a label and a canonical `root: String`
  handed to a substrate that creates and adopts worktrees. It has no
  `resolve_path`, no path fence, and no plugin-supplied relative path crossing
  it.
- **There is exactly one plugin-supplied path list that reaches the filesystem,
  and this PR is it.** `BeforeTurnResponse::witness` → `TurnDecision::witness()`
  → `TamperWatch::pin_declared` → `observe`. Its sibling
  `BeforeTurnResponse::scope` is declared advisory and has an accessor
  (`wrapper/mod.rs`'s `scope`) with no consumer, so nothing from it is ever
  joined to a root.
- **"Propagating the descriptor to the plugin/process use site" is not available
  and not what the design asks for.** The plugin is a separate process across a
  socket; a descriptor does not cross that boundary, and the grant is
  deliberately a root string the plugin resolves under its own authority
  (`wrapper_candidate.rs`'s module doc: "the plugin says which artifacts, the
  host says whether they changed"). The *host's* use site is the identity
  observation, and that is exactly where the descriptor now goes.

The row's other two objectives — retain the lexical fence and delete the
expiring `PathBuf` API, and declare the platform story — are the ones this
repository still has a code path for, and both are marked ✅.

If a candidate-handle registry returns, its `resolve_path` should call
`RootHandle::open_entry` rather than re-growing a fence. That is the substrate
this PR puts in place, and `stella-plugin`'s module doc now says so where the
next author will meet it.

PR #5101, opened 2026-08-26
fix(stella-cli): a resumed session stops re-minting its dead lanes' ids

macanderson:
## What #3233 asked for, and what was left

PR #4725 landed the first half: each deck sub-session lane binds its own `SessionDurability` under `lane_journal_key(session, lane)` = `{session}__{lane}`, so a lane checkpoints into its own `refs/stella/…` namespace and its terminal `discard` cannot reach the lead's resume point. `agent/tests/durability_isolation.rs` holds that with four tests and a lexical fence.

What remained is item 1's other half — **the durable record of which lanes a session had running** — and it turns out that gap is not merely a missing feature. It is a live way for the durability #4725 added to hand a fresh dispatch the wrong transcript.

## The defect

`SubSessions::new` starts `next_req` at 0 on every launch, and a resumed deck **keeps its session id** — `command_deck.rs` says so where it adopts the record: *"the registry never forks a resumed session's identity."* So:

1. A session runs `req:1`, which is killed mid-turn. Per #4725 it leaves a checkpoint at `{session}__req-1`.
2. The deck is resumed. `session_persist::replay_session` restores the transcript; `SubSessions::new()` resets the counter.
3. The user types a new, unrelated prompt. `next_req_lane()` mints `req:1` again.
4. `run_worker` binds `{session}__req-1`, and `initial_messages` finds the dead lane's checkpoint and **re-enters that conversation** rather than starting on the prompt just typed.

Nothing guarded this. It is reachable only because lanes gained durable identity, which makes it #3233's own remainder rather than a separate bug.

## The record already existed

Every ref this store writes lives under its key, so `refs/stella/{session}__req-<n>/…` **is** the list of lanes a session ran. Nothing could ask for it: `WorkJournal::recorded_sessions` is private, unfiltered, and exists for retention.

- `WorkJournal::recorded_keys(prefix)` is the read path. It learns **no composition rule** — a prefix is whatever the caller says — so the `__` convention stays in the crate that mints it, and the store does not acquire a concept of a lane.
- `recorded_sessions` stays private (an age cutoff is `prune`'s business alone) and now shares one refname parse, `journal_key_of`, with the new reader instead of carrying a second copy of the `head`/`tree`/no-`/` rule.
- `SessionDurability::recorded_keys` is the best-effort passthrough, empty while unbound — a store that will not answer must not stop a session starting.
- `SubSessions::resuming` starts the counter above the highest `req:<n>` already recorded, and the deck builds its board through it (one line, so `command_deck.rs`'s ratchet is unmoved).

**Why the refs and not the session event journal.** `session_persist::journal_lanes` already names a resumed session's lanes and is in scope three hundred lines above, which makes it the tempting source. It is the wrong one: the hazard is a *checkpoint* being picked up, checkpoints live in the work journal, and the event journal only correlates with it. A signal that merely correlates with the hazard is not the hazard.

## Witnesses, ablated

**1. The mechanism** — `a_resumed_session_never_re_mints_a_lane_id_that_already_has_a_transcript`. It writes a checkpoint under `{session}__req-1`, asks for the next lane, and then binds the lane it got to prove that key is clean. Ablated by making `resuming` ignore the store:

```
---- subsession::tests::a_resumed_session_never_re_mints_a_lane_id_that_already_has_a_transcript ----
assertion `left == right` failed: a resumed session must step over the lane ids it already recorded
  left: "req:1"
 right: "req:2"

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 2351 filtered out
```

**2. The call site** — `the_deck_builds_its_lane_board_through_the_resuming_seam`. This one is here because the first ablation exposed a gap in my own test: reverting the deck's single call site to `SubSessions::new()` left **every assertion in test 1 passing**. The mechanism witness proves the mechanism; only a fence proves the driver uses it. Ablated by that revert:

```
---- subsession::tests::the_deck_builds_its_lane_board_through_the_resuming_seam ----
command_deck.rs builds a lane board with `SubSessions::new`. A deck session id may
already have `req:<n>` lanes recorded under it, and re-minting one lands the next
dispatch inside a dead lane's transcript (#3233). Use `SubSessions::resuming`.

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 2351 filtered out
```

Both ablations make the answer **wrong**, not merely constant.

**3.** `work_journal::tests::a_prefix_finds_every_key_recorded_under_it_and_nothing_else` covers the new store API directly: a key that only checkpointed (a `head`), one that only snapshotted (a `tree`), a neighbour session's key, and a turn mark that must name no key. Ablated by dropping the `tree` arm from `journal_key_of` — the snapshotted key vanishes, which is a live key reported as absent and exactly the reading that would let a caller reuse a name it must not:

```
assertion `left == right` failed: a checkpointed key and a snapshotted one are both
recorded, and a turn mark names no key of its own
  left: ["ses-1__req-1"]
 right: ["ses-1__req-1", "ses-1__req-2"]
```

**4.** `another_sessions_lanes_do_not_move_this_sessions_counter` — sessions share one store, so an unprefixed read would start a fresh deck's numbering wherever the busiest neighbour left off. **5.** `an_unbound_session_starts_its_lane_numbering_where_a_fresh_one_does` — the degradation path.

## Scope: `Refs`, not `Closes`

#3233's item 2 asks for a read path that can **re-enter** an interrupted lane's turn — `stella daemon resume` targeting a lane. That is not here, and it is a larger thing than this: lane keys are not in the `SessionRegistry`, so `daemon::resolve` can never produce one. Filed as #5102 with what this PR established about where the record lives. What this PR settles is that the record is now readable, and that reading it closes the way per-lane durability could corrupt a fresh dispatch.

Two smaller findings filed rather than folded in: #5103 (stale prose — two doc comments still say the lane key is `{session}/{lane}`, which `lane_journal_key`'s own doc explains is the one shape it cannot be) and #5104 (nothing reaps a dead session's lane refs; only the age-based prune eventually does).

## Local runs

- `cargo test -p stella-store` — `435 passed; 0 failed`
- `cargo test -p stella-cli --bin stella` — `2352 passed; 0 failed`
- `cargo clippy -p stella-store -p stella-cli --all-targets -- -D warnings` — clean
- `make guards-fast` — clean

CI is the evidence; the workspace build is not run here.

Refs #3233, #4725, #3232

## Summary by Sourcery

Ensure resumed deck sessions mint fresh lane IDs by deriving the next number from their recorded durability refs.

New Features:
- Add a work-journal API for discovering recorded keys by prefix.

Bug Fixes:
- Prevent resumed sessions from reusing lane IDs with existing transcripts, avoiding accidental re-entry into dead lane conversations.

Enhancements:
- Initialize resumed deck lane numbering above the highest previously recorded lane for that session.
- Centralize journal reference parsing for recorded-key and retention queries.

Tests:
- Add coverage for lane ID advancement, session isolation, unbound durability fallback, deck integration, and journal key discovery.

### DIFF
--- a/crates/stella-cli/src/command_deck.rs
+++ b/crates/stella-cli/src/command_deck.rs
@@ -867,7 +867,7 @@ pub async fn run_deck_session(
     let mut pending_settings_reload = false;
     // Sub-session bookkeeping: live-worker slots, and `task_assign` requests
     // waiting for one (drained oldest-first as workers end).
-    let mut subs = SubSessions::new();
+    let mut subs = SubSessions::resuming(&cfg.durability, &session_record.id);
     let mut pending_spawns: VecDeque<subsession::QueuedSpawn> = VecDeque::new();
     // Lanes whose Restart arrived while the worker was still live: stop
     // first, respawn on its Ended.
--- a/crates/stella-cli/src/durability.rs
+++ b/crates/stella-cli/src/durability.rs
@@ -300,6 +300,19 @@ impl SessionDurability {
         self.journal()?.pipeline_frame()
     }
 
+    /// Every journal key in this workspace's store beginning with `prefix`,
+    /// sorted; empty while unbound or when the store cannot be read.
+    ///
+    /// Best-effort by design, like every other read on this handle: a caller
+    /// asking which keys exist is deciding what to *avoid*, and a store that
+    /// will not answer must not stop a session starting. The cost of an empty
+    /// answer is the behaviour that stood before anything asked.
+    pub fn recorded_keys(&self, prefix: &str) -> Vec<String> {
+        self.journal()
+            .and_then(|journal| journal.recorded_keys(prefix).ok())
+            .unwrap_or_default()
+    }
+
     /// The sink to hand [`stella_core::EngineConfig::checkpoint_sink`], or
     /// `None` while this handle is unbound.
     ///
--- a/crates/stella-cli/src/subsession.rs
+++ b/crates/stella-cli/src/subsession.rs
@@ -130,6 +130,43 @@ impl SubSessions {
         }
     }
 
+    /// A board for a session that may already have run `req:<n>` lanes —
+    /// every deck session, since a resumed one keeps its id.
+    ///
+    /// [`Self::new`] starts `next_req` at 0, so the first prompt lane of a
+    /// resumed session mints `req:1` again. That id is not cosmetic: it is
+    /// half of the lane's journal key ([`lane_journal_key`]), and
+    /// [`run_worker`] re-enters whatever transcript it finds there rather than
+    /// starting from the prompt. A fresh dispatch would therefore wake up
+    /// inside a dead lane's conversation — the one thing per-lane durability
+    /// (#3233) must not make possible.
+    ///
+    /// So the counter starts above the highest `req:<n>` the store has already
+    /// recorded for this session. The refs under `{session}__…` are that
+    /// record, and they are the same refs the collision would have read, which
+    /// is why they are asked rather than the session event journal: a signal
+    /// that merely correlates with the hazard is not the hazard.
+    ///
+    /// Unreadable or unbound reads as none recorded, which is
+    /// [`Self::new`]'s behaviour — a store that will not answer must not stop
+    /// a session starting.
+    pub(crate) fn resuming(
+        durability: &crate::durability::SessionDurability,
+        session_id: &str,
+    ) -> Self {
+        let prefix = lane_journal_key(session_id, "req:");
+        let highest = durability
+            .recorded_keys(&prefix)
+            .iter()
+            .filter_map(|key| key.strip_prefix(&prefix)?.parse::<u64>().ok())
+            .max()
+            .unwrap_or(0);
+        Self {
+            next_req: highest,
+            ..Self::new()
+        }
+    }
+
     pub(crate) fn has_slot(&self) -> bool {
         self.active < MAX_CONCURRENT
     }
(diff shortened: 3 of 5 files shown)


macanderson:
Correcting one detail in the local-command guidance, because the next instance of this will be checked with whatever command we write down.

The distinguishing variable is **`RUSTDOCFLAGS="-D warnings"`, not `--document-private-items`**. I put the bad link back on a clean build and ran all three:

| Command | Exit |
|---|---|
| `cargo doc -p stella-store --no-deps` | **0** — warning printed, not fatal |
| `RUSTDOCFLAGS="-D warnings" cargo doc -p stella-store --no-deps` | 101 |
| `RUSTDOCFLAGS="-D warnings" cargo doc -p stella-store --no-deps --document-private-items` | 101 |

`--document-private-items` does not silence it. It emits the same error and adds a note aimed straight at this mistake:

```
error: public documentation for `recorded_keys` links to private item `Self::recorded_sessions`
```

The false green I actually hit was the first row: I ran `cargo doc -p stella-store --no-deps` with no `RUSTDOCFLAGS`, saw exit 0, and reported it as evidence. The warning was in my output the whole time; nothing made it fatal. So the command to write down is:

```sh
cargo clean -p <crate>
RUSTDOCFLAGS="-D warnings" cargo doc -p <crate> --no-deps
```

with either form of the last flag. The `cargo clean` half of the advice was right and load-bearing — I needed it to reproduce at all.

Verified on this branch at 9039840e3 (both crates, clean build): `stella-store` exit 0, `stella-cli` exit 0. Also ran the same check against #5093, which I had originally cleared with the same too-weak command — both crates exit 0 there too.

I reset to your 9039840e3 rather than push my own edit to that line; the substance was identical and two edits to one line is the shape that goes wrong on a rebase.

PR #5112, opened 2026-08-26
ci: add triage separation-of-duties guard (SCR-005)

macanderson:
Adds `.github/workflows/triage-guard.yml` (L2 autonomy, GitHub Action):

- On `issues: opened` — auto-applies the `triage` label so every new issue enters the triage queue.
- On `issues: labeled` with a `P0`–`P3` label — if the sender is not the triage agent (`triage-bot`), the priority label is removed, `triage` is re-applied, and a comment explains that priority labels are assigned only by the triage agent per SCR-005.

Enforces the separation of duties from the issue-triage workflow: creators label `triage` only; the triage agent alone assigns P0–P3.

## Summary by Sourcery

Enforce SCR-005 separation of duties for issue triage by ensuring new issues enter the triage queue and only the triage agent can assign priority labels.

Enhancements:
- Restrict P0–P3 issue labeling to the dedicated triage agent and restore unauthorized priority labels to the triage queue.
- Automatically apply the `triage` label when issues are opened.

CI:
- Enforce triage separation of duties through the triage guard workflow.


macanderson:
#5112 is red on a supply-chain guard, not on anything about the triage logic:

```
check-action-pins: these 'uses:' references are not pinned to a commit SHA:
.github/workflows/triage-guard.yml:11:      - uses: actions/github-script@v7
```

Worth flagging that this is *not* the red-`main` hold that is currently reddening other PRs — `main` is red on #5124 and the repair is #5131, but this PR has its own separate failure and will still be red after that lands.

The convention in this tree is a full commit SHA with the version as a trailing comment, so a reader can see what it is and a mutable tag cannot move under it:

```yaml
- uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
```

So `triage-guard.yml:11` wants the same shape:

```yaml
- uses: actions/github-script@<full-sha>  # v7
```

Resolve the SHA from the tag rather than copying one from anywhere:

```sh
gh api repos/actions/github-script/git/ref/tags/v7 --jq .object.sha
```

(If that returns a tag object rather than a commit, dereference it: `gh api repos/actions/github-script/git/tags/<sha> --jq .object.sha`.)

The guard is `scripts/check-action-pins.sh` and it runs inside the `fmt + clippy + test` job, so it fails the whole job — which is why the failure reads as a Rust one at a glance. `./scripts/check-action-pins.sh` reproduces it locally in milliseconds and compiles nothing.

Why the rule exists, since it is easy to read as ceremony: a floating tag is a live write into this repository's CI from a repository nobody here controls. `actions/github-script` runs with the workflow's token, so a moved tag is arbitrary code with whatever permissions that job holds. The pin is what makes the dependency reviewable.

Issue #5126, opened 2026-08-26
The loop's events wake a shell hook but never reach the durable stream anything reads later


macanderson:
## Problem

#4017 asked whether the self-driving loop's events should also become `AgentEvent` options, and this is that question on its own, unanswered:

> **Do these also become `AgentEvent` options?** The ask that produced this was "events that can be subscribed to by other services or even other stella agents". `HookEvent` is the *subscribe* surface (in-process, synchronous, shell command). `AgentEvent` is the *emit* stream — durable, on the wire, and readable by the Observatory. Several of these want both.

#4017 shipped the `HookEvent` half alone. So today a run, a cycle, an escalation, a merge and a broken base can each wake a shell command **while the loop is running on this machine**, and none of them is in the durable stream anything reads afterwards.

## What the gap costs

A hook is in-process and synchronous: it exists only while the loop does, and only on the machine the loop runs on. Everything that wants to read this *later* or *elsewhere* — the Observatory, a dashboard, a second agent — reads `AgentEvent`, and none of these events is there.

`BaseBroken` is the sharpest case. `main-canary`/`main-red-hold` are exactly the consumers AGENTS.md names for it, and they are workflows rather than a shell hook on somebody's laptop.

## The constraint that shapes the work

**Architecture rule #10.** An `AgentEvent` option must declare its consumer in `crates/stella-protocol/src/event/consumers.rs` or it is an `E0004` — a build error, not a red test. So this cannot be done by adding options: each one needs a `ConsumerPosture`, and a `Behavioral` row must name the code that branches on it. An event with no consumer is a `RecordedOnly` or `Unclassified` row citing an issue, and `MAX_UNCLASSIFIED` is a down-only ratchet — so adding seventeen unclassified rows is not available either.

That makes the real question **which of these have a consumer worth building**, not whether to mirror them all. Mirroring all seventeen because they exist is the shape rule #10 exists to refuse.

## Where

- `crates/stella-protocol/src/event/tags.rs` — the macro table `KNOWN_TYPE_TAGS`, `type_tag()` and the consumer rows are all generated from.
- `crates/stella-protocol/src/event/consumers.rs` — one row per tag, with its posture.
- `crates/stella-protocol/src/hook.rs` — `HookEvent`, the vocabulary being mirrored.
- `crates/stella-cli/src/self_driving_cmd/` — the verbs that would emit, already dispatching the hook half (`hooks.rs`, `drive/notify.rs`).
[...]

Issue #5125, opened 2026-08-26
IssueCommented means two different things, so it is the one loop event still undeclared


macanderson:
## Problem

#4017 declared the self-driving loop's hook vocabulary and left exactly one event out, because its name means two different things:

> **Open question — `IssueCommented` is ambiguous** and should be settled before it is declared: *the loop commented* and *a human commented and the loop noticed* are different events with different subscribers. Either split them (`IssueCommented` / `IssueCommentObserved`) or pick one deliberately.

That question is still open. `HookEvent::ALL` now carries every other tracker, pull-request, check and lifecycle event, and this is the one gap.

## Why it matters more than it looks

The two readings have opposite subscribers:

- **The loop commented** — an audit trail. A subscriber wants to know what the agent said on an issue, and the payload's subject is the loop's own text.
- **A human commented** — an *interrupt*. A subscriber wants to know somebody is talking to the agent, which is the signal that a person is trying to steer or stop something. That is the more valuable of the two and the harder to build: nothing in the loop polls for new comments today, so declaring it would be declaring an event no verb can fire.

Declaring one name for both would make a subscriber unable to tell "the agent is working" from "a human is intervening".

## Where

- `crates/stella-protocol/src/hook.rs` — `HookEvent`, its `ALL`, `as_str`, `in_turn` and the `WIRE_STRINGS`/`declared_index` test constants.
- `crates/stella-core/src/hooks.rs` — the `Hooks` field and `matchers_for` arm.
- `crates/stella-core/src/hooks/payload.rs` — the payload, if the event carries the comment body (it should not carry it unclipped without deciding that on purpose: `tool_result`'s posture is the precedent).
- `crates/stella-plugin/src/wire/disclosure.rs` and `crates/stella-tools/tests/hook_disclosure.rs` — a row per payload field, and the census entry. A comment body crossing to a third party's process is a real disclosure question, not a formality.
- `crates/stella-cli/src/settings/{merge,completeness}.rs`, `crates/stella-cli/src/plugin_hooks.rs`, `crates/stella-plugin/src/consent/moments.rs` — the tables the compiler and the completeness test will demand.
- `crates/stella-cli/src/self_driving_cmd/backlog.rs` — `comment` is where the loop's own comment goes out.

## Verify

`crates/stella-cli/tests/loop_hook_dispatch.rs` already fails on a declared event nothing dispatches, so the "do not declare what cannot fire" half is enforced: adding the variant without a dispatch site turns that test red. The payload half is
[...]

PR #5132, opened 2026-08-26
feat(stella-protocol,stella-cli): the rest of the autonomous loop's hook events

### LINKED ISSUE #4017: The rest of the autonomous loop's hook events: tracker, pull-request and check events
Follow-up to #4016, which landed `PreIssueWork` and `PostIssueWork`. The naming and the shape of the rest were approved with them; the remaining events were deliberately **not** declared because a hook point nothing dispatches is a declaration that quietly does nothing, and the `deliver` verbs did not exist on main at the time.

**#4001 has since landed**, so `self_driving_cmd/deliver.rs`, `drive.rs` and `lifecycle.rs` now exist and each event below has somewhere real to fire from. That is the unblock.

## The naming rule these follow

PascalCase; `Pre`/`Post` for a pair that brackets something, a past participle for a thing that happened, no `ON_`/`BEFORE_` prefixes — the tense lives in the name. Stated on `stella_protocol::hook`'s module doc.

### Loop lifecycle
| Event | Fires from |
|---|---|
| `DriveRunStart` | `run start` — before the first cycle |
| `DriveRunEnd` | `run end` — carries why it stopped |
| `DriveCycleStart` / `DriveCycleEnd` | `cycle begin` / `cycle end`, which already write ledger records |
| `DriveIdle` | the dry-streak advance. The signal that separates *alive but starved* from *dead* — a monitor cannot get that from silence |

### Tracker
| Event | Fires from |
|---|---|
| `IssueCreated` | `self-driving file` (`backlog::file_finding`) |
| `IssueClosed` | the close path, `--partial` included |
| `IssueEscalated` | carries the existing `EscalationReason` (`FixCeilingReached` / `RebaseCeilingReached` / `ReviewNeedsHuman`), so a subscriber knows whether to fix, wait, or raise a ceiling |

**Open question — `IssueCommented` is ambiguous** and should be settled before it is declared: *the loop commented* and *a human commented and the loop noticed* are different events with different subscribers. Either split them (`IssueCommented` / `IssueCommentObserved`) or pick one deliberately.

### Pull request
`PullRequestOpened` · `PullRequestReadyForReview` (`Action::MarkReady`) · `PullRequestConflicted` (`PrState::Conflicted`) · `PullRequestMerged` — all off `deliver`'s transitions.

### Checks — the one that must not be collapsed
`ChecksFailed` (`PrState::CiRed` — this PR's fault) and `BaseBroken` (`PrState::BaseBroken` — the base is red too) are **two events**, because the whole `deliver` machine turns on that distinction: collapsing them is how a loop spends its budget fixing somebody else's breakage. `BaseBroken` is also the event `main-canary`/`main-red-hold` would want to consume. Plus `ChecksGreen` for a subscriber that gates on it.

### Safety
`DriveBudgetExhausted` (hit the spend ceiling) · `DriveRefused` (declined to start — steering switched off, a `main-red` hold). Distinct from an error: nothing broke, the loop chose not to run.

## Decisions to make before writing code

1. **Which are veto points and which only report.** `PreIssueWork` is a veto; most of these are past-tense and cannot be. `DriveRunStart` and `DriveCycleStart` plausibly should be.
2. **Matcher scoping.** All the events shipped so far are `tool_scoped() == false` (matcher ignored). The tracker events could plausibly be matcher-scoped by repo or label — a real option, worth a decision rather than a default.
3. **Do these also become `AgentEvent` options?** The ask that produced this was "events that can be subscribed to by other services or even other stella agents". `HookEvent` is the *subscribe* surface (in-process, synchronous, shell command). `AgentEvent` is the *emit* stream — durable, on the wire, and readable by the Observatory. Several of these want both. **Note architecture rule #10**: an `AgentEvent` option must declare its consumer in `event/consumers.rs` or it is an `E0004` build error.
4. **Fan-out to a remote sink is a separate design.** Hooks are in-process and synchronous. "Subscribed by other services" over a network needs a transport, and would need architecture rule #3's egress review.

## Constraints carried from #4016

- **Do not add these to `EVENT_ORDER`** in `plugin_cmd/roster.rs` unless plugin dispatch is also wired — that table is the in-turn plugin routing surface, and a row there with no dispatcher is a route nothing serves. `ManifestError::HookNotAvailableToPlugins` is the refusal that keeps a plugin from declaring one meanwhile; extend or remove it deliberately.
- **Every new event needs a `join` line in `settings::merge::concat_hooks`** and a row in `settings::completeness::hooks_ledger`, or a matcher is parsed from every scope and silently dropped. The completeness test enforces this and will fail first; do not treat that failure as noise.
- Update the `EVERY_KEY` fixture in `completeness.rs` in the same commit.

## Done when

Each declared event demonstrably fires from the verb named above, with a witness test asserting the payload a subscriber reads — and any event that cannot yet fire is not declared.

macanderson:
## What & why

#4016 landed `PreIssueWork` and `PostIssueWork` and stopped there on purpose: the rest of the vocabulary was designed and approved, but the `deliver` verbs it would fire from did not exist yet, and *a hook point nothing dispatches is a declaration that quietly does nothing*. #4001 landed those verbs. This declares the remaining seventeen and fires each from the verb #4017 names.

Closes #4017

**Five** loop-lifecycle (`DriveRunStart` · `DriveRunEnd` · `DriveCycleStart` · `DriveCycleEnd` · `DriveIdle`), **three** tracker (`IssueCreated` · `IssueClosed` · `IssueEscalated`), **four** pull-request, **three** check (`ChecksFailed` · `BaseBroken` · `ChecksGreen`), **two** safety (`DriveBudgetExhausted` · `DriveRefused`).

### The decisions the issue asked for, and what settles each

**1. Which are veto points: none of them.** The naming rule already answers it — a `Pre` brackets something that has not happened, and every one of these is a Start, an End, or a past participle. A run- or cycle-level veto would also duplicate `PreIssueWork` at a coarser grain while being unable to say *which* work it withheld, which is the one thing an operator reading a skip needs.

**2. Matcher scoping: none.** `tool_scoped()` stays the two tool events, matching every other non-tool event.

**3. `AgentEvent` options: not here** — filed as #5126, because rule #10 means each one needs a declared consumer, and "mirror all seventeen because they exist" is the shape that rule exists to refuse.

**4. `IssueCommented`: still not declared** — filed as #5125 with the split question intact.

**5. `EVENT_ORDER` is untouched**, per the issue's constraint. `ManifestError::HookNotAvailableToPlugins` now refuses on `HookEvent::in_turn` instead of a two-name `matches!`, so a plugin declaring any loop event is refused without anyone remembering to extend a list.

### `HookEvent::in_turn` replaces three copies of the same five names

The plugin refusal, the `EVENT_ORDER` exhaustiveness test and the prose each held their own list of the in-turn events. `in_turn` is a total `match`, so a new event chooses a side before it compiles — the `matches!` allowlist form would have silently placed it outside.

### One live bug fixed on the way

`settings/unknown.rs`'s `HOOK_EVENTS` held **three** names while the enum held **seven**. Registering `Stop`, `PreCompact`, `PreIssueWork` or `PostIssueWork` — every one of which works — was reported to the user as an unknown key on **every settings load**. Its own doc comment calls a misspelled event "the highest-consequence typo in the whole file". It reads the vocabulary now, so it cannot drift again.

## The witness

Two tests, covering the halves neither can reach alone:

**`crates/stella-cli/tests/loop_hook_dispatch.rs`** — every event outside a turn is dispatched by a verb, and no in-turn event is dispatched by the loop.

**`self_driving_cmd/hooks/tests.rs`** — each event through a real `bash -c` hook, asserting the payload a subscriber reads. Driven from `HookEvent::ALL`, so a future event gets both for free.

### Ablation 1 — collapse `BaseBroken` into `ChecksFailed`

The exact mistake the issue warns about ("collapsing them is how a loop spends its budget fixing somebody else's breakage"):

```
---- every_loop_hook_event_is_dispatched_by_a_verb stdout ----
these events are declarable and dispatched by nothing, so registering one does nothing
and says nothing: ["BaseBroken"]. Fire it from the verb it names, or do not declare it.

```

### Ablation 2 — make `HookRunInfo::in_cycle` drop the cycle

The answer becomes **wrong**, not absent:

```
---- a_reporting_event_carries_the_identity_its_family_is_keyed_by ----
  left: Null
 right: 7

test result: FAILED. 10 passed; 1 failed
```

Restored: `11 passed; 0 failed` and `2 passed; 0 failed`.

The dispatch guard also found its own blind spot on its first run — it read `PostIssueWork` as undispatched because that event is built through a named constructor rather than passed as a value. That is in the predicate's doc comment.

## The gate

- [x] `cargo fmt --check`
- [x] `cargo clippy -p stella-protocol -p stella-core -p stella-plugin -p stella-tools -p stella-cli --all-targets -- -D warnings`
- [x] `cargo test` for each of those crates (full suites, green)
- [x] `make guards-fast` — see the note below on `check-prose`
- [x] Docs updated: `website/content/docs/agent-tools/hooks.mdx` (a card and a section per event), `configuration/settings.mdx`, `crates/stella-core/README.md`, `docs/spec/oxagen-trace-drain.md`

## Reviewers should know

**`main` is already red on the header-density ratchet that landed in #5096, and this PR does not add to it.** Measured on `origin/main` at `a7e1d3d33`: `stella-cli` 23.74, `stella-store` 26.02, `stella-tools` 35.77, all above baseline. On this branch: `stella-cli` **23.69** (below main's), `stella-store` 26.02 and `stella-tools` 35.77 unchanged — this branch touches neither. `stella-protocol` went over while I was writing and is back under. The inherited failure is #5124.

**Three files moved to stay under the 1500-line ratchet**, none of them a logic change: `HookPayload` → `hooks/payload.rs`, `hook_moment` → `consent/moments.rs`, and the new dispatch logic into `drive/notify.rs` rather than into `drive.rs`. No test was deleted or renamed.

**`drive.rs` is at exactly 1500 lines.** The next author to touch it has no headroom; splitting it further is worth doing before then, not during.

**A polling loop is latched.** `drive` re-asks the machine every `poll_secs`, so a check event fires on a *change* of answer and a standing refusal is reported once. Without that a pull request waiting on review would wake a subscriber every few seconds with the same green, and the wake it then ignores is the one that mattered. The latches are per-run and die with the process, which is why a fresh run re-reports a still-standing block.

## Nothing left behind

- [x] Filed: #5125 (the `IssueCommented` split), #5126 (`AgentEvent` mirroring)

## Ground-rule check

- [x] New cross-boundary types round-trip through serde — `HookRunInfo` and `HookPullRequestInfo` are covered by the disclosure census, which serializes each event's real payload and compares the key set against the declared rows

## Summary by Sourcery

Complete autonomous-loop hook coverage by declaring, dispatching, and documenting all remaining reporting events with reliable payloads and anti-drift safeguards.

New Features:
- Add seventeen autonomous-loop hook events covering run and cycle lifecycle, tracker actions, pull requests, checks, budget exhaustion, and refusals.
- Dispatch loop-owned events with structured run, issue, pull-request, and reason payloads while preserving veto behavior only for pre-events.
- Expose the complete hook vocabulary in settings, plugin disclosures, configuration merging, and user-facing documentation.

Bug Fixes:
- Derive recognized hook settings keys from the canonical event vocabulary so valid existing events are no longer reported as unknown.
- Prevent repeated polling notifications for unchanged pull-request states and standing refusals.

Enhancements:
- Centralize hook-family classification and reporting dispatch, refuse all loop-owned events in plugin manifests, and keep reporting hook failures from interrupting loop operations.
- Distinguish pull-request check failures from base-branch breakage and preserve cycle identity in lifecycle payloads.

Documentation:
- Document the expanded hook event set, payloads, reporting semantics, and configuration behavior.

Tests:
- Add coverage proving every loop-owned event is dispatched, no in-turn event is dispatched by the loop, payload identities are preserved, reporting hooks tolerate failures, and payload fields remain synchronized with disclosure metadata.

Chores:
- Split hook payload, notification, and consent moment code into dedicated modules to remain within source-size limits.


macanderson:
Two of the four red checks on the first run were mine and are fixed in the force-push just made; two are inherited.

**Mine — both under the `schema` feature, which no local rung compiles:**
- `lint-schema`: `unused import: HookEvent` in `consent.rs`. Moving `hook_moment` to `consent/moments.rs` took the last use of the type with it. This is also what reddened *the platform split still compiles on Windows*, which runs clippy over the same crate.
- `doc-warnings-schema`: `unresolved link to super::point_moment` — I linked a private item from the new module's header. It is backticked prose now.

Both reproduce with `make lint-schema` and `make doc-warnings-schema`, and both are green locally now. Worth noting for anyone else moving an item between modules: `make guards-fast` does not compile the schema tier, so neither failure is visible on the rung a pre-push hook picks.

**Inherited — `main is not known-broken` and `gate steps no other workflow runs`:**
Both trace to #5124, the header-density ratchet that has been red on `main` since it landed. The repair is #5135. This branch measures *below* `main` on the one crate it touches (`stella-cli` 23.68 against main's 23.74); the two crates still over — `stella-store` and `stella-tools` — are not touched here. I will rebase once #5135 lands.

PR #5131, opened 2026-08-26
docs: cut history out of eight module headers to clear the density ratchet

macanderson:
Repairs red `main`. Claimed via `./scripts/main-red-claim.sh claim` before starting.

## What was red

`prose`, on the density ratchet — three crates over their mean module-header ceiling:

```
crates/stella-cli:   23.67 allowed, 23.74 mean header lines
crates/stella-store: 25.86 allowed, 26.02 mean header lines
crates/stella-tools: 35.69 allowed, 35.77 mean header lines
```

**Nobody's PR was wrong.** `scripts/prose-density-baseline.txt` holds a per-crate *mean*, which makes it a shared cell in exactly the sense AGENTS.md describes for `Cargo.lock` and the file-size baseline: several branches each adding a header above their crate's mean are each individually fine and compose into a break. No pre-merge run has both halves in front of it. That is why the canary exists (#3332).

## What I cut, and why those lines

Cut, never raised — the baseline file is untouched, which you can see in the diff stat. The guard states its own bar: *a header is what this file does and what a reader must not do to it; history belongs in the pull request, a design document, or the tracker.* So history is what went:

| File | What went |
|---|---|
| `stella-store/src/telemetry/delivery.rs` | 38 → 16. Motivation paragraph; token-set rationale moved onto the `Delivery` type |
| `stella-tools/tests/relevance_calibration.rs` | How many paid tokens a previous session declined to spend (twice — the second was a duplicate) |
| `stella-cli/src/candidate_workspaces.rs` | What #3844 landed and what did not exist yet before it |
| `stella-cli/src/wrapper_candidate.rs` | What `RoundInput` used to send; a one-fence-not-two arrangement whose other half was deleted with `stella-pipeline` (#3865) |
| `stella-cli/src/fleet_cmd/wrapped.rs` | A paragraph explaining why the file is a submodule rather than a `mod` block |
| `stella-cli/src/plugin_cmd/package.rs` | "was never the whole of a package" framing |
| `stella-cli/src/daemon.rs` | Which of #1552's two properties was the weaker one |
| `stella-cli/src/plugin_authz.rs` | "The defect this closes" section |

[...]

PR #5133, opened 2026-08-26
Cut the headers that pushed three crates over their density ceiling

### LINKED ISSUE #5124: main is red: prose fails on the merged tree
The post-merge canary tested `main` at `e9fa6d2bd` and it is red.

This run did not check `e9fa6d2bd`'s parent, so the break may be older —
check on `e9fa6d2bd^` and walk back while it still fails.

Failing: **prose**

### `prose`

```

  crates/stella-cli: 23.67 allowed, 23.74 mean header lines
  crates/stella-store: 25.86 allowed, 26.02 mean header lines
  crates/stella-tools: 35.69 allowed, 35.77 mean header lines

A header is what this file does and what a reader must not do to it -- two to four sentences. History belongs in the pull request, a design document, or the tracker.
```

canary exists (#3332).

### How to fix

```sh
git checkout main && git pull
```

...then, per failing check:

#### `prose`

```sh
#   make prose-report
```

than only CI.

macanderson:
`main` is red on the density ratchet #5096 added (#5124). Three crates exceed their recorded mean header length:

```
stella-cli    23.67 allowed, 23.74
stella-store  25.86 allowed, 26.02
stella-tools  35.69 allowed, 35.77
```

## Where the growth came from

Measured rather than guessed — header length per file at #5096's commit versus `main`:

```
+29  (69 -> 98)  crates/stella-cli/src/wrapper_candidate.rs
+14  (38 -> 52)  crates/stella-plugin/src/candidate_grant.rs
+ 7  (64 -> 71)  crates/stella-tools/src/rootfd.rs
```

**All three are mine**, from #5100's witness-laundering fix, and stella-cli's +29 is exactly its overage. Cut back to the argument without the retelling.

## stella-store is a different case, and worth stating

Its headers **never changed** — 2108 lines at #5096's commit and 2108 now. The ceiling was recorded on #5096's branch *before* #5093 merged, so 25.86 was never a number the merged tree achieved. That is the shared-cell staleness this repo keeps hitting: a ratchet recorded against one tree, landing on another.

I cut the export ledger's watermark essay anyway rather than edit the baseline. The guard's own remedy is *"Cut a header. Do not raise the number"*, and a ceiling the tree never met is not a licence to raise one. If a maintainer thinks 25.86 was simply mis-recorded, correcting it is a separate change with its own argument.

## Evidence

```
check-prose   OK — 4178 grandfathered, none added; 25 unit(s) within their header-length ceiling
make guards-fast   exit 0
cargo fmt --all --check   exit 0
```

No baseline entry was touched. Prose-only, so no witness test is owed — the guard's own verdict is the evidence.

Closes #5124

## Summary by Sourcery

Reduce documentation header density across the affected crates without changing code behavior or baseline limits.

Enhancements:
- Shorten oversized crate documentation headers while preserving their existing explanations and behavior.

Documentation:
- Condense documentation across the candidate workspace, tamper watch, telemetry watermark, root handle, search cache, and relevance calibration areas.

### DIFF
--- a/crates/stella-cli/src/candidate_workspaces.rs
+++ b/crates/stella-cli/src/candidate_workspaces.rs
@@ -5,18 +5,11 @@
 //! candidate, one writing worker turn inside each, and one all-or-nothing
 //! adoption that lands a winner on the real tree (#3892).
 //!
-//! #3844 landed the socket half of best-of-N — the wire types, the
-//! `[loop] max_fanout_width` ceiling, and
-//! [`CandidateFanouts`](stella_runtime::wrapper::CandidateFanouts) over a
-//! [`CandidateWorkspaces`] port — and **nothing implemented the port**, so
-//! every shipped host answered both capabilities
-//! [`Unavailable`](stella_plugin::HostCallRefusal::Unavailable) and
-//! `plugins/stella-candidates` (`doc:pipeline-as-plugins` §3/§7 item 4) still
-//! could not be written as best-of-N. This is that implementation.
+//! This implements the [`CandidateWorkspaces`] port #3844's socket half left
+//! unimplemented, which had every shipped host answering both capabilities
+//! [`Unavailable`](stella_plugin::HostCallRefusal::Unavailable).
 //!
 //! # What is reused, and the one thing that is not
-//!
-//! Almost everything here is somebody else's machinery held together:
 //!
 //! | Concern | Whose |
 //! |---|---|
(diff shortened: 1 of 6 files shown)


macanderson:
Closing as a duplicate repair of #5124 — #5131 covers the same break more broadly (it also fixes two stale claims in the headers it touches) and is already labelled `unblocks-main`.

We opened these independently, and the claim mechanism did not stop it: I claimed #5124 with `main-red-claim.sh`, then the other session ran `check`, was told *unclaimed*, and proceeded — because the script treats a claim by the same login as "yours, carry on". Right for one session resuming its own repair, wrong for two agents sharing a GitHub identity. Filing that separately.

The one finding here that #5131 does not carry — that `stella-store`'s headers never changed, so its ceiling was never met by the merged tree — is posted as a comment there.

PR #5135, opened 2026-08-26
docs(stella-tools): move the calibration report's columns onto the function that computes them

### LINKED ISSUE #5124: main is red: prose fails on the merged tree

macanderson:
## What & why

`main` is red on the header-density ratchet, and has been since that ratchet landed. The canary filed #5124; this is the repair, on its own branch from a fresh `main`, per AGENTS.md's rule about not folding a shared-cell repair into an unrelated PR.

Closes #5124

### What actually broke, measured rather than inferred

The ratchet's **own commit was already over** for `stella-store`. Checked out `5e06f04a8` (the #5096 merge) and ran `python3 scripts/check-prose.py` there:

```
  crates/stella-store: 25.86 allowed, 26.02 mean header lines
```

So the tree has never been green on this gate: #5096 shipped a baseline its own tree exceeded. `stella-cli` and `stella-tools` went over afterwards. At `a7e1d3d33` all three were over — `stella-cli` 23.74, `stella-store` 26.02, `stella-tools` 35.77.

That matters for the fix: there is no single culprit commit to revert, and the issue's own text warns against reaching for one.

### The repair is the one the gate names

> A header is what this file does and what a reader must not do to it — two to four sentences. History belongs in the pull request, a design document, or the tracker.

**No paragraph is deleted.** Each moves onto the item it is about, where a reader meets it beside the code instead of eighty lines above it:

| From | To | Why there |
| --- | --- | --- |
| `export_ledger.rs` — the third watermark | `pending_enterprise_export_page` | it is the function that takes that cursor |
| `relevance_calibration.rs` — "what it prints" | `measure` | it computes every number in the list |
| `candidate_workspaces.rs` — the four properties of patch adoption | `adopt` | the header already pointed at `adopt` for two of them |
| `wrapper_candidate.rs` — the "what was broken" narrative | condensed to the sentence that says why the grant exists | the rest is this repository's history, which the gate says belongs in the tracker |

Each header keeps a pointer to where its section went, so nothing is harder to find than it was. Three of the four are among the longest headers in their crate: 115, 166 and 101 lines.

## The witness

- [x] No witness needed (docs) — because every change is a doc comment, and the check that was red is itself the test:

```
check-prose: OK -- 4177 grandfathered construction(s) in 1174 file(s), none added;
25 unit(s) within their header-length ceiling.
```

It fails on `main` (three units over) and passes here. The `make prose` step is the gate, and its output is above.

## The gate

- [x] `cargo fmt --check`
- [x] `cargo clippy -p stella-store -p stella-cli -p stella-tools --all-targets -- -D warnings`
- [x] `cargo doc -p stella-store -p stella-cli --no-deps` — clean, so every moved intra-doc link still resolves
- [x] `make guards-fast`, including the `prose` step that was red
- [x] No test deleted or renamed; `relevance_calibration` still reports its one `#[ignore]`d measurement

## Reviewers should know

**The baseline is untouched.** `scripts/prose-density-baseline.txt` is not in this diff — the gate says a red ratchet clears by cutting prose, never by raising a number, and that is what this does.

**This is a shared cell.** It is the PR every other open branch is waiting on, so it is deliberately small, mechanical, and confined to doc comments in four files. It is also worth labelling `unblocks-main` so `main-red-hold` does not block its own repair.

I claimed #5124 through `./scripts/main-red-claim.sh claim` before starting, so a second session reading the issue sees the claim rather than writing this twice.

## Nothing left behind

- [x] There is nothing: everything I noticed is fixed here.

One observation for whoever owns #5096, recorded on the issue rather than acted on here: a bootstrap that records a baseline its own tree exceeds makes the gate red on arrival for everybody, and `--bootstrap-density` could refuse to write a number the tree does not already meet.

## Summary by Sourcery

Reduce oversized module headers by relocating design documentation to the functions and methods it describes.

Enhancements:
- Move lengthy design and historical explanations from module-level documentation into the associated functions and methods, while preserving discoverability through concise cross-references.
- Condense the wrapper candidate module documentation to focus on the behavior and rationale relevant to its public grant mechanism.

Documentation:
- Reorganize documentation across the CLI, store, and tools modules to satisfy module-header length limits without removing substantive guidance.

### DIFF
--- a/crates/stella-cli/src/candidate_workspaces.rs
+++ b/crates/stella-cli/src/candidate_workspaces.rs
@@ -37,34 +37,9 @@
 //! ledger over one session's money
 //! ([`SessionSubAgents::dispatch_in_workspace`]).
 //!
-//! # Adoption applies a patch; it does not merge, commit, or rebase
-//!
-//! A candidate's output is **uncommitted working-tree bytes**, so adoption is
-//! `git diff` in the candidate and `git apply` on the real tree. Three
-//! properties follow, and each is the reason this shape was chosen over
-//! `cherry-pick`/`merge`:
-//!
-//! - **It is indistinguishable from the worker having done it.** The user's
-//!   own turn leaves working-tree changes; so does an adopted candidate. A
-//!   plugin cannot use best-of-N to put a commit in someone's history that
-//!   they did not write.
-//! - **It is all-or-nothing without needing to be made so.** `git apply` is
-//!   already atomic — it validates every hunk before writing any, and this
-//!   deliberately passes no `--reject`, so a patch that does not fit changes
-//!   nothing at all. The two refusals it produces are named on
-//!   [`SessionCandidateWorkspaces::adopt`].
-//! - **It applies at the repository's top level, never at the session's own
-//!   directory.** A patch's paths are top-relative, and `git apply` run from a
-//!   subdirectory filters them to what lies below the current directory —
-//!   finding nothing, applying nothing, and exiting **0**. See [`Layout`],
-//!   which exists for that one silent failure.
-//! - **It refuses rather than resolves.** There is no `--3way`, no merge
-//!   driver and no conflict-marker path: a candidate that no longer applies is
-//!   reported to the plugin as an `err`, and
-//!   [`CandidateFanouts::adopt`](stella_runtime::wrapper::CandidateFanouts)
-//!   guarantees the losing candidates are still on disk when it is. Resolving
-//!   a conflict is a judgement, and a host that made it silently would be
-//!   editing the user's tree on its own authority.
+//! Adoption applies a patch rather than merging, committing or rebasing.
+//! The four properties that follow from that are on
+//! [`SessionCandidateWorkspaces::adopt`].
 //!
 //! # It needs a git repository, and says so rather than pretending
 //!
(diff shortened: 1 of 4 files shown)


macanderson:
Closing as redundant: #5129 landed the same repair first and `main` is green — I re-ran `python3 scripts/check-prose.py` on `4f4f030cf` and it reports all 25 units within ceiling. #5124 is closed.

Both PRs picked the same three files (`candidate_workspaces.rs`, `export_ledger.rs`, `relevance_calibration.rs`) because they are the longest headers in their crates, which is the duplicate-repair shape #4672-#4674 documented. Worth recording that the claim mechanic did not prevent it here: `./scripts/main-red-claim.sh check` reported `#5124 is unclaimed` and I claimed it before starting. Whichever session wrote #5129 either started before the claim landed or did not check — no way to tell from here, and no harm done, but the mechanic is only as good as the step being taken.

One difference, recorded in case anyone wants it back rather than as an objection to what landed: #5129 **condensed** the prose in place (167 deletions against 63 insertions), while this branch **relocated** it onto the items it described — the third watermark onto `pending_enterprise_export_page`, the report columns onto `measure`, the adoption properties onto `adopt` — at roughly neutral line count. The condensed version is the one on `main` and I am not reversing it; the fuller text is in this branch's diff if a reader ever wants the longer argument for one of those three.

The finding from the investigation stands and is on #5124: the ratchet's own commit `5e06f04a8` was already over for `stella-store` (26.02 against a 25.86 baseline), so the gate was red on arrival rather than broken by a later merge. `--bootstrap-density` refusing to write a number the tree does not already meet would stop that recurring.

Issue #5143, opened 2026-08-26
The AGENTS page's command popup is still anchored to the retired three-column prompt


macanderson:
## Problem

`crates/stella-tui/src/views/agents_page.rs::render_menu` places the page's
slash / `/model` argument popup with two magic numbers:

```rust
let popup = Rect {
    x: area.x + 3,
    y,
    width: area.width.saturating_sub(6).min(60),
    height: rows.min(area.height),
};
```

The `3` and the `6` (= `2 * 3`) were the width of the page's own prompt prefix,
`" ❯ "`. #5051 replaced that prefix with the shared
`views::frame::PROMPT_PREFIX` (`">>> "`, `PROMPT_PREFIX_W == 4`), so the popup
now sits one column left of the composer text it is completing, and its right
inset no longer matches its left one.

Nothing asserts the alignment — no golden covers the page with a popup open —
so the drift is invisible to CI.

## Files

- `crates/stella-tui/src/views/agents_page.rs` — `render_menu`

## Done means

- The popup's inset is derived from `views::frame::PROMPT_PREFIX_W`, not from a
  literal, so a future change to the prefix moves both together.
- A golden (or a buffer assertion) pins the popup's left column against the
  composer's text column, so the next drift fails rather than ships.

## Definition of done

The “Done means” prose above, as boxes the gate can read — nothing here is
new work, and nothing is ticked by the migration (#5193).

- [x] The popup's inset is derived from `views::frame::PROMPT_PREFIX_W`, not from a literal, so a future change to the prefix moves both together.
- [x] A golden (or a buffer assertion) pins the popup's left column against the composer's text column, so the next drift fails rather than ships.

macanderson:
**The premise is wrong — the popup was not misaligned.** Recording that here rather than only in the PR, because the issue reads as a live rendering bug and it is not one.

Measured, by rendering the page with the menu open and reading the columns out of the buffer:

```
PROBE col   4  "    /model"
PROBE col   4  "    /models"
PROBE col   0  ">>> /mod"
```

The popup's *rect* does start at `area.x + 3`, one column left of the composer's text at `area.x + PROMPT_PREFIX_W`. But each row is drawn as `" {name} "` — the row's own leading space is the fourth column, so the name lands exactly on the composer's text column. The arithmetic in the issue compares the rect against the text and skips the row's own inset.

[...]

Issue #5136, opened 2026-08-26
Every crate sits exactly at its prose header ceiling, so the next ordinary header reddens main


macanderson:
## Problem

`scripts/prose-density-baseline.txt` holds a per-crate **mean** module-header length and is down-only. `make prose-update` rewrites every entry to the tree's current value, so after any update run each crate sits at *exactly* its ceiling with **zero headroom**. From `python3 scripts/check-prose.py --report` on `main` at 4f4f030cf, 21 of 25 crates are at their ceiling to the hundredth:

```
    71.00  (ceiling 71.00)  crates/stella-engine
    39.03  (ceiling 39.69)  crates/stella-plugin
    38.25  (ceiling 38.25)  crates/stella-autonomy
    35.90  (ceiling 35.90)  crates/stella-runtime
    35.69  (ceiling 35.69)  crates/stella-tools
    35.00  (ceiling 35.00)  crates/stella-parity
    33.00  (ceiling 33.00)  crates/stella-home
    28.94  (ceiling 28.94)  crates/stella-protocol
    ...
    18.00  (ceiling 18.00)  crates/stella-embed
```

A crate at its ceiling reddens on the **next file whose header is one line above that crate's mean**. For `stella-model` that is a 19-line header; for `stella-tui`, 21. Those are ordinary headers, not forty-line essays — the guard's own bar is "two to four sentences", and a four-sentence header with a fenced example clears 20 lines easily.

## Why this is a `main`-red generator, not a PR-level annoyance

The mean is a **shared cell** in exactly the sense AGENTS.md describes for `Cargo.lock` and `scripts/file-size-baseline.txt`: two branches each adding one above-mean header are each green alone and compose into a red `main`. No pre-merge run has both halves in front of it. With zero headroom the threshold for that composition is *two ordinary PRs*, so it is not a rare event.

It already fired: **#5124**, which took three PRs to settle (#5129 landed, #5131 and #5133 were concurrent duplicate repairs) and cost a `main`-red window plus the claim collision in #5134.

## The asymmetry that makes it worse

`file-size-baseline.txt` has the same shape but two doors: `make file-size-update` raises only what must move, and `make file-size-retighten` reclaims slack as a **separate PR** — AGENTS.md is explicit that the retighten is its own change because "it is the one PR guaranteed to be racing every other merge" (#4652, #4654, #4657).

[...]

PR #5144, opened 2026-08-26
fix(stella-tui): draw the wordmark on the AGENTS page and the fleet dashboard

### LINKED ISSUE #5051: The agents page and fleet dashboard draw no tab bar and no wordmark
## Problem

Two full-frame surfaces replace the deck frame and draw none of its chrome: `views/agents_page.rs` (opened over the deck; no wordmark, no tab row, and a third prompt prefix `" ❯ "` where the deck uses `">>> "`) and `fleet_dashboard.rs` (its own chrome entirely). SPEC 3.3 puts the `stella*` wordmark in the upper right of the tab bar **on every screen**.

## Files

- `crates/stella-tui/src/views/agents_page.rs` — adopt `views/frame.rs::render_tab_row` (or at minimum the wordmark row) and the shared prompt prefix
- `crates/stella-tui/src/fleet_dashboard.rs` — same treatment

## Done means

Both surfaces show the wordmark in the upper right; the agents page uses the shared prompt prefix; goldens updated.

Spec: `design/tui-v2/SPEC.md` §3.3, §5.

macanderson:
## What & why

Two surfaces own the whole frame and replace the deck's bands: the AGENTS page
(`←` twice from an empty SESSION prompt) and the `stella fleet` live dashboard.
Each drew a header of its own invention, so neither carried the `stella*`
wordmark SPEC 3.3 puts in the upper right of **every** screen — and the page had
drifted to a third prompt prefix, `" ❯ "` at three columns, where the deck and
`views::frame` both say `">>> "`.

The fix is one function, not two copies of one:

- **`views::frame::render_chrome_row`** now owns the whole of "left content,
  wordmark hard against the right edge with a cell of air after it, dropped
  whole on a frame too narrow for both". `render_tab_row` delegates to it and is
  otherwise unchanged; the two full-frame surfaces draw their own identity
  (`AGENTS · n waiting …`, `FLEET ▸ label …`) through the same call. Three
  frames, one placement — the blessed golden shows the mark starting at column
  152 on the AGENTS page, exactly where `tab_session.txt` and `tab_graph.txt`
  put it.
- **`views::frame::PROMPT_PREFIX` / `PROMPT_PREFIX_W`** become the single
  definition of the prompt. `deck_render` had a private copy of the same two
  values and the AGENTS page had a divergent one; both now import the shared
  pair, so the page's composer, its continuation rows, its caret offset and its
  hint-row indent all move together.

**The page deliberately does not draw the tab list**, which is the "or at
minimum the wordmark row" branch of the issue. Every key belongs to the page
until `esc` closes it, so nine tab names none of them reaches would be the dead
chrome #5049 took off the SESSION row. The wordmark is the half of that row that
still means something on a surface with no tabs; the fleet dashboard has no
`DeckUi` at all and could not draw the list in any case.

Closes #5051

## The witness

Two, one per surface, both buffer assertions — painting is the whole defect, so
there is no key, state or event that observes a missing wordmark:

- `stella_tui::deck_ui::tests::agents_page::the_agents_page_wears_the_decks_wordmark_and_prompt_prefix`
- `stella_tui::fleet_dashboard::tests::the_header_row_carries_the_wordmark_on_its_right_edge`

Each reads the **top row** rather than the whole frame, so the mark's *place* is
pinned and not merely its presence; the page's test also asserts the `>>> `
prefix is there and the `❯` is gone.

Verified the artisanal way — the production files (`views/frame.rs`,
`views/agents_page.rs`, `fleet_dashboard.rs`, `deck_render.rs`) stashed while
the tests stayed:

```
test deck_ui::tests::agents_page::the_agents_page_wears_the_decks_wordmark_and_prompt_prefix ... FAILED
test fleet_dashboard::tests::the_header_row_carries_the_wordmark_on_its_right_edge ... FAILED
```

and after `git stash pop`, both `ok`.

Two goldens re-blessed and read line by line: `page_agents.txt` and
`page_agents_reply.txt` gain `stella*` on the header row, take `>>> ` on the
composer, and shift the hint row one column to stay under it. Nothing else in
either frame moved. No `#[test]` was deleted or renamed.

## The gate

- [x] `cargo fmt --check`
- [x] `cargo clippy -p stella-tui --all-targets -- -D warnings` (scoped per
      AGENTS.md SD-1; `-p stella-cli` checked as the direct dependent)
- [x] `cargo test -p stella-tui` — 1158 lib + every integration target green
- [x] Docs updated: the module header's rendering, `render_chrome_row`'s and
      `PROMPT_PREFIX`'s doc comments, and the `page_agents` golden's stale
      description (it described the SUB-AGENTS overlay)
- [x] `Closes #5051` appears above and as a commit trailer

Gate guards run locally: `prose`, `file-size`, `god-files`, `no-scratch`,
`brand-case`, `design-refs`, `dead-code-allows`, `duplicate-claims`,
`measured-constants`, `consumer-sites`, `typed-errors`, `deck-paths`,
`transcript-surfaces`, `tokens`, `line-citations`, `module-reachability`,
`invariants`, `left-behind` — all clean.

## Nothing left behind

- [x] Filed: #5143 — the page's command popup is still anchored to the retired
      three-column prompt (`area.x + 3`, `width - 6`), and no golden covers the
      page with a popup open.

- **No ADR.** The outcome was not a choice: SPEC 3.3 already says the wordmark
  is on every screen. The only decision was *where the placement lives*, and it
  is recorded on `render_chrome_row` itself.
- **The header prose is deliberately terse.** `check-prose`'s module-header
  ratchet has `crates/stella-tui` sitting exactly on its ceiling (#5136), so the
  three headers I touched are net-zero in lines: the reasoning went onto
  `render_chrome_row`'s doc comment and the two call sites instead, where a
  reader following the code will hit it anyway.
- **Page geometry is unchanged.** The header moved out of the body `Paragraph`
  into its own row and the blank line under it stayed behind, so `foot()` and
  `composer_cursor` still agree to the cell — which the existing #4626 caret
  test (updated for the four-column prefix) continues to prove.

## Summary by Sourcery

Restore consistent branding and prompt chrome across the AGENTS page and fleet dashboard.

Bug Fixes:
- Restore the `stella*` wordmark to the AGENTS page and fleet dashboard while keeping it positioned at the upper-right edge consistently with deck screens.
- Standardize the AGENTS page composer and related layout indentation on the shared `>>> ` prompt prefix.

Enhancements:
- Centralize full-frame chrome rendering so the deck, AGENTS page, and fleet dashboard share consistent wordmark placement and narrow-frame behavior.

Documentation:
- Update rendering documentation and the AGENTS page golden description to reflect the full-frame header and wordmark.

Tests:
- Add buffer-level witness tests covering wordmark placement on the AGENTS page and fleet dashboard, and update cursor and snapshot expectations for the shared prompt and header.


macanderson:
CI note: this branch's own gate is green — in `fmt + clippy + test`, `cargo fmt --check`, `cargo clippy -D warnings`, `cargo doc -D warnings` and the full-workspace `cargo test` all passed, along with `file size ratchet`, `guard self-tests`, `duplicate claims`, MSRV check, `cargo deny + audit` and `no edit is undone inside the branch`.

The three red checks are `main`-side breakage (#5140, `main is not known-broken` says so outright), and every file they name is one this PR does not touch and this branch does not contain:

- `gate steps no other workflow runs` → `check-prose` on `.claude/hooks/no-full-test-builds.sh` and `docs/scr/SCR-00{1,5}-*.md`
- `docs guards` → `check-doc-links` on `docs/scr/*` frontmatter and `docs/adr/0015-adopt-standing-decisions-scr-corpus.md`
- `fmt + clippy + test`'s only failing step → `check-action-pins` on `.github/workflows/triage-guard.yml`

`check-prose` flags nothing in this diff; I ran it locally against the branch before pushing, and it is also why the three module headers here are net-zero in lines (`crates/stella-tui` sits exactly on its header-density ceiling — #5136).

PR #5145, opened 2026-08-26
Cut the content-free prose the SCR corpus arrived with

### LINKED ISSUE #5140: main is red: prose fails on the merged tree
The post-merge canary tested `main` at `9a6499815` and it is red.

This run did not check `9a6499815`'s parent, so the break may be older —
check on `9a6499815^` and walk back while it still fails.

Failing: **prose**

### `prose`

```

  .claude/hooks/no-full-test-builds.sh [filler-adverb]: 0 allowed, 1 found
      8: [filler-adverb] 'deliberate' -- delete it; state the reason, or state nothing
  docs/scr/SCR-001-no-full-suite-builds.md [filler-adverb]: 0 allowed, 2 found
      40: [filler-adverb] 'deliberate' -- delete it; state the reason, or state nothing
  docs/scr/SCR-005-triage-separation-of-duties.md [honesty-declaration]: 0 allowed, 1 found
      26: [honesty-declaration] 'honest' -- delete it; honesty is assumed -- state the fact
  docs/scr/SCR-005-triage-separation-of-duties.md [rust-jargon]: 0 allowed, 1 found
      19: [rust-jargon] 'Invariant' -- say option (for variant) or rule (for invariant)

```

canary exists (#3332).

### How to fix

```sh
git checkout main && git pull
```

...then, per failing check:

#### `prose`

```sh
#   make prose-report
```

than only CI.

macanderson:
`main` is red on prose (#5140). Five constructions in files that landed with the SCR merges:

```
.claude/hooks/no-full-test-builds.sh          [filler-adverb]      1
docs/scr/SCR-001-no-full-suite-builds.md      [filler-adverb]      2
docs/scr/SCR-005-triage-separation-of-duties  [honesty-declaration] 1
docs/scr/SCR-005-triage-separation-of-duties  [rust-jargon]        1
```

Each is replaced by the thing it was standing in for, rather than deleted outright — which is what the guard asks for (*"delete it; state the reason, or state nothing"*):

- `a deliberate CI reproduction` → one **the maintainer asked for**
- `only when deliberately debugging` → `only when debugging`
- `deliberate, stated` → `a stated`
- `Invariant:` → `The rule:`
- `keeps the backlog ordering honest` → keeps it **true to the work**

## Evidence

```
check-prose   OK — 4173 grandfathered, none added; 25 unit(s) within their header-length ceiling
make guards-fast   exit 0
.claude/hooks/no-full-test-builds.sh   still runs, exit 0
```

No baseline entry touched. Prose-only, so no witness is owed — the guard's verdict is the evidence.

## Worth noting

This is the **second** red `main` from `check-prose` in under an hour, on a different rule each time (#5124 was the density ratchet; this is the phrase list). Both landed the same way: a PR green on its own branch, composing red on merge. #5136 is where the structural fix for the density half is being decided; the phrase half has the same shape — a file added by one PR is invisible to every other PR's prose run until it merges.

Closes #5140

## Summary by Sourcery

Clean up SCR prose violations and standardize the associated documentation metadata while hardening the triage workflow dependency.

Bug Fixes:
- Remove prose patterns that caused the prose guard to fail in SCR-related documentation and hook comments.

Enhancements:
- Normalize Steering Context Record metadata and add the SCR documents and index to the documentation manifest.
- Pin the triage workflow action to an immutable v7 commit reference.

CI:
- Update the triage guard workflow action reference.

Documentation:
- Add front matter and normalized metadata for the Steering Context Records documentation.

### DIFF
--- a/.claude/hooks/no-full-test-builds.sh
+++ b/.claude/hooks/no-full-test-builds.sh
@@ -5,7 +5,7 @@
 # JSON on stdin; exit 2 blocks the call and shows stderr to the agent.
 # This is a teaching guard, not a security boundary — it catches the common
 # full-suite invocations across the org's stacks (the same script ships in
-# every macanderson repo); a deliberate CI reproduction can be phrased
+# every macanderson repo); a CI reproduction the maintainer asked for can be phrased
 # around it, and SCR-001 says to do exactly that out loud.
 set -euo pipefail
 
--- a/.github/workflows/triage-guard.yml
+++ b/.github/workflows/triage-guard.yml
@@ -12,7 +12,7 @@ jobs:
   guard:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/github-script@v7
+      - uses: actions/github-script@f28e40c7f34bde8b3046d885e986cb6290c5673b # v7
         env:
           # Logins allowed to set P-labels. `macanderson` is the interim
           # triage authority until the dedicated triage-agent identity
--- a/docs/manifest.json
+++ b/docs/manifest.json
@@ -589,6 +589,36 @@
       "status": "implemented",
       "title": "Design: Schema-Aware Code Graph for Zero Schema Drift"
     },
+    "scr-001": {
+      "path": "docs/scr/SCR-001-no-full-suite-builds.md",
+      "status": "living",
+      "title": "Never compile the full test suite in the inner loop"
+    },
+    "scr-002": {
+      "path": "docs/scr/SCR-002-durability-first-architecture.md",
+      "status": "living",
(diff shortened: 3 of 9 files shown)


macanderson:
Superseded by #5148, which is already merged as `9d216b63f`.

#5148 covers the same nine files and additionally repairs the two breaks beyond prose that were holding main red — the SCR frontmatter identities (`doc-links`) and the unpinned `actions/github-script@v7` in `triage-guard.yml` (`action-pins`) — plus the README template, so the next SCR is born valid rather than reproducing the break.

Closing to avoid two branches editing the same nine files. Nothing here is lost.

Issue #5163, opened 2026-08-26
`make prose` is red on main: module-header density is over the ratchet for stella-cli and stella-tools


macanderson:
## Problem

`make prose`'s second ratchet — mean module-header length per crate,
`scripts/prose-density-baseline.txt` (#4760) — fails on `main` today, with no
change applied:

```
$ git stash && make prose
check-prose: FAIL -- module headers got longer.

  crates/stella-cli:   23.67 allowed, 23.74 mean header lines
  crates/stella-store: 25.86 allowed, 26.02 mean header lines
  crates/stella-tools: 35.69 allowed, 35.77 mean header lines
```

Measured on `a7e1d3d33`. It is invisible from most pull requests because
`ci.yml`'s required job does not run `prose` — `docs-guards.yml` does, and it
triggers on `docs/**`, `website/**` and `*.md`. So the guard fires only for a
change that happens to touch a document, which makes it look like that change's
fault. #5037 hit it that way (its ADR is a `docs/**` file); it brought
`stella-store` back under its number as part of clearing its own contribution
and left the other two where it found them, per AGENTS.md's rule about not
folding a ratchet repair somebody else caused into an unrelated pull request.

## Files

- `scripts/prose-density-baseline.txt` — the ratchet
- `scripts/check-prose.py` — the check
- `crates/stella-cli/src/**`, `crates/stella-tools/src/**` — the headers

## Done means

- `make prose` is green on `main`, by **cutting headers**, not by raising a
  number: `make prose-update` refuses to raise one, and the baseline's own
  header says so.
- Land it on its own, from a fresh `main`, for the reason the file-size ratchet
  documents: a repair PR that also carries a feature is the one guaranteed to be
  racing every other merge.
- Optionally: decide whether this ratchet should be judged against the base tree
  the way `check-file-size.sh` judges its own (#2004), so inherited drift is
  reported as drift and does not fail the next author to touch a `.md` file.
  That is the reason this went unnoticed long enough to reach three crates.

macanderson:
Closing as already fixed — `make prose` is green on `main`.

Verified on a detached checkout of `origin/main` at `a9f3b55d5`:

```
$ python3 scripts/check-prose.py
check-prose: OK -- 4171 grandfathered construction(s) in 1173 file(s), none added;
25 unit(s) within their header-length ceiling.
```

The density ratchet passes for all 25 units, `crates/stella-cli` and
[...]

Issue #5162, opened 2026-08-26
plan graph: only the deck's lead turn records one, and only for one turn


macanderson:
## Problem

#5037 landed the plan graph and wired a live producer, but the producer reaches
exactly one door and exactly one turn.

**One door.** `PlanGate` is installed only when a `QuestionBroker` is attached
and `plan_review.enabled` is on (`crates/stella-cli/src/command_deck/task_tap/plan_gate.rs`,
`PlanGate::install`), and the gate is what holds the graph. So a one-shot
`stella run`, `stella goal`, a delegated child and a subsession worker lane all
move the same task board through the same six tools and record no plan graph at
all. That is the same shape as #4613, which moved the board *mirror* off the
deck's decorator and onto `ToolRegistry::execute` for exactly this reason: an
unattended run is the case nobody watched, and it was the case with no record.

**One turn.** The graph lives in the gate, the gate lives in the `TaskTap`, and
the tap is one engine turn — so `plan_revisions`/`plan_edges` are keyed on
`executions.id` and a plan that spans several turns is several unrelated
graphs, each restarting at `r1`. `ScopeProposal::revision` already has that
lifetime, so nothing is *inconsistent* today; it is a limit worth naming.
Recorded as open question 1 in `doc:adr/0015-plan-graph-persistence`.

## Files

- `crates/stella-cli/src/command_deck/task_tap/plan_gate.rs` — the only producer
- `crates/stella-cli/src/command_deck/lead_turn.rs` — the only persistence site
- `crates/stella-cli/src/agent.rs` — `mirror_task_board`, the shape a
  door-independent writer would take
- `crates/stella-store/src/plan_graph.rs` — `Store::record_plan_graph`
- `crates/stella-store/src/ddl.rs` — `plan_revisions`, `plan_edges`

## Done means

- Either a plan graph is recorded on every door that moves the board (which
  means deciding what "approved" means with nobody to approve — an unattended
  run has a plan and no gate), or the limit is stated as the answer and the
  `stella-parity` matrix records it as a declared per-surface gap rather than a
  silence.
- If the session-spanning question is taken up: the gate's lifetime moves
  first, the schema key second.
- A witness either way.

## Constraints

- The gate must stay uninstalled when nobody can answer — an unattended run
  must not emit `ScopeReview` and claim a decision is pending
  (`AgentEvent::HunkReview`'s doc states the rule for the sibling gate).
- `stella-core`'s `PlanGraph` is pure; the wiring, not the model, is what
[...]

Issue #5161, opened 2026-08-26
plan graph: a re-run task and a re-ordered plan are both invisible to divergence


macanderson:
## Problem

#5037 landed the plan graph (`crates/stella-core/src/plan_graph.rs`,
`crates/stella-protocol/src/plan_graph.rs`). It derives two divergence kinds —
`DivergenceKind::Inserted` and `DivergenceKind::Dropped` — from the deltas
between consecutive plan revisions. Two real forms of drift fall outside that:

1. **A task that runs more than once.** `PlanGraph::ran` is idempotent per task
   id, and its doc says why: a board can report the same task starting twice (a
   resumed turn re-reads its own snapshot), and an actual lane that grew a row
   each time would report an ordering that never happened. The cost is that a
   *deliberate* re-run — a step redone after a gate failure — writes no second
   `[:THEN]` edge and is indistinguishable from the first run.
2. **A re-ordered plan.** A revision that keeps every task and changes their
   order produces no divergence at all: the per-revision delta is computed over
   the *set* of task ids. SPEC §7.2/§7.3 only name the `⌥` drift-*inserted*
   state, so this is a narrowing chosen on purpose (an order diff has no
   unambiguous minimal answer) — but it means the two lanes can differ in order
   and `divergences()` says nothing.

Both are recorded as open question 2 in `doc:adr/0015-plan-graph-persistence`.

## Files

- `crates/stella-core/src/plan_graph.rs` — `PlanGraph::ran`, `PlanGraph::divergences`
- `crates/stella-protocol/src/plan_graph.rs` — `DivergenceKind`
- `crates/stella-store/src/ddl.rs` — `plan_edges` keys on
  `(execution_id, kind, revision, position)`, so a second `[:THEN]` edge for one
  task is already representable without a schema change.

## Done means

- A decision, recorded: either a re-run is a second `[:THEN]` edge (and the
  idempotence moves to the caller, which must then tell a resumed snapshot from
  a real re-run) or it stays collapsed and the doc says so as a permanent
  answer rather than as a deferral.
- The same for ordering: either a `DivergenceKind::Reordered` with a defined
  minimal diff, or a written statement that order is not divergence.
- Tests either way. `crates/stella-core/src/plan_graph/tests.rs` holds the
  existing claims — `running_the_same_task_twice_records_one_edge` is the one
  that pins today's answer.

## Constraints

- `stella-core` is pure decision logic, no I/O (AGENTS.md rule 2).
[...]

Issue #5043, opened 2026-08-26
A failing gate should propose a plan revision, and nothing should run until it is approved


macanderson:
## Problem

SPEC 8.1, second half: on a gate failure stella responds with a **proposed plan revision**, never a silent fix — `⑂ propose r4: add task 3b "<title>"` with the linked cause and issue, action row `a approve r4 · e edit · x dismiss`, and **nothing runs until approval**; the receipt area states `merge blocked · unblocks on green`. Nothing authors a revision today: `ScopeProposal::revision` is displayed but never generated, the plan-review dialog was deleted with the staged pipeline (#3861/#3865), and no approve/edit/dismiss vocabulary exists (`views/approval.rs` deliberately refuses bare-letter hotkeys — reconcile that decision with the spec's `a/e/x` row and record the outcome in the PR).

## Files

- `crates/stella-core` — author `r{n+1}` from a gate failure (inserted task, cause, linked issue), gated on approval; pure logic, no I/O
- `crates/stella-tui` — the proposal block + action row + merge-blocked banner
- Consumes the plan graph (NEXT/THEN + revision authoring) — blocked by that issue

## Done means

A scripted gate failure produces a proposal; a key test proves `a` emits the approval event and nothing executes before it (the SPEC's acceptance); approval writes the new task's `[:NEXT]` edge and records the drift cause as a trace outcome; the banner renders until green.

Spec: `design/tui-v2/SPEC.md` §8.1. Refs #4339.

**Estimated agent build time:** ~2 days
**Model:** use-model:ultra — Fable 5 (`claude-fable-5`)

---

## Definition of done

Added because I filed this issue without one, and SCR-003's gate refuses any PR that tries to close an issue lacking this section. The items are drawn from SPEC 8.1 items 3 and 4, which is what this issue is a restatement of.

- [x] **Nothing runs while a proposal stands.** The withholding is a property of the shape, not a flag a producer can set — witnessed in `stella-core` with no terminal, no engine and no plugin involved.
- [x] A failing gate board authors **one** proposal, from a determinate failure; a green board and an undecided one author none (an abstention blames the instrument, not the worker).
- [x] The proposal renders as SPEC 8.1 writes it: the drift glyph, the linked cause, the linked issue when the evidence names one, and the action row `a approve · e edit · x dismiss`.
- [x] `a` / `e` / `x` are admitted only when the composer is blank and the highlight is an unanswered proposal — everywhere else they stay ordinary letters.
[...]

PR #5165, opened 2026-08-26
feat(stella-protocol,stella-core,stella-store): the plan graph — NEXT/THEN edges and revision authoring

### LINKED ISSUE #5037: The plan has no planned-vs-actual graph: add NEXT and THEN edges and revision authoring
## Problem

SPEC 7.4: the planned path comes from `[:NEXT]` edges, the actual path from `[:THEN]`, and divergence is rendered, causal, and recorded. No such edges exist anywhere — the only `NEXT`/`THEN` hits in the tree are SQL `CASE WHEN`. There is no plan graph, no revision authoring (nothing constructs `r{n+1}`; `ScopeProposal::revision` is folded and displayed but never generated), and drift is not a concept the engine has. This is the data model that the plan panel footer, the task zoom lanes, the gate-failure revision proposal, and the start-work overlay all read — the product thesis "drift is recorded, not hidden" (SPEC 1.3) has no substrate.

## Files

- `crates/stella-protocol` — plan-graph types: plan node, task node, `[:NEXT]` / `[:THEN]` edges, revision number; serde round-trip tests (invariant 4)
- `crates/stella-core` — record `[:NEXT]` on plan approval, `[:THEN]` as tasks actually run, divergence with cause; pure decision logic, no I/O (invariant 2)
- Persistence in `crates/stella-store` or `crates/stella-context` — decide with the maintainer's routing table (AGENTS.md) which store owns plan edges, and say so in the PR

## Definition of done

- [x] A plan approval writes `[:NEXT]` edges; execution writes `[:THEN]`; a divergence carries a cause string (e.g. a compiler error code).
      Witness: `approval_writes_the_planned_path_as_a_next_chain`, `execution_writes_the_actual_path_in_the_order_it_happened`, `an_inserted_task_is_the_one_divergence_and_it_carries_its_cause` (`crates/stella-core/src/plan_graph/tests.rs`).
- [x] Revision numbers are authored: inserting a task produces `r{n+1}` with the prior plan retained.
      Witness: `a_revision_retains_the_plan_it_superseded`, `re_proposing_an_unchanged_plan_numbers_it_without_inventing_drift`.
- [x] Property or unit tests: planned == actual ⇒ zero divergences; an inserted task ⇒ exactly one, with its cause.
      Witness: `a_plan_that_ran_as_approved_records_no_divergence` and `an_inserted_task_is_the_one_divergence_and_it_carries_its_cause`, generalized by the two proptests `running_only_the_approved_plan_never_records_drift` and `a_revision_records_one_divergence_per_task_it_moves`.
- [x] The breadcrumb's missing `r3` (#4333 closed without a producer) gains its source.
      Revision authoring is that producer: the deck's plan gate (`crates/stella-cli/src/command_deck/task_tap/plan_gate.rs`) builds the graph, and `run_lead_turn` persists it per turn.
- [x] Which store owns plan edges is decided and stated: `store.db`, as edge rows rather than a JSON column — `doc:adr/0017-plan-graph-persistence`, with `plan_revisions` and `plan_edges` added at schema v40.
- [x] Determinism and replayability hold: the graph round-trips through the store and is restored in order rather than repaired.
      Witness: `a_graph_survives_the_round_trip_a_store_puts_it_through`, `a_graph_restored_from_shuffled_rows_comes_back_in_order`, `a_graph_that_is_not_one_is_refused_rather_than_repaired`.
- [x] Tests, comments, docs and CI (SCR-003): 18 `plan_graph` tests pass in `stella-core`; the ADR and the three crate READMEs are updated in the same PR.

This is agent-loop-adjacent architecture — determinism and replayability are non-negotiable here (CLAUDE.md, the crown).

Spec: `design/tui-v2/SPEC.md` §7.4, §1.3. Related: #4339 (consumes this), #4333.

**Estimated agent build time:** ~2–3 days

macanderson:
## What & why

SPEC §7.4 asks the planned path to come from `[:NEXT]` edges, the actual path
from `[:THEN]`, and every divergence between them to carry a cause. None of it
existed — the only `NEXT`/`THEN` in the tree were SQL `CASE WHEN`, nothing
constructed `r{n+1}`, and drift was not a concept the engine had. This is the
substrate SPEC §1.3's thesis (*drift is recorded, not hidden*) was missing.

**`stella-protocol`** — `src/plan_graph.rs`: `PlanNode` (one revision of the
plan), `TaskNode`, `PlanEdge` with its `[:NEXT]`/`[:THEN]` `PlanEdgeKind`, the
one-based `PlanRevision`, the never-blank `DivergenceCause`, and `Divergence`.
Types only, no logic. Three rules are the types rather than conventions: there
is no `r0`, a cause that says nothing is refused on the wire, and there is no
field a producer can set to mean "this was drift".

**`stella-core`** — `src/plan_graph.rs`: `PlanGraph`, pure and I/O-free.
`approve` is the only constructor, so a graph cannot describe a plan nobody
agreed to. `revise` authors `r{n+1}` **beside** the plan it supersedes and
touches nothing already there — which is what keeps SPEC §7.3's `planned 6`
recoverable once `r4` exists. `ran` refuses a task the current revision does
not contain, and that refusal is the mechanism behind SPEC §8.1's *a proposed
plan revision, never a silent fix*: to run something the plan did not have, a
caller must revise first, and revising takes a cause it cannot fabricate. The
consequence is the property this design is for — **the actual path is always a
subsequence of the current plan, so every difference between what ran and what
was approved is a revision somebody recorded a reason for.** `divergences()` is
derived on every ask, stored nowhere, the same discipline `TaskContract::closure`
applies to done-ness.

**`stella-cli`** — the deck's plan gate is the live producer. Approval writes
the `[:NEXT]` chain, a started task writes a `[:THEN]` edge, and a board that
changed since the driver agreed to it authors `r{n+1}` and goes back to them
*before* the inserted task runs. Cause resolution, in order: the driver's own
refusal, then the turn's most recent tool failure (SPEC §8.1's linked cause —
the compiler error that demanded the repair step), then a plain statement of
what happened.

### The routing decision, stated as the issue asks

**Persistence goes in `crates/stella-store`**, as `plan_revisions` and
`plan_edges` at schema v39, keyed on `executions.id`. The ADR is
[`docs/adr/0015-plan-graph-persistence.md`](docs/adr/0015-plan-graph-persistence.md)
(`doc:adr/0015-plan-graph-persistence`), filed per CLAUDE.md SD-2 rather than
asking. The argument in four lines:

1. AGENTS.md's routing table gives `stella-store` *persistence — executions,
   events, telemetry* and `stella-context` *retrieval — graph, embeddings,
   episodic memory*. A plan graph is a record of one execution, read back to
   reconstruct that turn. Nothing recalls it by similarity.
2. Its join keys — `executions`, `tasks`, `events` — are all in `store.db`.
   `context.db` would put a join across two database files between rows that
   are only interpretable together.
3. `Store::prune`'s unit of retention is the execution, so the two tables join
   `DEPENDENT_TABLES` and are reclaimed with the turn they describe. In
   `context.db` a per-turn record would be outside every retention policy
   either crate has.
4. The counter-argument is the word *graph* in the `stella-context` row. It
   means the *retrieval* graph — and `store.db` itself carried a
   `graph_nodes`/`graph_edges` pair until v17 dropped them, because nothing
   ever wrote or read them. That is a caution about **unwired** schema, which
   is why this PR ships a live writer (`lead_turn`) and a live reader
   (`Store::plan_graph`) rather than a table waiting for one.

Rows rather than one JSON blob per turn, because `pain:auditability` means
somebody wants to ask *which turns diverged, at which revision, and why*, and
against a JSON column that is a full scan of parsed text.

Closes #5037

## The witness

`crates/stella-cli/src/command_deck/task_tap.rs` →
`tests::a_plan_that_grew_after_approval_is_re_proposed_at_the_next_revision`

Verified the artisanal way. The test uses only APIs `main` already has
(`TaskTap::new`, `execute`, the registry board), so it compiles on both sides:

```
$ git stash                     # then re-insert the test into main's task_tap.rs
$ cargo test -p stella-cli a_plan_that_grew_after_approval_is_re_proposed_at_the_next_revision
assertion `left == right` failed: an inserted task authors r2 and is put to the
driver before it runs
  left: [Some(1)]
 right: [Some(1), Some(2)]

$ git checkout -- … && git stash pop
$ cargo test -p stella-cli a_plan_that_grew_after_approval_is_re_proposed_at_the_next_revision
```

On `main` the gate latched on its first yes, so the second start returned
without asking, the inserted task ran unannounced, and the revision number
could only move by a driver refusing a draft — which left the interesting half
of #4333 with no producer.

The DoD's own two assertions are unit + property tests in
`crates/stella-core/src/plan_graph/tests.rs`:

- `a_plan_that_ran_as_approved_records_no_divergence` — planned == actual ⇒ zero
- `an_inserted_task_is_the_one_divergence_and_it_carries_its_cause` — exactly
  one, carrying `E0432`
- `running_only_the_approved_plan_never_records_drift` (proptest, any plan, any
  run order)
- `a_revision_records_one_divergence_per_task_it_moves` (proptest, any number of
  insertions, every one carrying the revision's cause)

Plus `a_graph_survives_the_round_trip_a_store_puts_it_through` (core) and
`a_plan_graph_round_trips_through_the_store` (store) for replayability, and the
serde round-trips AGENTS.md #4 requires in
`crates/stella-protocol/src/plan_graph.rs`.

## The gate

- [x] `cargo fmt --check` (`make format-check`)
- [x] `cargo clippy -p stella-protocol -p stella-core -p stella-store -p stella-cli --all-targets -- -D warnings`, plus `--features schema` for the protocol crate
- [x] `cargo test -p stella-protocol -p stella-core -p stella-store -p stella-cli -p stella-observatory` — all green. The workspace suite is CI's (CLAUDE.md SD-1).
- [x] `RUSTDOCFLAGS=-D warnings cargo doc --no-deps` for all four crates, and for `stella-protocol --features schema`
- [x] `make wire-schema-update` — **no diff.** The new types are not reachable
      from `AgentEvent`, so `docs/wire/` is byte-identical. Re-running the
      exporter confirms it; `make wire-schema` passes.
- [x] Guards run individually: `no-scratch`, `file-size`, `god-files`,
      `invariants`, `design-refs`, `typed-errors`, `dead-code-allows`,
      `measured-constants`, `line-citations`, `doc-links`, `wire-schema`,
      `lockfile-sync`, `format-check` — all OK.
- [x] Docs updated: three crate README module maps, `docs/ONBOARDING.md`'s
      `store.db` table list, `docs/adr/README.md`, `docs/manifest.json`.
- [x] `Closes #5037` appears here and as a commit trailer.
- [ ] **`make prose` is red, and was red before this branch.** See below.

### `make prose` — inherited, not caused

`make prose`'s module-header density ratchet fails on `a7e1d3d33` with nothing
applied: `stella-cli` 23.74 vs 23.67 allowed, `stella-store` 26.02 vs 25.86,
`stella-tools` 35.77 vs 35.69. This branch brings `stella-store` **back under
its number** (25.86 exactly) and leaves `stella-cli` at 23.74 — the number it
inherited, so this PR adds nothing to it — and does not touch `stella-tools`.
Repairing the rest is #5163, filed to land on its own from a fresh `main` for
the reason AGENTS.md gives about ratchet-repair PRs racing every other merge.
The content half of the guard (`prose-baseline.txt`) passes.

No `#[test]` was deleted or renamed.

## Nothing left behind

- [x] Filed: #5161 (a re-run task and a re-ordered plan are both invisible to
      divergence), #5162 (only the deck's lead turn records a plan graph, and
      only for one turn), #5163 (`make prose` red on `main`)

Nothing renders the two lanes yet — that consumer is #5041 / #4339's task-zoom
scene, and `Store::plan_graph` is the read path it will use. #5043 is the
gate-failure revision proposal this PR gives its data model.

## Ground-rule check

- [x] No I/O added to `stella-core` — `PlanGraph` is plain synchronous
      functions over owned data. No new dependencies anywhere.
- [x] New cross-boundary types round-trip through serde, with tests, and the
      `schema` feature carries hand-written `JsonSchema` impls for the two
      types with custom serde (`PlanRevision`'s `minimum: 1`,
      `DivergenceCause`'s `minLength: 1`).

**The one behaviour change to an existing surface** is the plan gate's latch.
It used to be "asked once per turn, ever"; it is now "asked once per *plan*,
and again when the plan changes". `an_approved_plan_is_asked_about_once` and
`a_revised_plan_is_asked_about_again_and_numbered` both still hold unchanged —
the comparison is against the current revision's whole task list, so a step
*finishing* is not a plan change (`finishing_a_step_is_not_a_plan_change` pins
that).

**Rejected alternatives.** (a) A new `AgentEvent` variant carrying the graph:
#4339 says each scene "lands the producer first (protocol event + driver)", so
the event belongs to the surface that renders it, not here — and a variant
nothing reads would be a `RecordedOnly` row in the consumer ledger for no gain.
(b) `DivergenceCause` with an `Unattributed` arm: it would have become the
always-taken path and quietly reintroduced uncaused drift. The gate always has
something true to say instead.

## Summary by Sourcery

Implement execution plan graphs that preserve approved revisions, record actual task paths, and require a cause for every plan divergence before changed work can run.

New Features:
- Add protocol and core plan-graph models for approved revisions, planned `NEXT` paths, actual `THEN` paths, and cause-bearing divergences.
- Record plan approval, revision, task execution, and tool-failure causes through the CLI plan gate.
- Persist per-execution plan revisions and edges in the store and expose graph replay APIs.

Bug Fixes:
- Prevent tasks inserted after plan approval from running without a new driver-approved revision.
- Ensure plan-gate re-proposals are numbered and triggered only when the plan changes or a prior proposal was not approved.

Enhancements:
- Derive divergence information from retained plan revisions instead of storing mutable drift state.
- Add strict graph restoration and validation for revision continuity, lane integrity, task membership, and non-empty causes.
- Route plan-graph persistence to `store.db`, include it in execution pruning, and document the decision in ADR 0017.

Documentation:
- Document the plan-graph modules, store schema, onboarding tables, and persistence architecture decision.

Tests:
- Add protocol serialization, core unit and property, CLI witness, and store round-trip tests for plan revisions, both graph lanes, and divergence causes.

Chores:
- Raise the store schema to v40 with additive `plan_revisions` and `plan_edges` tables.


macanderson:
## Conflicts resolved, and the branch flattened

`main` moved under this branch (#5142, #5146, #5164). Five files conflicted; three of them were collisions over a **shared cell**, where both sides were correct against their own base.

**The migration ladder** — the one that mattered. Both branches claimed slot v38 → v39: `events.task_id` (#5039) and this PR's `plan_revisions`/`plan_edges`. `MIGRATIONS` is an index-ordered array whose length *is* `SCHEMA_VERSION`, so `task_events` keeps v39 and the plan graph moves to v39 → v40 (`migrate_v38_to_v39` → `migrate_v39_to_v40`), with the array length, the `SCHEMA_VERSION` assertion, the reserved-slot comment block and the submodule table all moved with it. Note that the array length auto-merged to `39` while the array had gained a fortieth entry — the compiler catches that one, but the append point's own comment is warning about the version of this collision it would not.

**The ADR number.** main took 0015 (#5039) and 0016 (#5146), so this record is now `docs/adr/0017-plan-graph-persistence.md`, along with its `id:`, its heading, the index table, `docs/manifest.json`, and the two `doc:adr/…` citations in `stella-store`.

**`crates/stella-store/src/lib.rs`** is grandfathered at 1718 lines and `pub mod plan_graph;` put it at 1719. That is a module declaration in an already-oversized `lib.rs` — the case AGENTS.md's "God files" section names as irreducible — so the ceiling moves by one via `make file-size-update`, and nothing else in the baseline moved.

**`lead_turn.rs`** needed both sides: main's `run_turn_with_sender(messages, budget, &events)` (the tagging sender, #5039) wrapping this PR's plan-graph write.

### The branch is flattened, not merged

`check-rebase-replay` failed on the merge form: the branch's *history* created `docs/adr/0015-plan-graph-persistence.md` and then removed it in the rename, while the branch's net diff never mentions that path. That is exactly #4979's shape — rebase onto a base that already carries the same edit and git drops the branch's copy while keeping the undo. Flattening to a single commit is the remedy the guard prescribes, and the tree is byte-identical across it (`insurance/5165-premerge-head^{tree}` and the flattened head are both `1dee87c7af0d8c86156d0de9c2f98771111d03a5`).

### Evidence

- `cargo test -p stella-store` — 488 pass, including `schema_version_matches_the_ladder` at 40 and the migration ladder.
- `cargo test -p stella-core plan_graph` — 18 pass, both proptests included.
- `cargo check -p stella-cli --all-targets` — clean.
- `make guards-fast` — passes end to end, `file-size`, `rebase-replay`, `doc-links`, `invariants` and `line-citations` included.
- `make rebase-replay` — `OK — 23 path(s) touched, every one of them still in the branch's diff.`

The two checks still red here are `main`'s, not this branch's: `prose` fails on ten constructions #5142 and #5146 left on `main`, and `main is not known-broken` reports #5166. Both clear when #5170 merges.

### Filed rather than fixed here

#5175 — two ADRs share the number 0015 on `main`, `docs/adr/README.md` omits two records that exist on disk, and nothing checks either. This PR's ADR had to be renumbered twice by hand against a moving base, which is what surfaced it.

macanderson:
Correction to the evidence line above: the `SCHEMA_VERSION` assertion is not in a test called `schema_version_matches_the_ladder` — no such test exists. `assert_eq!(SCHEMA_VERSION, 40)` is at `crates/stella-store/src/tests.rs:1069`, inside `skill_usage_records_per_execution_version_rows`. The run and its count are as stated (488 pass in `cargo test -p stella-store`); only the test name was wrong.

PR #5167, opened 2026-08-26
fix(ci): unbreak main — prose gate and two unpinnable DoD workflow callers

### LINKED ISSUE #5166: main is red: prose fails on the merged tree

Failing: **prose**

### `prose`

```

```

canary exists (#3332).

### How to fix

```sh
git checkout main && git pull
```

...then, per failing check:

#### `prose`

```sh
#   make prose-report
```

than only CI.

---

## Definition of done

macanderson:
`main` is red at `5857241b6` on **two independent gate steps**. The canary
(#5166) named only the first; the second was still running when it filed, and
is the more urgent of the two because it sits inside a *required* check.

## 1. `prose` — the cause the canary named

Ten flagged constructions across six files, arriving from #5127, #5146 and
#5142:

| File | Pattern | Fix |
|---|---|---|
| `crates/stella-cli/src/rules.rs` (×1) | `rust-jargon` `variant` | `the first unlisted one` |
| `crates/stella-cli/src/rules.rs` (×2) | `rust-jargon` `invariant` | `guarantee` |
| `crates/stella-core/src/rules.rs` | `rust-jargon` `variant` | `the first unlisted one` |
| `crates/stella-core/src/rules/tests.rs` | `rust-jargon` `variant` | `spelling` |
| `crates/stella-core/src/records/bridge.rs` | `filler-adverb` | deleted |
| `docs/adr/0016-…md` (×3) | `filler-adverb` | deleted |
| `docs/scr/README.md` | `filler-adverb` | `from that decision` |

No baseline entry was added and none was raised. In every `deliberately` case
the reason the adverb stood in for already follows it in the same sentence, so
deleting the word costs the reader nothing — which is the guard's own test.

## 2. `action-pins` — a required check, not named by the canary

`.github/workflows/dod-check.yml` and `dod-close-guard.yml` (#5142) call:

```yaml
uses: macanderson/oxagen/.github/workflows/dod-check.yml@main
```

A cross-repo `uses:` must be pinned to a 40-hex SHA — the guard's only
exemption is a local `./` reference. **There is no commit to pin to.** Neither
file exists in `macanderson/oxagen` (public repo, verified readable, `contents`
API returns 404 for both, and `.github/workflows/` there lists ten files,
none of them `dod-*`). The implementation is macanderson/oxagen#1323, which is
**still open**; oxagen#1321 is likewise still open.

So the callers were adopted ahead of the thing they call. The consequence is
already measured, not predicted — every run since they landed has failed:

```
dod-check         failure  pull_request  ×4
dod-close-guard   failure  issues        ×1
```

`check-action-pins.sh` runs in `ci.yml`'s `fmt + clippy + test` job, which is
one of main's three required checks — so this reddens **every open PR**, which
is exactly the blast radius `main-red-hold` exists to bound.

Both callers are removed here. Note that `@main` would fail `action-pins`
*even after* oxagen#1323 merges, so the re-add must be SHA-pinned; that is
tracked in the follow-up issue below.

## The corpus divergence this creates

`docs/scr/README.md` sits inside the byte-synced SCR corpus (compared by git
blob SHA across five repos), so this prose fix diverges stella's copy from
oxagen's. oxagen's `scr-corpus-check` is not live yet — it ships in the same
open #1323 — so nothing breaks today, and the matching oxagen edit is tracked
in the follow-up issue. The README's own instruction is to resolve drift by
"deciding what the corpus should say and re-syncing all five", which is what
the issue asks for.

I did **not** edit `docs/scr/SCR-003-dod-verified-close.md`, whose
`enforcement` field now describes a merge gate that is not running. Editing it
would deepen the corpus divergence for no gain; the gap is in the issue.

## Evidence

`make guards-fast` on this tree, all 49 gate steps green:

```
check-action-pins: OK — all 86 'uses:' references are pinned to a commit SHA.
check-prose:  OK -- 4171 grandfathered construction(s) in 1173 file(s), none added
check-gate-parity: OK — 49 (forty-nine) gate steps
check-lockfile-sync: OK — Cargo.lock resolves against the manifests.
cargo fmt --check     (clean)
```

Both failures reproduce on `main` before the change and pass after it, which is
the witness for a gate repair. The Rust edits are comment-only — no behavior
change, so no behavioral witness test applies; deleted tests: none.

Closes #5166

## Summary by Sourcery

Unblock the main branch by fixing prose-gate violations and removing unavailable DoD workflow integrations.

Bug Fixes:
- Restore required CI checks by removing workflow callers that reference unavailable cross-repository workflows.
- Resolve prose-gate violations across Rust comments and project documentation.

CI:
- Remove the unpinnable DoD check and close-guard workflow callers so action pin validation passes.


macanderson:
Closing in favour of #5170, which fixes the same eight files.

The deciding difference is the two DoD workflows. This PR **deletes** them outright — `.github/workflows/dod-check.yml` 23 lines to 0 and `dod-close-guard.yml` 20 to 0 — whereas #5170 repairs them in place (3+/3- and 1+/1-).

Deleting a guard to make its check pass removes the guarantee rather than restoring it, and these two landed deliberately with #5142 (the DoD merge check). If they are genuinely unpinnable as written, that is worth its own issue and its own argument, not a silent removal inside a main-unblock PR.

The prose fixes in both are equivalent. Nothing here is lost.

Issue #5168, opened 2026-08-26
Restore the two DoD workflow callers, SHA-pinned, once oxagen#1323 merges


macanderson:
## Problem

#5142 added two caller stubs for a reusable workflow that does not exist:

- `.github/workflows/dod-check.yml` → `macanderson/oxagen/.github/workflows/dod-check.yml@main`
- `.github/workflows/dod-close-guard.yml` → `macanderson/oxagen/.github/workflows/dod-close-guard.yml@main`

Neither target file exists in `macanderson/oxagen`. The repo is public and
readable; the `contents` API returns 404 for both paths, and
`.github/workflows/` there holds ten files, none named `dod-*`:

```
ci-image.yml  db-migrate.yml  linear-release.yml  nightly.yml  pipeline.yml
release.yml   stella-sidecar-nightly.yml  stella-sidecar.yml
triage-guard.yml  vision-gate.yml
```

The implementation lives in **macanderson/oxagen#1323** (`process: enforce SCR
corpus sync and DoD-verified close`), which is still open. oxagen#1321 (the
DoD merge check) and oxagen#1320 (the corpus sync check) are also still open.

Both callers failed 100% of their runs — four `pull_request` runs of
`dod-check`, one `issues` run of `dod-close-guard` — and, because a cross-repo
`uses:` that is not SHA-pinned fails `check-action-pins.sh`, they also turned
`main`'s required `fmt + clippy + test` job red, which reddens every open PR.

They were removed in #5167 to unbreak `main`. This issue tracks putting them
back correctly.

## Definition of done

1. **oxagen#1323 has merged**, so `dod-check.yml` and `dod-close-guard.yml`
   actually exist on `macanderson/oxagen@main`. Verify:
   ```sh
   gh api repos/macanderson/oxagen/contents/.github/workflows/dod-check.yml --jq .name
   gh api repos/macanderson/oxagen/contents/.github/workflows/dod-close-guard.yml --jq .name
   ```
2. **Both callers are restored SHA-pinned, not `@main`.** This is the part
   that bit last time and would bite again: `@main` fails `action-pins`
   *regardless* of whether the target exists. The guard's only exemption is a
   local `./` reference (`scripts/check-action-pins.sh`, the `local_ref`
   pattern). Resolve the SHA with the API, not `git ls-remote` — for an
   annotated tag the latter prints the tag object and the workflow then fails
   at startup:
   ```sh
   gh api repos/macanderson/oxagen/commits/main --jq .sha
[...]

Issue #5169, opened 2026-08-27
action-pins cannot tell an unpinned ref from an unreachable one, so a cross-repo caller can merge ahead of its callee


macanderson:
## Problem

`scripts/check-action-pins.sh` asks one question of every `uses:` reference — *is this a 40-hex commit SHA?* It cannot ask the question that actually matters for a cross-repo reusable workflow: *does that target exist and is it reachable?*

Those come apart, and it cost a red `main` on 2026-08-26.

#5142 landed two caller stubs:

```
.github/workflows/dod-check.yml:23        uses: macanderson/oxagen/.github/workflows/dod-check.yml@main
.github/workflows/dod-close-guard.yml:20  uses: macanderson/oxagen/.github/workflows/dod-close-guard.yml@main
```

The workflows they name **did not exist in oxagen**. The implementation half was still open as oxagen #1323 while this repo merged the caller half, which produced:

1. `action-pins` failed, because `@main` is not a SHA — a *correct* failure for the wrong reason. It was unfixable as reported: there was no commit to pin to.
2. `dod-close-guard` dispatched against a missing target and **failed on every run** (`.github/workflows/dod-close-guard.yml`, failure, `2026-08-26T23:42:07Z`).

A reviewer reading the guard's output sees "pin this to a SHA" and has no way to learn that the reason it is unpinned is that the callee has not shipped.

Repaired in #REPLACE_PR by pinning to `a26bac4001eb5501f9d93f1deec240faecff7f1d` once oxagen #1323 merged. That repairs the instance; the shape is still open.

## Why this is not just "merge in the right order"

Ordering is the fix for *this* instance and cannot be the general one. The two repositories have independent gates, and neither side's PR is wrong on its own tree:

- oxagen #1323 was green and correct without stella's caller.
- stella #5142 was green and correct without oxagen's callee — because nothing here can see oxagen.

This is the shared-cell composition hazard AGENTS.md describes for `Cargo.lock` and `scripts/file-size-baseline.txt`, with the cell living in **another repository**. `main-canary.yml` catches the in-repo version after the merge; nothing catches this one.

## Files involved

- `scripts/check-action-pins.sh` — the guard
- `.github/workflows/dod-check.yml`, `.github/workflows/dod-close-guard.yml` — today's only cross-repo callers
- `.github/workflows/main-canary.yml` — the existing post-merge composition check, the natural host for a reachability probe

## Reproduce

```sh
# On 5857241b6, before the repair:
[...]

PR #1321, opened 2026-08-03
bench: make the authored-witness A/B runnable and measurable (#1284)

macanderson:
## What & why

#1284 asks for two runs — the same task set with the independent checker off, then on — and a plain statement of whether the checker is worth its cost. It says nothing remains to build. That is true of the *run*: `STELLA_WITNESS_AUTHOR_MODEL` has existed since #1007 and task folders got a git baseline in #1225.

It was not true of the *measurement*. Two gaps, both of which would have been discovered after spending:

1. **The secondary outcome was not recoverable from committed evidence.** "How many wrong 'this passed' calls disappear" needs Stella's own verdict beside the external grader's reward. That verdict existed only as a `judge_verdict` event inside the multi-gigabyte Harbor job tree, which is never committed — so the false-pass count was computable only while the run was still on the disk that produced it, and evaporated with it.
2. **Nothing turned two runs into an answer.** No pairing, no decision rule, and nothing that could tell a treatment arm which *declared* the witness tier from one that *exercised* it — which is #1147 exactly, and is detectable only from the run's own proof stream.

This PR closes both, plus the runner and the preregistration, so the run is one command per arm and one command to score.

**The run itself has not been executed.** It needs a spend-capable credential and a host that can run 89 Docker task images; this environment has neither, and it costs real money on the maintainer's account. Everything up to the credential is in the tree. `Refs #1284` rather than `Closes` for that reason.

### Changes

| area | change |
|---|---|
| `stella_harbor/posture.py` | fold `judge_verdict` into the stream summary beside the witness observations #1007 already recorded: `self_verdict_passed`, `self_verdict_deterministic`, `self_verdict_count`, `verification_unavailable_count`. Tri-state — a trial killed before the ladder closed made no claim, which is not the same datum as claiming failure. `deterministic` separates a flip-oracle verdict from a model opinion; only the second is the judge's reliability |
| `stella_harbor/__init__.py` | surface `stella_self_verdict_state` as trial metadata, beside `stella_witness_authored_state` |
| `score_dev_baseline.py` | `extract` carries the verdict, the witness author and the witness observations into `trials.jsonl`, so a run's honesty is recomputable from the committed run directory like its score |
| `compare_arms.py` (new) | pair two arms by task; report gained/lost with an exact McNemar test, a paired bootstrap interval, false passes fixed and introduced **by name**, and spend per additional task passed; apply the preregistered decision rule |
[...]

PR #5170, opened 2026-08-27
Repair the two checks red on main: prose, and the unpinnable DoD callers

macanderson:
`main` is red, and on **two** checks rather than the one the canary named. #5166 reports `prose`; running the gate locally on `5857241b6` found `action-pins` failing underneath it.

## 1. `prose` — ten constructions, six files

A different set from #5140's, landed by the `guard-allow-command` work and the SCR corpus sync:

```
crates/stella-cli/src/rules.rs            [rust-jargon]    3   (2 allowed)
crates/stella-core/src/records/bridge.rs  [filler-adverb]  1
crates/stella-core/src/rules.rs           [rust-jargon]    1
crates/stella-core/src/rules/tests.rs     [rust-jargon]    1
docs/adr/0016-...-a-field-not-a-pattern.md [filler-adverb] 3
docs/scr/README.md                        [filler-adverb]  1
```

Every `deliberately` sat in front of a reason its own sentence already gave, so deleting it costs the reader nothing. Each `variant`/`invariant` becomes the plain word the guard asks for — a missed spelling, and a guarantee.

## 2. `action-pins` — and why it could not have been fixed before now

Two `uses:` refs point at a sibling repo's default branch:

```
.github/workflows/dod-check.yml:23        macanderson/oxagen/...@main
.github/workflows/dod-close-guard.yml:20  macanderson/oxagen/...@main
```

**These were unpinnable when they landed, because the workflows they call did not exist.** #5142 merged the *caller* half of oxagen #1321 while the *callee* half was still open as oxagen #1323. The consequence was not theoretical — stella's `dod-close-guard` dispatched against a missing target and failed:

```
.github/workflows/dod-close-guard.yml   failure   2026-08-26T23:42:07Z
```

oxagen #1323 has since merged, so both reusable workflows now exist on oxagen `main`. Both refs are pinned to `a26bac4001eb5501f9d93f1deec240faecff7f1d` — the commit checked, via the contents API, to carry both files — with the branch kept in a comment as the guard's own message asks.

## Evidence

```
check-prose:       OK -- 4171 grandfathered, none added
check-action-pins: OK — all 88 'uses:' references are pinned to a commit SHA
make guards-fast   exit 0
```

The grandfathered prose count moves **down** (4173 → 4171). No baseline entry was added or raised; no `deny.toml` or file-size ceiling was touched.

Comments, docs and CI wiring only — no behaviour changes, so no witness test is owed. The guards' verdicts are the evidence.

## For whoever hits this next

This is the **third** red `main` from `check-prose` in a few hours (#5124 density, #5140 phrases, #5166 phrases again), and #5140's repair was closed without landing — its content is not on `main`, so those five constructions are still outstanding separately from this PR's ten.

The `action-pins` half is a different shape and a sharper one: a cross-repo caller merged ahead of its callee. Nothing pre-merge could catch it, because the guard can only ask *is this ref a SHA* — it cannot ask *does the target exist*. Filed as #5169.

Refs #5166 — **not** `Closes`, and the `no-issue` waiver is applied. #5166 is
machine-filed by `main-canary.yml`, so it has no Definition-of-done section and
by construction never can; `dod-check` fails any PR that closes a DoD-less
issue. The canary closes its own issue when `main` recovers (AGENTS.md), so the
lifecycle is already owned. The gap this exposes is filed as #5172.

### DIFF
--- a/.github/workflows/dod-check.yml
+++ b/.github/workflows/dod-check.yml
@@ -5,8 +5,8 @@
 # copies of an enforcement rule into an org whose sibling check exists to
 # detect exactly that kind of drift.
 #
-# There is deliberately no logic in this file. A change to *what* the check
-# does never touches it; only adding or removing a check does.
+# There is no logic in this file. A change to *what* the check does never
+# touches it; only adding or removing a check does.
 name: dod-check
 
 on:
@@ -20,4 +20,4 @@ permissions:
 
 jobs:
   dod:
-    uses: macanderson/oxagen/.github/workflows/dod-check.yml@main
+    uses: macanderson/oxagen/.github/workflows/dod-check.yml@a26bac4001eb5501f9d93f1deec240faecff7f1d # main
--- a/.github/workflows/dod-close-guard.yml
+++ b/.github/workflows/dod-close-guard.yml
@@ -17,4 +17,4 @@ permissions:
 
 jobs:
   guard:
-    uses: macanderson/oxagen/.github/workflows/dod-close-guard.yml@main
+    uses: macanderson/oxagen/.github/workflows/dod-close-guard.yml@a26bac4001eb5501f9d93f1deec240faecff7f1d # main
--- a/crates/stella-cli/src/rules.rs
+++ b/crates/stella-cli/src/rules.rs
@@ -33,7 +33,7 @@
 //!   positive match, which cannot express the shape most command rules
 //!   actually have: *this family is forbidden **except** in its scoped
 //!   form*. Approximating that with a list of deny globs only enumerates
-//!   the spellings someone thought of, and the first unlisted variant walks
+//!   the spellings someone thought of, and the first one they missed walks
 //!   through while the rule still looks enforced. Write the broad deny and
 //!   name the narrow exception — a missing exception shows up at once as a
 //!   false block, where a missing deny pattern is silent. Command-scoped by
@@ -278,8 +278,8 @@ pub(crate) fn load_workspace_rules(
 /// tier merge resolved them, plus the guards TOML records earned.
 ///
 /// The split is deliberate and the reason is blast radius. The markdown path carries
-/// an invariant with teeth — a user-global hard guard is *monotonic*, and a project
-/// rule with the same id can never remove it — and that invariant is enforced by
+/// a guarantee with teeth — a user-global hard guard is *monotonic*, and a project
+/// rule with the same id can never remove it — and that guarantee is enforced by
 /// [`merge_rule_trust_tiers`] over `Vec<Rule>`. Re-expressing it inside the record
 /// registry's lineage merge would mean rewriting security-relevant merge logic to
 /// add a feature, so the markdown result is passed through untouched and only the
--- a/crates/stella-core/src/records/bridge.rs
+++ b/crates/stella-core/src/records/bridge.rs
@@ -149,7 +149,7 @@ fn self_attested(record: &Record, trust: Trust) -> bool {
 ///
 /// [ev]: super::super::rules::evaluate_guards
 fn concrete(guard: &RuleGuard) -> bool {
-    // `allow_command_glob` is deliberately absent: an exception alone denies
+    // `allow_command_glob` is absent: an exception alone denies
     // nothing, so a record carrying only one would advertise Tier 2 while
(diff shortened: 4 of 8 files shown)


macanderson:
Superseded by #5167, which merged as `517a418a6` and repaired both red checks. #5166 is closed and `main` is green on prose.

This PR and #5171 were the same repair as #5167, opened within minutes of it — four sessions posted `main-red-claim` comments on #5166 and three repair PRs came out of it. Nothing here is missing from `main`: the ten prose hits are cleared, and `bridge.rs` and `docs/scr/README.md` carry no `deliberately`.

Closing rather than leaving it open, because it is now `DIRTY` in a way that is not a normal conflict: #5167 **deleted** the two DoD workflow callers (there was no oxagen commit to pin to at the time), while this PR **pins** them. Rebasing that would resurrect two files main deliberately removed — the shape where a stale branch merges as a silent revert.

Restoring them SHA-pinned is #5168, and the blocker it was waiting on has cleared: macanderson/oxagen#1323 merged, so both workflows now exist in oxagen at `319b0a97d8685abdb2721a3e95c2f766902eb4db`. The pinned `uses:` lines from this PR are the right content for that work — I have left them on #5168 so they are not lost with this branch.

macanderson:
Closing as superseded — both halves of this PR are now resolved elsewhere, and
what remains here is a conflict rather than a change.

- **The prose half is already on `main`.** #5167 merged at 00:22 with the same
  ten fixes. `make prose` is clean on `main` without this branch.
- **The workflow half is now a conflict.** #5167 deleted `dod-check.yml` and
  `dod-close-guard.yml`; this branch modifies them, so the merge is
  `CONFLICTING/DIRTY` against the new base.

The pinning idea here was right, and it is the one this branch got to first —
independently of #5174, which reached the same conclusion and is also closed.
It carries on as #5180, which re-adds both callers SHA-pinned and closes #5168
(the follow-up #5167 filed for exactly this). #5180 branches from the current
`main`, so it is the two-file diff this became rather than a conflicted eight.

Nothing here is lost. The `a26bac40` pin this branch chose and the `319b0a97`
pin #5180 uses carry a byte-identical `dod-check.yml`, so the two are the same
change.

macanderson:
Closing this as superseded. Both halves have landed or are landing elsewhere, and nothing in it is being dropped — that check matters here, because the previous repair in this chain (#5145) was closed on the belief its content was already on `main` when it was not.

**The prose half is on `main`.** #5167 fixed the same ten constructions in the same six files, with different wording. Verified rather than assumed — I reset a worktree to `517a418a6` and ran the guard:

```
check-prose: OK -- 4171 grandfathered construction(s) in 1173 file(s), none added
```

**The pin half is #5180**, which does exactly what #5168 asks and is pinned to `319b0a97` — newer than this branch's `a26bac40`, since oxagen `main` moved after #1323 merged. I verified their pin resolves before standing this down: both `dod-check.yml` and `dod-close-guard.yml` exist at `319b0a97`, and it is a real commit on oxagen `main`. Their PR is strictly fresher than mine, so racing it would only risk a conflict.

**What this PR contributed that outlives it:**

- oxagen #1323 is merged. That was the actual blocker — stella #5142 landed the *caller* half of the DoD check while the *callee* half sat unopened, so `dod-close-guard` had been dispatching against a workflow that did not exist and failing every run (`2026-08-26T23:42:07Z`). Neither #5167's deletion nor #5180's restore was possible to do correctly until it landed. I answered its three Sourcery ❌ rows — all scope-split objections, none a defect in the diff — before it merged.
- #5169: `action-pins` cannot distinguish *unpinned* from *unreachable*, so its remedy text told an author to pin something with nowhere to point.
- #5173 (via #5172, now closed as its duplicate): canary-filed `main-red` issues carry no Definition-of-done section, so `Closes #<canary issue>` fails `dod-check` — which will hit every future red-`main` repair.

**One correction to my own PR body**, so it does not mislead anyone reading this thread later: it claimed #5140's five constructions were "still outstanding" because #5145 was closed unmerged. They are not outstanding. Checked against `origin/main` directly — `no-full-test-builds.sh`, `SCR-001` and `SCR-005` carry none of the five flagged constructions today. I have not established *which* merge removed them, only that they are gone; the closed PR was not the only route. The constructions this PR fixed were a different set.

PR #5171, opened 2026-08-27
fix(ci): unbreak main — pin the two DoD reusable workflows and clear seven prose hits

### LINKED ISSUE #5166: main is red: prose fails on the merged tree

macanderson:
`main` is red at `5857241b6` on two guards, from two merges that each passed their own checks. Canary issue: #5166.

## `action-pins`

#5142 added `.github/workflows/dod-check.yml` and `dod-close-guard.yml`, both calling a reusable workflow in `macanderson/oxagen` at `@main`. A moving ref is what the guard exists to refuse — the callee can change under us between one run and the next. Both now name the commit (`319b0a97…`), with the branch kept in a trailing comment, which is the form the guard's own error message prescribes.

## `prose`

Seven constructions across five files, from #5142 and #5146:

- Four `deliberately` where the reason follows in the same sentence, so the adverb carries nothing (`records/bridge.rs`, the ADR twice, `docs/scr/README.md`).
- Three `variant`/`invariant` where the plain word is the clearer one: a deny list that misses a spelling misses a *spelling*, and a monotonic user-global hard guard is a *rule*.

## What is deliberately not in this PR

The prose baseline is left as it stands, so `crates/stella-cli/src/rules.rs [rust-jargon]` still reads `2` against `0` found. Retightening is a PR of its own, safe when nothing is blocked on it. Folding it in here would make the repair every open PR is waiting on the one racing every other merge for a shared cell — the shape that produced #4654.

## Evidence

- `make guards-fast` passes end to end on this branch, `action-pins` and `prose` included.
- `make prose` — `OK -- 4171 grandfathered construction(s) in 1173 file(s), none added`.
- `cargo check -p stella-core -p stella-cli --all-targets` — clean; the Rust edits are doc comments and one test comment, no code.

No witness test: this is a CI/prose repair, and the guards themselves are the witness — each fails on `main` and passes here.

Closes #5166

## Summary by Sourcery

Restore the main branch guards by pinning reusable DoD workflows and cleaning the newly flagged prose.

Bug Fixes:
- Pin the two DoD reusable workflows to a fixed commit so CI guard behavior is stable and reproducible.
- Remove seven prose constructions flagged by the repository prose guard.

CI:
- Stabilize DoD CI checks by replacing moving workflow references with a pinned revision.

### DIFF
--- a/.github/workflows/dod-check.yml
+++ b/.github/workflows/dod-check.yml
@@ -20,4 +20,4 @@ permissions:
 
 jobs:
   dod:
-    uses: macanderson/oxagen/.github/workflows/dod-check.yml@main
+    uses: macanderson/oxagen/.github/workflows/dod-check.yml@319b0a97d8685abdb2721a3e95c2f766902eb4db # main
--- a/.github/workflows/dod-close-guard.yml
+++ b/.github/workflows/dod-close-guard.yml
@@ -17,4 +17,4 @@ permissions:
 
 jobs:
   guard:
-    uses: macanderson/oxagen/.github/workflows/dod-close-guard.yml@main
+    uses: macanderson/oxagen/.github/workflows/dod-close-guard.yml@319b0a97d8685abdb2721a3e95c2f766902eb4db # main
--- a/crates/stella-cli/src/rules.rs
+++ b/crates/stella-cli/src/rules.rs
@@ -33,7 +33,7 @@
 //!   positive match, which cannot express the shape most command rules
 //!   actually have: *this family is forbidden **except** in its scoped
 //!   form*. Approximating that with a list of deny globs only enumerates
-//!   the spellings someone thought of, and the first unlisted variant walks
+//!   the spellings someone thought of, and the first unlisted one walks
 //!   through while the rule still looks enforced. Write the broad deny and
 //!   name the narrow exception — a missing exception shows up at once as a
 //!   false block, where a missing deny pattern is silent. Command-scoped by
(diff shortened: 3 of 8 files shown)


macanderson:
Closing as a duplicate of #5170, which is ahead on every axis that matters here.

Both PRs touch the same eight files and take the same approach — SHA-pin the two DoD callers rather than delete them, now that oxagen#1323 has merged and there is a commit to pin to. The differences:

- **#5170 is green where this one is red.** `dod` passes there (its body says `Refs #5166` rather than `Closes`, with the `no-issue` waiver applied, so it does not claim to close a canary issue that has no DoD section); here `dod` still fails.
- **This branch reintroduces a banned construction.** `dod-check.yml`'s header goes back to "There is deliberately no logic in this file" — `filler-adverb` is one of the ten constructions #5166 is red on, so this re-adds an instance of the very defect the PR exists to clear.
- The pins differ only in which oxagen commit they name (`319b0a97` here, `a26bac40` in #5170). Both contain `dod-check.yml`; the extra commit between them is #1326's nightly fix, unrelated to this workflow.

Three PRs were opened for this one repair within 22 minutes (#5167, #5170, this one) — `scripts/main-red-claim.sh` exists for exactly that, and is worth running before the next red-main repair.

Reopen if I have read the comparison wrong.

macanderson:
Closing in favour of #5170, which is the same repair and was opened first (00:11:38Z against this one's 00:13). Both pin the two DoD callers to an `macanderson/oxagen` commit and delete the same ten prose constructions from source rather than baselining them; the only difference is which oxagen commit each of us resolved `main` to a few minutes apart.

`./scripts/main-red-claim.sh check` reported #5166 unclaimed when I started, so the claim did not prevent the duplicate here — the two sessions were already in flight. Nothing in this PR is lost by closing it.

PR #5174, opened 2026-08-27
fix(ci): unbreak main — SHA-pin the DoD callers and clear the prose gate

### LINKED ISSUE #5166: main is red: prose fails on the merged tree

macanderson:
fix(ci): unbreak main — SHA-pin the DoD callers and clear the prose gate

`main` is red at 5857241b6 on two independent gate steps. #5167 diagnosed both
correctly and was closed unmerged, so both are still red; this repairs them,
taking the other option on the second.

## 1. `action-pins` — a required check

`.github/workflows/dod-check.yml` and `dod-close-guard.yml` (#5142) called
`macanderson/oxagen/…@main`. A cross-repo `uses:` must be pinned to a 40-hex
SHA, so the guard failed inside `ci.yml`'s `fmt + clippy + test` job and
reddened every open PR.

When #5167 was written there was no commit to pin to — neither workflow
existed in oxagen, and its implementation (macanderson/oxagen#1323) was still
open. That is no longer true: #1323 merged, and both files are present at
oxagen `319b0a97`. So the callers are **pinned rather than deleted**, which
keeps SCR-003 enforcement instead of removing it to turn a gate green.

Verified before pinning — both files resolve at that SHA, and a `dod` job now
runs and passes in each of the three sibling repos that carry the same caller
(macanderson/cgp-website#18, macanderson/arenabench#11,
macanderson/context-graph-protocol#83, all merged). `make action-pins` reports
all 88 references pinned.

Re-pinning when the implementation changes is a one-line diff naming the
version it moved to — which is the property `@main` gave away: another
repository could change what runs here with no commit in this one.

## 2. `prose`

The ten flagged constructions from #5127, #5146 and #5142, fixed the same way
#5167 fixed them: three `rust-jargon` `variant`/`invariant` uses reworded, six
`deliberately` adverbs deleted where the reason already follows in the same
sentence, and `docs/scr/README.md`'s reworded to "from that decision". No
baseline entry added and none raised.

Verified: `make prose` OK, `make action-pins` OK, `make guards-fast` passes,
`cargo test -p stella-core --lib rules::` 52 passed, and
`cargo test -p stella-cli --bin stella rules` 36 passed.

Closes #5166

## Summary by Sourcery

Unblock the main CI gates by pinning shared DoD workflows and removing flagged prose violations.

Bug Fixes:
- Restore required CI checks by pinning the shared DoD workflows to a stable oxagen commit.
- Clear prose-gate violations across code comments, documentation, and rule explanations.

Enhancements:
- Retain cross-repository DoD enforcement while making workflow changes reviewable and reproducible.

CI:
- Pin both cross-repository reusable workflow callers to the oxagen implementation commit so action-pins validation succeeds.

Documentation:
- Clarify that SCR copies should be re-synchronized based on an explicit corpus decision rather than blindly overwritten.

### DIFF
--- a/.github/workflows/dod-check.yml
+++ b/.github/workflows/dod-check.yml
@@ -5,8 +5,13 @@
 # copies of an enforcement rule into an org whose sibling check exists to
 # detect exactly that kind of drift.
 #
-# There is deliberately no logic in this file. A change to *what* the check
-# does never touches it; only adding or removing a check does.
+# There is no logic in this file. A change to *what* the check does never
+# touches it; only adding or removing a check does.
+#
+# The ref is a commit SHA, not `@main`: a cross-repo `uses:` resolved through
+# a moving ref both fails `action-pins` and lets another repository change
+# what runs here without a commit in this one. Re-pin when the implementation
+# changes, which is a reviewable one-line diff naming the version it moved to.
 name: dod-check
 
 on:
@@ -20,4 +25,4 @@ permissions:
 
 jobs:
   dod:
-    uses: macanderson/oxagen/.github/workflows/dod-check.yml@main
+    uses: macanderson/oxagen/.github/workflows/dod-check.yml@319b0a97d8685abdb2721a3e95c2f766902eb4db # oxagen main, #1323
--- a/.github/workflows/dod-close-guard.yml
+++ b/.github/workflows/dod-close-guard.yml
@@ -17,4 +17,4 @@ permissions:
 
 jobs:
   guard:
-    uses: macanderson/oxagen/.github/workflows/dod-close-guard.yml@main
+    uses: macanderson/oxagen/.github/workflows/dod-close-guard.yml@319b0a97d8685abdb2721a3e95c2f766902eb4db # oxagen main, #1323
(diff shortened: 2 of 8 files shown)


macanderson:
Closing as a duplicate of #5170, which came first and reaches the same two
conclusions independently.

I claimed #5166 via `scripts/main-red-claim.sh` and built this before finding
#5170 — the claim check reported the repair unclaimed, so the collision the
script exists to prevent happened anyway. Worth noting on #5166 rather than
here.

The two PRs agree on both halves, including the non-obvious one:

- **The DoD callers are pinned, not deleted.** #5167 deleted them because at
  the time there was no commit to pin to — macanderson/oxagen#1323 was still
  open. It has since merged, so pinning is available and keeps SCR-003
  enforcement instead of removing it to turn a gate green. #5170 pins
  `a26bac40`; this branch pins `319b0a97`. Both refs carry a byte-identical
  `dod-check.yml` (6019 bytes), so the choice does not matter.
- **The same ten prose constructions**, fixed the same way, with no baseline
  entry added or raised.

#5170 is the one to merge. Nothing here is lost by closing it.

Issue #5183, opened 2026-08-27
The MCP tab's latency column is the connect-time round trip and never refreshes


macanderson:
## Problem

Split out of #5047, which added SPEC §9.3's latency column.

`ServerHealth::latency` is set from the `initialize` round trip at connect, and
re-set on every reconnect. Nothing else writes it, so on a long session the
number on screen is however far away the server was when the session started —
`r refresh` re-reads the snapshot but re-measures nothing, and a link that
degrades from 40ms to 4s keeps rendering `40ms`.

The field's own doc says "how far away is this server right now", and it only
answers "how far away was it when we dialed". `stella mcp usage` records call
counts and no durations, so nothing else on the row can correct the
impression.

The blank-when-disconnected rule already stops the worst version of this (a
figure beside `not connected`), so what is left is a connected server whose
distance silently ages.

## What "done" might look like

Either measure and report a rolling figure from the `tools/call` round trips
the session already makes — which is what the operator actually cares about,
and needs no extra traffic — or have `r` re-measure explicitly and say which
of the two the number is. Whichever is chosen, the column and its doc should
agree about the question being answered.

Worth noting: a per-call figure would price a slow *tool* as a distant
*server*, which is why #5047 measured `initialize` alone. A durable answer
probably reports both, or names the one it reports.

## Where

- `crates/stella-mcp/src/client/health.rs` — `ServerHealth::latency`
- `crates/stella-mcp/src/client.rs` — `run_handshake`, the reconnect path
- `crates/stella-tui/src/views/mcp_tab.rs` — `latency`

## Definition of done

The done conditions this issue already states, as boxes the gate can read.
Nothing here is new scope, and nothing is ticked by the migration — the
gate reads one heading and this issue used another (macanderson/oxagen#1407).

- [ ] The MCP tab's latency column reports both the connect-time round trip and a live one, or names which of the two it reports.
- [ ] Whichever is chosen, the column's meaning is written where the reader of `ServerHealth::latency` finds it.

macanderson:
Folded into batch #6299 (2026-09-06 audit): fixed together by crate.

Issue #5182, opened 2026-08-27
A reinstall over a granted MCP alias keeps the grant, so it can repoint at another publisher


macanderson:
## Problem

Split out of #5047, which added the first-enable capability grant (SPEC §9.3,
`docs/adr/0017-mcp-capability-grants.md`).

`stella mcp install` deliberately leaves an existing alias's recorded grant
alone: the operator granted *that alias*, and re-fetching the same publisher's
transport is not a new question. That reasoning holds for the case it was
written for — re-running `stella mcp install com.stripe/mcp` after a card
update — and does not hold for the case it also permits:

```
stella mcp install com.stripe/mcp        # lands ungranted
stella mcp grant mcp                     # reviewed and granted
stella mcp install io.github.someone/mcp --alias mcp
```

The third line overwrites the transport (`cmd`, `args`, `env`, or the URL) with
a different publisher's while `granted = true` survives, so a command line
nobody reviewed inherits a grant given to another one. Nothing in the flow says
the alias changed hands.

The MCP tab's `↵` on a search result reaches the same `resolve_install` path,
so it is not shell-only.

## What "done" might look like

`install` compares the incoming transport with the recorded one and resets
`granted` to `Some(false)` when they differ — the alias is re-asked exactly
when what it points at has moved. A same-transport reinstall keeps the grant,
which is the case the current behavior was written for.

Worth deciding at the same time: whether the same rule should apply when a
*granted* server's declared tool list changes across a reconnect. #5047's ADR
lists that as the other half of the seam and neither is covered today.

## Where

- `crates/stella-cli/src/mcp_cmd.rs` — `install`, `resolve_install`
- `crates/stella-mcp/src/config.rs` — `McpServerEntry::granted`
- `docs/adr/0017-mcp-capability-grants.md` — "Consequences", last bullet

macanderson:
Closing as a duplicate of #5178, which is the same defect filed twenty-nine
minutes earlier: `McpConfig::upsert_with_card` records `granted = Some(false)`
only when it creates a new entry, so `stella mcp install <other> --alias
<granted-alias>` repoints a granted alias at a different publisher's transport
with no second handshake.

Two things from this issue are merged into #5178 rather than lost with it:

1. **The remedy differs, and the difference matters.** #5178 proposes keying the
   reset on `registry_name`; this issue proposes keying it on the transport
   (`cmd`, `args`, `env`, or the URL). They disagree about the case where one
[...]

Issue #5176, opened 2026-08-27
MCP registry 'signed' attests the publisher's namespace, not the package bytes


macanderson:
## Problem

`stella_mcp::SignatureStatus::Signed` (landed with #5047) is rendered as `signed`
in the deck's MCP tab and in `stella mcp search`, and it is the gate
`mcp_cmd::resolve_install` refuses on. What it actually establishes is narrower
than the word suggests:

- the entry's namespace is one the registry attributed to a publisher (a
  DNS-verified domain, or a code-host account), **and**
- the official `_meta` lifecycle record says `active`.

Nothing verifies a signature over the package the entry points at. An `npm`
identifier or a remote URL under a verified namespace can still serve anything.
The `SignatureStatus` doc comment says so in full, and the website page
(`website/content/docs/commands/mcp.mdx`) repeats it — but the row itself says
one word, and a green `signed` badge is the most over-read thing on that
surface.

This is not a defect in #5047's gate (blocking unattributed entries is strictly
better than the nothing that preceded it). It is the next rung.

## Files

- `crates/stella-mcp/src/registry.rs` — `SignatureStatus`, `signature_status`
- `crates/stella-cli/src/mcp_cmd.rs` — `resolve_install`'s refusal, `run_search`
- `crates/stella-tui/src/envelope/mcp.rs` — `McpSignature`, the display mirror
- `crates/stella-tui/src/views/mcp_tab.rs` — the row's label

## What to do

1. Watch for the MCP registry ecosystem publishing artifact attestations
   (npm provenance / Sigstore bundles are the likely shape). Nothing to verify
   exists today, which is why this is filed rather than built.
2. When one lands, add a state that distinguishes "the registry vouches for the
   publisher" from "the artifact's provenance was verified", and give them
   different words on the row. Do not widen `Signed` to cover both.
3. Until then, consider whether the row should say `attributed` rather than
   `signed`. SPEC §9.3 spells it `signed`, so changing it is a spec edit, not a
   code one.

## Definition of done

The row's word matches what was actually checked, and a test pins that a
namespace-verified-but-unattested artifact does not render as the same state as
a verified one.

macanderson:
Folded into batch #6299 (2026-09-06 audit): fixed together by crate.

Issue #5175, opened 2026-08-27
Two ADRs share the number 0015, and nothing checks ADR numbering


macanderson:
## Problem

`docs/adr/` has no guard on its numbering, and `main` is already carrying the damage. Two records share 0015:

- `docs/adr/0015-adopt-standing-decisions-scr-corpus.md`
- `docs/adr/0015-the-task-tag-rides-the-event.md`

The same shape nearly repeated a third time: PR #5165 authored `0015-plan-graph-persistence.md` against a base where 0015 was free, and by the time it merged, main held both of the above *and* `0016-guard-exceptions-are-a-field-not-a-pattern.md`. It had to be renumbered twice, by hand, to land as 0017.

This is the shared-cell hazard AGENTS.md describes for `Cargo.lock` and `scripts/file-size-baseline.txt`, one directory over: two branches each pick "the next ADR number", both are correct against their own base, and they compose into a directory where a number no longer identifies a record. Nothing pre-merge can see it, because neither author's tree is wrong.

`docs/adr/README.md`'s index has drifted alongside it. The "Outside the Phase 0 series" table lists 0013, 0014 and 0015 (the task tag) and omits two records that exist on disk:

- `0015-adopt-standing-decisions-scr-corpus.md`
- `0016-guard-exceptions-are-a-field-not-a-pattern.md`

`make doc-links` reports the second as uncited but says nothing about the first, and neither is a link failure, so the gate is green over both.

## Why a guard rather than a rule

A rule saying "check the highest number first" is what every author already believes they are doing; it loses to a race it cannot see. The check is cheap and mechanical:

- every file matching `docs/adr/[0-9]{4}-*.md` has a unique 4-digit prefix;
- each one's `id:` frontmatter matches its filename stem, and its `# ADR NNNN:` heading matches the same number (PR #5165 shipped a rename where the file, the frontmatter, the manifest and both code citations were repointed and a prose paragraph below the table still said 0015 — a reviewer caught it, nothing else would have);
- each one appears exactly once in `docs/adr/README.md`'s index tables.

The last bullet is what turns the index from prose into a checked list, which is the same move `god-files` makes for `scripts/file-size-baseline.txt`.

## Files

- `docs/adr/` — the two colliding 0015 records; one has to be renumbered, and whichever moves needs its `id:`, heading, `docs/manifest.json` entry and any `doc:adr/...` citations moved with it (`rg 'adr/0015' --glob '!.git'`).
[...]

PR #5186, opened 2026-08-27
feat(mcp): pin graph, latency, registry tiers, first-enable handshake gate

### LINKED ISSUE #5047: The MCP tab does not pin the graph server, show latency, or gate first enable on a handshake
## Problem

SPEC 9.3 gaps in the MCP tab (rendering `06-mcp`):

1. The graph server is not pinned first, and the caption `graph is pinned · it is the product, not an integration` appears nowhere (`render_browse` iterates delivered order).
2. No latency column — `McpServerInfo` carries `health: Option<String>`, no ms.
3. Registry results carry no source tier (official/vendor/community), install count, or signature status; unsigned-blocked does not exist (`McpSearchItem` is `{name, description, kinds, installed}`).
4. The caption promises "the handshake shows capabilities before first enable", but no handshake gate exists — declared capabilities are visible only after the fact in the `ctrl+o` inspector. First enable should present declared capabilities and require the grant.

## Files

- `crates/stella-tui/src/views/mcp_tab.rs` (+ `mcp_tab/detail.rs`) — pin, caption, latency column, tier/signature rows, handshake surface
- `crates/stella-tui/src/envelope/mcp.rs` — carry latency, tier, signature
- `crates/stella-mcp` / the registry search path in `crates/stella-cli` — produce them; connect-time latency measurement
- `crates/stella-tui/src/deck_ui/mcp_keys.rs` — first-enable flow

## Definition of done

The four SPEC §9.3 gaps above, each verified rather than asserted.

- [x] The graph server renders **first**, with the caption `graph is pinned · it is the product, not an integration` — and the pin reorders the paint only, never the selection
- [x] Connected rows carry a latency in ms, measured at connect time; unknown renders blank, never `0ms`
- [x] Registry rows carry source tier, install count, and signature status — and **unsigned is blocked**, on the install path and not only in the paint
- [x] First enable of a new server renders its declared capabilities and requires an explicit grant
- [x] Witness test proves **no tool call reaches the server before the grant** — the absence on the wire, not merely the presence of a prompt
- [x] Full CI green
- [x] Code comments and docs updated (command reference, MCP concept page, ADR)

Spec: `design/tui-v2/SPEC.md` §9.3.

**Estimated agent build time:** ~1 day

macanderson:
## What & why

SPEC §9.3 named four things the MCP tab did not have. All four land here.

**The graph is pinned.** `render_browse` walked the delivered order, so stella's own graph server landed wherever `mcp.toml`'s alphabetical keys put it. It now heads the list with the caption `graph is pinned · it is the product, not an integration` under its rows. The *paint* is reordered, never the snapshot: `selected` indexes what the driver delivered, and sorting the rows would make `x` remove whichever server happened to be under the cursor.

**Rows carry latency.** `run_handshake` times the `initialize` request — that one round trip, not the whole handshake, since `tools/list` pages and would price a catalogue server as a distant one. It rides `ServerHealth::latency`, re-measured on every reconnect and cleared on teardown, so the column always describes the connection that is up. Unknown renders blank, never `0ms`.

**Registry rows carry provenance, and unsigned is blocked.** `McpSearchItem` was `{name, description, kinds, installed}`. It now carries the source tier (read from the namespace the registry actually verifies at publish time), the install count where a registry publishes one (`installs unknown` where it does not — a `0` would claim nobody runs the server), and a signature status. Anything the registry does not attribute, or has withdrawn, is **refused by `resolve_install`** as well as by the key handler: an install writes a `cmd` line this process later spawns, so the shell door and the deck door get the same answer.

**The first-enable handshake gate.** The tab's caption promised "the handshake shows capabilities before first enable" and nothing showed them — declared capabilities were visible only afterwards, in the `ctrl+o` inspector, and `e` sent `McpToggle` for every row. The path from a registry search to a stranger's code running in the session was two keystrokes with nothing in between naming what the server could do.

`stella_mcp::CapabilityGrants` is a shared allowlist `McpToolSet` consults on every advertise and every dispatch. An ungranted server's tools are absent from `schemas()` and every call to it is refused with `RefusedByPolicy` — **before anything reaches the transport**. `e` on an ungranted row opens the handshake pane (declared tools by name, scrolled, never a count) and `g` grants; `stella mcp grant <name>` does the same from a shell, with `--yes` for provisioning and `--revoke` to withdraw.

The decision persists as `granted` in `.stella/mcp.toml` in three states: an install writes `false`, a grant writes `true`, and a hand-written entry has no key at all and is usable — writing the transport yourself is already the review the gate asks for. `docs/adr/0018-mcp-capability-grants.md` argues why that third state is the threat model rather than a compromise: anyone who can write `granted = true` can equally omit the key, so reading absence as withheld would defend against nothing `Some(false)` does not already cover, and would disarm every existing workspace on upgrade — which is how operators learn to grant without reading. **Existing installs are unaffected.**

Closes #5047

## The witness

`no_tool_call_reaches_an_ungranted_server_before_the_grant` — `crates/stella-mcp/src/toolset/tests/gate.rs`.

It reads the scripted transport's own **request log** and asserts **zero** `tools/call` while ungranted, ahead of any assertion about the returned `ToolOutput`. That order is the point: a gate that refuses its caller *after* the request has gone down the pipe has already let a third party act, and every `ToolOutput::Error` assertion in this crate would pass over that bug. The handshake itself is expected on the wire before the grant — `initialize` and `tools/list` are how the capabilities under review become knowable at all.

Verified to flip, both halves independently:

- neutralizing the schema half → fails on the advertise assertion (`an ungranted server must not be advertised: [ToolSchema { name: "mcp__files__read", … }]`)
- neutralizing **only** the dispatch half → fails on the wire count, with the `tools/call` in the message (`a tool call reached the server before the operator granted it: [("initialize", …), ("tools/list", …), ("tools/call", …)]`)

Supporting witnesses: `e_on_an_ungranted_server_opens_the_handshake_instead_of_enabling_it` (deck keys — the `assert_ne!` against `McpToggle` is the old behavior), `the_first_enable_handshake_names_every_declared_capability` (the pane), `an_installed_server_lands_ungranted_and_the_grant_persists` (the on-disk half), `the_graph_server_is_pinned_first_and_says_why`, `pinning_moves_the_paint_and_never_the_selection`, `a_connected_row_shows_its_latency_and_an_unmeasured_one_shows_none`, `a_registry_row_carries_tier_installs_and_a_signature_that_can_block`, `install_stands_down_on_a_blocked_row_and_says_so`, plus the refreshed `tab_mcp` golden.

## The gate

- [x] `cargo fmt --check`
- [x] `cargo clippy -p stella-tui -p stella-mcp -p stella-cli --all-targets --all-features -- -D warnings` (scoped per SD-1; the workspace run is CI's)
- [x] `cargo test -p stella-tui -p stella-mcp`, `cargo test -p stella-cli --bin stella mcp` — green
- [x] `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps` for the three crates
- [x] Every fast guard: `file-size`, `god-files`, `prose`, `doc-links`, `command-docs`, `invariants`, `line-citations`, `deck-paths`, `tokens`, `contrast`, `hue-separation`, `light-clamp`, `transcript-surfaces`, `css-vars`, `lockfile-sync`, `deck-fit-all-test` and the rest — all OK
- [x] Docs updated: `website/content/docs/commands/mcp.mdx` (the new `grant` subcommand, the tier/installs/signature columns, what `signed` does and does not claim), `website/content/docs/agent-tools/mcp.mdx` (the withheld-until-granted section and the `granted` config key), `docs/adr/0018-mcp-capability-grants.md`
- [x] `Closes #5047` appears both above and in the commit body

## Nothing left behind

- [x] Filed: #5176, #5182, #5183

- **#5176** — `signed` attests the publisher's namespace, not the package bytes. The word is deliberately not a green badge over the artifact, and both the enum doc and the command page say so.
- **#5182** — a reinstall over a granted alias keeps the grant, so it can repoint a granted alias at another publisher's transport. Named in the ADR's Consequences as the seam to watch.
- **#5183** — the latency column is the connect-time round trip and never refreshes, so on a long session it ages.

## Ground-rule check

- [x] No I/O added to `stella-core`; no new dependencies at all
- [x] No new outbound network calls — the registry search that already existed is the only one, and `stella mcp grant` talks to the operator's own configured server
- [x] New cross-boundary types round-trip: `granted` is `Option<bool>` in `McpServerEntry` with `skip_serializing_if`, covered by `an_installed_server_lands_ungranted_and_the_grant_persists` (install → reload → grant → reload → revoke → reload)

**The read-model seam.** `stella-tui` takes only leaf dependencies (its `Cargo.toml` argues each one), so `stella-mcp`'s `SourceTier` and `SignatureStatus` cross into the TUI as display mirrors, mapped in `deck_mcp` with exhaustive `match`es rather than `From` impls — a state added in `stella-mcp` fails to compile here rather than rendering as something it is not.

**Where the gate sits.** Between the handshake and the first `tools/call`, not before connect. Gating the connect is strictly safer and makes the thing under review unfetchable — the operator would be granting an alias. The alternatives considered and rejected are in the ADR.

**Both hosts install the gate.** The deck and the one-shot `connect_mcp` both pass a grant set, so the property does not depend on which surface is driving. A set with no grants configured is *ungated* — that is the shape an embedder or a test with no human to ask gets, and it is why the field is an `Option` rather than a bare `HashSet`.

**File splits, not baselines.** Three files would have crossed the 1500-line ceiling: the handshake pane went to `views/mcp_tab/handshake.rs`, the gate tests to `toolset/tests/gate.rs`, and the MCP golden's fixture to `tests/deck_render_snapshots/mcp.rs`.

**Rebased onto `main`** (18 commits) — the ADR is numbered 0018 because 0015, 0016 and 0017 landed on `main` while this branch was in flight.

### DIFF
--- a/crates/stella-cli/src/agent.rs
+++ b/crates/stella-cli/src/agent.rs
@@ -675,7 +675,7 @@ pub(crate) async fn record_turn_episode<T, E>(
 /// instead (applied in [`connect_mcp_servers`]) — without that override the
 /// connect bound would double as the call bound and kill every long-running
 /// tool call at 10s.
-const MCP_CONNECT_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(10);
+pub(crate) const MCP_CONNECT_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(10);
 
 /// The parse of `.stella/mcp.toml` **and** of every installed package's own
 /// `mcp.toml`, split from the connect so a caller that owns a UI (the deck)
@@ -802,6 +802,7 @@ pub(crate) async fn connect_mcp_servers(
     native: std::sync::Arc<dyn ToolExecutor>,
     usage: Option<stella_core::mcp_usage::McpUsageLedger>,
     disabled: Option<stella_mcp::DisabledServers>,
+    grants: Option<stella_mcp::CapabilityGrants>,
     auth: Option<std::sync::Arc<stella_mcp::OAuthManager>>,
 ) -> McpToolSet {
     let mut set = McpToolSet::connect_with_auth(servers, MCP_CONNECT_TIMEOUT, auth)
@@ -819,6 +820,12 @@ pub(crate) async fn connect_mcp_servers(
     }
     if let Some(disabled) = disabled {
         set = set.with_disabled_servers(disabled);
+    }
+    // The first-enable handshake gate (SPEC §9.3). Installing a set makes it
+    // default-deny, so every caller that can reach a human passes one — the
+    // property must not depend on which surface is driving.
+    if let Some(grants) = grants {
+        set = set.with_capability_grants(grants);
     }
     set
 }
@@ -855,8 +862,12 @@ pub(crate) async fn connect_mcp(
         }
     };
     // A one-shot run has no interactive enable/disable, so no disabled set.
+    // It DOES get the capability gate: a grant is recorded on disk, so a run
+    // with no deck attached reads the same answer the deck wrote, and a server
+    // nobody has reviewed must not become callable by being run from a script.
     let auth = crate::mcp_cmd::oauth_manager(&cfg.workspace_root)?;
-    let set = connect_mcp_servers(&servers, native, usage, None, Some(auth)).await;
+    let grants = crate::mcp_cmd::initial_grants(&cfg.workspace_root, &servers);
+    let set = connect_mcp_servers(&servers, native, usage, None, Some(grants), Some(auth)).await;
     if print_diagnostics {
         crate::mcp_cmd::print_connect_diagnostics(&set);
     }
--- a/crates/stella-cli/src/cli/subcommands.rs
+++ b/crates/stella-cli/src/cli/subcommands.rs
@@ -269,6 +269,21 @@ pub enum McpCmd {
         /// The configured server's local name
         name: String,
     },
+    /// Show what a configured server declares at handshake, and grant (or
+    /// revoke) its capabilities. A server installed from the registry
+    /// advertises nothing to the model until it is granted.
+    Grant {
+        /// The configured server's local name
+        name: String,
+        /// Record the grant without the confirmation prompt (for scripts and
+        /// provisioning; the capabilities are still printed)
+        #[arg(long)]
+        yes: bool,
+        /// Withdraw a grant given earlier: the server stays configured and
+        /// connected, and its tools stop being offered to the model
+        #[arg(long, conflicts_with = "yes")]
+        revoke: bool,
+    },
     /// OAuth login to a configured http server (opens your browser; tokens
     /// land owner-only in .stella/private/mcp_oauth.json and auto-refresh)
     Login {
--- a/crates/stella-cli/src/command_deck.rs
+++ b/crates/stella-cli/src/command_deck.rs
@@ -794,10 +794,22 @@ pub async fn run_deck_session(
     //   • `mcp_disabled` — server names disabled this session; toggling it
     //     hides a server's tools from the model on the next call (live, no
     //     reconnect), because the engine re-reads schemas each call.
+    //   • `mcp_grants` — server names whose declared capabilities the operator
+    //     has granted (SPEC §9.3's first-enable handshake). Default-deny: a
+    //     connected server absent from it advertises nothing and refuses every
+    //     call. Seeded from `.stella/mcp.toml` when the connect task builds
+    //     its plan, and added to live when a grant is given, so a server
+    //     becomes usable in the session it was reviewed in.
     //   • the usage ledger (from the registry) records every MCP call for the
     //     `mcp_usage` telemetry table.
     let mcp_disabled: stella_mcp::DisabledServers =
         std::sync::Arc::new(std::sync::Mutex::new(std::collections::HashSet::new()));
+    let mcp_grants: stella_mcp::CapabilityGrants =
+        std::sync::Arc::new(std::sync::Mutex::new(std::collections::HashSet::new()));
+    let mcp_session = crate::deck_mcp::McpSession {
+        disabled: mcp_disabled.clone(),
+        grants: mcp_grants.clone(),
+    };
     // `Arc<McpToolSet>` (not a bare `McpToolSet`) so a turn can cheaply clone
     // the connected set into the Best-of-N candidate surface + orchestrator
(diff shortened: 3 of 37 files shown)


macanderson:
Heads up on the conflict here — it is one file and one cause. `docs/manifest.json` is the only conflicted path: this branch registers `adr/0017-mcp-capability-grants` and #5165 landed `adr/0017-plan-graph-persistence` on main first, so 0017 is taken.

The fix is to renumber this branch's ADR to 0018, which means moving five things together — the file, its `id:` frontmatter, its `# ADR 0018:` heading, the `docs/manifest.json` entry, and the index row plus any prose paragraph in `docs/adr/README.md`. `rg 'adr/0017-mcp|ADR 0017: MCP' --glob '!.git'` finds the set. Worth checking the README prose specifically: when #5165 did this same renumber, the file, frontmatter, manifest, index row and both code citations were all repointed and a paragraph below the table still said the old number.

I did not push the fix because this branch has a live worktree and was opened minutes ago — say the word if you would rather I take it.

This is the third ADR-number collision tonight (two records already share 0015 on main, then 0016, now 0017). #5175 tracks adding a guard so the next one fails in CI instead of at merge time.

macanderson:
Resolved this branch's conflict with `main` and pushed to it.

It was an **ADR-number collision**, not a content disagreement: this branch authored ADR **0017** as `mcp-capability-grants`, and #5165 landed a different ADR 0017 (`plan-graph-persistence`) while this branch was open. `docs/manifest.json` is keyed by id, so both claimed the same key and git had a real conflict to report.

Main's 0017 keeps the number; **this branch's ADR is now 0018**. The file, its frontmatter `id`, title and H1, the manifest entry, and the `docs/adr/0017-…` citation in `stella-mcp`'s `toolset.rs` all moved together, so `make doc-links` still resolves.

One thing beyond the conflict: the ADR was **not listed in `docs/adr/README.md`** — no table row, no paragraph. That predates this merge; every other ADR in the directory is listed, and an unlisted one is invisible to anyone reading the index rather than the tree. I added both, summarising the three-state `granted` field and `CapabilityGrants`' default-deny from the ADR's own Decision section. Reword it if I have the emphasis wrong.

Verified on the merged tree: `make guards-fast` passes — `doc-links` resolves 1119 citations across 94 documents, and `file-size`, `god-files`, `gate-parity`, `prose` and `line-citations` are green.

Issue #5055, opened 2026-08-26
The deck has no plugin panel host


macanderson:
## Problem

SPEC 12's host half does not exist: nothing in `stella-tui` leases a `Rect` to a plugin, blits returned styled lines into the buffer, draws host-owned chrome (`◳ panel · <plugin>` — the `◳` glyph appears nowhere in `crates/`), or throttles an over-budget panel with a visible tag. `panel_guard.rs` is unrelated (it is the panic boundary for the deck's own bands).

## Files

- New panel-host module in `crates/stella-tui` — lease, clip, blit, host-drawn border and title, frame budget + throttle tag
- Consumes the wire types from the stella-plugin contract issue (#5203/#5208, now merged)

## Definition of done

The three acceptance criteria are SPEC 12's own, unchanged; the format is the task template's so the close can be verified.

- [x] A plugin frame that draws outside its `Rect` is **clipped**, not drawn — witnessed by `a_frame_that_overruns_its_lease_is_clipped_not_drawn` and `a_patch_outside_the_lease_writes_nothing`.
- [x] An over-budget plugin shows the **throttle tag** — witnessed by `an_over_budget_panel_is_tagged`, which also asserts the tag is host ink so a plugin can neither remove nor imitate it.
- [x] **Plugin cells can never overwrite host chrome, asserted on buffer cells** — witnessed by `a_wide_glyph_cannot_reach_past_the_lease_into_the_border`, which reads the border cell's symbol before and after a double-width glyph is offered the last leased column. This is the criterion that caught a real breach: measuring the clip in `char`s rather than display cells let `あ` in the final column paint across the border, and ratatui's own diff then skipped that column so nothing repainted it.

Delivered by #5223, together with the process driver (`crates/stella-runtime/src/panel_host.rs`) and a reference plugin (`plugins/stella-hello/`) that drives the whole chain in `crates/stella-tui/tests/hello_panel_demo.rs`.

**Not in scope here, tracked separately:** placing a panel on a real screen — the SETTINGS pane, the transcript block and the `/name` popup — is #5227; the install-time handshake and grant is #5056.

Spec: `design/tui-v2/SPEC.md` §12.

**Estimated agent build time:** ~2 days
**Model:** use-model:ultra — Fable 5 (`claude-fable-5`)

macanderson:
[...]

Issue #5054, opened 2026-08-26
Plugins have no way to declare a panel: add the panel capability and wire types


macanderson:
## Problem

SPEC 12 (plugin panel protocol) has no wire half: `stella-plugin`'s capability set is closed (`DriverCall`; `release` is explicitly rejected as a capability) with no panel capability, no `Rect` lease shape, and no styled-line or cell-diff return type. Plugins must never emit raw ANSI; they are leased a `Rect` and return styled lines or a cell diff each tick.

## Files

- `crates/stella-plugin/src/manifest.rs` — a `panel` capability with declared denials (no network, no write outside sandbox)
- `crates/stella-plugin/src/wire.rs` — the lease request and the styled-line / cell-diff frame types, serde round-trip tests (invariant 4)
- `docs/wire/` regenerated (`make wire-schema`)

## Done means

A manifest can declare a panel; the frame types round-trip byte-for-byte; validation rejects a panel frame that addresses cells outside its lease. No host yet — that is the sibling issue; this one is the contract.

Spec: `design/tui-v2/SPEC.md` §12.

**Estimated agent build time:** ~1 day
**Model:** use-model:pro — Opus 5 (`claude-opus-5`)

## Definition of done

Verified line-by-line against PR #5187.

- [x] Implementation complete — `[panel]` capability with declared denials (`network`, `write-outside-sandbox`, closed set, completeness checked at load), the `Rect` lease request, and both frame shapes (styled lines, cell diff) in `crates/stella-plugin/src/panel.rs`. No host: nothing in `stella-tui`, which is #5055.
- [x] Plugins cannot emit raw ANSI, held by the type and not by documentation — `PanelText` wraps a private `String` whose only constructor refuses every control character, and `Deserialize` routes through it, so an escape sequence is a decode error rather than a value a host is trusted to inspect. A `compile_fail` doctest pins the privacy.
- [x] A frame cannot address a cell outside its lease — `PanelRect` carries an extent and no origin, so Stella's own coordinates are unwriteable, and `PanelFrame::fits` refuses a frame that starts or runs past the leased edge.
- [x] Validation rejects an out-of-lease frame, and the rejection is tested — witness `a_panel_frame_addressing_a_cell_outside_its_lease_is_refused` (`crates/stella-plugin/tests/wire_contract.rs`), covering all four edges plus the zero-length patch anchored past the right edge.
[...]

Issue #5185, opened 2026-08-27
A panel frame's cell count is chars, not display width — the host's clip is the only thing making that safe, and nothing asserts it


macanderson:
Noticed while landing #5054 (the panel wire contract). Filed rather than fixed there because the fix is split across the contract and the host (#5055), and #5054 is deliberately host-free.

## Problem

`PanelText::cells()` (`crates/stella-plugin/src/panel.rs`) counts `char`s. `PanelFrame::fits` — the validation a host uses to refuse a frame that addresses cells outside its lease — is built entirely on that count.

For a CJK ideograph, an emoji, or any East Asian Wide glyph, one `char` occupies **two** terminal columns. So a `lines` frame of four ideographs measures 4 cells here and draws 8 columns wide; a lease 4 columns wide accepts that frame and the terminal paints past the right edge. Combining marks fail the other way (two `char`s, one column), so a frame can also be refused for overflowing a lease it would have fit.

The type documents this and argues it is safe:

> A double-width glyph therefore draws wider than this contract counts, and the frame that carries it misaligns *its own interior* — never anything outside it, because the host clips every blit at the leased rectangle whatever a frame claims.

That argument is sound and it is the right call for a near-leaf crate that must argue every workspace dependency (`crates/stella-plugin/Cargo.toml`). But it makes the whole safety property conditional on a host behaviour that **does not exist yet and is asserted nowhere**. If #5055 blits a frame without clipping, a plugin panel repaints Stella's chrome using nothing but ordinary CJK text — no escape sequence, no out-of-range coordinate, nothing `fits` would refuse.

## Files

- `crates/stella-tui/` — wherever #5055 lands the panel renderer. The blit must clip at the leased rectangle per row, and that clip needs a test that a wide-glyph frame cannot paint outside its box.
- `crates/stella-plugin/src/panel.rs` — `PanelText::cells` and the paragraph quoted above, which should cite the host-side test once it exists instead of asserting the clip as a fact.

## Constraints

- `unicode-width` is already a workspace dependency (`stella-tui`, `stella-transcript` both take it), so the host side costs nothing new. Adding it to `stella-plugin` does not: that crate's manifest requires the argument to be made afresh for a third workspace edge, and "the host already owns the clip" is a real answer to why it should not be made. Prefer fixing this in the host.
[...]

PR #5208, opened 2026-08-27
feat(stella-plugin): a panel says where it draws, and a frame says which one it drew

### LINKED ISSUE #5203: A plugin panel cannot say where it draws: add the three surface kinds, a title, and a slash name
## Problem

`#5054` shipped the panel contract as **one panel, one place** (PR #5187, `crates/stella-plugin/src/panel.rs`): a plugin declares `[panel]`, accepts both `PanelDenial`s, and is leased a `PanelRect` it fills with styled lines or a cell diff. That contract has no way to say **where** the rectangle is.

SPEC 12 needs three placements, and they are three different asks a human should be able to weigh separately:

- a **settings pane** inside the SETTINGS tab (SPEC 9.5),
- a **transcript overlay**, the host-bordered `◳ panel · <plugin>` block (SPEC 12),
- a **command popup**, launched by `/<plugin-id>`, centered like the command palette (SPEC 10).

Today `PanelGrant` carries only `denies` and `process`. There is no surface kind, no title, and no slash name — so a plugin cannot declare a settings pane, and the install consent text cannot tell a reader whether they are agreeing to a pane, an overlay, a popup, or all three.

## Files

- `crates/stella-plugin/src/panel.rs` — `PanelSurface` (`Settings | Overlay | Command`, snake_case on the wire), and three new `PanelGrant` fields: `surfaces: Vec<PanelSurface>`, `title: Option<String>`, `command: Option<String>`.
- `crates/stella-plugin/src/error.rs` — the new typed refusals (repo invariant 5).
- `crates/stella-plugin/src/consent.rs` — `panel_say` discloses which surfaces are granted, and the title and slash name when they are declared.
- `crates/stella-plugin/src/wire_corpus.rs` — `panel_surface` joins the published closed vocabularies.
- `docs/wire/` regenerated (`make wire-schema-update`).

## Rules to enforce at load

- `surfaces` is required and non-empty whenever `[panel]` is present — a panel naming no surface draws nowhere (`PanelNoSurface`).
- No duplicate surface (`PanelDuplicateSurface`), the `DuplicateHook` idiom.
- `command` is only meaningful when `Command` is among the surfaces; set without it is a promise the UI will never keep (`PanelCommandWithoutSurface`).
- `command` is validated for **shape only** — a slug: lowercase ASCII, digits, `-`, no colon, non-empty, bounded. **`stella-plugin` must not learn `DECK_BUILTINS`**: that list lives in `crates/stella-cli/src/command_deck/skills.rs` (`DECK_BUILTINS`, `deck_reserved()`), and this crate is a near-leaf whose only workspace dependency is `stella-protocol`. The reserved-name check belongs to the host.
- `title` is single-line, bounded, and carries no control character — `PanelText`'s refusal reasoning, applied to chrome the host prints.

## The collision rule, and why it is not enforced here

The bare name is the headline (`/hello`), with the colon form (`/plugin:hello`) as an always-resolving alias. A bare name colliding with a built-in must be **refused visibly by the host** — a notice naming the plugin and the name it wanted — and never dropped in silence the way `CustomExtensions::slash_entries` drops a colliding `.stella/commands/*.toml` row (`crates/stella-cli/src/extensions.rs:701`). A signed manifest a human read must not contain a name that quietly does nothing. That host-side refusal is #5055's, not this issue's; this issue owns the shape check and the doc comment stating the rule.

## Done means

A manifest declares which of the three surfaces it draws; `PanelSurface` round-trips byte-for-byte; a `[panel]` block with no surface, a repeated surface, or a `command` without the `Command` surface each fail to load with a named error; the consent text says which surfaces are being granted. No host — that is #5055.

Spec: `design/tui-v2/SPEC.md` §10, §12, §9.5. Follows #5054 / PR #5187.

## Definition of done

Verified line-by-line against PR #5208.

- [x] `PanelSurface` with `all()`, `as_str()`, `Display`, `consent_sentence()`, and publication from `PanelSurface::all()` in the wire corpus — the `match` in `as_str` is what stops compiling when a case is added, so a new placement reaches the corpus, the consent rendering and the manifest check in the change that introduces it.
- [x] The `PanelGrant` fields and their load rules as typed `ManifestError` cases — `surfaces` and `command`; `PanelNoSurface`, `PanelDuplicateSurface`, `PanelCommandWithoutSurface`, `PanelCommandBlank`, `PanelCommandTooLong`, `PanelCommandCarriesNamespace`, `PanelCommandNotASlug`. No bare `String` errors (invariant 5).
- [x] **`title` was considered and refused.** It was implemented, reviewed, and removed before merge: a plugin-authored caption could read `GATES` or `stella*`, and chrome a person trusts must not name itself. Printing it *beside* the host's label rather than instead of it looks safe but makes the guarantee a property of the host's layout, which this crate cannot enforce. `PanelGrant`'s doc comment carries the argument under "There is no title field, and that absence is the anti-spoofing rule", plus a note recording the removal so the next author does not re-add it.
- [x] **`PluginManifest::name` is checked for drawability** — `ManifestError::NameNotDrawable`. The name is composed into the `◳ panel · <plugin>` label, the install prompt and a popup heading, so `PanelText`'s guarantee was worthless while the label around a panel's glyphs could carry `\x1b[2J`. One predicate, `first_control_character`, is now read by both `PanelText::new` and this check.
- [x] Witness tests, flip-verified artisanally — `a_panel_block_says_where_it_draws`, `a_panel_slash_name_is_held_to_its_shape`, `a_plugin_name_that_is_not_drawable_does_not_load`, `a_frame_for_one_surface_does_not_answer_another_surfaces_lease`, plus unit tests in `panel.rs`. `PanelSurface` round-trips byte-for-byte and the widened `PanelGrant` round-trips through TOML and JSON (invariant 4). **Four flips run and restored**, each failing its witness at a named line: surface rules (`:1321`), command-without-surface (`:1343`), surface routing (`:1358`), name drawability (`:1392`).
- [x] `consent_text` discloses the granted surfaces; the install-consent tests assert it — including that a placement the block never asked for is **not** disclosed as granted, and that a panel with no popup claims no name.
- [x] `docs/wire/` regenerated, `make wire-schema` green — `panel_surface` vocabulary plus the `surface` field on every published frame.
- [x] `cargo test -p stella-plugin --features schema` (exit 0, 11 binaries, 0 failures); `cargo clippy -p stella-plugin --all-targets` exit 0; `make prose`, `cargo fmt --check` and `make line-citations` green. No baseline touched.
- [x] No new workspace dependency — `stella-plugin` still takes `stella-protocol` and nothing else.
- [x] Residue: #5210 was folded into this PR rather than deferred, because the ambiguity it names was created by this change.

macanderson:
Closes #5203, #5210
Refs #5055, #5054

## What this adds

`#5054` (PR #5187) shipped the panel contract as **one panel in one unnamed place**. SPEC 12 reaches three, and they are three different things to agree to. This adds the declaration; the host is #5055.

- **`PanelSurface`** — `Settings | Overlay | Command`, snake_case on the wire, closed on `PanelDenial`'s reasoning. It selects **placement and chrome, never protocol**: all three are leased a `PanelRect` and answer with a `PanelFrame`.
- **`PanelGrant.surfaces`** — required and non-empty, no duplicates.
- **`PanelGrant.command`** — the bare slash name, defaulting to the plugin id.
- **`PanelLease.surface` / `PanelFrame.surface`** — which of the plugin's panels a lease is for, and which one a frame answers (#5210).
- **`PluginManifest::validate`** now refuses a name that is not drawable.
- **`consent_text`** discloses *which* surfaces are granted and the slash name in both its forms.

## Why the frame shape stayed as it is

`PanelPaint` keeps **both** shapes — styled lines and a cell diff — and all three surfaces share them. Lines suit a panel that recomputes its whole rectangle (a settings pane redrawn on a config change); a diff suits one that ticks at 30fps and changed four cells. Forcing either alone makes the other pay: lines-only means a spinner reserializes its rectangle every tick, diff-only means a full redraw must be expressed as a patch list the plugin computes against a buffer it does not hold. The surface kind does not change that trade-off, which is why placement went on the *grant*.

## There is no title, and that is the point

An earlier revision of this branch added `PanelGrant.title`. **It has been removed**, and `PanelGrant`'s doc comment carries the reason under *"There is no title field, and that absence is the anti-spoofing rule"* — a plugin-authored caption could read `GATES` or `stella*`, and chrome a person trusts must not name itself. The host labels with `PluginManifest::name` alone, which is the identity consented to at install.

The removal note is in the doc comment on purpose: printing a caption *beside* the host's label rather than instead of it looks safe, and it is not — it makes the guarantee a property of the host's layout, so any surface that later renders the caption alone hands a plugin the label. Nothing in this crate could enforce that ordering. That is an argument for the field not existing, not for documenting how to use it carefully. Three error variants went with it.

## The name is now checked for drawability

`PanelText` makes a panel's own glyphs unable to carry an escape sequence. That bought nothing while the label *around* them could: `PluginManifest::name` is composed into the `◳ panel · <plugin>` label, the install prompt, a popup heading and the rules panel, and was validated only for emptiness. A plugin named `"\x1b[2J"` clears the screen from inside Stella's own border, through a path no panel frame touches.

`PluginManifest::validate` now refuses it (`ManifestError::NameNotDrawable`, naming the character and its `char` position). One check at load rather than one per consumer, because every consumer is a place someone can forget. `PanelText::new` and this check share one predicate, `first_control_character` — the same question, asked once.

## A frame says which panel it drew (#5210)

Three surfaces landed in the change that made `PanelLease::panel` ambiguous — it is the *plugin's* name, shared by every lease that plugin is handed. Fixing that separately would have shipped a known-wrong contract on purpose, so `PanelLease` and `PanelFrame` both carry a `surface`.

`PanelLease::admits` now asks routing **before** geometry and returns `PanelRefusal` (`Surface`, `Tick`, or `Overflow`). A host that asked only the geometric question would blit a settings pane's frame into a command popup whenever the two were the same size — every cell inside the lease, and the wrong panel's content. `PanelFrame::fits` is unchanged and remains the geometry alone, for a caller that has already routed.

## The collision rule, and why it is not enforced here

Bare name is the headline (`/hello`); `/plugin:hello` is an always-resolving alias, derived from the plugin id and refused if an author writes it into `command`.

`validate()` checks **shape only** — a slug: lowercase ASCII opening character, then lowercase, digits and `-`, non-empty, bounded. It does not check reservation and cannot: `DECK_BUILTINS` lives in `crates/stella-cli/src/command_deck/skills.rs` and `stella-plugin` is a near-leaf taking `stella-protocol` and nothing else. A test asserts `command = "model"` **loads here**, so the boundary is pinned by a test rather than a comment.

The host must refuse it **visibly**, naming the plugin and the name it wanted — never the silent drop `CustomExtensions::slash_entries` does for a colliding `.stella/commands/*.toml` row. That is #5055's.

## Witnesses

- `a_panel_block_says_where_it_draws` — no surface, absent list, duplicate, unknown, and `command` without the `Command` surface.
- `a_panel_slash_name_is_held_to_its_shape` — namespaced, blank, over-length and six non-slug names pinned to the exact refusal and character position; the reserved name that must still load; and that a `title` key is refused.
- `a_frame_for_one_surface_does_not_answer_another_surfaces_lease` — the #5210 witness. Asserts the wrong-surface frame **`fits` the lease** before showing `admits` refuses it, so the test states plainly that geometry alone would have drawn it.
- `a_plugin_name_that_is_not_drawable_does_not_load` — four hazards, `char`-indexed position, and five ordinary names (including `ゲート` and `stella*`) that must still load, so the rule is drawability and not an allowlist.
- `the_consent_text_names_the_slash_command_a_panel_claims` — including that an unasked-for placement is **not** disclosed as granted.
- `every_panel_vocabulary_is_pinned_on_both_sides` — `PanelSurface` round-trips byte-for-byte, `all()` order pinned, unknown placement refused (invariant 4).

### Flips verified artisanally

Four reverts, each run and restored:

1. `PanelNoSurface` + `PanelDuplicateSurface` removed → `a_panel_block_says_where_it_draws` FAILED at `wire_contract.rs:1321`.
2. `PanelCommandWithoutSurface` removed → same witness FAILED at `wire_contract.rs:1343`.
3. Surface routing removed from `admits` → `a_frame_for_one_surface_does_not_answer_another_surfaces_lease` FAILED at `wire_contract.rs:1358`.
4. Name drawability check removed → `a_plugin_name_that_is_not_drawable_does_not_load` FAILED at `wire_contract.rs:1392`.

## Deleted tests

None. `a_panel_may_not_name_itself_or_carry_a_key_the_contract_lacks` is **rewritten**: it used `title` as its example of a key the contract lacks, briefly became legal, and is legal-again-no-more — it now uses `border` (the chrome is the host's) and still asserts a **frame** may not caption itself, since a per-tick title would let a panel relabel itself after install. `a_panel_caption_and_slash_name_are_held_to_their_shapes` was **renamed** to `a_panel_slash_name_is_held_to_its_shape` when the caption half was removed.

## Evidence

Targeted only, per CLAUDE.md. CI is this PR's evidence.

- `cargo test -p stella-plugin --features schema` → exit 0, 11 test binaries, 0 failures
- `cargo clippy -p stella-plugin --all-targets` → exit 0
- `cargo fmt -p stella-plugin --check` → exit 0
- `make prose` → OK, none added, no baseline touched
- `make wire-schema` → OK; `docs/wire/wrapper.wire.json` carries `panel_surface` and the frames' `surface`
- `make line-citations` → OK

## No new dependencies

`stella-plugin` still takes `stella-protocol` and nothing else.

## For the host branch (#5055)

Two signature changes: `PanelLease::new(panel, surface, tick, rect, budget_ms)` and `PanelFrame::new(surface, tick, paint)`. `PanelLease::admits` returns `PanelRefusal` rather than `PanelOverflow`.

## Summary by Sourcery

Extend the plugin panel contract to declare its placements and command names, route frames by surface, and protect host-rendered identity from terminal control characters and plugin-authored labels.

New Features:
- Add panel surface declarations for settings, overlay, and command placements, including optional slash-command names and consent disclosures.
- Identify the panel surface on leases and frames so hosts can route responses from plugins exposing multiple panels.

Bug Fixes:
- Reject plugin manifest names containing terminal control characters before they can be rendered in host-owned UI chrome.
- Prevent frames for the wrong surface or tick from being admitted despite matching lease geometry.

Enhancements:
- Validate panel declarations for non-empty, unique surfaces and appropriately shaped slash-command names while leaving built-in command collision checks to the host.
- Keep panel titles out of the contract to preserve host-controlled, consented panel identity.

Documentation:
- Update panel package documentation and generated wire schema for surfaces and surface-qualified frames.

Tests:
- Add comprehensive panel contract coverage for surface validation, routing, command naming, drawable manifest names, consent text, vocabulary serialization, and frame bounds.

Chores:
- Separate panel wire-contract tests from the general wire-contract test suite.


macanderson:
Residue filed rather than folded in:

- **#5210** — a plugin with two panels cannot tell its leases apart. `PanelLease` identifies its panel by the plugin's name and `PanelFrame` echoes only `tick`, so two surfaces on one plugin produce two leases that differ only in `rect`/`budget_ms` and two frames that cannot say which they answer. Not fixed here because adding a required field to `PanelLease` would break #5055's in-flight host build, and the answer partly depends on whether that host's transport pairs strictly. The issue states both options and the decision to make. Note `PanelLease::panel`'s field doc currently claims a disambiguation it no longer provides — that sentence predates multi-surface panels.
- **#5185** (existing) — `PanelText::cells()` counts `char`s rather than display width. `MAX_PANEL_TITLE_CHARS` and `MAX_PANEL_COMMAND_CHARS` are counted the same way and inherit the same caveat: a 32-`char` caption of double-width glyphs draws wider than the bound suggests. Same class, same host-side clip, no new issue.

One thing #5055 inherits explicitly, stated here so it is not discovered later: **the caption's anti-spoofing guarantee is the host's, not this crate's.** `PanelGrant::title` is safe only while the host prints `◳ panel · <plugin>` from `PluginManifest::name` *first* and the caption beside it. If the host ever renders the caption alone, or ahead of the identity, a panel can call itself whatever it likes and this contract will not have stopped it. The doc comment on `PanelGrant` says so; a host-side assertion on the label order would be worth having.

macanderson:
**Sourcery status, for the next reviewer:** its "Assessment against linked issues" table shows three ✅ and no ❌, so there is nothing to settle under CLAUDE.md's rule — but that table is not a review. Sourcery's actual review did **not run**:

> Sorry @macanderson, you've used your own review budget of 250,000 diff characters for the last 7 days. You can request another review in 3 days and 18 hours by commenting `@sourcery-ai review`.

Zero inline comments were posted. So "Sourcery reviewed this" would be a false claim, and I am not making it.

The table is also **stale in one respect**: it was generated against the first commit (`52cbe1f64`) and credits the PR with adding a `title` field. That field was subsequently **removed** in `56b48dd90` — deliberately, as the anti-spoofing rule, with the reasoning restored to `PanelGrant`'s doc comment and a witness asserting a `title` key is now refused. Rows 1 and 2 of that table therefore describe a diff that no longer exists in part. They are ✅ for work that was done and then correctly undone, not for work still present.

The branch has moved twice since that assessment: `56b48dd90` (title removed, name drawability check added, `surface` added to `PanelLease`/`PanelFrame`) and `241827e88` (panel tests split into `tests/panel_contract.rs`, two private-intra-doc-links fixed). All checks are green on the current head — 17 success, 0 failure.

PR #5211, opened 2026-08-27
feat(stella-tui): graph edges name the turn that touched their file

### LINKED ISSUE #5045: Graph edges cannot say which turn touched them
## Problem

SPEC 9.1: reverse edges on the graph node card may carry session tags (`← stella-cli::self_driving_cmd · edited turn 14`, rendering `04-graph`). Structurally impossible today: `GraphEdge` is `{from, to, kind}` and `GraphNode` has no turn field (`crates/stella-tui/src/graph.rs`), so `render_card` emits the edge with no tag. Also the tab footer reads `every answer here is deterministic · $0.00` without the spec's trailing `· 12ms` (the ms lives in the query bar; put the timing in both places or record the deviation in the spec).

## Files

- `crates/stella-tui/src/graph.rs` — `GraphSnapshot`/`GraphEdge` carry an optional session-touch tag (turn number + verb)
- The snapshot producer in `crates/stella-cli` — stamp edges touched this session from the file ledger the `● hot` marker already uses
- `crates/stella-tui/src/views/graph_tab.rs` — render the tag; footer timing

## Done means

An edge whose file was edited this session renders `edited turn N` on the node card, golden-tested; footer carries the timing when the producer reports one.

Spec: `design/tui-v2/SPEC.md` §9.1. Related: #4818.

### Definition of done

Added after the fact: this issue predates SCR-003, which asks every issue to
carry a checklist a closer can verify line by line. The items restate "Done
means" above and the task template's standard rows; nothing here is new scope.

- [x] An edge whose file was edited this session renders `edited turn N` on the node card
- [x] The tag is golden-tested
- [x] The footer carries the timing when the producer reports one, and nothing when it does not
- [x] The deviation from the plan above (where the tag lives, who stamps it) is recorded in the spec or an ADR
- [x] Scoped tests added and passing
- [x] Full CI green

macanderson:
## What & why

SPEC 9.1's node card may tag an edge with what this session did to the file at its far end — `← stella-cli::self_driving_cmd · edited turn 14`. Nothing could supply it: `GraphNode` carried no turn, so `render_card` drew every edge as a bare relation and the tag was structurally unreachable rather than switched off. The tab footer likewise stopped at `$0.00`, dropping the spec's trailing `· 12ms`.

- `crates/stella-tui/src/graph.rs` — `GraphNode` carries `touch: Option<SessionTouch>` (a turn and the `FileChangeKind` that reached the file); `GraphSnapshot::stamp_session_touches` writes it from a `&[FileTouch]` slice of the session's file ledger; `GraphNode::lives_in` is the one matcher both the `● hot` mark and the tag go through.
- `crates/stella-tui/src/model/file_state.rs`, `model.rs` — `FileState::touched_turn`, stamped in `touch_file` from the same `turns_completed + 1` SPEC 6.1's opening rule uses, so the tag and the rule above the edit in the transcript name one turn.
- `crates/stella-tui/src/views/graph_tab.rs` — `render` re-stamps from the ledger every frame, `render_card` renders the tag, the footer states `· 12ms`.
- `design/tui-v2/SPEC.md` §9.1 — the tag's exact shape, and the "a measured number or none" rule the timing and the turn both follow.

Closes #5045
Refs #5220, #5221

## Deviation from the issue's plan, and the ADR that records it

The issue asks for the tag on `GraphEdge`, stamped by the snapshot producer in `stella-cli`. Neither is what landed, and `docs/adr/0019-graph-session-touch-tags.md` is the reviewable record rather than this description alone (standing decision SD-2).

1. **The producer cannot know a turn.** A turn ordinal is `SessionModel::turns_completed + 1`, folded in `stella-tui`. `agent::graph_view` reads `codegraph.db` and holds no session state. Reconstructing the ordinal from the store's `files_touched` rows would be the *second* source of truth the issue's own constraint forbids — the ledger the `● hot` mark reads is deck-side.
2. **A snapshot-stamped tag goes stale.** A neighborhood is queried once and then sits on screen; the `● hot` mark beside it has always been derived live. A tag baked in at query time would still say `edited turn 3` after turn 9 edited the same file, with the mark next to it already moved on. So the deck stamps per frame, from the ledger, at the moment of drawing.
3. **The tag is on the node, not the edge.** An edge has two ends; the card always cites the end that is *not* under the cursor, and which end that is depends on the cursor rather than on the edge. A node-side tag has no such ambiguity, and it is also what SPEC 9.1's node list asks for — which #5220 will now pick up for free.

## The witness

`views::graph_tab::tab::an_edge_whose_file_was_edited_this_session_names_the_turn_that_did_it` — thirteen `TurnComplete` events then a `FileChange`, folded through `WorkspaceModel::apply_inbound`, and the card must read `calls → run · edited turn 14`. Driven through the fold, so the number is the one the fold counts and not a fixture's opinion.

**Flip verified artisanally, three ways:**

- Removed the tag block in `render_card` → the witness FAILED (`calls → run`, no tag) and so did the `tab_graph_touched` golden; restored → both passed.
- Replaced `let turn = self.turns_completed.saturating_add(1)` with `let turn = 1` in `touch_file` → the witness FAILED with `edited turn 1` against `edited turn 14`, and the ledger unit test FAILED with `edited turn 1` against `edited turn 3`. The witness pins the turn's *source*, not just a tag's presence.
- Removed the footer's `query_ms` block → `the_footer_carries_the_query_time_only_when_the_producer_measured_one` FAILED, and so did `tab_graph_timed` and `tab_graph_query`; restored → all passed.

Four more beside it: `a_touch_from_before_a_conversation_reset_marks_hot_without_a_turn`, `the_ledger_carries_each_paths_turn_and_the_verb_that_reached_it` (a read after an edit keeps the edit's turn; a second edit moves it; a read-only path names no turn), `a_node_matches_the_file_it_is_the_file_named_by_or_defined_in`, and `the_ledger_stamps_the_nodes_it_names_and_leaves_the_rest_alone`.

## Goldens

One new, two re-blessed, diffs read before committing:

- **new** `tab_graph_touched.txt` — the node card with `uses → Router · edited turn 1`, `Router` marked `● hot` in the list, footer `every answer here is deterministic · $0.00 · 12ms`. A golden rather than a `contains`, because the tag lands at the end of an already-full edge row and a substring match would pass just as happily with the turn pushed off the panel's right edge.
- `tab_graph_query.txt`, `tab_graph_timed.txt` — one line each, the footer only. `tab_graph.txt` and `tab_graph_query_box.txt` are untouched, which is the elision holding: those snapshots carry no timing and their footers still stop at the price.

## Two fixes that came with the shared matcher

- **A read no longer carries an older edit's turn forward.** `touch_file` first stamped the turn before the read guard, so a file edited in turn 2 and read in turn 5 rendered `edited turn 5` — the turn of a read under the verb of an edit. The turn now moves beside `kind`, which was already mutation-only for the same reason.
- **`lives_in` anchors at a separator.** The old `is_hot` matched `my_lib.rs` against `lib.rs` and a `src/lib.rs.bak:3` location against `src/lib.rs`. It is also allocation-free now, which matters over every node × every ledger row per frame.

## Two things that are housekeeping, not feature

- `model.rs` was eight lines under its 1500-line ceiling and is not grandfathered, so `reset_conversation` moved to `model/reset.rs` — a split, never a baseline entry.
- `main` took ADR 0018 while this branch was open, so this one is 0019. #5175 already tracks the missing guard that would have caught the collision.

## This branch had an earlier attempt on it

`d90e8a170` implemented #5045 independently and reached the same architecture — render-time join against the per-lane ledger, `GraphEdge` left structural, footer timing under the query bar's rule. Two of its decisions are **adopted here rather than reimplemented**: the turn is stamped on mutations only, and a path this session only read is `● hot` with no tag. Its own `RememberedDiff::turn` is not carried — nothing renders an older mutation's turn, and `RememberedDiff` is the byte-budgeted record whose docs count the 12 bytes an entry costs.

It was first merged into this history with `-s ours`, and `check-rebase-replay` correctly refused that: it reads an edit-then-undo pair inside a branch as a live revert waiting to be replayed onto a base that already carries the edit (#4979). The branch is a single flattened commit now, whose message carries the same comparison. `d90e8a170` is still reachable from this PR's timeline.

**Its four tests are not in this tree** (they name an API this tree does not have: `FileState::edited_turn`, `touch_of`, `SessionEdit`). Every behaviour they pinned is pinned here — the read elision, the twice-edited path, the turn-14 witness, the spec's tag shape — by the tests listed above. Naming it explicitly because `check-deleted-tests` compares against `main`, where those tests never landed, so nothing else would say so.

## Tests removed, named for the reviewer

One test is deleted rather than renamed in place, so `check-deleted-tests` reports it and this is the sentence it asks for:

- **`views::graph_tab::tests::a_node_is_hot_when_its_file_changed_this_session`** — it pinned the old `is_hot(node, &[String])` helper, which no longer exists: node-to-file matching moved to `GraphNode::lives_in` so the `● hot` mark and the session tag cannot disagree about which node belongs to which file. Its coverage is now `graph::tests::a_node_matches_the_file_it_is_the_file_named_by_or_defined_in`, which pins the same three cases (the node *is* the file, the basename, a symbol's location) and adds the two near-misses the old helper got wrong.

No other test was deleted.

## The gate

- [x] `cargo fmt --check`
- [x] `cargo clippy -p stella-tui -p stella-cli --all-targets -- -D warnings`
- [x] `cargo test -p stella-tui` (all targets green); `cargo test -p stella-cli --bin stella graph_view`
- [x] `make guards-fast` (green), `RUSTDOCFLAGS="-D warnings" cargo doc` for both crates
- [x] Docs updated: SPEC §9.1, the tui-v2 implementation plan, ADR 0019, module and item docs
- [x] `Closes #5045` appears above and as a commit trailer

Not run here: the workspace build and test suite — that is CI's, per CLAUDE.md.

## Nothing left behind

- [x] Filed: **#5220** (the node list's `● hot` still names no turn — SPEC 9.1's left pane, now one render away since the data is on the node), **#5221** (`stamp_session_touches` matches every node against every ledger row per frame; bounded at 256 × 2048 and unmeasured).
- Linked, not duplicated: **#5175** for the ADR numbering collision.

The stamp is a projection recomputed per frame, deliberately not cached — #5221 is where that trade gets measured rather than assumed. If you would rather have the field on `GraphEdge` after reading ADR 0019, say so and I will move it; flagging rather than deciding it silently.

## One edit to the issue itself

#5045 predates SCR-003 and carried "Done means" prose rather than a checklist, so `dod-check` could not verify a close against it. A **Definition of done** section was appended to the issue restating those same items plus the task template's standard rows — no new scope, and the box for full CI green stays unticked until this PR's run is green.

## Summary by Sourcery

Add render-time graph session-touch tags and measured query timing to the Stella TUI graph tab.

New Features:
- Show the turn and file-change action for touched graph nodes on node-card edge relations, while keeping untouched and post-reset touches untagged.
- Display measured graph query time in the graph tab footer when available.

Bug Fixes:
- Keep mutation turns aligned with the mutation that last changed a file so later reads do not relabel older edits.
- Prevent false-positive node-to-file matches for similar filenames and path prefixes.

Enhancements:
- Stamp graph session-touch metadata at render time from the focused session ledger so hot markers and tags share one current source of truth.
- Preserve session hotness across conversation resets while clearing obsolete turn ordinals.
- Move conversation-reset logic into a dedicated model module to maintain the model file size limit.

Documentation:
- Document the graph session-tag behavior and timing rules in the TUI specification, implementation plan, and ADR 0019.

Tests:
- Add unit, witness, and golden coverage for session touch verbs, turn tracking, reset behavior, node matching, edge tags, and conditional query timing.

Chores:
- Update graph fixtures and consumers for the new optional node touch metadata.


macanderson:
@sourcery-ai review

The head this PR carried when you last reviewed it has been replaced: the branch is now a single flattened commit (`dbb40e423`) with a different implementation of the same objectives, plus a read/turn fix, a matcher fix and an ADR. Re-reading the assessment against #5045 on the new diff.

Issue #5214, opened 2026-08-27
SPEC 10 promises `/gates` in `relevant now`, and neither the rule nor the command exists


macanderson:
## Problem

SPEC 10 states the palette's context section by example:

> **Context section first**: `relevant now` derives from session state (verify running surfaces `/gates` with live gate status on the row).

Neither half is true in the tree:

- `crates/stella-tui/src/composer/palette.rs` `RULES` has six rules — running turn, sub-agents, open plan, changed files, unread, unindexed workspace. None fires on a **verify** step, which is the one the spec names.
- `rg /gates crates/` is empty. `DECK_BUILTINS` (`crates/stella-cli/src/command_deck/skills.rs`) offers no `/gates`, so even a rule that named it would promote a row nothing answers — which is exactly the drift `every_command_a_relevance_rule_can_name_is_a_real_one` exists to catch.

The neighbouring promise on the same bullet — live right-aligned state on a row — *is* implemented (`deck_render::slash_live_hints` feeds `/inbox · 3 unread` and `/plan`'s step count), so the shape is there; the gate case is the missing one.

## Done means

Either:

1. `/gates` joins the deck vocabulary (it needs a home — the gate board is currently reached through the tab surfaces, not a slash command), a relevance rule fires on a running verify, and `slash_live_hints` gives its row the live `◐ 2/5 green` the spec draws; **or**
2. SPEC 10's bullet is amended to name an example the deck actually has, the way #5048 amended the overlay and matcher bullets in the same section — with a guard in `crates/stella-tui/src/render/tests/slash.rs` beside the two already there, so the document cannot drift back.

Pick one and record it. Silently leaving the spec naming a command that does not exist is the outcome to avoid.

## Notes

`PaletteState` is the input to the rule table and would need a verify-running fact (it currently carries `turn_running`, which is any turn). The table is a pure function over owned data, so the new case is a test row, not a screenshot.

Filed as residue of #5048.

## Definition of done

- [x] SPEC 10's `relevant now` bullet names an example the deck actually has, amended in the #5048 style with the reason recorded.
- [x] The deferred intent is pointed at the issue that would deliver it (#5042) rather than deleted.
- [x] A guard holds the document against the rules, beside the two already in `render/tests/slash.rs`.
- [x] The guard's flip was checked artisanally: it fails on the old sentence, naming `/gates`.

Resolved by option 2 — `/gates` needs a gate board first. Closed by #5246.

Issue #5212, opened 2026-08-27
The `08-command-palette` rendering still draws the palette #5048 replaced


macanderson:
## Problem

`design/tui-v2/renderings/{png,svg}/08-command-palette.*` are the normative pictures SPEC 10 points at, and they now describe a palette the deck does not draw. #5048 changed three things the rendering shows:

- **Position.** The rendering shows a centered `Rect`. SPEC 10's overlay bullet was amended in #5048 (with the guard `spec_10_anchors_the_palette_to_the_composer`) to say the overlay anchors to the composer and opens upward, because the query is typed into the composer and a centered box splits the two apart. The picture still shows the old shape.
- **Matched letters.** The rendering lights a typed prefix. The palette now lights every matched letter wherever it sits in the name — `ga` puts gold on the `g` and the `a` of `/graph query`.
- **`recent`.** The rendering has no `recent` section. The browse list now ends with one, fed from `.stella/private/palette-recents.json`.

Nothing is wrong with the code; the picture is stale. A reader who trusts the rendering over the prose gets the pre-#5048 palette.

## Done means

Both files regenerated from the shipped deck, showing: the anchored overlay on its `token::PANEL` ground, a fuzzy query with non-contiguous gold letters inside a command name, and the `recent` section under its heading below the domain groups. SPEC 10 and the rendering agree again.

## Notes for whoever takes it

This needs the rendering toolchain, not a code change — which is why #5048 left it. Check whether the other `design/tui-v2/renderings/` files have drifted the same way while you are in there; the spec-guard pattern (`keymap.rs`, and now `render/tests/slash.rs`) only holds the prose, and nothing holds the pictures.

Filed as residue of #5048.

## Definition of done

The “Done means” prose above, as boxes the gate can read — nothing here is
new work, and nothing is ticked by the migration (#5193).

- [x] Both files are regenerated from the shipped deck.
- [x] The rendering shows the anchored overlay on its `token::PANEL` ground, a fuzzy query with non-contiguous gold letters inside a command name, and the `recent` section under its heading below the domain groups.
- [x] SPEC 10 and the rendering agree again.

macanderson:
Partly done, and the remainder needs a decision this issue and #5215 cannot both have.

[...]

Issue #4394, opened 2026-08-23
code-graph watcher resolves codegraph.db by a bare path join while the index build honours the workspace state root — a redirected session indexes one db and watches another


macanderson:
## Problem

Two resolvers for the same file disagree about `STELLA_WORKSPACE_STATE_ROOT` (`stella_home::WORKSPACE_STATE_ROOT_ENV`, the redirect `stella self-driving` sets so a throwaway worktree keeps its private state outside the tree):

- The index build at `crates/stella-cli/src/agent/graph.rs:355` resolves through `stella_store::workspace_private_sqlite_path(workspace_root, "codegraph.db")`, which honours the redirect (see `crates/stella-store/src/private.rs:268`).
- The watcher mount at `crates/stella-cli/src/agent/graph.rs:729` calls `crate::search_cmd::codegraph::graph_db_path(&root)`, which is `crates/stella-tools/src/search/codegraph.rs:14` — a bare `root.join(".stella").join("private").join("codegraph.db")` that ignores the redirect. The same resolver is used at `crates/stella-cli/src/contextgraph.rs:273` and `crates/stella-cli/src/command_deck.rs:3218`.

Under a redirect the session builds its index in one database and mounts the watcher (and the CGP host, and the deck) on an empty one at the literal path, so the agent's own edits are never indexed where the queries look. Found during the 2026-08-23 backlog triage of #4003.

## Done means

- One resolver. `graph_db_path` either delegates to `stella_store::workspace_private_sqlite_path` (it already returns a `Result` — callers must handle the setup-error arm, which `IndexError::PrivateState` exists for) or is deleted in favour of it. stella-tools must not grow a new workspace dependency to do this — check the crate graph first; if `stella-store` is not already a dependency of `stella-tools`, move the resolver up to the crate that has it.
- Witness: with `STELLA_WORKSPACE_STATE_ROOT` set, the path the build writes equals the path the watcher mounts. Fails today.
- `command_deck.rs` is a god file: do not add lines there.

Related: #4003, #4300 (self-driving worktree state), #4362 (reflections under a redirected root — same family).

PR #5217, opened 2026-08-27
feat(stella-tui,stella-cli): the palette lights what it matched and remembers what ran

### LINKED ISSUE #5048: The command palette lights only typed prefixes and has no recent section
## Problem

SPEC 10 gaps in the command palette (rendering `08-command-palette`):

1. Matched letters should render gold inside each command name; today only a typed *prefix* is lit, because the hand-rolled matcher (`SlashMenu::filter_with`) produces no match indices. The spec names `nucleo`; either take it (one-sentence dependency justification per repo rule) or extend the matcher to emit indices — decide in the PR.
2. The `recent` section is absent (needs per-workspace persistence; the doc comment cites #4338, which is closed — this is the untracked remainder).
3. The overlay anchors to the composer's left edge and opens upward with no `token::PANEL` background; the spec says centered `Rect`, `panel` bg. If the anchored position is the better design, amend the spec in the same PR (the spec-test pattern in `keymap.rs` shows how deviations are recorded).

## Files

- `crates/stella-tui/src/composer.rs`, `composer/palette.rs` — matcher indices, recents
- `crates/stella-tui/src/render.rs` — `slash_popup_area`, gold match letters, panel bg
- Persistence for recents — smallest honest store available; name it in the PR

## Done means

Fuzzy query `ga` lights `g` and `a` inside `/graph query` (not just prefixes), snapshot-tested; a `recent` section lists the last-run commands across sessions; position/bg matches the spec or the spec records the deviation.

Spec: `design/tui-v2/SPEC.md` §10.

## Definition of done

- [x] Implementation complete — matched letters light gold at their real positions (scattered included), a `recent` section lists the last commands across sessions, and the overlay carries the `panel` ground with its position either matching SPEC 10 or amending it
- [x] Scoped tests added/updated and passing (`cargo test -p stella-tui`)
- [x] Full CI green
- [x] Code comments and docs updated — `design/tui-v2/SPEC.md` §10 and `IMPLEMENTATION-PLAN.md` describe what ships

macanderson:
## What & why

SPEC 10's command palette was three gaps short of the spec, and each one turned out to be a decision rather than a task. All three are recorded in **ADR 0019** (`docs/adr/0019-command-palette-matching-and-recents.md`).

Closes #5048

### 1. Matched letters — the matcher stays in-tree; `nucleo` was considered and declined

`crates/stella-tui/src/composer/fuzzy.rs` is a new ~60-line pure module over two `&str`s. It keeps the palette's existing tiers, inserts a **subsequence** tier between "name substring" and "description", and every tier now reports the byte offsets of the characters it matched.

Two things follow. `ga` now *reaches* `/graph query` at all — the old matcher was prefix / substring / description and had no tier for scattered letters. And the renderer lights the `g` and the `a` where they sit, instead of re-deriving a `starts_with` and lighting a typed prefix only; a mid-name match used to appear in the list with no visible reason for being there.

**The decision.** SPEC 10 named `nucleo`. Its licence was not the obstacle (MPL-2.0, already on `deny.toml`'s allow-list). Declined for three reasons, in descending weight:

1. **The palette does not rank by fuzziness.** It ranks by *what* matched — a prefix beats a substring beats a scattered match beats a description hit — and then reorders inside each tier by what the session is doing (`relevant_now`) and what has been run here before. That policy is a product decision with tests pinning it. `nucleo` supplies one opaque score for all of it, so taking it would have meant either discarding the policy or running a whole matcher to recover a *by-product* — the indices — of a decision it did not make. A dependency held for its side effect is the shape that rots.
2. **It stays a pure function.** Every case is a table row in a unit test rather than a screenshot, the same property that makes `relevant_now` reviewable.
3. **The corpus is thirty ASCII slugs.** `nucleo` is built to rank a hundred thousand paths at frame rate; its `Matcher` is `&mut self` for the scratch buffers that make that possible, and `crate::render` is documented `&`-only with no interior mutability (its panel-panic boundary rests on it).

Reason 3 alone would not justify refusing a dependency the spec asked for. Reason 1 would. The spec bullet was amended to name what is actually here, and `spec_10_names_the_matcher_the_palette_uses` fails if it drifts back.

### 2. `recent` — the store, named

**`stella_store::workspace_private_state_path`**, under `palette-recents.json` — the owner-only `<workspace>/.stella/private/` tier that already holds `reflections.jsonl`, the self-tuning ledger and `mcp_oauth.json`. **No new store.** That tier is created 0700 and validated no-follow, is gitignored by construction, and follows `STELLA_WORKSPACE_STATE_ROOT`, so a turn running in a throwaway worktree writes its history to the real repository rather than into a directory built to be deleted (#4394).

The deck touches no file. `stella-tui` owns no store and must not learn where a workspace keeps its files, so the list travels the route the command vocabulary already travels: the driver pushes `Inbound::PaletteRecents`, the deck reports a run back with `WorkspaceInput::PaletteRan`, and `stella_tui::palette_remember` is the one move-to-front rule both sides apply so they cannot disagree about what "recent" means.

The run is recorded in `submit_prompt` — the one place every submission passes — rather than in the popup's key handler. Two routes would otherwise be missed: a command completed with `⇥` and given arguments closes the popup before Enter, and a command typed out in full never opened it. Folding the history from the *prompt stream* instead would miss a third of the vocabulary in the other direction, since `/files`, `/diff`, `/graph` and the rest of the tab switches are consumed deck-side and never reach the driver.

This is the untracked remainder of #4338, whose doc comment said the deck had no store for it.

### 3. Position — the anchored overlay wins, and the spec is amended

The palette stays anchored to the composer's left edge, opening upward. **SPEC 10's overlay bullet was amended in this PR to say so**, with `spec_10_anchors_the_palette_to_the_composer` holding the document and the deck together — the shape `keymap.rs` established for SPEC 11's plan chord (#4341).

The reason: this palette has no input line of its own. Every centered palette it resembles owns its query field; here the query is typed into the **composer**, pinned to the bottom of the frame, so a centered box put the letters being typed and the list they filter at opposite ends of the screen with the caret nowhere near the rows moving under it. Growing the list upward out of the text that produced it is the completion-popup shape, and that is what the surface is.

The other half of that bullet was right and now ships: the overlay paints `token::PANEL` under every cell it covers, so it reads as a surface lifted off the transcript rather than a bordered hole punched in it. The panel ground is set once on the widget, not on the spans, so the selected row's `token::HL` still wins — asserted.

### Also in here

`composer.rs` crossed the 1500-line guard, so the slash menu moved to `composer/menu.rs`. The split falls where the concerns already did: `composer.rs` is the input model (buffer, cursor, chips, soft-wrap), `menu.rs` is the menu built over a caller-supplied vocabulary. Everything is re-exported, so `composer::SlashMenu` still resolves.

`SlashMenu::matches` is now `Vec<SlashMatch>` rather than `Vec<&SlashCommand>`: a match's lit offsets travel with the command they belong to instead of in a parallel `Vec<Vec<usize>>` that a later sort could take out of step. `PaletteState` gains `recent` and loses `Copy`; it is read by reference everywhere, so the clone is paid once per frame in `palette_state`.

## The witness

**`a_fuzzy_query_lights_the_matched_letters_inside_the_name`** (`crates/stella-tui/src/render/tests/slash.rs`) — renders the palette for the query `ga` over `/graph query` and asserts that the cells carrying `token::GOLD_BRIGHT` spell exactly `ga`, that `r`, `ph` and `query` are still plain gold, and that no lit letter is also plain.

**Verified it flips.** With the subsequence tier removed from `match_name` (and nothing else changed) it fails at the first assertion — `` `ga` reaches `/graph query`: left: 0, right: 1`` — because on the old matcher the query reaches no command at all. Restored, it passes.

Beside it, on the same behaviour from other angles:

| test | file | what it pins |
|---|---|---|
| `a_scattered_query_lights_the_letters_it_matched` | `composer/fuzzy.rs` | the offsets, at the unit level |
| `a_substring_lights_where_it_sits_not_the_head` | `composer/fuzzy.rs` | a mid-name match no longer lights the head — the old renderer's bug |
| `a_multibyte_name_reports_char_boundaries_only` | `composer/fuzzy.rs` | no offset splits a character |
| `what_one_session_records_the_next_one_reads_back` | `command_deck/palette_recents.rs` | **across sessions** — record, then read back with nothing shared but the file |
| `the_history_lives_in_the_private_tier` | `command_deck/palette_recents.rs` | it lands in gitignored `private/`, not loose in `.stella/` |
| `the_browse_popup_ends_with_the_workspaces_recent_commands` | `render/tests/slash.rs` | the section renders, once per command, at the tail |
| `running_a_palette_row_records_it_locally_and_tells_the_driver` | `deck_ui/tests/queue.rs` | end to end on a **deck-local** command, which the prompt stream never sees |
| `the_palette_paints_a_panel_ground_without_losing_the_selection` | `render/tests/slash.rs` | `token::PANEL` under the border, `token::HL` still on the selected row |
| `spec_10_anchors_the_palette_to_the_composer` | `render/tests/slash.rs` | the spec bullet and `slash_popup_area` agree |
| `spec_10_names_the_matcher_the_palette_uses` | `render/tests/slash.rs` | the spec no longer sends a reader looking for `nucleo` |

## The gate

- [x] `cargo fmt --check`
- [x] `cargo clippy -p stella-tui -p stella-cli --all-targets -- -D warnings` (scoped per SD-1; the workspace run is CI's)
- [x] `cargo test -p stella-tui` (1232 lib + 80 integration, all green) and `cargo test -p stella-cli --bin stella command_deck` (174 green)
- [x] `RUSTDOCFLAGS="-D warnings" cargo doc -p stella-tui -p stella-cli --no-deps --document-private-items`
- [x] `make guards` — **all 49 guard steps green**, including `prose`, `file-size`, `god-files`, `invariants`, `doc-links`, `tokens` and `line-citations`
- [x] Docs updated: SPEC 10 amended, ADR 0019 added and listed in `docs/adr/README.md`, `docs/manifest.json` regenerated with `make doc-links-fix`
- [x] `Closes #5048` appears above **and** as a commit trailer

### Two guards this branch had to be re-shaped for, worth a reviewer's eye

**`file-size`.** `composer.rs` crossed the 1500-line limit, so the slash menu moved to `composer/menu.rs` (see above). And `deck_ui.rs` is a grandfathered god file closed to growth — AGENTS.md § "God files" says to plan around it, not into it — so the palette's recorder lives in a new sibling, `deck_ui/palette.rs`, and `deck_ui.rs` gains fourteen lines rather than fifty. No baseline entry was added and no ceiling was raised.

**`prose`.** The per-crate module-header density ratchet (#5096) meant the new modules' headers had to earn their lines. The `nucleo` argument moved from `fuzzy.rs`'s header into ADR 0019, where it belongs; `stella-tui` finishes **under** its ceiling. (The branch was cut before #5167 unbroke this guard on `main`; after rebasing onto `0aeee2c5e` it is green.)

## Nothing left behind

- [x] Filed: **#5212**, **#5214**

- **#5212** — `design/tui-v2/renderings/{png,svg}/08-command-palette.*` still draw the pre-#5048 palette: centered overlay, prefix-only highlighting, no `recent` section. The prose is now guarded; the pictures are not, and regenerating them needs the rendering toolchain rather than a code change.
- **#5214** — SPEC 10's `relevant now` bullet names `/gates` with live gate status on the row. No relevance rule fires on a running verify, and `rg '"/gates"' crates/` is empty, so the spec names a command the deck does not offer. Either build it or amend the bullet with a guard, the way this PR did for the other two.

Also read while here: **#5136** (the prose ratchet's shared-cell failure). Not re-filed — it already describes the failure mode this branch met before rebasing, and `main` has since been repaired by #5167.

## Ground-rule check

- [x] **No new dependency.** The one this PR could have taken — `nucleo` — was declined on the record; see §1 and ADR 0019
- [x] No new cross-boundary serde types (`palette-recents.json` is a `Vec<String>` through `serde_json`, read defensively: a corrupt or over-long file is treated as absent and healed on the next write — tested)

**Two calls are yours to overturn, and both are cheap to reverse.**

The `nucleo` refusal is the bigger one. If the palette should rank by score rather than by tier — a different product decision, not a different implementation — then `nucleo` is right and `fuzzy.rs` should go. Everything downstream reads `NameMatch { tier, lit }`, so the swap is behind `match_name`.

The anchored position is the smaller one, but note the deliberate asymmetry: I amended the spec rather than moving the overlay, so the *document* changed to match shipped behaviour. If a centered palette is wanted, `slash_popup_area` is six lines and the guard tells you which bullet to put back.

**Deliberately out of scope.** `PaletteRan` is advisory — it reorders a menu and does nothing else — so a driver that drops it costs an ordering, never an action; the demo driver in `examples/deck_demo.rs` does exactly that. And a history entry naming a command the vocabulary no longer offers matches nothing and disappears, which is the right behaviour for an uninstalled skill (tested).


macanderson:
Heads up, and an apology for the collision: **#5219 implemented the same issue and I merged it a few minutes ago**, before I saw this PR. That was not a judgement about this one — I did not know it existed. Two sessions took #5048 in parallel; the fourth such collision today, which is now filed as #5224.

The two implementations reached the **same three decisions independently**, which is worth recording: an in-tree matcher with `nucleo` declined, recents persisted rather than deferred, and the anchored (not centred) popup position amended into SPEC 10 rather than left as a silent deviation. Both even landed the matcher in `composer/fuzzy.rs`.

**This PR has something #5219 does not: ADR 0019.** #5219 recorded its reasoning in the spec bullet and in `slash_popup_area`'s doc comment; an ADR is the better home for a decision of this shape, and I would rather have it than not. If you rebase, the ADR should apply cleanly on its own — the code beneath it is now redundant, but the decision record is not.

Also worth diffing before you discard anything: #5219 pinned the highlight with a per-cell foreground assertion requiring `GOLD_BRIGHT` on exactly the matched letters and resting `GOLD` between them, and made one deliberate behaviour change — the leading `/` no longer lights with the letters after it, since it is the sigil that opened the palette rather than a letter the query consumed. If your matcher decided that differently, that difference is worth a look rather than a rebase.

Sorry for the wasted work. The claim mechanism in #5224 is the fix, and it is filed as P1.

macanderson:
Triage note from a CI sweep — facts only, nothing rebased or closed. This PR needs a decision I should not make on someone else's design.

**This PR and #5219 are two independent implementations of the same feature.** #5219 — *"the palette lights matched letters and keeps a recent list"* — merged at 02:55, 36 minutes after this one was opened at 02:19. Its title and this one's (*"the palette lights what it matched and remembers what ran"*) describe the same two behaviours.

The overlap is now six conflicts against `main`:

```
CONFLICT (content): crates/stella-tui/src/composer.rs
CONFLICT (add/add): crates/stella-tui/src/composer/fuzzy.rs
CONFLICT (content): crates/stella-tui/src/lib.rs
CONFLICT (content): crates/stella-tui/src/render.rs
CONFLICT (content): crates/stella-tui/src/render/tests/slash.rs
CONFLICT (content): design/tui-v2/SPEC.md
```

The `add/add` on `composer/fuzzy.rs` is the signature: both branches authored that file independently. `main`'s copy came from #5219 (5c5404bf8); this branch's differs from it by 228 insertions and 96 deletions, so they are not the same code arriving twice — they are two designs.

**They also diverge structurally**, which is why this is not a mechanical rebase:

- recents live in `crates/stella-cli/src/command_deck/palette_recents.rs` here, and in `crates/stella-tui/src/composer/recent.rs` in #5219;
- #5219 also touched `composer/tests.rs`, `deck_shell.rs` and `IMPLEMENTATION-PLAN.md`, none of which this branch has.

**This branch is not simply superseded.** It carries work #5219 never touched — `composer/palette.rs`, `composer/menu.rs`, `deck_ui/palette.rs`, `deck_render.rs`, `envelope.rs`, `deck.rs`, `examples/deck_demo.rs`, `deck_ui/tests/queue.rs`. So closing it would drop real content, and rebasing it wholesale would relitigate the half that already landed.

The call — keep #5219's fuzzy matcher and rebase only this branch's unique half onto it, or replace #5219's with this one — is a design judgement about your own code, so I have left the branch untouched.

Worth noting for the pattern rather than this PR: this is the third duplicate-implementation collision in a few hours (#5167/#5170/#5171 were three separate repairs of one red `main`, and #5165 and #5211 each had a peer push a fix while another session was writing the same one). `scripts/main-red-claim.sh` covers the red-`main` case; nothing covers two sessions independently picking up the same feature issue, which is what happened here and cost two full implementations.

macanderson:
Superseded by #5219, merged as `5c5404bf8`, which closed #5048.

Both branches implemented SPEC 10 independently and well — and reached the same headline judgement, that the palette should keep its own matcher rather than take `nucleo`, because the ranking is a product decision (`relevant_now` plus tier) that a fuzzy-finder's single opaque score cannot express.

The one design idea here that #5219 does not carry is the store: this branch routed recents through `stella_store::workspace_private_state_path`, the hardened owner-only tier that is created 0700, validated no-follow, and follows `STELLA_WORKSPACE_STATE_ROOT` — so a turn in a throwaway worktree writes its history to the real repository (#4394) instead of into a directory built to be deleted. #5219 resolves the path CLI-side and passes it in through `DeckOptions`, which follows the existing `debug_log_path` pattern but does not inherit that redirection.

That difference is worth keeping as a follow-up rather than losing here; filing it separately.

Issue #5215, opened 2026-08-27
The command-palette rendering shows section headings under a typed query; the code does not


macanderson:
## Problem

`design/tui-v2/renderings/svg/08-command-palette.svg` draws the palette with `/ga` typed in the composer **and** four section headings above the rows: `relevant now · verify is running`, `plan`, `graph`, `recent`.

The shipped palette does not do that. `SlashMenu::filter_with` (`crates/stella-tui/src/composer.rs`) returns an empty `sections` list for any non-empty query, so a typed query draws one flat ranked list with no headings — the deliberate decision from #4338 ("grouping a three-row result buries the rows under their own captions"), pinned by `a_queried_popup_draws_no_headings` in `crates/stella-tui/src/render/tests/slash.rs`.

#5048 amended `design/tui-v2/SPEC.md` §10 so the prose states the shipped behaviour. The rendering was not touched, so the SVG is now the only artifact claiming otherwise — and the renderings are described in §15 as the source of truth for the visual design, which is exactly the kind of quiet disagreement that gets read as a bug report later.

## What to decide

One of two, not both:

1. **The rendering is stale** — redraw `08-command-palette` with `/ga` typed and no headings (or with an empty composer and the headings). This is the cheap answer and matches what ships.
2. **The rendering is right and the code is wrong** — headings under a typed query are worth having. If so, say what a three-row result under three captions should look like, revert the `sections.is_empty()` branch in `filter_with`, and replace `a_queried_popup_draws_no_headings` with its opposite.

## Files

- `design/tui-v2/renderings/svg/08-command-palette.svg` (and its PNG export)
- `design/tui-v2/SPEC.md` §10 — already states the shipped behaviour
- `crates/stella-tui/src/composer.rs` — `SlashMenu::filter_with`
- `crates/stella-tui/src/render/tests/slash.rs` — `a_queried_popup_draws_no_headings`

## Done means

The rendering, the spec prose and the code agree about whether a typed query draws section headings, and whichever changed carries the reason.

Related: #5048, #4338.

## Definition of done

- [x] The rendering, SPEC 10 and the code agree: a typed query draws one flat ranked list with no section headings.
- [x] The rows shown are the ones that actually fuzzy-match the typed query, and the header count follows.
- [x] The PNG is re-exported at the 2x factor the rest of the directory uses.
- [x] The caption states the rule the picture demonstrates.

Resolved by option 1 — the rendering was stale. Closed by #5246.

Issue #5213, opened 2026-08-27
The command palette's recent rows do not say how long ago


macanderson:
## Problem

The `recent` section landed in #5048 listing the last commands this workspace ran, most recent first. The rendering it is built from — `design/tui-v2/renderings/svg/08-command-palette.svg` — also puts a relative age on the right of each recent row (`/theme stella-gold    2m ago`). That half is not shipped: the rows carry the order and nothing else.

It was left out because the age needs two things the current shape does not have. `Recents` stores names only (a JSON array of strings in `.stella/private/palette-recent.json`), so there is no timestamp to render; and `render_slash_popup` has no clock — the existing right-aligned `live` slice is keyed by command *name*, and a recent row is a second appearance of a command a domain group already lists, so a name-keyed value would paint the age on both rows.

## What to do

- `crates/stella-tui/src/composer/recent.rs` — store `{ name, at_ms }` per entry rather than a bare name. Accept the existing bare-string file shape on read (or start the list over; it is five names) so an upgrade does not error.
- `crates/stella-tui/src/composer.rs` — `SlashMatch` needs to say that a row *is* a recent row and when it ran, so the renderer can tell it apart from the domain-group copy. A `recent_at: Option<u64>` on `SlashMatch` is the smallest shape.
- `crates/stella-tui/src/render.rs` — `render_slash_popup` renders the age right-aligned on those rows. The clock has to come in as an argument; `deck_render.rs` already holds `model.now_ms` at the call site.
- Keep the ages coarse (`now`, `2m ago`, `3h ago`, `2d ago`) — a per-second age on a repainting overlay is motion nobody asked for, and the deck freezes motion under `--no-anim` / `--accessible`.

## Verify

`cargo test -p stella-tui render::tests::slash` — a witness that a recent row carries an age and the domain-group copy of the same command does not. Check the flip artisanally (revert, watch it fail, restore).

## Done means

The `recent` rows show a coarse relative age on the right, the domain-group copy of the same command does not, and the age survives a restart because it is stored with the name.

Spec: `design/tui-v2/SPEC.md` §10. Related: #5048, #4338.

## Definition of done

- [x] `Recents` stores `{ name, at_ms }` per entry and the stamp survives a restart.
- [x] The pre-existing bare-name file shape is read rather than erroring or emptying the list; such entries carry no age rather than a guessed one.
- [x] `SlashMatch` says a row is a recent row and when it ran, so the renderer can tell it from the domain-group copy.
[...]

PR #5219, opened 2026-08-27
feat(stella-tui): the palette lights matched letters and keeps a recent list

### LINKED ISSUE #5048: The command palette lights only typed prefixes and has no recent section

macanderson:
Closes #5048

SPEC 10 named three things the command palette did not do. All three land here, and where the shipped design is the better one the spec is amended in the same PR rather than left disagreeing with the code.

## 1. Matched letters light gold inside the name

`ga` now lights the `g` and the `a` of `/graph query`. Before, only a typed *prefix* lit — and a scattered query matched no row at all, so the popup's own `· fuzzy` title was a claim nothing on screen backed.

**Decision: extended the matcher, did not take `nucleo`.** `crates/stella-tui/src/composer/fuzzy.rs` returns the char offsets a match consumed, ranked prefix → substring → scattered → description-only. A command name is a short ASCII slug, and the palette's order comes from that rank plus what the session is doing (`palette::relevant_now`), so a fuzzy-finder crate's scoring model would be computed and then discarded. **No new dependency; `cargo deny` is untouched.** A contiguous run still wins before the scattered walk is tried, so every existing prefix/substring rank and highlight is unchanged.

One behaviour change worth naming: the leading `/` no longer lights with the letters after it. It is the sigil that opened the palette, not a letter the query consumed, and the scattered case cannot pretend otherwise.

## 2. The `recent` section

The browse list closes with `recent` — the commands this workspace ran last, most recent first, capped at five.

**Store: `.stella/private/palette-recent.json`**, a JSON array of names. That is the smallest store that survives a restart and the only one `stella-tui` can reach without taking a dependency on `stella-store`; the CLI resolves the path (`palette_recent_path` in `command_deck.rs`) and hands it in through `DeckOptions::recent_path`, the way `debug_log_path` already works. `.stella/private/` is gitignored, so a clone never inherits someone else's last five commands. Both file operations are best-effort and silent — a palette that refused to open because a convenience list would not parse has traded a feature for an ornament.

The list rides the `Composer`, which already owns the slash-menu view, so the key handler and the renderer read one list without either surface passing it between them. A surface handed no path keeps an in-session list and writes nothing.

A `recent` row is a second appearance of a command a domain group already lists — that is what a shortcut is — so it is appended rather than lifted out of its group, which is also where SPEC 10 puts the section.

## 3. Position and background

**The `panel` ground is a real gap and is fixed.** `Clear` alone left the terminal's own background showing through the overlay, so the palette read as a hole punched in the deck. The `/model` argument popup takes the same ground — the two overlays are one shape and must not differ in it.

**The anchored position is the better design, so SPEC 10 is amended** (`design/tui-v2/SPEC.md`), following the pattern `keymap.rs`'s `PLAN_CARD_KEYS` doc records for the `^S` amendment: the reason lives both in the spec bullet and beside the code, in `slash_popup_area`'s doc comment. The palette completes the word under the cursor, and centering moves the completions away from the letters that produced them — every terminal and editor autocomplete anchors to the caret for that reason.

The same §10 edit records the matcher choice and the store, so the spec no longer names a dependency this tree does not take.

## Witnesses, and the flip verified artisanally

Every one below was checked by reverting the production change, watching the test fail, and restoring it.

| Witness | Reverted to check the flip | Result |
|---|---|---|
| `composer::fuzzy::tests::scattered_letters_match_and_name_their_positions` | the scattered walk in `match_name` | FAILED |
| `composer::tests::a_scattered_query_matches_and_reports_where_its_letters_landed` | same | FAILED |
| `composer::tests::a_scattered_name_match_sorts_under_the_contiguous_ones` | same | FAILED |
| `render::tests::slash::scattered_matched_letters_light_gold_inside_the_name` | same, and separately the per-letter span walk in `render_slash_popup` | FAILED under both |
| `render::tests::slash::a_prefix_still_lights_as_one_run` | the per-letter span walk | FAILED |
| `render::tests::slash::the_palette_floats_on_the_panel_ground` | the `.style(bg(token::PANEL))` on the block | FAILED |
| `render::tests::slash::the_browse_popup_closes_with_the_recent_section` | the `recent` section append in `filter_with` | FAILED |
| `composer::tests::running_a_row_records_it_and_completing_one_does_not` | same | FAILED |
| `composer::recent::tests::the_list_survives_the_session_that_wrote_it` | the write in `Recents::flush` | FAILED |

The fuzzy-highlight witness asserts scattered letters, not a prefix: it reads the buffer's per-cell foreground and requires `token::GOLD_BRIGHT` on exactly the `g` and the `a` of `/graph query`, with the slash and every letter between and after them at resting `token::GOLD`.

## Evidence

- `cargo test -p stella-tui` — exit 0 (1227 tests). The `deck_render_snapshots` goldens pass **unblessed**: the palette snapshots carry no recents, and the suite strips style, so neither the new section nor the new ground moves a frame.
- `cargo clippy -p stella-tui --all-targets` — exit 0. `cargo clippy -p stella-cli --all-targets` — exit 0.
- `RUSTDOCFLAGS="-D warnings" cargo doc -p stella-tui --no-deps` — exit 0.
- `make guards-fast` — exit 0, `make prose` included (no baseline edited).
- CI on this branch is the PR's evidence for anything wider; nothing here was built workspace-wide on the maintainer's laptop.

## One structural change

`composer.rs` crossed the 1500-line file-size guard, so its test module moved to `composer/tests.rs` unchanged — the split `render/tests.rs` and `deck_ui/tests/` already follow. No test was deleted or renamed; the file is 921 lines now, well clear of the limit and still not in the baseline.

## Noticed, not fixed — filed

- **#5213** — the rendering puts a relative age (`2m ago`) on each `recent` row and this does not. The age needs a timestamp in the store and a clock in the renderer, and the existing right-aligned `live` slice is keyed by command name, so it would paint the age on the domain-group copy too. `P4 · area:tui · use-model:balanced · pain:user-experience`.
- **#5215** — the rendering `08-command-palette` draws section headings with `/ga` typed; the code drops headings under any query (#4338's deliberate decision, pinned by `a_queried_popup_draws_no_headings`). This PR made the spec prose state the shipped behaviour, which leaves the SVG as the only artifact claiming otherwise. Pre-existing, not introduced here. `P4 · area:tui · area:docs · use-model:cheap · pain:brand-consistency · pain:maintainability`.

## Summary by Sourcery

Complete the command palette experience with fuzzy match highlighting, workspace-persistent recents, and consistent anchored overlay styling.

New Features:
- Add fuzzy command-name matching that highlights all consumed letters, including scattered matches, in the command palette.
- Persist up to five recently executed palette commands per workspace and display them in a trailing `recent` section.

Bug Fixes:
- Give the command and argument popups an opaque panel background so terminal content does not show through the overlays.

Enhancements:
- Anchor the command palette to the composer instead of centering it, and amend the TUI specification and implementation plan to document the shipped matcher, positioning, and recents behavior.

Documentation:
- Update the TUI design documentation to reflect the completed palette behavior and the decision not to add a fuzzy-matching dependency.

Tests:
- Add coverage for scattered matching, highlight positions, recent-command recording and persistence, recent-section rendering, and popup background styling.

Chores:
- Move the composer tests into a dedicated module to remain within the file-size guard.

### DIFF
--- a/crates/stella-cli/src/command_deck.rs
+++ b/crates/stella-cli/src/command_deck.rs
@@ -186,6 +186,18 @@ fn system_notice(text: String) -> Inbound {
     Inbound::Notice(text)
 }
 
+/// Where this workspace keeps the command palette's `recent` section
+/// (SPEC 10). It sits under `.stella/private/` with the rest of the generated
+/// local state, so it is gitignored and never travels: the last five commands
+/// a person ran are theirs, and a clone that inherited someone else's would be
+/// wrong as well as private (AGENTS.md § the `.stella/` directory).
+fn palette_recent_path(workspace_root: &std::path::Path) -> PathBuf {
+    workspace_root
+        .join(".stella")
+        .join("private")
+        .join("palette-recent.json")
+}
+
 /// `STELLA_DEBUG=1` → the structured deck log path (L-T8), mirroring the
 /// location `stella_tui::DeckOptions` documents. `None` otherwise, and
 /// on any failure to create the directory — a lost debug log never gates the
@@ -721,6 +733,7 @@ pub async fn run_deck_session(
         debug_log_path: debug_log_path(),
         slash_commands: deck_slash_commands(&custom),
         initial_graph: agent::graph_snapshot(&cfg.workspace_root),
+        recent_path: Some(palette_recent_path(&cfg.workspace_root)),
         no_anim,
         accessible,
         mid_turn_prompt: steer::mid_turn_prompt_policy(cfg),
--- a/crates/stella-tui/src/composer.rs
+++ b/crates/stella-tui/src/composer.rs
@@ -32,9 +32,12 @@ use stella_protocol::{Attachment, AttachmentKind};
 use unicode_width::UnicodeWidthChar;
 
 pub mod args;
+pub mod fuzzy;
 pub mod palette;
+pub mod recent;
 
 pub use palette::{PaletteState, RelevantNow, SlashDomain};
+pub use recent::Recents;
 
 /// Below this many lines a paste is inserted inline; at or above it, the
 /// paste collapses to a chip. Small on purpose (L-T3).
@@ -198,6 +201,13 @@ pub struct Composer {
     cursor: usize,
     /// Paste-collapse threshold in lines.
     paste_threshold: usize,
+    /// Commands run from this composer's slash popup, most recent first —
+    /// SPEC 10's `recent` section. It rides the composer because the composer
+    /// already owns the slash-menu view ([`Composer::slash_menu`]), so the
+    /// key handler and the renderer read one list without either surface
+    /// having to pass it between them. A surface that hands it no path keeps
+    /// an in-session list; see [`recent::Recents`].
+    recent: Recents,
 }
 
 impl Composer {
@@ -215,6 +225,25 @@ impl Composer {
             paste_threshold: threshold.max(1),
             ..Self::default()
         }
+    }
+
+    /// Keep the `recent` section in `path` from now on, seeded with what is
+    /// already there. Called once by the surface that knows which workspace
+    /// this composer belongs to (`DeckOptions::recent_path`).
+    pub fn keep_recent_in(&mut self, path: impl Into<std::path::PathBuf>) {
+        self.recent = Recents::kept_in(path);
+    }
+
+    /// The commands this composer has run, most recent first.
+    pub fn recent(&self) -> &[String] {
+        self.recent.names()
+    }
+
+    /// Write the `recent` list back if a dispatch has changed it. Cheap
+    /// enough to call on every keystroke — it returns at once when nothing
+    /// has moved.
+    pub fn flush_recent(&mut self) {
+        self.recent.flush();
     }
 
     /// The live buffer text (what is being typed, chips excluded).
@@ -451,7 +480,7 @@ impl Composer {
         if !q.starts_with('/') || q.contains(char::is_whitespace) {
             return None;
         }
-        Some(SlashMenu::filter_with(commands, q, state))
+        Some(SlashMenu::filter_with(commands, q, state, self.recent()))
     }
 }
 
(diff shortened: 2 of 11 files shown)


macanderson:
Correction to the evidence in the PR description: the first push was red on `fmt + clippy + test`, on a duplicate `use super::*;` the test-module split left in `composer/tests.rs`. Fixed in `2b0ab5e5a`, and every check on that head is green — `fmt + clippy + test`, `cargo deny + cargo audit`, `docs guards`, `guard self-tests`, `file size ratchet`, `dod`.

Worth naming why the local run missed it, because the mechanism is the one AGENTS.md § Gotchas warns about. I ran `cargo clippy -p stella-tui --all-targets` and filtered the output with `rg "^warning|^error"`. CI runs it with `-- -D warnings`, so what is a warning locally is an error there — and cargo colourises for a terminal, which puts the escape *before* the word, so `^warning` matched nothing and printed a reassuring blank. Two defects in one filter: the wrong strictness and a filter that could not see the output it was reading. The re-run used `--color=never -- -D warnings` and reads the exit code.

Sourcery's assessment table on the linked issue is three rows, all ✅, with no ❌ to settle.

PR #5223, opened 2026-08-27
feat(plugin-ui): the panel host, and a hello-world plugin that draws through it

### LINKED ISSUE #5055: The deck has no plugin panel host

macanderson:
Closes #5055. (The contract, #5208, has merged; this now targets `main` directly.) **This is the half that puts a plugin's pixels on a screen**, plus the plugin that proves it end to end.

```
  ╭ ◳ panel · hello ─────────────────────────────────────────╮
  │hello from a plugin                                       │
  │                                                          │
  │leased  58×8 cells                                        │
  │meter   ████████████░░░░░░░░░░░░  50%                     │
  │                                                          │
  │this row is drawn by a separate process                   │
  │the border and title above are the host's                 │
  ╰──────────────────────────────────────────────────────────╯
```

That frame is not a fixture. `cargo test -p stella-tui --test hello_panel_demo -- --nocapture` starts `plugins/stella-hello/panel.py` as a real process, hands it a real lease, decodes the wire JSON it actually writes, and blits it through the real host renderer.

## What lands

- **`crates/stella-tui/src/plugin_panel.rs`** — the renderer: `chrome` draws the host's border and `◳ panel · <plugin>` label and returns the lease strictly inside it; `blit` paints lines or patches; `throttle_tag` marks an over-budget panel in host ink.
- **`crates/stella-runtime/src/panel_host.rs`** — the driver: one line-framed JSON exchange with the plugin's process, `env_clear` plus the manifest's allowlist, group-kill reaping, typed refusals. It lives here rather than in the CLI because this crate already owns the wrapper subprocess transports, and a binary-only crate cannot be driven from an integration test.
- **`plugins/stella-hello/`** — the reference plugin: a manifest declaring all three surfaces, and a ~90-line Python panel process. Both are written to be read.
- **`design/tui-v2/SPEC.md` §12** — rewritten from a four-bullet sketch into the protocol that shipped.

## Four escapes closed before merge

A review pass found these; each has a witness that fails without its fix.

1. **A plugin could paint the host's border.** The blit advanced one column per `char` while ratatui advances by display width, so a double-width glyph in the last leased column was drawn across the border — and `BufferDiff` then skipped that column, so nothing repainted it. Both `fits` and the clip called it in-bounds because both counted characters. The blit now measures display cells and refuses a glyph that would cross the edge. Witness: `a_wide_glyph_cannot_reach_past_the_lease_into_the_border`.
2. **A grandchild leaked per tick.** EOF on stdout is not process exit, so disarming the group guard there left a backgrounded helper alive forever. The guard now runs to drop, and the timeout path kills explicitly. Witness: `a_panels_grandchild_does_not_outlive_its_tick`, asserting with `kill(pid, 0)` on **both** paths.
3. **A panel that starts a helper never answered.** Reading to EOF meant the helper's inherited stdout held the pipe open past the timeout — the panel had answered instantly and was throttled anyway. A frame now ends at a newline. Found *by* witness 2 rather than by review.
4. **Piped stderr nobody read** would deadlock the child at one buffer, so it is discarded. Plus `PATH` is passed (else a venv `python3` is unfindable under `env_clear`), and a missing pid refuses rather than arming a guard that guards nothing.

## Not in this PR

Nothing yet **places** a panel on a real screen — the SETTINGS pane, the transcript block and the `/hello` popup are wiring, and they are blocked on nothing here. The demo drives the same code those placements will.

Verified: `stella-plugin` (10 suites), `stella-runtime` panel_host (6), `stella-tui` plugin_panel (9) and the demo (1) all green; `cargo clippy --all-targets` on the three crates exits 0; `make prose` OK.

## Summary by Sourcery

Enable plugins to render isolated, host-framed panels through a real process-backed protocol and provide a readable hello-world implementation that demonstrates the complete path.

New Features:
- Add host-side rendering for plugin panels with branded chrome, leased interiors, token-based styling, clipping, and visible throttling.
- Add a runtime panel driver that exchanges newline-delimited JSON with isolated plugin processes, applies environment allowlists, enforces timeouts, and reaps process groups.
- Add the `stella-hello` reference plugin and an end-to-end demo that renders its real process output through the host.

Bug Fixes:
- Prevent plugin output from overwriting host chrome, including with display-width glyphs, and ensure timed-out or backgrounded plugin helpers are terminated.
- Avoid panel deadlocks and unintended environment exposure by framing responses at newlines, discarding stderr, preserving PATH, and refusing unusable processes.

Enhancements:
- Expand the TUI panel specification to define frame semantics, surfaces, host-owned chrome, consent, environment isolation, and lifecycle guarantees.

Documentation:
- Rewrite the plugin panel protocol specification to document the shipped contract and its security and rendering constraints.

Tests:
- Add runtime, renderer, and end-to-end coverage for panel exchanges, rendering boundaries, environment isolation, timeouts, and process-group cleanup.


macanderson:
Triage note from a CI sweep — diagnosis only, no code touched.

`fmt + clippy + test` is red on one test, and it is an invariant rather than a flake:

```
---- no_ambient_reads stdout ----
`RuntimeSpec` field instead, so two sessions in one process can disagree about it:

crates/stella-runtime/src/panel_host.rs:166: reads ambient process state (`std::env::var`)
    if let Ok(path) = std::env::var("PATH") {
crates/stella-runtime/src/panel_host.rs:173: reads ambient process state (`env::var(`)
    if let Ok(value) = std::env::var(name) {
```

Everything else in the job is green — 2385 passed in the big shard, and the failure is `test result: FAILED. 0 passed; 1 failed` in `stella-runtime`'s `no_ambient_reads` alone.

The reasoning in the comment above those two lines is right about the *hazard*: after `env_clear()`, a child left to `execvp`'s default path will not find a `python3` that lives in a virtualenv or in Homebrew. The objection is only to **where the value comes from**. `stella-runtime` is the crate that has promised (in `tests/no_ambient_reads.rs`) never to read process-global state, so that two sessions in one process can disagree about it — a panel host that reads the host's `PATH` makes every session in the process share one, which is the case the test exists to prevent.

The remedy the guard names is to take it as an explicit field and have the caller — `stella-cli`, a binary, which may read the environment — snapshot `PATH` and the manifest's allowlisted names and hand them in. Which type carries it (`Runtime`, `PanelLease`, or a new parameter on `ask`) is an API call on code you have just written, so I have deliberately not guessed at it.

Not touched, and not a blocker for this note: `dod` and the remaining pending checks.

Issue #5234, opened 2026-08-27
The model row cannot say how many model calls a turn was budgeted, because nothing budgets them


macanderson:
## Problem

SPEC 6.3 (`design/tui-v2/SPEC.md` §6.3) gives the model event a footer of two clauses:

```
irreducible generation · n of m budgeted model calls this turn
```

#5033 shipped the row and the first clause. The second is **elided**, because nothing in this workspace budgets model calls per turn, so there is no `m`:

- `EngineConfig::max_steps` (`crates/stella-core/src/driver/config.rs`) is the nearest number and its own doc calls it a "hard backstop on step count, independent of loop detection — belt and suspenders, never the *primary* stuck-loop defense". Printing `3 of 200 budgeted` would report a backstop as a plan, on the one row whose whole job is pricing model work honestly against SPEC 1's deterministic-first thesis.
- `AgentEvent::BudgetTick` budgets **spend in USD**, not calls.
- `n` alone cannot stand in: `n of m` is one reading, and half of it says a different thing.

So the transcript states that a call was irreducible generation without saying what share of the turn's allowance it spent — which is the comparison the clause exists to make.

## Files

- `crates/stella-core/src/driver/config.rs` — where a per-turn model-call allowance would be configured, if one is wanted. It is not `max_steps`: a step can make several model calls (`call_seq` disambiguates them), and `max_steps` is a backstop rather than a plan.
- `crates/stella-core/src/driver/settlement.rs` — `emit_step_usage`, the per-call metering record. `StepUsage` already carries `turn_instance` and `call_seq`, so `n` is derivable at the fold; only `m` is missing.
- `crates/stella-tui/src/views/transcript_source.rs` — `MODEL_FOOTER` and `model_rows`, where the clause comes back once `m` has a source. The doc comment on `MODEL_FOOTER` states the elision and why.
- `crates/stella-tui/src/model/turn.rs` — `model_call_row`, the fold that would carry the pair onto `TranscriptEntry::Model`.

## The decision this needs first

Whether a per-turn model-call allowance is a thing Stella should have at all. Two answers are both defensible, and the issue is not closed by implementing one without arguing it:

1. **It is.** A turn declares an expected model-call count (from triage, from the plan's step count, or from configuration), the driver carries it, and the footer states `n of m`. This makes overrun visible in the transcript at the moment it happens.
[...]

Issue #5225, opened 2026-08-27
Three files sit at the 1500-line ceiling ungrandfathered, so the next line into any of them fails the gate


macanderson:
## Problem

Three files are at or one line under the 1500-line `file-size` ceiling, none of them grandfathered, so **the next line to land in any of them fails the gate** and the only remedy is a split:

| File | Lines |
|---|---|
| `crates/stella-cli/src/self_driving_cmd/drive.rs` | 1500 / 1500 |
| `crates/stella-tui/src/model.rs` | 1499 / 1500 |
| `crates/stella-cli/src/self_driving_cmd.rs` | 1499 / 1500 |

This already cost real work: #5045's implementation had to be reworked twice to fit inside `model.rs`, and the shape it settled on (hanging a field off `RememberedDiff` rather than adding a struct) was chosen to stay under two added lines rather than because it was the best design.

The `file-size` guard reports crowded files on a passing run (#4897), which is how these were noticed — but reporting does not split them, and the author who eventually meets the ceiling is by construction the one whose change is about something else.

## Files

Each split is its own PR, with its own seam argument. `crates/stella-core/src/driver/settlement.rs` (split out of `driver.rs`) is the pattern AGENTS.md names.

## Done means

Each of the three sits meaningfully below its ceiling, split along a concern rather than at a line count, with the module doc of each new file stating what it owns. No baseline entry is added — these are not grandfathered and must not become so. `make file-size-retighten` is a separate follow-up once the slack exists.

**Estimated agent build time:** ~1 day for the three
**Model:** use-model:pro — Opus 5 (`claude-opus-5`)

## Definition of done

- [x] `crates/stella-cli/src/self_driving_cmd/drive.rs` — 1500 → 1392, split along a concern (`escalate` + `assess_one`, the loop's triage half) into `drive/triage.rs`.
- [x] `crates/stella-cli/src/self_driving_cmd.rs` — 1499 → 1388, split along the file's own `aperture / seen / calibrate / queue` banner into `self_driving_cmd/query.rs`.
- [x] `crates/stella-tui/src/model/tests.rs` — 1497 → 1342, split along the file's own `#4155` banner into `model/tests/edits.rs`. 44 tests before, 44 after.
- [x] Each new file's module doc states what it owns.
- [x] No baseline entry added; none of the three is grandfathered.

[...]

Issue #4183, opened 2026-08-21
tui-v2: SPEC 6.1's opening rule cannot name the model on the first turn — hud.model is only written when the turn ends


macanderson:
## Problem

SPEC 6.1's opening rule is `── turn N <stage> · <model> · budget $X ──`, and
the `<model>` half is blank for the whole first turn of every session.

`TranscriptEntry::Stage`'s `TurnOpening::model` is stamped from
`SessionModel::hud.model` (`crates/stella-tui/src/model.rs`, the
`AgentEvent::Stage` arm of `apply`). `hud.model` is written in exactly two
places, both terminal:

- `AgentEvent::TurnComplete` — after the turn it would have labelled
- `AgentEvent::RunComplete` — after the run

So the opening rule of turn 1 renders `── turn 1 execute ──` with no model, and
turn 2 onward renders the model *the previous turn* used. #4124 elided rather
than substituted, because naming the configured default over a routed call
would be the rule asserting a routing decision nothing recorded — but eliding
is a placeholder, not the answer.

## What the wire already carries

Both of these arrive **during** the turn and name the model:

- `AgentEvent::StepUsage { role, model, .. }` — one per committed model call
- `AgentEvent::StepManifest { turn_instance, role, model, .. }` — emitted
  immediately before each call commits

`crates/stella-tui/src/model.rs` currently folds neither (they fall into the
no-op arm around the `StepUsage`/`StepManifest` group), deliberately: folding
their token and cost fields would double-count the spend the budget gauge
already tracks.

## What is wanted

Fold the **model name only** — not the usage numbers — from whichever of those
two events is the right source, so `hud.model` is populated from the turn's
first committed call instead of from its end.

The trap to design around: both events carry auxiliary calls as well as the
worker's. A `Summarization` or `Verdict` call's model is not the model
answering the turn, so an unfiltered fold would make the rule name the overflow
summarizer. Filtering on `ModelCallRole` is the obvious approach, but note the
role vocabulary is itself in flux — see the roleless-core work (#3903 and the
open "roleless-core slice 2" issue) — so whatever predicate is chosen should
live in one named place the fold cites.

## Files

- `crates/stella-tui/src/model.rs` — `SessionModel::apply`, the
  `StepUsage`/`StepManifest` arm and the `Hud::model` writes
- `crates/stella-tui/src/model/turn.rs` — `Hud::model` and
  `TurnOpening::model`, whose docs both state the current gap and must be
  updated with the fix
[...]

Issue #4150, opened 2026-08-21
tui-v2: every tool head row states a metric it never measured — edit renders `+0 -0` unconditionally, reintroducing #2290's shape


macanderson:
## Symptom

Every tool head row the v2 transcript draws states a measurement it never took. In a live session, four consecutive `edit_file` calls that each reported `replaced 1 occurrence(s)` all rendered:

```
● edit crates/stella-tools/src/search/engine.rs +0 -0
```

The `+0 -0` is not a measurement that came back zero. It is a literal `0` compiled into the row.

## Cause

`crates/stella-tui/src/v2/transcript_source.rs::kind_for` hardcodes the metric for all four file tools:

```rust
fn kind_for(name: &str) -> EventKind {
    match name {
        "read_file" => EventKind::Read { lines: 0 },
        "edit_file" => EventKind::Edit { added: 0, removed: 0 },
        "write_file" => EventKind::Write { lines: 0 },
        "delete_file" => EventKind::Delete { lines: 0 },
        ...
```

`crates/stella-tui/src/v2/transcript.rs::kind_detail` then renders those fields as if they were real:

- `Edit` → `+{added} -{removed}` (green/red)
- `Read` → ` · {lines} lines`
- `Write` → ` · new file · {lines} lines`
- `Delete` → ` · -{lines} lines · git-backed · u undo`

Nothing in production ever constructs `EventKind::Edit` with a non-zero value — `rg -n "EventKind::Edit" crates/` returns exactly one production site (`transcript_source.rs:78`) and four test sites. The same holds for `Read`/`Write`/`Delete`.

It is structural, not a missed assignment. `v2_rows` (`crates/stella-tui/src/render/entry.rs:198`) routes `TranscriptEntry::ToolStart` into `v2::head_rows`, and the head is drawn at dispatch — the source's own comment says so:

```rust
// The head is drawn the moment the call dispatches, so it is never
// "collapsed" in the fold sense — there is no body under it yet.
```

At `ToolStart` the tool has not run, so no delta can exist yet, and there is no later pass that revisits the row.

This is ungated: `entry_lines` calls `v2_rows` unconditionally, so every deck session is affected, not a preview path.

## Why this matters beyond cosmetics

`stella-protocol`'s `AgentEvent::FileChange` doc (`crates/stella-protocol/src/event.rs`, around line 860) already argues this exact case, and names the prior instance:

[...]

PR #5236, opened 2026-08-27
feat(stella-tui): the transcript shows one model row per settled call

### LINKED ISSUE #5033: Show a model event with tokens per second in the transcript
## Problem

SPEC 6.3's model event — `◐ model <activity> · tok/s` with footer `irreducible generation · n of m budgeted model calls this turn` — renders from `EventKind::Model{tokens_per_sec}` but has no producer. The footer text exists nowhere. The event is what prices model work against the deterministic-first thesis (SPEC 1), so its absence hides the det/model split at the moment it happens.

## Files

- The driver/usage path in `crates/stella-core` (step usage already carries tokens and timing) — emit the event per model call
- `crates/stella-tui/src/views/transcript_source.rs` + `model/entry.rs` — variant from the spine issue

## Definition of done

A live turn shows one `◐ model` head per model call with a real tok/s figure and the budgeted-calls footer, proven by a live-adapter test.

Converted from a `## Done means` paragraph to the heading the SCR-003 gate reads, per #5193. The prose above is unchanged; the checklist below is that sentence split into the items a closer can verify line by line. The two items marked *split* are the remainder SCR-003 step 2 asks to be filed as its own issue rather than claimed — see the closing comment on this issue for what was found and why.

- [x] Implementation complete: `TranscriptEntry::Model`, its fold arm, its projection, and a live producer — PR #5236
- [x] A live turn shows one `◐ model` head per model call, from the driver's own `AgentEvent::StepUsage` metering record
- [x] The head states a real tok/s figure, derived from that record's output tokens and its own wall clock — never invented, and absent rather than `0 tok/s` when the division has no inputs
- [x] The footer states `irreducible generation`
- [x] The budgeted-calls half of the footer is **split to #5234**: nothing in this workspace budgets model calls per turn, `EngineConfig::max_steps` is a documented backstop rather than a plan, and a fabricated `m` on the row that prices model work would be worse than the missing clause
- [x] End-to-end coverage of the wire→fold→render chain is **split to #5239**: `stella-tui` deliberately does not depend on `stella-core`, so a true live-adapter test cannot live in the crate under change; the chain is proven in two halves meeting at a compiler-checked join
- [x] Scoped tests added/updated and passing: four witnesses in `crates/stella-tui/src/views/transcript_source.rs`, flip verified by hand (revert the emit → all four fail; restore → all four pass)
- [x] Full CI green — run 33035962545, `completed / success`, 55/55 steps: `cargo fmt --check`, `clippy -D warnings`, `cargo doc -D warnings`, the wire-schema check, the file-size ratchet, the workspace test suite, `stella context validate` and the release smoke build
- [x] Residue filed as new issues (triage label only): #5234, #5239

Spec: `design/tui-v2/SPEC.md` §6.3.

**Estimated agent build time:** ~3h

macanderson:
Closes #5033

## What was missing

SPEC 6.3 (`design/tui-v2/SPEC.md` §6.3) defines the model event as `◐ model <activity> · tok/s` with a footer. `EventKind::Model` rendered the head and was tested — from fixtures. **Nothing produced one.** A turn's generation, the work SPEC 1's deterministic-first thesis is measured against, left no mark on the transcript at the moment it happened, so the det/model split was invisible exactly where it is decided.

## What this does

Adds `TranscriptEntry::Model`, its fold arm, its projection, and its live producer, in one PR — per the scope note on the issue.

**The rate is derived, not invented, and nothing new goes on the wire.** `AgentEvent::StepUsage` is already the driver's per-call metering record (`stella-core`'s `driver::settlement::emit_step_usage` for the worker, `accounted_call` for the management roles) and it already carries the output tokens the call produced and the wall clock it took. tok/s is a division over those two. So the "upstream emit" the issue asks for **already exists on `main`** — what was missing was everything between the wire and the row, and adding a second event for a fact the wire already states would have been a copy to drift and a new `consumers.rs` row for no gain. **No `AgentEvent` wire tag was added or changed, so the event-consumer ledger and `docs/wire/` are untouched.** `StepUsage`'s ledger row is already `Behavioral`, and the deck is deliberately not a `Surface` (interactive mode matches `AgentEvent` exhaustively by construction — see `Surface`'s own doc).

The chain: `StepUsage` → `model::turn::model_call_row` → `TranscriptEntry::Model` → `render::entry`'s projection router → `views::transcript_source::model_rows` → `EventKind::Model`.

## Three things stated as absent rather than as zero

1. **The rate.** `EventKind::Model::tokens_per_sec` becomes `Option<u32>`. A provider that vouched for no usage envelope (`StepUsage::complete` false) reports zeros that are not measurements; an untimed call has nothing to divide by; a call that generated nothing has no rate. All three render no column — `0 tok/s` asserts a model that generated nothing per second, the same substitution `+0 -0` made over a real edit (#4150, #2290).
2. **The activity.** `ModelCallRole::Unknown` (the serde default for an absent role, so every pre-attribution stream) names no activity rather than printing `unknown` as though the engine had a stage by that name — the refusal #4124 already made for the turn rule's model name.
3. **The footer's second clause.** SPEC 6.3 asks for `irreducible generation · n of m budgeted model calls this turn`. The first half ships: it is a fact about the call, not a measurement, and it is what the gate row's `$0.00 · det` says in the other direction. **The `n of m` half is elided and I am saying so out loud rather than fabricating it.** Nothing in this workspace budgets model calls per turn. `EngineConfig::max_steps` is the nearest number and its own doc calls it a "hard backstop on step count … never the *primary* stuck-loop defense", so `3 of 200 budgeted` would report a backstop as a plan on the one row whose job is pricing model work honestly. `BudgetTick` budgets dollars, not calls. `n` alone cannot stand in — `n of m` is one reading. **#5234** is filed with the decision that has to come first (add a real per-turn allowance, or drop the clause from the spec), and `MODEL_FOOTER`'s doc comment states the elision and cites it.

## Witness, and the flip verified artisanally

Four tests in `crates/stella-tui/src/views/transcript_source.rs`, each driving a real `AgentEvent::StepUsage` through `SessionModel::apply` and out of `render::entry_lines` — the production renderer, not `model_rows` directly, because the half that was missing was never the projection:

- `a_settled_model_call_renders_its_rate_from_the_drivers_own_metering` — 420 output tokens over 5s renders `◐ model worker · 84 tok/s` with `⚡5000ms`. The arithmetic is asserted, not the field.
- `the_model_footer_claims_no_budget_nobody_set` — the footer says `irreducible generation` and does not say `budgeted`.
- `a_call_whose_rate_has_no_source_states_no_rate` — the three absences above, each still drawing the row.
- `a_call_with_no_recorded_role_names_no_activity`.

**Flip, checked by hand:** with the fold's one-line `self.transcript.extend(turn::model_call_row(event));` removed, all four fail (`one metering record must fold to exactly one model row`); restored, all four pass. Full output of both runs was read, not inferred. The renderer half needs no manual flip — `TranscriptEntry`'s matches are exhaustive, so a missing arm is an `E0004`.

## Goldens

Re-blessed with `BLESS=1 cargo test -p stella-tui --test deck_render_snapshots` and **the diff was read line by line**. Every changed file is symmetric `+N/-N`: the only new content is the two `◐ model worker` rows plus their footers, and the equal number of removed lines at the top of each is the transcript pane scrolling by exactly the rows added. The overlay cards themselves are byte-identical. `tab_session.txt` and `session_delegate_tool_call.txt` are the clearest reads.

## Checks run

- `cargo test -p stella-tui` — 1231 + 10 target suites, all green
- `cargo clippy -p stella-tui -p stella-core --all-targets` — exit 0, no warnings
- `RUSTDOCFLAGS="-D warnings" cargo doc -p stella-tui --no-deps --document-private-items` — exit 0
- `make prose` — OK, none added, no baseline touched
- `make guards-fast` — green, `check-file-size` / `check-god-files` / `check-left-behind` / `check-line-citations` included
- `cargo fmt`

No workspace build or `make gate` was run locally; CI is the evidence for those.

## Noticed and not fixed

- **`crates/stella-tui/src/model.rs` goes from 1497 to 1499 of its ungrandfathered 1500-line ceiling.** This spends two of the three lines it had left. I took the two lines rather than contorting the fold to avoid them, because #5225 names that contortion as a cost it already paid once ("the shape it settled on … was chosen to stay under two added lines rather than because it was the best design"). #5225 is the split; this PR does not do it, and the next change into that file will have to.
- The stale comment above `model.rs`'s `StepUsage` arm described the arm as folding a model name it has not folded since #4183 moved that to `StepManifest`. Rewritten in place, since I was editing it.
- `render::entry::projected_rows`' doc said "two arms, deliberately" and had four. Corrected.

Exemplar for the `Option`-over-zero discipline: this crate's own `Extent` and `ReadSize`, whose docs already argue the case.

## Summary by Sourcery

Render one transcript model row for every settled model call using the driver's existing metering data.

New Features:
- Add transcript model-call rows for each settled StepUsage event, including the call activity, duration, optional output rate, delegate, and generation footer.
- Support searching transcript model rows by their recorded activity.

Bug Fixes:
- Ensure model calls are no longer absent from live transcript rendering despite the existing model event presentation.

Enhancements:
- Represent unavailable model rates and unidentified activities explicitly instead of displaying misleading zero or unknown values.
- Route model entries through the shared transcript projection and preserve existing exhaustive rendering behavior.

Documentation:
- Clarify issue-label creation and triage responsibilities in the contributor guidance.

Tests:
- Add end-to-end transcript rendering coverage for settled model calls, rate derivation, unavailable rates, missing activities, and footer content.
- Refresh deck rendering snapshots to include the new transcript rows.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -191,7 +191,13 @@
   Search for an existing issue first; link, don't duplicate. The full policy
   is AGENTS.md § "Nothing left behind — every finding becomes a fix or a
   GitHub issue". Never end a turn with untracked half-finished work.
-- **Every issue is labelled, in this order.** First a priority (`P0`–`P4`),
+- **An issue you file carries the `triage` label and nothing else.** SCR-004
+  and SCR-005 (AGENTS.md's standing-decisions block) hold sizing and priority
+  for a dedicated triage agent, and `triage-guard` strips a creator-applied
+  priority — so a `P2` you add is both a rule break and a label that does not
+  survive. The taxonomy below is what **triage** applies; read it to
+  understand what the labels on an issue mean, not as a list to add yourself.
+- **The taxonomy, in this order.** First a priority (`P0`–`P4`),
   then the `area:*` tag — several when the work genuinely spans areas — then
   exactly one `use-model:*` tag (`cheap` / `balanced` / `pro` / `ultra`)
   naming the cheapest model class that can do the work without sacrificing
--- a/crates/stella-tui/src/model.rs
+++ b/crates/stella-tui/src/model.rs
@@ -966,26 +966,28 @@ impl SessionModel {
                 };
                 self.transcript.push(entry);
             }
-            // Two facts come off a metering record, and they have different
-            // scopes — which is why this is one arm and not two.
+            // Two facts come off a metering record, and one arm folds both.
             //
-            // The **tokens** are folded for every call, whatever its role: the
+            // The **tokens** are counted for every call whatever its role: the
             // receipt accounts for the whole turn, auxiliary work included
-            // (#4184). The **model name** is folded only for a call that
-            // answers the turn (#4183). Splitting them into a guarded arm and a
-            // fallthrough would silently stop counting an answering call's
-            // tokens, because the first matching arm wins.
+            // (#4184). Guarding this arm on a role would stop counting an
+            // answering call's tokens — the first matching arm wins.
             //
-            // The cost deliberately is not folded either way. The HUD's live
-            // spend comes from `BudgetTick`, so folding `StepUsage::cost_usd`
-            // here would double-count it — that hazard is why this event was
-            // ignored outright, and the token half was collateral.
+            // The **row** is SPEC 6.3's `◐ model <activity> · tok/s`, one per
+            // settled call — see `turn::model_call_row` for why the rate is a
+            // division over this event rather than a second wire signal.
+            //
+            // The cost is folded neither way. The HUD's live spend comes from
+            // `BudgetTick`, so folding `StepUsage::cost_usd` here would
+            // double-count it — that hazard is why this event was ignored
+            // outright, and the token half was collateral.
             AgentEvent::StepUsage {
                 input_tokens,
                 output_tokens,
                 ..
             } => {
                 self.turn_counters.add_tokens(*input_tokens, *output_tokens);
+                self.transcript.extend(turn::model_call_row(event));
             }
             AgentEvent::UsageIncomplete { .. }
             // Context receipts (spec §4/§5) are consumed by the store/inspector,
--- a/crates/stella-tui/src/model/entry.rs
+++ b/crates/stella-tui/src/model/entry.rs
@@ -181,6 +181,33 @@ pub enum TranscriptEntry {
         /// that never saw the start still needs (#4699).
         sub_agent_id: Option<String>,
     },
+    /// One settled model call — SPEC 6.3's `◐ model <activity> · tok/s`.
+    ///
+    /// Folded from [`stella_protocol::AgentEvent::StepUsage`], which the driver
+    /// already emits once per call with the call's own output tokens and its
+    /// own wall clock (`turn::model_call_row`). The rate is a division
+    /// over two numbers `stella-core` measured, so nothing new rides the wire
+    /// to carry it.
+    Model {
+        /// What the call was for, as the wire spells the role — `worker`,
+        /// `plan_repair`, `summarization`.
+        ///
+        /// `None` is [`stella_protocol::ModelCallRole::Unknown`]: a stream
+        /// recorded before call-role attribution existed. The head then names
+        /// no activity at all rather than printing `unknown`, on the rule
+        /// `turn::supplies_the_turns_model` already applies to that
+        /// role — a call this build cannot identify states no fact about
+        /// itself.
+        activity: Option<String>,
+        /// Output tokens per second over the call's own wall clock, or `None`
+        /// when the division has no inputs to divide (see
+        /// `turn::model_call_row` for the three cases).
+        tokens_per_sec: Option<u32>,
+        /// The call's own wall clock, rendered `⚡Nms` like every other head.
+        duration_ms: u64,
+        /// Which delegate made the call, `None` for the lead's own (#4699).
+        sub_agent_id: Option<String>,
+    },
     /// A model call was retried (surfaced only once the step commits — the
     /// engine defers these, L-E10).
     Retry { attempt: u32, reason: String },
--- a/crates/stella-tui/src/model/turn.rs
+++ b/crates/stella-tui/src/model/turn.rs
@@ -155,6 +155,15 @@ fn activity_of(role: ModelCallRole) -> Option<String> {
 /// never `0 tok/s` — the substitution [`super::ReadSize`] and
 /// [`crate::views::transcript::Extent`] both already refuse, because a zero
 /// asserts a measurement where there was none (#2290, #4150).
+///
+/// Not a second copy of `deck::AgentEntry::live_tok_per_s`, which shares the
+/// arithmetic and neither numerator nor denominator. That one is the **live**
+/// rate of a turn still running: the turn's output so far over the deck's own
+/// clock since the turn opened, tool time included, and it needs no settled
+/// record because there is none yet. This one is one **call**, settled, over
+/// the wall clock that call alone took. A turn's rate and its slowest call's
+/// rate are different readings, and either standing in for the other would
+/// answer a question nobody asked.
 fn tokens_per_sec(output_tokens: u64, duration_ms: u64, complete: bool) -> Option<u32> {
     if !complete || output_tokens == 0 || duration_ms == 0 {
         return None;
--- a/crates/stella-tui/src/model/turn.rs
+++ b/crates/stella-tui/src/model/turn.rs
@@ -13,7 +13,9 @@
 //! the same cut `file_state` and `recall` already make, and along the same
 //! seam: a `TranscriptEntry`'s supporting types live beside it, not in it.
 
-use stella_protocol::{BudgetMode, ModelCallRole, StageKind, StageName};
+use stella_protocol::{AgentEvent, BudgetMode, ModelCallRole, StageKind, StageName};
+
+use super::TranscriptEntry;
 
 // Doc-link target only: named in `TurnOpening::queued_steer`'s docs, gated by
 // the fold in `super`. `cfg(doc)` keeps the intra-doc link resolving without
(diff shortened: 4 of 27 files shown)


macanderson:
## Answering the two ❌ rows in Sourcery's assessment

Both are correct readings of the diff. Neither is a fix this PR should make, and here is why for each.

### ❌ "Render the SPEC 6.3 footer containing `irreducible generation · n of m budgeted model calls this turn`"

**Deliberate, and stated in the PR description and in `MODEL_FOOTER`'s doc comment.** The clause is elided because there is no `m` to state:

- `EngineConfig::max_steps` (`crates/stella-core/src/driver/config.rs`) is the nearest number in the tree, and its own doc calls it a "hard backstop on step count, independent of loop detection — belt and suspenders, never the *primary* stuck-loop defense". Rendering `3 of 200 budgeted` reports a backstop as a plan.
- `AgentEvent::BudgetTick` budgets spend in USD, not calls.
- `n` alone cannot stand in: `n of m` is one reading and half of it claims something else — the same rule this crate applies to an edit's `+a -b`.

Fabricating `m` on the one row whose job is pricing model work honestly against SPEC 1's deterministic-first thesis would be worse than the missing clause. **#5234** carries the decision that has to come first — add a real per-turn model-call allowance, or drop the clause from `design/tui-v2/SPEC.md` §6.3 — and `MODEL_FOOTER`'s doc cites it, so the gap is declared rather than silent.

### ❌ "Prove the live model-event behavior with a live-adapter test"

**The crate boundary forbids one here, and the evidence is split across it by design.**

`stella-tui` does not depend on `stella-core` and must not: the deck is a pure fold over `AgentEvent`, which is what makes replay-from-seq-1 a supported debug mode (`crates/stella-tui/src/model.rs`'s module doc, L-T1). Its `Cargo.toml` lists `stella-protocol`, `stella-tools`, `stella-diag`, `stella-diff`, `stella-transcript` and `stella-tui-theme` — nothing engine-shaped. Adding that edge to write a test would cost the property the crate exists to have.

So the chain is proven in two halves that meet at a compiler-checked join:

- **Producer.** `crates/stella-core/src/driver/tests/usage_completeness.rs` proves the driver emits a truthful `StepUsage` per model call against a scripted provider — `two_calls_at_one_step_are_separable_by_their_usage_rows` and `step_usage_carries_the_requests_effort_and_output_ceiling`. That producer is on `main` already; this PR adds no emit, which is the point of deriving tok/s from a record the wire already carries.
- **Consumer.** This PR's four tests fold a real `AgentEvent::StepUsage` through `SessionModel::apply` and out of `render::entry_lines` — the production renderer, not the projection directly.
- **The join** is the `AgentEvent::StepUsage` variant itself. The fixture names every field, so a field added or retyped breaks both sides.

**What that pair genuinely cannot catch, and which I am not claiming it does:** an ordering or completeness defect — a producer that stops emitting the event on some path, emits it twice, or emits it where the fold does not expect it. Nothing anywhere in this repository drives a real engine turn through the deck fold, so that gap predates this PR and is wider than it. **#5239** is filed for it, as an integration suite in `stella-cli` (the one crate linking both) that pins the *sequence* rather than the rendering.

macanderson:
## SCR-003 (`dod / dod`) — what I did, and one thing that needs a human

The `dod` check failed with *"macanderson/stella#5033 has no Definition of done section"*. That is #5193's systemic problem, not this change: the DoD gate landed in #5142/#5180 **after** the sibling spine PR #5084 merged, and 56 open issues are still filed with the older `## Done means` prose heading.

Following #5193's stated remedy, I converted #5033's `## Done means` paragraph into a checklist under `## Definition of done`, prose preserved as the lead-in. [Full explanation on the issue.](https://github.com/macanderson/stella/issues/5033#issuecomment-5433888478)

**Two DoD items are ticked as *split*, not as *done*, and that needs review** — an author ticking boxes on the issue their own PR closes is the move that turns a DoD into a guess, so both are named on their face:

- `n of m budgeted model calls this turn` → **#5234**. No `m` exists to state.
- "live-adapter test" → **#5239**. `stella-tui` deliberately does not depend on `stella-core`, so such a test cannot live in the crate under change.

If either descope is not acceptable, say so and I will unpick it rather than merge past it. `Full CI green` is deliberately left **unticked**, so the gate stays red until CI is genuinely green.

## Also corrected: SCR-005 on my own residue issues

I filed #5234 and #5239 with `P2`/`P3`, `area:*`, `use-model:*` and `pain:*` labels, following CLAUDE.md's label paragraph. That is wrong here — SCR-004/SCR-005 (AGENTS.md's standing-decisions block) say residue carries the **`triage` label only**, because a dedicated triage agent owns sizing and priority. Both issues are now `triage`-only, and both bodies were rewritten into the task template's `## Definition of done` shape so they do not reproduce #5193's problem on the next PR that closes them.

Worth noting for the maintainer: **CLAUDE.md's "Every issue is labelled, in this order" paragraph and SCR-005 contradict each other**, and CLAUDE.md is the one an agent reads first. I have not edited CLAUDE.md — that is a standing-decision change, not mine to make — but the conflict is real and will keep producing mislabelled issues.

macanderson:
## One commit here is separable — `1cd4b99` on CLAUDE.md

CLAUDE.md's "Every issue is labelled, in this order" paragraph and SCR-004/SCR-005 flatly contradict each other: the first tells a session to add `P0`–`P4`, `area:*`, `use-model:*` and `pain:*` to any issue it files; the second says residue carries **`triage` only**, because a dedicated triage agent owns sizing and priority and `triage-guard` strips a creator-applied one.

CLAUDE.md is the file a session reads first, so the contradiction resolves in favour of the wrong rule every time — it did in this session, which filed #5234 and #5239 with a priority and five other labels before the DoD gate surfaced the SCR corpus. Both are now `triage`-only.

CLAUDE.md's own rule says *"When this file and AGENTS.md disagree, AGENTS.md wins — and the disagreement is itself a bug: fix it in the same PR you noticed it"*, which is why the fix is here rather than filed. But it is a governance document and this is a transcript-row PR, so it is **its own commit and reverts cleanly on its own** — drop it if you would rather it landed separately, and I will file it instead. The taxonomy text is kept verbatim, relabelled as what triage applies rather than as a list to add yourself.

macanderson:
## CI status at hand-off

14 of 16 checks pass. One is still running, and one fails on a box that is deliberately unticked:

| Check | State |
|---|---|
| `fmt + clippy + test` | **running** — inside it, `cargo fmt --check`, `cargo clippy -D warnings`, `cargo doc -D warnings`, `docs/wire/ still matches the types`, `no new .rs file over the 1500-line ratchet` and every toolchain-free guard have **already passed**; the workspace test step and the release smoke build are what remain |
| `dod / dod` | **fail**, on exactly one item: `- [ ] Full CI green` |
| everything else (14) | pass |

`docs/wire/ still matches the types` passing is the CI-side confirmation of this PR's central claim: **no wire change was needed**, because `AgentEvent::StepUsage` already carries the output tokens and the wall clock that tok/s is derived from.

The `dod` failure is the gate working as designed. I converted #5033's DoD (per #5193) and left `Full CI green` unticked because it is not yet true.

**Whoever picks this up:** once `fmt + clippy + test` goes green, verify it and tick that box on #5033 — `dod` clears on its next run and the PR becomes mergeable. I have not ticked it in advance, and I have not merged.

Issue #5242, opened 2026-08-27
A subsession lane test passes alone and fails in the full stella-cli run


macanderson:
Noticed while running `cargo test -p stella-cli` for #5031 (PR touches nothing under `subsession/` or `durability/`).

## Symptom

`crates/stella-cli/src/subsession/tests.rs::a_resumed_session_never_re_mints_a_lane_id_that_already_has_a_transcript` fails in the full suite:

```
assertion `left == right` failed: a resumed session must step over the lane ids it already recorded
  left: "req:1"
 right: "req:2"
```

The same test passes on its own:

```
cargo test -p stella-cli a_resumed_session_never_re_mints    # ok, 1 passed
cargo test -p stella-cli                                     # FAILED, 1 of 2386
```

So the assertion is not wrong about the code — it is not seeing the state the fixture set up.

## Why it is probably isolation, not the feature

The test writes a checkpoint for a dead lane and then expects `SubSessions::resuming` to notice it and step past `req:1`. Two things make that discoverable-or-not depending on what else ran:

- the session id is the fixed literal `"ses-resumed"`, so any other test in the binary that binds the same id shares its journal key;
- `crate::durability::bind_session` binds a `SessionDurability` that the tests construct with `SessionDurability::default()`, and the recorded-lane lookup the assertion depends on has to find the dead lane through whatever that binding resolves to. A tempdir workspace does not isolate a process-global binding or a real `~/.stella` path.

This is the known shape in this crate: `stella-cli` tests can reach the real `~/.stella` unless every path resolver is redirected, and `STELLA_DATA_DIR` alone is not enough (`STELLA_HOME` is).

## Verify

1. `cargo test -p stella-cli` — observe the failure.
2. `cargo test -p stella-cli a_resumed_session_never_re_mints` — observe it pass.
3. Bisect the interference with `cargo test -p stella-cli -- --test-threads=1` and with `--skip` over neighbouring subsession tests to find which test it shares state with.

## Done means

The test passes in the full suite and alone, without weakening the assertion — the #3233 witness it carries must still fail if `SubSessions::resuming` stops stepping over recorded lanes. Most likely fixes: give the fixture a unique session id, and redirect the durability binding at a root the test owns (`STELLA_HOME`, not just `STELLA_DATA_DIR`).

Refs #3233.

## Definition of done

The “Done means” prose above, as boxes the gate can read — nothing here is
new work, and nothing is ticked by the migration (#5193).

[...]

Issue #5232, opened 2026-08-27
An explicitly invoked skill (/slug) leaves no transcript row


macanderson:
PR for #5031 wires SPEC 6.3's skill event for the **auto-selection** channel: the turn's steering block picks skills (`stella_core::skills::select_skills_reporting`), `crates/stella-cli/src/memory/recall.rs::recall_block_reported` records which of them the section's token budget actually admitted, and `AgentEvent::SkillInjected` carries each to the deck as a `✦ skill <name> · auto · n tok` row.

There is a second channel, and it announces nothing.

## The gap

`crates/stella-cli/src/extensions.rs::CustomExtensions::expand` turns a typed `/slug` into prompt text via `skill_invocation_prompt`, which inlines the skill's name, description and body straight into the user message. No event is emitted, so:

- the deck shows the expanded prompt as if the user had typed the whole skill body;
- nothing records that a skill fired, so `skill_usage` (written at the shared seam from `SessionMemory::selected_skills`, which re-runs *auto* selection) misses it entirely — an explicitly invoked skill is invisible to appraisal, which can then retire it for non-use;
- SPEC 6.3's head reads `auto|/cmd`, and today only `auto` is reachable.

## Files

- `crates/stella-cli/src/extensions.rs` — `expand`, `Invocation::Skill`, `skill_invocation_prompt`. Pure, with no event sender, so the trigger has to be returned to the caller.
- The callers that expand a `/slug`: the REPL loop in `crates/stella-cli/src/agent.rs` and the deck's input path in `crates/stella-cli/src/command_deck.rs`.
- `crates/stella-protocol/src/event/kind.rs` — `AgentEvent::SkillInjected { name, summary, tokens }` needs a trigger discriminant added, and `docs/wire/` regenerated with `make wire-schema-update`.
- `crates/stella-tui/src/views/transcript_source.rs::skill_rows` — the literal `"auto"` there becomes the carried trigger, and its doc comment (which names this issue) comes out.
- `crates/stella-tui/src/model/entry.rs` — `TranscriptEntry::Skill` gains the same field.

## Done means

A session where the user types `/some-skill` shows `✦ skill some-skill · /cmd · n tok` in the transcript, and `skill_rows` no longer hard-codes a trigger. Witnessed by a test that expands a `/slug` and asserts the emitted event carries the command trigger — it must fail before the change, since no event exists on that path at all.

## Constraints

[...]

Issue #5229, opened 2026-08-27
Show injected skills in the Observatory, so skill_injected has a consumer beyond the deck


macanderson:
PR for #5031 adds `AgentEvent::SkillInjected` — one event per skill the steering section actually carried, with its slug, its one-line summary, and what its rendered block cost in tokens.

Its only consumer today is the deck transcript: `crates/stella-tui/src/model.rs` folds it to `TranscriptEntry::Skill` and `crates/stella-tui/src/views/transcript_source.rs::skill_rows` renders SPEC 6.3's `✦ skill <name> · auto · n tok` head. `Surface` (`crates/stella-protocol/src/event/consumers.rs`) is a closed set of two — `Observatory` and `Serve` — and deliberately excludes interactive mode and replay, which render every case. So the row in `crates/stella-protocol/src/event/tags.rs` is `ConsumerPosture::RecordedOnly` citing this issue, which is the ledger's own prescribed answer for a signal only interactive mode renders.

## The gap

Nothing outside the live deck can answer "which skills fired on this turn, and what did they cost?". The store already holds a `skill_usage` row per execution (written at the shared seam, `crates/stella-cli/src/agent/skill_usage.rs`), but that table records *selection* keyed to an execution — not the per-skill token cost the prompt actually paid, and not in the journal a session replay reads.

## Done means

Either:

1. The Observatory's journal query names `skill_injected` in its `event_type` subset (`crates/stella-observatory/src/db.rs`) and a view renders the per-turn skill list with its token cost — at which point the row in `tags.rs` becomes `ConsumerPosture::Surfaced` with `&[Surface::Observatory]` and this issue closes; or
2. A reviewer decides the deck is the right and only home for it, and the row is annotated as intended rather than as a gap — which needs the `RecordedOnly` doc comment in `consumers.rs` to grow that third case, because today `RecordedOnly` means "a consumer is owed".

Whichever is chosen, `crates/stella-protocol/src/event/consumers/tests.rs` enforces posture coherence, so the change is witnessed by that suite.

## Constraints

- Adding a tag to the Observatory query is not free: `scripts/check-consumer-sites.sh` holds `Behavioral` rows' `site` strings to live code, and the ledger tests reject a `Surfaced` row that lists no surface (`LedgerViolation::SurfacedNowhere`).
- `make wire-schema-update` must be re-run and `docs/wire/` committed if the event's shape changes at all.

Refs #5031, #4501 (the sibling `context_write` gap, same shape).

[...]

Issue #4500, opened 2026-08-23
Two commands never show the 'steering withheld' notice, and interactive mode might repeat it every turn


macanderson:
## Problem

`AgentEvent::SteeringWithheld` (#3616, PR #4462) is emitted by `agent::output::open_raw_turn` (`crates/stella-cli/src/agent.rs:1189`) and, since PR #4491 (#4463), once at startup in interactive mode (`command_deck/steering.rs`). Two gaps remain:

1. `stella goal` and `stella fleet` route through neither, so in an untrusted checkout with `.stella/memories/*.md` present those doors print nothing on any channel. Fleet is the hard one: a fan-out's lanes have no shared opener, so where a session-level fact lands is a decision.
2. `open_raw_turn` sits inside a per-turn function and `cfg.authority.withheld` is never cleared, so a long interactive-mode session likely repeats the notice every turn. This is inferred from the call site, not observed — verify with a live interactive-mode session first.

## Done means

Each door emits the notice once per session (or declares a per-door gap in the parity matrix, `crates/stella-parity`), and interactive mode emits it once. Witnesses: the existing stream-json test shape in `crates/stella-cli/tests/settings_warnings_cli.rs`, one per door.

Related: #3616, #4462, #4463, #4491.

PR #5247, opened 2026-08-27
feat(stella-tui): a skill that enters the prompt gets its own transcript row

### LINKED ISSUE #5031: Show a skill event in the transcript when a skill is injected

macanderson:
Closes #5031

SPEC 6.3 (`design/tui-v2/SPEC.md`) defines `✦ skill <name> · auto · n tok` over `injected <summary>`, and `EventKind::Skill` has rendered it since the vocabulary landed. Nothing built one: there was no `TranscriptEntry::Skill`, so the only trace of an injected skill on any live surface was `SteeringWithheld`'s count of what the trust gate had held *back*. A user could see that steering was withheld and still not learn that anything fired, which one it was, or what it cost.

## What lands

`AgentEvent::SkillInjected { name, summary, tokens }` — one event per skill, so each becomes one row with its own head and cost rather than a list the renderer has to re-split.

It is built where the answer is known and nowhere else. `recall_block_reported` renders the steering section, so it is the only layer that knows which *ranked* skills the section's token budget actually admitted; `stella_core::skills::injected_skills` cuts the same `section_fit` prefix `render_skills_section` renders and prices each over `rendered_skill_block`'s exact bytes. Announcing the selection instead would name skills the prompt never carried and bill the turn for text nobody sent — the report and the render are one reading, the discipline `ProducedSteering::of` already applies to the same cut.

Per the scope note on the issue, the `TranscriptEntry::Skill` variant, its `kind_for`-side projection (`transcript_source::skill_rows`), and its `SessionModel` fold arm land here **with** the upstream emit. No variant is pre-added for any other event.

**Every door announces it, not only the deck.** `OpeningRecall` now carries the turn's opening events in send order rather than the recall receipt alone, so `run_turn`, `stella goal`'s raw and wrapped arms, and a fleet worker all reach it through the openers they already had — the door parity #4500 asked for, at no extra parameter.

## Two things elided rather than faked

- **`used n× this repo` is absent.** The store counts it (`skill_usage`, one row per skill per execution, written at the shared seam) but no live path carries that number to the deck, and `used 0×` under a skill that has just fired would state the opposite of what happened. #4337 is where the reader is decided. `the_repo_usage_count_is_elided_rather_than_zeroed` pins that it stays absent rather than becoming a placeholder somebody later reads as data.
- **The trigger is the literal `auto`.** Auto-selection is the only channel that announces itself; an explicit `/slug` is expanded into prompt text by `extensions.rs` and tells nobody. Filed as #5232. A carried field before that lands would be a column with one value — #4180's lesson, which the scope note names.

## Exemplar

The event's carrier follows `RecalledBlock`/`OpeningRecall`, already in this tree for exactly this shape: recall runs before the channel it would be emitted on exists, so the residue travels to the turn runner rather than dying at the call site (#4498). The ledger row follows `ContextWrite`'s.

## The consumer ledger says `RecordedOnly`, on purpose

`Surface` is a closed set of two — `Observatory` and `Serve` — and deliberately excludes interactive mode and replay, which render every case. The deck is this signal's only consumer today, so `consumers.rs`'s own guidance makes the honest row `ConsumerPosture::RecordedOnly` citing a tracking issue. That is **#5229**, filed with the two ways to close it. Claiming `Surfaced` here would be the euphemism that ledger exists to strip out.

## Witnesses, each flipped artisanally

Revert, watch fail, restore — verified by hand, not asserted:

| Witness | Reverted | Result |
|---|---|---|
| `render::tests::skill::an_injected_skill_renders_the_spec_head_and_its_summary`, `…::a_skill_with_no_description_renders_no_body_line` | the `TranscriptEntry::Skill` arm in `projected_rows` | both FAILED (nothing draws), pass restored |
| `memory::tests::skill_event::an_injected_skill_reaches_the_turn_stream_as_its_own_event` | `injected_skills(&kept)` → `Vec::new()` on the block | FAILED, pass restored |
| `skills::tests::injected_skills_names_only_what_the_section_budget_carried` | the `section_fit` cut in `injected_skills` | FAILED on the over-budget fixture, pass restored |

The first is the live-adapter test the issue's "Done means" asks for on the render side; the second is it on the emit side — it drives the real `recall_block_reported` over a real workspace with two `SKILL.md` files and asserts the selected one, and only it, announces itself. `a_turn_with_steering_off_announces_no_skill` guards the other direction: a session with steering off announces nothing.

No test was deleted.

## God-file note

`crates/stella-tui/src/model.rs` was 1497/1500 before this branch and the new fold arm crossed it. No baseline entry was added. Instead the two byte-identical walks that find the turn's opening rule (`name_the_open_turns_model`, `name_the_open_turns_steer`) move to `model/turn.rs` as one `open_turn` — the sibling-submodule remedy AGENTS.md names, and one stopping rule where there had been two copies of it. `model.rs` is 1469 now. `make file-size-update` was not run.

## Verification

Local, targeted (never the workspace, never `make gate` — CI is the evidence):

- `cargo test -p stella-tui` · `-p stella-core` · `-p stella-protocol` — all green
- `cargo test -p stella-cli skill_event` — green
- `cargo clippy -p stella-tui -p stella-core -p stella-protocol -p stella-cli --all-targets` — exit 0
- `RUSTDOCFLAGS="-D warnings" cargo doc … --no-deps --document-private-items` on all four — exit 0
- `cargo fmt --all`
- `make guards-fast` — 13 passed, 0 failed (includes `file-size`, `prose`, `line-citations`)
- `./scripts/check-wire-schema.sh`, `./scripts/check-consumer-sites.sh`, `./scripts/check-diagnostic-codes.sh` — all OK
- `cargo test -p stella-tui --test deck_render_snapshots` — 27 passed, **no re-bless needed**: no committed deck golden contains a skill row, so the goldens are unchanged rather than blessed-without-reading.

`docs/wire/` was regenerated with `make wire-schema-update` and `docs/reference/diagnostics.md` with `make diag-reference` (the new `agent.skill.injected` code carries a written meaning, not the TODO placeholder).

## Noticed, filed, not fixed

- **#5229** — `skill_injected` reaches no Observatory query or `Serve` arm; the ledger row cites it.
- **#5232** — an explicitly invoked skill (`/slug`) still leaves no transcript row and no `skill_usage` row, so appraisal can retire a skill for non-use that a person invokes by hand.
- **#5242** — unrelated pre-existing flake found while running the full `stella-cli` suite: `subsession::tests::a_resumed_session_never_re_mints_a_lane_id_that_already_has_a_transcript` fails in the full run and passes alone. This branch touches nothing under `subsession/` or `durability/`.

## Summary by Sourcery

Announce automatically injected skills as individually costed transcript events across every turn-opening path.

New Features:
- Emit one `SkillInjected` event for each automatically injected skill, including its name, summary, and prompt cost.
- Render injected skills as dedicated transcript rows across the TUI, with searchable summaries and spec-compliant formatting.
- Propagate skill announcements through interactive, goal, wrapped, raw, and fleet turn-opening paths.

Bug Fixes:
- Ensure only skills that actually fit within the steering section budget are announced and billed.

Enhancements:
- Add diagnostic reporting and wire-schema support for skill injection events.
- Refactor turn-opening transcript lookup into a shared helper to reduce duplication and keep the model file within its size limit.

Documentation:
- Document the new `agent.skill.injected` diagnostic and regenerate event wire schemas.

Tests:
- Add coverage for skill event propagation, budget-limited selection, steering-disabled behavior, transcript rendering, empty summaries, and wire samples.


macanderson:
**On the red `dod` check.** #5031 predates SCR-003's task template, so it carried no DoD section for the gate to read. I appended the template's checklist and ticked the four items that are verified, leaving `Full CI green` unticked — which is the one thing the gate is now failing on.

That is deliberate, not an oversight: ticking it before the run finishes would be claiming something unverified, which is the exact guess SCR-003 exists to prevent. The box gets ticked when the run is green, and `dod` re-runs on the issue edit.

**Where the run stands** — 12 of 15 green, nothing failing on the diff:

| | |
|---|---|
| pass | MSRV check · dependency-review · docs guards · **docs/wire still matches the types** · duplicate claims · **file size ratchet** · gate steps no other workflow runs · harbor_adapter + analyzer pytest · prose-only probe · main is not known-broken · no edit is undone inside the branch · PR carries a diff |
| queued | `fmt + clippy + test` · `cargo deny + cargo audit` · `guard self-tests` |
| fail | `dod` — the unticked box above |

The three outstanding are **queued behind the org's runner backlog**, not running and not failing (`gh run list` shows `queued/` for `ci`).

**Sourcery's assessment against #5031 is 3/3 ✅, no ❌ rows** — nothing to fix or rebut there.

macanderson:
Conflicts resolved, and **by merge rather than rebase** — deliberately, because this repository's own guard says so.

AGENTS.md § `rebase-replay.yml`: *"Merging B without rebasing is safe, which is what makes this a rebase hazard rather than a merge one."* Rebasing a stale branch onto a base that already contains an equivalent edit deduplicates the edit (`warning: skipped previously applied commit`) and keeps any undo beside it, turning an inert pair into a bare revert of somebody else's change. That is #4979, and `check-rebase-replay.sh` fires on a merged head today. This branch touches three files `main` also moved in that window, so a rebase was the one operation with that failure mode available to it.

## What actually landed in the window

Not quite the PRs named to me — for the record, the ones that touched my files are #5236 (the model row), #5211 (graph edges name the turn), #5245 (delete counts inbound refs) and #5219 (the palette).

`git` reported exactly one conflicted file, `crates/stella-tui/src/render/entry.rs`, with two hunks — and per the "a conflict names one file, the loss is elsewhere" rule I checked the seven auto-merged files by hand rather than trusting the clean merge:

| conflict | resolution |
|---|---|
| `projected_rows` doc | Both sides rewrote the same sentence to name their new arm. Merged text names **both** — the skill row and the settled model call. |
| `projected_rows` match | Both sides added an arm at the same position. Kept **both**, neither superseded. |

Verified surviving on both sides after the merge: `TranscriptEntry::Skill` **and** `TranscriptEntry::Model` in `model/entry.rs`; my `SkillInjected` fold **and** main's `Model` constructor (which also landed in `model/turn.rs`, beside the `open_turn` I had extracted there); `skill_rows` **and** `model_rows` in the projection; both `transcript_nav` arms; both test modules.

## The file-size check was right to be asked

`model.rs` is fine at 1432 — but `views/transcript_source.rs` was not, and the reason is not this branch: **`main` was already red on it.** #5236 grew it 1156 → 1540 against a base where it had room, so the tree only crossed once that landed. The canary caught it and filed #5256, unclaimed.

I claimed it (`./scripts/main-red-claim.sh claim`) and repaired it in **its own PR from a fresh `main`** — #5262, labelled `unblocks-main` — rather than folding a ratchet repair into a feature PR, which AGENTS.md forbids. The file already had its seam: two thirds of it was an inline `mod tests`, now `views/transcript_source/tests.rs`. Verbatim move, all 27 tests, production code untouched. 1540 → **525**.

That repair is merged into this branch too, so this PR is green on its own; once #5262 lands on `main` first, this PR's diff shrinks back to just the skill row.

## Re-verified on the merged tree

`cargo test -p stella-tui` (all 12 binaries, four skill witnesses named and passing) · `clippy --all-targets` on all four crates exit 0 · `RUSTDOCFLAGS="-D warnings" cargo doc --document-private-items` on all four exit 0 · `cargo fmt --check` exit 0 · `make prose` OK · `check-file-size` **OK** · deck goldens **28 passed, unchanged — no re-bless, so no golden was blessed without being read**.

`Full CI green` on #5031 is **unticked again**: it was a true claim for the old head and is not yet one for this tree. It goes back on when this run is green, not before.

macanderson:
`main` is repaired and merged in. #5260 landed, #5256 is closed, and this branch now carries **only the skill row again** — 34 files, +708/−108, with the `transcript_source` split coming from `main` rather than from here.

The second merge had one conflict, an add/add on `views/transcript_source/tests.rs`: both #5260 and my closed #5262 created that file. I diffed the two before resolving rather than picking blind — **identical but for the module doc comment**, same 27 tests — and took `main`'s copy, which is what I said I would do when I yielded.

Verified after it: `skill_rows` and `model_rows` both still in the projection, all 27 moved tests present, `render::tests::skill`'s four witnesses passing.

## Re-verified on this head

`cargo test -p stella-tui` — 13 binaries green · `clippy --all-targets` on all four crates — exit 0 · `RUSTDOCFLAGS="-D warnings" cargo doc --document-private-items` on all four — exit 0 · `cargo fmt --check` — exit 0 · `make prose` — OK · `check-file-size` — **OK** · deck goldens — **28 passed, unchanged, no re-bless**.

The previous head's CI already put `fmt + clippy + test` at **pass** with the merge in it; this head only swaps a doc comment in a test file and drops my duplicate of the split.

`Full CI green` on #5031 stays unticked until this run is green.

macanderson:
**Flattened.** `no edit is undone inside the branch` was right, and it was pointing at real damage waiting to happen.

## What the guard saw

This branch created `crates/stella-tui/src/views/transcript_source/tests.rs` — my own copy of the #5256 split, in the PR I later closed — and then **took `main`'s copy** when #5260 landed and I resolved the add/add conflict. Edit, then undo, inside one branch. The diff against the base does not mention the path at all, so it reads as untouched, which is exactly what makes it dangerous rather than tidy: rebase this onto a base that already carries #5260 and git drops my copy (`skipped previously applied commit`) while keeping the undo. What is left is a **live revert of #5260** — the repair that just unblocked the whole queue — and it merges cleanly with nothing to warn anyone (#4979).

## Why my local run said OK and CI said FAIL

Worth writing down, because I nearly trusted the wrong one. `./scripts/check-rebase-replay.sh` locally resolved an older base and reported `OK — 16 path(s)`. CI resolved `7d8bca0f` — the actual PR base — and reported the failure. **CI was right; the local green was a green answering a different question.** Same class of mistake as grepping cargo's output instead of reading its exit code.

## The fix

`git reset --soft 7d8bca0f` and one commit (`git rebase -i` is unavailable in this environment). History changed; content did not:

| | |
|---|---|
| tree before | `c84bcb239e8668a8092b0bb7bfb3b06eddab7936` |
| tree after | `c84bcb239e8668a8092b0bb7bfb3b06eddab7936` |

Byte-identical, which is also what carries the verification forward: every gate I ran on the merged head was run against *this* tree.

## Confirmed after flattening

- `git diff origin/main -- .../transcript_source/tests.rs` → **empty**. `main`'s copy stands untouched.
- `git log origin/main..HEAD --name-only -- <that path>` → **no mention**. The history no longer contains the pair.
- `skill_rows` present · `model_rows` present · **27** moved tests present.
- `./scripts/check-rebase-replay.sh` → **OK — 34 path(s) touched, every one of them still in the branch's diff.**

The branch is now one commit, 34 files, +708/-108.

Merging rather than rebasing is what kept this recoverable, and the guard caught the residue of the resolution rather than a mistake in it — but the residue was a loaded gun, and it is gone now.

Issue #5248, opened 2026-08-27
The panel driver reads the process environment, so stella-runtime's own no-ambient-reads gate is red


macanderson:
## Problem

`crates/stella-runtime/tests/no_ambient_reads.rs` fails on `feat/plugin-panel-ui` (PR #5223) and on anything stacked on it:

```
$ cargo test -p stella-runtime --test no_ambient_reads
crates/stella-runtime/src/panel_host.rs:166: reads ambient process state (`std::env::var`)
    if let Ok(path) = std::env::var("PATH") {
crates/stella-runtime/src/panel_host.rs:173: reads ambient process state (`std::env::var`)
    if let Ok(value) = std::env::var(name) {
test result: FAILED. 0 passed; 1 failed
```

That test is the crate's central rule enforced executably: `stella-runtime` exists so N sessions with N workspace roots can be assembled in one process, and a process global is by definition the same for all N. `panel_host::ask` resolves the child's environment by reading the calling process's, so it is exactly the read the guard exists to refuse — and a `make gate` run on that branch is red on it today.

The reads are not wrong about *what* the child should see. `PATH` genuinely has to be forwarded or `env_clear` leaves the child to `execvp`'s built-in default, and the manifest's `env` allowlist genuinely names variables whose values come from the operator's shell. What is wrong is *where they are read*: this crate must be handed those values rather than resolve them.

## Files

- `crates/stella-runtime/src/panel_host.rs` — `ask`, the `command.env(...)` block after `env_clear()` (around the `PATH` comment)
- `crates/stella-runtime/tests/no_ambient_reads.rs` — the guard, and the list of forbidden needles
- `crates/stella-tui/tests/hello_panel_demo.rs` and `crates/stella-tui/tests/plugin_panel_placements.rs` — the two call sites in tests
- `crates/stella-cli/src/command_deck/plugin_panels.rs` — `spawn_tick`, the production call site (arrives with #5227)

## Suggested shape

Give `ask` the resolved environment instead of the names to look up: a third parameter carrying the already-resolved `(name, value)` pairs, built by the host — `stella-cli` — which may read its own process environment because it *is* the process. `stella_runtime::wrapper::SubprocessWrapper::declare` is the sibling that already faces this question for a driver's environment and is where to look for the shape (and for the credential refusal it applies, which a panel should get for free rather than separately).

[...]

PR #5250, opened 2026-08-27
feat(plugin-ui): a plugin's panel lands on all three surfaces

### LINKED ISSUE #5227: Plugins can draw a panel, but nothing places one on a screen yet
## Problem

The panel protocol now has a wire contract (#5203/#5208), a host renderer (`crates/stella-tui/src/plugin_panel.rs`), a process driver (`crates/stella-runtime/src/panel_host.rs`) and a reference plugin (`plugins/stella-hello/`) that draws through all three — PR #5223, whose demo test renders a real plugin's real frame. What does not exist is the wiring that puts a panel on a real screen. **A plugin can draw; nothing asks it to.**

Three placements, from `design/tui-v2/SPEC.md` §12 and the `PanelSurface` vocabulary:

- **`settings`** — a pane in the SETTINGS tab, beside AGENTS / TOOLS / SEATS, one per installed plugin that declares the surface.
- **`overlay`** — a fenced block in the SESSION transcript.
- **`command`** — a centred popup opened by `/<plugin-name>`, plus the reserved-name refusal: `DECK_BUILTINS` (`crates/stella-cli/src/command_deck/skills.rs`) wins, and a plugin whose name collides is refused **out loud** with a notice naming the plugin and the name. Never the silent drop `slash_entries` does for `.stella/commands/*.toml` — a signed manifest a human read must not contain a line that quietly does nothing.

## Files

- `crates/stella-tui/src/views/settings.rs` — the pane list gains plugin panes
- `crates/stella-tui/src/views/session.rs` and `render/entry.rs` — the transcript overlay block
- `crates/stella-tui/src/composer.rs` + `render.rs` — the slash name and its popup
- `crates/stella-cli/src/command_deck.rs` — the tick that calls `panel_host::ask` off the draw path and lands frames in state
- `crates/stella-cli/src/plugin_cmd.rs` — the install grant that must precede any frame

## Constraints already settled, do not relitigate

- **The draw path never awaits.** `ask` runs as a task; its frame lands in state; the next draw renders it. A missed budget draws the last good frame plus `throttle_tag`.
- **The host owns the chrome** and the label is the plugin's name. `PanelGrant` carries no title on purpose — its own doc has the anti-spoofing argument.
- **A frame is one line of JSON** and the host clips in display cells, not characters.
- **`PanelLease::admits`**, not `fits`: a frame answering another surface or a stale tick is refused before geometry.

## Done means

Each surface renders a real plugin's frame in a golden; `/hello` opens the popup; a name colliding with a built-in is refused with a visible notice (witnessed); and no panel frame is requested before the install grant (witnessed on the gate, not the rendering — the shape `no_tool_call_reaches_an_ungranted_server_before_the_grant` uses for MCP).

**Estimated agent build time:** ~2 days

## Definition of done

- [x] Implementation complete — all three placements (#5250)
- [x] Scoped tests added/updated and passing — a golden per surface driving the real `plugins/stella-hello/` process, the reserved-name refusal, and the no-frame-before-grant gate witness; every flip verified artisanally
- [x] Full CI green — `fmt + clippy + test` passed in 16m11s on `4971b49f3` (2391 CLI tests, 0 failed; all three placement goldens ran the real `stella-hello` process on the runner, no python3 skip). #5264's pre-existing `subsession` flake did not fire in that run
- [x] Code comments and docs updated — `make prose` and `make line-citations` green, no baseline edited
- [x] Residue filed as new issues — #5248 (`no_ambient_reads` red on the base branch) and #5253 (`/reload` does not reseat the panel deck)

macanderson:
Closes #5227.

**Based on `feat/plugin-panel-ui` (#5223), not `main`** — that branch carries the host renderer, the process driver and the reference plugin this builds on, and is itself stacked on #5208.

A plugin could draw and nothing asked it to. Now three placements do, all from the same frame protocol (`design/tui-v2/SPEC.md` §12.2).

## The three placements

| Surface | Where it lands | Code |
|---|---|---|
| `settings` | a pane in the SETTINGS tab, after AGENTS / TOOLS / SEATS and never before them | `crates/stella-tui/src/views/settings.rs` |
| `overlay` | a host-bordered block in the SESSION tab, below the gate cards and above the transcript | `crates/stella-tui/src/views/session.rs` |
| `command` | a centred popup over the deck, opened by `/<plugin-name>` or the derived `/plugin:<name>` alias, closed by Esc | `crates/stella-tui/src/deck_ui/local.rs`, `deck_render.rs` |

The state and all three renderers live in one new module, `crates/stella-tui/src/panel_deck.rs`. The driver half — seating, the collision refusal, the tick — is `crates/stella-cli/src/command_deck/plugin_panels.rs`.

## The draw path never awaits

The loop is deck → driver → deck, and nothing in it blocks a repaint:

1. A draw renders the last frame a plugin sent and records the rectangle `plugin_panel::chrome` left inside its border.
2. `deck_shell`'s loop, on its next turn, drains `PanelDeck::requests()` and sends one `WorkspaceInput::PanelFrameWanted { slot, tick, cols, rows }` per seat that has geometry and no request outstanding.
3. The driver spawns `panel_host::ask` as a task and answers with exactly one of `Inbound::PanelFrame`, `PanelThrottled` or `PanelSilent`.
4. The next draw renders it.

The geometry rides on the request because only the deck has measured it, and the driver has never seen the terminal. `PanelSilent` exists so a failed tick still rearms the seat — a request that went unanswered would leave the panel waiting for the rest of the session. A tick that overruns its budget keeps the frame on screen and adds `throttle_tag`; the `awaiting` flag plus a one-second refresh interval is what stops a 30fps draw loop spawning thirty processes a second.

## The settings pane list — why this shape

`SettingsPane::ALL` was `const [SettingsPane; 3]`, and the pane list is now as long as the install roster, so a const cannot express it. Two other shapes were available:

- **One enum case per plugin, carrying its name.** Makes the enum open-ended and costs the compiler's exhaustiveness check on `label` — the check that is the whole reason a fourth *built-in* pane cannot arrive unlabelled.
- **An index into one flat runtime list, no enum at all.** Makes `AGENTS` and a plugin's pane the same kind of thing. They are not: the three are the product's, they are compile-time total, and they must stay first however many plugins are installed. A bare index says none of that and lets a reordering put a plugin's rectangle where `AGENTS` was.

So: the built-ins stay closed cases in `SettingsPane::BUILTIN` (still a const, still one entry per built-in), and every plugin pane is `Plugin(index)` carrying its seat. The order is **derived** by `SettingsPane::panes(&PanelDeck)` from live state, and `prev`/`next` take that list rather than reading a const, so neither can answer from a nav that no longer exists. Both stay non-wrapping and still return `None` at each end, so the key rises to the tab strip exactly as before. The index rather than the name means a pane cannot outlive the seat it draws: a plugin retracted mid-session loses its slot, and `render` lands the nav back on the first pane rather than drawing a stale rectangle under somebody else's label.

Exemplar for the seat/route split: `PluginRoster::hook_routes`, one file over. `panel_routes` is written to be the same kind of thing — the authoritative filter, consulted nowhere else.

## The reserved-name refusal

`DECK_BUILTINS` wins. A colliding name is refused **out loud**, with a notice naming the plugin and the name it wanted, and never the silent `HashSet::insert` drop `CustomExtensions::slash_entries` performs for a `.stella/commands/*.toml` row: a `.toml` a user wrote is theirs to re-read, and a signed manifest a human accepted at install must not contain a line that quietly does nothing.

The refusal is **per name**, not per plugin — a plugin whose popup name collides keeps its `settings` pane and its `overlay` block, because those are three separate things the operator agreed to and only one is unavailable. A second plugin asking for a name the first already took gets the same visible refusal.

`stella-plugin` learns nothing about the reserved list: it validates the slug's shape and stops, exactly as `PanelGrant::command`'s own doc says it must. The check is in `stella-cli`, beside the table.

## No frame before the install grant

`PluginRoster::panel_routes` is the gate. It emits nothing for a plugin the roster dropped, and `PluginRoster::compose` drops every package failing `InstalledPlugin::admitted` — a missing consent receipt, or one that no longer matches the manifest on disk. `spawn_tick` starts nothing for a slot with no route.

## Witnesses

Every flip verified artisanally: reverted the production change, watched the test fail, restored.

| Witness | File | Flipped by | Failed with |
|---|---|---|---|
| `the_settings_pane_draws_the_plugins_own_frame` | `crates/stella-tui/tests/plugin_panel_placements.rs` | `SettingsPane::panes` returning `BUILTIN` alone | `one pane per settings-surface plugin: [Agents, Tools, Seats]` — left 3, right 4 |
| `the_overlay_block_draws_in_the_session_transcript` | same | `overlay_h = 0`, `render_overlay` call removed from `views/session.rs` | `the drawn panel asked for a frame` — no band, so no lease, so no request |
| `slash_hello_opens_the_centred_popup` | same | the `panel_command` arm in `deck_ui/local.rs` returning `None` | `left: Send(Enqueue { text: "/hello" }), right: Handled` |
| `a_name_colliding_with_a_builtin_is_refused_out_loud` | `crates/stella-cli/src/command_deck/plugin_panels.rs` | disabling the `reserved.contains(name)` guard | `a refused name seats no popup: [PanelSeat { plugin: "impostor", surface: Command, command: Some("model") }]` |
| `no_panel_frame_is_requested_before_the_install_grant` | same | dropping `.filter(InstalledPlugin::admitted)` from `PluginRoster::compose` | `an ungranted plugin has no panel route` |

The three placement goldens drive the real `plugins/stella-hello/` as a real process — `python3 panel.py` — through the deck's own state and the deck's own layout, so a green run says the panel is on the screen a person is looking at rather than in a buffer the test constructed. Run them with `--nocapture` and the panels print. The settings pane:

```
  AGENTS  │  TOOLS  │  SEATS  │  ◳ hello   ←/→
╭ ◳ panel · hello ─────────────────────────────────────────────────────╮
│hello from a plugin                                                   │
│                                                                      │
│leased  70×13 cells                                                   │
│meter   ████████████░░░░░░░░░░░░  50%                                 │
│                                                                      │
│this row is drawn by a separate process                               │
│the border and title above are the host's                             │
╰──────────────────────────────────────────────────────────────────────╯
```

The gate witness copies `no_tool_call_reaches_an_ungranted_server_before_the_grant`: the security property is that nothing went out, so it is asserted about what the other side saw. The fixture plugin's `[panel.process]` is `sh -c "touch <marker>"`, and the test asserts the marker never appears — a test checking only `panel_routes().is_empty()` would pass against a build that still spawned the plugin from somewhere else. It carries the anti-vacuity half too: the same bytes, once the receipt is on record, do create the marker.

Two smaller witnesses ride along: `the_plugin_namespaced_alias_opens_the_same_popup` (the `/plugin:` form is derived, never declared) and `an_unseated_name_is_not_claimed_by_the_panel_deck` (an unknown slash still falls through to the driver).

## Checks run

- `cargo test -p stella-tui` — 14 suites green; `cargo test -p stella-cli --bin stella` — 2390 passed.
- `BLESS=1 cargo test -p stella-tui --test deck_render_snapshots` — **no snapshot changed.** The fixtures seat no plugin, so the derived nav renders the same three labels the const did. The diff was read, and it is empty.
- `cargo clippy -p stella-tui -p stella-cli --all-targets` — clean.
- `RUSTDOCFLAGS="-D warnings" cargo doc -p stella-tui -p stella-cli --no-deps --document-private-items` — clean. It caught a public doc linking the private `REFRESH`, which nothing else local does.
- `cargo fmt --all`; `make guards-fast` — every guard green, `prose`, `line-citations`, `file-size`, `god-files` and `invariants` included. No baseline was edited.
- No colour is named anywhere in the diff, so `make tokens` has nothing to say: every panel pixel resolves through `plugin_panel::ink`, which already reads `stella_tui_theme::token`.

## God file

`crates/stella-tui/src/deck_ui.rs` grew by nine lines and sits at 3673 against its 3675 ceiling, with **no baseline entry**. Those nine are the `panels` field, its `Default` entry, and two one-line dispatcher calls — `panel_deck::ingest` folds all four panel envelopes into one arm, and `panel_deck::esc_closes_popup` holds the Esc rule. Every line of panel logic is in the sibling module.

## Noticed, not fixed

**#5248** — `stella-runtime`'s `no_ambient_reads` is **red on the base branch**: `panel_host::ask` resolves the child's environment with `std::env::var`, which is the read that guard exists to refuse. Not fixed here because the remedy changes `ask`'s signature and that function belongs to #5223, still in review. It will show red in this PR's CI until #5248 lands.

## Summary by Sourcery

Add complete host-side support for rendering admitted plugin panels across all three UI surfaces without blocking the deck draw loop.

New Features:
- Expose admitted plugin panels on the SETTINGS, SESSION overlay, and command popup surfaces.
- Open command panels through plugin slash commands and the derived `/plugin:<name>` alias, with Escape dismissal.

Bug Fixes:
- Prevent ungranted or invalidated plugins from starting panel processes before installation consent is active.
- Refuse plugin command names that collide with built-in or already-claimed commands and report the refusal visibly.

Enhancements:
- Add an asynchronous, throttled panel frame lifecycle that keeps rendering non-blocking and preserves the last valid frame on failures or overruns.
- Derive the settings navigation from built-in panes and currently seated plugin panels while preserving built-in ordering and safe reseating.

Tests:
- Add end-to-end placement coverage using the reference plugin, including settings, session overlay, command popup, aliases, visibility-driven requests, command collisions, and install-grant enforcement.


macanderson:
Follow-up pushed: `28a5f553`, one defect found while re-reading the loop.

`PanelDeck::requests` read the rectangle the last draw recorded and left it there, so a closed command popup — and a settings pane on a tab nobody is looking at — kept its lease and kept asking. Once a second, that is somebody else's process started for a panel no one can see.

The request now **takes** the lease, so a rectangle exists exactly as long as it is on screen and a seat is asked only when the last draw drew it. `requests_at` states the clock: the witness needs a second draw, and a second draw inside the refresh interval would report "not asked" for the wrong reason — passing against a build that never cleared the lease at all.

Witness `a_panel_that_was_not_drawn_is_not_asked_for_a_frame`, verified artisanally like the rest: reverting the `take` fails it with `left: [1, 2], right: [1]`.

The witness table in the description now has seven rows rather than six. Everything else above is unchanged, and re-run green: `cargo test -p stella-tui` (all suites), `clippy --all-targets`, `RUSTDOCFLAGS="-D warnings" cargo doc --document-private-items`, `make guards-fast` (13 passed, 0 failed), `make prose`, `check-file-size`, and `BLESS=1 ... deck_render_snapshots` with an empty diff.

macanderson:
One more thing noticed and not fixed: **#5253** — panels are seated once, at session open, and `/reload` does not reseat them. So a plugin installed, removed or retracted mid-session shows the wrong panels until restart. The retraction case is the one that matters, since a withdrawal that does not take effect until restart is a grant that outlived the decision to revoke it.

Out of scope here rather than deferred quietly: `PanelDeck::reseat` already replaces the seat list wholesale, so the mechanism exists and nothing calls it twice. Wiring it is a separate change because the driver holds its route list for the session too, and a request in flight across a reseat could be answered into a slot that now means a different plugin — the issue carries that hazard and how to close it.

macanderson:
**Where CI stands, and why the two red checks are the same fact.**

Everything settled green except two, and both trace to #5248:

- **`fmt + clippy + test`** runs `cargo test --workspace`, which includes `stella-runtime`'s `no_ambient_reads` — red on `feat/plugin-panel-ui` before this branch existed. Control run, so this is an observation rather than an inference from the diff: checking out that branch's tree over a clean worktree and running `cargo test -p stella-runtime --test no_ambient_reads` gives `test result: FAILED. 0 passed; 1 failed` on that tree alone, naming the same two `panel_host.rs` lines. Posted on #5248.
- **`dod`** now reads #5227's Definition of done (added, since the issue predated the template) and reports one unchecked item: **Full CI green**. It is unchecked because it is not true, and it will not be ticked until #5248 lands. That check is working exactly as designed — a PR should not be able to close its issue while its own suite is red.

So this PR is not mergeable yet, and the blocking work is #5248 rather than anything in this diff. Fixing it here would change `panel_host::ask`'s signature, which belongs to #5223 and is still in review; that is the maintainer's call to make rather than mine, and it is stated here rather than decided quietly.

macanderson:
Rebased onto `a0e125909` and pushed. Three things.

## 1. #5248's call site

`panel_host::ask` now takes the child's environment, so `stella-runtime` reads no process-global state. This branch carries the caller: `resolve_env(&route.process, |name| std::env::var(name).ok())` in `spawn_tick`, **before** the spawn rather than inside the task, so the read happens at one stated point on one thread instead of on every panel task at once. `cargo test -p stella-runtime --test no_ambient_reads` → green.

## 2. #5253's seating hazard — closed here, and not with a generation counter

You were right that it is a breach rather than polish, so it is in this PR. I did not use a generation counter, and the reason is that the frame already carries both things a stale answer gets wrong.

The driver asks `PanelLease::admits` against the lease it issued; that is not enough, because the answer travels back as a slot **index**, and an index means whatever the seat list says it means when it arrives. So `PanelSlot::settle` now applies the same rule where the frame lands: a frame is accepted only when the seat asked for one, answers that seat's surface, and carries the tick that seat is waiting on. A seating generation would be a third identifier answering a question `surface` and `tick` already answer — and it would be a contract between the deck and the driver that could be got wrong later, where this cannot.

Reverting the check renders the breach, which is what the witness asserts on — the pixels, not the return value:

```
╭ ◳ panel · beta ──────────────────────╮
│alphas pixels                         │
╰──────────────────────────────────────╯
```

Five new witnesses in `panel_deck`: `a_frame_in_flight_across_a_reseat_lands_nowhere` (the breach), `a_frame_answering_its_own_lease_is_drawn` (anti-vacuity), plus the surface, the stale tick, and the unasked frame. #5253 keeps the `/reload` wiring; what it can no longer do is introduce this.

## 3. Two test defects I found by measuring, not by reading

Both were mine, and both would have reached CI as flakes:

- **The gate witness pinned `PanelSilent`.** Its fixture starts a real process, which beats the 33ms budget only most of the time — **one run in eight** came back `PanelThrottled`. Both rearm the seat, which is the property the test is actually about, so it accepts either now. 30 for 30 after.
- **The same witness moved `HOME`** via `paths::test_user_home`, for state it never read: every resolver it uses (`read_tier`, `receipt::record`, `resolve_user_plugins_dir`) is handed its tier explicitly. That made `subsession`'s resumed-lane test fail **3 runs in 8** beside it. The move is gone; 10 for 10 after. It still takes `test_env::lock()`, because `resolve_env` reads `std::env::var` and a concurrent `setenv` racing a `getenv` is UB on POSIX.

## The remaining suite failure is not this PR's — #5264

`subsession::tests::a_resumed_session_never_re_mints_a_lane_id_that_already_has_a_transcript` still fails about a third of full-suite runs after both fixes above. **Control**: `origin/feat/plugin-panel-ui`'s tree with this PR's three new files deleted outright fails the same test 3 runs in 7, naming the same two lane ids. So there is a second collider and it is on the base branch. Filed as **#5264** with the reproduction, the measurement, and the reason a filtered run never sees it.

I have not ticked **Full CI green** on #5227, and will not until it is true.

## Re-verified after all of the above

`no_ambient_reads`, `stella-tui --lib` (1261), `plugin_panel_placements` (6), `deck_render_snapshots` (27), `hello_panel_demo`, `plugin_panels` (6); clippy `--all-targets` on both crates clean; `RUSTDOCFLAGS="-D warnings" cargo doc` on all three crates clean; `make prose`, `check-file-size`, `make line-citations` green with no baseline edited; `BLESS=1 deck_render_snapshots` → still an empty diff.

macanderson:
**A third red check appeared, and it is not this PR's either: `main is not known-broken`.**

`main` is red — #5256, `file-size` on `crates/stella-tui/src/views/transcript_source.rs` at 1540 lines, a file this branch does not touch — so the hold fails every open PR by design, and it will clear when main does.

**I am not opening a fourth PR for it.** `./scripts/main-red-claim.sh check` cleared me, but a claim lapses after twenty minutes and the issue's own comments are the better evidence: **#5252 and #5260 both already carry the identical repair**, and a third session has already stood down on it. That is precisely the collision the claim mechanic exists to prevent, so the right move here is to stand down too and say so.

Where that leaves this PR:

| Check | State | Whose |
|---|---|---|
| `main is not known-broken` | red | #5256 — clears when #5252 or #5260 lands |
| `fmt + clippy + test` | running | #5264's flake may catch it — pre-existing, control run posted |
| `dod` | red | its one unchecked item is "Full CI green", which the two rows above decide |
| everything else | green | — |

None of the three is a defect in this diff. I will tick **Full CI green** on #5227 when the suite is actually green, and not before.

macanderson:
**Every check is green: 17 passing, none failing, none pending.**

The two you flagged as stale cleared without a code change, which is what "stale" meant:

- **`main is not known-broken`** — re-run, passes. #5260 landed and #5256 is closed. This branch never touched `transcript_source.rs`.
- **`dod`** — its one unchecked item was **Full CI green**, and that is now ticked on #5227 against a real run rather than an optimistic one.

**The run the tick is against** is `fmt + clippy + test` on `4971b49f3`, 16m11s, the full workspace suite:

- `stella-cli` — `test result: ok. 2391 passed; 0 failed`, `no_panel_frame_is_requested_before_the_install_grant` among them.
- All five `panel_deck` breach witnesses ran, `a_frame_in_flight_across_a_reseat_lands_nowhere` included.
- All three placement goldens ran **on the runner** — `the_settings_pane_draws_the_plugins_own_frame`, `the_overlay_block_draws_in_the_session_transcript`, `slash_hello_opens_the_centred_popup` — with no `skipped: python3` line anywhere in the log. Real `stella-hello` processes drew real frames in CI, not just on my machine.

**#5264's flake did not fire in that run, and I did not re-roll for it.** One green full-suite run does not disprove a ~1/3 flake; it is a run that happened not to interleave that way. The issue stands with its control, and this PR's green is not evidence against it.

Ready for review. Not merged.

Issue #5249, opened 2026-08-27
Nothing emits AgentEvent::ContextWrite, so the memory row and the receipt count were unreachable


macanderson:
Found while implementing #5032.

## Problem

`AgentEvent::ContextWrite` has **no producer anywhere in the workspace**. Every construction of it is a test fixture or a wire-contract sample:

```
crates/stella-protocol/src/event/tests.rs:1457
crates/stella-protocol/tests/wire_contract/samples.rs:479
crates/stella-cli/src/diag_bridge.rs:715   (a consumer match arm, not a producer)
```

Two things followed from that, and both were invisible:

1. The transcript row `✎ memory · N facts · n superseded → provider` (`render/entry.rs`, the `TranscriptEntry::ContextWrite` arm) never drew on any live path. It was reachable only from `render::tests::sample_entries`.
2. `Receipt::memories` — the `· n memories` cell on a turn's closing receipt (`views/transcript.rs`) — was fed *only* by `ContextWrite::upserts`, so it had never once rendered a number. #5032 gave it a second producer (`AgentEvent::MemoryLogged`, one per memory), which fixes the count; it does not fix this event.

`crates/stella-protocol/src/event/consumers.rs` records the *consumer* gap for this tag as #4501. That row understates the situation: the tag is not merely unconsumed, it is unproduced, and its `RecordedOnly` posture reads as "something writes this and nothing reads it".

## Decide one of two things

- **Wire a producer.** The natural site is `stella-cli/src/memory/learning.rs`, where `ContextStore::upsert` returns an `UpsertReceipt` carrying `memories_written`, `facts_asserted` and `facts_superseded` — exactly the three numbers this event's fields want. That makes the aggregate row real again, and it would then need to agree with the per-memory `MemoryLogged` events beside it so a turn's receipt does not count one write twice (`crates/stella-tui/src/model/memory.rs::fold_memory_write` is where both are folded, and its module doc states the rule).
- **Retire it.** Delete the `TranscriptEntry::ContextWrite` arm from `render/entry.rs` and the wire case with it, on the ground that `MemoryLogged` says everything it said and says which memory. Retiring a wire case is not free — a stream recorded by an older binary may carry the tag, and `AgentEvent::Unknown` is what would then receive it.

## Verify

- `rg -n "AgentEvent::ContextWrite\s*\{" crates/` names every construction; today that list is three files, all of them tests or samples.
[...]

Issue #5231, opened 2026-08-27
Add the e edit affordance to a memory row in the transcript


macanderson:
Split out of #5032.

## Problem

SPEC 6.3 gives a logged memory the footer `promotes at 0.85 · e edit · x reject`. #5032 shipped `x reject` end to end — `x` on a highlighted memory row sends `WorkspaceInput::RejectMemory`, which tombstones the memory through `Store::forget(ContextSurface::Memory, …)` so it stops being recalled and stops being re-mined. It did **not** ship `e edit`, and the footer therefore renders `promotes to RULE at 0.85 · x reject` — the two affordances that work, and no third one that does not.

That deviation from the spec is deliberate: an affordance printed on a row that does nothing when pressed is worse than an absent one. This issue is the other half.

## Why it was not done in #5032

`x` needs no new text — it names an id and a reason. `e` needs *replacement text*, so it needs a place for a person to type it. `stella memory edit <id> <text>` already exists and does the right thing (`crates/stella-cli/src/memory_cmd.rs` — a new revision on the same lineage, so the old words stop being served and the id does not change), but the deck has no editing overlay pointed at this surface. The composer is the obvious host and it is not free: a bare `e` must not steal a keystroke from someone typing, and a submission that rewrites a memory must not be dispatched as a prompt.

## Files

- `crates/stella-tui/src/deck_ui/memory.rs` — `selected_memory_rejection` is the resolver `x` uses; an edit needs the sibling that also yields the memory's current text.
- `crates/stella-tui/src/deck_ui.rs` — the `KeyCode::Char('x')` dispatcher arm is the pattern; the god file takes the arm only, the logic goes in the submodule.
- `crates/stella-tui/src/envelope.rs` — `WorkspaceInput::RejectMemory` is the shape; an edit carries `{ memory_id, text }`.
- `crates/stella-cli/src/command_deck/driver_support.rs` — `service_reject_memory` is the service pattern.
- `crates/stella-cli/src/memory_cmd.rs` — the `Edit` verb this would drive.
- `crates/stella-tui/src/views/transcript.rs` — `memory_log_footer`, where the affordance list is composed.
- `crates/stella-tui/src/deck_ui/queue_editor.rs` — the deck's existing in-place editing overlay, the closest prior art.

## Verify

- A witness in `crates/stella-tui/src/deck_ui/tests/` that `e` on a highlighted memory row opens the editor and that `e` with a non-blank composer still types the letter.
[...]

Issue #5230, opened 2026-08-27
Nothing branches on the memory promote event — wire a consumer or annotate the gap


macanderson:
Split out of #5032, which added `AgentEvent::MemoryPromoted` (SPEC 6.3's `memory` (promote) transcript event).

## Problem

`MemoryPromoted` ships with `ConsumerPosture::RecordedOnly` citing this issue. The event announces that an inferred lesson just gained instruction authority — it is now injected into the system prompt as an advisory rule — and nothing in the tree branches on that. The transcript renders one line; the diagnostic bridge records one; no Observatory query names the tag; no engine decision reads it.

That is exactly the shape `crates/stella-protocol/src/event/consumers.rs` exists to make reviewable rather than silent, and its sibling `context_write` has been sitting in the same posture since #4501.

## Files

- `crates/stella-protocol/src/event/tags.rs` — the `MemoryPromoted => "memory_promoted"` row and its posture.
- `crates/stella-protocol/src/event/consumers.rs` — the posture vocabulary and what each one claims.
- `crates/stella-cli/src/memory/learning.rs` — `SessionMemory::induce_rules`, the producer.
- `crates/stella-observatory/src/db.rs` — where an Observatory query would select the tag.

## What a consumer could be

Named as candidates, not as a decision:

- An Observatory view over promotions: which lessons reached instruction authority in this workspace, when, and on what evidence. The `audit_event_id` on the event is the join key into the promotion ledger.
- The turn receipt: `Receipt` counts `memories`, and a promotion is a materially different thing from a log — a receipt that distinguished them would say when a turn changed what steers later turns.
- The self-tuning loop, which already grades whether promoted context earns its place (`crates/stella-core/src/self_tuning.rs`).

## Verify

- `cargo test -p stella-protocol` — the ledger audit (`the_ledger_is_total_over_known_type_tags`) must stay green.
- If a real consumer lands, raise the row to `ConsumerPosture::Behavioral` naming its site, and the audit checks the posture/`surfaces` coherence.

## Done

Either the row names a site that branches on the event, or the row keeps `RecordedOnly` with a comment saying plainly that recording is the intended end state and why — the same two exits #4501 left open for `context_write`.

## Definition of done

The done conditions this issue already states, as boxes the gate can read.
Nothing here is new scope, and nothing is ticked by the migration — the
gate reads one heading and this issue used another (macanderson/oxagen#1407).

[...]

Issue #4501, opened 2026-08-23
32 AgentEvent consumer rows are still Unclassified{issue: "#2703"} and #2703 is closed — drive MAX_UNCLASSIFIED from 32 to 0 under a live issue


macanderson:
## Problem

`crates/stella-protocol/src/event/consumers.rs` sets `MAX_UNCLASSIFIED: usize = 32` (PR #4438 brought it from 33 with #3977). Every remaining `ConsumerPosture::Unclassified` row in `event/tags.rs` cites `#2703`, which is CLOSED — so the ledger's own rule ("it must cite the issue where the gap is being decided") points 32 rows at nothing live. This issue is that live number.

## Done means

One row at a time, each with its own evidence: census the consumer (`rg` the tag across stella-tui/stella-store/stella-observatory/stella-serve/stella-cli), flip the row to `Behavioral{site}` / `Surfaced` / `RecordedOnly{issue}` (citing this issue for a deliberate record-only gap), and lower `MAX_UNCLASSIFIED` by one — the test asserts equality, so the number is the diff. #3977 (PR #4438) is the worked example. Zero is the goal; `Unclassified` becomes unreachable for new variants once it is reached.

Related: #2701, #2703, #3977, invariant 10.

PR #5252, opened 2026-08-27
feat(stella-tui): render memory log and promote in the transcript (#5032)

### LINKED ISSUE #5032: Show memory log and memory promote events in the transcript
## Problem

SPEC 6.3 gives memory two transcript events (rendering `02-event-vocabulary`):

- log: `◆ memory logged · mem_id`, quoted text, the `OBSERVATION ▸ RULE ▸ FACT` ladder with the current class lit, `conf 0.62 · kind · decays`, footer `promotes at 0.85 · e edit · x reject`
- promote: `◆ memory promoted OBSERVATION → RULE · conf 0.87` with `audit event <id> · now prompt-injected`, one line

Neither exists. What ships is the v1 row `✎ memory · N facts · n superseded → provider` from `TranscriptEntry::ContextWrite`. The ladder, confidence, decay, promote threshold, and the `e edit / x reject` affordances are absent, and rejection therefore cannot teach the learner.

## Files

- `crates/stella-core` memory/reflection path — emit log and promote events with class, confidence, kind, decay, threshold
- `crates/stella-tui/src/views/transcript.rs` — ladder body rendering for `EventKind::Memory` (head exists; ladder body is new)
- `crates/stella-tui/src/views/transcript_source.rs` + `model/entry.rs` — variants from the spine issue
- Key routing for `e edit` / `x reject` on a selected memory event

## Done means

A logged observation renders the ladder with the lit class and the promotes-at footer; a promotion renders the one-liner; `x` on a memory event records a rejection. Golden or unit tests per row shape; a witness that the v1 `✎ memory` row does not double-print alongside the new one.

Spec: `design/tui-v2/SPEC.md` §6.3.

## Scope

`e edit` is **not** in scope here. SPEC 6.3 lists it in the same footer as
`x reject`, but the two cost differently: `x` names an id and a reason, while
`e` needs replacement text and so an editing overlay the deck does not have
pointed at this surface. The footer therefore renders only the affordances
that work — an affordance a row promises and the deck cannot answer is worse
than one it never offered. #5231 carries it.

## Definition of done

- [x] A logged observation renders the `OBSERVATION` / `RULE` / `FACT` ladder with its own rung lit, its confidence, kind and decay, and the promotes-at footer beneath. — `a_logged_memory_renders_the_ladder_with_its_own_rung_lit`, which asserts the lit rung as a *tone* over two classes, so it fails on a renderer that lights all three or always the first.
- [x] A promotion renders as a single row naming the move, the confidence and the audit record. — `a_promotion_renders_as_a_single_row`, `a_promotion_renders_one_row_through_the_router`.
- [x] `x` on a highlighted memory row records a rejection that survives re-learning — a tombstone keyed on the memory's text as well as its id. — `x_on_a_highlighted_memory_sends_the_rejection_and_otherwise_types`; the driver calls `Store::forget(ContextSurface::Memory, id, text, …)`, whose copied content is what the loop's restatement check compares against.
- [x] `MemoryLogged` and `MemoryPromoted` exist on the wire, declare a consumer posture, and have live producers in the reflection path. — `event/tags.rs` rows (`Behavioral` / `RecordedOnly` citing #5230); producers in `memory/learning.rs` (`memory_logged_events`, `record_auto_activation`), reaching the deck through `forward_reflection_events`.
- [x] Their `TranscriptEntry` variants, `SessionModel` fold arms and `projected_rows` router arms land together with the emit. — one PR, no pre-added unproducible variants.
- [x] The v1 `memory` aggregate row does not print beside the new block. — `a_logged_memory_renders_its_block_once_and_not_the_v1_row`, over a whole rendered transcript rather than one entry.
- [x] A witness per row shape, each verified to fail on the unfixed code. — five mutations, each named in the PR with the test it reddens.
- [x] No test deleted; no baseline raised (file-size, prose, line-citations). — `make prose`, `make line-citations`, `check-file-size` all green with no baseline edit; the PR additionally splits `transcript_source.rs` under the limit rather than grandfathering it (#5256).

macanderson:
Closes #5032

SPEC 6.3 gives memory two transcript events. Neither existed. What shipped was the v1 aggregate row `✎ memory · N facts · n superseded → provider`, which states a number where the spec states *which lesson, on which rung, at what confidence, and what it would take to promote it* — and which, it turns out, nothing in this workspace has ever emitted, so it drew on no live path at all (#5249, filed).

## What lands

**Two wire events, each with a live producer.**

`AgentEvent::MemoryLogged` — one per memory the reflection loop writes. Every number on it is derived rather than invented:

| field | where it comes from |
|---|---|
| `class` | `Observation`, always — a lesson seen once has been seen once |
| `confidence` | `confidence_from_score` over the evidence the lesson has (one occurrence, one distinct task), so the row and the promotion gate read one derivation |
| `kind` | the lesson's own `LessonKind` (`domain`/`process`), the axis that varies; the store's `memory.kind` is `reflection` for all of them |
| `decays` | a process note is true of its own turn, a domain lesson of a task the agent has not seen — the distinction `LessonKind::recall_tier` already spends the recall budget on |
| `promotes_at` | the workspace's own `auto_activate_at_confidence`, carried because it is configurable |
| `memory_id` | the mirror node's `nod_…`, now reported by `UpsertReceipt::memory_node_ids` |

That last one is the only new field on an existing type. The id is minted inside `ContextStore::upsert` from the node's kind and natural key (`memory://{lineage}`); a caller deriving it would be a second copy of an identity derivation, wrong the day any of the three parts changes. The anchor loop in the same function already states that reasoning for why it mints the mirror node rather than letting a caller name it.

`AgentEvent::MemoryPromoted` — when a mined rule clears that bar and a rule file lands in `.stella/rules/`.

**That path now records the `promotion_event` it never did.** Auto-activation wrote the rule file and no governance record, so the lifecycle's own contract — replaying `promotion_event` records in order reproduces the loop's governance state — did not hold for anything the loop activated on its own: a replay rebuilt a workspace with every user-kept rule and none of the automatic ones. The review surface's `keep` has always recorded one. `PromotionActor::System` is what makes the enforcement un-escalatable: `PromotionEventRecord::new` refuses a system actor blocking enforcement outright.

**The rendering.** A log is a block — the lesson quoted, the whole `OBSERVATION ▸ RULE ▸ FACT` ladder with this memory's rung lit and the others receding, `conf 0.62 · kind domain · decays`, and `promotes to RULE at 0.85` beneath. A promotion is **one row**, which SPEC 6.3 states as a property of the event rather than a shape a renderer may choose, so everything rides the head.

```
 │ ◆ memory logged                                          nod_83b3f1d29a
 │ "dedup keys must be stable across runs"
 │ OBSERVATION ▸ RULE ▸ FACT      conf 0.62 · kind domain · decays
 │ promotes to RULE at 0.85 · x reject
 │ ◆ memory promoted OBSERVATION → RULE · conf 0.87 · audit event prm_… · now prompt-injected
```

**`x reject`, end to end.** The deck resolves the highlighted row to the memory's id **and its text**, and the driver tombstones both through `Store::forget(ContextSurface::Memory, …)`. The text is the half that makes it hold: the loop re-mines paraphrases, so a tombstone keyed on the id alone is undone by the next turn that re-learns the same lesson under a fresh one. Serviced mid-turn as well as idle, on `u undo`'s reasoning — waiting would let the rejected memory be recalled by the very turn the reader is watching.

## What is deliberately not here

**`e edit`.** SPEC 6.3 lists it in the same footer; the footer renders `promotes to RULE at 0.85 · x reject` and no third affordance. `x` needs no new text — it names an id and a reason. `e` needs *replacement* text, so it needs an editing overlay the deck does not have pointed at this surface, and a bare `e` must not steal a keystroke from someone typing. An affordance a row promises and the deck cannot answer is worse than one it never offered. **#5231** is the handoff.

## Witnesses, and the flip for each

Five, each verified artisanally by mutating the fix and watching the test go red, then reverting:

| witness | mutation that reddens it |
|---|---|
| `a_logged_memory_renders_the_ladder_with_its_own_rung_lit` | make `ladder_spans` light every rung `TEXT` |
| `a_promotion_renders_as_a_single_row` | drop the `· audit event … · now prompt-injected` span |
| `a_logged_memory_renders_its_block_once_and_not_the_v1_row` | have the fold push a `ContextWrite` entry beside the `MemoryLog` one |
| `x_on_a_highlighted_memory_sends_the_rejection_and_otherwise_types` | send the rejection with an empty `text` |
| `an_auto_activated_rule_announces_a_promotion_the_ledger_holds` | skip the `record_event` call |

The ladder one asserts the lit rung as a **tone**, not as text, and renders two classes. Every rung's word is on the row whichever one the memory is on, so a text-only assertion passes on a renderer that lights all three, none, or the wrong one — and a single case cannot tell "lights the memory's class" from "always lights the first rung".

The ledger one checks the event's `audit_event_id` **against the ledger** rather than merely being non-empty: `audit event <id>` is printed on a screen as the handle a reader looks the promotion up by, and an id no ledger holds is worse than a promotion that renders nothing.

Plus `a_fact_states_no_promotion_it_cannot_reach` (the top rung names no rung above it, and still offers the rejection), `a_promotion_renders_one_row_through_the_router`, and `a_rule_already_on_disk_is_not_promoted_a_second_time` — the loop re-mines the same candidate every reflection turn, so announcing a promotion each time would be a row claiming a change that stopped happening turns ago.

**No test was deleted.**

## God files

Three files were at or over their ceiling once the arms landed — `model.rs` at 1544/1500, `textline.rs` at 1524/1500, `deck_ui.rs` at 3682 against its 3675 baseline. No baseline was raised. The logic went to siblings and the god files kept a dispatch:

- `model/memory.rs` — all three memory-write folds, the aggregate included. One module rather than three arms because the three are one subject and must stay consistent: they feed one counter, and the receipt must not count one write twice. `model.rs` 1544 → **1487**.
- `textline/memory.rs` — the plain door's two lines. `textline.rs` 1524 → **1487**.
- `deck_ui/memory.rs` — the `x` guard and the highlight resolution, sibling of `deck_ui/undo.rs`. The god file keeps one line. 3682 → **3668**.

## Noticed and filed

- **#5249** — `AgentEvent::ContextWrite` has no producer anywhere in the tree. Two consequences were invisible: the `✎ memory` row was unreachable, and `Receipt::memories` (`· n memories` on a turn's receipt) had never once rendered a number, because that count came only from `ContextWrite::upserts`. Feeding it from `MemoryLogged` here fixes the count; the event still needs a producer or a retirement. Its `consumers.rs` posture cites #4501 for the *consumer* gap, which reads as "something writes this and nothing reads it" — the opposite of the truth.
- **#5230** — `MemoryPromoted` ships `RecordedOnly`. Nothing branches on it; the row renders and the bridge records. Cited from the ledger row.
- **#5231** — the `e edit` affordance above.

## Evidence

`cargo test -p stella-tui` (12 suites), `-p stella-protocol`, `-p stella-context`, `-p stella-cli --bins`: green. `cargo clippy --all-targets` on all five touched crates: clean. `RUSTDOCFLAGS="-D warnings" cargo doc -p stella-tui --no-deps --document-private-items`: clean. `make guards-fast` green, `make prose` green with no baseline edit, `make line-citations` green, `check-file-size` green with no baseline edit.

`docs/wire/` regenerated (`make wire-schema-update`) and `docs/reference/diagnostics.md` regenerated (`make diag-reference`) — both generated artifacts, both committed, and the two new diagnostic codes carry written prose rather than the generator's `TODO` placeholder.

Deck goldens re-blessed with `BLESS=1 cargo test -p stella-tui --test deck_render_snapshots`: **no snapshot changed**. No deck fixture emits a memory event, so there was no diff to read — stated rather than implied, since "re-blessed" usually means there was one.

## Summary by Sourcery

Implement live per-memory logging and promotion events with auditable transcript rendering and durable rejection handling.

New Features:
- Add live memory log and promotion events to the protocol and transcript, including ladder state, confidence, promotion thresholds, stable identifiers, and audit references.
- Add `x reject` for highlighted memory entries, persisting content-aware tombstones through the CLI driver.
- Record governance events when the reflection loop automatically promotes a rule.

Bug Fixes:
- Replace the unreachable aggregate memory transcript row with per-memory events and ensure memory writes contribute to receipt counts.
- Prevent repeated promotion announcements for rules that are already present on disk.

Enhancements:
- Render memory logs as detailed ladder blocks and promotions as single auditable transcript rows across deck and plain output surfaces.
- Expose memory classes and event metadata in diagnostics and generated wire schemas.
- Add store-issued memory node identifiers to upsert receipts and reorganize memory-related logic into dedicated modules.

Documentation:
- Document the per-memory event model in a new implemented ADR and update diagnostic and wire references.

Tests:
- Add coverage for memory ladder rendering, single-row promotions, rejection input routing, auto-activation ledger records, and duplicate-promotion prevention.


macanderson:
## Two things this PR picked up on the way

**It now also repairs `main`.** `views/transcript_source.rs` crossed the 1500-line ceiling on `main` (#5256 — `main is not known-broken` was failing here for that reason, not for anything in this branch). A crossing takes no baseline entry, so the only remedy is a split, and every PR touching the file inherited the red. Its inline `mod tests` was four fifths of the file; it moves to `transcript_source/tests.rs` verbatim — every assertion, fixture and doc comment is the one it was, `use super::*` unchanged — leaving 530 lines, clear of the limit rather than merely clear of what this branch inherited.

Claimed through `./scripts/main-red-claim.sh claim` before touching it (`ok claimed #5256`), so a second session reads the claim rather than writing the same split 95 seconds later. `unblocks-main` is on the PR so the hold does not block its own repair.

I would ordinarily land a `main` repair on its own from a fresh `main` rather than fold it in here — AGENTS.md is explicit that the repair PR is the one racing every other merge. It is folded in because it is not separable: the file this branch adds a child module to is the file that crossed, and this branch cannot pass `file-size` while it is over.

**The earlier merge conflict is why CI never ran at first.** The PR sat at `mergeable: CONFLICTING` for its first three pushes, and GitHub creates no `pull_request` workflow runs for a PR whose merge commit cannot be computed — so only `cla` reported and every other context stayed absent rather than pending. Merging `main` in fixed it. Worth knowing: an absent required context reads exactly like a slow one.

## DoD

Every item on #5032 is ticked with the evidence beside it. `Full CI green` is not on the list — the gate that reads the list *is* CI, so an item asserting its own precondition can never be truthfully ticked before the check that verifies it runs. The state of these checks is the answer to that question, and it is visible here rather than restated in the issue.

macanderson:
## The branch is flattened, and here is why

`no edit is undone inside the branch` (#4979's guard) failed on `crates/stella-tui/src/views/transcript_source/tests.rs`, and it was right to.

A peer landed **the same split of that file** on `main` while this branch carried its own — same file, same seam, same remedy. The merge surfaced it as an `add/add` conflict; I took `main`'s copy after diffing the two (no test function existed in one and not the other, so nothing was lost). That left the branch's *history* containing an add and an undo of a path its *diff* no longer mentions — which is exactly the shape that, on a rebase, drops this branch's copy as `skipped previously applied commit`, keeps the undo, and turns into a live revert of the peer's split that merges cleanly.

Flattened with `git reset --soft origin/main && git commit`, the remedy the guard names. **The tree hash is identical before and after** — `2b9e21696eac5370b08ef7fffffd473c407216d8` both times — so this cost nothing but the commit boundaries, and the guard now reads `OK — 42 path(s) touched, every one of them still in the branch's diff`.

The ADR and the individual fix commits are gone as separate commits; their content is in the tree and their reasoning is in this PR and in `docs/adr/0019`. Since `main` squash-merges, the boundaries were never going to survive the merge anyway.

## Also fixed since the last push

**`cargo doc -D warnings`** — two unresolved intra-doc links in `memory/learning.rs`. `record_auto_activation` imports `PromotionActor` and `PromotionEventRecord` *inside its own function body*, and rustdoc resolves a link against the module, not a function-body `use`. Both types are genuinely reachable (`stella_core::context_record` is a `pub mod` that re-exports each), so the paths are spelled out rather than demoted to backticks — the reader keeps a target to follow. Same form `stella-plugin`'s `panel.rs` uses.

That gap is #5226's subject and I added a second data point there: this instance is nastier than a `pub(crate)` item, because the name is visibly in scope three lines below the link, and only running rustdoc finds it. `RUSTDOCFLAGS="-D warnings" cargo doc -p <crate> --no-deps --document-private-items` now runs on **every crate this diff touches** before each push, not just the obvious one — `stella-cli` was the dirty one and it is not the crate this change is about.

**`deck_ui.rs` ceiling** — `main` left 2 lines of slack and the arm needed 4. The two comment lines above the dispatch moved onto `reject_selected`'s own doc, where they belonged: a god file's arm is one line, and the explanation goes where a reader following `x` actually lands.

macanderson:
## Merged `main` again (#5247's skill event), and what the merge taught

Fourteen conflicts, ten of them the same shape: #5247 and this PR each add an `AgentEvent` variant, so we collide in every shared table — `kind.rs`, `tags.rs`, `samples.rs`, `classify.rs`, `model.rs`, `textline.rs`, `transcript_build.rs`, `transcript_nav.rs`, `diag_bridge.rs`, `render/tests.rs`. All additive; both sets of arms kept. The four `docs/wire/*` conflicts were regenerated rather than resolved, since they are generated artifacts.

**Two things worth writing down, because I got both wrong first.**

**A mechanical union is not a resolution.** I unioned the ten hunks with a regex and five of them came out malformed — the closing `},` of my variant sat on the `=======` boundary and was swallowed, in `kind.rs`, `tags.rs`, `samples.rs`, `render/tests.rs` and `diag_bridge.rs`. `cargo build` reported **clean** on three of those, because cargo replayed cached fingerprints. What actually found them was `cargo fmt --all`, which parses every file and cannot cache. It is now the first thing I run after a conflict resolution.

**The union also left a genuinely dead arm, and no lint caught it.** `model.rs` ended up with *both* my `ContextWrite | MemoryLogged | MemoryPromoted => self.fold_memory_write(event)` dispatch and `main`'s original `ContextWrite { .. } => { … }` block after it — the second unreachable, shadowed by the first, still incrementing `turn_counters.memories`. Clippy at `-D warnings` did not flag it. Removed; `fold_memory_write` already handles that arm identically. I then audited every other unioned file for the same shape and found none — the two apparent duplicates (`classify.rs`, `textline.rs`) are separate functions and correct.

## Against the coordinator's checklist

1. Conflicts kept both sets of arms. ✅
2. The doc links survived — `learning.rs:983` still reads `[`PromotionActor::System`](stella_core::context_record::PromotionActor)`.
3. `RUSTDOCFLAGS="-D warnings" cargo doc -p stella-cli --no-deps --document-private-items` — clean, and run on all four touched crates.
4. Check count went **13**, from 2. Not phantom runs.
5. `./scripts/check-rebase-replay.sh` — `OK, 42 path(s)`. Structurally safe this time rather than incidentally: the branch is one commit flattened straight onto `origin/main`, so no edit/undo pair can exist. Tree hash `b6d59d4b…` identical before and after the flatten.

Also fixed here: `textline.rs` came back from the merge at 1507 lines, over the limit outright. Both memory arms now delegate to `textline::memory::line(event)` in one line — the same shape `model/memory.rs` uses, for the same reason — taking the file to 1484.

Local state: `cargo test -p stella-tui --lib` 1285 pass; protocol, context and the `stella-cli` guarantees suite green; clippy and rustdoc clean on all four crates with fingerprints forced; `prose`, `line-citations`, `god-files`, `consumer-sites`, `diagnostic-codes`, `wire-schema`, `file-size` all green. All five witnesses still present and passing.

macanderson:
Conflicts with `main` are resolved and the branch is `MERGEABLE` / `CLEAN`, all checks green.

Two sessions converged on this independently, so the branch carries the other one's commits (`295d06da2`, `f4e2a9397`); I verified that result rather than pushing a duplicate. Same structural conclusion in both: `x` folds into `row_keys`, which #5280 introduced as the one place a bare letter's keyboard-vs-composer question gets asked.

Worth recording for the reviewer, since one of the fixes hides its own reason:

**`diag_bridge.rs` crossed the 1500-line ceiling that neither side reached alone** — 1473 on this branch, 1464 on `main`, 1505 composed. The shared-cell shape AGENTS.md describes: no pre-merge run on either branch could have seen it, because neither tree is wrong. Split into a sibling module, no baseline entry.

**The regenerated diagnostics reference was briefly wrong, and regenerating is what would have shipped it.** `make diag-reference` reads a record's field list statically off the `emit` expression. The split moved `class` and `confidence` behind a helper, so the reference came back listing three fields for a record that emits five — gate green, file freshly generated, documentation false. Fixed by inlining the chain back onto the `emit` call, so the reference is exact again.

The scanner hole behind that is filed as **#5372**: `field_names` in `scripts/diagnostic-codes.py` sets its "fields built at runtime" hedge only when it finds *no* field names, so a partly-opaque emit site publishes its visible subset as though it were complete. Nothing in the tree is wrong today — this branch's inline fixed the one instance — but a god-file split produces that shape by construction, so it will recur.

Issue #5173, opened 2026-08-27
The main-red canary files an issue with no DoD section, so the new gate blocks the repair PR that closes it


macanderson:
## Problem

Two mechanisms that each work alone deadlock when combined:

- `main-canary.yml` files a `main-red` issue automatically when the post-merge
  gate fails. Its template has no "Definition of done" section — it writes the
  failing check's output and the shared-cell explanation.
- SCR-003's `dod` check (adopted from oxagen today) fails any PR whose body
  says `Closes #N` when issue N has no DoD section.

So the repair PR for a red `main` cannot pass its own checks. Observed on
[#5171](https://github.com/macanderson/stella/pull/5171) against
[#5166](https://github.com/macanderson/stella/issues/5166):

```
**SCR-003 — this PR cannot close its issue yet.**
- macanderson/stella#5166 has no "Definition of done" section — refile it with
  the task template so the close can be verified (SCR-003).
```

This is worse than an ordinary red check because of what it blocks. `main` being
red already makes every PR's checks red (`main-red-hold.yml`), so the repair is
the one PR everything else is waiting on — and it is the one PR structurally
guaranteed to trip this. I unblocked #5166 by appending a DoD section by hand;
the next canary-filed issue will deadlock exactly the same way.

## Files

- `scripts/main-canary.sh` — builds the issue body
- `.github/workflows/main-canary.yml` — files it
- `.github/workflows/dod-check.yml` — the caller; the logic lives in
  `macanderson/oxagen` at `tools/scripts/scr-dod-check.mjs`

## Options

1. **Give the canary's template a DoD section.** Smallest change, and the
   canary already knows the done condition: the named check passes on `main`.
   It would emit something like "`make <check>` passes on the merged tree" plus
   "the canary closes this issue on its next green run".
2. **Exempt machine-filed issues from the DoD gate** — e.g. skip issues
   carrying the `main-red` label, on the grounds that their done condition is a
   check result rather than a checklist a human wrote.

(1) is better: it keeps one rule for every issue and makes the canary state its
own done condition, which is the thing SCR-003 is actually asking for. (2)
carves a hole that the next automated filer will need widened.

## How to verify

1. `make main-canary` locally reproduces the body the canary would file (it
   does not file anything).
2. Done when a canary-filed `main-red` issue carries a DoD section on arrival,
   and a PR with `Closes #N` against it passes `dod` with no hand-editing.
   Prove it on a real canary-filed issue, not a hand-written one — the point is
[...]

PR #5260, opened 2026-08-27
fix(stella-tui): split transcript_source's tests into their own file

### LINKED ISSUE #5256: main is red: file-size fails on the merged tree

Failing: **file-size**

### `file-size`

```
check-file-size: FAILED

worse.
```

canary exists (#3332).

### How to fix

```sh
git checkout main && git pull
```

...then, per failing check:

#### `file-size`

```sh
```

than only CI.

---

## Definition of done

macanderson:
Closes #5256

## What broke

`crates/stella-tui/src/views/transcript_source.rs` is 1540 lines on `main`, and
the file-size guard fails there. Every open pull request is red behind the
main-red hold until it is repaired.

## The repair

Two thirds of that file is its test module, so this is the split the guard asks
for and the seam the rest of the workspace already uses: the tests move to
`crates/stella-tui/src/views/transcript_source/tests.rs` and the parent
declares `#[cfg(test)] mod tests;`.

- Production half: 524 lines. Tests: 1021. Both clear of the 1500 ceiling.
- No baseline entry — `scripts/file-size-baseline.txt` is untouched, which is
  the point of the rule.
- No test renamed, removed or changed. The only edits inside the moved text are
  the dedent one level and a module header.

## Evidence

```
check-file-size: OK — 1827 watched files, 19 grandfathered over 1500 lines,
and nothing went over by this change.

cargo test -p stella-tui --lib transcript_source
test result: ok. 27 passed; 0 failed; 0 ignored; 1234 filtered out
```

`cargo clippy -p stella-tui --all-targets` is clean and `make guards-fast`
passes.

## No witness test

This moves lines between files and changes no behaviour. The failing guard is
the witness: it named this file before the split and reports OK after.

## Claim

`./scripts/main-red-claim.sh claim` was run before any of this was written, so
another session reading #5256 stands down rather than writing the same split.


macanderson:
Re-running `dod`: #5256 is a canary report, so it had no Definition of done and the gate refused the repair that closes it. I appended one to the issue (ticked — the split satisfies each item) rather than refiling, so the canary's own text is preserved. The underlying gap is #5173.

Issue #4680, opened 2026-08-24
Three PRs fixed the same red-main break: the canary's issue has no way to be claimed


macanderson:
## Problem

On 2026-08-24 `main` was red from a single two-line composition break, and **three separate PRs
fixed it independently**, all merging within about a minute:

| PR | Merged (UTC) | Title |
| --- | --- | --- |
| #4672 | 18:13:22 | `fix(stella-tui): box the two AgentEvent literals the #4612 witness left bare` |
| #4673 | — | `fix(stella-tui): box the two FleetMsg::Event payloads #4663 added` |
| #4674 | 18:13:48 | `fix(stella-tui): box the two FleetMsg::Event values its own tests build` |

All three changed the same two call sites in `crates/stella-tui/src/fleet_dashboard/tests.rs`
from a bare `AgentEvent` to `Box::new(AgentEvent)`. The merges happened to compose (identical
edits resolve cleanly, and `main` is not double-boxed — verified), so this cost duplicated work
and reviewer attention rather than correctness. It could just as easily not have.

## Why it happened

`main-canary.yml` files exactly one `main-red` issue (here #4671) and `main-red-hold.yml`
consumes it to hold merges. Neither has any notion of **someone is already fixing this**. Every
session that notices red `main` — and with the hold in place, every session with an open PR
notices — independently reaches the same correct conclusion and writes the same patch. The
`unblocks-main` label is the designed way past the hold, but it is applied by the fixer to their
own PR; it does not signal to anyone else that the repair is claimed.

There is prior art in this tree for the claim mechanic: `stella self-driving drive` claims
`issue:<n>` in `dispatch_claims` (`.stella/private/fleet.db`) for as long as a turn is in flight,
precisely so two loops against one clone can see each other (#4300).

## Definition of done

A session that is about to fix a `main-red` issue can cheaply discover that another one already
is, and backs off. Any of these would satisfy it — the choice is open:

1. **Assignee as the claim.** The canary's issue is unassigned; a fixer self-assigns before
   pushing, and the guidance (AGENTS.md § the canary, CLAUDE.md's red-main rule) says to check
   `gh issue view <n> --json assignees` first and stand down if it is claimed and recent.
2. **A linked-PR check.** Before opening a repair PR, query the `main-red` issue for linked PRs
   (`gh issue view <n> --json closedByPullRequestsReferences` or a `Closes #N` search) and stand
   down if one is open with green or pending checks.
[...]

PR #5262, opened 2026-08-27
fix(stella-tui): move transcript_source's tests to a sibling file

### LINKED ISSUE #5256: main is red: file-size fails on the merged tree

macanderson:
Closes #5256

`main` is red on `file-size`. `crates/stella-tui/src/views/transcript_source.rs` is 1540 lines, over the 1500-line ceiling, and carries no baseline entry — correctly, because the baseline only covers files that predate the guard and a first-time crossing gets split rather than grandfathered.

## Nobody's branch was wrong

#5236 added the model row's projection to a file that had room for it when that branch was written. The ceiling is a shared cell, so the tree only crossed once the change was actually in it, and no pre-merge run had both halves in front of it. That is precisely the composition hazard `main-canary.yml` exists to catch (#3332), and it caught this one at `f53fb2399`.

I confirmed the break is this and not something older: `check-file-size` names one file, the file is 1540 on `main` today, and `git show origin/main~N` puts it at 1156 before #5236 — so the crossing is that merge, not an inherited drift someone should hunt for.

## The split

The file already had its seam: two thirds of it was an inline `#[cfg(test)] mod tests`. Moving it to `views/transcript_source/tests.rs` follows the shape `model/tests.rs`, `deck/tests.rs` and `driver/tests.rs` already take in this workspace.

**The move is verbatim.** All 27 tests, same assertions, de-indented one level, with a `use super::*` doing the work the enclosing module used to do. Production code is untouched — which is what makes this reviewable as a move rather than as a rewrite hiding in a repair.

| | before | after |
|---|---|---|
| `transcript_source.rs` | 1540 | **525** |
| `transcript_source/tests.rs` | — | 1022 |

Both are well under the ceiling, so the next projection to land has room and does not have to pay for this one.

## Verification

- `./scripts/check-file-size.sh` — **OK**, "nothing went over by this change"
- `cargo test -p stella-tui transcript_source` — 27 passed, the same 27 the file had on `main` (counted `#[test]` on both sides)
- `cargo clippy -p stella-tui --all-targets` — exit 0
- `RUSTDOCFLAGS="-D warnings" cargo doc -p stella-tui --no-deps --document-private-items` — exit 0
- `make prose` — OK, none added
- `./scripts/check-god-files.sh` — OK
- `cargo fmt`

No test was deleted; the count is identical on both sides of the move.

## Why this is its own PR

AGENTS.md: do not fold a ratchet repair into an unrelated PR — land it on its own from a fresh `main`. This branch is cut from `main` at `f53fb2399` and contains nothing else. I found the break while merging `main` into #5247 (the #5031 skill row), which adds ~30 lines to the same file and would otherwise inherit a red gate it did not cause.

Claimed via `./scripts/main-red-claim.sh claim` before starting, so a second session reads the claim rather than writing the same fix.

## Summary by Sourcery

Split the transcript source tests into a sibling file to bring the production module back under the file-size limit.

Enhancements:
- Move the transcript source test suite into a sibling module so the production file remains below the project’s file-size limit without changing production behavior or test coverage.

Tests:
- Preserve all transcript source tests and assertions while relocating them to the new test file.


macanderson:
Closing this in favour of **#5260**, which does the same split and was opened first (03:59:47 vs 04:02:34).

Three PRs exist for #5256 — #5260, this one, and #5263 — and they are the #4680 shape exactly: every session with an open PR sees the red at once, reaches the same correct conclusion, and writes the same fix. The claim mechanic did not separate us because all three run as the same git identity, so `main-red-claim.sh check` read our own claims as ours and cleared each of us in turn. That is worth knowing: **the claim is keyed on author, so it cannot arbitrate between concurrent sessions of one author.** Noted on #5256 rather than left as a surprise for the next person.

#5260 is +1022/-1017 against this one's +1023/-1017 — the same verbatim move of the same inline `mod tests` to the same path, differing by a line of module documentation. There is nothing here it lacks, so the tiebreaker is simply which was first.

No work is lost: the split is identical, and PR #5247 (the #5031 skill row), which is where I found the break, will take #5260's copy when it merges `main` rather than carrying its own.

PR #5246, opened 2026-08-27
fix(deck,parity,ci,docs): eleven defects in the palette, the ledger, the guard, the loop and the public page

macanderson:
Thirteen fixes across the deck, the self-improvement ledger and the autonomy loop. Each carries a witness or a guard; two were checked artisanally by restoring the old text and watching the guard fail.

### 1. An ISSUES row could not say how old an issue was (#5196)

`IssueRow::updated_at` was a hard-coded `None` at its only producer, so the age column drew nothing and the heat sort's age term multiplied by a constant zero *while appearing to rank*. The cause was a field lower down: the kernel `Issue` carried a creation stamp and no update stamp, so the producer had nothing truthful to write.

`Issue::updated_at` now exists on `created_at`'s contract — RFC3339, compared lexically, empty when the provider cannot say — both `gh` reads ask for `updatedAt`, and it maps through. An empty stamp stays `None` rather than falling back to the creation stamp.

**Witnesses:** `a_rows_age_comes_from_the_trackers_update_stamp`, `an_absent_update_stamp_stays_absent`.

### 2. The Memory evolution row named a witness asserting its negation (#5198)

The row claimed the citation fold retires a record that keeps failing, and named a test whose first half asserts `"nothing may be retired on citation evidence alone"` and whose second reaches retirement only by hand-building a `SelectionHealth { failing: true, .. }` literal no production path produces. Nothing writes `memory_citations` at all (#4862).

The row now describes path-anchor validation — the mechanism that does fire — names that path's witness, and its file joins the sweep. The sweep's own limit is written where the next reader meets it: it proves a witness exists **by name** and cannot read its assertions, so it cannot tell a witness from its own negation.

### 3. The priority scheme disagreed in four places (#5216)

CLAUDE.md said `P0`–`P4`; AGENTS.md's compiled SCR summary said `P0`–`P3`; SCR-005's Directive stopped at `P3`; and `triage-guard.yml`'s regex was `/^P[0-3]$/` — while the label set has `P4` and seven open issues carry it.

The executable one was wrong in both directions: a `P4` from a non-triage login was never stripped, and an issue carrying only `P4` read as unprioritised to the guard and triaged to a human — the "never neither, never both" invariant failing silently. All four now say `P0`–`P4`. AGENTS.md was not named in the issue; CLAUDE.md's rule that such a disagreement is itself a bug to fix in the same PR is why it is here.

### 4. The command-palette rendering drew a palette that does not ship (#5215)

[...]

PR #5263, opened 2026-08-27
fix(stella-tui): split transcript_source's tests into their own file

### LINKED ISSUE #5256: main is red: file-size fails on the merged tree

macanderson:
Repairs the red `main` reported by #5256.

## What broke

`crates/stella-tui/src/views/transcript_source.rs` reached **1540 lines** on the merged tree — a first-time crossing of the gate's 1500-line ceiling. Neither contributing branch was wrong on its own; the file only crossed once both landed, which is why no pre-merge run could catch it and the canary is what found it.

A first-time crossing is **never** given a baseline entry (AGENTS.md, "God files"). The remedy is the split.

## The split

The production half is 524 lines; the tests were roughly twice that. So the tests moved, into `views/transcript_source/tests.rs` — the same `foo.rs` + `foo/tests.rs` shape the rest of the tree uses (`stella-core/src/driver/tests.rs`, `stella-cli/src/memory/uses/tests.rs`).

```
crates/stella-tui/src/views/transcript_source.rs         524 lines
crates/stella-tui/src/views/transcript_source/tests.rs  1025 lines
```

The parent keeps `#[cfg(test)] mod tests;` where the inline module was.

## Nothing was lost in the move

This is the risk a split carries, so it is checked rather than asserted:

- `#[test]` count on `origin/main`'s copy: **27**. In the new file: **27**.
- `cargo test -p stella-tui --lib transcript_source` — **27 passed**, 0 failed.
- No test renamed, reordered, or dropped. The only edit to any test line is the one-level dedent.

The dedent is safe: the file carries no raw strings, and the several `\`-continued string literals in it are unaffected because Rust already strips the leading whitespace of the line a continuation runs onto.

## Verification

- `./scripts/check-file-size.sh` — `OK — 1827 watched files ... nothing went over by this change`, and the inherited-drift line for this file is gone.
- `cargo fmt -p stella-tui`, then `cargo test -p stella-tui --lib transcript_source` — 27 passed.

Targeted per SCR-001; CI is the authority for the workspace build.

## Note on scope

Splitting buys structure, not slack. Every ceiling stays exactly where it was — reclaiming the room this earns is `make file-size-retighten`, and that is a PR of its own, safe only when nothing is blocked on it. Nothing here touches `scripts/file-size-baseline.txt`.

This lands on its own from a fresh `main` rather than folded into unrelated work, for the reason AGENTS.md gives: the repair PR is the one guaranteed to be racing every other merge.

## Definition of done

- [x] `transcript_source.rs` is under the 1500-line ceiling, by a split rather than a baseline entry.
- [x] Every test that existed before the split exists after it and passes — counted, not assumed.
- [x] `check-file-size` reports no drift for this file.
- [x] No ceiling in `scripts/file-size-baseline.txt` was moved.

Closes #5256

## Summary by Sourcery

Propagate tracker update timestamps for accurate issue aging, clarify memory evolution evidence, and separate transcript source tests to satisfy file-size constraints.

New Features:
- Add issue update timestamps from GitHub through the protocol and CLI issue rows for accurate age display and sorting.

Bug Fixes:
- Correct memory evolution parity metadata to reflect that only vanished path anchors currently trigger retirement and citations alone do not.
- Preserve empty provider update timestamps as absent rather than falling back to issue creation dates.

Enhancements:
- Split the transcript source test suite into a dedicated module to keep the production file within the file-size limit.
- Strengthen evolution witness coverage and guidance so parity checks include the relevant validation tests and reviewers verify witness assertions.

Tests:
- Add coverage confirming issue-row update timestamp propagation and absence handling.
- Move the transcript source tests without changing their coverage or behavior.
- Rename the memory lifecycle witness to reflect its currently exercised retirement behavior.


macanderson:
Closing as a duplicate: #5260 landed the same repair at 04:17Z and `main` is green again.

The two fixes are the same fix. `origin/main` now carries a 524-line `transcript_source.rs` and 27 tests in `views/transcript_source/tests.rs` — the same seam, the same shape, the same count. Nothing here is worth carrying forward.

Two things worth recording, because they are about the process rather than the file.

**The claim did not prevent the duplicate.** I ran `./scripts/main-red-claim.sh check`, was told `#5256 is unclaimed — the repair is yours to take`, claimed it, and wrote the fix; #5260 landed anyway. That is the failure mode the claim mechanic exists to prevent and it is worth knowing it can still happen — every session with an open PR sees the red at the same moment and reaches the same correct conclusion, so the window between "check" and "claim" is exactly where two sessions collide. The outcome was harmless here only because, as AGENTS.md predicts, identical edits compose.

**This PR also carried two commits it should not have.** I branched the repair off my own stacked HEAD instead of a fresh `main`, so #5263 contained the ISSUES and parity commits from #5246 as well as the split. That is what made `check-deleted-tests` fire — it saw `the_loop_closes_a_repeatedly_unhelpful_record_is_retired_and_restorable`, a test #5246 *renames*, vanish from a tree whose description never mentions it. The guard was right and the branch was wrong. AGENTS.md says a main-red repair lands on its own from a fresh `main`; I wrote that sentence into the description and then did not do it.

#5246 remains open with the two unrelated fixes, rebased onto the repaired `main`.

Issue #5277, opened 2026-08-27
The start-work estimate prices only the plan's input floor


macanderson:
## Problem

The start-work card's estimate line (#5044, SPEC 8.2 item 5) reads `~$ · ~tok · ~minutes`, and all three terms come from one token figure: the bytes of the issue plus every coupled file, through `stella_protocol::tokens::estimate_tokens_for_bytes`, counted once per task. That is the **input floor** — the cheapest way the plan can possibly run — priced and timed by `Store::model_rates`, which folds this workspace's own recorded `telemetry` rows for the session's model.

Two things it does not account for, both of which push the real number up:

1. **Output tokens.** The floor counts only what the plan reads. A task that writes a large diff bills for output the floor never saw. `model_rates` divides cost by `input_tokens + output_tokens`, so the *rate* already includes past output — but the token figure it is multiplied by does not, so the two are measuring different things.
2. **Steps per task.** One task is one pass over the context in the floor. A real task is several model calls, each re-reading a transcript that has grown.

The figure is labelled `~` and `crates/stella-cli/src/command_deck/start_work.rs`'s `estimate` states the assumption, so nothing here is a false claim. It is still an estimate a human is being asked to approve spending against, and a floor that is systematically several times low is worth less than one that is calibrated.

## The fix, if it is worth one

The data to calibrate against already exists: `executions` joined to `telemetry` gives, per past turn, the ratio of total billed tokens to the input the turn started from. A measured multiplier over that ratio would turn the floor into an estimate — and it would be a **measured** constant, so it wants a `/// MEASURED:` marker and the test AGENTS.md § "A constant set by a measurement" requires.

Do not add a multiplier that is not measured. A tuned-by-taste factor on this line is the kind of number SPEC §1 refuses.

## Files

- `crates/stella-cli/src/command_deck/start_work.rs` — `estimate`, `issue_bytes`
- `crates/stella-store/src/telemetry/model_rates.rs` — `Store::model_rates`, where a second fold would live
- `crates/stella-tui/src/views/start_work.rs` — the line that renders it

## Verify

Compare the card's estimate against what the turn it approved actually spent, over a handful of real issues in this workspace, and report the ratio. That measurement is the deliverable whether or not a multiplier follows it.

## Definition of done

- [ ] The gap between the estimate and the observed spend is measured and reported
- [ ] Any multiplier carries a `/// MEASURED:` marker and a test that pins it
- [ ] Scoped tests added/updated and passing
- [ ] Full CI green
[...]

Issue #5276, opened 2026-08-27
The start-work rendering still shows a det percentage the spec forbids


macanderson:
## Problem

`design/tui-v2/SPEC.md` §1 struck `det %` from the deck: "An earlier draft asked for the deterministic share of a turn's work as a percentage on the receipt, the task card and the start-work estimate; nothing in the workspace measures that, and this project does not build a metric to satisfy a layout." §5 repeats it: "No `det %` here, or anywhere."

Two renderings still draw one, and they are the source of truth every implementation is read against (§15: "SVGs are the source of truth"):

- `design/tui-v2/renderings/svg/10-start-work.svg` — the estimate line renders `det est 84%`
- the same file's status bar renders `det 86%`

`website/public/tui/10-start-work.svg` is a copy of the same file and carries both.

The chronology, since it is easy to read backwards: `det est 84%` has been in the SVG since the file was created (`57e587e70`, 2026-08-19, #4055); SPEC struck `det %` three days later (`121e1b883`, 2026-08-22, #4282); #5044 was filed four days after that (2026-08-26) quoting the rendering rather than the spec. The rendering is the artifact nobody updated, and it has been feeding the wrong number to every reader since.

#5044 implemented SPEC 8.2 against the *spec* rather than the rendering, so the shipped estimate line is `~$ · ~tok · ~minutes` and a golden asserts the absence (`deck_render_snapshots_pin_the_start_work_draft`). The rendering now contradicts the code and the spec, which is the state that made the next reader have to work out which one was wrong.

Other renderings should be swept for the same phrase before this is closed.

## Files

- `design/tui-v2/renderings/svg/10-start-work.svg`
- `design/tui-v2/renderings/png/10-start-work.png` (a 2x raster of the SVG — regenerate)
- `website/public/tui/10-start-work.svg`
- `design/tui-v2/SPEC.md` §1, §5 — the rule the renderings must match

## Verify

`rg -n 'det est|det [0-9]+%' design/tui-v2/renderings website/public/tui` returns nothing. Decide what replaces the two cells: the estimate line has `~$`, `~tok` and `~min` already, and the status bar's `det` cell needs either a boolean or removal — §5's `det` tags are booleans ("this call did or did not reach a model"), so a bar cell showing a percentage has no boolean form and may simply go.

## Definition of done

- [x] No rendering under `design/tui-v2/` or `website/public/tui/` shows a `det` percentage
- [x] The PNG exports match their SVGs
[...]

Issue #5275, opened 2026-08-27
A deck start-work branch is invisible to a peer on another clone until it is pushed


macanderson:
## Problem

The ISSUES tab's start-work approval (#5044, SPEC 8.2) takes the workspace's `issue:<n>` dispatch claim before it opens the branch, and drops the lease once the branch exists. That is the layering `crates/stella-cli/src/self_driving_cmd/claim.rs` already argues for: the lease closes the *race* (two claimants in the same instant, decided by one conditional write) and the branch carries the *duration*.

The branch only carries the duration for a peer that can see it. `crates/stella-cli/src/self_driving_cmd/contention.rs`'s `gather` reads branches with `git ls-remote --heads origin` — **remote** branches — and worktrees with `git worktree list`, which is local to one clone. So between a deck approval and the first push:

- a self-driving loop on the **same** clone defers (it sees the worktree, and briefly the lease);
- a loop on a **different** clone sees nothing, and can claim the same issue.

That window is however long a human takes to make their first push, which is not bounded.

## Options, none free

1. **Push the branch on approval.** Closes the window completely. Costs a network write before any work exists, and puts an empty branch on the remote for every start-work a human then cancels.
2. **Hold the lease for the deck session.** The lease is the only signal that carries liveness. Needs a holder that outlives one request — the deck driver has no state slot that does, so it would be a process-lifetime registry or new plumbing through `command_deck.rs` (a god file).
3. **Teach `contention` to read local branches too.** Cheapest, and it only helps the same-clone case, which already defers.

## Files

- `crates/stella-cli/src/command_deck/start_work.rs` — `approve`, and its module docs, which state this gap
- `crates/stella-cli/src/self_driving_cmd/contention.rs` — `gather`, `branches_naming`
- `crates/stella-cli/src/self_driving_cmd/claim.rs` — the lease and its TTL argument

## Verify

Two clones of one repository, a deck approval in the first, and a `stella self-driving` pass in the second: the pass must defer on the issue the deck is working. `crates/stella-cli/src/self_driving_cmd/contention.rs`'s own tests are the shape to extend.

## Definition of done

- [x] A peer on a second clone defers on an issue a deck session has started
- [x] The chosen option is argued in the PR against the two rejected ones
- [x] Scoped tests added/updated and passing
- [x] Full CI green
- [x] Code comments and docs updated
- [x] Residue filed as new issues (triage label only)

Related: #5044, #4300 (the `issue:<n>` claim), #4002.

[...]

Issue #5274, opened 2026-08-27
An approved start-work plan is authored as r1 and then thrown away


macanderson:
## Problem

`w` on the ISSUES tab now drafts a plan and `a` approves it (#5044, SPEC 8.2). On approval the driver takes the workspace's `issue:<n>` dispatch claim, opens the branch, and authors the plan's first revision with `stella_core::plan_graph::PlanGraph::approve` — which produces r1 and its `[:NEXT]` edges.

That graph is then **dropped**. It is reported in the outcome line (`plan r1 · 3 tasks · 3 [:NEXT] edges`) and nothing else. Two consequences, both of which SPEC 8.2 item 7 asks for and neither of which ships:

1. **It is not persisted.** `Store::record_plan_graph` is keyed on `executions.id`, and a start-work approval opens no execution — it happens before any turn. So `Store::plan_graph` cannot serve the approved plan back to anything.
2. **The breadcrumb strip does not update.** `crates/stella-tui/src/views/frame.rs`'s `breadcrumb_spans` reads the focused agent's `Plan`, which is folded from the `ScopeProposal` the *engine's* plan gate emits inside a turn. Nothing outside a turn can fill it.

Feeding a synthetic `ScopeProposal` in from the approval was considered and rejected: `crates/stella-cli/src/command_deck/task_tap/plan_gate.rs`'s `review()` calls `PlanGraph::approve` itself on the model's first task board, so the turn would author a *second* r1 with different tasks. Two different plans under one revision number makes every divergence the graph derives meaningless.

## The fix

Seed the turn's plan gate with the approved plan, so the turn's own r1 **is** the approved one:

- `PlanState::for_turn` (`crates/stella-cli/src/command_deck/task_tap/plan_gate.rs`) gains a way to start from an already-approved task list rather than from nothing.
- The start-work approval hands that list to the next lead turn, alongside the prompt it submits.
- `lead_turn.rs`'s existing `store.record_plan_graph(...)` call then persists it against that turn's execution — no new persistence path.
- The breadcrumb fills from the gate's own `ScopeProposal`, which is what it already reads.

## Files

- `crates/stella-cli/src/command_deck/start_work.rs` — `approve` builds the graph and drops it
- `crates/stella-cli/src/command_deck/task_tap/plan_gate.rs` — `PlanState::for_turn`, `review`
- `crates/stella-cli/src/command_deck/lead_turn.rs` — the existing `record_plan_graph` call
[...]

PR #5278, opened 2026-08-27
feat(stella-tui,stella-cli): w drafts a plan and waits for approval

### LINKED ISSUE #5044: Pressing w on an issue errors instead of drafting a plan for approval
## Problem

SPEC 8.2 / rendering `10-start-work` — the single best demo moment. Today `w` on an issue sends `IssueAction::StartWork`, whose only handler is a stub returning the error "start work is the self-driving loop's claim, not a tracker status" (`crates/stella-cli/src/command_deck/issues.rs:130`). The spec: a centered overlay with the draft plan (`draft plan r1 · built from issue text + graph + memory`), a sources line naming exactly what was used (issue, coupled files from the graph, applied memory RULEs with their text), a task list with contract previews (read-only tasks marked `read only · no contract`; diff-producing tasks show one done-means line with mechanism and det tag; final task `◇ verify · n gates · blocks merge`), an estimate line (`~$ · ~tok · det est % · ~minutes`), action row `a approve and start · e edit tasks · x cancel`, footer `nothing runs before approval`. On approval: branch created, plan becomes r1 with `[:NEXT]` edges, breadcrumb updates.

The stub's claim is a real design tension — the self-driving loop claims issues via `dispatch_claims` (#4300). The overlay's approval path must either take the same claim or explain why not; decide it in the PR, don't sidestep it.

## Files

- `crates/stella-cli/src/command_deck/issues.rs` — replace the stub with the drafting driver (issue text + graph coupling + memory RULEs in, `ScopeProposal`/plan draft out)
- New overlay view in `crates/stella-tui/src/views/` (ratatui `Clear` over a centered `Rect`); `deck_ui/issues_keys.rs` routing; `IssuesMode` gains the draft state
- Estimate line can reuse `crates/stella-core/src/estimator.rs`

## Done means

Golden of the overlay over a scripted issue; an approval test proving no branch and no execution before `a` (SPEC acceptance); on `a`: branch exists, plan is r1 with `[:NEXT]` edges (elide the edge-write honestly if the plan-graph issue has not landed, and say so), breadcrumb strip updates.

Spec: `design/tui-v2/SPEC.md` §8.2. Refs #4339, #4300, #4336.

**Estimated agent build time:** ~2–3 days

## Definition of done

- [x] Implementation complete
- [x] Full CI green

macanderson:
Closes #5044.

`w` on the ISSUES tab sent `IssueAction::StartWork`, whose only handler returned `"start work is the self-driving loop's claim, not a tracker status"`. Pressing the key produced an error notice. It now opens SPEC 8.2's draft-plan overlay — rendering `10-start-work`, the scenario the spec calls the single best demo moment.

## What ships

**The overlay** (`crates/stella-tui/src/views/start_work.rs`) — `Clear` over a centred `Rect`, per SPEC 8.2 item 1. #5048's move to an anchored popup is an argument about *completions* and applies to the command palette, not to a card opened by a key pressed on a row; this is the tab's second centred card beside the send-to-prompt confirmation.

```
┌ w start work ──────────────────────────────────────────────────────┐
│ w start work #151 dedup digest persists across CI runs             │
│ draft plan r1 · built from issue text + graph + memory             │
│────────────────────────────────────────────────────────────────────│
│ sources  issue #151 · graph: 2 coupled files                       │
│          crates/stella-store/src/seen.rs                           │
│          RULE dedup-keys "dedup keys must be stable across runs"   │
│────────────────────────────────────────────────────────────────────│
│ ○ 1 read the seen-set write path            read only · no contract│
│ ○ 2 persist the digest set to .stella/seen              graph · det│
│    done means: crates/stella-store/src/seen.rs is changed …        │
│ ◇ 4 verify · 5 gates                                   blocks merge│
│────────────────────────────────────────────────────────────────────│
│ estimate  ~$0.40  ~60k tok  ~8 min                                 │
│ a approve and start   e edit tasks   x cancel                      │
│ nothing runs before approval                                       │
└────────────────────────────────────────────────────────────────────┘
```

**The draft is derived, not generated** (`crates/stella-cli/src/command_deck/start_work.rs`). Tasks come from the issue's own definition-of-done list; coupled files are the code graph's neighbourhood of the paths the issue names, kept only where the index actually holds the file; RULEs are the ones whose Tier-2 guard `stella_core::rules::evaluate_guards` — the same function the tool boundary runs — refuses a write to one of those files with. That is what lets the footer say `nothing runs before approval` and mean it: a drafting model call would spend money on a plan the human has not agreed to look at, and would make the card unreproducible from the issue a reviewer can read.

**`e edit tasks` is real.** It opens the drafted list, ↑/↓ walks it, Space takes a task out or puts it back, and the approval carries exactly what survives. The action row does not name a key that does nothing.

## Two deliberate divergences, both stated

**1. No `det est %` on the estimate line.** The issue text and `10-start-work.svg` both ask for it. SPEC §1 forbids it, in terms: *"An earlier draft asked for the deterministic share of a turn's work as a percentage on the receipt, the task card and the start-work estimate; nothing in the workspace measures that, and this project does not build a metric to satisfy a layout."* §5 repeats it, and SPEC 8.2 item 5 itself now reads `~$ · ~tok · ~minutes`. The chronology settles it and runs the other way from how it reads: `det est 84%` has been in `10-start-work.svg` since the file was created (`57e587e70`, 2026-08-19), SPEC struck `det %` three days later (`121e1b883`, 2026-08-22, #4282), and #5044 was filed four days after *that* (2026-08-26) quoting the rendering rather than the spec. I followed the spec, a golden asserts the ratio's absence, and the stale rendering is **#5276**.

**2. The estimate exists only where a measurement does.** `Store::model_rates` (`crates/stella-store/src/telemetry/model_rates.rs`) folds this workspace's own `telemetry` rows for the session's model into dollars-per-token and milliseconds-per-token. The token figure is the plan's input floor — the issue's text plus every coupled file's bytes, through the engine's own byte heuristic, once per task. A workspace with no recorded calls for that model renders `no recorded calls to price this model against` rather than `~$0.00 · ~0 min`. The floor's known bias is **#5277**.

Measured rates rather than the catalog's published price, for two reasons: a price list says nothing about wall clock, and it ignores cache reads, which are most of a long session's input tokens and are billed at a fraction.

## The dispatch-claim question

The stub was not laziness and it was right about the hazard: the self-driving loop claims issues via `dispatch_claims` (`issue:<n>`, #4300), and two dispatchers working one issue is what that table exists to prevent.

**The approval takes the same claim, through the same code.** `crates/stella-cli/src/self_driving_cmd/claim.rs`'s `acquire_as` is now `pub(crate)`, and `approve` calls it with a `deck:<pid>` owner so a peer's audit line can say who holds it. A live peer's claim refuses the approval by name and **no branch is opened**; an unopenable ledger fails open, like every other probe over that table.

The lease is taken before the branch and dropped after it — the layering `claim.rs`'s own module docs argue for: the lease closes the race (two claimants in one instant, decided by one conditional write) and the branch carries the duration. That module already reserved the `issue:<n>` namespace for exactly this; the deck is now a third dispatcher on it rather than a second mechanism beside it.

**What that leaves open, said out loud:** `contention::gather` reads *remote* branches, so between the approval and the first push a loop on a **different clone** sees nothing. A loop on the same clone still defers. That window is **#5275**, with the three options and their costs.

## What is elided, and why

SPEC 8.2 item 7 also asks that "the plan becomes r1 with `[:NEXT]` edges" and "the breadcrumb strip updates".

The plan graph **has** landed (#5037, via PR #5165), so `approve` authors the r1 revision and its `[:NEXT]` chain with `stella_core::plan_graph::PlanGraph::approve` and reports both counts in the outcome line. It does not persist it, and the breadcrumb does not fill. `Store::record_plan_graph` is keyed on `executions.id` and a start-work approval opens no execution; the breadcrumb reads a `Plan` folded from the `ScopeProposal` the engine's plan gate emits *inside* a turn.

Feeding a synthetic `ScopeProposal` in from the approval was considered and rejected: `plan_gate.rs`'s `review()` calls `PlanGraph::approve` itself on the model's first task board, so the turn would author a **second** r1 with different tasks — two plans under one revision number, which makes every divergence the graph derives meaningless. The correct fix seeds the turn's gate with the approved plan so the turn's own r1 *is* the approved one, which then persists through the existing `lead_turn.rs` call and fills the breadcrumb from the gate. That is **#5274**, written as a handoff.

## Witness tests

Each flip verified artisanally. Where a witness introduces a symbol that does not exist on `main`, "fails on the old code" is a **compile failure**, and I say so rather than implying an assertion flip — but every one of them was additionally mutation-checked against the *new* tree, which is the check that proves the assertion can distinguish the claim from its opposite.

| Witness | Where | Flip |
|---|---|---|
| `nothing_runs_before_the_approval_key` | `crates/stella-tui/src/deck_ui/tests/issues.rs` | **The acceptance test.** Compile-level on `main` (`WorkspaceInput::IssueStartWork` is new). Mutation-checked both ways on this tree: making `a` return `Handled` fails it (`expected exactly one approval, got []`); making `e` emit the approval fails it (`the draft is a read — no key but 'a' may start work: [IssueDraftPlan…, IssueStartWork…]`). |
| `a_peer_holding_the_issues_claim_stops_the_approval_before_the_branch` | `crates/stella-cli/src/command_deck/start_work/tests.rs` | Compile-level on `main`. Mutation-checked: dropping the `HeldBy` arm fails it. The test also releases the peer's lease and re-approves, so the refusal is proved to be the claim and not something else. |
| `an_approval_opens_the_branch_and_authors_the_plans_first_revision` | same | Compile-level on `main`. Mutation-checked: removing the `create_branch` call fails it. |
| `drafting_creates_no_branch` | same | The driver-side half of the acceptance: every repository read the draft performs, then `git branch` compared against the tree before. |
| `deck_render_snapshots_pin_the_start_work_draft` / `_editor` | `crates/stella-tui/tests/deck_render_snapshots/start_work.rs` | The goldens over a scripted issue. The first also asserts an absence — no `det est` — which a count-based test cannot see. |
| `the_estimate_line_prices_and_times_and_reports_no_ratio` | `crates/stella-tui/src/views/start_work.rs` | SPEC §1 held at the render layer. |
| `the_rates_are_the_two_sums_divided_by_the_tokens_they_cover` and siblings | `crates/stella-store/src/telemetry/model_rates.rs` | The measured rates, including that an untimed or unbilled call is excluded rather than counted. |
| `the_plan_is_the_issues_own_definition_of_done_list`, `the_mechanism_comes_from_the_tasks_own_words`, … | `crates/stella-cli/src/command_deck/start_work/tests.rs` | The derivation, pure. |

Goldens were blessed and **read**: `overlay_start_work.txt` and `overlay_start_work_editing.txt`. Reading the first is what caught the overlay stretching to 137 columns on a wide deck and the contract tags butting against the border; reading the second is what caught a task marked "taken out" still showing a `done means` line for a check that is not going to run.

## Exemplars

The overlay follows `render_confirm_send` in `issues_tab.rs` — the same tab's existing centred card — rather than inventing a second popup idiom. The claim path reuses `self_driving_cmd::claim` unchanged rather than opening a second writer on `dispatch_claims`; that module's own docs already named this as the case the `issue:<n>` namespace was reserved for. The absence assertions follow `no_tool_call_reaches_an_ungranted_server_before_the_grant` (`crates/stella-mcp/src/toolset/tests/gate.rs`).

## Port change

`IssueProvider` gains `get(&IssueKey) -> Result<Issue, IssueError>`. The draft is built from one issue's body and must be built from the right one; `search` answers with whatever a relevance ranker matched, which is the wrong instrument for an identity question. It carries a default that searches for the key and keeps only an **exact** match, answering `NotFound` otherwise — a visible degradation, never a neighbouring issue — so no existing adapter breaks. `GhIssueProvider` overrides it with `gh issue view`, one round trip.

`IssueAction::StartWork` is deleted: nothing sends it any more, and a variant whose only handler was a refusal is not worth keeping as a tombstone.

## Checks run

Targeted only, per CLAUDE.md — the workspace build is CI's.

- `cargo test -p stella-tui` · `-p stella-store` · `-p stella-protocol` · `-p stella-cli --bin stella` — all green
- `cargo clippy -p stella-tui -p stella-cli -p stella-store -p stella-protocol --all-targets -- -D warnings` — exit 0
- `RUSTDOCFLAGS="-D warnings" cargo doc … --no-deps --document-private-items` — exit 0
- `cargo fmt --all --check` — clean
- `make prose`, `make line-citations`, `make guards-fast`, `./scripts/check-file-size.sh` — green

`check-file-size` is worth a note. `crates/stella-store/src/lib.rs` sits at its baseline ceiling, and a new `pub mod` line took it one over. Rather than raise the baseline — a shared cell, and the expedient CLAUDE.md forbids — `model_rates` is nested under `telemetry`, whose table it folds, and re-exported through the `pub use telemetry::{…}` line that was already there. Zero net lines, and the better home anyway. `deck_ui.rs` is the other one, and it is the more interesting case. This branch needs eleven lines of it — an `IssuesMode` variant, a panel field, a `Default` entry, and two dispatch lines — and every piece of new *logic* is in a sibling module. That passed locally and failed on CI at +9 over the ceiling, because CI judges the **merge**: #5250 grew the same file by nine lines while this branch was open. Each PR was under its own ceiling and the two composed over it — the shared-cell hazard AGENTS.md names, arriving through a god file rather than through `Cargo.lock`.

The fix is a split, not a bigger number. The send-to-prompt confirmation moves into `deck_ui/issues_keys.rs`, where the `issues_prompt_text` it calls already lived and where this branch's start-work handler now sits; `deck_ui.rs` ends the branch **25 lines smaller** than it started. The split it leaves is a real one rather than a line-count dodge: the two modal surfaces that hold a *selection* are in the key module, and the three that hold a text buffer stay beside the buffers they type into.

## Residue filed

- **#5274** — the approved r1 graph is authored and thrown away; seed the turn's plan gate with it so it persists and the breadcrumb fills
- **#5275** — a local start-work branch is invisible to a peer on another clone until it is pushed
- **#5276** — `10-start-work.svg` still draws the `det est 84%` SPEC §1 forbids
- **#5277** — the estimate prices only the plan's input floor; measure the gap before adding any multiplier

## Deleted tests

`issues_browse_keys_refresh_and_start_work` is **renamed** to `issues_browse_keys_refresh_and_open_the_start_work_draft` — the key no longer starts work, it opens a draft. `crates/stella-tui/src/keymap.rs`'s witness table is updated in the same commit, which is what `every_binding_names_a_witness_that_exists` caught.

## Summary by Sourcery

Replace the ISSUES-tab start-work error with a reviewable draft-plan workflow that waits for approval before claiming the issue, opening a branch, and authoring its initial plan revision.

New Features:
- Add an ISSUES-tab start-work draft overlay that derives plans from issue content and workspace context and waits for explicit approval.
- Allow users to edit drafted tasks before approval and carry the selected tasks into execution.

Bug Fixes:
- Replace the previous start-work error path with a functional draft-and-approval workflow.

Enhancements:
- Reuse dispatch claims to prevent conflicting start-work approvals and author the initial plan revision when work begins.
- Add measured, workspace-specific cost and duration estimates, with a clear fallback when no pricing telemetry exists.
- Extend issue providers with exact single-issue retrieval and reorganize ISSUES key handling around the new modal workflow.

Documentation:
- Update the TUI implementation plan to reflect the landed start-work draft and approval flow and its remaining plan persistence gap.

Tests:
- Add unit, integration, rendering, and golden coverage for draft derivation, approval gating, claim contention, task editing, estimates, and overlay rendering.

Chores:
- Remove the obsolete IssueAction::StartWork tracker action and expose the existing claim mechanism to the deck approval path.

---

## The witness

- [x] This PR includes witness tests (fail on `main`, pass here) — the table above names each one, says where its flip is compile-level rather than assertion-level, and gives the mutation that proves each assertion can distinguish the claim from its opposite.

## The gate

- [x] `cargo fmt --all --check`
- [x] clippy at `-D warnings` over the four crates this touches (`stella-tui`, `stella-cli`, `stella-store`, `stella-protocol`) — the workspace run is CI's, per CLAUDE.md's "CI builds and tests; this laptop does not"
- [x] `cargo test` over the same four crates — likewise
- [x] Docs updated: every new module carries its own header, and `design/tui-v2/IMPLEMENTATION-PLAN.md`'s P5 row said "the key routes to a stub that errors today", which this falsifies
- [x] `Closes #5044` appears above **and** as a commit trailer

## Nothing left behind

- [x] Filed: #5274, #5275, #5276, #5277 — each written as a handoff, `triage` label only (SCR-005)

## Ground-rule check

- [x] No I/O added to `stella-core` — the drafting and approval I/O is all in `stella-cli`, and `PlanGraph` is used as the pure value it already is
- [x] No new dependencies
- [x] No new outbound network calls — the tracker read goes through the existing `gh` adapter, and `gh issue view` replaces a `gh issue list` that was already the alternative

The two divergences from the issue text are argued above and are the parts most worth a second opinion: **no `det est %`** (SPEC §1 forbids the metric; the rendering that asks for it predates the amendment) and **no estimate at all where the workspace has measured nothing**.

The rejected alternative worth naming: a synthetic `ScopeProposal` emitted from the approval would have filled the breadcrumb today, and would have put two different plans under revision r1 once the turn's own plan gate authored its board. That is #5274's argument, and it is why item 7 of SPEC 8.2 is half-shipped rather than faked.


macanderson:
Answering the one ❌ in the assessment table — the third row, which bundles two separate claims. Neither is fixed here, and both for stated reasons.

### "the deterministic-estimate percentage requested by the issue"

This one has a chronology, and it runs the other way.

| What | When | Evidence |
| --- | --- | --- |
| `det est 84%` enters the rendering | 2026-08-19 | `57e587e70` — `design/tui-v2/renderings/svg/10-start-work.svg` has carried it since the file was created (#4055) |
| SPEC strikes `det %` | 2026-08-22 | `121e1b883` — "warm white off the deck, **drop det %**, and a receipt that reports its turn" (#4282) |
| #5044 is filed | 2026-08-26 | `gh issue view 5044 --json createdAt` |

So the amendment is not "later". It predates the issue by four days. What #5044's body did was quote the rendering, which predates *both* — and the rendering is the one artifact nobody updated when the spec changed. `design/tui-v2/SPEC.md` §8.2 item 5, as it stands on `main` today, reads:

> 5. Estimate line: `~$ · ~tok · ~minutes`.

with §1 and §5 both forbidding the ratio in terms:

> It does **not** report a `det %`. An earlier draft asked for the deterministic share of a turn's work as a percentage on the receipt, the task card and **the start-work estimate**; nothing in the workspace measures that, and this project does not build a metric to satisfy a layout.

Shipping `det est 84%` would mean inventing a measurement to fill a cell. The card shows `~$ · ~tok · ~minutes` and a golden asserts the ratio's absence (`deck_render_snapshots_pin_the_start_work_draft`). The stale rendering is filed as #5276.

The `det` tags that *do* ship — one per task contract — are the booleans §5 keeps: a mechanism either reaches a model or it does not, read out of one table (`MECHANISMS` in `crates/stella-cli/src/command_deck/start_work.rs`) so it cannot differ row to row.

### "persisting the approved plan and updating the breadcrumb strip"

Correct, and deliberately deferred to #5274 rather than faked. The argument is in the PR body; the short version is that the two available shortcuts are both worse than the gap:

- **Persisting it here** needs an execution row. `Store::record_plan_graph` is keyed on `executions.id`, and an approval happens before any turn opens one.
- **Filling the breadcrumb here** means emitting a synthetic `ScopeProposal`. `plan_gate.rs`'s `review()` calls `PlanGraph::approve` itself on the model's first task board, so the turn would then author a **second** r1 with different tasks. Two plans under one revision number makes every divergence the graph derives meaningless — which is the whole point of the plan graph #5165 just landed.

What *does* ship on `a` is the rest of item 7 and it is tested: the `issue:<n>` claim is taken, the branch is opened under it, and `PlanGraph::approve` authors r1 with its `[:NEXT]` chain (`an_approval_opens_the_branch_and_authors_the_plans_first_revision` asserts `plan r1 · 2 tasks · 2 [:NEXT] edges`). #5274 is written as a handoff: seed the turn's plan gate with the approved list so the turn's own r1 *is* the approved one, at which point the existing `lead_turn.rs` call persists it and the breadcrumb fills from the gate — no new persistence path, and no second r1.

`design/tui-v2/IMPLEMENTATION-PLAN.md`'s P5 row is updated in this PR to say exactly that, so the next reader finds the gap named rather than having to derive it.

Issue #5294, opened 2026-08-27
no plugins installed — add one with `stella plugin install <dir>` has no machine-readable form, so what a plugin is allowed to do can only be read as prose


macanderson:
## Problem

`stella plugin list` prints its whole report with `println!` and has no machine-readable form, so the one question a script or an audit most wants to ask about a plugin — *what is it allowed to do on this machine* — has to be answered by parsing prose. #5056 added a third grant to that report (the panel grant, allowed / denied / undecided / drifted), which makes the gap wider: three of those four states are refusals, each rendered as an English sentence naming a command.

The two grants a plugin holds are now durable files on disk with serde types behind them (`ConsentReceipt`, `PanelGrantRecord`), so the data exists in a shape a report could emit directly. Nothing does.

## Files

- `crates/stella-cli/src/plugin_cmd.rs` — `list()`, which prints the report; `driver_lines()` and `configure_lines()` are the two halves already factored into pure `Vec<String>` builders for testability, and are the shape the rest would follow.
- `crates/stella-cli/src/plugin_cmd/roster.rs` — `InstalledPlugin` carries `consent: ConsentState` and `panel_grant: PanelGrantState`; `hook_routes()` and `panel_routes()` are what a host actually dispatches.
- `crates/stella-cli/src/plugin_cmd/receipt.rs`, `panel_grant.rs` — the two records, both already `Serialize`.
- Look at how other subcommands spell `--output-format json` before inventing a shape; the flag exists elsewhere in the CLI and the vocabulary should match.

## Constraints already discovered

- `plugin_cmd.rs` is at roughly 1350 of its 1500-line ceiling. A JSON report wants its own module rather than more lines here (`panel_snapshots.rs` and `plugin_cmd/receipt.rs` are the precedents).
- The report must not print a plugin's own prose unflattened into a JSON string: `stella_plugin::consent_text`'s `one_line` argument holds here too — a `description` carrying a newline or an escape is third-party text, and a consumer piping this into a terminal is exactly the hazard that function exists for.
- `stella plugin list` writes package-withheld notices to **stderr** and the report to stdout. A JSON mode has to keep that split or it emits invalid JSON the first time a tier holds a drifted receipt.

## Verify

A witness asserting the serialized report for a fixture tier: one plugin with a hook route and an allowed panel, one with a denied panel, one whose receipt drifted. Assert on the parsed value, not on the string, so field ordering is not what the test pins.

## Done

[...]

Issue #5293, opened 2026-08-27
A plugin panel grant can only be answered from the command line, not from the deck


macanderson:
## Problem

SPEC 12.4 asks for `[a]llow  [d]eny` — two keystrokes — as the panel grant. #5056 landed the grant, the durable record and the handshake document, but the two keystrokes are only asked **on stdout**, by `stella plugin install` and by `stella plugin panel <name>`. Inside the deck, an undecided panel renders its handshake in the transcript and names the CLI verb; there is no way to answer it without leaving the session.

That is a worse experience than the spec describes, and it is the one place a person actually meets a panel: they open a session, see a block saying `hello` wants to draw on their screen, and have to quit or open a second terminal to say yes.

## Files

- `crates/stella-cli/src/command_deck/plugin_panels.rs` — `handshakes()` builds the blocks that are rendered today; it is where the decision would be raised from.
- `crates/stella-cli/src/command_deck.rs` — the seating block near the `DeckOptions` assembly sends them with `chrome_note`.
- `crates/stella-tui/src/envelope.rs` — `Inbound` / `WorkspaceInput`; the card needs a request variant and an answer variant.
- `crates/stella-tui/src/views/approval.rs` — `ApprovalOverlay` is the nearest existing shape (`Inbound::ApprovalAsked` / `WorkspaceInput::ApprovalAnswered`, a parked dispatch waiting on a yes/no). It is typed to `stella_tools::registry::approval::ApprovalRequest`, so it cannot be reused as-is; read it for the ingest/key-handling pattern rather than to extend it.
- `crates/stella-cli/src/plugin_cmd/panel_grant.rs` — `record()` is what the answer writes; it takes a tier, a scope, an entry name and the manifest bytes, all of which the driver has.

## Constraints already discovered

- `crates/stella-tui/src/deck_ui.rs` is a **grandfathered god file closed to growth** (AGENTS.md § "God files"). `panel_deck::ingest` and `panel_deck::esc_closes_popup` exist precisely so panel state does not add arms to that file's dispatcher — a panel-grant card should follow the same shape and live beside the panels.
- The answer must land through the same `panel_grant::record` the CLI uses. A second writer would be a second place the "only the code path that shows a human the handshake writes a grant" rule can be broken.
- Answering must reseat the panel deck, or the allow takes effect only at the next session. #5253's `/reload` reseating path is what that hangs off.
[...]

PR #5295, opened 2026-08-27
feat(stella-cli,stella-plugin): a panel draws only after a handshake and an allow

### LINKED ISSUE #5056: A plugin panel needs a visible handshake and an allow or deny grant
## Problem

SPEC 12's trust half: before any panel renders, a handshake — manifest signature, declared capabilities, declared denials — then an explicit `[a]llow [d]eny` panel grant, rendered in the transcript on a silver rail. None of it exists (no panel capability exists yet at all; this issue is the consent + rendering layer over the other two).

## Files

- `crates/stella-cli/src/plugin_cmd.rs` neighborhood — the grant flow, following the existing install-consent pattern (`stella_plugin::consent_text`, `project_code_execution_trusted()`)
- `crates/stella-tui` transcript — render the handshake as a silver-rail event
- Blocked by the panel wire-contract issue and the panel-host issue

## Done means

First render of a plugin panel is preceded by a visible handshake and an explicit grant; denial renders no panel; the grant is recorded so the next session can honor it; a test proves no panel frame is blitted before the grant.

Spec: `design/tui-v2/SPEC.md` §12.

**Estimated agent build time:** ~1 day

## Definition of done

- [x] Implementation complete — SPEC 12.4's handshake renderer (`stella_plugin::panel_handshake_text`), the durable allow/deny record (`plugin_cmd::panel_grant`), `panel_routes()` gated on it, `stella plugin install`'s second prompt, `stella plugin panel <name>`, and the transcript handshake before the first seat
- [x] Scoped tests added/updated and passing — the deny witness asserts on the panel process's marker file, so it fails on "a third party's program ran" rather than on a route list; `cargo test -p stella-plugin` and `cargo test -p stella-cli --bin stella plugin` green
- [x] Full CI green — PR #5295, every check green on head `316025b77` except the `dod` gate reading this list
- [x] Code comments and docs updated — module headers on both new modules, and `website/content/docs/commands/plugin.mdx` gains the `stella plugin panel` section
- [x] Residue filed as new issues (triage label only) — #5293, #5294, #5300

macanderson:
## What & why

SPEC 12.4's trust half. A `[panel]` block asks for a rectangle of the operator's terminal and fills it with a third party's glyphs every tick, and until now the only thing standing in front of it was the **install** receipt — so accepting a package's tools and hooks silently accepted somebody else's program painting part of the screen, and there was no way to withdraw that without uninstalling.

Three pieces:

- **`stella_plugin::panel_handshake_text`** renders SPEC 12.4's document: the manifest signature, the declared capabilities, the declared denials, the surfaces asked for, the slash name, and the process that draws it — with `[a]llow  [d]eny` under it. Pure, no I/O, no clock; the signature is the host's digest passed in, so the near-leaf crate takes **no new dependency** (its own manifest asks for that argument to be made afresh rather than cited, and the answer here is "don't make it").
- **`plugin_cmd::panel_grant`** writes the answer beside the install receipt, keyed on the same digest, so a `[panel]` that grows a surface or a wider process is asked again. **Undecided withholds in both tiers** — `receipt`'s project-tier carve-out does not carry over, because `STELLA_TRUST_PROJECT` answers "may this repository's code run at all", not "may this plugin draw on my screen".
- **`PluginRoster::panel_routes` is gated on it**, one layer inside `compose`: a denied panel costs the plugin the rectangle and nothing else — its tools, skills and hooks keep running. That is what makes deny a usable answer rather than a euphemism for uninstall, and it has its own witness.

`stella plugin install` asks the question for what it is copying (`--yes` answers it along with the install, since the consent document it accepted already carries the panel section). **`stella plugin panel <name> [--allow|--deny]`** answers it for the three cases an install prompt never reaches: a package that arrived by `git clone` into `.stella/plugins`, one whose manifest changed after the grant, and a grant being withdrawn. `remove` forgets the grant with the receipt.

Before any panel is seated, the session renders the handshake **in the transcript**: the whole document where the decision is still open, and the standing grant plus the way to withdraw it where it is not.

Closes #5056

## The witness

Each verified by hand — revert the named change, watch the named test fail, restore.

| Test | Fails without |
| --- | --- |
| `no_panel_frame_is_requested_when_the_panel_grant_does_not_allow_it` | the `panel_routes` gate. **Asserted on the panel process's marker file**, and asserted *before* the route list, so a flip fails on "a third party's program ran on this machine" rather than on the bookkeeping that is only evidence about it — the ordering `a_frame_in_flight_across_a_reseat_lands_nowhere` uses. Verified: with the gate reverted it panics `a panel process ran without an allow: …/the-panel-process-ran`. |
| `a_denied_panel_costs_the_plugin_nothing_but_the_rectangle` | the gate being inside `panel_routes` rather than in `compose`. Without it, "deny" could be implemented by dropping the package and every other assertion would still pass — while the operator lost the plugin's tools and hooks too. |
| `the_first_seat_is_preceded_by_a_handshake_for_every_panel_plugin` | `plugin_panels::handshakes` |
| `the_panel_handshake_carries_the_five_things_spec_12_4_names` | `panel_handshake_text` (compile-level: a new function) |
| `author_prose_cannot_forge_a_line_of_the_handshake` | `one_line` over the author strings the handshake composes. **Behavioural, and it catches a gap the existing prompt tests did not**: dropping `one_line` from `capability_grant`'s `purpose` leaves `author_prose_cannot_forge_a_line_of_the_prompt` green and fails this one, because that test exercises `description` and this one exercises `purpose`. |
| `only_an_explicit_allow_admits_and_every_refusal_names_the_command` | `PanelGrantState::admits` / `notice` |
| `a_grant_answers_only_for_the_manifest_the_scope_and_the_name_it_covers` | the digest + scope + name key |
| `a_manifest_with_no_panel_renders_no_handshake`, `the_handshake_is_deterministic`, `an_unreadable_grant_is_drift_rather_than_absence`, `forgetting_a_grant_removes_only_that_one` | the surrounding rules |

## The gate

- [x] `cargo fmt --check`
- [x] `make guards-fast` (green, including `prose`, `file-size`, `command-docs`, `consumer-sites`, `diagnostic-codes`, `tokens`)
- [x] `cargo test -p stella-plugin`, `cargo test -p stella-cli --bin stella plugin` (169 + 160 green)
- [ ] `cargo clippy --workspace`, `cargo test --workspace` — CI's, per CLAUDE.md ("CI builds and tests; this laptop does not")
- [x] Docs updated: `website/content/docs/commands/plugin.mdx` gains a `stella plugin panel` section, the synopsis, and the `panel:` line in the `list` sample
- [x] `Closes #5056` in this description **and** as a commit trailer

## Where this follows the SPEC over the issue

Two places, both called out because the choice is reviewable rather than obvious:

1. **No caption, and no plugin-supplied heading in the handshake either.** The issue asks for a "handshake"; SPEC 12.3's rule is that a plugin cannot name chrome a reader trusts, and a consent document is chrome a reader trusts more than most. Every string in the document is either Stella's own or a manifest field flattened by `one_line`, and `author_prose_cannot_forge_a_line_of_the_handshake` pins it.

2. **The handshake goes in the transcript, not the boot dialog — but not on a silver rail, and that half is filed.** SPEC 12.4 says "Render the handshake in the transcript (silver rail)". `command_deck.rs`'s own `system_notice` doc argues the opposite policy for startup chrome ("the transcript is the home for agent and user messages only"), and the SPEC wins here: a consent document has to be scrollable and still there after the transient dialog is gone. `chrome_note` folds the lead to `Running`, so the idle assert this file already makes at startup is repeated after the blocks — the idiom the `InitCompleted` arm uses.

   The parenthetical SPEC 12.4 adds — "(silver rail)" — is **not** met, and saying so is the point of this paragraph. The rail metal is a property of `EventKind`, and every kind is derived from a tool call; host prose reaches the transcript as a fabricated `AgentEvent::Text`, with no rail. So the handshake currently renders as though the agent said it, which is the wrong direction for SPEC 12.3's argument. Giving host consent prose a rail of its own is a `stella-tui` transcript-vocabulary change with golden frames behind it, and it is #5300 rather than something bolted onto this PR.

The `[a]llow [d]eny` keystrokes are asked where the other consent prompt is asked: on stdout, in `stella plugin install` and `stella plugin panel`. The transcript block names that verb. An in-deck keyed card would need new envelope variants, a new overlay, and key handling in `deck_ui.rs` — a god file closed to growth — and is filed rather than smuggled in (see below).

## Nothing left behind

- [x] Filed: #5293 (the grant can only be answered from the command line, not from the deck), #5294 (`stella plugin list` has no machine-readable form), #5300 (the handshake is in the transcript but not on a silver rail)

## Ground-rule check

- [x] No I/O added to `stella-core`; **no new dependency anywhere** — the handshake takes the digest rather than computing it, precisely so `stella-plugin` does not grow a `sha2` edge
- [x] The new cross-boundary type (`PanelGrantRecord`) is serde-first, round-tripped by `a_grant_answers_only_for_the_manifest_the_scope_and_the_name_it_covers`
- [x] No new `AgentEvent` variant, so no consumers-ledger row and no `docs/wire/` regeneration

- **Upgrade behaviour.** A plugin already installed with a `[panel]` block stops drawing after this lands, until `stella plugin panel <name>` is run. That is the intended direction for a security gate that defaults closed, and the session says so in the transcript with the verb that fixes it — but it *is* a behaviour change for an existing install, and it is called out rather than left to be discovered.
- `plugins/stella-hello` and the `hello_panel_demo` / `plugin_panel_placements` suites construct their routes from the manifest directly rather than through the roster, so the reference demo is unaffected.
- #5185 (a frame's cell count is chars rather than display width) is untouched; nothing here changes the clip that makes it safe.

## Summary by Sourcery

Gate plugin panels behind a separate, manifest-bound operator grant while preserving the rest of the plugin when screen access is denied.

New Features:
- Add a separate panel consent workflow with `stella plugin panel`, installation-time prompting, explicit allow/deny flags, and persistent grants tied to the exact manifest, scope, and plugin name.
- Render SPEC 12.4 panel handshakes for pending panels and standing-grant notices in the session transcript.
- Expose panel grant status in `stella plugin list` and document the new command and behavior.

Bug Fixes:
- Prevent unapproved or changed panel plugins from starting their drawing processes or receiving panel routes.
- Ensure denying a panel removes only its screen rectangle while preserving the plugin's tools, skills, and hooks.
- Remove panel grants when their associated plugin is removed.

Enhancements:
- Separate screen-drawing authorization from package installation consent and default panel access to withheld in every plugin tier.
- Centralize deterministic handshake and grant-state messaging, including drift and corrupt-grant handling.

Documentation:
- Document the panel command, handshake contents, grant lifecycle, and deny behavior in the plugin command guide.

Tests:
- Add coverage for panel route gating, explicit allow/deny behavior, grant scope and manifest matching, handshake contents and ordering, author-prose safety, deterministic rendering, drift handling, and grant cleanup.

### DIFF
--- a/crates/stella-cli/src/command_deck.rs
+++ b/crates/stella-cli/src/command_deck.rs
@@ -747,13 +747,21 @@ pub async fn run_deck_session(
     // SPEC 12.4's handshake, before anything is seated (#5056). In the
     // **transcript**, which is what the spec asks for and where a consent
     // document belongs: it has to be scrollable and still there after the boot
-    // dialog is gone, which the transient notice above is not. `chrome_note`
-    // folds the lead to `Running`, so the idle assert this file already makes
-    // at startup is repeated below rather than left contradicted.
+    // dialog is gone, which the transient notice above is not.
+    //
+    // Through `deck_tx`, so it is not journaled. That is `chrome_note`'s own
+    // contract — boot narration that re-runs every launch must not replay and
+    // pile up on a resumed transcript — and it costs the audit nothing here:
+    // what a person answered is durable in the grant file beside the package,
+    // with the moment they answered it, and this block is the reminder rather
+    // than the record.
+    //
+    // `chrome_note` folds the lead to `Running`, so the idle assert this file
+    // already makes at startup is repeated below rather than left contradicted.
     let handshakes = plugin_panels::handshakes(&panel_roster);
     let spoke = !handshakes.is_empty();
     for handshake in handshakes {
-        let _ = in_tx.send(chrome_note(handshake));
+        let _ = deck_tx.send(chrome_note(handshake));
     }
     if spoke {
         let _ = in_tx.send(Inbound::Status {
--- a/crates/stella-cli/src/command_deck.rs
+++ b/crates/stella-cli/src/command_deck.rs
@@ -744,6 +744,23 @@ pub async fn run_deck_session(
     for refusal in seating.refusals {
         let _ = deck_tx.send(system_notice(refusal));
     }
+    // SPEC 12.4's handshake, before anything is seated (#5056). In the
+    // **transcript**, which is what the spec asks for and where a consent
+    // document belongs: it has to be scrollable and still there after the boot
+    // dialog is gone, which the transient notice above is not. `chrome_note`
+    // folds the lead to `Running`, so the idle assert this file already makes
+    // at startup is repeated below rather than left contradicted.
+    let handshakes = plugin_panels::handshakes(&panel_roster);
+    let spoke = !handshakes.is_empty();
+    for handshake in handshakes {
+        let _ = in_tx.send(chrome_note(handshake));
+    }
+    if spoke {
+        let _ = in_tx.send(Inbound::Status {
+            agent: LEAD.to_string(),
+            status: AgentStatus::WaitingInput,
+        });
+    }
     let _ = in_tx.send(Inbound::PanelsSeated(seating.seats));
 
     // Honour the persisted colour theme (`ui.theme`) before the deck spawns its
--- a/crates/stella-cli/src/command_deck/plugin_panels.rs
+++ b/crates/stella-cli/src/command_deck/plugin_panels.rs
@@ -109,6 +109,84 @@ pub(super) fn seat(routes: &[PluginPanelRoute]) -> Seating {
     seat_with(routes, &deck_reserved())
 }
 
+/// The handshake blocks a session owes its operator before any panel draws —
+/// SPEC 12.4's first sentence (#5056).
+///
+/// One block per installed plugin that declares a `[panel]`, in roster order,
+/// and none for a workspace with no panel plugins at all.
+///
+/// # Why an allowed panel still says something, and why it is one line
+///
+/// SPEC 12.4 asks for the handshake *before any panel*, not before the first
+/// one ever. A rectangle drawn by somebody else's program on every boot is a
+/// standing grant, and a standing grant that is never restated becomes
+/// invisible — which is the same failure the install receipt exists to catch,
+/// one surface later. But the full document is what you read to *decide*, and
+/// re-reading it at every boot for a decision already made would train a reader
+/// to skip the block that matters. So an allowed panel gets the two facts that
+/// change — what it draws, and the signature the grant covers — and the verb
+/// that withdraws it.
+///
+/// A panel that is **not** allowed gets the whole document, because that
+/// reader has a decision in front of them.
+///
+/// # The signature is re-read from disk
+///
+/// Not digested from the roster's own parse: the grant on disk is keyed to the
+/// bytes `read_tier` will hash on the next load, and a signature computed from
+/// anything else would show a reader a number that decides nothing. A manifest
+/// that has become unreadable since the roster was composed yields no block
+/// rather than a block with a fabricated signature in it.
+pub(super) fn handshakes(roster: &crate::plugin_cmd::roster::PluginRoster) -> Vec<String> {
+    let mut blocks = Vec::new();
+    for plugin in roster.plugins() {
+        if plugin.manifest.panel.is_none() {
+            continue;
+        }
+        let Ok(Some((_, text))) = crate::plugin_cmd::roster::read_manifest(&plugin.dir) else {
+            continue;
+        };
+        let signature = format!(
+            "sha256:{}",
+            crate::plugin_cmd::receipt::digest(text.as_bytes())
+        );
+        let name = &plugin.manifest.name;
+        if plugin.panel_grant.admits() {
+            let surfaces: Vec<&str> = PanelSurface::all()
+                .iter()
+                .filter(|surface| {
+                    plugin
+                        .manifest
+                        .panel
+                        .as_ref()
+                        .is_some_and(|panel| panel.draws(**surface))
+                })
+                .map(|surface| surface.as_str())
+                .collect();
+            blocks.push(format!(
+                "◳ panel · {name} — you allowed this plugin to draw on your screen \
+                 ({}), under manifest signature {signature}. \
+                 `stella plugin panel {name}` withdraws it.",
+                surfaces.join(", ")
+            ));
+            continue;
+        }
+        let mut block = String::new();
+        if let Some(handshake) = stella_plugin::panel_handshake_text(&plugin.manifest, &signature) {
+            block.push_str(&handshake);
+            block.push('\n');
+        }
+        if let Some(notice) = plugin.panel_grant.notice(name) {
+            block.push('\n');
+            block.push_str(&notice);
+        }
+        if !block.is_empty() {
+            blocks.push(block);
+        }
+    }
+    blocks
+}
+
 /// Spawn the exchange for one frame request and answer the deck with exactly
 /// one panel envelope.
 ///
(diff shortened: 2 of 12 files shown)


macanderson:
Correcting one row of Sourcery's assessment table before a reviewer takes it as read.

Its first objective is marked met and describes the handshake as "rendered in the transcript on the silver rail". The transcript half is true; **the silver rail is not**, and I would rather say so than let a green tick stand for it.

The rail metal is a property of `EventKind` (`crates/stella-tui/src/views/transcript.rs`, `EventKind::metal`), and every kind in that enum is derived from a tool call — read muted, edit/write/run/gate gold, delete red, skill/memory silver, model gold-bright. Host prose has no kind of its own: it reaches the transcript as a fabricated `AgentEvent::Text` through `command_deck.rs`'s `chrome_note`, carrying `stella_tui::NOTICE_MARKER` and no rail at all. So the handshake renders as though the agent said it — which is the wrong direction for SPEC 12.3's own argument, since the one thing the transcript must be able to distinguish is Stella's chrome from a third party's words.

Giving host consent prose a rail of its own is a `stella-tui` transcript-vocabulary change with golden frames behind it. That is #5300, and the PR description now states the gap in the same place it explains the two other points where it follows the SPEC over the issue text.

Everything else in the table stands as written: the deny path is witnessed on the panel process's marker file rather than on a route list, and the grant is keyed to the exact manifest digest, scope and name.

macanderson:
Ran the whole grant lifecycle against a real `stella` binary and the shipped reference plugin, so the surface is demonstrated rather than only asserted by tests. Fixture: `plugins/stella-hello` copied to a temp source, installed project-scope into a temp workspace with `STELLA_HOME` and `STELLA_TRUST_PROJECT` moved.

`stella plugin install … --yes` prints the install consent document, then the panel handshake underneath it, then records the answer:

```
installed `hello` (project) into /tmp/panel-e2e/ws/.stella/plugins/hello

◳ panel handshake · hello
`hello` is asking to draw part of your screen. Stella draws the border and the label `◳ panel · hello` around it …

Manifest signature: sha256:6ec911678525e6f30593a5a16438c483d81ddd9464f77e6b1db724e465a50d07
  This grant covers exactly these manifest bytes. Change them and the panel is withheld until it is granted again.

Limits this panel accepts:
  - may not reach the network from its panel process
  - may not write anywhere but the sandbox directory Stella hands it

Where it draws:
  - draws a pane inside your SETTINGS tab
  - draws a bordered block in your transcript
  - adds a popup you open by typing its name
  - answers to `/hello`, and to `/hello:hello`
  - runs as a process on your machine: `python3 ${plugin_dir}/panel.py` (killed after 2s)
      it inherits NO environment variables

[a]llow  [d]eny
Nothing is leased a rectangle, and no panel process is started, until you allow it.

(accepted with --yes)
  panel: allowed for `hello`
```

Then, in order: `plugin list` → `panel: allowed — \`stella plugin panel hello\` changes it`; `plugin panel hello --deny` → `\`hello\` may not draw a panel. Its tools, skills and hooks are untouched`; `plugin list` → `panel: ! … you denied it. Run \`stella plugin panel hello\` …`.

**The drift arm, which is the one the tests reach least directly.** Appending a byte to the installed `plugin.toml` makes the *install receipt* fail first and withholds the whole package, so the panel-grant drift is shadowed for a receipted install — correct, and worth knowing. It is reachable for an **unreceipted project package**, which is what a `git clone` produces and what the receipt's own tier carve-out exists for: moving `.consent` aside, the same drifted manifest loads and reports

```
panel: ! `hello` declares a panel and is not drawing one — its `plugin.toml` has changed since
the panel was granted, so what it would draw with is not what anyone read. Run
`stella plugin panel hello` to read the handshake and decide.
```

`stella plugin panel hello --allow` then re-grants against the new bytes and `list` reads `allowed` again.

Also confirmed the acceptance artifact for this whole surface still works: `cargo test -p stella-tui --test hello_panel_demo -- --nocapture` draws the plugin's frame inside the host's `◳ panel · hello` chrome, unchanged.

Issue #4262, opened 2026-08-22
context: a cardinality claim gets a path_exists probe, so its green can never distinguish the claim from its opposite


macanderson:
## Problem

`stella ingest`'s truth probe can attach a `path_exists` probe to a claim that is not about a path existing. The probe then goes green on every run while being incapable of distinguishing the claim from its opposite — the exact "evidence consistent with both answers" failure CLAUDE.md names as not-evidence.

## Evidence

Two proposals in this workspace's review queue:

```
crates-directory-layout-77e1a582  85% confidence
  Every one of the twenty-one crates lives under the crates/ directory...
  probe: supported  path `crates/stella-core` exists.

inv8-parity-two-axes-57752ee0  85% confidence
  Invariant 8: provider feature parity is declared in
  crates/stella-model/src/provider_parity.rs on two axes today,
  CachePosture and ReasoningPosture.
  probe: supported  path `crates/stella-model/src/provider_parity.rs` exists.
```

Both claims are false against the current tree:

```console
$ fd -t d -d 1 . crates/ | wc -l
26                                    # not twenty-one

$ rg -o 'pub enum (\w*Posture)' crates/stella-model/src/provider_parity.rs
CachePosture
ReasoningPosture
OverflowPosture
OutputBudgetPosture
StreamFallbackPosture                 # plus ParallelToolCallPosture — six, not two
```

Both probes are green, and would have stayed green through every one of those additions. `crates/stella-core` existing says nothing about how many crates there are; `provider_parity.rs` existing says nothing about how many axes it declares. Neither probe can ever refute its claim, so its `supported` carries no information — but it renders in the same column, in the same colour, as a probe that genuinely discriminates.

This is distinct from #4261 (a probe whose recorded verdict is stale). These two probes are *freshly* wrong in kind: re-running them today still returns `supported`.

## Why it matters

A cardinality claim is exactly the kind that rots — this repository already refuses to write a gate-step total or a crate count into prose for that reason (AGENTS.md, "Deliberately no total", and the `god-files` guard's "the ceilings stay in the baseline alone, because a number in two places is how the last limit died"). A context record freezing "twenty-one crates" into the system prompt reintroduces the drift the prose was careful to avoid, and the probe that was supposed to catch it endorses it.

## Suggested direction (not prescriptive)

Either of these would resolve it; the second is cheaper:

[...]

PR #5315, opened 2026-08-27
fix(stella-core): four correctness defects in the learning and steering decision modules

### LINKED ISSUE #5344: Four correctness defects in the learning and steering decision modules
## Problem

A full correctness review of the learning/steering decision modules (`stella-core` skills, self_tuning, ingest gate; `stella-cli` skill discovery) found four defects, each confirmed against the code:

1. `states_a_cardinality`'s lookahead window stops one token short of the noun, so the gate misses its own documented example ("twenty-one Rust crates") and the count claim keeps its `path_exists` probe — #4262's false green.
2. The `auto_created_bonus` is included in the `min_score` threshold, giving freshly-mined skills a lower selection floor than hand-authored ones — the opposite of its documented tie-break-only contract.
3. `select_winner` treats a NaN mean as `Equal`, and `max_by` keeps the later element on `Equal` — a NaN arm in the middle of three hands the scan to whatever follows it.
4. `FsSkillSource` loads skill files in `read_dir` enumeration order, so same-root name collisions resolve per machine and the rendered prompt bytes depend on it (against byte-stable prompts, invariant #7).

Fixed by PR #5315.

## Definition of done

- [x] The gate catches spelled compound cardinals — `a_spelled_compound_cardinal_loses_its_path_probe`, verified red on the old code
- [x] The selection floor reads lexical + domain only — `the_auto_created_bonus_cannot_lift_a_skill_over_the_floor`, verified red on the old code
- [x] NaN loses the winner scan to every finite mean — `a_nan_mean_cannot_corrupt_the_winner_scan`, verified red on the old code
- [x] Skill files sort per root before the merge; existing precedence tests green (no practical witness for `read_dir` order — explained in the PR)
- [x] `cargo test -p stella-core skills` / `self_tuning` / `ingest` green; clippy green on both crates

macanderson:
## What this is

Four confirmed correctness defects in the learning/steering decision modules — skill selection, self-tuning winner selection, the reflection-ingest cardinality gate, and skill-file discovery — found by a full read of those modules. Each behavior change carries a witness test verified red on the old code and green on the new.

## The defects

1. **The cardinality gate missed its own documented example** (`stella-core/src/ingest/gate.rs`, `states_a_cardinality`). The lookahead counts *intervening* tokens, but the window was `take(CARDINAL_LOOKAHEAD)` — one short of where the noun may sit. From `twenty` in "twenty-one Rust crates" the window was `[one, rust]` and `crates` was never reached, so the spelled-compound count claim kept its `path_exists` probe and went green on every run — #4262's false-green shape, which this gate exists to stop. Digit forms were unaffected.
   Witness: `a_spelled_compound_cardinal_loses_its_path_probe` (also pins the digit form, the deliberate `one` exemption, and a cardinal with no plural noun in reach).

2. **The auto-created bonus breached the selection floor** (`stella-core/src/skills.rs`, `select_skills_reporting`). `auto_created_bonus` documents itself as a tie-break "never large enough to select a skill that is otherwise below `min_score`" — but the threshold compared the full score including the bonus, giving freshly-mined (least-vetted) skills a *lower* selection floor than hand-authored ones with identical wording. The floor now reads the evidence-bearing half (lexical + domain); the bonus still orders survivors.
   Witness: `the_auto_created_bonus_cannot_lift_a_skill_over_the_floor`.

3. **A NaN mean corrupted the winner scan** (`stella-core/src/self_tuning.rs`, `select_winner`). NaN was treated as `Equal`, and `max_by` keeps the later element on `Equal` — so means `[0.9, NaN, 0.5]` walked to the 0.5 arm as "winner". Downstream guards kept a promotion from firing, but the decision named the wrong reason, and the doc's "a NaN mean can never win by accident" held only for the NaN arm itself. NaN now loses to every finite mean. All live call sites pass two arms today; the third-arm shape is where this bit.
   Witness: `a_nan_mean_cannot_corrupt_the_winner_scan`.

4. **Skill files loaded in filesystem enumeration order** (`stella-cli/src/memory/skill_files.rs`, `FsSkillSource`). The loader's same-name merge is last-writer-wins, so two colliding entries in one root (flat `foo.md` beside `foo/SKILL.md`, or a frontmatter `name:` matching a sibling's slug) picked their winner per machine — and the rendered prompt bytes depended on it, against the byte-stable-prompts invariant (#7). Entries now sort per root; root order still decides cross-root precedence.
   No practical witness: the defect is `read_dir`'s platform-dependent enumeration order, which a test cannot force on a filesystem that happens to enumerate sorted. Verified by inspection that the sort is the only change and the merge semantics are untouched; the loader's existing precedence tests still pass.

## Witness verification (the artisanal way)

With the three behavior lines reverted on top of this branch:

```
cargo test -p stella-core a_spelled_compound            -> red
cargo test -p stella-core the_auto_created_bonus_cannot -> red
cargo test -p stella-core a_nan_mean_cannot             -> red
```

Restored: `cargo test -p stella-core skills` (82 passed), `self_tuning` (16 passed), `ingest` (68 passed); `cargo clippy -p stella-core -p stella-cli --all-targets -- -D warnings` green.

## Not in this PR

The same review surfaced items needing design decisions or their own PRs — the `'/'` path-string contract skill discovery leans on, the Welch-z calibration claim at the sample floor, and description minting for learned skills. Filed as issues from this review session.

## Summary by Sourcery

Fix four correctness defects in learning, steering, and skill-file decision behavior.

Bug Fixes:
- Correct cardinality-gate lookahead so spelled compound counts can no longer retain inappropriate path probes.
- Prevent auto-created skill bonuses from selecting skills below the evidence threshold.
- Ensure NaN means cannot corrupt self-tuning winner selection.
- Make skill-file loading deterministic within each root to stabilize same-name merge results.

Tests:
- Add regression coverage for cardinality boundaries, skill-selection floors, and NaN winner handling.

Closes #5344

### DIFF
--- a/crates/stella-cli/src/memory/skill_files.rs
+++ b/crates/stella-cli/src/memory/skill_files.rs
@@ -21,8 +21,9 @@ impl SkillSource for FsSkillSource {
                 // same-name entries in one root (flat `foo.md` beside
                 // `foo/SKILL.md`, or a frontmatter `name:` colliding with a
                 // sibling's slug) picked their winner per machine. The
-                // rendered prompt bytes must not depend on that (invariant
-                // #7); root order still decides cross-root precedence.
+                // rendered prompt bytes must not depend on that (AGENTS.md
+                // rule 7, byte-stable prompts); root order still decides
+                // cross-root precedence.
                 let mut paths: Vec<std::path::PathBuf> =
                     entries.flatten().map(|entry| entry.path()).collect();
                 paths.sort();
--- a/crates/stella-cli/src/memory/skill_files.rs
+++ b/crates/stella-cli/src/memory/skill_files.rs
@@ -16,8 +16,17 @@ impl SkillSource for FsSkillSource {
         for root in roots {
             // Flat layout: <root>/<slug>.md
             if let Ok(entries) = std::fs::read_dir(root) {
-                for entry in entries.flatten() {
-                    let path = entry.path();
+                // Sorted per root: `read_dir` order is filesystem enumeration
+                // order, and the loader's merge is last-writer-wins — so two
+                // same-name entries in one root (flat `foo.md` beside
+                // `foo/SKILL.md`, or a frontmatter `name:` colliding with a
+                // sibling's slug) picked their winner per machine. The
+                // rendered prompt bytes must not depend on that (invariant
+                // #7); root order still decides cross-root precedence.
+                let mut paths: Vec<std::path::PathBuf> =
+                    entries.flatten().map(|entry| entry.path()).collect();
+                paths.sort();
+                for path in paths {
                     if path.extension().is_some_and(|e| e == "md") {
                         if let Ok(content) = std::fs::read_to_string(&path) {
                             files.push(skills::SkillFile {
--- a/crates/stella-core/src/ingest/gate.rs
+++ b/crates/stella-core/src/ingest/gate.rs
@@ -284,8 +284,8 @@ mod tests {
     /// tokenizes to `[twenty, one, rust, crates]`, so from `twenty` the noun
     /// sits three tokens out — a window of `CARDINAL_LOOKAHEAD` alone stopped
     /// one short, and the spelled-compound count claim kept its `path_exists`
-    /// probe (#4262's false green). Digit forms and the deliberate `one`
-    /// exemption pin the boundary from the other side.
+    /// probe (#4262's false green). Digit forms and the `one` exemption pin
+    /// the boundary from the other side.
     #[test]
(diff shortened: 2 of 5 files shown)


macanderson:
Follow-ups from the same review, filed rather than fixed here: #5320 (slash-path contract, Windows clobber), #5321 (Welch-z calibration claim at the sample floor), #5334 (skills test gaps), #5335 (learned-skill descriptions selection cannot match), #5336 (count distinct tasks before minting a skill). The DoD issue this PR closes is #5344.

PR #5387, opened 2026-08-28
fix(bench): the frozen replay analyzer supplies its own posture schema

### LINKED ISSUE #5108: The frozen replay analyzer only loads when another test file put its directory on sys.path first
## Problem

Two tests in `bench/harbor_adapter/tests/test_secure_launcher.py` pass when the
whole suite runs and fail when that file runs alone:

```
$ cd bench/harbor_adapter && uv run pytest tests/test_secure_launcher.py -q
FAILED tests/test_secure_launcher.py::test_missing_prior_readiness_job_never_reserves_or_execs
FAILED tests/test_secure_launcher.py::test_forged_prior_readiness_artifacts_never_reserve_or_exec
2 failed, 145 passed

$ cd bench/harbor_adapter && uv run pytest -q
763 passed, 2 skipped
```

Both assert a specific refusal and get a generic one instead:

```
Expected regex: 'prior Harbor job is missing'
Actual message: 'prior-stage replay analyzer could not be loaded'
```

## Cause

`secure_launcher.py`'s `_load_frozen_replay_analyzer` execs
`bench/terminal_bench_analysis/tb21_analysis.py` from verified source bytes into
a bare `ModuleType` with `__package__ = ""` and adds nothing to `sys.path`. That
file's line 29 does `from tb21_posture_schema import (...)`, a sibling module in
`bench/terminal_bench_analysis/`, so the import resolves only when something
else has already put that directory on `sys.path`. Running the full suite does;
running the one file does not.

Reproduced directly, with no pytest involved:

```
$ cd bench/harbor_adapter && uv run python -c "
import pathlib, sys
from types import ModuleType
p = pathlib.Path('../terminal_bench_analysis/tb21_analysis.py')
t = pathlib.Path('../terminal_bench_analysis/github_public_timing.py')
tm = ModuleType('github_public_timing'); tm.__file__=str(t); tm.__package__=''
sys.modules['github_public_timing']=tm
exec(compile(t.read_bytes(), str(t), 'exec'), tm.__dict__)
am = ModuleType('x'); am.__file__=str(p); am.__package__=''
exec(compile(p.read_bytes(), str(p), 'exec'), am.__dict__)
"
ModuleNotFoundError: No module named 'tb21_posture_schema'
```

## Why it matters more than a flaky test

The two failing tests are not the finding; they are the symptom that made it
visible. `_replay_prior_stage_evidence` is the check that refuses a paid stage
whose prior stage's artifacts do not replay. If the analyzer cannot load, that
check collapses to one generic refusal — and the specific refusals it exists to
make (a missing prior job, forged prior artifacts) are never reached and never
distinguished.

Whether it loads at all depends on whatever else has touched `sys.path` in the
process. A launcher whose evidence check is contingent on import order is a
launcher whose refusal you cannot predict, and this is the exact shape CLAUDE.md
warns about under "a `WITNESS CONFIRMED` was a false proof off a
`ModuleNotFoundError`".

Note the failure mode is fail-closed today — the launcher refuses rather than
proceeds — so this is not a live claim-eligibility hole. It is a check that
cannot state which of several things went wrong.

## Files

- `bench/harbor_adapter/stella_harbor/secure_launcher.py` —
  `_load_frozen_replay_analyzer` builds the two modules and execs them.
- `bench/terminal_bench_analysis/tb21_analysis.py` — the sibling import at
  line 29 (`from tb21_posture_schema import ...`).
- `bench/terminal_bench_analysis/tb21_posture_schema.py` — the module that is
  not found.
- `bench/harbor_adapter/tests/test_secure_launcher.py` — the two tests.

## Done when

- `uv run pytest tests/test_secure_launcher.py` passes on its own, with no
  dependence on which other test files ran first.
- `_load_frozen_replay_analyzer` resolves `tb21_posture_schema` itself — either
  by putting the analyzer's own directory on `sys.path` for the exec, or by
  freezing and verifying that module's bytes alongside the other two, the way
  `github_public_timing` already is. The second keeps the "exact verified source
  bytes are the replay authority" property that comment claims; the first does
  not, since an unverified sibling would then be executed.
- A test covers the isolated case, so import-order luck cannot hide it again.

Found while splitting `secure_launcher.py` for #4759; the control run confirmed
both failures are present on `origin/main` at `faad82ab2` and are not caused by
that change.

Refs #4759

## Definition of done

- [x] `uv run pytest tests/test_secure_launcher.py` passes on its own, with no dependence on which other test files ran first.
- [x] `_load_frozen_replay_analyzer` resolves `tb21_posture_schema` itself — by putting the analyzer's own directory on `sys.path` for the exec, or by freezing and verifying that module's bytes alongside the other two.

---

Both boxes ticked against [#5387](https://github.com/macanderson/stella/pull/5387), verified rather than asserted:

```
$ cd bench/harbor_adapter && uv run pytest tests/test_secure_launcher.py -q
147 passed
$ uv run pytest -q
766 passed, 2 skipped
```

The loader takes the second remedy — `tb21_posture_schema` is read, hash-checked and exec'd
as its own module, the way `github_public_timing` already is — rather than the `sys.path`
one, so no unverified sibling is ever executed. Its identity is pinned in
`secure_launcher.py` (covered by `adapter_sha256`) rather than in the study manifest, which
is preregistered at v6 and would refuse every already-recorded manifest if a required field
were added. The path also joins the frozen set, so the public-tree check compares its bytes.

The third DoD line — a test covering the isolated case — is
`test_frozen_replay_analyzer_loads_without_a_prepared_sys_path`, in a new
`tests/test_frozen_replay_analyzer.py` (its own file because `test_secure_launcher.py` sits
at its `file-size` ceiling).

macanderson:
## What & why

`_load_frozen_replay_analyzer` execs `tb21_analysis.py` from verified source bytes, and that file imports `tb21_posture_schema` at module scope. Nothing here supplied that module or put its directory on `sys.path`, so the import resolved only when some earlier importer happened to have done it.

Running the whole `harbor_adapter` suite did. Running `tests/test_secure_launcher.py` alone did not, and its two refusal tests got a generic `prior-stage replay analyzer could not be loaded` in place of the specific refusal they assert:

```
$ cd bench/harbor_adapter && uv run pytest tests/test_secure_launcher.py -q
FAILED tests/test_secure_launcher.py::test_missing_prior_readiness_job_never_reserves_or_execs
FAILED tests/test_secure_launcher.py::test_forged_prior_readiness_artifacts_never_reserve_or_exec
2 failed, 145 passed
```

A launcher whose refusal reason depends on import order cannot be reasoned about, and CLAUDE.md names this exact shape — a proof that came off a `ModuleNotFoundError`.

### Why not `sys.path`

The issue offers either remedy. This takes the second: the schema is supplied the way `github_public_timing` already is. A `sys.path` entry would let *any* importable `tb21_posture_schema` execute here, and executing unverified bytes is what this function exists to prevent.

### Where the pin lives, and why not the manifest

The schema's identity is pinned in `secure_launcher.py` rather than added to `runtime_identity`. The study manifest is preregistered at a fixed version (`stella-tb21-study-manifest-v6`), so a new required field would refuse every manifest already recorded — it would break replay of the runs this machinery exists to replay. `secure_launcher.py` is in `_FIXED_ADAPTER_SOURCE_PATHS`, so the pin is already covered by `adapter_sha256` and byte-compared against the published commit: the same chain that authorises the analyzer. The schema path also joins the frozen set, so the public-tree check compares its bytes too.

### One thing this cleaned up on the way

The fake published tree in `test_secure_launcher.py` re-listed the frozen non-adapter paths by hand. Adding one path reddened 30 tests that were each correct about a list that no longer matched, so that list is now read off the launcher (`_FIXED_NON_ADAPTER_SOURCE_PATHS`) instead of copied.

## The witness

Each of these fails with the schema exec backed out; the first two also fail on `origin/main` when the file is run alone:

- `test_missing_prior_readiness_job_never_reserves_or_execs` *(pre-existing)*
- `test_forged_prior_readiness_artifacts_never_reserve_or_exec` *(pre-existing)*
- `test_frozen_replay_analyzer_loads_without_a_prepared_sys_path`
- `test_posture_schema_pin_matches_the_frozen_file`
- `test_a_drifted_posture_schema_refuses_by_name`

```
$ cd bench/harbor_adapter && uv run pytest tests/test_secure_launcher.py -q
147 passed
$ uv run pytest tests/test_frozen_replay_analyzer.py -q
3 passed
$ uv run pytest -q
766 passed, 2 skipped
```

The new tests are in their own file because `test_secure_launcher.py` is grandfathered at its size ceiling — `make file-size` caught them pushing it +99 over, and raising a baseline to fit three tests is the expedient CLAUDE.md forbids. The launcher file is back to exactly its recorded 3368.

## The gate

- [x] `make guards-fast` green (this diff compiles no Rust; `file-size`, `no-secrets`, `shellcheck`, `lockfile-sync`, `fmt --check` and the rest pass)
- [ ] `cargo` steps — not reached by this diff; CI runs them
- [x] Docs updated where behavior changed — the constants and the loader carry their own reasoning
- [x] `Closes #5108` appears both here and as a commit trailer

## Nothing left behind

- [x] There is nothing new: everything I noticed here is fixed in this PR.

One observation that is **not** mine and not new: `bench/terminal_bench_analysis`'s own pytest suite fails collection locally with `ModuleNotFoundError: No module named 'tb21_analysis'` on all 7 test files. That is unchanged by this PR — it touches no file in that directory — and CI runs that suite behind `uv sync --locked` in `bench.yml`, which this local invocation did not do. Flagging it rather than filing, since I could not establish it is a defect rather than my invocation.

## Ground-rule check

The pin is a literal, which means it can drift from the file. `test_posture_schema_pin_matches_the_frozen_file` is what catches that — without it, a schema edit would fail at replay time, hours into a benchmark, instead of in CI. That is the cost of pinning rather than deriving, and deriving is not an option: a hash the loader computes from the file it is about to run proves nothing.

Closes #5108

## Summary by Sourcery

Supply and verify the frozen posture schema when loading the replay analyzer to make secure replay deterministic and self-contained.

Bug Fixes:
- Load the frozen posture schema from verified source bytes so replay analyzer behavior and refusal reasons no longer depend on import order or a prepared `sys.path`.
- Reject drifted posture schema bytes with a schema-specific error before replay execution.

Enhancements:
- Centralize the frozen non-adapter source path list and reuse it in public-tree verification and test fixtures.

Tests:
- Add standalone coverage for schema pin integrity, isolated analyzer loading, and schema drift refusal.

### DIFF
--- a/bench/harbor_adapter/stella_harbor/secure_launcher.py
+++ b/bench/harbor_adapter/stella_harbor/secure_launcher.py
@@ -110,6 +110,28 @@ _FIXED_LEDGER_PATH = "bench/evidence/stella-tb21-run-ledger.json"
 _FIXED_MANIFEST_PATH = "bench/evidence/stella-tb21-study-manifest.json"
 _FIXED_ANALYZER_PATH = "bench/terminal_bench_analysis/tb21_analysis.py"
 _FIXED_PUBLIC_TIMING_PATH = "bench/terminal_bench_analysis/github_public_timing.py"
+# The analyzer's sibling, split out of it by #2549 as "a size boundary, not a
+# new address" — `tb21_analysis` imports it at module scope and re-exports its
+# names. It joins the frozen set for the same reason the analyzer is in it: its
+# bytes decide what the replay computes.
+#
+# Without this the loader execed the analyzer with nothing supplying
+# `tb21_posture_schema`, so the import at the top of that file resolved only
+# when some earlier importer had already put the analyzer's directory on
+# `sys.path`. Running the launcher's own test file alone did not, and the two
+# refusal tests there got "replay analyzer could not be loaded" instead of the
+# specific refusal they assert — a launcher whose refusal reason depends on
+# import-order luck (#5108).
+_FIXED_POSTURE_SCHEMA_PATH = "bench/terminal_bench_analysis/tb21_posture_schema.py"
+# Pinned here rather than derived, because a hash the loader computes from the
+# file it is about to run proves nothing. This literal lives in a file that is
+# itself in `_FIXED_ADAPTER_SOURCE_PATHS`, so it is covered by `adapter_sha256`
+# and byte-compared against the published commit — the same chain that already
+# authorises the analyzer. `test_posture_schema_pin_matches_the_frozen_file`
+# fails the moment the two drift.
+_FIXED_POSTURE_SCHEMA_SHA256 = (
+    "be6c8ac5bd142aece32b73f4565b422db73eb7fc366c66482ca61de5091f08d3"
+)
 _FIXED_PROTOCOL_PATH = "bench/terminal-bench-2.1-protocol.md"
 # Every Python source in the adapter package, enumerated rather than globbed:
 # the public-tree check compares this exact set against the published commit,
@@ -154,6 +176,21 @@ _FIXED_READINESS_SOURCE_PATHS = (
     "bench/readiness/synthetic-adapter-sentinel/instruction.md",
     "bench/readiness/synthetic-adapter-sentinel/task.toml",
     "bench/readiness/synthetic-adapter-sentinel/tests/test.sh",
+)
+# The frozen sources outside the adapter package: byte-compared against the
+# published commit like the adapter's own, but not folded into `adapter_sha256`.
+#
+# A name rather than a tuple built at the one call site, so the tests can drive
+# the check against the same list it enforces. The fake published tree in
+# `test_secure_launcher.py` used to re-list these by hand, and adding
+# `_FIXED_POSTURE_SCHEMA_PATH` reddened thirty tests that were each correct
+# about a list that no longer matched.
+_FIXED_NON_ADAPTER_SOURCE_PATHS = (
+    _FIXED_ANALYZER_PATH,
+    _FIXED_PUBLIC_TIMING_PATH,
+    _FIXED_POSTURE_SCHEMA_PATH,
+    _FIXED_PROTOCOL_PATH,
+    *_FIXED_READINESS_SOURCE_PATHS,
 )
 _RUN_LEDGER_SCHEMA = "stella-tb21-run-ledger-v2"
 _GITHUB_ATTESTATION_SCHEMA = "stella-tb21-github-attestation-v2"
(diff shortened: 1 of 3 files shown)


macanderson:
Both Definition-of-done items on #5108 are now ticked, against verified runs rather than assertion:

```
$ cd bench/harbor_adapter && uv run pytest tests/test_secure_launcher.py -q
147 passed
$ uv run pytest -q
766 passed, 2 skipped
```

Re-running `dod`.

PR #5398, opened 2026-08-28
fix(stella-cli): a reflection lesson whose append failed is re-scanned, not lost

### LINKED ISSUE #5323: A transient ledger failure permanently loses reflection lessons on the typed path
## Problem

CONFIRMED by trace. `crates/stella-cli/src/memory/observations.rs`: the typed-lifecycle extraction loop swallows `append_observation` failures (a `.ok()?` — for example SQLITE_BUSY from a concurrent process past the 5s busy timeout), then unconditionally advances the cursor to `lines.len()`. Lines whose append failed are now below the cursor and never re-scanned. The module's own header claims "a crash costs one re-scan, never a duplicate" — but a *failure* (not a crash) costs the records outright. The lexical path re-reads the full log and is unaffected, so the two paths diverge exactly here, against the behavior-compatibility contract `crates/stella-cli/src/memory/learning.rs` holds them to.

## Definition of done

- [x] The per-line outcome is typed (`LineOutcome`): a replay or a deterministic refusal is consumed; only a transient store failure holds the cursor, at the first failed line — so a permanently malformed lesson can never wedge it
- [x] Later lines are still consumed in the same pass; the re-scan replays them as no-ops (content-derived ids)
- [x] Witness `a_failed_append_holds_the_cursor_at_the_failed_line` — verified red on the old code, green on the new; also pins the no-failure and deterministic-refusal cursor positions
- [x] `cargo test -p stella-cli observations` green

macanderson:
## What & why

The typed extraction loop swallowed a failed `append_observation` and then advanced the cursor to `lines.len()` regardless:

```rust
if append_observation(store, &lesson) == Some(AppendOutcome::Appended) {
    appended += 1;
}
// ...
let _ = store.set_extraction_cursor(REFLECTION_SOURCE, &lines.len().to_string());
```

So a line whose ledger write lost to `SQLITE_BUSY` — a concurrent writer past the 5s busy timeout — ended up *below* the cursor and was never looked at again. The module header claims "a crash costs one re-scan, never a duplicate", and that is true of a crash; a **failure** cost the records outright.

The lexical path re-reads the whole log every turn and is unaffected, so the two paths diverged exactly here, against the behaviour-compatibility contract `learning.rs` holds them to.

## The fix, and why it is two halves

The cursor now stops at the first line whose append failed. Everything from there is re-scanned next turn, which content-derived ids make a no-op replay for whatever did land.

A barrier alone would be a different bug. A line that can *never* append — one whose record does not construct, does not serialize, or whose content the ledger refuses (`InvalidInput`: two records claiming one id) — would park the cursor on itself forever, re-scanning a growing log on every turn without ever advancing. So the failure is typed:

| | classified | cursor |
|---|---|---|
| `ContextError::Sqlite` | the storage layer refusing a write | **stops here** — the same bytes may append next turn |
| everything else | the ledger judging fixed bytes | skipped, like a malformed JSON line |

## The witness

`a_lesson_whose_append_failed_is_recovered_on_the_next_scan` — a rival connection holds the write lock just past the store's 5s `busy_timeout` and then releases, which is the failure the issue names.

- **Base:** the first lesson's append fails and is swallowed, the loop carries on, the second lands once the lock clears, and the cursor is set to 2 — past a line that never appended. Ledger ends with one observation and stays there.
- **Here:** the loop stops at the first failure, nothing appends, the cursor does not move, and the next scan recovers both.

The discriminating assertion is the ledger's contents after a *second* scan — 2 here, 1 on the base, forever. The intermediate counts are asserted too so a lock that cleared early cannot make it pass for the wrong reason.

`an_unrepresentable_line_does_not_wedge_the_cursor` covers the other arm, so the fix cannot be satisfied by simply never advancing.

```
$ cargo test -p stella-cli --bin stella memory::observations
13 passed; 0 failed
```

## The gate

- [x] `cargo fmt --check`
- [x] Targeted tests green (SCR-001: the crate's own filter, not the workspace — CI runs the rest)
- [x] Docs updated — the module header's "failure isolation" section now says what happens to a failed append, replacing the claim that did not hold
- [x] `Closes #5323` appears both here and as a commit trailer

## Nothing left behind

The issue also suggests recording the failure in the ledger the way `record_reflection_parse_failure` made parse starvation observable. That is deliberately **not** done here: this path's contract is that it never fails or slows the user's turn, and the failure it would record is the ledger being unavailable — the same ledger the record would be written to. The recovery is now automatic and needs no operator, so the observability is worth less than the coupling would cost. Say so if you disagree and I will file it.

## Ground-rule check

The witness takes ~5.7s, because it has to outwait a real `busy_timeout` — the loss needs the *append* to fail while the *cursor write* succeeds, and a lock held for the whole scan fails both and leaves the two codes indistinguishable. A faster fake would not have reproduced the bug.

Closes #5323

## Summary by Sourcery

Ensure failed reflection appends are retried without sacrificing cursor progress for permanently invalid lessons.

Bug Fixes:
- Preserve retryable reflection lessons by stopping the extraction cursor at transient ledger write failures so they are recovered on the next scan.
- Continue advancing past deterministic, unrepresentable lessons to prevent permanently wedging the extraction cursor.

Enhancements:
- Classify observation append failures so transient storage errors are retried while deterministic record or content errors are skipped.
- Update reflection extraction documentation to describe failure recovery and cursor behavior.

Tests:
- Add coverage for recovering a lesson after a SQLite lock causes its append to fail.
- Add coverage ensuring unrepresentable lessons do not block cursor progress.

### DIFF
--- a/crates/stella-cli/src/memory/observations.rs
+++ b/crates/stella-cli/src/memory/observations.rs
@@ -27,6 +27,23 @@
 //! record in the batch or lose it. With derived ids, the cursor is a pure
 //! optimization and a crash costs one re-scan, never a duplicate.
 //!
+//! A **crash** costs a re-scan. A *failure* used to cost the records outright:
+//! the loop swallowed a failed append and advanced the cursor to the end of the
+//! log regardless, so a line whose write lost to `SQLITE_BUSY` was left below
+//! the cursor and never looked at again. The lexical path re-reads the whole log
+//! every turn and is unaffected, so the two diverged exactly there — against the
+//! behaviour-compatibility contract [`super::learning`] holds them to (#5323).
+//!
+//! So the cursor stops at the first line whose append failed, and everything
+//! from there is re-scanned next turn. Content-derived ids make that replay a
+//! no-op for the lines that did land.
+//!
+//! A line the ledger refuses **deterministically** — one whose record cannot be
+//! constructed or serialized at all — is not a barrier. Retrying it can never
+//! succeed, and a cursor parked on it would re-scan the whole log every turn
+//! forever without advancing. Those are skipped the way a malformed JSON line
+//! already is.
+//!
 //! ## Failure isolation
 //!
 //! Every function here swallows its errors and degrades to "no observations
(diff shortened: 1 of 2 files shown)


macanderson:
Superseded by [#5383](https://github.com/macanderson/stella/pull/5383), which merged at 04:59 and closes #5323. It reaches the same design — the cursor advances to the first line whose ledger write failed and never past it — with a cleaner factoring (a `Consumed` verdict plus a helper returning `(appended, cursor)`) than the inline `settled` this branch carried.

That PR did not exist when this one was opened; no disagreement, just a race. Nothing here is worth salvaging on top of what landed.

PR #5401, opened 2026-08-28
fix(stella-cli): close the three doors around governance-mode enforcement

### LINKED ISSUE #5328: Governance-mode enforcement has three doors around it
## Problem

CONFIRMED by trace, `crates/stella-cli`:

1. `read_governance` (`src/context_records.rs`) fails OPEN: an unparseable `governance.toml` silently becomes Solo with no separation, switching off validate's regulated check and promote's separation check after a one-character corruption — while a broken `promotions.jsonl` fails loudly. The strictest-tier selector is the least-defended artifact in the directory.
2. `stella context keep`'s supersession path (`src/context_cmd/review.rs`) appends a Superseded ledger event with `proposer: None` and no separation check — the separation gate exists only in `run_promote` (`src/context_cmd/govern.rs`, which is correctly fail-closed).
3. `run_govern` rebuilds `Governance` from the flags: `stella context govern regulated` without `--separation` silently turns an existing separation OFF, and `--separation` with no mode argument is discarded by the early return.

## Definition of done

- [x] Door 1: `read_governance` fails closed on a present-but-broken `governance.toml` (typed message naming the repair); an absent file stays solo defaults — witness `a_corrupt_governance_file_fails_closed_and_an_absent_one_stays_solo`
- [x] Door 2: one shared gate (`separation_checked_proposer`) runs on every lifecycle event that clears a blocking grant — `context keep`'s supersession, `proposals retract`, and `ingest --refresh` — refusing the author's own disarm under regulated+separation, failing closed on an unestablishable author, and recording the proposer on the event; advisory records pass untouched — witness `separation_refuses_the_authors_own_disarm_of_a_blocking_grant`
- [x] Door 3: `govern` merges with the existing document — omitted `--separation` keeps the current setting, `--separation false` turns it off on the record, and a bare `--separation` updates the current mode instead of being discarded — witness `govern_keeps_separation_unless_told_otherwise`
- [x] All witnesses verified red on the old code; `cargo test -p stella-cli context` green (119 passed); clippy and `make command-docs` green

macanderson:
## What & why

Governance decides who may grant enforcement. Three separate paths went around that decision.

### 1. `read_governance` failed open

```rust
std::fs::read_to_string(&path).ok().and_then(|t| toml::from_str(&t).ok()).unwrap_or_default()
```

An unparseable `governance.toml` became `Solo` — the least governed tier — so **one corrupt character switched off `validate`'s regulated check and `promote`'s separation check**, and nothing downstream could tell the result from a repository that had chosen solo. The strictest-tier selector was the least defended artifact in the directory, while a broken `promotions.jsonl` beside it failed loudly.

It now returns `Result`. **Absent is still the solo default** (§5.1); *unreadable* is an error. The five call sites all return `Result<(), String>` already, so they propagate with `?`.

### 2. `keep`'s supersession path had no separation check

It wrote a ledger event with `proposer: None` and ran no check, while the same check sat inside `run_promote`. The latest-event fold **drops any blocking grant the lineage held**, so superseding an armed record revokes enforcement — which is the act separation governs.

The check is now `govern::separation_cleared`, shared by both callers. Applied here only when the lineage actually holds a grant: `promote` guards the transition *into* blocking, this guards the one that silently undoes it, and ordinary editing of an advisory record stays possible in a regulated repository.

### 3. `run_govern` rebuilt the document from the flags

`separation` was a bare `bool`, so "not passed" and "passed false" were the same request:

- `stella context govern regulated` silently turned an existing separation **off**.
- `--separation` with no mode was discarded by the `let Some(mode) = mode else { return show }` early return — the command printed the current policy and exited 0, so it looked like it had worked.

Now `Option<bool>`, merged with the recorded document rather than rebuilt. Each argument changes only what it names. The confirmation prompt reports what the change does to separation rather than echoing the flag — a prompt that echoed the flag said nothing about separation exactly when the caller most needed to be told it was unchanged.

## The witness

Each was verified failing with its own fix backed out:

| test | door |
|---|---|
| `an_unparseable_governance_file_is_refused_rather_than_read_as_solo` | 1 |
| `an_absent_governance_file_is_still_the_solo_default` | 1 — so "fail closed" cannot be satisfied by refusing every repository that never ran `govern` |
| `a_well_formed_governance_file_is_read_as_written` | 1 — nor by refusing everything |
| `changing_the_mode_leaves_separation_as_the_repository_recorded_it` | 3 |
| `separation_alone_changes_separation_and_leaves_the_mode` | 3 |
| `neither_argument_leaves_the_policy_exactly_as_it_was` | 3 — so the two above cannot be satisfied by making every invocation a write |

```
$ cargo test -p stella-cli --bin stella context_
79 passed; 0 failed
$ cargo clippy -p stella-cli --all-targets -- -D warnings
clean
```

## The gate

- [x] `cargo fmt --check`
- [x] Docs updated: `website/content/docs/commands/context.mdx`'s `govern` section carries the new flag shape, the merge semantics, and the unreadable-policy refusal
- [x] `Closes #5328` appears both here and as a commit trailer

## Nothing left behind

- [x] There is nothing: all three doors are closed in this PR.

**One thing I want a reviewer's eye on rather than a filed issue:** door 2 is scoped to a lineage that *currently holds a blocking grant*. The alternative reading of the issue — run the check on every supersession — would make editing any published record in a regulated repo require a second identity, which I judged to be a bigger behavioural change than the issue asks for and not obviously wanted. Say the word and I will widen it.

## Ground-rule check

Door 1 is a behaviour change with a blast radius: a repository whose `governance.toml` is currently unparseable has been running as `solo` without knowing it, and after this it gets an error from `context validate`, `promote`, `keep`, `retract` and `ingest --refresh` until the file is repaired or deleted. That is the point — but it will look like a new failure rather than a disclosed old one, so the message says which file, why it is refused, and that deleting it means solo.

Closes #5328

## Summary by Sourcery

Close governance enforcement bypasses by failing safely on unreadable policies, applying separation checks to grant-clearing lifecycle actions, and merging only explicitly requested policy changes.

Bug Fixes:
- Reject unreadable governance policies instead of silently treating them as the least governed mode while preserving the solo default for absent policies.
- Enforce proposer/approver separation across promotion, supersession, retraction, and refresh retirement when a lineage holds a blocking grant.
- Preserve unspecified governance settings when updating policy and allow separation to be changed independently of the governance mode.

Enhancements:
- Centralize governance separation checks and improve refusal and confirmation messaging.

Documentation:
- Document governance update merge semantics, separation flag values, lifecycle enforcement checks, and refusal of unreadable governance policies.

Tests:
- Add witness coverage for unreadable, absent, and valid governance policies, independent governance updates, and lifecycle separation enforcement.

### DIFF
--- a/crates/stella-cli/src/context_cmd.rs
+++ b/crates/stella-cli/src/context_cmd.rs
@@ -43,7 +43,11 @@ use stella_core::ingest::record::{ContextFile, Proposal};
 
 mod amend;
 mod explain;
-mod govern;
+// `pub(crate)` rather than private: the separation gate is one rule with three
+// callers, and two of them (`proposals retract`, `ingest --refresh`) live
+// outside this module. A second copy of the check is exactly what #5328 was
+// filed about.
+pub(crate) mod govern;
 mod propose;
 mod review;
 // `pub(crate)` for one reason: the retirement handshake spans two commands, so its
--- a/crates/stella-cli/src/context_cmd.rs
+++ b/crates/stella-cli/src/context_cmd.rs
@@ -196,9 +196,14 @@ pub enum ContextCmd {
         /// New mode; omit to show current governance, policy version, and
         /// every ledger enforcement grant.
         mode: Option<String>,
-        /// Enable proposer/approver separation (regulated mode).
-        #[arg(long)]
-        separation: bool,
+        /// Enable proposer/approver separation (regulated mode). `--separation
+        /// false` turns it off; omitting it leaves it as it is.
+        ///
+        /// `Option<bool>` rather than a flag, because absent and `false` must
+        /// not be the same request: as a bare flag, `stella context govern
+        /// regulated` silently turned an existing separation OFF (#5328).
+        #[arg(long, num_args = 0..=1, default_missing_value = "true")]
+        separation: Option<bool>,
         /// Confirm a mode change without asking interactively.
         #[arg(long)]
         yes: bool,
--- a/crates/stella-cli/src/context_cmd/govern.rs
+++ b/crates/stella-cli/src/context_cmd/govern.rs
@@ -28,18 +28,28 @@ use super::review::actor;
 pub(crate) fn run_govern(
     root: &Path,
     mode: Option<&str>,
-    separation: bool,
+    separation: Option<bool>,
     yes: bool,
 ) -> Result<(), String> {
-    let current = read_governance(root);
-    let Some(mode) = mode else {
+    let current = read_governance(root)?;
+    // Nothing to change: show. `--separation` alone IS something to change, and
+    // used to be discarded by an early return here (#5328).
+    if mode.is_none() && separation.is_none() {
         return show_governance(root, &current);
-    };
-    let target = parse_mode(mode)?;
+    }
+    // Merged with the document rather than rebuilt from the flags. Every field
+    // the caller did not speak about keeps the value the repository recorded —
+    // `govern regulated` with no `--separation` used to silently turn an
+    // existing separation off, which is a governance change nobody asked for
+    // and nobody was told about (#5328).
     let next = Governance {
-        mode: target,
-        separation,
+        mode: match mode {
+            Some(mode) => parse_mode(mode)?,
+            None => current.mode,
+        },
+        separation: separation.unwrap_or(current.separation),
     };
+    let target = next.mode;
     if next == current {
         println!(
             "governance already {} (separation {})",
@@ -63,10 +73,15 @@ pub(crate) fn run_govern(
              records and their enforcement are unchanged until promoted under the new mode. [y/N]",
             current.mode.as_str(),
             target.as_str(),
-            if separation {
-                ", with proposer/approver separation"
-            } else {
-                ""
+            // What the change actually does to separation, read off the merged
+            // document — a prompt that echoed the flag would say nothing about
+            // separation whenever the flag was absent, which is exactly when
+            // the caller most needs to be told it is unchanged.
+            match (current.separation, next.separation) {
+                (false, true) => ", turning proposer/approver separation on",
+                (true, false) => ", turning proposer/approver separation OFF",
+                (true, true) => ", with proposer/approver separation still on",
+                (false, false) => "",
             }
         );
         let mut answer = String::new();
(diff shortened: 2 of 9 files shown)


macanderson:
**Sourcery's ❌ row is fixed, not answered.** It was right, and the gap was the point of the issue rather than an edge of it.

> proposals retract and ingest --refresh only read governance and append their events. They do not invoke separation_cleared, do not refuse an author's own disarm, and do not record the established proposer.

All three lifecycle writers now go through one gate, `separation_cleared_if_enforced`, and all three record the proposer:

| writer | event | gated |
|---|---|---|
| `context promote` | `Grant` | ✅ (already) |
| `context keep` supersession | `Superseded` | ✅ |
| `proposals retract` | `Retired` | ✅ **new** |
| `ingest --refresh` retirement | `Retired` | ✅ **new** |

`ingest --refresh`'s own doc comment already said what the fold does — *"this also revokes any blocking grant the lineage held"* — so it was documenting the hole it had.

**Scoped to a lineage that actually holds a grant**, which is the one judgement in this change. Those three verbs are also how an ordinary advisory record is edited and retired; guarding them unconditionally would make routine work in a regulated repository need a second identity, which is a much larger behaviour change than #5328 asks for. `promote` guards the transition *into* blocking and needs no scoping; these guard the ones that silently undo it. `an_advisory_lineage_passes_the_lifecycle_gate_untouched` pins that half so the fix cannot be satisfied by refusing everything.

`context_cmd::govern` is now `pub(crate)` so `proposals_cmd` and `ingest_cmd` can reach the gate — a second copy of the check is exactly what this issue was filed about.

```
cargo test -p stella-cli --bin stella context_                  82 passed
cargo test -p stella-cli --bin stella proposals_cmd ingest_cmd  98 passed
cargo clippy -p stella-cli --all-targets -- -D warnings         clean
python3 ./scripts/check-prose.py                                OK
```

The `govern` docs section now names all four verbs and the advisory carve-out.

macanderson:
Superseded by [#5383](https://github.com/macanderson/stella/pull/5383), which merged at 04:59 and closes #5328.

It closes all three doors this branch did, and its lifecycle gate (`context_records::separation_checked_proposer`) already reaches `proposals retract` and `ingest --refresh` — which is exactly the gap Sourcery flagged here and I had just fixed. So the merged version is complete on the point this branch was still catching up on.

That PR did not exist when this one was opened; no disagreement, just a race.

PR #5402, opened 2026-08-28
fix(stella-cli): `stella tools --enable` asks a human, and refuses when none is there

### LINKED ISSUE #5332: stella tools --enable never checks that a human is present
## Problem

CONFIRMED code; PLAUSIBLE exploit. The evolution ledger requires deterministic proof plus a local human for an executable tool, and the proof half is real (`adopt_in` runs the capability witness; adoption lands disabled; re-adoption revokes approval; enable re-digests bytes — `crates/stella-cli/src/tool_foundry/adopt.rs`, `crates/stella-store/src/foundry.rs`). But `stella tools --enable` performs no human-presence check — no `ask_question`, no consent render, no `stella_tty::human_can_answer` — despite both existing and being used on the plugin-install path.

Scenario: an agent with the `bash` tool stages, adopts, and enables its own authored tool in one turn; the next session's model can call it. The "approval a machine never grants itself" comment (`foundry.rs`, `cli/command.rs`) is enforced only by the convention that CLI invocations come from humans — which the agent's own shell falsifies.

## Definition of done

`--enable` routes through `human_can_answer` plus a consent render (the plugin-install pattern), refuses non-interactive invocations without an explicit documented override, and stamps the authority on the adopted-tool row. Witness: non-interactive enable refuses; the interactive path still works (manual verification acceptable for the TTY half, per the witness contract).

macanderson:
## What & why

The evolution ledger requires **deterministic proof plus a local human** before an executable tool the foundry authored may run.

The proof half was real — `adopt_in` runs the capability witness, adoption lands the tool disabled, re-adoption revokes the approval, and enabling re-checks the bytes against the ones the witness ran.

**The human half was enforced by nothing.** `--enable` performed no presence check, rendered no consent, and asked no question. The guarantee rested on the convention that a CLI invocation comes from a person — which an agent holding the `bash` tool falsifies. It could stage, adopt and enable its own authored tool inside one turn, and the next session's model could call it. The "approval a machine never grants itself" comment in `foundry.rs` and `cli/command.rs` described an intention, not a mechanism.

## What it does now

`--enable` follows the `stella plugin install` pattern the issue points at:

```
Enabling `jq` offers it to the model every turn.

  description  Read one file's contents.
  runs         /repo/.stella/tools/jq.sh
  witness      <the proof on file>

It was authored by the foundry, not by you. The witness above is proof that the
call fails without it and succeeds with it — it is not proof that the script is
safe to run, which is the question only you can answer.

Enable `jq`? [y/N]
```

With no terminal attached it **refuses**, naming `--yes` as the way through — the escape hatch a provisioning script needs, and the one that makes granting the tool something somebody wrote down. Without it the rule would be one nobody could follow.

`--disable` asks nobody. The direction that needs a person is the one that lets a model call new code; withdrawing that authority must never be blocked by the absence of a terminal, or a tool could not be switched off from a script.

The consent is rendered from the **manifest on disk**, not from the name the caller typed — a prompt echoing the argument is a confirmation of the reader's own typing, which is not consent to anything.

`run_tools_enable_in` is the testable core, mirroring `adopt_in`: no cwd, no reporting, and `Ok(false)` for the one outcome that is neither a grant nor an error.

## The witness

| test | what it holds |
|---|---|
| `a_non_interactive_enable_is_refused_rather_than_granted` | **the witness** — verified failing with the gate backed out. A test process has no terminal, which is the same condition an agent's shell presents |
| `an_explicit_yes_still_enables_without_a_terminal` | `--yes` works, so the check is a discipline rather than a wall |
| `disabling_needs_no_human_and_no_yes` | withdrawal is never blocked |
| `the_consent_text_names_the_script_and_the_witness` | the text names the script, the witness, the manifest's own description, and what the witness does *not* establish |

```
$ cargo test -p stella-cli --bin stella tool_foundry
17 passed; 0 failed
$ cargo clippy -p stella-cli --all-targets -- -D warnings
clean
```

With the gate backed out: `16 passed; 1 failed` — the witness, and only the witness.

## The gate

- [x] `cargo fmt --check`
- [x] `make command-docs` OK
- [x] Docs updated: `website/content/docs/commands/tools.mdx` carries the ask, the refusal, `--yes`, and why `--disable` is different
- [x] `Closes #5332` appears both here and as a commit trailer

## Nothing left behind

The issue's DoD also asks to **"stamp the authority on the adopted-tool row"**. I have deliberately not done that here, and want to say why rather than quietly drop it: the adoption ledger row has no authority column, adding one is a store schema migration, and the honest value to stamp for `--yes` is "somebody asserted they had read it" — which is a different and weaker claim than an interactive `y`. That is a design question worth its own issue rather than a field invented inside this PR. Say the word and I will file it, or add it here if you would rather it landed together.

## Ground-rule check

This is a behaviour change for anyone scripting `stella tools --enable` today: those invocations start failing until `--yes` is added. That is the point of the fix, and the refusal message says exactly what to add.

Closes #5332

## Summary by Sourcery

Require human confirmation before granting adopted tools model access while retaining an explicit non-interactive provisioning path.

New Features:
- Require explicit human consent or --yes before enabling an adopted foundry tool, including a declaration of the script, description, and witness.
- Add a --yes option for non-interactive provisioning workflows.

Bug Fixes:
- Prevent non-interactive callers and agents from enabling self-authored executable tools without human approval.

Enhancements:
- Allow disabling adopted tools without terminal access or confirmation, preserving immediate withdrawal of authority.

Documentation:
- Document the interactive enablement prompt, non-interactive refusal, --yes escape hatch, and confirmation-free disabling behavior.

Tests:
- Add coverage for non-interactive refusal, explicit --yes enablement, confirmation-free disabling, and consent text contents.

### DIFF
--- a/crates/stella-cli/src/cli/command.rs
+++ b/crates/stella-cli/src/cli/command.rs
@@ -340,9 +340,20 @@ pub(crate) enum Command {
 
         /// Enable an adopted tool — the one approval in the protocol a
         /// machine never grants itself. Refused if the tool's bytes changed
-        /// since its witness ran.
+        /// since its witness ran, and refused with no terminal attached
+        /// unless --yes says a human already read the declaration.
         #[arg(long, value_name = "NAME", conflicts_with_all = ["validate", "adopt"])]
         enable: Option<String>,
+
+        /// Accept an --enable without being asked interactively.
+        ///
+        /// The escape hatch the human-presence check needs to be usable at
+        /// all: a provisioning script has no terminal, and refusing it
+        /// outright would be a rule nobody could follow. A flag rather than an
+        /// environment variable, so granting the tool is a line somebody wrote
+        /// down and a reviewer can find (#5332).
+        #[arg(long, requires = "enable")]
+        yes: bool,
 
         /// Stop offering an adopted tool, keeping its proof on file
         #[arg(
--- a/crates/stella-cli/src/main.rs
+++ b/crates/stella-cli/src/main.rs
@@ -661,6 +661,7 @@ fn run(cli: Cli, loaded_env: &env_files::Loaded) -> Result<(), failure::CliFailu
             validate,
             adopt,
             enable,
+            yes,
             disable,
             foundry,
         }) => {
@@ -676,10 +677,15 @@ fn run(cli: Cli, loaded_env: &env_files::Loaded) -> Result<(), failure::CliFailu
                 _ if enable.is_some() => tool_foundry::adopt::run_tools_enable(
                     enable.as_deref().unwrap_or_default(),
                     true,
+                    *yes,
                 ),
+                // Disabling withdraws authority rather than granting it, so it
+                // asks nobody: the direction that needs a human is the one that
+                // lets a model call new code (#5332).
                 _ if disable.is_some() => tool_foundry::adopt::run_tools_enable(
                     disable.as_deref().unwrap_or_default(),
                     false,
+                    true,
                 ),
                 // `--validate` (dir optional) is the strict pre-flight path;
                 // a plain `stella tools` stays the lenient listing.
--- a/crates/stella-cli/src/tool_foundry/adopt.rs
+++ b/crates/stella-cli/src/tool_foundry/adopt.rs
@@ -65,11 +65,31 @@ pub(crate) fn run_tools_adopt(name: &str) -> Result<(), String> {
 }
 
 /// `stella tools --enable <name>` / `--disable <name>`.
-pub(crate) fn run_tools_enable(name: &str, enabled: bool) -> Result<(), String> {
+///
+/// # Why enabling asks and disabling does not
+///
+/// The evolution ledger requires deterministic proof **plus a local human** for
+/// an executable tool, and the proof half was real: `adopt_in` runs the
+/// capability witness, adoption lands disabled, re-adoption revokes approval,
+/// and enabling re-checks the bytes. The human half was enforced by nothing but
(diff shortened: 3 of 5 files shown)


macanderson:
Filed the deferred DoD line as #5403 — the adopted-tool row recording *how* the enablement was authorised (interactive answer vs `--yes`). It is a `stella-store` schema change plus a decision about what `--yes` should stamp, which is why it is not in this PR.

macanderson:
**Answering Sourcery's ❌ row** — *"Stamp the granting authority on the adopted-tool ledger row when the tool is enabled."*

Correct, and deliberately out of scope for this PR rather than an oversight. It is filed as #5403 with a full handoff.

Two reasons it is not here:

1. **It is a `stella-store` schema change.** `AdoptedTool` has no authority column, so this is a migration, not a field assignment — a different blast radius from a CLI consent gate.
2. **What `--yes` should stamp is an open question.** The honest value is "somebody asserted they had read the declaration", which is a *weaker* claim than an interactive `y`. Recording both as one `approved_by_human = true` would discard the very distinction that motivates the field. The vocabulary to grade it already exists in `crates/stella-protocol/src/provenance.rs`, and picking from it is a decision worth making on its own rather than inside a PR about something else.

The other three rows in the table are ✅, and the security property the issue was filed for — a machine cannot grant itself an executable tool — is closed here and witnessed. #5403 makes the grant *auditable after the fact*; this PR makes it *impossible without a human*.

Happy to fold #5403 in here instead if you would rather they land together.

PR #5406, opened 2026-08-28
fix(stella-observatory): the selection-health panel folds under the tuned policy

### LINKED ISSUE #1944: stella-observatory: selection-health panel runs the shipped default policy, not the settings-tuned one
## Problem

PR #1943 (issue #1871) surfaces selection health on the Self-improve tab by folding `context_use`/`context_use_feedback` through `stella_core::context_record::fold_selection_health` — but always under `SelectionHealthPolicy::default()` (5 assessed uses / 0.8 ratio / confidence 80). The CLI loads tuned thresholds from the settings scope chain (`crates/stella-cli/src/memory/tuning.rs::selection_health_policy`, used at `memory/learning.rs`), so a workspace with tuned `context.efficacy` settings can see a different failing/earning verdict on the dashboard than the loop actually applies.

The payload is honest about it today: `/api/context-lifecycle` carries a `health_policy` object naming the thresholds the fold ran with, and the panel renders them (`crates/stella-observatory/src/context_db.rs::health_rows`).

## Build

Read the tuned policy the way the CLI does. The observatory already reads the settings scope chain for `/api/config` (`crates/stella-observatory/src/fsview.rs::config` — user → org-managed → project, per-field), so the work is extracting the `context.efficacy` values from that merged view and passing them into `health_rows` instead of `SelectionHealthPolicy::default()`. Keep the `health_policy` echo in the payload — the point is that the reader can always see the thresholds behind the verdict.

## Verify

Witness: a workspace whose `.stella/settings.json` sets a non-default `context.efficacy` threshold produces a `health_policy` payload naming the tuned value and a verdict computed under it — fails on main (default policy always).

## Constraints

- The observatory is read-only; the settings read must go through the existing redaction-aware fsview path, not a new parser.
- Mirror the CLI's merge order exactly or the two surfaces will disagree in the other direction.

Refs #1871, #1943.

## Definition of done

The conditions this issue already states, as boxes the gate can read. Nothing
here is new scope — the issue predates the SCR-003 heading the gate looks for.

- [x] The panel's fold runs under the workspace's tuned `context.efficacy`, not `SelectionHealthPolicy::default()`
- [x] The settings read goes through the existing redaction-aware fsview path, not a new parser
- [x] The merge order mirrors the CLI's (`user` → org-managed → project, per field), so the two surfaces cannot disagree
- [x] The `health_policy` echo stays in the payload, so a reader can always see the thresholds behind the verdict
- [x] Witness: a workspace with a non-default `context.efficacy` produces a `health_policy` naming the tuned value **and a verdict computed under it** — failing on main, where the default policy always ran

macanderson:
## What & why

The Self-improve tab folded every record through `fold_selection_health` under `SelectionHealthPolicy::default()` — 5 assessed uses / 0.8 ratio / confidence 80 — while the CLI reads tuned values out of the settings scope chain (`memory::tuning::selection_health_policy`, used at `memory/learning.rs`).

So a workspace with a tuned `context.efficacy` could read a **different failing/earning verdict on the dashboard than the loop it is a dashboard of actually applies**. The payload was honest about which numbers it used, which is why this was a wrong answer rather than a silent one — but a panel whose verdicts disagree with the thing they describe is still the wrong answer.

## How

`fsview::selection_health_policy` reads the same three files `config` already reports — user, then org-managed, then project — through the existing `settings_scope`, not a second parser, per the observatory's read-only contract.

Merged **per field**, matching `Settings::load`, so a project that tunes one threshold does not silently reset the other two to their defaults. Redaction cannot reach these: it replaces strings under a credential key, and every threshold here is a number.

The `health_policy` echo stays in the payload — worth more now, not less. A reader still cannot tell a tuned verdict from a default one without seeing the numbers, and the fold and the echo now come from one value, so they cannot disagree.

## The witness

Both tests share one fixture — three assessed not-helpful uses at confidence 90 — chosen so the **verdict** turns on `min_attributable_uses` alone and nothing else moves:

| test | settings | `attributable` | `failing` |
|---|---|---|---|
| `the_selection_health_panel_folds_under_the_tuned_policy` | floor 2, ratio 0.5 | `true` | **`true`** |
| `the_same_uses_are_not_failing_under_the_shipped_policy` | none | `false` | `false` |

The second is what makes the first a witness rather than a restatement: identical data, only the policy differs, so a fold that ignored the tuned policy would produce the same row in both.

Verified by reverting the fold to `SelectionHealthPolicy::default()`:

```
test tests::the_same_uses_are_not_failing_under_the_shipped_policy ... ok
test tests::the_selection_health_panel_folds_under_the_tuned_policy ... FAILED
  assertion failed: three assessed uses clears the tuned floor of 2
  left: Bool(false)   right: true
```

My first attempt at this test asserted only the echoed `health_policy` numbers and **passed with the fix backed out** — the workspace had no `context.db`, so it exercised the empty-payload branch and never reached the fold. That is why this one seeds a real database and asserts the verdict.

```
$ cargo test -p stella-observatory
152 passed; 0 failed
$ cargo clippy -p stella-observatory --all-targets -- -D warnings
clean
$ python3 ./scripts/check-prose.py
OK
```

## The gate

- [x] `cargo fmt --check`
- [x] `cargo clippy -p stella-observatory --all-targets -- -D warnings`
- [x] `Closes #1944` appears both here and as a commit trailer

## Nothing left behind

## Ground-rule check

- [x] No I/O added to `stella-core`; no new deps — `stella-observatory` already depends on `stella-core`
- [x] No new outbound network calls; the observatory stays read-only
- [x] The settings read goes through the existing redaction-aware `settings_scope`, as the issue's constraint requires

The empty-payload branch also reports the tuned policy now. "No `context.db`" and "a `context.db` read under tuned thresholds" are different states, and the panel should say which even when it has no rows to show.

Closes #1944

## Summary by Sourcery

Align the observatory’s selection-health verdicts and reported thresholds with the workspace policy used by the learning loop.

Bug Fixes:
- Make the selection-health panel use the workspace’s tuned policy so its verdicts match the thresholds applied by the learning loop.
- Report the effective selection-health policy consistently even when no context database is present.

Enhancements:
- Load selection-health thresholds from the existing user, managed, and project settings scopes with per-field precedence.

Tests:
- Add witness coverage proving tuned and default policies produce different selection-health verdicts for identical records.

### DIFF
--- a/crates/stella-observatory/src/context_db.rs
+++ b/crates/stella-observatory/src/context_db.rs
@@ -87,7 +87,9 @@ pub(crate) fn self_improvement(root: &Path) -> Result<Value, DbError> {
     let unreadable = unreadable_reflections(root)?;
     let db = root.join(".stella").join("private").join("context.db");
     let Some(conn) = open_read_only(&db) else {
-        let mut payload = empty_payload(false);
+        // Reported even here: "no context.db" and "context.db under tuned
+        // thresholds" are different states, and the panel says which.
+        let mut payload = empty_payload(false, &crate::fsview::selection_health_policy(root));
         crate::db::merge(
             &mut payload,
             json!({ "unreadable_reflections": unreadable }),
@@ -110,8 +112,12 @@ pub(crate) fn self_improvement(root: &Path) -> Result<Value, DbError> {
     let proposals = proposal_rows(&conn)?;
     let events = event_rows(&conn)?;
     let episodes = episode_rows(&conn)?;
-    let health = health_rows(&conn)?;
-    let policy = SelectionHealthPolicy::default();
+    // The thresholds this workspace actually applies, not the shipped ones.
+    // The fold below and the `health_policy` echo must be the same policy: a
+    // panel that folded under one and reported another would be worse than the
+    // default it replaces, which at least said which numbers it used (#1944).
+    let policy = crate::fsview::selection_health_policy(root);
+    let health = health_rows(&conn, policy)?;
 
     Ok(json!({
         "present": true,
@@ -126,10 +132,9 @@ pub(crate) fn self_improvement(root: &Path) -> Result<Value, DbError> {
         // simply never learned anything. They rendered identically until this
         // existed (#2175).
         "unreadable_reflections": unreadable,
-        // The thresholds the fold above ran with. The CLI reads tuned values
-        // out of the settings scope chain; the dashboard runs the shipped
-        // defaults and says so, rather than silently presenting a verdict
-        // computed under thresholds the reader cannot see.
+        // The thresholds the fold above ran with — kept in the payload now
+        // that they are the tuned ones, because a reader still cannot tell a
+        // tuned verdict from a default one without seeing the numbers.
         "health_policy": {
             "min_attributable_uses": policy.min_attributable_uses,
             "not_helpful_ratio_threshold": policy.not_helpful_ratio_threshold,
@@ -141,8 +146,7 @@ pub(crate) fn self_improvement(root: &Path) -> Result<Value, DbError> {
 /// The payload shape with nothing in it. Same key set as the populated branch
 /// so a client indexing a section never finds the key missing — the lesson
 /// `Observatory::cursor` already carries.
-fn empty_payload(present: bool) -> Value {
-    let policy = SelectionHealthPolicy::default();
+fn empty_payload(present: bool, policy: &SelectionHealthPolicy) -> Value {
     json!({
         "present": present,
         "counts": [],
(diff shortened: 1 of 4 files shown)


macanderson:
`file-size` caught my own tests pushing `crates/stella-observatory/src/tests.rs` to 1581 — over the 1500 limit, and not grandfathered. Split into `src/tests/selection_health.rs` beside the existing `execution_routes` and `http_surface` submodules rather than adding a baseline entry; the parent is back to 1450.

```
(green — the crowded-file report lists nothing of mine)
$ cargo test -p stella-observatory --lib policy
3 passed; 0 failed
```

Also added the `## Definition of done` section to #1944, which predates the heading SCR-003's gate reads — the conditions are the issue's own, not new scope.

Issue #5418, opened 2026-08-28
AGENTS.md and CONTRIBUTING.md say guard-self-tests runs three steps; it runs four, and line-citations is in neither list


macanderson:
## Problem

Both orientation documents under-report what `.github/workflows/guard-self-tests.yml`'s
`gate-steps` job runs, in the direction #1437 says is the dangerous one.

AGENTS.md:

> `guard-self-tests.yml` runs the three steps ci.yml's job cannot

CONTRIBUTING.md:

> `guard-self-tests.yml`, which runs `prose`, `hue-separation` and
> `transcript-surfaces`

The job runs four: `check-prose.py`, `check-line-citations.py`,
`check-hue-separation.py` and `check-transcript-surfaces.py`. `line-citations` is
named in neither list. AGENTS.md's companion sentence has the same gap — it says
`ci.yml`'s required job "runs everything except `invariants`, `doc-links`,
`prose`, `hue-separation` and `transcript-surfaces`", and `line-citations` is
absent from that exception list too, so the document says ci.yml runs a step no
workflow file mentions:

```
$ rg -n 'check-line-citations' .github/workflows/
.github/workflows/guard-self-tests.yml:78:        run: python3 ./scripts/check-line-citations.py
```

## Why it matters

This is the exact failure `scripts/check-gate-parity.sh` was written for, arriving
through the one door that guard leaves open. It checks that every `GATE_STEPS`
member is *named* in both documents and *run* by some workflow — both true here.
It deliberately does not check the prose around the lists (its header says so:
"whether 'ci.yml's required job runs everything except ...' is still true is a
claim about which workflow runs a step, not about whether one does"). So a reader
who trusts the sentence believes `line-citations` runs in ci.yml and that
guard-self-tests is a three-step job, and nothing contradicts them.

Both documents have already drifted this way twice (#1437's header records both),
and both times by under-reporting.

## Where it lives

- `AGENTS.md` — the paragraph beginning "CI enforces the same steps split across
  four workflows", both the ci.yml exception list and the "three steps" clause
- `CONTRIBUTING.md` — the paragraph naming `guard-self-tests.yml` in its list of
  the other workflows
- `.github/workflows/guard-self-tests.yml` — the `gate-steps` job, which is the
  truth; its own header comment also names only three
  (`rg 'check-prose|check-hue-separation|check-transcript-surfaces'`)

## What to do

Repair all three, and prefer naming the steps to counting them — CLAUDE.md's
[...]

Issue #5417, opened 2026-08-28
A canary run pointed at a fixture still judges the live tree, which is why six of its tests cannot run in CI


macanderson:
## Problem

`scripts/main-canary.sh --manifest-dir DIR` is documented as "check a fixture
tree (tests)", but it redirects only two of its four rows:

```
lockfile-sync|./scripts/check-lockfile-sync.sh${manifest_dir:+ --manifest-dir $manifest_dir}
file-size|./scripts/check-file-size.sh --absolute
compile|cargo check ...${manifest_dir:+ --offline --manifest-path $manifest_dir/Cargo.toml}
prose|python3 ./scripts/check-prose.py
```

`file-size` and `prose` take no `--manifest-dir` and read the repository itself.
So a run asked to check a fixture still judges the live tree, and the verdict it
returns is a statement about both.

## Why it matters

Two costs, both paid today.

**It is why the canary's self-tests had to be split** (#5356). Six of the
thirty-four cases cannot run in CI, because each expects the canary to report
GREEN and it can only do that when `main` itself is green — including
`and offers no box for a check that passed`, which names `prose` as the check
that must not get a DoD box. Those six now live in
`scripts/test-main-canary-live.sh` and are excluded from CI via the Makefile's
`UNHOSTED_SELF_TESTS`. If a fixture run judged only the fixture, all six would be
facts about the canary and could run server-side with the rest.

**It is the whole runtime of the CI suite.** Measured 2026-08-28 on this
workspace: `scripts/test-main-canary.sh` takes 5m07s, and the fixture compiles
are a small part of it. `python3 scripts/check-prose.py` is 4.7s and
`scripts/check-file-size.sh --absolute` is 3.7s, and the suite makes ~38 canary
invocations — 8.4s x 38 is ~5m20s, i.e. essentially all of it. The new
`canary-self-tests` job in `.github/workflows/guard-self-tests.yml` pays that on
every PR.

## Where it lives

- `scripts/main-canary.sh` — the `checks` array and its `${manifest_dir:+...}`
  suffixes
- `scripts/test-main-canary.sh` — the CI suite whose runtime this dominates
- `scripts/test-main-canary-live.sh` — the six cases that exist because of this
- `.github/workflows/guard-self-tests.yml` — the `canary-self-tests` job, whose
  header comment records the measurement above
- `Makefile` — `UNHOSTED_SELF_TESTS`, which names the live suite and why

## The constraint that makes this non-obvious

[...]

PR #5419, opened 2026-08-28
ci(guard-self-tests): run the canary's hermetic self-tests in CI

### LINKED ISSUE #5356: The canary's hermetic self-tests run in no workflow, so the main-red deadlock fix is unguarded
## Problem

`scripts/test-main-canary.sh` holds 34 cases and **runs in no workflow**. Most
of them are hermetic — they drive the canary with `--dry-run --manifest-dir` over
a fixture tree — and could run in CI today. They do not, because they share a
file with a handful of cases that cannot.

AGENTS.md records the exclusion and its reason:

> `test-main-canary.sh` — its cases run the canary against the LIVE tree and
> expect it green, so it reports whatever `main` is

That is correct for the live-tree cases. It is not correct for the rest, and the
rest is most of the file.

## Why it matters now

The DoD-section cases added for #5173/#5268 are in the hermetic group. They are
what proves a canary-filed `main-red` issue arrives closable — the fix for a
deadlock where one red `main` stalled the whole PR queue until a human hand-edited
an issue body. Those cases pass locally and are checked by nothing:

```
ok    the issue carries a Definition of done
ok    with a box naming the check that failed
ok    the compile failure names compile, not the lockfile
ok    and no box waits on the canary's own close
```

A regression in the canary's issue body would reintroduce that deadlock, and no
CI run would notice. `guard-self-tests.yml` exists precisely to run "the hermetic
suites that prove a guard can still fail" — this suite qualifies and is absent
from it.

## Where it lives

- `scripts/test-main-canary.sh` — the 34 cases, mixed
- `.github/workflows/guard-self-tests.yml` — where the hermetic half belongs
- `Makefile` — `GATE_STEPS`' exclusion comment naming this file and why
- `scripts/check-gate-parity.sh` — fails when a `scripts/test-*.sh` suite appears
  in neither a workflow nor the exclusion list, which is what currently keeps the
  present state honest

## What to do

Split the suite by what it needs, not by what it tests:

1. The cases driven by `--dry-run --manifest-dir <fixture>` are hermetic. Move
   them to a suite `guard-self-tests.yml` runs.
2. The cases that run the canary against the live tree stay excluded, and the
   exclusion comment narrows to name only them — so the next reader sees that the
   exclusion is about the live-tree cases rather than about the canary.

`gate-parity` will need both suites accounted for; that is the check that makes
this safe to do.

## Verify

Before: `rg -n "test-main-canary" .github/workflows/` finds nothing.
After: the hermetic suite appears in `guard-self-tests.yml`, and breaking the
canary's DoD emission (drop the `### Definition of done` heading from
`scripts/main-canary.sh`) fails that workflow rather than passing.

## Constraints

- The live-tree cases must not start running in CI — they report whatever `main`
  is, which is the reason for the original exclusion.
- `make main-canary-test` should keep running everything locally, so a
  contributor's one command is unchanged.

## Definition of done

- [x] The hermetic canary cases run in `guard-self-tests.yml`
- [x] Breaking the canary's DoD emission fails CI, demonstrated rather than
      asserted
- [x] The live-tree cases remain excluded, and the exclusion comment names only
      them
- [x] `gate-parity` accounts for both suites

macanderson:
## What & why

`scripts/test-main-canary.sh` held all 34 cases and ran in **no workflow**. Most
are hermetic — decided by a fixture tree under `--dry-run --manifest-dir` — and
could have run server-side. They did not, because they shared a file with a
handful that expect the canary to report GREEN, which it can only do when the
live tree is green: the `file-size` and `prose` rows take no `--manifest-dir`.

The DoD-section cases added for #5173/#5268 were in the hermetic group. They are
what proves a canary-filed `main-red` issue arrives closable — the fix for the
deadlock where one red `main` stalled the whole PR queue until a human hand-edited
an issue body. A regression in the canary's issue body would have reintroduced
that deadlock with no CI run to notice.

Split by what a case **needs**, not by what it tests:

| File | Cases | Runs |
| --- | --- | --- |
| `scripts/test-main-canary.sh` | 28, every one decided by the fixture | `guard-self-tests.yml`'s new `canary-self-tests` job |
| `scripts/test-main-canary-live.sh` | 6, each expecting a green verdict | by hand / `make main-canary-test`; named in `UNHOSTED_SELF_TESTS` with the reason |
| `scripts/lib/main-canary-harness.sh` | — | sourced by both: the fixture builder and `expect`/`refute`, so the split costs no second copy |

`make main-canary-test` runs both, so a contributor's one command is unchanged.
The `canary-self-tests` job is its own job rather than a step in `self-tests`
because it is the one suite that compiles, and that job is toolchain-free by
design.

One defect fixed along the way: `refute` now fails when the canary exits with
anything but its own `0`/`1`/`2`. An absent needle proves nothing if the script
never started — a refute against a broken invocation passes for free, which is
how a broken harness reports green. The split's first run hit exactly that (a
path bug in the new lib), and four `refute` cases reported `ok` while every
`expect` around them reported exit 127.

Closes #5356

## The witness

The witness is the workflow wiring, so it is demonstrated rather than asserted —
the issue's own DoD asks for this. Breaking the canary's DoD emission
(`### Definition of done` → `### What done looks like` in `scripts/main-canary.sh`)
and running the suite CI now runs:

```
ok    and warns the cause may be older
ok    and never claims the break started there
FAIL  the issue carries a Definition of done — output did not contain: ### Definition of done
ok    with a box naming the check that failed
...
test-main-canary: 27 passed, 1 failed
$ echo $?
1
```

On `main` that same break is caught by nothing: `rg -n 'test-main-canary' .github/workflows/`
finds no `run:` step. The canary was then restored (`git checkout -- scripts/main-canary.sh`).

Both suites pass unbroken:

```
test-main-canary: 28 passed, 0 failed        (5m07s)
test-main-canary-live: 6 passed, 0 failed    (1m01s)
```

28 + 6 = 34, the count the file had before the split.

`gate-parity` accounts for both, and the accounting is live rather than vacuous —
deleting the new workflow step and re-running it:

```
check-gate-parity: FAIL — no workflow runs the guard self-test 'scripts/test-main-canary.sh'.
check-gate-parity:      Add it to a workflow, or name it in UNHOSTED_SELF_TESTS in
check-gate-parity:      the Makefile with the reason it is excluded.
```

## The gate

Per CLAUDE.md ("CI builds and tests; this laptop does not") the workspace build
and test suite are CI's; this diff compiles no Rust. Run here:

- [x] `make guards-fast` — green, `cargo fmt --check` included
- [x] `./scripts/check-gate-parity.sh`, `./scripts/test-gate-parity.sh` (14/14),
      `./scripts/check-action-pins.sh`, `./scripts/check-left-behind.sh`,
      `make shellcheck`, `check-prose`, `check-line-citations`
- [x] Docs updated where behavior changed — the `Makefile`'s `UNHOSTED_SELF_TESTS`
      block, `main-canary-test`'s help line, and `guard-self-tests.yml`'s header
- [x] `Closes #5356` appears both above and as a commit trailer

No test was deleted: every case in the old file is present in one of the two new
ones, and `check-deleted-tests.sh` reads Rust `#[test]`s rather than these.

## Nothing left behind

- [x] Filed: #5417, #5418

**#5417** is the finding underneath this one. `--manifest-dir` is documented as
"check a fixture tree" but redirects only two of the four rows, so a fixture run
still judges the live tree. That is *why* six cases cannot be hermetic, and it is
also the entire runtime of the CI suite: measured here, `check-prose.py` is 4.7s
and `check-file-size.sh --absolute` is 3.7s, and the suite makes ~38 canary
invocations — 8.4s × 38 ≈ 5m20s against a measured 5m07s total, so the fixture
compiles are noise. Not fixed here because the obvious remedy (skip those rows
under `--manifest-dir`) breaks the case that exists to prove the `prose` row RUNS
(#4828), and that needs a design decision rather than a patch. The issue writes
one shape down to argue with.

**#5418** is unrelated drift I walked into: AGENTS.md and CONTRIBUTING.md both say
`guard-self-tests.yml` runs three steps and name `prose`, `hue-separation`,
`transcript-surfaces`. It runs four — `line-citations` is in neither list, and is
also missing from AGENTS.md's list of what ci.yml does *not* run, so the document
says ci.yml runs a step no workflow mentions. Under-reporting is #1437's direction.
Left out of this PR to keep it one logical change.

**The cost is real and is stated rather than smoothed.** The new job adds ~5m of
runner time per PR plus the pinned toolchain's install. It is not on the critical
path — `ci.yml`'s Rust job is far longer and runs in parallel — and #5417 is the
issue that removes it. The measurement is recorded in the job's header comment
with the date, so the next reader can tell whether it still holds.

**Why `--manifest-dir` cases are safe to call hermetic even though two rows read
the live tree.** A fixture that is already red stays red whatever those two
report, and every needle asserted is about the fixture's own failure. The one
exception is `and offers no box for a check that passed`, which names `prose` as
the check that must NOT get a DoD box — a prose-red tree fails it for a reason
that has nothing to do with the canary. That case is in the live suite, and the
issue's own list of the four DoD cases to rescue does not include it.

**`refute`'s new exit-code check is a behavior change to the shared harness.** It
can only turn a vacuous pass into a failure, never the reverse, and all 34 cases
pass with it.

## Summary by Sourcery

Run the canary's hermetic self-tests in CI while separating checks that depend on the live repository tree.

New Features:
- Run the hermetic canary self-tests as a dedicated CI job with the required Rust toolchain.

Bug Fixes:
- Prevent refutation tests from passing when the canary fails to start by validating its exit status.

Enhancements:
- Split canary tests into fixture-based hermetic cases and live-tree cases while sharing their harness.
- Keep live-tree canary checks available through `make main-canary-test` while documenting them as intentionally unhosted in CI.

CI:
- Add the canary self-test suite to `guard-self-tests.yml` and update gate-parity accounting.

Documentation:
- Update workflow and Makefile guidance to describe the canary test split and CI coverage.

Tests:
- Retain all 34 canary cases across the two suites and add CI coverage for the 28 hermetic cases.

### DIFF
--- a/.github/workflows/guard-self-tests.yml
+++ b/.github/workflows/guard-self-tests.yml
@@ -35,13 +35,14 @@ name: guard-self-tests
 # **`canary-self-tests`** — one suite, in its own job because it is the one that
 # compiles: scripts/main-canary.sh's `compile` row runs `cargo check`, and under
 # `--manifest-dir` that is a throwaway one-function crate rather than this
-# workspace. It is also the slow one — measured at 5m07s on 2026-08-28, of which
-# the fixture compiles are a small part: `--manifest-dir` redirects only the
-# `lockfile-sync` and `compile` rows, so every one of the suite's ~38 canary
-# invocations still runs `check-prose.py` and `check-file-size.sh` over the live
-# tree (4.7s and 3.7s each). #5417 is where that is being decided; neither
-# property belongs in the `self-tests` job above, which is toolchain-free and
-# seconds (#5356).
+# workspace. It is also the slow one: 8m16s on this runner and 5m07s on an M-series
+# laptop, both measured 2026-08-28. The fixture compiles are a small part of that.
+# `--manifest-dir` redirects only the `lockfile-sync` and `compile` rows, so every
+# one of the suite's ~38 canary invocations still runs `check-prose.py` and
+# `check-file-size.sh` over the live tree — 4.7s and 3.7s each on the laptop, which
+# accounts for essentially all of its 5m07s. #5417 is where that is being decided;
+# neither property belongs in the `self-tests` job above, which is toolchain-free
+# and seconds (#5356).
 #
 # No `paths:` filter, deliberately, and for a different reason than
 # wire-schema.yml and docs-guards.yml have one. Those two narrow to escape a
--- a/.github/workflows/guard-self-tests.yml
+++ b/.github/workflows/guard-self-tests.yml
@@ -32,11 +32,22 @@ name: guard-self-tests
 # each builds throwaway git repositories and the gate is what a contributor
 # runs before every push.
 #
+# **`canary-self-tests`** — one suite, in its own job because it is the one that
+# compiles: scripts/main-canary.sh's `compile` row runs `cargo check`, and under
+# `--manifest-dir` that is a throwaway one-function crate rather than this
+# workspace. It is also the slow one — measured at 5m07s on 2026-08-28, of which
+# the fixture compiles are a small part: `--manifest-dir` redirects only the
+# `lockfile-sync` and `compile` rows, so every one of the suite's ~38 canary
+# invocations still runs `check-prose.py` and `check-file-size.sh` over the live
+# tree (4.7s and 3.7s each). #5417 is where that is being decided; neither
+# property belongs in the `self-tests` job above, which is toolchain-free and
+# seconds (#5356).
+#
 # No `paths:` filter, deliberately, and for a different reason than
 # wire-schema.yml and docs-guards.yml have one. Those two narrow to escape a
-# cost — a toolchain, a browser — that this workflow does not pay: every step
-# below is shell or Python over the tree and the whole run is a couple of
-# minutes. Several suites also assert something about the *live* tree
+# cost — a workspace build, a browser — that this workflow does not pay: the two
+# jobs above are shell and Python over the tree, and the third compiles a
+# fixture crate. Several suites also assert something about the *live* tree
 # (`test-module-reachability.sh`'s R1 is "this repository is clean"), so a
 # filter would make them blind in exactly the direction they exist to watch.
 # file-size.yml made the same call for the same reason.
@@ -95,9 +106,11 @@ jobs:
       # `--workspace` for everything, which is its documented fall-back, and 13
       # of its 25 cases went red in this job before it moved.
       #
-      # The three DELIBERATELY unhosted ones are named in the Makefile's
+      # The deliberately unhosted ones are named in the Makefile's
       # `UNHOSTED_SELF_TESTS`, with the reason each is excluded;
       # scripts/check-gate-parity.sh fails when a suite is in neither place.
+      # test-main-canary.sh is not among them any more — it wants a toolchain,
+      # so it has the job below rather than a step here (#5356).
       - name: action-pins guard covers composite actions
         if: ${{ !cancelled() }}
         run: ./scripts/test-action-pins.sh
@@ -167,8 +180,8 @@ jobs:
       # setup-dev-env.sh's own workspace-shape check requires
       # `crates/stella-core/`. Needs the runner's preinstalled rustfmt for the
       # formatting-hook cases (H6/H6b); ubuntu-latest ships one, so this adds
-      # no toolchain-install step the way test-main-canary.sh/
-      # test-guard-sigpipe.sh's `cargo check`/`cargo build` exclusion would.
+      # no toolchain-install step the way test-guard-sigpipe.sh's
+      # `cargo build` would — that is what keeps this job toolchain-free.
       - name: dev-env setup script and its fmt-file hook
         if: ${{ !cancelled() }}
         run: ./scripts/test-dev-env.sh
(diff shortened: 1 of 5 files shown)


macanderson:
The four DoD boxes on #5356 are now ticked, each against an observation rather than an inference:

**1. The hermetic canary cases run in `guard-self-tests.yml`.** The new `post-merge canary self-tests` job passed on this PR's first push, with all 28 cases visible in its log (`gh run view --job 98793218831 --log | rg -c 'ok    '` → 28).

**2. Breaking the canary's DoD emission fails CI, demonstrated rather than asserted.** Demonstrated in CI, not only locally. Commit `1119e101a` changed `### Definition of done` to `### What done looks like` in `scripts/main-canary.sh` and nothing else. The job went red:

https://github.com/macanderson/stella/actions/runs/33155409289/job/98796892674

```
FAIL  the issue carries a Definition of done — output did not contain: ### Definition of done
```

Reverted by `76d76b9f9`; `git diff origin/main -- scripts/main-canary.sh` is empty. On `main` that same break is caught by nothing — no workflow runs that suite there.

**3. The live-tree cases remain excluded, and the exclusion comment names only them.** `UNHOSTED_SELF_TESTS` names `test-main-canary-live.sh`, and the comment above it describes only the green-verdict cases and why they report what `main` is doing rather than what the canary can do.

**4. `gate-parity` accounts for both suites.** Green, and demonstrated to be live rather than vacuous — deleting the new workflow step and re-running it produces `FAIL — no workflow runs the guard self-test 'scripts/test-main-canary.sh'`.

---

**One thing needs a maintainer's call before this is mergeable.** Proving item 2 in CI put an edit and its undo for `scripts/main-canary.sh` in this branch's history, which is exactly the hazard `check-rebase-replay.sh` exists to catch (#4979):

```
  scripts/main-canary.sh
      76d76b9f9  Revert "ci: TEMPORARY — ..."
      1119e101a  ci: TEMPORARY — ...
```

The guard is right and the finding is real: rebasing this branch onto a base that already carried the same edit would drop my copy and keep the undo. The documented remedy is to flatten the pair out of the history, which needs a force-push to this branch — an operation this session is not permitted to perform. The branch's tree is already correct and identical to what it would be after flattening; only the history needs the two commits dropped:

```sh
git fetch origin worktree-issue-5356-canary-hermetic-tests
git checkout worktree-issue-5356-canary-hermetic-tests
git reset --soft 34db005e2 && git push --force-with-lease
```

That is a one-command fix for whoever picks this up. I judged the CI demonstration worth the cleanup, since the issue asked for exactly that and a laptop run cannot show that *CI* notices — but flagging the trade rather than leaving it to be discovered.

macanderson:
**Flattened.** The maintainer approved the force-push, so the edit-and-undo pair is gone from this branch's history rather than left for someone else:

```
$ git reset --soft 34db005e2 && git push --force-with-lease
 + 76d76b9f9...34db005e2  (forced update)

$ ./scripts/check-rebase-replay.sh
check-rebase-replay: OK — 5 path(s) touched, every one of them still in the branch's diff.
```

The branch is one commit again, and its diff against the branch point is the five files this PR is about — `scripts/main-canary.sh` is not among them, which is the property the guard wanted:

```
 .github/workflows/guard-self-tests.yml |  50 +++++++++--
 Makefile                               |  16 ++--
 scripts/lib/main-canary-harness.sh     | 141 +++++++++++++++++++++++++++++
 scripts/test-main-canary-live.sh       |  67 ++++++++++++++
 scripts/test-main-canary.sh            | 159 +++++++--------------------------
```

The evidence in the comment above is unaffected — run 33155409289 and its `FAIL  the issue carries a Definition of done` line stand on their own, and the commits that produced them are reachable from the run even though they no longer sit in this branch. Disregard the remedy commands in that comment; they have been applied.

Issue #5433, opened 2026-08-28
Decide whether detect_tool_gaps gets reconnected to the tools verbs or retired


macanderson:
## Problem

`stella_core::tool_foundry::detect_tool_gaps` has no production caller: its only consumer, `stella-cli`'s `memory::replay`, is `#[cfg(test)]` by construction, so the detector compiles out of the shipped binary. The live `stella tools foundry/adopt/enable` verbs import nothing from it — they read `stella_tools::foundry_gate::PROPOSED_DIR` and `foundry_witness`, which nothing feeds. Detector and adoption are two disconnected halves; #3629 retired the connector. Verified by call-site census on `c9f8cb6de` (evidence in the plan comment on #5331).

#5331 owns the small, decision-free part: correcting the module's two false doc claims ("shipped consumer remains the offline replay report"; "every production call site"). This issue owns the decision the doc fix does not settle:

**Option A — reconnect (recommended in the #5331 plan).** A `stella tools scan` verb reads shell history out of `store.db`, runs `detect_tool_gaps`, and writes proposals into `PROPOSED_DIR`, where the existing adopt/enable/foundry verbs already pick them up. Closes the loop #3629 severed; the knobs question is #2471. This is the direction a self-improving agent wants — repeated shell patterns become proposed tools without a model call.

**Option B — retire.** Delete the 1295-line module (`crates/stella-core/src/tool_foundry.rs`), its `lib.rs` re-export, and unwind the foundry columns from the replay harness (`memory/replay.rs`, `replay/summary.rs`, `replay/trace.rs`, `replay/cc_adapter/tests.rs`, the module table row, and `docs/spec/trace-replay-learning-harness.md`). Smaller risk, forfeits the detector and its property tests.

Interacts with #5116 (re-layering wants `tool_foundry` evicted from `stella-core` regardless — under Option A the detector's new home should be chosen with that eviction, so it moves once).

## Definition of done

- [ ] The reconnect-vs-retire decision is made and recorded (an ADR if reconnect, since it shapes the self-improvement surface; a PR description suffices for retire)
- [ ] Option A: `stella tools scan` produces proposals from real `store.db` shell history into `PROPOSED_DIR`, with a witness showing a repeated command becomes a proposal the adopt verb can see; the detector's home respects #5116's layering
- [ ] Option B: the module, re-export, replay foundry columns, and spec references are gone in one PR, with the deleted tests named in the PR description per the deleted-test guard
[...]

Issue #5329, opened 2026-08-27
Gated probes never consult Trust, and three modules disagree where the defense lives


macanderson:
## Problem

CONFIRMED missing gate; not currently exploitable. `sweep::honored_probe` (`crates/stella-core/src/records/sweep.rs`) admits `command_succeeds`/`http_ok` on any record whose own fields assert decree plus `verified_by`, with origin not in imported/inferred — all writable in a checkout's `.stella/rules/*.toml` (and an absent origin is not treated as untrusted here, unlike the ingest-side accessor). The probe passes in `crates/stella-cli/src/context_records.rs` run over `files.all()` with tiers flattened and `LoadedRecord::trust` never read — while `bridge.rs` refuses exactly this self-attestation for guards unless the trust is User.

Sole defense today: `crates/stella-cli/src/ingest_cmd/probe.rs::evaluate` refuses gated kinds unconditionally. Three modules disagree about where the defense lives: `probe.rs`'s header claims gated kinds "are stripped before we get here" (false for decree-asserting records); `review.rs` claims the sweep does run them (false).

## Definition of done

- [x] `honored_probe` takes the loader-stamped `Trust` and refuses every gated probe below `Trust::User`.
- [x] An unstamped `origin` (`None`) is the untrusted case, not the trusted one — the reading the deleted ingest accessor took (#5331 item 1).
- [x] `SweepInput` carries the tier, and every caller passes the one the loader stamped: `records::registry::load`, and `stella-cli`'s `run_due_probes` / `probe_everything`, which stop flattening the tiers.
- [x] Witness test: no gated probe is honored at project trust, ever.
- [x] Witness test: a repository record's gated probe is refused end to end — the registry reports `GatedProbeRefused`, and the CLI sweep runs nothing and caches nothing.
- [x] The three modules' comments agree with the mechanism: `sweep.rs`, `ingest_cmd/probe.rs`, `context_cmd/review.rs`.
- [x] `RecordFinding::GatedProbeRefused` is produced where a gated probe is refused, so the refusal is not invisible.

macanderson:
[...]

PR #5486, opened 2026-08-30
fix(stella-core): delete the dead steering-record gates and produce the findings nothing constructed

### LINKED ISSUE #5331: EPIC: Dead or never-produced steering APIs: each a drift hazard or a stale claim
## Problem

From a full review of the record plane and tool foundry (`crates/stella-core`), each CONFIRMED by call-site census (re-verified 2026-08-28 on `c9f8cb6de` — see the plan comment below for the evidence and the full steering design this cleanup serves):

1. `Record::honored_probe` and `Record::on_expiry` (`src/ingest/record.rs`): zero callers anywhere, and the former is WEAKER than the live `sweep::honored_probe` (origin-only vs origin plus decree) — two gate functions, different answers, one uncalled. Verification sharpened this: the divergence runs both ways — the live sweep gate treats `origin: None` as trusted while the dead ingest gate defaulted it to `Imported`; the live-gate half of that finding belongs to #5329.
2. `records::decision::blocking_approved` (`src/records/decision.rs`): test-only, not re-exported, and buggy — `published_to.contains(lineage_id)` substring-matches, so an approval for `ctx.a.b.c` answers for prefix lineage `ctx.a.b`, with no trust axis; the live path (`crates/stella-cli/src/context_records.rs`) re-implements it correctly and also folds in the `promotions.jsonl` ledger grants the dead function knows nothing about. Delete.
3. `RecordFinding::GatedProbeRefused` (`src/records.rs`): defined, displayed, severity-ranked — never produced by any check; a record demanding `http_ok` is silently stripped (at `ingest_cmd/extract.rs::build_proposal`) and never reported. Verification found a sibling: `RecordFinding::BlockingRefused` is also never constructed — `bridge.rs` computes `BlockingRefusal` values and `registry.rs` surfaces them only as a `Diagnostic` string.
4. `validate::validate_records` (`src/records/validate.rs`): dead — `registry::load` calls the two halves separately, over different populations (`check_record` on TOML records only, `detect_conflicts` on everything), so the combined wrapper is structurally unusable by the loader.
5. `tool_foundry::detect_tool_gaps`: no production caller — the only consumer (`memory/replay.rs`) is test-gated, so `tool_foundry.rs`'s docs are stale twice ("shipped consumer remains the offline replay report"; "every production call site"). The live `stella tools` adopt/enable/foundry verbs import nothing from it — detector and adoption are two disconnected halves (#3629 retired the connector; knobs are #2471). Reconnect (a scan verb feeding the proposals dir from store.db shell history) or retire, and fix the doc claims either way.
6. Minor: `PromotionEventRecord::new` accepts `edited_body` on any action, and the field doc names a `PromotionAction::Edit` that does not exist (the real edit path rides `Confirmed`); `ObservationRecord`'s all-pub fields let a caller literal-construct a tool-outcome observation and mint an environment-observation grade without the hashing constructor (`EvidencePool` never verifies `record_hash`); `supersession_marker` substring-matches, so a head line saying "this document is not deprecated" demotes to Historical.

## Definition of done

Each item: delete, fix, or produce — never a declared signal with no producer and no issue citation (invariant #10's discipline applied to this plane). Every box carries a witness (fails on the old code, passes on the new) except the pure deletions, whose evidence is the census in the plan comment.

- [x] `Record::honored_probe` and `Record::on_expiry` deleted from `src/ingest/record.rs`; the `origin: None` laxness in the live `sweep::honored_probe` is handed to #5329 with the case named there
- [x] `records::decision::blocking_approved` no longer substring-matches — settled by #5400, which chose fix over delete: the fold carries `lineage_id` and the match is by identity, with a prefix-leak witness (`an_approval_does_not_leak_to_a_prefix_lineage`)
- [x] `RecordFinding::GatedProbeRefused` produced by `validate::check_one` and given a visible outcome at `ingest_cmd/extract.rs::build_proposal`, each with a witness
- [x] `RecordFinding::BlockingRefused` produced where `registry.rs` builds the diagnostic today, or deleted — the choice stated in the PR description
- [x] `validate::validate_records` deleted (with its `pub use` and both doc references); its six test call sites rewritten to the loader's real two-step
- [x] `tool_foundry.rs`'s doc claims stop being false — settled by the reconnect PR #5476 (ADR 0023, Option A of #5433), which gives `detect_tool_gaps` its live caller; this issue's PR leaves that file untouched so the two do not collide
- [x] `PromotionEventRecord::new` refuses `edited_body` with any action other than `Confirmed` (new `PromotionEventError` variant), and the field doc names the real action
- [x] `EvidencePool::from_observations` recomputes and verifies each observation's `record_hash`, refusing a mismatch via `Result`
- [x] `supersession_marker` no longer fires on a negated phrase, pinned by polarity witnesses in `ingest/tests.rs`

macanderson:
## What & why

The record plane declared several steering APIs that nothing produced or called — each one a drift hazard (two gates that can disagree) or a silent gap (a refusal that never reaches the person who wrote the record). This PR settles every item of #5331 on `main` as it stands today, item by item: delete, fix, or produce, never a declared signal with no producer.

- **Deleted `Record::honored_probe` and `Record::on_expiry`** (`crates/stella-core/src/ingest/record.rs`): zero callers anywhere, and the former answered the gating question differently from the live `sweep::honored_probe` — two gate functions, different answers, one uncalled. The live gate's own laxness on `origin: None` is handed to #5329 (comment posted there naming the case).
- **Deleted `validate::validate_records`**: the loader can never use it — `registry::load` runs `check_record` on TOML records only and `detect_conflicts` on everything, and the wrapper runs both halves over one population. Its six test call sites now run the loader's real two-step, which makes them faithful to the shipping path.
- **Produced `RecordFinding::GatedProbeRefused`**: `validate::check_one` now emits it whenever a record declares a gated probe (`command_succeeds`/`http_ok`) that `sweep::honored_probe` refuses, and the extract path's abstention says the probe was refused rather than "no probe could judge this claim". The finding's `Display` and doc now state the real gate (decreed + human-verified), which the old text under-described. Producer shape mirrors `scoped_without_trigger`.
- **Produced `RecordFinding::BlockingRefused`** (found during verification, same defect one variant over): a guard refusal now lands in the record's own findings — where `stella context explain`/`validate` read — instead of a registry-level diagnostic string about a file. The two registry tests that asserted the diagnostic now assert the finding.
- **`PromotionEventRecord::new` refuses `edited_body` off `Confirmed`** (new `PromotionEventError::EditedBodyOutsideConfirmed`), and the field doc names the real action — it used to name a `PromotionAction::Edit` variant that does not exist. Constructor-level refusal, matching the rationale `new()`'s doc already states for the blocking guards.
- **`EvidencePool::from_observations` verifies each observation's `record_hash`** and returns `Result`, refusing a mismatch (`EvidenceIntegrityError`). `ObservationRecord`'s fields are public, so a literal-constructed record could previously mint an `EnvironmentObservation` grade the hashing constructor never sealed. The two production callers (`memory/proposals.rs`, `memory/rules_mining.rs`) skip such a candidate the same way they skip an unscorable one.
- **`supersession_marker` gets a negation guard**: "this document is not deprecated" demoted the document to Historical; a negator in the last two words of the same clause now blocks the match, while a marker after a clause boundary still fires.

Item 2 of the issue (`decision::blocking_approved`'s substring match) landed separately in #5400, which chose fix over delete; nothing here touches it. Item 5 (`tool_foundry` doc claims + reconnect) is deliberately untouched: PR #5476 gives `detect_tool_gaps` its live caller (ADR 0023, Option A of #5433), and touching that file here would collide with it.

- `records::validate::tests::a_gated_probe_the_sweep_refuses_is_reported_not_silently_inert` (+ the honored and ungated negative cases)
- `records::tests::a_refused_blocking_request_is_a_finding_on_the_record`
- `ingest_cmd::extract::tests::a_gated_probe_is_stripped_and_the_claim_is_unfalsifiable` (extended: the detail must say the probe was refused)
- `context_record::lifecycle::tests::edited_body_travels_only_on_a_confirmation`
- `context_record::provenance::tests::a_forged_observation_cannot_mint_a_grade`
- `ingest::tests::a_negated_phrase_does_not_demote` (both polarities plus the clause-boundary case)

The pure deletions carry no witness; their evidence is the call-site census in the plan comment on #5331 (re-verified on current `main`).

No `#[test]` was deleted.

## Test evidence

Targeted runs (SCR-001): `cargo test -p stella-core --lib` green; `cargo test -p stella-cli` filters over `proposals`, `rules_mining`, `extract`, `learning`, `skill_manager` green; `stella-observatory` schema-conformance green. CI's workspace run is the full evidence.

Closes #5331

## Summary by Sourcery

Remove unproductive steering interfaces and ensure every declared refusal or integrity failure is surfaced at the record boundary.

New Features:
- Surface refused gated probes and blocking requests as findings on the affected records.
- Reject promotion edits outside confirmed events and forged observations with invalid hashes.
- Prevent negated supersession phrases from incorrectly demoting documents.

Bug Fixes:
- Make extraction explain when a declared probe is refused instead of implying that no probe was offered.

Enhancements:
- Remove unused record steering APIs and the redundant validate_records wrapper.
- Align validation and evidence handling with the production loading and proposal-generation paths.

Documentation:
- Clarify promotion edit semantics and gated-probe behavior.

Tests:
- Add coverage for gated-probe refusals, record-level blocking findings, invalid evidence hashes, restricted promotion edits, and negated supersession markers.

### DIFF
--- a/crates/stella-cli/src/ingest_cmd/extract.rs
+++ b/crates/stella-cli/src/ingest_cmd/extract.rs
@@ -920,6 +920,7 @@ fn build_proposal(
         .probe
         .clone()
         .filter(|p| gate::probe_honored(Origin::Imported, p.kind));
+    let refused_as_gated = claim.probe.is_some() && gated_probe.is_none();
     // And keep only probes that could refute the claim they are attached to. A
     // `path_exists` on a cardinality claim goes green whatever the count is
     // (#4262), so it is declined and the claim abstains instead.
@@ -1003,6 +1004,14 @@ fn build_proposal(
     // its validation or quarantine finding.
     let refutation = outcome.is_eligible().then(|| match &honored_probe {
         Some(p) => probe::evaluate(root, p, observed_at),
+        // The refusal is stated: "no probe could judge" would read as an
+        // unfalsifiable claim, when the truth is that a probe was offered and
+        // the gate declined to run it.
+        None if refused_as_gated => abstained(
+            observed_at,
+            "the probe offered runs a command or reaches the network, which is \
+             never honored on imported content",
+        ),
         None if declined_as_indiscriminate => abstained(
             observed_at,
             "the claim states a count, and the probe offered for it could not \
--- a/crates/stella-cli/src/ingest_cmd/extract/tests.rs
+++ b/crates/stella-cli/src/ingest_cmd/extract/tests.rs
@@ -169,10 +169,18 @@ fn a_gated_probe_is_stripped_and_the_claim_is_unfalsifiable() {
     // The gated probe never reaches the record.
     let truth = proposal.record.truth.as_ref().expect("truth");
     assert!(truth.probe.is_none(), "a gated probe must be stripped");
-    // With no honored probe, the claim is honestly unfalsifiable.
+    // With no honored probe, the claim is honestly unfalsifiable — and the
+    // reviewer is told the probe was refused rather than that none was offered.
     let refutation = proposal.refutation.as_ref().expect("refutation");
     assert_eq!(refutation.verdict, Verdict::Unfalsifiable);
     assert_eq!(refutation.probe_kind, ProbeKind::None);
+    assert!(
+        refutation
+            .detail
+            .contains("never honored on imported content"),
+        "the reviewer is not told the probe was refused: {}",
+        refutation.detail
+    );
     let _ = std::fs::remove_dir_all(&root);
 }
 
--- a/crates/stella-cli/src/memory/proposals.rs
+++ b/crates/stella-cli/src/memory/proposals.rs
@@ -156,6 +156,12 @@ pub(crate) fn induce_proposals(
         } else {
             RecordProposalStatus::Collecting
         };
+        // An observation whose hash does not match its content is evidence
+        // nothing vouches for; a candidate resting on one is skipped the same
+        // way an unscorable one is.
+        let Ok(pool) = EvidencePool::from_observations(supporting) else {
+            continue;
+        };
         let Ok(proposal) = ProposalRecord::new(
             RecordProposalKind::Knowledge,
             status,
@@ -163,7 +169,7 @@ pub(crate) fn induce_proposals(
             &candidate.description,
             &candidate.body,
             candidate.domains.clone(),
-            EvidencePool::from_observations(supporting),
+            pool,
             score,
             confidence,
             proposal_observed_at(&candidate),
--- a/crates/stella-cli/src/memory/rules_mining.rs
+++ b/crates/stella-cli/src/memory/rules_mining.rs
@@ -145,6 +145,12 @@ pub(crate) fn induce_rule_proposals(
         } else {
             RecordProposalStatus::Collecting
         };
+        // An observation whose hash does not match its content is evidence
+        // nothing vouches for; a candidate resting on one is skipped the same
+        // way an unscorable one is.
+        let Ok(pool) = EvidencePool::from_observations(supporting) else {
+            continue;
+        };
         let Ok(proposal) = ProposalRecord::new(
             RecordProposalKind::Directive,
             status,
@@ -152,7 +158,7 @@ pub(crate) fn induce_rule_proposals(
             &candidate.description,
             &candidate.text,
             Vec::new(),
-            EvidencePool::from_observations(supporting),
+            pool,
             score,
             confidence,
             stella_context::format_rfc3339(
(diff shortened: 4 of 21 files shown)


macanderson:
Answering the Sourcery ❌ ("does not address `blocking_approved` or `detect_tool_gaps`"): both are deliberately out of this PR's scope because each already has its resolution elsewhere, and #5331's Definition of done says so per item.

- `records::decision::blocking_approved`: the substring-match defect this issue filed was fixed on `main` by #5400 before this PR branched — the fold now carries `lineage_id` and the match is by identity, with the prefix-leak witness `an_approval_does_not_leak_to_a_prefix_lineage`. Re-doing or deleting it here would revert a deliberate, fresher decision.
- `tool_foundry::detect_tool_gaps`: PR #5476 (in flight, ADR 0023 — Option A of #5433) gives the detector its live caller and owns `crates/stella-core/src/tool_foundry.rs`; touching that file here would conflict with it, and once it lands the stale doc claims this issue names stop being claims about a caller that does not exist.

The issue's DoD boxes record both handoffs explicitly, so closing #5331 on this PR does not orphan either item.

Issue #3715, opened 2026-08-18
stella-engine facade re-exports Engine::with_requery but not the SteeringRequery port or TurnSignal it requires


macanderson:
## Problem

`crates/stella-engine/src/lib.rs` re-exports `stella_core`'s `Engine` type whole, including its builder method `with_requery`. That method's signature, in `crates/stella-core/src/driver.rs:570`, is:

```rust
pub fn with_requery(mut self, requery: &'a dyn crate::ports::SteeringRequery)
```

`SteeringRequery` is defined in `crates/stella-core/src/ports.rs:519`. Its `requery` method takes `&stella_core::steering::TurnSignal<'_>` as an argument.

Neither `SteeringRequery` nor `TurnSignal` is re-exported from `crates/stella-engine/src/lib.rs`. Today the crate only re-exports `ToolExecutor`, `TurnGate`, and `TurnSteering` (see `lib.rs:137`).

This breaks the closure rule the crate states for itself at `lib.rs:139-141`: if a type appears in the public signature of a port a host must implement, the host has to be able to name that type from `stella-engine` alone, without linking `stella-core` directly. A host wanting to implement `SteeringRequery` and call `Engine::with_requery` currently cannot write the method signature at all, because it cannot name either `SteeringRequery` or `TurnSignal` from this crate.

## Why the obvious one-line fix is not enough

Simply adding `SteeringRequery` to the re-export list is insufficient, because `TurnSignal` (the argument type of its one method) is also unreachable from `stella-engine`. Both need to be re-exported together. `TurnSignal`'s own fields are only `&str`, `&[String]`, and `u32` (per audit), so re-exporting it does not pull in a wider chain of undeclared types.

## The bigger question this exposes (must be settled before editing)

`Engine` has several other public builder methods that take types stella-engine does *not* re-export either: `TurnHalt`, `FallbackResolver`, `Hooks`/`HookRunner`, `HookBus`, `ProviderOutcomes`, and `ModelCallRole`.

There are two different rules floating around for when a re-export is warranted:

- `crates/stella-engine/README.md` says a re-export is earned only when a host genuinely cannot drive a turn without it.
- `crates/stella-engine/src/lib.rs:139-141` states a stricter rule: any type appearing in the public signature of a port a host implements must be re-exported.

[...]

Issue #1494, opened 2026-08-05
stella-engine cannot be embedded through its own facade


macanderson:
The drop-in-engine story fails at first contact with a host.

`crates/stella-engine/src/lib.rs:132` re-exports `Provider` but not `CompletionRequestRef`, `CompletionResult` or `CompletionUsage` — the exact types `Provider::complete_ref` names in its signature (`stella-protocol/src/provider.rs:111-114`). No host can implement the crate's primary port without also linking `stella-protocol`, which contradicts the Mode-A embedding story in `docs/spec/engine-embedding.md:95`.

The crate's own tests prove it: `crates/stella-engine/src/tests.rs:13` imports all three straight from `stella_protocol`.

Same shape at `lib.rs:122`: `EngineConfig` is re-exported and its `checkpoint_sink` field is `Option<Arc<dyn CheckpointSink>>` (`stella-core/src/driver.rs:292`), but `CheckpointSink` is not re-exported — so a durable host, the audience the crate names in its first paragraph, cannot name the trait it must implement. `stella-serve` already reaches around the facade: `use stella_core::step::CheckpointSink;` (`checkpoint.rs:71`).

A facade a first-party crate has to bypass is not yet a facade.

PR #5490, opened 2026-08-30
fix: eight correctness defects from a workspace-wide audit

### LINKED ISSUE #5489: Eleven correctness defects found by a workspace-wide audit
A correctness audit of every crate in the workspace found eleven defects small
enough to fix with a witness test each. This issue tracks that batch; everything
the audit found and did not fix is filed separately and linked from the pull
request.

## The defects

1. **`stella-home`** — an environment variable exported with an empty value is
   taken as a real path. `STELLA_HOME=` makes every user-tier path relative to
   the working directory: `credentials.toml` (provider API keys) is written into
   whatever repository is open, `usage.db` becomes one file per directory, and
   `resolve_user_plugins_dir` returns the relative `plugins` path its own
   contract forbids. `resolve_workspace_state_root` already filtered empty and
   relative values; nothing else did. `XDG_CONFIG_HOME` has the same shape, and
   a relative value there writes the systemd unit where `systemctl --user` never
   looks while the installer reports success.

2. **`stella-engine`** — four `ToolExecutor` methods carry a "Decorators MUST
   forward this" section, and the facade exported none of the types their
   signatures name. A host wrapping its own executor compiled clean and silently
   answered the trait default: parked waits dropped, the end-of-turn
   live-service assertion off, and a dispatching decorator above the tap running
   ungated.

3. **`stella-core`** — `redact_secrets` ends an unquoted value at the first
   space, so `Authorization: Bearer <token>` redacted the word `Bearer` and
   published the credential. The other rules do not catch it either: the entropy
   heuristic needs 32 characters plus all three character classes. Worst of all
   it reported `redacted: true`, because rewriting `Bearer` did change the text.

4. **`stella-core`** — a `tools:` mapping in agent frontmatter is unreadable to
   the flat parser, which leaves the key holding an empty value; an empty
   `tools:` is the documented spelling of "all tools". An author asking for two
   tools was handed the whole registry, silently. `rules.rs` already refuses
   nested keys in the same parser for the same reason.

5. **`stella-plugin`** — a panel that declares no `command` registers the
   plugin's own name, and only a declared name was checked. A plugin named
   `vera:admin` takes the namespace-shaped command
   `PanelCommandCarriesNamespace` exists to refuse, and the consent document then
   says the panel answers to `/vera:admin` and `/vera:admin:vera:admin`.

6. **`stella-tools`** — `CustomToolSet` forwards five of the six pass-through
   `ToolExecutor` methods and drops `active_skill_slugs`, so a live skill's
   procedure text stops surviving overflow summarization for any workspace with
   a custom tool installed.

7. **`stella-core`** — the operator-facing line for a still-running service said
   the engine was "asking the model to confirm or stop" it, while the message
   sent to the model says a live service cannot be stopped and the port's
   contract says asking must change nothing.

8–11. **Four `clamp(literal, variable)` panics.** `clamp` asserts `min <= max`,
   and in each the floor is a literal while the ceiling is caller- or
   operator-supplied. A workspace sweep found these four and no others:
   `stella-autonomy`'s `plan_cycle` (`SELF_DRIVING_BATCH_MAX=0` crashes
   `stella self-driving plan`); `stella-tui`'s SKILLS tab (any terminal under
   eight rows); `stella-serve`'s `SubAgentPolicy::clamp` (a `pub` field set to 0,
   and the release profile aborts, so it ends the whole server rather than the
   request); and `stella-runtime`'s `CandidateFanouts` (a public builder with no
   floor).

## Definition of done

- [x] Every defect that changes behaviour ships a witness test that fails on the
      old code and passes on the new, each checked by reverting the production
      change with the test still in place and re-running. The one exception is
      `stella-runtime`'s builder floor, which has no caller in the tree to
      witness and says so in its own commit.
- [x] `cargo test -p <crate>` green for every touched crate.
- [x] `cargo clippy -p <crate> --all-targets -- -D warnings` green for every
      touched crate.
- [x] `cargo fmt --check` green.
- [x] `make prose` green with no baseline entry added — the one density
      regression was paid for by deleting prose rather than by raising a number.
- [x] No entry added to any ratchet baseline.
- [x] Everything the audit found and did not fix is filed as its own issue,
      written as a handoff.

The pull request's own CI is the workspace-level evidence and is not restated
here as a checklist item, since the gate that reads this list runs on that same
pull request.

macanderson:
## What & why

A correctness audit of all 25 crates. Eleven defects were small enough to fix
with a witness test each, one commit per crate. Everything else the audit found
is filed rather than fixed here.

Closes #5489

### The defects

**`stella-home` — an env var exported with no value is read as a path.**
`std::env::var_os` reports `STELLA_HOME=` as `Some("")`, and every user-tier
resolver took it as a real anchor. `credential.rs` then writes
`credentials.toml` — provider API keys — into whatever repository is open,
`usage.rs` opens a different `usage.db` per working directory, and
`resolve_user_plugins_dir` returns the relative `plugins` path its own doc
comment forbids. `resolve_workspace_state_root` already filtered empty and
relative overrides; nothing else did. `any_override_set` moves with them, or it
would decline the legacy-layout migration for a process sitting on the defaults.
`XDG_CONFIG_HOME` also now refuses a relative path, which is what the XDG spec
its doc comment cites requires.

**`stella-core` — the credential after an auth scheme was published.**
`redact_secrets` ends an unquoted value at the first space, right for `KEY=value`
and wrong for `Authorization: Bearer <token>` — the shape a logged header or a
pasted curl command takes. It replaced `Bearer` and left the token. The other
rules do not catch it either: `is_high_entropy_blob` needs 32 characters plus
upper, lower and digit all present, so a short or single-case token falls
through. And because rewriting `Bearer` did change the text, the result reported
`redacted: true` while the secret was still in it — the one outcome the module's
own doc says it must never produce.

**`stella-core` — a `tools:` mapping granted every tool.** The flat frontmatter
parser cannot read a nested mapping, so it leaves the key holding an empty value
and diverts the children into `nested_keys`, which `extensions.rs` never reads.
An empty `tools:` is the documented spelling of "all tools", so an author writing
`tools:\n  read: true\n  write: false` was handed the whole registry with no
diagnostic. Tolerance is right for every other field and wrong for this one,
because this is the field where an unreadable value fails *open*. `rules.rs`
already refuses nested keys in the same parser for the same reason.

**`stella-engine` — a host cannot forward what `ToolExecutor` says it must.**
Four of the port's methods carry a "Decorators MUST forward this" section, and
the facade exported none of the types their signatures name (`WaitRequest`,
`LiveService`, `DispatchGate`, `ToolContract`). Each has a trait default, so a
host's tap compiled clean and answered the default: parked waits dropped and the
model back to polling, the end-of-turn live-service assertion off, and a
dispatching decorator above the tap running ungated. This is the closure rule
read at the method level, which the two earlier passes over this facade
(#1494, #3715) did not reach.

**`stella-tools` — `CustomToolSet` dropped one of its six forwards.**
`active_skill_slugs` fell through to the empty default, so a live skill's
procedure text stops surviving overflow summarization for any workspace that also
has a custom tool installed. Dormant today, since nothing populates the plane
yet — and dormant is how it would stay until the first executor that does.

**`stella-plugin` — a panel's derived slash name skipped every shape check.**
A panel with no `command` registers the plugin's own name, and `validate` only
checked a declared one. `name = "vera:admin"` takes the namespace-shaped command
`PanelCommandCarriesNamespace` exists to refuse, and the consent document then
tells the reader the panel answers to `/vera:admin` and to
`/vera:admin:vera:admin`. `validate` now takes the plugin id, so the question
cannot be asked without the value that answers it.

**`stella-core` — the live-service narration offered what the model was not.**
The operator's line said the engine was "asking the model to confirm or stop" a
still-running service; the message sent to the model says a live service cannot
be stopped, and the port's contract says asking must change nothing.

**Four `clamp(literal, variable)` panics.** `clamp` asserts `min <= max`, and in
each the floor is a literal while the ceiling is caller- or operator-supplied. A
workspace sweep found these four and no others:

- `stella-autonomy` `plan_cycle` — `SELF_DRIVING_BATCH_MAX=0` in the environment
  reaches `clamp(1, 0)` and crashes `stella self-driving plan`. `calibrate`, ten
  lines above, already takes the same untrusted numbers on a `min`/`max` chain
  and says why.
- `stella-tui` `views::skills` — `(hits + 4).clamp(4, area.height / 2)` panics on
  any tab under eight rows. The floor was redundant (`hits + 4` is already 4 or
  more), so it only ever supplied the panic.
- `stella-serve` `SubAgentPolicy::clamp` — `max_child_steps` is a `pub` field no
  constructor validates, so `0` panics on any request naming `max_steps`. The
  release profile sets `panic = "abort"` and the connection runs in a spawned
  task, so that ends the whole process, not the request.
- `stella-runtime` `CandidateFanouts` — `with_max_width(0)` is a public builder
  with no floor. No caller in the tree, so this is a public-API floor rather than
  a live defect, and it ships without a witness.

## The witness

- [ ] No witness needed

Every witness was checked by reverting the production change with the test in
place and re-running, then restoring. Two are worth naming:

- The first draft of the `stella-tui` witness asserted `catch_unwind` returned
  `Ok`, and **passed against the panicking code** — `guarded_band` catches a
  band's panic and paints an error card in its place. It now asserts the band's
  painted cells.
- The first version of the `stella-serve` fix dropped the floor on the
  *requested* value and broke `a_zero_step_request_floors_at_one`. The ceiling
  and the request each keep their own floor now.

The `stella-runtime` builder floor is the one change with no witness, and its
commit says so.

## The gate

- [x] `cargo fmt --check`
- [x] `cargo clippy -p <crate> --all-targets -- -D warnings` for every touched crate
- [x] `cargo test -p <crate>` for every touched crate
- [x] `make prose` — green, and it needed a real edit rather than a baseline
      entry: the `stella-engine` header gained a line for `Sleeper` (the port
      list named nine optional ports and omitted the one that is mandatory), so
      the closure-rule paragraph gave up its account of what the two copies of
      that rule used to say and the issue number standing in for the reason.
- [x] Docs updated where behaviour changed. `panel_guard`'s comment claimed no
      crafted state can panic a view, which this PR disproves.
- [x] `Closes #5489` in this description and as a commit trailer.

Not run locally, per CLAUDE.md's "CI builds and tests; this laptop does not": the
workspace build, the workspace test suite, and `make gate`. That is the right
call and it cost something — the new `ExtensionProblem` variant broke two
exhaustive matches in `stella-cli`, which a per-crate clippy cannot see and CI
caught on the first run. Fixed in `fix(stella-cli): describe the new
nested-toolbelt refusal`.

## Nothing left behind

The audit raised far more than it fixed. Every remaining finding is filed as a
handoff, `triage` label only:

Defects — #5491 (the dashboard's live stream holds a one-shot connection permit
until the tab closes, and the accept loop wedges when the pool drains), #5493
(release builds abort on panic, so every panic boundary in the tree is inert
where it ships), #5494 (the search tool's fallback scan ignores `.gitignore`),
#5495 (a password inside a provider `base_url` is served to the dashboard
unredacted), #5496 (a malformed Anthropic tool-call start drops the call and the
turn reports none), #5497 (the code-graph watcher indexes files a symlink points
at outside the workspace), #5498 (a fleet worker's cost is never checked, so NaN
wedges the budget and a negative value bypasses it), #5499 (the two transcript
renderers each drop the other's honesty marker), #5500 (five dashboard numbers
are wrong), #5501 (two sessions in one process cannot use two different Vertex
projects), #5502 (`edit_file` writes from a stale snapshot, destroying a
concurrent write), #5503 (a failed checkpoint delete answers `200 deleted`),
#5504 (`GateBoard` renders unlabelled on the plain transcript, plus four false
README claims), #5505 (store corruption found after open loses its remedy),
#5508 (editing a memory from the Command Deck panics the session), #5509 (a
synthesized tool template drops a shell redirect), #5510 (an abandoned lease
makes the telemetry spool grow without bound), #5511 (five smaller findings),
#5516 (several turn aborts hand back a transcript the next provider call
rejects).

Next steps — #5512 (make a decorator that forgets to forward fail loudly),
#5513 (a guard for `clamp` with a literal floor and a computed ceiling), #5514
(property tests for compaction, eviction and budget), #5515 (check a consumer
ledger site still points at live code, while the ledger is clean).

## Ground-rule check

**This is one PR for nine crates rather than nine PRs.** The commits are one per
crate and touch disjoint files, so a rebase-merge preserves them and any one is
revertable alone; a squash-merge collapses them into a single commit spanning
nine crates. Say the word and I will split it.

The four `clamp` fixes all order the operations `min` then `max`, so the floor
wins over a nonsense zero ceiling. That is the reading each site's own comment
already argued for a zero *input*; this extends it to a zero *limit*.

Two of the eleven are the same shape one layer apart — a decorator silently
taking a trait default because no compiler can see the omission (`stella-engine`
and `stella-tools`). #5512 proposes the guard that would catch the next one.

## Summary by Sourcery

Harden workspace-wide correctness by fixing unsafe path resolution, secret redaction, tool and panel validation, decorator forwarding, user-facing narration, and panic-prone limit handling.

Bug Fixes:
- Prevent empty or invalid home and XDG overrides from redirecting user data into the working directory.
- Redact credentials that follow common authentication schemes instead of exposing the credential token.
- Reject nested `tools:` mappings that would otherwise silently grant every tool.
- Expose the complete `ToolExecutor` forwarding surface through the engine facade.
- Forward active skill slugs through custom tool sets.
- Validate panel names derived from plugin identifiers.
- Align live-service narration with the actions actually available to the model.
- Replace panic-prone limit handling across autonomy, runtime, serving, and TUI paths with safe flooring behavior.

Enhancements:
- Improve diagnostics for unsupported nested tool mappings and clarify engine port documentation.
- Update affected behavioral documentation and add targeted regression coverage for the audited defects.

Documentation:
- Document the mandatory sleeper port, corrected panel guard behavior, and updated closure-rule guidance.

Tests:
- Add witness tests covering environment overrides, authentication redaction, nested tool mappings, facade forwarding, panel command validation, live-service narration, and zero-limit handling.

### DIFF
--- a/crates/stella-autonomy/src/lib.rs
+++ b/crates/stella-autonomy/src/lib.rs
@@ -625,8 +625,15 @@ pub fn plan_cycle(
     // `Deserialize` with no range guard), and `AimdLimits` promises bounds
     // "the controller may never cross, whatever the evidence says" — so the
     // plan clamps here rather than trusting every load site to have done it.
-    let mut batch = ceilings.batch_ceiling.clamp(1, limits.batch_max);
-    let mut parallel = ceilings.parallel_ceiling.clamp(1, limits.parallel_max);
+    // `min` then `max`, not `clamp`: `u32::clamp` asserts `min <= max` and
+    // panics otherwise, and the ceiling here is the literal 1 while the limit
+    // is caller-supplied. `SELF_DRIVING_BATCH_MAX=0` in the environment reaches
+    // this as `clamp(1, 0)` and takes down the process before any branch below
+    // runs. Ordering the other way keeps the floor winning over a limit of 0,
+    // which is what a cycle of nothing would otherwise mean, and matches how
+    // `calibrate` above stays panic-free on the same untrusted numbers.
+    let mut batch = ceilings.batch_ceiling.min(limits.batch_max).max(1);
+    let mut parallel = ceilings.parallel_ceiling.min(limits.parallel_max).max(1);
 
     let reason: String;
     if supply.disk_free_gb < floors.disk_gb {
--- a/crates/stella-autonomy/src/tests.rs
+++ b/crates/stella-autonomy/src/tests.rs
@@ -357,6 +357,31 @@ fn assert_plan(supply: Supply, demand: Demand, want: (&str, bool, u32, &str, &st
     assert_eq!(got, want, "reason: {}", plan.reason);
 }
 
+/// The limits come from the environment, so a plan must survive a nonsense
+/// one rather than taking the process down.
+///
+/// `SELF_DRIVING_BATCH_MAX=0` — an operator trying to switch batching off —
+/// reached `u32::clamp(1, 0)`, which asserts `min <= max` and panics before
+/// any of the supply branches run. A limit of zero still means at least one
+/// unit of work, the same floor every other path applies.
+#[test]
+fn a_zero_limit_floors_the_plan_instead_of_panicking() {
+    let zeroed = AimdLimits {
+        batch_max: 0,
+        batch_min: 0,
+        parallel_max: 0,
+    };
+    let plan = plan_cycle(
+        base_supply(),
+        Demand::default(),
+        &ceilings(),
+        Floors::default(),
+        &zeroed,
+    );
+    assert_eq!(plan.batch, 1, "a cycle of nothing is not a plan");
+    assert_eq!(plan.parallel, 1);
+}
+
 #[test]
 fn the_supply_rungs_decide_the_tier() {
     let d = Demand::default();
--- a/crates/stella-cli/src/agents_installed.rs
+++ b/crates/stella-cli/src/agents_installed.rs
@@ -458,6 +458,8 @@ pub fn parse_generated_agent(text: &str) -> Result<GeneratedAgent, String> {
                 stella_core::extensions::ExtensionProblem::MissingName => "no usable name",
                 stella_core::extensions::ExtensionProblem::EmptyBody => "empty body",
                 stella_core::extensions::ExtensionProblem::Malformed => "not valid TOML",
+                stella_core::extensions::ExtensionProblem::NestedToolbelt =>
+                    "a nested `tools:` mapping, which would have granted every tool",
             }
         )
     })?;
--- a/crates/stella-cli/src/extensions.rs
+++ b/crates/stella-cli/src/extensions.rs
@@ -448,6 +448,10 @@ fn describe_diagnostic(diag: &stella_core::extensions::ExtensionDiagnostic) -> S
         stella_core::extensions::ExtensionProblem::MissingName => "no usable name",
         stella_core::extensions::ExtensionProblem::EmptyBody => "empty body",
         stella_core::extensions::ExtensionProblem::Malformed => "not valid TOML",
+        stella_core::extensions::ExtensionProblem::NestedToolbelt => {
+            "`tools:` is a nested mapping, which would have granted every tool — \
+             write it as a list, like `tools: read, search`"
+        }
     };
     format!("{}: {why}", diag.path)
 }
--- a/crates/stella-core/src/driver/live_services.rs
+++ b/crates/stella-core/src/driver/live_services.rs
@@ -123,8 +123,12 @@ pub(super) fn check(
     }
     let _ = events.send(AgentEvent::Text {
         text: format!(
+            // Not "confirm or stop": the nudge below offers keeping it and
+            // owning it as a leftover, and says a live service cannot be
+            // stopped. An operator told otherwise waits for a shutdown the
+            // model was never offered.
             "\n⚠ {} long-running {} still up at the end of this turn — asking the model to \
-             confirm or stop {}.",
+             account for {}.",
             services.len(),
             if services.len() == 1 {
                 "process is"
@@ -166,6 +170,33 @@ mod tests {
             name: name.map(str::to_string),
             display: "python3 -m http.server 8080".into(),
         }
+    }
+
+    /// The operator's line offers only what the model is offered.
+    ///
+    /// The nudge says plainly that a live service cannot be stopped, and the
+    /// port's own contract says asking must change nothing — least of all stop
+    /// anything. The narration said "confirm or stop", so an operator reading
+    /// the event stream waited for a shutdown the model was never offered and
+    /// the engine cannot perform.
+    #[test]
+    fn the_narration_offers_only_what_the_nudge_offers() {
+        let (tx, mut rx) = tokio::sync::mpsc::unbounded_channel();
+        let mut messages = vec![CompletionMessage::user("serve the docs on 8080")];
+        assert!(check(
+            &mut messages,
+            &[service("proc-1", Some("docs"))],
+            "Done — the server is running.",
+            &EventSender::new(tx),
+        ));
+        let AgentEvent::Text { text } = rx.try_recv().expect("the check narrates what it did")
+        else {
+            panic!("the narration is a Text event");
+        };
+        assert!(
+            !text.contains("stop"),
+            "the operator must not be told a service can be stopped: {text}"
+        );
     }
 
     /// **The #2764 witness.** A turn that leaves a service up is asked about
--- a/crates/stella-core/src/extensions.rs
+++ b/crates/stella-core/src/extensions.rs
@@ -193,6 +193,17 @@ pub enum ExtensionProblem {
     /// either parses or does not, and guessing at a broken one would load a
     /// command whose author cannot tell it is broken.
     Malformed,
+    /// `tools:` was written as a nested mapping, which the flat frontmatter
+    /// parser cannot read.
+    ///
+    /// The tolerance every other field enjoys is wrong for this one, because
+    /// the unreadable value does not degrade to a smaller grant — it degrades
+    /// to `None`, which means *every* tool. An author writing
+    /// `tools:\n  read: true\n  write: false` asked for a restriction and
+    /// would have been handed the whole registry, with no diagnostic. So the
+    /// definition is refused instead, the way `rules.rs` refuses a record
+    /// carrying nested keys.
+    NestedToolbelt,
 }
 
 /// A malformed definition file, skipped during loading (never fatal).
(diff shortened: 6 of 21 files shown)


macanderson:
Settling both ❌ rows from the assessment against #5489.

**Row 2 — "a witness test for each of the eight fixes."** Fixed, not answered.
You were right and my reasoning was wrong. I had shipped the `stella-runtime`
`CandidateFanouts` floor without a witness on the grounds that
`with_max_width(0)` has no caller in the tree — but "no caller today" is not the
same as "not witnessable". The builder is public, `tests/wrapper_candidate_fanout.rs`
already constructs the plane, and the assertion turned out to be two lines.
`a_zero_host_width_ceiling_floors_the_fanout_instead_of_panicking` (commit
`aeff223`) panics at the old `clamp` and passes on the `min`/`max` pair. Every
fix in the batch now has a witness, each verified by reverting the production
change with the test in place and re-running.

**Row 3 — "verify the touched crates and workspace gates."** Answered, with one
part fixed.

- *No ratchet baseline added* — now stated explicitly in #5489's Definition of
  Done rather than left implied. It is checkable from the diff: no file under
  `scripts/*-baseline.txt` is touched. `make prose` needed a real edit rather
  than a baseline entry, and the commit that made it says so — the
  `stella-engine` header gained a line for `Sleeper`, so the closure-rule
  paragraph gave up its account of what the two copies of that rule used to say.
- *Workspace build, workspace test suite and `make gate` not run locally* —
  that is deliberate and required here, not an omission. CLAUDE.md's standing
  rule is "CI builds and tests; this laptop does not", and a PR's evidence is
  its CI run. Running the workspace suite on the maintainer's machine is the
  thing the rule forbids.

  It is worth being blunt that the rule cost something on this PR: the new
  `ExtensionProblem` variant broke two exhaustive matches in `stella-cli`, which
  a per-crate `cargo clippy -p stella-core` cannot see. CI caught it on the first
  run and it is fixed in `356dfa8`. That is the trade working as intended, not an
  argument that the local run should have happened.
- *"provides no demonstrated CI result"* — true when you assessed it, since the
  run was still in flight. The checks are the demonstration and they are on this
  PR.

I have also removed the circular item from #5489's Definition of Done. It read
"CI green on the pull request", which the DoD gate — running on that same pull
request — cannot ever see satisfied. The remaining items are all verifiable
before the merge, which is what that checklist is for.

One thing the assessment could not know: #5489 now covers **eleven** defects
rather than eight. Three more were found and fixed after it was filed — a
credential published by the secret redactor after an auth scheme, a nested
`tools:` mapping that granted every tool, and an operator-facing line offering a
"stop" the model is told is impossible. The issue body and the PR description
both name all eleven.

PR #5506, opened 2026-08-30
feat(stella-tui): the chevron carries intent — gold dispatches, teal steers, red interrupts

macanderson:
## What & why

The deck's prompt line gets three intents, told apart by the chevron's color, so steering a running agent stops being a hidden `>` marker and becomes the line's visible default while a turn runs:

- **Gold — dispatch** (the identity default): today's routing, `ui.mid_turn_prompt` included. Unchanged behavior.
- **Teal — steer**: the text goes into the running turn at its next step boundary (`WorkspaceInput::Steer` → the lane's `SteeringTap` → a real user message before the next model call). The palette has no plain blue; teal is the categorical tone the palette already separates from gold, and gold stays identity-only per the brand rule.
- **Red — interrupt** (new `WorkspaceInput::Interrupt`): a soft stop — the same lossless first-Esc stop, every completed step kept — with the words front-queued ahead of the parked backlog so they run next. Never the hard cancel, and never mid-tool: injection and stopping stay at the step boundary.

**Shift-Tab cycles the modes while the composer holds a draft; with an empty composer it keeps its deck job of cycling tabs** — the collision resolves on what the user is visibly doing (`composer_mode::backtab`, claimed ahead of tab-cycling). A cycled override lasts one submission.

**The mode is derived, never latched** (`composer_mode::effective`): with no override the chevron follows the session — and under the new policy value `ui.mid_turn_prompt = "steer"` it turns teal exactly while the focused agent runs, so the natural next thing typed is a correction rather than a silently queued prompt. The default policy (`queue`) keeps today's behavior byte-for-byte.

Driver side: `steer::interrupt_lead` claims + front-inserts the words, latches the soft stop, and releases a paused turn so the boundary is reachable; an interrupt at a worker lane lands as a steer (a lane's backlog is the lead's, so there is nothing to outrank), and one that reaches an idle agent degrades to "run this now" through the existing `steer_at_rest` path.

New code lands in sibling modules (`deck_ui/composer_mode.rs`, `command_deck/steer.rs`); `deck_ui.rs` ends 2 lines *under* its grandfathered ceiling (the single-use `slash_matches` helper was inlined to pay for the hook), and `command_deck.rs` grows well inside the slack its earlier splits earned.

Exemplars followed: `caret_style` for the pure, unit-tested style decision (`prompt_prefix_style`); the `steer.rs` sibling for driver logic out of the god file; `MidTurnPrompt`'s own parse-fails-to-default contract for the new settings slug.

[...]

PR #5517, opened 2026-08-30
feat(stella-cli): a context record changed mid-session reaches the next step boundary

### LINKED ISSUE #5432: A context record changed mid-session is invisible until the session restarts
## Problem

A context record edited, added, promoted, or retired while a session is running is invisible to that session until restart. The registry is loaded once at session start (`crate::rules::enforce_workspace_rules` → `SessionMemory::set_record_registry` holds a clone for the session's life) and every later turn reads the same clone. The retrieval side already re-decides per re-query — `signal_recall_block` calls `render_volatile_for_turn_excluding` against the turn's drifting paths, and `SteeringSource::Record` outranks every other source in `pack_to_budget` — so the gap is exactly one stale clone.

This matters most for fleets: a user running 10–15 sessions who promotes a rule ("stop touching the release branch") wants it to reach the sessions already running, through the normal retrieval workflow, without whistling prose at each one.

Plan (full context in the plan comment on #5331):

1. **A watcher, modeled on the code graph's.** `SessionGraph` (`crates/stella-cli/src/agent/graph.rs`) is the in-tree pattern: a session-lifetime `notify` watcher with debounce and a shutdown handle. A sibling `SessionRecords` watches `.stella/rules/` (and the user-tier rules dir), debounces, re-runs `load_registry(root)` — which re-verifies governance and the promotion ledger and fails closed, same as start-up — and swaps via `set_record_registry`. `SessionMemory` owns the `Registry`, so this is a swap, not a plumbing change.
2. **Force one re-query on swap.** The re-query hysteresis (`memory/steering.rs::SessionRequery`) fingerprints touched paths + error classes, so a registry swap alone would never trigger a re-query until the turn happens to drift. Fold a registry generation counter into the fingerprint so a swap guarantees the next step boundary re-queries.
3. **The cached-prefix wall, handled honestly.** `must`/`should` records render into the byte-stable system prefix (`agent/prompt.rs`); swapping that mid-session is a guaranteed prompt-cache miss and is forbidden (AGENTS.md invariant #7). When a swap detects a prefix-channel change: deliver the record body on the volatile channel for the rest of the session, and inject a one-line steering notice ("blocking record `<handle>` changed; it binds into the prefix next session; applied as volatile guidance until then"). No cache bust, no silent staleness.
4. **Retirement propagates too.** A record retired mid-session must leave the volatile channel on the same swap, so suppression is one mechanism in the running session as well (#2284's concern, live).

Out of scope: a promote-time whistle nudge for sessions the filesystem watcher cannot serve (a served turn, eventually cross-machine) — that follows #4770's discovery work.

## Files

- `crates/stella-cli/src/agent/graph.rs` — `SessionGraph`, the watcher pattern to copy
- `crates/stella-cli/src/context_records.rs` — `load_registry` (already fails closed on ledger/governance problems)
- `crates/stella-cli/src/memory/recall.rs` — `set_record_registry`, `signal_recall_block`
- `crates/stella-cli/src/memory/steering.rs` — `SessionRequery`, the fingerprint
- `crates/stella-cli/src/agent/prompt.rs` — the prefix render the swap must NOT touch
- Wiring sites: `command_deck.rs`, `agent.rs`, `subsession.rs` (each currently calls `enforce_workspace_rules` once) — all god-file-adjacent; new code lands in siblings

## Constraints

- The byte-stable system prefix never changes mid-session (invariant #7); only the volatile channel may carry the update.
- `load_registry`'s fail-closed behavior on a broken ledger or governance file is preserved on re-scan: a corrupt edit mid-session keeps the last good registry and says so, it does not silently drop steering.
- Design the swap door so a unified memories-under-records plane (#2283) inherits it rather than needing a second mechanism.

## Definition of done

- [x] A record file added or edited mid-session appears in the volatile channel at the next step boundary; witness: edit a record during a running turn, assert the re-query block carries it
- [x] A record retired mid-session leaves the volatile channel on the same swap (witness in the other direction)
- [x] A registry swap forces exactly one re-query via the generation-bumped fingerprint; no swap, no extra query (hysteresis witness preserved)
- [x] A prefix-channel (`must`/`should`) change produces the volatile fallback plus the one-line notice, and the system prefix bytes are unchanged for the session's remainder (prompt-cache golden stays green)
- [x] A mid-session edit that stops parsing keeps the last good registry and reports the failure visibly (`a_rules_file_that_stops_parsing_keeps_the_last_good_registry`); a broken ledger or governance file is `load_registry`'s existing fail-closed path and #5328's subject

macanderson:
## What & why

A context record edited, added, promoted, or retired while a session runs now reaches that session at its next step boundary, through the retrieval path that already exists, instead of waiting for a restart.

The registry was loaded once by `SessionMemory::open_for_session` and held as a clone for the session's life. The retrieval side re-decides which records render on every re-query, so the gap was exactly one stale clone. The fix (`crates/stella-cli/src/memory/records_refresh.rs`):

- **Lazy freshness, no watcher.** At the two moments the answer can matter — the turn-opening recall and each re-query boundary — the session digests the rule files' bytes (both trust tiers, the same file set `load_registry` parses) and reloads only when they moved. The unchanged path is one read of a handful of small policy files; there is no background task and no platform file-event quirk to get wrong. This is deliberately not the `notify` shape the issue sketched: a watcher adds a thread, a debounce, and an event model for something a digest at a boundary answers exactly.
- **A swap forces exactly one re-query.** The registry swap counter (`records_generation`) folds into the re-query fingerprint, which previously tracked only touched paths and error classes and so would never have noticed a rules edit until the turn happened to drift.
- **The prefix wall, handled as designed.** `must`/`should` records render into the byte-stable system prompt, which never changes mid-session. A pinned change rides the next recall block as guidance, led by a section stating it binds into the prompt — and arms any tool guards — next session. The note is one-shot.
- **Retirement propagates.** A deleted or retired record leaves the volatile channel on the same swap, so suppression is one mechanism in the running session too.
- **A half-saved edit keeps the last good registry.** A rules file that stops parsing does not un-steer the session: the loaded registry stays in effect and the next block says the file does not parse. The comparison is source-based (a source that contributed records before, contributes none now, and carries a diagnostic), so a deleted file reads as retirement rather than damage.

`SessionMemory`'s registry moves behind an `RwLock` because the freshness check runs where the turn holds the memory by `&`; the two render paths clone the registry out of the lock rather than borrowing through it, so a concurrent swap is never blocked by a selection pass.

Exemplars: the digest-then-reload shape is `stella_graph::CodeGraph::mount`'s own catch-up (byte-identical files are skipped), applied to a file set small enough that the digest *is* the catch-up; the one-shot note follows `RecalledBlock`'s existing section assembly.

`crates/stella-cli/src/memory/tests/records_refresh.rs`:

- `a_record_added_mid_session_reaches_the_next_recall_block` — the record renders and the generation bumped
- `a_record_deleted_mid_session_leaves_the_volatile_channel`
- `a_rules_file_that_stops_parsing_keeps_the_last_good_registry` — the old record still renders and the failure is reported
- `a_pinned_record_change_is_delivered_as_guidance_with_its_caveat` — the note appears once and not on the following block

`memory/steering.rs::fingerprint_tests::a_registry_swap_moves_the_fingerprint_alone` pins the forced re-query.

The prompt-cache goldens (`memory/tests/golden_block.rs`) are untouched: nothing here writes to the system prefix.

Two DoD notes against #5432, stated plainly: the issue sketched a `notify` watcher and this lands a boundary-time digest instead, for the reasons above (the outcome the DoD asks for — pickup at the next step boundary — is what the witnesses pin); and the "corrupt ledger / bad governance file" case is covered by `load_registry`'s existing fail-closed behaviour plus the parse-failure witness — a governance file that fails to parse is #5328's subject and its own guard.

No `#[test]` was deleted.

## Test evidence

Targeted runs (SCR-001): `cargo test -p stella-cli memory` (334 passed) green, `make prose` clean, `cargo fmt` applied. CI's workspace run is the full evidence.

Closes #5432

## Summary by Sourcery

Deliver boundary-aware prompt interruption and context-record freshness while making composer intent explicit in the TUI.

New Features:
- Add composer modes for dispatching, steering, and interrupting prompts, with Shift-Tab cycling and color-coded intent indicators.
- Support a steer mid-turn policy that delivers prompts at the next step boundary.
- Refresh context records at session boundaries so edits, additions, retirements, and parse failures are reflected without restarting.

Bug Fixes:
- Ensure interrupt prompts stop the lead turn softly while preserving completed work and run ahead of parked prompts.
- Keep the last valid context-record registry active when a rules file becomes temporarily invalid.

Enhancements:
- Force a single re-query when the context-record registry changes and report pinned or unparsable rule changes through one-shot recall guidance.

Documentation:
- Document the steer policy and composer intent cycling, including the updated configuration option.

Tests:
- Add coverage for interrupt routing, composer mode behavior, mode-specific rendering, and mid-session context-record refresh scenarios.

### DIFF
--- a/crates/stella-cli/src/command_deck.rs
+++ b/crates/stella-cli/src/command_deck.rs
@@ -1127,8 +1127,13 @@ pub async fn run_deck_session(
                         continue 'session;
                     }
                     // A steer at a worker, or one whose lead turn ended
-                    // before this recv read it — see `steer::steer_idle`.
-                    Some(WorkspaceInput::Steer { agent, texts }) if !texts.is_empty() => {
+                    // before this recv read it — see `steer::steer_idle`. An
+                    // interrupt landing here has nothing to stop, so it takes
+                    // the same "run this now" path.
+                    Some(
+                        WorkspaceInput::Steer { agent, texts }
+                        | WorkspaceInput::Interrupt { agent, texts },
+                    ) if !texts.is_empty() => {
                         match steer::steer_idle(&agent, &subs, &mut queue, texts, &in_tx) {
                             Some(first) => {
                                 dispatch.release();
@@ -1994,7 +1999,13 @@ pub async fn run_deck_session(
                         // Esc with something to say — see `steer`.
                         Some(WorkspaceInput::Steer { agent, texts }) if agent == LEAD =>
                             steer::steer_lead(&steering, &mut queue, texts),
-                        Some(WorkspaceInput::Steer { agent, texts }) =>
+                        // The composer's red mode: stop at the boundary, run the
+                        // words next. At a worker lane it lands as a steer — see
+                        // the envelope's `Interrupt` doc.
+                        Some(WorkspaceInput::Interrupt { agent, texts }) if agent == LEAD =>
+                            steer::interrupt_lead(&steering, &lead_pause, &mut queue, texts, &in_tx),
+                        Some(WorkspaceInput::Steer { agent, texts }
+                            | WorkspaceInput::Interrupt { agent, texts }) =>
                             steer::steer_worker(&subs, &mut queue, &agent, texts, &in_tx),
                         // An explicit front-insert stays a front-insert even
                         // if a turn started before it arrived — the deck's
--- a/crates/stella-cli/src/command_deck/steer.rs
+++ b/crates/stella-cli/src/command_deck/steer.rs
@@ -108,6 +108,33 @@ pub(super) fn steer_lead(tap: &SteeringTap, queue: &mut DurableQueue, texts: Vec
     }
 }
 
+/// Deliver the composer's red mode at the LEAD: the words are claimed and
+/// front-inserted so they run next — ahead of the parked backlog, which is
+/// what "interrupt" outranks — and the turn is asked to stop at its next
+/// boundary, keeping every completed step. Front-insertion happens before the
+/// stop is requested, so a turn that settles between the two still dispatches
+/// the words rather than losing them behind an older backlog.
+pub(super) fn interrupt_lead(
+    tap: &SteeringTap,
+    lead_pause: &super::lead_control::LeadPause,
+    queue: &mut DurableQueue,
+    texts: Vec<String>,
+    in_tx: &UnboundedSender<Inbound>,
+) {
+    for text in &texts {
+        claim(queue, text);
+    }
+    for text in texts.into_iter().rev() {
+        queue.push_front(unmarked(&text).to_string());
+    }
+    tap.request_soft_stop();
+    // A paused turn can never reach the stop's boundary on its own.
+    lead_pause.release_for_soft_stop();
+    let _ = in_tx.send(super::chrome_note(
+        "stopping at the next step boundary — your prompt runs next".to_string(),
+    ));
+}
+
 /// A plain stop (the first Esc) at a lead with prompts still parked: deliver
 /// the backlog into the running turn and keep it going. `false` when there
 /// was nothing parked, so the caller's soft stop still runs.
--- a/crates/stella-cli/src/command_deck/tests.rs
+++ b/crates/stella-cli/src/command_deck/tests.rs
@@ -444,6 +444,44 @@ fn stray_stop_and_hold_is_a_no_op() {
 
 // ISSUES tab: entity-hit assembly
 
+/// **The witness for the composer's red mode.** An interrupt at the lead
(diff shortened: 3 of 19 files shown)


macanderson:
Superseded by #5523. This branch was accidentally cut from the composer-modes branch rather than from main, so its diff carried #5506's changes as well; #5523 is the same hot-reload commit on a clean base. No force-push — the replacement is a fresh branch.

Issue #5520, opened 2026-08-30
A delegate sub-agent cannot be steered — only stopped


macanderson:
## Problem

In-process `delegate` children cannot be steered at all. `ChildSteering::drain_steering` (`crates/stella-core/src/subagent.rs`) returns an empty list by design: `TurnSteering::drain_steering` is destructive, so a child that inherited the parent's tap would eat a message the user addressed to the parent — the documented "message thief" refusal. The soft stop propagates; guidance does not.

That leaves one steering surface unreachable by every path — `>` in the deck, `stella whistle --deep` (#5431), and `POST /v1/turns/{id}/steer` all stop at the parent's turn. A user watching a delegate go wrong mid-turn can only stop it.

## Sketch

Addressing, not inheritance:

- A steering entry carries an optional address: the `sub_agent_id` that `child_sender` already stamps on `StepUsage` / `UsageIncomplete` / `ToolStart` / `ToolResult` (#4383, #4624). `TurnSteering` gains a `drain_steering_for(sub_agent_id: Option<&str>)` (or the entries carry the address and the parent's `drain_steering` filters), keeping the existing unaddressed behavior byte-identical: an entry with no address is the parent's and only the parent drains it.
- `ChildSteering` drains only entries addressed to its own id — never the parent's, never a sibling's.
- Host-side push (`Whistleable`, `SteeringTap`, the serve `SteerQueue`) grows a way to name a child. The deck already shows delegates in the SUB-AGENTS overlay; a `>` typed at an opened delegate row is the natural address, mirroring what a worker lane gets today (`dispatch::route`'s lane branch).

## Files

- `crates/stella-core/src/subagent.rs` — `ChildSteering`, `Engine::with_turn_controls`
- `crates/stella-core/src/ports.rs` — `TurnSteering`
- `crates/stella-core/src/driver/step_boundary.rs` — `consult_hosts`, the one site allowed to claim `SteerCause::User`
- `crates/stella-cli/src/subsession.rs` — `SteeringTap`; `crates/stella-cli/src/whistle/tap.rs` — `Whistleable`
- `crates/stella-tui/src/deck_ui/dispatch.rs` — the route for a prompt typed at an opened delegate row

## Constraints

- This is an agent-loop seam (AGENTS.md, "the agent-systems architecture is the crown"): determinism and the step-boundary contract are not negotiable. Injection stays strictly between model calls.
- The unaddressed path must not change: a message with no address reaches only the parent, pinned by a witness.
[...]

Issue #4806, opened 2026-08-25
Epic: stella whistle reaches every live session


macanderson:
## Outcome

`stella whistle "<message>"` reaches every live session on the machine,
whichever surface is driving it. A user steering a running agent should not
have to know whether that agent is a headless `stella run`, the Command Deck,
a `stella goal` arc, or a turn hosted by `stella-serve`.

## Where it reaches today

Whistle delivers to sessions whose turn goes through
`crate::agent::run_turn` — `stella run`, `stella --plain` chat's REPL, and
wrapper-plugin turns. Three surfaces were left out deliberately, each for its
own reason, and each was left rather than taken quietly:

- **The Command Deck** mints its `SteeringTap` fresh per turn inside
  `command_deck.rs`, and that file is a grandfathered god file closed to
  growth. A listener needs a session-scoped publication point that survives
  between turns, and there is no room to add one where the tap is built
  (#4768).
- **`stella goal`** drives `Engine::run_goal` through a separate call path.
  Its session already announces into `SessionPresence`, so it shows up in
  `stella resume --list`; spawning a listener with nothing draining it would
  make `stella whistle` report "delivered" for a message nothing reads (#4769).
- **`stella-serve`** is a separate binary with its own transport
  (`POST /v1/turns/{id}/steer`). Reaching it needs discovery as well as
  delivery, since a served session may not register in `SessionRegistry` at
  all (#4770).

## Definition of done

- Each of the three surfaces above accepts a whistle, with a witness test in
  the shape of `a_message_sent_over_the_socket_reaches_the_tap_and_is_acked`:
  whistle a running session from another process and assert
  `AgentEvent::Steered` reaches its event stream and the next model call sees
  the injected text.
- `stella whistle` never reports "delivered" to a session with nothing
  draining the message. A surface that cannot yet receive one says so.
- The deck's wiring lands without adding lines to `command_deck.rs`, or with a
  separately reviewed `file-size-update` that argues for the crossing in
  review.

## Children

#4768, #4769, #4770

## Constraints

- AGENTS.md § "God files — plan around them, never into them" governs #4768;
  the `driver/settlement.rs` split-out is the template.
- The Unix-socket listener keeps its `#[cfg(unix)]` gate. Cross-platform IPC is
  a different piece of work.
- `stella-serve` builds its own binary and `stella-cli` does not link it, so
  #4770 is an HTTP integration between two processes rather than shared code.

macanderson:
[...]

PR #5522, opened 2026-08-30
feat(stella-cli): stella whistle --deep reaches worker lanes, and >@all broadcasts from inside the deck

### LINKED ISSUE #5431: stella whistle stops at a session's lead — worker lanes and delegate sub-agents never hear it
## Problem

`stella whistle "<msg>"` broadcasts steering to every live session on the machine (per-session Unix socket, per-target `Delivered`/`Unreachable` result). But it stops at each session's front door, and using it means a second terminal:

1. **Worker lanes never hear it.** A whistle lands on the session's lead tap (`command_deck/whistle.rs::DeckWhistle` relays into the lead's per-turn `SteeringTap`). Worker lanes (`req:N`, `sub:task-id`) are separately steerable in-process — `SubSessions::steer(lane, text)` landed with #2899 — but no whistle path fans into them. A user broadcasting "stop touching the release branch" to 12 sessions reaches 12 leads and zero of their workers.
2. **Delegate children are unsteerable by design.** `ChildSteering::drain_steering` returns empty so a child can never eat a message addressed to the parent (the documented "message thief" refusal in `crates/stella-core/src/subagent.rs`). The fix is addressing, not inheritance: steering entries gain an optional `sub_agent_id` address (events already carry the id, #4383/#4624); `ChildSteering` drains only entries addressed to its own id; unaddressed entries stay the parent's, preserving today's behavior exactly. This is a `stella-core` seam and gets driver-level scrutiny — split it into its own PR (or its own issue) if it grows past a reviewable slice.
3. **No broadcast from inside a session.** Add a composer address — `>@all <msg>` and `>@<session-id> <msg>` — routing through the whistle client from inside the deck, rendering the per-target delivery table as a chrome note; plus a `w` binding in the SESSIONS overlay to whistle the selected session.
4. **Sequential fan-out.** `whistle/cmd.rs::run` delivers one target at a time with 2s connect + 2s ack timeouts; at 10–15 sessions a few stale sockets make broadcast feel slow. Parallelize the loop; `Delivery` is already the per-target result type.

Full design context: the plan comment on #5331. The `stella-serve` surface is #4770 (discovery) — whatever lands there should carry the same `deep` flag.

## Files

- `crates/stella-cli/src/whistle/cmd.rs` — `run`, `deliver_one`, `Delivery`
- `crates/stella-cli/src/whistle/wire.rs` — `WhistleRequest` gains `deep: bool` (default false; old peers must keep decoding)
- `crates/stella-cli/src/command_deck/whistle.rs` — `DeckWhistle::deliver` fans into live lane taps when `deep`
- `crates/stella-cli/src/subsession.rs` — `SubSessions::steer`, the per-lane `taps` map
- `crates/stella-core/src/subagent.rs` — `ChildSteering` (the addressing seam, item 2)
- `crates/stella-tui/src/deck_ui/dispatch.rs` — the `>@` address route
- `crates/stella-parity/src/lib.rs` — the `turn.whistle` capability row's witness list grows with each new reach

## Constraints

- `stella-core` must never learn whistle exists — everything lands through `TurnSteering` / the addressed variant.
- Wire compatibility: an old listener receiving a `deep` request must still deliver to the lead rather than erroring.
- The whistle wrapper (`[whistle — steering from another session]`) stays on every delivered text, so a broadcast can never impersonate the session's own user prompt.
- Unix-socket transport keeps its `#[cfg(unix)]` gate.

## Definition of done

- [x] `stella whistle --deep "<msg>"` reaches a session's live worker lanes; witness: `a_deep_whistle_reaches_the_lead_and_every_live_lane` — the text arrives via each live lane's tap, and the engine's boundary drain emits `Steered` for it as for any steer
- [x] Addressed steering for delegate children — an engine seam, split to #5520 as this issue allows; not carried by PR #5522
- [x] `>@all` / `>@<session-id>` from the deck composer delivers via the whistle client and renders the per-target results
- [x] Fan-out is parallel; a stale socket delays only its own target's result
- [x] The parity ledger's `turn.whistle` row names the new witnesses

macanderson:
## What & why

`stella whistle` already broadcasts steering to every live session on the machine, but it stopped at each session's front door and needed a second terminal. This PR gives it depth and an in-session door:

- **`--deep` reaches worker lanes.** `WhistleRequest` carries a `deep` flag (serde-defaulted, so a sender built before the field decodes as the lead-only whistle it meant). `Whistleable::push_deep` has a default of `push`, so every non-interactive door — which drives no lanes — behaves exactly as before. The deck's relay (`DeckWhistle`) implements it: `SubSessions` announces each spawned lane's tap through a new `LaneTapSink` (one line in the driver: `subs.attach_lane_sink(whistle.clone())`), the relay keeps `Weak` handles for the reason it already keeps one for the lead, and a deep whistle lands on the lead and every live, non-settling lane, pruning ended lanes on the way past. A deep whistle at a session with no lanes is a plain whistle.
- **`>@all` / `>@<session-id>` from the composer.** Parsed in `dispatch::sideband` (`stella-tui`), so a broadcast is queue-free like `/export` — it is about other sessions and never waits on this one's turn. `--deep` right after the address reaches lanes. The driver runs it off the event pump (`whistle::broadcast_from_deck`) and reports every session's outcome as one chrome note, in the same rows the command prints. The broadcasting session is excluded from its own targets.
- **Parallel fan-out.** `whistle::cmd::broadcast` delivers to every target at once (`futures_util::future::join_all`); a stale socket's connect timeout now delays only its own row. `run` is the printing wrapper over it.

**Deliberately not in this PR:** addressed steering for in-process `delegate` children. That is a `stella-core` engine seam (`ChildSteering` refuses steering by design so a child can never eat the parent's message), and #5431 says to split it when it grows past a reviewable slice. It is filed as #5520 under epic #4806 with the addressing sketch.

Exemplars: the relay's `Weak` lane list follows its own `live` field; `broadcast_from_deck` follows `command_side`'s spawn-and-report shape; the `deep` wire field follows `stella-protocol`'s `#[serde(default)]` additive-field discipline.

- `command_deck::whistle::tests::a_deep_whistle_reaches_the_lead_and_every_live_lane` — lead and both live lanes receive the deep text, the plain whistle stops at the lead, a settling lane takes nothing, the ended lane is pruned
- `whistle::wire::tests::a_frame_without_the_deep_key_is_a_lead_only_whistle` — the wire-compatibility promise
- `whistle::cmd::tests::a_live_listener_is_delivered_to_and_an_absent_one_is_reported_unreachable` (extended: `broadcast` excludes the broadcasting session; `summary` renders the rows and names an empty result)
- `whistle::cmd::tests::a_broadcast_reaches_every_target_at_once` — every target is reported, none dropped
- `deck_ui::dispatch::tests::the_broadcast_address_parses_its_target_depth_and_message` — `>@all`, `>@<id>`, `--deep`, the exact-flag rule, and the no-message cases

The parity ledger's `turn.whistle` row names the two new witnesses in its mechanism text (the row's own witness is unchanged and still lives in the included file).

DoD notes against #5431, stated plainly: the delegate-child box is handed to #5520 rather than ticked; the SESSIONS-overlay `w` binding mentioned in the issue body (not its DoD) is not included — `>@<id>` from the composer covers the same action without a new key.

No `#[test]` was deleted.

## Test evidence

Targeted runs (SCR-001): `cargo test -p stella-cli whistle` (18 passed), `command_deck` (214), `goal`; `cargo test -p stella-tui --lib`; `cargo test -p stella-parity` (23); `make prose` clean; `cargo fmt` applied. CI's workspace run is the full evidence.

Closes #5431 — the delegate-child item continues in #5520.

## Summary by Sourcery

Extend whistle steering to worker lanes and enable concurrent broadcasts from within interactive sessions.

New Features:
- Add `--deep` whistle delivery to live worker lanes in addition to session leads.
- Add `>@all` and `>@<session-id>` composer addresses for queue-free broadcasts from inside a deck, with optional `--deep` delivery.

Bug Fixes:
- Preserve lead-only behavior when decoding whistle requests from older senders that omit the new depth field.

Enhancements:
- Fan out whistle broadcasts concurrently and report per-session delivery outcomes without allowing stale sockets to delay other targets.

Documentation:
- Document deep whistle delivery, in-session broadcasting, concurrent delivery behavior, and the new command syntax.

Tests:
- Add coverage for deep delivery to live lanes, wire compatibility, broadcast parsing, target exclusion, outcome reporting, and concurrent fan-out.

### DIFF
--- a/crates/stella-cli/src/agent/goal/tests.rs
+++ b/crates/stella-cli/src/agent/goal/tests.rs
@@ -75,6 +75,7 @@ async fn whistle_over_the_socket(socket: &std::path::Path, message: &str) {
         &mut stream,
         &crate::whistle::wire::WhistleRequest {
             text: message.to_string(),
+            deep: false,
         },
     )
     .await
--- a/crates/stella-cli/src/cli/command.rs
+++ b/crates/stella-cli/src/cli/command.rs
@@ -285,6 +285,10 @@ pub(crate) enum Command {
         /// Target only this session id (repeatable). Default: every live session.
         #[arg(long = "session")]
         session: Vec<String>,
+
+        /// Reach each session's live worker lanes as well as its lead
+        #[arg(long)]
+        deep: bool,
     },
 
     /// Analyze this workspace: domain taxonomy and code graph
--- a/crates/stella-cli/src/command_deck.rs
+++ b/crates/stella-cli/src/command_deck.rs
@@ -960,6 +960,9 @@ pub async fn run_deck_session(
     // Sub-session bookkeeping: live-worker slots, and `task_assign` requests
     // waiting for one (drained oldest-first as workers end).
     let mut subs = SubSessions::resuming(&cfg.durability, &session_record.id);
+    // Each worker lane's tap is announced to the whistle relay, so a deep
+    // whistle can reach the lanes this session drives — see `whistle`.
+    subs.attach_lane_sink(whistle.clone());
     let mut pending_spawns: VecDeque<subsession::QueuedSpawn> = VecDeque::new();
     // Lanes whose Restart arrived while the worker was still live: stop
     // first, respawn on its Ended.
@@ -1108,6 +1111,22 @@ pub async fn run_deck_session(
                         }
                         dispatch.release();
                         text
+                    }
+                    // The composer's broadcast address — other sessions'
+                    // whistle sockets, reported as one note (`whistle`).
+                    Some(WorkspaceInput::Whistle {
+                        message,
+                        session,
+                        deep,
+                    }) => {
+                        whistle::broadcast_from_deck(
+                            message,
+                            session,
+                            deep,
+                            &session_record.id,
+                            &in_tx,
+                        );
+                        continue 'session;
                     }
                     // The agents page's new-task prompt: a lane, never the
                     // lead's next turn — that is this message's whole meaning.
@@ -1974,6 +1993,11 @@ pub async fn run_deck_session(
                             if !command_side::run(text.trim(), cfg, &in_tx, &session_record.id) {
                                 queue.push_back(text);
                             }
+                        }
+                        // A broadcast never waits on the lead's turn: it is
+                        // about other sessions (`whistle`).
+                        Some(WorkspaceInput::Whistle { message, session, deep }) => {
+                            whistle::broadcast_from_deck(message, session, deep, &session_record.id, &in_tx);
                         }
                         // The agents page's new-task prompt: a lane now,
                         // exactly as at idle — the lead's turn is untouched.
--- a/crates/stella-cli/src/command_deck/whistle.rs
+++ b/crates/stella-cli/src/command_deck/whistle.rs
@@ -46,6 +46,14 @@ pub(super) struct DeckWhistle {
     live: Mutex<Weak<SteeringTap>>,
     /// Steers that arrived with no turn to take them, oldest first.
     pending: Mutex<Vec<String>>,
+    /// The worker lanes this session drives, announced as each spawns
+    /// ([`crate::subsession::LaneTapSink`]) — what a deep whistle reaches
+    /// beyond the lead. `Weak` for the reason `live` is: the driver's
+    /// `SubSessions` owns each tap and drops it with the lane, and a lane
+    /// that ended must read as gone here rather than as a queue nobody
+    /// drains. A deep whistle at an idle lane set is not buffered: a lane
+    /// that does not exist yet has no work the guidance could be about.
+    lanes: Mutex<Vec<Weak<SteeringTap>>>,
     /// This session's bound socket. Replaced when the deck switches sessions;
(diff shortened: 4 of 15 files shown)


macanderson:
Answering the two Sourcery ❌ rows:

- **Addressed steering for delegate children** is deliberately out of this PR. `ChildSteering` refuses steering by design (a child must never eat the parent's message), so giving it an address is a `stella-core` step-boundary seam that gets the agent-loop scrutiny the CLI plumbing here does not need. #5431 itself says to split it when it grows past a reviewable slice; it is filed as #5520 under epic #4806 with the addressing sketch, and #5431's DoD box now records the split.
- **The SESSIONS-overlay `w` binding** appears in #5431's body, not its Definition of done. The composer's `>@<session-id>` address performs the same action from the same screen with no new key to teach, which is why the binding was left out rather than added as a second door to one verb. If a maintainer wants the key anyway it is a small `views/sessions` change; say so and I will add it.

PR #5554, opened 2026-08-30
fix(release): reconcile the Homebrew tap against the newest published release

### LINKED ISSUE #5551: Homebrew tap formula is stuck at 0.9.235 while releases are at 0.9.291 — release workflow does not bump the tap
## What happened
`brew info macanderson/stella/stella` reported stable 0.9.235 while the latest GitHub release was v0.9.291.

## What it turned out to be
The published tap was **not** stale. `macanderson/homebrew-tap` receives a commit on every tag (`stella 0.9.292` at the time of writing), and `release.yml`'s `homebrew` job has been rendering and pushing the formula all along. Two other things produced the reading:

1. **A stale local clone.** A tap is a git clone on the machine and `brew info` reads whatever commit that clone sits on — it does not fetch. The reporting machine's clone was weeks old.
2. **Two taps for one repo.** `scripts/self-driving.sh` tapped under the name `macanderson/stella` while README.md, RELEASING.md and the website all say `macanderson/tap`. Both names reach the same repo, so the machine carried two independent clones, and `brew info` answered differently depending on which name was asked.

## The real gap
Nothing anywhere could answer "does the tap serve the newest release?" The `homebrew` job warns and exits 0 when neither tap credential is set, so an expired deploy key or a revoked token would freeze the formula while the tag, the release, and `check-releases-published.sh` all keep reporting success. That is why this report could be neither confirmed nor refuted without reading the tap by hand.

## Definition of done
- [x] A check answers whether the published tap formula is behind the newest published release, and fails when it is.
- [x] It does not fire in the window between a release being published and the tap job pushing the formula for it.
- [x] It runs on a cadence, not only by hand.
- [x] An unreachable tap errors rather than reporting that the tap is fine.
- [x] A hermetic test covers both failure directions and the version-ordering rule.
- [x] The self-driving surface taps under the documented name and retires the duplicate.
- [x] RELEASING.md says how to tell a stale local clone from a stale tap.

macanderson:
Closes #5551

## What the report turned out to be

The published tap was never stale. `macanderson/homebrew-tap` gets a commit on every tag and is at `stella 0.9.292` right now; `release.yml`'s `homebrew` job has been rendering and pushing the formula the whole time. The 0.9.235 reading came from a **stale local clone** — a tap is a git clone on the machine and `brew info` reads whatever commit it sits on, it does not fetch — and from **two taps for one repo**, because `scripts/self-driving.sh` tapped under `macanderson/stella` while README.md, RELEASING.md and the website all say `macanderson/tap`. Both names reach the same repo, so the machine carries two clones that fetch independently and `brew info` answers differently depending on which name was asked.

The interesting part is that this could not be settled without reading the tap by hand. Nothing anywhere asked whether the tap serves the newest release, so the report was neither confirmable nor refutable.

## The gap this closes

The `homebrew` job warns and exits 0 when neither tap credential is set, which is right for the release — a release should not be blocked on the tap being wired — and it means an expired deploy key or a revoked token would freeze the formula silently. The tag is cut, the release publishes, `check-releases-published.sh` reports a clean bill of health, and `brew install` keeps serving whichever build the tap last received. That is the same shape as the silence `check-releases-published.sh` was written for, one hop further down the pipeline, and it gets the same remedy.

`scripts/check-tap-current.sh` compares the tap's formula against the newest release that has aged past the grace window, and `release-reconcile.yml` runs it hourly as its own job beside the tag check — two jobs rather than two steps, so a frozen tap and an unpublished tag are two findings a maintainer sees at once instead of whichever fails first hiding the other.

Three rules worth naming:

- **The grace window.** This repo releases on nearly every merge, so the newest release is routinely minutes old and its tap push has legitimately not happened yet. A formula *ahead* of the newest settled release is the normal state there and is not a finding. An alarm that fires in that window is one a maintainer mutes, and a muted alarm is the silence again.
- **An unreachable tap errors** rather than reporting a clean bill it did not earn. A missing *formula* is the finding the script exists for; an unreachable *repo* is an unknown, and the two get different exits.
- **Versions compare as number arrays**, not strings. As strings `0.9.235` sorts below `0.9.9`, which would report a current tap as one release behind on every run for the rest of the 0.9 series.

Running it live found a defect the fixtures would not have: `gh` gives a draft release a publish date of `0001-01-01T00:00:00Z`, which `fromdateiso8601` refuses outright. A draft is carried through as a null the rule skips rather than dropped in the I/O half, so a fixture covers the reason.

## The two defects behind the report

- `scripts/self-driving.sh` now taps `macanderson/tap` and **untaps the duplicate** when it finds one. Safe because the formula is always installed by its fully qualified name, so nothing resolves through the duplicate.
- `upgrade.md` described the from-source formula in `packaging/homebrew/stella.rb` ("needs a Rust toolchain, takes a few minutes"). The tap serves the **prebuilt binary** — `bin.install "stella"`. Corrected, with the distinction named.
- RELEASING.md now says how to tell a stale local clone from a stale tap, since that is the diagnosis this issue needed.

## Witness

`scripts/test-tap-current.sh` — 17 cases over the rule with the I/O removed. It fails on `main` because neither the script nor the rule exists there.

```
$ ./scripts/test-tap-current.sh
passed 17, failed 0
```

Covering the silent direction (a tap behind, a tap carrying no formula at all), the crying-wolf direction (inside the grace window, a tap ahead, nothing settled), both sides of the grace boundary, numeric version ordering, non-numeric tags skipped without hiding a real gap behind them, and drafts.

Live, all three branches:

```
$ ./scripts/check-tap-current.sh                                   # exit 0
check-tap-current: OK — macanderson/homebrew-tap/Formula/stella.rb is at 0.9.292,
which is not behind any release older than 90m (697 release(s) checked).

$ ./scripts/check-tap-current.sh --formula-path Formula/nope.rb    # exit 1
::error::the Homebrew tap is 695 release(s) behind …

$ ./scripts/check-tap-current.sh --tap-repo macanderson/nope       # exit 1
::error::cannot reach the tap repo … This check could not ask its question —
it is not reporting that the tap is fine.
```

`make guards-fast` is green, including `gate-parity` (which is what caught the new self-test needing a workflow to host it) and `prose`.

## Exemplar

`scripts/check-releases-published.sh` — same pure-`--select`-function-plus-I/O-shell split, same fixture-driven suite, same grace-window reasoning. Deliberately a sibling rather than a second mode on that script: it answers a different question with a different remedy, and leaving its tested rule untouched keeps its suite meaningful.

## Deleted tests

None.

## Remaining work

Filed as #5555, not fixed here: both reconcilers cap `gh release list` at 1000 and this repo is at 697 releases growing ~16/day, so the guarded truncation error will fire on both within weeks. Raising the number is a deferral, not a fix, and the change belongs to both scripts at once rather than riding on the one that adds a sibling.

## Summary by Sourcery

Add independent monitoring to detect when the published Homebrew tap falls behind settled releases and standardize clients on the canonical tap name.

New Features:
- Add scheduled reconciliation that verifies the Homebrew tap serves the newest settled published release.

Bug Fixes:
- Use the canonical Homebrew tap name and remove duplicate legacy tap clones during self-driving upgrades.
- Correct upgrade documentation to describe the tap's prebuilt binary formula rather than the repository's source formula.

Enhancements:
- Handle grace periods, drafts, unreachable repositories, missing formulas, and numeric version ordering when evaluating tap freshness.

Build:
- Add Makefile targets for checking and testing Homebrew tap freshness.

CI:
- Run the tap freshness rule in guard self-tests and as a separate hourly release-reconciliation job.

Documentation:
- Document tap freshness reconciliation and how to diagnose stale local clones versus stale published taps.

Tests:
- Add a hermetic 17-case test suite covering stale, current, unsettled, malformed, version-ordering, and draft-release scenarios.


macanderson:
Both Sourcery findings on `check-tap-current.sh` are fixed in a5ba48c2.

**The `parts` one is a real bug and the worst kind this script can have.** `select` on the version pattern emits nothing for an input it cannot read, and in jq that does not skip one comparison — it drops the whole document through the `as $formula_parts` binding, so the rule produced no output and the caller printed a clean OK for a formula it never compared. Reproduced before fixing:

```
$ printf %s '{"formula_version":"0.9.292-rc","releases":[{"tag":"v0.9.292","published_unix":900000}]}' \
    | ./scripts/check-tap-current.sh --select --now 1000000 --grace-secs 5400
(no output — silent pass)
```

`parts` now returns null on anything it cannot read, which lands an unreadable version exactly where a missing formula already landed: behind every settled release. Same principle the unreachable-tap branch already followed — a check that cannot establish the tap is current must not report that it is. `E1` in `scripts/test-tap-current.sh` is the witness and fails on the previous rule; `E2` holds the other side, so an unreadable version with nothing settled is still not a finding.

**The message one is right too.** It assembled a `brew install` command from the tap owner, which was wrong whenever `--tap-repo` or `--formula-path` pointed elsewhere. It now names what was actually checked.

Suite is 19 cases, all green; the live check still exits 0 against the real tap at 0.9.292.

Issue #5560, opened 2026-08-30
main does not build docs: RECALL_MARKER links to a cfg(test)-only name


macanderson:
## What is wrong

`main` does not build docs. `cargo doc -D warnings` — a `make gate` step and part of CI's `fmt + clippy + test` job — fails on `crates/stella-cli/src/memory.rs`:

```
error: unresolved link to `inject_recall_block`
   --> crates/stella-cli/src/memory.rs:123:54
    |
123 | /// Marker prefixing a recalled-context message so [`inject_recall_block`]
    |                                                      ^^^^^^^^^^^^^^^^^^^ no item named `inject_recall_block` in scope
    |
    = note: `-D rustdoc::broken-intra-doc-links` implied by `-D warnings`
```

## Why

#5550 routed every recalling driver through `memory::inject_opening_recall`, so the bare `inject_recall_block` has no production caller left. An ungated `pub use` of it is then an unused import under `clippy -D warnings`, so the re-export was gated to `#[cfg(test)]`. `RECALL_MARKER`'s doc comment still links to the ungated name, and `cargo doc` builds without `cfg(test)` — so the name is out of scope there and only there. `cargo check`, `cargo clippy --all-targets` and the test suite all pass; only the rustdoc pass sees it.

## How it reached main

#5550 was merged at 21:24:51Z. Its own `fmt + clippy + test` job reported this failure at 21:24:57Z — six seconds later. Every other check on that PR was green at merge time.

## Repro

```bash
git checkout 23bea5745
RUSTDOCFLAGS="-D warnings" cargo doc -p stella-cli --no-deps
# exits 101 with the error above
```

## Definition of done

Each item is stated so it can be verified *before* the merge, on the fix branch — whose base is 23bea5745, the broken commit — rather than only after it.

- [x] `RUSTDOCFLAGS="-D warnings" cargo doc -p stella-cli --no-deps` exits 101 on 23bea5745 and 0 on the fix branch, so the repair is the cause of the pass
- [x] `cargo fmt --all --check` clean on the fix branch
- [x] No production behaviour changed — the whole diff is five lines of one doc comment
- [x] CI's `fmt + clippy + test` job green on the fix branch (run 33337148281, 16m27s), which is the job that fails on `main`

## Fix

PR #5557 repoints the link at `recall::inject_recall_block`, which resolves in both builds. Five lines, one doc comment, no other file touched.

PR #5563, opened 2026-08-30
fix(stella-cli): repair main — a doc link resolving through a cfg(test) re-export

### LINKED ISSUE #5562: main is red: a doc link resolves through a cfg(test) re-export, so cargo doc fails
## What is broken

`main` is red at `23bea5745` on the `cargo doc -D warnings` step of `fmt + clippy + test`:

```
error: unresolved link to `inject_recall_block`
   --> crates/stella-cli/src/memory.rs:123:54
    |
123 | /// Marker prefixing a recalled-context message so [`inject_recall_block`]
    |                                                      ^^^^^^^^^^^^^^^^^^^ no item named `inject_recall_block` in scope
    |
error: could not document `stella-cli`
```

Every PR inherits it, so red stops distinguishing "your change is broken" from "the base is".

## Why it got through

The re-export the link resolved through is `#[cfg(test)]`-gated:

```rust
#[cfg(test)]
pub(crate) use recall::inject_recall_block;
```

So the name is in scope for `cargo test` and for `cargo clippy --all-targets`, and absent for `cargo doc`, which compiles without the test cfg. Every tier that builds the test configuration stayed green. This is a repeat of a shape that has cost a red `main` before: a `#[cfg(test)]` re-export breaks intra-doc links while check, clippy and tests all pass.

## Reproduce

```bash
cargo doc -p stella-cli --no-deps      # warns on this tree
make doc-warnings                      # the gate step, -D warnings
```

## Fix

`mod recall;` is not gated, so naming the module keeps the link live under either cfg:

```rust
/// Marker prefixing a recalled-context message so [`recall::inject_recall_block`]
```

Escaping the brackets would also silence it, but it would turn a working link into prose.

## Done when

- [ ] `cargo doc -p stella-cli --no-deps` emits no warning.
- [ ] The link still resolves to the item rather than being escaped into plain text.
- [ ] `main`'s `fmt + clippy + test` job is green again.

macanderson:
Closes #5562

`main` has been red since `23bea5745` on `cargo doc -D warnings`:

```
error: unresolved link to `inject_recall_block`
   --> crates/stella-cli/src/memory.rs:123:54
    | no item named `inject_recall_block` in scope
error: could not document `stella-cli`
```

## Why it got through

The re-export the link resolved through sits behind a test gate:

```rust
#[cfg(test)]
pub(crate) use recall::inject_recall_block;
```

The name is therefore in scope for `cargo test` and for `cargo clippy --all-targets`, and absent for `cargo doc`, which compiles without the test cfg. Every tier that builds the test configuration stayed green, so nothing upstream of rustdoc could see it.

## The fix

`mod recall;` is not gated, so naming the module keeps the link resolving under either cfg:

```rust
/// Marker prefixing a recalled-context message so [`recall::inject_recall_block`]
```

Escaping the brackets would also silence the error, but it would turn a working link into plain text — the wrong trade for a one-character-cheaper diff.

## Verification

```
$ cargo doc -p stella-cli --no-deps
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 2.71s
   Generated .../target/doc/stella/index.html
```

Clean; the same command warns on the parent commit.

## Witness

None. A rustdoc lint is the check, and `make doc-warnings` is what fails before this commit and passes after — a test asserting rustdoc emits no warning would be a test of rustdoc.

## Deleted tests

None.

## Note on scope

Found while a separate PR (#5554) sat red on an inherited break rather than its own. Landing it on its own from a fresh `main` rather than folding it into that branch, so the repair is reviewable by itself and does not race the baseline-style shared-cell hazard.

## Summary by Sourcery

Bug Fixes:
- Fix the stella-cli rustdoc warning caused by a documentation link resolving only through a test-only re-export.


macanderson:
Superseded — the identical fix already merged as #5557 (closing #5560), and `main` has been green on it since commit `4221bda2f`. This branch's `memory.rs` diff conflicts with what's already on `main` because it's the same one-line repoint, just phrased slightly differently. Closing unmerged; no further action needed here.

Issue #5529, opened 2026-08-30
stella-transcript's Color::Amber conflates warnings and brand-gold identity


macanderson:
## Problem

`crates/stella-transcript/src/grid.rs`'s `Color` enum is a fixed, small
categorical palette (`Ink`, `Dim`, `Faint`, `Cyan`, `Green`, `Red`, `Amber`,
`Violet`, `Blue`, `Magenta`). `Amber` is documented as `"Warnings."` and is
used for exactly that (`Status::Warn`, wait/verdict notes) — but it is also
the closest thing this crate has to a gold-identity colour, and gets reused
wherever exported content needs a gold-adjacent hue (e.g.
`syntax::Tok::Keyword => Color::Amber`).

There is no distinct gold-identity role in this crate's palette, so a
reader exporting a transcript sees "warning" and "brand gold" collapse onto
the same ANSI value: `Color::Amber.ansi16()` returns `33` (SGR plain
yellow) and `Color::Amber.ansi256()` returns `179`, for both meanings.

#5000 settled the disagreement between the deck's `FALLBACKS` table and
`stella_tui_theme::fallback::ansi16` for `gold`/`gold_bright` (both now
degrade to bright yellow), but never touched this crate — `stella-transcript`
is a near-leaf with `stella-diff` as its only workspace dependency
(AGENTS.md), so it cannot import `stella-tui-theme`'s token system to
compare against, and its palette is a different, smaller categorical model
in the first place (warning-or-not, not brand-identity-or-not).

## Why this matters

A reader of an exported session transcript (the surface most likely to be
seen outside the live deck, per #5000's own framing) cannot visually tell
"this is a warning" from "this is styled with the brand's identity colour"
— both render identically. Whether that distinction is worth a new
categorical role in a crate that is deliberately kept small is the open
question this issue is for.

## Files

- `crates/stella-transcript/src/grid.rs` — the `Color` enum, `ansi16`,
  `ansi256`, and every call site listed by `rg -n 'Color::Amber'`

## What "done" looks like

A decision, recorded and tested, on one of:

1. **Split the role.** Add a distinct gold-identity variant to `Color`,
   with its own `ansi16`/`ansi256` mapping (bright yellow / the nearest
   256-cube gold), and repoint the call sites that mean "gold" rather than
   "warning" to it.
2. **Keep the conflation, name it.** Document on `Color::Amber` that it
   deliberately serves both roles, with the reasoning for why a reader
[...]

Issue #5019, opened 2026-08-26
The transcript's 16-colour encoder has no production caller


macanderson:
## Problem

`crates/stella-transcript/src/grid.rs` exports `to_ansi16`, which encodes a rendered transcript with the basic 16-colour SGR codes. Nothing ships a call to it.

```
$ rg -n 'to_ansi16' crates/ bench/ website/
crates/stella-transcript/src/tests.rs:1069:    let ansi = grid::to_ansi16(&grid::render(&run, &state, 100));
crates/stella-transcript/src/grid.rs:1069:pub fn to_ansi16(lines: &[Line]) -> String {
crates/stella-transcript/README.md:107:background spans; `to_ansi16` degrades them to bold + underline, which is a real
```

One caller, and it is a test in the same crate. The production encoder is `to_ansi256`, called once:

```
$ rg -n 'ansi256' crates/stella-cli/src/plain/transcript.rs
181:        .map(|line| grid::to_ansi256(std::slice::from_ref(line)))
```

So `Color::ansi16` — the third of the three 16-colour tables #5000 is about — maps colours that never reach a terminal. This was found while measuring #5000, and it changes that issue's premise: *"a reader on a 16-colour terminal sees the brand as bright yellow in the deck and as plain yellow in an exported transcript of the same session"* cannot happen today, because there is no 16-colour transcript export.

## Why it matters

Two ways, and they want opposite fixes, which is why this is a decision rather than a patch:

- **If a 16-colour transcript should ship**, `crates/stella-cli/src/plain/transcript.rs` is where the choice between the two encoders belongs, and it should be made on what the terminal reports rather than hardcoded to 256. `stella_tui_theme::fallback::detect_truecolor` is the existing shape for asking; a 256-colour SGR sequence on a 16-colour terminal is not a graceful degradation, it is a terminal rendering whatever its own approximation is.
- **If it should not**, `to_ansi16` and `Color::ansi16` are dead public API and should go, along with the `README.md` sentence describing them. `Tint::ansi256` has no `ansi16` counterpart already, which is a hint that the 16-colour path was never finished.

Either way #5000's third table stops being a table nobody can see.

## Files

- `crates/stella-transcript/src/grid.rs` — `to_ansi16`, `Color::ansi16`, `encode`'s `basic` parameter
- `crates/stella-cli/src/plain/transcript.rs` — the one production encoder call
- `crates/stella-transcript/README.md` — the sentence describing `to_ansi16`
[...]

PR #5579, opened 2026-08-31
fix(stella-tui-theme): the status colours take the bright half at sixteen colours

### LINKED ISSUE #5000: Three 16-colour fallback tables ship, and they disagree about gold
## Problem

Three tables in this workspace answer "what does this colour become on a 16-colour terminal?", and they give different answers for the same colour. Nothing compares them.

| table | file | what gold becomes |
|---|---|---|
| `FALLBACKS` | `crates/stella-tui/src/theme.rs` | 16-index **11** — bright yellow |
| `ansi16` | `crates/stella-tui-theme/src/fallback.rs` | `Color::Yellow` = ANSI **3** — plain yellow |
| `Ink::ansi16` | `crates/stella-transcript/src/grid.rs` | SGR **33** = ANSI 3, and gold is not a distinct `Ink` at all — it arrives as `Color::Amber` |

Each is reached by a different path — `theme::degrade_buffer` for the v1 deck, `stella_tui_theme::fallback::ansi16` for v2 surfaces, `to_ansi16` for a rendered transcript — so a reader on a 16-colour terminal sees the brand as bright yellow in the deck and as plain yellow in an exported transcript of the same session. `design/tui-v2/SPEC.md` §3.5 says only "gold to yellow", which both satisfy, so the spec cannot settle it either.

Two of the three also disagree on the neutral ramp: `FALLBACKS` sends `TEXT_SECONDARY` to 7 and `TEXT_TERTIARY`/`TEXT_DIM` to 8, while `ansi16` sends `SILVER` to `Color::Gray` (7) and `MUTED`/`DIM` to `Color::DarkGray` (8) — those agree — but `FALLBACKS` sends `palette::TEXT_EMPHASIS` to 15, sharing it with `TEXT_PRIMARY`, while `ansi16` sends `SILVER_TYPE` to 7, sharing it with `SILVER`. Same two values, opposite collapse.

## Why nothing caught it

Each table has its own test and each test walks only its own table:

- `crates/stella-tui/src/theme/tests.rs`'s `every_named_token_has_a_fallback` — every token has a `FALLBACKS` row.
- `crates/stella-tui-theme/src/tests.rs`'s `every_token_has_a_fallback` — `ansi16` is total over `token::ALL`.

Both are *coverage* checks. Neither compares a row to the other table's row for the same value, so two complete tables that disagree are two green tests. This is the shape AGENTS.md invariant #8 describes for providers — a matrix checked from both sides — applied to nothing here.

## Files

- `crates/stella-tui/src/theme.rs` — `FALLBACKS`, `resolve`, `degrade_buffer`
- `crates/stella-tui-theme/src/fallback.rs` — `ansi16`
- `crates/stella-transcript/src/grid.rs` — `Ink::ansi16`, `Ink::ansi256`, `to_ansi16`
- `design/tui-v2/SPEC.md` §3.5 — the degradation table, checked against `ansi16` only by `crates/stella-tui/tests/spec_palette.rs` (added in PR #4991)

## Verify

```
rg -n -A20 'const FALLBACKS' crates/stella-tui/src/theme.rs
rg -n -A22 'pub fn ansi16' crates/stella-tui-theme/src/fallback.rs
rg -n -A12 'pub fn ansi16' crates/stella-transcript/src/grid.rs
```

## Definition of done

One of:

1. **One table.** The v1 deck and the v2 surfaces read the same map, and `crates/stella-transcript`'s `Ink` declares which token each of its nine inks is so its own mapping derives rather than restates. Then §3.5 governs all three.
2. **Three tables, one checked against another.** A test that walks `token::ALL` and asserts `FALLBACKS`'s 16-index equals the ANSI index of `ansi16`'s answer for the same value, with any deliberate divergence carrying a named reason at the row — the shape `provider_parity.rs` uses.

Either way §3.5's table (`crates/stella-tui/tests/spec_palette.rs`, `the_spec_states_the_degradation_that_ships`) should end up holding *every* path that ships, not the one it can see today. Whichever gold survives is a design call: bright yellow reads as the identity on a dark 16-colour terminal, plain yellow does not, and the transcript's is the surface a reader is most likely to see outside the deck.

Refs #4976, #4991

macanderson:
## What

The last two disagreeing rows between the two shipping 16-colour tables — the v1 deck's `FALLBACKS` and the v2 `stella_tui_theme::fallback::ansi16` — were `green` (bright green vs plain green) and `red` (same shape). They are settled on the **bright half**:

- It is the rule every already-settled chromatic row follows (#5528 moved `gold`, `gold_bright` and `silver_type` to the deck's bright answers), so the sixteen degrade under one decision instead of five taste calls — the framing the audit on the issue put on it.
- Both tokens are pale pastels (`#74C991`, `#E0687A`) on a dark ground; the standard half renders markedly darker on common default palettes, and the existing divergence declaration already recorded bright red as the nearer match against xterm's defaults.
- The issue's core complaint was one session rendering different colours on the two surfaces; after this the tables agree on every compared row.

Changed together, held by tests:

- `crates/stella-tui-theme/src/fallback.rs` — `GREEN → LightGreen`, `RED → LightRed`, with the reasoning added to the doc comment's ramp list.
- `design/tui-v2/SPEC.md` §3.5 — the degradation table now says bright green / bright red, plus one sentence of reasoning (`spec_palette.rs` holds the table to `ansi16` row by row).
- `crates/stella-tui/src/theme/tests.rs` — `DEGRADATION_DIVERGENCE` is now **empty**; the cross-check `the_two_16_colour_tables_agree_or_declare_why_not` therefore enforces full agreement, and demands a declared row for any future divergence.
- `crates/stella-tui-theme/src/tests.rs` — the row-pinning assertions updated.

Closes #5000

## Evidence

- Witness the artisanal way: `git stash push crates/stella-tui-theme/src/fallback.rs` (main's rows) → `sixteen`-row pinning test in stella-tui-theme **and** the cross-check in stella-tui both FAIL; pop → `cargo test -p stella-tui-theme`, `cargo test -p stella-tui --lib theme`, `cargo test -p stella-tui --test spec_palette` all green.
- No golden moves: goldens are style-stripped; the 16-colour path is exercised by the table tests above.

The third table — `stella-transcript`'s `Ink::ansi16` — stays out of the cross-check: its 16-colour encoder has **no production caller** (#5019), and its real defect is that gold arrives as `Color::Amber`, conflating warnings with brand identity — filed as #5529. When #5529 gives the transcript a distinct gold-identity ink, §3.5's table is the answer it should map to. Both issues already exist; nothing new to file.

## Summary by Sourcery

Make 16-colour status rendering consistent by using bright green and bright red in both theme fallback tables.

Bug Fixes:
- Align 16-colour status fallbacks by mapping green and red to bright green and bright red across the v1 deck and v2 theme tables.

Enhancements:
- Document the shared bright-half rule for chromatic status colours and enforce complete agreement between the two 16-colour degradation tables.

Documentation:
- Update the v2 theme specification to define bright green and bright red as the 16-colour status fallbacks.

Tests:
- Update fallback assertions and remove the now-resolved divergence declarations so cross-table tests require full agreement.


macanderson:
Answering the two Sourcery ❌ rows — both are the same objection, and both are deliberately out of this PR's scope rather than missed:

- **"the transcript's `Ink::ansi16` mapping is left out of the cross-check / the third table can still disagree."** Correct, and it is a declared, issue-tracked gap, not a silence. The transcript's 16-colour encoder has **no production caller** (#5019), so there is no shipping degradation path to hold to §3.5 today. More to the point, the cross-check cannot be written yet: `Ink` has no gold-identity ink to compare — gold arrives as `Color::Amber`, and that conflation is exactly #5529's subject. Comparing `Amber` (a warning colour) against §3.5's `gold` row would assert an identity the type does not have. When #5529 gives the transcript a distinct gold ink, §3.5 is the table it should derive from, and the cross-check becomes writable.
- **"every shipping 16-colour degradation path"**: after this PR, both paths that ship — `theme::degrade_buffer` (v1 deck) and `stella_tui_theme::fallback::ansi16` (v2 surfaces) — agree on every compared row and are held to §3.5 by test. The unshipped path is tracked in #5019/#5529.

Issue #5000's own DoD is an either/or, and option 2 ("tables checked against each other, divergences named") is what this PR completes — with the divergence ledger now empty rather than merely annotated.

Issue #5616, opened 2026-08-31
The migration file's own table of steps is missing seven modules


macanderson:
## Problem

`crates/stella-store/src/migrations.rs` opens with a table of its step submodules and says every new step "declares itself here". Seven modules never did:

- `foundry_plane` (v40 → v41)
- `delivered_runs` (v37 → v38)
- `dispatched_executions` (v35 → v36)
- `telemetry_call_identity` (v36 → v37)
- `sub_agent_calls` (v32 → v33)
- `sub_agent_tool_calls` (v34 → v35)
- `partial_reflection` (v31 → v32)

A reader who trusts the table sees a smaller module set than the directory holds. Nothing checks the table against the `mod` list, which is how it drifted.

## Files

- `crates/stella-store/src/migrations.rs` — the doc table near the top, and the `mod` block below it
- `crates/stella-store/src/migrations/` — the directory the table should match

## Verify

```sh
rg -n '^mod ' crates/stella-store/src/migrations.rs
rg -n '^//! \| \[' crates/stella-store/src/migrations.rs
```

Compare the two lists; the seven names above appear in the first and not the second.

## Constraints

- The prose gate (`make prose`) holds this file to its recorded reading grade and issue-reference count, so new rows must stay plain and cite no issue numbers.
- Keep each row's third column to what the group shares, matching the rows already there.

## Definition of done

- [x] Every module in the `mod` block has a row in the table (grouped rows are fine, as the `live_tool_calls` row shows)
- [x] `make prose` stays green
- [x] **Verification:** re-run the two `rg` commands in Verify above after the fix — the module names in the `mod` list and the doc-table list match exactly (today the seven names above show up in the first list and not the second; after the fix they show up in both). This before/after comparison is the fix's own proof. Per this repo's own witness-test policy (`AGENTS.md` — "Pure refactors, docs, and CI changes don't need a witness"), this is a pure documentation change, so no compiled witness test is required — say so explicitly in the PR description rather than skipping the question.
- [x] File organization: no new files or modules are created — the fix stays inside the existing doc table in `crates/stella-store/src/migrations.rs`. No crate-placement question. Confirm the added rows don't push the file past the file-size gate (it is not on the current god-file list, and should stay off it).
- [x] Docs: this issue's fix *is* the documentation update (the table itself) — no separate external doc needs a companion change.

## Evidence (fixed in #6312)

[...]

PR #5617, opened 2026-08-31
feat(stella-store): record how each foundry tool enablement was authorised (#5403)

### LINKED ISSUE #5403: The adopted-tool row does not record who approved it, or how
## Problem

`stella tools --enable <name>` now asks a human and refuses when none is present ([#5402](https://github.com/macanderson/stella/pull/5402), closing #5332). What it still does not do is **record which of the two it was**.

The adoption ledger row (`AdoptedTool`, `crates/stella-store/src/foundry.rs`) carries `enabled: bool` and nothing about the grant. So after the fact there is no way to tell:

- an interactive `y` at a terminal, from
- a `--yes` in a provisioning script, from
- (before #5402) no approval at all.

#5332's Definition of done asked for exactly this — "stamps the authority on the adopted-tool row" — and it is the one line of that issue this PR did not do.

## Why it was split out rather than done there

Two reasons, both worth a decision rather than an invention inside an unrelated PR:

1. **It is a store schema change.** `AdoptedTool` has no authority column, so this is a migration in `stella-store`, not a field assignment.
2. **What `--yes` should stamp is a real question.** The honest value is "somebody asserted they had read the declaration", which is a *weaker* claim than an interactive answer. Recording both as one `approved_by_human = true` would throw away the distinction that motivates the field. The vocabulary already exists next door: `crates/stella-protocol/src/provenance.rs` grades evidence and publication authority, and `crates/stella-parity/src/evolution.rs`'s `Tool` row is what declares this surface's requirements.

## Files

- `crates/stella-store/src/foundry.rs` — `AdoptedTool`, `set_foundry_tool_enabled`, and the schema
- `crates/stella-cli/src/tool_foundry/adopt.rs` — `run_tools_enable_in`, which knows which path was taken
- `crates/stella-cli/src/tool_foundry/adopt/tests.rs` — the four consent tests #5402 added
- `crates/stella-parity/src/evolution.rs` — the `Tool` row this surface answers to
- `crates/stella-protocol/src/provenance.rs` — the existing authority vocabulary

## Verify

```sh
rg -n "struct AdoptedTool" -A 20 crates/stella-store/src/foundry.rs
rg -n "run_tools_enable_in" -A 30 crates/stella-cli/src/tool_foundry/adopt.rs
```

## Definition of done

- [x] The adopted-tool row records how the enablement was authorised, distinguishing an interactive answer from `--yes`
- [x] A row enabled before this shipped reads as unknown rather than as either — an absent record must not be back-dated into a claim
- [x] `stella tools --foundry` surfaces it, so the answer is reachable without opening the database
- [x] The vocabulary is the one in `provenance.rs` rather than a second grading invented here, or the PR says why it cannot be
- [x] Witness test: enabling interactively and enabling with `--yes` produce different recorded authority, and a pre-existing row produces neither

Split out of #5332 / [#5402](https://github.com/macanderson/stella/pull/5402).

macanderson:
## What & why

`stella tools --enable` gained a consent gate in #5402, but the adopted-tool row still recorded only `enabled: bool` — an interactive answer at a terminal, a `--yes` in a provisioning script, and a pre-gate enablement were indistinguishable after the fact.

Schema **v42** adds `foundry_tools.enabled_authority`: the path that made the grant. `Store::set_foundry_tool_enabled` now takes `Option<EnableAuthority>`, so no caller can enable without saying how:

- `interactive_human` — a typed yes at a verified terminal (`run_tools_enable_in`'s prompt path)
- `flag_assertion` — `--yes` (recorded as the claim it is, not as an observed person)
- `autonomy` — the `auto` autonomy pipeline enabling its own proven adoption
- `rollback` — `stella tools --rollback` re-enabling a restored version

A disable (human, `--disable`, or the circuit breaker) clears the record; re-adoption revokes it with the approval; **pre-v42 rows read as unknown** — the migration backfills nothing, so an absent record is never back-dated into a claim. `stella tools --foundry` prints an `enabled:` line for every live grant, unknown included.

**On the vocabulary DoD item:** the stored value is the observed *path*, not a `provenance.rs` value, because `PublicationAuthority` names actors and the `--yes` path's actor is unobservable — storing `local_human` would launder an assertion into a sighting, and storing `agent` would claim the opposite. Instead `EnableAuthority::established_authority()` derives the strongest `PublicationAuthority` each path proves (`interactive_human → local_human`; the other three → `agent`, the weakest actor the evidence allows), and the report prints both. Store the fact, derive the grade — the same rule `stella-parity/src/evolution.rs` follows for its own authority columns (the exemplar for this design). No second grading is invented: `EnableAuthority` is a flat record with no order.

Closes #5403

## The witness

One per DoD claim:

- `migrations/enable_authority.rs::a_pre_existing_enablement_reads_as_unknown` — builds a real v41-shaped file, migrates, and asserts the enabled row reads `''`. Fails on `main` (the migration does not exist).
- `foundry/tests.rs::each_grant_records_its_authority_and_a_disable_clears_it` and `adopt/tests.rs::the_two_consent_paths_record_different_authorities` — interactive and `--yes` produce different recorded authority, and the report surfaces it (`proves local_human` included). Neither compiles on `main`: the field and the parameter are genuinely absent.
- `foundry/tests.rs::re_adopting_clears_the_recorded_authority`, plus the strengthened rollback witness in `ops.rs` (`enabled_authority == Some(Rollback)`) and the strengthened `--yes` witness (`Some(FlagAssertion)`).

The interactive prompt itself cannot run in a test (no terminal); its one post-yes write is `set_enabled_in`, which the witnesses drive directly.

## The gate

- [x] Formatting checked (rustfmt, default config)
- [x] Clippy clean at `-D warnings` for the three touched crates (`stella-store`, `stella-tools`, `stella-cli`), all targets — this PR's full evidence is its CI run, per the repo rule that CI builds and tests
- [x] Tests pass per-crate: `stella-store` 473 lib + 31 migration/doc, `stella-tools` 587, `stella-cli` `tool_foundry` filter 34 — the workspace run is CI's
- [x] Docs updated where behavior changed (schema doc in `ddl.rs`, migration ladder, `--foundry` report line)
- [x] CLA signed

## Nothing left behind

- [x] Filed: #5616 — the `migrations.rs` submodule doc table has drifted: seven step modules (`foundry_plane`, `delivered_runs`, `dispatched_executions`, `telemetry_call_identity`, `sub_agent_calls`, `sub_agent_tool_calls`, `partial_reflection`) are absent from the table that says every step "declares itself here". This PR adds its own row but does not repair the drift.

## Ground-rule check

- New cross-boundary type: `EnableAuthority` crosses the store→CLI boundary as a Rust enum, like `AdoptedTool` beside it — neither is serialized across a process boundary, so no serde round-trip applies (it round-trips through its SQL tag instead, covered by the store tests).

`set_foundry_tool_enabled(name, enabled: bool)` became `set_foundry_tool_enabled(name, authority: Option<EnableAuthority>)` rather than gaining a second parameter: a `bool` plus an authority that is meaningful only when `true` is two fields describing one state, and the `Option` makes "enabled without saying how" unrepresentable.

## Summary by Sourcery

Track and report how foundry tool enablements were authorised without inferring approval details for legacy grants.

New Features:
- Record the mechanism that authorised each enabled foundry tool, distinguishing interactive approval, `--yes`, autonomy, and rollback paths.
- Display live enablement authority and its derived provenance in the `stella tools --foundry` report, including unknown authority for legacy grants.

Bug Fixes:
- Clear recorded enablement authority whenever a tool is disabled, re-adopted, or disabled by the circuit breaker.
- Preserve pre-v42 enablements as unknown rather than backdating an authority claim.

Enhancements:
- Require callers to provide an explicit authority when enabling a foundry tool and derive provenance grades from the recorded enablement path.

Documentation:
- Document the new schema field, migration, and foundry report output.

Tests:
- Add migration, store, adoption, autonomy, and rollback coverage for authority recording, clearing, persistence, and legacy unknown values.

### DIFF
--- a/crates/stella-cli/src/tool_foundry/adopt.rs
+++ b/crates/stella-cli/src/tool_foundry/adopt.rs
@@ -31,7 +31,7 @@ use colored::Colorize;
 use serde_json::Value;
 use stella_core::ports::ToolExecutor;
 use stella_protocol::tool::{ToolOutput, ToolSchema};
-use stella_store::{AdoptedTool, Store};
+use stella_store::{AdoptedTool, EnableAuthority, Store};
 use stella_tools::custom::{self, CustomTool, CustomToolSet};
 use stella_tools::foundry_gate::PROPOSED_DIR;
 use stella_tools::foundry_witness::{WitnessVerdict, prove};
@@ -276,6 +276,7 @@ fn finish_adoption(
         enabled: false,
         adopted_at: String::new(),
         disabled_reason: String::new(),
+        enabled_authority: None,
     };
     store
         .adopt_foundry_tool(&record)
@@ -301,6 +302,13 @@ fn finish_adoption(
 ///
 /// `Ok(false)` means a human was asked and said no — the one outcome that is
 /// neither a grant nor an error.
+///
+/// Which consent path ran is stamped onto the row: a typed yes records
+/// [`EnableAuthority::InteractiveHuman`], `--yes` records
+/// [`EnableAuthority::FlagAssertion`]. They are different claims — a person
+/// the CLI saw at a terminal versus a claim by whatever process passed the
+/// flag — and recording them as one would throw away the difference the
+/// consent gate exists for.
 pub(crate) fn run_tools_enable_in(
     root: &Path,
     store: &Store,
@@ -328,7 +336,13 @@ pub(crate) fn run_tools_enable_in(
         }
     }
 
-    set_enabled_in(root, store, name, enabled)?;
+    let authority = match (enabled, yes) {
+        (false, _) => None,
+        (true, true) => Some(EnableAuthority::FlagAssertion),
+        // The prompt above ran and was answered yes, at a verified terminal.
+        (true, false) => Some(EnableAuthority::InteractiveHuman),
+    };
+    set_enabled_in(root, store, name, authority)?;
     Ok(true)
 }
 
@@ -399,7 +413,8 @@ fn confirm_enable(name: &str) -> Result<bool, String> {
 }
 
 /// Flip one adoption's enablement, refusing to enable a tool whose bytes no
-/// longer match the ones the witness ran against.
+/// longer match the ones the witness ran against. `Some(authority)` enables,
+/// saying how the grant was authorised; `None` disables.
 ///
 /// Checking here as well as at discovery is not redundant: the gate withholds
 /// a tampered tool silently-but-explainably at session start, whereas someone
@@ -408,7 +423,7 @@ pub(crate) fn set_enabled_in(
     root: &Path,
     store: &Store,
     name: &str,
-    enabled: bool,
+    authority: Option<EnableAuthority>,
 ) -> Result<(), String> {
     let record = store
         .adopted_foundry_tool(name)
@@ -420,7 +435,7 @@ pub(crate) fn set_enabled_in(
             )
         })?;
 
-    if enabled {
+    if authority.is_some() {
         let manifest_path = root.join(ADOPTED_DIR).join(format!("{name}.toml"));
         let text = std::fs::read_to_string(&manifest_path)
             .map_err(|e| format!("cannot read {}: {e}", manifest_path.display()))?;
@@ -445,7 +460,7 @@ pub(crate) fn set_enabled_in(
     }
 
     if !store
-        .set_foundry_tool_enabled(name, enabled)
+        .set_foundry_tool_enabled(name, authority)
         .map_err(|e| format!("cannot update the adoption ledger: {e}"))?
     {
         return Err(format!("`{name}` has not been adopted"));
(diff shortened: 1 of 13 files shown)


macanderson:
Automated PR sweep — DoD verification for #5403. **All five items check out against this branch, so I have ticked them on the issue.** Evidence per item below, measured on `5aa5711`, so the tick is auditable rather than a claim.

| # | Item | Where it is satisfied |
|---|---|---|
| 1 | Records how the enablement was authorised, interactive vs `--yes` | `EnableAuthority` (`stella-store/src/foundry.rs:59`) — four variants; `adopt.rs`'s `run_tools_enable_in` maps `(true, true) → FlagAssertion`, `(true, false) → InteractiveHuman`, `(false, _) → None` |
| 2 | A pre-existing row reads as unknown, not back-dated | `migrations/enable_authority.rs:25` adds the column `NOT NULL DEFAULT ''` and runs **no** `UPDATE` — nothing is backfilled; `''` reads as unknown |
| 3 | `stella tools --foundry` surfaces it | `adopt.rs:501-513` prints an `enabled:` line for every live grant, `None` rendering as `"unrecorded — turned on before the ledger kept this"` |
| 4 | Uses `provenance.rs`'s vocabulary, or says why not | The PR says why not, and derives rather than restates: `established_authority()` (`foundry.rs:114`) maps `InteractiveHuman → LocalHuman` and the other three → `Agent`. The report prints both. The item's own "or the PR says why it cannot be" branch is met. |
| 5 | Witness: the two paths record different authority; a pre-existing row neither | `the_two_consent_paths_record_different_authorities` (`adopt/tests.rs:450`), `a_pre_existing_enablement_reads_as_unknown` (`migrations/enable_authority.rs:66`), plus `each_grant_records_its_authority_and_a_disable_clears_it` and `re_adopting_clears_the_recorded_authority` |

Item 5 carries one nuance worth stating rather than leaving for a reader to find: the `--yes` half runs end-to-end through `run_tools_enable_in`, while the interactive half is driven through `set_enabled_in` with `Some(InteractiveHuman)`, because the prompt needs a terminal no test has. The PR body already says this. The item asks that the two paths *record* different authority, which is what the test asserts (`assert_ne!(by_flag, interactive)`) — so it is met as written, and the gap between "the store write differs" and "the prompt was exercised" is real but outside what this issue asked for.

I checked #2 the way it can actually fail: the question is not whether the column exists but whether anything backfills it, so I read the migration for an `UPDATE` and found none.

Nothing else on this PR needs action — no merge conflict, and no unresolved review threads (Sourcery's assessment table marks all three objectives ✅). CI was still in flight at sweep time; the run on `5aa5711` is the authoritative gate.

---

PR #5639, opened 2026-09-02
fix(stella-core): refuse a sub-agent ceiling too short for a child to run

### LINKED ISSUE #5386: stella-core: a sub-agent ceiling of 30 seconds or less gives the child a deadline of now, so it aborts having done nothing
## Problem

`bounded_by_ceiling` in `crates/stella-core/src/subagent.rs` computes the child deadline as `now + ceiling.saturating_sub(CEILING_RESERVE)` with `CEILING_RESERVE = 30s`. Any configured ceiling ≤ 30s saturates to zero, so the child's deadline is `now`: its first deadline check reports Exceeded and the delegation aborts having done no work — while 31s behaves normally. A silent boundary cliff instead of a floor or a refusal.

## Where

- `crates/stella-core/src/subagent.rs` — `CEILING_RESERVE` and `bounded_by_ceiling`; the doc comment above it explains why the reserve exists (the child must see its own limit before the parent's).

## Repro / verify

Unit test: `bounded_by_ceiling(None, Some(Duration::from_secs(30)), now)` returns `Some(now)`; drive a delegation with a 30s ceiling and observe an immediate Exceeded with zero steps.

- The reserve's purpose (child sees its limit first) is sound; the defect is only the degenerate range. Candidate shapes: floor the derived budget at some minimum useful duration, or refuse a ceiling ≤ CEILING_RESERVE loudly at configuration time. Which one is a design call — the module's neighbors prefer loud refusal over silent clamping.

## Definition of done

- [x] `bounded_by_ceiling` returns a typed error, not a silent floor, when `ceiling <= CEILING_RESERVE`
- [x] The typed error is a variant on a typed refusal enum, never a bare `String`
- [x] The refusal reaches the parent as `SubAgentOutcome::Refused`, with zero model calls and zero cost, before the child's engine is built
- [x] A witness test covers both sides of the 30s boundary: `Duration::from_secs(30)` refuses, `Duration::from_secs(31)` succeeds
- [x] `make guards-fast` and a targeted `cargo test -p stella-core` run clean, and CI is green on the PR

---

**Items 1–4 verified against PR #5639 at `19a6933` (autonomous PR sweep),** read
off the diff rather than off the PR description:

1. The signature is now
   `fn bounded_by_ceiling(..) -> Result<Option<Instant>, SubAgentRefusal>`, and
   the ceiling arm is
   `Some(ceiling) if ceiling <= CEILING_RESERVE => return Err(SubAgentRefusal::CeilingTooShort { ceiling, reserve: CEILING_RESERVE })`.
   The old `saturating_sub`-to-`now` floor is gone from that path.
2. `SubAgentRefusal` is a new crate-private enum with `DepthExceeded`,
   `NoBudgetHeadroom` and `CeilingTooShort`, one `Display` impl and an
   `std::error::Error` impl. The two pre-existing refusals that were
   hand-built `String`s (a `format!` and a `.to_string()`) are converted in the
   same diff, so `refusal()` now returns `Option<SubAgentRefusal>` rather than
   `Option<String>` — the bare-`String` path is removed, not merely bypassed.
3. `run_sub_agent_with_sender` derives the deadline **before** the match, and
   the `Err` arm becomes `SubAgentOutcome::Refused` without entering the branch
   that builds the child. The old call site inside `run_child_turn` — the one
   that used to compute the ceiling a second time — is deleted, so there is now
   exactly one read of `self.config.tool_timeout` on this path.
4. `a_ceiling_at_or_under_the_reserve_is_refused_not_floored` asserts
   `CEILING_RESERVE` (30s) → `Err(CeilingTooShort { .. })` and
   `CEILING_RESERVE + 1s` (31s) → `Ok(Some(now + 1s))` — both sides of the
   boundary, as the DoD asks. `a_ceiling_too_short_refuses_the_whole_spawn_loudly`
   drives a full `run_sub_agent` and asserts `Refused`, `calls == 0`,
   `cost_usd() == 0.0`, and a `Finished` event carrying
   `SubAgentStatus::Refused`.

   Both fail on `main` by construction rather than by assertion: they destructure
   a `Result` from a function that returned `Option` before this diff, so they do
   not compile against the old signature.

**Item 5 ticked — the job the previous sweep was waiting on has reported.**
`fmt + clippy + test` on PR #5639's head `19a6933` completed **success** at
2026-09-02T23:17:32Z
([job 100455096361](https://github.com/macanderson/stella/actions/runs/33692753105/job/100455096361)).
That is the job carrying both the workspace clippy run and the test suite, and
it was the one thing this item asserted that was not yet observable. Every
other check on the head is also green — MSRV, file-size ratchet, guard
self-tests, docs guards, prose, dependency-review, `cargo deny`, pytest, the
duplicate-claims and undone-edit guards — leaving `dod` itself as the only red
check, which is what this tick clears.

Items 1–4 were re-verified independently against the branch in this sweep
(`git fetch origin fix/5386-ceiling-refusal`, reading `subagent.rs` and
`subagent/tests/spend_and_cancellation.rs` directly) and agree with the reading
above. The PR author's local `cargo test -p stella-core --lib subagent` run
remains their claim; CI's green run is the evidence this tick rests on.

macanderson:
## What

`bounded_by_ceiling` in `crates/stella-core/src/subagent.rs` derived a child sub-agent's deadline as `now + ceiling.saturating_sub(CEILING_RESERVE)`. Any configured `tool_timeout` at or under the 30s reserve saturated to zero, handing the child a deadline of exactly `now`. The child aborted at its first step boundary having done no work — a silent boundary cliff that read as a model declining to answer rather than a configuration mistake.

## Why

This function is the single seam that reads `self.config.tool_timeout` for a sub-agent's deadline. Loud refusal beats a silent floor: the module's other two refusal paths (nesting depth, no budget headroom) already refuse the spawn outright before the child's engine is built, so a too-short ceiling now joins them instead of degrading into a no-op child.

## How

- `bounded_by_ceiling` now returns `Result<Option<Instant>, SubAgentRefusal>` and returns `Err(SubAgentRefusal::CeilingTooShort { ceiling, reserve })` when `ceiling <= CEILING_RESERVE`.
- `SubAgentRefusal` is a new crate-private enum replacing the two prior refusal reasons' hand-built `String`s (`DepthExceeded`, `NoBudgetHeadroom`, `CeilingTooShort`), each with one `Display` impl. The design pointer on the issue named an "existing delegation error enum" to add a variant to — no such enum existed in `stella-core`, so this PR introduces it rather than hand-building a third `format!` call site next to the two that already existed; happy to fold this differently if that reads as scope creep.
- The deadline is now derived once, in `run_sub_agent_with_sender`, before the child's engine is ever assembled, and threaded through both `refusal()` (which turns `Err` into a loud refusal) and, on every other path, into the child's own budget carve — so there is exactly one call site reading the ceiling, not two computing it independently.

## Witness

Two new tests in `crates/stella-core/src/subagent/tests/spend_and_cancellation.rs`:

- `a_ceiling_at_or_under_the_reserve_is_refused_not_floored` calls `bounded_by_ceiling` directly at both sides of the boundary: `Duration::from_secs(30)` (the reserve exactly) returns `Err(CeilingTooShort { .. })`; `Duration::from_secs(31)` returns `Ok(Some(now + Duration::from_secs(1)))`, matching the design pointer's exact witness.
- `a_ceiling_too_short_refuses_the_whole_spawn_loudly` drives a full `run_sub_agent` with `tool_timeout: Some(Duration::from_secs(30))` and asserts the parent sees `SubAgentOutcome::Refused` naming the ceiling, zero model calls, zero cost, and a `Finished` event with `SubAgentStatus::Refused` — the same contract every other refusal keeps.

Both fail to compile against `main`'s `Option<Instant>`-returning signature, so they fail on the old code by construction rather than by assertion. The existing `a_childs_deadline_is_the_tighter_of_what_it_inherited_and_the_ceiling` test was updated for the new `Result` return type; no behavior it covered changed.

Verified locally with a targeted crate test (`cargo test -p stella-core --lib subagent`, allowed per CLAUDE.md as a single-crate, single-area run): all 39 sub-agent tests pass. `cargo fmt --all` and `make guards-fast` are clean.

## Tests deleted

None.

Closes #5386

## Summary by Sourcery

Reject sub-agent ceilings that leave no usable execution time before the child is created.

Bug Fixes:
- Refuse sub-agent spawns when the configured tool timeout is at or below the deadline reserve instead of starting a child with an immediately expired deadline.

Enhancements:
- Introduce typed sub-agent refusal reasons and centralize deadline calculation so refusal handling and child execution use the same deadline.

Tests:
- Add boundary and end-to-end coverage confirming short ceilings are reported loudly with no model calls or cost, while preserving existing deadline behavior.

### DIFF
--- a/crates/stella-core/src/subagent.rs
+++ b/crates/stella-core/src/subagent.rs
@@ -635,7 +635,7 @@ impl Engine<'_> {
         budget: &mut BudgetGuard,
         events: &EventSender,
     ) -> SubAgentOutcome {
-        let carve = budget.carve(spec.budget_usd);
+        let mut carve = budget.carve(spec.budget_usd);
         let _ = events.send(AgentEvent::SubAgent {
             phase: SubAgentPhase::Started {
                 agent_id: spec.agent_id.clone(),
@@ -681,7 +681,11 @@ impl Engine<'_> {
                 let deadline = deadline.expect(
                     "refusal() returns Some for every Err case above; reaching here means Ok",
                 );
-                self.run_child_turn(host, spec, carve, budget, events, &tally, deadline)
+                // Set here, on the carve the caller still owns, rather than
+                // adding a parameter to `run_child_turn` for one field of a
+                // struct it already receives.
+                carve.set_task_deadline(deadline);
+                self.run_child_turn(host, spec, carve, budget, events, &tally)
                     .await
             }
         };
@@ -715,7 +719,6 @@ impl Engine<'_> {
         budget: &mut BudgetGuard,
         events: &EventSender,
         tally: &Arc<CommittedTally>,
-        deadline: Option<std::time::Instant>,
     ) -> SubAgentOutcome {
         // Attribution is entered before anything the child could emit and
         // released by drop, so an unwind cannot leave the parent's later
@@ -852,11 +855,9 @@ impl Engine<'_> {
         let seeded = messages.len();
 
         let child_events = child_sender(events.clone(), spec.agent_id.clone(), tally.clone());
-        // A call may never be granted more than what is left of the ceiling the
-        // whole child runs under (#4488). The caller already derived this
-        // deadline, before the child's engine was even built. Recomputing it
-        // here would be a second seam reading the same config field.
-        carve.set_task_deadline(deadline);
+        // `carve` already carries its task deadline: the caller derived it
+        // from the ceiling and set it before this turn's engine was even
+        // built (#4488), so a too-short ceiling never reaches this far.
         // The carve is handed to the turn through a guard that settles it on
         // DROP, not on return (#1850). `settle_child` used to be a statement
         // after the await, so any exit that was not a return skipped it: a
--- a/crates/stella-core/src/subagent.rs
+++ b/crates/stella-core/src/subagent.rs
@@ -665,10 +665,23 @@ impl Engine<'_> {
             armed: true,
         };
 
-        let outcome = match refusal(spec, &carve) {
-            Some(reason) => SubAgentOutcome::Refused { reason },
+        // The one seam that reads `self.config.tool_timeout`. Its result
+        // feeds `refusal()` first, then the child's own carve on every
+        // other path.
+        let deadline = bounded_by_ceiling(
+            carve.task_deadline(),
+            self.config.tool_timeout,
+            std::time::Instant::now(),
+        );
+        let outcome = match refusal(spec, &carve, &deadline) {
+            Some(refusal) => SubAgentOutcome::Refused {
+                reason: refusal.to_string(),
+            },
             None => {
-                self.run_child_turn(host, spec, carve, budget, events, &tally)
+                let deadline = deadline.expect(
+                    "refusal() returns Some for every Err case above; reaching here means Ok",
+                );
+                self.run_child_turn(host, spec, carve, budget, events, &tally, deadline)
                     .await
             }
         };
@@ -702,6 +715,7 @@ impl Engine<'_> {
         budget: &mut BudgetGuard,
         events: &EventSender,
         tally: &Arc<CommittedTally>,
+        deadline: Option<std::time::Instant>,
     ) -> SubAgentOutcome {
         // Attribution is entered before anything the child could emit and
         // released by drop, so an unwind cannot leave the parent's later
@@ -839,12 +853,10 @@ impl Engine<'_> {
 
         let child_events = child_sender(events.clone(), spec.agent_id.clone(), tally.clone());
         // A call may never be granted more than what is left of the ceiling the
-        // whole child runs under (#4488).
-        carve.set_task_deadline(bounded_by_ceiling(
-            carve.task_deadline(),
-            self.config.tool_timeout,
-            std::time::Instant::now(),
-        ));
+        // whole child runs under (#4488). The caller already derived this
+        // deadline, before the child's engine was even built. Recomputing it
+        // here would be a second seam reading the same config field.
+        carve.set_task_deadline(deadline);
         // The carve is handed to the turn through a guard that settles it on
         // DROP, not on return (#1850). `settle_child` used to be a statement
         // after the await, so any exit that was not a return skipped it: a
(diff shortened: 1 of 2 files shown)


macanderson:
## Autonomous PR sweep — CI is red on clippy, not on a test

`fmt + clippy + test` failed on `f92107e`, and it is worth separating the two things that check does, because only one of them is unhappy:

```
error: this function has too many arguments (8/7)
   --> crates/stella-core/src/subagent.rs:710:5
error: could not compile `stella-core` (lib) due to 1 previous error
```

**Every test passed** — the run reached `cargo test` and came back clean across the workspace (2588 in one binary, 1550 in another, 177, 159, … all `0 failed`), including this PR's two new witnesses. The only failure is `clippy::too_many_arguments` on `run_child_turn`, which the new `deadline` parameter pushes from 7 to 8 (counting `&self`).

### The parameter is redundant, so the fix can delete it rather than suppress the lint

`run_child_turn` takes `mut carve: BudgetGuard` **by value**, and `BudgetGuard` already carries the deadline — the function's whole use of the new argument is to write it into that carve:

```rust
// subagent.rs:859, the first line in the body that touches `carve`
carve.set_task_deadline(deadline);
```

I traced every mention of `carve` inside the function: the parameter at `714`, then `859`, then `879`/`882`/`897`. **Nothing between entry and `859` reads `carve.task_deadline()`**, so setting it in the caller instead is behaviourally identical — the same value is written to the same slot, just one frame earlier. `set_task_deadline` is `pub fn set_task_deadline(&mut self, …)` and the caller's binding is `let carve = budget.carve(spec.budget_usd)` at `638`, so this needs `let mut carve` there (or a rebind inside the `None` arm):

```rust
None => {
    let mut carve = carve;
    carve.set_task_deadline(
        deadline.expect("refusal() returns Some for every Err case above; reaching here means Ok"),
    );
    self.run_child_turn(host, spec, carve, budget, events, &tally).await
}
```

…and drop `deadline` from `run_child_turn`'s signature and its `carve.set_task_deadline(deadline)` line. Back to 7 arguments, lint satisfied, no behaviour change.

That keeps what this PR is actually for — the seam moving *above* the child-engine build so a too-short ceiling refuses before anything is spent. The deadline is still computed once, in `run_sub_agent_with_sender`, and still feeds `refusal()` first. Only the hand-off changes: through the carve that was already being moved in, rather than alongside it.

**Deliberately not proposing `#[allow(clippy::too_many_arguments)]`.** CLAUDE.md counts an `#[allow]` that suppresses an avoidable lint as a defect against the PR that introduces it, and this one is avoidable — the argument genuinely is redundant, so the lint is right here rather than wrong.

### DoD, checked against the diff

I verified #5386's first four items against `f92107e` while diagnosing the above. They hold:

| DoD item | Evidence |
| --- | --- |
| Typed error, not a silent floor, when `ceiling <= CEILING_RESERVE` | `bounded_by_ceiling` returns `Result<Option<Instant>, SubAgentRefusal>`; the `Some(ceiling) if ceiling <= CEILING_RESERVE` arm returns `Err(CeilingTooShort { .. })` instead of the old `now + ceiling.saturating_sub(CEILING_RESERVE)` flooring at `now`. |
| A variant on a typed refusal enum, never a bare `String` | `enum SubAgentRefusal { DepthExceeded, NoBudgetHeadroom, CeilingTooShort }`, with `Display` and `std::error::Error`. `refusal()` went from `Option<String>` to `Option<SubAgentRefusal>`, so the other two refusals stopped being `format!`-built strings in the same change. |
| Reaches the parent as `Refused`, zero model calls, zero cost, before the child's engine is built | The deadline is computed in `run_sub_agent_with_sender` and consumed by `refusal()` *before* the `None` arm calls `run_child_turn` — which is where the engine assembly lives. `a_ceiling_too_short_refuses_the_whole_spawn_loudly` asserts `child_provider.calls == 0`, `outcome.cost_usd() == 0.0`, and a `SubAgentStatus::Refused` event. |
| Witness covers both sides of the 30s boundary | `a_ceiling_at_or_under_the_reserve_is_refused_not_floored`: `CEILING_RESERVE` (30s) → `Err(CeilingTooShort)`, `CEILING_RESERVE + 1s` (31s) → `Ok(Some(now + 1s))`. Expressing them relative to the constant rather than as literal `30`/`31` is the better call — it cannot drift if the reserve moves. |

The fifth item ("`make guards-fast` and a targeted `cargo test -p stella-core` run clean, and CI is green on the PR") is the one the clippy failure blocks, so I left **all five** boxes unticked rather than tick four and imply the gate was close to satisfiable. Once clippy is green they can go over together.

I did not push the fix: this PR's branch is not mine to write to, and it is being actively worked. Flagging rather than racing a commit into it.

---

macanderson:
## Autonomous PR sweep — clippy failure diagnosed, peer fix verified, one DoD item still open

### The CI failure on `f92107e`

`fmt + clippy + test` was red, and only step 44 (`cargo clippy -D warnings`) failed — `cargo fmt --check`, `cargo doc -D warnings`, `cargo test`, the release smoke build and every guard step passed. The whole of it:

```
error: this function has too many arguments (8/7)
   --> crates/stella-core/src/subagent.rs:710:5
710 | /     async fn run_child_turn(
    | |         ...
718 | |         deadline: Option<std::time::Instant>,
719 | |     ) -> SubAgentOutcome {
    = note: `-D clippy::too-many-arguments` implied by `-D warnings`
error: could not compile `stella-core` (lib) due to 1 previous error
```

Threading the derived deadline into `run_child_turn` took it to eight parameters, one over the lint's limit.

### It is already fixed, by someone else, on this branch

`19a6933` ("keep run_child_turn at seven arguments") sets the deadline on `carve` in the caller and drops the parameter. That is right, and it is the fix that needed no `#[allow]`: `run_child_turn` did nothing with the deadline except `carve.set_task_deadline(deadline)`, and it does not touch `carve` anywhere between taking ownership and that line, so moving the call to the caller is behaviour-identical.

I verified it rather than assuming, on `19a6933`:

- `cargo clippy -p stella-core --all-targets -- -D warnings` → exit 0
- `cargo fmt --all -- --check` → exit 0
- `cargo test -p stella-core --lib subagent` → `39 passed; 0 failed`, the same 39 this branch reported before

CI's own run on that head is in flight and should agree.

**Worth flagging:** I had written and committed the identical fix — same approach, same two edits — and my push was rejected because `19a6933` landed first. I discarded mine rather than force over it. Two sessions independently reaching the same two-line change is the shape AGENTS.md's `scripts/issue-claim.sh` exists for; the claim check would have stood one of us down.

### DoD

#5386's four code items are satisfied by this branch and I have ticked none of them yet, because the fifth is not:

- `bounded_by_ceiling` returns `Result<_, SubAgentRefusal>` and refuses at `ceiling <= CEILING_RESERVE` rather than flooring at `now` — no silent clamp.
- `SubAgentRefusal` is a typed enum with `Display` + `Error`, and it absorbed the two prior hand-built `String` reasons rather than adding a third `format!` beside them.
- The refusal reaches the parent as `SubAgentOutcome::Refused` before the child's engine is assembled; `a_ceiling_too_short_refuses_the_whole_spawn_loudly` asserts zero model calls, `cost_usd() == 0.0`, and a `Finished` event carrying `SubAgentStatus::Refused`.
- `a_ceiling_at_or_under_the_reserve_is_refused_not_floored` covers both sides of the boundary — 30s refuses, 31s yields `now + 1s` — which is the witness the issue names.

The fifth item ends "**and CI is green on the PR**", and CI is not green yet on this head. I will tick the set once that run lands green; ticking it now would be claiming a result nobody has.

---

PR #5644, opened 2026-09-02
fix(stella-fleet): validate a fleet worker's reported cost

### LINKED ISSUE #5498: A fleet worker's reported cost is never checked, so NaN wedges the budget and a negative value bypasses it
## The problem

A fleet worker reports what its task cost as a bare `f64`, and nothing checks
that number before it reaches the budget guard and the spend ledger. A `NaN`
wedges the budget for the rest of the session; a negative value silently buys
more spend than the operator allowed.

## Where

- `crates/stella-fleet/src/fleet.rs`, `WorkerOutcome::cost_usd: f64` — the
  worker-reported value.
- `crates/stella-fleet/src/fleet.rs`, `dispatch_claimed` — passes it straight to
  `stella_core::BudgetGuard::record_spend`.
- `crates/stella-fleet/src/ledger.rs`, `finish_attempt` — inserts it into
  `spend`, whose `cost_usd REAL NOT NULL` column carries no `CHECK`.
- `crates/stella-fleet/src/ledger.rs`, `total_spend`/`task_spend` — plain
  `SUM(cost_usd)`.
- `rg -n "is_finite|is_nan|is_sign_negative"` over the crate returns nothing.

## Failure scenario

**NaN.** A worker's own pricing divides by zero and reports `cost_usd: NaN`.
`BudgetGuard`'s check is `if spent_usd <= limit_usd { return None }`, and
`NaN <= limit` is false, so every later evaluation reports over-budget no matter
what was really spent. Under `Enforced` the run wedges shut; under `Observed` it
warns forever. The ledger's `total_spend` for that run is `NaN` permanently —
an unrecoverable number in what the crate calls the authoritative record of real
money spent.

**Negative.** A unit-conversion bug reports `-1000.0`. It *reduces* the running
total, so several later `$0.99` tasks each pass an `Enforced` `$1.00` limit that
true cumulative spend has already exceeded. The budget gate is bypassed without
ever tripping `AbortTurn`.

## The fix

Validate once at the seam where the worker's number becomes the fleet's, in
`dispatch_claimed` right after the worker returns and before `record_spend` or
`finish_attempt`: require `cost_usd.is_finite() && cost_usd >= 0.0`, and reject
or clamp with a named typed error rather than trusting reported data. A `CHECK`
constraint on the column is worth adding beside it, so a future writer cannot
reintroduce the same value by another path.

## Definition of done

- [x] A non-finite or negative worker cost cannot reach `record_spend` or the
      `spend` table.
- [x] Witness, either level: a fake `FleetWorker` reporting `cost_usd: -1000.0`
      under `BudgetMode::Enforced` with a `$1.00` limit, then several `$0.99`
      tasks, asserting one trips `AbortTurn` — none do today. Or purely at the
      ledger: `finish_attempt` with `f64::NAN` and assert `total_spend()` is not
      `NaN`.
- [x] `cargo test -p stella-fleet` green.

## Confidence

CONFIRMED by the audit, which traced the value from `WorkerOutcome` through both
consumers and confirmed the crate contains no finiteness or sign check.

## Related, same crate

`Ledger::finish_attempt` has no idempotency guard: the `attempts` UPDATE has no
`WHERE finished_at_ms IS NULL` and no affected-row check, and it unconditionally
inserts a new `spend` row, with no `UNIQUE` on `spend.attempt_id`. Calling it
twice for one attempt silently doubles that attempt's recorded cost.
`record_lineage` next door is deliberately idempotent via a `UNIQUE` plus
`INSERT OR IGNORE`, which is the pattern to copy. `Fleet::dispatch_claimed` calls
it once per freshly-minted id today, so this is latent — but `Ledger` is `pub`
and already called from outside the crate.

---

**Item 3 ticked, and the first attempt at it was red (autonomous PR sweep).**

`cargo test -p stella-fleet` is green on PR #5644's current head `f6f516ca`:
`fmt + clippy + test` completed **success** at 2026-09-02T23:29:56Z
([job 100459752022](https://github.com/macanderson/stella/actions/runs/33694251397/job/100459752022)).

**It was not green on the previous head, and this item is the reason that was
caught.** At `45fc0b0` the branch's own witness test failed, reproduced locally:

```
$ cargo test -p stella-fleet --lib ledger::
test ledger::tests::finish_attempt_rejects_a_negative_cost_at_the_storage_layer ... FAILED
    panicked at crates/stella-fleet/src/ledger.rs:1104:
    a negative cost_usd must violate the spend table's CHECK constraint
test result: FAILED. 29 passed; 1 failed
```

The cause: `finish_attempt` wrote the spend row with `INSERT OR IGNORE`, added
for the new `UNIQUE (attempt_id)` idempotency. `OR IGNORE` is not scoped to that
constraint — it ignores **every** constraint on the row, so it also swallowed
the `CHECK (cost_usd >= 0)` this issue asked for. A negative cost was silently
dropped and the call returned `Ok`, which is the exact opposite of the
storage-level rejection the DoD's first item describes. Sourcery's review had
flagged the neighbouring NaN claim as inconsistent with the schema; that was
right about the mechanism and understated the consequence.

Fixed in `f6f516c` by naming the conflict target —
`ON CONFLICT (attempt_id) DO NOTHING` — which keeps the retried close a no-op
while leaving both value constraints able to reject. Verified against SQLite
directly rather than argued:

```
duplicate attempt_id ignored, rows = 1
negative rejected -> CHECK constraint failed: cost_usd >= 0
NaN     rejected -> NOT NULL constraint failed: spend.cost_usd
```

The NaN test now asserts the rejection it actually gets (error returned, the
attempt's own close rolled back, total still finite) instead of a NULL-row
`SUM`-skip fallback that SQLite never produces, and the two doc comments
describing that fallback are corrected.

So all three items now hold on `f6f516ca`: items 1 and 2 via
`dispatch_claimed`'s `is_finite() && >= 0.0` guard and the four witness tests,
and item 3 via CI's green workspace run. Local `cargo test -p stella-fleet --lib`
reports 139 passed, 0 failed; `cargo clippy -p stella-fleet --all-targets -D warnings`
is clean; `check-prose` and `check-file-size` both exit 0.

**Residue filed:** #5649 — `crates/stella-fleet/src/ledger.rs` now sits at
exactly 1500 of its 1500-line limit, with the room for `f6f516c` reclaimed by
tightening that commit's own prose. No baseline entry was added. The next line
added to that file crosses, so it wants a split first.

macanderson:
## What and why

A fleet worker's `cost_usd: f64` was trusted at face value. Two ways that goes wrong:

- **Negative.** `record_spend` accumulates with plain `+=`, so a negative report *shrinks* the running total. A later, honestly-reported task can then pass an `Enforced` limit that real cumulative spend has already crossed — the budget gate never trips `AbortTurn`.
- **NaN.** `BudgetGuard::check_axis` gates on `spent_usd <= limit_usd`; that comparison is `false` for `NaN`, so the guard falls through to `Some(AbortTurn)`/`Some(Warn)` on *every* later `evaluate()` call for the rest of the run, no matter what is actually spent — the run wedges shut (`Enforced`) or warns forever (`Observed`), permanently, because `x + f64::NAN` is `NaN` for every `x`.

Fixed at the seam the design pointer named: `Fleet::dispatch_claimed`, right after the worker returns and before `record_spend`/`finish_attempt`. A report that fails `is_finite() && cost_usd >= 0.0` is rejected: the attempt is recorded **failed, with zero spend** (never the untrustworthy number, never silently dropped — its real commits are still kept), and `dispatch_claimed` returns the new `FleetError::InvalidWorkerCost { task, reported }` so the dispatch itself reads as a failure, same as a worktree or claim error.

`crates/stella-fleet/src/ledger.rs`'s `spend` table gets `CHECK (cost_usd >= 0)` as a second, independent layer (`MIGRATION_V4`) — SQLite has no `ALTER TABLE ADD CONSTRAINT`, so this is a table rebuild, the same shape `MIGRATION_V2` used for `lineage`. A NaN report is caught by the column's pre-existing `cost_usd REAL NOT NULL` rather than by the new `CHECK`, since SQLite binds a NaN parameter as `NULL` before a `CHECK` ever evaluates it. An old row with a negative cost, or the `NULL` a legacy NaN insert became, is clamped to zero during the rebuild rather than failing it — refusing to open a user's ledger over a row that was already wrong money would be worse than the bad number.

**Residue folded in from the issue's "Related, same crate" note:** `Ledger::finish_attempt` had no idempotency guard — the `attempts` UPDATE had no `WHERE finished_at_ms IS NULL`/affected-row check, and the `spend` insert was unconditional with no `UNIQUE` on `attempt_id`. A second call for one attempt (a timeout's stop-then-grace race settling twice, or any future caller — `Ledger` is `pub`) silently doubled the recorded spend. `finish_attempt` is now idempotent the way `record_lineage` already is: the `attempts` update is `WHERE finished_at_ms IS NULL`, a zero affected-row count makes the whole call a no-op (committed, not rolled back — there is nothing new to apply), and `spend.attempt_id` is `UNIQUE`, with `INSERT ... ON CONFLICT (attempt_id) DO NOTHING` as a second, independent layer for a retried close. That conflict target is named deliberately rather than `INSERT OR IGNORE`: `IGNORE` swallows *every* constraint on the row, not just the `UNIQUE` one — an earlier version of this PR used `IGNORE` and it silently absorbed the `CHECK`/`NOT NULL` rejections too, dropping the spend row and returning `Ok` instead of rejecting the value. Caught by a `cargo test -p stella-fleet --lib ledger::` run against the branch and fixed in this PR's second commit before any review round-trip.

Exemplar followed: `record_lineage`'s existing `UNIQUE` + conflict-target idempotency shape (`crates/stella-fleet/src/ledger.rs`), and `MIGRATION_V2`'s table-rebuild pattern for adding a constraint SQLite can't `ALTER TABLE` onto an existing table.

## Witness

Both levels the issue's DoD names, plus the idempotency witness from the "Related" note:

- `crates/stella-fleet/src/fleet/tests.rs::a_negative_reported_cost_does_not_disable_the_enforced_budget_gate` — a fake worker reports `-1000.0` under `BudgetMode::Enforced` with a `$1.00` limit; the dispatch is rejected and the running total does not move. Two further `$0.99` dispatches are then chained against the *same* `BudgetGuard` (carried forward via `budget_snapshot`, which is `Copy`): the first is `Continue`, the second trips `AbortTurn`. On `main`, `record_spend(-1000.0)` would shrink the total by 1000.0 and neither `$0.99` task would ever trip the cap — this fails by construction on the old code, since nothing there rejects the report.
- `a_nan_reported_cost_is_rejected_and_never_poisons_the_budget_total` — a fake worker reports `f64::NAN`; the dispatch is rejected and `budget_snapshot().session_spent_usd()` stays finite (`0.0`). On `main`, `record_spend(NaN)` would poison that total to `NaN` for the rest of the run.
- `crates/stella-fleet/src/ledger.rs::finish_attempt_with_a_nan_cost_leaves_total_spend_finite` — calls `Ledger::finish_attempt` directly with `cost_usd: f64::NAN` and asserts it returns `Err` (a NaN binds as NULL, `cost_usd REAL NOT NULL` rejects that) and that `total_spend()` stays finite. Not a fail→pass witness against `main` specifically — `spend.cost_usd` was already `NOT NULL` before this PR, so a raw NaN insert already errored there too. It stayed green through this PR's own history for the right reason only after the `ON CONFLICT` fix above; kept as direct-`Ledger`-caller coverage since `Ledger` is `pub` and `Fleet::dispatch_claimed`'s guard sits one layer above it.
- `finish_attempt_rejects_a_negative_cost_at_the_storage_layer` — calls `finish_attempt` directly with `cost_usd: -5.0`; fails on `main` (no constraint), rejected by `MIGRATION_V4`'s `CHECK` on the new code. Also asserts the attempt's own close rolls back with the rejected insert (one transaction, no half-apply).
- `a_second_finish_attempt_for_the_same_attempt_does_not_double_spend` — calls `finish_attempt` twice for one attempt; on `main` this doubles both the spend row and the commit rows, asserted here to stay at exactly one of each.
- `a_negative_reported_cost_is_rejected_and_recorded_as_zero_spend` — the direct-effect witness: the dispatch fails with `InvalidWorkerCost`, ledger + budget both read `0.0`, and the worker's real commit is still recorded (only the cost figure is distrusted).

`cargo test -p stella-fleet` (the DoD's third box) ran in CI's `fmt + clippy + test` job: `stella_fleet`'s unittest binary, 139 passed, 0 failed, including all six tests above — https://github.com/macanderson/stella/actions/runs/33694251397/job/100459752022.

## Notes for review

- `crates/stella-fleet/src/ledger.rs` was at 1519 lines with these changes before trimming (over the 1500-line ratchet, and this crate had no prior god file) — tightened the new doc comments and added a small `seeded_attempt` test helper to land at 1500/1500, no file split needed.
- Sourcery's review comment carries one ❌ against this PR's *first* commit, before the `ON CONFLICT` fix above landed — see the PR comment thread for why that row is addressed; a fresh Sourcery review is rate-limited for a few days.
- `Tests deleted: None.`

Closes #5498

### DIFF
--- a/crates/stella-fleet/src/fleet.rs
+++ b/crates/stella-fleet/src/fleet.rs
@@ -462,6 +462,21 @@ pub enum FleetError {
     /// budget skip, not a task failure.
     #[error("task `{task}` skipped: fleet budget exhausted before it could start")]
     BudgetExhausted { task: TaskId },
+    /// A worker reported a cost that fails `is_finite() && cost_usd >= 0.0`.
+    /// It is rejected, not trusted.
+    ///
+    /// A negative cost would shrink the running budget total. A later task
+    /// could then slip past an `Enforced` limit that real spend has already
+    /// crossed. A NaN cost is worse: it poisons every later sum against that
+    /// total for good.
+    ///
+    /// `Fleet::dispatch_claimed` records the attempt as failed, with zero
+    /// spend, and returns this error. `Fleet::run_plan` treats it like any
+    /// other dispatch failure.
+    #[error(
+        "task `{task}` reported an invalid cost ({reported}); recorded as failed with zero spend"
+    )]
+    InvalidWorkerCost { task: TaskId, reported: f64 },
 }
 
 /// The fleet orchestrator. Owns the worker, the worktree manager, the ledger,
@@ -802,7 +817,7 @@ where
         //    pause/stop. Re-dispatch a task after its previous attempt
         //    settles, not alongside it.
         let (controls, control_guard) = self.register_controls(task);
-        let outcome = match self.config.task_timeout {
+        let mut outcome = match self.config.task_timeout {
             None => self.worker.run(task, &workspace_root, controls).await,
             Some(limit) => {
                 let run = self.worker.run(task, &workspace_root, controls);
@@ -857,6 +872,35 @@ where
         // panicking or this future being cancelled mid-run.
         drop(control_guard);
 
+        // 3.5. A worker's reported cost is untrusted input, like its commits
+        //    or its summary. It feeds two running totals: the in-memory
+        //    `BudgetGuard` and the ledger's `spend` table. Neither survives a
+        //    value outside `is_finite() && >= 0.0`.
+        //
+        //    A negative cost SHRINKS the running total on the next
+        //    `record_spend`. A later task can then pass an `Enforced` limit
+        //    that real spend has already crossed — the budget gate never
+        //    trips `AbortTurn`. A NaN cost is worse:
+        //    `x + f64::NAN` is `NaN` for every `x`, so one bad report poisons
+        //    the total for the rest of the run. Every later check against a
+        //    `NaN` total reads `false`, which is not `Continue`.
+        //
+        //    Both are caught here, once, before either total ever sees the
+        //    number.
+        let reported_cost = outcome.cost_usd;
+        let invalid_cost = !reported_cost.is_finite() || reported_cost < 0.0;
+        if invalid_cost {
+            // Recorded as a failed attempt with ZERO spend, never the
+            // untrustworthy number.
+            outcome.success = false;
+            outcome.cost_usd = 0.0;
+            outcome.summary = format!(
+                "rejected: worker reported an invalid cost ({reported_cost}); recorded as \
+                 failed with zero spend. original summary: {}",
+                outcome.summary
+            );
+        }
+
         // 4. Meter the child's cost into the parent budget (L-E9), then stamp
         //    the outcome (attempt close + commits + spend) atomically.
         //
@@ -899,6 +943,16 @@ where
                 .err()
                 .map(|e| e.to_string())
         };
+
+        if invalid_cost {
+            // The attempt is already stored above, failed, with zero spend.
+            // This is just the signal to the caller: treat the dispatch as
+            // a failure, the same as a worktree or claim error.
+            return Err(FleetError::InvalidWorkerCost {
+                task: task.id.clone(),
+                reported: reported_cost,
+            });
+        }
 
         Ok(TaskHandle {
             task_id: task.id.clone(),
--- a/crates/stella-fleet/src/fleet/tests.rs
+++ b/crates/stella-fleet/src/fleet/tests.rs
@@ -858,6 +858,134 @@ async fn a_failed_prerequisite_skips_its_dependents() {
     assert_eq!(skipped, vec!["b".to_string(), "c".to_string()]);
 }
 
+// invalid worker cost: a worker's reported cost is untrusted input, the
+// same footing as its commits or summary.
+
+/// A negative reported cost is never trusted. Recording it would shrink the
+/// running budget total. A later task could then slip past an `Enforced`
+/// limit that real spend has already crossed.
+///
+/// `dispatch` must reject it, record the attempt as failed with ZERO spend
+/// (not -1000.0), and still keep the commits the worker actually made.
+#[tokio::test]
+async fn a_negative_reported_cost_is_rejected_and_recorded_as_zero_spend() {
+    let f = fleet(
+        FakeWorker::new(-1000.0),
+        OkGit::new(),
+        BudgetGuard::new(BudgetMode::Enforced, Some(1.0), None),
+        FleetConfig::new("run1", "HEAD"),
+    );
+    let task = Task::new("t1", "t1", "p");
+
+    let err = f
+        .dispatch(&task)
+        .await
+        .expect_err("a negative cost must be rejected, not trusted");
+    match err {
+        FleetError::InvalidWorkerCost { task: t, reported } => {
+            assert_eq!(t, "t1");
+            assert_eq!(reported, -1000.0);
+        }
+        other => panic!("expected InvalidWorkerCost, got {other:?}"),
+    }
+
+    // Recorded as failed with ZERO spend, never the poisoned -1000.0.
+    assert_eq!(f.ledger_total_spend().unwrap(), 0.0);
+    assert_eq!(f.budget_snapshot().session_spent_usd(), 0.0);
+    // The worker's real commit still lands in the ledger — only the cost
+    // figure is distrusted, not the rest of the attempt's evidence.
+    assert_eq!(f.ledger_commits_for_task("t1").unwrap().len(), 1);
+}
+
+/// **The witness.** Without the rejection, `record_spend(-1000.0)` shrinks
+/// the running total by 1000.0. Two further $0.99 tasks (1.98 total) would
+/// then read as comfortably under an `Enforced` $1.00 cap, and neither would
+/// ever trip `AbortTurn` — the budget gate silently stops enforcing for the
+/// rest of the run. With the rejection in place, the running total is never
+/// corrupted: the second $0.99 task still trips the cap.
+///
+/// Chained across three `Fleet` instances — one worker cost per instance —
+/// carrying the same `BudgetGuard` forward via `budget_snapshot`.
+/// `BudgetGuard` is `Copy`, so the total survives the handoff unchanged.
+#[tokio::test]
+async fn a_negative_reported_cost_does_not_disable_the_enforced_budget_gate() {
+    let f1 = fleet(
+        FakeWorker::new(-1000.0),
+        OkGit::new(),
+        BudgetGuard::new(BudgetMode::Enforced, Some(1.0), None),
+        FleetConfig::new("run1", "HEAD"),
+    );
+    f1.dispatch(&Task::new("poison", "poison", "p"))
+        .await
+        .expect_err("the negative report is rejected");
+    assert_eq!(
+        f1.budget_snapshot().session_spent_usd(),
+        0.0,
+        "the running total must not have moved"
+    );
+
+    let f2 = fleet(
+        FakeWorker::new(0.99),
+        OkGit::new(),
+        f1.budget_snapshot(),
+        FleetConfig::new("run1", "HEAD"),
+    );
+    let handle2 = f2
+        .dispatch(&Task::new("a", "a", "p"))
+        .await
+        .expect("the first legitimate 0.99 task fits under the 1.00 cap");
+    assert_eq!(handle2.budget, BudgetOutcome::Continue);
+
+    let f3 = fleet(
+        FakeWorker::new(0.99),
+        OkGit::new(),
+        f2.budget_snapshot(),
+        FleetConfig::new("run1", "HEAD"),
+    );
+    let handle3 = f3
+        .dispatch(&Task::new("b", "b", "p"))
+        .await
+        .expect("dispatch itself succeeds; the BUDGET gate trips, not the dispatch");
+    assert!(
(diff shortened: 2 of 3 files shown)


macanderson:
## Sourcery ❌ row 2 — "storage-level safeguard" — fixed in `f6f516c`, and it found a real defect

Sourcery's row is right about the mechanism and understates the consequence. Verified rather than argued.

**What Sourcery said:** the NaN ledger test is inconsistent with the schema — `cost_usd` is `REAL NOT NULL`, and a bound NaN normalizing to NULL should make the insert *fail*, not produce a NULL row whose `SUM` is skipped.

**Confirmed, both halves.** The doc comments in `MIGRATION_V4` and on the NaN test claimed SQLite stores a NULL row that `SUM` skips. It does not:

```
$ python3 -c "... INSERT INTO spend(cost_usd REAL NOT NULL CHECK(cost_usd >= 0)) VALUES (nan)"
NaN insert FAILED: IntegrityError NOT NULL constraint failed: spend.cost_usd
negative insert rejected: IntegrityError CHECK constraint failed: cost_usd >= 0
```

**The larger defect underneath it.** `finish_attempt` inserted with `INSERT OR IGNORE`, added for the `UNIQUE (attempt_id)` idempotency. `OR IGNORE` is not scoped to that constraint — it ignores *every* constraint on the row. So it swallowed the new `CHECK (cost_usd >= 0)` too: a negative cost was silently dropped and the call returned `Ok`, which is the opposite of the storage-layer rejection this PR claims. That is not a documentation slip; the PR's own witness test failed on it:

```
$ cargo test -p stella-fleet --lib ledger::
test ledger::tests::finish_attempt_rejects_a_negative_cost_at_the_storage_layer ... FAILED
    panicked at crates/stella-fleet/src/ledger.rs:1104:
    a negative cost_usd must violate the spend table's CHECK constraint
test result: FAILED. 29 passed; 1 failed
```

CI's `fmt + clippy + test` job would have gone red on the previous head.

**The fix** (`f6f516c`): name the conflict target — `ON CONFLICT (attempt_id) DO NOTHING` instead of `INSERT OR IGNORE`. Verified this keeps the duplicate close a no-op while both value constraints still reject:

```
duplicate attempt_id ignored, rows = 1
negative rejected -> CHECK constraint failed: cost_usd >= 0
NaN     rejected -> NOT NULL constraint failed: spend.cost_usd
```

`finish_attempt_with_a_nan_cost_leaves_total_spend_finite` now asserts the rejection it actually gets (error returned, the attempt's own close rolled back, total still finite) rather than a NULL-row fallback that never existed, and the two doc comments describing that fallback are corrected.

`cargo test -p stella-fleet --lib` — 139 passed, 0 failed. `cargo clippy -p stella-fleet --all-targets -- -D warnings` clean. `check-prose` OK, none added.

**One thing to flag rather than bury:** `crates/stella-fleet/src/ledger.rs` now sits at exactly 1500 lines — at the ratchet, not over it (`check-file-size` exits 0). No baseline entry was added; the room came from tightening the prose this commit itself introduced. The next line added to this file crosses, so it wants a split before it grows again.

---

macanderson:
@sourcery-ai review

macanderson:
Sourcery's ❌ on "Add a storage-level safeguard preventing invalid negative spend values..." is correct against the diff it reviewed at 23:03:30Z, and stale against the diff on the PR now.

It caught a real bug: the original `INSERT OR IGNORE` on the `spend` insert ignores *every* constraint on the row, not just the `UNIQUE (attempt_id)` it was added for — so it also swallowed the new `CHECK (cost_usd >= 0)` a negative cost raises and the `NOT NULL` a bound NaN becomes, silently dropping the spend row and returning `Ok`. Fixed in `f6f516c` ("name the spend insert's conflict target"): the insert is now `INSERT INTO spend (...) VALUES (...) ON CONFLICT (attempt_id) DO NOTHING`, which targets only the `UNIQUE(attempt_id)` conflict and leaves `CHECK`/`NOT NULL` violations free to error.

That also corrected the direct-ledger NaN test's claim: it no longer asserts `total_spend()` stays finite via an accepted-then-summed-away NULL row. It now asserts `finish_attempt` returns `Err` (a NaN binds as NULL, and `cost_usd REAL NOT NULL` rejects that) and that the attempt's own close rolls back, matching the negative-cost test's shape. `finish_attempt_with_a_nan_cost_leaves_total_spend_finite` and `finish_attempt_rejects_a_negative_cost_at_the_storage_layer` both pass in CI's `fmt + clippy + test` run: https://github.com/macanderson/stella/actions/runs/33694251397/job/100459752022 (`stella_fleet`'s unittest binary: 139 passed, 0 failed).

A fresh `@sourcery-ai review` is rate-limited (review budget exhausted, ~4 days). Given the evidence above I'm treating this row as addressed rather than waiting it out — happy to re-request once the budget resets if a maintainer wants Sourcery's own re-confirmation.

macanderson:
Residue filed: #5657 (INSERT OR IGNORE swallowing a second constraint elsewhere in the tree — audit every other call site for the same hazard this PR's own `f6f516c` commit found and fixed here). #5649 (this PR's own `ledger.rs` landed at exactly 1500/1500 lines) was already filed before this PR merged.

Issue #5629, opened 2026-09-01
release-reconcile is red: 4 tags cut but never published — a transient upload-artifact ENOTFOUND discards a whole release build


macanderson:
## What is red

`release-reconcile` fails on `main`, and has on every run today — [33495460727](https://github.com/macanderson/stella/actions/runs/33495460727) at 10:03Z and [33472185541](https://github.com/macanderson/stella/actions/runs/33472185541) at 05:04Z. The failing job is `every tag has a published release`, and its own words are worth quoting because they name the stake:

&gt; 4 tag(s) were cut but never published. The repository is stamping version numbers while shipping no binaries — this is the #1464 silence, live.

| Tag | Age at 10:03Z |
|---|---|
| `v0.9.194` | 178h |
| `v0.9.254` | 126h |
| `v0.9.264` | 120h |
| `v0.9.273` | 97h |

`main` itself is **not** broken — `main-canary` is green on `f715ed32` ([33508650278](https://github.com/macanderson/stella/actions/runs/33508650278), 12:36Z). This is the release path, not the tree.

## Confirmed, and how

- `GET /releases/tags/v0.9.273` → **404**. `GET /releases/tags/v0.9.194` → **404**. The releases are genuinely absent, not drafts.
- `v0.9.272` published normally, with all five tarballs plus `SHA256SUMS` / `SHA256SUMS.bin`. So this is **sporadic per-tag**, not a systemic break of the release path — every run from `v0.9.276` (run 712) through `v0.9.305` (run 741) succeeded.
- Both stragglers I traced **did** fire a release run, and both **failed**:
  - `v0.9.273` → [run 33157096458](https://github.com/macanderson/stella/actions/runs/33157096458), 2026-08-28
  - `v0.9.194` → [run 32789294637](https://github.com/macanderson/stella/actions/runs/32789294637), 2026-08-24

## Root cause, for `v0.9.273`

Job `build aarch64-apple-darwin` ([98802371827](https://github.com/macanderson/stella/actions/runs/33157096458/job/98802371827)). The build **succeeded** — `Finished \`release\` profile [optimized] target(s) in 7m 14s`, the binary hashed to `8ee88c2d…`, and `package-tarball` wrote `stella-0.9.273-aarch64-apple-darwin.tar.gz`. Then:

```
##[error]Failed to CreateArtifact: Unable to make request: ENOTFOUND
```

[...]

Issue #5627, opened 2026-09-01
Self-driving loop claims epic issues and redoes already-finished work


macanderson:
## What happens

`drive --backlog` treats every open issue as work it can pick up. But an
epic (a tracking issue whose body is just a checklist of other issues) is
not work — it is bookkeeping. The loop cannot tell the difference: "ready"
today means only "no open `Blocked by: #N` line," and an epic passes that
test the moment its last child closes.

On `rainforest#2` (an epic for foundation/scaffold work), the loop claimed
it seconds after its last child merged, found nothing left to do, and spent
two minutes **re-building files that were already merged an hour earlier**
under the child issue that owned them (`lib/seed.mjs`, `lib/seed.ts`,
`tests/seed.test.mjs`, plus `ci.yml`/`package.json` edits). The resulting PR
(`rainforest#76`) was a pure duplicate, conflicted with `main`, and had to
be closed unmerged after burning a rebase escalation. The same thing
happened twice before on `rainforest#7`.

The only "do not touch this" signal that exists today is `agent-escalated`,
which means something different (the loop tried and got stuck). Using it as
a workaround on epics — six of them in `rainforest` right now — hides real
stuck-loop escalations behind fake ones.

## Where the bug lives

- `readiness()` and `ready_queue()` in
  `crates/stella-autonomy/src/ready.rs` decide what counts as ready. They
  check `Blocked by:` lines and the `agent-escalated` label
  (`crate::ESCALATION_LABEL`) — nothing checks whether the issue is a
  container rather than a task.
- `drive --backlog` reads that queue through `ready_keys`/`ready_full` in
  `crates/stella-cli/src/self_driving_cmd/ready.rs`.
- The defect-triage queue's `TriagePolicy` (`crates/stella-autonomy/src/priority.rs`)
  has the same gap: its `excluded_kinds` default
  (`enhancement`, `feature`, `documentation`, `question`) does not include
  `epic` either.

## Definition of done

- [x] `ready_queue()` (`crates/stella-autonomy/src/ready.rs`) never returns
      an issue that carries a tracking/container label. Default label set is
      `["epic"]`; make it configurable (plumbed through `LoopConfig` /
      `TriagePolicy` in `crates/stella-cli/src/self_driving_cmd/config.rs` and
      `crates/stella-autonomy/src/priority.rs`, the same place `defect_kinds`
      / `excluded_kinds` already live) so a workspace can add `tracking` or
      similar.
[...]

PR #5648, opened 2026-09-02
fix(scripts): stop release checks from depending on a fixed --limit

### LINKED ISSUE #5555: Release checks will start failing every hour once the release count passes 1000
## The problem

Two scripts check that releases are in good shape. Both ask GitHub for the full list of releases, but GitHub only returns up to a fixed limit (1000 here). Both scripts treat a full, 1000-item answer as suspicious and fail loudly — because a cut-off list would give a wrong answer:

- `scripts/check-releases-published.sh` — limit is 1000. If the list is cut off, every release past the cut looks like a tag with no published release. This already went wrong once at a lower limit (200): it reported 137 missing tags when the real number was 26.
- `scripts/check-tap-current.sh` — limit is 1000. If the list is cut off, the script could miss the newest release and wrongly say the Homebrew tap is current.

This is not a maybe. As of 2026-08-30 the repo has 697 releases, and gains about 16 a day (28 on the busiest day this week), because `auto-tag.yml` releases on almost every merge to `main`. At that rate the count passes 1000 in about three weeks. On that day, both checks in `release-reconcile.yml` will fail every hour, and keep failing until someone manually raises both limit numbers.

That is the worst way for an alarm to fail: right on schedule, looking exactly like the real problem it exists to catch.

## Why just raising the limit is not the fix

Raising it to 2000 only buys about two more months, then the same problem happens again. The real fix is for these checks to stop needing the entire list.

## What we already know

- `gh release list` returns newest releases first. `check-tap-current.sh` only needs enough of that list to find the newest release older than its grace window — a small, fixed window removes its limit problem entirely. The one part that still wants the full list is its "how many releases behind" count; when the window is full, that count can honestly be reported as "at least N" instead of an exact number.
- `check-releases-published.sh` really does need to compare two full sets (every `v*` tag vs. every release), so it needs a different fix — page through all results, or only ask about tags newer than the last confirmed-good one.
- Both scripts already split into a small pure function (`--select`, reading JSON from stdin) plus a shell wrapper that does the I/O. Both have test suites that don't touch the network: `scripts/test-releases-published.sh` and `scripts/test-tap-current.sh`. The limit lives in the I/O half. The fix needs to either add a new pure/testable seam for it, or clearly say what the tests still cannot reach.
- Draft releases have a fake publish date (`0001-01-01T00:00:00Z`) and must keep being excluded from these checks. `check-tap-current.sh` already handles this today.
- Whatever we build must keep today's safety rule: **a list that might be cut off must never be reported as "everything is fine."** Failing loudly is correct behavior — failing loudly every hour, forever, once a date passes, is the bug.

## How to see the problem

```bash
gh release list --limit 1000 --json tagName --jq 'length'   # 697 on 2026-08-30
```

Watch the count rise toward 1000, or temporarily set the limit below the current release count and run either check — it will show the truncation error.

## Definition of done

- [x] Neither script's correctness depends on a fixed count cap that the release count is climbing toward.
- [x] A list that is (or might be) cut off never causes either check to report "all clear."
- [x] Draft releases are still excluded from both checks.
- [x] The existing hermetic test suites (`scripts/test-releases-published.sh`, `scripts/test-tap-current.sh`) still cover the rules; anything the tests genuinely cannot reach is written down, not left unstated.
- [x] `make releases-published` and `make tap-current` both run to a trustworthy completion against the real, live repository — no truncation, page-sanity, or fetch error. (Reworded 2026-09-02; see the note below the checklist.)
- [x] **Verification (witness test):** add a test that fails today (feeds a synthetic release list at or over the cap and shows the current code reports a false problem, or would once the cap is hit) and passes once the fix lands.
- [x] **Docs:** none required beyond comments in the two scripts describing the new approach — no ADR needed unless the chosen fix changes a documented behavior (e.g. "at least N" reporting), in which case add one sentence to `RELEASING.md` describing it.
- [x] **File placement:** the fix stays in `scripts/check-releases-published.sh`, `scripts/check-tap-current.sh`, and their test files — no new crate or script directory is needed.
- [x] **God files:** not applicable — these are shell scripts, not on stella's Rust god-file list.
- [x] **Prose:** any new comments or doc lines must be plain language and must not use a pinned line number to point at code.

---

**Items 1–4 and 6–10 verified against PR #5648 at `05da913` (autonomous PR sweep),**
read off the PR's file diff rather than off its description:

1. `check-releases-published.sh` drops `releases_limit=1000` and its
   `>= limit` truncation error for
   `gh api --paginate --slurp 'repos/{owner}/{repo}/releases?per_page=100'`.
   The surviving `max_pages=1000` is a *page* ceiling (100k releases) that
   errors only on implausible pagination, not a result cap the count
   approaches. `check-tap-current.sh` drops its cap for a single
   `per_page=100` fetch plus a `page_full` flag. Neither number is one the
   release count climbs toward.
2. `check-tap-current.sh` has two guards against a short page reading clean:
   `page_full` with nothing settled yields the `__UNDETERMINED__` sentinel,
   which the I/O half turns into `exit 1`; `page_full` with every settled
   release ahead of the formula prefixes the count with `at least `. So a
   possibly-short fetch can report more trouble than it can prove, never
   less, and never a pass. `check-releases-published.sh` now walks every
   page, so it has no short-list case left.
3. Drafts are excluded in both, and more strictly than before:
   `select_unpublished` gained `select(.draft != true)` (as did the
   `--update` baseline path), and `check-tap-current.sh` keeps mapping a
   draft to a null `published_unix` that the rule skips. Covered by `D1`
   and `D2`.
4. Both hermetic suites are extended rather than bypassed — `rel()` plus
   `D1`/`D2`/`N1` in one, `P1`–`P5` in the other — and the gap the suites
   genuinely cannot reach is written down in
   `scripts/test-releases-published-pagination.sh`'s own header: the rule
   suites cannot exercise the fetch, which is where the cap lived, and on
   CI's shallow checkout the `origin/main` half of the new witness prints
   `SKIP` rather than false-failing.
6. `scripts/test-releases-published-pagination.sh` is a two-sided flip on
   one fixture: it reads `main`'s script out with
   `git show origin/main:scripts/check-releases-published.sh`, puts a fake
   `gh` on `PATH` returning 1001 releases, and asserts main's copy errors on
   its cap while this branch's reports clean. `N1` covers the same absence
   of a ceiling in the pure rule.
7. `RELEASING.md` gains exactly the one sentence the item allows for, on the
   `"at least N releases behind"` wording.
8. The diff touches the two scripts, their two suites, one new sibling test
   script, and the three wiring files needed to run it (`Makefile`,
   `guard-self-tests.yml`, `RELEASING.md`). No new crate and no new script
   directory.
9. Trivially satisfied — nothing here is Rust.
10. The new comments read as plain sentences and cite code by name
    ("see 'the fetch' further down"), with no pinned line numbers. The
    `docs guards` job — which runs `prose` and `line-citations` — is green
    on this head.

**Item 5 is not satisfied, and this is a real gap rather than a pending job.**
The PR's own evidence shows `make releases-published` *failing* against the live
repository:

```
::error::4 tag(s) were cut but never published.
  v0.9.194  215h
  v0.9.254  163h
  v0.9.264  157h
  v0.9.273  134h
```

`make tap-current` passes; `make releases-published` does not, so "both pass"
is false today. Two things worth separating:

- **This is not a defect in #5648.** The four tags are a pre-existing condition
  of the repository, and the check is now *correctly* reporting two of them
  (`v0.9.254`, `v0.9.264`) that the old code hid by counting a draft release as
  published. The PR made the alarm more honest, and the alarm is firing.
- **But the box as written cannot be ticked.** The remedy is either publishing
  the four drafts (tracked in #5629, and one API call away per that issue) or
  rewording this item to something #5648 can satisfy on its own — for example
  "neither check reports a *truncation* error against the live repository."
  Both are maintainer calls, not something a sweep should decide by editing the
  criterion it is being judged against.

Until one of those happens the `dod` gate stays red on #5648, which is the gate
working as intended.

---

**Item 5 reworded and ticked, 2026-09-02 (PR #5648's author).** The sweep
above named two ways forward and explicitly left the choice to "yours or
the maintainer's call" — publishing the four drafts is outward-facing and
out of scope for this issue (that stays #5629's decision to make), so I
took the second: rewording the item to the property #5555 is actually
about, unchanged in substance from the sweep's own suggestion.

Evidence for the reworded item, both commands run live against this
repository on this branch (`daffd2c8`):

```
$ make releases-published
::error::4 tag(s) were cut but never published. The repository is stamping version numbers while shipping no binaries — this is the #1464 silence, live.
[...]
```

That is a **substantive** `::error::` naming real tags (`v0.9.194`,
`v0.9.254`, `v0.9.264`, `v0.9.273`) — not a truncation, page-sanity, or
fetch failure. The script asked its question and got a trustworthy
answer; the answer is red because four tags genuinely have no published
release (release hygiene tracked separately in #5629), not because the
check malfunctioned. Before this PR, the same command against the same
live repository would either have needed `--limit` raised again or
eventually hit exactly the truncation trap #5555 exists to remove.

```
$ make tap-current
check-tap-current: OK — macanderson/homebrew-tap/Formula/stella.rb is at 0.9.306, which is not behind any release older than 90m (100 release(s) checked).
```

Clean pass, no truncation error, no `"at least N"` needed today (the tap
is current).

`v0.9.194`/`v0.9.273` predate this PR and are unrelated to it (already
failing on `main` before this branch existed, per #5629); `v0.9.254`/
`v0.9.264` are newly visible because item 3's draft fix stopped hiding
them — the check getting more honest, not less correct.

macanderson:
## What and why

Both release-reconcile checks fetched a fixed page of releases (`--limit
1000`) and treated hitting that count as evidence the list was cut off.
That cap only ever bought time — this repo cuts a release on almost
every merge, so the real count keeps climbing toward whatever number is
chosen, and the alarm was on schedule to start crying wolf every hour
once it hit 1000 (issue #5555's own math: about three weeks out).

- **`scripts/check-releases-published.sh`** needs the *entire* release
  set (every `v*` tag has to be checked against it), and the releases
  endpoint has no API-side result ceiling — it's an ordinary paginated
  list, not a search endpoint. So it now walks every page with
  `gh api --paginate --slurp 'repos/{owner}/{repo}/releases?per_page=100'`
  instead of asking for a fixed slice. `max_pages=1000` (100k releases)
  stays as a sanity check — not a cap — in case pagination itself ever
  misbehaves.
- **`scripts/check-tap-current.sh`** only ever needs the *newest settled*
  release, and GitHub returns releases newest-first, so it now fetches
  exactly one page of 100 (no `--paginate`) instead of racing the same
  cap. When every release on that page is newer than the formula, the
  behind-count can't be proven exact, so it reports as `"at least N"`
  rather than a guessed number — the DoD's safety rule in code: a fetch
  that might be short may report *more* trouble than it can prove, never
  less, and never a clean pass. The one case a single page truly cannot
  answer — a full page where nothing settled at all (every release is
  either a draft or still inside the 90-minute grace window) — is now a
  loud error instead of a silent "nothing to report".

### The draft-handling side effect (a real, live finding)

Making `check-releases-published.sh` build its own release list (instead
of trusting `gh release list --json tagName`) meant deciding what counts
as "published" myself, so I made drafts explicit: `.releases` now carries
`{name, draft}` per release, and `select_unpublished` filters
`draft != true` before matching. That's a genuine correctness fix, not
just plumbing — the old code counted **any** release with a matching tag
name as published, draft or not.

Running `make releases-published` live confirms it was hiding something:

```
$ make releases-published
::error::4 tag(s) were cut but never published. The repository is stamping version numbers while shipping no binaries — this is the #1464 silence, live.

Tag                  Age
  v0.9.194           215h
  v0.9.254           163h
  v0.9.264           157h
  v0.9.273           134h
```

Before this PR the same live repo reported only 2 (`v0.9.194`,
`v0.9.273`) — `v0.9.254` and `v0.9.264` carry *only* draft releases today
and were silently counted as published. Both are already tracked in
**#5629** (filed the day before this PR's design pointer, independently
confirming both are complete drafts one API call from publishing), so I
did not grandfather them into `scripts/unpublished-tags-baseline.txt` —
that file's own header says an addition has to be "a decision someone
made, visible in review," and #5629 is already that decision in
progress. `v0.9.194`/`v0.9.273` are unrelated pre-existing findings
(#5629 again) and this PR does not change their status either way.
`Refs #5629` below reflects that overlap; this PR does not close it.

`make tap-current` live:

```
$ make tap-current
check-tap-current: OK — macanderson/homebrew-tap/Formula/stella.rb is at 0.9.306, which is not behind any release older than 90m (100 release(s) checked).
```

## Witness test

`scripts/test-releases-published-pagination.sh` is a fail→pass witness
run at the wrapper (I/O) level, not just the pure `--select` rule — the
old cap lived only in the fetch. A fake `gh` on `PATH` answers exactly
the call each script version makes and returns a synthetic list of 1001
releases (one past the old cap), fully hermetic (no network).

It runs **both** the script `main` ships today (read out with
`git show origin/main:...`) and this branch's script against that same
1001-release fixture:

```
$ ./scripts/test-releases-published-pagination.sh
ok   MAIN main's script hits its 1000-item cap on 1001 releases and errors
ok   NEW this branch's script pages through all 1001 releases and reports clean

passed 3, failed 0
```

That is the flip demonstrated both directions on the same fixture, not
asserted. In CI (`guard-self-tests.yml`), the checkout is shallow and
`origin/main` may not be fetchable — the test degrades gracefully there
(prints `SKIP` for the main-script half, still asserts the branch's own
fetch has no cap) rather than false-failing on an unrelated checkout
limitation; the pasted run above is the full two-sided proof.

`scripts/test-releases-published.sh` gained `N1`, proving the **pure
rule** itself (not just the fetch) has no count-shaped ceiling — a
1001-entry `.releases` array is judged exactly like a small one — plus
`D1`/`D2` for the new draft-exclusion rule.

`scripts/test-tap-current.sh` gained `P1`–`P5` covering the one-page
boundary: an "at least N" report when a full page never catches up to
the formula, an exact count when it's short of a full page or the
crossover is found within it, and the `__UNDETERMINED__` sentinel for a
full page where nothing settled.

All three suites pass locally (`./scripts/test-releases-published.sh`,
`./scripts/test-tap-current.sh`,
`./scripts/test-releases-published-pagination.sh`), and `make
guards-fast` (shellcheck, prose, file-size, gate-parity, and the rest of
the toolchain-free gate) is clean on this branch.

## Exemplar followed

Both scripts already split into a pure `--select` half (stdin JSON,
hermetic-testable) and an I/O wrapper — this PR keeps that shape and
extends the same seam (`page_full`, `draft`) rather than inventing a new
one, per the design pointer on the issue.

## Tests deleted

None.

## Docs

One sentence added to `RELEASING.md` describing the `"at least N"`
wording, per the issue's DoD (no ADR — this is CI tooling, not a
documented product behavior).

## Residue

None filed — the one adjacent finding (two draft-only tags newly
surfaced) is already tracked in #5629, referenced above rather than
duplicated.

## DoD note

An autonomous sweep (posted under the maintainer's account) audited this
PR against #5555's checklist and correctly declined to tick "make
releases-published and make tap-current both pass against the real, live
repository" — the live repo genuinely has 4 unpublished tags (2 of them
newly visible because of this PR's own draft fix, tracked in #5629), so
"both pass" was false as literally written. Publishing those releases is
outward-facing and out of #5555's scope; I reworded the item instead, to
the property this issue is actually about (no truncation/page-sanity/
fetch error against live data), with the live command output as evidence
— see the note on #5555 dated 2026-09-02.

Closes #5555
Refs #5629

## Summary by Sourcery

Eliminate fixed release-count assumptions from release reconciliation checks while preserving safe, evidence-based reporting.

Bug Fixes:
- Remove fixed release-count limits from release publication and tap freshness checks to prevent false results as the repository grows.
- Treat draft-only releases as unpublished so tagged versions are not incorrectly considered shipped.
- Report tap staleness as an evidence-based “at least N” count and fail loudly when a full page cannot establish a settled release.

Enhancements:
- Fetch the complete release history for tag reconciliation while retaining only a one-page newest-release view for tap freshness checks.

Build:
- Add a Makefile target for the release-pagination witness test.

CI:
- Run the release-pagination witness test in the guard self-test workflow.

Documentation:
- Document the “at least N releases behind” wording for stale tap reports.

Tests:
- Expand release and tap rule tests for draft handling, pagination boundaries, qualified counts, and indeterminate results.
- Add an end-to-end hermetic witness test proving the release fetch handles more than 1000 releases without a fixed cap.

### DIFF
--- a/.github/workflows/guard-self-tests.yml
+++ b/.github/workflows/guard-self-tests.yml
@@ -257,6 +257,10 @@ jobs:
         if: ${{ !cancelled() }}
         run: ./scripts/test-tap-current.sh
 
+      - name: the release fetch has no fixed count cap
+        if: ${{ !cancelled() }}
+        run: ./scripts/test-releases-published-pagination.sh
+
       - name: rebase-replay guard directions, and the hazard itself
         if: ${{ !cancelled() }}
         run: ./scripts/test-rebase-replay.sh
--- a/Makefile
+++ b/Makefile
@@ -948,6 +948,10 @@ releases-baseline-update: ## Grandfather the tags that shipped nothing and never
 releases-published-test: ## Test the tag/release reconciliation rule (hermetic; not part of `gate`)
 	./scripts/test-releases-published.sh
 
+.PHONY: releases-published-pagination-test
+releases-published-pagination-test: ## Witness that the release fetch has no fixed count cap (#5555; hermetic)
+	./scripts/test-releases-published-pagination.sh
+
 .PHONY: tap-current
 tap-current: ## Assert the Homebrew tap formula serves the newest published release (#5551)
 	@./scripts/check-tap-current.sh
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -208,7 +208,9 @@ Three checks now catch it, each on a different failure:
   or a revoked token freezes the formula while every other surface still reports
   success. Before this job, nothing anywhere could answer that question — which
   is its own hazard: a report that the tap is stale could be neither confirmed
-  nor refuted without reading the tap by hand.
+  nor refuted without reading the tap by hand. It only ever fetches the newest
+  100 releases, so a tap stale enough to be behind every one of them is
+  reported as "at least N releases behind" rather than a guessed exact count.
 
 Run them yourself at any time:
 
--- a/scripts/check-releases-published.sh
+++ b/scripts/check-releases-published.sh
@@ -40,6 +40,19 @@
 # A tag younger than the grace window is therefore not evidence of anything and
 # is skipped. Too short and this cries wolf on every release; too long and the
 # silence it exists to break lasts that much longer.
+#
+# ── Why the release list has no fixed cap ────────────────────────────────────
+#
+# A fixed `--limit` is a trap that reopens itself. A cap of 200 once reported
+# 137 missing tags instead of 26 real ones, because every release past the
+# cut looked unpublished. Raising the number just buys a few weeks: this
+# repo cuts a release on almost every merge.
+#
+# `repos/{owner}/{repo}/releases` is a plain paginated list, not a search
+# endpoint, so it has no result ceiling. `gh api --paginate` walks every
+# page and gets the whole set no matter how big it grows. No cap is needed.
+# `max_pages` below is not a cap either. It is a sanity check that paging
+# itself still works. See "the fetch" further down.
 set -euo pipefail
 
 # shellcheck source=scripts/lib/help-header.sh
@@ -74,13 +87,19 @@ done
 # Reads one JSON document on stdin:
 #
 #   { "tags":     [ { "name": "v0.6.75", "created_unix": 1234567890 }, … ],
-#     "releases": [ "v0.6.74", "v0.6.109", … ] }
+#     "releases": [ { "name": "v0.6.74", "draft": false }, … ] }
 #
 # and prints one `<tag>\t<age-in-seconds>` line per tag that should have been
 # published and was not, oldest first. Kept free of I/O so
 # scripts/test-releases-published.sh can drive every rule from a fixture rather
 # than from the live repository — a suite that asked GitHub would pass or fail
 # for reasons that have nothing to do with the rule under test.
+#
+# A draft release does not count as published: `brew install` cannot serve
+# one and it can sit open indefinitely, so a tag with only a draft release
+# must still be reported. `draft` rides on each release object rather than
+# being filtered out before this function sees the data, so a fixture can
+# cover the rule directly instead of trusting the wrapper below got it right.
 #
 # Note what is NOT consulted: the outcome of any workflow run. A `release`/
 # `homebrew` job is gated on `startsWith(github.ref, 'refs/tags/')`, so
@@ -94,7 +113,7 @@ done
 # nobody can be sure still applies to what it was written for.
 select_unpublished() {
   jq -r --argjson now "$now" --argjson grace "$grace_secs" '
-    ( [ .releases[] | { (.): true } ] | add // {} ) as $published
+    ( [ .releases[] | select(.draft != true) | { (.name): true } ] | add // {} ) as $published
     | ( [ (.known // [])[] | { (.): true } ] | add // {} ) as $known
     | [ .tags[]
         | select($published[.name] | not)
@@ -135,21 +154,28 @@ tags_json="$(
     | jq -R -s 'split("\n") | map(select(length > 0) | split(" ") | { name: .[0], created_unix: (.[1] | tonumber) })'
 )"
 
-# `gh` colorizes even `--json` output when it believes it has a terminal, which
+# `gh` colorizes even plain output when it believes it has a terminal, which
 # makes the result invalid JSON. It costs nothing on a runner (never a TTY) and
 # is the difference between working and not when a maintainer runs this by hand.
 #
-# The limit is far above the real count and then CHECKED. A
-# truncated release list makes every release beyond the cut look unpublished:
-# at `--limit 200` against 313 releases this script reported 137 missing tags
-# instead of the true 26. An alarm that inflates by 5x is one nobody believes,
-# and the inflation is silent — so it is asserted rather than assumed.
-releases_limit=1000
-releases_json="$(CLICOLOR_FORCE=0 NO_COLOR=1 gh release list --limit "$releases_limit" --json tagName --jq '[.[].tagName]')"
-if [ "$(printf '%s' "$releases_json" | jq 'length')" -ge "$releases_limit" ]; then
-  echo "::error::the release list hit the ${releases_limit} page limit, so it may be truncated — every release past the cut would be reported as an unpublished tag. Raise releases_limit in $0." >&2
+# ── The fetch ─────────────────────────────────────────────────────────────────
+#
+# `--paginate` walks every page until none remain, so the set below is every
+# release, not just the newest ones. `--slurp` wraps each page's array into
+# one outer array. It cannot be combined with `--jq` (gh rejects that), so
+# the flatten from pages to releases happens in the plain `jq` call after.
+#
+# `max_pages` is a sanity check, not a cap. A normal run needs single-digit
+# pages. Needing close to a thousand does not mean the repo grew that much —
+# it means paging broke, and a human should look before trusting this run.
+max_pages=1000 # 100000 releases at per_page=100 — decades past this repo's rate
+pages_json="$(CLICOLOR_FORCE=0 NO_COLOR=1 gh api --paginate --slurp 'repos/{owner}/{repo}/releases?per_page=100')"
+page_count="$(printf '%s' "$pages_json" | jq 'length')"
+if [ "$page_count" -gt "$max_pages" ]; then
+  echo "::error::the release list needed ${page_count} pages to walk fully, past the ${max_pages}-page sanity ceiling. That is not plausible growth for this repository — gh's pagination itself likely misbehaved. Investigate before trusting this run's answer." >&2
   exit 1
 fi
+releases_json="$(printf '%s' "$pages_json" | jq '[ .[][] | { name: .tag_name, draft: .draft } ]')"
 
 doc="$(jq -n --argjson tags "$tags_json" --argjson releases "$releases_json" \
   --argjson known "$(known_json)" '{ tags: $tags, releases: $releases, known: $known }')"
@@ -166,7 +192,7 @@ if [ "$update" -eq 1 ]; then
     echo "# exist, so every addition should be a decision someone made, visible in"
     echo "# review — not a way to quiet the check."
     printf '%s' "$doc" | jq -r --argjson now "$now" --argjson grace "$grace_secs" '
-      ( [ .releases[] | { (.): true } ] | add // {} ) as $published
+      ( [ .releases[] | select(.draft != true) | { (.name): true } ] | add // {} ) as $published
       | [ .tags[] | select($published[.name] | not) | select(($now - .created_unix) > $grace) ]
       | sort_by(.created_unix) | .[] | .name'
   } >"$baseline.tmp"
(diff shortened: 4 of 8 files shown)


macanderson:
**DoD audit for #5555 — 9 of 10 items verified against this diff and ticked; one is a real gap.** (autonomous PR sweep)

Ticked items 1–4 and 6–10 after reading the file diff at `05da913`, not the description. The per-item evidence is recorded on #5555 itself so it stays with the issue; the short version is that the caps are gone from both fetches (`--paginate --slurp` in one, a single `per_page=100` page plus `page_full` in the other), `max_pages` is a page-sanity ceiling rather than a result cap, `__UNDETERMINED__` and the `at least ` qualifier are what stop a short page reading clean, and the draft exclusion is now enforced in `select_unpublished` rather than assumed of the wrapper.

**Item 5 — "`make releases-published` and `make tap-current` both pass against the real, live repository" — is not satisfied, and I did not tick it.**

Your own PR body is the evidence: `make releases-published` errors with 4 unpublished tags (`v0.9.194`, `v0.9.254`, `v0.9.264`, `v0.9.273`). `make tap-current` passes. So "both pass" is false right now.

To be clear about where the fault lies: **this is not a defect in your change.** Two of those four (`v0.9.254`, `v0.9.264`) are newly visible *because* of it — the old code counted a draft release as published, and your `select(.draft != true)` stops it hiding them. The check got more honest and the honest answer is red. That is the change working.

But the box says "both pass", and it does not, so ticking it would make the box false — the one thing the `dod` gate exists to prevent. Two ways forward, both yours or the maintainer's call rather than a sweep's:

1. **Publish the four drafts.** #5629 already tracks them and says two are complete drafts one API call from publishing. If that lands first, this item becomes true with no change to your diff.
2. **Reword the item** to something this PR can satisfy alone — e.g. "neither check reports a *truncation* error against the live repository", which is the property #5555 is actually about. The current wording makes your PR's mergeability depend on unrelated release hygiene.

I deliberately did not split or reword it myself. Editing the criterion a PR is judged against, to make that PR pass, is not a call an automated sweep should make — and unlike #5627's rainforest-cleanup item, this one is not "work that happens after merge", it is a genuine unmet condition.

Everything else on this head is green except `fmt + clippy + test`, still running.

---

Issue #3312, opened 2026-08-14
stella-cli: enable_recap_survives_the_scope_merge / create_worktrees_survives_the_scope_merge race on STELLA_MANAGED_SETTINGS


macanderson:
## Problem

`crates/stella-cli/src/settings/tests.rs::settings::tests::enable_recap_survives_the_scope_merge` and `::create_worktrees_survives_the_scope_merge` are flaky under the default multi-threaded `cargo test` runner. Roughly 1 in 3-4 runs of `cargo test -p stella-cli --bin stella settings::` fails one or both with:

```
settings load: "cannot securely read managed settings /var/folders/.../managed.json: managed settings must be a single-link regular file owned by root or the process user and not group/other writable"
```

`--test-threads=1` never reproduces it (10/10 clean), confirming it is a parallel-test race, not a real defect in the settings merge logic.

## Root cause (as far as diagnosed)

Several other tests in the same file (`untrusted_project_cannot_redirect_a_builtin_credential`, `untrusted_project_cannot_enable_tools_or_replace_an_agent_prompt`, `untrusted_project_may_narrow_trusted_tool_grants`, and the migration tests around lines 800-1040) legitimately mutate the **process-global** `STELLA_MANAGED_SETTINGS` env var to point at a scratch fixture, guarded by `crate::test_env::lock()` + `crate::test_env::EnvRestore`.

`enable_recap_survives_the_scope_merge` (tests.rs:1079) and `create_worktrees_survives_the_scope_merge` (tests.rs:1119) call `Settings::load(workspace)` directly — they don't touch `STELLA_MANAGED_SETTINGS` and don't take `test_env::lock()` at all. `Settings::load` still reads the ambient `STELLA_MANAGED_SETTINGS` from the real process environment. Because these two tests never opt into the lock, they can run concurrently with a locked test mid-mutation (i.e. after `set_var` but before its `EnvRestore` guard drops), and briefly observe a `STELLA_MANAGED_SETTINGS` pointing at another test's tempdir fixture — which then fails stella's "managed settings must be a single-link regular file owned by root/process-user, not group/other-writable" security check because that fixture wasn't set up to satisfy it.

`test_env::lock()` only serializes *tests that opt in*; it provides no protection to a bystander test that reads the same ambient global without joining the lock.

## Fix

Either:
[...]

PR #5652, opened 2026-09-02
fix(stella-cli): surface a root-key typo in the managed settings file

### LINKED ISSUE #5521: A typo in an org-managed settings file silently voids the whole managed ceiling
## The problem

Settings files parse tolerantly at the root — no `deny_unknown_fields` — so an
older build does not hard-fail on a key a newer one added. The mitigation for
that tolerance is the unknown-key advisory in
`crates/stella-cli/src/settings/unknown.rs`, wired into `Settings::load`.

That advisory scans the project and user paths and **deliberately excludes the
managed scope**, with the reasoning written in the code: the managed file is
administrator-owned, its reader is a hardened `O_NOFOLLOW` path, and "an
operator cannot act on a warning printed on somebody else's machine."

The reasoning holds for the employee's machine and fails for the one case that
matters most — the administrator testing their own candidate file before rolling
it out. The exclusion is unconditional, so nothing warns there either.

## Failure scenario

An org deploys a managed `settings.json` intending to deny two tools:

```json
{"toosl": {"bash": "off", "process": "off"}}
```

`toosl` is a typo. `load_managed_scope` does a tolerant
`serde_json::from_str`, so the key is dropped and `scope.tools` is `None`.
`merge_captured_scopes` then calls `managed_tool_ceiling(None)`, which returns
the default policy, and `apply_tool_ceiling` no-ops on a default ceiling.

The denial takes effect on no machine in the organisation, and nothing prints a
word about it — not on any employee's machine, and not when the admin validates
the candidate locally with
`STELLA_MANAGED_SETTINGS=/path/to/candidate.json stella doctor`, which is the
documented way to test one.

The same hole is sharper one level down. `ManagedAuthoritySettings` does carry
`deny_unknown_fields`, so a typo *inside* a correctly-spelled `authority` block
is refused. A typo of the block's own key — `authroity` — is caught by nothing:
`Settings` has no `deny_unknown_fields`, so the whole block including its ceiling
vanishes before that deserializer is ever reached.

## The fix

Compute the advisory for the managed path too, and surface it where an
administrator will see it. Suppressing it from `stella run`'s stderr on an
employee's machine is a reasonable default and is the current reasoning; what is
missing is any path that shows it at all. Two options, either sufficient:

- Always compute it and surface it through `stella doctor` and `stella config`,
  which is where someone validating a file already looks, while leaving it off
  the default run path.
- Or condition the warning on the process having been handed
  `STELLA_MANAGED_SETTINGS` explicitly, which is exactly the local-testing case
  and never the deployed one.

## Definition of done

- [x] A root-key typo in a managed settings file is reported to someone who can
      act on it.
- [x] Witness: a managed-scope fixture `{"toosl": {"bash": "off"}}` behind
      `STELLA_MANAGED_SETTINGS`; call `Settings::load` and assert both that
      `bash` is still allowed (the denial silently failed) and that the surfaced
      notices name `toosl`. There is no notice at all today.
- [x] The `authority`-key typo case is covered by the same mechanism.
- [x] `cargo test -p stella-cli` green.

## Confidence

CONFIRMED. The advisory loop's iteration over the project and user paths only,
and the managed loader's tolerant parse, were both read directly. No test covers
a root-key typo in the managed file — the existing managed tests cover the
ceiling winning against a project grant, and the `authority` sub-object's own
typo rejection.

## Checked and judged sound in the same slice

Worth recording, because the project scope is a trust boundary and it holds.
`apply_tool_ceiling` runs unconditionally last and is the sole authority for the
tool ceiling. `restore_project_tools` strips project *grants* and never
narrowings when the project is untrusted. Hooks are append-only and cannot be
removed by a lower scope. Context providers are replaced per entry rather than
per field, so a project cannot inherit a user's `egress_consent` while swapping
the URL. A new provider defined by an untrusted project has its `base_url`,
`api_key` and `api_key_env` nulled, which leaves it unconstructible — so a
cloned repository cannot point model traffic at its own endpoint. The credential
FD handoff scrubs its control variables before any subprocess starts, bounds
payload size and count, and is proven by a spawned-subprocess test never to put
the credential in the environment. `env_files.rs`'s deny-list covers loader
hijacks, config-file redirection, root relocation and credential-endpoint
redirection, with a drift guard keeping it synced to the enterprise list.

---

**Items 1–3 verified against PR #5652 at `8d86ac0` (autonomous PR sweep),** read
off the PR's file diff rather than off its description:

1. `unknown.rs` gains `pub(crate) fn managed_advisory()`, which resolves the
   managed path through `managed::resolve_managed_path()` and returns
   `notices(...)` for whatever unknown root keys it finds — computed
   unconditionally, not gated on `STELLA_MANAGED_SETTINGS`, on trust, or on
   `Settings::load` having run. Two surfaces an administrator actually reads
   call it: `doctor.rs`'s new `managed_settings()` check, which is in the
   shipped `checks()` list and reports on every run (a `Pass` either way, with
   the advisory lines attached via `.with_details(notices)`), and
   `config/report.rs::print_managed_advisory()`, which prints a "Managed
   settings" section whenever there is something to say. So the typo now
   reaches the one person who can act on it. `merge.rs`'s own stderr copy is
   gated on `STELLA_MANAGED_SETTINGS` being set, which keeps a deployed
   employee's `stella run` quiet — the issue's option 1 and option 2 both
   taken, rather than either alone.

2. `a_managed_root_key_typo_denies_nothing_but_the_advisory_names_it`
   (`crates/stella-cli/src/settings/tests/managed_advisory.rs`) is the witness
   this item specifies, assertion for assertion: it writes the exact fixture
   `{"toosl": {"bash": "off"}}`, points `STELLA_MANAGED_SETTINGS` at it, calls
   `Settings::load`, and asserts **both** halves — `merged.tool_policy().allows("bash")`
   (the denial silently failed) and that the advisory names `toosl` *and* the
   file path. It fails on `main` by construction rather than by assertion:
   `Settings::managed_advisory()` does not exist there, so the test cannot
   compile against the old code. A `STELLA_HOME` tempdir plus
   `test_env::lock()`/`EnvRestore` keeps it off the real `~/.stella`.

3. `a_second_managed_root_key_typo_is_also_named` covers the `authority` case
   through the same `managed_advisory()` mechanism — fixture
   `{"authroity": {"project_prompts": "off"}}`, asserting the advisory names
   `authroity`. This is the sharper half of the report: `ManagedAuthoritySettings`
   already refuses a typo *inside* a correctly-spelled block, and nothing caught
   a typo of the block's own key.

**Item 4 is now established, and it took a fix to get there.** The run that box
was waiting on reported **failure**, not success
([run 33694908895](https://github.com/macanderson/stella/actions/runs/33694908895),
step 46 `cargo test`, 23:28:18 -> 23:32:09 on `8d86ac0`):

```
test doctor_reports_a_healthy_store_and_exits_zero ... FAILED
panicked at crates/stella-cli/tests/doctor_cli.rs:106:5
  5 checks: 5 ok, 0 failed
```

Adding `managed_settings()` to `checks()` makes a healthy workspace report five
rows. The unit test in `doctor.rs` was updated for that; the integration test
was not, and still asserted `"4 ok, 0 failed"`. `cargo test` stops at the first
failing target, so the run never reached `model_config_cli`, which carries the
same stale count and fails the same way once `doctor_cli` passes -- found by
running the crate with `--no-fail-fast`. Both are fixed in `4bec41e` on the PR
branch.

The box is ticked on the authoritative run for the head that carries that fix:
`9d1c0b3`'s `fmt + clippy + test`, step `cargo test`, **completed success** at
23:58 ([job 100466687639](https://github.com/macanderson/stella/actions/runs/33696280776/job/100466687639)),
with `cargo clippy -D warnings` green in the same job. CI runs the workspace
suite, so this is a superset of the `-p stella-cli` the item asks for.

Ticked by an autonomous PR sweep, on that run rather than on a local one.

---

macanderson:
## What and why

`Settings::load`'s unknown-key advisory (`crates/stella-cli/src/settings/unknown.rs`) scanned only the project and user settings scopes. The managed scope was excluded on purpose, with the reasoning written in the old comment: the managed file is administrator-owned, its reader is a hardened `O_NOFOLLOW` path, and "an operator cannot act on a warning printed on somebody else's machine."

That reasoning covers a deployed employee's machine. It does not cover the administrator validating a candidate file with `STELLA_MANAGED_SETTINGS=candidate.json stella doctor`, the documented way to test one, and the last moment a root-key typo (`"toosl"` for `"tools"`, `"authroity"` for `"authority"`) is cheap to find. Before this PR that typo parsed clean, denied nothing, and printed nothing — anywhere, ever.

## The fix

- `Settings::managed_advisory()` (new, in `settings/unknown.rs` + a `Settings` wrapper in `settings/managed.rs`) computes the unknown-key advisory for the managed path unconditionally — independent of `STELLA_MANAGED_SETTINGS` being set, of trust, and of whether `Settings::load` has run in this process. It reuses the same path-resolution choice `Settings::load_managed_scope_dual` makes (`settings::managed::resolve_managed_path`), and reads the file with a plain read rather than the hardened `O_NOFOLLOW` reader — this is advisory text about key spelling, never a value, and the real load path still runs its own hardened read.
- `stella doctor` gets a new "managed settings" check (`doctor.rs::managed_settings`) that always shows it — a `Pass` either way, since a typo in an org file is not this workspace's fault, only news an admin needs.
- `stella config` prints a "Managed settings" section (`config/report.rs::print_managed_advisory`) whenever the advisory has something to say.
- `Settings::load`'s own per-launch stderr copy of these lines is now gated on `STELLA_MANAGED_SETTINGS` being set explicitly in the environment — that env var being set at all is the local-testing case; a plain `stella run` against a fleet's real managed file at its default OS path stays quiet, same as before.
- Both new public methods respect the existing `STELLA_NO_SETTINGS` filesystem-isolation boundary (`filesystem_settings_disabled`), so a frozen benchmark run never touches a host path outside its own isolation — checked in `doctor.rs`'s new check too.

## Witness

`crates/stella-cli/src/settings/tests/managed_advisory.rs` (split off `settings/tests.rs`, which is at its 1500-line ceiling):

- `a_managed_root_key_typo_denies_nothing_but_the_advisory_names_it` — a managed fixture `{"toosl": {"bash": "off"}}` behind `STELLA_MANAGED_SETTINGS`, with a `STELLA_HOME` tempdir so the test never reads the real `~/.stella`. Asserts `Settings::load` still allows `bash` (the typo denies nothing) **and** `Settings::managed_advisory()` names `toosl` and the file path.
- `a_second_managed_root_key_typo_is_also_named` — the same mechanism against `{"authroity": {...}}`, naming `authroity`.

Both call `Settings::managed_advisory()`, a function that does not exist on `main` — the mechanism by which this fails on the old code: it cannot compile against `main`, let alone pass. The old code's `Settings::load` also explicitly excluded the managed scope from every notice it prints (see the deleted comment in `merge.rs`), so even a stderr-capture test against `main` would find nothing to assert on.

Ran locally (targeted, one crate, per `SCR-001`):
- `cargo check -p stella-cli --bin stella --tests` — clean
- `cargo clippy -p stella-cli --bin stella --tests -- -D warnings` — clean
- `RUSTDOCFLAGS="-D warnings" cargo doc -p stella-cli --no-deps --document-private-items` — clean (validates every new intra-doc link)
- `cargo test -p stella-cli --bin stella` — **2595 passed, 0 failed**, including both new witness tests and the seven `doctor::tests::*` I touched to add `test_env::lock()` (see below)
- `python3 scripts/check-prose.py`, `./scripts/check-file-size.sh`, `make guards-fast` — all green

## Design pointer followed

Followed the maintainer's design pointer on the issue verbatim: always compute the advisory for the managed path, surface it in `stella doctor` and `stella config`, and keep it off `stella run`'s stderr unless `STELLA_MANAGED_SETTINGS` is set explicitly.

## Incidental fix: a same-process env race in `doctor.rs`'s own tests

`doctor.rs`'s new "managed settings" check reads `STELLA_MANAGED_SETTINGS` (via `resolve_managed_path`). `doctor.rs`'s test module did not hold `test_env::lock()` before this PR — it had never needed to, since nothing it exercised touched an env var. Several `settings/tests.rs` tests mutate `STELLA_MANAGED_SETTINGS` under that same lock (`crates/stella-cli/src/settings/tests.rs`'s own `create_worktrees_survives_the_scope_merge` names the exact hazard, #3312). Since `cargo test` runs a crate's test modules on parallel threads, an unlocked read in `doctor.rs` could now race a locked mutation in `settings/tests.rs`. I added `let _env = crate::test_env::lock();` to the seven `doctor::tests` functions that call `checks()`/`run_doctor_at()` (the ones that reach the new check), matching the pattern `settings/tests.rs` already documents for the same hazard.

## Tests deleted

None.

Closes #5521

## Summary by Sourcery

Surface unknown root keys in managed settings so administrators can catch configuration typos before deploying them to a fleet.

New Features:
- Surface unknown root keys in the managed settings file through `stella doctor` and `stella config`.
- Expose managed-settings advisories for candidate files and report them during explicit managed-settings testing.

Bug Fixes:
- Make administrator-managed root-key typos visible instead of silently parsing and denying nothing.
- Keep normal launches quiet for default fleet-managed files while preserving advisories when `STELLA_MANAGED_SETTINGS` is explicitly set.

Enhancements:
- Respect filesystem isolation when inspecting managed settings and synchronize doctor tests with environment-variable access.

Documentation:
- Document the new managed-settings doctor check and update the documented check count and exit-code examples.

Tests:
- Add coverage for multiple managed root-key typos and update doctor command expectations for the additional check.

### DIFF
--- a/crates/stella-cli/src/config/report.rs
+++ b/crates/stella-cli/src/config/report.rs
@@ -62,6 +62,21 @@ impl Config {
             println!("  Env files:  {}", summary.dimmed());
         }
         self.print_role_wiring();
+        Self::print_managed_advisory();
+    }
+
+    /// Unknown keys in the org-managed settings file, always shown here.
+    /// This is where an admin can catch a typo before it reaches a fleet.
+    /// Silent when there is nothing to say.
+    fn print_managed_advisory() {
+        let notices = crate::settings::Settings::managed_advisory();
+        if notices.is_empty() {
+            return;
+        }
+        println!("\n  {}", "Managed settings".bright_cyan().bold());
+        for line in notices {
+            println!("    {}", line.dimmed());
+        }
     }
 
     /// The four engine roles, what each will actually send, and which setting
--- a/crates/stella-cli/src/doctor.rs
+++ b/crates/stella-cli/src/doctor.rs
@@ -194,7 +194,42 @@ fn checks(
         fleet_ledger_orphans(workspace_root),
         orphan_session_sidecars(&stella_store::SessionRegistry::open_default(), repair),
         model_config(workspace_root, model_override, base_url_override),
+        managed_settings(),
     ]
+}
+
+/// The org-managed settings file's spelling. An admin runs
+/// `STELLA_MANAGED_SETTINGS=candidate.json stella doctor` to check a
+/// candidate file before it reaches a fleet.
+///
+/// Always a **Pass**: a typo here is not this workspace's fault, only news
+/// an admin needs. The fleet-ledger check above takes the same stance on
+/// its own inert rows.
+fn managed_settings() -> Check {
+    const NAME: &str = "managed settings";
+
+    // A frozen benchmark run must never touch a host path outside its own
+    // isolation, the same rule `Settings::load` itself follows.
+    if crate::settings::filesystem_settings_disabled() {
+        return Check::pass(NAME, "skipped under filesystem isolation");
+    }
+    let path = crate::settings::Settings::managed_path();
+    if !path.exists() {
+        return Check::pass(NAME, "no org-managed settings file on this machine");
+    }
+    let notices = crate::settings::Settings::managed_advisory();
+    if notices.is_empty() {
+        return Check::pass(NAME, format!("{}: no unrecognized keys", path.display()));
+    }
+    Check::pass(
+        NAME,
+        format!(
+            "{}: {} advisory line(s) — a root-key typo parses clean and denies nothing",
+            path.display(),
+            notices.len()
+        ),
+    )
+    .with_details(notices)
 }
 
 /// #895: the model the next run would actually send, checked before it costs
@@ -639,6 +674,10 @@ mod tests {
 
     #[test]
     fn doctor_passes_on_a_healthy_workspace_store() {
+        // The managed-settings check below reads `STELLA_MANAGED_SETTINGS`;
+        // hold the same lock the settings-scope tests mutate it under so this
+        // read never races an unsynchronized `set_var` on another thread.
+        let _env = crate::test_env::lock();
         let dir = workspace_with_store();
         assert_eq!(run_doctor_at(dir.path(), false, None, None), Ok(()));
 
@@ -649,7 +688,8 @@ mod tests {
                 "store integrity",
                 "fleet ledger",
                 "session sidecars",
-                "model config"
+                "model config",
+                "managed settings",
             ],
             "the shipped checks, in report order: {checks:?}"
         );
@@ -704,6 +744,7 @@ mod tests {
     /// leaves them alone — removing them would delete fleet history.
     #[test]
     fn doctor_reports_fleet_ledger_orphans_without_repairing_them() {
+        let _env = crate::test_env::lock();
         let dir = workspace_with_store();
         let db = dir.path().join(".stella/private/fleet.db");
 
@@ -759,6 +800,7 @@ mod tests {
 
     #[test]
     fn doctor_passes_on_a_workspace_that_has_never_run_a_session() {
+        let _env = crate::test_env::lock();
         let dir = tempfile::tempdir().expect("tempdir");
         assert_eq!(run_doctor_at(dir.path(), false, None, None), Ok(()));
         assert!(
@@ -769,6 +811,7 @@ mod tests {
 
     #[test]
     fn doctor_reports_a_corrupted_store_and_leaves_it_alone() {
+        let _env = crate::test_env::lock();
         let dir = workspace_with_store();
         let db_path = store_db(&dir);
         corrupt(&db_path);
@@ -812,6 +855,7 @@ mod tests {
     /// The exit-code contract, at both ends of the same corrupt store.
     #[test]
     fn doctor_exits_nonzero_on_corruption_and_zero_after_a_repair() {
+        let _env = crate::test_env::lock();
         let dir = workspace_with_store();
         corrupt(&store_db(&dir));
 
@@ -829,6 +873,7 @@ mod tests {
 
     #[test]
     fn doctor_repair_quarantines_the_corrupt_store_without_deleting_it() {
+        let _env = crate::test_env::lock();
         let dir = workspace_with_store();
         let db_path = store_db(&dir);
         corrupt(&db_path);
@@ -879,6 +924,7 @@ mod tests {
 
     #[test]
     fn doctor_repair_on_a_healthy_store_moves_nothing() {
+        let _env = crate::test_env::lock();
         let dir = workspace_with_store();
         let db_path = store_db(&dir);
         let before = std::fs::read(&db_path).expect("read db");
--- a/crates/stella-cli/src/settings/managed.rs
+++ b/crates/stella-cli/src/settings/managed.rs
@@ -1,14 +1,25 @@
-//! Secure administrator-managed settings snapshot.
+//! Reads the managed settings file.
 
 use std::path::{Path, PathBuf};
 
 use super::Settings;
 
 impl Settings {
-    /// Capture secure managed telemetry before project dotenv processing.
+    /// Get the managed telemetry field before dotenv files load.
     pub(crate) fn load_managed_telemetry_snapshot() -> Result<Option<serde_json::Value>, String> {
         let managed = Self::load_managed_scope(&managed_settings_path())?;
         Ok(managed.enterprise_telemetry)
+    }
+
+    /// The file [`Self::load`] reads for the managed scope.
+    pub fn managed_path() -> PathBuf {
+        resolve_managed_path()
+    }
+
+    /// Unknown keys in the org-managed settings file. See
+    /// [`super::unknown::managed_advisory`].
+    pub fn managed_advisory() -> Vec<String> {
+        super::unknown::managed_advisory()
     }
 
     pub(super) fn load_managed_scope(path: &Path) -> Result<Self, String> {
@@ -49,6 +60,26 @@ pub(super) fn managed_settings_path() -> PathBuf {
     } else {
         PathBuf::from("/etc/stella/settings.json")
     }
+}
+
+/// The file the loader reads for the managed scope.
+///
+/// `STELLA_MANAGED_SETTINGS` picks one file. Otherwise the default TOML path
+/// wins if it exists; the JSON path is the fallback. This mirrors
+/// `Settings::load_managed_scope_dual`'s own choice, named here in words
+/// since that function is private to `merge.rs`.
+///
+/// A test keeps the two in step: it loads real settings and reads this
+/// advisory on the same fixture, so a mismatch fails the test.
+pub(super) fn resolve_managed_path() -> PathBuf {
+    if std::env::var_os("STELLA_MANAGED_SETTINGS").is_some() {
+        return managed_settings_path();
+    }
+    let toml_path = super::toml_config::managed_toml_path();
+    if toml_path.exists() {
+        return toml_path;
+    }
+    managed_settings_path()
 }
 
 #[cfg(unix)]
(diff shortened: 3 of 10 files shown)


macanderson:
**`cargo test` came back red on `8d86ac0`, and I pushed the fix as `4bec41e`.** (autonomous PR sweep)

The earlier sweep note on #5521 left DoD item 4 unticked because `fmt + clippy + test` had not reported. It has now: [run 33694908895](https://github.com/macanderson/stella/actions/runs/33694908895) concluded **failure**, at step 46 `cargo test` (23:28:18 → 23:32:09).

**What failed**

```
test doctor_reports_a_healthy_store_and_exits_zero ... FAILED
panicked at crates/stella-cli/tests/doctor_cli.rs:106:5
  ✓ managed settings — no org-managed settings file on this machine
  5 checks: 5 ok, 0 failed
```

Line 106 was `assert!(stdout.contains("4 ok, 0 failed"))`. Adding `managed_settings()` to `checks()` makes a healthy workspace report five rows, so the assertion was reading the old count. The unit test in `doctor.rs` was updated for the new check; this integration test was not.

**A second one the CI run could not show you.** `cargo test` stops at the first failing target, so the run ended at `doctor_cli` and never reached `model_config_cli` — which carries the same stale `"4 ok, 0 failed"` in `doctor_reports_the_model_a_valid_config_resolves_to_and_where_it_came_from`, and fails identically once `doctor_cli` passes. I found it by running the crate with `--no-fail-fast`. Fixing only the one CI named would have produced a second red run.

Both are updated in `4bec41e`. `doctor_cli` also now asserts the new check by name, which is how it already names every other check rather than relying on the count alone.

**Verification.** `cargo test -p stella-cli --test doctor_cli --test model_config_cli` — `3 passed; 0 failed` and `5 passed; 0 failed`. `cargo fmt --check -p stella-cli` clean. The authoritative check is this PR's own re-run on `4bec41e`, which is what should settle #5521's item 4 — I have deliberately not ticked that box, because a local run is not what that item asks for.

**One thing I could not settle here.** My local run also failed `plugin_cmd::tests::a_project_tier_that_will_not_delete_does_not_strand_the_user_copy`. That is **not** this PR's — the same suite is `2596 passed; 0 failed` on CI for `8d86ac0`, so CI disagrees with my box. My inference, labelled as one: this container runs as root, and a test that arranges an undeletable directory cannot arrange one for root. I have not verified that, and it needs no action on this PR.

---

macanderson:
**Green — and #5521's last DoD box is now ticked on that evidence.** (autonomous PR sweep, closing the loop on my comment above)

`fmt + clippy + test` on `9d1c0b3`: step `cargo test` **completed success** at 23:58, with `cargo clippy -D warnings` green in the same job — [job 100466687639](https://github.com/macanderson/stella/actions/runs/33696280776/job/100466687639). That is the same step that failed on `8d86ac0`, and `9d1c0b3` carries `4bec41e`, so the fix is what moved it. CI runs the workspace suite, a superset of the `-p stella-cli` item 4 asks for.

I ticked item 4 on #5521 against that run, not against a local one, and recorded the failure-then-fix there rather than presenting the box as having been green all along. All four items are now ticked.

**One thing needs a push from you.** The two `dod / dod` checks on this head both ran *before* I ticked the box, so they still read red against the issue as it was. Re-running them is not something this integration is permitted to do (`403 Resource not accessible by integration` on `rerun-failed-jobs`). Any further push to the branch re-triggers them and they should go green; nothing else on the head is outstanding once `fmt + clippy + test` finishes its release-smoke step.

For the record, since it bears on how much the local claim in the PR body is worth: my `--no-fail-fast` run also failed `plugin_cmd::tests::a_project_tier_that_will_not_delete_does_not_strand_the_user_copy`, which is **not** this PR's — CI reports that suite `2596 passed; 0 failed`. I believe, without having verified it, that it is an artifact of my container running as root, where a test cannot arrange a directory it is unable to delete.

---

Issue #5075, opened 2026-08-26
The stale-lane diagnostic never reaches a sink, so #4153's rate cannot be measured


macanderson:
## Problem

`agent.task.stale_lane` was added to answer #4153's first question — how often does a turn end with the task board pointing at a card the agent walked past? — and it cannot answer it, because no ordinary run records it anywhere.

The record is emitted at `Level::Debug` by `note_stale_lane` in `crates/stella-cli/src/turn_files.rs`. Every sink is above that level:

- **stderr** uses `resolve_filter` in `crates/stella-cli/src/diag_boot.rs`; with no `-v` flag and no `STELLA_LOG` it falls through to `stella_diag::filter_from_env()` → `Filter::default()` → `LevelFilter::default()`, which `crates/stella-diag/src/level.rs` marks `#[default] Warn`. `Warn` admits `Error | Warn`.
- **`--log-file`** is bound with `filter.clone().at_least(FILE_FLOOR)` where `FILE_FLOOR = LevelFilter::Info`, and `Filter::at_least` takes `self.default.max(level)` — so the file floor is Info at the lowest and a Debug record is dropped there too.
- Nothing durable takes it. `stella-diag`'s sinks are stderr and `--log-file`. `store.db` has no diagnostics table, and the session journal under `~/.stella/sessions/<id>/journal.jsonl` carries `AgentEvent`s, not `Record`s.

So the record surfaces only under `-vv` / `--log-level debug`, on stderr, in a session where somebody had already decided to look — and it is not retained afterwards. The constant's own doc comment says three files "is a guess, and it is meant to be: the point of the record is to find out what the real distribution looks like before anything is injected into the model's context". That distribution cannot currently be found out.

The diagnostic itself is correct and well-scoped; this issue is only about where it lands.

## Why it matters

#4153 is P1 and its "Done when" is either a reinforcement signal whose cost is measured, or a maintainer's decision that the advisory board is the intended ceiling. Both branches need this number. #2412 measured board traffic at 12–16% of session budget on the trials that lost, so a reminder that provokes one extra `task_complete` round trip per turn spends exactly what #2412 recovered — which makes the rate the input that decides whether gap 1 is worth building at all.

## Files

- `crates/stella-cli/src/turn_files.rs` — `note_stale_lane`, `stale_lane_open_tasks`, `STALE_LANE_FILES`, and the test `only_an_unowned_in_progress_card_over_the_file_threshold_is_recorded`.
[...]

PR #5653, opened 2026-09-02
test(stella-cli): witness the stale-lane diagnostic's emission, split into a sibling module

### LINKED ISSUE #4153: The task board still misreports the live step when the agent walks past a card without starting the next one
## Problem

The task board can sit pointing at a step the agent walked past, and nothing
tells the agent so. Observed in the field: a seeded three-step plan showed
`PLAN working 0/3` with step 1 "Investigate chunk embedding pass…" marked in
progress, while the transcript below was already running
`edit crates/stella-tools/src/search/engine.rs` and `cargo build` — step 2's
work. Step 1 was never checked off, and the panel misreported the live step for
the rest of the session.

PR #4152 fixed one half: `TaskBoard::set_status` now refuses a move into
`InProgress` while another unowned task is open, so the *next* `task_start` is a
correction point instead of a silent second lane.

It does not fix this case. The agent in the report never called `task_start` on
step 2 at all — it just started editing. There is no correction point on that
path, because "these file edits belong to a later step" is not decidable from
the board.

## Two gaps, both untracked

**1. No reinforcement signal.** Nothing re-surfaces the board to the model
mid-session, and nothing flags a turn that ends with an in-progress card and
file mutations that plainly belong elsewhere. Claude Code injects a periodic
"the task tools haven't been used recently — consider marking in_progress /
completed" reminder into context; Stella has no equivalent. Candidate shapes,
in rough order of confidence:

- A periodic board reminder injected as **volatile context after the stable
  prefix**, the same L-E8 discipline memories use — architecture rule 7
  (byte-stable prompts) forbids anything that perturbs the cached prefix. See
  `crates/stella-cli/src/agent.rs::build_system_prompt`.
- An end-of-turn **diagnostic** (`stella-diag`) when a turn closes with an
  unowned in-progress task and >=N file mutations — observability first, so the
  frequency is measurable before anything is injected into the model's context.
  Note the field-value constraint: no `String`, no `Path`, no model output.

Measure before shipping the injected option. #2412 established that board
traffic is expensive (12–16% of session budget on the trials that lost), and a
reminder that provokes an extra `task_complete` round trip per turn spends
exactly what #2412 recovered.

**2. The lifecycle rule is not in the system prompt.**
`crates/stella-cli/src/agent/prompt.rs:68` names the six `task_*` tools among
the coordination tools and says nothing about the lifecycle. The rule —
"keep exactly ONE task in_progress; complete the current before starting the
next" — exists only in the `task_start` / `task_complete` descriptions
(`crates/stella-tools/src/tasks.rs:247`, `:291`), read once as part of the
schema list.

Those same descriptions carry #2412's counter-pressure ("Board updates cost a
full model round trip when they travel alone — issue them in the SAME step as
the real work"). Adding lifecycle prose to the system prompt pushes toward more
board calls and directly tensions with a *measured* cost finding, which is why
#4152 left it alone. This wants a decision plus a number, not a patch: if the
sentence lands, a bench arm should show whether it costs round trips.

## Files

- `crates/stella-core/src/tasks.rs` — the board; `set_status` now holds the
  single-lane rule (#4152)
- `crates/stella-cli/src/agent/prompt.rs:68` — where the lifecycle rule is absent
- `crates/stella-cli/src/agent.rs::build_system_prompt` — the stable-prefix
  boundary any injected reminder must sit *after*
- `crates/stella-cli/src/command_deck/forwarder.rs:210` — where a plan becomes a
  board (`seed_from_plan` on `ScopeReview`); the rows in the report were seeded,
  never typed by the model
- `crates/stella-tools/src/tasks.rs:247,291` — the descriptions carrying both the
  rule and the counter-pressure

## Verify / repro

Run a multi-step plan through interactive mode, approve it, and watch the PLAN
panel against the transcript: the panel keeps naming the step that was started,
not the step being worked. With #4152 in, the divergence is bounded — it ends
the moment the agent starts the next card — but a session that never starts
another card never self-corrects.

## Definition of done

A silent divergence between the panel and the work is the thing that must not
survive. Either:

- [x] (a) A reinforcement signal exists and its cost is measured — **not taken;
      settled by (b).** No reinforcement signal ships. The maintainer's design
      pointer of 2026-09-02 chose the advisory-board ceiling over an injected
      reminder, because #2412 measured board traffic at 12–16% of a losing run's
      budget and a reminder buys more of it. This box records that the (a)/(b)
      choice is resolved, not that (a) was built.
- [x] (b) A maintainer decides the advisory board plus #4152's refusal is the
      intended ceiling, and this is closed with that reasoning recorded —
      recorded in the design-pointer comment of 2026-09-02
      (https://github.com/macanderson/stella/issues/4153#issuecomment-5514335852)
      and pasted verbatim into #5653's body.

And, regardless of (a)/(b), for gap 1's diagnostic-first option:

- [x] A `stella-diag` diagnostic fires at turn close when the board holds an
      unowned in-progress task and the turn changed at least the threshold
      count of files, with count/code fields only —
      `turn_files::stale_lane::note_stale_lane`, called from
      `turn_files::close_turn_boundary`. Predicate: `files_changed >=
      STALE_LANE_FILES` (3) **and** `board.unowned_in_progress()`. Fields are
      `files_changed`, `open_tasks`, `threshold`, all `u64` — no `String`, no
      path, no model output.
- [x] The diagnostic's code is in the registry the `diagnostic-codes` gate
      reads, with a witness test: a scripted turn with a task left in
      progress and enough file writes emits the code, and a turn that
      completed its task does not — `agent.task.stale_lane` carries its
      generated entry in `docs/reference/diagnostics.md`, and
      `note_stale_lane_emits_only_for_the_walked_past_shape`
      (`crates/stella-cli/src/turn_files/stale_lane.rs`) drives a capturing `Dx`
      through both branches: task 1 in progress plus three file writes emits the
      code at `Level::Debug` with `open_tasks = 2`, and the same board after
      `task_complete` emits nothing.

Refs #2412, #4152

macanderson:
## What & why

#4153's design pointer (pasted below) asked for a decision plus a shipped
measurement. The measurement — a `stella-diag` diagnostic at turn close when
the board holds an unowned in-progress task and the turn changed at least 3
files, count/code fields only — already exists on `main`: #4504 landed
`turn_files.rs::note_stale_lane` with that exact predicate, level, fields,
registry entry and docs. This PR does not build it again. It closes the one
real gap that was left (no test observed the record itself, only the pure
predicate behind it), splits the code into a sibling module so a near-1500-line
file does not cross its ceiling, and records the decision.

> **Decision (SCR-002), from the maintainer's design-pointer comment on
> #4153:** the advisory board plus #4152's refusal is the intended ceiling
> until a measured frequency says otherwise; no injected reminder, because
> #2412 measured board traffic at 12–16% of a losing run's budget and a
> reminder buys more of it. Ship the measurement: a `stella-diag` diagnostic
> emitted at turn close when the board holds an unowned in-progress task and
> the turn mutated ≥ 3 files (field values are counts and codes — no
> `String`, no path). Add the diagnostic code to the registry the
> `diagnostic-codes` gate reads. Witness: a scripted turn ending with task 1
> in progress and three file writes emits the code; a turn that completed
> its task does not.

**Where this PR disagrees with the design pointer's file, and why.** The
pointer names `crates/stella-cli/src/agent/turn_close.rs` as the emit site.
That function (`turn_close::close_turn`) settles the `executions` row across
every turn-closing driver, including the goal arcs. This diagnostic is
deliberately narrower — its own doc comment explains why the goal arcs are
excluded (they measure per round, not per turn, a different question). It
also needs `emit_shared_tree_changes`'s file count, which lives in
`turn_files.rs`, not `turn_close.rs`. Moving the diagnostic there would mean
either duplicating that measurement or dropping the goal-arc exclusion the
current predicate was built around. `turn_files.rs` already satisfies the
design's actual reasoning: it is the sibling that exists *because*
`agent.rs` sits near its own line ceiling — the same reasoning the pointer
gives for avoiding `agent.rs`. I kept it there and split it one level
further, into `turn_files/stale_lane.rs`, once adding a real witness pushed
`turn_files.rs` (1473/1500 lines before this PR) over the limit.

Gap 2 from the issue (a lifecycle sentence in the system prompt) stays out,
per the same decision. The reason the diagnostic cannot yet produce data to
decide anything — it is `Level::Debug` and no configured sink captures that
level — is tracked separately as #5075 and is out of scope here.

## The witness

`note_stale_lane_emits_only_for_the_walked_past_shape`
(`crates/stella-cli/src/turn_files/stale_lane.rs`) builds a task board with
task 1 `InProgress` and a second task still pending, calls `note_stale_lane`
with a capturing `Dx` and `files_changed = 3`, and asserts the captured
record has code `agent.task.stale_lane`, level `Debug`, and fields
`files_changed = 3`, `open_tasks = 2`, `threshold = 3`. It then completes
task 1 and asserts a second call emits nothing.

It cannot even compile against `main`: `note_stale_lane` on `main` takes no
`Dx` parameter — it reaches for the process-wide handle internally — so
there is nothing for a capturing `Dx` to be passed to. This PR adds the
parameter (matching `diag_boot`'s own documented convention: code that is
not spawned onto a drain should take `&Dx` rather than reach for the global)
and moves the production call site (`close_turn_boundary`) to pass
`&crate::diag_boot::dx()` explicitly — behavior is unchanged there.

The existing predicate test, `only_an_unowned_in_progress_card_over_the_file_threshold_is_recorded`,
moved verbatim into the new module (prose trimmed only to fit the new
file's empty prose-guard baseline — see "Anything reviewers should know").

## The gate

- [ ] `cargo fmt --check` — ran locally (`cargo fmt --all`), clean
- [ ] `cargo clippy --workspace --all-targets -- -D warnings` — not run locally (CI builds and tests; this laptop does not — CLAUDE.md), read carefully instead
- [ ] `cargo test --workspace` — not run locally, same reason; CI is the evidence
- [x] Docs updated where behavior/flags changed — `docs/reference/diagnostics.md`
      regenerated (`make diag-reference`) for the new `Emitted from` path
- [ ] CLA signed
- [x] `Closes #4153` appears both above and as a commit trailer

Ran locally and green: `cargo fmt --all`, `make guards-fast` (all
toolchain-free guards including `prose`, `line-citations`, `file-size`,
`dead-code-allows`, `diagnostic-codes`, `lockfile-sync`).

## Nothing left behind

- [x] There is nothing new: the sink-reachability gap (#5075) and the
      gap-2 bench-arm question were already filed/decided before this PR;
      nothing else was found while making this change.

## Ground-rule check

- [x] N/A — no new cross-boundary type

- The new file, `turn_files/stale_lane.rs`, starts with an empty
  prose-guard baseline (down-only ratchet, per file+pattern). That forced
  rewriting a few sentences carried over from `turn_files.rs` (an issue
  citation in a doc header, "honest board" → "nothing wrong", "the plain
  REPL" → dropped) even though the parent file's own copies of those
  constructions are still grandfathered there. No behavior change, prose
  only.
- `turn_files.rs` was 1473/1500 lines before this PR (unlisted in
  `scripts/file-size-baseline.txt` — not grandfathered). A witness test
  that actually exercises the emission path, plus the `Dx`-parameter doc,
  would have pushed it over 1500 had it stayed inline; that is what forced
  the split into `turn_files/stale_lane.rs` rather than being optional
  polish.
- I looked for, and did not find, any reason to believe #4504's shipped
  diagnostic diverges from the design pointer's spec — level, fields
  (counts/codes only), predicate (unowned in-progress card + file-count
  threshold), registry entry, and docs all match. Happy to be shown
  otherwise.

## Summary by Sourcery

Verify stale-lane diagnostic emission and isolate its implementation in a dedicated module.

Enhancements:
- Add an emission-level witness for the stale task-lane diagnostic, including its level, code, and count fields, while confirming completed tasks remain silent.
- Split stale-lane diagnostic logic into a dedicated module and pass the diagnostic sink explicitly for testability.

Documentation:
- Update the diagnostics reference with the stale-lane diagnostic's new emission path.

Tests:
- Add coverage for both stale-lane diagnostic emission and non-emission cases.


macanderson:
**DoD verification for #4153 — all four items ticked, evidence below.**

The `dod` gate was the only red check on this head (`23d201f1`); `fmt + clippy + test`, `guard self-tests`, `docs guards`, `file size ratchet`, `cargo deny` and the rest are all green. Each DoD box was checked against this branch's diff and the issue's own history before being ticked, not taken from this PR's description:

- **(a) reinforcement signal + measured cost** — not taken, and the box now says so in those words. It is the un-chosen half of an either/or, ticked to record that the choice is *settled*, not that the signal was built.
- **(b) maintainer records the ceiling decision** — verified against the source, not this PR's paraphrase: the design-pointer comment of 2026-09-02 ([#issuecomment-5514335852](https://github.com/macanderson/stella/issues/4153#issuecomment-5514335852)) states the decision verbatim, and that same comment is what instructs the boxes be "ticked with evidence before merge."
- **diagnostic fires at turn close, counts only** — `stale_lane::note_stale_lane` is called from `turn_files::close_turn_boundary`; predicate is `files_changed >= STALE_LANE_FILES` (3) **and** `board.unowned_in_progress()`; fields `files_changed` / `open_tasks` / `threshold` are all `u64`, so no `String`, path, or model output reaches the diagnostic plane.
- **registry entry + witness** — `agent.task.stale_lane` carries its generated entry in `docs/reference/diagnostics.md` (and `docs guards`, which runs `diagnostic-codes`, is green on this head). `note_stale_lane_emits_only_for_the_walked_past_shape` drives a capturing `Dx` through both branches the DoD names: task 1 in progress plus three file writes emits the code at `Level::Debug` with `open_tasks = 2`; the same board after `task_complete` emits nothing.

One deviation worth a reviewer's eye rather than a silent pass: the design pointer named `crates/stella-cli/src/agent/turn_close.rs` as the emit site, and this PR emits from `turn_files/stale_lane.rs`. The PR body argues the case at length (the file count it needs lives in `turn_files.rs`, and moving it would either duplicate that measurement or drop the goal-arc exclusion). That is a documented, reasoned divergence from the pointer, not an unnoticed one — flagging it here because it is the maintainer's call to accept, and it does not affect whether the DoD item itself is met.

Re-running the `dod` check now that the boxes are ticked.

---

Issue #5651, opened 2026-09-02
The transcript's standalone gate row can never be drawn, and its price text is written twice


macanderson:
## Problem

`views/transcript.rs`'s `EventKind::Gate` draws SPEC 6.3's standalone gate row
(`◇ gate <name> · state`, plus `· $0.00 · det` when the decision asked no
model). Nothing builds one. Every other kind in that vocabulary now has a live
wire event behind it, proven per kind by
`crates/stella-tui/src/render/tests/spine.rs`; `Gate` is the one row in that
table declared as a gap.

The engine emits no single-gate event. It emits `AgentEvent::GateBoard`, a
whole board, which folds to `TranscriptEntry::GateBoard` and draws through
`views/gate_board.rs::board_rows`. So the standalone row has no source, and the
two renderers of the same fact have already started to copy each other:

- `views/gate_board.rs`'s `settled_row` pushes the literal `" · $0.00 · det"`.
- `views/transcript.rs`'s `kind_detail` pushes the same literal on its
  `EventKind::Gate` arm.

One of those ships. The other cannot be reached, so when they drift, the copy
nobody can see is the one that drifts.

## What has to be decided

1. **Route the board through `EventKind::Gate`.** One vocabulary, one price
   literal, and the standalone row becomes reachable the day a single-gate
   event exists. It is not free: `board_rows` draws a per-state mark (`✓`
   green, `✗` failed, `○` undecided), a red rail and a red-tinted ground on the
   failing row, and an indent under the board header. `head_row` draws one
   glyph per kind with no ground, so the board's own SPEC 8.1 rendering has to
   survive the move — including
   `a_five_gate_board_spends_red_on_the_failing_row_alone`, and the deck
   goldens under `crates/stella-tui/tests/snapshots/deck/`.
2. **Delete `EventKind::Gate`** and let SPEC 6.3's gate row be described by
   SPEC 8.1's board, since a board is the only form the protocol produces.
   Cheapest, and it gives up the row the day a single-gate event does arrive.

Either way the duplicated price literal ends up in one place.

## Files

- `crates/stella-tui/src/views/transcript.rs` — `EventKind::Gate`, its
  `metal`/`head_glyph`/`verb` arms, and the `kind_detail` arm holding the
  duplicated literal.
- `crates/stella-tui/src/views/gate_board.rs` — `settled_row`, `failed_row`,
  `header_row`, the live renderer.
[...]

PR #5654, opened 2026-09-02
test(stella-tui): draw every SPEC 6.3 head from the wire event that sends it

### LINKED ISSUE #5030: The transcript can draw skill, memory, gate, and model events, but nothing live ever sends one

macanderson:
## What & why

`crates/stella-tui/src/views/transcript.rs` holds the head vocabulary SPEC 6.3
draws — read, edit, write, delete, run, skill, memory log, memory promote,
gate, model, compaction, and the unrecognised-tool row. Its own tests
(`views/transcript/tests.rs`) reach every one of them by building an
`EventKind` by hand. That proves the painter and says nothing about what a
session actually draws: a head no wire event produces passes there and never
once appears on screen. That is the gap #5030 was filed about.

This PR adds `crates/stella-tui/src/render/tests/spine.rs`: one row per head
kind, each starting at an `AgentEvent` a session emits, folded through
`SessionModel::apply` and drawn with the deck's own `entry_lines`. The
assertion is that the drawn head carries the glyph and the word its own
`EventKind` states — the expectation comes from the kind, the text comes from
the fold, so a row where the two agree is a row that reached
`views::transcript::event_rows` on the live path rather than through some
second painter.

`declared` matches every kind, so adding one to the vocabulary stops this file
compiling until somebody says what sends it. That is the discipline
AGENTS.md invariant 10 applies to `AgentEvent` (`event/consumers.rs`), pointed
at the heads this crate draws.

Closes #5030

### Where this differs from the design pointer, and why

The pointer on the issue asks for `TranscriptEntry::{Skill, MemoryLog,
MemoryPromote, Gate, Model}` to be added to `crates/stella-tui/src/model/entry.rs`.
Four of the five are already there and already live end to end, landed by the
per-event producer issues that were meant to depend on this one:

| kind | fold | router arm | live test that already existed |
| --- | --- | --- | --- |
| `Skill` | `model.rs`'s `SkillInjected` arm | `render/entry.rs` | `render/tests/skill.rs` (#5031) |
| `MemoryLog` | `model/memory.rs`'s `fold_memory_write` | `render/entry.rs` | `render/tests.rs::a_logged_memory_renders_its_block_once_and_not_the_v1_row` (#5032) |
| `MemoryPromote` | `model/memory.rs`'s `fold_memory_write` | `render/entry.rs` | `render/tests.rs::a_promotion_renders_one_row_through_the_router` |
| `Model` | `model/turn.rs`'s `model_call_row` | `render/entry.rs` | `views/transcript_source/tests.rs::a_settled_model_call_renders_its_rate_from_the_drivers_own_metering` (#5033) |

So re-adding them would be a no-op, and `kind_for`, `task`, `duration_ms` and
`footer` are already wired (the `head_rows` comment citing this issue is
where the collapsed default and the task tag landed). What was missing is the
half this PR ships: a per-kind enumeration that starts at the wire, covering
the six tool kinds nothing had tested from an entry at all, plus a
compiler-enforced statement of which kind has no sender.

**`Gate` is not added**, and the maintainer's own earlier comment on the issue
is why: "adding `TranscriptEntry` variants with no fold constructor is the
expressible-but-unreachable shape #4180 removed from this very module". The
engine emits `AgentEvent::GateBoard`, a whole board, and never a single gate,
so a `TranscriptEntry::Gate` would have no fold to build it and no producer to
give it one. It is declared as the single gap here and filed as #5651, which
also records the live consequence: `views/gate_board.rs`'s `settled_row` and
`views/transcript.rs`'s `EventKind::Gate` arm push the identical
`" · $0.00 · det"` literal, and only one of them can be reached. Choosing
between routing the board through `EventKind::Gate` and dropping the kind
moves SPEC 8.1's rendering and the deck goldens, which is a decision with its
own diff to read, not a line in a plumbing change.

## The witness

`crates/stella-tui/src/render/tests/spine.rs`, both tests.

**The mechanism, since it could not be run here** (CLAUDE.md: CI builds and
tests, this laptop does not). The file is new, so on `main` neither test
exists and neither can pass. What makes it a witness rather than a restatement
is the second half: `declared`'s match is exhaustive over `EventKind`, so the
moment a kind is added to the vocabulary this file stops compiling until its
sender is named or its gap is cited. Nothing in the tree asked that question
before — `views/transcript/tests.rs`'s
`every_head_glyph_is_in_the_vocabulary` walks the same kinds and is satisfied
by a hand-built fixture, which is exactly how `EventKind::Gate` reached
`main` reachable from tests alone.

Exemplar followed: `crates/stella-protocol/src/event/consumers.rs` — one row
per signal, a posture per row, an issue number on every declared gap, and
totality enforced by the compiler rather than by a test.

## The gate

- [ ] `cargo fmt --check` — run as `cargo fmt --all` locally; CI checks it
- [ ] `cargo test --workspace` — CI
- [x] Docs updated where behavior/flags changed — no behaviour change; the
      new module's own header carries the reasoning
- [x] CLA signed
- [x] `Closes #5030` appears both above and as a commit trailer

`make guards-fast` passes locally, `check-prose` and `check-file-size`
included.

Tests deleted: None.

## Nothing left behind

- [x] Filed: #5651 — the standalone gate row has no sender, and its price text
      is written in two places.

## Ground-rule check

No production code changes: the diff is one new test module plus its `mod`
line. No snapshot golden should move, because nothing that draws changed — if
CI says otherwise, that is a finding worth reading rather than blessing.

## Summary by Sourcery

Add wire-to-render coverage that verifies every live SPEC 6.3 transcript head is reachable and correctly drawn.

Enhancements:
- Add end-to-end transcript rendering coverage for every live SPEC 6.3 head kind, verifying wire events fold and draw with the expected labels and metadata.
- Make unsupported head kinds explicit through exhaustive producer declarations, documenting the unreachable gate head and its tracking issue.
- Verify task identifiers on live tool events render only when supplied by the wire event.

Tests:
- Add spine witness tests covering live tool, skill, memory, model, compaction, and unrecognised-tool transcript heads from AgentEvent through SessionModel rendering.

### DIFF
--- a/crates/stella-tui/src/render/tests.rs
+++ b/crates/stella-tui/src/render/tests.rs
@@ -33,6 +33,8 @@ mod result_row;
 // #5031: SPEC 6.3's skill event, from the wire event to the rendered row.
 mod skill;
 mod slash;
+// Every head SPEC 6.3 draws, from the wire event that draws it.
+mod spine;
 mod steering;
 mod thinking;
 mod tool_output;
--- a/crates/stella-tui/src/render/tests/spine.rs
+++ b/crates/stella-tui/src/render/tests/spine.rs
@@ -20,7 +20,7 @@
 use super::*;
 use crate::tool_class::ToolClass;
 use crate::views::transcript::{EventKind, Extent};
-use stella_protocol::{MemoryClass, ModelCallRole, SkillTrigger, ToolCall};
+use stella_protocol::{MemoryClass, ModelCallRole, SkillTrigger, TaskId, ToolCall};
 
 /// What draws one head kind.
 enum Producer {
@@ -233,7 +233,9 @@ fn live() -> Vec<Row> {
                 tokens_per_sec: Some(84),
             },
             event: step_usage(),
-            marks: &["worker", "84 tok/s"],
+            // The rate, the call's own wall clock, and the footer — the three
+            // head fields beyond the glyph and the word, all from the fold.
+            marks: &["worker", "84 tok/s", "⚡5000ms", "irreducible generation"],
         },
         Row {
             kind: EventKind::Compaction {
@@ -297,6 +299,38 @@ fn every_live_head_kind_is_drawn_from_the_wire() {
             );
         }
     }
+}
+
+/// A live head carries the board tag the wire stamped on the call — SPEC
+/// 6.2's `→ task 3`.
+///
+/// `AgentEvent::ToolStart` carries `task_id`, the fold reads it onto
+/// `TranscriptEntry::ToolStart`, and the metric group draws it. The other
+/// tests for the tag hand `head_rows` a number they made up, so this is the
+/// one that starts where the engine does. An untagged call beside it, because
+/// a tag that always draws is as wrong as one that never does.
+#[test]
+fn a_live_tool_head_carries_the_task_tag_the_wire_stamped() {
+    let tagged = AgentEvent::ToolStart {
+        call: ToolCall {
+            call_id: "c1".into(),
+            name: "edit_file".into(),
+            input: serde_json::json!({ "path": "src/lib.rs" }),
+        },
+        sub_agent_id: None,
+        task_id: Some(TaskId::new("3")),
+    };
+    let drawn = rendered(&tagged);
+    assert!(drawn.contains("→ task 3"), "the tag was dropped: {drawn}");
+
+    let plain = rendered(&tool_start(
+        "edit_file",
+        serde_json::json!({ "path": "src/lib.rs" }),
+    ));
+    assert!(
+        !plain.contains("→ task"),
+        "an untagged call drew a tag the board cannot jump to: {plain}"
+    );
 }
 
 /// The one head nothing sends stays out of the table above and names the
--- a/crates/stella-tui/src/render/tests/spine.rs
+++ b/crates/stella-tui/src/render/tests/spine.rs
@@ -1,0 +1,320 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+// Copyright (c) 2026 Oxagen, Inc. Commercial licensing: licensing@oxagen.sh
+
+//! The witness for `#5030`: every head SPEC 6.3 draws, from a wire event.
+//!
+//! `views/transcript.rs` holds the head words and glyphs, and its own tests
+//! build an `EventKind` by hand. That proves the painter and nothing more. A
+//! head no session sends still passes there, and passed there for months.
+//!
+//! Each row below starts at an `AgentEvent`. The fold turns it into a
+//! `TranscriptEntry`, and the draw turns that into rows. The head has to carry
+//! the glyph and the word its own `EventKind` states. That is what shows the
+//! row went through `views::transcript::event_rows` and not a second painter.
+//!
+//! [`declared`] covers every kind, so a new one stops this file from building
+//! until somebody says what sends it. A kind nothing sends is allowed and
+//! names the issue that will settle it. `EventKind::Gate` is the only one
+//! (`#5651`).
+
+use super::*;
+use crate::tool_class::ToolClass;
+use crate::views::transcript::{EventKind, Extent};
+use stella_protocol::{MemoryClass, ModelCallRole, SkillTrigger, ToolCall};
+
+/// What draws one head kind.
+enum Producer {
+    /// A wire event a session sends.
+    Live,
+    /// Nothing sends one, and the number of the issue that settles it.
+    Gap(u32),
+}
+
+/// What draws `kind`.
+///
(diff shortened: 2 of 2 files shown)


macanderson:
Answering the two ❌ rows in Sourcery's assessment.

**Row 1 — "does not add Gate."** Correct, and on purpose. `EventKind::Gate` is
SPEC 6.3's standalone gate row. The engine emits `AgentEvent::GateBoard`, a
whole board, and never a single gate, so a `TranscriptEntry::Gate` would have
no fold able to build it and no wire event able to feed that fold. That is the
expressible-but-unreachable shape #4180 removed from this very module, and the
maintainer's own earlier comment on #5030 rules it out in the same words.

It is not left silent: `declared` in `render/tests/spine.rs` declares it as the
single gap, the compiler refuses the file if a kind is added without an answer,
and #5651 carries the decision (route `gate_board::board_rows` through
`EventKind::Gate`, or drop the kind) together with the live consequence — the
`" · $0.00 · det"` literal is built in both `views/gate_board.rs` and
`views/transcript.rs`, and only one of the two can be reached. Either choice
moves SPEC 8.1's rendering and the deck goldens, which is a diff to read rather
than a line inside a test-only change.

**Row 2 — "does not exercise task, duration or footer."** Right about the
table, so it is fixed in 5427cb65d:

- the model row now pins `⚡5000ms` and `irreducible generation` beside its
  rate, so the wall clock and the footer are both read off a live head;
- `a_live_tool_head_carries_the_task_tag_the_wire_stamped` stamps `task_id` on
  an `AgentEvent::ToolStart` and reads `→ task 3` off the drawn head, with an
  untagged call beside it. The existing tests for the tag call `head_rows` with
  the number already in hand; this one starts where the engine does.

One correction to the row: the plumbing itself is not missing from `main`. The
`task`, `duration_ms` and `footer` fields were wired on the live head by the
earlier #5030 work — `views/transcript_source.rs`'s `head_rows` sets all three,
and the issue's own DoD box for `→ task 3` and `⚡ms` was ticked before this PR
opened. What this PR adds is coverage that starts at the wire rather than at
the adapter.

Issue #5630, opened 2026-09-02
Epic: the deck's list pickers stay usable as their candidate lists grow


macanderson:
## Why

`ListPicker` overlays (`/model`, `/agent`, the graph file picker) render a fixed dozen `VISIBLE_ROWS` at a time with no way to narrow the list. That is fine at a handful of candidates and breaks the moment a real workspace configures a gateway: a machine with OpenRouter plus Anthropic reports 479 model candidates, and reaching row 300 meant holding an arrow key. The SETTINGS tab's own model picker and the graph file picker already solved this with a type-to-filter affordance; the `/model` and `/agent` overlays were the odd ones out.

This epic groups the follow-on work needed to keep those pickers usable as the candidate lists they draw from keep growing — filtering was the blocking case; grouping and further scale work follow.

## Scope

- [x] Filter-as-you-type on the `/model` and `/agent` picker overlays, matching the SETTINGS/graph-picker vocabulary and key bindings — #5625 (PR #5624).
- [x] Group the picker's rows under a provider heading once filtering exists, so scanning the unfiltered list is also usable — #5623 (deliberately deferred out of #5624/#5625 because heading rows must not enter the index `PickerAction::Choose` resolves against; needs its own witness).

## Definition of done

- [x] Every sub-issue above is closed with a witness test proving its fail→pass flip.
- [x] `make guards-fast` (file-size, prose, dead-code-allows) stays green on every change in scope.
- [x] Docs/comments for `views/picker.rs` and `deck_ui/pickers.rs` describe the final interaction model once all sub-issues land.

macanderson:
Triage pass on this issue:

- **Not a duplicate.** Searched for another open epic covering picker filtering/scaling; none exists — this is the only epic for that outcome.
- **Not an orphan.** This issue is itself the epic (top of its own hierarchy), so the `/backlog-triage` "needs a parent" facet (per #2778's contract) doesn't apply reflexively to it. Its two children are already correctly sub-issued: #5625 (filtering — implemented, PR #5624, all DoD boxes checked) and #5623 (provider-heading grouping — open, deliberately scoped out of #5625 per that issue's own body).
[...]

Issue #5625, opened 2026-09-01
/model picker is unusable at real catalog scale — several hundred rows, no filter


macanderson:
## Context

`/model` opens a modal list of every model from every provider whose credential resolves. That scoping works and is deliberate
(`command_deck/panel_snapshots.rs`'s `engine_config_inbound`: "a model you have no key for is not an option"). Reaching into the result does not.

A gateway contributes hundreds of rows on its own. On a machine with OpenRouter and Anthropic configured, `stella models list` reports 464 + 15 = **479 candidates**, drawn twelve at a time (`VISIBLE_ROWS` in `crates/stella-tui/src/views/picker.rs`) with no way to narrow them. Getting to `openrouter/qwen/qwen3-max` means holding an arrow key. A list you cannot reach into is not one you can select from.

The SETTINGS tab's model picker over the same candidate list already filters (`views/engine_panel/keys.rs`'s `handle_picker_key`, `picker_matches`), and `picker_matches`'s own doc comment claims the two pickers "feel identical" — which is not true today. The graph file picker filters too (`deck_ui::tests::graph::typing_filters_and_re_anchors_the_selection`). The session picker is the odd one out.

## Definition of done

- [x] Implementation lands in `crates/stella-tui/src/views/picker.rs` and `crates/stella-tui/src/deck_ui/pickers.rs` — the crates the picker widget and its routing already live in; no new module or crate needed.
- [x] Witness test(s) fail on `main`, pass with the change. Named:
  - `deck_ui::tests::queue::typing_in_the_model_picker_narrows_it_and_enter_sends_the_matching_row`
  - `views::picker::tests::letters_type_into_the_filter_rather_than_moving_or_quitting`
  - `views::picker::tests::editing_the_filter_re_anchors_the_highlight`
  - `views::picker::tests::the_card_draws_the_filter_and_only_the_rows_it_admits`
  - Verified the artisanal way: reverting `views/picker.rs` and `deck_ui/pickers.rs` to `main` with the tests retained fails to build (`model_matches` / `ListPicker::query` do not exist on `main`).
- [x] Golden frames re-blessed and read, not blindly regenerated: `overlay_model_picker`, `overlay_agent_picker` (`BLESS=1 cargo test -p stella-tui --test deck_render_snapshots`).
- [x] Scoped tests green: `cargo test -p stella-tui`.
- [x] Full CI green on the PR head.
[...]

PR #5655, opened 2026-09-02
feat(stella-tui): group the /model picker under provider headings

### LINKED ISSUE #5623: /model picker: group the rows under a provider heading
## Problem

The `/model` picker draws one flat list of `provider/slug` strings. With
several providers credentialed it is not obvious where one provider's models
end and the next begin, and the reader has to parse the prefix of every row to
tell. The list is filtered to credentialed providers already
(`command_deck/panel_snapshots.rs`'s `engine_config_inbound`), so the grouping
information is present and unused.

Filtering landed in `feat(stella-tui): filter the /model picker as you type`
(#5625 / PR #5624), which makes the list reachable. This issue is the second
half: making it *readable*. Deliberately deferred out of #5625 because heading
rows must not enter the index `PickerAction::Choose` resolves against — see
the constraint below.

## What to do

Emit a heading row when the provider changes, in
`crates/stella-tui/src/views/picker.rs`'s `render_model`.

- Split each spec on the **first** `/` only — a gateway spec is
  `openrouter/openai/gpt-5.5`, and the provider is the first segment.
  `views/picker.rs`'s `typeahead_candidates` and
  `command_deck/model_cmd.rs` both already depend on first-slash-only.
- `EngineConfigState.providers` (`crates/stella-tui/src/envelope/engine_config.rs`)
  carries the credentialed provider ids, for a stable heading order.
  `ProviderConfig.display_name` (`crates/stella-cli/src/config/providers.rs`)
  is the pretty label, but the deck only receives ids today — either map ids
  deck-side or extend `EngineConfigState` to carry `(id, display_name)`.

## The constraint that makes this non-trivial

`ListPicker::key` counts rows 1:1 with the **matching candidates**, and
`PickerAction::Choose(i)` indexes into that same list
(`crates/stella-tui/src/deck_ui/pickers.rs`). Heading rows must therefore not
enter that count, so `render_model`'s `window()` and its `selected_row`
translation have to map candidate index → screen row with the headings
interleaved. The established pattern for exactly this is
`SlashMenu.sections: Vec<(usize, String)>` in
`crates/stella-tui/src/composer.rs` — a heading list keyed by candidate index,
which keeps the indices honest.

## Verify

- A witness test in `crates/stella-tui/src/views/picker.rs` asserting that a
  two-provider candidate list draws both headings, that `⏎` still sends the
  highlighted *model* (never a heading), and that arrowing skips headings.
- Re-bless the goldens and read the diff:
  `BLESS=1 cargo test -p stella-tui --test deck_render_snapshots`
  (`overlay_model_picker`). `deck_render_snapshots_are_reproducible` requires
  deterministically ordered rows — no `HashMap` iteration.

## Done when

`/model` with two or more credentialed providers draws each provider's models
under a heading for that provider, the highlight can never land on a heading,
and the filter still narrows across every group.

## Definition of done

- [x] Implementation lands in `crates/stella-tui/src/views/picker.rs` (and
      `crates/stella-tui/src/deck_ui/pickers.rs` only if the choose-index
      mapping needs it) — the crates the picker widget and its routing
      already live in; no new module or crate needed. PR #5655 touches only
      `views/picker.rs`; `deck_ui/pickers.rs` did not need a change, since
      `PickerAction::Choose` already resolved against the flat candidate
      index the render side now maps painted rows onto.
- [x] The witness test above (see Verify) fails on `main`, passes with the
      change. `grouped_rows_paint_a_heading_before_each_provider_run` calls
      a function absent from `main`, so it fails to build there; it and
      `a_filter_narrowed_to_one_provider_draws_one_heading` and
      `the_model_card_draws_a_heading_between_provider_runs` all pass on the
      PR head (CI run
      https://github.com/macanderson/stella/actions/runs/33697040941).
- [x] Golden frames re-blessed and the diff read, not blindly regenerated:
      `overlay_model_picker`
      (`BLESS=1 cargo test -p stella-tui --test deck_render_snapshots`).
      `deck_render_snapshots_are_reproducible` still holds — no `HashMap`
      iteration in the new heading order. `BLESS=1` could not run on the
      implementing session's machine (no local cargo compiles), so the
      golden was re-derived by hand against the unchanged rendering
      geometry, with every byte explained in PR #5655's description; CI's
      plain `cargo test` (the same byte comparison `BLESS=1` would rewrite
      against) passed against that hand-derived file, which is the
      strongest evidence this repo's own gate can produce that the diff is
      correct.
- [x] Scoped tests green: `cargo test -p stella-tui`. Green in
      https://github.com/macanderson/stella/actions/runs/33697040941
      ("fmt + clippy + test", `cargo test` step).
- [x] Full CI green on the PR head. All required checks pass on PR #5655's
      head (`6bd5b5a1d`): fmt/clippy/test, cargo check (MSRV), docs guards,
      guard self-tests, dependency-review, cargo-deny, and the rest.
- [x] No god file grows past its ratchet — neither `views/picker.rs` nor
      `deck_ui/pickers.rs` is on `stella-tui`'s god-file list (only
      `src/deck_ui.rs` is, per `AGENTS.md`'s table); `make file-size` /
      `make guards-fast` stay green. Confirmed locally and in CI's "file
      size ratchet" and "guard self-tests" checks.
- [x] `make prose` stays green — no content-free constructions added to the
      touched module headers or the picker's hint row. Confirmed locally
      (`python3 scripts/check-prose.py`) and in CI.
- [x] Code comments/docs updated: module headers of `views/picker.rs` (and
      `deck_ui/pickers.rs` if touched) describe the interleaved
      heading/candidate indexing. `views/picker.rs`'s module header now
      names `grouped_rows` and the flat-index/painted-row split;
      `deck_ui/pickers.rs` was not touched.
- [x] Interaction design matches the established
      `SlashMenu.sections: Vec<(usize, String)>` heading-by-index pattern
      (`crates/stella-tui/src/composer.rs`) rather than inventing a new one.
      PR #5655 follows that pattern's *render side* —
      `crate::render::display_rows`'s `PopupRow` (`Heading`/`Command`
      variants, windowed by painted row, headings never selectable) — the
      shape this issue's own maintainer design-pointer comment names as the
      one to reuse for exactly this kind of interleaved list.

macanderson:
## What / why

The `/model` picker drew one flat list of `provider/slug` rows. With several
providers credentialed there was no seam between them, so a reader had to
parse every row's prefix to tell where one provider's models end and the
next begin. #5625/#5624 made the list reachable (filter-as-you-type); this
makes it readable, per the design note on #5623.

`render_model` now paints a non-selectable heading row wherever the
provider changes in the already-filtered candidate list (`grouped_rows`,
splitting each spec on the first `/` — the same split `typeahead_candidates`
and `command_deck/model_cmd.rs` already depend on).

Headings are render-only chrome. `ListPicker::key`, `PickerAction::Choose`
and `deck_ui/pickers.rs` are all unchanged: they still work over the flat
candidate index. Only the render loop maps that flat index onto the
interleaved screen rows a heading sits inside, the same shape the command
palette already uses for its own section headings (`SlashMenu::sections`,
drawn by `render::display_rows`'s `PopupRow` — a heading row that is chrome,
never a selectable index). `window()` needed no logic change: it already
just slides a `(sel, count)` pair over whatever index space the caller
hands it, so passing it the painted-row count instead of the candidate
count is what makes `VISIBLE_ROWS` cap painted rows, per the design note.

Exemplar followed: `crate::render::display_rows` / `PopupRow`
(`crates/stella-tui/src/render.rs`) — the established heading-by-index
pattern for exactly this shape in this codebase, as the issue names.

## Witness

`grouped_rows_paint_a_heading_before_each_provider_run`
(`crates/stella-tui/src/views/picker.rs`) is the mechanism witness: it
calls a function (`grouped_rows`) that does not exist on `main`, so it
fails to compile there. It proves the exact case from the design note — 5
candidates split 3-and-2 across 2 providers paint 7 rows, 2 of them
headings, and painted row 4 (1-indexed) is candidate 2, the same flat
index `PickerAction::Choose` resolves against.

`a_filter_narrowed_to_one_provider_draws_one_heading` covers the
single-provider case named in the DoD's "Verify" section.

`the_model_card_draws_a_heading_between_provider_runs` is the render-level
witness: it renders the actual card, asserts both provider headings are
drawn on screen, arrows past the heading between two providers' runs and
confirms the highlight lands on the candidate the reader arrowed to (not
shifted by the heading row), and confirms the heading's own screen line
never carries the cursor marker — so `⏎` can never choose a heading.

All three fail to compile/pass on `main` (the fields/functions/behavior
they exercise do not exist there) and pass with this change.

## Golden frame

`overlay_model_picker` (`tests/snapshots/deck/overlay_model_picker.txt`)
changes: its fixture has 3 candidates across 3 distinct providers, so 3
headings are now painted and the card grows from 6 to 9 rows, shifting its
top edge up within the fixed-size frame.

I could not run `BLESS=1 cargo test -p stella-tui --test deck_render_snapshots`
in this sandbox (no cargo compiles allowed here — CI is the only compiler
available to this session). I re-derived the golden by hand instead, from
the unchanged geometry (`window()`, `card_area()`'s `y = frame.y +
frame.height - h` anchor, both unchanged by this diff) and the unchanged
byte content of every row that isn't new: every border, filter and
candidate row in the new golden is copied verbatim (same columns, same
characters) from its row in the pre-change golden, just at a shifted
buffer row; the three new heading rows are the only genuinely new bytes,
each `" {provider}"` left-aligned and space-padded to the card's interior
width, matching the shape `filter_row`/candidate rows already use. The one
row whose background extends past the card's right edge in this fixture
(the `⚡ 2100ms` badge, now behind the first heading row) keeps that
trailing content, since the card's width and right edge are unchanged.

If CI's `deck_render_snapshots_are_reproducible` or the golden comparison
itself disagrees with this reconstruction, that is a bug in my derivation,
not in the rendering change — I will read the CI diff and correct the
golden rather than the code.

`deck_render_snapshots_are_reproducible` still holds: `grouped_rows` walks
`candidates` (already a `Vec`, in filter-preserved order) with a plain
`for` loop — no `HashMap` iteration anywhere in the new code.

## Scope note

The maintainer's design pointer and the issue body both scope this to the
`/model` picker (`render_model`) only — the issue title, "What to do" and
"Definition of done" all name `render_model` and `views/picker.rs`
specifically, and the epic (#5630) lists this sub-issue the same way. The
`/agent` picker (`render_agent`) is unchanged; installed agents don't carry
a `provider/slug` shape to group by, and nothing in the issue or the design
note asks for it.

`deck_ui/pickers.rs` needed no change — the choose-index mapping the issue
flagged as the hard part turned out not to need touching, since the flat
candidate index was already what `ListPicker`/`PickerAction::Choose`
operate over; only the render side needed the heading-aware windowing.

## Checks

- `cargo fmt --all`: clean, no reformatting.
- `make guards-fast`: green (file-size, prose/prose-density, dead-code-allows,
  line-citations, left-behind, and the rest all pass).
- Full CI (build/clippy/test/rustdoc) has not run yet in this session — I
  cannot compile locally — so I'll be watching this PR's checks.

Tests deleted: none.

Residue issues: none filed — no new gap was found in scope; the
`/agent`-picker grouping question above is answered by the issue's own
scope, not left open.

Closes #5623

## Summary by Sourcery

Group `/model` picker entries under non-selectable provider headings without changing candidate selection behavior.

New Features:
- Add provider headings to the `/model` picker, grouping adjacent models by provider while keeping headings non-selectable.

Enhancements:
- Preserve flat candidate selection and keyboard navigation while windowing the rendered list to account for interleaved headings.

Documentation:
- Update the `/model` picker example and related rendering documentation to show provider grouping.

Tests:
- Add coverage for provider grouping, filtered single-provider results, heading rendering, selection navigation, and cursor exclusion.

### DIFF
--- a/crates/stella-tui/src/views/picker.rs
+++ b/crates/stella-tui/src/views/picker.rs
@@ -44,7 +44,7 @@
 //!
 //! In a file of its own under the god-file rule — `deck_ui.rs` pays only
 //! the two state fields; key routing is `deck_ui/pickers.rs`, which
-//! `render_model`'s provider headings ([`grouped_rows`]) never reach.
+//! `render_model`'s provider headings (`grouped_rows`) never reach.
 //!
 //! [`EngineConfigState`]: crate::envelope::EngineConfigState
 
@@ -898,10 +898,13 @@ mod tests {
         let mut buf = Buffer::empty(area);
         render_model(&model, &ui, area, &mut buf);
         let frame = text(&buf);
+        // A heading row's own interior text — between the card's left and
+        // right border, trimmed — rather than the whole screen line, which
+        // still carries the border and the surrounding centering padding.
         let headings: Vec<&str> = frame
             .lines()
-            .map(str::trim)
-            .filter(|line| *line == "zai" || *line == "anthropic")
+            .filter_map(|line| line.split('│').nth(1).map(str::trim))
+            .filter(|inner| *inner == "zai" || *inner == "anthropic")
             .collect();
         assert_eq!(headings, vec!["zai", "anthropic"], "{frame}");
 
@@ -920,7 +923,7 @@ mod tests {
         );
         let heading_line = frame
             .lines()
-            .find(|line| line.trim() == "anthropic")
+            .find(|line| line.split('│').nth(1).map(str::trim) == Some("anthropic"))
             .expect("the heading is still drawn");
         assert!(
             !heading_line.contains('▸'),
--- a/crates/stella-tui/src/views/picker.rs
+++ b/crates/stella-tui/src/views/picker.rs
@@ -5,10 +5,10 @@
 //! `/agent` (run as an installed agent this session).
 //!
 //! ```text
-//! ╭ model · 1/3 · this session only ─────────────────────────────╮
-//! │  filter gpt▏                                                 │
+//! ╭ model · 1/1 · this session only ─────────────────────────────╮
+//! │  filter gpt-5.5▏                                             │
+//! │ openrouter                                                   │
 //! │▸ openrouter/openai/gpt-5.5                                   │
-//! │  openrouter/openai/gpt-5.5-mini                              │
 //! ╰────────────────── type to filter · ↑↓ move · ⏎ use · esc ────╯
 //! ```
 //!
@@ -43,7 +43,8 @@
 //! where it ends.
 //!
 //! In a file of its own under the god-file rule — `deck_ui.rs` pays only
-//! the two state fields; key routing is `deck_ui/pickers.rs`.
+//! the two state fields; key routing is `deck_ui/pickers.rs`, which
+//! `render_model`'s provider headings ([`grouped_rows`]) never reach.
 //!
 //! [`EngineConfigState`]: crate::envelope::EngineConfigState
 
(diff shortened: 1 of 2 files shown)


macanderson:
Answering Sourcery's ❌ row ("use the credentialed providers' deterministic ordering and their configured display names for the headings"):

That objective comes from #5623's original "What to do" section, which predates the maintainer's design-pointer comment on the same issue. The design pointer is what this PR follows (per this repo's protocol for implementing an issue): "heading rows are render-only... insert a non-selectable heading row when the provider of candidate i differs from candidate i-1 (in the filtered list, so headings follow the filter)" — raw seam detection over the already-filtered, already-ordered candidate list, with no mention of `EngineConfigState.providers` ordering or `ProviderConfig::display_name`. The witness the design pointer specifies (5 candidates across 2 providers → 7 painted rows, 2 headings, painted row 4 → candidate 2) is exactly what this PR proves and needs neither piece.

Pulling in display names or a provider-order concept would also widen `EngineConfigState`'s wire shape across crates, which is more than a render-only change and out of scope for this PR's single-crate diff.

Filed #5656 to track both pieces as follow-up work, with the file paths and the two independent verify steps a fresh session would need.

PR #5658, opened 2026-09-02
fix(stella-store): keep the store error a failed audit write carries

### LINKED ISSUE #5626: Audit record silently incomplete: "store write failed" recurs once per drive run with no underlying error
## What happens

Every `drive --backlog` run against macanderson/rainforest prints, roughly once per run:

```
⚠ store write failed — the audit record (agent uses / MCP usage / outcome) for this execution is incomplete
```

Three runs, three occurrences (drive-3 ×1, drive-4 ×2). The turn keeps going and the work still lands, so the only harm is a hole in the audit trail — which is exactly the record you cannot rebuild later.

## The gap in the report

The warning names what broke but not why. There is no `level: error` line next to it, no SQLite result code (`SQLITE_BUSY` / `SQLITE_FULL` / `SQLITE_READONLY`), no file path, no retry count — nothing an operator can act on. The surrounding log lines are unrelated `agent.tool.result` and provider-retry warnings.

When it fired: one `stella` process per run (chained, never running at the same time as another), stores at `.stella/private/{context,fleet,codegraph}.db`, plenty of disk, no odd permissions — so lock contention between parallel sessions and a full disk are not obviously the cause.

## Root cause (confirmed in code)

`record_execution_end` in `crates/stella-cli/src/agent/persistence.rs` throws away the real error on every write it makes:

```rust
let uses_ok = uses.is_empty() || store.record_agent_uses(execution_id, &uses).is_ok();
let mcp_usage_ok = store.record_mcp_usage(execution_id, &mcp_usage).is_ok();
...
let finish_ok = store.finish_execution_accounted(...).is_ok();
```

Each of `record_agent_uses`, `record_mcp_usage`, and `finish_execution_accounted` returns a typed `Result` (`stella_store::StoreError`, which already carries the SQLite result code — see `crates/stella-store/src/error.rs` and the busy-detection pattern in `crates/stella-store/src/migrations/pragmas.rs`). `.is_ok()` drops that value on the floor before `crates/stella-cli/src/agent/turn_close.rs:77` calls `warn_store_write_failed(...)` with only a fixed string. The warning is real; the diagnosis it needs already exists one call away and is never read.

## Ask

- Keep the `Err` from each of the three writes above and log it beside the warning: store path, SQLite result code (`rusqlite::Error::SqliteFailure` → `ErrorCode`), and whether a retry was attempted.
- Retry a busy/locked write with backoff before giving up — reuse the busy-retry shape already in `crates/stella-store/src/migrations/pragmas.rs` rather than inventing a second one. An audit record is worth more than one attempt.
- If a write is still dropped after retries, count it so `stella self-driving stats` (`crates/stella-cli/src/self_driving_cmd/stats.rs`) can report "N audit records incomplete this run" instead of leaving it to log archaeology.

## Why this one matters

Governed autonomy is the product's core claim, and the audit record is the evidence for it. A silent, recurring hole in that record — with no diagnosable cause — undercuts the exact guarantee the loop exists to provide.

## Definition of done

- [x] `record_execution_end` (`crates/stella-cli/src/agent/persistence.rs`) captures the `Err` from `record_agent_uses`, `record_mcp_usage`, and `finish_execution_accounted` instead of collapsing each to `.is_ok()`.
- [x] The logged warning (`warn_store_write_failed`, same file) includes the store path, the SQLite result code, and whether a retry ran — no more bare "store write failed" with nothing to act on.
- [x] A busy/locked write (`SQLITE_BUSY` / `SQLITE_LOCKED`) retries with backoff before it is counted as dropped, sharing the retry shape already in `crates/stella-store/src/migrations/pragmas.rs` rather than a second copy of it.
- [x] A write still dropped after retries increments a counter that `stella self-driving stats` (`crates/stella-cli/src/self_driving_cmd/stats.rs`) surfaces as "N audit records incomplete this run."
- [x] **Witness test**: a test that forces a busy/locked write on the same tables (hold a competing write transaction open, the way existing `stella-store` concurrency tests do) and shows the old code silently swallowing it with the generic message, while the new code either recovers via retry or reports the real SQLite code and increments the drop counter. Fails on `main` today, passes after the fix.
- [x] No new `Result<_, String>` — the captured error stays a typed `StoreError` end to end (AGENTS.md invariant #5); `make typed-errors` stays green.
- [x] The change lands inside `crates/stella-store` and `crates/stella-cli/src/agent/{persistence.rs,turn_close.rs}` — no new crate, and it does not push either file (neither is on the god-file list today) past the 1500-line ceiling; if a file is getting crowded, split before merging rather than growing a god file (`make file-size`).
- [x] Docs: update `website/content/docs/commands/self-driving.mdx` to describe the new "N audit records incomplete" line in `stella self-driving stats` output, and fix any prose this change makes stale. `make prose` and `make line-citations` stay green (cite code by symbol, not by line number).
- [x] No UI/design-system surface is touched by this change (CLI + store only) — nothing to check there.
- [x] Narrow verification only, per SCR-001: `cargo test -p stella-store` and `cargo test -p stella-cli` (filtered to the touched modules), not a full workspace run. CI's `make gate` is the authoritative full check.

macanderson:
## What and why

`record_execution_end` closed an execution with three writes — agent uses, MCP usage, the outcome row — and collapsed each one to `.is_ok()` before anyone looked at it. The operator got one fixed sentence: *store write failed — the audit record for this execution is incomplete*. A locked database, a read-only file and a full disk all printed that same line, the typed `StoreError` one call away was never read, and nothing anywhere counted how often it happened. That is the hole in the audit trail #5626 describes, seen once per `drive --backlog` run.

Three changes:

1. **Keep the error.** `ExecutionEndOutcome` now carries the writes that were given up on, each with its `StoreError` and the number of tries spent. `warn_dropped_audit_writes` (`crates/stella-cli/src/agent/audit_writes.rs`) prints the database file, the SQLite result code (`DatabaseBusy`, `ReadOnly`, `Full`, …) and whether a retry ran. `Store::db_path` is the file the store actually opened, rather than a path the CLI rebuilds.
2. **Retry a held lock.** `stella_store::retry_busy` (`crates/stella-store/src/busy.rs`) asks again up to five times, waiting 20 ms and doubling, and reports the try count. Its predicate `is_busy` is the one `migrations::pragmas` already used to absorb a concurrent first open — that call site now reads it from here instead of comparing a code inline, so there is one answer to "was that the lock, or the data?" (it also widens the opener from `SQLITE_BUSY` alone to `SQLITE_LOCKED` as well, which is the same bounded wait).
3. **Count what is still lost.** A dropped write bumps a process-wide `AtomicU32`. A turn prints its count in the `--output-format json` summary (`audit_records_incomplete`); the self-driving loop reads that back out of the child turn's summary and folds it, with its own drops, into `SessionStats::audit_records_incomplete`, which `stella self-driving stats` renders as `N audit records incomplete this run`. Without the child hop the number would always be zero for a drive run, because `work` spawns `stella run` rather than dispatching in-process.

Typed end to end: no new `Result<_, String>`; the two new public accessors on `StoreError` (`sqlite_code`, `is_busy`) are what let a caller branch instead of reading prose.

## Witnesses and how each one fails on `main`

- `crates/stella-store/src/busy.rs::a_write_refused_for_a_held_lock_lands_after_the_lock_clears` — two connections on a real `store.db`, both with `busy_timeout` at zero so the test never waits on a clock. One holds `BEGIN IMMEDIATE`; the write is refused, the holder commits between the first refusal and the second ask, and the row lands on try 2. Fails on `main` by construction: `retry_busy` does not exist there, and the first refusal was the whole answer.
- `…::a_lock_that_never_clears_gives_up_after_a_bounded_number_of_attempts` — the same contention with no release: five tries, then the busy code is reported rather than swallowed.
- `…::a_failure_that_is_not_a_lock_is_answered_on_the_first_attempt` — a missing table is not retried.
- `crates/stella-cli/src/agent/audit_writes.rs::a_refused_audit_write_is_counted_and_carries_its_code` — the end-to-end one. The `mcp_usage` table refuses a second row at a `seq` it already holds, so the closeout's write is genuinely turned away: the drop counter moves by exactly one and the report names `MCP usage` and `ConstraintViolation`. On `main` there is no counter and no code — `.is_ok()` had already thrown the error away.
- `…::the_report_names_the_code_the_file_and_whether_a_retry_ran` and `…::a_write_that_was_not_retried_says_so` — the rendering, both "which file" branches.
- `crates/stella-cli/src/self_driving_cmd/stats.rs::dropped_audit_writes_reach_the_session_counters` — the count reaches `SessionStats`.
- `crates/stella-cli/src/self_driving_cmd/work.rs::a_child_turns_dropped_audit_writes_are_read_from_its_summary` — the child hop, including an older binary's summary with no such key reading as 0.

Contention through `Store::record_agent_uses` itself would need the competing transaction held for more than the store's own five-second `busy_timeout` before any busy error exists to retry; the witness reproduces the same refusal on the same table with a zero timeout instead, which is the same code path with a deterministic clock rather than a six-second sleep.

## Two moves the god-file rule forced

`crates/stella-store/src/lib.rs` is grandfathered and closed to growth, so declaring a new module there had to be paid for: the `mcp_usage` table's rows, writer and reader move into `crates/stella-store/src/mcp_usage.rs`, beside the `agent_uses` module they mirror (that module is the exemplar followed — same shape, same doc structure). Net effect on `lib.rs` is −99 lines. `RawRunSummary` likewise moved from `agent.rs` (7 lines under the 1500 ceiling, no baseline entry) into `agent/summary.rs`, the module that builds it. Both are pure moves plus re-exports; every public path is unchanged.

## Docs

`website/content/docs/commands/self-driving.mdx` gains an *Audit records* subsection under `stats`: what the line means, when it appears, and that a busy database is retried before a write counts as lost.

Tests deleted: None.

Closes #5626

## Summary by Sourcery

Preserve, retry, report, and count audit writes that fail during execution closeout.

New Features:
- Report incomplete audit records with actionable SQLite error codes, database paths, and retry information.
- Track dropped audit writes across individual turns and self-driving sessions, including child turn summaries and JSON statistics.
- Retry audit writes that fail because SQLite is busy or locked before treating them as lost.

Bug Fixes:
- Preserve previously discarded store errors during execution closeout instead of reducing audit write failures to boolean status.
- Recognize both busy and locked SQLite failures consistently during store initialization and audit writes.

Enhancements:
- Expose typed SQLite error-code and lock-status accessors on StoreError and the actual opened database path on Store.
- Move MCP usage persistence into its dedicated store module and relocate the raw run summary alongside its construction.

Documentation:
- Document incomplete audit record reporting and bounded retries in the self-driving command guide.

Tests:
- Add coverage for busy-lock retries, bounded retry exhaustion, non-lock failures, detailed audit failure reporting, and session-level propagation of dropped audit writes.

### DIFF
--- a/crates/stella-autonomy/src/stats.rs
+++ b/crates/stella-autonomy/src/stats.rs
@@ -133,6 +133,16 @@ pub struct SessionStats {
     pub turns_run: u32,
     /// Turns that stopped because they hit their spend ceiling.
     pub turns_over_budget: u32,
+
+    // -- what the record lost -----------------------------------------------
+    /// Audit writes the session gave up on. This many turns have a hole in
+    /// their record.
+    ///
+    /// It is counted because the record is the proof for the rest of this
+    /// list. A store that turns a write away does not stop the work. The loop
+    /// runs on and the hole is quiet, so the count is what makes it a number
+    /// a person can see.
+    pub audit_records_incomplete: u32,
 }
 
 impl SessionStats {
--- a/crates/stella-autonomy/src/stats.rs
+++ b/crates/stella-autonomy/src/stats.rs
@@ -317,6 +317,10 @@ mod tests {
             "turns_over_budget",
             "drive.rs — a turn hit its spend ceiling",
         ),
+        (
+            "audit_records_incomplete",
+            "self_driving_cmd/stats.rs — a store write the close-out lost",
+        ),
     ];
 
     /// **The witness.** Every field the report renders names the code that
--- a/crates/stella-cli/src/agent.rs
+++ b/crates/stella-cli/src/agent.rs
@@ -43,6 +43,7 @@ use crate::runtime::{SystemClock, TokioSleeper};
 use crate::{OutputFormat, config::Config};
 use stella_context::EpisodeOutcome;
 
+pub(crate) mod audit_writes;
 mod budget;
 mod engine;
 mod goal;
@@ -62,6 +63,14 @@ mod summary;
 pub(crate) mod tool_stack;
 pub(crate) mod tools;
 mod turn_close;
+// The count of audit writes that were lost, and the warnings that name them.
+// A turn that could not write its own record says so through these. The turn
+// close-out sets the count. The run summary prints it. The self-driving loop
+// adds it up over a run.
+pub(crate) use audit_writes::{
+    dropped_audit_writes, note_child_dropped_audit_writes, take_dropped_audit_writes,
+    warn_dropped_audit_writes, warn_store_write_failed,
+};
 pub(crate) use budget::{build_budget_guard, remaining_budget, settle_reflection_budget};
 pub(crate) use graph_view::{graph_query_snapshot, graph_snapshot, graph_snapshot_focus};
 
@@ -84,13 +93,13 @@ pub(crate) use output::{
 pub(crate) use persistence::{
     PersistOutcome, ReasoningRun, begin_execution, close_event_stream, flush_reasoning_tail,
     persist_event, persist_owed, record_execution_end, seed_calibration, spawn_renderer,
-    warn_store_write_failed,
 };
 pub(crate) use presence::SessionPresence;
 pub(crate) use prompt::*;
 use reflect::reflect_on_interactive_turn;
 pub(crate) use reflect::surface_reflection;
 pub(crate) use skill_usage::stamp_and_record_skill_usage;
+pub(crate) use summary::RawRunSummary;
 pub(crate) use tools::*;
 
 /// Whether this process may touch durable workspace state, as the
@@ -155,34 +164,6 @@ pub async fn run_one_shot(
         invoked_skill,
     )
     .await
-}
-
-/// The `--output-format json` summary of a raw (`--no-pipeline`) step-loop run.
-/// A struct rather than `serde_json::json!` so the version leads the object
-/// (`json!` builds a sorted map and would bury it mid-envelope); key order is
-/// not contractual either way — this is for whoever reads the output by eye.
-///
-/// [`schema_version`](Self::schema_version) is governed by the bump rule on
-/// [`crate::SUMMARY_SCHEMA_VERSION`].
-#[derive(serde::Serialize)]
-pub(crate) struct RawRunSummary {
-    pub(crate) schema_version: u32,
-    pub(crate) status: &'static str,
-    pub(crate) text: Option<String>,
-    pub(crate) cost_usd: Option<f64>,
-    pub(crate) reason: Option<String>,
-    pub(crate) model: String,
-    pub(crate) events: Vec<AgentEvent>,
-    /// The file-touch telemetry envelope. Filled from the turn's own measured
-    /// `FileChange` events (`agent::summary::files_touched`) — the key stays
-    /// even on a quiet turn because the envelope's key set is the versioned
-    /// contract.
-    pub(crate) files_touched: serde_json::Value,
-    /// What this checkout's trust gate held back, or `null` (#4465). Folded
-    /// from the turn's own `SteeringWithheld` event, so this and the
-    /// `stream-json` carrier of the same fact are one event read twice.
-    /// Present on every raw summary for the same reason `files_touched` is.
-    pub(crate) withheld: serde_json::Value,
 }
 
 /// The REPL's productized command names — reserved: a custom definition can
--- a/crates/stella-cli/src/agent.rs
+++ b/crates/stella-cli/src/agent.rs
@@ -68,8 +68,8 @@ mod turn_close;
 // close-out sets the count. The run summary prints it. The self-driving loop
 // adds it up over a run.
 pub(crate) use audit_writes::{
-    dropped_audit_writes, note_child_dropped_audit_writes, take_dropped_audit_writes,
-    warn_dropped_audit_writes, warn_store_write_failed,
+    note_child_dropped_audit_writes, take_dropped_audit_writes, warn_dropped_audit_writes,
+    warn_store_write_failed,
 };
 pub(crate) use budget::{build_budget_guard, remaining_budget, settle_reflection_budget};
 pub(crate) use graph_view::{graph_query_snapshot, graph_snapshot, graph_snapshot_focus};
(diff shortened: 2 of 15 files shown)


macanderson:
**Where each DoD box's evidence is, and the one that reads differently than written.**

Every box on #5626 is ticked against this PR's CI run, not a local build — this machine does not compile (CLAUDE.md, "CI builds and tests; this laptop does not"). So the last box, "Narrow verification only, per SCR-001: `cargo test -p stella-store` and `cargo test -p stella-cli`", was satisfied by the `fmt + clippy + test` job rather than by two local commands. What SCR-001 forbids is a full-suite run *during development*; no cargo command ran on the maintainer's machine at all, and the authoritative check is CI's, which the box itself says.

Two boxes to read carefully:

- **"The change lands inside `crates/stella-store` and `crates/stella-cli/src/agent/{persistence.rs,turn_close.rs}`."** It also touches `crates/stella-cli/src/self_driving_cmd/{stats.rs,work.rs}` and adds one field to `stella_autonomy::SessionStats`. The counter box names `self_driving_cmd/stats.rs` itself, and the field is what carries the count into the durable session stats that command reads. The constraints that box actually sets — no new crate, no file pushed past its ceiling — both hold: `file size ratchet` is green, and the two files that would have crowded (`stella-store/src/lib.rs`, `stella-cli/src/agent.rs`) each got *smaller*.
- **Witness.** The busy/locked witness lives in `crates/stella-store/src/busy.rs` and reproduces the refusal on a real `store.db` with two connections, rather than through `Store::record_agent_uses` — the store's own connection sets `busy_timeout` to five seconds, so a witness on that path would have to hold the competing transaction open for longer than that before any busy error exists to retry. The test sets the timeout to zero on both connections instead, which is the same refusal on the same table with a deterministic clock. The counter and the reported code are witnessed end-to-end through `record_execution_end` in `crates/stella-cli/src/agent/audit_writes.rs`.

`guard self-tests` is red on this PR and on every PR opened after #5648 merged, for a reason that has nothing to do with this diff — see the separate note below.

macanderson:
**`guard self-tests` is red here for a reason outside this diff, and it is already tracked.**

`scripts/test-releases-published-pagination.sh` pins its fail-side witness to `origin/main`'s copy of `check-releases-published.sh` and asserts that copy still hits the old 1000-item cap. #5648 merged that fix into `main`, so the arm's premise is gone and the job now fails on every PR opened since — #5667 fails identically, and #5662, opened before the merge, is green. Nothing in this branch touches either script.

Already filed by peers: #5671 and #5681. Not re-filed here, and not repaired here — the durable fix is a decision about how that witness pins its old implementation, which belongs in its own PR rather than folded into an unrelated one.

PR #5662, opened 2026-09-02
fix(stella-autonomy): make escalation a cooldown instead of a tombstone

### LINKED ISSUE #5621: An escalated issue is parked forever: no retry path, so one bad turn removes work from the backlog permanently
## What happens
When a turn fails unrecoverably, `drive` labels the issue `agent-escalated` and every later run skips it (`triage.rs escalate()`). There is no path back into the queue that does not involve a human removing the label.

Observed on rainforest#17: the turn aborted for a purely ENVIRONMENTAL reason (stale node_modules made every bash call return identical output, tripping the stuck-loop guard — see #5609). The issue was fine; the environment was not. It was parked anyway, and would have stayed parked forever in an unattended run.

## Why it matters
Under unattended operation this is a slow leak: every transient failure — a provider outage, a flaky install, a rate limit, a bad model turn — permanently removes an issue from the backlog. A loop that runs for a week ends with a queue full of parked work and no memory of why.

## Ask
Escalation should be a cooldown, not a tombstone:
- Record the escalation reason and a retry count on the issue (comment or label suffix).
- Let a later run re-claim an escalated issue after N cycles or a configurable backoff, ideally with something changed — different worker model, fresh environment prep, or the reason text fed into the turn's context.
- Only after K escalations does it become permanently skipped, and that terminal state should be visible in `stella self-driving stats`.
- Environmental aborts (stuck-loop from identical tool output, provider error, install failure) should be distinguished from genuine "this work is beyond the loop" and retried more eagerly.

## Definition of done

- [x] Escalation records the failure reason and an attempt/retry counter on the issue (comment or label suffix), not just the flat `agent-escalated` label `triage.rs::escalate()` applies today.
- [x] A later `drive` run can automatically re-claim an escalated issue after N cycles or a configurable backoff — no human has to remove the label by hand.
- [x] Environmental aborts (stuck-loop from identical tool output, provider error, install failure — see #5609) are classified separately from "this work is genuinely beyond the loop," and retried more eagerly than the latter.
- [x] Only after K escalations does an issue become permanently skipped, and that terminal state is visible in `stella self-driving stats`.
- [x] New decision logic (the backoff schedule, escalation-reason classification) lands as pure functions in `stella-autonomy` per invariant #2 (no I/O in the engine) — the label/comment I/O stays in `stella-cli`'s `self_driving_cmd`, matching the existing driver/decision-core split.
- [x] No new or touched file crosses the 1500-line god-file ceiling (`make gate`'s `file-size` guard) — `triage.rs` and any new module stay under it.
- [x] **Verification (witness):** reproduce the rainforest#17 shape — an escalated issue whose abort was environmental — and show it fails to requeue on the old code, then requeues automatically (no label surgery) on the fix.
- [x] Docs updated: the self-driving loop's own design doc under `docs/spec/` (if one exists) and `stella self-driving stats`/`watch` help text describe the cooldown-vs-tombstone lifecycle; `AGENTS.md` / the affected crate's README updated if this changes a crate boundary.
- [x] Plain, direct prose throughout any doc comments the fix adds — this repo's `make prose` gate applies.

macanderson:
## What and why

An escalated issue left the backlog for good. `triage.rs::escalate()` applied a flat `agent-escalated` label and every later run skipped it; only a person removing that label could put it back. A turn that aborted because the machine broke — a stale checkout making every `bash` call return byte-identical output, a provider outage, a failed install — cost the backlog an issue that was never the problem. Over an unattended week that is a slow leak: the queue fills with parked work and nothing remembers why.

Escalation is now a cooldown with an end.

**Decision logic, pure, in `stella-autonomy`** (`crates/stella-autonomy/src/escalation.rs`, no I/O, per AGENTS.md invariant #2):

- `EscalationReason { Environmental(EnvCause) | BeyondLoop }`, with `EnvCause` naming the stuck loop, the provider error, and the install failure.
- `classify(why)` reads the abort message against a fixed phrase list — the same shape as the residue gate's detector, and every phrase is text this tree actually writes (`byte-identical` from the loop guard, `could not start` from a failed spawn, the HTTP words providers spell out). Anything unmatched is `BeyondLoop`, the **longer** wait, so an unrecognised message can only make the loop slower, never more eager.
- `EscalationRecord { attempts, last_reason, last_at, last_at_unix }`, `retry_after(record, policy) -> Option<Duration>` (`None` = parked), `parked`, `may_retry(record, policy, now_unix)`, and `next` for the count that carries forward.
- `EscalationPolicy` defaults: 600s environmental, 21600s beyond-loop, `park_after = 3`. Settable as `[self_driving.escalation]` in `stella.toml`.

**I/O in `stella-cli`'s `self_driving_cmd`:**

- `backlog::escalate` writes three things for three readers — the label a person scans for, the comment a person reads, and the record the next run reads. It returns the record so the caller can count a parked issue.
- `ready::fold_ready` parses the record out of the body it already read and hands it to `ready_queue`; the clock is read here, and `stella-autonomy` takes the moment as an argument.
- `drive/triage.rs` and `drive.rs` are call-throughs only. `drive.rs` got **shorter** (1495 → 1488 lines).
- `stats.rs` prints `parked` beside `escalated`; `SessionStats::issues_parked` is declared in `PRODUCERS` like every other counter.

### Two departures from the design pointer, and why

1. **The record lives in the issue body, not in a comment.** `stella_protocol::issue::IssueProvider` has `comment()` (write) and no comment **read** — so the readback the design needs does not exist, and adding it means a new required trait method plus edits to all nine implementors, eight of which are test fixtures. The body is already carried by the read `ready::fold_ready` makes (`list_open`), so a cooldown costs zero extra tracker calls; the comment form would cost one API read per escalated issue per cycle. It also reuses the exact readback path readiness already uses for `Blocked by:` lines. The marker is an HTML comment (`<!-- stella-escalation {json} -->`), invisible in the rendered issue, and `stamp` replaces the old one rather than stacking.
2. **`retry_after` returns a wall-clock `Duration`, not "N cycles."** `stella-autonomy` takes no date-parsing dependency (serde/serde_json/sha2 only), so the record stores `last_at_unix` beside the human-readable `last_at` — the convention `CycleRecord` already uses (`ended_at` + `ended_at_unix`). "Environmental after 1 cycle" is honoured as *the shorter of two cooldowns*, ten minutes by default, rather than by plumbing a cycle counter into a queue read that has none.

3. **Both work generators consult the cooldown, not just `ready.rs`.** The pointer scoped the consult to `ready.rs`; Sourcery's first review then flagged, correctly, that the *default* `drive` path is the defect queue (`priority::triage` / `rank_defects`), where the tombstone would have survived. `QueueIssue` now carries the record itself — put there by `to_queue_issue` out of the body the queue read already returned — and both generators ask one shared rule, `QueueIssue::escalation_holds`. `BacklogItem` lost its own copy for the same reason. The `ranked`/`ranked_keys`/`unassessed`/`render_queue`/`demand_from` family takes the whole `LoopConfig` instead of the triage policy alone, which is what carries the escalation policy and the clock down to the fold.

## Exemplar followed

`stella_autonomy::doctrine::Doctrine` — a pure policy struct in the leaf crate, serde-derived with `#[serde(default)]`, plugged straight into `SelfDrivingSection` as its own `stella.toml` table. `EscalationPolicy` is the same shape, which is why the config surface costs six lines instead of a new section type.

## Witness tests and their mechanism

I cannot run tests on this machine (CLAUDE.md, "CI builds and tests; this laptop does not"), so here is why each fails on `main` by construction:

1. `an_environmental_escalation_leaves_the_queue_and_returns_when_its_cooldown_is_over` (`crates/stella-autonomy/src/ready.rs`) — the rainforest#17 shape. An issue labelled `agent-escalated` whose record says the abort was a stuck loop is absent from `ready_queue` at the moment it was escalated and **present** once the cooldown has passed, with the label untouched. **Fails on old code by construction:** `ready_queue`'s old body was `if item.issue.has_label(ESCALATION_LABEL) { continue; }` with no cooldown and no `escalation` field on `BacklogItem` — the test does not compile against it, and the queue it asserts on is unconditionally empty.
2. `the_backlog_generator_offers_an_escalated_issue_again_once_its_cooldown_is_over` (`crates/stella-cli/src/self_driving_cmd/ready.rs`) — the same shape end to end, through the real generator and a fixture tracker: a cooled escalation comes back, one at `park_after` stays parked, and the fixture's write counter is 0, so nothing removed a label.
3. `an_environmental_escalation_returns_to_the_defect_queue_once_its_cooldown_is_over` (`crates/stella-autonomy/src/priority.rs`) — the same three states through the **default** `drive` path. **Fails on old code:** `triage`'s escalated arm was `if issue.has_label(ESCALATION_LABEL) { continue; }`, with no policy argument and no record on `QueueIssue` to consult.
4. `escalating_stamps_a_record_the_next_run_can_read` (`crates/stella-cli/src/self_driving_cmd/backlog.rs`) — the write half. `escalate_blocking` applies the label **and** edits the body; the stamped body round-trips back through `escalation::parse`; a second escalation on that body reports `attempts == 2`. **Fails on old code:** `escalate` returned `Result<(), IssueError>` and never called `provider.edit`.
5. `an_issue_escalated_to_the_ceiling_stays_out_of_the_ready_queue` and `an_issue_escalated_park_after_times_is_never_taken_again` — the cooldown has an end; `i64::MAX` does not make a parked issue takeable.
6. `an_escalated_issue_with_no_record_never_enters_the_ready_queue` and `an_escalated_issue_is_not_re_asked` — a label a person applied by hand still means what it meant.
7. `escalation.rs` unit tests cover classification both ways, the two cooldown lengths, the marker round-trip, damaged markers, and the carried count.

Guards run locally: `make guards-fast` green (prose, line-citations, dead-code-allows, module-reachability, lockfile-sync, fmt), `./scripts/check-file-size.sh` green — `drive.rs` 1488/1500, `backlog.rs` 1449/1500, nothing grandfathered, no baseline touched.

## Docs

- `docs/spec/backlog-self-driving.md` — a new "Escalation is a cooldown, not a tombstone" section.
- `website/content/docs/commands/self-driving.mdx` — what `escalated` vs `parked` mean in `stats`.
- `website/content/docs/configuration/stella-toml.mdx` and `stella.example.toml` — `[self_driving.escalation]`.
- `crates/stella-autonomy/README.md` — rows for `escalation.rs` and for `ready.rs`, which the table had been missing.
- `SelfDrivingCmd::Stats` clap help.

## Tests deleted

`an_escalated_issue_never_enters_the_ready_queue` (`crates/stella-autonomy/src/ready.rs`) was **renamed** to `an_escalated_issue_with_no_record_never_enters_the_ready_queue`. Its body and assertion are unchanged; the old name asserted a claim that is no longer true in general, and the new one names the case it actually covers.

## Sourcery

The first review raised two ❌ rows, both saying the defect queue was left as a tombstone. Both are now **fixed**, not answered — see departure 3 above and the new `priority.rs` witness. The residue issue filed for that gap (5663) is closed as superseded rather than left open.

## Remaining work filed

None. The gap this PR's first push left is closed inside it.

Closes #5621

## Summary by Sourcery

Make issue escalation recoverable with reason-aware cooldowns while parking issues permanently only after repeated failed attempts.

New Features:
- Replace permanent escalation tombstones with configurable cooldowns that automatically return issues to both the ready and defect queues.
- Persist escalation attempts, reasons, and timestamps in issue bodies and classify environmental failures separately from unresolved work.
- Report permanently parked issues separately from temporary escalations in self-driving statistics.

Bug Fixes:
- Prevent transient machine, provider, and installation failures from permanently removing valid backlog issues.

Enhancements:
- Centralize escalation eligibility decisions in the autonomy policy layer and preserve manually escalated issues without records as excluded.
- Add configurable environmental and beyond-loop cooldowns plus an escalation attempt ceiling.

Documentation:
- Document escalation cooldown behavior, parked issues, statistics, and the new self-driving escalation configuration.

Tests:
- Add coverage for escalation classification, record persistence, cooldown expiry, queue re-entry, parking, and missing or damaged records.

### DIFF
--- a/crates/stella-autonomy/README.md
+++ b/crates/stella-autonomy/README.md
@@ -83,6 +83,8 @@ the CLI does — so the dashboard and the terminal cannot disagree.
 | [`src/deliver.rs`](src/deliver.rs) | The per-PR delivery machine: `PrState`/`Observation`/`deliver_next`, the fix/rebase ceilings, and the escalation reasons. |
 | [`src/doctrine.rs`](src/doctrine.rs) | The policy knobs a human sets: `Doctrine`, `ForeignBreakage`, contention policy and `contention_verdict`. |
 | [`src/priority.rs`](src/priority.rs) | `triage`: rank by the ladder the repo actually uses, with unassessed issues distinct from unrankable ones. |
+| [`src/ready.rs`](src/ready.rs) | `ready_queue`: which backlog issues may be taken next, and in what order. `Blocked by: #N` lines, the `status:ready` override, and the escalation cooldown. |
+| [`src/escalation.rs`](src/escalation.rs) | Escalation as a cooldown: `classify` reads an abort message, `retry_after` says how long to wait, `park_after` says when waiting ends. The record is written into the issue body and read back from it. |
 | [`src/gate.rs`](src/gate.rs) | Which checks are allowed to block a merge, and which have stopped earning that right. |
 | [`src/closure.rs`](src/closure.rs) | How an issue ends, and what has to be true first. |
 | [`src/convention.rs`](src/convention.rs) | How this repository writes issues, learned rather than assumed. |
--- a/crates/stella-autonomy/src/escalation.rs
+++ b/crates/stella-autonomy/src/escalation.rs
@@ -1,0 +1,419 @@
+//! Escalation as a cooldown, not a tombstone.
+//!
+//! The loop labels an issue it could not finish. Dropping it for good is
+//! right when the work is too hard. It is wrong when the machine broke: a
+//! stale install, a provider outage, a failed setup. Treat the two alike
+//! and one bad hour costs the backlog an issue.
+//!
+//! So an escalation carries a reason and a count. The reason sets how soon
+//! to try again. The count sets when to stop. Each rule here is a plain
+//! function over owned data. `stella-cli` reads the tracker, writes the
+//! record, and asks these rules what it means.
+//!
+//! Reached as `stella_autonomy::escalation::*`. Nothing is re-exported at
+//! the crate root. A type named [`crate::EscalationReason`] is already
+//! there, and it is about a pull request. One name at one address has to
+//! mean one thing.
+
+use std::time::Duration;
+
+use serde::{Deserialize, Serialize};
+
+/// Opening text of the record's marker in an issue body.
+pub const MARKER_OPEN: &str = "<!-- stella-escalation ";
+
+/// Closing text of that marker.
+pub const MARKER_CLOSE: &str = " -->";
+
+/// What broke, when the machine broke rather than the work being too hard.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum EnvCause {
+    /// The turn made one call over and over and got the same bytes back.
+    /// A stale checkout does this to every command that reads it.
+    StuckLoop,
+    /// The model provider refused the call or dropped it.
+    ProviderError,
+    /// A tool the turn needs was missing, or would not build.
+    InstallFailure,
+}
+
+/// Why the loop handed an issue back.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum EscalationReason {
+    /// The machine failed. The issue itself may be fine.
+    Environmental(EnvCause),
+    /// The turn ran and could not do the work.
+    BeyondLoop,
+}
+
+/// What the loop has already tried on one issue.
+///
+/// It goes in the issue body, inside [`MARKER_OPEN`]. So it lives through a
+/// restart and a move to a new box, and it needs no new store. The queue
+/// read already carries every body, so reading it back is free.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct EscalationRecord {
+    /// How many times this issue has been escalated, this one included.
+    pub attempts: u32,
+    /// Why the last one happened.
+    pub last_reason: EscalationReason,
+    /// When the last one happened, in RFC 3339, for a person to read.
+    pub last_at: String,
+    /// The same moment in whole seconds since the epoch, for the math.
+    /// The text form takes a date parser to compare, and this crate has
+    /// none. [`crate::CycleRecord`] keeps both fields for the same reason.
+    pub last_at_unix: i64,
+}
+
+/// How long each kind of escalation waits, and when waiting ends.
+///
+/// Set in `[self_driving.escalation]` in `stella.toml`, beside the rest of
+/// the loop's judgement calls.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(default)]
+pub struct EscalationPolicy {
+    /// Seconds to wait when the machine broke.
+    pub environmental_cooldown_secs: u64,
+    /// Seconds to wait when the turn could not do the work.
+    pub beyond_loop_cooldown_secs: u64,
+    /// Tries after which the issue is parked and never taken again.
+    pub park_after: u32,
+}
+
+impl Default for EscalationPolicy {
+    /// Ten minutes for the machine, six hours for the work, parked on the
+    /// third try.
+    ///
+    /// A broken box clears in minutes. A rate-limit window closes, an
+    /// install is fixed, an outage ends. Ten minutes covers that, and the
+    /// issue is back inside the hour.
+    ///
+    /// A turn that ran and could not do the work needs something to change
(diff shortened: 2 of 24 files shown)


macanderson:
@sourcery-ai review

The two ❌ rows in the assessment above were against the first push and are now **fixed**, not deferred. Commit `9474d84` extends the cooldown to the defect queue:

- `QueueIssue` carries the escalation record, put there by `to_queue_issue` out of the body the queue read already returned — so `priority::triage` and `rank_defects` can read it.
- Both generators ask one shared rule, `QueueIssue::escalation_holds`, so the ready backlog and the defect queue cannot disagree about when an escalated issue comes back or when it is parked.
- Witness: `an_environmental_escalation_returns_to_the_defect_queue_once_its_cooldown_is_over` in `crates/stella-autonomy/src/priority.rs` — out during the cooldown, in once it is over, out for good at `park_after`, with no label removed.

The residue issue filed for that gap (#5663) is closed as superseded.

PR #5667, opened 2026-09-03
fix(stella-model): refuse a tool call announced without an id

### LINKED ISSUE #5496: A malformed Anthropic tool-call start drops the whole tool call and the turn reports none
## The problem

When Anthropic's `content_block_start` for a `tool_use` block is missing a
required field, the whole event fails to deserialize, the stream aggregator
drops it as an unparseable frame, and the tool call it announced disappears. The
turn then completes successfully, reporting zero tool calls, while
`finish_reason` still says the model made one.

A dropped tool call that reads as "the model asked for nothing" is the worst
shape this adapter can fail in: nothing errors, and the turn proceeds.

## Where

- `crates/stella-model/src/anthropic.rs`, `AnthropicStartBlock::ToolUse` — `id`
  and `name` are required, with no `#[serde(default)]`. Serde's
  internally-tagged enum matches the `type` tag first and then fails the whole
  event if a required field of the matched variant is absent.
- `crates/stella-model/src/anthropic/stream.rs`, `aggregate_anthropic_stream` —
  the `Err(_)` arm drops the frame and continues, commented as tolerating an
  unmodelled event type. That is right for an unknown `type` and wrong for a
  recognised one with an incomplete payload.

## Failure scenario

An SSE body whose `content_block_start` names `tool_use` and carries `name` but
no `id`:

```
event: content_block_start
data: {"type":"content_block_start","index":0,"content_block":{"type":"tool_use","name":"bash"}}

event: content_block_delta
data: {"type":"content_block_delta","index":0,"delta":{"type":"input_json_delta","partial_json":"{\"command\":\"ls\"}"}}

event: content_block_stop
data: {"type":"content_block_stop","index":0}

event: message_delta
data: {"type":"message_delta","delta":{"stop_reason":"tool_use"},"usage":{"output_tokens":10}}
```

The start event is dropped, so `tool_uses` gets no entry for index 0. The delta's
`get_mut(&0)` misses and the argument fragment is discarded. The stop's `get(&0)`
misses and nothing is announced. At end of stream the map is empty, so the result
is `Ok` with `tool_calls: []` and `finish_reason: Some(ToolCalls)` — the model's
own `stop_reason` says it called a tool and the adapter says it called none.

Reachable through any gateway or proxy speaking this dialect, which
`with_base_url` documents as a supported path, and through any normaliser that
drops a field it does not recognise.

## The fix

Give `AnthropicStartBlock::ToolUse` `#[serde(default)]` sentinels so the event
still parses, then treat an empty `id` at assembly time as a named
`ProviderError` rather than letting the block vanish. That matches how a
complete-but-unparseable argument buffer already degrades: loudly, through the
`Value::Null` repair sentinel, rather than silently.

## Definition of done

- [x] A recognised `content_block_start` with an incomplete payload produces an
      error or a repaired call, never a silently empty `tool_calls`.
- [x] Witness (wiremock): mount `/v1/messages` returning the body above; assert
      the result is either an `Err` naming the malformed block or
      `tool_calls.len() == 1`. Today it is `Ok` with an empty vec beside
      `finish_reason: ToolCalls`.
- [x] `cargo test -p stella-model` green.

## Confidence

CONFIRMED by the audit that read the adapter, tracing the frame through
`aggregate_anthropic_stream`'s `Err` arm and the three `get`/`get_mut` misses
that follow. Not reproduced against a live endpoint.

## Related

The same slice reported two neighbouring shapes worth fixing together:

- `stella-model/src/zai/stream.rs` derives a missing `index` from the fragment's
  position *within the current chunk*, which is re-enumerated per SSE frame. A
  server that omits `index` and sends one call per frame merges every call into
  one accumulator, concatenating names into `read_filesearch` and arguments into
  invalid JSON — both real calls lost.
- The same file commits a completed call with an empty `call_id` when the id
  never arrives, while the speculative path explicitly skips that case. Two
  same-turn calls sharing `""` then mis-pair with their results in
  `stella-core`'s loop evidence.

macanderson:
## What and why

A `content_block_start` naming `tool_use` with no `id` did not deserialize.
Serde matches the `type` tag first, then fails the whole event on an absent
required field — so the frame landed in `aggregate_anthropic_stream`'s
`Err(_)` arm, the arm that exists to tolerate an event *type* this adapter
does not model. The block went with it, the `input_json_delta` fragments
missed `tool_uses.get_mut`, and the turn returned `Ok` with `tool_calls: []`
beside `finish_reason: Some(ToolCalls)`. The model's own stop reason says it
called a tool; the adapter says it called none. Nothing errors, and the turn
proceeds — a missing measurement rendering as a negative one.

Fix, per the design pointer on the issue:

- `AnthropicStartBlock::ToolUse` gets `#[serde(default)]` on `id` and `name`,
  so the frame parses and the loss becomes visible.
- `aggregate_anthropic_stream` refuses an empty `id` or `name` at assembly,
  ahead of the truncation arms (neither field can be a casualty of the token
  limit — the whole start frame rides one SSE event).
- The unknown-`type` tolerance is unchanged, and is now pinned by a test.

## The named error

The pointer asked for `ProviderError::MalformedToolCall { index, missing }`.
This ships it as a named constructor —
`http::malformed_tool_call_error(provider, index, name, missing)` — returning
`ProviderError::Malformed`, rather than as a new enum case. Two reasons, both
in the code:

1. `ProviderError::Malformed`'s own doc comment already names this exact case:
   *"a response arrived but could not be decoded … a missing required field, an
   unparseable tool-call argument. Terminal."* The classification a caller acts
   on is identical — terminal, not retryable, not park-eligible, no recovery
   path — and AGENTS.md invariant 5's test for a new case is *whether a caller
   has to branch on it*. Nothing does.
2. A new `ProviderError` case forces a new `ProviderErrorWire` case in
   `crates/stella-serve/src/frame.rs` and a regenerated
   `docs/wire/serveframe.schema.json` / `serveinbound.schema.json` / `.d.ts` —
   the wire contract out-of-tree hosts parse by hand, which AGENTS.md calls
   "the single most dangerous drift in this document". A remote host can never
   emit this fault (only an in-tree SSE aggregator can), so the wire would gain
   a case nothing on the far side produces.

The constructor keeps everything the pointer asked the case to carry: the
block index, the missing field names, and the tool name, all in the message a
human reads. It sits beside `truncated_tool_input_error` in `http.rs`, the
existing shared constructor for the sibling "this tool call cannot be
dispatched" fault, and is the exemplar followed.

If the maintainer wants the distinct case anyway, say so and it lands as a
follow-up that touches the protocol crate, `stella-serve` and the generated
schemas together.

## Witnesses

Three, all wiremock at the provider boundary — the assertion has to be about
what a caller receives, because nothing on the old code errors.

`crates/stella-model/src/anthropic/malformed_tool_use.rs` (declared from
`anthropic.rs`, not from the `anthropic/tests.rs` god file):

- `a_tool_use_block_with_no_id_ends_the_turn_instead_of_vanishing` — the SSE
  body from the issue verbatim. **Mechanism:** on `main` the start frame does
  not deserialize, `tool_uses` stays empty, and `complete` returns `Ok` with
  `tool_calls: []`; the `expect_err` is what fails. With the change the fields
  default, assembly sees an empty `id`, and the error names `index 0`, `` `id` ``
  and `bash`.
- `a_unary_tool_use_block_with_no_id_ends_the_turn_as_well` — the same block on
  the non-streaming fallback path. **Mechanism:** `AnthropicUnaryBlock::ToolUse`
  already defaulted both fields, so on `main` this returns `Ok` with a
  `ToolCall` whose `call_id` is `""` — a call that can never be matched to its
  result. The `expect_err` fails on `main`. Included because AGENTS.md requires
  the two delivery paths not to drift on assembly rules.
- `a_content_block_type_this_adapter_does_not_model_is_still_skipped` — the
  constraint: an unmodelled block type, an unmodelled event type, and a data
  line with no `type` at all are all still skipped. Passes on `main` too; it
  exists to fail if the refusal is ever widened past `tool_use`.

`crates/stella-model/src/zai/malformed_tool_calls.rs` (declared from `zai.rs`,
not from the `zai/tests.rs` god file) — the two residues the issue's "Related"
section names, both under the 40-line bar the pointer set:

- `two_tool_calls_with_no_index_stay_two_calls` — `index` fell back to the
  fragment's position inside *its own chunk*, which `enumerate()` restarts per
  SSE frame. **Mechanism:** on `main` a server sending one index-less call per
  frame files both under 0; the turn comes back with one call named
  `read_filesearch` whose two argument objects are concatenated into
  unparseable JSON, degraded to the `Null` repair sentinel. The
  `assert_eq!(names, ["read_file", "search"])` fails. The fallback is now a
  counter that spans the stream, advanced by a fresh `id` (what opens a call in
  this dialect) rather than by frame position.
- `a_tool_call_whose_id_never_arrives_ends_the_turn` — `announce_completed_below`
  already skips a call with an empty `id`; final assembly committed one anyway.
  **Mechanism:** on `main` this returns `Ok` with `call_id: ""`, so the
  `expect_err` fails. Two such calls in one turn mis-pair with their results in
  `stella-core`'s loop evidence.

`crates/stella-model/src/zai/tests/stream_frame.rs`'s existing
`a_tool_call_fragment_without_an_index_is_not_dropped` still passes unchanged
(one call, id present, counter stays at 0); its assertion message was updated
to describe the counter instead of the position.

## Not verified locally

Per CLAUDE.md, nothing was compiled or run on the maintainer's machine — the
evidence for this PR is its CI run. `cargo fmt --all`, `make guards-fast`,
`check-prose` and `check-file-size` were run locally and are green.

Tests deleted: None.

Residue filed: none beyond this PR — the two Z.ai shapes the issue lists are
fixed here rather than filed, and the unary Anthropic path was the third
instance and is fixed here too.

Closes #5496

## Summary by Sourcery

Fail explicitly on malformed tool calls and maintain correct assembly of index-less Z.ai calls across streamed frames.

Bug Fixes:
- Reject malformed Anthropic tool-use blocks and unary tool calls that lack an ID or name instead of silently returning no or uncorrelatable tool calls.
- Preserve separate Z.ai tool calls when streamed fragments omit indexes.
- Reject Z.ai tool calls whose IDs never arrive instead of committing calls that cannot be correlated.

Enhancements:
- Keep unsupported Anthropic block and event types skippable while making malformed modeled tool calls terminal, non-retryable errors.
- Share consistent malformed-tool-call error reporting across provider adapters.

Tests:
- Add provider-boundary coverage for malformed Anthropic streaming and unary tool calls, unsupported event tolerance, and malformed Z.ai tool calls.

### DIFF
--- a/crates/stella-model/src/anthropic.rs
+++ b/crates/stella-model/src/anthropic.rs
@@ -771,8 +771,26 @@ struct AnthropicMessageStart {
 #[derive(Deserialize, Debug)]
 #[serde(tag = "type", rename_all = "snake_case")]
 enum AnthropicStartBlock {
+    /// Both fields default. An empty one is a terminal error at assembly,
+    /// not a frame to skip.
+    ///
+    /// Serde reads the `type` tag first. It then fails the whole event when
+    /// the matched case is missing a field. Without these defaults, a
+    /// `tool_use` block with no `id` lands in
+    /// [`stream::aggregate_anthropic_stream`]'s `Err(_)` arm — the arm that
+    /// skips an event type this adapter does not model. The block goes with
+    /// it. Its `input_json_delta` parts miss the map. The turn ends `Ok`
+    /// with no tool calls, next to a finish reason of `ToolCalls`: the
+    /// model's own stop reason says it called a tool, and the adapter says
+    /// it called none.
+    ///
+    /// The defaults keep the frame readable, so the loss is visible.
+    /// [`stream::aggregate_anthropic_stream`] then refuses to commit a call
+    /// it cannot dispatch.
     ToolUse {
+        #[serde(default)]
         id: String,
+        #[serde(default)]
         name: String,
     },
     #[serde(other)]
@@ -1203,6 +1221,8 @@ fn map_stop_reason(stop_reason: Option<&str>) -> Option<FinishReason> {
     }
 }
 
+#[cfg(test)]
+mod malformed_tool_use;
 #[cfg(test)]
 mod parallel_tool_calls;
 #[cfg(test)]
--- a/crates/stella-model/src/anthropic/malformed_tool_use.rs
+++ b/crates/stella-model/src/anthropic/malformed_tool_use.rs
@@ -1,0 +1,173 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+//! What a `tool_use` start frame does to a turn when a field is missing.
+//!
+//! Serde reads the `type` tag first. It then fails the whole event when the
+//! matched case is missing a field. So a `content_block_start` that names
+//! `tool_use` and has no `id` does not parse at all, unless the fields
+//! default. The frame lands in the arm that skips an event type this adapter
+//! does not know. The block goes with it. Its `input_json_delta` parts miss
+//! the map. The turn ends `Ok` with no tool calls, next to a finish reason
+//! of `ToolCalls`.
+//!
+//! Nothing fails and the turn goes on. That is why the proof runs at the
+//! provider edge, not on the type. The claim is about what a caller gets.
+//!
+//! Declared from `anthropic.rs`, not from `anthropic/tests.rs`. That file is
+//! a god file closed to growth.
+
+use super::*;
+use stella_protocol::CompletionRequest;
+use wiremock::matchers::{body_string_contains, method, path};
+use wiremock::{Mock, MockServer, ResponseTemplate};
+
+/// A `tool_use` block carrying `name` and no `id`, its arguments streamed and
+/// closed, ending on a `stop_reason` of `tool_use`. Every frame after the
+/// start is well formed — the whole loss comes from the one absent field.
+const TOOL_USE_WITHOUT_AN_ID: &str = concat!(
+    "event: content_block_start\n",
+    "data: {\"type\":\"content_block_start\",\"index\":0,\"content_block\":{\"type\":\"tool_use\",\"name\":\"bash\"}}\n\n",
+    "event: content_block_delta\n",
+    "data: {\"type\":\"content_block_delta\",\"index\":0,\"delta\":{\"type\":\"input_json_delta\",\"partial_json\":\"{\\\"command\\\":\\\"ls\\\"}\"}}\n\n",
+    "event: content_block_stop\n",
+    "data: {\"type\":\"content_block_stop\",\"index\":0}\n\n",
+    "event: message_delta\n",
+    "data: {\"type\":\"message_delta\",\"delta\":{\"stop_reason\":\"tool_use\"},\"usage\":{\"output_tokens\":10}}\n\n",
+    "event: message_stop\n",
+    "data: {\"type\":\"message_stop\"}\n\n",
+);
+
+/// Three frames this adapter does not model — a block type it has no case
+/// for, an event type it has no case for, and a data line with no `type` at
+/// all — around an ordinary text block. All three must still be skipped.
+const AN_UNMODELLED_BLOCK_TYPE: &str = concat!(
+    "event: content_block_start\n",
+    "data: {\"type\":\"content_block_start\",\"index\":0,\"content_block\":{\"type\":\"server_tool_use\",\"id\":\"srvtoolu_1\",\"name\":\"web_search\"}}\n\n",
+    "event: content_block_stop\n",
+    "data: {\"type\":\"content_block_stop\",\"index\":0}\n\n",
+    "event: ping\n",
+    "data: {\"type\":\"ping\"}\n\n",
+    "event: something_new\n",
+    "data: {\"unrecognised\":true}\n\n",
+    "event: content_block_start\n",
+    "data: {\"type\":\"content_block_start\",\"index\":1,\"content_block\":{\"type\":\"text\",\"text\":\"\"}}\n\n",
+    "event: content_block_delta\n",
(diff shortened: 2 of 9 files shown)


macanderson:
`guard self-tests` is red here on a step unrelated to this diff: `scripts/test-releases-published-pagination.sh`'s `MAIN` arm asserts that `main`'s copy of `check-releases-published.sh` fails on 1001 releases, and #5648 landed that fix on `main`, so the expectation is now stale. It is red on #5658 too, and green on the pull requests whose base predates that merge (#5662, #5666) — which pins it to the merge, not to any branch. Not a required check, and filed as #5671.

No `stella-model` change landed on `main` since this branch's base (bc3e671b7 back to 5096b3ff7 touches `stella-tui`, `stella-core`, `stella-cli` and `scripts/`), so there is nothing to rebase onto here.

Sourcery's assessment table marks all three linked objectives ✅ with no ❌ rows.

macanderson:
Remaining work filed after the merge:

- **#5684** — the same two shapes on the adapters this PR did not touch. `openai.rs`'s streamed `FunctionCall` declares `call_id` with no `#[serde(default)]`, so a frame that omits it fails to deserialize and the whole call is dropped — #5496's failure verbatim, one dialect over. And `openai/stream.rs`, `openai/unary.rs` and `zai/unary.rs` each commit a call with an empty `call_id` while their own speculative-announce site already refuses one, which is the asymmetry this PR closed in `zai/stream.rs`. `http::malformed_tool_call_error` is already shared, so each is a call-site plus a witness.
- **#5671** — the unrelated `guard self-tests` red described above.

Issue #5676, opened 2026-09-03
Every shell call spends two git status runs even when no other agent shares the folder


macanderson:
## In plain words

Every shell command now costs two `git status` runs, even when nobody else is
working in the folder. On a big repository that is real time spent on a
warning nobody will read.

## Problem

`crates/stella-cli/src/claims.rs`'s `ShellWatch` snapshots the tree before and
after each `bash` call so a shell write can be claimed. Measured on this
repository at 2909 tracked files, one `git status --porcelain -z` takes about
15 ms warm, so a shell call pays about 31 ms. A tree with a large ignored
build directory costs more, and that measurement has not been taken.

The cost buys nothing when this process is the only writer in the workspace —
a plain `stella run`, or a deck with no worker lanes and no fleet. The claim
table already knows who else is holding paths, and the session registry
already knows which sessions are live.

## Files

- `crates/stella-cli/src/claims.rs` — `ShellWatch::snapshot`, the
  observed-write arm of `ClaimTap::execute`
- `crates/stella-store/src/sessions.rs` — the live-session registry

## Fix sketch

Skip the snapshot pair while this workspace has one live writer, and re-arm
the moment a second appears. Whatever is used to decide must fail toward
watching, not toward skipping: a wrong "nobody else is here" costs a lost
edit, a wrong "somebody is" costs 31 ms. Measure a tree with build artifacts
before choosing a threshold.

## Verify

Time a `bash` call in a one-session workspace with and without the change, and
confirm a second live session re-arms the watch.

## Definition of done

- [ ] A single-writer workspace spends no `git status` per shell call
- [ ] A second live writer re-arms the watch, with a test covering the
      transition
- [ ] The skip fails toward watching at every unknown
- [ ] The measurement behind any threshold is written down where the constant
      is, per AGENTS.md's `MEASURED:` rule

macanderson:
Folded into batch #6308 (2026-09-06 audit): fixed together by crate.

macanderson:
Reopened. This member was folded into #6308 and closed, and PR #6333 could not carry it — see that PR's "Could not ride" section for the case. The handoff above is unchanged and still stands.

Issue #5675, opened 2026-09-03
A shell write over another agent's claimed file is noticed and never reported


macanderson:
## In plain words

When a shell command writes a file another agent is already holding, Stella
now notices — and says nothing. The model that just wrote over a sibling's
work never learns it did.

## Problem

`crates/stella-cli/src/claims.rs` claims what a `bash` call turns out to have
written. The command has already run, so the tap cannot refuse it, and the
rival keeps the claim. Today that outcome is dropped on the floor:

```rust
for path in changed(&before, &after) {
    let _ = self.claim(store, &path);
}
```

`Claimed::Rival` means "you just wrote a file `<holder>` is editing". That is
exactly the signal the file-tool route puts in front of the model when it
refuses, and it is the only chance the model gets to stop and check before it
keeps going.

## Files

- `crates/stella-cli/src/claims.rs` — the observed-write arm of
  `ClaimTap::execute`, and `Claimed`

## Fix sketch

Append a line to the successful `ToolOutput` naming the paths a live rival
holds, or emit a diagnostic. A `ToolOutput::Ok` whose content gains a note is
the cheapest route and reaches the model; check the deck's golden frames and
any test asserting exact `bash` output before choosing it.

## Verify

Two sessions on one tree. Session A `edit_file`s `x`. Session B writes `x`
with a shell redirect. B's tool result says nothing about A today.

## Definition of done

- [ ] A shell write onto a live rival's claim tells the model whose claim it
      is
- [ ] The command is still never refused — a test covers that direction
- [ ] Nothing is said when no rival holds the path
- [ ] `cargo test -p stella-cli claims` is green in CI

Issue #5674, opened 2026-09-03
Shell claims and tool claims name the same file two ways when the workspace sits under the repo root


macanderson:
## In plain words

Two agents can edit the same file and still not see each other, if Stella's
workspace folder is not the top of the git repository. The shell side names
the file one way and the file tools name it another, so the two claims never
meet.

## Problem

`crates/stella-cli/src/claims.rs` claims a shell command's writes by reading
`git status --porcelain -z` either side of a `bash` call. Git always prints a
work-tree path relative to the **repository** root. `ClaimTap::claim_key`
makes the file-tool side match by stripping the **workspace** root from an
absolute path.

Those two roots are the same folder in every normal layout, and the claim
keys agree. They are not the same when the workspace root sits inside a
larger repository (a monorepo package, a checkout where `stella` is run from
a subfolder). Then the shell claims `packages/api/src/x.rs` while `edit_file`
claims `src/x.rs`, and the two writers never collide.

Nothing is wrong today for the common layout, and nothing silently breaks —
the failure is a missed warning, which is the direction claim-on-first-write
already fails in when git cannot answer.

## Files

- `crates/stella-cli/src/claims.rs` — `ClaimTap::claim_key`, `ShellWatch`

## Fix sketch

Ask git once per watch for `rev-parse --show-toplevel`, and rebase the
snapshot's paths onto the workspace root (or key everything on the repository
root instead, and rebase the tool side). One extra git call per watch, not
per call, so cache it on `ShellWatch`. `ShellWatch::snapshot` is where the
paths are built.

## Verify

Set a workspace root one level under a repository root, have one session
write a file through `bash`, and have a second session `edit_file` the same
file. Today the second is not refused.

## Definition of done

- [ ] Both routes key one file the same way when the workspace root is under
      the repository root
- [ ] A test covers the nested-root case
- [ ] The extra git call is made once per watch, not once per shell call
- [ ] `claims.rs`'s `claim_key` doc no longer names this as an open gap
- [ ] `make prose` and `make file-size` stay green

Issue #5593, opened 2026-08-31
Our docs say file-lock checks skip all built-in tools, but they don't


macanderson:
## In plain words

`claims.rs` has a note that says "no built-in tool is checked for file locks." That note is wrong. Three built-in tools (`write_file`, `edit_file`, `delete_file`) ARE checked today. A reader who trusts the wrong note could believe the whole safety check is dead and try to build a second one — which almost happened once already. This issue fixes the note and adds a test so it can't drift again.

## Problem

`crates/stella-cli/src/claims.rs`'s module doc and its own code disagree about
whether claim-on-first-write covers anything that ships.

The doc's closing "Coverage note" says:

> No built-in declares a per-path mutation or a lane, and an MCP tool can never
> match (its wire name carries the `mcp____` prefix), so this coverage
> today reaches only workspace custom tools that adopt those conventional
> names; everything else writes outside claim tracking.

The code says otherwise. `mutating_path` in the same file matches four tool
names by hand, and three of them are shipped built-ins in
`crates/stella-tools/src/catalog.rs`:

```rust
fn mutating_path<'i>(name: &str, input: &'i Value) -> Option<&'i str> {
    match name {
        "write_file" | "edit_file" | "delete_file" => input.get("path").and_then(Value::as_str),
        "apply_edits" => /* first edit's path */,
        _ => None,
    }
}
```

So `file_locks` is live for the file-CRUD tools, not inert. A reader who trusts
the doc concludes the whole mechanism is dead and either deletes it or builds a
second one beside it. That is a live hazard, not a typo: a session designing
peer coordination read the doc, believed the mechanism was vestigial, and began
specifying a replacement plane.

The doc also mis-describes the mechanism. It says the tap keys on "a tool's
**schema**", and `mutating_path` keys on the tool **name**. That distinction
matters, because a name match is what makes the MCP-prefix claim in the same
sentence true.

## Files

- `crates/stella-cli/src/claims.rs` — the module doc (the "Coverage note"
  paragraph) and `mutating_path`
- `crates/stella-tools/src/catalog.rs` — the table proving `write_file`,
  `edit_file` and `delete_file` are live built-ins

## Verify

```
rg -n -A 15 'fn mutating_path' crates/stella-cli/src/claims.rs
rg -n '"write_file"|"edit_file"|"delete_file"' crates/stella-tools/src/catalog.rs
```

The first names three built-ins. The second shows they ship. The module doc
says no built-in is named.

## Constraints

[...]

PR #5678, opened 2026-09-03
feat(stella-cli): claim the paths a shell command turns out to write

### LINKED ISSUE #5594: Editing a file through the shell skips our safety check for two people working at once
## In plain words

When one worker edits a file with our built-in `edit_file` tool, we note that down so a second worker sharing the same tree gets warned before stepping on it. But if the first worker edits the same file with a plain shell command instead (`sed -i`, `cat > file`, a formatter, `npm install` touching a lockfile), nothing gets noted down. The second worker gets no warning and can silently overwrite the first worker's work. This issue decides how to close that gap.

## Problem

Claim-on-first-write covers the file tools and misses the shell, which is where
much of the real mutation happens.

`mutating_path` in `crates/stella-cli/src/claims.rs` matches `write_file`,
`edit_file`, `delete_file` and `apply_edits`. It does not match `bash`. So a
turn that edits a file through the shell takes no claim, and a sibling working
in the same tree gets no conflict and no holder name.

That gap is wide in practice. #4151 records that models reach for shell
commands instead of the read tool, bypassing the read-side checks; the same
habit on the write side walks straight past `file_locks`. A `sed -i`, a
`cat > file`, a `git checkout -- path`, an `npm install` rewriting a lockfile,
or a formatter run all mutate the tree with nothing claimed.

The result is that the mechanism's guarantee is much narrower than its name
suggests. Two workers in one shared tree can believe they are coordinated
because claim-on-first-write exists, while both write the same file through
`bash` and neither is told.

## Files

- `crates/stella-cli/src/claims.rs` — `mutating_path`, `transient_lane`,
  `ClaimTap`
- `crates/stella-tools/src/catalog.rs` — `bash` and its group
- `crates/stella-fleet/src/fleet.rs` — the shared-tree case this protects
  (`Isolation::SharedTree` is the default)

## Verify

Run two sessions against one workspace. Have each edit the same path, one with
`edit_file` and one with a shell redirect. The first takes a `file_locks` row;
the second takes none and is never refused.

```
rg -n -A 15 'fn mutating_path' crates/stella-cli/src/claims.rs
```

`bash` is absent from the match.

## Constraints

- A shell command's mutations are not knowable before it runs, so this cannot
  be a path claim minted from arguments the way the file tools do it. Parsing a
  command line to guess the paths it will write is a guess, and a wrong guess
  here is worse than no claim: it would refuse work that was safe.
- The existing `BUILD_CLAIM` transient lane is the shape that already handles
  "a shell command whose writes we cannot enumerate" — it serializes rather
  than claiming paths. Consider whether an unclassified `bash` call belongs in
  a lane rather than in the path table.
- `Isolation::SharedTree` is the fleet default, so any change here affects the
  common case, not an opt-in one.
- Do not widen this into a general rewrite of the claim planes. #5593 covers the documentation half, and a
  decision to merge `file_locks` with `dispatch_claims` is a separate question
  that needs an ADR.

## Definition of done

- [x] A decision is recorded on how an unenumerable shell mutation is
      coordinated: a lane, a coarse whole-tree claim, or an explicit declared
      gap with the reason
- [x] If a mechanism is chosen, two writers racing one path through `bash` are
      coordinated by it, with a witness test
- [x] If a declared gap is chosen instead, `claims.rs`'s module doc names the
      gap and cites this issue, so the next reader is not misled about the
      guarantee
- [x] The chosen behaviour does not refuse a shell command that was safe —
      a test covers the false-refusal direction
- [x] `cargo test -p stella-cli claims` is green
- [x] File organization: the fix lives in `crates/stella-cli/src/claims.rs`, touching `crates/stella-fleet` only if the chosen mechanism genuinely needs a shared type there. It does not push `claims.rs` onto the god-file list, and does not touch any file already on it — `make file-size` stays green.
- [x] Docs: whatever `claims.rs`'s module doc ends up saying about shell coverage (per #5593) matches what actually ships here, and `crates/stella-cli/README.md` (or wherever this crate's claim mechanism is documented) is updated to match.
- [x] `make prose` and `make dead-code-allows` stay green.
- [x] The chosen mechanism does not change `Isolation::SharedTree`'s default behaviour for existing file-tool claims — only extends coverage to `bash`.

macanderson:
## What and why

Claim-on-first-write covered the file tools and missed the shell. `sed -i`, a
redirect, a formatter run, or an `npm install` rewriting a lock file all
mutated the tree with nothing claimed, so a sibling working in the same tree
got no conflict and no holder name — while the mechanism's name suggested it
was covered.

Following the maintainer's design pointer on #5594: **claim on observed
write**, not on guessed paths. A command's writes cannot be read out of its
arguments before it runs, and parsing a command line to guess them would
refuse work that was safe. They can be read off the tree exactly, once it has
run. So `ClaimTap` snapshots `git status --porcelain -z` either side of a
`bash` call and claims every path that changed, for the calling session,
exactly as `edit_file` would have.

It never refuses: by the time anything is known the command has already run.
What it buys is the warning the **next** writer gets. A path a live rival
already holds stays that rival's — first claim wins.

`bash` is deliberately **not** put in the `BUILD_CLAIM` lane; that would
serialise every shell call in the workspace.

### Where it is armed

Only where writes land in the tree the coordination store guards: the deck's
lead turn, its sub-session worker lanes, and a fleet worker under
`Isolation::SharedTree` (`ShellWatch::for_attempt` is the one place that
decision is written). An isolated worktree has one writer and paths of its
own; claiming those in the shared table would refuse a sibling an edit to a
file it cannot see.

### Declared gaps, named in `claims.rs`'s module doc

- A workspace that is not a git repository takes no shell claim, and neither
  does one where `git` is absent or its index is locked by a concurrent
  command. The snapshot fails, nothing is claimed, the command runs.
- Git's own view is the resolution: an ignored path is invisible, and a write
  changing neither a file's size nor its modification time reads as no write.
- A command that only moves a path in or out of the dirty set — a commit —
  claims it too. Over-claiming costs a rival a wait; under-claiming costs it
  an overwritten edit.

The module doc's old "Coverage note" said no built-in was matched at all,
which was already wrong for `write_file`/`edit_file`/`delete_file` and would
have been doubly wrong after this. It is replaced by a "What a claim is taken
on" section describing both routes. Refs #5593, which owns that correction and
its regression test; this PR does not close it.

### Cost, measured

`git status --porcelain -z` on this repository (2909 tracked files, warm
cache, macOS/APFS, 7 runs): 15.0–16.2 ms, **median 15.4 ms**. A shell call
therefore pays about 31 ms for the pair. Honest caveat: this worktree carries
no `target/` directory, so a tree with build artifacts costs more and that
number has not been taken — #5676 tracks skipping the pair entirely while the
workspace has a single live writer.

## The witness and why it fails on the old code

`a_shell_write_is_claimed_and_the_next_writer_is_refused_by_name`
(`crates/stella-cli/src/claims.rs`): session A runs `bash` whose call
arguments name no path; the scripted tree reports `src/x.rs` clean before and
` M` after. The test asserts `file_locks` names `ses-1/lead`, then that
session B's `edit_file` on `src/x.rs` is refused naming A.

On `main` the mechanism it exercises does not exist: `execute` has no arm for
`bash`, so no snapshot is taken and `acquire_file_lock` is never called for a
shell write. `store.file_lock_holder("src/x.rs")` is `None` and the first
assertion fails; B's write then succeeds, which is the defect. The test cannot
pass without the observed-write route, because nothing else in the tap ever
writes a row for a tool whose input carries no `path`.

Alongside it:

- `a_shell_write_onto_a_rivals_claim_is_never_refused` — the false-refusal
  direction the pointer asked for, written even though the design makes it
  structurally impossible.
- `a_shell_command_that_writes_nothing_claims_nothing` — the negative: `ls`
  over an already-dirty tree claims nothing.
- `a_second_write_to_an_already_dirty_path_is_still_claimed` — status letters
  alone read ` M` on both sides, so the file's size and modification time are
  read beside them.
- `a_workspace_git_cannot_read_claims_nothing_and_still_runs_the_command` —
  the non-git gap.
- `an_unwatched_shell_claims_nothing` — an isolated worktree is unchanged.
- `an_absolute_tool_path_meets_the_shells_repo_relative_claim` — both routes
  key one file the same way.
- `the_z_status_form_is_parsed_including_both_ends_of_a_rename` — pinned to
  the exact bytes `git status --porcelain -z` printed on this repository for a
  staged rename: `R  README2.md\0README.md\0` (new before old, which is what
  `-z` does to the arrow form).

## Exemplar followed

`crates/stella-cli/src/fleet_commits.rs`'s `CommitObserver` — the same shape
one seam over: a decorator that reads git either side of a tool call, routes
through `stella_fleet::GitCli` rather than spawning a subprocess itself, and
degrades to recording nothing when git will not answer. Its scripted-git test
fake is the model for `ScriptedGit` here.

## Scope notes

- No new dependencies. `stella-fleet` was already a `stella-cli` dependency
  and already supplies the `GitCli` port.
- New logic lands in `claims.rs` (1176 lines) — no god file grows. The
  conditional in `fleet_cmd.rs` is four lines because that file sits at
  1498/1500; the reasoning behind it lives in `ShellWatch::for_attempt`.
- `Isolation::SharedTree`'s existing file-tool claims are untouched: the
  declared route's behaviour is identical apart from `claim_key`, which now
  strips a `./` prefix and the watched root from an absolute path so the two
  routes cannot open two claims on one file.

## Verification

`make guards-fast` green locally (prose, file-size, line-citations, fmt and
the rest). Nothing was compiled or tested on the maintainer's machine, per
CLAUDE.md — CI is the evidence.

Tests deleted: None.

Residue filed: #5674 (nested workspace root keys the two routes differently),
#5675 (a shell write over a rival's claim is noticed and not reported), #5676
(the two `git status` runs are spent even with one writer in the workspace).

Closes #5594

## Summary by Sourcery

Extend claim-on-first-write to observe and claim paths changed by shell commands in shared workspaces.

New Features:
- Claim paths modified by shell commands after execution so subsequent writers receive the same conflict protection as file tools.

Bug Fixes:
- Prevent shell writes in shared trees from bypassing file-claim tracking and overwriting concurrent work without a holder warning.

Enhancements:
- Arm shell-write observation only for shared-tree execution contexts while leaving isolated worktrees untracked.
- Treat shell observation as best-effort, preserving command execution when Git status is unavailable and retaining the first existing claim on conflicts.
- Normalize shell and tool paths so both routes claim the same workspace file.

Documentation:
- Document shell commands as write operations covered by parallel-workspace claim tracking and note the non-Git workspace limitation.

Tests:
- Add coverage for shell writes, read-only commands, dirty-path rewrites, renames, deleted or committed paths, rival claims, unavailable Git, isolated worktrees, path normalization, and porcelain status parsing.


macanderson:
Automated PR sweep — `guard self-tests` is fixed here; `dod` is the only red left, and part of it needs you.

**`guard self-tests` is green on this head.** It was red on every open PR in the repo, not just this one: `scripts/test-releases-published-pagination.sh` pinned its fail-side to `origin/main`, so it inverted the moment its own fix merged. Repaired by #5688 (`3aea7c4`); this branch picked it up when I merged main in. That failure was never this PR's.

**`dod / dod` remains red on #5594's nine unchecked items.** I did not tick any of them, and the reason splits cleanly in two.

**Four are decisions, and they are yours.** Items 1–3 turn on a choice this sweep cannot make for you — whether an unenumerable shell mutation gets a mechanism or a declared gap — and item 3's wording (`claims.rs`'s module doc names the gap) only applies if you pick the second. Recording a decision on someone else's behalf is how a DoD gate stops meaning anything, so these stay open.

**Five look mechanically settleable from checks already green on this head**, if you want them ticked without re-running anything locally:

| item | what would settle it |
|---|---|
| `cargo test -p stella-cli claims` is green | `fmt + clippy + test` — currently running on this head; it runs the workspace suite |
| `make file-size` stays green | `file size ratchet` ✅ |
| `make prose` stays green | `docs guards` ✅ |
| `make dead-code-allows` stays green | covered by the guard job, ✅ |
| does not change `Isolation::SharedTree`'s default | readable off the diff |

I have deliberately not ticked those either — four of them hang off a `fmt + clippy + test` that has not finished, and the fifth is a claim about the diff that pairs with the design decision above. Ticking half a list while its governing decision is open would make `dod` green on a question nobody answered.

**One process note, since this is the fourth instance today.** #5629, #5244, #5681 and #5589 all carried DoD items that no pre-merge state could satisfy ("green on `main`" for a job that only runs post-merge). #5594's list does not have that defect — its items are all verifiable on the head under review, which is the right shape. Noting it because the contrast is the useful half.

---

macanderson:
Two notes on the DoD ticks, so the evidence behind each is on the record and
nobody has to take a checkbox on faith.

**"the module doc names the gap and cites this issue".** That box is the
either/or partner of the one above it: a mechanism was chosen, so the branch
it describes did not apply in full. What ships satisfies its substance —
`claims.rs`'s module doc names the non-git gap the design pointer asked for,
plus the two others the mechanism has (git's own resolution, and the two
`git status` runs per shell call). It does not carry the issue number,
because `make prose`'s `issue-reference` ratchet refuses any new `#NNNN` in a
comment and the guard has no exemption for a doc that a DoD item asks to cite.
Two repo rules genuinely disagree there; I obeyed the enforced one and am
saying so rather than deciding it quietly. If a maintainer wants the citation
in the doc, it needs a `prose-baseline` decision, not a code change.

**"does not change `Isolation::SharedTree`'s default behaviour for existing
file-tool claims".** One qualification, already in the PR body under Scope
notes: `ClaimTap::claim_key` now strips a leading `./` and the watched root
from a tool's path before claiming it. For a path spelled the way every
existing test and caller spells it, the key is byte-identical and behaviour is
unchanged; what changes is that `/repo/src/x.rs` and `src/x.rs` stop being two
claims on one file, which is the coverage this PR exists to extend rather than
a change to the declared route's semantics.

Evidence for the rest: `fmt + clippy + test` green on 31a5567 — the workspace
test run includes `claims::tests::a_shell_write_is_claimed_and_the_next_writer_is_refused_by_name`,
`a_shell_write_onto_a_rivals_claim_is_never_refused`,
`a_second_write_to_an_already_dirty_path_is_still_claimed` and the rest of the
new cases, all `ok`. `file size ratchet`, `docs guards` and the guards job
carry `make prose`, `make file-size` and `make dead-code-allows`.

`guard self-tests` was red on an earlier head for a reason unrelated to this
diff — `test-releases-published-pagination.sh` pinned its fail-side to
`origin/main`, which failed every open PR (#5671, #5681). #5688 landed and this
branch picked it up in the merge from main; the job is green now.

Issue #5589, opened 2026-08-31
Version-sync pull requests get stuck waiting for someone to click approve


macanderson:
**In plain words:** every automatic release pull request opens with its checks frozen. GitHub is waiting for a human to click "approve" on 13 workflow runs before any of them will even start. Nobody is supposed to be in this loop — the whole point of the version-sync bot is that it runs without a person watching.

Parent epic: #5313 ("the CI and test signal is trustworthy — no green over a real failure, no red on a real pass"). This is the same failure class one step earlier: here the checks never even *start*, so there is no signal at all, not even a wrong one.

## What happens

Every release version-sync pull request opens with all of its workflow runs parked in `action_required`. They do not queue and they do not fail — they sit. Until somebody approves them, the pull request carries two checks (`dod` and `Sourcery review`) and none of the three that `main` actually requires.

Because the required contexts never report, the pull request cannot merge, and nothing on it says why. It reads as "checks still coming."

## Seen three times in about ninety minutes

On 2026-08-31: #5571 (0.9.295), #5576 (0.9.296), #5588 (0.9.299). Each opened with thirteen runs parked — `ci`, `msrv`, `file-size`, `bench`, `guard-self-tests`, `rebase-replay`, `main-red-hold`, `windows-check`, `docker-serve`, `duplicate-claims`, `empty-diff`, `dependency-review`, `dod-check`.

Approving them by hand cleared it each time, and the checks then went green with no other change:

```sh
h=$(gh pr view <PR> --json headRefOid --jq .headRefOid)
gh api "repos/macanderson/stella/actions/runs?head_sha=$h" \
  --jq '.workflow_runs[] | select(.conclusion=="action_required") | .id' \
| while read -r id; do gh api -X POST "repos/macanderson/stella/actions/runs/$id/approve"; done
```

## A second papercut, same owner

Once the runs are approved and checks report, the pull request still fails `dod`:

```
SCR-003 — this PR cannot close its issue yet.
- This PR links no issue. Add a closing reference (`Closes #123`) to the description,
  or apply the `no-issue` label if the change is genuinely trivial (SCR-003).
```

A mechanical version bump is exactly the `no-issue` case the gate names. The bot doesn't apply that label at creation time, so every one of these needs a *second* manual touch after the first (a human applied `no-issue` by hand to #5588 to clear it).

So today a version-sync pull request needs two human actions before it can merge: approve thirteen parked workflow runs, then add `no-issue`. Neither is a judgement call — both should be automatic.

[...]

PR #5680, opened 2026-09-03
fix(release): record the four orphan tags and stop discarding built releases

### LINKED ISSUE #5244: Tag v0.9.194 was cut but never published, and main is red on it
### Context

`main` is red on the `every tag has a published release` check
(`scripts/check-releases-published.sh`):

```
1 tag(s) were cut but never published. The repository is stamping version
numbers while shipping no binaries — this is the #1464 silence, live.

Tag                  Age
  v0.9.194           51h
```

`v0.9.194` exists as a tag (`ec6e0c9f4`) and
`GET /repos/macanderson/stella/releases/tags/v0.9.194` returns **404** — no
release, draft or otherwise.

### What actually happened

The release run for that tag is
[32789294637](https://github.com/macanderson/stella/actions/runs/32789294637)
(`workflow_dispatch`, 2026-08-24T23:25:46Z, attempt 1), and it did **not** fail
on a build error:

- All four `build …` jobs report `conclusion: failure`, but every one of their
  **steps** is `success` — including `Build stella (reproducible)`,
  `Package tarball` and `Upload build artifact`. `--log-failed` returns nothing,
  because no step failed.
- All four completed at the **identical** timestamp, `2026-08-25T00:20:46Z`,
  having started at `23:25:50` — killed together at 54m56s.

Four independent jobs dying on the same second, with every step green, is a
run-level kill (timeout or cancellation), not four build failures. The
downstream `publish GitHub Release` and `publish Homebrew formula` jobs were
`skipped` in consequence, which is why the tag has no release.

Every release since has succeeded — v0.9.243 through v0.9.253 are all green, so
this is one stranded tag rather than a broken release path.

### The decision this issue needs

Two remedies, and they are not equivalent. Both are outward-facing, which is why
this is filed rather than done:

1. **Re-run the release for `v0.9.194`.** Restores the invariant the check
   asserts. But it publishes a 51-hour-old version *after* ~60 newer ones,
   including a Homebrew formula write — consider whether a release appearing out
   of order is worse than the gap.
2. **Delete the tag as an aborted release.** Honest if nothing was ever shipped
   from it and nothing references it, and it clears the check without publishing
   stale binaries. Verify first that no changelog entry, download URL or
   installer pin names v0.9.194.

## Definition of done

- [x] A decision is recorded here on which remedy applies, and why. **Remedy 3, neither of the two listed: the tag stays and stays unpublished, recorded in `scripts/unpublished-tags-baseline.txt` with its cause (PR #5680). Re-running `release.yml` on an old tag runs the workflow as it stood at that tag, whose `homebrew` job would render an older formula over the tap; the run's four artifacts expired on 2026-08-31, so a re-run means a full rebuild of a version a hundred releases behind. Deleting the tag was rejected because the tag is the record that the version was cut.**
- [x] `./scripts/check-releases-published.sh --grace-secs 5400` exits 0 on the
      head that carries the fix — the identical tree `main` receives on merge.
- [x] If the cause was a run-level timeout, the release workflow's timeout is
      either raised or the reason it took >54m is understood — otherwise the next
      release near that boundary strands the same way. **It was not a timeout:
      the jobs started at 23:25:50 and the run was concluded at 00:20:46 —
      54m56s — against `timeout-minutes: 90` on the `build` job, so the ceiling
      was never met and raising it would fix nothing. What did conclude the run
      is not reachable from the API, and is split into #5699.**
- [x] Residue filed as new issues (`triage` label only, SCR-004). **#5698 (a 5xx from the release-publish API still ends the run with the build finished) and #5699 (why run 32789294637 was killed). #5681 / #5689 already covered the third finding, so it was linked rather than duplicated.**

### Files

- `scripts/check-releases-published.sh` — the check and its `--grace-secs`.
- `.github/workflows/release.yml` — the run that was killed; look for a
  job/workflow `timeout-minutes` near 55.
- `RELEASING.md` § *When a release fails* — the documented re-run path.

### Verify

```
gh api repos/macanderson/stella/releases/tags/v0.9.194   # 404 today
./scripts/check-releases-published.sh --grace-secs 5400  # exit 1 today
```

### Constraints

- Do not clear this by widening `--grace-secs`: the tag is 51h old and the grace
  window is 90 minutes, so any widening that hides it would hide a real gap for
  two days.
- #1464 is the precedent the check's own error message cites — a monitor that
  goes silent is worse than none, so the check going green must mean the tag was
  resolved, not that the check stopped asking.

## Verification at `c5cc15c` (PR #5680)

Ticked by an automated PR sweep, against the **live** GitHub API rather than a
fixture: a curl-backed `gh` shim served the real release list (713 releases
over 8 pages) to both copies of the script, with the repository's real 742 `v*`
tags read from git.

**Before**, `main`'s copy — exit 1:

```
::error::4 tag(s) were cut but never published. The repository is stamping
version numbers while shipping no binaries — this is the #1464 silence, live.

Tag                  Age
  v0.9.194           217h
  v0.9.254           165h
  v0.9.264           159h
  v0.9.273           136h
```

**After**, #5680's copy on the same live data — exit 0:

```
check-releases-published: OK — every v* tag older than 90m has a published
release (742 tag(s) checked, 28 grandfathered).
```

The four tags are genuinely unpublished, not a reporting artifact:
`GET /releases/tags/{tag}` answers `404` for each of `v0.9.194`, `v0.9.254`,
`v0.9.264`, `v0.9.273`. #5680 grandfathers exactly those four in
`scripts/unpublished-tags-baseline.txt`, each with a note naming the run that
lost it, which is what that file asks for.

### On the wording

As written this item read *"exits 0 on `main`"*, which no pre-merge state can
satisfy — the fix is not on `main` until this merges, so the box could never be
ticked and `dod` would stay red forever. That is the same bootstrap shape
already corrected on #5681 and flagged on #5589/#5672; the rewording above
follows that precedent and is recorded here rather than done silently. What was
verified is the identical tree `main` receives. If it is somehow red on `main`
afterwards, that is a new finding and a new issue, not this box.

---

### LINKED ISSUE #5629: release-reconcile is red: 4 tags cut but never published — a transient upload-artifact ENOTFOUND discards a whole release build

macanderson:
## What this changes

Four tags cut between 2026-08-24 and 2026-08-28 shipped no binaries, and
`release-reconcile` has been red on `main` ever since. They failed three
different ways, not one:

| Tag | Failure | Evidence |
|---|---|---|
| `v0.9.194` | Run-level kill. All four build jobs uploaded their artifacts and wrote clean logs, then were marked failed at one identical second 48 minutes later. | [run 32789294637](https://github.com/macanderson/stella/actions/runs/32789294637) |
| `v0.9.254` | The publish step created the release and uploaded assets, then GitHub answered a 5xx. Draft 377549233 was left behind with no `SHA256SUMS`. | [job 98400786154](https://github.com/macanderson/stella/actions/runs/33035688879/job/98400786154) |
| `v0.9.264` | Same 5xx in the same step. Draft 377720633 was left behind, missing the `aarch64-apple-darwin` tarball. | [job 98479799232](https://github.com/macanderson/stella/actions/runs/33059553973/job/98479799232) |
| `v0.9.273` | `actions/upload-artifact` failed with `ENOTFOUND` after a seven-minute build had already packaged its tarball. | [run 33157096458](https://github.com/macanderson/stella/actions/runs/33157096458) |

I confirmed both drafts' asset lists directly (`gh api
repos/macanderson/stella/releases/<id> --jq '.assets[].name'`): five assets
each, and each set is missing a different required file. Neither was
publishable as it stood.

**The four tags stay unpublished**, per the maintainer's design note on the
issue. Re-running `release.yml` on an old tag runs the workflow *as it stood at
that tag*, whose `homebrew` job renders the formula for that version
unconditionally and would downgrade the tap. `v0.9.194`'s artifacts expired on
2026-08-31, so a rebuild is the only path there, and rebuilding a version a
hundred releases behind buys nothing. Each tag is recorded in
`scripts/unpublished-tags-baseline.txt` with a comment line naming its cause
and run id.

The two incomplete drafts are deleted (release ids 377549233 and 377720633),
noted on the issue.

## Why the path stops discarding built releases

Three changes in `.github/workflows/release.yml`, one per failure shape:

- **`upload-artifact` runs twice.** The first step carries
  `continue-on-error: true` and an `id`; the second runs on
  `steps.upload.outcome == 'failure'` with `overwrite: true`. No third-party
  retry action, so `action-pins` and the licence allowlist are untouched, and a
  genuine upload failure still fails the job because the retry step is plain.
- **The `release` job deletes a stale draft for its own tag** before
  `softprops/action-gh-release` runs. A draft is not attached to its tag, so
  the action's by-tag lookup 404s and it creates a *second* release beside the
  first — which is exactly how two half-built drafts accumulated. Only a draft
  is ever deleted; a published release for the tag is never touched.
- **The `homebrew` job clones the tap first** and refuses to render a version
  lower than the formula already there (`sort -V` comparison, `::notice::` and
  exit 0). The formula is the one artifact here that is a pointer rather than
  an addition: a stale GitHub Release sits harmlessly beside newer ones, a
  stale `Formula/stella.rb` is what everybody installs.

## Draft versus absent

`scripts/check-releases-published.sh` reported all four as "cut but never
published", which sent readers looking for four rebuilds when two already had
their assets built. It now reports a state per tag, read from `GET /releases`:

```
Tag                  Age    State
  v0.9.254           126h   draft
  v0.9.273           97h    absent
```

`GET /releases/tags/{tag}` cannot answer this — it returns 404 for a draft too,
because a draft is not attached to its tag. That 404 is consistent with both
answers, which is why the issue's own "the releases are genuinely absent, not
drafts" was wrong for half the set. The script's header says so, and the report
prints the remedy for each state.

The baseline's `--update` also keeps the comment line above each tag now.
Without that, the next regeneration would erase the reason every entry exists,
which is the only thing that makes an entry reviewable.

## The witness, and why it fails on `main`

`scripts/test-releases-published.sh` gains nine cases in two groups. Run
`make releases-published-test`.

**D3/D4/D5** read the *third* field of a `--select` line. On `main` that line
carries only `<tag>\t<age>`, so `cut -f3` returns the empty string and all
three fail with `wanted 'draft', got ''`. Verified against `main`'s copy of the
script directly:

```
$ printf '{"tags":[{"name":"v0.6.80","created_unix":568000}],
           "releases":[{"name":"v0.6.80","draft":true}]}' \
    | <origin/main copy> --select --now 1000000 --grace-secs 5400 | cut -f3
(empty)
```

**B1–B5** drive the note-carrying rule through a new pure mode
(`--carry-notes --baseline PATH`) against temp fixtures. On `main` that mode
does not exist: the argument parser answers `unknown argument: --carry-notes`
and exits 2, so every case fails. Verified the same way.

Both groups pass on this branch — all 25 cases green locally
(`./scripts/test-releases-published.sh`).

The workflow changes carry no witness test: a `run:` block cannot be exercised
without a release, and the DoD asked for no new script. `release-reconcile`
going green on `main` after merge is their fail→pass flip, and I will trigger
it and cite the run before ticking that box.

## Exemplar followed

`--carry-notes` copies `--select`'s own shape in the same file: a pure mode
with the I/O taken out, so a fixture is a complete input and no live repository
or API is involved. That is what makes the reconciliation rule testable today,
and it is why the note-carrying awk is not shipped untested.

## Evidence

- `make guards-fast` green, including `prose`, `line-citations`, `shellcheck`,
  `action-pins` and `file-size`.
- `./scripts/test-releases-published.sh` — passed 25, failed 0.
- Draft asset lists read from the API before deleting either one.

Tests deleted: None

Closes #5629
Closes #5244

## Summary by Sourcery

Harden release recovery so transient failures do not discard built artifacts or downgrade downstream packaging, while recording and accurately reporting unreleased tags.

New Features:
- Distinguish draft and absent releases in reconciliation reports and preserve explanatory notes when updating the unpublished-tag baseline.

Bug Fixes:
- Prevent transient artifact-upload failures and stale draft releases from discarding otherwise completed release builds.
- Prevent release retries from overwriting a newer Homebrew formula with an older version.

Enhancements:
- Document recovery guidance for draft, absent, and permanently unpublished releases.

Documentation:
- Document release failure states, draft asset validation, and safeguards for rerunning older tags.

Tests:
- Add coverage for draft-versus-absent reporting and baseline note preservation.

Chores:
- Record the four historically unpublished tags and their failure reasons in the baseline.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -158,7 +158,21 @@ jobs:
           echo "asset=${stem}.tar.gz" >> "$GITHUB_OUTPUT"
           echo "sums=${stem}.sha256" >> "$GITHUB_OUTPUT"
 
+      # Uploaded twice, because one unlucky DNS lookup used to throw away a
+      # finished release. v0.9.273 compiled for seven minutes, packaged its
+      # tarball, and then died here on `Failed to CreateArtifact: Unable to
+      # make request: ENOTFOUND` — an Actions endpoint the runner could not
+      # resolve. Every downstream job needs all four artifacts, so that one
+      # step took the whole tag with it (#5629).
+      #
+      # `continue-on-error` on the first step is what lets a second one run;
+      # the retry itself is a plain step, so a real upload failure still fails
+      # the job. No retry action is added: a third-party action would owe
+      # `action-pins` a pinned SHA and the licence allowlist a review, to buy
+      # one `if:` condition.
       - name: Upload build artifact
+        id: upload
+        continue-on-error: true
         uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7
         with:
           name: stella-${{ matrix.target }}
@@ -167,6 +181,21 @@ jobs:
             ${{ steps.package.outputs.sums }}
           if-no-files-found: error
           retention-days: 7
+
+      # `overwrite` covers the narrower failure where the first attempt got far
+      # enough to reserve the name before it broke; without it the retry would
+      # fail on a name conflict of its own making.
+      - name: Upload build artifact (retry)
+        if: steps.upload.outcome == 'failure'
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7
+        with:
+          name: stella-${{ matrix.target }}
+          path: |
+            ${{ steps.package.outputs.asset }}
+            ${{ steps.package.outputs.sums }}
+          if-no-files-found: error
+          retention-days: 7
+          overwrite: true
 
   # Rebuild one target on a second runner and refuse to publish if the bytes
   # differ (#910, acceptance clause). This is the check that keeps
@@ -444,6 +473,37 @@ jobs:
           } > notes.md
           echo "----- notes.md -----"; cat notes.md
 
+      # A draft release is not attached to its tag, so `GET /releases/tags/{tag}`
+      # answers 404 for one and the publish step below creates a second release
+      # instead of finishing the first. That is how v0.9.254 and v0.9.264 ended
+      # up as drafts nothing would ever touch again: the step created the
+      # release, uploaded its assets, then hit a GitHub 5xx before publishing,
+      # and a re-run would have left the half-built draft sitting beside a new
+      # one (#5629). Deleting it first is what makes a re-run finish the job.
+      #
+      # Only a draft is ever deleted. A published release for this tag is a
+      # release users can already download, and this step must never touch it.
+      - name: Delete a stale draft release for this tag
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        shell: bash
+        run: |
+          set -euo pipefail
+          tag="${GITHUB_REF_NAME}"
+          # `--slurp` cannot be combined with `--jq` (gh rejects that), so the
+          # filter runs in a plain jq after the fetch.
+          ids="$(CLICOLOR_FORCE=0 NO_COLOR=1 gh api --paginate --slurp \
+            'repos/{owner}/{repo}/releases?per_page=100' \
+            | jq -r --arg tag "$tag" '.[][] | select(.draft == true and .tag_name == $tag) | .id')"
+          if [ -z "$ids" ]; then
+            echo "no draft release for ${tag}; nothing to clear."
+            exit 0
+          fi
+          for id in $ids; do
+            echo "::notice::deleting stale draft release ${id} for ${tag} so this run can publish."
+            gh api -X DELETE "repos/{owner}/{repo}/releases/${id}"
+          done
+
       - name: Create / update GitHub Release
         uses: softprops/action-gh-release@3d0d9888cb7fd7b750713d6e236d1fcb99157228 # v3
         with:
@@ -494,6 +554,41 @@ jobs:
 
           version="${GITHUB_REF_NAME#v}"
 
+          if [ -n "${HOMEBREW_TAP_DEPLOY_KEY:-}" ]; then
+            # SSH with the tap-scoped deploy key. The key file lives outside
+            # the workspace and is chmod 600, as ssh requires.
+            keyfile="${RUNNER_TEMP}/tap_deploy_key"
+            printf '%s\n' "${HOMEBREW_TAP_DEPLOY_KEY}" > "${keyfile}"
+            chmod 600 "${keyfile}"
+            export GIT_SSH_COMMAND="ssh -i ${keyfile} -o IdentitiesOnly=yes -o StrictHostKeyChecking=accept-new"
+            git clone --depth 1 "git@github.com:macanderson/homebrew-tap.git" tap
+          else
+            git clone --depth 1 \
+              "https://x-access-token:${HOMEBREW_TAP_TOKEN}@github.com/macanderson/homebrew-tap.git" tap
+          fi
+
+          # The tap is cloned before anything is rendered so this job can read
+          # the version it currently serves and refuse to go backwards.
+          #
+          # Re-running release.yml on an old tag checks out the workflow AS OF
+          # THAT TAG and renders the formula for that version unconditionally,
+          # so retrying a months-old failed release would hand every `brew
+          # upgrade` an older stella than the one it has (#5629). The formula
+          # is the one artifact here that is a pointer rather than an addition:
+          # a stale GitHub Release is harmless beside the newer ones, and a
+          # stale Formula/stella.rb is what everybody installs.
+          tap_version=""
+          if [ -f tap/Formula/stella.rb ]; then
+            tap_version="$(sed -n 's/^[[:space:]]*version "\([^"]*\)".*/\1/p' tap/Formula/stella.rb | head -n1)"
+          fi
+          if [ -n "${tap_version}" ] && [ "${tap_version}" != "${version}" ]; then
+            older="$(printf '%s\n%s\n' "${version}" "${tap_version}" | sort -V | head -n1)"
+            if [ "${older}" = "${version}" ]; then
+              echo "::notice::the tap serves ${tap_version}; not rendering the older ${version} over it."
+              exit 0
+            fi
+          fi
+
           # SHA-256 of a given target's release tarball (same bytes uploaded to
           # the GitHub Release in the `release` job).
           sha_for() {
@@ -522,19 +617,6 @@ jobs:
 
           echo "----- rendered Formula/stella.rb -----"
           cat rendered/stella.rb
-
-          if [ -n "${HOMEBREW_TAP_DEPLOY_KEY:-}" ]; then
-            # SSH with the tap-scoped deploy key. The key file lives outside
-            # the workspace and is chmod 600, as ssh requires.
-            keyfile="${RUNNER_TEMP}/tap_deploy_key"
-            printf '%s\n' "${HOMEBREW_TAP_DEPLOY_KEY}" > "${keyfile}"
-            chmod 600 "${keyfile}"
-            export GIT_SSH_COMMAND="ssh -i ${keyfile} -o IdentitiesOnly=yes -o StrictHostKeyChecking=accept-new"
-            git clone --depth 1 "git@github.com:macanderson/homebrew-tap.git" tap
-          else
-            git clone --depth 1 \
-              "https://x-access-token:${HOMEBREW_TAP_TOKEN}@github.com/macanderson/homebrew-tap.git" tap
-          fi
 
           mkdir -p tap/Formula
           cp rendered/stella.rb tap/Formula/stella.rb
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -237,6 +237,41 @@ diff:
 make releases-baseline-update    # then read the diff
 ```
 
+Write the reason above the tag as a comment line. Regeneration keeps those
+notes, and the reason is what a reviewer needs to judge the entry.
+
+### The three shapes a failed release takes
+
+`make releases-published` names the state of every reported tag, because the
+work each one needs is different. Four orphan tags once turned out to be three
+different failures:
+
+| State | What happened | What to do |
+|---|---|---|
+| `draft` | The build ran and the assets are attached, but a `5xx` killed the publish step before the release left draft. | List the draft's assets. Publish it if the set is complete; delete it and re-run the tag if it is not. |
+| `absent` | No release object exists — a build job died, often in `actions/upload-artifact`. | Re-run `release.yml` on the tag, but read the version guard below first. |
+| gone | The build finished, and the run itself was killed after the fact. Build artifacts expire after 7 days, so an old one has nothing left to publish. | Grandfather the tag with a note. Rebuilding a version a hundred releases behind buys nothing. |
+
+Find a draft and see what it holds:
+
+```bash
+gh api repos/{owner}/{repo}/releases \
+  --jq '.[] | select(.draft) | {id, tag_name, assets: [.assets[].name]}'
+```
+
+A complete set is four tarballs plus `SHA256SUMS` and `SHA256SUMS.bin`.
+`install.sh` and the Homebrew formula both read `SHA256SUMS`, so a draft
+missing it is not publishable — delete it (`gh api -X DELETE
+repos/{owner}/{repo}/releases/<id>`) rather than shipping a release whose
+checksum file is absent.
+
+**Re-running an old tag runs the workflow as it stood at that tag.** That
+matters most for the `homebrew` job, which renders `Formula/stella.rb` for
+whatever version it is building. Today's job refuses to render a version lower
+than the one the tap already serves, so a re-run cannot downgrade `brew
+upgrade`. A tag old enough to predate that guard has no such protection: check
+what its own `release.yml` does before dispatching it.
+
 ## Verify a published binary yourself
 
 Release builds are reproducible: the same tag, built on the pinned toolchain,
--- a/scripts/check-releases-published.sh
+++ b/scripts/check-releases-published.sh
@@ -4,6 +4,7 @@
 #
 #   ./scripts/check-releases-published.sh [--grace-secs N]
 #   ./scripts/check-releases-published.sh --select --now <unix> --grace-secs <n>   # pure: JSON on stdin
+#   ./scripts/check-releases-published.sh --carry-notes [--baseline PATH]          # pure: tags on stdin
 # --help text ends here.
 #
 # ── The silence ──────────────────────────────────────────────────────────────
@@ -41,6 +42,22 @@
 # is skipped. Too short and this cries wolf on every release; too long and the
 # silence it exists to break lasts that much longer.
 #
+# ── Draft or absent: two findings with two remedies ──────────────────────────
+#
+# A tag with no published release is in one of two states, and they cost
+# different things to fix:
+#
+#   draft    the build ran and its assets are attached to a release object
+#            that was never flipped out of draft. Nothing has to be rebuilt.
+#            Check the asset list, then publish or delete the draft.
+#   absent   there is no release object at all. The tag has to be built again.
+#
+# Telling them apart needs `GET /releases`, which lists drafts. The by-tag
+# endpoint cannot answer it: `GET /releases/tags/{tag}` returns 404 for a draft
+# too, because a draft is not attached to its tag. That 404 is consistent with
+# both states, so reading it as "absent" is a guess. Four orphan tags were read
+# that way once, and two of them already had their assets built.
+#
 # ── Why the release list has no fixed cap ────────────────────────────────────
 #
 # A fixed `--limit` is a trap that reopens itself. A cap of 200 once reported
@@ -60,6 +77,7 @@ set -euo pipefail
 
 grace_secs=$((90 * 60))
 select_only=0
+carry_only=0
 update=0
 now=""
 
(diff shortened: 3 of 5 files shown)


macanderson:
**Answering the two ❌ rows in the Sourcery assessment.**

### #5244 — "Understand or prevent the run-level timeout/cancellation that stranded `v0.9.194`, and file the required follow-up residue issues"

Half of this is out of scope and half is now filed.

**On the timeout.** Sourcery reads the run-level kill as a timeout and asks
that the timeout be raised. The numbers say it was not one: the build jobs
started at 23:25:50 and the run was concluded at 00:20:46 — 54m56s — against
`timeout-minutes: 90` on the `build` job. Raising 90 would fix nothing and
would be a guess written as a remedy, which is the move AGENTS.md's evidence
rule exists to stop.

**On the cause.** I could not establish it, and I say so in the PR body rather
than around it. `gh run view`, `gh api .../jobs` and `--log-failed` carry no
error for any of the four jobs, no cancellation actor, and a bare `failure` on
the run. What is left to read is the raw log archive, the run's annotations, or
GitHub support. That is filed as **#5699**, with the evidence, the three places
left to look, and the note that this shape recurs past both fixes in this PR.

**On residue.** Filed: **#5698** (a 5xx from the release-publish API still ends
the run with the build finished — this PR makes a *re-run* idempotent, it does
not make the *first* run survive) and **#5699** above. The third thing I found —
`test-releases-published-pagination.sh` reading its fail-side off `origin/main`,
which reddens `guard self-tests` for every open PR right now — was already filed
as #5681 / #5689 with three PRs open against it, so I linked rather than
duplicated.

### #5629 — "Establish the root cause of each failed release and complete the required verification, including a demonstrated fail-to-pass `release-reconcile` run"

**Root causes.** Three of four are established from their own logs and named in
the PR table. `v0.9.194` is not, and is #5699 — stated as unestablished rather
than assumed.

**The `release-reconcile` flip.** Now demonstrated, before the merge:

- **Before** — [run 33522107716](https://github.com/macanderson/stella/actions/runs/33522107716)
  on `main`, `every tag has a published release` red on all four tags.
- **After** — [run 33700728756](https://github.com/macanderson/stella/actions/runs/33700728756),
  the same workflow dispatched on this branch, both jobs green.

That run is the real check against the live repository — it reads the actual
tag list and the actual release list, with this branch's baseline. I will
dispatch it once more on `main` after the merge and cite that run on the issue
before ticking the box that names `main` specifically.

The `guard self-tests` red on this PR is #5681, not this diff: the pagination
witness compares against `origin/main`, which now carries the fix it was
written to prove missing. Its two `NEW` cases pass here; the `MAIN` case fails
identically on every open PR in the repository.

macanderson:
Automated PR sweep — the `dod` blocker is cleared, and one process finding worth more than this PR.

**All three outstanding DoD items are now satisfied and ticked**, on #5629 and #5244, with the evidence recorded on each issue rather than asserted here.

### The fail→pass flip, against live data

I reproduced it rather than citing the red runs, using a curl-backed `gh` shim so both copies of the script saw the **real** release list (713 releases over 8 pages) and this repository's real 742 `v*` tags:

| | `check-releases-published.sh --grace-secs 5400` |
|---|---|
| `main` | **exit 1** — names `v0.9.194` (217h), `v0.9.254` (165h), `v0.9.264` (159h), `v0.9.273` (136h) |
| this branch | **exit 0** — `every v* tag older than 90m has a published release (742 tag(s) checked, 28 grandfathered)` |

The four tags are genuinely unpublished, not a reporting artifact: `GET /releases/tags/{tag}` answers `404` for each. The baseline entries each carry a note naming the run that lost it — `v0.9.194`'s artifacts expired, `v0.9.254` and `v0.9.264` hit a 5xx mid-publish and left drafts since deleted, `v0.9.273` lost an `upload-artifact` to `ENOTFOUND` — which is what that file asks of an addition.

### `release-reconcile` green, by a real run

Run [33700728756](https://github.com/macanderson/stella/actions/runs/33700728756) concluded `success`, both jobs green. It sits on `48cf089` while the head is now `c5cc15c`, so I checked the evidence transfers instead of assuming it: `git diff 48cf089 c5cc15c` over `check-releases-published.sh`, `unpublished-tags-baseline.txt` and `release-reconcile.yml` is **empty**, and the only commits between them are the merge of `main`.

### The finding: this is the third bootstrap DoD in a row

Two of the three items could not be satisfied by *any* pre-merge state as written — "green on `main`" for a workflow that only runs on `main` after the merge. The box can never be ticked, so `dod` stays red forever and the PR cannot land, which is the opposite of what the gate is for.

This is not the first time: #5681's fourth item was reworded for exactly this, and #5589/#5672 hit it before that. Three occurrences in one day suggests the fix belongs in how these lists get written — a DoD item should name something verifiable on the head under review — rather than in a fourth one-off rewording. I applied #5681's precedent here (reworded to *the head that carries the fix*, with the reason recorded in the issue body, not silently), but I have not filed an issue for the underlying pattern: that is a call about your own process, and SCR-004 residue filed blind on someone else's workflow is noise.

### Also on this head

`guard self-tests` is **green** here now. It was red on every open PR because `scripts/test-releases-published-pagination.sh` pinned its fail-side to `origin/main`, so it inverted the moment its own fix merged. Repaired by #5688 (`3aea7c4`); this branch picked it up in the merge at `c5cc15c`. That failure was never this PR's.

The stale red `dod / dod` on this head is the 00:49 run, which read the boxes before they were ticked; it clears on the next `synchronize` or `edited` event. I did not manufacture one — SCR-005 keeps labels with triage, and a no-op push to turn a check green is the shortcut CLAUDE.md forbids.

---

PR #5682, opened 2026-09-03
fix(workspace): unwind on panic in the release profile

### LINKED ISSUE #5493: Release builds abort on panic, so every panic boundary in the tree is inert where it ships
## The problem

The release profile sets `panic = "abort"` (root `Cargo.toml`, `[profile.release]`).
`catch_unwind` cannot catch anything under that setting, so every panic boundary
in the workspace is inert in the binaries users actually run. The code that
builds those boundaries is written, documented and tested as though it works,
and it does — but only in debug and test builds, which is where the tests run.

## Where

- Root `Cargo.toml`, `[profile.release]`: `panic = "abort"`, alongside
  `lto = true`, `codegen-units = 1`, `strip = true`.
- `crates/stella-tui/src/panel_guard.rs`, `guarded` — wraps every deck band in
  `std::panic::catch_unwind` and paints an error card on a panic, with a module
  doc explaining that "the rest of the TUI is still running". In a release build
  a band panic ends the process instead.
- `crates/stella-serve` — a panic inside a spawned connection task aborts the
  whole server: every other live turn, the listener, and every graceful drain
  path the crate builds. #5489's `stella-serve` fix removed one reachable panic
  of this kind; the profile is why that one mattered so much.

## Failure scenario

A user on a short terminal opens the SKILLS tab with a live search. In a debug
build the band paints "panel 'skills' panicked" and the rest of the deck keeps
working, which is what `panel_guard` promises. In the shipped release binary the
process aborts and the terminal is left in raw mode. (That specific panic is
fixed in #5489; the point is that the boundary meant to contain the *next* one
does not.)

## The decision this needs

This is a maintainer's call, not a mechanical fix. The options:

1. **Set `panic = "unwind"` for release.** The boundaries start working as
   documented. Costs binary size and some optimisation; `strip`/`lto` stay.
2. **Keep `abort` and delete the boundaries**, replacing them with the honest
   statement that a panic ends the process. Cheaper binary, and no code that
   claims a guarantee it does not have.
3. **Keep both and document the split** — the boundary is a development aid
   only. This is the status quo, and it is the one option that leaves the
   comments untrue.

Option 3 is the current state by accident rather than by decision, which is why
this is filed rather than fixed.

## Definition of done

- [x] A decision recorded, as an ADR if the profile changes.
      `docs/adr/0024-release-builds-unwind.md`, cited from `[profile.release]`
      itself and listed in `docs/adr/README.md` and `docs/manifest.json`.
- [x] If the profile keeps `abort`: every comment claiming a panic is contained
      says plainly that it is contained in debug builds only.
      Not applicable — the profile moved to `unwind`. The comments were swept
      the other way instead: ten of them across `stella-tui`, `stella-core`,
      `stella-serve` and `stella-cli` claimed a release panic ends the process,
      and each now says which branch a shipped build takes.
- [x] If the profile moves to `unwind`: a test that a band panic is contained in
      a build with the release panic strategy, so the guarantee is checked rather
      than assumed.
      `crates/stella-tui/examples/panic_guard_probe.rs`, run with `--release` by
      ci.yml's step "a panel panic is contained under the release profile". It
      passed on run 33701067233 and printed
      `panic_guard_probe: OK — a panel panic was contained under this profile`.

## Confidence

CONFIRMED for the mechanism — `panic = "abort"` is in the root manifest and
`catch_unwind` is documented by the standard library as unable to catch an abort.
Not measured: whether any panic boundary other than `panel_guard`'s is reachable
in a shipped binary today.

macanderson:
## What and why

The release profile set `panic = "abort"`. A catch cannot catch an abort, so
every panic boundary in the workspace did nothing in the binaries users run.
The code that builds those boundaries is written, documented and tested as
though it works — and it does, in test builds, which is where the tests run.

Two of them are promises to a person:

- A panel that panics paints an error card and the rest of the screen keeps
  working (`crates/stella-tui/src/panel_guard.rs`). Under `abort` the process
  died instead and left the terminal in raw mode.
- A panic in one `stella-serve` connection ends that connection. Under `abort`
  it ended the server, every other live turn with it.

Three more have the same shape: a child agent's spend settles on the panic path
(`crates/stella-cli/src/subagent.rs`), a worker panic costs one lane rather than
the session (`crates/stella-cli/src/subsession.rs`), and a hook that panics is
skipped rather than fatal (`crates/stella-core/src/bus.rs`).

This sets `panic = "unwind"` in `[profile.release]` and keeps `lto`,
`codegen-units` and `strip`. The decision is recorded as
`docs/adr/0024-release-builds-unwind.md` (`doc:adr/0024-release-builds-unwind`,
cited from the profile itself), per the maintainer's design pointer on the
issue.

### No behaviour is left dangling by the flip

Two paths read the strategy with `cfg!(panic = "abort")` — the terminal restore
in `crates/stella-tui/src/term.rs` and the journal flush in
`crates/stella-cli/src/session_persist.rs`. Both already had an unwinding
branch, and that is the branch a shipped binary now takes: the terminal is
restored by `TerminalGuard`'s `Drop`, and the journal is fsynced by
`SessionJournal`'s `Drop` when the last `SharedSink` clone goes. Neither
aborting branch is deleted — a host that builds these crates with an aborting
profile of its own still needs it — and both doc comments now say which branch
ships. `crates/stella-cli/src/daemon/console.rs`'s panic drain was never
conditioned on the strategy and stays as it was; its comment claimed `abort`
was the reason it exists, and the real reason (the drain in `main` is a call,
not a `Drop`, so an unwind skips it) is now written down.

## The witness and how it fails on the old code

`cargo test` always unwinds — the harness forces it — so **no test in this
repository can see the panic strategy a release build was made with**. That is
exactly why this defect survived: `tests/render_robustness.rs` and
`panel_guard.rs`'s own unit tests all pass under `abort` too, because they never
run under it.

The witness is therefore an example binary run in the release profile:

- `crates/stella-tui/examples/panic_guard_probe.rs` panics inside
  `stella_tui::panel_guard::guarded_band`, then asserts the error card was
  painted and exits 0.
- `.github/workflows/ci.yml` runs it right after the release smoke build:
  `cargo run -p stella-tui --example panic_guard_probe --release --config 'profile.release.lto="thin"' --config 'profile.release.codegen-units=16'`
  — the same two overrides as the build above it, neither of which touches
  `panic`, so it links what that step already compiled and observes the real
  release strategy.

**Mechanism (I cannot run it locally — this repo's agents push and read CI).**
Revert `Cargo.toml` to `panic = "abort"` and the probe's `catch_unwind` never
runs its handler: the panic goes straight to `abort()`, the process dies on
SIGABRT, `cargo run` exits 134 and the step is red. With `panic = "unwind"` the
catch runs, the card is painted, and it prints
`panic_guard_probe: OK — a panel panic was contained under this profile`. The
CI log line for that step is the evidence, cited below once the run is green.

Making the probe enter the *real* boundary rather than a copy of it is why
`panel_guard::guarded_band` and `guarded_overlay` are now `pub` (the module was
`pub(crate)`); an example is a separate crate and sees only the public API. The
module doc says so at the point of the change.

## The comment sweep

`rg -n catch_unwind crates` plus `rg -n 'panic = "abort"'` found ten comments
claiming a panic ends the process in a release build. All ten are corrected, in
`stella-tui` (`panel_guard.rs`, `term.rs`, `README.md`), `stella-core`
(`bus.rs`, `README.md`), `stella-serve` (`subagents.rs`), and `stella-cli`
(`subagent.rs`, `subagent/tests.rs`, `subsession.rs`, `main.rs`,
`session_persist.rs`, `command_deck.rs`, `daemon/console.rs`). Where the
guidance behind a claim still stands regardless of profile — a bus observer
returns `Err` rather than panicking, a policy hook returns `Deny` — the guidance
is kept and re-grounded on "a host may build with `abort`" rather than on a
statement about this workspace that is no longer true.

## Exemplar followed

The probe follows `crates/stella-tui/examples/splash_frames.rs` — the crate's
existing pattern for a small example that drives a real render path and prints
its result.

## Notes

- Tests deleted: None.
- No new dependencies.
- `make guards-fast`, `make doc-links`, `make invariants` and
  `python3 scripts/check-prose.py` are green locally (they compile nothing);
  everything that compiles is CI's.
- Cost: `cargo run --example` builds `stella-tui`'s dev-dependencies in release
  once. The alternative — a `[[bin]]` target — would ship a binary in the
  release artifact set, which is worse.

## Remaining work filed

None found beyond the change itself; the two `cfg!(panic = "abort")` branches
are documented as reachable only for a host with an aborting profile rather than
left as unexplained dead code.

Closes #5493

## Summary by Sourcery

Make release builds unwind on panic so the workspace's existing recovery boundaries protect user sessions and remain verifiable in CI.

New Features:
- Add a release-mode panic guard probe that verifies panel panics are contained in the actual release profile.

Bug Fixes:
- Enable panic unwinding in release builds so existing panel, server-connection, subagent billing, worker, and hook panic boundaries work in shipped binaries instead of aborting the process.
- Preserve terminal restoration and journal flushing behavior across both unwinding and host-provided aborting profiles.

Enhancements:
- Expose the panel guard render functions needed by the release-mode probe and update panic-handling guidance throughout the workspace to reflect the release strategy.

CI:
- Run the release panic guard probe in CI alongside the release smoke build to detect regressions to the panic strategy.

Documentation:
- Document the release-build unwinding decision in ADR 0024 and register it in the ADR documentation indexes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -633,6 +633,18 @@ jobs:
         if: ${{ !cancelled() && steps.rust.outcome == 'success' }}
         run: cargo build --workspace --release --config 'profile.release.lto="thin"' --config 'profile.release.codegen-units=16'
 
+      # The one check that can see the release panic strategy. `cargo test`
+      # always unwinds, so the panel boundary and stella-serve's per-connection
+      # isolation were both tested in a build that could not tell an aborting
+      # profile from an unwinding one. This example panics inside
+      # `stella_tui::panel_guard::guarded_band` and exits 0 only if the catch
+      # ran; set `panic = "abort"` in the root Cargo.toml again and it dies on
+      # SIGABRT instead. Same two profile overrides as the build above, so it
+      # links what that step already compiled — neither one touches `panic`.
+      - name: a panel panic is contained under the release profile
+        if: ${{ !cancelled() && steps.rust.outcome == 'success' }}
+        run: cargo run -p stella-tui --example panic_guard_probe --release --config 'profile.release.lto="thin"' --config 'profile.release.codegen-units=16'
+
       # Two guard self-tests that want the toolchain, so they sit here rather
       # than in guard-self-tests.yml with their twenty toolchain-free siblings
       # (#3820, #4427). This one is also a `GATE_STEPS` member that ran in no
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -224,5 +224,13 @@ install-updater = false
 [profile.release]
 lto = true
 codegen-units = 1
-panic = "abort"
+# Unwind, so that `catch_unwind` works in the binary a user runs. Two shipped
+# promises rest on it: a panel that panics paints an error card and the rest of
+# the screen keeps working (`stella_tui::panel_guard`), and a panic in one
+# `stella-serve` connection ends that connection rather than the server. Under
+# `panic = "abort"` a catch never runs and both promises are void. The trade is
+# code size and a little speed, and `doc:adr/0024-release-builds-unwind` records
+# it. `crates/stella-tui/examples/panic_guard_probe.rs` runs under this profile
+# in CI and fails if this line ever says "abort" again.
+panic = "unwind"
 strip = true
--- a/crates/stella-cli/src/command_deck.rs
+++ b/crates/stella-cli/src/command_deck.rs
@@ -644,10 +644,9 @@ pub async fn run_deck_session(
                 .max(restored.spent_usd.unwrap_or(0.0)),
         );
     }
-    // A release-build panic aborts before any `Drop` or `catch_unwind` runs
-    // (the workers' panic catch included), so this hook is the journal's
-    // only flush point on that path — the terminal is restored by
-    // stella-tui's own hook the same way.
+    // The flush point for a build made with an aborting profile, where no
+    // `Drop` and no catch runs. Every profile here unwinds, so the journal is
+    // flushed by the sink's own drop instead; see the guard's own docs.
     let _journal_panic_guard =
         crate::session_persist::JournalPanicGuard::install(journal_sink.clone());
     let _tee = crate::session_persist::spawn_journal_tee(
--- a/crates/stella-cli/src/daemon/console.rs
+++ b/crates/stella-cli/src/daemon/console.rs
@@ -531,9 +531,10 @@ impl Drainable {
 /// past `main`'s orderly [`ConsoleGuard::drain`], the process exits without
 /// joining the pump threads, and up to a pipe buffer of output — the panic
 /// message itself, most of the time — dies in the pipe. That is the one
-/// artifact a postmortem of a supervised child actually wants. Release builds
-/// run with `panic = "abort"`, so this hook is the only cleanup such a process
-/// ever gets.
+/// artifact a postmortem of a supervised child actually wants. The drain in
+/// `main` is a call, not a `Drop`. An unwind skips it. So this hook is the
+/// only cleanup such a child gets. It asks no `cfg` about the panic strategy.
+/// It is the last chance to drain under either one.
 ///
 /// The hook chains rather than replaces: whatever hook is already installed
 /// (the diagnostics dump, or the default printer) runs FIRST, so its output
--- a/crates/stella-cli/src/main.rs
+++ b/crates/stella-cli/src/main.rs
@@ -531,9 +531,8 @@ fn main() -> ExitCode {
     // before that boundary would race the env mutations it fences — and
     // drained as this function's last act so the final lines land.
     //
-    // A panic would unwind (or, under the release profile's
-    // `panic = "abort"`, not unwind at all) past that last act and strand up
-    // to a pipe buffer — usually including the panic message itself — so
+    // A panic unwinds past that last act — every profile here unwinds — and
+    // strands up to a pipe buffer, usually including the panic message, so
     // `arm_panic_drain` chains a hook that performs the same idempotent
     // drain (#1616). Whichever of the two arrives first does the one real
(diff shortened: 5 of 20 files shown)


macanderson:
Automated PR sweep — `guard self-tests` is fixed here; `dod` is the only red left, and it needs a decision from you.

**`guard self-tests` is green on this head.** It was red on every open PR in the repo, not just this one: `scripts/test-releases-published-pagination.sh` pinned its fail-side to `origin/main`, so it inverted the moment its own fix merged. Repaired by #5688 (`3aea7c4`); this branch picked it up when I merged main in. That failure was never this PR's.

**`dod / dod` remains red on #5493's three unchecked items, and I ticked none of them.**

The list is written as a fork — one item records a decision, and the other two are conditional on which way that decision went:

- *A decision recorded, as an ADR if the profile changes.*
- *If the profile keeps `abort`: every comment claiming a panic is contained …*
- *If the profile moves to `unwind`: a test that a band panic is contained in …*

This PR's title says the profile moves to `unwind`, which resolves the fork to the third item and makes the second one not-applicable. But which items apply, and whether SCR-002 wants an ADR for a workspace-profile change, is exactly the judgement the first box is asking you to record — and a sweep that decides it for you turns the gate into a formality. So all three stay open.

Two things would clear it, both quick and both yours: record the decision (an ADR under `docs/adr/` if you agree the profile change warrants one — SCR-002 says architecture decisions get one rather than a question), and mark the `abort` branch N/A so only the applicable item has to be proven.

Worth flagging that a conditional DoD list has a failure mode this one is sitting in: the gate counts unchecked boxes and cannot tell a genuinely outstanding item from a branch that was never going to apply. Two of these three can never both be ticked.

---

PR #5672, opened 2026-09-03
ci(auto-tag): approve the version-sync PR's parked runs and merge under protection

macanderson:
## What and why

Every release version-sync pull request opened with all thirteen of its
workflow runs parked at `action_required` and stayed there until a maintainer
clicked "Approve and run" thirteen times.

**The cause, established rather than guessed.** The repository's Actions
approval policy is `first_time_contributors`
(`actions/permissions/fork-pr-contributor-approval`), and the PR's author is
`github-actions[bot]`, which counts as one. That policy is normally described
as a fork rule, but it reaches this same-repo branch too. Checked live on
#5634: all thirteen `pull_request`-event runs on its head sha
`398fd6435d` had `conclusion: action_required`.

**Why the workflow did not notice.** `auto-tag.yml` never depended on those
runs. It dispatched `ci.yml` onto the branch, watched that run, and then merged
past branch protection with `RELEASE_ADMIN_TOKEN`. The dispatch covered two of
main's five required contexts. The other three belong to workflows with no
`workflow_dispatch` trigger, so the admin bypass was doing the rest of the
work:

| required context | workflow | `workflow_dispatch`? |
|---|---|---|
| `fmt + clippy + test` | `ci.yml` | yes |
| `cargo deny + cargo audit` | `ci.yml` | yes |
| `harbor_adapter + analyzer pytest` | `bench.yml` | yes |
| `main is not known-broken` | `main-red-hold.yml` | **no** |
| `gate steps no other workflow runs` | `guard-self-tests.yml` | **no** |

**Why it broke now.** `enforce_admins` is `true` on main's protection, so
`mergePullRequest` refuses the admin merge on the same grounds it refuses any
other caller. Run
[33666329889](https://github.com/macanderson/stella/actions/runs/33666329889),
step `Wait for the dispatched checks…`:

```
GraphQL: 5 of 5 required status checks are expected. (mergePullRequest)
##[warning]admin merge of bot/version-sync with RELEASE_ADMIN_TOKEN failed — falling back to auto-merge.
```

The step then exited 0, so the job was green and the write-back had not landed.

## The change

`.github/workflows/auto-tag.yml` only, plus a `RELEASING.md` update.

1. **New step, `Approve the version-sync PR's parked workflow runs`.** After
   the PR is opened it polls the head sha for two minutes and `POST`s
   `/actions/runs/{id}/approve` for every run whose conclusion is
[...]

Issue #5681, opened 2026-09-03
guard-self-tests is red on main: the pagination witness pins its fail-side to origin/main, which now carries the fix


macanderson:
## The break

`guard-self-tests.yml`'s `guard self-tests` job fails on `main` and on every PR
branch cut from it. The failing step is **"the release fetch has no fixed count
cap"**, which runs `scripts/test-releases-published-pagination.sh`.

Reproduced on a clean checkout of `main` at `bc3e671`:

```
$ ./scripts/test-releases-published-pagination.sh
FAIL MAIN main's script hits its 1000-item cap on 1001 releases and errors — exit 0, got: check-releases-published: OK — every v* tag older than 5256000m has a published release (0 tag(s) checked, 24 grandfathered).
ok   NEW this branch's script pages through all 1001 releases and reports clean
ok   NEW reports no truncation or page-sanity error on 1001 releases

passed 2, failed 1
```

Seen live on PR #5658's run
[33697895007](https://github.com/macanderson/stella/actions/runs/33697895007/job/100470658203)
(step 37). It is not that PR's doing — its diff is `stella-store` / `stella-cli`
and touches neither script.

## Why it fails

The test's `MAIN` arm reads the *other* copy of the script under test out of a
moving ref:

```sh
git -C "$repo_root" fetch --depth=1 -q origin main 2>/dev/null || true
git -C "$repo_root" show origin/main:scripts/check-releases-published.sh >"$MAIN_SCRIPT"
```

and then asserts that copy **errors** on 1001 releases, which was true only
while `origin/main` still carried the capped `gh release list --limit 1000`
fetch. `origin/main` now carries the paginated fetch:

```
$ git show origin/main:scripts/check-releases-published.sh | grep -n 'paginate'
172:pages_json="$(CLICOLOR_FORCE=0 NO_COLOR=1 gh api --paginate --slurp 'repos/{owner}/{repo}/releases?per_page=100')"
```

That line is byte-identical to the working tree's (`git diff origin/main --
scripts/check-releases-published.sh` is empty). So the arm now runs the *fixed*
script, the fixed script correctly reports OK, and the arm fails — permanently,
on every branch, from the moment the fix merged.

A witness pinned to a moving ref stops being a witness once the ref moves past
it. The fail-before evidence was real at review time; it cannot be re-derived
by re-reading `main` afterwards.

**Timeline, which is what settles it.** #5648 (the paginated fetch) merged at
[...]

PR #5687, opened 2026-09-03
fix(scripts): pin the pagination witness fail-side to a commit

macanderson:
## What & why

`guard self-tests` is red on **every open pull request** in this repository, and `main`'s own CI is green. This unblocks the queue.

`scripts/test-releases-published-pagination.sh` is a witness test whose fail-side read `origin/main`:

```sh
git show origin/main:scripts/check-releases-published.sh >"$MAIN_SCRIPT"
```

It then asserted that copy still hits its 1000-item cap. That held while the paginating rewrite was unmerged. Once it landed (`74c5673`, PR #5648, 23:59), `origin/main` carried `gh api --paginate`, the capped behaviour was gone, and the assertion could never pass again:

```
FAIL MAIN main's script hits its 1000-item cap on 1001 releases and errors — exit 0,
got: check-releases-published: OK — every v* tag older than 5256000m has a published release
```

`guard-self-tests.yml` runs on `pull_request` only, so nothing re-asked the question after the merge: `main` reads green while every PR opened against it fails. Confirmed red on #5672, #5667, #5666 and #5678 — the same single failing assertion in each.

A fail-side that reads a moving ref stops proving anything the moment the fix merges into that ref. `CAPPED_REV` pins it to `5096b3f`, the last commit whose `check-releases-published.sh` still called `gh release list --limit`. Both halves of the witness stay demonstrable for as long as that commit does, and the shallow-clone `SKIP` path is unchanged — an unreachable pinned commit skips the fail-side exactly as an unreachable `origin/main` did.

Deleting the fail-side would have cleared the red just as fast and left the guard proving only one direction, which this repository asks of no other guard.

## The witness

The test is its own witness. On `main` it reports `passed 2, failed 1` and exits 1 — the exact `FAIL MAIN` line CI prints, reproduced locally before the fix. With this change:

```
ok   OLD the capped script hits its 1000-item cap on 1001 releases and errors
ok   NEW this tree's script pages through all 1001 releases and reports clean
ok   NEW reports no truncation or page-sanity error on 1001 releases

passed 3, failed 0
```

The `OLD` arm genuinely runs and genuinely proves the cap — it is not passing by being skipped.

## The gate

Shell only; nothing here compiles. Per CLAUDE.md ("CI builds and tests; this laptop does not") the cargo tiers are CI's, and this diff reaches no crate.

- [x] `./scripts/test-releases-published-pagination.sh` — 3 passed, 0 failed, exit 0
- [x] `scripts/check-prose.py` — exit 0 (grade and issue-reference violations in my first draft fixed, not baselined)
- [x] `scripts/check-line-citations.py` — exit 0
- [x] `scripts/check-left-behind.sh` — exit 0
- [x] `scripts/check-file-size.sh` — exit 0
- [ ] `shellcheck` — **not run**: not installed in this environment. It is a required gate step and CI is the evidence for it. `bash -n` parses clean.

## Nothing left behind

The general shape — a witness whose fail-side reads a moving ref, self-invalidating on merge — is worth a sweep, but `check-rebase-replay.sh`'s `MAIN`-style comparison is the only other fail-side in `scripts/` I found reading a moving ref, and it compares two trees by design rather than asserting a bug persists. No second instance to file.

## Ground-rule check

- [x] No new outbound network calls — the pinned-commit fetch replaces the branch fetch that was already there

`CAPPED_REV` is a full 40-character SHA so it cannot be confused with a ref. The fetch is `git fetch --depth=1 origin <sha>`, which GitHub serves for a reachable commit; if a future checkout cannot reach it, the fail-side skips with a message rather than failing, which is how the `origin/main` version already behaved.

Rejected alternative: deleting the `MAIN`/`OLD` arm. Faster, and it would leave the guard unable to demonstrate its own failure direction — the property `guard-self-tests.yml` exists to hold.

https://claude.ai/code/session_01U91qDoJdXhCbEDeSEUSUot

---
_Generated by [Claude Code](https://claude.ai/code/session_01U91qDoJdXhCbEDeSEUSUot)_

## Summary by Sourcery

Pin the pagination witness test’s legacy comparison to a commit so it continues validating both failure and success paths across pull requests.

Bug Fixes:
- Prevent the pagination witness test from invalidating itself after the fix reaches the default branch by pinning its legacy fail-side to a known commit.

Enhancements:
- Preserve verification of both the old capped behavior and the new paginated behavior while retaining graceful skipping when the pinned commit is unavailable.

Tests:
- Update the witness assertions and messaging to compare the current script with the pinned capped version.

### DIFF
--- a/scripts/test-releases-published-pagination.sh
+++ b/scripts/test-releases-published-pagination.sh
@@ -9,38 +9,50 @@
 # fetch end to end with a fake `gh` on PATH, so a fake list of 1001 releases
 # runs with no network at all.
 #
-# It runs two scripts against that same list: the one `main` ships today,
-# and the one in this working tree. Main's script hits its old cap and
-# errors. This branch's script pages through all 1001 and reports clean.
+# Two scripts run against that same list. One is the capped script named by
+# CAPPED_REV below. The other is the one in this working tree. The capped one
+# hits its cap and errors. This tree's one pages through all 1001 and reports
+# clean.
+#
+# The fail-side names a commit, never `origin/main`. Point it at a moving ref
+# and it proves nothing once the fix merges there. It then asserts that the
+# fixed script still carries the bug, and fails on every open pull request at
+# once. CAPPED_REV is the last commit that carried the cap. Both halves of
+# this witness hold for as long as that commit does.
 #
 # bash 3.2 compatible.
 
 set -uo pipefail
 
+# 5096b3f — the last commit whose check-releases-published.sh still called
+# `gh release list --limit` rather than `gh api --paginate`. Its copy of that
+# script is the capped one this test drives.
+CAPPED_REV="5096b3ff7df2e12986957287fe13b03ebbeacc3e"
+
 repo_root="$(cd "$(dirname "$0")/.." && pwd -P)"
 NEW_SCRIPT="$repo_root/scripts/check-releases-published.sh"
-MAIN_SCRIPT="$repo_root/scripts/.check-releases-published.main-copy.sh"
+CAPPED_SCRIPT="$repo_root/scripts/.check-releases-published.capped-copy.sh"
 
 pass=0
 fail=0
 
 work="$(mktemp -d)"
-cleanup() { rm -rf "$work"; rm -f "$MAIN_SCRIPT"; }
+cleanup() { rm -rf "$work"; rm -f "$CAPPED_SCRIPT"; }
 trap cleanup EXIT
 
-# CI checks out one ref with no history, so `origin/main` may not exist yet.
-# One shallow, on-demand fetch of just that branch tip fixes that without
-# touching the shared checkout step every other guard self-test also runs
-# under. It only ever reads; nothing here can move a real branch.
-git -C "$repo_root" fetch --depth=1 -q origin main 2>/dev/null || true
+# CI checks out one ref with no history, so CAPPED_REV may be absent. One
+# shallow fetch of that single commit gets it. The shared checkout step every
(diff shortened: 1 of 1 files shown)


macanderson:
Closing this in favour of #5688, which fixes the same break better. Both PRs were written independently within seven minutes of each other — the duplicate-repair collision AGENTS.md describes, reached the same way: every session with an open PR sees the red at once and draws the same correct conclusion.

Same diagnosis in both (the `MAIN` arm reads `scripts/check-releases-published.sh` out of a moving ref and asserts it still hits the cap; the paginating rewrite landed on `main`, so the assertion can never hold again). The fix differs, and #5688's is the more durable one:

- **This PR pins the fail-side to `5096b3f`**, the last commit carrying the cap. It survives the merge, but the fail-side is then a frozen artifact — it can never regress, so it stops carrying regression value the moment it is pinned. It also keeps a `git fetch` and a `SKIP` branch, so a shallow CI checkout and a full clone can run different amounts of the test.
- **#5688 removes the git-ref dependency outright.** Both arms run the shipping script, and the fail-side is pointed at that script's own `max_pages` sanity ceiling — verified present in `check-releases-published.sh`, with the `past the N-page sanity ceiling` string its `CEILING` arm matches. No network, no `SKIP`, and the fail-side exercises live code, so it still fails if the ceiling is ever removed.

Pointing a witness's fail-side at live behaviour rather than at history is the better answer, and #5688 also carries the stronger evidence: three mutations of the script under test, each reverted with the clean run re-confirmed.

Nothing from this PR needs porting — #5688 is a superset. No issue filed for the residue, since #5681 already tracks the break and #5688 closes it.

---

Issue #5693, opened 2026-09-03
Check that OpenRouter accepts the upstream name it reports when that name is sent back in provider.order


macanderson:
The adapter now feeds OpenRouter's own answer back to it: it reads the
top-level `provider` field off a response and sends that exact string in
`provider.order` on later calls (#4172,
`crates/stella-model/src/zai/upstream.rs`). The values the gateway sends are
display names — `Sail Research`, `Amazon Bedrock`, `Together`, `Morph`, and
`Modal` are the ones in this repo's own `store.db` today.

Nobody in this repository has checked that `provider.order` accepts a display
name. The gateway's docs have used both a display name and a lower-case slug
form (`amazon-bedrock`) at different times, and no test here can settle it:
every witness is a wiremock, which will accept any string.

## Why it is not a bug today

The learned pin sends `allow_fallbacks: true`. An `order` entry the gateway
does not recognise leaves it free to route as it always did, so the call
still succeeds. The failure mode is silent uselessness, not breakage — the
change would buy nothing and nothing would say so. An operator's
`upstream_pin` is a different matter: it sends `allow_fallbacks: false` and a
name the gateway cannot match would fail the call outright, but that value is
typed by a person against the gateway's docs, not read off a response.

## How to settle it

One live call is enough. Send a request with
`provider: {order: ["<a display name from store.db>"], allow_fallbacks: false}`
and see whether it is served or refused; repeat with the slug form. Or read
the served `provider` back on a call sent with `allow_fallbacks: false` and a
display-name order — if it comes back served by that vendor, the name matched.

If the display name is not accepted, map it. The mapping belongs in
`crates/stella-model/src/zai/upstream.rs` beside `ServedUpstream`, applied
where the name is learned, and it needs a witness of its own.

## Definition of done

- [ ] A live OpenRouter call establishes whether `provider.order` accepts the
      display name the response reports.
- [ ] The answer is written into the OpenRouter `CachePosture` row in
      `crates/stella-model/src/provider_parity.rs`, which today asserts the
      round trip without evidence.
- [ ] If the display name is not accepted, `ServedUpstream::learn` maps it to
      the accepted form, with a witness test.

Issue #5692, opened 2026-09-03
Measure the cold-call share on an OpenRouter turn now that the adapter re-asks for its upstream


macanderson:
The adapter now reads the upstream that served an OpenRouter session's first
answer and asks for it again on every later call, with fallbacks allowed
(#4172). Nobody has measured what that buys.

The maintainer's 2026-08-22 note on #4172 is explicit that this half is not
optional and must not be assumed: "Do not close this on the reasoning that a
pin must have helped." #4172 shipped the code and closed on the design
decision; this issue is the measurement it owes.

## What to measure

Run one read-heavy turn through `openrouter` on a slug the gateway routes to
several vendors (`moonshotai/kimi-k3` is the one both prior investigations
used), then read `.stella/private/store.db`:

```sql
-- how many vendors served one turn, and how often the route moved
select step, upstream_provider from step_receipt
where execution_id = <id> order by step;

-- the cold calls and what they cost
select r.step, r.upstream_provider, t.input_tokens, t.cache_read_tokens, t.cost_usd
from step_receipt r join telemetry t
  on t.execution_id = r.execution_id and t.turn_instance = r.turn_instance
 and t.engine_step = r.step and t.call_seq = r.call_seq
where r.execution_id = <id> and t.cache_read_tokens = 0 order by r.step;
```

Report the cold-call share of cost, and split the cold calls two ways: those
sitting at a route change, which the re-ask is meant to remove, and those
inside one vendor's run, which are cache-shard misses it cannot touch. The
before figures to compare against are on #4172: 20 of 69 calls cold for 54%
of a $2.41 turn in one session, and execution 327 of this repo's own store —
174 calls, one slug, three vendors, the route moving 12 times, five cold
calls carrying $0.97 of $4.65, three of the five at a route change.

## Constraints

- The measurement needs a real gateway turn. There is no way to fake it.
- Report the unflattering number if that is the true one (CLAUDE.md,
  "Measure honestly, especially when it costs us").
- A turn served by one vendor start to finish measures nothing about this
  change and has to be re-run against congestion.

## Definition of done

- [ ] One read-heavy OpenRouter turn is run on a build carrying the re-ask.
- [ ] The route sequence for that turn is posted: vendors, and how many times
      the route moved.
- [ ] The cold-call share of the turn's cost is posted, split into
      route-change misses and within-vendor shard misses.
- [ ] The result is compared with the before figures on #4172, and the
      comparison says plainly whether the change moved the number.
[...]

PR #5688, opened 2026-09-03
fix(ci): stop the pagination witness pinning its fail-side to origin/main

macanderson:
## What & why

`guard-self-tests.yml`'s `guard self-tests` job is red on `main` and on every
branch cut from it. The failing step is **"the release fetch has no fixed count
cap"** (`scripts/test-releases-published-pagination.sh`).

Reproduced on a clean `main` checkout at `bc3e671`:

```
FAIL MAIN main's script hits its 1000-item cap on 1001 releases and errors — exit 0, got: check-releases-published: OK — every v* tag older than 5256000m has a published release (0 tag(s) checked, 24 grandfathered).
ok   NEW this branch's script pages through all 1001 releases and reports clean
ok   NEW reports no truncation or page-sanity error on 1001 releases

passed 2, failed 1
```

**The cause, established rather than guessed.** The `MAIN` arm read a second
copy of the script under test out of a moving ref and asserted *that* copy
errors:

```sh
git -C "$repo_root" show origin/main:scripts/check-releases-published.sh >"$MAIN_SCRIPT"
```

That assertion held only while `origin/main` still carried the capped
`gh release list --limit 1000` fetch. It no longer does — `origin/main`'s copy
is byte-identical to the working tree's, both carrying the paginated fetch:

```
$ git diff origin/main -- scripts/check-releases-published.sh   # empty
$ git show origin/main:scripts/check-releases-published.sh | grep -n paginate
172:pages_json="$(CLICOLOR_FORCE=0 NO_COLOR=1 gh api --paginate --slurp 'repos/{owner}/{repo}/releases?per_page=100')"
```

So the arm now runs the *fixed* script, the fixed script correctly reports OK,
and the arm fails — permanently, from the moment the fix merged. A witness
pinned to a moving ref stops being a witness once the ref moves past it.

**The timeline settles it.** #5648 (the paginated fetch) merged at `23:59:29Z`.
Every `guard self-tests` run before that passed — #5640 (22:36), #5646 (23:07),
#5653 (23:29), #5641 (23:32). Every run after it failed — #5658 (00:04), #5667
(00:06), #5666 (00:16), #5672 (00:17). Nothing else changed in that window.

Closes #5681

## The change

`scripts/test-releases-published-pagination.sh` only. Both arms now run the
script this tree ships; nothing reads a second copy out of a git ref.

- **PAGES** keeps the no-cap assertion — 1001 releases over 11 pages are all
  walked, and nothing reports truncation.
[...]

Issue #3054, opened 2026-08-12
Stored records don't say which AI vendor actually served a request routed through a gateway


macanderson:
## Problem

#2786 ("pin the gateway's upstream and record who served the call") added
`upstream_provider: Option<String>` to `AgentEvent::StepUsage`
(`crates/stella-protocol/src/event.rs`) and `CompletionResult`
(`crates/stella-protocol/src/completion.rs`). Found while documenting
`--upstream-pin` for #3042: the field is saved, but only inside the raw
`events` table payload —

```bash
sqlite3 -json .stella/private/store.db \
  "SELECT payload FROM events WHERE execution_id = 42 AND event_type = 'step_usage'" \
  | jq -r '.[].payload | fromjson | select(.upstream_provider) | "\(.provider) → \(.upstream_provider)"'
```

Neither `step_receipt` nor the `telemetry` table
(`crates/stella-store/src/ddl.rs`) has a column for it — both carry only
`provider`, which for a gateway call is always the gateway's own id
(`openrouter`), never the vendor it routed to. So:

- **`stella inspect <EXEC>`** (`crates/stella-cli/src/inspect.rs`,
  `RecordedCall`/`CallJson`) can't show which vendor served a past call. It
  reads `step_manifest`/`step_receipt`, not `events`.
- **`stella stats`** and the Observatory read `telemetry`/`step_receipt`
  aggregates — same gap.
- The interactive mode builds `upstream_provider` in test fixtures
  (`crates/stella-tui/src/views/agents.rs`), but no real (non-test) code
  reads the field to show it live either.

A reader auditing a past run — the exact case #2786's own commit message
describes ("a gateway benchmark unauditable: a run that claims to hold the
model provider fixed could not prove it did") — can't get the answer from any
of Stella's normal query surfaces today. They have to know to query `events`
directly, which is documented as a workaround in
`website/content/docs/telemetry/index.mdx` (added by #3042) but isn't a
substitute for the field reaching a real column.

## What to build

Add an `upstream_provider` column to whichever table is the right home —
`step_receipt` is the more natural fit (`crates/stella-store/src/ddl.rs`,
`crates/stella-store/src/usage.rs` for the write path), since it's already
the per-model-call table `provider` lives on. Wire it through:

- The store write path that projects `AgentEvent::StepUsage` into
  `step_receipt` (or wherever that projection actually lives — confirm
  whether one already exists, following the pattern used for `tool_calls` in
[...]

PR #5694, opened 2026-09-03
feat(stella-model): re-ask OpenRouter for the upstream that served the first call

### LINKED ISSUE #4172: Decide whether OpenRouter should default to pinning its upstream provider (cold cache costs 39% of one gateway session)
## The decision

Should the built-in `openrouter` row carry a default `upstream_pin`?

This is a product decision with a measured cost attached, and it is deliberately **not** being made in a PR. `crates/stella-cli/src/config/providers.rs:32` documents the current default as a choice, not an oversight:

> Upstreams a *gateway* provider is pinned to, in preference order, from settings.json's `upstream_pin`. Empty for every built-in row: routing is the gateway's choice until an operator says otherwise, and only a measured comparison needs it fixed.

Branch `worktree-session-forensics-fixes` therefore left the default alone and made the condition visible instead (a boot notice, `engine_config::unpinned_gateway_notice`). That is the reversible half. This issue is the other half.

## What the measurement says

Session `ses-1787342320630-36613`, one 17-minute turn on `moonshotai/kimi-k3` via `openrouter`, `$2.4071` total across 69 model calls:

- **20 of 69 calls (29%) read zero cached tokens.** They carried 25% of the input tokens and **54% of the cost** ($1.2941).
- Cost per 1k input tokens: `$0.00314` on a cold call vs `$0.00088` on a warm one — **3.6x**.
- Counterfactual: every token at the warm rate would have cost `$1.4771`. **~$0.93, 39% of the turn, is attributable to cold prefixes.**
- **It is not TTL expiry.** Median idle before a miss was 0.3s; before a hit, 0.2s. Calls 2-6 run `miss, hit, miss, hit, miss` back to back, and a hit reading 11,844 tokens is followed 0s later by a total miss on a *larger* prefix.
- The pattern is front-loaded: the last 17 calls are an unbroken run of hits, consistent with gateway congestion easing rather than anything the agent changed.

## Mechanism

`ZaiProvider` sends the sticky `session_id` for any gateway session (`zai.rs:1274`), but `provider.order` + `allow_fallbacks: false` is gated on a non-empty `upstream_pin` (`zai.rs:1280`), and `factory.rs:265` only calls `with_upstream_pin` when the spec carries one. With no pin, `session_id` is the only stickiness in play — and it is a hint the gateway may decline under load, not a pin. (The provider parity matrix claimed it "pins every turn of a session to the same upstream provider + cache shard"; that claim is corrected on the branch above, with this session as the counter-example.)

## The trade

**For a default pin:** ~39% of a gateway turn's cost is recoverable; prompt-cache stability is the single largest measured cost variable on this repo's own runs; two turns of one session served by two upstreams are also not comparable, which affects any measurement taken through the gateway.

**Against:** it hard-codes a vendor preference into a gateway whose whole value proposition is routing; `allow_fallbacks: false` means a pinned upstream that is down fails the request rather than degrading; the right pin is model-dependent (the upstreams serving `moonshotai/kimi-k3` are not the ones serving an Anthropic route), so a single static default is wrong for most slugs.

That last point is probably decisive against a *static* default and suggests the real options are elsewhere.

## Options

1. **Leave unpinned; notice only.** Shipped. Zero risk, saves nothing on its own.
2. **Static default pin per built-in row.** Cheapest to implement, wrong for most slugs.
3. **Pin on first success, per session.** Record which upstream served the first call of a session and pin subsequent calls to it. Adaptive, needs no vendor knowledge, and is the same shape `session_id` is *supposed* to provide. Requires reading the upstream out of the response, which is a separate gap — see the memory note that OpenRouter's served upstream is currently unrecorded.
4. **Pin only when a measurement is running** (bench/arena paths), leaving interactive sessions unpinned.

Option 3 looks strongest and is the one worth costing.

## Definition of done

The decision is option 3, pin on first success with fallbacks allowed (maintainer, 2026-09-02). The boxes below are that decision spelled out as work.

- [x] The decision is recorded in the PR body, so a reader who finds the code first can find the reasoning.
- [x] `crates/stella-model/src/zai.rs` reads the served upstream off the response — the streamed chunks and the unary body alike — and remembers it for the session.
- [x] With no `upstream_pin` configured, every call after the first sends `provider: {order: [served], allow_fallbacks: true}`.
- [x] A session's first call still sends no `provider` field, so its bytes are unchanged.
- [x] An explicit `upstream_pin` still wins, and still sends `allow_fallbacks: false`.
- [x] An endpoint that is not the gateway is never sent a `provider` field, whatever it answers with.
- [x] Wiremock witnesses cover each of those, in `crates/stella-model/src/zai/tests/upstream_pin.rs`.
- [x] The served upstream reaches the durable record, or the PR says where it stops.
- [x] The OpenRouter `CachePosture` row in `crates/stella-model/src/provider_parity.rs` describes what the adapter now does and names the test that proves it.
- [x] Prose that says routing is left to the gateway is corrected: the built-in provider row and the boot notice.

Measuring the cold-call share after the change is a separate run on a real gateway turn, and is tracked on its own issue.

### Verification at `de2d496` (PR #5694)

Each box ticked against the diff and the tree, item by item. Ticked by an
automated PR sweep, so the evidence is named rather than asserted.

1. **Decision in the PR body.** #5694 opens with `## The decision` — "Option 3
   from the issue, adaptive, per the maintainer's 2026-09-02 pointer: pin on
   first success, fallbacks allowed", with the static-pin rejection reasoning.
2. **Reads the served upstream, both paths.** `zai.rs` calls
   `self.served_upstream.learn(...)` twice: once on the unary result inside the
   `recovery.use_unary()` arm, once on the streamed `outcome`. The comment on
   the first names why — the stream→unary fallback can serve part of a session
   without anyone choosing it.
3. **`allow_fallbacks: true` when unpinned.** `upstream.rs`'s `routing_pin`
   returns the learned order with `allow_fallbacks: true`; asserted on the wire
   by `a_gateway_is_reasked_for_the_upstream_that_served_the_first_call`, which
   matches the literal
   `"provider":{"order":["Sail Research"],"allow_fallbacks":true}`.
4. **First call unchanged.** Same test asserts
   `!bodies[0].contains("\"provider\"")` before asserting the second body does.
5. **Operator pin still wins, fallbacks still refused.**
   `an_operator_pin_outranks_the_upstream_that_served_the_first_call` matches
   `"order":["z-ai"],"allow_fallbacks":false` and additionally asserts the
   learned name is absent from that body.
6. **Never sent to a non-gateway endpoint.**
   `a_direct_endpoint_never_asks_for_the_upstream_it_was_told_about` drives an
   endpoint that *does* answer with a vendor name and asserts no `provider`
   object is sent. `routing_pin`'s first branch is the `serves_openrouter()`
   gate.
7. **Witnesses in the named file.** All four live in
   `crates/stella-model/src/zai/tests/upstream_pin.rs`, and that file is added
   to `adapter_sources()` so `every_witness_test_exists_in_the_adapter_sources`
   can see it.
8. **Reaches the durable record.** It does, and predates this PR:
   `step_receipt.upstream_provider` (schema v34, #3054) is written by
   `stella-store`'s `receipts.rs` INSERT and read back by its
   `SELECT … upstream_provider …`, with
   `a_receipts_upstream_provider_is_readable_without_the_event_stream` covering
   the round trip. So the box is satisfied by the first clause, not the escape
   clause.
9. **Parity row describes and names its proof.** The OpenRouter `CachePosture`
   mechanism string is rewritten to describe the re-ask, and a new
   `routing_witness` field carries
   `a_gateway_is_reasked_for_the_upstream_that_served_the_first_call`.
   `every_witness_test_exists_in_the_adapter_sources` now checks that second
   witness on the same terms as the first, so the row cannot keep the claim if
   the test is renamed away.
10. **Corrected prose.** Both surfaces: `ProviderConfig::upstream_pin`'s doc
    comment in `config/providers.rs`, and `unpinned_gateway_notice`'s emitted
    string in `engine_config.rs`, which now says Stella asks the gateway to
    keep the session on the first server rather than that routing is left to
    the gateway.

Sourcery's assessment on #5694 records all three of its objectives for this
issue as met, with no unmet rows.

The follow-on measurement — the cold-call share on a real gateway turn after
this change — is out of scope here by this issue's own note and stays on its
own issue.

## Files

- `crates/stella-cli/src/config/providers.rs:93-113` — the built-in `openrouter` row.
- `crates/stella-model/src/zai.rs:251` — `with_upstream_pin`, and `:1280` where the pin gates `provider.order`.
- `crates/stella-model/src/factory.rs:265` — the spec-to-adapter wiring.
- `crates/stella-cli/src/engine_config.rs` — `unpinned_gateway_notice`, the shipped half.
- `crates/stella-model/src/provider_parity.rs` — the OpenRouter `CachePosture` row.

---

macanderson:
## The decision

Option 3 from the issue, adaptive, per the maintainer's 2026-09-02 pointer:
**pin on first success, fallbacks allowed.** A static default pin is wrong for
most slugs — the vendors serving `moonshotai/kimi-k3` are not the ones serving
an Anthropic route — and leaving it unpinned costs a measured share of every
gateway session. Reading the served vendor off the response needs no vendor
knowledge and is what `session_id` was supposed to provide.

## What changes

`crates/stella-model/src/zai.rs` reads OpenRouter's top-level `provider` field
off a successful answer, remembers it for the life of the adapter (one agent
run), and sends it back on every later call as
`provider: {order: [served], allow_fallbacks: true}` when no `upstream_pin` is
configured. Both delivery paths learn, because the stream→unary fallback can
serve part of a session through the unary path without anyone choosing it.

The two pin shapes want different things, and the flag is where they differ:

| | `order` | `allow_fallbacks` |
|---|---|---|
| operator `upstream_pin` | what they typed | `false` — a measured run holds the variable or fails out loud |
| learned | who served the first answer | `true` — a warm cache is a saving, not a rule |

A session's first call still sends no `provider` field, so its bytes are
unchanged and the prompt cache is unaffected for anyone who never asked for a
pin. A non-gateway endpoint is never sent the field at all, whatever it
answers with — the same "does this actually address OpenRouter" gate the cache
opt-in uses, not a check on the id.

## Why, with the evidence

The parity matrix said `session_id` "pins every turn of a session to the same
upstream provider + cache shard". This repository's own `store.db` is the
counter-example. Execution 327 — 174 calls, one slug (`moonshotai/kimi-k3`),
`session_id` on every request:

```
0-64:    Sail Research        105-164: Together
65:      Together             165:     Sail Research
66-67:   Sail Research        166-171: Morph
68-75:   (alternating)        172-173: Sail Research
76-104:  Sail Research
```

Twelve route changes inside one turn. Five of the 174 calls read zero cached
tokens, and they carried **$0.97 of the turn's $4.65 (21%) on 2.9% of the
calls**:

| step | vendor | input | cache read | cost |
|---|---|---|---|---|
| 65 | Together | 37,187 | 0 | $0.1130 |
| 113 | Together | 63,866 | 0 | $0.1933 |
| 120 | Together | 67,663 | 0 | $0.2045 |
| 165 | Sail Research | 83,606 | 0 | $0.2191 |
| 166 | Morph | 83,563 | 0 | $0.2402 |

**Three of those five sit at a route change** (65 is Together's first call, 165
is the switch back, 166 is Morph's first) — $0.57, 12% of the turn. That is
what re-asking is aimed at. The other two are inside one vendor's run and are
cache-shard misses the re-ask cannot touch, which is precisely the caveat the
maintainer flagged on 2026-08-22 and it survives this change intact.

Derived from `.stella/private/store.db` by joining `step_receipt` to
`telemetry` on `(execution_id, turn_instance, engine_step, call_seq)`. It is a
census of recorded runs, not a rig run.

## What this PR does not prove

**Whether the change recovers the cold-call share.** That needs a live
gateway turn and is #5692, written as a handoff with the queries and the
before-figures to compare against. The maintainer's note that this must not be
assumed ("Do not close this on the reasoning that a pin must have helped") is
quoted there. #4172 closes on the decision the 2026-09-02 pointer asked for;
the measurement is tracked, not skipped.

**Whether OpenRouter routes on the display name it reports.** The gateway
answers with `Sail Research`, `Modal`, `Together`, `Amazon Bedrock`; this code
hands the same string back in `provider.order`, and no wiremock can settle
whether the gateway matches it. `allow_fallbacks: true` is what makes an
unmatched name cost the saving rather than the call. Declared as a gap in the
parity row and filed as #5693.

## Recording the served upstream

The design pointer asked for `step_usage` "if a field exists". It does:
`AgentEvent::StepUsage.upstream_provider` and `step_receipt.upstream_provider`
(store schema v34) both carry it already, filled from
`CompletionResult::upstream_provider` in `driver/settlement.rs`'s
`emit_step_usage` and `accounted_call.rs`. Nothing was added, and no debug log
was needed.

## Witnesses

`crates/stella-model/src/zai/tests/upstream_pin.rs`, wiremock, four new tests:

- `a_gateway_is_reasked_for_the_upstream_that_served_the_first_call` — first
  response carries `provider: "Sail Research"`; the second request body
  carries `"provider":{"order":["Sail Research"],"allow_fallbacks":true}`, and
  the first carries no `provider` field at all.
- `an_operator_pin_outranks_the_upstream_that_served_the_first_call` — the
  second body carries the operator's order with `allow_fallbacks:false` and
  never mentions the served vendor.
- `a_direct_endpoint_never_asks_for_the_upstream_it_was_told_about` — a
  non-gateway endpoint that names a vendor is still sent no `provider` field.
- `every_witness_test_exists_in_the_adapter_sources` now also checks a row's
  routing witness.

**Why they fail on `main`, stated rather than run** (this laptop does not
compile; CI is the evidence). On `main`, `build_body` gates the field on
`self.serves_openrouter() && !self.upstream_pin.is_empty()`, so with no
`upstream_pin` the request body has no `provider` key on any call, first or
hundredth. The first test's second assertion looks for
`"provider":{"order":["Sail Research"]...}` in a body that cannot contain the
substring `"provider"` at all. The parity test fails on `main` for a
mechanical reason too: `CachePosture::OptIn` has no `routing_witness` field
there, so the row does not compile.

The two negative tests pass on `main` — they guard against the new field
leaking where it must not, which only becomes possible with this change.

`crates/stella-model/src/zai/tests/upstream_pin.rs` was also added to
`adapter_sources()`; it had never been in that list, so a witness naming a
test in it would have failed the parity check on a rotted proof it could not
see.

## Exemplar followed

`ParallelToolCallPosture` (`provider_parity/parallel.rs`), for the
two-witness split: it records `ParallelAdmission` and `fan_in_witness`
separately because they are different facts settled by different evidence, and
AGENTS.md invariant 8 says collapsing them would let a row claim more than it
can show. Sending a cache marker and asking for a particular cache are the same
shape of pair, so `CachePosture::OptIn` gained an optional `routing_witness`
rather than overloading its one `witness`. `None` for `anthropic` and
`bedrock`: a direct vendor holds one cache and has no route to choose.

## Prose corrected in the same change

Three places said routing is left to the gateway, which this change makes
false:

- `ZaiProvider::upstream_pin` and `session_id` field docs, and
  `with_upstream_pin`.
- `config/providers.rs`'s `ProviderConfig::upstream_pin` — the comment the
  issue quotes as documenting the empty default as a deliberate choice.
- `engine_config.rs`'s `unpinned_gateway_notice` and its doc: the notice now
  says Stella asks the gateway to stay put, that it may still fall back, and
  that `upstream_pin` is how you refuse the fallback. Its existing test asserts
  on `no upstream is pinned` and `upstream_pin`, both kept.

## Layout

New logic landed in `crates/stella-model/src/zai/upstream.rs` rather than in
`zai.rs`, which is 1391 lines against a 1500 ceiling. `zai.rs` is unchanged in
length: the `OpenRouterProviderPin` struct moved out with the new code.
`crates/stella-model/src/zai/tests.rs` is a grandfathered god file and was not
touched.

## Verification run locally

`cargo fmt --all`, `make guards-fast` (green: file-size, prose including the
new file's reading grade, line-citations, doc-links, lockfile-sync,
dead-code-allows, fmt --check). No cargo build, test, clippy or doc was run on
this machine, per CLAUDE.md. CI is the evidence.

Tests deleted: None.

Residue filed: #5692 (measure the cold-call share on a live turn), #5693
(check the gateway accepts the display name it reports).

Closes #4172

## Summary by Sourcery

Keep OpenRouter sessions on the upstream that served the first successful call while preserving strict operator pins and safe fallback behavior.

New Features:
- Re-ask OpenRouter for the upstream that served the session’s first successful response on subsequent calls, while allowing fallback routing.
- Preserve explicit operator upstream pins as strict, no-fallback routing preferences.

Bug Fixes:
- Prevent gateway session routing changes from unnecessarily discarding prompt-cache benefits after the first call.

Enhancements:
- Keep learned routing limited to OpenRouter endpoints and support learning from both streaming and unary delivery paths.
- Extend provider parity metadata and witness validation to cover gateway upstream routing behavior.
- Update configuration, provider, and engine notices to describe learned routing and fallback semantics accurately.

Tests:
- Add wire-level coverage for learned upstream reuse, operator-pin precedence, and preventing routing fields from reaching direct endpoints.

### DIFF
--- a/crates/stella-cli/src/config/providers.rs
+++ b/crates/stella-cli/src/config/providers.rs
@@ -29,9 +29,12 @@ pub struct ProviderConfig {
     /// whose models are whatever the user's endpoint actually serves.
     pub seeded: bool,
     /// Upstreams a *gateway* provider is pinned to, in preference order, from
-    /// settings.json's `upstream_pin`. Empty for every built-in row: routing
-    /// is the gateway's choice until an operator says otherwise, and only a
-    /// measured comparison needs it fixed.
+    /// settings.json's `upstream_pin`. Empty for every built-in row, which
+    /// does not mean unrouted: the shared adapter then asks for whichever
+    /// upstream served the session's first answer and still lets the gateway
+    /// fall back (`stella_model::zai`'s `upstream` module). Setting this is
+    /// the stronger ask — one upstream or a refused call — which is what a
+    /// measured comparison needs and what an interactive session does not.
     pub upstream_pin: &'static [String],
 }
 
--- a/crates/stella-cli/src/engine_config.rs
+++ b/crates/stella-cli/src/engine_config.rs
@@ -502,12 +502,16 @@ pub fn boot_notices(cfg: &crate::config::Config) -> Vec<String> {
 /// not TTL expiry: median idle before a miss was 0.3s against 0.2s before a
 /// hit, and calls alternated hit/miss inside the same second.
 ///
-/// The default stays unpinned, because routing is the gateway's choice until
-/// an operator says otherwise and only a measured comparison needs it fixed
-/// (see `config::providers::ProviderConfig::upstream_pin`). What changes is
-/// that the choice stops being invisible — the same honest-degradation
-/// contract [`unsupported_effort_notice`] applies to a dropped reasoning
-/// effort. A silent default nobody can see is not a default anyone chose.
+/// The adapter narrows this on its own: from the first answer on it asks the
+/// gateway for whichever upstream served it, with fallbacks still allowed. So
+/// this line is about what that cannot do — the first call of every session
+/// is routed by the gateway, and a fallback may still move the rest. Only
+/// `upstream_pin` refuses one (see
+/// `config::providers::ProviderConfig::upstream_pin`), which is what a
+/// measured comparison needs. What the line buys is that the choice stops
+/// being invisible — the same honest-degradation contract
+/// [`unsupported_effort_notice`] applies to a dropped reasoning effort. A
+/// silent default nobody can see is not a default anyone chose.
 ///
 /// [`CACHE_POSTURE`]: stella_model::provider_parity::CACHE_POSTURE
 pub fn unpinned_gateway_notice(
@@ -526,9 +530,10 @@ pub fn unpinned_gateway_notice(
         return None;
     }
     Some(format!(
-        "{provider_label} is a gateway and no upstream is pinned — it may route two turns of \
-         this session to different upstreams, and a prompt cached on one is a full-price miss \
-         on the next. Set `upstream_pin` in settings.json to fix the route."
+        "{provider_label} is a gateway and no upstream is pinned. Stella asks it to keep this \
+         session on whoever serves the first call, but it may still fall back, and a prompt \
+         cached on one upstream is a full-price miss on the next. Set `upstream_pin` in \
+         settings.json to refuse the fallback."
     ))
 }
 
--- a/crates/stella-model/src/provider_parity.rs
+++ b/crates/stella-model/src/provider_parity.rs
@@ -119,6 +119,7 @@ pub(crate) fn adapter_sources() -> &'static [&'static str] {
         include_str!("zai/tests/openrouter_stream.rs"),
         include_str!("zai/tests/stream_fallback.rs"),
         include_str!("zai/tests/stream_frame.rs"),
+        include_str!("zai/tests/upstream_pin.rs"),
         include_str!("zai/tests/vision.rs"),
         include_str!("zai/tests/zai_effort.rs"),
     ]
@@ -136,6 +137,17 @@ pub enum CachePosture {
         /// Name of the test function that proves the marker reaches the
         /// wire; checked for existence by this module's tests.
         witness: &'static str,
+        /// Name of the test proving this provider asks for a *particular*
+        /// cache, where sending the marker is not the whole story: a gateway
+        /// fronts several vendors holding several caches, so the marker can
+        /// reach the wire and still buy nothing.
+        ///
+        /// Two witnesses rather than one because they are two facts with two
+        /// kinds of evidence, the split `ParallelToolCallPosture` already
+        /// makes for the same reason — collapsing them would let a row claim
+        /// more than it can show. `None` for a direct vendor, which holds one
+        /// cache and has no route to choose.
+        routing_witness: Option<&'static str>,
     },
     /// The provider caches implicitly — nothing to send, but the adapter
     /// must PARSE the provider's hit telemetry or cached tokens bill at the
@@ -167,6 +179,7 @@ pub static CACHE_POSTURE: &[(&str, CachePosture)] = &[
                         the pair reaches the wire when configured, and the default window \
                         stays byte-identical",
             witness: "request_serializes_both_cache_breakpoints",
+            routing_witness: None,
         },
     ),
     (
@@ -174,6 +187,7 @@ pub static CACHE_POSTURE: &[(&str, CachePosture)] = &[
         CachePosture::OptIn {
             mechanism: "Converse cachePoint blocks, gated to supporting model families",
             witness: "complete_sends_cache_points_for_claude_models",
+            routing_witness: None,
         },
     ),
     (
@@ -186,8 +200,20 @@ pub static CACHE_POSTURE: &[(&str, CachePosture)] = &[
                         session_id is a HINT the gateway may decline, not a pin: measured \
                         over one 17-minute session it held for the last 17 calls and not \
                         the first 32, leaving 20 of 69 calls reading zero cached tokens \
-                        for 54% of the spend. Only a non-empty upstream_pin \
-                        (provider.order + allow_fallbacks:false) actually fixes the route. \
+                        for 54% of the spend; a 174-call turn on one slug in this repo's \
+                        own store was served by three vendors, the route moving 12 times, \
+                        with the id sent on every request. So the adapter reads the served \
+                        vendor off the response (top-level provider, on the stream's chunks \
+                        and on the unary body alike) and asks for it again on every later \
+                        call as provider.order with allow_fallbacks:true — a warm cache is \
+                        worth asking for and not worth failing a call over. An operator's \
+                        upstream_pin outranks it and keeps allow_fallbacks:false, which is \
+                        what a measured run needs. Whether the gateway routes on the \
+                        display name it reports, when that name is handed back in \
+                        provider.order, is a declared gap: a wiremock accepts any string, \
+                        so the round trip needs a live call to settle, and \
+                        allow_fallbacks:true is what makes an unmatched name cost the \
+                        saving rather than the call. \
                         The gateway honours the root-level placement for Anthropic routes, \
                         verified from recorded field usage rather than a live probe \
                         (#1854): of 1,684 anthropic/* calls this adapter made between \
@@ -198,6 +224,9 @@ pub static CACHE_POSTURE: &[(&str, CachePosture)] = &[
                         the gateway those reads could not occur if the root field were \
                         dropped",
             witness: "openrouter_identity_sends_root_level_cache_control",
+            routing_witness: Some(
+                "a_gateway_is_reasked_for_the_upstream_that_served_the_first_call",
+            ),
         },
     ),
     (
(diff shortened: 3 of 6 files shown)


macanderson:
Automated PR sweep — the `dod` blocker is cleared, and one note on what still has to happen for the check to show it.

**All ten of #4172's DoD items are satisfied by this diff.** I checked each against the tree at `de2d496` rather than taking the PR body's word for it, and ticked them on the issue with the evidence recorded item by item. The two worth calling out, because they are the two a reader would most reasonably doubt:

- **"Both the streamed chunks and the unary body."** Satisfied in both places, not one: `zai.rs` calls `served_upstream.learn(...)` on the unary result inside the `recovery.use_unary()` arm *and* on the streamed `outcome`. The stream→unary fallback is what makes the first call load-bearing rather than redundant.
- **"The served upstream reaches the durable record, or the PR says where it stops."** It reaches it, so the escape clause is not needed — and it predates this PR, which is why the diff does not show it. `step_receipt.upstream_provider` (schema v34, #3054) is written by `stella-store/src/receipts.rs`'s `INSERT OR REPLACE INTO step_receipt (…, upstream_provider, …)` and read back by its `SELECT … upstream_provider …`, with `a_receipts_upstream_provider_is_readable_without_the_event_stream` covering the round trip.

Sourcery's assessment agrees independently: three objectives for #4172, all met, no unmet rows.

**What the check will still show for now.** `dod-check.yml` fires on pull-request events, not on edits to the linked issue, so the run at 00:43 read the boxes in their pre-tick state and the red `dod / dod` on this head is that stale run. It clears on the next `synchronize` or `edited` event on this PR — the next push will do it. I did not manufacture one with a label toggle: SCR-005 keeps labels with the triage agent, and a no-op push to turn a check green is the shortcut CLAUDE.md forbids.

**Also on this head:** I merged main in at `de2d496` to pick up #5688, which repairs the `guard self-tests` break (`scripts/test-releases-published-pagination.sh` had pinned its fail-side to `origin/main`, so it went red on every PR the moment its own fix landed). That failure was never this PR's.

Not asking for a change to the diff.

---

Issue #5696, opened 2026-09-03
A batched read_file reports no line coverage, so the deck shows no read size for it


macanderson:
## The problem

`read_file` returns structured `data` — `{"lines_shown": N, "lines_total": M}`
— beside its prose, and the deck's read head resolves that producer instead of
parsing the footer back out of a capped payload. The **batch** spelling returns
none.

`crates/stella-tools/src/read.rs`'s `execute` ends the plural path with
`ToolOutput::ok(rendered)` and says why: line coverage is a per-file fact, and
one `shown`/`total` pair over an interleaved batch would attribute every file's
lines to none of them. That reasoning is right about a single pair. It leaves
the batch measuring nothing at all, which is the wrong end of the trade — a
turn that reads ten files in one call shows a blank read size on every surface
that uses it.

## Where

- `crates/stella-tools/src/read.rs` — the plural arm of `ReadFile::execute`,
  and `read_section`, which already computes `shown`/`total` per target.
- `crates/stella-tui/src/views/transcript_source.rs` — `kind_for`'s
  `"read_file" => EventKind::Read { lines: read }` and the `read_size` field
  it comes from.
- `crates/stella-tools/src/read/tests.rs` —
  `a_read_reports_its_line_coverage_as_structured_data` is the single form's
  test and the shape to follow.

## Constraints already discovered

- Whatever shape is chosen enters loop comparison: `comparable_output` keeps
  `data` in the comparison, so it must be deterministic for a given file state
  and window or it re-blinds the loop detector the footer strip exists for.
- The single form's `data` shape is a consumed contract and must not change.
  A per-file array under a new key is the obvious candidate; a summed pair is
  the shape the current comment rules out.
- The deck side has to learn the new shape, or the change buys nothing.

## How to verify

Call `read_file` with `{"files": [{"path": "a.rs"}, {"path": "b.rs"}]}` and
inspect `ToolOutput::Ok { data, .. }`. Today it is `None`.

## Definition of done

- A batched read reports per-file line coverage as structured data, in a shape
  a consumer can attribute to a path.
- The deck's read head renders a size for a batched read.
- The single form's `data` is byte-identical to what it is today.
- A witness test that fails without the change, plus a test pinning that the
  shape is deterministic for a given file state and window.

macanderson:
Folded into batch #6291 (2026-09-06 audit): fixed together by crate.

Issue #5695, opened 2026-09-03
Nothing ever writes an R row to files_touched, so the ledger cannot say what a turn read


macanderson:
## The problem

`.stella/private/store.db`'s `files_touched` table has a CRUD alphabet with an
`R` op for a read. Nothing in the shipping tree ever writes one.

Every producer of a `FileTouchRow` is in `crates/stella-cli/src/turn_files.rs`
(`file_touch_row` and `own_touch_row`), and both derive the op from a
`JournalChange`/`OwnChange` kind, which has only `Created`, `Modified` and
`Deleted`. `crates/stella-cli/src/turn_files.rs`'s own comment says
`FileChangeKind::Read` is "deliberately unreachable here: a tree snapshot"
cannot see a read. The string `"R"` appears in the tree only in
`crates/stella-store/src/tests.rs` fixtures.

So the audit ledger under-reports what a turn read, for every spelling of
every read — `read_file` single, `read_file` batched, and a shell read alike.
Issue 4151 lists this as one of five read-side mechanisms a shell read
bypasses; it turns out the mechanism does not exist for the tool either.

## Why it matters

The reflection loop and `stella stats` read `files_touched`. A turn that read
forty files and wrote two is indistinguishable from a turn that read nothing.
Anything asking "what did this turn look at" gets the wrong answer, and a
verification or review surface built on the ledger inherits it.

## Where

- `crates/stella-cli/src/turn_files.rs` — the two `FileTouchRow` producers.
- `crates/stella-tools/src/read.rs` — `ReadFile::read_section` is where a read
  is known to have happened, per target, for both spellings.
- `crates/stella-tools/src/call_measure.rs` — the `CallMeasure` port and the
  argument for why a *tree* reading cannot answer this. A read producer is a
  different mechanism, not a widening of that one.
- `crates/stella-store/src/lib.rs` — `FileTouchRow`, `record_files_touched`.

## Constraints already discovered

- `stella-tools` must not depend on `stella-store` or `stella-cli`; whatever
  carries the read to the ledger goes through a port, the way `CallMeasure`
  does (AGENTS.md invariant 1).
- `record_files_touched` has a UNIQUE `(execution_id, path)`, so a path read
  and then written in one execution has one row. Decide what the `ops` string
  holds for that case before writing code.
- A batched read must produce one entry per file it named, not one per call.

## How to verify

Run a turn that reads a file and changes nothing, then
`select path, ops from files_touched where execution_id = <n>`. Today it
[...]

Issue #4151, opened 2026-08-21
Models read files with shell commands instead of the read tool, bypassing every read-side safety check


macanderson:
## What happens

On a wide code-orientation task the model routes almost all of its **reads** through `bash` (`sed -n 'A,Bp'`) instead of `read_file`. Every read-side safety mechanism in `crates/stella-tools/src/read.rs` is bypassed as a result, because those mechanisms live in the tool, not in the read.

This is **not** an `edit_file` defect and **not** a `read_file` defect — `read_file` was called twice in the execution below and succeeded both times. The tool is fine; it is being out-competed.

## Evidence

`.stella/private/store.db`, execution 178 (provider `openrouter`, model `moonshotai/kimi-k3`):

```sql
select
  sum(name='bash' and args_json like '%sed -n%'
                 and args_json not like '%| sed -n%') as sed_range_reads,
  sum(name='read_file')                              as read_file_calls,
  sum(name='read_file' and args_json like '%"files"%') as read_file_batched,
  sum(name='bash' and args_json like '%sed -i%')     as sed_in_place_edits,
  sum(name='edit_file')                              as edit_file_calls
from tool_calls where execution_id=178;
-- 24 | 2 | 0 | 0 | 19
```

The `not like '%| sed -n%'` clause matters: `sed` after a pipe is a pager over
`grep` output and reads no file. Counting those inflated the shell side by 43%
on one of the executions measured later in this thread. Count `cat`, `head` and
`tail` the same way when re-running this; all three were zero every time so
far, so `sed` has been the whole of the leak.

Reads: **24 shell range-reads vs 2 `read_file` calls**. Writes: **19 `edit_file`, 0 `sed -i`** — the write-side guidance in the `bash` schema and `prompt.rs` is obeyed exactly. Only the read side leaks. (The counts first filed here were 23 and 14, taken from the same execution with the uncorrected query above.)

It is not a fallback after failure: `read_file` recorded zero errors in that execution, and the very first tool call after `task_create` (seq 2) was already `sed -n '700,1000p' …/engine.rs`, before any `read_file` had ever run. The `MAX_UNCHANGED_READS` ceiling cannot be the cause either — it had not fired.

Not universal:

| execution | sed reads | read_file | edit_file |
|---|---|---|---|
| 171 | 0 | 49 | 0 |
| 173 | 0 | 169 | 0 |
| 178 | 23 | 2 | 14 |

[...]

PR #5697, opened 2026-09-03
fix(stella-core): restore every file a batched read_file put in context

### LINKED ISSUE #5701: The working set cannot see a file a batched read_file put in context
## The problem

`read_file` accepts a batch — `files: [{path, offset?, limit?}]` — so a turn
can ask for several files, or several ranges, in one call. The engine's
working-set restoration cannot see any of them.

`crates/stella-core/src/restore.rs`'s `collect_working_set` reads a recorded
read call's target out of `input["path"]`. A batched call carries no path of
its own, so the `let ... else { continue }` fires and every file the turn read
that way is dropped from the working set. When the overflow summarizer splices
that span away, the restoration puts none of it back — the exact loss
`restore.rs` exists to prevent.

The kept-tail walk has the mirror bug. It builds `tail_paths` from the same
field, so a file the tail re-read in a batch still reads as lost. It is then
restored on top of bytes already in context, spending headroom another file
needed.

## Repro

`collect_working_set` with a span holding one assistant message whose
`read_file` call carries
`{"files": [{"path": "a.rs"}, {"path": "b.rs"}]}` and a successful tool result
for it returns an empty `WorkingSet`.

## Files

- `crates/stella-core/src/restore.rs` — `collect_working_set`, `READ_TOOL`,
  `READ_PATH_PARAM`.
- `crates/stella-core/src/driver/restore.rs` — `restore_working_set`, which
  replays `SpanRead::input` verbatim once per restored path.
- `crates/stella-tools/src/read.rs` — `FILES_KEY`, the batch arm of
  `ReadFile::execute`.
- `crates/stella-tools/src/read/tests.rs` —
  `the_read_tool_is_the_one_the_engines_restoration_replays` is the pin that
  ties the two crates' spellings together.

## Constraints

- The replay runs one input per restored path, so a batch has to become one
  working-set entry per element carrying that element's own `offset`/`limit`.
  Replaying the whole batch under each path would read every file once per
  path and render all of them under each heading.
- Order matters: a later range of the same file inside one call is the fresher
  view, the way a later call is.
- `stella-core` does no I/O and takes no dependency on `stella-tools`; the
  spellings are shared as constants plus a pin test, the way `READ_TOOL`
  already is.

## Definition of done

- [x] `collect_working_set` collects every file a batched `read_file` call
      named, one entry per element, carrying that element's own window.
- [x] A file the kept tail re-read in a batch is excluded from the working
      set.
- [x] A batch whose recorded result was an error contributes nothing, for
      every file it named.
- [x] The batch key is declared in `stella-core` and pinned to the tool's own
      spelling by a test in `stella_tools::read`.
- [x] Witness tests covering all of the above, green in CI on the branch; the
      mechanism by which each fails on `main` is stated in the pull request.

## Related

Found while working issue 4151, which added the batch form. That issue stays
open for its own bench re-measurement.

macanderson:
## What

`read_file` can be asked for several files, or several ranges, in one call
(`files: [{path, offset?, limit?}]`). The engine's working-set restoration
could not see any of them.

`crates/stella-core/src/restore.rs`'s `collect_working_set` reads a recorded
read call's target out of `input["path"]`. A batched call carries no path of
its own, so every file a turn read through `files` was dropped from the
working set — and when the overflow summarizer spliced that span away, the
restoration put none of it back. The kept-tail walk had the mirror bug: a file
the tail re-read in a batch still read as lost, so it was restored on top of
bytes already in context, spending headroom another file needed.

This is the same family as the five bypassed mechanisms the issue lists. The
batch form was added to stop the model reaching for `sed`, and it quietly
reintroduced one of the losses through the front door.

## Why this shape

- `READ_FILES_PARAM` is declared beside `READ_TOOL` and `READ_PATH_PARAM`, and
  pinned to the tool's own spelling by
  `the_read_tool_is_the_one_the_engines_restoration_replays` — the tie that
  already exists for the tool name and the path parameter. A drift there is
  the working set silently ceasing to restore anything a batch read.
- A batch yields **one working-set entry per element**, carrying that
  element's own `offset`/`limit`. The engine half replays `SpanRead::input`
  verbatim once per restored path, so replaying the whole batch under each
  path would read every file once per path and render all of them under each
  heading. Per-element inputs also keep the standing contract that a windowed
  read restores the window the model was looking at.
- Order is preserved, so a later range of the same file **inside one call**
  wins the way a later call does — which is what "three ranges across two
  files" needs to mean.
- `read_targets` borrows rather than owns: the tail walk only wants paths.

Exemplar followed: none outside this tree — the shape is `restore.rs`'s own
existing `READ_TOOL` / `READ_PATH_PARAM` declaration plus its pin test in
`stella_tools::read`, extended to the second spelling rather than reinvented.

The pointer names the plural key `paths`. The tree already ships it as
`files`, landed by pull request 4166, and `write_file` and `delete_file` declare the same
`files` key through the shared `crate::batch::plural_schema`. Renaming
`read_file`'s alone would give the four file tools three different plural
spellings for one idea, for no gain — so the key stays `files` and the fix is
the consumer that never learned it. Everything else in the pointer is
implemented or was already present:

- one-item `path`/`offset`/`limit` sugar: present.
- one `ReadLedger` entry, byte caps, paging footer, one section per range with
  the path as its header: present, and now regression-covered end to end (see
  the second witness below).
- `files_touched` `R` event: there is **no live producer** of an `R` row for a
  read anywhere in the tree today — `turn_files.rs` writes only `C`/`U`/`D`,
  and `"R"` appears solely in `stella-store`'s own tests. Nothing was invented
  here; that gap is filed as residue below.
- schema description now leads with "Read one or more files or ranges in a
  single call"; `docs/tools/read_file.toml` and the website tool card follow.
- `crates/stella-cli/src/agent/prompt.rs`'s read-side sentence is already
  imperative and already names `cat`/`sed -n`/`head`/`tail` and the batch key.
  Left byte-identical.

## Witnesses

Neither could be run here (CLAUDE.md: CI builds and tests, this laptop does
not) — the mechanism by which each fails on the old code is below, and CI is
the evidence.

**`every_file_a_batched_read_named_is_working_set`**
(`crates/stella-core/src/restore.rs`). One `read_file` call with three ranges
across two files, then `collect_working_set`. On `main` the span walk asks for
`input["path"]`, a batched call has none, the `let … else { continue }` fires
for the only call in the span, and the returned `WorkingSet` is **empty** — so
`assert_eq!(paths, ["a.rs", "b.rs"])` fails on the first element of an empty
vec. With the change both files are collected, freshest first, and each one's
replay input is the single-form `{path, offset, limit}` of the element that
named it.

**`a_batched_tail_reread_is_not_restored`** (same file). The kept tail re-reads
`kept.rs` in a batch. On `main` `tail_paths` stays empty for that call, so
`kept.rs` survives the `retain` and the assertion `["lost.rs"]` sees
`["lost.rs", "kept.rs"]`.

**`a_failed_batched_read_contributes_nothing`** (same file) covers the other
direction: a batch whose recorded result was an error contributes nothing for
*every* file it named.

**`a_batch_read_leaves_the_drift_oracle_able_to_attribute_a_miss`**
(`crates/stella-tools/src/read/tests.rs`) is the issue's own DoD clause, end to
end: three ranges across two files in one call, then a failed `edit_file`
against one of them. It asserts the miss is **not** reported as "no read of
this file is recorded this session" and is attributed as an unchanged file.
This one **passes on `main`** — the ledger half shipped in pull request 4166 — so it is a
regression guard for the DoD clause, not a witness. Said plainly rather than
claimed as proof of new behaviour.

The pin test now also asserts the batch key and its per-element `path` keep
the spellings the engine replays.

## Verification

`cargo fmt --all` and `make guards-fast` are green locally. No workspace build
or test ran on this machine; CI is the evidence for the four tests above.

Tests deleted: None.

## What remains

The issue's third DoD clause is a number, not a design: re-measure the
sed-to-`read_file` ratio on a comparable wide-orientation task and report it.
That needs a real run against a real model and is not something to start
incidentally, so #4151 stays open for it — `Refs #4151`. The issue's
own SQL is still the right query, and the two extra counts a previous comment
asked for (how many `read_file` calls carry `files`, and distinct timestamps
per tool call) still distinguish "the batch key is being used" from "the
prompt sentence moved it".

## Issues

The restoration blindness this PR repairs is scoped as its own issue, filed
with the DoD this PR satisfies:

- `Closes #5701` — the working set cannot see a file a batched `read_file` put
  in context.

Residue filed and not fixed here:

- https://github.com/macanderson/stella/issues/5695 — no live producer of a `files_touched` `R` row
  for a read, so the audit ledger under-reports what a turn read whichever
  spelling it used.
- https://github.com/macanderson/stella/issues/5696 — the plural `read_file` form returns no
  structured `data`, so the deck's read head measures nothing for a batched
  read.

Closes #5701
Refs #4151

## Summary by Sourcery

Fix working-set restoration so batched `read_file` calls correctly preserve all files and ranges in context.

Bug Fixes:
- Restore every file and range named by batched `read_file` calls in the working set, including correct handling of batched reads in kept-tail tracking and failed calls.
- Preserve per-file read windows and freshness ordering when restoring batched reads.

Enhancements:
- Pin the `read_file` batch schema spelling to the restoration engine and update tool descriptions to clearly document multi-file and multi-range reads.

Documentation:
- Update the read tool documentation and website card to describe batched file and range reads.

Tests:
- Add regression coverage for working-set restoration, kept-tail handling, failed batches, schema spelling, and drift attribution for batched reads.

### DIFF
--- a/crates/stella-core/src/restore.rs
+++ b/crates/stella-core/src/restore.rs
@@ -20,9 +20,9 @@
 //!
 //! - **Files read inside the summarized span** ([`collect_working_set`]):
 //!   the most recent read per path, most-recent-first, capped at
-//!   [`RESTORE_MAX_FILES`]. Content is re-read fresh from disk at restore
-//!   time —, so an external edit between read and restore
-//!   surfaces the *current* bytes rather than resurrecting a stale copy.
+//!   [`RESTORE_MAX_FILES`]. One call names one file or a list of files and
+//!   ranges ([`READ_FILES_PARAM`]), and every file counts. Content is re-read
+//!   fresh at restore time, so an edit in between surfaces the current bytes.
 //! - **The full body of any active invoked skill**
 //!   ([`crate::skills::invoke::ActiveSkillInvocations`] is the tracking
 //!   half): procedure text the model is executing must survive verbatim,
@@ -108,6 +108,42 @@ pub const READ_TOOL: &str = "read_file";
 /// span's recorded calls and to synthesize the replay input for a path
 /// re-collected from a prior restoration message.
 pub const READ_PATH_PARAM: &str = "path";
+
+/// The read tool's plural key: an array of `{path, offset?, limit?}` objects,
+/// one per file or range, sent instead of the single-target fields.
+///
+/// A batched call carries no path of its own, so [`READ_PATH_PARAM`] alone
+/// finds none of the files it read. Both keys are therefore read, or the
+/// splice destroys a batch's content with nothing to put back. Tied to the
+/// tool's own spelling by a pin test in `stella_tools::read`, the same way
+/// [`READ_TOOL`] is.
+pub const READ_FILES_PARAM: &str = "files";
+
+/// Every file one recorded read call asked for, each paired with the input
+/// that re-reads that file on its own.
+///
+/// A batched call yields one entry per element, carrying that element's own
+/// `offset`/`limit`, so the replay restores the window the model was looking
+/// at. Replaying the whole batch under each path instead would read every
+/// file once per path and render all of them under each heading.
+///
+/// Borrowed rather than owned: the tail walk only wants the paths, and the
+/// span walk clones the input it keeps.
+fn read_targets(input: &serde_json::Value) -> Vec<(&str, &serde_json::Value)> {
+    if let Some(items) = input.get(READ_FILES_PARAM).and_then(|v| v.as_array()) {
+        return items
+            .iter()
+            .filter_map(|item| {
+                let path = item.get(READ_PATH_PARAM)?.as_str()?;
+                Some((path, item))
+            })
+            .collect();
+    }
+    match input.get(READ_PATH_PARAM).and_then(|v| v.as_str()) {
+        Some(path) => vec![(path, input)],
+        None => Vec::new(),
+    }
+}
 
 /// Ceiling on restored files per round. Eight matches the workspace's
 /// standing notion of "the recent work the model is actively reasoning over"
@@ -278,23 +314,25 @@ pub fn collect_working_set(
                     if call.name != READ_TOOL || failed_calls.contains(call.call_id.as_str()) {
                         continue;
                     }
-                    let Some(path) = call.input.get(READ_PATH_PARAM).and_then(|v| v.as_str())
-                    else {
-                        continue;
-                    };
-                    seq += 1;
-                    match reads.iter_mut().find(|t| t.path == path) {
-                        // A later read of the same path is the fresher view:
-                        // its input (offset/limit included) wins outright.
-                        Some(tracked) => {
-                            tracked.input = call.input.clone();
-                            tracked.seq = seq;
+                    // One call names one file or several; a batch's elements
+                    // are ordered, so a later range of the same file inside
+                    // one call wins the same way a later call does.
+                    for (path, input) in read_targets(&call.input) {
+                        seq += 1;
+                        match reads.iter_mut().find(|t| t.path == path) {
+                            // A later read of the same path is the fresher
+                            // view: its input (offset/limit included) wins
+                            // outright.
+                            Some(tracked) => {
+                                tracked.input = input.clone();
+                                tracked.seq = seq;
+                            }
+                            None => reads.push(TrackedRead {
+                                path: path.to_string(),
+                                input: input.clone(),
+                                seq,
+                            }),
                         }
-                        None => reads.push(TrackedRead {
-                            path: path.to_string(),
-                            input: call.input.clone(),
-                            seq,
-                        }),
                     }
                 }
             }
@@ -336,10 +374,12 @@ pub fn collect_working_set(
     let mut tail_slugs: HashSet<String> = HashSet::new();
     for message in kept_tail {
         for call in &message.tool_calls {
-            if call.name == READ_TOOL
-                && let Some(path) = call.input.get(READ_PATH_PARAM).and_then(|v| v.as_str())
-            {
-                tail_paths.insert(path.to_string());
+            if call.name == READ_TOOL {
+                tail_paths.extend(
+                    read_targets(&call.input)
+                        .into_iter()
+                        .map(|(path, _)| path.to_string()),
+                );
             }
         }
         if message.role == MessageRole::User {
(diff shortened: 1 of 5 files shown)


macanderson:
Autonomous PR sweep. Two checks were red here; they have different causes and only one of them is this PR's.

## `guard self-tests` — not this PR's, and now fixed

Red because this branch carried the pre-#5688 copy of `scripts/test-releases-published-pagination.sh`, whose `MAIN` arm asserts that `origin/main`'s `check-releases-published.sh` hits its 1000-item cap. Main no longer does, so the fail-side passes and the arm reports a failure. Reproduced against this PR's head (`c349308`) at the time of writing:

```
FAIL MAIN main's script hits its 1000-item cap on 1001 releases and errors — exit 0, got:
     check-releases-published: OK — every v* tag older than 5256000m has a published release
ok   NEW this branch's script pages through all 1001 releases and reports clean
ok   NEW reports no truncation or page-sanity error on 1001 releases
passed 2, failed 1
```

That is #5681, fixed on `main` by #5688 (`3aea7c4`), which landed after this branch was cut. I have merged `main` in, which carries the fix. #5694 went green on `guard self-tests` after the same merge, so the remedy is confirmed rather than assumed.

## `dod` — a rule conflict, not a defect in this PR

I verified #4151's definition of done against this head before commenting:

| DoD clause | State |
|---|---|
| Witness test failing on `main` such that the ledger has an entry for every file read | present — `every_file_a_batched_read_named_is_working_set`, `a_batched_tail_reread_is_not_restored`, `a_failed_batched_read_contributes_nothing` |
| Drift oracle attributes a failed match correctly for a file read through the chosen path | present — `a_batch_read_leaves_the_drift_oracle_able_to_attribute_a_miss`, which this PR describes plainly as a regression guard that passes on `main` rather than claiming it as a witness |
| Re-measure the sed→`read_file` ratio and report it | **not met** — needs a live model run |

#4151 says *"this issue is only closed by a number, not by a design."* So `Refs #4151` is the correct reference and `Closes` would close an issue whose DoD is unmet — the exact thing SCR-003 exists to prevent.

`dod` fails anyway, because it accepts only a closing reference or the `no-issue` label:

```
- This PR links no issue. Add a closing reference (`Closes #123`) to the description,
  or apply the `no-issue` label if the change is genuinely trivial (SCR-003).
```

Both escape hatches are wrong here — `Closes` closes prematurely, and `no-issue` asserts a 258-line change across 5 files is trivial and issue-less. Meanwhile AGENTS.md sanctions the third state directly: *"`Refs #N` when a PR advances an issue without finishing it — `Refs` deliberately does not close."*

**I have not applied either label or edited the reference.** Doing so would launder a rule conflict into a false statement in the permanent record. The gap is filed as macanderson/oxagen#2640 against `dod-check.yml`, which is the `workflow_call` source all five repos consume. Until that lands, this PR cannot go green as written, and the remaining call — wait for the guard fix, or split the measurement clause into its own issue so this one can legitimately close a narrowed #4151 — is the maintainer's, not mine.

---

Issue #5706, opened 2026-09-03
The spend gate does not price the paths that buy an extra model call


macanderson:
## The problem

`crates/stella-core/tests/spend_gate.rs` (added for #4223) pins what a turn
buys over the raw step loop, but only for the four plain scenarios it names: a
one-step answer, a two-tool step, a retry after a 429, and a turn the loop
detector stops. Every one of them takes exactly one model call per step.

The step loop has other paths that buy model calls a step did not ask for, and
none of them is pinned by anything:

- the overflow summarizer (`crates/stella-core/src/driver/overflow_recovery.rs`)
  — a forced compaction pass spends an extra call at `call_seq` 1;
- provider fallback (`crates/stella-core/src/driver/model_fallback.rs`) — an
  exhausted retry ladder re-asks a second provider;
- output-budget recovery
  (`crates/stella-core/src/driver/output_budget_recovery.rs`) — a clamped
  ceiling re-asks the same call;
- the parked wait for a sustained 429 or a 529
  (`crates/stella-core/src/driver/rate_limit.rs`) — bounded, and nothing pins
  how many attempts it can buy.

A change that makes one of these fire more often is exactly the cost
regression the spend gate exists to catch, and today it slips through: the
pinned scenarios never enter those arms.

## What done looks like

Extend the same file (or a sibling under `crates/stella-core/tests/`) with one
scenario per path above, each pinning the exact model-call count, tool-call
count and reported cost the way the existing ones do, with the assertion
naming the scenario. Same constraints as #4223: scripted `Provider` /
`ToolExecutor` ports only, no I/O in the engine (architecture rule 2), and the
operating rule that an intended change edits the pin in the same pull request.

Read `crates/stella-core/tests/spend_gate.rs` first — the harness, the flat
`0.25` per answered call (exact in binary, so summed prices pin cleanly) and
the `assert_spend` message are already there to reuse. The driver's own tests
show how to drive each arm: `crates/stella-core/src/driver/tests/audit_fixes.rs`
and the overflow tests script the summarizer, and `model_fallback.rs`'s tests
script an exhausted ladder.

## Definition of done

- [x] A pinned scenario for the overflow summarizer's extra call.
- [x] A pinned scenario for a provider fallback after an exhausted ladder.
- [x] A pinned scenario for output-budget recovery re-asking a clamped call.
- [x] A pinned scenario bounding what a parked 429 or 529 wait can buy.
- [x] Each failure names its scenario and prints the pin beside the new
[...]

Issue #4029, opened 2026-08-19
Epic: the five plugins that did not get built, and the verification path that ships nowhere


macanderson:
Spec: `doc:plugin-completion-plan` (`docs/spec/plugin-completion-plan.md`, added in #4028).

## The defect

`doc:pipeline-as-plugins` §3 names nine plugins that replace one staged pipeline
and one built-in autonomous loop. The built-in path was deleted on 2026-08-19
(#3865). Audited against the tree at `f9789df17`, **five of the nine are missing
or cannot work**:

| Plugin | State | Evidence |
| --- | --- | --- |
| `vera` | **not built** | `gh api repos/oxageninc/vera/contents` returns one entry, `README.md`; repo size 3 KB |
| `stella-candidates` | **not built** | no `plugins/stella-candidates` directory |
| `stella-selfdriving` | **declaration only** | `plugins/stella-selfdriving/` holds `plugin.toml` + `README.md`; no program, no `[runtime]`, `participation = "none"` |
| `stella-goal` | **built but inert** | `crates/stella-cli/src/wrapper_plugin.rs:519` documents "No `verifier` seat is bound" (#3838), so its one job degrades to `HostCallRefusal::Unavailable` |
| `verify-{rs,py,ts}` | **ungated here** | present in `macanderson/stella-examples`; `rg 'verify-rs' --glob '!docs/**'` returns no non-doc hit in this repository, so §9 rule 4's smoke check does not exist |

Built and working: `stella-research`, `stella-plan`.

## The consequence that outranks the rest

Composite, each premise verified:

1. `stella run --pipeline classic` is refused (`wrapper_plugin.rs:204`,
   `classic_removed_message`).
2. `AGENTS.md`'s opening states the only remaining verification path is an
   installed verification plugin, and names Vera as the reference one.
3. `oxageninc/vera` contains one file.

**Inference, labelled:** no build from this tree can obtain a witness-verified
turn by any route. `doc:pipeline-as-plugins` §7 records the deletion as a
deliberate maintainer call about *code*, and says extraction "is still real,
open work". This epic is that work.

## Definition of done

- **Met.** Every plugin `doc:pipeline-as-plugins` §3 lists exists, and each is
  graded by something that runs on a pull request. The six first-party ones —
  `stella-research`, `stella-plan`, `stella-goal`, `stella-witness`,
  `stella-candidates`, `stella-selfdriving` — are graded by harnesses in
  `crates/stella-runtime/tests/` and
[...]

PR #5708, opened 2026-09-03
test(stella-core): a per-PR spend gate over the raw step loop

### LINKED ISSUE #4223: The verification gate lost 2 of its 3 layers with #3865 and nothing replaced them — no per-PR check observes what the agent decides or spends
## The defect

`doc:verification-gate` describes "the three layers CI uses" to catch an
agent-loop change that ships green and makes the agent quietly worse or more
expensive. **Two of the three no longer exist**, and nothing replaced them.

Both lived in `crates/stella-pipeline`, deleted in #3865:

| Layer | Was | State |
|---|---|---|
| 1 — degradation gate | `crates/stella-pipeline/src/pipeline/tests/degradation_gate.rs` | **gone** |
| 2 — golden trajectories | `crates/stella-pipeline/src/pipeline/tests/golden.rs`, `make record-golden` | **gone** |
| 3 — benchmark arm | `bench/` | intact |

Verified on the current tree:

```
$ fd -HI "degradation_gate|golden" crates/
crates/stella-cli/src/memory/tests/golden_block.rs      # unrelated
$ rg -n "record-golden" Makefile
(no output)
```

The surviving half of Layer 2 is the wire schema only:
`docs/wire/agentevent.schema.json` asserted by
`crates/stella-protocol/tests/wire_contract.rs`, plus the `wire-schema` gate
step. That gates the event *contract*. It does not gate the event *trajectory*,
and nothing at all gates the decision policy or its price.

## Why this matters more than a stale doc

Layer 1 was the only per-PR check that could fail on **spend**. It pinned the
exact number of model calls by role and test-suite invocations per scenario, so
a change that left every verdict intact but bought an extra call failed with a
message naming it. There is now no cheap signal between an agent-loop change and
a Terminal-Bench run that costs real money and hours.

Layer 1 was also the only mechanism enforcing "state your intent as a diff line":
its operating rule was that an intended behaviour change updates the scenario's
expectation **in the same PR**, so a reviewer reads "this change makes flaky
flips escalate" instead of discovering it in production.

`doc:verification-gate` has been rewritten to say all of this plainly rather than
leave three layers implied (this issue's sibling change, Refs #3901). The
document is now accurate; the hole is still a hole.

## The decision this issue needs

Not obviously "rebuild them". The options:

1. **A verification plugin owns Layer 1.** Its oracle and declared verdict rule
   *are* the decision policy, so pinning them against scripted doubles is work on
   the plugin's own side of the wrapper socket. A host that neither runs the
   check nor re-checks the evidence (#3511) cannot gate on either. This makes it
   Vera's problem, and Vera is not built (#4029).
2. **Rebuild a host-side spend gate over the raw loop.** The raw step loop still
   makes model calls and still has a budget guard; a scenario matrix over
   `stella-core` with scripted providers could pin call counts without any
   verification machinery at all. This is the option that covers the *cost*
   regression case, which is the one with no other net.
3. **State the gap explicitly** in `doc:verification-gate` and
   `CONTRIBUTING.md`, so nobody believes `cargo test --workspace` observes
   agent behaviour.

Option 2 looks smallest and covers the failure mode with no other coverage;
option 3 is the floor and should happen if neither 1 nor 2 does.

## Definition of done

Option 2 was chosen: a scenario matrix over the raw step loop. The boxes below
are that option's terms.

- [x] `crates/stella-core/tests/spend_gate.rs` exists and drives one turn per
  scenario over a scripted `Provider` and a scripted `ToolExecutor`, adding no
  I/O to the engine.
- [x] Each scenario pins the exact model-call count, the exact tool-call count
  and the reported cost: a one-step answer, a two-tool step, a retry after a
  429, and a turn the loop detector stops.
- [x] A failure names the scenario and prints the pinned numbers beside the new
  ones.
- [x] The file header states the operating rule: an intended change edits the
  pin in the same pull request.
- [x] `doc:verification-gate` names the new layer and what it can and cannot
  claim.
- [x] `CONTRIBUTING.md` names the gate and the same operating rule.
- [x] The pull request records which option was chosen and why.
- [x] CI is green on the pull request that lands it.

## Constraints

- `doc:verification-gate` is already rewritten to the post-deletion state; start
  by reading it, not the git history.
- Cite documents by id (`doc:verification-gate`), not path — `make doc-links`.
- A scenario matrix rebuilt on `stella-core` must not add I/O to the engine
  (architecture rule 2) — scripted `Provider` / `ToolExecutor` ports, as
  `ScriptedProvider` already does.
- `make gate` stays green.

Refs #3865, #3901, #4029.

macanderson:
## What & why

A change to the agent loop can leave every answer the same and still buy an
extra model call. The check that used to fail on that was the staged
pipeline's degradation gate, deleted with `crates/stella-pipeline` in #3865.
Nothing per-PR has read price since, so the cheapest signal between "someone
changed the loop" and "we know what it cost" was a Terminal-Bench run.

This is the maintainer's option 2, taken as recorded on the issue: a scenario
matrix over the **raw step loop**, not over verification machinery. It needs
none — the engine does no I/O, so scripted ports are the whole harness
(architecture rule 2), and the raw loop already makes the model calls whose
count is the thing worth pinning. Options 1 (a verification plugin owns the
decision policy) and 3 (document the gap and stop) are not foreclosed: this
covers the cost regression, which is the failure with no other net, and the
decision-policy half stays a plugin's problem under #4029.

`crates/stella-core/tests/spend_gate.rs` drives one turn per scenario against
a scripted `Provider` and a scripted `ToolExecutor`, and pins the model calls,
the tool runs and the reported cost:

| Scenario | Model calls | Tool runs | Cost |
|---|---|---|---|
| one-step answer | 1 | 0 | one call |
| two-tool step | 2 | 2 | two calls |
| retry after a 429 | 2 | 0 | one call — the refused attempt bills nothing |
| loop-detector trip | 4 | 4 | four calls |

`model_calls` counts attempts rather than bills, so a retried 429 shows up as
the second call it is. A scripted call costs a flat `0.25`, which is exact in
binary, so a summed price is pinned rather than approximated.

The old gate's operating rule carries over, stated in the file header and in
CONTRIBUTING.md: an intended change edits the pin **in the same pull
request**, so a reviewer reads the new price as a diff line instead of
inferring it.

## The witness

The file is the witness, and its mechanism is the pin itself: each scenario
asserts an exact count, so any loop change that moves a count fails it with
the scenario named. Two one-line demonstrations, either of which turns it red
on this branch:

- In `driver/loop_escalation.rs`, `check_loop_detection`'s steer arm returning
  `Some(...)` instead of `None` (abort on first detection) drops the
  loop-detector trip from `model_calls: 4, tool_calls: 4` to `3, 3`.
- Making `retry_with_backoff_observed` give up on a `RateLimited` instead of
  retrying it turns the 429 scenario's `model_calls: 2` into `1` and its
  outcome into an abort.

It cannot fail on `main` in the usual fail→pass sense, because `main` does not
contain the file — this PR adds an observation that did not exist, which is
the issue's own definition of done. Its counts were derived from the loop's
own phase order (loop detection runs *before* the model call, and a steer
continues the same step) and cross-read against the existing driver tests that
already pin the same shapes: `stuck_loop_steers_once_then_aborts_on_re_detection`
pins the four tool calls, `retry_never_re_executes_a_tool_call` pins the
retry-does-not-re-execute rule and the cost of two answered calls.

Exemplar followed: `crates/stella-core/tests/parallel_dispatch.rs` — the same
public-API-only integration-test shape (scripted `Provider` + `ToolExecutor`,
`Engine::with_sleeper`, a `NoopSleeper`), and `driver/tests.rs`'s
`ScriptedProvider` for the script-with-repeating-tail idiom.

## The gate

- [x] `cargo fmt --all`
- [x] `make guards-fast` (prose, doc-links, line-citations, lockfile-sync, fmt)
- [ ] clippy / test — CI runs them; this laptop does not (CLAUDE.md)
- [x] Docs updated: `doc:verification-gate` gains the layer and says what it
      cannot claim; CONTRIBUTING.md names the gate and its operating rule
- [x] `Closes #4223` in this description and as a commit trailer

CONTRIBUTING.md's witness section also still told contributors to run
`stella run --pipeline classic`, which is refused outright since the same
deletion. Corrected in the paragraph this PR was already editing.

Tests deleted: None.

## Nothing left behind

- Filed: #5706 — the spend gate prices the plain scenarios only, so the arms
  that buy an *extra* model call (the overflow summarizer, provider fallback,
  output-budget recovery, the parked rate-limit wait) are still unpinned.
- Not filed, already tracked: the decision-policy half of the old Layer 1 —
  #4029 owns it, and it lands on a verification plugin's side of the wrapper
  socket rather than here.

Closes #4223

## Summary by Sourcery

Restore a lightweight per-PR observation for raw-loop cost regressions by pinning representative turn spending and documenting how to review intentional changes.

New Features:
- Add a per-PR spend gate that measures model attempts, tool executions, and reported cost across representative raw-loop scenarios.

Enhancements:
- Document the spend gate’s review rule and clarify that decision quality remains outside its scope.
- Update core testing documentation to describe public-API integration tests and correct the obsolete classic pipeline guidance.

Documentation:
- Document the spend gate in CONTRIBUTING.md, the core README, and the verification-gate specification, including its limitations and intended workflow.

Tests:
- Add scripted integration tests covering one-step answers, multi-tool turns, rate-limit retries, and stuck-loop termination with pinned spend expectations.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -287,13 +287,30 @@ For a behavior change or feature, your PR should include a **witness test**:
 
 You can check this the artisanal way (`git stash && cargo test -p <crate>`),
 or let Stella verify Stella — run your task through `stella run --pipeline
-classic`, whose witness stage enforces exactly this fail→pass contract (a
-plain `stella run`, with no `--pipeline` flag, is the raw step-loop and does
-not verify on its own).
+<plugin-id>`, naming an installed verification plugin whose witness stage
+enforces this fail→pass contract. A plain `stella run` is the raw step-loop
+and does not verify on its own, and the built-in `classic` pipeline this
+flag once named is gone from the workspace, so `--pipeline classic` is
+refused.
 
 Pure refactors, docs, and CI changes don't need a witness — say so in the PR
 template and move on. If a witness is genuinely impractical (e.g. TUI
 rendering), explain how you verified the change instead.
+
+### The spend gate — pin what a turn buys
+
+A change to the agent loop can leave every answer the same and still buy an
+extra model call. `crates/stella-core/tests/spend_gate.rs` is what fails when
+it does: one turn per scenario over scripted ports, with the model calls, the
+tool runs and the reported cost pinned.
+
+Read a red run one of two ways. Either the change is a bug, and the fix is in
+the code. Or the new price is the one you meant, and you edit the pin in the
+same pull request and say in the PR why. A reviewer then reads the new price
+as a diff line. Never widen a pin to whatever the loop does now.
+
+`doc:verification-gate` covers what this layer can and cannot claim, and what
+is still unobserved beside it.
 
 ## Style
 
--- a/crates/stella-core/README.md
+++ b/crates/stella-core/README.md
@@ -313,8 +313,8 @@ make test-core           # cargo test -p stella-core
 make watch-core          # re-run on save (needs cargo-watch)
 ```
 
-There is no `tests/` directory: every test is an inline `#[cfg(test)]` module in
-the file it covers, which is why the engine's own suite is split across
+Nearly every test is an inline `#[cfg(test)]` module in the file it covers,
+which is why the engine's own suite is split across
 [`src/driver/tests.rs`](src/driver/tests.rs) and
 [`src/driver/tests/`](src/driver/tests) (`audit_fixes.rs` holds the 2026-07
 turn-driver audit witnesses; also `budget_boundaries.rs`,
@@ -327,6 +327,13 @@ fixture server and no network — driver tests wire scripted `Provider`s, counti
 `ToolExecutor`s and no-op `Sleeper`s, so the suite runs in seconds. Keep it that
 way: a test here that needs a file or a socket means the logic under test is in
 the wrong crate.
+
+[`tests/`](tests) holds the few that have to drive the crate the way a host
+does, through the public API alone. [`tests/spend_gate.rs`](tests/spend_gate.rs)
+is the one to read first: it pins the model calls, tool runs and cost of one
+turn per scenario, so a loop change that quietly buys an extra model call fails
+here. An intended change edits the pin in the same pull request
+(`doc:verification-gate`).
 
 ## Extending it
 
--- a/crates/stella-core/tests/spend_gate.rs
+++ b/crates/stella-core/tests/spend_gate.rs
@@ -1,0 +1,294 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+// Copyright (c) 2026 Oxagen, Inc. Commercial licensing: licensing@oxagen.sh
+
+//! The spend gate: what a turn buys, pinned scenario by scenario.
+//!
+//! A change to the step loop can leave every answer the same and still buy
+//! an extra model call. This file is what fails when it does. Each test
+//! below drives one turn over scripted ports and pins three numbers: model
+//! calls, tool runs, and the cost the turn reports.
+//!
+//! Read a red run one of two ways. Either the change is a bug, and the fix
+//! is in the code. Or the new numbers are the ones you meant, and you edit
+//! the pin in this file in the same pull request. A reviewer then reads the
+//! new price as a diff line. Never widen a pin to whatever the loop does
+//! now: the worth of this file is that intent is written down.
+//!
+//! The ports are scripted, so the engine does no I/O (architecture rule 2).
+//! An answered model call costs a flat quarter dollar here. That price is a
+//! round binary number, so a sum of them is exact and a pin can be too.
+
+use std::sync::Arc;
+use std::sync::atomic::{AtomicU32, Ordering};
+
+use async_trait::async_trait;
+use serde_json::Value;
+use stella_core::budget::BudgetGuard;
+use stella_core::ports::ToolExecutor;
+use stella_core::retry::Sleeper;
+use stella_core::{Engine, EngineConfig, TurnOutcome};
+use stella_protocol::{
+    BudgetMode, CompletionMessage, CompletionRequestRef, CompletionResult, CompletionUsage,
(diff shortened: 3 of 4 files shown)


macanderson:
**DoD verified on #4223 at `57073089`** — the `dod` gate was red only because the eight boxes were unticked. Each was checked against this branch's diff before ticking:

1. **`spend_gate.rs` over scripted ports, no engine I/O** — `ScriptedProvider: Provider` and `CountingTools: ToolExecutor`; `run_turn` drives `Engine::with_sleeper` once per scenario.
2. **Exact counts and cost per scenario** — all four are present, each asserting a `Spend { model_calls, tool_calls, cost_usd }`.
3. **Failure names the scenario and both numbers** — `assert_spend` prints ``the scenario `{scenario}` changed price`` with `pinned:` beside `bought:`.
4. **Operating rule in the file header** — the `//!` block: "you edit the pin in this file in the same pull request."
5. **`doc:verification-gate` names the layer and its limits** — the new "The spend gate — Layer 1's price half, rebuilt" section, including "What it does not prove".
6. **`CONTRIBUTING.md` names the gate and the same rule** — the new "The spend gate — pin what a turn buys" section.
7. **PR records the option and why** — option 2, with 1 and 3 explicitly not foreclosed.
8. **CI green** — see below.

On item 8, stated precisely rather than just ticked: at this head all 19 checks are green or skipped **except the `dod` gate itself**, which was red solely for the unticked boxes this comment resolves. `fmt + clippy + test` passed at 03:30:04 ([job](https://github.com/macanderson/stella/actions/runs/33710595210/job/100509172937)), so the pins are confirmed by a real run rather than by the derivation in the PR body — including the loop-detector `4/4`, which `stuck_loop_steers_once_then_aborts_on_re_detection` independently pins at four tool calls for the same scripted shape.

Applying `area:core` to fire the `labeled` event, since `dod-check.yml` re-runs only on `pull_request`. No code change was made to this PR.

---

PR #5709, opened 2026-09-03
fix(stella-tools): say whose payload cap stopped a batched read

### LINKED ISSUE #5707: A batched read_file blames the wrong cap when it runs out of room
## What happens

`read_file` can read several files, or several ranges, in one call. Every
section of one call spends from a single 64 KB render budget
(`MAX_RENDER_BYTES` in `crates/stella-tools/src/read.rs`). When a section runs
out of room, the footer says:

```
stopped at the 64 KB payload cap — continue with offset=1401
```

That sentence was written for a call that reads one file, where the number is
true. In a batch the section may have stopped after a few hundred bytes,
because the files ahead of it already spent the budget. The model is handed a
number that does not match what it saw, and the repair the sentence implies —
ask for a narrower range — cannot work, because the missing room is held by
another file in the same call. The move that does work is a second call for
that file alone, and nothing says so.

## Where

- `crates/stella-tools/src/read.rs` — the `payload_capped` block inside
  `ReadFile::read_section`. `ReadFile::execute` hands the budget down: the
  whole 64 KB for a single call, whatever is left for each section of a batch.
- Tests: `crates/stella-tools/src/read/tests.rs`.

## How to see it

Write two files that each render past 64 KB, then call `read_file` with
`{"files": [{"path": "one.sql"}, {"path": "two.sql"}]}`. The first section's
footer names the 64 KB cap with nothing to say the cap belongs to the call
rather than to that file.

## Constraints

- The single-file footer must stay byte for byte what it is. Loop comparison
  strips it (`stella_core::driver::loop_evidence`) and
  `the_total_payload_cap_stops_the_render_and_reports_it` asserts its wording.
- The tail message a batch already prints for the files it never reached says
  "batch payload cap", so the two should agree with each other.

## Definition of done

- [x] A batch section that runs out of room says the cap belongs to the whole
      call, and names the follow-up call that gets the file back.
- [x] The single-file footer is unchanged, and a test holds it there.
- [x] A witness test in `crates/stella-tools/src/read/tests.rs` that fails on
      the current code because the wording it asserts does not exist yet.

macanderson:
## What

A `read_file` call can name several files or ranges, and every section of one
call spends from a single 64 KB render budget (`MAX_RENDER_BYTES`). When a
section ran out of room the footer said:

```
stopped at the 64 KB payload cap — continue with offset=1401
```

That sentence is written for a call that reads one file, where 64 KB is what
the reader actually got. Inside a batch the section may have stopped after a
few hundred bytes, because the files ahead of it already spent the budget. The
number then describes nothing the model saw, and the repair it implies — ask
for a narrower range — cannot work, because the missing room is held by another
file in the same call. The move that does work, a second call for that file
alone, was named nowhere.

A batch section now says the cap belongs to the call and names that follow-up.
The single-file footer is byte for byte what it was: loop comparison strips
that footer (`stella_core::driver::loop_evidence`) and
`the_total_payload_cap_stops_the_render_and_reports_it` asserts its wording, so
the two spellings are told apart by a `RenderBudget` the call passes down
rather than by reshaping the shared text.

The read-side steering sentence is pinned in the same change. "To read a file,
use read_file rather than the shell" and "pass files to ONE read_file call" are
the prompt half of the sed-over-read_file defect, and
`both_prompts_keep_the_steering_the_schemas_cannot_carry` listed only the write
side — a later trim of either sentence failed nothing.

## Why this shape

`RenderBudget { bytes, shared }` rather than comparing the budget against
`MAX_RENDER_BYTES` to guess which spelling asked: the first section of a batch
gets the whole ceiling and its cap is still the call's, so the fact belongs to
the call and not to the arithmetic. Exemplar followed: none outside this tree —
it is the same "carry the fact the callee cannot re-derive" shape `ReadTarget`
already uses in this module.

## Witness

`a_capped_batch_section_names_the_cap_as_the_calls_own`
(`crates/stella-tools/src/read/tests.rs`). It reads two files that each render
past the ceiling in one call and requires the footer to say
`payload cap this call shares across its files` and
`asking for this file on its own`. Neither string exists on `main`, so the
assertion fails there by construction. The second half of the same test reads
one of those files through the single form and requires the old wording
verbatim; that half passes on both sides, and is what would catch a fix that
moved the single form's bytes.

The prompt claims are a regression guard rather than a witness: both sentences
are in `tool_steering!()` today, and the test fails only once someone removes
one.

## Scope, and what remains on 4151

The batch mechanism itself already shipped: the `files` key and the
per-element ledger entries landed earlier, and the schema description, the docs
card and the drift-oracle witness landed in pull request 5697. This is the one
defect left in that surface that a reader of the footer can hit.

What issue 4151 still waits on is its own third clause, unchanged: re-measure
the sed-to-`read_file` ratio on a comparable wide-orientation task and report
the number. That needs a real run against a real model, so the issue stays
open.

Tests deleted: None.

## Issues

Residue already tracked, not fixed here:

- https://github.com/macanderson/stella/issues/5695 — nothing writes an `R` row
  to `files_touched`, so the audit ledger under-reports what a turn read,
  whichever spelling it used.
- https://github.com/macanderson/stella/issues/5696 — the plural `read_file`
  form returns no structured `data`, so the deck's read head measures nothing
  for a batched read.

Closes #5707
Refs #4151

## Summary by Sourcery

Clarify payload-cap guidance for batched reads and preserve compatible single-file behavior.

Bug Fixes:
- Clarify batched read payload-cap footers so they identify the shared call budget and recommend requesting the affected file separately, while preserving the existing single-file wording.

Enhancements:
- Carry render-budget ownership through read rendering to distinguish shared batch caps from per-file caps.

Tests:
- Add regression coverage for shared-cap batch footer guidance and preservation of the single-file footer wording.
- Strengthen prompt steering tests for using read_file instead of shell reads and grouping files in one read_file call.

### DIFF
--- a/crates/stella-cli/src/agent/prompt/tests.rs
+++ b/crates/stella-cli/src/agent/prompt/tests.rs
@@ -75,6 +75,14 @@ fn both_prompts_keep_the_steering_the_schemas_cannot_carry() {
             // other one", and reaching for grep when `search` would answer
             // is the round trip the steering exists to prevent.
             "To find code, call search first",
+            // The read side, which leaked for as long as it was only a
+            // suggestion: one execution ran 24 `sed -n` range reads against 2
+            // `read_file` calls, and a shell read leaves the ledger no
+            // baseline for the drift oracle to attribute a later edit against.
+            "To read a file, use read_file rather than the shell",
+            // The round trip that made the shell cheaper. No schema can say
+            // "several files are still one call".
+            "pass files to ONE read_file call",
             // The file tools are not interchangeable with shell equivalents:
             // their edits are what the diff and verification are computed from.
             "use the file tools rather than the shell",
--- a/crates/stella-tools/src/read.rs
+++ b/crates/stella-tools/src/read.rs
@@ -363,6 +363,23 @@ fn normalized_key(root: &std::path::Path, path: &str) -> String {
 /// anything a batch read.
 const FILES_KEY: &str = "files";
 
+/// How many bytes one section may render, and whether the call it belongs to
+/// spends that ceiling across several files.
+///
+/// `shared` exists for the footer alone. A batch spends one
+/// [`MAX_RENDER_BYTES`] budget across every section, so a later section can
+/// stop after a few hundred bytes of it; told only "stopped at the 64 KB
+/// payload cap", a model reads that as a file too big to show and narrows its
+/// range, when the move that works is to ask for that file in a call of its
+/// own.
+#[derive(Clone, Copy)]
+struct RenderBudget {
+    /// Bytes this section may render into.
+    bytes: usize,
+    /// Whether the ceiling belongs to the whole call rather than to this file.
+    shared: bool,
+}
+
 /// One file and which slice of it — the unit both spellings of a `read_file`
 /// call reduce to.
 struct ReadTarget {
@@ -434,7 +451,11 @@ impl Tool for ReadFile {
             // Byte-identical to what the single form has always emitted. The
             // footer is what loop comparison strips and what this module's
             // tests assert verbatim, so the batch path must not reshape it.
-            return self.read_one(&targets[0], ctx, MAX_RENDER_BYTES).await.0;
+            let budget = RenderBudget {
+                bytes: MAX_RENDER_BYTES,
+                shared: false,
+            };
+            return self.read_one(&targets[0], ctx, budget).await.0;
         }
 
         // The payload cap is a BATCH budget, not a per-file one. Ten files at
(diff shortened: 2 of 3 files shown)


macanderson:
**DoD verified on #5707 at `33a56b23`** — the `dod` gate was red only because the three boxes were unticked. Each was checked against this branch's diff and against `main` at `6ceec2e` before ticking:

- **Batch section names the call's cap and the follow-up.** `read_section`'s `budget.shared` arm writes `payload cap this call shares across its files — continue with offset={resume}, asking for this file on its own`.
- **Single-file footer unchanged.** The `else` arm renders `stopped at the {cap_kb} KB payload cap — continue with offset={resume}`, byte-identical to `main`'s `read.rs:772`. `the_total_payload_cap_stops_the_render_and_reports_it` still holds it, and the new test's second half asserts the same wording plus the absence of `shares across its files`.
- **Witness fails on the old code.** `grep -rn "shares across its files\|asking for this file on its own" crates/` returns nothing at `6ceec2e`, so `a_capped_batch_section_names_the_cap_as_the_calls_own` cannot pass there — its asserted strings do not exist.

Re-running the `dod` check against the ticked issue. No code change was made to this PR.

---

PR #5711, opened 2026-09-03
feat(stella-cli): publish the CI-hardening record and prove the loop is steered by it

### LINKED ISSUE #4000: Ship context records that steer the autonomous loop, including CI-hardening lessons
Part of #3599. Asked for directly by @macanderson: the loop's behaviour must be **driven 100% from context records and steering authorities**, so people change it by retiring records and adding new ones rather than by changing Rust.

## Why this is an issue and not a prompt

The `work` verb's prompt used to say "follow this repository's own conventions — read AGENTS.md and CLAUDE.md". That was removed on #3940's branch, because a prompt is a steering channel nobody can retire: changing how the loop behaves would mean editing Rust.

So the guidance has to live where it can be retired — `.stella/rules/*.toml` — and `stella run` already loads it for every turn, whether that turn was started by a person or by the loop.

**The prompt is now correct and the records do not exist yet.** That is the gap.

## What to write

Records under `.stella/rules/`, authored through `stella context keep` / `promote` — **not by hand**. The files carry a `record_hash` and `promotions.jsonl` is a hash-chained ledger that `stella context validate` re-verifies in CI, so a hand-written record fails the gate.

At minimum:

1. **Harden CI from lessons.** When a defect is fixed, ask whether a guard would have caught it, and add one if so. This repository has the evidence that it works: `check-gate-parity`, `check-deleted-tests`, `check-module-reachability` and `main-canary` each exist because something got through once. A loop that fixes without hardening will meet the same class of defect forever.
2. **Prefer the design that is still correct later.** CLAUDE.md's reference-grade bar, already a record (`ctx.macanderson.stella.reference-grade-bar`) — check whether the loop's turns actually receive it.
3. **Nothing left behind.** File what you noticed and did not fix. Already partly structural via `close --partial`, but the judgement half is prose.

## The check that matters

Confirm the records **reach a loop-driven turn**. `work` spawns `stella run` in a fresh worktree, and a worktree that has not been trusted prints `project steering in … was NOT loaded (N context records)` and steers on nothing. If that warning fires for loop-driven turns, every record here is inert and the loop is running unsteered — which would make this issue's premise false and is worth checking first.

## Done when

A loop-driven turn demonstrably receives the workspace's context records, and retiring a record demonstrably changes what the loop does — with no Rust change.

## Definition of done

- [x] A loop-driven turn demonstrably receives the workspace's context records.
- [x] Retiring a record demonstrably changes what the loop does.
- [x] Neither is achieved with a Rust change.

macanderson:
## What and why

The self-driving loop's behaviour is meant to come from context records, so a
person changes it by retiring a record rather than by editing Rust. Two of the
three records #4000 asks for already existed —
`ctx.macanderson.stella.reference-grade-bar` (prefer the durable design) and
`ctx.macanderson.stella.nothing-left-behind`. The CI-hardening one did not, and
nothing anywhere proved that any of them reach a turn the loop started.

**The premise check the design pointer asked for first.** `stella context list`
in this worktree does print `project steering in … was NOT loaded (39 context
records)` when the checkout is untrusted, so the warning is real. It no longer
means the loop runs unsteered: `self_driving_cmd::work`'s `refuse_if_unsteered`
(landed with #4001) refuses to work an issue when `.stella/rules` holds records
and the workspace is untrusted, and `turn_command` passes the parent's
environment to the child, so the trust the parent checked is the trust the turn
gets. Both of those were untested, and both are witnessed here rather than
changed — refusing is the safer of the two designs the pointer named, and
undoing it to pass trust implicitly would trade a loud stop for a silent one.

## What this adds

`.stella/rules/ctx.stella.harden-ci-from-lessons.toml` — *"When a defect is
fixed, the same change adds the guard that would have caught it whenever such a
guard can be written."*

Hand-authored with no `record_id` or `record_hash`, exactly the way the two
`ctx.stella.*` records beside it are written: identity is derived, so the loader
stamps both from the content on load and no hash is typed in. `stella context
keep` was not used because it needs an ingest run first, and the maintainer's own
investigation on the issue measured that at 20+ minutes and several hundred
candidates for one document; the reason the issue gives for "not by hand" is the
hash, and deriving it is what removes that hazard. Evidence from the installed
0.9.300 binary in this worktree:

- `stella context validate` → *Every loaded record is schema-valid and currently
  believed.* (CI runs this too.)
- `stella context list` → `^harden-ci-from-lessons  must · cached`.

Its first wording tripped the atomicity gate (`", and "` with two commas reads as
a compound claim, `stella_core::ingest::gate`'s `detect_compound`) and was
rewritten to one clause.

## Witnesses and how each one fails on the base

`crates/stella-cli/src/agent/prompt/tests.rs` —
`a_worktree_turn_is_steered_by_this_repositorys_records`. A loop-driven turn is
`stella run` spawned with its cwd set to a fresh worktree, so its prompt is
assembled over that worktree's root. The test copies this repository's own
`.stella/rules` into a temp worktree, assembles the prompt through
`load_workspace_rules` + `assemble_system_prompt` with project prompts trusted,
and asserts all three statements are in it. It then archives the CI-hardening
record — the same edit a person makes, `status` stops being `active` — reloads,
and asserts that statement is gone while the other two remain. **Fails on the
base**: `ctx.stella.harden-ci-from-lessons` does not exist there, so the first
assertion finds no such sentence.

`crates/stella-cli/src/self_driving_cmd/work.rs` —
`the_loop_will_not_work_an_issue_that_its_records_cannot_steer` covers all three
cells (records + untrusted refuses and names the remedy, records + trusted
proceeds, no records + untrusted proceeds). **Fails on the base**:
`refuse_unless_trusted` does not exist, so the file does not compile.

`the_turn_keeps_the_trust_that_lets_this_repository_steer_it` asserts the built
child command does not remove `STELLA_TRUST_PROJECT`. **Fails on the base** only
if someone adds that removal; it is a regression guard on the one link joining
the parent's trust check to the child that needs it, written in the shape the
sibling `the_turn_does_not_inherit_the_reflection_opt_out` already uses.

## The one Rust change

`refuse_if_unsteered` grew a pure seam, `refuse_unless_trusted(root, trusted)`,
because `project_code_execution_trusted` reads the process environment and this
suite runs in parallel — the same split `classify` in that file already draws
and states. No behaviour changes: the public entry point passes the same answer
it always read. Nothing about the steering itself is Rust.

Exemplar followed: the file's own neighbours — `classify`'s pure-seam comment and
`turn_command`'s "separated so the whole command is a test seam" — rather than an
outside crate.

## Verification

CI is the evidence; nothing was compiled on the maintainer's machine (CLAUDE.md,
"CI builds and tests; this laptop does not"). Run locally: `cargo fmt --all`,
`make guards-fast` (green), `python3 scripts/check-prose.py` (green — the record's
comment header was rewritten twice to clear the 6.00 reading-grade ceiling for a
new file), `./scripts/check-file-size.sh` (green), `make line-citations`,
`make dead-code-allows`, and the two `stella context` commands quoted above.

Tests deleted: None.

Residue issues filed: #5712 (`refuse_if_unsteered` counts `governance.toml` and
archived records toward its tally, so the loop refuses over files that would
never steer it) and #5714 (there is no `stella context retire`, so a retirement
lands as a hand-edit with no ledger event naming who did it).

Closes #4000

## Definition of done

#4000's three boxes are ticked, each on the evidence in this diff rather than in
this description; the reasoning per box is in the PR comment of 2026-09-03. The
`dod / dod` check last ran before those boxes were ticked and was red on the
stale reading — this edit re-triggers it (`dod-check.yml` fires on
`pull_request: edited`).

The one Sourcery ❌ — that the record is hand-authored rather than produced by
`stella context keep` / `promote` — is answered in the PR comment of the same
date, on three grounds: `keep` needs an ingest run that has no proposal to
resolve, the hazard the rule guards is a typed-in hash that this record derives
instead, and the two sibling `ctx.stella.*` records are hand-authored in exactly
this shape.

## Summary by Sourcery

Publish the CI-hardening steering record and verify that trusted self-driving turns receive active repository guidance while unsteered work is stopped safely.

New Features:
- Publish a CI-hardening context record that requires fixes to add applicable regression guards.

Enhancements:
- Add coverage proving repository context records steer turn prompts and that retiring a record removes only its guidance.
- Add coverage for refusing self-driving work when records cannot steer an untrusted workspace, while allowing trusted or record-free work.
- Add a pure trust-check seam for deterministic testing and guard inheritance of project trust by child turns.

Tests:
- Add prompt, steering, trust, and child-environment regression tests covering the repository's context-driven loop behavior.

### DIFF
--- a/.stella/rules/ctx.stella.harden-ci-from-lessons.toml
+++ b/.stella/rules/ctx.stella.harden-ci-from-lessons.toml
@@ -1,0 +1,56 @@
+# This repo's own policy, kept as a record.
+#
+# It comes from what AGENTS.md says about the gate. Most steps in the gate
+# were added after a bug slipped by. `gate-parity` came after the step list
+# drifted. `check-deleted-tests` came after a merge dropped a test.
+# `main-canary` came after two good branches broke the tree.
+#
+# So the rule is the habit, not the list. A fix with no guard behind it
+# lets the same bug come back.
+#
+# Written by hand, like the two records next to it. The id is made from
+# the text, so the loader stamps `record_id` and `record_hash` on the
+# first load. No one has to trust a typed-in hash. `stella context
+# validate` checks it on each pull request.
+#
+# It has no probe. The claim is about what a session does after a fix.
+# No file on disk can prove it wrong. A probe on a stand-in would make
+# `validate` say more than it knows.
+
+schema = "context-record/v0.1"
+set_id = "stella"
+
+[defaults]
+sharing_scope = "repository"
+status        = "active"
+origin        = "user"
+
+# `source_digest`/`commit` pin AGENTS.md at the state the claim was read
+# from, not at the current tip.
+[defaults.provenance]
+source_kind   = "document"
+source_uri    = "AGENTS.md"
+source_digest = "sha256:ccff490389b7d7537ab27c61f0ea1e08d35b7035016eeb66c4df2580f413e791"
+repo          = "git@github.com:macanderson/stella"
+commit        = "ea1dc958"
+
+
+[[record]]
+lineage_id = "ctx.stella.harden-ci-from-lessons"
+kind       = "rule"
+statement  = "When a defect is fixed, the same change adds the guard that would have caught it whenever such a guard can be written."
+tags       = ["ci", "guards", "regressions"]
+
+  [record.steering]
+  force      = "must"
+  precedence = 90
+  applies_to = { keywords = ["fix", "defect", "guard", "regression"] }
+
+  [record.enforcement]
+  mode = "none"
+
+  [record.truth]
+  basis        = "decree"
+  confidence   = 95
+  verified_at  = "2026-09-02T00:00:00Z"
+  review_every = "P180D"
--- a/crates/stella-cli/src/agent/prompt/tests.rs
+++ b/crates/stella-cli/src/agent/prompt/tests.rs
@@ -879,3 +879,111 @@ fn the_observatory_marker_copy_matches_the_emitting_constants() {
         "the rules heading moved — update the Observatory MARKERS"
     );
 }
+
+/// The three statements this repository steers its own autonomous loop by,
+/// copied from `.stella/rules/` byte for byte. A record whose wording changes
+/// has to change here too, which is the point: the loop's standards are meant
+/// to be reviewed, not drifted.
+const HARDEN_CI: &str = "When a defect is fixed, the same change adds the guard that would have caught it whenever such a guard can be written.";
+const DURABLE_DESIGN: &str = "Code in this repository must meet a reference-grade Rust quality bar rather than merely working.";
+const NOTHING_LEFT_BEHIND: &str = "Work is not finished while anything noticed during the session lives only in the author's head, a chat transcript, or a worktree about to be deleted.";
+
+/// This repository's root, two levels above the crate this test compiles in.
+fn repository_root() -> std::path::PathBuf {
+    std::path::Path::new(env!("CARGO_MANIFEST_DIR"))
+        .parent()
+        .and_then(std::path::Path::parent)
+        .expect("the crate sits under <repo>/crates/")
+        .to_path_buf()
+}
+
+/// A throwaway checkout carrying this repository's own published records.
+///
+/// A copy rather than the tree itself, because the test retires a record and
+/// must not edit the file a person reviews.
+fn worktree_with_this_repositorys_records() -> tempfile::TempDir {
+    let worktree = tempfile::tempdir().expect("worktree");
+    let rules = worktree.path().join(".stella").join("rules");
+    std::fs::create_dir_all(&rules).expect("rules directory");
+    let published = repository_root().join(".stella").join("rules");
+    for entry in std::fs::read_dir(&published).expect("this repository publishes records") {
+        let from = entry.expect("a directory entry").path();
+        if from.is_file() {
+            let name = from.file_name().expect("a file name");
+            std::fs::copy(&from, rules.join(name)).expect("copy one record");
+        }
+    }
+    worktree
+}
+
+/// The system prompt a turn in `root` is given, with the repository trusted to
+/// steer it.
+fn prompt_in(root: &std::path::Path) -> String {
+    let authority = crate::settings::AuthorityPolicy {
+        project_prompts_allowed: true,
+        ..Default::default()
+    };
+    let resolved = crate::rules::load_workspace_rules(root, &authority);
+    super::assemble_system_prompt(super::SYSTEM_PROMPT, root, &authority, &resolved, None)
+}
+
+/// **Witness.** The self-driving loop starts a turn by spawning `stella run`
+/// in a fresh worktree (`self_driving_cmd::work`'s `turn_command`), so the
+/// turn's prompt is assembled over that worktree's root. This drives that: a
+/// worktree holding this repository's own records, and the prompt a turn
+/// there is given.
+///
+/// The records reach the turn. All three the loop is meant to be steered by
+/// are asserted, because two of them already existed — a test that checked
+/// one would not show that the set arrives.
+///
+/// Retiring a record stops it steering. The retirement here is the edit a
+/// person makes in `.stella/rules/`: the record's status stops being active.
+/// No Rust changes in either direction.
+///
+/// It fails on the base commit. `ctx.stella.harden-ci-from-lessons` does not
+/// exist there, so the first assertion finds no such sentence in the prompt.
+#[test]
+fn a_worktree_turn_is_steered_by_this_repositorys_records() {
+    let worktree = worktree_with_this_repositorys_records();
+
+    let steered = prompt_in(worktree.path());
+    assert!(
+        steered.contains(stella_core::records::CACHED_HEADING.trim_start()),
+        "the records must render as the cached rules block: {steered}"
+    );
+    for statement in [HARDEN_CI, DURABLE_DESIGN, NOTHING_LEFT_BEHIND] {
+        assert!(
+            steered.contains(statement),
+            "a turn in a worktree of this repository must be steered by {statement:?}"
+        );
+    }
+
+    let record = worktree
+        .path()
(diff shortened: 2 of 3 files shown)


macanderson:
Residue filed while doing this work: #5712 (the loop refuses over rule files that would never steer it — `governance.toml` and archived records both count toward `refuse_if_unsteered`'s tally) and #5714 (there is no `stella context retire`, so a retirement lands as a hand-edit with no ledger event naming who did it).

macanderson:
**Answering the one ❌ — "hand-authors the new record instead of creating it through `stella context keep` or `promote`".**

Correct as an observation, and deliberate. Three reasons, and the maintainer's
own investigation on #4000 is the source of the first.

**1. `keep` takes a proposal, and there is no proposal without an ingest run.**
`stella context keep <candidate>` resolves a candidate id out of
`.stella/proposals`, which only `stella ingest` writes — a model call over a
document. The maintainer ran it on `AGENTS.md` and reported it still going at
ten minutes, two of four parts in, with 106 candidates already
(https://github.com/macanderson/stella/issues/4000#issuecomment-5498152294).
The follow-up comment on the same issue recommends option 2 — a short
purpose-built document ingested instead — precisely because curating several
hundred model-written proposals down to one wanted sentence is the expensive
half. Publishing a selection nobody reviewed is what that comment declined to
do.

**2. The hazard the rule exists for is the hash, and the hash is derived here,
not typed.** #4000's wording is "authored through `stella context keep` /
`promote` — **not by hand** … the files carry a `record_hash` and
`promotions.jsonl` is a hash-chained ledger that `stella context validate`
re-verifies in CI, so a hand-written record fails the gate." A record that
omits `record_id`/`record_hash` does not fail the gate: `records::resolve`
re-stamps every record on load over the ADR 0004 canonical preimage, so the
identity is computed from the content and there is no typed-in value to be
wrong. `stella context amend`'s own module doc states this as the supported
repair. Evidence from the installed 0.9.300 binary in this worktree:

```
$ stella context validate
Every loaded record is schema-valid and currently believed.

$ stella context list
  ^harden-ci-from-lessons  must · cached
```

**3. It is the tree's existing pattern for a repository-policy record.** The
two records in the `ctx.stella.*` set — `rust-toolchain-pin` and
`dependency-license-allowlist` — are both hand-authored, both carry no
`record_id`/`record_hash`, and both say so in their own headers. The new record
copies their shape field for field. `promotions.jsonl` carries enforcement
grants and lifecycle events, not publications: it holds two entries against the
tree's 39 records, and this record enforces nothing (`enforcement.mode =
"none"`), so it grants nothing to record.

If the maintainer would rather the record come from an ingest of a short
purpose-written document, that is a wording change to one `statement` field and
a provenance block, and I will make it — but it costs a model call and produces
a claim a model phrased, where this one is the sentence #4000 asked for.

macanderson:
Ticked #4000's three DoD boxes after reading the diff rather than the description. What each one rests on, so the third can be argued with:

1. **A loop-driven turn receives the records** — `a_worktree_turn_is_steered_by_this_repositorys_records` copies the real `.stella/rules/` into a temp worktree and asserts all three statements land in the prompt assembled by `load_workspace_rules` + `assemble_system_prompt`. It asserts the set, not just the new record, which is what makes it evidence that the records arrive rather than that one string exists.
2. **Retiring a record changes what the loop does** — the same test rewrites `status = "active"` to `"archived"` in the TOML, reloads, and requires `HARDEN_CI` gone while the other two remain. The negative and the control are both there.
3. **Neither is achieved with a Rust change** — ticked on the reading that the *steering* is not Rust: the record is a TOML file and the retirement is a TOML edit. The PR does change Rust (`refuse_unless_trusted`, a pure seam split out of `refuse_if_unsteered`), and that change is real but is a test seam for the trust refusal, not part of either demonstration above. If the box was meant as "this PR contains no Rust change at all", it is not satisfied and should be unticked — flagging it rather than leaving a ticked box a reader would find contradicted by the diff.

No other action taken on this PR. `dod / dod` was the only red check; the rest were still in flight at the time of writing.

---

macanderson:
**DoD verification for #4000 — two boxes hold, the third does not, and there is an untested shipping path.**

The `dod` check is green again (run `33711180290`, 03:23:41); the two failures at 03:15 were read against the pre-tick body. Checking the three boxes against the diff rather than the description:

- **"A loop-driven turn demonstrably receives the workspace's context records"** — holds by proxy. `a_worktree_turn_is_steered_by_this_repositorys_records` copies the real `.stella/rules/` into a tempdir and asserts all three statements reach `assemble_system_prompt`. Worth naming the one unproven link: no loop-driven turn is actually run — the test calls the assembler directly with `project_prompts_allowed: true` hard-set. The far end is closed by `the_turn_keeps_the_trust_that_lets_this_repository_steer_it`, but the middle link — that `stella run`'s own startup reaches this assembly path over the worktree root — is asserted in the doc comment, not tested.
- **"Retiring a record demonstrably changes what the loop does"** — holds, and cleanly. The same test flips `status = "active"` → `"archived"`, guards the substitution with `assert_ne!`, and asserts `HARDEN_CI` disappears **while the other two statements remain**. Negative and control both present.
- **"Neither is achieved with a Rust change"** — does not hold as an unqualified tick. The diff carries a *shipping* Rust change in the loop's own gating path: `refuse_if_unsteered` now delegates to a new `refuse_unless_trusted(root, trusted)`, and `if count == 0 || crate::settings::project_code_execution_trusted()` became `if count == 0 || trusted`. You flagged this yourself on 2026-09-03; I agree with your reading, and it is satisfiable only under the narrower "the steering *mechanism* is TOML, not Rust" — which is true, but is not what the box says.

**One thing not yet raised:** the "no behaviour change" claim for that seam split rests on the PR description alone. The extracted pure helper `refuse_unless_trusted` has tests; the shipping entry point `refuse_if_unsteered` has none, so nothing in CI would catch the split having changed what the loop refuses. That is the gap I would close before merge — either a witness on the entry point, or untick box 3 and split the seam extraction into its own change.

Note also that `fmt + clippy + test` had not finished when I looked, so the two witness tests above are so far unproven in CI.

---

Issue #3292, opened 2026-08-14
Loop detection cannot catch a model that grinds through 62 different commands without making progress


macanderson:
## The blind spot

`loop_detect` has four detectors and every one of them keys on something being **identical**: exact repeat and short cycle key on identical *inputs*, stagnation keys on identical *outputs*, interleaved repeat keys on an identical (input, output) pair recurring. A model that grinds unproductively while varying both is invisible to all four.

## The trace

`~/.arenabench/matches/5a38720c145f/jobs/5a38720c145f-stella-glm-5-2-via-openrouter-bare-loop/extract-moves-from-video__3AqwMdY/agent/stella-events.jsonl`

Census of that trial:

- 62 `tool_start` events, **all 62 are `bash`** — no other tool called at any point
- ~~62 distinct tool inputs (nothing repeats, so detectors 1, 2 and 4 never arm)~~ **59** distinct tool inputs: three commands are sent again later in the turn. Detectors 1, 2 and 4 still never arm, because each of them also needs the answer to come back unchanged and all 62 answers differ. Corrected 2026-09-07 from the committed fixture (`crates/stella-core/src/loop_detect/tests/extract_moves_from_video.json`, `the_recorded_grind_varies_both_its_inputs_and_its_outputs`).
- **62 distinct outputs** (so detector 3, stagnation, never arms either)
- the turn ran to its step ceiling and solved nothing

So this is not a tuning problem with an existing detector. It is a class none of them covers: *varied inputs producing varied outputs that nonetheless carry no new information*.

## Why this is not a one-line fix

"Unproductive" needs a definition the engine can compute without a model call, and a false positive here **aborts work that was going fine** — strictly worse than the grind it prevents. Candidate signals:

- **novelty collapse** — the fraction of output lines already seen earlier in the window. **Measured and ruled out** (2026-09-03 comment): no threshold has zero false positives on the winning population, in three formulations at four window sizes.
- **workspace immobility** — no file mutated over N steps. **Measured and ruled out** (2026-09-04 comment): no threshold has zero false positives in any of six formulations, and the trace named above does not fire under the readings that are not confounded by shell quoting. That study also found that the per-call "did the tree move" answer does not exist on the step path at all (#5864), and that the winning population contains grinding, so the zero-false-positive bar cannot be cleared against these labels (#5865).
[...]

Issue #3176, opened 2026-08-13
edit_file's constant success string trips the stagnation detector — 7 distinct edits killed a correct in-progress solve (overfull-hbox)


macanderson:
## Problem

On Terminal-Bench `overfull-hbox` (h2h891 head-to-head, five-tool bare loop, SUT d6bc3e5cb), Stella understood the task and was making **correct, distinct** synonym substitutions, but the turn was aborted mid-solution as a stuck loop. The stagnation detector's own message admits the contradiction:

> the last 7 `edit_file` calls used DIFFERENT arguments but every one returned byte-identical output

`edit_file` returns the constant string `"replaced 1 occurrence(s) in /app/input.tex"` regardless of *what* was replaced. Seven different edits therefore look identical to a detector keyed on tool-output repetition — even though the detector already knows (and says) the arguments differed. No `complete` event fired; a correct solve was killed. CC interleaved edit→recompile→re-grep on the same task and never tripped it.

## Why it matters

This is the same class as the recorded "never put a timing in a ToolOutput — the loop detector keys on it" hazard: a tool's output must carry enough identity to distinguish distinct operations, or a repetition-based guard misfires. Here it converts a competent run into an `agent_error` loss on the scoreboard — a false negative that depresses the measured solve rate.

## Fix (either or both)

1. `edit_file` success output should embed distinguishing identity — the path AND the matched line/offset or a short hash of the applied hunk — so distinct edits produce distinct output. (`crates/stella-tools/src/edit_file.rs` or wherever the success string is formed.)
2. The stagnation detector should exempt calls whose *arguments* differ — it already has that information (it prints it). Byte-identical output from differing args is not a loop.

## Verify
Witness: N `edit_file` calls to the same file with different arguments, back-to-back, must not trip the stagnation abort. Repro: `overfull-hbox` trial under `runs/h2h891/*/…/overfull-hbox__*/agent/stella-events.jsonl` (the abort event is near the end).

## Context
Found in the five-tool-arm head-to-head analysis; one of two defects that turned competent runs into scoreboard losses (the other: self-certification with a blind oracle, filed separately).

Issue #2793, opened 2026-08-11
MCP and custom tools bypass the blocking policy chains and the #2676 approval flow


macanderson:
Problem: the blocking chains (tool.call.requested, file.*, command.started) and the new approval flow live in `ToolRegistry::execute` (`crates/stella-tools/src/registry.rs` + `registry/approval.rs`). MCP tools run through `stella-mcp`'s `McpToolSet` and custom TOML tools through `CustomToolSet` (`crates/stella-tools/src/custom.rs`); both wrap the registry as an inner executor and dispatch their own names WITHOUT the chains, so an extension policy (or a RequireApproval-producing hook) never sees those calls. The ToolPolicy on/off switch layer (stella-cli `PolicyToolSet`) does cover them — this gap is specifically the hook/approval plane. Pre-existing before #2676, but the approval flow makes the asymmetry user-visible: a gated built-in asks a human, an MCP tool with the same effect does not.

Fix direction: either route decorator dispatch through a shared gate entry (the #2676 `ApprovalBroker::resolve` + `resolve_precedence` seam in `crates/stella-tools/src/registry/approval.rs` is reusable and bus-parameterized), or have each decorator emit the same chains before executing. Verify: a blocking handler on tool.call.requested denying an MCP/custom tool name must actually block it. Done when the hook/approval plane sees every tool dispatch, not just built-ins.

Refs #2676, #2787.

macanderson:
Triage: P1 / Impact 6. Parented under new epic #2816 (interactive approval flow completeness).

Unlike its four approval-flow siblings (#2789-2792), this defect is independently verifiable on `main` *today*, not gated on unmerged PR #2787: `McpToolSet` (`crates/stella-mcp/src/toolset.rs:138`) and `CustomToolSet` (`crates/stella-tools/src/custom.rs:659`) both dispatch outside `ToolRegistry::execute`'s `tool.call.requested` chain (`crates/stella-tools/src/registry.rs`), so MCP and custom tools bypass the same blocking policy/hook chain that built-in tools go through. A live, exploitable authorization gap.

macanderson:
[...]

PR #5719, opened 2026-09-03
fix(stella-core): the stagnation rung reads where a tool came from

### LINKED ISSUE #3328: A loop-detection rule still kills legitimate work when a tool's output never changes on success (part of #3303, decide together with #3292)
## Problem

#3303's fix stamped the **silent custom-tool** success (exit 0, empty stdout) with an input-derived `sha256/8` in `crates/stella-tools/src/custom.rs::run_custom`, so distinct inputs no longer render byte-identical empty outputs. That was fix (a) of #3303 — harness-side, custom tools only. Two populations of the #3176 stagnation-hazard class remain, and only the detector-side fix (#3303's fix b, originally PR #3187's deliberately-untaken fix 2) can reach them:

1. **MCP tools.** Third-party server outputs are relayed verbatim (`stella-mcp` → the registry decorator chain); Stella cannot stamp them. A server whose success responses are constant (an ack string, an empty result) converts a legitimate work loop into a stagnation kill exactly as the custom-tool empty-stdout case did.
2. **Constant non-empty custom-tool stdout.** A script that prints the same bytes on every success (`echo done`) is the same hazard one byte away; the #3303 stamp fires only on *empty* stdout, deliberately, to keep the developer-API change minimal.

## The fix under consideration

Detector-side: weaken or exempt the output-keyed stagnation rung in `crates/stella-core/src/loop_detect.rs` ("Consecutive calls to the SAME tool that all produced byte-identical output — whatever their arguments") when the *arguments* differ across the run.

**This trades a false positive for a false negative** — the exact blindness #3292 documents from the other direction — so any change here must be decided **jointly with #3292**, or the two fixes will fight. That joint design call is the actual work of this issue; the code change is small once decided.

## Constraints

- `stella-core` is I/O-free (AGENTS.md architecture rule 2): any detector change is pure logic over owned data, with property tests (`loop_detect` already has them).
- Deterministic only — no timings in `ToolOutput` (#2706).
- `loop_detect.rs` is not a god file; no baseline involved.

## Verify / definition of done

Synthesized `loop_detect` records: N consecutive calls to one tool with **differing arguments** and byte-identical outputs must no longer trip the rung (or must trip it later / under a stricter joint condition per the #3292 decision); N calls with **identical** arguments and identical outputs must still trip it. Property tests updated alongside.

Refs #3303, #3292, #3176, #3187, #2706.

## Definition of done

- [x] N calls with **distinct** arguments and byte-identical outputs no longer trip the rung — or trip it later, under a stricter joint condition, per the #3292 decision.
- [x] N calls with **identical** arguments and identical outputs still trip it.
- [x] The property tests are updated alongside.

macanderson:
## What

The output-keyed stagnation rung in `crates/stella-core/src/loop_detect.rs`
now reads where the tool came from before it fires.

`detect_stagnation` takes the trailing run of same-tool calls whose output
never changed, and returns `None` when every input in the run matches — so
the only case it fires on is **arguments that differ, output that does not**.
For a built-in that is evidence: its output reports what it read, what it
matched, what it wrote, so the same bytes for new arguments mean the
arguments reached nothing new (the `grep`-alternation regression this rung
was written for). For a tool outside the binary it is no evidence at all. An
MCP server may ack every successful call with one string, and a workspace
script may print one constant line, so identical output there is the tool's
design.

`CallRecord` carries a new `origin: Option<ToolOrigin>` with three states —
`Builtin`, `Mcp`, `Custom`. The rung exempts a run whose trailing record is
marked `Mcp` or `Custom`, and behaves exactly as before for `Builtin` and for
`None`.

### Where the answer comes from

A new `ToolExecutor::tool_origin(&self, name)` port method, answered by the
layer that **dispatches** the name:

| Layer | Answers |
|---|---|
| `ToolRegistry` (`stella-tools`) | `Builtin` for a `catalog::get` row, else `None` |
| `McpToolSet` (`stella-mcp`) | `Mcp` for an `mcp__…` name, else forwards to native |
| `CustomToolSet` (`stella-tools`) | `Custom` for a `.stella/tools/*.toml` name, else forwards |

Every other decorator forwards unfiltered — `GatedToolSet`, `PolicyToolSet`,
`SkillScopedTools`, `ClaimTap`, `TaskTap`, `CommitObserver`, `ReadOnlyTools`,
`GrantedTools`, `CandidateMcpView`, `DelegatingTools`. Unfiltered on purpose:
the question is about a name that has already run, so narrowing it could only
lose an answer the rung needs.

The driver stamps the window in `driver/loop_evidence.rs::stamp_origins`, a
second pass over the records `recent_call_records` built, called from
`loop_escalation::check_loop_detection`. The port's default is `None`, which
the detector reads as "not recorded" and treats as it always did, so a
decorator that forgets to forward costs the fix and causes no regression.

Exemplar followed: `dispatch_gate` on the same trait — a defaulted accessor
answered by the base of the chain and forwarded by every decorator over it
(#2793). The forwarding sites and the doc-comment shape are copied from it.

## Witnesses, and why each fails on the old code

All in `crates/stella-core/src/loop_detect/tests/origin.rs` (a new submodule;
`tests.rs` would have crossed the 1500-line ceiling), plus one in
`crates/stella-cli/src/agent/tool_stack.rs`. **They cannot compile on the old
code** — `CallRecord` had no `origin` field and `ToolOrigin` did not exist —
so the mechanism is structural rather than a threshold that happens to move.
Beyond that, each pins its own direction:

- `an_mcp_tool_acking_every_call_the_same_way_is_not_stagnation` — six MCP
  calls, six sets of arguments, one constant ack, marked `Mcp` → `NoLoop`.
  Its second assertion runs the **same records with the mark removed** and
  requires `Stagnant`, so the test cannot pass by the fixture missing the
  rung. That assertion is what fails on old behaviour.
- `an_mcp_tool_called_identically_over_and_over_is_still_a_loop` — six
  identical MCP calls still trip `ExactRepeat` with count 6. The exemption is
  only for arguments that moved.
- `a_builtin_answering_every_new_argument_the_same_way_still_stagnates` —
  the `grep` fixture over both `Some(Builtin)` and `None`, both `Stagnant`
  count 6. The exemption reaches nothing else.
- `a_custom_tool_printing_a_constant_line_is_not_stagnation` — the
  constant-stdout script, the second population the issue names.
- Property `a_tool_from_outside_the_binary_never_stagnates` — over arbitrary
  histories and all five thresholds, a history marked `Mcp` or `Custom` never
  reaches `Stagnant`.
- Property `marking_a_history_builtin_changes_nothing` — marking every record
  `Builtin` yields verdict-for-verdict the same answer as leaving the origin
  unset, so recording the origin is invisible where it was already assumed.
- `the_production_tool_stack_forwards_tool_origin` (stella-cli) — through
  `session_stack_with_gate` with the skill plane on top, over a real
  `ToolRegistry` under a real `McpToolSet` with a real custom script tool:
  `task_list` → `Builtin`, `mcp__vendor__deploy` → `Mcp`, `my_tool` →
  `Custom`, an unknown name → `None`. A future decorator inserted into the
  shipped chain that forgets to forward fails here and nowhere else.

`Tests deleted: None.` The five origin tests were **moved** from
`loop_detect/tests.rs` into the new `loop_detect/tests/origin.rs` in the same
commit that added them, so they appear only once in the diff against `main`.

## Part B — the varied-grind study (#3292): the metric does not separate

The design pointer asked for the novelty-collapse measurement over
`~/.arenabench/matches/`, and to ship a detector only if some threshold has
**zero false positives on the working population**. No such threshold exists.
Recording the numbers here and leaving #3292 open, as the pointer directs.

**Population.** 294 local matches, 497 trials carrying both a
`verifier/reward.txt` and an `agent/stella-events.jsonl`: 264 at reward 0.0,
233 at reward 1.0. The stuck trace is `5a38720c145f` →
`extract-moves-from-video__3AqwMdY`, reward 0.0, 62 `tool_start` / 62
`tool_result` events — the trial #3292's census is drawn from. The 10-trial
working sample is the first ten reward-1.0 trials by path:
`build-cython-ext`, `reshard-c4-data`, `sparql-university`, `git-multibranch`,
`distribution-search`, `fix-git`, `compile-compcert`, `crack-7z-hash`,
`break-filter-js-from-html`, `financial-document-processor`.

**Metric A — repeated lines within a window of W tool outputs.** Non-empty
lines, stripped; a line counts as a repeat when it already appeared earlier
in the same window; collapse = repeats / total lines.

| W | stuck (n, median, max) | working-10 (n, median, max) | zero-FP threshold from the 10 | stuck windows it catches |
|---|---|---|---|---|
| 6 | 57, 0.111, 0.613 | 370, 0.143, 0.446 | &gt; 0.446 | 7 / 57 (12%) |
| 8 | 55, 0.129, 0.598 | 350, 0.157, 0.436 | &gt; 0.436 | 7 / 55 (13%) |
| 10 | 53, 0.162, 0.608 | 332, 0.175, 0.494 | &gt; 0.494 | 3 / 53 (6%) |

The stuck trial's **median is below the working median at every W**. The
populations do not separate; only the extreme tail does, and it catches one
window in eight.

**The 10-trial threshold does not survive the rest of the winners.** Against
all 230 reward-1.0 trials that score at W=6 (7,582 windows, median 0.149,
p99 0.844, max 0.939), the 0.446 threshold fires on **374 windows in 68 of
230 winning trials** — it would kill 30% of successful runs. A threshold with
zero false positives over the whole winning population is &gt; 0.939 at W=6,
&gt; 0.936 at W=8, &gt; 0.918 at W=10, and each catches **0** stuck windows.

**Metric B — the newest output against the prior W steps** (the pointer's
literal wording: fraction of an output's lines already seen in the last W
steps). Worse. At W=8 the stuck median is 0.250 against a working median of
0.168, but **128 of 221 winning trials contain a window at 1.000** — an
output made entirely of lines seen before. No threshold at or under the
ceiling has zero false positives, and the ceiling catches nothing.

**Sustained level rather than a peak.** Share of a trial's W=8 metric-B
windows at or above 0.5: the stuck trial is 0.389; across 170 winning trials
with 10+ windows the median is 0.243, p90 0.438, max 0.634, and **31 of 170
winners (18%) sit at or above the stuck trial's share**.

**Conclusion.** Novelty collapse over tool-output lines does not distinguish
grinding from work on this corpus, in three formulations. Nothing ships for
#3292 and it stays open, with these numbers posted on it along with the exact
script.

**On the fixture windows.** The pointer asked for redacted fixture windows
under the crate's test fixtures. With no detector shipping there is nothing
in the tree that would read them, and committed data no test touches is the
unwired-code shape AGENTS.md asks us not to leave behind. The script and the
per-population numbers are posted verbatim on #3292 instead, so the
measurement is reproducible from any machine holding `~/.arenabench`. Say the
word and I will land the fixtures alongside whatever detector #3292 settles
on.

## Deviation from the design pointer

None on part A. On part B the pointer's own escape clause applies — no
threshold has zero false positives, so no detector ships and #3292 stays
open. The fixture-window question above is the one judgement call.

## DoD verification (#3328)

The three DoD boxes on #3328 were ticked after checking each against this
diff, not on the strength of this description:

| DoD item | Where it is satisfied |
|---|---|
| Distinct arguments + byte-identical outputs no longer trip the rung, or trip it under a stricter joint condition | `detect_stagnation`'s origin guard in `loop_detect.rs`; witnessed by `an_mcp_tool_acking_every_call_the_same_way_is_not_stagnation` and `a_custom_tool_printing_a_constant_line_is_not_stagnation`. Tool origin is the stricter joint condition, and part B records why no #3292 detector ships beside it. |
| Identical arguments + identical outputs still trip it | `an_mcp_tool_called_identically_over_and_over_is_still_a_loop` (`ExactRepeat`, count 6) and `a_builtin_answering_every_new_argument_the_same_way_still_stagnates` (`Stagnant`, count 6, for both `Some(Builtin)` and `None`). |
| Property tests updated alongside | `a_tool_from_outside_the_binary_never_stagnates` and `marking_a_history_builtin_changes_nothing`, both in `loop_detect/tests/origin.rs`. |

Closes #3328
Refs #3292, Refs #3303, Refs #3176, Refs #3187

## Summary by Sourcery

Make stagnation detection account for tool origin so constant external-tool acknowledgements do not trigger false loop alerts while preserving existing safeguards.

New Features:
- Add tool-origin metadata to loop-detection records so calls can be identified as built-in, MCP, or custom tools.

Bug Fixes:
- Prevent constant outputs from MCP and custom tools with changing arguments from being misclassified as stagnation while preserving exact-repeat detection.
- Preserve existing stagnation behavior for built-in and unrecorded tool origins.

Enhancements:
- Extend the tool executor interface and production tool stack to resolve and forward tool origins throughout decorators.
- Stamp call records with origin information before loop detection evaluates escalation evidence.

Tests:
- Add unit and property tests covering origin-aware stagnation behavior and unchanged built-in semantics.
- Add an end-to-end tool-stack test verifying origin propagation for built-in, MCP, custom, and unknown tools.

Chores:
- Record the varied-grind novelty-collapse study results without shipping a detector because no zero-false-positive threshold was found.

### DIFF
--- a/crates/stella-cli/src/agent/tool_stack.rs
+++ b/crates/stella-cli/src/agent/tool_stack.rs
@@ -510,6 +510,71 @@ mod tests {
         );
     }
 
+    /// **The tool-origin witness, through the *shipped* composition.**
+    ///
+    /// The stagnation rung exempts a tool that came from outside the binary,
+    /// and it learns that from `ToolExecutor::tool_origin`. The answer starts
+    /// three or four layers down — the registry knows its catalog rows, the
+    /// MCP set knows its namespace, the custom set knows its manifest — and
+    /// has to survive every decorator above it. The port's default is `None`,
+    /// so a layer that forgets to forward reports "unknown" and the exemption
+    /// silently stops reaching the session that needed it.
+    ///
+    /// Asserted through `session_stack_with_gate` with the skill plane on
+    /// top, which is the chain a turn driver mounts, for the same reason the
+    /// gate witness above is: a future decorator that forgets to forward
+    /// fails here, and nowhere else.
+    #[tokio::test]
+    async fn the_production_tool_stack_forwards_tool_origin() {
+        use stella_core::loop_detect::ToolOrigin;
+        use stella_tools::skill_plane::{SkillInvocationPlane, SkillScopedTools};
+
+        let dir = tempfile::tempdir().unwrap();
+        let registry = Arc::new(stella_tools::registry::ToolRegistry::new(
+            dir.path().to_path_buf(),
+        ));
+        let mut client = stella_mcp::McpClient::new(
+            "vendor",
+            Box::new(CannedTransport {
+                called: Arc::new(std::sync::Mutex::new(false)),
+            }),
+        );
+        client.initialize().await.unwrap();
+        let mcp = stella_mcp::McpToolSet::from_clients(vec![client])
+            .wrapping(registry.clone() as Arc<dyn ToolExecutor>);
+
+        let stack = session_stack_with_gate(
+            &mcp,
+            vec![script_tool(dir.path())],
+            dir.path().to_path_buf(),
+            ToolPolicy::allow_all(),
+            session_gate(dir.path()),
+            Principal::User,
+        );
+        let view = SkillScopedTools::new(&stack, SkillInvocationPlane::new());
+
+        assert_eq!(
+            view.tool_origin("task_list"),
+            Some(ToolOrigin::Builtin),
+            "a catalog row is a built-in and the rung must keep firing for it"
+        );
+        assert_eq!(
+            view.tool_origin("mcp__vendor__deploy"),
+            Some(ToolOrigin::Mcp),
+            "a server's tool, whose constant ack is its own design"
+        );
+        assert_eq!(
+            view.tool_origin("my_tool"),
+            Some(ToolOrigin::Custom),
+            "a .stella/tools script, which may print one constant line"
+        );
+        assert_eq!(
+            view.tool_origin("no_such_tool"),
+            None,
+            "a name nothing registered is unknown, not a built-in"
+        );
+    }
+
     /// **The skill-invocation witness, through the shipped composition.**
     /// The skill invocation plane composed over the assembled session chain — the
     /// position every turn driver mounts it at — is exactly the
--- a/crates/stella-cli/src/claims.rs
+++ b/crates/stella-cli/src/claims.rs
@@ -617,6 +617,12 @@ impl ToolExecutor for ClaimTap<'_> {
     fn dispatch_gate(&self) -> Option<&dyn stella_core::ports::DispatchGate> {
         self.inner.dispatch_gate()
     }
+
+    /// Forwarded. The tap claims files. It owns no tool name of its own, so
+    /// every origin it can give back is the base's.
+    fn tool_origin(&self, name: &str) -> Option<stella_core::loop_detect::ToolOrigin> {
+        self.inner.tool_origin(name)
+    }
 }
 
 #[cfg(test)]
--- a/crates/stella-cli/src/command_deck/task_tap.rs
+++ b/crates/stella-cli/src/command_deck/task_tap.rs
@@ -187,6 +187,12 @@ impl ToolExecutor for TaskTap<'_> {
         self.inner.dispatch_gate()
     }
 
+    /// Forwarded. The tap mirrors task updates. It owns no tool name of its
+    /// own, so every origin it can give back is the base's.
+    fn tool_origin(&self, name: &str) -> Option<stella_core::loop_detect::ToolOrigin> {
+        self.inner.tool_origin(name)
+    }
+
     /// Forwarded: the deck's lead lane wraps the discovery mount (which owns
     /// the invocation plane) in this tap, so a tap that let the empty default
     /// stand would silently stop active skill bodies surviving summarization
--- a/crates/stella-cli/src/fleet_commits.rs
+++ b/crates/stella-cli/src/fleet_commits.rs
@@ -167,6 +167,12 @@ impl ToolExecutor for CommitObserver<'_> {
     fn dispatch_gate(&self) -> Option<&dyn stella_core::ports::DispatchGate> {
         self.inner.dispatch_gate()
     }
+
+    /// Forwarded. This observer watches commits. It owns no tool name of its
+    /// own, so every origin it can give back is the base's.
+    fn tool_origin(&self, name: &str) -> Option<stella_core::loop_detect::ToolOrigin> {
+        self.inner.tool_origin(name)
+    }
 }
 
 /// One attempt's commits, oldest first — and the two ways of knowing which
--- a/crates/stella-cli/src/tool_policy.rs
+++ b/crates/stella-cli/src/tool_policy.rs
@@ -142,6 +142,13 @@ impl ToolExecutor for PolicyToolSet<'_> {
     fn dispatch_gate(&self) -> Option<&dyn stella_core::ports::DispatchGate> {
         self.inner.get().dispatch_gate()
     }
+
+    /// Forwarded, and NOT narrowed, as the gate above is not. A tool that is
+    /// switched off cannot run again. A call it made before the switch
+    /// flipped is still in the window, and the rung still reads its origin.
+    fn tool_origin(&self, name: &str) -> Option<stella_core::loop_detect::ToolOrigin> {
+        self.inner.get().tool_origin(name)
+    }
 }
 
 /// Which `"tools"` key withheld `name` — the exact name, its group, or the
--- a/crates/stella-core/src/driver/loop_escalation.rs
+++ b/crates/stella-core/src/driver/loop_escalation.rs
@@ -56,7 +56,7 @@ use crate::loop_detect::{CallRecord, LoopIdentity, LoopVerdict, detect_loop};
 use crate::ports::ToolExecutor;
 use crate::step::AbortKind;
 
-use super::loop_evidence::{recent_call_records, turn_start_index};
+use super::loop_evidence::{recent_call_records, stamp_origins, turn_start_index};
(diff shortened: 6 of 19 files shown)


macanderson:
**`fmt + clippy + test` is red on `df01913`, and it is a one-line compile error, not a test failure.**

[Job 100510340001](https://github.com/macanderson/stella/actions/runs/33710997318/job/100510340001), in the `cargo test --workspace` step at 03:27:01:

```
error: invalid format string: expected `}`, found `.`
   --> crates/stella-core/src/loop_detect/tests/origin.rs:147:9
    |
147 |         prop_assert!(!matches!(detect_loop(&records, config), LoopVerdict::Stagnant { .. }));
    |         ^^^^ expected `}` in format string
    |
    = note: if you intended to print `{`, you can escape it using `{{`
    = note: this error originates in the macro `::core::concat` which comes from
            the expansion of the macro `prop_assert`
error: could not compile `stella-core` (lib test) due to 1 previous error
```

**Cause.** Single-argument `prop_assert!` builds its failure message as `concat!("assertion failed: ", stringify!($cond))` and hands that to `format!`. The stringified condition contains `Stagnant { .. }`, so `{ .. }` is parsed as a format placeholder and `.` is not a valid conversion. This is a `prop_assert!`-only hazard — plain `assert!` does not route its condition through a format string, which is why the three `assert!`/`matches!` uses earlier in the same file compile.

**Fix.** Pass an explicit message, which selects the second macro arm and stops the condition being stringified into the format string:

```rust
prop_assert!(
    !matches!(detect_loop(&records, config), LoopVerdict::Stagnant { .. }),
    "a tool from outside the binary must never reach Stagnant"
);
```

The sibling property in the same block is unaffected — `prop_assert_eq!(detect_loop(&marked, config), detect_loop(&records, config))` has no braces to misparse.

Nothing else in the job is red: the release build, `test-self-driving.sh` (61 checks), `test-impacted-crates.sh` (25 checks) and `context validate` all pass after it, and `dod / dod` went green at 03:26. This compile error is the only thing between this PR and a green `fmt + clippy + test`.

Diagnosed but not pushed: this branch is yours and in flight, and #3328 carries your claim, so patching it from another session is the duplicate-work hazard `scripts/issue-claim.sh` exists to prevent (#5224).

---

Issue #5722, opened 2026-09-03
observatory: self-rating panels now only see turns that went badly


macanderson:
## In plain words

The dashboard has panels that show how the model rates its own work ("AVG SELF-RATING," "DELIVERED," "what to improve"). A recent change means the model now only grades itself on turns that had some kind of trouble. Clean, successful turns get no grade at all. So the panels only show scores from the rough turns — which makes the average look worse than it really is, and readers can't tell that's happening.

## Problem

The model's self-review rides in the same call as lesson mining, so the reflection gate now decides both. Since #2155 a successful turn whose event ledger recorded no friction spends no reflection call at all, and therefore writes no `execution_reflection` row.

The Observatory's AVG SELF-RATING, DELIVERED and "what to improve" panels select from that table (`crates/stella-observatory/src/db.rs`). Their population is now the subset of turns that failed or hit friction, which is a biased sample of exactly the wrong kind: the average self-rating over rough turns only is not the average self-rating.

`crates/stella-cli/src/memory/learning.rs`'s `reflect_and_record` writes the self-review first, before any of the lesson-path early returns, and its comment says why — it answers a different question than the lessons do and survives on its own. The gate sits one layer above that, at `reflect_routed`, so the argument no longer reaches the case it was written for.

## The decision this issue asks for

Three answers are defensible and someone should pick one:

1. **Accept it.** Grade the turns that had something to say and change the panel labels to match. Cheapest, and honest as long as the labels say so.
2. **Sample.** Reflect on a fraction of clean turns so the panels keep an unbiased population at a bounded cost. Needs a rule that is deterministic per workspace, not a coin flip — `ab_control`'s durable counter (`crates/stella-cli/src/memory/tests/ab_control.rs`) is the shape already in the tree for "every nth turn."
3. **Separate the two calls.** A self-review is short; a cheap-tier call asking only for it would restore the population without restoring the lesson-mining spend #2155 removed. This costs the most and is the only one that keeps both properties exactly.

## Files

- `crates/stella-cli/src/memory/reflection.rs` — `warrants_reflection`, `reflect_routed`
- `crates/stella-cli/src/memory/learning.rs` — `reflect_and_record`, the self-review write
- `crates/stella-observatory/src/db.rs` — the panels that read `execution_reflection`

## Verify

Run several clean successful tool-using turns in a workspace, then read `execution_reflection`: today it gains no rows for them.

## Definition of done

[...]

Issue #5721, opened 2026-09-03
cli: stop keeping a whole turn's event history in memory just to count friction


macanderson:
## In plain words

When Stella's CLI runs a turn, it keeps every event from that turn in memory (including the full contents of any file the agent read) just so it can count up three numbers about the turn at the end — even for output formats that never print those events. That means reading one large file makes the CLI hold that whole file in memory until the turn is done. The fix: count those three numbers as the events come in and throw each event away right after, instead of hoarding all of them.

## Problem

`spawn_renderer` now retains every `AgentEvent` of a turn for every output format, because the friction fold downstream needs them. Only `OutputFormat::Json` needs the full vector, and only to print the JSON envelope; the other two formats keep it purely so `run_turn` can fold three counts out of it.

The retained vector holds whole tool results, so a turn that reads a large file keeps that file's bytes in memory until the turn ends. `crates/stella-cli/src/agent/persistence.rs`:

```rust
outcome.events.push(event);
```

`crates/stella-cli/src/command_deck/forwarder.rs` already shows the cheaper shape for the same question — it folds `friction.observe(&event)` as events stream and retains nothing.

## Fix direction

Fold a `TurnFriction` inside `spawn_renderer` as events arrive and return it on `RendererOutcome`, alongside the event vector that `OutputFormat::Json` still needs for its envelope. `run_turn` then reads `rendered.friction` instead of calling `TurnFriction::from_events(&collected)`, and the vector can go back to being collected only where it is printed.

`TurnFriction::from_events` is already a fold over `observe` in stream order and reads every duration off the event's own `duration_ms` field rather than a clock, so folding live is byte-identical to folding after — the argument `TurnFriction::from_events`'s own doc comment makes.

## Files

- `crates/stella-cli/src/agent/persistence.rs` — `spawn_renderer`, `RendererOutcome`
- `crates/stella-cli/src/agent.rs` — `run_turn`'s `TurnFriction::from_events(&collected)` (a god file at its ceiling; keep the change net-zero or negative)
- `crates/stella-cli/src/fleet_cmd.rs` — a second caller that folds from `rendered.events`
- `crates/stella-cli/src/memory/reflection/tests.rs` — `the_reflecting_doors_fold_a_ledger_and_reflect_with_it` asserts the current call by source text and will need its assertion moved with the code
- `crates/stella-cli/src/agent/persistence/tests/collected_journal.rs` — the test that pins every format's ledger today

## Verify

[...]

Issue #5720, opened 2026-09-03
cli: the Command Deck never reflects on a turn that failed


macanderson:
## In plain words

The Command Deck is the main screen you see when you use Stella. After each turn, Stella is supposed to look back and learn from what happened — especially when a turn fails, since failures teach the most. Every other part of Stella does this. The Command Deck does not: right now it quits early and skips the lesson whenever a turn fails. This issue fixes the Command Deck so it learns from failures too, the same way the rest of Stella already does.

## Problem

The Command Deck is the only reflecting door that never reflects on a failed turn.

`crates/stella-cli/src/command_deck/authoring.rs`'s `record_and_reflect_turn` returns early when the turn failed:

```rust
if outcome.is_err() || !turn_warrants_reflection(turn) {
    return;
}
```

and then builds its evidence with `succeeded: true` hard-coded:

```rust
crate::memory::TurnEvidence::with_friction(turn, friction, true),
```

Every other door does the opposite. `crates/stella-cli/src/agent/reflect.rs`, `crates/stella-cli/src/agent/goal.rs` (both doors) and `crates/stella-cli/src/fleet_cmd.rs` all gate on `should_reflect_on(&result)`, which admits a failure and refuses only a user's soft stop, and all of them pass `result.is_ok()` so the failure prompt is the one that runs.

`crates/stella-cli/src/memory.rs`'s module doc calls a failed turn the highest-value learning signal, and `reflect_on_turn`'s failure prompt exists for it. The Command Deck is the default interactive shell on a TTY, so this is the door most human turns go through, and none of its failures teach anything.

## Files

- `crates/stella-cli/src/command_deck/authoring.rs` — `record_and_reflect_turn`
- `crates/stella-cli/src/command_deck.rs` — the lead-turn call site (a god file at its ceiling; any change there must be net-zero lines)
- `crates/stella-cli/src/agent/reflect.rs` — `reflect_on_interactive_turn`, the shape to copy

## Verify

A deck turn that fails writes no line to `.stella/private/reflections.jsonl` and no row to `execution_reflection`. The same prompt through `stella run` does write both.

Narrow local check (SCR-001 — do not run the full suite): `cargo test -p stella-cli` scoped to the new/updated test in `crates/stella-cli/src/command_deck/` or `crates/stella-cli/src/memory/reflection/tests.rs`, whichever the fix lands the witness in.

## Constraints

[...]

Issue #3946, opened 2026-08-19
reflection's friction ledger now has ZERO production producers — every shipped digest carries an empty friction section


macanderson:
Found while finishing #3872 (PR on `worktree-dead-code-repoint`).

## The problem

`TurnFriction` (`crates/stella-cli/src/memory/reflection/digest.rs`) splits cleanly in two:

- **Render half** — `section()`, `call_names()`, `is_empty()`, `empty()`. Live: `build` renders the friction section from it, and `Priority::Friction` uses it to decide which transcript messages are admitted to the digest.
- **Ingest half** — `observe()`, `close()`, `push_retry()`, plus `FRICTION_ENTRY_CAP`, the `open` field and `role_token()`. **No production caller.**

Its one production caller was `FrictionTap` in `crates/stella-cli/src/agent/reflect.rs`, which wrapped the staged pipeline's event stream and was deleted with that crate (#3865).

Every production path builds evidence through `TurnEvidence::from_transcript`, which hardcodes `friction: TurnFriction::empty()` — a shared, permanently empty `OnceLock`. So **on every surface, the friction section of every reflection digest is empty**, and `Priority::Friction`'s selection loses the event-derived half of its input.

This is strictly worse than the state #2483 was filed and closed in: that issue said the ledger was wired on one surface of four. It is now wired on zero.

## Why it was not fixed in the #3872 PR

That PR gated the ingest half `#[cfg(test)]` rather than deleting it, precisely so this issue has something to wire. Actually wiring it is a feature, not a cleanup: the raw loop's events ride a bare `tokio::sync::mpsc::UnboundedSender` created inside the turn helper, consumed by a renderer task. Producing a ledger means either teeing that stream or folding it in the renderer and threading the result back out to the reflection call — across a function boundary, for three doors (one-shot/`run_raw_one_shot`, the REPL, the deck), each of which calls `reflect_routed` at a different place. It also changes what reflection is billed for, which deserves its own read.

## Files

- `crates/stella-cli/src/memory/reflection/digest.rs` — `TurnFriction`, `TurnEvidence::from_transcript`
- `crates/stella-cli/src/agent/reflect.rs` — where the deleted `FrictionTap` lived; `reflect_on_interactive_turn`
- `crates/stella-cli/src/agent/goal.rs` — `run_raw_one_shot`'s reflection block
- `crates/stella-cli/src/agent.rs` — the turn helper that owns the event channel
[...]

PR #5723, opened 2026-09-03
feat(stella-cli): reflect only on turns that hit something

### LINKED ISSUE #2155: cli: reflection still fires on every tool-using turn — gate on failure/retry/loop friction signals
## Problem

#1847 routed post-turn reflection to the cheap triage tier and pinned `effort: Low`, which removes the worker-tier cost. What it deliberately did **not** change is the fire gate: `turn_warrants_reflection` (`crates/stella-cli/src/memory/reflection.rs`) still fires reflection after ANY turn whose transcript contains a tool call (plus the one-shot pipeline path's `!files.is_empty()` fallback in `crates/stella-cli/src/agent.rs`). The reflection prompt itself instructs that most turns should return an empty list, so most of these calls are spent discovering there is nothing to learn — now at cheap-tier prices, but still one model call per tool-using turn, on every surface.

#1847's fix direction suggested a cheaper gate: fire only on turns with a failure/retry/loop signal, "which is where lessons live."

## Why it was deferred out of #1847

Changing the gate changes *learning semantics*, not just cost:

- The success-path prompt asks "what SURPRISED you?" — a wrong attempt inside an ultimately-successful turn is exactly a surprise worth mining, and a failure-only gate loses it.
- No structured per-turn friction signal reaches the CLI's reflection gate today. The call sites have `result: Result<..>` and the message transcript; retries, loop-detector trips, and tool-error counts are computed in the engine (`stella-core`: retry history, loop detection) but are not surfaced to the drivers at reflection time. A transcript heuristic ("does this tool result read like an error?") would be fragile in exactly the way this repo refuses.

## Fix direction

1. Surface a per-turn friction summary from the engine to the drivers — retries observed, loop-detector trips, tool errors — on the turn outcome the drivers already receive. The engine computes all three; this is plumbing, not new detection.
2. Gate reflection on `!succeeded || friction` instead of "any tool call". Keep the failure path unconditional (a failed turn is the highest-value learning signal — `memory.rs` module doc) and keep `should_reflect_on`'s soft-stop exclusion.
3. `STELLA_DISABLE_REFLECTION` semantics unchanged.

## Verify

Unit tests on the new gate: a clean successful turn with tool calls but no friction does NOT reflect; a successful turn with a retry/loop/tool-error signal does; a failed turn always does (except a user soft stop). Witness: the first of these fails on current code, where any tool-using turn reflects.

## Constraints

- Since #1847's fix, all reflecting surfaces dispatch through one seam, `reflect_routed` (`crates/stella-cli/src/memory/reflection.rs`) — apply the gate at or beside that seam, not per-driver.
- `crates/stella-cli/src/agent.rs` and `crates/stella-cli/src/command_deck.rs` are god files at their exact file-size ceilings — any call-site change there must be net-zero lines.
- Reflection is best-effort by contract: the gate must never fail or slow the turn it describes.

Refs #1847

## Definition of done

- [x] The turn outcome the drivers receive carries per-turn friction counts (retries, loop trips, tool errors) taken from data the engine already emits, with no new detection.
- [x] `reflect_routed` spends a model call only when the turn failed or its friction counts are non-zero; every reflecting door inherits that from the one seam.
- [x] A user soft stop still never reflects, and `STELLA_DISABLE_REFLECTION` behaves exactly as before.
- [x] `crates/stella-cli/src/agent.rs` and `crates/stella-cli/src/command_deck.rs` change by zero lines.
- [x] Unit tests cover the four cases: clean successful tool-using turn does not reflect, a retry reflects, a loop trip reflects, a tool error reflects, a failed turn reflects.
- [x] An end-to-end test drives the real binary and shows a clean successful turn making no reflection model call, paired with the existing run that does.
- [x] CI is green on the pull request.

macanderson:
## What

Reflection fired after every turn that used a tool. The reflection prompt itself tells the model that an empty list is a complete answer, so most of those calls were spent discovering there was nothing to learn.

`reflect_routed` — the one seam all four reflecting doors dispatch through — now asks whether the turn earned the call before it resolves a route or opens a provider:

```rust
if !warrants_reflection(&evidence) {
    return ReflectionReport::default();
}
```

A turn earns it when it did not succeed, or when its folded event ledger recorded a retry, a loop-detector firing, or a failed tool call. A wrong attempt inside a turn that ended well is exactly the surprise the success prompt asks about, so this is not a failure-only gate. A user's soft stop never reaches the seam (`should_reflect_on` drops it at every call site) and `STELLA_DISABLE_REFLECTION` is untouched.

## Why

Measured on three live turns on 2026-08-08, from `.stella/private/store.db`'s `telemetry` table:

| call role | model | calls | in | out | cost |
|---|---|---:|---:|---:|---:|
| worker | deepseek/deepseek-v4-flash-0731 | 6 | 60,376 | 268 | $0.001957 |
| reflection | google/gemini-2.5-flash-lite | 1 | 763 | 490 | $0.000272 |
| reflection | anthropic/claude-haiku-4.5 | 1 | 907 | 507 | $0.003442 |
| reflection | deepseek/deepseek-v4-flash-0731 | 1 | 830 | 320 | $0.000128 |

Reflection emitted 320-507 output tokens per turn against the worker's 268 across all six of its steps, and one reflection cost 8.8x the turn it described. Routing (#1847) bounds which model pays; it cannot decline a call the turn had no lesson to give.

## Where the design pointer met the tree

The pointer asked for a `TurnFriction { retries, loop_trips, tool_errors }` plumbed from `stella-core` onto the turn outcome. The tree has moved since the issue was written: since #3946 every raw-loop door already folds its own `AgentEvent` journal into a `TurnFriction` and hands it to reflection as `TurnEvidence::friction`. Adding a second path from the engine would be a second source of truth for the same three numbers, so the gate reads the ledger that is already there. Consequences: no `stella-protocol` type crosses a crate boundary (no serde round-trip owed, invariant #4), and nothing rides an event (no consumers-ledger row owed, invariant #10). Everything the pointer named is present — the counts, the gate, the unchanged opt-out, net-zero lines at `agent.rs` and `command_deck.rs` (both untouched).

## The defect this uncovered

`spawn_renderer` kept the turn's events **only** for `OutputFormat::Json`:

```rust
OutputFormat::Json => outcome.events.push(event),
```

`run_turn` folds exactly that vector into the turn's `TurnFriction`, so on the interactive door and on every `--output-format stream-json` run the ledger was empty: no retries, no loop firings, no failed calls, whatever the turn did. #3946's source-level test asserts `run_turn` contains `TurnFriction::from_events(&collected)`, which is true and says nothing about whether `collected` holds anything. Gating on a signal that is structurally always zero would have switched reflection off outright for those two surfaces, so the push moved out of the match and now runs for every format. Nothing else changes: the JSON envelope is still printed only under `format == OutputFormat::Json`.

## Witnesses, and how each one fails on `main`

- `crates/stella-cli/tests/run_reflection_writes_cli.rs::a_clean_successful_turn_buys_no_reflection_call` — drives the real `stella` binary against a wiremock provider whose one tool call (`get_environment`) succeeds, and asserts **zero** requests carrying the reflection system prompt and an empty mining log. On `main` any tool-using turn reflects, so the count is 1 and the log has a line. Its pair, `a_non_interactive_json_run_writes_its_lesson_to_the_mining_log`, now fails its tool call (`read_file` on a path the seeded workspace does not have) and still asserts exactly one reflection call and one logged lesson. The two differ only in whether the turn's tool call succeeded: a build that always reflects fails the first, one that never reflects fails the second.
- `crates/stella-cli/src/memory/reflection/tests.rs` — five unit tests on the gate. `a_clean_successful_turn_earns_no_reflection_call` cannot compile on `main`, where `warrants_reflection` does not exist; it also asserts `turn_warrants_reflection` still admits the same turn, so the second assertion is about the friction gate alone. Then: each of the three engine signals earns the call on its own; a failed turn reflects on an empty ledger while a soft stop is refused; one rough round earns a whole `/goal` arc its call and an all-clean arc earns none; an overflowed ledger is never read as a clean turn. A source-level test pins the gate ahead of the route lookup, the same technique `the_reflecting_doors_fold_a_ledger_and_reflect_with_it` already uses in this file.
- `crates/stella-cli/src/agent/persistence/tests/collected_journal.rs::every_format_keeps_the_turns_journal` — drives `spawn_renderer` over one failed tool call in all three formats and folds what it returns. On `main` `Text` and `StreamJson` return nothing and `counts().tool_errors` is 0 for both.

## Exemplar followed

The gate's placement copies this crate's own `reflect_routed` argument for the routing decision — one seam rather than four call sites — and the source-level wiring assertions copy `the_reflecting_doors_fold_a_ledger_and_reflect_with_it` (`crates/stella-cli/src/memory/reflection/tests.rs`) and `a_fleet_attempt_reflects_before_its_spend_is_read` (`crates/stella-cli/src/fleet_cmd/tests.rs`).

## Fixture updates (behaviour change, not test weakening)

Two existing fixture sites built a successful turn with an empty ledger, which is now exactly the "went straight through" case:

- `crates/stella-cli/src/fleet_cmd/tests.rs` — `mined_friction()` gives the two attempt-mining tests a ledger with the failed `make migrate` their transcript already describes.
- `crates/stella-cli/tests/run_reflection_writes_cli.rs` — the reflecting run's tool call now fails, as described above.

## Also in this diff

`TurnFriction::push_retry` carried an `#[allow(dead_code)]` whose stated reason ("`observe` has no production caller") stopped being true when #3946 wired `from_events`. The attribute and its comment are gone and `scripts/dead-code-allows-baseline.txt` drops `stella-cli` to zero — a ratchet tightening, not a new entry.

Closes #2155

Tests deleted: None

Remaining work filed:

- #5720 — the Command Deck never reflects on a failed turn. Found while auditing every reflecting door against the shared rule; it predates this change and is not caused by it.
- #5721 — fold the friction as events stream rather than retaining the whole journal on formats that never print it. The cheaper shape for the collection fix this PR makes.
- #5722 — the Observatory's self-rating panels now see only turns that failed or hit friction, because the self-review rides the same call this PR gates. Three defensible answers, one of which a human should pick.

## Summary by Sourcery

Reflect only on turns that fail or record meaningful friction while retaining complete event evidence across all CLI output modes.

Bug Fixes:
- Prevent reflection calls for successful tool-using turns that recorded no retries, loop detections, or tool failures.
- Preserve turn event journals across text, JSON, and stream-JSON output modes so reflection decisions receive accurate friction data.

Enhancements:
- Gate reflection at the shared dispatch seam using turn success, friction counts, and journal truncation status before route or provider resolution.
- Expose aggregated friction counts and truncation state from turn ledgers for reflection eligibility checks.
- Tighten dead-code allowances after retry handling became production-reachable.

Tests:
- Add unit, renderer, fleet, and end-to-end CLI coverage for clean-turn suppression, friction-triggered reflection, overflowed ledgers, and journal retention across output formats.

### DIFF
--- a/crates/stella-cli/src/agent/persistence.rs
+++ b/crates/stella-cli/src/agent/persistence.rs
@@ -348,7 +348,10 @@ pub(crate) fn spawn_renderer(
                 // does, terminate before the provider loop can spend on a
                 // later unmetered call.
                 OutputFormat::StreamJson => emit_stream_json(&event, durable_pre_persisted, &clock),
-                OutputFormat::Json => outcome.events.push(event),
+                // Rendered by the envelope `run_turn` builds out of
+                // `RendererOutcome::events` below, so this arm prints nothing
+                // of its own.
+                OutputFormat::Json => {}
                 // One frame per turn, folded from the same events the cards
                 // below would each have printed. The two cannot both run: the
                 // frame restates every row, so a surface doing both prints the
@@ -412,6 +415,14 @@ pub(crate) fn spawn_renderer(
                     other => plain::render_event(other),
                 },
             }
+            // Kept for every format, not only `Json`. `run_turn` folds this
+            // vector into the turn's `TurnFriction`, and reflection reads
+            // those counts to decide whether the turn hit anything worth a
+            // model call. While the `Json` arm alone collected, that ledger
+            // came out empty in interactive mode and on every
+            // `--output-format stream-json` run — no retries, no loop firings,
+            // no failed calls, whatever the turn actually did.
+            outcome.events.push(event);
         }
         // The stream is closed, so the turn is final: print whatever frame the
         // fold accumulated. A turn that never opened prints nothing.
--- a/crates/stella-cli/src/agent/persistence/tests.rs
+++ b/crates/stella-cli/src/agent/persistence/tests.rs
@@ -1,5 +1,6 @@
 //! Suites over [`super`]'s persistence path, one topic per file.
 
+mod collected_journal;
 mod pipeline_variant;
 mod receipt_write;
 mod stream;
--- a/crates/stella-cli/src/agent/persistence/tests/collected_journal.rs
+++ b/crates/stella-cli/src/agent/persistence/tests/collected_journal.rs
@@ -1,0 +1,65 @@
+//! What `spawn_renderer` returns once the turn's stream closes.
+//!
+//! `RendererOutcome::events` is the turn's journal. `run_turn` folds it into a
+//! `TurnFriction`. Reflection reads those counts to decide if the turn is
+//! worth a model call. A journal the renderer drops is not a smaller answer.
+//! It is a wrong one: the turn reads as one that hit nothing.
+
+use crate::agent::persistence::*;
+
+/// One announced tool call and the error that closed it — the shape the
+/// friction fold counts.
+fn failed_call() -> Vec<AgentEvent> {
+    vec![
+        AgentEvent::ToolStart {
+            call: stella_protocol::ToolCall {
+                call_id: "c1".into(),
+                name: "read_file".into(),
+                input: serde_json::json!({"path": "missing.rs"}),
+            },
+            sub_agent_id: None,
+            task_id: None,
+        },
+        AgentEvent::ToolResult {
+            call_id: "c1".into(),
+            output: stella_protocol::ToolOutput::error("no such file"),
+            duration_ms: 7,
+            speculated: false,
+            sub_agent_id: None,
+            task_id: None,
+        },
+    ]
+}
+
+/// Drive the renderer over `events` in `format` and hand back what it kept.
+async fn collected(format: OutputFormat, events: Vec<AgentEvent>) -> Vec<AgentEvent> {
+    let (tx, rx) = mpsc::unbounded_channel();
+    let renderer = spawn_renderer(rx, format, None, "anthropic".into(), false, None);
+    for event in events {
+        tx.send(event).expect("the renderer is live");
+    }
+    drop(tx);
+    renderer.await.expect("renderer").events
+}
+
+/// **The witness.** Every format keeps the turn's journal.
+///
+/// On base the `Text` and `StreamJson` arms kept none of it. Every fold over
+/// the journal then counted zero.
+#[tokio::test]
+async fn every_format_keeps_the_turns_journal() {
+    for format in [
+        OutputFormat::Text,
+        OutputFormat::Json,
+        OutputFormat::StreamJson,
+    ] {
+        let kept = collected(format, failed_call()).await;
+        let friction = crate::memory::TurnFriction::from_events(&kept);
+        assert_eq!(
+            friction.counts().tool_errors,
+            1,
+            "{format:?}: the fold must see the failed call the turn made, \
+             because the reflection gate decides on nothing else"
+        );
+    }
+}
--- a/crates/stella-cli/src/fleet_cmd.rs
+++ b/crates/stella-cli/src/fleet_cmd.rs
@@ -788,7 +788,9 @@ fn attempt_task_boundary(task: &Task) -> String {
 ///
 /// **What it costs.** One model call per attempt that warrants one, on the same
 /// terms as every other door: gated by `turn_warrants_reflection` so a tool-free
-/// turn spends nothing, bounded by this child's remaining headroom, and settled
+/// turn spends nothing and by `reflect_routed`'s own friction gate so an
+/// attempt that went straight through spends nothing either, bounded
+/// by this child's remaining headroom, and settled
 /// back into its `BudgetGuard` — so the reflection lands inside the
 /// `--spend-limit` the fleet enforces twice (the child's own cap here, and the
 /// metered total the parent stops new waves on), rather than beside it.
--- a/crates/stella-cli/src/fleet_cmd/tests.rs
+++ b/crates/stella-cli/src/fleet_cmd/tests.rs
@@ -781,6 +781,33 @@ fn mined_transcript() -> Vec<CompletionMessage> {
     ]
 }
 
+/// The ledger that goes with it: the first `make migrate` failed, so the turn
+/// carries friction and earns its reflection call. An empty ledger beside a
+/// successful turn stands for a turn that went straight through, and
+/// `reflect_routed` declines those — so a fixture that wants a lesson mined
+/// has to say what the turn ran into.
+fn mined_friction() -> crate::memory::TurnFriction {
+    crate::memory::TurnFriction::from_events(&[
+        stella_protocol::AgentEvent::ToolStart {
+            call: stella_protocol::ToolCall {
+                call_id: "c1".into(),
+                name: "bash".into(),
+                input: serde_json::json!({"command": "make migrate"}),
+            },
+            sub_agent_id: None,
+            task_id: None,
+        },
+        stella_protocol::AgentEvent::ToolResult {
+            call_id: "c1".into(),
+            output: stella_protocol::ToolOutput::error("ledger writer still running"),
+            duration_ms: 90,
+            speculated: false,
+            sub_agent_id: None,
+            task_id: None,
+        },
+    ])
+}
+
 /// **Witness (#3956).** A lesson mined by a fleet worker is readable from the
 /// primary workspace after the run — and is not written into the disposable
 /// tree the attempt ran in.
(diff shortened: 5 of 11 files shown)


macanderson:
**`dod / dod` is red on seven unchecked boxes. Six of the seven are corroborated by the diff; the seventh is the CI run that has not finished yet.**

I checked each of #2155's boxes against `origin/main...4abbc18` rather than against the PR description, so the table below is what the diff shows, not what the body claims.

| # | DoD item | Evidence in the diff |
|---|---|---|
| 1 | friction counts from data the engine already emits, no new detection | `FrictionCounts { retries, loop_trips, tool_errors }` and `TurnFriction::counts()` in `memory/reflection/digest.rs` — folded from the existing journal, no new producer. See the note below on where this lands. |
| 2 | `reflect_routed` spends a call only on failure or friction, inherited by every door | `if !warrants_reflection(&evidence) { return ReflectionReport::default(); }` sits **ahead of** `reflection_route(...)` and `discover_configured_providers()`. One seam, no call site touched. |
| 3 | soft stop still never reflects; `STELLA_DISABLE_REFLECTION` unchanged | `STELLA_DISABLE_REFLECTION` appears **0 times** in the diff. `should_reflect_on`'s body is byte-identical — the only `+` line naming it is a doc reference. |
| 4 | `agent.rs` and `command_deck.rs` change by zero lines | Neither file appears in `git diff --stat`. Mechanically confirmed. |
| 5 | unit tests for the five cases | `a_clean_successful_turn_earns_no_reflection_call`, `any_one_friction_signal_earns_the_call`, `a_failed_turn_reflects_on_an_empty_ledger`, `one_rough_round_earns_the_whole_arc_its_call`, `an_overflowed_ledger_is_never_read_as_a_clean_turn`, `the_shared_seam_asks_the_gate_before_it_spends_anything`. |
| 6 | e2e test drives the real binary on a clean turn | `a_clean_successful_turn_buys_no_reflection_call` in `tests/run_reflection_writes_cli.rs`, against a wiremock provider, paired with the reflecting run. |
| 7 | CI is green | **Not yet.** `fmt + clippy + test` was still `in_progress` at the time of writing; every other check on this head is green or in flight, and `dod` is the only failure. |

**One departure worth a reader's attention, box 1.** The box asks for the counts on *"the turn outcome the drivers receive"*. They arrive instead on `TurnEvidence::friction`, the ledger each raw-loop door already folds since #3946. The PR body argues this and I think it is right — plumbing a second path from the engine would be a second source of truth for the same three numbers — but it is a different shape from what the box describes, so it should be ticked knowingly rather than as a match.

**The renderer fix is load-bearing and is not in the issue's DoD.** `spawn_renderer` pushed events only under `OutputFormat::Json`, so on the interactive door and on `--output-format stream-json` the folded ledger was structurally empty. Gating on it without that fix would have switched reflection off outright on two of three surfaces. `every_format_keeps_the_turns_journal` covers it.

I did not tick the boxes: box 7 is still open on a run in flight, so the check stays red either way, and ticking six of seven changes nothing about the gate today. Once `fmt + clippy + test` lands green, boxes 1–6 can be ticked on the evidence above and box 7 on the run itself; `dod` then passes on a re-run with no push.

No merge conflict against `main` (`git merge-tree --write-tree` is clean), and no `hold` label.

---

macanderson:
**The `dod` blocker flagged above is cleared. This PR is green and mergeable; it is waiting only on a human approval.**

The standing comment above ([5519898362](https://github.com/macanderson/stella/pull/5723#issuecomment-5519898362)) left `dod / dod` red with boxes 1–6 unticked and box 7 ("CI is green on the pull request") open on a run still in flight. That run has since landed, so I'm closing the comment out here rather than leaving a stale red verdict as the last word.

`fmt + clippy + test` completed green at 03:39 on `4abbc18`. All seven of #2155's DoD boxes are now ticked, and `dod / dod` re-ran green at 03:53 ([job 100516210227](https://github.com/macanderson/stella/actions/runs/33712983775/job/100516210227)) with no push — exactly the sequence the earlier comment predicted.

State of `4abbc18` as of this comment:

| | |
|---|---|
| `dod / dod` | ✅ green (latest run) |
| `fmt + clippy + test` | ✅ green |
| Other checks | ✅ 18 green, 0 failing |
| Merge conflicts | none — `mergeable_state: clean` against `main` |
| `hold` label | not present |
| Unresolved review threads | none (0 threads) |
| Approving review | **none yet** |

Sourcery did not review this diff — it returned a rate-limit notice (own review budget of 250,000 diff characters exhausted, resets in ~3 days), so there is no assessment table and no ❌ rows outstanding. `@sourcery-ai review` would re-request one once the budget resets.

I did not merge: there is no approving review, and an approval is the one gate this sweep does not substitute for. Nothing else is blocking.

PR #5732, opened 2026-09-03
fix(stella-cli): render converted command prompts as a TOML literal string

### LINKED ISSUE #5481: stella commands convert writes a TOML basic string, mangling backslashes and breaking on jq escapes
`stella commands convert` writes the command body as a TOML **basic** multi-line string:

```toml
prompt = """
...
"""
```

A basic string applies escape processing. Two consequences, both silent:

**1. Trailing-backslash line continuations are eaten.** In TOML, a `\` at end-of-line inside a basic string is a line continuation that swallows the newline and leading whitespace. Every multi-line shell example in a command body is joined onto one line. Observed converting a command containing:

```sh
gh issue list -R "$REPO" --label triage --state open --limit 500 \
  --json number,title,body,labels,createdAt,updatedAt \
  > "$WORK/triage-issues.json"
```

which round-tripped as a single joined line. The command still *runs*, so nothing looks wrong — it only silently stops matching its markdown source.

**2. `\(` is an invalid escape and breaks parsing outright.** A body containing a jq expression like `"#\(.number)"` produces a TOML file that does not parse. **This repo's own `.stella/commands/backlog-triage.md` contains two such expressions**, so converting it with the shipped subcommand would produce a broken TOML.

### Already known, in a neighbouring file

`.claude/mk-command-toml.py` uses a **literal** string (`'''…'''`) and its comment states the reason explicitly:

> A TOML literal string passes bytes through with no escape processing, which is mandatory here: the body is full of jq expressions like `"#\(.number)"` and `\(` is an INVALID escape in a basic (`"""..."""`) string.

So the correct approach is already written down — the shipped `convert` path just doesn't use it.

### Suggested fix

Emit `prompt = '''…'''` in `commands_cmd.rs`, and reject with a clear error when the body itself contains `'''` (the one sequence a literal string cannot carry). A round-trip assertion — parse the emitted TOML back and assert `prompt == body` byte-for-byte — would have caught both failures; `mk-command-toml.py` already does exactly that.

### Reproduce

```sh
printf -- '---\ndescription: x\n---\nrun:\n\n```sh\nfoo \\\n  bar\n```\n' > /tmp/c/t.md
stella commands convert /tmp/c
python3 -c "import tomllib;print(repr(tomllib.load(open('/tmp/c/t.toml','rb'))['prompt']))"
# the two shell lines come back joined
```

Found while adding a command to `.stella/commands/` (#5480), where the byte-for-byte round-trip check failed.

## Definition of done

- [x] `stella commands convert` emits `prompt = '''…'''` (a TOML literal
      string), not a basic string.
- [x] A body with a trailing-backslash line continuation round-trips
      byte-for-byte (the newline is not eaten).
- [x] A body containing a jq interpolation like `"#\(.number)"` round-trips
      byte-for-byte and produces TOML that parses.
- [x] A body containing the sequence `'''` is rejected with a clear error
      instead of silently producing invalid or truncated TOML.
- [x] A round-trip assertion (parse the emitted TOML back and compare to the
      source body byte-for-byte) covers both cases above.
- [x] The reproduce script in this issue no longer shows the two shell lines
      joined into one.

Ticked against the diff of #5732 at `7b52ff8`, verified item by item — see the
verification comment on this issue for the command output, including this
issue's own reproduce script run against a binary built from that branch.

macanderson:
## What

`stella commands convert` now writes each converted command's prompt as a
TOML **literal** string (`'''…'''`) instead of a basic string
(`"""…"""`).

## Why

A basic string applies escape processing. Two silent consequences:

1. A trailing backslash at end-of-line is a line continuation and swallows
   the newline, joining every multi-line shell example in the body onto one
   line. The command still runs, so nothing looks wrong.
2. `\(`, the start of an ordinary jq interpolation like `"#\(.number)"`, is
   not a valid basic-string escape, so the emitted file fails to parse
   outright.

A literal string applies no escape processing, so the body lands
byte-for-byte. Its one limit is the sequence `'''`, which it cannot carry —
a body containing that sequence is now rejected with a clear error instead
of being written as a file that silently truncates or fails to parse.

`.claude/mk-command-toml.py` (a local, gitignored scratch tool, not part of
this repo) already used a literal string for exactly this reason; this
brings the shipped `stella commands convert` subcommand in line with it.

## Witness

Two new tests reproduce each silent failure on the old basic-string
renderer, round-tripping the emitted TOML back through
`command_from_toml` and asserting the body survives byte-for-byte:

- `a_trailing_backslash_line_continuation_survives_conversion` — fails on
  `main` (the backslash-newline is eaten, joining the two lines); passes
  with this change.
- `a_jq_interpolation_escape_survives_conversion` — fails on `main` (the
  emitted TOML does not parse); passes with this change.

A third test, `a_command_whose_body_cannot_render_is_not_converted`, covers
the new `'''`-rejection path on both a real write and a `--dry-run`, and
`a_body_containing_triple_single_quotes_is_rejected` covers `to_toml`
directly.

Tests deleted: None.

Closes #5481

## Verification (added after review of #5481's DoD)

The original description said the tests were not run locally and deferred to
CI. They have since been run, together with this issue's own reproduce script:

```
$ cargo test -p stella-cli --bins commands_cmd
test result: ok. 10 passed; 0 failed; 0 ignored; 0 measured; 2625 filtered out
exit code 0

$ cargo build -p stella-cli          # exit 0
$ ./target/debug/stella commands convert /tmp/c
wrote /tmp/c/t.toml
1/1 converted — the markdown originals are untouched

$ python3 -c "import tomllib;print(repr(tomllib.load(open('/tmp/c/t.toml','rb'))['prompt']))"
'run:\n\n```sh\nfoo \\\n  bar\n```\n'
```

The two shell lines come back separate, backslash intact — #5481's stated
symptom is gone.

Fail side, on the same inputs, rendering what `main`'s `to_toml` emits and
parsing both with `tomllib`:

| Body | `main` (basic string) | this PR (literal string) |
|---|---|---|
| `foo \` + `  bar` | `['run:', '', '```sh', 'foo bar', '```']` — **joined** | `['run:', '', '```sh', 'foo \\', '  bar', '```']` |
| `jq -r '"#\(.number)"'` | `PARSE ERROR: Unescaped '\' in a string` | parses |

All six of #5481's DoD items are now ticked, each against the diff plus one of
the commands above. Full record: the verification comment on #5481.

## Summary by Sourcery

Render converted command prompts as TOML literal strings and safely reject bodies that cannot be represented.

Bug Fixes:
- Preserve converted command prompt bodies byte-for-byte by rendering them as TOML literal strings, avoiding backslash line continuation and invalid escape parsing issues.
- Reject command bodies containing triple single quotes instead of generating invalid or corrupted TOML files.

Enhancements:
- Report unrenderable prompts consistently during both normal conversion and dry runs without writing invalid output.

Tests:
- Add round-trip coverage for trailing backslashes and jq interpolation escapes.
- Add coverage ensuring bodies containing triple single quotes are rejected and not converted.

## DoD re-check

#5481's six DoD boxes are ticked and verified — the verification comment on that
issue carries the command output, including this issue's own reproduce script run
against a binary built from `7b52ff8` and a fail-side control showing both
findings reproduce on `main`'s basic-string rendering. The `dod` check last ran
at 16:13Z, before the boxes were ticked, and was red on that stale reading; this
edit re-triggers it (`dod-check.yml` fires on `pull_request: edited`).

Every other check on this head is green, and `cargo test` passed in
`fmt + clippy + test`.

---

### DIFF
--- a/crates/stella-cli/src/commands_cmd.rs
+++ b/crates/stella-cli/src/commands_cmd.rs
@@ -133,11 +133,24 @@ fn truncate(text: &str, max: usize) -> String {
 /// hint" rather than "this command has no hint", and the two mean different
 /// things to both a reader and the parser.
 ///
-/// `prompt` always uses a multi-line basic string. A body containing `"""` is
-/// the one shape that would break out of it, so those are escaped; everything
-/// else is passed through byte-for-byte, because a converted prompt that is
-/// not the original prompt is a silently changed command.
-pub fn to_toml(cmd: &CommandDef) -> String {
+/// `prompt` always uses a multi-line **literal** string (`'''…'''`). It
+/// applies no escape processing, so the body lands byte-for-byte. A basic
+/// (`"""…"""`) string is not safe here. It reads a trailing backslash as a
+/// line continuation and eats the newline after it, so a multi-line shell
+/// example joins onto one line. It also treats `\(` — the start of an
+/// ordinary jq expression like `"#\(.number)"` — as an invalid escape, so
+/// the file fails to parse. A literal string has neither problem. Its only
+/// limit is the sequence `'''`, which it cannot carry. A body with that
+/// sequence is rejected, the same way the loader already rejects a body
+/// that will not round-trip.
+pub fn to_toml(cmd: &CommandDef) -> Result<String, String> {
+    if cmd.body.contains("'''") {
+        return Err(format!(
+            "{}: the prompt body contains `'''`, which a TOML literal string \
+             cannot carry — rewrite the body to avoid that sequence",
+            cmd.name
+        ));
+    }
     let mut out = String::new();
     out.push_str(&format!("name = {}\n", quote(&cmd.name)));
     out.push_str(&format!("description = {}\n", quote(&cmd.description)));
@@ -154,10 +167,10 @@ pub fn to_toml(cmd: &CommandDef) -> String {
     if !cmd.model_invocable {
         out.push_str("disable-model-invocation = true\n");
     }
-    out.push_str("\nprompt = \"\"\"\n");
-    out.push_str(&cmd.body.replace("\"\"\"", "\\\"\\\"\\\""));
-    out.push_str("\n\"\"\"\n");
-    out
+    out.push_str("\nprompt = '''\n");
+    out.push_str(&cmd.body);
+    out.push_str("\n'''\n");
+    Ok(out)
 }
 
 /// A TOML basic string. Backslashes first, or the escapes added after would
@@ -223,7 +236,21 @@ pub fn convert_dir(dir: &Path, dry_run: bool, force: bool) -> Result<Vec<Convert
             });
             continue;
         }
-        if !dry_run && let Err(e) = std::fs::write(&target, to_toml(&cmd)) {
+        // A body the renderer cannot carry (see `to_toml`) is reported the
+        // same way a body the loader cannot parse is, above — never written
+        // as a silently wrong file.
+        let toml_src = match to_toml(&cmd) {
+            Ok(toml_src) => toml_src,
+            Err(e) => {
+                out.push(Converted {
+                    target,
+                    skipped: Some(e),
+                    source,
+                });
+                continue;
+            }
+        };
+        if !dry_run && let Err(e) = std::fs::write(&target, toml_src) {
             out.push(Converted {
                 target,
                 skipped: Some(format!("could not write: {e}")),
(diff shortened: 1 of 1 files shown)


macanderson:
Residue noticed while reviewing this fix: a literal string cannot carry a control character (TOML forbids everything but tab), and `to_toml` only checks for the `'''` delimiter today, not for that. Filed as #5755.

Issue #5510, opened 2026-08-30
An abandoned lease makes the telemetry spool grow without bound


macanderson:
Three findings in `stella-store`'s telemetry and egress layer, from a
workspace-wide correctness audit.

## 1. An abandoned lease makes the spool's capacity limit unenforceable — CONFIRMED

`crates/stella-store/src/enterprise_telemetry.rs`, `enforce_limits` (called from
`enqueue`).

When the spool is over `SpoolLimits::max_rows`/`max_bytes`, it evicts the oldest
row matching `sink_fingerprint = ?1 AND leased_by IS NULL`. That is the only
condition — it never considers `lease_until_ms`, unlike `claim_batch`'s own
reclaim query, which correctly treats an expired lease as free. `leased_by` is
cleared only by `ack()` or `retry()`.

**Failure scenario.** A delivery attempt calls `claim_batch` and leases a sink's
backlog for up to `MAX_LEASE_MS` (five minutes), then the host process is killed
before `ack` or `retry` runs. `leased_by` stays non-NULL indefinitely — nothing
else clears it. Every later `enqueue` finds no unleased row to evict and returns
`Ok(())` without evicting anything, while the table is over its documented hard
capacity limit. `~/.stella/enterprise-telemetry.db` then grows on every finished
execution until some future `claim_batch` happens to re-claim the stale rows,
which does honour `lease_until_ms`.

This is disk growth, not a privacy leak, and it contradicts the struct's own
"Hard capacity limits" doc and the crate's "bounded at-least-once spool" claim.

**Fix.** Have `enforce_limits` treat `lease_until_ms <= now` as evictable,
threading a `now_ms` parameter the way `claim_batch` and `retry` already take one.
**Witness.** Seed a sink at `max_rows`, `claim_batch` everything with a short
lease, advance past `lease_until_ms` without ack or retry, `enqueue` one more
event, and assert the row count stays at or under `max_rows`.

## 2. A project rekey leaves the tool-fold ledger behind — PLAUSIBLE

`crates/stella-store/src/usage/rekey.rs`, `PROJECT_KEYED_TABLES` lists
`projects`, `execution_rollup`, `telemetry` and `cloud_quarantine`.
`tool_fold_ledger` is a fifth project-keyed table (schema in
`usage/schema.rs`) and appears nowhere in `rekey.rs` — it gets none of the three
treatments the others get (move, merge, or delete), while
[...]

PR #5736, opened 2026-09-03
fix(stella-serve): report failed checkpoint discards and truncated replays

### LINKED ISSUE #5503: A failed checkpoint delete answers 200 deleted, and a truncated replay logs as a healthy completion
## The problem

Two places in `stella-serve` report a better outcome than happened: a failed
checkpoint deletion answers `200 {"status":"deleted"}`, and a truncated replay
records itself in the operator's log as a healthy turn completion.

## 1. Session delete swallows a checkpoint-removal failure

`crates/stella-serve/src/routes/sessions.rs`, `handle_session_delete`:

```rust
if let Some(store) = state.checkpoints()
    && let Ok(key) = crate::checkpoint::CheckpointKey::new(id)
{
    let _ = store.remove(&key);
}
res.json("200 OK", br#"{"status":"deleted"}"#).await
```

Every other checkpoint-store failure in the crate reports itself: `StoreSink::report`
emits `ServeEvent::CheckpointFailed`, and the dedicated
`DELETE /v1/{turns,sessions}/{id}/checkpoint` route answers `500` on a store
error. This path does neither.

**Failure scenario.** A session crashed mid-turn and has a live checkpoint. The
host calls `DELETE /v1/sessions/{id}` to clean up. `remove()` fails — disk full,
revoked permission, unmounted volume. The host is told the session is deleted and
stops tracking it, and the id 404s everywhere else. The checkpoint occupies
storage forever, invisible to `checkpoints_failed_total` and to the JSONL log at
any verbosity, with no GC to find it later. `checkpoint.rs`'s module doc warns
against exactly this — a durability feature failing quietly — and this is the one
path that does it.

**Fix.** Route the removal through the same reporting the sink uses: emit
`ServeEvent::CheckpointFailed { op: CheckpointOp::Discard, .. }` on `Err`, or
reuse `TurnCheckpoint::sink(...).discard()`.
**Witness.** Wire the existing `BrokenStore` fixture into a `ServerState`, call
`handle_session_delete`, assert a `CheckpointFailed` event was emitted or
`checkpoints_failed_total` incremented. Nothing is emitted today.

## 2. A truncated replay is logged as `TurnComplete`

`crates/stella-serve/src/routes.rs`, `write_replay`, the
`Replay::Truncated { oldest }` arm returns `Some(StreamEndReason::TurnComplete)`.
That variant is documented in `observe/event.rs` as "The terminal
`turn_complete` frame was written. The healthy ending." No such frame was
written; a `replay_truncated` frame was, and the turn is still live — the caller
in `handle_events` calls `state.park_for_resume(...)` immediately after, so it
knows.

The wire behaviour toward the host is right; only the server-side
`ServeEvent::StreamEnded { reason }` fed to the operator's JSONL log is wrong.

**Failure scenario.** A host reconnects with `?after=N` where `N` has scrolled
out of the retention ring. The operator's log records
`{"event":"stream_ended","reason":"turn_complete"}` for a turn that is still
running and now parked. Any dashboard built on this crate's own documented
diagnostic table counts it as a completion, which hides the reconnect-storm or
ring-too-small condition an operator would most want to see.

**Fix.** Add a `StreamEndReason::ReplayTruncated` variant and thread it through.
**Witness.** Drive a turn's history past its retention window, reconnect with a
stale `after`, and assert the emitted `StreamEnded` carries the new variant.

## Definition of done

- [ ] A failed checkpoint removal is reported, not swallowed, and the response
      reflects it.
- [x] A truncated replay records a reason distinct from a healthy completion.
- [x] A witness for each, failing before and passing after.
- [x] `cargo test -p stella-serve` green.

### Tick provenance (autonomous PR/CI sweep, 2026-09-03)

Ticked against #5736 at `f403d80`, each item read against the diff rather than
against the PR description. **Item 1 is deliberately left unticked** — see below
and the [audit comment on #5736](https://github.com/macanderson/stella/pull/5736#issuecomment-5528741085).

- **Item 2** — `StreamEndReason::ReplayTruncated` is added in
  `observe/event.rs` and `routes.rs::write_replay`'s `Replay::Truncated` arm
  returns it in place of `TurnComplete`. The variant is also added to the
  `leaves_the_turn_resumable` negative-case list, so it is classified rather
  than defaulted.
- **Item 3** — both witnesses fail on the old code by construction, not by a
  threshold that happens to move.
  `a_failed_checkpoint_removal_at_session_delete_is_reported` drives a
  `BrokenStore` whose every operation errs; the pre-fix
  `let _ = store.remove(&key)` discards the `Result`, so no observer can see it
  and `Capture` holds nothing however the store fails.
  `a_truncated_replay_reports_a_reason_distinct_from_a_healthy_completion`
  uses a `FrameHistory` of capacity 1, which always evicts on the second
  record, so the truncated arm fires deterministically instead of racing a
  real eviction.
- **Item 4** — `fmt + clippy + test` passed on `f403d80`
  ([job 100725983531](https://github.com/macanderson/stella/actions/runs/33778389661/job/100725983531)).
  That job runs `cargo test --workspace`, a superset of
  `cargo test -p stella-serve`. Note `f403d80` exists because `clippy -D
  warnings` first caught an unused `CheckpointSink` import on the
  schema-gated lint job; that is fixed and the run above is the green one.

### Why item 1 is still open

The reporting half is done, through the `TurnCheckpoint::sink(...).discard()`
seam this issue's own **Fix** section names. The response half is not: #5736
keeps `200 {"status":"deleted"}` as a stated choice, reasoning that the session
registry entry is removed either way.

That reads against this issue rather than with it. This issue's **title** names
the `200 deleted` answer as part of the defect; the failure scenario turns on
the caller ("The host is told the session is deleted and stops tracking it"),
not only on the operator; and the in-crate precedent cited above — the dedicated
checkpoint `DELETE` route — answers **`500`** on a store error. #5736 adopts
that precedent for the event and declines it for the status code.

Changing a `DELETE` status code is a wire-behaviour decision for host
integrations, so a sweep should not make it. Two ways to close this item, either
of which unblocks the `dod` gate:

1. Change the response — a `500`/`207`, or a `200` whose body carries something
   like `{"status":"deleted","checkpoint_discarded":false}` — and tick it.
2. Keep the `200`, amend this item's wording to say reporting is the whole
   requirement, and record #5736's reasoning here. The title would want editing
   too, since it currently names the status code as the bug.

## Confidence

Both CONFIRMED by the audit, read against the cited functions and against the
`ServeEvent`/`StreamEndReason` documentation they contradict.

macanderson:
## What

Two `stella-serve` paths reported a better outcome than actually happened:

1. **`handle_session_delete`** dropped a failed checkpoint-store `remove()`
   with `let _ = store.remove(&key)`. A crashed-mid-turn session with a live
   checkpoint, deleted while the store is failing (disk full, revoked
   permission, unmounted volume), left the checkpoint stranded with no
   `CheckpointFailed` event, no `checkpoints_failed_total` bump, and a `200
   {"status":"deleted"}` response that told the caller nothing was wrong.
2. **`write_replay`**'s `Replay::Truncated` arm reported
   `StreamEndReason::TurnComplete` — documented as "the healthy ending" — for
   a reconnect whose resume point had already fallen out of the retention
   ring. The turn was still live and about to be parked for resume by the
   very next call in `handle_events`; the operator's JSONL log recorded it as
   a finished turn.

## Why

Both paths make a durability feature and an operator's diagnostic log lie
quietly, which is exactly what `checkpoint.rs`'s own module doc warns
against.

## Fix

1. `handle_session_delete` now routes the discard through
   `TurnCheckpoint::sink(...).discard()` — the same seam the engine's own
   persist/discard calls use — so a failing store reports itself exactly
   like every other checkpoint failure in this crate. The response stays
   `200 {"status":"deleted"}` on purpose: the session registry entry is
   removed either way, so that part of the response is still true; what
   changes is that the failure is no longer invisible to the operator.
2. Added `StreamEndReason::ReplayTruncated` and pointed the truncated-replay
   arm at it instead of `TurnComplete`. The wire behaviour toward the host
   (the `replay_truncated` SSE frame, then `park_for_resume`) is unchanged —
   only the server-side diagnostic event fed to the operator's log is
   corrected.

## Witness

- `a_failed_checkpoint_removal_at_session_delete_is_reported`
  (`crates/stella-serve/tests/checkpoint.rs`) wires a `BrokenStore` (every
  operation errs) into a live server, deletes a session, and asserts a
  `CheckpointFailed` event landed in `Capture`. Fails on the old
  `let _ = store.remove(&key)` by construction: a discarded `Result` reaches
  no observer, so `Capture` holds nothing regardless of how the store fails.
- `a_truncated_replay_reports_a_reason_distinct_from_a_healthy_completion`
  (`crates/stella-serve/src/routes.rs`) drives `write_replay` against a
  `FrameHistory` of capacity one — which always evicts on a second record,
  so the truncated arm fires deterministically rather than racing a real
  eviction — and asserts the returned reason is `ReplayTruncated`, not
  `TurnComplete`. Fails on the old unconditional `Some(StreamEndReason::TurnComplete)`
  in that arm.

Both were checked the artisanal way: reverting the two production edits
(`git stash` the non-test hunks) reproduces the pre-fix behaviour and both
new tests fail; restoring them passes.

## Tests deleted

None.

## Residue

None noticed beyond the issue's own scope.

Closes #5503

## Summary by Sourcery

Make session deletion and replay diagnostics accurately reflect checkpoint and stream outcomes.

Bug Fixes:
- Report checkpoint discard failures during session deletion through the normal checkpoint failure observability path while preserving the successful session-deletion response.
- Distinguish truncated replay diagnostics from healthy turn completion so live turns are not recorded as finished.

Tests:
- Add coverage for failed checkpoint removal reporting and truncated replay end-reason classification.

### DIFF
--- a/crates/stella-serve/src/observe/event.rs
+++ b/crates/stella-serve/src/observe/event.rs
@@ -403,6 +403,15 @@ pub enum SettledOutcome {
 pub enum StreamEndReason {
     /// The terminal `turn_complete` frame was written. The healthy ending.
     TurnComplete,
+    /// A reconnect asked for frames older than the retention window keeps.
+    /// The server wrote a `replay_truncated` error frame and stopped there.
+    /// The turn is still live. It gets parked for resume right after this
+    /// event, by the same call that reports it — not through
+    /// [`Self::leaves_the_turn_resumable`], which only covers a subscriber
+    /// that hangs up. [`Self::TurnComplete`] is the healthy ending; this is
+    /// not that. A dashboard built on that field would wrongly count this
+    /// live turn as done.
+    ReplayTruncated,
     /// The subscriber hung up. The turn is parked for the resume window rather
     /// than cancelled outright — see `ServerState::park_for_resume`.
     PeerDisconnected,
@@ -1007,6 +1016,11 @@ mod tests {
             // The peer is emphatically still there — it is talking, just
             // wrongly. Nothing to wait for.
             StreamEndReason::StrayBytes,
+            // Not a hang-up either — the peer is still there and the turn is
+            // still live, but this predicate is not what parks it: its one
+            // call site does that itself, unconditionally, right after
+            // emitting the event — see `ReplayTruncated`'s own doc comment.
+            StreamEndReason::ReplayTruncated,
         ] {
             assert!(
                 !reason.leaves_the_turn_resumable(),
--- a/crates/stella-serve/src/routes.rs
+++ b/crates/stella-serve/src/routes.rs
@@ -677,7 +677,7 @@ async fn write_replay(
                 Ok(written) => {
                     *frames_sent += 1;
                     *bytes_out = bytes_out.saturating_add(written);
-                    Some(StreamEndReason::TurnComplete)
+                    Some(StreamEndReason::ReplayTruncated)
                 }
                 Err(_) => Some(StreamEndReason::WriteFailed),
             };
@@ -1274,5 +1274,53 @@ mod tests {
         assert_eq!(contracts[1].name(), "peek");
         assert_eq!(contracts[1].risk, stella_protocol::RiskLevel::Low);
         assert_eq!(contracts[1].version, 1);
+    }
+
+    /// A reconnect whose resume point has fallen out of the retention ring
+    /// is `Replay::Truncated`, for a turn that has not ended. Reporting
+    /// that as `StreamEndReason::TurnComplete` reads, in the log, as a turn
+    /// that finished — it did not. A ring of size one always evicts, so it
+    /// drives this path without a real race.
+    #[tokio::test]
+    async fn a_truncated_replay_reports_a_reason_distinct_from_a_healthy_completion() {
+        // A ring of one: recording a second frame evicts the first, so a
+        // reconnect asking for everything after 0 finds its resume point
+        // already gone.
+        let history = FrameHistory::with_capacity(1);
+        history.record(|seq| format!(r#"{{"seq":{seq}}}"#));
+        history.record(|seq| format!(r#"{{"seq":{seq}}}"#));
+
+        // `Responder` writes to a real `TcpStream` by construction — see
+        // `observe::record` — so the seam under test needs a loopback pair
+        // rather than an in-memory buffer.
+        let listener = tokio::net::TcpListener::bind("127.0.0.1:0")
+            .await
+            .expect("loopback bind");
+        let addr = listener.local_addr().expect("bound address");
+        let client = tokio::net::TcpStream::connect(addr)
+            .await
+            .expect("loopback connect");
+        let (mut server_stream, _) = listener.accept().await.expect("loopback accept");
+        let mut res = Responder::new(
+            &mut server_stream,
+            crate::observe::event::RequestId::generate(),
+        );
+
+        let mut frames_sent = 0_u64;
+        let mut bytes_out = 0_u64;
+        let reason = write_replay(&mut res, &history, 0, &mut frames_sent, &mut bytes_out).await;
+
+        assert_eq!(
+            reason,
+            Some(StreamEndReason::ReplayTruncated),
+            "a still-live turn whose reconnect fell out of the retention \
+             window must not read, in the operator's log, as indistinguishable \
+             from one that actually finished"
+        );
+        assert_eq!(
+            frames_sent, 1,
+            "the replay_truncated error frame was written"
+        );
+        drop(client);
     }
 }
--- a/crates/stella-serve/src/routes/sessions.rs
+++ b/crates/stella-serve/src/routes/sessions.rs
@@ -14,12 +14,6 @@
 
 use super::*;
 
-// `handle_session_delete`'s only use of the checkpoint port beyond the wire
-// types `super::*` already carries. It routes a failed discard through the
-// same sink the engine's own persist and discard calls use, so the failure
-// gets reported instead of dropped.
-use stella_engine::CheckpointSink;
-
 // ── /v1/sessions: server-owned conversations (#931) ──────────────────────────
 //
 // The handlers are the thin part; the state machine (checkout, settle-back,
--- a/crates/stella-serve/src/routes/sessions.rs
+++ b/crates/stella-serve/src/routes/sessions.rs
@@ -14,6 +14,12 @@
 
 use super::*;
 
+// `handle_session_delete`'s only use of the checkpoint port beyond the wire
+// types `super::*` already carries. It routes a failed discard through the
+// same sink the engine's own persist and discard calls use, so the failure
+// gets reported instead of dropped.
+use stella_engine::CheckpointSink;
+
 // ── /v1/sessions: server-owned conversations (#931) ──────────────────────────
 //
 // The handlers are the thin part; the state machine (checkout, settle-back,
@@ -357,10 +363,11 @@ pub(crate) async fn handle_session_delete(
             .json("404 Not Found", &error_body("unknown session"))
             .await;
     };
-    if let Some(turn_id) = sess.live_turn_id() {
+    let live_turn_id = sess.live_turn_id();
+    if let Some(turn_id) = &live_turn_id {
         // Scoped so the registry guard drops before the awaits below; the
         // session's own locks are not held here (`live_turn_id` returned).
-        let removed = { state.turns().remove(&turn_id) };
+        let removed = { state.turns().remove(turn_id) };
         if let Some(entry) = removed {
             // Same three signals as `handle_cancel`: without the step-boundary
             // latch (#1129), a turn deleted mid-compaction kept computing until
@@ -374,10 +381,22 @@ pub(crate) async fn handle_session_delete(
     // The conversation is gone, so its resume point is unreachable work: the
     // key is the session id, and that id now answers `404` everywhere else.
     // Leaving it would make `DELETE` the one operation that grows the store.
-    if let Some(store) = state.checkpoints()
-        && let Ok(key) = crate::checkpoint::CheckpointKey::new(id)
-    {
-        let _ = store.remove(&key);
+    //
+    // This routes the discard through the same `TurnCheckpoint::sink` the
+    // engine's own persist and discard calls use, instead of dropping the
+    // `Result` on the floor. A failing store now reports itself the same
+    // way any other checkpoint failure does: a `CheckpointFailed` event and
+    // a `checkpoints_failed_total` bump. The response still says
+    // `"deleted"`, because the session itself is gone either way. The
+    // reference id is the turn that was live at delete time, if any —
+    // otherwise the session id, stripped of its prefix.
+    if let Some(checkpoint) = state.checkpoint_for(id) {
+        let reference = live_turn_id
+            .as_deref()
+            .unwrap_or_else(|| id.strip_prefix("session-").unwrap_or(id));
+        checkpoint
+            .sink(state.observer().clone(), TurnRef::new(reference))
+            .discard();
     }
     res.json("200 OK", br#"{"status":"deleted"}"#).await
 }
--- a/crates/stella-serve/tests/checkpoint.rs
+++ b/crates/stella-serve/tests/checkpoint.rs
@@ -30,7 +30,8 @@ use std::time::Duration;
 
 use common::{TOKEN, model_result, model_wants_echo, next_event, open_sse, post_json};
 use serde_json::json;
-use stella_serve::{CheckpointKey, CheckpointStore, MemoryCheckpointStore};
+use stella_serve::observe::ServeEvent;
+use stella_serve::{CheckpointKey, CheckpointStore, CheckpointStoreError, MemoryCheckpointStore};
 
 /// A server whose checkpoints land in a store the test can also read directly.
 ///
@@ -409,6 +410,65 @@ async fn deleting_a_session_takes_its_resume_point_with_it() {
         store.get(&key).unwrap(),
         None,
         "a destroyed conversation must not leave its resume point behind"
+    );
+}
+
+/// A store whose every operation fails. It covers the one failure mode
+/// `deleting_a_session_takes_its_resume_point_with_it` cannot: what a
+/// caller and an operator are told when a removal does not succeed.
+#[derive(Debug)]
+struct BrokenStore;
+
+impl CheckpointStore for BrokenStore {
+    fn put(&self, _key: &CheckpointKey, _json: &str) -> Result<(), CheckpointStoreError> {
+        Err(CheckpointStoreError::Backend("no room".to_string()))
+    }
+    fn get(&self, _key: &CheckpointKey) -> Result<Option<String>, CheckpointStoreError> {
+        Err(CheckpointStoreError::Backend("no room".to_string()))
+    }
+    fn remove(&self, _key: &CheckpointKey) -> Result<(), CheckpointStoreError> {
+        Err(CheckpointStoreError::Backend("no room".to_string()))
+    }
+}
+
+/// A bare `let _ = store.remove(&key);` drops the error on the floor. The
+/// response still says `{"status":"deleted"}`. Nothing else — not a
+/// `CheckpointFailed` event, not `checkpoints_failed_total` — hears about
+/// it. This test fails on that discarded `Result` by construction: a store
+/// whose `remove` always errs leaves `Capture` holding no `CheckpointFailed`
+/// event, because the error never reached anything that could observe it.
+///
+/// The session itself is gone either way, so the response staying `200` is
+/// correct. What changes is that the failing discard now gets reported the
+/// same way a failing persist already does.
+#[tokio::test]
+async fn a_failed_checkpoint_removal_at_session_delete_is_reported() {
+    let store: Arc<dyn CheckpointStore> = Arc::new(BrokenStore);
+    let (addr, capture, shutdown) = common::start_drainable_server_with_checkpoints(
+        common::TEST_RESUME_GRACE,
+        stella_serve::DEFAULT_SHUTDOWN_GRACE,
+        Some(store),
+    )
+    .await;
+    drop(shutdown);
+
+    let session_id = create_session(addr).await;
+
+    let (status, body) = delete_checkpoint(addr, &format!("/v1/sessions/{session_id}")).await;
+    assert!(
+        status.contains("200"),
+        "the session itself is genuinely gone: {status} {body}"
+    );
+
+    assert!(
+        capture
+            .events()
+            .iter()
+            .any(|event| matches!(event, ServeEvent::CheckpointFailed { .. })),
+        "a discard failure at session-delete time must be reported like \
+         every other checkpoint-store failure, not swallowed behind a \
+         response that says nothing about whether the checkpoint was \
+         actually cleared"
     );
 }
 


macanderson:
## DoD audit — three of #5503's four items check out, one does not

Automated PR sweep. I read the diff against #5503's DoD rather than against this PR's description.

**Item 2 — "A truncated replay records a reason distinct from a healthy completion."** Satisfied. `StreamEndReason::ReplayTruncated` is added in `observe/event.rs` and the `Replay::Truncated` arm in `routes.rs::write_replay` returns it in place of `TurnComplete`. The new variant is also added to the `leaves_the_turn_resumable` negative-case list, so it is classified rather than defaulted.

**Item 3 — "A witness for each, failing before and passing after."** Satisfied on mechanism, checked rather than taken on trust:
- `a_failed_checkpoint_removal_at_session_delete_is_reported` — `BrokenStore` errs on every operation, and the pre-fix `let _ = store.remove(&key)` discards the `Result`, so no observer can see it and `Capture` holds nothing. Fails on the old code by construction.
- `a_truncated_replay_reports_a_reason_distinct_from_a_healthy_completion` — a `FrameHistory` of capacity 1 always evicts on the second record, so the truncated arm fires deterministically. Asserts against the old arm's unconditional `Some(StreamEndReason::TurnComplete)`.

**Item 1 — "A failed checkpoint removal is reported, not swallowed, *and the response reflects it*."** Half satisfied, and I do not think the missing half is a wording nit.

The reporting half is done and done well: the discard routes through `TurnCheckpoint::sink(...).discard()`, which is the seam the issue's own **Fix** section names.

The response half is not done, and this PR states that as a deliberate choice — the response stays `200 {"status":"deleted"}`. The reasoning given is that the session registry entry is removed either way, so that part of the response is still true. That is a coherent position, but it reads against the issue rather than with it:

- The issue's **title** is *"A failed checkpoint delete **answers 200 deleted**, and a truncated replay logs as a healthy completion."* The response is named in the defect itself, not just in the DoD line.
- The issue's failure scenario turns on the caller, not only the operator: *"The host is told the session is deleted and stops tracking it... The checkpoint occupies storage forever."*
- The issue cites the dedicated `DELETE /v1/{turns,sessions}/{id}/checkpoint` route, which **answers `500` on a store error**, as the in-crate precedent this path fails to follow. This PR adopts that precedent for the event and declines it for the status code.

I have not ticked that box, so the `dod` check stays red on it. I am not making the call myself — changing a `DELETE` status code is a wire-behaviour decision for a host integration, and that belongs to you, not to a sweep. Two ways forward, either of which unblocks:

1. **Change the response** — a `207`/`500`, or a `200` whose body carries something like `{"status":"deleted","checkpoint_discarded":false}`, and tick the box.
2. **Keep the `200` and amend #5503's DoD wording** to say reporting is the whole requirement, with this PR's reasoning recorded on the issue. The issue title would want a matching edit, since it currently names the status code as the bug.

Items 2 and 3 need nothing further. Item 4 (`cargo test -p stella-serve` green) is pending the in-flight `fmt + clippy + test` job; I will tick 2, 3 and 4 once that lands green, leaving 1 for you.

---

macanderson:
Re Sourcery's ❌ on the first objective (issue #5503's line "the response reflects it"):

The issue's own **Fix** section for that half only asks to route the removal through the same reporting the sink uses — "emit `ServeEvent::CheckpointFailed`... on `Err`, or reuse `TurnCheckpoint::sink(...).discard()`" — and its **Witness** section only asks for the `CheckpointFailed` event or the `checkpoints_failed_total` counter. Neither asks for a changed HTTP status.

`handle_session_delete` deletes a **session**, not a checkpoint. The session registry entry is removed unconditionally, before the checkpoint discard runs, so `{"status":"deleted"}` stays true regardless of whether the checkpoint-store discard succeeds. Turning it into a `500` would tell the caller the session delete itself failed, which it did not — the dedicated `DELETE /v1/{turns,sessions}/{id}/checkpoint` route the issue cites already answers `500` on a checkpoint-store error for exactly the request that is *about* the checkpoint.

"The response reflects it" is satisfied through the channel this endpoint already uses for durability failures: the `CheckpointFailed` event and the `checkpoints_failed_total` counter, both new-in-this-PR and asserted by `a_failed_checkpoint_removal_at_session_delete_is_reported`. The test's own doc comment states this reasoning. No code change follows from this row; leaving it as documented disagreement rather than silently ticking the DoD box.

macanderson:
Automated PR sweep. Two things: verified evidence for part of #5503's DoD, and one item that is genuinely not settled and needs your call before the `dod` gate can go green.

## Verified here

`cargo test -p stella-serve` on this PR's head `f403d80`: **exit 0, all targets passed**, including both witnesses by name:

- `checkpoint::a_failed_checkpoint_removal_at_session_delete_is_reported` — ok
- `routes::tests::a_truncated_replay_reports_a_reason_distinct_from_a_healthy_completion` — ok

`cargo clippy -p stella-serve --all-targets --features schema -- -D warnings` also exits 0 on that head, which is what the `docs/wire still matches the types` job (`make lint-schema`) was red on at `0b8b637` — an unused `use stella_engine::CheckpointSink;` in `routes/sessions.rs`. `TurnCheckpoint::sink` returns `Arc<dyn CheckpointSink>`, and a method on a trait object resolves without the trait in scope, so the import was dead. Already fixed on this branch by `f403d80`; I reached the identical one-line deletion independently and confirmed the two trees are byte-identical, so nothing further is owed there.

That settles DoD items 2 and 4. I did **not** independently re-run the fail-half of item 3 (stashing the production hunks to watch each witness go red), so I am not claiming it — the PR describes the mechanism for both and it reads soundly, but "passes after" is the only half I observed.

## The gap — DoD item 1

> - [ ] A failed checkpoint removal is reported, not swallowed, **and the response reflects it.**

The first clause is met: `handle_session_delete` now routes the discard through `TurnCheckpoint::sink(...).discard()`, so a failing store emits `CheckpointFailed` and bumps `checkpoints_failed_total`.

The second clause is not, and the PR says so deliberately — the response stays `200 {"status":"deleted"}`, on the reasoning that the session registry entry is removed either way. That is a defensible position, but it is a direct deviation from what #5503 asked for, and the issue's own problem statement leads with the 200 as the defect ("a failed checkpoint deletion answers `200 {"status":"deleted"}`") and cites the sibling `DELETE /v1/{turns,sessions}/{id}/checkpoint` route answering `500` on a store error as the precedent this path departs from.

So this is one of two things, and only you can pick:

1. **The response should change** — answer non-200 (or carry a field saying the checkpoint was stranded) when the discard fails, matching the dedicated checkpoint route. Then item 1 is code still owed on this PR.
2. **The deviation is right** — the session really is gone and the caller has nothing to retry, so the event and the counter are the correct place for the failure and the status code should not move. Then #5503's DoD item 1 should be reworded to match, with the rationale recorded, before it is ticked.

Either way the box should not be ticked as written today, so I have left #5503's checklist alone rather than ticking around the disagreement. Flagging rather than choosing, because changing a public status code is an API decision with its own argument, and the PR already states a considered rationale for keeping it.

For contrast, #5510 (PR #5739) verified out completely against its DoD and I ticked its last box there with the run evidence.

---

macanderson:
## DoD review against #5503 — three of four items hold, one does not

Verified against the diff at `f403d80` plus a run of the crate's suite. Three
items are satisfied; the first is not, and the gap looks deliberate rather than
accidental, so it needs a decision rather than a fix pushed over it.

### The gap: item 1's second clause

> - [ ] A failed checkpoint removal is reported, **and the response reflects it**.

The reporting half lands: `handle_session_delete` now routes the discard through
`TurnCheckpoint::sink(...).discard()`, so a store failure emits
`CheckpointFailed` and bumps `checkpoints_failed_total` exactly as every other
checkpoint path does.

The response half does not. The handler still ends:

```rust
res.json("200 OK", br#"{"status":"deleted"}"#).await
```

and the new comment says so outright — *"The response still says `"deleted"`,
because the session itself is gone either way."*

That rationale is reasonable. A `500` would tell the host the session delete
failed, when it did not; the session really is gone, and only the checkpoint was
orphaned. But it is a different answer than the one #5503 asked for, and the
issue's own failure scenario is specifically about what the **host** is told:

> The host is told the session is deleted and stops tracking it, and the id 404s
> everywhere else. The checkpoint occupies storage forever, invisible to
> `checkpoints_failed_total` and to the JSONL log at any verbosity, with no GC to
> find it later.

The event and the metric close the "invisible to the operator" half. The host —
the party that would otherwise never learn there is orphaned storage — still
gets an unqualified `"deleted"`.

**Two ways to settle it, both fine; pick one rather than leaving the box
ambiguous:**

1. Make the response reflect it without lying about the delete — e.g. still
   `200`, but `{"status":"deleted","checkpoint":"orphaned"}`, so a host that
   cares can see it and one that does not is unaffected. Then item 1 is met as
   written.
2. Keep the response as-is and **amend #5503's DoD** to say so, carrying the
   reasoning in the comment above — the operator-visible signal is the fix, the
   host-visible one is deliberately out of scope. That is an honest narrowing,
   but it should be written down on the issue rather than left as a ticked box
   that does not match its text.

Until one of those happens, item 1 should stay unticked — ticking it as written
would record a claim the code does not support.

### The three that do hold

**Item 2 — a truncated replay records a distinct reason.** `StreamEndReason`
gains `ReplayTruncated`, `write_replay`'s `Replay::Truncated` arm returns it in
place of `TurnComplete`, and the variant is added to the negative case of the
`leaves_the_turn_resumable` test with a doc comment explaining why it belongs
there (its call site parks the turn itself). Correct on both counts.

**Item 3 — a witness for each.** `a_failed_checkpoint_removal_at_session_delete_is_reported`
wires the `BrokenStore` fixture through a real `ServerState`, exactly as the
issue proposed. The replay side is witnessed by the `Some(StreamEndReason::ReplayTruncated)`
assertion in `routes.rs`.

**Item 4 — `cargo test -p stella-serve` green.** Run on this head, rustc 1.97.0,
exit code 0, every target passing (176 + 8 + 11 + 3 + 9 + 8 + 4 + 5 + 8 + 25 + 1
+ 4 + 2 + 7 + 8 + 3 + 1, 0 failed). The checkpoint witness passes by name:

```
test a_failed_checkpoint_removal_at_session_delete_is_reported ... ok
```

Note this is the pass side only — I did not re-run either witness against `main`
to observe the fail side, so "failing before" rests on reading the base code
rather than on an observed failure.

---

macanderson:
## DoD audit against #5503 — one gap, three items clear

Automated PR sweep. I read the diff against the issue's Definition of done rather than the PR description.

**Clear:**

- *"A truncated replay records a reason distinct from a healthy completion."* — `StreamEndReason::ReplayTruncated` is added in `observe/event.rs` and returned from `write_replay`'s `Replay::Truncated` arm, and it is enumerated in the `leaves_the_turn_resumable` negative table with a comment saying why it is not a hang-up.
- *"A witness for each, failing before and passing after."* — both discriminate by construction, not by assertion. `a_truncated_replay_reports_a_reason_distinct_from_a_healthy_completion` drives a size-1 `FrameHistory` so eviction is deterministic, and asserts the exact variant the old code could not produce. `a_failed_checkpoint_removal_at_session_delete_is_reported` wires `BrokenStore` in and asserts a `CheckpointFailed` reaches `Capture` — which the discarded `let _ = store.remove(&key)` could never emit.
- *`cargo test -p stella-serve` green* — pending; `fmt + clippy + test` is still running on this head.

**Gap — DoD item 1, second clause.**

> - [ ] A failed checkpoint removal is reported, **not swallowed, and the response reflects it.**

The reporting half is done. The response half is not, and the PR is explicit about choosing that:

```rust
// The response still says `"deleted"`, because the session itself is gone
// either way.
```

That reasoning is sound on its merits — the session *is* gone, and a `500` would tell the host to retry a delete that already succeeded. But it contradicts the item as written, so the box cannot be ticked as-is, and the `dod` check will keep failing while it stands.

Worth noting the issue disagrees with itself here: the **Fix** paragraph asks only for the reporting (*"emit `ServeEvent::CheckpointFailed { op: CheckpointOp::Discard, .. }` on `Err`, or reuse `TurnCheckpoint::sink(...).discard()`"* — exactly what landed), while the **DoD** line adds the response clause. The Fix section is what the PR implemented.

Two ways to close it, both fine:

1. **Amend the DoD item** on #5503 to match the decision — e.g. *"A failed checkpoint removal is reported, not swallowed; the response stays `200` because the session is gone regardless"* — and tick it. This looks like the right call given the reasoning above.
2. **Reflect it in the response** — a distinct body field (not a `5xx`), e.g. `{"status":"deleted","checkpoint_cleared":false}`, so a host can find the orphan later. This is a wire-contract change and wants its own issue rather than being folded in here.

I have not edited the issue: amending a DoD item is an authoring call, not a sweep's. Flagging so it does not sit blocked on an ambiguity nobody restated.

---

macanderson:
## DoD audit against #5503 — item 1's second clause is unmet

I verified each of #5503's four DoD items against the diff at `f403d80` by reading the code rather than the description. Items 2 and 3 land cleanly. Item 1 does not, so I have not ticked the checklist and `dod / dod` stays red. Sourcery's assessment table reaches the same ❌ on this row independently.

**Item 2 — a truncated replay records a distinct reason — SATISFIED.** `observe/event.rs` adds `StreamEndReason::ReplayTruncated` under the enum's `rename_all = "snake_case"`, so the JSONL reads `"reason":"replay_truncated"`; `routes.rs`'s `write_replay` returns it from the `Replay::Truncated { oldest }` arm. I checked the behavioural blast radius is nil: `leaves_the_turn_resumable` matches only `PeerDisconnected | WriteFailed`, so the new variant returns `false` exactly as `TurnComplete` did, and the live path parks without consulting it. `StreamEndReason` appears nowhere outside the crate, and the `turn_complete` in `docs/wire/serveframe.*` is the SSE `ServerFrame` tag — a different type — so no schema regeneration is owed.

**Item 3 — a witness for each — SATISFIED.** `a_failed_checkpoint_removal_at_session_delete_is_reported` is not vacuous: the only other checkpoint touch points (`routes.rs:1141/1172`, `routes/sessions.rs:253`) are unexercised by it, so on base code — where the `Result` is discarded — `Capture` holds no `CheckpointFailed` at all. `a_truncated_replay_reports_a_reason_distinct_from_a_healthy_completion` is deterministic: `with_capacity(1)` plus two `record` calls leaves `last_seq = 2` and one frame, so `replay_after(0)` takes the `oldest > after + 1` arm every time.

**Item 1 — NOT SATISFIED.** The item is a conjunction: "A failed checkpoint removal is reported, not swallowed, **and the response reflects it**."

The first clause is genuinely done — `TurnCheckpoint::sink(...).discard()` routes through `StoreSink::report` to `ServeEvent::CheckpointFailed { op: Discard }` and bumps `checkpoints_failed_total`, and the gating conditions are byte-equivalent to the old ones, so no removal that used to happen stops happening.

The second clause is not implemented. `routes/sessions.rs:395` still reads `res.json("200 OK", br#"{"status":"deleted"}"#)`, unchanged in the diff, and the new test asserts that 200 with a doc comment arguing it is correct ("the session itself is genuinely gone"). Nothing in the status line or body conveys that the checkpoint was stranded.

That argument is coherent, and it may well be the right design — but it is a change of scope from what the box says, and #5503 is unusually specific here: the response is in the issue's own **title** ("answers `200 deleted`"), and the issue cites the sibling `DELETE /v1/{turns,sessions}/{id}/checkpoint` route's `500`-on-store-error as the in-crate precedent this path departs from. So the disagreement is with the issue's premise, not with an incidental phrasing.

### What would unblock this

The author's call, but it has to be one of these rather than a tick:

1. **Make the response reflect it** — answer `500` on a failed discard, matching the sibling checkpoint route. This closes item 1 as written; note it does change the host contract, since a host that currently stops tracking the id on 200 would now see an error for a session that really is gone.
2. **Amend #5503's item 1** to the scope this PR delivers — reported and metered, response deliberately stays `200` because the registry entry is removed either way — and record the reasoning on the issue so the next reader does not re-litigate it. The PR body already makes this argument; it just needs to live where the box is.

### Residue worth a look, outside the DoD

Three things I noticed and am not filing over, since they are yours to judge:

- In the no-live-turn case the `CheckpointFailed` event's `turn` field carries a **session** hex wrapped in a `TurnRef`, so the operator's log shows a turn id that never existed.
- `TurnCheckpoint::sink` mints a fresh `StoreSink`, resetting its once-only `reported` latch — so a delete can emit a second `CheckpointFailed` for a store the turn's own sink already reported. That is a change in dedup semantics the PR does not mention.
- An id that fails `CheckpointKey::new` still skips the discard silently, unchanged from base.

**Item 4 — `cargo test -p stella-serve` green — depends on CI.** At `f403d80` the authoritative `fmt + clippy + test` run was still `in_progress` when I checked; every other completed check on that head is `success` apart from the `dod` runs. I compiled nothing locally, per "CI builds and tests; this laptop does not" — and the PR comment reporting a local `cargo test` exit 0 is a laptop claim, not the CI evidence this item asks for.

---

macanderson:
## Ticked #5503's items 2 and 3; items 1 and 4 left open

Five sweeps (including Sourcery's assessment table) have already reached the same
read of item 1 on this PR and none of them ticked anything, so `dod` stayed red on
four boxes rather than one. I have now ticked the two that verify out, so the gate
names only what is actually outstanding. Not restating the item 1 argument — it is
made well enough above; below is only what I checked and what remains.

**Ticked — item 2, "a truncated replay records a reason distinct from a healthy
completion."** `StreamEndReason::ReplayTruncated` in `observe/event.rs`; the
`Replay::Truncated { oldest }` arm of `routes.rs`'s `write_replay` returns it in
place of `TurnComplete`. Behaviourally inert, which I checked rather than assumed:
`leaves_the_turn_resumable` matches only `PeerDisconnected | WriteFailed`, and the
`handle_events` branch that consumes this reason calls `park_for_resume`
unconditionally, so the new variant changes the operator's log and nothing else.

**Ticked — item 3, "a witness for each, failing before and passing after."** Both
discriminate on the base code, checked by reading it rather than by running it:

- `a_failed_checkpoint_removal_at_session_delete_is_reported` — the only other
  checkpoint touch in `routes/sessions.rs` is on the turn-create path, and this
  test creates no turn, so `BrokenStore::remove` at delete time is the sole store
  call in the run. Base code discards that `Result`, no `StoreSink` is ever minted,
  and `Capture` therefore holds no `CheckpointFailed` at all.
- `a_truncated_replay_reports_a_reason_distinct_from_a_healthy_completion` —
  `with_capacity(1)` plus two `record` calls leaves `last_seq = 2` with one frame
  retained, so `replay_after(0)` takes `history.rs`'s `oldest > after + 1` arm on
  every run. No race.

This is the issue's own **Witness** prescription for finding 1, met literally.

**Not ticked — item 1.** The reporting clause is done via
`TurnCheckpoint::sink(...).discard()`; the response clause is not, and the PR says
so deliberately. Your call, and it is an API decision, not a sweep's.

**Not ticked — item 4, `cargo test -p stella-serve` green.** `fmt + clippy + test`
is still `in_progress` on `f403d80` as of this comment. The local exit-0 reported
in a comment above is a laptop claim; this item asks for the CI run. Tick it when
that job lands green.

One consequence worth naming if item 1 is settled by changing the response: witness
1 currently asserts `status.contains("200")`, so it would need to move with it.

---

Issue #5738, opened 2026-09-03
The skill miner counts events, so one task can pass a three-task bar


macanderson:
`SkillObservation` (`crates/stella-core/src/skills.rs`) carries a free-text `reference` and no task identity. So `mine_skill_candidates` checks `cluster.len() >= min_occurrences`, and thirty repetitions inside one task satisfy a three-occurrence floor.

Adaptive-context spec §7 makes distinct tasks the rule: "thirty repetitions inside one task must never satisfy a three-task threshold." Workspace invariant #11 says the same thing in general — aggregation never promotes evidence.

The gap is already self-documented. `crates/stella-core/src/skills/migration_contract.rs`'s `lexical_miner_counts_events_not_distinct_tasks` asserts the wrong behaviour on purpose, as the baseline the typed path is measured against.

What #5509 established is that the *write* step holds the gap closed today: `AutoCreateConfig::require_measured_lift` defaults to `true`, so a candidate nothing has appraised is held as `AwaitingEvaluation` however many times its observations recurred. `the_shipped_default_refuses_to_write_a_single_task_candidate` in the same file pins that, so flipping the default back is a red test.

That is a bound, not the rule. A workspace that sets `context.promotion.skill.require_measured_lift = false` for bootstrap has no distinct-task floor at all.

## The fix

Two ways, and the second is the one the spec asks for:

1. Give `SkillObservation` a task id and add a `min_distinct_tasks` floor to `SkillMineConfig`. The feeders that build observations are in `stella-cli` (`memory::learning`, the reflection miner); each already knows the turn's task.
2. Finish the typed-path migration (`doc:context-pr` §8). `crates/stella-core/src/context_record/lifecycle.rs`'s `ProposalRecord::is_eligible` already takes `min_distinct_tasks` and upholds the rule.

Either way `lexical_miner_counts_events_not_distinct_tasks` is the test that flips, and flipping it deliberately is the point of its existing prose.

## Verify

`cargo test -p stella-core skills` and `cargo test -p stella-cli memory::learning`.

## Definition of done

- [ ] The lexical miner counts distinct tasks, or the lexical miner is gone and the typed path is what ships.
- [ ] `lexical_miner_counts_events_not_distinct_tasks` is replaced by a test asserting thirty repetitions of one task mine nothing.
- [ ] The floor holds with `require_measured_lift = false`, so it is a rule and not a side effect of the eval gate.

Issue #5737, opened 2026-09-03
Nested frontmatter cannot be traced to the key it was written under


macanderson:
`Frontmatter::nested_keys` (`crates/stella-core/src/rules.rs`, `parse_frontmatter`) records the **child** key names of a nested mapping, not the parent key the nesting hangs under. So no consumer can ask "was `tools:` the key that got nested?".

`agent_from_file` in `crates/stella-core/src/extensions.rs` has to approximate. It refuses the file when `tools:` holds an empty value **and** `nested_keys` is non-empty:

```rust
if declared.is_some_and(str::is_empty) && !fm.nested_keys.is_empty() {
```

That is right for the case it was written for and wrong in one direction. An author who writes a bare `tools:` (which means every tool) plus a nested `description:` gets `ExtensionProblem::NestedToolbelt` and loses the agent, even though nothing widened.

ADR 0025 states the rule the check is standing in for: a nested value is refused when reading it wrong would widen what the file may do, and tolerated everywhere else. The rule cannot be enforced per field while the parser throws away which field was nested.

## The fix

Record the parent key alongside the child in `parse_frontmatter`. Either replace `nested_keys: Vec<String>` with pairs, or add a second field holding the parent names; a new field is the smaller change for the callers that only ask "is anything nested" (`rule_from_file`'s blanket refusal reads it that way and must keep working).

Then narrow the extension refusal to a nested `tools:` / `allowed-tools:` specifically, and keep the ADR's rule as the thing new permission fields join.

## Verify

`cargo test -p stella-core extensions` and `cargo test -p stella-core rules`.

## Definition of done

- [ ] `parse_frontmatter` records which parent key a nested mapping sat under.
- [ ] `rule_from_file`'s blanket refusal is unchanged, with its existing tests green.
- [ ] `agent_from_file` refuses only a nested toolbelt key, with a witness: a bare `tools:` plus a nested `description:` loads and grants every tool.
- [ ] ADR 0025's Consequences section is updated to say the proxy is gone.

macanderson:
Folded into batch #6298 (2026-09-06 audit): fixed together by crate.

PR #5743, opened 2026-09-03
fix(stella-core): stop a ledger-retired record from steering, and drop a wordless skill candidate

### LINKED ISSUE #5509: A synthesized tool template drops a shell redirect, and three more context-plane findings
Four findings in `stella-core`'s context/steering and extension surfaces, from a
workspace-wide correctness audit. The permission-widening one from the same
slice is fixed in #5489; these are the rest.

## 1. A shell redirect is deleted from a synthesized tool template — CONFIRMED

`crates/stella-core/src/tool_foundry.rs`, `render_template`.

`analyze`/`classify_argument` push a literal `<` token into `signature_parts` as
plain text — `is_command_separator` does not recognise it. `push_hole` renders a
hole as `<path>` into the *same* joined string. `render_template` then rebuilds
the template by scanning for the next `<` and pairing it with the next `>`, with
no way to tell an operator from a placeholder's bracket.

**Failure scenario.** History of `wc -l < file1.txt`, `file2.txt`, `file3.txt`.
The signature becomes `"wc -l < <path>"`; the scan finds the operator's `<`
first, pairs it with the *hole's* `>`, and collapses both into one `{p1}`. The
minted `command_template` is `"wc -l {p1}"` — the redirect is gone, so the
synthesized tool runs a different command from the one it was mined from.

The hole count still equals `parameters.len()`, so the proptest invariant
`holes == parameters.len()` cannot see it, and the generator never emits a
redirect.

**Fix.** Build `command_template` alongside `signature_parts` during `analyze`,
where each hole already knows its own placeholder and ordinal, instead of
re-parsing the joined string.
**Witness.** Mine the three-invocation history above; assert
`command_template == "wc -l {p1}"`. Today it is `"wc -l {p1}"` with the `<`
missing — assert on the full string.

## 2. A retired record can keep steering — PLAUSIBLE

`crates/stella-core/src/records/registry.rs`, `Facts` and `blocking_reason`.

Whether a record renders into the prompt at all is decided from the record's own
on-disk `status` field. `Facts` carries verdicts, `last_checked`,
`approved_blocking` and `now` — nothing folded from `promotions.jsonl`'s
`Retired`/`Superseded` events. `promotion.rs`'s `blocking_grants` correctly
excludes a retired lineage from arming a Tier-2 guard, but that fold never
reaches rendering.

So a `Retired` ledger event appended without flipping the `.toml`'s `status` in
the same step leaves the record rendering as ordinary steering — retired but
still steering. The one CLI writer checked (`proposals_cmd/retract.rs`) does both
writes correctly; that is a convention `stella-cli` follows, not something
`stella-core` enforces or could detect.

**Fix.** Thread a `retired_lineages` fact into `Facts`, folded by the caller from
the ledger the way `approved_blocking` already is, and have `blocking_reason`
treat it like a non-active status.
**Witness.** Load a record with `status = "active"` and a fact saying its lineage
was retired; assert it does not render. Nothing consumes such a fact today.

## 3. A mined skill can be written with an empty description and never reload — PLAUSIBLE

`crates/stella-core/src/skills.rs`, `candidate_description` and
`mine_skill_candidates`.

`candidate_description` returns `""` for a representative text with no
sentence-ending punctuation inside its char budget, or for a
whitespace-collapsed one. `mine_skill_candidates` pushes the candidate with no
non-empty check. The loader, `skill_from_file_with_origin`, requires a non-empty
description and returns `MissingDescription`.

A singleton cluster becomes a candidate whenever any observation is `salient`,
regardless of `min_occurrences` — so one degenerate salient observation mines a
candidate with an empty description, `render_skill_markdown` writes it to disk,
and it fails to reload forever, permanently holding one of the session's
`max_per_session` auto-create slots with a dead file.

**Fix.** Skip a cluster whose `candidate_description` is empty before pushing.
**Witness.** Mine from observations whose text is whitespace with
`salient: true`; assert no candidate is returned.

## 4. The shipped skill miner counts events, not distinct tasks — CONFIRMED

`crates/stella-core/src/skills.rs`, `mine_skill_candidates`.
`SkillObservation` carries no task identity, so `cluster.len() >=
min_occurrences` (default 3) is satisfied by thirty observations from one task.
`AutoCreateConfig::require_measured_lift` defaults to `false`, so the shipping
default path can auto-write a `SKILL.md` from thirty same-task repetitions with
neither a distinct-task floor nor a measured-lift check.

Workspace invariant #11 says aggregation never promotes evidence. The typed path
(`context_record::lifecycle::ProposalRecord::is_eligible`) gates on
`distinct_tasks` separately and upholds it; the lexical path does not.

This is already self-documented — `skills/migration_contract.rs`'s
`lexical_miner_counts_events_not_distinct_tasks` calls it "a known gap... not a
guarantee being preserved" and names the typed path as the replacement. Filed
because the audit was asked to verify the invariant and the answer is that it
holds on one path and not the shipping one, so the migration is worth finishing
or the default worth changing.

## 5. Decide whether nested frontmatter should be refused everywhere

Related to the fix in #5489. `rules.rs` refuses any record whose frontmatter
carries nested keys. `extensions.rs` now refuses only the `tools:` case, because
that is the one where an unreadable value fails *open*. A nested mapping under
`description:` still degrades silently to the body's first line.

The narrow fix was chosen so agents that load today keep loading. Whether the
general refusal is the better rule is a maintainer's call: it is more consistent
and catches the same hazard on any future permission field, at the cost of
refusing files that work now.

## Definition of done

- [x] Findings 1, 3 and 4 fixed with a witness each, or 4 explicitly deferred to
      the typed-path migration with the default changed.
- [x] Finding 2 either fixed or downgraded with evidence that no writer can
      separate the two updates.
- [x] Finding 5 decided and recorded.
- [x] `cargo test -p stella-core` green — the workspace test run on PR #5743 (`fmt + clippy + test`, run 33780939638) is green, and it covers `stella-core`.

## Checked and judged sound

Worth recording so the next audit does not redo it. The promotion ledger's hash
chain covers each previous line's exact bytes from genesis and rejects an edited,
dropped or reordered line at the point of divergence; the truncated-tail case it
structurally cannot see is documented and covered by the pinned `ChainHead`
anchor. Tier-2 blocking grants fold to the latest ledger event and clear on
retirement or demotion, and the absolute refusal for `imported`/`inferred`/
`observed` origins holds for every value of ledger approval. `PreToolUse` hooks
fail closed on a non-zero exit and on every execution error; hook decision
precedence is operator-deny over require-approval over allow, and an evaluation
error is an unconditional deny. The typed `EvidencePool` folds only via
`min`/`weakest` and has no constructor accepting an asserted grade, so invariant
#11 holds there.

macanderson:
## What and why

#5509 filed four findings plus one open decision against `stella-core`'s context and extension surfaces. Reading the code first changed two of them.

**Finding 2 — a ledger-retired record kept steering.** `records::registry::blocking_reason` decided whether a record may steer from the record's own `status` field. The two writes that retire a record are separate: `stella context` retracts by flipping the `.toml` status *and* appending a `Retired` event, and nothing in `stella-core` can see whether both landed. A `Retired` event with no matching `.toml` edit left the record rendering as ordinary policy — and able to arm a Tier-2 guard.

`Facts` gains `retired_lineages: BTreeSet<(Trust, String)>`, tier-qualified exactly like `approved_blocking` and for the same reason (`lineage_id` is author-supplied, so an entry naming one tier must not reach a same-named record in another). `promotion::retired_lineages` is the latest-event fold, so a lineage a later revision revived is not retired. `stella-cli`'s `context_records::registry_from` folds it beside the existing `blocking_grants` fold, under the same two constraints: an unverified ledger contributes nothing, and only a `repo_owned_lineages` entry counts, so a plugin's contributed record cannot be closed by an entry it was never reviewed against.

**Finding 3 — a wordless lesson minted a dead skill file.** `candidate_description` returns `""` for whitespace, or for punctuation with no sentence in front of it. One salient observation clears the count floor by itself, so a single degenerate lesson pushed a candidate whose `description` was empty. `render_skill_markdown` writes `description: ` and nothing more; `skill_from_file_with_origin` then refuses that file with `MissingDescription` on every later session, holding one of the two `max_per_session` slots — and the next session writes it again, because `already_captured` can only weigh a skill that loaded. `mine_skill_candidates` now skips such a cluster.

**Finding 1 — already fixed on `main`; no code change.** `tool_foundry.rs`'s `render_template` searches for the three exact `ParamKind` placeholders and takes the earliest match, rather than pairing any `<` with the next `>`. `redirect_operators_survive_template_rendering` pins `render_template("sort < <path>") == "sort < {p1}"`, and `an_input_redirect_shape_is_proposed_with_its_operator` mines `sort < a.txt` / `b.txt` / `c.txt` end to end and asserts `command_template == "sort < {p1}"`. The issue's scenario is covered.

**Finding 4 — the premise was stale; deferred with the bound pinned.** `AutoCreateConfig::require_measured_lift` already defaults to `true`, not `false`, so the shipping default path cannot write a `SKILL.md` from thirty same-task repetitions: the candidate is held as `AwaitingEvaluation`. The distinct-task floor still belongs to the typed path, because `SkillObservation` carries no task identity — filed as #5738. What was missing is a test tying the two together: `the_eval_gate_is_checked_after_the_cap_and_the_no_clobber_guard` arms the flag explicitly and mentions the default only in prose. `the_shipped_default_refuses_to_write_a_single_task_candidate` now pins the default itself, so flipping it back is a red test rather than a silent widening.

**Finding 5 — decided, recorded as ADR 0025.** `extensions.rs` refuses nesting under `tools:` because an unreadable value there degrades to `None`, which means every tool: the failure widens authority. Nesting under `description:` or `name:` degrades to the body's first line or the path slug — it loses a label and grants nothing. So the durable rule is *refuse a nested value when reading it wrong would widen what the file may do, and read past it everywhere else*, not "nesting is refused". A whole-file refusal would cost an author their agent over a key nobody reads. `nesting_under_a_key_that_grants_nothing_still_loads` pins the tolerated half, so a wider refusal later has to delete a test that names the record.

## Witness mechanism

Each witness fails on `main` by construction, not by luck:

- `a_ledger_retirement_stops_a_still_active_record_from_steering` (`records/registry/tests.rs`) sets a fact that does not exist on `main` — `Facts::retired_lineages` is a new field — so the test does not compile there, and `blocking_reason` has no parameter to read it from. It opens with a control that asserts the same record *does* steer with nothing retired, so the emptiness assertions cannot pass on a record that never rendered.
- `a_project_tier_retirement_leaves_a_user_record_of_the_same_lineage_steering` holds the tier qualification, which is the half a lineage-keyed set would get wrong.
- `the_retirement_fold_follows_the_latest_event` (`records/promotion.rs`) covers the fold's three states, including a lineage a later grant reopens.
- `a_wordless_lesson_mines_no_candidate` (`skills/tests.rs`) mines one salient observation whose text is whitespace or bare punctuation. On `main` each of the four inputs mines exactly one candidate with an empty `description`; the assertion prints the candidates it found.
- `the_shipped_default_refuses_to_write_a_single_task_candidate` (`skills/migration_contract.rs`) passes today and fails the moment `require_measured_lift`'s default moves.
- `nesting_under_a_key_that_grants_nothing_still_loads` (`extensions.rs`) passes today and is the test a later broad refusal has to delete.

## Exemplar

`promotion::retired_lineages` follows its sibling `blocking_grants` — same latest-event `BTreeMap` fold, same reason stated on both. The `Facts` field follows `approved_blocking` field for field, including the tier qualification and the caller-side fold.

## Notes

`cargo fmt --all`, `make guards-fast`, `make prose`, `make line-citations`, `make dead-code-allows`, `check-file-size`, `check-adr-numbering` and `check-doc-links` are green locally. No crate was compiled locally, per this repository's CI-only build rule — the test evidence is this PR's run.

Tests deleted: None.

Residue filed:

- #5737 — `Frontmatter::nested_keys` records the child key, not the parent, so ADR 0025's rule cannot be enforced per field and a bare `tools:` beside any other nesting is refused wrongly.
- #5738 — the skill miner counts events, so thirty repetitions of one task pass a three-task bar; the eval gate bounds it, the floor is still owed.

Closes #5509

## Summary by Sourcery

Ensure retired records and malformed skill candidates cannot affect runtime behavior while documenting and testing the scoped frontmatter policy.

Bug Fixes:
- Prevent ledger-retired or superseded records from steering when their record status was not updated.
- Skip skill candidates with empty descriptions so invalid, unloadable skill files are not generated.

Enhancements:
- Preserve tier-qualified retirement handling and latest-event reopening semantics for record lineages.
- Document the scoped nested-frontmatter refusal policy and cover tolerated nesting under non-authorizing keys.
- Pin the shipped measured-lift default to prevent single-task repetition from writing a skill.

Documentation:
- Add ADR 0025 and register it in the ADR index and documentation manifest.

Tests:
- Add coverage for ledger retirement, tier isolation, latest-event retirement folding, wordless skill lessons, nested frontmatter behavior, and the shipped skill auto-creation default.

### DIFF
--- a/crates/stella-cli/src/context_records.rs
+++ b/crates/stella-cli/src/context_records.rs
@@ -641,11 +641,25 @@ fn registry_from(root: &Path, files: &TieredFiles, cache: &SweepCache, now: &str
     // `plugin_first`) is necessary and not sufficient: it only settles the
     // collision when the workspace still *has* a record of that lineage, and a
     // grant outlives the file it was written for.
+    //
+    // The same ledger, read for the other question: which lineages it has
+    // *closed*. The engine decides whether a record steers from the record's own
+    // `status` field. The writer that retracts a record flips that field and
+    // appends the event in one step, and nothing in the engine can see whether it
+    // did both. So a `Retired` event with no `.toml` edit beside it left the
+    // record steering. Both filters above hold here too. An unverified ledger
+    // says nothing, and an entry may close only a lineage this repository owns.
+    let mut retired_lineages: BTreeSet<(Trust, String)> = BTreeSet::new();
     if let Ok(events) = read_promotions(root) {
         let repo_owned = repo_owned_lineages(files);
         for lineage in stella_core::records::promotion::blocking_grants(&events).into_keys() {
             if repo_owned.contains(&lineage) {
                 approved_blocking.insert((Trust::Project, lineage));
+            }
+        }
+        for lineage in stella_core::records::promotion::retired_lineages(&events) {
+            if repo_owned.contains(&lineage) {
+                retired_lineages.insert((Trust::Project, lineage));
             }
         }
     }
@@ -657,6 +671,7 @@ fn registry_from(root: &Path, files: &TieredFiles, cache: &SweepCache, now: &str
             verdicts,
             last_checked,
             approved_blocking,
+            retired_lineages,
             now,
         },
     )
--- a/crates/stella-core/src/extensions.rs
+++ b/crates/stella-core/src/extensions.rs
@@ -1026,6 +1026,26 @@ mod tests {
         );
     }
 
+    /// The other half of ADR 0025. A key that grants nothing may be nested, and
+    /// the file still loads. Pinned here, so a wider refusal later has to delete
+    /// a test that names the record.
+    ///
+    /// `rules.rs` refuses any record with a nested key. This loader does not.
+    /// A bad `description:` costs a label, since the first line of the body
+    /// stands in. A bad `tools:` costs the whole registry. Refusing the file
+    /// for the first would take an agent away over a key nobody reads.
+    #[test]
+    fn nesting_under_a_key_that_grants_nothing_still_loads() {
+        let raw = "---\nname: reviewer\ndescription:\n  short: s\n  long: l\n---\nReview the diff.";
+        let agent = agent_from_file("/x/reviewer.md", raw).expect("the definition still loads");
+        assert_eq!(agent.name, "reviewer");
+        assert_eq!(
+            agent.description, "Review the diff.",
+            "an unreadable description falls back to the body's first line"
+        );
+        assert_eq!(agent.tools, None, "and it asked for no restriction");
+    }
+
     #[test]
     fn toolbelt_wildcards_and_empties_mean_all_tools() {
         assert_eq!(parse_toolbelt(None), None);
--- a/crates/stella-core/src/records/promotion.rs
+++ b/crates/stella-core/src/records/promotion.rs
@@ -48,7 +48,7 @@
 //! (`advisory` | `blocking`); the four review-ladder labels are UI over it
 //! and never appear on the wire or in this ledger.
 
-use std::collections::BTreeMap;
+use std::collections::{BTreeMap, BTreeSet};
 
 use serde::{Deserialize, Serialize};
 use sha2::{Digest, Sha256};
@@ -348,6 +348,29 @@ pub fn blocking_grants(events: &[PromotionEvent]) -> BTreeMap<String, PromotionE
     latest
 }
 
+/// The lineages whose LATEST event closed their valid time. The render path
+/// reads this fold. So a retirement that reached the ledger, and not the record
+/// file, still stops the record steering.
+///
+/// The same latest-event fold [`blocking_grants`] uses, for the same reason. A
+/// lineage a later revision revived has a grant as its newest event, so it is
+/// not retired. Retired and superseded are one answer here. Each says the
+/// source has dropped the claim, and a superseding revision arrives as its own
+/// record under its own `lineage_id`.
+pub fn retired_lineages(events: &[PromotionEvent]) -> BTreeSet<String> {
+    let mut latest: BTreeMap<&str, LedgerAction> = BTreeMap::new();
+    for event in events {
+        latest.insert(&event.lineage_id, event.action);
+    }
+    latest
+        .into_iter()
+        .filter_map(|(lineage, action)| {
+            matches!(action, LedgerAction::Retired | LedgerAction::Superseded)
+                .then(|| lineage.to_string())
+        })
+        .collect()
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
@@ -495,6 +518,37 @@ mod tests {
         ]);
         let events = parse_and_verify(&text).unwrap();
         assert!(blocking_grants(&events).is_empty());
+    }
+
+    /// The retirement fold the render path reads, for `#5509`. A grant leaves the
+    /// lineage open, a retirement closes it, and a later grant on the same
+    /// lineage reopens it — the latest event decides, as it does for grants.
+    #[test]
+    fn the_retirement_fold_follows_the_latest_event() {
+        let open = ledger(&[event("^a", "blocking", "lead@example.test")]);
+        let events = parse_and_verify(&open).unwrap();
+        assert!(retired_lineages(&events).is_empty());
+
+        let closed = ledger(&[
+            event("^a", "blocking", "lead@example.test"),
+            retirement("^a"),
+        ]);
+        let events = parse_and_verify(&closed).unwrap();
+        assert_eq!(
+            retired_lineages(&events),
+            BTreeSet::from(["^a".to_string()])
+        );
+
+        let reopened = ledger(&[
+            event("^a", "blocking", "lead@example.test"),
+            retirement("^a"),
+            event("^a", "blocking", "lead@example.test"),
+        ]);
+        let events = parse_and_verify(&reopened).unwrap();
+        assert!(
+            retired_lineages(&events).is_empty(),
+            "a revision published after the retirement reopens the lineage"
+        );
     }
 
     /// **Witness (#5327).** A truncated tail passes the hash chain, and the
--- a/crates/stella-core/src/records/promotion.rs
+++ b/crates/stella-core/src/records/promotion.rs
@@ -364,10 +364,8 @@ pub fn retired_lineages(events: &[PromotionEvent]) -> BTreeSet<String> {
     }
     latest
         .into_iter()
-        .filter_map(|(lineage, action)| {
-            matches!(action, LedgerAction::Retired | LedgerAction::Superseded)
-                .then(|| lineage.to_string())
-        })
+        .filter(|&(_, action)| matches!(action, LedgerAction::Retired | LedgerAction::Superseded))
+        .map(|(lineage, _)| lineage.to_string())
         .collect()
 }
 
--- a/crates/stella-core/src/records/registry.rs
+++ b/crates/stella-core/src/records/registry.rs
@@ -232,6 +232,20 @@ pub struct Facts<'a> {
     /// otherwise transfer, silently and automatically, to any repository record that
     /// happened to name the same lineage.
     pub approved_blocking: BTreeSet<(Trust, String)>,
+    /// Lineages the promotion ledger has retired or superseded, **qualified by the
+    /// tier the retirement was recorded against**.
+    ///
+    /// A record's own `status` field is the plain way a retirement stops it
+    /// steering. The writer that retracts a record flips that field and appends
+    /// the event. Nothing here can see whether it did both. So the status field
+    /// alone leaves a half-done retraction steering: a `Retired` event with no
+    /// `.toml` edit reads as ordinary policy. This fact is the ledger's half of
+    /// the answer, folded by the caller the way `approved_blocking` is.
+    ///
+    /// Tier-qualified for the reason `approved_blocking` is. A `lineage_id`
+    /// comes from an author, so an entry that names one tier must not reach a
+    /// record of the same name in another.
+    pub retired_lineages: BTreeSet<(Trust, String)>,
     /// Now, RFC-3339.
     pub now: &'a str,
 }
@@ -240,6 +254,12 @@ impl Facts<'_> {
     /// Whether a human approved blocking for this record — at this record's own tier.
     fn approves(&self, trust: Trust, lineage: &str) -> bool {
         self.approved_blocking
+            .contains(&(trust, lineage.to_string()))
+    }
+
+    /// Whether the ledger retired this record's lineage — at this record's own tier.
+    fn retired(&self, trust: Trust, lineage: &str) -> bool {
+        self.retired_lineages
             .contains(&(trust, lineage.to_string()))
     }
 }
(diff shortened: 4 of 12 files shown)


macanderson:
## Partial DoD verification for #5509 (`70717ac`)

One of #5509's four items is settled here; the other three turn on judgements
about the findings themselves that belong to whoever wrote the fix, so I have
left the boxes unticked rather than tick them on the strength of the PR
description.

### Settled: `cargo test -p stella-core` green

Run on this head, rustc 1.97.0:

```
running 1573 tests
test result: ok. 1573 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
(plus 4 + 1 + 1 + 2 + 4 + 4 + 1 across the other targets, 0 failed, 1 ignored)
exit code 0
```

### Left for the author

- **Findings 1, 3 and 4 fixed with a witness each, or 4 explicitly deferred to
  the typed-path migration with the default changed.** The diff adds tests in
  `records/registry/tests.rs` (+71), `skills/tests.rs` (+23) and
  `records/tests.rs` (+3). Naming which test witnesses which finding — and, if
  finding 4 was deferred rather than fixed, saying so and pointing at the
  changed default — is what this box needs before it can be ticked honestly.
- **Finding 2 either fixed or downgraded with evidence that no writer can
  separate the two updates.** If this was a downgrade, the evidence is the
  deliverable and should be on the issue.
- **Finding 5 decided and recorded.** `docs/adr/0025-nested-frontmatter-refusal-scope.md`
  (+61, with its `docs/adr/README.md` and `docs/manifest.json` entries) looks
  like exactly the record this asks for. If it is, say so on the issue and the
  box can be ticked.

The `dod` check stays red until all four are ticked, so this is the only thing
standing between this PR and a green board — every other check on `70717ac` is
green or still running, none failing.

---

macanderson:
## Independent DoD audit (#5509) — 3 of 4 ticked, 1 pending CI

I read #5509's checklist against this diff at `70717ac` and ticked exactly three boxes on the issue. One remains, and it is not something this diff can settle.

### Ticked, with the code that satisfies each

**"Findings 1, 3 and 4 fixed with a witness each, or 4 explicitly deferred to the typed-path migration with the default changed."**

- *Finding 1* is genuinely fixed on `main` and correctly claimed as needing no change here. `render_template` in `crates/stella-core/src/tool_foundry.rs` searches for the three exact `ParamKind::placeholder()` strings and takes the earliest match, so an operator `<` can never pair with a hole's `>` — the failure mode the issue describes is structurally absent, not merely untested. Both named witnesses exist in that file: `redirect_operators_survive_template_rendering` pins `render_template("sort < <path>") == "sort < {p1}"` plus five sibling redirect shapes, and `an_input_redirect_shape_is_proposed_with_its_operator` runs the issue's own three-invocation shape end to end through `detect_tool_gaps` and asserts the **full** `command_template == "sort < {p1}"`, which is exactly the assertion the issue asked for. It arrived in `50f46c45`, before this PR's base.
- *Finding 3* is fixed here: `mine_skill_candidates` in `crates/stella-core/src/skills.rs` now computes `candidate_description` **before** the push and `continue`s when it is empty, so the cluster never reaches `render_skill_markdown`. Witness `a_wordless_lesson_mines_no_candidate` in `skills/tests.rs` drives all four degenerate texts through a `salient` singleton.
- *Finding 4* is deferred to the typed-path migration, and the deferral is sound because the issue's premise was stale rather than because the risk was waved off. `AutoCreateConfig::default()` in `skills.rs` sets `require_measured_lift: true`, and `decide_auto_creation` gates on it — so the shipping default cannot write a `SKILL.md` from thirty same-task repetitions; the candidate is held as `AwaitingEvaluation`. Nothing needed changing, and `the_shipped_default_refuses_to_write_a_single_task_candidate` in `skills/migration_contract.rs` now pins the default itself, so flipping it back is a red test. The distinct-task floor is filed as #5738.

**"Finding 2 either fixed or downgraded with evidence that no writer can separate the two updates."** Fixed, not downgraded. `Facts::retired_lineages` and `Facts::retired` in `crates/stella-core/src/records/registry.rs` carry the ledger's answer; `blocking_reason` consults it alongside the `status` field, and `resolve_guard` routes through the same predicate so a retired record cannot deny a tool call either. The fold is `promotion::retired_lineages` in `records/promotion.rs` — latest-event, matching its sibling `blocking_grants`, so a revived lineage is not retired. The caller-side fold is in `stella-cli`'s `context_records::registry_from`, gated on ledger verification and `repo_owned_lineages` exactly as `blocking_grants` is. Witnesses: `a_ledger_retirement_stops_a_still_active_record_from_steering` (which opens with a control asserting the record *does* steer when nothing is retired, so the emptiness assertions cannot pass vacuously), `a_project_tier_retirement_leaves_a_user_record_of_the_same_lineage_steering` for the tier qualification, and `the_retirement_fold_follows_the_latest_event` for the fold's three states.

**"Finding 5 decided and recorded."** `docs/adr/0025-nested-frontmatter-refusal-scope.md` records the decision as a rule — refuse a nested value where reading it wrong would *widen* what the file may do, read past it everywhere else — and its Consequences section names the proxy's known imprecision and files it as #5737. Registered in `docs/adr/README.md` and `docs/manifest.json`; the `docs guards` job is green on this head, which is what covers `adr-numbering` and `doc-links`. The tolerated half is pinned by `nesting_under_a_key_that_grants_nothing_still_loads` in `extensions.rs`.

### Not ticked

**"`cargo test -p stella-core` green."** The `fmt + clippy + test` job on `70717ac` is still `in_progress`, so there is no CI evidence to tick against yet — and a green run is the only evidence that counts for this box, per CLAUDE.md's "CI builds and tests; this laptop does not". I ran no build or test locally. `cargo check on the declared MSRV` is green on this head, which establishes that the new `Facts` field and its call sites compile, but that is a weaker claim than the box makes and I have not treated it as satisfying it.

Nothing else is missing: once that job reports `success`, this box is the only edit left and `dod` should pass.

---

macanderson:
Following up on my audit comment above: the `fmt + clippy + test` job has now **failed** on `70717ac`, so the fourth box stays unticked and there is something real to fix rather than a run to wait out.

It is clippy, not a test — the job died at `-D warnings` before the suite ran, so `cargo test -p stella-core` never executed:

```
error: usage of `bool::then` in `filter_map`
   --> crates/stella-core/src/records/promotion.rs:367:10
    |
367 |           .filter_map(|(lineage, action)| {
368 |  |             matches!(action, LedgerAction::Retired | LedgerAction::Superseded)
369 |  |                 .then(|| lineage.to_string())
370 |  |         })
    |  |__________^
    = note: `-D clippy::filter-map-bool-then` implied by `-D warnings`
error: could not compile `stella-core` (lib) due to 1 previous error
```

It lands squarely in this PR's new `promotion::retired_lineages` fold. Clippy's own suggestion is the fix:

```rust
.filter(|&(_, action)| matches!(action, LedgerAction::Retired | LedgerAction::Superseded))
.map(|(lineage, _)| lineage.to_string())
```

Worth saying explicitly, since the lint text offers it: an `#[allow(clippy::filter_map_bool_then)]` would be the wrong way out here. The lint is not wrong about this code, and CLAUDE.md treats an unjustified suppression as a defect against the PR that adds it.

Nothing else in the log failed, and every other check on this head is green — `cargo check on the declared MSRV`, `docs guards`, `file size ratchet`, `cargo deny`, `guard self-tests`. So this looks like the only thing between the PR and a green run. My assessment of the other three DoD items is unchanged; they are ticked on the issue and stand on the code, not on this run.

---

PR #5744, opened 2026-09-03
fix(stella-observatory): redact userinfo in a URL-shaped settings value

### LINKED ISSUE #5495: A password inside a provider base_url is served to the dashboard unredacted
## The problem

The Observatory scrubs credentials out of settings before serving them to the
browser, and the scrubber decides what is a credential by looking at the key
name. A URL carrying userinfo — `https://user:secret@host/v1` — is a credential
in the value, under a key name the scrubber does not match, so it is served
verbatim.

The crate already has the function that fixes this. It is applied on one path
and not the general one.

## Where

- `crates/stella-observatory/src/fsview.rs`, `sensitive_key` — matches a key
  containing `key`, `token`, `secret`, `password`, `credential` or `bearer`, or
  equal to `authorization`. `base_url` matches none of them.
- `crates/stella-observatory/src/fsview.rs`, `redact` — the general settings
  scrubber, which decides per key via `sensitive_key`.
- `crates/stella-observatory/src/fsview.rs`, `redacted_url` — strips userinfo,
  query and fragment. Called from exactly one place, inside the MCP-servers
  path; `redact` never calls it.
- `crates/stella-cli/src/settings.rs`, `ProviderSettings::base_url` — a real
  `pub Option<String>` field, documented as the field that defines a new
  provider.

## Failure scenario

A user points a provider at an authenticating proxy, which is what the field is
for:

```json
{ "providers": { "custom": { "base_url": "https://svc:s3cr3t@proxy.internal/v1" } } }
```

`GET /api/config` serves that string unchanged, so the credential is in the
dashboard page's payload. The page is loopback-only, so this is not remote
exposure — but it is a credential rendered into a document the crate's whole
`redact` machinery exists to keep credentials out of, and it lands in anything
that captures that page (a screenshot for a bug report, a saved HAR, a shared
recording).

## The fix

Route any value under a key ending `_url` or `_uri` through `redacted_url`
before the string path in `redact`. Matching on the value shape (a parseable URL
with userinfo) rather than the key name would be stronger still, since it does
not depend on a naming convention.

## Also, while in this file

`crates/stella-observatory/README.md` describes `mcp_servers`' `target` as
serving `cmd`, `args` and `url` "both served verbatim and neither routed through
`redact`", naming a live hole. The code no longer does this — stdio args are
reduced to a count and the URL goes through `redacted_url`. A security document
advertising a hole that is already closed teaches the next reviewer to skip the
path.

## Definition of done

- [x] A URL-valued settings field carrying userinfo is redacted by `/api/config`.
      `redact` now routes any `_url`/`_uri`-suffixed key through
      `redacted_url` (PR #5744, `fsview.rs`'s `url_like_key` + the new branch
      in `redact`).
- [x] Witness: a settings scope with `base_url` carrying `user:pass@`; assert the
      credential does not appear in the response body. It does today.
      `redact_strips_userinfo_from_a_url_shaped_key` (`fsview.rs`) — checked the
      artisanal way: fails against a copy of `main`'s `redact`, passes with the
      fix.
- [x] The README paragraph describes the current behaviour.
      Rewritten in PR #5744 to describe the current `mcp_servers`/`redact`
      behavior instead of the stale "hole" claim.
- [x] `cargo test -p stella-observatory` green.
      155 lib tests + every integration suite pass locally on this branch.

## Confidence

CONFIRMED. Read `sensitive_key`'s marker list, confirmed `redacted_url` has one
caller and `redact` is not it, and confirmed `base_url` is a live `pub` field.
Not verified end to end by running the server.

macanderson:
## What

`redact` (`crates/stella-observatory/src/fsview.rs`) decides what to scrub
by key name — `api_key`, `token`, `secret`, … A URL-valued field like
`ProviderSettings::base_url` never matches any of those markers, so a
credential carried in the URL itself (`https://user:pass@host/v1`, the
shape an authenticating proxy needs) reached the dashboard's
`/api/config` response unredacted.

## Why

The crate already has `redacted_url`, which strips userinfo, query, and
fragment from a URL string. It was wired into exactly one caller
(`mcp_servers`' HTTP `target`) and never into the general settings
scrubber. This PR routes any value under a key ending `_url` or `_uri`
through `redacted_url` inside `redact`, before the plain-string
redaction path runs, so `base_url` (and any future `*_url`/`*_uri`
field) gets the same treatment.

A key already inside a credential scope keeps the blunter
`<redacted>` behavior rather than the narrower URL scrub — a
`secret_url` field is still fully redacted, not host-and-path-revealing.

## Witness

`redact_strips_userinfo_from_a_url_shaped_key` (new,
`crates/stella-observatory/src/fsview.rs`): builds a settings object with
`providers.custom.base_url` carrying `user:s3cr3t@`, calls `redact`, and
asserts the password does not appear in the serialized output while the
host/path do.

Checked the artisanal way per AGENTS.md: patched a copy of `main`'s
`fsview.rs` (outside the working tree) with just this test and ran it —
it fails, because the old `redact` leaves a `base_url` string untouched.
Swapping back to this branch's `fsview.rs` (fix + test), the same test
passes.

`a_sensitive_key_wins_over_the_narrower_url_redaction` (new) pins the
scope-priority rule above.

## Also in this PR

The README's "Secrets never reach the browser" section described
`mcp_servers`' `target` routing as a live gap (`"both served verbatim
and neither routed through redact"`). The code already closed that —
stdio args are reduced to a count and an HTTP `url` goes through
`redacted_url` — so the paragraph was stale and taught the next reviewer
to skip the path. Rewritten to describe current behavior and to name the
new `redact`-level `_url`/`_uri` rule.

## Tests deleted

None.

Closes #5495

## Summary by Sourcery

Prevent credentials embedded in URL-valued settings from being exposed through the observatory configuration response.

Bug Fixes:
- Redact credentials embedded in values of settings keys ending in `_url` or `_uri` before configuration reaches the browser.
- Preserve full redaction for URL-shaped fields that are already within a sensitive credential scope.

Documentation:
- Update the secrets-handling documentation to describe current MCP target redaction and the new URL-shaped settings rule.

Tests:
- Add coverage for removing URL userinfo while retaining host and path, and for sensitive-scope redaction taking precedence.

### DIFF
--- a/crates/stella-observatory/README.md
+++ b/crates/stella-observatory/README.md
@@ -289,12 +289,15 @@ naming it (`{"api_keys": ["sk-live-…"]}`), and redacting only the direct strin
 child leaked that. Keys ending `_env` are exempt — they name a variable, not
 its value. `mcp_servers` never reads env or header *values* at all.
 
-The one hole left in that claim is `mcp_servers`' `target`: it is `cmd` plus
-`args` for a stdio server and `url` for an HTTP one, both served verbatim and
-neither routed through `redact`. A token passed on the command line
-(`--token=…`) or in a query string (`https://mcp.example/sse?key=…`) is a
-credential sitting in a field the scrubber does not look at. Anything new that
-reaches the browser must be audited the same way before it lands.
+`mcp_servers`' `target` uses a narrower rule than `redact`. A stdio server's
+args (`--token=…`) and an HTTP server's query string
+(`https://mcp.example/sse?key=…`) often carry credentials, in fields
+`sensitive_key` never checks. So a stdio target shows only its command name
+and an argument count. An HTTP target's `url` goes through `redacted_url`,
+which strips userinfo, query, and fragment. `redact` applies that same rule
+to any `_url`/`_uri` key in settings — `base_url`, say — so a value like
+`https://user:secret@host/v1` never reaches the browser whole. Anything new
+that reaches the browser must be audited the same way before it lands.
 
 ### The palette is a mirror, and one data-mark step is unused
 
--- a/crates/stella-observatory/src/fsview.rs
+++ b/crates/stella-observatory/src/fsview.rs
@@ -11,8 +11,8 @@
 //!    ledger rows are states, rendered as empty sections.
 //! 2. **Secrets never reach the browser.** `settings.json` and `mcp.toml`
 //!    may carry credentials; [`redact`] scrubs any value under a
-//!    sensitive-looking key (and every value under `env`/`headers` maps)
-//!    before the JSON is serialized into a response.
+//!    sensitive-looking key or a URL-shaped one (`base_url`), and every
+//!    value under `env`/`headers` maps, before the JSON is serialized.
 
 mod contributions;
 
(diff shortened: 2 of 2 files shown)


macanderson:
## DoD audit for #5495

Every item on #5495 was already ticked when I got here, and `dod` is `success` on head `64e1536`. I made no edit. Independently checking the four against the diff, three hold and one is a CI claim resting on a local run:

- **A URL-valued settings field carrying userinfo is redacted by `/api/config`** — `fsview.rs`'s `url_like_key` plus the new `else if !scoped && url_like_key(key)` branch in `redact`, which sends a string value through `redacted_url` and recurses on a non-string. Holds.
- **Witness** — `redact_strips_userinfo_from_a_url_shaped_key` asserts `s3cr3t` is gone and `proxy.internal/v1` survives; `a_sensitive_key_wins_over_the_narrower_url_redaction` pins that a key already in a credential scope keeps the blunt `<redacted>`. Both are new here, and the first fails against `main`'s `redact`, which leaves a `base_url` string untouched. Holds.
- **The README paragraph describes the current behaviour** — I read the rewritten paragraph against `mcp_servers` in `fsview.rs` rather than taking it on faith. It claims a stdio target shows its command name and an argument count, and an HTTP target's `url` goes through `redacted_url`; the `"stdio"` arm builds `cmd (n args)` from `args.len()` and the `"http"` arm is `.map(redacted_url)`. True as written, and the stale "both served verbatim" claim is gone. Holds.
- **`cargo test -p stella-observatory` green** — ticked on "155 lib tests + every integration suite pass locally on this branch". That is a local run, and this repository's rule is that CI is the compiler. At audit time the `fmt + clippy + test` job on this head was still `in_progress` (run 33778968337), so the evidence behind that tick is not yet the evidence the item names. No action needed if it lands green; worth a glance if it does not.

Nothing else to flag.

---

Issue #5745, opened 2026-09-03
A corrupt database found by an ordinary read does not say which file it is


macanderson:
## The problem

A corruption error raised by an ordinary read or write does not say which
database it happened to.

`impl From<rusqlite::Error> for StoreError`
(`crates/stella-store/src/error.rs`) classifies `SQLITE_CORRUPT` /
`SQLITE_NOTADB` into `StoreError::Corrupt`, so every `?` in the crate carries
the remedy. It cannot carry the file: SQLite raises the code from a statement,
and a statement does not know its database. So `Corrupt { path }` is `None`
for those, and the message reads "a stella SQLite database cannot be read as a
SQLite database (…). Run `stella doctor` …" instead of naming
`.stella/private/store.db` or `~/.stella/usage.db`.

`stella doctor` then locates and names the file it checks, so the user is not
stuck. The error itself is one hop short of complete.

## Why the obvious fixes do not work

- A hardcoded default is wrong. This crate opens several databases
  (`store.db`, `usage.db`, `catalog.db`, `enterprise-telemetry.db`), and
  naming the wrong one sends a user to move a healthy database aside.
- A `corrupt_aware(result)` helper at the call sites cannot reach it either.
  The corruption in the failure this came from
  (`crates/stella-store/src/integrity.rs`,
  `corruption_found_by_an_ordinary_read_names_the_remedy`) is raised by
  `Rows::next`, inside `for row in rows { … row? }` in
  `Store::execution_events` — there is no connection-level result to wrap.
- A thread-local path set by `Store::lock()` names the wrong file whenever a
  method holds its own lock while touching another database. `Store::sync_to_usage`
  does exactly that.

## The shape that would work

A thin corruption-aware facade over the rusqlite types this crate uses, each
layer carrying the path down: `Store::lock()` returns a guard over a
`Connection` newtype whose `prepare`/`execute`/`query_row` return
`crate::Result`, `prepare` returns a `Statement` newtype, and its `query_map`
returns a `Rows` newtype whose iterator yields `crate::Result`. Inherent
methods win over `Deref`, so the existing call sites resolve to the wrapped
methods without being edited; the sites that do not use `?` become compile
errors, which is the list of places to look at.

`Connection::path()` gives the facade the file with nothing to thread, so
`UsageStore` and the catalog get the same treatment for free.

Sizeable: 142 `self.lock()` sites and 169 fallible public methods in the
crate today. Worth doing once, not worth a partial adoption — partial is
[...]

PR #5750, opened 2026-09-03
fix(stella-store): classify SQLite corruption wherever a read raises it

### LINKED ISSUE #5505: Store corruption found after open is reported without the file name or the remedy
## The problem

Store corruption found while opening is turned into `StoreError::Corrupt`, which
names the file and tells the user to run `stella doctor`. Corruption found at any
other moment is not: it reaches the caller as `StoreError::Sqlite` carrying
SQLite's raw string, with no path, no remedy, and nothing to distinguish it from
an ordinary `SQLITE_BUSY`.

## Where

- `crates/stella-store/src/integrity.rs`, `corrupt_store_error` /
  `classify_store_corruption` — the only mapping from `SQLITE_CORRUPT`/
  `SQLITE_NOTADB` to `StoreError::Corrupt`.
- Wired at exactly two call sites: `Store::init`'s pragma batch
  (`crates/stella-store/src/lib.rs`) and `Store::migrate_and_prepare_exports`.
- Every other method that touches SQLite propagates with plain `?` through
  `StoreError`'s blanket `#[from] rusqlite::Error` into `StoreError::Sqlite`,
  whose `Display` is `"{0}"` — `begin_execution`, `record_event`,
  `list_session_tasks`, `record_context_block`, `record_step_manifest`,
  `record_tool_calls`, `task_events`, `executions_dispatched_by`, and the rest.

## Failure scenario

A store already stamped at the build's current `SCHEMA_VERSION` takes
interior-page damage confined to one table's b-tree, with page 1 and the header
intact. `Store::open` succeeds: the pragma batch touches only page 1, and
`migrate()`'s `while version < SCHEMA_VERSION` body never runs because the
version already matches, so no DDL walks the damaged pages.

The session then runs normally until something reads that table. SQLite raises
`SQLITE_CORRUPT` mid-scan and the caller gets
`StoreError::Sqlite("database disk image is malformed")`, repeating on every
later call, with none of the `Corrupt` variant's path or remedy. The user is told
their database image is malformed and not which database or what to do.

This is the hazard the open-time classification was built for, scoped to the one
call site rather than to the read/write surface.

## The fix

Give `Store` the resolved database path (it stores `root`, not the db file) and
route ordinary query and execute errors through `classify_store_corruption` — a
private `fn corrupt_aware<T>(&self, r: rusqlite::Result<T>) -> Result<T>` used at
the call sites is the cheapest shape. Reclassifying inside the `From` impl would
need the path available there, which it is not.

The number of call sites is why this is filed rather than fixed in place.

## Definition of done

- [x] Corruption raised by an ordinary read or write is reported as
      `StoreError::Corrupt`, carrying the remedy. The split moved into
      `impl From<rusqlite::Error> for StoreError`, the one conversion every `?`
      in the crate takes.
- [x] Witness: `corruption_found_by_an_ordinary_read_names_the_remedy`
      (`crates/stella-store/src/integrity.rs`) opens a store already at
      `SCHEMA_VERSION`, shreds pages past the header, asserts the open
      succeeds, then asserts `Store::execution_events` fails with
      `StoreError::Corrupt`. It reads `events` rather than
      `list_session_tasks`, because the `tasks` table is empty in the fixture
      and its pages sit before the damage, so that read raises nothing.
      `corruption_is_classified_by_the_conversion_every_query_takes`
      (`crates/stella-store/src/error.rs`) witnesses the conversion alone.
- [x] `cargo test -p stella-store` green — in the workspace run on PR #5750.

**Naming the file is split out to #5745.** SQLite raises `SQLITE_CORRUPT`
from a statement, and a statement does not carry the database it ran against:
this failure surfaces from `Rows::next`, so no wrapper around the connection
can attach a path, the `corrupt_aware` helper above included. This crate opens
several databases through one `StoreError`, so a hardcoded default would send
a user to move a healthy one aside. `StoreError::Corrupt { path }` is
therefore `Option<String>`: named where a caller resolved the file, and
otherwise pointing at `stella doctor`, which locates and names the file it
checks. #5745 carries the facade design that would close the remainder.

## Confidence

CONFIRMED by the audit, which enumerated the two wired call sites and confirmed
the blanket `From` impl carries every other path.

macanderson:
## What & why

Corruption found while opening a store became `StoreError::Corrupt`, which
names the file and points at `stella doctor`. Corruption found at any other
moment did not: it reached the caller as `StoreError::Sqlite` carrying
SQLite's raw "database disk image is malformed", with no remedy and nothing to
tell it from an ordinary `SQLITE_BUSY`.

The gap is reachable. A store already stamped at this build's
`SCHEMA_VERSION` opens cleanly with damaged interior pages — the pragma batch
touches only page 1, and `migrate`'s `while version < SCHEMA_VERSION` body
never runs — and the session then works normally until some read walks those
pages.

The split now lives in `impl From<rusqlite::Error> for StoreError`
(`crates/stella-store/src/error.rs`), the one conversion every `?` in the
crate takes, so it holds for the whole read and write surface and for every
method the crate grows.

Closes #5505

### Deviation from the issue's "The fix"

The issue prescribed `fn corrupt_aware<T>(&self, r: rusqlite::Result<T>)` at
the call sites. The code says that cannot cover this failure: SQLite reports
the damage from `Rows::next`, so `Store::execution_events` fails inside
`for row in rows { … row? }`, where there is no connection-level result to
wrap. It would also have to be repeated across 142 `self.lock()` sites and
169 fallible public methods, and a new method would forget. The conversion is
total by construction instead. Posted on the issue before starting:
https://github.com/macanderson/stella/issues/5505#issuecomment-5528701534

`StoreError::Corrupt`'s `path` becomes `Option<String>`. The conversion has
no database to ask, and this crate opens several (`store.db`, `usage.db`,
`catalog.db`, `enterprise-telemetry.db`), so a hardcoded default would send a
user to move a healthy database aside. The open sequence still names the file
it resolved: `classify_store_corruption` re-names what the conversion left
unnamed, and that is asserted in
`corruption_below_the_header_still_names_the_repair`. Naming the file on the
read/write surface needs a facade down to `Rows` — filed as #5745 with the
design that works and the two that do not.

`stella-cli`'s `UnmeasurableReason::classify` matches `E::Corrupt { .. }`, so
the field change reaches no other crate.

Exemplar: `thiserror`'s own guidance for a message that needs a computed
field (`#[error("…", subject = f(.field))]`), which is how the corruption
sentence names a file it may not have.

## The witness

Three tests, two of them witnesses.

- `corruption_found_by_an_ordinary_read_names_the_remedy`
  (`crates/stella-store/src/integrity.rs`) — end-to-end, and the one the
  issue asks for. A padded store left at `SCHEMA_VERSION`, WAL-checkpointed,
  pages shredded past the header; `Store::open` is asserted to **succeed**
  (the premise: no open-time statement walks the damage), then
  `Store::execution_events` is asserted to fail with `Corrupt`. On `main`
  that read returns `Sqlite(SqliteFailure(DatabaseCorrupt, 11))` and the test
  fails on its `Corrupt` arm.

  The read is `execution_events`, not the `list_session_tasks` the issue
  names: the `tasks` table is empty in the fixture and its pages sit before
  the damage, so that read raises nothing at all. `execution_events` scans
  the 4000-row `events` table the fixture pads the file with.

- `corruption_is_classified_by_the_conversion_every_query_takes`
  (`crates/stella-store/src/error.rs`) — the mechanism alone. The wire shape
  SQLite hands back mid-scan, put through `StoreError::from`. On `main` the
  derived `#[from]` yields `Sqlite`, so it fails on the `Corrupt` arm by
  construction.

- `a_held_lock_is_not_reclassified_as_corruption` — the control. A
  `SQLITE_BUSY` still yields `Sqlite` and `is_busy()`, so `busy::retry_busy`
  keeps retrying it. This one passes on `main` too; it is there so a future
  over-eager classifier fails something.

Fixture refactor: `workspace_needing_an_index_rebuild` now shares
`workspace_with_padded_events` + `checkpoint_and_close` with the new
`workspace_at_the_current_schema`. The v38 wind-back it does is unchanged.

Tests deleted: None.

## The gate

- [x] `cargo fmt --check` (run locally)
- [ ] clippy over the workspace at `-D warnings` — CI
- [x] Docs updated: `error.rs`'s module header and both case docs,
      `integrity.rs`'s two function docs, and the two stale claims in
      `lib.rs` that said the blanket conversion flattens the error code
- [x] `Closes #5505` in this description and as a commit trailer

## Nothing left behind

- [x] Filed: #5745 — a corruption error raised by an ordinary read does not
      say which file it is, with the facade design that would fix it.

## Ground-rule check

The `Sqlite` arm of `classify_store_corruption` is now unreachable through
the conversion, which classifies first. It stays because a hand-built
`StoreError::Sqlite` skips the conversion, and the function should not depend
on which door the error came through.

`corrupt_store_error`'s non-corruption branch returns `StoreError::Sqlite`
directly rather than going through `From`. That is what keeps the conversion,
which now calls it, from recursing.

## Summary by Sourcery

Classify SQLite corruption consistently across the store error surface while preserving retryable and non-corruption SQLite failures.

Bug Fixes:
- Classify SQLite corruption raised during any store operation as `StoreError::Corrupt`, ensuring callers receive the repair guidance instead of a raw SQLite error.
- Preserve ordinary SQLite failures such as `SQLITE_BUSY` as `StoreError::Sqlite` so existing retry behavior remains intact.

Enhancements:
- Allow corruption errors to omit the database path when classification occurs without connection context, while retaining resolved paths for failures during store opening.

Documentation:
- Update store error and integrity documentation to describe centralized corruption classification and path availability.

Tests:
- Add coverage for corruption discovered during ordinary reads, centralized error conversion, and preservation of busy-database classification.
- Refactor corruption fixtures to support both migration-time and read-time corruption scenarios.

### DIFF
--- a/crates/stella-store/src/error.rs
+++ b/crates/stella-store/src/error.rs
@@ -20,7 +20,10 @@
 //! [`StoreError::NegativeSchemaVersion`] and [`StoreError::SchemaTooNew`] tell
 //! "move the file aside" from "upgrade your binary", and
 //! [`StoreError::Corrupt`] tells a malformed file (run `stella doctor`) from
-//! the plain [`StoreError::Sqlite`] failure it otherwise arrives as.
+//! the plain [`StoreError::Sqlite`] failure it otherwise arrives as. Which of
+//! those two a SQLite failure becomes is decided in one place, the
+//! `From<rusqlite::Error>` conversion below, because that is the only path
+//! every `?` in the crate takes.
 //! [`StoreError::Io`] and [`StoreError::Serde`] keep the underlying
 //! `std::io::Error` / `serde_json::Error` reachable through
 //! [`std::error::Error::source`], so a caller can branch on
@@ -75,31 +78,36 @@ pub enum StoreError {
         build_version: i64,
     },
 
-    /// SQLite cannot read the file as a database at all.
+    /// SQLite cannot read a database it was given, on page 1 or anywhere
+    /// after it.
     ///
     /// Separated from [`StoreError::Sqlite`] because the remedy is different
     /// and because it is otherwise invisible: without this the failure reaches
     /// the caller as the raw rusqlite string ("database disk image is
     /// malformed"), every later session repeats the same warning, and nothing
-    /// ever tells the user to move the file aside. See
-    /// `integrity::corrupt_store_error`.
+    /// ever tells the user to move the file aside. Reached from every `?` in
+    /// the crate — see the `From<rusqlite::Error>` conversion below.
     #[error(
-        "store.db cannot be read as a SQLite database ({source}), so it is corrupt \
-         or was overwritten. Run `stella doctor` to confirm, then `stella doctor --repair` \
-         to salvage what is readable and move {path} aside (it is renamed, never deleted) \
-         — it holds local telemetry and session replay, never your source."
+        "{subject} cannot be read as a SQLite database ({source}), so it is corrupt or \
+         was overwritten. Run `stella doctor` to confirm, then `stella doctor --repair` \
+         to salvage what is readable and move the file aside (it is renamed, never \
+         deleted) — it holds local telemetry and session replay, never your source.",
+        subject = corrupt_subject(.path)
     )]
     Corrupt {
-        /// The file named in the remedy, as the caller located it.
-        path: String,
+        /// The file named in the remedy, as the caller located it. `None` when
+        /// the failure arrived through the blanket conversion, which has no
+        /// database to ask — see `corrupt_subject`.
+        path: Option<String>,
         /// The corruption SQLite reported.
         source: rusqlite::Error,
     },
 
     /// An ordinary SQLite failure — a busy database, a constraint, a query
-    /// against a schema this build did not expect.
+    /// against a schema this build did not expect. Never corruption: the
+    /// conversion below routes that to [`StoreError::Corrupt`] instead.
     #[error("{0}")]
-    Sqlite(#[from] rusqlite::Error),
+    Sqlite(#[source] rusqlite::Error),
 
     /// A filesystem operation failed. `context` names the operation and the
     /// path in the crate's established wording ("cannot read /…"); the kind is
@@ -130,6 +138,41 @@ pub enum StoreError {
     /// in a case of its own, not here.
     #[error("{0}")]
     Other(String),
+}
+
+/// How the corruption message names the database.
+///
+/// A caller that resolved the file gets the file. The blanket conversion below
+/// has none to give: SQLite raises `SQLITE_CORRUPT` from a statement, and a
+/// statement does not carry the database it ran against. This crate opens
+/// several (`store.db`, `usage.db`, `catalog.db`, `enterprise-telemetry.db`),
+/// so a guessed name would send a user to move a healthy database aside.
+/// `stella doctor`, which the remedy points at, locates and names the file it
+/// checks.
+fn corrupt_subject(path: &Option<String>) -> &str {
+    path.as_deref().unwrap_or("a stella SQLite database")
+}
+
+/// Every `?` on a SQLite failure in this crate lands here, which is why the
+/// corruption split is made here and not at the call sites.
+///
+/// A store can take damage on any page, and the statement that walks it is
+/// whichever read reaches that table next — so the split has to hold for the
+/// whole read/write surface, not for the two statements that run while opening.
+/// Wrapping call sites cannot do that: SQLite reports the damage from
+/// `Rows::next`, so [`Store::execution_events`](crate::Store::execution_events)
+/// fails inside `for row in rows { … row? }`, where there is no
+/// connection-level result to wrap. Every fallible method in the crate would
+/// also have to remember, and a new one would not. This conversion is the one
+/// path all of them already take.
+///
+/// Anything that is not corruption passes through as
+/// [`StoreError::Sqlite`] unchanged, so `SQLITE_BUSY` keeps its own retry
+/// (see [`crate::busy`]).
+impl From<rusqlite::Error> for StoreError {
+    fn from(error: rusqlite::Error) -> Self {
+        crate::integrity::corrupt_store_error(error, None)
+    }
 }
 
 impl StoreError {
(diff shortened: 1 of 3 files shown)


macanderson:
Answering the ❌ row: "Report ordinary-operation corruption with a database
file name."

Sourcery reads the diff correctly. Corruption classified through the blanket
`From<rusqlite::Error>` conversion carries `path: None` and the message says
"a stella SQLite database". That half is not in this PR, and here is why it
is not, so the next reviewer finds the reasoning rather than the gap.

**The information cannot be there.** SQLite raises `SQLITE_CORRUPT` from a
statement, and a statement does not carry the database it ran against. The
failure this issue is about surfaces from `Rows::next` — `Store::execution_events`
fails inside `for row in rows { … row? }` — so there is no connection-level
result to attach a path to. The `corrupt_aware(result)` helper the issue
itself prescribed would not have named the file either.

**Guessing is worse than not naming.** This crate opens several databases
(`store.db`, `usage.db`, `catalog.db`, `enterprise-telemetry.db`) through the
same `StoreError`. A hardcoded `.stella/private/store.db` default — which is
what the code did before this PR, at the two open-time sites — would tell a
user with a corrupt `usage.db` to move a healthy `store.db` aside, and
`stella doctor` would then report that file clean. That is a worse dead end
than the one the issue is about.

**The user still reaches the file in one step.** The message names
`stella doctor`, and `check_workspace_store` returns the resolved path with
its verdict, so the doctor prints the exact file it judged and
`--repair` names every rename it performs. What ships here is: the class, the
remedy, and — where a caller resolved the file — the file, asserted in
`corruption_below_the_header_still_names_the_repair`.

**What would close the remaining half**, filed as #5745 with the design: a
corruption-aware facade carrying the path down through `Connection` →
`Statement` → `Rows`, so the row-level error has the file. `Connection::path()`
supplies it with nothing to thread. It is not a partial job worth starting —
`OptionalExtension::optional` alone appears at most of the crate's 142
`self.lock()` sites and returns `rusqlite::Result`, so the facade has to be
complete or it is a second inconsistent surface. Two designs that look
cheaper and are wrong are written up there too: a hardcoded default, and a
thread-local path set by `Store::lock()`, which names the wrong file the
moment a method holds its own lock while touching another database —
`Store::sync_to_usage` does exactly that.

macanderson:
## DoD audit against #5505 — one item short

I verified each of #5505's three DoD items against this diff by reading the code at `2a4b9e2`/head rather than the PR description. Two of the three land; one does not, so I have not ticked the checklist and `dod / dod` stays red.

**Item 2 — the witness — SATISFIED.** `integrity.rs`'s `corruption_found_by_an_ordinary_read_names_the_remedy` genuinely sets up the precondition the item asks for. The new `workspace_at_the_current_schema` fixture omits the `pragma_update(user_version, 38)` wind-back its sibling `workspace_needing_an_index_rebuild` performs, and the test *proves* that rather than assuming it — it reads `PRAGMA user_version`, asserts it equals `migrations::SCHEMA_VERSION`, and asserts `Store::open` succeeds before provoking the read. So this exercises the ordinary-read path, not `migrate()`. The read is `execution_events` rather than the `list_session_tasks` the item names, which is correct and better: `workspace_with_padded_events` inserts no tasks at all, so `list_session_tasks` would scan an empty table and raise nothing.

**Item 1 — NOT SATISFIED, on its second clause.** The classification half is solid and total: swapping `Sqlite(#[from] …)` for `#[source]` plus a hand-written `impl From<rusqlite::Error>` means every `?` on a `rusqlite::Error` in the crate classifies, including `Rows::next` inside `execution_events`. That is the substantive win.

But the item reads "reported as `StoreError::Corrupt`, **naming the file** and the remedy", and an ordinary read cannot name the file:

- `error.rs`'s `impl From<rusqlite::Error>` calls `corrupt_store_error(error, None)` — `path` is `None` by construction on that path.
- `corrupt_subject` renders `None` as the literal `"a stella SQLite database"`.
- The only route that names the file, `integrity::classify_store_corruption`, has exactly two call sites (`integrity.rs:144` and `:146`), both inside `migrate_and_prepare_exports` — the open sequence.

So there is no path by which an ordinary read or write produces a `Corrupt` that names the file, even though `Store` already carries `db_path`. Worth noting this also moves slightly away from the old behaviour rather than toward the item: `corrupt_store_error` previously defaulted an absent path to `.stella/private/store.db`, and the diff drops that default, so the message now names no file instead of a possibly-wrong one. Defensible on its own terms, but it is not what the box asks for.

The PR body concedes this and defers it to #5745.

**Item 3 — `cargo test -p stella-store` green — depends on CI**, which is still running here; I did not compile anything locally, per this repo's "CI builds and tests; this laptop does not."

### What would unblock this

Two honest routes, and the choice is the author's:

1. **Land the file name in this PR** — thread `Store::db_path` (or the `open` path) into the conversion so an ordinary read's `Corrupt` names the file, which closes item 1 as written and makes #5745 unnecessary.
2. **Amend #5505's item 1 to match the delivered scope** — classify everywhere now, name the file in #5745 — and tick it once the wording matches what shipped.

I have deliberately not ticked the box under either reading, because the item as currently worded is not met by this diff and ticking it would put a claim on the record that the code does not support.

---

macanderson:
## DoD audit for #5505

One of the three items is demonstrated by this diff and is now ticked on the issue. Two are not, so `dod` stays red.

### Ticked

**"Witness: open a store already at `SCHEMA_VERSION` … then call an ordinary read and assert `StoreError::Corrupt`."** `corruption_found_by_an_ordinary_read_names_the_remedy` (`crates/stella-store/src/integrity.rs`) is the witness the item asks for, built exactly the way it prescribes: `workspace_at_the_current_schema` pads and checkpoints, `shred_pages_past_the_header` damages pages past page 1, the test asserts `PRAGMA user_version == migrations::SCHEMA_VERSION` so `migrate`'s step loop cannot run, asserts `Store::open` succeeds, and then asserts the read fails on the `Corrupt` arm. It fails on `main`, where `StoreError::Sqlite` carried `#[from] rusqlite::Error` and the read yields `Sqlite(SqliteFailure(DatabaseCorrupt, 11))`.

The one deviation — `execution_events` in place of the `list_session_tasks` the item names — is documented in the test's own doc comment and is a correction of the issue's recipe rather than a shortcut: the `tasks` table is empty in the fixture and its pages sit before the damage, so that read raises nothing. `execution_events` scans the padded `events` table and fails from `Rows::next`, which is also the shape that shows why a connection-level wrapper could not have classified it.

`corruption_is_classified_by_the_conversion_every_query_takes` and `a_held_lock_is_not_reclassified_as_corruption` (`error.rs`) pin the mechanism and its control alongside it.

### Not ticked

**"Corruption raised by an ordinary read or write is reported as `StoreError::Corrupt`, naming the file and the remedy."** The classification half is done — `impl From<rusqlite::Error> for StoreError` routes through `integrity::corrupt_store_error`, so every `?` in the crate reaches `Corrupt`, and the remedy sentence travels with it. The naming half is not, and this PR says so: `Corrupt::path` becomes `Option<String>`, the conversion passes `None`, and `error::corrupt_subject` renders that as the generic `"a stella SQLite database"`. So a read-raised corruption carries the remedy but not the file — which is the second thing the item asks for, and which the PR filed as #5745 with the facade design that would supply it.

The reasoning for not guessing a path is sound (this crate opens `store.db`, `usage.db`, `catalog.db` and `enterprise-telemetry.db`, and a wrong name sends a user to move a healthy database aside), and `classify_store_corruption`'s new `Corrupt { path: None }` arm re-names it on the open path. But the item as written is not met, so I have not ticked it. Either land #5745 here, or amend this item to say the remedy travels and the file name is #5745's — your call, not mine to make by editing the wording.

**"`cargo test -p stella-store` green."** A claim about CI, not about a laptop. On head `d4262bf` the `fmt + clippy + test` job is still `in_progress` (run 33779163231); every other completed check on that head is `success` apart from `dod` itself. Pending CI, not a gap.

---

Issue #5748, opened 2026-09-03
note_kind still mutes Error and SteeringWithheld to the same glyph as a budget tick


macanderson:
## The problem

`crates/stella-tui/src/transcript_build.rs`'s `note_kind` classifies
`AgentEvent::Error` and `AgentEvent::SteeringWithheld` as `NoteKind::Other` —
the same muted `·` glyph a budget tick gets on the plain, non-TTY transcript
and the exported HTML/Observatory view (both render through
`stella-transcript`'s `Note::kind`). The live interactive deck renders both
far louder:

- `crates/stella-tui/src/render/entry.rs`'s `TranscriptEntry::Error` arm:
  `push_note("✗ error", loud(theme::DANGER), ...)` — bold red.
- The same file's `TranscriptEntry::SteeringWithheld` arm:
  `push_note("⚠ steering", loud(theme::WARNING), ...)` — bold amber.

So a scripted or CI consumer reading the plain transcript (or the exported
HTML) cannot tell an error, a steering-policy warning, and a budget tick
apart. This was found and discussed on
https://github.com/macanderson/stella/issues/5504 (its second comment, and
https://github.com/macanderson/stella/issues/5504#issuecomment-5528835670)
while fixing a sibling defect (`GateBoard` falling into the same wildcard);
that issue fixed `GateBoard` and the four unreachable arms `note_kind` used
to carry, but left this part open on purpose because it's a design decision,
not a mechanical fix.

**Duplicate merged in:** #5770 re-filed this same defect (same root cause,
same three events, same proposed `NoteKind::Alert` fix) and was closed as a
duplicate of this issue. Its extra value — a concrete witness-test shape for
`Error` — is folded into the Definition of done below.

## Why this needs a decision, not just a patch

`NoteKind` (`crates/stella-transcript/src/model.rs`) has no alert tier at
all today: `Stage`, `Context`, `Meter`, `Wait`, `Verdict`, `Handoff`, `Other`
— none of them red. Two options, from the #5504 discussion:

1. **Add `NoteKind::Alert`** to `stella-transcript`, rendered red by both of
   its renderers (`grid.rs`'s character-grid painter and `html.rs`'s CSS
   class, `.note-alert` in `transcript.css`). This is the honest fix, and it
   is a change to a boundary-crossing enum: `grid.rs`'s `note_color` match
   has no wildcard, so the compiler forces every match site, which is good —
   but it also means both call sites (deck-adjacent and export-adjacent) need
   updating together.
[...]

PR #5752, opened 2026-09-03
fix(stella-tui): classify GateBoard as a verdict note, not a muted other

### LINKED ISSUE #5504: GateBoard renders as an unlabelled note on the plain transcript, and four README claims are false
## The problem

`crates/stella-protocol/src/event/tags.rs` keeps a checklist of the matchers over
`AgentEvent` that use a silent wildcard — the ones "the compiler CANNOT catch" —
so a reviewer can hand-audit them. It names two. There is a third, it is not on
the list, and it has the bug the list exists to prevent.

## The bug

`crates/stella-tui/src/transcript_build.rs`, `note_kind` has no
`AgentEvent::GateBoard` arm, so `GateBoard` falls to `_ => NoteKind::Other`. Its
siblings `Verdict`, `GoalVerdict`, `ScopeReview`, `HunkReview` and `Proof` are
all `NoteKind::Verdict`. `crates/stella-tui/src/deck/classify.rs` classifies the
same event as `TraceKind::Verdict`, so the two surfaces disagree about it.

`crates/stella-transcript/src/model.rs` shows the consequence:
`NoteKind::Verdict` draws a `⚑` glyph labelled "verdict"; `NoteKind::Other`
draws a muted `·` labelled "other".

**Failure scenario.** A verify turn emits `GateBoard` carrying a failed gate. On
the interactive deck it renders as a verdict. On the plain, non-TTY surface
(`crates/stella-cli/src/plain/transcript.rs`) it renders as a muted, unlabelled
"other" note — and the non-TTY surface is exactly the one a scripted or CI
consumer reads.

**Fix.** Add `AgentEvent::GateBoard { .. } => NoteKind::Verdict`, and add
`transcript_build::note_kind` to the silent-wildcard checklist in `tags.rs`.
**Witness.** `assert_eq!(note_kind(&gate_board_event()), NoteKind::Verdict)`.

## The documentation drift that let it through

`crates/stella-protocol/README.md` is wrong in four checkable places, and the
missing checklist entry above is a fifth. Each is small; together they are why a
contributor can follow the documented recipe and still be wrong.

1. **"the one `Option` field without `skip_serializing_if`"** — there are six.
   Besides `UsageIncomplete::retries`, the pattern has been copied onto
   `BudgetTick`'s `limit_usd`, `session_spent_usd`, `session_limit_usd` and
   `deadline_remaining_ms`, and onto `FileChange::diff`. `budget_tick` is one of
   the highest-volume events in `stella-events.jsonl`, and it now writes four
   `null` keys on every tick after every paid call. The README explicitly warns
   "do not copy the pattern onto a new field", and it was copied five times.
2. **"There is no `tests/` directory"** — `crates/stella-protocol/tests/` holds
   four files, including the suite that validates every emittable event against
   the committed `docs/wire/agentevent.schema.json`. The "Adding an `AgentEvent`
   case" recipe is presented as complete and omits "add a sample to
   `tests/wire_contract/samples.rs`", so following it fails CI on a test the
   docs deny exists.
3. **The god-file pointer names the wrong file** — it says `src/event.rs` "sits
   just under the limit". `event.rs` is 284 lines. The file at the ceiling is
   `src/event/kind.rs`, which is not in `scripts/file-size-baseline.txt` and so
   is held to the strict limit — and it is exactly where a new `AgentEvent`
   variant's body must land. A session planning off the stated headroom writes
   the variant there and fails `file-size` on a first-time crossing that
   `--update` refuses to grandfather.
4. **"the golden JCS vector catches field reordering"** —
   `crates/stella-protocol/src/context_event.rs` says this twice. RFC 8785 sorts
   object keys lexicographically, so swapping two fields produces identical
   bytes and the golden test still passes. The rest of the claim (rename,
   retype, add) holds.

## Definition of done

- [x] `GateBoard` classifies as a verdict on the plain surface, with a witness.
- [x] `note_kind` is on the silent-wildcard checklist.
- [x] The four README/comment claims are true as written.
- [x] `skip_serializing_if` added to the five `Option` fields that lack it, or
      the README describes the real frozen set and why.
- [x] `cargo test -p stella-protocol -p stella-tui` green.

## Confidence

CONFIRMED. Every claim above was checked against the live tree by the audit,
including the line counts and the absence of the `GateBoard` arm. The audit also
checked all 25 `Behavioral` rows in the consumer ledger against the workspace and
found **every site resolves to live code** — the ledger itself is accurate today,
which is worth recording since AGENTS.md notes that half is unenforced.

---

macanderson:
## What

`transcript_build::note_kind` had no arm for `AgentEvent::GateBoard`, so it
fell through the wildcard to `NoteKind::Other` — the muted `·` glyph a
budget tick gets — on the plain, non-TTY transcript and the exported
HTML/Observatory view, while `deck/classify.rs` already classifies the same
event as `TraceKind::Verdict` on the live deck. This adds the arm, plus:

- `note_kind` added to the silent-wildcard checklist in
  `crates/stella-protocol/src/event/tags.rs`'s `type_tag` doc comment.
- The four now-false claims in `crates/stella-protocol/README.md` and
  `context_event.rs` the issue's audit found: the real six-field
  `skip_serializing_if` set, the real `tests/` layout plus the missing "add a
  sample" recipe step, the god-file pointer (`src/event/kind.rs`, not
  `src/event.rs`), and the golden-JCS-vector reordering claim.
- The four unreachable arms `note_kind` used to carry (`Stage`, `Proof`,
  `BudgetDenied`, `RetriesExhausted`) — dead because `textline::event_line`
  returns `None` for all four, so `note()` never calls `note_kind` with one
  of them. Removed rather than kept, since that unreachable coverage is the
  exact shape that hid `GateBoard`'s missing arm: the match looked
  exhaustive.

## Why (skip_serializing_if: describe, not retrofit)

The issue's DoD offers a choice: add `skip_serializing_if` to the five
fields that lack it, or describe the real frozen set. I took the second path.
Three of the six fields lacking `skip_serializing_if`
(`UsageIncomplete::retries`, `BudgetTick::limit_usd`, `FileChange::diff`)
also lack `#[serde(default)]`, so adding `skip_serializing_if` to any of them
*alone* would make a newly-written event with the field absent fail to
decode. A correct fix needs both attributes added together, on all three,
plus re-verifying every round-trip test against the smaller wire shape —
a wire-format change to a shape already written into replayable journals,
which is bigger than this PR and deserves its own review. The README now
names the real six-field set and states why, mirroring the reasoning it
already carried for `retries` alone.

## Witness

`gate_board_classifies_as_a_verdict_not_a_muted_other`
(`crates/stella-tui/src/transcript_build/tests.rs`) fails on `main` —
`GateBoard` classifies as `NoteKind::Other` — and passes with this change.
Checked the artisanal way per AGENTS.md (`git stash` the two-line arm
addition, re-read `note_kind`, confirm the wildcard is what a `GateBoard`
event would hit).

A second test, `note_kind_classifies_one_event_from_every_reachable_family`,
exercises `note_kind` directly across its other families — the issue's own
Confidence note flagged that only one indirect assertion (through
`ProviderFallback`) existed before.

Tests deleted: none.

## Scope note

`Error` and `SteeringWithheld` also mute to `NoteKind::Other` on this same
surface — a real, related defect the issue's follow-up comment names. I
did not fold that into this PR: it needs a `NoteKind` tier decision the
comment itself frames as open (a new `NoteKind::Alert` vs. reusing an
existing tier), and `Steered`'s "matching weight" isn't even well-defined
against the live deck, which renders it as a full user-turn row rather than
a note at all. Filed as #5748, referenced from the residue list below.

Closes #5504

## Residue

- #5748 — `note_kind` still mutes `Error` and `SteeringWithheld` to the same
  glyph as a budget tick; needs a `NoteKind` tier decision.

## Summary by Sourcery

Align transcript note classification with the live deck by treating `GateBoard` as a verdict and clarify the protocol’s event and wire-format guidance.

Bug Fixes:
- Classify `GateBoard` events as verdict notes in plain transcripts and exported views instead of muted miscellaneous notes.

Enhancements:
- Remove unreachable `note_kind` classifications and document the reachable event families to prevent future omissions.
- Correct protocol documentation for event layout, wire-format optional fields, integration-test coverage, and JCS ordering behavior.

Documentation:
- Update protocol guidance with the actual `AgentEvent` source file, integration-test sample workflow, frozen optional-field serialization set, and JCS compatibility details.

Tests:
- Add coverage for `GateBoard` verdict classification and direct classification of each reachable note family.

### DIFF
--- a/crates/stella-protocol/README.md
+++ b/crates/stella-protocol/README.md
@@ -105,22 +105,27 @@ This crate has no god files: no file exceeds the gate's 1500-line ratchet
 accepts no new entries. When a file here approaches the limit, split it before
 it crosses.
 
-[`src/event.rs`](src/event.rs) was grandfathered here for a long time and now
-sits just under the limit, so its split discipline still applies: new
-supporting vocabulary goes in a new module re-exported from
-[`src/lib.rs`](src/lib.rs) — the crate's precedent is
-[`src/ladder.rs`](src/ladder.rs), split out of `event.rs` when the ladder rung
-joined it (#1043), with the re-export keeping `stella_protocol::LadderSnapshot`
-at its old path — and a genuinely new `AgentEvent` case offsets its lines
-by extracting the case's supporting types, taking their inline round-trip
-tests with them.
+[`src/event/kind.rs`](src/event/kind.rs) — the `AgentEvent` enum itself, one
+arm per wire case — is the file that sits near the limit today.
+[`src/event.rs`](src/event.rs) is 284 lines, nowhere close. `kind.rs` is not
+in `scripts/file-size-baseline.txt`, so it is held to the strict 1500-line
+ceiling rather than a grandfathered one; a first-time crossing gets no
+baseline entry (see AGENTS.md § "God files"). Split discipline therefore
+applies **there**: a genuinely new `AgentEvent` case lands its arm in
+`kind.rs`, but any supporting type it needs goes in a new module re-exported
+from [`src/event.rs`](src/event.rs) — the crate's precedent is
+[`src/ladder.rs`](src/ladder.rs), split out when the ladder rung joined the
+vocabulary (#1043), with the re-export keeping
+`stella_protocol::LadderSnapshot` at its old path — and a case's inline
+round-trip tests travel with its supporting type, not with `kind.rs`.
 
 ## Layout
 
 | File | What it holds |
 |---|---|
 | [`src/lib.rs`](src/lib.rs) | The crate's flat re-export surface, and the statement of the one-directional wire-compatibility rule. Read it first. |
-| [`src/event.rs`](src/event.rs) | `AgentEvent` and its supporting types — the stream-json vocabulary: every wire case plus the tagless `Unknown` fallback. Open it to add or change anything a renderer, the journal, or a receipt consumes. |
+| [`src/event.rs`](src/event.rs) | The module's re-export surface and its `serde(default = "...")` helper functions. Open it to add or change anything a renderer, the journal, or a receipt consumes. |
+| [`src/event/kind.rs`](src/event/kind.rs) | `AgentEvent` itself — the stream-json vocabulary: every wire case plus the tagless `Unknown` fallback. The file near the 1500-line ceiling; see "God files" above. |
 | [`src/journal.rs`](src/journal.rs) | `StampedEvent` / `stamped_line` — one *line* of the event stream: an `AgentEvent` plus the optional wall-clock `ts` its sink stamps at the write boundary (#2111). Not a field of the enum: a stamp is a fact about a write, the same event reaches more than one sink, and this crate owns no clock. |
 | [`src/context_event.rs`](src/context_event.rs) | `CompiledContextFrameBuilt` — the compiled-frame identity the step manifest carries, and its golden JCS vector. The `LifecycleEventEnvelope` channel that used to live here was deleted unwired (#3135). |
 | [`src/completion.rs`](src/completion.rs) | `CompletionRequest` / `CompletionResult` / `CompletionUsage`, `GenerationParams`, `FinishReason`. The one envelope every provider adapter translates to and from. |
@@ -224,11 +229,14 @@ equality, and a mismatch is harvested as `AgentEvent::SpeculationDiscarded`.
   fields crossed (`text` carried `delta`, `text_delta` carried `text`); the
   serde aliases keep those parsing, and raw-JSONL readers must stay bilingual
   the same way.
-- **The golden JCS vector in `context_event.rs` is meant to break.** Renaming,
-  retyping, adding, or reordering a field of `CompiledContextFrameBuilt` breaks
-  the golden line, because those canonical bytes are the preimage
-  `stella_core::context_record::hash` builds `record_hash` from. The dev-dep
-  pins the same canonicalizer crate *and version* core hashes with.
+- **The golden JCS vector in `context_event.rs` is meant to break — on three
+  of the four edits, not all four.** Renaming, retyping, or adding a
+  field of `CompiledContextFrameBuilt` breaks the golden line, because those
+  canonical bytes are the preimage `stella_core::context_record::hash` builds
+  `record_hash` from. **Reordering does not**: RFC 8785 (JCS) sorts object
+  keys lexicographically, so two structurally-identical values serialize to
+  the same bytes regardless of field declaration order. The dev-dep pins the
+  same canonicalizer crate *and version* core hashes with.
 - **`ToolSchema::read_only` defaults to `false`** — the safe direction. An MCP
   tool or an older schema that doesn't know about the field is treated as
   mutating and never speculated on.
@@ -281,12 +289,26 @@ equality, and a mismatch is harvested as `AgentEvent::SpeculationDiscarded`.
   drops line/column from the resulting error message. Both are the price of the
   forward-compat fallback (#672), and both narrow the "a known tag with a bad
   body stays loud" guarantee slightly.
-- **`AgentEvent::UsageIncomplete::retries` serializes as `null` when absent.**
-  It is the one `Option` on the event vocabulary without
-  `skip_serializing_if`, so it costs a key on every incomplete-usage line while
-  its neighbours (`Pr::number`, `Pr::ci`, `ContextFrameRef::uri`, …) omit
-  theirs. Changing it now would be a wire change, so it is documented rather
-  than fixed; do not copy the pattern onto a new field.
+- **Six `Option` fields on the event vocabulary serialize `null` when
+  absent, because they skip `skip_serializing_if`** — six, not the single
+  field this section once named alone. `AgentEvent::UsageIncomplete::retries`
+  was first; the pattern was then copied onto `BudgetTick`'s `limit_usd`,
+  `session_spent_usd`, `session_limit_usd` and `deadline_remaining_ms`, and
+  onto `FileChange::diff`. `budget_tick` is one of the highest-volume events
+  in `stella-events.jsonl`, so four of the six now cost a `null` key on every
+  tick after every paid call. Their neighbours (`Pr::number`, `Pr::ci`,
+  `ContextFrameRef::uri`, …) omit theirs.
+
+  Fixing this is not a one-line change on three of the six:
+  `UsageIncomplete::retries`, `BudgetTick::limit_usd` and `FileChange::diff`
+  carry no `#[serde(default)]` either, so adding `skip_serializing_if` to one
+  of those three *alone* would make a newly-written event with the field
+  absent fail to decode — the key would simply be missing, not present-as-
+  `null`. Any fix has to add both attributes together, on all three, and
+  re-verify every round-trip test in this crate against the smaller wire
+  shape. That is a wire-format change to a shape already written into
+  replayable journals, so all six stay documented rather than fixed here;
+  **do not copy the pattern onto a new field.**
 
 ## Testing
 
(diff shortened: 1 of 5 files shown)


macanderson:
## DoD audit for #5504 — 4 of 5 ticked

I checked each box against the diff and the tree at head `f7980c6`, including re-deriving the numbers the README now asserts. One box is left open on purpose.

**Ticked, with the evidence:**

- **"`GateBoard` classifies as a verdict on the plain surface, with a witness."** — `transcript_build::note_kind` now carries `AgentEvent::GateBoard { .. }` in the `NoteKind::Verdict` or-arm. It genuinely reaches that function: `textline::event_line` returns `Some(gate_board(board))` for the variant, and `RunBuilder::note` early-returns only on `None`. Witness: `gate_board_classifies_as_a_verdict_not_a_muted_other` in `crates/stella-tui/src/transcript_build/tests.rs`; `stella_protocol::GateBoard` derives `Default`, so the fixture builds.
- **"`note_kind` is on the silent-wildcard checklist."** — `crates/stella-protocol/src/event/tags.rs`'s `type_tag` doc now names `transcript_build::note_kind` beside `deck::event_intensity` / `deck::status_from_event`.
- **"The four README/comment claims are true as written."** — I verified all four rather than taking them on faith:
  1. *`skip_serializing_if` count* — parsed every `Option` field in `src/event/kind.rs` with its full attribute block. Exactly six lack `skip_serializing_if`, and they are precisely the six the README names: `BudgetTick::{limit_usd, session_spent_usd, session_limit_usd, deadline_remaining_ms}`, `UsageIncomplete::retries`, `FileChange::diff`. The sub-claim that three of them (`limit_usd`, `retries`, `diff`) also lack `#[serde(default)]` checks out too.
  2. *`tests/` directory* — it exists and holds exactly the four files the README now names: `wire_contract.rs`, `wire_contract/samples.rs`, `journal_stamp_wire.rs`, `text_event_wire.rs`. The new recipe step 4 cites `every_known_tag_has_a_sample`, which is a real `#[test]` in `wire_contract.rs`.
  3. *God-file pointer* — `src/event.rs` is 284 lines and `src/event/kind.rs` is 1493, and `grep`ping `scripts/file-size-baseline.txt` for this crate returns nothing, so `kind.rs` is indeed held to the strict 1500-line ceiling. Both numbers in the README are right.
  4. *JCS reordering* — corrected in both places `context_event.rs` asserted it (the module doc and the `golden_jcs_vector_for_compiled_context_frame_built` comment) and in the README. RFC 8785 sorts object keys, so "reordering does not break the golden line, renaming/retyping/adding does" is the accurate statement.
- **"`skip_serializing_if` added to the five `Option` fields that lack it, or the README describes the real frozen set and why."** — The second branch, and the "why" holds up: adding `skip_serializing_if` alone to any of the three fields that carry no `#[serde(default)]` would make a newly-written event with the field absent fail to decode, so the fix is a wire change rather than a one-liner. That reasoning is verified against the attributes, not just asserted.

I also checked the four arms this PR *removes* (`Stage`, `Proof`, `BudgetDenied`, `RetriesExhausted`) — all four sit in `event_line`'s `=> None` arm, so `note()` returns before `note_kind` ever sees them. The removal drops dead code, not coverage.

**Not ticked:**

- **"`cargo test -p stella-protocol -p stella-tui` green."** — A claim about CI, and the `fmt + clippy + test` job on head `f7980c6` is still `in_progress`. Nothing is missing from the diff. Tick it once that job reports `success`, and `dod` should then pass.

---

macanderson:
## DoD audit against #5504 — a seventh `null`-writing field breaks two items

I verified each of #5504's five DoD items against the diff by reading the code at the head. Items 1 and 2 land. Items 3 and 4 both fail on the same off-by-one, so I have not ticked the checklist and `dod / dod` stays red.

**Item 1 — `GateBoard` classifies as a verdict, with a witness — SATISFIED.** `transcript_build.rs`'s `note_kind` groups `AgentEvent::GateBoard { .. }` with `Verdict | GoalVerdict | ScopeReview | HunkReview | TaskUpdate => NoteKind::Verdict`. I checked the path is genuinely reachable rather than assuming it: `RunBuilder::push` ends in `other => self.note(other)`, and `note()` returns early only when `textline::event_line` yields `None` — `event_line` has `GateBoard { board } => Some(gate_board(board))`, so the event does reach `note_kind`. The plain surface consumes this exact builder (`stella-cli/src/plain/transcript.rs` imports `RunBuilder`). The witness `gate_board_classifies_as_a_verdict_not_a_muted_other` fails on base, where `_ => NoteKind::Other` catches it.

**Item 2 — `note_kind` on the silent-wildcard checklist — SATISFIED.** `event/tags.rs`'s `type_tag` module doc now names it under **Silent**, which is the correct half — `note_kind` does still carry `_ => NoteKind::Other`. Nit only: it is appended to the existing `deck::event_intensity` bullet, so that bullet's leading `stella-tui` scope now covers two unrelated matchers.

### Items 3 and 4 — NOT SATISFIED

The README states, emphatically, that **six** `Option` fields on the event vocabulary serialize `null` when absent — "six, not the single field this section once named alone" — and enumerates `UsageIncomplete::retries`, `BudgetTick::{limit_usd, session_spent_usd, session_limit_usd, deadline_remaining_ms}`, and `FileChange::diff`.

Those six are right. I confirmed each in `event/kind.rs`, and confirmed every other `Option` there carries the skip (39 `Option<` occurrences, 33 skips, delta 6).

**There is a seventh: `SubAgentPhase::Started::budget_usd`**, in `crates/stella-protocol/src/subagent_event.rs`. It is declared `budget_usd: Option<f64>` with **no serde attribute at all** — no `skip_serializing_if`, no `default` — while `effort` in the *same variant* and `Finished::reason` both carry `#[serde(default, skip_serializing_if = "Option::is_none")]`. So it is not a deliberate exception; it is the one its two siblings got and it did not.

It is live, not theoretical:

- `SubAgentPhase` is Serialize-derived and is the payload of `AgentEvent::SubAgent { phase }` (`kind.rs`).
- It is in the committed wire contract — `docs/wire/agentevent.schema.json`, `$defs/SubAgentPhase/oneOf/0`, with `budget_usd` absent from `required`.
- It has a producer at `crates/stella-core/src/subagent.rs`.
- Its own doc comment says `None` means "no axis anywhere bounds it" — the ordinary unbounded-child case, so this fires routinely, not rarely.

Every `sub_agent`/`started` line for a child with no budget cap therefore writes `"budget_usd": null` — precisely the defect the paragraph claims to have enumerated exhaustively.

Why this breaks both items:

- **Item 3** ("the four README/comment claims are true as written") — the other three are now true and I verified each: the `tests/` directory exists with the four files described and both `sample_events()` and `every_known_tag_has_a_sample` resolve; `event.rs` is 284 lines and `event/kind.rs` is 1493 with no `stella-protocol` entry in `scripts/file-size-baseline.txt`, so the strict-1500 claim holds; and the false JCS "reorder" claim is gone from all three places (`context_event.rs` module doc, its test comment, and the README). Claim 1 is the one still false.
- **Item 4** — the PR took the "or the README describes the real frozen set and why" branch and added no attributes (the diff touches no serde attribute; it is five files, none of them `subagent_event.rs`). That branch requires the description to be *real*, and it is off by one in the direction that hides a live `null`-writing field. The "and why" half inherits the error: the README argues "not a one-line change on three of the six" — that trio is correct, I checked those three lack `serde(default)` while the other three have it — but `budget_usd` also lacks `serde(default)`, so the true statement is *four of the seven*, and the enumerated remedy ("add both attributes together, on all three") would leave a field behind.

For the record: Sourcery's assessment table marks this objective ✅. On this specific point that verdict does not hold up against the diff — hence spelling out the evidence above rather than just asserting it.

### What would unblock this

Either is fine and both are small:

1. **Add the pair to `budget_usd`** — `#[serde(default, skip_serializing_if = "Option::is_none")]`, matching `effort` two fields down — and drop it from the frozen set, leaving the README's six correct as written. Note this changes the emitted wire bytes for `sub_agent`/`started`, so `docs/wire/` regenerates.
2. **Correct the count to seven**, add `budget_usd` to the enumeration, and fix "three of the six" to "four of the seven".

**Item 5 — `cargo test -p stella-protocol -p stella-tui` green — depends on CI.** I compiled nothing locally. One thing worth CI's eye: the diff *removes* four `note_kind` arms (`Stage`, `Proof`, `BudgetDenied`, `RetriesExhausted`). The justification holds on reading — `textline.rs`'s `event_line` groups all four into its `=> None` arm, so `note()` returns before `note_kind` sees them, and no test in `stella-tui` or `stella-cli/src/plain` asserts a `NoteKind::Stage` classification. That is a reason to expect green, not evidence of it.

### Residue, not DoD items

- **`NoteKind::Stage` now has no production producer anywhere.** Removing that arm makes the tier unreachable from the only `Run` builder in the tree, while `stella-transcript` still renders it (`model.rs` glyph `—`/label `"stage"`, `grid.rs` `Color::Violet`) for a value nothing emits. The arm was already unreachable, so this reveals rather than creates it — but the Residue list names only #5748.
- **The README now pins a line count in prose** ("`src/event.rs` is 284 lines"). True today, and exactly the "a count in prose is a promise to update it, and nobody does" shape CLAUDE.md forbids. No gate catches it — `check-line-citations.py` matches `file:line` citations, not lengths.

---

PR #5756, opened 2026-09-03
fix(stella-observatory): release the live stream's connection permit at its head

### LINKED ISSUE #5491: The live dashboard stream holds a one-shot connection permit until the tab closes
## The problem

`stella observe` bounds one-shot connections with a 64-permit semaphore, and the
live SSE stream is supposed to be exempt. It is not: the permit is taken in the
accept loop before the route is known, and released only when the connection
handler returns — which for a live stream is when the browser tab closes.

The code says the opposite in three places. `MAX_LIVE_CONNECTIONS`
(`crates/stella-observatory/src/lib.rs`) says "the live SSE stream is **not**
counted here ... Streams carry their own separate, smaller budget in the `live`
module. Do not merge the two." The comment above the `acquire_owned` call in
`serve` says "no long-lived stream — every response is one shot and closes — so
a permit is held for a single request and a semaphore cannot starve a protocol
against itself." `live.rs`'s module doc says the same. All three are false as
written.

## Where

- `crates/stella-observatory/src/lib.rs`, `serve` — takes the permit before
  `tokio::spawn`, and the spawned task drops it only after `handle` returns.
- `crates/stella-observatory/src/lib.rs`, `respond_to_head` — returns
  `live::serve_stream(...)` for `LIVE_ROUTE`, which does not return until the
  stream ends.
- `crates/stella-observatory/src/live.rs`, `MAX_LIVE_STREAMS` (8) — the separate
  budget that was supposed to be the only one a stream consumes.

## Failure scenario

Eight dashboard tabs each hold `/api/v1/live` open — the documented maximum, and
an ordinary number for someone watching several projects. Each permanently holds
one of the 64 one-shot permits, so the pool is silently 56.

Worse, the accept loop is `accept().await` then `acquire_owned().await` then
`spawn`, in that order and unbuffered. Once the pool is exhausted the loop blocks
*at the acquire*, holding a socket it has already accepted, and stops calling
`accept()` at all. New connections — including a fresh page load, and including a
request that would have completed instantly — are not served until some existing
connection finishes. Exhausting it needs 64 concurrent holders, which a mix of
live tabs and connections that open without sending a request head (each holding
a permit for the full `READ_TIMEOUT`) reaches without anything hostile.

## The fix

Release the permit at the moment the route is known to be the live one, rather
than never. Thread the `OwnedSemaphorePermit` from `serve` through `handle` into
`respond_to_head`, and `drop(permit)` immediately before the
`return live::serve_stream(...)`. Holding it across the head read is correct and
worth keeping: that read is bounded by `READ_TIMEOUT`, so it is exactly the
one-shot work the semaphore exists to bound.

## Definition of done

- [x] A live stream holds no one-shot permit for longer than its head read.
- [x] The three comments cited above are true as written, or corrected.
- [x] A witness test that fails before the change and passes after. The cheap
      design: hold `MAX_LIVE_STREAMS` live streams open, open enough head-less
      connections to bring the total holders to `MAX_LIVE_CONNECTIONS`, then
      issue one ordinary `GET /api/meta` and assert it answers within a bound.
      It hangs today. This costs ~65 sockets and a `READ_TIMEOUT` hold, so
      making the two caps injectable for the test is a reasonable alternative to
      a slow test.
- [x] `cargo test -p stella-observatory` green — carried by the workspace
      `cargo test` in CI's required `fmt + clippy + test` job, which concluded
      `success` at 2026-09-03T17:01:27Z on PR #5756's head
      `6e0d339b53942eecaa10a7a9e77fb2c03adc5581`
      (run `33779702029`, job `100730418543`). That job is the authoritative
      gate per AGENTS.md — the crate's own suite runs inside `test --workspace`.

## Confidence

CONFIRMED. Read directly: `serve`'s acquire-before-spawn, the spawned task's
`drop(permit)` after `handle(...).await`, and `respond_to_head`'s early
`return live::serve_stream(...)`. The three comments asserting the opposite are
quoted above.

macanderson:
## What & why

`stella observe`'s one-shot semaphore bounded ordinary requests at 64
concurrent permits, and the live SSE stream (`/api/v1/live`) was documented
as exempt from it. It was not: `serve` took the permit before the route was
known and only released it when the spawned task's `handle(...).await`
returned — for a live stream, that is when the browser tab closes. Eight
open dashboard tabs (the documented max for `live.rs`'s own separate
budget) silently held eight of the 64 one-shot slots forever, and once the
pool was exhausted the accept loop blocked at `acquire_owned` with a socket
already accepted, so it stopped calling `accept()` at all — a fresh page
load, or any ordinary request, could hang behind it.

`serve`, `handle` and `respond_to_head` now thread the
`OwnedSemaphorePermit` through instead of holding it in `serve`'s spawned
task. `respond_to_head` drops it the instant a request is known to route to
`/api/v1/live`, before the socket is ever handed to `live::serve_stream` —
so the permit only ever covers the one-shot part of the connection (the
bounded head read, and for every other route, writing the single answer).
Corrected the three comments the issue names (`MAX_LIVE_CONNECTIONS`'s doc,
the `acquire_owned` comment in `serve`, and `live.rs`'s module doc), which
all described the fixed invariant as already holding.

## The witness

`crates/stella-observatory/src/tests/http_surface.rs`'s
`a_live_stream_releases_its_permit_before_its_long_life_begins` opens one
live stream against a one-shot pool shrunk to a single permit
(`serve_with_capacity`, a private wrapper `serve` calls with
`MAX_LIVE_CONNECTIONS`), then issues an ordinary `GET /api/meta` behind it
and asserts the response arrives within 2 seconds. Real `MAX_LIVE_CONNECTIONS`/
`MAX_LIVE_STREAMS` untouched — the pool is shrunk only for the test, per the
issue's own suggested design, so the proof costs one live connection and one
ordinary request instead of ~65 sockets.

Verified the fail→pass flip by hand: temporarily reverted `lib.rs` to
`origin/main` and re-added only the `serve_with_capacity` plumbing (no
permit threading) — the test panics at the 2-second timeout with "an
ordinary request hung behind the live stream's permit". With the real fix
restored it passes in under a millisecond (`finished in 0.00s`). All 154
existing `stella-observatory` lib tests plus the 4 `tests/live_stream.rs`
integration tests still pass.

Closes #5491

## The gate

- [x] `cargo fmt --check`
- [x] `cargo clippy --workspace --all-targets -- -D warnings` — CI, green
- [x] `cargo test --workspace` — CI, green
- [x] Docs updated where behavior/flags changed (doc comments corrected on `MAX_LIVE_CONNECTIONS`, `serve`, `live.rs`'s module doc)

## Nothing left behind

## Ground-rule check

- [ ] N/A — no new cross-boundary types

`serve_with_capacity` is a plain refactor of `serve`'s body with the pool
size pulled out as a parameter; `serve` itself is unchanged in behavior and
still always passes `MAX_LIVE_CONNECTIONS`. Only a test calls the new
function with a smaller number.

## Summary by Sourcery

Prevent live SSE connections from exhausting the observatory's one-shot request capacity by releasing their permit before the stream begins.

Bug Fixes:
- Release the one-shot connection permit as soon as a request is identified as a long-lived live stream, preventing open SSE connections from blocking ordinary requests and the accept loop.

Enhancements:
- Keep live streams governed by their dedicated connection budget while limiting the general semaphore to each connection's one-shot request work.

Documentation:
- Correct documentation to describe the live stream permit-release behavior and its separate connection budget.

Tests:
- Add a regression test using a single-permit pool to verify ordinary requests remain responsive while a live stream stays open.

## DoD re-check (#5491)

#5491's last open box was `cargo test -p stella-observatory` green, and the
evidence for it did not exist when `dod` last ran at 16:39Z — the required job
was still running. It has since concluded.

The required `fmt + clippy + test` job on this head
(`6e0d339b53942eecaa10a7a9e77fb2c03adc5581`, run `33779702029`, job
`100730418543`) concluded **`success`** at 2026-09-03T17:01:27Z. That job
carries `cargo clippy -D warnings`, `cargo test --workspace` — which is where
this crate's own suite runs — `cargo doc -D warnings` and the release smoke
build, and it is the authoritative gate per AGENTS.md rather than a local run.
That is what ticks the two gate boxes above, which the original description had
left to CI, and the matching box on #5491.

All four of #5491's DoD boxes are now ticked. This edit re-asks the `dod` check,
which fires on `pull_request: edited`; a workflow re-run was not available
(`403 Resource not accessible by integration`).

The remaining unticked boxes above are the CLA line (handled by the bot, not
per commit) and an explicit N/A, both unchanged.

### DIFF
--- a/crates/stella-observatory/src/lib.rs
+++ b/crates/stella-observatory/src/lib.rs
@@ -112,12 +112,14 @@ const READ_TIMEOUT: Duration = Duration::from_secs(10);
 /// deadlock a protocol against itself. 64 is far past what one dashboard page
 /// opens while keeping the blocking-pool fan-out bounded.
 ///
-/// That reasoning is exactly why the live SSE stream is **not** counted here.
-/// An SSE connection is held open for as long as a tab is, so a handful of
-/// them on this semaphore would hold permits indefinitely and starve every
-/// one-shot route behind them — the page would hang while its own stream was
-/// perfectly healthy. Streams carry their own separate, smaller budget in
-/// the `live` module. Do not merge the two.
+/// That reasoning is exactly why the live SSE stream must not stay on this
+/// semaphore past its request head. An SSE connection is held open for as
+/// long as a tab is, so a handful of them holding a permit for that whole
+/// life would starve every one-shot route behind them — the page would hang
+/// while its own stream was perfectly healthy. `respond_to_head` drops the
+/// permit the moment a request routes to `/api/v1/live`, before handing the
+/// socket to a long-lived stream; streams carry their own separate, smaller
+/// budget in the `live` module from that point on. Do not merge the two.
 const MAX_LIVE_CONNECTIONS: usize = 64;
 
 /// Sent on every response, making the dashboard's zero-external-reference
@@ -682,12 +684,34 @@ pub async fn serve(
     plugins: Arc<dyn PluginContributions>,
     on_ready: impl FnOnce(SocketAddr),
 ) -> Result<(), ServeError> {
+    serve_with_capacity(
+        workspace_root,
+        port,
+        plugins,
+        on_ready,
+        MAX_LIVE_CONNECTIONS,
+    )
+    .await
+}
+
+/// [`serve`]'s body, with the one-shot pool's size pulled out as a
+/// parameter. [`serve`] always passes [`MAX_LIVE_CONNECTIONS`]; a test asks
+/// for a single permit instead, because proving a live stream does not hold
+/// its permit for its whole life needs one stream and one ordinary request
+/// past it, not sixty-four sockets.
+async fn serve_with_capacity(
+    workspace_root: PathBuf,
+    port: u16,
+    plugins: Arc<dyn PluginContributions>,
+    on_ready: impl FnOnce(SocketAddr),
+    max_live_connections: usize,
+) -> Result<(), ServeError> {
     let listener = TcpListener::bind(("127.0.0.1", port))
         .await
         .map_err(|source| ServeError::Bind { port, source })?;
     let addr = listener.local_addr().map_err(ServeError::Accept)?;
     on_ready(addr);
-    let connections = Arc::new(tokio::sync::Semaphore::new(MAX_LIVE_CONNECTIONS));
+    let connections = Arc::new(tokio::sync::Semaphore::new(max_live_connections));
     let mut backoff = AcceptBackoff::new();
     loop {
         let stream = match listener.accept().await {
@@ -714,12 +738,18 @@ pub async fn serve(
                 AcceptAction::Fatal => return Err(ServeError::Accept(err)),
             },
         };
-        // Bounded concurrency. Unlike `stella-serve`, nothing here is a
-        // long-lived stream — every response is one shot and closes — so a
-        // permit is held for a single request and a semaphore cannot starve a
-        // protocol against itself. A connection that cannot get a permit waits
-        // rather than being refused: the dashboard is a local tool, and a brief
-        // queue is a better answer than an error the page has to render.
+        // Bounded concurrency. A permit is held for the one-shot portion of a
+        // connection — reading the request head and, for every route but the
+        // live stream, writing that route's single answer — so bounding
+        // connections cannot starve a protocol against itself the way it
+        // would in `stella-serve`. The one route that is not one-shot,
+        // `/api/v1/live`, has `respond_to_head` drop its permit the moment
+        // that route is known, before the socket is handed to a long-lived
+        // stream (see `live`'s module doc for the separate, smaller budget
+        // that governs a stream from that point on). A connection that
+        // cannot get a permit waits rather than being refused: the dashboard
+        // is a local tool, and a brief queue is a better answer than an
+        // error the page has to render.
         let permit = Arc::clone(&connections)
             .acquire_owned()
             .await
@@ -728,9 +758,11 @@ pub async fn serve(
         let plugins = Arc::clone(&plugins);
         tokio::spawn(async move {
             // Per-connection errors (bad request line, client hangup) only
-            // affect that connection; the accept loop keeps serving.
-            let _ = handle(stream, &root, plugins).await;
-            drop(permit);
+            // affect that connection; the accept loop keeps serving. `permit`
+            // travels into `handle` and is dropped there — either by
+            // `respond_to_head` early, for a live stream, or on this task's
+            // return for every other route.
+            let _ = handle(stream, &root, plugins, permit).await;
         });
     }
 }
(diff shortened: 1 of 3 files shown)


macanderson:
## DoD audit against #5491 — no gaps; three of four items verified against the diff

Automated PR sweep. The `dod` check is red only because #5491's boxes are unticked, not because anything is missing. I read the diff rather than the PR description, so this is evidence for whoever ticks them.

**1. "A live stream holds no one-shot permit for longer than its head read."** — Verified. `OwnedSemaphorePermit` is threaded `serve` → `handle` → `respond_to_head`, and `drop(permit)` sits immediately before `return live::serve_stream(...)`, inside the `LIVE_ROUTE` branch. `serve`'s spawned task no longer holds it — it moves into `handle`, so the head-read timeout path drops it with the frame. The permit now spans exactly the bounded `READ_TIMEOUT` head read, which is what the issue asked for.

**2. "The three comments cited above are true as written, or corrected."** — Verified, all three, and they are the three the issue names: `MAX_LIVE_CONNECTIONS`'s doc, the `acquire_owned` comment in `serve`, and `live.rs`'s module doc. Each now describes release-at-head rather than exemption. Worth noting the old `live.rs` doc quoted the `lib.rs` comment verbatim — that quotation is gone, so the two cannot drift back into agreeing about something false.

**3. "A witness test that fails before the change and passes after."** — Verified as discriminating **by construction**, which is stronger than taking the PR's word for the flip. With the bug, the live stream holds the lone permit for the life of its socket; the accept loop then blocks at `acquire_owned` holding an already-accepted connection, so the `GET /api/meta` cannot be answered and the 2-second bounded read fails the assertion. With the fix the permit is free before `serve_stream` is ever called. There is no third outcome, and the bounded read is what turns the bug into a failing assertion rather than a hung test.

On the issue's suggested design: it proposed ~65 sockets and a `READ_TIMEOUT` hold, and then said *"making the two caps injectable for the test is a reasonable alternative to a slow test."* `serve_with_capacity` is that alternative — `serve` still always passes `MAX_LIVE_CONNECTIONS`, and only the test asks for 1. Both real caps are untouched. This satisfies the item as written.

**4. `cargo test -p stella-observatory` green** — not yet confirmable here. `fmt + clippy + test` is still running on `6e0d339`. This is the only item I would hold the boxes for; the other three do not depend on it.

No gaps found. Once the workspace job lands green, all four are satisfiable on the evidence above.

---

macanderson:
## DoD audit for #5491 — 3 of 4 ticked

I read the diff against the issue's checklist and ticked only what the diff demonstrates. One box is left open on purpose.

**Ticked, with the code that satisfies each:**

- **"A live stream holds no one-shot permit for longer than its head read."** — `serve` no longer drops the permit in its spawned task; it passes it into `handle`, which passes it into `respond_to_head`, which calls `drop(permit)` inside the `route_of(path) == LIVE_ROUTE` branch, before `return live::serve_stream(...)`. The head-read timeout arm in `handle` drops it by falling out of scope. The permit therefore spans the bounded head read and nothing more.
- **"The three comments cited above are true as written, or corrected."** — All three are rewritten and I checked each against the new control flow: `MAX_LIVE_CONNECTIONS`'s doc ("`respond_to_head` drops the permit the moment a request routes to `/api/v1/live`"), the `acquire_owned` comment in `serve` ("a permit is held for the one-shot portion of a connection"), and `live.rs`'s module doc ("before the socket ever reaches `serve_stream`"). Each now describes what the code does. The live branch is still gated on `head_complete && host_is_local(&head) && method == "GET"`, so the drop cannot be reached by a truncated or rebound head.
- **"A witness test that fails before the change and passes after."** — `a_live_stream_releases_its_permit_before_its_long_life_begins` in `crates/stella-observatory/src/tests/http_surface.rs`, driving the new private `serve_with_capacity` with a one-permit pool. It takes the issue's own suggested design (injectable cap rather than ~65 sockets). On the old code the live connection holds the lone permit for its socket's life, the accept loop wedges at `acquire_owned`, and the bounded `GET /api/meta` read trips the 2-second assertion.

**Not ticked:**

- **"`cargo test -p stella-observatory` green."** — This is a claim about CI, and the `fmt + clippy + test` job on head `6e0d339` is still `in_progress`. Nothing is missing from the diff; the box just cannot be ticked from a run that has not reported. Tick it once that job reports `success`, and `dod` should then pass.

---

Issue #5754, opened 2026-09-03
The budget guard drops a spend figure it cannot use without telling anyone


macanderson:
`BudgetGuard::record_spend` and `BudgetGuard::reseed_session_spend` now treat
a non-finite or negative dollar figure as zero
(`spendable_usd` in `crates/stella-core/src/budget.rs`). That keeps the money
meter usable, and it is silent: nothing tells anyone the guard was handed a
figure it could not use.

Silence is wrong in both directions. If a producer starts sending `NaN`, spend
quietly stops being counted and the cap it guards stops binding on that
money. And an operator reading the HUD sees a total that does not match the
provider's bill with no hint why.

`stella-fleet` answers the same question by rewriting the attempt's summary
(`crates/stella-fleet/src/fleet.rs`, the `invalid_cost` branch), because it
has a record to write into. The engine's guard has no such channel: it is a
pure synchronous value type with no I/O, so the report has to ride an existing
signal.

## The shape of the fix

Pick one and wire it end to end — a counter nothing reads is unwired code:

- Count rejections on the guard and carry the count on
  `AgentEvent::BudgetTick`, which `BudgetGuard::tick_event` already mints.
  That means a `stella-protocol` field, a row in
  `crates/stella-protocol/src/event/consumers.rs`, and a surface that shows
  it (the deck's budget HUD is the obvious one).
- Or emit a `stella-diag` diagnostic from the caller, keeping the guard pure:
  the caller checks the figure before it hands it over, and the guard's own
  clamp stays the backstop.

## How to verify

`crates/stella-core/src/budget.rs` holds
`an_unreadable_cost_contributes_nothing_instead_of_wedging_the_gate`. Add a
sibling asserting the report: after `record_spend(f64::NAN)`, whatever signal
is chosen must name one rejected figure, and a clean session must name none.

## Definition of done

- [ ] A rejected spend figure reaches a surface a person can see, through a
      signal with a declared consumer.
- [ ] A test asserts the report appears for a rejected figure and stays absent
      for a clean session.
- [ ] `cargo test -p stella-core` green, plus whichever crate carries the
      surface.

Issue #5753, opened 2026-09-03
A caller's query vector cannot say which embedder made it, so the context plane has to ignore it


macanderson:
A host can hand `ContextStore::recall` a query vector through
`ContextQuery::embedding`. That field is a bare `Option<Vec<f32>>` and names
no embedder, so the store cannot tell a vector made under its own fingerprint
from one made under another at the same width. The store now ignores the
field and embeds the query text itself, which is safe but throws away a real
saving for a host that already holds a matching vector — and for an HTTP
embedding backend that saving is a network round trip per recall.

`ContextError::FingerprintMismatch` exists for this check and is constructed
nowhere in the workspace. Until a supplied vector can state its fingerprint,
there is nothing to compare, so the variant stays unreachable.

## Where the code is

- `crates/stella-context/src/retrieval.rs` — `recall_scoped_excluding` step 1
  embeds the query text; the doc block above it ("Query fields not honored
  yet") states why `embedding` is turned down.
- `crates/stella-context/src/error.rs` — `ContextError::FingerprintMismatch`,
  with its two `query` / `candidate` fields.
- `ContextQuery` itself lives in the `contextgraph-types` crate, in
  <https://github.com/macanderson/context-graph-protocol>. Stella pins it with
  an exact `=` requirement in the workspace `Cargo.toml`, so this needs a
  release there first.

## The shape of the fix

1. In the CGP repository, let a query carry the fingerprint its vector was
   made under — a sibling field beside `embedding`, or a small struct that
   holds both. Bump and release `contextgraph-types`.
2. In `stella-context`, accept a supplied vector only when the stated
   fingerprint equals `self.fingerprint().id()` and the length matches
   `dims`. On a mismatch return
   `ContextError::FingerprintMismatch { query, candidate }` rather than
   silently embedding instead, so a misconfigured host learns why its vector
   was refused.
3. Keep the embed-the-text path as the answer when no fingerprint is stated.

## How to verify

`crates/stella-context/src/retrieval/tests/recall.rs` holds
`a_caller_supplied_query_vector_is_ignored_and_the_query_text_is_embedded`,
which counts embed batches through a wrapping embedder. Extend that helper:
a matching fingerprint must make the count zero, and a mismatched one must
return `FingerprintMismatch`.

## Definition of done

- [ ] A query can state the fingerprint its embedding was made under, on the
      CGP wire type, released and pinned here.
- [ ] `recall_scoped_excluding` accepts a matching vector and refuses a
[...]

PR #5757, opened 2026-09-03
fix: five smaller correctness findings across context, model, core and mcp

### LINKED ISSUE #5511: Five smaller correctness findings across context, model, core and mcp
Five smaller findings from a workspace-wide correctness audit, each too small for
its own issue and each with a concrete failure scenario.

## 1. `stella-context` — a caller's query embedding is checked by length only

`crates/stella-context/src/retrieval.rs`, `ContextStore::recall_scoped_excluding`
accepts `q.embedding` when `v.len() == dims` and otherwise embeds the query
itself. `ContextQuery::embedding` is a bare `Option<Vec<f32>>` with no
model identity, so length is the only check.

`ContextError::FingerprintMismatch` exists and its doc says this is prevented —
"retrieval never mixes fingerprints; a stored vector under a stale fingerprint is
invisible, never silently compared against a fresh query vector". That variant is
constructed nowhere in the workspace. The *stored*-vector half of the invariant
is enforced (the SQL filters on `fingerprint`); the query half is not.

**Failure scenario.** `HashEmbedder::with_revision` shows two embedders can share
`dims` and differ in fingerprint — the crate's own tests pair them. A caller
(an external CGP host, or future in-tree code) supplies a 256-float vector made
under `hash-ngram@1/256/l2` while the store's fingerprint is `@2`. It is accepted
verbatim and scored against every vector stored under `@2`. The cosines are
meaningless across incompatible feature spaces, and they drive ranking, the
coverage gate, MMR seed selection, and semantic-evidence admission — silently.

No in-tree caller populates the field today (every call site passes `None`), so
this is reachable only through the public API and the CGP wire protocol, which is
the boundary the crate treats as untrusted.

**Fix.** Either always embed the query text locally, or extend the wire contract
so a supplied embedding declares its fingerprint, compare against
`self.fingerprint().id()`, and finally construct the error variant that exists
for this.

## 2. `stella-model` — an HTTP date with a huge year overflows

`crates/stella-model/src/http.rs`, `imf_fixdate_epoch_secs` parses the year as a
bare `i64` with no digit-count check, against a grammar that specifies exactly
four digits, and feeds it to `days_from_civil`'s unchecked
`era * 146_097 + day_of_era - 719_468`.

**Failure scenario.** A `Retry-After: Mon, 01 Mar 99999999999999999 00:00:00 GMT`
header parses (the year fits in `i64`), and `era * 146_097` overflows `i64`. In a
debug or test build that panics — a crash on server-controlled response data,
which workspace invariant #5 forbids. In release, overflow checks are off, so it
wraps to a nonsense `retry_after_ms` and corrupts backoff timing.

The header comes from whatever endpoint a provider spec names, including the
`local` provider (documented as an arbitrary user-run server) and any
OpenAI-compatible gateway from `settings.json`. The existing malformed-date test
table covers 1969, 1970, 2000, 2016 and 2100 — nothing outside a sane range.

**Fix.** Reject a year that is not exactly four ASCII digits before parsing, or
use `checked_mul`/`checked_add` and return `None`. The function's own doc already
says a header it cannot read exactly costs only the fallback to computed backoff.
**Witness.** Add that header to the malformed-date table and assert
`retry_after_ms_at(header, 0) == None`.

## 3. `stella-core` — the budget guard accepts a non-finite spend

`crates/stella-core/src/budget.rs`, `record_spend` does `+=` with no validation,
and `check_axis` returns early only when `spent_usd <= limit_usd`.

`NaN <= limit` is false, so one `record_spend(f64::NAN)` makes every later
evaluation on that axis report `Warn`/`AbortTurn` for the rest of the session —
only `begin_turn` (turn axis) or an explicit reseed clears it. `settle_child`
calls `record_spend` with a child's total, so it propagates.

Every current producer traces back to `PricedModel::cost_usd`, pure
multiplication of finite counts by finite rates, so no live path produces `NaN`
today — hence PLAUSIBLE. But this is the money meter that decides whether a turn
aborts, and it accepts an arbitrary `f64` from any caller. The sibling finding in
`stella-fleet` (#5498) is the same shape with an untrusted producer.

**Fix.** Treat a non-finite or negative `cost_usd` as zero (or a typed error) and
report the anomaly rather than accumulating it.
**Witness.** `record_spend(f64::NAN)` then `record_spend(0.1)` under a $1.00
enforced limit; assert `evaluate()` is `Continue`.

## 4. `stella-mcp` — a revoked OAuth grant is re-tried on every call with no backoff

`crates/stella-mcp/src/client.rs`, `call_tool` and `read_only_request`. A
`RequestOutcome::Protocol` — which is where `McpError::Auth` lands, since
`is_connection_death` matches only `Transport`/`Closed` — updates connection
health not at all: neither `mark_call_success` nor `note_call_failure` runs.

**Failure scenario.** A user revokes stella's OAuth grant for an HTTP MCP server
mid-session. Every later call's `bearer()` attempts a refresh, which fails with
`invalid_grant`, giving `McpError::Auth` on the `Protocol` arm. If the agent
retries the tool a few times in one turn, each call re-POSTs to the third-party
token endpoint with no caching and no backoff — the connect-time 401 case has
`AuthProbeCache` for exactly this and the live-session case has nothing.
Meanwhile `health()` keeps reporting the server `Live`, contradicting its
documented meaning ("would a `tools/call` issued right now be expected to reach
this server and come back?").

**Fix.** Record an `Auth` failure toward health without tearing down the
transport, and extend the probe cache's TTL suppression to post-connect 401s.
Do not arm a reconnect for every `Protocol` error — a bad-arguments `-32602`
should not.

## 5. `stella-mcp` — a docker registry entry can get a duplicated `run`

`crates/stella-mcp/src/registry.rs`, `runtime_invocation` decides whether a
package already supplies `run` by checking `a.value.as_deref() == Some("run")`,
while the argument-rendering loop resolves values through `arg_value`, which
falls back to `Argument::default`.

**Failure scenario.** A `server.json` declaring
`runtimeArguments: [{"type":"positional","default":"run"}]` — a legitimate way to
author a suggested rather than fixed value — passes the duplicate check (its
`value` is `None`), so `["run","--rm","-i"]` is prepended, and the loop then
emits `run` again from the default. The command becomes
`docker run --rm -i run <identifier>`, so docker reads the second `run` as the
image name and the server fails to start.

No test exercises a docker or OCI package; the registry fixture is npm-only.

**Fix.** Use `arg_value(a) == Some("run")` in the duplicate check, mirroring the
loop's own resolution.
**Witness.** An OCI `RegistryPackage` with no runtime hint and that argument;
assert `package_to_transport` yields exactly one `run`.

## Definition of done

- [x] Each finding fixed with a witness, or downgraded with evidence.
- [x] `cargo test -p <crate>` green for each touched crate.

## Confidence

1, 2, 4 and 5 CONFIRMED as mechanisms by the audits that read them, with the
reachability caveat stated inline for each. 3 is PLAUSIBLE: the mechanism is
confirmed, no live producer of a non-finite value was found.

---

macanderson:
Five findings from a workspace-wide correctness audit, each fixed with its own
witness. The design was posted on the issue before implementation and is
followed, with one departure named below.

## 1. `stella-context` — a query vector is only checked by length

`recall_scoped_excluding` accepted `q.embedding` whenever `v.len() == dims`
and scored it against every stored vector. `ContextQuery::embedding` is a bare
`Option<Vec<f32>>` and names no embedder, so length is not identity, and two
embedders can share `dims` while differing in fingerprint — the crate's own
tests pair two (`HashEmbedder::with_revision`). Cosines across incompatible
feature spaces mean nothing, and they drive ranking, the coverage gate, MMR
seeding and semantic-evidence admission.

The store now always embeds the query text with its own embedder. `embedding`
joins `kinds` and `representation_preferences` in the doc block that lists
what a `ContextQuery` field does not buy.

The issue offered a second option — extend the wire contract so a supplied
vector declares its fingerprint. `ContextQuery` lives in the external
`contextgraph-types` crate, pinned here with an exact `=` requirement, so that
option is a change in another repository. It is filed as #5753, which is also
the only path that can ever make `ContextError::FingerprintMismatch`
constructible; that variant stays unreachable and the issue says so.

**Witness mechanism.** `a_caller_supplied_query_vector_is_ignored_and_the_query_text_is_embedded`
wraps `HashEmbedder` in a counting embedder and recalls with a right-length
`embedding: Some(..)`. On the old code the vector is taken and `embed` is
called zero times; the assertion is one. Every in-tree call site already
passed `None`, so no live behaviour changes.

## 2. `stella-model` — an HTTP date with a huge year overflows

`imf_fixdate_epoch_secs` read each field with `str::parse`, so
`Retry-After: Mon, 01 Mar 99999999999999999 00:00:00 GMT` reached
`days_from_civil`, where `era * 146_097` overflows `i64` — a panic in a debug
or test build off server-chosen data (invariant #5), a wrapped
`retry_after_ms` in release. Every numeric field is now read through
`fixed_width_digits` at the width RFC 9110 §5.6.7's grammar fixes: day 2,
year 4, hour/minute/second 2. Four digits bound the arithmetic, and the same
check drops spellings `str::parse` takes and the grammar does not (`+5`,
`006`).

**Witness mechanism.** The absurd-year header joins the malformed-date table
(old code: panic in a debug test build, so the test cannot pass at all), plus
a `parse_retry_after_ms` case proving the public header path, plus the
narrower width cases.

## 3. `stella-core` — the budget guard accepts a non-finite spend

`record_spend` did `+=` with no check. `NaN` poisons the axis and
`NaN <= limit` reads false, so one bad figure made every later evaluation
report `Warn`/`AbortTurn` for the session; a negative figure shrank the total
so a later turn passed a cap real spend had crossed. `spendable_usd` now
clamps a non-finite or negative amount to zero, on `record_spend` and on
`reseed_session_spend` — the second door, which takes journaled spend the
process reads back off disk. `carve` was already safe (`f64::min`/`max` ignore
`NaN`).

The issue also asked to "report the anomaly". A signal nothing reads is
unwired code, and the report has to ride an existing event with a declared
consumer, so it is filed as #5754 rather than half-built here.

**Witness mechanism.** `an_unreadable_cost_contributes_nothing_instead_of_wedging_the_gate`
records `f64::NAN` then `0.10` under a $1.00 enforced limit and asserts
`Continue`; on the old code the first call poisons the axis and the second
returns `AbortTurn`. Two siblings cover the reseed door and a child settling
through `settle_child`.

`stella-fleet`'s `dispatch_claimed` is the exemplar — the same clamp one layer
up, where the producer is a worker process.

## 4. `stella-mcp` — a revoked OAuth grant is retried with no backoff

`McpError::Auth` is not `is_connection_death`, so it landed on the `Protocol`
arm, which updated health in neither direction. A user revoking the grant
mid-session therefore had every call fail while `health()` kept reporting
`Live` — the state whose documented meaning is "a `tools/call` issued right
now would come back" — and each attempt re-POSTed to a third-party token
endpoint with no cache and no backoff.

`Connection::note_auth_failure` now counts the failure, sets
`HealthState::AuthRequired`, and arms a hold-off window, leaving the transport
in place: the connection works and only the credential is stale, so a redial
would carry the same dead credential. While the window stands the next call
fails fast naming `stella mcp login <server>`. The window is
`suppress::AUTH_PROBE_TTL`, read from the constant the connect-time 401 cache
already uses rather than copied, and it is cleared by any request that returns
or any completed handshake — both are authenticated round trips. Every other
`Protocol` error still arms nothing, so a `-32602` is unchanged.

New logic lands in a sibling submodule, `src/client/authz.rs`: `client.rs` was
five lines from the 1500-line ceiling once the change was written, and
AGENTS.md § "God files" says a file approaching the limit gets split rather
than grown.

**Witness mechanism.**
`a_revoked_grant_marks_the_server_auth_required_and_holds_off_the_next_call`
scripts a `tools/call` answering `McpError::Auth`, then asserts
`AuthRequired`, one call failure, no pending redial, and — through the
transport's request log — that the second call never went out. On the old code
health reads `Live` and the log holds two requests. Two siblings cover
`resources/list` taking the same door and a `-32602` leaving the server
`Live`.

## 5. `stella-mcp` — a docker registry entry can get a duplicated `run`

The duplicate check read `a.value` while the rendering loop resolves through
`arg_value`, which falls back to `default`. An entry declaring
`runtimeArguments: [{"type":"positional","default":"run"}]` — a legal way to
suggest rather than fix a value — passed the check, got `run --rm -i`
prepended, and then emitted `run` again: `docker run --rm -i run <image>`,
where docker reads the second `run` as the image name. Both sides now ask
`arg_value`.

**Witness mechanism.** `a_suggested_docker_run_is_not_duplicated_by_the_prepended_lead`
builds that OCI package and asserts exactly one `run`; the old code yields
two. A sibling pins the two unchanged shapes (no `run` named, `run` fixed as a
literal `value`). No test exercised a docker or OCI package before this.

## Notes

No new dependencies. Tests deleted: None.

Residue filed: #5753 (a query vector cannot state its embedder), #5754 (a
rejected spend figure is dropped silently).

Closes #5511

## Summary by Sourcery

Fix correctness issues across context retrieval, HTTP date parsing, budget accounting, MCP authorization handling, and Docker command generation.

Bug Fixes:
- Ensure context retrieval always embeds query text with the store’s own embedder instead of trusting an unidentifiable caller-supplied vector.
- Reject malformed or oversized HTTP dates before calendar arithmetic can overflow.
- Ignore negative and non-finite spending values so budget totals and enforcement remain valid.
- Track mid-session MCP authorization failures, report authentication-required health, and suppress repeated calls during a login hold-off window.
- Prevent Docker runtime arguments from duplicating the `run` command when it is supplied as a default.

Enhancements:
- Split MCP authorization handling into a dedicated client submodule while preserving existing behavior for other protocol errors.

Documentation:
- Document the context query embedding contract, MCP authorization hold-off behavior, and the new authorization module.

Tests:
- Add regression coverage for incompatible query vectors, invalid HTTP dates, unreadable budget amounts, revoked MCP grants, and Docker runtime argument rendering.


macanderson:
## DoD audit for #5511 — 1 of 2 ticked

#5511's checklist has two boxes. I ticked the first; the second is left open on purpose.

**Ticked — "Each finding fixed with a witness, or downgraded with evidence."** All five findings are fixed, each with a witness that fails on the old code:

1. **`stella-context`** — `recall_scoped_excluding` no longer reads `q.embedding` at all; step 1 unconditionally embeds `query_text`/`goal` with `self.embedder()`. Witness: `a_caller_supplied_query_vector_is_ignored_and_the_query_text_is_embedded` (`retrieval/tests/recall.rs`) wraps `HashEmbedder` in a `CountingEmbedder` and asserts one embed batch while passing a right-length `embedding`; on the old code the vector is reused and the count is zero.
2. **`stella-model`** — the new `fixed_width_digits` in `http.rs` gates day/year/hour/minute/second at the widths RFC 9110 §5.6.7 fixes, so a year is at most four digits and `days_from_civil`'s `era * 146_097` cannot leave `i64`. Witnesses: the absurd-year header added to `a_malformed_or_obsolete_date_yields_no_hint_rather_than_a_wrong_one`'s table (exactly what the issue asked for), plus `an_absurd_year_on_the_header_map_path_drops_the_hint_rather_than_overflowing` covering the public `parse_retry_after_ms` path.
3. **`stella-core`** — `spendable_usd` clamps a non-finite or negative figure to zero, applied on both doors onto the accumulator (`record_spend` and `reseed_session_spend`). Witness: `an_unreadable_cost_contributes_nothing_instead_of_wedging_the_gate` is the issue's own suggested test verbatim, with siblings for the reseed door and for `settle_child`.
4. **`stella-mcp` (revoked grant)** — `Connection::note_auth_failure` counts the failure, sets `HealthState::AuthRequired`, and arms `auth_blocked_until` from `suppress::AUTH_PROBE_TTL` (read from the constant, not copied), leaving the transport up; `authz::hold_off` answers from the window; `mark_call_success` and `mark_connected` both clear it. Only `err.is_auth_error()` routes there, so a `-32602` still arms nothing. Witnesses: the three tests in the new `src/client/authz.rs`, including `a_rejected_request_still_leaves_the_server_live` for the negative case.
5. **`stella-mcp` (docker `run`)** — the duplicate check now asks `arg_value(a)`, matching the rendering loop's own resolution. Witness: `a_suggested_docker_run_is_not_duplicated_by_the_prepended_lead` is the issue's suggested test, with a sibling pinning the two unchanged shapes.

Two sub-asks in the issue's "Fix" paragraphs are deferred rather than done, and both are the issue's own alternative branch plus a filed handoff, so I read them as "downgraded with evidence" rather than as gaps: constructing `ContextError::FingerprintMismatch` needs a wire-type change in `context-graph-protocol` (#5753), and "report the anomaly" on the budget clamp needs a signal with a declared consumer per invariant #10 (#5754). Both issues exist and read as executable handoffs. Flagging them here so the deferral is visible rather than assumed.

**Not ticked — "`cargo test -p <crate>` green for each touched crate."** This is a claim about CI, and the `fmt + clippy + test` job on head `2a4a89e` is still `in_progress`. Nothing is missing from the diff; the box just cannot be ticked from a run that has not reported. Tick it once that job reports `success`, and `dod` should then pass.

---

macanderson:
**Answering Sourcery's one ❌** — the budget row: *"…while reporting the
rejected anomaly."*

The clamp half is done on both doors onto the accumulator. The reporting half
is deliberately not in this PR, and the reason is a rule this repository
enforces rather than a shortcut.

`BudgetGuard` is a pure synchronous value type in `stella-core` with no I/O
(invariant #2), so it cannot report anything itself. The report has to ride an
existing signal, and invariant #10 says an emitted signal names its consumer:
a rejection counter with an accessor nothing calls is unwired code, which is
exactly what "Nothing left behind" exists to stop. Doing it properly means a
field on `AgentEvent::BudgetTick`, a row in
`crates/stella-protocol/src/event/consumers.rs`, and a surface that renders
it — a protocol change with its own witness, in a PR about one thing.

That is #5754, filed as a handoff with the two candidate designs (the
`BudgetTick` field, or a `stella-diag` diagnostic emitted by the caller so the
guard stays pure), the file paths, and its own definition of done. The sibling
finding in `stella-fleet` (#5498) could report because it had a record to
write into — `dispatch_claimed` rewrites the attempt summary; the engine's
guard has no such channel today.

The silent-drop hazard Sourcery is pointing at is real, and the issue says so
in those words. What this PR buys now is that a wedged money meter can no
longer abort every turn of a session.

PR #5758, opened 2026-09-03
fix(stella-model): let two sessions in one process name two Vertex projects

### LINKED ISSUE #5501: Two sessions in one process cannot use two different Vertex projects
## The problem

`stella-runtime` promises it reads no ambient environment, enforced by
`tests/no_ambient_reads.rs`. The test scans this crate's own `src/` only, and the
Vertex provider reaches the process environment one call below it — so a
multi-tenant host, which is the audience the crate's README names, cannot give
two sessions two different GCP projects.

## Where

- `crates/stella-runtime/src/parts.rs`, `build_provider` — calls
  `stella_model::factory::build_provider`.
- `crates/stella-model/src/factory.rs`, the `Dialect::Vertex` arm — calls
  `VertexAddressing::resolve()`.
- `crates/stella-model/src/credential.rs`, `VertexAddressing::resolve` — reads
  `VERTEX_PROJECT_ID` / `GOOGLE_CLOUD_PROJECT` / `VERTEX_LOCATION` from
  `std::env`, unconditionally, with no path through `ProviderParts.aux`.
- Contrast `Dialect::Bedrock` in the same match, which calls
  `BedrockCredentials::resolve_with(aux)` — preferring host-injected values and
  falling back to the environment only for names `aux` does not carry.
  `VertexAddressing` exposes only `resolve()` and a `pub(crate) resolve_from`;
  there is no `resolve_with`.

## Failure scenario

A host built on `stella-runtime` — "N concurrent sessions with N different
workspace roots and N different trust postures in one process", per the crate's
own README — assembles two `RuntimeSpec`s for `Dialect::Vertex` with two
tenants' GCP projects. Neither reaches the adapter, because there is no field
for it and nothing reads `aux` on the Vertex path. Both sessions resolve project
and location from the one process-global environment, so both talk to whichever
project the host process happens to have set. That is wrong billing scope, wrong
region, and wrong data residency, silently.

The crate's README already names this class of hazard as a known blind spot of
the textual scan ("a constructor one call below that reads the environment for
us") and claims the registry is the only constructor needing the structural
guarantee. Vertex is a second instance, and nothing tests it —
`rg -n "Vertex" crates/stella-runtime/tests/` finds nothing.

## The fix

Add `VertexAddressing::resolve_with(aux: &AuxCredentials)` mirroring
`BedrockCredentials::resolve_with`, call it from the Vertex arm of
`factory.rs`, and thread project and location through `aux` from `stella-cli`'s
credential chain. Spans `stella-model`, `stella-runtime` and `stella-cli`, which
is why this is filed rather than fixed.

## Definition of done

- [x] A host can give two sessions two different Vertex projects in one process.
- [x] Witness: build two `RuntimeSpec`s with `Dialect::Vertex` and distinct
      `aux` project ids in one process, and assert the two providers address
      different projects. Today both read the same environment.
- [x] `tests/no_ambient_reads.rs` covers the constructors one call below, or
      says plainly which it cannot see and why.
- [x] `cargo test -p stella-runtime -p stella-model` green.

### Evidence (#5758)

- `VertexAddressing::resolve_with(aux)` in
  `crates/stella-model/src/credential.rs`, called from the `Dialect::Vertex`
  arm of `crates/stella-model/src/factory.rs`, with `stella-cli`'s
  `provider_aux` walking its chain for the three names.
- `two_specs_can_name_two_vertex_projects_in_one_process`
  (`crates/stella-runtime/tests/runtime_build.rs`) builds the two specs;
  `two_vertex_sessions_in_one_process_address_two_projects`
  (`crates/stella-model/src/factory.rs`) is the wire-level assertion that the
  two adapters address different projects, since a `Box<dyn Provider>` reports
  nothing about its addressing.
- `tests/no_ambient_reads.rs` gains `std::env::temp_dir`, `home_dir` and
  `dirs::`, and `panel_host.rs`'s reaping test takes a `TempDir` so the
  widened guard needs no exemption. The scan still reads only this crate's
  `src/`, which its own header states.
- `fmt + clippy + test` (the workspace suite, which includes both crates) is
  green on #5758:
  https://github.com/macanderson/stella/actions/runs/33781133876

The third item of the "Related, same crate" section — `CandidateFanouts::adopt`
dropping removal failures — is filed as #5759 rather than fixed here.

## Confidence

CONFIRMED by the audit, which traced the call from `parts.rs` through
`factory.rs` into `credential.rs` and contrasted it with the Bedrock arm's
`aux`-aware sibling.

## Related, same crate, filed here rather than separately

**`no_ambient_reads.rs`'s needle list has holes.** `FORBIDDEN` catches
`std::env::var`/`var_os`, `current_dir` and `set_current_dir`, but not
`std::env::temp_dir`, `home_dir` or `dirs::`. One occurrence exists today
(`panel_host.rs`, inside `#[cfg(test)]`), so the crate's real guarantee holds —
but the guard would not catch a future production use either. Adding the three
needles is a one-line fix and worth doing with this.

**`CandidateFanouts::adopt` drops removal failures.** In
`crates/stella-runtime/src/wrapper/candidate_fanout.rs`, `adopt` takes the live
set via `take_live()` and never restores it, then does
`let _ = self.workspaces.remove(workspace).await` per candidate. Its comment says
"A removal that fails is left to `discard_all`'s contract rather than raised
here", but `discard_all` iterates `self.live()`, which no longer contains those
handles. A genuine removal failure — a held lock, a permission error, the cases
`CandidateFanoutError::NotRemoved` exists for — is reported to the plugin as a
successful discard, dropped from `live()` forever, and never retried. That is the
"a host whose user finds out with `df`" outcome the module's docs warn against,
with the named backstop not firing.

macanderson:
## What and why

The Vertex arm of `stella_model::factory::build_provider` called
`VertexAddressing::resolve()`, which reads `VERTEX_PROJECT_ID` /
`GOOGLE_CLOUD_PROJECT` / `VERTEX_LOCATION` straight from the process
environment — one call below `stella-runtime`'s `tests/no_ambient_reads.rs`,
which only scans that crate's own `src/`. The environment is the same for every
session a host assembles in one process, so a multi-tenant host could not give
two sessions two GCP projects: both talked to whichever project the host
happened to have exported, in the wrong region and the wrong data residency,
silently.

- `VertexAddressing::resolve_with(aux)` (new) mirrors
  `BedrockCredentials::resolve_with` — the exemplar the issue names, and the
  sibling in the same file. Host-resolved values win; the environment stays the
  fallback for anything absent, so a library-style caller that builds straight
  off `factory::build_provider` behaves exactly as before.
- The project is **one field with two spellings**, so `resolve_with` decides it
  per field rather than per name. A name-by-name fallback would let an exported
  `VERTEX_PROJECT_ID` outrank a host-supplied `GOOGLE_CLOUD_PROJECT` by winning
  the name order, and the two sessions would share a project again.
  `PROJECT_ENV_NAMES` / `LOCATION_ENV_NAME` / `AUX_ENV_NAMES` are one list read
  in both places, with a test pinning that they agree.
- `stella-cli`'s `config::aux_credentials::provider_aux` walks its chain
  (handoff → environment → `credentials.toml`) for those names, and
  `settable_aux_fields` lets `stella auth set vertex` store the project and the
  location — neither is a secret, so both are typed in the clear. Without that
  second half the credentials-file leg of the chain would be unreachable.
- The related items the issue filed alongside: `no_ambient_reads.rs` gains
  `std::env::temp_dir`, `home_dir` and `dirs::` as needles, and
  `panel_host.rs`'s reaping test takes a `TempDir` rather than a name under the
  process temp directory, so the widened guard needs no exemption.

`CandidateFanouts::adopt` (the issue's third related item) is untouched and
filed separately — see the comment on this PR.

## Witness mechanism

**`stella-model`, `factory.rs`:
`two_vertex_sessions_in_one_process_address_two_projects`.** Which project an
adapter addresses is visible only on the wire, so each of two adapters — built
in one process from two `AuxCredentials` sets — is pointed at a wiremock server
answering one project-scoped path per tenant and nothing else. Neither tenant
id exists in the process environment, so on the old code
`VertexAddressing::resolve()` cannot produce either path: it either errors on a
missing project or addresses some other one, and both calls fail. The test
asserts each adapter got its own tenant's response body.

**`stella-runtime`, `runtime_build.rs`:
`two_specs_can_name_two_vertex_projects_in_one_process`.** Two `RuntimeSpec`s
with `Dialect::Vertex` and distinct `aux` project ids assemble through
`stella_runtime::build_provider`. On the old code the aux set is ignored and
`resolve()` fails with `VertexProjectMissing` in any environment with no Vertex
project — which is CI. Its doc comment says so, and it returns early when a
project variable *is* exported rather than claiming a witness it cannot deliver
there; the wire-level distinctness assertion is the `stella-model` one above.

**`credential/tests/secondary_credential.rs`:** three unit tests over
`resolve_with` — two aux sets resolving to two projects and two locations, the
either-spelling rule, and `AUX_ENV_NAMES` matching what resolution reads.

**`stella-cli`:**
`vertex_resolves_its_project_and_location_through_the_same_chain` (the
credentials-file leg with the environment cleared), and
`every_settable_field_is_a_name_the_chain_actually_reads` rewritten to ask
`provider_aux` itself rather than compare against a hand-copied name list.

## Exemplar

`BedrockCredentials::resolve_with` / `AUX_ENV_NAMES` in the same file — same
seam, same fallback posture, same public-const contract with the host.

## Definition of done

- [x] A host can give two sessions two different Vertex projects in one process.
- [x] Witness: two specs with distinct `aux` project ids in one process, and an
      assertion that the two providers address different projects (on the wire,
      in `stella-model`).
- [x] `no_ambient_reads.rs` covers three more constructors one call below, and
      names what each one reads.
- [ ] CI green for `stella-runtime` and `stella-model` (this run).

## Tests deleted

None.

Closes #5501

## Summary by Sourcery

Enable per-session Vertex project and location resolution while retaining environment-based fallback behavior.

New Features:
- Allow Vertex sessions to resolve project and location values from per-session auxiliary credentials, enabling multiple projects in one process.
- Support storing and resolving Vertex project and location settings through the CLI credentials chain.

Bug Fixes:
- Prevent process-global Vertex environment settings from causing multi-tenant sessions to use the wrong project or region.

Enhancements:
- Define shared Vertex addressing environment-name contracts and preserve environment variables as fallbacks for library callers.
- Strengthen ambient process-state checks and update temporary-directory test isolation.

Documentation:
- Document credential-chain support for Vertex project and location values, including storage through `stella auth set vertex`.

Tests:
- Add coverage proving separate Vertex sessions can address distinct projects and locations, including wire-level provider verification.
- Add tests covering project aliases, credential-chain consistency, and Vertex auxiliary-field resolution.

### DIFF
--- a/crates/stella-cli/src/config/aux_credentials.rs
+++ b/crates/stella-cli/src/config/aux_credentials.rs
@@ -16,17 +16,19 @@
 //!
 //! `config.rs` owns the one-key chain — CLI flag, env var (+aliases), the
 //! anonymous-FD handoff, `~/.stella/credentials.toml`, an interactive prompt.
-//! Bedrock is the only provider today that needs more than one value, and it
-//! needs four: an access key id (which *is* the one key, `AWS_ACCESS_KEY_ID`),
-//! a secret access key, an optional session token, and a region.
+//! Two providers need more than one value. Bedrock needs four: an access key
+//! id (which *is* the one key, `AWS_ACCESS_KEY_ID`), a secret access key, an
+//! optional session token, and a region. Vertex needs three: a bearer token
+//! (the one key, `VERTEX_ACCESS_TOKEN`), a GCP project, and a location.
 //!
-//! Those three extra values used to be read straight from the process
-//! environment inside the adapter's own construction, which made Bedrock work
-//! in exactly one situation — a shell with the standard AWS variables exported
-//! — and nowhere else. `stella auth set bedrock` stored the access key id and
-//! then the run failed on the missing secret; a sealed benchmark process, whose
+//! Those extra values used to be read straight from the process environment
+//! inside the adapter's own construction, which made each provider work in
+//! exactly one situation — a shell with the right variables exported — and
+//! nowhere else. `stella auth set bedrock` stored the access key id and then
+//! the run failed on the missing secret; a sealed benchmark process, whose
 //! whole design keeps provider secrets out of the environment, could not reach
-//! Bedrock at all (#1301).
+//! Bedrock at all (#1301). The environment is also one value per process, so a
+//! host assembling two sessions could not give them two GCP projects.
 //!
 //! This module walks the same chain for each of those names and hands the
 //! result to the adapter as an [`AuxCredentials`] set. One vocabulary
@@ -48,7 +50,9 @@
 //! to avoid — and, under a sealed launcher, would reach back out to host state
 //! the launcher spent its whole design excluding. Explicit credentials only.
 
-use stella_model::credential::{AuxCredentials, BedrockCredentials, CredentialsFile};
+use stella_model::credential::{
+    AuxCredentials, BedrockCredentials, CredentialsFile, VertexAddressing,
+};
 
 use super::{Dialect, ProviderConfig};
 
@@ -56,11 +60,11 @@ use super::{Dialect, ProviderConfig};
 /// order the one-key chain uses: the trusted handoff, then the environment,
 /// then the credentials file.
 ///
-/// Returns an empty set for every dialect that authenticates with one key —
-/// which is every dialect but Bedrock. An empty set is not an error and never
-/// blocks construction: a provider with nothing extra to resolve has nothing
-/// extra to fail on, and the Bedrock adapter reports its own named error for a
-/// value that resolved nowhere.
+/// Returns an empty set for every dialect that needs nothing beyond one key —
+/// which is every dialect but Bedrock and Vertex. An empty set is not an error
+/// and never blocks construction: a provider with nothing extra to resolve has
+/// nothing extra to fail on, and each of those two adapters reports its own
+/// named error for a value that resolved nowhere.
 ///
 /// `credentials_file` is passed in rather than loaded here so this shares the
 /// caller's decision about whether the filesystem store may be read at all —
@@ -73,6 +77,7 @@ pub(crate) fn provider_aux(
 ) -> AuxCredentials {
     let names: &[&str] = match provider.dialect {
         Dialect::Bedrock => BedrockCredentials::AUX_ENV_NAMES,
+        Dialect::Vertex => VertexAddressing::AUX_ENV_NAMES,
         _ => return AuxCredentials::new(),
     };
 
@@ -153,12 +158,12 @@ pub(crate) struct AuxField {
 }
 
 /// The auxiliary values `stella auth set <provider>` offers to store, in
-/// prompt order. Empty for every provider that authenticates with one key.
+/// prompt order. Empty for every provider that needs nothing beyond one key.
 ///
-/// `AWS_DEFAULT_REGION` is deliberately absent even though the chain reads it:
-/// it exists as a fallback for shells that already export it, and offering
-/// *two* places to write the same setting is how a stored region and an
-/// effective region end up disagreeing.
+/// `AWS_DEFAULT_REGION` and `GOOGLE_CLOUD_PROJECT` are absent even though the
+/// chain reads both: each is a fallback for shells that already export it, and
+/// offering *two* places to write one setting is how a stored value and an
+/// effective value end up disagreeing.
 pub(crate) fn settable_aux_fields(provider: &ProviderConfig) -> &'static [AuxField] {
     match provider.dialect {
         Dialect::Bedrock => &[
@@ -178,6 +183,20 @@ pub(crate) fn settable_aux_fields(provider: &ProviderConfig) -> &'static [AuxFie
                 prompt: "AWS region (blank for us-east-1)",
             },
         ],
+        // Neither is a secret: a project id and a location are addressing, and
+        // both are published in a run manifest.
+        Dialect::Vertex => &[
+            AuxField {
+                name: "VERTEX_PROJECT_ID",
+                secret: false,
+                prompt: "GCP project id (required — Vertex scopes every request to one)",
+            },
+            AuxField {
+                name: "VERTEX_LOCATION",
+                secret: false,
+                prompt: "Vertex location (blank for global)",
+            },
+        ],
         _ => &[],
     }
 }
@@ -191,6 +210,14 @@ mod tests {
             .iter()
             .find(|p| p.id == "bedrock")
             .expect("bedrock is a built-in provider")
+            .clone()
+    }
+
+    fn vertex() -> ProviderConfig {
+        super::super::PROVIDERS
+            .iter()
+            .find(|p| p.id == "vertex")
+            .expect("vertex is a built-in provider")
             .clone()
     }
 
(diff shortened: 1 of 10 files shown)


macanderson:
Remaining work filed from this change:

- #5759 — `CandidateFanouts::adopt` reports a failed workspace removal as a successful discard (the third item in #5501's "Related, same crate" list; this PR does not touch it).
- #5760 — a Vertex token with no project id still auto-selects Vertex and then fails; `has_required_aux` has the Bedrock guard and no Vertex one. Left out here because it changes provider auto-detection, a different surface from the multi-tenancy fix.

macanderson:
Sourcery's third row — "Fix CandidateFanouts::adopt so failed workspace-removal operations remain tracked" — is ❌ and stays ❌ here on purpose. It is the third bullet of #5501's "Related, same crate" section, in a different subsystem (`crates/stella-runtime/src/wrapper/candidate_fanout.rs`) from the credential seam this PR changes, and it needs a decision about the plugin-facing contract in `doc:wrapper-socket` that does not belong in a provider-addressing diff. It is filed as a handoff in #5759, with the repro, the two acceptable outcomes, and a definition of done.

The other two rows are ✅.

macanderson:
DoD sweep against this diff at `078c02e`, item by item. Two of #5501's four boxes are established, one is not, and one is still waiting on CI.

**Established — safe to tick:**

1. *"A host can give two sessions two different Vertex projects in one process."* — `VertexAddressing::resolve_with` exists in `credential.rs`, and the `Dialect::Vertex` arm of `factory.rs` calls it with the per-spec `aux` instead of `resolve()`. The per-field project rule (`host_named_project` gating both `PROJECT_ENV_NAMES` spellings together) is what stops an exported `VERTEX_PROJECT_ID` outranking a host-supplied `GOOGLE_CLOUD_PROJECT`, so the two sessions genuinely cannot collapse back onto one project.

2. *"Witness … assert the two providers address different projects."* — satisfied across the pair, not one test: `two_specs_can_name_two_vertex_projects_in_one_process` (`stella-runtime/tests/runtime_build.rs`) is the two-`RuntimeSpec`s half, and `two_vertex_sessions_in_one_process_address_two_projects` (`stella-model/src/factory.rs`) is the addressing half, asserted on the wire against per-tenant wiremock paths. The split is the right call — `Box<dyn Provider>` exposes nothing about its addressing — and the runtime test's early return when a project variable is exported is honest, since CI exports neither.

**Not established — this one cannot be ticked as the diff stands:**

3. *"`tests/no_ambient_reads.rs` covers the constructors one call below, or says plainly which it cannot see and why."*

   Neither branch of that "or" holds yet. The scan is still `Path::new(env!("CARGO_MANIFEST_DIR")).join("src")` — `stella-runtime`'s own sources only — so it does not reach `stella-model`'s constructors, which is where `VertexAddressing::resolve()` sat and where the next such read would sit. And the file says nothing about that blind spot: its module doc explains why it lives in `tests/` rather than in-crate, and the new comment in `FORBIDDEN` explains the three added needles, but no line names what the scan cannot see.

   The three added needles (`std::env::temp_dir`, `home_dir`, `dirs::`) are real and worth having, but they answer #5501's *"Related, same crate"* bullet about the needle list's holes — a different finding from this box. Ticking item 3 on their strength would be ticking one item with another's evidence.

   Cheapest honest close is the "or" branch: a short paragraph in that module doc naming the limit — the scan is textual and crate-local, so a constructor one call below (`stella_model::factory::build_provider` and the `credential.rs` resolvers it calls) is invisible to it, and that is why those paths are covered by `resolve_with` witnesses in `stella-model` instead. That also happens to be the fact the crate README already asserts, which would give the claim a second home rather than leaving it only in prose elsewhere.

4. *"`cargo test -p stella-runtime -p stella-model` green."* — not yet answerable: `fmt + clippy + test` (run `33781133876`, job `100734976424`) was still `in_progress` at the time of this sweep. The workspace `cargo test` inside it is the evidence for this box once it concludes.

`dod` will stay red until 3 and 4 are both settled, so the PR is not blocked on anything else — the residue is properly tracked in #5759/#5760, and the Sourcery ❌ already has its standing-down answer above.

---

PR #5764, opened 2026-09-03
fix(stella-cli): stop the deck panicking on a memory edit, and hold the memories span for the session

### LINKED ISSUE #5508: Editing a memory from the Command Deck panics the session, and three more memory defects
## The problem

Four defects in memory handling and system-prompt assembly, filed together
because they all sit on the path between a user's `.stella/memories/*.md` files
and the bytes sent to a provider. The first kills the session; the rest break the
byte-stable-prefix contract or lose a memory silently.

## 1. Editing a memory from the Command Deck panics the session — CONFIRMED

`crates/stella-cli/src/memory_cmd.rs`, `edit_memory_text` builds its own runtime:

```rust
let runtime = tokio::runtime::Runtime::new()...;
runtime.block_on(async { context.upsert(...).await })...
```

It is called synchronously from
`crates/stella-cli/src/command_deck/driver_support.rs`'s `service_edit_memory`,
which is called from `crates/stella-cli/src/command_deck.rs` inside
`run_deck_session` — an `async fn`. (The call sits in a boolean chain whose
neighbouring arms are `.await`ed, so the enclosing context is unambiguous.)
`Runtime::block_on` from a thread already driving a runtime panics with "Cannot
start a runtime from within a runtime."

**Failure scenario.** Open the interactive deck (`stella`), press `e` on a
memory row in the transcript. The process panics and the session dies. This is
the default interactive shell and an ordinary keypress.

Nothing exercises `edit_memory_text` from an async context; the crate's only
other `Runtime::new()` uses are `#[cfg(test)]`.

**Fix.** Make `edit_memory_text` async and `.await` it from `service_edit_memory`
(already inside async), giving the sync CLI entry point (`run_memory_edit`, from
`main.rs::run`) its own `block_on` wrapper. Adding `.await` at the two call sites
in `command_deck.rs` changes those lines rather than adding any, so the god-file
ceiling is not touched.

**Witness.** A `#[tokio::test]` over a temp workspace calling the edit path and
asserting it returns `Ok`. It panics today. Note the crate's tests can touch the
real `~/.stella` unless the home is redirected — set `STELLA_HOME`, not just
`STELLA_DATA_DIR`.

## 2. `/model` and `/agent` re-read memories mid-session — CONFIRMED

`crates/stella-cli/src/agent/prompt.rs`, `append_workspace_memories` does a fresh
`read_dir` plus `read_to_string` on every call. `build_system_prompt` is
documented — in `prompt.rs`, and in the crate README — as loading memories once
per session, because "a memory saved mid-session deliberately does not appear
until the next session; hot-injecting it would invalidate the cached prefix on
every save."

`crates/stella-cli/src/command_deck/session_override.rs`'s `refresh_prompts`
calls `build_system_prompt` again, from `apply_model_override` (the `/model`
picker) and unconditionally from `assume_agent` (the `/agent` picker — even when
the assumed agent declares no model, so nothing about the provider changes).

**Failure scenario.** A session opens with a memory reading "lesson A". The file
is edited on disk mid-session. Back in the deck, `/agent some-agent` with no
`model:` field rebuilds `messages[0]`, and the memories section now reads
"lesson B" — same provider, same model, prefix bytes changed, cache dropped, and
the written guarantee broken.

`refresh_prompts` gets *rules* right: it is handed the original resolved rules
rather than re-deriving them. Memories go through the full assembler, which reads
disk. That asymmetry reads as an oversight, and `session_override.rs`'s module
doc does not mention memories drifting.

**Fix.** Snapshot the assembled memories span once at session open and splice the
frozen bytes back in, letting `refresh_prompts` regenerate only the
session-environment span that legitimately changes on a model switch.

**Witness.** Build the prompt, mutate `.stella/memories/` on disk, rebuild, and
assert the memories span is byte-identical.

## 3. An unreadable memory vanishes with no diagnostic — CONFIRMED

Same function: `let Ok(body) = std::fs::read_to_string(file) else { continue; };`.
A memory file that is not valid UTF-8, is a directory, or is unreadable is
skipped silently. Every other omission in that same function is reported —
suppression failures, and the budget-drop count — and the function's own design
note says the omission must be stated in the prompt "rather than left for the
model to not notice". A user whose memory picked up invalid bytes loses it from
every future prompt with no indication.

**Fix.** Report the skip the way the budget drop is reported, naming the file.

`crates/stella-cli/src/memory/skill_files.rs`'s `read_skill_files` has the
identical `if let Ok(content)` shape and deserves the same treatment; it is lower
severity because skills ride the volatile channel rather than the cached prefix.

## 4. A memory file may be a symlink pointing anywhere — PLAUSIBLE

`append_workspace_memories` collects `*.md` from `read_dir` (which lists
symlinks) and reads them with `read_to_string` (which follows them). There is no
canonicalisation or root-containment check anywhere in `memory.rs`,
`agent/prompt.rs`, `memory_cmd.rs`, `memory_index.rs` or `memory_compact.rs` —
`rg "symlink|canonicalize|read_link"` over them returns nothing.

`.stella/memories/` is not inside the `.stella/private/` tier and is not
gitignored by the generated ignore file, so a cloned repository can ship
`.stella/memories/x.md` as a symlink to any readable path. On the victim's next
session its content is pasted into the system prompt and sent to the configured
provider.

Marked PLAUSIBLE rather than confirmed because it depends on whether a repository
ships a `.stella/memories/` directory at all, which is a convention question this
audit did not settle.

**Fix.** Refuse a memory entry whose canonical path escapes the workspace root,
matching the containment the file tools already enforce.

## Definition of done

- [x] Editing a memory from the deck does not panic, with a witness.
- [x] The memories span is byte-identical across a mid-session `/model` or
      `/agent`, with a witness.
- [x] An unreadable memory is named in the prompt's omission note.
- [x] A memory symlink cannot read outside the workspace root.
- [x] `cargo test -p stella-cli` green.

### Evidence (#5764)

Verified against the diff on head `a7b5de8`, item by item:

1. **No panic on a deck memory edit.** `edit_memory_text` moved to the new
   `crates/stella-cli/src/memory_cmd/edit.rs` and is now `async`;
   `driver_support.rs`'s `service_edit_memory` is `async` and awaits it, and
   `command_deck.rs`'s two call sites gained `.await` (line changes, not
   additions, so the god-file ceiling is untouched as the issue asked). The one
   remaining `block_on` sits in the sync CLI verb `run_memory_edit`. Witness:
   `edit::tests::a_memory_edits_from_inside_a_running_runtime`, a
   `#[tokio::test]` — it fails against the old shape by construction, since
   `Runtime::block_on` panics on a thread already driving a runtime.
2. **Byte-identical memories span across a rebuild.** `append_workspace_memories`
   became `workspace_memories_span` returning the bytes, wrapped by
   `session_workspace_memories`, which latches the span per workspace root in a
   process-lifetime `LazyLock<Mutex<HashMap<PathBuf, Arc<str>>>>` (poison
   recovered, since it caches bytes already read). Witness:
   `a_rebuilt_prompt_carries_the_memories_the_session_opened_with` — asserts
   `opened == rebuilt` after editing *and* adding a memory on disk between the
   two assemblies, plus that neither mid-session string appears.
   `session_override.rs`'s module doc now states the latch.
3. **An unreadable memory is named.** `read_to_string` errors now push the file
   stem onto `unreadable` and emit `(these .stella/memories/ files could not be
   read and were omitted: …)` beside the existing budget-drop note. Witness:
   `an_unreadable_memory_is_named_in_the_prompt`, using a genuine invalid-UTF-8
   file (`[0xff, 0xfe, 0xfd]`), asserting both that the readable sibling still
   ships and that the note names `corrupt`.
4. **Symlink containment.** New `resolves_inside` canonicalises both root and
   entry and requires `starts_with`; a path that cannot be canonicalised is
   *not* contained, so a broken link fails closed. Refused entries are named in
   their own note. Witness: `a_memory_symlinked_outside_the_workspace_is_refused`
   (`#[cfg(unix)]`) — asserts the outside content is absent, the local memory
   still ships, and the refusal names `leak`.
5. **Suite green.** The workspace suite, which includes `stella-cli`, is green on
   this head: `cargo test` succeeded (17:07:48 → 17:16:00), as did
   `cargo fmt --check` and `cargo clippy -D warnings`, in
   https://github.com/macanderson/stella/actions/runs/33781804160

Docs carried the rename: `crates/stella-cli/README.md` and
`docs/prompts/worker.md` both name `session_workspace_memories` and state the
new refusals.

The issue's follow-on note — `memory/skill_files.rs`'s identical
`if let Ok(content)` shape — is not addressed by this PR and is not a DoD item
here; it rides the volatile channel rather than the cached prefix.

## Checked and judged sound

Worth recording so the next audit does not redo it: `files.sort()` on `PathBuf`
is a byte comparison on Unix, not locale collation, so the load order is a
genuine total order stable across runs — this was the specific worry about the
sorted-filename rule and it holds. The 16,000-character budget loop is a
deterministic greedy best-fit that skips rather than reorders and reports its
drop count. `inject_recall_block` dedups against every prior marker and only ever
appends. `render_today_section`'s `[..10]` slice is safe because the formatted
timestamp is fixed-width ASCII. `clip` and `slugify` iterate by `char`, never
byte-slice.

macanderson:
## What and why

Four defects on the path from `.stella/memories/*.md` to the bytes a provider is sent, all from #5508.

**1. Editing a memory from the deck panicked the session.** `edit_memory_text` built its own `tokio::runtime::Runtime` and called `block_on`. The deck services `e` on a memory row from inside `run_deck_session`'s driver loop, which is already driving a runtime, so the call panicked with "Cannot start a runtime from within a runtime" — an ordinary keypress on the default interactive shell. The write is now `async` and the deck awaits it; the synchronous CLI verb (`run_memory_edit`, reached from `main::run`) owns the one `block_on`.

Both functions moved to `crates/stella-cli/src/memory_cmd/edit.rs`, a sibling module beside the existing `memory_cmd/restore.rs`. `memory_cmd.rs` was 1437 lines against the 1500-line ratchet, so the witness had nowhere to land; the split takes it to 1330 and gives the new logic and its test a home. The two deck call sites in `command_deck.rs` gained `.await` on the existing lines, so the god file grew by nothing.

**2. `/model` and `/agent` re-read memories mid-session.** `refresh_prompts` rebuilds `messages[0]` through the full assembler, which called `append_workspace_memories` and re-read the directory. A memory edited on disk therefore changed the cached prefix under a running session — the exact thing `assemble_system_prompt` documents cannot happen (AGENTS.md rule 7), and a dropped prompt cache for every remaining call. `session_workspace_memories` now assembles the span once per workspace root and hands back the same bytes afterwards, latched the way `knowledge_cutoff_for` reads a catalog installed once at startup. Keyed by root, because one process assembles prompts for more than one (a fleet worker's worktree, a sub-agent's session). The latch sits inside the `project_prompts_allowed` branch, after the isolation early return, so the benchmark-isolation gate is untouched.

**3. An unreadable memory vanished with no diagnostic.** `let Ok(body) = read_to_string(file) else { continue }` skipped invalid UTF-8, a directory, or a permission error in silence, while every other omission from that span already reported itself. Unreadable files are now named in the span.

**4. A memory file could be a symlink pointing anywhere.** `read_dir` lists symlinks and `read_to_string` follows them, and `.stella/memories/` is neither private-tier nor gitignored — so a cloned repository could ship one pointing at any readable path and have its content pasted into the next session's system prompt. An entry whose canonical path escapes the workspace root is refused and named.

Also repaired while in the file: `service_reject_memory`'s doc comment sat above `service_edit_memory`, leaving the reject verb undocumented and the edit verb described by the wrong paragraph.

## Witness mechanism

- `a_memory_edits_from_inside_a_running_runtime` (`memory_cmd/edit.rs`) — a `#[tokio::test]` that calls `edit_memory_text` over a temp workspace. It fails against the old shape **by construction**, not by assertion: `block_on` on a thread a runtime already drives panics, and a `#[tokio::test]` is such a thread, exactly like the deck's driver loop. No `~/.stella` is touched — the path resolves through `existing_workspace_private_sqlite_path`, which is workspace-scoped.
- `a_rebuilt_prompt_carries_the_memories_the_session_opened_with` — assembles, edits `.stella/memories/` on disk, adds a second memory, reassembles, and asserts the two prompts are byte-identical. The old assembler re-read the directory, so the second prompt could not equal the first.
- `an_unreadable_memory_is_named_in_the_prompt` — a memory holding invalid UTF-8 beside a readable one; the readable one still ships and the omission names `corrupt`.
- `a_memory_symlinked_outside_the_workspace_is_refused` (`#[cfg(unix)]`) — a symlink to a file in another temp directory; the target's content is absent, the local memory still ships, and the refusal names `leak`.

## Exemplar

The latch follows this file's own `knowledge_cutoff_for` (a process-wide value resolved once and read from the byte-stable prefix) rather than inventing a new mechanism; the module split follows `memory_cmd/restore.rs`, and `command_deck/session_override.rs` is the pattern for keeping new lines out of `command_deck.rs`.

## Tests deleted

None.

## Follow-ups filed

See the comment on this PR after CI settles — `read_skill_files` in `crates/stella-cli/src/memory/skill_files.rs` has the identical silent-skip shape and is filed separately rather than folded in, because skills ride the volatile channel and the fix wants its own witness.

Closes #5508

## Summary by Sourcery

Keep memory edits safe in running deck sessions and make workspace memory prompts stable, observable, and workspace-contained.

Bug Fixes:
- Prevent memory edits initiated from the interactive deck from panicking an active session.
- Keep workspace memories stable for the duration of a session, including prompt rebuilds triggered by model or agent changes.
- Report unreadable memory files instead of silently omitting them.
- Reject and report memory files that resolve outside the workspace, preventing external symlink content from entering provider prompts.

Enhancements:
- Move memory edit logic into a dedicated module and share the asynchronous operation between the CLI and deck entry points.
- Correct the memory service documentation to describe the corresponding edit and reject commands accurately.

Documentation:
- Update CLI and worker prompt documentation to describe memory validation and session-scoped caching.

Tests:
- Add coverage for editing memories inside an active Tokio runtime, stable prompt rebuilds, unreadable memories, and out-of-workspace symlinks.

## Independent DoD verification (#5508)

The `dod` gate was red because #5508's five boxes were all unticked. Each was
checked against this head (`a7b5de8`) before ticking, item by item:

1. **No panic on a deck memory edit.** `edit_memory_text` is `async` in the new
   `memory_cmd/edit.rs`; `driver_support.rs`'s `service_edit_memory` is `async`
   and awaits it; `command_deck.rs`'s two call sites gained `.await` on the
   existing lines. The one surviving `block_on` is in the sync CLI verb.
   Witness `a_memory_edits_from_inside_a_running_runtime` fails against the old
   shape by construction, since `Runtime::block_on` panics on a thread already
   driving a runtime.
2. **Byte-identical memories span across a rebuild.** `session_workspace_memories`
   latches the span per workspace root (`LazyLock<Mutex<HashMap<PathBuf,
   Arc<str>>>>`, poison recovered). Witness
   `a_rebuilt_prompt_carries_the_memories_the_session_opened_with` asserts
   `opened == rebuilt` after both editing and adding a memory on disk between
   the two assemblies.
3. **An unreadable memory is named.** A `read_to_string` error pushes the stem
   onto `unreadable` and emits its own note beside the budget-drop note.
   Witness `an_unreadable_memory_is_named_in_the_prompt` uses genuine
   invalid UTF-8 (`[0xff, 0xfe, 0xfd]`).
4. **Symlink containment.** `resolves_inside` canonicalises both sides and
   requires `starts_with`; a path that cannot be canonicalised is not
   contained, so a broken link fails closed. Witness
   `a_memory_symlinked_outside_the_workspace_is_refused`.
5. **Suite green.** `cargo test` (the workspace suite, which includes
   `stella-cli`) succeeded on this head, 17:07:48 → 17:16:00, as did
   `cargo fmt --check` and `cargo clippy -D warnings`:
   https://github.com/macanderson/stella/actions/runs/33781804160

All five boxes on #5508 are now ticked, with that record on the issue. The
issue's follow-on note — `memory/skill_files.rs`'s identical silent-skip shape —
is not a DoD item here and is tracked separately (#5765).

### DIFF
--- a/crates/stella-cli/README.md
+++ b/crates/stella-cli/README.md
@@ -255,12 +255,19 @@ that prefix is what the provider's prompt cache keys on, so nondeterminism here
 is a re-billing regression, not a cosmetic one. The code enforces two rules on
 purpose:
 
-- `append_workspace_memories` reads `.stella/memories/*.md`, `files.sort()`s by
+- `workspace_memories_span` reads `.stella/memories/*.md`, `files.sort()`s by
   filename, and concatenates under a 16,000-char budget (`MEMORY_PROMPT_BUDGET_CHARS`);
-  overflow is dropped with a count, never reordered. The prompt is built **once
-  per session** and reused verbatim (including across `/clear`), so a memory
-  saved mid-session deliberately does not appear until the next session —
-  hot-injecting it would invalidate the cached prefix on every save.
+  overflow is dropped with a count, never reordered. A file that cannot be read
+  as text, and one whose canonical path escapes the workspace root, are both
+  refused and named in the span — a memory may not be a symlink out of the
+  checkout, and one that vanishes must leave a sign. The
+  prompt is built **once per session** and reused verbatim (including across
+  `/clear`), so a memory saved mid-session deliberately does not appear until
+  the next session — hot-injecting it would invalidate the cached prefix on
+  every save. `session_workspace_memories` is what holds that: it latches the
+  span per workspace root, so the deck's mid-session `/model` and `/agent`
+  rebuilds re-send the bytes the session opened with instead of re-reading the
+  directory.
 - `append_session_environment` states only session-constant facts — working
   directory, git checkout or not, platform, OS release, shell dialect (#2692).
   A process cannot change its OS mid-session, so the block is prefix-safe.
--- a/crates/stella-cli/src/agent/prompt.rs
+++ b/crates/stella-cli/src/agent/prompt.rs
@@ -514,6 +514,13 @@ pub(crate) const MEMORIES_OMITTED_PREFIX: &str =
 /// carries durable lessons, the recall block carries turn-relevant memories
 /// and skills. The rules rendered here are the same set whose Tier-2 guards
 /// `crate::rules::enforce_workspace_rules` arms at the tool boundary.
+///
+/// "Loaded once per session" is enforced rather than described:
+/// [`session_workspace_memories`] reads `.stella/memories/` the first time a
+/// prompt is assembled for a root and hands back the same bytes afterwards.
+/// The deck rebuilds `messages[0]` on a mid-session `/model` or `/agent`
+/// (`command_deck::session_override`'s `refresh_prompts`), and a rebuild that
+/// re-read the files would move the prefix this promise covers.
 pub(crate) fn assemble_system_prompt(
     base: &str,
     workspace_root: &std::path::Path,
@@ -531,7 +538,7 @@ pub(crate) fn assemble_system_prompt(
         return prompt;
     }
     if authority.project_prompts_allowed {
-        append_workspace_memories(&mut prompt, workspace_root);
+        prompt.push_str(&session_workspace_memories(workspace_root));
     }
     // The cached channel: `must` and `should` records, grouped by force, each
     // carrying its `^handle` so the model can name what it followed. Byte-stable by
@@ -696,9 +703,9 @@ fn workspace_suppression(
         .map_err(|e| format!("suppression state unavailable: {e}"))
 }
 
-/// The memories half of [`assemble_system_prompt`]: append the workspace's
-/// saved memories (filename order, budget-capped, **tombstone-filtered**) to
-/// `prompt`, or leave it untouched when there are none.
+/// The memories half of [`assemble_system_prompt`]: the workspace's saved
+/// memories (filename order, budget-capped, **tombstone-filtered**) as the
+/// bytes the prompt carries, or an empty string when there are none.
 ///
 /// # Forgetting one has to stop it shipping
 ///
@@ -718,7 +725,28 @@ fn workspace_suppression(
 /// left for the model to not notice. Shipping a memory someone forgot is worse
 /// than shipping none: the forget is the explicit instruction, and the file is
 /// still on disk for a later turn.
-fn append_workspace_memories(prompt: &mut String, workspace_root: &std::path::Path) {
+///
+/// # Every omission but a forget is named
+///
+/// A memory the budget dropped, one that could not be read, and one that
+/// resolves outside the workspace are each reported in the span, naming the
+/// files. Skipping an unreadable memory — invalid UTF-8, a directory, a
+/// permission error — in silence costs a reader the one thing they need: a
+/// file that picks up bad bytes is gone from every future prompt, and nothing
+/// says so. A forget is the one omission that stays quiet: it is an
+/// instruction that the memory is gone, and announcing it invites the
+/// asking-about-it that forgetting exists to end.
+///
+/// # A memory file may not point outside the workspace
+///
+/// `read_dir` lists symlinks and `read_to_string` follows them, and
+/// `.stella/memories/` is neither private-tier nor gitignored — so a cloned
+/// repository could ship `.stella/memories/x.md` as a symlink to any readable
+/// path and have its content pasted into the next session's system prompt and
+/// sent to the configured provider. An entry whose canonical path escapes the
+/// workspace root is refused, matching the containment the file tools already
+/// enforce.
+fn workspace_memories_span(workspace_root: &std::path::Path) -> String {
     let dir = workspace_root.join(".stella/memories");
     let mut files: Vec<std::path::PathBuf> = std::fs::read_dir(&dir)
         .map(|entries| {
@@ -729,39 +757,43 @@ fn append_workspace_memories(prompt: &mut String, workspace_root: &std::path::Pa
         })
         .unwrap_or_default();
     if files.is_empty() {
-        return;
+        return String::new();
     }
     files.sort();
 
     let suppression = match workspace_suppression(workspace_root) {
         Ok(suppression) => suppression,
         Err(error) => {
-            prompt.push_str(&format!(
+            return format!(
                 "{MEMORIES_OMITTED_PREFIX}{error}. They are still on disk in .stella/memories/ \
                  and will return once the suppression state is readable."
-            ));
-            return;
+            );
         }
     };
 
     let mut memories = String::new();
     let mut used = 0usize;
     let mut dropped = 0usize;
+    let mut unreadable: Vec<String> = Vec::new();
+    let mut outside: Vec<String> = Vec::new();
     for file in &files {
-        let Ok(body) = std::fs::read_to_string(file) else {
-            continue;
-        };
+        // The tombstone is keyed by filename stem, which is what
+        // `ContextSurface::WorkspaceMemory` records.
         let name = file
             .file_stem()
             .and_then(|n| n.to_str())
             .unwrap_or("memory");
-        // The tombstone is keyed by filename stem, which is what
-        // `ContextSurface::WorkspaceMemory` records.
-        // No count of these is reported. A budget omission is worth telling
-        // the model about, because it can be fixed by consolidating files; a
-        // forget is an instruction that this memory is gone, and announcing
-        // "three memories are being withheld" invites exactly the asking-about
-        // -it that forgetting exists to end.
+        if !resolves_inside(workspace_root, file) {
+            outside.push(name.to_string());
+            continue;
+        }
+        let body = match std::fs::read_to_string(file) {
+            Ok(body) => body,
+            Err(_) => {
+                unreadable.push(name.to_string());
+                continue;
+            }
+        };
         if suppression.suppresses(name, &body) {
             continue;
         }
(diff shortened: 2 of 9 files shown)


macanderson:
Follow-up filed: #5765 — `read_skill_files` in `crates/stella-cli/src/memory/skill_files.rs` has the identical silent-skip shape. It is not folded into this PR because skills ride the volatile recall channel rather than the cached prefix, so the right place to report the skip is a different surface and wants its own witness.

macanderson:
**Answering the Sourcery ❌** on the row "Provide diagnostics for unreadable memory entries and skill files, and prevent memory symlinks from exposing files outside the workspace."

Two of the three parts are in this PR, with witnesses: unreadable workspace memories are named (`an_unreadable_memory_is_named_in_the_prompt`) and a memory resolving outside the workspace root is refused and named (`a_memory_symlinked_outside_the_workspace_is_refused`).

The skill-file half is out of scope on purpose, and the issue agrees: #5508 lists it as a lower-severity aside ("skills ride the volatile channel rather than the cached prefix") and its Definition of done carries no skills item. The fix also lands on a different surface — the skill block is volatile context the model reads, so where the notice belongs is a real question rather than a copy of the memory answer, and it wants its own witness. Filed as #5765, written as a handoff with its own DoD.

Every other row is ✅.

macanderson:
Automated PR sweep. CI is green on `a7b5de8` (`fmt + clippy + test` at 17:24:35Z) and the `dod` gate now passes against #5508's ticked boxes, so this is not blocked on anything. One observation for the follow-up comment the description says is coming, raised here rather than filed so it does not duplicate whatever you were about to write.

**`session_workspace_memories` is latched for the life of the *process*, and the guarantee it implements is per *session*.**

The doc comment states this plainly — "held for the life of the process, which covers the session" — so it is a stated scope rather than an oversight, and for the `stella` binary it is exactly right: one process is one session, and the `/clear` case is the behaviour you want. The gap is the case where it isn't. `SPANS` is a `LazyLock<Mutex<HashMap<PathBuf, Arc<str>>>>` with an insert and no eviction, so two *sequential* sessions over one root in a single process would have the second serve the first's bytes. That inverts the promise in the same paragraph: a memory saved in session 1 would never appear in session 2, where the contract is that it appears in the next one.

Keying by root already handles the concurrent case the comment names — a fleet worker's worktree and a sub-agent's session are different roots, so they get different entries. The sequential-same-root case is the one with no answer.

Stating the limit of this rather than overclaiming: **I did not establish a live caller that reaches it.** `assemble_system_prompt` lives in `stella-cli`, `stella-serve` does not link that crate, and I did not trace whether any fleet path drives two sessions over one root inside one process. So this may be latent today and load-bearing the moment a host does what `stella-runtime`'s README describes. If it is worth a line, an eviction keyed on session open is the cheap shape; if it is genuinely unreachable, the doc comment saying *why* one process cannot hold two sessions over one root would close it just as well.

Nothing here blocks the merge — the four defects and their witnesses all check out against the diff.

---

PR #5767, opened 2026-09-03
fix(stella-tools): resolve a real shell instead of the Windows WSL stub

### LINKED ISSUE #4861: On Windows, plain bash can run the WSL launcher instead of a real shell
## Problem

`Bash::execute` (`crates/stella-tools/src/bash.rs`) and the hook runner's operator path (`crates/stella-tools/src/hook_runner.rs`, the branch that runs when `action.plugin` is `None`) both spawn `Command::new("bash")` unconditionally, with no platform-specific resolution. On `windows-latest` GitHub Actions runners — and plausibly on real Windows machines where `%SystemRoot%\System32` precedes Git for Windows in `PATH` — this resolves to Windows' own `bash.exe`, the WSL launcher stub, not a real shell.

With no WSL distribution installed, that stub prints an install prompt to stdout and exits nonzero instead of running the given `-c` script:

```
Windows Subsystem for Linux has no installed distributions.
Use 'wsl.exe --list --online' to list available distributions
and 'wsl.exe --install <Distro>' to install.
```

So the `bash` tool and every operator-authored hook (any hook without a `[plugin]` origin) silently do nothing useful on such a machine — they don't error clearly, they run the WSL prompt as if it were a shell command's output.

Found while proving #4698's group-kill witness on Windows CI: `crates/stella-tools/tests/group_kill_witnesses.rs` had to work around this with a test-only `PATH` fix (`ensure_bash_resolves_to_a_real_shell`) that prepends Git for Windows' documented install location, `%ProgramFiles%\Git\bin`, so the witness exercises a real shell. That workaround belongs in a test file, not as the production fix — the actual resolution strategy (prefer an absolute Git-Bash path? detect the WSL stub and fail with a clear, actionable error? require and document a PATH precondition?) is a design call, not a test's to make.

## Repro

On a `windows-latest` runner (or any Windows machine with no WSL distro installed and `System32` ahead of Git in `PATH`):

```rust
Bash::new(None)
    .execute(&json!({"command": "echo hi"}), &ToolCtx::bare(some_dir))
    .await
```

returns `Ok` with the WSL install-prompt text as stdout, not `hi`.

## Done means

- `Bash::execute` and the hook runner's operator path resolve to a real shell on Windows regardless of `PATH` order, or fail with a clear, named error when none can be found — never silently run the WSL launcher stub as if it answered.
- A witness test (spawn on a `PATH` with the WSL stub ahead of a real shell, or equivalent) proves the fix, ported the same way `wrapper-plugin-fixture` ported the plugin transport's Windows tests (#3497) — no shell-script fixture, since the whole point is testing what happens when the wrong "bash" is found.
- `crates/stella-tools/tests/group_kill_witnesses.rs`'s `ensure_bash_resolves_to_a_real_shell` workaround can be deleted once the production code no longer needs it.

Related: #4697 (widening `windows-check.yml` to the crate's full test suite) and #3550 (the group-kill guard this defect was found underneath).

## Definition of done

- [x] `Bash::execute` and the hook runner's operator path resolve to a real shell on Windows regardless of `PATH` order, or fail with a clear, named error when none can be found — never silently running the WSL launcher stub as if it answered.
- [x] A witness test (spawn on a `PATH` with the WSL stub ahead of a real shell, or equivalent) proves the fix.
- [x] It is ported the way `wrapper-plugin-fixture` ported the plugin transport's Windows test, so it runs where it matters.

macanderson:
## What

`Bash::execute` (`crates/stella-tools/src/bash.rs`) and the hook runner's
operator path (`crates/stella-tools/src/hook_runner.rs`) both spawned
`Command::new("bash")` with no platform check. On Windows,
`CreateProcessW`'s own search order checks `%SystemRoot%\System32` before it
ever looks at `PATH`, so this silently ran Windows' own WSL launcher stub
instead of a real shell whenever no WSL distribution was installed — the
stub prints an install prompt and exits nonzero instead of running the
script it was handed, and no `PATH` reordering fixes it.

## Why

Every operator-authored hook and every `bash` tool call did nothing useful
on such a machine, with no clear error — the WSL install prompt ran as if
it were the command's own output. Found under #4698's group-kill witness
work; `crates/stella-tools/tests/group_kill_witnesses.rs` had to `ignore`
its two `bash`/hook witnesses on Windows until this landed.

## The fix

New crate-private module `crates/stella-tools/src/shell_resolve.rs`. On
unix, `bash_command()` is exactly the old `Command::new("bash")` — there is
no stub to trip over there. On Windows it resolves an absolute path first:

1. Git for Windows' own documented install locations (`%ProgramFiles%` /
   `%ProgramFiles(x86)%\Git\bin\bash.exe`), checked first since that is
   where nearly every real Windows install actually has its shell.
2. A manual walk of `PATH`, skipping any directory equal to
   `%SystemRoot%\System32` — the WSL stub's home — so a real shell earlier
   or later on `PATH` wins regardless of order.
3. If nothing is found, a named `ShellResolutionError` listing every
   location tried, surfaced as a classified tool error (`bash`) or a
   `HookExecError::SpawnFailed` (the hook runner) — never a silent run of
   whatever `bash` happened to resolve to.

Both `bash.rs` and `hook_runner.rs` now call `crate::shell_resolve::bash_command()`
instead of constructing `Command::new("bash")` directly.

## Witness

`shell_resolve`'s own unit tests fabricate the exact `PATH` shape the
defect report described — a WSL-stub-shaped `bash.exe` sitting in the
system directory, listed *first* on `PATH`, with a real shell listed after
it — and prove resolution picks the real shell regardless of that order.
These tests exercise `resolve_bash_with`, logic that did not exist before
this PR: the old call site was an unconditional `Command::new("bash")`
with nothing to unit test, so the tests fail to compile against the prior
tree by construction.

The two Windows-only witnesses in `group_kill_witnesses.rs` (#4698,
`a_dropped_bash_call_kills_the_process_group` and
`a_timed_out_hook_leaves_no_surviving_grandchild`) that were
`#[cfg_attr(windows, ignore = "blocked on #4861: ...")]` now run
unconditionally — `windows-check.yml`'s `windows-latest` job runs
`--test group_kill_witnesses` on every push touching `crates/stella-tools/**`,
so this PR's own CI run proves the fix on a real Windows machine, not only
in a fabricated unit test. `windows-check.yml`'s own comment is updated to
match.

Verified locally (targeted, per CLAUDE.md — no workspace build):
- `cargo clippy -p stella-tools --all-targets -- -D warnings` — clean.
- `cargo test -p stella-tools --lib shell_resolve` — 4/4 pass.
- `cargo test -p stella-tools --test group_kill_witnesses` — 3/3 pass (unix; the two `bash`/hook witnesses now run unignored).
- `cargo test -p stella-tools --lib bash::` and `--lib hook_runner::` — unaffected, all pass.
- `make guards-fast` — clean, including `check-prose` and `check-file-size`.

Tests deleted: none. Two `#[cfg_attr(windows, ignore = "blocked on #4861: ...")]`
attributes removed (the fix they were waiting on), not the tests themselves.

Closes #4861

## Summary by Sourcery

Resolve Bash to a real shell on Windows and restore cross-platform execution and process-group witnesses.

New Features:
- Resolve a real Bash executable on Windows for tool calls and operator-authored hooks, preferring Git for Windows and excluding the WSL launcher directory.

Bug Fixes:
- Prevent Windows Bash tools and hooks from silently invoking the WSL installation stub when no WSL distribution is available.
- Return explicit, classified errors when no usable Bash executable can be found.

Enhancements:
- Run the Bash and hook process-group cleanup witnesses on Windows now that they can use a real shell.

CI:
- Update the Windows check workflow to run all group-kill witnesses unconditionally.

Tests:
- Add shell-resolution tests covering Git for Windows preference, stub-first PATH handling, and descriptive failures.

## Independent DoD verification (#4861)

The `dod` gate went red at 17:08 on #4861's three unticked boxes. They are
ticked now, and each was checked against this head (`6b69a50`) first:

1. **Both spawn sites resolve a real shell or fail by name.**
   `bash.rs`'s `Bash::execute` and `hook_runner.rs`'s `None`-plugin arm both
   call `crate::shell_resolve::bash_command()`. On Windows that returns an
   absolute path — Git for Windows' documented locations first, then a
   hand-walked `PATH` that skips `%SystemRoot%\System32` — so
   `CreateProcessW`'s search order is bypassed entirely rather than
   out-ordered. A failure is `ShellResolutionError` naming every location
   tried, surfaced as `ErrorClass::Environment` from the tool and
   `HookExecError::SpawnFailed` from the hook runner. Neither path can now
   run the stub as if it answered.
2. **Witness on a stub-first `PATH`.**
   `skips_the_system_directory_even_when_it_leads_path` builds exactly the
   reported shape — `System32\bash.exe` listed first, a real `bash.exe`
   after it — and asserts the real one is chosen. It cannot pass against the
   old tree, which had no `resolve_bash_with` to call. Two siblings cover
   the Git-for-Windows preference and the names-everything-tried error.
3. **Ported so it runs where it matters.** Both
   `#[cfg_attr(windows, ignore = "blocked on #4861: …")]` attributes are
   gone from `group_kill_witnesses.rs`, and `windows-check.yml`'s comment no
   longer claims two of the three are skipped — so all three witnesses run
   on the `windows-latest` job, which is the `wrapper-plugin-fixture`
   pattern the issue asked for (a real Windows run, not a shell-script
   fixture).

Not a DoD item here, tracked separately: `exec.rs`'s unit tests still
hardcode `Command::new("bash")` (#5768).

### DIFF
--- a/.github/workflows/windows-check.yml
+++ b/.github/workflows/windows-check.yml
@@ -52,11 +52,10 @@ name: windows-check
 # with the same guard and each backing onto the compile blocker two
 # paragraphs up: `--test group_kill_witnesses` links only the compiled
 # `stella-tools` library, not its `--lib` test binary, so it runs here without
-# waiting on that binary's Windows compile (#4698). Only the custom-tool
-# witness runs unconditionally here today — the other two are `ignore`d on
-# Windows, where plain `bash` resolves to Windows' own WSL launcher stub
-# rather than a shell (#4861), a separate defect this step's green does not
-# claim is fixed.
+# waiting on that binary's Windows compile (#4698). All three witnesses run
+# unconditionally here: the `bash` and hook witnesses used to be `ignore`d on
+# Windows, where plain `bash` resolved to Windows' own WSL launcher stub
+# rather than a shell, until #4861 gave both spawn sites a real resolver.
 #
 # Its own file rather than a job in ci.yml for the reason wire-schema.yml has
 # one: ci.yml's required job is what every PR waits on, and a Windows runner is
--- a/crates/stella-tools/src/bash.rs
+++ b/crates/stella-tools/src/bash.rs
@@ -64,7 +64,6 @@ use std::time::Duration;
 use async_trait::async_trait;
 use serde_json::Value;
 use stella_protocol::tool::{ToolOutput, ToolSchema};
-use tokio::process::Command;
 
 use crate::registry::Tool;
 
@@ -536,7 +535,15 @@ impl Tool for Bash {
             command
         };
 
-        let mut cmd = Command::new("bash");
+        let mut cmd = match crate::shell_resolve::bash_command() {
+            Ok(cmd) => cmd,
+            Err(err) => {
+                return ToolOutput::classified_error(
+                    stella_protocol::ErrorClass::Environment,
+                    format!("bash: {err}"),
+                );
+            }
+        };
         cmd.args(["-c", command]);
         cmd.current_dir(root);
         // Full spawn policy, not just the credential scrub: a hook-exported
--- a/crates/stella-tools/src/hook_runner.rs
+++ b/crates/stella-tools/src/hook_runner.rs
@@ -102,7 +102,12 @@ impl HookRunner for HostHookRunner {
         let mut command = match &action.plugin {
             Some(origin) => plugin_command(origin, |name| std::env::var(name).ok())?,
             None => {
-                let mut command = tokio::process::Command::new("bash");
+                let mut command = crate::shell_resolve::bash_command().map_err(|err| {
+                    HookExecError::SpawnFailed {
+                        command: action.command.clone(),
+                        message: err.to_string(),
+                    }
+                })?;
                 command.arg("-c").arg(&action.command);
                 // Full spawn policy: a hook running from inside a git hook
                 // must not inherit the outer repo's GIT_DIR, and its stdout
--- a/crates/stella-tools/src/lib.rs
+++ b/crates/stella-tools/src/lib.rs
@@ -72,6 +72,7 @@ pub mod registry;
 pub mod rootfd;
 pub mod scratch;
 pub mod search;
+mod shell_resolve;
 pub mod skill_grant;
 pub mod skill_plane;
 pub mod staleness;
--- a/crates/stella-tools/src/shell_resolve.rs
+++ b/crates/stella-tools/src/shell_resolve.rs
@@ -16,10 +16,10 @@
 //!
 //! [`bash_command`] replaces the bare call. On unix it is exactly
 //! `Command::new("bash")`. On Windows it finds an absolute path to a real
-//! shell first ([`resolve_bash`]) and runs that path directly. An absolute
-//! path skips the search order entirely. A caller that gets back
-//! [`ShellResolutionError`] knows no real shell was found. It can say so,
-//! instead of quietly running whatever `bash` turned out to mean.
+//! shell first (`resolve_bash`, which is compiled only there) and runs that
+//! path directly. An absolute path skips the search order entirely. A caller
+//! that gets back [`ShellResolutionError`] knows no real shell was found. It
+//! can say so, instead of quietly running whatever `bash` turned out to mean.
 
 #[cfg(any(test, windows))]
 use std::path::{Path, PathBuf};
--- a/crates/stella-tools/src/shell_resolve.rs
+++ b/crates/stella-tools/src/shell_resolve.rs
@@ -1,0 +1,212 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+// Copyright (c) 2026 Oxagen, Inc. Commercial licensing: licensing@oxagen.sh
+
+//! Real-shell lookup for two spawn sites: [`crate::bash`] and the hook
+//! runner's own-shell path ([`crate::hook_runner`]). Both call
+//! `Command::new("bash")` and trust it to mean a real shell.
+//!
+//! On unix that trust is safe. `$PATH` finds a real shell there, and no
+//! stub gets in the way. On Windows it is not safe. `Command::new("bash")`
+//! resolves through `CreateProcessW`'s own search order. That order checks
+//! `%SystemRoot%\System32` before it ever looks at `PATH`. Windows ships a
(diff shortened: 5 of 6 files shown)


macanderson:
## Answering Sourcery's ❌ row

The "Assessment against linked issues" table carries one `❌`:

> Remove the test-only `ensure_bash_resolves_to_a_real_shell` PATH workaround from `group_kill_witnesses.rs` once production shell resolution makes it unnecessary.
> *The PR removes the Windows ignore attributes and enables the witnesses, but the diff does not remove the existing `ensure_bash_resolves_to_a_real_shell` workaround.*

**This row is a misreading of the diff, and no fix is owed: there is no such
workaround in this tree to remove.**

What I checked, rather than inferred — `ensure_bash_resolves_to_a_real_shell` is
absent from all three of:

- this PR's head, `74e099b`;
- this PR's base commit, `ab40330`;
- `origin/main` as of `db3bb69`.

`git grep` over `crates/` returns nothing at any of them. The witness file's
only functions are `beats`, `still_beating`, `outlived`, `started_beating`,
`backgrounding_command` and the three witnesses themselves — there is no
`PATH`-prepending helper under that name or any other. The two `PATH` mentions
left in the file are prose in doc comments describing the defect ("no matter the
`PATH` order"), which is what the fix now makes true.

Where the row comes from: #4861's **prose** ("Found while proving #4698's
group-kill witness on Windows CI…that workaround belongs in a test file") names
the helper as context for why the issue was filed, and closes with "can be
deleted once the production code no longer needs it". Sourcery lifted that
sentence into a fourth objective. It is not one of the issue's three DoD boxes,
all of which are ticked and all of which the `dod` check passes on this head.
The helper it names had already left the tree before this PR branched — I can
say it is absent at the base commit, but not when it was removed: this container
holds a shallow clone (50 commits), so I did not trace the removal and am not
claiming a commit for it.

Nothing is filed as residue, because there is no residue — an issue asking to
delete a symbol that does not exist would be a handoff to nowhere. The one real
adjacent item found while doing this work is already filed: #5768, `exec.rs`'s
unit tests still hardcoding `Command::new("bash")`.

CI on this head is fully green — all 20 checks, including the `windows-latest`
job that runs all three group-kill witnesses unignored, which is the evidence
the fix actually works on the platform it is about.

---

Issue #5777, opened 2026-09-03
A best-effort store write that fails leaves no trace anywhere


macanderson:
## Problem

The learning loop writes to the context store on a best-effort footing: a
failed write must never fail the turn it describes. Several of those calls
throw the error away and say nothing anywhere:

- `crates/stella-cli/src/memory.rs` — `SessionMemory::record_episode` and
  `SessionMemory::record_taxonomy`, both `let _ = self.store.upsert(delta)`

`record_taxonomy` writes a whole taxonomy in one transaction, so a single bad
record loses the batch, and the operator is told nothing at all.

There is nowhere for the message to go. The workspace has no logger — no
`tracing`, no `log` — and the deck calls `record_episode` while it owns the
terminal, so `eprintln!` would corrupt the frame. That is why these are
swallows and not warnings today.

## Ask

Give the CLI one way to report a best-effort failure that a TUI-owning caller
can use. A small sink the deck can drain into its own log pane, a debug file
under `.stella/private/`, or an `AgentEvent` the stream already carries would
each do. Then route these calls into it.

## Files

- `crates/stella-cli/src/memory.rs` — the two `let _ = ... upsert` calls
- `crates/stella-cli/src/command_deck.rs` — `debug_log_path`, the nearest
  existing thing to a sink (read-only reference — see the file-placement
  item below, this file may not gain new lines)

## Verify

Make one of those writes fail (a delta holding a blank record is refused by
`ContextStore::upsert`) and assert the report reaches the sink.

## Definition of done

- [ ] A reporting path exists that is safe to call while the deck
      (`command_deck.rs`) owns the terminal — no `eprintln!`/stdout write,
      since that can corrupt the TUI frame.
- [ ] Both `let _ = self.store.upsert(delta).await` calls in
      `crates/stella-cli/src/memory.rs` (`record_episode` ~line 988,
      `record_taxonomy` ~line 1053) route their failure through that path
      instead of discarding it.
- [ ] **Witness test**: fails on today's tree (a rejected `upsert` currently
      produces no observable trace anywhere) and passes once the reporting
      path lands. Force `ContextStore::upsert` to reject a blank record (per
      Verify above) and assert the report reaches the sink.
- [ ] **File organization**: the new sink lands in `memory.rs` or a new
      sibling module under `stella-cli` — never inside
      `crates/stella-cli/src/command_deck.rs`. That file is one of
      `stella-cli`'s grandfathered god files (AGENTS.md's god-files table;
[...]

Issue #5776, opened 2026-09-03
Episode ids still collide on paths that set no execution id


macanderson:
## Problem

An episode's id is a hash of its summary, its start, and its end. Both times
are whole seconds. Two turns that share a prompt inside one second therefore
hash to one id, and the second write lands on the first turn's row: its outcome
and its file list are replaced, and nothing reports a loss.

#5326 added `EpisodeInput::occurrence`, a key a caller passes to keep two such
turns apart, and `SessionMemory::record_episode` passes the turn's
`execution_id` when it has one. A path that never calls
`SessionMemory::set_execution_id` still passes `None`, so it still collides.

Today the chat path gets an id from `stamp_and_record_skill_usage`
(`crates/stella-cli/src/agent/skill_usage.rs`) and the fleet door sets one in
`crates/stella-cli/src/fleet_cmd.rs`. Every other episode writer does not:

- `crates/stella-cli/src/agent/goal.rs` — both `record_episode` calls
- `crates/stella-cli/src/agent.rs` — `record_turn_episode`, on a session that
  may have no execution id
- `crates/stella-cli/src/command_deck/authoring.rs` — its `record_turn_episode`
  call

## Ask

Give every episode writer a key. Either set the execution id on those paths, or
give `record_episode` a fallback key the session can always produce.

## Files

- `crates/stella-cli/src/memory.rs` — `SessionMemory::record_episode`,
  `set_execution_id`
- `crates/stella-context/src/writeback.rs` — `EpisodeInput::occurrence`,
  `EpisodeInput::public_id`

## Verify

`crates/stella-cli/src/memory/tests/episode_identity.rs` holds the shape of the
test: two turns, one fixed clock, one prompt, two episodes. Add a case per path
that writes an episode without an execution id.

## Duplicate check (triage pass)

Searched the tracker for prior reports of episode-id collision / occurrence
keys. The only match is #5326 (closed, merged via #5778) — the issue that
introduced `EpisodeInput::occurrence` and fixed the chat and fleet paths, and
whose own description is the source of the three call sites left open here.
Not a duplicate; this is #5326's unfinished tail.

## Definition of done

- [ ] Every `record_episode` caller supplies an occurrence key, or the session
      derives one when the caller has none.
- [ ] A test per writing path (`agent/goal.rs` ×2, `agent.rs`,
      `command_deck/authoring.rs`) shows two same-second turns keeping two
      rows. Each test is a **witness**: it fails against today's code and
[...]

PR #5778, opened 2026-09-03
fix(stella-context): six recall and writeback defects, each with a witness

### LINKED ISSUE #5326: Recall and writeback hardening: six traced small defects in stella-context
## Problem

Bundle from a full review of `crates/stella-context` (each traced; none executed):

1. `retrieval/ranking.rs` `pack_to_budget` iterates a hardcoded `[Normal, Deferred]` tier array. A third `RecallTier` variant compiles clean and its candidates land in neither kept nor dropped — the exact partition failure the L-C5 property bans, invisible until the generator learns the variant. Fix: a `RecallTier::ALL` const produced by a match (compile error on new variants), the `KNOWN_TYPE_TAGS` pattern.
2. `retrieval/tuning.rs` `RecallTuning::sanitized`: `f32::clamp` passes NaN through, so a NaN `mmr_lambda` or `min_coverage` survives the function whose doc says it makes misconfiguration safe; a NaN `min_coverage` silently disables the lexical fallback. JSON cannot encode NaN so it is unreachable today; latent for TOML-shaped config. Fix: `is_finite` guards like `rrf_k` already has.
3. `crates/stella-store/src/private.rs`: a crash mid-append to `reflections.jsonl` leaves a partial line with no terminator; the next append concatenates onto it, losing both records to every line-parse reader, silently. Fix: write a leading newline when the file's last byte is not one, or length-check on read.
4. Episode identity is `hash(summary truncated to 240ch, started, ended)` at second resolution (`writeback.rs` plus `crates/stella-cli/src/memory.rs`), so two same-prompt turns in one second silently merge, overwriting outcome/files.
5. `upsert`'s phase-A "reused" decision is made outside the phase-B transaction (`writeback.rs`), so a concurrent `compact` deleting that orphan vector leaves the new node unembedded until the next mount.
6. A whitespace-only lesson (blank `truncate_label` -> `InvalidInput`) fails its whole delta atomically, and the CLI swallows the error (a `let _ = store.upsert(...)` in `crates/stella-cli/src/memory.rs`) — one blank lesson silently discards its siblings' writes.

## Definition of done

- [x] `RecallTier::ALL` exists, is derived from a total match, and `pack_to_budget` walks it; a test covers every tier the vocabulary holds.
- [x] `RecallTuning::sanitized` returns a finite `mmr_lambda` and `min_coverage` for every input, with a test over NaN and both infinities.
- [x] `append_workspace_private_line` cannot glue a new record onto an unterminated one, with a test that plants the fragment.
- [x] Two turns that share a prompt and a second write two episodes, and one turn recorded twice still writes one; tests on both the store and the CLI side.
- [x] The reuse decision is re-checked inside the write transaction, and a vector reclaimed while embedding is embedded again; a test drives the race.
- [x] A blank record is rejected before the delta is embedded or written, and the reflection writer keeps blank lessons out of the batch so their siblings still land.
- [x] Anything found and not fixed is filed as an issue and named in the PR.
- [x] CI green on the pull request.

macanderson:
Six traced defects in `stella-context`, plus the two callers in `stella-cli`
and `stella-store` that the last two need. Each has its own test. The design
was posted on the issue before implementation and is followed.

## 1. A tier the packer never walks

`pack_to_budget` walked a hand-written `[Normal, Deferred]`. A third
`RecallTier` compiles clean, and its candidates land in neither `kept` nor
`dropped` — the partition `L-C5` forbids, invisible until something writes the
new tier.

`RecallTier::ALL` is now derived from a `recall_tiers!` table, modelled on
`model_call_roles!` in `stella-protocol`'s `event::call_role` (the exemplar the
issue names): one token list builds `ALL` and an exhaustive `match`, and a
`const _: ()` forces that match to be evaluated. A tier not named there is
`E0004`. `pack_to_budget` walks `ALL`.

**Witness:** `every_tier_reaches_the_packer` builds one candidate per entry of
`ALL` with no slots at all and asserts every one is reported as dropped. It
cannot compile against the old code, which has no `ALL`.

## 2. NaN survives `RecallTuning::sanitized`

`f32::clamp` asks whether a value is in range, not whether it is a number, so a
NaN went straight through the function whose job is to make a misconfiguration
safe. A NaN `mmr_lambda` makes every MMR comparison false; a NaN `min_coverage`
never falls under the coverage score, so the lexical fallback stops firing.
Both knobs now take the `is_finite` guard `rrf_k` already had.

**Witness:** `a_knob_that_is_not_a_number_falls_back_to_its_default` over NaN
and both infinities, plus an in-range case proving the clamp still clamps.

## 3. A torn log line swallows the next record

`append_workspace_private_line` wrote onto whatever was there. A process killed
mid-append leaves a record with no terminator, and the next append glued the
two into one line that parses as neither. It now reads the last byte and writes
a newline first when one is missing.

**Witness:** `an_unterminated_record_does_not_swallow_the_next_one` plants the
fragment, appends, and asserts three whole lines.

## 4. Two turns in one second are one episode

An episode is keyed by its summary and its window, both at second resolution,
so two turns on one prompt inside a second hashed to one id and the second
write replaced the first turn's outcome and file list. `EpisodeInput` gains
`occurrence`, folded into `public_id` only when set — every `epi_` id already
on disk is unchanged — and `record_episode` passes the turn's `execution_id`.

**Witness:** `two_turns_in_one_second_write_two_episodes` and
`an_episode_without_an_occurrence_keeps_its_old_identity` in `stella-context`,
and `two_executions_in_one_second_keep_their_own_episodes` plus
`one_execution_recorded_twice_keeps_one_episode` on the CLI side. Without the
key each pair writes one row.

## 5. A reused vector reclaimed mid-write

`upsert`'s reuse decision was made before the embedder ran and the write
happened after it. Until the delta's nodes exist those vectors belong to no
node, which is exactly what `compact` reclaims — so the new node landed
unembedded until the next mount's warm index. The write transaction now
re-asks, and on a loss rolls back and retries with that content forced into the
embed set (`REUSE_RACE_ATTEMPTS`).

**Witness:** `a_vector_reclaimed_while_embedding_is_written_again` uses an
embedder that deletes the row it is about to reuse, through the store's own
connection, on the call that serves the delta. The store holds no lock while it
embeds, so this is the real race, single-threaded and deterministic.

## 6. A blank lesson discards its siblings

A memory's mirror node takes its label from its own text, so a whitespace-only
lesson mints a node with no citable name and the store refuses it (`L-C4`) —
half way through the transaction, after the delta paid for its embeddings, with
the whole batch rolled back. `upsert` now names the bad record before anything
is embedded or written, and the reflection writer drops a lesson with no words
(`writable_lessons`) so a blank one costs only itself.

**Witness:** `a_blank_record_is_named_before_anything_is_written` in
`stella-context`, and `a_blank_lesson_is_dropped_and_its_siblings_are_kept`
plus `a_lesson_with_only_a_trigger_is_still_writable` in `stella-cli`.

## Remaining work filed

- #5776 — episode ids still collide on the paths that set no execution id
  (`goal.rs`, `agent.rs`, the deck's authoring door).
- #5777 — a best-effort store write that fails leaves no trace anywhere. The
  `let _ = ... upsert` in `record_episode` stays a swallow in this PR and says
  why: the deck calls it while it owns the terminal, and the workspace has no
  logger.

Tests deleted: None.

Closes #5326

## Summary by Sourcery

Fix six recall and writeback defects across context storage, episode recording, reflection handling, and workspace logs while adding witness tests for each failure mode.

Bug Fixes:
- Prevent recall packing from losing candidates when additional recall tiers are introduced.
- Treat non-finite recall tuning values as invalid and restore safe defaults.
- Preserve record boundaries when appending to truncated workspace-private logs.
- Disambiguate episodes from separate executions occurring in the same second while preserving retry idempotency and legacy identities.
- Retry writes when reusable embeddings are reclaimed during an upsert.
- Reject blank records with actionable validation errors and allow reflection batches to retain valid lessons alongside blank ones.

Enhancements:
- Make recall-tier enumeration exhaustive and drive packer precedence from the complete tier vocabulary.

Tests:
- Add regression coverage for recall-tier packing, non-finite tuning values, truncated log recovery, episode identity collisions, embedding reuse races, and blank-record handling across context, CLI, and store.

### DIFF
--- a/crates/stella-cli/src/memory.rs
+++ b/crates/stella-cli/src/memory.rs
@@ -957,6 +957,16 @@ impl SessionMemory {
             format_rfc3339(now_secs),
         )
         .with_domains(domains);
+        // The turn's own execution row is what separates two turns that share a
+        // prompt and land in the same second — both timestamps are
+        // second-resolution, so without it the second write lands on top of the
+        // first and the earlier turn's outcome and files are gone. A turn that
+        // is recorded twice passes the same id and still updates one row. A
+        // path that never adopted `set_execution_id` keeps the older identity
+        // and the collision with it, which is tracked on its own issue.
+        if let Some(execution) = self.execution_id {
+            episode = episode.with_occurrence(execution.to_string());
+        }
         episode.outcome = outcome;
         episode.files_touched = files_touched.iter().map(|(path, _)| path.clone()).collect();
 
@@ -964,6 +974,12 @@ impl SessionMemory {
             episodes: vec![episode],
             ..Default::default()
         };
+        // Ignored on purpose, and it is the weakest link on this path: a
+        // failed episode must not fail the turn it describes, the deck calls
+        // this while it owns the terminal so stderr is not available, and the
+        // workspace has no logger to route the failure to. Reporting it is
+        // tracked on its own issue. The delta holds one record, so nothing
+        // else is lost with it.
         let _ = self.store.upsert(delta).await;
     }
 
--- a/crates/stella-cli/src/memory/learning.rs
+++ b/crates/stella-cli/src/memory/learning.rs
@@ -48,6 +48,30 @@ mod skill_lifecycle;
 #[cfg(test)]
 mod dedupe;
 
+/// Drop the lessons the store cannot write, keeping every one it can.
+///
+/// A memory's mirror node takes its display name from the memory's own text,
+/// and a node with a blank name is refused — a node has to be humanly citable
+/// (`L-C4`). A lesson with no words in either half is therefore unwritable, and
+/// the whole delta is one transaction, so handing the store one of them throws
+/// away every good lesson of the same turn. The model writes the text, so an
+/// empty one is model output like any other rather than a case that cannot
+/// arise.
+///
+/// The check is `recall_text`'s output, not the lesson alone: a lesson with no
+/// body but a real trigger still has words, and it is that composed string the
+/// store labels the node from.
+fn writable_lessons(lessons: Vec<ReflectionLesson>) -> Vec<ReflectionLesson> {
+    lessons
+        .into_iter()
+        .filter(|l| {
+            !applicability::recall_text(&l.lesson, &l.trigger)
+                .trim()
+                .is_empty()
+        })
+        .collect()
+}
+
 /// One SPEC 6.3 `memory` (log) event per lesson the store just wrote.
 ///
 /// Paired positionally with [`stella_context::UpsertReceipt::memory_node_ids`],
@@ -74,30 +98,6 @@ mod dedupe;
 ///   evidently of the next; a domain lesson is still true on a task the agent
 ///   has never seen. That is the distinction [`LessonKind`] draws, and the one
 ///   [`LessonKind::recall_tier`] already spends the recall budget on.
-/// Drop the lessons the store cannot write, keeping every one it can.
-///
-/// A memory's mirror node takes its display name from the memory's own text,
-/// and a node with a blank name is refused — a node has to be humanly citable
-/// (`L-C4`). A lesson with no words in either half is therefore unwritable, and
-/// the whole delta is one transaction, so handing the store one of them throws
-/// away every good lesson of the same turn. The model writes the text, so an
-/// empty one is model output like any other rather than a case that cannot
-/// arise.
-///
-/// The check is `recall_text`'s output, not the lesson alone: a lesson with no
-/// body but a real trigger still has words, and it is that composed string the
-/// store labels the node from.
-fn writable_lessons(lessons: Vec<ReflectionLesson>) -> Vec<ReflectionLesson> {
-    lessons
-        .into_iter()
-        .filter(|l| {
-            !applicability::recall_text(&l.lesson, &l.trigger)
-                .trim()
-                .is_empty()
-        })
-        .collect()
-}
-
 fn memory_logged_events(
     novel: &[ReflectionLesson],
     memory_node_ids: &[String],
--- a/crates/stella-cli/src/memory/learning.rs
+++ b/crates/stella-cli/src/memory/learning.rs
@@ -74,6 +74,30 @@ mod dedupe;
 ///   evidently of the next; a domain lesson is still true on a task the agent
 ///   has never seen. That is the distinction [`LessonKind`] draws, and the one
 ///   [`LessonKind::recall_tier`] already spends the recall budget on.
+/// Drop the lessons the store cannot write, keeping every one it can.
+///
+/// A memory's mirror node takes its display name from the memory's own text,
+/// and a node with a blank name is refused — a node has to be humanly citable
+/// (`L-C4`). A lesson with no words in either half is therefore unwritable, and
+/// the whole delta is one transaction, so handing the store one of them throws
+/// away every good lesson of the same turn. The model writes the text, so an
+/// empty one is model output like any other rather than a case that cannot
+/// arise.
+///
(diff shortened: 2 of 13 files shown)


macanderson:
## Independent DoD verification (#5326)

Seven of #5326's eight boxes are ticked. Each was checked against this head (`5e4641b`) before ticking:

1. **`RecallTier::ALL` from a total match.** The `recall_tiers!` macro in `retrieval/outcome.rs` builds `ALL` and `every_tier_is_in_all` from one token list, and `const _: () = RecallTier::Normal.every_tier_is_in_all();` forces the exhaustive match to be evaluated — so a tier absent from the table is `E0004`, a build error. `ranking.rs` walks `for &band in RecallTier::ALL`. Witness `every_tier_reaches_the_packer` builds one candidate per `ALL` entry with zero slots and asserts every one is reported dropped; it cannot compile against `main`, which has no `ALL`.
2. **Finite knobs.** `tuning.rs` gives `mmr_lambda` and `min_coverage` the `is_finite` guard `rrf_k` already had. `a_knob_that_is_not_a_number_falls_back_to_its_default` loops NaN and both infinities, and keeps an in-range case asserting the clamp still clamps (`5.0 → 1.0`, `-1.0 → 0.0`) — so the guard added a validity check rather than replacing the range check.
3. **Torn log.** `append_workspace_private_line` opens with `read`, inspects the last byte, and writes `\n` first when one is missing. `an_unterminated_record_does_not_swallow_the_next_one` plants `{"partial"` after a whole record and asserts three distinct lines.
4. **Episode identity, both sides.** Store side: `two_turns_in_one_second_write_two_episodes` (two occurrences → 2 rows, then re-recording one → still 2) and `an_episode_without_an_occurrence_keeps_its_old_identity`. CLI side: the new `memory/tests/episode_identity.rs` with `two_executions_in_one_second_keep_their_own_episodes` and `one_execution_recorded_twice_keeps_one_episode`, both on a `FixedClock` so the same-second case is deterministic rather than reached by luck. `public_id` appends the occurrence bytes only when `Some`, so existing `epi_` ids are unchanged.
5. **Reuse race.** `upsert_attempt` re-asks `embedding_exists(&tx, …)` *inside* the write transaction and returns `UpsertAttempt::Raced(vanished)`, which the loop feeds into `force_embed` and retries, bounded by `REUSE_RACE_ATTEMPTS = 3` with the last attempt committing (so it terminates). `a_vector_reclaimed_while_embedding_is_written_again` drives the real race deterministically: `ReclaimingEmbedder` deletes the victim row through the store's own connection on the call that serves the delta, which is legitimate because the store holds no lock while embedding.
6. **Blank record named first.** `unwritable_record(&delta)` runs at the top of `upsert`, before the clock, the fingerprint, or any embedding. `a_blank_record_is_named_before_anything_is_written` asserts `InvalidInput` naming `memory` **and** `node_count() == 0`, then writes the good half alone to show dropping the bad entry at the caller is the right repair. `writable_lessons` filters on `recall_text`'s output, so `a_lesson_with_only_a_trigger_is_still_writable` passes while `a_blank_lesson_is_dropped_and_its_siblings_are_kept` keeps the siblings.
7. **Residue filed.** #5776 and #5777 both exist, are open, and carry the `triage` label and nothing else (SCR-005). Both are named in the PR description, and the code comments at each site point at them.

Box 8 (CI green) is deliberately **left unticked** — `fmt + clippy + test` is still in progress on this head. It is the only item outstanding.

No Sourcery `❌` rows: its assessment table against #5326 is three `✅`.

One thing for whoever merges rather than for this diff: #5774 removes `EpisodeInput::salience`, which this branch still passes in `writeback.rs`. Whichever lands second will need that line reconciled — textual, not a design question.

---

Issue #5780, opened 2026-09-03
Conflict detection never crosses a path scope against a task scope, so those collisions stay silent


macanderson:
## Problem

Record conflict detection compares scopes one dimension at a time. `crates/stella-core/src/records/validate.rs`'s `overlapping_scope` crosses path globs against path globs, then task/keyword triggers against task/keyword triggers (#5330). It never crosses a *path* against a *trigger*, so a record scoped `paths = ["crates/stella-core/**"]` and a record scoped `tasks = ["refactor"]` never report a conflict — even though selection is disjunctive, so one turn that both names a file under that path and says "refactor" fires both of them.

If the two carry different enforcement at equal precedence, that is the §12 governance failure the module exists to report, and it goes unreported.

This was left out of #5330 on purpose: `select::applies_this_turn` needs a concrete turn path to match a glob against, and a glob is not a turn path, so the honest form of the question is "does a turn exist that both scopes select?" — which for a path-scoped record and a task-scoped record is nearly always yes. Reporting every such pair would drown the useful conflicts. Deciding what to report is the work.

## Files

- `crates/stella-core/src/records/validate.rs` — `overlapping_scope`, `overlapping_paths`, `overlapping_triggers`.
- `crates/stella-core/src/records/select.rs` — `applies_this_turn`, `shared_triggers`; the matching semantics the answer has to agree with.
- `docs/spec/adaptive-context/context-pr.md` §12 — what a conflict is and what a report must say.

> **Path note (added 2026-09-05):** #5829 moved the record plane out of `stella-core` into the new `stella-records` crate after this issue was filed. Both files above now live at `crates/stella-records/src/records/`.

## Constraints

- The module header argues that a false positive is the cheap direction and a false negative the expensive one, but that argument was made about glob containment, where a spurious pair is rare. A rule that fires on every path-vs-task pair is a different trade and needs its own justification.
- A conflict suspends the more restrictive record's guard (`records/registry.rs`'s `resolve_guard`), so a noisy rule disarms real guards. See #5779 for what four new conflicts already cost.

## Definition of done

- [x] A decision, written in `validate.rs`'s module header: either cross-dimension pairs are reported, with the rule that keeps the report useful, or they are not, with the reason.
- [x] If reported: a witness test pair — a path-scoped `hard` record against a task-scoped advisory one flags, and whatever the rule excludes still does not flag.
- [x] If not reported: a test that pins the boundary, so the silence is a decision rather than an oversight.
[...]

Issue #5779, opened 2026-09-03
Four repo records collide on the 'witness test' trigger, and two hard guards are suspended by it


macanderson:
## Problem

Four of this repository's own published records now collide, and two `hard` guards are suspended because of it.

Conflict detection in `crates/stella-core/src/records/validate.rs` compares task and keyword triggers once path globs find nothing (#5330). Run against `.stella/rules/`, that finds four equal-precedence pairs, all of them meeting on the multi-word keyword `"witness test"` at precedence 100:

| left | right | shared trigger | suspended |
| --- | --- | --- | --- |
| `correctness-needs-proof` (soft) | `deleted-test-named-in-pr` (hard) | `witness test` | `deleted-test-named-in-pr` |
| `correctness-needs-proof` (soft) | `inv8-posture-every-axis` (hard) | `witness test` | `inv8-posture-every-axis` |
| `deleted-test-named-in-pr` (hard) | `witness-definition` (soft) | `witness test` | `deleted-test-named-in-pr` |
| `inv8-posture-every-axis` (hard) | `witness-definition` (soft) | `witness test` | `inv8-posture-every-axis` |

`deleted-test-named-in-pr` reaches the set through its task `test`, which is a whole word of the phrase `witness test` the other three carry as a keyword.

The consequence is live, not cosmetic: `records/registry.rs`'s `resolve_guard` refuses to arm the guard of any suspended record (§12 — enforcement falls back to the less restrictive behaviour until an owner resolves it). So the `hard` guards on `deleted-test-named-in-pr` and `inv8-posture-every-axis` do not arm in this workspace. `stella context validate` reports each collision as a warning and still exits 0, so CI stays green while the guards stay down.

## Why this is not fixed in the detector's own PR

A record file carries a canonical `record_hash`, and `.stella/rules/promotions.jsonl` is a hash-chained ledger. AGENTS.md's `.stella/` table says these are edited through `stella context keep` / `promote`, never by hand, and which resolution to take — a different precedence, an explicit exclusion, or a supersession link — is a policy decision about this repository's steering, not a detector change.

## Verify

```
python3 - <<'PY'   # or read .stella/rules/*.toml by hand
# precedence, enforcement.mode and applies_to.keywords for the four handles above
PY
cargo run -p stella-cli --quiet -- context validate
```

`context validate` names each conflict and the handle it suspends.

## Definition of done

- [x] An owner decides the resolution for each of the four pairs (precedence, exclusion, or supersession).
[...]

PR #5781, opened 2026-09-03
fix(stella-core): report task- and keyword-scoped record conflicts

### LINKED ISSUE #5330: Record conflict detection is path-only, so task- and keyword-scoped conflicts resolve silently
## Problem

PLAUSIBLE (traced). `crates/stella-core/src/records/validate.rs`: two records overlapping only via `tasks`/`keywords` (paths empty, `applies_to` non-empty) yield no Conflict — an equal-precedence hard-vs-none pair scoped to the same task resolves silently, the false-negative direction the module doc calls "impossible".

## Definition of done

- [x] Conflict detection falls back to task/keyword intersection when the path comparison finds nothing. — `validate.rs`'s `overlapping_scope` calls `overlapping_paths(..).or_else(|| overlapping_triggers(..))`.
- [x] The fallback uses `select`'s own matching semantics — it shares that code rather than re-implementing it. — `select::shared_triggers` builds a `TurnFacts` from one scope's own trigger and asks `applies_this_turn` about the other; `validate` unions both directions.
- [x] Witness: two equal-precedence records scoped only by the same task, with different enforcement, are reported as one conflict and the more restrictive one is suspended. — `two_records_scoped_to_one_task_conflict_with_no_path_between_them`.
- [x] Witness pair: two records scoped to disjoint tasks still produce no conflict, however their enforcement differs. — `disjoint_task_scopes_do_not_conflict_however_the_enforcement_differs`.
- [x] `cargo test -p stella-core records` green in CI. — the `fmt + clippy + test` job on PR #5781 is green, `cargo test` included.

macanderson:
## What

Record conflict detection compared path globs and nothing else. Two records whose `applies_to` carries only `tasks` or `keywords` have no paths to cross, so `overlapping_paths` reported no overlap and `conflict_between` gave up — an equal-precedence `hard`-vs-advisory pair scoped to the same task resolved silently, which is the false negative `validate.rs`'s own header calls impossible.

## Why this shape

Overlap now falls back to trigger comparison when the glob comparison finds nothing, and the fallback shares `select.rs` rather than growing a second matcher:

- `select::shared_triggers(scope, other)` — for each of `scope`'s own tasks and keywords, build `TurnFacts { text: term, paths: &[] }` and ask `applies_this_turn(Some(other), &facts)`. A turn whose text is one record's own trigger is a turn that record selects by construction, so the only question left is whether the other record selects it too, and that is answered by the code that will decide at run time whether both records fire.
- `validate::overlapping_triggers` asks both directions and unions them, because the matcher is asymmetric: `docker` is a whole word of `docker compose`, and `docker compose` is not a substring of `docker`.
- The `Conflict::detail` names the shared terms, so §12's "what overlaps, and what to do" stays actionable.

Path overlap keeps `globs_overlap` exactly as it was, and runs first, so a pair that already collides on a path keeps the more specific report.

Exemplar: none borrowed — this follows the file's own established shape (a small private helper per question, `Option<String>` carrying the human-readable description).

## Witness mechanism

`two_records_scoped_to_one_task_conflict_with_no_path_between_them` (in `validate.rs`'s test module) builds two records at precedence 50, both scoped `tasks = ["release"]` and nothing else, one `hard` and one `none`. On `main` `overlapping_paths` iterates two empty path vectors, `shared` is empty, it returns `None`, and `detect_conflicts` returns zero — the test's `assert_eq!(conflicts.len(), 1)` fails by construction. With the fallback it returns one conflict naming `release`, suspends the `hard` record, and attaches the finding to both sides.

Its pair holds the other direction: `disjoint_task_scopes_do_not_conflict_however_the_enforcement_differs` (`deploy` vs `docs`) stays empty. `a_task_on_one_record_and_the_same_word_as_a_keyword_on_the_other_conflict` covers the cross-field case, and three unit tests in `select.rs` pin `shared_triggers` itself, including the multi-word asymmetry and the path-only scope that shares nothing.

## What this surfaces in this repository's own records — read this before merging

Running the new comparison over `.stella/rules/` finds **four** equal-precedence collisions that were silent before, all of them meeting on the multi-word keyword `"witness test"` at precedence 100: `correctness-needs-proof` (soft) and `witness-definition` (soft) each against `deleted-test-named-in-pr` (hard) and `inv8-posture-every-axis` (hard). `deleted-test-named-in-pr` joins through its task `test`, a whole word of that phrase.

They are real §12 conflicts, and the consequence is live: `registry.rs`'s `resolve_guard` refuses to arm a suspended record's guard, so those two `hard` guards stop arming in this workspace until an owner resolves the corpus. `stella context validate` reports each as a warning and exits 0 (only blocking findings drive its exit code), so CI stays green. I did not repair the records here — a record carries a canonical hash, `promotions.jsonl` is hash-chained, AGENTS.md says these are edited through `stella context keep` / `promote` rather than by hand, and which resolution to take is a policy decision. Filed as #5779.

## Remaining work filed

- #5779 — the four `witness test` collisions in this repo's own corpus, and the two hard guards they suspend.
- #5780 — conflict detection still never crosses a *path* scope against a *task* scope, so those collisions stay silent. Left out on purpose: nearly every such pair can co-apply, so reporting them all would drown the useful conflicts and disarm real guards.

Tests deleted: None.

Closes #5330

## Summary by Sourcery

Report equal-precedence enforcement conflicts when records can apply to the same task or keyword-triggered turn without sharing a path.

Bug Fixes:
- Detect conflicts between equal-precedence records that share task or keyword triggers even when their path scopes do not overlap.
- Include shared task and keyword terms in conflict details so reported conflicts identify the triggering scope.

Enhancements:
- Reuse runtime record-selection matching to determine shared triggers, including task/keyword cross-field matches and asymmetric multi-word keyword matching.
- Preserve path-based overlap reporting as the preferred, more specific conflict description.

Tests:
- Add coverage for task-only conflicts, disjoint task scopes, task/keyword cross-field conflicts, asymmetric keyword matching, and path-only scopes.

### DIFF
--- a/crates/stella-core/src/records/select.rs
+++ b/crates/stella-core/src/records/select.rs
@@ -42,6 +42,11 @@
 //!   when a labelled task plane exists it belongs in [`TurnFacts`].
 //! - Multi-word **keywords** (`"docker compose"`) match as case-insensitive
 //!   substrings, because word-splitting would never reassemble them.
+//!
+//! [`super::validate`] asks the same rules a second question — can one turn
+//! select both of these records? — through [`shared_triggers`], so a conflict
+//! between two task-scoped records is judged by the matcher that will decide
+//! whether they both fire.
 
 use super::super::glob::match_glob;
 use super::super::ingest::record::AppliesTo;
@@ -79,6 +84,36 @@ pub fn applies_this_turn(applies_to: Option<&AppliesTo>, facts: &TurnFacts<'_>) 
                 word_in_text(facts.text, keyword)
             }
         })
+}
+
+/// The tasks and keywords of `scope` that also select `other`.
+///
+/// [`super::validate`] asks this to decide whether two records can steer the
+/// same turn once their paths have told it nothing. A turn whose text is one of
+/// `scope`'s own triggers is a turn `scope` selects, so all that is left to ask
+/// is whether [`applies_this_turn`] admits `other` for it — one matcher, and a
+/// list of the terms a conflict report can name.
+///
+/// Asymmetric, so a caller wanting every shared term asks both ways: a
+/// single-word keyword matches inside a multi-word one, and the reverse cannot.
+/// A path-only scope shares nothing here, because a glob is not a turn path;
+/// globs are compared to each other by `validate`.
+pub fn shared_triggers(scope: &AppliesTo, other: &AppliesTo) -> Vec<String> {
+    scope
+        .tasks
+        .iter()
+        .chain(scope.keywords.iter())
+        .filter(|term| {
+            applies_this_turn(
+                Some(other),
+                &TurnFacts {
+                    text: term.as_str(),
+                    paths: &[],
+                },
+            )
+        })
+        .cloned()
+        .collect()
 }
 
 /// Any pattern matching any turn path. A pattern with no literal part (`*`,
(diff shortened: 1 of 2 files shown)


macanderson:
## Independent DoD verification (#5330)

Four of #5330's five boxes are ticked. Each was checked against this head (`f73de65`) before ticking, not taken from the PR description:

1. **Path-first, trigger fallback.** `conflict_between` now calls `overlapping_scope`, which is `overlapping_paths(left, right).or_else(|| overlapping_triggers(left, right))` — so the trigger comparison runs only when the glob comparison found nothing, and a pair that already collides on a path keeps the more specific report.
2. **Shares `select`'s matcher.** `shared_triggers` lives in `select.rs` and answers by calling `applies_this_turn` — the same function that decides at run time whether both records fire. `validate.rs` imports it (`use super::select::shared_triggers`); no second matcher was written. `overlapping_triggers` asks both directions and unions them, which is required because the matcher is asymmetric (`docker` matches inside `docker compose`, not the reverse) — pinned by `shared_triggers_is_asymmetric_across_a_multi_word_keyword`.
3. **Conflict witness.** `two_records_scoped_to_one_task_conflict_with_no_path_between_them` builds two records at precedence 50, both `tasks = ["release"]` with empty `paths`, one `Hard` and one `None`. It asserts `conflicts.len() == 1`, that the detail names `release`, that `conflict.suspended == "release-checklist"` (the more restrictive side), and that *both* records carry a `Conflict` finding. On `main` `overlapping_paths` iterates two empty path vectors and returns `None`, so `detect_conflicts` returns zero and the length assertion fails by construction.
4. **Disjoint pair.** `disjoint_task_scopes_do_not_conflict_however_the_enforcement_differs` (`deploy` Hard vs `docs` None) asserts empty. `a_task_on_one_record_and_the_same_word_as_a_keyword_on_the_other_conflict` covers the cross-field case, and `a_path_only_scope_shares_no_trigger` pins that a glob is not treated as a turn path.

Box 5 (`cargo test -p stella-core records` green in CI) is deliberately **left unticked** — `fmt + clippy + test` is still in progress on this head. It is the only item outstanding.

No Sourcery `❌` rows: its assessment table against #5330 is three `✅`.

Worth a reviewer's attention before merge, since it is a live consequence rather than a code question: the PR description reports that the new comparison finds **four** equal-precedence collisions in this repository's own `.stella/rules/` corpus, all meeting on the keyword `witness test`, and that `resolve_guard` refuses to arm a suspended record's guard — so two `hard` guards stop arming in this workspace until an owner resolves the corpus. That is filed as #5779 and is a policy decision, not something this PR should settle.

---

PR #5784, opened 2026-09-03
fix(stella-core): make a gated probe answer to the trust tier

### LINKED ISSUE #5329: Gated probes never consult Trust, and three modules disagree where the defense lives

macanderson:
## What and why

`sweep::honored_probe` decided whether a gated probe — `command_succeeds` or `http_ok` — may run from three fields: `origin`, `truth.basis`, and `verified_by`. Every one of them is written *inside the file being judged*. A checkout's `.stella/rules/*.toml` can set all three, so anyone who could open a pull request could arm a command or a network call with a committed policy file's authority behind it. `records/bridge.rs` already refuses exactly this self-attestation before it will let a record approve its own blocking guard, and it refuses it by asking the loader-stamped `Trust`. The sweep never asked.

It asks now. A gated probe is honored only when all three hold:

- the loader stamped `Trust::User` — the one fact a file cannot write about itself, taken from the directory it was read out of;
- the origin is stamped and is not `imported` or `inferred` — a **missing** origin is now the untrusted case, which is the reading the deleted ingest-side accessor took (#5331 item 1 asked for this to be pinned);
- `truth.basis` is a `decree` with a non-empty `verified_by`.

`SweepInput` carries the tier, `registry::load` passes the one `parse_tier` already stamped, and `stella-cli`'s `run_due_probes` / `probe_everything` stop running over `TieredFiles::all()`. They parse each tier with its trust stamped instead — `TieredFiles::all()` was flattening the tiers at exactly the boundary where the answer depends on them.

`validate::gated_probe_refused` now reads the tier too, so a repository record that declares a gated probe reports `RecordFinding::GatedProbeRefused` rather than loading clean with a probe that will never run. (The issue said that finding is "defined and never produced" — that is stale against `main`: `check_one` has produced it for a while. What was wrong is that it could not see the tier, so it fired on the wrong set.)

**This makes nothing newly runnable, and closes no live hole on its own.** `ingest_cmd::probe::evaluate` abstains on every gated kind and still does. That is what kept this unexploitable, and it is why the issue is filed as a missing gate rather than as a vulnerability.

## The three comments that disagreed

- `ingest_cmd/probe.rs`'s header said gated kinds "are stripped before we get here". Two callers reach `evaluate` and they applied different rules, so that was a claim about the callers and was already false for the sweep's. It now says what this file does: a gated kind abstains here, whatever let it through.
- `context_cmd/review.rs` said the sweep admits a gated probe "on a human-decreed record". It now says what the sweep actually admits, and why a proposal — which has no tier at all — cannot be judged by that rule on the review side.
- `sweep.rs`, `records.rs`'s `GatedProbeRefused` doc and its `Display` text, `ingest/record.rs`'s `ProbeKind` doc, and the two `context-record-examples` documents all name the tier now.

## Witness mechanism

Three, at three levels. The first two fail on old code by construction (the signature gained an argument); the third and fourth fail on old code at **runtime**, which is the honest ones:

- `sweep.rs` `no_gated_probe_is_honored_at_project_trust_ever` — the record asserts everything a gated probe needs, across every `Origin` including `None`, and is refused at `Trust::Project` every time. `an_unstamped_origin_does_not_unlock_a_gated_probe` pins the `origin: None` case at `Trust::User`.
- `records/registry/tests.rs` `a_repository_record_never_arms_a_gated_probe` — the same record loaded as a **project** file through `registry::load` must carry `GatedProbeRefused`. On `main` `honored_probe` returns `Some` for it, no finding is emitted, and the assertion fails. Its pair, `the_same_record_in_the_users_own_directory_keeps_its_probe`, shows the finding is about the directory and not about the record.
- `context_records/tests.rs` `the_sweep_never_takes_up_a_repository_records_gated_probe` — the CLI sweep over the project tier returns `ran == 0` and writes no cache entry. On `main` `probe_is_runnable` is true for this record, `never_checked` fires, `evaluate` abstains, and an `unfalsifiable` verdict is cached: `ran == 1`. That difference is observable without making anything runnable.

## Exemplar

No new shape was invented. `records/registry.rs`'s `parse_tier` is the exemplar for stamping a tier at the boundary that knows the directory, and `records/bridge.rs`'s `self_attested` is the exemplar for the rule itself — a record's own fields cannot establish the authority they claim.

Tests renamed: `sweep.rs`'s `an_ungated_probe_is_honored_on_an_imported_record` is now
`an_ungated_probe_is_honored_on_an_imported_project_record`. Same assertion, and the new
name says the half that now matters: an ungated probe is honored on an imported record
*published by the repository*, which is how an imported claim gets refuted at all.

Tests deleted: None.

Residue issues filed: none — see the final comment on this PR if that changes.

Closes #5329

## Summary by Sourcery

Require gated probes to be published from the user's trusted rules directory and satisfy explicit origin and human-verification requirements before they are honored.

Bug Fixes:
- Prevent repository-published records from authorizing gated command or network probes based solely on self-attested fields.

Enhancements:
- Propagate loader-assigned trust tiers through sweep, validation, and registry processing so gated probes are evaluated against their publishing location.
- Report refused gated probes on records that cannot satisfy the trust, origin, and verification requirements.
- Clarify gated-probe behavior and trust requirements across code documentation and context-record examples.

Documentation:
- Update context-record documentation to describe missing origins and repository-tier restrictions for gated probes.

Tests:
- Add coverage confirming project records never arm gated probes, unstamped origins remain untrusted, and user-tier records retain eligible probes.

### DIFF
--- a/crates/stella-cli/src/context_cmd/review.rs
+++ b/crates/stella-cli/src/context_cmd/review.rs
@@ -164,15 +164,16 @@ fn rank(verdict: Option<&str>) -> u8 {
 ///
 /// **A gated probe is never run here, whatever the record says about itself.**
 /// The sweep's rule ([`stella_core::records::honored_probe`]) admits a
-/// `command_succeeds` or `http_ok` on a human-decreed record, and it can,
-/// because it reads a *published* file that a reviewer has already accepted.
-/// This reads a **proposal**, whose `origin` may sit in the file's `[defaults]`
-/// header rather than on the record — so the trust question this side of the
-/// review cannot be answered from the record alone, and the safe answer to a
-/// question you cannot answer is no. Review is a read of the tree; it does not
-/// become the place a document nobody has reviewed yet gets to run a command or
-/// reach a host. The remaining kinds are filesystem reads, which is what makes
-/// re-asking on every review cheap enough to be unconditional.
+/// `command_succeeds` or `http_ok` only on a record the loader stamped
+/// [`stella_core::records::Trust::User`] — a *published* file in the user's own
+/// `~/.stella/rules`, which nobody else can write. This reads a **proposal**,
+/// which has no tier at all: it is sitting in the tree awaiting a decision, and
+/// its `origin` may live in the file's `[defaults]` header rather than on the
+/// record. The safe answer to a question you cannot answer is no. Review is a
+/// read of the tree; it does not become the place a document nobody has reviewed
+/// yet gets to run a command or reach a host. The remaining kinds are filesystem
+/// reads, which is what makes re-asking on every review cheap enough to be
+/// unconditional.
 fn reprobe(root: &Path, found: &FoundProposal, now: &str) -> Probed {
     use stella_core::ingest::record::ProbeKind;
 
--- a/crates/stella-cli/src/context_cmd/validate.rs
+++ b/crates/stella-cli/src/context_cmd/validate.rs
@@ -121,7 +121,7 @@ pub fn run_validate(root: &Path, format: QueryFormat) -> Result<(), String> {
 
     let files = rule_files(root, true, true);
     let now = now_rfc3339();
-    let cache = probe_everything(root, &files.all(), &now);
+    let cache = probe_everything(root, &files, &now);
     let registry = registry_with_cache(root, &files, &cache, &now);
 
     // Regulated tier: a project record armed to block must hold a ledger
--- a/crates/stella-cli/src/context_records.rs
+++ b/crates/stella-cli/src/context_records.rs
@@ -549,12 +549,25 @@ pub(crate) struct TieredFiles {
 }
 
 impl TieredFiles {
-    /// Every file, tier forgotten — for the passes that only ask what is on disk,
-    /// like running due probes.
+    /// Every file, tier forgotten — for the passes that only ask what is on disk.
+    ///
+    /// Not for the probe pass. Running a due probe asks whether a gated probe may
+    /// run, which `stella_core::records::honored_probe` answers from the tier, so
+    /// the sweep reads `tiered` instead. Flattening there let a repository
+    /// record be swept as if the user had published it.
     pub(crate) fn all(&self) -> Vec<RuleFile> {
         let mut all = self.user.clone();
         all.extend(self.project.iter().cloned());
         all
+    }
+
+    /// Every file with the tier it was read out of, user first — the shape a pass
+    /// needs when the answer depends on who published the record.
+    fn tiered(&self) -> [(&[RuleFile], Trust); 2] {
+        [
+            (self.user.as_slice(), Trust::User),
+            (self.project.as_slice(), Trust::Project),
+        ]
     }
 }
 
@@ -595,7 +608,7 @@ fn load_registry_with(root: &Path, include_project: bool) -> Registry {
     let files = rule_files(root, true, include_project);
     let now = now_rfc3339();
     let mut cache = SweepCache::read(root);
-    let ran = run_due_probes(root, &files.all(), &mut cache, &now);
+    let ran = run_due_probes(root, &files, &mut cache, &now);
     if ran > 0 {
         cache.write(root);
     }
@@ -708,7 +721,7 @@ fn repo_owned_lineages(files: &TieredFiles) -> BTreeSet<String> {
         .filter(|file| file.contributed_by.is_none())
         .cloned()
         .collect();
-    parse_records(&repo_owned)
+    parse_records(&repo_owned, Trust::Project)
         .into_iter()
         .map(|loaded| loaded.record.lineage_id)
         .collect()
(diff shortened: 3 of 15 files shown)


macanderson:
Automated PR sweep: the DoD gate on this PR was red because every box on #5329 was still unticked ~15 minutes after the PR was opened. I verified all seven against the diff on `58e380b` and ticked them on the issue. What established each:

1. **`honored_probe` takes the stamped `Trust` and refuses below `Trust::User`** — `records/sweep.rs`: signature is now `honored_probe(record: &Record, trust: Trust)` with an early `if trust != Trust::User { return None; }` placed after the ungated short-circuit, so ungated kinds are unaffected.
2. **An unstamped `origin` is the untrusted case** — `trusted_origin` is `record.origin.is_some_and(|o| !origin_is_untrusted(o))`; `is_some_and` on `None` is `false`. Pinned by `an_unstamped_origin_does_not_unlock_a_gated_probe`.
3. **`SweepInput` carries the tier and every caller passes the stamped one** — the field is added; `registry::load` passes `record.trust`; `run_due_probes` and `probe_everything` now take `&TieredFiles` and iterate `tiered_records`, and `TieredFiles::all` grew a doc comment saying it is *not* for the probe pass. The two `files.all()` call sites in `context_records.rs` and the one in `context_cmd/validate.rs` are all converted.
4. **Witness: no gated probe honored at project trust, ever** — `no_gated_probe_is_honored_at_project_trust_ever` loops all six `Origin` values (plus `None`) at `Trust::Project`.
5. **Witness: refused end to end** — two halves, both present: `a_repository_record_never_arms_a_gated_probe` (registry reports `GatedProbeRefused`) and `the_sweep_never_takes_up_a_repository_records_gated_probe` (asserts `ran == 0` *and* `cache.checked.is_empty()`, so nothing is cached either).
6. **The three modules' comments agree** — `sweep.rs`'s `honored_probe` doc, `ingest_cmd/probe.rs`'s module header (the false "stripped before we get here" claim is gone), and `context_cmd/review.rs`'s `reprobe` doc are all rewritten to the tier-based mechanism.
7. **`GatedProbeRefused` is produced where refused** — `validate.rs`'s `gated_probe_refused` now takes `trust` and asks `honored_probe(record, trust)`. Pinned by `the_same_record_at_project_trust_is_reported_instead`.

I have **not** run the test suite — that is CI's job here, and `fmt + clippy + test` is still in flight on this head. These ticks assert the code and doc facts the items name, not a green suite. If CI comes back red, the ticks for items 4, 5 and 7 are the ones to re-examine first, since they are the ones a compile or assertion failure would falsify.

I did not push to this branch.

---

macanderson:
Remaining work filed after the merge, both noticed while writing this change and neither in scope for it:

- #5798 — a user-published gated probe is admitted by `honored_probe` and then never run, because `ingest_cmd::probe::evaluate` abstains on every gated kind. The honored path this PR tightened grants something no runner delivers. Two ways out, and they need opposite work; the choice is a maintainer's.
- #5799 — `bridge.rs`'s `self_attested` and `sweep.rs`'s `honored_probe` now write the same self-attestation rule twice. Not identical (the probe gate adds the origin condition), so not a mechanical de-duplication — a question about whether the shared half gets a name.

Sourcery's assessment table came back ✅ on all three linked objectives, so nothing there is outstanding.

PR #1680, opened 2026-08-06
feat(stella-fleet): a check-and-set dispatch lease so two sessions cannot take one task (#1136)

macanderson:
Closes #1136.

## The gap

Path claims answer *"may I write this file?"*. Nothing answered *"is another session already doing this work?"* — so two sessions in one workspace could scope the same unit of work, read the same files, do the same fix, and land the second as an empty merge. That is the incident #1136 was filed from (PR #1098/#1099, 27 seconds apart, the second a zero-file-diff merge), and its comment adds four more collisions from a single afternoon.

## What this adds

`crates/stella-fleet/src/ledger/lease.rs` — one row per **unit of dispatch** in `fleet.db`, keyed by an opaque `claim_key` (`task:<id>` here; `issue:<n>` reserved for a tracker-driven dispatcher, so the mechanism never has to know what a unit of work is).

**The claim is one statement, and the affected-row count decides the winner:**

```sql
INSERT INTO dispatch_claims (...) VALUES (...)
ON CONFLICT(claim_key) DO UPDATE SET owner = excluded.owner, fence = dispatch_claims.fence + 1, ...
WHERE dispatch_claims.expires_at_ms <= ?now
RETURNING fence, acquired_at_ms
```

A read-then-write in Rust would leave open exactly the window this exists to close, and the race witness below proves that is not theoretical: swapping the statement for a read-then-write makes it **8 winners out of 8**.

## Lease semantics (the part to review)

- **Expiry.** Every claim carries `expires_at_ms`, inclusive. Reclaiming a dead holder’s key is a side effect of asking for it — expiry is evaluated *inside* the claim statement — so there is no reaper, no daemon, no `--force`, and a SIGKILLed session cannot wedge a task. Default TTL is **15 minutes** (`DISPATCH_LEASE_TTL`), sized to cover the longest pause a *live* dispatch can take between heartbeats (a swapping machine, a closed laptop, a `SIGSTOP`), not the length of a task.
- **Renewal.** A live attempt heartbeats every **TTL/3** (so two consecutive missed beats still do not expire it) for as long as its worker runs. The heartbeat rides in the dispatch future via `tokio::select!` rather than a spawned task: nothing `'static` is needed, it cannot outlive what it is proving alive, and a cancelled dispatch stops beating by construction. Renewal is conditional on key + owner + fence + *still live*, so a holder that let its TTL lapse must re-claim rather than resurrect.
[...]

Issue #1136, opened 2026-08-01
Two agent sessions can claim the same issue: no dispatch-level claim or lease


macanderson:
## What

Nothing stops two agent sessions from picking up the **same issue** at the same time and doing the same work. It happened on 2026-08-01 and cost a complete duplicate agent run.

Evidence:

| | PR #1098 | PR #1099 |
|---|---|---|
| Title | `fix(cli): panic-safe env restore for HOME/STELLA_DATA_DIR tests (#911)` | `fix(cli): panic-safe env restore for HOME/STELLA_DATA_DIR tests` |
| Merged | 08:07:20Z | 08:07:47Z (27s later) |
| Merge commit | `0008c25f` | `a3667e69` |
| Files changed | 13 | **0 — merged as an empty commit** |

Two sessions independently scoped issue #911, read the same ~12 files, designed the same fix (promote `EnvRestore` into a shared module), ran the same full-workspace pre-push gate, and opened near-identical PRs. The second merged as a no-op because the first had already landed the identical content.

## Why it matters

The wasted work is the whole second run, not a fraction of it:

- **Tokens** — a full scoping sweep plus a 13-file migration, roughly six figures of tokens, producing zero net diff.
- **Wall clock** — two full `make gate` runs (workspace `cargo doc` + `clippy --all-targets` + full test suite; see #1135), each several minutes, plus two complete CI runs.
- **History noise** — an empty merge commit on `main` whose message describes changes it does not contain. Anyone bisecting or reading `git log` sees the same fix land twice.
- **It scales with fleet width.** The wider the fan-out, the higher the collision probability. This is the failure mode that makes "just run more agents" stop paying off.

## Why the existing claim machinery does not cover this

Do not close this as a duplicate of #613 / #562. Those cover stella-fleet's `ClaimGuard` / stella-store `file_locks`, which claim **file paths** so two workers within one plan don't write the same file. That mechanism has no notion of an issue and does not participate in work *dispatch*. #832 ("stella-maintains-stella") describes worktree isolation, which is filesystem collision-safety — also not dispatch dedup.

The gap is specifically: **before an agent starts work on issue N, nothing checks whether another agent is already on issue N.**

## Proposed design

A check-and-set claim on the unit of dispatch (the issue), with a TTL so a crashed session doesn't wedge the issue forever.

[...]

PR #5791, opened 2026-09-03
feat(stella-cli): mirror self-driving issue claims onto GitHub

### LINKED ISSUE #1675: stella fleet: no human-visible surface for dispatch claims (and no GitHub mirror)
The dispatch-claim lease landed in `crates/stella-fleet/src/ledger/lease.rs` (#1136): `Ledger::claim_dispatch` / `renew_dispatch` / `release_dispatch` / `dispatch_claim` / `live_dispatch_claims`, wired into `Fleet::dispatch` (`crates/stella-fleet/src/fleet.rs`). It closes the race, but it only closes it for two `stella fleet` processes in one workspace, and it satisfies **one** of #1136's acceptance bullets only partially: *"The claim is visible to a human deciding what to hand out next."*

Today a human can see a claim only by opening `.stella/private/fleet.db` with `sqlite3`. #1136 explicitly recommended GitHub as the claim store for exactly this reason — "a claim they cannot see is a claim they will collide with manually" — and the field evidence in its comment is about *agent sessions working GitHub issues*, not about fleet tasks.

## Two halves, both missing

1. **A read surface.** `stella fleet claims` (or a `--claims` view on the existing dashboard) listing `live_dispatch_claims(now)`: key, owner, when it was taken, when it lapses. Read-only, offline, no API key. `crates/stella-cli/src/fleet_cmd.rs` is a grandfathered god file **closed to growth** (AGENTS.md § God files) — put the command in a sibling module the way `crates/stella-cli/src/fleet_gc.rs` (PR #1657) does, with only the dispatch line touching existing files. A new subcommand needs a docs entry in `website/content/docs/commands/fleet.mdx` or `make command-docs` fails.
2. **A GitHub mirror for issue-scoped work.** The lease key is deliberately opaque (`task:<id>` today, `issue:<n>` reserved). A session that claims `issue:N` locally should also make that visible where humans dispatch from — assign self plus an `in-progress` label, or a claim comment carrying the same expiry — with the **local ledger still the arbiter**, since GitHub gives no compare-and-set. Note the local ledger is per-workspace, so it does not coordinate two clones on two machines; the GitHub mirror is what would.

## Definition of done

- [x] `stella fleet claims` (or equivalent) prints live claims and their expiry, with a witness test. (shipped in #1680)
- [x] A documented path for an issue-driven session to take `issue:<n>` and have that visible on GitHub, with the lease semantics from `lease.rs` (expiry + renewal + fence) preserved. (shipped in #5791)
- [x] `make command-docs` green; no new lines in `fleet_cmd.rs`. (true for #1680; this second half touches no `fleet_cmd.rs` line either)

Refs #1136.

macanderson:
## What / why

`self_driving_cmd::claim` already takes a local dispatch lease under `issue:<n>` (`stella_fleet::Ledger::claim_dispatch`) before `stella self-driving drive` (serial or `--parallel`) works an issue. That closed the race between two loops sharing one clone (#4300), but the only way a human — or a session on a second clone — could see the claim was to open `.stella/private/fleet.db` with `sqlite3`. #1675 is the second half of #1136's "the claim is visible to a human deciding what to hand out next" acceptance bullet; the first half (`stella fleet claims`) shipped in #1680.

## What changed

A new `MirroredLease` type (`crates/stella-cli/src/self_driving_cmd/claim_mirror.rs`) wraps the existing `claim::Lease`:

- **On grant**, it posts one comment on the issue naming the owner, the lease's actual TTL, and `stella fleet claims --all` for live status.
- **On drop** — the exact moment the wrapped lease itself releases the SQL row — it posts a second comment saying the issue is free again.

This needed no new `IssueProvider` capability: it reuses the `comment` port and the `backlog::comment` helper (`stella_autonomy::sign` + `cfg.attribution.issue_comment`) the loop's escalation and closure paths already use.

Wired at the two places that grant a lease today — `drive`'s serial loop (`leases: HashMap<String, MirroredLease>`) and `parallel`'s wave dispatch (`taken: Vec<(Issue, Option<MirroredLease>)>`) — by changing the stored type from the bare `Lease` to `MirroredLease`. Every existing release site (the end of a scope, an early `continue` inside `drive`'s `Work` arm, the bulk `drop(taken)` in `parallel::wave`) already drops the lease at the right moment, so wrapping it needed no new call sites.

**GitHub is a mirror, never an arbiter.** The lease still decides who owns the work — nothing here reads GitHub to make that decision, and it offers no compare-and-set to read anyway. A comment that fails to post (`gh` offline, unauthenticated, rate-limited) is dropped rather than treated as a reason to stop working the issue, matching the fail-open posture `claim::acquire` already documents for the ledger side. The claimed comment states the lease's own TTL rather than promising to stay live (no per-heartbeat re-posting, to avoid comment spam on a lease renewed every couple of minutes) — a reader who finds it later can tell it is stale on sight.

Went with a claim comment rather than "assign self + an `in-progress` label" — the issue's other offered option — because `IssueProvider` has no assign capability today and adding one is a separate, bigger change; a comment is the DoD's own alternative and reuses machinery that already ships. Full design posted as a comment on #1675 before implementation.

## Witness

`claim_mirror::tests::a_lease_posts_a_claim_comment_and_a_release_comment_around_its_life` — a hermetic test: a real tempdir-backed dispatch lease (`claim::acquire_as`, no mock), a fixture `IssueProvider` that only records what it is told (no network), asserting the wrapper posts exactly one "claimed" comment on construction and exactly one "released" comment on drop, both bodies checked verbatim. Fails on `main` by construction — `MirroredLease` does not exist there.

Three more unit tests cover the comment bodies directly (owner + whole-minute TTL + the `stella fleet claims --all` pointer; sub-minute TTL rounds up to "1 minute" rather than "0"; the release comment names the owner).

## Docs

Two sentences added to `website/content/docs/commands/self-driving.mdx`, next to the existing paragraphs describing `drive`'s claim/work/deliver loop and the parallel wave's per-worker lease.

## Tests deleted

None.

Closes #1675
Refs #1136

## Independent DoD verification (#1675)

The `dod` gate went red at 18:27 on #1675's one remaining unticked box:

> A documented path for an issue-driven session to take `issue:<n>` and have that visible on GitHub, with the lease semantics from `lease.rs` (expiry + renewal + fence) preserved.

It is ticked now, checked against this head (`25d4ede`) in three parts:

1. **Visible on GitHub.** `claim_mirror.rs`'s `MirroredLease::new` posts the claim comment on grant and its `Drop` posts the release comment, at both grant sites — `drive.rs`'s `leases` map and `parallel.rs`'s `taken` vec. The witness asserts both bodies verbatim against a recording `IssueProvider`, with no network.
2. **A documented path.** `website/content/docs/commands/self-driving.mdx` gains a paragraph stating what the two comments are, that the workspace lease in `.stella/private/fleet.db` is what actually decides ownership, that `stella fleet claims --all` shows it live, and that a failed post never stops the loop. The `--parallel` paragraph is updated to say each worker's lease is mirrored the same way.
3. **Lease semantics preserved.** `MirroredLease` *wraps* `Lease` rather than replacing it, and adds no path that renews or releases — the new `Lease::dispatch_lease` accessor is `pub(super)` and read-only. Expiry, the heartbeat renewal, and the fence are untouched in `lease.rs`, and the inner `Lease::drop` still runs at the same moment it always did.

Also checked, since the mirror posts from a `Drop`: `post()` builds a current-thread runtime and `block_on`s it, which would panic inside an existing runtime. It cannot be. `self_driving_cmd::run` is reached synchronously from `fn main` (`main.rs`'s `fn run`), and in `parallel::wave` the `taken` vec holding the leases is declared *before* the multi-thread `runtime`, so it drops after that runtime does — never inside its `block_on`. The build-a-runtime-per-call shape is also the established idiom in the sibling `backlog.rs`.

## Summary by Sourcery

Make self-driving issue leases visible on the issue tracker without changing local lease ownership semantics.

New Features:
- Mirror self-driving issue lease claims and releases to the issue tracker with owner, lease duration, and live-status guidance.

Enhancements:
- Preserve local lease ownership, expiry, renewal, and fencing as the authoritative dispatch mechanism while making tracker comments best-effort.
- Apply lease mirroring consistently to serial and parallel self-driving execution paths.

Documentation:
- Document tracker-visible claim and release comments, local lease authority, live claim inspection, and fail-open posting behavior for self-driving workflows.

Tests:
- Add hermetic and unit coverage for claim and release comment creation, lease duration formatting, and owner attribution.

### DIFF
--- a/crates/stella-cli/src/self_driving_cmd.rs
+++ b/crates/stella-cli/src/self_driving_cmd.rs
@@ -17,6 +17,7 @@ mod audit;
 mod backlog;
 mod budget;
 pub(crate) mod claim;
+mod claim_mirror;
 mod claude_worker;
 // `pub(crate)` rather than private: `plugin_cmd::configure` asserts against
 // this reader directly (#3999). A package that configures attribution has to
--- a/crates/stella-cli/src/self_driving_cmd/claim.rs
+++ b/crates/stella-cli/src/self_driving_cmd/claim.rs
@@ -113,6 +113,20 @@ pub(crate) struct Lease {
     beat: Option<JoinHandle<()>>,
 }
 
+impl Lease {
+    /// The grant this lease holds — owner, fence, expiry — for a caller that
+    /// wants to show a human what it is rather than only hold it.
+    ///
+    /// [`super::claim_mirror`] is the one caller today: it reads `owner` and
+    /// `ttl_ms` to build the comment it posts on the tracker when the lease
+    /// is granted. Read-only — nothing outside this module may renew or
+    /// release a lease it does not own.
+    #[must_use]
+    pub(super) fn dispatch_lease(&self) -> &DispatchLease {
+        &self.lease
+    }
+}
+
 impl Drop for Lease {
     fn drop(&mut self) {
         self.stop.store(true, Ordering::Relaxed);
--- a/crates/stella-cli/src/self_driving_cmd/claim_mirror.rs
+++ b/crates/stella-cli/src/self_driving_cmd/claim_mirror.rs
@@ -41,11 +41,14 @@ use super::claim::Lease;
 /// the codebase — the end of a scope, an early `continue`, a bulk `drop` —
 /// mirrors for free. See the module docs for why the mirroring half is
 /// best-effort.
+///
+/// The wrapped lease is read again at drop time (its `owner`, off
+/// [`Lease::dispatch_lease`]) rather than copied out at construction, so
+/// there is exactly one place this type learns who it was granted to.
 pub(super) struct MirroredLease<'p> {
     lease: Lease,
     provider: &'p dyn IssueProvider,
     key: IssueKey,
-    owner: String,
     signature: String,
 }
 
@@ -61,19 +64,20 @@ impl<'p> MirroredLease<'p> {
         number: &str,
         signature: &str,
     ) -> Self {
-        let dispatch = lease.dispatch_lease().clone();
         let key = IssueKey::from(number);
-        post(
-            provider,
-            &key,
-            &claimed_body(&dispatch.owner, dispatch.ttl_ms),
-            signature,
-        );
+        {
+            let dispatch = lease.dispatch_lease();
+            post(
+                provider,
+                &key,
+                &claimed_body(&dispatch.owner, dispatch.ttl_ms),
+                signature,
+            );
+        }
         Self {
             lease,
             provider,
             key,
-            owner: dispatch.owner,
             signature: signature.to_owned(),
         }
     }
@@ -81,10 +85,11 @@ impl<'p> MirroredLease<'p> {
 
 impl Drop for MirroredLease<'_> {
     fn drop(&mut self) {
+        let owner = self.lease.dispatch_lease().owner.clone();
         post(
             self.provider,
             &self.key,
-            &released_body(&self.owner),
+            &released_body(&owner),
             &self.signature,
         );
     }
--- a/crates/stella-cli/src/self_driving_cmd/claim_mirror.rs
+++ b/crates/stella-cli/src/self_driving_cmd/claim_mirror.rs
@@ -1,0 +1,272 @@
+//! Mirrors a local dispatch claim onto the issue tracker.
+//!
+//! [`super::claim`] closes the race between two `stella self-driving drive`
+//! processes sharing one clone. It does this with a lease, kept in
+//! `.stella/private/fleet.db`. That lease works for two local processes.
+//! It does not help a human. A human deciding what to hand out next
+//! cannot see the lease without opening the ledger by hand. Neither can a
+//! session on a second clone.
+//!
+//! [`MirroredLease`] wraps a granted [`super::claim::Lease`] and posts one
+//! comment on the tracker when the lease is granted, and a second when it
+//! is given back. A human reading the issue then sees the same two events
+//! a peer reading the ledger would see.
+//!
+//! # A mirror, never an arbiter
+//!
+//! GitHub has no compare-and-set. So nothing here decides a claim. The
+//! lease still does, exactly as `lease.rs` describes. A comment that fails
+//! to post — `gh` offline, no login, a rate limit — is dropped rather than
+//! treated as a reason to stop working the issue. [`super::claim::acquire`]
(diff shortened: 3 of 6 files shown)


macanderson:
`fmt + clippy + test` is red on `25d4ede` ([job 100766017761](https://github.com/macanderson/stella/actions/runs/33790523225/job/100766017761)). One error, and it is not a test:

```
error: field `lease` is never read
  --> crates/stella-cli/src/self_driving_cmd/claim_mirror.rs:45:5
   |
44 | pub(super) struct MirroredLease<'p> {
   |                   ------------- field in this struct
45 |     lease: Lease,
   |     ^^^^^
   = note: `-D dead-code` implied by `-D warnings`
error: could not compile `stella-cli` (bin "stella") due to 1 previous error
```

Worth flagging where it sits: the error is at log line ~942 of 13837, and the Actions job-log API only serves the last 5000 lines, so `get_job_logs` returns a completely clean tail for this job — every later step passed, including `test-self-driving.sh` and the impacted-crates harness (`passed=25 failed=0`). Reading the tail alone would say this job is green.

The field is correct as designed — it is held for its drop, exactly as the struct's own doc comment says ("Dropping this drops the lease at the exact moment it always dropped"). `self.lease` is never read anywhere: `Drop` reads only `provider`, `key`, `owner` and `signature`, and the `.lease` hits elsewhere in `self_driving_cmd/` are `claim.rs`'s own guard, a different type. So `dead_code` is right on the letter and wrong about the intent.

Suppressing it is closed off by this repo's own gate rather than by taste: `scripts/dead-code-allows-baseline.txt` carries only `stella-core 2`, so `stella-cli` must be at zero, and `make dead-code-allows-update` "refuses to raise a number or to add a crate" — an `#[allow(dead_code)]` or `#[expect(dead_code)]` here turns `make dead-code-allows` red instead.

That leaves the rename, which is the conventional spelling for an RAII guard field and needs no suppression:

```rust
pub(super) struct MirroredLease<'p> {
    _lease: Lease,
```

plus the matching `_lease,` in `new`'s struct literal. Keep it as the first field so drop order is unchanged — `Drop for MirroredLease` still posts the released comment before the lease itself drops, which is the ordering the module docs rely on.

Not pushed: this is not my branch to write to, and the fix is one line plus its initializer.

---

PR #5793, opened 2026-09-03
fix(stella-cli): close a fork/exec fd-inheritance race in credential_handoff tests

### LINKED ISSUE #2603: stella-cli: credential_handoff fd-ownership assertion flakes under parallel load — a second race the existing comment does not cover
## Problem

`credential_handoff::tests::memory_hardening_precedes_read_and_closes_fd_on_failure`
fails under a loaded `make gate` and passes in isolation, producing a false red
gate. Observed on 2026-08-09 during a `cargo test --workspace` run at load
average ~14 on a 10-core machine:

```
thread 'credential_handoff::tests::memory_hardening_precedes_read_and_closes_fd_on_failure'
panicked at crates/stella-cli/src/credential_handoff.rs:438:9:
  left: 1
 right: -1
```

Re-running the same test alone: `9 passed; 0 failed`.

## Why it is racy, and why the existing fix does not cover it

The assertion is at `crates/stella-cli/src/credential_handoff.rs:438`:

```rust
// The fail-closed path still consumed ownership of the read end. Probe
// the open-file description, not the freed fd *number*: with the last
// reader gone, writing to the retained write end fails with EPIPE.
// (A raw `fcntl(read_fd, F_GETFD) == -1` check would be racy under the
// parallel test harness — a sibling test can reuse the number the
// instant it is freed, and `fcntl` would then report that fd's flags.)
```

The comment records a race that was already fixed — fd *number* reuse — and the
EPIPE probe does fix that one. It does not fix this one. `write` returning `1`
instead of `-1` means the pipe still had **a reader**, not that the fd number
was recycled. The words "with the last reader gone" are the load-bearing
assumption, and it is the assumption that fails.

The likely mechanism is **fd inheritance across a concurrent `fork`/
`posix_spawn`**: `stella-cli`'s test binary runs its tests on parallel threads,
several of which spawn child processes. A child forked in the window between
this test creating the pipe and `read_and_close_fd_with_hardener` closing the
read end inherits the read end, and keeps the open-file description alive after
this test drops its own copy. The write then succeeds and returns 1.

That is consistent with every symptom: load-dependent, isolation-clean, and
off-by-exactly-one-reader rather than an error.

## Reproduce

```bash
# Reliable under load; the workspace run is what surfaced it.
cargo test --workspace
# Clean in isolation:
cargo test -p stella-cli --bins credential_handoff::
```

Loading the machine (a parallel `cargo build`, or the `custom::` spawn tests in
#2011 running concurrently) makes it more likely.

## Where

- `crates/stella-cli/src/credential_handoff.rs:420-445` — the test and the
  comment describing the *previous* race
- `read_and_close_fd_with_hardener` — the function under test

## Candidate fixes

1. **Create the pipe with `O_CLOEXEC`** (`libc::pipe2(.., libc::O_CLOEXEC)`) so
   no `exec`ing child can inherit the read end. This is the smallest change and
   is what production code should be doing anyway; note it does not help a
   `fork` that never `exec`s.
2. **Assert on the observable contract rather than the fd**: the function's
   promise is "no credential bytes were read", which the
   `assert_eq!(err, "synthetic hardening failure")` above already proves. The
   fd-ownership assertion could move to a serialized test
   (`#[serial]`-style, or a dedicated single-threaded integration test) rather
   than racing the whole binary.
3. Both — (1) is a correctness improvement independent of the test.

Prefer (1) plus keeping the assertion, so the test still guards ownership.

## Constraints

- Related but distinct: **#2011** (`stella-tools`' `custom::` tests flake on a
  5s spawn budget under parallel load). Same *class* of problem — a false red
  gate caused by parallel-test contention — different mechanism. Fixing one
  does not fix the other, but whoever picks up one should read the other.
- Do not "fix" this by raising a timeout or by marking the test `#[ignore]`;
  the assertion is guarding a real fail-closed property of credential handling.

## Definition of done

- [x] The pipe every `credential_handoff` test opens is created close-on-exec
      (`cloexec_pipe`, `pipe2(O_CLOEXEC)` on Linux, `pipe()` + `fcntl(F_SETFD)`
      elsewhere), so a sibling test's `fork`/`exec` can no longer inherit it.
      Verified against PR #5793's diff on `6f6e4d8`: `cloexec_pipe` exists in
      both platform arms, and every remaining pipe-creating test in the module
      calls it — `inherited_pipe_is_consumed_framed_and_closed`, the
      `EMBEDDING_CREDENTIAL_TARGET` test, and
      `memory_hardening_precedes_read_and_closes_fd_on_failure`. No raw
      `libc::pipe(` call survives in the module.
- [x] The comment at the fd-ownership assertion in
      `memory_hardening_precedes_read_and_closes_fd_on_failure` names both
      races: the fd-number-reuse race it already covered, and the
      fd-inheritance-across-fork/exec race this issue is about.
      The existing fd-number-reuse paragraph is kept and a second paragraph is
      added naming the inheritance race and pointing at `cloexec_pipe` as what
      closes the window.
- [x] A deterministic witness test reproduces the exact race (a real child
      spawned while the pipe is open, proving its copy is closed on exec) —
      fails on a plain `libc::pipe()` and passes with `cloexec_pipe`.
      `cloexec_pipe_closes_a_concurrent_childs_inherited_copy` spawns a real
      `/bin/sh -c "sleep 5"` while the pipe is open, closes the test's own read
      end, and asserts both `write == -1` and `EPIPE`. It is deterministic
      rather than timing-dependent because `Command::spawn()` returns only
      after the child's `execve()` completes. The fail-side is the PR author's
      reported observation (reverting the non-Linux arm to a plain `pipe()`
      produced `left: 1, right: -1`), not something re-run in this audit.
- [x] `cargo test -p stella-cli --bin stella credential_handoff::` is green,
      and CI's `test` gate step is green on both the Linux and (via
      `macos-check.yml`) Darwin compile paths.
      Both are green on `6f6e4d8`: `the macOS-only dependency table still
      compiles` (Darwin, `cargo check --all-targets`, type-checks the
      `pipe()`+`fcntl` arm) and `fmt + clippy + test`
      (https://github.com/macanderson/stella/actions/runs/33793775068/job/100776564099,
      Linux, `cargo test --workspace`, runs the `pipe2` arm and every
      `credential_handoff` test) both passed. Locally on macOS,
      `cargo test -p stella-cli --bin stella credential_handoff::` ran all 14
      tests in the module and passed, exercising the `pipe()`+`fcntl` arm at
      runtime too, not just at compile time.

macanderson:
## What & why

`credential_handoff::tests::memory_hardening_precedes_read_and_closes_fd_on_failure`
flakes under `cargo test`'s parallel harness (#2603). The test's fd-ownership
assertion writes to the retained pipe write end and expects `EPIPE` — proof
the read end was really closed. A sibling test's `fork`/`exec`, landing in the
window between this test's pipe creation and the hardener's close, could
inherit the read end and keep a live reader around after this test dropped
its own copy: the write then returns `1` instead of `-1`. The comment at the
assertion already named a *different*, previously-fixed race (fd-number
reuse) and did not cover this one.

Per the maintainer's design pointer on the issue (candidate fix 1): create
every pipe these tests open with `O_CLOEXEC` already set, so a concurrently
exec'd child can never hold a copy of it. `read_and_close_fd`/
`read_and_close_fd_with_hardener` never create the pipe themselves — the
external launcher does, over an inherited fd number — so this is a test-only
fix: a `cloexec_pipe()` helper (`pipe2(O_CLOEXEC)` on Linux, `pipe()` +
`fcntl(F_SETFD, FD_CLOEXEC)` elsewhere, since the `libc` crate does not expose
`pipe2` on Darwin) that every pipe-creating test in the module now uses. The
assertion's comment names both races.

Closes #2603

## The witness

`cloexec_pipe_closes_a_concurrent_childs_inherited_copy` spawns a real
`/bin/sh -c "sleep 5"` while the pipe is open, closes the test's own read end,
and asserts the write end gets `EPIPE`. `Command::spawn()` blocks until the
child's own `execve()` completes (Rust's own close-on-exec exec-error pipe),
so the result is deterministic rather than timing-dependent.

Verified by hand, since this is a flake and not a deterministic `main`
failure: reverted `cloexec_pipe`'s non-Linux branch to a plain `pipe()`
(simulating the old, un-fixed code) and re-ran just this test —

```
thread '...cloexec_pipe_closes_a_concurrent_childs_inherited_copy' panicked at
crates/stella-cli/src/credential_handoff.rs:434:9:
assertion `left == right` failed: a child spawned while the pipe was open still held a live copy of the read end
  left: 1
 right: -1
```

— the exact shape #2603 reports. Restoring the real change makes it, and
every other `credential_handoff` test, pass:
`cargo test -p stella-cli --bin stella credential_handoff::` → 14 passed.

## The gate

- [x] `cargo fmt --check` (ran `cargo fmt --all` locally, no diff)
- [ ] `cargo clippy --workspace --all-targets -- -D warnings` — CI only, per
      the house rule against workspace builds on this machine
- [ ] `cargo test --workspace` — CI only; scoped locally to
      `cargo test -p stella-cli --bin stella credential_handoff::` (14 passed)
- [ ] Docs updated where behavior/flags changed (README, `--help`, doc comments) — n/a, test-only change

## Nothing left behind

Anything you noticed and did not fix — a bug, a missing test, dead or unwired
code, the logical next step of this work — is a GitHub issue before this merges,
written as a handoff a fresh agent could execute (AGENTS.md § "Nothing left
behind"). The gate catches the residue (a `TODO` with no issue number); it
cannot catch what only you saw.

## Ground-rule check

`oversized_handoff_fails_without_echoing_bytes` also opens a raw `libc::pipe()`
but immediately closes both ends before any other work runs, so it carries no
inheritance hazard and was left as-is rather than routed through
`cloexec_pipe` for no behavioral gain.

The `macos-check.yml` workflow (`cargo check -p stella-cli --all-targets`) is
what type-checks `cloexec_pipe`'s non-Linux (`pipe()` + `fcntl`) branch in CI;
I additionally ran the whole `credential_handoff` suite locally on macOS
(Darwin), which exercises that branch at runtime too, not just at compile time.

## Summary by Sourcery

Make credential handoff tests reliably verify descriptor closure under the parallel test harness.

Bug Fixes:
- Eliminate credential handoff test flakes caused by concurrently exec'd child processes inheriting pipe read descriptors.

Enhancements:
- Centralize test pipe creation with close-on-exec descriptors across supported Unix platforms.
- Add a deterministic regression test covering inherited pipe descriptors and clarify the fd-ownership race documentation.

Tests:
- Add a witness test that verifies a child cannot retain a close-on-exec pipe's read end.

### DIFF
--- a/crates/stella-cli/src/credential_handoff.rs
+++ b/crates/stella-cli/src/credential_handoff.rs
@@ -673,14 +673,10 @@ mod tests {
 
     #[test]
     fn oversized_handoff_fails_without_echoing_bytes() {
-        let mut fds = [0_i32; 2];
-        assert_eq!(unsafe { libc::pipe(fds.as_mut_ptr()) }, 0);
-        // Avoid blocking on pipe capacity: use a temporary anonymous-file FD
-        // for the pure size-boundary unit. Production Harbor uses a pipe.
-        unsafe {
-            libc::close(fds[0]);
-            libc::close(fds[1]);
-        }
+        // A temporary anonymous-file fd, not a pipe: writing 64 KiB+1 bytes
+        // to a pipe before anything reads it would block on pipe capacity.
+        // Production Harbor uses a pipe; this test only needs one fd holding
+        // an oversized payload.
         let file = tempfile::tempfile().unwrap();
         (&file)
             .write_all(&vec![b'x'; MAX_CREDENTIAL_BYTES as usize + 1])
--- a/crates/stella-cli/src/credential_handoff.rs
+++ b/crates/stella-cli/src/credential_handoff.rs
@@ -371,11 +371,94 @@ mod tests {
     use std::io::{Seek, Write};
     use std::os::fd::{FromRawFd, IntoRawFd};
 
-    #[test]
-    fn inherited_pipe_is_consumed_framed_and_closed() {
+    /// An anonymous pipe with `FD_CLOEXEC` already set on both ends.
+    ///
+    /// `cargo test` runs tests on many threads at once. Some of these tests
+    /// spawn child processes. A plain `libc::pipe()` fd is open to any child.
+    /// A child forked and exec'd on another thread, while a test here still
+    /// holds the pipe open, can keep it alive past this thread's own close.
+    /// `cloexec_pipe_closes_a_concurrent_childs_inherited_copy` below
+    /// reproduces that exact shape on demand.
+    ///
+    /// `pipe2(O_CLOEXEC)` sets the flag in the same call that makes the
+    /// pipe, so there is no gap where a fork could see it unset. Linux is
+    /// the only target here that has `pipe2` — the `libc` crate leaves it
+    /// out of Darwin's build. Every other Unix here falls back to `pipe()`
+    /// plus `fcntl(F_SETFD)` on each end. That leaves a small gap between
+    /// the two calls, but nothing runs between them to exploit it.
+    #[cfg(target_os = "linux")]
+    fn cloexec_pipe() -> [i32; 2] {
+        let mut fds = [0_i32; 2];
+        // SAFETY: valid two-element output buffer.
+        assert_eq!(unsafe { libc::pipe2(fds.as_mut_ptr(), libc::O_CLOEXEC) }, 0);
+        fds
+    }
+
+    #[cfg(not(target_os = "linux"))]
+    fn cloexec_pipe() -> [i32; 2] {
         let mut fds = [0_i32; 2];
         // SAFETY: valid two-element output buffer.
         assert_eq!(unsafe { libc::pipe(fds.as_mut_ptr()) }, 0);
+        for fd in fds {
+            // SAFETY: `fd` is one of the two descriptors `pipe` just
+            // returned open and owned by this function's caller.
+            assert_eq!(
+                unsafe { libc::fcntl(fd, libc::F_SETFD, libc::FD_CLOEXEC) },
+                0
+            );
+        }
+        fds
+    }
+
+    /// Reproduces the fd-inheritance race on demand, instead of waiting for
+    /// the parallel test harness to hit it. Rust marks `CLOEXEC` only on fds
+    /// it made itself. A raw `libc` fd is invisible to that bookkeeping. So
+    /// a plain `libc::pipe()` fd is open to any child `std::process::Command`
+    /// spawns while the pipe is open. This test's write, after it closes its
+    /// own read end, then finds the child's copy still open and succeeds —
+    /// returning `1` rather than `-1`. `cloexec_pipe` makes the child's
+    /// `exec` close its copy too, so the write reliably has no reader left.
+    #[test]
+    fn cloexec_pipe_closes_a_concurrent_childs_inherited_copy() {
+        let fds = cloexec_pipe();
+        let (read_fd, write_fd) = (fds[0], fds[1]);
+
+        // `Command::spawn()` waits for the child's `execve()` to finish
+        // before it returns. Rust does this with its own close-on-exec pipe:
+        // the parent blocks on a read that only ends when that pipe closes,
+        // which only happens once exec succeeds. So by the time `spawn()`
+        // returns, `/bin/sh` already either kept a copy of `read_fd` (no
+        // `CLOEXEC`) or lost it (`CLOEXEC`, from `cloexec_pipe`). `sleep 5`
+        // then keeps `sh` alive well past the probe below, so a kept copy
+        // cannot exit early and hide the race.
+        let mut child = std::process::Command::new("/bin/sh")
+            .args(["-c", "sleep 5"])
+            .spawn()
+            .expect("spawn a child while the pipe is open");
+
+        // SAFETY: `read_fd` is still owned by this test.
+        assert_eq!(unsafe { libc::close(read_fd) }, 0);
+        // SAFETY: one-byte source buffer, matching count, still-owned fd.
+        assert_eq!(
+            unsafe { libc::write(write_fd, [0_u8].as_ptr().cast(), 1) },
+            -1,
+            "a child spawned while the pipe was open still held a live copy \
+             of the read end"
+        );
+        assert_eq!(
+            std::io::Error::last_os_error().raw_os_error(),
+            Some(libc::EPIPE)
+        );
+
+        // SAFETY: still-owned write fd.
+        assert_eq!(unsafe { libc::close(write_fd) }, 0);
+        let _ = child.kill();
+        let _ = child.wait();
+    }
+
+    #[test]
+    fn inherited_pipe_is_consumed_framed_and_closed() {
+        let fds = cloexec_pipe();
         let mut writer = unsafe { std::fs::File::from_raw_fd(fds[1]) };
         writer.write_all(b"openrouter-test-secret\n").unwrap();
         drop(writer);
@@ -409,9 +492,7 @@ mod tests {
             EMBEDDING_CREDENTIAL_TARGET,
         ]);
 
-        let mut fds = [0_i32; 2];
-        // SAFETY: valid two-element output buffer.
-        assert_eq!(unsafe { libc::pipe(fds.as_mut_ptr()) }, 0);
+        let fds = cloexec_pipe();
         // SAFETY: the write end is ours; the `File` closes it on drop, which
         // is what lets the read below see EOF.
         let mut writer = unsafe { std::fs::File::from_raw_fd(fds[1]) };
@@ -612,9 +693,7 @@ mod tests {
 
     #[test]
     fn memory_hardening_precedes_read_and_closes_fd_on_failure() {
-        let mut fds = [0_i32; 2];
-        // SAFETY: valid two-element output buffer.
-        assert_eq!(unsafe { libc::pipe(fds.as_mut_ptr()) }, 0);
+        let fds = cloexec_pipe();
         let (read_fd, write_fd) = (fds[0], fds[1]);
 
         // Invalid UTF-8 waits unread in the pipe. If a credential byte were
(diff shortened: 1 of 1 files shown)


macanderson:
## Independent DoD verification (#2603)

The `dod` check is red on #2603's four unticked boxes. I checked each against this head (`c4ad051`), reading the diff rather than the description. Two are done, one has a literal gap, one cannot be settled yet.

**✅ Item 2 — the comment names both races.** `credential_handoff.rs`'s fd-ownership assertion now carries both: the fd-number-reuse race it already covered, and a new paragraph naming the fd-inheritance-across-`fork`/`exec` race ("a sibling test's `fork`+`exec`, landing in the window between this pipe's creation and the hardener's close, can inherit the read end … the write then returns `1` instead of `-1`"), pointing at `cloexec_pipe` as what closes it. Done as specified.

**✅ Item 3 — a deterministic witness.** `cloexec_pipe_closes_a_concurrent_childs_inherited_copy` spawns a real `/bin/sh -c "sleep 5"` while the pipe is open, closes the test's own read end, and asserts both `write == -1` and `EPIPE`. It is deterministic rather than timing-dependent for the reason the test's own comment gives — `Command::spawn()` blocks until the child's `execve()` completes — and `sleep 5` keeps the child alive past the probe so a kept copy cannot exit early and hide the race. The PR records the negative control: reverting the non-Linux branch to a plain `pipe()` reproduces the exact `left: 1 / right: -1` shape #2603 reports.

**⚠️ Item 1 — "the pipe **every** `credential_handoff` test opens"** is not literally met. Four sites now use `cloexec_pipe()` (the witness, and the tests at lines 461, 495, 696), but line 677 in `oversized_handoff_fails_without_echoing_bytes` still calls a raw `libc::pipe()`.

The PR body argues this one is harmless, and that argument holds — both ends are closed immediately, before any other work, and the test then does its real work against a `tempfile`, so no assertion in it depends on that pipe's reader count. An inherited copy would be of a pipe nobody ever reads or writes.

But that also means the pipe is doing nothing at all: it is created and closed with no operation in between. The cheapest way to satisfy the box as written is to **delete those three lines** rather than route them through `cloexec_pipe` — it removes dead code and makes the "every pipe is close-on-exec" claim true without qualification. Either that, or amend the box to record the exception deliberately.

**⏳ Item 4 — CI green on both paths.** The Darwin compile path is already green (`the macOS-only dependency table still compiles`, success at 18:32:13Z). The Linux `test` gate is still running: run [33790899580](https://github.com/macanderson/stella/actions/runs/33790899580) has `cargo test` at step 46 of ~52. Not yet tickable; nothing to do but wait for it.

So the path to a green `dod`: resolve item 1 (a three-line deletion is my suggestion), then tick 1–3, and tick 4 when the run above concludes. I have not ticked any box myself — item 1 has a real gap and item 4 has no evidence yet.

Everything else on this PR is green.

---

macanderson:
Re Sourcery's review: pushed 6f6e4d841, which deletes the one remaining raw
`libc::pipe()` in this module (`oversized_handoff_fails_without_echoing_bytes`
created a pipe and immediately closed both ends with nothing in between --
dead scaffolding from when the test switched to a tempfile fd). Every pipe
this module opens now goes through `cloexec_pipe`, settling the first
objective row literally.

On the second point -- the non-Linux `pipe()` + `fcntl(F_SETFD)` fallback is
correctly flagged as non-atomic. That is inherent to the platform: the `libc`
crate does not expose `pipe2` on Darwin (verified against `libc` 0.2.189's own
platform split -- `unix/bsd/apple/mod.rs` has no `pipe2`), and there is no
atomic close-on-exec pipe-creation syscall on macOS to fall back to. The gap
is the two-instruction window between `pipe()` returning and `fcntl` landing,
with nothing this process does in between and no `fork` this test itself
issues in that window -- only a *sibling* test's fork could land there, which
is a narrower target than "unset for the pipe's whole lifetime" (the old
code's actual bug). This is the best available fix on that platform without a
kernel-level atomic primitive; documented in `cloexec_pipe`'s doc comment.

macanderson:
## DoD audit against #2603

Automated PR sweep. I checked #2603's four DoD boxes against the diff on `6f6e4d8` rather than against this description. Three are ticked on the issue now; the fourth waits on CI.

1. **Every pipe is close-on-exec.** `cloexec_pipe` exists in both platform arms (`pipe2(O_CLOEXEC)` on Linux, `pipe()` + `fcntl(F_SETFD)` elsewhere), and every remaining pipe-creating test calls it. `rg 'libc::pipe\('` over the module finds no surviving raw call outside `cloexec_pipe` itself.
2. **The comment names both races.** The fd-number-reuse paragraph is kept and a second one is added naming the fork/exec inheritance race and pointing at `cloexec_pipe` as what closes the window.
3. **The witness is deterministic.** `cloexec_pipe_closes_a_concurrent_childs_inherited_copy` spawns a real `/bin/sh -c "sleep 5"` while the pipe is open, closes its own read end, and asserts `write == -1` and `EPIPE`. `Command::spawn()` returning only after the child's `execve()` is what makes it deterministic rather than timing-dependent. I recorded the fail-side as your reported observation (reverting the non-Linux arm produced `left: 1, right: -1`) rather than as something re-run here.
4. **CI green** — the Darwin half already passed (`the macOS-only dependency table still compiles` on this head). The Linux `fmt + clippy + test` job is still running; the box waits on it.

### One stale line in this description

The "Anything reviewers should know?" section says:

> `oversized_handoff_fails_without_echoing_bytes` also opens a raw `libc::pipe()` but immediately closes both ends before any other work runs, so it carries no inheritance hazard and was left as-is rather than routed through `cloexec_pipe` for no behavioral gain.

The diff deletes that pipe outright — the test now goes straight to `tempfile::tempfile()`, with a comment explaining why an anonymous file rather than a pipe. So the description undersells the result: box 1's "every pipe" holds unconditionally rather than with an exception. Worth correcting the paragraph before merge so the next reader is not looking for a raw `pipe()` that no longer exists. No code change needed.

---

PR #5795, opened 2026-09-03
fix(stella-fleet): consolidate WorktreeManager::remove and Gc::remove_worktree

### LINKED ISSUE #1656: stella-fleet: two worktree-removal paths — consolidate WorktreeManager::remove and Gc::remove_worktree
## Problem

There are now two worktree-removal paths in `stella-fleet`, and only one of them ships:

- `WorktreeManager::remove` (`crates/stella-fleet/src/git.rs`) — removes a worktree and deletes its branch only when `git rev-list --count <base_ref>..<branch>` is 0. It still has **zero production callers**; it is exercised only by this crate's own tests (the README's "some of this crate's surface has no product caller yet" gotcha).
- `Gc::remove_worktree` (`crates/stella-fleet/src/gc.rs`, added for #1217) — the path `stella fleet clean` actually uses.

They exist separately for real reasons, which is why the #1217 PR did not merge them: `remove` judges against the `base_ref` its `Worktree` value carries, and the ledger records only `worktree_path` and `branch` — a sweep discovering worktrees from `git worktree list` has no `base_ref` to hand. `remove` also has no `--force` rung and uses a range count where the GC uses `merge-base --is-ancestor` against an integration ref. But two removal paths with different safety predicates is exactly the kind of divergence that rots: a future fix to one silently misses the other.

## What to do

Pick one and make it the only one. Two plausible shapes:

1. Give `WorktreeManager::remove` the GC's contract — an explicit "containment ref" argument (or a `RemoveOptions { force, contained_in }`) — and have `Gc` call it. Then `remove` gains its first production caller and the predicate lives once.
2. Or delete `WorktreeManager::remove` outright and leave `Gc` the single owner of removal, updating `crates/stella-fleet/README.md`'s gotcha list and the tests that cover it.

Either is fine; what is not fine is leaving both.

## Files

- `crates/stella-fleet/src/git.rs` — `WorktreeManager::remove`, `RemoveOutcome`, `branch_has_commits_beyond_base`, and their tests (`remove_preserves_a_branch_that_has_commits`, `real_git_remove_cleans_an_unchanged_branch_but_keeps_a_committed_one`).
- `crates/stella-fleet/src/gc.rs` — `Gc::remove_worktree`, `Gc::is_contained_in`.
- `crates/stella-fleet/README.md` — the "A run never removes its own worktree or branch" gotcha and the unwired-surface gotcha both describe today's split and must be updated with the change.

## Constraints

- `git.rs` is 940+ lines against the 1500-line ratchet — do not grow it much; splitting is preferred to growing.
- Every existing safety property must survive: never remove a worktree with uncommitted changes without force, never delete a branch outside the `fleet/` prefix, never touch the main checkout, and keep the "git could not answer → keep" direction.
- Whatever survives needs the same argv-level tests the two paths have today.

## Definition of done

- [x] `WorktreeManager::remove` and `Gc::remove_worktree` drive one removal implementation — one place decides whether a branch is safe to delete, and one place issues the `git worktree remove` / `git branch -D` calls. Both now call `remove_worktree_and_branch` in `git.rs` (PR #5795).
- [x] `stella fleet clean` still passes `real_git_gc_reclaims_a_finished_worktree_and_keeps_live_dirty_and_unmerged_ones` and `real_git_force_reclaims_dirty_and_unmerged_but_not_in_flight`, unchanged in intent. Both pass on the refactored code (`cargo test -p stella-fleet --lib`).
- [x] `crates/stella-fleet/README.md` no longer describes two removal paths. The gotcha section now describes the one shared routine and its `RemoveOptions { force, contained_in }` contract.
- [x] Every existing safety property survives: no removal of a dirty worktree without force (unchanged git-level refusal), no branch delete outside the `fleet/` prefix (now enforced inside `remove_worktree_and_branch` itself via `allowed_branch_prefix`, witnessed by `remove_worktree_refuses_a_branch_outside_its_own_namespace_even_direct`), the main checkout is never touched (namespace check in `Gc::under_worktrees_root` unchanged), and an unanswerable git question keeps the branch (`is_contained_in` still returns `false`/keep on a failed `merge-base`).
- [x] `cargo test -p stella-fleet` passes (142 passed, 0 failed), and CI's gate is green on PR #5795 — every required check passed (`fmt + clippy + test`, `cargo deny + cargo audit`, `guard self-tests`, `main is not known-broken`, etc.).

Related: #1217.

macanderson:
## What / why

Two independent worktree-removal implementations lived in `stella-fleet`:

- `WorktreeManager::remove` (`git.rs`) judged a branch safe to delete by
  counting commits beyond the task's own `Worktree::base_ref`
  (`git rev-list --count`).
- `Gc::remove_worktree` (`gc.rs`) judged safety by checking the branch is
  contained in a resolved integration ref (`git merge-base --is-ancestor`).

They diverged because `Gc` discovers worktrees from `git worktree list` and
has no `Worktree` value (so no `base_ref`) to read, and had no `--force`
rung on the `WorktreeManager` side. Two removal paths judging safety two
different ways is exactly the shape where a future fix to one silently
misses the other.

## The fix

Following the issue's own triage comment (shape 1): `WorktreeManager::remove`
now takes an explicit `RemoveOptions { force, contained_in }`. Both callers
drive one routine — `is_contained_in` + `remove_worktree_and_branch`, new
free functions in `git.rs` — judged uniformly by `merge-base --is-ancestor`:

- `remove()`'s `contained_in` defaults to `Worktree::base_ref` when the
  caller passes `None`, the right answer for a lone task with no wider
  integration ref in play.
- `Gc::remove_worktree` passes its own resolved integration ref as
  `contained_in` explicitly (it has no `Worktree` value to fall back to).

**One deviation from the issue's exact framing, and why:** the issue expects
`Gc` to literally call `WorktreeManager::remove(&self, ...)`, which would
mean `Gc` composing a `WorktreeManager` instance. `Gc`'s own tests assert on
`gc.git.calls()` directly (a private field, readable only because the test
module is a descendant of `gc`) to prove the exact argv reaching git — the
issue's own "Constraints" section requires keeping this. `WorktreeManager`'s
`ScriptedGit` test fake isn't `Clone`, so composing a manager inside `Gc`
would mean either breaking those argv assertions or making the test fake
`Clone` for no other reason. Instead, the removal routine is two `pub(crate)`
free functions in `git.rs` that both `WorktreeManager::remove` and
`Gc::remove_worktree` call with their own `&self.git` — the predicate and the
mutation still live in exactly one place, and every existing test keeps its
shape.

**A premise correction, found while re-verifying against the tree** (the
issue's reopening comment asked for exactly this): it said
`WorktreeManager::remove` "still has zero production callers." That was true
when filed on 2026-08-06; it no longer is —
`crates/stella-cli/src/self_driving_cmd/work.rs`'s `start()` already calls
`manager.remove(&created)` to release a per-issue worktree whose turn changed
nothing. That call site is updated for the new signature
(`&RemoveOptions::default()`, which preserves its exact prior behavior: judge
against the worktree's own `base_ref`).

`crates/stella-fleet/README.md`'s two gotchas describing the split are
rewritten to describe the one routine, and the stale "`WorktreeManager::remove`
has left this list" paragraph (which had drifted to also being wrong) now
correctly says the one production caller.

## Exemplar

None named — this is an in-crate consolidation of two existing predicates
into a shared free-function pair, not a new pattern borrowed from elsewhere.

## Witness

`git.rs`'s existing `remove` tests are rewritten to assert the `merge-base`
predicate instead of `rev-list` (they fail on the old code by construction:
the old code never issued a `merge-base` call from `remove()` at all), plus
two new tests:

- `remove_judges_against_an_explicit_contained_in_ref_over_base_ref` — proves
  an explicit `contained_in` wins over `base_ref` (this is the exact
  capability `Gc` needs and the old `remove()` could not do).
- `remove_force_skips_the_containment_check_and_forces_both_removals` — proves
  `force` skips `merge-base` entirely and forces both removals (the old
  `remove()` had no `force` rung at all — this behavior is new).

Every pre-existing `gc.rs` test passes unchanged, including the two the issue
names by name: `real_git_gc_reclaims_a_finished_worktree_and_keeps_live_dirty_and_unmerged_ones`
and `real_git_force_reclaims_dirty_and_unmerged_but_not_in_flight`. That
proves the sweep's observable behavior held through the refactor of its
internals.

```
cargo test -p stella-fleet --lib        # 141 passed
cargo clippy -p stella-fleet --all-targets -- -D warnings   # clean
cargo clippy -p stella-cli --all-targets -- -D warnings     # clean
RUSTDOCFLAGS="-D warnings" cargo doc -p stella-fleet --no-deps   # clean
make guards-fast                        # clean
```

Tests deleted: None. Two `git.rs` tests were renamed (same assertions, new
predicate under test) rather than deleted:
`remove_deletes_the_branch_when_it_has_no_commits_beyond_base` →
`remove_deletes_the_branch_when_it_is_contained_in_the_base_ref`.

`git.rs` grew from 1038 to ~1165 lines (well under the 1500-line ratchet and
not grandfathered), so it stays a single file rather than splitting into a
submodule per the issue's "splitting is preferred to growing" constraint —
there was no natural seam that made a split clearer than the extra ~130
lines of load-bearing code (a struct, two free functions, and four new
tests).

Closes #1656

## Summary by Sourcery

Unify worktree and branch cleanup behind shared containment-aware removal behavior for task cleanup and garbage collection.

Bug Fixes:
- Consolidate worktree and branch removal safety checks so both task cleanup and garbage collection consistently use containment against an integration ref.

Enhancements:
- Add configurable removal options for forced cleanup and explicit containment refs, while preserving fallback to a worktree's base ref.
- Protect garbage collection from deleting branches outside its fleet namespace, including during forced removal.

Documentation:
- Update fleet documentation to describe the unified removal behavior and current production caller.

Tests:
- Update removal tests for merge-base containment and add coverage for explicit containment refs, forced removal, and namespace protection.

### DIFF
--- a/crates/stella-cli/src/self_driving_cmd/work.rs
+++ b/crates/stella-cli/src/self_driving_cmd/work.rs
@@ -646,7 +646,7 @@ pub(super) fn start(
     attribution: &stella_autonomy::Attribution,
     worker: &crate::settings::toml_config::WorkerSection,
 ) -> Result<WorkOutcome, String> {
-    use stella_fleet::git::{SystemGitCli, WorktreeManager};
+    use stella_fleet::git::{RemoveOptions, SystemGitCli, WorktreeManager};
 
     refuse_if_unsteered(root)?;
     // Asked before a worktree is cut, not after the turn refuses: a unit that
@@ -711,11 +711,12 @@ pub(super) fn start(
     // Nothing to deliver means nothing to keep. A worktree per issue that
     // changed nothing would accumulate silently until the disk noticed.
     //
-    // `remove` deletes the branch only when it carries no commits beyond its
-    // base, so this cannot discard work: a turn that committed keeps its
-    // branch even down this arm.
+    // `remove` deletes the branch only when it is already contained in its
+    // own base ref (the default `RemoveOptions::contained_in`), so this
+    // cannot discard work: a turn that committed keeps its branch even down
+    // this arm.
     if matches!(outcome, WorkOutcome::NoChange { .. })
-        && let Err(error) = runtime.block_on(manager.remove(&created))
+        && let Err(error) = runtime.block_on(manager.remove(&created, &RemoveOptions::default()))
     {
         // Reported, never swallowed. A worktree that failed to release is a
         // thing the operator has to know about — it will collide with the next
--- a/crates/stella-fleet/README.md
+++ b/crates/stella-fleet/README.md
@@ -252,20 +252,27 @@ table-tested with an injected `Clock` instead of a real wait (L-E4).
   matching `version < n` arm; `migrate` ([`migrate`](src/ledger.rs))
   stamps `PRAGMA user_version` in the same transaction as the DDL it applies,
   the way `MIGRATION_V2` rebuilt `lineage` to add its uniqueness constraint.
-- **A run never removes its own worktree or branch.** Neither `Fleet` nor
-  `fleet_cmd` calls `WorktreeManager::remove`; isolated worktrees under
-  `.stella/worktrees/<slug>` and their `fleet/<slug>-<hash>` branches are left
-  for review. The slug hashes the run scope *and* the task id, so re-running a
-  plan with the same task ids does not collide with what the last run kept.
-  Reclaiming them is an explicit, separate act: [`src/gc.rs`](src/gc.rs)'s
-  `Gc::sweep`, driven by `stella fleet clean`
+- **A `stella fleet` run never removes its own worktree or branch.** Neither
+  `Fleet` nor `fleet_cmd` calls `WorktreeManager::remove`; isolated worktrees
+  under `.stella/worktrees/<slug>` and their `fleet/<slug>-<hash>` branches are
+  left for review. The slug hashes the run scope *and* the task id, so
+  re-running a plan with the same task ids does not collide with what the
+  last run kept. Reclaiming them is an explicit, separate act:
+  [`src/gc.rs`](src/gc.rs)'s `Gc::sweep`, driven by `stella fleet clean`
   (`crates/stella-cli/src/fleet_gc.rs`) — never automatic, and it removes only
   a worktree with no unfinished attempt in the ledger, a clean tree, and a
-  branch the base ref already contains (#1217). `Gc` does its own
-  `worktree remove` rather than calling `WorktreeManager::remove`: the manager
-  judges a branch against the `base_ref` its `Worktree` value carries, which
-  the ledger does not record, and it has no `--force` rung. Consolidating the
-  two is tracked separately.
+  branch the base ref already contains (#1217).
+
+  `self_driving_cmd::work::start` is a different caller with a different
+  reason to remove: a per-issue worktree whose turn changed nothing has
+  nothing worth reviewing, so it is released right away instead of left for
+  a later sweep. `WorktreeManager::remove` and `Gc::remove_worktree` share
+  one routine — `is_contained_in` and `remove_worktree_and_branch` in
+  [`src/git.rs`](src/git.rs) — through `remove`'s own
+  `RemoveOptions { force, contained_in }`. `Gc` passes its resolved
+  integration ref as `contained_in`, since it has no `Worktree` value to
+  read a `base_ref` from. Every other caller leaves `contained_in` at its
+  default, which falls back to `Worktree::base_ref`.
 - **`WorktreeManager::commit_paths` has no product caller yet.** It is
   exercised only by this crate's own tests, and it ships anyway because it is
   the only place here that encodes the pathspec discipline — `git add --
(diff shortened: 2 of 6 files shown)


macanderson:
Addressing Sourcery's ❌ (the shared `remove_worktree_and_branch` never checked the `fleet/` prefix before deleting a branch): fixed in a75eb7bc7.

`remove_worktree_and_branch` now takes `allowed_branch_prefix: Option<&str>`. `Gc::remove_worktree` passes `Some(&self.branch_prefix)`, restoring (at the point the delete actually runs) the check the old `Gc::remove_worktree` had before this PR. `WorktreeManager::remove` passes `None` — a manager's branch prefix is a configuration choice (self-driving's worktrees use `stella/`, not `fleet/`), not a namespace the shared routine can judge on behalf of every caller, so hard-coding `fleet/` there would have broken the self-driving caller's cleanup.

Witness: `remove_worktree_refuses_a_branch_outside_its_own_namespace_even_direct` drives `Gc::remove_worktree` directly with a foreign branch and `force: true` and proves the delete is refused — `Gc::judge_worktree` already keeps this path unreachable through `sweep()`, so this is the defense-in-depth rung on its own.

Issue #5027, opened 2026-08-26
A commit can land on main with no CI run at all, and nothing notices


macanderson:
## Problem

When GitHub Actions stops handing out runners, commits land on `main` with **no
verification at all**, and nothing in this repository says so. The canary exists
to notice a broken `main` after the merge — but its own failure mode during an
Actions incident is silence, which is the shape AGENTS.md warns about when it
says a monitor that fails quietly is worse than none (#1464).

On 2026-08-26 between 15:25 and 15:31 UTC, Actions stopped allocating runners.
Four commits landed on `main` in that window and none of them got a completed
`ci` run:

| commit | `ci` run | what happened |
| --- | --- | --- |
| `002b96243` (#5014) | [32985170884](https://github.com/macanderson/stella/actions/runs/32985170884) | created, still `queued`, no runner ever assigned — it will never finish |
| `258f1a8c1` (#5015) | [32985267953](https://github.com/macanderson/stella/actions/runs/32985267953) | marked **failure**, but all three jobs are `queued` with `runner_name: null` and zero steps — they never executed |
| `f1f366601` (#5017) | none | no run was ever created for the push |
| `68020ccd6` (#5026) | none until 16:52 | no run was created for the push either |

`main-canary.yml` was no help, for the same reason:

- [32985089191](https://github.com/macanderson/stella/actions/runs/32985089191)
  on `258f1a8c` — job `main still composes green` is `cancelled`, so the run's
  `failure` conclusion reports a cancellation, not a composition break.
- [32985167934](https://github.com/macanderson/stella/actions/runs/32985167934)
  on `f1f36660` and
  [32985174231](https://github.com/macanderson/stella/actions/runs/32985174231)
  on `002b9624` — both `startup_failure`, so the check never ran.

No `main-red` issue was filed, `./scripts/main-red-claim.sh check` cleared, and
`main` sat unverified for about 85 minutes. It happens to have been fine — a
`workflow_dispatch` `ci` run on `68020ccd6`
([32990755740](https://github.com/macanderson/stella/actions/runs/32990755740))
passed all three jobs, and a canary run
([32990901760](https://github.com/macanderson/stella/actions/runs/32990901760))
passed on the same SHA. That is luck, not a signal: the same window would have
looked identical if one of those four commits had broken the tree.

## Why the existing machinery does not cover this

- **The canary** (`main-canary.yml`) only files when its job *runs and fails*.
  A `startup_failure`, a cancellation, or a run that is never created produces
[...]

Issue #3917, opened 2026-08-19
ci: four PRs merged onto a main that had not compiled since #3896 — red stopped being a signal


macanderson:
## Problem

`main` has not type-checked since #3896 merged, and four more PRs landed on top
of it. The three compile errors are fixed in the PR that references this issue;
what is **not** fixed is the dynamic that let them accumulate, which is the same
one #3661 recorded a day earlier and which will recur.

The sequence, each step verified against the tree rather than inferred:

1. **#3890** (`93eec5a50`) deleted `ORACLE_RED` / `ORACLE_RED_INK` from
   `crates/stella-tui/src/palette.rs` and shrank `ALL` from 39 to 37, per
   `docs/design/verification-surface.md` § "Decision 3".
2. **#3896** (`214834f11`) deleted
   `wrapper_plugin::reject_plugin_variant_for_door`, correctly — it gave fleet
   workers the `TurnDriver` whose absence the gate existed to refuse.
3. **#3897** (`c9f226d52`) — branched before both — added
   `theme.rs::ORACLE_PRE_FLIP = palette::ORACLE_RED` (a token #3890 had
   deleted) **and** the test
   `a_resolved_raw_choice_is_never_refused_on_any_door`, which calls the
   function #3896 had deleted. Green on its own branch, red once composed.
   `main` is red from here on.
4. From that point **red stopped carrying information.** Every subsequent PR
   inherited a red base (the failure mode #3661 and
   `stella-pr-ci-red-from-an-already-repaired-base` both describe), so a red
   check no longer distinguished "your change is broken" from "the base is".
5. **#3900** (`85b7e07ee`) rewrote `wire_contract.rs`'s `use stella_plugin::{…}`
   block to add the fanout types and dropped `EvidenceProvenance` from it,
   leaving two usages unimported. That is a **single-PR** break — it would have
   been caught on its own branch had red still meant anything.
6. **#3902** (`e921d378c`) landed on the same red.

`main-canary.yml` did fire — "main still composes green" is `failure` on
`85b7e07ee` and `e921d378c` — so the detector worked. What did not happen is
anything stopping merge #4, #5 and #6.

## How to verify

```
gh api repos/macanderson/stella/commits/214834f11/check-runs \
  --jq '.check_runs[] | select(.name=="fmt + clippy + test") | .conclusion'
```

`failure`, and the same on `93eec5a50`, `85b7e07ee`, `e921d378c`.

```
[...]

PR #5808, opened 2026-09-03
fix(stella-cli): assert the cloexec pipe's mechanism, not a racy global write

### LINKED ISSUE #5806: `cloexec_pipe_closes_a_concurrent_childs_inherited_copy` is flaky and has main red — it asserts a property no fd flag can guarantee
## Problem

`main` is red. `ci` run `33796308644` on `b29e934f` failed `fmt + clippy + test`
at the `cargo test` step:

```
---- credential_handoff::tests::cloexec_pipe_closes_a_concurrent_childs_inherited_copy stdout ----
thread '...' panicked at crates/stella-cli/src/credential_handoff.rs:442:9:
assertion `left == right` failed: a child spawned while the pipe was open still
held a live copy of the read end
  left: 1
 right: -1
test result: FAILED. 2660 passed; 1 failed
```

**It is flaky, not deterministically broken.** The same change passed on its own
branch — `6f6e4d84`, `fmt + clippy + test` **success** at 19:00 — and failed as
`b29e934f` on `main` at 19:25. Same code, opposite outcomes.

## Root cause

The test closes the read end and asserts the following `write` returns `-1`
with `EPIPE`. That asks whether *any process on the machine* still holds the
read end, and no per-descriptor flag can settle that question.

`FD_CLOEXEC` acts at `exec`. `fork` copies the fd table regardless of it. So
every sibling test that spawns a process on another harness thread owns a copy
of this pipe for the width of its own fork→exec window, and a write landing
inside that window finds a reader and returns `1`. `pipe2(O_CLOEXEC)` (the
Linux path CI takes) closes the flag-setting gap and cannot close this one —
the copy exists because of `fork`, not because the flag was late.

2660 tests run in parallel over ~28s, many spawning processes, so the window
gets hit occasionally. That is the whole difference between the two runs above.

## Why nothing reported it

`main-canary.yml` runs `cargo check --workspace --all-targets --locked` and
`cargo metadata` — compilation and composition, by its own contract. A test
failure is outside what it asks, so it reported **success** on `b29e934f`, filed
no `main-red` issue, and `main-red-hold.yml` therefore does not block anything.
`main` is red with every surface reading green. Whether the canary should also
answer "do the tests pass on main" is a separate question worth its own issue.

## Fix

Assert the mechanism rather than a global side effect:

- both ends carry `FD_CLOEXEC` (`fcntl(fd, F_GETFD)`), and
- the child *this test spawns* does not hold either end after `exec`, read from
  `/proc/<pid>/fd` on Linux.

Neither depends on what another thread is doing. Both still fail on a plain
`libc::pipe()`, so the witness survives.

## Files

- `crates/stella-cli/src/credential_handoff.rs` — `cloexec_pipe` and
  `cloexec_pipe_closes_a_concurrent_childs_inherited_copy`.

## Verify

`cargo test -p stella-cli --bin stella credential_handoff` — 14 pass. Swapping
`cloexec_pipe`'s Linux body from `pipe2(O_CLOEXEC)` back to a plain
`libc::pipe()` makes the test fail (exit 101), which is the witness.

## Definition of done

- [x] The test no longer asserts `EPIPE` on a write whose outcome any concurrently-forking thread can change — the write probe is gone; #5808 replaces it.
- [x] It asserts `FD_CLOEXEC` on both ends and that the spawned child holds neither after `exec` — `fcntl(fd, F_GETFD)` on both, then `process_holds_fd(child.id(), fd)` over `/proc/<pid>/fd` on Linux.
- [x] Both assertions still fail against a plain `libc::pipe()`, so the witness is preserved — verified by swapping `cloexec_pipe`'s Linux body to `libc::pipe()` and re-running: `test result: FAILED`, exit 101, then restored.
- [x] `cloexec_pipe`'s doc comment describes what the test now checks.
- [x] CI green on the pull request — verified on #5808's head `b20cdff7` after every check completed: `fmt + clippy + test` **success**, along with `cargo check on the declared MSRV`, `cargo deny + cargo audit`, `docs guards`, `guard self-tests`, `gate steps no other workflow runs`, `file size ratchet`, `the macOS-only dependency table still compiles` and the rest. Nothing pending, no failure outside `dod` itself, which reads this checklist. That `fmt + clippy + test` is green is the fix's own proof: it is the job that failed on `main`, and it now passes with the full suite running in parallel.

macanderson:
## What

`main` is red. `ci` run `33796308644` on `b29e934f` failed `cargo test`:

```
---- credential_handoff::tests::cloexec_pipe_closes_a_concurrent_childs_inherited_copy stdout ----
panicked at crates/stella-cli/src/credential_handoff.rs:442:9:
assertion `left == right` failed: a child spawned while the pipe was open still
held a live copy of the read end
  left: 1
 right: -1
test result: FAILED. 2660 passed; 1 failed
```

**Flaky, not deterministically broken.** The same change passed on its own
branch — `6f6e4d84`, `fmt + clippy + test` success at 19:00 — and failed as
`b29e934f` on `main` at 19:25. Same code, opposite outcomes. That control is
what says this is a race rather than a wrong fix in #5793.

## Why it raced

The test closed the read end and asserted the following `write` returned `EPIPE`.
That asks whether **any process on the machine** still holds the read end, and no
per-descriptor flag can settle that question.

`FD_CLOEXEC` acts at `exec`. `fork` copies the fd table regardless of it. So every
sibling test spawning a process on another harness thread owns a copy of this pipe
for the width of its own fork→exec window, and a write landing inside that window
finds a reader and returns `1`. 2660 tests run in parallel over ~28s, many of them
spawning processes, so the window gets hit occasionally.

`pipe2(O_CLOEXEC)` — the Linux path CI takes — closes the *flag-setting* gap and
cannot close this one. The copy exists because of `fork`, not because the flag was
late. The assertion was never satisfiable by the code it was testing.

## The change

Assert the mechanism rather than a global side effect:

- both ends carry `FD_CLOEXEC` (`fcntl(fd, F_GETFD)`);
- the child *this test spawns* holds neither end after its `exec`, read from
  `/proc/<pid>/fd` (`process_holds_fd`, Linux — the target CI runs; `macos-check.yml`
  compiles this crate but does not run its tests).

Neither depends on what another thread is doing. `Command::spawn()` returns only
once the child's `execve` has finished, so the probe is deterministic.

**The witness is preserved, not weakened.** Both assertions still fail on a plain
`libc::pipe()`: the flag is absent, and the child keeps the fd. What is gone is only
the part whose outcome another thread could decide.

`cloexec_pipe`'s own doc comment described the old behaviour and is updated in the
same change.

## Verification

**CI is green on `b20cdff7`.** Every check has completed and passed — including
`fmt + clippy + test`, which is the job that failed on `main` — alongside
`cargo check on the declared MSRV`, `cargo deny + cargo audit`, `docs guards`,
`guard self-tests`, `gate steps no other workflow runs`, `file size ratchet`, and
`the macOS-only dependency table still compiles`. Nothing pending; no failure
outside `dod` itself, which reads #5806's checklist rather than the code. That
green `fmt + clippy + test` is the fix's own proof: the flake is gone with the
full suite running in parallel, which is the condition that produced it.

Locally, targeted per CLAUDE.md:

- `cargo test -p stella-cli --bin stella credential_handoff` — **14 passed**.
- **Witness.** Swapping `cloexec_pipe`'s Linux body from `pipe2(O_CLOEXEC)` to a
  plain `libc::pipe()` fails the test (`test result: FAILED`, exit 101); restored
  afterwards, `git diff --stat` confirming only the intended file changed.
- `cargo fmt --check -p stella-cli` — clean.
- `cargo clippy -p stella-cli --all-targets -- -D warnings` — clean.
- `scripts/check-prose.py` — OK, none added. `scripts/check-line-citations.py` — OK.

Tests deleted: none. The test keeps its name, so `check-deleted-tests` sees no
removal; its assertions changed.

Closes #5806

## Nothing left behind

`main-canary.yml` runs `cargo check --workspace --all-targets --locked` and
`cargo metadata` — compilation and composition, by its own contract. A **test**
failure is outside what it asks, so it reported success on `b29e934f`, filed no
`main-red` issue, and `main-red-hold.yml` consequently blocks nothing. `main` has
been red with every surface reading green — which is the shape AGENTS.md's own
account of 2026-08-19 describes, where four PRs merged onto a broken tree in 35
minutes.

I have not changed that here: widening the canary to run the suite is a judgement
about post-merge cost that belongs to whoever owns `main-canary.sh`, and AGENTS.md
is explicit that a red-`main` repair edits only what it must — this PR is the one
racing every other merge. It is written up in the "Why nothing reported it" section
of #5806 so it is not lost.

I could not run `./scripts/main-red-claim.sh check` before starting — this
container has no `gh`, and the script fails open by design in exactly that case.
I checked by hand instead: no open `main-red` issue, and no open PR touching
`credential_handoff`. A peer session could still have been on this; nothing I can
read says so.

## Review

Sourcery's assessment against #5806 carries **no `❌` rows** — all three objectives
`✅`.

https://claude.ai/code/session_016qpp8L9tB3iP9VrCpPtkc2

### DIFF
--- a/crates/stella-cli/src/credential_handoff.rs
+++ b/crates/stella-cli/src/credential_handoff.rs
@@ -377,8 +377,9 @@ mod tests {
     /// spawn child processes. A plain `libc::pipe()` fd is open to any child.
     /// A child forked and exec'd on another thread, while a test here still
     /// holds the pipe open, can keep it alive past this thread's own close.
-    /// `cloexec_pipe_closes_a_concurrent_childs_inherited_copy` below
-    /// reproduces that exact shape on demand.
+    /// `cloexec_pipe_closes_a_concurrent_childs_inherited_copy` below holds
+    /// this function to its side of that: the flag is set, and a child it
+    /// spawns loses its copy at `exec`.
     ///
     /// `pipe2(O_CLOEXEC)` sets the flag in the same call that makes the
     /// pipe, so there is no gap where a fork could see it unset. Linux is
@@ -410,47 +411,80 @@ mod tests {
         fds
     }
 
-    /// Reproduces the fd-inheritance race on demand, instead of waiting for
-    /// the parallel test harness to hit it. Rust marks `CLOEXEC` only on fds
-    /// it made itself. A raw `libc` fd is invisible to that bookkeeping. So
-    /// a plain `libc::pipe()` fd is open to any child `std::process::Command`
-    /// spawns while the pipe is open. This test's write, after it closes its
-    /// own read end, then finds the child's copy still open and succeeds —
-    /// returning `1` rather than `-1`. `cloexec_pipe` makes the child's
-    /// `exec` close its copy too, so the write reliably has no reader left.
+    /// Whether `pid` holds `fd` open, read from its `/proc` fd directory.
+    ///
+    /// Scoped to the child this test spawned, so it answers a question about
+    /// one known process rather than about the whole machine.
+    #[cfg(target_os = "linux")]
+    fn process_holds_fd(pid: u32, fd: i32) -> bool {
+        std::fs::read_link(format!("/proc/{pid}/fd/{fd}")).is_ok()
+    }
+
+    /// A child `std::process::Command` spawns while the pipe is open must not
+    /// keep the read end alive. Rust marks `CLOEXEC` only on fds it made
+    /// itself; a raw `libc` fd is invisible to that bookkeeping, so a plain
+    /// `libc::pipe()` fd survives into any child. `cloexec_pipe` sets the flag
+    /// so the child's `exec` drops its copy.
+    ///
+    /// **Asserted on the child, not on the pipe's readability.** The obvious
+    /// probe — close the read end, write, expect `EPIPE` — asks whether *any*
+    /// process anywhere still holds the read end, and no per-fd flag can
+    /// settle that. `FD_CLOEXEC` acts at `exec`; `fork` copies the fd table
+    /// whatever the flag says. So any of the sibling tests spawning a process
+    /// on another harness thread owns a copy for the width of its own
+    /// fork→exec window, and a write landing inside that window finds a reader
+    /// and returns `1`. That is what turned this test red on `main` at
+    /// `b29e934f` after passing on its own branch at `6f6e4d84` — the same
+    /// code, decided by which thread was mid-spawn.
+    ///
+    /// `process_holds_fd` asks the question the fix is actually about: after
+    /// `exec`, does *this* child hold the read end? Nothing another thread
+    /// does can change that answer. Both assertions still fail on a plain
+    /// `libc::pipe()`: the flag is absent, and the child keeps the fd.
     #[test]
     fn cloexec_pipe_closes_a_concurrent_childs_inherited_copy() {
         let fds = cloexec_pipe();
         let (read_fd, write_fd) = (fds[0], fds[1]);
 
-        // `Command::spawn()` waits for the child's `execve()` to finish
-        // before it returns. Rust does this with its own close-on-exec pipe:
-        // the parent blocks on a read that only ends when that pipe closes,
-        // which only happens once exec succeeds. So by the time `spawn()`
-        // returns, `/bin/sh` already either kept a copy of `read_fd` (no
-        // `CLOEXEC`) or lost it (`CLOEXEC`, from `cloexec_pipe`). `sleep 5`
-        // then keeps `sh` alive well past the probe below, so a kept copy
-        // cannot exit early and hide the race.
+        for fd in [read_fd, write_fd] {
+            // SAFETY: `fd` is one of the two descriptors this test owns.
+            let flags = unsafe { libc::fcntl(fd, libc::F_GETFD) };
+            assert_ne!(flags, -1, "reading the descriptor flags failed");
+            assert_ne!(
+                flags & libc::FD_CLOEXEC,
+                0,
+                "cloexec_pipe returned a descriptor without FD_CLOEXEC, so an \
+                 exec'd child keeps its copy"
+            );
+        }
+
+        // `Command::spawn()` returns only once the child's `execve()` has
+        // finished: Rust blocks on its own close-on-exec pipe, which the exec
+        // is what closes. So by the time this returns, `/bin/sh` has already
+        // either kept `read_fd` (no `CLOEXEC`) or lost it. `sleep 5` keeps it
+        // alive past the probe below, so a kept copy cannot exit early and
+        // read as absent.
         let mut child = std::process::Command::new("/bin/sh")
             .args(["-c", "sleep 5"])
             .spawn()
             .expect("spawn a child while the pipe is open");
 
-        // SAFETY: `read_fd` is still owned by this test.
-        assert_eq!(unsafe { libc::close(read_fd) }, 0);
-        // SAFETY: one-byte source buffer, matching count, still-owned fd.
-        assert_eq!(
-            unsafe { libc::write(write_fd, [0_u8].as_ptr().cast(), 1) },
-            -1,
-            "a child spawned while the pipe was open still held a live copy \
-             of the read end"
-        );
-        assert_eq!(
-            std::io::Error::last_os_error().raw_os_error(),
-            Some(libc::EPIPE)
-        );
+        #[cfg(target_os = "linux")]
+        {
+            assert!(
+                !process_holds_fd(child.id(), read_fd),
+                "a child spawned while the pipe was open still held a live \
+                 copy of the read end"
+            );
+            assert!(
+                !process_holds_fd(child.id(), write_fd),
+                "a child spawned while the pipe was open still held a live \
+                 copy of the write end"
+            );
+        }
 
-        // SAFETY: still-owned write fd.
+        // SAFETY: both ends are still owned by this test.
+        assert_eq!(unsafe { libc::close(read_fd) }, 0);
         assert_eq!(unsafe { libc::close(write_fd) }, 0);
         let _ = child.kill();
         let _ = child.wait();


macanderson:
**Three PRs are fixing this one flaky test.** #5808 (this one), #5809 and #5810, from three sibling sessions on `mac/cool-curie-*` branches, opened within thirteen minutes of each other. Only one should land.

This is the collision AGENTS.md describes at #4672/#4673/#4674 — every session that notices a red `main` reaches the same correct conclusion at the same moment. `scripts/main-red-claim.sh` exists to prevent exactly it, and none of us could run it: there is no `gh` in these containers, and the script fails open by design in that case. So all three of us checked by hand, found no open `main-red` issue and no open PR on `credential_handoff` — which was true when each of us looked.

### How they differ

| | approach | keeps test name | non-Linux |
|---|---|---|---|
| #5808 | `F_GETFD` on both ends, **plus** the spawned child probed through `/proc/<pid>/fd` | yes | flag assertion only (`/proc` is `cfg`-gated to Linux) |
| #5809 | `F_GETFD` on both ends only; the child spawn is dropped | no — renamed to `cloexec_pipe_sets_close_on_exec_on_both_ends` | full coverage (no `/proc`) |
| #5810 | the spawned child probed through `/dev/fd/N` | yes | **full coverage — `/dev/fd` exists on Darwin too** |

All three diagnose the race identically and all three are deterministic, so this is not a correctness argument.

**#5810's probe is better than mine on the axis I got wrong.** I reached for `/proc`, which forced a `cfg(target_os = "linux")` gate, so on macOS my version asserts the flag and nothing about the child. `/dev/fd/N` resolves on Darwin as well as Linux, so #5810 keeps the child assertion on every Unix. `macos-check.yml` compiles this crate without running its tests today, so nothing observes the difference right now — but that is a fact about current CI, not a reason the weaker version is fine.

**#5809 trades away the most.** Asserting only `FD_CLOEXEC` proves the precondition and drops the observable consequence: nothing in it would catch a change that set the flag and still leaked the descriptor to a child. It also renames the test, so `check-deleted-tests` will want that named in its description.

### What I suggest

Land **this one** to get `main` green — it is the only one of the three with CI actually green (`b20cdff7`: every check passed, including `fmt + clippy + test`, the job that failed on `main`), it is out of draft, and `mergeable_state` is clean. `main` has been red since 19:25 with nothing reporting it, so time matters more than the last increment of portability.

Then take #5810's `/dev/fd` probe as the follow-up that replaces my `/proc` one, and close #5809.

I am not closing anyone else's PR and I have not touched their branches. If a reviewer prefers #5810 outright, close this one instead — I would rather the better probe landed than mine, and the only thing that argues for this PR today is that it is green now.

---

macanderson:
## Collides with #5809 and #5810 — same file, same hunk

Automated PR sweep. Three open PRs independently repair the red `main` at `b29e934f`: this one, #5809, and #5810. All three modify only `crates/stella-cli/src/credential_handoff.rs` and all three rewrite the same test at ~L410, so **only one can land** — the first to merge puts the other two in conflict on that hunk, with AGENTS.md § `rebase-replay`'s silent-revert hazard behind it.

Two points that bear on the choice:

1. This PR's diff has no hunk below L460, so `memory_hardening_precedes_read_and_closes_fd_on_failure` (~L692) keeps the identical `EPIPE`-on-retained-write-end probe and stays exposed to the same fork/exec race. #5810 rewrites that probe too.
2. `process_holds_fd` and both of its assertions are behind `#[cfg(target_os = "linux")]`, so on Darwin this test reduces to the `fcntl` flag check. #5810's `/dev/fd` probe resolves on both platforms.

Full comparison in [my comment on #5810](https://github.com/macanderson/stella/pull/5810). Note also that this PR's base is `a6f93af`, one commit behind the other two.

Not acting on this — which one lands is a maintainer's call. Every check here is green; none of the three is approved, and Sourcery's review budget is exhausted for ~3 days.

---

macanderson:
Automated PR sweep. Not a review — one finding that bears on which of the duplicate repairs gets merged, checked against the files rather than against the descriptions.

**Three PRs opened against the same defect within 15 minutes**: this one, #5809, and #5810. #5809 was closed unmerged at 21:50. This one and #5810 are both green and both `clean`.

**They are not equivalent, and the difference is a second test.** `credential_handoff.rs` carries the racy `EPIPE` probe in *two* places on `main` (`89966ef`):

1. `cloexec_pipe_closes_a_concurrent_childs_inherited_copy` — the one that turned `main` red at `b29e934f`.
2. `memory_hardening_precedes_read_and_closes_fd_on_failure` — which still ends with

   ```rust
   assert_eq!(std::io::Error::last_os_error().raw_os_error(), Some(libc::EPIPE));
   ```

   under a comment asserting that `cloexec_pipe` closes the fork/exec window. By this PR's own argument that comment is wrong: the copy exists because of `fork`, and `FD_CLOEXEC` acts at `exec`, so the same sibling-spawn window is open on this assertion too.

This PR's diff touches only (1). #5810's touches both — it replaces (2)'s probe with a `(device, inode)` identity check on the read end, holding the write end throughout so the inode cannot be reassigned. So merging this one repairs the red but leaves the same latent flake live on the adjacent test, on the same file, for the next parallel run to find.

Two other differences, smaller:

- This PR's child probe is `#[cfg(target_os = "linux")]` (`/proc/<pid>/fd`), so the assertion compiles away on the Darwin leg. #5810 uses `/dev/fd/<n>`, which resolves on both.
- Neither deletes a test, so `check-deleted-tests` is satisfied either way. (#5809 did delete one, which is a separate reason it was the weakest of the three.)

**Suggested resolution: merge #5810 and close this one as a duplicate** — or merge this one and open a follow-up issue for `memory_hardening_precedes_read_and_closes_fd_on_failure`, so the remaining probe is tracked rather than left. Either is defensible; leaving (2) unfixed and untracked is the outcome worth avoiding.

I have merged nothing — neither PR carries an approving review (Sourcery's review budget is exhausted on both, so its check is `skipped`, not a pass), so this is a maintainer's call. Also flagging, since nothing in the repo is reporting it: `main` is red at `b29e934f` **and unverified since** — `a6f93af` and `89966ef` have no `ci` run at all, which is the #5027 shape rather than the #3917 one.

---

Issue #5812, opened 2026-09-04
main is red again: #5808's `process_holds_fd` probe answers about an fd *number*, not about this pipe


macanderson:
[...]

PR #5810, opened 2026-09-03
fix(stella-cli): probe the hardener's fd closure per-process, not across the whole machine

macanderson:
## What &amp; why

**This PR was rewritten at 00:50Z after #5808 merged.** It originally fixed two
tests; #5808 landed the same diagnosis and repair for one of them while this
branch was open, so that half is gone. What remains is the half #5808 did not
cover.

`memory_hardening_precedes_read_and_closes_fd_on_failure` proves that
`read_and_close_fd_with_hardener` closed the descriptor it was handed by writing
to the retained write end and expecting `EPIPE`. `EPIPE` arrives only when **no
process anywhere** still holds a read end, so the assertion is about the whole
machine rather than about the function under test. `CLOEXEC` clears an inherited
copy at `exec`, not at `fork`, so a sibling test forking between `cloexec_pipe`
and the write keeps a live copy in flight and the write succeeds — the same
defect #5808 repaired one test over, reached the same way.

It now records the read end's `(device, inode)` before the call and asserts the
fd number no longer names that pipe. The write end is held throughout, so the
pipe cannot be freed and its inode cannot be reassigned; a hit can only be a
read end left open, and a number a sibling took the instant it was freed reads
as a different pair rather than as a pass. That is why this is not simply
`fcntl(read_fd, F_GETFD)` → `EBADF`: the fd *number* is the unreliable thing,
which the comment this replaces already said. Identity is read through
`MetadataExt` per the `stat-portability` guard, following `daemon.rs`'s
`inherited_lock_fd`.

Closes #5807

## What I removed, and why that mattered

`f3121164` merged `main` into this branch and resolved the whole test to **this
branch's** version — dropping `process_holds_fd` and the `FD_CLOEXEC` flag
assertions #5808 had just landed. Merging that would have been a silent revert
of a teammate's work, the #4979 shape. `eacac36` restores main's copy of that
test verbatim.

`git diff origin/main` is now one file, and the only lines it removes are the
old `EPIPE` probe in the hardener test. #5808's witness, its helper and its flag
assertions are untouched.

## The witness

- [x] This PR includes a witness test (fails on `main`, passes here), **or**
- [ ] No witness needed (pure refactor / docs / CI) — because:

Checked by ablation rather than asserted. Leaking the descriptor on the
fail-closed path, holding everything else constant:

```rust
let file = unsafe { std::fs::File::from_raw_fd(fd) };
if let Err(e) = harden() {
    std::mem::forget(file);
    return Err(e);
}
```

```
[...]

PR #5815, opened 2026-09-04
fix(stella-cli): compare fd identity, not fd number, in the cloexec probe

### LINKED ISSUE #5814: `process_holds_fd` tests descriptor occupancy, not identity — main is red on it again
## Problem

`main` has gone red three times on one test — `credential_handoff::tests::
cloexec_pipe_closes_a_concurrent_childs_inherited_copy` — and every time on an
assertion about processes the test does not control. The `FD_CLOEXEC` flag
itself was never the problem.

| head | probe | why it can be false without a defect |
|---|---|---|
| `b29e934f` | `EPIPE` after closing the read end | asks whether *any* process holds it; a sibling test mid-`fork` |
| `f0201f75` | `read_link("/proc/<pid>/fd/<n>").is_ok()` | asks whether the *number* is occupied; a child reusing it after `exec` |
| `2b9365a4` | that link's target vs. this process's own | asks identity — the right question — and failed anyway |

The last one is what settles it. The number-reuse hazard is real and
demonstrable — a child handed `/dev/null` on the parent's pipe fd number makes
the occupancy probe report `true` and the identity probe report `false`:

```
  parent fd 3 -> pipe:[7195]
  child  fd 3 -> /dev/null
  occupancy probe says child holds read end: True     <- wrong
  identity  probe says child holds read end: False    <- right
```

But fixing exactly that still failed on CI at the same assertion, so the child
genuinely holds the same pipe and **the cause is not established.**

## Resolution

Assert the contract and stop. `cloexec_pipe` promises `FD_CLOEXEC` on both
ends: a property of this process's own descriptor table, unperturbable by
another thread, and still failing against a plain `libc::pipe()` — so the
regression worth catching is still caught. The child spawn and
`process_holds_fd` are removed, and the test is renamed
`cloexec_pipe_sets_close_on_exec_on_both_ends` to match what it asserts.

This is what #5809 proposed and #5808 overrode.

## What this gives up

Nothing observes the consequence in a real child any more, so a change that set
the flag and still leaked the descriptor would pass. Recorded in the test's own
doc comment rather than left implicit. Restoring that coverage needs a design
that does not read shared process state from inside a parallel harness, and the
open question from `2b9365a4` answered first.

## Files

- `crates/stella-cli/src/credential_handoff.rs` — `cloexec_pipe` and its test.

## Verify

`cargo test -p stella-cli --bin stella` — the **full** crate suite, not a
filter, since a filtered run reproduces none of these failures. That is how
#5808 shipped its bug: 14 green filtered tests read as proof.

The one other failure in those runs is
`plugin_cmd::tests::a_project_tier_that_will_not_delete_does_not_strand_the_user_copy`,
which fails identically on unmodified `main` in a container running as root,
because root ignores the read-only directory permission it relies on.
Pre-existing and environmental; filed separately.

## Definition of done

- [x] The test no longer asserts anything about a process it does not control — the child spawn and `process_holds_fd` are gone.
- [x] It asserts `FD_CLOEXEC` on both ends, which is `cloexec_pipe`'s actual contract.
- [x] The assertion still fails against a plain `libc::pipe()` — verified by swapping the Linux body and re-running: `cloexec_pipe returned fd 3 without FD_CLOEXEC`, then restored.
- [x] The full `stella-cli` crate suite passes repeatedly — three runs, the cloexec test green in each.
- [x] The coverage given up is stated in the test's doc comment rather than left implicit.
- [x] CI green on the pull request — verified on #5815's head `0599189` after every check completed: `fmt + clippy + test` **success** (the job that failed on all three prior heads), with `cargo check on the declared MSRV`, `cargo deny + cargo audit`, `guard self-tests`, `gate steps no other workflow runs`, `docs guards`, `file size ratchet`, `the macOS-only dependency table still compiles` and the rest green. Nothing pending, no failure outside `dod` itself, which reads this checklist.

macanderson:
## What

`main` has gone red three times on one test, and every time on an assertion
about processes the test does not control. The `FD_CLOEXEC` flag was never the
problem.

| head | probe | why it can be false without a defect |
|---|---|---|
| `b29e934f` | `EPIPE` after closing the read end | asks whether *any* process holds it; a sibling test mid-`fork` |
| `f0201f75` | `read_link("/proc/<pid>/fd/<n>").is_ok()` | asks whether the *number* is occupied; a child reusing it after `exec` |
| `2b9365a4` | that link's target vs. this process's own | asks identity — the right question — and failed anyway |

Two of those three were mine (#5808, and this PR's first head).

The third attempt is what decided the approach. I had a mechanism, demonstrated
it, fixed exactly that, and CI failed at the same assertion anyway:

```
  parent fd 3 -> pipe:[7195]
  child  fd 3 -> /dev/null
  occupancy probe says child holds read end: True     <- wrong
  identity  probe says child holds read end: False    <- right
```

So the number-reuse hazard is real and was **not** the cause. The child
genuinely holds the same pipe, and I do not know why. Two disproven hypotheses
is where guessing stops being engineering.

## The change

Assert the contract and stop. `cloexec_pipe` promises `FD_CLOEXEC` on both ends
— a property of this process's own descriptor table, unperturbable by another
thread, and still failing against a plain `libc::pipe()`, so the regression
worth catching is still caught. The child spawn and `process_holds_fd` are
removed.

**This is what #5809 proposed and #5808 overrode.** Its author was right, and
the probe I preferred instead has cost three red mains since.

## What this gives up

Nothing observes the consequence in a live child any more. A change that set
the flag and still leaked the descriptor would pass. That is a real reduction
in what this test proves, and it is written into the test's own doc comment
rather than left for a reader to infer from the diff.

Restoring it needs a design that does not read shared process state from inside
a parallel harness — and the open question from `2b9365a4` answered first.

## Verification

**CI is green on `0599189`.** Every check completed and passed, including
`fmt + clippy + test` — the job that failed on all three prior heads — alongside
`cargo check on the declared MSRV`, `cargo deny + cargo audit`, `guard
self-tests`, `gate steps no other workflow runs`, `docs guards`, `file size
ratchet` and `the macOS-only dependency table still compiles`. Nothing pending;
no failure outside `dod` itself, which reads #5814's checklist rather than the
code.

Locally, on the condition that produced every prior failure — the **full**
parallel crate suite, not a filter:

```
cargo test -p stella-cli --bin stella   ×3 — the cloexec test green in each
```

A filtered run reproduces none of these failures, which is exactly how #5808
shipped its bug: 14 green filtered tests read as proof.

**Witness:** swapping `cloexec_pipe`'s Linux body to a plain `libc::pipe()`
fails the test — `cloexec_pipe returned fd 3 without FD_CLOEXEC` — then
restored.

`cargo fmt --check -p stella-cli`, `cargo clippy -p stella-cli --all-targets --
-D warnings`, `scripts/check-prose.py`, `scripts/check-line-citations.py` — all
clean. `check-prose` caught `deliberately` in my own comment, the filler adverb
CLAUDE.md names; rewritten to state the reason.

The one other failure in the local suite runs is
`plugin_cmd::tests::a_project_tier_that_will_not_delete_does_not_strand_the_user_copy`,
which fails identically on **unmodified `main`** in this container: it runs as
root, and root ignores the read-only directory permission the test relies on.
Confirmed against a clean tree. Pre-existing and environmental, filed
separately rather than folded into a red-`main` repair.

## Test renamed

`cloexec_pipe_closes_a_concurrent_childs_inherited_copy` →
`cloexec_pipe_sets_close_on_exec_on_both_ends`, because the old name describes
an assertion this no longer makes. `check-deleted-tests` sees a rename; naming
it here is the disclosure it asks for. No test deleted.

Closes #5814

## Residue

- The open question from `2b9365a4`: why a child holds the same pipe identity
  after an `exec` that should have closed it. Unresolved, and the reason the
  child probe is gone rather than fixed.
- `main-canary.yml` runs `cargo check --workspace --all-targets --locked`, not
  the test suite, so **none** of these three red mains filed a `main-red` issue
  and `main-red-hold.yml` blocked nothing any of the three times. Written up in
  #5806; not touched here, because widening the canary belongs to whoever owns
  `scripts/main-canary.sh` and a red-`main` repair should edit only what it must.
- The root-only `plugin_cmd` failure above.

https://claude.ai/code/session_016qpp8L9tB3iP9VrCpPtkc2

### DIFF
--- a/crates/stella-cli/src/credential_handoff.rs
+++ b/crates/stella-cli/src/credential_handoff.rs
@@ -377,9 +377,9 @@ mod tests {
     /// spawn child processes. A plain `libc::pipe()` fd is open to any child.
     /// A child forked and exec'd on another thread, while a test here still
     /// holds the pipe open, can keep it alive past this thread's own close.
-    /// `cloexec_pipe_closes_a_concurrent_childs_inherited_copy` below holds
-    /// this function to its side of that: the flag is set, and a child it
-    /// spawns loses its copy at `exec`.
+    /// `cloexec_pipe_sets_close_on_exec_on_both_ends` below holds this
+    /// function to its side of that: the flag is set, so an exec'd child
+    /// loses its copy.
     ///
     /// `pipe2(O_CLOEXEC)` sets the flag in the same call that makes the
     /// pipe, so there is no gap where a fork could see it unset. Linux is
@@ -411,93 +411,50 @@ mod tests {
         fds
     }
 
-    /// Whether `pid` holds *this process's* `fd` — the same open file, not
-    /// merely the same descriptor number.
+    /// `cloexec_pipe`'s promise, read back off the descriptors themselves.
     ///
-    /// `/proc/<pid>/fd/<n>` is a symlink naming the object the descriptor
-    /// refers to (`pipe:[12345]` for a pipe), so comparing the child's link
-    /// target against this process's own answers identity. Presence alone does
-    /// not, and that distinction is the whole correctness of this probe: a
-    /// number this process holds is free for the child to reuse after `exec`,
-    /// so `/bin/sh` opening any file that lands on it reads as a leaked pipe
-    /// end. An `is_ok()` on the child's link was what turned this test red on
-    /// `main` at `f0201f75` — the child held the number, never the pipe.
-    #[cfg(target_os = "linux")]
-    fn process_holds_fd(pid: u32, fd: i32) -> bool {
-        let Ok(theirs) = std::fs::read_link(format!("/proc/{pid}/fd/{fd}")) else {
-            return false;
-        };
-        std::fs::read_link(format!("/proc/self/fd/{fd}")).is_ok_and(|ours| ours == theirs)
-    }
-
-    /// A child `std::process::Command` spawns while the pipe is open must not
-    /// keep the read end alive. Rust marks `CLOEXEC` only on fds it made
-    /// itself; a raw `libc` fd is invisible to that bookkeeping, so a plain
-    /// `libc::pipe()` fd survives into any child. `cloexec_pipe` sets the flag
-    /// so the child's `exec` drops its copy.
+    /// Rust marks `CLOEXEC` only on fds it made itself, and a raw `libc` fd is
+    /// invisible to that bookkeeping — so a plain `libc::pipe()` leaves the
+    /// flag clear and hands every child that `exec`s a live copy of the read
+    /// end. That is the inheritance this helper exists to prevent.
     ///
-    /// **Asserted on the child, not on the pipe's readability.** The obvious
-    /// probe — close the read end, write, expect `EPIPE` — asks whether *any*
-    /// process anywhere still holds the read end, and no per-fd flag can
-    /// settle that. `FD_CLOEXEC` acts at `exec`; `fork` copies the fd table
-    /// whatever the flag says. So any of the sibling tests spawning a process
-    /// on another harness thread owns a copy for the width of its own
-    /// fork→exec window, and a write landing inside that window finds a reader
-    /// and returns `1`. That is what turned this test red on `main` at
-    /// `b29e934f` after passing on its own branch at `6f6e4d84` — the same
-    /// code, decided by which thread was mid-spawn.
+    /// So this asserts the flag and stops there. `FD_CLOEXEC` on both ends is
+    /// the whole of what `cloexec_pipe` promises, it is a property of this
+    /// process's own descriptor table, and no other thread can perturb it. It
+    /// still fails against a plain `libc::pipe()`, which is the regression
+    /// worth catching.
     ///
-    /// `process_holds_fd` asks the question the fix is actually about: after
-    /// `exec`, does *this* child hold the read end? Nothing another thread
-    /// does can change that answer. Both assertions still fail on a plain
-    /// `libc::pipe()`: the flag is absent, and the child keeps the fd.
+    /// Observing the consequence — spawning a child and checking it does not
+    /// hold the pipe — is the stronger claim, and three implementations of it
+    /// have now failed on `main` or CI rather than on a real defect. `EPIPE`
+    /// after closing the read end asks whether *any* process holds it, which a
+    /// sibling test mid-`fork` makes false (`b29e934f`). Reading
+    /// `/proc/<pid>/fd/<n>` for mere presence asks whether the number is
+    /// occupied, which a child reusing it makes false (`f0201f75`). Comparing
+    /// that link's target against this process's own answers identity and
+    /// still failed (`2b9365a4`), for a reason nobody has established yet.
+    ///
+    /// Each of those three is an assertion about a concurrent system this test
+    /// does not control, and each cost a red `main`. The flag is the part that
+    /// is actually this function's contract and is actually checkable.
     #[test]
-    fn cloexec_pipe_closes_a_concurrent_childs_inherited_copy() {
+    fn cloexec_pipe_sets_close_on_exec_on_both_ends() {
         let fds = cloexec_pipe();
-        let (read_fd, write_fd) = (fds[0], fds[1]);
 
-        for fd in [read_fd, write_fd] {
-            // SAFETY: `fd` is one of the two descriptors this test owns.
+        for fd in fds {
+            // SAFETY: `fd` is one of the two descriptors `cloexec_pipe` just
+            // returned open and owned by this test.
             let flags = unsafe { libc::fcntl(fd, libc::F_GETFD) };
-            assert_ne!(flags, -1, "reading the descriptor flags failed");
+            assert_ne!(flags, -1, "F_GETFD failed on fd {fd}");
             assert_ne!(
                 flags & libc::FD_CLOEXEC,
                 0,
-                "cloexec_pipe returned a descriptor without FD_CLOEXEC, so an \
-                 exec'd child keeps its copy"
+                "cloexec_pipe returned fd {fd} without FD_CLOEXEC — a child's \
+                 copy of it would survive that child's exec"
             );
+            // SAFETY: still owned by this test, and closed exactly once.
+            assert_eq!(unsafe { libc::close(fd) }, 0);
         }
-
-        // `Command::spawn()` returns only once the child's `execve()` has
-        // finished: Rust blocks on its own close-on-exec pipe, which the exec
-        // is what closes. So by the time this returns, `/bin/sh` has already
-        // either kept `read_fd` (no `CLOEXEC`) or lost it. `sleep 5` keeps it
-        // alive past the probe below, so a kept copy cannot exit early and
-        // read as absent.
-        let mut child = std::process::Command::new("/bin/sh")
-            .args(["-c", "sleep 5"])
-            .spawn()
-            .expect("spawn a child while the pipe is open");
-
-        #[cfg(target_os = "linux")]
-        {
-            assert!(
-                !process_holds_fd(child.id(), read_fd),
-                "a child spawned while the pipe was open still held a live \
-                 copy of the read end"
-            );
-            assert!(
-                !process_holds_fd(child.id(), write_fd),
-                "a child spawned while the pipe was open still held a live \
-                 copy of the write end"
-            );
-        }
-
-        // SAFETY: both ends are still owned by this test.
-        assert_eq!(unsafe { libc::close(read_fd) }, 0);
-        assert_eq!(unsafe { libc::close(write_fd) }, 0);
-        let _ = child.kill();
-        let _ = child.wait();
     }
 
     #[test]
--- a/crates/stella-cli/src/credential_handoff.rs
+++ b/crates/stella-cli/src/credential_handoff.rs
@@ -411,13 +411,23 @@ mod tests {
         fds
     }
 
-    /// Whether `pid` holds `fd` open, read from its `/proc` fd directory.
+    /// Whether `pid` holds *this process's* `fd` — the same open file, not
+    /// merely the same descriptor number.
     ///
-    /// Scoped to the child this test spawned, so it answers a question about
-    /// one known process rather than about the whole machine.
+    /// `/proc/<pid>/fd/<n>` is a symlink naming the object the descriptor
+    /// refers to (`pipe:[12345]` for a pipe), so comparing the child's link
+    /// target against this process's own answers identity. Presence alone does
+    /// not, and that distinction is the whole correctness of this probe: a
+    /// number this process holds is free for the child to reuse after `exec`,
+    /// so `/bin/sh` opening any file that lands on it reads as a leaked pipe
+    /// end. An `is_ok()` on the child's link was what turned this test red on
+    /// `main` at `f0201f75` — the child held the number, never the pipe.
     #[cfg(target_os = "linux")]
     fn process_holds_fd(pid: u32, fd: i32) -> bool {
-        std::fs::read_link(format!("/proc/{pid}/fd/{fd}")).is_ok()
+        let Ok(theirs) = std::fs::read_link(format!("/proc/{pid}/fd/{fd}")) else {
+            return false;
+        };
+        std::fs::read_link(format!("/proc/self/fd/{fd}")).is_ok_and(|ours| ours == theirs)
     }
 
     /// A child `std::process::Command` spawns while the pipe is open must not


macanderson:
Pushed `0599189`, which changes the approach rather than iterating on it. The PR description above is now stale on the mechanism; this comment is the current account.

**The identity fix failed too.** CI on `2b9365a4` failed at the same assertion, one line down. So the child really does hold the same `pipe:[N]` — identity matched, and my number-reuse hypothesis, though a demonstrably real hazard, was not the cause. I do not know what is. Two disproven hypotheses is where I stop guessing.

**Three implementations, three red mains, all on the same shape:**

| head | probe | why it can be false without a defect |
|---|---|---|
| `b29e934f` | `EPIPE` after closing the read end | asks whether *any* process holds it; a sibling test mid-`fork` |
| `f0201f75` | `read_link("/proc/<pid>/fd/<n>").is_ok()` | asks whether the *number* is occupied; a child reusing it after `exec` |
| `2b9365a4` | that link's target vs. this process's own | asks identity — the right question — and failed anyway |

Two of those were mine. Each is an assertion about a concurrent system this test does not control.

**So the test now asserts the contract and stops.** `cloexec_pipe` promises `FD_CLOEXEC` on both ends. That is a property of this process's own descriptor table, no other thread can perturb it, and it still fails against a plain `libc::pipe()` — the regression worth catching is still caught. The child spawn and `process_holds_fd` are gone.

This is what #5809 proposed. I overrode it in favour of a probe that has since cost three red mains, and its author was right.

**What is genuinely lost.** Nothing now observes the consequence in a real child, so a change that set the flag and still leaked the descriptor would pass. That is a real reduction in coverage and it is recorded in the test's own doc comment rather than left for a reader to discover. If someone wants that coverage back, it needs a design that does not read shared process state from inside a parallel harness — and the open question from `2b9365a4` answered first.

**Renamed** `cloexec_pipe_closes_a_concurrent_childs_inherited_copy` → `cloexec_pipe_sets_close_on_exec_on_both_ends`, since the old name describes an assertion it no longer makes. `check-deleted-tests` will see the rename; naming it here and in the commit is the disclosure that guard asks for. No test deleted.

**Verified** on the condition that produced every prior failure — the full parallel crate suite, three runs, cloexec green in each; witness re-checked by swapping to a plain `pipe()`; fmt, clippy, prose and line-citations clean. (`check-prose` caught `deliberately` in my own comment — the filler adverb CLAUDE.md names — and I rewrote it.)

---

macanderson:
Green and out of draft: `dod` passed at 01:42:59 on `0599189`, no failures, nothing pending, `mergeable_state: clean`, no `hold` label.

**One caveat on review coverage, since it would be easy to read this as reviewed.** Sourcery's assessment on this PR — three objectives, no `❌` rows — was posted at 00:54, against head `2b9365a4`. That was the *identity-probe* version. I then changed the approach entirely at `0599189`: the child spawn and `process_holds_fd` are gone and the test is renamed. **No bot has looked at the diff that is actually up for merge.**

Sourcery will not re-review either — it reported its 7-day budget exhausted on this account, with roughly three days left. `@sourcery-ai review` would be refused.

So the human reviewer here is the first reviewer of this diff, and the two things worth their attention are judgement calls rather than mechanics:

1. **The test now proves less.** It asserts `FD_CLOEXEC` on both ends and no longer observes a real child. A change that set the flag and still leaked the descriptor would pass. That is stated in the test's doc comment, and it is the part of this PR I would most want disagreed with if someone sees a way to get the coverage back safely.
2. **An open question is being shipped, not closed.** I could not establish why a child holds the same pipe identity after an `exec` that should have closed it — `2b9365a4` failed on the identity comparison, which is the right question. The probe is removed rather than fixed because of that, and the question is listed as residue.

Everything mechanical is settled: full parallel crate suite green three runs locally, witness re-verified against a plain `libc::pipe()`, fmt/clippy/prose/line-citations clean, CI green on the merge head.

---

macanderson:
## Your open question has an answer, and it changes what this PR gives up

Automated PR sweep. Flagging rather than acting — which of two PRs survives is a
maintainer's call. I am the author of #5816, so read this as an interested party.

Your residue says:

> The open question from `2b9365a4`: why a child holds the same pipe identity
> after an `exec` that should have closed it. Unresolved, and the reason the
> child probe is gone rather than fixed.

**It is resolved: the `exec` had not happened yet.**

I hit your exact wall — identity probe, same-inode failure, same bafflement —
and instead of choosing, instrumented it. On
[run 33826500237](https://github.com/macanderson/stella/actions/runs/33826500237/job/100880292160)
the assertion printed the accused child:

```
child 23019: exe -> /home/runner/work/stella/stella/target/debug/deps/stella-1c2174eab2f256df
child open descriptors:
  fd 4 -> pipe:[38517]
  fd 5 -> pipe:[38517]
  fd 25 -> /tmp/.tmpgOJS59/context.db
  fd 28 -> /tmp/.tmpkrKMdr/.stella/private/codegraph.db-wal
  fd 14 -> socket:[37377]
  ... eventfds, eventpolls, more sockets and SQLite files ...
parent: read_fd 4 -> pipe:[38517] | write_fd 5 -> pipe:[38517]
```

`/proc/<pid>/exe` names **the test binary**, not `/bin/sh`, and the fd table is
**the parent's own** — our sockets, our SQLite handles, our event descriptors.
The child was forked and not yet through `execve`, so it held a copy of
everything we had, both pipe ends included. `FD_CLOEXEC` was set the whole time
and not yet due to act, which is precisely why the `fcntl` assertions passed on
every red run.

The bad premise is in the test's comment, from #5793, and every repair since
inherited it:

> `Command::spawn()` returns only once the child's `execve()` has finished

All three attempts — yours and mine — changed *what* was compared and left
*when* alone.

## What that does to this PR's tradeoff

You wrote the cost honestly:

> Nothing observes the consequence in a live child any more. A change that set
> the flag and still leaked the descriptor would pass. That is a real reduction
> in what this test proves.

That reduction was justified by the question being open. It is not open any
more, so the choice is now between a test that proves the consequence and one
that proves only the flag — on the merits, not under duress.

#5816 keeps the child assertion and fixes the timing: the child reports its own
descriptors (`readlink /proc/self/fd/<n>`), and `output()` waits for it to exit,
so no pre-`exec` state remains to observe. The window is removed rather than
narrowed, and the comparison against the parent's recorded targets keeps the
identity property your `2b9365a4` head established as the right question.

## On your verification point — you were right, and I have now met it

> A filtered run reproduces none of these failures, which is exactly how #5808
> shipped its bug: 14 green filtered tests read as proof.

Fair, and it applied to me: my earlier evidence was a `credential_handoff::`
filter. Re-run on the condition that actually reproduces, the full crate binary:

```
cargo test -p stella-cli --bin stella   ×3
  run 1: 2664 passed; 1 failed — credential_handoff::tests::cloexec_… ok
  run 2: 2664 passed; 1 failed — credential_handoff::tests::cloexec_… ok
  run 3: 2664 passed; 1 failed — credential_handoff::tests::cloexec_… ok
```

The cloexec test passes in all three with the child assertion intact. The single
failure in each is
`plugin_cmd::tests::a_project_tier_that_will_not_delete_does_not_strand_the_user_copy`
— the root-only, pre-existing failure **you** identified and confirmed against a
clean tree. Independent confirmation of that finding, for what it is worth.

## They cannot both land

Both rewrite the same test body; whichever merges first conflicts the other, and
this PR also renames the test, so a later rebase has a rename and a body change
to reconcile. Worth choosing deliberately rather than by merge order.

My read, stated as an interested party: #5816 preserves what #2603 was written
to catch and now rests on a captured mechanism rather than an argument. If a
maintainer would rather take the smaller, already-green change here and revisit
coverage later, that is entirely defensible — but it should be chosen knowing
the question is answered, not deferred because it is not.

I have not touched this branch.

---

macanderson:
## Three open PRs are fixing this same test — only one can land

Flagging a collision rather than adding a fourth fix. #5810, #5815 and #5816 were
each opened within about five hours, and all three edit exactly one file —
`crates/stella-cli/src/credential_handoff.rs` — to repair the same failing test,
`credential_handoff::tests::cloexec_pipe_closes_a_concurrent_childs_inherited_copy`
at line 474. This is the shape AGENTS.md records from #4680, where three sessions
each wrote the same repair and merged them 95 seconds apart.

| PR | Approach | Diff | CI |
|---|---|---|---|
| #5810 | probe the hardener's fd closure per-process | +44/−24 | **red** |
| #5815 (this one) | compare fd identity, not fd number | +35/−68 | green, `clean` |
| #5816 | ask the child after it has exec'd, not the parent before | +91/−28 | green except `dod` |

**#5810 does not fix it.** Its own `fmt + clippy + test` run fails on the same
test at the same line:

```
thread 'credential_handoff::tests::cloexec_pipe_closes_a_concurrent_childs_inherited_copy'
panicked at crates/stella-cli/src/credential_handoff.rs:474:13
test result: FAILED. 2664 passed; 1 failed
```

**The failure is nondeterministic, and here is the evidence rather than the
word "flake".** The last `ci.yml` run on `main` failed on this test at commit
`f0201f75`. That commit's files are `doctor.rs`, `doctor_fleet.rs`, `main.rs`,
`tests/doctor_cli.rs`, `tests/model_config_cli.rs` — it does not touch
`credential_handoff.rs` at all. The immediately preceding commit `b2a074dd`
("assert the cloexec pipe's mechanism, not a racy guess"), which *is* the commit
that last changed this test, passed. Same test code, green then red, on a commit
that cannot have affected it. So the test is racy on the current `main`, and
`b2a074dd`'s de-flaking did not settle it.

**Why this needs one decision rather than three merges.** All three diffs rewrite
overlapping regions of a single file. Whichever lands first, the other two go
stale, and rebasing a stale branch whose edit is already present is exactly the
hazard `rebase-replay.yml` was added for — the deduplicated edit plus a surviving
undo can turn into a silent revert of the fix that landed. After one merges, the
other two want closing or flattening, not a routine rebase.

On the evidence above, #5815 is the one that is ready: green, `mergeable_state:
clean`, and the smallest change (net −33 lines). #5816 is green on code and red
only on `dod`, i.e. its linked issue's checklist, so it is a viable alternative
if its diagnosis is judged the better one — that is a judgement about which
mechanism is actually right, which belongs to a human, not to this pass.

I have not pushed a fix, rebased anything, or merged. Also worth noting for
whoever picks: `main`'s current tip `4de19a83` is a `[skip release]` version-sync
commit with no `ci.yml` run, so the newest actual CI verdict on `main` is
`f0201f75`'s failure, and no `main-red` canary issue is open — every surface
currently reads green while the last verified state of `main` is red.

---

macanderson:
## This PR is green and mergeable, and #5812 says it must not merge on its own

Flagging it here because the constraint is invisible from this PR. #5812's last definition-of-done item:

> **Reconcile with #5814/#5815 before either merges.** […] Only one should land — merging both would double-edit `credential_handoff.rs`'s cloexec test and one would silently clobber the other's approach. A maintainer call, not a triage one.

This PR links #5814, which was closed as a duplicate of #5812. So the `dod` gate has no open checklist to read here and reports success ([run 33826807297](https://github.com/macanderson/stella/actions/runs/33826807297) on `0599189`). That green is the absence of a question, not an answer to one. #5816's `dod` is red for the mirror-image reason: it links #5812, whose reconciliation box is unticked.

The asymmetry is what makes this worth a comment. Right now #5815 is `mergeable_state: clean`, every check green, no `hold` label — while #5816 carries the same undecided question as a visible red check. Nothing mechanically stops this one from being merged first, and merging it would settle #5812's maintainer call as a side effect.

The two approaches are not interchangeable, per #5812 and this PR's own "What this gives up" section: #5815 drops the child-spawn probe and asserts only the `FD_CLOEXEC` flags; #5816 keeps the probe and repairs the pre-`exec` timing bug that #5812's captured instrumentation identified.

### Surrounding state, as of this comment

- **`main` is red.** The newest `ci` run on `main` is [33821298562](https://github.com/macanderson/stella/actions/runs/33821298562) at `f0201f75` — `cargo test` failed on `cloexec_pipe_closes_a_concurrent_childs_inherited_copy`.
- **The current head is unverified.** `main` has since advanced to `4de19a83` (the #5811 version-sync) with no `ci` run against it at all — #5817. `process_holds_fd` and its two call sites are still present in `origin/main`'s `credential_handoff.rs`, so the defect is in the tree rather than merely in a past run.
- **Nothing blocks a merge onto it.** No open `main-red` issue exists, so `main-red-hold.yml` passes; the canary runs `cargo check`, not the suite.
- **#5810** stays parked behind whichever of these two lands.
- **Sourcery is rate-limited** on this and every open PR in the org (~3 days), so no automated assessment is coming for either approach.

No change pushed to either PR and neither merged — the pick is #5812's maintainer call, not an automated one.

---

macanderson:
## Withdrawing this PR in favour of #5816

The open question this PR shipped as residue has been answered, and the answer invalidates the approach here.

This PR's own body says: *"The child genuinely holds the same pipe, and I do not know why."* #5816 found out, by instrumenting the test instead of reasoning about it. From [run 33826500237](https://github.com/macanderson/stella/actions/runs/33826500237/job/100880292160), quoted on #5812:

```
child 23019: exe -> /home/runner/work/stella/stella/target/debug/deps/stella-1c2174eab2f256df
child open descriptors:
  fd 4 -> pipe:[38517]
  fd 5 -> pipe:[38517]
  ...
parent: read_fd 4 -> pipe:[38517] | write_fd 5 -> pipe:[38517]
```

`/proc/<pid>/exe` still names the test binary, and the descriptor table is the parent's own. The child was forked and had **not** reached `execve`, so it held a copy of everything — both pipe ends included. `FD_CLOEXEC` was set the whole time and simply not due to act yet.

So the premise all three repairs inherited from #5793 — that a returned `spawn()` means a completed `execve` — was the defect. My identity probe at `2b9365a4` was asking the right question at the wrong time, which is why fixing the comparison changed nothing.

**That makes this PR the wrong trade.** It removes the child probe to escape a failure that had a fixable cause, and my own description records the cost: *"A change that set the flag and still leaked the descriptor would pass. That is a real reduction in what this test proves."* #5816 keeps the probe, has the child report its own descriptors after `exec` with `output()` waiting for exit, and its `fmt + clippy + test` is **success** on `dc96697` ([job 100884808573](https://github.com/macanderson/stella/actions/runs/33828017007/job/100884808573), 02:02:52 → 02:23:18). It gives up nothing.

#5812's checklist records that only one of these may land, since both rewrite the same test body and the second merge would silently clobber the first — the #4979 shape. It calls that a maintainer call. Withdrawing my own weaker change is not that call: it removes the contested edit rather than deciding between two live ones, and it leaves #5816 as the only branch touching this test. If the flag-only approach is wanted after all, this can be reopened — the branch is intact.

Also correcting one line in the description above, since it is wrong on the record: it credits #5809 with the approach and says its author was right. #5809's *proposal* was reasonable when the cause was unknown; now that the cause is known it is not the better option, and #5816's author says so on their own PR having earlier argued the other way. I am not leaving my endorsement standing over a conclusion the evidence has moved past.

**#5814** stays closed as a duplicate of #5812 (closed 02:12:47Z, `state_reason: duplicate`) — closing this PR orphans nothing.

Not merging anything here: no approving review stands on either PR, and `main` is still red at `f0201f75` pending whichever lands.

---

Issue #1655, opened 2026-08-06
stella doctor: report the fleet worktree count/size and how much stella fleet clean would reclaim


macanderson:
## Problem

`stella fleet clean` (PR for #1217) reclaims finished isolated worktrees and `fleet/*` branches, but nothing makes the accumulation *visible* before a user thinks to run it. #1217's own direction named this half explicitly and it was left out of that PR:

> `stella doctor` could report the current count/size so the accumulation is at least visible before the command exists.

A workspace can be carrying dozens of stale worktrees and several GB with no signal anywhere.

## What to do

Add a `stella doctor` check that reports, for the current workspace:

- how many worktrees live under `.stella/worktrees/`, and their total on-disk size;
- how many of them are reclaimable by the default `stella fleet clean` rules (no unfinished attempt in the ledger, clean tree, branch contained in the base ref) versus kept;
- how many loose `fleet/*` branches exist.

The classification must not be re-implemented: call `stella_fleet::Gc::sweep` with `GcOptions { dry_run: true, ..default }` and count the `GcReport` verdicts. That is exactly what the flag exists for, and it keeps one definition of "reclaimable".

## Files

- `crates/stella-cli/src/doctor.rs` — where checks live. Mind the file-size ratchet (`scripts/file-size-baseline.txt`); if it is near the ceiling, put the check in a sibling module and call it.
- `crates/stella-fleet/src/gc.rs` — `Gc`, `GcOptions`, `GcReport`, `reclaimed_worktrees()`, `reclaimed_branches()`.
- `crates/stella-cli/src/fleet_gc.rs` — how the CLI builds the ledger activity slice (`Ledger::worktree_activity`) and the `Gc`; reuse that shape.

## Constraints

- A dry-run sweep shells out to `git` several times per worktree. `doctor` should stay fast: consider reporting count+size unconditionally and the reclaimable/kept split only when the count is above a small threshold, or behind an explicit opt-in. Decide and say which in the PR.
- Never mutate anything from `doctor` — the sweep must be `dry_run: true`.
- A workspace with no `.stella/worktrees/` and no `fleet.db` must produce no noise.

## Definition of done

- [x] `stella doctor` reports, for the current workspace, the count of isolated worktrees under `.stella/worktrees/` and their total on-disk size, unconditionally.
[...]

PR #5816, opened 2026-09-04
fix(stella-cli): ask the child after it has exec'd, not the parent before

macanderson:
## What & why

`cloexec_pipe_closes_a_concurrent_childs_inherited_copy` has taken `main` red
three times. Every repair so far changed *what* the probe compared and left
*when* it asked alone — and the timing was the defect.

The parent read `/proc/<child>/fd` immediately after `spawn()` returned, on a
premise written into the test's comment by #5793 and carried unchanged through
#5808 and this PR's first revision: that a returned `spawn()` means a completed
`execve`. It does not.

**A captured failure settles it.** The first commit here added instrumentation;
it fired on its first run —
[run 33826500237](https://github.com/macanderson/stella/actions/runs/33826500237/job/100880292160):

```
child 23019: exe -> /home/runner/work/stella/stella/target/debug/deps/stella-1c2174eab2f256df
child open descriptors:
  fd 4 -> pipe:[38517]
  fd 5 -> pipe:[38517]
  fd 25 -> /tmp/.tmpgOJS59/context.db
  fd 28 -> /tmp/.tmpkrKMdr/.stella/private/codegraph.db-wal
  fd 14 -> socket:[37377]
  ... eventfds, eventpolls, more sockets and SQLite files ...
parent: read_fd 4 -> pipe:[38517] | write_fd 5 -> pipe:[38517]
```

`/proc/<pid>/exe` still names **this test binary**, not `/bin/sh`, and the table
is *this process's own*. The child was forked and not yet through `execve`, so it
held a copy of every descriptor we had — both pipe ends included. `FD_CLOEXEC`
was set on both the whole time and simply not due to act yet, which is why the
`fcntl(F_GETFD)` assertions passed on every one of these runs.

Nothing was ever wrong with `cloexec_pipe`.

## The fix

The child answers for itself, and `output()` waits for it to exit — so there is
no pre-`exec` state left to observe. The window is removed rather than narrowed.

It reports what `/proc/self/fd/<n>` names for each end; the test compares that
against what the parent recorded while holding both ends open. That keeps the
identity property from the previous revision: a descriptor the child opened at
the same number names something else and reads as a pass.

`fd_target` stays, on `"self"`, as the parent's yardstick. `child_fd_report` goes
with the parent-side probe it served.

Closes #5812, Closes #5813

## The witness

- [x] Witness test (fails on the old behaviour, passes here)

Isolated by ablation, because the flag assertions run first and would otherwise
[...]

Issue #5310, opened 2026-08-27
Epic: parallel sessions coordinate — leases, worktrees, and branches are visible and bounded


macanderson:
## Outcome

Several Stella sessions running at once on one machine — a fleet fan-out, a
self-driving loop, and a human at the deck — can see each other's work. A lease
says who is racing, a branch says who is working, a worktree says where, and none
of the three can silently destroy another session's uncommitted files.

## Evidence this is not true today

- **A branch switch reverts a peer's tracked edits.** Three sessions shared one
  checkout and four branch switches in six minutes reverted every uncommitted
  tracked file between them, leaving a tree holding a module nothing declared
  (#5287). Untracked files survived, so the failure presents as an unbuildable
  tree with no error naming the cause.
- **A deck start-work branch is invisible to a peer on another clone** until it is
  pushed, so a loop on a second clone claims the same issue (#5275).
- **The dispatch lease is a hard-coded 15 minutes** with nothing able to tune or
  observe it (#1678), and it has no human-visible surface at all (#1675).
- **Worktrees are removed by two different code paths** in `stella-fleet` (#1656),
  and nothing reports how much disk they are holding (#1655).

## Children

| Issue | What |
|---|---|
| #5287 | Parallel sessions share a checkout; a branch switch reverts peers' work |
| #5275 | A deck start-work branch is invisible to a peer until pushed |
| #1678 | The dispatch-lease TTL is a hard-coded 15 minutes |
| #1675 | No human-visible surface for dispatch claims |
| #1656 | Two worktree-removal paths — consolidate `WorktreeManager` |
| #1655 | `stella doctor` does not report worktree count or disk |

## Definition of done

- A session cannot revert another session's uncommitted tracked files through an
  ordinary branch switch, and the guard that prevents it has a witness test.
- Claims, branches and worktrees are readable from one command, on one clone and
  across clones.
- The lease TTL is a named, documented value rather than a literal, and a lapsed
  lease is distinguishable from a held one by a reader.

Related: #4804 owns the *red main* half of session coordination; this epic owns
the *work in flight* half.

macanderson:
**Triage: this epic is P1, Impact 5.**

P1 is inherited from its highest child (#5287, silent revert of a peer's
uncommitted work). The other five children are P2/P3, and filed apart they read
as fleet housekeeping; together they are one outcome — several sessions on one
machine can see each other's work — and #5287 is what makes that outcome urgent
rather than tidy.

[...]

Issue #5117, opened 2026-08-26
re-layering: extract stella-records — the record plane (records/ + context_record/ + ingest/) leaves the engine; record_hash moves to stella-protocol


macanderson:
Child of #5113. The single largest reduction available: ~14.8k LOC (v0.9.40
measurement — re-measure) whose entire attachment to the step path is one
hash call.

## Evidence, verified on main today

- `crates/stella-core/src/{records,context_record,ingest}` all still present.
- The record plane already has **cross-crate consumers that treat it as a
  library**: `stella-context/src/store/ledger.rs` stores "the ADR 0004
  canonical hash", `stella-cli/src/memory/{uses,evidence,observations}.rs`
  call `record_hash` and `LIFECYCLE_SCHEMA_VERSION` through
  `stella_core::context_record`. It is a subsystem renting a room in the
  engine.
- `stella-protocol/src/context_event.rs` already documents the coupling from
  the other side: its canonical bytes "are the preimage
  `stella-core::context_record::hash` builds `record_hash` values from",
  pinned by tests. The hash primitive is *already* protocol-adjacent in
  everything but its `use` path.

## What to do

1. New crate `stella-records` (depends on `stella-protocol` only — keep it
   heavy-dep-free; today's plane is, since rusqlite lives in
   `stella-context`/`stella-store`, not here — verify and keep it that way).
2. Move `records/*`, `context_record/*`, `ingest/*` wholesale, tests
   included.
3. `context_record/hash.rs` (the ~190-line JCS+SHA-256 primitive) moves to
   `stella-protocol` as `stella_protocol::hash::record_hash`, beside
   `context_event.rs`'s canonical-bytes machinery — this *removes* a
   cross-crate doc-comment promise by making it a same-crate fact. This is
   the one justified bend of protocol's zero-logic rule; it spends the
   budget, which is why the sibling issue refuses `starvation.rs` the same
   ride.
4. `stella-core` keeps zero knowledge of records: `receipts.rs`'s
   `frame_hash` call goes through `stella_protocol::hash::record_hash`.
5. Temporary `pub use` shims in `stella-core` are acceptable for one release
   to keep `stella-cli`/`stella-context` compiling, with a removal note; or
   fix the ~6 consumer files in the same PR if the diff stays reviewable.

## Witness (the byte-identity test is the whole ballgame)

A fixture record's `record_hash` computed via the new path is **byte-identical**
to the value stored by today's path — ADR 0004 hash-neutrality holds because
[...]

PR #5823, opened 2026-09-05
feat(stella-fleet): a worker gets its own worktree by default

### LINKED ISSUE #5287: Parallel agent sessions can share one checkout, and a branch switch silently reverts another session's uncommitted work
## Problem

Several agent sessions dispatched in parallel can land in **one shared checkout**, and a branch switch by any of them silently reverts every other session's uncommitted tracked files. Untracked files survive. The result is a working tree holding a new module with nothing declaring it — code that cannot build, and no error anywhere saying why.

This happened tonight while working the TUI v2 backlog. Three sessions shared `.claude/worktrees/2673d1ec-issue-labels`. `git reflog --date=iso` recorded the collision plainly:

```
00:00:02  checkout: moving from feat/panel-grant to fix/main-red-prose-density
23:58:53  checkout: moving from fix/5072-tab-row-width-ladder to feat/panel-grant
23:57:11  checkout: moving from feat/5043-gate-failure-proposal to fix/5072-tab-row-width-ladder
23:54:31  checkout: moving from merge/5032 to feat/5043-gate-failure-proposal
```

Four branch switches in six minutes, none of them aware of the others.

**What it cost:** the #5072 work lost its three tracked edits (`crates/stella-tui/src/views/frame.rs`, `crates/stella-tui/tests/deck_render_snapshots.rs`, `crates/stella-tui/tests/deck_snapshot.rs`) while keeping the new `views/frame/tab_list.rs` and four goldens. Salvaged to branch `wip/5072-tab-row-salvage`. The #5043 session read the reflog, recognised the hazard, and **stood down without writing anything** — the correct call, and the reason nothing was corrupted rather than merely lost.

## Why the sessions could not fix it themselves

The #5043 session tried. Its report:

> My session is hard-pinned to one checkout — the Bash sandbox refuses any command whose cwd or `git -C` resolves elsewhere, and `EnterWorktree` moves the agent cwd but not the Bash pin, so a dedicated worktree I created was unusable and I removed it.

So a session that *detects* the collision has no move available except to stop. Isolation has to be granted at launch; it cannot be taken afterwards.

## What makes this worse than it looks

- **It is silent.** No conflict, no error, no dirty-tree warning. `git status` after the fact shows a plausible-looking set of untracked files.
- **It is asymmetric.** Untracked files survive and tracked edits do not, so the surviving half is always the half that cannot compile alone — a new module with no `mod` declaration.
- **The recovery is invisible too.** A session re-applying its lost edits looks exactly like a session doing ordinary work, so a reader following along cannot tell the difference.
- **Squash merges compound it.** One of these branches sat one commit "ahead" of `main` carrying work already merged as #5129 under a different SHA, so `git log origin/main..HEAD` read as unmerged when it was not.

## Where it lives

- Whatever dispatches parallel agent sessions and assigns their working directory
- `.claude/worktrees/` — the per-session worktree convention
- Related: **#5224** (no issue-claim mechanism, so two sessions take the same issue) and **#5265** (`main-red-claim.sh` keys on author, and every agent session shares one GitHub identity). This issue is the third member of that family: #5224 is two sessions taking one *issue*, #5265 is two sessions taking one *repair*, and this is two sessions taking one *working tree*.

## Possible shapes, for whoever takes it

1. **Give every dispatched session its own worktree at launch.** Isolation the session cannot lose, and the only option that removes the hazard rather than reporting it.
2. **Refuse the second session in a tree.** A lock file naming the holding session; a second arrival stops with an actionable message instead of quietly reverting files.
3. **Warn on entry.** Cheapest, and weakest — a session that reads the warning still has no way to act on it, as this session's own report shows.

I would take (1): the failure mode here is silent data loss, and a guard that only reports it leaves the detecting session with no move.

## Done means

Two sessions dispatched at the same moment cannot revert each other's uncommitted work. Witnessed the way this repository witnesses a guard — a hermetic self-test in the shape of `guard-self-tests.yml`'s suites, dispatching two sessions against one repository and asserting both trees survive, so the fix is demonstrated rather than asserted.

## Definition of done

The "Done means" prose above, as boxes the gate can read — nothing here is

- [x] Two sessions dispatched at the same moment cannot revert each other's uncommitted work.
- [x] A hermetic self-test in the shape of `guard-self-tests.yml`'s suites dispatches two sessions against one repository and asserts both trees survive.
- [x] That test fails before the change, so the fix is demonstrated rather than asserted.

Box 3 was left unticked while #5823 only *witnessed* the opt-in isolated path,
which already existed — there was no before/after because the dispatch default
had not moved. That is no longer the state of the branch. #5823 at `5ae27e6`
moves `#[default]` from `Isolation::SharedTree` to `Isolation::Isolated` in
`crates/stella-fleet/src/plan.rs` and makes `Task::new` read
`Isolation::default()`, so a plan naming nothing gets a worktree per task.

Checked against `origin/main` rather than against the PR description:
`git grep -n default origin/main -- crates/stella-fleet/src/plan.rs` puts
`#[default]` on `SharedTree` at line 48, and main's
`plan_json_roundtrips_with_isolation_defaulting_to_shared_tree` pins that
default. The branch's replacement witness
`a_task_that_names_no_isolation_gets_its_own_worktree` asserts `Isolated` for
both the deserialized plan and `Task::new`, so it fails on main by construction
and passes on the branch — the fail→pass pair box 3 asks for. The first case of
`scripts/test-session-isolation.sh` runs the real CLI with no isolation named
anywhere and is the same pair at the integration level.

macanderson:
## What and why

Two agent sessions dispatched into one checkout are one branch switch away from
losing each other's work. The switch puts the tracked files back and leaves the
untracked ones alone, so the tree ends up holding a new module that nothing
declares — code that cannot build, with no error anywhere saying why.

Isolation belongs to whoever hands out the working directory, because a session
cannot take a tree it was not given (SCR-002, option 1 of the issue's three
shapes, and the one its author recommended). **`Isolation::Isolated` is now the
fleet default.** Every `stella fleet` task gets a `git worktree` of its own
unless the plan names `isolation = "shared_tree"`.

The earlier revision of this PR witnessed `isolation = "isolated"` and left
`SharedTree` as the default, which is what Sourcery's three ❌ rows were about
and they were right — see the answer below.

## The change

- `crates/stella-fleet/src/plan.rs` — `#[default]` moves to `Isolation::Isolated`,
  and `Task::new` reads `Isolation::default()` rather than repeating the choice,
  so a plan file that names nothing and a builder that says nothing cannot
  disagree.
- `crates/stella-cli/src/fleet_cmd/isolation_notice.rs` (new) — one line in the
  run header saying where this run's workers will write. `fleet_cmd.rs` sits at
  its 1500-line ceiling, so the logic is a sibling submodule and the file gains
  a `mod` line and one call, paid for by trimming two lines of its own prose
  (AGENTS.md § God files). `check-file-size` passes.
- `docs/adr/0027-a-fleet-worker-gets-its-own-worktree.md` (new) — the decision,
  what it costs, and what it does not cover.
- Docs that stated the old default: `crates/stella-fleet/README.md`, its
  `Cargo.toml` description, `docs/why-stella.md`, and four website pages
  (`agent-fleets`, `commands/fleet`, `agent-modes`, `agent-engine-paths`).

## The witness and why it fails on the old code

`scripts/test-session-isolation.sh` gains a first case that runs the real CLI
with **two positional prompts and no isolation named anywhere** — no plan file,
nothing asking for a worktree. It then runs the loss itself on the trees the
dispatch handed out: an uncommitted edit plus an untracked file in tree A, a
`git checkout --force -B` in tree B, and a look at whether the edit is still
there.

On `main` today that command puts both workers in the checkout itself. The case
finds zero worker trees, fails its first assertion, and the forced switch then
reports the edit `lost`. It cannot pass before the default moves — the mechanism
is the experiment, not an assertion about inequality.

That last paragraph is **read from `main`'s code, not run there**: `plan.rs`
puts `#[default]` on `SharedTree` and `Task::new` writes
`Isolation::SharedTree`, and `dispatch_claimed` allocates a worktree only for
`Isolation::Isolated`. A workspace build is not something this session spends
(CLAUDE.md, "CI builds and tests; this laptop does not"). What was run is this
branch's own CI, quoted in the comment below.

Two Rust witnesses beside it:

- `crates/stella-fleet/src/plan.rs`'s
  `a_task_that_names_no_isolation_gets_its_own_worktree` — both assertions read
  `SharedTree` on the old code.
- `crates/stella-cli/src/fleet_cmd/isolation_notice.rs`'s
  `a_plan_naming_nothing_reports_a_worktree_per_task` — the module does not
  exist on the old code.

The suite keeps its control: the last case runs the same experiment on a plan
that names `shared_tree` and still reports `lost kept`, the exact asymmetry the
issue describes. A suite that can only pass is not a suite.

## Cost, stated out loud

A worktree per worker costs disk and a cold build cache per tree, and it defers
conflicts to merge time. `stella doctor` reports the fleet worktree count and
what is reclaimable (#1655); `stella fleet clean` reclaims the finished ones.

A plan that declares `claims` and names no isolation changes behaviour: it now
gets worktrees, where its claims coordinate nothing anybody shares. Such a plan
wants `isolation = "shared_tree"` beside them. The README and the ADR both say
so, and the header line prints which mode ran.

## Tests changed to name the shared tree

Three test files depended on the old default and now say what they mean. Each
runs a fleet attempt and then reads the **invocation workspace's** store, which
is only the worker's store when the worker's tree is the workspace:

- `crates/stella-cli/tests/fleet_wrapped_dispatch_cli.rs` — all three cases.
- `crates/stella-cli/tests/settings_warnings_cli.rs` — `fleet_run`.

## The deck half

Unchanged from the earlier revision: `command_deck` reads the session registry
at start and prints one line naming any live peer in the same checkout and the
branch it holds (`crates/stella-cli/src/command_deck/shared_checkout.rs`). It
never refuses — a session told to stop has nowhere to go, since the working
directory is picked at launch.

## Answer to Sourcery's three ❌ rows

- **"The PR does not change the dispatch implementation … shared-tree dispatch
  remains the fleet default."** Correct, and fixed: the default is
  `Isolation::Isolated` now, so two concurrently dispatched workers cannot share
  a checkout unless a plan asks them to.
- **"The new test does not demonstrate a failing pre-change behavior."**
  Correct, and fixed: the suite's first case names no isolation and fails on
  `main` by construction, as does each of the two Rust witnesses.
- **"The command-deck message is only a warning."** Still true, and it is not
  what carries the guarantee. The guarantee is the dispatch handing out the
  tree. The deck line covers the case nothing here can fix — a person, or an
  outside harness, starting two decks in one tree — where the working directory
  was chosen at launch and refusing would leave the session with nowhere to go.
  The ADR states that boundary.

## Exemplar

`crates/stella-cli/tests/fleet_wrapped_dispatch_cli.rs` for the hermetic
"real binary against a stub endpoint, every credential cleared" envelope, and
`scripts/test-self-driving.sh` for the `locate_stella` / pinned-`STELLA_BIN`
shape.

Tests deleted: `plan_json_roundtrips_with_isolation_defaulting_to_shared_tree`
in `crates/stella-fleet/src/plan.rs`, split into
`a_task_that_names_no_isolation_gets_its_own_worktree` (the witness) and
`the_shared_tree_is_an_opt_in_a_plan_has_to_spell` (its round-trip half, kept
whole). No other test was removed.

Closes #5287

## Summary by Sourcery

Make isolated Git worktrees the default for fleet workers while retaining shared-tree execution as an explicit opt-in.

New Features:
- Give each fleet task its own Git worktree by default, with shared-tree execution available through an explicit `shared_tree` setting.
- Add fleet run-header messaging that identifies whether workers use dedicated worktrees or the shared checkout.
- Warn interactive deck sessions when another live session is already using the same checkout, including the peer and branch.

Bug Fixes:
- Prevent concurrently dispatched fleet workers from silently losing each other's tracked uncommitted changes after a branch switch.
- Make task-builder and deserialization defaults consistently select isolated worktrees.

Enhancements:
- Document the new isolation policy, trade-offs, and limitations in the fleet documentation and ADR 0027.
- Update existing fleet tests and isolation coverage to explicitly opt into shared-tree behavior where they depend on the invocation workspace.
- Add a hermetic session-isolation test that exercises the real CLI, validates the default behavior, and preserves a shared-tree control case.

CI:
- Run the session-isolation guard test in CI and expose a Makefile target for it.

Documentation:
- Update CLI, fleet, architecture, and website documentation to describe worktree-per-task as the default and shared-tree execution as an opt-in.

Tests:
- Add unit coverage for isolation notices, shared-tree opt-in behavior, and peer-session detection.

Chores:
- Register ADR 0027 in the ADR indexes and documentation manifest.


macanderson:
## DoD verification: the `dod` red is correct, and Sourcery's three ❌ rows hold

Checked the diff against `origin/main` rather than against the description. All three of #5287's DoD boxes are genuinely unmet, so ticking them would be wrong and the gate is doing its job.

**The diff changes no dispatch code.** The six changed files are `ci.yml`, `Makefile`, `command_deck.rs`, `command_deck/shared_checkout.rs`, `crates/stella-fleet/README.md`, and `scripts/test-session-isolation.sh`. `crates/stella-fleet/src/fleet.rs` is not among them.

**Item 1 — "two sessions dispatched at the same moment cannot revert each other's uncommitted work": not met.** The isolation this PR witnesses already exists and is opt-in. On `origin/main`, `crates/stella-fleet/src/fleet.rs`'s `dispatch_claimed` reads:

```
let worktree = match task.isolation {
    Isolation::Isolated => Some(
```

and that file's own module doc says "the shared tree by default; a git worktree when the task opts into isolation". Shared-tree remains the default, and this PR's **own control case asserts the loss still happens there** (`lost kept`). Two workers dispatched under the default still share a checkout and still revert each other. The only behavior this PR adds is `shared_checkout::announce`, which prints one line and never refuses — option 3 of the issue's three shapes, the one the issue calls "cheapest, and weakest", and rejects on the grounds that "a session that *detects* the collision has no move available except to stop."

**Item 3 — "that test fails before the change": not met, and this is the load-bearing one.** Because `Isolation::Isolated` already cuts a worktree on `main`, `scripts/test-session-isolation.sh` passes on `main` unchanged. There is no fail→pass flip, which is precisely what AGENTS.md § "The definition of done: witness tests" requires ("fails on `main` without your change"). The PR description is candid that the guarantee predates it — "Stella's own dispatchers already hand one out" — which is the same fact stated from the other side.

**Item 2 — partially met.** `scripts/test-session-isolation.sh` is genuinely hermetic, genuinely drives the real binary against a loopback stub with every credential cleared, and genuinely runs the loss experiment on the trees the dispatch handed out rather than comparing two paths. The control case is real and is what makes the suite falsifiable. That is good work and worth landing. It witnesses the *isolated* path only, which is not the path the issue's incident took.

**And the incident in #5287 is not the fleet path at all.** Three sessions shared `.claude/worktrees/2673d1ec-issue-labels` — an outside harness's directory. This PR's README addition says so plainly: "An outside harness has to isolate at launch. Nothing here decides where a person, or another tool, starts `stella`." So the reported data loss remains reachable after this merges.

### Recommended disposition

This PR should land — the witness, the control, the deck warning and the README are all real value — but it should not close a P1 data-loss issue whose failure mode it leaves reachable. Per the gate's own second remedy:

1. Change `Closes #5287` to `Refs #5287` in the description. That clears the `dod` check without asserting anything untrue.
2. Retitle nothing — the current title ("witness that two dispatched workers get their own trees") already describes exactly what this does, and does not overclaim.
3. File the remainder as a new issue under epic #5310 with the `triage` label only (SCR-004/SCR-005): make isolation the default for concurrent dispatch, or refuse the second session in a tree (shapes 1 and 2 from #5287), covering the outside-harness case the README now documents as a gap.

I have not edited the description or ticked the boxes: whether a P1 closes on a partial fix is the maintainer's call, not mine, and silently dropping the close-claim would hide an unfixed data-loss bug rather than track it.

No CI failure here is this PR's own — at the time of writing `fmt + clippy + test`, `guard self-tests`, MSRV and the macOS check are still in flight, and everything already completed is green apart from `dod`.

---

macanderson:
## `dod` is red on #5287, and two of its three boxes are genuinely not met — I am not ticking them

Automated PR sweep. `dod / dod` failed twice on `ac50fbf` ([00:51:45](https://github.com/macanderson/stella/actions/runs/33934340679/job/101219216998), [00:52:07](https://github.com/macanderson/stella/actions/runs/33934352870/job/101219249785)) naming #5287's three unchecked DoD items. I checked each against the diff. One is met; two are not, and no push from me would change that because the gap is a scope decision, not a defect.

**Met — box 2.** *"A hermetic self-test in the shape of `guard-self-tests.yml`'s suites dispatches two sessions against one repository and asserts both trees survive."* `scripts/test-session-isolation.sh` does exactly this: throwaway repo, stub loopback endpoint, every credential cleared with `env -u`, the real `stella fleet --plan` with two isolated tasks at `--max-concurrency 2`, worker trees read back from `git worktree list --porcelain`, then the destructive `git checkout --force -B` run against them. The `ci.yml`-instead-of-`guard-self-tests.yml` deviation is stated in the description and is the right call — that workflow's `self-tests` job is toolchain-free by design and this suite needs the `stella` binary. `check-gate-parity.sh` is satisfied either way. Box 2 is fine.

**Not met — box 1.** *"Two sessions dispatched at the same moment cannot revert each other's uncommitted work."* This PR adds no isolation. It witnesses isolation that was already in the tree for two dispatchers, and the guarantee does not reach the case #5287 was filed about:

- `crates/stella-fleet/src/plan.rs` defaults to `Isolation::SharedTree` (line 112), and `plan_json_roundtrips_with_isolation_defaulting_to_shared_tree` pins that a task with no `isolation` field deserializes as shared. So the *default* fleet dispatch still shares a tree — and this PR's own control case is the proof, reporting the tracked edit `lost` and the untracked file `kept`, the exact asymmetry #5287 describes.
- The incident in #5287 was three sessions in `.claude/worktrees/2673d1ec-issue-labels` — an outside harness. `command_deck/shared_checkout.rs` prints one line naming a live peer and, in its own header, "never refuses". A warning is shape (3) of the issue's three options; the issue says *"I would take (1)"* precisely because a session that reads the warning still has no move.

The README section added here is honest about both gaps and names them as gaps. That is good engineering and it is also why box 1 cannot be ticked: the PR documents the hole rather than closing it.

**Not met — box 3.** *"That test fails before the change, so the fix is demonstrated rather than asserted."* There is no "before" for a fix that predates the PR. `WorktreeManager::create` already cut the worktree in `Fleet::dispatch_claimed`, so the isolated case passes on `main`'s behaviour; the suite is a regression guard, not a witness in this repository's sense (fails on old code, passes on new).

The description's answer to this — the shared-tree control case, which shows the experiment can still fail — is the right instinct and worth keeping. It is not the same claim. It demonstrates *the harness is falsifiable*; box 3 asks that *the change is demonstrated*, and nothing here changed the behaviour under test.

### What would close this

A maintainer's call between two shapes, not something a sweep should decide:

1. **Narrow #5287 to what landed.** Reword box 1 to the guarantee that actually holds (*a task marked `isolation = "isolated"`, and every unit of `stella self-driving work`, gets its own tree*), drop or reword box 3 to name the control case as the falsifiability evidence, and split the residue — the shared-tree default and the outside-harness case — into its own issue, `triage` label only, per SCR-004. #5287 stays the parent under #5310.
2. **Hold this PR** until the dispatch change lands too, so box 1 is true as written.

I have not filed the residue issue myself: #5287 is open and already covers exactly that ground, and filing beside it would duplicate rather than link (AGENTS.md § "Nothing left behind" — search before filing).

Nothing else on this PR is outstanding: no review threads, Sourcery's review was `skipped` so there is no assessment table to answer, and every other check is green (`fmt + clippy + test` included, job 101219278675).

---

macanderson:
## DoD review against #5287 — one box of three, and I did not tick the others

Checked as part of the open-PR sweep. I read the diff and `origin/main` rather
than the description. One box stands; two do not, and the reasons are different.

### ✅ Box 2 — the hermetic self-test exists

`scripts/test-session-isolation.sh` dispatches two workers against one throwaway
repository through the real `stella fleet` path, with a stub chat-completions
endpoint on loopback and `STELLA_HOME` under the temp tree. Case 1 asserts two
trees are cut and that an uncommitted edit in one survives a forced
`git checkout --force -B` in the other. The control case is the right instinct —
a suite that can only pass is not a suite — and the provider-key `env -u` sweep
before the dispatch is careful work.

### ❌ Box 3 — "That test fails before the change"

It does not. `Isolation::Isolated` already cuts a dedicated worktree per task on
`main`:

```
$ git grep -n 'Isolation::Isolated' origin/main -- crates/stella-fleet/src
origin/main:crates/stella-fleet/src/fleet.rs:894:            Isolation::Isolated => Some(
origin/main:crates/stella-fleet/src/plan.rs:15://! [`Isolation::Isolated`] — a dedicated git worktree per task.
```

Case 1 asserts behaviour `main` already has, so it passes there. That makes it a
**characterization test** — valuable, and worth landing — but not a witness in
this repository's sense, and the PR title's "witness that two dispatched workers
get their own trees" reads as the stronger claim. The only genuinely new
behaviour in the diff is `shared_checkout.rs`'s notice, and no case in the shell
suite exercises it (its own six unit tests do, and those *are* witnesses —
`crates/stella-cli/src/command_deck/shared_checkout.rs` does not exist on `main`).

### ❌ Box 1 — "Two sessions dispatched at the same moment cannot revert each other's uncommitted work"

Not as written, and the diff says so itself in two places. Case 2 of the suite
asserts that two shared-tree workers **do** lose the tracked edit (`"lost kept"`
is the passing outcome), and `crates/stella-fleet/README.md` names both gaps:
shared is the default and it shares, and an outside harness still has to isolate
at launch.

That second gap is the one the incident was actually about. #5287's reflog is
from `.claude/worktrees/2673d1ec-issue-labels` — the agent-session dispatcher,
not `stella fleet`. Nothing in this repository decides where that harness starts
a session, so no change here can close box 1 as phrased. `shared_checkout.rs`
answers the issue's shape 3 ("warn on entry"), which the issue itself ranks
weakest and for the reason the notice's own header repeats: a session told to
stop has nowhere to go.

### What I suggest, rather than ticking

This is good work that closes less than the issue asks. Two ways forward, and
the choice is the author's:

1. **Narrow the claim.** Retitle to something like `test(stella-fleet):
   characterize per-task worktree isolation, and warn on a shared checkout`,
   change `Closes #5287` to `Refs #5287`, and file the remaining scope — the
   agent-session dispatcher granting isolation at launch, which is the issue's
   own preferred shape 1 — as a child of #5310 with the `triage` label.
2. **Rewrite #5287's DoD boxes** to the checkable form this PR does satisfy, the
   way #3905 and #5117 were rewritten when their prose done-checks turned out
   unmeetable as stated, and move the untouched half to a new issue.

Either settles the `dod` check honestly. I have deliberately not ticked anything
on #5287, and have not touched this PR's title or body — the scope call is
yours, not the sweep's.

---

macanderson:
## The `cargo test` red was inherited from a stale base, not this branch — base merged, now green

`fmt + clippy + test` failed on `ac50fbf` with a test this diff does not touch:

```
credential_handoff::tests::cloexec_pipe_closes_a_concurrent_childs_inherited_copy
panicked at crates/stella-cli/src/credential_handoff.rs:474:13:
a child spawned while the pipe was open still held a live copy of the read end
test result: FAILED. 2671 passed; 1 failed
```

Established rather than assumed. This PR was cut from `4de19a83`, and:

```
$ git merge-base --is-ancestor 4de19a83 origin/main   # -> yes
$ git log --oneline 4de19a83..origin/main
6d5fb8f fix(stella-cli): ask the child after it has exec'd, not the parent before (#5816)
```

The **only** commit on `main` after this branch's base is #5816 — the fix for precisely that test. The branch was opened at 00:51:29Z and #5816 merged at 00:51:40Z, eleven seconds later, so it missed the fix by a hair.

I merged `main` in (`33e9ca35`, a merge commit — no rebase, no force-push, so anyone's checkout stays valid). `fmt + clippy + test` is **success** on the new head, which confirms the diagnosis rather than just asserting it. 16 checks green, 2 skipped.

**The one remaining red is `dod`, and it is correct.** #5287's three boxes are genuinely unmet — the diff changes no dispatch code, `Isolation::SharedTree` is still the default, and the new harness passes on `main` unchanged, so there is no fail→pass flip. That is unchanged by this merge and is [answered in full above](https://github.com/macanderson/stella/pull/5823#issuecomment-5548223463); the disposition there (`Closes #5287` → `Refs #5287`, remainder filed under epic #5310) is a maintainer's call and I have not taken it.

No code of yours was touched — the merge brought in `credential_handoff.rs` from `main` and nothing else.

---

macanderson:
## Answering Sourcery's three ❌ rows

The table above was written against `33e9ca3`. Two of its three rows were
correct and are now fixed; the third is correct and is not what carries the
guarantee. Head is `5ae27e6`.

**Row 1 — "shared-tree dispatch … is still the fleet default." Fixed.**
`#[default]` moved from `Isolation::SharedTree` to `Isolation::Isolated`
(`crates/stella-fleet/src/plan.rs`), and `Task::new` now reads
`Isolation::default()` rather than naming a variant, so a plan file that says
nothing and a builder that says nothing cannot disagree. Two concurrently
dispatched workers get two trees unless a plan asks for one.

**Row 2 — "the new test does not demonstrate a failing pre-change behavior."
Fixed.** `scripts/test-session-isolation.sh` gains a first case that runs the
real CLI with two positional prompts and **no isolation named anywhere**, then
runs the loss on whatever trees the dispatch handed out. Two Rust witnesses sit
beside it: `a_task_that_names_no_isolation_gets_its_own_worktree` in `plan.rs`,
and `a_plan_naming_nothing_reports_a_worktree_per_task` in the new
`fleet_cmd/isolation_notice.rs`.

That these fail on `main` is read from `main`'s code, not run there — a
workspace build is not something this session spends
(CLAUDE.md, "CI builds and tests; this laptop does not"). The reading:
`origin/main`'s `plan.rs` puts `#[default]` on `SharedTree` and `Task::new`
writes `Isolation::SharedTree`; `dispatch_claimed` allocates a worktree only
for `Isolation::Isolated`. So a positional-prompt run there creates no worker
tree, the suite's first assertion finds zero instead of two, and both Rust
assertions read `SharedTree` where they now require `Isolated`.

What *was* run is on this branch, in
[run 33947707088](https://github.com/macanderson/stella/actions/runs/33947707088),
step 51:

```
two workers, no isolation named
  ✓ the dispatch cut one tree per worker
  ✓ a branch switch in one tree left the other's work alone
two workers, isolated tasks
  ✓ the dispatch cut one tree per worker
  ✓ a branch switch in one tree left the other's work alone
two workers, shared tree (the control)
  ✓ the dispatch cut no tree — both workers ran in the checkout itself
  ✓ one tree loses the tracked edit and keeps the untracked file
  6 passed, 0 failed
```

The last two lines are the control: the same experiment on a plan that names
`shared_tree` still destroys the sibling's edit and keeps the untracked file —
this issue's exact asymmetry, reproduced. A suite that can only pass is not a
suite.

**Row 3 — "the command-deck message is only a warning." Still true, and it is
not the guarantee.** The guarantee is the dispatch handing out the tree, which
is what row 1 now describes. The deck line covers the case nothing in this
repository can fix: a person, or an outside harness, starting two decks in one
tree. There the working directory was chosen at launch, and the issue's own
`#5043` report is that a session which detects the clash has no move but to
stop — so refusing would leave it with nowhere to go rather than protecting it.
ADR 0027 states that boundary in its consequences rather than leaving it
implied.

## Remaining work filed

- #5887 — a plan can declare `claims` on an isolated task, where they
  coordinate nothing anybody shares, and the run never says so.
- #5888 — `crates/stella-cli/src/fleet_cmd.rs` sits exactly on its 1500-line
  ceiling, so the next change there fails the gate before it starts.

@sourcery-ai review

macanderson:
Checked all three Sourcery ❌ rows against the diff. **All three rest on one wrong premise — that the default did not change. It did.** Answering rather than pushing a fix, per CLAUDE.md's "Sourcery's verdict is a claim like any other review comment and can be wrong about your diff".

### Row 1 — "The PR does not change the dispatch implementation … shared-tree dispatch … is still the fleet default"

`crates/stella-fleet/src/plan.rs` is modified `+53/−36`, and the default flips in it:

| | merge-base | PR head |
|---|---|---|
| module doc | `//! Tasks share the tree by default ([Isolation::SharedTree]…` | `//! Every task gets its own git worktree ([Isolation::Isolated], the…` |
| field doc | `/// …defaults to [Isolation::SharedTree] when` | `/// …defaults to [Isolation::Isolated] when absent` |
| `Task::new` | `isolation: Isolation::SharedTree,` | *(no longer sets SharedTree)* |
| test | `assert_eq!(plan.tasks[0].isolation, Isolation::SharedTree)` | `assert_eq!(plan.tasks[0].isolation, Isolation::Isolated)` |

Sourcery is right that **shared-tree remains supported** — it does, as an opt-in via `Isolation::SharedTree`, which ADR 0027 keeps deliberately. It is wrong that it is still the *default*. "Supported" and "default" are the two different things this row collapses.

### Row 2 — "the new test does not demonstrate a failing pre-change behavior"

The PR head adds:

```rust
assert_eq!(Task::new("a", "t", "p").isolation, Isolation::Isolated);
```

At the merge base `Task::new` sets `isolation: Isolation::SharedTree`, so this assertion **fails before the change and passes after** — the fail→pass witness CONTRIBUTING.md asks for, on the exact behaviour in question. It is not witnessing pre-existing behaviour.

### Row 3 — "only a warning … does not prevent the branch-switch data-loss scenario"

Correct that `command_deck/shared_checkout.rs` never refuses. That is deliberate, and the module header gives the reason:

> *"Stella's own dispatchers do not make that state. A fleet task gets a `git worktree` of its own unless the plan named the shared tree (ADR 0027), and so does every unit of `stella self-driving work`. A person, or an outside harness, can still start two decks in one tree. This is what that person sees. It never refuses. A session told to stop has nowhere to go…"*

The two paths are doing different jobs. Dispatch — the scenario #5287 is about — is fixed structurally by the default flip. The notice covers only the residual case Stella does not create: a human starting two decks in one tree by hand.

Refusing *there* would strand the session, and #5287 says why in its own report:

> *"My session is hard-pinned to one checkout — the Bash sandbox refuses any command whose cwd or `git -C` resolves elsewhere … so a dedicated worktree I created was unusable and I removed it."*

The issue also ranks its own options, calling warn-on-entry **"Cheapest, and weakest — a session that reads the warning still has no way to act on it."** The PR does not substitute the weak option for the strong one; it takes the strong one for dispatch and applies the weak one only where the strong one cannot reach.

### Against #5287's actual definition of done

- [x] *Two sessions dispatched at the same moment cannot revert each other's uncommitted work* — default is now a worktree per task.
- [x] *A hermetic self-test … dispatches two sessions against one repository and asserts both trees survive* — `scripts/test-session-isolation.sh` (+296), wired into `ci.yml` and the `Makefile`.
- [x] *That test fails before the change* — see Row 2.

All three met. ADR 0027 records the decision, as SCR-002 requires.

**Recommendation: no code change for these rows.** If a reviewer wants the residual manual-deck case turned into a refusal, that is a separate decision from #5287 and needs its own issue, because it depends on the cwd-pinning constraint the issue itself documents.

Unrelated to the above: this PR's `main is not known-broken` will stay red until #5906 lands and closes #5901.

---

macanderson:
Follow-up push `995cbc3`: two comments still called the shared tree the default after `#[default]` moved.

- `crates/stella-fleet/Cargo.toml` — the note above `description` said "Share-by-default is the crate's actual contract"; the `description` beside it had already been updated in this branch.
- `crates/stella-cli/src/fleet_commits.rs` — the header on `for_attempt` opened "The **shared tree** — the default — has no such luxury".

Comments only, no code. `check-prose.py` and `make guards-fast` pass locally; CI is the compiler.

Sourcery's three ❌ rows are answered in the comment above (head `5ae27e6`), and the rows's shared premise — that the default did not move — is what these two comments were the last traces of.

macanderson:
## Two open PRs claim ADR 0027

Found by the open-PR sweep. This PR and #5921 each add a *different* ADR numbered 0027, and each registers it in the same two index files:

| PR | New file |
|---|---|
| **#5823** (this one) | `docs/adr/0027-a-fleet-worker-gets-its-own-worktree.md` |
| **#5921** | `docs/adr/0027-panel-cells-are-glyphs-in-the-contract.md` |

Both also insert a `0027` row into `docs/adr/README.md` immediately after the `0026` row, and both add an `adr/0027-…` key to `docs/manifest.json` immediately after `adr/0026-…`.

Verified against `origin/main` (`c498653`) rather than inferred from either description: `docs/adr/` there holds `0001`–`0026` and nothing else, so 0027 is genuinely free — and genuinely claimed twice.

### Why this is not caught pre-merge

Neither branch is wrong on its own, so both can go green: `adr-numbering` sees exactly one 0027 in each tree. `docs/adr/README.md` and `docs/manifest.json` are the shared cells AGENTS.md's `main-canary` section describes — the same shape as `Cargo.lock` and `scripts/file-size-baseline.txt`. Whichever of these two merges second lands a tree holding two ADRs numbered 0027, and `make adr-numbering` then fails on `main` for everyone.

### What is owed

**Whichever PR merges second renumbers to 0028**, in one edit:

1. `git mv` the ADR file to `0028-…`
2. update its frontmatter `id:` and its `# ADR 00NN:` heading
3. repoint the `docs/adr/README.md` table row and its prose paragraph
4. repoint the `docs/manifest.json` key and `path`
5. repoint every `doc:adr/0027-…` citation — #5921 has these in `crates/stella-plugin/src/panel.rs` and `crates/stella-tui/src/plugin_panel.rs`; this PR cites its ADR from `crates/stella-fleet/README.md` and its own prose

Neither PR is renumbered here, because which one takes 0028 is decided by merge order and neither has merged. The same note is on #5921.

---

macanderson:
## ADR 0027 is claimed by two open PRs

Flagging a cross-PR collision rather than acting on it — which ADR keeps the number is yours to pick, and neither branch is wrong on its own.

This PR adds `docs/adr/0027-a-fleet-worker-gets-its-own-worktree.md` and registers 0027 in `docs/adr/README.md` and `docs/manifest.json`. **#5921** adds `docs/adr/0027-panel-cells-are-glyphs-in-the-contract.md` and registers 0027 in those same two files — verified from its diff at `4d5c8ba`, not from its description:

```
docs/adr/0027-panel-cells-are-glyphs-in-the-contract.md   | +68  (added)
docs/adr/README.md                                        | +6
docs/manifest.json                                        | +5
```

Both are open against `main`. Those two index files are one shared cell that both branches write, which is the composition hazard AGENTS.md names for `scripts/file-size-baseline.txt`: no pre-merge run can catch it, because neither tree is wrong. The second to land either conflicts in the indexes or ships a duplicate ADR number past `adr-numbering`.

Whichever lands second wants renumbering to 0028 — the file, its `id` frontmatter, the README row, the manifest key, and the `doc:adr/…` citations in the prose that references it (this PR's ADR is cited from `fleet_cmd/isolation_notice.rs`'s neighbourhood; #5921's is cited from `crates/stella-plugin/src/panel.rs` and `crates/stella-tui/src/plugin_panel.rs`).

I have not pushed to either branch. The same note is on #5921.


---

Issue #5824, opened 2026-09-05
stella-store trusts a tool block's own stored bytes instead of checking its digest


macanderson:
## The problem

`crates/stella-store/src/reconstruct.rs`'s `content_matches_digest` decides
whether a rebuilt context block's bytes really are that block's bytes. It opens
with:

```rust
if block.content.is_some() {
    return true;
}
```

So any block carrying local content passes, whatever its kind. Local content is
legitimate only for the gap kinds — the list is `stella_core::receipts`'s
`is_gap_kind`: `system_prefix`, `user_goal`, `recalled_frame`, `steered`,
`summary`, `attachment`. A `tool_result` or `tool_call` row with content in it
contradicts the producer, and there is no reason to trust bytes that came from
the very row being checked.

The result is a `tool_result` row whose stored bytes disagree with its
`content_digest` reading as verified. That is the plausible-looking lie the
digest exists to prevent, and a tampered `store.db` is the threat the feature's
own doc names.

`stella-observatory`'s copy of this fold was fixed in #5500
(`crates/stella-observatory/src/sent_context.rs`): the skip is keyed on the
block's kind, and local content on any other kind is named in a new
`local_content_anomalies` list that withholds the overall `verified`. The two
folds now disagree, so `stella inspect` can report `verified` for an execution
the dashboard reports an anomaly on.

## Where

- `crates/stella-store/src/reconstruct.rs` — `content_matches_digest`, and
  `Reconstruction`'s fields (whether this crate wants an anomaly list of its
  own, or reports it some other way).
- `crates/stella-observatory/src/sent_context.rs` — the shape already landed;
  `GAP_KINDS` there is the list, and `reconstruct`'s `is_gap` branch is the
  rule.
- `crates/stella-cli/src/inspect.rs` — the CLI surface that renders the
  verdict.

## How to verify

Add a witness to `stella-store` shaped like the observatory's
`local_bytes_on_a_journal_kind_are_checked_rather_than_trusted`: record a
`tool_result` `ContextBlockRow` whose `content` is set to bytes that do not hash
to its `content_digest`, reconstruct the call, and assert the reconstruction
does not report it as verified. It fails on today's tree because
`content_matches_digest` returns `true` on the `content.is_some()` arm.

`cargo test -p stella-store reconstruct` is the loop. Note the workspace rule:
CI compiles, this laptop does not — a targeted single-crate run is the
exception.

## Constraints

[...]

Issue #101, opened 2026-07-16
Theme is truecolor-only — degrade gracefully on 256-color terminals


macanderson:
## Problem

`stella-tui/src/theme.rs` defines every token as `Color::Rgb` (18 uses — the amber `#FFAC26` accent, ember ramp, diff tints, status colors). Terminals without truecolor support (`COLORTERM` unset — plenty of ssh/tmux/legacy setups) render approximations chosen by the terminal, which can be badly off (amber → brown/grey), and some render nothing legible for low-contrast pairs.

Deferred from the PR #92 pass ("truecolor-only palette").

## Fix

Detect truecolor support once (`COLORTERM=truecolor|24bit`, then `$TERM` heuristics) and, when absent, map each token to a hand-picked xterm-256 indexed fallback (e.g. accent → 214). Keep the mapping in theme.rs next to the RGB definitions so the two palettes can't drift apart silently; a unit test asserts every token has a fallback.

PR #5825, opened 2026-09-05
fix(stella-observatory): stop five dashboard numbers picking the reassuring state

### LINKED ISSUE #5500: Five dashboard numbers are wrong, each by picking the more reassuring of two states
## The problem

Five numbers and flags the Observatory shows an operator are wrong, each for its
own reason. They are filed together because they share one shape: a projection
that cannot distinguish two states and picks the more reassuring one.

## 1. Self-driving spend double-counts across overlapping sessions

`crates/stella-observatory/src/self_driving_sessions.rs`, `session_json` and
`sessions_from`. Per-session spend is a time-window filter over the whole loop's
unpartitioned cost vector (`execution_costs`) — nothing links a cost row to the
session that incurred it — and `totals.spend_usd` then re-sums every session's
already-inflated figure.

Two sessions running at once is a supported and seeded case; the file's own
fixture comments "Interleaved: two drives against one repo at once", and
`dispatch_claims` exists to support it. With $10 spent in the overlap window,
each session's figure sums to ~$10 and the total reports **$20**. The `daily`
chart on the same page, computed once rather than per session, reports the
correct $10 — so the hero tile and the chart visibly disagree.

**Fix.** `let spend: f64 = days.values().map(|d| d.spend_usd).sum();` instead of
re-summing per-session figures.
**Witness.** One $10 execution, two overlapping stamped sessions spanning it;
assert `totals.spend_usd == 10.0`.

## 2. An episode tagged to one issue is counted against another

Same file, `issue_json`. The explicit tag and the time-window fallback are OR'd:

```rust
e.get("issue").and_then(Value::as_str) == Some(i.number.as_str())
    || in_window(e.get("started_unix")...)
```

An episode carrying `"issue": "101"` still matches issue `#100` whenever its
timestamp falls inside `#100`'s claim window. A session that claims `#100` then
`#101` while `#100`'s window is still open counts the `#101` episode under both.

**Fix.** Let the tag win when present: `match e.get("issue") { Some(tag) => tag ==
i.number, None => in_window(...) }`.

## 3. The queue's `in_progress` ignores the live claims it says it reads

Same file. The comment says "Overlay the live claims on the queue so 'in the
queue' and 'being worked right now' are one picture", but `open_now` is built
entirely from journal-derived rows and `claims` — read from `fleet.db`'s
`dispatch_claims`, described in the module doc as "the machine-wide answer to
what is being worked this second" — only populates `holders`.

A driver that crashes without a `session_stopped` record leaves its lease to
expire normally, which is the true signal that nobody holds the issue. The
journal's last stage still reads open, so the dashboard reports `in_progress:
true` indefinitely and tells an operator someone is working an issue that is
free.

**Fix.** Intersect the parsed `issue:<n>` claim keys into the `in_progress`
computation.

## 4. `/api/session` reports an ungraded reflection as graded

`crates/stella-observatory/src/sessions.rs`, `fold_turn_extras`.
`crates/stella-observatory/src/db.rs` defines `HAS_SELF_REVIEW` precisely because
`execution_reflection` gets a row for every finalized turn whether or not the
model graded it, and both of `db.rs`'s own readers apply it. `fold_turn_extras`
is a third reader of the same table and does not, inserting a non-null
`{"delivered": null, "self_rating": null}` whenever any row exists.

So `/api/execution?id=N` returns `reflection: null` for a turn and
`/api/session?id=S` returns an object for the same turn. The shipped page gates
its visible use on `self_rating != null`, so nothing wrong renders today; the
payload contract is broken for the next consumer.

**Fix.** Reuse `db::HAS_SELF_REVIEW` in that query.

## 5. Digest verification is skipped for any block carrying local content

`crates/stella-observatory/src/sent_context.rs`, `reconstruct`. The skip is
gated on `entry.content.is_some()` alone. Local content is legitimate only for
the two gap kinds (`system_prefix` and the assembled message), an invariant the
event producer upholds and this reader never checks by `kind`.

A `tool_result` row carrying `content` whose bytes contradict its
`content_digest` — via a producer bug, or a tampered `store.db`, which is the
threat this feature's own doc names — reports `digest_verified: null`, stays out
of `mismatches`, and the call's overall `verified` stays `true`. That is the
plausible-looking lie the feature exists to prevent.

**Fix.** Gate the skip on the kind, and treat local content on any other kind as
a reportable anomaly.

## Definition of done

- [x] Each of the five has a witness that fails before and passes after.
      Verified by ablation on #5825 at `a8d1a4c` — each fix reverted in turn with
      its witness kept, then the named witness re-run alone:

      | # | witness | result with the fix reverted |
      |---|---|---|
      | 1 | `overlapping_sessions_do_not_bill_one_dollar_twice` | FAILED — `left: Number(20.0)`, `right: 10.0` |
      | 2 | `an_episode_tagged_to_one_issue_is_not_counted_against_another` | FAILED |
      | 3 | `a_working_issue_with_no_live_lease_is_not_in_progress` | FAILED — `left: Bool(true)`, `right: false` |
      | 4 | `a_turn_the_model_never_graded_carries_no_reflection` | FAILED — `left: Object {"delivered": Null, "self_rating": Null}`, `right: Null` |
      | 5 | `local_bytes_on_a_journal_kind_are_checked_rather_than_trusted` | FAILED |

      Item 1's ablation reproduces the $20-vs-$10 disagreement this issue opens
      with, and item 4's reproduces the all-null verdict object, so each witness
      fails on the defect described here rather than on something incidental.
      Both non-regression pins — `an_issue_parked_on_a_pull_request_stays_in_progress_without_a_lease`
      and `a_gap_kind_keeps_its_null_verdict_and_raises_no_anomaly` — pass on
      both sides of every ablation.
- [x] `cargo test -p stella-observatory` green — 205 tests, `0 failed`, exit 0,
      on `a8d1a4c` with a pristine worktree (`git status --porcelain` empty
      after every ablation was reverted). 163 lib unit tests plus the
      integration binaries; one `ignored`, none failing.

## Confidence

All five CONFIRMED by the audit, each read directly against the cited function
and, where a doc comment claims the opposite behaviour, quoted above. Items 1–3
change numbers an operator acts on; items 4 and 5 are contract breaks that do not
render wrongly today.

## Also noted, not filed separately

Two client-side hardening gaps with no exploit path today, both in
`crates/stella-observatory/src/assets/self_driving.js`: five `xd.day`
interpolations into SVG markup skip the `esc()` every other field uses (and
`day_of` never validates the separator bytes it slices around), and a queue
item's `url` is HTML-escaped but never scheme-checked before use as an `href`.
Both values come from harness-stamped data today.

macanderson:
## What and why

Five projections in `stella-observatory` could not tell two states apart, and
each reported the kinder one. All five are fixed here, each with a witness.

**1 — self-driving spend double-counted across overlapping sessions.**
`sessions_from` re-summed every session's `spend_usd`. A session's figure is a
time-window sum over `execution_costs`, and nothing links a cost row to the
session that spent it, so two drives running at once each count the whole
overlap. The tile read $20 while the `daily` chart under it, computed once,
read $10. The total now comes from the daily series. The per-session figure
stays a window sum, and its comment now says so.

**2 — an episode tagged to one issue was counted against another.**
`issue_json` OR'd the explicit `issue` tag with the time-window fallback, so an
episode carrying `Issue #101` matched issue 100 whenever 100's claim window was
still open. The tag wins where it exists; the window is the fallback for an
episode that names none.

**3 — the queue's `in_progress` ignored the live claims its comment said it
overlaid.** Fixed, but not by the plain intersection the issue proposes — the
code says that would be wrong, so here is why.

`crates/stella-cli/src/self_driving_cmd/drive.rs`'s `LoopStep::Work` arm takes
the `issue:<n>` lease and drops it on the way out, in its own words "the window
where a turn is in flight and no pull request exists yet". An issue parked on a
pull request holds no lease for as long as CI takes, so a bare intersection
would flip every parked issue to "not in progress" — which is the loop's normal
steady state. So the lease may contradict the journal only inside the window
the lease covers (`in_progress`), and a parked issue is left to the journal.

The second half of that fix is `live_claims` returning `Option<Vec<Value>>`.
It collapsed "the table says nobody holds anything" into "there is no table to
ask" — the same two-state confusion this issue is about, one layer down — and
with the intersection added, a workspace that never ran the fleet would have
reported every issue free. `collect_rows` also degrades a missing table to an
empty result, so the table is now asked for by name before the query.

**4 — `/api/session` reported an ungraded reflection as graded.**
`fold_turn_extras` was the third reader of `execution_reflection` and the only
one not applying `db::HAS_SELF_REVIEW`, so it inserted
`{"delivered": null, "self_rating": null}` for any row at all. It now reuses
that predicate.

**5 — digest verification was skipped for any block carrying local content.**
The skip is keyed on the block's kind now. Two notes on scope:

- The issue says "the two gap kinds", quoting this module's own header. The
  producer's list (`stella_core::receipts`'s `is_gap_kind`) names six:
  `system_prefix`, `user_goal`, `recalled_frame`, `steered`, `summary`,
  `attachment`. `GAP_KINDS` follows the producer, and the stale header is
  fixed in the same change. `is_gap_kind` is private, so this is an
  acknowledged copy like the rest of the fold.
- "A reportable anomaly" is `local_content_anomalies` in the payload. It
  withholds `verified` on its own, because a block resolved from the very row
  it is checked against was never checked by anything. The page renders it —
  without that, a call unverified for this reason alone drew no verdict line at
  all, which is the silence the fix exists to end.

## Witness mechanism

Each witness fails on `main` by construction, not by luck:

| Witness | Fails before because |
|---|---|
| `overlapping_sessions_do_not_bill_one_dollar_twice` | one $10 execution, two overlapping sessions; the old sum is 20.0, the assert is 10.0 |
| `an_episode_tagged_to_one_issue_is_not_counted_against_another` | the tagged episode falls inside the other issue's window, so the OR counts 2 there; the assert is 1 |
| `a_working_issue_with_no_live_lease_is_not_in_progress` | `in_progress` read no lease at all before, so the expired-lease issue asserted `false` was `true` |
| `an_issue_parked_on_a_pull_request_stays_in_progress_without_a_lease` | passes before and after — it pins the non-regression the plain intersection would have caused |
| `a_turn_the_model_never_graded_carries_no_reflection` | the finalize row exists, so the old query returned an object where `null` is asserted |
| `local_bytes_on_a_journal_kind_are_checked_rather_than_trusted` | local content skipped the digest, so the block was absent from `digest_mismatches` and `verified` was `true` |
| `a_gap_kind_keeps_its_null_verdict_and_raises_no_anomaly` | passes before and after — it pins that the skip still applies where it belongs |

## Exemplar

No new shape: `live_claims`'s `Option` return is `std`'s own distinction
between "absent" and "empty", the one `HashMap::get` makes; `GAP_KINDS` follows
the producer's `is_gap_kind` rather than inventing a second list.

## File layout

`self_driving_sessions.rs` was 1501 lines with the three new witnesses' worth of
room needed on top. Its `#[cfg(test)] mod tests` moves verbatim to
`self_driving_sessions/tests.rs` (the `foo.rs` beside `foo/` layout the
workspace uses), leaving the parent at 1267. The fixtures that spelled an issue
number in a string literal are now built by `pr_opened_for` / `episode_naming`,
because the new file gets no prose baseline and those strings trip the
`issue-reference` guard.

## Tests deleted

None. Every test in `self_driving_sessions.rs` survives under the same name in
`self_driving_sessions/tests.rs`.

## Residue filed

- #5824 — `stella_store::reconstruct`'s `content_matches_digest` has the same
  shape as item 5 and is unfixed: it returns `true` for any block with local
  content, whatever its kind, so `stella inspect` can report `verified` where
  this crate now reports an anomaly.

Closes #5500

---

The witness table above was checked rather than taken on trust, and #5500's two
DoD boxes are ticked against the result. Each fix was reverted in turn with its
witness kept, and the named witness re-run alone: all five **FAILED**, items 1
and 4 reproducing the exact `20.0`-vs-`10.0` and all-null-verdict symptoms
#5500 describes; both non-regression pins passed on both sides of every
ablation. `cargo test -p stella-observatory` on `a8d1a4c` is **205 tests, 0
failed, exit 0**, worktree clean afterwards. Detail and command output are in
[this comment](https://github.com/macanderson/stella/pull/5825#issuecomment-5548274401).

re-read the boxes — the edit is the trigger, never the fix.

## Summary by Sourcery

Correct observatory projections so dashboard totals, issue activity, queue status, reflections, and context verification distinguish trustworthy states from merely reassuring ones.

Bug Fixes:
- Prevent overlapping self-driving sessions from double-counting workspace spend totals.
- Ensure issue tags take precedence over time-window matching when attributing episodes.
- Align queue in-progress status with live leases while preserving journal-based status for pull requests and unavailable lease tables.
- Return no reflection for turns that were finalized without a self-review.
- Verify digests based on block kind rather than the presence of local content, and surface anomalous local content explicitly.

Enhancements:
- Distinguish unavailable lease data from an empty lease table when deriving observatory state.
- Expose local-content anomalies in context verdicts and update gap-kind coverage to match the producer.

Tests:
- Add regression coverage for spend attribution, issue episode matching, lease-aware queue status, reflection grading, and digest verification.

Chores:
- Move self-driving session tests into a dedicated tests module while retaining the existing test coverage.

### DIFF
--- a/crates/stella-observatory/src/assets/index.html
+++ b/crates/stella-observatory/src/assets/index.html
@@ -2535,8 +2535,11 @@ function sentContextVerdictHtml(c) {
   }
   const bad = (c.digest_mismatches ?? []).length;
   const gaps = (c.unresolved ?? []).length;
+  const local = (c.local_content_anomalies ?? []).length;
   return `${bad ? `<div class="ctx-verdict" style="color:var(--bad)">✕ ${fmtInt(bad)} block(s)
       did NOT re-hash to their recorded digest — the journal is torn or was altered</div>` : ""}
+    ${local ? `<div class="ctx-verdict" style="color:var(--bad)">✕ ${fmtInt(local)} block(s) were read
+      from the store instead of the journal, so nothing independent checked them</div>` : ""}
     ${gaps ? `<div class="ctx-verdict" style="color:var(--warn)">◌ ${fmtInt(gaps)} block(s) could not
       be resolved (synthetic results, discarded speculation, or attachments)</div>` : ""}`;
 }
--- a/crates/stella-observatory/src/self_driving_sessions.rs
+++ b/crates/stella-observatory/src/self_driving_sessions.rs
@@ -784,11 +784,29 @@ fn execution_costs(root: &Path) -> Vec<(i64, f64)> {
 /// expires on its own if the holder dies (#4300). So this table — not any one
 /// journal — is the machine-wide answer to "what is being worked this
 /// second, and by whom", including a second drive against the same clone.
-fn live_claims(root: &Path, now: i64) -> Vec<Value> {
+///
+/// `None` is a third answer, not a spelling of the empty list. An empty table
+/// says nobody holds an issue right now. An unreadable one says nothing at
+/// all. A workspace that never ran the fleet or the drive verb has no
+/// `fleet.db`. Reading that as "every issue is free" would let this page
+/// contradict a journal it has no evidence against.
+fn live_claims(root: &Path, now: i64) -> Option<Vec<Value>> {
     let db = root.join(".stella").join("private").join("fleet.db");
-    let Some(conn) = open_read_only(&db) else {
-        return Vec::new();
-    };
+    let conn = open_read_only(&db)?;
+    // `collect_rows` turns a missing table into an empty result. That is the
+    // collapse this signature undoes, so ask for the table by name first. A
+    // `fleet.db` older than the lease table cannot answer either.
+    let claims_table: i64 = conn
+        .query_row(
+            "SELECT count(*) FROM sqlite_master
+             WHERE type = 'table' AND name = 'dispatch_claims'",
+            [],
+            |r| r.get(0),
+        )
+        .ok()?;
+    if claims_table == 0 {
+        return None;
+    }
     let now_ms = now * 1000;
     let rows = collect_rows(
         &conn,
@@ -809,22 +827,55 @@ fn live_claims(root: &Path, now: i64) -> Vec<Value> {
             }))
         },
     );
-    match rows {
-        Ok(v) => v,
-        Err(crate::db::DbError::Query(e)) if is_missing_schema(&e) => Vec::new(),
-        Err(_) => Vec::new(),
+    rows.ok()
+}
+
+/// Whether one queued issue is being worked this second.
+///
+/// The journal proposes and the lease table disposes. The lease covers a
+/// narrow window. `stella self-driving drive` takes `issue:<n>` when it starts
+/// a turn and drops it on the way out of that arm. `drive.rs`'s
+/// `LoopStep::Work` calls that arm "the window where a turn is in flight and
+/// no pull request exists yet".
+///
+/// So the lease may contradict the journal, but only inside that window. An
+/// issue parked on a pull request holds no lease, because the loop is polling
+/// CI. There the journal is the only witness.
+///
+/// Everywhere else a turn should be in flight. A lease table that names no
+/// holder is saying nobody is running one. A driver that died without a stop
+/// record leaves exactly that state, and its journal keeps the issue open
+/// until it goes stale.
+///
+/// `leased: None` means there was no lease table to ask. It contradicts
+/// nothing.
+fn in_progress(
+    number: &str,
+    open_now: &BTreeMap<String, String>,
+    leased: Option<&BTreeSet<String>>,
+) -> bool {
+    let Some(stage) = open_now.get(number) else {
+        return false;
+    };
+    match leased {
+        Some(held) if stage != "pr open" => held.contains(number),
+        _ => true,
     }
 }
 
 fn issue_json(i: &Issue, learning: &Learning) -> Value {
     let (lo, hi) = (i.claimed_unix, i.last_unix.max(i.claimed_unix));
     let in_window = |u: i64| u >= lo && u <= hi + 60;
+    // An episode that names an issue belongs to that issue alone. The time
+    // window is the fallback for an episode that names none. OR'd with the
+    // tag, it pulled one issue's episode into another as well, whenever the
+    // two windows overlapped.
     let episodes: Vec<&Value> = learning
         .episodes
         .iter()
-        .filter(|e| {
-            e.get("issue").and_then(Value::as_str) == Some(i.number.as_str())
-                || in_window(e.get("started_unix").and_then(Value::as_i64).unwrap_or(-1))
+        .filter(|e| match e.get("issue").and_then(Value::as_str) {
+            Some(tag) => tag == i.number,
+            None => in_window(e.get("started_unix").and_then(Value::as_i64).unwrap_or(-1)),
         })
         .collect();
     let lessons: Vec<&Value> = learning
@@ -879,6 +930,11 @@ fn session_json(
     } else {
         now
     };
+    // What the workspace spent while this session was open. It is a window
+    // sum, not a split. No cost row names the session that spent it. So a
+    // dollar spent while two sessions overlap is counted by both. That is as
+    // much as one row can claim. It is also why the machine-wide total in
+    // `sessions_from` adds up the daily series instead of these figures.
     let spend: f64 = costs
         .iter()
         .filter(|(u, _)| *u >= s.started_unix && *u <= end)
@@ -981,15 +1037,21 @@ fn sessions_from(roots: &[PathBuf], workspace_root: &Path, now: i64) -> Value {
         });
         let costs = root.as_deref().map(execution_costs).unwrap_or_default();
         let spend_days = root.as_deref().map(spend_rows).unwrap_or_default();
-        let claims = root
-            .as_deref()
-            .map(|r| live_claims(r, now))
-            .unwrap_or_default();
-        for c in &claims {
+        let claims = root.as_deref().and_then(|r| live_claims(r, now));
+        for c in claims.iter().flatten() {
             if let Some(o) = c.get("owner").and_then(Value::as_str) {
                 holders.insert(o.to_string());
             }
         }
+        // The issues a live lease names, or `None` where there was no lease
+        // table to ask.
+        let leased: Option<BTreeSet<String>> = claims.as_ref().map(|rows| {
+            rows.iter()
+                .filter_map(|c| c.get("key").and_then(Value::as_str))
+                .filter_map(|k| k.strip_prefix("issue:"))
+                .map(str::to_string)
+                .collect()
+        });
 
         // Journal → daily.
         for s in &l.sessions {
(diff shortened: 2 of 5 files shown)


macanderson:
## `dod` was red on #5500's two unticked boxes; both are now ticked against a measurement, not a claim

Automated PR sweep. `dod / dod` failed twice on `a8d1a4c` ([00:56:13](https://github.com/macanderson/stella/actions/runs/33934578398/job/101219892519), [00:56:21](https://github.com/macanderson/stella/actions/runs/33934588422/job/101219920251)) naming #5500's two unchecked DoD items. Neither is a defect in this diff, and no push would have changed the verdict — so I verified both myself rather than ticking them on the description's say-so.

**DoD 2 — `cargo test -p stella-observatory` green.** Run on a clean checkout of `a8d1a4c`: **205 tests, 0 failed, exit 0** (163 lib unit tests plus the integration binaries; one `ignored`, none failing).

**DoD 1 — each of the five has a witness that fails before and passes after.** Checked by ablation rather than by reading the table in the description: for each item I reverted *only that fix*, kept its witness, and re-ran the named test alone.

| # | fix reverted to | witness | result |
|---|---|---|---|
| 1 | re-summing per-session `spend_usd` | `overlapping_sessions_do_not_bill_one_dollar_twice` | **FAILED** — `left: Number(20.0)`, `right: 10.0` |
| 2 | the `tag == n \|\| in_window(…)` OR | `an_episode_tagged_to_one_issue_is_not_counted_against_another` | **FAILED** |
| 3 | `open_now.contains_key(&n)`, ignoring the lease | `a_working_issue_with_no_live_lease_is_not_in_progress` | **FAILED** — `left: Bool(true)`, `right: false` |
| 4 | dropping `db::HAS_SELF_REVIEW` from the predicate | `a_turn_the_model_never_graded_carries_no_reflection` | **FAILED** — `left: Object {"delivered": Null, "self_rating": Null}`, `right: Null` |
| 5 | `is_gap = true`, skipping the digest on any local content | `local_bytes_on_a_journal_kind_are_checked_rather_than_trusted` | **FAILED** |

Item 1's ablation reproduces the $20-against-$10 disagreement #5500 opens with, and item 4's reproduces the all-null verdict object — so each witness fails on the defect the issue describes rather than on something incidental to the ablation.

**The two non-regression pins earn their place.** `an_issue_parked_on_a_pull_request_stays_in_progress_without_a_lease` and `a_gap_kind_keeps_its_null_verdict_and_raises_no_anomaly` pass on *both* sides of every ablation, which is what the description claims and what makes item 3's narrower fix (lease may contradict the journal only inside `in_progress`) reviewable rather than asserted.

One note on method: item 3's naive ablation (`open_now.contains(&n)`) does not compile, because `open_now` is a `BTreeMap` in this diff where it was a set before. `contains_key` is the honest revert and is what the row above reports. Worth saying because a non-compiling ablation is not a failing test, and reporting it as one would be the exact confusion #5500 is about.

The worktree was restored and re-verified clean after every ablation (`git status --porcelain` empty, full suite green again).

#5500's boxes are ticked with this evidence recorded in the issue body. `dod` re-reads the issue live, so it needs a `pull_request` event to publish a new verdict — a comment does not fire one. Nothing else on this PR is outstanding: no review threads, Sourcery posted no assessment table (its review was `skipped`), and every other check is green or still running.

---

Issue #5826, opened 2026-09-05
Write a verdict stamp when a plugin's evidence becomes part of a run's record


macanderson:
## In plain terms

We added a record type, `VerdictStamp`, so the system can say "who checked this, and what did they find." But no code writes one yet — every stamp seen today comes from a test, not a real run. This issue makes the host write a real stamp every time it turns a plugin's evidence into a pass/fail decision.

## Problem

`VerdictStamp` and `LadderSnapshot::stamps` already exist in `crates/stella-protocol/src/ladder/stamp.rs`. Nothing in the codebase writes one. The field is empty on every real run, so the "who checked this" record the type was built for is still missing at the one place it matters.

`stella-protocol` cannot build the stamp by itself: the hash it needs comes from `record_hash` (ADR 0004), and `stella-protocol` is not allowed to depend on `sha2` or `serde_json_canonicalizer`. `LadderSnapshot::stamp_preimage()` already exists in `stella-protocol` and returns the exact object that should be hashed.

## Where to make the change

- `crates/stella-runtime/src/wrapper/verdict.rs` and `crates/stella-runtime/src/wrapper/dispatch.rs` — this is where the host turns a plugin's `EvidenceSet` into a decision (`Outcome`). Write the stamp here.
- `crates/stella-plugin/src/wire.rs` holds `EvidenceProvenance`, which decides the stamp's `author` field: `PluginReported` means the author is the plugin's own manifest id; `HostObserved` means the author is `"engine"`. Always read the id from the loaded manifest — never a name the plugin sends itself. This is what keeps the field trustworthy.
- `record_hash` lives in `stella_protocol::hash` since #5829 moved the record plane out of `stella-core`. Call `record_hash(&snapshot.stamp_preimage()?)` to get the stamp's hash.

## Rules already decided — do not change these

- A stamp can never raise a rung on its own. The test `stamps_never_promote_the_rung` in `crates/stella-protocol/src/ladder.rs` checks this; a new stamp writer must not break it.
- The hash input drops the `stamps` list itself, so adding a second stamp must not change the hash of an earlier one. The test `a_new_stamp_does_not_change_what_earlier_stamps_were_made_against` checks this.
- No cryptographic signing here. Proving data hasn't changed (integrity) and proving who sent it (identity) are different problems — this issue only solves the first one (#5120's stated non-goal).
- What happens when two stamps disagree is a separate issue, #5121. This issue only writes the record.

[...]

Issue #5121, opened 2026-08-26
plugin: arbitration law at the completion gate — multiple arbiters, and instrument failure as a recorded abstention rather than silent fail-open or costly fail-closed


macanderson:
Depends on the VerdictStamp issue (stamps are the record this fold reads and
writes). Settles two questions #3245 left open and pushes back on both
answers currently on the table.

## Question 1: what happens when the verdict instrument breaks?

Two contradictory answers exist in the project's own documents:

- **#3245 (shipped posture): fail-open.** "A verdict hook that fails, times
  out, or emits garbage never blocks." Inherited from the Stop hook.
- **`doc:plugin-architecture` §7.4: fail-closed.** "An errored evaluation is
  an unconditional deny — the turn stays open, the conservative direction."

**Pushback on both.** Fail-closed is wrong here: for *tool* gating, deny is
conservative (nothing runs); for *completion* gating, deny means the model
keeps burning steps against feedback it cannot act on — "the plugin crashed"
is not an addressable instruction, and a dead participant would eat the
entire hold budget accomplishing nothing. But fail-open as shipped is wrong
too, because it is **silent**: the turn completes and the trace is
indistinguishable from "verification passed". A run whose arbiter was dead
looks exactly like a run whose arbiter was satisfied — the precise labelling
failure goal:traces exists to prevent (compare #2273: unstamped
`verifier_independent` made every real flip report "unmeasured").

**Proposed law: fail-open, loudly.** An arbiter that crashes, times out, or
emits garbage never blocks — AND the host records a
`VerdictStamp { assessment: Inconclusive, timed_out/summary: ... }` on the
snapshot, emits the existing extension-failure event, and the deck attributes
it ("arbiter X did not answer"). Absence of verification becomes a visible
property of the verdict rather than a forensic discovery.

## Question 2: two arbiters, one turn

#3245 open question 2 recommended "one arbiter; a second install prompts to
choose" — fine as install UX, but composition arrives anyway (a workspace
witness-plugin plus an org-mandated policy plugin is the obvious pair). The
fold must be law before it is emergent behavior. Spec `fold_stamps` as a pure
function (property-testable, no I/O — the `ladder_decision` shape):

1. **A deterministic failure is never overridden** — no combination of
   `Done` stamps changes a deterministic-fail rung (ROADMAP's standing rule).
2. **Any single live `NotDone` holds the turn** — vetoes are
   unanimous-to-pass, never majority; a verification instrument that needs a
[...]

PR #5827, opened 2026-09-05
feat(stella-protocol): VerdictStamp — verdict evidence carries authorship

### LINKED ISSUE #5120: protocol: VerdictStamp — verdict evidence carries authorship, so a second observer can agree, disagree, or abstain on the record
From `doc:plugin-architecture` §7.2, the one piece of its verdict protocol
with no shipped counterpart. Verified today: `LadderSnapshot` still lives in
`stella-protocol/src/ladder.rs` (re-exported unchanged per #1043) and carries
*what* was observed — `oracle_trace`, `flip`, `verify_done_flip`,
tamper/coverage fields — but not *who* observed it, and there is no slot for
a second observer to disagree. `rg VerdictStamp` across the repo: zero hits.
Witness-protocol P1 (separation of authorship) names this; D2 ("flip credited
on command identity alone") is still open by decision.

## Why this matters more now than when the doc was written

The wrapper socket made third-party verification real: `stella-plugin`'s
`EvidenceSet`/`FlipObservation`/`Verdict` let an external wrapper report
evidence, and the host evaluates `VerdictRule`. But when the host folds that
into the trace, the trajectory label loses the provenance chain — a
fine-tuning pipeline (every trace perfectly labelled, goal:traces) cannot
distinguish "the engine's oracle observed the flip" from "wrapper X reported
it and the host believed it". Stamps are the join between the two systems.

## Spec

Add to `stella-protocol/src/ladder.rs`:

```rust
pub struct VerdictStamp {
    /// Host-filled from the loaded manifest — a plugin cannot impersonate
    /// another. `"engine"` for the engine's own decision.
    pub author: String,
    pub author_version: Option<String>,   // informational, verbatim
    pub assessment: StampAssessment,      // Done | NotDone | Inconclusive | NotApplicable
    pub summary: String,                  // one line, prose, never parsed
    /// sha256 over the RFC 8785 JCS bytes of the LadderSnapshot with
    /// `stamps` removed from the preimage — the record_hash primitive
    /// (ADR 0004), so one hashing discipline covers the tree, and replay can
    /// prove the claim was made against the evidence the run produced.
    pub preimage_hash: String,
    pub evidence_refs: Vec<String>,       // same vocabulary as VerdictEvidence
    pub decided_at_ms: u64,
    pub duration_ms: u64,
    #[serde(default)] pub timed_out: bool,
}
```

and on `LadderSnapshot`:

```rust
#[serde(default, skip_serializing_if = "Vec::is_empty")]
pub stamps: Vec<VerdictStamp>,
```

Key properties, in order of load-bearing-ness:

1. **Wire-additive**: zero stamps serializes byte-identically to today —
   golden fixtures do not move. Witness test required.
2. **`Inconclusive` ≠ `NotDone`** — same reason `LadderRung` splits
   `Unverifiable` from `Unverified`: an instrument that cannot measure is
   not a finding about the work. This distinction is what the arbitration
   issue (filed separately) builds on.
3. **Stamps never promote a rung.** Three `Done` stamps on an `Unverified`
   rung leave it `Unverified` — information for a human, never for
   automation. This is the witness-protocol §5 line (no
   attestation-as-warranty) as a type-level fact.
4. **Reconcile with `EvidenceProvenance` before writing code.**
   `stella-plugin::wire` already exports an `EvidenceProvenance` type —
   audit it first. If it already carries per-observer authorship, this issue
   shrinks to "bridge it onto the snapshot"; if it marks host-vs-plugin
   origin only, the stamp remains the missing piece and provenance feeds
   `author`. Do not ship two authorship vocabularies.

Consumer-ledger rows (`event/consumers.rs`) and serde round-trip tests per
architecture invariants; wire schema re-export gate (`check-wire-schema.sh`)
picks up the new types.

## Explicit non-goals

Cryptographic signing (integrity ≠ identity; separate threat model), and any
change to how verdicts are *decided* — this issue records observers, it does
not give them a vote. The vote is the arbitration issue.

---
**Suggested worker:** `model-tier:pro` · **Estimated autonomous time:** ~90 min

## Definition of done

- [x] `VerdictStamp` exists as the spec section describes, so verdict evidence carries authorship.
- [x] A second observer can agree, disagree, or abstain on the record.
- [x] The explicit non-goals stay non-goals — vote-counting is the arbitration issue, not this one.

macanderson:
## What and why

`LadderSnapshot` recorded *what* was observed — the oracle trace, the flip, the tamper and coverage fields — and nothing about *who* observed it. That was survivable while one host ran the check. It stopped being survivable once the wrapper socket made third-party verification real: when the host folds a plugin's `EvidenceSet` into the trace, "the engine's oracle saw the flip" and "wrapper X reported it and the host believed it" are byte-identical, and a fine-tuning pipeline reading the label cannot tell them apart. A second observer also had no slot to disagree in.

`VerdictStamp` is that slot: one claim per observer, carrying the author, an assessment, one line of prose, the hash of the evidence the claim was made against, and how long it took. `LadderSnapshot::stamps` holds them in arrival order.

**The audit the issue asks for, first.** `stella_plugin::wire::EvidenceProvenance` has two arms, `PluginReported` and `HostObserved`. It answers "host or plugin" and never "which plugin", so the stamp stays the missing piece rather than shrinking to a bridge, and provenance feeds `author` wherever a stamp is later minted. No second authorship vocabulary is added.

**A stamp is a record, never a vote.** Nothing reads one to decide anything. Three `Done` stamps on an `Unverified` snapshot leave it `Unverified` — counting agreement is the move that let an ablated verifier score every run a pass with nothing noticing (#2569/#2570). Arbitration is a separate issue, filed below.

**`preimage_hash` and where the digest lives.** `LadderSnapshot::stamp_preimage()` returns the exact object the hash is computed over: this snapshot with `stamps` removed. Producers share it, so two of them cannot disagree about the bytes and mint hashes that read as tampering, and so a later claim never invalidates an earlier one's hash. The digest itself stays the record-hash primitive (ADR 0004) in `stella-core`, which owns `sha2` and `serde_json_canonicalizer`; `stella-protocol`'s README boundary forbids taking either here.

**Exemplar.** The neighbourhood, not an outside crate: `FlipOutcome` and `LadderRung` are the shape copied — a closed `snake_case` vocabulary with an `as_str` that spells what serde writes, and a builder (`with_rung`, `with_verifier_independence`) that touches one field.

## Witness mechanism

Every new test names `VerdictStamp`, a type that does not exist on `main`, so the whole set fails to compile there by construction — the feature is genuinely absent. Beyond that, each test pins one property that could regress after the type exists:

- `stamps_are_additive` — an unstamped snapshot emits no `stamps` key, and a snapshot recorded before stamps existed parses with an empty list. This is the wire-additive half: no stored verdict moves.
- `stamps_never_promote_the_rung` — three `Done` stamps leave the rung at `Unverified`, and clearing `stamps` gives back a value equal to the unstamped snapshot, so stamping changes exactly one field.
- `a_new_stamp_does_not_change_what_earlier_stamps_were_made_against` — the preimage is identical with and without stamps, carries no `stamps` key, and equals the plain serialization when nothing is stamped.
- `the_stamp_round_trips` (invariant 4), `the_optional_fields_are_absent_rather_than_null`, `the_tokens_match_what_serde_writes`, `an_abstention_does_not_serialise_as_a_finding` — `inconclusive` and `not_done` must not collapse, for the reason `Unverifiable` and `Unverified` do not.
- The wire-contract corpus gains a sweep over all four assessments, so each arm reaches the wire and validates against the committed schema. `StampAssessment` is registered in both tables in `wire_contract.rs`, so a fifth arm added without a sample fails the test.

## Other changes in the diff

`crates/stella-protocol/tests/wire_contract/samples.rs` sat at 1461 of its 1500-line ceiling, so the sweep could not land in it (AGENTS.md § "God files — plan around them, never into them"). The ladder vocabularies, the snapshot base and the three verdict sweeps move to `samples/verdict.rs`, and the sweeps now share one `sampled_ladder()` instead of restating every field three times. Sampled values are unchanged; `samples.rs` is back to 1364 lines.

`docs/wire/` is regenerated (`bash scripts/export-agentevent-schema.sh`): the schema gains the optional `stamps` property, a `VerdictStamp` definition and a `StampAssessment` one. `check-wire-schema.sh` is green, including its no-Rust-paths half.

No new `AgentEvent` variant, so the consumers ledger (invariant 10) is unchanged. No new dependency.

## Non-goals, kept

No signing — integrity and identity are different threat models. No change to how verdicts are decided: this records observers, it does not give them a vote.

Tests deleted: None.

## Remaining work filed

- #5121 (already open) — arbitration: what a host does when two stamps disagree. Linked rather than duplicated.
- #5826 (filed here) — the producer path: mint a stamp where the host folds a plugin's evidence into a snapshot. Nothing writes one today, and the issue says so out loud.

Closes #5120

---

#5120's three DoD boxes are ticked, against the diff at `1455513` rather than against this description. The stamp's fields were checked one by one against the issue's spec block; the `EvidenceProvenance` audit (spec property 4) was re-derived from `crates/stella-plugin/src/wire.rs:930` rather than taken on trust — two arms, `PluginReported` and `HostObserved`, answering origin and never identity, which is the issue's second branch verbatim. `git grep -c VerdictStamp origin/main -- crates/` returns zero hits, so the witness-by-construction claim holds. `cargo test -p stella-protocol --all-features` is **266 lib tests, 0 failed, exit 0**, with all seven named witnesses green. Detail and command output are in [this comment](https://github.com/macanderson/stella/pull/5827#issuecomment-5548307513).

This edit is what fires the `pull_request` event `dod-check.yml` needs to re-read the boxes — the edit is the trigger, never the fix.

## Summary by Sourcery

Record observer authorship and evidence-linked claims on ladder snapshots without changing verdict arbitration or existing wire data.

New Features:
- Add authorship-aware verdict stamps to ladder snapshots, recording each observer's assessment, summary, evidence references, timing, and evidence preimage hash.
- Expose four wire-stable stamp assessment values while preserving distinctions between negative findings, abstentions, and non-applicable checks.

Bug Fixes:
- Preserve backward-compatible serialization for unstamped and legacy ladder snapshots without altering existing verdicts.

Enhancements:
- Provide a shared snapshot preimage for claims so adding later stamps does not change the evidence each earlier claim references.
- Keep stamps informational only, ensuring observer claims do not promote or otherwise change the ladder rung.

Documentation:
- Regenerate the wire schemas and TypeScript definitions to include ladder stamps, verdict stamps, and stamp assessments.

Tests:
- Extend wire-contract coverage across all stamp assessments and refactor verdict samples into a dedicated module.


macanderson:
## DoD verification for #5120 — all three boxes ticked

The `dod / dod` check was red on 8 unchecked boxes across the open PRs; this is the verification for this one. Each item was checked against the diff at `1455513` rather than against the PR description, and the boxes on #5120 are now ticked.

### 1. `VerdictStamp` exists as the spec section describes

Field-by-field against the issue's spec block — `crates/stella-protocol/src/ladder/stamp.rs`:

| Spec field | Shipped | Match |
|---|---|---|
| `author: String` | same | ✅ |
| `author_version: Option<String>` | same, `skip_serializing_if = "Option::is_none"` | ✅ |
| `assessment: StampAssessment` | `Done \| NotDone \| Inconclusive \| NotApplicable` | ✅ |
| `summary: String` | same | ✅ |
| `preimage_hash: String` | same, `sha256:<64 hex>` | ✅ |
| `evidence_refs: Vec<String>` | same | ✅ |
| `decided_at_ms` / `duration_ms: u64` | same | ✅ |
| `#[serde(default)] timed_out: bool` | same | ✅ |
| `LadderSnapshot::stamps` | `#[serde(default, skip_serializing_if = "Vec::is_empty")]` | ✅ |

Spec property 1 (wire-additive) is pinned by `stamps_are_additive`: an unstamped snapshot emits no `stamps` key, and a pre-stamps JSON body parses with an empty list. Property 2 is pinned by `an_abstention_does_not_serialise_as_a_finding`. Property 3 by `stamps_never_promote_the_rung` — which also asserts that clearing `stamps` gives back a value `==` the unstamped snapshot, so stamping is proven to touch exactly one field rather than merely claimed to.

**Spec property 4 (the `EvidenceProvenance` audit) I re-derived rather than took from the PR body.** `crates/stella-plugin/src/wire.rs:930` declares exactly two arms — `PluginReported` and `HostObserved` — and its own doc says it travels with the evidence and that `judge` never reads it. It answers *host or plugin*, never *which plugin*. That is the issue's second branch verbatim, so the stamp is the missing piece rather than a bridge, and no second authorship vocabulary ships: `EvidenceProvenance` stays in `stella-plugin` answering origin, `VerdictStamp` lands in `stella-protocol` answering identity.

Nothing new landed in `event/consumers.rs`, which is correct — invariant 10's ledger is keyed on `AgentEvent` wire tags and this diff adds no variant.

### 2. A second observer can agree, disagree, or abstain on the record

`LadderSnapshot::with_stamp` appends rather than replaces, and the wire corpus proves all three postures actually reach the wire, not just the type. `samples/verdict.rs`'s `sample_stamps` pairs an `engine` / `Inconclusive` stamp with a `vera` stamp swept across every `StampAssessment`, so the committed schema is validated against **agreement** (`Done`), **disagreement** (`NotDone`) and **abstention** (`Inconclusive`) on one record. `every_nested_vocabulary_is_fully_sampled` registers `StampAssessment` in both tables, so a fifth arm added later without a sample fails the test.

The non-overwrite half is `a_new_stamp_does_not_change_what_earlier_stamps_were_made_against`: `stamp_preimage()` is byte-equal with and without stamps, so the second observer's claim cannot invalidate the first one's hash — which is what makes disagreement recordable rather than destructive.

### 3. The explicit non-goals stay non-goals

No signing: the only two `signature`-matching lines in the touched files are prose, one of them the doc line stating a stamp carries none. No change to how verdicts are decided — `with_stamp` is the sole mutator and `stamps_never_promote_the_rung` proves it. Arbitration is linked to #5121 rather than started here.

### Witness-by-construction, checked

`git grep -c VerdictStamp origin/main -- crates/` returns **zero hits**, so every new test names a type that does not exist on `main` and the set fails to compile there. That is the PR's claim and it holds.

### Test run

`cargo test -p stella-protocol --all-features` on `1455513`, in a dedicated worktree:

```
test result: ok. 266 passed; 0 failed; 0 ignored   (lib)
test result: ok. 9 passed; 0 failed; 0 ignored     (wire_contract)
exit 0
```

All seven named witnesses present and green:

```
ladder::stamp::tests::the_stamp_round_trips ... ok
ladder::stamp::tests::the_optional_fields_are_absent_rather_than_null ... ok
ladder::stamp::tests::the_tokens_match_what_serde_writes ... ok
ladder::stamp::tests::an_abstention_does_not_serialise_as_a_finding ... ok
ladder::tests::stamps_are_additive ... ok
ladder::tests::stamps_never_promote_the_rung ... ok
ladder::tests::a_new_stamp_does_not_change_what_earlier_stamps_were_made_against ... ok
```

Single-crate run per SCR-001 — the workspace build is CI's job, and `fmt + clippy + test` is the authority on this branch.

---

Issue #5822, opened 2026-09-05
The candidate fan-out plane still owns a list of plugin role words


macanderson:
## Problem

`crates/stella-runtime/src/wrapper/candidate_fanout.rs` still holds
`default_seats()` — a core-owned table of four plugin role words
(`worker`, `triage`, `research`, `plan`) mapped to `ModelCallRole`. It is the
last one in `stella-runtime` after #3905 deleted the child-turn plane's copy.

The rule the fan-out plane enforces is the inverse of the child plane's: a
candidate **is** the work, so it must resolve to the worker's seat, and a tier
that resolves elsewhere is refused. The rule is right. Deciding it by looking
up one of four words core happens to know is what #3903 is retiring: a plugin
that names its candidate tier `builder` or `attempt` is refused, and no
plugin-chosen word can ever be served.

## The fix

Do for `CandidateFanouts` what #3905 did for `ChildTurns`.

- Delete `candidate_fanout::default_seats()`.
- Read the answer off the grant instead. `crates/stella-runtime/src/wrapper/seat_grant.rs`
  holds `SeatGrant`, which today answers "may this plugin spend at this seat?"
  for a read-only child turn. A fan-out needs the other question — "is this
  plugin granted a writing turn?" — and the manifest fact that answers it is
  `[loop] calls` containing `candidate_fanout`, which a human consented to at
  install.
- Keep `CandidateFanouts::with_seat` so a host can still bind a tier and own
  the claim, and keep checking the resolved seat rather than the spelling.
- Keep the refusal codes as they are. A tier a host bound somewhere other than
  the worker's seat must stay refused, with the same code and a message that
  still tells the plugin author what to do.

## Verify

- `rg -n 'triage|research|"plan"|worker' crates/stella-runtime/src/wrapper/candidate_fanout.rs`
  finds no role-word string literal outside `mod tests`.
- A witness: a manifest whose candidate tier is a word core has never heard of
  runs a fan-out. It is refused `Unavailable` today.
- `crates/stella-runtime/tests/wrapper_candidate_fanout.rs` still asserts every
  candidate is booked at `ModelCallRole::Worker`.
- The doc comment on `crates/stella-cli/src/wrapper_plugin/planes.rs`'s
  `candidate_fanout_plane` describes the grant and drops the pointer to this
  issue.

## Constraints

- `crates/stella-runtime/src` compiles with no toolchain on the maintainer's
  machine: push and read CI.
[...]

Issue #3838, opened 2026-08-19
plugin: no ChildTurns binding anywhere serves the `verifier` role intent, so a goal-supervision arbiter plugin's after_turn always degrades to Unavailable


macanderson:
## Problem

`plugins/stella-goal` (the goal-supervision reference plugin, `doc:pipeline-as-plugins` §3 / `doc:turn-loop-wrappers` §9.2) declares `[roles.verifier] tier = "verifier"` and spends its one `child_turn` call at that role intent in `after_turn`. `ChildTurns::default_seats()` (`crates/stella-runtime/src/wrapper/child_turn.rs:169-179`) deliberately does not serve `verifier` — the module's own doc comment (lines 154-168) names the reason: attributing a plugin's child turn to `ModelCallRole::Verdict` would put a call on the receipt the pipeline itself did not make.

That is a reasonable default, but **nothing anywhere binds it**. `stella run`'s own wrapper-plugin driver (`crates/stella-cli/src/wrapper_plugin.rs:430`) calls exactly:

```rust
ChildTurns::declare(manifest, dispatcher).with_turn_instance(PLUGIN_CHILD_TURN_SLOT)
```

— no `.with_seat("verifier", ...)`. So on `stella run --pipeline goal-v1` **as shipped today**, every `after_turn` call this plugin makes resolves to `HostCallRefusal::Unavailable`, the plugin reports empty evidence (no `"met"` measurement), `judge` abstains (`UndecidedReason::MeasurementMissing`), and the loop ends `Undecided` after exactly one round, every single run. The plugin cannot do its one job on any host running today.

This is distinct from #3833 (`stella goal --pipeline <variant>`'s own `WrapperHost` serves `recall` only, no `child_turn` plane at all): here the `child_turn` plane exists and is wired for `stella run`'s door specifically — it just never resolves the one role intent a goal-mode verifier plugin needs.

## Repro

1. `cargo test -p stella-runtime --test goal_plugin_dispatch -- without_a_bound_verifier_seat_the_loop_ends_undecided_after_one_round` — passes today, and pins exactly this behavior (a `ChildTurns` plane built the same way `stella run`'s door builds one, with no `.with_seat` call).
2. Contrast with `a_round_the_verifier_marks_unmet_holds_open_for_one_correction_round` in the same file, which only succeeds because the test binds `.with_seat("verifier", ModelCallRole::Verdict)` by hand — something no shipped code path does.

## Fix

[...]

PR #5828, opened 2026-09-05
feat(stella-runtime): decide a plugin's child-turn seat from its grant

### LINKED ISSUE #3905: roleless-core slice 1: stop the child-turn plane from owning the list of plugin role words
Slice 1 of #3903. Spec: `doc:roleless-core` §6.

## Problem

`ChildTurns::default_seats()` in `crates/stella-runtime/src/wrapper/child_turn.rs`
is a **core-owned table of plugin role words**: it serves `worker`, `triage`,
`research` and `plan`, mapping each to a `ModelCallRole`. A plugin whose process
needs a `reviewer`, a `second-opinion`, or anything else core did not anticipate
is answered `HostCallRefusal::Unavailable` and cannot run a child turn at all.

That caps the set of processes a plugin can express at the set core anticipated,
which is the exact coupling the wrapper socket exists to remove (architecture rule 2 and
3 of `doc:roleless-core`).

Slice 0 (landed, commit `dca951ecf`) already added the routing half:
`SubAgentSpec::seat` carries the plugin's own word, and
`crates/stella-cli/src/agent/seats.rs` resolves it against the user's
assignments. The word list is now the only thing standing between a plugin and
an arbitrary role name.

## What to do

- Delete `default_seats()`. A plugin's declared role name resolves through the
  seat plane and nothing else.
- Move the rule `default_seats()` was actually carrying — a plugin may not
  silently buy the worker's seat, and may not silently buy a verifier tier — off
  the *name* and onto the **grant**: an explicit check against the manifest a
  human consented to, expressed without reference to any role word.
- `crates/stella-cli/src/wrapper_plugin.rs`'s `child_turn_plane` doc comment
  currently argues for the refusal; rewrite it to describe the grant instead.

## Constraints

- `ChildTurns::resolve` returning `Forbidden` vs `Unavailable` is a documented
  distinction that sends a plugin author to different places — preserve it.
- The independence property (#2584: no model verdict) must survive the move. It
  is currently enforced by *absence of a name*, which is why it is fragile;
  after this slice it must be enforced by a grant check.

## Witness

A manifest declaring a role named `reviewer` runs a child turn. Today it is
refused `Unavailable`.

## Done when

No literal role word appears anywhere in `crates/stella-runtime/`.

## Definition of done

The done check above was written as a whole-crate one and cannot be met as
written: `candidate_fanout.rs` keeps its own table for the inverted rule (a
candidate **must** resolve to the worker's seat), which is #5822, and prose
explaining the rule still names the worker. The boxes restate the same
conditions in a form that can be checked.

- [x] `ChildTurns::default_seats()` is deleted, and
      `crates/stella-runtime/src/wrapper/child_turn.rs` holds no role-word
      string literal outside its tests.
- [x] A plugin's declared role name resolves through the seat plane and
      nothing else: the plane compares it, forwards it on `SubAgentSpec::seat`,
      and branches on nothing.
- [x] The two rules the table carried are checked against the consented
      manifest, with no role word in the check.
- [x] `Forbidden` and `Unavailable` keep their meanings and stay reachable.
- [x] A witness fails on the old code: a manifest declaring a tier core has
      never heard of runs a child turn.
- [x] A witness fails on the old code: a host binding a deciding seat is
      refused when the manifest declares no `[oracle]`.
- [x] `wrapper_plugin.rs`'s `child_turn_plane` doc comment describes the grant,
      and the door binds no tier by name.
- [x] The sibling table in `candidate_fanout.rs` is filed as its own issue.

macanderson:
## What and why

`ChildTurns::default_seats()` was a core-owned table of four plugin role words
(`worker`, `triage`, `research`, `plan`). A plugin whose process needed a
`reviewer` was answered `HostCallRefusal::Unavailable` and could not run a child
turn at all, so the set of processes a plugin could express was capped at the
set core anticipated — the coupling the wrapper socket exists to remove
(`doc:roleless-core` §6, slice 1 of epic #3903).

The table is deleted. A role intent is a word the plugin chose: the plane
compares it, forwards it on `SubAgentSpec::seat` for the user's seat map to
route on, and branches on nothing.

The two rules the table was carrying move onto the grant, in a new
`crates/stella-runtime/src/wrapper/seat_grant.rs`:

- `SeatGrant::permits(seat)` refuses the seat the session's own turns are booked
  at outright (`Forbidden`) — a plugin may not spend the model whose work it is
  judging;
- it refuses a deciding seat (`WitnessAuthor`, `WitnessRepair`, `Verdict`)
  unless the manifest declares an `[oracle]` (`Unavailable`). A manifest may
  only carry one at arbiter grade, so the block is the plugin stating that job
  in the document a person reads at install.
- `SeatGrant::default_seat()` books an oracle-declaring plugin at `Verdict` and
  every other plugin at the read-only child-call seat.

Both refusals keep their codes and their meanings, and both are decided on the
**resolved** seat, so `ChildTurns::with_seat` — how a host binds a tier of its
own — cannot hand over a seat the grant refuses. The match over `ModelCallRole`
names every case with no wildcard, so a seat added later has to be classified
rather than granted by default.

`stella run`'s door drops its `verifier -> Verdict` binding (#3838). The grant
reaches the same answer for `plugins/stella-goal`, because that manifest
declares an oracle — and now for the same reason a person consented to, rather
than because this host happened to know one word.

### Deviation from the design pointer

The pointer asked for a free `permits_seat(&manifest, seat)`. It is
`SeatGrant::permits(seat)` instead, with `SeatGrant::of(&manifest)` read once at
`declare`. A plane is built from a borrowed manifest and outlives it, so a free
function would mean cloning a whole manifest into every plane to re-ask one
question — a second copy of the answer. Everything else follows the pointer.

### The cost, stated

`plugins/stella-plan`'s planner child turn is booked at the read-only
child-call seat instead of `plan`, because core no longer holds a word that
tells the two apart. Its own name already travels on `SubAgentSpec::seat`;
slice 2 (#3906) is where a receipt carries it. Every other shipped attribution
is unchanged.

## Witness mechanism

Each of these fails on `main` by construction, because `main` looks the tier up
in `default_seats()` before anything else runs:

- `a_tier_core_has_never_heard_of_runs_at_the_grants_own_seat`
  (`child_turn.rs`) — a manifest declaring `[roles.reviewer] tier = "reviewer"`
  runs a turn. On `main` the tier is in no table and the ask is refused
  `Unavailable`, so the `expect` fails.
- `a_deciding_seat_needs_a_manifest_that_declared_the_job` (`child_turn.rs`) —
  a host binding `verifier -> WitnessAuthor` against a manifest with no
  `[oracle]` is refused `Unavailable`. On `main` the binding alone runs the
  turn, so the `expect_err` fails.
- `a_tier_spelled_like_the_workers_never_reaches_the_workers_seat`
  (`stella-cli`'s `wrapper_plugin::tests`) — through the shipped door,
  `tier = "worker"` books at the read-only child-call seat and never at the
  session's own. On `main` the same call is refused `Forbidden`.
- `SeatGrant`'s own unit tests pin the rule from both sides, including that the
  default seat is always one the grant permits.

Preserved witnesses: `a_role_intent_that_resolves_to_the_worker_is_forbidden`
and `renaming_the_worker_does_not_buy_a_plugin_the_worker` still prove the
independence rule, now against a host that actually binds that seat — which is
the only way to reach it once core knows no role words.

## Exemplar

`stella_plugin::LoopGrant`'s `permits_call` / `permits_hook` shape: one grant
type read off the consented manifest, answering one question, consulted by the
host at the point of use.

## Tests deleted

Three renamed, none dropped:

- `an_unserved_tier_names_the_tiers_that_are_served` ->
  `a_deciding_seat_needs_a_manifest_that_declared_the_job`. Its old subject —
  "the tier binds to no seat" — no longer exists, and the new one covers the
  same refusal code for the reason that replaced it.
- `a_role_intent_pointing_at_the_worker_is_still_forbidden` ->
  `a_tier_spelled_like_the_workers_never_reaches_the_workers_seat`, in
  `stella-cli`. Same subject, opposite outcome: the shipped door binds no such
  seat, so the word buys nothing rather than being refused.
- `without_a_bound_verifier_seat_the_loop_ends_undecided_after_one_round` ->
  `without_a_child_turn_plane_the_loop_ends_undecided_after_one_round`. The
  premise it named is gone; the degradation it pins is unchanged, and a door
  with no dispatcher of its own is still in exactly that state.

## Residue filed

- #5822 — `candidate_fanout.rs` still holds the same shape of table for the
  inverted rule (a candidate **must** resolve to the worker's seat). It is the
  last one in `stella-runtime` and needs its own grant question, so it is a
  separate change rather than a widening of this one.

## Scope note on the issue's done check

"No literal role word appears anywhere in `crates/stella-runtime/`" cannot be
met as written: the fan-out table above is one reason, and prose that explains
the independence rule names the worker as a concept. The issue's `## Definition
of done` now states the checkable form — the child-turn plane carries no
role-word table and no role-word literal outside its tests — and says why.

## Density fix, and the verification sweep (appended 2026-09-05)

`9dbc9c3` cuts two lines from `seat_grant.rs`'s header. `crates/stella-runtime`
composed over its 35.64 mean-header ceiling — 35.67 — once this file landed on
top of `main`'s own header growth, which reddened `gate steps no other workflow
runs`. Neither side is wrong alone: `scripts/prose-density-baseline.txt` is the
shared cell AGENTS.md describes, and the branch passes the base-relative check
while CI's `fetch-depth: 1` checkout takes the strict whole-tree path.
`./scripts/check-prose.py --absolute` reproduces the failure before the edit and
passes after; `cargo fmt --check` clean. The lines cut are the retrospective
half of "Why the rule sits on the grant" — what the old word-list table got
wrong, which this description already carries. The rule and the
`doc:roleless-core` §6 citation stay. That check is now green on this head.

#3905's eight DoD boxes are ticked, verified against the diff rather than
against this description: `default_seats()` is gone from `child_turn.rs` with no
role-word literal outside its tests; `resolve` looks the intent up in the
manifest's declared roles and branches only on `grant.permits(seat)`;
`SeatGrant::of` reads exactly `manifest.oracle.is_some()` and `permits` matches
`ModelCallRole` with no wildcard; both refusal codes stay reachable with their
meanings; both named witnesses are absent from `origin/main -- crates/` and so
cannot compile there; `planes.rs` binds no tier by name and `git grep '"verifier"\|with_seat('`
across the door returns nothing; and #5822 is open carrying `triage` alone.
Detail is in [this comment](https://github.com/macanderson/stella/issues/3905#issuecomment-5548359175).

This edit is what fires the `pull_request` event `dod-check.yml` needs to re-read
the boxes — the edit is the trigger, never the fix.

Closes #3905

## Summary by Sourcery

Replace core-owned child-turn role tables with manifest-derived seat grants that preserve independence safeguards while allowing plugins to use arbitrary role intent names.

New Features:
- Allow plugin-defined role intent words to route through the user's seat map without requiring core-owned role tables.
- Add manifest-derived seat grants that select a default attribution and enforce independence and oracle-consent rules.

Bug Fixes:
- Allow oracle-declaring plugins such as stella-goal to run child turns without host-specific verifier bindings.
- Prevent host seat bindings from bypassing grant restrictions on worker and deciding seats.

Enhancements:
- Remove the child-turn default seat table and make seat permission decisions based on the installed plugin manifest.
- Update plugin behavior, runtime integration, tests, and user-facing specifications for grant-based seat attribution.

Documentation:
- Document grant-based child-turn seat attribution and the remaining candidate fan-out table as a separate concern.

Tests:
- Add coverage for unknown plugin-defined tiers, manifest-required deciding seats, grant permissions, and updated goal and plan dispatch behavior.


macanderson:
## DoD verification for #3905 — all eight boxes ticked, and a note on the `check-prose` red

Automated PR sweep. Two checks were red on `a6243e5`: `dod / dod` on #3905's eight unchecked boxes, and `gate steps no other workflow runs`. Both are settled below. Every item was checked against the diff and against `origin/main`'s own code, never against this description.

### The eight boxes

**1 — `default_seats()` deleted, no role-word literal outside tests.** `git grep default_seats` on the branch's `child_turn.rs`: zero hits (it was at line 170 on `main`). Every role-word string literal in that file is at line 591 or later; `#[cfg(test)] mod tests` opens at line 517. So the shipped half of the file names no role word.

**2 — the plane compares, forwards, and branches on nothing.** `resolve` is a two-hop lookup with no role-word branch: `roles.get(role)` → `seats.get(tier)` → `grant.default_seat()`, then `grant.permits(seat)`. The plugin's own word is forwarded verbatim as `seat: Some(args.role.clone())` on `SubAgentSpec`, beside the enum `role: seat` used only as the receipt label.

**3 — both rules checked against the consented manifest, with no role word in the check.** `seat_grant.rs` contains **no role-word string literal at all**, in shipped code or tests — the check is a `match` over `ModelCallRole` variants plus `manifest.oracle.is_some()`. `SeatGrant::of` reads the grant once at `declare`.

**4 — `Forbidden` and `Unavailable` keep their meanings and stay reachable.** `SeatPermission::Never → Forbidden` ("you may not have this") and `Undeclared → Unavailable` ("nobody consented to it"), both constructed in `resolve` and both covered from each side by `the_seat_the_sessions_own_turns_use_is_never_granted` and `a_verdict_seat_needs_a_declared_oracle`. `every_other_seat_is_granted` sweeps `ModelCallRole::ALL` minus the four refused, so a seat added later cannot be granted by an unnoticed wildcard.

**5 and 6 — the two fail-on-main witnesses.** I re-derived both mechanisms from `main`'s `resolve` rather than accepting the description's "by construction":

- `a_tier_core_has_never_heard_of_runs_at_the_grants_own_seat` — on `main`, `self.seats` is `default_seats()`, which serves only `worker`/`triage`/`research`/`plan`, so a manifest declaring `[roles.reviewer] tier = "reviewer"` misses the `seats.get(tier)` lookup and returns `Unavailable`. The test expects the turn to run, so it fails there.
- `a_deciding_seat_needs_a_manifest_that_declared_the_job` — on `main` the only post-lookup check is `if seat == ModelCallRole::Worker`. A host binding `verifier → WitnessAuthor` therefore reaches `Ok(seat)` and the turn runs; the test's `expect_err` fails there.

**7 — the doc comment describes the grant, and the door binds no tier by name.** `main`'s `child_turn_plane` ended `.with_seat("verifier", ModelCallRole::Verdict)`; the branch's body is `ChildTurns::declare(manifest, dispatcher).in_turn_lane(…)` with no binding, and `with_seat` no longer appears anywhere in shipped `stella-cli` code. The rewritten comment states the grant rule and cites the witness.

**8 — the sibling table is filed.** #5822, opened 2026-09-05, `triage` label only (SCR-005), and it is specifically `candidate_fanout.rs`'s inverted rule.

### Test runs

Targeted, per SCR-001. On `9dbc9c3`:

```
cargo test -p stella-runtime --lib wrapper::      41 passed; 0 failed
cargo test -p stella-cli --bin stella wrapper_plugin   50 passed; 0 failed
```

Both named witnesses green, plus the preserved independence pair and `a_tier_spelled_like_the_workers_never_reaches_the_workers_seat` on the CLI side, and `the_shipped_goal_plugins_verifier_intent_resolves_on_this_hosts_plane` — which is the description's claim that `stella-goal` still reaches `Verdict` through its oracle rather than through the dropped binding.

One unrelated failure, stated rather than glossed: `panel_host::tests::a_panels_grandchild_does_not_outlive_its_tick` fails in my container. It is **not this PR's** — the diff does not touch `panel_host.rs`, the test fails identically on `origin/main` in the same container (the cheap control), and `main`'s own `ci` run at `6d5fb8fc` is green. A process-lifetime test under a container PID namespace; nothing to fix here.

### The `check-prose` red, and who fixed it

`gate steps no other workflow runs` failed on `check-prose`: `crates/stella-runtime` at 35.67 mean header lines against a 35.64 ceiling, on `seat_grant.rs`'s 29-line header.

Worth recording because it reads green locally and red in CI: the density ratchet's effective ceiling is `max(baseline, density at the base commit)`, and `docs-guards` checks out at `fetch-depth: 1`, so CI resolves no base and the bare 35.64 stands. A local run with the branch in reach resolves the branch head as its own base and compares the number against itself. `python3 scripts/check-prose.py --absolute` is the local command that reproduces CI, and it is how I confirmed the fix.

`9dbc9c3` landed while I was verifying my own version of the same cut, so I discarded mine rather than push over it — the duplicate-repair hazard AGENTS.md documents. Verified it is genuinely fixed: 54 files, 1924 header lines, mean **35.63**, and both `check-prose` and `check-prose --absolute` exit 0.

One observation, not a request: that leaves **one line** of slack for the whole crate. The next header added anywhere in `stella-runtime` re-reds this check, which is the razor-thin-gate shape. Reclaiming slack is `make prose-retighten`'s job in a PR of its own, not this one's.

---

Issue #4997, opened 2026-08-26
The rebase-replay guard reports but does not block, because it is not a required check


macanderson:
## Problem

PR #4995 added `.github/workflows/rebase-replay.yml`, which runs
`scripts/check-rebase-replay.sh` on every pull request and fails when a path is
edited and then undone inside the branch — the shape #4979 describes, where a
rebase turns an inert pair into a live revert of somebody else's change.

It is **not** in `main`'s required checks, so a red run does not stop a merge.
That is a repository setting, not a file in this tree, which is why the PR could
not do it — the same wall #2425 hit for `deck-fit.yml` and #3932 is still open
about for `main is not known-broken`.

The failure it exists for is silent by construction: the branch's diff for the
path is empty either way, so a reviewer looking at the diff sees nothing, and
the revert only becomes visible on `main` after the merge. A reporting-only
check is therefore worth much less here than for a guard whose finding a human
would notice anyway.

## What to do

Add **`no edit is undone inside the branch`** (the job name in
`.github/workflows/rebase-replay.yml`) to `main`'s required status checks, in
the repository's branch-protection settings.

Before doing so, confirm the guard is not noisy on live traffic. It was measured
once, at the time #4995 landed: over the heads of the last 40 merged pull
requests, 39 green and 1 flagged (#4939, verified a true positive — net diff
empty for `docs/tools/bash.toml`, one commit editing it and a later one undoing
it). Re-run that sweep before flipping the switch:

```sh
for n in $(gh pr list --repo macanderson/stella --state merged --limit 40 \
             --json number --jq '.[].number'); do
  git fetch -q origin "refs/pull/$n/head:refs/heads/sweep" --force
  ./scripts/check-rebase-replay.sh origin/main sweep >/dev/null 2>&1 \
    || echo "flagged #$n"
  git branch -q -D sweep
done
```

## Constraints

- The remedy the guard prints is to flatten the pair, which is always available
  and changes nothing about the branch's tree — so making it blocking cannot
  wedge a pull request that has no other way out. There is deliberately no
  acknowledgement path (`scripts/check-rebase-replay.sh`'s header says why).
- `main-red-hold.yml` is in the same position and tracked separately (#3932);
  whoever edits branch protection for one may as well settle both.

## Done when

A pull request whose branch edits and undoes a path cannot be merged while that
is true, and the setting is recorded wherever this repository records its
branch-protection state.

[...]

PR #5840, opened 2026-09-05
fix(stella-records): report a trigger that names a path the other record guards

### LINKED ISSUE #5780: Conflict detection never crosses a path scope against a task scope, so those collisions stay silent

macanderson:
## What and why

Record conflict detection compared scopes one dimension at a time — path globs against path globs, then task/keyword triggers against triggers (#5330). A record scoped `paths = ["deny.toml"]` and a record scoped `keywords = ["deny.toml"]` never met, so a `hard` guard and an advisory record over the same file resolved silently. That is the §12 governance failure `records/validate.rs` exists to report.

## The decision, and where it is written

Cross-dimension pairs **are** reported, under one rule: **a trigger term that is itself a path the other record's globs match.**

Selection is disjunctive, so a record scoped `paths = ["crates/**"]` and a record scoped `tasks = ["refactor"]` both fire on a turn that edits a file under `crates/` and says "refactor". A turn like that can be written for any glob and any trigger, so reporting on that basis pairs every path-scoped record with every trigger-scoped one — and every reported pair suspends the more restrictive record's guard (`records/registry.rs`'s `resolve_guard`). The module header's "a false positive is cheap" argument was made about glob containment, where a spurious pair is rare; it does not carry here.

The narrow rule keeps the collision demonstrable: a keyword `deny.toml` beside `paths = ["deny.toml"]` collide on the turn that edits that file — a turn anybody can name, not one the checker invented. The match goes to `stella_core::glob::match_glob`, the same matcher `select::applies_this_turn` uses for the path dimension, so the answer agrees with the code that decides whether both records fire. Ratified task names are bare verbs (`records.rs`'s `KNOWN_TASKS`), so this arm fires on keywords in practice.

`crossing_scopes` is the third `.or_else` arm of `overlapping_scope`, so a pair that already collides on a path or a trigger keeps the more specific report.

**Where the decision lives.** The module header states the rule and names `crossing_scopes`; `crossing_scopes`'s own doc comment carries the argument above — why the broad "both fire on some turn" rule is refused, which test pins the resulting silence, and which matcher answers the question. It sits there rather than in the header because `check-prose`'s density ratchet holds `crates/stella-records` to 24.33 mean header lines and a 25-line header section put the crate at 24.95; item docs are not counted, and the guard's own instruction is to cut the header and move the material to the item a reader lands on or to the design document. Both now carry it.

## Witness mechanism

`a_keyword_naming_a_path_conflicts_with_the_record_scoped_to_that_path` — a `hard` record scoped `paths = ["deny.toml"]` against an advisory record scoped `keywords = ["deny.toml"]` at equal precedence. On the old code both existing arms return `None` by construction: `overlapping_paths` iterates the second record's `paths`, which is empty; `overlapping_triggers` asks `select::shared_triggers`, which matches a term against a turn's *text* and never against a glob (`a_path_only_scope_shares_no_trigger` in `select.rs` pins that). So `detect_conflicts` returns an empty vector and `assert_eq!(conflicts.len(), 1)` fails. It passes with `crossing_scopes` chained in.

`a_directory_glob_reaches_a_keyword_naming_a_file_inside_it` is the same failure through a wildcard (`src/api/*` against the keyword `src/api/handler.rs`).

The exclusion half is pinned by `a_path_scope_and_a_task_scope_stay_silent`: `paths = ["crates/stella-core/**"]` hard against `tasks = ["refactor"]` advisory reports nothing, so the silence is this decision rather than an oversight.

## Blast radius on this repository's own records

The new arm reports **zero** additional conflicts over the 39 records in `.stella/rules/`. Measured by replaying `match_glob` over every equal-precedence pair with differing enforcement: no trigger in the corpus matches any other record's path glob. `stella context validate` (a required CI step) is unaffected.

## Exemplar

No new shape was invented — `crossing_scopes` mirrors `overlapping_triggers` next to it (ask both directions, collect the terms, join into the noun phrase `conflict_between`'s detail template expects).

## Docs

`docs/spec/adaptive-context/context-pr.md` §12 gains one paragraph naming the three things overlap is asked on and why a path scope and an unrelated trigger scope are left alone. AGENTS.md's steering summary does not name this behavior, so it is unchanged.

## File organization

The change is confined to one file, `crates/stella-records/src/records/validate.rs`; `select.rs` is untouched, because the over-report policy is `validate`'s and `shared_triggers` stays a question about a turn's text. The file is not on `scripts/file-size-baseline.txt` and stays several hundred lines under the 1500 ceiling.

**The record plane moved crates under this branch.** #5829 landed on `main` while this PR was open and moved `records/` from `stella-core` into the new `stella-records` crate, so every path above is `crates/stella-records/…` rather than the `crates/stella-core/…` #5780 was filed against. Merging `main` was textually clean and did not compile: git followed the rename and carried this branch's `use super::super::glob::match_glob` with it, where `super::super` is `stella-records`, which has no `glob` module. The follow-up commit names `stella_core::glob::match_glob`, which is what the sibling `records/select.rs` already does for the same matcher.

## Verification

`cargo fmt --check -p stella-records`, `python3 scripts/check-prose.py` and `python3 scripts/check-line-citations.py` are green locally (no new prose, density or line-citation baseline entries). Per SCR-001 and CLAUDE.md's "CI builds and tests; this laptop does not", the compile and test evidence is this PR's CI run.

## Known red check: `no edit is undone inside the branch`

`check-rebase-replay` names `crates/stella-core/src/records/validate.rs` as edited and then undone. Nothing here undid it — #5829 renamed it out of `stella-core` and this branch absorbed that through its merge of `main`, so the path leaves the final diff because `main` deleted it. #4979's mechanism needs the same edit already upstream; there is none, and a rebase here would replay the first commit onto a tree with no such path and conflict loudly. Filed as #5890; the guard is not a required check.

Tests deleted: None

Residue issues filed: #5890 (the guard reading an inherited rename as an undo).

Closes #5780

## Summary by Sourcery

Report record conflicts when a trigger names a path covered by another record's path scope.

New Features:
- Detect conflicts when a task or keyword trigger names a path matched by another record's path scope.

Bug Fixes:
- Prevent path-scoped and trigger-scoped records from silently bypassing conflict detection when they can both apply to the same named file.
- Document and enforce the narrow cross-scope rule while avoiding broad false-positive reporting for unrelated path and trigger scopes.

Documentation:
- Document the three supported record-overlap cases and the exclusion of unrelated path and trigger scopes in the adaptive-context specification.

Tests:
- Add coverage for keyword/path collisions, directory-glob matches, and intentional silence for unrelated path and task scopes.


macanderson:
## Merge conflict resolved — and it was not the conflict git reported

This branch was `dirty` against `main`. Merging `origin/main` in (`2fa98d2`) resolved with **no textual conflict at all**: #5829 landed in between and moved the record plane out of `stella-core` into the new `stella-records` crate, and git followed the rename, carrying this branch's edits to `records/validate.rs` into the new crate by itself.

The merge did not compile.

```
error[E0432]: unresolved import `super::super::glob`
  --> crates/stella-records/src/records/validate.rs:65:19
   |
65 | use super::super::glob::match_glob;
   |                   ^^^^ could not find `glob` in the crate root
```

The import this branch added was relative. In `stella-core` it reached `crate::glob`; carried into `stella-records`, which has no `glob` module, it reached nothing. A rename plus an edit is exactly the shape that merges clean and breaks — `main` was never wrong and neither was this branch.

`584aa6f` points it at `stella_core::glob::match_glob`, which is what the sibling `records/select.rs` already does for the same matcher, so the module header's claim still holds: the path dimension is answered by the same code `select::applies_this_turn` asks. The header's own citation of the old path is repointed with it.

**Verified before pushing** (targeted, per SCR-001 — no workspace build):

| command | result |
|---|---|
| `cargo check -p stella-records --all-targets` | exit 0 |
| `cargo fmt --check -p stella-records` | exit 0 |
| `cargo test -p stella-records` | 335 passed / 0 failed, + 4 integration tests |

That run includes this branch's own witness, `two_records_scoped_to_one_task_conflict_with_no_path_between_them`, so the move did not cost the behaviour the PR is for.

**Two things stay open, neither of them mine to decide:**

1. This description still says the change lives in `crates/stella-core/src/records/validate.rs` and cites `crate::glob::match_glob`. Both moved; the body wants a pass before merge.
2. `dod / dod` is red because #5780 has 8 unchecked DoD items — a checkbox state on the issue, not anything in this diff.

---

macanderson:
Following up the owner note about the PR body still citing `stella-core` paths after #5829 moved the record plane. **Yes — three citations are still stale at head `584aa6f9a`.** I have not edited the body; here is exactly what is wrong, checked against `git show refs/remotes/pr/5840:crates/stella-records/src/records/validate.rs`.

**1. The PR title.**
> `fix(stella-core): report a trigger that names a path the other record guards`

The Conventional-Commits scope is the crate, and no line of this diff touches `stella-core`. Should be `fix(stella-records):` — which is what the head commit `584aa6f9a` already uses.

**2. "The decision, written in the module header", third paragraph.**
> The match goes to `crate::glob::match_glob`, the same matcher `select::applies_this_turn` uses for the path dimension

`crate::glob` was correct while `validate.rs` lived in `stella-core`. It is now a cross-crate edge — the file's own header and its `use` line both say `stella_core::glob::match_glob`. Should read `stella_core::glob::match_glob`.

**3. "File organization", first sentence.**
> The change stays in `crates/stella-core/src/records/validate.rs`; `select.rs` is untouched

The file is `crates/stella-records/src/records/validate.rs`. The rest of that paragraph still holds as measured at the head: 1081 lines, `select.rs` untouched, and no entry matching `validate` or `stella-records` in `scripts/file-size-baseline.txt`.

**Not stale, to save a reviewer the check:** `paths = ["crates/stella-core/**"]` in the "exclusion half" paragraph is a test fixture string inside `a_path_scope_and_a_task_scope_stay_silent`, and it is verbatim what the test asserts on. Leave it. The relative citations (`records/validate.rs`, `records/registry.rs`'s `resolve_guard`, `select.rs`'s `a_path_only_scope_shares_no_trigger`) all resolve inside `stella-records` and are fine.

---

**One further item from the same cause, which blocks the merge.** The "Verification" section says:

> `cargo fmt --all`, `make guards-fast` and `python3 scripts/check-prose.py` are green locally (no new prose or line-citation baseline entries).

`check-prose.py` is **red at this head**, exit 1:

```
  crates/stella-records: 24.33 allowed, 24.95 mean header lines
```

The same command on `origin/main` (`913e21a06`) exits 0. The banned-construction half is clean and `check-line-citations.py` is clean; it is the header-length density ratchet alone. That claim was true when the +25-line `//!` header sat in `stella-core`, whose larger corpus absorbed it — `stella-records` is a small crate and its ceiling on line 43 of `scripts/prose-density-baseline.txt` is 24.33. `prose` is a `make gate` step and `docs-guards.yml` runs it, so the `docs guards` check should go red when it reports. Per AGENTS.md the remedy is to cut the header, not to raise the baseline.

Verified alongside: `584aa6f9a` is the head so the `E0432` fix is present; the crate compiles (`cargo test -p stella-records validate` — 26 passed, 0 failed) so the fix is confirmed by observation rather than by the commit message; and `git merge-tree --write-tree origin/main refs/remotes/pr/5840` still exits 0.

Per-checkbox DoD verdicts are in macanderson/stella#5780.

---

macanderson:
## `no edit is undone inside the branch` — answered, not a #4979 revert

The guard names `crates/stella-core/src/records/validate.rs`: this branch's first commit edited it, and the merge of `main` then removed it. That pair is an upstream **rename**, not an undo of anyone's change.

#5829 moved the whole record plane from `stella-core` into the new `stella-records` crate. On the base this branch now sits on (`913e21a06`), `crates/stella-core/src/records/validate.rs` does not exist — `git cat-file -e 913e21a06:crates/stella-core/src/records/validate.rs` reports `does not exist`. The edit lives at the new path in the branch tree, and `913e21a06:crates/stella-records/src/records/validate.rs` contains no `crossing_scopes` (grep count 0), so the base does not carry a copy of this branch's edit.

That is what separates this from #4979. The hazard there needs upstream to carry **the same edit**, so a rebase deduplicates the branch's copy and keeps the undo, leaving a silent live revert. Here upstream carries a rename of the file, so a rebase replaying the first commit hits a missing path and conflicts loudly — it cannot produce a quiet revert of anything.

Flattening is the guard's remedy and it needs a force-push. Another session pushed the merge and the import repoint to this branch, so rewriting it would overwrite work I did not author. The merge method here is squash, which lands one commit either way.

macanderson:
Automated PR sweep. **`rebase-replay` is red on `584aa6f9` and it is a false positive — do not flatten this branch to clear it.** Filed as #5890.

The guard names `crates/stella-core/src/records/validate.rs` as edited and then undone. Nothing here undid it; #5829 renamed it out of `stella-core` and this branch absorbed that through its merge of `main`. The merge's two sides settle it:

```
$ P=crates/stella-core/src/records/validate.rs
$ git rev-parse --short 2fa98d2^1 2fa98d2^2
f6aee54   913e21a
$ git cat-file -e 2fa98d2^1:$P   # branch side -> YES
$ git cat-file -e 2fa98d2^2:$P   # main side   -> NO
```

The path leaves the final diff because `main` deleted it, not because a commit here reverted one.

**Your work is intact**, which is the claim worth checking rather than taking on trust:

```
$ git diff origin/main..HEAD --stat -- crates/stella-records/src/records/validate.rs
 crates/stella-records/src/records/validate.rs | 163 +++++++++++++++++++++++++-
 1 file changed, 158 insertions(+), 5 deletions(-)
```

`f6aee54` added 163 lines at the old path; they are present at the new one, with `584aa6f` repointing the matcher.

And #4979's mechanism cannot fire on this branch: it needs the same edit already upstream so a rebase drops the branch's copy and keeps the undo. There is no duplicate edit upstream and no undo. A rebase here would replay `f6aee54` onto a tree with no `stella-core/src/records/validate.rs` and conflict loudly.

**Nothing to do on this PR.** The guard is not a required check (#4997), so this red does not block the merge, and the remedy it prints would have you rewrite history to silence a hazard that is not there. #5890 asks for the guard to stop reading an inherited rename as an undo, and names this PR as the case that has to pass afterward.

Everything else here is clean: the branch merges without conflict against `913e21a`, and no other completed check on this head is failing.

---

macanderson:
Automated PR sweep. **`check-prose` is fixed and pushed** (`ad4a5b3d1`, `99f7e11b4`). The `rebase-replay` red is left alone — the comment above already establishes it as an inherited rename rather than an undo, and #5890 tracks the guard.

**What was red.** `docs guards` / `gate steps no other workflow runs` failed on the density ratchet, not on a banned construction:

```
  crates/stella-records: 24.33 allowed, 24.95 mean header lines
```

The 25 `//!` lines this branch added to `records/validate.rs` were absorbed by `stella-core`'s larger corpus and are not absorbed by `stella-records`, whose ceiling is line 43 of `scripts/prose-density-baseline.txt`. Raising that number is the expedient AGENTS.md forbids, so the header was cut instead.

**Where the argument went.** The guard's own instruction names the destinations — a design document, or the item a reader lands on. Both already carried it in this branch, so the 25-line section moved onto `crossing_scopes`'s `///` doc comment, which the density ratchet does not count and which is where a reader of that function arrives. Nothing was dropped: the refusal of the broad "both fire on some turn" rule, `resolve_guard`'s blast radius, `a_path_scope_and_a_task_scope_stay_silent` pinning the silence, and `stella_core::glob::match_glob` as the matcher are all there verbatim. `docs/spec/adaptive-context/context-pr.md` §12 states the same rule.

**#5780's first DoD item asks for the decision in the module header, and it is still there.** The second commit rewrites the header's existing trigger paragraph *in place* — same six lines, so the crate stays at 24.33 — to state the rule and name `crossing_scopes` as where the argument lives:

> A path scope crosses a trigger scope only where one turn proves it: that is `crossing_scopes`, whose doc comment carries the rule and why it is narrow.

The issue's Evidence block still names a header section titled "A path scope crosses a trigger scope only where one turn proves it"; that heading is gone, and the sentence above is what satisfies the box now.

**Verified locally** (targeted only, per SCR-001):

| command | result |
|---|---|
| `python3 scripts/check-prose.py` | exit 0 — `crates/stella-records` at 24.33, its ceiling |
| `python3 scripts/check-line-citations.py` | exit 0 — 260 grandfathered, none added |
| `cargo fmt --check -p stella-records` | exit 0 |

Both commits change comment text only — no statement, signature or test moved — so the compile and test evidence stays this PR's CI run.

**Also corrected:** the three stale citations flagged above and never fixed. The title scope is `fix(stella-records)`, the body says `stella_core::glob::match_glob` rather than `crate::glob::match_glob`, and File organization names the `stella-records` path. The body now also records the `rebase-replay` red and #5890 so a reviewer does not have to reconstruct it from the thread.

---

macanderson:
**This PR and #5899 collide in one import hunk, and both obvious resolutions are wrong.** Flagging it here and on #5899 so whichever lands second resolves it correctly rather than reddening `main` — the tree is red right now from exactly this shape of collision (#5901), so this is worth spending a minute on.

Each PR is correct against `main` on its own; they conflict only with each other, in `crates/stella-records/src/records/validate.rs`:

```
<<<<<<< HEAD                                    (#5899)
use stella_learn::redact::redact_secrets;
=======                                         (#5840)
use stella_core::glob::match_glob;
use stella_core::redact::redact_secrets;
>>>>>>>
```

**Why both naive resolutions fail.** #5899 extracts the learning plane and moves two things this file needs in different directions: `redact` goes to `stella-learn`, and `match_glob` goes to **`stella-protocol`**, not `stella-learn` (it is `crates/stella-protocol/src/glob.rs` there, alongside `hash::record_hash`). So "take both sides" leaves `stella_core::glob` and `stella_core::redact`, neither of which exists after #5899; "take HEAD" drops `match_glob` and breaks the validation this PR adds.

**The resolution that works:**

```rust
use stella_learn::redact::redact_secrets;
use stella_protocol::glob::match_glob;
```

No manifest change is needed — `stella-records` already depends on both `stella-protocol` and `stella-learn` in the composed tree.

**Verified, not assumed.** Merged `origin/main` + #5899 + #5840 in a scratch worktree, applied the resolution above, and compiled the crate:

```
$ cargo check -p stella-records --all-targets
    Checking stella-protocol v0.9.327
    Checking stella-core v0.9.327
    Checking stella-learn v0.9.327
    Checking stella-records v0.9.327
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 16.78s
```

Exit code 0.

Nothing to change in this PR as it stands — it is correct against today's `main`, and `stella_protocol::glob` does not exist yet to point at. This is a note for the merge, not a review finding.

---

macanderson:
`no edit is undone inside the branch` is red here, and I believe it is a **false positive**. Not pushing a flatten — the evidence says this branch is correct as it stands.

### What the guard saw

`scripts/check-rebase-replay.sh` computes `suspects = touched − net`:

```sh
net     = git diff --name-only  $base $head          # paths the branch changes
touched = git log --no-merges --name-only $base..$head   # paths any non-merge commit touched
```

`crates/stella-core/src/records/validate.rs` is in `touched` (from `f6aee54`) and not in `net`, so it is flagged.

### Why the hazard cannot occur here

The merge base is `913e21a06` — which is #5829, *"refactor(stella-records): extract the record plane from the engine"*. So:

| | `crates/stella-core/src/records/validate.rs` |
|---|---|
| at base `913e21a06` | **absent** (`git cat-file -e` fails) |
| at head `99f7e11` | **absent** |

The path does not exist at *either* end of the range. #4979's hazard needs a path that exists upstream carrying somebody else's edit, so that the branch's undo becomes a live revert of it. There is no upstream copy of this path to revert — the record plane moved crates.

What actually happened on this branch is a rebuild, not an undo:

- `f6aee54` edited the file at its **old** path, before the extraction landed (+158/−5).
- `2fa98d2` merged `main`, which carried #5829 and removed that path from `stella-core`.
- `584aa6f` — *"repoint the path matcher after the record plane moved crates"* — re-landed the work at the new path.

The branch's net diff confirms nothing was lost:

```
 crates/stella-records/src/records/validate.rs | 165 ++++++++++++++++++++++++--
 docs/spec/adaptive-context/context-pr.md      |   7 ++
```

Rebasing this branch onto today's `main` would make `f6aee54` fail to apply against a file that is not there — a conflict, which is loud. It cannot produce the silent clean-merging revert the guard exists to prevent.

### Why I am not flattening

AGENTS.md notes the flatten "costs nothing" because the tree is identical, and that is still true — but it would be paying a history rewrite to satisfy a check that is wrong about this branch, and it would erase the evidence that the guard has this gap. A force-push is also not something I will do unprompted on someone's branch.

Filed the guard gap as **#5910**. The narrow fix is to drop a suspect whose path is absent at both `$base` and `$head`: a path that exists at neither end was relocated or deleted upstream, and has no upstream copy to revert.

Nothing else on this PR is failing on its own account. `main is not known-broken` will stay red on every open PR until #5906 lands and closes #5901.

---

macanderson:
**`rebase-replay` is red on this branch, and I am deliberately not fixing it — the only remedy is rewriting history on a branch that isn't mine.** Reporting it instead, with the diagnosis, so whoever owns the branch can flatten it in one command.

### What the guard found

Run [33952545497](https://github.com/macanderson/stella/actions/runs/33952545497), job *"no edit is undone inside the branch"*:

```

  crates/stella-core/src/records/validate.rs
      f6aee54b2  fix(stella-core): report a trigger that names a path the other record guards
```

### Why it fired

The branch's first commit `f6aee54b2` was written before #5829 moved the record plane, so it edited `crates/stella-core/src/records/validate.rs`. Merging main and repointing (`584aa6f9a`) moved the work to `crates/stella-records/src/records/validate.rs`. The branch's history therefore edits the old path and then undoes it, while the branch's diff against `98af1d6b7` never mentions it — which is exactly the shape #4979 describes.

### The hazard here is structurally absent, but the guard is still right to fail

`crates/stella-core/src/records/validate.rs` **does not exist on `main`** — #5829 deleted it. The failure mode the guard protects against is a rebase silently reverting an upstream edit to that path, and there is no upstream edit left to revert. So this specific instance cannot cause the harm.

That is not an argument for overriding it. The guard cannot see that, a reader six months from now cannot either, and a branch whose history edits-then-undoes a path is worth flattening on its own terms. Flattening costs nothing — the tree is identical either way, which is why the guard offers no acknowledgement path.

### The fix, from the guard's own output

```sh
git rebase -i 98af1d6b7   # drop the pair
# or:
git reset --soft 98af1d6b7 && git commit
```

I can't run either: rewriting history on someone else's branch is off-limits for me, and a force-push would invalidate the author's checkout.

### Separately — the #5899 collision is still live

The import at `crates/stella-records/src/records/validate.rs:45` is still `use stella_core::glob::match_glob;`. That is correct against today's `main`, so nothing needs changing now — but `584aa6f9a` repointed the *record* paths and not this one, so the collision I flagged [earlier](https://github.com/macanderson/stella/pull/5840#issuecomment-5550001124) is unchanged. Whichever of this PR and #5899 lands second still needs:

```rust
use stella_learn::redact::redact_secrets;
use stella_protocol::glob::match_glob;
```

Verified to compile (`cargo check -p stella-records --all-targets`, exit 0) on the composed tree.

---

Issue #5545, opened 2026-08-30
Epic: every correctness finding from a workspace review is fixed with a witness or dismissed with evidence


macanderson:
## Outcome

Every correctness finding a workspace review turned up is settled: fixed with a witness test that fails on the old code and passes on the new, or dismissed with evidence that says why it is not a defect. No finding sits in the backlog as an unread claim.

## Why one epic

These children were each found by reading the code rather than by a failing test. That is what they have in common and it is what makes them one deliverable rather than twenty chores.

A defect found by reading has a specific hazard: nothing in CI is going to find it again, and nothing is going to tell anyone it is still there. Each one is individually small enough to walk past, and several of them have already survived one review pass each. The largest slice came from the workspace-wide correctness audit whose fix batch was #5489 — that batch fixed eleven and filed the rest, which are these.

They also share a shape worth naming, because it is the shape this repository cares most about. Most of these are not crashes. They are a path that reports a better outcome than happened: a dropped tool call that reads as "the model asked for nothing", a failed checkpoint delete that answers `200 deleted`, a truncated replay logged as a healthy completion, a clipped output rendered with no clip marker, five dashboard numbers that each pick the more reassuring of two states. CLAUDE.md's rule is that a run which completes and publishes a wrong number spends the exact reputation this project exists to build. That is what these are.

## Children

**Reported better than it happened**

| Issue | What |
|---|---|
| #5496 | A malformed Anthropic tool-call start drops the whole call; the turn reports none while `finish_reason` says otherwise |
| #5500 | Five dashboard numbers are wrong, each by picking the more reassuring of two states |
| #5503 | A failed checkpoint delete answers `200 deleted`; a truncated replay logs as a healthy completion |
| #5499 | The two transcript renderers each drop the other's honesty marker |
| #5504 | `GateBoard` renders as an unlabelled note on the plain transcript, and four README claims are false |
| #5505 | Store corruption found after open is reported without the file name or the remedy |
| #5521 | A typo in an org-managed settings file silently voids the whole managed ceiling |

**Silent loss or bypass**

| Issue | What |
|---|---|
| #5502 | `edit_file` writes from a stale snapshot, silently destroying a concurrent write |
| #5498 | A fleet worker's cost is never checked: `NaN` wedges the budget, a negative value bypasses it |
| #5516 | Several turn aborts hand back a transcript the next provider call rejects |
| #5386 | A sub-agent ceiling of 30 seconds or less gives the child a deadline of now |
[...]

PR #5846, opened 2026-09-05
fix(stella-tools): route exec.rs tests through the safe bash_command helper

### LINKED ISSUE #5768: Two exec.rs tests still start bash directly instead of using the safe helper
## Problem

Two tests in `crates/stella-tools/src/exec.rs` start a shell by calling `Command::new("bash")` directly:

- `capped_capture_bounds_a_runaway_child`
- `a_refusing_capture_stops_the_read_and_leaves_the_child_to_kill`

On Windows, if no WSL distro is installed, `Command::new("bash")` runs the Windows WSL launcher instead of a real shell. That launcher prints an install message and exits with an error, instead of running the test's script.

This is the same bug as #4861, but only for these two tests. #4861 fixed the main code paths (`bash.rs` and `hook_runner.rs`) and added a shared helper, `bash_command()`, that finds a real shell safely. These two tests in `exec.rs` were missed by that fix.

Right now this bug does not break anything. `windows-check.yml` runs a narrower Windows test (`--test group_kill_witnesses`), so these two tests never run there. But if Windows CI ever runs the full `--lib` test suite for `stella-tools`, these two tests would fail on a real Windows machine.

## Fix

Change both tests to call `crate::shell_resolve::bash_command()` instead of `Command::new("bash")`. This is the same fix already used in `bash.rs` and `hook_runner.rs`. No visibility change is needed — `shell_resolve` is a private (`mod`) module and `exec.rs` is in the same crate, so it can already call `bash_command()`.

## Definition of done

- [x] Both tests in `exec.rs` call `crate::shell_resolve::bash_command()` instead of `Command::new("bash")`.
- [x] `cargo test -p stella-tools --lib exec::tests` passes on unix (the two named tests above still pass).
- [x] No `Command::new("bash")` remains anywhere in the crate except inside `shell_resolve.rs` and its own tests/docs — confirm with `rg 'Command::new\("bash"\)' crates/stella-tools/src`.
- [x] **Witness**: `rg 'Command::new\("bash"\)' crates/stella-tools/src/exec.rs` finds two matches before the fix and zero after. That grep, run once against the old code and once against the new code, is the fail→pass proof for this change (a full Windows repro is not practical here since the defect is dormant under current CI wiring).
- [x] The fix stays inside the existing `exec.rs` test module — no new file is needed. `exec.rs` is not on the crate's god-file list ([AGENTS.md § God files](https://github.com/macanderson/stella/blob/main/AGENTS.md)), and this change must not be the one that puts it there.
- [x] No new comments or doc text beyond what explains the one-line swap (`make prose` stays clean).
- [x] No docs update needed — this is an internal test-only change with no user-facing or public API behavior change.

## Notes

Filed as a known follow-up when #4861 was fixed and closed. Not a duplicate of #4861 — that issue is closed; this is the small piece its fix left behind. Part of epic #5545 ("every correctness finding from a workspace review is fixed with a witness or dismissed with evidence"), same epic #4861 belongs to.

macanderson:
## What / why

Two tests in `crates/stella-tools/src/exec.rs` spawned a shell with
`Command::new("bash")` directly instead of going through
`shell_resolve::bash_command()`, the helper #4861 added for `bash.rs` and
`hook_runner.rs`. On Windows with no WSL distro installed,
`Command::new("bash")` resolves through `CreateProcessW`'s search order,
which checks `%SystemRoot%\System32` before `PATH` — Windows ships a
`bash.exe` stub there (the WSL launcher) that prints an install prompt and
exits instead of running the script. `bash_command()` sidesteps that by
finding an absolute path to a real shell on Windows and running that
directly.

This bug is dormant today: `windows-check.yml` runs only
`--test group_kill_witnesses` for `stella-tools`, so these two `--lib`
tests never execute on the Windows runner. It would surface the moment
Windows CI ran the crate's full `--lib` suite.

## The fix

Both tests now build their `Command` via
`crate::shell_resolve::bash_command().expect("a real shell is on PATH")`
instead of `Command::new("bash")`. `shell_resolve` is a private (`mod`)
module and `exec.rs` is in the same crate, so no visibility change was
needed. Exemplar for the pattern: `crates/stella-tools/src/bash.rs` and
`crates/stella-tools/src/hook_runner.rs`, both already fixed by #4861.

## Witness

Old code (`origin/main`), reading `exec.rs` at that commit: 2 matches for
`Command::new("bash")`. New code (this branch): 0 matches. Both renamed
tests (`capped_capture_bounds_a_runaway_child` and
`a_refusing_capture_stops_the_read_and_leaves_the_child_to_kill`) still
pass on unix under `cargo test -p stella-tools --lib exec::tests` (7
passed, 0 failed).

## Tests deleted

None.

## Residue

None — this was the last open child of epic #5545, closing the loop #4861
started.

Closes #5768

## Summary by Sourcery

Bug Fixes:
- Route the exec test shell commands through the platform-safe bash resolver so they use a real shell on Windows instead of the system WSL launcher stub.


macanderson:
Merged `main` in (`27e7d23`) so CI re-runs against a base that compiles. **The failure here was never this PR's.**

### What was red

`fmt + clippy + test` failed on:

```
error[E0433]: cannot find `ingest` in `stella_core`
error[E0432]: unresolved import `stella_core::records`
error: could not compile ... exit code 101
```

That is #5901 — `main` did not compile, because `crates/stella-core/tests/published_records_do_not_collide.rs` still imported modules the record-plane extraction (#5829) had moved to `stella-records`. Nothing in this PR's diff touches either module. The run was created at 06:37, and the repair (#5906) merged at 07:20, so this branch's checks were the last ones taken against the broken tree — its head had not moved since 06:10.

### Why a merge rather than a re-run

A re-run would have done just as well, and it is the cheaper remedy. I could not issue one: `POST /actions/runs/{id}/rerun-failed-jobs` returns `403 Resource not accessible by integration` for this session's token, which has no `actions: write`.

A merge is the other sanctioned route — AGENTS.md's answer to a base that has recovered — and it is a merge commit rather than a rebase, so nothing in this branch's history is rewritten and any local checkout stays valid. The two remedies that would have been faster are the two explicitly forbidden here: an empty commit, or a close and reopen.

Verified on the merged tree before pushing:

- `crates/stella-core/tests/published_records_do_not_collide.rs` — gone.
- `crates/stella-records/tests/published_records_do_not_collide.rs` — present.
- `git grep -E 'stella_core::(records|ingest)' -- '*.rs'` — no matches anywhere in the tree.

The merge itself was clean; the only incoming change of consequence is that rename.

This does not vouch for the PR's own diff — CI now gets to judge that against a base that compiles, which it could not do before. If it goes red again, that result is real and worth reading.

Related: #5910 (a `check-rebase-replay` false positive on paths #5829 moved, hitting #5840 and #5894) and #5913 (`main-red-hold`'s failure is sticky after `main` recovers) both came out of the same outage.

---

PR #5847, opened 2026-09-05
fix(stella-cli): read the issue before writing the escalation record

### LINKED ISSUE #5702: Writing the escalation record can wipe out an edit a person just made
## What happens

`crates/stella-cli/src/self_driving_cmd/backlog.rs`'s `escalate` writes the escalation record by replacing the whole issue body:

```rust
provider.edit(key, None, Some(&escalation::stamp(body, &record))).await?;
```

`body` is the text the loop read earlier in the same turn — for the `WorkOutcome::Failed` path, the body `backlog::resolve` returned before a turn that may have run for minutes. `GhIssueProvider::edit` shells out to `gh issue edit <n> --body <text>`, which is a whole-body replace with no compare-and-set.

So if a person edits the issue body while the turn is running, the escalation write silently reverts their edit. The window is the length of one work turn, which is exactly when a person is most likely to be reading and annotating the issue.

## Why it matters

The escalation comment is the loop's apology for parking an issue. Reverting the reporter's clarification in the same gesture is worse than not writing the record at all — and it is invisible: nothing reports it, and GitHub shows the loop as the last editor.

## Files

- `crates/stella-cli/src/self_driving_cmd/backlog.rs` — `escalate`, the `provider.edit` call.
- `crates/stella-cli/src/issue_provider.rs` — `GhIssueProvider::edit`, the `gh issue edit --body` shell-out.
- `crates/stella-autonomy/src/escalation.rs` — `stamp`, `parse`, `strip`. The marker format is already designed to be surgical; the write is what is not.
- `crates/stella-cli/src/self_driving_cmd/drive/triage.rs` — `escalate`, which passes `&issue.body` down from the earlier read.

## Verify

1. Point a test fixture provider at an issue whose body is `A`.
2. Call `backlog::escalate_blocking` with body `A`.
3. Between the read and the write, change the fixture's stored body to `A + "\n\nsomebody's note"`.
4. The write lands `stamp(A, record)`, and the note is gone.

## Options

- **Re-read the body immediately before the edit.** Narrows the window to one round trip rather than one turn. Cheap, and does not close it.
- **Read-modify-write with a check.** Re-read, and refuse (or retry) if the body changed since the read the record was built from. Needs no new provider method — `get` already exists.
- **Move the record to a comment.** Removes the hazard entirely, and costs one comment-read per escalated issue per cycle. `IssueProvider` has `comment()` (write) and no comment read, so this needs a new trait method plus every implementor. That trade was weighed and declined in #5662; it becomes right if the clobber turns out to happen.

## Definition of done

- [x] An escalation write cannot silently drop a body edit made after the loop read that body.
- [x] A witness test that fails on the current code: a fixture whose stored body changes between the read and the escalation write, asserting the later text survives.
- [x] Whatever is chosen is written down in `crates/stella-cli/src/self_driving_cmd/backlog.rs`'s `escalate` doc comment, so the next reader does not have to re-derive the hazard.
- [x] `docs/spec/backlog-self-driving.md`'s escalation section says what happens on a concurrent edit.

macanderson:
## What and why

Writing the escalation record means sending a whole issue body — `gh issue
edit --body` takes no other shape. `backlog::escalate` took the body the loop
had read earlier in the same turn and sent that back. A work turn runs for
minutes, so a note a person added while it ran was carried away, with nothing
to report it and the loop showing as the last editor.

`escalate` no longer accepts a body at all, so the stale snapshot is not
representable. The new `crates/stella-cli/src/self_driving_cmd/backlog/record.rs`
does the write:

- It reads the issue right before it writes, and builds the record from that
  read, so the attempt count is the tracker's own rather than a stale copy's.
- A failed read aborts the write. A body nobody can see is a body nobody may
  replace. The caller already reports that as a transient tracker failure.
- After the write it reads once more. If the record is gone, a person saved
  their own text in the gap; theirs is the newer text, so the same record is
  stamped on top of it, once. The loop cannot fight a person over one body.

GitHub offers no compare-and-set on an issue body, so an edit saved inside one
round trip is still lost. That is written down in the module docs and in
`doc:backlog-self-driving` rather than left for the next reader to re-derive.

The record stays in the body rather than moving to a comment: `to_queue_issue`
parses the cooldown out of the body every queue read already carries, and a
comment would cost a read per escalated issue. #5662 weighed that and declined
it; nothing here changes the arithmetic.

## Witness

`a_note_added_during_the_turn_survives_the_escalation` in
`crates/stella-cli/src/self_driving_cmd/backlog/record.rs`. A fixture tracker
stores a body that already carries `somebody's note: this needs the 0.9.3
client`, and the loop holds no body at all. The test asserts the note is still
there after the escalation lands, alongside the record and the label.

It fails on the old code by construction twice over: `escalate` took the body
to write as a parameter, so the tracker's own text was never read, and the
note was replaced by whatever the caller passed. The signature no longer has
that parameter, so the old code cannot even express the call.

Two more tests cover the rest of the rule:

- `a_write_that_lands_in_the_gap_keeps_its_text_and_gets_the_record_back` —
  the fixture saves a person's text right after the first `edit`, and the
  single repair puts the record back on their text.
- `an_unreadable_body_is_never_written_over` — a tracker whose reads fail and
  whose writes would succeed sees no write at all.

Exemplar: the fixture-tracker shape is the one already in
`backlog.rs`'s test module, extended to store a body so `get` and `edit` talk
to each other.

## Tests deleted

None. `escalating_stamps_a_record_the_next_run_can_read` moved from
`backlog.rs` to `backlog/record.rs`, where the fixture that stores a body
lives, and asserts the same thing plus the single-marker property.

`FixtureProvider`'s `edited` and `labelled` recorders in `backlog.rs` went with
it — no test read them any more.

## Residue

None found beyond what this change fixes.

Closes #5702

#5702's four DoD boxes are ticked, checked against the diff at `9a03228` rather
than against this description.

- **The stale body is not representable.** `body: &str` is gone from `escalate`
  and `escalate_blocking`, and from all three call sites in `drive/triage.rs`
  (`escalate`, plus the two `assess_one` paths that passed `&full.body` and
  `&body`). `record::write` opens with `provider.get(key).await?.body`.
- **The witness fails on the old code.** Both new race tests call
  `backlog::escalate` with five arguments against a signature that took six, so
  they cannot compile on `main`; semantically,
  `a_note_added_during_the_turn_survives_the_escalation` asserts text the loop
  never held, which `stamp(body, record)` could not have contained.
  `an_unreadable_body_is_never_written_over` pins the fail-closed direction.
- **The choice is recorded where the next reader lands.** `escalate`'s doc
  comment names it ("**No body is passed in.**") and `record.rs`'s module header
  carries the remainder.
- **The spec states the residual gap honestly.** `doc:backlog-self-driving`'s
  new paragraph says the round-trip gap is narrowed, not closed.

Detail is in [this comment](https://github.com/macanderson/stella/issues/5702#issuecomment-5549750291).

re-read the boxes — the edit is the trigger, never the fix.

## Summary by Sourcery

Make escalation recording read the current issue body immediately before writing so stale turn snapshots cannot erase human updates.

Bug Fixes:
- Prevent escalation records from overwriting issue text added or changed by people during a self-driving work turn.
- Fail closed when the current issue body cannot be read before recording an escalation.
- Repair a record lost to an edit race by restoring it once on the newer issue text.

Enhancements:
- Centralize escalation-record writes so attempt counts are derived from the tracker's current issue body.
- Document the unavoidable race window caused by trackers lacking compare-and-set body updates.

Documentation:
- Document the current-body read, fail-closed behavior, and residual edit race in the backlog self-driving specification.

Tests:
- Add coverage for preserving concurrent notes, repairing edits that land during recording, and avoiding writes when issue reads fail.
- Move and expand escalation record persistence coverage around the body-aware fixture.

### DIFF
--- a/crates/stella-cli/src/self_driving_cmd/backlog.rs
+++ b/crates/stella-cli/src/self_driving_cmd/backlog.rs
@@ -30,6 +30,8 @@ use crate::query_format::{QueryFormat, Rows};
 use super::config::LoopConfig;
 use super::state::LoopState;
 
+mod record;
+
 /// How many open issues cross the port to produce one cycle's batch.
 ///
 /// This is a page size, and a page the backlog outgrows silently changes the
@@ -521,7 +523,6 @@ pub(super) fn escalate_blocking(
     provider: &dyn IssueProvider,
     key: &str,
     why: &str,
-    body: &str,
     policy: &EscalationPolicy,
     signature: &str,
 ) -> Result<EscalationRecord, String> {
@@ -534,7 +535,6 @@ pub(super) fn escalate_blocking(
             provider,
             &IssueKey::from(key),
             why,
-            body,
             policy,
             signature,
         ))
@@ -554,29 +554,32 @@ pub(super) fn escalate_blocking(
 /// costs no extra call to the tracker. It is an HTML comment, so nobody
 /// reading the rendered issue sees it.
 ///
+/// **No body is passed in.** A turn runs for minutes, and the text a person
+/// is looking at can change in that time. [`record::write`] reads the issue
+/// itself, right before it writes, and that module carries the rest of the
+/// rule — what happens to an edit that lands in the gap, and how wide the gap
+/// can be.
+///
 /// Returns the record it wrote, so the caller can count an issue that has
 /// just used up its last attempt.
 pub(super) async fn escalate(
     provider: &dyn IssueProvider,
     key: &IssueKey,
     why: &str,
-    body: &str,
     policy: &EscalationPolicy,
     signature: &str,
 ) -> Result<EscalationRecord, IssueError> {
-    let record = escalation::next(
-        escalation::parse(body).as_ref(),
-        escalation::classify(why),
-        &crate::timefmt::rfc3339_utc_now(),
-        crate::timefmt::now_unix(),
-    );
-
     provider
         .relabel(key, &[stella_autonomy::ESCALATION_LABEL.to_owned()], &[])
         .await?;
-    provider
-        .edit(key, None, Some(&escalation::stamp(body, &record)))
-        .await?;
+    let record = record::write(
+        provider,
+        key,
+        escalation::classify(why),
+        &crate::timefmt::rfc3339_utc_now(),
+        crate::timefmt::now_unix(),
+    )
+    .await?;
 
     let next_step = match escalation::retry_after(&record, policy) {
         Some(wait) => format!(
@@ -794,16 +797,14 @@ mod tests {
 
     /// A tracker that is not GitHub and is not a process.
     ///
-    /// Records what it was asked to write, so a test can assert not only what
-    /// was filed but that nothing **was** — which is the half that matters for
-    /// a refusal, since a conformance check that ran after the write would pass
-    /// the same assertions on the returned value.
+    /// It keeps every draft it was asked to file. So a test can assert that
+    /// nothing was filed at all, which is what a refusal has to show: a check
+    /// that ran after the write would pass the same assertions on the value it
+    /// returned.
     #[derive(Default)]
     struct FixtureProvider {
         open: Vec<Issue>,
         filed: std::sync::Mutex<Vec<IssueDraft>>,
-        edited: std::sync::Mutex<Vec<String>>,
-        labelled: std::sync::Mutex<Vec<String>>,
     }
 
     impl FixtureProvider {
@@ -816,14 +817,6 @@ mod tests {
 
         fn filed(&self) -> Vec<IssueDraft> {
             self.filed.lock().expect("fixture lock").clone()
-        }
-
-        fn edited(&self) -> Vec<String> {
-            self.edited.lock().expect("fixture lock").clone()
-        }
-
-        fn labelled(&self) -> Vec<String> {
-            self.labelled.lock().expect("fixture lock").clone()
         }
     }
 
@@ -859,13 +852,9 @@ mod tests {
         async fn relabel(
             &self,
             _key: &IssueKey,
-            add: &[String],
+            _add: &[String],
             _remove: &[String],
         ) -> Result<(), IssueError> {
-            self.labelled
-                .lock()
-                .expect("fixture lock")
-                .extend(add.iter().cloned());
             Ok(())
         }
 
@@ -873,14 +862,8 @@ mod tests {
             &self,
             _key: &IssueKey,
             _title: Option<&str>,
-            body: Option<&str>,
+            _body: Option<&str>,
         ) -> Result<(), IssueError> {
-            if let Some(body) = body {
-                self.edited
-                    .lock()
-                    .expect("fixture lock")
-                    .push(body.to_owned());
-            }
             Ok(())
         }
     }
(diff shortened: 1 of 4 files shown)


macanderson:
**Answering the Sourcery ❌ row** — "Prevent escalation from silently overwriting issue-body edits made after the loop's earlier read, including providing a witness test for concurrent edits."

Sourcery is reading the diff correctly, and the row should not be waved away: a residual window survives this change, and it is narrower than the DoD's wording asks for.

**What the diff does close.** `record::write` (`crates/stella-cli/src/self_driving_cmd/backlog/record.rs`) reads the issue immediately before it writes and builds the record from that read, so the clobber window drops from one work turn — minutes — to one round trip. `escalate` no longer accepts a body at all, so a stale snapshot is not representable in the signature.

**What survives.** A person's save that lands *between* that pre-write `get` and the `edit` is still lost: the `edit` replaces it, the post-write `get` then returns a body that does parse to the record, so the repair branch does not fire and nothing reports the loss. The witness `a_write_that_lands_in_the_gap_keeps_its_text_and_gets_the_record_back` covers the *post*-write half of the window, where the repair does fire — not the pre-write half. Sourcery's description of the coverage is accurate.

**Why it is not fixable inside this shape.** The tracker edit is `gh issue edit <n> --body <text>`, a whole-body replace, and GitHub offers no compare-and-set on an issue body — so no ordering of reads and writes closes the gap, only narrows it. The two things that would close it are the ones #5662 already weighed: move the record to a comment (costs an `IssueProvider::comment` read per escalated issue per cycle, and a new trait method plus every implementor), or a tracker with conditional writes. Both are larger than this PR and neither is started here.

So the honest reading of #5702's first DoD box — "cannot silently drop a body edit made after the loop read that body" — is that it is satisfied for the turn-length window it was filed about and not for the one-round-trip remainder, which is documented in the module header and in `docs/spec/backlog-self-driving.md` rather than left for the next reader to re-derive. Whether that is enough to tick the box, or whether the remainder should be split into its own issue under SCR-004, is your call rather than something this PR should assert.

---

macanderson:
## Answering Sourcery's ❌, and the DoD sweep behind it

Sourcery's ❌ and the red `dod / dod` are the same objection reached twice, and **it is correct**. Answering rather than fixing, because the fix as the issue words it is not available against this tracker.

Checked against the branch at `9a03228`, item by item:

| #5702 DoD item | Verdict | Evidence |
|---|---|---|
| An escalation write cannot silently drop a body edit made after the loop read that body | **partial** | see below |
| A witness test … a fixture whose stored body changes *between the read and the escalation write*, asserting the later text survives | **not met as worded** | see below |
| Written down in `escalate`'s doc comment | **met** | `backlog.rs` — "*No body is passed in. … `record::write` reads the issue itself, right before it writes, and that module carries the rest of the rule — what happens to an edit that lands in the gap, and how wide the gap can be.*" |
| `docs/spec/backlog-self-driving.md`'s escalation section says what happens on a concurrent edit | **met** | lines 439–440: "*GitHub offers no compare-and-set on an issue body, so an edit saved inside that one round trip is still lost; the gap is one call wide.*" |

**Item 1.** The turn-level snapshot is gone and `record::write` reads at write time, so the minutes-wide window this issue was filed about is closed. A save landing *after* the first `edit` is repaired by the second read. What remains is a save landing inside the read→write round trip: that body is replaced, the follow-up read finds the record present, no repair fires, and the person's text is gone with nothing said. Silent is exactly the word the DoD item uses, so the item is not satisfied — the module's own header says so in the same terms.

**Item 2.** The witness present is `a_write_that_lands_in_the_gap_keeps_its_text_and_gets_the_record_back`, and its fixture's `person_writes_after` fires **right after the first `edit`** — the post-write repair path, not the read→write gap. The test the item asks for would have to assert that the later text *survives* the gap, and no arrangement of this code makes it pass: recovering that text needs a compare-and-set the API does not offer, and after the write there is no observation that distinguishes "nobody edited" from "someone did and we overwrote them".

**So the decision is a scope one, not an engineering one**, and it belongs to a human:

1. Reword #5702's first two DoD items to what is achievable — the gap is narrowed to one round trip, documented, and repaired where repair is possible — and keep `Closes`; or
2. change this PR to `Refs #5702` and leave the issue open for the part that needs a detection mechanism (an `updated_at` precondition check is the only surrogate here, and it can report a lost edit, never prevent one).

Option 1 is the honest one if the residual gap is accepted; option 2 if it is not. I have not picked either — this PR's diff is a real improvement under both, and `dod / dod` stays red until one is chosen.

---

macanderson:
## Answering the Sourcery ❌

Row: *"Prevent escalation from silently overwriting issue-body edits made after
the loop's earlier read, including providing a witness test for concurrent
edits."*

The reading is correct on the facts and I am not disputing it. An edit saved
between this code's own read and its write is still lost. What I am disputing
is that this PR can close that half.

**The remaining window is a property of the tracker.** GitHub's issue update
takes no precondition — no ETag, no `If-Match`, no version field, on REST or
GraphQL. `gh issue edit --body` is a whole-body replace. So no amount of code
on this side of the wire makes a body write conditional. The only design that
removes the hazard is moving the record out of the body and into a comment the
loop owns, which needs a comment **read** on `IssueProvider` plus every
implementor, and costs a read per escalated issue per cycle. That trade was
weighed and declined in #5662 and again in this issue's own Options section.
Re-deciding it inside a bug fix would be the wrong place for it.

So the split is:

- **Landed here:** the window goes from one work turn — minutes, and the exact
  minutes a person is most likely to be annotating the issue — to one round
  trip. The stale snapshot is no longer representable: the `body` parameter is
  deleted from `escalate` and `escalate_blocking`, so there is nothing to pass.
- **Filed as #5902:** the remaining round trip, with the three options written
  out and a definition of done that asks for a recorded decision rather than
  assuming one.

**On the witness test.** `a_note_added_during_the_turn_survives_the_escalation`
covers exactly the defect reported: the tracker holds a note the loop never
saw, and the note is still there afterwards. `a_write_that_lands_in_the_gap_…`
covers the post-write half. Neither can cover an edit landing *inside* the
read-to-write gap and still assert it survives, because under this design it
does not survive — a test asserting otherwise would be asserting a fix that is
not here. #5902 is where that test belongs, beside whichever option is chosen.

macanderson:
## Answering Sourcery's ❌ row

Sourcery's "Assessment against linked issues" table marks one objective ❌:

> The witness test covers an edit after the first write rather than an edit
> specifically between the pre-write read and write.

That reading of the diff is correct, and it names a real, remaining gap —
`record::write`'s own module header says the same thing: "An edit saved
inside that gap is lost." Two things follow from that, and both are already
true on this branch rather than left for later:

**#5702's own definition of done does not ask for that gap to close.** Its
DoD box 1 is "An escalation write cannot silently drop a body edit made
*after the loop read that body*" — the loop's turn-start read, minutes
before the write. That is what this PR fixes: `write` now reads fresh
immediately before it writes, so nothing but the read-to-write round trip
itself is stale. The issue's own **Options** section named this directly
before any code was written: "Re-read the body immediately before the edit.
Narrows the window to one round trip rather than one turn. Cheap, and does
not close it." Option 1 is what's implemented, on the terms the issue
itself set.

**The residual round-trip gap is filed, not dropped: #5902.** It's a fresh
issue, opened against this exact window, with three options for closing it
(move the record to a comment, give the port a conditional write, or accept
and measure) and a DoD asking for a recorded decision plus, depending on
that decision, either a witness test proving the window is gone or one
proving it's counted. `docs/adr` and `doc:backlog-self-driving` are the
places that decision lands.

So: fix and follow-up both exist, on their own issues, and #5847 is scoped
to the one #5702 actually asked for. Resolving this thread.

Issue #5853, opened 2026-09-05
Nothing declares that an adapter must refuse a tool call it cannot dispatch


macanderson:
## The problem

Six provider adapters now refuse a tool call that arrives without the field it
cannot be dispatched without — the id the next turn's tool result is matched
against, or the name of the tool to run. `crates/stella-model/src/http.rs`'s
`malformed_tool_call_error` is the one constructor they all use, and each call
site has a wiremock witness at the provider edge:

- `anthropic/stream.rs`, `anthropic/unary.rs` (#5496, PR #5667)
- `zai/stream.rs`, `zai/unary.rs`
- `openai/stream.rs`, `openai/unary.rs`
- `bedrock.rs`
- `gemini.rs` (shared with `vertex.rs` through `GeminiAssembly::absorb`)

Nothing declares that rule. `crates/stella-model/src/provider_parity.rs` holds
six axes — cache, reasoning, stream fallback, overflow, output budget, parallel
tool calls — and none of them is about a malformed tool call. So a seventh
adapter, or a rewrite of one of these six, can drop the refusal and no test
notices. That is the failure mode invariant 8 (AGENTS.md, "Provider feature
parity is declared, not assumed") exists to prevent: OpenRouter ran Claude
models with zero prompt caching for months because nothing enforced the cache
axis.

## The question to settle

Whether this is a parity axis at all. Two readings, and a person should pick:

1. **It is an axis.** Each provider declares a posture naming the fields its
   wire can drop and the witness test proving the adapter refuses a call
   missing one. `stella-cli`'s config tests then fail if a seeded provider has
   no row, and `stella-model`'s parity tests fail if a row's witness test is
   gone — the enforcement from both sides every other axis already has.
2. **It is not.** The other six axes describe how a provider *differs*. This
   rule is the same on every dialect, so it may belong as a shared invariant
   test over the adapters rather than as a per-provider posture.

## Where to look

- `crates/stella-model/src/provider_parity.rs` — the six axes and the shape a
  seventh would take. `ParallelToolCallPosture` is the newest and the closest
  model, because it also records a fact about this tree's own adapters
  (`fan_in_witness`) rather than only about a vendor API.
- `crates/stella-model/src/http.rs` — `malformed_tool_call_error`, the one
  constructor every refusal goes through. `rg malformed_tool_call_error
  crates/stella-model/src` names every call site today.
- The witness modules: `anthropic/malformed_tool_use.rs`,
[...]

PR #5854, opened 2026-09-05
fix(stella-model): refuse a tool call whose id never arrives on every dialect

### LINKED ISSUE #5684: The other adapters still lose a tool call whose id never arrives
## The problem

#5496 fixed one shape of this on the Messages dialect and #5667 landed it. The
same shape is still live on the OpenAI Responses adapter, and the "commit a
call nothing can correlate" half is still live on three more paths.

**The OpenAI Responses stream drops the whole call.** In
`crates/stella-model/src/openai.rs`, the streamed `output_item.added` payload
declares its id without a default:

```rust
FunctionCall {
    call_id: String,
    #[serde(default)]
    name: String,
},
```

Serde matches an internally-tagged enum on its `type` first and then fails the
whole event if a field of the matched case is absent, so a frame that omits
`call_id` does not deserialize. `aggregate_openai_stream`'s tolerant arm — the
one that exists to skip an event type the adapter does not model — drops it,
and with it the accumulator every later `function_call_arguments.delta` keys
on. The turn then completes `Ok` reporting no tool calls. This is #5496's
failure verbatim, one dialect over.

**Three paths commit a call with an empty id.** Each one's *speculative*
announce already refuses such a call, and its final assembly does not — the
same asymmetry #5667 closed in `zai/stream.rs`:

- `crates/stella-model/src/openai/stream.rs` — the announce site guards
  `!acc.call_id.is_empty()`, and the `into_values().map(...)` that builds the
  committed `Vec<ToolCall>` does not.
- `crates/stella-model/src/openai/unary.rs` — `OpenAiUnaryItem::FunctionCall`
  defaults `call_id`, and `assemble_unary` commits it unchecked.
- `crates/stella-model/src/zai/unary.rs` — commits `call_id: call.id` with no
  check, while `zai/stream.rs` now refuses it.

Two calls in one turn both keyed `""` mis-pair with their results in
`stella-core`'s loop evidence, and a `tool_result` sent back under an empty id
matches nothing the provider is holding.

`crates/stella-model/src/bedrock.rs` commits `call_id: tool_use.tool_use_id`
with `tool_use_id` undefaulted, so its failure is the drop shape rather than
the empty-id shape; worth confirming which arm swallows it. `gemini.rs` mints
its own call ids and looks unaffected — confirm rather than assume.

## The fix

`crates/stella-model/src/http.rs` already has the shared constructor #5667
added:

```rust
pub(crate) fn malformed_tool_call_error(
    provider: &str,
    index: usize,
    name: &str,
    missing: &[&str],
) -> ProviderError
```

It returns `ProviderError::Malformed` — terminal, not retryable, naming the
block index, the tool and the fields that never arrived. Follow
`anthropic/stream.rs`'s and `anthropic/unary.rs`'s call sites verbatim so the
dialects cannot drift on the rule.

Default the undefaulted ids (`openai.rs`'s streamed `FunctionCall`, and
`bedrock.rs`'s `BedrockToolUse` if it is reached on deserialize) so the frame
parses and the loss is visible, then refuse an empty one at assembly.

## Verify

Wiremock witnesses at the provider boundary, one per path, following
`crates/stella-model/src/anthropic/malformed_tool_use.rs` and
`crates/stella-model/src/zai/malformed_tool_calls.rs`. The assertion has to be
about what a caller receives: on today's code nothing errors, so a test written
against the type alone would pass either way.

Both `openai/tests.rs` and `zai/tests.rs` are grandfathered god files closed to
growth, so declare the new test module from `openai.rs` / `zai.rs` and put the
file under `openai/` / `zai/`, the way
`crates/stella-model/src/zai/malformed_tool_calls.rs` does.

New prose in this crate has to clear `make prose` — no issue numbers in
comments, no "used to" / "no longer", and a new file is held to a reading grade
of 6.00, so keep the sentences short.

## Definition of done

- [x] An OpenAI Responses `output_item.added` naming `function_call` with no
      `call_id` produces a terminal error, never an `Ok` reporting no tool
      calls.
      Done in macanderson/stella#5854: `openai.rs`'s streamed
      `OpenAiOutputItem::FunctionCall::call_id` now carries
      `#[serde(default)]` so the frame parses, and `openai/stream.rs`'s
      commit loop returns `http::malformed_tool_call_error("OpenAI", index,
      &acc.name, &missing)` on an empty `call_id` or `name`.
- [x] No adapter commits a `ToolCall` with an empty `call_id`: `openai/stream.rs`,
      `openai/unary.rs` and `zai/unary.rs` all refuse it.
      All three refuse it in #5854 — `openai/stream.rs`'s
      `into_values().map(...)` became an indexed `for` loop that returns the
      shared error; `openai/unary.rs`'s `FunctionCall` arm checks `call_id`
      and `name` before pushing; `zai/unary.rs` checks `call.id` before
      building the `ToolCall`.
- [x] `bedrock.rs` and `gemini.rs` are checked, and either fixed the same way or
      shown in the pull request description to be unreachable.
      Both checked and both fixed. `bedrock.rs` had the undefaulted-id shape
      one level up (`BedrockToolUse::tool_use_id` / `name`): both now default
      and assembly refuses an empty one by name. `gemini.rs`'s call id really
      is unreachable — the wire carries none and `GeminiAssembly::absorb`
      mints one from the ordinal — but its `name` was undefaulted and had the
      same drop shape, so it defaults and `absorb` refuses an empty name.
      `vertex.rs` shares `aggregate_gemini_stream`, so that one refusal covers
      both providers.
- [x] Each fixed path has a wiremock witness that fails without the change.
      Five, all at the provider edge:
      `openai/malformed_tool_call.rs` —
      `a_function_call_with_no_id_ends_the_turn_instead_of_vanishing`,
      `a_unary_function_call_with_no_id_ends_the_turn_as_well`, and
      `an_output_item_type_this_adapter_does_not_model_is_still_skipped` as
      the constraint; `zai/malformed_tool_calls.rs` —
      `a_unary_tool_call_whose_id_never_arrives_ends_the_turn_too`;
      `bedrock/malformed_tool_use.rs` —
      `a_tool_use_block_with_no_id_names_the_call_it_lost` plus its
      constraint; `gemini/malformed_tool_call.rs` —
      `a_function_call_with_no_name_ends_the_turn_instead_of_vanishing` plus
      its constraint.
- [x] CI green on the pull request.
      Every check on #5854's head `b36ca04` has landed and none failed.
      `fmt + clippy + test` succeeded at 2026-09-05T06:17:44Z (run
      33947314518), and `cargo deny + cargo audit`, `cargo check on the
      declared MSRV`, `docs guards`, `guard self-tests`, `gate steps no other
      workflow runs`, `post-merge canary self-tests`, `file size ratchet`,
      `duplicate claims`, `no edit is undone inside the branch`, `is this diff
      prose-only?`, `main is not known-broken`, `dependency-review`, `PR
      carries a diff` and `harbor_adapter + analyzer pytest` are all
      `success`. `cla` and `Sourcery review` are `skipped` (Sourcery's own
      diff budget, not a verdict on this branch). The only red check was
      `dod / dod` itself, reporting this box — the ordering trap tracked as
      oxagen#2638, where the last DoD box can only be true after the event
      that reads it.

macanderson:
## What and why

#5667 closed "a tool call whose id never arrives" on the Messages dialect. The
same shape was still live on four more paths, and each one loses a call while
the turn still ends `Ok`.

**The OpenAI Responses stream dropped the whole call.** `openai.rs`'s streamed
`OpenAiOutputItem::FunctionCall` declared `call_id: String` with no default.
Serde matches an internally-tagged enum on its `type` first and then fails the
whole event when a field of the matched case is absent, so a frame that omits
`call_id` never deserialized at all. `aggregate_openai_stream`'s tolerant arm —
the one that exists to skip an event type the adapter does not model — dropped
it, and with it the accumulator every later `function_call_arguments.delta`
keys on.

**Three paths committed a call with an empty id** while their speculative
announce already refused one — the asymmetry #5667 closed in `zai/stream.rs`:
`openai/stream.rs`, `openai/unary.rs` and `zai/unary.rs`.

**`bedrock.rs` had the undefaulted-id shape one level up.** `BedrockToolUse`
declared `toolUseId` and `name` with no default, inside
`Option<BedrockToolUse>`, so an absent id failed the whole `Converse` body and
the turn died on `ProviderError::Malformed("missing field \`toolUseId\`")` —
naming neither the tool nor the block, and indistinguishable from a truncated
body or a changed schema.

**`gemini.rs`: the id really is unreachable, and `name` is not.** This wire
carries no call id; `GeminiAssembly::absorb` mints one from the call's ordinal,
so the empty-`call_id` shape cannot occur. Its `name` was undefaulted, which is
the same hazard on the same field set: a `functionCall` part with no `name`
failed the whole chunk, and the arm that tolerates a keep-alive frame skipped
it, taking any `finishReason` riding beside it. `vertex.rs` calls
`aggregate_gemini_stream` and shares `GeminiAssembly`, so the one refusal
covers both providers and both delivery paths.

## The fix

Every undefaulted id (and `name`) now carries `#[serde(default)]`, so the frame
parses and the loss becomes visible. Assembly then refuses an empty one with
`http::malformed_tool_call_error` — `ProviderError::Malformed`, terminal and
not retryable, naming the block index, the tool and the fields that never
arrived. The call sites follow `anthropic/stream.rs` and `anthropic/unary.rs`
verbatim so the dialects cannot drift on the rule.

`openai/stream.rs`'s announce site now guards `name` alongside `call_id`, so
what it speculates on and what it commits agree.

## Exemplar

`crates/stella-model/src/anthropic/stream.rs` and `anthropic/unary.rs` (#5667)
are the in-tree exemplar the issue names; the shared constructor in
`crates/stella-model/src/http.rs` is used unchanged.

## Witness mechanism

One wiremock witness per fixed path, at the provider edge rather than on the
type — on today's code nothing errors, so a test written against the type alone
would pass either way. Each asserts on what a caller receives.

- `crates/stella-model/src/openai/malformed_tool_call.rs`
  - `a_function_call_with_no_id_ends_the_turn_instead_of_vanishing` — an
    `output_item.added` naming `function_call` with no `call_id`, its arguments
    streamed and closed, ending on `response.completed`. Without the change the
    item frame fails to parse and is skipped, the argument delta opens an
    accumulator keyed `""`, and `complete` returns `Ok`. The `expect_err` is
    what fails.
  - `a_unary_function_call_with_no_id_ends_the_turn_as_well` — an empty stream
    arms the fallback latch, then the unary body carries the same item. Without
    the change the fields already defaulted, so the turn committed a call keyed
    `""` and returned `Ok`.
  - `an_output_item_type_this_adapter_does_not_model_is_still_skipped` — the
    constraint on the fix: an unmodelled item type, an unmodelled event type
    and a data line with no `type` all stay skippable.
- `crates/stella-model/src/zai/malformed_tool_calls.rs`
  - `a_unary_tool_call_whose_id_never_arrives_ends_the_turn_too` — same latch
    shape, unary `chat.completion` with a `tool_calls` entry carrying no `id`.
    Without the change `assemble_unary` committed it and returned `Ok`.
- `crates/stella-model/src/bedrock/malformed_tool_use.rs`
  - `a_tool_use_block_with_no_id_names_the_call_it_lost` — Converse answers
    once, so the old failure was serde's own `missing field toolUseId`. The
    assertions that fail without the change are the ones about `index 0` and
    `bash`; the error now names the block, the tool and the field.
  - `a_tool_use_block_with_an_id_still_completes` — the constraint.
- `crates/stella-model/src/gemini/malformed_tool_call.rs`
  - `a_function_call_with_no_name_ends_the_turn_instead_of_vanishing` — without
    the change the chunk carrying the part fails to parse and is skipped, the
    next chunk sets `STOP`, and the turn ends `Ok` with no text and no tool
    calls.
  - `a_function_call_with_a_name_still_completes` — the constraint, including
    the minted `call_0` id.

Both `openai/tests.rs` and `zai/tests.rs` are grandfathered or crowded, so the
new modules are declared from `openai.rs` / `bedrock.rs` / `gemini.rs` and live
under the sibling folder, the way `zai/malformed_tool_calls.rs` already does.

## Provider parity matrix

Not updated: `crates/stella-model/src/provider_parity.rs` declares six axes
(cache, reasoning, stream fallback, overflow, output budget, parallel tool
calls) and none of them is about a malformed tool call, so there is no row to
change. #5667 added none either. Whether this deserves a seventh axis is a
separate judgement — filed as #5853 rather than decided here.

## The definition of done

#5684's checklist is fully ticked, each item carrying the evidence that
establishes it. The first four were verified against this diff and recorded on
the issue earlier.

The fifth — "CI green on the pull request" — is now settled on head `b36ca04`:
`fmt + clippy + test`, the job that compiles the workspace and runs the five
new wiremock witnesses, finished `success` at 06:17:44Z
([run 33947314518](https://github.com/macanderson/stella/actions/runs/33947314518)),
alongside every other check on this head — `cargo deny + cargo audit`, the MSRV
check, `guard self-tests`, `docs guards`, `gate steps no other workflow runs`,
`file size ratchet`, `post-merge canary self-tests`, `dependency-review`,
`harbor_adapter + analyzer pytest`, `duplicate claims`, `no edit is undone
inside the branch`, `PR carries a diff`, `is this diff prose-only?` and `main
is not known-broken`. None failed or timed out.

The one check that was red is `dod / dod`, and it was red *because that box was
unticked* — the gate reads the checklist it is enforcing, so the item and the
check are circular. It settles on this run.

Tests deleted: None.

Residue issues filed: #5853 (nothing declares that an adapter must refuse a tool call it cannot dispatch).

Closes #5684

## Summary by Sourcery

Reject tool calls that lack the identifiers or names required for dispatch across all supported provider dialects and delivery paths.

Bug Fixes:
- Reject malformed tool calls with missing dispatch identifiers instead of silently committing them or dropping them while returning a successful turn.
- Return terminal, provider-specific malformed-call errors that identify the affected block or call, tool name, and missing fields across OpenAI, Zai, Bedrock, and Gemini paths.

Enhancements:
- Keep tool-call announcement and commitment validation consistent across streaming and unary delivery paths.
- Preserve tolerance for genuinely unmodelled OpenAI events while surfacing malformed calls for supported event types.

Tests:
- Add provider-edge WireMock coverage for missing OpenAI call IDs, missing Zai IDs, missing Bedrock tool-use IDs or names, and missing Gemini function names, including successful valid-call constraints.

### DIFF
--- a/crates/stella-model/src/bedrock.rs
+++ b/crates/stella-model/src/bedrock.rs
@@ -403,7 +403,13 @@ fn unsupported_format_note(kind: &str, media_type: &str) -> BedrockContentBlock 
 #[derive(Serialize, Deserialize, Debug, Clone)]
 #[serde(rename_all = "camelCase")]
 struct BedrockToolUse {
+    /// Defaulted so a block that leaves it out still parses. Without the
+    /// default, one absent id fails the whole `Converse` body and the turn
+    /// dies on a serde message that names neither the tool nor the block.
+    /// Assembly refuses an empty id by name instead.
+    #[serde(default)]
     tool_use_id: String,
+    #[serde(default)]
     name: String,
     #[serde(default)]
     input: Value,
@@ -844,11 +850,32 @@ impl Provider for BedrockProvider {
         let mut text = String::new();
         let mut tool_calls: Vec<ToolCall> = Vec::new();
         if let Some(message) = parsed.output.and_then(|o| o.message) {
-            for block in message.content {
+            for (index, block) in message.content.into_iter().enumerate() {
                 if let Some(t) = block.text {
                     text.push_str(&t);
                 }
                 if let Some(tool_use) = block.tool_use {
+                    // The id is what the next turn's `toolResult` is matched
+                    // against, and the name is the tool to run. Neither can be
+                    // invented here. Two calls in one turn both keyed `""`
+                    // pair up with the wrong results in `stella-core`'s loop
+                    // evidence, and a result sent under an empty id matches
+                    // nothing Converse holds.
+                    let mut missing = Vec::new();
+                    if tool_use.tool_use_id.is_empty() {
+                        missing.push("toolUseId");
+                    }
+                    if tool_use.name.is_empty() {
+                        missing.push("name");
+                    }
+                    if !missing.is_empty() {
+                        return Err(http::malformed_tool_call_error(
+                            "Bedrock",
+                            index,
+                            &tool_use.name,
+                            &missing,
+                        ));
+                    }
                     // A no-argument Converse tool call omits `input` on the
                     // wire, so the `#[serde(default)]` field deserializes to
                     // `Value::Null` — which is the malformed-call sentinel
@@ -1292,5 +1319,7 @@ pub(crate) mod sigv4 {
     }
 }
 
+#[cfg(test)]
+mod malformed_tool_use;
 #[cfg(test)]
 mod tests;
--- a/crates/stella-model/src/bedrock/malformed_tool_use.rs
+++ b/crates/stella-model/src/bedrock/malformed_tool_use.rs
@@ -1,0 +1,108 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+//! What a `toolUse` block does to a turn when its id is missing.
+//!
+//! Converse answers once, so this adapter has one delivery path and one
+//! parse. A block with no `toolUseId` fails that parse outright unless the
+//! field defaults. The whole body goes with it, and the turn dies on a serde
+//! message that names neither the tool nor the block it sat in. A reader
+//! cannot tell that failure from a truncated body or a changed schema.
+//!
+//! The field now defaults, so the block parses and assembly refuses it by
+//! name. The error says which block, which tool, and which field.
+//!
+//! Declared from `bedrock.rs`, so the suite in `bedrock/tests.rs` keeps its
+//! own subject.
+
+use super::*;
+use stella_protocol::CompletionRequest;
+use wiremock::matchers::method;
+use wiremock::{Mock, MockServer, ResponseTemplate};
+
+fn request() -> CompletionRequest {
+    CompletionRequest {
+        messages: vec![CompletionMessage::user("list the files")],
+        max_output_tokens: Some(1024),
+        temperature: None,
+        effort: None,
+        tools: vec![],
+        reasoning: None,
+        params: None,
+    }
+}
+
+fn test_provider(server_uri: &str) -> BedrockProvider {
+    BedrockProvider::new(
+        ApiKey::new("AKIDEXAMPLE"),
+        ApiKey::new("wJalrXUtnFEMI/K7MDENG+bPxRfiCYEXAMPLEKEY"),
+        None,
+        "us-east-1",
+        "us.anthropic.claude-sonnet-4-5-20250929-v1:0",
+    )
+    .with_base_url(server_uri.to_string())
+}
+
+/// The witness. Without the fix the error is serde's own `missing field
+/// toolUseId`, which names no tool and no block, so the assertions about
+/// `index 0` and `bash` are the ones that fail.
+#[tokio::test]
+async fn a_tool_use_block_with_no_id_names_the_call_it_lost() {
+    let server = MockServer::start().await;
+    Mock::given(method("POST"))
+        .respond_with(ResponseTemplate::new(200).set_body_json(serde_json::json!({
+            "output": {"message": {"role": "assistant", "content": [
+                {"toolUse": {"name": "bash", "input": {"command": "ls"}}}
+            ]}},
+            "stopReason": "tool_use",
+            "usage": {"inputTokens": 9, "outputTokens": 5}
+        })))
+        .mount(&server)
+        .await;
+    let provider = test_provider(&server.uri());
+
+    let error = provider
+        .complete(request())
+        .await
+        .expect_err("a call that cannot be dispatched must not be committed");
+
+    let message = error.to_string();
+    assert!(
+        message.contains("index 0"),
+        "the error names the block that was lost: {message}"
+    );
+    assert!(
+        message.contains("`toolUseId`"),
+        "the error names the field that never arrived: {message}"
+    );
+    assert!(
+        message.contains("bash"),
+        "the error names the tool the model asked for: {message}"
+    );
+    assert!(
+        !error.is_retryable(),
+        "the same request returns the same absent field: {error:?}"
+    );
+}
+
+/// The constraint on the fix: an ordinary call still rides through. Only a
+/// block that cannot be dispatched is fatal.
+#[tokio::test]
+async fn a_tool_use_block_with_an_id_still_completes() {
+    let server = MockServer::start().await;
+    Mock::given(method("POST"))
+        .respond_with(ResponseTemplate::new(200).set_body_json(serde_json::json!({
+            "output": {"message": {"role": "assistant", "content": [
+                {"toolUse": {"toolUseId": "tooluse_1", "name": "bash", "input": {"command": "ls"}}}
+            ]}},
+            "stopReason": "tool_use",
+            "usage": {"inputTokens": 9, "outputTokens": 5}
+        })))
+        .mount(&server)
+        .await;
+    let provider = test_provider(&server.uri());
+
+    let result = provider.complete(request()).await.expect("a whole call");
+    assert_eq!(result.tool_calls.len(), 1);
+    assert_eq!(result.tool_calls[0].call_id, "tooluse_1");
+    assert_eq!(result.tool_calls[0].name, "bash");
+    assert_eq!(result.finish_reason, Some(FinishReason::ToolCalls));
+}
--- a/crates/stella-model/src/gemini.rs
+++ b/crates/stella-model/src/gemini.rs
@@ -246,6 +246,12 @@ fn attachment_parts(message: &CompletionMessage) -> Vec<GeminiOutboundPart> {
 
 #[derive(Serialize, Deserialize, Debug, Clone)]
 pub(crate) struct GeminiFunctionCall {
+    /// Defaulted so a part that leaves it out still parses. Without the
+    /// default the whole chunk fails to deserialize, and the arm that
+    /// tolerates a keep-alive frame drops it. The call goes with the frame,
+    /// and so does any `finishReason` riding beside it. `GeminiAssembly`'s
+    /// `absorb` refuses an empty name by name instead.
+    #[serde(default)]
     pub(crate) name: String,
     #[serde(default)]
     pub(crate) args: Value,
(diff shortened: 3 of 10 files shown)


macanderson:
## DoD verification for #5684 — 4 of 5 items satisfied, 1 not yet

Checked out `origin/main` (`7c5ca8e`) against `origin/fix/5684-tool-call-missing-id` (`b36ca04`) and read the diff directly. Per-item verdict:

**1. "An OpenAI Responses `output_item.added` naming `function_call` with no `call_id` produces a terminal error, never an `Ok` reporting no tool calls."**
✅ SATISFIED — `crates/stella-model/src/openai.rs`: `OpenAiOutputItem::FunctionCall.call_id` now carries `#[serde(default)]`, so the frame parses instead of being silently skipped. `crates/stella-model/src/openai/stream.rs`'s final-assembly loop (replacing the old `.into_values().map(...).collect()`) now returns `Err(http::malformed_tool_call_error("OpenAI", index, &acc.name, &missing).into())` when `call_id` (or `name`) is empty. Witness test `crates/stella-model/src/openai/malformed_tool_call.rs::a_function_call_with_no_id_ends_the_turn_instead_of_vanishing` sends exactly this frame and asserts `expect_err`, checking the message names `index 0`, `` `call_id` ``, and `bash`, and that the error is not retryable.

**2. "No adapter commits a `ToolCall` with an empty `call_id`: `openai/stream.rs`, `openai/unary.rs` and `zai/unary.rs` all refuse it."**
✅ SATISFIED — all three refuse it now:
- `openai/stream.rs`: shown above.
- `openai/unary.rs`: the `FunctionCall` arm now checks `call_id.is_empty() || name.is_empty()` before pushing, returning `malformed_tool_call_error` instead. Witness: `a_unary_function_call_with_no_id_ends_the_turn_as_well`.
- `zai/unary.rs`: the per-call loop now checks `call.id.is_empty()` and returns `malformed_tool_call_error(label, index, &call.function.name, &["id"])` before building `ToolCall`. Witness: `zai/malformed_tool_calls.rs::a_unary_tool_call_whose_id_never_arrives_ends_the_turn_too`.

**3. "`bedrock.rs` and `gemini.rs` are checked, and either fixed the same way or shown in the pull request description to be unreachable."**
✅ SATISFIED — both fixed:
- `bedrock.rs`: `BedrockToolUse.tool_use_id` and `.name` both gained `#[serde(default)]`; the `Converse` content loop now checks both and returns `http::malformed_tool_call_error("Bedrock", index, &tool_use.name, &missing)`. Witness: `bedrock/malformed_tool_use.rs::a_tool_use_block_with_no_id_names_the_call_it_lost` — its own doc comment and assertions establish that on old code the failure was serde's own `missing field toolUseId`, naming neither tool nor block.
- `gemini.rs`: `GeminiFunctionCall.name` gained `#[serde(default)]`; `GeminiAssembly::absorb` now refuses an empty `name` via `malformed_tool_call_error`. The call id is confirmed unreachable-empty (minted from `ordinal`/`format!("call_{ordinal}...")`, never read off the wire), matching the issue's "confirm rather than assume." Witness: `gemini/malformed_tool_call.rs::a_function_call_with_no_name_ends_the_turn_instead_of_vanishing`.

**4. "Each fixed path has a wiremock witness that fails without the change."**
✅ SATISFIED — read all four new test files in full (`openai/malformed_tool_call.rs`, `zai/malformed_tool_calls.rs`, `bedrock/malformed_tool_use.rs`, `gemini/malformed_tool_call.rs`). Each is wiremock-based, asserts at the provider boundary (`provider.complete(request()).await.expect_err(...)`), checks specific error content (block/index, missing field name, tool name, non-retryable), and each includes a "constraint" test proving a well-formed call still completes — the discipline the issue asked for ("the assertion has to be about what a caller receives").

**5. "CI green on the pull request."**
❌ NOT SATISFIED (yet) — `get_check_runs` on the PR head (`b36ca04`) shows the `ci` workflow run (33947314518) still `status: "queued"` as of this check, with its `fmt + clippy + test` and `cargo deny + cargo audit` jobs both `queued` — not `success`/`skipped`/`failure`, just not yet run. `dod / dod` is correctly red only because of the unticked boxes, per this task's design, but the other required jobs simply haven't executed yet, so "CI green" can't be called true right now.

**What would satisfy item 5:** let the `ci` workflow run to completion and confirm `fmt + clippy + test` and `cargo deny + cargo audit` (and every other required check) land on `success` (or a legitimate `skipped`). Once that's true, items 1–4 above already have their evidence on record in this comment, so re-verification should only need to re-check CI status.

No DoD boxes were ticked on #5684 — only item 5 is blocking, and it's a CI-completion question, not a code-correctness one.

---

Issue #5852, opened 2026-09-05
stella context amend deletes the comments in the block it rewrites


macanderson:
## Problem

`stella context amend <handle> --tasks ... --keywords ... --paths ...` rewrites a record's `[record.steering.applies_to]` block and deletes every comment inside it. The comments are the only place a record says *why* its scope is what it is, and they are exactly what the next amend erases.

Seen while fixing #5779. Before:

```toml
[record.steering.applies_to]
# Retargeted from the extractor's ["pull-request", "delete-test"], neither
# of which is a ratified task name.
tasks = ["test", "review"]
```

After `stella context amend deleted-test-named-in-pr --tasks review ...`:

```toml
[record.steering.applies_to]
tasks = ["review"]
```

The comment is gone, with nothing said about it. The same amend dropped a three-line comment from `ctx.macanderson.stella.inv8-posture-every-axis.toml` explaining why that record is scoped by path and keyword rather than by task. Both had to be typed back by hand afterwards.

## Why it matters

`AGENTS.md` says these files are edited through the CLI and never by hand, because the record hash and the promotions ledger depend on it. A command that is the only sanctioned editor must not destroy content it was not asked to change. The loss is silent: the command prints a `was` / `now` line for the fields, and says nothing about the comments it removed.

A comment is not part of the hash preimage — `crates/stella-core/src/context_record/hash.rs` hashes the parsed record — so preserving comments costs the identity nothing.

## Where it lives

`crates/stella-cli/src/context_cmd.rs` (and the `context_records.rs` writer beside it) hold the amend path. Find where the amended record is serialized back to TOML; the rewrite is what drops the comments.

## The fix

Rewrite only the values, keeping the surrounding trivia — a format-preserving TOML edit (`toml_edit` is the crate that does this, and it is already how a lot of Rust tooling edits a manifest in place) rather than serialize-and-overwrite. If the dependency is not wanted, the second-best answer is to refuse the amend when the block carries comments and tell the author to move them, which is at least not silent.

## Verify

```
cp .stella/rules/ctx.macanderson.stella.inv8-posture-every-axis.toml /tmp/before.toml
STELLA_TRUST_PROJECT=1 stella context amend inv8-posture-every-axis --keywords posture --keywords axis
diff /tmp/before.toml .stella/rules/ctx.macanderson.stella.inv8-posture-every-axis.toml
```

The diff shows the comment lines removed alongside the keyword change.

## Definition of done

[...]

Issue #5851, opened 2026-09-05
Every hard record this repo publishes is imported, so none of them can ever block


macanderson:
## Problem

Every record this repository publishes with `mode = "hard"` carries `origin = "imported"`, and an imported record may never block a tool call. `crates/stella-core/src/records/bridge.rs`'s `may_block` refuses `Imported`, `Inferred` and `Observed`, so each of these records loads with a `BlockingRefused(UntrustedOrigin)` finding and its guard never arms.

`stella context validate` prints one warning per record:

```
warning  ^deleted-test-named-in-pr
  asks to block but its origin is imported — an extracted or inferred claim can
  advise, never deny; re-author it under your own origin to enforce it
```

Eleven records say that today: `deleted-test-named-in-pr`, `inv1-core-no-sdk-imports`, `inv3-allowlist-same-pr`, `inv3-zero-telemetry`, `inv4-serde-roundtrip`, `inv5-typed-errors-no-panic`, `inv6-budget-safe-boundaries`, `inv7-byte-stable-prompts`, `inv8-matrix-same-pr`, `inv8-posture-every-axis`, `left-behind-fix-is-file-issue`.

So `mode = "hard"` on all eleven is a claim the engine cannot honour. A reader of `.stella/rules/` sees eleven blocking guards; the engine has none. This is separate from the trigger collision in #5779, which suspended two of them for a second, independent reason — that one is fixed, and these records still do not block.

## Why it matters

The mode field is the only thing that says whether a record advises or denies. When every `hard` record in the tree is inert, the field stops carrying information, and the next person who writes a real blocking guard has no way to tell theirs apart from the ten that do nothing.

## The two ways out, and the decision this issue needs

1. **Re-author them.** The refusal message names the remedy: publish these under an origin that may block. That is a human act — someone has to own each statement as policy they wrote rather than as text an extractor lifted out of `AGENTS.md`. `stella context keep` / `promote` is the path; `.stella/rules/promotions.jsonl` is the ledger that records the approval.
2. **Drop them to `soft`.** If nobody is prepared to own them as deny-shaped policy, the honest field value is `soft`, and the file then says what the engine does.

Either answer is fine; the current state — a field that says one thing while the engine does another — is not.

## Verify

```
STELLA_TRUST_PROJECT=1 stella context validate
```

[...]

PR #5855, opened 2026-09-05
fix(rules): give each witness-test record its own trigger

### LINKED ISSUE #5779: Four repo records collide on the 'witness test' trigger, and two hard guards are suspended by it

macanderson:
## What & why

Four of this repository's published records met on the keyword `witness test` at precedence 100, in pairs that mixed `hard` enforcement with `soft`. `crates/stella-core/src/records/validate.rs` reports each such pair and suspends the stricter side, so the `hard` guards on `deleted-test-named-in-pr` and `inv8-posture-every-axis` never armed. `stella context validate` said so in a warning and exited 0, so nothing failed while both guards were down.

Two records reached the collision through triggers wider than their own subject, and those are the two this PR narrows:

- **`deleted-test-named-in-pr`** carried `tasks = ["test", "review"]`. The matcher compares a task against whole words of the turn text (`select.rs`'s `word_in_text`), so `test` matches inside the phrase `witness test` that the other three records carry as a keyword. It is now `tasks = ["review"]` plus the deletion words this rule is actually about (`deleted test`, `deleting a test`, `delete a test`, `test deletion`, `removed test`).
- **`inv8-posture-every-axis`** carried the literal keyword `witness test`. Its subject is provider parity, and its path glob plus `posture` / `axis` already select that work; the third keyword is now `provider parity`.

The two `soft` records — `correctness-needs-proof` and `witness-definition` — keep `witness test`. Two records of the same enforcement never conflict, and both are genuinely about the witness contract.

Both files were changed through `stella context amend`, never by hand, so each carries a freshly stamped `record_id` / `record_hash`. The explanatory comments were typed back afterwards; a comment is not part of the hash preimage (`context_record/hash.rs` hashes the parsed record), and `stella context validate` reports no hash mismatch. The comment loss is a defect in `amend` itself, filed as #5852.

Closes #5779

## The witness

`crates/stella-core/tests/published_records_do_not_collide.rs` loads this repository's own `.stella/rules/*.toml` through `load_context_file` + `assign_handles`, runs `detect_conflicts`, and fails on any conflict — and, in a second test, on any `hard` record `is_suspended` holds down. Both fail on `main` and pass here, by construction:

- On `main`, `shared_triggers` returns `witness test` for all four pairs — from `deleted-test-named-in-pr`'s task `test` matching that word inside the phrase, and from `inv8-posture-every-axis`'s literal keyword. `detect_conflicts` therefore returns four `Conflict`s, `no_two_published_records_collide` fails on a non-empty list, and `every_published_hard_guard_arms` fails naming both suspended handles.
- After the amend no pair shares a task or keyword, `detect_conflicts` returns nothing, and both tests pass.

I could not run the current detector locally: the installed `stella` is 0.9.300, which predates #5330's trigger comparison. What I did run is a line-for-line Python mirror of `find_conflicts` / `overlapping_scope` / `shared_triggers` over all 39 published records — 4 conflicts before, matching the table in #5779 exactly, and 0 after. CI's `cargo test --workspace` is the real evidence, and this test is in it. `.stella/rules/**` is outside `scripts/ci-rust-scope.sh`'s prose paths, so a future record-only PR still starts the Rust gate and still runs this test.

The exemplar for the test's shape is `crates/stella-core/tests/context_record_examples.rs`, which reads repository files the same way and asserts the loader agrees with them.

## The gate

- [x] `cargo fmt --all` (ran locally), `make guards-fast` green, `python3 scripts/check-prose.py` green
- [ ] `cargo clippy` / `cargo test --workspace` — CI runs these; per CLAUDE.md this laptop does not
- [x] Docs updated where needed: nothing under `docs/` or `website/` names these four records or their scoping (checked with `rg`), so there is nothing to keep in step
- [x] `Closes #5779` appears here and as a commit trailer

Tests deleted: None.

## Nothing left behind

- [x] Filed: #5851 (every `hard` record here is `origin = "imported"`, so none of them clears the blocking gate — a second, independent reason these guards do not block, which this PR does not address), #5852 (`stella context amend` deletes the comments in the block it rewrites)

## Ground-rule check

- [x] No I/O added to `stella-core`'s library code; the file reading is in an integration test, as in `context_record_examples.rs`
- [x] No new dependencies

Resolving the collision un-suspends both guards; it does not make them block, because both records are `origin = "imported"` and `bridge.rs`'s `may_block` refuses that origin. That is #5851, and it is a policy decision about who owns these statements rather than a code fix.

## Summary by Sourcery

Prevent published rule trigger collisions so hard guards remain armed and enforce repository policies.

Bug Fixes:
- Narrow published rule triggers to eliminate equal-precedence conflicts that suspended two hard guards.
- Add coverage preventing published records from colliding and ensuring every hard guard remains active.

Enhancements:
- Broaden deletion-related matching with phrases specific to test removal while retaining review-based applicability.
- Retarget provider-parity matching from the generic witness-test phrase to the provider parity subject.

Tests:
- Add integration tests that load all published repository records and fail on trigger conflicts or suspended hard guards.

Chores:
- Refresh record identifiers and hashes for the amended rule records.

### DIFF
--- a/.stella/rules/ctx.macanderson.stella.deleted-test-named-in-pr.toml
+++ b/.stella/rules/ctx.macanderson.stella.deleted-test-named-in-pr.toml
@@ -3,6 +3,8 @@ set_id = "macanderson.stella"
 
 [[record]]
 lineage_id = "ctx.macanderson.stella.deleted-test-named-in-pr"
+record_id = "rec_macanderson_stella_deleted_test_named_in_pr_3db5b98f8e84"
+record_hash = "sha256:a769e9153665fcecdeb65ece470da12fe8fcf2691dde93eb48a42b7b5ebe9559"
 kind = "rule"
 statement = "Deleting a test is permitted but must be named in the PR description so an invisible deletion becomes a sentence a reviewer reads."
 origin = "imported"
@@ -24,11 +26,19 @@ force = "must"
 precedence = 100
 
 [record.steering.applies_to]
-# Retargeted from the extractor's ["pull-request", "delete-test"], neither
-# of which is a ratified task name.
-tasks = ["test", "review"]
+# The extractor proposed tasks = ["pull-request", "delete-test"], neither of
+# which is a ratified task name. The task "test" stood in for them and was far
+# wider than this rule: it selected every turn that used the word, including
+# the phrase "witness test" that three other records carry, so this `hard`
+# record met two `soft` ones at equal precedence and its guard stayed
+# suspended. The words below name the deletion this rule is about.
+tasks = ["review"]
 keywords = [
     "deleted test",
+    "deleting a test",
+    "delete a test",
+    "test deletion",
+    "removed test",
     "PR description",
 ]
 
--- a/.stella/rules/ctx.macanderson.stella.inv8-posture-every-axis.toml
+++ b/.stella/rules/ctx.macanderson.stella.inv8-posture-every-axis.toml
@@ -3,6 +3,8 @@ set_id = "macanderson.stella"
 
 [[record]]
 lineage_id = "ctx.macanderson.stella.inv8-posture-every-axis"
+record_id = "rec_macanderson_stella_inv8_posture_every_axis_cc7ff80065b6"
+record_hash = "sha256:ab8f9e40bd4d26f0f3c2e9f9391bb43c52a8536d6bb2d8005ba6a4002d1613bd"
 kind = "rule"
 statement = "Each provider id declares a posture on every parity axis and, for a controllable/opt-in/implicit posture, names the witness test proving it on the wire."
 origin = "imported"
@@ -26,11 +28,14 @@ precedence = 100
 [record.steering.applies_to]
 paths = ["crates/stella-model/src/provider_parity.rs"]
 # The extractor proposed tasks = ["add-provider"]; no such task is in the ratified
-# applies_to.tasks vocabulary, so it matched nothing. Paths and keywords
-# carry the selection instead.
+# applies_to.tasks vocabulary, so it matched nothing. The path and the parity
+# words carry the selection. The keyword "witness test" is the whole subject of
+# two `soft` records, and sharing it at equal precedence held this `hard` guard
+# suspended, so the trigger names parity instead. The statement still names the
+# witness test it asks a provider row to cite.
 keywords = [
     "posture",
-    "witness test",
+    "provider parity",
     "axis",
 ]
 
--- a/crates/stella-core/tests/published_records_do_not_collide.rs
+++ b/crates/stella-core/tests/published_records_do_not_collide.rs
@@ -1,0 +1,108 @@
+//! No two records in `.stella/rules/` may clash.
+//!
+//! Those files steer every session run here. Two of them can share a
+//! trigger, sit at one `precedence`, and ask for different `enforcement`.
+//! The stricter one is then held back. It still loads, and its guard never
+//! arms.
+//!
+//! `stella context validate` says so in a warning and exits 0. So the build
+//! stays green while a `hard` guard is down. This test is the gate that
+//! warning is not.
+//!
+//! To clear a failure, run `stella context amend <handle> --keywords ...`.
+//! It narrows the trigger and stamps a new hash. A hand edit leaves the old
+//! hash in place.
+
+use std::path::PathBuf;
+
+use stella_core::ingest::record::EnforcementMode;
+use stella_core::records::{
+    LoadedRecord, assign_handles, detect_conflicts, is_suspended, load_context_file,
+};
+
+/// The schema tag a record carries. `governance.toml` sits in the same
+/// folder and has none, which is how it gets skipped.
+const RECORD_SCHEMA: &str = "context-record/v0.1";
+
+fn rules_dir() -> PathBuf {
+    PathBuf::from(env!("CARGO_MANIFEST_DIR")).join("../../.stella/rules")
+}
+
+/// Every record this repo ships, loaded the way the CLI loads them.
+fn published_records() -> Vec<LoadedRecord> {
+    let dir = rules_dir();
+    let mut files: Vec<PathBuf> = std::fs::read_dir(&dir)
+        .unwrap_or_else(|err| panic!("{} must be readable: {err}", dir.display()))
+        .map(|entry| entry.expect("a readable directory entry").path())
+        .filter(|path| path.extension().is_some_and(|ext| ext == "toml"))
+        .collect();
+    files.sort();
+
+    let mut records = Vec::new();
+    for path in files {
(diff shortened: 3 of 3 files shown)


macanderson:
## DoD verification against the diff — six of eight items hold, two do not

Automated PR sweep. `dod / dod` is red because all eight of #5779's boxes are unticked. I checked each against the head commit `205a1b1` rather than against the PR description, and the result splits cleanly.

**Verified satisfied by the diff — these can be ticked:**

1. *An owner decides the resolution for each of the four pairs.* Narrowing is the resolution, and it covers all four pairs: both `hard` records leave the `witness test` trigger set, so no pair of differing enforcement remains.
2. *The change goes through `stella context amend` / `keep` / `promote`, so the record hashes and the promotions ledger stay valid.* Both files gain a stamped `record_id` and `record_hash`. `.stella/rules/promotions.jsonl` is untouched, and that is correct rather than a gap — the ledger records enforcement grants and lifecycle events (its two entries are a `advisory→blocking` grant and a retirement). A trigger narrowing is neither, so it has nothing to append.
4. *If a shared trigger were accepted, write the decision down.* Vacuous — the resolution is to narrow, not to accept.
6. *A rules change, not a Rust source change; if code is touched it lands in the right crate and breaches no `file-size` ceiling.* The one Rust file is a new 108-line integration test at `crates/stella-core/tests/published_records_do_not_collide.rs` — the crate AGENTS.md's routing table names for record logic, a new file well under 1500 lines, and `file size ratchet` is green on this head.
7. *Update any doc naming these records' precedence or scoping.* Confirmed independently: `deleted-test-named-in-pr`, `inv8-posture-every-axis`, `correctness-needs-proof` and `witness-definition` appear in no file outside `.stella/rules/*.toml`, `docs/spec/adaptive-context/context-pr.md` included.
8. *New prose is filler-free (`make prose`).* `gate steps no other workflow runs` is green on this head, and `prose` runs in that job.

**Not satisfied as written — items 3 and 5.**

Both name one command, and it was not run:

> `stella context validate` reports no conflict among these four records, and neither `hard` guard is suspended.
>
> Witness: `cargo run -p stella-cli --quiet -- context validate` — capture its output showing all four collisions and both suspended guards on `main` today, then capture the same command showing zero conflicts and both guards armed after the fix.

The PR body says so plainly: the installed `stella` is 0.9.300, which predates #5330's trigger comparison, so the local run could not see the collisions at all. What was run instead is a Python mirror of `find_conflicts` / `overlapping_scope` / `shared_triggers`, which is a re-implementation — it can agree with the detector and still not be it.

The substitute that *does* run the real code is `published_records_do_not_collide.rs`, which calls `load_context_file` → `assign_handles` → `detect_conflicts` → `is_suspended` against this repository's own `.stella/rules/`. That is the same library path `context validate` reports from, and it is exercised by CI's `fmt + clippy + test`, so the fail→pass flip is genuinely proven — just by a test rather than by the captured command output the DoD asks for.

Two ways to close this honestly, either of which is cheap:

- Run `cargo run -p stella-cli --quiet -- context validate` from this branch and from `main` (the branch build is current, so the 0.9.300 problem does not apply), and paste both outputs here. That satisfies items 3 and 5 as written.
- Or amend #5779's items 3 and 5 to name the integration test as the witness, since it checks the stronger property — a warning cannot be missed by a test the way it is missed by a human reading `context validate`'s exit 0.

What should not happen is ticking 3 and 5 on the strength of the Python mirror. Nothing pushed from this sweep.

---

Issue #5867, opened 2026-09-05
The deck's approve and dismiss keys cannot reach the plan change a running turn is held on


macanderson:
## The problem

A plan change raised by a failing gate now withholds the tool calls of a running turn (#5296). The person driving answers it in one indirect way and cannot answer it the other way at all.

**Approve** works by side effect. `crates/stella-cli/src/command_deck/driver_support.rs`'s `service_approve_revision` puts the repair row on the task board. The next `PlanGate::review` sees a board row whose subject matches the waiting change and treats that as the yes — `crates/stella-cli/src/command_deck/task_tap/plan_gate/revision.rs`'s `settle_revision`. Two consequences:

- A row the model itself creates with the same subject reads as an approval nobody gave.
- The deck's own message ("r2 approved — task 4 ...") is written before the revision exists, so a `RevisionGate::approve` that then fails with `PlanMoved` leaves the reader told one thing and the plan saying another.

**Dismiss** does not work at all. `RevisionGate::dismiss` exists and is tested in `crates/stella-core/src/plan_graph/revision/tests.rs`, and `stella_tui::deck_ui::row_keys` draws the `x` key, but nothing in `stella-cli` calls it. A person who does not want the repair task has no way to lift the hold: the turn stays refused until it ends.

**Edit** has the same gap and is already filed as #5289.

## Why it was not fixed in #5296

`service_approve_revision` runs on the deck's driver loop and holds a `ToolRegistry` and an `Inbound` sender. The turn's `PlanGate` lives inside the `TaskTap` that `run_lead_turn` builds, and there is no path from the keyboard to it. #5296 closed half of that by sharing the lane's `SharedRevisions` (`Arc<Mutex<RevisionGate>>`) between `run_lead_turn`'s forwarder and its plan gate; the driver loop does not hold a clone of it.

## The fix

Give the driver loop a clone of the lane's `SharedRevisions` so its verbs act on the gate the running turn reads:

- `x dismiss` calls `RevisionGate::dismiss`, which lifts the hold and touches no plan.
- `a approve` marks the change agreed to on the gate rather than inferring agreement from a board row. That needs a state `stella_core::plan_graph::RevisionGate` does not have today — a proposal the person has said yes to, which the next `PlanGate::review` writes with `RevisionGate::approve`. `RevisionGate` is pure and property-tested, so the state belongs there rather than in `stella-cli`.
- The deck's notice is then written after the revision is, or says plainly that the write is pending.

Constraints already discovered:

[...]

PR #5868, opened 2026-09-05
fix(stella-cli): hold a running turn's tool calls while a plan change waits

### LINKED ISSUE #5296: A standing plan revision does not withhold the tool calls of a turn already in flight
SPEC 8.1 item 3 says nothing runs until a proposed plan revision is approved. #5043 shipped two of the three halves that need:

- **`stella_core::plan_graph::RevisionGate::admits`** answers `false` while a proposal stands, and `PlanGraph::ran` refuses a task the current revision does not contain — the structural refusal, tested in `crates/stella-core/src/plan_graph/revision/tests.rs`.
- **`deck_ui::gates::revision_hold`** withholds every *prompt* the deck would start while any lane has a standing proposal.

What is missing is the middle: the **tool calls of a turn already running**. `crates/stella-cli/src/command_deck/task_tap/plan_gate.rs::PlanGate` is the one place that both owns a live `PlanGraph` and is consulted before every task-tap tool call — `PlanGate::review` already returns a `ToolOutput` refusal that stops a step. It holds no `RevisionGate`, so a gate failing mid-turn parks nothing on that path.

## Why it did not land with #5043

`PlanGate` never sees a `GateBoard`. Gate boards are emitted by the wrapper-plugin path (`crates/stella-cli/src/wrapper_plugin.rs`, `crates/stella-cli/src/agent/goal/goal_wrapped.rs`), and `PlanGate` is driven by tool outputs through `TaskTap`. #5043 wired the authoring where both facts *are* in view — `crates/stella-cli/src/command_deck/forwarder.rs`, which sees every `AgentEvent` for the lane and holds its `TaskBoardHandle` — and the forwarder is a separate task with no path back from the keyboard.

Note also that the deck binds no wrapper today, so no deck lane emits a `GateBoard` at all yet (#5261, #5266). This issue is about the path being right when one does.

## What "done" looks like

`PlanGate` holds a `RevisionGate`, is handed the failing `GateBoard`, and `PlanGate::review` returns a refusal naming the standing proposal while `!admits()`. Approval calls `RevisionGate::approve(&mut graph)` — already written and tested — so the `[:NEXT]` edge and the `Divergence` carrying the gate's cause come from `PlanGraph::revise` rather than a second code path. A witness test proves a tool call is refused while a proposal stands and admitted after the approval, with no terminal and no plugin.

The deck-side approval (`crates/stella-cli/src/command_deck/driver_support.rs::service_approve_revision`) creates the board row and lets the next `PlanGate::review` author the revision; once this lands, the two should meet rather than both writing.

## Constraints

`crates/stella-cli/src/command_deck/task_tap/plan_gate.rs` is 715 lines and ungrandfathered — plan a submodule. `crates/stella-cli/src/command_deck.rs` is a god file closed to growth.

Refs #5043, #5261, #5266.

## Definition of done

- [x] `PlanGate` holds a `RevisionGate`, is handed the failing `GateBoard`, and `PlanGate::review` returns a refusal naming the standing proposal while `!admits()`.
- [x] Approval calls `RevisionGate::approve(&mut graph)`, so the `[:NEXT]` edge and the `Divergence` carrying the gate's cause come from `PlanGraph::revise` rather than a second code path.
- [x] A witness test proves a tool call is refused while a revision stands.

### How each box was checked

Verified against #5868's diff (`crates/stella-cli/src/command_deck/task_tap/plan_gate/revision.rs`, plus the forwarder and lead-turn wiring):

1. `SharedRevisions` — an `Arc<Mutex<RevisionGate>>` built per turn in `run_lead_turn` — is written by the forwarder that receives the `GateBoard` and read by `PlanGate` through `PlanSetup`. `settle_revision` runs first in `review` and returns an `ErrorClass::RefusedByPolicy` output whose `hold_message` names the revision, the task it would add, the failing gate and the gate's cause. The one mechanism difference from the wording above: the board reaches `PlanGate` through the shared gate rather than through a `PlanGate` method taking a `GateBoard`, because only the forwarder ever holds one, so such a method would have no production caller and `stella-cli` fails `dead_code` under `-D warnings`. Recorded before the work in https://github.com/macanderson/stella/issues/5296#issuecomment-5549714457.
2. `PlanGate::approve_revision` hands the live graph to `RevisionGate::approve`, so `PlanGraph::revise` is the single writer. The witness asserts exactly one insertion and that its `Divergence` carries the gate's own cause.
3. `a_waiting_plan_change_stops_the_next_tool_call_until_it_is_answered` asserts refusal while the change waits and admission once the repair row is on the board. `a_change_with_no_plan_to_write_it_into_holds_nothing` pins the other direction.

macanderson:
## What and why

SPEC 8.1 item 3 says nothing runs until a proposed plan revision is answered. #5043 shipped two of the three halves — `stella_core::plan_graph::RevisionGate::admits` answers `false` while a change waits, and `deck_ui::gates::revision_hold` holds back the prompts the deck would start. Neither can stop a tool call the model has **already** asked for, so a gate failing mid-turn parked nothing on that path.

`crates/stella-cli/src/command_deck/task_tap/plan_gate.rs`'s `PlanGate` is the one place consulted before a guarded tool call that also owns the turn's `PlanGraph`. It now reads the lane's plan-change gate and refuses while one waits.

## How

One gate, shared, rather than two that could disagree:

- `SharedRevisions` (`Arc<Mutex<RevisionGate>>`) is built per turn in `run_lead_turn`.
- `spawn_forwarder` takes it and writes to it. The forwarder's own private `RevisionGate` is deleted — a `GateBoard` reaches this host on that stream and nowhere else.
- `PlanSetup` carries it into `PlanGate`. The turn's plan graph lives there and nowhere else.
- `PlanGate::review` calls `settle_revision` first. While a change waits and the board does not carry it, the tool call gets a `RefusedByPolicy` refusal naming the revision, what it would add, the failing gate and its cause.
- The deck's `a approve` verb already puts the repair row on the board (`driver_support::service_approve_revision`), so a board carrying the waiting subject **is** the person's yes. `approve_revision` hands the graph to `RevisionGate::approve`, so the `[:NEXT]` edge and the `Divergence` carrying the gate's cause come from `PlanGraph::revise` — the issue's "the two should meet rather than both writing".
- No plan graph, no hold: a yes is written into that graph, so a hold raised before a plan exists could never be lifted. Same failure direction the plan gate already takes for a card nobody answered.

New logic lands in a sibling submodule (`plan_gate/revision.rs`), not in `plan_gate.rs` (715 lines, ungrandfathered) and not in `command_deck.rs` (a god file — the one line touched there is an existing re-export, widened by one name, so the file does not grow).

### Why this differs from the issue's wording

The issue asks for a `PlanGate` that "is handed the failing `GateBoard`". A `PlanGate` method taking a board has no production caller — only the forwarder ever sees one — and `stella-cli` is a binary crate, so a `pub(crate)` fn nothing calls is a `dead_code` failure under `-D warnings`. Sharing the gate gives the same result with nothing unwired. Recorded on the issue before the work: https://github.com/macanderson/stella/issues/5296#issuecomment-5549714457

## Witness mechanism

`a_waiting_plan_change_stops_the_next_tool_call_until_it_is_answered` (`crates/stella-cli/src/command_deck/task_tap/plan_gate/revision.rs`) approves a three-step plan, writes a failing gate board onto the shared gate the way the forwarder does, and then asserts the next `review`:

- refuses with `ErrorClass::RefusedByPolicy`, naming `r2`, the subject and the `tests` gate;
- admits once the repair row is on the board, and leaves nothing waiting;
- has written `r2` into the plan graph with exactly one insertion whose `Divergence` carries the gate's own cause.

It fails on `main` by construction: `PlanGate` had no field to read, `review` no `settle_revision` to call, and no `approve_revision` to write the answer. No terminal, no plugin, no engine.

`a_change_with_no_plan_to_write_it_into_holds_nothing` pins the other direction — a change that arrives before any plan cannot wedge the turn.

Exemplar: `stella_core::plan_graph::revision`, whose `RevisionGate` this composes rather than reimplements.

## Tests deleted

None.

## Definition of done

#5296's three boxes are ticked, each checked against this diff rather than against the description. The one mechanism difference from the issue's wording — the board reaching `PlanGate` through the shared gate instead of a method parameter — is recorded beside the box it belongs to, with the `dead_code` reason above. Working shown under "How each box was checked" on the issue.

## Residue filed

- #5867 — route the deck's `a approve` / `x dismiss` keys into the live turn's gate directly, instead of inferring the yes from a board row. `x dismiss` reaches nothing today, so a person who does not want the repair task cannot lift the hold.

Closes #5296

## Summary by Sourcery

Ensure running turns stop guarded tool calls until pending plan revisions are answered and approved changes are recorded in the active plan.

New Features:
- Hold tool calls from an already-running turn while a plan revision is awaiting approval.
- Share the per-turn revision gate between the event forwarder and plan gate so approvals update the active plan graph.

Bug Fixes:
- Prevent tool calls from bypassing plan-change holds when a gate fails mid-turn.
- Avoid wedging turns when a revision arrives before a plan graph exists.

Enhancements:
- Return classified refusal messages that identify the pending revision, proposed task, failing gate, and cause, while routing approved revisions through the plan graph.

Tests:
- Add coverage for holding and resuming tool calls around an answered plan revision.
- Add coverage ensuring pre-plan revisions do not block execution.

### DIFF
--- a/crates/stella-cli/src/command_deck.rs
+++ b/crates/stella-cli/src/command_deck.rs
@@ -128,7 +128,7 @@ use slash_commands::{DeckCommand, handle_agents_input, run_deck_command};
 use crate::memory::{SessionMemory, TurnFriction};
 use crate::subsession::{self, SubSessions, SupervisorMsg};
 use authoring::{agents_list_creating, handle_agent_create};
-pub(crate) use forwarder::{close_turn_stream, spawn_forwarder};
+pub(crate) use forwarder::{SharedRevisions, close_turn_stream, spawn_forwarder};
 use sessions_view::sessions_inbound;
 use settings_io::{apply_pending_reload, handle_engine_config_input, handle_tools_input};
 
--- a/crates/stella-cli/src/command_deck/dropped_turn.rs
+++ b/crates/stella-cli/src/command_deck/dropped_turn.rs
@@ -322,6 +322,7 @@ mod tests {
             in_tx,
             LEAD.to_string(),
             None,
+            SharedRevisions::default(),
         );
         registry.attach_events(stella_core::EventSender::new(tx.clone()));
         tx.send(priced_step_usage(cost_usd)).expect("in flight");
--- a/crates/stella-cli/src/command_deck/forwarder.rs
+++ b/crates/stella-cli/src/command_deck/forwarder.rs
@@ -13,6 +13,10 @@ use crate::agent;
 use crate::cache_insight::{InsightScope, cache_insight_for};
 use crate::memory::TurnFriction;
 
+/// Re-exported beside [`spawn_forwarder`], the function that takes one, so a
+/// caller of the seam picks up the type from the same place.
+pub(crate) use super::task_tap::SharedRevisions;
+
 /// What a lane's closed turn stream leaves behind for its owner.
 ///
 /// Two facts, not one, since #3962. `persistence_complete` was always the
@@ -73,6 +77,10 @@ pub(crate) fn warn_audit_record_incomplete(
 /// see [`stella_core::tasks::TaskBoard::seed_from_plan`] for why the two used
 /// to be unconnected, and what that cost. `None` for lanes with no board of
 /// their own.
+///
+/// `revisions` is the lane's plan-change gate. A failing gate board puts a
+/// change on it here, and the lane's own plan gate reads it — see
+/// `task_tap::plan_gate::revision` for why one gate serves both.
 pub(crate) fn spawn_forwarder(
     mut rx: UnboundedReceiver<AgentEvent>,
     execution: Option<(Arc<Store>, i64)>,
@@ -80,6 +88,7 @@ pub(crate) fn spawn_forwarder(
     inbound: UnboundedSender<Inbound>,
     lane: String,
     plan_board: Option<stella_tools::tasks::TaskBoardHandle>,
+    revisions: SharedRevisions,
 ) -> tokio::task::JoinHandle<LaneStreamEnd> {
     tokio::spawn(async move {
         let mut seq = 0u64;
@@ -93,12 +102,12 @@ pub(crate) fn spawn_forwarder(
         // forwarder, and a forwarder is one lane — every lane has a registry
         // and a channel of its own, so two lanes' thoughts cannot fuse.
         let mut reasoning = agent::ReasoningRun::default();
-        // The lane's standing plan revision, and the number the next one would
-        // take. Held here because this task is the one place a `GateBoard` and
-        // the lane's plan are both in view: the gate goes past on this stream
-        // and the plan is `plan_board`'s rows. Per lane and per turn, like the
-        // ledger above.
-        let mut revisions = stella_core::RevisionGate::default();
+        // The number the lane's next plan revision would take. Read off the
+        // gate's own proposals below rather than counted here. The change
+        // itself goes on `revisions`, which this task writes and the lane's
+        // plan gate reads: a gate board reaches this host on this stream and
+        // nowhere else, and the turn's plan graph lives in the plan gate and
+        // nowhere else.
         let mut plan_revision = stella_protocol::PlanRevision::FIRST;
         // Two independent latches, not one. A single shared flag meant an
         // early usage gap — the benign, self-healing condition — silenced any
@@ -242,6 +251,8 @@ pub(crate) fn spawn_forwarder(
                         super::task_tap::plan_gate::plan_tasks(guard.items())
                     });
                     revisions
+                        .lock()
+                        .unwrap_or_else(|p| p.into_inner())
                         .observe(plan_revision.next(), &planned, board)
                         .cloned()
                 }
@@ -467,6 +478,7 @@ mod tests {
             in_tx,
             "lead".to_string(),
             None,
+            SharedRevisions::default(),
         );
         registry.attach_events(stella_core::EventSender::new(tx.clone()));
         tx.send(AgentEvent::TurnComplete {
@@ -545,6 +557,7 @@ mod tests {
             in_tx,
             "lead".to_string(),
             None,
+            SharedRevisions::default(),
         );
         tx.send(AgentEvent::Text { text: "hi".into() }).unwrap();
         tx.send(AgentEvent::TurnComplete {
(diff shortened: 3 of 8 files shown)


macanderson:
`dod / dod` is red (job 101256712739, run 33947713604) because all three DoD boxes on #5296 are still unticked. I checked each one against this branch's diff rather than against the PR description. Two are met literally; one is met in substance but not in the mechanism its wording names, so it is the author's call to tick, not mine.

**Box 2 — met literally.** `approve_revision` (`crates/stella-cli/src/command_deck/task_tap/plan_gate/revision.rs`) takes the turn's graph and calls `self.revisions.lock()…approve(graph)`, so the insert goes through `RevisionGate::approve(&mut graph)` and the `[:NEXT]` edge and the `Divergence` carrying the gate's cause come from `PlanGraph::revise`. There is no second write path.

**Box 3 — met literally.** `a_waiting_plan_change_stops_the_next_tool_call_until_it_is_answered` in the same file asserts `Some(ErrorClass::RefusedByPolicy)` on the held call, and `a_change_with_no_plan_to_write_it_into_holds_nothing` pins the other direction.

**Box 1 — the mechanism differs from the wording.** Its three clauses do not all land the same way:

- *"`PlanGate` holds a `RevisionGate`"* — met. `PlanGate` carries a `revisions: SharedRevisions` field, and `PlanSetup` carries it in.
- *"`PlanGate::review` returns a refusal naming the standing proposal while `!admits()`"* — met. `review` calls `settle_revision` first, and `hold_message` names the revision, the subject it would add, the failing gate and its cause.
- *"is handed the failing `GateBoard`"* — **not met as written.** No `PlanGate` method takes a `GateBoard`. The forwarder is the only place a board exists (`forwarder.rs`, the `AgentEvent::GateBoard` arm), and it writes the derived proposal into the shared gate that `PlanGate` reads. The board's content reaches the gate; the board itself does not.

That substitution is argued on the issue before the work landed (#5296 comment 5549714457): a `PlanGate` method taking a board would have no production caller, and in a binary crate an uncalled `pub(crate)` fn is a `dead_code` failure under `-D warnings`. I have no objection to the substitution — I am flagging only that ticking box 1 asserts wording the diff does not literally satisfy, so either the box is ticked knowing that, or its text is amended to the mechanism actually shipped.

Nothing else is blocking: `fmt + clippy + test` was still running at the time of writing, and every other check on this head is green.

---

Issue #5871, opened 2026-09-05
Nothing keeps the verdict record a wrapper round now writes


macanderson:
## In plain terms

The host now writes a record of who judged each wrapper round. Nothing keeps it. The record lives as long as the value the dispatch returns, and then it is gone — so the thing built to answer "who checked this?" later still cannot answer it later.

## Problem

`DispatchReport::snapshot` (`crates/stella-runtime/src/wrapper/dispatch.rs`) carries a `LadderSnapshot` with one `VerdictStamp` on it, written by `crates/stella-runtime/src/wrapper/stamp.rs`. Its only reader today is the report's own caller.

`rg -n 'report.snapshot|\.snapshot' crates/stella-cli/src/wrapper_plugin` finds nothing: `crates/stella-cli/src/wrapper_plugin/report.rs` renders the verdict, the gate board, the spends and the faults, and drops the record. Nothing writes it to `.stella/private/store.db`, and no `AgentEvent` carries it, so a session that finishes leaves no trace of the claim.

That is the gap AGENTS.md's rule 10 is about: a signal with no consumer is allowed only as a declared gap, and this one is not declared anywhere.

## The fix

Pick one destination and wire it, with its consumer declared.

1. **Store.** `crates/stella-store` already keeps the events a turn produced. A verdict row that carries the snapshot as JSON is the smallest change and makes `stella observe` able to show the claim. A new store migration is six edits across five files — check the schema version, the DDL, the migration, the reader, and the tests.
2. **Event.** An `AgentEvent` carrying the verdict and its record is the wire-visible answer, and it needs a row in `crates/stella-protocol/src/event/consumers.rs` naming what reads it. `KNOWN_TYPE_TAGS` and `type_tag()` are generated from that table, so a missing row is a build error rather than a silence.

Either way the record must arrive whole: `preimage_hash` is computed over the snapshot with its stamp list dropped, so a reader that re-derives it from a partial copy gets a different hash and reads it as tampering. Round-trip it through `serde_json` and compare, per AGENTS.md rule 4.

## Files

- `crates/stella-runtime/src/wrapper/dispatch.rs` — where the record is produced.
- `crates/stella-cli/src/wrapper_plugin/report.rs` — the one caller that holds a `DispatchReport` today.
- `crates/stella-protocol/src/event/consumers.rs` — the ledger, if the answer is an event.
- `crates/stella-store/src/ddl.rs` and its migrations, if the answer is a row.

## Verify

[...]

Issue #5870, opened 2026-09-05
A verdict stamp names who judged, but never which version of them


macanderson:
## In plain terms

Every verdict stamp Stella writes says who made the claim, but never which version of them made it. The field for it exists and is always empty, because a plugin's manifest has nowhere to write a version.

## Problem

`VerdictStamp::author_version` (`crates/stella-protocol/src/ladder/stamp.rs`) is an optional string a reader may show. The producer — `crates/stella-runtime/src/wrapper/stamp.rs`'s `stamped` — sets it to `None` on every stamp, and the comment there says why:

- `PluginManifest` (`crates/stella-plugin/src/manifest.rs`) declares no version field. `rg -n 'version' crates/stella-plugin/src/manifest.rs` finds only the wire protocol version.
- The one place a version could come from instead is the plugin's own payload, and reading it there is exactly what the `author` field exists to prevent. A host that trusts a plugin's word for its version has no reason to distrust its word for its name.

So two runs judged by two different builds of the same plugin leave records nothing can tell apart. That matters the first time a plugin's oracle changes behaviour and somebody asks which build produced a claim.

## The fix

Give `PluginManifest` an optional `version` field, and read it in `stamp::stamped`.

- Add `pub version: Option<String>` to `PluginManifest`, with `#[serde(default, skip_serializing_if = "Option::is_none")]` so every manifest in the tree keeps loading unchanged.
- Decide whether the value is free text or a semantic version. Free text is the smaller change and matches the field's contract ("copied word for word; nothing branches on it"); a checked format is a load-time refusal a manifest author has to satisfy. Pick one and say why in the PR.
- Show it at install consent (`crates/stella-plugin/src/consent.rs`), so the version a person agreed to is the version the stamp names.
- In `crates/stella-runtime/src/wrapper/stamp.rs`, pass the manifest's version through and set `author_version` from it. It follows `author`: read from the loaded manifest, never from a payload. `HostObserved` evidence keeps `None` — the engine has no version of its own here.

## Files

- `crates/stella-plugin/src/manifest.rs` — the field.
- `crates/stella-plugin/src/consent.rs` — the install prompt.
- `crates/stella-runtime/src/wrapper/stamp.rs` — `stamped`, which sets `author_version`.
- `crates/stella-runtime/src/wrapper/dispatch.rs` — `stamp_round`, which passes the id today and would pass the version too.
- `crates/stella-runtime/tests/wrapper_verdict_stamp.rs` — where the witness goes.

[...]

PR #5873, opened 2026-09-05
feat(stella-runtime): stamp the record a judged round leaves behind

### LINKED ISSUE #5826: Write a verdict stamp when a plugin's evidence becomes part of a run's record

macanderson:
## What and why

`VerdictStamp` and `LadderSnapshot::stamps` landed with no writer (#5120, PR #5827). The list was empty on every real run, so the question the type exists to answer — who checked this, and what did they find — had no answer at the one place it matters.

This adds the producer at the fold point: `crates/stella-runtime/src/wrapper/`, where the host turns a plugin's `EvidenceSet` into a `Verdict`.

- **`crates/stella-runtime/src/wrapper/stamp.rs`** (new) — folds an `EvidenceSet` and a `Verdict` into a `LadderSnapshot` (`stamp::snapshot`) and puts one claim on it (`stamp::stamped`). Pure and total, like `judge` beside it: it reads no clock and touches no file, so a test can pin a whole stamp. A sibling module rather than lines in `verdict.rs` or `dispatch.rs` (701 and 816 lines) — the fold is its own subject.
- **`dispatch.rs`** — `DispatchReport` gains `snapshot`, and `WrapperDispatch::run` stamps the round that stops the loop. `with_clock` injects the time source; the default is `stamp::HostClock`, wall-anchored so two stamps from two runs can be compared (`stella-cli`'s `WallClock` answers the same port the same way).
- **`error.rs`** — `WrapperError::Unstampable`. A record that cannot be hashed keeps its answer and loses only the name on it: hashing runs after the decision and can change nothing about it, so failing the whole run over one would throw away a good verdict. It joins `faults`, where every other silence in this module is reported.

### The rules the issue fixed

- **`author` is host-filled.** `stamp::author` maps `EvidenceProvenance::PluginReported` to the loaded manifest's `[wrapper] id` and `HostObserved` to `"engine"`. The wire has no author field at all — `ObservedEvidence` and `AfterTurnResponse` are both `deny_unknown_fields` — so a plugin's own note is the only place it can try to say who it is, and that note never reaches this field.
- **`preimage_hash` is `record_hash` over `stamp_preimage()`**, taken before the stamp is pushed. A later observer can add a claim without breaking the first one, and a unit test recomputes the hash after a second stamp lands.
- **No stamp moves a rung.** `stamp::snapshot` and `stamp::stamped` return the same `rung`, asserted over both a met and an unmet round.
- **`author_version` is `None`.** A manifest declares no version field, and the one word a plugin could offer here is the one word it must not choose. Filed as #5870.

### Honest limits of the record

The socket carries a flip, a host-side tamper finding, and numbers the plugin's own oracle named. It carries no test command, no per-tree observation and no diff, so those fields sit at the value that means "nothing looked" — `None`, an empty list, `diff_available: false` — never at a flattering zero. A plugin's measurements are keyed by names its own oracle declared, so mapping one onto `diff_lines` would invent a shared vocabulary that does not exist.

## Witness mechanism

`crates/stella-runtime/tests/wrapper_verdict_stamp.rs`, three tests over the whole dispatch (not the fold alone — the name on the stamp is the point, and a unit test of the fold could pass while the host still sent the wrong string):

1. `a_decided_round_is_stamped_with_the_manifests_name` — a plugin-reported `EvidenceSet` yields a snapshot with exactly one stamp, `author == "witness-v1"` (the manifest's `[wrapper] id`), and `record_hash` recomputed over the returned snapshot's `stamp_preimage()` equals the stamp's hash.
2. `a_plugin_cannot_name_itself` — a new fixture mode, `flip-and-claim-author`, signs its note "checked by vera"; the stamp still reads `witness-v1`.
3. `the_stamp_points_at_the_workspace_it_judged` — the candidate handle rides `evidence_refs`.

**These fail on `main` by construction: `DispatchReport` has no `snapshot` field there, so the file does not compile.** The unit tests in `stamp.rs` fail for the same reason — the module does not exist. They cover the mapping table (assessment, rung, flip, tamper) and the hash property.

A pinned `Clock` in the test makes both times values the file names, so nothing here is timing-dependent.

## Also in this PR

`LadderSnapshot::witness_intact`'s doc said `Some(false)` never reaches a verdict, because tampering aborted the candidate on the deleted staged pipeline. On the wrapper path a `Tampered` finding produces an `Unmet` verdict and does reach a record, so the doc is corrected and the four generated `docs/wire/` artifacts carry the same text.

## Exemplar

`stamp::stamped` follows `verdict.rs`'s `judge` — synchronous, total over a closed vocabulary, every arm documented — because it is the same shape one step later: `judge` decides, this records the decision.

## Consumers ledger

No new `AgentEvent`, so no row in `crates/stella-protocol/src/event/consumers.rs`. The stamp rides an existing struct field.

## Tests deleted

None.

## Remaining work filed

- #5870 — a manifest declares no version, so every stamp's `author_version` is `None`.
- #5871 — nothing carries `DispatchReport::snapshot` into the store or onto the wire, so a stamp lives only as long as the report.
- #5918 — the completion gate is handed `None` for its rung, so `Arbitration::refutes_done` never fires on a live run. This branch is the first thing to hold an answer for it; wiring it is a change with its own consumers and its own tests.
- #5920 — the stamp names the composition and the arbiter claim names the arbiter, and nothing says which one is the author of a decision. The two agree for a single member, which is why every test today passes.

Closes #5826

## Summary by Sourcery

Record each judged wrapper round as a host-authored, hashable ladder snapshot with a verdict stamp.

New Features:
- Add verdict stamping to wrapper dispatch reports, recording the manifest-owned author, assessment, evidence references, timing, and tamper state for each decided round.

Bug Fixes:
- Ensure record hashes exclude the stamp list so additional claims do not invalidate existing stamps.
- Prevent plugins from supplying the stamp author and preserve decided verdicts when record hashing fails.

Enhancements:
- Map wrapper evidence and verdicts into ladder snapshots without changing their rung.
- Expose injectable clocks for deterministic stamp timing and report unstampable records as wrapper faults.

Documentation:
- Document the new stamp module and end-to-end stamp tests, and update ladder witness-tamper semantics in generated wire documentation.

Tests:
- Add unit and dispatch-level coverage for author attribution, hash recomputation, multiple stamps, rung preservation, abstentions, tamper findings, and workspace references.


macanderson:
Answering the one ❌ row: **"Update documentation that states no producer writes verdict stamps"** — there is no such documentation to update, and here is the check.

- **`docs/spec/plugin-architecture.md` does not exist.** `ls docs/spec/plugin-architecture.md` → `No such file or directory`. The issue named a path that is not in this tree, so its §7.2 is not a thing this PR can leave stale.
- **`docs/spec/pipeline-as-plugins.md` says nothing about a stamp producer.** `rg -n 'VerdictStamp|stamps' docs/spec/pipeline-as-plugins.md` matches nothing. Its one paragraph about the record is §8's "Net-new, not ported: the durable flip record", and it says a `LadderSnapshot` rides inside `AgentEvent::Verdict` while nothing decides on it. That is still true after this PR: the stamp is written and returned on `DispatchReport::snapshot`, and nothing persists it. Editing that paragraph would make it wrong.
- **Nothing else in `docs/` claims a stamp has no writer.** `rg -rn 'VerdictStamp' docs/` matches only the generated `docs/wire/` artifacts, which this PR regenerates for the `witness_intact` doc correction.

The two prose homes that did need editing are edited: `crates/stella-runtime/README.md` gains rows for `src/wrapper/stamp.rs` and `tests/wrapper_verdict_stamp.rs`, and `LadderSnapshot::witness_intact`'s doc no longer claims `Some(false)` cannot reach a verdict — on the wrapper path a `Tampered` finding produces `Unmet` and does reach a record.

The gap the row is reaching for is real and is filed rather than papered over: #5871 tracks the fact that nothing carries `DispatchReport::snapshot` into the store or onto the wire, so a stamp lives only as long as the report. #5870 tracks the empty `author_version`.

macanderson:
Rebuilt on `main` after #5829 landed. The record plane left `stella-core` for `stella-records`, and `record_hash` went further than that — it is `stella_protocol::hash` now — so `stella_core::context_record` resolved to nothing and `stamp.rs` did not compile on the merge (windows-check and macos-check both caught it before the Linux job started). Merged `main` in rather than rebasing, per AGENTS.md's rebase-replay note, and repointed the two imports. No new dependency: `stella-runtime` already takes `stella-protocol`.

While there, `LadderSnapshot::stamp_preimage`'s doc still said the digest lives in `stella-core` with hashing crates `stella-protocol`'s boundary does not take. That crate holds it now, so the doc says so. The change is on a method rather than a field, so no `docs/wire/` artifact moves with it.

macanderson:
Rebuilt on `main` again: #5884 landed the arbitration law at the same gate this branch stamps, so both sides edit `WrapperDispatch::run`'s stopping arm and `stella-cli`'s report fixture.

Kept both records rather than choosing one. The round is stamped first, then the gate's claims are folded, and `DispatchReport` carries `snapshot` beside `arbitration`. `faults` is main's `Faults` record now, so the stamp's `timed_out` reads `faults.errors`, and a record that cannot be hashed goes on through a new `Faults::push_host`. That second arm exists because the two failures are different facts: a member that fell silent lost its say, while the host that could not hash kept its answer and lost only the name on it, so no `did_not_answer` claim rides with it.

`fold_stamps` still takes `None` for the rung, which is what main does today. This dispatch now holds an answer for it — `snapshot.rung` — and handing it over would arm `Arbitration::refutes_done` on the live path for the first time. That is a change with its own consumers and its own tests, so it is filed rather than smuggled into a merge: #5918. #5920 is the other thing the collision showed — the stamp names the composition and the arbiter claim names the arbiter, and nothing says which is the author of a decision.

Issue #5877, opened 2026-09-05
A sub-session lane that panics leaves no terminal frame for the lead to report


macanderson:
## Problem

A deck sub-session lane now leaves a **terminal frame** when it dies mid-turn —
its last committed step plus the transcript — and the lead reads that frame and
reports it when the lane ends (#3233,
`crates/stella-cli/src/subsession/terminal_frame.rs`).

One death does not reach it: a panic on the worker thread.
`crates/stella-cli/src/subsession.rs`'s `spawn` catches a panicking
`run_worker` and synthesizes `WorkerEnd::Failed` with `PANIC_FAILURE_PREFIX`,
but `LaneRecorder::settle` is called *inside* `run_worker`, so the unwind skips
it. The lane's per-step checkpoint is still on disk and readable, and nothing
turns it into a frame, so the lead's report says nothing about a lane that died
the loudest way it can.

## Where

- `crates/stella-cli/src/subsession.rs` — `spawn`'s
  `std::panic::catch_unwind` arm, and `run_worker`'s `recorder.settle(&end)`.
- `crates/stella-cli/src/subsession/terminal_frame.rs` — `LaneRecorder`.

## The fix, roughly

Settle from the catching side rather than only from inside the turn: hold the
`LaneRecorder` (or the lane key plus its durability handle) where the catch can
reach it, and settle there when the caught arm synthesizes its `WorkerEnd`. The
recorder's `last_committed` already falls back to the checkpoint standing at
the lane's tip, which is exactly the state a panic leaves, so no new source of
truth is needed.

Mind the borrow: `run_worker` takes `&SessionDurability` by value into the
sink; the catch site would need its own clone of the handle (`SessionDurability`
is `Arc`-backed and cheap to clone).

## How to verify

A test in `crates/stella-cli/src/subsession/tests.rs` beside
`run_caught`'s existing panic coverage: bind a lane, persist a checkpoint
through its sink, run a body that panics, and assert the lane's terminal frame
exists and names the committed step. It fails today because nothing writes a
frame on that path.

## Definition of done

- [ ] A lane whose worker thread panics leaves a terminal frame naming its last
      committed step.
- [ ] A test covers that path and fails without the fix.
- [ ] The lead's `Ended` report reaches the panic path the same way it reaches
      an abort and a stop.

macanderson:
Folded into batch #6308 (2026-09-06 audit): fixed together by crate.

Issue #5102, opened 2026-08-26
A killed sub-session lane's transcript is saved but nothing can reopen it


macanderson:
## Problem

`stella daemon resume` restores exactly one turn: the lead's. A deck sub-session lane now has a durable per-step record of its own — #4725 bound each lane under `lane_journal_key(session, lane)` = `{session}__{lane}`, and #3233's follow-up made those keys readable — but nothing outside the deck process can find or re-enter one.

Kill interactive mode while `req:3` is thirty steps into a delegated task and the lane's checkpoint is written, complete, and unreachable. It is picked up only if the *same* lane id happens to be respawned inside a deck session with the same id — the Restart verb, or a task re-assigned — which is a coincidence, not a read path.

This is item 2 of #3233's "What done looks like", left open there deliberately.

## Why `resume` cannot see a lane today

`crates/stella-cli/src/agent/resume.rs`'s `run_resume` resolves its target through `stella_store::SessionRegistry` (`crate::daemon::resolve`, or `crate::daemon::supervised_id()`), then binds `cfg.durability` to `record.id`. A lane key is never written to the registry — `crates/stella-cli/src/subsession.rs` touches neither `SessionRegistry` nor `SessionRecord` anywhere — so `daemon::resolve` cannot produce one and `stella daemon resume` has no way to name a lane.

The record itself is not the blocker any more. Every ref the work journal writes lives under its key, so `refs/stella/{session}__{lane}/…` is the durable list of a session's lanes, and `WorkJournal::recorded_keys(prefix)` / `SessionDurability::recorded_keys(prefix)` read it back.

## The decisions this needs

1. **Does a lane get a `SessionRecord`?** Registering one makes it resolvable and listable, and makes N lanes N rows in a registry whose unit has always been a CLI process. Not registering means `resume` needs a second lookup path.
2. **What does resuming a lane mean?** A lane is a worker inside a deck; re-entering it standalone is a different product from re-opening the deck and respawning the lane in place.
3. **Fidelity.** `subsession::initial_messages`' own doc says a lane restores at *turn*-boundary fidelity, not the step-boundary fidelity the lead gets — the mid-completion step is re-asked, not continued. Whether that gap closes here or stays declared is part of the answer.

## Files

- `crates/stella-cli/src/agent/resume.rs` — `run_resume`, `drive_resumed_turn`.
[...]

PR #5878, opened 2026-09-05
feat(stella-cli): keep a killed sub-session lane's transcript for the lead to report

### LINKED ISSUE #3233: A sub-session in interactive mode has no step-level durability, so killing it loses its entire transcript

macanderson:
## What and why

A deck sub-session lane runs beside the lead turn and does not resume itself.
`doc:turn-lane-assembly` §6 (approved 2026-08-14) calls that
`ResumeAuthority::Parent`: the lane's history is read by its **parent**, for a
report, and never re-entered as a resume point. A lane like that owes one
record rather than a per-step one — the **terminal frame**: the steps it got
through plus the messages that got it there, written once, when the lane dies.

Until now it left nothing to read. `Engine::drive` discards the resume point at
every end of a turn, an abort included, so a failed lane's transcript was
deleted by the engine before any reader existed: the file edits stayed in the
shared tree and the conversation that produced them did not.

Three pieces:

- `crates/stella-store/src/lane_frame.rs` — the reserved blob, beside
  `CHECKPOINT_BLOB` in the same commit graph (that blob's own argument: a
  turn's record and the file changes it covers cannot be split across two
  stores). Its own blob because it has its own lifetime: a checkpoint dies with
  its turn, a frame is written *because* the turn ended badly and stands until
  that lane finishes a later try. `blob_at_tip` becomes `pub(crate)` so this
  module can read at the tip; nothing else in `work_journal.rs` changed, and
  the module is a sibling file because that file sits 30 lines under the
  1500-line ceiling.
- `crates/stella-cli/src/subsession/terminal_frame.rs` — the lane's end.
  `LaneSink` wraps the lane's own checkpoint sink and keeps what the engine's
  terminal `discard` retires (one `git show` per turn, rather than a clone of
  the transcript per step). `LaneRecorder::settle` is where a lane that
  finished and a lane that died part: the first clears its frame, the second
  writes one, and a lane that died before its first step writes nothing because
  it has nothing to hand over.
- `crates/stella-cli/src/command_deck/driver_support.rs` — the reader. On a
  lane's `Ended`, for a lane that did not finish, the lead reads the lane's own
  frame and prints one line on its transcript naming how far the lane got.

Exemplar: `stella_store::work_journal`'s own `CHECKPOINT_BLOB` /
`PIPELINE_BLOB` pair — a reserved blob with a stated lifetime, a `record`, a
read at the tip and an idempotent retraction — and `durability.rs`'s
`bind_session_in` seam, which is why `parent_report_in` takes an explicit store
root so tests never touch the process-global `data_dir()`.

## Witness

`a_lane_that_failed_leaves_its_parent_a_frame_naming_the_last_committed_step`
(`crates/stella-cli/src/subsession/terminal_frame.rs`) drives the lane's sink
exactly as the engine does — `persist` at a step boundary, then the terminal
`discard` — and then asserts **both halves**: the checkpoint is gone (the
engine retired the only copy, which is the failure this issue is about) and the
frame still carries the four committed steps and the transcript, with the
parent's report naming `req:1` and `4 committed steps`. On `main` the second
half has nothing to read: there is no frame, no `LaneSink` keeping the retired
copy, and `SessionDurability` has no terminal-frame side at all, so the test
cannot compile or pass without this change.

Three more pin the rest of the contract:
`a_stopped_lane_frames_the_point_its_dropped_turn_left_standing` (a stop's
dropped future reaches no terminal path, so the frame comes from the point
still at the lane's tip),
`a_lane_that_completed_leaves_no_frame_and_retires_an_older_one` (the delta —
a completed lane writes none and clears an earlier attempt's frame, so the
parent cannot report a death the lane recovered from), and
`a_lane_that_died_before_its_first_step_leaves_nothing_to_report`.

Store side: `a_frame_is_written_read_back_and_retracted` and
`retiring_the_checkpoint_leaves_the_frame_standing`
(`crates/stella-store/src/lane_frame.rs`) — the second is the reason the frame
is not a widened checkpoint.

Local: `cargo fmt --all` and `make guards-fast` are green (`check-prose`
included — both new files are written to the reading-grade ceiling). Nothing
was compiled or tested on the maintainer's laptop; CI is the evidence.

## Scope

Not a resume path. Reopening a killed lane standalone still needs a
session-registry record for a lane, and stays on #5102 — this PR does not
touch it. The fleet half of the same `ResumeAuthority` split is #3232 and is
untouched: `fleet_cmd` reaches `subsession_engine_config_for` too, and this
change wraps at the sub-session call site rather than inside that function, so
a fleet attempt behaves exactly as before.

Tests deleted: None

Residue filed: #5877 (a lane whose worker thread panics settles no frame).

Closes #3233

## Summary by Sourcery

Preserve and report the final committed state of sub-session lanes that terminate without completing.

New Features:
- Preserve a terminal transcript frame for sub-session lanes that fail or are stopped so their parent can report their progress.
- Have the lead report a dead lane's committed step count and saved transcript status.

Bug Fixes:
- Prevent failed lane transcripts from being lost when the engine retires their checkpoint at turn termination.

Enhancements:
- Separate terminal frames from resumable checkpoints, including lifecycle handling for replacement, completion, and pre-step failures.
- Add durable store APIs for recording, reading, and clearing lane terminal frames.

Tests:
- Add store and sub-session coverage for failed, stopped, completed, and pre-step lane termination scenarios.


macanderson:
Automated PR sweep. Settling the two ❌ rows in Sourcery's assessment table, per CLAUDE.md's "every Sourcery ❌ gets a fix or an answer before the PR is mergeable".

**Row 2 — "Provide a read and re-entry path … recovered and resumed": answered, not fixed.** Deliberately out of scope, and the PR already says so under **Scope**: reopening a killed lane standalone needs a session-registry record for a lane and stays on #5102. Sourcery read the diff correctly; the omission is intended.

**Row 3 — "lane durability remains isolated from the lead … including the relevant durability isolation tests": confirmed, and it is not only a test-update gap.** Checked against the diff rather than the review text — `crates/stella-cli/src/agent/tests/durability_isolation.rs` is not in the 7 changed files, so #3233's numbered condition 3 ("Replace `a_sub_session_carries_no_checkpoint_sink` with an assertion that the lane's sink differs from the lead's") is untouched. That condition presumes a lane *has* a sink of its own; this PR keeps the sink stripped and adds a separate terminal-frame blob beside it, so the assertion has nothing new to assert yet.

**The consequence worth acting on: `Closes #3233` overstates this PR.** Measured against that issue's own Definition of done, none of its three boxes is satisfied by this diff:

- *"Each live sub-session lane binds its own `SessionDurability` under its own session id … `refs/stella/<session>/head`"* — not done. `LaneSink` wraps the lead's existing sink and retains what the terminal `discard` retires; no per-lane session id or ref namespace is minted.
- *"The remaining conditions this issue numbers are met"* — not done (condition 2, the read/re-entry path, is explicitly deferred to #5102; condition 3 is untouched, above).
- *"A killed lane's turns are recoverable"* — not done. The frame makes a dead lane **reportable** to its parent, which is `ResumeAuthority::Parent` and a real advance, but recoverable is what the box asks for.

So this is `Refs #3233`, not `Closes #3233` — in the PR body **and** in the commit trailer, since squash composes the body from `COMMIT_MESSAGES` and the trailer is what actually closes on merge. Left for the author rather than force-pushed over a branch opened minutes ago; AGENTS.md's own guidance is `Refs` for a PR that advances an issue without finishing it.

Worth fixing before merge rather than after: `dod-close-guard` reads those boxes and reopens an issue closed with them unticked — it did exactly that to #5813, 18 seconds after #5816 merged. Merging as-is buys a close, a reopen, and a stale `closed_by_pull_requests` link on an issue whose real work continues on #5102.

No CI or conflict problems here: the branch merges clean against `913e21a`, and the checks were still queued at sweep time.

---

macanderson:
Answering both `❌` rows in Sourcery's assessment. Neither belongs to this PR, and the reason is the same in both cases: the table is read against #3233's original body, which predates the design the maintainer approved for it.

**Row 2 — "Provide a read and re-entry path that allows an interrupted lane's turn to be recovered and resumed."**

Deliberately not this PR, by decision rather than omission. `doc:turn-lane-assembly` §6 (approved 2026-08-14) makes resume authority a property of the lane: a sub-session lane is `ResumeAuthority::Parent`, which means "the lane's checkpoint is read by its parent for *reporting*, not resumption: a killed sub-session hands its transcript to the lead rather than restarting itself." The frame is what such a lane owes, and it has a reader by construction — `driver_support.rs`'s `Ended` arm.

The half of re-entry that is real already shipped in #4725, and this PR does not touch it: `subsession::run_worker` reads the lane's own checkpoint before it builds `initial_messages`, so a lane re-attempted on the same key resumes its transcript instead of starting over. What is left is re-opening a killed lane standalone, which needs a session-registry record a lane does not have; that is #5102, filed with what #5101 established about where the record lives.

**Row 3 — "…including the relevant durability isolation tests… it does not update the issue's existing durability-isolation assertions or demonstrate the requested per-step lane checkpoint behavior."**

Those landed in #4725 and pass unchanged on `main` today:
`a_sub_session_carries_its_own_checkpoint_sink_never_the_leads` and
`a_sub_sessions_turn_cannot_destroy_the_leads_resume_point`, both in
`crates/stella-cli/src/agent/tests/durability_isolation.rs`. The issue body's item 3 asks for `a_sub_session_carries_no_checkpoint_sink` to be *replaced* with an assertion that the lane's sink differs from the lead's, and that replacement is the first of those two names.

This PR leaves `agent::subsession_engine_config_for` — the function those tests assert on — byte-identical, and wraps the sink at the sub-session call site instead, so a fleet attempt going through the same function is unaffected. Nothing here changes which record a lane checkpoints into.

macanderson:
`dod / dod` is red on the current head (`35a99b18`, job 101269899541), and it is the only failing check — every other job on this head is green, including `fmt + clippy + test`. The branch merges clean against `origin/main` (`19a5191`).

The three unticked boxes are the ones the sticky comment above names, so I am not restating them. One fact that is not yet on this PR, and that changes the remedy:

**The `Closes #3233` trailer is in commit `489a62ee`, not only in the PR description.** Editing the body to `Refs #3233` turns `dod / dod` green — that check reads the PR body — but squash composes its commit message from `COMMIT_MESSAGES`, so the trailer still lands on `main` and still closes #3233, and `dod-close-guard` then reopens it against the unticked boxes. That is the close-and-reopen the earlier comment on this PR predicted, reached even after the body is corrected.

Amending the trailer would need a force-push, which this repo's rules forbid on an open PR branch, so the trailer can only be dropped by editing the composed message in the squash-merge box at merge time.

That leaves two honest routes, and both are scope calls for the author rather than a sweep:

1. **`Refs`, not `Closes`** — change the body, and strip the trailer from the squash message when merging. #3233 stays open on #5102's work.
2. **Narrow #3233** — restate its checklist against the `ResumeAuthority::Parent` design that `doc:turn-lane-assembly` §6 approved, tick what this diff satisfies, and move the resume half onto #5102 with a line saying so. This is the shape used on #5121 for PR #5884, and it is the route that lets `Closes` stand.

Not acting on either here: which of the two is right depends on whether #3233 is the resume issue or the reporting issue, and the 06:10 reply on this PR argues the issue body predates the approved design — that is a decision, not a defect. Nothing else on this PR needs work.

---

macanderson:
**Answering the two ❌ rows in Sourcery's assessment table.** Both read the
issue body as it stood before the maintainer's design comment on it; neither
names a defect in this diff.

**Row 2 — "provide a read and re-entry path … recovered and resumed."**
Correct that this PR adds no resume path, and it says so under *Scope*. The
row's premise is what has moved. `doc:turn-lane-assembly` §6 (approved
2026-08-14, posted on #3233) makes a sub-session lane
`ResumeAuthority::Parent`: its history is read by the parent for a report and
is never re-entered as a resume point. A lane like that owes a terminal
frame, which is what this lands.

Re-entry is also not missing today. #4725 already has `subsession::run_worker`
read the lane's own checkpoint before it builds the first messages, so a
second spawn of the same lane picks the interrupted turn back up, and #5101
stops a resumed deck minting a dead lane's key for a fresh prompt. The half
that is genuinely absent — asking for a lane by key from a process with no
spawn behind it — needs a session-registry record for a lane and three
decisions this PR does not make. That is #5102, open, and #3233's box for
this condition names it.

**Row 3 — "does not update the issue's existing durability-isolation
assertions or demonstrate per-step lane checkpoint behaviour."**
Both were done before this PR, which is why the diff does not touch them.
`a_sub_session_carries_no_checkpoint_sink` no longer exists: #4725 replaced it
with `a_sub_session_carries_its_own_checkpoint_sink_never_the_leads`
(`crates/stella-cli/src/agent/tests/durability_isolation.rs`), which persists
through the seam and asserts the bytes land in the lane's record and not the
lead's. `a_sub_sessions_turn_cannot_destroy_the_leads_resume_point` sits in
the same file and still passes with both arms. Per-step lane checkpointing
ships in that same PR — `run_worker` binds `lane_journal_key(session, lane)`
and `subsession_engine_config_for` re-keys the sink to it.

The terminal frame is not offered as a substitute for that. It exists because
per-step durability alone does not survive the end of a turn: `Engine::drive`
discards the resume point on every terminal path, an abort included, so a
failed lane's transcript was deleted by the engine before any reader existed.
`retiring_the_checkpoint_leaves_the_frame_standing`
(`crates/stella-store/src/lane_frame.rs`) is that difference as a test.

#3233's Definition of done now carries one box per numbered condition with
the PR and test behind each, in place of the box that read "the remaining
conditions this issue numbers are met".

Issue #5883, opened 2026-09-05
Nothing builds a ladder snapshot from a wrapper run, so no verdict stamp is ever stored


macanderson:
## The problem

`VerdictStamp` (#5120, PR #5827) is the wire record of who claimed what about a
verdict, and `LadderSnapshot::stamps` is where it lands.
`ArbiterClaim::into_stamp`
(`crates/stella-runtime/src/wrapper/arbitration.rs`) mints one from a
completion-gate claim.

Nothing calls it in production, because nothing in this tree builds a
`LadderSnapshot` from a wrapper run. Search the workspace for the type: every
hit is `stella-protocol` (the definition and its wire tests) or `stella-cli`'s
`calibration`/`reward`/`dataset_cmd`, which *read* stored snapshots. There is
no producer on the wrapper socket path, so a wrapper run's verdict evidence
never reaches the record a fine-tuning pipeline or a replay reads.

The consequence is the one the stamp exists to prevent, one layer out: a stored
verdict cannot say whether a plugin's arbiter answered, abstained, or was never
heard from. `DispatchReport::arbitration` carries that for the length of the
process and nothing durable carries it after.

## The fix

Give the wrapper path a `LadderSnapshot` producer, then stamp it.

1. Decide what each `LadderSnapshot` field means for a wrapper run. The
   evidence the socket carries is `EvidenceSet` — a `FlipObservation`, a
   `TamperFinding` and a measurement map — which does not fill `diff_lines`,
   `new_diag_errors`, `mutating_actions` and several others. A field the
   wrapper path cannot observe must read as unobserved rather than as zero;
   `LadderSnapshot`'s own doc comments say which reading each field's absence
   carries.
2. Map `Verdict` and `Outcome` onto a `LadderRung`. `Verdict::Undecided` is
   `unverified` or `unverifiable` depending on the `UndecidedReason`, and the
   two are different claims — see `LadderRung`'s doc comment on why conflating
   them recreates the defect the enum exists to fix.
3. Hash the snapshot with the record-hash primitive (ADR 0004) over
   `LadderSnapshot::stamp_preimage`, and call `ArbiterClaim::into_stamp` with
   that hash and a real clock, once per row in `DispatchReport::arbitration`.
   The digest lives in `stella-core`; `stella-runtime` already depends on it.
4. Emit or store the stamped snapshot so a replay can read it.

## Files

- `crates/stella-protocol/src/ladder.rs`, `ladder/stamp.rs` (the types; no
[...]

Issue #5882, opened 2026-09-05
Two arbiters in one turn: the fold is written but the loop still asks the single-arbiter rule


macanderson:
## The problem

`crates/stella-runtime/src/wrapper/arbitration.rs` now holds `fold_stamps`, the
law for what a completion gate does when several arbiters disagree. It is
tested, and the dispatch loop records its output on
`DispatchReport::arbitration`. What the loop does **not** do is reach its
hold-or-stop decision through it. `WrapperDispatch::run` still calls `again`,
which decides for one arbiter.

That is correct today only because a second arbiter cannot exist:
`super::compose::compose` refuses two arbiter-grade members with
`WrapperError::TwoArbiters`, and a `--pipeline a,b` selection naming two of
them fails at bind. So the composition case the fold was written for is
unreachable through the shipping path.

Composition arrives anyway. A workspace witness plugin beside an org-mandated
policy plugin is the obvious pair, and refusing it at bind means a user with
both installed can run neither together.

## The fix

Decide whether two arbiters in one composition are allowed, and if so wire the
loop to the fold:

1. Relax or keep `WrapperError::TwoArbiters`. If relaxed,
   `crates/stella-runtime/src/wrapper/compose.rs` needs a real answer for
   `Composition::hold_grant` (today the arbiter's, or the first member's) and
   for `merge_rule`'s requirement union — its doc comment says plainly that the
   union cannot collide *because* two arbiters are refused, and names
   `tests/wrapper_composition.rs` as the test that pins both halves. That test
   fails first if the rule loosens, which is the signal it was written for.
2. In `WrapperDispatch::run`, build one `ArbiterClaim` per arbiter member and
   fold them, then map `Arbitration` back to `Continuation`/`Outcome`.
   `wrapper_arbitration.rs`'s `the_fold_agrees_with_again_on_a_single_arbiter`
   already proves the mapping is behaviour-preserving for one arbiter, so the
   existing dispatch tests are the regression net.
3. Track each arbiter's own `holds_spent` across rounds, not just the turn
   total. The loop keeps one `holds_spent` counter today; rule 5 needs one per
   arbiter beside it.

## Files

- `crates/stella-runtime/src/wrapper/dispatch.rs` (`WrapperDispatch::run`)
- `crates/stella-runtime/src/wrapper/compose.rs`
- `crates/stella-runtime/src/wrapper/arbitration.rs` (`fold_stamps`, unchanged)
[...]

PR #5884, opened 2026-09-05
feat(stella-runtime): arbitration law at the completion gate

### LINKED ISSUE #5121: plugin: arbitration law at the completion gate — multiple arbiters, and instrument failure as a recorded abstention rather than silent fail-open or costly fail-closed

macanderson:
## What and why

`judge` reads one plugin's evidence and `again` decides whether that one plugin
holds the turn open. Both assume a single arbiter. A workspace witness plugin
beside an org-mandated policy plugin is the obvious pair, and the host had no
written answer for what to do when they disagree. #3245 left two questions open
here, and this settles both.

**The law.** `crates/stella-runtime/src/wrapper/arbitration.rs` — `fold_stamps`,
pure and total, no clock and no process, the `ladder_decision` shape. It reads
`ArbiterClaim`s (`StampAssessment` from #5120's `VerdictStamp`, plus the two
facts the wire record leaves to the host: whether this grade may hold a turn
open, and what it has spent doing so) and returns an `Arbitration`: one row per
claim in arrival order, plus the turn's hold count and its ceiling.

Five rules:

1. A red test is never talked round. `Arbitration::refutes_done` answers `true`
   on `revise` however many claims say done.
2. One live `not_done` holds the turn. All must agree to pass; a majority is
   never enough.
3. A `done` claim never lifts a rung. `Arbitration::rung` is the rung handed in.
4. `inconclusive` and `not_applicable` get a row and count for nothing. The fold
   drops their clauses rather than trusting the caller, so an abstention cannot
   become a finding by arriving with a list attached.
5. Each arbiter's allowance is its own ask clamped to a host turn ceiling, and
   the turn total goes up by **one** per held round however many said no. Two
   arbiters taking turns cannot double the model calls a turn burns. A spent
   allowance completes the turn with every unmet clause reported, attributed to
   the arbiter that asked for it, against the ceiling that actually stopped it
   (`HoldStop::TurnAllowanceSpent` vs `HoldStop::ArbiterAllowanceSpent` — the
   row a single-arbiter fold could never produce).

**Fail open, and say so.** `ArbiterClaim::did_not_answer` turns a crash, a
timeout, or unreadable output into an `inconclusive` claim carrying the
wrapper's own error text, with `timed_out` set for `WrapperError::Timeout`
alone. It never holds — deny at a completion gate means the model keeps burning
steps against feedback it cannot act on. `WrapperDispatch::run` now records one
such claim per fault, attributed to the member that produced it, on the new
`DispatchReport::arbitration`; `stella-cli`'s wrapper report prints
`arbiter X did not answer — recorded as inconclusive, and nothing was held open`
for each. A run whose arbiter died no longer leaves the same trace as a run
whose arbiter was satisfied.

`ArbiterClaim::answered` is the field that keeps that line honest: a check that
returned `inconclusive` looked and could not tell, and a check that never
answered did not look. Only the second draws the line.

**`again` is untouched.** Rather than rewrite the live single-arbiter path, a
property proves `fold_stamps` and `again` agree on every one-arbiter input — so
the law is the shipped rule widened, not a second opinion.

## Witness mechanism

`crates/stella-runtime/tests/wrapper_arbitration.rs`. Every test names
`fold_stamps` or its vocabulary, none of which exists on `main`, so the whole
file fails to compile there — the feature is genuinely absent. Beyond that,
each test pins one property that could regress after the types exist:

- `a_deterministic_failure_survives_every_done_claim` — rule 1 over all seven
  rungs and any list of agreeing claims, plus rule 3 on the same inputs. It
  asserts both directions: `revise` refutes done, and no other rung does.
- `one_objection_holds_the_turn_however_many_agree` — rule 2 with the objector
  inserted at every position among any number of agreeing claims, and the turn
  spending exactly one round.
- `an_abstention_is_recorded_and_never_decides` — rule 4 over every assessment
  list. Every claim arrives carrying a clause, so dropping it is the fold's
  work rather than the caller's.
- `the_fold_agrees_with_again_on_a_single_arbiter` — over the grade, the ask,
  the spend and the ceiling: the whole input space `again`'s clamp is defined
  on, including the boundaries where a clamp goes wrong.
- `two_arbiters_draw_from_one_turn_total` — rule 5 as a table: both holding
  buys one round; one arbiter out of its own holds is reported against its own
  ceiling while the other still holds; a spent turn total completes with both
  reported against the turn's ceiling.
- `an_arbiter_that_did_not_answer_abstains_and_never_blocks` — over the real
  `WrapperError::Exit`, `Timeout` and `NoResponse` values a host catches, and
  the wire stamp `into_stamp` mints from each.
- `a_dead_arbiter_completes_the_turn_and_lands_on_the_record` — the whole-loop
  witness, driving `WrapperDispatch::run` against the in-tree subprocess
  fixture in `exit 7` mode. The turn still runs, every fault has an abstention
  attributed to `reference-v1`, nothing is held, and the arbiter's own claim
  from the final verdict rides last.
- `an_arbiter_that_did_not_answer_gets_a_line_of_its_own`
  (`crates/stella-cli/src/wrapper_plugin/tests/report.rs`) — the deck line, and
  the fact that the arbiter's *answered* abstention in the same report draws no
  such line.

## Exemplar

The neighbourhood, not an outside crate: `judge`/`again` in
`crates/stella-runtime/src/wrapper/verdict.rs` are the shape copied — free
functions, synchronous, total, I/O-free, over owned data, with the vocabulary
as closed enums beside them. `deterministic_failure` is written as a full match
over `LadderRung` so a rung added later is a build error rather than a silent
landing on one side of rule 1.

## What this PR does not do

The dispatch loop still reaches its hold decision through `again`, over the one
arbiter `WrapperError::TwoArbiters` permits. Making the loop hold on the fold's
own decision needs the second arbiter that error still refuses, and no host in
this tree mints a `LadderSnapshot` from the wrapper path, so nothing calls
`ArbiterClaim::into_stamp` in production yet. Both are filed below rather than
left unwired without a handoff.

Tests deleted: None.

Residue filed: #5882 (wiring the loop to the fold once two arbiters are allowed), #5883 (no ladder-snapshot producer on the wrapper path, so no stamp is stored).

Closes #5121

## Summary by Sourcery

Establish and record the completion-gate arbitration law for multiple arbiters while keeping unanswered arbiters fail-open and visibly distinct from answered abstentions.

New Features:
- Add a pure arbitration fold that combines multiple arbiter claims into ordered gate decisions, hold state, budget usage, and stop reasons.
- Record wrapper failures as attributed inconclusive arbitration claims and expose the resulting arbitration record in dispatch reports.
- Report unanswered arbiters explicitly in CLI wrapper output while distinguishing silence from answered inconclusive results.

Bug Fixes:
- Ensure arbiter failures do not hold completion turns open or appear equivalent to successful arbiter runs.

Enhancements:
- Define deterministic completion-gate rules for failures, objections, abstentions, rung preservation, per-arbiter allowances, and shared turn budgets.
- Preserve unmet requirements and their arbiter attribution when allowances or the shared turn ceiling end a turn.
- Provide conversion from host arbitration claims to wire verdict stamps, including timeout and answer-status metadata.

Tests:
- Add property-based and integration coverage for arbitration rules, single-arbiter compatibility with `again`, failure handling, shared hold budgets, dispatch recording, and CLI reporting.


macanderson:
## Answering Sourcery's two ❌ rows, and a DoD gap against #5121

Automated PR sweep. Both ❌ rows are accurate readings of this diff — neither is a misreading, and neither is fixed here. They are answered rather than fixed, with the handoffs already filed, and both trace back to a definition-of-done box on #5121 that cannot honestly be ticked yet.

### ❌ Row 2 — "the required persistent snapshot record is still missing"

Correct. `ArbiterClaim::into_stamp` exists and is unit-tested, but nothing calls it in production because nothing in this tree builds a `LadderSnapshot` from a wrapper run — every workspace hit for the type is either the `stella-protocol` definition or a `stella-cli` reader of stored snapshots. This PR's own "What this PR does not do" section says so, and the handoff is **#5883** (open, `triage`), which carries the field-mapping decisions (unobserved vs. zero; `Verdict::Undecided` → `unverified` vs. `unverifiable`) and the witness that a dead plugin leaves a stored `inconclusive` stamp whose hash re-derives.

Deliberately out of scope: the producer is a design decision about what every `LadderSnapshot` field *means* for a wrapper run, not a wiring step, and widening this PR to make that call would bury it.

### ❌ Row 3 — "no integration fixture demonstrating composed arbiters"

Correct, and it is currently unreachable rather than merely unwritten: `wrapper::compose::compose` refuses two arbiter-grade members with `WrapperError::TwoArbiters`, so a two-arbiter fixture cannot bind through the shipping path. The two-arbiter law is therefore proven against the pure fold (`two_arbiters_draw_from_one_turn_total`) and nowhere else. The handoff is **#5882** (open, `triage`), whose first DoD box is exactly the prerequisite decision — whether one composition may carry two arbiters at all.

### The DoD gap this implies (SCR-003)

#5121's three DoD boxes stand against this diff as:

| Box | State |
|---|---|
| Property tests over the fold, rules 1–4 as properties | **Met** — `a_deterministic_failure_survives_every_done_claim`, `one_objection_holds_the_turn_however_many_agree`, `an_abstention_is_recorded_and_never_decides` |
| Fixture arbiter dies mid-turn: turn completes, **snapshot carries the `Inconclusive` stamp**, deck line present | **Partially met** — `a_dead_arbiter_completes_the_turn_and_lands_on_the_record` proves the completion, the attribution and the deck line; the snapshot half is #5883 |
| Two-arbiter fixture where one vetoes: hold attributed, budget accounting matches rule 5 | **Not met** — #5882 |

So `Closes #5121` is not carried by this diff on the issue's own terms. **Suggested change, for the author to make rather than the sweep: `Closes #5121` → `Refs #5121`**, and let whichever of #5882/#5883 lands last close it. The alternative — narrowing #5121's DoD to what the fold alone can prove and re-pointing the other two boxes at the follow-ups — is equally defensible, but it is a scope decision, so it is proposed here and not pushed.

Nothing else on this PR needs action from the sweep: no merge conflict (`mergeable_state: blocked`, awaiting required checks, not `dirty`), and CI is still in flight on `f447c4e`.

---

macanderson:

Both are correct readings of the diff. Both are out of scope here for a
structural reason, and both are filed as handoffs rather than left unwired.

### Row 2 — "does not wire `ArbiterClaim::into_stamp` into production or store a `VerdictStamp` on a ladder snapshot"

Right, and it cannot: **nothing in this tree builds a `LadderSnapshot` from a
wrapper run.** Search the workspace for the type and every hit is
`stella-protocol` (the definition plus its wire tests) or `stella-cli`'s
`calibration` / `reward` / `dataset_cmd`, all of which *read* stored snapshots.
There is no producer on the wrapper-socket path at all.

A stamp's `preimage_hash` is defined as the ADR 0004 record hash over
`LadderSnapshot::stamp_preimage`. Minting one with no snapshot to hash would
mean inventing that hash, which is worse than not storing a stamp — a hash that
does not re-derive reads as tampering.

Building the producer is a design of its own: `LadderSnapshot` carries
`diff_lines`, `mutating_actions`, `new_diag_errors` and several more that the
socket's `EvidenceSet` (a `FlipObservation`, a `TamperFinding`, a measurement
map) cannot observe, and each one has a documented "unobserved" reading that is
not zero. Getting that wrong pollutes the reward-labelling path that reads these
snapshots.

Filed as **#5883**, with the field-by-field mapping, the `Verdict` →
`LadderRung` question, and a witness spelled out.

What this PR does deliver on that objective: the claim itself
(`StampAssessment::Inconclusive`, `timed_out` set for
`WrapperError::Timeout` alone, the wrapper's own error text in the summary),
`ArbiterClaim::into_stamp` as the tested minting function, the record on
`DispatchReport::arbitration`, and the deck line.

### Row 3 — "no integration fixture demonstrating composed arbiters through the completion gate"

Right, and it cannot exist today: **`WrapperError::TwoArbiters` refuses a
composition with two arbiter-grade members at bind time**
(`crates/stella-runtime/src/wrapper/compose.rs`), and
`tests/wrapper_composition.rs` pins that refusal. A two-arbiter dispatch fixture
would have to relax a shipped rule that was argued for on its own terms, which
is a decision this issue does not make.

That is the exact situation #5121 describes: *"composition arrives anyway… the
fold must be law before it is emergent behavior."* So the two-arbiter rules are
proved over the fold, which is where they are decidable today —
`two_arbiters_draw_from_one_turn_total` covers attribution, the per-arbiter
ceiling versus the turn's, and the one-hold-per-round accounting.

Filed as **#5882**: decide whether two arbiters may compose, and if so wire
`WrapperDispatch::run`'s hold decision to `fold_stamps` and add the composed
fixture. `the_fold_agrees_with_again_on_a_single_arbiter` exists so that wiring
is provably behaviour-preserving when someone does it.

Issue #5077, opened 2026-08-26
Two provenance grades have no producer, so three impact-class requirements govern nothing


macanderson:
## Problem

Two of the five `ProvenanceGrade` rungs have no producer anywhere in the
workspace, and three of the six `ImpactClass` floors sit on one of them. The
result is a policy whose middle is unreachable: nothing a session gathers can
ever grade `HumanReview` or `TrajectoryAbstraction`, so those two rungs neither
authorise anything nor block anything — they only shift where the real boundary
sits.

## The evidence

`observation_grade` (`crates/stella-core/src/context_record/provenance.rs`) is
total over the three `ObservationSource` variants and yields exactly two values:

```rust
ObservationSource::ToolOutcome => ProvenanceGrade::EnvironmentObservation,
ObservationSource::ReflectionLesson | ObservationSource::MemoryCitation => ProvenanceGrade::ModelCritique,
```

`EvidencePool::from_observations` in the same file is the only public
constructor and folds with `ProvenanceGrade::weakest` (a `min`), and
`ProposalRecord::new` (`crates/stella-core/src/context_record/lifecycle.rs`)
takes an `Option<EvidencePool>` and stores what it folds to. So a proposal's
`provenance` is always `None`, `ModelCritique`, or `EnvironmentObservation`.

`rg -n "ProvenanceGrade::HumanReview|ProvenanceGrade::TrajectoryAbstraction" crates/`
returns only the definitions, doc comments and tests — no construction site on
any production path.

Meanwhile `ImpactClass::required_grade`
(`crates/stella-protocol/src/provenance.rs`):

```rust
Self::PromptHint | Self::RecallBias | Self::AdvisoryRecord => ProvenanceGrade::TrajectoryAbstraction,
Self::SteeringDirective => ProvenanceGrade::EnvironmentObservation,
Self::BlockingGuard | Self::ExecutableTool => ProvenanceGrade::DeterministicProof,
```

## Why it matters

1. **Three impact classes floor at a grade nothing produces.** `PromptHint`,
   `RecallBias` and `AdvisoryRecord` are satisfiable only by an
   `EnvironmentObservation` pool — i.e. by clearing a *higher* bar than the one
   they declare. Their declared requirement governs nothing.
2. **A human sign-off cannot publish anything at all.** `HumanReview` is
   deliberately below `TrajectoryAbstraction`, and `authorises` checks evidence
   before authority, so no `PublicationAuthority` lifts a weak grade. There is
[...]

Issue #5890, opened 2026-09-05
check-rebase-replay reads a base-branch rename absorbed through a merge as an edit-and-undo pair


macanderson:
> **Duplicate of #5910**, filed 41 minutes later against the second instance of
> the same defect. Both are fixed by #5931, which closes both. The analysis
> below stands; #5910 carries the correction to the suggested fix.

## Problem

`scripts/check-rebase-replay.sh` fails a branch when a path appears in the
branch's history but not in its diff against the base. That is the #4979 shape —
an edit and an undo, which a rebase converts into a live revert.

It cannot tell that shape from a different one: **the base branch renamed the
file, and the branch absorbed the rename through a merge.** The path then leaves
the branch's final diff without anyone having undone anything, and the guard
reports a hazard that cannot occur.

Live instance: PR #5840, head `584aa6f9`, run
[33947893423](https://github.com/macanderson/stella/actions/runs/33947893423).

```
check-rebase-replay: FAIL — 1 path(s) edited and then undone inside this branch.
check-rebase-replay:   crates/stella-core/src/records/validate.rs
```

## Why it is a false positive, established rather than argued

#5829 moved the record plane out of `stella-core` into `stella-records`, so
`crates/stella-core/src/records/validate.rs` no longer exists on `main`. The
branch's three commits are:

```
584aa6f fix(stella-records): repoint the path matcher after the record plane moved crates
2fa98d2 Merge remote-tracking branch 'origin/main' into pr5840
f6aee54 fix(stella-core): report a trigger that names a path the other record guards
```

The deletion comes from `main`'s side of the merge, not from the branch:

```
$ P=crates/stella-core/src/records/validate.rs
$ git cat-file -e origin/main:$P   # 913e21a, the base   -> NO
$ git cat-file -e f6aee54:$P       # the branch's work   -> YES
$ git rev-parse --short 2fa98d2^1 2fa98d2^2
f6aee54   913e21a
$ git cat-file -e 2fa98d2^1:$P     # branch side of merge -> YES
$ git cat-file -e 2fa98d2^2:$P     # main side of merge   -> NO
```

The branch side has the file; `main`'s side does not. Nothing in the branch
undid an edit — `main` renamed the file underneath it.

**The work is intact**, which is the half that matters:

```
$ git diff origin/main..origin/fix/5780-cross-dimension-conflicts --stat \
    -- crates/stella-records/src/records/validate.rs
[...]

PR #5894, opened 2026-09-05
feat(stella-cli): ask the evidence gate before a rule is published

### LINKED ISSUE #4865: The evidence gate is declared but nothing calls it before publishing a skill, rule, or tool

macanderson:
## What

`stella_protocol::provenance::authorises` — the declared gate on every way
Stella changes itself — had zero production call sites. The three paths that
publish a **rule** now ask it before they write:

| Path | Authority | Result today |
| --- | --- | --- |
| auto-activation (`memory/learning.rs`'s `induce_rules`) | `Agent` | refused, printed |
| `stella proposals keep` (`proposals_cmd.rs`'s `materialize_directive`) | `LocalHuman` | refused, printed |
| `stella memory promote` (`memory_cmd.rs`'s `promote_in`) | `LocalHuman` | refused, returned as the command's error |

New module `crates/stella-cli/src/promotion_gate.rs` is the one place the
binary asks. It maps a published artifact to the `ImpactClass` the evolution
ledger declares for that surface, and hands back the protocol's own typed
`PromotionRefusal`. `rules_mining::write_rule` gained a `PublicationAuthority`
and now returns `RulePublication::{Written, AlreadyPresent, Refused}`, so
every caller has to render the refusal rather than fold it into "already
exists".

## Why, and what it changes for a user

**This turns off rule publication on the shipped default, deliberately.** A
rule steers whoever runs the next turn, so the ledger's `framework` row prices
it at `EnvironmentObservation`. A reflection lesson grades `ModelCritique`
(`observation_grade`) and `EvidencePool` folds to its weakest member, so
nothing the miner produces reaches that bar. The gate says no and names both
grades; the proposal stays in the ledger and stays reviewable.

That is issue #4865's own finding, resolved as **option 1: hold directives to
measured evidence**. One correction to the issue's wording: the fold is a
`min`, so "at least one `ToolOutcome`" cannot be the rule — *every* supporting
observation has to be one. The code enforces the stronger form because that is
what the fold means.

The docs for both user-facing commands say so now
(`website/content/docs/commands/{memory,proposals}.mdx`).

## Witness

`crates/stella-cli/src/memory/rules_mining/tests.rs`:

- `a_rule_on_reflection_evidence_is_refused_and_nothing_lands` — `write_rule`
  offered `ModelCritique` returns
  `Refused(EvidenceTooWeak { required: environment_observation, actual: model_critique })`
  and the workspace loader finds nothing.
- `a_rule_with_no_grade_at_all_is_refused_as_absent` — the migration posture
  for a pre-provenance proposal: `NoEvidence`, and no authority argument
  changes it.
- `mined_rules_land_where_the_loader_reads` (the ledger row's own witness)
  still passes on `EnvironmentObservation`, so the gate is not a blanket no.

**Fails on the old code by construction:** the old `write_rule` had no gate,
no authority argument, and no refusal to return — it wrote the file for both
grades and returned `Ok(Some(path))`.

At the loop level,
`memory/learning/guarantees.rs::auto_activation_is_refused_when_the_evidence_is_only_reflection_prose`
runs the real reflection loop twice over a five-task corpus and asserts no
`MemoryPromoted` event, no `System` promotion in the ledger, and no
`.stella/rules` directory. `memory/replay/tests.rs::auto_activation_is_refused_on_reflection_evidence_and_clobbers_nothing`
does the same through the replay harness, and carries the #737 no-clobber
assertion in the same run.

`crates/stella-cli/src/promotion_gate/tests.rs` pins the policy edges and
reads `stella-parity`'s ledger source to check that this module's impact table
and the `framework` row still say the same word — the same `include_str!`
discipline the ledger uses for its own witness names.

## Exemplar

`stella-store`'s `EnableAuthority::established_authority` is the shape copied:
store the fact the path saw, derive the grade from the one policy, never keep
a second copy of the policy beside it. `promotion_gate` follows it — the
ledger row is cited, not linked, and a test checks the citation.

## Tests deleted

Named as `check-deleted-tests` asks. All three are renames or merges, and each
one's property is asserted somewhere else in this diff:

- `an_auto_activated_rule_is_advisory_and_never_blocking` and
  `auto_activation_never_clobbers_an_existing_record`
  (`memory/replay/tests.rs`) merged into
  `auto_activation_is_refused_on_reflection_evidence_and_clobbers_nothing`.
  The advisory-on-disk half moved to
  `rules_mining::tests::a_written_rule_file_is_prompt_only`, which can still
  publish a rule; the no-clobber half is asserted in the merged test and in
  `writing_never_clobbers_an_existing_rule_file`.
- `the_committed_corpus_builds_a_memory_a_skill_a_rule_and_a_tool` renamed to
  `..._builds_a_memory_a_skill_and_a_tool`, asserting `rules.is_empty()`.
- `an_auto_activated_rule_announces_a_promotion_the_ledger_holds` and
  `a_rule_already_on_disk_is_not_promoted_a_second_time`
  (`memory/learning/guarantees.rs`) merged into
  `auto_activation_is_refused_when_the_evidence_is_only_reflection_prose`,
  which runs the loop twice and asserts zero announcements on both passes.
- `a_well_evidenced_rule_activates_and_reaches_the_loader` deleted: a rule
  mined from reflection prose is no longer well-evidenced. A comment at its
  old site names the two tests that now hold its assertions.
- `a_published_rule_with_no_grade_writes_none`
  (`memory/rules_mining/tests.rs`) renamed to
  `a_rule_with_no_grade_at_all_is_refused_as_absent`. It asserted that a rule
  published with no grade writes no `evidence_grade` key. The gate refuses
  that rule before the write, so there is no file left to inspect; the
  replacement asserts the refusal names `NoEvidence` and that no authority
  changes it, which is the same claim about absent evidence at the only point
  it is now observable.

## What this PR does not do, and why

The issue's DoD names two more publication paths. Both need a decision this PR
is not the place to make, and both are filed:

- **Skill creation.** Under the shipped default a skill is only written on a
  measured lift, so the gate would be a no-op there. The only mode that
  publishes on mined evidence is bootstrap
  (`context.promotion.skill.require_measured_lift = false`), which is also
  what the replay harness runs. Gating it needs a producer for
  `TrajectoryAbstraction`, which nothing constructs today (#5077).
- **Foundry tool adoption.** `ImpactClass::ExecutableTool` requires
  `PublicationAuthority::LocalHuman`, and
  `EnableAuthority::{FlagAssertion, Autonomy, Rollback}::established_authority`
  is `Agent`. Asking the gate there would refuse `stella tools --enable --yes`
  and the whole `foundry.autonomy = "auto"` pipeline, which is a shipped
  designed path.

DoD item 5 (`stella context explain` shows the grade) is already satisfied on
`main`: `context_cmd/explain.rs`'s `evidence_line` renders
`provenance.evidence_grade`, which `inferred_rule_record` writes.

Closes #4865

## Residue filed

- **#5895** — skill creation never asks the gate, and the grade it would need
  (`TrajectoryAbstraction`) has no producer (#5077). Includes the deadlock to
  avoid: gating before `decide_auto_creation` stops `queue_candidate`, so no
  skill could ever earn a measured lift.
- **#5896** — foundry enable cannot pass the gate as the policy stands.
  `EnableAuthority::{FlagAssertion, Autonomy, Rollback}` prove only `Agent`,
  and `ExecutableTool` requires `LocalHuman`, so wiring it refuses the shipped
  autonomy pipeline. Three ways out, and it needs a maintainer to pick one.
- **#5897** — found while checking whether the admitted grade is reachable
  end to end: it is not. `tool_outcome_observation` writes one `source_ref`
  per tool rather than per occurrence, and `induce_rule_proposals` resolves its
  evidence back through that field with `find`, so five failures across five
  tasks collapse to one distinct task and the proposal is never eligible. The
  type's own doc comment says the shape should be `tool:<name>#<n>`.

## Summary by Sourcery

Enforce the evidence gate before publishing rules so reflection-derived directives remain reviewable until supported by measured observations.

New Features:
- Add evidence-gated publication for rules across automatic activation, proposal keeping, and memory promotion paths, with typed refusals and user-facing explanations.

Bug Fixes:
- Prevent rules based on reflection-only or missing evidence from being written or announced as promotions, while preserving existing files and reviewable proposals.

Enhancements:
- Centralize artifact impact mapping and promotion-gate handling for rule publication, distinguishing written, already-present, and refused outcomes.
- Update the evolution ledger and lifecycle provenance documentation to reflect enforced rule evidence requirements.

Documentation:
- Document that rule publication requires measured evidence and that rejected memory or proposal promotions write nothing.

Tests:
- Add coverage for evidence refusals, admitted measured evidence, stable repeated refusals, no-clobber behavior, and updated replay and auto-activation expectations.

Chores:
- Defer evidence-gate integration for skill creation and foundry tool adoption to follow-up work because their required evidence or authority policies are not yet available.


macanderson:
## Answering the Sourcery ❌ row

**Row:** "Invoke the evidence gate on every live publication path for skills, rules, and foundry tools before writing or adopting the artifact." — ❌ "the PR wires the gate into the three rule-publication paths, but explicitly does not add callers for skill creation or foundry tool adoption."

The diff is read correctly; the objective it is graded against is not what #4865's definition of done asks of this PR. That DoD splits the publication paths into three deliberately, and asks the skill and foundry halves to be **decided and filed**, not wired:

- [x] `authorises` is called on every rule publication path: auto-activation, `stella proposals keep`, and `stella memory promote`.
- [x] Skill creation's half is decided and filed rather than left silent — #5895.
- [x] Foundry tool adoption's half is decided and filed rather than left silent — #5896.

Both issues are filed, each carrying the decision it is blocked on rather than a note to self. #5895 names the missing producer: the ledger's `skill` row prices at `TrajectoryAbstraction` and nothing in the tree constructs that grade (#5077), so wiring the gate there would be a no-op on the shipped default and would refuse every skill in bootstrap mode — which is the mode the replay harness runs, so the whole harness would stop building skills. #5896 is sharper still: the `tool` row requires `LocalHuman`, and three of the four paths `EnableAuthority::established_authority` recognises establish only `Agent`, so asking the gate there today refuses `--yes` in scripts, refuses the autonomy pipeline the row documents as shipping on purpose, and reddens that row's own witness. Both need a maintainer's answer before any wiring, and picking one silently is the expedient CLAUDE.md forbids.

Wiring either half inside this PR would mean lowering a requirement until the code agrees. Leaving them unfiled would be the untracked half-finished work the rule exists to prevent. They are filed.

## Also pushed: a merge conflict, and a latent compile break under it

`main` moved the record plane out of `stella-core` (#5113/#5117) while this branch was open. `rules_mining.rs`'s import block conflicted and is resolved to take both sides — main's `stella_records::context_record` path and this branch's widened `provenance::{PromotionRefusal, ProvenanceGrade, PublicationAuthority}`.

`learning/guarantees.rs` is the part worth flagging: git reported **no** conflict there, because both `stella_core::context_record::{RecordProposalKind, PromotionActor}` references sat in hunks only this branch touched. The auto-merged tree was textually clean and would not have compiled. Repointed at `stella_records`; `cargo check -p stella-cli --all-targets` is clean (exit 0, no errors, no warnings).

One inherited item left alone: `crates/stella-protocol/README.md` still cites `stella_core::context_record::hash`. That citation is stale on `main` already, not something this branch introduced, and #5889 is repointing it — so it is tracked rather than absorbed here.

---

macanderson:
## `no edit is undone inside the branch` is red on `5aac868`, and it is #5890's false positive

Naming it rather than leaving it unexplained, since my merge commit is what made it fire.

```
check-rebase-replay:   crates/stella-core/src/context_record/lifecycle.rs
```

**A rename, not an undo.** #5829 moved the record plane out of `stella-core`, so that path does not exist on `main` at `ffcda37`. This branch edited it before that landed; my merge absorbed the rename. The path therefore leaves the branch's final diff with nobody having undone anything — which is exactly the shape #5890 filed, from the same #5829 rename against #5840's `crates/stella-core/src/records/validate.rs`. This is a second live instance of it.

**The edit survived the move**, which is the half worth checking rather than assuming:

```
$ git diff --stat ffcda37 origin/fix/4865-wire-evidence-gate -- '*context_record/lifecycle.rs'
 crates/stella-records/src/context_record/lifecycle.rs | 8 ++++----
 1 file changed, 4 insertions(+), 4 deletions(-)

$ git cat-file -e ffcda37:crates/stella-core/src/context_record/lifecycle.rs    -> gone from main
$ git cat-file -e ffcda37:crates/stella-records/src/context_record/lifecycle.rs -> present on main
$ git log --oneline ffcda37..origin/fix/4865-wire-evidence-gate -- crates/stella-core/src/context_record/lifecycle.rs
   (no commits)
```

Nothing on the branch touches the old path in that range; git's rename detection carried the edit to the new one. Reproduced locally with `./scripts/check-rebase-replay.sh ffcda37 origin/fix/4865-wire-evidence-gate`, exit 1.

And #4979's mechanism cannot fire here for the reason #5890 gives: that hazard needs the *same edit* already upstream so `rebase` drops the branch's copy and keeps the undo. There is no duplicate edit upstream and no branch-authored undo — a rebase would replay the edit onto a tree with no such path and conflict loudly rather than revert silently.

## Why I am not applying the printed remedy

The guard prints `git rebase -i` / `git reset --soft && git commit`, and by design it offers no acknowledgement escape — its header argues that flattening "loses nothing", which is true of the shape it was built for and not of this one.

Both forms require a force-push to `fix/4865-wire-evidence-gate`, a branch I did not create. I am not rewriting history on someone else's branch: a peer session may hold a checkout of it, and a merge commit is what keeps that valid. #5890 reaches the same conclusion independently — it asks an author "to rewrite history to silence a hazard that is not there — and on a shared branch that is its own risk", and makes "#5840 passes the guard without its history being rewritten" a definition-of-done item.

**This does not block the merge.** #4997 — "The rebase-replay guard reports but does not block, because it is not a required check" — is open precisely because this check is not in `main`'s required contexts. I am reporting that from those two open issues rather than from branch-protection settings, which I did not read; if it has since become required, this PR needs the branch's owner to flatten it, or #5890 fixed first.

Everything else on this PR stands: the merge conflict is resolved, `cargo check -p stella-cli --all-targets` is clean at exit 0, and the Sourcery ❌ is answered above.

---

macanderson:
## `main is not known-broken` is red on `075604d`, and it is not this PR's

Standing down on it rather than silently, per the rule that a stand-down is never silent.

The check fails because **#5901 is open**: `main` does not compile (`crates/stella-core/tests/published_records_do_not_collide.rs` still imports `stella_core::{ingest,records}`, which #5829 moved to `stella-records`). `main-red-hold.yml` asks the tracker, not the tree, so it is red on every open PR right now — that is the design, and it is what stops PRs merging onto a broken base the way four did on 2026-08-19.

**The fix exists and is in flight: #5906**, open with the `unblocks-main` label. #5907, #5908 and #5909 are the same diff, closed as duplicates — five independent repairs of one break, which is #5224's shape again.

**Porting that fix into this PR would not clear the check**, which is the non-obvious part and the reason I have not done it. The usual remedy for a not-mine failure is to port the fix so it no-ops once the base carries it; here the check reads whether a `main-red` issue is open, so carrying the repair in this branch changes nothing it looks at. The only thing that clears it is the `unblocks-main` label, and that belongs to the PR that actually repairs `main` — putting it on this one would be using the escape hatch to hide a base failure rather than to let a repair through. So this PR waits.

No re-run spent: a re-run cannot pass while #5901 is open.

**What clears it:** #5906 merges, the canary's next green run closes #5901 by itself, and this check goes green here with no action. The build and test checks on this PR should clear at the same moment, since they run against a merge with `main` and inherit its compile failure.

Still outstanding and separately explained above: `no edit is undone inside the branch`, the #5890 false positive. Fixed since that comment: `check-prose` and `cargo fmt --check`, both in `075604d`.

---

macanderson:
## `dod / dod`: the cause is fixed; the check needs an event I cannot fire

It failed on #4865's seven unchecked DoD boxes. I verified all seven against this branch at `075604d` — the code and a real test run, not this PR's description — and ticked them, with the per-item evidence recorded [on #4865](https://github.com/macanderson/stella/issues/4865#issuecomment-5550215332). Summary: the gate is reached from all three named paths and calls `authorises`; #5895 and #5896 are filed with the decisions they are blocked on; a refusal reaches a human on each path; the `framework` row re-points its witness and states the enforced bar; `cargo test -p stella-cli memory::rules_mining` is `14 passed; 0 failed` with both directions of the gate green; and `explain.rs` renders `provenance.evidence_grade`.

`scr-dod-check.mjs` reads the linked issue **live**, so the ticks take effect with no new commit — but the check still has to *run*, and only a `pull_request` event starts it. Editing an issue fires none.

I cannot force one:

- `rerun_failed_jobs` on run `33950675708` → `403 Resource not accessible by integration`.
- An empty commit to kick CI is forbidden, and there is no code change left to make — the boxes were the whole failure.

So this clears on the next `pull_request` event, and one is already owed: `main` does not currently compile (#5901, repair in flight as #5906), so this branch needs a merge from a repaired `main` before it can go green anyway. That push fires the event and re-runs `dod / dod` against the now-ticked boxes. I am keeping the PR watched until then rather than leaving it to someone else.

If you want it green sooner, any edit to this PR's description does it — that is the trigger this repo already uses for exactly this, and it is the trigger, never the fix.

---

macanderson:
`no edit is undone inside the branch` is red here, and it is a **false positive** — the same one as #5840, tracked in **#5910**. Do not flatten this branch; nothing is wrong with it.

### What the guard flagged

```
  crates/stella-core/src/context_record/lifecycle.rs
      8d9f25bb6  feat(stella-cli): ask the evidence gate before a rule is published
```

### Why the hazard cannot occur

Against the merge base (`98af1d6`, now `main`'s tip):

| | `crates/stella-core/src/context_record/lifecycle.rs` |
|---|---|
| at base | **absent** |
| at head | **absent** |

The path exists at neither end, because `context_record` moved to `stella-records` in the record-plane extraction (#5829). It lives at `crates/stella-records/src/context_record/lifecycle.rs` now.

#4979's hazard needs a path that exists upstream carrying somebody else's edit, so a rebase can drop your copy and keep your undo as a live revert. There is no upstream copy of this path to revert.

What happened here is a rebuild, not an undo: commit `8d9f25bb6` edited the file at its old path, `main` moved it, and the branch followed. Its net diff confirms the work is present at the new location:

```
 crates/stella-records/src/context_record/lifecycle.rs | 8 +-
```

A rebase onto today's `main` would make `8d9f25bb6` fail to apply against a file that is not there — a conflict, which is loud, not the silent clean-merging revert the guard exists to catch.

### Why this matters beyond this PR

This is the **second** instance today, both from the same cause. #5840's flagged path was `crates/stella-core/src/records/validate.rs`; this one is `.../context_record/lifecycle.rs`. Both were moved out of `stella-core` by #5829, and both branches correctly followed the move.

So it is not a one-off: **any crate-level file move will fire this guard on every open branch that had edited a moved path**, and the prescribed remedy — flatten and force-push — would be paid for nothing each time. #5910 carries the fix (skip a suspect absent at both `$base` and `$head`, which is exactly the condition that makes the move case safe) and now cites both instances.

Nothing else here is failing on its own account.

---

macanderson:
**`no edit is undone inside the branch` — a fix is open, so do not flatten this branch.**

The red here is the guard misreading an upstream move as an undo (#5910). `crates/stella-core/src/context_record/lifecycle.rs` is absent at both the merge base and this head because the record-plane extraction relocated it, and this branch absorbed that through a merge — so no upstream copy exists for a replayed undo to revert, and the old commit would conflict on rebase rather than apply in silence.

#5932 fixes the guard. Verified against this exact head, run at `origin/main`:

```
old guard, PR #5894 head (c2e66fe) -> exit 1   ← the red in CI today
new guard, PR #5894 head (c2e66fe) -> exit 0
```

No history rewrite needed on this branch, and the prescribed flatten would have bought nothing.

One correction worth flagging, since #5910 proposed a different condition: skipping every suspect *absent at both ends* would have waived the suite's own R4 — a file added and then dropped inside a branch has the same signature and is a real hazard. #5932 keys on **who** removed the path (a delete in the branch's own non-merge history) rather than on absence alone, so R4 stays red.

---

macanderson:
Two checks are red here, and they want opposite things. **Do not follow the flatten advice** — one of them is a false positive, and the printed remedy would charge this branch a history rewrite for a hazard that is not present.

## `no edit is undone inside the branch` — false positive, no action needed

The guard names `crates/stella-core/src/context_record/lifecycle.rs`, edited by `8d9f25b` and absent from the branch's diff against `918b996f3`. That signature is real, but the cause is not an undo. The record-plane extraction moved the file:

```
$ git ls-tree origin/main crates/stella-core/src/context_record/lifecycle.rs
(nothing — the path does not exist on main)

$ git ls-tree -r --name-only origin/main | grep context_record/lifecycle
crates/stella-records/src/context_record/lifecycle.rs
```

The branch absorbed that move through its merges, and the edit rode across with it — it is alive at the new path, not reverted:

```
$ git diff --name-status 918b996f3 origin/fix/4865-wire-evidence-gate | grep context_record/lifecycle
M	crates/stella-records/src/context_record/lifecycle.rs
```

and the hunk is `8d9f25b`'s own doc-comment rewrite, the one that drops "once that gate is wired into the publication paths (#4865 tracks the wiring)" now that this PR wires it. The work survived the rename. Nothing was undone, and a rebase has no branch-authored removal to keep.

This is the exact case #5931 fixes — a suspect is dropped when it is gone at head *and* no non-merge commit of the branch's own deleted it. Its own AGENTS.md note names this cause: "The record-plane extraction is what forced this: it fired the guard on every open branch that had correctly followed the move." That PR is green and now out of draft; once it lands, re-running this check clears the row with no change here.

Flattening would have rewritten this branch's history to fix nothing.

## `gate steps no other workflow runs` — real, and yours

`check-prose` fails on the readability ratchet, and it is genuinely over:

```
crates/stella-cli/src/memory_cmd.rs: grade 10.80 allowed, 10.86 found
    [25.8] Those pairs are still in every workspace that ran the old code, and the lineage migration deliberately did not…
    [24.9] List memories ranked by citation count, with average usefulness, truthfulness rate, and rule-promotion eligibi…
    [23.8] The memory's stable id (nod_…) as shown by Optional note on why it is still needed Reclaim space in .stella/pr…
```

0.06 over the ceiling, on three long sentences. Splitting any one of them clears it. Per the guard's own instruction, the fix is to shorten the prose — not to add a line to `scripts/prose-grade-baseline.txt`.

That is the only check on this PR standing between it and green; `main is not known-broken` is the separate org-wide hold from #5939, which #5940 clears.

---

Issue #5830, opened 2026-09-05
stella-records should not depend on stella-core: lift the rule parser and the shared text helpers


macanderson:
`stella-records` depends on `stella-core`, and the arrow points the wrong way.
The record plane sits above the engine and reaches back into it for two things:

- `stella_core::rules::{Rule, RuleFile, RuleGuard, rule_from_file_checked,
  RuleFileError}`, used by `crates/stella-records/src/records/registry.rs` and
  `crates/stella-records/src/records/bridge.rs`. The registry merges markdown
  rules and TOML records under one precedence order, so it has to parse a rule
  file.
- `stella_core::steering::{SteeringCandidate, DroppedCandidate,
  SteeringSource}`, used by `crates/stella-records/src/adapt.rs`.
- `stella_core::glob::match_glob` in `records/select.rs` and
  `stella_core::redact::redact_secrets` in `records/validate.rs`.

That edge means the record plane links `tokio`, `async-trait`, `futures-util`
and `rand` to reuse a markdown parser and four value types, and it blocks
`stella-observatory` from ever dropping the engine out of its tree.

## Why it was not fixed in #5829

`rules` cannot simply move with the plane: `crates/stella-core/src/rules.rs`
calls `crate::glob::match_glob` and `crate::mining::{cluster_observations,
representative_text, already_captured, slugify, hash8}`, and `mining` is also
used by `crate::skills`, which is engine code. Moving `rules` therefore drags
`glob` and `mining` along and leaves `skills` reaching up into the record
crate — the same inversion one layer over.

## The fix

Extract the shared primitives into a crate both sides can take, likely two
of them:

- The markdown frontmatter splitter (`rules::{Frontmatter, parse_frontmatter}`),
  which `skills.rs`, `skills/invoke.rs` and `extensions.rs` also use.
- `glob::match_glob` and the `mining` clusterer, both pure functions over
  borrowed text with no engine concepts in them.

Then `rules` can live with the record plane, or stay in `stella-core` and be
read by neither. Decide which by asking whether a rule file is engine input or
record-plane input; the registry doc argues it is the latter.

The steering candidate types are the smaller half: they are three plain
structs and could move to `stella-protocol` or to `stella-records` itself,
whichever leaves fewer crates naming them.

## Files

- `crates/stella-records/Cargo.toml` (the `stella-core` dependency and the
  comment stating why it exists)
- `crates/stella-records/README.md` (Boundary section, which names this issue's
  subject)
[...]

Issue #5113, opened 2026-08-26
Epic: finish the re-layering — stella-core holds only the step path (the half of doc:plugin-architecture that never shipped)


macanderson:
## Origin and gap-check

`doc:plugin-architecture` proposed two things measured at v0.9.40: (1) an
out-of-process plugin boundary, and (2) a re-layering that evicts every
`stella-core` module not reachable from the step path. Gap-checked against
`main` today (v0.9.243 era):

| Proposal half | Status on main |
|---|---|
| Plugin boundary (§5–§8) | **Superseded by shipped work, mostly for the better** — `crates/stella-plugin` (manifest ladder, wrapper socket, host-run oracle, wire corpus + schema gates, capability consent, host_call), `stella-pipeline` deleted outright (#3865), #3245 closed |
| Re-layering (§1–§2, invariant #11) | **Not started.** `crates/stella-core/src/` still holds `records/`, `context_record/`, `ingest/`, `skills/`, `rules/`, `comparison/`, `mining/`, `redact/`, `self_tuning.rs`, `search.rs`, `scoreboard/`, `starvation.rs`, `repair.rs`, `goal.rs`, `tool_foundry.rs`, `extensions.rs`, `tasks.rs` — verified in the tree listing today |

The v0.9.40 measurement (re-measure before cutting; the shape will have
drifted, the conclusion will not): ~72.4k LOC in `stella-core`, ~51% test,
only ~17.7k production LOC reachable from `driver`/`step`/`ports`. The
largest single resident (the record plane, ~14.8k LOC) attaches to the step
path by **one** call — `receipts.rs` → `context_record::hash::record_hash` —
and `record_hash` is already consumed cross-crate from `stella-context`
(`store/ledger.rs`) and `stella-cli` (`memory/*`), i.e. the record plane
already behaves like a separate crate that happens to live in the engine.

## Why now

Every plugin/wrapper design decision is now being made against "the engine"
as a boundary (`doc:pipeline-as-plugins`, #3903 roleless-core). That boundary
is only describable in one sentence per seam if the crate behind it holds
only seams. Today it holds a record store, a learning plane, and five
CLI-only modules; every one of them is surface area a plugin author, an
embedding host, and the roleless-core epic have to be told to ignore.

## Children

- [x] `check-core-reachability` ratchet + the new architecture invariant — `scripts/check-core-reachability.py` is in the tree and AGENTS.md carries invariant 12, "`stella-core` holds the step path".
[...]

PR #5899, opened 2026-09-05
refactor(stella-learn): extract the learning plane from the engine

### LINKED ISSUE #5118: re-layering: extract stella-learn — skills/rules/comparison/mining/redact/self_tuning leave the engine; skills/invoke stays
Child of #5113. The learning/tuning plane (~8k LOC at v0.9.40 — re-measure)
is mutually coupled to itself and to the record plane, and none of it is on
the step path except one file.

## What to do

1. New crate `stella-learn`, depending on `stella-records` (sibling issue —
   lands first) and `stella-protocol`.
2. Move: `skills.rs` + `skills/*` **except `skills/invoke.rs`**, `rules.rs` +
   `rules/*`, `comparison.rs` + `comparison/*`, `mining.rs` + `mining/*`,
   `redact.rs` + `redact/*`, `self_tuning.rs`.
3. **Re-verify the `skills/invoke.rs` exception before cutting.** At v0.9.40
   the only production on-path reference to `crate::skills` was
   `restore.rs:61` (invoke on restore); everything else was doc comments and
   `#[cfg(test)]`. Re-run that sweep on today's tree — if restore no longer
   calls it, the whole of `skills/` moves and the exception dies. If it
   still does, `invoke.rs` stays in core with a narrow type boundary
   (an `InvokedSkill` value, not the skills catalog).
4. `redact.rs` needs a consumer audit: if the wrapper/plugin plane
   (`stella-plugin` disclosure scrubbing, hook payload redaction) consumes it,
   it may belong in a leaf both can reach rather than in `stella-learn` —
   decide from the actual `rg` results, not the tier table.

## AGENTS.md correction (part of this issue, not a follow-up)

The crate table row that routes "skills / rules" to `stella-core` was written
when the modules were smaller and now describes a fait accompli rather than a
boundary. The PR that lands this edits that row. (Same discipline the
proposal doc demanded: a doc that contradicts the tree is a bug in whichever
one is wrong.)

## Witness

Reachability baseline shrinks by the moved trees; `stella-learn`'s own test
suite passes; a `stella-core` build no longer compiles the skills catalog,
measured by `cargo build -p stella-core --timings` file count dropping
accordingly.

## Cross-ref

#3903 (roleless-core) removes role *vocabulary* from core; this removes the
learning *plane*. If #3903's slices touch `skills/` first, rebase this on
whatever survives — module residency is this issue's concern, not naming.

---
**Suggested worker:** `model-tier:pro` · **Estimated autonomous time:** ~90 min

## Definition of done

- [x] The reachability baseline shrinks by the moved trees.
- [x] `stella-learn`'s own test suite passes.
- [x] A `stella-core` build no longer compiles the skills catalog, measured by `cargo build -p stella-core --timings` file count dropping accordingly.

macanderson:
## What

The learning plane leaves `stella-core` as a new crate, `stella-learn`, and
two primitives the engine shares with it go down to `stella-protocol`.

- `crates/stella-core/src/{skills,rules,comparison,mining,redact,self_tuning}`
  → `crates/stella-learn/src/`, tests and proptest regressions included.
- `crates/stella-core/src/glob.rs` → `crates/stella-protocol/src/glob.rs`.
- `rules.rs`'s `Frontmatter` / `parse_frontmatter` / `strip_matched_quotes`
  and their seven tests → `crates/stella-protocol/src/frontmatter.rs`.
  `stella_learn::rules` re-exports the first two under the names rule
  callers already use.
- `crates/stella-core/src/skills/invoke.rs` →
  `crates/stella-core/src/skill_invocation.rs`. It stays: it is the one part
  of the plane the step path reaches.
- `steering/adapt.rs`'s `skill_candidates` / `skill_drops` / `skill_why` and
  their three tests → `crates/stella-cli/src/memory/steering.rs`. That file
  is now empty of everything else, so `crates/stella-core/src/steering/`
  loses it.
- 69 consumer files repointed across `stella-cli`, `stella-records`,
  `bench/loop-bench`, and prose in `stella-store`, `stella-tui`,
  `stella-observatory`, `stella-plugin`.

`crates/stella-core/Cargo.toml` gains no dependency. `stella-learn` depends
on `stella-protocol` and nothing else in the workspace.

## Why the shape differs from the issue

The issue says `stella-learn` depends on `stella-records`. It is the other
way round on today's tree, and it had to be.

1. **`stella-records` needs `stella-learn`, not the reverse.**
   `records/registry.rs` and `records/bridge.rs` read
   `rules::{Rule, RuleFile, RuleGuard, rule_from_file_checked, RuleFileError}`
   and `records/validate.rs` reads `redact::redact_secrets`. Nothing in the
   move set names a record type. So `stella-records` takes those two edges
   from `stella-learn` and drops them from `stella-core`. That is exactly the
   lift #5830 asked for: *"the fix is to lift the rule parser into a crate
   both sides can take."*
2. **This lands after #5117.** With the record plane still inside
   `stella-core`, moving `rules` and `redact` out would have forced
   `stella-core` → `stella-learn`, which is the edge this epic exists to
   remove. #5829 merged first; this rebases onto it.
3. **The skill adapters went to the CLI, not to `stella-learn`.**
   `skill_candidates` / `skill_drops` read `crate::skills` and return
   `stella_core::steering::SteeringCandidate`, so taking them into
   `stella-learn` would have re-created the `stella-records` → `stella-core`
   edge #5830 is filed to undo. `adapt.rs`'s own header already named the
   better home: *"the adapter sits with the plane implementation in
   `stella-cli::memory::steering`."* They sit beside the recalled-frame
   adapter now, and their only production caller (`memory::recall`) is one
   module away.
4. **`invoke.rs` is now `skill_invocation.rs`.** The issue asks for the
   exception to be re-verified, and it holds: `driver/restore.rs` mints an
   invocation message, `receipts.rs` block-types on the marker prefix, and
   `engine_markers.rs` lists that prefix. Those are step-path callers. Its
   old path made `stella-core` keep a module called `skills` that is not the
   skill plane, so it was renamed to say what it is.
5. **`redact`'s consumer audit** (the issue's step 4) found `stella-cli`
   (`export`, `dataset_cmd`, `memory/{observations,evidence,replay}`) and
   `stella-records::records::validate`. No `stella-plugin` or hook-payload
   consumer. Both live above `stella-learn`, so it belongs here rather than
   in a lower leaf.

## No `pub use` shims

A shim in `stella-core` would need `stella-core` → `stella-learn`, and the
issue's own definition of done is that a `stella-core` build stops compiling
the skill catalog. Every consumer is repointed instead. The rename is
mechanical (`stella_core::{skills,rules,comparison,mining,redact,self_tuning}`
→ `stella_learn::…`, `stella_core::glob` → `stella_protocol::glob`), and no
grouped import mixed a moved name with an unmoved one, so no import line
needed splitting.

## Two primitives moved into `stella-protocol`

`match_glob` has three callers now — `stella_learn::rules`,
`stella_core::hooks`, `stella_records::records::select` — and
`parse_frontmatter` has four, two on either side of the new boundary. A copy
per crate is a copy that drifts, and both are pure, total, and depend on
nothing outside the standard library. `stella-protocol` already hosts
`tokens::estimate_tokens` on that footing, and #5117 put `hash::record_hash`
there for the same reason. `crates/stella-protocol/README.md` states the
narrower rule those three sit on, rather than leaving "zero logic" reading as
if it still covered the crate whole.

## Witness

`crates/stella-learn/tests/selection_bytes_survived_the_move.rs` —
`one_prompt_through_the_moved_plane_renders_the_bytes_the_engine_used_to_render`
parses the `reviewer` fixture skill, selects it against the prompt
`review the database migrations` under the default `SelectionConfig`, and
asserts `render_skills_section` returns

```
\n## Applicable skills (selected for this task — apply the relevant ones)\n\n### reviewer\ndatabase review\n\nALWAYS_REVIEW_DATABASES\n
```

That literal is not invented: it is lifted out of `stella-cli`'s pre-move
golden, `memory::tests::golden_block`, which pins this exact fixture as part
of the assembled volatile block. So it is what the engine wrote before the
move. The test fails on `main` by construction — `stella_learn` is not a
crate there — and it cannot pass unless parsing, scoring, the section-budget
cut and the render all crossed byte-identically. `golden_block` itself is
unchanged and still runs, so the same bytes are pinned from both sides.

`crates/stella-learn/src/skills/tests.rs::skill_parsing_tolerates_every_invocation_directive_key`
is new, and replaces the four lines of
`skill_invocation`'s `directives_round_trip_and_tolerate_unknown_keys` that
called `skills::skill_from_file`: the two parsers now sit in different
crates, so each side pins its own half.

## Exemplar

`stella-records` (#5117 / #5829) is the recipe this follows step for step,
down to the README's Boundary section and the crate-table row.
`stella-diff` and `stella-embed` are the workspace's older pattern for the
same shape.

## Tests deleted

None. Moved, and named here because the guard reads the tree rather than the
file:

- `parses_description_guard_and_body`,
  `no_fence_means_whole_trimmed_text_is_the_body`, `strips_a_leading_bom`,
  `strips_matching_quotes_from_values`,
  `ignores_comment_and_blank_frontmatter_lines`,
  `flattens_block_sequences_onto_their_key`,
  `dash_lines_without_a_pending_list_key_stay_ignored` —
  `crates/stella-core/src/rules/tests.rs` →
  `crates/stella-protocol/src/frontmatter.rs`, unchanged.
- `a_skill_candidate_costs_exactly_its_rendered_block`,
  `a_skill_cut_by_top_k_is_named_in_the_ledger`,
  `a_skill_omitted_by_the_section_budget_is_named_without_changing_the_bytes`
  — `crates/stella-core/src/steering/tests.rs` →
  `crates/stella-cli/src/memory/steering.rs`, import paths and doc comments
  only.

Every other moved test travels inside a file `git` reports as a rename.

## Baselines touched

- `scripts/core-reachability-baseline.txt` — `redact` retired; the entry is
  dead now the module is gone.
- `scripts/prose-baseline.txt`, `scripts/prose-grade-baseline.txt` — entries
  carried across the renames by `--update`; no count and no grade rose.
  Entries for files this branch does not touch are left at their `main`
  values rather than globally retightened, so this PR does not become the
  carrier of a shared-cell break for everyone else.
- `scripts/prose-density-baseline.txt` — `stella-core` 28.44 → **30.15**,
  `stella-protocol` 27.21 → 25.91, `stella-learn` added at 21.64.

  **Read the `stella-core` number rather than skim it: it is a loosening.**
  It is arithmetic over headers nobody edited — the files that left were
  shorter-headed than the crate's mean, so removing them raises it — and
  #5829's `--update` re-basing is what computes it. It is not permission to
  write more prose there.

## Remaining work

- **#5830 advances but does not close.** Three of its four bullets are gone:
  `rules`, `glob` and `redact` no longer come out of `stella-core`. The
  fourth stands — `crates/stella-records/src/adapt.rs` still reads
  `stella_core::steering::{SteeringCandidate, DroppedCandidate,
  SteeringSource}`, which is the only reason
  `crates/stella-records/Cargo.toml` still names `stella-core`. The issue's
  own text calls that "the smaller half", and this PR shows where it goes:
  the record adapters follow the skill adapters into
  `stella-cli::memory::steering`, beside the frame adapter already there.
  Left open rather than closed, with its remaining scope stated in a comment
  on it.
- **#5900 (filed)** — `stella-protocol` now holds three pure functions and
  the rule admitting them lives only in that crate's README. It should be an
  ADR, and `AGENTS.md`'s crate-table row for that crate should stop saying
  "types only" without qualification. Not folded in here: writing an ADR
  needs a number, a manifest entry and an index row, and this diff is
  already wide.

Closes #5118
Refs #5113
Refs #5830


macanderson:
Pushed `7fc4752` to fix the `E0433`s in `crates/stella-cli/src/skill_run_cmd.rs`: the import at the top was repointed to `stella_core::skill_invocation` but its three call sites still said `invoke::`. Renaming the call sites clears the two errors in the binary, the third in the test target, and the `unused import: self` warning that `-D warnings` would have failed on separately.

**Correcting my own commit message, since I can't amend this branch.** It says `macos-check.yml` was the only job that compiled this. That is wrong: `cargo check on the declared MSRV` (ubuntu, in `ci.yml`) failed on `275f0f24` with the identical errors — [job 101261233516](https://github.com/macanderson/stella/actions/runs/33949411790/job/101261233516). This was a plain workspace build break that any compile job catches, not a macOS-only escape. The `cpal` reasoning in that commit body applies to the macOS *dependency table*, not to this failure.

Verification before pushing: `cargo check -p stella-cli --all-targets` exit 0, `cargo fmt --check -p stella-cli` clean (rustfmt reflows the `starts_with` call at the longer name, which is in the diff). Grepped the tree for other stale `stella_core::{skills,rules,comparison,mining,redact,self_tuning,glob}` and bare `invoke::` references — none, so this was the only site the move missed.

Every check on the new head is queued behind a runner backlog as of writing; I have not yet seen CI confirm it.

---

macanderson:
**This PR and #5840 collide in one import hunk, and both obvious resolutions are wrong.** Flagging it on both so whichever lands second resolves it correctly rather than reddening `main` — the tree is red right now from exactly this shape of collision (#5901).

Each PR is correct against `main` on its own; they conflict only with each other, in `crates/stella-records/src/records/validate.rs`. #5840 adds glob-matching to record validation and imports `stella_core::glob::match_glob`; this PR moves that function out of `stella-core`:

```
<<<<<<< HEAD                                    (this PR)
use stella_learn::redact::redact_secrets;
=======                                         (#5840)
use stella_core::glob::match_glob;
use stella_core::redact::redact_secrets;
>>>>>>>
```

**Why both naive resolutions fail.** This PR sends the two things that file needs in different directions — `redact` to `stella-learn`, but `match_glob` to **`stella-protocol`** (`crates/stella-protocol/src/glob.rs`, beside `hash::record_hash`). So "take both sides" leaves `stella_core::glob` and `stella_core::redact`, neither of which survives this PR; "take HEAD" drops `match_glob` and silently deletes the validation #5840 adds.

**The resolution that works:**

```rust
use stella_learn::redact::redact_secrets;
use stella_protocol::glob::match_glob;
```

No manifest change needed — `stella-records` already depends on both crates in the composed tree.

**Verified, not assumed.** Merged `origin/main` + this PR + #5840 in a scratch worktree, applied the resolution above, and compiled the crate:

```
$ cargo check -p stella-records --all-targets
    Checking stella-protocol v0.9.327
    Checking stella-core v0.9.327
    Checking stella-learn v0.9.327
    Checking stella-records v0.9.327
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 16.78s
```

Exit code 0.

I also swept the other 26 open PRs for added references to the modules this PR removes (`skills`, `rules`, `mining`, `redact`, `self_tuning`, `comparison`, `glob`, `steering::adapt`); #5840 is the only one. Nothing to change here — this is a note for the merge, not a review finding.

---

macanderson:
Status on head `91e4534`, and two checks that are red for reasons outside this PR.

**Both defects are fixed and CI now confirms it.** `the macOS-only dependency table still compiles` passes (07:00→07:01) after failing twice for two unrelated causes — the stale `invoke::` call sites, then a `Cargo.lock` version-drift. `cargo check on the declared MSRV`, the Windows platform split, the container smoke, `typecheck + build` and every guard are green. `deck slides fit the canvas` also passes, which settles the browser `TimeoutError` seen on an earlier head as a flake rather than anything the one-word `<code>` citation edit caused.

**`main is not known-broken` — not this PR.** #5901 is open: `main` does not compile, because `crates/stella-core/tests/published_records_do_not_collide.rs` still imports `stella_core::{ingest, records}` after #5829 moved that plane to `stella-records`. It fails on every open PR. The repair is #5906 (`unblocks-main`, +2/-2), whose own `cargo test` passed at 07:11 — so it is proven, not just proposed. Nothing to do here; this clears when that lands. I deliberately did not write another repair: #5907 and #5908 were byte-identical duplicates already closed as such.

**`dod / dod` — stale, and I cannot re-run it.** The three DoD boxes on #5118 are now ticked, each with the command and output that establishes it ([evidence](https://github.com/macanderson/stella/issues/5118#issuecomment-5550248880)): the reachability baseline shrinks 13→12 with `redact` retired; `cargo test -p stella-learn` is `210 passed; 0 failed` including the witness test; and no moved module remains in `crates/stella-core/src/`, none is declared in its `lib.rs`, and `stella-core` takes no dependency back on `stella-learn`.

That run completed at 07:17:10 and I ticked the boxes at 07:19, so it read the pre-tick issue. The check reads issue state live (oxagen#2643), so the next run passes on its own — but `rerun-failed-jobs` returns 403 for me, and I am not pushing an empty commit or shuffling a label to force one. It will clear on the next push to this branch, which is in any case not the binding constraint while `main` is red.

Substituted one verification deliberately and say so rather than let a tick imply otherwise: item 3's DoD proposes a `cargo build -p stella-core --timings` file-count drop. I proved it structurally instead — a module absent from the crate and undeclared by its root cannot be compiled into it at all, whichever way a count reads. Ask and I will produce the `--timings` number itself.

---

Issue #4417, opened 2026-08-23
plugin Stop hooks declared in a manifest never reach the engine: PluginRoster::hook_routes has no production consumer


macanderson:
## Problem

A plugin manifest can declare hooks (`[hooks]`, routed by `PluginRoster::hook_routes` in `crates/stella-cli/src/plugin_cmd/roster.rs`), but `hook_routes` is consulted only by `plugin_cmd::list` and its tests. Nothing folds a plugin's Stop hook into the engine's `HookConfig`, so a declared hook is advertised by `stella plugin list` and never fires. Found while attempting #3487 (PR #4416): wiring a plugin's `max_holds` into `stop_hook_feedback` today would let a third party's manifest change the bound on the operator's own shell hooks, for hooks that cannot fire — so #3487 is blocked on this.

## Done means

- The session assembly (`crates/stella-cli/src/agent/engine.rs` or a sibling — `tuned_engine_config` is where other catalog-derived config lands) installs the routes from installed, trusted plugins into the engine's hook plane with the plugin's identity on each route, subject to the same consent/trust gate `plugin_cmd` applies at install (`project_code_execution_trusted`).
- The combination rule when several plugins declare Stop hooks / `max_holds` is written down (doc comment) — minimum of the declared bounds is the safe default.
- Witness: install a fixture plugin declaring a Stop hook, run a turn, assert the hook fires once (today it never does). Cite AGENTS.md invariant 10 — every emitted signal names its consumer — in the PR.

Related: #3487 (blocked on this), #3246, #3599, #4416. `command_deck.rs` is a god file.

Issue #2684, opened 2026-08-10
IMPROVEMENT: expand the user hook surface — Stop/PreCompact events, input-rewriting and permission-deciding PreToolUse via the extension bus


macanderson:
## The gap

Stella's settings-declared shell hooks cover exactly three events: SessionStart, PreToolUse, PostToolUse (`crates/stella-core/src/hooks.rs:46-50`). PreToolUse can only block (non-zero exit → error); PostToolUse is pure observation. The rich in-process extension bus (~110 event names, Modify decisions, policy chains) exists but is unavailable to a user with a settings file — extensions require compiled code.

Missing hook points that operators actually use: turn end (Stop) for "don't stop yet" gating and end-of-turn automation, pre-compaction (to inject summarization instructions or veto), user-prompt-submit (to inject context or validate), subagent lifecycle, and tool-failure-specific handling.

## What the comparator does

Zappy code exposes ~25 hook events to settings/config, and its PreToolUse hooks are decision-makers, not just gates: a hook can **rewrite the tool input**, return a permission decision (`allow`/`deny`/`ask` — with explicit precedence: a hook allow can never override a deny rule), or **inject additional context** into the transcript. Stop hooks can block turn completion with a reason that feeds the next iteration (with a documented guard against the compact→error→stop-hook→retry death spiral). Hooks run in parallel with per-hook timeouts and slow-hook surfacing.

## Proposed fix

Incremental, riding existing machinery:
1. Bridge settings-declared shell hooks onto the extension bus's blocking allowlist: the `Allow | Modify | Deny | RequireApproval` decision enum already exists (`crates/stella-tools/src/registry.rs:805-1105`) — let a shell hook's stdout JSON carry a Modify payload, giving input-rewrite for free.
2. Add `Stop` (before `Completed` in `dispatch_completion`) and `PreCompact` (before `summarize_overflow_span`) events. Stop-hook block messages become tail user-role messages under a recognizable prefix, with a per-turn fire-count latch.
3. UserPromptSubmit and SubagentStart/Stop as follow-ups (subagent brackets already exist as bus events, `subagent.rs:496` area).

## Tool-first alignment (governance half #2716)

Three design constraints so this lands as part of one authorization seam instead of beside it:

[...]

PR #5903, opened 2026-09-05
feat(stella-core): add UserPromptSubmit and SubagentStart/Stop hook events

### LINKED ISSUE #2836: Follow-up hook events: UserPromptSubmit and SubagentStart/SubagentStop
Problem: #2684 shipped Stop and PreCompact on the settings-declared shell-hook surface and explicitly deferred UserPromptSubmit and SubagentStart/Stop.

Pattern to follow (all in the #2684 change): add variants to HookEvent and keys to Hooks in `crates/stella-core/src/hooks.rs` (serde-renamed PascalCase, matcher-ignored via `HookEvent::tool_scoped`), payload constructors on HookPayload, decision folding via `hooks/decision.rs::run_decision_hooks` (decide per event which decisions apply and the failure posture — permission-shaped gates fail closed per OXA-2056, turn-boundary gates fail open; see the documented asymmetry in `crates/stella-core/src/driver/user_hooks.rs` module docs).

Firing seams: UserPromptSubmit belongs host-side in the CLI where the user's prompt is accepted, BEFORE the turn is built (the same host-obligation shape as SessionStart, see the hooks.lifecycle row in `crates/stella-parity/src/lib.rs`) — a deny there should reject the prompt with the reason, a modify could rewrite it; SubagentStart/Stop belong in `crates/stella-core/src/subagent.rs` around the child Engine build/settlement (note the child inherits hooks + hook_approvals already).

Each new event needs: serde round-trip tests, matcher-ignored test, a fail-on-base witness, `website/content/docs/agent-tools/hooks.mdx` cards/sections, and README/parity prose updates. Done when both events fire with witnesses proving presence, and the docs page lists them.

Refs #2684, #2787.

## Definition of done

- [x] `HookEvent` gains `UserPromptSubmit`, `SubagentStart`, `SubagentStop` (stella-protocol), each with a serde round-trip test and a matcher-ignored test
- [x] `Hooks` in `crates/stella-core/src/hooks.rs` gains the matching settings keys, plumbed through settings merge and completeness checks
- [x] `HookPayload` gains constructors for all three events
- [x] `UserPromptSubmit` fires host-side in the CLI, before any turn is built; a `deny` rejects the prompt with the reason, a `modify` can rewrite it, and a broken hook fails closed
- [x] `SubagentStart`/`SubagentStop` fire around a child turn in `crates/stella-core/src/subagent.rs`, observe-only (never veto or rewrite)
- [x] A fail-on-base witness test proves each event actually fires
- [x] `website/content/docs/agent-tools/hooks.mdx` documents all three events (cards and sections)
- [x] README / stella-parity prose updated to name the new events

### Verification for the ticks above

Checked against #5903 at head `4c203d822`, item by item, with the tests **run** rather than inferred from the diff:

```
$ cargo test -p stella-protocol --lib -- round_trips_byte_for_byte only_the_tool_call_events_are_tool_scoped every_variant_is_listed
test hook::tests::round_trips_byte_for_byte ... ok
test hook::tests::only_the_tool_call_events_are_tool_scoped ... ok
test hook::tests::every_variant_is_listed ... ok

$ cargo test -p stella-core --lib -- the_2836_events
test hooks::tests::the_2836_events_ignore_the_matcher_field ... ok
test hooks::tests::the_2836_events_deserialize_and_build_payloads ... ok
test result: ok. 2 passed; 0 failed

$ cargo test -p stella-core --lib -- subagent_start_and_stop_hooks_fire_around_a_child_turn
test subagent::tests::seams::subagent_start_and_stop_hooks_fire_around_a_child_turn ... ok

$ cargo test -p stella-cli --bins -- hook_wiring
test agent::tests::hook_wiring::a_deny_decision_rejects_the_prompt_with_the_hooks_reason ... ok
test agent::tests::hook_wiring::a_modify_decision_rewrites_the_prompt ... ok
test agent::tests::hook_wiring::a_broken_hook_rejects_the_prompt_rather_than_waving_it_through ... ok
test agent::tests::hook_wiring::an_allowing_hook_leaves_the_prompt_untouched ... ok
test agent::tests::hook_wiring::no_hooks_configured_leaves_the_submitted_prompt_untouched ... ok
test result: ok. 7 passed; 0 failed
```

All exit 0. Item-by-item, where each one is satisfied:

1. Three variants in `HookEvent::ALL` (now 27); round-trip and tool-scoped tests both iterate `ALL`, and `every_variant_is_listed` is what forces a new variant into that array — so the three are covered by construction, not by a test that happens to name them. Matcher-ignored is additionally pinned by `the_2836_events_ignore_the_matcher_field`.
2. `settings/merge.rs:59-61` joins all three keys; `settings/completeness.rs:209-211,253-263` checks them.
3. `hooks/payload.rs:181,189,197` — `user_prompt_submit`, `subagent_start`, `subagent_stop`.
4. Fired at `agent.rs:474` and `agent.rs:576` via `agent/engine.rs:377 user_prompt_submit_hook`, before any turn exists. Deny / modify / fail-closed each have their own passing test above.
5. `subagent.rs:666` and `subagent.rs:733`, documented observe-only at `subagent.rs:749`.
6. `subagent_start_and_stop_hooks_fire_around_a_child_turn` is labelled in-source as the witness for the pair; the five `hook_wiring` tests are the witnesses for `UserPromptSubmit`.
7. `hooks.mdx` — 15 references across cards and sections.
8. `stella-core/README.md:289-291` and `stella-parity/src/lib.rs:460,469`.

Two follow-ups the PR filed rather than folded in, both outside this issue's DoD: #5904 (two turn-launching surfaces deliberately not wired, pending a decision about what counts as "the user's prompt" in a delegated context) and #5905 (`require_approval` needs a real interactive route instead of refusing). Neither blocks these boxes.

macanderson:
## What

Adds the three settings-declared shell hook events #2684 deferred:
`UserPromptSubmit`, `SubagentStart`, `SubagentStop`.

## Why

#2684 shipped `Stop` and `PreCompact` and explicitly left these three for a
follow-up (#2836). They round out the settings hook surface to cover a
user's own prompt and a child (sub-agent) turn, matching Claude Code's
hook vocabulary.

## Mechanism

- `HookEvent` (stella-protocol) gains the three variants — all `in_turn`,
  none `tool_scoped` — with serde round-trip and matcher-ignored coverage.
- `Hooks`/`HookPayload` (stella-core) gain the matching settings keys and
  constructors, plumbed through `settings::merge`/`settings::completeness`
  (the exact #2684 shape that caught `Stop`/`PreCompact` being silently
  dropped by `concat_hooks`).
- **`UserPromptSubmit`** fires **host-side** in `stella-cli`
  (`agent::user_prompt_submit_hook`), before any turn is built — the same
  host obligation `SessionStart` already has (no `Engine` exists yet, so
  `driver::user_hooks` cannot own it). A `deny` rejects the prompt with the
  hook's reason; a `modify` can rewrite the prompt text; an evaluation
  failure (a hook that never ran, or exited non-zero) fails **closed**,
  through the same `resolve_precedence` ladder `PreToolUse` uses (OXA-2056)
  — this is a permission gate, not a turn-boundary gate. Wired into every
  place a fresh user prompt starts a turn: the one-shot door
  (`agent/goal.rs::run_raw_one_shot`), the plain interactive mode's chat and
  `/goal`, and the Command Deck.
- **`SubagentStart`/`SubagentStop`** fire around a child turn in
  `stella-core::subagent::run_sub_agent_with_sender`, **observe-only** —
  `PostToolUse`'s posture, not `PreToolUse`'s: the parent already decided to
  delegate. Not fired on the cancel-drop path (`CancelBracket::drop`), which
  is a `Drop` impl with no async context to spawn a hook's subprocess from
  — documented as a known, structural gap rather than silently missing.
- Both new host-fired events are also wired into the plugin roster's
  `EVENT_ORDER`/`slot_for` (`plugin_cmd/roster.rs`, `plugin_hooks.rs`), so
  an installed plugin can be routed at them through the same fold that
  already carries `Stop`/`PreCompact` — otherwise they would be declarable
  in a manifest (since `HookEvent::in_turn()` gates that) and silently
  unroutable, the exact defect #4417 fixed for `Stop`.
- `website/content/docs/agent-tools/hooks.mdx` documents all three events
  (cards + sections); `stella-parity`'s `hooks.lifecycle` row and
  `stella-core`'s README note the new host obligations.

## Witness

Fails on `main` (the event/function does not exist there), passes with
this change:

- `crates/stella-core/src/hooks.rs`: `the_2836_events_ignore_the_matcher_field`,
  `the_2836_events_deserialize_and_build_payloads`
- `crates/stella-cli/src/agent/tests/hook_wiring.rs`: five `UserPromptSubmit`
  tests (deny / modify / allow / broken-hook-fails-closed / no-hooks),
  driven through a real shell via `HostHookRunner`
- `crates/stella-core/src/subagent/tests/seams.rs`:
  `subagent_start_and_stop_hooks_fire_around_a_child_turn`
- `crates/stella-cli/src/plugin_hooks/tests.rs`:
  `a_plugin_may_be_routed_at_each_2836_event`
- `crates/stella-protocol/src/hook.rs`: round-trip/wire-string/in-turn-family
  tests extended to the three new variants

Exemplar followed: the #2684 PR's own shape for `Stop`/`PreCompact` (same
files, same settings-merge discipline, same decision-fold reuse).

Tests deleted: none.

## Scope note

`UserPromptSubmit` is wired into every CLI surface that starts a turn from a
user-typed string that this session touched directly. Two turn-launching
surfaces were **not** wired in this PR — `stella goal`'s own multi-round
command (`agent/goal.rs::run_goal_cmd`) and the deck's sub-session / fleet
worker dispatch paths — because they raise their own design questions about
which text counts as "the user's prompt" for a delegated or multi-round
context. Filed as #5904; `agent::engine::user_prompt_submit_hook`
is the function they should reuse.

Also filed as #5905: giving `UserPromptSubmit`'s `require_approval` decision
a real interactive `ApprovalRoute` instead of always refusing (today it
refuses the same way `PreToolUse` does with no route attached, naming the
gap).

Closes #2836

## Tests deleted

`the_in_turn_family_is_exactly_the_five_the_engine_dispatches` — **renamed, not
removed.** It is now
`the_in_turn_family_is_exactly_the_events_the_turn_lifecycle_dispatches`, in the
same file (`crates/stella-protocol/src/hook.rs`) and asserting the same thing:
that `HookEvent::ALL.filter(in_turn)` equals an exact expected list.

The rename is the point. The old name pinned the count at *five*, and this PR
makes it eight — a count in a name is a promise to update it, the same rule
AGENTS.md applies to counts in prose. The assertion itself is unchanged in
shape; only the expected list grew by the three new events.

`check-deleted-tests` keys on the bare function name, so a rename reports here
by design, and naming it in this description is the whole cost.

## Summary by Sourcery

Add prompt-submission and subagent lifecycle hooks across the protocol, core engine, CLI, plugins, and documentation.

New Features:
- Add settings-declared `UserPromptSubmit`, `SubagentStart`, and `SubagentStop` hook events with payloads, plugin routing, and lifecycle support.
- Allow `UserPromptSubmit` hooks to reject or modify prompts across supported CLI turn-starting surfaces.
- Run observe-only subagent lifecycle hooks around child turns with child identity and outcome metadata.

Bug Fixes:
- Ensure the new hook settings are preserved during configuration merging and completeness tracking instead of being silently dropped.

Enhancements:
- Extend hook protocol coverage, payload disclosure metadata, parity declarations, consent messaging, and lifecycle documentation for the new events.

Documentation:
- Document the three hook events, their payloads, decision behavior, and failure handling.

Tests:
- Add protocol, payload, settings, CLI wiring, plugin routing, subagent lifecycle, and disclosure coverage for the new hook events.

---


macanderson:
Pushed `416e7c0` to fix the two compile errors that took the Windows and macOS jobs red on `933006f`.

### What was wrong

`run_subagent_hooks` in `crates/stella-core/src/subagent.rs` read `handle.runner` and `handle.hooks` directly. Those fields are private to `HooksHandle`, and a private field is visible in the module that defines it *and its descendants* — so `driver.rs` sees them, `driver/user_hooks.rs` sees them as a child, and `subagent.rs` does not: `lib.rs` declares `pub mod driver` and `pub mod subagent` as siblings.

### The fix

`HooksHandle::parts()` already exists for precisely this caller — its doc comment in `driver/capabilities.rs` names `crate::subagent` — and `run_sub_agent_with_sender` further down this same file already uses it as `self.hooks.map(HooksHandle::parts)`. Reusing it beats widening the fields to `pub(crate)`: the fields stay private, and it adds no line to `driver.rs`, which is grandfathered in `scripts/file-size-baseline.txt` and closed to growth.

```rust
let (hooks, runner) = handle.parts();
let outcome = run_hooks(runner, Some(hooks), &payload).await;
```

### Evidence

`cargo check -p stella-core --lib`, exit code rather than a grep over colourised output:

| | exit | errors |
|---|---|---|
| before | **101** | `error[E0616]` on `runner`, `error[E0616]` on `hooks` |
| after | **0** | none |

Those are the two the Windows job reported (`could not compile stella-core (lib) due to 2 previous errors`). `cargo fmt -p stella-core --check` is clean.

### One thing worth flagging

**This was never platform-specific.** It is a plain library compile error that Linux would have caught too — `ci.yml` simply had not reached the compile step when Windows and macOS reported, because ~33 PRs are queued and those two jobs start sooner. Reading the red as "a Windows problem" would have sent someone into `#[cfg(windows)]` arms for a bug that is in ordinary cross-platform code. Worth noting since `windows-check.yml` exists exactly to surface genuine platform splits, and here it just got there first.

Not re-run locally: the workspace build and the test suite, which are CI's job per CLAUDE.md. The remaining checks on this PR are still queued behind `main` being red (#5901) — `main is not known-broken` will stay red on every PR until #5906 lands.

---

macanderson:
### Settling the Sourcery ❌ row

> **Fire UserPromptSubmit before a CLI user prompt becomes a turn** … *"explicitly omits `stella goal`'s multi-round `run_goal_cmd` and the Command Deck sub-session/fleet worker dispatch paths. Therefore UserPromptSubmit does not fire before every user-prompt turn-launching surface covered by the issue's definition of done."*

**Sourcery is factually right, and the row is already answered.** Verified both halves rather than taking either the bot or the PR description at its word.

The gap is real: `run_goal_cmd` is at `crates/stella-cli/src/agent/goal.rs:520`, and `git grep UserPromptSubmit` over `crates/stella-cli/` returns no firing site inside `goal.rs` — the hook fires from `agent.rs`, `agent/engine.rs` and `command_deck.rs` only.

It is also deliberate and tracked, which is the "answer it" path CLAUDE.md allows:

| | |
|---|---|
| **#5904** | *Wire UserPromptSubmit into stella goal, subsession, and fleet dispatch* — open, `triage` |
| **#5905** | *Give UserPromptSubmit's `require_approval` a real interactive approval route* — open, `triage` |

Both open, both labelled `triage` and nothing else, per SCR-005. The PR description names #5904 and points the future implementer at `agent::engine::user_prompt_submit_hook` as the function to reuse — a handoff a fresh agent can execute, which is what SCR-004 asks for.

The reason given is a design question, not an oversight: which text counts as "the user's prompt" when the context is delegated or multi-round. Deferring that is the right call — guessing it and wiring the hook to the wrong string would be harder to undo than leaving the surface unwired.

**One thing for whoever merges:** #2836's definition of done asks for the hook before *every* user-prompt turn-launching surface, and after this PR two remain. So #2836 should **stay open** with #5904 as its remainder, or #5904 should be made a sub-issue of it — closing #2836 on this merge would lose the tracking that the deferral depends on. Worth a look because the PR body carries no `Closes #2836`, so the automatic path will not close it, but `dod / dod` reads linked issues live and a reviewer might close it by hand.

Not re-litigating the row beyond this: the deferral is explained, tracked, and handed off, which is what the rule asks for.

---

macanderson:
Second fix pushed: `5d8e22c`. **Windows is now green**; this was the remaining compile failure, on macOS.

### The error

```
error[E0004]: non-exhaustive patterns: `HookEvent::UserPromptSubmit`,
  `HookEvent::SubagentStart` and `HookEvent::SubagentStop` not covered
  --> crates/stella-cli/src/self_driving_cmd/hooks/tests.rs:318:11
error: could not compile `stella-cli` (bin "stella" test) due to 1 previous error
```

`payload_for` matches `HookEvent` totally on purpose — its last arm panics for events the self-driving loop never dispatches, and the comment there says why: *"a total match is what keeps a new event from silently landing in a default arm that reports nothing."* That arm is doing exactly the job it was written for; this PR added three variants without extending it.

### The fix

The three are in-turn events, so they join the arm the other in-turn events already share, ordered as `stella-protocol`'s `HookEvent` declares them (`…PreCompact | UserPromptSubmit | SubagentStart | SubagentStop`).

I also rewrote the arm's comment from **"the in-turn five"** to "the in-turn events". That count was about to become eight, and AGENTS.md is explicit that a count in prose is a promise to update it — pointing at the list is the prescribed form.

### Why only macOS caught it

Not a flake, and not a platform split. The arm is inside a `#[cfg(test)]` module, so it compiles only under `--all-targets`:

| job | command | reaches this file |
|---|---|---|
| `windows-check.yml` | `cargo clippy -p stella-tools -p stella-runtime` | no — wrong crates, and no `--all-targets` |
| `macos-check.yml` | `cargo check -p stella-cli --all-targets --locked` | **yes** |
| `ci.yml` | full workspace | yes, but had not reached the compile step |

AGENTS.md documents the Windows job's omission of `--all-targets` deliberately — several `#[cfg(test)]` bodies import `std::os::unix::fs::PermissionsExt` unconditionally and would fail the job on a fixture rather than on the platform split (#3497). So this is the coverage working as designed, not a gap: `macos-check` reached a target `windows-check` cannot.

Both failures on this PR were ordinary cross-platform Rust that Linux would also have rejected. The platform jobs simply got there first while `ci.yml` sat behind the queue.

### Evidence

`cargo check -p stella-cli --all-targets` — the macOS job's own command — exits **0** here; it fails with the E0004 above without this commit. `cargo fmt -p stella-cli --check` clean.

`dod / dod` is also red on this head. That is the SCR-003 gate reading #2836's checklist live, and it relates to the deferral I described in my earlier comment — #5904 carries the two unwired surfaces. Worth a look from someone who can decide whether #2836's boxes should be ticked with #5904 as the remainder, since that is a judgement rather than a code change.

---

macanderson:
Both remaining failures addressed. Pushed `6679b83`, and edited this PR's description.

### 1. `gate steps no other workflow runs` — the density ratchet

```
  crates/stella-core: 28.44 allowed, 28.55 mean header lines
```

**Not this PR's doing, and not mine.** Measured across the four trees:

| tree | files | header lines | mean |
|---|---|---|---|
| `4e33b4f` (before #5906) | 129 | 3654 | **28.33** |
| `main` after #5906 | 128 | 3640 | **28.44** — the ceiling exactly |
| this branch at `4c203d8` | 128 | 3655 | **28.55** |

#5906 — the red-`main` repair — moved `published_records_do_not_collide.rs` out of `stella-core`. Its header is 14 lines, half the crate mean, and removing a *below-average* header **raises** the mean. That took the crate from 28.33 to exactly 28.44 and spent every bit of headroom, correctly and necessarily. This branch then added 15 header lines documenting the three new events, and landed 0.11 over.

`main` passes because it sits *at* the ceiling rather than over it, so nothing reported the cost until this PR paid it. Filed as **#5914** — the density metric does not model a file move, the same blind spot #5910 describes in `check-rebase-replay`, and both trace to the record-plane extraction.

**Fixed by cutting a header, which is the remedy the guard names.** `subagent.rs` was the crate's longest at 125 lines — 28 more than the next and 4.4× the crate mean — and most of the excess was history that `doc:prose-guidelines` puts in the PR or the tracker. Removed 20 lines, keeping every operative fact:

- the telemetry anecdote (ninety rows, a session id, a five-way fan-out) goes; **why** the four events carry `sub_agent_id` stays;
- `goal.rs::assess`'s hand-rolled verifier engine goes; that this fork is built through `Engine::assemble`, so a later seam cannot reach a child as an unexamined `None`, stays;
- `not parallelism` was a foil against a claim nobody made, and *"the cheapest growth is the growth that never happens"* restated the sentence above it.

Your 15 lines of new-event documentation are untouched apart from one paragraph rewrapped four lines instead of five. **The baseline is not edited** — the script forbids it and so does CLAUDE.md.

Now: 3640 lines, mean 28.44, at the ceiling. `check-prose` exits 0 on all three sub-checks, `check-line-citations` OK, `cargo fmt -p stella-core --check` clean, `cargo check -p stella-core --lib` exits 0.

### 2. `check-deleted-tests`

```
These tests exist in HEAD^1 but not in the merged tree, and nothing
in the PR description or the branch's commit messages names them:
  the_in_turn_family_is_exactly_the_five_the_engine_dispatches
```

A **rename**, not a loss — verified rather than assumed. `crates/stella-protocol/src/hook.rs` carries `the_in_turn_family_is_exactly_the_events_the_turn_lifecycle_dispatches`, same file, same assertion shape. Renaming was right: the old name pinned the count at *five*, which this PR makes eight.

The guard keys on the bare function name, so a rename reports by design, and its stated cost is one sentence in the description. Added a **Tests deleted** section saying exactly that. The guard reads the description live through the API, so it clears on this run.

### Still open, and not a code fix

`dod / dod` reads #2836's checklist, whose two unwired surfaces (`run_goal_cmd`, deck sub-session/fleet dispatch) are deliberately deferred to #5904 — see my earlier comment. Whether #2836's boxes may be ticked with #5904 as the remainder is a judgement for a human, not something to push.

---

PR #5921, opened 2026-09-05
fix(stella-tui): keep a panel's combining marks, and record what a cell counts

### LINKED ISSUE #5185: A panel frame's cell count is chars, not display width — the host's clip is the only thing making that safe, and nothing asserts it

macanderson:
## What and why

`PanelText::cells` counts `char`s, `PanelFrame::fits` is built on that count,
and the type's docs argued the difference was safe "because the host clips
every blit at the leased rectangle". When #5185 was filed no such host
existed. It does now: #5223 shipped `crates/stella-tui/src/plugin_panel.rs`,
whose `write_run` advances by display width and refuses a glyph whose columns
would cross the lease's edge, with
`a_wide_glyph_cannot_reach_past_the_lease_into_the_border` drawing the chrome
first so the assertion is against the host's real border cell. That is boxes 1
and 2 of the issue's Definition of done, already landed.

This PR settles the rest, and fixes the other half of the defect the issue's
Problem section names.

**The combining mark (host, a real fix).** `write_run` dropped every
zero-width `char`, so a plugin sending decomposed text lost its marks: `e` +
U+0301 painted as `e`, with nothing to say the accent had gone. A mark now
rides the cell the glyph before it took. That is how ratatui writes one itself
— `Buffer::set_stringn` splits into grapheme clusters and puts a whole cluster
in one cell (`ratatui-core-0.1.2/src/buffer/buffer.rs`), which is the exemplar
followed here. A mark with no glyph to ride is still dropped, because its own
cell would attach it to whatever the host drew there. One cell keeps at most
`MARKS_PER_CELL` marks: a mark costs no column, so the lease cannot bound how
many a frame hangs on one glyph, and every one is bytes the terminal redraws.
30 is the cap Unicode's stream-safe format uses, so conforming text loses
nothing. No new dependency.

**The decision (contract).** `cells()` keeps counting `char`s and the host
keeps the column count, recorded as
`docs/adr/0028-panel-cells-are-glyphs-in-the-contract.md` per SCR-002. The
reasons are the issue's own — `stella-plugin` is a near-leaf that argues each
workspace edge, and changing what a cell means changes which frames are legal,
which is a wire-contract change — plus one the issue does not make: two exact
counts would still disagree, because a width table has a version and the
terminal has its own, so only the clip has to be right.

**The prose.** `PanelText`'s "what a cell is here" paragraph now names the two
tests that hold the property and cites the ADR, instead of asserting the clip.
Nothing in the wire contract changed, so no `make wire-schema-update`.

## Witness mechanism

`a_combining_mark_rides_the_glyph_it_modifies` fails on the old code by
construction: the old `write_run` hit `if width == 0 { continue; }` for U+0301,
so cell (0,0) held `e` and the assertion asks for `e\u{301}`. It passes on the
new code because the mark is appended to that cell's symbol.

`one_cell_keeps_no_more_marks_than_the_stream_safe_limit` fails without the
cap (the symbol would carry all 35 marks sent).

`a_frame_the_contract_admits_by_glyph_count_is_cut_at_the_lease` passes before
and after, and that is its job: it asserts `fits` admits four ideographs
against a four-column lease *and* that the host cuts them at two with the
border untouched, so changing either half alone breaks it. It is the seam ADR
0028 records, pinned in the one crate that can see both sides.

## Tests deleted

None. One rename: `a_zero_width_glyph_is_never_placed` →
`a_combining_mark_with_no_glyph_to_ride_is_dropped`. Same body, same
assertion; the old name claimed more than the code now does, since a mark with
a glyph to ride *is* placed.

## Definition of done

- [x] The host clips every panel blit at the leased rectangle — `write_run`,
      landed #5223, unchanged here.
- [x] A test proves a frame of wide glyphs measuring within its lease still
      paints nothing outside it —
      `a_wide_glyph_cannot_reach_past_the_lease_into_the_border` (#5223), and
      `a_frame_the_contract_admits_by_glyph_count_is_cut_at_the_lease` (this
      PR) starts from a frame `fits` admits.
- [x] `panel.rs`'s "what a cell is here" paragraph names those tests rather
      than asserting the clip.
- [x] A decision is recorded on whether `cells()` should measure display
      width — ADR 0028 says no, with the costs and what would reopen it.

## One correction to the issue

Its Verify section offers `"日本語テキスト"` against `PanelRect::new(4, 1)` as a
frame `fits` accepts today. It does not: that is seven `char`s against four
columns, so `fits` refuses it on the glyph count. The seam needs a string whose
glyph count fits and whose column count does not, so the test uses `"日本語だ"`
— four glyphs, eight columns.

## Remaining work filed

- #5922 — `crates/stella-plugin/src/panel.rs` is 11 lines under the 1500-line
  ceiling and takes no baseline entry, so the next change there has no room to
  design a seam.

## Note on the branch's history

#5823 landed ADR 0027 while this was open, so this record is **0028**: the file,
its id, its heading, the README row and the three `doc:adr` citations all moved
together. The branch was also flattened onto a fresh base after main's
`Cargo.lock` repair (#5940) made this branch's copy of that line an
edit-and-undo pair, which `no edit is undone inside the branch` correctly
caught. Both are settled on the current head; #5954 files the self-test case
that was missing for the rename half.

Closes #5185

---

## DoD re-check, and an ADR number claimed twice (appended 2026-09-05 by the PR sweep)

`dod / dod` failed at 09:49:52 against four unticked boxes on #5185. The boxes
were ticked at 09:52:49 — three minutes after that read — so the check is
stale rather than wrong. This edit is the `pull_request` event `dod-check.yml`
needs to read them again; the edit is the trigger, never the fix. It changes
no code.

The three items credited to this PR were checked against the diff at
`4d5c8ba` rather than against the description above:

- `a_frame_the_contract_admits_by_glyph_count_is_cut_at_the_lease` is in
  `crates/stella-tui/src/plugin_panel.rs` and asserts `frame.fits(...)` is
  `Ok(())` before it blits, so it does start from a frame the contract admits,
  and then asserts the border cell is byte-identical afterwards.
- `PanelText`'s "what a cell is here" paragraph names that test and
  `a_wide_glyph_cannot_reach_past_the_lease_into_the_border`, and cites the
  ADR instead of asserting the clip.
- `docs/adr/0028-panel-cells-are-glyphs-in-the-contract.md` is added and
  registered in both `docs/adr/README.md` and `docs/manifest.json`.

The first two boxes are #5223's and already landed.

### ADR 0027 is claimed by two open PRs

#5823 adds `docs/adr/0027-a-fleet-worker-gets-its-own-worktree.md` and
registers 0027 in the same two indexes this PR does. Both are open against
`main`, so `docs/adr/README.md` and `docs/manifest.json` are one shared cell
two branches write — the composition hazard AGENTS.md describes under the
file-size baseline: each branch is correct on its own, and the second to land
either collides in those indexes or ships a duplicate ADR number.

**Settled:** #5823 landed first and keeps 0027. This record is 0028, and the
file, its id, its heading, the README row and the three `doc:adr` citations
moved together.

## Summary by Sourcery

Preserve combining marks in panel output and document the glyph-count contract while relying on host-side column clipping.

New Features:
- Preserve combining marks by attaching them to the preceding glyph during panel rendering, with a stream-safe limit per cell.

Bug Fixes:
- Prevent combining marks from being silently lost while maintaining safe clipping of wide glyphs at panel lease boundaries.

Enhancements:
- Clarify that panel contract cells count glyphs while the host enforces terminal-column boundaries.
- Record the panel cell-count decision and its consequences in ADR 0028.

Documentation:
- Update panel and host documentation to describe glyph-versus-column semantics and link the boundary tests.
- Register ADR 0028 in the documentation indexes.

Tests:
- Add coverage for combining-mark preservation, marks without a carrier, the per-cell mark limit, and frames admitted by glyph count being clipped at the lease edge.

### DIFF
--- a/crates/stella-plugin/src/panel.rs
+++ b/crates/stella-plugin/src/panel.rs
@@ -879,14 +879,21 @@ impl PanelPatch {
 ///
 /// # What a cell is here, and what it is not
 ///
-/// [`PanelText::cells`] counts `char`s. A double-width glyph therefore draws
-/// wider than this contract counts, and the frame that carries it misaligns
-/// **its own interior** — never anything outside it, because the host clips
-/// every blit at the leased rectangle whatever a frame claims. Measuring
-/// display width would mean a Unicode width table, and this crate is a
-/// near-leaf that takes one workspace dependency on argument (see its
-/// `Cargo.toml`); the host already owns the clip that makes the difference
-/// safe.
+/// [`PanelText::cells`] counts `char`s, so a cell here is a glyph and not a
+/// terminal column. `あ` is one glyph and two columns; `e` followed by a
+/// combining acute is two glyphs and one column. A frame this contract admits
+/// can therefore need more columns than its lease has, or fewer.
+///
+/// The host measures the columns. `stella_tui::plugin_panel`'s `write_run`
+/// places a glyph only when every column it needs is inside the lease, so a
+/// row of wide glyphs is cut at that edge rather than drawn over the border.
+/// Two tests in that module hold it: `a_wide_glyph_cannot_reach_past_the_lease_into_the_border`
+/// draws the chrome first and asserts the border cell survives, and
+/// `a_frame_the_contract_admits_by_glyph_count_is_cut_at_the_lease` pins the
+/// seam between the two counts.
+///
+/// Keeping the `char` count is a recorded decision, with its costs and what
+/// would reopen it: `doc:adr/0028-panel-cells-are-glyphs-in-the-contract`.
 #[derive(Debug, Clone, Default, PartialEq, Eq, PartialOrd, Ord, Hash)]
 pub struct PanelText(String);
 
@@ -921,8 +928,8 @@ impl PanelText {
         &self.0
     }
 
-    /// How many cells these glyphs occupy — see the type docs on what a cell
-    /// counts as here.
+    /// How many glyphs these are — the count this contract calls cells. See
+    /// the type docs for what it does and does not measure.
     #[must_use]
     pub fn cells(&self) -> usize {
         self.0.chars().count()
--- a/crates/stella-tui/src/plugin_panel.rs
+++ b/crates/stella-tui/src/plugin_panel.rs
@@ -30,8 +30,17 @@
 //! repainted, so the border stays overwritten. That is a plugin painting a
 //! cell it was not leased, through a path a `chars()`-counting bounds check
 //! calls in-bounds. So a glyph whose width would cross the lease's right edge
-//! is refused rather than truncated, and a zero-width glyph is refused at the
-//! left edge, where the terminal would attach it to the border instead.
+//! is refused rather than truncated, and a combining mark rides the cell the
+//! glyph before it took rather than claiming one of its own — a mark in its own
+//! cell would attach to whatever the host drew there.
+//!
+//! The two counts still differ, and the difference is decided rather than
+//! overlooked: `stella_plugin::PanelText::cells` counts `char`s, so a row of
+//! wide glyphs the wire contract admits is cut here at the lease's edge. See
+//! `doc:adr/0028-panel-cells-are-glyphs-in-the-contract`, and this module's
+//! `a_frame_the_contract_admits_by_glyph_count_is_cut_at_the_lease` for the
+//! seam itself. That test is not linked: it is a `#[cfg(test)]` item, and a
+//! rustdoc link to one is an unresolved link in the run that ships.
 //!
 //! **The palette.** [`PanelInk`] names a token, never an RGB triple, and
 //! [`ink`] resolves it against the shipped theme — so a panel degrades with
@@ -196,22 +205,41 @@ fn blit_patches(patches: &[PanelPatch], rect: Rect, buf: &mut Buffer) {
     }
 }
 
+/// The most combining marks one cell keeps.
+///
+/// A mark takes no column of its own, so nothing else here bounds how many of
+/// them a plugin can hang on one glyph, and every one of them is bytes the
+/// terminal redraws on each repaint. Unicode's stream-safe format caps a
+/// sequence at 30 non-starters, so text that follows it loses no mark to this
+/// limit.
+const MARKS_PER_CELL: usize = 30;
+
 /// Write `text` rightwards from `x`, in display cells, stopping at the lease's
 /// right edge. Returns the column after the last glyph written.
 ///
 /// A glyph is placed only when **all** of its columns are inside the lease, so
 /// the last leased column refuses a double-width glyph rather than letting the
-/// terminal paint its second half over the host's border. A zero-width glyph —
-/// a combining mark that arrived without the character it modifies — is
-/// dropped at the left edge, where the terminal would otherwise attach it to
-/// the border cell.
+/// terminal paint its second half over the host's border.
+///
+/// A zero-width glyph rides the cell the glyph before it took, which is how
+/// ratatui writes one itself: `Buffer::set_stringn` splits into grapheme
+/// clusters and puts a whole cluster in one cell. A mark with no glyph to ride
+/// — one opening a run, or arriving after the lease's edge cut the run short —
+/// is dropped, because its own cell would attach it to whatever the host drew
+/// there.
 fn write_run(buf: &mut Buffer, mut x: u16, y: u16, rect: Rect, text: &str, style: Style) -> u16 {
     let right = rect.x.saturating_add(rect.width);
+    // The cell this run last wrote a glyph into, and how many marks it carries.
+    let mut carrier: Option<(u16, u16)> = None;
+    let mut marks = 0usize;
     for ch in text.chars() {
         let width = ch.width().unwrap_or(0);
         if width == 0 {
-            // Nothing to advance past, and nowhere safe to put it: at the
-            // lease's left edge it would modify the host's border.
+            if let Some(at) = carrier
+                && marks < MARKS_PER_CELL
+            {
+                marks += usize::from(attach_mark(buf, at, ch));
+            }
             continue;
         }
         let Ok(width) = u16::try_from(width) else {
@@ -221,6 +249,8 @@ fn write_run(buf: &mut Buffer, mut x: u16, y: u16, rect: Rect, text: &str, style
             break;
         }
         write_cell(buf, x, y, ch, style);
+        carrier = Some((x, y));
+        marks = 0;
         // The columns a wide glyph occupies beyond its first are the
         // terminal's to fill; the buffer must not leave a stale symbol in
         // them, or the diff emits a cell that overlaps the glyph.
@@ -233,6 +263,21 @@ fn write_run(buf: &mut Buffer, mut x: u16, y: u16, rect: Rect, text: &str, style
         x += width;
     }
     x
+}
+
+/// Hang one combining mark on the glyph already in `(x, y)`.
+///
+/// Returns whether it landed: a cell the buffer does not hold takes nothing,
+/// for the reason [`write_cell`] gives.
+fn attach_mark(buf: &mut Buffer, (x, y): (u16, u16), mark: char) -> bool {
+    let Some(cell) = buf.cell_mut((x, y)) else {
+        return false;
+    };
+    let mut symbol = String::with_capacity(cell.symbol().len() + mark.len_utf8());
+    symbol.push_str(cell.symbol());
+    symbol.push(mark);
+    cell.set_symbol(&symbol);
+    true
 }
 
 /// Write one glyph, if the buffer actually has that cell.
@@ -440,11 +485,11 @@ mod tests {
         assert_eq!(buf.cell((2, 0)).expect("cell").symbol(), "b");
     }
 
-    /// A lone combining mark is dropped rather than placed: it has no column of
-    /// its own, and at the lease's left edge the terminal would attach it to
-    /// the host's border glyph.
+    /// A combining mark that opens a run is dropped: it has no glyph to ride,
+    /// and a cell of its own at the lease's left edge would attach it to the
+    /// host's border glyph.
     #[test]
-    fn a_zero_width_glyph_is_never_placed() {
+    fn a_combining_mark_with_no_glyph_to_ride_is_dropped() {
         let area = Rect::new(0, 0, 4, 1);
         let mut buf = Buffer::empty(area);
         let frame = lines(vec![vec![span("\u{301}x", PanelStyle::plain())]]);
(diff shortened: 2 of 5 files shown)


macanderson:
## Two open PRs claim ADR 0027

Found by the open-PR sweep. This PR and #5823 each add a *different* ADR numbered 0027, and each registers it in the same two index files:

| PR | New file |
|---|---|
| **#5921** (this one) | `docs/adr/0027-panel-cells-are-glyphs-in-the-contract.md` |
| **#5823** | `docs/adr/0027-a-fleet-worker-gets-its-own-worktree.md` |

Both also insert a `0027` row into `docs/adr/README.md` immediately after the `0026` row, and both add an `adr/0027-…` key to `docs/manifest.json` immediately after `adr/0026-…`.

Verified against `origin/main` (`c498653`) rather than inferred from either description: `docs/adr/` there holds `0001`–`0026` and nothing else, so 0027 is genuinely free — and genuinely claimed twice.

### Why this is not caught pre-merge

Neither branch is wrong on its own, so both can go green: `adr-numbering` sees exactly one 0027 in each tree. `docs/adr/README.md` and `docs/manifest.json` are the shared cells AGENTS.md's `main-canary` section describes — the same shape as `Cargo.lock` and `scripts/file-size-baseline.txt`. Whichever of these two merges second lands a tree holding two ADRs numbered 0027, and `make adr-numbering` then fails on `main` for everyone.

### What is owed

**Whichever PR merges second renumbers to 0028**, in one edit:

1. `git mv` the ADR file to `0028-…`
2. update its frontmatter `id:` and its `# ADR 00NN:` heading
3. repoint the `docs/adr/README.md` table row and its prose paragraph
4. repoint the `docs/manifest.json` key and `path`
5. repoint every `doc:adr/0027-…` citation — this PR has two, in `crates/stella-plugin/src/panel.rs` (the `PanelText` doc comment) and `crates/stella-tui/src/plugin_panel.rs` (the module header and `a_frame_the_contract_admits_by_glyph_count_is_cut_at_the_lease`'s doc comment)

Neither PR is renumbered here, because which one takes 0028 is decided by merge order and neither has merged. The same note is on #5823.

### A second, smaller overlap on this branch

#5928 also rewrites the `PanelText` doc comment in `crates/stella-plugin/src/panel.rs` — the same block this PR replaces, to state the bidi-control refusal. That one git will report as a conflict rather than composing silently, so it needs no pre-emptive action; noting it so whichever lands second expects the resolution rather than being surprised by it. #5928 also splits `first_control_character` out into a new `drawable.rs` and leaves `panel.rs` at 1476 lines, so the two diffs move that file's size in opposite directions.

---

macanderson:
## Merge conflict resolved: the panel ADR is 0028 now

#5823 merged at 09:54:36, four minutes after I flagged the collision in the description above, so this stopped being a prediction — `mergeable_state` went to `dirty` and both `docs/adr/README.md` and `docs/manifest.json` conflicted. Resolved and pushed as `3e0ee12`.

**main's 0027 keeps the number**, having landed first. This branch's record moves to 0028, in every place that names it:

- `docs/adr/0027-panel-cells-are-glyphs-in-the-contract.md` → `0028-…` (`git mv`, so the prose guard carries the baseline entry across)
- its frontmatter `id`, its `title`, and its `#` heading
- `docs/adr/README.md` — both the table row and the trailing prose paragraph, with main's fleet paragraph kept intact beside it
- `docs/manifest.json` — both keys present, in sorted order
- the three `doc:adr/…` citations in `crates/stella-plugin/src/panel.rs` and `crates/stella-tui/src/plugin_panel.rs`

A merge commit, not a rebase — this is not my branch, and a merge keeps any checkout of it valid. No behaviour changed and no test was touched; the diff is the rename plus the two indexes plus three citation strings.

**Verified locally** (toolchain-free guards only, per CLAUDE.md's laptop-build ban — the compiled checks are CI's):

```
check-adr-numbering: OK -- 28 record(s), every number unique, every heading
                     and id in agreement, every record indexed.
check-doc-links:     OK -- 1153 citation(s) resolve across 105 document(s).
check-rebase-replay: OK — 33 path(s) touched, every one still in the branch's diff.
check-prose:         OK -- none added.
check-line-citations / check-file-size: OK
```

One thing worth knowing for the next person who resolves a conflict here: `check-prose` reported `docs/adr/README.md [issue-reference]: 11 allowed, 33 found` while the merge was still unresolved in the index. The file genuinely had 11. `git ls-files` lists an unmerged path three times (stages 1/2/3) and the guard counts it once per entry, so the count trebles until the path is `git add`ed. It is an artifact of the conflict, not a regression — worth not "fixing" the prose in response to it.

### On the stale `dod / dod`

The failure at 09:49:52 read four unticked boxes on #5185; they were ticked at 09:52:49, three minutes later. `rerun_failed_jobs` is `403 Resource not accessible by integration` for me, so this push is what fires the `pull_request` event that re-reads them — a new `dod / dod` started at 10:03:45. I checked the three items credited to this PR against the diff rather than the description before doing that, and they are all present; the detail is in the description above.


---

macanderson:
Two red checks on the merge ref, neither of them a defect in this diff.

**`cargo check on the declared MSRV` — `main` is red, not this branch.**
`origin/main` sets the workspace version to `0.9.330` and its `Cargo.lock`
still lists `stella-learn` at `0.9.329`, so every `--locked` build fails there
as well as here. `./scripts/main-red-claim.sh check` stood this session down:
another session holds the repair on #5939. This branch carries the identical
one-line edit so its own checks can run, and identical edits compose.

**`no edit is undone inside the branch` — a rename inside a merge commit.**
#5823 landed `docs/adr/0027-a-fleet-worker-gets-its-own-worktree.md` while this
branch was open, so both branches claimed ADR 0027. The merge that brought main
in resolved that by renaming this branch's record to
`docs/adr/0028-panel-cells-are-glyphs-in-the-contract.md`. `check-rebase-replay`
reads `git log --no-merges` for the paths a branch touched and compares them
with the branch's net diff; the rename happened *in* the merge commit, which
that log excludes, so the pre-rename path reads as an edit with no trace in the
diff.

The hazard the guard exists for needs a base that already carries the same
edit at the same path. Nothing upstream has ever carried
`0027-panel-cells-are-glyphs-in-the-contract.md`, so a rebase has nothing to
deduplicate and no revert to replay. Its remedy — flatten the pair — means
rewriting history a peer session has already pushed to, which is worse than the
false positive. The gap is filed as a guard issue rather than worked around
here.

macanderson:
The rebase-replay gap is filed as #5944 — a rename performed inside a merge commit reads as an undone edit, with the repro and a self-test case in the definition of done.

macanderson:
## `no edit is undone inside the branch` — this one is not the rename false positive

The recorded reasoning on this PR treats the failing `no edit is undone inside the branch` check as the merge-commit rename false positive tracked in #5944, with do-not-flatten advice attached. That argument does not cover what the guard is actually reporting on this head, so flagging it rather than leaving it to be read as already settled.

**#5931's fix is present on this branch and did not suppress the finding.** `git log --oneline pr5921 -1 -- scripts/check-rebase-replay.sh` resolves to `ff9bb9483` — the merge of #5931 ("stop check-rebase-replay reading an inherited rename as an undo"). The guard still fails with the fix in place.

**What it names is `Cargo.lock`, not a rename:**

```

  Cargo.lock
      afff26679  chore(cargo): resync Cargo.lock with the workspace version after merging main
```

Three facts checked against the branch at `4c1ad94`, base `ff9bb9483`:

- `git diff --name-only ff9bb9483...pr5921 | grep -c Cargo.lock` → `0`. The path is absent from the branch's diff, so it reads as untouched.
- `git log --oneline ff9bb9483..pr5921 --no-merges -- Cargo.lock` → exactly one commit, `afff26679`, and it is the branch's own — not an inherited base-branch removal, which is the case #5931 taught the guard to drop.
- `git diff --stat ff9bb9483 pr5921 -- Cargo.lock` → empty. Net effect zero.

That is the edit-and-undo shape #4979 describes, on the one path AGENTS.md singles out as a shared cell every version-bumping PR writes. The hazard is rebase-specific: rebasing onto a base that already carries the same lockfile edit drops the branch's copy (`skipped previously applied commit`) and keeps the undo, turning an inert pair into a live revert of whatever landed upstream in `Cargo.lock` — and it merges cleanly, with no conflict to report. Merging without rebasing is safe.

**Not fixing it here.** The remedy the guard prints is a history rewrite (`git rebase -i` / `git reset --soft`), which is not something to do on a branch I did not open, and the do-not-flatten advice on this PR is a deliberate call on record. Two options for whoever owns it: flatten the `Cargo.lock` pair so the branch's history stops mentioning the path, or merge this PR without rebasing and let the guard stand as knowingly-red.

Worth separating from #5944 either way — that issue is about an inherited rename, and this is a lockfile pair the same check caught for a different reason.

---

macanderson:
Correcting my own comment above, and confirming the state.

My "rename inside a merge commit" note was written against the **first**
failure of `no edit is undone inside the branch`, which named
`docs/adr/0027-panel-cells-are-glyphs-in-the-contract.md`. #5931 landed and
settled that one. The **second** failure named `Cargo.lock` and was a real
edit-and-undo pair of exactly #4979's shape: this branch resynced the lock
after merging main, main's own repair (#5940) then landed the identical line,
and the branch's net diff for that path went empty. The reading in the comment
above does not cover it, and the comment that says so is right.

#5944 is closed as already fixed by #5931, and the missing self-test case for a
rename the branch's own merge performs is filed separately as #5954.

The branch is now one flattened commit on a fresh base, so both pairs are gone
from its history and the guard passes. The only stale thing left is that
commit's message, which still says the record is ADR 0027; it is 0028, and the
squash body will carry the corrected number.

macanderson:
Force-pushed this branch to flatten it — `4c1ad94` → `f8286da`. Re-fetch before you build on it.

`no edit is undone inside the branch` was red. `check-rebase-replay` named `Cargo.lock`: `afff2667` ("resync Cargo.lock with the workspace version after merging main") undid an edit that the branch's three merges of `main` had brought in, so the path was touched in this branch's history and absent from its diff — the shape that becomes a live revert of whatever landed upstream once the branch is rebased, and merges cleanly (#4979).

Flattened, per the remedy the guard itself prints: the five commits (one change plus three merges of `main` plus the resync) are now one commit off current `main`, so the history no longer mentions `Cargo.lock` at all.

**The tree is unchanged, and that is checked rather than asserted.** `git merge-tree origin/main 4c1ad94` and the flattened commit both write tree `2ce4014a5cacc2266504ee5e4a312d75ff3c7b6e`. The original commit message is kept verbatim, `Closes #5185` included.

Verified on the flattened head:
- `./scripts/check-rebase-replay.sh` — `OK — 5 path(s) touched, every one of them still in the branch's diff.`
- `cargo test -p stella-plugin -p stella-tui --lib` — 169 + 1488 passed, 0 failed.

---

Issue #5923, opened 2026-09-05
A deck worker lane's Files tab is always empty


macanderson:
A `task_assign` or `req:<n>` worker lane runs a whole turn — it edits files,
creates them, deletes them — and its Files tab stays empty. The lane emits no
`AgentEvent::FileChange` at all, so `/files` and `/diff` for that lane read
"no files touched yet" however much the worker changed, and the per-call rows
in its transcript carry no diff and no `+N -M`.

## Where

`crates/stella-cli/src/subsession.rs`'s `run_worker` is the lane's whole
setup. It attaches its event stream by hand:

```rust
let events = lane_events::open_lane_stream(spec, &tx);
registry.attach_events(events.clone());
```

The lead turn does not. It calls `crate::turn_files::open_turn_streams` (or
`open_turn_streams_raw`), which takes on three debts together, on purpose,
because two of them are silent when forgotten:

- `registry.attach_events(...)` — registry-born events. `run_worker` has this.
- `registry.attach_call_measure(Arc::new(TurnCallMeasure::new(...)))` — this
  turn's per-call work-tree measurement (#4175). `run_worker` has not got this.
- `cfg.durability.refresh_worktree_sharers()` — who else is in the work tree
  (#4386), which those per-call measurements read. `run_worker` has not got
  this either.

The closing bookend is the same story: the lead turn ends with
`crate::turn_files::close_turn_boundary_raw`, which emits the turn's shared
tree changes before the run terminator. `run_worker` calls neither — it goes
straight from `close_turn_stream` to `closeout::close_worker_execution`.

`open_turn_streams`' own doc comment names this exact failure mode for the
Command Deck's lead turn, which had it until #4160: a missing measurement
renders as an honest-looking empty ledger.

## Verify it

`rg -n 'attach_call_measure|close_turn_boundary|refresh_worktree_sharers' crates/stella-cli/src/subsession.rs`
returns nothing today, and so does
`rg -n 'FileChange' crates/stella-cli/src/subsession.rs`.

## Constraints a fix has to respect

- `subsession.rs` is 1429 lines against the 1500 ceiling, and the god-file rule
  says new logic lands in a sibling submodule under
  `crates/stella-cli/src/subsession/`. `subsession/lane_events.rs` (added by
  #5158) is the pattern.
- A worker lane's durability is its own (`lane_durability`), never the lead's —
  see `run_worker`'s comment on `subsession_engine_config_for`. Anything that
[...]

Issue #5159, opened 2026-08-26
One board task, two types: `TaskItem::id` is a `String` while the wire tag is `TaskId`


macanderson:
#5039 introduced `stella_protocol::TaskId` — the newtype an event's `task_id`
carries — for the reason `LaneId` exists: this workspace already has several ids
that read alike (AGENTS.md's glossary), and an unwrapped `String` is how the
next one joins them.

It stopped one field short. `TaskItem::id`, the thing a `TaskId` **is**, is
still a `String`, so the one identity has two types and every crossing between
them is a `TaskId::new(item.id.clone())` or an `as_str()`. That is not a
correctness problem today — the conversion is total in both directions and
`TaskId` is `serde(transparent)`, so no wire byte changes either way — but it is
exactly the seam a typed id was supposed to remove, and it will read as an
oversight in a year rather than as a decision.

It was left out of #5039 on scope: `TaskItem` is referenced in ~74 places across
`stella-tui`, `stella-cli`, `stella-store`, `stella-tools` and `stella-core`, and
a type change there means compiling all of them — which is a different PR from
the one that adds the tag, not a bigger version of it.

The change is mechanical and mostly free: `TaskId` already carries
`From<String>`, `From<&str>`, `Display` and `PartialEq<str>`/`PartialEq<&str>`,
so `id: "1".into()`, `t.id == id`, `format!("{}", item.id)` and `item.id.clone()`
all keep compiling. What needs hands is the places that pass `&item.id` where a
`&str` is expected (`.as_str()`) and the store's `params![]` rows.

Done means: `TaskItem::id` is a `TaskId`, `TaskBoard::running` returns a
reference rather than minting a clone, `stella_protocol::task_id`'s module doc
loses the paragraph explaining the seam, and the workspace is green.

## Definition of done

The “Done means” prose above, as boxes the gate can read — nothing here is
new work, and nothing is ticked by the migration (#5193).

- [ ] `TaskItem::id` is a `TaskId`.
- [ ] `TaskBoard::running` returns a reference rather than minting a clone.
- [ ] `stella_protocol::task_id`'s module doc loses the paragraph explaining the seam.
- [ ] The workspace is green.

PR #5924, opened 2026-09-05
fix(stella-cli): tag a delegated lane's events with the task it was assigned

### LINKED ISSUE #5158: A `task_assign` worker lane emits untagged events, so a delegated task's ledger is empty
#5039 gave every work event a `task_id` and stamps it from the lane's board
(`stella_core::RunningTask`, attached to the turn's `EventSender` in
`stella-cli`'s `turn_files::open_turn_streams`). The lead lane is covered, and
so is every door that already drives its engine through the sender it hands the
registry.

A `task_assign` worker is not. `crates/stella-cli/src/subsession.rs` builds the
lane its own `ToolRegistry` with its own, empty board and attaches a plain
`EventSender::new(tx.clone())`, so `TaskBoard::running()` answers `None` and the
lane's tool calls, file changes and metering rows go out with no tag. The
consequence is one-directional but real: `Store::task_events` for a delegated
task comes back empty, and `Store::task_cost` reports `$0.00` for work that was
really done and really paid for.

Untagged is honest (the events are in no task's ledger) rather than wrong, which
is why #5039 shipped without it — but it is not complete.

The fix is small and the right shape is already there: the lane exists to work
exactly one board task, and `SpawnRequest::task_id` names it at spawn time. So
its source is a **constant**, not a board read:

```rust
let events = stella_core::EventSender::new(tx.clone());
events.attach_running_task(stella_core::RunningTask::from_fn(move || Some(task.clone())));
registry.attach_events(events.clone());
// ...and the turn is driven through `events`, not the raw `tx`.
```

The care needed is drop order, not logic: the lane's teardown closes its channel
by dropping every sender clone, and a held `EventSender` would wedge the
forwarder exactly as `command_deck::lead_turn`'s re-query adapter would (#2290).
`lead_turn` shows the shape — an explicit `drop` beside the adapter's.

Done means: a delegated task's `Store::task_events` returns its worker's rows,
with a test over a scripted lane, and the note in
`stella_core::tasks::RunningTask`'s doc citing this issue is removed.

## Definition of done

- [x] A delegated task's `Store::task_events` returns its worker's rows.
- [x] A test over a scripted lane witnesses it.
- [x] The note in `stella_core::tasks::RunningTask`'s doc citing this issue is removed.

## Verification by the PR sweep (2026-09-05)

Ticked against the diff on #5924 at `773c111`, read rather than taken from the
PR description.

**Box 1 — the ledger returns the worker's rows.**
`a_delegated_lanes_work_lands_in_that_tasks_ledger`
(`crates/stella-cli/src/subsession/lane_events.rs`) asserts both halves: that
`store.task_events("deck-1", TaskId::new("4"))` hands back **2** events for a
lane spawned as `sub:4`, and that every persisted row names the task
(`record.event.task_id() == Some("4")`). On the old code the sender carried no
running-task source, so `Store::task_events` filtered on `events.task_id = ?`
and matched none of them — the empty journal this issue was filed about.

The implementation is the constant the issue prescribed:
`open_lane_stream` attaches `RunningTask::from_fn(move || Some(task.clone()))`
only when `spec.board_task` is present, and one sender serves both the registry
and the engine, so neither half of the evidence is dropped.

**Box 2 — over a scripted lane.** `scripted_lane()` drives the shipping path,
not a mock of it: the real `spawn_forwarder`, the real `close_turn_stream`, a
real `ToolRegistry` and a real `Store`, with an execution row stamped with the
deck's session id because that is the join `Store::task_events` makes. Only the
model is absent. The negative direction is covered too —
`a_prompt_lane_names_no_task_and_stays_untagged` proves a `req:<n>` lane claims
no board id, so the fix cannot file a person's composer prompt in a stranger's
ledger.

The drop-order hazard this issue flagged as "the care needed" is handled and
witnessed by the same harness: `scripted_lane` performs the explicit
`drop(lane)` before `close_turn_stream`, and `open_lane_stream`'s doc comment
states why a surviving clone would leave the forwarder's `recv()` pending and
wedge the lane after the deck has painted it done.

**Box 3 — the doc note is gone.** `773c111` removed the `#5158` citation from
`RunningTask`'s doc comment; the "unresolved gap" framing it once carried had
already been rewritten earlier on the branch to describe the shipped behaviour.
What remains is the code pointer, `stella-cli`'s `subsession::lane_events`,
which is the half a reader needs. ADR 0015 keeps its own `#5158` link
deliberately — an ADR cites the issue that drove the decision as provenance, the
way the bullet beneath it cites #5159, and this box names the `RunningTask` doc
note alone.

macanderson:
## What and why

A `task_assign` worker lane emitted untagged events, so a delegated task's
ledger was empty.

`crates/stella-cli/src/subsession.rs`'s `run_worker` builds the lane its own
`ToolRegistry` with its own, empty task board, and attached a bare
`EventSender::new(tx.clone())` to it. `TaskBoard::running()` answered `None`,
so the lane's tool calls, file changes and metering rows went out with
`task_id: None`. `Store::task_events` for the delegated task came back empty
and `Store::task_cost` reported `$0.00` for work that had really run and
really been paid for.

The lane exists to work exactly one board task, and `SpawnRequest::task_id`
names it at spawn time, so its source is a **constant** rather than a board
read — the shape the issue asks for. A read would be wrong twice over: the
worker's board is empty, and board ids are per-session ordinals, so the
worker's `"1"` names a different task from the lead's `"1"`.

What changed:

- `SubSessionSpec` carries `board_task: Option<TaskId>` — `Some` for a
  `sub:<task-id>` lane, `None` for a `req:<n>` lane a person typed into the
  composer.
- New sibling module `crates/stella-cli/src/subsession/lane_events.rs` opens
  the one sender the lane's registry **and** its engine send through, and
  attaches the constant running-task source when the spec names a task. New
  logic lands here rather than in `subsession.rs`, which is 1429 lines against
  a 1500 ceiling.
- `run_worker` drives its turn with `run_turn_with_sender` on that sender
  rather than on the raw channel. It has to: a raw channel is wrapped in a
  second, unattached sender inside the engine, so the lane would tag its
  registry's events and none of its engine's — the shape
  `turn_files::open_turn_streams_raw` already names for the lead turn.
- `run_worker` drops that sender before `close_turn_stream`, the drop the lead
  turn owes its re-query adapter (#2290). Without it the forwarder's `recv()`
  never sees the channel close and the lane wedges after the deck has painted
  it done.
- The note in `stella_core::tasks::RunningTask`'s doc that cited this issue as
  an open gap is replaced by what the code now does; ADR 0015's consequence
  bullet likewise.

## Witness mechanism

`a_delegated_lanes_work_lands_in_that_tasks_ledger` in
`crates/stella-cli/src/subsession/lane_events.rs` scripts one lane end to end
with everything but the model: a real `Store`, a real execution row stamped
with the deck's session id (the join `Store::task_events` makes), the real
`spawn_forwarder`, the real `close_turn_stream`. It sends two `ToolStart`
events through the lane's sender and then asserts
`Store::task_events("deck-1", TaskId("4"))` returns both rows, each naming the
task.

It fails on `main` by construction: without a running-task source the sender
stamps nothing, every persisted row carries `task_id: NULL`, and
`task_events`' `WHERE e.task_id = ?` matches none of them — the journal comes
back empty and the length assertion fails at 0.

`a_prompt_lane_names_no_task_and_stays_untagged` holds the other direction: a
`req:<n>` lane must not borrow a board id.

## Exemplar

`crates/stella-cli/src/turn_files.rs`'s `open_turn_streams` /
`open_turn_streams_raw` — the same "the tag belongs to the stream, so the
registry and the engine share one sender" shape, and the same drop-order
obligation `command_deck::lead_turn` discharges beside its re-query adapter.

## Tests deleted

None.

## Residue issues filed

- #5923 — a deck worker lane's Files tab is always empty. `run_worker` never
  calls `turn_files::open_turn_streams` / `close_turn_boundary_raw`, so it
  attaches no `TurnCallMeasure` and emits no `FileChange`. Noticed while
  reading the lane's stream setup for this change; out of scope here because
  the fix has to answer how concurrent lanes in one work tree are attributed,
  which is a design question of its own.

Closes #5158

## Sourcery ❌ settled, and the DoD verified by the PR sweep (2026-09-05)

**The ❌ row.** Sourcery's assessment asked that the `RunningTask` doc note
citing #5158 as an unresolved gap be removed, and marked it unmet because a
bare `#5158` still stood in that doc. `773c111` drops it. The citation is worth
losing for a reason Sourcery's own wording did not reach — the gap framing had
already been rewritten, so its stated ground no longer held, but the number
resolves to a **closed** issue the moment this PR merges, and a bare issue
number beside a description of shipped behaviour reads to the next reader as
work still outstanding. `stella-cli`'s `subsession::lane_events` is the half a
reader needs, and it was already there.

ADR 0015 keeps its `#5158` link deliberately: an ADR cites the issue that drove
the decision as provenance, the way the bullet beneath it cites #5159, and the
checklist item names the `RunningTask` doc note alone. Reasoning recorded in
[this comment](https://github.com/macanderson/stella/pull/5924#issuecomment-5550944195).

**`dod / dod` was red** on #5158's three unticked boxes. All three are now
ticked, each checked against the diff at `773c111` rather than against this
description, with the evidence recorded in the issue body:

- **The ledger returns the worker's rows** — the witness asserts `task_events`
  hands back 2 rows for a `sub:4` lane and that each names task `"4"`.
- **Over a scripted lane** — `scripted_lane()` drives the real
  `spawn_forwarder` / `close_turn_stream` / `ToolRegistry` / `Store`, with only
  the model absent, and the negative case pins that a `req:<n>` lane claims no
  board id. The drop-order hazard the issue called "the care needed" is
  discharged and exercised by the same harness.
- **The doc note is gone** — as above.

re-read the boxes — the edit is the trigger, never the fix. **No code changed
in this edit.**

## Summary by Sourcery

Ensure delegated worker-lane events are attributed to the task they were spawned to execute.

Bug Fixes:
- Tag events emitted by delegated worker lanes with the task they were assigned so their work appears in the correct task ledger and cost records.

Enhancements:
- Use a shared, task-aware event sender for worker registries and engine turns while preserving untagged events for composer-requested lanes.
- Document the completed worker-lane task attribution behavior in the task model and ADR.

Documentation:
- Update task-tagging documentation to describe constant task attribution for delegated worker lanes.

Tests:
- Add end-to-end coverage verifying delegated worker events are recorded under the assigned task and prompt-lane events remain untagged.


macanderson:
## Settling the ❌ row

> Remove the `stella_core::tasks::RunningTask` documentation note citing this issue as an unresolved gap.

Sourcery reads this as two things; one is now fixed and one is answered.

**Fixed — the `RunningTask` doc note.** `773c111` drops the `#5158` from the parenthetical in `crates/stella-core/src/tasks.rs`, which now reads:

> Its host attaches a **constant** source instead, naming the one task the lane was spawned to work (`stella-cli`'s `subsession::lane_events`). Constant rather than a read for the same reason the lead's is a read: …

Sourcery was right that the citation should go, even though its stated reason ("still describes an unresolved gap") no longer held — the gap framing had already been rewritten. The reason it should go is the one it becomes true for on merge: `#5158` resolves to a **closed** issue the moment this PR lands, and a bare issue number beside a description of shipped behaviour reads to the next reader as work still outstanding. `subsession::lane_events` is the half that answers "where does the constant come from", and it was already there, so the number was the part carrying nothing.

**Answered — ADR 0015 keeps its link.** The checklist item names the `RunningTask` doc note, and an ADR is a different kind of document: it records *why a decision was taken*, and citing the issue that drove it is provenance rather than a stale pointer. The bullet directly beneath it cites #5159 in exactly that role. Removing #5158 there would lose the trail from the consequence back to the argument, which is what the ADR exists to preserve. If the intent was genuinely to strip the link from the ADR too, that is a change to how this repo writes ADRs and belongs in its own PR rather than folded in here.

Guards run on `773c111`: `check-prose.py` (OK, none added), `check-line-citations.py` (OK, none added), `check-file-size.sh` (OK). No compile evidence claimed locally — the change is comment text with no doc-link syntax, and CI is this PR's compile authority.

---

macanderson:
`the macOS-only dependency table still compiles` is red here, and it is not about macOS or about this diff. The job fails before compiling anything:

```
    Updating crates.io index
error: cannot update the lock file /Users/runner/work/stella/stella/Cargo.lock because --locked was passed to prevent this
```

That is #5939 — `Cargo.lock` records `stella-learn` at `0.9.329` while every manifest reads `0.9.330`, so every `--locked` build on `main`'s tip fails. This PR changes no manifest and no lockfile (its files are five Rust sources and one ADR), so the unresolvable lock can only be coming from the base.

The check's name is what makes this misleading: it reads as though the `cpal` dependency table broke on macOS. `cargo check on the declared MSRV` is red on other PRs for the same reason — both are `--locked` builds, and they are reporting the base's break, not the branch's.

No action needed here. #5940 carries the one-line repair and is armed to auto-merge; this check goes green on a re-run once it lands. `main is not known-broken` will clear at the same time.

---

Issue #5917, opened 2026-09-05
A plugin's name is still allowed to reorder itself with a bidi override


macanderson:
Noticed while landing #5184, which taught `PanelText::new` to refuse the bidi formatting characters. Not fixed there because it widens a public error enum that issue did not scope.

## Problem

A plugin's `name` is the one string Stella **composes into chrome it owns** — the `◳ panel · <plugin>` label, the install prompt, a popup heading, the rules panel. `validate_plugin_name` (`crates/stella-plugin/src/drawable.rs`) refuses control characters in it, so `"\x1b[2J"` cannot clear the screen from inside Stella's own border.

It still accepts `U+202E` RIGHT-TO-LEFT OVERRIDE and the eleven other bidi formatting characters. A plugin can therefore ship a name that renders in an order its bytes do not have — the Trojan Source shape, CVE-2021-42574 — and Stella prints it, under its own border, next to the consent prompt a person is reading in order to decide whether to trust that plugin. That is worse than the panel case #5184 closed: a panel's glyphs sit inside a leased rectangle the host clips, and this text sits in Stella's chrome.

The predicate already exists. `first_bidi_control` is in the same module, `pub(crate)`, and `validate_plugin_name` sits directly beneath it asking only `first_control_character`.

## Files

- `crates/stella-plugin/src/drawable.rs` — `validate_plugin_name`, whose doc comment names this issue as the thing that decides it.
- `crates/stella-plugin/src/error.rs` — `ManifestError`, where a second variant goes.
- `crates/stella-plugin/tests/panel_contract.rs` — the control-character half is asserted there, in the test that pins `ManifestError::NameNotDrawable`.

## The fix

Ask `first_bidi_control` in `validate_plugin_name` and return a new `ManifestError` variant naming the character and its position in `char`s.

A new variant rather than a reused `NameNotDrawable`, for the reason `PanelTextError::BidiControl` is its own variant in #5184: the two refusals answer different questions, and a message telling an author to look for an escape sequence that is not in their name wastes their time. `ManifestError` is public and re-exported from `lib.rs`, so this is an additive wire-visible change and wants the round-trip discipline the rest of that enum gets.

## Constraints

- `stella-plugin` is a near-leaf whose `Cargo.toml` requires a fresh argument for each workspace dependency. This needs none — the character list is already in `drawable.rs`.
- `error.rs` sits well under the 1500-line ceiling, so a variant fits; `drawable.rs` has room.
[...]

Issue #5604, opened 2026-08-31
The module layout script blocks the layout our own rules require


macanderson:
## Problem

The script `scripts/check-module-layout.py` blocks the wrong thing. It stops a code file from sitting next to a folder with the same name — for example `foo.rs` next to `foo/`. But that is the layout our own rules say to use.

The script's own error message points at `AGENTS.md` as its reason. `AGENTS.md`'s Code style section says the opposite:

> **A module with submodules is `foo.rs` beside `foo/`, never `foo/mod.rs`.** `src/anthropic.rs` next to `src/anthropic/` is the layout Rust 2018 introduced and the one this workspace uses.
>
> `mod.rs` is the pre-2018 spelling and is not used for library code.

`CLAUDE.md` says the same thing, and even uses this exact rule as its example of a house rule that outlived its own tree.

The tree agrees with the docs, not the script: the script's baseline file (`scripts/module-layout-baseline.txt`) already lists 234 pairs using the correct, docs-approved layout. The script only blocks *new* pairs, so the correct layout is now off-limits for anyone starting a new split, while every file that already uses it is left alone.

Even `AGENTS.md`'s own gate-step list contradicts its own Code style section a few hundred lines later — it still describes this step as "no code file beside a same-named folder," as if that were still the banned shape.

## Why this matters

When a file crosses the 1500-line ceiling (see `AGENTS.md` § "God files"), the standard fix is to split it into a folder named after the file. That fix is now blocked by this script. Whoever does the split has to pick a name the codebase would not otherwise choose, or fall back to the banned `mod.rs` style.

This already happened: splitting `crates/stella-cli/src/self_driving_cmd/work.rs` tried `work/claude.rs` first, got rejected by this script, and landed at `self_driving_cmd/claude_worker.rs` instead — a worse name chosen only to satisfy a guard that is itself wrong.

## Background

Issue #5478 already fixed this same wrong rule once: it corrected `AGENTS.md`, deleted a different script (`scripts/check-module-siblings.py`) that enforced the old rule, and merged as PR #5479. `scripts/check-module-layout.py` does the same job under a different name and was missed by that fix, so the bug is back under a new file name.

## Files

- `scripts/check-module-layout.py` — the script enforcing the wrong rule
- `scripts/module-layout-baseline.txt` — the 234 pairs it currently treats as debt, when they are already correct
- `scripts/test-module-layout.sh` — the script's own test harness
[...]

PR #5928, opened 2026-09-05
fix(stella-plugin): refuse a bidi override in a panel's glyphs

### LINKED ISSUE #5184: A panel's no-ANSI check refuses control characters but not bidi or zero-width format characters
Noticed while landing #5054 (the panel wire contract); not fixed there because it is a widening of the rule, not a hole in what #5054 committed to.

## Problem

`PanelText::new` (`crates/stella-plugin/src/panel.rs`) is the single constructor that makes "a panel never emits raw ANSI" a property of the type. It refuses `char::is_control()`, which is Unicode General_Category `Cc` — every C0 byte, `DEL`, and the C1 range including `U+009B` CSI. That is the whole escape-sequence class, and it is correct as far as it goes.

It does **not** refuse General_Category `Cf` (format), which is a different spoofing class inside the same rectangle:

- `U+202E` RIGHT-TO-LEFT OVERRIDE and `U+202D` LEFT-TO-RIGHT OVERRIDE reorder the glyphs after them, so a panel can render text whose visual reading differs from its logical content (the Trojan Source shape, CVE-2021-42574).
- `U+2066`–`U+2069` (the isolates) do the same with a scope.
- `U+200B` ZERO WIDTH SPACE and `U+200D` ZWJ occupy a cell in `PanelText::cells()` and none on screen, so a panel's own width accounting and the terminal's disagree.

Nothing here escapes the leased rectangle — the host clips every blit — so this is a legibility and trust problem inside a panel's own box, not an escape from it. It matters because a panel is chrome a user is asked to trust after a consent handshake (`design/tui-v2/SPEC.md` §12): text that reads one way and means another is exactly what that handshake is supposed to make impossible.

## Files

- `crates/stella-plugin/src/panel.rs` — `PanelText::new`, and the type's doc comment, which currently claims the constructor "refuses every control character" and should say which classes it refuses once this is settled.
- `crates/stella-plugin/tests/wire_contract.rs` — `a_panel_frame_carrying_an_escape_sequence_does_not_decode` is where the wire half of the refusal is asserted.

## Constraints

- `stella-plugin` is a near-leaf and its `Cargo.toml` requires a fresh argument for each workspace dependency. This needs no new one: `char::is_control()` has a sibling in `matches!(ch, '\u{200b}'..)`-style range checks, or `unicode-ident`-free explicit ranges. Do not reach for a Unicode crate without making that argument.
- Whatever is refused must be refused in `PanelText::new` itself, so `Deserialize` (which routes through it) refuses it on the wire too. A check anywhere else is a check a host can skip.
- `PanelTextError` is a typed error per AGENTS.md #5; a second refusal class wants its own variant naming the character and its position, not a reused `ControlCharacter`.

## Decide before implementing

Whether to refuse `Cf` wholesale or only the bidi controls. `U+200D` ZWJ is load-bearing in emoji sequences, and refusing it means a panel cannot draw a family emoji or a profession emoji. A defensible split is: refuse the bidi formatting characters (`U+061C`, `U+200E`, `U+200F`, `U+202A`–`U+202E`, `U+2066`–`U+2069`) and allow the joiners. Record the choice as an ADR-style paragraph in the type's doc comment, as the rest of that module does.

## Done means

A frame carrying `U+202E` is refused by `PanelText::new` and fails to decode as a `PanelResponse`, with a test in `tests/wire_contract.rs` asserting both; the type's doc comment states exactly which Unicode classes it refuses and why the allowed ones are allowed; `make prose`, `make lint-schema` and `cargo test -p stella-plugin --features schema` are green.

## Definition of done

- [x] A frame carrying `U+202E` is refused by `PanelText::new` and fails to decode as a `PanelResponse`, with a test asserting both — `a_panel_frame_carrying_a_bidi_override_does_not_decode` in `crates/stella-plugin/tests/panel_contract.rs`, which is where the panel half of `tests/wire_contract.rs` moved when that file met the size ceiling.
- [x] The type's doc comment states exactly which Unicode classes it refuses, and why the allowed ones are allowed. `PanelText`'s docs name `Cc` and the bidi list; `drawable.rs`'s header carries the argument for the joiners and `U+200B`.
- [x] `make prose`, `make lint-schema` and `cargo test -p stella-plugin --features schema` are green. `make guards-fast` (prose included) ran clean locally; on PR #5928 the `docs/wire still matches the types` job (which is where `lint-schema` and `doc-warnings-schema` run, with the `schema` feature on) and `fmt + clippy + test` are both green.

macanderson:
## What & why

`PanelText::new` is the single door that makes "a panel never emits raw ANSI" a property of the type. It asked `char::is_control()` alone — Unicode `Cc`, the escape-sequence class — and left `Cf` open.

`U+202E` RIGHT-TO-LEFT OVERRIDE and its siblings reorder the glyphs after them, so a frame could render text whose visual reading is not its content: the Trojan Source shape, CVE-2021-42574. Nothing escaped the leased rectangle, because the host clips every blit, so the harm sat inside a panel's own box — which is chrome a person agreed to trust after a consent handshake (`design/tui-v2/SPEC.md` §12).

The constructor now refuses the bidi formatting characters as well: `U+061C`, `U+200E`, `U+200F`, `U+202A`–`U+202E`, `U+2066`–`U+2069`. `Deserialize` routes through the same constructor, so a frame carrying one fails to decode rather than arriving as a value the host is trusted to inspect.

Closes #5184

### The decision the issue asked for

Refuse the bidi controls; allow every other `Cf`. `U+200D` ZWJ builds every family and profession emoji and `U+200C` ZWNJ is ordinary Persian and Indic text — refusing them bans writing systems to tidy a count. `U+200B` stays for the same reason the type already accepts a double-width glyph: it costs a cell in `PanelText::cells()` and none on screen, which misaligns a panel's own interior and nothing outside it. That caveat is already written on `cells()`. The full argument is in `drawable.rs`'s module header, and the type's own doc comment names the classes it refuses and why the rest are allowed. The decision was posted on the issue before any code was written.

### A second typed error, not a reused one

`PanelTextError::BidiControl { index, code }` names the character and its position in `char`s. Not a reused `ControlCharacter`: the two refusals answer different questions, and a host printing the wrong one sends an author hunting an escape sequence that is not in their frame.

### Where the code went

`crates/stella-plugin/src/drawable.rs` is a new **flat sibling** module holding `first_control_character`, the new `first_bidi_control`, and `validate_plugin_name` — the one definition of what Stella will draw, read by both the panel channel and the manifest's `name` rule.

`panel.rs` sat 18 lines under the 1500-line ceiling, so the split had to happen. It is flat rather than a child module under `panel/` because `make module-layout` rejects a new `foo.rs` beside `foo/` (#5604), which is the layout the size ceiling would otherwise ask for. `panel.rs` ends at 1476 lines.

No new dependency. The bidi set is closed and has not moved since Unicode 6.3 added the isolates, so a `const &[char]` answers it — `stella-plugin` is a near-leaf whose `Cargo.toml` takes a workspace dependency only on argument, and a crate for that list is not one.

## The witness

`a_panel_frame_carrying_a_bidi_override_does_not_decode`, in `crates/stella-plugin/tests/panel_contract.rs`.

**Why it fails on the old code by construction:** the test decodes a `PanelResponse` whose `text` is `"gates ‮neerg"`, written as the JSON escape a plugin's own encoder would put on the pipe. Before this change `char::is_control()` returns false for every character in `BIDI_CONTROLS` (a unit test in `drawable.rs` pins that fact), so `PanelText::new` accepted the string and the frame decoded cleanly — `expect_err` panics. It runs the whole refused set, both frame shapes (`lines` and `diff`), and the Rust half (`PanelText::new` directly, plus the position and code point in the message). It also asserts what the rule keeps: `U+200B`, `U+200C`, `U+200D` and a ZWJ emoji sequence still construct.

The issue named `tests/wire_contract.rs`. The panel half of that file was split into `tests/panel_contract.rs` when it met the ceiling, and `a_panel_frame_carrying_an_escape_sequence_does_not_decode` — the sibling this test is written beside — lives there now.

`drawable.rs` carries the unit half: every listed character is found, the joiners are not, the list is sorted and unique, and no listed character is a `Cc` control (so neither predicate can stand in for the other).

**Tests deleted: None.**

## The gate

- [x] `cargo fmt --check` (`cargo fmt --all`, then `make guards-fast` green locally)
- [ ] `cargo test --workspace` — CI
- [x] Docs updated: `panel.rs`'s module header and `PanelText`'s doc comment state which Unicode classes are refused and why the rest are allowed; `crates/stella-plugin/README.md` gains the new module and the widened rule
- [x] `Closes #5184` appears both here and as a commit trailer

Not built or tested on the maintainer's laptop (CLAUDE.md, "CI builds and tests; this laptop does not"). The evidence is this PR's run.

## Nothing left behind

- [x] Filed: #5917 — a plugin's `name` is composed into Stella's **own** chrome and `validate_plugin_name` still accepts `U+202E`. That is the same spoof one surface over, and worse, since a name is not clipped to a leased rectangle. Fixing it adds a public `ManifestError` variant, which is a wire-visible change this issue did not scope, so it is its own handoff. `validate_plugin_name`'s doc comment says plainly that it asks only the control-character predicate and why.

## Ground-rule check

- [x] The wire shape is unchanged — `PanelText` still serializes as a string; what changed is which strings decode, and the round-trip tests in `panel_contract.rs` cover it

Two prose fixes rode along in `tests/panel_contract.rs`, both in files this PR already touches: a stray copy of `wire_contract.rs`'s header was sitting above the real one (a leftover from the split), and the header's rule list now names the third rule this PR adds.

Exemplar for the predicate module: `stella-plugin`'s own `runtime.rs` / `consent.rs` shape — a small module of pure functions over borrowed text with the argument in the header, which is what the rest of this near-leaf crate already is.

## Summary by Sourcery

Refuse Unicode bidirectional controls in panel glyphs so decoded frames cannot visually reorder their own text.

New Features:
- Reject Unicode bidirectional formatting characters from panel text to prevent visual text reordering and Trojan Source-style spoofing.

Bug Fixes:
- Prevent panel frames containing bidi controls from decoding through either line-based or diff-based wire formats.

Enhancements:
- Introduce shared drawable-text validation helpers in a dedicated module and retain support for non-bidi format characters needed by common scripts and emoji sequences.
- Add a dedicated typed error identifying the refused bidi character and its character position.

Documentation:
- Document the panel glyph restrictions, rationale, and shared drawable-text validation module.

Tests:
- Add unit and wire-contract coverage for the complete bidi-control set, error details, both panel frame shapes, and explicitly allowed joiner characters.

### DIFF
--- a/crates/stella-plugin/README.md
+++ b/crates/stella-plugin/README.md
@@ -271,10 +271,14 @@ before it crosses.
   (`PanelGrant`, `PanelDenial`, `PanelLease`, `PanelFrame`): a plugin leased a
   rectangle of the screen, returning styled lines or a cell diff each tick
   (`design/tui-v2/SPEC.md` §12). Two of that section's rules are the types
-  rather than a host's care — `PanelText` wraps a private `String` that refuses
-  every control character, so no plugin emits an escape sequence in any
-  language, and `PanelRect` carries an extent with no origin, so a frame cannot
-  name a cell of Stella's own chrome.
+  rather than a host's care — `PanelText` wraps a private `String` that turns
+  away every control `char` and every bidi one, so no plugin emits an escape
+  sequence or flips its own text, and `PanelRect` carries an extent with no
+  origin, so a frame cannot name a cell of Stella's own chrome.
+- `src/drawable.rs` — the two rules behind that first one, and the argument for
+  where each line sits: what a screen runs (`Cc`), and what reorders the text
+  after it (the bidi `char`s). `PanelText::new` asks both; a plugin's `name`
+  asks the first.
 - `src/error.rs` — `ManifestError`, typed per rule (AGENTS.md #5).
 - `tests/manifest_grades.rs` + `tests/fixtures/*.toml` — slice A's
   acceptance: one fixture per grade, round-tripped through both TOML and
--- a/crates/stella-plugin/src/drawable.rs
+++ b/crates/stella-plugin/src/drawable.rs
@@ -1,0 +1,144 @@
+//! What Stella will draw.
+//!
+//! One rule, read wherever a plugin's text reaches a screen. Two kinds of
+//! `char` are turned away, for two reasons.
+//!
+//! A control `char` is an escape. `\x1b` opens every `CSI`, `OSC` and `SGR`
+//! run. `U+009B` opens one on its own. `\n`, `\r` and `\t` set a row's shape.
+//! A screen obeys them all. Stella writes every escape byte a screen sees.
+//!
+//! A bidi `char` is a lie. `U+202E` and its kin flip the text after them. A
+//! run can then read in an order its bytes do not have. That is the Trojan
+//! Source shape (`CVE-2021-42574`). None of it leaves the leased box, since
+//! the host clips each blit. It still counts. The box is chrome a person
+//! chose to trust.
+//!
+//! The rest of `Cf` stays. `U+200D` builds each family and job emoji. `U+200C`
+//! is plain text in Persian and in `Indic` scripts. To bar them is to bar a
+//! script. `U+200B` takes a cell here and none on a screen. That is the gap a
+//! wide glyph already opens; [`crate::PanelText::cells`] says so.
+
+use crate::error::ManifestError;
+
+/// Every bidi `char`, in code point order.
+///
+/// Written out, not looked up. The set is closed and has not moved since
+/// `Unicode 6.3` added the isolates. This crate takes a new dep only on a
+/// good case, and a whole crate for the list below is not one.
+const BIDI_CONTROLS: &[char] = &[
+    '\u{061c}', // ARABIC LETTER MARK
+    '\u{200e}', // LEFT-TO-RIGHT MARK
+    '\u{200f}', // RIGHT-TO-LEFT MARK
+    '\u{202a}', // LEFT-TO-RIGHT EMBEDDING
+    '\u{202b}', // RIGHT-TO-LEFT EMBEDDING
+    '\u{202c}', // POP DIRECTIONAL FORMATTING
+    '\u{202d}', // LEFT-TO-RIGHT OVERRIDE
+    '\u{202e}', // RIGHT-TO-LEFT OVERRIDE
+    '\u{2066}', // LEFT-TO-RIGHT ISOLATE
+    '\u{2067}', // RIGHT-TO-LEFT ISOLATE
+    '\u{2068}', // FIRST STRONG ISOLATE
+    '\u{2069}', // POP DIRECTIONAL ISOLATE
+];
+
+/// The first control `char` in `text`, as `(index in chars, character)`.
+///
+/// Read in two places. [`crate::PanelText::new`] holds it for a panel's own
+/// text, and [`validate_plugin_name`] holds it for the name Stella prints. A
+/// second copy would drift. Both ask one thing: can this string reach a screen
+/// with no escape in it?
+pub(crate) fn first_control_character(text: &str) -> Option<(usize, char)> {
+    text.chars()
+        .enumerate()
+        .find(|(_, found)| found.is_control())
+}
+
+/// The first bidi `char` in `text`, as `(index in chars, character)`.
+///
+/// The index counts `char`s, as [`first_control_character`] does. A refusal
+/// then names a spot a reader can count to.
+pub(crate) fn first_bidi_control(text: &str) -> Option<(usize, char)> {
+    text.chars()
+        .enumerate()
+        .find(|(_, found)| BIDI_CONTROLS.contains(found))
+}
+
+/// Whether a plugin's name is one Stella can print in its own chrome.
+///
+/// The name is the one string Stella puts in chrome it owns: the
+/// `◳ panel · <plugin>` label, the install prompt, a popup head, the rules
+/// panel. [`crate::PanelText`] keeps a panel's own text free of escapes. That
+/// is worth nothing if the label around it is not. A plugin named `"\x1b[2J"`
+/// wipes the screen from inside Stella's own border.
+///
+/// Asked once, at load, not at each reader. Each reader is a place to forget.
+/// Here, not in `manifest.rs`, so it sits by the rule it shares with
+/// [`crate::PanelText::new`].
+///
+/// It asks [`first_control_character`] and not [`first_bidi_control`]. A name
+/// with `U+202E` is the same trick one door over. To bar it adds a case to the
+/// public [`ManifestError`], which is its own change.
+///
+/// # Errors
+///
+/// [`ManifestError::NameNotDrawable`], naming the `char` and where it sits.
+pub(crate) fn validate_plugin_name(name: &str) -> Result<(), ManifestError> {
+    match first_control_character(name) {
+        Some((index, found)) => Err(ManifestError::NameNotDrawable {
+            index,
+            code: found as u32,
+        }),
+        None => Ok(()),
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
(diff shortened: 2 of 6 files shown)


macanderson:
**DoD audit against #5184** — automated PR sweep. The `dod` gate is red here because #5184's three boxes are unticked; two of them are settled by this diff, one is not met *as written*.

**Item 2 — doc comment names the refused classes and justifies the allowed ones: satisfied.** `PanelText`'s doc comment now enumerates exactly the set `BIDI_CONTROLS` holds (`U+061C`, `U+200E`, `U+200F`, `U+202A`–`U+202E`, `U+2066`–`U+2069`) and gives the reason the rest of `Cf` stays — ZWJ for emoji runs, ZWNJ for Persian and Indic text. The `# Errors` section and the doctest moved with it.

The issue's separate ask for a typed variant is also met, and it is a new one rather than a reused `ControlCharacter`: `PanelTextError::BidiControl { index, code }`, whose message names the character as `U+{code:04X}` and its position in `char`s.

**Item 1 — not met as written.** The DoD names `crates/stella-plugin/tests/wire_contract.rs`. Both halves are asserted, but in `crates/stella-plugin/tests/panel_contract.rs`, in `a_panel_frame_carrying_a_bidi_override_does_not_decode` — the constructor refusal and the `serde_json::from_str::<PanelResponse>(…).expect_err(…)` decode refusal. `tests/wire_contract.rs` still exists on this branch and this diff does not touch it.

Substantively this looks right, and splitting the panel contract into its own file is a defensible reason. But the box and the tree disagree, so one of the two should move: either relocate the assertions to `wire_contract.rs`, or amend #5184's item 1 to name `panel_contract.rs` and say why the file split. Ticking it as it stands would record a test location that isn't there.

**Item 3 — no evidence either way yet.** `fmt + clippy + test` is still `in_progress` on `ee9aadb`, and it is the only job that would cover `make lint-schema` or `cargo test -p stella-plugin --features schema`. No check run is named for either. Not a defect in the diff — just not yet demonstrable, so it should stay unticked until that job reports.

One thing this audit could not settle from the diff: that `PanelText`'s `Deserialize` routes through `new`. That impl is pre-existing and not shown here; the `expect_err` in the new test would establish it, once the test has actually run.

No changes pushed and nothing ticked — flagging only, since item 1 needs an author decision.

---

macanderson:
Answering the DoD audit above, and recording the Sourcery result.

**Item 1 — settled by amending the box, not by moving the test.** The DoD named `crates/stella-plugin/tests/wire_contract.rs`, which was correct when #5184 was filed. The panel half of that file has since been split into `crates/stella-plugin/tests/panel_contract.rs` — the file's own header says so, and `a_panel_frame_carrying_an_escape_sequence_does_not_decode`, the escape-sequence sibling this new test is written beside, lives there now. Moving the assertions back to `wire_contract.rs` would separate the two halves of one rule and push a file that already met the size ceiling back over it. So #5184's box now names `panel_contract.rs` and says why the path moved. The assertion itself is unchanged: `PanelText::new` refuses `U+202E` directly, and `serde_json::from_str::<PanelResponse>(…)` fails to decode a frame carrying one, on both frame shapes and over the whole refused set.

**Items 2 and 3 are ticked with the evidence beside them** — `PanelText`'s doc comment for the class list, `drawable.rs`'s header for why the joiners and `U+200B` stay, and this PR's own green runs for the commands (`docs/wire still matches the types` is where `lint-schema` and `doc-warnings-schema` run with the `schema` feature on).

**Sourcery:** the "Assessment against linked issues" table is three ✅ rows and no ❌, so there is nothing outstanding there.

**Residue:** #5917 — `validate_plugin_name` still accepts `U+202E` in a plugin's `name`, which Stella composes into its own chrome. Same spoof, one surface over, and not clipped to a leased rectangle. It needs a new public `ManifestError` variant, so it is its own change; the function's doc comment says plainly that it asks only the control-character predicate and why.

Issue #3830, opened 2026-08-19
The agent dev container image is not defined in this repository, so #3615's preferred fix (install shellcheck in the image) cannot be made here


macanderson:
## Problem

#3615 offers two fixes, "either, not both", and names the first as preferred:

> 1. Install `shellcheck` in the dev container image (**preferred** — it makes the gate work as advertised again), or
> 2. Have the `shellcheck` target detect a missing binary and fail with an explicit "shellcheck unavailable — this step did NOT run" notice. Note this is strictly worse than (1) …

The PR closing #3615 shipped **option 2 only**, because option 1 is not reachable from this repository. This issue tracks the half that was not done, so the preference recorded in #3615 does not quietly evaporate when that issue closes.

## Evidence that the image is not in-tree

Run from a clean checkout:

```
$ find . -iname 'devcontainer*' -not -path './target/*' -not -path './node_modules/*'
(no output)

$ ls -a | grep -i devcont
(no output)

$ find . -iname 'Dockerfile*' -not -path './target/*' -not -path './node_modules/*'
./packaging/docker/Dockerfile.serve
./arenabench/recorder/Dockerfile
./arenabench/infra/runner/Dockerfile
./bench/readiness/synthetic-adapter-sentinel/environment/Dockerfile
```

None of those four is the image agents run in: they build the `stella-serve` container, the arenabench recorder, the arenabench task runner, and a bench readiness sentinel respectively. Adding `shellcheck` to any of them would not put it on the PATH of a dev session, and doing so would be cargo-culting the fix into an unrelated image.

So the image is defined **outside this repository** — in whatever provisions the agent execution environment.

## Why it still matters now that option 2 has landed

Option 2 makes the failure visible: `make shellcheck` now prints `shellcheck: UNAVAILABLE — THIS STEP DID NOT RUN` and exits non-zero instead of dying with a `not found` / `Error 127` that reads like a lint finding. That is strictly better than before, and it is what #3615's own text calls "strictly worse than (1)".

The residual cost is real and unchanged: **`make gate` still cannot complete in the image**, on any rung of the ladder. A session in that image can run every other guard and still not clear the gate, which is the practical problem #3615 opened with — agents fall back to running steps by hand, and a step they forget is a step nobody ran.

## Definition of done

Either:

- `shellcheck` is present in the image agents run in, so `make guards-fast` completes there; **or**
[...]

PR #5931, opened 2026-09-05
fix(scripts): stop check-rebase-replay reading an inherited rename as an undo

### LINKED ISSUE #5890: check-rebase-replay reads a base-branch rename absorbed through a merge as an edit-and-undo pair

### LINKED ISSUE #5910: check-rebase-replay flags a path that was moved or deleted upstream, where the revert hazard cannot occur
## Problem

`scripts/check-rebase-replay.sh` fails a branch whose only "edit and undo" is a path that **upstream deleted or relocated**. The #4979 hazard it guards cannot occur for such a path, so the failure is a false positive and the prescribed remedy (flatten the history, force-push) pays a history rewrite for nothing.

Live instance: PR #5840, job https://github.com/macanderson/stella/actions/runs/33949398655/job/101260704974

```
check-rebase-replay:   crates/stella-core/src/records/validate.rs
```

## Why it is a false positive

The guard computes (`check-rebase-replay.sh`, the `net` / `touched` / `suspects` block):

```sh
net      = git diff --name-only  "$base" "$head"
touched  = git log --no-merges --name-only "$base..$head"
suspects = comm -23 touched net       # touched, but absent from the net diff
```

A path lands in `suspects` whenever some non-merge commit touched it and the branch's net diff does not mention it. The script does not ask **why** the net is empty, and there are two different reasons it can be:

1. **The branch edited the path and undid it.** The real hazard. Rebasing onto a base that carries somebody else's edit to that path drops the branch's copy (`skipped previously applied commit`), keeps the undo, and the result is a clean-merging silent revert.
2. **The path does not exist upstream any more.** No upstream copy exists, so there is nothing for an undo to revert. A rebase makes the old commit fail to apply — a conflict, which is loud.

#5840 is case 2, and it is checkable directly. The merge base `913e21a06` **is** #5829, *"refactor(stella-records): extract the record plane from the engine"*:

```console
$ git cat-file -e 913e21a06:crates/stella-core/src/records/validate.rs; echo $?
1                                  # absent at base
$ git cat-file -e 99f7e11:crates/stella-core/src/records/validate.rs; echo $?
1                                  # absent at head
```

The branch edited the file at its old path before the extraction landed, merged `main` (which moved the record plane to `stella-records`), and re-landed the work at the new path. Its net diff is `crates/stella-records/src/records/validate.rs | 165 ++++++--`, so nothing was lost.

## Proposed fix

> **Corrected 2026-09-05 while implementing this.** The condition first proposed here — drop a suspect absent at **both** ends — is **not safe**, and the suite already contained the counterexample. **R4** ("a file created and deleted inside the branch is flagged") has exactly that signature: `g.txt` is absent at base and at head, and it is a genuine hazard — rebase onto an upstream that added the same path and the branch's deletion is what survives. Skipping on absence alone turns R4 green and ships the hazard.
>
> What separates the two cases is **who removed the path**. A deletion in the branch's own non-merge history is the branch undoing its work; no such commit means the path left through a merge, which is the base's doing. The implemented condition is absence at both ends **and** no non-merge deletion on the branch:
>
> ```sh
> upstream_removed() {
>   local path="$1"
>   if git_c cat-file -e "$base:$path" 2>/dev/null; then return 1; fi
>   if git_c cat-file -e "$head:$path" 2>/dev/null; then return 1; fi
>   if [ -n "$(git_c log --no-merges --diff-filter=D --format=%h "$base..$head" -- "$path")" ]; then
>     return 1
>   fi
>   return 0
> }
> ```

Keep the check strict in the other direction: a path present at `$base` and absent at `$head` is a real deletion and must stay flagged, and a path present at both with an empty net diff is the original hazard.

This is narrow on purpose. Widening it to "absent at `$head`" alone would waive the genuine case where a branch deletes a file upstream still has.

## Files

- `scripts/check-rebase-replay.sh` — the `suspects` computation and the report loop.
- `scripts/check-rebase-replay-test.sh` (whatever `make rebase-replay-test` drives) — the hermetic suite that reproduces the rebase.
- `.github/workflows/rebase-replay.yml` — no change expected; it needs `fetch-depth: 0`, which it already has.

## Verify

`make rebase-replay-test` must keep failing on its existing blocking branch (the real hazard is still caught), and gain a case for this one:

1. Base commit A creates `foo/bar.rs`.
2. Branch commit B edits `foo/bar.rs`.
3. Upstream commit C **deletes or renames** `foo/bar.rs`; merge C into the branch.
4. Branch commit D re-lands the work at the new path.
5. `check-rebase-replay.sh A HEAD` exits **0** — today it exits 1.

That case fails on the current script and passes on the fixed one, which is the witness this repo asks for.

## Definition of done

- [x] `check-rebase-replay.sh` skips a suspect the **base** removed — absent at both ends *and* with no non-merge deletion on the branch — with a comment saying why that is the safe condition and why absence alone is not.
- [x] `make rebase-replay-test` covers the moved-upstream case and still covers the genuine hazard.
- [x] `make rebase-replay` is green on #5840's head with no history rewrite.
- [x] The reasoning is recorded where the guard's own prose lives (AGENTS.md § `rebase-replay.yml`), so the next reader knows the two reasons a net diff can be empty.

### Evidence (PR #5932, commit `354ce8b`)

**Box 1** — `upstream_removed` in `scripts/check-rebase-replay.sh`, consulted per suspect before it is reported. The comment block above it states both reasons a net diff can be empty and why absence at both ends does not separate them, naming R4. A run that clears every suspect prints which paths it cleared rather than the bare OK line.

**Box 2** — the suite grew from 11 cases to 13 and all pass. R9 builds the upstream-move shape and expects OK; R10 puts a genuine pair on that same branch and still expects a failure, so the skip cannot become a blanket waiver; R4 is untouched and still red. The flip was reproduced rather than asserted:

```
old script + new suite:   passed 12, failed 1   ← FAIL R9 ... expected OK
new script + new suite:   passed 13, failed 0
```

**Box 3** — run at `origin/main` against both live instances:

| | PR #5894 head (`c2e66fe`) | PR #5840 head (`56a966c`) |
|---|---|---|
| old guard | exit 1 (the red in CI) | exit 1 |
| new guard | **exit 0** | **exit 0** |

No history was rewritten on either branch.

**Box 4** — AGENTS.md § "A ninth, `rebase-replay.yml`" gains a paragraph on the two reasons, the who-removed-it discriminator, and R4/R9/R10. The issue numbers are left out of that prose because `check-prose`'s `issue-reference` ratchet is down-only; the change removes 4 grandfathered constructions overall.

Guards run locally: `check-prose` (OK, none added), `check-line-citations`, `check-file-size`, `check-left-behind`, `check-invariants`, `bash -n`. **`shellcheck` is not installed in that environment, so that step did not run there** and is left to CI rather than reported as passing.

## Context

- #4979 — the hazard this guard exists for.
- #5829 — the record-plane extraction that relocated the path.
- #5840 — the PR failing on this today; commented there rather than flattening, so the evidence survives.
- AGENTS.md § "A ninth, `rebase-replay.yml`" — the guard's normative description.

Found while triaging open PR CI. No priority or size label applied, per SCR-005.

macanderson:
## What & why

`check-rebase-replay` flags a path some branch commit touched that the branch's final diff no longer mentions. That is #4979's hazard when **the branch** removed the path, and a false positive when **the base branch** renamed or deleted it and the branch absorbed that through a merge — the ordinary way a long-lived branch survives a refactor.

The record-plane extraction (#5829) produced two live instances in one morning:

| PR | Flagged path |
| --- | --- |
| #5840 | `crates/stella-core/src/records/validate.rs` |
| #5894 | `crates/stella-core/src/context_record/lifecycle.rs` |

Both branches had followed the move correctly. The guard offers no acknowledgement by design, so its printed remedy would have charged each one a history rewrite — on a shared branch — for a hazard that was not there.

Closes #5910
Closes #5890

#5890 and #5910 are the same defect filed 41 minutes apart, against the first and second instance. Both DoDs are ticked with the evidence recorded on each issue.

> **Note on the closing keywords.** `Closes #5910` is on the commit trailer as well as here, so it closes on either merge path. `Closes #5890` is in this description only — it was identified as the duplicate after the commit was written, and this repo forbids an empty commit while a force-push to re-trailer it would restart a full CI cycle for a second keyword the linked-issue mechanism already carries. If the squash path drops it, #5890 needs closing by hand.

### Why not the fix the issue proposed

#5910 suggests dropping a suspect **absent at both `$base` and `$head`**. That would have silently waived the guard's own R4 case: a file the branch **created and then deleted** is also absent at the merge base and absent at head. Tree state cannot separate the two shapes, so that patch buys the false positive by discarding a genuine hazard — which is the direction the issue itself says must stay strict.

Confirmed by running it, not by reading it: with that condition in place, R4 (*"a file created and deleted inside the branch is flagged"*) flips from failing the guard to passing it.

#5890's variant — compare the **merge base** against the **base tip** — fails for a different reason: once the branch has merged the base, the merge base advances past the rename, so the path is absent at the merge base too. Both live instances are in that state.

What differs between the shapes is **who removed the path**. So the guard asks that:

```sh
if ! git cat-file -e "$head:$path" 2>/dev/null &&
  [ -z "$(git_c log --no-merges --diff-filter=D --format=%H "$base..$head" -- "$path")" ]; then
  continue
fi
```

A suspect is dropped only when it is gone at the head **and** no non-merge commit of the branch's own deleted it. An inherited removal is not in the range at all, because merging the base advances the merge base past it. The hazard cannot hide there — it needs the branch to contribute the removal a rebase would keep.

I recorded the correction on #5910 and #5890 so neither outlives this PR carrying advice that would regress the guard.

## The witness

**R9** in `scripts/test-rebase-replay.sh` builds the exact shape — branch edits a file, upstream moves it, branch merges and re-lands the work at the new path — and asserts the guard passes. **R9b** asserts the edit really survived the move, so R9 cannot pass by the work having been lost.

The flip, run rather than described:

```console
$ git stash push -q scripts/check-rebase-replay.sh && bash scripts/test-rebase-replay.sh
  ...
  passed 12, failed 1          # R9 fails on the unfixed guard

$ git stash pop -q && bash scripts/test-rebase-replay.sh
  passed 13, failed 0
```

**Against the live range CI actually ran for #5894** (`bfca7973c..c2e66fe90`):

```console
$ ./scripts/check-rebase-replay.sh bfca7973c c2e66fe90     # old guard
check-rebase-replay:   crates/stella-core/src/context_record/lifecycle.rs
EXIT=1

$ ./scripts/check-rebase-replay.sh bfca7973c c2e66fe90     # this branch
check-rebase-replay: OK — every path missing from the branch's diff was renamed
or deleted by the base branch, not by this branch.
EXIT=0
```

The genuine hazard is still caught: **R1** (edit undone), **R4** (created and deleted inside the branch), and **R2**, which reproduces #4979's rebase to completion rather than asserting it.

## The gate

Rust is untouched — this is two shell scripts and one `.md`, so `fmt`/`clippy`/`test` have nothing to compile. Per SCR-001 and CLAUDE.md's "CI builds and tests; this laptop does not", the compile evidence is this PR's CI run. Run here:

- [x] `bash scripts/test-rebase-replay.sh` — 13 passed
- [x] `scripts/check-prose.py` — exit 0, including the readability-grade ratchet
- [x] `scripts/check-line-citations.py` — OK, none added
- [x] `scripts/check-file-size.sh`, `check-left-behind.sh`, `check-gate-parity.sh`, `check-god-files.sh` — all exit 0
- [x] `bash -n` on both scripts
- [x] Docs updated — AGENTS.md § `rebase-replay.yml` now states the two reasons a net diff can be empty and the discriminator
- [x] `Closes #5910` appears both here and as a commit trailer

`shellcheck` is **not installed in this environment**, so that gate step did not run here; CI runs it. That gap is #3830.

No prose baseline entry was added: both files came back over the ratchet on first write and the prose was cut instead, which is why the comments carry the reasoning without the issue numbers.

## Nothing left behind

#4997 (make this guard a required check) is now unblocked by this fix but is a separate decision, and stays where it is.

## Ground-rule check

The discriminator deliberately does **not** use `--find-renames` to follow the path to its new home. Confirming the branch's edit survived the move would be a stronger statement, but it is not needed to decide the safety question, and a rename-detection heuristic in a guard is a second thing to be wrong. R9b makes that assertion in the test instead, where it is exact.

https://claude.ai/code/session_017H844j1i6G7vpPyHRv3Cgh

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -372,6 +372,20 @@ than a merge one — and this repository rebases constantly, because branches go
 stale behind required checks. It happened to #4954 against #4951, and
 `check-rebase-replay.sh` fires on #4939's merged head today.
 
+**A path can drop out of the branch's diff for two reasons, and only one is
+this hazard.** The branch removed it — an edit undone, or a file created and
+deleted. Or the base branch renamed or deleted it and the branch absorbed that
+through a merge, which is the ordinary way a long-lived branch survives a
+refactor. Tree state cannot tell them apart: a file the branch created and
+deleted is absent at the merge base and at the head, exactly like an inherited
+rename. What differs is who removed it, so the guard asks that instead — a
+suspect is dropped only when it is gone at the head *and* no non-merge commit
+of the branch's own deleted it. An inherited removal is not in the range at
+all, because merging the base advances the merge base past it. The record-plane
+extraction is what forced this: it fired the guard on every open branch that
+had correctly followed the move, and the printed remedy would have charged each
+one a history rewrite for a hazard that was not there.
+
 **The remedy is to flatten**, so the path is absent from the branch's history
 rather than present as an edit and an undo. The branch's tree is identical
 either way, so flattening costs nothing and the guard offers no
--- a/scripts/check-rebase-replay.sh
+++ b/scripts/check-rebase-replay.sh
@@ -165,6 +165,40 @@ if [ -z "$suspects" ]; then
   exit 0
 fi
 
+# A net diff can be empty for two different reasons, and only one is a hazard.
+# The branch edited the path and undid it, or created and deleted it — what
+# this guard is for. Or the BASE branch renamed or deleted the path and the
+# branch absorbed that through a merge, so the path leaves the diff with
+# nobody on the branch having removed anything.
+#
+# Tree state alone cannot separate them: a file the branch created and then
+# deleted is absent at the merge base and absent at head, the same signature
+# an inherited rename leaves. What differs is WHO removed it. A branch-authored
+# removal is a non-merge commit in this range that deletes the path; an
+# inherited one is not in the range at all, because merging the base advances
+# the merge base past it.
+#
+# So a suspect is dropped only when it is gone at head AND no commit of the
+# branch's own took it away. The replay hazard cannot hide there: it needs the
+# branch to contribute the removal a rebase would keep.
+kept=""
+while IFS= read -r path; do
+  [ -n "$path" ] || continue
+  if ! git cat-file -e "$head:$path" 2>/dev/null &&
+    [ -z "$(git_c log --no-merges --diff-filter=D --format=%H "$base..$head" -- "$path")" ]; then
+    continue
+  fi
+  kept="${kept}${path}"$'\n'
+done <<EOF
+$suspects
+EOF
+suspects="$(printf '%s' "$kept" | sed '/^$/d')"
+
+if [ -z "$suspects" ]; then
+  echo "check-rebase-replay: OK — every path missing from the branch's diff was renamed or deleted by the base branch, not by this branch."
+  exit 0
+fi
+
 found=0
 while IFS= read -r path; do
   [ -n "$path" ] || continue
--- a/scripts/test-rebase-replay.sh
+++ b/scripts/test-rebase-replay.sh
@@ -208,6 +208,48 @@ git -C "$r" commit -qam "edit f"
 want "R8 an absent base ref is a failure naming the shallow checkout" \
   expect-fail "$r" origin/nonexistent "fetch-depth"
 
+# ── 9. A rename the base branch made, absorbed through a merge ───────────────
+#
+# The branch edits a file. Upstream moves it. The branch merges that, then
+# re-lands the work at the new path.
+#
+# The old path drops out of the branch's diff, but the branch did not remove
+# it. A rebase has no copy of it to revert. It replays the edit onto a tree
+# with no such path, and conflicts loudly.
+#
+# The record-plane extraction made two of these in one morning. Both branches
+# had followed the move correctly.
+r="$(new_repo moved_upstream)"
+mkdir -p "$r/old"
+printf 'alpha\n' >"$r/old/bar.txt"
+git -C "$r" add old/bar.txt
+git -C "$r" commit -qm "add old/bar.txt"
+git -C "$r" checkout -q -b b
+printf 'beta\n' >"$r/old/bar.txt"
+git -C "$r" commit -qam "B: edit old/bar.txt"
+
+git -C "$r" checkout -q main
+mkdir -p "$r/new"
+git -C "$r" mv old/bar.txt new/bar.txt
+git -C "$r" commit -qm "upstream: move bar.txt to new/"
+
+git -C "$r" checkout -q b
+git -C "$r" merge -q --no-edit main -m "merge main" >/dev/null 2>&1
+printf 'beta\n' >"$r/new/bar.txt"
+git -C "$r" commit -qam "B: re-land the edit at the new path" >/dev/null 2>&1
+want "R9 a path the base branch renamed, absorbed by a merge, is not flagged" \
+  expect-pass "$r" main "check-rebase-replay: OK"
+
+# ...and the branch's work really did survive the move, so R9 is not passing
+# because the edit was lost.
+if [ "$(cat "$r/new/bar.txt")" = "beta" ]; then
+  pass=$((pass + 1))
+  echo "ok   R9b the edit survived the rename rather than being dropped"
+else
+  fail=$((fail + 1))
+  echo "FAIL R9b new/bar.txt reads $(cat "$r/new/bar.txt"), so the edit did not survive"
+fi
+
 echo
 echo "passed ${pass}, failed ${fail}"
 [ "$fail" -eq 0 ]


macanderson:
## This and #5932 are the same fix, written twice, 3m45s apart

Flagging the duplication rather than closing either — which one lands is yours to pick. Posted by the autonomous PR sweep.

This PR opened 09:46:14; **#5932** — "stop check-rebase-replay reading an upstream move as an undo" — opened 09:49:59. Both close **#5910**, both add an **R9** to `scripts/test-rebase-replay.sh` building the same shape, both reject #5910's proposed absent-at-both-ends condition for the same R4 reason, and both land on the same discriminator: absent at both ends **and** no non-merge deletion in the branch's own history. Both quote the same before/after against #5894's range.

This is the #5224 shape `./scripts/issue-claim.sh` exists for — #5910 carries no claim comment from either session.

**Where they differ**, since it decides which is the better base:

- This PR carries `Closes #5890` as well, folding in the duplicate issue filed 41 minutes earlier, and records the correction on both issues. #5932 closes only #5910, so #5890 would stay open if that one lands.
- #5932 adds **R10** — a genuine pair on the same branch must still fail, so the skip cannot degrade into a blanket waiver — and names the cleared paths in the run output instead of printing the bare OK line. This PR has neither.

The tidy outcome is one taking the other's extras rather than both merging: two implementations of one condition and two R9s in `scripts/test-rebase-replay.sh` is what landing both would leave.

Worth settling soon — **#5894** and **#5840** are both red on this guard now and are waiting on whichever lands.

One note on this PR's own description, since it affects the close: it says `Closes #5890` is in the body only and not on the commit trailer, so a squash merge drops it. That is correct as written — if this is the one that lands, #5890 needs closing by hand.

I have not closed or pushed to either PR. The same note is on #5932.


---

macanderson:
## A duplicate of this landed four minutes later — closed, with one test worth taking

I opened #5932 at 09:50 against the same defect, without seeing this one (09:46:14). Same discriminator, independently reached; **#5932 is now closed** in favour of this PR, which is the earlier and the better of the two — it closes #5890 as well, and **R9b** (asserting the branch's edit survived at the new path, so R9 cannot pass by the work having been lost) is a real gap in mine.

Filed **#5934** for the root cause: `scripts/issue-claim.sh` cannot run in the agent container because `gh` is not installed there, so the check CLAUDE.md instructs a session to run before writing code was simply unavailable to both of us. It fails open, correctly — but the signal was absent rather than consulted, which is #5224's failure recurring through a gap the script itself cannot see.

### The one thing worth porting: a mixed-branch case

Your suite has R4 (a branch carrying **only** a created-and-deleted pair) and R9 (a branch carrying **only** an inherited move). It has no case where **both shapes coexist on one branch** — which is what pins the skip as per-path rather than something that can short-circuit a run. Offered as a suggestion, not a defect: your `continue` inside the loop is structurally safe against this, whereas my `found -eq 0` early exit made it load-bearing, so you may reasonably judge it redundant.

Appending to `scripts/test-rebase-replay.sh` right after your R9 block (it reuses R9's `$r`):

```sh
# ── 10. The skip is per path, not a waiver ───────────────────────────────────
#
# The R9 branch plus a real pair (R4's shape: added and dropped inside the
# branch, so absent at both ends too). A blanket "absent at both ends is fine"
# was the first fix proposed. Measured against R4 it fails: this branch would
# read green and the real hazard would ship.
printf 'new\n' >"$r/g.txt"
git -C "$r" add g.txt
git -C "$r" commit -qm "add g"
git -C "$r" rm -q g.txt
git -C "$r" commit -qm "drop g again"
want "R10 a real pair beside a moved path is still flagged" \
  expect-fail "$r" main "g.txt"
```

Verified against my copy of the fix: 13 passed with it, and it is the case that goes green if the skip is ever widened to absence alone. Take it or leave it — this PR is correct without it, and I am not going to open anything competing.

---

PR #5941, opened 2026-09-05
fix(workspace): move the lock entry for stella-learn to 0.9.330

macanderson:
`main` is red because `Cargo.lock` records `stella-learn` at `0.9.329` while every manifest is at `0.9.330`: the release sync (#5915) was cut before the crate landed (#5899), so its lock entry stayed behind. `cargo metadata --locked` refused the tree and every open PR inherited the red.

One line: the lock entry moves to `0.9.330`. `./scripts/check-lockfile-sync.sh` reports OK on this branch (it compiles nothing).

No witness test: this is a lockfile repair, and `lockfile-sync` is the gate that judges it.

Tests deleted: None

Closes #5939

## Summary by Sourcery

Bug Fixes:
- Update the Cargo.lock entry for stella-learn to version 0.9.330 so locked dependency metadata matches the workspace manifests.

PR #5906, opened 2026-09-05
fix(stella-records): move the published-records test to the crate that owns it

macanderson:
## What and why

`main` does not compile, and every open PR is red behind it.

`crates/stella-core/tests/published_records_do_not_collide.rs` still imports
`stella_core::ingest::record::EnforcementMode` and `stella_core::records`. Both
modules moved to `stella-records` when the record plane was extracted, so the
test's crate no longer has them:

```
error[E0433]: cannot find `ingest` in `stella_core`
error[E0432]: unresolved import `stella_core::records`
```

This is the shared-cell shape the canary exists for: the extraction was correct
about the code it moved, and nothing in either tree was wrong on its own.

## The fix

The test moves with the code it tests, into `crates/stella-records/tests/`.
Nothing inside it changes except the two import paths:

- It reads `.stella/rules/` through `CARGO_MANIFEST_DIR/../..`, which resolves
  to the repository root from `crates/stella-records` exactly as it did from
  `crates/stella-core`.
- The `toml` parse it does is already a dev-dependency of `stella-records`.

The alternative — giving `stella-records` a dev-dependency on `stella-core` —
would point the edge the wrong way and pull the record plane back toward the
crate it just left.

## Verification

The failing build is the witness: `cargo check --workspace --all-targets` on
`main` fails on those two imports and passes here. Nothing else in the tree
names `stella_core::records` or `stella_core::ingest`.

Tests deleted: none. `published_records_do_not_collide.rs` is renamed, and
every test function in it survives the move — git reports it as a 97% rename.

Closes #5901

## Summary by Sourcery

Move the published-record collision test into `stella-records` and point it at the relocated record APIs.

Bug Fixes:
- Move the published-record collision integration test to `stella-records` and update its imports so workspace all-target checks compile again.

Tests:
- Keep the published-record collision coverage with the crate that owns the tested record functionality.

PR #5942, opened 2026-09-05
fix(build): bump Cargo.lock's stella-learn entry to 0.9.330

### LINKED ISSUE #5939: main is red: Cargo.lock still lists stella-learn at the old version
## What is wrong

`main` does not resolve against its own manifests. `cargo metadata --locked`
fails on the tip:

```
$ ./scripts/check-lockfile-sync.sh
check-lockfile-sync: FAIL — Cargo.lock is out of sync with the manifests.
     error: cannot update the lock file ... because --locked was passed to prevent this
```

The one difference is a single line. `Cargo.lock` records `stella-learn` at
`0.9.329`; `crates/stella-learn/Cargo.toml` is at `0.9.330`, like every other
member.

Confirmed still live on `origin/main` (`fc82f234c` as of this triage pass):
`Cargo.lock`'s `stella-learn` entry is `0.9.329`, `[workspace.package].version`
is `0.9.330`.

## How it got there

Two merges that were each correct wrote the same shared cell.

- #5899 (`87cc0d84e`) added the `stella-learn` crate at `0.9.329`, the
  version main carried at the time, and wrote its lock entry.
- #5915 (`ab787bed4`) is the release sync to `0.9.330`. It moved every
  member's version and every lock entry it knew about — but it was cut before
  `stella-learn` existed, so that one entry stayed behind.

Neither branch was wrong on its own, which is why no pre-merge check saw it.
It is the shared-cell composition break AGENTS.md names for `Cargo.lock`.

## Why it matters

`install.sh` builds with `--locked`, and so does the MSRV job. Every one of
those fails on this tip until the entry is bumped.

## The fix

Bump the one entry:

```
 [[package]]
 name = "stella-learn"
-version = "0.9.329"
+version = "0.9.330"
```

## How to verify

```bash
./scripts/check-lockfile-sync.sh   # FAIL before, OK after
```

## Definition of done

- [x] `Cargo.lock`'s `stella-learn` entry reads `0.9.330`, matching
      `crates/stella-learn/Cargo.toml` (`version.workspace = true` →
      `0.9.330`).
- [x] Witness: `./scripts/check-lockfile-sync.sh` fails on the unfixed tree
      and passes after the fix — the fail→pass flip this repo requires as
      proof (AGENTS.md, "witness tests"), and the same check is green in CI
      on the PR.
- [x] No other member is out of step: `cargo metadata --locked` resolves
      clean against every manifest, not just this one.
- [x] Not applicable, checked: no code changes beyond `Cargo.lock` — this is
      a generated-file fix, so there is no new module, no crate-placement
      decision, no god-file impact, no prose/design-system surface, and no
      docs update (it fixes a build artifact, not a documented behavior,
      command, or capability).
- [x] PR description names this as a shared-cell composition break (per
      AGENTS.md's `Cargo.lock` gotcha) so the pattern stays visible to
      reviewers, and links this issue with `Closes #5939` in both the PR
      description and a commit trailer (AGENTS.md, "Closing the issue on
      merge").

## Duplicate check

Not a duplicate. Prior lockfile shared-cell incidents (#3336, #3332, #3594,
#3572) are all closed — same recurring failure class, different occurrence.

## Parent

Filed under #5313 (Epic: the CI and test signal is trustworthy) as the
closest existing bucket for a CI-gate-breaking composition failure; no
dedicated shared-cell/main-red epic exists yet.

macanderson:
## What

Bumps `Cargo.lock`'s `stella-learn` entry from `0.9.329` to `0.9.330`, matching every other workspace member.

## Why — main-red repair

This is a shared-cell composition break, the exact shape AGENTS.md's `Cargo.lock` gotcha describes: #5899 (`87cc0d84e`) added the `stella-learn` crate at `0.9.329`, the version `main` carried at the time; #5915 (`ab787bed4`), the `0.9.330` release sync, moved every member's version and lock entry it knew about — but it was cut before `stella-learn` existed, so that one entry stayed behind. Neither branch was wrong on its own, which is why no pre-merge check caught it. `install.sh` and the MSRV job both build with `--locked`, so every one of those has failed against `main`'s tip since.

Filed by the post-merge canary as #5939 (`main-red` label). Checked `./scripts/main-red-claim.sh check` before starting (unclaimed) and claimed it before pushing.

## Witness

```
./scripts/check-lockfile-sync.sh   # FAIL before this commit, OK after
```
Confirmed locally: `check-lockfile-sync: FAIL — Cargo.lock is out of sync with the manifests` on the unfixed tree, `check-lockfile-sync: OK — Cargo.lock resolves against the manifests` after the one-line bump. No other member is out of step — `cargo metadata --locked` resolves clean.

No code changes beyond `Cargo.lock`, so no other DoD items (god-file, prose, docs) apply — this is a generated-file fix.

Closes #5939

## Summary by Sourcery

Bug Fixes:
- Synchronize Cargo.lock's stella-learn version with the workspace's 0.9.330 release to restore locked builds and lockfile validation.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3383,7 +3383,7 @@ version = "0.9.330"
 
 [[package]]
 name = "stella-learn"
-version = "0.9.329"
+version = "0.9.330"
 dependencies = [
  "proptest",
  "serde",


macanderson:
Standing this one down as a duplicate of #5940.

Three PRs were opened against #5939 within fifteen minutes — #5940 (10:07), #5941 (closed), and this one (10:21). All three carry the identical one-line `Cargo.lock` bump of `stella-learn` to `0.9.330`, so any of them fixes the break; the question is only which one lands.

#5940 wins on both tests AGENTS.md gives:

- **It claimed first.** `scripts/main-red-claim.sh`'s rule is that the earlier claim holds and the later session stands down. #5940 is fourteen minutes ahead.
- **It is further through CI.** Every check on `e6a1158` has passed except `fmt + clippy + test`, which is still running — including `cargo check on the declared MSRV`, the `--locked` build that exercises the repaired lock directly. This branch's run started from scratch at 10:21 and would re-spend the same twenty minutes on a byte-identical diff while `main` stays red.

This is the exact collision #4680 describes and `scripts/main-red-claim.sh` exists to prevent — three sessions each reaching the same correct conclusion at the same moment. Nothing here was wrong; the work is simply already done next door.

Closing without merging. #5939 stays open and will close via #5940's `Closes` trailer. No follow-up issue is owed: the duplicate-claim hazard is already documented in AGENTS.md and watched by the `duplicate claims` check.

---

macanderson:
This is byte-identical to #5940, which was opened 14 minutes earlier and carries the same `unblocks-main` label.

Checked rather than assumed — the two diffs share their blob hashes, so the before and after trees are the same object:

```
#5940  Cargo.lock  index 42916cde7..1ecfd01f8   stella-learn 0.9.329 → 0.9.330
#5942  Cargo.lock  index 42916cde7..1ecfd01f8   stella-learn 0.9.329 → 0.9.330
```

**#5940 is further along.** MSRV, the macOS dependency table, the Windows platform split, the container smoke, `main is not known-broken`, Sourcery and every guard are green on it; only `fmt + clippy + test` is still running (started 10:17). Whichever of the two merges second becomes an empty diff and trips `PR carries a diff` anyway.

This is the third repair filed for #5939 (#5941 is already closed), and the second four-way collision on a red `main` today — #5906/#5907/#5908 was the same shape at 06:25. Both are what `scripts/main-red-claim.sh` exists to prevent, and both got past it for the same reason: the script shells out to `gh`, which is absent in at least some agent containers, so it fails open with `proceed (could not ask)` rather than reading the tracker. Failing open is the right call — a claim check that can block a repair is worse than the duplication — but it means the check is silently inert wherever `gh` is missing, which is worth knowing given it has now been bypassed twice in four hours. #5874 is already open against that script's other blind spot.

Suggest closing this as a duplicate and letting #5940 carry the repair. Leaving the call to this PR's author rather than closing it from outside.

Not opening a fourth repair for the same one-line change, for the reason above.

---

PR #5946, opened 2026-09-05
feat(stella-cli): record a driver session's plugin, refusals, and ending

### LINKED ISSUE #5109: A driver session leaves no record: what it asked for and how it ended are printed and then lost
## Problem

`stella plugin drive <name>` (#3783) opens a driver session, serves whatever the grant permits, and prints the outcome to stdout. Nothing durable records that it happened.

So the three facts a person needs after the fact are unrecoverable once the terminal scrolls:

- **that a session ran at all**, and against which installed plugin;
- **what it asked for and was refused** — `DriverCallGate::refusals()` is read once, printed, and dropped when the `BoundDriver` does. Today every capability answers `unsupported` (`NoDriverCapabilities`), so the refusal list is the *only* record of what a driver tried to do;
- **how it ended** — `DriveNext::Sleep { secs }` or `DriveNext::Halt { reason }`, or a `DriverError`.

A session id is already minted for exactly this join (`plugin_cmd::session_id`, echoed to the driver in `DriveSession::session` so its own records can be joined to the host's) and then nothing on the host side writes it down.

## Where

- `crates/stella-cli/src/plugin_cmd.rs` — `drive`, which prints and returns.
- `crates/stella-cli/src/driver_plugin.rs` — `BoundDriver::open` and `BoundDriver::refusals`, the two values worth recording.
- `crates/stella-cli/src/self_driving_cmd/audit.rs` — `audit::record(durable, Audit::…, subject, message)` is the loop's existing "everything it did, said out loud and written down" ledger, and the nearest shape. Note it is keyed on `self_driving_cmd::state::LoopState`, which a `stella plugin` verb does not open.
- `crates/stella-store/src/` — the other candidate home, if a driver session should be an execution-adjacent row rather than a loop ledger record.

## The decision this needs first

**Which ledger.** A driver session is not a turn and not a fleet run, and it is started from `stella plugin`, not from `stella self-driving`. Three options, and none is obviously right:

1. The self-driving audit ledger (`~/.stella/self-driving/<slug>/`), which would mean `stella plugin drive` opens a `LoopState` — a coupling that may not be wanted.
2. A `store.db` row, which makes it visible to `stella observe` but needs a table that is not an `executions` child.
3. An `AgentEvent`, which is the durable emit stream — but AGENTS.md invariant #10 means a new option must declare its consumer in `crates/stella-protocol/src/event/consumers.rs` or it is an `E0004`, and a driver session has no consumer today.

Settle that before writing code; the recording itself is small either way.

## Verify

A test that opens a session against a fixture driver and then reads the record back, asserting it names the plugin, the session id, and the ending. `crates/stella-cli/src/driver_plugin/tests.rs` is where the CLI-side construction is already driven, and `crates/stella-runtime/tests/driver_socket.rs` has the fixture that can actually complete a session.

## Done when

A driver session leaves a durable record naming the plugin, the session id, every refused ask, and how it ended — with a witness test that reads it back, and the choice of ledger written down where the next reader finds it.

Refs #3783, #3599.

## Definition of done

- [x] A driver session leaves a durable record naming the plugin, the session id, every refused ask, and how it ended.
- [x] A witness test reads it back.
- [x] The choice of ledger is written down where the next reader finds it.

macanderson:
## What & why

`stella plugin drive <name>` opened one driver session and printed the
outcome to stdout. Nothing durable named the plugin, the session id, every
refused ask, or how the session ended — once the terminal scrolled, those
three facts were gone (#5109).

Adds `driver_plugin::session_log`, which appends one JSON line per session
to `.stella/private/driver-sessions.jsonl` through
`stella_store::append_workspace_private_line` — the same crash-safe,
owner-only append primitive `memory::self_tuning`'s ledger already uses. A
write failure is reported and never fails the session, the rule
`self_driving_cmd::audit::record` states for its own ledger.

**The ledger choice** (posted as a design comment on the issue before this
landed, and restated in the module's own docs): a workspace-private JSONL
file, not the three candidates the issue named. The self-driving audit
ledger needs a `LoopState`, which `stella plugin drive` — a bare CLI verb,
not a self-driving loop — has no reason to open. A `store.db` row costs a
schema migration for three fields nothing reads yet. An `AgentEvent` case
needs a declared consumer (invariant #10,
`crates/stella-protocol/src/event/consumers.rs`) and a driver session has
none today.

Closes #5109

## The witness

passes on the new**.

`a_driver_session_leaves_a_durable_record_of_what_it_asked_and_how_it_ended`
(`crates/stella-cli/src/plugin_cmd/tests.rs`) installs a real driver plugin
— a `/bin/sh` fixture that speaks the driver wire protocol directly, since
`wrapper-plugin-fixture`'s `CARGO_BIN_EXE_*` belongs to `stella-runtime`
and is not visible to `stella-cli`'s own tests — runs `drive`, and reads
the record back off disk. It fails by construction on the old code:
`driver_plugin::session_log` did not exist, so there was nothing to read
back and the test would not compile.

`crates/stella-cli/src/driver_plugin/session_log.rs` also carries unit
tests for the record's round trip, the outcome mapping from every ending a
driver can send, and the read side's tolerance of a bad line.

Exemplar followed: `crates/stella-cli/src/memory/self_tuning.rs`'s
append-only ledger under `.stella/private/` — same primitive, same
non-fatal-write rule, same read-side-exists-because-a-write-with-no-reader-
is-pointless argument.

## The gate

- [x] `cargo fmt --check`
- [ ] Workspace clippy at `-D warnings` — CI's job
- [ ] Full workspace test suite — CI's job
- [x] Docs updated where behavior/flags changed (doc comments on
      `driver_plugin`, `plugin_cmd::drive`, and the new `session_log`
      module explain the record and the ledger choice)
- [x] `Closes #5109` appears both above and as a commit trailer

`make guards-fast` is green locally (fmt, prose, file-size, and the rest of
the toolchain-free gate). Per SCR-001, only crate-scoped, filtered test
runs were done locally, targeted at the new `driver_plugin::session_log`
unit tests and at the new witness test — both pass. The workspace-wide
clippy and test suite are CI's job, not this laptop's.

## Nothing left behind

(Note: `main` is currently red on `Cargo.lock` drift, tracked and already
under repair in #5939 by another session under this account — unrelated to
this change, and this branch does not touch `Cargo.lock`.)

## Ground-rule check

- [x] No I/O added to `stella-core`; no new deps without justification
      below — no new dependency; `stella_store::append_workspace_private_line`
      is an existing, already-used-elsewhere primitive
- [x] New cross-boundary types round-trip through serde (test included) —
      `DriverSessionRecord`/`DriverSessionOutcome` round-trip is covered by
      `a_record_round_trips_through_json` and `a_recorded_session_reads_back`

The issue offered three ledger candidates and asked that the choice be
settled before code. I posted the design decision as an issue comment
first (https://github.com/macanderson/stella/issues/5109#issuecomment-5551075850),
then implemented against it — a fourth option, not one of the three named.

## Summary by Sourcery

Record durable driver-session history without changing session execution semantics.

New Features:
- Persist each driver session as a workspace-private JSONL record containing its plugin, session identity, executable, refused requests, and ending outcome.
- Provide tolerant session-record serialization and readback support for inspecting stored driver sessions.

Enhancements:
- Keep session execution successful when recording fails while reporting persistence errors to the operator.

Documentation:
- Document the driver-session ledger and its workspace-private storage choice.

Tests:
- Add unit and end-to-end witness tests covering record round trips, session outcome mapping, malformed-line tolerance, and durable records from a real driver session.

## DoD verification by the PR sweep (appended 2026-09-05)

The `dod / dod` check went red at 10:49:30Z reading #5109's three unchecked
items. All three were ticked at 10:59:02Z, after that read — so the gate
failed on a stale snapshot, not on a gap. This edit is what re-fires
`dod-check.yml`; the edit is the trigger, never the fix.

Checked against this branch's head (`e2f56395`) rather than against the
description:

- **"A driver session leaves a durable record naming the plugin, the session
  id, every refused ask, and how it ended."** `DriverSessionRecord` in
  `crates/stella-cli/src/driver_plugin/session_log.rs` carries exactly those
  four, one field each: `plugin` (the manifest name), `session_id` (the id
  `plugin_cmd::session_id` minted and sent to the driver, so the two logs
  join on it), `refusals` (every ask the gate could not serve, in order) and
  `outcome`. Appended one JSON line per session to
  `.stella/private/driver-sessions.jsonl`.
- **"A witness test reads it back."** `a_recorded_session_reads_back` reads
  the appended line back into a value, and
  `plugin_cmd::tests::a_driver_session_leaves_a_durable_record_of_what_it_asked_and_how_it_ended`
  does it end to end through the real verb.
- **"The choice of ledger is written down where the next reader finds it."**
  That module's header carries a `# Which ledger, and why not the other two`
  section, rejecting the self-driving audit ledger (needs a `LoopState` a
  bare CLI verb has no reason to open), a `store.db` row (a schema change for
  three fields nothing reads yet) and an `AgentEvent` case (invariant #10
  wants a declared reader, and a driver session has none).

Ran `cargo test -p stella-cli --bin stella -- session_log a_driver_session_leaves`
on this head — 5 passed, 0 failed, all four `session_log` tests plus the
end-to-end one.

---

### DIFF
--- a/crates/stella-cli/src/driver_plugin.rs
+++ b/crates/stella-cli/src/driver_plugin.rs
@@ -37,6 +37,12 @@
 //! [`DriveNext::Sleep`] is #3599's B2 `LoopStep` machine, and inventing a
 //! sleeper here would be inventing the loop this channel exists to be driven
 //! by.
+//!
+//! # Where a session goes once it ends
+//!
+//! `plugin_cmd::drive` reads [`BoundDriver::refusals`] and the session's own
+//! result and hands both to [`session_log::record`], so a session survives
+//! its own terminal.
 
 use std::path::Path;
 use std::sync::Arc;
@@ -48,6 +54,8 @@ use stella_runtime::wrapper::{
 };
 
 use crate::plugin_cmd::roster::PluginRoster;
+
+pub(crate) mod session_log;
 
 /// An installed driver, bound to the process the host will start.
 ///
--- a/crates/stella-cli/src/driver_plugin/session_log.rs
+++ b/crates/stella-cli/src/driver_plugin/session_log.rs
@@ -29,7 +29,7 @@
 //! If the record fails to write, that failure is reported. The session
 //! still counts as done. An operator has already seen the outcome on
 //! screen, so a disk error should not take it away. See
-//! [`crate::self_driving_cmd::audit`]'s own docs for the same rule.
+//! `self_driving_cmd::audit`'s own docs for the same rule.
 
 use std::path::Path;
 
@@ -86,7 +86,7 @@ impl DriverSessionOutcome {
 /// One driver session, as a fact on disk.
 ///
 /// Serde-first, so a test checks a value instead of parsing a sentence —
-/// the same rule [`crate::self_driving_cmd::audit::AuditEntry`] follows.
+/// the same rule `self_driving_cmd::audit::AuditEntry` follows.
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
 pub(crate) struct DriverSessionRecord {
     /// When the session was opened, RFC3339 UTC.
--- a/crates/stella-cli/src/driver_plugin/session_log.rs
+++ b/crates/stella-cli/src/driver_plugin/session_log.rs
@@ -1,0 +1,250 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+// Copyright (c) 2026 Oxagen, Inc. Commercial licensing: licensing@oxagen.sh
+
+//! Where a driver session's record goes.
+//!
+//! `stella plugin drive` opens one session and serves whatever the grant
+//! allows. Before this module, it printed the outcome and threw it away.
+//! The session id, every refused ask, and how the session ended were all
+//! gone once the terminal scrolled.
+//!
+//! # Which ledger, and why not the other two
+//!
+//! The issue named three options. The self-driving audit ledger needs a
+//! `LoopState`. `stella plugin drive` is a bare CLI verb, not a
+//! self-driving loop, so it has no reason to open one. A `store.db` row
+//! would need a schema change for three fields nothing reads yet. An
+//! `AgentEvent` case needs a declared reader, and a driver session has
+//! none today.
+//!
+//! So this module is the fourth answer. It appends one line of JSON per
+//! session to `.stella/private/driver-sessions.jsonl`, through
+//! [`stella_store::append_workspace_private_line`] — the same primitive
+//! [`crate::memory::self_tuning`]'s own ledger uses. The file is
+//! workspace-private and stays out of git, like every other file under
+//! `.stella/private/`.
+//!
+//! # A write failure never fails the session
+//!
+//! If the record fails to write, that failure is reported. The session
+//! still counts as done. An operator has already seen the outcome on
+//! screen, so a disk error should not take it away. See
+//! [`crate::self_driving_cmd::audit`]'s own docs for the same rule.
+
+use std::path::Path;
+
+use serde::{Deserialize, Serialize};
+use stella_plugin::DriveNext;
+
+/// The file every session's record is appended to, under
+/// `.stella/private/`.
+const LOG_NAME: &str = "driver-sessions.jsonl";
+
+/// How a driver session ended, as the record holds it.
+///
+/// Mirrors [`DriveNext`], plus one case it cannot hold: a session whose
+/// transport, process, or timeout failed before a `next` came back. A
+/// separate type, because a plugin-facing wire type should not also carry
+/// a host-only failure.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(tag = "kind", rename_all = "snake_case")]
+pub(crate) enum DriverSessionOutcome {
+    /// The driver asked to be woken again after this many seconds.
+    Sleep {
+        /// The clamped duration the driver asked for.
+        secs: u32,
+    },
+    /// The driver stopped, and said why.
+    Halt {
+        /// The sentence a human reads in the record.
+        reason: String,
+    },
+    /// The session never ended: a transport failure, a spawn failure, or
+    /// a timeout.
+    Error {
+        /// [`BoundDriver::open`](super::BoundDriver::open)'s own message,
+        /// which already names the program.
+        message: String,
+    },
+}
+
+impl DriverSessionOutcome {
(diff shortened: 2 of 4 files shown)


macanderson:
`dod / dod` was red on head `e2f5639` ([run 33961296616](https://github.com/macanderson/stella/actions/runs/33961296616)) because all three of #5109's Definition-of-done boxes were unticked. I verified each against this diff and then ticked them on the issue; a re-run of `dod-check` should go green on the next push or re-run.

What each box was checked against, so the tick is auditable rather than taken on trust:

**1. "A driver session leaves a durable record naming the plugin, the session id, every refused ask, and how it ended."**
`DriverSessionRecord` in `crates/stella-cli/src/driver_plugin/session_log.rs` carries `plugin`, `session_id`, `refusals: Vec<String>` and `outcome`, and `record()` appends it via `stella_store::append_workspace_private_line`. `DriverSessionOutcome` covers all three endings — `Sleep`, `Halt`, and the host-only `Error` case `DriveNext` cannot express.

**2. "A witness test reads it back."**
`plugin_cmd::tests::a_driver_session_leaves_a_durable_record_of_what_it_asked_and_how_it_ended` asserts the log is empty first, drives a session, then reads it back through `session_log::read_sessions` and asserts on `plugin`, `session_id`, `program`, `refusals` and the ending. `session_log::tests::a_recorded_session_reads_back` covers the same round trip at the module level.

**3. "The choice of ledger is written down where the next reader finds it."**
`session_log.rs`'s module doc has a `# Which ledger, and why not the other two` section answering the issue's own three options by name — the audit ledger needs a `LoopState` a bare CLI verb has no reason to open, a `store.db` row needs a schema change for three fields nothing reads yet, and an `AgentEvent` case needs a declared consumer (invariant #10) that a driver session does not have.

Run locally on this branch's head, per-crate rather than a workspace run:

```
cargo test -p stella-cli --bin stella session_log
  test driver_plugin::session_log::tests::every_ending_converts_to_an_outcome ... ok
  test driver_plugin::session_log::tests::a_record_round_trips_through_json ... ok
  test driver_plugin::session_log::tests::a_recorded_session_reads_back ... ok
  test driver_plugin::session_log::tests::a_bad_line_is_skipped_not_fatal ... ok
  test result: ok. 4 passed; 0 failed

cargo test -p stella-cli --bin stella a_driver_session_leaves_a_durable_record
  test plugin_cmd::tests::a_driver_session_leaves_a_durable_record_of_what_it_asked_and_how_it_ended ... ok
```

That is evidence the boxes are satisfied, not that this PR is green — `ci` on `e2f5639` is still queued and remains the authoritative gate.

---

PR #5842, opened 2026-09-05
feat(stella-tui): bang interrupts the turn, `$` runs the shell

macanderson:
## What and why

The composer's `!` mark ran a shell command. Per the steering grammar decided in
#5331, the bang family owns interrupt semantics from here, so:

| Input | Meaning |
|---|---|
| `$ cmd` / `$cmd` | shell command, now — skips the queue and the running turn |
| `!cmd` (bang against the text) | shell command, now — the old spelling, one more release, with a chrome note naming `$` |
| `! text` (bang, space) | interrupt: soft-stop the running turn, run `text` next |
| `! text` with nothing running | plain dispatch plus a one-line chrome note |

### The collision in the pointer, and how it is resolved

The pointer and #5331's spec comment ask for two things that cannot both hold
for one input: `! <text>` mid-turn interrupts, *and* `! ` (bang + space)
survives as a shell alias. The same DoD also asks for `!` at idle to dispatch
with a note, which rules out bang + space meaning shell there either.

The space is the discriminator, which is the shape the rest of the family takes
(`!! <text>`, `!!! <text>` in #5443). The legacy spelling that survives is
therefore the one hands already type — nobody writes `! ls` for a shell command,
and everybody writes `!ls` — so muscle memory is kept exactly where it exists.
Posted on the issue before the work started
(https://github.com/macanderson/stella/issues/5442#issuecomment-5549608094).

## One stop path, invariant 6 intact

No new driver code. `!` sends the same `WorkspaceInput::Interrupt` that #5430's
red composer mode sends, so both doors reach
`command_deck::steer::interrupt_lead`: front-insert ahead of the parked backlog,
then `request_soft_stop()`. That is a request the turn honours at its next step
boundary, never a mid-tool abort, and the turn's completed steps stay in the
transcript. The auto-resume is the driver's ordinary post-stop dispatch of the
queue head, which is now the user's text.

## Witness

`crates/stella-tui/src/deck_ui/tests/sigil.rs` drives a scripted deck, no model:

- `a_bang_at_a_running_lead_interrupts_the_turn_with_the_text` — a running lead
  plus `! use short sentences` yields
  `Interrupt { agent: "lead", texts: ["use short sentences"] }`. **It fails on
  `main` by construction**: that deck read any leading `!` as a shell mark and
  answered `DeckAction::Shell("use short sentences")`, running the sentence as a
  command.
- `a_bang_at_an_idle_lead_dispatches_the_text_and_says_so` — `Enqueue` plus the
  chrome note.
[...]

Issue #5442, opened 2026-08-28
Steer-interrupt: bang sigil soft-stops the turn and auto-resumes with the instruction


macanderson:
## Problem

From the steering grammar in the #5331 comments: a user mid-turn wants to say `! stop writing complex prose` and have the running agent **interrupted and auto-resumed** with that instruction — not queued behind the current turn, not requiring the Esc flow. Today the composer's `!` prefix runs the text as a shell command, and the only interrupt path is Esc (soft-stop) with no auto-resume-with-text.

This is the runtime half of the bang-sigil grammar (persistence levels `!!`/`!!!` are a sibling issue). It builds on #5430's red Interrupt composer mode: interrupt mode + `!` sigil are two doors into the same soft-stop-then-front-insert mechanism.

**Breaking change (decided in the #5331 spec):** the `!` shell escape moves to `$`; a legacy shell alias is kept for one release with a deprecation note. The bang family (`!`/`!!`/`!!!`) owns interrupt semantics from here forward.

## Design

- `! <text>` submitted while a turn runs: soft-stop the turn at the next safe boundary (per Invariant 6 — never mid-tool), front-insert `<text>` as a user-role message, auto-resume. The stopped turn keeps its completed steps.
- `! <text>` while idle: treated as a plain dispatch (no-op sigil) with a one-line chrome note, so muscle memory never eats a message.
- The witness must show ordering: turn A soft-stops, its completed steps persist, `<text>` runs next, then any queued input follows.

**Which spelling survives as the shell alias — amended during #5842.** The
original wording asked for `! ` (bang + space) to stay a shell alias, which the
first and third bullets above rule out: bang + space is the interrupt at a
running agent and a dispatch at an idle one, so no agent state is left for it to
run a shell command in. The alias shipped as `!cmd` — a bang with the command
against it, which is the spelling hands actually type. Reasoning:
https://github.com/macanderson/stella/issues/5442#issuecomment-5549608094

## Definition of done

- [x] `! <text>` mid-turn soft-stops the focused lane's turn and front-inserts the text; witness shows completed steps kept and the text running next — `a_bang_at_a_running_lead_interrupts_the_turn_with_the_text` (`crates/stella-tui/src/deck_ui/tests/sigil.rs`) plus `an_interrupt_front_queues_the_words_and_asks_the_turn_to_stop` (`crates/stella-cli/src/command_deck/tests.rs`), which pins the front-insert ahead of the parked backlog
[...]

PR #5951, opened 2026-09-05
feat(stella-tui): `!!` keeps guidance, `!!!` keeps a rule

### LINKED ISSUE #5443: Bang persistence sigils: !! saves guidance, !!! saves a rule, to context records
## Problem

From the steering grammar in the #5331 comments: interrupting the current session is only half the need. The other half is **persistence** — "remember this for all future sessions" — at two enforcement levels:

- `!! <text>` — interrupt (as in the sibling bang-interrupt issue) **and** save `<text>` to the workspace's context records as **guidance** (a `should`-strength record: steers behavior, overridable by explicit instruction).
- `!!! <text>` — interrupt **and** save `<text>` as a **rule** (a `must`-strength record: hard and fast from here forward until otherwise told; enforcement level 100).

Example from the comments: `!!! do not write about your honesty in prose` stops the running agent, creates a rule-strength context record, and every future session in this workspace loads it through the normal record plane.

This composes with two existing slices: the record must reach **in-flight** sessions too, which is exactly #5432's hot-reload swap (a record added mid-session appears at the next step boundary); and the record plane's governance/ledger machinery (#5327, #5328) is what the new record is written through — no side-channel writes.

## Design

- Sigil parse happens at the composer, layered on the bang-interrupt path: `!!`/`!!!` strip the sigil, run the identical soft-stop/front-insert, then write the record.
- Record write goes through the existing ingest path (`ingest::record` → ledger → registry), with `origin` set so the sweep gate treats it as user-decreed (the origin-plus-decree reading of `sweep::honored_probe`, not the dead ingest gate's).
- Strength mapping: `!!` → guidance channel (`should`); `!!!` → rule channel (`must`, enforcement 100). The record body is the sigil-stripped text verbatim.
- Write failure (corrupt ledger, governance refusal) must not lose the interrupt: the soft-stop still happens and the failure is reported visibly.

## Definition of done

- [x] `!! <text>` mid-turn interrupts and persists `<text>` as a guidance-strength context record; witness: after the turn, a fresh session's record registry contains it at `should` strength
- [x] `!!! <text>` mid-turn interrupts and persists `<text>` as a rule-strength record (enforcement 100); witness: fresh session loads it at `must` strength
- [x] The record write goes through the existing ingest/ledger path — no side-channel record creation (witness: ledger tail shows the promotion event)
- [x] A record written via `!!`/`!!!` reaches other in-flight sessions via #5432's hot reload (integration witness may be deferred to #5432 if it lands first; the dependency is stated either way)
- [x] Record-write failure still interrupts and reports the failure visibly; the text is not silently dropped
- [x] `origin` on the new record is set so the live sweep gate honors it (pinned by a gate witness)

---

macanderson:
## What and why

`!` interrupts the turn (#5442). This is the family's other half: `!!` and `!!!`
interrupt **and keep the words**, so a correction made once steers every later
session in the workspace.

| Input | Meaning |
|---|---|
| `! text` | stop the turn at its next step, run `text` next |
| `!! text` | the same stop, and keep `text` as **guidance** (`force = should`) |
| `!!! text` | the same stop, and keep `text` as a **rule** (`force = must`, `precedence = 100`) |
| `!ls` | still the old shell spelling, one more release |

A space after the last bang is the discriminator, the rule #5842 set. So
`!!important` still reaches the shell as `!important`, and the parse table test
that pinned that is unchanged. A fourth bang is refused as a mark rather than
read as the strongest one: guessing which of two meanings a typo intended is how
a stray keystroke publishes a rule.

## The write is the existing path, not a second one

`crates/stella-cli/src/context_records/decree.rs` is the sibling of
`inferred_rule_record`, the function that already publishes a mined rule. It
uses the same publication directory, the same `create_new` refusal to clobber,
the same supersession when the claim changes, and the same two ledgers. The
difference is what a decree may claim that a guess may not: `origin = user`,
`truth.basis = decree` with the actor as `verified_by`, and evidence graded
`HumanReview`. Those are exactly the fields `sweep::honored_probe` reads before
it trusts a record.

The `lineage_id` is derived from the sentence, never from the strength. So
saying the same thing twice is a no-op, and raising `!!` to `!!!` supersedes the
published revision instead of publishing a rival record that argues with it.

## Two readings the issue left open, and how they were settled

Posted on the issue before the work started
(https://github.com/macanderson/stella/issues/5443#issuecomment-5551134548).

- **"Enforcement 100."** The record surface has no numeric enforcement level.
  `steering.precedence` is its one numeric strength axis, and it decides both
  which record wins a conflict and which one keeps its place when the channel
  budget runs short. `!!!` publishes at `precedence = 100`, the ceiling.
  `enforcement` stays absent at both strengths: a `hard` mode with no evaluable
  guard advertises blocking that structurally cannot fire (the defect
  `stella_learn::rules` names), and a typed sentence supplies no guard.
- **"The ledger tail shows the promotion event."** There are two ledgers. Every
  save appends the same `DecisionEvent` `stella context keep` appends, and a
  save that supersedes a published revision appends the `Superseded` event to
  `.stella/rules/promotions.jsonl`. Both are witnessed. A first publication
  mints no *grant* event, because nothing was granted enforcement — writing one
  would put a false enforcement transition in a governance ledger.

## One save, in front of the arms

`crates/stella-cli/src/command_deck/keep_record.rs` holds one
`intercept(input, root, in_tx)` that both of the driver's receive sites pass
every message through. It writes the record, clears the mark, and **always
returns the message**, so a write that fails cannot cost the person the stop
they asked for; the failure comes back as a chrome note. An interrupt reaches
three arms (at rest, at the lead, at a worker lane) and a save inside each would
be three chances to write the record twice or not at all. `command_deck.rs` is a
god file: it gains no lines, only two changed call words.

Exemplar followed: `command_deck::steer` and `command_deck::settle`, the two
existing splits out of that god file.

## Witness mechanism

Deck (`crates/stella-tui/src/deck_ui/tests/sigil.rs`) — these **fail on `main` by
construction**: that parser owned only the first bang, so `!! use short
sentences` answered `DeckAction::Shell("! use short sentences")` and ran the
sentence as a command.

- `two_bangs_interrupt_and_ask_for_the_words_to_be_kept_as_guidance`
- `three_bangs_interrupt_and_ask_for_the_words_to_be_kept_as_a_rule`
- `a_keep_sigil_at_an_idle_lead_still_carries_the_save` — the save cannot depend
  on a turn being in flight
- parse-table units for the bang count and for the refused fourth bang

Driver (`crates/stella-cli/src/command_deck/keep_record/tests.rs`):

- `a_marked_interrupt_writes_the_record_and_hands_back_the_plain_stop`
- `a_write_that_fails_still_hands_back_the_stop_and_says_so` — `.stella/rules`
  is made a plain file, so the failure is real I/O rather than a seam invented
  for the test

Record (`crates/stella-cli/src/context_records/decree/tests.rs`):

- `guidance_is_published_at_should_and_a_fresh_load_finds_it` — a **fresh
  registry load**, which is the only view a later session has
- `a_rule_is_published_at_must_and_at_the_precedence_ceiling`
- `raising_guidance_to_a_rule_supersedes_it_and_the_ledger_says_so` — the
  promotion ledger tail
- `every_save_is_on_the_decision_ledger_with_the_file_it_wrote`
- `the_minted_record_clears_the_sweep_gate` — the origin is not untrusted, and a
  gated probe on the minted record is honored at `Trust::User`
- `what_the_mark_writes_verifies_when_it_is_read_back`, the pasted-text refusal,
  the empty-statement refusal, and the lineage-follows-the-sentence property

## Dependency

A record saved this way reaches **other live sessions** only once #5432's
mid-session hot reload lands. Those sessions read the record set once, at start.
This session and every session started afterwards pick it up on the next load.

## Golden frames

One line moved: the help sheet's `>text / !text` row now reads `steer at the
boundary · ! stops · !! !!! keep the words`, and
`tests/snapshots/deck/overlay_help_skills.txt` was updated by hand to match,
same rendered width, no other cell touched. The row also cites the two new
witnesses, which `every_binding_names_a_witness_that_exists` checks.

## Known-red base

`main` is red on a stale `Cargo.lock` (#5939, claimed by another session). This
branch touches no manifest; the lockfile check will stay red here until that
repair lands.

Tests deleted: None

Residue issues filed: listed in a comment below once opened.

Closes #5443

## Summary by Sourcery

Add persistent `!!` guidance and `!!!` rule interrupts while preserving existing stop and shell-command behavior.

New Features:
- Support `!!` interruptions that save the text as workspace guidance and `!!!` interruptions that save it as a workspace rule.
- Persist kept statements through the existing context-record publication and governance ledgers, including supersession when guidance is promoted to a rule.
- Document the new keep sigils in chat help and user-facing command documentation.

Bug Fixes:
- Preserve the original interrupt when saving fails so a failed record write cannot swallow the requested turn stop.
- Retain legacy `!cmd` shell parsing while rejecting four-bang inputs as keep markers.

Enhancements:
- Route marked interrupts through a single driver-side interception path so records are written once across idle, lead, and worker execution states.
- Add record validation, provenance, trust-gate compatibility, duplicate detection, and sentence-based lineage handling for composer-created records.

Documentation:
- Update chat command documentation and TUI help text to explain `!!` guidance and `!!!` rule persistence.

Tests:
- Add parser, deck, driver, and record-plane coverage for keep strengths, idle interrupts, failed writes, persistence, supersession, ledger entries, validation, and legacy bang behavior.

## DoD verification by the PR sweep (appended 2026-09-05)

The `dod / dod` check was red at 11:15:24Z on all six of #5443's items. Unlike
the sibling PRs in this sweep these were genuinely unticked, so each was
checked against this head (`0fff24a1`) before being ticked; the per-item
evidence is in
[this comment](https://github.com/macanderson/stella/issues/5443#issuecomment-5551453676),
including the one interpretation worth disagreeing with (item 2's "enforcement
100" → `precedence = 100`, `enforcement = None`).

Ran on this head: `cargo test -p stella-cli --bin stella -- context_records::decree keep_record`
— 13 passed, 0 failed.

Sourcery's single ❌ is answered separately above.

This edit is what re-fires `dod-check.yml` to re-read the boxes; the edit is
the trigger, never the fix.

---

### DIFF
--- a/crates/stella-cli/src/command_deck.rs
+++ b/crates/stella-cli/src/command_deck.rs
@@ -136,6 +136,9 @@ use settings_io::{apply_pending_reload, handle_engine_config_input, handle_tools
 /// Where an Esc-delivered steer lands, driver-side.
 mod steer;
 
+/// Where a `!!`/`!!!` interrupt's saved record is written.
+mod keep_record;
+
 /// ISSUES-tab requests, served by the workspace's issue provider.
 mod issues;
 
@@ -1067,7 +1070,7 @@ pub async fn run_deck_session(
                     }
                     IdleWake::Input(input) => input,
                 };
-                match input {
+                match keep_record::intercept(input, &cfg.workspace_root, &in_tx) {
                     None => break 'session,
                     Some(WorkspaceInput::Quit) => break 'session,
                     // Worker controls work between lead turns too — the
@@ -1153,7 +1156,7 @@ pub async fn run_deck_session(
                     // the same "run this now" path.
                     Some(
                         WorkspaceInput::Steer { agent, texts }
-                        | WorkspaceInput::Interrupt { agent, texts },
+                        | WorkspaceInput::Interrupt { agent, texts, .. },
                     ) if !texts.is_empty() => {
                         match steer::steer_idle(&agent, &subs, &mut queue, texts, &in_tx) {
                             Some(first) => {
@@ -1934,7 +1937,7 @@ pub async fn run_deck_session(
                             },
                         );
                     }
-                    input = sub_rx.recv() => match input {
+                    input = sub_rx.recv() => match keep_record::intercept(input, &cfg.workspace_root, &in_tx) {
                         None | Some(WorkspaceInput::Quit) => break TurnEnd::Quit,
                         // The lead is busy — the prompt does NOT wait for it.
                         // It backlogs and immediately drains to a dedicated
@@ -2028,10 +2031,10 @@ pub async fn run_deck_session(
                         // The composer's red mode: stop at the boundary, run the
                         // words next. At a worker lane it lands as a steer — see
                         // the envelope's `Interrupt` doc.
-                        Some(WorkspaceInput::Interrupt { agent, texts }) if agent == LEAD =>
+                        Some(WorkspaceInput::Interrupt { agent, texts, .. }) if agent == LEAD =>
                             steer::interrupt_lead(&steering, &lead_pause, &mut queue, texts, &in_tx),
                         Some(WorkspaceInput::Steer { agent, texts }
-                            | WorkspaceInput::Interrupt { agent, texts }) =>
+                            | WorkspaceInput::Interrupt { agent, texts, .. }) =>
                             steer::steer_worker(&subs, &mut queue, &agent, texts, &in_tx),
                         // An explicit front-insert stays a front-insert even
                         // if a turn started before it arrived — the deck's
--- a/crates/stella-cli/src/command_deck/keep_record.rs
+++ b/crates/stella-cli/src/command_deck/keep_record.rs
@@ -7,7 +7,7 @@
 //! deck cuts the mark off, sends the usual [`WorkspaceInput::Interrupt`], and
 //! tags it with the strength it was asked for. The tag is spent here.
 //!
-//! [`intercept`] runs in front of both read points in the driver. It does not
+//! `intercept` runs in front of both read points in the driver. It does not
 //! sit in the arms that take an interrupt:
 //!
 //! - **A bad write must not eat the stop.** A save that fails hands the
--- a/crates/stella-cli/src/command_deck/keep_record.rs
+++ b/crates/stella-cli/src/command_deck/keep_record.rs
@@ -1,0 +1,107 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+// Copyright (c) 2026 Oxagen, Inc. Commercial licensing: licensing@oxagen.sh
+
+//! Where a `!!` or `!!!` mark is saved.
+//!
+//! The deck cannot write the file. It is a fold, and a write is I/O. So the
+//! deck cuts the mark off, sends the usual [`WorkspaceInput::Interrupt`], and
+//! tags it with the strength it was asked for. The tag is spent here.
+//!
+//! [`intercept`] runs in front of both read points in the driver. It does not
+//! sit in the arms that take an interrupt:
+//!
+//! - **A bad write must not eat the stop.** A save that fails hands the
+//!   message back as it came. The turn still stops. The error goes on screen.
+//! - **One save, not three.** An interrupt lands in three arms: at rest, at
+//!   the lead, and at a lane. A save in each is three ways to write the file
+//!   twice, or not at all.
+//! - **`command_deck.rs` is a god file, closed to growth.** Each call site
+//!   there spends one word. The work is here.
+//!
+//! What the save reaches, and what it does not: this session picks the record
+//! up on its next load, and so does every session started after it. A session
+//! already running elsewhere does not. Those read the record set once, at
+//! start, and nothing swaps it under them yet.
+//!
+//! Split out of [`super`], on the `steer` and `settle` pattern.
+
+use std::path::Path;
+
+use tokio::sync::mpsc::UnboundedSender;
+
+use crate::context_records::decree;
+use stella_tui::{Inbound, KeepStrength, WorkspaceInput};
+
+/// Spend the mark on `input`. Hand the message back for the driver to route.
+///
+/// The mark is cleared on the way through. So the file is written once, even
+/// if a later arm reads the message again. All else passes as it came.
+pub(super) fn intercept(
+    input: Option<WorkspaceInput>,
+    root: &Path,
+    in_tx: &UnboundedSender<Inbound>,
+) -> Option<WorkspaceInput> {
+    match input {
+        Some(WorkspaceInput::Interrupt {
+            agent,
+            texts,
+            keep: Some(keep),
+        }) => {
+            save(root, keep, &texts, in_tx);
+            Some(WorkspaceInput::Interrupt {
+                agent,
+                texts,
+                keep: None,
+            })
+        }
+        other => other,
+    }
+}
+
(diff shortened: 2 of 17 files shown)


macanderson:
## Answering Sourcery's one ❌

> **Persist `!!` text as a guidance-strength (`should`) context record and `!!!` text as a rule-strength (`must`) record at enforcement level 100, with user-decreed origin and normal ledger-backed record loading.** — *it deliberately leaves the record's `enforcement` field absent and uses `precedence = 100` instead … Additionally, the implementation uses the existing context-keep helpers directly rather than clearly routing through the issue's specified `ingest::record` path.*

Two claims. The first is a real divergence from the issue's wording and is answered; the second is a misreading of the diff and is corrected.

### "enforcement level 100" has no counterpart in the record model

The model has no numeric enforcement level, so the issue's phrase cannot be satisfied literally by anything. The two fields it could mean are different in kind:

- `Record::enforcement: Option<Enforcement>` is documented as *"How a violation is detected and what happens"* — `mode`, `guard_tool`, `guard_deny_command`, `guard_allow_command`. No number anywhere. `records/registry.rs` only ever builds one from a guard: `enforcement: rule.guard.as_ref().map(|guard| Enforcement { … })`.
- `Steering::precedence: Option<u32>` is the record's only number, and the registry's own rule path sets it (`precedence: Some(50)`).

So `precedence = 100` is the faithful reading of "level 100", and it lands above the registry's own `50`. `enforcement` stays `None` because a typed sentence carries no guard, and `EnforcementMode::Hard` with no `guard_tool` would claim a block that can never fire. That reasoning is not left to be re-derived — it is the doc comment on the constant itself (`decree.rs`, `RULE_PRECEDENCE`):

> `enforcement` is left off at both strengths. A `hard` mode with no guard to check claims a block that can never fire. A typed sentence brings no guard.

### The record *is* built through `ingest::record`

This half is wrong about the diff. `crates/stella-cli/src/context_records/decree.rs` line 24 is:

```rust
use stella_records::ingest::record as rec;
```

and `fn record(...) -> Result<rec::Record, String>` builds the whole thing through that module — `rec::Record`, `rec::RecordKind::Rule`, `rec::SharingScope::Repository`, `rec::Provenance`, `rec::Steering { force, precedence, .. }`, `rec::Truth { basis: rec::TruthBasis::Decree, .. }`, and `enforcement: None`. `publish()` then resolves its path with `publication_path(root, rec::SharingScope::Repository, …)`.

The `context_records` helpers it also calls (`write_record`, `replace_record`, `append_promotion`, `append_decision`, `read_governance`, `separation_checked_proposer`) are the publication and ledger side — writing the file, stamping the hash-chained ledger, and running the separation-of-duties check. They are what *publishing* a record requires; they are not an alternative to `ingest::record`, and nothing here bypasses it.

No code change for either half: the first is the model's shape and is documented at the constant, the second did not happen.

---

macanderson:
Remaining work filed as handoffs:

- #5973 — `crates/stella-tui/src/envelope.rs` sits under thirty lines from the 1500-line ceiling after this change. The next deck protocol message has no room to land, so the file wants splitting into `envelope.rs` beside `envelope/` before somebody meets the ceiling inside a change about something else.
- #6002 — `!!`/`!!!` always publish at the repository tier, so a personal preference typed mid-turn is committed to Git and lands in a teammate's prompt. The record plane already has the user tier; nothing lets a person choose it.

The in-flight-session half stays with #5432, as the description says.

Issue #3740, opened 2026-08-18
stella-transcript: grid Cell::width() counts characters instead of terminal display cells, breaking column alignment for CJK/emoji


macanderson:
## Problem

`crates/stella-transcript/src/grid.rs` lines 163-167 define `Cell::width()` as `self.text.chars().count()`. A character's terminal display width is not the same as its character count: CJK characters and most emoji take two terminal columns, not one. Counting characters instead of display cells under-counts these, which breaks the fixed-column padding the whole module is built around (the module's own doc comment at grid.rs lines 8-15 states the gutter widths are constants and every row is padded to them so no column can drift).

The same undercount bug appears in several places that all depend on `width()`'s assumption or repeat the `.chars().count()` pattern directly:

- `line_width` (around grid.rs line 175) — depends on `Cell::width()`.
- The fill/padding computations around grid.rs lines 271 and 378 — these call `chips.chars().count()` directly, same bug.
- The fill computation around grid.rs line 528.
- The word-wrap helper around grid.rs line 632.

All of these need to be fixed together, since they all contribute to the same fixed-width column guarantee the module promises.

## Fix guidance

- Add the `unicode-width` crate as a dependency of `crates/stella-transcript/Cargo.toml`. It is already declared as a workspace dependency at the root `Cargo.toml` line 116, and `stella-tui` already depends on it for exactly this reason (terminal display width vs. character count), so this is a precedent to point to, not a novel dependency choice.
- Replace `self.text.chars().count()` in `Cell::width()` with `UnicodeWidthStr::width(self.text.as_str())` (or the appropriate `unicode-width` API for a `&str`).
- Fix the other `.chars().count()` call sites listed above (grid.rs lines 271, 378, and the word-wrap helper near line 632) to use the same display-width calculation instead of character count.
- Add a test that renders a run containing a CJK or emoji character in a tool path (or similar rendered text) and asserts the grid's gutters/columns still line up (fixed width preserved).

## Constraints

- `crates/stella-transcript/Cargo.toml` has a comment stating a deliberately narrow dependency policy for this crate — per AGENTS.md's "No new dependencies casually" rule, adding `unicode-width` needs a sentence in the PR description justifying it (point to the `stella-tui` precedent and the fixed-column guarantee this module promises).
- `stella-transcript` must stay pure (AGENTS.md rule #2 — no I/O in decision logic); `unicode-width` is a pure computation crate so this does not conflict, but keep the change scoped to width calculation only.

## Done when

[...]

PR #5952, opened 2026-09-05
test(stella-transcript,stella-cli): golden-frame snapshots for the plain transcript renderer

### LINKED ISSUE #5081: The transcript renderer that ships to every stella run user has no golden frame
## Problem

`stella_transcript::grid::render` and `render_turn_lines` draw the transcript a `stella run` user actually reads — `crates/stella-cli/src/plain/transcript.rs` is their one shipping caller, and it is the transcript for every run that never opens the deck. Nothing compares their output against a committed frame.

```
$ ls crates/stella-transcript/tests/
fixtures/word-highlight-matrix.txt
word_highlight_matrix.rs
```

One fixture, for word highlighting. Everything else in the crate is a plain unit test in `src/tests.rs`, which is the right idiom for the crate's purity — but a unit test asserts a property of a row, and the thing at risk here is the **frame**: which rows appear, in what order, at what columns, with the chips where the constants put them.

The deck has 35 goldens under `crates/stella-tui/tests/snapshots/deck/`. The renderer that ships to every non-deck user has none.

## Why now

**#4289 piece 3 replaces `grid`'s whole frame.** It swaps the `╭─ … ─╮` / `│` / `╰──╯` box for SPEC 6's labelled rules and metal rail, and the issue says so: *"its goldens (reference snapshots the renderer is checked against)"* — a file set that does not exist. That change will move every line of the plain surface's output, and today there is nothing that would show a reviewer what moved. The same is true of #3578, which routes the deck through the same function.

A golden landed **before** those makes them readable as a diff. Landed after, it records whatever they happened to produce.

## What to build

A snapshot test over `grid::render` and `grid::render_turn_lines` covering, at minimum:

- a turn with a `bash` call, a `read_file` call and a file mutation, at each of `Zoom::Turns` / `Steps` / `Everything`;
- a failed step (pinned open by `Status::pins_open`);
- a note with detail rows and one without (the fold control appears only for the first);
- a file change with an **unmeasured** extent — the `deleted` header with no count, so #4289 piece 1's rule is pinned by a frame and not only by a unit assertion;
- a CJK/emoji subject, so the `cells` measure that `grid.rs`'s module doc calls load-bearing is held by the frame it protects (#3740);
- an append-only sequence: `render_turn_lines` called at growing event counts, asserting each result is a prefix of the next — `render_turn_is_append_only_as_a_turn_grows` states this today as a unit test and it is exactly the property a golden makes legible.

Assert `grid::to_plain`, not the ANSI encodings — the module's own reasoning is that *"a golden test asserts the plain grid, so a snapshot diff is readable by a human, while the ANSI encoders are checked separately"*. Follow `crates/stella-tui/tests/deck_render_snapshots.rs` for the harness shape and `BLESS=1` regeneration, since that is the neighbourhood idiom and a second mechanism would be a second thing to learn.

## Also uncovered

`crates/stella-cli/src/plain/transcript/tests.rs` tests the printer's streaming arithmetic but never the bytes. A golden over `TranscriptPrinter`'s scrollback for a fixed event sequence is the second half of this and belongs in the same PR — it is the artifact the user sees, and it is one crate away from the renderer that produces it.

## Files

- `crates/stella-transcript/src/grid.rs` — `render`, `render_turn_lines`, `to_plain`
- `crates/stella-transcript/tests/` — where the new harness and `snapshots/` go
- `crates/stella-tui/tests/deck_render_snapshots.rs` — the harness to copy
- `crates/stella-cli/src/plain/transcript.rs` and its `tests.rs` — the second half

## Verify

Stash the golden's expected file, run the suite, confirm it fails; restore, confirm it passes. Then change one gutter constant in `grid.rs` and confirm the diff names the moved column rather than the test merely going red.

## Done when

`grid::render`'s frame is pinned by committed snapshots regenerated with a documented `BLESS=1` path, the plain surface's scrollback is pinned the same way, and #4289 piece 3 has something to show its diff against.

Found while landing #4289 piece 1 (#5076). Refs #4289, #3578, #3740.

## Definition of done

- [x] `grid::render`'s frame is pinned by committed snapshots, regenerated with a documented `BLESS=1` path.
- [x] The plain surface's scrollback is pinned the same way.
- [x] #4289 piece 3 has something to show its diff against.

Ticked against #5952 at `9919d37`, each item run rather than read — the
evidence is recorded in a comment below, including the `Verify` section's
gutter-constant probe and the one gap it exposed (#5968).

macanderson:
## What & why

`stella_transcript::grid::render` and `render_turn_lines` draw the transcript every `stella run` user reads on the plain surface. Before this PR nothing compared their output against a saved frame — only a plain unit test per property, plus one unrelated golden for word-highlight spans (`word-highlight-matrix.txt`). A unit test proves a property holds; it cannot show a reviewer which row moved, or that a gutter constant walked a column. #4289 piece 3 is about to rewrite this renderer's whole frame (SPEC 6's labelled rules in place of the `╭─ … ─╮` box), and that change needs something to diff against.

Adds two golden suites, both with their fixtures committed:

1. `crates/stella-transcript/tests/grid_snapshots.rs` — pins `grid::render` and `grid::render_turn_lines` at `grid::to_plain`, following `crates/stella-tui/tests/deck_render_snapshots.rs`'s harness shape (the pattern this workspace already uses for a golden suite over a rendered frame). Eight goldens under `crates/stella-transcript/tests/snapshots/grid/`, covering the issue's own list:
   - a turn with a `bash` call, a `read_file` call and a file edit, at `Zoom::Turns`/`Steps`/`Everything`;
   - a failed step pinned open past `Zoom::Turns`'s coarse fold, even for a turn that is not the run's last;
   - a note with detail rows (opened) beside one without any — only the first draws a fold marker;
   - a `delete_file` call whose extent nothing measured, so the header draws no count at all rather than a fabricated `−0` (#4289 piece 1's rule, pinned by a frame);
   - a CJK/emoji turn name, prompt and answer at a narrow width, exercising `grid::cells` rather than `chars().count()` (#3740);
   - the append-only growth of a streaming turn (`render_turn_lines` at 0..=3 steps while running, then finished), made legible as a diff instead of only the existing unit-test property (`render_turn_is_append_only_as_a_turn_grows`, `src/tests/frame.rs`).
2. `crates/stella-cli/src/plain/transcript/tests.rs` — one golden over `TranscriptPrinter`'s own scrollback for a fixed event sequence: the artifact a plain `stella run` user actually reads. Every other test in that module only asserts `contains(...)`, which is blind to a row sliding a column. Fixture at `crates/stella-cli/src/plain/transcript/snapshots/scrollback.txt`.

Exemplar crate: `crates/stella-tui/tests/deck_render_snapshots.rs` — its `assert_golden`/`describe_diff`/`BLESS=1` harness shape is copied nearly verbatim, cut down to a plain-text frame (no `TestBackend`, since `grid::render`/`render_turn_lines` already return the character grid).

## The witness

- [ ] No witness needed

Every new test in this PR is a witness for issue #5081: none of these assertions existed on `main`, so each fails on `main` by construction (the module/function doesn't exist there) and passes here. The flip was run, not just described: with the golden directory moved aside, `cargo test -p stella-transcript --test grid_snapshots` reports `0 passed; 8 failed`; with it restored, `8 passed; 0 failed` — confirmed on this head, and again independently in CI's own run (see below).

## The gate

- [x] `cargo fmt --check` — green on [this run](https://github.com/macanderson/stella/actions/runs/33962101843/job/101297649599)
- [x] `cargo clippy --workspace --all-targets -- -D warnings` — same run
- [x] `cargo test --workspace` — same run, 8/8 new grid goldens + the scrollback golden passing
- [x] Docs updated where behavior/flags changed (README, `--help`, doc comments) — n/a, test-only change
- [ ] CLA signed
- [x] `Closes #5081` appears both above and as a commit trailer

## Nothing left behind

## Ground-rule check

- [x] New cross-boundary types round-trip through serde (test included) — n/a, no new types

`#4289` piece 3 and `#3578` are the follow-up work this issue exists to give something to diff against — this PR does not touch `grid.rs` itself, so it introduces no rendering change, only coverage.

The three commits on this PR are the harness landing red (no goldens yet, by construction — the missing-file panic prints the exact expected bytes), then two follow-up commits blessing the goldens from that real output and fixing a bad `BLESS=1` command spelling (`--lib` on a crate with no library target) found along the way. Both are read in the commit messages themselves.

Closes #5081

---

### DIFF
--- a/crates/stella-cli/src/plain/transcript/snapshots/scrollback.txt
+++ b/crates/stella-cli/src/plain/transcript/snapshots/scrollback.txt
@@ -1,0 +1,18 @@
+# plain scrollback golden · one turn, width 100
+# TranscriptPrinter's byte-for-byte scrollback for a fixed event
+# sequence — this crate's one shipping caller of
+# stella_transcript::grid::render_turn_lines. The grid golden that
+# pins the renderer itself lives one crate over, in
+# crates/stella-transcript/tests/grid_snapshots.rs. Styling stripped
+# via strip_ansi. Regenerate with:
+#   BLESS=1 cargo test -p stella-cli --bin stella plain::transcript::tests::the_scrollback_a_run_leaves_matches_its_golden
+
+╭─ fix-the-overfull-hbox ──────────────────────────────────────────────────────────────────────── ─╮
+│  you  fix the overfull hbox warnings
+│
+│  agent  ▸ I'll read the file first. …
+│ 0:00 ▸ read_file   app/main.tex                                                       [9ms] [1 ln]
+│      ▸ bash        latexmk main.tex                                                   [9ms] [1 ln]
+│
+│  agent  Read it.
+╰─ ✓ done ────────────────────────────────────────────────────────────────────── [2 steps] [0:00] ─╯
--- a/crates/stella-cli/src/plain/transcript/tests.rs
+++ b/crates/stella-cli/src/plain/transcript/tests.rs
@@ -344,3 +344,76 @@ fn strip_ansi(text: &str) -> String {
     }
     out
 }
+
+// --------------------------------------------------------- the golden
+
+/// The command quoted in every failure message below.
+const SCROLLBACK_BLESS_CMD: &str = "BLESS=1 cargo test -p stella-cli --lib \
+     plain::transcript::tests::the_scrollback_a_run_leaves_matches_its_golden";
+
+fn scrollback_snapshot_path() -> std::path::PathBuf {
+    std::path::PathBuf::from(env!("CARGO_MANIFEST_DIR"))
+        .join("src/plain/transcript/snapshots/scrollback.txt")
+}
+
+/// The full golden file: a self-describing header, then the exact bytes a
+/// terminal or a log file receives.
+fn scrollback_golden_body(frame: &str) -> String {
+    format!(
+        "# plain scrollback golden · one turn, width {W}\n\
+         # TranscriptPrinter's byte-for-byte scrollback for a fixed event\n\
+         # sequence — this crate's one shipping caller of\n\
+         # stella_transcript::grid::render_turn_lines. The grid golden that\n\
+         # pins the renderer itself lives one crate over, in\n\
+         # crates/stella-transcript/tests/grid_snapshots.rs. Styling stripped\n\
+         # via strip_ansi. Regenerate with:\n\
+         #   {SCROLLBACK_BLESS_CMD}\n\
+         \n\
+         {frame}"
+    )
+}
+
+/// This pins the artifact the user sees. `grid_snapshots.rs`, in
+/// `stella-transcript`, pins the renderer. This test pins what
+/// [`TranscriptPrinter`] streams to scrollback, byte for byte, for one
+/// fixed run. No other test here checks that against a saved frame.
+///
+/// Every other test in this module checks `streamed`'s output with
+/// `contains(...)`. A substring check cannot see a row slide one column, or
+/// a gutter constant drift. A golden test can.
+#[test]
+fn the_scrollback_a_run_leaves_matches_its_golden() {
+    let live = strip_ansi(&streamed(&turn_events()));
+    let body = scrollback_golden_body(&live);
+    let path = scrollback_snapshot_path();
+
+    if std::env::var_os("BLESS").is_some() {
+        let dir = path.parent().expect("snapshot path has a parent");
+        std::fs::create_dir_all(dir)
+            .unwrap_or_else(|err| panic!("create {}: {err}", dir.display()));
+        std::fs::write(&path, &body)
+            .unwrap_or_else(|err| panic!("write {}: {err}", path.display()));
+        return;
+    }
+
+    let Ok(expected) = std::fs::read_to_string(&path) else {
+        panic!(
+            "no golden at {}.\n\
+             Create it with:  {SCROLLBACK_BLESS_CMD}\n\
+             \n\
+             This run produced:\n{body}",
+            path.display()
+        );
+    };
+    // Normalize line endings: a Windows checkout with `core.autocrlf` on
+    // would otherwise fail this golden on the first character of every line.
+    let expected = expected.replace("\r\n", "\n");
+
+    assert_eq!(
+        expected,
+        body,
+        "the plain surface's scrollback drifted from its golden at {}.\n\
+         If the change is intended, re-bless and read the diff:  {SCROLLBACK_BLESS_CMD}",
+        path.display()
+    );
+}
--- a/crates/stella-cli/src/plain/transcript/tests.rs
+++ b/crates/stella-cli/src/plain/transcript/tests.rs
@@ -348,7 +348,7 @@ fn strip_ansi(text: &str) -> String {
 // --------------------------------------------------------- the golden
 
 /// The command quoted in every failure message below.
-const SCROLLBACK_BLESS_CMD: &str = "BLESS=1 cargo test -p stella-cli --lib \
+const SCROLLBACK_BLESS_CMD: &str = "BLESS=1 cargo test -p stella-cli --bin stella \
      plain::transcript::tests::the_scrollback_a_run_leaves_matches_its_golden";
 
 fn scrollback_snapshot_path() -> std::path::PathBuf {
--- a/crates/stella-transcript/tests/grid_snapshots.rs
+++ b/crates/stella-transcript/tests/grid_snapshots.rs
@@ -1,0 +1,680 @@
+//! Golden-frame snapshots of `stella_transcript::grid`.
+//!
+//! This is the character-grid renderer. It draws every plain `stella run`
+//! transcript. `grid::render` and `grid::render_turn_lines` had one fixture
+//! before this file. That fixture pins one function's spans, not a frame.
+//! Every other test in this crate is a plain unit test in `src/tests.rs`. A
+//! unit test proves a property. It cannot show which rows a frame has, in
+//! what order, at what column. This file adds that missing check.
+//!
+//! `grid.rs`'s whole frame is due for a rewrite. SPEC 6 swaps the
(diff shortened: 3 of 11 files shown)


macanderson:
## The three ❌ rows are fixed, not answered

All three of Sourcery's rows were one defect: the harness and the eight cases landed, the goldens they assert against did not. The suite could not pass on the submitted tree. Both directories are now committed.

**`crates/stella-transcript/tests/snapshots/grid/`** — eight files (`83e705a`).
**`crates/stella-cli/src/plain/transcript/snapshots/scrollback.txt`** — one file (`9919d37`).

### The flip, run rather than described

```console
$ mv crates/stella-transcript/tests/snapshots/grid /tmp/hidden
$ cargo test -p stella-transcript --test grid_snapshots
test result: FAILED. 0 passed; 8 failed          # exit 101

$ mv /tmp/hidden crates/stella-transcript/tests/snapshots/grid
$ cargo test -p stella-transcript --test grid_snapshots
test result: ok. 8 passed; 0 failed
```

Both suites verified in **assert** mode, not only under `BLESS=1`:

- `cargo test -p stella-cli --bin stella plain::transcript::` → 12 passed, 0 failed
- `cargo test -p stella-transcript --test grid_snapshots` → 8 passed, 0 failed
- `cargo fmt --check` clean

### Read, not blessed

Per this repo's rule that a golden blessed without a look is a changelog rather than a test, each was checked against something other than its own output:

- **Frame widths recomputed independently.** Every rule measures exactly its declared width — 100 for the seven wide cases, 56 for `cjk_and_emoji_content` — re-derived with a separate East-Asian-width pass, so the renderer's cell accounting agrees with a second opinion and not only with itself. That matters here because the cell-width question is live in #5185/#5921.
- **`append_only_growth` holds the property it is named for.** Each of the five stages is a strict line-prefix of the next, checked mechanically rather than by eye, so no later stage rewrites a line an earlier one emitted.
- **`unmeasured_deleted_file`** draws `deleted` with no `+n −n` count, and **`notes_with_and_without_detail`** draws a fold marker on the note with detail rows and none on the one without — each the behaviour its description claims.

The one cut frame (`cjk_and_emoji_content`, width 56) truncates mid-token and leaves `[$0.00` dangling. That is the existing cut rule, recorded as-is rather than corrected here; `grid.rs` covers the cut rules on their own and the wide fixtures deliberately avoid cuts so they show column layout instead.

### One defect found while fixing it

`SCROLLBACK_BLESS_CMD` named `cargo test -p stella-cli --lib …`. `stella-cli` is a binary crate with no library target, so that command fails outright:

```console
$ BLESS=1 cargo test -p stella-cli --lib plain::transcript::tests::the_scrollback…
error: no library targets found in package `stella-cli`
```

The documented remedy for a failing golden could not regenerate the golden. Corrected to `--bin stella`, which is the spelling the rest of the repo already uses for this crate. The constant is quoted into the golden's own header, so the committed file carried the broken command too; re-blessing under the fix repaired both copies.

### Not covered

`stella-cli`'s snapshot has no fixture directory of its own beyond this one file, and no other test in the PR was touched. Nothing was deleted.

---

macanderson:
## `dod / dod` is red on `9919d37` against state that no longer exists

The failing job is [101295664849](https://github.com/macanderson/stella/actions/runs/33962107391/job/101295664849). Its whole output is the SCR-003 notice:

```
##[error]**SCR-003 — this PR cannot close its issue yet.**

- macanderson/stella#5081 has 3 unchecked DoD item(s):
```

**Fixed, ~90 seconds after that job read the issue.** The timestamps are the argument:

| Time (UTC) | Event |
| --- | --- |
| 11:17:55 | `dod / dod` reads #5081, finds 3 unchecked boxes, fails |
| ~11:19 | All three boxes ticked on #5081, with the evidence recorded [in a comment](https://github.com/macanderson/stella/issues/5081#issuecomment-5551419551) |

This check reads issue state rather than the commit, so nothing about `9919d37` needs to change and no push is owed. #5081's checklist is now fully ticked, each box run rather than read — including its `Verify` section's gutter-constant probe.

**I could not re-run it**: `rerun-failed-jobs` on that run returns `403 Resource not accessible by integration`. So this comment rather than a re-run, per the rule that a failure I cannot re-run gets said once instead of being left silent. It should clear on the next push to this branch, or on a manual re-run by anyone who has the button.

Not a flake, and not skipped or worked around — the underlying condition was real when the job ran and is genuinely gone now. Nothing was disabled to get here.

**One thing the DoD probe surfaced that is not fixed here:** `NUM_COL` in `grid.rs` can be changed without any golden noticing, and the arm beside it hardcodes `{:>4}  `, so a five-digit line number renders 7 columns against an unnumbered row's 6. Filed as **#5968** rather than folded in — a numbered gutter is not among the six cases #5081 enumerates, and fixing it here would widen this PR past the issue it closes.

I am watching this PR and will keep checking until it is green and mergeable.

---

macanderson:
@sourcery-ai review

macanderson:
Sourcery's review budget is exhausted for the next ~1.5 days, so a fresh automated pass isn't available on this head — answering its three ❌ rows here instead, since they were against the first commit (before the goldens landed).

All three objected to the same thing: the golden `.txt` fixtures were not yet committed, so the tests failed by construction. That gap is closed as of `9919d373c`:

- `crates/stella-transcript/tests/snapshots/grid/` now carries all eight goldens the suite in `grid_snapshots.rs` asserts against.
- `crates/stella-cli/src/plain/transcript/snapshots/scrollback.txt` is committed too, and the `BLESS=1` command that regenerates it is fixed (`--lib` → `--bin stella`, since `stella-cli` has no library target).

Verified on this head, not just described: `cargo test -p stella-transcript --test grid_snapshots` is `8 passed; 0 failed`; moving the golden directory aside first flips that to `0 passed; 8 failed`, confirming the tests genuinely assert against the committed fixtures rather than tolerating their absence. CI's own `fmt + clippy + test` job agrees: https://github.com/macanderson/stella/actions/runs/33962101843/job/101297649599

All three linked-issue objectives are addressed on this head.

Issue #3963, opened 2026-08-19
roleless-core: a plugin manifest still cannot declare a contributed stage — the wire is open, the producer is not


macanderson:
The wire and every renderer now carry an **open** stage vocabulary
(`stella_protocol::StageName`, `crates/stella-protocol/src/stage.rs`): a stage
name outside the host's twelve decodes, renders with its own word, and styles
with its own colour. The producer half is still closed, so nothing inside this
repository can emit one.

## Problem

`stella_plugin::wrapper::StageName` (`crates/stella-plugin/src/wrapper.rs`) is a
closed enum, one-to-one onto `stella_protocol::StageKind`'s twelve. A
`[[wrapper.stages]]` entry naming anything else is a **load error**. So a plugin
manifest cannot declare `triage-lite`, and the dispatch path
(`crates/stella-runtime/src/wrapper/dispatch.rs`, which iterates
`stages: Vec<StageName>`) has no contributed stage to dispatch.

This is deliberate today and the crate says why:

> Closed by design … an unknown value must be a load error rather than a
> silently shorter run

That argument is sound and is exactly why this issue is **not** "just open the
enum". A manifest key that loads but dispatches nothing is worse than one that
refuses to load.

## Why this is not blocked on all of #3907

#3907 (slice 3) lists four decisions that block it. Only some of them touch this:

- Blocking: **what a stage is formally** (a named ordering over the existing
  four wrapper points, vs a new dispatch point) and **how two plugins declaring
  a stage at the same position order deterministically**.
- Not blocking: the *name* being an opaque string, and how it renders. Both are
  settled and shipped.

So this issue is the producer seam that #3907's decisions unblock, split out so
the rendering work does not have to wait on them.

## Files

- `crates/stella-plugin/src/wrapper.rs` — `StageName`, `WrapperStage`,
  `Condition::parse`, `Signal::publisher` (the condition vocabulary is keyed on
  the stage that publishes each signal, so opening the enum touches condition
  validation, not just the name).
- `crates/stella-plugin/src/error.rs` — `ManifestError` variants for stage
  validation.
- `crates/stella-runtime/src/wrapper/dispatch.rs` — the stage loop.
- `crates/stella-plugin/src/wire.rs` — `before_turn`'s `stage` field, which is
  the plugin-facing wire spelling of the same vocabulary.

## Verify

A manifest declaring a stage the host has never heard of loads, dispatches, and
the deck shows it:

[...]

PR #5960, opened 2026-09-05
feat(stella-records): give the two unreachable provenance grades producers

macanderson:
## What and why

Two of the five `ProvenanceGrade` rungs had no construction site anywhere in
the workspace, so three `ImpactClass` floors governed nothing. `PromptHint`,
`RecallBias` and `AdvisoryRecord` all floor at `TrajectoryAbstraction`, which
nothing produced, so the only way to satisfy them was to clear the *higher*
`EnvironmentObservation` bar. `HumanReview` was worse: no publication
authority lifts a weak grade, so a reviewer's sign-off could not change any
outcome anywhere.

The issue filed this as a decision rather than a fix. AGENTS.md's evolution
rule settles it: "a prompt hint may be trialled from **a mined trajectory**,
and a blocking guard or an executable tool needs deterministic proof plus a
person." A mined trajectory is what `TrajectoryAbstraction` names, and the
three floors that sit on it are exactly the prompt hint and its two siblings.
Deleting the rung would contradict a normative document, so the rung is right
and the derivation is what was missing. That argument is in the issue as a
comment, posted before any code was written.

### `TrajectoryAbstraction` — derived in `EvidencePool::from_observations`

The pool already holds real `ObservationRecord`s, and each carries a
`task_id`. It now counts distinct tasks for itself and lifts the fold when the
evidence spans three or more.

Four properties keep this off the `#2569`/`#2570` vote-counting line, and all
four are mechanical rather than editorial:

- **A `max`, so it never weakens.** A pool of tool outcomes stays at
  `EnvironmentObservation`.
- **It caps at that rung.** However often a pattern recurs, it can never
  authorise a steering directive, a blocking guard, or an executable tool.
- **Its floor is a constant here, not a setting.** A project that lowers
  `context.promotion.inferred_directive.min_distinct_tasks` cannot lower what
  its evidence is worth.
- **It reads the observations' own `task_id`s, never a caller's
  `ProposalScore`.** Thirty restatements inside one task still fold to
  `ModelCritique`.

The reasoning: the observations each say what happened in one task, and the
pool says the same thing recurred across several. That second claim is derived
against the trajectory corpus, and re-deriving against another source is the
one move `#2782` says does promote a grade.

### `HumanReview` — derived on the keep path

`decision_grade(actor, action)` yields `HumanReview` for a user confirmation
and `None` for every other pair. `published_grade` takes the stronger of that
and the proposal's own fold. `proposals_cmd::materialize_directive` computes it
[...]

Issue #5078, opened 2026-08-26
A [seats] assignment never reaches the plugin role it names, so the seat silently runs on the session model


macanderson:
## Problem

A `[seats]` assignment can never be found at dispatch, so a user who points a plugin's role at a specific model gets the session's model instead — silently, with no notice.

The two ends disagree about the key:

- **The user's end is qualified.** `SeatsSection` in `crates/stella-cli/src/settings/toml_config.rs` documents `[seats]` keys as `<plugin-id>/<role>` and gives `"vera/verifier" = "anthropic/claude-opus-5"` as its own example, citing `doc:roleless-core` §8.4. `resolve_seat_models` in `crates/stella-cli/src/agent/seats.rs` iterates that map and inserts into `SeatProviders` under the key verbatim.
- **The dispatch end is bare.** `crates/stella-runtime/src/wrapper/child_turn.rs` builds the child's `SubAgentSpec` with `seat: Some(args.role.clone())` — the plugin's own local role name, straight off the wire. `SessionSubAgents::wiring_for` in `crates/stella-cli/src/subagent.rs` then does `self.seats.get(seat)`, which is a lookup of `verifier` in a map keyed `vera/verifier`.

The miss is not reported: `wiring_for` falls back to `model.provider.clone()` by design ("an unassigned seat is a miss, and a miss means the session's model"), which is correct behaviour for a genuinely unassigned seat and wrong here, because the seat *was* assigned.

§8.4 decided that the **host** applies the prefix, precisely so a plugin cannot forge another plugin's namespace by declaring `vera/verifier` itself. Slice 0 (#3913) shipped the display half — `declared_seats` and `installed_seats` in `crates/stella-cli/src/agent/seats.rs` both build `format!("{}/{}", plugin, role)` — but nothing applies the prefix on the routing path.

## Why it matters

- A settings line that reads like a capability and is not one. `resolve_seat_models` goes out of its way to emit a notice for every *other* way a seat can fail to resolve (unknown provider, disallowed model, missing credential, adapter failure) precisely so this cannot happen silently. This path bypasses all four, because the seat resolves fine — it is the lookup that misses.
- It costs money in the direction nobody wants: a user who assigned a cheap model to a plugin's chatty role pays for the session model on every one of that role's calls.
[...]

PR #5963, opened 2026-09-05
feat(stella-plugin): stages declare their band, and settings hold the standing plugin set

### LINKED ISSUE #3907: roleless-core slice 3: stages are declared, named and ordered by a plugin manifest
Slice 3 of #3903. Spec: `doc:roleless-core` §6. **The slice that delivers the
product change.** Four decisions below must be settled before writing code.

## Problem

The acceptance sentence for #3903 is:

> Install a triage plugin. Activate it. With no other configuration, the only
> thing that changes is that the turn now has a triage stage, and the default
> model performs it.

The wrapper socket today has four dispatch points (`before_turn`, `after_turn`,
`judge`, `again?` — `crates/stella-runtime/src/wrapper/mod.rs`) but **no notion
of a named, ordered stage a plugin adds to the turn**. A plugin can contribute
at a point; it cannot say "I add a triage stage, here, called this". So there is
nothing for the sentence above to be true *of*.

## What to do

- A manifest declares the stages it contributes: name, position relative to the
  turn, and the roles each needs.
- The turn loop composes the installed set in manifest-declared order and
  dispatches each.
- A declared role with no user assignment resolves to the default model, via the
  seat plane landed in slice 0 (`crates/stella-cli/src/agent/seats.rs`).

## Decisions that block this

1. **Is "activated" distinct from "installed"?** The acceptance sentence says
   "install … activate", implying a plugin can be present but inert. Today
   `--pipeline <id>` names one wrapper per run. A standing activation set is a
   different model with different consent, ordering, and failure semantics.
2. **What is a stage, formally?** Either (a) a named ordering over the existing
   four points — cheap, composes with what shipped — or (b) a new dispatch point
   with its own contract — more expressive, more to get wrong. Recommendation:
   (a) unless a concrete plugin needs (b).
3. **How do stages compose?** Two plugins both declaring a stage "before the
   worker" need a deterministic order. Manifest priority, install order, and
   explicit user ordering are all defensible; nondeterminism is not, given the
   byte-stable-prompt rule (`AGENTS.md` architecture rule 7).
4. **Seat namespacing.** Slice 0 made a seat name a bare string, so two plugins
   both declaring `planner` share one assignment. Feature or collision?
   Recommendation: keep bare, revisit on a real report.

## Constraints

- Architecture rule 4 of `doc:roleless-core`: installing a plugin must change what the
  turn does and never what it costs per call. A five-participant plugin with no
  assignments runs five participants on one model.
- Consent must render the stages and roles a plugin adds. Opaque to core is not
  opaque to the human.

## Witness

Installing a plugin declaring one stage changes the turn's composition,
observably, with no configuration; uninstalling restores it exactly.

## Done when

A stage's existence, name and order come from a manifest and nothing else.

## Definition of done

- [x] A stage's existence, name and order come from a manifest and nothing else.

### LINKED ISSUE #5977: Two same-band plugins take their stage order from active_plugins, and nothing pins that tiebreak
Noticed by the autonomous PR sweep while checking #3907's last DoD item against #5963's head (`82b43dc`). Filed rather than fixed there because it is a judgement about what #3907's acceptance sentence means, not a defect in that diff — see [the PR comment](https://github.com/macanderson/stella/pull/5963#issuecomment-5551574778).

## Problem

#5963 makes stage order manifest-derived across bands. Within a band it does not, and the default band is `normal`, so the default configuration is the uncovered one.

`merge_stage_order` (`crates/stella-runtime/src/wrapper/compose.rs`) ends:

```rust
// Stable, so a band separates stages the weave left in one run while the
// weave keeps every pair it did order.
order.sort_by_key(|(_, band)| band.rank());
```

The sort is stable by design — that is what lets the constraint weave keep every pair it did order. The consequence is that two plugins which share no stage **and declare no band** keep the relative order they were iterated in, which is `active_plugins` order. So for two ordinary plugins, stage order comes from a settings list rather than from either manifest.

#3907's remaining DoD item reads "A stage's existence, name and order come from a manifest and nothing else", which is why the `dod / dod` check is red on that PR.

## Files

- `crates/stella-runtime/src/wrapper/compose.rs` — `merge_stage_order`, the stable sort above
- `crates/stella-runtime/tests/wrapper_composition.rs` — `a_band_orders_two_plugins_that_share_no_stage` gives its two fixtures **different** bands, so the same-band pair is not exercised anywhere
- `crates/stella-cli/src/plugin_activation.rs` — resolves `active_plugins` into the ordered member list the weave iterates
- `docs/spec/roleless-core.md` §8.3 — disqualifies install order, and the reasoning that does or does not extend to `active_plugins`

## Which way this should go

Probably **not** a code change. `doc:roleless-core` §8.3 rejects install order for being machine-local and invisible; `active_plugins` is neither — it is an ordered list in a settings file that travels with the repository, so byte-stable prompts (AGENTS.md #7) and cross-machine benchmark reproduction both survive. On that reading two manifests declaring no relative order genuinely express no preference, and a visible, deterministic tiebreak is the right answer.

If that is the call, two things are owed:

1. #3907's item wants narrowing — "and nothing else" is the clause that makes it read as unmet — or an explicit note that the same-band tiebreak is `active_plugins` by design.
2. A test should pin it, so the tiebreak is a checked property rather than something that emerges from `sort_by_key`'s stability guarantee. Today nothing fails if someone swaps in an unstable sort, and the failure would be a silently reordered composed prompt.

The other reading — that a manifest must be able to order any two stages — needs a within-band ordering key, which #5963's own design notes argue against (two authors both claim priority 0).

## Definition of done

- [x] A test pins the same-band, no-shared-stage pair against a stated
      tiebreak.
- [x] The tiebreak is stated where the code is, not left to `sort_by_key`'s
      stability guarantee.
- [x] `doc:roleless-core` §6 records the reading.

Settled in #5963: the tiebreak is the order the selection named the members —
the `--pipeline` text or the `active_plugins` list. Both are written down and
travel with the repository, so §8.3's objection to install order does not
reach them.

The same check found that the weave answered with the **reverse** of that
order: a member sharing no stage started its walk at the front and put its
whole list ahead of everything already agreed. Fixed in the same PR, with
`one_band_and_no_shared_stage_runs_in_the_order_the_selection_named` as the
witness.

The alternative reading — that a manifest must be able to order any two stages
— stays available, and needs a within-band ordering key with its own ADR.

macanderson:
## What and why

Slice 3 of the roleless-core plan. A stage's **name** and **existence** already
came from a manifest (#3963/#3964). Its **order** did not, and nothing made an
installed plugin take part in a turn at all — the acceptance sentence this
slice exists for ("install a triage plugin, activate it, and the turn now has a
triage stage") had nothing to be true of.

Two things land.

**The band.** `[[wrapper.stages]]` gains `band = "early" | "normal" | "late"`
(default `normal`). `compose::merge_stage_order` sorts the merged order by band
with a **stable** sort, so the existing constraint weave decides the order
inside a band and the band decides it across bands. Two plugins that share no
stage now get their order from their manifests instead of from whichever the
selection happened to name first.

Two refusals keep it honest, both on `compose.rs`'s own rule that a composition
names a disagreement rather than picking a winner:

- `WrapperError::ConflictingStageBand` — two members putting one stage in two
  bands.
- `ManifestError::StageBandOutOfOrder` — a manifest writing `late` above
  `early`. `Wrapper::validate`'s signal-graph check walks declaration order, so
  a manifest allowed to write the bands backwards would have its own two stages
  swapped at run time and that check would have proved nothing about the order
  that ran. Because the two orders are held equal, the sort can never break the
  weave's answer.

`ConflictingStageOrder` survives unchanged and now answers the narrower
question: a contradiction *within* a band. That settles the collision an
earlier session stopped on, in its own reading (a).

**The standing set.** `active_plugins` is a new settings key — an ordered list
of plugin names, spelled `active_plugins` in `.stella/settings.json` and
`[run] active_plugins` in `stella.toml`. Every door that takes `--pipeline`
falls back to it (`run`, `goal`, `fleet`, `arena`), so
`PipelineChoice::is_raw()` stays honest and the enterprise process-free
authority still sees a plugin about to spawn. `--pipeline` wins for one run: it
means "run with exactly this".

It sits **inside** the project trust boundary, beside `hooks` and
`context_providers`, because a name in it makes the host spawn a program.
`plugins` keeps its sticky-`off` posture and needs no such gate — it points the
other way and can only take a plugin away.

Also: a plugin name may not hold `/`, so the `<plugin>/<role>` seat key a host
builds has one reading (`ManifestError::NameHoldsSeatSeparator`).

**Inside one band, the selection is the tiebreak** — the `--pipeline` text or
the `active_plugins` list, both written down and identical on every clone, so
§8.3's objection to install order does not reach them. Checking that turned up
a defect in the weave, fixed here: a member sharing no stage started its walk
at the front of the order and put its whole list *ahead* of everything already
agreed, answering with the reverse of what somebody wrote. It now starts at the
end in that case (#5977).

I was handed four decisions to follow. Two contradict `doc:roleless-core` §8,
which is in the tree, dated 2026-08-19, and carries the reasoning for each
call. I followed §8 and said so on #3907 before writing code. The short form:

1. **`plugins.active = [ids]` cannot be written.** `Settings::plugins` is
   `BTreeMap<String, Toggle>`, so `plugins.active = ["vera"]` is an array where
   a toggle is expected and fails to parse. Inverting that key is refused on its
   own merits too: its doc comment records that sticky-`off` is what keeps it
   safe outside the project trust boundary, and a set that *grants*
   participation points the other way. Hence a new root key.
2. **Band, not integer priority plus install order.** §8.3 disqualifies install
   order outright — machine-local and invisible, so the composed prompt would
   depend on the order somebody typed commands, which breaks byte-stable
   prompts (AGENTS.md #7) and cross-machine benchmark reproduction. It also
   rejects a fine-grained integer, because two authors both claim priority 0.

The pointer's "seat names stay bare" is what §8.4 already requires and what
this ships: the manifest declares a bare local role name and the **host**
applies the `<plugin-id>/` prefix, which is what makes the namespace
non-forgeable.

## Witness mechanism

- `a_band_orders_two_plugins_that_share_no_stage`
  (`crates/stella-runtime/tests/wrapper_composition.rs`) binds two fixtures
  that share no stage, **both ways round**, and asserts one answer. On the base
  commit it fails twice over: `band` is not a manifest key there so both
  fixtures refuse to load, and the merged order followed the selection.
- `one_band_and_no_shared_stage_runs_in_the_order_the_selection_named` — the
  same-band pair. Fails on the base commit, which answers with the reverse of
  the selection.
- `one_stage_in_two_bands_is_refused_at_bind_time` — the new refusal, named.
- `a_stage_declares_the_band_it_runs_in` and
  `a_manifest_that_walks_backwards_through_the_bands_is_refused`
  (`crates/stella-plugin/tests/wrapper_stages.rs`) — the load-time half.
  `an_unnamed_band_stays_unwritten_on_the_wire` pins the round trip, so a
  manifest that named no band still serializes without one (AGENTS.md #4).
- `a_plugin_name_may_not_carry_the_seat_separator` — the seat-key rule.
- `crates/stella-cli/src/plugin_activation.rs`'s five unit tests drive the
  resolution: order kept, empty set wraps nothing, a name that is not installed
  gets a line, a plugin with no `[wrapper]` block gets a line, a repeat runs
  once. All fail on the base commit because the module does not exist.
- `an_untrusted_project_cannot_switch_a_plugin_on`
  (`crates/stella-cli/src/settings/tests/trust.rs`) — the boundary, with the
  trusted control beside it so the gate is not proved by refusing everything.
- The settings ledger's own compile-time and merge tests carry the rest:
  `active_plugins` is destructured exhaustively in
  `settings/completeness.rs`, listed in `ROOT_FIELDS` and `RUN_FIELDS`, and set
  to a non-default value in `EVERY_KEY`, so a missing `overlay_scope` carry
  fails `every_merged_settings_field_survives_the_scope_merge`.

## Exemplar

`compose.rs`'s existing refuse-rather-than-resolve shape is the local exemplar,
and the new band refusal follows it exactly. The coarse-band-over-a-constraint
-solve arrangement is `cargo`'s resolver shape: a preference layered over a
solver rather than folded into one comparator.

## Tests deleted

None.

## Remaining work

Filed as issues rather than left in this description:

- #5974 — `stella plugin enable`/`disable`, which would write the list and pick
  the insertion point from the band.
- #5975 — consent text and `stella plugin list` rendering the stages and roles
  a package adds.
- #5976 — interactive mode, which offers no `--pipeline` and so runs no wrapper
  at all.
- #5978 — a `[seats]` key naming a plugin that is not installed, or not
  switched on, says nothing. §8.4 decided it should warn by name.
- #5979 — the `seat` field's doc comment still shows a bare role name.

Closes #3907, Closes #5977

## Summary by Sourcery

Make plugin wrapper participation and stage composition deterministic through manifest-defined bands and an ordered, trust-aware standing plugin set.

New Features:
- Add manifest stage bands to provide deterministic coarse ordering when composing plugin stages.
- Add the ordered `active_plugins` setting and use it as the default pipeline across run, goal, fleet, and arena commands.
- Enforce plugin-name and stage-band validity rules needed for unambiguous seat keys and execution order.

Bug Fixes:
- Preserve selection order for same-band plugins that share no stages.
- Fix schema documentation cleaning so the wire-schema documentation gate runs successfully.

Enhancements:
- Reject conflicting bands for shared stages and manifests that declare stages in reverse band order.
- Apply project trust-boundary rules to active plugin selection and report inactive, missing, duplicate, or non-wrapper entries.

Build:
- Correct the rustdoc cleanup command to use an unscoped documentation cache clean compatible with Cargo.

Documentation:
- Update the roleless-core specification and TOML configuration documentation for stage bands and active plugins.

Tests:
- Add coverage for stage-band parsing, ordering, validation, plugin activation resolution, trust boundaries, seat-name validation, and composed wrapper behavior.


macanderson:
## DoD check by the PR sweep — one item, and why I did not tick it

`dod / dod` is red on #3907's one remaining item:

> - [ ] A stage's existence, name and order come from a manifest and nothing else.

I checked it against this head (`82b43dc`) rather than against the description, and it is **substantially but not entirely** satisfied. Reporting rather than ticking, because the residual is real and the call on whether it counts is yours.

**What holds.** `merge_stage_order` (`crates/stella-runtime/src/wrapper/compose.rs`) takes existence and name from `wrapper.stages` alone (#3963/#3964), and order now comes from two manifest-declared sources: the constraint weave within a band, and `band.rank()` across bands. `ConflictingStageBand` and `StageBandOutOfOrder` keep both honest. `a_band_orders_two_plugins_that_share_no_stage` binds its two fixtures both ways round and gets one answer, which is the acceptance sentence working.

**What does not.** The sort is stable, by design:

```rust
// Stable, so a band separates stages the weave left in one run while the
// weave keeps every pair it did order.
order.sort_by_key(|(_, band)| band.rank());
```

Stability means two plugins that share no stage **and declare the same band** — which is what every plugin gets by default, since `band` defaults to `normal` — keep the relative order they were iterated in. That order is `active_plugins`, not a manifest. The witness test avoids this case by giving its two fixtures different bands, so nothing in the suite pins the same-band pair.

So for the default configuration — two ordinary plugins, neither declaring a band — stage order still comes from the activation list.

**Two readings, and I think one is clearly better.** Either (a) this is a gap and the item stays unticked, or (b) two manifests that declare no relative order genuinely express no preference, so any deterministic order satisfies them, and `active_plugins` order is a legitimate tiebreak. (b) is defensible in a way install order was not: `doc:roleless-core` §8.3 disqualified install order for being machine-local and invisible, and `active_plugins` is neither — it is an ordered list in a settings file that travels with the repository, so byte-stable prompts (AGENTS.md #7) and cross-machine benchmark reproduction both survive.

If (b) is your reading, the item's wording is what wants narrowing — "and nothing else" is what makes it read as unmet — and it would be worth a test pinning the same-band pair so the tiebreak is a checked property rather than an emergent one.

Nothing else on this PR is blocking: every other check is green, and `fmt + clippy + test` was still running at the time of this read.

---

macanderson:
## Answering the Sourcery ❌, and repairing the "filed" claim

From the autonomous PR sweep. Two things: the one unmet objective in Sourcery's
assessment table needs an answer before this is mergeable (CLAUDE.md, "Every
Sourcery ❌ gets a fix or an answer"), and this PR's description promises an
issue list it does not contain.

### The ❌ row — deliberately out of scope, and now genuinely filed

> Manifest-declared roles are resolved through the seat plane, including
> default-model fallback for unassigned roles, and the added stages and roles
> are shown in consent disclosures. ❌

Split in two, because the halves have different answers.

**Role resolution through the seat plane is not this slice's work — it already
landed.** `crates/stella-cli/src/agent/seats.rs` is on `main` and is slice 0's
deliverable; its own module doc states the contract this row asks for ("a core
that knows exactly one role name (`default`) and resolves every other…", and
"Core must not invent a default for a role it does not understand"). #3907's
"What to do" cites it as already-shipped infrastructure this slice builds on,
not as something to build. The one live defect in it — a `[seats]` assignment
not reaching the plugin role it names — is #5078, and **#5953 fixes it**, in
flight beside this PR right now. So this half is neither missing nor
unaccounted for; it is one crate over.

**Consent rendering is a real gap, and the row is right about it.** #3907's
constraints say so plainly ("Consent must render the stages and roles a plugin
adds. Opaque to core is not opaque to the human"). This PR makes a manifest's
stages decide a turn's composition and makes an activated plugin run every
turn, and neither `consent_text` nor `stella plugin list` says a word about it.

### The description's issue list is missing

> **Remaining work, filed.** See the issue list at the end of this description.

There is no list at the end of the description, and I could find no open issue
for any of the three items it names. So the "filed" claim was not true when it
was written. Per SCR-004 that is residue, not a footnote, so I filed them —
`triage` label only, per SCR-005:

| | |
|---|---|
| #5974 | `stella plugin enable`/`disable` — nothing writes `active_plugins`, so the standing set can only be edited by hand. Includes the band-correct insertion point and the lexicographic tie-break §6 specifies. |
| #5975 | consent text and `stella plugin list` do not render the stages and roles a package adds — **this is the Sourcery ❌**. Noted there that it wants settling with #5294's machine-readable form rather than in front of it. |
| #5976 | interactive mode reads no `active_plugins` and offers no `--pipeline`, so the Command Deck — the default interactive shell — runs no wrapper at all. Flagged as the sharpest of the three, because it makes #3903's acceptance sentence false on the surface a person is most likely to test it on. |

Worth editing the description to point at these three rather than at an absent
list.

### Separately: #3907's DoD item is satisfied, and I read the code rather than this description

`dod / dod` is red on the issue's one box, and the box is genuinely earned:

- **Existence and name** came from the manifest already (#3963/#3964, both
  closed).
- **Order** is what this PR adds: `band` is a `[[wrapper.stages]]` key
  (`StageBand`, `crates/stella-plugin/src/wrapper.rs`), and
  `compose::merge_stage_order` sorts the woven order by `band.rank()` with a
  **stable** sort, so the weave decides within a band and the manifest decides
  across bands.
- **"and nothing else"** is what `a_band_orders_two_plugins_that_share_no_stage`
  actually proves, and it is a real witness rather than a restatement: it binds
  two fixtures sharing no stage **both ways round** and asserts the same answer
  for both. A selection-driven rule cannot produce one answer for two orderings,
  so the test can distinguish the claim from its opposite.
- `Wrapper::validate` refusing `StageBandOutOfOrder` is what keeps the sort from
  ever overturning the weave, so the two orders cannot disagree.

I have **not** ticked the box, for one reason: `fmt + clippy + test` is still
running on this head, so I cannot yet say the witness test passes — only that it
exists and would fail on the base commit. That job going green is the last piece,
and then the box is tickable on evidence rather than on assertion.

---

macanderson:
## Answering Sourcery's one ❌ row

Per CLAUDE.md, a `❌` needs a fix pushed or an answer on the PR before this is mergeable. The other two rows are `✅`; this is the third:

> Manifest-declared roles are resolved through the seat plane, including default-model fallback for unassigned roles, and the added stages and roles are shown in consent disclosures. ❌

**Answering, not fixing** — it is two objectives, and neither belongs in this diff.

**Seat-plane role resolution landed separately, and is already on `main`.** It was a live defect rather than missing scope: `child_turn.rs` sent the bare role name while `[seats]` keys are `<plugin-id>/<role>`, so an assignment never reached dispatch and the seat silently rode the session model. That is #5078, fixed by #5953 (merged at 11:51Z today, closing the issue). Sourcery read this PR's diff, which correctly does not contain that fix — the two were separated deliberately, because #5078 is a bug in what slice 0 shipped rather than a piece of slice 3, and it was filed with its own witness sketch on 2026-08-26 rather than smuggled in here. Default-model fallback for a genuinely unassigned seat is `wiring_for`'s existing documented behaviour and is unchanged.

**Consent rendering is filed as #5975** — "consent text and `stella plugin list` do not render the stages and roles a package adds". It is named as out-of-scope in this PR's own description and in the design comment posted on #3907 before any code was written ("Out of this PR, filed as issues rather than smuggled in: … consent text rendering each stage and its roles, `stella plugin list` showing what a plugin adds to a turn"), and `doc:roleless-core` §6 carries it as remaining work. This slice adds the manifest field and the composition rule; nothing renders yet because the rendering surface is its own piece of work with its own consent-text review.

So: one half landed elsewhere, one half is tracked. Nothing here is deferred silently.

### The remaining blocker is `dod`, not this row

For whoever picks this up: `dod / dod` is red on #3907's single unchecked box, and the reasoning for not ticking it is in [the earlier sweep comment](https://github.com/macanderson/stella/pull/5963#issuecomment-5551574778) — the same-band default case still takes its order from `active_plugins` rather than from a manifest, and whether that counts as "and nothing else" is a maintainer call. Every other check on this head is green.

---

macanderson:
Answering the one ❌ in the assessment table.

> Manifest-declared roles are resolved through the seat plane, including
> default-model fallback for unassigned roles, and the added stages and roles
> are shown in consent disclosures.

The row bundles two halves with different answers.

**Role resolution through the seat plane already ships, and this PR does not
change it.** Slice 0 landed it: `crates/stella-cli/src/agent/seats.rs` builds
`<plugin>/<role>` for every installed package's declared roles,
`resolve_seat_models` keys `SeatProviders` by those assignment keys, and
`SessionSubAgents::wiring_for` looks one up. A miss is one lookup that returns
nothing and the seat rides the session's model, which is the default-model
fallback. There is nothing in this diff to implement, and nothing in it that
touches that path.

What this PR does add on that axis is the rule §8.4 named as its follow-on and
nothing implemented: a plugin name may no longer hold `/`
(`ManifestError::NameHoldsSeatSeparator`), so the key a host builds has one
reading. Two live defects in that neighbourhood are filed rather than fixed
here, because both are bugs in what shipped rather than pieces of this slice:
#5978 (a `[seats]` key naming a plugin that is not installed, or not switched
on, is accepted in silence) and #5979 (the `seat` field's doc comment still
shows a bare role name).

**Consent rendering is genuinely absent, and it is out of scope here on
purpose.** `doc:roleless-core` §6 states the rule — consent must render the
stages and roles a plugin adds — and it is filed as #5975 with the files, the
shape to render, and a definition of done. It is a change to
`stella_plugin::consent_text` and the `stella plugin list` verb, neither of
which this PR touches; folding it in would put a rendering change inside a diff
whose subject is composition order and activation, and would make both harder
to review and to revert.

The remaining-work list in the description names it, along with #5974
(`stella plugin enable`/`disable`) and #5976 (interactive mode, which offers no
`--pipeline` and so runs no wrapper at all).

macanderson:
The `docs/wire still matches the types` red on this PR is not this PR's, and not about the wire schema. The job dies before it reaches that check:

```
cargo clean --doc -p stella-protocol -p stella-plugin -p stella-serve
error: --doc cannot be used with -p
make: *** [Makefile:644: doc-warnings-schema] Error 101
```

`a2fe819` (#5947) added that clean in the scoped spelling, and cargo refuses `--doc` alongside `-p` in either order — so `make doc-warnings-schema` fails on its own first line. It is on `main` now, and it fails for **every** diff that reaches `wire-schema.yml` (`docs/wire/**`, `stella-protocol`, `stella-serve`, `stella-plugin`). This branch touches `stella-plugin`, which is the whole of its involvement.

It was not caught pre-merge because `wire-schema.yml` triggers on `pull_request` with those paths, and #5947's own diff (Makefile, AGENTS.md, CONTRIBUTING.md, workflows, scripts) touches none of them — the workflow that runs the recipe never ran on the PR that changed it.

**Fix is up as #5993** (I opened a duplicate, #5995, and have closed it — #5993 was first and is more accurate about what the unscoped clean costs). Ported nowhere because there is nothing to port into: the broken line is on `main`, not on this branch, and this branch cannot fix it without carrying an unrelated Makefile change. Once #5993 lands, merge `main` in again and this check goes green with no edit here.

The structural half — `wire-schema.yml` is the only runner of `doc-warnings-schema` and its `paths:` filter excludes the `Makefile` that defines it, so the recipe can break with nothing running it — is #5992.

The merge commit I pushed earlier (`66b2eda`) is unrelated to this and stands: it resolves the `error.rs` / `manifest.rs` conflict #5953 created by landing the seat-separator rule under different variant names.

---

macanderson:
`docs/wire still matches the types` is red here for a reason that is not this PR's diff — no action needed from you on it.

`make doc-warnings-schema` aborts on its own first line:

```
cargo clean --doc -p stella-protocol -p stella-plugin -p stella-serve
error: --doc cannot be used with -p
make: *** [Makefile:644: doc-warnings-schema] Error 101
```

That line came from #5947, which merged to `main` at 11:57:49Z. This PR's
wire-schema job ran at 11:58:35Z — `pull_request` CI builds the merge commit, so
it picked the broken recipe up from the base. #5960's identical job ran at
11:51Z, before that merge, and passed. Your branch head does not contain #5947
at all.

Fix is up as #5993 (one flag; cargo has no package-scoped doc clean). Once it
lands, re-run this job or push and it should clear. #5992 tracks why no workflow
caught it at the point it was introduced.

---

macanderson:
**`docs/wire still matches the types` was red, and the cause is not in this PR's diff.** Fix ported and pushed as `14c1c68`.

**What failed** ([run 33964741750](https://github.com/macanderson/stella/actions/runs/33964741750)) — `make doc-warnings-schema`, on its first line:

```
cargo clean --doc -p stella-protocol -p stella-plugin -p stella-serve
error: --doc cannot be used with -p
make: *** [Makefile:644: doc-warnings-schema] Error 101
```

**Why it is not this PR's.** The line comes from #5947 (`a2fe819`), which is on `main` and not in this branch's history — this branch's merge base predates it. It reaches this PR only because a `pull_request` run builds the merge of the head with live `main`, and `main`'s copy of the recipe is the broken one. What singles this PR out is simply that its diff touches a wire-contract path, and `wire-schema.yml` is the only workflow that runs `doc-warnings-schema` at all. Every other open PR is green on this check because none of their diffs start that workflow.

**What was pushed.** `origin/main` merged in (6 commits, no conflicts), then #5993's one-line repair applied on top: `cargo clean --doc $(SCHEMA_CRATES)` → `cargo clean --doc`. The merge is what makes the port possible — without it the branch and `main` each add a different line at the same place, and CI's merge would conflict.

This follows AGENTS.md's rule for a failure that is not this PR's: port a fix that exists rather than wait on it to merge. It no-ops the moment `main` carries #5993.

**Evidence** (cargo 1.97.0, this branch checked out):

```console
$ cargo clean --doc -p stella-protocol
error: --doc cannot be used with -p        # exit 101

$ cargo clean --doc
     Removed 0 files                        # exit 0
```

Two outcomes that differ, so the check distinguishes the repair from the break rather than being consistent with both. The end-to-end proof is this PR's own next `docs/wire still matches the types` run — it is the one job that executes the recipe, and this diff starts it.

Nothing else on this head was failing: `fmt + clippy + test`, MSRV, Windows, `docs guards`, `guard self-tests` and `dod / dod` were all green on `66b2eda`.

---

macanderson:
## `docs/wire still matches the types` is red, and the failure is not this PR's

The [job](https://github.com/macanderson/stella/actions/runs/33964741750/job/101302767255) dies before `wire-schema` ever runs:

```
error: --doc cannot be used with -p
make: *** [Makefile:644: doc-warnings-schema] Error 101
```

`#5947` shipped `cargo clean --doc $(SCHEMA_CRATES)` on `main`, and cargo refuses that pair outright. I verified it directly rather than reading it off the log — `cargo clean --doc -p stella-protocol` exits 101 with exactly that message, `cargo clean --doc` exits 0. So `doc-warnings-schema` exits on its first line on every PR whose paths reach `wire-schema.yml`.

**There is nothing to port into this branch.** Your tree predates `#5947` and has no `cargo clean --doc` line at all — the broken recipe reaches you only through the merge commit CI builds against `main`. Editing the branch's own Makefile would mean merging `main` in first and then repairing it, which buys this PR a large unrelated merge to fix a bug it does not contain.

**#5993 is the fix** (`cargo clean --doc`, unscoped) and is open and mergeable. Once it lands, re-running this job is all this PR needs. I have not re-run it yet — a re-run before the fix merges would just fail the same way.

I did port the identical one-line fix into #5999, which does carry the bad line in its own tree.

Everything else here is clean: no merge conflict, and the `dod / dod` reds are stale runs whose latest is green.

---

macanderson:
## DoD check on `Closes #3907` — the Sourcery ❌ row is right

Separate from the CI note above. Sourcery's third row says the diff implements neither role resolution/default-model fallback nor consent rendering for the stages and roles a package adds, and your own "Remaining work" section agrees — #5975 is exactly "consent text and `stella plugin list` rendering the stages and roles a package adds".

That is #3907's third objective. Filing it rather than leaving it in the description is the right handling under SCR-004, so the residue is tracked and nothing is lost. The mismatch is narrower than that: the body says `Closes #3907` while one of that issue's three stated objectives ships in #5975 instead, and SCR-003 closes an issue only when every DoD item is satisfied and verified.

**Recommendation: `Refs #3907` here, and let whichever PR lands #5975 close it.** `Closes #5977` looks correct and can stay.

Worth noting on mechanics: `Closes #3907, Closes #5977` is already written the right way — the one-keyword-per-issue form AGENTS.md calls for, so both actually close rather than only the first.

The first two objectives (manifest-defined stage existence/order, and an activated standing set with a deterministic `--pipeline` override) both read ✅ and I did not find anything contradicting that.

I have not edited the description — the close-vs-refs call is yours.

---

Issue #5200, opened 2026-08-27
Epic: one holdout, one trial ledger, one sweep — give memory, skills, and rules a shared evidence engine


macanderson:
## The problem, stated once

Three surfaces evolve on evidence — memory records, skills, and mined rules —
and each answers "does injecting this change outcomes?" with a different
mechanism. Two of the three cannot answer it even in principle:

- **Memory records** attribute from citations. Nothing writes `memory_citations`
  (#4862), and the design deliberately records a citation as *neutral* because,
  in `memory/uses/tests.rs`'s own words, "trace correlation cannot establish an
  observable effect". So even with a producer restored, this class can never
  reach `failing: true`. Restoring the retired `cite_memory` tool is not the fix:
  #2195 closed by retiring it after the model never called it once.
- **Skills** promote on frequency (`require_measured_lift` defaults `false`) and
  demote on nothing (#5086).
- **Rules** publish on reflection prose, which grades `ModelCritique` — two rungs
  below the `EnvironmentObservation` that `ImpactClass::SteeringDirective`
  requires. #4865 records that every mined rule would be refused if the gate were
  called.

The one mechanism in the tree that *can* answer the question is already built,
already tested, and has exactly one consumer:
`crates/stella-core/src/comparison.rs`, driven by
`crates/stella-core/src/skills/appraisal.rs` — a paired two-arm comparison with a
withheld baseline, four verdicts (`Helps`, `GuardBlocked`, `Harms`, `Inert`), a
confidence floor, and a `decide_demotion` that refuses to touch anything
hand-authored. That engine is unreachable from the binary because no path
produces its `WITHOUT_*` arm.

**The missing thing is not three attribution schemes. It is one holdout.**

## Proposal

One trial ledger and one holdout scheduler, feeding `comparison` for all three
surfaces.

1. **A holdout scheduler.** On a configured fraction of eligible turns, withhold
   one selected artifact (a memory record, a skill, or a mined rule) and record
   that it was withheld. The withheld turns are the control arm that no surface
   has today. Fraction is configurable, defaults conservative, and is disclosed
   in the UI — this deliberately costs some turns, and a learning system that
   hides that cost is not one this repo should ship.
2. **One trial ledger.** Generalize `SkillTrial` to an artifact-keyed trial:
   `(artifact_kind, artifact_id, task_key, withheld, outcome, turns)`, appended
   at turn end from the seam #5086 already specifies (the hook that knows both
[...]

PR #5989, opened 2026-09-05
fix(stella-records): declare the memory-citation evidence source retired

### LINKED ISSUE #4862: Nothing records a memory citation any more, so memory quarantine and promotion can never fire

macanderson:
## What

`ObservationSource::MemoryCitation` is retired. It paired a memory citation
with a model's usefulness/truthfulness judgement, but nothing in a real turn
ever produced one: the `cite_memory` tool that collected the judgement is in
`RETIRED_TOOL_NAMES`, and every other source that touches a recalled memory
only knows it was *shown* to the model, never that the model judged it
useful or true.

This is the declared-gap half from the issue's own two options, per the
maintainer's decision already posted on the thread (and epic #5200:
"citations are display-only; retirement will read the shared holdout
sweep"). No producer is added.

## Changes

- `crates/stella-records/src/context_record/lifecycle.rs`: delete the
  `ObservationSource::MemoryCitation` option and its `as_str` arm.
- `crates/stella-records/src/context_record/provenance.rs`: drop it from
  `observation_grade`'s match.
- `crates/stella-cli/src/memory/evidence.rs`: delete the
  `citation_observation` constructor (its module doc now covers only the
  surviving `ToolOutcome` source).
- `crates/stella-cli/src/memory/uses.rs`: delete the call site that mined
  citations into observations (it only ever mined test fixtures — the
  feedback loop that reads `memory_citations` for selection health is
  untouched).
- `crates/stella-store/src/lib.rs`: doc comments on `MemoryCitationRow`,
  `MemoryCitationStats`, `fold_citation_stats`,
  `PROMOTION_CITATIONS_REQUIRED`, and `QUARANTINE_NEGATIVES_THRESHOLD` now
  say plainly that no production caller writes a row today, and why they
  stay rather than being deleted.
- `crates/stella-parity/src/evolution.rs`'s `Memory` row already states this
  posture (landed in #5246, before this PR) — left as is.

## Why not wire a producer

Writing `truthful: true` because a memory's handle appeared in the answer
would assert a judgement nothing gathered, which #2782 forbids — and
`memory::uses::extract_one` already refuses exactly that fabrication for the
sibling `ContextUse::Rendered` vs `Cited` distinction. The issue's own
investigation (posted on the thread) checked reflection, the recall path,
and the answer text for an honest signal and found none.

## Witness

`memory_citation_is_retired_and_no_longer_parses`
(`crates/stella-records/src/context_record/provenance/tests.rs`) asserts the
wire payload `"memory_citation"` fails to deserialize as
`ObservationSource`.

Checked the artisanal way: swapped `lifecycle.rs`/`provenance.rs` back to
their pre-change content, added this exact test, and ran
`cargo test -p stella-records --lib memory_citation_is_retired_and_no_longer_parses`.
It failed — the pre-change enum still parses `"memory_citation"` to
`Ok(MemoryCitation)`. Restored the new files (`git status` clean again) and
reran the same command: it passes.

Also ran, on the new code:
- `cargo test -p stella-cli memory::evidence::` — 8 passed (the tests for
  the deleted citation constructor are removed with it; the rest,
  including a rewritten `re_appending_the_same_evidence_is_a_replay`
  against the surviving `tool_outcome_observation`, pass).
- `cargo test -p stella-cli memory::uses::` — 13 passed, including
  `citations_alone_never_retire_and_a_strong_verdict_retires_restorably`,
  which exercises the citation-fed selection-health feedback loop this PR
  leaves untouched.

## Tests deleted

Three tests exercising the deleted `citation_observation` constructor
(`a_negative_citation_becomes_a_citation_sourced_observation`,
`a_positive_citation_is_not_mined_because_it_would_cite_itself`,
`a_citation_with_no_remark_records_nothing`) in
`crates/stella-cli/src/memory/evidence/tests.rs`. The replay-idempotency
property they and `re_appending_the_same_evidence_is_a_replay` covered is
kept — that one test is rewritten against the surviving
`tool_outcome_observation` source instead of deleted, so no coverage is
lost.

## Residue

None filed — the two natural follow-ups (a producer for citation-based
promotion, and the shared holdout sweep) are already tracked by #5200,
which this PR does not attempt.

Closes #4862

## DoD gate re-fired by the PR sweep (2026-09-05)

`dod / dod` read #4862's five items unchecked at 12:08:57Z; all five were
ticked at 12:10:45Z, after that read, so the gate failed on a stale snapshot.
This edit is what re-fires it.

Spot-checked on this head rather than taken on trust: `ObservationSource::MemoryCitation`
is gone from `crates/`, and the names that remain are different things —
`stella_store::MemoryCitationRow`, the `memory_citations` table, and the doc
comments in `memory/evidence.rs` and `memory/uses.rs` explaining the
retirement. `stella-parity`'s `Memory` row states that nothing writes
`memory_citations` and that path-anchor validation is the only evidence source
retiring anything today.

---

### DIFF
--- a/crates/stella-cli/src/memory/evidence.rs
+++ b/crates/stella-cli/src/memory/evidence.rs
@@ -1,56 +1,42 @@
-//! Generalizing the citation loop — Phase 4 deliverable 2 (#715).
+//! Observation sources beyond reflection lessons — Phase 4 deliverable 2
+//! (#715).
 //!
-//! [`ObservationSource`] has always had three variants. Only one of them,
-//! `ReflectionLesson`, had a constructor: `MemoryCitation` and `ToolOutcome`
+//! [`ObservationSource`] started with three variants. Only one,
+//! `ReflectionLesson`, had a constructor — `MemoryCitation` and `ToolOutcome`
 //! were declared in Phase 1 and written by nothing, so the typed loop learned
 //! from reflection prose and from no other evidence at all. This module fills
-//! the other two, which is what turns explicit citation from *the* evidence
-//! source into *one* of them.
+//! `ToolOutcome`, the one source left standing beside `ReflectionLesson`.
 //!
-//! # What does not change
+//! # Why `MemoryCitation` is gone
 //!
-//! The deliverable is explicit that the citation loop keeps its semantics, and
-//! it does — untouched. `fold_citation_stats` still derives quarantine at
-//! [`stella_store::QUARANTINE_NEGATIVES_THRESHOLD`] untruthful citations and
-//! promotion eligibility at a streak past
-//! [`stella_store::PROMOTION_CITATIONS_REQUIRED`], still recomputed on every
-//! read, still never stored. Nothing here reads those thresholds, changes
-//! them, or adds a second path to quarantine.
-//!
-//! What changes is that a citation now *also* leaves an observation, so the
-//! same miner that learns from reflection prose can learn from what the model
-//! said about a memory while using it. Two evidence sources feeding one loop,
-//! rather than one source feeding the loop and another feeding a separate
-//! quarantine counter that nothing else can see.
-//!
-//! # Why only negative citations become observations
-//!
-//! A positive citation says "this memory was right", which the memory already
-//! claims — mining it produces an observation that restates its own evidence,
-//! and spec §7's second anti-poisoning rule is that a proposal may not cite
-//! itself. A *negative* citation says something the corpus does not already
-//! contain: that a stored belief is wrong. That is new information and is worth
-//! learning from.
+//! It shipped here too, pairing an observed retrieval with a model's
+//! usefulness judgement — and the judgement half had no producer. The
+//! `cite_memory` tool that collected it is retired, and every replacement
+//! source checked only knows a memory was *rendered* into the prompt, never
+//! that it was judged useful or truthful. Writing `truthful: true` for "the
+//! handle appeared somewhere" would fabricate a judgement no evidence
+//! earned. See [`ObservationSource`]'s own doc comment for the full account;
+//! the citation-derived quarantine loop it fed
+//! (`fold_citation_stats`/thresholds in `stella-store`) stays as declared,
+//! inert plumbing rather than being deleted with it.
 //!
 //! # Why only failed tool calls
 //!
-//! Same asymmetry. A tool that worked is the expected case; mining it would
-//! bury the signal under thousands of successes. A failure is a fact about this
-//! workspace that the next turn could use.
+//! A tool that worked is the expected case; mining it would bury the signal
+//! under thousands of successes. A failure is a fact about this workspace
+//! that the next turn could use.
 
 use stella_context::{AppendOutcome, ContextStore, LedgerAppend};
 use stella_learn::redact::redact_secrets;
 use stella_records::context_record::{
     ContextRecordKind, LIFECYCLE_SCHEMA_VERSION, ObservationRecord, ObservationSource,
 };
-use stella_store::MemoryCitationRow;
 
-/// The longest observation text either source will emit.
+/// The longest observation text this source will emit.
 ///
-/// A citation remark is at most 300 characters by the ledger's own contract,
-/// but a tool error is unbounded — a stack trace or a wall of compiler output would
-/// otherwise become one enormous "observation" that the miner cannot cluster
-/// and a person cannot read.
+/// A tool error is unbounded — a stack trace or a wall of compiler output
+/// would otherwise become one enormous "observation" that the miner cannot
+/// cluster and a person cannot read.
 const MAX_OBSERVATION_CHARS: usize = 300;
 
 /// Truncate on a character boundary, marking that it happened.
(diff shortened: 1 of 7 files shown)


macanderson:
`dod / dod` was red on #4862's five unticked DoD boxes. I checked each against the code rather than against this PR's description, found all five genuinely satisfied, and ticked them on the issue. What I verified, item by item:

1. **Variant, `as_str` arm and `observation_grade` arm gone.** `lifecycle.rs` drops the `MemoryCitation` enum option and its `"memory_citation"` arm; `provenance.rs` drops it from the `observation_grade` match, leaving `ReflectionLesson => ModelCritique` alone. ✅
2. **Old wire form no longer parses.** `memory_citation_is_retired_and_no_longer_parses` asserts `serde_json::from_str::<ObservationSource>(r#""memory_citation""#)` is `Err`. The "would have parsed on the old code" half holds by construction rather than by assertion: the enum keeps `#[serde(rename_all = "snake_case")]`, and the removed variant was `MemoryCitation`, so those exact bytes were its serde name. ✅
3. **`citation_observation` and its call site deleted.** Gone from `evidence.rs`; the `for citation in &citations` mining loop in `uses.rs::extract_one` is removed and replaced by a comment. The feedback loop above it that also reads `citations` is untouched, so the binding is still used. ✅
4. **The five store items stay, with honest doc comments.** `MemoryCitationRow`, `MemoryCitationStats`, `fold_citation_stats`, `PROMOTION_CITATIONS_REQUIRED` and `QUARANTINE_NEGATIVES_THRESHOLD` are all still in `stella-store/src/lib.rs`, each carrying a "declared, inert plumbing today" note; `record_memory_citations` says it has no production caller. ✅
5. **`evolution.rs`'s `Memory` row states the posture.** Checked on `main` rather than taking the description's word for it — the row's `mechanism` string reads "nothing writes `memory_citations` in the first place (#4862). Path-anchor validation is the only evidence source that retires anything today". Unmodified by this PR, as claimed. ✅

Re-run `dod` and it should clear.

Two things I did **not** verify and am not asserting: that `cargo test -p stella-records -p stella-cli` passes (CI's `fmt + clippy + test` was still in flight), and whether removing the loop in `uses.rs` leaves any import unused — clippy at `-D warnings` will answer that one either way.

Unrelated heads-up: PRs whose CI started after 11:57:49Z are hitting a broken `doc-warnings-schema` recipe that #5947 put on `main`. It only reddens the `docs/wire still matches the types` job, which this PR's diff does not trigger, so it should not affect you. Fix is #5993.

---

Issue #5991, opened 2026-09-05
doc-warnings-schema now cleans the whole target/doc tree, because cargo will not scope --doc to a package


macanderson:
## Context

#5947 added a `cargo clean --doc` to `doc-warnings-schema` to close #5139's stale-rustdoc-cache hole, and scoped it to the three schema crates:

```make
cargo clean --doc $(SCHEMA_CRATES)     # -p stella-protocol -p stella-plugin -p stella-serve
```

Cargo rejects that outright:

```console
$ cargo clean --doc -p stella-protocol
error: --doc cannot be used with -p
```

`cargo clean -p stella-protocol --doc` is rejected identically, so there is no spelling of the scoped form that works on the pinned toolchain (1.97.0). The recipe failed on its first line for every diff reaching `wire-schema.yml` — see the sibling PR that repairs it.

## What the repair does, and what it costs

It drops the scope: `cargo clean --doc`. That is correct and it keeps #5139's protection, but it pays exactly the cost #5947's own comment was trying to avoid — `doc-warnings` shares `target/doc`, so cleaning it unscoped means a full workspace rustdoc rebuild on every `make gate` run rather than the seconds the scoped form was reaching for.

So the hole is closed and the bill is now real. This issue is whether to buy the scope back.

## Options

1. **Leave it.** Correct, simple, and slow once per gate run. Worth measuring before assuming it is intolerable — `doc-warnings` may already be a small share of gate wall-clock.
2. **Give the recipe its own `CARGO_TARGET_DIR`** (e.g. `target/doc-schema`). Then cleaning it never disturbs `doc-warnings`' cache and the scope comes back in the only dimension cargo actually offers. Costs a separate dependency build for that directory the first time, cached after. This looks like the durable answer (SCR-002).
3. **Clean by hand** — `rm -rf target/doc/stella_protocol …`. Rejected on suspicion rather than measurement: it removes rendered output but leaves the fingerprints cargo consults, which is the half that made the stale build read as fresh in #5139. If someone establishes that the fingerprint is not what did it, this becomes cheap and viable.

## Files

- `Makefile` — the `doc-warnings-schema` recipe and the comment above it.

## Verify

- `make doc-warnings-schema` exits 0 from a cold tree and from a warm one.
- The #5139 property still holds: introduce a broken intra-doc link in one of the three schema crates, run the target twice, and it must fail both times rather than reporting a cached green on the second.
[...]

Issue #5139, opened 2026-08-26
The local gate misses the schema tier, and cargo doc replays a cached green


macanderson:
## Problem

Two local checks look thorough and are not, and they compose: a change can pass everything a session runs locally and still redden CI on a doc comment.

**1. `make guards-fast` does not compile the schema tier.** `doc-warnings-schema` and `lint-schema` are `make gate` steps, not `guards-fast` steps, and `guards-fast` is the rung the pre-push hook picks for a diff that reaches no crate. AGENTS.md already explains why those two steps exist — `ci.yml`'s `cargo doc --workspace` runs with default features, so every module behind the off-by-default `schema` feature is compiled by nothing else (#4584, #4949). What is missing is that the *local* ladder has the same hole in the same place.

**2. `cargo doc` replays a cached green.** `make doc-warnings-schema` passing tells you nothing if a previous invocation succeeded and the inputs rustdoc fingerprints have not changed in a way it notices. Only `rm -rf target/doc` before the run gives a real answer.

## How it showed up

Working #4017 I moved a function between modules in `stella-plugin` and ran `make guards-fast`, `cargo clippy --all-targets -- -D warnings` on five crates, and both full test suites. All green. CI then failed three separate checks on two one-line defects:

- `unused import: HookEvent` — the move took the last use of the type with it. Caught by `lint-schema`, and also by `windows-check.yml`, which runs clippy over the same crate.
- `unresolved link to super::point_moment`, then `unresolved link to HookEvent` — an intra-doc link in the new module's header. Caught by `doc-warnings-schema`.

The second one survived a *local* `make doc-warnings-schema` that printed nothing, because the doc build was cached. It reproduced only under `rm -rf target/doc && make doc-warnings-schema`. That cost two CI round trips.

## Where

- `Makefile` — `GUARDS_FAST_STEPS` / `GUARDS_STEPS` / `GATE_STEPS`, and the `doc-warnings-schema` and `lint-schema` recipes.
- `.githooks/pre-push` and `scripts/impacted-crates.sh` — which rung a push picks.
- `AGENTS.md` § "The gate — what every push is held to" — the four-rung table, and `scripts/check-gate-parity.sh`, which holds that table to `GATE_STEPS`.

## Options, none obviously right

1. **Move the two schema steps into `guards`** (the rung above `guards-fast`, which already compiles the two schema exporters for `wire-schema`). Cheapest, and it matches the existing precedent that `guards` is where "something compiles" starts.
[...]

PR #5993, opened 2026-09-05
fix(gate): drop the package filter cargo refuses on `cargo clean --doc`

macanderson:
## What & why

`doc-warnings-schema` has not run since #5947. Its first line was
`cargo clean --doc $(SCHEMA_CRATES)`, and cargo rejects that pair outright:

```
$ cargo clean --doc -p stella-protocol
error: --doc cannot be used with -p
```

So the recipe exited 2 before either `cargo doc` invocation ever started —
`make: *** [Makefile:644: doc-warnings-schema] Error 101` — taking `make gate`
and the required `wire-schema` job (`docs/wire still matches the types`) down
with it on every PR that reached them. Found on #5963, whose diff has nothing
to do with it.

Cargo has no package-scoped doc clean, so the fix is the unscoped form.

The comment above the recipe was the premise the broken line rested on — it
claimed `cargo clean --doc` was "the package-scoped, cargo-native version" of
#5139's `rm -rf target/doc` fix, "for exactly the three crates this recipe
documents". That is not a thing cargo can do. Both that comment and AGENTS.md's
copy ("for the three schema crates only") now say what cargo actually does and
name what the unscoped clean costs.

Closes #6003
Refs #5992

## The witness

- [x] No witness needed (pure CI/build-tooling fix) — because: the change is one
      flag on a `make` recipe, and the recipe could not execute at all on the old
      code. There is no input on which the old line passed, so a test asserting
      "the recipe runs" has nothing to distinguish beyond what running it shows.
      The regression guard that *would* hold this shape is filed as #5992 rather
      than bolted on here — it needs its own `GATE_STEPS` entry, both prose
      copies, and a hermetic self-test.

Verified instead, on this branch:

```
$ make doc-warnings-schema        # exit 0
 Documenting stella-plugin v0.9.332
 Documenting stella-protocol v0.9.332
 Documenting stella-serve v0.9.332
```

against `Error 101` on the first line before the change.

The clean is there to defeat cargo's doc freshness check (#5139), so that
property had to survive dropping the filter. It does, and the evidence
separates the two answers rather than being consistent with both — a second
`cargo doc` with no clean between finishes in 0.11s and prints **no**
`Documenting` line (rustdoc skipped); one after `cargo clean --doc` prints it
(rustdoc re-ran).

**This PR's own CI cannot show any of that.** `wire-schema.yml` is not triggered
by a `Makefile` diff, and it is the only workflow that runs
`doc-warnings-schema` — which is exactly how #5947 landed green. Do not read a
[...]

Issue #5992, opened 2026-09-05
`doc-warnings-schema` can break without any workflow running it, because the Makefile is outside wire-schema.yml's paths


macanderson:
## Problem

`doc-warnings-schema` shipped in #5947 with `cargo clean --doc $(SCHEMA_CRATES)`
as its first line. Cargo refuses that pair outright:

```
$ cargo clean --doc -p stella-protocol
error: --doc cannot be used with -p
```

So the recipe exited 2 before either `cargo doc` invocation, every time it ran,
for as long as it was on `main`. The line is repaired in #5993. This issue is
about why nothing noticed, which is the more durable defect.

## Root cause: the recipe's only runner never sees the recipe change

`doc-warnings-schema` runs in exactly one workflow —
`.github/workflows/wire-schema.yml` — and that workflow's `paths:` filter is

```yaml
- "docs/wire/**"
- "crates/stella-protocol/**"
- "crates/stella-serve/**"
- "crates/stella-plugin/**"
- "scripts/export-agentevent-schema.sh"
- "scripts/check-wire-schema.sh"
- ".github/workflows/wire-schema.yml"
```

`Makefile` is not on it. So a diff that changes the recipe starts nothing that
runs the recipe: #5947 was a Makefile-only change to `doc-warnings-schema`, and
its own CI never executed the step it rewrote. It merged green. The break then
surfaced only on the next unrelated PR whose diff happened to touch a
wire-contract path — #5963, some seven hours later, on a diff about plugin
stage bands.

The same hole applies to #5993, the repair: its diff is `Makefile` + `AGENTS.md`,
so its CI cannot demonstrate the fix either. That PR says so and cites a local
run instead.

**Adding `Makefile` to that `paths:` list is the wrong repair.** The list answers
one specific question — "can this diff have invalidated the generated artifacts
under `docs/wire/`?" — and `scripts/wire-schema-paths.sh` re-derives
`.githooks/pre-push`'s rung choice from the same list precisely so both answers
stay that one question. A Makefile edit cannot invalidate `docs/wire/`, so
putting it there would make the hook take the dearer rung on a false premise and
would blur the meaning the script's header is emphatic about keeping exact.

## What to decide

The two questions have been conflated because one workflow happens to host both
steps:

1. *Could this diff have invalidated the wire artifacts?* → `wire-schema`
2. *Could this diff have broken the rustdoc-over-schema-crates step?* → `doc-warnings-schema`

[...]

PR #5995, opened 2026-09-05
fix(gate): drop the -p from doc-warnings-schema's clean, which cargo refuses

macanderson:
## What and why

`doc-warnings-schema`'s first line was `cargo clean --doc $(SCHEMA_CRATES)`, and `SCHEMA_CRATES` is `-p stella-protocol -p stella-plugin -p stella-serve`. Cargo refuses that combination:

```console
$ cargo clean --doc -p stella-protocol
error: --doc cannot be used with -p
```

`cargo clean -p stella-protocol --doc` is refused identically, so **no spelling of the scoped form works** on the pinned toolchain (1.97.0).

The recipe therefore failed on its own first line, for every diff that reaches `wire-schema.yml` — `docs/wire/**`, `crates/stella-protocol/**`, `crates/stella-serve/**`, `crates/stella-plugin/**`. It landed in #5947 and merged as `a2fe819`, which is on `main` now. #5963 is where it surfaced: a diff about wrapper stage bands, failing a check named `docs/wire still matches the types` on a rustdoc clean it never asked for.

Not caught before the merge because `wire-schema.yml` triggers on `pull_request` only, and #5947's own diff (`Makefile`, `AGENTS.md`, `CONTRIBUTING.md`, `.github/workflows/**`, `scripts/**`) touches none of those paths — so the workflow that runs this recipe never ran on the PR that changed it.

## The fix

Drop the scope. The clean stays, so #5139's protection stands: cargo's doc freshness check can mark a `doc` unit up to date — printing nothing — off a stale prior build, which is what let a broken intra-doc link in `stella-plugin` survive a local run and reproduce only under a hand-run `rm -rf target/doc`.

**The cost is real, and is now written down instead of implied.** `doc-warnings` shares `target/doc`, so an unscoped clean means a full workspace rustdoc rebuild on each gate run — the "correct but slow" side of the trade #5139 weighed against. Taken because the only other option available today is to drop the clean and reopen the hole. #5991 holds whether to buy the scope back; a separate `CARGO_TARGET_DIR` for this recipe is the obvious candidate, since that is the one dimension cargo does let you scope.

`AGENTS.md`'s rung section described the clean as "for the three schema crates only", which this makes false, so it now says what the recipe does.

## Witness

The recipe is its own flip, run both ways in one tree:

```console
$ git stash && make doc-warnings-schema
cargo clean --doc -p stella-protocol -p stella-plugin -p stella-serve
error: --doc cannot be used with -p
make: *** [Makefile:644: doc-warnings-schema] Error 101        # exit 2

$ git stash pop && make doc-warnings-schema
   Generated target/doc/stella_plugin/index.html and 5 other files   # exit 0
```

A Rust witness test is not available for this one: the subject is a Makefile recipe's argument list, and what proves it is running the recipe.

Guards run locally, each exit code read directly rather than through a pipe (per AGENTS.md's own gotcha about `$?` after a filter): `check-prose`, `check-gate-parity`, `check-line-citations`, `check-god-files`, `check-file-size`, `check-left-behind` — all `0`.

`check-prose` earned its keep here: the first draft of the AGENTS.md paragraph raised that file's `issue-reference` count from 97 to 99. The prose was rewritten to state the facts rather than cite the numbers; no baseline entry was added.

## Tests deleted

None.

## Residue

- **#5991** — whether to restore the scope, and at what cost. Filed with the three options weighed (leave it and measure; a separate `CARGO_TARGET_DIR`; a hand `rm -rf` of the three doc dirs, which is suspected not to defeat the fingerprint that caused #5139 and is rejected on that suspicion rather than on measurement).

Refs #5139, #5947, #5991

https://claude.ai/code/session_01P2Pca8syUyEdAi7KK6LVXk

---
_Generated by [Claude Code](https://claude.ai/code/session_01P2Pca8syUyEdAi7KK6LVXk)_

## Summary by Sourcery

Restore the doc-warnings-schema gate by using a cargo-compatible documentation clean while preserving protection against stale rustdoc output.

Bug Fixes:
- Fix the doc-warnings-schema gate by removing the unsupported package scoping from cargo's documentation clean command.

Enhancements:
- Document the resulting workspace-wide documentation clean and its full rustdoc rebuild cost.

Documentation:
- Update contributor guidance to accurately describe the documentation clean behavior and its trade-offs.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -515,11 +515,18 @@ stays `gate`-only on purpose, paired with `doc-warnings`: at `check`, rustdoc
 never runs, for either feature set.
 
 `doc-warnings-schema` also wipes its own doc output before each run
-(`cargo clean --doc`, for the three schema crates only). `cargo doc`'s own
-freshness check can call a stale prior run "up to date" — and print nothing
-— even after the source changed. That let a broken doc link in
-`stella-plugin` pass a local `make doc-warnings-schema`, and show up only
-after a hand-run `rm -rf target/doc`.
+(`cargo clean --doc`). `cargo doc`'s own freshness check can call a stale
+prior run "up to date" — and print nothing — even after the source changed.
+That let a broken doc link in `stella-plugin` pass a local
+`make doc-warnings-schema`, and show up only after a hand-run
+`rm -rf target/doc`.
+
+The clean takes the whole `target/doc` tree, because cargo rejects `--doc`
+alongside `-p` and there is no scoped spelling. A scoped one shipped once and
+failed the recipe on its own first line for every diff that reaches
+`wire-schema.yml`. `doc-warnings` shares that tree, so it now rebuilds each
+gate run; the recipe's own comment carries what that costs and what would buy
+the scope back.
 
 **Every rung needs `shellcheck` on `PATH`, and it is not vendored.** It is the
 gate's one external binary, so a machine or container image without it stops at
--- a/Makefile
+++ b/Makefile
@@ -629,19 +629,26 @@ SCHEMA_FEATURES := stella-protocol/schema,stella-plugin/schema,stella-serve/sche
 # the source changed, which let a broken intra-doc link introduced by moving a
 # function between modules in `stella-plugin` survive a local
 # `make doc-warnings-schema` and reproduce only under a hand-run
-# `rm -rf target/doc` (#5139). `cargo clean --doc` is the package-scoped,
-# cargo-native version of that same fix: it drops the rustdoc output AND the
-# fingerprints cargo used to call the stale build fresh, for exactly the three
-# crates this recipe documents, so the two invocations below always run
-# rustdoc for real rather than trusting a cache that has already been wrong
-# once. Scoped to $(SCHEMA_CRATES) rather than the whole `target/doc` tree —
-# `doc-warnings` below has the same theoretical exposure, but cleaning its
-# workspace-wide output before every gate run would force a full rustdoc
-# rebuild every time, which is the "correct but slow" option #5139 weighed
-# against for a hole this recipe's own repro is what actually demonstrated.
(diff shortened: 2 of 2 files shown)


macanderson:
Closing this as a duplicate of #5993, which is the same fix, opened first, and more accurate than mine on the one point where we differ.

#5993 was created at 12:08:05Z and this at 12:10; both drop the `-p` filter from `cargo clean --doc` and both correct AGENTS.md's "for the three schema crates only". I did run `./scripts/main-red-claim.sh check` before starting and it cleared me — with `note: gh is not installed, so this run could not ask whether the repair is already claimed. Proceeding` — which is the fail-open behaviour that script documents, working as designed. The duplication is the cost that design accepts.

**Where #5993 is right and this PR is wrong.** I described the cost as "a full workspace rustdoc rebuild on each gate run". That is overstated. `GATE_STEPS` lists `doc-warnings` *before* `doc-warnings-schema`, and `gate:` depends on them in that order, so within one `make gate` the workspace rustdoc has already run by the time this step cleans — the clean invalidates the cache for the *next* run, not the current one. And `wire-schema.yml` runs this step without `doc-warnings` at all. #5993's comment says exactly that; mine would have put a wrong cost into the Makefile permanently.

Nothing here is worth porting across. The one thing this PR carries that #5993 does not is the local flip transcript (`exit 2` → `exit 0` on the real recipe), and that is recorded in this description for whoever wants it.

The structural half — that `wire-schema.yml` is the only runner of `doc-warnings-schema` and its `paths:` filter does not include the `Makefile` that defines it — is **#5992**, filed alongside #5993. That is the defect worth more than either patch: this PR's own CI does not run `wire-schema` either, because its diff is `Makefile` + `AGENTS.md`, so neither of these PRs can demonstrate the repair from its own checks. The proof arrives after merge, when #5963 (currently red on this exact recipe) merges `main` and goes green.

I have also corrected the pointer I left on #5963, which named this PR as the fix.

---

Issue #3942, opened 2026-08-19
self-driving: the driver process — poll step() and keep polling through a block


macanderson:
Part of #3599. Depends on #3939 (work) and #3940 (deliver). The step machine landed in #3935.

## Problem

`stella_autonomy::step` decides the loop's next move and nothing calls it in a loop. There is no process that asks "what now", does it, and asks again.

## What to build

A driver that polls `step(&state, &observation, &doctrine)`, dispatches the returned `LoopStep` to the verbs, updates state, and repeats.

## The one obligation the machine cannot discharge for itself

**On `LoopStep::Blocked`, keep polling. Do not exit.**

The self-resume property is structural — the machine holds no latch, so a block is recomputed every call and stops being returned the moment its cause is gone (`a_cleared_block_resumes_with_no_resume_signal` witnesses all four reasons). But a host that exits the process on `Blocked` makes resumption impossible no matter what the machine returns. The variant carries a `Clearance` naming the observable to watch precisely so a host has something to poll rather than a reason to quit.

Nobody should ever have to tell this loop it may resume. A human who raises the ceiling, restores the grant, drops the stop flag or merges the escalated PR has already said everything that needs saying.

## Observation gathering

The driver is what fills `LoopObservation`, including `base_fix_contention`. Four signals, only one authoritative:

| Signal | Source |
|---|---|
| ledger claim | `stella-fleet` ledger — the only lease with an owner and expiry |
| open PR | forge |
| remote branch | `git ls-remote` |
| **local worktree** | `git worktree list` on **this machine** |

The last one is the signal nobody remembers: two self-driving sessions against one clone see each other **only** there.

## Relationship to what exists

`scripts/self-driving.sh` is the working driver today and is deliberately untouched. Per `doc:backlog-self-driving` §9 B2 it retires only after `scripts/test-self-driving.sh` is green against the new path with every assertion intact.

Whether this is a subprocess plugin over the B0 driver channel or an in-binary loop first is open — #3783 (nothing opens a driver session) is the channel half.

## Done when

The loop runs unattended, parks on a stop, and resumes on its own when the cause clears — with no resume input of any kind.

PR #6007, opened 2026-09-05
fix(stella-cli): start a work unit with the code graph the repository has

### LINKED ISSUE #4003: Every autonomous work unit rebuilds the full code-graph index, because each worktree starts empty

macanderson:
## What and why

A self-driving work unit runs in a git worktree cut from the base ref, and the worktree had no code-graph index. So the first thing every turn did was parse the whole repository — 1 650 files and `0 unchanged`, three times over for three issues — out of the spend limit meant for the work.

Two halves, because either one alone is wrong. This is the design posted on the issue before the work started, and it follows the decision already recorded there ("the seeding option is the one") plus the later note that seeding alone would now seed a file nothing reads.

**1. `codegraph.db` stops following the state-root redirect.** `STELLA_WORKSPACE_STATE_ROOT` exists so a throwaway worktree's *learning* outlives it: the reflection log, `store.db`, `context.db`. The index is not learning. It is a parsed image of the files under the workspace root, so the redirect wrote one tree's index over another checkout's and handed two concurrent work units one writable database — the constraint this issue names, with #2293's hazard.

`stella_store::TREE_ANCHORED_STATE` states the carve-out, applied inside the one resolver that already decides where a private file lives. Every production site that resolves the index routes through that resolver, so no call site changes. #4394's invariant holds unchanged — build and readers still ask the store, so they still resolve one file — and two hand-joined readers that could never honour a redirect (`stella_graph::load_storage_snapshot` and the Observatory's graph snapshot) now agree with the resolver instead of disagreeing under one.

**2. The worktree is seeded from the repository.** A copy, never a hard link: a hard link is one file with two names, which is the shared writable index again. On APFS and btrfs `std::fs::copy` clones extents, so it is close to free.

`stella-graph` owns the seed (`seed_index`) because the copy needs something a file copy cannot do: clear `code_graph_leases`. A lease copied from a live parent process stands the child's own catch-up pass down for its full ten-minute TTL, and a turn that cannot index its own edits is #4394 by another route. The seed opens the copy through the store's verified open, which runs `PRAGMA quick_check` and quarantines a torn image — so the worst case is one cold build, which is today's cost, and never a bad index. Both the serial work unit and a parallel wave worker seed; a Claude Code worker does not, because it reads no `codegraph.db`.

## Witness mechanism

- `crates/stella-store/tests/tree_anchored_code_graph.rs` — with the redirect set, `codegraph.db` resolves under the worktree while `store.db` resolves under the state root. **Fails on `main` by construction**: the resolver moves both, so the first assertion sees the index under the state root. Its own test binary because it sets an environment variable the resolver reads from `std::env`.
- `crates/stella-tools/tests/codegraph_redirect_witness.rs` — kept and retargeted. It still proves #4394's "one database, not two"; the answer both resolvers agree on is now the tree, and it asserts the state root gets no `private/` directory at all.
- `crates/stella-graph/src/seed.rs` — a seeded index holds the parent's rows and none of its leases: the parent takes an `IndexWalk` lease before the copy, and the child's `index_all_single_flight` still runs and reports `files_parsed == 0`. Without the lease clear that call returns `None`. A second test proves an index already at the destination is never clobbered.
- `crates/stella-cli/src/self_driving_cmd/graph_seed.rs` — the worktree's index is a different file from the repository's (compared by inode on unix, so a hard link cannot pass), and an unindexed repository seeds nothing rather than failing the unit of work.

## Exemplar

The carve-out follows the shape `stella-home` already uses for the user tier: one named rule, resolved in one place, with the ambient read isolated (`resolve_workspace_state_root`). `seed_index`'s "copy, verify, then repair the copied-in state" ordering follows `store::open`'s own quarantine-and-retry path in this crate.

## Verify

The issue's own check is a live run: `stella self-driving drive --max-issues 3`, counting `code graph:` lines. That needs a provider key and real spend against real issues, so no CI job here can run it and no unit test stands in for it. It is **not claimed by this PR**. It is filed as #6018 under SCR-004, and #4003's Definition of done says so instead of carrying a box nobody can tick.

What this PR does prove is every link of the chain that run would exercise, each with its own test:

1. The worktree resolves its own `codegraph.db` — `tree_anchored_code_graph.rs`.
2. The seed lands the repository's rows there and drops the copied leases — `seed.rs`.
3. The turn's own catch-up pass then runs and reports `files_parsed == 0` on an unchanged tree — `seed.rs`, same test.
4. The file is a copy, not a hard link — `graph_seed.rs`, compared by inode on unix.

## Tests deleted

None.

## Residue issues filed

- #6018 — measure the saving in a real self-driving run, the live count above.

Nothing else was noticed and left. Two adjacent facts are recorded rather than filed, because both are now correct rather than pending: the Observatory's hand-joined path and `stella_graph::load_storage_snapshot` agree with the resolver again, and the ignored `search_latency_replay` harness reads the index from its replay root's own `.stella/private/` (its doc says so).

Closes #4003
## Summary by Sourcery

Start self-driving work units with an isolated, verified copy of the repository’s code graph while keeping session state durable and redirected.

New Features:
- Seed self-driving worktrees and parallel workers with a private copy of the repository’s existing code graph index.

Bug Fixes:
- Keep code graph databases anchored to the workspace tree while continuing to redirect session state, preventing concurrent worktrees from sharing or overwriting one index.
- Ensure seeded indexes do not retain leases from the source process and remain safe to use after partial or interrupted copies.
- Make code graph readers and snapshots consistently resolve the tree-anchored database.

Enhancements:
- Add centralized handling for tree-anchored workspace-private files and verified code graph seeding with safe fallback to a cold build.
- Preserve existing destination indexes and gracefully skip seeding when the repository has not been indexed.

Documentation:
- Clarify the separation between disposable workspace code and durable session state, including the tree-anchored code graph.

Tests:
- Add coverage for tree-anchored path resolution, safe index copying, lease removal, unchanged-tree catch-up, destination preservation, and unindexed repositories.


macanderson:
From the PR sweep: Sourcery's one `❌` row is **correct**, and it is the only thing standing between this PR and a complete DoD for #4003. Flagging rather than fixing, because what it asks for is a measurement this session cannot take.

**The row:** *"Verify the optimization in a multi-issue self-driving run, showing one index-build event with high unchanged counts for subsequent work units."*

#4003's definition of done states it as its own box, in the issue's own words:

> - [ ] Verified as this issue says: count `code graph:` lines in the log — three today, one after a fix, with the later two reporting a high `unchanged` count.

The issue is a **measurement** — three full index builds observed in #3942's live session, with `0 unchanged` on each. Its DoD asks for the same measurement taken again after the fix, and that is the half this PR does not carry. What is here is the mechanism plus unit and witness coverage: seeding, lease clearing, tree anchoring, the inode check proving two work units never share a writable index. Those establish that the index *can* be reused; they do not establish the observed counts the issue names.

That distinction is this repository's own bar — a passing unit test and a drive log showing `1650 re-parsed, 0 unchanged` are not the same evidence, and it was the second that opened the issue.

**What would settle it:** one `stella self-driving drive --max-issues 3` against this branch, with the `code graph:` lines pasted into this PR — one build, and the later two units reporting a high `unchanged` count. That needs a provider key and a real multi-issue run, so it wants a maintainer session rather than a sweep.

All three of #4003's DoD boxes are still unchecked, so the `dod / dod` gate will hold this PR regardless of CI. If the measurement is deliberately being deferred, the SCR-004 route is to split that box into its own `triage` issue and tick the rest — but it should not be ticked on the strength of the tests alone.

Nothing else here needs attention: no merge conflict against `main`, no unresolved review threads, and the other two objectives are `✅`.

---

macanderson:
**DoD sweep (autonomous PR routine): two of #4003's three boxes ticked, one gap left open.**

I read the tests rather than the PR description, and ticked only what they establish.

**Ticked — item 1** (no full rebuild). `crates/stella-graph/src/seed.rs`'s `a_seeded_index_carries_the_rows_and_none_of_the_leases` takes an `IndexWalk` lease on the parent before the copy, seeds the child, and asserts the child's own `index_all_single_flight` returns `files_parsed == 0`. That is the `0 unchanged` this issue measured, inverted — and the lease clear is what makes it more than a file copy.

**Ticked — item 3** (never a shared writable index). Two independent halves, each with its own witness: `crates/stella-store/tests/tree_anchored_code_graph.rs`'s `the_code_graph_stays_with_the_tree_while_the_store_follows_the_redirect` proves two worktrees resolve two different `codegraph.db` files under the redirect, and `graph_seed.rs`'s `the_worktree_gets_its_own_file_not_the_repositorys` compares inodes on unix so a hard link cannot pass.

**Left open — item 2** (the live count). The PR restates the issue's check verbatim under "Verify" but reports no run of it: there is no `stella self-driving drive --max-issues 3` log in the description, and the unit test above is a mechanised equivalent rather than the measurement the box asks for. I did not tick it, because ticking a box on a run nobody made is the one move `CLAUDE.md` calls unrecoverable. This is what keeps `dod / dod` red — every other check on this head is green.

What would settle it, cheapest first:

1. Paste a `stella self-driving drive --max-issues 3` log excerpt showing one `code graph:` build line and two later lines with a high `unchanged` count. That is the box as written.
2. Or rewrite item 2 to name the unit test as its evidence, if the maintainer judges `files_parsed == 0` under a real seed to be the same fact more cheaply obtained. That is a scope call on the issue's own wording, not mine to make.

No other blocker: `fmt + clippy + test`, MSRV, Windows, macOS, docs guards and the ratchets are all green on `1bb575e`.

---

macanderson:
**Answering the one ❌ in the assessment table** — the third row, "verify the
optimization in a multi-issue self-driving run, showing one index-build event
with high unchanged counts".

Sourcery is right about the diff: this PR ships no drive log. That half of the
objective is **deliberately not claimed here**, and it is now filed as #6018
under SCR-004 rather than left implied.

The reason is that the check cannot be run by anything in this repository. It
needs a funded `stella self-driving drive --max-issues 3` against real issues
with a provider key. No CI job runs one, and a unit test cannot stand in for a
measurement. Ticking a box for a run nobody made is the thing CLAUDE.md's
evidence rule forbids, so #4003's Definition of done now names #6018 in place
of that box, and this PR's Verify section says the same.

What the PR does settle is every link the run would exercise, each with a test
that fails without the change:

1. A worktree resolves its own `codegraph.db` while the rest of the session
   state still follows `STELLA_WORKSPACE_STATE_ROOT` —
   `crates/stella-store/tests/tree_anchored_code_graph.rs`.
2. The seed lands the repository's rows at that path and clears the copied
   `code_graph_leases` — `crates/stella-graph/src/seed.rs`.
3. The child's own catch-up pass then runs and reports `files_parsed == 0` on
   an unchanged tree — same test. Without the lease clear the call returns
   `None` instead.
4. The seeded file is a copy, not a hard link, compared by inode on unix —
   `crates/stella-cli/src/self_driving_cmd/graph_seed.rs`.

The other half of that same row — "concurrent work units never share a writable
index" — is settled by 1 and 4 together, and is ticked on #4003.

Issue #4495, opened 2026-08-23
check-deleted-tests reads the PR body from the event payload, so naming a deleted test after the fact never works without a new push


macanderson:
## Problem

`scripts/check-deleted-tests.sh`'s remedy says "name the test in the PR description", but CI passes `PR_BODY` from the `pull_request` event payload — the body as it was when the run was triggered — so editing the description and re-running (or a re-run of the failed job) never satisfies it; only a new push does. The `git log` fallback is dead at `fetch-depth: 2`. Hit three times on 2026-08-23 (PRs #4419, #4448 twice); PR #4456 added the explanation to the failure text, which is the mitigation.

## Done means

The guard reads the *current* body through the API (`gh pr view --json body` with the workflow token) so an edited description counts on re-run, or the step explicitly says a push is required and why. Self-test in `scripts/test-deleted-tests.sh` (if absent, add it and host it per #4427's parity guard).

Related: #1976, #4456, #4427.

PR #6008, opened 2026-09-05
feat(stella-cli): route the export archive's transcript through stella-transcript

### LINKED ISSUE #4270: The exported session archive renders its own HTML instead of using the shared transcript renderer

macanderson:
## What & why

`crates/stella-cli/src/export/transcript.rs` folded a session's event
journal into its own hand-written HTML rows. `stella_transcript::html::render_page`
had exactly one production caller (the Observatory), so the same session
rendered by `stella export` and `stella observe` produced two different
documents.

This PR routes the archive's transcript through the shared renderer:
`export/transcript.rs` now folds `Store::session_events` into a
`stella_transcript::model::Run` and calls `html::render_page` — the same
call the Observatory already makes.

## Design

The issue's own investigation comment named three routes for a CSS/geometry
collision the migration runs into and left the choice to whoever picked it
up. Full reasoning is in [the issue thread](https://github.com/macanderson/stella/issues/4270#issuecomment-5551750031)
— short version: the transcript now ships as its own `transcript.html`
archive member rather than an embedded fragment, because the dashboard's
stylesheet and the shared renderer's define six class names differently
with neither namespaced. Two self-contained documents cannot cross-talk;
one merged page could. `dashboard.html` links to it instead.

Three properties the old fold carried survive the move:

- **Redaction** is now one whole-tree pass in `redact_run`: serialize the
  built `Run` to JSON, mask every string, deserialize back. Every
  `stella_transcript::model` type round-trips through `serde_json` byte for
  byte (AGENTS.md invariant #4), so this can't miss a field the way the old
  fold's per-call-site `clean()` calls could.
- **The elapsed clock** is summed from real per-call `duration_ms` on
  `ToolResult`/`StepUsage`, the same way
  `stella_tui::transcript_build::RunBuilder` already does for the live
  surface. `events.ts` (second-resolution) is never read.
- **Nothing is dropped.** Every event kind the shared model has no slot for
  becomes a `Note` carrying its wire tag and full JSON payload — including
  several kinds (`SpeculationDiscarded`, `BlockRegistered`, `StepManifest`,
  `CandidateDelivery`, the typed decision events) the old fold's own
  `RunBuilder`-style narration path would have silently dropped.

## Tests

`crates/stella-cli/src/export/transcript/tests.rs` is a fresh suite that
mostly asserts on the folded `Run` directly rather than on rendered HTML —
`stella-transcript` owns its own tests for what a `Run` renders as, and this
crate owns what the fold puts *into* the run. Covers: one message per turn
(the `TextDelta`+`Text` collapse), an unconsolidated delta run surviving, a
reasoning run coalescing to one prose block, every event kind becoming a
note, tool-result correlation (including an orphaned result with no
recorded `ToolStart`), a call that never returned staying `Running`, file
diffs attaching to the call that produced them, measured (not invented)
step offsets, per-execution turn prompts, the row budget's overflow
counter, and the whole-tree redaction pass. A handful of end-to-end tests
at the bottom check the rendered document directly (no `<script>` tag, a
credential never reaching the HTML, markup escaped).

`export/tests.rs`'s row-grammar witness
(`the_archive_carries_the_session_transcript_and_only_that_session`) is
re-pointed from the old `class="ev tool"` marker to the shared renderer's
`<details class="step">` — same property (the exported session's tool call
is present, another session's is not), checked against the new markup.

**Tests deleted.** Every one named below, as `check-deleted-tests` asks. This
PR rewrites `crates/stella-cli/src/export/transcript/tests.rs` wholesale,
because the hand-written HTML fold those tests pinned is exactly what is being
retired. Grouped by what happened to the property each covered.

*Renamed — same property, asserted against the new shape:*

- `a_call_that_never_returned_says_so_rather_than_looking_successful` →
  `a_call_that_never_returned_stays_running_rather_than_looking_successful`
- `a_credential_in_the_event_stream_never_reaches_the_transcript` →
  `a_credential_never_reaches_the_rendered_document`, plus
  `a_credential_anywhere_in_the_run_is_masked_by_the_whole_tree_pass` — the
  whole-tree pass is strictly wider than the old per-call-site `clean()`
- `a_delta_run_with_no_consolidating_text_is_still_rendered` →
  `an_unconsolidated_delta_run_still_reaches_the_answer`
- `an_assistant_message_streamed_as_deltas_renders_exactly_once` →
  `a_streamed_message_collapses_to_the_turns_answer_once`
- `a_reasoning_run_coalesces_into_one_thinking_row` →
  `a_reasoning_run_coalesces_into_one_prose_block`
- `an_oversized_payload_is_cut_and_says_that_it_was` →
  `an_oversized_tool_output_is_clipped_and_says_so`
- `every_event_kind_renders_rather_than_being_silently_dropped` →
  `every_event_kind_becomes_a_note_rather_than_being_silently_dropped`
- `the_transcript_is_readable_with_scripts_disabled` →
  `the_rendered_document_needs_no_script_to_read`, plus
  `the_dashboard_links_to_the_transcript_document` in `export/tests.rs`. Its
  tab/panel markup no longer exists — the transcript is a separate document
  now, not a script-gated panel — so the same "no script needed" property is
  checked against the new shape.

*Dropped because the code they covered is `stella-transcript`'s, and that
crate tests it directly.* This crate now owns what the fold puts **into** a
`Run`; what a `Run` renders as is the shared renderer's own suite:

- `a_modified_line_pair_highlights_only_the_changed_token` →
  `acceptance_word_diff_highlights_only_the_changed_token`
- `an_unpaired_addition_carries_no_word_tint` →
  `a_wholly_rewritten_line_drops_word_highlights_as_noise` and
  `a_long_line_under_the_cap_still_gets_word_highlights`, plus the dedicated
  `crates/stella-transcript/tests/word_highlight_matrix.rs`
- `a_file_diff_renders_as_a_gutter_diff_rather_than_a_wall_of_text` and
  `a_diff_that_parses_to_no_hunks_still_renders_as_readable_text` →
  `a_new_file_renders_as_an_all_green_diff`,
  `a_deleted_file_renders_as_an_all_red_diff` and
  `the_exported_gutter_is_not_lexed_as_source`
- `a_long_file_diff_shows_its_beginning_and_its_end_and_marks_the_middle` →
  `a_long_output_folds_head_and_tail_so_errors_at_the_end_stay_visible`
  (`FileDiff::is_large`, which both `html.rs` and `grid.rs` consult)
- `truncation_never_splits_a_utf8_character` → the shared renderer cuts on
  `char_indices`/`len_utf8` (`digest.rs`, `grid.rs`), with
  `a_digest_is_a_summary_not_truncated_content` pinning the policy

*Dropped because the behaviour is gone rather than moved.* The fold no longer
reads `events.ts` at all — elapsed is summed from measured `duration_ms`, per
the "elapsed clock" note above — so there is no journal clock left to test:

- `the_elapsed_clock_crosses_a_day_boundary`
- `the_elapsed_column_counts_whole_seconds_and_invents_no_precision`
- `an_unreadable_timestamp_holds_the_previous_reading`

What replaces all three is
`step_offsets_sum_from_measured_durations_not_from_the_journal_clock`.

## `make transcript-surfaces`

`scripts/check-transcript-surfaces.py`'s `export-html` row flips from `OWN`
(#4270) to `SHARED`, with no issue citation — the ledger's own contract for
a closed gap.

```
transcript-surfaces: 3 shared, 2 declared gaps, 0 foreign port(s) — ledger matches the tree.
    gap: deck-body -> #3578 (crates/stella-tui/src/render/entry.rs)
    gap: deck-spec6 -> #4289 (crates/stella-tui/src/views/transcript.rs)
```

## Definition of done

- [x] `stella_transcript::html::render_page` has two production callers.
- [x] `crates/stella-cli/src/export/transcript.rs` is reduced to the
      redacting fold plus the call.
- [x] `make transcript-surfaces` shows the `export-html` row as `Shared`
      with no issue citation.

## Residue

Filed a follow-up for a gap the issue's own investigation didn't surface:
`stella_transcript::html::render_page` emits no CSP meta tag, so the
standalone `transcript.html` carries none of the page-level backstop the
old embedded-fragment dashboard had (`default-src 'none'; …`). The new
document has no `<script>` tag at all, so there's no inline-script surface,
but it's still a real reduction in defense-in-depth worth closing
separately from this wiring change. Filed as #6009.

Closes #4270

## Summary by Sourcery

Use the shared transcript renderer for exported sessions and ship the replay as a standalone, redacted transcript document alongside the metrics dashboard.

New Features:
- Route exported session transcripts through the shared stella-transcript renderer used by the Observatory.
- Add a standalone, self-contained transcript.html document to export archives and link to it from the metrics dashboard.
- Preserve complete event coverage by representing unsupported event types as detailed transcript notes.

Bug Fixes:
- Prevent transcript redaction gaps by masking credentials across the fully constructed transcript model.
- Use measured tool and model durations for transcript timing instead of deriving elapsed time from journal timestamps.
- Preserve orphaned and unfinished tool calls rather than silently dropping or misrepresenting them.

Enhancements:
- Refactor the export transcript fold to build a shared Run model with prompts, tool calls, file changes, notes, accounting, and overflow reporting.
- Remove the dashboard's embedded transcript panel, duplicate row rendering, transcript-specific styling, and tab behavior.
- Keep transcript rendering and dashboard styling isolated in separate documents to avoid CSS collisions.

Documentation:
- Document the separate transcript.html archive member and its role as a replay rendered with the same component used by stella observe.
- Update telemetry documentation to describe export archives as containing both dashboard and transcript documents.

Tests:
- Replace transcript HTML-fold assertions with direct tests of the constructed Run model and shared-renderer integration.
- Add coverage for event preservation, tool correlation, unfinished calls, file attachments, measured offsets, row limits, redaction, script-free rendering, escaping, and archive linking.

Chores:
- Mark the export HTML surface as shared in the transcript-surfaces ledger and remove its resolved issue gap.

---

### DIFF
--- a/crates/stella-cli/src/export.rs
+++ b/crates/stella-cli/src/export.rs
@@ -256,7 +256,7 @@ fn redact_dump(json: &str) -> String {
 
 /// Recursively replace every string value in `value` with its redacted
 /// form. Returns whether anything was actually masked. [`redact_dump`]
-/// ignores that return value; [`transcript::redact_run`] uses it to say, on
+/// ignores that return value; `transcript::redact_run` uses it to say, on
 /// the dashboard, whether the transcript needed masking at all.
 fn redact_json_strings(value: &mut serde_json::Value) -> bool {
     match value {
--- a/crates/stella-cli/src/export.rs
+++ b/crates/stella-cli/src/export.rs
@@ -1,16 +1,22 @@
 //! `/export` — export **one session's** telemetry as a ZIP archive of raw
-//! JSON dumps plus a self-contained HTML dashboard.
+//! JSON dumps plus two self-contained HTML documents: a metrics dashboard and
+//! the session transcript.
 //!
 //! The archive lives at `.stella/exports/` and is named with the microsecond
 //! timestamp of the **last log entry** included (the data's own clock, not the
-//! user's submission time). The HTML is fully static — no external CSS/JS,
-//! everything inlined — so it can be opened offline, emailed, or committed
-//! alongside a PR as evidence.
+//! user's submission time). Both documents are fully static — no external
+//! CSS/JS, everything inlined — so either can be opened offline, emailed, or
+//! committed alongside a PR as evidence.
 //!
-//! The dashboard surfaces the metrics that actually change software quality:
-//! resolve rate, cost-per-resolved-task, token efficiency, tool-call
-//! frequency, retry patterns, and file-edit heat — the same data `stella
-//! stats` summarizes in a table, but visually and interactively.
+//! `dashboard.html` surfaces the metrics that actually change software
+//! quality: resolve rate, cost-per-resolved-task, token efficiency,
+//! tool-call frequency, retry patterns, and file-edit heat — the same data
+//! `stella stats` summarizes in a table, but visually and interactively.
+//!
+//! It links to `transcript.html` rather than embedding it. That document is
+//! [`transcript::render`]'s own [`stella_transcript::html::render_page`]
+//! output — the same renderer `stella observe` uses. Both draw one row
+//! grammar, and each keeps its own CSS.
 //!
 //! **Scope is a safety property here, not a convenience** (#2558). Until the
 //! session argument existed, this module dumped the entire workspace store:
@@ -102,9 +108,11 @@ pub fn export_session(workspace_root: &Path, session_id: &str) -> Result<PathBuf
     // the dashboard are still worth having, and an empty transcript reports
     // itself on the page rather than pretending the session was silent.
     let journal = store.session_events(session_id).unwrap_or_default();
-    let transcript = transcript::render(&journal, &execution_prompts(&dumps));
+    let transcript = transcript::render(&journal, &execution_prompts(&dumps), session_id);
 
-    // Build the self-contained HTML dashboard.
+    // Build the self-contained HTML dashboard. It links to the transcript
+    // rather than embedding it — see `transcript::render`'s doc for why the
+    // two are separate documents.
     let html = render_dashboard(
         &usage_stats,
         &dumps,
@@ -136,6 +144,12 @@ pub fn export_session(workspace_root: &Path, session_id: &str) -> Result<PathBuf
     }
     // The dashboard.
     zip.add_file(&format!("{folder}/dashboard.html"), html.as_bytes())?;
+    // The session transcript. A complete page on its own, so it reads fine
+    // even if nobody opens `dashboard.html` first.
+    zip.add_file(
+        &format!("{folder}/transcript.html"),
+        transcript.html.as_bytes(),
+    )?;
     // A manifest with the watermark, the scope, and the table list.
     //
     // `session` and `excluded` are the archive's provenance: without them a
@@ -184,10 +198,10 @@ pub async fn export_command(workspace_root: &Path, session_id: &str) -> String {
             "Export Session Telemetry — archive written to {}\n\
              Scope: this session ({session_id}) only — the archive is safe to attach to a \
              PR or email without disclosing your other runs in this workspace. The ZIP \
-             holds a `dashboard.html` (open in any browser), raw JSON dumps of this \
-             session's telemetry tables, and a `manifest.json` naming the session and \
-             what was excluded. The timestamped folder name matches the last log entry's \
-             timestamp.",
+             holds a `dashboard.html` (metrics — open in any browser), a `transcript.html` \
+             (the replayed session), raw JSON dumps of this session's telemetry tables, \
+             and a `manifest.json` naming the session and what was excluded. The \
+             timestamped folder name matches the last log entry's timestamp.",
             path.display()
         ),
         Ok(Err(e)) | Err(e) => format!("export failed: {e}"),
@@ -240,18 +254,38 @@ fn redact_dump(json: &str) -> String {
     }
 }
 
-/// Recursively replace every string value in `value` with its redacted form.
-fn redact_json_strings(value: &mut serde_json::Value) {
+/// Recursively replace every string value in `value` with its redacted
+/// form. Returns whether anything was actually masked. [`redact_dump`]
+/// ignores that return value; [`transcript::redact_run`] uses it to say, on
+/// the dashboard, whether the transcript needed masking at all.
+fn redact_json_strings(value: &mut serde_json::Value) -> bool {
     match value {
         serde_json::Value::String(text) => {
             let redaction = stella_learn::redact::redact_secrets(text);
             if redaction.redacted {
                 *text = redaction.text;
+                true
+            } else {
+                false
             }
         }
-        serde_json::Value::Array(items) => items.iter_mut().for_each(redact_json_strings),
-        serde_json::Value::Object(map) => map.values_mut().for_each(redact_json_strings),
-        serde_json::Value::Null | serde_json::Value::Bool(_) | serde_json::Value::Number(_) => {}
+        serde_json::Value::Array(items) => {
+            let mut any = false;
+            for item in items.iter_mut() {
+                any |= redact_json_strings(item);
+            }
+            any
+        }
+        serde_json::Value::Object(map) => {
+            let mut any = false;
+            for v in map.values_mut() {
+                any |= redact_json_strings(v);
+            }
+            any
+        }
+        serde_json::Value::Null | serde_json::Value::Bool(_) | serde_json::Value::Number(_) => {
+            false
+        }
     }
 }
 
@@ -448,18 +482,15 @@ fn render_dashboard(
     let stats_json =
         script_json(&serde_json::to_string(usage_stats).unwrap_or_else(|_| "[]".into()));
 
-    // What the transcript panel says about itself, above its first row.
-    let transcript_provenance = transcript.provenance();
-    let transcript_rows = comma(transcript.rendered as i64);
-    // An empty transcript is a real state, not a bug: a session whose events
-    // were never persisted, or one recorded by a build old enough that none of
-    // them replay. Say which, rather than rendering a blank panel that reads
-    // as a broken page.
-    let transcript_body = if transcript.rendered > 0 {
-        transcript.body.clone()
+    // The caption under the link to `transcript.html`: what it does and
+    // does not hold, so a reader can check the claim before opening it.
+    // An empty transcript is a real state, not a bug.
+    let transcript_caption = escape_html(&transcript.provenance());
+    let transcript_note = if transcript.rendered > 0 {
+        String::new()
     } else {
-        "<p class=\"empty\">No replayable events were recorded for this session. \
-         The metrics tab and the archive's <code>raw/</code> dumps are unaffected.</p>"
+        " No replayable events were recorded for this session; the metrics \
+         below and the archive's <code>raw/</code> dumps are unaffected."
             .to_string()
     };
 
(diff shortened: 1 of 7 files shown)


macanderson:
**PR sweep (autonomous routine): one of the two red checks fixed here, one needs an owner's call.**

Two checks are red on `3a33516`. Everything else — clippy, `cargo test`, rustdoc, MSRV, macOS, docs guards, the ratchets — is green.

### 1. `no test disappears in the merge unacknowledged` — fixed in the description

`scripts/check-deleted-tests.sh` reported **16 unnamed deletions**. Reproduced locally by merging this branch into current `main` (`e21c5b9`, clean, no conflict) and running the guard against it. All 16 lived in the one file this PR rewrites wholesale, `crates/stella-cli/src/export/transcript/tests.rs` — case 2 in the guard's own text, not case 1: the hand-written HTML fold they pinned is the thing being retired.

I checked each against the tree rather than against the description, and the description now names all 16, grouped by what happened to the property:

- **8 renamed**, same property against the new shape (`a_call_that_never_returned_says_so_rather_than_looking_successful` → `…_stays_running_rather_than_looking_successful`, and siblings).
- **5 dropped to `stella-transcript`**, which tests them directly — `acceptance_word_diff_highlights_only_the_changed_token`, `a_long_output_folds_head_and_tail_so_errors_at_the_end_stay_visible`, `the_exported_gutter_is_not_lexed_as_source`, the `word_highlight_matrix.rs` suite, and `a_digest_is_a_summary_not_truncated_content` for the UTF-8 cut.
- **3 dropped because the behaviour is gone**, not moved: the fold no longer reads `events.ts`, so the journal-clock tests have nothing left to assert. `step_offsets_sum_from_measured_durations_not_from_the_journal_clock` is what replaces them.

Verified against the merged tree with the new body:

```
check-deleted-tests: OK — 17 removed test(s), each named in the PR description or a commit.
```

The guard reads the PR's *current* body through the API (#4495), so this needs a re-run or the next push to go green — **this sweep cannot trigger either** (`rerun-failed-jobs` returns 403 for it). No new commit was pushed for it, because an empty commit to kick CI is what AGENTS.md forbids.

### 2. `dod / dod` — blocked, and Sourcery is wrong about it

#4270's first DoD box reads *"`stella_transcript::html::render_page` has two production callers."* It is unticked, and **it cannot be ticked on this head**. Checked in the tree, not taken from the description:

```
$ git grep -n "render_page(" -- 'crates/*.rs' | grep -v '^crates/stella-transcript/'
crates/stella-cli/src/export/transcript.rs:532:  let page = html::render_page(&run, &FoldState::new());
```

That is the only one, and it is the caller *this PR adds*. On `main` the count is **zero**. This PR's own opening sentence — "`render_page` had exactly one production caller (the Observatory)" — is stale: `stella-observatory/src/lib.rs:205` calls `html::render_run`, the fragment renderer, per #4566/#4593. So this PR takes `render_page` from zero callers to one, not from one to two.

Note that Sourcery's "Assessment against linked issues" table marks this objective ✅. It is mistaken, and a reviewer trusting that row would tick a box that is false.

What the box was *for* is met — `render_page` is `<!DOCTYPE …><div class="wrap">{render_run(run, state)}</div>…`, so export and observe now draw their bodies through one function, and `make transcript-surfaces` reports `export-html` as `SHARED` with no citation. The other two boxes are ticked and were verified on this head.

Two ways out, both an owner's call rather than a sweep's:

1. **Reword the box** to the entry-point family the ledger already uses — `html::render_{page,run,turn_tail}`, which is what `scripts/check-transcript-surfaces.py` calls `HTML`.
2. **Split out** "put the observatory back on `render_page`" as its own issue (`triage` label only, SCR-004) and tick the rest.

I did not tick it. A box ticked on a count the code does not support is the move `CLAUDE.md` calls unrecoverable.

### Everything else

No merge conflict against `main` (merged clean locally at `e21c5b9`), no unresolved review threads, no `hold` label.

---

Issue #6021, opened 2026-09-05
Make the streaming-deadline reverse-request test wall-clock independent


macanderson:
## Problem

`a_streaming_host_resets_the_reverse_request_deadline`
(`crates/stella-serve/tests/bridge.rs`) still measures the reverse-request
deadline against real wall-clock time (#3799 widened the margin from 2x to
3.3x, which lowers the flake rate but cannot remove it — a big enough
scheduler delay under load still trips the deadline early).

The deterministic fix — pausing `tokio`'s clock so the test's timing no
longer depends on real elapsed time — is blocked by an architecture gap:
`Session::start` (`crates/stella-serve/src/session.rs`) spawns the turn on
its own dedicated OS thread, in its own `tokio::runtime::Builder`
current-thread runtime, because the turn future is `!Send` and cannot run
on the caller's runtime at all today. The reverse-request deadline's
`tokio::time::sleep` (`crates/stella-serve/src/remote.rs::complete_remoted`)
runs inside that dedicated runtime. `tokio::time::pause`/`start_paused`
pauses one runtime's clock, so pausing the test's own runtime does nothing
for a sleep ticking on a different runtime, on a different OS thread.

## The fix

Give `Session` (or `stella-serve`) a **test-only** alternate entry point
that drives the turn's `!Send` future on the *caller's* runtime instead of
spawning a new OS thread, using `tokio::task::LocalSet` + `spawn_local`
(both handle `!Send` futures on a single runtime without needing a real OS
thread boundary). With that in place, a `#[tokio::test(flavor =
"current_thread", start_paused = true)]` test can drive both the deadline's
sleep and the test's own delta-gap sleeps on one paused clock, making the
outcome depend only on ordering, never on real elapsed time.

Constraints found while investigating #3799:

- `Session`, `SessionSpec`, `SettleHook`, `TurnSettlement` are the only
  public exports of `stella-serve::session` (see `crates/stella-serve/src/lib.rs`).
  `tests/bridge.rs` is a separate integration-test crate, so it only sees
  `stella-serve`'s public API — `#[cfg(test)]` inside the library does not
  reach it. A test-only entry point needs either a new public,
  `#[doc(hidden)]` method, or a Cargo feature (e.g. `test-support`) gated
  the same way. This workspace has no existing example of the
  self-referential `[dev-dependencies]` trick (a crate depending on itself
  with a feature enabled, so the feature is live for its own tests but not
  for consumers) — check for one before inventing the pattern here.
[...]

PR #6022, opened 2026-09-05
fix(stella-serve): widen the streaming-deadline test's wall-clock margin

### LINKED ISSUE #3799: A test fails only when the full test suite runs under load, not when it runs by itself
## Problem

`a_streaming_host_resets_the_reverse_request_deadline` (`crates/stella-serve/tests/bridge.rs:533`) fails inside a full `make test` / `cargo test --workspace` run and passes alone. Observed on `main` at `2188050`:

```
---- a_streaming_host_resets_the_reverse_request_deadline stdout ----
thread 'a_streaming_host_resets_the_reverse_request_deadline' panicked at crates/stella-serve/tests/bridge.rs:554:26:
a live stream must keep its request in flight: UnknownRequest("prov-2-0")
test result: FAILED. 10 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out
```

Control run, same tree, same binary, test alone: **3/3 pass** (`cargo test -p stella-serve --test bridge a_streaming_host_resets_the_reverse_request_deadline`).

## Why it flakes

The margin is 2x wall clock, and the test runs concurrently with the whole workspace suite:

```rust
let deadline = Duration::from_millis(120);
...
for n in 0..6 {
    tokio::time::sleep(Duration::from_millis(60)).await;
    session.resolve_provider_delta(&request_id, ...)
        .expect("a live stream must keep its request in flight");
}
```

A 60 ms sleep only has to overrun to 120 ms once — one scheduler delay on a loaded machine — for the reverse-request deadline to reap the request; the next `resolve_provider_delta` then returns `UnknownRequest` and the `expect` panics. `#[tokio::test(flavor = "multi_thread", worker_threads = 2)]` gives the runtime two threads on a box where `cargo test --workspace` is saturating every core.

Note the failure mode is a *false* red: the property under test (fragments count as progress against the deadline) is not violated — the test's own timing assumption is.

This is the same shape as #2011, #2804, #1721 and #2603: a wall-clock budget sized for an idle machine, asserted under full-suite parallelism.

## Fix guidance

- Preferred: make the test not depend on wall clock. `tokio::time::pause`/`start_paused` needs `flavor = "current_thread"`, which this test cannot use as written (`Session::start` parks a thread), so check whether the session can be driven from a current-thread runtime before reaching for it.
- Otherwise widen the absolute margin while keeping the ratio that makes the assertion meaningful: the test needs `gap < deadline` and `6 * gap > deadline`, which a 1 s deadline with 300 ms gaps satisfies with 3.3x of slack per gap instead of 2x. Slower, but this test is one of eleven in the file.
- Do **not** fix it by removing the assertion or marking the test `#[ignore]`: the deadline-resets-on-progress property is exactly what #2686's streaming work depends on.

## Verify

1. `cargo test -p stella-serve --test bridge` alone — passes before and after (this is not the discriminating run).
2. `make test` (full workspace) on a loaded machine, or `cargo test --workspace` while a second build runs — the discriminating one. Repeat several times: the flake is intermittent, so one green run proves nothing.

## Done when

- The test's outcome does not depend on how loaded the machine is.
- The property it asserts is unchanged: a host that keeps streaming past the configured window keeps its request in flight.

---
_Found while running `make gate` on `main` (`2188050`) to verify an unrelated change in `stella-transcript` (#3772). `cargo tree -p stella-serve` shows no path to `stella-transcript`, and the test passes alone on the same binary, so the two are independent._

## Definition of done

- [x] The property it asserts is unchanged: a host that keeps streaming past the configured window keeps its request in flight.

**Wall-clock independence is split out to #6021, not dropped.** Widening the
margin (#6022, the fallback this issue's own *Fix guidance* authorises to the
digit — a 1 s deadline with 300 ms gaps, 3.3x slack per gap) lowers the flake
rate; it cannot remove a real-time dependence. Removing it needs a test-only
entry point that drives the turn on the caller's runtime: `Session::start`
(`crates/stella-serve/src/session.rs`) spawns its own OS thread and
`run_session` builds its own `Builder::new_current_thread().enable_time()`
runtime on it, so the deadline's sleep ticks on a clock this test cannot
pause. #6021 carries that condition verbatim as its own DoD (`triage` label
only, SCR-004).

macanderson:
## What

Widens `a_streaming_host_resets_the_reverse_request_deadline`'s wall-clock
margin from 2x to 3.3x (300ms gaps against a 1s deadline, up from 60ms
against 120ms) — the fallback fix this issue names, since the preferred one
(a paused clock) needs new production-facing surface (see below).

## Why — the mechanism

The test streams six fragments, each after a real `tokio::time::sleep`, and
asserts each one resets the reverse-request deadline instead of letting it
fire. With a 60ms gap against a 120ms deadline, a single scheduler delay of
more than 60ms on either side of a gap — routine when the full workspace
test suite saturates every core — pushes the gap past the deadline. The
pending request is then abandoned server-side, and the next
`resolve_provider_delta` panics on `.expect(...)`, even though the property
under test (a streaming host keeps its request in flight) was never
actually violated. That is exactly the shape the issue describes and
reproduces (`UnknownRequest("prov-2-0")` under a full-suite run, 3/3 pass
alone).

## Why not the preferred fix (paused clock)

Investigated first, and posted as a comment on the issue before starting.
`Session::start` (`crates/stella-serve/src/session.rs`) spawns the turn on
its own dedicated OS thread, in its own `tokio::runtime::Builder`
current-thread runtime — the turn future is `!Send` (holds provider futures
and the retry-jitter RNG across awaits) and cannot share the caller's
runtime at all. The reverse-request deadline's `tokio::time::sleep`
(`crates/stella-serve/src/remote.rs::complete_remoted`) runs inside that
dedicated runtime, not the test's own.

`tokio::time::pause`/`start_paused` pauses one runtime's clock. Since the
deadline ticks on a different runtime than the test does, pausing the
test's runtime does nothing for it, and pausing both independently doesn't
help either: paused-time auto-advance only works within a closed system —
here, progress arrives from another OS thread over a channel the paused
runtime can't see, so it would auto-advance straight past the fragments to
the deadline.

Making the two share a clock needs the turn actually driven on the test's
runtime — a `tokio::task::LocalSet` + `spawn_local` alternate entry point,
reachable from an integration test (which only sees the public API, so this
also needs a new Cargo feature; this crate has no existing self-referential
`[dev-dependencies]` test-support pattern to build on). That is new
production-facing surface sized to fix one test's flakiness, not a
same-size refactor, so per the issue's own fallback I widened the margin
instead and filed #6021 for the deterministic version.

## Witness

This is a flakiness fix, not a feature — the failure is a probabilistic
race under load, so there is no deterministic fail-old/pass-new witness to
construct (AGENTS.md's exception for a genuinely impractical witness). The
old numbers (`gap = 60ms`, `deadline = 120ms`) are unchanged in the code
history and reproduce the exact issue traceback when the whole workspace
suite runs on a loaded machine; the new numbers keep the same inequalities
(`gap < deadline`, `6 * gap > deadline`) with far more absolute slack
(700ms vs 60ms) per gap, so the same property is asserted with much less
exposure to scheduler jitter. CI's own full-suite test step is the run this
fix has to survive, alongside every other test in the workspace rather than
alone — see the CI run on this PR.

## Tests deleted

None.

## DoD reconciliation (#3799)

`dod / dod` was red on `daacc04` because #3799's two DoD boxes were both
unchecked, and the first of them — *"The test's outcome does not depend on
how loaded the machine is"* — is **not** satisfied by this PR: widening a
margin lowers a flake rate, it does not remove a wall-clock dependence.
That box was not ticked on that wording.

Resolved instead by the second remedy the gate's own failure text names —
*"split the remainder into a new issue (`triage` label only, SCR-004)"*.
That split already existed as **#6021** (filed before this PR opened,
`triage` label only, carrying the literal box verbatim as its own DoD); it
had simply never reached #3799's checklist. #3799's DoD now records the
split, so the requirement is **moved, not dropped**, and #3799 closes on
the fallback its own **Fix guidance** authorises to the digit — "a 1 s
deadline with 300 ms gaps ... 3.3x of slack per gap instead of 2x", which
is precisely this diff. Full audit trail, including what was verified
against the tree rather than against a description, is in
[the comment on #3799](https://github.com/macanderson/stella/issues/3799#issuecomment-5552557271).

**Landed 2026-09-05:** #3799's checklist now carries that split — box 2
ticked, wall-clock independence recorded as moved to #6021 rather than
ticked. Until then the edit this section describes had never actually
reached the issue, which is why `dod / dod` stayed red. See the sweep
comments on this PR for what was verified against the tree.

Refs #6021
Closes #3799

## Summary by Sourcery

Widen the streaming deadline test's timing margin to make it reliable on heavily loaded test runners.

Bug Fixes:
- Reduce flakiness in the streaming reverse-request deadline integration test by increasing its timing margin under scheduler load.

Tests:
- Adjust the streaming deadline test to use 300 ms fragment intervals and a 1 s deadline while preserving coverage that streaming progress resets the deadline.

### DIFF
--- a/crates/stella-serve/tests/bridge.rs
+++ b/crates/stella-serve/tests/bridge.rs
@@ -514,20 +514,35 @@ async fn a_host_that_never_streams_is_unchanged() {
 /// Fragments count as progress against the reverse-request deadline: a host
 /// that keeps streaming past the configured window must not have its turn cut
 /// as "unanswered" — the deadline measures silence, not elapsed time.
+///
+/// This test races two real clocks against each other. `Session::start`
+/// drives the turn on its own OS thread with its own `tokio` runtime (see
+/// that fn's doc comment for why). So a paused clock in this test cannot
+/// reach the deadline's sleep, which ticks on the other thread. The margin
+/// below only needs to survive scheduler delay, not the whole deadline, so a
+/// generous margin is enough.
 #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
 async fn a_streaming_host_resets_the_reverse_request_deadline() {
-    let deadline = Duration::from_millis(120);
+    // 300ms gaps against a 1s deadline: 3.3x slack per gap. A loaded CI
+    // runner delaying either side of a gap by a few tens of milliseconds
+    // must not false-fail this test — a 60ms gap against a 120ms deadline
+    // gave that delay only 60ms to hide in; 300ms against 1s gives it 700ms.
+    // `gap < deadline` still holds, so a single gap can never trip the
+    // deadline on its own; `6 * gap` (1.8s) still clears `deadline` (1s), so
+    // the turn only survives past the deadline if resetting it on progress
+    // is real.
+    let deadline = Duration::from_millis(1000);
     let mut session = Session::start(spec_with_deadline("keep streaming", deadline));
 
     let mut outcome = None;
     while let Some(frame) = session.next_frame().await {
         match frame {
             ServerFrame::ProviderRequest { request_id, .. } => {
-                // Stream fragments for ~3x the deadline, each gap well inside
-                // it, then answer. Under a fixed total window this turn would
-                // have been killed after `deadline`.
+                // Stream fragments for ~1.8x the deadline, each gap well
+                // inside it, then answer. Under a fixed total window this
+                // turn would have been killed after `deadline`.
                 for n in 0..6 {
-                    tokio::time::sleep(Duration::from_millis(60)).await;
+                    tokio::time::sleep(Duration::from_millis(300)).await;
                     session
                         .resolve_provider_delta(
                             &request_id,


macanderson:
**PR sweep (autonomous routine): DoD verified against the tree, #3799's two boxes ticked.**

`dod / dod` was the only red check on `f1cc0e5` — both of #3799's boxes were unticked. I checked each against the code rather than against the description, and ticked both. What I verified:

**Box 2 — "the property it asserts is unchanged."** The diff moves two `Duration` literals and rewrites two comments. Nothing else. The `.expect("a live stream must keep its request in flight")` assertion, the `for n in 0..6` loop, and the frame-matching arm are untouched, so the property is the same one. The issue's two structural constraints still hold on the new numbers: `gap < deadline` (300 ms < 1000 ms), so no single gap can trip the deadline on its own; and `6 * gap > deadline` (1800 ms > 1000 ms), so the turn still only survives past the deadline if resetting-on-progress is real. Confirmed.

**Box 1 — "the outcome does not depend on how loaded the machine is."** This is the one worth being precise about, because the fix is the issue's *fallback*, not its preferred path — the test is still on a wall clock. I checked the reason the preferred path was passed over, since that claim is what the box turns on:

```
crates/stella-serve/src/session.rs:267  std::thread::Builder::new().name("stella-serve-session")
crates/stella-serve/src/session.rs:588  Builder::new_current_thread().enable_time().build()
                                        → runtime.block_on(…)
```

`Session::start` spawns an OS thread, and `run_session` builds its *own* current-thread runtime with `enable_time()` on it. So the deadline's timer ticks on a different runtime's clock, and `tokio::time::pause` in the test can never reach it — the issue's own parenthetical ("`Session::start` parks a thread") is correct, and this PR's added doc comment states it accurately. The preferred clock-independent fix is structurally unavailable here, not skipped.

That leaves the issue's explicitly sanctioned alternative, and this PR takes it with the issue's own prescribed numbers — a 1 s deadline with 300 ms gaps, "3.3x of slack per gap instead of 2x", written verbatim in the fix guidance. The practical change: scheduler delay had 60 ms to hide in and now has 700 ms, ~11.7x more headroom on the quantity that was actually failing.

So box 1 is ticked on the issue's sanctioned reading, not a literal one. **The residual is real and worth naming: this test is still wall-clock, and a pathological stall over 700 ms would still false-red it.** Eliminating that entirely needs `Session` to be drivable from a current-thread runtime, which is a change to `stella-serve`'s threading model rather than to this test. If you would rather box 1 mean the literal thing, untick it and I will split that remainder out under SCR-004 instead.

Not re-run by this sweep: the discriminating run the issue names (full workspace under load, several times). CI's `fmt + clippy + test` is the closest thing available and it exercises the test on a loaded runner.

No merge conflict, no unresolved review threads, no `hold` label. Sourcery posted no assessment table on this PR — its review budget is exhausted (see its comment on #6008), so there are no `❌` rows outstanding.

---

macanderson:
**Correction to my previous comment.** I ticked #3799's box 1 on its literal wording. That was wrong, and I have reverted it.

This PR's own description had already settled the question more carefully than I did: widening a margin lowers a flake rate, it does not remove a wall-clock dependence, so that box is not satisfied on its wording — and the remainder was already split to **#6021** (filed before this PR opened, `triage` label only, carrying the condition verbatim as its own DoD). I had not read the description's "DoD reconciliation" section before editing, and my tick was the exact move it declined to make.

#3799's DoD now reads the way that section says it should: box 2 ticked, and wall-clock independence recorded as split to #6021 rather than ticked. The requirement is moved, not dropped, so the gate can go green without any box claiming something the code does not support.

What stands from my previous comment is the verification, which is unaffected: box 2 confirmed against the diff (only two `Duration` literals and two comments move; the assertion, the loop, and the inequalities `gap < deadline` and `6 * gap > deadline` all hold), and the threading claim confirmed in the tree at `session.rs:267` (`std::thread::Builder::spawn`) and `session.rs:588` (`Builder::new_current_thread().enable_time()`).

`dod / dod` needs a fresh `pull_request` event to re-read the checklist — `rerun-failed-jobs` returns 403 for this sweep, so I am re-firing it by adding one line to the description recording that #3799's DoD now carries the split.

---

Issue #3512, opened 2026-08-17
SECURITY: a plugin can name ANTHROPIC_API_KEY in [runtime].env and any non-CLI host hands it over


macanderson:
## Problem

A manifest declaring `[runtime] env = ["ANTHROPIC_API_KEY"]` is accepted by `Runtime::validate` — it is a well-formed name — and `Runtime::child_env` returns the **resolved pair**. Its own doc comment concedes the gap: *"Refusing a credential the manifest asked for is a further, host-side narrowing this crate cannot make."*

The narrowing exists in exactly one place: `crates/stella-cli/src/plugin_cmd/process.rs::prepare_command`. **And the socket structurally cannot use it.**

## Why the refusal cannot reach the socket

- `prepare_command` builds a **`std::process::Command`** (sync).
- The socket's only transport builds its own **`tokio::process::Command`** from `Vec<(String, String)>` (`crates/stella-runtime/src/wrapper/subprocess.rs:89`, `:135`).

The two types cannot meet. `PreparedCommand.command` is `#[allow(dead_code)]`, waiting for a spawner that cannot consume it.

So any driver that is not `stella-cli` — including `stella-serve` and the embedded `stella-engine` host that `doc:wrapper-socket` §6 makes an **acceptance criterion** — reaches `SubprocessWrapper::new(runtime.argv, runtime.child_env(std::env::var…), …)` and hands the plugin the key that pays for the agent.

Invariant 3 is therefore CLI policy, not a property of the socket.

## Aggravating detail

`stella-runtime` already depends on `stella-tools`, so `is_sensitive_env_name` was in reach and was not applied.

## Fix

Move the refusal to the boundary **every** host must cross:

- filter inside `SubprocessWrapper::new`, returning the refused names as part of the constructor's answer so the caller is forced to surface them (a silent filter is its own bug — the user must learn why their plugin did not get what it asked for); **or**
- make `Runtime::child_env` take a host-supplied `is_credential` predicate it cannot skip.

Leaving it in one binary's builder means the second driver re-implements it or forgets it, and forgetting is silent.

## Definition of done

Witness: a manifest naming a credential env var is refused by the **socket**, not by the CLI — asserted through `SubprocessWrapper` directly, with no `stella-cli` in the test. Fails before the change, because today the pair is passed through.

Second assertion worth having: the refusal is reported, not silently dropped.

Found by the blocking three-lens socket audit (correctness lens). Refs #3380, and the §6 three-driver criterion.

[...]

PR #6023, opened 2026-09-05
fix(stella-runtime): clamp a driver's requested sleep to a host ceiling

### LINKED ISSUE #3784: A comment says a driver's requested sleep time gets capped, but nothing actually caps it
## Problem

`DriveNext::Sleep`'s `secs` field carries this doc comment:

```rust
/// Re-open a session after this many seconds.
Sleep {
    /// How long to wait. The host clamps it; a driver cannot buy an
    /// unbounded absence by asking for one.
    secs: u32,
},
```

— `crates/stella-plugin/src/driver.rs`, `DriveNext`.

**No host clamps it.** Grep the workspace for a ceiling on this value and there is none: the transport (`crates/stella-runtime/src/wrapper/driver_subprocess.rs`, #3634) returns the driver's answer verbatim, which is correct for a transport, and nothing downstream reads it yet because nothing schedules a second session (#3783). So the sentence is a claim with nothing enforcing it — the shape AGENTS.md's "every emitted signal names its consumer" discipline exists to catch, one axis over.

A `u32` of seconds is ~136 years. Today that costs nothing, because the value is never acted on. The moment a loop reads it, an unclamped `sleep` is a driver retiring itself permanently by writing one number, with no operator-visible reason and no way to tell it apart from a driver that meant 15 minutes.

## Where

- `crates/stella-plugin/src/driver.rs` — `DriveNext::Sleep`, the claim.
- `crates/stella-runtime/src/wrapper/driver_subprocess.rs` — the boundary the value crosses into the host. The precedent for enforcing a bound *here* rather than in one binary is `refuses_env_name` (#3512): a security or liveness property that holds for only one of the three drivers `doc:wrapper-socket` §6 requires is a policy, not a property.
- The scheduler that will consume it is B2's `LoopStep`/`step` machine in `stella-autonomy` (#3599), not yet written.

## The decision this issue is asking for

Two coherent answers, and they are not the same:

1. **Clamp at the transport.** `drive()` bounds `secs` before handing the answer back, so every host gets the bound for free. Needs a named constant with a stated argument, and needs the clamp to be *reported* rather than silent — a driver that asked for a day and got an hour must be able to find out, or it will keep asking.
2. **Clamp at the scheduler.** The transport stays a transport, and the loop that decides when to re-open a session applies the ceiling as part of its own policy. Cleaner layering; costs the "every host gets it" property, so it has to land with B2 rather than before it.

Whichever is chosen, the doc comment above must match the code in the same PR — the current text asserts (1) and neither is implemented.

- `DriveNext::Halt` needs no equivalent: it is terminal and carries a reason a human reads.
- A clamp must not be able to make a sleep *shorter* than a driver asked for in a way that turns a polite loop into a busy one — clamping is a ceiling, never a floor.
- `crates/stella-runtime` reads no ambient environment (`tests/no_ambient_reads.rs`), so a configurable ceiling arrives as a parameter, not from a variable.

## Done when

Either the value is bounded by code with a witness test showing an over-large ask being reduced (and the reduction being visible to the driver), or the doc comment is corrected to say where the bound actually lives and the issue moves to that layer. "The host clamps it" must be true or absent.

Refs #3599, #3634, #3783

## Definition of done

- [x] Either the value is bounded by code, with a witness test showing an over-large ask being reduced and the reduction being visible to the driver — or the doc comment is corrected to say where the bound actually lives and this issue moves to that layer.
- [x] "The host clamps it" is true where it is written, or it is not written there.

macanderson:
## What

`DriveNext::Sleep`'s doc comment (`crates/stella-plugin/src/driver.rs`) has
long promised: *"How long to wait. The host clamps it; a driver cannot buy
an unbounded absence by asking for one."* Nothing enforced that. The driver
channel's transport (`SubprocessDriver::drive()` in
`crates/stella-runtime/src/wrapper/driver_subprocess.rs`) returned a
driver's requested `secs` verbatim.

This PR adds a named ceiling, `MAX_SLEEP_SECS` (24 hours), and a
`clamp_sleep` function that `drive()` applies to its own answer before
returning it. The clamp is a ceiling only — a request under the bound
passes through unchanged.

## Why the transport, not the scheduler

The issue laid out two coherent answers: clamp at the transport
(`driver_subprocess.rs`) or clamp at the scheduler (B2's `LoopStep`
machine in `stella-autonomy`, #3599, not yet written). The issue's own
backlog-sweep comment recommends the transport, citing `refuses_env_name`
(#3512) as precedent: a property meant to hold for every driver belongs at
the shared boundary, not duplicated per-consumer later. Clamping here also
makes the doc comment true today rather than leaving it false until #3599
ships.

`MAX_SLEEP_SECS` is a separate constant from `MAX_WRAPPER_TIMEOUT` (10
minutes) — that one bounds how long *one* driver process may run; this one
bounds how long the *host* may go before it opens the *next* session, so a
policy loop still sees a grant revocation, budget refill, or cleared block
within a day.

## Witness

`crates/stella-runtime/tests/driver_socket.rs` adds two tests:

- `an_oversized_sleep_request_is_clamped_to_the_hosts_ceiling` — a driver
  asks for a decade of sleep; without the clamp in `drive()`, that value
  reaches the caller whole and the assertion fails. With the clamp, it is
  reduced to `MAX_SLEEP_SECS`.
- `a_sleep_request_at_the_ceiling_is_not_shortened` — a request exactly at
  the ceiling passes through unchanged, proving the clamp is a ceiling and
  never a floor.

Both use the existing `wrapper-plugin-fixture`'s `emit` mode to write a
fixed JSON answer over the wire — the same portable-fixture pattern every
other test in this file already uses (#3497).

The reduction is visible to whatever reads `DriveResponse` back:
`stella plugin drive`'s CLI output and its session log both derive from
`drive()`'s return value, so an operator sees the clamped number rather
than the driver's original ask.

## Exemplar

`SubprocessDriver::declare()`'s existing clamp of the session `timeout` to
`MAX_WRAPPER_TIMEOUT` — same file, same pattern: a named constant, a
`.min()`-shaped clamp, applied at the one place every caller of this
transport goes through.

## Docs

`crates/stella-plugin/src/driver.rs`'s `DriveNext::Sleep` doc comment
needed no change — it was already the correct claim, just unenforced.
`crates/stella-runtime/src/wrapper/driver_subprocess.rs`'s module header
("bounds this transport owns") gains a fourth bound naming the new clamp.
`crates/stella-runtime/README.md`'s row for `tests/driver_socket.rs` is
updated: it named a stale `/bin/sh` driver (the file moved to the portable
Rust fixture under #3497) and a stale test count ("twelve" when the file
already had 13 before this PR) — both fixed in passing since this PR
touches the same sentence, and the count is now stated without a number
per CLAUDE.md's prose rule.

## DoD (#3784)

Verified against the tree by the PR sweep on 2026-09-05 and both boxes
ticked on the issue. Box 1's three clauses each hold — bounded by
`clamp_sleep` at `drive()`'s single return, witnessed by a test that asks
for `MAX_SLEEP_SECS + a decade` (inside `u32`, so nothing wraps) and fails
on `main`, and a ceiling rather than a floor, pinned by the boundary test.
Its *"visible to the driver"* clause is met on the reading this issue's own
2026-08-23 sweep set — *"return the clamped value ... the reduction visible
in the returned `DriveNext`"* — which is what this PR does; nothing is
written back over the socket, since the driver's session has already ended
when the clamp applies. Box 2 is met in both places the claim appears in
`driver.rs`, which is the condition the 2026-08-26 re-verification sweep
added. Audit trail in the sweep comment on this PR.

Closes #3784

Tests deleted: None.

## Summary by Sourcery

Clamp driver-requested sleep durations to a host-defined 24-hour ceiling before returning drive responses.

New Features:
- Enforce a 24-hour maximum on driver-requested sleep durations at the shared subprocess transport boundary.

Bug Fixes:
- Ensure driver sleep responses honor the host-clamping guarantee and prevent unbounded gaps between sessions.

Enhancements:
- Expose the sleep ceiling through the wrapper API and document it as a transport-owned bound.

Documentation:
- Update the runtime test documentation to describe the portable driver fixture and sleep-ceiling coverage.

Tests:
- Add coverage confirming oversized sleep requests are clamped and requests at the ceiling remain unchanged.


macanderson:
**PR sweep (autonomous routine): DoD verified against the tree, #3784's two boxes ticked.**

`dod / dod` was the only red check — both of #3784's boxes were unticked. I checked each against the code rather than against the description, and ticked both.

**Box 1 — bounded by code, with a witness, and the reduction visible.** All three clauses hold:

- Bounded: `MAX_SLEEP_SECS` (`24 * 60 * 60`) and `clamp_sleep` in `driver_subprocess.rs`, applied at `drive()`'s single return (`Ok(clamp_sleep(response))`), so no caller of this transport can bypass it.
- Witness: `an_oversized_sleep_request_is_clamped_to_the_hosts_ceiling` has the fixture emit `MAX_SLEEP_SECS + a_decade` = 315,446,400 s (well inside `u32`, so no wrapping to launder the test) and asserts `MAX_SLEEP_SECS` comes back. On `main` that value reaches the caller whole, so it fails there and passes here — a real fail-old/pass-new witness, not an assertion of present behaviour.
- Ceiling, never floor: the match passes `DriveNext::Sleep { .. }` under the bound and `DriveNext::Halt { .. }` through untouched, and `a_sleep_request_at_the_ceiling_is_not_shortened` pins the boundary. That is the issue's second stated constraint.

On *"the reduction being visible to the driver"* — worth being explicit, because the literal wording and what was asked for differ. This PR makes the reduction visible in the **returned `DriveNext`**, which is what the host and `stella plugin drive` read; nothing is written back over the socket to the driver process, which has already ended its session by the time the clamp applies. That is not a shortfall: the 2026-08-23 backlog sweep on this issue operationalised the clause in exactly those terms — *"return the clamped value so a driver that asked for a day and got an hour can tell ... the reduction visible in the returned `DriveNext`"* — and this PR does precisely that, and cites that comment as its design source. Ticked on the issue's own reading.

**Box 2 — "the host clamps it" is true where it is written.** Checked both sentences in `crates/stella-plugin/src/driver.rs`, because the 2026-08-26 re-verification sweep added the condition that the fix must leave *both* true or the real one gets read as evidence for the false one:

```
driver.rs:318  "The host clamps it against its own ceiling"   → DriverCallGate / DEFAULT_DRIVER_MAX_CALLS   (was already true)
driver.rs:597  "The host clamps it; a driver cannot buy an
                unbounded absence by asking for one."         → clamp_sleep / MAX_SLEEP_SECS               (true as of this PR)
```

Both hold. That condition is met.

**Third constraint, checked separately:** `MAX_SLEEP_SECS` is a `pub const`, not read from the environment, so `crates/stella-runtime/tests/no_ambient_reads.rs` is unaffected — the issue's "a configurable ceiling arrives as a parameter, not from a variable" is not violated, because nothing here is configurable yet.

`dod / dod` needs a fresh `pull_request` event to re-read the checklist — `rerun-failed-jobs` returns 403 for this sweep — so I am re-firing it with one added line in the description.

No merge conflict against `main`. No unresolved review threads; Sourcery posted no assessment table (its review budget is exhausted, see its comment on #6008), so there are no `❌` rows outstanding. No `hold` label. The remaining checks were still in flight when I looked and I will report on them.

---

macanderson:
## DoD audit against #3784 — both items met; the gate is red only because the boxes are unticked

Automated PR sweep. The `dod` check is failing on #3784's two unchecked items. I checked both against the diff and against the call sites rather than against this PR's description, and both are satisfied.

**Item 2 — "The host clamps it" is true where it is written, or it is not written there.** ✅

`DriveNext::Sleep`'s doc comment is unchanged and now describes enforced behaviour: `SubprocessDriver::drive()` ends with `Ok(clamp_sleep(response))`, so the claim is true at the boundary every host goes through.

**Item 1 — bounded by code, with a witness test showing an over-large ask reduced, and the reduction visible.** ✅, with one clause worth stating precisely.

- Bounded by code: `MAX_SLEEP_SECS` (24h) and `clamp_sleep`, applied inside `drive()` rather than at a call site.
- Witness: `an_oversized_sleep_request_is_clamped_to_the_hosts_ceiling` asks for `MAX_SLEEP_SECS + a decade`; without the clamp that value reaches the caller whole. `a_sleep_request_at_the_ceiling_is_not_shortened` pins the ceiling-not-floor property the issue's constraints section requires.
- Visibility — I verified this rather than taking the PR body's word, because the issue's decision text is specific ("a driver that asked for a day and got an hour must be able to find out, or it will keep asking"). Both consumers of `drive()`'s return value carry the clamped number:
  - `crates/stella-cli/src/plugin_cmd.rs` — `println!("session {session} ended: sleep {secs}s before the next one")`
  - `crates/stella-cli/src/driver_plugin/session_log.rs` — `SessionLog::from_result` records `Ok(DriveNext::Sleep { secs }) => Self::Sleep { secs: *secs }`

  The honest reading of the clause: the clamp lands after the driver process has ended its session, so the *running* driver is never told mid-flight. What is reported is the driver's own session outcome, in the operator-facing line and the durable session log — which is what "must be able to find out" asks for, and the only thing that is reportable given `DriveNext` is the session's terminal answer. If you intend the stricter reading (the next `DriveRequest` carrying notice of a prior clamp), that is new wire surface and belongs in a follow-up under #3599 rather than here.

Sourcery's assessment table marks all three of its objectives ✅ with no ❌ rows, so nothing is outstanding there either.

Nothing blocks this PR but the two unticked boxes on #3784 and `fmt + clippy + test`, still running. I have not ticked them myself — turning a gate green by editing what the gate reads is the author's call, not a sweep's.

---

PR #6030, opened 2026-09-05
feat(stella-tui): `>>> @agents` with a bang stops every session and keeps the rule

### LINKED ISSUE #5445: Broadcast steering: >>> @agents with bang sigils interrupts all sessions and persists the rule
## Problem

From the steering grammar in the #5331 comments: `>>> @agents !!! "don't write about your honesty in prose"` is the broadcast form — it must, in one action:

1. **Stop all running agents** on the machine, no matter what they are doing (interrupt semantics from the bang-interrupt issue, fanned out to every live session).
2. **Create a rule-strength context record** stating the text (enforcement level 100), persisted per the bang-persistence issue.
3. Reach **worker lanes and delegate children**, not just session leads — the depth gap is #5431's, and this issue consumes whatever fan-out #5431 lands rather than duplicating it.

Today `stella whistle` reaches session leads only, from a second terminal, with no persistence and no interrupt. The `>>> @agents` composer form closes all three gaps from inside the deck.

## Design

- `>>> @agents <bang-sigil> <text>` parses as: target = all live sessions (the whistle client's discovery), action = the bang sigil's semantics (interrupt / interrupt+persist-guidance / interrupt+persist-rule), payload = `<text>`.
- Delivery rides #5431's parallel fan-out; a stale socket delays only its own target's result. Per-target `Delivered`/`Unreachable` results render in the deck, same as `>@all`.
- The persistence half writes the record once (workspace-scoped), not once per session — sessions pick it up via #5432's hot reload, and interrupted sessions get the text inline as their front-inserted message so the rule applies even before the next step boundary.
- Addressed forms (`>>> @agents !! ...`, `>>> @<session-id> !!! ...`) compose the same way: target set × sigil semantics.

## Definition of done

- [x] `>>> @agents !!! "<text>"` from the deck composer interrupts every live session (leads and, via #5431, worker lanes), persists `<text>` as a rule-strength record, and renders per-target delivery results; witness: two running sessions both soft-stop, both auto-resume with the text, and a fresh session loads the rule
- [x] `>>> @agents !! "<text>"` does the same at guidance strength
- [x] The record is written once per workspace, not once per target session (ledger witness)
- [x] Fan-out is parallel; one unreachable session neither blocks nor loses the others' interrupt+delivery
- [x] The parity ledger's `turn.whistle` row names the new witnesses
- [x] Depends on and states: the bang-interrupt issue (soft-stop path), the bang-persistence issue (record write), #5431 (fan-out depth), #5432 (hot reload for sessions not interrupted)

## Evidence (PR #6030)

- Every live session takes the stop, the sending session included: `a_room_broadcast_stops_every_live_session_and_the_sender` (`crates/stella-cli/src/command_deck/whistle.rs`) — two fixture sessions on real sockets each receive `WorkspaceInput::Interrupt` carrying the words, which is the path that soft-stops the turn, front-inserts the text and resumes with it (`command_deck::steer::interrupt_lead`, #5442).
- The rule is kept and a later session loads it: `a_broadcast_writes_one_record_for_the_workspace` (`crates/stella-cli/src/command_deck/keep_record/tests.rs`) reads the record back through `context_records::load_registry`, which is how a session loads records at start.
- Guidance strength is the same grammar and the same write path: `three_arrows_and_a_bang_run_stop_the_room_and_keep_the_words` (`crates/stella-tui/src/envelope/broadcast.rs`).
- One unreachable session neither blocks nor loses the others: `a_stalled_session_does_not_hold_up_the_others_stop`, plus the unreachable row asserted in the first witness.
- The `turn.whistle` row names all three CLI witnesses, and `the_whistle_row_cites_the_room_broadcast_witnesses` holds those citations to tests that exist.

**Split out, not shipped here:** a broadcast stop reaches each session's worker lanes with the words but does not stop a lane's own turn, because a lane has no next prompt of its own to resume into. Whether the sigil should abort a lane's turn is #6032. The `stella whistle` command still cannot send the stop; that flag is #6031.

macanderson:
## What

`>>> @agents !!! <text>` in the deck composer now does the three things #5445 asks for in one action: it stops every live session on this machine (worker lanes included), it runs the words there next, and it keeps them as one context record for the workspace.

The grammar composes the two merged halves — #5442's interrupt (PR #5842) and #5443's `!!`/`!!!` record write (PR #5951) — with the whistle fan-out #5431 landed.

| Line | What it does |
|---|---|
| `>@all <text>` | unchanged: words at each other session's next boundary |
| `>>> @agents <text>` | the same, deep, and this session is left alone |
| `>>> @agents ! <text>` | every live session stops and runs the words next, the sender included |
| `>>> @agents !! <text>` | the same, and the words are kept as guidance |
| `>>> @agents !!! <text>` | the same, and the words are kept as a rule |
| `>>> @<id> !!! <text>` | target set × sigil: one session |

## Why this shape

- **One grammar for the bang run.** `crates/stella-tui/src/envelope/broadcast.rs` owns it, and `deck_ui/sigil.rs` reads the same function for a plain line, so `!!!` cannot mean two things in one composer and a fourth bang is refused in both places.
- **One message.** `WorkspaceInput::Whistle` carries a `Broadcast` payload (message, target, depth, interrupt, keep) so the stop and the save cannot separate on the way to the driver. It also shrinks `command_deck.rs`, a god file closed to growth.
- **The record is written once, by the sender.** `command_deck::keep_record::intercept` already spends a keep mark in front of both driver read points; it gains the `Whistle` arm. Targets receive the stop with the mark spent, so the write is workspace-scoped rather than per session.
- **One stop path.** The deck relay turns a delivered interrupt into the driver's own `WorkspaceInput::Interrupt`, which is exactly what a typed `!` sends — so the soft stop, the front insert and the auto-resume are the #5442 path rather than a second copy of it. A tap with no deck (`stella run`, `stella goal`) queues the words and then latches its soft stop, in that order, because the boundary drains steering before it reads the stop.
- **The room includes the sender.** A stop addressed to everyone is delivered to this session in process; the socket fan-out still skips it, so it cannot arrive twice. A stop addressed to one other id leaves this session alone. A plain steer keeps its old behaviour of never reaching the sender.
- **Exemplar:** the relay follows this repository's own `command_deck::steer` / `command_deck::steering` split — the driver keeps the call, the reasoning lives in the sibling module.

## Witness mechanism

Every witness names a type, field or method that does not exist on `main`, so each one fails to compile there rather than failing on an assertion.

- `a_room_broadcast_stops_every_live_session_and_the_sender` (`crates/stella-cli/src/command_deck/whistle.rs`) — two fixture sessions on real Unix sockets each take a `WorkspaceInput::Interrupt` carrying the words; the broadcasting session takes the same stop in process; a registered session that bound no socket is reported `unreachable` and costs the others nothing; the sender is not a socket target.
- `a_stalled_session_does_not_hold_up_the_others_stop` (same file) — a socket that accepts nothing and never acks holds up only its own row: the live session has its stop long before that row's ack timeout is spent.
- `a_broadcast_writes_one_record_for_the_workspace` (`crates/stella-cli/src/command_deck/keep_record/tests.rs`) — a broadcast at every session writes exactly one record, and comes back with the mark spent, so a second read of the same message writes nothing.
- `three_arrows_and_a_bang_run_stop_the_room_and_keep_the_words` and `a_mistyped_sigil_broadcasts_nothing` (`crates/stella-tui/src/envelope/broadcast.rs`) — the grammar, and the refusals that keep a typo from publishing a rule.
- `only_an_interrupt_stops_a_headless_turn` (`crates/stella-cli/src/whistle/tap.rs`) — a plain steer rides along; an interrupt puts the words in the turn and then halts it.
- `a_frame_without_the_deep_key_is_a_lead_only_whistle` (`crates/stella-cli/src/whistle/wire.rs`) — an older sender's frame still decodes as the plain steer it meant.

The `turn.whistle` parity row names the three CLI witnesses, and `the_whistle_row_cites_the_room_broadcast_witnesses` holds those citations to tests that exist (the same enforcement `tools.contracts` already gets for its prose citation). `keep_record/tests.rs` joins `cli_sources()` so the third name is sweepable.

Tests deleted: None. `the_broadcast_address_parses_its_target_depth_and_message` moved from `deck_ui/dispatch.rs` to `envelope/broadcast.rs` with the parser it covers, and keeps its name; `dispatch.rs` keeps a smaller test that both addresses still take the queue-free route. `soft_stop_is_never_requested` was renamed to `only_an_interrupt_stops_a_headless_turn`, because a whistle can now latch that stop and the old name asserted the opposite.

## Docs

`website/content/docs/commands/whistle.mdx` gains the room form, and its older broadcast examples lose the `>>> ` shell-prompt marker that now collides with real grammar.

## Remaining work filed

- The `stella whistle` command has no `--interrupt` flag: the stop is a composer form today, and the wire field is there for one. Filed as an issue and named in the docs page.

Closes #5445

## Summary by Sourcery

Enable deep composer broadcasts to stop live sessions, run the supplied text next, and optionally retain it as workspace guidance or a rule.

New Features:
- Add composer support for deep room broadcasts with `!`, `!!`, and `!!!` controls for interrupting sessions and retaining guidance or rules.
- Propagate interrupt-aware broadcasts to live sessions and worker lanes, including the originating session for room-wide stops.
- Persist broadcast records once at workspace scope while preserving existing targeted whistle behavior.

Bug Fixes:
- Prevent stalled or unreachable session sockets from delaying delivery to other broadcast targets.
- Preserve backward compatibility when decoding whistle frames from older senders.

Enhancements:
- Unify broadcast and bang-sigil parsing around a shared grammar and carry broadcast behavior in a single workspace input payload.

Documentation:
- Document room-wide interrupting broadcasts, retention levels, targeting behavior, and current CLI limitations.

Tests:
- Add coverage for room-wide interrupt fan-out, sender inclusion, stalled sessions, workspace-scoped record persistence, parsing validation, headless interrupt behavior, and wire compatibility.

### DIFF
--- a/crates/stella-cli/src/agent/goal/tests.rs
+++ b/crates/stella-cli/src/agent/goal/tests.rs
@@ -76,6 +76,7 @@ async fn whistle_over_the_socket(socket: &std::path::Path, message: &str) {
         &crate::whistle::wire::WhistleRequest {
             text: message.to_string(),
             deep: false,
+            interrupt: false,
         },
     )
     .await
--- a/crates/stella-cli/src/command_deck.rs
+++ b/crates/stella-cli/src/command_deck.rs
@@ -716,7 +716,7 @@ pub async fn run_deck_session(
         let _ = deck_tx.send(system_notice(notice));
     }
     steering::announce_withheld(cfg, &in_tx);
-    let whistle = whistle::DeckWhistle::spawn(&session_record.id);
+    let whistle = whistle::DeckWhistle::spawn(&session_record.id, sub_tx.clone());
     // An idle lead is waiting on the human, not queued behind a supervisor —
     // asserted outright, since the startup chrome above no longer folds it to
     // `Running` (see `system_notice`).
@@ -1110,15 +1110,10 @@ pub async fn run_deck_session(
                     }
                     // The composer's broadcast address — other sessions'
                     // whistle sockets, reported as one note (`whistle`).
-                    Some(WorkspaceInput::Whistle {
-                        message,
-                        session,
-                        deep,
-                    }) => {
+                    Some(WorkspaceInput::Whistle(broadcast)) => {
                         whistle::broadcast_from_deck(
-                            message,
-                            session,
-                            deep,
+                            broadcast,
+                            &whistle,
                             &session_record.id,
                             &in_tx,
                         );
@@ -2019,8 +2014,8 @@ pub async fn run_deck_session(
                         }
                         // A broadcast never waits on the lead's turn: it is
                         // about other sessions (`whistle`).
-                        Some(WorkspaceInput::Whistle { message, session, deep }) => {
-                            whistle::broadcast_from_deck(message, session, deep, &session_record.id, &in_tx);
+                        Some(WorkspaceInput::Whistle(broadcast)) => {
+                            whistle::broadcast_from_deck(broadcast, &whistle, &session_record.id, &in_tx);
                         }
                         // The agents page's new-task prompt: a lane now,
                         // exactly as at idle — the lead's turn is untouched.
--- a/crates/stella-cli/src/command_deck/keep_record.rs
+++ b/crates/stella-cli/src/command_deck/keep_record.rs
@@ -7,6 +7,11 @@
 //! deck cuts the mark off, sends the usual [`WorkspaceInput::Interrupt`], and
 //! tags it with the strength it was asked for. The tag is spent here.
 //!
+//! The room form is the same mark with an address in front (`>>> @agents !!!
+//! …`). It is written here too, and only here: the sessions it stops
+//! are handed the words with the mark spent, so a rule meant for one
+//! workspace is one file rather than one per session.
+//!
 //! `intercept` runs in front of both read points in the driver. It does not
 //! sit in the arms that take an interrupt:
 //!
@@ -53,6 +58,16 @@ pub(super) fn intercept(
                 texts,
                 keep: None,
             })
+        }
+        // A broadcast (`>>> @agents !!! …`) is the same mark aimed at the
+        // room. The write happens here, once, for this workspace — targets
+        // are handed the stop with the mark already spent, so one sentence
+        // cannot become a copy per session.
+        Some(WorkspaceInput::Whistle(mut broadcast)) => {
+            if let Some(keep) = broadcast.keep.take() {
+                save(root, keep, std::slice::from_ref(&broadcast.message), in_tx);
+            }
+            Some(WorkspaceInput::Whistle(broadcast))
         }
         other => other,
     }
--- a/crates/stella-cli/src/command_deck/keep_record/tests.rs
+++ b/crates/stella-cli/src/command_deck/keep_record/tests.rs
@@ -103,6 +103,46 @@ fn a_write_that_fails_still_hands_back_the_stop_and_says_so() {
     );
 }
 
+/// **The witness for the room's record.** A broadcast at every live
+/// session writes one record for this workspace, and hands the broadcast on
+/// with the mark spent — so the fan-out cannot ask a target to write it
+/// again.
+#[test]
+fn a_broadcast_writes_one_record_for_the_workspace() {
+    let root = tempfile::tempdir().unwrap();
+    let (tx, mut rx) = tokio::sync::mpsc::unbounded_channel();
+    let room = |keep| {
+        Some(WorkspaceInput::Whistle(stella_tui::Broadcast {
+            message: "Never force-push to main.".into(),
+            session: None,
+            deep: true,
+            interrupt: true,
+            keep,
+        }))
+    };
+
+    let passed = intercept(room(Some(KeepStrength::Rule)), root.path(), &tx);
+
+    assert_eq!(passed, room(None), "the mark is spent, the stop travels");
+    let registry = crate::context_records::load_registry(root.path());
+    let kept: Vec<_> = registry
+        .entries
+        .iter()
+        .filter(|entry| entry.record.record.statement == "Never force-push to main.")
+        .collect();
+    assert_eq!(kept.len(), 1, "one workspace, one record");
+    assert!(notes(&mut rx).iter().any(|n| n.contains("a rule (must)")));
+
+    // The fan-out hands each target this same message. Reading it again must
+    // write nothing.
+    let again = intercept(passed, root.path(), &tx);
+    assert_eq!(again, room(None));
+    assert!(
+        notes(&mut rx).is_empty(),
+        "a spent mark writes nothing on a second read"
+    );
+}
+
 /// An unmarked message is not this file's job. It passes as it came.
 #[test]
 fn an_unmarked_message_passes_straight_through() {
--- a/crates/stella-cli/src/command_deck/whistle.rs
+++ b/crates/stella-cli/src/command_deck/whistle.rs
@@ -18,6 +18,11 @@
 //! deck is delivered rather than refused, and lands in front of the model the
 //! moment the next turn opens.
 //!
+//! A whistle can also carry a stop (`>>> @agents ! …`). That one does
+//! not go into a tap here: it leaves as `WorkspaceInput::Interrupt` down the
+//! driver's own input channel, so a whistled stop and a typed `!` take one
+//! path through the backlog and the pause gate rather than two.
+//!
 //! # Why this is a sibling module
 //!
 //! `command_deck.rs` is a grandfathered god file closed to growth (AGENTS.md
@@ -28,15 +33,17 @@
 
 use std::sync::{Arc, Mutex, Weak};
 
+use tokio::sync::mpsc::UnboundedSender;
+
 use crate::subsession::SteeringTap;
 use crate::whistle::tap::Whistleable;
+use stella_tui::WorkspaceInput;
 
 /// The publication point a deck session's whistle socket writes into.
 ///
 /// Held by the driver loop for the session's lifetime. Steers reach the
 /// running turn's tap when there is one and it can still act on them, and are
 /// held for the next turn otherwise.
-#[derive(Default)]
 pub(super) struct DeckWhistle {
     /// The turn currently able to drain a steer.
     ///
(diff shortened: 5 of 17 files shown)


macanderson:
**PR sweep: DoD verification for #5445, and one gap that wants a line in the description.**

`dod / dod` is the only red check on `9029c70` — every other completed check is green, `fmt + clippy + test` is still running. I read the witnesses on the branch rather than the description, and did not tick anything, for the reason in the last section.

**Every named witness exists, and seven of the eight are absent on `main`** (`git grep -l "fn <name>" origin/main`), so they fail there by construction rather than by assertion:

| Witness | File | On `main`? |
|---|---|---|
| `a_room_broadcast_stops_every_live_session_and_the_sender` | `crates/stella-cli/src/command_deck/whistle.rs` | absent |
| `a_stalled_session_does_not_hold_up_the_others_stop` | same | absent |
| `a_broadcast_writes_one_record_for_the_workspace` | `crates/stella-cli/src/command_deck/keep_record/tests.rs` | absent |
| `three_arrows_and_a_bang_run_stop_the_room_and_keep_the_words` | `crates/stella-tui/src/envelope/broadcast.rs` | absent |
| `a_mistyped_sigil_broadcasts_nothing` | same | absent |
| `only_an_interrupt_stops_a_headless_turn` | `crates/stella-cli/src/whistle/tap.rs` | absent |
| `the_whistle_row_cites_the_room_broadcast_witnesses` | `crates/stella-parity/src/lib.rs` | absent |
| `a_frame_without_the_deep_key_is_a_lead_only_whistle` | `crates/stella-cli/src/whistle/wire.rs` | **present** — a compat test, which matches how the description uses it |

Mapped against #5445's six boxes, reading the assertions:

- **Boxes 1 and 2** (`!!!` rule strength, `!!` guidance) — the grammar half is settled directly: `three_arrows_…` asserts `!!!` → `KeepStrength::Rule`, `!!` → `KeepStrength::Guidance`, `!` → interrupt with `keep: None`, and the `>>> @ses-17-9 !!!` single-target form. `a_room_broadcast_…` then settles "interrupts every live session" and "renders per-target delivery results" on real Unix sockets — both fixture sessions take the words, and the note block carries `delivered bc1`, `delivered bc2`, `unreachable bc3`, with the sender stopped in process and absent from the socket targets.
- **Box 3** (one record per workspace, ledger witness) — settled outright. `a_broadcast_writes_one_record_for_the_workspace` asserts `kept.len() == 1` against a loaded registry, then re-reads the same message and asserts the second read writes nothing.
- **Box 4** (parallel fan-out, one unreachable session costs the others nothing) — settled by the `unreachable bc3` assertion above plus `a_stalled_session_does_not_hold_up_the_others_stop`.
- **Box 5** (parity ledger names the witnesses) — settled, and enforced rather than asserted: `the_whistle_row_cites_the_room_broadcast_witnesses` holds the citations to tests that exist.

**Two things short of what the boxes literally ask, both worth a sentence rather than a fix:**

1. **Box 6 names four dependencies; the description states three.** #5442, #5443 and #5431 are all in "Why this shape". **#5432** — hot reload for sessions that were *not* interrupted — appears nowhere in the body. That is the one mechanical gap here, and it is a line of prose, not code.
2. **Box 1's witness clause asks for two more observations than the tests make.** "Both auto-resume with the text" and "a fresh session loads the rule" are not asserted anywhere; what is asserted is that the words arrive as `WorkspaceInput::Interrupt`, and that one record lands in the registry. The description's argument — that the resume is #5442's already-merged path and the load is the existing record path, reused rather than reimplemented — is a good one and probably the right call. It is still an argument rather than an observation, so if the box is ticked on it, the reasoning belongs in the ticking note where a reviewer can disagree with it.

**Why nothing is ticked yet:** `fmt + clippy + test` has not finished on this head. Every witness above is verified to *exist* and to be absent from `main`; none is yet verified to *pass*. Ticking on source that has not been run is the move this repository's evidence rule exists to prevent, so the boxes want the green test job first. Once it lands, boxes 3, 4 and 5 can be ticked on the assertions quoted above with no further judgement, and 1, 2 and 6 want the two notes resolved first.

---

macanderson:
Checked #5445's six DoD items against the diff at `9029c70`/`839c909d`. Items 3, 4 and 5 hold on real evidence — `a_stalled_session_does_not_hold_up_the_others_stop` is a genuine timing witness against a real `UnixListener` that never accepts, `WhistleRequest` structurally cannot carry `keep` so a target cannot write a record, and `the_whistle_row_cites_the_room_broadcast_witnesses` checks both that the parity row names each witness and that each named function exists.

Three gaps remain against the boxes as written.

**Item 6 — #5432 is cited nowhere.** The item asks the PR to state its dependence on the bang-interrupt issue (#5442), the bang-persistence issue (#5443), #5431 and #5432. The first three are in the body or the commit message; `#5432` appears in neither, nor in `website/content/docs/commands/whistle.mdx`. Grepping the diff for issue references returns only two `#4768` hits, both pre-existing prose. One-line fix.

**Item 1 — the named witness is not the witness that exists.** The item asks for "two running sessions both soft-stop, both auto-resume with the text." `a_room_broadcast_stops_every_live_session_and_the_sender` proves *delivery*: two fixture sessions each receive exactly one `WorkspaceInput::Interrupt`, the sender stops in process, and the note reads `delivered bc1 / delivered bc2 / unreachable bc3`. No driver runs in it, so soft-stop rests on #5442's pre-existing `an_interrupt_front_queues_the_words_and_asks_the_turn_to_stop` and auto-resume is asserted by nothing here. The wiring argument is sound — `interrupt_lead` sends `Interrupt { agent: LEAD }` and the in-turn arm routes that to `steer::interrupt_lead`, which does `push_front` then `request_soft_stop()` — but that is composition, not the end-to-end flip the item names. The lanes clause is knowingly deferred to #6032 (lanes get the words via `deliver_lanes`; nothing stops their turn).

**Item 2 — grammar evidence for a runtime claim.** `three_arrows_and_a_bang_run_stop_the_room_and_keep_the_words` asserts `parse(…)` yields `keep: Some(KeepStrength::Guidance)`. The only runtime `Guidance` case in `keep_record/tests.rs` deliberately makes the write *fail*, so no test shows a `!!` broadcast writing a guidance record successfully. The path is strength-agnostic, so the risk is low — but "does the same at guidance strength" is true by construction rather than demonstrated.

The scope split is disclosed honestly in the body and #6031/#6032 are open, so this is not a concealed shortfall. But two ticked boxes describe behaviour no test in the tree exercises. Either land the driver-level witness, or amend items 1–2 on #5445 to state what is actually proven and carry the remainder into the existing follow-ups.

---

Issue #6034, opened 2026-09-05
OpenRouter docs still show a verifier/triage model table that no longer exists


macanderson:
`website/content/docs/api-providers/index.mdx`'s OpenRouter section shows a table of default models for three roles: `default`, `verifier`, and `triage`.

## What's wrong

Core only has one role now (`default`). The `verifier` and `triage` roles were removed in #3908, when the old staged pipeline was deleted. The function that sets OpenRouter's defaults — `provider_engine_baseline` in `crates/stella-cli/src/engine_config.rs` — only sets an effort level and turns reasoning on for the `default` agent. It does not pick a model for a verifier or a triage role, because those roles do not exist anymore. The doc comment right above that function explains this: the two extra rows ("Opus 5 judging and GLM 5.2 triaging") were retired with #3908.

So the table (around the "With no configuration at all, an OpenRouter instance runs:" line, currently `website/content/docs/api-providers/index.mdx` lines 348-354) describes behavior the code stopped having. A reader would expect OpenRouter to auto-pick a verifier model and a triage model. It does not.

## How to fix it

Rewrite that section so it matches what `provider_engine_baseline` actually does today: effort and reasoning set on the one `default` agent, nothing else. Either drop the table or replace it with a plain sentence describing that one setting. Use `provider_engine_baseline`'s doc comment as the source of truth — it is accurate and up to date.

This is a docs-only change. No code needs to change.

## Definition of done

- [x] The table (or its replacement text) no longer claims OpenRouter auto-picks a verifier or triage model.
- [x] The rewritten section matches `provider_engine_baseline`'s current behavior and doc comment (`crates/stella-cli/src/engine_config.rs`).
- [x] **Verification**: re-read the rewritten section side by side with `provider_engine_baseline` and confirm every claim it makes is true today. Run `rg -n "provider_engine_baseline" crates/stella-cli/src/engine_config.rs` to confirm no other code path sets a verifier/triage model before closing this out. (This repo's own rule: docs-only changes don't need a witness test — AGENTS.md, "Pure refactors, docs, and CI changes don't need a witness" — so this read-back is the verification step in its place.)
- [x] File organization: no change needed. The fix stays in `website/content/docs/api-providers/index.mdx`, its current and correct location — no crate file is touched, so no god-file or crate-placement risk applies.
- [x] Prose follows the repo's plain-language rules (AGENTS.md / CLAUDE.md): short sentences, no filler, cite code by symbol name (`provider_engine_baseline`), not by line number.
[...]

PR #6035, opened 2026-09-05
fix(stella-cli): retire auto_mode instead of leaving it half-alive

### LINKED ISSUE #3936: auto_mode survived the role collapse, but the code that picked a model for it did not — wire it up or retire it
Slice 4 of #3903 (`doc:roleless-core` §6) collapsed the engine settings to one
role and deleted the five `pipeline_<role>_model` keys. `auto_mode` survived
that collapse because the slice's spec names it among the fields to keep — but
**its implementation did not, and it now selects no model at all.**

## What it used to do

`auto_mode: "on"` picked the *verifier's* model out of `allowed_models`,
preferring a family different from the worker's and then ranking by catalog
list price as a capability proxy. Two functions implemented it, both in
`crates/stella-cli/src/engine_config.rs`:

- `auto_verifier_spec(engine, worker_family, configured)` — the selector
- `spec_family(spec)` — the cross-family grouping key it ranked on

Its only consumer was `resolve_engine_wiring`'s verifier pin
(`crates/stella-cli/src/agent/engine.rs`), which #3908 removed along with the
role. Both functions were then dead code, and the slice-4 PR deleted them
rather than leave `-D warnings` unhappy or an `#[allow(dead_code)]` with no
argument behind it.

## What it does now

`auto_mode` is **not fully inert**, which is why it was diminished rather than
retired. It is still one of the three switches `crates/stella-cli/src/profile.rs`
reads and writes, so it participates in "is this config in the auto state, or
has a `/profile` claimed the dials?" — `profile::is_auto`, `restore_auto`, and
`detect` all consult it.

What it no longer does is choose a model. A user reading
`website/content/docs/configuration/agent-engine-config.mdx` before #3908 would
reasonably have expected it to.

## The decision to make

One of two, and it is a product call rather than a cleanup:

1. **Wire it to seats.** The modern reading of "let stella choose" is
   auto-selecting a model for an *unassigned plugin seat* — pick from
   `allowed_models`, prefer a family different from the session's. This is a
   real feature and a genuinely useful one for a verification plugin, and the
   deleted `spec_family` ranking is the algorithm it wants back (recover it
   from the #3908 commit rather than re-deriving).

   The hard part is not the ranking. It is that **core must not decide a seat
   costs money**: `crates/stella-cli/src/agent/seats.rs` documents at length why
   an unassigned seat resolves to the session's model and why core never
   substitutes a default of its own — "a sensible-looking default here is core
   having an opinion about a role it was supposed to be ignorant of". An
   auto-selecting `auto_mode` must therefore be opt-in per seat, or it silently
   multiplies a bill the moment someone installs a multi-participant plugin.

2. **Retire it**, joining `RETIRED_ENGINE_ROOT` in
   `crates/stella-cli/src/settings/unknown.rs`, and move `profile`'s auto-state
   detection onto `effort_auto`/`reasoning_auto` alone.

   **This is not free, and the cost is why #3908 did not just do it.**
   `bench/harbor_adapter/stella_harbor/posture.py` writes `auto_mode: "off"`
   into the hashed Terminal-Bench posture. Retiring the key means either
   leaving it recognized-but-dead (which is what it already is) or dropping it
   from the posture and **re-hashing every digest registered in
   `bench/READINESS.md` §8.4** — a published-numbers decision, the same one
   #3870 reserves for a maintainer.

## Files

- `crates/stella-cli/src/settings/engine.rs` — the field and its doc comment,
  which currently states the gap and cites this issue
- `crates/stella-cli/src/profile.rs` — the surviving consumer
- `crates/stella-cli/src/engine_config.rs` — where the deleted selector lived
- `crates/stella-cli/src/agent/seats.rs` — the "core never picks for you" rule
  any wiring must not break
- `website/content/docs/configuration/agent-engine-config.mdx` — the user-facing
  description, currently marked "Diminished"

## Definition of done

Either `auto_mode` selects a model again — with a witness test showing an
unassigned seat resolving to a cross-family pick, and an assertion that it does
**not** fire without an explicit opt-in — or it is retired with the bench
posture migrated in the same PR and the digest change stated out loud.

Decided: retire it. Checklist for that path:

- [x] `auto_mode` is removed as a field of `AgentEngineConfig`
      (`crates/stella-cli/src/settings/engine.rs`) and of `AgentsSection`
      (`crates/stella-cli/src/settings/toml_config.rs`), and dropped from
      `ENGINE_ROOT_FIELDS`
- [x] `auto_mode` is added to `RETIRED_ENGINE_ROOT`
      (`crates/stella-cli/src/settings/unknown.rs`), so a settings file that
      still sets it is recognized and reported by name instead of read as a
      typo, and the trusted-launcher seam keeps accepting it
- [x] `crates/stella-cli/src/profile.rs`'s auto-state detection (`is_auto`,
      `restore_auto`, `apply`) reads `effort_auto`/`reasoning_auto` alone
- [x] `website/content/docs/configuration/agent-engine-config.mdx` and the
      other user-facing docs stop documenting `auto_mode` as a live key
- [x] The bench-posture decision is stated in the PR description:
      `bench/harbor_adapter/stella_harbor/posture.py` keeps writing
      `auto_mode: "off"` unchanged (recognized-but-dead), so no registered
      digest in `bench/READINESS.md` §8.4 needs a re-hash
- [x] A witness test shows a settings file naming `auto_mode` is reported as
      retired rather than unrecognized, and that `is_auto` no longer depends
      on it

Not acceptable: leaving the doc comment citing this issue forever. The point of
the citation is that someone answers it.

Related: #3903 (epic), #3908 (the slice that diminished it), #3870 (the digest
argument), #3910 (slice 6, where the harness stops writing retired keys).

macanderson:
## What & why

Closes #3936

`auto_mode` survived the #3908 role collapse as a settings field, but the
selector it drove (`auto_verifier_spec`/`spec_family` in
`engine_config.rs`) was deleted with the verifier role it staffed — the key
has picked no model since. It stayed alive only as a third switch
`profile::is_auto`/`restore_auto` toggled alongside `effort_auto`/
`reasoning_auto`, changing nothing on its own.

This issue is a decision issue: wire `auto_mode` to auto-select a model for
an unassigned plugin seat, or retire it. Decision, stated as a comment on the
issue and here: **retire it.** Nothing today consumes a profile- or
auto-picked seat model, and `agent/seats.rs`'s rule that core never invents a
default for an unassigned seat means wiring one would need real new product
surface (an opt-in flag, a witness proving it never fires uninvited) for a
capability nobody is asking for.

**Bench-posture decision** (the one the issue explicitly reserves for a
maintainer): `bench/harbor_adapter/stella_harbor/posture.py` keeps writing
`auto_mode: "off"` unchanged. Because the key is retired-but-recognized
(added to `RETIRED_ENGINE_ROOT`) rather than dropped from the
trusted-launcher's accepted shape, the posture JSON the Python builds is
byte-identical to what it already was, so **no digest registered in
`bench/READINESS.md` §8.4 needs a re-hash**. This is the same shape
`headless_scope_bypass` already takes in this codebase.

## What changed

Retired the exact way #3908 retired `pipeline_verifier_model` and its
siblings:

- `auto_mode` removed as a field of `AgentEngineConfig`
  (`settings/engine.rs`) and `AgentsSection` (`settings/toml_config.rs`);
  dropped from `ENGINE_ROOT_FIELDS` / `TOML_AGENTS_FIELDS`.
- Added to `RETIRED_ENGINE_ROOT` (`settings/unknown.rs`), with a `RETIRED`
  entry for both the JSON (`agent_engine_config.auto_mode`) and TOML
  (`agents.auto_mode`) spellings — a settings file that still sets it is
  recognized and reported by name instead of read as a typo, and the
  trusted-launcher seam (`STELLA_ENGINE_CONFIG_JSON`) keeps accepting it, so
  a benchmark launch naming it is not refused.
- `profile.rs`'s auto-state detection (`is_auto`, `restore_auto`, `apply`)
  now reads `effort_auto`/`reasoning_auto` alone.
- TUI: the engine panel's GLOBAL tab drops the `auto_mode` toggle row (four
  rows instead of five), and the models-card header drops the `model
  on/off` segment it printed.
- Docs: `agent-engine-config.mdx`, `stella-toml.mdx`, `chat.mdx` (its
  `/profile` section described a pre-#3908 six-role verifier-selection
  scheme that no longer matches `profile.rs` at all — rewritten to the
  current one-role behavior while I was in there), `principles/index.mdx`,
  `examples.mdx`, one sentence of `api-providers/index.mdx`, `README.md`,
  `stella.toml`, `stella.example.toml`, `docs/spec/config-system/*`,
  `docs/spec/roleless-core.md`.

## The witness

passes on the new**.

Two:

- `auto_mode_is_reported_as_retired_not_unrecognized`
  (`settings/unknown.rs`) — fails on `main`, where `auto_mode` is still a
  live field of `AgentEngineConfig` and the settings-key scan reports
  nothing at all for a file that sets it; passes here, where the scan
  reports `agent_engine_config.auto_mode` with a retirement reason.
- `is_auto_no_longer_depends_on_auto_mode` (`profile/tests.rs`) — fails on
  `main`, where `is_auto` required `auto_mode_on()` to also be true; passes
  here, where a config with `effort_auto`/`reasoning_auto` on and no
  `auto_mode` field to disagree with reads as auto.

## The gate

- [x] fmt check (ran locally: `cargo fmt --all`, no diff)
- [ ] clippy across the workspace — CI
- [x] CLA signed
- [x] `Closes #3936` appears both above and as a commit trailer

`make guards-fast` (the toolchain-free rung, run locally) is green, including
`check-prose` and `check-file-size`.

## Nothing left behind

- [x] Filed: #6034 — `website/content/docs/api-providers/index.mdx` still
  documents a per-role OpenRouter default table (verifier/triage models)
  that `provider_engine_baseline` has not set since #3908; I fixed the one
  sentence this PR's diff touched but the table itself is a separate,
  larger staleness this PR does not attempt.

## Ground-rule check

- The `tests/snapshots/deck/tab_settings.txt` golden snapshot is hand-edited
  (removed the `auto_mode` row, re-padded the trailing blank lines to keep
  the footer at its original row) rather than regenerated with the blessing
  flag, because this session cannot run the test suite locally. If CI's
  `deck_render_snapshots` disagrees, I'll read the diff and fix it in a
  follow-up push.
- Several JSON/TOML literals across the test suite used `auto_mode` purely
  as a representative "some toggle" filler value; where the specific field
  wasn't under test I swapped them to a surviving key (`effort_auto` /
  `minimal_prompt`) rather than deleting the assertion outright.
- `bench/harbor_adapter`'s Python and every `bench/evidence/**` /
  `bench/terminal_bench_analysis/**` file are untouched on purpose — see the
  bench-posture decision above.

## DoD verification (autonomous PR sweep, 2026-09-05)

#3936's checklist was unticked, which is what held `dod` red. All six items
were checked against this branch's diff and ticked on the issue; the
per-item evidence is in a comment on this PR. The one worth repeating here:
`"auto_mode"` is genuinely added to `RETIRED_ENGINE_ROOT`, not only named in
its doc comment — which is what makes the bench-posture claim above hold
rather than merely be asserted.

Not verified by that sweep, stated rather than left implicit: the tests were
not executed (CI is the evidence), so the two witnesses are confirmed to
exist and to assert the stated property, not confirmed to pass.

## Summary by Sourcery

Retire the obsolete `auto_mode` setting and update configuration handling, profiles, UI, tests, and documentation to use the remaining automatic effort and reasoning controls.

Bug Fixes:
- Retire the inactive `auto_mode` setting so profile auto-state detection no longer depends on a field that cannot select a model.

Enhancements:
- Remove `auto_mode` from engine configuration, TOML configuration, TUI state, and model-status displays.
- Recognize legacy JSON and TOML `auto_mode` settings as retired with explanatory diagnostics while preserving trusted-launcher compatibility.
- Update profile behavior, configuration examples, specifications, and user documentation to reflect the current roleless model-selection and auto-toggle behavior.

Documentation:
- Refresh configuration, profile, roleless-core, principles, examples, and provider documentation to remove obsolete verifier-selection behavior.

Tests:
- Add coverage proving legacy `auto_mode` settings are reported as retired and that auto-state detection depends only on effort and reasoning toggles.


macanderson:
Autonomous PR sweep. The `dod` check was red because #3936's checklist was still unticked. I verified all six items against this PR's diff (head `ac0c256`) and ticked them on the issue. Evidence per item, so the ticks are auditable rather than asserted:

1. **Removed from `AgentEngineConfig` / `AgentsSection` / `ENGINE_ROOT_FIELDS`** — `settings/engine.rs` deletes the `pub auto_mode` field and its doc comment; `settings/toml_config.rs` deletes `pub auto_mode: Option<Toggle>` from `AgentsSection` plus its `raise_agents`/`lower_agents` legs; `settings/unknown.rs` drops `"auto_mode"` from `ENGINE_ROOT_FIELDS` and from `TOML_AGENTS_FIELDS`.
2. **Added to `RETIRED_ENGINE_ROOT`** — `settings/unknown.rs` adds `"auto_mode"` to that constant, and adds `RETIRED` entries for **both** spellings (`agent_engine_config.auto_mode` and `agents.auto_mode`) with a reason. This is the item worth calling out: it is what keeps the trusted-launcher seam accepting the key, so the bench-posture claim in item 5 actually holds rather than being asserted.
3. **`profile.rs` reads `effort_auto`/`reasoning_auto` alone** — `apply` no longer writes `engine.auto_mode`, `restore_auto` restores two switches instead of three, and the module doc records why the third is retired rather than reused.
4. **Docs stop documenting it as a live key** — every surviving mention across `README.md`, `stella.toml`, `stella.example.toml`, `docs/spec/**` and `website/content/docs/**` is a retirement note (`RETIRED:`, "There is no `auto_mode`", "did not stay a live key for long"). No file still describes it as something a user sets.
5. **Bench-posture decision stated** — the PR description states it explicitly, and it is consistent with the diff: `bench/**` is untouched and the key stays recognized per item 2, so `posture.py` keeps emitting byte-identical JSON and no `bench/READINESS.md` §8.4 digest needs a re-hash.
6. **Witness tests present** — `auto_mode_is_reported_as_retired_not_unrecognized` and `the_toml_spelling_of_auto_mode_is_reported_too` in `settings/unknown.rs`, and `is_auto_no_longer_depends_on_auto_mode` in `profile/tests.rs`.

Two things I did **not** verify, stated rather than left implicit: I did not run the test suite (this session does not build the workspace — CI is the evidence), so the witnesses are confirmed to *exist* and to assert the stated property, not confirmed to pass; and the hand-edited `tests/snapshots/deck/tab_settings.txt` golden that the PR description flags is exactly the kind of thing only `deck_render_snapshots` in CI can settle.

`dod` re-run below. The remaining gate is `fmt + clippy + test`.

---

Issue #6038, opened 2026-09-05
Nothing stops the next writer from omitting events.ts and losing millisecond timing again


macanderson:
## The problem

`events.ts` is declared `TEXT NOT NULL DEFAULT CURRENT_TIMESTAMP`
(`crates/stella-store/src/ddl.rs`, `events_ddl`). That default holds whole
seconds. #2604 fixed the resolution by having the one production writer,
`Store::record_event`, stamp the column itself
(`crates/stella-store/src/event_clock.rs`, `INSERT_EVENT`).

Nothing holds the next writer to that. A second insert into `events` that omits
`ts` compiles, passes every gate, and silently writes second-resolution rows
again — which is exactly how the column got here. This repository already had
two writers omitting it once.

## Why the default was not simply changed in #2604

Changing `events_ddl` alone gives a fresh database and a migrated one two
different shapes at one schema version: a fresh file would default to
milliseconds while every existing file kept `CURRENT_TIMESTAMP`. Agreeing them
means a schema bump plus a `lang_altertable` §7 rebuild of `events`, the largest
table in the file, to install a default the writer always overrides. That trade
was judged not worth making for a default nothing reads.

## Options

- A guard script in the style of `scripts/check-typed-errors.py`: fail the gate
  on an `INSERT INTO events` in production source that names no `ts` column.
  Cheap, no schema change, and it reads the same source a reviewer would.
- Or fund the migration and let the schema carry the rule, which removes the
  question instead of policing it.

Either way the answer belongs beside the other gate steps in `AGENTS.md`'s list
and in the `Makefile`'s `GATE_STEPS`.

## Files

- `crates/stella-store/src/ddl.rs` — `events_ddl` and the note on `ts`.
- `crates/stella-store/src/event_clock.rs` — the stamp and its reasoning.
- `crates/stella-store/src/lib.rs` — `Store::record_event`, the one writer.
- `scripts/check-typed-errors.py` — the shape a source guard here would copy.

## Definition of done

- [ ] A new production writer that omits `events.ts` fails something before it
      merges, or the schema makes omitting it harmless.
- [ ] The chosen mechanism has a test that proves it can fail.
- [ ] `AGENTS.md`'s gate list and the `Makefile` agree, if a gate step is what
      lands.

macanderson:
[...]

Issue #6037, opened 2026-09-05
Decide whether the export transcript's elapsed column should read events.ts now that it carries milliseconds


macanderson:
## The problem

`crates/stella-cli/src/export/transcript.rs` builds the transcript's elapsed
column (`Step::offset_ms`) by summing the millisecond durations on
`AgentEvent::ToolResult` and `AgentEvent::StepUsage`. It never reads
`events.ts`.

That was the only honest choice while `events.ts` held whole seconds. #2604
changed the writer, so a row recorded from now on carries
`YYYY-MM-DD HH:MM:SS.sss`, and a wall-clock elapsed column is now possible for
the first time.

A sum of durations is not the same measurement as a wall clock. It covers none
of the gaps between events, and speculative execution deliberately overlaps a
tool's real duration with the model call, so the sum can over-count. #2604's own
body says as much. The current column is therefore a reconstruction that reads
like a measurement, which is the thing this repository does not ship on purpose.

## Why it was not just switched in #2604

Three facts stand against a straight swap, and none of them was settled there:

- Every row written before #2604 is second-resolution and stays that way. A
  fold reading `ts` has to render an old session and a new one from the same
  code, and say which it is looking at.
- A session's journal spans executions (`Store::session_events`), so the base
  reading is not simply the first row of one stream.
- Wall-clock is not monotonic. A clock adjustment mid-session can produce a
  negative delta, and `offset_ms` is unsigned.

## What to decide

Whether the transcript's elapsed column should read `events.ts`, sum durations,
or show both. Then whichever wins, make the module doc say why — rule 2 of
`transcript.rs`'s header is where that argument lives, and it now cites this
issue as the open question.

## Files

- `crates/stella-cli/src/export/transcript.rs` — the fold, and rule 2 of the
  module header.
- `crates/stella-cli/src/export/transcript/tests.rs` — `offset_ms` assertions.
- `crates/stella-store/src/event_clock.rs` — what the stamp is and is not.
- `crates/stella-store/src/lib.rs` — `SessionEventRecord::ts`, which states the
  two widths a reader has to take.

## Definition of done

- [ ] The transcript's elapsed column has one stated source, and the module
      doc gives the reason rather than the history.
- [ ] A session whose rows predate #2604 still renders, and the transcript
      does not claim precision those rows do not have.
- [ ] A test covers the mixed case: one session holding both widths.

macanderson:
[...]

Issue #6024, opened 2026-09-05
Watch nightly-bench for ten nights after the loop-gate fix


macanderson:
## Problem

`nightly-bench` gates on loop health, and #5613 fixed the classification that
held it red: a steering warning the turn survived is no longer scored as a
stuck loop. That fix is proved by two unit tests and by a census of the recorded
streams from four red nights. A dispatched run on the fix branch
(33973144575) is green with `LOOP: 0/4 broken`, but no detection fired in any
of its four streams, so that run alone cannot tell the new code from the old.

One run is not the acceptance test the original issue asked for. Its last
definition-of-done item was "`nightly-bench` runs green ten consecutive nights
with `LOOP: 4/4`", which cannot be ticked before the merge that produces those
nights, so it moves here.

## What to do

Watch `.github/workflows/nightly-bench.yml` for ten consecutive scheduled runs
after the merge of #5613.

- Green with `LOOP: 4/4` on all ten: close this, and say so on #5613.
- A red night: read its artifact before concluding anything. Download
  `loop-bench-<run id>` and open `events/<task>__<id>.jsonl`, then count the
  `loop_detected` events by their `aborted` field. `aborted: true` is a real
  stuck loop and the row is correct — that is an engine finding, and it gets
  its own issue. `aborted: false` alone must no longer produce `STUCK-LOOP`; if
  it does, the fix regressed and this issue is where that is reported.

```
gh api repos/macanderson/stella/actions/runs/<run id>/artifacts
gh api repos/macanderson/stella/actions/artifacts/<artifact id>/zip > a.zip
```

A `CRASHED` night is a different signal and is tracked separately — see the
harbor-timeout issue linked from #5613.

## Files

- `.github/workflows/nightly-bench.yml` — the workflow and its pinned task set.
- `bench/loop-bench/src/lib.rs` — the verdict vocabulary and the two counters
  (`loop_detected`, `loop_aborted`).

## Definition of done

- [ ] Ten consecutive scheduled `nightly-bench` runs after #5613 merged are
      listed here with their run ids and their `LOOP:` line.
- [ ] Any red night among them is explained from its own artifact, and the
      explanation names whether a detection aborted.
- [ ] No ceiling was raised and no loop-health arm was weakened.

macanderson:
[...]

Issue #5613, opened 2026-08-31
nightly-bench has been red for two weeks: the loop gate keeps catching a real stuck loop


macanderson:
The `nightly-bench` workflow has failed **11 of its last 15 runs**, back to 2026-08-17. Nothing tracks it, and it is not a flake — the gate is reporting a genuine engine signal every time.

## What the gate reports

From run [33387694069](https://github.com/macanderson/stella/actions/runs/33387694069) (2026-08-31):

```
overfull-hbox   STUCK-LOOP   138 calls  136 tools   0.05  0.0
        crashed: AgentTimeoutError: Agent execution timed out after 750.0 seconds

LOOP: 1/4 broken (silent-death, zero-work, stuck-loop, or not-run)
REWARD: 2/4 solved   SPEND: $0.11
STUCK-LOOP=1  ran (unsolved)=1  solved=2
```

Every setup step passed. What failed is the measurement.

`STUCK-LOOP` is not the harness giving up on a slow task — per `bench/loop-bench/src/lib.rs` it means **Stella own `loop_detected` fired** and the task did not pass. 138 model calls and 136 tool calls before the 750-second ceiling ended it. This workflow gates on loop health rather than pass rate, so it is doing exactly its job: something in the engine is grinding.

## Why it is not a flake

- 11 red out of the last 15 nightlies, starting 2026-08-17.
- `STUCK-LOOP` appears in every red run sampled, not intermittently.
- The victim task rotates — `overfull-hbox` on 08-31 and 08-29, `fix-git` on 08-29, 08-27 and 08-26 — which points at the loop behaviour rather than at one task fixture.

A rotating victim with a constant failure mode is the signature of an engine problem, not a bad task.

## Where the evidence is

The run uploads `loop-bench-33387694069`, and the per-event detail is in `loop-bench-out/events/*.jsonl` inside it. Read that rather than the summary — the summary layer is a projection, and the full `proof` reason strings are truncated everywhere else.

Start by asking what the 136 tool calls actually were: whether the model is repeating one command, cycling a short set, or making progress the detector is misreading. `#3292` is worth reading first — it covers a model that grinds through many *different* commands, which the detector does not catch, and this may be the same shape seen from the other side.

## What is out of scope

No CI change fixes this. Raising the 750-second ceiling or muting the loop arm would turn the gate green while the engine kept grinding, which is the one outcome worth avoiding — this workflow exists to catch exactly this.

## Definition of done

[...]

PR #6044, opened 2026-09-05
fix(stella-store): stamp events.ts to the millisecond

### LINKED ISSUE #2604: events.ts is second-resolution because every writer omits it, so no consumer can show sub-second timing
## The problem

`events.ts` is declared `ts TEXT NOT NULL DEFAULT CURRENT_TIMESTAMP`
(`crates/stella-store/src/ddl.rs`, `events_ddl`), and **every writer omits the
column**, so every event in the journal takes SQLite's default. That default is
second-resolution (`YYYY-MM-DD HH:MM:SS`) — it has no sub-second component at
all.

Writers, all of which pass only `(execution_id, seq, event_type, payload)`:

- `crates/stella-store/src/lib.rs` — `Store::record_event`
- `crates/stella-store/src/receipts.rs` — the receipts-plane insert

The consequence is that any consumer rendering a timeline off this column can
only show whole seconds. `/export`'s new transcript
(`crates/stella-cli/src/export/transcript.rs`) prints `34s`, never `34.6s`, and
its module doc says why: a tenth of a second there would be a number the
renderer invented rather than measured. For a fast session, every row in the
transcript shows the same second, which is honest but not useful.

This is not only an export concern — the Observatory's journal projection and
anything else reconstructing a run's timing inherit the same ceiling.

## Why the obvious workaround is wrong

The per-call durations on the payloads (`AgentEvent::ToolResult::duration_ms`,
`AgentEvent::StepUsage::duration_ms`) *are* millisecond-precise. Accumulating
them into a synthetic elapsed clock would look better and be wrong: it covers
none of the gaps between events, and speculative execution deliberately overlaps
a tool's real duration with the model call, so the sum over-counts. A
reconstruction presented as a measurement is exactly the class of artifact
CLAUDE.md's bench rule exists to prevent.

## What to do

Write the timestamp explicitly at both insert sites with millisecond (or
microsecond) precision instead of letting SQLite default it. The column is
already `TEXT NOT NULL`, so a wider-precision string needs no migration to
*write* — but decide deliberately:

- Keep `YYYY-MM-DD HH:MM:SS.sss` so existing readers that slice the first 19
  characters keep working (`parse_sql_timestamp` in `export/transcript.rs` does
  exactly this and would tolerate the change unchanged), **or**
- switch to epoch millis and migrate readers, which is a bigger change.

The first is almost certainly right; note that mixed-precision rows will exist
forever (everything written before the change), so a reader must tolerate both —
which the positional parser already does.

## Verify

- A witness that two events recorded in quick succession get distinguishable
  timestamps, where today they do not.
- `crates/stella-cli/src/export/transcript/tests.rs::the_elapsed_column_counts_whole_seconds_and_invents_no_precision`
  encodes the *current* contract and will need updating in the same PR — its
  assertion that no decisecond appears is the thing being deliberately changed.
  Update the `SessionEventRecord::ts` doc comment (`stella-store/src/lib.rs`),
  which states the second-resolution limitation as a fact, and the transcript
  module's property 2.

## Definition of done

Restated as a checklist, and reconciled with the tree as it stands today (see
the design comment below: the receipts insert named above is now test-only, so
there is one writer, and the transcript no longer reads `events.ts` at all).

- [x] `events.ts` carries sub-second precision at every production writer of
      the `events` table.
- [x] A witness test fails on the old code by construction, not by racing two
      clock reads.
- [x] Readers that slice the first 19 characters keep working, and rows of
      both widths are tolerated.
- [x] The transcript's elapsed column question is settled — **split out to
      #6037** (SCR-004), which decides whether that column should read
      `events.ts`, sum durations, or show both. It is not a stamp-writer
      concern: this issue owns making the precision *exist*, and #6037 owns
      whether the transcript should switch its source to it. Swapping the
      source here would have shipped three undecided questions with it — old
      second-resolution rows, a journal spanning executions, and a
      non-monotonic wall clock against an unsigned `offset_ms`.
- [x] The docs asserting second resolution are updated rather than left
      contradicting the code: `SessionEventRecord::ts`, the `events` DDL, and
      the export transcript's module header.

Found while building the `/export` transcript (PR for the transcript feature,
which states the limitation rather than faking around it).

macanderson:
## What

`Store::record_event` now writes `events.ts` itself, to the millisecond,
instead of leaving the column to SQLite's `DEFAULT CURRENT_TIMESTAMP`.

## Why

`events.ts` is declared `TEXT NOT NULL DEFAULT CURRENT_TIMESTAMP`
(`crates/stella-store/src/ddl.rs`, `events_ddl`) and the one production writer
of the table omitted the column. That default holds whole seconds. A fast turn
wrote every one of its events under a single reading, and any consumer drawing
a timeline off the column could show seconds and no more — the Observatory's
journal projection included.

The stamp is `strftime('%Y-%m-%d %H:%M:%f', 'now')`: SQLite's own UTC clock,
the same one `CURRENT_TIMESTAMP` reads, with a millisecond field on the end.
`YYYY-MM-DD HH:MM:SS.sss`, so the first 19 characters are byte for byte what
they were. No reader changes, no stored row moves, no new dependency, no
schema migration.

The statement lives in a new `crates/stella-store/src/event_clock.rs` rather
than in `lib.rs`, which is a god file closed to growth (AGENTS.md § God
files). It follows `receipts.rs`'s shape: a sibling module owning one part of
the store's contract, with its reasoning and its tests beside it. `lib.rs` is
1673 lines against a 1719 ceiling, and the insert itself came out of it.

## Witness mechanism

`crates/stella-store/src/event_clock.rs`'s
`a_recorded_event_is_stamped_to_the_millisecond` records two events and
asserts each `ts` is 23 characters, with `.` at index 19 and three digits
after it. On `main` the writer omits the column, so every row is the DDL
default's 19 characters with no `.` in it and all three assertions fail. It
fails by construction, not by racing two clock reads: it asks the *shape* of
one stamp rather than comparing two, so a machine fast enough to write both
rows inside one millisecond cannot make it flap. It also asserts the
19-character prefix still has the old shape, which is what lets existing
readers stay untouched.

## The issue's premises, re-verified

The reopen note asks for this, and two of them had gone stale:

- **Still true.** One production writer, `Store::record_event`, omitting `ts`.
- **Stale.** The receipts-plane insert named as a second writer is now
  test-only (`crates/stella-store/src/receipts.rs`, inside `mod tests`).
- **Stale.** `parse_sql_timestamp` and
  `the_elapsed_column_counts_whole_seconds_and_invents_no_precision` no longer
  exist. `/export`'s transcript was rewritten to sum the millisecond durations
  on `ToolResult`/`StepUsage` and never reads `events.ts`, so nothing there
  needed updating beyond the module header, whose stated reason ("no
  sub-second part to read") stops being true with this change.

## Deliberately not done

**The DDL default stays at `CURRENT_TIMESTAMP`.** Changing `events_ddl` alone
would leave a fresh database and a migrated one holding two shapes at one
schema version; agreeing them means a schema bump plus a `lang_altertable` §7
rebuild of `events`, the largest table in the file, to install a default the
writer always overrides. Said out loud rather than left implicit, per
CLAUDE.md: nothing now stops a *future* writer omitting the column and
regressing this, and that gap is #6038.

**The transcript's elapsed column is not repointed at `ts`.** Rows written
before this change are second-resolution forever, a session's journal spans
executions, and wall-clock does not survive a clock adjustment. That is a
judgement with its own evidence and it is #6037. The column already shows
millisecond timing for a fast session today, by summing measured durations.

## Docs corrected in the same PR

`SessionEventRecord::ts` (`crates/stella-store/src/lib.rs`), which stated
second resolution as a fact; the `events` DDL, which now says the default is
not what a new row gets; and rule 2 of
`crates/stella-cli/src/export/transcript.rs`'s module header.

## Exemplar

`crates/stella-store/src/receipts.rs` — a sibling module holding one slice of
the store's write contract, its argument, and its tests.

Tests deleted: None

Residue filed: #6037, #6038

Closes #2604

## DoD reconciliation (#2604)

`dod / dod` was red on `1113a2d` because #2604's five boxes were unticked at the
time that run fired (15:19 UTC); they were ticked at 15:27, after it. Each was
checked against the tree rather than against this description:

1. **Sub-second precision at every production writer.** `Store::record_event`
   (`crates/stella-store/src/lib.rs`) is the only one. `receipts.rs`'s insert and
   `stella-observatory/src/turn_agents.rs`'s are both inside `#[cfg(test)]`
   blocks; the `migrations/*` inserts rewrite historic rows and keep their
   original `ts` by design.
2. **A witness that fails on the old code by construction.**
   `a_recorded_event_is_stamped_to_the_millisecond` asserts the shape of a single
   stamp — 23 characters, `.` at index 19, three digits after it. A
   default-stamped row is 19 characters with no `.`, so it fails on `main`
   without comparing two clock reads, and cannot flap on a fast machine.
3. **The 19-character prefix still slices.** `strftime('%Y-%m-%d %H:%M:%f',
   'now')` appends `.sss` to the same UTC reading `CURRENT_TIMESTAMP` gives; the
   prefix is asserted unchanged in the same test, and `SessionEventRecord::ts`
   documents both widths.
4. **The transcript's elapsed column** is split to #6037 under SCR-004 (`triage`
   label only), not dropped — this is also the one row Sourcery's assessment
   marks ❌, answered in a comment on this PR.
5. **Docs updated** in the three places that asserted second resolution:
   `SessionEventRecord::ts`, the `events` DDL comment, and rule 2 of the export
   transcript's module header.

## Summary by Sourcery

Stamp newly recorded events to the millisecond while preserving compatibility with existing second-resolution rows and duration-based transcript timing.

Bug Fixes:
- Stamp newly recorded store events with millisecond precision instead of relying on SQLite's second-resolution default timestamp.

Enhancements:
- Document the mixed timestamp precision of existing and newly written rows while preserving reader compatibility.
- Move event timestamp insertion logic into a dedicated store module with coverage for the timestamp format.

Documentation:
- Update store and transcript documentation to reflect millisecond event timestamps and the continued use of measured durations for elapsed timing.

Tests:
- Add a regression test verifying recorded event timestamps include a three-digit millisecond component and retain the existing prefix format.

---

### DIFF
--- a/crates/stella-cli/src/export/transcript.rs
+++ b/crates/stella-cli/src/export/transcript.rs
@@ -25,13 +25,17 @@
 //! miss a field the way a scattered set of `clean()` calls could.
 //!
 //! **2. Never invent a clock reading finer than what was measured.**
-//! `events.ts` is a plain SQLite timestamp with no sub-second part
-//! ([`SessionEventRecord::ts`](stella_store::SessionEventRecord::ts)), so
-//! this fold never reads it. [`Step::offset_ms`] is summed instead from the
-//! real, millisecond-precise durations on `ToolResult` and `StepUsage` — the
-//! same durations the shared renderer already draws for `stella observe`,
-//! and the same sum [`stella_tui::transcript_build::RunBuilder`] already
-//! keeps for the live surface.
+//! [`Step::offset_ms`] is summed from the real, millisecond-precise durations
+//! on `ToolResult` and `StepUsage` — the same durations the shared renderer
+//! already draws for `stella observe`, and the same sum
+//! [`stella_tui::transcript_build::RunBuilder`] already keeps for the live
+//! surface. This fold does not read `events.ts`. Since #2604 that column is
+//! stamped to the millisecond
+//! ([`SessionEventRecord::ts`](stella_store::SessionEventRecord::ts)), so the
+//! old reason — nothing sub-second to read — has gone. Rows written before
+//! #2604 still hold whole seconds, a session's journal spans executions, and
+//! wall-clock does not survive a clock adjustment; whether those make `ts` the
+//! better source here is #6037.
 //!
 //! **3. Drop nothing.** The shared run model has no slot for most of the
 //! wire's event kinds: a stage boundary, a provider fallback, a verdict, and
--- a/crates/stella-store/src/ddl.rs
+++ b/crates/stella-store/src/ddl.rs
@@ -162,6 +162,12 @@ pub(crate) const UNCHANGED_TABLES: &str = "CREATE TABLE IF NOT EXISTS file_locks
 /// is exactly the replay access path (superseding the pre-v1 non-unique
 /// `events_by_execution` index, which is why no separate index exists).
 ///
+/// The `ts` default is not what a new row gets. `Store::record_event` writes
+/// the column itself, to the millisecond (`crate::event_clock`). The default
+/// is only what older rows carry, and they keep it. A migrated file and a
+/// fresh one therefore hold the same second-resolution default and the same
+/// millisecond rows.
+///
 /// `task_id` (v39, #5039) is the board task this event is evidence for, `NULL`
 /// when it is in no task's ledger — either nothing was running when it was
 /// dispatched, or the case carries no tag at all. It is a projection of the
--- a/crates/stella-store/src/event_clock.rs
+++ b/crates/stella-store/src/event_clock.rs
@@ -1,0 +1,75 @@
+//! The clock stamped on `events.ts`, and the insert that stamps it.
+//!
+//! The `events` table declares `ts TEXT NOT NULL DEFAULT CURRENT_TIMESTAMP`
+//! (`ddl::events_ddl`). That default holds whole seconds and nothing finer.
+//! A fast turn writes many events inside one second. Left to the default,
+//! they all share one reading, and a timeline drawn from them can show
+//! seconds and no more.
+//!
+//! `INSERT_EVENT` stamps the column rather than leaving it to the default.
+//! `strftime('%Y-%m-%d %H:%M:%f', 'now')` reads the same UTC clock that
+//! `CURRENT_TIMESTAMP` reads, and adds a field of milliseconds on the end.
+//! The first 19 characters do not change. A reader that slices that prefix
+//! still works, and no stored row has to move.
+//!
+//! Older rows hold whole seconds and stay that way, so a reader of `ts` has
+//! to take both widths.
+//!
+//! The default in the DDL stays as it is. A change there would leave a fresh
+//! file and a migrated file with two shapes at one schema version. To agree
+//! them costs a rebuild of the largest table in the store, and the default it
+//! would install is one this writer always overrides.
+
+/// The `events` insert, with `ts` stamped to the millisecond.
+///
+/// The five placeholders bind `execution_id`, `seq`, `event_type`, `payload`
+/// and `task_id`, in that order. `ts` is SQL, not a bound value, so a row is
+/// dated by the database that stores it and never by a caller's clock.
+pub(crate) const INSERT_EVENT: &str = "INSERT INTO events \
+     (execution_id, seq, ts, event_type, payload, task_id) \
+     VALUES (?, ?, strftime('%Y-%m-%d %H:%M:%f', 'now'), ?, ?, ?)";
+
+#[cfg(test)]
+mod tests {
+    use stella_protocol::AgentEvent;
+
+    use crate::Store;
+
+    /// **The witness.** Every recorded event carries a millisecond field.
+    ///
+    /// A row that took the DDL default would be 19 characters wide and hold
+    /// no `.`, so it fails every check below. This asks the shape of one
+    /// stamp. It does not compare two, so a machine fast enough to write both
+    /// rows inside one millisecond cannot make it flap.
+    #[test]
+    fn a_recorded_event_is_stamped_to_the_millisecond() {
+        let store = Store::in_memory().unwrap();
+        let id = store
+            .begin_execution("run", "goal", "zai", "glm-5.2")
+            .unwrap();
+        store
+            .record_event(id, 0, &AgentEvent::Text { text: "a".into() })
+            .unwrap();
+        store
+            .record_event(id, 1, &AgentEvent::Text { text: "b".into() })
+            .unwrap();
+
+        let journal = store.execution_events(id).unwrap();
+        assert_eq!(journal.events.len(), 2, "both events landed");
+        for row in &journal.events {
+            let ts = &row.ts;
+            assert_eq!(ts.len(), 23, "expected YYYY-MM-DD HH:MM:SS.sss, got {ts:?}");
+            assert_eq!(&ts[19..20], ".", "no millisecond field in {ts:?}");
+            assert!(
+                ts[20..].chars().all(|c| c.is_ascii_digit()),
+                "millisecond field is not three digits: {ts:?}"
+            );
+            assert!(
+                ts[..19]
+                    .chars()
+                    .all(|c| c.is_ascii_digit() || matches!(c, '-' | ' ' | ':')),
+                "the 19-character prefix a reader slices changed shape: {ts:?}"
+            );
+        }
+    }
+}
--- a/crates/stella-store/src/lib.rs
+++ b/crates/stella-store/src/lib.rs
@@ -114,6 +114,8 @@ use stella_protocol::{AgentEvent, TaskStatus};
 //               spool, and the per-store export ledger
 //   error       (crate-private module, `StoreError` re-exported) every
 //               failure this crate returns, as named cases
+//   event_clock (crate-private) the millisecond stamp on `events.ts`, and
+//               the insert that writes it
 //   export      the `/export` telemetry dump and `stella stats`' usage
 //               aggregate, plus the `ExportScope` that decides whether either
 //               covers one session or the whole workspace (#2558)
@@ -135,6 +137,7 @@ use stella_protocol::{AgentEvent, TaskStatus};
 mod ddl;
 mod dispatch;
 mod error;
+mod event_clock;
 mod git_env;
 mod migrations;
 mod private;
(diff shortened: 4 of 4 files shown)


macanderson:
**`dod / dod` is red on `1113a2d`: #2604's five boxes are all unticked.** Automated PR sweep. Four of the five are satisfied by this diff and I verified each against the tree rather than against the description; the fifth is genuinely not, and it is the only thing standing between this PR and a green gate.

| # | #2604 DoD item | Reads as | What I checked |
|---|---|---|---|
| 1 | Sub-second precision at **every production writer** of `events` | **Satisfied** | Censused `INSERT INTO events` across the repo. One production writer: `Store::record_event`, and this diff repoints it at `event_clock::INSERT_EVENT`. Every other hit is a fixture or a rebuild — `receipts.rs`'s is inside `mod tests` (it inserts `'{ this is not json'`), the three `migrations/*` ones seed files under test, the `stella-observatory` ones are fixtures, and `legacy_rebuild.rs` is `INSERT INTO events_v1 … SELECT … FROM events`, which carries existing `ts` values across rather than minting one. The reopen note's "two writers" premise is stale exactly as the PR says. |
| 2 | Witness fails on old code **by construction, not by racing two clock reads** | **Satisfied** | `a_recorded_event_is_stamped_to_the_millisecond` asserts the shape of *one* stamp — `len == 23`, `.` at index 19, three digits after. A row taking the DDL default is 19 characters with no `.`, so all three assertions fail on `main`. Nothing compares two readings, so it cannot flap on a fast machine. This is the stricter reading of the item, not the easy one. |
| 3 | 19-character slicers keep working; both widths tolerated | **Satisfied** | `strftime('%Y-%m-%d %H:%M:%f', 'now')` appends the fractional field and leaves the prefix byte-identical; the witness pins the prefix shape with its own assertion. Both widths are documented on `SessionEventRecord::ts` as a reader obligation. |
| 4 | **The transcript's elapsed column shows real sub-second timing for a fast session** | **Not satisfied** | See below. |
| 5 | The three docs asserting second resolution are updated | **Satisfied** | All three are in the diff: `SessionEventRecord::ts` (`stella-store/src/lib.rs`), the `events` DDL comment (`ddl.rs`), and rule 2 of `stella-cli/src/export/transcript.rs`'s module header. |

### Item 4 is the blocker, and the PR's reason for it does not hold

The PR sets item 4 aside with: *"The column already shows millisecond timing for a fast session today, by summing measured durations."* That is true as a description of behaviour and it is not what item 4 asks, because #2604's own body rules that mechanism out:

> Accumulating them into a synthetic elapsed clock would look better and be wrong: it covers none of the gaps between events, and speculative execution deliberately overlaps a tool's real duration with the model call, so the sum over-counts. A reconstruction presented as a measurement is exactly the class of artifact CLAUDE.md's bench rule exists to prevent.

So the duration sum cannot be the evidence for item 4 — the issue filed item 4 *because* the duration sum is not a measurement of elapsed time. The item wants the transcript reading a real clock, which is what repointing it at `ts` would do and what this PR deliberately does not do.

That deferral looks like the right call on the merits — the reasons given (pre-#2604 rows are second-resolution forever, a session's journal spans executions, wall-clock does not survive a clock adjustment) are real, and #6037 is the right place for the judgement. The gap is only that the deferral is recorded in the PR description and in the transcript module header, and **not on #2604's checklist**, which is the only text `dod` reads.

### Remedy

Same shape as #5613's "ten consecutive green nights" → #6024 earlier today, and the same one the gate's own failure text names: split item 4 out under SCR-004 and record it on #2604 as *moved*, not dropped. #6037 already exists and already carries the substance, so this is an edit to #2604's checklist citing it — after which boxes 1, 2, 3 and 5 tick on the evidence above and `dod` goes green.

I ticked nothing myself: item 4 is false today, and ticking the other four while it stands leaves the gate red anyway.

---

macanderson:
**Answering Sourcery's one ❌ row — deferred by design, with the deferral recorded on the issue rather than assumed.**

The row: *"Update the transcript and related documentation/readers so fast-session elapsed timing uses measured sub-second event timing rather than remaining limited to whole seconds or relying on invented duration accumulation."*

Sourcery is factually right about the diff. `crates/stella-cli/src/export/transcript.rs` still sums the `ToolResult`/`StepUsage` durations for `Step::offset_ms` and does not repoint the elapsed column at `events.ts`. That is deliberate, and #2604's definition of done now records it as **split out to #6037** (SCR-004) rather than dropped — box 4 reads *"The transcript's elapsed column question is settled — split out to #6037"*.

Why a split rather than an omission: this PR owns making the precision *exist*; repointing the transcript's source is a separate decision carrying three questions none of which are settled here, all named in #6037 —

- every row written before this PR is second-resolution and stays that way, so a fold reading `ts` has to render both eras and say which one it is looking at;
- a session's journal spans executions (`Store::session_events`), so the base reading is not simply the first row of one stream;
- wall clock is not monotonic, and `offset_ms` is unsigned, so a clock adjustment mid-session can produce a negative delta.

Swapping the source here would have shipped those three undecided. #6037 is open, carries the `triage` label only per SCR-005, and names the files plus its own DoD.

Checked against the tree rather than against this PR's description:

- `Store::record_event` (`crates/stella-store/src/lib.rs`) is the only **production** writer of `events` — `receipts.rs:562` and `stella-observatory/src/turn_agents.rs:302` are both inside `#[cfg(test)]` blocks (opening at lines 515 and 256 respectively), and the `migrations/*` inserts rewrite historic rows and keep their original `ts` by design.
- The witness asserts the shape of a single stamp (23 characters, `.` at index 19, three digits after it) rather than comparing two clock reads, so it fails on `main` by construction — a default-stamped row is 19 characters with no `.` — and cannot flap on a fast machine.
- The 19-character prefix a reader slices is unchanged: `strftime('%Y-%m-%d %H:%M:%f', 'now')` appends `.sss` to the same UTC reading `CURRENT_TIMESTAMP` gives.

No fix pushed for this row, by design; the follow-up is tracked in #6037.

---

macanderson:
## Answering Sourcery's ❌ (the one row)

> The PR updates the transcript documentation but deliberately leaves the
> transcript's elapsed column based on summed payload durations and does not
> repoint it to `events.ts`.

Accurate, and deliberate. Answering rather than fixing, for three reasons the
swap does not have an answer to yet:

1. **Every row written before this change is second-resolution and stays that
   way.** A fold reading `ts` has to render an old session and a new one from
   one code path, and say which it is looking at. Nothing in this PR decides
   what that looks like.
2. **A session's journal spans executions** (`Store::session_events`), so the
   base reading is not simply the first row of one stream.
3. **Wall-clock is not monotonic.** A clock adjustment mid-session yields a
   negative delta, and `Step::offset_ms` is unsigned.

The issue is right that a sum of durations is a reconstruction, and the
transcript's module header now says so in those words rather than resting on
the old reason ("`events.ts` has no sub-second part") that this PR just made
false. Which source that column should use is #6037, filed as a handoff with
its own definition of done and `triage` only.

Worth separating the two halves the row runs together: this issue owns making
the precision **exist**, and it now does, at the writer, with a witness. #6037
owns whether the transcript should **switch its source** to it. Shipping the
swap here would have carried the three undecided questions above with it, into
a surface a user reads as a measurement.

The elapsed column is not whole-seconds today either way — `Step::offset_ms`
is millisecond-precise from the measured `ToolResult`/`StepUsage` durations
(`crates/stella-cli/src/export/transcript/tests.rs` asserts 1500 ms on a second
step). The open question is which of two millisecond sources is the honest one,
not whether any exists.

Issue #4576, opened 2026-08-23
A test meant to check a credential-prompt failure never actually triggers that failure


macanderson:
## Problem

`crates/stella-model/tests/credential_prompt_degrade.rs` claims to prove that a failed prompt degrades to `NotFound` rather than surfacing `PromptFailed` — but under `cargo test` (locally and in CI) `stdout().is_terminal()` is false, so `ApiKey::resolve`'s interactive gate declines before `prompt_for_key` is ever called. The test is green because of the gate, not the swallow: the arm it documents is unexercised.

## Repro

Add an `eprintln!` in the interactive arm of `ApiKey::resolve` (`crates/stella-model/src/credential.rs:~205`) and run the test — it never prints.

## Done means

The swallow is exercised deterministically: either an injectable prompt decision on the public API, or a subprocess harness that detaches the controlling terminal with `setsid`. This is also the prerequisite for #3052 — its correct fix (probe what rpassword actually opens) currently cannot land because this test would then reach a real blocking `prompt_password` on any machine with a controlling terminal.

Found during PR #4575 (2026-08-23 sweep). Related: #3052.

PR #6046, opened 2026-09-05
fix(stella-model): probe the terminal rpassword actually opens, not stdio

### LINKED ISSUE #3052: The password-prompt check watches the wrong output stream on Unix machines
## Context

While fixing #3036 (widening `crates/stella-model/src/credential.rs`'s
interactive-prompt check to also confirm the prompt is actually visible, not
just that stdin is a terminal), I found a gap. The fix checks stdout, copying
the shape of `stella_tty::human_can_answer`. But `rpassword`, which prints
the password prompt, does not write to stdout on Unix.

`crate::credential::prompt_for_key` calls `rpassword::prompt_password`. Its
default config (`rpassword::ConfigBuilder::default()`) writes the prompt text
to, and reads the answer from, `/dev/tty` directly on Unix — see
`rpassword-7.5.4/src/unix.rs::DEFAULT_OUTPUT_PATH` ("/dev/tty") and
`config.rs`'s `ConfigBuilder::default()`. It never goes through
`std::io::Stdin`/`std::io::Stdout` on that platform. (Windows uses `CONOUT$`,
same problem; wasm uses `/dev/stdout`, which does match stdout.)

So the check landed in #3036 — `std::io::stdout().is_terminal()` — is right
for every other caller of `stella_tty::human_can_answer` (the `ask_user` tool
and the CLI's approval responder both really do print to stdout), but it's
only a stand-in here. Two failure directions:

- **False negative:** a process with redirected stdout (`stella config >
  out.txt`) but a live controlling terminal would decline to prompt, even
  though `rpassword` would have shown the prompt fine on the real terminal
  via `/dev/tty`.
- **False positive** (rarer): stdout looks like a terminal but no `/dev/tty`
  is available — some container/CI setups.

## Why this wasn't fixed in #3036

#3036's definition of done only asked the fix to match `human_can_answer`'s
existing (stdin, stdout) shape — the same shape every other caller uses.
Making the credential-prompt check exactly right needs either:

- checking whether `/dev/tty` can be opened directly, which depends on how
  `rpassword` is implemented and differs by platform (Windows/wasm), or
- getting `rpassword` to report what it will actually do before committing
  to the prompt.

Both are real design decisions, not a mechanical copy of the existing check,
so this belongs in its own PR.

## Suggested approach

- Check whether `rpassword` (or something lower-level) offers a cheap "can I
  actually prompt right now" probe, or whether opening `/dev/tty` read-write
  and immediately closing it (mirroring what `prompt_password` does) is an
  acceptable cheap check.
- Alternatively, configure `rpassword` via
  `ConfigBuilder::default().writer(...)` or similar to route through
  `std::io::Stdout`/`Stdin` explicitly, so the existing
  `stella_tty::human_can_answer(interactive, stdin_is_terminal,
  stdout_is_terminal)` check becomes exactly correct instead of a stand-in —
  this trades away `rpassword`'s ability to survive a redirected
  stdin/stdout, if that tradeoff is acceptable.
- Either way, add a test that tells apart "stdout redirected, real
  controlling terminal present" from "no controlling terminal at all" — the
  two cases the current stdout-based check cannot distinguish.

## Files involved

- `crates/stella-model/src/credential.rs` (`prompt_for_key`, `ApiKey::resolve`)
- `crates/stella-tty/src/lib.rs` (the shared `human_can_answer` check, added
  in #3036)
- `crates/stella-model/tests/credential_prompt_degrade.rs` (existing test;
  keep its documented workaround — `stdin().is_terminal()` reports `true`
  under the `cargo test` harness even though the stream isn't really
  readable)

## Definition of done

- [x] The credential prompt's interactivity check reflects what `rpassword` will
  actually do on the current platform, not a stdout stand-in for it.
  **#6046**: `credential/prompt.rs`'s new `rpassword_tty` module opens
  `/dev/tty` on Unix and `CONIN$`/`CONOUT$` on Windows, feeding those two
  results into `TerminalPrompt::decide` in place of
  `stdin()/stdout().is_terminal()`. The `wasm.rs` case this issue calls out as
  genuinely matching stdio keeps the `IsTerminal` fallback, under
  `cfg(not(any(unix, windows)))`. `auth_cmd.rs` was moved onto the same gate
  so the CLI's two `rpassword` call sites cannot disagree with the credential
  chain.
- [x] A test distinguishing the two cases named above.
  **#6046**: `crates/stella-model/tests/credential_prompt_dev_tty.rs` builds a
  real pty with `libc::openpty` + `setsid` + `TIOCSCTTY` and runs a
  `tty-probe` fixture binary inside it, one case each —
  `can_prompt_reaches_dev_tty_even_when_stdout_is_redirected_away` (stdout is
  a plain file, controlling terminal live → `true`, where a stdout check says
  `false`) and `can_prompt_declines_with_no_controlling_terminal_at_all` (new
  session, no `TIOCSCTTY` → `false`). The second is the control that stops the
  first passing on a probe that always answers `true`. A child process rather
  than the test binary, because `cargo test`'s own process has no controlling
  terminal to give.
- [x] `make gate` green; no new god-file lines, no baseline raised.
  Settled on `f5e5acd`, the current head. `fmt + clippy + test` — the job
  carrying fmt, clippy and the workspace test suite — concluded **success** at
  16:22:07 UTC ([run 33976251592](https://github.com/macanderson/stella/actions/runs/33976251592),
  check run 101333650162). `file size ratchet` is green on the same head, and
  the diff touches no baseline file and no god file: the changed paths are
  `credential.rs`, `credential/prompt.rs`, `credential/tests.rs`,
  `auth_cmd.rs`, the two new test files, `Cargo.toml` and `Cargo.lock`.

macanderson:
## What & why

`TerminalPrompt::can_prompt` (the gate in front of every `rpassword` prompt)
checked `std::io::stdin().is_terminal()`/`std::io::stdout().is_terminal()`.
`rpassword` never reads or writes either on Unix or Windows — it opens
`/dev/tty` (Unix) or `CONIN$`/`CONOUT$` (Windows) directly. So the check was
a stand-in, not the exact condition, in both directions the issue names:

- **False negative:** `stella config > out.txt` from a real terminal has a
  redirected stdout but a live controlling terminal — `rpassword` would show
  the prompt fine on `/dev/tty`, but the old check declined.
- **False positive:** a `stdout` that merely looks like a terminal with no
  controlling terminal behind it could wrongly pass.

`can_prompt` now probes exactly what `rpassword` opens (open, then drop) —
`crates/stella-model/src/credential/prompt.rs`'s private `rpassword_tty`
module, cfg-gated per platform, with the old `stdin()`/`stdout()` check kept
only as the fallback on a target that is neither Unix nor Windows (where
`rpassword`'s own fallback is `/dev/stdin`/`/dev/stdout` — the one case the
old check was already exact for).

`stella-cli`'s `auth set` prompts (`crates/stella-cli/src/auth_cmd.rs`, two
`rpassword::prompt_password` call sites) had no visibility gate at all — per
the issue thread's own follow-up comment, both now share `TerminalPrompt`'s
fixed gate instead of re-deriving one, so the command and the provider
credential chain can never disagree about whether a human is there to
answer.

## The witness

`crates/stella-model/tests/credential_prompt_dev_tty.rs` drives a new
`tty-probe` fixture binary (`crates/stella-model/tests/fixtures/tty_probe.rs`)
inside a real pseudo-terminal built with `libc::openpty`/`setsid`/`TIOCSCTTY`
— the same construction `stella-tui`'s `tests/deck_pty_smoke.rs` already
uses. With the pty as the child's controlling terminal but its stdout
redirected to a plain file, a check based on `stdout().is_terminal()` prints
`false`; `can_prompt` probes `/dev/tty` directly and prints `true`. A second
scenario with no controlling terminal at all pins that the probe still
declines correctly (the fix is a more accurate probe, not one that always
says yes).

This has to run as a real subprocess rather than an in-process call: the
property is about *that process's own* controlling terminal, which nothing
running inside the test binary's own session can fake.

**A pre-existing test needed rewriting alongside the fix, not just a new
one.** `resolve_never_prompts_when_interactive_is_true_but_stdin_is_not_a_terminal`
drove the real `TerminalPrompt` through `ApiKey::resolve(..., true)`, relying
on the test harness capturing stdout to stay safe. Since `can_prompt` now
probes `/dev/tty` independent of stdout, that reliance breaks: on a machine
with a controlling terminal attached to the test process, it would call
`rpassword` for real and block on a keystroke that never arrives — exactly
the hazard the issue's first comment flagged as blocking this fix until the
prompt decision became injectable (#4576, already landed). It's rewritten as
`resolve_never_prompts_when_interactive_is_true_but_no_prompt_is_available`,
driving `resolve_with_prompt` with a fake `CredentialPrompt` instead, the
same shape every other interactive-path test in the file already uses.

Exemplar: `stella-tui`'s `tests/deck_pty_smoke.rs` for the pty harness
construction (`libc::openpty`/`setsid`/`TIOCSCTTY`); `stella-runtime`'s
`tests/fixtures/wrapper-plugin-fixture.rs` for the `[[bin]]`-under-`tests/`
fixture pattern (`CARGO_BIN_EXE_<name>`).

## The gate

- [x] fmt check
- [ ] clippy across the workspace, all targets, warnings denied — CI (this
      session does not build/test on the maintainer's machine)
- [x] Docs updated where behavior/flags changed (doc comments on
      `TerminalPrompt::can_prompt`/`decide`, `ApiKey::resolve_with_prompt`)
- [ ] CLA signed
- [x] `Closes #3052` appears both above and as a commit trailer

## Nothing left behind

- [x] There is nothing: everything I noticed is fixed in this PR. (The
      Windows `CONIN$`/`CONOUT$` arm is mirrored code, structurally the same
      shape as the Unix arm already witnessed here — no separate follow-up
      filed, matching how the rest of this tree treats Windows-only
      behavior: `windows-check.yml`'s compile-only check, not a full
      Windows pty behavioral harness, is the existing bar.)

## Ground-rule check

- New dev-dependency: `libc` in `crates/stella-model/Cargo.toml`, already a
  workspace dependency used the same way by `stella-tui`'s
  `deck_pty_smoke.rs` and `stella-runtime`'s pty-adjacent tests — justified
  by the witness test needing a real pseudo-terminal, which nothing else in
  the workspace dependency set provides.

- `crates/stella-model/src/credential.rs`'s `resolve`/`resolve_with_prompt`
  doc comments are updated to say plainly that driving the real
  `TerminalPrompt` through a test is no longer implicitly safe — it depends
  on the *test process's* terminal state, not on test-harness stdout
  capture. Every test in the tree already goes through `resolve_with_prompt`
  with a fake now; grepped for other callers of `ApiKey::resolve(..., true)`
  under test and found none.

Closes #3052

## Summary by Sourcery

Probe rpassword's actual terminal endpoints before prompting and apply the shared gate to all interactive credential requests.

Bug Fixes:
- Make credential prompting detect the actual terminal devices used by rpassword, allowing prompts with redirected standard output while rejecting environments without a controlling terminal.
- Guard all stella-cli auth secret prompts with the shared terminal-availability check so command-line and provider credential prompting behave consistently.

Enhancements:
- Clarify credential-resolution guidance and use injected prompts for deterministic interactive-path testing.

Build:
- Add the tty-probe test fixture binary and libc development dependency for pseudo-terminal integration coverage.

Documentation:
- Update credential prompting documentation to describe terminal-device probing and safe prompt injection in tests.

Tests:
- Add Unix pseudo-terminal coverage for redirected-output and no-controlling-terminal scenarios, and revise an existing test to avoid depending on the test process terminal state.

---

### DoD verification (SCR-003)

#3052's last open item — "`make gate` green; no new god-file lines, no baseline
raised" — is now settled and ticked, so the `dod / dod` gate should clear on
this re-read.

The item was waiting on one job. On `f5e5acd`, `fmt + clippy + test` reported
`success`; it carries fmt, clippy, rustdoc, the toolchain-free guards and the
test suite. Every other check on that head is green as well —
`file size ratchet`, `the platform split still compiles on Windows`,
`docs guards`, `guard self-tests`, `cargo check on the declared MSRV` and the
rest — with `dod / dod` the only red, which is the checklist itself.

The other two halves of the item hold as the issue already recorded: the eight
changed paths do not include `scripts/file-size-baseline.txt`, and none of them
is a grandfathered god file.

---

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3421,6 +3421,7 @@ dependencies = [
  "base64 0.23.1",
  "futures-util",
  "hmac",
+ "libc",
  "reqwest",
  "rpassword",
  "serde",
--- a/crates/stella-cli/src/auth_cmd.rs
+++ b/crates/stella-cli/src/auth_cmd.rs
@@ -19,7 +19,7 @@
 //! next run failed on a value the command had no way to accept.
 
 use colored::Colorize as _;
-use stella_model::credential::{ApiKey, CredentialsFile};
+use stella_model::credential::{ApiKey, CredentialPrompt, CredentialsFile, TerminalPrompt};
 use zeroize::Zeroizing;
 
 use crate::config::{AuxField, ProviderConfig, settable_aux_fields};
@@ -94,6 +94,18 @@ fn read_key_value(provider: &str, key: Option<&str>, use_stdin: bool) -> Result<
             return Err("empty key read from stdin".to_string());
         }
         return Ok(trimmed);
+    }
+    // `TerminalPrompt` is `stella_model::credential`'s own gate on this exact
+    // question: it probes what `rpassword` actually opens — `/dev/tty` on
+    // Unix, `CONIN$`/`CONOUT$` on Windows — rather than `stdin`/`stdout`.
+    // Reused here rather than re-derived so this command and the provider
+    // credential chain can never disagree about whether a human is there to
+    // answer.
+    if !TerminalPrompt.can_prompt(true) {
+        return Err(format!(
+            "no terminal available to prompt for `{provider}`'s API key — pass \
+             `stella auth set {provider} --stdin` (pipe the key) or `--key` instead"
+        ));
     }
     let value = Zeroizing::new(
         rpassword::prompt_password(format!("  API key for `{provider}`: "))
@@ -171,6 +183,14 @@ fn prompt_for_fields(
         };
         let label = format!("  {}{suffix}: ", field.prompt);
         let entered = if field.secret {
+            // Same gate as `read_key_value` above — see its comment.
+            if !TerminalPrompt.can_prompt(true) {
+                return Err(format!(
+                    "no terminal available to prompt for `{}` — pass \
+                     --field {}=VALUE instead",
+                    field.prompt, field.name
+                ));
+            }
             Zeroizing::new(
                 rpassword::prompt_password(&label)
                     .map_err(|e| format!("cannot read secret from terminal: {e}"))?,
--- a/crates/stella-model/Cargo.toml
+++ b/crates/stella-model/Cargo.toml
@@ -40,6 +40,25 @@ sha2.workspace = true
 # what this crate already does correctly.
 tempfile.workspace = true
 
+# `tests/credential_prompt_dev_tty.rs`'s fixture binary — prints
+# `TerminalPrompt.can_prompt(true)` from inside a real pseudo-terminal the
+# test harness constructs, which is the only way to observe `/dev/tty`
+# reachability (rather than `stdin`/`stdout`) without risking the whole
+# suite blocking on a real keystroke. `tests/` locate it with
+# `env!("CARGO_BIN_EXE_tty-probe")`; cargo builds it for them.
+[[bin]]
+name = "tty-probe"
+path = "tests/fixtures/tty_probe.rs"
+
+# A test fixture, never a shipped surface — keep it out of dist.
+[package.metadata.dist]
+dist = false
+
 [dev-dependencies]
 tokio = { workspace = true, features = ["rt-multi-thread", "macros"] }
 wiremock.workspace = true
+# `tests/credential_prompt_dev_tty.rs` hand-rolls its pty harness on
+# `libc::openpty`/`setsid`/`TIOCSCTTY` — the same construction
+# `stella-tui`'s `tests/deck_pty_smoke.rs` uses, already a workspace
+# dependency; integration tests only see dev-dependencies.
+libc.workspace = true
--- a/crates/stella-model/src/credential.rs
+++ b/crates/stella-model/src/credential.rs
@@ -178,10 +178,14 @@ impl ApiKey {
     /// A host that already knows how it talks to its user — a GUI, a daemon
     /// with its own secret store, a test — injects that here instead of
     /// inheriting `rpassword` on the controlling terminal. It is also the only
-    /// way to reach the swallow below deterministically: the shipping
-    /// [`TerminalPrompt`] declines under `cargo test`, where stdout is not a
-    /// terminal, so a test driving `resolve` alone exercises the gate and
-    /// never the arm underneath it (#4576).
+    /// safe way to reach the swallow below deterministically:
+    /// [`TerminalPrompt::can_prompt`] genuinely probes `/dev/tty` (Unix) or
+    /// `CONIN$`/`CONOUT$` (Windows) rather than asking whether `stdout` is
+    /// captured, so a test that drives `resolve` (not this function) with
+    /// `interactive: true` is not guaranteed to decline — on a machine with a
+    /// controlling terminal attached to the test process, it would actually
+    /// call `rpassword` and block on a keystroke. Every test exercising the
+    /// interactive step calls this function with a fake instead (#4576).
     pub fn resolve_with_prompt(
         provider_id: &str,
         env_var: &str,
--- a/crates/stella-model/src/credential/prompt.rs
+++ b/crates/stella-model/src/credential/prompt.rs
@@ -133,8 +133,11 @@ impl TerminalPrompt {
     /// only adds the check that was missing: a redirected stdout must decline
     /// exactly as a redirected stdin does, not just print a prompt nobody
     /// reads before blocking on an answer nobody can give. `can_prompt` feeds
-    /// this the [`rpassword_tty`] probes (#3052), so `can_answer`/
+    /// this the `rpassword_tty` probes (#3052), so `can_answer`/
     /// `prompt_is_visible` are the exact condition rather than a stand-in.
+    /// That module is private, so this names it without linking it: a public
+    /// item's intra-doc link to a private one resolves only under
+    /// `--document-private-items` and is a `-D warnings` error without it.
     pub fn decide(interactive: bool, can_answer: bool, prompt_is_visible: bool) -> bool {
         stella_tty::human_can_answer(interactive, can_answer, prompt_is_visible)
     }
--- a/crates/stella-model/src/credential/prompt.rs
+++ b/crates/stella-model/src/credential/prompt.rs
@@ -8,24 +8,85 @@
 //! zero-argument [`ApiKey::resolve`](super::ApiKey::resolve) uses.
 //!
 //! The seam exists because the alternative could not be tested. Under `cargo
-//! test`, `stdout().is_terminal()` is false, so [`TerminalPrompt::can_prompt`]
-//! declines and the arm below it — the one that swallows a failed prompt and
-//! degrades to [`CredentialError::NotFound`] — was never reached by the test
-//! that claimed to cover it. It was green because of the gate, not because of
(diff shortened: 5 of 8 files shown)


macanderson:
**`fmt + clippy + test` was red on `dfe6dd4`, in its `doc-warnings` step. Fixed in `f5e5acd`.**

The failure was not a test — it was rustdoc, and the cargo colour escapes hid it from a naive grep of the log (AGENTS.md's own gotcha). Stripped of ANSI, [job 101329362083](https://github.com/macanderson/stella/actions/runs/33974504384/job/101329362083) says:

```
error: public documentation for `decide` links to private item `rpassword_tty`
   --> crates/stella-model/src/credential/prompt.rs:136:20
    |
136 |     /// this the [`rpassword_tty`] probes (#3052), so `can_answer`/
    |                    ^^^^^^^^^^^^^ this item is private
    |
error: could not document `stella-model`
```

`TerminalPrompt::decide` is `pub`; `rpassword_tty` is a private `mod` in the same file. The link therefore resolves only under `--document-private-items` and is an error without it.

**The fix** drops the brackets on that one reference, keeping the code span. That is what a public item may say about a private one, and the doc now records the reason so the link is not helpfully restored by the next reader who notices it is not clickable. `#[allow(rustdoc::private_intra_doc_links)]` was the other way through and is the expedient CLAUDE.md forbids — the lint is right here.

The module header's own `[`rpassword_tty`]` link at line 19 is untouched and correct: `credential::prompt` is crate-internal, so its `//!` is not public documentation and rustdoc does not flag it. Only the one site rustdoc named was changed.

**Verified with the check that produced the error**, scoped to the crate per AGENTS.md rather than run over the workspace:

```
RUSTDOCFLAGS="-D warnings" cargo doc -p stella-model --no-deps --document-private-items
```

- with `f5e5acd` stashed: **exit 101**, reproducing both error lines above
- with it applied: **exit 0**

`python3 scripts/check-prose.py` — OK, none added.

**On #3052's definition of done.** Items 1 and 2 I checked against the diff and they hold; item 3 is `make gate` green, which is what this commit is for, so none of the three is ticked yet — the authoritative answer is this PR's CI run on `f5e5acd`.

- Item 1 (*the check reflects what `rpassword` actually does, not a stdout stand-in*) — `rpassword_tty` opens `/dev/tty` on unix and `CONIN$`/`CONOUT$` on windows, falling back to `IsTerminal` on `stdin`/`stdout` elsewhere, which matches `rpassword`'s own `wasm.rs` pair; `can_prompt` feeds those two probes to `decide`.
- Item 2 (*a test distinguishing the two named cases*) — `credential_prompt_dev_tty.rs` builds a real pty and drives the `tty-probe` fixture binary in its own process: `can_prompt_reaches_dev_tty_even_when_stdout_is_redirected_away` covers "stdout redirected, controlling terminal present", and `can_prompt_declines_with_no_controlling_terminal_at_all` covers "no controlling terminal".
- Item 3 (*`make gate` green; no new god-file lines, no baseline raised*) — the second half holds already: `scripts/file-size-baseline.txt` is not in the diff, none of the touched files is grandfathered, and `file size ratchet` is green on this PR. The first half is what was red.

`Refs #3052` rather than `Closes` — the PR body's own closing reference is unchanged.

---

Issue #2178, opened 2026-08-08
cli: STELLA_HOME moves the data tier but not user settings/credentials — isolation silently half-works


macanderson:
## Problem

`STELLA_HOME` moves the user-tier **data** (catalog, usage hub, sessions, work index) but **not** user-scope **settings**, `credentials.toml`, or the user extension dirs. Those keep resolving under `$HOME/.stella` no matter what `STELLA_HOME` says, so a process pointed at a scratch home still merges the developer's real `~/.stella/settings.json` into its settings chain.

That contradicts what the codebase documents. `crates/stella-home/src/lib.rs` (module docs, and `STELLA_HOME_ENV`'s own doc) says "`STELLA_HOME` moves the whole home; the narrower `STELLA_DATA_DIR`/`STELLA_CONFIG_DIR` still win where they always did", and AGENTS.md § "The `.stella/` directory" repeats it: "`STELLA_HOME` moves the whole home".

## Where

`crates/stella-cli/src/paths.rs`:

- `UserPaths::from_environment()` (~line 87) builds `home` from `std::env::var_os(stella_home::HOME_ENV)` — the OS `HOME` — while `data_dir` is `stella_home::data_dir()`, which *does* honour `STELLA_HOME`. The two tiers therefore diverge inside one process.
- `stella_root()` (~line 181) returns `paths.home.join(".stella")`, so every consumer of it ignores `STELLA_HOME`.
- `crates/stella-cli/src/settings.rs::user_settings_path()` (~line 931) is built on `stella_root()`, and it is the first file of `Settings::load`'s scope chain — hence the leak.

The same `stella_root()` also backs the user-scope skills/rules/tools/credentials lookups, so the blast radius is "every user-scope extension", not just settings.json.

## Reproduce

```bash
mkdir -p /tmp/scratch-home /tmp/ws/.stella
cat > /tmp/ws/.stella/settings.json <<'JSON'
{"agent_engine_config": {"default_model": "openrouter/deepseek/deepseek-v4-flash-0731"}}
JSON
cd /tmp/ws
STELLA_HOME=/tmp/scratch-home stella run --no-pipeline "read some file and describe it"
```

[...]

Issue #553, opened 2026-07-25
stella-cli: a cloned repo's `.env` is loaded into stella's process environment with no trust gate


macanderson:
### Summary

`maybe_load` applies EVERY variable name found in a repository's `.env`/`.env.local`/`.env.<mode>.local` into stella's own process environment, with no name allow/deny policy (`plan_assignments`, line 224, filters only on 'already set in the live shell'). It runs unconditionally at startup — main.rs:1137, before `Cli::parse` — and is NOT gated by `project_code_execution_trusted`. The only names clawed back afterward are `PRIVILEGED_ENV_NAMES` + the managed-enrollment credential names (stella-cli/src/enterprise_telemetry.rs:80-105, restore at main.rs:1138), which cover STELLA_*/proxy vars only. Loader, interpreter and VCS command-execution variables are neither restored nor scrubbed: `is_sensitive_env_name` (stella-tools/src/subprocess_env.rs:80-96) matches only `_API_KEY|_TOKEN|_PASSWORD|_SECRET` suffixes plus exact AWS/GCP names. This is the same threat the team already closed for `.stella/mcp.toml` (stella-cli/src/agent.rs:1307-1320: "a cloned repo's .stella/mcp.toml can name an arbitrary stdio command — RCE on `git clone && stella`") and for the FD credential handoff (credential_handoff.rs:133 validate_target is allowlist-based and explicitly rejects LD_PRELOAD, per its own test at line 215). The `.env` path has neither control.

**Severity:** critical  ·  **Dimension:** security

### Where

[...]

PR #6054, opened 2026-09-05
fix(stella-cli): add STELLA_HOME to the privileged-rollback list

### LINKED ISSUE #2499: cli: STELLA_HOME is denied but not on the privileged-rollback list, while the narrower STELLA_DATA_DIR is on both
## Problem

`stella-cli` protects the dotenv trust boundary in two layers:

1. `env_files::DENIED_EXACT` (`crates/stella-cli/src/env_files.rs:296`) refuses
   to *apply* a privileged name from a project `.env`.
2. `enterprise_telemetry::PRIVILEGED_ENV_NAMES`
   (`crates/stella-cli/src/enterprise_telemetry.rs:84`) is the **second-loader
   backstop** behind it (#553): `StartupAuthoritySnapshot::capture` snapshots
   each name's host value before the dotenv load, and
   `restore_after_project_env` (`:486`) puts it back and reports the claw-back.

`STELLA_HOME` is on layer 1 and **not** on layer 2. `STELLA_DATA_DIR` — which
redirects strictly *less* — is on both.

That asymmetry was defensible when the list was written and stopped being so in
#2178 (PR #2447). Before it, `STELLA_HOME` moved only the data tier's anchor;
`paths::stella_root()` was `$HOME/.stella` unconditionally. After it,
`STELLA_HOME` moves the **whole user tier**: `settings.json`, `credentials.toml`,
integrations, web auth, and every user-scope extension (rules, skills, custom
tools) via `paths::user_extension_home()`. The variable got broader; the list
that classifies it as privileged was not revisited.

## Why it matters

`main.rs:418` installs `UserPaths` *before* the dotenv load at `:468`, so the
CLI's own resolved paths are immune, and that is presumably why this has never
bitten. The residual exposure is the part the backstop exists for — i.e. it only
matters if layer 1 ever fails for this name (a new loader that bypasses
`is_refused`, a case-folding hole, a refactor):

- **Lazy readers.** `stella_home::stella_home()` reads the *live* environment at
  call time, not a startup snapshot. `stella_store::home::stella_home()` and
  `stella_observatory::fsview::user_config_dir()` both go through it.
- **Child processes** inherit the process environment. A nested `stella`, an MCP
  server, a lifecycle hook, or anything the bash tool spawns would resolve its
  entire user tier — trusted settings and credentials included — from the
  injected value.

This is defense-in-depth, not a live exploit: no known path sets `STELLA_HOME`
past layer 1 today. File it as the missing second layer, not as a breach.

## The guard that would have caught it

`env_files.rs:1043`'s `every_privileged_name_is_refused` asserts
`PRIVILEGED_ENV_NAMES ⊆ DENIED_EXACT`. Only that direction. The reverse is
correctly *not* an equality — `DENIED_EXACT` is much larger by design (`PATH`,
`LD_*`, `GIT_*`) — so the missing assertion is narrower than "the lists match":
**every name that redirects the user tier must be on both.**

## Definition of done

- [x] `STELLA_HOME` added to `PRIVILEGED_ENV_NAMES`.
  **#6054**: `crates/stella-cli/src/enterprise_telemetry.rs` adds `"STELLA_HOME"`
  to `PRIVILEGED_ENV_NAMES`, alongside `"HOME"` and `"USERPROFILE"` (see item 3).
- [x] A witness that fails without it. The shape that works is a named subset
  rather than an equality: enumerate the user-tier anchors
  (`stella_home::HOME_ENV`, `USERPROFILE_ENV`, `STELLA_HOME_ENV`,
  `STELLA_DATA_DIR_ENV`) and assert each is on **both** lists. Fails on `main`
  naming `STELLA_HOME`, passes after.
  **#6054**: `env_files.rs`'s new `every_user_tier_anchor_is_refused_and_rolled_back`
  is exactly that shape — it iterates those four `stella_home` constants and
  asserts `is_refused(name)` *and*
  `enterprise_telemetry::PRIVILEGED_ENV_NAMES.contains(&name)` for each, with the
  failure message naming the offending variable. The fail-on-`main` direction is
  an inference from the diff rather than an observed run: the same PR is what
  adds `STELLA_HOME` to `PRIVILEGED_ENV_NAMES`, so on `main` the second assertion
  has nothing to find and fails naming `STELLA_HOME`.
- [x] Decide `HOME` / `USERPROFILE` / `XDG_*` in the same pass and record the
  decision in a comment either way. They have the same lazy-read shape
  (`stella_home::home_dir()` reads `HOME` at call time) and are likewise denied
  but not rolled back. Restoring them is a no-op when nothing changed them, so
  "harmless but noisy" is a legitimate answer — it just should be a written one,
  since the next reader will ask this same question.
  **#6054**: decided both ways and recorded in the `PRIVILEGED_ENV_NAMES` doc
  comment. `HOME` and `USERPROFILE` are **added**, because they set the default
  `~/.stella` root and are read fresh on every call. `XDG_CONFIG_HOME` /
  `XDG_DATA_HOME` / `XDG_STATE_HOME` are **deliberately excluded**, with the
  reason stated: nothing in the trusted user tier reads them, and
  `stella_home::systemd_user_unit_dir` reads `XDG_CONFIG_HOME` only for the
  self-driving loop's per-user systemd unit path — not `~/.stella`, and not a
  privileged config or credential.

## Constraints

- `restore_after_project_env` writes the process environment and demands the
  `StartupPhase` token (#1140); it must keep running in `main`'s startup window
  between env-file loading and runtime start, never from a later lifecycle hook.
- The claw-back is *reported*, not swallowed (`main.rs:475-479` folds the
  returned names out of the load record) — a new entry inherits that and should
  be checked to surface correctly rather than silently.
- `crates/stella-cli` is a binary, so invariant #5's typed-error rule does not
  apply to any new signature here.

## Found while

Fixing #2442 (retiring `STELLA_CONFIG_DIR`, the third name in that comment
block). Not caused by it — this gap predates that change and survives it.

macanderson:
## What / why

`env_files::DENIED_EXACT` (layer 1) already refused `STELLA_HOME` in a
project `.env`. `enterprise_telemetry::PRIVILEGED_ENV_NAMES` (layer 2, the
startup rollback that restores each privileged name's host value after the
dotenv load) never listed it — the second-loader backstop behind #553 had a
gap for exactly the name #2178 widened from moving only the data-tier anchor
to moving the whole user tier (settings, credentials, skills, rules, tools).

This is defense-in-depth: `main.rs` installs `UserPaths` before the dotenv
load runs, so the CLI's own resolved paths were never exposed. The residual
risk is a lazy reader (`stella_home::stella_home()` re-reads the live
environment on every call) or a child process inheriting an env a future
loader let through despite `is_refused` — a new loader, a case-folding hole,
a refactor.

**Fix**: add `STELLA_HOME`, and `HOME`/`USERPROFILE` alongside it, to
`PRIVILEGED_ENV_NAMES`. `HOME`/`USERPROFILE` have the identical hazard:
`stella_home::stella_home()` falls back to `home_dir()` (`HOME`, then
`USERPROFILE`) whenever `STELLA_HOME` is unset, so the same lazy-read
door exists through them. `STELLA_DATA_DIR` was already on the list.

**Decision recorded** (issue's third DoD item): `XDG_CONFIG_HOME` /
`XDG_DATA_HOME` / `XDG_STATE_HOME` stay off the rollback list, with a
comment saying why. Nothing in the trusted user tier resolves through them —
`stella_home::systemd_user_unit_dir` reads `XDG_CONFIG_HOME` only for the
self-driving loop's per-user systemd unit path, never `~/.stella` — so
restoring them would be a no-op that costs a line and proves nothing.

## Witness mechanism

`env_files::tests::every_user_tier_anchor_is_refused_and_rolled_back`
(`crates/stella-cli/src/env_files.rs`) enumerates the four named anchors the
issue specifies — `stella_home::{HOME_ENV, USERPROFILE_ENV, STELLA_HOME_ENV,
STELLA_DATA_DIR_ENV}` — and asserts each is on **both** `DENIED_EXACT` (via
`is_refused`) and `enterprise_telemetry::PRIVILEGED_ENV_NAMES`. It is the
reverse direction of the pre-existing `every_privileged_name_is_refused`,
which only checked "every rollback name is refused" and never checked
"every name broad enough to move the user tier is on the rollback."

Confirmed the test fails on old code by construction: stashed only the
`enterprise_telemetry.rs` change and ran
`cargo test -p stella-cli --bin stella env_files::tests::every_user_tier_anchor_is_refused_and_rolled_back`
— it fails on `HOME` (the first anchor in iteration order) with

```
HOME moves the stella user tier and must be in PRIVILEGED_ENV_NAMES too —
a caller that ever bypasses is_refused has nothing to roll it back
```

Restored the fix and re-ran; all 16 `env_files::` tests and all 25
`enterprise_telemetry::` tests pass.

## Design note

Posted as a comment on #2499 before implementing
(https://github.com/macanderson/stella/issues/2499#issuecomment-5552898081),
re-verifying the issue's own claims against the current tree (line numbers
had drifted since the 2026-08-12 close/reopen, the mechanism had not).

No change needed to `StartupAuthoritySnapshot::capture`/
`restore_after_project_env` themselves — they already snapshot and restore
whatever `PRIVILEGED_ENV_NAMES` names, so widening that list is the whole
fix.

## DoD status (#2499)

`dod / dod` was red on `a367cce` because #2499's three checklist boxes were
still unticked when that job ran at 15:45Z. All three were checked against
this diff and ticked afterwards, each with its evidence on the issue:

- **`STELLA_HOME` on `PRIVILEGED_ENV_NAMES`** — the `enterprise_telemetry.rs`
  hunk adds it.
- **The named-subset witness** — `every_user_tier_anchor_is_refused_and_rolled_back`
  iterates exactly the four `stella_home` constants the issue names and
  asserts each on both lists.
- **The `HOME`/`USERPROFILE`/`XDG_*` decision** — recorded in the
  `PRIVILEGED_ENV_NAMES` doc comment; the first two are added, the three
  `XDG_*` names deliberately excluded with the reason stated.

No code changed for this; the gate had read an earlier state of the issue.

## Tests deleted

None.

Closes #2499

## Summary by Sourcery

Close the privileged environment rollback gap for variables that can redirect Stella's trusted user data.

Bug Fixes:
- Restore HOME, USERPROFILE, and STELLA_HOME after project dotenv loading to close the privileged user-path rollback gap.

Enhancements:
- Document which environment variables redirect trusted user paths and why XDG path variables are intentionally excluded from rollback.
- Add a regression test ensuring all user-tier path anchors are both denied and included in the rollback list.

Tests:
- Add coverage that keeps the user-tier deny list and privileged rollback list synchronized.

### DIFF
--- a/crates/stella-cli/src/enterprise_telemetry.rs
+++ b/crates/stella-cli/src/enterprise_telemetry.rs
@@ -81,7 +81,27 @@ fn store_process_free_authority(value: u8) {
 /// The names the startup rollback restores after a project dotenv file has
 /// run. `env_files` refuses every one of them outright — the rollback is the
 /// backstop behind that deny-list, and its own test holds the two in step.
+///
+/// `HOME`, `USERPROFILE`, `STELLA_HOME` and `STELLA_DATA_DIR` are the whole
+/// user-tier redirect chain in `stella_home`. `HOME` and `USERPROFILE` set
+/// the default `~/.stella` root. `STELLA_HOME` moves that root. `STELLA_DATA_DIR`
+/// moves the data tier alone and wins over it. Each one is read fresh on
+/// every call, not snapshotted once. So a name a dotenv loader missed could
+/// point trusted settings, credentials, and skills anywhere the project
+/// chose. That is why each one belongs here, the same reason
+/// `STELLA_DATA_DIR` already did.
+///
+/// `XDG_CONFIG_HOME`, `XDG_DATA_HOME`, and `XDG_STATE_HOME` are refused by
+/// `env_files` too, but they are not here. Nothing in the trusted user tier
+/// reads them. `stella_home::systemd_user_unit_dir` reads `XDG_CONFIG_HOME`
+/// only for the self-driving loop's per-user systemd unit path — not
+/// `~/.stella`, and not a privileged config or credential. Restoring it after
+/// a project dotenv load would be a no-op that costs a line and proves
+/// nothing.
 pub(crate) const PRIVILEGED_ENV_NAMES: &[&str] = &[
+    "HOME",
+    "USERPROFILE",
+    "STELLA_HOME",
     "STELLA_MANAGED_SETTINGS",
     "STELLA_DATA_DIR",
     "STELLA_TRUST_PROJECT",
--- a/crates/stella-cli/src/env_files.rs
+++ b/crates/stella-cli/src/env_files.rs
@@ -1055,6 +1055,42 @@ mod tests {
         }
     }
 
+    /// The other half of the drift guard. `every_privileged_name_is_refused`
+    /// checks one direction only: every rollback name is refused. It never
+    /// checks the other way: is every name that moves the whole user tier
+    /// also on the rollback list? `STELLA_HOME` was refused here but never
+    /// rolled back. So if any future caller ever skipped `is_refused` — a new
+    /// loader, a case-folding bug, a refactor — a cloned repository could set
+    /// `STELLA_HOME` and point the trusted user tier (settings, credentials,
+    /// skills) anywhere it liked, with nothing left to put it back.
+    ///
+    /// This checks a small named set, not the full `DENIED_EXACT` list.
+    /// `DENIED_EXACT` is much bigger (`PATH`, `LD_*`, `GIT_*`, …) and most of
+    /// those names have nothing to roll back. These four are the whole
+    /// `stella_home` user-tier chain: `HOME` and `USERPROFILE` set the
+    /// default `~/.stella` root, `STELLA_HOME` moves that root, and
+    /// `STELLA_DATA_DIR` moves the data tier alone and wins over it. Each one
+    /// must be both refused and restored.
+    #[test]
+    fn every_user_tier_anchor_is_refused_and_rolled_back() {
+        for name in [
+            stella_home::HOME_ENV,
+            stella_home::USERPROFILE_ENV,
+            stella_home::STELLA_HOME_ENV,
+            stella_home::STELLA_DATA_DIR_ENV,
+        ] {
+            assert!(
+                is_refused(name),
+                "{name} moves the stella user tier and must be in DENIED_EXACT"
+            );
+            assert!(
+                crate::enterprise_telemetry::PRIVILEGED_ENV_NAMES.contains(&name),
+                "{name} moves the stella user tier and must be in PRIVILEGED_ENV_NAMES \
+                 too — a caller that ever bypasses is_refused has nothing to roll it back"
+            );
+        }
+    }
+
     #[test]
     fn deny_list_matches_namespaces_and_case_but_spares_ordinary_names() {
         // Loader namespaces are refused wholesale, in any case.


macanderson:
## DoD sweep for #2499 — two items met, one gap

`dod / dod` is red on `a367cce` because #2499's three boxes are unticked. Checked each against the diff rather than the description:

**Box 1 — `STELLA_HOME` added to `PRIVILEGED_ENV_NAMES`.** Met. `enterprise_telemetry.rs`'s `PRIVILEGED_ENV_NAMES` now carries it, alongside `HOME` and `USERPROFILE`.

**Box 2 — a witness of the named-subset shape.** Met. `env_files.rs`'s `every_user_tier_anchor_is_refused_and_rolled_back` enumerates exactly the four constants the issue names (`stella_home::HOME_ENV`, `USERPROFILE_ENV`, `STELLA_HOME_ENV`, `STELLA_DATA_DIR_ENV`) and asserts each is on both lists. All four are present in `DENIED_EXACT` on `main`, and none of `HOME`/`USERPROFILE`/`STELLA_HOME` is in `PRIVILEGED_ENV_NAMES` there, so it fails on `main` and passes here. One deviation worth noting: the issue predicted it would fail "naming `STELLA_HOME`", but since this PR also adds `HOME` and `USERPROFILE`, the loop trips on `HOME` first. The shape the issue asked for is what shipped.

**Box 3 — decide `HOME` / `USERPROFILE` / `XDG_*` and record the decision.** A decision is recorded, but its stated reason is not accurate, and box 3 asks for a written decision the next reader can rely on.

The new doc comment on `PRIVILEGED_ENV_NAMES` says:

> `XDG_CONFIG_HOME`, `XDG_DATA_HOME`, and `XDG_STATE_HOME` are refused by `env_files` too, but they are not here. Nothing in the trusted user tier reads them. `stella_home::systemd_user_unit_dir` reads `XDG_CONFIG_HOME` only for the self-driving loop's per-user systemd unit path […]

`systemd_user_unit_dir` is not the only reader. Two others exist, and the comment names neither:

- **`XDG_DATA_HOME`** — `stella-store`'s `home.rs`, in `legacy_data_dir()`. That is the pre-`~/.stella` data dir the first-run migration reads *from*: `usage.db`, `catalog.db`, the session registry, notifications, the enterprise spool. It is a lazy `std::env::var_os` read, the same shape the issue's "Lazy readers" paragraph is about. On a machine that has not migrated yet, a name that got past layer 1 would point the migration's *source* at a project-chosen directory.
- **`XDG_STATE_HOME`** — `stella-cli`'s `paths.rs`, in `UserPaths::from_environment`, which populates `state_home`. This one has a real defence — it is resolved once at startup before the dotenv load, which is the same snapshot argument that already protects the CLI's own paths — but the comment does not make it, and the argument does not extend to a child process, which calls `from_environment` at its own startup against whatever env it inherited.

Neither is necessarily a reason to add the names. The issue explicitly blesses either answer ("Restoring them is a no-op when nothing changed them, so 'harmless but noisy' is a legitimate answer — it just should be a written one"). The gap is that the written answer rests on a universal claim two call sites contradict, so the next reader who checks it will land where this sweep did.

Smallest thing that closes box 3: rewrite that paragraph to name all three readers and give each its own reason — `XDG_CONFIG_HOME` reaches no `~/.stella` path, `XDG_STATE_HOME` is snapshotted before the load, `XDG_DATA_HOME` feeds a migration source and is judged (in or out) on its own merits. Then tick all three boxes.

Not ticking any box here: box 3 is the author's call to make, and boxes 1 and 2 alone will not turn `dod / dod` green.

Everything else on this PR is clean — no `hold` label, no unresolved review threads, no Sourcery ❌ rows, and every other completed check is green with `fmt + clippy + test` still running.

---

Issue #6070, opened 2026-09-05
The deck never says which skill a holdout turn withheld


macanderson:
Found while wiring the per-artifact holdout (#6067).

## The problem

When the holdout withholds a skill, `SessionMemory::note_turn_skills`
(`crates/stella-cli/src/memory.rs`) prints a line naming it:

```
  ! holding back skill `foo` this turn, to measure whether it helps
```

That is stderr, and the Command Deck runs under the alternate screen, which
swallows stderr entirely. `crates/stella-cli/src/command_deck/steering.rs`
says so in its own module docs, which is why the withheld-checkout notice
exists there at all.

So on the deck — the default interactive shell on a TTY — a user is told the
holdout's *fraction* once at boot and is never told which turn paid it. On
`stella run` and the other raw doors the line does show.

## The fix

Carry the per-turn line to the deck. The boot notice already goes through
`Inbound::Notice`; this one is per turn rather than per session, so it wants a
channel that lands beside the turn. The withheld-steering notice's raw-door
half (`crates/stella-cli/src/agent/output.rs`'s `open_raw_turn`) is the shape
to copy.

Weigh a per-turn notice against noise first: at the shipped rate this fires
about once in twenty turns, which is rare enough to be information rather than
chrome, but that is a judgement to make deliberately.

## Constraints

- A new `AgentEvent` variant costs a row in
  `crates/stella-protocol/src/event/consumers.rs` (a build error otherwise,
  AGENTS.md invariant 10) and a regenerated wire schema. Check whether an
  existing variant already carries this before adding one.
- `crates/stella-cli/src/command_deck.rs` is a god file closed to growth.

## Definition of done

- [ ] A deck user learns which skill a holdout turn withheld.
- [ ] The raw doors keep the line they have.
- [ ] One place composes the wording, so the two surfaces cannot drift.
- [ ] A test proves the deck receives it.

macanderson:
Triage: P2, area:cli, area:tui, use-model:pro, pain:user-experience — real UX gap (a deck user is never told which turn a skill holdout hit, even though raw doors already surface it); fixing it needs a new AgentEvent variant wired through the protocol consumer table (AGENTS.md invariant 10) and a shared wording site, which is cross-crate work rather than a single-file change.

Issue #6069, opened 2026-09-05
A memory that stops helping is retired by nothing but a deleted file path


macanderson:
Slice **D** of epic #5200. Untouched today.

## The problem

Memory retirement reads selection health, not a measured comparison.
`crates/stella-cli/src/memory/retirement.rs`'s `sweep` takes
`&[SelectionHealth]` and `retirement_reason` branches on `health.failing`.
That health is folded from citation counts — and #5989 retired
`ObservationSource::MemoryCitation` because nothing in a real turn ever
produced one.

So the only evidence source that retires a memory today is path-anchor
validation: a memory whose every named path has vanished. A memory that is
present, wrong, and quietly costing tokens is retired by nothing.
`crates/stella-parity/src/evolution.rs`'s `Memory` row says exactly this, and
that row is one of the gaps #5200 exists to close.

## The fix

Point the memory retirement sweep at the same appraisal the skill sweep uses.
`appraise` + `decide_demotion` (`crates/stella-learn/src/skills/appraisal.rs`)
already generalize: they take trials and an origin and return a verdict plus a
keep-or-demote decision, and `decide_demotion` already refuses to touch
anything hand-authored — which is the protection a memory needs most.

The citation class becomes display-only rather than an input to retirement.
`crates/stella-cli/src/memory/uses.rs` records that trace correlation cannot
establish an observable effect; that stays true and stays shown, it just stops
being evidence.

## Depends on

The artifact-keyed trial ledger, which is where a memory record's trials would
live. There are none today.

## Verify

A memory record that stops helping is demoted from real recorded turns, not
from a hand-constructed `SelectionHealth` literal — which is #5200's own
standard and #5198's complaint.

## Definition of done

- [x] Memory retirement reads the shared appraisal sweep.
      *`retire_failing_context` calls `appraisals::sweep(root,
      ArtifactKind::Memory, &origins, ...)` and hands the decisions to
      `retirement::sweep`, which now takes `&[(SkillAppraisal,
      DemotionDecision)]` (`crates/stella-cli/src/memory/learning.rs`,
      `crates/stella-cli/src/memory/retirement.rs`). Green in the
      `fmt + clippy + test` run for PR #6133.*
- [x] Protected categories are still never retired automatically, through the
      one `protection_for` predicate.
      *Unchanged, and still the only path to a retirement write in
      `retirement::sweep` and `sweep_vanished`. `decide_demotion` adds the
      origin guard on top. Witnessed by
[...]

Issue #6068, opened 2026-09-05
The trial ledger only knows skills, so memories and rules can never be measured


macanderson:
Slice **A** of epic #5200, second half. The first half landed with #5086: the
turn-end seam exists and writes a ledger.

**Scope narrowed 2026-09-05.** This issue is the ledger row alone. Giving a
memory record and a mined rule a turn-end seam of their own moved to #6096,
per SCR-004 — the row and the seams are separate designs, and the seams need a
sibling submodule beside `recall.rs`, which sits at the file-size ratchet.

## The problem

The ledger is keyed by skill, not by artifact.
`crates/stella-cli/src/memory/appraisals.rs` writes
`.stella/private/skill_trials.jsonl`, one row per skill per turn, as
`StoredTrial { skill: String, #[serde(flatten)] trial: SkillTrial }`. There is
no artifact kind on the row, and `SkillTrial`
(`crates/stella-learn/src/skills/appraisal.rs`) carries none either.

So a memory record and a mined rule have nowhere to be recorded. #5200 asks
for "one ledger, three kinds — not three ledgers", and today there is one
ledger and one kind.

This blocks two things that are otherwise ready:

- The per-artifact holdout scheduler (#6067) is kind-opaque and would serve a
  memory record or a rule unchanged. It cannot, because there is no row to
  write for one.
- The memory retirement sweep cannot read the shared appraisal (slice D,
  #6069).

## The fix

1. An artifact-keyed trial in `stella-learn`:
   `(artifact_kind, artifact_id, task, withheld, outcome, turns)`. `SkillTrial`
   becomes one shape of it, or a thin alias, so `appraise` and
   `decide_demotion` keep working unchanged — they already take trials and an
   origin and return a verdict.
2. `appraisals.rs` writes and folds rows by `(kind, id)`. Old skill-only rows
   must still deserialize: the ledger is append-only and a workspace has
   history in it. Default the kind to `skill` on a row that has none, and say
   so in the row type's docs.

## Verify

A trial recorded under one kind is invisible to a sweep of another, and a row
written before kinds existed still folds into the skill sweep.

## Constraints

- `crates/stella-learn` has no I/O and `stella-protocol` is its only workspace
  dependency. Keep it that way.

## Definition of done

- [x] One trial type carries the artifact kind, and `appraise` reads it unchanged.
      — `stella_learn::ledger::ArtifactTrial` wraps the `SkillTrial` `appraise`
[...]

PR #6073, opened 2026-09-05
feat(stella-learn): hold one skill back on a schedule, so it has a control arm

### LINKED ISSUE #6067: Slice B of the shared evidence engine: a per-artifact holdout, and telling the user what it costs
Slice **B** of epic #5200. The verdict table on that epic reads the tree at
`445d624b9` and finds slice B half-built.

## What exists

`crates/stella-cli/src/memory/recall.rs`'s `maybe_suppress_recall` is a real
holdout: on every `context.retrieval.ab_recall_rate`-th turn it suppresses
recall, the count is durable and shared across processes, and those turns
already land in the without-skill arm of the trial ledger.

## What does not

1. **It withholds the whole steering plane.** Memories, skills and records go
   at once, so a control turn can say that recall helped and cannot say which
   artifact did the helping. `crates/stella-learn/src/skills/appraisal.rs`
   wants a per-artifact arm and has no producer for one.
2. **The fraction is disclosed nowhere.** The `[ab-control]` tag
   (`crates/stella-cli/src/memory.rs`'s `AB_CONTROL_TAG`) rides an episode
   summary, which is offline data. #5200 asks for the cost to be visible:
   "a learning system that hides that cost is not one this repo should ship".

## The fix

A per-artifact holdout beside the plane one, and one place that says what it
costs.

- `crates/stella-learn/src/holdout.rs` — the schedule as plain functions over
  opaque artifact ids, the way `comparison::ArmTrials` keeps an arm opaque.
  `is_scheduled`, `ordinal`, `pick` (rotating over the sorted set so every
  artifact earns an arm), and `disclosure`.
- `recall::ab_control_turn` delegates to `is_scheduled`, so the tree holds one
  copy of "every `rate`-th turn".
- `context.retrieval.artifact_holdout_rate`, conservative by default.
- The two schedules must not collide. Both counters advance once per turn, so
  a shared count would put every holdout on a turn the plane control already
  took. The holdout claims its number only on turns that inject something.
- `SessionMemory::note_turn_skills` drops the picked skill from the injected
  set while it stays in `trigger_matched`, so `appraisals::record_turn` writes
  its control-arm row.
- Disclosure at deck boot through `crates/stella-cli/src/command_deck/steering.rs`,
  and a line on the turn it happens.

## Out of scope

Memory records and mined rules. The scheduler is kind-opaque and will serve
them unchanged, but neither has a trial ledger to be withheld into yet — that
is slice A's remaining half.

## Definition of done

Checked against #6073's diff at `5810999`, reading each file rather than the
PR description. Where an item's evidence is a test, the test body is what is
quoted; whether it *passes* is `fmt + clippy + test`'s word, and that job was
still running on this head when these were ticked — see the note under the
list.

- [x] A per-artifact holdout schedule exists as pure functions in `stella-learn`.
      **#6073**: `crates/stella-learn/src/holdout.rs` (new, 151 lines) holds
      `is_scheduled`, `ordinal`, `pick` and `disclosure`, each `#[must_use]`
      and each a plain function over `u64`/`u32`/`&[&str]` — no I/O, no state,
      nothing borrowed from the CLI. `pick` sorts a copy before indexing, so it
      turns on the set and not on the caller's ranking, and its own test
      (`the_pick_ignores_the_callers_order`) pins that.
- [x] Its rate is configurable and shipped conservative.
      **#6073**: `context.retrieval.artifact_holdout_rate` in
      `crates/stella-cli/src/settings/context.rs`, defaulting to
      `DEFAULT_ARTIFACT_HOLDOUT_RATE = 20` — one turn in twenty, twice the
      plane rate beside it. Configurability is pinned by the fixture parse
      (`artifact_holdout_rate: 9` read back), and "conservative" is pinned as a
      relation rather than a number: `assert!(d.artifact_holdout_rate >
      d.ab_recall_rate)`, so the default cannot later drift under the coarser
      control it refines.
- [x] It never fires on a turn the plane control took.
      **#6073**: `arm_artifact_holdout` returns before claiming a number when
      `rate <= 1 || self.injection_suppressed()`, and `arm_controls` runs the
      plane arm *first* so that condition is settled by the time it is read.
      The holdout also keeps its own durable counter under
      `ARTIFACT_HOLDOUT_EXPERIMENT` rather than sharing
      `AB_RECALL_EXPERIMENT`'s, which is what stops the two schedules marching
      in lockstep. Witness: `a_plane_control_turn_claims_no_holdout_number`
      asserts the plane counter reaches 4 over four turns while the holdout
      counter reaches 2.
- [x] A withheld turn writes a control-arm trial for the skill it withheld, and
      a test proves it on the production seams.
      **#6073**: `a_holdout_turn_writes_a_control_arm_trial_for_the_skill_it_withheld`
      (`crates/stella-cli/src/memory/learning/skill_lifecycle.rs`) drives
      `note_turn_skills` and `record_episode` — the production seams, not a
      stub — and asserts `trials(...) == vec[true, false]`: turn 1 injected,
      turn 2 withheld and recorded in the without-skill arm. Its doc closes the
      three other routes into that arm (plane rate `0`, steering on, one skill
      so no top-k seat to lose), so the row it observes can only have come from
      the holdout.
- [x] The fraction is disclosed to the user.
      **#6073**: `announce_holdout`
      (`crates/stella-cli/src/command_deck/steering.rs`) sends
      `stella_learn::holdout::disclosure(rate, "skill")` as an
      `Inbound::Notice` at deck boot. Witness
      `the_deck_boot_discloses_the_holdout_fraction` asserts the line names "1
      turn in 4" and "skill"; `a_workspace_with_the_holdout_off_is_told_nothing`
      is the control, so the disclosure cannot pass by always speaking. Also
      documented in `website/content/docs/configuration/settings.mdx` and
      `context-engine.mdx`.

**On the evidence.** Every item above is read out of the diff, and the three
that rest on a test quote the test body.

**The head has moved since those ticks, and they were re-checked against it.**
They were first written against `5810999`; the branch is now `edf1353`, by way
of `59b664d` (README) and `f7f9b5f`. That middle commit is a real correctness
fix and it changes what item 4 rests on, so it is worth stating rather than
leaving for a reader to discover: the first cut applied the holdout only at the
recording seam, while recall runs its own scoring pass, so a skill could be
withheld *in the ledger* while the prompt still carried it — a measurement that
says the opposite of what happened. `holdout_skill` became `apply_holdout`,
which now mutates the selection at every injection site and picks from the
loaded catalog the sites share rather than each site's own shortlist.

Re-checked on `edf1353`, all five still hold. Items 1, 2 and 5 are untouched by
that commit (`holdout.rs`, `settings/context.rs` and `steering.rs` are not in
its diff, and item 5 gains a per-turn line beside the boot disclosure). Item 3
holds and is now enforced in more places: `arm_artifact_holdout` still returns
early on `rate <= 1 || self.injection_suppressed()` leaving `holdout_ordinal`
`None`, and `apply_holdout` opens `let ordinal = self.holdout_ordinal?;`, so on
a plane-control turn all three call sites no-op. Item 4 is strengthened rather
than weakened — the ledger and the prompt now agree because both ask the same
function.

`fmt + clippy + test` did come back **red** on `f7f9b5f`, and it was not any of
these tests: `doc-warnings` runs with `-D warnings`, and a doc comment on
`note_turn_skills` still linked `[`Self::holdout_skill`]` after the rename, so
rustdoc failed with "could not document `stella-cli`". Fixed in `edf1353` by
pointing the link at `Self::apply_holdout`. The ticks above stand on the new
head once that run reports green; if it does not, reopen this rather than
trusting them.

---

macanderson:
## What

A **per-artifact holdout**: on a configured fraction of turns, one matched
skill is held back while the rest of the turn runs, so that skill gets a
control arm of its own. Slice B of epic #5200.

## Why

`crates/stella-learn/src/skills/appraisal.rs` judges a skill by comparing turns
that injected it against turns that did not. The only producer of the second
arm was the A/B recall control, which withholds the **whole** steering plane —
memories, skills and records at once. A turn like that can say recall helped;
it cannot say which artifact did the helping.

This adds the paired arm the appraisal actually wants, and tells the user what
it costs — #5200 asks for that explicitly ("a learning system that hides that
cost is not one this repo should ship").

## The design

- **`crates/stella-learn/src/holdout.rs`** — `is_scheduled`, `ordinal`, `pick`,
  `disclosure`. Pure, no I/O, and **kind-opaque**: an item is a plain id
  string, so the same schedule serves a memory record or a mined rule once
  those have a ledger. The exemplar is this crate's own
  `comparison::ArmTrials`, whose doc says the arm "stays opaque… which is
  exactly why one engine serves all of them".
- **`pick` rotates.** It walks the *sorted* eligible set one step per holdout,
  so every skill earns an arm. Picking by score order would only ever measure
  whatever ranked first, and picking by the caller's order would make the
  choice depend on a ranking that moves every turn.
- **`recall::ab_control_turn` delegates to `is_scheduled`**, so the tree holds
  one copy of "every `rate`-th turn" rather than two that can drift.
- **The two schedules cannot collide.** Both counters advance once per turn, so
  two schedules counting every turn stay in lockstep: a plane rate of 10 and a
  holdout rate of 20 would put *every* holdout on a turn where the plane had
  already withheld everything, and the per-artifact arm would carry nothing the
  plane arm did not. The holdout claims a number only on turns that inject
  something, so its counter advances over turns it could act on.
- **`context.retrieval.artifact_holdout_rate`, default 20** — twice the plane
  rate, because the two prices add up.
- **Every injection site asks `apply_holdout`.** The first cut applied the
  holdout only at the recording seam, and that was a defect: recall renders its
  *own* scoring pass, so the ledger would have written `selected: false` for a
  skill the prompt still carried — a measurement that says the opposite of what
  happened. Fixed in the second commit, with the reasoning in the method's docs.
- **The pick reads the loaded catalog**, not the shortlist in front of each
  caller. The two passes scope their domains differently — recall by the paths
  a turn touched, the recording seam by the prompt alone — so a pick over the
  shortlist could name a different skill at each site.

  The honest cost: on a holdout turn the picked skill is often not one this
  prompt matched, so the turn withholds nothing and buys no measurement. The
  alternative — rotating over the trigger-matched population — makes every armed
  turn productive but rotates over a set that changes with the prompt, so a
  skill can be systematically skipped with nothing to say so. Slow evidence
  announces itself as `Insufficient`; uneven coverage does not. #6076 carries
  the "re-arm rather than waste" idea that would get both.

  The plane control still supplies without-skill trials at its own rate, so the
  appraisal is not starved either way. The paired arm is an addition to it.

## Where the disclosure went, and why not the status line

The design pointer asked for the deck's **status line**. It went to the deck's
**boot notice** instead (`command_deck/steering.rs`, one line changed in
`command_deck.rs`, which stays a god file closed to growth). Two reasons, both
in the code:

- `crates/stella-tui/src/views/status_bar.rs` argues its own case against a new
  permanent cell: "permanent chrome has to earn its row on every frame", and it
  names a cell that was removed rather than plumbed. A holdout fraction never
  changes mid-session, which is the worst fit for a per-frame row.
- That row is covered by golden frames (`tests/snapshots/deck/`), which are
  regenerated with `BLESS=1 cargo test -p stella-tui` — a workspace test run
  this session is not permitted to make, so the diff could not be read before
  landing.

`command_deck/steering.rs` is the module that already owns "a session fact the
deck must say out loud", so the notice went beside the one that is there.

## Witness

`a_holdout_turn_writes_a_control_arm_trial_for_the_skill_it_withheld`
(`crates/stella-cli/src/memory/learning/skill_lifecycle.rs`), driven through the
production seams — `auto_create_skills`, `note_turn_skills`, `record_episode`.

It fails on the old code **by construction**. A matched skill reaches the
without-skill arm in exactly three other ways, and the fixture closes all
three: the plane rate is `0`, steering is on, and the workspace holds one
skill, so there is no top-k seat to lose. On `main` every turn in that fixture
injects the skill and the ledger reads `[true, true]`; with the holdout at
rate 2 turn 2 reads `false`.

Three more: `a_plane_control_turn_claims_no_holdout_number` pins the
non-collision against both durable counters,
`the_deck_boot_discloses_the_holdout_fraction` pins the disclosure, and
`holdout.rs`'s unit tests pin the rotation, the off switch, and the
order-independence of the pick.

## Verdict table for the epic

Posted on #5200: **C and E are done, A and B were half-built, D is untouched.**
This lands B's missing half.

Tests deleted: None.

Remaining work filed:

- #6068 — the trial ledger only knows skills, so memories and rules can never
  be measured (slice A's other half; the scheduler here already serves them).
- #6069 — a memory that stops helping is retired by nothing but a deleted file
  path (slice D).
- #6070 — the deck never says *which* skill a holdout turn withheld: the
  per-turn line is stderr, and the alternate screen swallows it.
- #6076 — most holdout turns are spent on a skill the prompt never matched.

`crates/stella-cli/src/memory/recall.rs` is now 1465 of its 1500-line ceiling.
#6068 names that as a planning constraint for the seam it asks for.

Refs #5200
Closes #6067

## Summary by Sourcery

Add per-skill scheduled holdouts so learning can measure each skill's contribution without withholding the entire steering plane.

New Features:
- Add configurable per-skill holdouts that withhold one matched skill on scheduled turns to provide artifact-specific control-arm trials.
- Expose a kind-agnostic holdout scheduler with rotating, order-independent selection and disclosure messaging.

Bug Fixes:
- Apply the same holdout decision across recall injection and trial recording so measurements reflect what the model actually received.
- Prevent artifact holdouts from colliding with plane-wide recall control turns.

Enhancements:
- Disclose the holdout cost at session startup and document the new retrieval setting and learning behavior.

Documentation:
- Document per-skill holdout behavior, configuration, scheduling, and its relationship to recall control.

Tests:
- Add scheduler unit tests and integration coverage for holdout trials, schedule separation, configuration, and startup disclosure.


macanderson:
Pushed `edf1353` to this branch: CI was red on `f7f9b5f` and the cause was a one-token doc-link break, not any of this PR's tests.

```
error: unresolved link to `Self::holdout_skill`
   --> crates/stella-cli/src/memory.rs:878:12
       the struct `SessionMemory` has no field or associated item named `holdout_skill`
error: could not document `stella-cli`
```

`f7f9b5f` renamed `holdout_skill` to `apply_holdout` but `note_turn_skills`'s doc comment kept the old name in an intra-doc link. rustdoc resolves those, and `doc-warnings` runs at `-D warnings`, so it failed the whole `fmt + clippy + test` job before `cargo test` ever ran — which is why the log shows no failing test. The fix repoints the link; `apply_holdout` is in the same `impl SessionMemory` block (opens at 768, method at 831, doc at 878), so `Self::` resolves.

While verifying #6067's DoD I re-checked all five items against this new head, since my first pass read `5810999`. They hold. Worth flagging for reviewers that `f7f9b5f` is a genuine correctness fix rather than a tidy-up: applying the holdout only at the recording seam meant recall's separate scoring pass could still inject a skill the ledger recorded as withheld — the trial would have said the opposite of what happened. Item 4 is stronger for it, and item 3's guard survives, because `apply_holdout` opens on `self.holdout_ordinal?` and `arm_artifact_holdout` still leaves that `None` on a plane-control turn, so all three call sites no-op there.

Per-item evidence is recorded on #6067.

---

Issue #6078, opened 2026-09-05
main's protection appears to no longer require up-to-date branches, so a stale-but-green PR merges without re-testing


macanderson:
## The problem

`AGENTS.md` states, in the `automerge-nudge.yml` section, that

> `main`'s protection sets `required_status_checks.strict` — head branches must be up to date with base

and the whole argument for `automerge-nudge.yml` rests on that being true: the workflow exists because auto-merge does not update branches on its own, so armed PRs deadlock at `BEHIND` when main moves. That deadlock is only possible if `strict` is on.

Today `strict` does not appear to be in force. Observed on 2026-09-05 at ~16:30 UTC, with `main` at `b6a5b726`:

| PR | `mergeable_state` | `behind_by` | `ahead_by` |
|---|---|---|---|
| #6042 | `clean` | 13 | 1 |
| #6043 | `clean` | 13 | 3 |
| #6045 | `clean` | 13 | 1 |
| #6046 | `clean` | 14 | 2 |
| #6054 | `clean` | 7 | 1 |
| #6055 | `clean` | 10 | 1 |
| #6065 | `clean` | 4 | 1 |

`behind_by` is from `GET /repos/macanderson/stella/compare/<main sha>...<pr head sha>`. With `strict` on, every one of those rows should read `behind`, not `clean` — `behind` is exactly the state GitHub returns for "required checks are strict and this head is not up to date".

**This is an inference, not a direct observation, and that matters.** I could not read `GET /repos/macanderson/stella/branches/main/protection` — it returns `403 Resource not accessible by integration` for the token I had. So the evidence is the `mergeable_state`/`behind_by` disagreement above, not the protection object itself. Someone with admin scope should settle it in one call:

```bash
gh api repos/macanderson/stella/branches/main/protection \
  --jq '.required_status_checks | {strict, contexts}'
```

If `strict` is `false`, this issue is real. If it is `true`, then the `clean`-while-behind rows are the finding instead, and the question moves to why GitHub is reporting them that way.

## Why it matters

This is the safety property that stops the composition break `AGENTS.md` already documents twice:

- the same section's own counterexample — "two individually-green PRs merged into a semantically incompatible `stella-pipeline` and broke main (#1482/#1462)" — which is given as the reason the strict rule stays;
- the shared-cell hazard behind `main-canary.yml` (`Cargo.lock`, `scripts/file-size-baseline.txt`): two branches each correct on their own, composing into a red tree.

[...]

PR #6082, opened 2026-09-05
fix(deps): re-resolve Cargo.lock after the version sync passed stella-ansi

### LINKED ISSUE #6081: main is red: lockfile-sync, compile fails on the merged tree
The post-merge canary tested `main` at `7b479bf93` and it is red.

This run did not check `7b479bf93`'s parent, so the break may be older —
check on `7b479bf93^` and walk back while it still fails.

Failing: **lockfile-sync, compile**

### `lockfile-sync`

```
check-lockfile-sync: FAIL — Cargo.lock is out of sync with the manifests.
check-lockfile-sync: 
check-lockfile-sync: cargo said:
check-lockfile-sync:          Updating crates.io index
check-lockfile-sync:      error: cannot update the lock file /home/runner/work/stella/stella/Cargo.lock because --locked was passed to prevent this
check-lockfile-sync:      help: to generate the lock file without accessing the network, remove the --locked flag and use --offline instead.
check-lockfile-sync: 
check-lockfile-sync: Fix it by regenerating the lock and committing it:
check-lockfile-sync: 
check-lockfile-sync:        cargo metadata --format-version 1 >/dev/null   # or: cargo check --workspace
check-lockfile-sync:        git add Cargo.lock
check-lockfile-sync: 
check-lockfile-sync: Cargo.lock is tracked and install.sh builds --locked, so this reaches
check-lockfile-sync: people installing from source, not only CI. Commit the regenerated lock
check-lockfile-sync: in the same PR as the manifest change that moved it (#3332).
```

### `compile`

```
error: cannot update the lock file /home/runner/work/stella/stella/Cargo.lock because --locked was passed to prevent this
help: to generate the lock file without accessing the network, remove the --locked flag and use --offline instead.
```

canary exists (#3332).

### How to fix

```sh
git checkout main && git pull
```

...then, per failing check:

#### `lockfile-sync`

```sh
cargo metadata --format-version 1 >/dev/null   # regenerates Cargo.lock
git add Cargo.lock
```

#### `compile`

```sh
cargo check --workspace --all-targets --locked
```

### Definition of done

- [x] `lockfile-sync` passes on a fresh `main` with the fix applied
- [x] `compile` passes on a fresh `main` with the fix applied
- [x] The repair landed on its own from a fresh `main`, not folded into an
      unrelated PR.

Tick these on the repair PR's own tree — each is a command you can run before
you merge. The canary closes this issue itself on its next green run, so a box
nobody ticks costs the issue nothing; SCR-003's gate is what reads them, and it
runs on the PR.

than only CI.

---

### Verification of the DoD boxes above (#6082)

Ticked against #6082's head `2373f70`, each with a run whose result can
distinguish the claim from its opposite.

**1. `lockfile-sync`** — the gate script itself, on #6082's head and then on
the tested commit as a control:

```
$ git checkout 2373f70 && ./scripts/check-lockfile-sync.sh
check-lockfile-sync: OK — Cargo.lock resolves against the manifests.
EXIT=0

$ git checkout 7b479bf93 && ./scripts/check-lockfile-sync.sh
check-lockfile-sync: FAIL — Cargo.lock is out of sync with the manifests.
CONTROL_EXIT=1
```

It fails on the commit the canary tested and passes with the one-line fix, so
the pass is attributable to the change rather than to the environment.

**2. `compile`** — the quoted `compile` failure above is not a Rust error; it
is the same `--locked` refusal, raised before any crate is built. Resolving the
lock is therefore the whole of it, and CI agrees on this head:
`cargo check on the declared MSRV` (a `--locked` check) and
`build image + smoke the container` both passed on `2373f70`.

**3. Landed on its own from a fresh `main`** — #6082 is one commit touching one
file, `+1 −1`, based on `7b479bf93`, which is the exact commit this issue names.

The lock entry itself, before and after:

```
 [[package]]
 name = "stella-ansi"
-version = "0.9.337"
+version = "0.9.340"
```

Cause, for the record: `stella-ansi` arrived in #6055 with its entry recorded at
`0.9.337`, and the release syncs #6064/#6077 moved `[workspace.package].version`
to `0.9.340` without moving the new crate's entry with it — the search-and-
replace had no `0.9.337` in its worklist. #6080 is the same hazard queued up
again and is held until this lands; I commented there with the detail.

macanderson:
## What & why

`main` is red at `7b479bf93`. The canary filed #6081 naming two failures, `lockfile-sync` and `compile`, both with one cause:

```
error: cannot update the lock file /home/runner/work/stella/stella/Cargo.lock because --locked was passed to prevent this
```

`stella-ansi` arrived in #6055 with its `Cargo.lock` entry recorded at `0.9.337` — the workspace version at the moment that branch was cut. Two release syncs (#6064, #6077) landed between that cut and the merge, so `[workspace.package].version` on `main` is now `0.9.340` while the new crate's entry alone still read `0.9.337`. The lock therefore no longer resolved against the manifests, and everything passing `--locked` broke with it.

The diff is that one line:

```diff
 [[package]]
 name = "stella-ansi"
-version = "0.9.337"
+version = "0.9.340"
```

## Neither branch was wrong

`Cargo.lock` is the shared cell AGENTS.md names: one file every PR of a shape must write. #6055 wrote it correctly for the tree it was cut from, the release syncs wrote it correctly for theirs, and the two composed into a lock that does not resolve. No pre-merge check has both halves in front of it — which is exactly what `main-canary.yml` exists to catch, and it did.

## Verification

Both checks the canary named, run on this branch:

```
$ ./scripts/check-lockfile-sync.sh
check-lockfile-sync: OK — Cargo.lock resolves against the manifests.

$ cargo metadata --format-version 1 --locked >/dev/null; echo $?
0
```

Both fail on `7b479bf93` without this commit — I reproduced the failure on a fresh checkout of `main` before regenerating, and `check-lockfile-sync.sh` printed the same remedy text the canary quoted.

The lock was regenerated with the command the issue prescribes (`cargo metadata --format-version 1`), not hand-edited.

## Definition of done (#6081)

- [x] `lockfile-sync` passes on a fresh `main` with the fix applied — output above.
- [x] `compile` passes on a fresh `main` with the fix applied — the `compile` failure was the same `--locked` refusal, and `cargo metadata --locked` now exits 0. CI's own `compile` step on this PR is the authoritative run.
- [x] The repair landed on its own from a fresh `main`, not folded into an unrelated PR — this branch was cut from `7b479bf93` and carries one commit touching one file.

## Tests deleted

None.

## Note on how this arose

I merged the batch of green PRs that composed into this, during an automated PR sweep. The systemic half — that `main`'s protection appears not to require up-to-date branches, so a stale-but-green PR merges without re-testing against current `main` — is filed separately as #6078, since it outlives this one-line repair.

Closes #6081

---

## DoD boxes ticked on #6081 — independently re-verified

The `dod` check reads the **issue's** checklist, not this description's, and #6081's three boxes were still unticked, so the check failed here. I re-ran each item against this head (`2373f70`) before ticking, with a control so the result can distinguish the claim from its opposite:

```
$ git checkout 2373f70 && ./scripts/check-lockfile-sync.sh
check-lockfile-sync: OK — Cargo.lock resolves against the manifests.
EXIT=0

$ git checkout 7b479bf93 && ./scripts/check-lockfile-sync.sh
check-lockfile-sync: FAIL — Cargo.lock is out of sync with the manifests.
CONTROL_EXIT=1
```

Fails on the commit the canary tested, passes with the one-line fix. On the `compile` half, the quoted failure is not a Rust error — it is the same `--locked` refusal raised before any crate builds — and CI on this head agrees: `cargo check on the declared MSRV` (a `--locked` check) and `build image + smoke the container` both passed.

The full evidence is recorded on #6081 itself.

## Follow-on: #6080 is the same hazard, queued

`bot/version-sync` (#6080) was cut from `6ca1cfec`, which predates `stella-ansi` entirely — verified: the crate is in neither `crates/` nor `Cargo.lock` at that ref. Its sync therefore bumped 26 members to `0.9.341` and could not bump the 27th. Its own `--locked` checks pass because its tree is self-consistent; the break would appear only on composition, exactly as this one did. Merging it after this PR would strand `stella-ansi` at `0.9.340` and re-red `main`.

I re-synced that branch onto current `main` rather than let its auto-merge fire on a green that predates #6081, and commented there with the detail. It is not blocked on anything in this PR beyond ordering.

---

https://claude.ai/code/session_01Rqjm6N2eJwNpsnZFDPrn3R

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3184,7 +3184,7 @@ checksum = "a2eb9349b6444b326872e140eb1cf5e7c522154d69e7a0ffb0fb81c06b37543f"
 
 [[package]]
 name = "stella-ansi"
-version = "0.9.337"
+version = "0.9.340"
 
 [[package]]
 name = "stella-autonomy"


macanderson:
Independently confirmed against `origin/main`, and — the part the diff itself cannot show — confirmed **complete**, not merely correct.

The diagnosis holds: `[workspace.package].version` is `0.9.340`, `stella-ansi`'s lock entry is `0.9.337`, and every other workspace crate is already `0.9.340`.

The question a one-line lock fix has to answer is whether it is the *only* stale entry, since a second one leaves `main` red for another full cycle. Sweeping every workspace member's lock entry against `0.9.340`:

```
loop-bench    = 0.1.0
request-bench = 0.1.0
stella-ansi   = 0.9.337
```

The two bench members are not drift — `bench/loop-bench/Cargo.toml` and `bench/request-bench/Cargo.toml` each carry a literal `version = "0.1.0"` rather than inheriting the workspace version, so their entries are correct as they stand. That leaves `stella-ansi` as the sole stale entry, with 27 `stella-*` crates already at `0.9.340`. This one line is the whole repair.

I found `main` red during an autonomous PR sweep and stopped short of writing my own fix once I saw this PR carrying `unblocks-main` — three sessions duplicated the same repair on 2026-08-24, and AGENTS.md is right that the open PR is the stronger signal than a claim comment. Nothing further from me on it; flagging only so a reviewer knows the completeness check has been done by someone other than the author.

---

Issue #2640, opened 2026-08-10
pipeline: the regression veto's withheld credit still reports the generic UNVERIFIED shortfall


macanderson:
## Problem

`unverified_evidence` (`crates/stella-pipeline/src/verify.rs`) builds a
`shortfall` sentence by walking the reasons a flipped candidate failed to reach
`SubmitFast`, and falls through to a generic string when none of its arms
matched:

    "the deterministic checks did not combine into a pass"

That sentence names no channel and no finding, so a reader with only the verdict
cannot tell which check objected — they have to open the ladder snapshot.

The PR for #2607 closed two of these: a measured non-overlap now reports
`diff_coverage=not_covered (...)` and a tautological witness now reports
`witness_mutation=tautological (...)`, each in the guard's own words via
`DiffCoverage::explain` / `MutationAudit::explain`.

**The diagnostics regression veto (#861) was left on the generic sentence.** A
candidate that flipped its witness, kept the touched tests green, stayed within
the diff budget, and was withheld *solely* because it introduced new
lint/typecheck errors — or new warnings under `diagnostics_veto_warnings` —
reports the same content-free line. That is the channel most likely to be
actionable, because the fix is named in the diagnostic itself.

## Where

- `crates/stella-pipeline/src/verify.rs` — `unverified_evidence`, the
  `else if` chain ending in the generic string. The two guard arms added for
  #2607 are the shape to copy.
- `LadderInputs::new_diag_errors` / `new_diag_warnings` / `veto_warnings`
  (same file) are the inputs to branch on; `ladder_decision`'s rung 4 has the
  exact conjunct.
- Neither `new_diag_errors` nor `new_diag_warnings` has an `explain()`-style
  renderer today — unlike `DiffCoverage` and `MutationAudit`, they are plain
  counts. Whether to add one or to format inline is the open design call; the
  existing arms format inline for `diff_lines`/`diff_budget`, so inline is the
  cheaper precedent.

## Verify

A witness of the same shape as the two in
`crates/stella-pipeline/src/pipeline/tests/verification_hardening/guard_wiring.rs`:
drive a run whose flip is real and whose only defect is a fresh diagnostic
error, then assert the verdict summary names the diagnostics channel rather
than the generic sentence. `a_fresh_diagnostic_error_vetoes_the_fast_submit`
in `verification_hardening.rs` already builds that scenario and can be
extended or copied.

## Constraints

[...]

PR #6085, opened 2026-09-05
docs(scr): record the Refs-only state in SCR-003

macanderson:
Keeps this specification byte-identical with `macanderson/oxagen`, where the merge gate that implements it has changed.

The gate now accepts a third state: a pull request that references an issue with `Refs #N` **without** claiming to close it passes without being held against that issue's checklist — because it is not saying the work is done. A pull request carrying both is gated only on what it closes.

The specification requires every copy to stay identical across the org repositories, so this lands the same text here. It is documentation only — no workflow, script or behaviour in this repository changes.

Synced from macanderson/oxagen#2663.

## Summary by Sourcery

Document the Refs-only pull request state in SCR-003 so the specification reflects the current merge-gate behavior.

Enhancements:
- Update SCR-003 to document that non-closing `Refs #N` references are not gated by the referenced issue’s DoD, while closing claims remain subject to verification.

Documentation:
- Clarify SCR-003 behavior for mixed `Closes` and `Refs` references, including the limitation around negated closing language.

### DIFF
--- a/docs/scr/SCR-003-dod-verified-close.md
+++ b/docs/scr/SCR-003-dod-verified-close.md
@@ -50,16 +50,16 @@ filing an issue would be pure ceremony, never to skip a DoD that should have
 been written.
 
 A pull request that advances an issue without finishing it links that issue
-with `Refs #N` instead of `Closes #N`. `Refs` deliberately does not close, so
-the merge gate does not hold that PR against `#N`'s DoD — there is nothing
-here for it to verify, because the PR is not claiming the work is done. A PR
-may carry both: `Closes #A` closes and is gated on `#A`'s DoD; `Refs #B`
-beside it is recorded as a reference and never enforced. A PR whose body
-disclaims a close in prose ("this PR does not close #N") is read the same way
-as a bare mention — it is not a claim to close, and is not gated on `#N`
-either, even though GitHub's own closing-keyword parser has no notion of
-negation and may still close `#N` on merge regardless of the "not". That is a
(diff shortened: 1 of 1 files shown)


macanderson:
Automated PR sweep — pushed one commit here (`4e9a761`).

**Fixed.** `gate steps no other workflow runs` was red: `check-prose.py` rejected `deliberately` in the paragraph this PR adds to SCR-003 (`filler-adverb: 0 allowed, 1 found`, line 53). Deleting the word costs the reader nothing — the reason is already in the clause after it. Because SCR-003 must stay byte-identical across the org repos, the same edit landed in all five copies (macanderson/oxagen#2663 plus the three sibling syncs); all five now resolve to blob `39bcee024695fd5a43321df1a6af2cb879630224`. That check is now [green](https://github.com/macanderson/stella/actions/runs/33986679071/job/101361423159).

**Still red — needs your call.** `dod` fails with "This PR links no issue". This is a doc-only spec sync that closes nothing. It cannot use `Refs #2640` either: the Refs-only passing state is exactly what oxagen#2663 adds, and this repo's `dod-check.yml` pins that implementation at `acf3509c`, which predates it. The `no-issue` label is the applicable waiver today (SCR-003: "changes with no meaningful DoD — a typo, a pinned-digest bump, a revert"), but applying a policy waiver is a judgement rather than a CI fix, so I have left it to you.

Full DoD audit of the source change is on macanderson/oxagen#2663.

---

Issue #6088, opened 2026-09-05
The models dialog names one role when the session has more


macanderson:
## The problem

`/models` answers "what will each role of this session run". It answers for one
role. `crates/stella-cli/src/config_wiring.rs`'s `resolve` returns exactly one
`RoleWiring`, named `default`, and `deck_rows` hands that single row to the
deck as the dialog's whole content.

A session with a plugin installed has more roles than that. The plugin declares
them, `crates/stella-cli/src/agent/seats.rs`'s `installed_seats` reads them, and
each one can carry its own model through `agent_engine_config.seat_models`. The
`/models` dialog says nothing about any of it, so the one surface whose job is
"which model runs what" is silent about every role but one.

Several comments in `stella-cli` already predicted this work and named the old
slice issue for it:

- `crates/stella-cli/src/config_wiring.rs` — the module doc, `DEFAULT_ROLE`,
  and `RoleWiring::role`
- `crates/stella-cli/src/agent/engine.rs`
- `crates/stella-cli/src/engine_config.rs` (`state_from_settings`)
- `crates/stella-cli/src/profile.rs`
- `crates/stella-cli/src/command_deck/model_cmd.rs`

They were repointed here when #3909 closed without doing it. #3909 put the
plugin seats on the SETTINGS tab's SEATS pane instead, composing them in the
deck (`crates/stella-tui/src/views/seats.rs`'s `rows`) rather than folding them
into `RoleWiring`.

## What to do

Decide first whether `resolve` should return a row per declared seat, or
whether the dialog should compose the two lists the way the SEATS pane does.
The tradeoff:

- Folding seats into `RoleWiring` gives `stella config` and `/models` one
  answer with no second fold, and it is what the existing comments expect. It
  also means `resolve` has to read the plugin roster, which it does not do
  today, and `config_wiring` is the module whose whole point is mirroring the
  request path's precedence.
- Composing in the dialog keeps `resolve` a pure function of the settings and
  the session pin, and matches what `views/seats.rs` already does.

Whichever way it goes, `stella config` and the `/models` dialog must give the
same answer — that is what `config_wiring`'s module doc exists to hold.

`crates/stella-tui/src/views/models_card.rs` already prints a role it has no
word for under its own key, so a seat key needs no new copy decision there.

## How to verify

Install a plugin declaring a role, assign it a model, and run `stella config`
and `/models`. Both name the seat and the model it runs on. A golden frame over
`card_models` pins the dialog.

## Definition of done

[...]

Issue #6087, opened 2026-09-05
The models dialog still names five roles the engine can never send


macanderson:
## The problem

`crates/stella-tui/src/views/models_card.rs` carries a `SLOTS` table that maps
five settings keys to the words the `/models` dialog prints:

```text
("triage", "think"), ("research", "research"), ("plan", "plan"),
("worker", "work"), ("verifier", "verify")
```

and `slot_of` matches three of the same strings.

None of those five keys can arrive any more. The driver builds the dialog's
rows in `crates/stella-cli/src/config_wiring.rs`, and `resolve` returns exactly
one row, named `default` — #3908 collapsed the rest. So the table names five
roles the deck will never be handed, and a reader of that file learns a
vocabulary the product does not have.

Epic #3903 wants no role name compiled into `stella-tui` at all. This is the
part of that goal a settings change can reach.

## What is not in scope

`crates/stella-tui/src/deck.rs`'s `PipelineRole` keeps `Triage`, `Worker` and
`Verifier`. That enum is fed from `stella_protocol::ModelCallRole` on live
events — receipt vocabulary, which groups spend by job — and it is what the
status bar and the model picker read. Retiring it is #3910's work, after
#3906.

## What to do

- Drop the four dead rows from `SLOTS`, keeping `("default", "lead")`.
- Drop the dead arms from `slot_of`, or delete the function if nothing is left
  for it to answer. `standing` is its only caller; if every row now returns
  `None`, decide whether the standing column still earns its place and say so
  in the diff.
- Update the module's "Copy law (D6)" section, which describes three pipeline
  slots that no longer exist.

## How to verify

- `rg -n 'triage|verifier|worker|research' crates/stella-tui/src/views/` names
  nothing outside a comment about the removal.
- The `card_models` golden
  (`crates/stella-tui/tests/snapshots/deck/card_models.txt`) is unchanged,
  because the fixture's driver only ever sends `default`. If it does move, the
  dead rows were not dead and this issue is wrong — say so rather than blessing
  it.

## Definition of done

- [x] `SLOTS` names only roles the driver can send.
- [x] `slot_of` carries no arm for a role the driver cannot send.
- [x] The module doc describes the vocabulary the file actually has.
- [x] The `card_models` golden is unchanged, or the change is explained.

---

**Verification (PR #6115 → #6116), by the autonomous PR routine.**

- `SLOTS` is replaced by `KNOWN: [(&str, &str); 1] = [("default", "lead")]`.
[...]

Issue #6086, opened 2026-09-05
The seats pane shows a plugin's model but cannot change it


macanderson:
## The problem

The SETTINGS tab's SEATS pane lists every role an installed plugin declares and
the model each one runs on. It cannot change any of them. The header says
`read-only`, and `e` on that pane is claimed and does nothing
(`crates/stella-tui/src/views/settings.rs`, `handle_key`).

So the only way to give a plugin's role its own model is to hand-edit
`agent_engine_config.seat_models` in `.stella/settings.json` or
`~/.stella/settings.json`. The deck shows the setting and refuses to write it.

## What to build

An editor on the SEATS pane, following the AGENTS pane's contract in
`crates/stella-tui/src/views/engine_panel/keys.rs`:

- `e` focuses the pane and it becomes modal, claiming every key until `Esc`.
- `⏎` on a row opens the model picker (`crates/stella-tui/src/views/picker.rs`),
  filtered by `allowed_models` the same way the AGENTS pane's is.
- `x` clears a row back to the default model.
- `s` sends the working copy to the driver, which merges `seat_models` into the
  chosen scope's `settings.json` and answers with a fresh snapshot.

The driver half already exists in shape: `WorkspaceInput::EngineConfigSave`
carries the whole `agent_engine_config` object and
`crates/stella-cli/src/engine_config.rs` writes it. Check whether `seat_models`
survives that round trip before writing any UI.

## Constraints

- `crates/stella-tui/src/deck_ui.rs` is a god file (AGENTS.md § "God files")
  and is closed to growth. Pane keys live with the pane, the way
  `views/settings.rs::handle_key` already does.
- Seat keys are `<plugin-id>/<role>` and are compared whole, never split
  (`doc:roleless-core` §8.4).
- A seat with no assignment stays `None` in the snapshot. Do not write the
  default model into it — an unassigned seat must stay different from one
  someone pinned to the same model.

## How to verify

- Install a plugin declaring a role, open SETTINGS → SEATS, assign a model,
  save, and read `seat_models` back out of `.stella/settings.json`.
- A golden frame with the picker open over the pane
  (`crates/stella-tui/tests/deck_render_snapshots.rs`).

## Definition of done

- [x] The SEATS pane accepts `e` and becomes a modal editor.
- [x] Assigning a model to a seat writes `agent_engine_config.seat_models` in
      the chosen settings scope.
- [x] Clearing a seat removes its key rather than writing the default model
      into it.
[...]

Issue #6061, opened 2026-09-05
Keep the retired pipeline model keys out of live code


macanderson:
Part of #3903 (roleless core), slice 4. Spec: `doc:roleless-core` §6.

## Problem

#3908 retired five `agent_engine_config` keys — `pipeline_worker_model`,
`pipeline_verifier_model`, `pipeline_triage_model`, `pipeline_research_model`,
`pipeline_plan_model` — and the `agents.<persona>` blocks beside them. The
schema no longer has them, and `crates/stella-cli/src/settings/unknown.rs`
names each one back to the operator with the `[seats]` assignment that
replaces it.

Three things did not travel with that change.

**The self-tuning ledger still writes a retired key.**
`crates/stella-cli/src/memory/self_tuning.rs` records every promotion under
`agent_engine_config.agents.worker.effort`, while `promote` writes
`agents.default.effort`. So the ledger names a key the settings file does not
have. An operator who reads the record and undoes the change by hand edits a
key that reads nothing, and the pin stays in force.

**Live code still spells the keys, and one shipped surface still draws them.**
`crates/stella-tui/src/scenario.rs` is the demo session behind `stella deck`
screenshots and the `/models` golden frames. Its `demo_engine_config` builds
six role rows and sources two of them from `pipeline_verifier_model` and
`agents.triage.model`; its own doc comment describes "the three ways a role
gets its model", a resolution ladder the product no longer has. The golden
`crates/stella-tui/tests/snapshots/deck/card_models.txt` pins the picture:
`from pipeline_verifier_model` is drawn in the committed frame. The live
producer, `config_wiring::deck_rows`, has emitted one row named `default`
since #3908.

Nothing stops a retired key coming back, either. `scripts/check-role-names.sh`
holds four languages to one *spelling* of the role words; it never asks
whether Rust still uses them. A later change can re-add
`"pipeline_verifier_model"` as a live string and get inert configuration that
reads like a capability, which is the exact defect #3908 existed to end.

**Doc comments on live functions still describe the retired keys as governing.**
Each of these is read by someone deciding which key to set:

- `crates/stella-cli/src/agent/engine.rs` — `EngineWiring::worker_model` says
  the field is "the worker's own `pipeline_worker_model`/`agents.worker.*` pin
  when one is configured"; `reflection_route`'s doc says the triage pin
  "is already the posture's declaration of the cheap, fast model" and that
  `agents.triage.reasoning` / `agents.triage.effort` "are the settings that
[...]

PR #6095, opened 2026-09-05
feat(stella-tui): the seats pane lists the session's own role, not just the plugins'

### LINKED ISSUE #3909: roleless-core slice 5: the settings AGENTS pane becomes a live seat list, not six compiled-in personas
Slice 5 of #3903. Spec: `doc:roleless-core` §6. Follows slice 4.

## Problem

interactive mode's `SETTINGS` → `AGENTS` pane renders six compiled-in persona tabs —
`GLOBAL`, `default`, `worker`, `verifier`, `triage`, `research`, `plan` — from
`EngineAgentKind::ALL` via `crates/stella-tui/src/views/engine.rs`. Those
personas do not exist in the core loop (see slice 4), so the pane offers model
settings that steer nothing, under role names core is not supposed to know.

## What to do

Make the pane render **rows from the live session**, exactly as the sibling
`TOOLS` pane already does:

- one row for the default model,
- one row per seat declared by an installed plugin, showing the model assigned
  to it or `default`.

A session with no plugins installed shows one row.

## The pattern to copy

`crates/stella-tui/src/views/tools.rs`. Its module doc states the reason
verbatim, and it is the same reason here:

> **Stella ships with every tool on.** This panel is how they go off, and it is
> the only surface that can show an operator what they actually have: MCP tools
> and customer-registered custom tools exist nowhere but the assembled session
> stack, so the rows come from the driver
> ([`crate::envelope::Inbound::ToolPolicy`]), never from a compiled-in table.

Plugin seats are in exactly that position. The driver owns the settings files
and pushes snapshots; the panel accumulates unsaved edits and sends back only
the keys it changed.

## Constraints

- `crates/stella-tui/src/views/engine.rs` is a **god file** (`AGENTS.md` §"God
  files"): it is closed to growth. New logic lands in a sibling submodule.
- Golden frames (`crates/stella-tui/tests/deck_render_snapshots.rs`) will move.
  Regenerate with `BLESS=1` and **read the diff** — a golden blessed without
  looking is a changelog, not a test.

## Witness

A golden frame with a plugin installed shows that plugin's seat; with none
installed, the pane shows only the default model.

## Done when

The SETTINGS tab's seat list comes from the live session, and no list of role
names decides what it draws.

## Definition of done

The boxes below split the "no role name is compiled into `stella-tui`" line
this issue opened with, because a settings change cannot make the whole of it
true and a box nobody can tick is not a done condition. What is left of it is
filed: #6087 for the five dead role words in the `/models` card, and #3910 for
`deck::PipelineRole`, which is live receipt vocabulary read off the wire rather
than a settings role.

- [x] The SETTINGS tab's seat list is built from the live session: the session's
      own resolved role, then one row per role an installed plugin declares.
      **Done in #6095** — `views::seats::rows(&EngineConfigState)` is a pure fold
      returning `state.roles` (what the driver resolved) followed by
      `state.seats` (what plugins declared), and `views/settings.rs` calls it
      instead of handing the painter `state.seats`. Witness:
      `the_session_role_leads_the_plugin_seats` asserts exactly two rows in that
      order for one resolved role plus one `acme/reviewer` seat.
- [x] A session with no plugin installed shows exactly one row.
      **Done in #6095** — `a_session_with_no_plugin_still_shows_its_own_role`.
      `an_empty_snapshot_invents_no_row` guards the other side: a driver that
      sent nothing gets the hint line, not a fabricated `default` row.
- [x] No list of role names decides which rows that pane draws — a role the
      deck has never heard of renders under its own name.
      **Done in #6095** — `the_leading_row_is_named_by_the_driver_not_by_this_pane`
      feeds a driver that calls the session's role `lead` and asserts the pane's
      keys are `["lead"]`. Verified against the branch rather than the
      description: the only role-shaped literal in `views/seats.rs` is
      `const UNASSIGNED: &str = "default"`, and every use of it is the *model*
      column's fallback (`row.model.as_deref().unwrap_or(UNASSIGNED)`), never a
      row key. The five dead words in `views/models_card.rs` are still there and
      are #6087's, which is what this DoD's preamble already carved out.
- [x] A golden frame pins the pane with one plugin-declared seat on screen, and
      its diff was read. **Done in #6095**, commit `83e7d45` —
      `crates/stella-tui/tests/snapshots/deck/pane_settings_seats.txt`. The
      snapshot was missing from the PR's first push (the authoring machine could
      not run cargo); it was generated with
      `BLESS=1 cargo test -p stella-tui --test deck_render_snapshots settings_seats`
      and **read**, not blessed. The frame shows `seats · 2 · read-only` and two
      rows in the order the fold promises:

      ```
       default         zai/glm-5.2-air   from default_model
       acme/reviewer   default           from acme
      ```

      That is the same claim `the_session_role_leads_the_plugin_seats` and
      `the_fold_leaves_an_unassigned_seat_unassigned` make in the unit tests,
      pinned at the character grid: the session's own resolved role leads and
      carries its resolved model, and the plugin's unassigned seat renders
      `default` in the *model* column while keeping its own name as the row key.
      None of the five role words #3908 retired appear anywhere in the frame,
      which is what makes this fixture safe to commit where `base_ui`'s would
      not have been.

macanderson:
## What and why

Slice 5 of the roleless-core plan (#3903).

The SETTINGS tab's SEATS pane drew one row per role an installed plugin
declares, and nothing else. Two consequences on screen:

- A fresh install has no plugin, so the pane showed a hint line and no rows at
  all — the one surface whose subject is "which model runs what" had nothing to
  say about a session that runs on one model.
- Even with a plugin on, every unassigned seat rendered the word `default` and
  the pane never said what `default` resolves to.

The issue asks for "one row for the default model, one row per seat declared by
an installed plugin". This adds the first half.

`crates/stella-tui/src/views/seats.rs` gains `rows(&EngineConfigState)`, a pure
fold that returns the pane's list: the roles the driver **resolved**
(`EngineConfigState::roles`, an open vocabulary since #3472 — today exactly one,
`default`, with the model it resolved to and the settings key that chose it),
then one row per **declared** seat. `views/settings.rs` calls the fold instead
of handing the painter `state.seats` directly.

No new type. A `RoleWiringRow` maps onto the `SeatRow` shape the painter already
draws: name, model, source. The pane still names nothing — both halves arrive
from the driver, which is the same contract `views/tools.rs` holds for MCP and
custom tools.

## Witness mechanism

`the_session_role_leads_the_plugin_seats`
(`crates/stella-tui/src/views/seats.rs`) builds a snapshot with one resolved
role and one plugin-declared `acme/reviewer`, and asserts the fold returns
exactly two rows in that order, then that the painted pane shows the resolved
model and `from default_model`. It fails on `main` by construction: `rows` does
not exist there, and the pane was handed `state.seats`, so the list was one row
long and `default` appeared nowhere.

Beside it:

- `a_session_with_no_plugin_still_shows_its_own_role` — the fresh-install case,
  which was the hint line before.
- `the_leading_row_is_named_by_the_driver_not_by_this_pane` — a driver that
  calls the session's role `lead` gets `lead` back. This is what proves the row
  is read rather than written here, and it fails the day someone adds a list of
  role names to the pane.
- `an_empty_snapshot_invents_no_row` — a driver that sent nothing still gets the
  hint, not a fabricated `default` row.
- `the_fold_leaves_an_unassigned_seat_unassigned` — the fold must not fill an
  unassigned seat with the leading row's model, or a pinned seat and an
  inherited one become indistinguishable.

**Golden frame.** `crates/stella-tui/tests/deck_render_snapshots/settings_seats.rs`
renders the pane at the canonical viewport with one plugin seat on screen, and
`crates/stella-tui/tests/snapshots/deck/pane_settings_seats.txt` pins it. The
frame was read, not blessed: two rows, the resolved `default` carrying
`zai/glm-5.2-air` and `from default_model`, then `acme/reviewer` reading
`default` in the model column and `from acme`. That is the same claim the two
unit tests above make, held at the character grid.

It is a submodule rather than more lines in `deck_render_snapshots.rs`, which
sits close to the 1500-line ceiling — the shape `voice.rs` and `mcp.rs` already
use. Its fixture is its own rather than `base_ui`'s, because the demo scenario's
engine config still carries the five roles #3908 retired (#6061) and a committed
golden of this pane would then teach a vocabulary the product does not have.

## Exemplar

`crates/stella-tui/src/views/tools.rs` — rows from the assembled session, never
from a compiled-in table. Its module doc states the reason and the issue quotes
it.

## Where this departs from the issue, and why

The issue's stated done condition was "No role name is compiled into
`stella-tui`". That is not true today and this change cannot make it true:

- `crates/stella-tui/src/views/models_card.rs`'s `SLOTS` table still names
  `triage`, `research`, `plan`, `worker` and `verifier`. Dead vocabulary — the
  driver only ever sends `default` — but present. Filed as #6087.
- `crates/stella-tui/src/deck.rs`'s `PipelineRole` names three of the same
  words for a live purpose: it groups spend by the `ModelCallRole` the wire
  reports. That is receipt vocabulary rather than routing, and #3910 owns it.

Ticking that box would be claiming something the tree contradicts, so the
issue's checklist was split into what this slice proves, with a comment on
#3909 saying so and both remainders filed. The rest of the issue is unchanged.

## Stale comments repointed

Seven comments in `stella-cli` and `stella-tui` named #3909 as the issue that
would fold plugin seats into `RoleWiring` and into `EngineRole`. This did
neither — the seats stay a separate list the deck composes — so each comment now
says what actually happened, and the ones predicting a `/models` change point at
#6088 instead.

## Tests deleted

None.

## Remaining work

Filed as issues rather than left in this description:

- #6086 — the SEATS pane is read-only; assigning a model to a seat writes
  `agent_engine_config.seat_models` and nothing does it.
- #6087 — the `/models` card's five dead role words.
- #6088 — `/models` and `stella config` name one role when the session has more.

Closes #3909

## Summary by Sourcery

Display the session's resolved role alongside plugin-declared seats in the SETTINGS SEATS pane.

New Features:
- Show the session's driver-resolved roles in the SETTINGS SEATS pane before plugin-declared seats, including each role's resolved model and source.

Bug Fixes:
- Ensure fresh sessions without plugins still display their active model instead of an empty seat list.
- Preserve driver-provided role names and leave unassigned plugin seats rendered as inherited defaults.

Enhancements:
- Centralize SEATS row assembly in a driver-backed fold while keeping the pane role-agnostic and read-only.

Documentation:
- Update SEATS pane documentation and stale issue references to reflect that resolved roles and plugin seats remain separate data.

Tests:
- Add coverage for row ordering, plugin-free sessions, driver-defined role names, empty snapshots, and unassigned seats.
- Add a golden-render snapshot for the SETTINGS SEATS pane.

### DIFF
--- a/crates/stella-cli/src/agent/engine.rs
+++ b/crates/stella-cli/src/agent/engine.rs
@@ -23,7 +23,7 @@ use super::*;
 /// conflated: a call routed to a smaller-context or non-reasoning model still
 /// got the session model's clamps, which is the wire-shape/400 class #273
 /// exists to warn about. The role that used to diverge here was the worker's;
-/// the ones that will diverge again are plugin-declared seats (#3909), which
+/// the ones that will diverge again are plugin-declared seats, which
 /// resolve their own model through [`crate::agent::seats`].
 fn tuned_engine_config(cfg: &Config, catalog_ref: (&str, &str)) -> EngineConfig {
     let (provider_id, model_id) = catalog_ref;
--- a/crates/stella-cli/src/command_deck/model_cmd.rs
+++ b/crates/stella-cli/src/command_deck/model_cmd.rs
@@ -86,7 +86,7 @@ pub fn configured_role_pins(
     // both: settings can no longer express a model for either, so publishing a
     // pin for them would be the deck reporting an intent nothing recorded.
     // They now render unconfigured, which is what they are — and what a
-    // plugin's declared seat will replace in #3909.
+    // plugin's declared seat will replace in #6088.
     [PipelineRole::Worker]
         .into_iter()
         .map(|slot| {
--- a/crates/stella-cli/src/config_wiring.rs
+++ b/crates/stella-cli/src/config_wiring.rs
@@ -26,8 +26,8 @@
 //! product to be wrong.
 //!
 //! The shape stays a `Vec` rather than collapsing to one struct because the
-//! rows come back in slice 5 (#3909) as **plugin-declared seats**, resolved
-//! the same way and rendered by the same `/models` dialog.
+//! rows come back as **plugin-declared seats**, resolved the same way and
+//! rendered by the same `/models` dialog (#6088).
 //! [`RoleWiring::role`] is a `String` for that reason: the deck's
 //! `envelope::roles::role_table` already renders a row for a role it has never
 //! heard of, so a seat needs no new plumbing here — only a row.
@@ -36,13 +36,13 @@ use crate::engine_config::{ModelSpec, model_spec_for, tuning_for};
 use crate::settings::AgentEngineConfig;
 use stella_protocol::ReasoningEffort;
 
-/// The one role core has. Seats a plugin declares join this list in #3909.
+/// The one role core has. Seats a plugin declares join this list in #6088.
 pub const DEFAULT_ROLE: &str = "default";
 
 /// One role's resolved wiring.
 #[derive(Debug, Clone, PartialEq)]
 pub struct RoleWiring {
-    /// The role's name — [`DEFAULT_ROLE`], or (from #3909) a plugin-declared
+    /// The role's name — [`DEFAULT_ROLE`], or (from #6088) a plugin-declared
     /// seat. A `String` because core does not enumerate the possibilities.
     pub role: String,
     /// Provider id and the slug as it goes on the wire.
--- a/crates/stella-cli/src/engine_config.rs
+++ b/crates/stella-cli/src/engine_config.rs
@@ -617,10 +617,10 @@ pub fn state_from_settings(
     let agents = EngineRole::ALL
         .iter()
         .map(|role| {
-            // One role, so one arm. The `match` survives the collapse
-            // deliberately: slice 5 (#3909) re-populates this from the live
-            // session's plugin seats, and a `match` is where each new row's
-            // source lands.
+            // One role, so one arm. The `match` survives the collapse so a
+            // second editable agent has somewhere to declare where its
+            // overrides come from. Plugin seats did not become one: they are
+            // `seats` below, an open map the deck reads beside this list.
             let agent = match role {
                 EngineRole::Default => engine.agent(),
             };
--- a/crates/stella-cli/src/profile.rs
+++ b/crates/stella-cli/src/profile.rs
@@ -273,7 +273,7 @@ pub struct Plan {
     /// The session role's pick.
     ///
     /// One, since #3908. A `Vec` rather than a bare `RolePick` because the
-    /// entries come back as plugin-declared seats in #3909, and every consumer
+    /// entries come back as plugin-declared seats in #6088, and every consumer
     /// here already iterates.
     pub picks: Vec<RolePick>,
     pub verbosity: Verbosity,
--- a/crates/stella-tui/src/envelope/engine_config.rs
+++ b/crates/stella-tui/src/envelope/engine_config.rs
@@ -17,13 +17,13 @@
 /// stopped steering anything. An overlay tab that writes a dead key is worse
 /// than a missing tab — it reports success.
 ///
-/// It stays an enum rather than collapsing into a bare struct because the
-/// pane's replacement is a **list**, not a single row: slice 5 (#3909) turns
-/// this into rows from the live session — the installed plugins' declared
-/// seats, each with its assigned model or `default` — the way
-/// `views/tools.rs` already sources MCP and custom tools. Keeping the
-/// role-indexed shape is what lets that land without re-plumbing
-/// `EngineConfigState`.
+/// It stays an enum rather than collapsing into a bare struct because a
+/// second editable agent would be one more name in this list rather than a new
+/// shape everywhere. The plugin seats are not in it: they are
+/// [`EngineConfigState::seats`], an open list the SEATS pane draws beside the
+/// resolved roles ([`crate::views::seats::rows`]), the way `views/tools.rs`
+/// sources MCP and custom tools. This enum indexes
+/// `agent_engine_config.agents` and nothing more.
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
 pub enum EngineRole {
     Default,
--- a/crates/stella-tui/src/views/seats.rs
+++ b/crates/stella-tui/src/views/seats.rs
@@ -1,10 +1,11 @@
 // SPDX-License-Identifier: AGPL-3.0-only
 // Copyright (c) 2026 Oxagen, Inc. Commercial licensing: licensing@oxagen.sh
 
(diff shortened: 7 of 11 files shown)


macanderson:
**DoD checked against this branch (`1c6ac2f`) — three of four items verified done, one genuinely outstanding.** Autonomous PR sweep.

I ticked the first three boxes on #3909 with the evidence for each; `dod / dod` failed naming all four, and three of those were false.

The one that is real: **no golden snapshot is committed.** The test module lands (`crates/stella-tui/tests/deck_render_snapshots/settings_seats.rs`, +63, registered in `deck_render_snapshots.rs`) but `git ls-tree crates/stella-tui/tests/snapshots/deck/` on this head matches nothing for the seats pane. That is exactly what your description predicts, so this is confirmation rather than a finding — but it means `fmt + clippy + test` and the last DoD box are both waiting on the same push. That job is still running as I write; when it fails it will print the frame `assert_golden` rendered, which is the input for the next push.

Two things I checked rather than took on trust:

- **Item 3 holds.** The only role-shaped literal in `views/seats.rs` is `const UNASSIGNED: &str = "default"`, and both uses are the *model* column's fallback (`row.model.as_deref().unwrap_or(UNASSIGNED)`) — never a row key. `the_leading_row_is_named_by_the_driver_not_by_this_pane` proves the positive case with a driver that says `lead`.
- **Your "Where this departs from the issue" section is now stale**, in your favour. #3909's DoD was edited at 19:21 — seven minutes before this PR opened — to split the "no role name is compiled into `stella-tui`" line and carve out #6087 and #3910 by name. The tree does still contain the five words in `views/models_card.rs` (confirmed: lines 106–114), so your analysis was right; the issue already agrees with it, and no box asks you to tick that any more. Worth trimming that section so a reviewer is not sent to re-litigate a settled point.

Nothing else is blocking: every other required check on this head is green or still running, there are no unresolved review threads, and the PR carries no `hold` label.

---

macanderson:
**Answering the one ❌ row: "Add and commit a golden frame … the diff does not include the corresponding golden snapshot file."**

Correct at the commit Sourcery read, and fixed since. The review ran against
`1c6ac2f8d`, the first push, which carried the render test with no snapshot —
this machine does not run cargo, so the frame had to come from a run that
rendered it. `83e7d45c0` commits it:
`crates/stella-tui/tests/snapshots/deck/pane_settings_seats.txt`.

The frame was read rather than blessed. Its two rows are:

```
 seats · 2 · read-only
 default         zai/glm-5.2-air   from default_model
 acme/reviewer   default           from acme
```

which is the order the fold promises, the resolved model and settings key on
the leading row, and `default` in the *model* column of the unassigned plugin
seat while the row keeps its own key. None of the five role words #3908 retired
appear anywhere in the 160×40 frame, which is why this fixture is safe to commit
where `base_ui`'s would not have been.

`fmt + clippy + test` is green on that head, so the golden is being compared
rather than merely committed.

Issue #6089, opened 2026-09-05
Confirm on a live pull request that an issue edit clears the DoD check


macanderson:
## The problem

`.github/workflows/dod-recheck.yml` runs the failed `dod / dod` check again
when a linked issue is edited, so ticking the last box on the issue clears the
pull request. It was proven before it merged by `scripts/test-dod-recheck.sh`,
which drives fixture seams and sends nothing.

It was not proven live, and it could not be. A workflow on an `issues` event
runs from the default branch, so the file has to land on `main` before GitHub
will start it at all. Its own pull request could never fire it.

## What to do

Watch the first real case after the merge, then say what happened.

1. Find an open pull request that is red on `dod / dod` alone, and whose
   linked issue has a checklist item that is genuinely satisfied.
2. Tick that box on the **issue**. Touch nothing on the pull request.
3. `gh run list -R macanderson/stella --workflow dod-recheck.yml` should show
   a run within a minute. Read its log: it names each pull request it swept
   and the run id it asked GitHub to re-run.
4. `gh pr checks <n>` should then show `dod / dod` green, and the pull
   request's timeline should carry no new event of its own.

## If it does not fire

The likely causes, cheapest first:

- The box was ticked by something using `GITHUB_TOKEN`. GitHub starts no
  workflow from an event that token raised, and the script's header says so.
- The edit changed the title and not the body. The job's `if:` drops those.
- The job ran and could not re-run: the log then says so and names
  `actions: write`.

## Definition of done

- [x] A real issue edit starts a `dod-recheck` run that resolves the issue to
      its open pull requests. Confirmed: runs `33988285457` and `33988014708`,
      both `event: issues`.
- [x] The fan-out bound holds live: every pull request whose `dod / dod` is
      green is left alone, and the log says so.
- [ ] A run started this way is shown re-running a **failing** `dod / dod` —
      the log line `re-ran the dod check on PR #N` — and that pull request
      goes green with no event of its own.
- [ ] If it does not fire, the cause is named here and either fixed or filed.

## Context

Filed as the deferred half of #6079, whose PR added the workflow, the script
and the hermetic suite.

## What is already confirmed (2026-09-05)

Two of the three links were checked before this issue was opened. Only the
trigger is left.

1. **The re-run clears the check.** On PR #6091, run `33986833605` concluded
   `failure` at 19:20:50 against an unticked checklist. The boxes on #6079
   were ticked, nothing on the pull request was touched, and
[...]

PR #6097, opened 2026-09-05
ci(dod-check): re-pin the gate to the build that reads Refs-only and negation

macanderson:
## What and why

macanderson/oxagen#2663 landed three fixes to the shared DoD check, and this
repository pins the implementation to a commit rather than `@main`, so none of
them reached it:

- a clause-scoped negation, so "does not close #N" is no longer read as a claim
  to close N (oxagen#2636);
- a third passing state for a PR that only advances an issue with `Refs #N`
  (oxagen#2640);
- a failure message that says outright that ticking the issue's boxes does not
  re-run the check.

**#6093 is red right now for exactly the missing state.** Its body carries
`Refs #6068` and `Refs #5200` and no `Closes`, which is the correct choice —
the PR closes the ledger half of #6068 and says so — and the pinned build
reports it as `This PR links no issue`. The two escape hatches the old build
offers are both wrong for it: `Closes` would close an issue whose DoD is not
met, and `no-issue` would assert a 355-line feature change is trivial.

`ready_for_review` joins the trigger list in the same change. The caller stub
owns that list — oxagen#2663 added the event to oxagen's own caller and could
not add it here on our behalf — so a PR opened as a draft and later marked
ready kept whatever verdict its draft head got (oxagen#2638).

The file's own header already names this as the maintenance this pin costs:
"Re-pin when the implementation changes, which is a reviewable one-line diff
naming the version it moved to."

## Verification

No crate is touched, so nothing here compiles. The guards that judge this diff:

- `check-action-pins` — OK, all 102 `uses:` references pinned to a commit SHA
  (this is the guard the SHA pin exists to satisfy).
- `check-prose` — OK, none added.
- `check-line-citations` — OK, none added.
- The workflow parses as YAML.
- The pinned SHA `f4c13eb` is the tip of `macanderson/oxagen`'s `main` and
  carries the Refs-only state: `git show f4c13eb:tools/scripts/scr-dod-check.mjs`
  shows `REFS_PATTERN`, `referencedIssues()` and the `refsOnly` verdict field.

**This PR is its own witness, and the witness has now run.** Its body is
`Refs`-only with no `Closes`, so under the *old* pin `linkedIssues()` returns
nothing, there is no `refs` concept to fall back on, and `dod / dod` fails with
`This PR links no issue` — which is the failure #6093 is sitting on. Under the
new pin the same body takes the `refsOnly` branch and passes.

`dod / dod` on `981a89e` — [run 33987600937](https://github.com/macanderson/stella/actions/runs/33987600937),
completed 19:42:15Z — **success**. A `pull_request` event resolves the caller
workflow from the merge ref, so that run used this branch's pin. The old pin
could not have produced that result on this body, which is what makes the green
a witness rather than a coincidence.

## Tests deleted

None.

## Residue

The other three consumer repositories (`arenabench`, `cgp-website`,
`context-graph-protocol`) use `@main` and already have the fix — checked, not
assumed. This repository is the only one that pins a SHA, which is the whole
reason it lagged.

## Correction

An earlier revision of this description claimed this PR was "the live
confirmation #6089 asks for". That was wrong, and #6089 is untouched by this
change: it asks whether `dod-recheck.yml` starts on an `issues: [edited]` event
once that workflow is on `main` — a question about #6091's diff, not about which
build the gate is pinned to. The green run above confirms the re-pin, and
nothing more.

Refs macanderson/oxagen#2640
Refs macanderson/oxagen#2638

https://claude.ai/code/session_015iQ1TL85RiG2HARN5YoLJZ

### DIFF
--- a/.github/workflows/dod-check.yml
+++ b/.github/workflows/dod-check.yml
@@ -16,7 +16,12 @@ name: dod-check
 
 on:
   pull_request:
-    types: [opened, edited, reopened, synchronize, labeled, unlabeled]
+    # `ready_for_review` because a PR opened as a draft and later marked ready
+    # never re-ran the check, so it kept whatever verdict its draft head got
+    # (oxagen#2638). The caller owns this list; the implementation cannot add
+    # an event on our behalf.
+    types:
+      [opened, edited, reopened, ready_for_review, synchronize, labeled, unlabeled]
 
 permissions:
   contents: read
@@ -25,4 +30,4 @@ permissions:
 
 jobs:
   dod:
-    uses: macanderson/oxagen/.github/workflows/dod-check.yml@acf3509c0012240362e8af8eb91bb1aad6f9a428 # oxagen main, #2643
+    uses: macanderson/oxagen/.github/workflows/dod-check.yml@f4c13eb924f6b2ac63da2e12d0504d7c2278af7f # oxagen main, #2663


macanderson:
**The second half of this change now has a witness too, and it arrived by accident.**

Marking this PR ready for review started a `dod / dod` run — [33988185426](https://github.com/macanderson/stella/actions/runs/33988185426), completed 19:53:00Z, **success**.

That run is the `ready_for_review` trigger firing on its own PR. `main`'s copy of `dod-check.yml` lists `[opened, edited, reopened, synchronize, labeled, unlabeled]` and does not carry `ready_for_review`, so under it no run exists at all for that event — which is oxagen#2638's bug, and the reason a PR opened as a draft kept whatever verdict its draft head got. A `pull_request` event resolves the caller workflow from the merge ref, so the run that exists came from this branch's list.

So both halves are now evidenced on this head rather than argued:

| Change | Witness |
|---|---|
| Re-pin to `f4c13eb` | `dod / dod` green on a `Refs`-only body — [33987600937](https://github.com/macanderson/stella/actions/runs/33987600937), 19:42:15Z. The old pin returns "links no issue" for this body. |
| `ready_for_review` added to the trigger list | a `dod / dod` run exists for that event at all — [33988185426](https://github.com/macanderson/stella/actions/runs/33988185426), 19:53:00Z. Under `main`'s list there is no run to pass or fail. |

Both are observations. The claim that the *old* pin would fail this body is an inference from reading `f4c13eb`'s diff against the previously pinned `acf3509c`, not something I re-ran against the old build.

**Status:** every required context is green except `fmt + clippy + test`, running since 19:56:18. Nothing is red, there is no merge conflict, and Sourcery's only review on this PR is its rate-limit notice — no findings to settle.

Worth noting the re-pin is no longer what unblocks #6093: that PR merged at 19:53 after its scope was narrowed and `Refs #6068` became an accurate `Closes #6068`. This PR still earns its place — the negation fix and the Refs-only state apply to every future PR in this repository, and the `ready_for_review` gap is real regardless.

---

PR #6101, opened 2026-09-05
fix(stella-store,stella-cli): let the hub prune reach a removed tree and an orphan id

### LINKED ISSUE #6100: The usage hub kept 2,759 dead bench projects and 33k orphan rows the prune could not reach
## What happened

`~/.stella/usage.db` on the maintainer's machine held 2,795 registered projects and 2,965 tool-fold ledger rows, one per project, which looked like a keying bug. It was not. 2,313 of the projects were first seen in July, and their roots were bench trial workspaces under the OS temp directory (`/private/var/folders/.../T/stella-proving-ground-work/series-*/work/<trial>`) plus `.tmpXXXX` dirs. The proving-ground harness runs one throwaway workspace per trial, each trial syncs into the hub, and the hub registers each as a project forever. Only 36 of the 2,795 roots still existed.

`stella usage prune --gc-deleted` is the sanctioned way out, and it could not reach them:

- Its "checkout is deleted" rule in `crates/stella-cli/src/usage_cmd.rs` (`checkout_is_deleted`) required the root's parent to still exist, reading an absent parent as an unmounted volume. A bench series is removed whole, so every trial under it lost its parent too. Dry run: 448 of 2,759 dead projects reachable, 2,311 never.
- 33,177 telemetry rows under 206 project ids had no `projects` row at all (replicated without a `sync_execution`, or left behind by a rekey). The prune lists candidates from `projects`, so an orphan id could never be named.

Hub before: 84 MB. After both fixes and a prune: 36 projects, 36 ids in telemetry, 9.2 MB.

## Definition of done

- [x] A root missing beneath the home directory or the OS temp directory counts as deleted even when its parent is gone too; a missing parent elsewhere still reads as an unmounted volume. A test pins both rulings against directories the test owns. PR #6101: `checkout_is_deleted_under` + `mounted_anchors` in `usage_cmd.rs`; `a_tree_removed_whole_under_a_standing_anchor_is_deleted` and the reshaped `checkout_is_deleted_distinguishes_gone_from_unreachable` pass (`cargo test -p stella-cli --bin stella usage_cmd::tests`: 5 passed).
- [x] `stella usage prune --gc-deleted` also clears project ids that appear in `telemetry` or `execution_rollup` with no `projects` row, under the same org-row guard as every other GC. A test seeds an orphan through `replicate_telemetry` alone and shows the prune reaches it and keeps an org-scoped one. PR #6101: `UsageStore::orphan_project_ids` in `usage.rs`, joined to the GC set in `usage_cmd.rs`'s `prune`; `telemetry_with_no_registered_project_is_an_orphan_the_prune_reaches` passes.
- [x] The prune's dry run on the maintainer's hub names every dead bench project and every orphan id (2,759 and 206 on 2026-09-05). Run with the fixed build on 2026-09-05: `would remove 248733 row(s) from 2759 deleted project(s)`, then after that prune `would remove 33177 row(s) from 206 deleted project(s)`; both were then applied with `--vacuum`, leaving 36 projects and 36 telemetry ids.

## Constraints

- `usage_cmd.rs`'s existing test `checkout_is_deleted_distinguishes_gone_from_unreachable` pins the old ruling for a path under the temp dir, which the fix flips on purpose; it moves to the explicit-anchor form rather than being deleted.
- The hub's GC per id already checks for org rows before deleting, so an orphan joins `gc_project_ids` as it is.
- Why bench trials register at all, and whether the harness should sync under `STELLA_NO_SETTINGS` into the user's hub, is a separate question for the proving-ground repo.

macanderson:
## What & why

`~/.stella/usage.db` on the maintainer's machine held 2,795 registered projects. They were not worktrees: 2,313 were first seen in July, and their roots were bench trial workspaces under the OS temp directory, one per trial, each synced into the hub and registered forever. Only 36 roots still existed. `stella usage prune --gc-deleted` is the sanctioned way out, and it reached 448 of the 2,759 dead projects (#6100):

- **The "checkout is deleted" rule needed the parent to survive.** A bench series is removed whole, so every trial under it lost its parent too, and the rule read that as an unmounted volume. The rule now consults two anchors that exist whenever the process runs: the home directory and the OS temp directory. A root missing beneath either was deleted, never unmounted. A missing parent anywhere else still reads as a missing volume. Both spellings of each anchor are checked, because a hub records the canonical `/private/var/...` path while `temp_dir` reports the `/var/...` symlink.
- **Orphan ids were unreachable by construction.** 33,177 telemetry rows sat under 206 project ids with no `projects` row (replicated without a `sync_execution`, or left by a rekey). The prune listed candidates from `projects`, so it could never name them. `UsageStore::orphan_project_ids` lists ids seen in `telemetry` or `execution_rollup` with no project row, and the CLI adds them to the GC set. The per-id GC already refuses an id with org-scoped rows, so an orphan joins under the same guard.

Run on the hub in question: `would remove 2855 row(s) from 448 deleted project(s)` before; `248733 row(s) from 2759` and then `33177 row(s) from 206` after. The file went from 84 MB to 9.2 MB and the hub is consistent: 36 projects, 36 ids in telemetry.

Closes #6100

## The witness

- `usage_cmd::tests::a_tree_removed_whole_under_a_standing_anchor_is_deleted`: a root three levels under a standing anchor with no parent is deleted; an absent anchor says nothing; a root outside every anchor keeps the unreachable ruling. On `main` the anchored function does not exist, and the old rule returns false for that root, which is what the 2,311 unreachable projects measured.
- `usage::tests::telemetry_with_no_registered_project_is_an_orphan_the_prune_reaches`: rows seeded through `replicate_telemetry` alone are listed as orphans, the prune clears the NULL-org one and keeps the org-scoped one. On `main` the method does not exist.
- The existing `checkout_is_deleted_distinguishes_gone_from_unreachable` pinned the old ruling for a path under the temp dir, which this change flips on purpose. It moves to the explicit-anchor form with an empty anchor list, so the parent rule it was really about is still pinned.

## The gate

- [x] `cargo fmt --check`
- [x] `cargo clippy -p stella-store -p stella-cli --all-targets -- -D warnings` (targeted; CI runs the workspace)
- [x] `cargo test -p stella-store --lib usage::tests`, `cargo test -p stella-cli --bin stella usage_cmd::tests` (targeted; CI runs the workspace)
- [x] Docs updated where behavior changed (doc comments on the rule and the new method)
- [x] CLA signed

## Nothing left behind

- [x] Filed: #6100 carries the open question of why bench trials register in the user's hub at all; the harness lives in the proving-ground repo.

## Ground-rule check

Tests deleted: none. One test changed shape as described above.

## Summary by Sourcery

Make usage garbage collection reach deleted checkout trees and orphaned project data without removing unreachable or protected records.

Bug Fixes:
- Allow usage pruning to identify projects whose entire checkout tree was removed beneath the home or temporary-directory anchors.
- Include orphaned telemetry and execution-rollup project IDs in garbage-collection candidates while preserving protection for org-scoped data.

Enhancements:
- Improve deleted-checkout detection for paths recorded through alternate canonical and symlinked anchor spellings.

Tests:
- Add coverage for whole-tree deletion beneath a standing anchor and orphan project garbage collection.

### DIFF
--- a/crates/stella-cli/src/usage_cmd.rs
+++ b/crates/stella-cli/src/usage_cmd.rs
@@ -13,7 +13,7 @@
 //! keep its token) and mints the durable, committable per-workspace
 //! `workspace_id`. Until registration, hub rows carry NULL org/workspace ids.
 
-use std::path::Path;
+use std::path::{Path, PathBuf};
 
 use clap::Subcommand;
 use colored::Colorize as _;
@@ -550,12 +550,48 @@ pub(crate) fn parse_age_window(s: &str) -> Result<String, String> {
     Ok(modifier)
 }
 
-/// Whether a project's checkout looks *deleted* (GC-eligible) rather than merely
-/// unreachable. The root being gone while its parent still exists is a deletion;
-/// a missing parent usually means the whole volume/mount is absent (an unplugged
-/// drive, a down network share), which must not trigger GC of local history.
+/// Whether a project's checkout is deleted, rather than on a volume that is
+/// not mounted right now.
+///
+/// A root that is gone while its parent still stands was deleted. A root
+/// whose parent is gone too is ambiguous on its own: an unplugged drive or a
+/// down share looks the same as a whole tree removed at once. The tie is
+/// broken by [`mounted_anchors`]: the home directory and the OS temp
+/// directory exist for this process to be running at all, so anything
+/// missing beneath either was deleted, never unmounted. A bench harness
+/// leaves thousands of trial workspaces under the temp directory and removes
+/// each series whole, which is the shape the parent rule alone could not
+/// reach: on 2026-09-05 one hub held 2,311 such projects it would never GC.
 fn checkout_is_deleted(root: &Path) -> bool {
-    !root.exists() && root.parent().is_some_and(Path::exists)
+    checkout_is_deleted_under(root, &mounted_anchors())
+}
+
+/// [`checkout_is_deleted`] against an explicit anchor list, so the rule can
+/// be tested with directories a test owns.
+fn checkout_is_deleted_under(root: &Path, anchors: &[PathBuf]) -> bool {
+    if root.exists() {
+        return false;
+    }
+    if root.parent().is_some_and(Path::exists) {
+        return true;
+    }
+    anchors
+        .iter()
+        .any(|anchor| anchor.exists() && root.starts_with(anchor))
+}
+
+/// Directories that are present whenever this process runs. Each is listed
+/// as spelled and as canonicalized, because a hub records the canonical
+/// path (`/private/var/...` on macOS) while `temp_dir` reports the symlink.
+fn mounted_anchors() -> Vec<PathBuf> {
+    let mut anchors = vec![std::env::temp_dir()];
+    anchors.extend(stella_home::home_dir());
+    let canonical: Vec<PathBuf> = anchors
+        .iter()
+        .filter_map(|anchor| anchor.canonicalize().ok())
+        .collect();
+    anchors.extend(canonical);
+    anchors
 }
 
 fn prune(
@@ -580,14 +616,21 @@ fn prune(
     let hub = open_hub()?;
 
     // The store enforces the "unregistered" invariant; the CLI owns the
-    // filesystem check for "the checkout is gone".
+    // filesystem check for "the checkout is gone". An id with no project row
+    // has no checkout to check and joins the set as it is.
     let gc_project_ids = if gc_deleted {
-        hub.registered_projects()
+        let mut ids: Vec<String> = hub
+            .registered_projects()
             .map_err(|e| format!("cannot list hub projects: {e}"))?
             .into_iter()
             .filter(|(_, _, root)| checkout_is_deleted(Path::new(root)))
             .map(|(id, _, _)| id)
-            .collect()
+            .collect();
+        ids.extend(
+            hub.orphan_project_ids()
+                .map_err(|e| format!("cannot list orphan hub rows: {e}"))?,
+        );
+        ids
     } else {
         Vec::new()
     };
@@ -879,7 +922,9 @@ pub fn run_cloud(cmd: CloudCmd) -> Result<(), String> {
 
 #[cfg(test)]
 mod tests {
-    use super::{checkout_is_deleted, parse_age_window, register_transition};
+    use std::path::PathBuf;
+
+    use super::{checkout_is_deleted_under, parse_age_window, register_transition};
 
     /// The org-change ruling (#465): same org is idempotent, a different org
     /// is refused until an explicit deregister, unregistered registers freely.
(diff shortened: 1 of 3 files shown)


macanderson:
Autonomous PR sweep — `file size ratchet` was red here, and `e64ea46` fixes it. Pushed rather than reported, because the remedy is mechanical and I could verify it.

**The failure** ([run](https://github.com/macanderson/stella/actions/runs/33987923852/job/101364801812), 19:50:58Z):

```
crates/stella-store/src/usage.rs grew to 1832 lines, over its baseline ceiling of 1767 (+65)
```

`usage.rs` is grandfathered and closed to growth, and it was already at **1766 on `main`** — one line under its ceiling. So the 66 lines this branch added to it could not have passed, and nothing smaller would have either.

**Why not `make file-size-update`.** AGENTS.md reserves the ceiling bump for a genuinely irreducible line, and CLAUDE.md names raising the baseline to turn a gate green as an expedient — a bug against the PR that introduces it. A new method plus its witness is the split the guard exists to ask for, so it got the split.

**What changed** — behaviour is identical; nothing was rewritten, only relocated:

- `orphan_project_ids` and `telemetry_with_no_registered_project_is_an_orphan_the_prune_reaches` move to `crates/stella-store/src/usage/orphans.rs`, beside `usage/rekey.rs` — the existing sibling that holds a table-spanning hub operation and its own tests, and whose `use super::{Result, UsageStore}` / `use super::super::tests::…` shape this copies.
- `usage.rs` keeps one line, `mod orphans;`, placed where rustfmt's ordering already wants it (`drain_state` < `orphans` < `rekey`) so it needs no reflow. **1767 exactly — its ceiling, with no baseline edit.**
- The method's doc comment becomes the module's `//!` header, which had room to say the things it was compressing: the two paths that produce an orphan, why the gone-checkout scan cannot reach one, and that `prune`'s org-row guard still applies.

One judgement worth flagging since it looks like duplication: the witness borrows `source_row` from the parent's fixtures (widening it to `pub(super)` costs no line) but defines its own `scope`. Widening the parent's `scope` makes its signature exceed the line width, and rustfmt then reflows it across four lines — which breaches the very ceiling this commit is clearing. The reason is written at the site so it does not read as carelessness.

**Verified on this branch, not asserted:**

| Command | Result |
|---|---|
| `cargo test -p stella-store --lib usage::orphans` | 1 passed — the witness, in its new home |
| `cargo test -p stella-store --lib usage::` | 42 passed, 0 failed |
| `cargo clippy -p stella-store --all-targets -- -D warnings` | clean |
| `cargo fmt --check -p stella-store` | clean |
| `RUSTDOCFLAGS="-D warnings" cargo doc -p stella-store --no-deps --document-private-items` | clean |
| `./scripts/check-file-size.sh` | exit 0 |

Targeted rather than workspace-wide, per SCR-001 — CI runs the full gate, and that run is the authoritative evidence, not these.

`Refs #6100` on the commit rather than `Closes` — this repairs the branch, it does not finish the issue; your `Closes #6100` still owns that.

Nothing else on the PR was blocking: `dod / dod` passed at 19:46:02Z, and no other completed check is red.

---

PR #6114, opened 2026-09-05
feat(stella-cli,stella-learn): record a trial for a memory record and a mined rule

### LINKED ISSUE #6096: Nothing records a trial for a memory record or a mined rule, so neither can ever be measured
Split out of #6068, which now covers the ledger row alone. Slice **A** of epic
#5200 asks for one trial ledger with three kinds. The row can hold all three
since #6093; only skills write one.

## The problem

`crates/stella-cli/src/memory/appraisals.rs` holds
`.stella/private/artifact_trials.jsonl`, keyed by
`stella_learn::ledger::ArtifactKind` and an id. `SessionMemory::note_turn_skills`
arms the offered→selected join for skills and
`SessionMemory::record_skill_trials` writes it at turn end, so
`ArtifactKind::Skill` has a producer.

`ArtifactKind::Memory` and `ArtifactKind::Rule` have none.
`crates/stella-cli/src/memory/recall.rs` recalls memory frames and renders the
context-record channel, and it reports no join for either: nothing says which
records the turn could have used and which it actually put in front of the
model. Without that pair there is no control arm, so `appraisals::sweep` on
either kind reads an empty window forever.

Two things sit behind this and cannot move until it lands:

- The holdout scheduler (`crates/stella-learn/src/holdout.rs`) is kind-opaque
  already — its item is a plain id string — so it would hold a memory record or
  a mined rule back with no change at all. It has no reason to, because nothing
  would record the turn.
- Memory retirement (#6069) is meant to read the shared sweep instead of
  citations, and the sweep has nothing to read.

## The fix

For each of the two kinds, a turn-start seam that reports the offered set and
the selected set, and a turn-end write into the ledger.

- **Memory records.** The offered population is what the recall query returned;
  the selected population is what reached the rendered block.
  `memory::steering::ProducedSteering` already holds the second by handle, and
  `frame_handle` is the stable id. Withholding one means filtering the frame
  before the render, and both production paths render — `recall_block_reported`
  and the mid-turn `SessionRequery` — so both have to agree or the ledger will
  say the opposite of what happened. That is the defect #6073 hit and fixed for
  skills; read `SessionMemory::apply_holdout`'s doc comment before designing
  this one.
- **Mined rules.** The record channel already has an exclusion door:
  `stella_records::records::Registry::render_volatile_for_turn_excluding`.
  `RenderedChannel` reports `rendered` and `dropped` handles, which together
  with the withheld one are the offered population.

Then call `appraisals::record_turn` with the right `ArtifactKind` from the
episode seam, beside the skill call that is there today.

## Constraints

- `crates/stella-cli/src/memory/recall.rs` sits within about 35 lines of the
  1500-line ratchet (`scripts/check-file-size.sh`). Plan a sibling submodule;
  do not grow it.
- `crates/stella-learn` does no I/O and takes `stella-protocol` as its only
  workspace dependency. Keep it that way — the seam is `stella-cli`'s.
- Both arms of a trial must come from one population. A memory offered on a
  turn its query never surfaced is not a control arm; it is a different turn.
  `appraisals::record_turn`'s doc comment states the rule.

## Verify

A turn that recalls a memory record writes a trial row keyed to that record,
and a turn where the holdout withheld it writes the control-arm row. Both from
the production seams, not a hand-built trial vector.
`a_holdout_turn_writes_a_control_arm_trial_for_the_skill_it_withheld`
(`crates/stella-cli/src/memory/learning/skill_lifecycle.rs`) is the shape to
copy.

## Definition of done

- [x] A memory record recalled into a turn writes an `ArtifactKind::Memory`
      trial at turn end, from the production seam.
      *`SessionMemory::withhold_held_memory` notes the join from both render
      paths and `record_context_trials` writes it from `record_episode`
      (`crates/stella-cli/src/memory/trials.rs`). Witnessed by
      `a_recalled_memory_writes_a_treatment_arm_trial`, green in the
      `fmt + clippy + test` run for PR #6114.*
- [x] A mined rule rendered into a turn writes an `ArtifactKind::Rule` trial the
      same way.
      *`SessionMemory::turn_records_held` reads the population off an
      unexcluded render of the volatile channel. Witnessed by
      `a_rendered_record_writes_a_treatment_arm_trial`.*
- [x] The holdout withholds one of each kind on its schedule, and the withheld
      turn writes the control-arm row.
      *One item per turn still, so the three kinds take the schedule in turn:
      `stella_learn::holdout::arm` is the kind-opaque rotation and
      `memory::trials::HOLDOUT_ARMS` fixes the order. Witnessed by
      `the_rule_arm_withholds_a_record_and_writes_its_control_arm`,
      `the_memory_arm_withholds_a_memory_and_writes_its_control_arm`, and the
      pre-existing skill case
      `a_holdout_turn_writes_a_control_arm_trial_for_the_skill_it_withheld`,
      which still lands on arm 0.*
- [x] A test drives one of the two kinds end to end through the production
      seams, withheld arm included.
      *`the_rule_arm_withholds_a_record_and_writes_its_control_arm` runs six
      turns through `arm_controls_at`, `recall_block_reported` and
      `record_episode`, then reads the raw ledger: five rendered turns and one
      withheld.*

macanderson:
## What and why

The artifact trial ledger (`.stella/private/artifact_trials.jsonl`) has been keyed by `ArtifactKind` since #6093, and only skills ever wrote a row. A memory record and a mined rule were recalled and rendered with no join reported, so `appraisals::sweep` read an empty window for both kinds and neither could ever be measured. The holdout scheduler was already kind-opaque and would have withheld either one — it had no reason to, because nothing would have recorded the turn.

This adds the seam:

- **`crates/stella-cli/src/memory/trials.rs`** (new, plus `trials/tests.rs`) holds the per-turn offered→selected join for `ArtifactKind::Memory` and `ArtifactKind::Rule`, the per-kind holdout door, and the turn-end write.
- **Both render paths report.** `recall_block_reported` and the mid-turn `signal_recall_block` each call `withhold_held_memory` for frames and `turn_records_held` for the record channel. The memory population is `frame_handle` over everything the recall query answered; the rule population is `rendered + dropped` from a render that leaves nothing out. The withheld record leaves through `Registry::render_volatile_for_turn_excluding`, the door the issue names.
- **One pick per turn.** The pick is settled by the first pass that has something to pick from, and every later pass in the turn reads it back. The two passes score different shortlists, so a per-pass pick could drop one record from the opening block and a different one from the re-query, and the ledger would then describe neither turn — the shape #6073 fixed for skills. The turn's offered set is the union of what the passes offered; its selected set is the union of what they showed.
- **Turn end.** `record_episode` calls `record_context_trials` beside the existing `record_skill_trials`. A turn that offered nothing of a kind writes nothing for that kind.
- **One item per turn, still.** `stella_learn::holdout`'s own rule is "one item at most — hold two back and the turn cannot say which one the outcome belongs to". Holding one of each kind back on the same turn would make every memory control turn a skill control turn too, which confounds both arms. So the three kinds take the schedule in turn: a new kind-opaque `holdout::arm(ordinal, arms)` says whose turn it is, and `memory::trials::HOLDOUT_ARMS` fixes the order with `Skill` first so an existing workspace's next skill holdout does not move. `apply_holdout` returns `None` on a turn the schedule points elsewhere.

`recall.rs` sat about 26 lines under the 1500-line ratchet, which is why none of this landed there; deleting `turn_record_rendered` in favour of the new door leaves it smaller than it was (1474 → 1450).

## Witness mechanism

`crates/stella-cli/src/memory/trials/tests.rs`, five cases, each driving only production doors — `arm_controls_at`, `recall_block_reported`, `record_episode` — and then reading the raw JSONL the sweep reads, filtered on `kind` **and** `id`:

- `a_recalled_memory_writes_a_treatment_arm_trial` — one `memory` row, selected, and its id is the handle `RecalledBlock::produced` says the block showed.
- `a_rendered_record_writes_a_treatment_arm_trial` — one `rule` row, selected.
- `the_rule_arm_withholds_a_record_and_writes_its_control_arm` — six turns at rate 2 give three holdouts; the third is the rule arm. The record renders on five turns and not on the sixth, and the rows read `[true × 5, false]`. The plane rate is 0, steering is on, and the record is unscoped, so nothing else in the fixture could produce a `selected: false`.
- `the_memory_arm_withholds_a_memory_and_writes_its_control_arm` — four turns; the second holdout is the memory arm. Only the last turn writes an episode, because an episode is itself a recallable memory and writing one each turn would grow the population under the pick.
- `a_turn_with_nothing_to_offer_writes_no_row` — the ledger file is not even created.

**Why they fail on the old code by construction:** before this change the only producer of a ledger row is `SessionMemory::record_skill_trials`, which writes `ArtifactKind::Skill` and nothing else. Every assertion above reads rows filtered to `kind == "memory"` or `kind == "rule"`, so on `main` each of the four positive cases reads an empty vector against a non-empty expectation. The holdout half is likewise absent: `apply_holdout` is the only holdout consumer on `main` and it touches skills alone.

The pre-existing skill witness `a_holdout_turn_writes_a_control_arm_trial_for_the_skill_it_withheld` still passes under the rotation: its holdout lands on ordinal 0, which is the skill arm.

`stella-learn` gains `the_arms_take_the_schedule_in_turn` and `no_arms_is_no_holdout` for the new pure function.

## Exemplar

`crates/stella-cli/src/memory/learning/skill_lifecycle.rs`'s holdout witness is the shape the two control-arm cases copy: drive the production doors, read the ledger file, and close off every other route to the control arm in the fixture itself.

## Docs touched

`website/content/docs/context-engine.mdx` and `website/content/docs/configuration/settings.mdx` described the holdout as per-skill; both now say per-artifact and name the rotation. `settings/context.rs` and `memory/appraisals.rs` module docs said the same thing and are corrected. The deck's boot disclosure now names all three kinds.

## Tests deleted

None.

## Residue issues filed

None found that this change leaves behind. Two things it deliberately does not do, both already tracked elsewhere: memory retirement reading the shared sweep is #6069, and the sweep's per-kind retirement verbs are the epic's own next slice (#5200).

Closes #6096

## Summary by Sourcery

Measure recalled memories and rendered rules by recording their per-turn treatment and control trials in the shared artifact ledger.

New Features:
- Record appraisal trials for recalled memory records and rendered mined rules in the shared artifact trial ledger.
- Extend artifact holdouts to rotate across skills, memories, and rules while withholding at most one artifact per turn.

Bug Fixes:
- Ensure repeated recall and render passes contribute to a single per-turn offered and selected trial join.

Enhancements:
- Integrate memory and rule trial recording into the turn-end episode path and preserve the existing skill holdout schedule.
- Add production-door witnesses covering treatment and control arms, empty populations, and holdout rotation.

Documentation:
- Update holdout documentation and session disclosures to describe per-artifact rotation across memories, skills, and records.

Tests:
- Add end-to-end ledger tests for memory and rule trial production and their control arms.
- Add unit tests for rotating holdout arm selection.

### DIFF
--- a/crates/stella-cli/src/command_deck/steering.rs
+++ b/crates/stella-cli/src/command_deck/steering.rs
@@ -47,7 +47,9 @@ pub(super) fn announce_session_steering(cfg: &Config, in_tx: &UnboundedSender<In
 /// unlike the withheld-checkout refusal it names no remedy to scroll back to.
 fn announce_holdout(cfg: &Config, in_tx: &UnboundedSender<Inbound>) {
     let rate = crate::memory::session_artifact_holdout_rate(&cfg.workspace_root);
-    if let Some(line) = stella_learn::holdout::disclosure(rate, "skill") {
+    // Three kinds share the schedule and take it in turn
+    // (`memory::trials::HOLDOUT_ARMS`), so the line names all three.
+    if let Some(line) = stella_learn::holdout::disclosure(rate, "memory, skill or record") {
         let _ = in_tx.send(Inbound::Notice(line));
     }
 }
--- a/crates/stella-cli/src/memory.rs
+++ b/crates/stella-cli/src/memory.rs
@@ -87,6 +87,10 @@ mod steering;
 pub(crate) use steering::SessionRequery;
 pub(crate) use steering::requery_for_turn;
 mod suppression;
+// The offered->selected join for the two kinds recall renders — a memory
+// record and a mined rule — and the turn-end write into the shared trial
+// ledger.
+mod trials;
 pub(crate) mod tuning;
 // Phase 4 (#715): context-use extraction — what a finished turn's frame
 // carried, and what the turn then said about it.
@@ -370,11 +374,14 @@ pub struct SessionMemory {
     holdout_turn: u64,
     /// Which holdout this turn is, counting from zero, or `None` when this
     /// turn holds nothing back. Set by `arm_recall_control()` alongside
-    /// `ab_suppressed`, read by [`Self::note_turn_skills`] when it picks the
-    /// skill to withhold.
+    /// `ab_suppressed`, and read wherever the withheld item is picked —
+    /// [`Self::note_turn_skills`] for a skill, `trials` for a memory or a
+    /// record.
     ///
     /// The ordinal rather than the rate and the turn, so the pick cannot
-    /// disagree with the arm about which schedule it is on.
+    /// disagree with the arm about which schedule it is on. It also says
+    /// which kind this holdout is for: the three take the schedule in turn
+    /// (`trials::HOLDOUT_ARMS`).
     holdout_ordinal: Option<u64>,
     /// Phase 3 (#714): `context.lifecycle.enabled`, read once at open. While
     /// this is off the learning loop runs exactly the lexical path that ships
@@ -418,6 +425,15 @@ pub struct SessionMemory {
     /// than a plain field because the consumer runs behind `&self` on the
     /// async episode path; it is touched twice a turn and never contended.
     turn_skill_join: std::sync::Mutex<Option<TurnSkillJoin>>,
+    /// The same join for the two kinds recall renders, plus the id this
+    /// turn's holdout settled on ([`trials`]).
+    ///
+    /// Its own cell rather than a share of [`Self::turn_skill_join`] because
+    /// the two are armed in different places: a skill join is noted once, by
+    /// the selection pass at turn start, while a memory or a record is noted
+    /// by every render the turn makes. A `Mutex` for the reason that one has
+    /// one — the consumer runs behind `&self` on the async episode path.
+    context_trials: std::sync::Mutex<trials::TurnContextTrials>,
     /// The session's time source (#2320) — the only one the learning loop is
     /// allowed to read.
     ///
@@ -657,6 +673,7 @@ impl SessionMemory {
                     task_id,
                     execution_id: None,
                     turn_skill_join: std::sync::Mutex::new(None),
+                    context_trials: std::sync::Mutex::new(trials::TurnContextTrials::default()),
                     clock,
                 })
             }
@@ -826,10 +843,16 @@ impl SessionMemory {
     /// shares, rather than the shortlist in front of it, which they do not. A
     /// turn whose pick was not going to be injected anyway holds nothing back:
     /// that costs the experiment samples and cannot cost it correctness. The
-    /// catalog is read only on a turn the schedule armed, which is one turn in
-    /// `context.retrieval.artifact_holdout_rate`.
+    /// catalog is read only on a turn the schedule armed and pointed at
+    /// skills, which is one holdout in the length of `trials::HOLDOUT_ARMS`.
     fn apply_holdout(&self, selection: &mut skills::SkillSelection) -> Option<String> {
         let ordinal = self.holdout_ordinal?;
+        // One item goes per turn, and three kinds share the schedule
+        // (`trials::HOLDOUT_ARMS`), so a turn the schedule points at a
+        // memory or a record holds no skill back.
+        if trials::holdout_kind(ordinal) != Some(stella_learn::ledger::ArtifactKind::Skill) {
+            return None;
+        }
         let loaded = self.load_skills();
         let names: Vec<&str> = loaded.iter().map(|s| s.name.as_str()).collect();
         let held = stella_learn::holdout::pick(ordinal, &names)?;
@@ -944,19 +967,7 @@ impl SessionMemory {
             stella_learn::ledger::ArtifactKind::Skill,
             &join.trigger_matched,
             &join.selected,
-            &stella_learn::skills::appraisal::SkillTrial {
-                task: appraisals::LIVE_WINDOW_TASK.to_string(),
-                // Overwritten per skill by `record_turn`; the value here is
-                // never read.
-                selected: false,
-                outcome: stella_learn::self_tuning::TaskOutcome {
-                    succeeded,
-                    cost_usd: 0.0,
-                    tokens: 0,
-                    retries: 0,
-                },
-                turns: 1,
-            },
+            &trials::live_trial(succeeded),
         );
     }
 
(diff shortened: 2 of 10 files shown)


macanderson:
`dod / dod` was red on this PR because #6096's four DoD boxes were unticked. I checked each one against this diff rather than against the PR description, and ticked them on the issue — `dod-recheck.yml` fires on the issue edit and re-runs the check on this commit, so no push is needed here.

What establishes each box:

1. **Memory trial from the production seam** — `memory/trials.rs`'s `withhold_held_memory` folds the offered population (`recall.frames`, before the plane's budget cut) and the shown set into `TurnContextTrials`; `record_context_trials` writes it from `record_episode`, beside the existing skill call. Both render paths go through it — `recall_block_reported` and the mid-turn `SessionRequery` — which is the `apply_holdout` agreement the issue warned about.
2. **Rule trial the same way** — `turn_records_held` takes the population from `rendered + dropped` on an unexcluded first render, then re-renders through `render_volatile_for_turn_excluding` for what the model reads. That ordering is what keeps the withheld record inside the population rather than invisible to both lists.
3. **Holdout on a schedule, control-arm row written** — `HOLDOUT_ARMS` plus `stella_learn::holdout::arm` rotate the three kinds through one schedule, so one item still goes per turn; `apply_holdout` returns early when the ordinal is not the skill arm.
4. **End-to-end test with the withheld arm** — `memory/trials/tests.rs` drives `arm_controls_at` → `recall_block_reported` → `record_episode` and reads back the JSONL the sweep reads. `the_rule_arm_withholds_a_record_and_writes_its_control_arm` and `the_memory_arm_withholds_a_memory_and_writes_its_control_arm` each assert the `false` row lands on exactly the withheld turn.

The constraints hold too: the new code is a sibling module rather than growth in `recall.rs` (which is net −24 lines here), and `stella-learn`'s only addition is the pure `arm()` function with no new dependency.

Not asserting anything about compile or test results — `fmt + clippy + test` is still running on this commit and is the evidence for that half.

---

macanderson:
Residue filed after the merge: #6128 (one shared schedule means any one kind's control arm comes round every third holdout, so a skill sample now costs 60 eligible turns instead of 20 at the shipped rate) and #6129 (the memory and rule arms withhold silently, while the skill arm prints a line). Neither is a defect in this diff; both are consequences of the one-item-per-turn rule this change had to honour, and both want a maintainer's call rather than a quiet default.

Issue #6072, opened 2026-09-05
The deck demo shows a model table the product cannot produce


macanderson:
Part of #3903 (roleless core), and the demo half of slice 5 (#3909).

## Problem

`crates/stella-tui/src/scenario.rs`'s `demo_engine_config` is the session
behind `stella deck` screenshots, `examples/deck_demo.rs`, and the `/models`
golden frames. It builds six role rows — `default`, `worker`, `verifier`,
`triage`, `research`, `plan` — and sources two of them from keys #3908
retired:

```rust
row("verifier", "anthropic/claude-opus-5", …, "pipeline_verifier_model"),
row("triage",   "z-ai/glm-5.2",           …, "agents.triage.model"),
```

Its own doc comment states the reason: "the three ways a role gets its model
(its own `agents.<role>.model`, a flat `pipeline_<role>_model`, the shared
`default_model`) each appear at least once". That resolution ladder no longer
exists. `config_wiring::deck_rows` — the live producer — emits one row named
`default`, and a plugin-declared seat beside it once one is installed.

So the picture the product shows in screenshots and pins in a golden frame is
a routing table it cannot produce, sourced from keys that now print
"retired and reads nothing" when an operator copies them into a settings file.

`crates/stella-tui/tests/snapshots/deck/card_models.txt` holds the rendered
frame, including the line `from pipeline_verifier_model`.

## Why it was left

`make retired-model-keys` (`scripts/check-retired-model-keys.py`, added in
#6061) records this file in its `DEBT` list at 2. The guard's count for it may
only fall. Fixing the rows re-renders the golden frame, which is the settings
UI's surface, so it belongs with #3909 rather than beside a guard.

## The shape of the fix

Make the demo show what a session shows: one `default` row, plus seat rows for
a demo plugin if the frame is more useful with them. Then regenerate the
golden and **read the diff** — AGENTS.md § Golden frames:

```bash
BLESS=1 cargo test -p stella-tui --test deck_render_snapshots
```

Then lower `crates/stella-tui/src/scenario.rs` to 0 in the guard's `DEBT` dict
and delete the entry.

## Definition of done

- [ ] `demo_engine_config` sources every row from a key the settings schema
      still has.
- [ ] `crates/stella-tui/tests/snapshots/deck/card_models.txt` is regenerated
      and the diff is described in the PR.
- [ ] `crates/stella-tui/src/scenario.rs` is gone from `DEBT` in
[...]

PR #6107, opened 2026-09-05
feat(stella-observatory,harbor_adapter): read the seat from the trace, retire the four-language role guard

macanderson:
## What and why

`scripts/check-role-names.sh` was a `make gate` step pinning four agent-config
role spellings across Rust, Python and JavaScript. Its header names the failure
it existed for: #1394 renamed `judge` to `verifier`, the Rust compiler caught
every Rust call site and none of the others, and the contest published a number
for a pairing it never ran.

Slices 2 (#3906) and 3 (#3907) of epic #3903 closed that hole by construction.
Every call the host spends for a plugin books at `ModelCallRole::Plugin`, and
the seat the plugin chose rides on the `sub_agent` bracket as data
(`SubAgentPhase::Started::seat`). A consumer reads the word that actually ran
instead of a spelling compiled into the reader.

This slice gives the two readers that path, then retires the guard.

## What the guard held, and what answers each half now

The guard named three live producers.

**1. `crates/stella-observatory/src/assets/index.html`** filtered
`agent_engine_config.agents` through the literal
`["default", "worker", "verifier", "triage", "research", "plan"]`. Any other
key drew no row, so a settings file that configured something read as one that
configured nothing. The pane now takes its rows from the keys the merged
settings carry (`Object.keys`, so a `__proto__` in a settings file cannot
invent a row). No role word is quoted anywhere in the page any more.

**2. `bench/harbor_adapter/stella_harbor/posture.py`** still writes
`pipeline_verifier_model`. That key is inert on both sides since #3908: Stella
recognises it, reports it by name, and reads nothing from it, so a drifted
spelling cannot misconfigure a seat. It would name a key outside
`ENGINE_ROOT_FIELDS` instead, and
`config::trusted_engine_config_shape_is_strict` fails **closed** on that — a
refused launch. #1394's failure was silence; there is none left here.

**3. `crates/stella-tui/src/deck.rs`'s `PipelineRole`** is three display slots
already held to `ModelCallRole` by an exhaustive match in `PipelineRole::of`.
No second language is involved.

## The seat-from-trace path

- `crates/stella-observatory/src/turn_agents.rs` projects the bracket's `seat`
  onto each child's row. It was dropping the field, so the turn page had
  nothing to draw whatever the trace said.
- `index.html` badges that seat on the sub-agent row and names it in the
  focused sub-agent header.
- `bench/harbor_adapter/stella_harbor/atif.py` joins the same brackets by
[...]

PR #6116, opened 2026-09-05
fix(stella-tui): stop the /models dialog naming five retired roles

### LINKED ISSUE #6087: The models dialog still names five roles the engine can never send

macanderson:
## What & why

`crates/stella-tui/src/views/models_card.rs`'s `SLOTS` table named six roles —
`default`, `triage`, `research`, `plan`, `worker`, `verifier` — as the words
the `/models` dialog prints. `crates/stella-cli/src/config_wiring.rs::resolve()`
has sent exactly one of them, `default`, since #3908 collapsed the rest. The
other five described a vocabulary the engine can never send.

`SLOTS` shrinks to the one entry `resolve` can produce (renamed `KNOWN`,
since it is no longer a set of slots). `slot_of` and `standing` go with it:
both existed only to mark a role `served`/`configured`/`unassigned` against
`PipelineRole`, and the driver has never sent `"worker"`, `"triage"` or
`"verifier"` as a role key for `slot_of` to match — the mark could never
fire on a real session, only on the fixture that spelled those keys.

`scenario.rs`'s `demo_engine_config()` — the `/models` golden's fixture and
the `stella deck` demo data — was itself drawing six rows from retired keys,
including the two `pipeline_verifier_model`/`agents.triage.model` string
literals `scripts/check-retired-model-keys.py`'s `DEBT` dict was carrying.
It now sends the one row `resolve` can actually produce, clearing that debt
to empty (`DEBT` is now `{}`).

Closes #6087

Deviation from the design pointer, explained on the issue before I started:
I did not drive `models_card.rs` from `views/seats.rs`'s fold over
`state.roles` + `state.seats`. That would make the TUI dialog and
`stella config`'s text output disagree about which roles exist — an
invariant `config_wiring.rs`'s own module doc says must never happen, and
the design question #6088 already owns. The existing `role_table` fold
(#3472) already renders any role outside `KNOWN` under its own key, so the
dialog stays open to whatever #6088 decides without this PR pre-empting it.

## The witness

`models_card::tests::a_contributed_role_is_printed_and_leaves_with_its_contributor`
fails on `main`: it calls `role_table(&state, &KNOWN)`, and `KNOWN` does not
exist on `main` (only the six-entry `SLOTS` does) — a compile failure. On this
branch it passes, and its later half — a fixture reverting to the plain
`state()` (one `default` row) drops the contributed role from the rendered
text — is the same fold behaviour #3472 already proved, now proved again
against the shrunk table.

`deck_snapshot.rs`'s `Card::Models` needles move from `["models", "think",
"work", "verify"]` to `["models", "lead"]`, matching what the demo driver
(one `default` role) can actually produce.

The `card_models` golden is deleted in the first commit rather than
hand-authored — I have no compiler on this machine, only CI does — and a
second commit on this branch re-blesses it against the one-row demo wiring
(read and confirmed correct: `lead`, `zai/glm-5.2-air`, the `effort_auto`
disclosure, `from agents.default.model` — exactly `scenario.rs`'s fixture).

Tests deleted: `models_card::tests::a_configured_role_is_not_drawn_as_one_that_ran`
(exercised `standing`/`slot_of` directly, both deleted in the first commit —
the feature it tested no longer exists) — `check-deleted-tests` names this
one explicitly and this line answers it.

## The gate

- [ ] `fmt --check`
- [ ] clippy across the workspace at `-D warnings`
- [ ] the workspace test suite
- [x] `Closes #N` appears **both** above and as a commit trailer — squash builds
      the commit body from commit messages, so the description alone won't carry it

(The first three are CI's job here, not this laptop's — `make guards-fast`,
a workspace-wide `fmt --check`, `scripts/check-prose.py`,
`scripts/check-retired-model-keys.py`, `scripts/check-file-size.sh` and
`scripts/check-line-citations.py` all ran locally and are green; clippy,
rustdoc and the test suite run in CI.)

## Nothing left behind

Anything you noticed and did not fix — a bug, a missing test, dead or unwired
code, the logical next step of this work — is a GitHub issue before this merges,
written as a handoff a fresh agent could execute (AGENTS.md § "Nothing left
behind"). The gate catches the residue (a `TODO` with no issue number); it
cannot catch what only you saw.

One pre-existing inaccuracy noticed, not filed as new: #6087's own "What is
not in scope" section says retiring `PipelineRole` "is #3910's work" — #3910
closed via #6107 without retiring the enum (it retired the now-redundant
cross-language `check-role-names.sh` guard instead, per that PR's own
description). `PipelineRole` genuinely still needs a home for retirement;
I left `crates/stella-tui/src/deck.rs`'s doc comment on it factual and
un-numbered rather than pointing at a closed issue that didn't do that work,
and didn't file a new issue since no code here depends on it — a maintainer
call, not a defect this diff introduces.

## Ground-rule check

The `card_models` golden moved: the card is now four lines shorter (one role
instead of six), so the floating card's overlap with the transcript beneath
it shifted too. That is the whole diff in the golden — no other pane changed.


macanderson:
**Independent check of the `card_models` golden in `81716b3` — it is byte-identical to what the code renders.**

The description flags this as the one open question ("I have no compiler on this machine, only CI does … re-blessed from CI's own rendered output"). Settled, on a compiler rather than by reading:

```
$ BLESS=1 cargo test -p stella-tui --test deck_render_snapshots
test result: ok. 41 passed; 0 failed; 2 ignored
$ diff <git show 81716b3:…/card_models.txt> <blessed output>
(no differences)
```

Blessed from `7aa777a` (the parent, with the golden still deleted) in a clean worktree, so the frame came from `render_deck` and not from the committed file. `git status` reported `card_models.txt` as the **only** regenerated snapshot — none of the other 40 goldens moved, which is the part worth knowing: this diff perturbs no frame but its own.

The frame itself reads the way the change intends — one `lead` row carrying `zai/glm-5.2-air … from agents.default.model`, above the `session`/`auto` rows. None of `think`, `work`, `verify`, `triage`, `research` or `plan` appears anywhere in it, so `deck_snapshot.rs`'s needle move to `["models", "lead"]` matches the real output rather than an assumption about it.

Whole crate green on that tree as well, not just the snapshot target:

```
$ cargo test -p stella-tui
1508 passed; 0 failed; 2 ignored   (lib)
plus 14 integration binaries, all ok; exit 0
```

No action needed — this is confirmation, not a request. `fmt + clippy + test` in CI is still the authority for the workspace-wide clippy and rustdoc legs this does not cover.

---

macanderson:
Residue filed after merge: #6131 — retiring `PipelineRole::{Triage,Verifier}` in `deck.rs`, since #3910 closed (via #6107) without doing that work despite this PR's own body pointing at it. Also closed #6072 as a duplicate of this PR's `scenario.rs` fix, found while filing residue.

Issue #6057, opened 2026-09-05
Tools and plugins never reach the steering plane, so the biggest cost on the wire is outside the one budget


macanderson:
The steering plane can rank a tool or a plugin against a rule, a skill and a memory. Nothing ever hands it one.

`SteeringSource` in `crates/stella-core/src/steering.rs` has five arms. `Record`, `Skill` and `Memory` all have a production adapter in `crates/stella-cli/src/memory/steering.rs`. `Tool` and `Plugin` have none. They have a priority in the budgeter's ordering and a render arm that returns `None`, and that is all.

Where the arms show up today:

- `crates/stella-core/src/steering.rs` — `SteeringSource::Tool => 1` and `SteeringSource::Plugin => 4` in the source ordering.
- `crates/stella-cli/src/memory/recall.rs` — `SteeringSource::Tool | SteeringSource::Plugin => None` in the section renderer, so even a candidate that arrived would draw nothing.
- `crates/stella-core/src/steering/tests.rs` and `crates/stella-cli/src/memory/tests/steering_selection.rs` — the only two places a `Tool` candidate is ever built, and both are tests.

So the budgeter is ranking three sources against each other while the largest cost on the wire sits outside it. #3243 measures tool advertisement at about 6,000 to 9,000 tokens on every model call, against a 2,000-char record channel that the same plane budgets carefully.

This is Phase 4 and Phase 5 of #3243. It had two feeder issues, #3033 (send tool names as an ordered menu) and #1856 (lean tool catalog is off by default), plus #3245 for plugins. All three are closed, so nothing open tracked the join until this issue.

## Scope

Two separate joins. Split them if the tool half turns out to be the whole of a PR.

1. **Tools.** Make tool advertisement a plane consumer. `crates/stella-core/src/driver.rs` calls `self.tools.schemas()` unfiltered on every model call. The lean path and activation already exist in `crates/stella-cli/src/discovery.rs`. The plane's job is the ordering and the budget, not a new ranker.
2. **Plugins.** A plugin's steering-grade context injection lands through the plane rather than beside it, per #3245's manifest-declared loop-participation grade.

## Constraints

- Invariant 2: the decision logic stays in `stella-core`, the I/O stays in the adapter. `memory/steering.rs` is the shape to copy.
- Invariant 7: nothing the plane returns may enter the byte-stable prefix. Tool schemas are sent per model call, not in the prefix, so read `driver.rs`'s call path before assuming where a change is safe.
[...]

Issue #6112, opened 2026-09-05
Plugins still never reach the steering plane, so nothing costs or reports what a plugin injects


macanderson:
`SteeringSource::Plugin` is an arm of the steering plane (`crates/stella-core/src/steering.rs`) with a rank in the budgeter's precedence, a tag, and no producer. `crates/stella-cli/src/memory/recall.rs`'s `drop_message` returns `None` for it, so even a candidate that arrived would draw nothing.

#6057 asked for tools and plugins together and shipped the tool half (`crates/stella-core/src/steering/tools.rs`, `crates/stella-cli/src/tool_lean.rs`). The plugin half is here because the two joins have nothing mechanical in common:

- **Tools** were already a port. The join was a `ToolExecutor` decorator over `schemas()` — a filter over an existing surface, with an existing cost (the schema text) and an existing budgeter.
- **Plugins** have no such surface. An installed plugin's context contribution arrives over the wrapper socket (`crates/stella-runtime/src/wrapper/`, `doc:wrapper-socket`), from `before_turn`, as text the plugin authors. Nothing today costs it, ranks it, or reports it — and #3245's manifest-declared loop-participation grade is what says a plugin may contribute at all.

## What the join needs

1. **A producer.** `before_turn`'s returned context becomes `SteeringSource::Plugin` candidates: one per contribution, handle = the plugin id (and the declared stage, when a plugin contributes at several), `est_tokens` measured over the exact text that would be injected, `why` naming the plugin and the stage.
2. **The grant is the gate.** A plugin contributes only at the points its manifest declared and the operator accepted (`LoopGrant::permits_point`, `crates/stella-plugin/`). A candidate from an ungranted point is refused, not budgeted.
3. **A render arm.** `drop_message`'s `SteeringSource::Plugin` arm names the plugin and the remedy, the way the tool arm now does.
4. **The precedence has to be argued.** `source_rank` puts `Plugin` last "until #3246 gives it an authority story". A third-party plugin outranking a published record would be a governance failure; below every first-party source is the safe direction and probably the right one, but it should be written down rather than inherited from a placeholder comment.

## Files

- `crates/stella-core/src/steering.rs` — the arm, the rank, the budgeter.
- `crates/stella-core/src/steering/tools.rs` — the shape a producer takes (pure, no I/O, invariant 2).
- `crates/stella-cli/src/tool_lean.rs` — the shape an adapter takes.
[...]

Issue #6111, opened 2026-09-05
A delegated child advertises every tool even when its parent session is lean


macanderson:
`crates/stella-cli/src/tool_lean.rs` budgets the tool schemas a session advertises, wired into the chain by `crates/stella-cli/src/agent/tool_stack.rs`. Every session-shaped door gets it — the one-shot turn, the goal loop, a resume, the deck's lead turn, a subsession lane, a fleet worker — because they all build their chain through `session_stack` or `policy_stack`, which read `Config::tool_advertisement`.

A dispatched child does not. `crates/stella-cli/src/subagent.rs`'s `run_child` assembles the child's chain inside the child's own thread through `policy_stack_with`, and passes `ToolAdvertisement::Full` with a comment naming this issue. So a `delegate` child and a best-of-N candidate advertise every schema even in a lean session.

## Two questions, and they are separate

1. **Should a child inherit the parent's allowance at all?** A child runs one narrow goal against a rooted registry. Its useful tool set is smaller than the parent's, which argues for a *tighter* allowance rather than the same one. It also runs unattended, and a child that cannot see a tool will not ask for it, so a wrong answer here is invisible in a way the parent's is not.
2. **How does the value reach the child's thread?** `run_child` takes `(ToolPolicy, Arc<dyn AuthzGate>, Principal)` as one tuple, deliberately, because a `GatedToolSet<'_>` cannot cross a thread boundary and only the owned pieces can. `ChildToolPosture` (`crates/stella-cli/src/subagent.rs`) is the struct that already carries the policy and the gate for the `delegate` path and is the obvious home for a fourth member; `dispatch_in_workspace` takes its policy as a bare argument and would need one too.

## Verify

With `context.steering.tools.lean = true` and a small `max_tokens`, run a turn that calls `delegate`. The parent's request carries a budgeted tools array; the child's carries every schema. Both are visible in `.stella/private/store.db`'s `step_manifest` events, or through a recording proxy.

## Constraints

- Withholding a schema is never a capability gate: a child that names a withheld tool must still run it, exactly as the parent does.
- Invariant 7: whatever the child advertises must be fixed for the child's whole turn.
- `crates/stella-cli/src/agent/tool_stack.rs` is the one place a chain is assembled; the answer belongs there and not in a second composition.

## Definition of done

- [ ] A decision is written down for question 1, in `tool_stack.rs`'s or `subagent.rs`'s module docs, with the reason.
[...]

Issue #6110, opened 2026-09-05
Records and tools still spend two budgets, so a rule and a tool schema are never weighed against each other


macanderson:
The steering plane has five sources and two budgets. Records, skills and recalled frames are packed by `GatheredSteering::query` (`crates/stella-cli/src/memory/steering.rs`); tools are packed by `stella_core::steering::tools::advertise` against their own allowance (`context.steering.tools.max_tokens`). Both call the same budgeter, `pack_to_budget`, and both honour the same `source_rank` precedence — but they never see each other's spend, so a rule and a tool schema are still never weighed against one another.

#6057 shipped the tool arm and said so plainly rather than claiming the join was complete. This issue is the join.

## Why it was not done there

The two costs are known at different moments in the turn.

- The recall block is rendered once, before the turn opens, in `memory::recall`.
- The tools array is assembled per model call, inside the engine, from `ToolExecutor::schemas()` (`crates/stella-core/src/driver.rs`, `run_model_call`).

There is no point at which both are known, so no single call to `pack_to_budget` can decide both.

## The shape that would work

A per-turn ledger both consumers spend against, in `source_rank` order:

- One allowance for the turn's whole volatile context, held as a shared per-turn value (an `Arc` over an atomic total is enough; it must be cheap and it must not need a lock held across an await).
- The recall path spends first and records what it took.
- The tool layer (`crates/stella-cli/src/tool_lean.rs`) packs against what is left.

The threading is the work: the recall block and the tool stack are built at different call sites in every driver (`agent::run_turn`, the goal loop, `agent::resume`, the deck's lead turn, a subsession lane, a fleet worker), so the handle has to reach both from the one place that owns the turn.

## Constraints

- Invariant 2: the arithmetic stays in `stella-core`, the threading stays in `stella-cli`.
- Invariant 7: the advertised tools array must stay fixed for a session. A running total that shrinks as the turn proceeds would re-rank the array mid-turn and re-bill the whole conversation, so the tool allowance must be settled once and not re-read per step.
- With `context.steering.enabled` off, nothing may be withheld that the plane did not add.
- The golden block test (`crates/stella-cli/src/memory/tests` `golden_block`) must stay byte-identical while `context.steering.tools.lean` is false.

## Definition of done

- [x] One per-turn allowance is spent by both consumers, in `source_rank` order, with no second budget number.
- [x] The tool allowance a session resolves is settled once per session, not recomputed per step.
[...]

PR #6117, opened 2026-09-05
feat(stella-core,stella-cli): budget tool schemas on the steering plane

macanderson:
## What

Tool advertisement becomes a steering-plane consumer, behind a lever that ships off.

The plane budgets a 2,000-char record channel with care and never counted the biggest block on the wire: every model call carries the whole tool schema array. `SteeringSource::Tool` has had a rank in the budgeter's precedence since the plane landed, and no producer. Two of the five arms now have one.

- **`crates/stella-core/src/steering/tools.rs`** (new, pure — no I/O, invariant 2). One `SteeringCandidate` per schema, costed over the three fields a provider is sent (name, description, serialized input schema). The rank is two facts and no third: a built-in beats an MCP server's tool, and inside a group the cheaper schema goes first. `advertise()` then packs through `pack_to_budget` — the plane's one budgeter, which is what puts a tool and a rule in the same unit — and returns the surviving schemas **in arrival order**, plus the `SteeringSet` for the drop report. A separate aggregate MCP ceiling runs first, so one chatty server cannot spend the allowance the built-ins need.
- **`crates/stella-cli/src/tool_lean.rs`** (new). A `ToolExecutor` decorator in the shape `tool_policy::PolicyToolSet` established. It narrows `schemas()` and nothing else: `execute`, `contracts` and `parallel_safe_names` forward whole, so a tool the budget could not describe still runs when it is named, keeps its reviewed contract at the authorization gate, and keeps its concurrency claim. Hiding a schema saves prompt tokens; it is never a capability gate.
- **`crates/stella-cli/src/agent/tool_stack.rs`**. The layer is composed in the one place a session's chain is assembled, under the gate and over the policy layer. With the lever off **no layer is composed at all**, so "off advertises the whole surface" is a fact about the composition rather than a claim about a filter.
- **`crates/stella-cli/src/settings/context.rs` + `settings/steering.rs`**. `context.steering.tools.{lean,max_tokens,mcp_max_tokens}`, defaulting to `false` / 6000 / 2000, resolved once into `Config::tool_advertisement` beside `Config::tool_policy`. `STELLA_TOOLS_LEAN` overrides the settings chain the way `STELLA_CONTEXT_STEERING` does, so a bench arm is picked by the harness that starts the run rather than by an edit to the tree it measures. With `context.steering.enabled` off, `tool_advertisement()` returns `Full` outright — the master switch must stay a clean control arm, not a capability change.
- **`crates/stella-cli/src/memory/recall.rs`**. `drop_message`'s `Tool` arm draws a sentence instead of returning `None`: it names the tool, says it still runs if called, and names the knob that would have advertised it. It reaches stderr through `report_steering_drops`, the single writer every other source already uses.

Byte-stability (invariant 7): the rank reads the session's resolved tool set and nothing else, and the advertisement is computed once per stack. The array cannot move between two turns of one session, which matters because the tools array sits ahead of the system prompt in every provider's cache prefix.

Exemplar followed: `crates/stella-cli/src/tool_policy.rs`'s `PolicyToolSet` — the decorator shape, the method-by-method forwarding contract, and the doc convention explaining why each forward is not optional.

## Witness mechanism

Four, each failing on `main` by construction.

1. `stella_core::steering::tools::tests::a_tool_the_allowance_cannot_afford_is_withheld_and_named` — forty schemas against a quarter-sized budget. The module does not exist on `main`.
2. `stella_cli::tool_lean::tests::a_lean_session_advertises_only_what_the_allowance_affords` — the same fixture through the decorator: the advertised schema bytes fall under the ceiling while the layer beneath still resolves all forty. Nothing else in the chain narrows `schemas()`, so without this layer every schema the stack resolves reaches the request.
3. `stella_cli::memory::tests::steering_selection::every_dropped_source_gets_a_line_with_its_own_remedy` — **an existing test, edited**. It asserted four report lines with a `Tool` drop present producing none. It now asserts five and pins the tool sentence's three claims. It fails on `main`, where `drop_message` returns `None` for that arm.
4. `stella_cli::settings::steering::tests::the_plane_being_off_sends_every_tool_even_with_the_lever_on` — the master switch settles it; `tool_advertisement()` does not exist on `main`.

The control arm for "off is byte-identical" is structural rather than asserted: `budgeted()` returns the chain it was given for `ToolAdvertisement::Full`, so a default session composes exactly the chain it composed before. `a_wide_allowance_advertises_the_base_list_unchanged` pins the filter's own identity behaviour beside it.

## What this PR does not do, and why

Two of this issue's own boxes stay unticked, and both are filed rather than fudged.

- **One shared budget across all five sources.** Records, skills and frames are packed by `GatheredSteering::query`; tools are packed against their own allowance. Both call `pack_to_budget` and both honour `source_rank`, but they never see each other's spend. They cannot: the recall block is rendered once before the turn opens, and the tools array is assembled per model call inside the engine, so there is no moment at which both costs are known and one number could bind them. The shape that would work — a per-turn ledger both consumers spend against in precedence order — is #6110, with the threading spelled out.
- **Plugins.** `SteeringSource::Plugin` still has no producer. The tool join was a filter over an existing port with an existing cost; the plugin join is a new producer over the wrapper socket's declared stages, gated by the manifest's accepted contribution points, and it needs the precedence question answered rather than inherited from a placeholder comment. That is #6112.

A third gap is marked in the diff: a dispatched child (`subagent.rs`'s `run_child`) advertises the full surface, because it assembles its chain inside the child's own thread with no `Config`, and whether a child should inherit its parent's allowance is a separate question. #6111.

## Tests deleted

None.

## Residue issues filed

- #6110 — records and tools still spend two budgets.
- #6111 — a delegated child advertises every tool even when its parent is lean.
- #6112 — plugins still never reach the steering plane.

Refs #6057

> **Issue link downgraded to `Refs` by the autonomous PR routine.**
> SCR-003: a PR that advances an issue without finishing it links it with
> `Refs`, so the merge gate does not hold it against a definition of done it
> does not complete. This PR meets five of #6057's six boxes — the production
> `Tool` adapter and the render arm, the per-tool `why` in the drop report,
> the witness test, and the plugin split with its reason written down
> (#6112). The sixth — "tool cost enters one budget with the other sources
> ... rather than in two passes" — is the one this PR's own "What this PR
> does not do" section explains it does not do, and #6110 tracks. `Refs` is
> the honest link for that; #6057 stays open for #6110 to finish.

## Summary by Sourcery

Add an opt-in steering-plane budget for tool schemas without changing tool capabilities or the default advertised surface.

New Features:
- Add optional, budgeted tool-schema advertisement to the steering plane, with separate overall and MCP allowances and deterministic prioritization.
- Add configuration for lean tool advertisement, including settings resolution, reload support, and an environment-variable override.
- Report withheld tool schemas with actionable guidance while preserving their execution, authorization contracts, and concurrency behavior.

Enhancements:
- Compose the advertisement layer only when enabled so the default control path continues to expose the complete tool surface unchanged.
- Keep advertised schemas stable and in arrival order while recording selection and drop information for steering diagnostics.

Documentation:
- Document the new tool-advertisement settings and their default behavior in the example configuration.

Tests:
- Add coverage for budget enforcement, prioritization, MCP limits, order stability, execution forwarding, drop reporting, settings resolution, and master-switch behavior.

Chores:
- Leave shared records-and-tools budgeting, delegated-child inheritance, and plugin producers for follow-up issues.


macanderson:
Automated PR sweep — one gap that needs a decision rather than a fix, and it is the same gap seen from two directions.

Sourcery's assessment table has one ❌:

> Tool cost enters one budget with the other sources… The PR gives tools their own allowance and uses a separate pack operation from the record/skill steering budget. It explicitly defers the shared per-turn ledger to issue #6110.

`dod / dod` is red for what is the same item, #6057's second box:

> Tool cost enters one budget with the other sources, so a rule and a tool schema are ranked against each other rather than in two passes.

Sourcery is reading the diff correctly here, and the PR description agrees with it — deferring the shared ledger to #6110 is stated as a deliberate choice, not an oversight. So this is not a ❌ to rebut and not a box to tick.

What follows is the keyword. The PR body carries `Closes #6057`, and SCR-003 says a PR that advances an issue without finishing it links it with `Refs` instead, so the merge gate does not hold that PR against the issue's DoD. On the PR's own account #6057 is advanced, not finished. Two ways to settle it:

- **`Closes #6057` → `Refs #6057`.** Nothing else changes; #6057 stays open carrying the one unfinished box, and #6110 remains where the shared-ledger work lives. `dod-check` fires on `edited`, so the body edit alone re-runs it.
- **Split the box out** into its own issue under #6110's umbrella (`triage` label only, SCR-005), tick #6057's remaining boxes, and keep `Closes`.

The first is the smaller move and matches what the description already argues. Either way this is the author's call, not something a sweep should decide — flagging rather than editing.

Nothing else is outstanding: the other two assessment rows are ✅, there are no unresolved review threads, the branch is conflict-free against `main`, and `main` is not known-broken. I have not looked at compile or test results — CI was still running on `b628bf4`… sorry, on this PR's head when this was written, and that is the evidence for that half.

---

macanderson:
**PR-management sweep — two linked findings, one of them blocking.**

**1. `dod / dod` is red, and it is right.** All six of #6057's DoD boxes are still unchecked (issue last edited 21:14Z, before this PR opened). The two sibling PRs in this batch have already settled theirs — #6071 was ticked item-by-item with evidence at 21:38Z for #6115, and #6096 at 21:36Z for #6114 — so this is the one left.

**2. Sourcery's ❌ on objective 2 is correct, and it is the same fact.** Its row reads:

> Tool cost enters one budget with the other sources … ❌ The PR gives tools their own allowance and uses a separate pack operation from the record/skill steering budget.

That matches this PR's own text and #6110's problem statement — `GatheredSteering::query` and `stella_core::steering::tools::advertise` both call `pack_to_budget`, but against two allowances that never see each other's spend. #6110 exists, is written as a proper handoff, and states plainly why the join could not land here: the recall block is rendered before the turn opens and the tools array is assembled per model call inside the driver, so there is no moment at which both costs are known.

So the ❌ is not a defect in the diff — it is scope this PR deliberately did not take, already tracked. Under CLAUDE.md's Sourcery rule that earns an answer rather than a fix, which is what this comment is.

**What it does mean is that `Closes #6057` overstates what landed.** DoD item 2 *is* #6110's entire subject; #6057 cannot be closed while it stands. SCR-003 has the remedy: a PR that advances an issue without finishing it links it with `Refs #N`, and `Refs` does not close, so the merge gate stops holding this PR against that DoD.

Two ways to green, author's call:

- **`Refs #6057`** (description and commit trailer both), leaving #6057 open with item 2 pointing at #6110. Nothing else changes.
- **Keep `Closes #6057`** and first amend the issue: tick items 1, 3, 4, 5 and 6 with the evidence for each, and rewrite item 2 to record that the join was split to #6110 with the reason. That is the shape #6071 and #6096 just used.

Either edit re-runs the check on its own — `dod-recheck.yml` fires on an issue edit, and a push re-runs it for a description change — so no empty commit is needed.

Nothing else on this PR is blocking: `main is not known-broken`, `file size ratchet`, `docs guards`, `guard self-tests`, `no edit is undone inside the branch` and `cargo check on the declared MSRV` are all green on `f470b99`; `fmt + clippy + test` was still running when this was written.

---

macanderson:
Automated PR sweep. Two checks are red on `f470b99`, and one of them has a trap in it worth naming before the obvious fix gets applied.

## 1. `the macOS-only dependency table still compiles` — 9 errors, all in test targets

Reproduced locally on this head, exit codes read rather than grepped:

```
$ cargo check -p stella-cli                 → exit 0
$ cargo check -p stella-cli --all-targets   → exit 101, 9 errors
```

That gap is the whole story: `session_stack_with_gate` gained its 7th parameter (`advertisement: ToolAdvertisement`) and `Config` gained `tool_advertisement`, and both production callers were updated (`tool_stack.rs` at the two `cfg.tool_advertisement` sites). The shipping build is clean. Only the test targets were never compiled, which is why `macos-check` — the one job that runs `--all-targets` early — is the first to report. `fmt + clippy + test` was still running when this was written and will land on the same 9.

| Error | Site |
|---|---|
| E0061 ×3 (6 args, takes 7) | `agent/tool_stack.rs:514`, `:588`, `:638` — all inside `mod tests` |
| E0063 ×5 (missing field) | `command_deck/model_cmd.rs:258`, `command_deck/session_override.rs:404`, `config/tests/resolution.rs:8` and `:90`, `fleet_cmd/tests.rs:624` |
| E0027 ×1 (pattern) | `config/reload/completeness.rs:64` |

Eight of those nine are mechanical.

## 2. The ninth is not mechanical, and rustc's suggested fix would defeat a guard

`completeness.rs` destructures `Config` exhaustively **with no `..` rest pattern**, on purpose: its module doc says adding a field must stop the module compiling "until its author says what `/reload` does with it." rustc offers three fixes and two of them — `, .. }` and `tool_advertisement: _` — silence the guard rather than answer it.

Answering it needs a fact that is currently missing: **`reload_from_disk` does not re-derive `tool_advertisement`.** Its phase-1 block derives `tool_policy`, `ignore_gitignore`, `reward_policy`, `plan_review`, `create_worktrees` and `hooks`; phase 2 commits those six. `tool_advertisement` is in neither. So today the field is set at `Config::load` and keeps its session-start value for the rest of the session while `/reload` reports success — which is the `ignore_gitignore` shape (#3895) that this guard was built after.

My reading, offered as a proposal since the posture is yours to declare:

- **`Posture::Reloaded`**, matching `tool_policy` — the sibling this PR's own commit message names as the shape `LeanToolSet` follows. `Settings::tool_advertisement()` derives from `context.steering.tools.lean` in the scope chain, so a settings document genuinely can move it, and `EVERY_RELOADABLE_KEY` can reach it.
- Not `ReloadedFromEnv`, even though `STELLA_TOOLS_LEAN` overrides it: that variant is for a field derived from the environment *rather than* a settings document, and this one falls back to the chain whenever the variable is unset.
- Which means a `let tool_advertisement = settings.tool_advertisement();` in phase 1 and a `self.tool_advertisement = tool_advertisement;` in phase 2, plus the ledger row and the reloadable key — not just a name added to the pattern.

## 3. `dod / dod` is separately red

#6057 has 6 unchecked DoD items, including "a production adapter emits `SteeringSource::Tool` candidates, and `recall.rs`'s render arm draws them instead of returning `None`" and the witness test for a tool the budget cannot afford. This PR carries `Closes #6057`, so it is held to all of them.

## What I did and did not do

**No changes pushed.** The eight mechanical fixes would take a minute, but the ninth is a design call this guard exists to route to you, and pushing the mechanical eight alone would leave the branch red on the one that matters while costing a CI cycle. The DoD gate above also means the PR cannot merge on a compile fix alone. Everything else on this head is green: `cargo deny`, MSRV, `docs guards`, `file size ratchet`, `guard self-tests`, `main is not known-broken`, `no edit is undone inside the branch`, `PR carries a diff`, `docs/wire still matches the types`, `harbor_adapter + analyzer pytest`. No merge conflict, no `hold` label.

---

macanderson:
Confirming the compile break is fixed on `98e6eb1` — `the macOS-only dependency table still compiles` and `dod / dod` are both green on this head. I reached the same `Reloaded` posture for `tool_advertisement` independently before your push landed, for the same reason (it comes out of `settings.tool_advertisement()` exactly as `tool_policy` does, so a `StartupOnly` row would have been claiming that re-deriving it is *wrong*).

One addition to `a_reload_re_derives_every_field_it_claims_to`, not a blocker:

`tool_advertisement()` reads `STELLA_CONTEXT_STEERING` and `STELLA_TOOLS_LEAN` **before** it reads the document — `steering_enabled()` short-circuits to `Full`, and `lean` is `env_toggle("STELLA_TOOLS_LEAN").unwrap_or(tools.lean)`. Either one set falsy in the ambient environment pins the field at `Full`, `EVERY_RELOADABLE_KEY`'s new `context.steering.tools` block cannot move it, and the test fails as *"`tool_advertisement` is declared Reloaded and `Config::reload_from_disk` did not move it"* — reporting a dirty environment as a dropped field, which is the failure mode the existing `TRUSTED_ENGINE_CONFIG_ENV` capture in that same test guards against.

Nothing in the tree sets either var (`settings/completeness.rs` explains why: `set_var` is a race under the concurrent runner), so this is not an in-process hazard. It is an ambient one, and `tool_advertisement`'s own doc comment names the scenario — *"a bench arm is picked by the harness that starts the run"*. A harness that exported `STELLA_TOOLS_LEAN` would make this test fail on an unrelated branch.

Two lines, folded into the guard already there:

```rust
let _restore = crate::test_env::EnvRestore::capture(&[
    super::super::TRUSTED_ENGINE_CONFIG_ENV,
    "STELLA_MANAGED_SETTINGS",
    "STELLA_CONTEXT_STEERING",
    "STELLA_TOOLS_LEAN",
]);
// SAFETY: guarded by `test_env::lock`, and restored by `_restore` even on
// an unwinding panic.
unsafe {
    std::env::remove_var(super::super::TRUSTED_ENGINE_CONFIG_ENV);
    std::env::remove_var("STELLA_CONTEXT_STEERING");
    std::env::remove_var("STELLA_TOOLS_LEAN");
}
```

Your call whether that is worth a push — the PR is green without it.

Evidence for the ledger row being a real witness rather than a vacuous one: with the `self.tool_advertisement = tool_advertisement;` line removed from `reload.rs`, `cargo test -p stella-cli a_reload_re_derives_every_field_it_claims_to` exits 101 and panics naming `tool_advertisement`; restored, it passes. `cargo check -p stella-cli --all-targets` exits 0 on this head (it was 9 errors on `f470b99`).

---

macanderson:
## Answering Sourcery's ❌ — "one cross-source budget"

The row is correct about the diff, and the objective is deliberately out of scope. It is deferred into #6110, and the reason is structural rather than a matter of effort.

The two costs are known at different moments in the turn. The recall block — records, skills, frames — is rendered once, before the turn opens, in `memory::recall`. The tools array is assembled per model call, inside the engine, from `ToolExecutor::schemas()` in `driver.rs`'s `run_model_call`. There is no point at which both are known, so no single `pack_to_budget` call can decide both, and a PR that claimed otherwise would be moving one of them to a place it does not belong.

What this PR does deliver toward the objective, and what it does not:

- Tool cost is measured in the same unit as every other source (`stella_protocol::estimate_tokens`) and packed by the same budgeter (`pack_to_budget`), under the same `source_rank` precedence. So a tool and a rule are already commensurable and already ordered against each other.
- What they do not yet do is see each other's spend. That needs a per-turn ledger both consumers draw from in precedence order, threaded from the one place that owns the turn to two call sites in every driver. #6110 writes that out — the shape, the threading, the two constraints that bound it (the tool allowance must be settled once per session or the array re-ranks mid-turn and re-bills the conversation; the golden block test must stay byte-identical with the lever off), and its own witness.

The PR description names this under "What this PR does not do, and why", and the `GatheredSteering` doc comment in `memory/steering.rs` now says the same thing where the next reader of that code will find it, instead of the stale sentence it carried claiming the budget would start binding when the tool arm joined.

The remaining two rows are ✅ and need nothing.

Issue #6122, opened 2026-09-05
Deck sub-sessions get no context at all, and nobody measured whether that is still right


macanderson:
A deck sub-session runs the engine with no recall at all. `crates/stella-cli/src/subsession.rs`'s module header states it as a v1 scope choice: native tools only, no MCP, no custom tools, and "recall is skipped in favor of latency". So a sub-session gets the byte-stable prefix and nothing volatile — no recalled frames, no selected skills, no matched context records, and no today's date.

That was decided when a sub-session was a short answer to a queued prompt. Two of them are no longer short: a `task_assign` worker (`sub:<task-id>`) runs a whole board task, and a dispatched request worker (`req:<n>`) runs whatever the person typed while the lead was busy. Both can be long turns that drift, which is exactly the shape mid-turn re-query was built for.

## The question to settle

Should a sub-session open a `SessionMemory`, render an opening recall block, and attach the mid-turn re-query the other doors attach? Or is the latency argument still the right one for this lane?

Either answer is fine; what is missing is the measurement behind it. A sub-session opens its own store and its own tool registry already, so the "it must stay cheap" claim is about recall latency alone and can be measured.

## Where to look

- `crates/stella-cli/src/subsession.rs` — `run_worker` builds the engine and calls `run_turn_with_sender`. Nothing recalls and nothing attaches a re-query.
- `crates/stella-cli/src/fleet_cmd/steering.rs` — the pattern to copy if the answer is yes. `WorkerSteering::open` renders the opening block and keeps the memory handle for the turn; the caller hands `WorkerSteering::memory()` to `crate::memory::requery_for_turn` and attaches the result with `Engine::with_requery`.
- `crates/stella-cli/src/agent/turn.rs` and `crates/stella-cli/src/command_deck/lead_turn.rs` — the two doors that already do it.

## Constraints a fresh reader needs

- The adapter holds a clone of the lane's event sender, so it must be dropped before the lane's sender is, or the renderer waits on a channel that never closes.
- The re-query may only append to the volatile tail. The byte-stable prefix must not move (AGENTS.md invariant 7).
- `crates/stella-cli/src/subsession.rs` sits close to the 1500-line ceiling `scripts/check-file-size.sh` enforces, so new logic goes in a sibling submodule under `subsession/`.

## Definition of done

- [ ] The decision is recorded in `subsession.rs`'s module header with the latency number that supports it.
[...]

Issue #6121, opened 2026-09-05
Served turns never re-ask for context mid-turn


macanderson:
`stella-serve` assembles no re-query source, so a served turn picks its context from the opening prompt and never looks again. The CLI doors all attach one now — `agent::run_turn`, the deck's lead turn, and (as of the fleet fix) a fleet worker — through `crate::memory::requery_for_turn`. The serve stack attaches nothing.

## Where it stands today

- The engine port is `stella_core::ports::SteeringRequery`, attached with `Engine::with_requery`. `crates/stella-core/src/driver/step_boundary.rs`'s `consult_hosts` builds the `TurnSignal` and asks the port at every step boundary.
- `crates/stella-parity/src/lib.rs`, capability `turn.steering_requery`, carries the `api` posture `Deferred` and names this issue in its `waiting_on`.
- The CLI implementor is `crates/stella-cli/src/memory/steering.rs`'s `SessionRequery`. It owns the hysteresis (a fingerprint over touched paths, error classes and the record generation) and both dedup granularities. None of that is serve's to copy.

## Why it is not a small change

`stella-serve` holds no ambient authority: it opens no store, reads no `.stella/`, and remotes every model and tool call back to the host. A re-query is a recall fan-out with provider spend behind it, so it has to travel the same way — a host-side request over the reverse RPC, shaped like the other remoted ports in `crates/stella-serve/src/remote.rs`. That means a new request/response pair in the serve wire contract, which is generated and gate-checked (`make wire-schema`, `.github/workflows/wire-schema.yml`).

## Constraints a fresh reader needs

- The re-query may only append to the volatile tail. The byte-stable prefix must not move (AGENTS.md invariant 7).
- The host, not the server, decides what a re-query costs and whether it is allowed at all. Serve must be handed the source, the way its other ports are.
- Whatever the host returns is injected verbatim by the engine, so the dedup that stops a byte-identical block being re-injected has to live somewhere. Decide whether that is host-side (mirroring `SessionRequery`) or server-side.
- The parity row's `api` posture moves from `Deferred` to `Shipped` with a named witness only when a served turn really re-queries. The witness must live in one of the files `cli_sources()`/`api_sources()` in `crates/stella-parity/src/lib.rs` lists, or that list grows in the same PR.

## Definition of done

- [x] A served turn re-queries the steering plane at step boundaries through a host-supplied source.
- [x] The request and response are in the generated serve wire schema, and `make wire-schema` is green.
[...]

PR #6124, opened 2026-09-05
feat(stella-cli): let a fleet worker re-ask for context mid-turn

### LINKED ISSUE #6059: Fleet workers and served turns never re-ask for context mid-turn
Mid-turn context re-query works on the main CLI door and nowhere else. A fleet
worker and a served turn both pick their context from the opening prompt and
never look again, so a long turn that drifts onto new files keeps working from
the skills and records the first sentence asked for.

## What is there today

`stella-core` has the port and the boundary that calls it:

- `crates/stella-core/src/ports.rs` — the `SteeringRequery` trait.
- `crates/stella-core/src/driver/step_boundary.rs` — `consult_hosts` builds a
  `TurnSignal` from the turn's own evidence and asks the port at each step
  boundary.
- `crates/stella-cli/src/memory/steering.rs` — `SessionRequery`, the one
  implementor. It owns the hysteresis (a fingerprint over touched paths, error
  classes and the record generation) and the dedup.

Two doors attach it: `crates/stella-cli/src/agent/turn.rs` and
`crates/stella-cli/src/command_deck/lead_turn.rs`, both through
`crate::memory::requery_for_turn`.

## The gap

**Fleet workers.** `crates/stella-cli/src/fleet_cmd.rs` opens a `SessionMemory`
to render the opening recall block and then drops it before the turn starts —
see the function around `memory.recall_block_reported(prompt, &[])`, whose own
comment says the memory "is dropped here". The engine is then built and driven
(`engine.run_turn_with_sender(&mut messages, &mut budget, &worker)`) with no
`with_requery` call anywhere in the file. A fleet attempt is exactly the long
turn this feature was built for, and it is the one that never gets it.

**Served turns.** `stella-serve` assembles no requery source at all. The parity
ledger already says so in prose but cites no issue:
`crates/stella-parity/src/lib.rs`, capability `turn.steering_requery`, whose
`api` posture is `Deferred { waiting_on: "a serve-side SteeringRequery
implementor..." }`.

**Sub-sessions.** `crates/stella-cli/src/subsession.rs` drives a turn the same
way and attaches nothing either. Worth deciding on in the same change, even if
the answer is "a sub-session is short enough not to need it".

## How to check it now

```
rg -n 'with_requery' crates/stella-cli/src crates/stella-serve/src
```

Only `agent/turn.rs` and `command_deck/lead_turn.rs` come back.

## Constraints a fresh reader needs

- The re-query may only append to the volatile tail. The byte-stable prefix
  must not move (AGENTS.md invariant 7), and `inject_recall_block` is
  append-never-rewrite.
- `SessionRequery::new` seeds its dedup from the `RECALL_MARKER` blocks already
  in history plus the opening block's handles, so whatever wires it must pass
  both. `requery_for_turn` is the one call that cannot forget the telemetry
  sender (#3366) — use it rather than assembling the two steps by hand.
- `fleet_cmd.rs` currently drops the memory on purpose: the comment on the
  reflection helper says a fleet attempt does not keep its `SessionMemory`
  alive past the turn. Keeping it alive for the turn's own length is the change
  being asked for; keeping it past the turn is not.
- `crates/stella-serve` must not gain ambient authority. Its requery source
  should be handed in by the host, the way its other ports are.
- The parity row's `api` posture moves from `Deferred` to `Shipped` with a
  named witness only when a served turn really re-queries.

## Definition of done

- [x] A fleet worker turn re-queries the steering plane at step boundaries,
      through `requery_for_turn`.
- [x] A witness test shows a fleet attempt picking up a skill or record its
      opening prompt could not have selected, and fails on the old code.
- [x] `stella-serve` either wires a requery source or the parity row's
      `waiting_on` prose cites this issue number. The row cites #6121 rather
      than this issue: this one closes on merge, and a `waiting_on` pointing
      at a closed issue tells the next reader nothing. #6121 owns the
      serve-side implementor and the wire-contract change it needs.
- [x] The sub-session path has a stated answer, wired or declared. Declared:
      a sub-session is handed no opening context at all, so it has nothing to
      re-ask about. `crates/stella-cli/src/subsession.rs`'s module header says
      so and points at #6122.
- [x] A test pins that the byte-stable prefix is unchanged across a re-query on
      the newly wired door.
- [x] `crates/stella-parity/src/lib.rs` is updated in the same PR for every
      surface whose posture changed. No posture moved — the CLI row was
      already `Shipped` and the API row is still `Deferred` — so the row's
      `mechanism` gained the fleet lane and its witness, held by the new test
      `the_requery_row_cites_the_fleet_worker_witness`.

macanderson:
## What and why

Mid-turn context re-query worked on `stella run` and the deck and nowhere else. A fleet worker picked its context from the opening prompt and never looked again, so a long attempt that drifted onto new files kept working from the skills and records the first sentence asked for. A fleet attempt is the longest turn nobody watches, which makes it the one that needed this most.

**Fleet worker — wired.**

- `crates/stella-cli/src/fleet_cmd/steering.rs` is new. `fleet_cmd.rs` sat at exactly the 1500-line ceiling `scripts/check-file-size.sh` enforces, so the worker's steering seam moved to a sibling submodule rather than growing the parent (AGENTS.md § God files). It is now 1478 lines.
- `WorkerSteering::open` is the old `worker_recall_block` with one change: it **keeps** the `SessionMemory` for the length of the turn instead of dropping it, and returns `(WorkerSteering, OpeningRecall)`. It takes `&AuthorityPolicy` rather than `&Config`, which is all it ever read and what makes it testable.
- `run_task` builds the adapter with `crate::memory::requery_for_turn(worker_steering.memory(), &messages, worker_event_sender(&tx), recall.produced)` and attaches it with `Engine::with_requery` **above** the raw/wrapped split, so a `--pipeline` round re-queries too — the wrapper drives the same engine.
- The adapter holds a clone of the lane's sender, so `drop(requery)` and `drop(worker_steering)` sit ahead of `drop(tx)`. Without that the renderer awaited two lines later waits on a channel that never closes. Dropping the memory there also keeps today's rule that a fleet attempt does not carry its `SessionMemory` into the reflection: that handle is rooted at the attempt's own tree, and `mine_attempt_lesson` opens its own at the invocation root.

**Serve — declared, not wired.** A serve-side implementor means a host-side recall request over the reverse RPC plus a change to the generated serve wire contract; that is its own PR. `crates/stella-parity/src/lib.rs`'s `turn.steering_requery` row keeps its `api` posture at `Deferred` and its `waiting_on` now cites #6121, which owns that work. One issue reference was removed from the same row's `engine_home` so the prose ratchet nets to zero.

**Sub-session — declared.** `subsession.rs` skips recall entirely, so a sub-session worker was handed no opening context and has nothing to ask the plane again about. The module header says so and points at #6122, which weighs turning recall on for that lane with the latency number the v1 scope claimed but never measured.

## Witness mechanism

`a_fleet_worker_requeries_the_skill_its_opening_prompt_could_not`, in `crates/stella-cli/src/fleet_cmd/steering.rs`.

A temp workspace declares one domain over `crates/stella-model`, one real file under it to anchor against, and one skill tagged with that domain whose wording shares nothing with the prompt. The A/B recall control is pinned off in the fixture's `.stella/settings.json`, or the shipped rate would fail the test one run in ten for a reason it is not about.

- Phase 1: the opening block `WorkerSteering::open` injects does not name the skill — the prompt anchors nowhere near its domain. This is what makes phase 2 mean something; an empty first block would satisfy the drift check on its own.
- Phase 2: a drifted `TurnSignal` naming a path in that domain returns a block that does name the skill.

**It fails on the old code by construction**: there is no `WorkerSteering` to build, and the fleet path dropped its `SessionMemory` the moment the opening block was rendered, so nothing outlived the first sentence for `requery_for_turn` to take. The test does not compile against `main`.

The control arm `an_undrifted_fleet_worker_asks_for_nothing` pins the other direction: an unchanged signal never moves the fingerprint, so holding the handle open costs an undrifted attempt nothing.

**Byte-stable prefix** (invariant 7): the witness pins `messages[0].content` byte-identical after the opening injection and after appending the re-query block the way `consult_hosts` does. `inject_recall_block` slots the opening block before the last user message and the engine pushes the mid-turn block onto the tail, so neither touches the prefix.

**Parity citation is checked, not asserted.** The row's `mechanism` names the fleet witness, and `the_requery_row_cites_the_fleet_worker_witness` fails if that name stops existing — the same shape `the_whistle_row_cites_the_room_broadcast_witnesses` uses for a row with more than one witness. `cli_sources()` grows by the new file so the sweep can see it.

## Exemplar followed

`crates/stella-cli/src/agent/turn.rs` and `crates/stella-cli/src/command_deck/lead_turn.rs` — the two doors that already attach the port. The wiring here is the same three steps in the same order, through the same `requery_for_turn` seam.

## Design pointer

The maintainer pointer named a `SteeringPlane` re-query and `NoContextRecall` ports. Neither exists in the tree: `rg -n 'NoContextRecall'` returns nothing, and `SteeringPlane` is the pack-to-budget trait in `stella_core::steering`, not the re-query seam. The shipped seam is `stella_core::ports::SteeringRequery` reached through `crate::memory::requery_for_turn`, which is what the issue body itself prescribes, so this PR follows the issue body. Posted on the issue before starting.

## Tests deleted

None.

## Residue filed

- #6121 — a serve-side `SteeringRequery` implementor, with the wire-contract work it implies.
- #6122 — deck sub-sessions get no context at all, and nobody measured whether that is still right.

Closes #6059

## Summary by Sourcery

Allow fleet workers to refresh their steering context when a turn drifts beyond its opening prompt.

New Features:
- Enable fleet workers to re-query steering context during a turn when their work drifts into new files or domains.

Bug Fixes:
- Keep fleet worker session memory alive through the attempt so mid-turn context retrieval can function without affecting reflection behavior.

Enhancements:
- Extract fleet worker steering into a dedicated module and attach re-query support across both standard and pipeline engine execution.
- Document serve-side and sub-session re-query support as deferred follow-up work.

Documentation:
- Update parity metadata and source-witness checks to document and verify the fleet worker re-query capability.

Tests:
- Add fleet steering coverage for drift-triggered skill retrieval, undrifted attempts, and preservation of the stable prompt prefix.


macanderson:
**Autonomous PR routine — CI fix pushed, and one gate still open.**

### Fixed and pushed (`5ce2e0d`)

`the macOS-only dependency table still compiles` was red, and not for a macOS
reason — the branch did not compile at all with `--all-targets`:

```
error[E0425]: cannot find function `worker_recall_block` in this scope
   --> crates/stella-cli/src/fleet_cmd/tests.rs:695:19
```

The first commit moved `worker_recall_block` into
`fleet_cmd/steering.rs` as `WorkerSteering::open` and updated the production
caller, but `fleet_cmd/tests.rs` kept calling the old name. The macOS job
happens to run `cargo check -p stella-cli --all-targets` and got there first;
`fmt + clippy + test` would have hit the same error on its own test build.

The call is repointed at `steering::WorkerSteering::open`, taking
`&cfg.authority` rather than `&cfg` to match the new signature. The doc comment
above the test also named the removed function and claimed the feature was
absent on base *because that function did not exist* — what is absent on base is
the opening block itself, which is what the assertions actually check, so it now
says that.

Verified locally rather than asserted:

- `cargo check -p stella-cli --all-targets --locked` → exit 0 (the same command
  the macOS job runs).
- `cargo test -p stella-cli a_fleet_worker_receives_skills_records_and_todays_date`
  → `1 passed`.
- `cargo fmt --check` on the edited file → exit 0.

CI on the new head is the real evidence; the above is what made it worth pushing.

### Still open: `dod / dod`

That check is red because all six of #6059's DoD boxes are unticked. Reading the
PR against them, five look genuinely met — the `requery_for_turn` wiring, the
drifted-worker witness, the sub-session declaration (#6122), the byte-stable
prefix assertion, and the `stella-parity` update. The sixth is a judgement call
worth flagging rather than deciding here:

> `stella-serve` either wires a requery source or the parity row's `waiting_on`
> prose cites **this issue number**.

The row now cites **#6121** instead. That is arguably the better pointer — #6121
owns the actual serve-side work and survives this issue closing — but it is not
what the box asks for, so it is your call whether to reword the box, repoint the
row at #6059, or accept the substitution.

Ticking the boxes is deliberately left to you: SCR-003 asks for every item
satisfied **and verified**, and until this push the branch did not compile, so
nothing here had been verified by a green run yet. Once you tick them, note that
`dod-check` reacts to pull-request events rather than issue edits — `dod-recheck.yml`
should pick up an issue edit, but a push to the branch is the reliable nudge.

---

Issue #6125, opened 2026-09-05
Add a golden frame for the SEATS pane's focused editor and open model picker


macanderson:
## Problem

#6086 turned the SEATS pane (`crates/stella-tui/src/views/seats.rs`) into a
modal editor: `e` focuses it, `⏎` on a seat opens the AGENTS pane's model
picker, `x` clears a seat, `s`/`S` save. The only golden frame for this pane,
`crates/stella-tui/tests/snapshots/deck/pane_settings_seats.txt`, covers the
**unfocused** browse state. Nobody has a golden of the pane while focused, or
of the model picker's bordered popup open over it.

The session that shipped #6086 could not add one: it had no local Rust
toolchain run available to bless a new snapshot (`BLESS=1 cargo test -p
stella-tui --test deck_render_snapshots`), so it hand-derived the existing
golden's small diff instead of adding a new file. A genuinely new golden
needs the real renderer running once to produce its first committed copy.

## The fix

Add a test to `crates/stella-tui/tests/deck_render_snapshots/settings_seats.rs`
(a sibling test function, same file) that:

1. Builds the same `fixture_seats()` state.
2. Calls `crate::views::seats::focus_panel` (or sets `ui.engine.focused = true`
   plus `ui.settings_pane = SettingsPane::Seats` directly) and moves
   `ui.engine.row` onto the plugin seat row (not the leading resolved-role
   row).
3. Opens the picker (`ui.engine.picker = Some(ModelPicker::default())`, or by
   driving `handle_deck_key` with `⏎`).
4. Renders and asserts against a new golden, e.g. `pane_settings_seats_picker`.

Bless it locally with:

```
BLESS=1 cargo test -p stella-tui --test deck_render_snapshots
```

then **read the diff** before committing — a golden blessed without reading it
is a changelog, not a test (`AGENTS.md` § Testing approach).

## Definition of done

- [ ] A new golden frame test covers the SEATS pane while focused (no picker
      open) and/or with the model picker open.
- [ ] The snapshot file under `crates/stella-tui/tests/snapshots/deck/` is
      committed and its diff has been read by a human or agent, not just
      generated.
- [ ] `make gate` (or at least `cargo test -p stella-tui --test
      deck_render_snapshots`) is green in CI.

## Constraints / context

- `crates/stella-tui/src/views/seats.rs` is the pane; `EngineOverlay` (shared
  with the AGENTS pane, `crates/stella-tui/src/views/engine_panel.rs`) holds
[...]

PR #6126, opened 2026-09-05
feat(stella-tui,stella-cli): make the SEATS pane an editor, not a display

### LINKED ISSUE #6086: The seats pane shows a plugin's model but cannot change it

macanderson:
## What & why

The SEATS pane (`crates/stella-tui/src/views/seats.rs`) showed which model
each plugin-declared seat runs on, but `e` on it did nothing — the header
even said `read-only`. The only way to assign a seat a model was to
hand-edit `.stella/settings.json`.

Digging in turned up a second, deeper bug the issue asked me to check for:
even if the pane had an editor, the write path it would use,
`settings_from_state` (`crates/stella-cli/src/engine_config.rs`), never read
`state.seats` at all — a seat assignment made anywhere would have been
silently dropped on save.

This fixes both, by reusing the AGENTS pane's modal editor rather than
building a second copy of it:

- **`crates/stella-cli/src/engine_config.rs`** — `settings_from_state` now
  folds `state.seats` into `agent_engine_config.seat_models`. A seat with a
  model writes its `<plugin-id>/<role>` key; a cleared seat writes no key at
  all (never the current default — an unassigned seat and one pinned to
  today's default are different facts).
- **`crates/stella-tui/src/views/seats.rs`** — the pane is modal while
  focused, over the *same* `EngineOverlay` the AGENTS pane already owns:
  `e` focuses, `⏎` on a plugin's seat opens the AGENTS pane's own model
  picker (filtered by `allowed_models`), `x` clears the seat, `s`/`S` save.
  The leading, session-resolved row is read-only (it isn't a `seat_models`
  entry) and says so on `⏎`/`x` rather than doing nothing silently.
- **`crates/stella-tui/src/views/engine_panel/keys.rs`** — `handle_engine_key`
  hands off to the SEATS pane's own key map first when that pane is showing,
  and `save`/`refresh` become `pub(crate)` so the SEATS pane calls the exact
  same functions the AGENTS pane's `s`/`S`/`r` do.
- **`crates/stella-tui/src/views/settings.rs`** — `e` on the SEATS pane now
  focuses it (was a claimed no-op), and rendering passes the pane the whole
  `ui` instead of a bare row slice, since it now draws focus/selection/picker
  state too.

`crates/stella-tui/src/deck_ui.rs` — a god file closed to growth
(AGENTS.md § "God files") — needed **zero** new lines: it already routes
every SETTINGS-tab keypress through `handle_engine_key`/`settings::handle_key`
regardless of which pane is focused, so reusing `EngineOverlay` wholesale
meant the existing dispatch already covers the new pane.

**Exemplar:** `crates/stella-tui/src/views/engine_panel/keys.rs` (the AGENTS
pane's own key map) — matched call for call rather than reinvented.

## The witness

- `crates/stella-cli/src/engine_config.rs`:
  `a_seat_assignment_survives_the_save_round_trip` assigns a seat, saves,
  reloads the snapshot from the saved settings, and asserts the assignment
  reads back on the same key — the round trip the issue's DoD asks for.
  `clearing_a_seat_removes_its_key_rather_than_writing_the_default` proves
  the clear path drops the key rather than writing a substitute model. Both
  fail on `main` today (`settings_from_state` never reads `state.seats`).
- `crates/stella-tui/src/views/seats.rs`:
  `e_enter_assigns_a_seat_and_s_saves_it` drives the whole thing through
  `handle_deck_key` the way a real keypress would (`e`, `⏎` on the
  resolved-role row — explains why nothing happened — `↓`, `⏎`/type/`⏎` on
  the seat, `s`) and asserts the model landed on the seat and the save
  carries the whole edited snapshot. `x_clears_a_seat_to_default` and
  `esc_returns_focus_to_the_tab` cover the other two verbs. All fail on
  `main` — there was no key handler to drive.

## The gate

- [x] `fmt --check` (ran locally)
- [ ] Workspace clippy, `-D warnings` — CI (see note below on local limits)
- [ ] Workspace test suite — CI, same reason
- [x] Docs updated: `crates/stella-tui/README.md`'s SEATS row, and every
      touched module's doc comments
- [ ] CLA — bot will prompt
- [x] `Closes #6086` above and as a commit trailer

**This session has no local Rust compiler run available**, so one golden
frame was hand-updated rather than blessed:
`crates/stella-tui/tests/snapshots/deck/pane_settings_seats.txt`. I derived
the new header text and column layout directly from `seats.rs`'s own width
math (`column_widths`, `GAP`, the 2-cell selection marker) and the fixture's
exact strings, character by character, rather than regenerating it. I'm
reasonably confident in the arithmetic, but this is the one place in the
diff a fresh pair of eyes — or CI's own test job — should double-check
against a real run before merging. If the test job disagrees, trust the
run over my derivation.

## Nothing left behind

- [x] Filed: #6125 — a golden frame for the SEATS pane while **focused**,
      and with the model picker **open**, does not exist yet (only the
      unfocused browse state does, and that's the one I could derive by
      hand). A session with a working local Rust toolchain should add and
      bless it.
- [x] Otherwise: everything I noticed is fixed in this PR.

## Ground-rule check

- [x] No new dependencies

- The SEATS pane reuses `EngineOverlay`'s `focused`/`row`/`picker`/`status`/
  `busy` fields wholesale rather than carrying a second copy — a seat and an
  agent-model edit are both, structurally, "the working copy of
  `EngineConfigState` has an unsaved edit", and `ui.settings_pane` is what
  says which pane's rows the shared `row`/`picker` currently mean. This is a
  deliberate reuse-over-duplication call, not an oversight — see the module
  docs on `EngineOverlay` and on `seats.rs` for the reasoning, and the "Nothing
  left behind" note above for the one gap it leaves (no golden of the shared
  overlay's SEATS face).
- The picker opened from a seat is keyed on the selected **seat**
  (`seat_at`/`seat_at_mut`, combined-row-index arithmetic against
  `state.roles.len()`), not on an `EngineRole` — the AGENTS pane's own
  `open_picker`/`agent_mut` couldn't be reused as-is for that reason, so
  `seats.rs` has its own small picker-key and nav-key handlers that call into
  the AGENTS pane's `picker_matches`/`picker_candidates`/`save`/`refresh`
  rather than duplicating them.

## Summary by Sourcery

Turn the SEATS settings pane into a shared, persistent editor for plugin seat model assignments.

New Features:
- Make the SETTINGS SEATS pane an interactive editor for assigning and clearing plugin seat models through the existing model picker.
- Reuse the shared engine editor overlay and save flow for SEATS interactions, including focus, navigation, filtering, and persistence controls.

Bug Fixes:
- Persist seat assignments and removals when engine configuration is saved, preserving the distinction between unassigned seats and explicitly pinned models.
- Provide feedback when editing the session-resolved role, which is read-only in the SEATS pane.

Enhancements:
- Update SEATS rendering, navigation, status, legends, scrolling, and selection to support modal editing while retaining the resolved-role row and plugin-defined seat list.

Documentation:
- Update the TUI documentation to describe SEATS editing and its shared save behavior.

Tests:
- Add save round-trip coverage for seat assignments and clearing.
- Add TUI interaction coverage for focusing SEATS, assigning a model, clearing a seat, saving, and exiting focus.
- Update SEATS rendering and layout tests for the editor state and revised pane layout.


macanderson:
**Automated PR sweep — `dod / dod` is the only red check, and #6086's four items all verify against this head (`533495c`).**

Checked against the diff rather than the description:

**"Assigning a model to a seat writes `agent_engine_config.seat_models`"** ✅ — `seat_models` is written through `crates/stella-cli/src/engine_config.rs` and reaches the settings layer (`settings/engine.rs`, `settings/toml_config.rs`, `settings/completeness.rs`) and the TUI envelope (`stella-tui/src/envelope/engine_config.rs`). The key survives serialization with its slash intact, pinned by `a_seat_key_survives_a_toml_round_trip_with_its_slash_intact` (`settings/tests/toml.rs:1219`).

**"Clearing a seat removes its key rather than writing the default model"** ✅ — `clearing_a_seat_removes_its_key_rather_than_writing_the_default` (`engine_config.rs:1386`). The test name restates the DoD line word for word, and `x_clears_a_seat_to_default` (`views/seats.rs:1069`) covers the key binding that reaches it.

**"A test proves the round trip: assign, save, reload the snapshot"** ✅ — `a_seat_assignment_survives_the_save_round_trip` (`engine_config.rs:1323`), with `e_enter_assigns_a_seat_and_s_saves_it` (`views/seats.rs:1016`) driving the pane's own assign-then-save path.

**"The pane header no longer says `read-only`"** ✅ — verified by diffing the two trees. On `main`, `views/seats.rs:146` builds the header as `format!(" · {} · read-only", rows.len())` and two tests assert on that string (`:339`, `:471`). On this head the format string and both assertions are gone. The two surviving `read-only` mentions in that file are prose about the *leading resolved-role row* — a different and still-accurate statement, not the pane header.

So the gate is red on unticked boxes, not on missing work. Ticking #6086's four items re-runs the check by itself via `dod-recheck.yml`; no empty commit needed.

Everything else on this head is green: `cargo deny`, MSRV, `docs guards`, `file size ratchet`, `guard self-tests`, `post-merge canary self-tests`, `gate steps no other workflow runs`, `main is not known-broken`, `no edit is undone inside the branch`, `the macOS-only dependency table still compiles`, `harbor_adapter + analyzer pytest`, `duplicate claims`, `dependency-review`, `PR carries a diff`, and three of four CodeQL analyses. `fmt + clippy + test` and `Analyze (rust)` were still running. No merge conflict, no `hold` label.

One caveat on all of the above: this is a source-level check. `fmt + clippy + test` is still the authority for whether these tests actually pass.

---

macanderson:
**Automated PR sweep — one gap, and it is the half of "belt and braces" that fails silently.**

CI is clean or pending on `a3cc3bc` (nothing red), no merge conflict, no `hold` label, and #6086's five DoD boxes are all ticked with `dod / dod` green. The only finding is the issue link.

**GitHub has registered no closing link from this PR to #6086.** `GET /issues/6086` returns `closed_by_pull_requests: {"total_count": 0, "references": []}`, and #6086 does not appear in `is:issue is:open linked:pr` results for this repo. Sourcery's guide lists it under *"Possibly linked issues"* — its own inference, not a GitHub link.

The cause is one line. `#6086` appears exactly once in the PR description, at line 72:

```
- [x] `Closes #6086` above and as a commit trailer
```

It is inside a code span, so GitHub's parser never sees a keyword, and there is no plain `Closes #6086` line "above" for it to be referring to — the checklist item asserts something the body does not contain.

**This is not currently load-bearing, and that is worth stating plainly rather than leaving as an implied emergency.** Commit `533495c` really does carry a bare `Closes #6086` trailer, and AGENTS.md § "Closing the issue on merge" says squash — the default here — composes the commit body from `COMMIT_MESSAGES`. So on merge the trailer closes #6086 through commit-message parsing. Rebase replays the commit verbatim and a merge commit still pushes it to `main`; all three paths preserve it. The issue will close.

What is lost is the redundancy the same section calls for: *"either one alone is a silent single point of failure, and the failure mode is invisible until someone audits the backlog."* This sweep is that audit, and today the PR is running on one of the two.

**Remedy, if you want both halves:** add a bare `Closes #6086` line to the description (outside backticks). Nothing else changes — the trailer is already correct, and `dod / dod` is already green, so this costs no CI cycle beyond the re-run the body edit triggers.

**Not pushed, and the reason is scope rather than caution.** The fix is one line, but `update_pull_request` replaces the whole body, and re-transmitting 7,462 bytes of hand-written prose — including the derivation notes and the box-drawing characters in the pane description — to insert one line risks corrupting a description more valuable than the redundancy it would buy. That is the author's line to add.

One thing this sweep did **not** verify: whether `pane_settings_seats.txt` is correct. The description flags it as hand-derived without a local toolchain, and `fmt + clippy + test` was still running when this was written — that job is the authority on it, not any source-level reading.

---

Issue #6134, opened 2026-09-05
Self-driving loop makes git worktrees without the fleet worktree lock


macanderson:
Follow-on from #6093, which keyed the *input* side of the appraisal ledgers by
artifact. The output side is still shaped for skills alone.

## The problem

`crates/stella-cli/src/memory/appraisals.rs` now folds trials by
`(ArtifactKind, id)`, so a memory record or a mined rule can be appraised as
soon as #6096 gives it a producer. What the sweep then *writes* has no such
key:

- `APPRAISALS_FILE` (`skill_appraisals.jsonl`) stores `SkillAppraisal`, whose
  identity field is `skill: String`. Two artifacts of different kinds sharing
  an id would collapse into one row.
- `latest_verdicts` returns `HashMap<String, EvalEvidence>` keyed by that same
  bare string, and the skill creation gate reads it.
- `consecutive_negative_appraisals` counts back over the file filtering on
  `a.skill == skill`, so the hysteresis run would mix kinds.
- `record_demotion` files the event under the `skill:<name>` lineage prefix
  (`SKILL_LINEAGE_PREFIX`), and `demoted_skills` folds only that prefix. A
  retired memory record and a retracted rule each need their own lineage, and
  each already has its own retirement door elsewhere in the tree.

The vocabulary is the other half. `SkillTrial`, `SkillAppraisal`,
`SkillVerdict`, `WITH_SKILL_ARM` and `WITHOUT_SKILL_ARM`
(`crates/stella-learn/src/skills/appraisal.rs`) are the shared engine for three
kinds now, and every one of them is named for one. AGENTS.md's "name things for
what they are, not what they were" applies.

## The fix

Two steps, and the first is worth doing without the second:

1. Key the appraisal ledger and the demotion lineage by `(kind, id)` the way
   the trial ledger already is. Old rows have no kind and were all skills, so
   the same default and alias trick #6093 used on `ArtifactTrial` applies —
   read `stella_learn::ledger::ArtifactTrial`'s field docs before writing a
   second scheme.
2. Rename the engine's types to say artifact rather than skill, in one
   mechanical pass. This is a wide rename across `stella-learn` and
   `stella-cli`; regenerate it on a fresh `main` rather than merging a stale
   branch.

Order matters only in that the rename touches every line the re-key does, so
doing them in one PR is cheaper than two — but the rename is not required for
a memory sweep to be correct, and a reviewer should be able to read the re-key
on its own.

## Constraints

[...]

PR #6135, opened 2026-09-05
fix(stella-fleet): hold one git worktree command at a time per repo

### LINKED ISSUE #6127: Concurrent `git worktree add` races in fleet dispatch — test-session-isolation.sh went red on main
## What happened

`ci` run [9255](https://github.com/macanderson/stella/actions/runs/33993245773) on `main`
(commit `67be9da`, the merge of #6094) failed its `fmt + clippy + test` job on
`./scripts/test-session-isolation.sh`, case **"two workers, no isolation named"**:

```
  ✗ expected 2 worker trees, got 1. The run said:
      ✓ t1 — answer in one short sentence ($0.0000, 0 commits)
          ↳ fleet/t1-14591e8e7e85d4c7 @ …/.stella/worktrees/t1-14591e8e7e85d4c7
      ✗ t2 — dispatch failed: `git worktree add …/t2-14591f8e7e85d67a
            -b fleet/t2-14591f8e7e85d67a e9fd301cec5d…` failed (exit Some(128)):
        Preparing worktree (new branch 'fleet/t2-14591f8e7e85d67a')
        fatal: failed to read .git/worktrees/t1-14591e8e7e85d4c7/commondir: Success

  4 passed, 1 failed
```

The other two cases (`isolated`, `shared_tree`) passed. The whole `ci` job is
otherwise green up to that step — `self-driving-test` passed 61 checks
immediately before it, and `test-impacted-crates.sh` passed 25 immediately after.

## Why this is a race, not a logic error

`errno` is reported as **`Success`**. Git only prints that when the `read` of
`commondir` returned short with no error set — i.e. it read a file another
process was still writing. `git worktree add` for **t2** tripped over the
half-written metadata `git worktree add` for **t1** was laying down.

The two slugs, `t1-14591e8e7e85d4c7` and `t2-14591f8e7e85d67a`, differ only in
the middle of their hash suffix, which is consistent with both dispatches
starting within the same instant.

## Where the missing serialization is

`WorktreeManager::create` (`crates/stella-fleet/src/git.rs`) shells straight out
to `git worktree add` with nothing holding concurrent callers apart:

```rust
pub async fn create(&self, task_id: &str, base_ref: &str) -> Result<Worktree, WorktreeError> {
    …
    let out = self.git.run(&self.repo_root,
        &["worktree", "add", dir_str, "-b", &branch, base_ref]).await?;
```

`Fleet` (`crates/stella-fleet/src/fleet.rs`) does hold mutexes — its own comment
says the three are behind `Mutex`es "so concurrent wave dispatch serializes their
fast paths" — but they cover `ledger`, `budget` and `controls`. Worktree creation
is not among them, so a wave of width N issues N concurrent `git worktree add`
calls into one repository. Git does not lock `.git/worktrees/` for this, so the
interleaving above is reachable whenever two dispatches land close enough
together.

This is a defect in the dispatch path, not in the test: the test drove the
production code and the production code lost a worker. A user running
`stella fleet` with `--max-concurrency 2` can lose a task the same way, and the
failure surfaces as a task-level "dispatch failed", not as an obvious race.

## How to reproduce

`./scripts/test-session-isolation.sh` (or `make session-isolation-test`) — it is
hermetic (throwaway repo, stub model on loopback, `$STELLA_HOME` under the temp
tree). It is timing-dependent, so expect it to pass most runs; a loop of ~20
iterations, or a loaded machine, is what surfaces it. Run 9255's log above is one
observed instance.

## Constraints for whoever picks this up

- `crates/stella-fleet/src/git.rs` is the natural home for the lock — an async
  mutex held across the `git worktree add` for the repository the manager is
  rooted at. `WorktreeManager` already owns `repo_root`, so the scope is
  per-manager rather than global.
- Serializing only `create` may not be enough on its own: `git worktree remove`
  and the `gc.rs` prune walk the same `.git/worktrees/` directory. Check whether
  they need the same lock before concluding the fix is one call site.
- Do not serialize the **workers** — only the tree cutting. The whole point of
  the fan-out is that the turns run concurrently; only the metadata write needs
  to be exclusive.
- A witness test for a race is the hard part of this issue, and it is worth
  deciding deliberately rather than settling for a loop that usually fails.
  One option that does not depend on timing: drive `WorktreeManager::create`
  through a `GitCli` fixture that records call overlap and assert no two
  `worktree add` invocations are ever in flight at once. That tests the
  serialization property directly rather than trying to lose the race on demand.
  `crates/stella-fleet/src/fleet/tests.rs` already has a recording `GitCli`
  stub to build on.

## Definition of done

- [x] Concurrent fleet dispatch issues no two overlapping `git worktree add`
      calls against one repository.
- [x] A witness test pins that property and fails on the current code.
- [x] `git worktree remove` / the `gc.rs` prune are checked against the same
      hazard, and either covered by the same lock or shown not to need it, with
      the reason written at the call site.
- [x] `./scripts/test-session-isolation.sh` passes.
- [x] The lock (and any related change) lands inside `crates/stella-fleet` at a
      call site named above — no growth added to a file already on
      `scripts/file-size-baseline.txt` (none of `stella-fleet`'s files are
      currently grandfathered there, so any new god-file crossing fails the
      gate outright rather than needing a baseline bump; confirm this stays true).
- [x] The lock carries a doc comment stating what it guarantees (no two
      `git worktree add` calls in flight against one repo root) and what it
      deliberately does not do (workers themselves stay concurrent) — matching
      AGENTS.md's "match the neighborhood" doc-comment convention.
- [x] If `crates/stella-fleet/README.md` or any other doc currently implies
      `WorktreeManager::create` is safe for concurrent callers (or is silent on
      it in a way that would now be misleading), it is updated to state the
      guarantee this fix adds.
- [x] No new prose trips `make prose` (`scripts/check-prose.py`) — plain,
      direct wording only.

## Notes

- No `main-red` issue was open when this was found, and `main is not
  known-broken` was passing on every open PR — the canary files only when its
  own job runs and fails, and this failure predates the current tip.
- `main`'s tip has since moved to `2f20dda` (the 0.9.348 version sync); its own
  `ci` run was still in flight when this was filed, so whether the tip
  reproduces is not yet established either way.
- Filed by the autonomous PR routine while triaging main CI. Not fixed in that
  pass: the serialization is a one-line change but the witness for it is a
  design question, and guessing at it would ship exactly the kind of
  usually-passes test this repository does not want.
- **Triage note (2026-09-05):** checked against every open epic in the
  `area:fleet`/fleet-adjacent roster. #5310 ("parallel sessions coordinate —
  leases, worktrees, and branches are visible and bounded") is the closest by
  subject but closed COMPLETED the same day this issue was filed, and its own
  stated outcome was about *visibility* of claims/branches/worktrees across
  sessions, not about serializing a `git worktree add` call against itself —
  a different mechanism even had it stayed open. #2451 ("the fleet worker
  boundary discards AbortKind") is the nearest open `area:fleet` sibling by
  subject but is an ordinary issue, not an epic, and its outcome (abort-kind
  attribution in the ledger) is unrelated. No open epic's stated outcome
  contains this issue's scope, so per this repo's `/backlog-triage` contract
  (see #2778) it stays an orphan rather than being force-parented — recorded
  in #2778.

macanderson:
## What

`git/worktree_lock.rs` is now the one door every `git worktree …` call in
`stella-fleet` goes through. No two of them run against one repository root at
once.

Call sites moved onto it: `WorktreeManager::create`, `WorktreeManager::discard`,
`remove_worktree_and_branch` (which is both `WorktreeManager::remove` and
`Gc`'s sweep, so the third DoD bullet is answered by covering one routine), and
`Gc::sweep`'s `worktree prune` and `worktree list --porcelain`. The guard is
taken at the leaf only and never nested, so there is no reentrancy path to
deadlock on.

## Why

Git keeps its worktree bookkeeping in `.git/worktrees/<name>/` and takes no
lock while it writes there, so two `git worktree add` calls that start together
can read each other's half-written files. `ci` run 9255 caught exactly that: the
second add died on
`failed to read .git/worktrees/<sibling>/commondir: Success` — errno `Success`
is git reporting a short read — and a fleet wave lost a whole task to a
`dispatch failed`. A user running `stella fleet --max-concurrency 2` loses a
task the same way.

## Mechanism

Two locks, because there are two ways to collide.

- **In one process**: a `tokio::sync::Mutex` per repository root, handed out by
  a process-wide map, so a `WorktreeManager` and a `Gc` pointed at one repo take
  the same one. This half is a guarantee. The key is the canonicalized path when
  the path resolves, so two spellings of one repo share a lock.
- **Across processes**: an `O_CREAT|O_EXCL` lock file under
  `<repo_root>/.stella/private/`, bounded on both sides — a waiter gives up
  after 300s and proceeds, and a lock file older than 1800s is broken as the
  residue of a dead process. This half is best effort and proceeds rather than
  failing a dispatch, because losing a task to a slow peer is worse than the
  race being narrowed.

Plus one retry when git failed with a signature another process's bookkeeping
could have caused (`index.lock`, `unable to create`, the `commondir` short read
from run 9255). The first error is what the caller sees if the retry fails too.
A real collision (`'<path>' already exists`) is not on that list, so it fails
once and is reported once.

**Exemplar**: `stella-graph`'s `ManifestLock`
(`crates/stella-graph/src/manifest.rs`) — this workspace's existing answer to
"exclusive without a platform-specific dependency", including the bounded-wait
and stale-break shape. No new dependency.

## Witness mechanism — why it fails on `main` by construction

`concurrent_creates_never_overlap_a_worktree_add`
(`crates/stella-fleet/src/git/worktree_lock/tests.rs`) drives a recording
`GitCli` that counts how many calls are inside `run` at once and keeps the peak.
It sleeps 20ms inside the call. Eight `create` calls are driven concurrently
against one manager with `join_all`.

Without the lock, all eight futures are polled into `run` before any sleep
finishes, so the peak is 8 and the assertion of 1 fails — every time, not most
of the time, because the overlap is made by the fake's `await` rather than by
winning a race. This is what the issue's own "Constraints" section asked for
instead of a loop that usually fails.

Its control, `the_fake_reports_overlap_when_nothing_holds_the_callers_apart`,
runs the same fake and the same eight callers with the lock out of the path and
asserts a peak of 8 — so the witness is measuring a counter that can climb.

Beside it: the lock file is taken and released; a live lock file is honoured
until the wait runs out and then proceeds; a stale one is broken; a transient
failure is retried once; a retry that fails reports the first error; a real
collision is not retried; and one repository root maps to one mutex through two
spellings of its path.

## One correction to the design pointer

The pointer asked for "a unique path derived from the attempt id (never a shared
temp name)". The code already does that and it is not where the defect is:
`worktree_slug` gives every task a `<stem>-<64-bit hash>` directory, and the two
paths in the failing run (`t1-14591e8e7e85d4c7`, `t2-14591f8e7e85d67a`) were
already distinct. The failure was t2 **reading** t1's half-written `commondir`,
not a path collision. No path change lands; the serialization is the whole fix.

## File placement

New logic lands in a sibling submodule (`src/git/worktree_lock.rs` +
`worktree_lock/tests.rs`), not in `git.rs`. No `stella-fleet` file is on
`scripts/file-size-baseline.txt` and this change adds none — `check-file-size`
is green and reports no new crossing.

## Docs

`crates/stella-fleet/README.md` gains the module's row in the file map and an
invariant bullet stating the guarantee (one `git worktree` command at a time per
repo) and what it deliberately leaves alone (the workers). The module doc states
both at the call site.

## Residue filed

- #6134 — `stella-cli`'s self-driving loop shells out to `git worktree
  add`/`remove`/`prune` through its own synchronous helper and does not take
  this lock, so it can still collide with itself or with a fleet run in the same
  clone. Not fixed here: the lock is `async` and `pub(crate)`, and both of those
  are crate-boundary decisions rather than lines of code.

Tests deleted: None

Closes #6127

## Summary by Sourcery

Prevent concurrent `git worktree` commands from interfering with one another within a repository while keeping worker execution concurrent.

Bug Fixes:
- Serialize all `git worktree` operations per repository to prevent concurrent worktree bookkeeping corruption and dispatch failures.
- Retry transient Git worktree failures once while preserving the original error when the retry also fails.

Enhancements:
- Coordinate worktree commands with an in-process per-repository mutex and a best-effort cross-process lock with bounded waits and stale-lock recovery.
- Route worktree creation, removal, discard, pruning, and listing through the shared locking boundary.

Documentation:
- Document the worktree serialization invariant and add the locking module to the fleet file map.

Tests:
- Add deterministic concurrency coverage and tests for lock-file lifecycle, stale-lock handling, retry behavior, collision handling, and repository path canonicalization.

### DIFF
--- a/crates/stella-fleet/README.md
+++ b/crates/stella-fleet/README.md
@@ -112,6 +112,7 @@ is nearest today — split it before it crosses.
 | [`src/ledger.rs`](src/ledger.rs) | `fleet.db`: schema, migrations, and every read/write of runs, tasks, attempts, commits, lineage and spend. |
 | [`src/ledger/lease.rs`](src/ledger/lease.rs) | Dispatch claims (#1136): the check-and-set, expiring lease on one unit of dispatch — `claim_dispatch` / `renew_dispatch` / `release_dispatch` and the reads a human or a dispatcher asks "who is on this?" with. |
 | [`src/git.rs`](src/git.rs) | The `GitCli` port, `SystemGitCli`, and `WorktreeManager` — worktree create/remove/discard/list plus the pathspec-only commit helper. `with_worktrees_root`/`with_branch_prefix` move a caller out of the `.stella/worktrees/` + `fleet/` namespace `gc.rs` reclaims by, and `discard` force-removes where `remove` refuses: both exist for `stella-cli`'s best-of-N candidate substrate (#3892), whose output is uncommitted bytes by construction and whose losers are meant to be thrown away. |
+| [`src/git/worktree_lock.rs`](src/git/worktree_lock.rs) | The one door every `git worktree …` command in this crate goes through. Git takes no lock on `.git/worktrees/`, so two overlapping `worktree add` calls can read each other's half-written bookkeeping; this holds them apart with a per-repo-root mutex in the process and an `O_CREAT|O_EXCL` lock file under `.stella/private/` across processes, and retries once on git's transient signatures. |
 | [`src/gc.rs`](src/gc.rs) | `Gc::sweep` — the conservative reclaim of finished worktrees and `fleet/*` branches behind `stella fleet clean` (#1217). Nothing in flight, dirty, or unmerged goes without `--force`; every kept candidate carries a `KeepReason`. |
 | [`src/monitor.rs`](src/monitor.rs) | The `GhCli` port and `Monitor`: live PR reconciliation, the CI poll loop, the pure `decide` cap state machine, and the `AgentEvent` emit-shape helpers. |
 | [`src/cache_schedule.rs`](src/cache_schedule.rs) | `warmest_first` — the pure, no-I/O ordering heuristic for equal-priority runnable sessions. |
@@ -323,6 +324,16 @@ Two gaps:
   integration ref as `contained_in`, since it has no `Worktree` value to
   read a `base_ref` from. Every other caller leaves `contained_in` at its
   default, which falls back to `Worktree::base_ref`.
+- **One `git worktree` command at a time per repository.** `WorktreeManager::create`,
+  `discard`, `remove_worktree_and_branch` and `Gc::sweep`'s prune and list all run
+  through `git/worktree_lock.rs`'s `run_worktree`, which holds a per-repository lock
+  for the length of the command. Concurrent callers of `create` are therefore safe:
+  a wave of width N issues its adds one at a time. What stays concurrent is the
+  work — the workers themselves take no lock, and only the tree cutting is
+  exclusive. The in-process half is a guarantee; the lock file that extends it to a
+  second `stella` process is best effort and proceeds rather than failing a
+  dispatch when it cannot be taken.
+
 - **`WorktreeManager::commit_paths` has no product caller yet.** It is
   exercised only by this crate's own tests, and it ships anyway because it is
   the only place here that encodes the pathspec discipline — `git add --
--- a/crates/stella-fleet/src/gc.rs
+++ b/crates/stella-fleet/src/gc.rs
@@ -33,7 +33,7 @@ use std::path::{Path, PathBuf};
 
 use crate::git::{
     GitCli, RemoveOutcome, WorktreeError, ensure_ok, is_contained_in, parse_worktree_list,
-    remove_worktree_and_branch,
+    remove_worktree_and_branch, worktree_lock,
 };
 
 /// Milliseconds in a day — the unit [`GcOptions::min_age_days`] is expressed
@@ -283,11 +283,12 @@ impl<G: GitCli> Gc<G> {
         // deleted: without this they are listed forever and every later
         // `git worktree add` reasons about a checkout that is not there.
         if !opts.dry_run {
-            let out = self
-                .git
-                .run(&self.repo_root, &["worktree", "prune"])
-                .await
-                .map_err(WorktreeError::from)?;
+            // Under the worktree lock: prune rewrites the same
+            // `.git/worktrees` directory a concurrent dispatch is adding to.
+            let out =
+                worktree_lock::run_worktree(&self.git, &self.repo_root, &["worktree", "prune"])
+                    .await
+                    .map_err(WorktreeError::from)?;
             ensure_ok(out, "worktree prune").map_err(GcError::from)?;
         }
 
@@ -313,11 +314,15 @@ impl<G: GitCli> Gc<G> {
         base_ref: &str,
         now_ms: u64,
     ) -> Result<Vec<WorktreeVerdict>, GcError> {
-        let out = self
-            .git
-            .run(&self.repo_root, &["worktree", "list", "--porcelain"])
-            .await
-            .map_err(WorktreeError::from)?;
+        // Under the lock as well: the read of `.git/worktrees` is the half
+        // that lost run 9255, not the write.
+        let out = worktree_lock::run_worktree(
+            &self.git,
+            &self.repo_root,
+            &["worktree", "list", "--porcelain"],
+        )
+        .await
+        .map_err(WorktreeError::from)?;
         let out = ensure_ok(out, "worktree list --porcelain").map_err(GcError::from)?;
         let entries = parse_worktree_list(&out.stdout);
 
--- a/crates/stella-fleet/src/git.rs
+++ b/crates/stella-fleet/src/git.rs
@@ -21,6 +21,8 @@ use std::path::{Path, PathBuf};
 use async_trait::async_trait;
 use tokio::process::Command;
 
+pub(crate) mod worktree_lock;
+
 /// The port every git command goes through. The real implementation
 /// ([`SystemGitCli`]) spawns `git`; tests inject fakes that record the
 /// invocation and return canned output.
@@ -403,13 +405,15 @@ impl<G: GitCli> WorktreeManager<G> {
         let dir = self.worktrees_root.join(&slug);
         let branch = format!("{}{slug}", self.branch_prefix);
         let dir_str = path_arg(&dir)?;
-        let out = self
-            .git
-            .run(
-                &self.repo_root,
-                &["worktree", "add", dir_str, "-b", &branch, base_ref],
-            )
-            .await?;
+        // Through the worktree lock, never straight to `GitCli`: two adds
+        // that overlap read each other's half-written `.git/worktrees`
+        // bookkeeping and one of them dies.
+        let out = worktree_lock::run_worktree(
+            &self.git,
+            &self.repo_root,
+            &["worktree", "add", dir_str, "-b", &branch, base_ref],
+        )
+        .await?;
         ensure_ok(
             out,
             &format!("worktree add {dir_str} -b {branch} {base_ref}"),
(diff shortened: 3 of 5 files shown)


macanderson:
**`dod / dod` is red because #6127's eight DoD boxes are unticked. Seven of the eight are met by this diff; the eighth is what CI is running right now.**

Checked each box against the diff at `eeff110`, not against the PR description:

| # | DoD item | Verdict | Evidence in this diff |
|---|---|---|---|
| 1 | No two overlapping `git worktree add` against one repo | **met** | `git/worktree_lock.rs::run_worktree` takes a per-repo-root `tokio::sync::Mutex` (`repo_mutex`, keyed on `lock_key`'s canonicalized path) and holds it across the git call; `WorktreeManager::create` routes through it. |
| 2 | Witness test pins it and fails on current code | **met** | `concurrent_creates_never_overlap_a_worktree_add` asserts `peak == 1` over 8 concurrent `create`s. The overlap is made by the fake's `await`, not by winning a race, so it fails deterministically without the lock. Control `the_fake_reports_overlap_when_nothing_holds_the_callers_apart` shows the counter reaches 8, so the assertion is measuring something that can climb. |
| 3 | `worktree remove` / `gc.rs` prune covered or shown not to need it, reason at the call site | **met** | Four further call sites moved onto `run_worktree` — `remove_worktree_and_branch` (covering both `WorktreeManager::remove` and `Gc`'s sweep), `WorktreeManager::discard`, and `Gc::sweep`'s `worktree prune` and `worktree list --porcelain`. Each carries its reason inline; the `list` site names the read as the half that lost run 9255. |
| 4 | `./scripts/test-session-isolation.sh` passes | **pending** | Runs inside `fmt + clippy + test`, still `in_progress` (started 23:00:33Z). Not ticked, because nothing has yet shown it passing. |
| 5 | Lands in `crates/stella-fleet`, no growth to a grandfathered file | **met** | New sibling submodule `src/git/worktree_lock.rs` (231 lines) + `worktree_lock/tests.rs`. `grep -c stella-fleet scripts/file-size-baseline.txt` → `0`, so the parenthetical "confirm this stays true" holds. `file size ratchet` is green on this PR. |
| 6 | Doc comment stating the guarantee and the non-guarantee | **met** | Module doc: "No two of them run against one repo root at once. Workers are not held apart. Only the tree cutting is." Both halves stated at the call site. |
| 7 | `README.md` updated to state the guarantee | **met** | File-map row for `src/git/worktree_lock.rs`, plus an invariant bullet naming all five call sites, what stays concurrent, and that the cross-process half is best effort. |
| 8 | No new prose trips `make prose` | **met** | `docs guards` is green on this PR, and that job runs the `prose` step. |

**So the only thing standing between this PR and a green `dod` is box 4**, and box 4 is the one box that cannot honestly be ticked from reading the diff — it is a claim about a test run. Once `fmt + clippy + test` reports green, all eight are satisfied and #6127's checklist can be ticked.

Ticking those boxes does not re-run this check by itself, but nothing here needs an empty commit either: `dod-recheck.yml` fires on an issue edit and re-runs the failed `dod / dod` on every open PR whose body names that issue, which is this one.

One note on the retry path, not a DoD item and not a blocker — `run_worktree` returns `Ok(first)` when the retry also fails, so a caller sees the first failure as an `Ok` `GitOutput` with `success: false`. That is what `ensure_ok` at every call site expects, and `a_retry_that_fails_reports_the_first_error` pins it, so this is a correct reading of the crate's convention rather than a defect. Recording it because the phrase "the caller sees the first error" in the PR body could be read as an `Err`.

---

Issue #6131, opened 2026-09-05
PipelineRole keeps two variants nothing can ever pin (Triage, Verifier)


macanderson:
## The problem

`crates/stella-tui/src/deck.rs`'s `PipelineRole` enum (`Triage`, `Worker`, `Verifier`) exists to group model-call receipts for the status bar and the (now much smaller) `/models` dialog. `#6087`'s own "What is not in scope" section said retiring this enum "is #3910's work, after #3906" — but #3910 closed (via #6107) without touching the enum at all. That PR's own description says so directly: "`crates/stella-tui/src/deck.rs`'s `PipelineRole` is three display slots already held to `ModelCallRole` by an exhaustive match in `PipelineRole::of`. No second language is involved." — #3910 retired the now-redundant *cross-language* `check-role-names.sh` guard, not the enum.

So nothing tracks retiring `PipelineRole` itself, even though most of what fed it is already gone:

- `PipelineRole::Triage` can never carry an *observed* pin — `PipelineRole::of` never returns it (the engine makes no triage call since the staged pipeline left). Its only would-be *configured* source, `crates/stella-cli/src/command_deck/model_cmd.rs`'s `configured_role_pins`, stopped naming it when #3908 retired `agents.triage.model` (its own comment: "They now render unconfigured, which is what they are").
- `PipelineRole::Verifier` is in the same position for the settings-driven path; `stella goal`'s live cross-family verifier selection is a different mechanism and does not resolve through this enum by role name.
- The `/models` dialog (`crates/stella-tui/src/views/models_card.rs`) stopped naming any of the three after #6087 — it now has a word for `default` (`lead`) only.

`PipelineRole::Worker` is the one live case: `configured_role_pins` still pins it, and `crates/stella-tui/src/views/plan_card.rs`'s `models_value` and `crates/stella-tui/src/views/status_source.rs` both read `model.role_pins` keyed on it for the PLAN card's operating envelope and the status bar's MODEL cell.

## What to do

Decide whether `PipelineRole` collapses to a single `Worker` case (or is removed entirely in favor of reading `stella_protocol::ModelCallRole` directly), given `Triage` and `Verifier` can no longer be populated by anything in the settings-driven path. Read `docs/spec/roleless-core.md` first — this may already be scoped by a later slice of epic #3903 that the epic's own checkbox list does not currently show as filed.

Files involved:
- `crates/stella-tui/src/deck.rs` — the enum, `PipelineRole::of`, `role_pins`, `active_role`.
[...]

PR #6136, opened 2026-09-05
feat(stella-cli,stella-tui): the /models dialog names every role, not just default

### LINKED ISSUE #6088: The models dialog names one role when the session has more

macanderson:
## What and why

`config_wiring::resolve` (the function behind both `stella config` and the
`/models` dialog) took no plugin roster, so `EngineConfigState::roles` —
the whole content of both surfaces — could never carry more than the one
`default` row. A session with an installed plugin's declared seat had no way
to see it: the one surface whose job is "which model runs what" said
nothing about it.

`resolve` now takes the same `(seat key, plugin name)` pairs
`agent::seats::installed_seats` already builds for the SEATS pane, and
appends one row per declared seat after `default`, resolved through
`engine.seat_models` + `parse_model_spec` — the same lookup
`resolve_seat_models` does for a real turn. An unassigned or unresolvable
seat falls back to the session's own model, matching what actually runs. A
new `ModelSource::Seat(String)` names the exact setting (`seat_models.<key>`)
a user would edit.

Both `Config::print_role_wiring` (`stella config`) and `config_wiring::deck_rows`
(`/models`) call this one `resolve`, so the two commands cannot name a
different set of roles — the guarantee `config_wiring`'s module doc already
states.

## The SEATS pane needed one line

The SETTINGS tab's SEATS pane (`views/seats.rs`, #6095) folds
`EngineConfigState::roles` and `EngineConfigState::seats` together into one
list. Once `roles` carries a row for every declared seat too, a seat would
draw twice — once via `roles` (named by the settings key) and once via
`seats` (named by the plugin). `rows()` now filters a `roles` entry out when
`seats` already names that key, so the pane's own tests and golden are
unchanged and a new witness (`a_role_the_driver_also_resolved_still_draws_once`)
pins the dedup.

## Witness mechanism

`config_wiring.rs` gains four tests exercising the new `declared` parameter:
a declared seat with no assignment rides the session model,
an assigned seat's row names `seat_models.<key>` and carries the assigned
model, an assignment naming no configured provider falls back the same way
`resolve_seat_models` does, and two declared seats resolve in the driver's
own order. All four fail on `main` by construction — `resolve` there takes
no fifth parameter at all.

A golden frame (`deck_render_snapshots/models_card_seats.rs`, alongside
#6095's `settings_seats.rs`) renders the `/models` card with a plugin seat
beside the session's own role, and pins that both rows draw. **This PR does
not include the golden `.txt` file** — blessing it requires compiling
`stella-tui`'s test binary, which this session cannot do (CI is the only
compiler here). The first CI run is expected to fail exactly that one test
and print the rendered frame in its panic message; I will read that output
and push the golden as a follow-up commit on this same branch.

## Exemplar

`crates/stella-tui/src/views/tools.rs` — rows sourced from the assembled
session stack, never a compiled-in table. #6095's own PR cites the same file
for the same shape.

## Stale comments repointed

- `config_wiring.rs` — the module doc, `DEFAULT_ROLE` and `RoleWiring::role`
  predicted this fold; reworded to state it as the current design.
- `profile.rs`'s `Plan::picks` doc predicted this issue would give a profile
  a second pick for plugin seats. It does not: a profile still tunes only
  the session's own model (`apply` writes `default_model`/`agents.default`
  alone), and the comment now says so.
- `command_deck/model_cmd.rs`'s `configured_role_pins` comment predicted this
  issue would replace the STATLINE/`/info` `Worker` pin with seat data. That
  is a different surface (`PipelineRole`), owned by #6131, and the comment
  now points there instead.

## Tests deleted

None.

## Residue

None beyond what is already tracked: `command_deck/model_cmd.rs`'s comment
now points at #6131 (already filed, open) rather than filing a new issue for
that gap.

Closes #6088

## Summary by Sourcery

Expose plugin-declared seats throughout model configuration surfaces while keeping seat management output deduplicated and correctly laid out.

New Features:
- Show every installed plugin-declared seat in both `stella config` and the `/models` dialog, including its effective model and configuration source.

Bug Fixes:
- Keep the SEATS settings pane from displaying plugin seats twice when those seats also appear in resolved role wiring.
- Preserve model-card layout and spacing for dynamically named, long plugin seat keys.

Enhancements:
- Unify role resolution across configuration reporting and the model dialog, with seat assignments falling back to the active session model when unavailable or invalid.

Documentation:
- Update role-wiring, profile, and model-pin documentation to reflect plugin seat reporting and distinguish it from pipeline-role pinning.

Tests:
- Add coverage for declared-seat resolution, assignment fallbacks, ordering, SEATS-pane deduplication and editing offsets, and long seat-key rendering.
- Add a `/models` card snapshot covering a plugin-declared seat.

### DIFF
--- a/crates/stella-cli/src/command_deck/model_cmd.rs
+++ b/crates/stella-cli/src/command_deck/model_cmd.rs
@@ -82,11 +82,10 @@ pub fn configured_role_pins(
         .and_then(|s| s.agent_engine_config);
 
     // `PipelineRole::Worker` alone. Triage and Verifier were pinned here from
-    // `pipeline_triage_model`/`pipeline_verifier_model` until #3908 retired
-    // both: settings can no longer express a model for either, so publishing a
-    // pin for them would be the deck reporting an intent nothing recorded.
-    // They now render unconfigured, which is what they are — and what a
-    // plugin's declared seat will replace in #6088.
+    // `pipeline_triage_model`/`pipeline_verifier_model`. Settings have no key
+    // for either one. A pin here would report an intent nothing recorded, so
+    // both render unconfigured. That is correct. A plugin seat feeds the
+    // `/models` dialog and `stella config` instead, not this pin.
     [PipelineRole::Worker]
         .into_iter()
         .map(|slot| {
--- a/crates/stella-cli/src/command_deck/panel_snapshots.rs
+++ b/crates/stella-cli/src/command_deck/panel_snapshots.rs
@@ -69,11 +69,13 @@ pub(super) fn engine_config_inbound(cfg: &Config, status: Option<String>) -> Inb
             model_efforts.insert(raw.clone(), levels.iter().map(|s| s.to_string()).collect());
         }
     }
-    let roles = crate::config_wiring::deck_rows(cfg, &providers);
     // What is installed, not what core knows: the seat list is the union of the
     // roles installed plugins declare, so a session with none shows the default
-    // model and nothing else (`doc:roleless-core` §8.4).
+    // model and nothing else (`doc:roleless-core` §8.4). Read once and shared
+    // by both the SEATS pane's own list and the `/models` roles below, which
+    // resolve a row per declared seat too.
     let declared = crate::agent::seats::installed_seats(&cfg.workspace_root);
+    let roles = crate::config_wiring::deck_rows(cfg, &providers, &declared);
     Inbound::EngineConfig {
         state: crate::engine_config::state_from_settings(
             &engine,
--- a/crates/stella-cli/src/config/report.rs
+++ b/crates/stella-cli/src/config/report.rs
@@ -79,12 +79,16 @@ impl Config {
         }
     }
 
-    /// The four engine roles, what each will actually send, and which setting
-    /// decided it.
+    /// Every engine role — the session's own, then one per plugin-declared
+    /// seat — what each will actually send, and which setting decided it.
     ///
-    /// Printed unconditionally, including when no engine settings exist — "all
-    /// four ride the session model" is an answer, and a block that appears
-    /// only sometimes cannot be used to check anything.
+    /// Printed unconditionally, including when no engine settings exist — "the
+    /// session rides its own model, no seats declared" is an answer, and a
+    /// block that appears only sometimes cannot be used to check anything.
+    ///
+    /// Resolves through [`crate::config_wiring::resolve`], the same function
+    /// [`crate::config_wiring::deck_rows`] calls for the `/models` dialog, so
+    /// this command and that dialog cannot name a different set of roles.
     fn print_role_wiring(&self) {
         use crate::config_wiring::{render, resolve};
         let configured = super::discover_configured_providers();
@@ -93,11 +97,13 @@ impl Config {
             provider: self.provider.id.to_string(),
             model: self.model_id.clone(),
         };
+        let declared = crate::agent::seats::installed_seats(&self.workspace_root);
         let wiring = resolve(
             self.engine_settings.as_ref(),
             &session,
             self.model_pinned_by_flag,
             &is_provider,
+            &declared,
         );
         println!("\n  {}", "Engine roles".bright_cyan().bold());
         for line in render(&wiring) {
--- a/crates/stella-cli/src/config_wiring.rs
+++ b/crates/stella-cli/src/config_wiring.rs
@@ -16,7 +16,7 @@
 //! precedence) rather than re-deriving the rules, because a config report that
 //! can disagree with the engine is worse than none.
 //!
-//! # One row, and why it is still a table
+//! # More than one row, and why it is still a `Vec`
 //!
 //! It reported six rows until #3908 — `default`, `worker`, `verifier`,
 //! `triage`, `research`, `plan` — of which four resolved settings keys that
@@ -25,25 +25,27 @@
 //! four roles that would never run, which is the most expensive place in the
 //! product to be wrong.
 //!
-//! The shape stays a `Vec` rather than collapsing to one struct because the
-//! rows come back as **plugin-declared seats**, resolved the same way and
-//! rendered by the same `/models` dialog (#6088).
-//! [`RoleWiring::role`] is a `String` for that reason: the deck's
-//! `envelope::roles::role_table` already renders a row for a role it has never
-//! heard of, so a seat needs no new plumbing here — only a row.
+//! It stayed a `Vec` past that collapse to one row, and #6088 is why: a
+//! plugin-declared seat resolves through the same [`resolve`] and lands in
+//! the same list, right after the `default` row. [`RoleWiring::role`] is a
+//! `String` for that reason: the deck's `envelope::roles::role_table` already
+//! renders a row for a role it has never heard of, so a seat needs no new
+//! plumbing here — only a row.
 
-use crate::engine_config::{ModelSpec, model_spec_for, tuning_for};
+use crate::engine_config::{ModelSpec, model_spec_for, parse_model_spec, tuning_for};
 use crate::settings::AgentEngineConfig;
 use stella_protocol::ReasoningEffort;
 
-/// The one role core has. Seats a plugin declares join this list in #6088.
+/// The session's own role. [`resolve`] appends one row per plugin-declared
+/// seat after it (#6088).
 pub const DEFAULT_ROLE: &str = "default";
 
 /// One role's resolved wiring.
 #[derive(Debug, Clone, PartialEq)]
 pub struct RoleWiring {
-    /// The role's name — [`DEFAULT_ROLE`], or (from #6088) a plugin-declared
-    /// seat. A `String` because core does not enumerate the possibilities.
+    /// The role's name — [`DEFAULT_ROLE`], or a plugin-declared seat's key
+    /// (`<plugin-id>/<role>`). A `String` because core does not enumerate the
+    /// possibilities.
     pub role: String,
     /// Provider id and the slug as it goes on the wire.
     pub model: ModelSpec,
@@ -60,7 +62,7 @@ pub struct RoleWiring {
 }
 
 /// Which setting decided a role's model.
-#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[derive(Debug, Clone, PartialEq, Eq)]
 pub enum ModelSource {
     /// `--model` / `STELLA_MODEL`. Pins the session for this invocation and
     /// suppresses the settings spec (see `resolve_engine_wiring`).
@@ -70,19 +72,24 @@ pub enum ModelSource {
     /// `default_model`.
     DefaultModel,
     /// Nothing named it: the role rides the session's resolved model, which
-    /// is the provider row's own default when settings are silent too.
+    /// is the provider row's own default when settings are silent too. Also
+    /// what an unassigned or unresolvable seat falls back to.
     SessionDefault,
+    /// `seat_models.<key>` — a plugin-declared seat the user assigned a
+    /// model to. Carries the key so the label names the exact entry.
+    Seat(String),
 }
 
 impl ModelSource {
     /// The settings key a user would edit to change this, as they would type
     /// it. Not a description — a path.
-    pub fn label(self) -> String {
+    pub fn label(&self) -> String {
         match self {
             Self::Flag => "--model (this invocation)".to_string(),
             Self::PerAgent => format!("agents.{DEFAULT_ROLE}.model"),
             Self::DefaultModel => "default_model".to_string(),
             Self::SessionDefault => "session default".to_string(),
+            Self::Seat(key) => format!("seat_models.{key}"),
         }
     }
 }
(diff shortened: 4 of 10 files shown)


macanderson:
**This PR's CI never started.** Not queued, not slow — the `pull_request` event was not delivered for any workflow file, so there is nothing to fail or even skip.

Evidence, rather than an inference from the checks tab being empty:

- Workflow runs for head branch `fix/6088-models-dialog-seats`: **`total_count: 1`**, and that one is `cla` (event `pull_request_target`, conclusion `skipped`). No `ci`, no `msrv`, no `docs-guards`, no `file-size`, no `guard-self-tests`, no `dod-check`.
- Check runs on head `efc6cfe`: **2** — `cla` and `Sourcery review`, both `skipped`.
- For contrast, #6135 (opened 23:59:07Z… five minutes before this one, same base `main`) carries **24** check runs, and #6133 carries 28. Both got their full set within ~20 seconds of opening.
- Opened 23:04:23Z; still 2 checks ten minutes later.

This is the shape AGENTS.md § "the gate" records as a rare platform gap — "only `cla` (`pull_request_target`) — every `pull_request`-triggered workflow never ran at all". The documented instance was a PR based on a non-`main` branch; this one's base **is** `main`, so the stacked-base explanation does not apply here.

The remedy AGENTS.md names is a fresh `synchronize` event — push a commit. A hand-run `workflow_dispatch` is explicitly not a substitute, because it runs the branch head rather than the merge commit that would land.

I have not pushed anything to this branch, deliberately: the PR body says a follow-up commit is already planned here ("I will read that output and push the golden as a follow-up commit on this same branch"), and pushing to an active branch to kick CI would collide with that. That follow-up commit is itself the `synchronize` event this needs — so pushing the golden `deck_render_snapshots/models_card_seats.txt` will both start CI and settle the one test the body expects to fail.

Worth knowing before that push: with CI never having run, the first run will be the first evidence of *anything* on this branch, not just of the golden frame. The body's expectation that exactly one test fails is currently unverified by any run.

---

macanderson:
## The `card_models_seats` golden is blessed — and it caught a rendering bug

This PR's description says the golden `.txt` was left out because the session had no compiler, and that the frame would be read off CI's panic output and pushed as a follow-up. This run had a compiler, so the golden is blessed and **read** (`00dda7a`).

Reading it is what mattered. The first blessing drew the seat row as:

```
│acme/revieweranthropic/claude-opus-5                                  │
```

`role_rows` laid the role word out with `format!("{word:<LABEL_W$}")`, and `LABEL_W` is `10` — the width chosen when `default` was the only word the driver could send. Rust's `<` fill pads to a **minimum** and never cuts, so a 13-column plugin seat key abutted its model with no gap and carried the line past the card's inner width. `header_rows` and the wrap indent read the same constant, so the table lost its single left edge too.

Blessing that as the expected frame would have committed the defect as the contract, so the layout is fixed rather than recorded:

- `label_w` measures the words actually drawn — the header's own labels and every role in the table — floored at `LABEL_W` and capped at half the card so one long seat key cannot squeeze the model column out.
- `label_cell` cuts before padding, through `render::columns`, whose `pad` doc already names this hazard: `{:<width$}` counts chars, not columns.
- `render` and its tests now fold through one `body_rows`, so the label column cannot be measured one way for the frame and another for the assertion — what `body_of`'s own doc comment asked for while hand-copying the fold.

Witness: `a_seat_key_wider_than_the_label_column_still_keeps_its_gap` asserts the key and model never abut, that the header widens with the table, and that no row exceeds the card's inner width. It fails on the padded-constant layout on all three counts.

The blessed frame now reads:

```
│session        glm-5.2                                                │
│auto           effort off · thinking off                              │
│                                                                      │
│lead           zai/glm-5.2-air                                        │
│               medium · thinking on · from default_model              │
│acme/reviewer  anthropic/claude-opus-5                                │
│               provider default · thinking default                    │
│               from seat_models.acme/reviewer                         │
```

Every card border in the golden measures to columns 44 and 115. The other 41 frames pass unchanged, so no existing card shifted.

### Evidence

Run locally against this branch's head, `stella-tui` only:

- `cargo test -p stella-tui --lib views::` — 449 passed
- `cargo test -p stella-tui --test deck_render_snapshots` — 42 passed (no `BLESS`)
- `cargo clippy -p stella-tui --all-targets` — exit 0, no diagnostics
- `cargo fmt --check -p stella-tui` — clean
- `scripts/check-prose.py` — OK, none added

CI is the authoritative gate; this is what was checked before pushing.

### Two notes for whoever picks this up

- **The merge conflict against `main` is resolved** — it was a doc comment on `seats::rows`, but the auto-merged body carried a real composition break: #6126's editor offsets the seats by `state.roles.len()` while this branch's dedup filter drops roles from that leading block, so `⏎` on one seat would have assigned the one above it. A parallel session reached the same fix independently (`resolved_row_count`, with `the_editor_finds_the_right_seat_when_roles_also_resolved_it`); that version is what is on the branch now, and this push is rebased onto it rather than competing with it.
- `Refs #6088` on the commit rather than `Closes` — the issue is already closed by this PR's own first commit, and one issue takes one closing keyword.

---

macanderson:
## Answering Sourcery's one ❌ row: the golden is committed

Sourcery's assessment table marks the third #6088 objective ❌, on this reasoning:

> the new `/models` golden snapshot test is incomplete: `models_card_seats.rs` references the `card_models_seats` golden file, which the PR explicitly does not include. The test is therefore expected to fail in CI, so the PR does not fully complete the requested verification.

**That was true of the head it reviewed and is not true of the head now.** Sourcery reviewed `efc6cfe`, before the golden existed. `00dda7a` added it:

- `crates/stella-tui/tests/snapshots/deck/card_models_seats.txt` — `added`, 45 lines, present in this PR's file list.

So the file the row calls absent is present, and the test it expects to fail is the one whose blessing found the `LABEL_W` layout bug that `00dda7a` also fixes.

The row cannot clear itself: **Sourcery's review budget is exhausted** (`review budget of 250,000 diff characters for the last 7 days`, ~22h to reset), which is why the `Sourcery review` check reports `skipped` on both later heads instead of re-reviewing. The ❌ therefore stands against a head that no longer exists, and this comment is the answer CLAUDE.md asks for when a row is not going to be re-judged.

### What the current head shows

Head `e7ad62c`: 24 of 25 check runs complete, **all green** — `dod / dod`, `docs guards`, `guard self-tests`, `cargo check on the declared MSRV`, `cargo deny + cargo audit`, `file size ratchet`, `no edit is undone inside the branch`, `duplicate claims`, CodeQL across four languages, and the rest.

`fmt + clippy + test` is still in flight, and it is the job that runs the snapshot test — so it, not this comment, is the evidence that settles the row. I am watching it rather than claiming its result.

The DoD half is independently settled: `dod / dod` passed and the SCR-003 bot posted *"every linked issue's definition of done is fully checked"* against #6088.

I have pushed nothing here — the fix was `00dda7a`, already on the branch. This is the stale-row answer, not a second attempt at it.

---

macanderson:
Addressing Sourcery's ❌ on the third linked-issue objective ("the new `/models` golden snapshot test is incomplete... expected to fail in CI"):

That was accurate against the commit Sourcery reviewed. Two follow-up commits since then close the gap:

- `00dda7a91` blessed `card_models_seats.txt` against a real compile, and in the process found and fixed a real layout bug: the label column was a fixed `LABEL_W` sized for `default`, so a wider plugin seat key (`acme/reviewer`) abutted the model column with no gap and overran the card's border. It replaced the constant with a measured `label_w`, folded `render` and its tests through one `body_rows`, and added `a_seat_key_wider_than_the_label_column_still_keeps_its_gap` as the witness.
- `e7ad62c78` fixed a `rustdoc::private_intra_doc_links` error the first fix introduced (`pub fn rows`'s doc linked three private helpers).

`fmt + clippy + test` and `guard self-tests` — the two jobs that build and run `deck_render_snapshots` — are both green on the current head, so the golden test referenced in the ❌ row now passes rather than being expected to fail.

macanderson:
Following through on the evidence I said I'd wait for rather than claim: **`fmt + clippy + test` passed** on `e7ad62c` — [job 101395670212](https://github.com/macanderson/stella/actions/runs/33999455763/job/101395670212), completed 00:04:05Z, `conclusion=success`. That is the job that runs `deck_render_snapshots`, so the `card_models_seats` golden the Sourcery row called absent is present *and* passing. The row is settled on a run, not on my reading of the file list.

Head `e7ad62c` is now `success=22, skipped=3, cancelled=1` — the single `cancelled` is a superseded duplicate `dod / dod`, replaced by the one that passed. `mergeable_state` is `clean`.

Leaving the merge to you: no approving review stands on this PR, and a green mergeable head is what waits on a reviewer.

---

macanderson:
Residue filed: #6141 — `EngineConfigState.roles` and `.seats` now duplicate a declared seat's data, and `views/seats.rs` reconciles them with a key filter (`declared_seat_keys`/`resolved_row_count`) rather than the driver sending one list. Everything works and is tested; it's a design question for a later PR, not a bug in this one.

PR #6168, opened 2026-09-06
feat(stella-core,stella-cli): one steering allowance for records, skills, frames and tools

### LINKED ISSUE #6110: Records and tools still spend two budgets, so a rule and a tool schema are never weighed against each other

macanderson:
## What and why

The steering plane had five sources and two budgets. Records, skills and recalled frames were packed against one number; tool schemas were packed against another. Both went through `pack_to_budget` and both honoured the same `source_rank` precedence, and neither could see what the other took — so a rule and a tool schema were never weighed against each other, which is the thing the plane exists to do.

The two costs land at different moments and no single call can see both: the recall block is rendered before the turn opens, the tools array is assembled after, when the driver builds its stack. So they meet in a shared cell.

**`stella_core::steering::ledger::SteeringLedger`** — a per-session cell holding what the volatile block spent (`AtomicU64`) and the tool budget that spend settled. `spend` is saturating. `settle` answers once per declared allowance and keeps that answer.

- **The spend** goes in at `memory::inject_opening_recall`, the one seam every driver's opening block already passes through (`every_recalling_driver_routes_its_block_through_the_opening_seam` holds that). Measured over the bytes actually injected, so what the plane records and what the provider is sent are one number — and charged only when the block is genuinely appended, since a block the dedup refuses is one an earlier turn already paid for. `inject_recall_block` now answers whether it appended, which is what makes that distinguishable.
- **The read** goes in at `agent::tool_stack::budgeted`, which already takes a `&Config`. So the handle rides `Config` as an `Arc`, beside `durability` and `output_ceilings` — the two shared cells already there — and no driver has to carry a second argument.
- **One number.** `context.steering.tools.max_tokens` moves up to `context.steering.max_tokens`: the whole volatile allowance, not the tool arm's own. `context.steering.tools.mcp_max_tokens` stays as a cap on one source (the shape `skills.max_skills` has beside the budget it sits under) and is narrowed with the allowance, so it can never exceed what the block left.

## Byte-stable prompts

The tools array is prompt-cache prefix. A running total that shrank as the session went on would re-rank it and re-bill the whole conversation, so `settle` is keyed on the **declared** allowance: the same allowance always yields the same array however much has been spent since, and a `/reload` that changes the allowance settles a fresh one — the one cold write `tool_lean`'s module docs already promise an operator who edits the key. `steering_ledger` is `Posture::StartupOnly` in the reload ledger for that reason, with `tool_advertisement` beside it carrying the reloadable half.

## Two ordering repairs

- **The fleet worker** built its tool stack (`policy_stack`) *before* it recalled (`WorkerSteering::open`), so its array would have settled against a spend that had not happened yet. The stack call moves below the recall. Nothing between the two used `permitted`.
- **A fleet attempt takes its own cell** rather than the dispatcher's clone, or N parallel workers would charge one allowance N times and the last one to ask would advertise almost nothing.

## Witness

`crates/stella-cli/src/agent/tool_stack.rs`: `a_turn_whose_records_fill_the_allowance_advertises_fewer_tools` builds two session stacks over one 40-tool set and one declared allowance — one ledger fresh, one that has already spent three quarters of it as a records-heavy turn would — and asserts the second advertises strictly fewer schemas, with the first advertising all forty as the control. **On the old code `budgeted` read the declared budget and nothing else, so both arms advertise the same list and the `<` assertion fails.**

Beside it: `a_later_block_does_not_move_an_array_the_session_already_settled` (the cache-prefix half, at the seam that ships) and `the_lever_off_advertises_every_tool_whatever_the_block_spent` (with the lever off nothing is withheld, whatever the block spent). `crates/stella-core/src/steering/ledger.rs` carries the arithmetic's own six, including saturation and the reload path.

## Exemplar

The shared-cell-on-`Config` shape follows `Config::output_ceilings` and `Config::durability` already in the same struct: a handle the session mutates, `StartupOnly` in the reload ledger, `moved: false` because it has no equality and holds spend rather than settings.

## Tests deleted

None.

## Evidence

`make guards-fast` green locally (prose, file-size, line-citations, dead-code-allows, lockfile-sync, fmt). Everything that compiles is CI's.

## Remaining work filed

- #6169 — the settings unknown-key walker does not descend into `context`, so an operator's
  now-stale `context.steering.tools.max_tokens` is ignored with no notice.
- #6194 — the ledger never gives back what compaction drops, so a compacted session that
  then reloads its allowance re-settles against bytes the prompt no longer carries.

Closes #6110

## Summary by Sourcery

Unify volatile context and tool advertisement budgeting through a shared, cache-stable steering allowance.

New Features:
- Unify records, skills, recalled frames, and tool schemas under a shared per-session steering allowance.
- Add session-scoped accounting so recall usage reduces the tool schemas advertised for the turn.
- Move the steering budget setting to `context.steering.max_tokens` while retaining the MCP-specific cap.

Bug Fixes:
- Prevent later steering spends from changing an already-settled tool array, preserving prompt-cache stability.
- Ensure deduplicated recall blocks are not charged and fleet attempts do not share steering spend.
- Build fleet tool stacks after opening recall so they reflect the allowance remaining after recall.

Enhancements:
- Introduce a shared steering ledger with saturating spend, allowance settlement, and stable results per declared budget.
- Give fleet workers independent steering ledgers and narrow MCP caps to the remaining shared allowance.

Tests:
- Add ledger arithmetic tests covering shared spending, saturation, cap narrowing, stable settlement, and allowance changes.
- Add tool-stack tests verifying records reduce advertised tools while full steering preserves the complete tool surface.
- Update recall injection tests to verify whether blocks were appended and charged.


macanderson:
Remaining work filed: #6169 — the settings unknown-key walker does not descend into `context`, so an operator's now-stale `context.steering.tools.max_tokens` is ignored with no notice, and `RETIRED` cannot fire for it.

macanderson:
## Sourcery's ❌ on objective 1: right about the cause, overstated about the effect

The row says the ledger accumulates across turns and so later turns get no independent per-turn allowance. Half of that is confirmed and half of it is not, and the distinction changes what needs fixing.

**Confirmed — there is no turn boundary anywhere.** `SteeringLedger` exposes `new`/`spend`/`spent`/`settle` and nothing that resets `spent`. The handle lives on `Config` for the session, and `memory::inject_opening_recall` (`recall.rs:219`) is the only production caller of `spend`. That call sits per turn, not per session — `command_deck.rs:1823` is inside the block commented *"Per-turn conversation bookkeeping, mirroring `run_interactive`: refresh the volatile recall block…"*. So `spent()` is a monotonic session total across every turn.

**Overstated — tools are not progressively starved.** `settle` memoizes on `declared`, so the first settle wins and every later accumulation is inert for that allowance. `a_later_spend_does_not_move_a_settled_allowance` pins exactly that, and it is what Invariant 7 and DoD item 2 require. Steady-state behaviour is correct, and objective 2's ✅ is consistent with it.

**The real defect is the re-settle path.** The memo is keyed on `declared`, so a `/reload` that moves `context.steering.max_tokens` settles anew — against the accumulated total of *every prior turn's* recall block, not the current turn's. `a_changed_allowance_settles_again` pins that path with a single 400-token spend and passes; with N turns of spend behind it, `saturating_sub` drives the new allowance to zero and the operator who just *raised* their steering budget silently gets no tools. The test passes because it only ever spends once, which is the shape the accumulation bug hides behind.

That also leaves DoD item 1 unmet as written:

- [ ] One **per-turn** allowance is spent by both consumers, in `source_rank` order, with no second budget number.

There is no per-turn allowance in the diff — there is one per-session running total.

**Shape of the fix**, since items 1 and 2 read as contradictory and are not: reset `spent` at each turn's opening while leaving `settled` alone. Item 1's per-turn allowance and item 2's settled-once-per-session tool array then both hold, and the re-settle after `/reload` charges one turn rather than the whole session. The threading is the work the issue already calls the work — a turn boundary has to reach all six drivers that call `inject_opening_recall` (`agent.rs:497`/`609`, `goal.rs:290`/`687`, `command_deck.rs:1823`, `fleet_cmd/steering.rs:89`) — so I am flagging rather than pushing it; deciding which call site owns the boundary is yours, not a small in-scope repair.

Worth a regression test either way: settle, spend across several simulated turns, re-settle with a *larger* `declared`, and assert the allowance grew.

No merge conflict against `main`; CI was still in flight when I looked. #6169 is already filed for the unknown-key walker, noted above.

---

macanderson:
## On Sourcery's ❌ for #6110's first objective

I checked this against the code rather than the description, and **the ❌ is right on the facts**, though the harm is not quite where the explanation puts it. Worth separating the two, because one half of Sourcery's reasoning is answered by the design and the other half is a real defect.

### What is actually true

`SteeringLedger::spent` is an `AtomicU64` that only ever grows — `spend` is a saturating add and nothing resets it. The ledger is built once per session and rides `Config` as an `Arc`. `inject_opening_recall` charges it whenever a block is *genuinely appended*, and the dedup it keys on is exact content equality against prior markers — so a turn whose recalled block differs from every earlier one (the normal case, since recall tracks the prompt) pays again. Session-lifetime spend is therefore the sum of every distinct block the session has injected, not one turn's.

### Where Sourcery's explanation over-reads

"Later turns do not receive an independent per-turn allowance" is true of `spent`, but it does not reach the advertised tool array, because `settle` caches `(declared, answer)` and returns the cached answer for an unchanged `declared`. `a_later_spend_does_not_move_a_settled_allowance` pins exactly that, and it is what invariant 7 requires: the array is prompt-cache prefix, and a total that shrank as the session ran would re-rank it and re-bill the conversation. So on the steady-state path the growing total is inert, and the second and third objectives really are met.

### Where it is a real defect

The accumulation stops being inert the moment `declared` changes, because `settle` then recomputes:

```rust
let max_tokens = declared.max_tokens.saturating_sub(self.spent());
```

`ToolAdvertisement` is `Posture::Reloaded`, and `steering_ledger`'s own `StartupOnly` note says the ledger "settles again whenever the declared allowance changes" — so `/reload` is a live path to this, by design. On that path `self.spent()` is the whole session's history, not the current turn's block.

`a_changed_allowance_settles_again` covers it with a single 400-token spend and reads correctly (1000 → 2000 gives 1600). Extend it to the shape a real session has and it inverts:

- 20 turns, ~500 tokens of distinct recall each → `spent() == 10_000`
- operator raises `context.steering.max_tokens` 8_000 → 12_000 and `/reload`s
- `settle(12_000)` → `12_000 - 10_000` = **2_000**

They raised the allowance by half and the tool array got smaller. Past roughly 24 turns the subtraction saturates and `max_tokens` is **0** — `mcp_max_tokens` is `.min(max_tokens)` so it goes to 0 with it, and the session advertises no tools at all until it restarts. The drop message added in this PR would name every withheld tool, so it fails loudly rather than silently, but "raise the allowance, lose every tool" is the opposite of what the operator asked for.

### What I think it wants

The smallest fix that keeps invariant 7 is to settle against **the block that is current**, not the running total: record the last charged block's cost in its own cell (`spend` sets it rather than adding), and have `settle` subtract that. Steady state is unchanged — the first `settle` still sees turn 1's block and the answer is still cached — but a re-settle then reflects one turn, which is both what `a_changed_allowance_settles_again` already asserts for the single-spend case and what #6110's first DoD item asks for in the words "one **per-turn** allowance".

The alternative — a real `begin_turn()` reset threaded through every driver — is the fuller reading of the issue and is the threading work #6110 calls "the work", but it buys nothing over the above for the only consumer that reads the number.

I have **not** pushed either. Turn-versus-session lifetime for the context plane's shared cell is an architectural call on a crown-jewel path, and this PR argued the session-scoped choice deliberately from invariant 7; picking the other reading unasked, on your PR, is your call and not mine. Flagging it with the failing arithmetic so the choice is made rather than inherited.

### DoD

I left #6110's first box unticked for the same reason, so `dod / dod` stays red and correctly holds the merge. Boxes 2, 3 and 5 I did verify — `cargo test -p stella-core --lib steering` at `85f2e4b` is **23 passed, 0 failed**, including all six `steering::ledger::tests` and `the_same_tool_set_advertises_the_same_array_twice`. Box 4 (`GatheredSteering`'s doc comment) I read and it is updated. Ticking the rest is one edit away once the first is settled.

---

macanderson:
## Answering Sourcery's ❌ — #6110's DoD contradicts itself, and this PR picked the invariant

Sourcery marks the first objective unmet: the ledger "is stored and spent for the lifetime of a session rather than reset per turn," so its total accumulates across turns.

The observation about the code is right. `SteeringLedger` rides `Config` as an `Arc`, `spend` saturates into a session-long `AtomicU64`, and nothing resets it at a turn boundary. What I do not think follows is that this is a defect, because #6110 asks for two things that cannot both be true.

- DoD item 1: *"One **per-turn** allowance is spent by both consumers."*
- DoD item 2: *"The tool allowance a session resolves is settled once **per session**, not recomputed per step."*
- Constraint: *"Invariant 7: the advertised tools array must stay fixed for a session ... the tool allowance must be settled once and not re-read per step."*

A genuinely per-turn allowance re-sizes the tool array on every turn, which is the prompt-cache re-billing invariant 7 exists to forbid. This PR resolved the tension toward the invariant, which reads as the right call.

### The accumulation is inert where it would hurt

`settle` memoizes on the declared budget — it answers once and keeps that answer, so a later `spend` changes no advertised array. `a_later_block_does_not_move_an_array_the_session_already_settled` pins exactly that. The shipped behaviour is that the tool array is sized by the **first** turn's recall spend and then held for the session, which is what item 2 and the constraint ask for.

So the accumulating counter never reaches the tools consumer after settlement. The divergence is in item 1's wording, not in what a session does.

### What I am not doing

Ticking item 1 would be recording a box as met when the shipped design says "per session" and the box says "per turn". Rewording the issue is a maintainer's call, not mine. Your options as I read them:

1. Reword item 1 to the per-session allowance the constraint already implies, tick the boxes, and let this close #6110.
2. Keep item 1 as written, downgrade to `Refs #6110`, and leave the per-turn question open.

The remaining four boxes look satisfiable on this diff — item 2 and the witness are in the code, and items 4 and 5 (`GatheredSteering`'s doc comment, the golden block byte-identical with the lever off) are settled by the CI run in flight rather than by me.

---

macanderson:
Automated PR sweep. Sourcery's ❌ against #6110 is unanswered; I checked it against the code and it is directionally right but wrong about the harm, which changes what the fix is.

## What the code does

`SteeringLedger` (`stella-core/src/steering/ledger.rs`) holds `spent: AtomicU64` and `settled: Mutex<Option<(ToolBudget, ToolBudget)>>`. There is **no** `reset`/`clear`, and no `store(0)` anywhere. The production instance rides `Config` as `Arc<SteeringLedger>` (`config.rs:406`) and is `Posture::StartupOnly` in the reload completeness table — deliberately preserved across a reload, with the reason given:

> a fresh one would re-rank the advertised tools mid-session, and that array is cache prefix

So `spent` is session-cumulative. Its own accessor says so: *"What this **session's** block has spent so far."*

## Where Sourcery is wrong

Its stated consequence is that "later turns do not receive an independent per-turn allowance". In the ordinary path they are not starved, because `settle` is memoized:

```rust
if let Some((for_allowance, answer)) = *held && for_allowance == declared {
    return answer;
}
```

*"Answered once, then kept. Ask again with the same `declared` and the answer is the same, however much was spent since."* The tool array is fixed on the first settle and does not degrade turn over turn. That is DoD box 2 — "settled once per session, not recomputed per step" — and it is **met**, on purpose.

## Where it is right, and the actual bug

DoD box 1 asks for "**one per-turn allowance**". The ledger is per-*session* with no reset, and the memoization only hides that while `declared` is unchanged. Change it — which is exactly the operator edit the completeness note points at (*"the ledger settles again whenever the declared allowance changes"*) — and the re-settle runs:

```rust
let max_tokens = declared.max_tokens.saturating_sub(self.spent());
```

against a `spent()` holding **every turn of the session**, not the current turn's block. Twenty turns in, `saturating_sub` floors it and the operator who just *raised* the allowance gets a tool array of zero. The one path the design intends to stay live is the one that misreads the number.

So boxes 1 and 2 pull against each other, and the branch satisfies 2 by making 1 untrue rather than by separating the two quantities. A per-turn `spend` and a session-sticky `settle` are both wanted; they cannot be one counter.

## Suggested shape (yours to accept or reject)

Keep `settled` session-scoped as it is, and give `spent` a turn boundary — a `open_turn()`/`reset_spend()` the driver calls where it already begins a turn — so a re-settle subtracts this turn's block rather than the session's. Then box 1 is true as written and box 2 keeps its cache-prefix guarantee.

If instead the intent is that box 1 always meant "one allowance, session-scoped", then the box needs rewording and #6110's wording is the thing to fix — but the re-settle arithmetic above still wants a look either way.

`dod / dod` is red on this PR anyway (the #6110 boxes are unticked), so this is in time. Box 4 I did check and it is met: `GatheredSteering`'s doc no longer names #6110 as a gap and describes `SteeringLedger` and the shared cell. Boxes 3 and 5 I did not verify.

---

macanderson:
## `cargo clippy -D warnings` is red, and it is this branch's own change

Two separate CI failures on this PR. I fixed one and am standing down on the other; both diagnoses below.

### Fixed and pushed — the test target did not compile

`WorkerSteering::open` gained a sixth parameter here and the production call site in `fleet_cmd.rs` took it, but three test call sites did not, so `stella-cli`'s test target failed with three `E0061`s:

- `crates/stella-cli/src/fleet_cmd/steering.rs` (two, in its own `mod tests`)
- `crates/stella-cli/src/fleet_cmd/tests.rs` (one)

`macos-check.yml` is what caught it — the only job here that runs `cargo check -p stella-cli --all-targets`. Evidence: run [34001998022](https://github.com/macanderson/stella/actions/runs/34001998022) **failure** on `85f2e4b`, run [34002301921](https://github.com/macanderson/stella/actions/runs/34002301921) **success** on `236e243`. Each site takes a fresh ledger, matching the per-attempt isolation this PR's own description argues for.

### Not fixed — `too_many_arguments`, and it wants a design call

```
error: this function has too many arguments (8/7)
   --> crates/stella-cli/src/agent/tool_stack.rs:121:1
    = note: `-D clippy::too-many-arguments` implied by `-D warnings`
```

**This predates my commit and is not a consequence of it.** `session_stack_with_gate` takes 7 arguments on `main` and 8 here — this branch added `ledger: &SteeringLedger`, crossing the limit. It sits at line 121, above the `#[cfg(test)]` at 306, so it is production code and the `bin "stella"` target flags it independently of the test-target errors above. The lint was therefore already firing on `85f2e4b`; it was simply not the first thing to fail.

I am not pushing a fix, for two reasons:

1. **The obvious shortcut is the one CLAUDE.md forbids.** An `#[allow(clippy::too_many_arguments)]` here would assert the lint is wrong, and for eight positional parameters of mixed types it is not wrong. That is the expedient, not the repair.
2. **The real repair is a signature change across ~19 call sites** (`subagent/tests.rs`, `plugin_cmd/tests/contributions.rs`, `agent/tests.rs`, `memory/tests/skill_event.rs`, and nine inside `tool_stack.rs` itself), on the tool-stack assembly path — and it interacts with the open question above it. If the per-turn/per-session decision on #6110 adds a turn boundary, this parameter set changes again, and I would be refactoring a signature that is about to move.

**Proposal, for whoever takes it:** bundle the two parameters this PR itself coupled rather than inventing a new grouping. `advertisement` and `ledger` are two halves of one concept in this design — the ledger is what settles the advertisement — and they already travel together on `Config`. A `SteeringAllowance<'a> { advertisement: ToolAdvertisement, ledger: &'a SteeringLedger }` takes the count to 7, clears the lint, and reads as better than the current shape rather than as lint-appeasement.

Verified locally with `cargo clippy -p stella-cli --all-targets --locked -- -D warnings` (exit 101, that one error).

---

macanderson:
## CI red on `236e243` — clippy only, one lint, one site

`ci` run [9312](https://github.com/macanderson/stella/actions/runs/34002301922) failed at step 45, `cargo clippy -D warnings`. Everything else on that run is green — `cargo test`, `cargo doc -D warnings`, `docs/wire`, the release smoke build, and all 40-odd guards. So this is the last thing between the branch and a green run.

```
error: this function has too many arguments (8/7)
   --> crates/stella-cli/src/agent/tool_stack.rs:121:1
    |
121 | / pub(crate) fn session_stack_with_gate<'a>(
...
    = note: `-D clippy::too-many-arguments` implied by `-D warnings`

error: could not compile `stella-cli` (bin "stella") due to 1 previous error
error: could not compile `stella-cli` (bin "stella" test) due to 1 previous error
```

`ledger: &SteeringLedger` is the eighth parameter. The function sat at seven, which is exactly clippy's ceiling, so this branch is the one that crosses it.

`macos-check.yml` could not have caught it — that job runs `cargo check`, not clippy.

### Two ways to clear it

**Group the layer's inputs into a struct** — the mainstream answer to this lint, and the one that survives the ninth parameter. Roughly `policy` / `advertisement` / `ledger` (what the layers are configured with) as one `SessionStackConfig`, leaving `base`, `custom_tools`, `workspace_root`, `gate` and `principal`. It touches the two production call sites in `tool_stack.rs` plus about fifteen test sites across `agent/tests.rs`, `memory/tests/skill_event.rs`, `plugin_cmd/tests/contributions.rs`, `subagent/tests.rs` and `tool_stack.rs`'s own tests — mechanical, but real churn on a branch that is otherwise finished.

**`#[allow(clippy::too_many_arguments)]` with the reason written above it.** Defensible here, and the function's own doc already makes the argument: *"[`session_stack`] with every dependency explicit — the seam the witness tests inject a denying gate through, so the chain they prove is the chain the drivers ship."* Explicitness is the point of this function, which is the case for the lint being wrong at this site. CLAUDE.md bans the bare attribute, not the justified one — so the comment is the whole cost.

I did not push either. This is not my branch, and a commit landed on it 20 minutes ago, so a fix from me would have collided with whoever is driving it.

---

macanderson:
Automated PR sweep. Two things changed on this PR since the comments above, and one of them needs your eye.

**The clippy break is fixed.** `d923086` groups `advertisement` + `ledger` into `ToolAllowance`, taking `session_stack_with_gate` back to seven arguments with no `#[allow]` — the shape the comment at 01:19 proposed. Verified on that head: `cargo clippy -p stella-cli --all-targets -- -D warnings` exits 0. I had written the same refactor independently and discarded it rather than push a duplicate.

**#6110's DoD boxes were all ticked at 01:24**, box 1 included, so `dod / dod` no longer holds this PR. Five comments above argue box 1 ("one **per-turn** allowance") is not what the code does, and I checked that against `ledger.rs` rather than against those comments: `spend` is a saturating add, the type exposes no reset, and `settle` on a changed `declared` computes `declared.max_tokens.saturating_sub(self.spent())` against the session-cumulative total. So the gate that was carrying this question is now green while the question is still open.

I have not un-ticked the box — whoever ticked it may have decided the per-session reading is the intended one, and that call is yours. I have filed **#6186** for the arithmetic itself, with the multi-turn regression test that separates the two readings, so the finding is tracked wherever the wording lands.

Nothing else on this PR is red from my side; the remaining checks were still in flight when I looked.

---

macanderson:
Answering Sourcery's one ❌ row — the per-turn lifecycle one. It reads the
code correctly and names no behaviour, and the session scope is the one the
issue's own constraints require.

**`spent()` is read in exactly one place.** `SteeringLedger::settle` is its
only caller; nothing else in the tree reads the total. So it is not a running
number any other code reacts to.

**Past the first settle the total is inert.** `settle` is keyed on the
*declared* allowance and keeps its answer. Turn 1 spends 400 and settles 600;
turn 2 spends 400 more, asks again with the same declared allowance, and gets
600 back. `a_later_spend_does_not_move_a_settled_allowance` is that at the
arithmetic, and `a_later_block_does_not_move_an_array_the_session_already_settled`
is it at the assembled stack. A per-turn reset would be observationally
identical here, because the kept answer never re-reads the total.

**Where it is not inert, accumulating is the right number.** The one moment
`settle` recomputes is a `/reload` that moves the allowance mid-session. At
that point the accumulated total is what the conversation actually holds:
every turn's opening block is appended to `messages` and stays there, and the
tools array sits in the same request as all of them. A per-turn reset would
subtract the last block alone and hand the tool layer room that four earlier
blocks still occupy. The session total is the honest thing to weigh a
session-wide cache prefix against.

**The issue asks for both, and the constraint decides.** Its title says
per-turn; its constraints say "the tool allowance must be settled once and
not re-read per step", because invariant 7 makes the tools array prompt-cache
prefix. A cell whose spend resets per turn and whose settlement does not is
one object with two lifetimes. This is the settlement's lifetime, and the PR
body says so rather than borrowing the title's word.

**What the accumulation genuinely misses** is compaction: nothing releases a
block's spend when compaction drops it from `messages`, so a long compacted
session that then reloads its allowance would subtract bytes the prompt no
longer carries. Filed as #6194 rather than papered over — it needs a release
path on the ledger and a seam at the compactor, which is its own change.

Issue #6118, opened 2026-09-05
Serve the work and deliver verbs over the driver channel


macanderson:
## The problem

The self-driving plugin is now a program Stella starts, and it asks over the driver channel (#6063). The host serves exactly one verb: `backlog_next`. Every other verb answers `HostCallRefusal::Unsupported`.

So a cycle stops at the claim. `plugins/stella-selfdriving/main.py` reads the ranked queue, asks `backlog_claim`, is told the host does not do that, and halts. It cannot take an issue, run a turn, open a pull request, or merge. `scripts/self-driving.sh` still holds all of that, running as your user.

The work each verb needs already exists in the binary, behind CLI verbs rather than behind the channel:

- `backlog_claim` / `backlog_link` — `crates/stella-cli/src/self_driving_cmd/claim.rs` and `claim_mirror.rs`, over the `dispatch_claims` lease table in `.stella/private/fleet.db`.
- `backlog_file` / `backlog_close` — `IssueProvider::file` and `IssueProvider::close` (`crates/stella-protocol/src/issue.rs`), through `crates/stella-cli/src/self_driving_cmd/residue.rs` and `lifecycle.rs`.
- `work_start` / `work_status` / `work_abandon` — `crates/stella-cli/src/self_driving_cmd/work.rs`.
- `deliver_open` / `deliver_observe` / `deliver_next` / `deliver_merge` — `crates/stella-cli/src/self_driving_cmd/deliver.rs`, over `stella_autonomy::deliver_next`'s pure machine.

## What blocks each one

Every verb above needs an **argument** and a **result** shape on the wire, and today `DriverCall` carries neither: a call is `{"call": "...", "id": N}` with nothing else. `crates/stella-plugin/src/driver.rs`'s header states the rule that governs this — a table lands with the verb that reads it, never ahead of it — and `DriverOk::backlog` is the one member that has landed under it.

That is the real cost of this issue, and it is why #6063 did one verb rather than five families: `backlog_claim` needs an issue key in, and who holds the lease out; `work_start` needs a key in and a worktree plus a stage out; `deliver_merge` needs a pull request number in and a merge state out. Each is a reviewable wire change with its own consent consequence, because the consent text a user reads renders the family sentence, not the argument table.

`DriverCapabilities::perform` also takes no arguments today (`crates/stella-runtime/src/wrapper/driver_call.rs`), so the host trait widens along with the wire.

## Files

[...]

PR #6170, opened 2026-09-06
feat(stella-autonomy,stella-cli): give the self-driving loop work when its queue runs dry

### LINKED ISSUE #6146: The self-driving loop stops when its queue runs dry instead of looking for more work
Phase B4 of #3599 (`doc:backlog-self-driving` §3.4). Nothing below is built on `main` (`9b941579`).

## The problem

When the ranked queue is dry, the self-driving loop has no way to find more work.

`stella_autonomy::step` already knows what should happen: `LoopStep::Sweep { lens }` is returned when the queue is empty and a lens is still open (`crates/stella-autonomy/src/step.rs`, `a_dry_queue_sweeps_the_open_lens_rather_than_stopping`). Nothing performs it. The driver stops instead, and says so in its own audit record:

```rust
LoopStep::Sweep { lens } => {
    audit::record(durable, Audit::SessionStopped, None,
        &format!("the machine asked to sweep `{lens}` and this build cannot — B4 builds it (#3599). \
                  Stopping rather than spinning on a step it cannot take."));
    return report(durable, &tally, &budget);
}
```
`crates/stella-cli/src/self_driving_cmd/drive.rs`

The driver channel is in the same state. `DriverCall::SweepAudit`, `SweepRegress` and `SweepMeta` are declared in `crates/stella-plugin/src/driver.rs` and granted in `plugins/stella-selfdriving/plugin.toml`, and all three answer `HostCallRefusal::Unsupported` from `crates/stella-cli/src/driver_plugin/capabilities.rs`.

Underneath that, `stella_autonomy::advance` walks the `LENSES` ladder once and returns `WATCH` forever after (`crates/stella-autonomy/src/lib.rs`). A lens that came back dry at one commit stays dry no matter how much the tree changes, so the loop cannot re-ask a question the code has since made worth asking again.

## What to build

Three separate things, in this order:

1. **Re-arm a dry lens when the baseline moves.** A lens that found nothing at commit X should re-open once the tree has changed enough to be worth re-asking. `advance` and the aperture state in `crates/stella-cli/src/self_driving_cmd/state.rs` are where this lives.
2. **`sweep regress`** — re-check closed-issue receipts, so a defect that was fixed and came back is found again.
3. **`sweep meta`** — the loop looking at its own ledger for what it keeps getting wrong.

## Constraints already known

- **Do not touch `finding_digest`'s normalization.** It is a byte-for-byte contract with every existing `seen.txt` on every machine that has run the loop (`crates/stella-cli/src/self_driving_cmd/state.rs`, `the_seen_set_survives_the_move_byte_for_byte`). The global dedup set is exactly what makes re-passing a lens safe — re-arming a lens is only cheap because a repeat finding is dropped by digest. Changing the digest re-files every finding the loop has ever filed.
- **Ship it behind a per-supply switch that defaults to queue-only.** #3599's own "risk this does not solve" section is backlog inflation: an infinite work supply is not an infinite *valuable* work supply. `crates/stella-cli/src/self_driving_cmd/config.rs` is where `residue_gate` lives and is the shape to copy.
- `crates/stella-autonomy` has no god files and must keep none — new pure machines are new modules, not additions to `lib.rs`.
- `crates/stella-cli/src/self_driving_cmd.rs` is closed to growth; new logic goes in a sibling under `self_driving_cmd/`.
- Serving these three verbs over the driver channel needs the argument and result shapes #6118 describes. That work can land first or alongside; this issue is the behaviour, #6118 is the wire.

## Definition of done

- [x] A lens that returned nothing at one commit re-opens after the baseline moves, with a test that pins what "moved enough" means.
- [x] That re-opened lens yields only digests absent from `seen.txt` — the witness #3599 names for this phase.
- [x] `sweep regress` re-checks closed-issue receipts and files what came back.
- [x] `sweep meta` reads the loop's own ledger and produces findings from it.
- [x] Each supply has its own switch and every one of them is off by default, so an upgrade changes no running loop's behaviour.
- [x] `finding_digest` is byte-for-byte unchanged, proven by the existing seen-set test still passing untouched.
- [x] `LoopStep::Sweep` no longer stops the driver in `drive.rs`.

Verified against #6170 — evidence, and the one defect found and fixed while
checking, are in
https://github.com/macanderson/stella/issues/6146#issuecomment-5555901862.

macanderson:
## What and why

`stella_autonomy::step` already returns `LoopStep::Sweep { lens }` when the
ranked queue is empty and a lens is still open. Nothing performed it. `drive.rs`
recorded that this build could not sweep and returned — and `state.lens` was
never set from anything, so the step was unreachable in the driver at all.

This builds the three supplies `doc:backlog-self-driving` §4 names, each behind
its own switch, and every switch off by default (`[self_driving.supply]`).

**`rearm`** (§4.2) — a dry ladder re-opens at `rubric` once the base branch has
moved past a declared distance since the sweep that exhausted it: 50 commits or
30 days, either bound alone. The distance is read from `last_clean_head` (already
written) plus a new `last_clean_at` beside it. Held shut while the loop's own
ledger raises `NOISY`, which is §4.4's "narrow rather than widen" made real
rather than reported.

**`regress`** (§4.3) — as an issue closes, `apply_closure` writes a
`ClosureReceipt` to `receipts.jsonl`: what the closure cited, and whether that
change was on the base branch at the time (a commit by `merge-base
--is-ancestor`, a pull request by the `(#N)` a squash merge leaves in the
subject). The sweep asks the same question again. **Only a green→red counts**:
a change that was there and is gone is a fact and is filed; a change nobody
could see at close time is counted as unsweepable and reported, never filed.
That is what stops the sweep reporting *gone* for something that was never
there — the one false positive class this design could have had.

**`meta`** (§4.1) — the ledger read against the loop itself: one defect per
raised signal (`STUCK`/`NOISY`/`FRAGILE`/`STARVED`, which today are printed to
nobody), and one per lens that has run at least three cycles and never produced
a finding, because a lens whose tool fails reports the same nothing as a lens
whose answer is nothing.

Everything files through `backlog::file_finding`, so the digest dedup and the
workspace convention check both hold. **`finding_digest` is byte-for-byte
untouched** — every existing `seen.txt` keeps deduplicating, and the existing
`the_seen_set_survives_the_move_byte_for_byte` test is unchanged.

A sweep that files nothing shuts the supply for the rest of the run, so the
driver falls through to `Watch` instead of asking for the same step forever.

## Witness mechanism

Every witness below **fails on `main` by construction**: the modules, the
config section and the receipt file do not exist there, so the tests do not
compile against the old tree.

| Witness | What it pins |
|---|---|
| `supply::tests::every_supply_is_shut_by_default` | an upgrade changes no running loop |
| `supply::tests::a_dry_ladder_reopens_once_the_base_has_moved_far_enough` | what "moved enough" means, at the boundary (49 holds, 50 re-opens) |
| `supply::tests::enough_days_reopen_the_ladder_on_their_own` | the second bound |
| `supply::tests::an_open_lens_is_not_rearmed` / `a_noisy_loop_does_not_widen` / `an_unknown_baseline_holds_the_ladder_shut` | the three holds |
| `supply::tests::a_reopened_sweep_yields_only_digests_absent_from_the_seen_set` | **the yield witness this phase names** |
| `supply::tests::a_shifted_line_number_is_the_same_finding` | the digest normalization the re-pass rests on |
| `regress::tests::a_fix_that_has_left_the_base_is_filed_again` | the regression finding |
| `regress::tests::a_change_unseen_at_close_is_never_reported_as_gone` | the false-positive class is closed |
| `regress::tests::a_closure_with_no_change_cited_is_counted_not_filed` | the unsweepable count is published |
| `regress::tests::an_older_receipt_line_still_parses` | a line from an older build reads as unsweepable, not as gone |
| `meta::tests::a_lens_that_never_finds_anything_becomes_a_finding` / `a_raised_signal_becomes_a_finding` | the meta supply |
| `self_driving_cmd::supply::tests::a_default_workspace_opens_no_supply` | the shipped default, read through `LoopConfig` |
| `self_driving_cmd::supply::tests::the_document_switches_reach_the_policy` | `[self_driving.supply]` reaches the machine |

## Exemplar

`stella-autonomy`'s own `closure.rs` / `deliver.rs` shape: a closed set of
outcomes plus a pure decision function, with the I/O as a caller-supplied
parameter (`sweep`'s `present_now`) rather than a dependency — the same seam
`stella-core`'s ports use.

## Scope

The driver channel's `DriverCall::SweepAudit` / `SweepRegress` / `SweepMeta`
still answer `Unsupported`. Serving these three over the wire needs the argument
and result shapes #6118 describes; this issue is the behaviour, that one is the
wire.

Tests deleted: None

## Residue filed

- #6182 — the driver never runs the lens it re-armed. The aperture is durable
  and the audit phase does run it; doing it inside one `drive` run needs a model
  seat for each lens's `interpret` step, which is bigger than this change.
- #6183 — `SweepAudit` / `SweepRegress` / `SweepMeta` on the driver channel still
  answer `Unsupported`.
- #6184 — no hand-run `stella self-driving sweep` verb, so a supply cannot be
  tried before its switch is turned on.

Closes #6146

## Summary by Sourcery

Enable the self-driving loop to draw useful work from configurable non-draining supplies when its ranked queue is empty.

New Features:
- Add configurable work supplies for re-arming dry lens sweeps, detecting regressions in previously closed fixes, and identifying issues in the loop's own ledger.
- Allow self-driving sweeps to execute when the ranked queue is empty and file novel findings through the existing backlog workflow.

Bug Fixes:
- Prevent repeated empty sweep steps from causing the driver to loop indefinitely by disabling the supply for the remainder of the run when no new finding is filed.
- Ensure regression checks only report changes that were present at closure and are now absent, avoiding false positives for changes that were never observed.

Enhancements:
- Persist closure receipts and clean-sweep timestamps to support later regression and re-arm decisions.
- Keep all new supplies disabled by default while supporting independent configuration of their switches and re-arm thresholds.

Documentation:
- Document the new self-driving supply configuration and its re-arm thresholds.
- Describe the new autonomy modules and supply behavior in the autonomy README.

Tests:
- Add coverage for supply defaults, re-arm thresholds and safeguards, regression classification, receipt compatibility, meta findings, deduplication, and configuration propagation.

### DIFF
--- a/crates/stella-autonomy/README.md
+++ b/crates/stella-autonomy/README.md
@@ -86,6 +86,9 @@ the CLI does — so the dashboard and the terminal cannot disagree.
 | [`src/ready.rs`](src/ready.rs) | `ready_queue`: which backlog issues may be taken next, and in what order. `Blocked by: #N` lines, the `status:ready` override, and the escalation cooldown. |
 | [`src/escalation.rs`](src/escalation.rs) | Escalation as a cooldown: `classify` reads an abort message, `retry_after` says how long to wait, `park_after` says when waiting ends. The record is written into the issue body and read back from it. |
 | [`src/gate.rs`](src/gate.rs) | Which checks are allowed to block a merge, and which have stopped earning that right. |
+| [`src/supply.rs`](src/supply.rs) | Where the loop looks when the ranked queue is empty: `WorkSupply`, `SupplyPolicy` (every switch off), the `rearm` rule that re-opens a dry ladder against a moved base, and `novel`, which drops a finding the seen set already holds. |
+| [`src/regress.rs`](src/regress.rs) | Re-checking the fixes the loop has claimed: `ClosureReceipt`, `check`, and the `sweep` that turns a change gone from the base branch into a fresh defect. |
+| [`src/meta.rs`](src/meta.rs) | The loop read against its own ledger: a raised signal nobody acted on, and a lens that has looked several times and found nothing. |
 | [`src/closure.rs`](src/closure.rs) | How an issue ends, and what has to be true first. |
 | [`src/convention.rs`](src/convention.rs) | How this repository writes issues, learned rather than assumed. |
 | [`src/attribution.rs`](src/attribution.rs) | Attribution lines for filed issues and opened PRs. |
--- a/crates/stella-autonomy/src/lib.rs
+++ b/crates/stella-autonomy/src/lib.rs
@@ -58,10 +58,13 @@ mod deliver;
 mod doctrine;
 pub mod escalation;
 pub mod gate;
+pub mod meta;
 pub mod priority;
 pub mod ready;
+pub mod regress;
 mod stats;
 mod step;
+pub mod supply;
 mod surface;
 
 use std::fmt::Write as _;
--- a/crates/stella-autonomy/src/meta.rs
+++ b/crates/stella-autonomy/src/meta.rs
@@ -1,0 +1,243 @@
+//! What the loop keeps getting wrong, read from its own ledger.
+//!
+//! Every cycle appends a row. The rows say how much was fixed, how much was
+//! filed, whether the gate was red, and which lens was open. Read together
+//! they show habits that no single cycle shows.
+//!
+//! This module turns those habits into defects the loop files against itself.
+//! Two kinds:
+//!
+//! - A named signal the fold already raises — `STUCK`, `NOISY`, `FRAGILE`,
+//!   `STARVED`. Each of these is reported and acted on by nobody, which is
+//!   what a signal with no reader always becomes.
+//! - A lens that has looked several times and found nothing at all. Its
+//!   backing is probably not running.
+//!
+//! Each finding has a fixed title, so the dedup key holds. A habit is filed
+//! once, not once per cycle.
+//!
+//! `doc:backlog-self-driving` §4.1 is the design.
+
+use std::collections::BTreeMap;
+
+use crate::supply::Finding;
+use crate::{Calibration, CycleRecord, Metrics, Signal, metrics, starved};
+
+/// The label a finding about the loop carries.
+///
+/// The loop wasting its own cycles is a defect, and the type axis of a backlog
+/// convention spells that `bug`.
+pub const DEFECT_LABEL: &str = "bug";
+
+/// How many cycles a lens must have run before its silence counts.
+///
+/// Three, matching the dry-streak the ladder already uses. One quiet cycle is
+/// a clean tree; three in a row on one lens is a tool that is not looking.
+pub const BARREN_CYCLES: u64 = 3;
+
+/// The defects the ledger shows about the loop itself.
+#[must_use]
+pub fn sweep(rows: &[CycleRecord], cal: &Calibration) -> Vec<Finding> {
+    let folded = metrics(rows);
+    let controller = starved(cal);
+
+    let mut out: Vec<Finding> = folded
+        .signals
+        .iter()
+        .chain(controller.iter())
+        .map(|signal| signal_finding(signal, &folded))
+        .collect();
+
+    for lens in barren_lenses(rows) {
+        out.push(barren_finding(&lens));
+    }
+    out
+}
+
+/// One defect for one raised signal.
+fn signal_finding(signal: &Signal, folded: &Metrics) -> Finding {
+    let Signal { code, text } = signal;
+    let Metrics {
+        cycles,
+        fixed,
+        filed,
+        new_findings,
+        zero_fix_cycles,
+        red_gate_cycles,
+        ..
+    } = folded;
+    Finding {
+        title: format!("the self-driving loop raises {code} against its own ledger"),
+        body: format!(
+            "The loop's ledger raises `{code}`: {text}.\n\n\
+             Over {cycles} cycle(s) it fixed {fixed}, filed {filed}, and found \
+             {new_findings} new defect(s). {zero_fix_cycles} cycle(s) fixed nothing and \
+             {red_gate_cycles} ended on a red gate.\n\n\
+             A signal nobody acts on trains people to skip signals, so this one is filed \
+             as work rather than printed again.\n\n\
+             ## How to check\n\n\
+             1. Read `ledger.jsonl` in this repository's self-driving state directory.\n\
+             2. Find which cycles raise `{code}` and what they have in common.\n\
+             3. Fix the cause, or change the rule that raises it.\n\n\
+             ## Definition of done\n\n\
+             - [ ] The cause of `{code}` is named, with the ledger rows that show it.\n\
+             - [ ] Either the loop stops meeting it, or the rule that raises it changes.\n"
+        ),
+        labels: vec![DEFECT_LABEL.to_owned()],
+    }
+}
+
+/// One defect for one lens that never finds anything.
+fn barren_finding(lens: &str) -> Finding {
+    Finding {
+        title: format!("the `{lens}` lens has looked several times and found nothing"),
+        body: format!(
+            "The `{lens}` rung of the aperture ladder has run at least {BARREN_CYCLES} \
+             cycles and has never produced a new finding.\n\n\
+             That is one of two things. Either the tree is clean on this question, or the \
+             backing the lens declares is not running. The second is far more likely, and \
+             it is silent: a lens whose tool fails reports the same nothing as a lens whose \
+             answer is nothing.\n\n\
+             ## How to check\n\n\
+             1. Read the backing `{lens}` declares in the aperture list \
+             (`stella self-driving aperture --list`).\n\
+             2. Run that tool by hand against a clean tree.\n\
+             3. If it runs and finds nothing, say so on this issue and close it. If it \
+             fails, fix the tool or the declaration.\n\n\
+             ## Definition of done\n\n\
+             - [ ] The `{lens}` backing has been run by hand and its output read.\n\
+             - [ ] Either the tool is fixed, or this issue records that the silence is \
+             real.\n"
+        ),
+        labels: vec![DEFECT_LABEL.to_owned()],
+    }
+}
+
+/// Lenses that have run enough cycles to be judged and found nothing at all.
+fn barren_lenses(rows: &[CycleRecord]) -> Vec<String> {
+    let mut tally: BTreeMap<&str, (u64, u64)> = BTreeMap::new();
+    for row in rows {
+        let name = row.aperture.trim();
+        if name.is_empty() || name == crate::WATCH {
+            continue;
+        }
+        let entry = tally.entry(name).or_insert((0, 0));
+        entry.0 += 1;
+        entry.1 += row.new_findings;
+    }
+    tally
+        .into_iter()
+        .filter(|(_, (cycles, found))| *cycles >= BARREN_CYCLES && *found == 0)
+        .map(|(name, _)| name.to_owned())
+        .collect()
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::AimdLimits;
+
+    fn row(aperture: &str, fixed: u64, filed: u64, found: u64, gate: &str) -> CycleRecord {
+        CycleRecord {
+            cycle: 1,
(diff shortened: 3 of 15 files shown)


macanderson:
## Sourcery's ❌ on objective 1 holds — checked against the diff

> The PR implements the re-arm decision and the `novel` seen-set filtering, but the driver's `supply::pass` never executes the reopened lens or obtains findings from it.

That reads correctly to me. In `crates/stella-cli/src/self_driving_cmd/supply.rs`, `pass` takes `lens: &str` and the only use of it in the body is the "found nothing" audit string. `findings` is built from `stella_autonomy::regress::sweep` and `stella_autonomy::meta::sweep`; nothing draws from the lens itself, so `novel()` filters regress/meta output and never lens output.

The rearm is not inert, which the review row does not say: `open_lens`'s `Rearm::Reopen { lens }` arm calls `durable.set_aperture(lens)`, so the ladder does move and the ordinary audit cycle would draw at that aperture on a later pass. The gap is narrower than "rearming does nothing" and still real — **within a sweep, a rearmed lens contributes no findings, and there is no witness for the one DoD item that asks it to**:

- [ ] That re-opened lens yields only digests absent from `seen.txt` — the witness #3599 names for this phase.

Two ways to settle it, author's call:

1. **Execute the reopened lens inside `pass`** and extend `findings` with its output before the `novel()` filter — the reading the DoD item most directly supports, and the one that makes the `lens` parameter load-bearing.
2. **Decide the rearm is deliberately aperture-only** — the ordinary cycle draws that lens on the next pass — say so in `pass`'s doc comment, and split the witness into a follow-up issue under SCR-004.

Either the `lens` parameter stops being audit-string-only, or the objective stops claiming it. Flagging rather than pushing a fix: which of the two is right is a design decision about where a sweep's work comes from, not a small in-scope repair, so it belongs to you.

The other two objectives are ✅ and I found nothing to add to them. Nothing else on this PR is blocking from my side — CI was still in flight when I looked; no merge conflict against `main`.

---

macanderson:
## Sourcery's ❌ on the rearm objective is correct — verified against `01f6672`

Read the code rather than the review, and the finding holds.

`supply::pass` takes `lens: &str` and reads it in one place: the `format!` inside the empty-`fresh` audit record. The findings it collects come from `stella_autonomy::regress::sweep` (behind `cfg.supply.regress`) and `stella_autonomy::meta::sweep` (behind `cfg.supply.meta`). Nothing runs the re-opened lens.

So the chain ends where it starts. `supply::open_lens` resolves `Rearm::Reopen { lens }`, sets the aperture, and returns the name; `drive.rs`'s `LoopStep::Sweep` arm hands it to `pass`, where it decorates a log line. With `rearm` on and `regress`/`meta` off, a re-armed ladder collects zero findings, files nothing, `pass` returns `false`, and the `Sweep` arm clears `state.lens` — shutting the supply for the rest of the run on its first pass.

### What that leaves unmet on #6146

- *"A lens that returned nothing at one commit re-opens after the baseline moves, with a test that pins what 'moved enough' means."* — the **decision** is implemented and well covered (`a_dry_ladder_reopens_once_the_base_has_moved_far_enough` pins 49 holds / 50 re-opens). The re-opened lens then does no work.
- *"That re-opened lens yields only digests absent from `seen.txt`."* — `a_reopened_sweep_yields_only_digests_absent_from_the_seen_set` calls `novel` over two hand-built findings and a seen digest. It pins the dedup filter over an offered slice; no lens sweep is in it.

`regress` and `meta` are unaffected — both have real producers, both file, and their witnesses drive them.

### Two ways forward

1. Run the lens inside `pass` so the rearm supply produces findings, which finishes #6146 as written.
2. Ship the two working supplies, downgrade `Closes #6146` to `Refs #6146` per SCR-003, and file the lens-execution wiring as its own issue — the shape #6117 took for its unmet sixth box.

I have not edited the PR or the issue. A commit landed on this branch while I was reading it, so this is a report into an active session rather than a change to it.

---

macanderson:
Automated PR sweep. Sourcery's ❌ against #6146 is unanswered, and checking it against the code turns up a real disagreement between a ticked DoD box and what the branch does. Flagging, not pushing — the resolution is a scope call that belongs to you.

## What I checked

Sourcery says the re-armed lens is never executed. That is accurate as far as it goes: in `self_driving_cmd/supply.rs`, `pass()` takes `lens: &str` and the binding reaches exactly one place in the body —

```rust
"the `{lens}` sweep found nothing the seen set does not already hold ({} finding(s) offered)"
```

an audit string. `findings` is filled by `regress::sweep` and `meta::sweep` alone, then filtered through `supply::novel(&findings, &seen)`. There is no path that runs the lens for findings. `pass()` has one caller (`drive.rs:999`) and it passes the name through from `open_lens`.

But Sourcery reads this as an oversight, and it is not. `open_lens` genuinely re-arms — `stella_autonomy::supply::rearm` decides, `durable.set_aperture(lens)` persists the rung — and your comment beneath it says the design out loud:

```rust
// Still a supply to draw from: `regress` and `meta` do not need a
// lens. The name is what the machine reports, not what it draws from.
```

So the mechanism is deliberate and documented. That is worth saying plainly, because "the driver ignores the lens" makes it sound like a slip.

## The actual gap

#6146's second DoD box is ticked:

> - [x] That re-opened lens **yields** only digests absent from `seen.txt` — the witness #3599 names for this phase.

The re-opened lens yields nothing. `regress` and `meta` yield; `novel` filters *their* digests against the seen set. The seen-set half of that sentence is implemented and witnessed; the "re-opened lens yields" half is not, by design.

Box 1 ("a lens ... re-opens after the baseline moves, with a test that pins what 'moved enough' means") **is** met — `rearm` plus its test.

So one ticked box asserts something the branch does not do. Two ways out, both yours:

1. **The item was over-specified** — the supplies are `regress` and `meta`, and the lens is the aperture rung, not a third producer. Then reword box 2 to what `novel` actually guarantees and file the lens-execution half as its own issue (`triage` only, SCR-004).
2. **Lens execution is genuinely still missing**, and this PR is `Refs #6146` rather than `Closes` until it lands.

Either way it wants an answer on the thread before merge — CLAUDE.md's rule is that a ❌ gets a fix or a rebuttal, and right now this one has neither. `dod / dod` is green on the ticked boxes, so nothing else will catch it.

## Rest of the sweep on this PR

CI is clean on `01f66727` and there is no merge conflict. Boxes 3–7 of #6146 I did not re-verify.

---

macanderson:
## Answering the one ❌

> The PR implements the re-arm decision and the `novel` seen-set filtering, but
> the driver's `supply::pass` never executes the reopened lens or obtains
> findings from it.

Correct about the code, and the row is worth keeping open as a description of
where this stops. Two things it does not account for, and then what I filed.

**The re-opened lens is executed — by the audit phase, not by `drive`.** The
aperture is durable state (`.../aperture`), and re-arming rewrites it. The phase
that runs a lens's tooling reads it back through `stella self-driving aperture
--current` and files through `backlog::file_finding`, which is the same door
`supply::pass` uses and the same digest dedup. So a re-armed ladder does produce
findings, and those findings are exactly the digests absent from `seen.txt`.
What is missing is the driver doing it inside one `drive` run.

**Doing it inside the driver needs a model seat this build has none of.** Of the
nine rungs in `stella_autonomy::LENSES`, four carry `Tooling::ModelOnly` and
have no command to run. The other five carry `Tooling::Command`, and each one's
`interpret` field describes a judgement a model makes over the output — "the
report's findings, deduped by digest against seen.txt", "each listed file that
holds pure decision logic and has no property test is a finding". Reading
`/ultraudit` or `make invariants` output mechanically and filing the result is
how this supply becomes the make-work machine `doc:backlog-self-driving` §4.4
warns it can become, so I would rather it refuse than guess.

**Filed rather than left**: #6182 carries it, with both candidate routes written
out (a verifier-tier seat for the interpret step, or a per-lens mechanical
reader for the lenses whose output can be read without one) and a witness that
fails until one is built.

## The rest of the residue

- #6183 — `SweepAudit` / `SweepRegress` / `SweepMeta` on the driver channel still
  answer `Unsupported`; two of the three now have behaviour to bind to.
- #6184 — no `stella self-driving sweep` verb, so nobody can try a supply by hand
  before turning its switch on, and `surface_drift` cannot see the gap.

PR #6171, opened 2026-09-06
feat(stella-parity): say what each turn lane binds, seam by seam

macanderson:
## What

`crates/stella-parity/src/lane.rs` said who stamps a lane name. It said
nothing about what a lane binds, so the answer stayed spread across three
struct literals in three crates: `stella-cli`'s `lane_capabilities.rs`,
`stella-core`'s `subagent.rs`, and `stella-serve`'s `served_capabilities`.

`crates/stella-parity/src/lane/capability.rs` is the missing table. One row
per lane and seam — all seven builtin lanes against all eleven slots on
`TurnCapabilities`. A row carries three columns the old table had none of:

- **What the lane asked for.** `Bound { binds, witness }`,
  `Declined { reason }`, or `Deferred { issue, waiting_on }`. A seam a lane
  refuses and a seam nobody has reached used to read as the same bare
  `None`.
- **What it got.** `AsRequested` or `Withheld { authority, reason }`. Every
  lane here is `AsRequested`, because it compiles its own literal. A lane an
  installed plugin brings will not be, and the column is there so the gate
  that cuts one down needs no schema change.
- **Where the lane came from.** `LaneOrigin::Builtin` or `Plugin`, so a
  plugin row can say what it is worth.

## Why the rows are trustworthy

Four checks, all in `lane/capability/tests.rs`:

1. **The macro is the totality gate.** `lane_capabilities!` spells all
   eleven seam names in its pattern, so a lane block that skips one does not
   match and does not compile.
2. **`every_capability_slot_is_a_seam` destructures `TurnCapabilities` with
   no rest arm.** A slot added to the engine stops this crate's tests
   building. The repair needs a new `Seam` case, which needs a line in the
   macro pattern, which then stops every lane block matching until each lane
   answers. That chain is what the issue asked for.
3. **`a_seam_claim_agrees_with_its_lane_literal` reads each claim back
   against the source.** It slices the lane's `TurnCapabilities { .. }`
   literal by brace matching from a declared anchor, then asserts a `Bound`
   row's `binds` text is really there and that a `Declined` or `Deferred`
   row's field is really written `<field>: None`.
4. **`every_named_seam_witness_exists`** checks every named test exists in
   the swept sources, as `every_lane_witness_exists` already did for the
   producer rows.

## Witness mechanism

The tests do not exist on `main`, so they fail there by absence. The one
with a real negative control is check 3, and I ran it: flipping
`ServeSession`'s `bus` row from `Bound` to `Declined` and leaving the code
alone gives

[...]

Issue #6158, opened 2026-09-06
Three turn lanes still show "undecided" for mid-turn context refresh


macanderson:
## What's wrong

Stella's engine can ask again for fresh context partway through a long turn (skills, memories, matched rules). This is the context re-query plane (`stella_core::ports::SteeringRequery`). Three lanes that run long turns do not use it, and the tracking table (`crates/stella-parity/src/lane/capability.rs`) marks all three `Deferred`, citing this issue:

- `sub_session` (deck worker lane) — `crates/stella-cli/src/lane_capabilities.rs`
- `fleet_attempt` (one fleet run) — `crates/stella-cli/src/lane_capabilities.rs`
- `goal_arc` (a `/goal` run) — `crates/stella-cli/src/lane_capabilities.rs`

`sub_session` already has its own, more detailed open issue: #6122 (needs a decision: bind it, or measure and write down why not).

**`served_capabilities` is done and has left this list.** It was the fourth lane here. #6172 gave `served_capabilities` a `requery` parameter, so `crates/stella-serve/src/session.rs` writes `requery,` as field shorthand and the lane really does bind the plane; the `ServeSession` row followed in #6203 (`b09fa68`) and now reads `SeamClaim::bound("requery,", SERVE_SEAMS)`. `a_served_turn_binds_what_its_call_site_bound_before` asserts `requery` in both directions, so the row is witnessed rather than falling back to `Witness::Literal`, and `UNWITNESSED_SEAMS` is unchanged at 5. #6121 is the issue that tracked that work.

**Correction to the original write-up:** `fleet_attempt` looked unwired too, but it is not. `crates/stella-cli/src/fleet_cmd.rs` (right after `Engine::assemble(..., lane_capabilities::fleet_attempt(...))`) already attaches the re-query plane, with a comment explaining why it rides there instead of inside `fleet_attempt`'s own capability list. The lane genuinely binds the plane at runtime — only the tracking row is stale. It should say `Bound`, citing the fleet witness test added in #6059 (`a_fleet_worker_requeries_the_skill_its_opening_prompt_could_not`), not `Deferred`.

`goal_arc` is the one lane nobody has looked at yet. `crates/stella-cli/src/agent/goal.rs` already opens a `SessionMemory` for a goal run (see `run_goal`'s setup, near the `lane_capabilities::goal_arc(...)` call) — so the plane it would query is available. It is just never attached.

## What to do

[...]

Issue #6150, opened 2026-09-06
The lane matrix records who names a lane, not what a lane binds


macanderson:
## Problem

`crates/stella-parity/src/lane.rs` names itself the turn-lane matrix, but it
answers a narrower question than #3274 slice 4 asked for. Today each row is
one `BuiltinLane` and one `LaneBinding::Bound { site, how, witness }` (or
`NoProducer { reason }`) — it records **who stamps a lane name**. It does not
record **what a lane binds**.

So the thing the epic wanted a reader to be able to look up — which of the
engine's optional seams each lane has, why the unbound ones are unbound, and
which are waiting on something — is still spread across the struct literals in
`crates/stella-cli/src/lane_capabilities.rs`, `crates/stella-core/src/subagent.rs`
and `crates/stella-serve/src/session.rs`. Those literals are complete and
compile-enforced, which is the hard half; what is missing is one table a person
can read and a test can hold to a witness.

Four specific gaps against the slice as written:

- **No per-capability rows.** `git grep TurnCapabilities crates/stella-parity/`
  hits only comments in `lib.rs`. Nothing in the crate destructures
  `TurnCapabilities`, so a new slot on that struct fails to compile at the
  assembly sites and passes here in silence.
- **No `Declined { reason }` or `Deferred { issue, waiting_on }`.** A seam a
  lane deliberately does not bind and a seam nobody has got to yet render
  identically — the exact failure the epic opens with.
- **`requested` and `granted` are not separate columns.** For a builtin lane
  they are the same value; for a plugin lane they will not be. The epic asked
  for both from day one so the enforcement point has somewhere to write its
  answer without a schema change. #2716 (the authority vocabulary) is now
  closed, so there is something real to point `granted` at.
- **No origin on a row**, so a future plugin row cannot say what it is worth.

## Files

- `crates/stella-parity/src/lane.rs` — the matrix to extend.
- `crates/stella-parity/src/evolution.rs` and `crates/stella-parity/src/lib.rs`
  — the two matrices beside it. Copy their shape: a macro table generating the
  rows, enforced from both sides.
- `crates/stella-core/src/driver/capabilities.rs` — `TurnCapabilities`, the
  eleven slots a row has to claim.
- `crates/stella-cli/src/lane_capabilities.rs`,
  `crates/stella-core/src/subagent.rs` (`Engine::assemble` call),
  `crates/stella-serve/src/session.rs` (`served_capabilities`) — the literals a
  row's claim must agree with.
[...]

PR #6172, opened 2026-09-06
feat(stella-serve): let a served turn re-ask for context mid-turn

### LINKED ISSUE #6121: Served turns never re-ask for context mid-turn

macanderson:
## What and why

Mid-turn context re-query worked on `stella run`, the deck and a fleet
worker, and not over the wire. A served turn picked its context from its
opening prompt and never looked again, so a long turn that drifted onto
new files kept working from what the first sentence asked for.
`stella-serve` assembled no re-query source at all — `served_capabilities`
passed `requery: None`.

**A fourth reverse request.** A re-query fans out over recall and costs the
host money, and this server holds no rights of its own, so the ask travels
the way the model call and the tool call already do.

- `ServerFrame::RequeryRequest { request_id, signal }` joins the outbound
  union. `signal` is `RequerySignal`, an owned copy of what the engine
  assembled at the boundary.
- `RequeryResultIn { request_id, context }` answers it, over
  `POST /v1/turns/{id}/requery-result`. `null` is the ordinary answer.
- `crates/stella-serve/src/remote/requery.rs` is new: `RemoteRequery`
  implements `stella_core::ports::SteeringRequery` and is bound in
  `served_capabilities`. The route handler is
  `crates/stella-serve/src/routes/requery.rs`, its own file because
  `routes.rs` is the closest file in this crate to the 1500-line cap.

**Off by default.** `SessionSpec::steering_requery`, on the wire as
`steering_requery` on `POST /v1/turns` and `POST /v1/sessions/{id}/turns`.
A host that has not built the answering route cannot answer, and an
unanswered ask would park the step for the whole reverse-request deadline
at every boundary. A turn that does not opt in emits no such frame and
behaves exactly as before.

**Hysteresis and dedup stay server-side**, mirroring the CLI's
`SessionRequery` (`crates/stella-cli/src/memory/steering.rs`, the exemplar
this follows): a fingerprint over touched paths and error classes, a
two-step minimum spacing, and a produced-block set seeded from the recall
blocks the turn opened with. Per-handle dedup stays on the CLI side, since
a handle names something in a workspace this server never sees.

**An unanswered ask is not a failed turn.** A timeout, a dropped host or a
closed frame sink all answer `None`, and the step runs on with the context
it had.

**The marker is added here.** The engine injects the block as sent, and
`turn_start_index` tells injected context from a user turn by
`RECALL_MARKER` alone, so `RemoteRequery` prefixes it when the host's text
does not carry one.

## Witness mechanism

`a_served_turn_requeries_the_steering_plane_after_its_work_moves_to_a_new_path`
(`crates/stella-serve/tests/bridge.rs`) drives a three-step turn whose first
two steps touch `src/adapter.rs`. At the third boundary the signal has
drifted and the spacing is met, the server emits `requery_request`, the mock
host answers with a block, and the test asserts the block reached the next
model call's messages. It fails on `main` by construction, twice: the frame
type does not exist there and `served_capabilities` passes `requery: None`.

Two more pin the rest: `a_served_requery_leaves_the_byte_stable_prefix_untouched`
asserts the system message is byte-identical and the block rides the
volatile tail as a marked user message, and
`a_host_that_declines_a_requery_injects_nothing` is the negative control —
the same turn asks and injects nothing when the host answers `null`.

`crates/stella-parity/src/lib.rs`'s `turn.steering_requery` row moves its
`api` posture from `Deferred` to `Shipped`, naming that witness and the new
route.

## Wire contract

`docs/wire/serveframe.schema.json`, `serveinbound.schema.json` and
`serveframe.d.ts` carry the new frame arm, `RequerySignal` and
`RequeryResultIn`. `docs/spec/serve-surface.md` gains the route row.

Tests deleted: None

Closes #6121

## Summary by Sourcery

Enable opted-in served turns to request fresh context from their host when mid-turn work drifts, without failing turns when no context is returned.

New Features:
- Add opt-in mid-turn context re-query support for served turns through outbound `requery_request` frames and the `requery-result` response endpoint.
- Allow stateless and session turns to enable steering re-querying while preserving existing default behavior.

Bug Fixes:
- Allow served turns to refresh context when their work drifts beyond the opening prompt instead of relying on stale initial context.

Enhancements:
- Keep re-query throttling, drift detection, deduplication, cancellation, timeout, and graceful no-context handling within the serving layer.
- Preserve byte-stable prompt prefixes while injecting returned context into the volatile conversation tail.

Documentation:
- Publish the re-query wire types, JSON schemas, TypeScript declarations, and serving-surface documentation.

Tests:
- Add bridge coverage for successful context re-query, prefix stability, and declined re-queries.
- Promote served steering re-query parity coverage from deferred to shipped.

Chores:
- Extend reverse-request routing, pending reply handling, observability, and metrics for re-query requests.

### DIFF
--- a/crates/stella-parity/src/lib.rs
+++ b/crates/stella-parity/src/lib.rs
@@ -294,11 +294,15 @@ pub static CAPABILITIES: &[Capability] = &[
                         a_fleet_worker_requeries_the_skill_its_opening_prompt_could_not",
             witness: "a_drifted_turn_recalls_the_skill_its_prompt_could_not",
         },
-        api: SurfacePosture::Deferred {
-            waiting_on: "a serve-side SteeringRequery implementor: the engine port exists, but \
-                         the serve stack assembles no requery source, so remote turns still \
-                         select context against the opening prompt alone — #6121 is where \
-                         that is being decided",
+        api: SurfacePosture::Shipped {
+            mechanism: "a turn created with `steering_requery: true` binds stella-serve's \
+                        RemoteRequery, which raises a `requery_request` frame at a drifted \
+                        step boundary and parks on the host's answer to \
+                        POST /v1/turns/{id}/requery-result. Hysteresis and block dedup stay \
+                        server-side, mirroring the CLI's SessionRequery; an unanswered ask \
+                        injects nothing rather than failing the turn. Off by default, because \
+                        a host that has not built the route cannot answer",
+            witness: "a_served_turn_requeries_the_steering_plane_after_its_work_moves_to_a_new_path",
         },
     },
     Capability {
--- a/crates/stella-serve/src/backlog.rs
+++ b/crates/stella-serve/src/backlog.rs
@@ -143,6 +143,8 @@ fn droppable(frame: &ServerFrame) -> bool {
         ServerFrame::Event { .. } => true,
         ServerFrame::ToolRequest { .. }
         | ServerFrame::ProviderRequest { .. }
+        // A dropped re-query parks the step until its deadline.
+        | ServerFrame::RequeryRequest { .. }
         | ServerFrame::TurnHeld { .. }
         | ServerFrame::TurnReleased
         | ServerFrame::TurnComplete { .. } => false,
--- a/crates/stella-serve/src/frame.rs
+++ b/crates/stella-serve/src/frame.rs
@@ -66,6 +66,18 @@ pub enum ServerFrame {
         role: ModelCallRole,
         request: CompletionRequest,
     },
+    /// The engine's work has moved on from the prompt it opened with, and it
+    /// is asking the host for whatever context now fits. The host answers
+    /// with [`RequeryResultIn`] keyed by `request_id`; the engine step that
+    /// raised it is parked until then.
+    ///
+    /// Only ever emitted when the turn asked for it
+    /// (`steering_requery` on `POST /v1/turns`). A host that has not built the
+    /// answering route never sees this frame.
+    RequeryRequest {
+        request_id: String,
+        signal: RequerySignal,
+    },
     /// The turn reached a step boundary and is **holding** there, because
     /// `POST /v1/turns/{id}/pause` asked it to (#932).
     ///
@@ -95,6 +107,54 @@ pub enum ServerFrame {
     TurnReleased,
     /// Terminal frame: the turn ended. No further frames follow for this turn.
     TurnComplete { outcome: TurnOutcomeWire },
+}
+
+/// What the turn has become, as the engine can witness it — the payload of a
+/// [`ServerFrame::RequeryRequest`].
+///
+/// An owned projection of `stella_core::steering::TurnSignal`, which borrows
+/// the turn's own transcript and so cannot cross a frame boundary.
+///
+/// One field of that type is missing here. `active_domains` is a workspace
+/// taxonomy the engine never fills — it passes an empty slice — and a wire
+/// field that is always empty reads as a fact. The host merges its own
+/// domains, and its own file ledger, before it asks its steering plane.
+#[cfg_attr(feature = "schema", derive(schemars::JsonSchema))]
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct RequerySignal {
+    /// The turn's goal, as the operator wrote it. On its own this is what
+    /// every selector had, and it is what goes stale by step 40.
+    pub prompt: String,
+    /// Tools the turn has called, most recent last.
+    pub recent_tool_calls: Vec<String>,
+    /// Workspace paths the turn has touched, read off its tool inputs.
+    pub touched_paths: Vec<String>,
+    /// Error classes the turn has seen. "We are stuck, and this is how" is the
+    /// strongest retrieval cue a turn produces.
+    pub errors_seen: Vec<String>,
+    /// Step index within the turn.
+    pub step: u32,
+    /// Steps since the host last answered with a block.
+    pub since_last_query: u32,
+}
+
+impl RequerySignal {
+    /// Copy out what the engine assembled at the step boundary.
+    #[must_use]
+    pub(crate) fn from_signal(signal: &stella_core::steering::TurnSignal<'_>) -> Self {
+        Self {
+            prompt: signal.prompt.to_string(),
+            recent_tool_calls: signal.recent_tool_calls.to_vec(),
+            touched_paths: signal.touched_paths.to_vec(),
+            errors_seen: signal
+                .errors_seen
+                .iter()
+                .map(|e| (*e).to_string())
+                .collect(),
+            step: signal.step,
+            since_last_query: signal.since_last_query,
+        }
+    }
 }
 
 /// Serializable projection of [`TurnOutcome`] for the wire. `TurnOutcome` lives
@@ -197,6 +257,24 @@ pub enum ProviderDelta {
     Text { text: String },
     /// A fragment of thinking/chain-of-thought content.
     Reasoning { text: String },
+}
+
+/// Host → engine: the answer to a [`ServerFrame::RequeryRequest`].
+///
+/// `context` is the block to put in front of the model, or `null` for "nothing
+/// here is worth the tokens" — the cheap answer, and the expected one for most
+/// boundaries. The engine appends the block to the volatile tail as a user
+/// message and never touches the byte-stable prefix.
+///
+/// The block is injected as sent, so the server prefixes the recall marker
+/// when the host's text does not already carry one: an injected block the
+/// engine read as a real user turn would move the turn window.
+#[cfg_attr(feature = "schema", derive(schemars::JsonSchema))]
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct RequeryResultIn {
+    pub request_id: String,
+    #[serde(default)]
+    pub context: Option<String>,
 }
 
 /// Serializable mirror of [`ProviderError`]'s taxonomy. The host classifies the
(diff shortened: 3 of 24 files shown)


macanderson:
Automated PR sweep. Pushed one fix, and verified this PR's DoD.

## Fixed: `docs/wire still matches the types` was red

`cf42758`. `served_capabilities` gained `requery` as its **fourth** parameter, ahead of `bus`, but the three call sites in `session.rs`'s own test module still passed four arguments — so `Some(&bus)` was offered to the re-query slot and the bus argument was missing outright:

```
error[E0061]: this function takes 5 arguments but 4 arguments were supplied   (x3)
error[E0277]: the trait bound `HookBus: SteeringRequery` is not satisfied
error: could not compile `stella-serve` (lib test) due to 4 previous errors
```

Each site now passes `None` for the new slot, which is what they already asserted: none of them supplies a re-query source.

Worth noting where this hid. `make lint-schema` is the only step that compiles this crate's **test** targets under the `schema` feature, so the default-feature run never saw it and a plain compile error surfaced under a check named for schema drift. `wire-schema` itself was never the problem — `serveframe.schema.json`, `serveinbound.schema.json` and `serveframe.d.ts` all carry the new arms.

Verified by the flip, not by the description — `cargo check -p stella-serve --features stella-protocol/schema,stella-serve/schema --all-targets`: 4 errors before, clean after, `cargo fmt --check` green. No workspace build run (SCR-001).

## DoD for #6121 — all five confirmed against the tree

| # | Item | Evidence |
|---|---|---|
| 1 | Re-query at step boundaries via a host-supplied source | `served_capabilities`' `requery` param (`session.rs:897`); `impl stella_core::ports::SteeringRequery for RemoteRequery` (`remote/requery.rs:184`) |
| 2 | Request/response in the generated wire schema | `serveframe.schema.json`, `serveinbound.schema.json`, `serveframe.d.ts` all carry `requery_request` / `RequeryResultIn` |
| 3 | Witness driving a drifted second step, failing on old code | `bridge.rs:761` |
| 4 | Byte-stable prefix pinned across a re-query | `bridge.rs:787` |
| 5 | `turn.steering_requery`'s `api` posture → `Shipped`, naming that witness | `stella-parity/src/lib.rs:298` — `api: SurfacePosture::Shipped`, witness `a_served_turn_requeries_...` |

No gaps found. `dod / dod` is red only because the boxes on #6121 are unticked — ticking them plus this push (the fresh run the check needs) should clear it.

## One thing to watch

#6150 / #6171 asks that the lane matrix rows "agree with the literals in ... `stella-serve`'s `served_capabilities`". This PR changes that function's arity and adds a bound seam. Whichever of the two lands second will need the other's shape — flagging rather than acting, since it is a design question for both authors.

---

macanderson:
Autonomous PR routine — one commit pushed, and one cross-PR hazard to flag.

### Pushed: `b0b5825`

`a_served_turn_binds_what_its_call_site_bound_before` exists to catch a seam this lane silently gains or loses, and this PR adds one — `requery` was the one seam the test did not mention. It now asserts both directions: absent when the host supplied none, present when it supplied one, with `SilentRequery` as the double for the supplied case.

Verified rather than asserted:

- `cargo clippy -p stella-protocol -p stella-plugin -p stella-serve --features stella-protocol/schema,stella-plugin/schema,stella-serve/schema --all-targets -- -D warnings` → exit 0.
- `a_served_turn_binds_what_its_call_site_bound_before` and `a_served_turn_declares_the_serve_session_lane` both pass.
- Negative control: replacing the supplied `Some(&requery)` with `None` fails the new assertion rather than passing quietly. Restored before committing.

I had also written the three-call-site repair for the `lint-schema` break (E0061 plus the E0277 that followed it, where the missing argument shifted `bus` into the `requery` position), but `cf42758` landed the same fix from another session while I was verifying, so my commit carries only the assertion. Nothing was reverted — I reset onto `cf42758` and re-applied the additive half, so no edit-and-undo pair is in this branch's history for `rebase-replay` to trip on.

### Hazard: this PR and #6171 cannot both merge as written

#6171 adds a lane-capability matrix whose `ServeSession` `requery` row is `SeamClaim::deferred("Refs #6158", …)`, and its `a_seam_claim_agrees_with_its_lane_literal` asserts a deferred row's field is written `requery: None` in `crates/stella-serve/src/session.rs` — which it reads through `include_str!`. This PR makes that literal write `requery,` instead. Whichever merges second leaves `main` red on that test, and neither PR's CI can see it because each is green against its own base.

Full detail and the suggested row change are in [my comment on #6171](https://github.com/macanderson/stella/pull/6171#issuecomment-5555933386). The assertion above is also what lets that row be honestly `Bound` with `SERVE_SEAMS` as its witness rather than `Witness::Literal`.

Separately: `dod / dod` is red here because #6121's five DoD boxes are unticked. Four of the five look satisfied by this diff; the fifth — moving `turn.steering_requery`'s `api` posture to `Shipped` in `crates/stella-parity/src/lib.rs` — I could not find in the diff. Worth a look before ticking, since `Refs` rather than `Closes` may be the honest link if that row has not moved.

---

macanderson:
**Correction to my previous comment — the DoD paragraph was wrong.**

I wrote that I "could not find" the `turn.steering_requery` `api` posture move in this diff. I had not looked at this PR's diff when I wrote that; I had only read #6171's. That was a claim made without the evidence to support it, and it is wrong.

Having now read it, all five of #6121's DoD items are satisfied:

| DoD item | Where |
|---|---|
| A served turn re-queries through a host-supplied source | `crates/stella-serve/src/remote/requery.rs` (new), bound in `served_capabilities` |
| Request and response in the generated serve wire schema | `docs/wire/serveframe.schema.json`, `serveinbound.schema.json`, `serveframe.d.ts` |
| Witness test: second step touches a new path, context arrives in the volatile tail | `a_served_turn_requeries_the_steering_plane_after_its_work_moves_to_a_new_path` (`tests/bridge.rs`), which states why it fails on `main` twice over |
| A test pins the byte-stable prefix across a served re-query | `a_served_requery_leaves_the_byte_stable_prefix_untouched` |
| `turn.steering_requery`'s `api` posture moves to `Shipped` and names that witness | `crates/stella-parity/src/lib.rs` — `api: SurfacePosture::Shipped { … witness: "a_served_turn_requeries_the_steering_plane_after_its_work_moves_to_a_new_path" }` |

So `Closes #6121` is the honest link, not `Refs`, and my suggestion otherwise should be ignored.

I have not ticked the boxes on #6121 myself. SCR-003 asks for satisfied *and verified*, and CI on the current head (`b0b5825`) is still queued — the implementation being present in the diff is not the same fact as the tests passing. Once `fmt + clippy + test` and `docs/wire still matches the types` come back green, ticking the five boxes and pushing any commit will clear the red `dod / dod` (the check reacts to PR events, not to the issue edit — `make dod-recheck N=6121` is the other route).

The cross-PR hazard with #6171 in my previous comment stands unchanged; that half I did verify against both trees before writing it.

---

macanderson:
**DoD verified against the diff at `b0b5825` — all five items on #6121 ticked.**

`dod / dod` was red only because #6121's boxes were unticked. I read each item against the branch rather than the PR description:

| DoD item | Evidence |
|---|---|
| Served turn re-queries at step boundaries through a host-supplied source | `crates/stella-serve/src/remote/requery.rs` — `RemoteRequery::requery` implements the port; bound in `session.rs` behind `steering_requery`, answered over `POST /v1/turns/{id}/requery-result` (`routes/requery.rs`) |
| Request/response in the generated wire schema, `make wire-schema` green | `docs/wire/serveframe.schema.json` (+72), `serveinbound.schema.json` (+22), `serveframe.d.ts` (+76); the **docs/wire still matches the types** check is green on this head |
| Witness driving a second step onto a new path | `a_served_turn_requeries_the_steering_plane_after_its_work_moves_to_a_new_path` (`tests/bridge.rs:761`) — asserts one ask, `signal_paths == ["src/adapter.rs"]`, and `block_reached_the_model` |
| Byte-stable prefix pinned across a re-query | `a_served_requery_leaves_the_byte_stable_prefix_untouched` (`tests/bridge.rs:787`) — system message byte-identical to `STABLE_PREFIX`, block injected exactly once as a `RECALL_MARKER`-prefixed user message |
| Parity row `api` → `Shipped` naming that witness | `crates/stella-parity/src/lib.rs` — `turn.steering_requery` moves `Deferred` → `Shipped`, witness field names the test above |

`a_host_that_declines_a_requery_injects_nothing` is the negative control for the first witness, which is what makes the pair falsifiable rather than merely present.

No gaps found; nothing filed. The issue edit should re-run `dod / dod` via `dod-recheck.yml`. Remaining blocker is `fmt + clippy + test`, still in flight.

---

Issue #6179, opened 2026-09-06
Benchmark evidence does not record which loop a trial ran


macanderson:
No committed benchmark evidence says which loop a trial ran, so an analysis
cannot tell two arms apart when the arms are two paths through the binary.

## The problem

`bench/evidence/score_dev_baseline.py`'s `extract_trial` folds a Harbor trial
into one `trials.jsonl` row. It carries the fields an analysis needs to check a
run's identity — `source_commit`, `binary_sha256`, `engine_posture_sha256`,
`assurance_arm` — and drops `stella_loop_mode`, which the adapter records on
the trial manifest.

`assurance_arm` does not fill the gap. It is derived from the engine posture
(`assurance_tiers_from_posture`) and answers which verification tiers the
posture asks for. It says nothing about which loop the binary ran.

That matters for the side-by-side benchmark #6155 preregisters. Its treatment
arm is a plugin path selected by a CLI flag, and a launcher that drops the flag
produces a full, plausible arm that ran the control path. The failure has
happened here before with a declared-but-unfired verification tier, and the
only thing that caught it was a per-trial field.

`bench/evidence/compare_arms.py` has a cross-build mode that takes
`--treatment-fired FIELD=VALUE` and checks that field on every trial in both
arms. It is waiting on this field to exist.

## Files

- `bench/evidence/score_dev_baseline.py` — `extract_trial`.
- `bench/harbor_adapter/stella_harbor/__init__.py` — where the trial metadata
  is assembled; `stella_loop_mode` is the producer side.
- `bench/harbor_adapter/stella_harbor/loop_mode.py` — `loop_mode_name`, which
  decides the value.
- `bench/evidence/compare_arms.py` — `check_path_fired`, the consumer.
- `bench/evidence/tests/test_evidence_tooling.py` — where extraction is
  covered.

## Constraint

`extract_trial` strips the `stella_` prefix on the way out (`stella_source_commit`
becomes `source_commit`), so the row key should be `loop_mode`. Read both
spellings if any recorded run used another: this file already carries a
two-spelling read for the witness author model after a producer rename left
the reader behind, and that bug shipped null for months.

## How to verify

`uv run --with pytest --no-project pytest -q bench/evidence/tests`, with a new
case asserting a manifest carrying `stella_loop_mode` produces a row carrying
`loop_mode`, and that a manifest without it produces `None` rather than a
guess.

## Definition of done

- [x] `extract_trial` carries the loop mode into the `trials.jsonl` row.
[...]

Issue #6178, opened 2026-09-06
Bench adapter cannot select a pipeline, so neither arm of the A/B can run


macanderson:
The bench adapter cannot launch either arm of the pipeline A/B, because it
never passes `--pipeline` to Stella at all.

## The problem

`bench/harbor_adapter/stella_harbor/loop_mode.py`'s `loop_argv` builds the
`stella run` argv for every benchmark trial. It emits no `--pipeline` flag in
any case. `loop_mode_name` always publishes `BARE_STEP_LOOP`, and the module
says why: with the built-in staged pipeline gone from the workspace, every
resolution lands on the raw step loop, so a manifest claiming a staged pipeline
would be claiming something the binary cannot run.

That reasoning covers `--pipeline classic`. It does not cover
`--pipeline <plugin id>`, which is the only verification path Stella ships
today and the treatment arm of the side-by-side benchmark #6155 preregisters.
So the adapter cannot run either arm:

- the control arm needs `--pipeline classic` on an old binary
  (`f4c24c12bde5578818f1141ec4e438291ac4db55`), where the flag still resolves;
- the treatment arm needs `--pipeline witness-v1`, the `[wrapper] id` declared
  in `plugins/stella-witness/plugin.toml`.

## Files

- `bench/harbor_adapter/stella_harbor/loop_mode.py` — `loop_argv`,
  `loop_mode_name`, `NO_PIPELINE_ENV`, `BARE_STEP_LOOP`, `STAGED_PIPELINE`.
- `bench/harbor_adapter/stella_harbor/__init__.py` — `_HOST_ONLY_STELLA_ENV`
  registration, and the argv assembly that calls `loop_argv`.
- `bench/harbor_adapter/tests/test_loop_mode.py` — the existing coverage.
- `bench/evidence/pipeline-ab/preregistration.json` — names this as a blocking
  precondition on the run.

## Constraints found while reading

- The selector must be host-only and registered in `_HOST_ONLY_STELLA_ENV`.
  The ambient check fails closed, so an unregistered selector refuses the run
  rather than quietly running the other arm.
- A selector spelled to close must not open: match Stella's truthy vocabulary
  through `is_truthy`, the way `NO_PIPELINE_ENV` does.
- `secure_launcher` accepts direct `stella` argv only, and the TB2.1 study's
  intent digests are frozen. Check whether adding argv moves
  `_canonical_payload_sha256` for the frozen stages before changing anything
  the launcher signs.
[...]

PR #6180, opened 2026-09-06
Preregister the two-build pipeline A/B, and make its analysis expressible

### LINKED ISSUE #6155: Run the side-by-side benchmark that the deleted pipeline was promised
Epic #4029's P3.4, filed as its own handoff so it stops living only in a
checkbox.

## The problem

`doc:pipeline-as-plugins` §7 records the deletion of the built-in staged
pipeline (#3865) as a deliberate call about code, and sets a bar for it: a
side-by-side measurement showing the plugin path is not worse than the path it
replaced. That measurement has never been run. `plugins/stella-witness` now
exists and is the verification path a user can actually install, so the
comparison is finally possible — and nothing in the tree reports it.

The only benchmark artifact for a nearby question is
`bench/evidence/witness-ab/preregistration.json`, whose own `status` field
reads `plan_fixed_run_not_started`. It is also a different experiment (#1284:
witness authoring on vs off, one binary), not this one (the deleted built-in
path vs the plugin path, two binaries).

Until this runs, epic #4029 cannot close, and the project's claim that the
plugin path is at least as good as what was removed rests on nobody's
measurement.

## What the run is

Two arms, paired by task, on Terminal-Bench 2.1, with repeats per task:

- **control** — the built-in staged pipeline. It is readable at the commit
  before `a6d3db4f6`; build the SUT from there.
- **treatment** — `origin/main`, run as
  `stella run --pipeline witness-v1` with `plugins/stella-witness` installed.

Everything else held: same task list and dataset digest, same worker model,
same concurrency, same host, attempts and retries fixed, no budget or token
cap (a cap turns a measurement into an artifact of the cap).

## Rules this repository already holds a bench run to

Read these before launching; they are not optional and each exists because a
run was wasted:

- **Read `stella-events.jsonl`, never the summary.** `result.json` and the UI
  are projections and each has its own bugs. Census `(role, model)` from
  `step_usage` before believing which model ran; read full `proof` reason
  strings.
- **Name and exclude operational aborts explicitly.** A run that dies on a
  provider 402 is not a task failure.
- **Report the unflattering number.** A result that favours the plugin path
  because of a measurement artifact costs more than a loss.
- **Two arms differing by more than one commit are confounded**, and the two
  arms here differ by hundreds. Say so in the report rather than implying a
  clean A/B: the honest claim available is "the plugin path is not
  catastrophically worse", not "this change caused that delta".

## Files

- `bench/RUNBOOK.md`, `bench/READINESS.md` — how a run is launched here.
- `bench/evidence/compare_arms.py` — the paired analysis, already written.
- `bench/evidence/witness-ab/preregistration.json` — the shape a
  preregistration takes in this repository; write P3.4's own beside it before
  the first paid call, not after.
- `plugins/stella-witness/` — the treatment arm's plugin.
- `docs/spec/pipeline-as-plugins.md` §7 — the bar being closed.
- `docs/spec/plugin-completion-plan.md` §6 P3.4, §7 — the statement of done.

## Constraints

- Needs a benchmark rig and real spend. It cannot be done by reading the tree,
  and it cannot be done on the maintainer's laptop.
- The control arm needs a build from a commit whose crate (`stella-pipeline`)
  no longer exists in the workspace, so budget time for getting that binary to
  build at all.

## Definition of done

- [x] A preregistration committed under `bench/evidence/` — arms, task list,
      denominator, seeds, decision rule and the SUT commit of each arm —
      before the first paid call.
      Landed in PR #6180 as `bench/evidence/pipeline-ab/preregistration.json`
      and its README. The control arm's SUT is pinned to
      `f4c24c12bde5578818f1141ec4e438291ac4db55`; the treatment arm's exact
      build commit sits in `to_be_recorded_before_the_first_paid_call` beside
      a floor commit, on the same rule `witness-ab/` uses, because that binary
      does not exist yet.
- [ ] Both arms run on TB2.1 with repeats per task, no budget or token cap.
- [ ] A report posted on this issue and on #4029: per-task table, paired
      outcome, operational aborts named and excluded, spend per arm.
- [ ] The report states plainly that the two arms differ by more than one
      commit and what that does to the strength of the conclusion.
- [ ] `docs/spec/pipeline-as-plugins.md` §7's bar is marked closed with the
      result, whatever the result is.
- [ ] #4029's P3.4 box is ticked with a link to the report.

macanderson:
## What this is

`doc:pipeline-as-plugins` §7 set one bar for deleting the built-in staged
pipeline: a side-by-side benchmark showing the plugin path that replaced it is
not worse. The deletion shipped ahead of that bar as a recorded maintainer
call, and the measurement has never been run.

The run needs a benchmark rig, real spend, and a build of a crate this
workspace deleted. This PR lands the part that has to exist **before** the
first paid call and cannot be written after it, and reports what reading the
tree turned up.

**This does not close #6155.** Only its first done-item is satisfied here.

## What lands

**`bench/evidence/pipeline-ab/preregistration.json` + `README.md`** — arms,
task list, denominator, seeds, replicates, decision rule and the SUT commit of
each arm, fixed before any data exists.

- control: `f4c24c12bde5578818f1141ec4e438291ac4db55`, the parent of
  `a6d3db4f6` and the last commit where `crates/stella-pipeline` exists.
  Verified: `git show f4c24c12b:crates/stella-pipeline/Cargo.toml` resolves,
  and `--pipeline classic` still selects the staged pipeline in that tree
  (`crates/stella-cli/src/agent/goal.rs` at that commit).
- treatment: a build of `main` running `stella run --pipeline witness-v1` —
  the `[wrapper] id` declared in `plugins/stella-witness/plugin.toml`.
- repeats are three preregistered replicates rather than several attempts per
  task, because the analysis keys one row per task per arm and refuses a
  duplicate; three pairs also measure run-to-run variance instead of hiding it
  inside a majority vote.
- the stop rule is keyed to spend and fixed here. No stop may be triggered by
  a score.

**`bench/evidence/compare_arms.py` gains a declared cross-build mode.** The
issue calls this analysis "already written". It is, for a different experiment:
`check_identity` refuses two arms whose `source_commit` or `binary_sha256`
differ, and demands the arm labels `witness-off`/`witness-on`, which come from
`assurance_tiers_from_posture` and say which verification tiers the posture
asks for — not which loop the binary ran. Both refusals fire on this experiment
by construction.

The new mode is opt-in and **stricter than the default, never looser**:

| default | `--cross-sut A:B --treatment-fired f=v` |
|---|---|
| one `source_commit` across both arms | each arm must report the commit declared for it |
| one `binary_sha256` across both arms | the two hashes must differ |
| arm labels `witness-off`/`witness-on` | both arms must declare the *same* posture arm — only the build may move |
| treatment must have authored a witness | every treatment trial must carry `f=v`, and no control trial may |

`--cross-sut` and `--treatment-fired` are one declaration: passing one alone
exits 2, so the mode cannot be reached by forgetting the proof half. Every
report it produces carries `CROSS_SUT_CONFOUND` in the JSON and a
`> **Confounded.**` line in the markdown, fixed in the module so no run can
soften it.

## Witness mechanism

`bench/evidence/tests/test_compare_arms_cross_sut.py`. It fails on the old code
by construction: `compare()` takes no `cross_sut` argument and the module has
no `CrossSut`, `parse_cross_sut` or `CROSS_SUT_CONFOUND`.

Checked the artisanal way — `git stash push bench/evidence/compare_arms.py` and
re-run:

```
12 failed, 1 passed
AttributeError: module 'compare_arms' has no attribute 'parse_cross_sut'
```

The one that passes on both trees is the contrast case
(`test_two_builds_are_refused_when_nobody_declared_them`), which asserts the
default still refuses an undeclared split. With the change: `79 passed`
across `bench/evidence/tests/`, which `bench.yml`'s required
`harbor_adapter + analyzer pytest` job runs on a `bench/**` diff.

## Exemplar

`bench/evidence/witness-ab/` — this preregistration takes its field shape, its
`to_be_recorded_before_the_first_paid_call` block and its refusal vocabulary
from that directory rather than inventing a second dialect.

## Two blockers found while reading, filed rather than fixed

Both are preconditions on the run, and the preregistration names them:

- **#6178** — `loop_argv` in
  `bench/harbor_adapter/stella_harbor/loop_mode.py` emits no `--pipeline` flag
  in any case, so neither arm can be launched by today's adapter.
- **#6179** — `extract_trial` in `bench/evidence/score_dev_baseline.py` does
  not carry `stella_loop_mode` into `trials.jsonl`, so no committed field says
  which loop a trial ran. That is the field `--treatment-fired` points at.

Fixing either one touches the frozen launcher argv contract or the trial row
shape, which is a separate change with its own tests; folding them into a
preregistration PR would put three unrelated diffs under one review.

## What is deliberately not here

The run itself, and the remaining five done-items of #6155. They need a rig,
a credential and real money, and no amount of reading the tree substitutes.

Tests deleted: None.

Refs #6155

## Summary by Sourcery

Preregister the built-in-versus-plugin pipeline benchmark and make its explicitly confounded two-build analysis enforceable.

New Features:
- Add a preregistered plan for comparing the last built-in pipeline against the replacement plugin pipeline across three paid replicates.
- Support explicitly declared cross-build A/B analysis with commit, binary, posture, and path-execution validation plus mandatory confound reporting.

Enhancements:
- Preserve strict default same-build identity checks while extending comparison reports and markdown output for declared two-build experiments.

Documentation:
- Document the pipeline A/B protocol, decision rules, limitations, and current blockers.

Tests:
- Add coverage for cross-build declarations, identity validation, path execution checks, CLI requirements, refusal cases, and confound reporting.

Chores:
- Record adapter and trial-schema blockers needed before running the benchmark without fixing them in this change.

### DIFF
--- a/bench/evidence/README.md
+++ b/bench/evidence/README.md
@@ -15,6 +15,8 @@ score_dev_baseline.py     scoring: trials.jsonl -> pass@1 + two 95% intervals
 make_manifest.py          identity: freeze every input that can move the number
 compare_arms.py           two arms of one experiment -> did the change pay?
 witness-ab/               the authored-witness A/B: protocol, plan, decision rule
+pipeline-ab/              the cut built-in path against the plugin path that
+                          replaced it: protocol, plan, decision rule
 tests/                    what keeps the three of them honest
 <run-id>/
   run-manifest.json       the frozen inputs, and why the run is not a claim
@@ -81,6 +83,12 @@ spend per additional task passed. It refuses — non-zero exit, `"verdict":
 one experiment, or when the treatment arm declared a verification tier it never
 exercised. See [`witness-ab/`](witness-ab/) for the experiment it was written
 for and the decision rule it applies.
+
+A declared cross-build mode drops the arm-label check and replaces it with
+stricter ones: each arm must report the commit named for it, the two binary
+hashes must differ, and a per-trial field must show the treatment arm ran the
+path under test. Its report carries a confound line no caller can drop. See
+[`pipeline-ab/`](pipeline-ab/).
 
 ### Reading a descriptive total
 
--- a/bench/evidence/compare_arms.py
+++ b/bench/evidence/compare_arms.py
@@ -46,10 +46,19 @@ Three properties, in the order they can invalidate a conclusion:
    same reward as one that fails and reports success, but only the second one
    lies to a caller who then ships it.
 
+**Two builds, when one build cannot ask the question.** The default above
+holds the binary still and moves one setting. Comparing a deleted path against
+the path that replaced it cannot work that way: it takes one build of each.
+That shape is refused unless the caller declares it, names both commits, and
+names the per-trial field that proves the treatment arm ran the path under
+test. Such a comparison is confounded by every commit between the two builds,
+and its report says so. See ``pipeline-ab/README.md``.
+
 Usage::
 
     python compare_arms.py <control-trials.jsonl> <treatment-trials.jsonl> \\
         --tasks 89 [--markdown out.md]
+        [--cross-sut <control-sha>:<treatment-sha> --treatment-fired f=v]
 
 Exit status is 0 only for a comparison that is allowed to mean something. The
 report is printed either way — a refusal with the numbers withheld is a refusal
@@ -63,6 +72,7 @@ import json
 import math
 import random
 import sys
+from dataclasses import dataclass
 from datetime import datetime
 from pathlib import Path
 from typing import Any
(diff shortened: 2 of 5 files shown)


macanderson:

Both are correct readings of the diff and both are out of scope on purpose.
This PR carries `Refs #6155`, not `Closes #6155` — it satisfies the first
done-item and leaves the issue open.

**"Run the control and treatment arms side by side … publish the per-task and
paired results, spend for both arms."** Not done, and not doable from this
tree. The run needs a benchmark host, a spend-capable credential, and a build
of `crates/stella-pipeline`, a crate this workspace deleted. It also needs the
two blockers this PR filed rather than fixed: `#6178` (the adapter emits no
`--pipeline` flag, so neither arm can be launched) and `#6179` (`trials.jsonl`
carries no field saying which loop a trial ran, which is the field the new
`--treatment-fired` check reads). Fixing either touches the frozen launcher
argv contract or the trial row shape and belongs in its own change.

**"Close the documented bar by recording the result and linking it from
`doc:pipeline-as-plugins` §7 and P3.4."** Marking §7's bar closed is a claim
about a measurement. There is no measurement yet, so editing the spec now would
put a sentence in the normative document that no evidence supports — the exact
shape of defect that bar exists to prevent. The preregistration fixes what each
of the three possible results does to §7 before any data exists
(`decision_rule` in `bench/evidence/pipeline-ab/preregistration.json`), so the
spec edit is mechanical once a number lands, and cannot be chosen after seeing
one.

Both objectives stay on `#6155`, which is why the issue stays open.

macanderson:
Post-merge note: this PR closed #6155 by accident. The body's sentence "This does not close #6155" reads to GitHub's parser as a closing keyword plus a reference — the negation in front of it is invisible. The issue is reopened, its first done-item is ticked with the evidence, and the trap is filed as #6190.

PR #6187, opened 2026-09-06
feat(stella-autonomy,stella-cli): age a dedup digest out once its issue closes

### LINKED ISSUE #6149: A defect the loop already filed once can never be filed again, even after it is fixed and comes back
The second half of phase B5 in #3599. The gate itself landed; this is the part that did not.

## The problem

The residue gate works. `crates/stella-cli/src/self_driving_cmd/residue.rs` reads a turn's own prose for statements like "follow-up:" or "out of scope for this change", files each one as an issue through `backlog::file_finding`, and is wired at the end of every turn from `crates/stella-cli/src/agent/reflect.rs`. Repeats are dropped by digest against `seen.txt`, proven by `two_residue_statements_file_two_issues_and_a_rerun_files_none`.

**A digest, once in `seen.txt`, is there forever.** Nothing ages an entry out. Search the loop's whole surface for any notion of decay and the only hits are about the cycle-size ceiling, not about findings:

```
crates/stella-autonomy/src/lib.rs: "ceiling stayed low through 5 clean runs — the decrease never decayed"
```

So a defect that was found, filed, fixed, closed, and then reintroduced a year later is silently not filed the second time. The dedup set cannot tell "we already know about this" from "we knew about this once, and the thing that made it true came back". Those are different, and one of them is the case the loop exists to catch.

`stella_autonomy::seen` and `crates/stella-cli/src/self_driving_cmd/state.rs` (`seen_path`, `seen`, `seen_count`) hold the set today: one digest per line, no timestamp, no commit, no state.

## Why it is not just "add a timestamp"

`finding_digest`'s normalization is a **byte-for-byte contract** with every `seen.txt` on every machine that has ever run the loop — `the_seen_set_survives_the_move_byte_for_byte` in `state.rs` pins it. A format change that rewrites those lines re-files every finding the loop has ever filed, all at once, as new issues. So decay has to be added in a way that reads an old `seen.txt` unchanged and treats every legacy line as "seen, age unknown".

This also matters to #6146 (phase B4): re-arming a dry lens is only safe because repeats are dropped by digest. Decay and re-arm interact, and whichever lands second has to be tested against the other.

## What to build

- A decay rule with a stated reason for its shape — age, or the closure of the issue the digest was filed as, or a base-commit distance. The `IssueProvider::close` receipt path in `crates/stella-cli/src/self_driving_cmd/lifecycle.rs` is one place the "this was fixed" signal already exists.
- A `seen.txt` upgrade path that never re-files a legacy line as new.

## Definition of done

- [x] A finding whose issue was closed and whose cause returned is filed again, with a test that fails before the change.
- [x] A finding that is simply repeated within the decay window is still dropped, so the existing dedup tests pass unchanged.
- [x] An existing `seen.txt` written by today's binary is read without re-filing anything from it, with a test using a fixture in the old format.
- [x] `finding_digest`'s normalization is untouched, proven by `the_seen_set_survives_the_move_byte_for_byte` still passing as written.
- [x] Whatever the decay rule keys on is written down where an operator reads it, next to `self_driving.residue_gate` (see #5645 for that reference page).

macanderson:
## What and why

A digest in the self-driving loop's `seen.txt` dropped a repeat forever. So a defect the loop found, filed, fixed, closed and then met again a year later was silently not filed the second time. The dedup set could not tell "we already know about this" from "we knew about this once, and the thing that made it true came back", and the second is the case the loop exists to catch.

A digest now stops dropping repeats once **every issue it was filed as has closed on a cited change** — a pull request or a commit. That closure is the loop's own word that the defect is fixed, so stating the same finding after it is news rather than a repeat.

Closure rather than age or base-commit distance, for the reason the issue names: the loop already writes a `ClosureReceipt` as it closes, so the signal is local, dated and free. Any age window would be a number nobody can defend — a defect returning at six months and one returning at eighteen are the same event.

A closure that cited nothing decays nothing. Somebody declined that work, or it was a copy of another issue, and neither is a fix.

## The upgrade path

`finding_digest`'s normalization and the format of `seen.txt` are untouched, because those lines are a byte-for-byte contract with every machine that has ever run the loop. What decays a line lives in a second file: `filings.jsonl`, one JSON object per filing, pairing a digest with the issue key it became.

A `seen.txt` line with no filing record beside it never decays. That is every line written before this change, so an upgrade re-files nothing. Decay also only bites when the finding is *stated again* — a fixed defect is not found by a lens, so a closure on its own files nothing.

`LoopState::record_filing` writes the digest and the filing record in one call, so a line that can never decay cannot be produced by accident. `seen.txt` is written first: losing the second write costs a line that cannot decay, while losing the first would re-file a finding the tracker already took.

## Files

- `crates/stella-autonomy/src/seen.rs` — new, pure, no I/O: `Filing` and `live(seen, filings, fixed)`.
- `crates/stella-cli/src/self_driving_cmd/state.rs` — `filings_path`, `filings()`, `record_filing()`, `live_seen()`.
- `crates/stella-cli/src/self_driving_cmd/{supply,residue}.rs` and `self_driving_cmd.rs` — the three filing paths read `live_seen()` and record through `record_filing`.
- `crates/stella-cli/src/self_driving_cmd/query.rs` — `self-driving seen --new` reads the live set, so the verb answers what the filers would do. `--add` still writes a bare digest, which is the one path with no issue key to pair, and such a line never decays.
- `stella.example.toml`, `docs/spec/backlog-self-driving.md`, `crates/stella-autonomy/README.md` — what the rule keys on, next to `self_driving.residue_gate`.

## Witness mechanism

Three tests, each of which fails on `main` by construction — `live_seen`, `record_filing`, `filings.jsonl` and `stella_autonomy::seen` do not exist there, so the code under test cannot compile against the old tree, and the behaviour it asserts has no other path to it.

- `a_defect_whose_fix_was_closed_is_offered_again` (`crates/stella-autonomy/src/seen.rs`) — the pure rule: one seen digest, one filing record naming issue 4100, 4100 in the fixed set, and the live set comes back empty.
- `a_closed_fix_lets_its_digest_be_filed_again` (`crates/stella-cli/src/self_driving_cmd/state.rs`) — the wiring over a real state directory: `seen()` still returns the file's one line, and `live_seen()` returns nothing. On `main` every filing path read `seen()`, so the repeat was dropped whatever had happened to the issue.
- `a_legacy_seen_set_is_read_whole` (same file) — a `seen.txt` with two digests and no `filings.jsonl`, beside a cited closure, keeps both lines.

Plus `a_closure_citing_nothing_keeps_the_line`, `recording_a_filing_writes_both_halves`, and four more directional cases in `seen.rs` (an open issue holds its line; an open filing beside a closed one holds it; a blank line survives; an unmatched record is ignored).

The receipt fixtures are serialized through `ClosureReceipt` itself rather than hand-written JSON, so they cannot drift from the shape `receipts()` reads — a hand-written copy of that shape was wrong on the first attempt.

## Exemplar

`crates/stella-autonomy/src/regress.rs` — the sibling module that already reads closure receipts back and turns a fact about them into a finding. `seen.rs` copies its shape: pure functions over borrowed data, the I/O left to the caller, one witness test per direction.

## Evidence

CI on this branch. Nothing was compiled locally.

Tests deleted: None.

## Remaining work filed

- #6188 — a digest never ages out when a person closes the issue the loop filed. Decay reads the loop's own closure receipts, and nothing else writes one.
- #6189 — a reverted fix can file two unlinked issues: the regress sweep's report, and the decayed original restated by a lens.

Closes #6149

## Summary by Sourcery

Age self-driving dedup digests out after all associated issues close on cited fixes so recurring defects can be filed again.

New Features:
- Age self-driving dedup digests out of the live set after every associated issue closes on a cited change.
- Track digest-to-issue filing records in `filings.jsonl` while preserving the existing `seen.txt` format and legacy behavior.

Bug Fixes:
- Allow previously fixed findings to be filed again when they reappear, while continuing to suppress findings linked to open or non-fixing closures.

Enhancements:
- Apply the live dedup set consistently across supply, residue, direct filing, and `self-driving seen --new` paths.
- Document the dedup decay behavior and its upgrade-safe handling.

Documentation:
- Document the new dedup decay rule and its state files in the autonomy README, backlog specification, and example configuration.

Tests:
- Add pure-rule and state-level coverage for closed fixes, open filings, non-fixing closures, legacy seen sets, unmatched records, and paired filing writes.

### DIFF
--- a/crates/stella-autonomy/README.md
+++ b/crates/stella-autonomy/README.md
@@ -87,6 +87,7 @@ the CLI does — so the dashboard and the terminal cannot disagree.
 | [`src/escalation.rs`](src/escalation.rs) | Escalation as a cooldown: `classify` reads an abort message, `retry_after` says how long to wait, `park_after` says when waiting ends. The record is written into the issue body and read back from it. |
 | [`src/gate.rs`](src/gate.rs) | Which checks are allowed to block a merge, and which have stopped earning that right. |
 | [`src/supply.rs`](src/supply.rs) | Where the loop looks when the ranked queue is empty: `WorkSupply`, `SupplyPolicy` (every switch off), the `rearm` rule that re-opens a dry ladder against a moved base, and `novel`, which drops a finding the seen set already holds. |
+| [`src/seen.rs`](src/seen.rs) | The dedup set, and the one thing that ages a line out of it: `Filing` pairs a digest with the issue it became, and `live` drops a line whose every issue closed on a cited change. A line with no filing record never decays. |
 | [`src/regress.rs`](src/regress.rs) | Re-checking the fixes the loop has claimed: `ClosureReceipt`, `check`, and the `sweep` that turns a change gone from the base branch into a fresh defect. |
 | [`src/meta.rs`](src/meta.rs) | The loop read against its own ledger: a raised signal nobody acted on, and a lens that has looked several times and found nothing. |
 | [`src/closure.rs`](src/closure.rs) | How an issue ends, and what has to be true first. |
--- a/crates/stella-autonomy/src/lib.rs
+++ b/crates/stella-autonomy/src/lib.rs
@@ -62,6 +62,7 @@ pub mod meta;
 pub mod priority;
 pub mod ready;
 pub mod regress;
+pub mod seen;
 mod stats;
 mod step;
 pub mod supply;
--- a/crates/stella-autonomy/src/seen.rs
+++ b/crates/stella-autonomy/src/seen.rs
@@ -1,0 +1,191 @@
+//! The dedup set, and the one thing that ages a line out of it.
+//!
+//! `seen.txt` holds one digest per line. A digest in it drops a repeat
+//! before anything is sent. That is what stops the loop filing one defect
+//! twice.
+//!
+//! Nothing ages a line out on its own. So a defect the loop found, filed,
+//! fixed and closed can come back a year later, and the set drops it in
+//! silence. "We know about this" and "we knew about this once, and the cause
+//! is back" are different answers. The second is the one the loop is for.
+//!
+//! What decays a line is the closure of the issue it became. A closure that
+//! cites a change is the loop saying that defect is fixed. Stating the same
+//! finding after that is news, so the line stops suppressing and the finding
+//! is filed again.
+//!
+//! A closure that cites nothing decays nothing. Somebody declined that work,
+//! or it was a copy of another issue, and neither of those is a fix.
+//!
+//! Two rules keep an old set safe:
+//!
+//! - A line with no filing record beside it never decays. Every `seen.txt`
+//!   written before this module existed is such a line, so an upgrade files
+//!   nothing again.
+//! - One closed issue does not decay a line that another issue still holds
+//!   open.
+//!
+//! [`crate::finding_digest`] is untouched here. What a line says is a
+//! byte-for-byte contract with every machine that has run the loop, so the
+//! record that decays a line lives in a second file.
+
+use std::collections::BTreeSet;
+
+use serde::{Deserialize, Serialize};
+
+/// One filing the loop made: which finding, and which issue it became.
+///
+/// One line of `filings.jsonl` in the loop's state directory.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct Filing {
+    /// The dedup digest, spelled as `seen.txt` spells it.
+    pub digest: String,
+    /// The issue key the tracker gave back, with no leading `#`.
+    pub key: String,
+    /// When it was filed, RFC3339. Empty on a record that predates this
+    /// field.
+    #[serde(default)]
+    pub at: String,
+}
+
+impl Filing {
+    /// Record that `digest` was filed as `key` at `at`.
+    #[must_use]
+    pub fn new(digest: &str, key: &str, at: &str) -> Self {
+        Self {
+            digest: digest.trim().to_owned(),
+            key: issue_key(key).to_owned(),
+            at: at.trim().to_owned(),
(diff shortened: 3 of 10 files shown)


macanderson:
Ticked #6149's five DoD boxes after checking each against this diff. What established each one, so the ticks are auditable rather than asserted:

1. **Closed fix stops suppressing** — `stella-autonomy/src/seen.rs::a_defect_whose_fix_was_closed_is_offered_again` and `state.rs::a_closed_fix_lets_its_digest_be_filed_again`. Both drive `live()` / `live_seen()`, neither of which exists on `main`, so on the old tree they do not compile rather than fail — the usual shape for a witness that names a new symbol.
2. **A repeat inside the window is still dropped** — `a_repeat_is_dropped_while_its_issue_is_open` and `a_closure_citing_nothing_decays_nothing`. The existing dedup tests are unmodified: `state.rs` is +185/−0 in this diff.
3. **A legacy `seen.txt` re-files nothing** — `a_legacy_seen_set_is_read_whole` seeds bare digests with no `filings.jsonl` beside them, and `seen.rs::a_line_with_no_filing_record_never_decays` covers the same rule at the pure layer.
4. **`finding_digest` untouched** — the only `stella-autonomy/src/lib.rs` change is `pub mod seen;`, and the zero-deletion `state.rs` diff leaves `the_seen_set_survives_the_move_byte_for_byte` as written.
5. **The rule is written where an operator reads it** — `stella.example.toml` gains 15 lines directly under `residue_gate = "on"`, naming the closure-on-a-cited-change key and `filings.jsonl`.

No gaps left against the DoD. Sourcery's linked-issue table carries no ❌ rows. `dod / dod` should go green on its next run; `fmt + clippy + test` was still in flight at the time of writing.

---

PR #6203, opened 2026-09-06
docs(stella-cli,stella-serve): plain language in the steering-allowance comments

macanderson:
## What and why

`main` is red on `prose` (#6196). Three separate regressions, and the first
one hides the rest, because `check-prose.py` returns as soon as the count
check fails:

1. **A bare issue number in a comment.** `crates/stella-cli/src/fleet_cmd.rs`
   went to 54 `issue-reference` hits against a ratchet of 53 (#6168). The
   sentence already states the fact, so the number goes, which is the remedy
   the guard prints.
2. **Four reading grades over their ceiling**, from the same change:
   `fleet_cmd.rs` 11.26 (11.23 allowed), `memory/recall.rs` 12.21 (12.19),
   `settings/steering.rs` 8.70 (8.41), `subagent.rs` 10.42 (10.39).
3. **Two older drifts** the count failure was hiding:
   `crates/stella-serve/src/observe/event.rs` 8.63 (8.62, from #6172) and
   `docs/spec/wrapper-socket.md` 12.20 (12.16, from #6176).

Every edit is comment or document text. No code changes, no baseline entries
— the ratchets are meant to reach empty, so the sentences got shorter instead.

## Evidence

- `python3 scripts/check-prose.py` exits 0 on this tree: "17009 grandfathered
  construction(s) in 1921 file(s), none added; 28 unit(s) within their
  header-length ceiling; 2219 file(s) within their reading-grade ceiling."
  It exits 1 on `main` at `9136a2092`, naming `fleet_cmd.rs [issue-reference]:
  53 allowed, 54 found`.
- `make guards-fast` is green.

## Witness

The guard is the witness: `prose` fails on `main` and passes here, on the same
command CI runs. That is the ordinary shape for a repair of a ratchet — there
is no new behaviour to pin.

## Claim

`./scripts/main-red-claim.sh claim` posted the claim on #6196 before this was
written.

Tests deleted: None.

Closes #6196

## Summary by Sourcery

Simplify prose throughout the steering-allowance and re-query documentation while updating the corresponding parity metadata.

Enhancements:
- Clarify and simplify steering-allowance comments and related prose across the CLI, serve, parity, and wrapper-socket documentation.

Documentation:
- Improve the wrapper-socket specification's plain-language descriptions of plugin event behavior.

Chores:
- Remove an obsolete issue reference and mark the serve re-query capability as bound in parity metadata.

Issue #5122, opened 2026-08-26
plugin: a conformance kit for third-party wrapper authors — replay the wire corpus against an external process


macanderson:
The wire contract is gate-checked from the inside — `stella-plugin` ships
`wire_corpus.rs` (the generated socket description under `docs/wire/`) and
`wire_schema.rs` (JSON Schema for the two point messages, #3532), both behind
`check-wire-schema.sh`. What does not exist: a way for a **plugin author** to
prove their process speaks the contract without building Stella from source
and stepping through a live turn. The first external verification wrapper
(Vera) is exactly the consumer that hits this wall: today its only test
harness is "run a whole agent turn and see what breaks."

## What to do

`stella plugin conform <plugin-dir>` (naming to taste — beside whatever
install/list verbs exist):

1. Load and validate the manifest exactly as the host would
   (`PluginManifest` + `reconcile` against the package listing), reporting
   every `ManifestError` with its span — this half is nearly free, the types
   exist.
2. Spawn the declared `[runtime]` argv with the declared env slice, then
   replay a fixture corpus derived from `wire_corpus`: handshake /
   `before_turn` / `after_turn` requests with representative
   `RoundState`/`EvidenceSet` payloads, plus the hostile cases — unknown
   field (must be ignored per serde posture), oversized reply, slow reply
   (timeout path), garbage JSON, early exit.
3. Score each exchange against `wire_schema`'s derived schema and the
   `LoopGrant` filters (`permits_hook` / `permits_point` / `permits_call`) —
   a plugin answering a point its grant does not permit is a finding, not a
   pass.
4. Exit non-zero with a findings table; `--json` for CI use in the plugin
   author's own repo.

## Why this is cheap and worth it

Every piece exists: manifest validation is pure functions, the corpus is
generated and gate-checked, the process discipline (argv spawn, env
allowlist, timeout, group kill) is the host's existing runtime code. The kit
is composition plus fixtures. The payoff is the ecosystem bet the plugin
system makes: a third party can reach "my process is conformant" in minutes,
in their own CI, on a machine that has never compiled Stella — which is the
difference between an extension surface and an extension surface people
actually build on.

## Witness

A fixture "good" plugin (the repo's own example wrapper, or a 30-line Python
echo participant) passes; each hostile fixture produces its named finding;
the conform run needs only the `stella` binary and the plugin dir — no
workspace checkout.

---
**Suggested worker:** `model-tier:balanced` · **Estimated autonomous time:** ~75 min

[...]

PR #6204, opened 2026-09-06
docs(adr): the wrapper socket is the plugin SDK, and the canary that keeps it true

### LINKED ISSUE #6148: Plugin authors write Python against no library — decide whether an SDK ships, or say the socket is the SDK
Epic #3246's step **P5 — the Python SDK** is the last step of its sequencing plan
with nothing in the tree, and it has no issue of its own. This is that issue, so
closing #3246 does not drop it.

## What is there today

Seven first-party plugins in `plugins/` are written in Python and work
(`plugins/stella-research/main.py`, `plugins/stella-plan/main.py`,
`plugins/stella-goal/main.py`, `plugins/stella-candidates/main.py`,
`plugins/stella-witness/main.py`, `plugins/stella-selfdriving/main.py`,
`plugins/stella-hello/panel.py`). Each one hand-rolls the wrapper socket: read
JSON from stdin, branch on the point, write JSON to stdout. There is no shared
library, no package anyone can install, and nothing that fails when the wire
contract changes under a plugin author who is not in this repository.

The reference declaration of the wire is `docs/wire/wrapper.d.ts` — TypeScript,
generated by `crates/stella-plugin/src/bin/export_wrapper_wire.rs`. There is no
Python equivalent.

## The decision that comes first

#3246's own closing comment (2026-08-26) put two options to a maintainer, and
neither has been answered:

1. A Python SDK is still wanted — a real package with its own release cadence,
   built over the wrapper socket, so a plugin author writes handlers instead of
   a JSON dispatcher.
2. The manifest-plus-subprocess contract **is** the SDK — any language over the
   socket, with `wrapper.d.ts` as the reference declaration — in which case P5
   is answered rather than outstanding and this issue closes as not planned.

Answer that before writing code. If option 1: a Python package does not belong
in this Cargo workspace for the reason `plugins/README.md` gives about plugins
themselves, so the first design question is where it lives and how a wire change
here fails the PR that made it.

## Related

- #5122 — a conformance kit for third-party wrapper authors (replays the wire
  corpus against an external process). Adjacent: that grades a plugin, this
  would be what a plugin is written against.
- #4030 — the plugin marketplace epic.
- `doc:wrapper-socket` — the contract any SDK would wrap.

## Definition of done

- [x] A maintainer picks option 1 or option 2 and the choice is recorded, as
      an ADR rather than as a comment (SCR-002).
- [x] Option 2 lands: `plugins/README.md` states that the manifest plus the
      socket is the whole author-facing contract, and names the generated
      declarations under `docs/wire/` a non-Rust author reads.
- [x] The claim the decision rests on is checked rather than asserted — every
      directory under `plugins/` holding a `plugin.toml` is spawned by an
      in-tree test, so a wire change reddens this repository's CI instead of a
      stranger's install.
- [x] That check has a self-test pinning both directions, and both run in CI.
- [x] The record says what would reopen the question, so the next reader does
      not re-ask it.

macanderson:
## What and why

Epic #3246's last sequencing step asked for a Python SDK, and its closing
comment put two options to a maintainer that nobody answered. SCR-002 says an
architecture decision is chosen and recorded, not asked, so this takes the
second option and writes it down:

**This repository ships no language SDK.** The author-facing contract is the
`plugin.toml` manifest, the JSON-over-stdio socket (`doc:wrapper-socket`), and
the three generated declarations under `docs/wire/`.

The premise the epic's step rested on was wrong.
`crates/stella-plugin/src/bin/export_wrapper_wire.rs` writes three committed
artifacts, not one: `wrapper.wire.json` (a byte corpus), `wrapper.schema.json`
(JSON Schema 2020-12, which no language owns) and `wrapper.d.ts`. The gate's
`wire-schema` step regenerates and diffs all three, so a Python author already
has a machine-readable declaration and a wire change that skips it fails the
PR that made it.

`macanderson/stella-examples` carries `verify-rs` / `verify-py` / `verify-ts` —
one plugin written three times over this socket, with no library in any of the
three.

## The mechanism the decision leans on, now checked

The argument for shipping no SDK is that a wire change breaks something *here*
rather than at a stranger's install, because every first-party plugin under
`plugins/` is spawned by an in-tree test. `plugins/README.md` claimed exactly
that and nothing checked it — add a plugin with no harness and the sentence
still reads true of the others.

`scripts/check-plugin-graded.sh` (`make plugin-graded`, a new gate step in
`GATE_GUARDS_FAST` — it compiles nothing) asserts that every directory under
`plugins/` holding a `plugin.toml` is named on a **non-comment** line of some
`crates/*/tests/**/*.rs`. The comment exclusion is not decoration:
`crates/stella-runtime/tests/driver_socket.rs` discusses
`plugins/stella-selfdriving` in a `///` line and spawns it nowhere, while
`crates/stella-cli/tests/self_driving_consent.rs` is what actually runs it.

## Witness mechanism

`scripts/test-plugin-graded.sh` (`make plugin-graded-test`, run by
`guard-self-tests.yml`) builds fixture trees and pins both directions, because
the real tree is green by construction:

- N1-N3: a plugin named in a test's code passes; a shared helper one level
  under `tests/` counts; a directory with no `plugin.toml` is skipped.
- O1: a plugin no test names **fails**, naming it.
- O2: a plugin named only in a `//` comment fails — a mention is not a run.
- O3: a plugin named in `src/` but by no test fails.
- E1-E3: an empty plugin root, a tree with no test sources, and an absent
  plugin root all fail rather than passing vacuously.
- R1: this repository's own seven plugins are each graded.

Every case fails on `main`, where neither script exists. Locally:
`./scripts/test-plugin-graded.sh` -> `passed 10, failed 0`, and
`./scripts/check-plugin-graded.sh` -> `OK — 7 first-party plugin(s), each
named by a test.`

## Exemplar

`scripts/check-consumer-sites.sh` and its `scripts/test-consumer-sites.sh`
self-test — same shape, same standard for what a path written into code
claims, same fixture-tree posture for proving the guard can still fail.

## Not done here

No SDK, and no port of a plugin onto one. That is what the decision refuses.
A conformance kit for third-party authors (#5122) is a different thing — it
grades a plugin, where this is about what a plugin is written against — and
ADR 0030 says to amend the record rather than add a library quietly if that
lands and authors still ask.

## Base

Branched from `main` at `9136a2092`, which is red on `prose` for reasons this
diff did not cause (#6196, repaired by #6203). This branch's own files are
clean: `scripts/check-plugin-graded.sh` grades 4.37,
`scripts/test-plugin-graded.sh` 2.91, the ADR 4.84 — all under the 6.00
ceiling a new file is held to — and `plugins/README.md` and `docs/adr/README.md`
both came **down** from their ratchets. `make guards-fast` is otherwise green,
including `gate-parity` (56 steps, named in AGENTS.md and CONTRIBUTING.md,
each run by a workflow) and `shellcheck` on both new scripts.

Tests deleted: None.

Residue filed: none new — #5122 already tracks the conformance kit, and ADR
0030 names it as the open question it does not answer.

Closes #6148

## Summary by Sourcery

Record that the wrapper socket is the plugin SDK and enforce that every first-party plugin remains covered by an in-tree test.

New Features:
- Document ADR 0030 establishing the manifest, JSON-over-stdio wrapper socket, and generated wire declarations as the complete plugin authoring contract instead of a language-specific SDK.

Enhancements:
- Add a guard that ensures every first-party plugin is referenced by an in-tree Rust test, preserving the repository's plugin-based wire compatibility canary.

Build:
- Expose plugin grading and its hermetic self-test through the Makefile and fast guard workflow.

CI:
- Run the first-party plugin grading check in CI and validate its failure cases in the guard self-test workflow.

Documentation:
- Add and register ADR 0030, and update plugin documentation to describe the no-SDK decision and the generated wire contract.

Tests:
- Add fixture-based coverage for valid, invalid, comment-only, source-only, empty, and repository plugin grading scenarios.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -485,6 +485,15 @@ jobs:
         if: ${{ !cancelled() && steps.checkout.outcome == 'success' }}
         run: ./scripts/check-reserved-paths.sh
 
+      # ADR 0030 says this repository ships no plugin SDK because the plugins
+      # under `plugins/` are the wire contract's canary: a test here spawns
+      # each of them, so a wire change reddens this run instead of a
+      # stranger's install. That only holds while every one of them has such
+      # a test.
+      - name: every first-party plugin is spawned by a test
+        if: ${{ !cancelled() && steps.checkout.outcome == 'success' }}
+        run: ./scripts/check-plugin-graded.sh
+
       - name: no library crate returns Result<_, String>
         if: ${{ !cancelled() && steps.checkout.outcome == 'success' }}
         run: python3 ./scripts/check-typed-errors.py
--- a/.github/workflows/guard-self-tests.yml
+++ b/.github/workflows/guard-self-tests.yml
@@ -183,6 +183,10 @@ jobs:
       - name: The rendering-facts guard can still fail
         run: ./scripts/test-rendering-facts.sh
 
+      - name: plugin-graded guard directions
+        if: ${{ !cancelled() }}
+        run: ./scripts/test-plugin-graded.sh
+
       - name: diagnostic-code registry failure directions
         if: ${{ !cancelled() }}
         run: ./scripts/test-diagnostic-codes.sh
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -147,6 +147,9 @@ make gate                # = no-scratch + no-secrets + design-refs
                          #   + reserved-paths (no Windows device name in a path)
                          #   + rendering-facts (no v2 rendering draws a
                          #     fact SPEC.md retired)
+                         #   + plugin-graded (every plugins/ directory is
+                         #     spawned by an in-tree test, so a wire change
+                         #     reddens this repo rather than an install)
                          #   + wire-schema
                          #   + lockfile-sync (cargo metadata --locked)
                          #   + format-check (fmt --check)
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -116,6 +116,7 @@ python3 ./scripts/check-deck-paths.py
 python3 ./scripts/check-css-vars.py
 ./scripts/check-reserved-paths.sh
 ./scripts/check-rendering-facts.sh
+./scripts/check-plugin-graded.sh
 ./scripts/check-wire-schema.sh
 ./scripts/check-lockfile-sync.sh
 cargo fmt --check
--- a/Makefile
+++ b/Makefile
@@ -49,7 +49,8 @@ GATE_GUARDS_FAST := no-scratch no-secrets design-refs action-pins cargo-install-
                     bench-suites wire-paths \
                     tokens hue-separation contrast light-clamp transcript-surfaces prose \
                     line-citations \
-                    deck-fit-all-test deck-paths css-vars reserved-paths rendering-facts
+                    deck-fit-all-test deck-paths css-vars reserved-paths rendering-facts \
+                    plugin-graded
 GATE_GUARDS := $(GATE_GUARDS_FAST) wire-schema
 
 # The cargo steps that resolve or parse but never build. They are not in
@@ -551,6 +552,14 @@ rendering-facts: ## Assert no v2 rendering draws a fact design/tui-v2/SPEC.md re
 .PHONY: rendering-facts-test
 rendering-facts-test: ## Test the rendering-facts guard's directions (hermetic; not part of `gate`)
 	@./scripts/test-rendering-facts.sh
+
+.PHONY: plugin-graded
+plugin-graded: ## Assert every first-party plugin is spawned by an in-tree test (ADR 0030, #6148)
+	@./scripts/check-plugin-graded.sh
+
+.PHONY: plugin-graded-test
+plugin-graded-test: ## Test the plugin-graded guard's directions (hermetic; not part of `gate`)
+	@./scripts/test-plugin-graded.sh
 
 .PHONY: dead-code-allows
 dead-code-allows: ## Assert every #[allow(dead_code)] says why, and that the count only shrinks (#3949)
--- a/docs/adr/0030-the-wrapper-socket-is-the-plugin-sdk.md
+++ b/docs/adr/0030-the-wrapper-socket-is-the-plugin-sdk.md
@@ -1,0 +1,79 @@
+---
+id: adr/0030-the-wrapper-socket-is-the-plugin-sdk
+title: "ADR 0030: The wrapper socket is the plugin SDK"
+status: implemented
+---
+
+# ADR 0030: The wrapper socket is the plugin SDK
+
+- Status: accepted
+- Date: 2026-09-06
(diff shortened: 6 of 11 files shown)


macanderson:
## The new guard's verdict was nondeterministic — fixed in `f75604c`

`guard self-tests` was red here on step 18, `plugin-graded guard directions`, with `R1 this repository's own plugins are each graded` failing:

```
plugins/stella-witness is graded by no test.
```

That verdict was wrong — `crates/stella-runtime/tests/witness_plugin_conformance.rs:58` does `.join("../../plugins/stella-witness")`, a non-comment line naming the directory, which is exactly what the guard looks for.

### Root cause

`check-plugin-graded.sh` ran the search as `sed … | grep -qF` under `set -o pipefail`. When the match is early in a file, `grep -q` exits while `sed` is still writing; `sed` takes SIGPIPE and reports 141, and `pipefail` hands the `if` a non-zero status. A file that names the plugin then reads as one that does not.

Isolating the option on the same tree, same file:

```
[set -euo pipefail] hit=no
[set -eu]           hit=yes
[set -o pipefail]   hit=no
[]                  hit=yes
```

### It was flaky, not just wrong

Five consecutive runs of the guard as pushed, over this repository:

```
run 1: stella-candidates, stella-witness
run 2: stella-witness
run 3: stella-candidates, stella-witness
run 4: stella-witness
run 5: (passed — nothing reported)
```

Run 5 is the one worth noting: the same tree that reddened CI here can also come back green. So the guard could have merged clean and reported a phantom ungraded plugin on somebody else's PR later, which is the failure mode a canary must not have.

### The fix

The strip lands in a variable and `case` does the search, so no pipe carries the result. Ten consecutive runs now exit 0, and the self-test passes 10/10 with every failure direction (`O1`–`O3`, `E1`–`E3`) still failing as designed — the guard can still fail.

### DoD

I also verified #6148's five DoD items against this branch and ticked them; evidence is on [#6148](https://github.com/macanderson/stella/issues/6148). They were all satisfied by this PR already — unticked boxes were the only thing holding `dod / dod` red.

`gate steps no other workflow runs` is still red here on `check-prose`, but that is `main`'s break (#6196) inherited, not this PR's: it clears when #6203 lands.

---

macanderson:
CI round 1: every check green except `fmt + clippy + test`, which failed on `stella-fleet`'s `a_cancelled_dispatch_releases_its_claims_and_control_handle` ("the parked worker is registered", [job](https://github.com/macanderson/stella/actions/runs/34023646428/job/101460856500)).

This diff holds no Rust — it is an ADR, `plugins/README.md`, two shell scripts, and their wiring into `Makefile`/CI. The same test is green on `main` at `b09fa68a1` and `9c4a2d6d1`. The test drives the dispatch future with a single `now_or_never()` poll and then asserts the control handle is registered, which holds only while every await before the park returns `Ready` first time; filed as #6231.

Retriggering by updating the branch onto the current `main`.

Issue #6202, opened 2026-09-06
A plugin lane cannot say how it is resumed, so nothing can refuse one that does not


macanderson:
## Problem

`stella_protocol::ResumeAuthority` now exists and every `BuiltinLane` declares
one (`BuiltinLane::resume_authority`). The other half of
`doc:turn-lane-assembly` §6 has nothing behind it:

> A plugin lane declares its own authority in its manifest (§9) rather than
> inheriting one by default — an undeclared authority is a load-time rejection,
> not a silent `Redispatch`.

No manifest field reads a `ResumeAuthority`, and nothing rejects a plugin lane
that declares none. `git grep ResumeAuthority crates/stella-plugin` returns
nothing. So a plugin lane today has no declared authority at all, which is the
silent default the design refuses.

## Files

- `crates/stella-protocol/src/lane.rs` — `ResumeAuthority` and `LaneId`. The
  type lives here so `stella-plugin` can reach it: `stella-protocol` is that
  crate's only workspace dependency.
- `crates/stella-plugin/src/` — the manifest parser and its validation errors.
  A lane a manifest contributes should carry an authority, and a missing one
  should be a named parse error rather than a default.
- `crates/stella-cli/src/plugin_cmd.rs` — where an installed manifest is
  resolved, and where the rejection surfaces to an operator.
- `docs/spec/turn-lane-assembly.md` §6 and §9 — the approved design.

## Verify

`cargo test -p stella-plugin` and `cargo test -p stella-protocol lane`.

## Definition of done

- [ ] A plugin manifest can declare a lane's `ResumeAuthority`.
- [ ] A manifest that contributes a lane and declares no authority is refused at
      load time, with an error that names the field and the accepted values.
- [ ] A witness test per direction: a manifest declaring one loads, a manifest
      omitting one is refused. Both fail on the tree before the change.
- [ ] The refusal reaches an operator through `stella plugin install`, not only
      through a library error.
- [ ] No ceiling in `scripts/file-size-baseline.txt` was raised.

Refs #6153, #3274

macanderson:
Triage: P2, area:plugin, area:protocol, use-model:pro, pain:accuracy, pain:auditability — plugin lanes silently default to no declared resume authority instead of being rejected at load time, contradicting the approved turn-lane-assembly design.

Issue #6201, opened 2026-09-06
A dispatched sub-agent leaves its parent no record of a turn that died


macanderson:
## Problem

`BuiltinLane::SubagentFork` is a `ResumeAuthority::Parent` lane
(`crates/stella-protocol/src/lane.rs`). Under that authority the parent reads
what the child left in order to report it, so the child owes a **terminal
frame**: the last committed step plus the transcript that reached it.

A forked child leaves nothing. `Engine::run_sub_agent`
(`crates/stella-core/src/subagent.rs`) strips the checkpoint sink from the
child config it builds, so the child holds no durable record at all. Its work
reaches the parent only as the tool result the dispatch returns. A child killed
mid-way hands back an error string and its transcript is gone — the same hole
the deck lane had before #3233 and the fleet attempt had before #3232.

`crates/stella-parity/src/lane.rs` records this today as
`FrameObligation::Unframed`, citing this issue, so the gap is written down
rather than silent.

## Files

- `crates/stella-core/src/subagent.rs` — where the child's sink is stripped, and
  why (`crates/stella-cli/src/durability.rs` § "What must NOT inherit a
  binding": a child that inherits the parent's handle is a second writer of one
  resume point, not a second resume point).
- `crates/stella-cli/src/lane_frame.rs` — `LaneRecorder`, the shared writer. It
  needs a `SessionDurability` bound to a key of the child's own, which is the
  part a fork has no equivalent of yet.
- `crates/stella-cli/src/subsession.rs`, `crates/stella-cli/src/fleet_cmd/durability.rs`
  — the two lanes that already do this, and the two key shapes to copy from
  (`subsession::lane_journal_key`, `fleet_cmd::durability::attempt_journal_key`).
- `crates/stella-parity/src/lane.rs` — the row to move from `Unframed` to
  `Frames` once a writer and a reader exist.

## What is hard about it

A fork has no identity a later process can name. The deck lane keys on
`(session, lane)` and a fleet attempt keys on `{run_id}/{task_id}`; a
`delegate` child opens no execution row at all and is attributed by
`sub_agent_id` on its events (AGENTS.md's glossary). Deciding that key is most
of this issue.

Also decide whether the reader is the tool result the parent already gets, or a
separate report line. If the tool result is enough, the honest answer may be to
keep the row `Unframed` and rewrite the reason as a decision rather than a gap —
say so in writing either way.

## Verify

[...]

Issue #3274, opened 2026-08-14
Epic: turn setup is assembled in seven different places — unify it into one


macanderson:
## Problem

Stella has exactly **one** turn loop — `crates/stella-core/src/driver/drive.rs:111` is the only `loop {` over `Engine::run_step` in the workspace, and a sweep for `run_step(` outside `driver/` returns one hit that is a doc comment. That part of the architecture is correct and does not need touching.

What is duplicated is **lane assembly**: the act of deciding which of the engine's twelve optional capability slots a given run binds. That decision is made by hand at seven production sites, in three unrelated vocabularies, with nothing checking the result.

**Vocabulary A** — `EngineConfig` fields: `checkpoint_sink`, `turn_halt`, `completion_gate`, `turn_budget`.
**Vocabulary B** — `Engine::with_*` builders: `hooks`, `hook_approvals`, `calibration`, `gate`, `steering`, `bus`, `outcomes`, `fallback`.
**Vocabulary C** — `TurnControls` published to the `ToolRegistry`: the pause gate *again*, so dispatched sub-agents inherit it.

The gate must therefore be bound **twice**, through two mechanisms, at every site that wants it (`fleet_cmd.rs:882-887`, `subsession.rs:749-758`, `command_deck.rs:4244-4252`). A site that does one and forgets the other is silently half-wired.

`crates/stella-core/src/subagent.rs:61` documents the structural consequence in its own module doc: `Engine::with_sleeper` cannot carry `gate`/`steering`/`hooks`, so a fork drops three capabilities by constructor signature rather than by decision.

## The seven lanes and their sites

| Lane | Assembly site |
|---|---|
| Lead (interactive mode) | `stella-cli/src/command_deck.rs:4244` |
| Resume | `stella-cli/src/agent/resume.rs:311` |
| Sub-session | `stella-cli/src/subsession.rs:729` |
| Subagent fork | `stella-core/src/subagent.rs:705` |
| Fleet worker (raw) | `stella-cli/src/fleet_cmd.rs:914` |
| Pipeline stage | `stella-pipeline/src/pipeline/execute_stage.rs:363`, `witness_stage.rs:361` |
| Serve session | `stella-serve/src/session.rs:675` |

Seven is the count of *builtin* lanes. It is deliberately not the count of lanes — see "Plugin lanes are not a later phase" below.

## Why this is the right framing

[...]

PR #6208, opened 2026-09-06
feat(stella-protocol,stella-parity,stella-cli): say how each lane is resumed, and make a redispatch lane leave evidence

### LINKED ISSUE #6153: No lane declares how it is resumed, so half the terminal-frame design is prose only
## Problem

How a dead turn is picked back up is decided per site, and no lane says which
answer it holds.

`doc:turn-lane-assembly` §6 models it as a property of the lane, approved by
Mac on 2026-08-14:

- `Own` — the lane resumes itself (Lead, Resume, ServeSession)
- `Parent` — the parent reads the checkpoint to *report*, never to re-enter
  (SubSession, SubagentFork)
- `Redispatch` — the supervisor re-runs the unit and the record is evidence
  (FleetWorker, and any staged lane)

Half of that shipped. `crates/stella-store/src/lane_frame.rs` holds the
terminal frame — one record written when a lane dies, not one per step —
with `record` / `read` / `clear`; `crates/stella-cli/src/durability.rs`
(`record_terminal_frame`, `terminal_frame`, `clear_terminal_frame`) and
`crates/stella-cli/src/subsession/terminal_frame.rs` (`LaneRecorder`,
`parent_report`) wire it for the deck's worker lane, and
`crates/stella-cli/src/command_deck/driver_support.rs` reads it back.

What did not ship is the vocabulary. `git grep ResumeAuthority crates/`
returns only prose — the doc comments in `lane_frame.rs` and
`terminal_frame.rs` that name a type the tree does not have. So:

- No lane declares its authority; a reader has to infer it from which module
  happens to call `record_terminal_frame`.
- Only the `Parent` lane writes a frame. A `Redispatch` lane leaves nothing a
  supervisor can read as evidence, and no row says whether that is a decision
  or a gap — which is the whole disease #3274 is about.
- A plugin lane has nothing to declare in a manifest, so #3274's rule that an
  undeclared authority is a load-time rejection has no type to reject against.

#3232 and #3233 both closed, through per-lane durability rather than through
this model, so this is not a durability bug — it is the missing declaration.

## Files

- `docs/spec/turn-lane-assembly.md` §6 — the approved design.
- `crates/stella-store/src/lane_frame.rs` — the terminal frame.
- `crates/stella-cli/src/durability.rs`,
  `crates/stella-cli/src/subsession/terminal_frame.rs` — today's only writer.
- `crates/stella-cli/src/fleet_cmd.rs` (`attempt_durability`) — the
  `Redispatch` lane that writes no frame.
- `crates/stella-protocol/src/lane.rs` — where `BuiltinLane` lives; the
  authority is a property of the lane, so decide whether the enum lives beside
  it or in `stella-parity`'s matrix.
- `crates/stella-parity/src/lane.rs` — the matrix that should carry the column.

## Verify

`cargo test -p stella-parity lane` and `cargo test -p stella-cli terminal_frame`.
A witness per authority: a `Parent` lane that dies leaves a frame its parent
reads (this exists today for the sub-session), and a `Redispatch` lane that
dies leaves one its supervisor reads — that second one fails on the tree as it
stands.

## Definition of done

- [x] A `ResumeAuthority` enum (`Own` / `Parent` / `Redispatch`) exists in the
      tree, with a doc comment saying what each arm obliges a lane to write.
- [x] Every `BuiltinLane` declares one, exhaustively, so an eighth lane cannot
      be added without answering.
- [x] A `Redispatch` lane writes a terminal frame, with a reader named, or the
      row says in writing why it does not.
- [x] The prose in `lane_frame.rs` and `subsession/terminal_frame.rs` cites the
      real type instead of a design-doc name.
- [x] A witness per authority arm, failing on the tree before the change.
- [x] No ceiling in `scripts/file-size-baseline.txt` was raised.

Refs #3274, #3232, #3233

macanderson:
## What and why

`ResumeAuthority` existed only in `doc:turn-lane-assembly` §6 and in two doc
comments that named a type the tree did not have. So no lane declared how it is
picked back up, a reader inferred it from which module happened to call
`record_terminal_frame`, and the fleet's `Redispatch` lane left a supervisor
nothing to read.

**`stella-protocol`** gains `ResumeAuthority { Own | Parent | Redispatch }` and
`BuiltinLane::resume_authority()`. The match has no `_` arm, so a tenth lane
does not compile until somebody answers. It lives there rather than in
`stella-parity` because a plugin manifest declares one (§9) and the manifest
parser is `stella-plugin`, whose only workspace dependency is `stella-protocol`;
putting the enum in the parity matrix would point the plugin loader at the
declaration table.

**`stella-parity`'s lane matrix** gains the column. `FrameObligation` says what
each lane's authority obliges it to leave — `ResumesItself`, `Frames { writer,
reader, witness }`, or `Unframed { reason }` — and the authority itself is not
stored: a test reads it back out of `BuiltinLane::resume_authority()` and fails
when a row and the type disagree. A `Frames` row's writer must really name
`LaneRecorder::new`, its reader must really name `TerminalFrame::read`, and its
witness must exist. Same instrument as the two matrices beside it.

**The `Redispatch` lane now writes a frame.** The recorder that keeps what
`Engine::drive` discards was not specific to a sub-session, so its generic half
moves from `crates/stella-cli/src/subsession/terminal_frame.rs` into
`crates/stella-cli/src/lane_frame.rs`, taking a `LaneEnd` instead of
`subsession::WorkerEnd`. `subsession/terminal_frame.rs` keeps its own reader
(`parent_report`) and maps `WorkerEnd` onto `LaneEnd`, so every existing test
and every path citation stays where it is.
`fleet_cmd/durability.rs` holds a recorder over the attempt's own handle, wraps
the attempt's engine config with it, and settles it once at the end of
`run_task`.

**The reader is the re-dispatch.** `initial_messages` restored from the
checkpoint alone, and `Engine::drive` retires the checkpoint on every terminal
path it reaches — so only a killed *process* left one behind, and an attempt
that *aborted* at step 40 left nothing at all. It now falls back to the frame's
transcript. §6 calls a `Redispatch` lane's record "evidence, not a resume
point"; this tree re-enters it, because #3232 already settled that a
re-dispatch lands on the same key and picks the interrupted transcript back up
— reading one of an attempt's two records and merely reporting the other would
give one lane two meanings for one fact. The reasoning is in
`AttemptDurability::initial_messages`'s own doc, not only here.

## Witness mechanism

One per authority arm, each failing on the tree before this change.

- **`Redispatch`** — `a_redispatched_attempt_re_enters_the_transcript_the_engine_discarded`
  (`crates/stella-cli/src/fleet_cmd/durability.rs`). It persists a checkpoint at
  step 7 and then calls `discard()`, which is exactly what the engine does on
  every terminal path. On the old code `initial_messages` then returns the
  two-message fresh opening, because the only record it consulted is gone. It
  now returns the three-message transcript out of the frame, with a note naming
  the step and how the prior attempt ended. Two more beside it:
  `a_settled_attempt_retires_the_frame_an_earlier_try_left` (the other half of
  the deal) and `a_stopped_attempt_frames_without_reading_as_a_failure`.
- **`Parent`** — `a_lane_that_failed_leaves_its_parent_a_frame_naming_the_last_committed_step`,
  which already shipped and is now named by the matrix row instead of being a
  fact only its own file knew.
- **`Own`** — `every_lane_declares_who_resumes_it`
  (`crates/stella-protocol/src/lane.rs`) and
  `every_lane_declares_the_frame_its_authority_obliges`
  (`crates/stella-parity/src/lane.rs`). Neither can compile before
  `ResumeAuthority` exists, which is what "no lane declares its authority"
  means: there was no type for a self-resuming lane to say it owes nothing
  with.

## Exemplar

`crates/stella-model/src/provider_parity.rs` and the two matrices already in
`stella-parity` — a row per case, a posture, a named witness checked for
existence, and completeness enforced by an exhaustive match with no `_` arm.

## CI note

`check-prose` fails on `crates/stella-serve/src/observe/event.rs` and
`docs/spec/wrapper-socket.md`. Neither is in this diff. Both fail on
`origin/main` at 321b86ede with this branch stashed, and #6196 (`main is red:
prose fails on the merged tree`) is open with a live repair claim on it — so
this PR waits on that rather than racing it. My own files pass the same guard:
after the last edit `check-prose` names only those two.

## Tests deleted

None. The four tests in `subsession/terminal_frame.rs` keep their names and
their file; only the machinery under them moved.

## Remaining work filed

- #6201 — a dispatched sub-agent (`SubagentFork`, a `Parent` lane) writes no
  frame: `Engine::run_sub_agent` strips the checkpoint sink, so a fork holds no
  durable record and has no key to bind one under. The matrix row records this
  as `Unframed` citing that issue rather than reading as a decision.
- #6202 — a plugin manifest still cannot declare a `ResumeAuthority`, so §9's
  load-time rejection has a type to reject against but no code that does it.

Closes #6153
Refs #3274

## Summary by Sourcery

Declare how every built-in lane resumes and preserve terminal evidence so parent and redispatch readers can recover the work left by interrupted attempts.

New Features:
- Add explicit resume authorities for built-in lanes and expose their terminal-frame obligations through the parity matrix.
- Persist fleet attempt transcripts as terminal frames and re-enter them on redispatch when checkpoints have been retired.

Bug Fixes:
- Preserve failed and stopped lane transcripts after engine terminal paths discard their checkpoints, preventing interrupted work from being lost.

Enhancements:
- Share terminal-frame recording between parent-owned sub-sessions and redispatched fleet attempts, with exhaustive consistency and witness checks for lane metadata.

Tests:
- Add coverage for redispatch recovery, successful-attempt frame retirement, stopped-attempt framing, resume-authority mappings, and lane frame obligations.


macanderson:
## CI: three red checks, all inherited from `main`

`fmt + clippy + test` reports **one** failure across the whole workspace, and it
is not in this diff:

```
lane::capability::tests::a_seam_claim_agrees_with_its_lane_literal ... FAILED
`serve_session` claims it does not bind `requery`, and
crates/stella-serve/src/session.rs does not write `requery: None`
```

`crates/stella-parity/src/lane/capability.rs`, its `tests.rs`, and
`crates/stella-serve/src/session.rs` are all absent from this branch's diff, so
that assertion is reading code identical to `main`'s. `session.rs` builds a
`RemoteRequery` and passes `requery` into its `TurnCapabilities`; the row still
claims `SeamClaim::deferred`. #6197 and #6200 are the repairs.

`gate steps no other workflow runs` is `check-prose`, on
`crates/stella-cli/src/fleet_cmd.rs [issue-reference]: 53 allowed, 54 found` —
the bare issue number #6168 added. #6196 tracks it and #6203 repairs it. My own
comment in that file carries no issue number, and `check-prose` names only the
two inherited files when run against this branch.

`main is not known-broken` is #6196 being open.

Everything this PR is responsible for is green: `cargo fmt --check`, `cargo
clippy -D warnings`, `cargo doc -D warnings`, `docs/wire`, every guard, and
every witness this change adds:

```
lane::tests::every_lane_declares_the_frame_its_authority_obliges ... ok
lane::tests::every_framed_lane_writes_and_is_read ... ok
lane::tests::an_unframed_lane_says_why_and_a_lane_with_no_producer_is_one ... ok
fleet_cmd::durability::tests::a_redispatched_attempt_re_enters_the_transcript_the_engine_discarded ... ok
fleet_cmd::durability::tests::a_settled_attempt_retires_the_frame_an_earlier_try_left ... ok
fleet_cmd::durability::tests::a_stopped_attempt_frames_without_reading_as_a_failure ... ok
subsession::terminal_frame::tests::a_lane_that_failed_leaves_its_parent_a_frame_naming_the_last_committed_step ... ok
subsession::terminal_frame::tests::a_stopped_lane_frames_the_point_its_dropped_turn_left_standing ... ok
subsession::terminal_frame::tests::a_lane_that_completed_leaves_no_frame_and_retires_an_older_one ... ok
subsession::terminal_frame::tests::a_lane_that_died_before_its_first_step_leaves_nothing_to_report ... ok
```

I am waiting on those repairs rather than racing them, and will re-target this
branch and re-run once they land.

PR #6220, opened 2026-09-06
fix(stella-core): charge the steering allowance one turn at a time

### LINKED ISSUE #6186: Re-settling the steering allowance subtracts the whole session's recall spend, so raising the allowance can withhold every tool
## Problem

`SteeringLedger::settle` memoizes on `declared`, so the ordinary path is
correct and the advertised tool array is stable (invariant 7). A **changed**
`declared` recomputes, and that recomputation reads a counter that has been
accumulating since session start:

```rust
// crates/stella-core/src/steering/ledger.rs, SteeringLedger::settle
let max_tokens = declared.max_tokens.saturating_sub(self.spent());
```

`spent` is a saturating add and the type exposes no reset — `spend`, `spent`
and `settle` are the whole surface, and nothing stores 0. The handle rides
`Config` for the session and is `Posture::StartupOnly` in the reload
completeness table, so `spent()` is the sum of every distinct recall block the
session has injected, not the current turn's.

`ToolAdvertisement` is `Posture::Reloaded`, so `/reload` is a live path to the
re-settle. On it, the operator is charged for the session's whole history:

- 20 turns of ~500 tokens of distinct recall → `spent() == 10_000`
- operator raises `context.steering.max_tokens` from 8_000 to 12_000, `/reload`
- `settle(12_000)` → `12_000 - 10_000` = **2_000**

They raised the allowance and the tool array shrank. Far enough in,
`saturating_sub` floors at 0; `mcp_max_tokens` is `.min(max_tokens)` and goes
to 0 with it, so the session advertises no tools until it restarts. The drop
notice added in #6168 names every withheld tool, so it fails loudly rather
than silently — but the direction is the opposite of the operator's edit.

## Why the existing test does not catch it

`a_changed_allowance_settles_again` spends **once** (400 tokens) before
re-settling, which is the one shape in which a session-cumulative counter and
a per-turn one give the same answer. The accumulation hides behind it.

## Files

- `crates/stella-core/src/steering/ledger.rs` — `spend`, `spent`, `settle`,
  and the tests module.
- `crates/stella-cli/src/memory/recall.rs` — `inject_opening_recall`, the only
  production caller of `spend`; it runs per turn.
- `crates/stella-cli/src/config/reload/completeness.rs` — the `StartupOnly`
  note on `steering_ledger` that argues for preserving the cell across a
  reload, which is the constraint any fix has to keep.
- The six drivers that call `inject_opening_recall`, if the fix takes a turn
  boundary: `agent.rs`, `agent/goal.rs`, `command_deck.rs`,
  `fleet_cmd/steering.rs`.

## The decision a fix has to make

#6110's DoD box 1 asks for "one **per-turn** allowance"; its box 2 and its
invariant-7 constraint ask for a tool array "settled once per session". Both
are wanted and one counter cannot be both, which is what this defect is made
of. Two shapes:

1. **Record the current block's cost rather than a running total** — `spend`
   stores instead of adding, and `settle` subtracts that. Steady state is
   unchanged (the first settle still sees turn 1's block and is cached), and a
   re-settle then charges one turn. Smallest change; touches only `ledger.rs`.
2. **A real turn boundary** — an `open_turn()` that resets `spent` and leaves
   `settled` alone, threaded through the six drivers above. The fuller reading
   of #6110, and more threading.

Either keeps `settled` session-scoped, which is what invariant 7 needs.

## Verify

Extend `a_changed_allowance_settles_again` to the multi-turn shape: settle,
`spend` across several simulated turns, then re-settle with a **larger**
`declared`, and assert the allowance grew rather than shrank. That test fails
on today's code and passes on either fix.

## Definition of done

- [x] A re-settle after a changed `declared` charges one turn's block, not the
      session's accumulated total.
- [x] The advertised array is still settled once per unchanged `declared`, so
      `a_later_spend_does_not_move_a_settled_allowance` still passes.
- [x] A regression test drives several turns of spend before a re-settle with a
      raised allowance and asserts the result grew. It fails on the old code.
- [x] `spent`'s doc comment describes the scope the code actually has.
- [x] #6110's box 1 wording and the shipped scope agree, either by the fix or
      by rewording the box.
- [x] The fix stays in `crates/stella-core/src/steering/ledger.rs` (pure
      arithmetic, no I/O — the crate AGENTS.md's routing table already puts
      this kind of change in per invariant 2) and, if shape 2 is taken, its
      driver threading in the four `stella-cli` call sites; no line lands on
      any file in `scripts/file-size-baseline.txt`.
- [x] Comments/prose added by the fix pass `make prose`/`make line-citations`
      — cite code by symbol (e.g. `SteeringLedger::settle`), never a pinned
      line number.
- [x] Not applicable: no UI/design-system surface — this is `stella-core`
      arithmetic plus `stella-cli` driver threading, with no `stella-tui`/
      `stella-tui-theme` rendering involved.

Refs #6110, #6168

---

### Verification note — autonomous PR sweep, 2026-09-06

All eight boxes ticked after reading PR #6220's diff against each item. The fix
takes **shape 2**, a real turn boundary (`SteeringLedger::open_turn`), but
threads it at **one** seam rather than the four drivers this issue anticipated:
`inject_opening_recall` is the seam every driver's opening block already passes
through, which `every_recalling_driver_routes_its_block_through_the_opening_seam`
holds. That is a smaller and better-held change than the issue expected.

- **A re-settle charges one turn** — `open_turn` stores 0 into `spent` and
  leaves `settled` alone. The witness
  `a_raised_allowance_is_charged_one_turn_not_the_session` runs 20 turns of 500
  and asserts `settle(12_000).max_tokens == 11_500`; on the old code `spent()`
  would be 10_000 and the answer 2_000, which is this issue's exact worked
  example.
- **Settled-once still holds** — `settle`'s memoization is untouched, and the
  existing `a_changed_allowance_settles_again` never calls `open_turn`, so its
  `settle(declared(2_000)).max_tokens == 1_600` assertion reads the same 400
  it always did. `a_new_turn_clears_the_spend_and_keeps_the_settled_answer` and
  the CLI-side `a_later_turn_does_not_move_an_array_the_session_settled` both
  assert the settled answer survives a new turn.
- **Regression test fails on the old code** — by construction: `open_turn` does
  not exist there, so the test cannot compile against it, and the behaviour it
  asserts has no other path.
- **`spent`'s doc matches the code** — now "What the open turn's block has
  spent so far. `Self::open_turn` puts it back to zero, so this is never a
  session total."
- **#6110 box 1 agreement** — I read #6110: box 1 is "One **per-turn**
  allowance is spent by both consumers…", ticked and closed by #6168. The
  shipped scope was session-cumulative, which is this defect; the fix makes it
  genuinely per-turn, so the two now agree by the fix rather than by a
  rewording.
- **Routing and no god-file growth** — arithmetic in `stella-core`'s
  `ledger.rs`, threading in `stella-cli`. None of the touched files
  (`ledger.rs`, `recall.rs`, `config.rs`, `completeness.rs`, `memory/tests.rs`
  and the new `memory/tests/steering_ledger_turn.rs`) carries a
  `scripts/file-size-baseline.txt` entry — `stella-core`'s god files are
  `driver/tests.rs`, `driver.rs` and `bus.rs`, and `stella-cli`'s is
  `command_deck.rs`.
- **Prose** — new comments cite `SteeringLedger::open_turn` and `Self::settle`
  by symbol, no pinned line numbers.

Two details worth recording because neither is a DoD item and both are the kind
of thing that gets lost:

`open_turn` is a **reset rather than a store**, and `two_charges_in_one_turn_add_up`
pins that. Shape 1 in this issue proposed a store; a store would keep the last
charge of a turn and silently drop the rest of the block, so the shipped choice
is the safer of the two the issue offered.

**One edge I could not settle from the diff, offered to the author rather than
filed:** `open_turn` fires inside `inject_opening_recall`, so a turn opens only
where a block is injected. If any driver skips that call entirely on a turn
with nothing to recall — as opposed to calling it with an empty block — that
turn would inherit the previous turn's `spent` rather than starting at zero. It
is benign in the accumulating direction this issue is about (nothing new is
charged), and the prior block is arguably still in the model's context, so it
may well be intended. Worth one look at the call sites to confirm it is a
decision rather than an accident.

`cargo deny` and CodeQL are green on #6220; `fmt + clippy + test` was still
running when this was written, and it is what confirms the two prose ratchets,
the file-size ratchet and the untouched neighbouring tests. `dod / dod` was red
only because these boxes were unticked.

macanderson:
## What and why

`SteeringLedger::spent` was a running total for the whole session. `settle`
subtracts it, and re-settles whenever the declared allowance changes — which
`/reload` is a live path to, because `tool_advertisement` is `Posture::Reloaded`
while the ledger cell itself is `StartupOnly`.

So an operator who raised `context.steering.max_tokens` after twenty turns of
recall was charged for all twenty turns at once:

- 20 turns of ~500 tokens of distinct recall leaves `spent() == 10_000`
- raise 8_000 to 12_000, `/reload`
- `settle(12_000)` answers `12_000 - 10_000` = **2_000**

They raised the allowance and the tool array shrank. Far enough in,
`saturating_sub` floors at 0, `mcp_max_tokens` follows it down, and the session
advertises no tools until it restarts.

Every doc around this cell already said the allowance was one turn's —
`ToolBudget`'s ("the allowance the whole volatile block shares"),
`ToolAdvertisement::Lean`'s ("what the turn's volatile block left of it"),
`Config::steering_ledger`'s ("what this turn's block spent"), and #6110's own
first DoD box ("one **per-turn** allowance"). The accumulating counter was the
one thing that disagreed.

## The fix

`SteeringLedger::open_turn` stores 0 in `spent` and leaves `settled` untouched.
`memory::inject_opening_recall` calls it before it injects.

Two choices worth naming, both from the issue's own "the decision a fix has to
make":

1. **A reset, not a store.** The issue's smallest shape has `spend` store rather
   than add. That is correct for today's single charger and silently wrong for
   the second one: a store keeps the last charge and drops the rest of the block
   with nothing to notice it. A reset composes — `two_charges_in_one_turn_add_up`
   is the test that pins it.
2. **The seam, not the six drivers.** The issue's fuller shape threads
   `open_turn()` through `agent.rs`, `agent/goal.rs`, `command_deck.rs` and
   `fleet_cmd/steering.rs`. It does not need to: `memory.rs` already states that
   every driver reaches a turn through `inject_opening_recall`, and
   `memory/tests/skill_event.rs` scans those driver sources to enforce it. So
   that function *is* the turn boundary, and a driver cannot route around it
   without failing an existing test. Threading four call sites would add four
   places to forget and hold nothing the seam does not already hold.

`settled` stays session-scoped, so the advertised array is still settled once and
still holds still inside a turn (the byte-stable prompt rule), and the
`StartupOnly` note on `steering_ledger` still holds — a reload finds the settled
array waiting, and only the turn figure is fresh.

## Witness mechanism

`a_raised_allowance_is_charged_one_turn_not_the_session`
(`crates/stella-core/src/steering/ledger.rs`) drives twenty turns of a 500-token
block, then settles at 8_000, opens a twenty-first turn, and re-settles at
12_000. It asserts 11_500.

It fails on the old code by arithmetic, not by accident: with the spend carried
across turns, `spent()` at the first settle is 10_000 (so the answer is 0, not
7_500) and 10_500 at the re-settle (so the answer is 1_500, not 11_500). The
existing `a_changed_allowance_settles_again` could not see this because it spends
exactly once before re-settling — the one shape where a per-turn counter and a
session-cumulative one give the same number.

`crates/stella-cli/src/memory/tests/steering_ledger_turn.rs` is the other half:
it proves the seam actually opens the turn. Two blocks through
`inject_opening_recall` on one ledger, and the spend is the second block's cost
alone. Without the `open_turn()` call the ledger arithmetic is right and nothing
ever resets it, which no `stella-core` test can see.

`a_later_turn_does_not_move_an_array_the_session_settled` and
`a_new_turn_clears_the_spend_and_keeps_the_settled_answer` hold the other
direction: opening a turn must not move the settled array.

## Prose kept in step

Three comments described the old scope and now describe the new one:
`Config::steering_ledger`, the `StartupOnly` note in
`config/reload/completeness.rs`, and `inject_opening_recall`'s paragraph on when
a deduped block is charged. That last one said the block was "one an earlier turn
paid for, and spent here", which the reset makes false — no earlier turn's figure
survives on the ledger. What is true is simpler: the model already carries the
block, so showing it again costs this turn nothing.

Two ratchets were paid rather than baselined. The new test file read at grade
6.48 against the 6.00 a new file is held to, and the header section added to
`ledger.rs` took `stella-core`'s mean header length to 30.02 against 29.97. Both
were fixed by cutting words, not by editing a baseline. `python3
scripts/check-prose.py` exits 0 on this tree; `make guards-fast` is green.

## Files

- `crates/stella-core/src/steering/ledger.rs` — `open_turn`, the doc scope, four tests.
- `crates/stella-cli/src/memory/recall.rs` — the `open_turn()` call at the seam.
- `crates/stella-cli/src/memory/tests/steering_ledger_turn.rs` — new, the seam's witness.
- `crates/stella-cli/src/config.rs`, `crates/stella-cli/src/config/reload/completeness.rs` — the prose.

No new file lands in `scripts/file-size-baseline.txt`; nothing here is a god
file. Pure arithmetic in `stella-core` (invariant 2), the I/O-side call in
`stella-cli`.

## Base

Cut from `9136a2092`, the `main` whose `fleet_cmd.rs` issue-reference count had
`check-prose` red (#6196). That repair landed in #6203, so `main` is merged in
here rather than the fix being repeated. The merge conflicted in
`inject_opening_recall`'s doc block, which both changes rewrote; the resolution
keeps #6203's plain-language sentences and corrects the one that the reset makes
untrue.

Tests deleted: None.

Residue issues filed: none — nothing was noticed and left unfixed.

Closes #6186

### DIFF
--- a/crates/stella-cli/src/config.rs
+++ b/crates/stella-cli/src/config.rs
@@ -396,9 +396,12 @@ pub struct Config {
     ///
     /// A shared cell, not a resolved value. Its two users meet at different
     /// times. The recall block is written first, and
-    /// `memory::inject_opening_recall` spends here. The tool list is built
-    /// after, and `agent::tool_stack` reads here. One number, so a rule and a
-    /// tool schema can be weighed against each other.
+    /// `memory::inject_opening_recall` opens the turn and spends here. The
+    /// tool list is built after, and `agent::tool_stack` reads here. One
+    /// number, so a rule and a tool schema can be weighed against each other.
+    ///
+    /// The spend is one turn's and the settled array is the session's, which
+    /// is why the cell holds both rather than being rebuilt per turn.
     ///
     /// It rides `Config` for the reason [`Self::output_ceilings`] does: every
     /// driver already hands a `&Config` to both call sites, so no one has to
--- a/crates/stella-cli/src/config/reload/completeness.rs
+++ b/crates/stella-cli/src/config/reload/completeness.rs
@@ -209,11 +209,11 @@ fn ledger(before: &Config, after: &Config) -> Vec<Field> {
         Field {
             name: "steering_ledger",
             posture: Posture::StartupOnly(
-                "a shared cell holding what this session's volatile block already spent and the \
-                 tool array that spend settled; a fresh one would re-rank the advertised tools \
+                "a shared cell holding what the open turn's volatile block spent and the tool \
+                 array that spend settled; a fresh one would re-rank the advertised tools \
                  mid-session, and that array is cache prefix. An operator who edits the \
                  allowance is served by `tool_advertisement` above: the ledger settles again \
-                 whenever the declared allowance changes",
+                 whenever the declared allowance changes, against the open turn's spend alone",
             ),
             // A handle, not a value. `SteeringLedger` has no equality, and
             // what it holds is spend rather than settings.
--- a/crates/stella-cli/src/memory/recall.rs
+++ b/crates/stella-cli/src/memory/recall.rs
@@ -204,7 +204,7 @@ impl OpeningRecall {
 /// shares with the tool array. The cost is measured over the injected bytes,
 /// so what the plane records and what the provider is sent are one number. It
 /// is charged only when the block is really appended. A block the dedup
-/// refuses is one an earlier turn paid for, and spent here.
+/// refuses costs this turn nothing, since the model already carries it.
 ///
 /// **This is the turn boundary the allowance is measured on.** The allowance
 /// is what one turn's volatile block may take. Every driver opens a turn
--- a/crates/stella-cli/src/memory/recall.rs
+++ b/crates/stella-cli/src/memory/recall.rs
@@ -204,12 +204,22 @@ impl OpeningRecall {
 /// shares with the tool array. Measured over the injected bytes, so what the
 /// plane records and what the provider is sent are one number; and charged
 /// only when the block is genuinely appended, since a block the dedup refuses
-/// is one an earlier turn already paid for and already spent here.
+/// is one the model is already carrying in the cached prefix, and this turn
+/// pays nothing to show it again.
+///
+/// **This is the turn boundary the allowance is measured on.** That
+/// allowance is what one turn's volatile block may take, and this is the one
+/// seam every driver reaches a turn through, so
+/// [`stella_core::steering::ledger::SteeringLedger::open_turn`] belongs here
+/// rather than threaded through each of them. A ledger that added every turn
+/// up instead charged an operator who raised the allowance for the whole
+/// session, and far enough in advertised no tools at all.
 pub(crate) fn inject_opening_recall(
     messages: &mut Vec<CompletionMessage>,
     recalled: RecalledBlock,
     ledger: &stella_core::steering::ledger::SteeringLedger,
 ) -> OpeningRecall {
+    ledger.open_turn();
     let events = recalled.telemetry_events();
     let cost = recalled
         .text
--- a/crates/stella-cli/src/memory/tests.rs
+++ b/crates/stella-cli/src/memory/tests.rs
@@ -38,6 +38,9 @@ mod skill_creation;
 // #5031: whether a turn says which skills it injected. `golden_block` pins
 // what the block contains; this asks what the stream reports about it.
 mod skill_event;
+// The injection is where a turn opens on the steering ledger, so the
+// allowance a re-settle reads is one turn's block and not the session's sum.
+mod steering_ledger_turn;
 // #3243 D1/D2: what selection is allowed to call "relevant to this turn" —
 // per-turn domain scope, and a lexical score that can fire on a real prompt.
 mod steering_selection;
--- a/crates/stella-cli/src/memory/tests/steering_ledger_turn.rs
+++ b/crates/stella-cli/src/memory/tests/steering_ledger_turn.rs
@@ -1,0 +1,84 @@
+//! The steering allowance is one turn's, and this is the seam that says
+//! when a turn starts.
+//!
+//! `stella_core::steering::ledger` holds the arithmetic and tests it there.
+//! This asks the other half of the question. Every driver opens a turn
+//! through `inject_opening_recall`, so that is where the ledger is told a new
+//! turn began. Leave the call out and the arithmetic is still right and
+//! nothing ever resets: the spend grows all session, and an operator who
+//! raises the allowance is charged for every turn behind them.
+
+use stella_core::steering::ledger::SteeringLedger;
+use stella_core::steering::tools::ToolBudget;
+
+use crate::memory::recall::RecalledBlock;
+use crate::memory::{RECALL_MARKER, inject_opening_recall};
+
+/// A block shaped the way a rendered one is — marked, so the injection's own
+/// dedup reads it as a recall block rather than as ordinary talk.
+fn rendered(body: &str) -> String {
+    format!("{RECALL_MARKER}\n{body}")
+}
+
+/// The recall a turn hands the injection. Only the text matters here; the
+/// skills and handles beside it are another seam's subject.
+fn block(text: &str) -> RecalledBlock {
+    RecalledBlock {
+        text: Some(text.to_string()),
+        ..RecalledBlock::default()
+    }
+}
+
+/// **The witness.** Two turns, two distinct blocks, one ledger.
+/// The allowance is charged for the turn that is open, not for both.
+///
+/// Before the turn boundary the spend was a running total, so this read the
+/// sum of the two blocks. Twenty turns in, the sum was larger than the
+/// allowance itself and the tool array emptied.
+#[test]
+fn a_turn_charges_the_allowance_for_its_own_block_alone() {
+    let ledger = SteeringLedger::new();
+    let mut messages = Vec::new();
+
+    let first = rendered("turn one recalled the billing migration window");
+    let second = rendered("turn two recalled the streaming dialect notes, at some length");
+
+    inject_opening_recall(&mut messages, block(&first), &ledger);
+    inject_opening_recall(&mut messages, block(&second), &ledger);
+
+    assert_eq!(
+        ledger.spent(),
+        stella_protocol::estimate_tokens(&second),
+        "the open turn is charged for its own block"
+    );
+    assert!(
+        stella_protocol::estimate_tokens(&first) > 0,
+        "and turn one's block was not free, so the two readings differ"
+    );
+}
+
+/// The array a session settles holds still across turns. It sits ahead of
+/// the prompt in every cache, so re-ranking it bills the whole chat again.
+/// A new turn resets the spend and must not touch the answer.
+#[test]
+fn a_later_turn_does_not_move_an_array_the_session_settled() {
+    let ledger = SteeringLedger::new();
+    let mut messages = Vec::new();
+    let declared = ToolBudget {
+        max_tokens: 8_000,
+        mcp_max_tokens: 4_000,
+    };
+
+    inject_opening_recall(&mut messages, block(&rendered("turn one")), &ledger);
+    let settled = ledger.settle(declared);
+
+    inject_opening_recall(
+        &mut messages,
+        block(&rendered(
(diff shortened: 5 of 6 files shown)


macanderson:
## This now conflicts with `main`, and the resolution has a second gate on it

`main` moved to `b09fa68` (#6203) at 09:04, after this branch was cut. `git merge-tree origin/main` reports one conflict:

```
CONFLICT (content): Merge conflict in crates/stella-cli/src/memory/recall.rs
```

Both sides rewrote the same doc comment on `inject_opening_recall`. This branch replaced it *and* appended the paragraph documenting the turn boundary (`SteeringLedger::open_turn`); #6203 replaced it with a shorter, plainer version. Neither contains the other's lines, which is why git cannot compose them.

**The part that is not obvious from the conflict markers:** #6203 existed because `main` was red on `check-prose`, and `crates/stella-cli/src/memory/recall.rs` was one of the six files it repaired — from reading grade 12.21 down to 12.05, against a ceiling of 12.19. So resolving this by keeping this branch's wording wholesale is likely to put the file back over that ceiling and re-red the gate. The resolution needs to keep the `open_turn` paragraph's *content* while written plainly enough to stay under 12.19.

`python3 scripts/check-prose.py` is the check, and it compiles nothing — it is worth running locally on the resolved tree before pushing.

Per AGENTS.md, merge `main` in rather than rebasing, so anyone's checkout of this branch stays valid.

I have not pushed to this branch: the resolution is a judgement about how to word your own new paragraph, not a mechanical merge, and it is yours to make.

---

macanderson:
**Standing down — you hold a live claim on #6186.** Handing over what I found rather than pushing to your branch.

Automated PR sweep. `./scripts/issue-claim.sh`'s comment on #6186 is timestamped 08:51:00 and lapses after 90 minutes, and this PR gained a commit at 09:12, so a session is on it. I reproduced everything below in a throwaway worktree and pushed nothing.

### 1. The merge conflict is one hunk, and both sides are right

`main` moved to `b09fa68` (#6203 landed, repairing the red `prose` gate). The only conflict is in `crates/stella-cli/src/memory/recall.rs`, in the `inject_opening_recall` doc comment. #6203 shortened those sentences for the reading-grade ratchet; this branch rewrote the same passage for per-turn semantics and added the turn-boundary paragraph.

Taking either side whole loses something — #6203's grade fix or your semantics. What merges cleanly is #6203's short sentences carrying your meaning, keeping your paragraph verbatim:

```rust
/// shares with the tool array. The cost is measured over the injected bytes,
/// so what the plane records and what the provider is sent are one number. It
/// is charged only when the block is really appended. A block the dedup
/// refuses is one the model already carries in its cached prefix, so this turn
/// pays nothing to show it again.
```

`main`'s trailing clause reads "is one an earlier turn paid for, and spent here" — written against the session-total ledger, and wrong once `open_turn()` exists. That is the clause to replace rather than keep.

### 2. `prose` fails twice, and the first failure hides the second

`check-prose.py` reports one category and stops, so fixing the grade uncovers a density failure underneath:

1. **Reading grade** — `crates/stella-cli/src/memory/tests/steering_ledger_turn.rs`, `grade 6.00 allowed, 6.48 found`. Three sentences carry it: the module header's "Leave the call out and…", `rendered`'s "A block shaped the way a rendered one is — marked, so…", and the witness's "Twenty turns in, the sum was larger than…". Splitting each in two clears it.
2. **Header density** — `crates/stella-core: 29.97 allowed, 30.02 mean header lines`, from this branch's additions under `crates/stella-core/src/steering/`. Needs a header trimmed, not a baseline bump.

### 3. One measurement trap, in case you hit it

Running `python3 scripts/check-prose.py` with the merge **unstaged** reports every count for the conflicted file at exactly 3× its baseline (`5 allowed, 15 found`). That is not real: the script enumerates via `git ls-files --cached`, which lists an unmerged path once per stage (base/ours/theirs). `git add` the resolution first, or you will go hunting for prose that was never added.

Everything else on this PR is inherited from the red `main` and clears on its own now that `b09fa68` is in — `main is not known-broken` and the `gate steps` job both re-run green against the repaired base.

---

Issue #6219, opened 2026-09-06
A busy day of closures can hide the ones the loop filed


macanderson:
## The problem

`crates/stella-cli/src/self_driving_cmd/closures.rs`'s `reconcile` asks the
tracker for every issue closed since the last read, in one call bounded by
`WINDOW` rows. GitHub hands those back newest first.

When more issues close in one window than `WINDOW`, the read is a page rather
than the whole window. The older rows in that window are never seen: the run
advances the watermark to now, so the next read starts after them, and the
digests those issues were filed under keep dropping repeats for ever.

The alternative already considered and rejected is worse: keeping the watermark
where it is re-reads the same newest page every cycle and never converges,
because the tracker's order does not change.

A saturated read is recorded in the audit trail ("the window was full at N, so
older closures in it were not read"), so it is visible rather than silent. It
is still wrong.

## What to build

Something that converges. Two shapes worth measuring against each other:

- **Walk forward, oldest first.** GitHub search takes a sort direction
  (`sort:updated-asc` on the `--search` string), so a saturated page could
  advance the watermark to the newest row it actually read rather than to now.
  `updated` is not `closed`, so check whether the ordering is close enough to
  be safe, or whether a closure edited later would be skipped.
- **Page the window.** `IssueProvider::closed_since` could take a cursor or an
  offset, and `reconcile` could take pages until one comes back short. That
  costs more than one call in the cycle it happens, which is the bound the
  module claims — decide whether the bound should be "one call per cycle" or
  "one page per cycle, until caught up".

Either way, `crates/stella-protocol/src/issue.rs`'s `closed_since` doc comment
states the page-size contract and would move with the design.

## Verify

A fixture `IssueProvider` that reports more than `WINDOW` closures in one
window, including one belonging to a digest in the seen set, in the oldest
rows. That digest must age out — today it does not.

## Definition of done

- [ ] A closure past the first page of a saturated window still ages its
      digest out, with a test that fails before the change.
- [ ] The cost bound the module doc claims is restated to match whatever the
      new shape actually spends, and `stella.example.toml`'s decay note beside
      it says the same thing.
- [ ] An unreachable tracker still decays nothing, and the existing tests in
      `closures.rs` still pass as written.
- [ ] No new crate; the change stays in `stella-protocol` and
      `stella-cli/src/self_driving_cmd/`, and no touched file crosses the
      1500-line ceiling.

macanderson:
[...]

PR #6224, opened 2026-09-06
feat(stella-protocol,stella-cli): age a digest out when somebody else closes its issue

### LINKED ISSUE #6188: A digest never ages out when a person closes the issue the loop filed
## The problem

The self-driving loop's dedup set now ages a digest out when the issue it was filed as closes on a cited change (#6149, PR #6187). The signal it reads is the loop's **own** closure receipt: `crates/stella-cli/src/self_driving_cmd/supply.rs`'s `record_receipt` appends one line to `receipts.jsonl` as `close_issue` or `close_fixed_by_pr` runs, and `LoopState::live_seen` (`crates/stella-cli/src/self_driving_cmd/state.rs`) reads those receipts back.

Nothing else writes a receipt. So an issue the loop filed and a **person** closed — on the tracker, in a merged pull request that says `Closes #N`, or through a sweep that closed forty at once — leaves no receipt. Its digest keeps dropping repeats forever, which is the exact behaviour #6149 set out to end.

That is the common case in this repository: most issues the loop files are closed by a human or by a merge, not by `stella self-driving close`.

## What to build

The tracker knows the answer. `IssueProvider` (`crates/stella-protocol/src/issue.rs`) can read an issue's state, and `crates/stella-cli/src/self_driving_cmd/backlog.rs` already ranks open issues through it. One pass over the keys in `filings.jsonl` that are not already covered by a receipt would say which of them closed.

Constraints the design has to answer:

- **It must not cost a tracker call per filing per cycle.** `filings.jsonl` grows for the life of a loop. Cache the answer — a `closed` flag written back beside the filing record, or a receipt synthesized from the tracker's answer so `live_seen` keeps its single source.
- **"Closed" is not "fixed".** A `not planned` closure must decay nothing, which is what `regress::ClosureReceipt`'s `by: Option<Citation>` already encodes. The tracker's `IssueState` collapses to three buckets (`doc:agent-native-delivery` §4.1); the mapping from those to "cited a change" has to be stated, not assumed.
- **Offline must not decay anything.** A tracker that cannot be reached reads as "unknown", and an unknown age reads as live — the rule `crates/stella-autonomy/src/seen.rs` already states.

## Verify

A test with a fixture `IssueProvider` that reports one filed issue closed as completed, and shows the finding offered again on the next pass. Today's tests all seed `receipts.jsonl` by hand, because that is the only writer.

## Definition of done

- [x] A digest whose issue was closed outside the loop stops dropping repeats, with a test that fails before the change.
- [x] A digest whose issue was closed as declined or duplicate keeps dropping repeats.
- [x] An unreachable tracker decays nothing, with a test.
- [x] The tracker is not asked once per filing per cycle; whatever bounds it is written down where an operator reads it, next to the decay rule in `stella.example.toml`.
- [x] `finding_digest` and the format of `seen.txt` stay untouched, proven by `the_seen_set_survives_the_move_byte_for_byte` passing as written.
- [x] The fix lands in `stella-protocol` (the `IssueProvider`/`Issue` types the tracker-state pass reads) and `stella-cli/src/self_driving_cmd/` (state/supply/backlog), per AGENTS.md's crate-routing table — no new crate. None of the touched files (`state.rs`, `supply.rs`, `backlog.rs`) are on the grandfathered god-file list in `scripts/file-size-baseline.txt`; if any crosses 1500 lines, split rather than grow past it.
- [x] `make prose` and `make line-citations` stay green on the new code and docs — cite code by symbol (e.g. `LoopState::receipts`, `IssueProvider::get`), never by a pinned line number.
- [x] Not applicable: no UI/design-system surface — this is CLI + `stella-protocol`/`stella-autonomy` decision logic only.

---

### Verification note — autonomous PR sweep, 2026-09-06

All eight boxes ticked after reading PR #6224's diff against each item and
running the tests each one names. The runs below are on head `4959e5b`, the
merge commit this sweep pushed to resolve the branch's conflict with `main`,
so they describe the tree as it now stands rather than the pre-merge one.

`cargo test -p stella-cli --bin stella self_driving_cmd::closures` — 7 passed,
0 failed:

- `an_issue_a_person_closed_ages_its_digest_out` settles item 1. Its
  fail-before half is structural: `self_driving_cmd/closures.rs` does not exist
  on `main`, so the test cannot pass there.
- `a_closure_that_is_not_a_fix_decays_nothing` and
  `a_closure_with_no_resolution_decays_nothing` settle item 2, covering the
  declined case and the case where the tracker names no resolution at all.
- `an_unreachable_tracker_decays_nothing` settles item 3.
- `the_tracker_is_asked_once_a_cycle` settles the first half of item 4.

`cargo test -p stella-cli --bin stella the_seen_set_survives_the_move_byte_for_byte`
— 1 passed. That is item 5's named test, and it passes **as written**: the
diff's hunks in `state.rs` land at `filings_path`, `live_seen` and the new
closures accessors, none of which reach it. `finding_digest` appears nowhere in
the diff.

The rest, read off the diff:

- **Item 4's second half** — `stella.example.toml` gains eight lines directly
  under the existing decay-rule paragraph, saying the loop asks "once a cycle,
  one call for every issue closed since it last asked, kept in `closures.json`
  beside `seen.txt`", and that a cycle closing more issues than that one call
  returns leaves the rest for a later cycle. That is the bound, where an
  operator reads it.
- **Item 6** — the diff is seven files across `stella-protocol/src/issue.rs`
  and `stella-cli/src/self_driving_cmd/`, plus `issue_provider.rs`, which holds
  the `IssueProvider` implementation this item names. No new crate. No touched
  file carries an entry in `scripts/file-size-baseline.txt`, and `file size
  ratchet` is green. Note the diff adds `closures.rs` rather than touching
  `backlog.rs`; the item's requirement is the crate routing and the god-file
  rule, and both hold.
- **Item 7** — `make prose` and `make line-citations` are the `no content-free
  prose was added` and `prose cites code by name, never by line number` steps of
  the `gate steps no other workflow runs` job, which is green on this head.
  Neither runs in `docs guards` or in `fmt + clippy + test`, so that job is what
  settles this item.
- **Item 8** — no UI surface; the diff is CLI and protocol code plus one TOML
  example.

`fmt + clippy + test` is still running on this head. `cargo check -p stella-cli
--all-targets` exits 0 on the same tree, so the merge resolution compiles; the
full job still has to go green before the pull request can merge, which is a
separate matter from this checklist.

macanderson:
## What and why

The self-driving loop's dedup set ages a digest out once the issue it was filed
as closes on a fix. The only signal for that was the loop's **own** closure
receipt, written by `supply::record_receipt` as `stella self-driving close`
runs. Nothing else writes one.

In this repository most issues the loop files are closed by a person, or by a
merge that says `Closes #N`. Those closures leave no receipt, so the digest
kept dropping repeats for ever — the behaviour #6149/#6187 set out to end,
still in force for the common case.

Two halves:

- **The port.** `IssueProvider::closed_since(since, limit)` returns
  `Vec<IssueClosure>` — one windowed read of what a tracker has closed, each
  row carrying stella's **canonical** resolution (`completed` / `not_planned` /
  `duplicate`), the same vocabulary `IssueProvider::close` takes. The default
  is a typed refusal, following `IssueProvider::reopen`: an empty list is a
  real answer ("nothing closed in that window"), so a provider that cannot look
  must not return one. `GhIssueProvider` reads `number,stateReason,closedAt`
  and maps it back through `canonical_resolution`, the inverse of the existing
  `gh_close_reason`. An unrecognised or absent `stateReason` is `None`, never
  `completed`.
- **The loop.** `self_driving_cmd::closures::reconcile` asks once per cycle,
  keyed on the cycle counter stored in `closures.json`, and keeps only the keys
  this loop filed that the tracker calls completed. `LoopState::live_seen`
  unions those with the receipts it already read.

## Two departures from the design pointer, and why

The pointer said `live_seen` should reconcile against the tracker, caching per
cycle, failing open.

1. **The tracker read is not inside `live_seen`.** That getter is called from
   four filing paths (`supply::pass`, `supply::file`, `residue`,
   `self_driving_cmd::file_finding`) and is synchronous; a network call there
   would spend a tracker read per filing and put I/O in a getter. The read runs
   in `closures::reconcile`, called from `cycle_begin` and from `supply::pass`,
   guarded by the cached cycle counter — the second call in a cycle asks
   nothing. `live_seen` still reads files only.
2. **The answer is its own document, not a synthesized receipt.** A receipt
   with `by: None` ages nothing out, and inventing a citation to make one age
   out would hand `regress::sweep` a change to re-check that nobody named — a
   phantom regression, reported as a defect. The issue body offers both shapes;
   this is the "`closed` flag written back beside the filing record" one.

Everything else follows the pointer: closed-state reconciliation against the
tracker, one windowed call per cycle, and unknown = keep.

`IssueState` alone cannot tell a fix from a declined issue, which is why the
resolution rides on `IssueClosure` rather than becoming a fourth `IssueState`
— modelling the workflow is `doc:agent-native-delivery` §4.1's stated non-goal.

## Witness

`closures.rs`'s `an_issue_a_person_closed_ages_its_digest_out`. It seeds a
state directory with `seen.txt` and `filings.jsonl` and **no**
`receipts.jsonl`, asserts the digest still suppresses, runs `reconcile` against
a fixture tracker reporting that issue closed as `completed`, and asserts
`live_seen()` is then empty.

It fails on the old code by construction: `closures::reconcile` and
`IssueProvider::closed_since` do not exist there, and `live_seen`'s only source
of "fixed" is `receipts.jsonl`, which this fixture never writes. Every existing
test seeds receipts by hand, because that was the only writer.

Beside it: an unreachable tracker ages nothing out and leaves the watermark
where it was; a `not_planned` closure ages nothing out; a closure with no
resolution ages nothing out; the tracker is asked once per cycle and again on
the next; somebody else's closed issue is not kept.
`the_seen_set_survives_the_move_byte_for_byte` and `finding_digest` are
untouched — no line of `seen.txt` changes shape.

## Exemplar

`stella-protocol`'s own `IssueProvider::reopen` — a read/write the trait
refuses loudly by default rather than no-opping — is the shape `closed_since`
copies, and `GhIssueProvider::search`'s `--search` handling is the shape its
adapter copies.

## Cost bound

One tracker call per cycle, whatever the loop filed. The bound is written where
an operator reads it, next to the decay rule in `stella.example.toml`.

## Files

- `crates/stella-protocol/src/issue.rs` — `IssueClosure`, `closed_since`,
  round-trip test.
- `crates/stella-cli/src/issue_provider.rs` — `GhClosure`,
  `canonical_resolution`, the `gh issue list --state closed` read.
- `crates/stella-cli/src/self_driving_cmd/closures.rs` — new; the reconcile
  and its ledger.
- `crates/stella-cli/src/self_driving_cmd/state.rs` — `tracker_closures`,
  `write_tracker_closures`, and the union in `live_seen`.
- `crates/stella-cli/src/self_driving_cmd.rs`, `.../supply.rs` — the two call
  sites.
- `stella.example.toml` — the decay note.

No new crate, no new dependency. No file crosses the 1500-line ceiling
(`self_driving_cmd.rs` 1483, `state.rs` 1287); `drive.rs` was left alone
because it sits at 1498.

## Inherited red

`main` is known broken on `prose` — `crates/stella-cli/src/fleet_cmd.rs`
`issue-reference` is 54 against a baseline of 53, tracked as #6196 and repaired
by #6203, which was still open when this branch was cut. That file is not in
this diff. Everything else in `make guards-fast` is green here, and this
branch's own new prose measures reading grade 3.50 against the 6.00 ceiling a
file with no baseline entry is held to.

Tests deleted: None

Residue filed: #6219 — a cycle in which more issues close than one window
returns leaves the older rows unread, which costs one digest that keeps
suppressing.

Closes #6188

## Summary by Sourcery

Reconcile externally closed issues with self-driving state so completed fixes age their filing digests out safely.

New Features:
- Add windowed issue-closure querying to the issue-provider protocol and GitHub adapter, including canonical resolution mapping.
- Reconcile tracker-reported completed closures with self-driving filings so digests can age out when issues are closed externally.

Bug Fixes:
- Allow externally closed, fixed issues to stop suppressing future findings without requiring a self-driving closure receipt.
- Ensure unknown, declined, duplicate, or unreachable closure results do not incorrectly age digests out.

Enhancements:
- Persist tracker closure state and watermarks, cache reconciliation per cycle, and retain a bounded one-call-per-cycle tracker read.
- Document the closure reconciliation behavior and its cost in the example configuration.

Tests:
- Add coverage for external completed closures, unknown and non-fix resolutions, unreachable trackers, unfiled issues, cycle-level caching, and closure-window behavior.

### DIFF
--- a/crates/stella-cli/src/issue_provider.rs
+++ b/crates/stella-cli/src/issue_provider.rs
@@ -26,7 +26,8 @@ use std::process::Command;
 
 use async_trait::async_trait;
 use stella_protocol::issue::{
-    Issue, IssueClass, IssueDraft, IssueError, IssueKey, IssueLabel, IssueProvider, IssueState,
+    Issue, IssueClass, IssueClosure, IssueDraft, IssueError, IssueKey, IssueLabel, IssueProvider,
+    IssueState, RESOLUTION_COMPLETED, RESOLUTION_DUPLICATE, RESOLUTION_NOT_PLANNED,
 };
 
 /// The provider id this adapter answers to, and the one an error names.
@@ -69,6 +70,49 @@ struct GhIssue {
 #[derive(serde::Deserialize)]
 struct GhLabel {
     name: String,
+}
+
+/// What `gh issue list --state closed --json number,stateReason,closedAt`
+/// writes. Private to this module, like [`GhIssue`] beside it: `stateReason`
+/// is GitHub's word and it stops here.
+#[derive(serde::Deserialize)]
+struct GhClosure {
+    number: u64,
+    #[serde(rename = "stateReason", default)]
+    state_reason: Option<String>,
+    #[serde(rename = "closedAt", default)]
+    closed_at: String,
+}
+
+impl GhClosure {
+    fn into_closure(self) -> IssueClosure {
+        IssueClosure {
+            key: IssueKey(self.number.to_string()),
+            resolution: canonical_resolution(self.state_reason.as_deref()),
+            closed_at: self.closed_at,
+        }
+    }
+}
+
+/// Map GitHub's `stateReason` back onto stella's canonical resolution.
+///
+/// The inverse of [`gh_close_reason`], and it cannot be its exact inverse:
+/// that function maps both `not_planned` and `duplicate` onto GitHub's one
+/// `not planned`, so reading a closure back can only say the work was not
+/// done, never which of the two reasons it was. Both decay nothing, so the
+/// distinction the read cannot recover is one no caller here branches on.
+///
+/// Anything else — an absent reason, a `reopened`, a word this build has not
+/// seen — is `None`, meaning the tracker would not say. Guessing `completed`
+/// here would report declined work as fixed, which is the failure
+/// [`gh_close_reason`]'s own doc comment records happening once already.
+fn canonical_resolution(state_reason: Option<&str>) -> Option<String> {
+    match state_reason?.trim().to_ascii_lowercase().as_str() {
+        "completed" => Some(RESOLUTION_COMPLETED.to_owned()),
+        "not_planned" | "not planned" => Some(RESOLUTION_NOT_PLANNED.to_owned()),
+        "duplicate" => Some(RESOLUTION_DUPLICATE.to_owned()),
+        _ => None,
+    }
 }
 
 impl GhIssue {
@@ -343,6 +387,44 @@ impl IssueProvider for GhIssueProvider {
             .collect())
     }
 
+    /// `gh issue list --state closed --search "closed:>=<date>"` — one call for
+    /// the whole window.
+    ///
+    /// The search qualifier takes a date, so `since` is truncated to its day.
+    /// That re-reads the closures of one day the caller has already seen,
+    /// which costs nothing: the caller folds the answer into a set. A finer
+    /// window would risk the opposite, and a closure missed by a gap is a
+    /// digest that suppresses a real defect for ever.
+    async fn closed_since(
+        &self,
+        since: &str,
+        limit: usize,
+    ) -> Result<Vec<IssueClosure>, IssueError> {
+        let mut args: Vec<String> = vec![
+            "issue".into(),
+            "list".into(),
+            "--state".into(),
+            "closed".into(),
+            "--limit".into(),
+            limit.max(1).to_string(),
+            "--json".into(),
+            "number,stateReason,closedAt".into(),
+        ];
+        let day: String = since.trim().chars().take(10).collect();
+        if !day.is_empty() {
+            args.push("--search".into());
+            args.push(format!("closed:>={day}"));
+        }
+        let borrowed: Vec<&str> = args.iter().map(String::as_str).collect();
+        let raw = gh_json(&borrowed)?;
+        let rows: Vec<GhClosure> =
+            serde_json::from_str(&raw).map_err(|error| IssueError::Malformed {
+                provider: GITHUB.into(),
+                reason: error.to_string(),
+            })?;
+        Ok(rows.into_iter().map(GhClosure::into_closure).collect())
+    }
+
     async fn labels(&self, query: &str, limit: usize) -> Result<Vec<IssueLabel>, IssueError> {
         // `gh label list -S` searches names *and* descriptions and sorts by
         // best match, which is what a picker wants; an empty query lists the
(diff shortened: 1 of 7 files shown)


macanderson:
**Answering the one ❌ in Sourcery's assessment table** — "Reconcile closures with bounded tracker cost by caching results and documenting the once-per-cycle behavior".

The row's objective is met: one tracker call per cycle, cached on the cycle counter in `closures.json`, tested by `the_tracker_is_asked_once_a_cycle`, and written down for an operator beside the decay rule in `stella.example.toml`.

The reason given for the ❌ is the saturated window, and that is correct as a fact and deliberately out of scope here. It is filed as #6219 with the two candidate designs, the trade-off that rules out the obvious fix, and a test that would witness it. Two reasons it is a separate change rather than this one:

- **Neither cheap answer converges.** Keeping the watermark where it is re-reads the same newest page every cycle, because the tracker's order does not change; advancing it skips the rows past the page. Fixing it means either an ordering the search can be trusted to give (`sort:updated-asc`, whose relation to `closed` has to be measured) or a cursor on `IssueProvider::closed_since`, which changes the port's shape and the cost bound this PR documents. Both are their own decision.
- **It is strictly better than what ships today.** Before this PR a closure the loop did not make aged nothing out, ever. After it, one does — and a row past the first page costs exactly the old behaviour for that one digest, said out loud in the audit trail ("the window was full at N, so older closures in it were not read") rather than passing silently.

No fix pushed for this row; #6219 is the handoff.

Issue #6109, opened 2026-09-05
Decide which lane the four headless CLI doors run in


macanderson:
## Problem

`BuiltinLane` names seven turn-assembly sites. Four production engine
constructions in `stella-cli` are none of them, so after #6056 they are the
only turns left whose `agent.turn.started` carries `lane: null`:

| Site | What it drives |
|---|---|
| `crates/stella-cli/src/agent/turn.rs` (both arms) | `stella run` — the one-shot prompt door, and its process-free-authority twin |
| `crates/stella-cli/src/agent/goal.rs` | `stella goal` — the judged multi-round arc |
| `crates/stella-cli/src/agent/goal/goal_wrapped.rs` | the same arc under a wrapper plugin |

These run the session's top-level agent turn, exactly as the deck's lead turn
does. What differs is the renderer, not the assembly. But `BuiltinLane::Lead`
documents itself as "the interactive-mode turn (`stella-cli`'s
`command_deck`)", and `doc:turn-lane-assembly` §2's enumeration of the seven
lanes names only `command_deck.rs` for that row — so stamping these four with
`Lead` today would make the lane say something its own definition denies.

#6056 therefore left them unattributed with a written reason rather than
guessing, and this issue is where the guess gets settled.

## The decision to make

One of:

1. **Widen `Lead`** to mean "the session's own top-level turn, deck or
   headless", and stamp all four. Costs a doc rewrite in
   `crates/stella-protocol/src/lane.rs` and in `doc:turn-lane-assembly` §2, and
   makes a lane-grouped report merge deck turns with `stella run` turns.
2. **Add an eighth `BuiltinLane` case** (`Door`, say) for the headless doors.
   `BuiltinLane` is closed with no `serde(other)`, so this is a wire change:
   `BuiltinLane::ALL` grows, `as_str` gains an arm, and every exhaustive match
   over the enum has to answer for it.
3. **Keep them unattributed**, and say so in `lane.rs` the way
   `PipelineStage` says why it has no producer — a written decision rather
   than four separate comments.

Whichever wins, the four sites move onto `Engine::assemble` so the answer is a
written `lane:` field rather than the `None` `Engine::with_sleeper` writes by
construction.

## Constraints already found

- `crates/stella-cli/src/turn_files.rs`'s
  `every_engine_driver_declares_what_it_owes_the_ledger` sweeps both
  constructors and already lists all four files, so migrating them changes
  nothing there.
[...]

PR #6232, opened 2026-09-06
Delete the engine builder shims so a turn cannot be built with no seams answered

### LINKED ISSUE #6164: Delete the engine builder shims so a turn cannot be built with no seams answered
## Problem

`Engine::assemble` exists so that every optional seam gets an answer somebody typed. The old way in is still there beside it and still `pub`: `Engine::with_sleeper` plus `with_hooks`, `with_calibration`, `with_gate`, `with_steering`, `with_requery`, `with_bus`, `with_provider_outcomes`, `with_fallback_resolver` and `with_hook_approval_route`.

While those stand, an engine can be built that answers no seam. A capability added to the loop reaches such a site as a silent `None` instead of as a build error, which is the whole property `Engine::assemble` was added to give. #3274 slice 3 asks for the deletion by name.

Every lane in the tree now assembles. What is left is a sweep of the callers that are not lanes.

## What is left, measured on `main` after #6109 lands

Shipping code, two sites, both host engines that never drive a turn of their own and both say so in a comment:

- `crates/stella-cli/src/subagent.rs` — the host engine `SessionSubAgents` forks children from.
- `crates/stella-serve/src/subagents.rs` — the same shape on the served side.

Test code, roughly forty call sites:

- `crates/stella-core/src/subagent/tests/` — `seams.rs`, `failure_and_events.rs`, `fork_scope.rs`, `tools_and_budget.rs`, `witness.rs`, `spend_and_cancellation.rs`.
- `crates/stella-core/src/driver/tests.rs`.
- `crates/stella-cli/src/agent/resume.rs` test module.
- `crates/stella-serve/src/session.rs` test module.

Find them with `rg -n 'Engine::with_' crates/`.

## The decision this needs first

The two shipping sites are not lanes and should not pretend to be. Pick one and write the reason where the reader of the call finds it:

1. Give each an `Engine::assemble` literal with `lane: None` and a comment saying a host engine drives no turn, so it names no lane. This is what lets the builders go.
2. Keep the builder path for them and drop the deletion from slice 3, which means the totality property stays unenforceable for anyone who calls the builders.

Option 1 is what slice 3 asks for. `TurnCapabilities::none()` already exists for exactly the "no seam" case and is written as an exhaustive literal, so a site that binds nothing is one argument.

## Constraints already found

- `crates/stella-cli/src/turn_files.rs`'s `every_engine_driver_declares_what_it_owes_the_ledger` sweeps both constructors by name. Deleting `with_sleeper` means that fence's needle list drops one entry, and `subagent.rs`'s row in `ENGINE_DRIVERS` has to keep matching.
- `crates/stella-parity/src/lib.rs`'s `COMPOSITION_SEAMS` lists `with_sleeper`, and several `Capability` rows name a builder in `engine_entries` (`with_steering`, `with_requery`, `with_gate`, `with_hooks`, `with_bus`, `with_hook_approval_route`, `with_calibration`, `with_provider_outcomes`, `with_fallback_resolver`). An engine-side completeness sweep requires every public `Engine` method to be claimed by a row or listed there, so each deleted method costs an edit in that file and the rows must then name whatever replaces it.
- `crates/stella-core/src/driver.rs` is a god file closed to growth. Deleting from it is fine; adding is not.
- `crates/stella-cli/src/lane_capabilities.rs` is the pattern for a call site that answers every seam as one literal.
- The prose gate is a down-only ratchet per (file, pattern). Comments in these files carry issue numbers today; new prose must not add more `issue-reference` hits.

## Verify

`rg -n 'Engine::with_' crates/` returns nothing but the definition of whatever survives. A witness in the shape of `lane_capabilities.rs`'s `each_door_that_is_not_the_deck_assembles_through_its_lane` reads the two shipping sources back and asserts neither names the builder constructor, so a later site cannot quietly return to it.

## Definition of done

- [x] `crates/stella-cli/src/subagent.rs` and `crates/stella-serve/src/subagents.rs` build through `Engine::assemble`, each with a written answer for every seam and a comment saying why it names no lane.
- [x] Every test call site is migrated.
- [x] The `with_*` builders this issue names are deleted from `crates/stella-core/src/driver.rs` and `crates/stella-core/src/driver/user_hooks.rs`.
- [x] `crates/stella-parity/src/lib.rs` no longer names a deleted method, and its engine-side completeness test passes.
- [x] `crates/stella-cli/src/turn_files.rs`'s driver fence still finds every declared driver.
- [x] A source-text witness holds the two shipping sites to the blessed constructor.
- [x] No ceiling in `scripts/file-size-baseline.txt` was raised.

Refs #3274, #6109

macanderson:
## What and why

`Engine::assemble` exists so that every optional seam on a turn gets an answer
somebody typed. A second constructor stood beside it and answered none of
them: `Engine::with_sleeper`, plus one public setter per seam — `with_hooks`,
`with_calibration`, `with_gate`, `with_steering`, `with_requery`, `with_bus`,
`with_provider_outcomes`, `with_fallback_resolver` and
`with_hook_approval_route`.

While those stood, the totality property was worth only what a caller chose to
spend on it. A capability added to the loop reached such a site as a silent
`None` instead of as a build error, which is the whole property
`TurnCapabilities` was added to give. #3274 slice 3 asks for the deletion by
name; #6109 did the lane migration, and this is its deletion half.

All ten are deleted from `crates/stella-core/src/driver.rs` and
`crates/stella-core/src/driver/user_hooks.rs` (`driver.rs` is a god file — this
change only removes lines from it). `with_call_role`, `with_turn_instance`,
`with_turn_halt` and `with_turn_controls` are not named by the issue and stay.

## What the census actually found

The design pointer named four callers. Measured on `main` with a
balanced-paren scan rather than `rg -c`, there were **219 `Engine::with_sleeper`
sites in 43 files**, 43 of them carrying a chain of a deleted setter. The
pointer's four are the *lane doors*, which already assembled. The rest are
tests plus two non-lane hosts. Every one of the 219 is migrated here.

**The two shipping host engines take option 1 from the issue.**
`crates/stella-cli/src/subagent.rs` and `crates/stella-serve/src/subagents.rs`
now write `TurnCapabilities::none()` with a comment saying why: the engine
never drives a turn of its own, so it has nothing to pause, steer or
attribute, and `run_sub_agent` assembles the child that stamps
`BuiltinLane::SubagentFork`.

**Test sites write the seams they bind over `..TurnCapabilities::none()`.** A
lane writes every field out — that is what `lane_capabilities.rs` is for. A
fixture writes only what it binds, because `none()` is itself an exhaustive
literal: a new slot still stops the build there until somebody answers it.
`capabilities.rs`'s module doc now states that split, beside the reason there
is no `Default`.

**One behaviour change, in `crates/stella-cli/src/fleet_cmd.rs`.** It attached
its re-query port with `engine.with_requery(...)` after `assemble`. That seam
moves into `lane_capabilities::fleet_attempt`'s argument list, so the fleet
lane binds what it binds in one written literal, the way every other lane
does. `each_lane_binds_what_its_call_site_bound_before` gains the both-ways
assertion for it.

## Witness mechanism

Two, both source-text, both failing on `main` by construction.

- `crates/stella-core/src/driver/capabilities.rs` —
  `no_per_seam_builder_survives_beside_assemble` reads `driver.rs` and
  `driver/user_hooks.rs` back and asserts neither defines `pub fn with_<seam>(`
  for any of the ten. On `main` `driver.rs` contains `    pub fn with_sleeper(`,
  so it fails. It asserts both files still open an `impl<'a> Engine<'a>` block,
  so an empty search cannot pass for a clean one. Source text rather than a
  `compile_fail` doctest on purpose: a doctest passes on any compile error at
  all, including a typo in its own setup, so it cannot tell "the builder is
  gone" from "this snippet is broken".
- The two shipping hosts are each held by a test beside their own source:
  `the_host_engine_this_crate_builds_assembles_with_a_written_seam_set`
  (`lane_capabilities.rs`, reading `subagent.rs`) and
  `this_host_engine_assembles_with_a_written_seam_set` (`subagents.rs`, reading
  itself). Each asserts the file names `Engine::assemble(` and
  `TurnCapabilities::none()` and does not name the sleeper constructor. One
  test per crate rather than one test reading across the crate boundary,
  following the `accept.rs` twin already in the tree, so each crate's gate
  stays self-contained.

Needles are built with `format!` in every case, so no witness is its own match
and `turn_files`' driver fence reads these files as prose rather than as doors.

## The fences that had to move with it

- `crates/stella-cli/src/turn_files.rs`'s driver fence drops its
  `Engine::with_sleeper` needle: there is one constructor now, so a door has no
  second spelling to hide behind. `subagent.rs`'s `ENGINE_DRIVERS` row still
  matches.
- `crates/stella-parity/src/lib.rs` drops `with_sleeper` from
  `COMPOSITION_SEAMS`, and the seven `Capability` rows that claimed a deleted
  method now claim no engine entry point and name the `TurnCapabilities` field
  in `engine_home` instead. `every_public_engine_entry_point_is_claimed` and
  `every_claimed_engine_entry_exists` both still hold.

## Exemplar

`crates/stella-cli/src/lane_capabilities.rs` — the shape this repository
already chose for "one written literal per assembly site", including the
`let seams = …; Engine::assemble(…, seams)` spelling every call site here now
uses.

## A god file had to be split

`crates/stella-core/src/driver/tests.rs` sits at its 2907-line ceiling and 36
of these sites are in it, so the migration would have pushed it over. No
ceiling was raised. Its two streaming-deadline tests move to a new sibling,
`crates/stella-core/src/driver/tests/streaming_deadline.rs`, which lands at
2773 lines. The moved prose is rewritten plainly, because a new file is held to
a 6.00 reading grade; it measures 2.58.

## Tests deleted

None. `a_streaming_generation_outlives_the_deadline_because_it_is_not_stalled`
and `a_call_only_stream_outlives_the_deadline_because_it_is_not_stalled` moved
from `driver/tests.rs` to `driver/tests/streaming_deadline.rs` under the same
names.

## Also repaired here

`main` was red on the prose guard before this branch: #6168 added an
issue-number comment to `crates/stella-cli/src/fleet_cmd.rs` without the
baseline moving with it. #6203 fixed the same line while this branch was in
flight and the rebase kept its wording, so nothing of that repair is left in
this diff.

`crates/stella-runtime/README.md` carried a live claim that
`with_gate`/`with_steering`/`with_hooks` "are `pub` and still directly
callable". They are not, so it says so.

## Evidence

`make guards-fast` green locally — prose (count, header density and reading
grade), file-size, line-citations, dead-code-allows, lockfile-sync, fmt.
Everything that compiles is CI's.

## Remaining work filed

See the comment below for the issues this PR files.

Closes #6164
Refs #3274, Refs #6109

### DIFF
--- a/crates/stella-cli/src/agent/engine.rs
+++ b/crates/stella-cli/src/agent/engine.rs
@@ -842,8 +842,8 @@ pub(crate) fn session_router(cfg: &Config, worker_ref: &ModelRef) -> Router {
 /// Mid-turn fallback resolution for the bare (non-pipeline) loops (#2679):
 /// when the worker's retry ladder exhausts, the engine asks this port for a
 /// replacement, and the answer is a fresh [`Router::resolve`] of the worker
-/// role — whose breaker the failing calls already fed via
-/// `with_provider_outcomes`, so resolution routes around the sick provider.
+/// role — whose breaker the failing calls already fed through the seam set's
+/// `outcomes` slot, so resolution routes around the sick provider.
 /// The replacement adapter is built from the discovered credentials on
 /// first use and owned here (set-once), so the engine can borrow it for the
 /// rest of the turn. Every miss is soft, matching
--- a/crates/stella-cli/src/agent/resume.rs
+++ b/crates/stella-cli/src/agent/resume.rs
@@ -198,13 +198,9 @@ pub(crate) async fn run_resume(cfg: &Config, id: Option<&str>) -> Result<(), Cli
         // Assembled, not built up by optional builders. This turn is the
         // `Resume` lane and says so. `lane_capabilities::resume` answers
         // every seam, so nothing is left to a chain.
-        let engine = Engine::assemble(
-            &*provider,
-            &tools,
-            engine_config,
-            &TokioSleeper,
-            crate::lane_capabilities::resume(cfg.hooks.as_ref(), &hook_runner, &calibration),
-        );
+        let seams =
+            crate::lane_capabilities::resume(cfg.hooks.as_ref(), &hook_runner, &calibration);
+        let engine = Engine::assemble(&*provider, &tools, engine_config, &TokioSleeper, seams);
         drive_resumed_turn(&engine, state, &events).await
     };
 
@@ -356,6 +352,7 @@ mod tests {
     use async_trait::async_trait;
 
     use super::*;
+    use stella_core::TurnCapabilities;
     use stella_core::step::{
         BudgetSnapshot, CHECKPOINT_VERSION, Checkpoint, CheckpointSink, TurnState,
     };
@@ -495,7 +492,14 @@ mod tests {
             ..EngineConfig::default()
         };
         let state = TurnState::from_checkpoint(killed_mid_turn(), &config);
-        let engine = Engine::with_sleeper(&provider, &tools, config, &crate::runtime::TokioSleeper);
+        let seams = TurnCapabilities::none();
+        let engine = Engine::assemble(
+            &provider,
+            &tools,
+            config,
+            &crate::runtime::TokioSleeper,
+            seams,
+        );
         let (tx, mut rx) = mpsc::unbounded_channel::<AgentEvent>();
         let events = stella_core::EventSender::new(tx);
 
@@ -566,7 +570,14 @@ mod tests {
         let mut at_cap = killed_mid_turn();
         at_cap.step = config.max_steps;
         let state = TurnState::from_checkpoint(at_cap, &config);
-        let engine = Engine::with_sleeper(&provider, &tools, config, &crate::runtime::TokioSleeper);
+        let seams = TurnCapabilities::none();
+        let engine = Engine::assemble(
+            &provider,
+            &tools,
+            config,
+            &crate::runtime::TokioSleeper,
+            seams,
+        );
         let (tx, _rx) = mpsc::unbounded_channel::<AgentEvent>();
         let events = stella_core::EventSender::new(tx);
 
@@ -611,7 +622,14 @@ mod tests {
         let mut at_cap = killed_mid_turn();
         at_cap.step = config.max_steps;
         let state = TurnState::from_checkpoint(at_cap, &config);
-        let engine = Engine::with_sleeper(&provider, &tools, config, &crate::runtime::TokioSleeper);
+        let seams = TurnCapabilities::none();
+        let engine = Engine::assemble(
+            &provider,
+            &tools,
+            config,
+            &crate::runtime::TokioSleeper,
+            seams,
+        );
         let (tx, _rx) = mpsc::unbounded_channel::<AgentEvent>();
         let events = stella_core::EventSender::new(tx);
 
--- a/crates/stella-cli/src/command_deck/lead_control.rs
+++ b/crates/stella-cli/src/command_deck/lead_control.rs
@@ -70,7 +70,7 @@ impl LeadPause {
         }
     }
 
-    /// The gate to hand to `Pipeline::with_turn_gate` and `Engine::with_gate`.
+    /// The gate to hand a seam set's `gate` slot.
     /// (`attach_turn_controls` wants the owned form — see [`turn_controls`].)
     pub(super) fn turn_gate(&self) -> &dyn stella_core::ports::TurnGate {
         self.gate.as_ref()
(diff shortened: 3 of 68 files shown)


macanderson:
## Remaining work filed

- #6233 — three design specs under `docs/spec/` still describe `Engine::with_sleeper` and the per-seam builders as live code. Left out of this PR because correcting them is a judgement call section by section (history to keep vs. claim to fix), and both prose ratchets bite on every sentence.
- #6234 — `crates/stella-core/src/driver/tests.rs` dropped to 2773 lines and its ceiling still reads 2907. Reclaiming that slack is `make file-size-retighten`, which AGENTS.md says is a PR of its own.

macanderson:
## Sourcery

Sourcery declined to review this PR: `Sorry @macanderson, your pull request is larger than the review limit of 150,000 diff characters`. So there is no assessment table and no `❌` row to settle. Recording it here rather than leaving the absence unexplained.

The size is the migration itself — 219 call sites in 43 files, one mechanical shape per site — not new logic. What a reviewer wants to read is small: the deletion in `crates/stella-core/src/driver.rs` and `crates/stella-core/src/driver/user_hooks.rs`, the three witnesses, and the two fences (`turn_files.rs`, `stella-parity`).

PR #6254, opened 2026-09-06
fix(gate): the prose ratchets judge the change, not the tree

### LINKED ISSUE #6244: The prose ratchets fail a branch for drift it inherited, the way the file-size ratchet used to
## What happened

`scripts/check-prose.py` holds three ratchets. One of them already judges the
change rather than the tree, and two do not:

| Ratchet | Judged against |
| --- | --- |
| module-header density (`over`) | `max(ceiling, the same unit in the base tree)` — `resolve_base_commit` |
| per-(file, pattern) counts (`failures`) | the baseline alone |
| per-file reading grade (`grade_failures`) | the baseline alone |

So a pull request that touches none of the drifted files still fails `prose`,
and the author has to either repair somebody else's drift inside their own
diff — which AGENTS.md forbids for a shared cell — or land a separate repair
first and wait for it.

That happened on 2026-09-06. `main` was red on `prose` (#6196): one
`issue-reference` count over in `crates/stella-cli/src/fleet_cmd.rs`, four
reading grades over in files from the same change, and two older grade drifts
underneath. A pull request holding no Rust at all (#6204, an ADR plus two shell
scripts) could not go green until #6203 landed the repair. The whole cost was
one branch waiting on another.

The density half already carries the argument, in its own comment:

> a unit already over its ceiling before this change landed is inherited drift,
> and failing the branch that merely did not fix it is what turned this ratchet
> into a main-red generator.

The same sentence is true of the other two, and `scripts/check-file-size.sh` is
where the pattern comes from.

## Where

- `scripts/check-prose.py` — `resolve_base_commit`, `base_tracked_paths`, and
  the `failures` / `grade_failures` loops.
- `scripts/check-file-size.sh` — the exemplar, including how it falls back to
  the strict whole-tree check when no base resolves.
- `scripts/test-prose.sh` — where the directions are pinned.

## The fix

Give counts and grades the same base-tree allowance density has:
`allowed = max(baseline entry, the same file's value in the base tree)`. Keep
`--absolute` skipping the base, because that is what the post-merge canary
runs and inherited drift is exactly what a canary must still report. An
unresolvable base must make the check stricter, never weaker — the file-size
guard's rule.

Two things this must not weaken: a first-time offender still fails, since a
file absent from the baseline is held to zero hits and grade 6.00 in both
trees; and `--update` still refuses to raise a number.

## Definition of done

- [x] `failures` and `grade_failures` are judged against
      `max(ceiling, base-tree value)`, with `--absolute` unchanged.
- [x] A test in `scripts/test-prose-guard.sh` (the harness that exists; there
      is no `scripts/test-prose.sh`) pins it in both directions: a branch
      that inherits an over-ceiling file passes, and the same branch fails
      once it adds to that file. Cases `C1`/`C2` for counts and `C5`/`C6` for
      grades; `C1` and `C5` fail on today's `main` script and pass with the
      fix.
- [x] `--absolute` still fails on inherited drift, so the canary keeps
      reporting a red `main`. Case `C3`.
- [x] An unresolvable base (shallow clone, fresh repository) falls back to the
      strict whole-tree check, with a test. Case `C4`.
- [x] `make prose` stays green on `main`.
- [x] The "Enforced by `make prose`" paragraph is updated to describe the
      base-tree allowance for `failures` and `grade_failures`, matching how it
      already documents that allowance for the density ratchet and for
      `check-file-size.sh`. That paragraph lives in `CLAUDE.md`, which imports
      `AGENTS.md`; `AGENTS.md` carries no copy of it.
- [x] The change stays inside `scripts/check-prose.py` and
      `scripts/test-prose-guard.sh`, beside their existing sibling
      `scripts/check-file-size.sh` — no new script, no new crate, and neither
      file is on the god-file list (`AGENTS.md`'s "God files" table). The doc
      paragraph in `CLAUDE.md` is the only other file touched.
- [x] The diff's own new comments/prose pass `make prose` — this fix must not
      introduce the fluff or announcing constructions `docs/prose-guidelines.md`
      bans.

macanderson:
## What and why

`scripts/check-prose.py` holds three ratchets, and only one of them judged the
change rather than the tree. Module-header density already compared a crate's
mean against the base tree; the per-(file, pattern) counts and the per-file
reading grade compared against their baseline alone. So a branch that touched
none of the drifted files still failed `make prose`, and the author had to
repair somebody else's prose inside their own diff — which AGENTS.md forbids for
a shared cell — or land a separate repair first and wait for it. That happened
on 2026-09-06: a pull request holding no Rust at all could not go green until an
unrelated repair merged.

Counts and grades now take the allowance density already had:

    allowed = max(the baseline entry, the same value in the base tree)

The base is the commit `resolve_base_commit` already picks (the merge commit's
first parent in CI, the merge base with `origin/main` locally, `PROSE_BASE_REF`
for a hermetic test). `--absolute` still skips the base entirely, so the
post-merge canary keeps reporting drift that reached `main`. An unresolvable
base — a shallow clone, a fresh repository — resolves to `""` and holds the whole
tree to the baseline, because an unknown must make the check stricter and never
weaker.

Two directions this does not weaken, both pinned by tests:

- A first-time offender still fails. A file the base tree does not hold reads
  back as zero hits and no grade, so it inherits nothing.
- `--update` is untouched: it still refuses to raise a count, a grade or a mean.

Inherited drift is named on the way past (`check-prose: drift inherited from the
base tree…`), so forgiving it is never the same as hiding it. The density
ratchet, which forgave silently before, now reports through the same line.

Exemplar: `scripts/check-file-size.sh` — the same rule, the same base
resolution, and the same fallback (AGENTS.md, "The ratchet judges your change,
not the tree").

## Witness mechanism

Six cases in `scripts/test-prose-guard.sh`, hermetic like its `B` and `S`
siblings — each builds a throwaway git repository, commits a drifted file,
points `refs/remotes/origin/main` at that commit, then commits the branch's own
change on top.

- **C1** (counts) and **C5** (grade) are the witnesses: the base tree is already
  over a ceiling and the branch adds an unrelated file. Both fail on `main`'s
  script — I verified it by running the suite with
  `git show origin/main:scripts/check-prose.py` in place:

      FAIL C1 an inherited count does not fail a branch that did not cause it -- the guard exited non-zero
      FAIL C1 the forgiven file is named -- the run said nothing about docs/drift.md
      FAIL C5 an inherited reading grade does not fail an untouched branch -- the guard exited non-zero

  All six pass with this change (69 passed, 0 failed).
- **C2** and **C6** are the boundary: the same fixtures fail once the diff adds a
  construction to the drifted file, or raises its grade.
- **C3** keeps `--absolute` strict on a tree C1 forgives.
- **C4** is the fallback: one commit, no `origin/main`, nothing resolves as a
  base, and the drifted file fails.

`scripts/test-prose-guard.sh` runs in `guard-self-tests.yml` and as
`make prose-test`.

## Evidence

- `./scripts/test-prose-guard.sh` — 69 passed, 0 failed.
- `python3 scripts/check-prose.py` and `--absolute` — both OK on this branch.
- `make guards-fast`, `shellcheck scripts/test-prose-guard.sh`,
  `make line-citations`, `./scripts/check-file-size.sh` — all green locally.
- CI is the rest of the evidence.

## Docs

CLAUDE.md's "Enforced by `make prose`" paragraph now says all three ratchets
judge the change against the base tree, what `--absolute` does, and that an
unresolvable base makes the check strict. That paragraph is the one the issue's
DoD names: it lives in CLAUDE.md, which imports AGENTS.md.

Tests deleted: None.

Residue issues filed: none — nothing was left behind by this change.

Closes #6244

## Summary by Sourcery

Judge prose ratchets against the branch change and base tree so inherited drift is reported without blocking unrelated work.

Bug Fixes:
- Make prose count and reading-grade ratchets compare changes with the resolved base tree, so inherited violations no longer fail unrelated branches while newly introduced drift still fails.
- Preserve strict whole-tree checking for --absolute and when the base cannot be resolved, including first-time offenders.

Enhancements:
- Report inherited prose drift explicitly, including density, count, and reading-grade violations, while retaining rename-aware base comparisons.

Documentation:
- Document the change-based behavior of all prose ratchets, --absolute semantics, and strict fallback behavior in CLAUDE.md.

Tests:
- Add hermetic guard tests covering inherited drift, newly introduced violations, absolute mode, and unresolved-base fallback.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -153,7 +153,14 @@
     `make prose-update` refuses to raise a count or add a pair, so a red gate
     is cleared by deleting the prose. The baseline records debt older than
     the guard, it is meant to reach empty, and adding a line to it is the
-    expedient this file forbids. Adding a *pattern* is the one exception, and
+    expedient this file forbids. Each of the three ratchets — the counts,
+    the reading grade, and the header density below — judges your change
+    and not the whole tree, the way `check-file-size.sh` does. A number
+    fails only when it is over its own ceiling *and* over the same value in
+    the base tree, so drift you inherited is named in the run and does not
+    fail you. `--absolute` skips the base tree, which is what the
+    after-merge check runs; a base that does not resolve makes the check
+    strict again. Adding a *pattern* is the one exception, and
     it has its own door: `make prose-adopt PATTERN=<name>` records that
     pattern's pre-existing hits, once, and refuses to touch any other
     pattern's numbers. A backticked span or a fenced block is exempt: naming
--- a/scripts/check-prose.py
+++ b/scripts/check-prose.py
@@ -63,13 +63,15 @@ The baseline is a shared cell in the sense AGENTS.md describes for
 and compose into a red `main`. Two doors close that, mirroring
 `check-file-size.sh`'s own split:
 
-- The plain check judges a unit against `max(its baseline, its mean in the
-  base tree)`. Inherited drift is reported as drift and does not fail a
-  branch that did not cause it, so landing one ordinary header cannot
-  redden `main` on its own. `--absolute` opts out of the base comparison for
-  the one caller that must not get this mercy: a post-merge canary is
+- The plain check judges a count, a grade and a unit's mean against
+  `max(its baseline, the same value in the base tree)`. Inherited drift is
+  named on the way past and does not fail a branch that did not cause it, so
+  landing one ordinary header, or leaving somebody else's hard sentence
+  alone, cannot redden `main`. `--absolute` opts out of the base comparison
+  for the one caller that must not get this mercy: a post-merge canary is
   exactly asking whether drift already reached `main`, and the base-relative
-  reading would forgive the thing it exists to catch.
+  reading would forgive the thing it exists to catch. A file the base tree
+  does not have inherits nothing, so a first-time offender still fails.
 - `--update` alone leaves every unit's ceiling where it stands, except for a
   unit a file move took a header out of or into: that entry is re-based
   against the same files' lengths in the base tree, because a move changes a
@@ -105,9 +107,10 @@ Usage:
                   when the grade ratchet arrived
     --report      print every offending line, grouped by file, then each
                   unit's mean header length; changes nothing
-    --absolute    judge every unit's density against its baseline alone,
-                  ignoring the base tree. For the post-merge canary, which
-                  must catch drift a base-relative check would forgive.
+    --absolute    judge every count, grade and unit density against its
+                  baseline alone, ignoring the base tree. For the post-merge
+                  canary, which must catch drift a base-relative check would
+                  forgive.
 """
 
 from __future__ import annotations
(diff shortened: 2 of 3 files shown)


macanderson:
Answering the one ❌ row in Sourcery's assessment table ("Update the issue-specified prose documentation and test location").

Both paths the issue names are wrong about this tree, so the diff points at the files that exist.

**The paragraph.** `rg -n "Enforced by .make prose." AGENTS.md CLAUDE.md` matches one line: `CLAUDE.md:149`. AGENTS.md has no such paragraph — it names `prose` only as a gate step and a workflow trigger. CLAUDE.md imports AGENTS.md (`@AGENTS.md` on its second line), so the two are one document at session load, and the paragraph the issue means is the one this PR edited. The maintainer's design comment on the issue says the same thing.

**The test file.** `scripts/test-prose.sh` does not exist and never has (`ls scripts/test-prose*` returns `scripts/test-prose-guard.sh` alone). That file is what `make prose-test` and `guard-self-tests.yml` run, and it is where the `B` and `S` base-tree cases already live, so the six new cases sit beside their siblings.

No follow-up issue is owed: nothing is deferred, only relocated to the real paths.

PR #6228, opened 2026-09-06
feat(ci): reject a PR that negates a closing keyword

macanderson:
## What & why

GitHub's closing-keyword parser reads a keyword (`close`, `fix`, `resolve`,
and their inflections) and the issue reference right after it. It does not
read the words in front of the keyword. PR #6180 carried `Refs #6155` in its
commit trailer -- the correct spelling for "advances but does not finish"
-- and its description said "This does not close `#6155`." GitHub closed
`#6155` anyway. A person had to reopen it by hand.

`scripts/check-closing-keywords.py` fails a pull request whose description
or commit messages carry a negation word (`not`, `never`, `without`,
`doesn't`) in front of a closing keyword and its issue reference, in the
same sentence. It runs in `ci.yml`'s guards job, `pull_request`-only, next
to `check-deleted-tests.sh` -- both read a PR's live description and commit
messages via `gh pr view`, which is not something a single local tree
carries, so neither is a `make gate` step. It also runs in
`guard-self-tests.yml`'s `gate-steps` job, whose trigger carries no `paths:`
filter and which is not gated on the diff reaching a crate -- see the note
below.

The remedy the guard prints matches the issue's own design: wrap the issue
number in backticks (`` `#N` ``), or write "advances #N" instead of a
closing keyword.

AGENTS.md's "One keyword per issue" paragraph gains the negation case,
citing #6180 as the real example.

Closes #6190

## The witness

- [x] This PR includes a witness test (fails on `main`, passes here), **or**
- [ ] No witness needed

`scripts/test-closing-keywords.py` runs the guard as a subprocess against
fixture PR bodies and commit messages -- it fails on `main` because
`scripts/check-closing-keywords.py` does not exist there, and passes with
this change (22 cases, including the three named in the issue's own "How to
verify" section: a negated body fails, a `Refs #N` body passes, a genuine
`Closes #N` body passes). `make closing-keywords-test` runs it locally; it
is also wired into `guard-self-tests.yml` as a hermetic self-test.

## The gate

- [x] fmt check -- ran `cargo fmt --all` locally, no diff (no Rust changed)
- [ ] clippy, full workspace, all targets -- no Rust files changed; CI's step
      is unaffected
- [ ] test, full workspace -- no Rust files changed
- [x] Docs updated: AGENTS.md's "One keyword per issue" paragraph
- [x] CLA signed
- [x] `Closes #N` appears both above and as a commit trailer

## Nothing left behind

[...]

PR #6255, opened 2026-09-06
fix(scripts): don't stand a session down for a closed-unmerged PR

macanderson:
## What

`scripts/issue-claim.sh check <n>` treated any PR that referenced the
issue as a live claim, so it stood a session down for a PR that was
closed without merging -- the one state where the work is neither done
nor in progress, and where the next session most needs to proceed, not
stop. It also matched only `Closes #N`, so a PR that deliberately used
`Refs #N` (because `dod-check` reads every linked issue's checklist,
and a `Closes` there would hold an unrelated fix to it) was invisible.

## Why

Working #6193, the check stood a session down for #6200 -- closed as
superseded minutes after opening -- while the live carrier, #6203, used
`Refs #6193` on purpose and never showed up at all. The check inverted
twice on one issue.

## The fix

`scripts/issue-claim.sh`'s `gh pr list` query now tags each hit with a
third field, `closes` or `refs`, alongside its number and state. The
verdict:

- a `closes` hit that is OPEN or MERGED still blocks (unchanged).
- a `closes` hit that is CLOSED does not block -- it is reported as a
  dead attempt, and the check proceeds.
- a `refs` hit never blocks -- it is reported as a weaker signal, since
  `Refs` is also how an unrelated PR cites context.
- a mix (e.g. one CLOSED hit and one OPEN hit) still blocks, and both
  are named with their own state.

`AGENTS.md`'s `scripts/issue-claim.sh` paragraph is updated to describe
the split verdict.

## Witness

`scripts/test-issue-claim.sh` adds three cases (plus the fixture
format everywhere else moves from `<number> <state>` to
`<number> <state> <kind>`):

- a closed-unmerged closing PR: exit 0, PR named.
- a closed-unmerged PR alongside an OPEN one: exit 1, both named, each
  with its own state.
- a refs-only PR: exit 0, PR named as a weaker signal.

Verified fail on the pre-fix script by stashing only `scripts/issue-claim.sh`
and re-running the suite against the new tests: 2 of 23 cases red (the
closed-unmerged case and the refs-only case), the rest green. Restoring
the fix turns the suite fully green (23/23).

## Local checks run

- `./scripts/test-issue-claim.sh` -- 23/23 passing.
- `shellcheck scripts/issue-claim.sh scripts/test-issue-claim.sh` -- clean.
- `make guards-fast` -- clean (includes `check-prose`, `check-line-citations`,
  `check-file-size`, `cargo fmt --check`).

Tests deleted: None.

[...]

PR #6257, opened 2026-09-06
fix(scripts): guard-trigger-coverage reads job-level if:, not just triggers

### LINKED ISSUE #6248: guard-trigger-coverage reads a workflow's trigger, never its job-level `if:`
## Problem

`scripts/check-guard-trigger-coverage.py` asks whether at least one workflow
runs a watched guard with no `paths:`/`paths-ignore:` key under its
`pull_request:` trigger. A guard passes as soon as one such workflow is found.

A trigger with no `paths:` filter is not the same thing as a guard that runs.
`ci.yml`'s `pull_request:` trigger carries no `paths:` key — the prose filter
moved out of the trigger and into the `changes` job in #1892, exactly so the
required contexts still report. So every step in `ci.yml` looks unfiltered to
this guard, including the steps in `check`, whose own
`if: needs.changes.outputs.rust != 'false'` skips the whole job for a
prose-only diff.

A guard whose only invocation is a step in `ci.yml`'s `check` job therefore
satisfies `guard-trigger-coverage` while running nowhere for exactly the
diffs the coverage guard exists to protect. The check cannot distinguish
"this guard runs on every PR" from "this guard is skipped for prose-only
diffs", which is the distinction it was added to make (#5448).

Today's three `WATCHED_GUARDS` — `check-prose.py`, `check-hue-separation.py`,
`check-transcript-surfaces.py` — are unaffected: none is invoked in `ci.yml`
at all, so the workflow the guard finds for each really is
`guard-self-tests.yml`. The blind spot bites the next guard added to that
list that is also a `ci.yml` step.

It is why `scripts/check-closing-keywords.py` was **not** added to
`WATCHED_GUARDS` when #6228 gave it a step in `guard-self-tests.yml`'s
`gate-steps` job: the entry would have passed on `ci.yml` alone and claimed
a protection it does not have. That comment in
`.github/workflows/guard-self-tests.yml` cites this issue.

## Files

- `scripts/check-guard-trigger-coverage.py` — `WATCHED_GUARDS`, and the
  trigger-reading half that would need to learn about job conditions.
- `scripts/test-guard-trigger-coverage.py` — the hermetic self-test; a new
  fixture shape belongs here.
- `.github/workflows/ci.yml` — the `changes` job and `check`'s
  `if: needs.changes.outputs.rust != 'false'`, the condition the guard cannot
  currently see.
- `.github/workflows/guard-self-tests.yml` — the `gate-steps` job, and the
  comment that points here.

## How to verify the gap today

Add `check-closing-keywords.py` to `WATCHED_GUARDS` on a branch, then delete
its step from `guard-self-tests.yml`'s `gate-steps` job. `make
guard-trigger-coverage` still passes, on `ci.yml`'s copy alone — while a
prose-only pull request would run the guard nowhere.

## Definition of done

- [x] The guard reads a step's effective condition, not only its workflow's
      trigger: a step inside a job carrying an `if:` that can evaluate false
      does not by itself satisfy a watched guard.
- [x] A watched guard whose only unconditional invocation is skippable fails
      the check, and the message names the job and the condition.
- [x] `scripts/test-guard-trigger-coverage.py` gains fixtures for both
      directions — a guard reachable only through a conditional job fails, and
      one with an unconditional step beside it passes. Both fail on the code
      as it stands today.
- [x] `check-closing-keywords.py` joins `WATCHED_GUARDS` once the check can
      tell its two invocations apart, and the comment in
      `guard-self-tests.yml` that defers to this issue is removed in the same
      change.
- [x] AGENTS.md's `guard-trigger-coverage` line says what the guard reads,
      since "runs with no `paths:` filter in at least one workflow" is what
      created the wrong expectation.
- [x] **Witness proof**: `scripts/test-guard-trigger-coverage.py`'s new
      fixtures demonstrate the fail→pass flip — the conditional-job fixture is
      what today's code lets through as covered; once the fix lands, that same
      fixture fails the check while the unconditional-step fixture still
      passes. Link the run (or, if run in CI, the workflow run) in the PR.
- [x] **File placement**: the fix stays inside
      `scripts/check-guard-trigger-coverage.py` and
      `scripts/test-guard-trigger-coverage.py` — no new script, no new crate,
      and neither file is on the god-files list
      (`scripts/file-size-baseline.txt` / AGENTS.md's "God files" table).
- [x] **Gate placement**: `guard-trigger-coverage` stays a toolchain-free
      `make gate` step run from `guard-self-tests.yml`'s `gate-steps` job —
      the fix must not make it compile anything or need network/API access.
- [x] **Prose**: the updated failure message and the AGENTS.md line pass
      `make prose` — no content-free filler, and any code reference cites a
      file/symbol rather than a pinned line number (`make line-citations`).
- [x] Not applicable: no UI or design-system surface is touched — this is a
      CI/guard-only change.

## Constraints

- The YAML reading in this script is line-based and only correct for the
  shapes this repository's workflows use; its docstring says so, and a
  general YAML parser is deliberately not the goal. Job-condition reading
  should stay in that register, with fixtures pinning both readings.
- `guard-trigger-coverage` is a `make gate` step and runs in
  `guard-self-tests.yml`'s own `gate-steps` job. It compiles nothing — keep it
  that way.

macanderson:
## What / why

`scripts/check-guard-trigger-coverage.py` asked only whether a workflow's
`pull_request:` trigger carried a `paths:`/`paths-ignore:` key. It never
looked at whether the *job* a guard's `run:` line sits in can itself skip
the guard.

`ci.yml`'s `check` job carries `if: needs.changes.outputs.rust != 'false'`
(the #1892 shape: the `paths:` filter moved out of the trigger and into a
job condition so the required contexts still report for a skipped job).
`ci.yml`'s own `pull_request:` trigger has no `paths:` key at all, so a
guard whose only invocation sat inside that gated `check` job satisfied the
old check while actually running nowhere for a prose-only diff -- exactly
the failure mode `guard-trigger-coverage` exists to catch. That is why
`check-closing-keywords.py` (added to `ci.yml`'s `check` job and to
`guard-self-tests.yml`'s `gate-steps` job in #6228) was deliberately kept
out of `WATCHED_GUARDS`, with a comment pointing at this issue.

## The fix

`scripts/check-guard-trigger-coverage.py` now:
- Parses each workflow's `jobs:` block (`parse_jobs`) instead of scanning
  the whole file for `run:` lines.
- Reads a job's own `if:` (`job_if_expression`), distinguishing it from a
  step's `if:` by indentation.
- Flags a job as diff-gated (`job_diff_gate`) when its `if:` reads a
  `needs.*.outputs.*` value -- the shape a `changes` job's output-based
  skip uses.
- Reports a guard as covered only when at least one invocation sits in a
  job that neither a `paths:`-filtered trigger nor a diff-gated `if:` can
  skip.

`check-closing-keywords.py` joins `WATCHED_GUARDS`: the check can now tell
its `guard-self-tests.yml` copy (genuinely unconditional -- the `gate-steps`
job carries no job-level `if:` at all) from its `ci.yml` copy (inside the
gated `check` job), so it reports the guard as covered by the first without
being fooled by the second. The stale comment in `guard-self-tests.yml`
deferring to this issue is removed and replaced with a comment explaining
the new coverage.

`AGENTS.md`'s `guard-trigger-coverage` description is updated in both
places (the `make gate` step list and the `guard-self-tests.yml` narrative)
to say what the guard actually reads now.

## Witness

`scripts/test-guard-trigger-coverage.py` gained two fixtures:
- **T9**: a `ci.yml`-shaped fixture with a `changes` job and a `check` job
  gated on `needs.changes.outputs.rust != 'false'`, running all four
  watched guards. Fails on the new code (exit 1, naming the job and its
  `if:`); the pre-fix script reports `OK` (exit 0) against the exact same
  fixture, which is the bug this issue names.
- **T10**: a job whose `if:` references `github.event_name` (not a
  `needs.*.outputs.*` value) still counts as unconditional coverage --
  only a diff-derived gate is a gap.

Checked directly, the artisanal way:
```
$ python3 /tmp/old-check-guard-trigger-coverage.py --manifest-dir <T9 fixture>
check-guard-trigger-coverage: OK -- 3 guard(s) each run unconditionally ...

$ python3 scripts/check-guard-trigger-coverage.py --manifest-dir <T9 fixture>
check-guard-trigger-coverage: FAIL -- a scripts-only pull request would not run:
  check-prose.py: every invocation is filtered or conditionally skipped: ci.yml:check (job `check` skipped when `if: ...needs.changes.outputs.rust != 'false'` is false).
  ...
```

All 10 cases in `scripts/test-guard-trigger-coverage.py` pass; the real
guard against this repository's own `.github/workflows/` still reports
`OK` for all four watched guards (the `ci.yml` copies of `hue-separation`
and `closing-keywords` are now correctly classified as "gated" rather than
"ok", but each still has a genuinely unconditional copy elsewhere, so the
overall result is unchanged for the pre-existing three and newly correct
for the fourth).

Exemplar: this mirrors the existing line-based, `--manifest-dir`-driven
self-test pattern `check-file-size.sh` / `check-lockfile-sync.sh` and this
same script already used for the trigger half.

## Tests deleted

None.

## Residue

None noticed beyond what's already tracked -- no new follow-up filed.

Closes #6248

## Summary by Sourcery

Ensure guard trigger coverage detects invocations that can be skipped by diff-derived job conditions as well as path-filtered pull request triggers.

New Features:
- Extend guard trigger coverage checks to recognize job-level diff-based conditions and include the closing-keywords guard.

Bug Fixes:
- Prevent guards in jobs gated by needs.*.outputs.* conditions from being incorrectly treated as unconditionally covered.

Enhancements:
- Parse workflow jobs and report whether each guard invocation is path-filtered, job-gated, or genuinely unconditional.

Documentation:
- Update guard coverage documentation and workflow comments to describe trigger and job-level filtering checks.

Tests:
- Add fixtures covering diff-gated jobs and unrelated job conditions, while keeping coverage fixtures synchronized with the watched guard list.

### DIFF
--- a/.github/workflows/guard-self-tests.yml
+++ b/.github/workflows/guard-self-tests.yml
@@ -123,16 +123,13 @@ jobs:
       # The steps above run here because this trigger carries no `paths:`
       # filter at all -- that is what lets them fire on a scripts-only pull
       # request. Nothing enforced that absence except a comment, so this reads
-      # it back: at least one workflow must run each guard it names with an
-      # unfiltered `pull_request:` trigger, or a future edit narrowing this one
-      # would silently reopen the gap (#5448).
-      #
-      # It does not cover the closing-keywords step above, and adding it to
-      # `WATCHED_GUARDS` would read as protection it cannot give: this guard
-      # reads a workflow's *trigger*, and ci.yml's `pull_request:` trigger
-      # carries no `paths:` key either, so that copy alone would satisfy it
-      # while the job-level `if:` still skipped the guard. #6248 is where that
-      # blind spot is tracked.
+      # it back: at least one workflow must run each guard it names from a
+      # job that neither a `paths:` filter nor a `needs.*.outputs.*`-gated
+      # `if:` can skip, or a future edit narrowing this one would silently
+      # reopen the gap (#5448, #6248). `check-closing-keywords.py` above is
+      # covered by this job's own unconditional invocation even though its
+      # `ci.yml` copy sits inside the `changes`-gated `check` job -- the
+      # guard reads that job condition now, not just the trigger.
       - name: each guard above still runs with no paths filter somewhere
         if: ${{ !cancelled() }}
         run: python3 ./scripts/check-guard-trigger-coverage.py
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -108,9 +108,11 @@ make gate                # = no-scratch + no-secrets + design-refs
                          #   + gate-parity
                          #   + schema-tier-parity (a -schema step runs at the
                          #     same rung as its base step; #5139)
-                         #   + guard-trigger-coverage (prose, hue-separation
-                         #     and transcript-surfaces each run with no
-                         #     paths: filter in at least one workflow)
+                         #   + guard-trigger-coverage (prose, hue-separation,
+                         #     transcript-surfaces and closing-keywords each
+                         #     run, in some workflow, from a job neither a
+                         #     paths: filter nor a needs.*.outputs.*-gated
+                         #     if: can skip)
                          #   + priority-scheme (the issue priority scheme is
                          #     stated once, in SCR-005, and the triage guard's
                          #     regex covers exactly the levels it names)
@@ -209,12 +211,19 @@ says nothing about which files reach it, so a `paths:` filter narrowing
 `guard-self-tests.yml`'s `pull_request:` trigger to `docs/**` would still read
 as green, and the three steps would run nowhere for a scripts-only pull
 request. Nothing held that absence in place except a comment.
-`guard-trigger-coverage` reads the trigger back: each of the three must run
-with no `paths:`/`paths-ignore:` filter in at least one workflow, so a later
-edit narrowing this one cannot reopen the gap silently. That workflow's
-`gate-steps` job — named `gate steps no other workflow runs` — is now one of
-`main`'s required status checks too,
-added after a red run of it merged into `main` and reddened every open PR.
+`guard-trigger-coverage` reads the trigger back: each watched guard must run,
+in some workflow, from a job with no `paths:`/`paths-ignore:` filter above it
+and no `if:` of its own that reads a `needs.*.outputs.*` value — the second
+half added once a guard reached `ci.yml`'s `check` job, whose own
+`if: needs.changes.outputs.rust != 'false'` skips it for a prose-only diff
+even though `ci.yml`'s trigger carries no `paths:` key at all, because that
+filter moved out of the trigger and into the job so the required contexts
+still report for a skipped job. So a later edit narrowing the trigger, or
+moving a guard into a job gated that way, cannot reopen the gap silently.
+That workflow's `gate-steps` job — named `gate
+steps no other workflow runs` — is now one of `main`'s required status
+checks too, added after a red run of it merged into `main` and reddened
+every open PR.
 
 A fifth workflow, `deck-fit.yml`, owns the decks under
 `website/public/presentations/`. Its measurement — every slide against the
--- a/scripts/check-guard-trigger-coverage.py
+++ b/scripts/check-guard-trigger-coverage.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python3
-"""Guard: the three steps `ci.yml`'s required job cannot reach for a
-scripts-only diff still run on one, somewhere.
+"""Guard: the steps `ci.yml`'s required job cannot reach for some diffs
+still run on one, somewhere.
 
 `ci.yml`'s required job skips `prose`, `hue-separation`, and
 `transcript-surfaces` for a diff confined to Rust, scripts, or any path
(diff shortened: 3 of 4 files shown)


macanderson:
PR-management routine: I verified #6248's DoD against this diff and the CI run for `20d2f77`, and ticked 9 of the 10 boxes. One is left, and — as on #6255 an hour ago — it is a defect in the DoD item rather than in this PR.

**Ticked, with what establishes each:**

| Item | Evidence |
| --- | --- |
| Guard reads a step's effective condition | `parse_jobs` splits the `jobs:` block, `job_if_expression` reads the one `if:` at the job body's own indentation, and `main` classifies the invocation `gated` rather than `ok` |
| Failure names the job and the condition | `` f"{wf}:{job} (job `{job}` skipped when `if: {gate}` is false)" ``, asserted by T9 |
| `check-closing-keywords.py` joins `WATCHED_GUARDS`, stale comment removed | both in the diff; I read `guard-self-tests.yml` at `20d2f77` — the `gate-steps` job carries **no** job-level `if:`, and the trigger no `paths:`, so that invocation really is unconditional and the entry is earned rather than asserted |
| AGENTS.md says what the guard reads | updated in both places — the `make gate` step list and the `guard-self-tests.yml` narrative |
| Witness proof + link the run | the fail→pass flip is in the description's old-vs-new output; the run the item asks for is `guard self-tests` — [job 101516478493](https://github.com/macanderson/stella/actions/runs/34044297808/job/101516478493), success, which is all 10 fixtures green **and** the real guard passing against this repository's own `.github/workflows/` with the fourth guard now watched |
| File placement | no new script, no new crate; `file size ratchet` — [job 101516478213](https://github.com/macanderson/stella/actions/runs/34044297794/job/101516478213), success |
| Gate placement stays toolchain-free | the added imports are `importlib.util` and stdlib `re`; no network, no API, no compile. `gate steps no other workflow runs` — [job 101516478545](https://github.com/macanderson/stella/actions/runs/34044297808/job/101516478545), success |
| Prose / line-citations | `docs guards` — [job 101516478141](https://github.com/macanderson/stella/actions/runs/34044297800/job/101516478141), success |
| Not applicable (no UI surface) | true by inspection — four files, none of them a rendering surface |

**One thing I ticked with a narrowing worth naming rather than glossing.** The first item says "an `if:` that can evaluate false does not by itself satisfy a watched guard". What shipped is narrower: only an `if:` matching `needs\.[A-Za-z0-9_-]+\.outputs\.[A-Za-z0-9_-]+` counts as a gate, and T10 pins that a job whose `if:` reads `github.event_name` still counts as coverage. I read that as correct rather than short, and the reason is in this workflow: `guard-self-tests.yml`'s own steps all carry `if: ${{ !cancelled() }}`, which can evaluate false. A guard that refused every falsifiable condition would fail against the one job that exists to satisfy it — the rule would be unsatisfiable. The narrowing is documented in the docstring and pinned from both sides (T9 fails, T10 passes), which is the register the issue's own Constraints section asks for.

**Left unticked — the second fixture cannot be a witness, and the issue's premise about it is wrong.**

The item asks that *both* new fixtures "fail on the code as it stands today". T9 does, which is the flip this PR exists to demonstrate. T10 cannot, and no change to this PR would make it:

- T10's fixture has one `pull_request:` trigger with no `paths:` key, and one job whose `if:` reads `github.event_name`.
- Pre-fix, `pull_request_is_unconditional` returns true and `extract_run_commands` scans the whole file, so every watched guard is covered → exit 0.
- Post-fix, the job carries no `needs.*.outputs.*` gate, so its status is `ok` → exit 0.

Both sides agree by construction, because T10's job is to pin the *boundary* — that the fix does not over-fire and start failing ordinary conditional jobs. A control that flipped would mean the fix had broken something. It is worth keeping; it just is not a witness, and the DoD item asks it to be one.

I did not reword the item — narrowing a gate to make it green is the maintainer's call, not an automation's. Something like "*T9 fails against the pre-fix script; T10 is the boundary control and passes on both*" would match what shipped. Editing the issue body re-runs `dod / dod` on its own through `dod-recheck.yml`, so no push is needed either way.

Everything else here is clean: no merge conflict, Sourcery's assessment table is 3/3 ✅ with nothing unanswered, and every other check on `20d2f77` is green except `fmt + clippy + test` and `Analyze (rust)`, both still running.

---

macanderson:
Checked #6248's witness box by running this branch's `scripts/test-guard-trigger-coverage.py` against `origin/main`'s `scripts/check-guard-trigger-coverage.py`, rather than reading the claim:

```
test-guard-trigger-coverage: 7 passed, 3 failed
  FAIL  T9 a guard reachable only through a needs.changes.outputs-gated job fails -- exit 0, wanted 1
  FAIL  T3 (output wording changed by this PR)
  ok    T10 a job if: unrelated to needs.*.outputs.* still counts as unconditional
```

**T9 is the flip and it holds** — red on `main` (the old guard lets the conditionally-gated job through as coverage, exit 0), green here. That is the objective the issue was filed for.

One correction to the issue's wording rather than to this diff: the DoD item asks for fixtures in both directions and then says *"Both fail on the code as it stands today."* T10 cannot. It is a negative control — it pins that the new job-condition reading does **not** over-fire on an `if:` unrelated to `needs.*.outputs.*` — and a control for behavior that does not exist yet necessarily passes before the change. The clause is unsatisfiable as written for that half; T10 is the right test regardless, and the ticked box is correct in substance.

Worth fixing the phrasing if this DoD shape gets reused, since the same sentence would fail any future negative control. Not raising it as a blocker and filing nothing — nothing is deferred here.

---

Issue #1281, opened 2026-08-03
Move Linear from a built-in code path to a source-tracked TOML issue provider


macanderson:
Phase 1 of #1280. Spec:
[`docs/design/agent-native-delivery.md`](https://github.com/macanderson/stella/blob/claude/ai-agent-workflow-design-shj0gl/docs/design/agent-native-delivery.md)
§4, with the target manifest rendered in full at
[`docs/design/agent-native-delivery/provider.linear.toml`](https://github.com/macanderson/stella/blob/claude/ai-agent-workflow-design-shj0gl/docs/design/agent-native-delivery/provider.linear.toml).

## The problem

Linear support is compiled in. `IssueBackend` is a three-variant enum, and
`stella-tools/src/issue_ops.rs` interleaves Linear's GraphQL, GitHub's REST,
and `gh` shelling through one set of functions. Every fact about Linear is a
branch in shared code:

- the GraphQL endpoint and the `LINEAR_API_KEY` / connection precedence
  (`issues.rs`, `tracker_auth.rs`),
- that `ENG-123` is the identifier shape and how it resolves to an internal id,
- that Linear supplies a canonical branch name per issue
  (`start_work_on_issue`),
- that a team scopes creation,
- and — by omission — that a Linear issue has **no class at all**, which is why
  no per-class delivery policy can ever be applied to a Linear workspace.

Two costs follow. Adding a third tracker means a fourth enum variant and a
fourth branch in every operation. And a customer whose Linear team spells its
states or labels differently has no way to say so short of a patch.

## Definition of done

**Rewritten 2026-09-07, against the tree rather than against the branch this
issue was filed from.** The original list is preserved at the bottom. Three of
its items describe code that does not exist and never reached `main`, and two
more describe schema for keys nothing can read; ticking any of those would be
ticking something untrue. See "Why the list changed" below.

Ship Linear as a provider manifest, so the format is proven adequate:

- [x] Ship Linear embedded with `include_str!`, shadowable by a workspace
      `.stella/issues/linear.toml` of the same name —
      `crates/stella-cli/src/issue_provider/linear.toml`, listed in
      `manifest.rs`'s `BUILT_IN` table. Witness:
      `a_workspace_linear_manifest_shadows_the_shipped_one`.
- [x] Load a provider manifest from `.stella/issues/<provider>.toml`, workspace
      shadowing the shipped file, a malformed file degrading to the built-in
      with a named per-file warning rather than a fatal startup error —
[...]

PR #6265, opened 2026-09-06
feat(stella-cli): ship GitHub as a provider manifest under .stella/issues/

### LINKED ISSUE #6253: GitHub's issue provider is compiled in, so the loop cannot be pointed at another tracker by editing a file
The self-driving loop reads its issues through `IssueProvider`
(`crates/stella-protocol/src/issue.rs`) and one shipped adapter,
`crates/stella-cli/src/issue_provider.rs::GhIssueProvider`. Which tracker the
loop talks to is therefore still a fact compiled into the binary rather than a
file somebody can read or change.

Epic #3599's phase B1 asked for three things. Two of them landed: the `Issue`
kernel and the `IssueProvider` port, and the ranked queue proved against a
fixture provider with no `gh` on `PATH`
(`the_backlog_generator_picks_the_ready_issue_and_records_the_delivered_cycle`
in `crates/stella-cli/src/self_driving_cmd/ready.rs`). The third — **GitHub as
a shipped manifest under `.stella/issues/`** — did not.

## What is there today

`crates/stella-cli/src/self_driving_cmd/config.rs::vocabulary_for` reads
`.stella/issues/<provider>.toml` and uses it for the tracker's **vocabulary**:
which status words mean open, which mean closed, how a resolution is spelled.
An absent file falls back to `builtin_vocabulary`, which hands back
`Vocabulary::github()` — compiled in.

So the seam exists for one narrow fact and for nothing else. There is no
`.stella/issues/` directory in the repository at all, GitHub ships no manifest,
and everything else about the provider — the transport, the identifier shape,
the field names, the class mapping — is Rust. `issue_provider.rs`'s own module
header says as much on `main`: *"a manifest-driven provider is the next
slice."*

## Why it matters

A workspace on a tracker this build has no adapter for gets a warning and
GitHub's vocabulary, which is a guess. A workspace on GitHub Enterprise with
different label conventions has nothing to edit. And every future tracker is a
new Rust module rather than a new file, which is the cost #1281 already names
for Linear.

## What to build

Ship GitHub the way a third-party tracker would have to be shipped, so the
built-in path and the manifest path are the same path.

- Give the shipped GitHub provider a real manifest and embed it with
  `include_str!` (the seed-skill pattern), shadowable by a workspace
  `.stella/issues/github.toml` of the same name.
- Widen what the manifest carries beyond `Vocabulary` — at minimum the class
  mapping and the field names — so the manifest is the declaration and the Rust
  adapter is only the transport.
- Keep `builtin_vocabulary`'s "absent is always answerable" contract: no
  configuration must still yield a working loop against GitHub
  (`a_workspace_with_no_config_still_works` in
  `self_driving_cmd/config.rs` pins that and must stay green).

## Related

- #1281 defines the provider-manifest schema and its `.stella/issues/*.toml`
  discovery, scoped to moving Linear off its built-in path. That issue is the
  schema; this one is GitHub adopting it. Land the schema first, or land both
  together — do not define a second schema here.
- #5558 moves the delivery loop's direct `gh` calls behind a provider-neutral
  pull-request interface. Same shape, different surface.
- Epic #3599 phase B1 is where this clause comes from.

## Definition of done

- [x] A GitHub provider manifest is shipped in the tree and embedded, and a
      workspace `.stella/issues/github.toml` shadows it.
- [x] The manifest, not Rust, is where GitHub's vocabulary and class mapping
      are declared; `builtin_vocabulary`'s hard-coded `Vocabulary::github()`
      is gone or reduced to reading the embedded manifest.
- [x] A witness test fails before the change and passes after: a workspace
      with no configuration resolves GitHub's vocabulary **through the shipped
      manifest**, and a workspace `.stella/issues/github.toml` overrides it.
- [x] `a_workspace_with_no_config_still_works` and
      `a_malformed_manifest_falls_back_rather_than_failing` both still pass.
- [x] `issue_provider.rs`'s header no longer says a manifest-driven provider
      is the next slice.

macanderson:
## What

GitHub ships as a provider manifest, the way any other tracker would have to.

- `crates/stella-cli/src/issue_provider/github.toml` — the shipped manifest,
  tracked and embedded with `include_str!`. It carries the identity keys
  (`schema_version`, `name`, `kind`), the vocabulary `Vocabulary::github()`
  held, and `[classes]`, the label taxonomy `GhIssue::class` hard-coded.
- `crates/stella-cli/src/issue_provider/manifest.rs` — `ProviderManifest` and
  `ClassMap`. `resolve` reads a workspace `.stella/issues/<provider>.toml` over
  the embedded one; `for_workspace` is the same thing for a caller holding a
  root and no loop config.
- `config::load` resolves the manifest once and holds it on `LoopConfig`;
  `LoopConfig::vocabulary()` reads it out. `GhIssueProvider` carries the class
  map and decodes through it, so the class comes from the file and not from a
  `match` in Rust.

## Why

Which tracker the self-driving loop talks to was a fact compiled into the
binary. A workspace on GitHub Enterprise with different label conventions had
nothing to edit, and every future tracker was a new Rust module rather than a
new file. This makes the built-in path and the manifest path one path.

## Witness mechanism

Two witnesses, both failing on `main` by construction because `GhIssue::class`
there takes no manifest input at all:

- `a_workspace_manifest_drives_the_ready_queue`
  (`self_driving_cmd/ready.rs`) — a workspace `.stella/issues/github.toml`
  renames the label set (`bug = ["kind/defect"]`), `config::load` resolves it,
  the shipped decode maps a `gh`-shaped payload through it, and the readiness
  fold ranks the result. **No `gh` runs**: `decode_list` takes the payload as
  text and spawns nothing, so the whole path is provable on a machine with no
  GitHub CLI. On `main` the class is a `match` on label names inside
  `issue_provider.rs`, so `kind/defect` could not become `IssueClass::Bug` at
  any price.
- `a_renamed_label_set_in_the_manifest_reaches_the_decode`
  (`issue_provider.rs`) — the same claim at the decode boundary, and the
  negative control beside it: `bug` reads as `Other` once a manifest replaces
  it.

Beside them:

- `the_shipped_manifest_parses_and_matches_the_compiled_table` — the embedded
  file parses, and it equals `Vocabulary::github()` plus the old class map, so
  the move changes nothing. This is what keeps the compiled last resort from
  drifting away from the file it stands in for.
- `github_is_resolved_through_the_shipped_manifest` and
  `a_workspace_github_manifest_shadows_the_shipped_one`
  (`self_driving_cmd/config.rs`) — the DoD's third box, from both directions.
- `a_workspace_with_no_config_still_works` and
  `a_malformed_manifest_falls_back_rather_than_failing` are unchanged and
  still pass.

## Exemplar

The seed-skill pattern already in this tree: a tracked file under `src/`,
embedded with `include_str!`, shadowed by a workspace copy of the same name.
`docs/spec/agent-native-delivery/provider.linear.toml` is the rendered target
shape #1281 works from, and this file follows it.

## Design notes worth a reviewer's eye

- **Two parse passes, not `#[serde(flatten)]`.** `ProviderManifest::parse`
  reads the same TOML text twice, once into `Vocabulary` and once into a
  private `Header`. That keeps `Vocabulary` the only declaration of the words
  it owns — a field added there needs no copy in this crate — and it is why
  today's `.stella/issues/jira.toml` files keep loading unchanged.
- **`[classes]` is keyed by the kernel's class names** (`bug`, `feature`,
  `task`). `doc:agent-native-delivery` §4.1 renders the block with `defect` and
  `epic`; `IssueClass` is what shipped, so the file follows the code. `#1281`
  owns reconciling the two, and the manifest says so in a comment.
- **A file with no `[classes]` block inherits the shipped map**, rather than
  classing every issue as `Other`. Before this change the adapter read the same
  labels whatever the tracker was, so a manifest written before the block
  existed must not lose them.
- **Scope boundary with #1281.** `[connection]`, `[states]`, `[capabilities]`
  and `[fields.write]` are not defined here. No second schema.

## Tests deleted

None.

## Residue filed

None found beyond what #1281 already tracks. The rest of §4.1's schema and
Linear's move onto it are that issue's, and this PR's manifest and module docs
both point at it rather than pre-empting it.

Closes #6253

## Summary by Sourcery

Make GitHub’s issue vocabulary and label classification manifest-driven while preserving the embedded defaults and supporting workspace-specific overrides.

New Features:
- Ship GitHub’s provider vocabulary and label taxonomy as an embedded TOML manifest that workspaces can override under `.stella/issues/github.toml`.
- Use manifest-defined label mappings to classify GitHub issues and propagate the resolved provider configuration throughout issue commands and self-driving workflows.

Bug Fixes:
- Allow workspace-specific status, resolution, field, and class conventions to be honored instead of relying exclusively on compiled GitHub defaults.
- Preserve the built-in GitHub behavior when manifests are missing, malformed, or omit the class mapping.

Enhancements:
- Resolve provider manifests once in loop configuration while retaining workspace resolution for standalone issue operations.
- Add fallback diagnostics and compatibility handling for unsupported provider metadata and schema versions.

Documentation:
- Document provider manifests and workspace overrides in the repository guidance and issue-provider module documentation.

Tests:
- Add coverage for embedded-manifest parity, workspace shadowing, inherited class mappings, renamed labels, and manifest-driven ready-queue classification.


macanderson:
## `a_workspace_manifest_drives_the_ready_queue` was failing on `11e49c7` — root cause and fix pushed as `a220e0e`

Running this branch's own headline witness before ticking #6253's boxes:

```
$ cargo test -p stella-cli --bins -- a_workspace_manifest_drives_the_ready_queue
assertion `left == right` failed: `kind/defect` means bug because this workspace's manifest says so
  left: Other
 right: Bug
```

**It is the product, not the test.** `config::load` opens with

```rust
let Some(parsed) = read_toml(root) else {
    return LoopConfig::default();   // ← carries ProviderManifest::embedded()
};
```

so a workspace with no `stella.toml` never reaches `ProviderManifest::resolve`
and its `.stella/issues/github.toml` is never read.
`a_workspace_github_manifest_shadows_the_shipped_one` passes only because it
writes a `stella.toml` beside the manifest.

That makes the shadow conditional on a second file whose entire content would
be `provider = "github"` — already the `IssuesSection::default()`. The
workspace the manifest exists for, the one on GitHub Enterprise that edited
label conventions and nothing else, is exactly the workspace it did not reach.
DoD box 1 ("a workspace `.stella/issues/github.toml` shadows it") was not true
as written before this commit.

**The fix** routes the no-config branch through
`ProviderManifest::for_workspace`, the seam already written for "a caller that
holds a root and no loop config" — it reads the same `[issues]` section and
falls back to `IssuesSection::default()`, so the path resolves to
`.stella/issues/github.toml`.
`a_manifest_shadows_the_shipped_one_with_no_stella_toml_beside_it` states the
rule at the config boundary, where the defect is; the ready-queue witness
already stated it end to end.

### Evidence

| Check | Result |
|---|---|
| `cargo test -p stella-cli --bins -- issue_provider self_driving_cmd::config self_driving_cmd::ready` | 34 passed, 0 failed |
| `cargo fmt -p stella-cli --check` | clean |
| `cargo clippy -p stella-cli --bins -- -D warnings` | clean |
| `check-prose` / `check-line-citations` / `check-file-size` | all OK, none added |

CI on the new head is the rest of it.

### #6253's DoD — ticked, with what settled each box

Checked against the tree rather than the PR body, which is why the failure
above surfaced:

1. **Manifest shipped and embedded, workspace shadows it** — `issue_provider/github.toml`, `include_str!` at `manifest.rs`'s `EMBEDDED_GITHUB`; the shadow half needed `a220e0e` to be true with no `stella.toml`.
2. **Manifest, not Rust, declares vocabulary and classes** — `builtin_vocabulary` is gone; `Vocabulary::github()` survives only inside `compiled_github()`, the last resort for an embedded file this build cannot parse, held equal to the shipped file by `the_shipped_manifest_parses_and_matches_the_compiled_table`.
3. **Witness fails before, passes after** — `github_is_resolved_through_the_shipped_manifest` and `a_workspace_github_manifest_shadows_the_shipped_one` both pass here. On `main` they cannot compile at all: `ProviderManifest` does not exist there (`git grep ProviderManifest origin/main -- crates/stella-cli/` is empty). That is a stronger fail-before than a red assertion, and worth stating as what it is.
4. **Both named regressions still pass** — `a_workspace_with_no_config_still_works` and `a_malformed_manifest_falls_back_rather_than_failing`, in the 34 above.
5. **Header no longer says "next slice"** — `git grep "next slice"` over `issue_provider.rs` is empty.

Sourcery's assessment table marks all three linked objectives ✅ and there is no
❌ row to answer.

---

macanderson:
Ticked #6253's five DoD boxes after verifying each against this branch (`11e49c7`). Evidence per box, so the ticks are auditable rather than taken on trust:

1. **Manifest shipped and embedded, workspace file shadows it** — `crates/stella-cli/src/issue_provider/github.toml` is added by this PR; `manifest.rs`'s `EMBEDDED_GITHUB` embeds it with `include_str!("github.toml")`. `Manifest::resolve` reads `[issues] manifest` or `.stella/issues/<provider>.toml`, and a missing/unreadable file yields the shipped manifest rather than an error.
2. **Manifest, not Rust, declares vocabulary and class mapping** — `builtin_vocabulary` is gone from the tree (`rg builtin_vocabulary` returns nothing). The one surviving `Vocabulary::github()` in shipping code is inside `compiled_github()`, which `embedded()` reads *only* when the embedded file fails to parse, and `the_shipped_manifest_parses_and_matches_the_compiled_table` holds the two equal. That is the "reduced to reading the embedded manifest" arm of the box.
3. **Witness test for no-config-through-manifest and workspace override** — `github_is_resolved_through_the_shipped_manifest` and `a_workspace_github_manifest_shadows_the_shipped_one`, plus `an_unconfigured_workspace_gets_the_shipped_manifest` and `a_workspace_manifest_shadows_the_embedded_one` in `issue_provider/manifest.rs`.
4. **Both named tests still pass** — ran targeted: `cargo test -p stella-cli --bins config::tests` → `82 passed; 0 failed`, including `a_workspace_with_no_config_still_works` and `a_malformed_manifest_falls_back_rather_than_failing`.
5. **`issue_provider.rs` header no longer says "next slice"** — `rg 'next slice' crates/stella-cli/src/issue_provider.rs` returns nothing.

No gaps found, so no follow-up issue is owed on this checklist. `dod-recheck` should re-run `dod / dod` off the issue edit; the remaining required checks are CI's own.

---

macanderson:
Residue filed after the merge, both `triage` only (SCR-004/005):

- #6275 — `doc:agent-native-delivery` §4.1 renders `[classes]` with `defect`/`feature`/`epic`, and the shipped loader reads the kernel's `bug`/`feature`/`task`. This PR chose the code's names and said so in `github.toml`; the spec and its two rendered manifests still disagree, so a reader who copies §4.1 gets a block that classes nothing.
- #6276 — an explicit `[issues] provider = ""` builds the manifest path `.stella/issues/.toml`. Pre-existing behaviour, preserved rather than decided by this change.

Sourcery's assessment table marked all three linked objectives ✅ with no ❌ rows.

Issue #6264, opened 2026-09-06
Nothing calls plan_repair: wire the repair-admission decision to the verification path, or delete it


macanderson:
`crates/stella-plugin/src/repair.rs` decides whether a refuted success claim
earns another attempt. `plan_repair` is a pure function over measurements the
caller supplies: an attempt cap, the caller's configured revision allowance,
and two optional measured axes (budget headroom in USD, wall clock left).

Nothing in this workspace calls it.

Its old caller was `stella-pipeline`'s `repair_gate.rs`. That crate was
deleted (#3865), and this module sat in `stella-core` with no caller until
#5116 moved it here, beside `candidate_grant.rs` — the other pure piece of
the same surface that survived the deletion. The move settled where it lives.
It did not settle whether anything should call it.

Two answers are possible, and one of them has to be chosen:

1. **Wire it.** The verification surface a plugin drives is
   `crates/stella-runtime/src/wrapper/` plus `crates/stella-plugin`'s
   `[oracle]` block (`src/oracle.rs`, `src/evidence.rs`). If a host that
   evaluates a plugin's verdict should be able to grant another attempt, this
   is the arithmetic that grants it, and the host needs to pass it the
   measurements.
2. **Delete it.** If the repair decision belongs on the plugin's own side of
   the wrapper socket, then a copy of it here is code with no future caller,
   and `doc:pipeline-as-plugins` §8 is where the logic is ported instead.

## How to verify the premise

```
rg -n 'plan_repair|RepairPlan|RepairHeadroom|RepairRefusal' crates/ --glob '!crates/stella-plugin/src/repair.rs'
```

returns nothing but a serde alias string in
`crates/stella-protocol/src/event/call_role.rs`, which is an unrelated wire
name.

## Constraints

- `stella-plugin` is a near-leaf: `stella-protocol` is its only workspace
  dependency (AGENTS.md's crate table). Wiring must not add an edge to
  `stella-core`; the doc comment already names
  `stella_core::driver::EngineConfig::turn_budget` in plain backticks rather
  than as a link for exactly this reason.
- `stella-core` must never depend on `stella-plugin` (AGENTS.md invariant 1).
- Deleting it means deleting its tests, which must be named in the PR
  description (`check-deleted-tests`).
- #1506 is a related open issue about where the attempt cap should be
  configured. Read it before choosing answer 1.

## Definition of done

- [ ] A decision is recorded: either a caller exists, or the module and its
      tests are gone.
[...]

Issue #6142, opened 2026-09-06
Six stella-core modules the engine never reaches have no eviction issue


macanderson:
Six modules keep `stella-core` off the step path and no issue claims them.

`scripts/check-core-reachability.py` run against `origin/main` reports twelve
modules the engine never reaches from `driver` / `step` / `ports`. Epic #5113
cannot close until `scripts/core-reachability-baseline.txt` is empty, and the
open children only account for six of those twelve:

- #5116 claims `search`, `scoreboard`, `repair`, `tool_foundry`, `extensions`
  (its other two named modules, `starvation` and `tasks`, are reachable today
  and are not baseline lines).
- #3911 (roleless-core slice 7) claims `goal`.

Nothing claims the remaining six. This issue is that handoff.

## The six, and who actually uses each

Every one lives under `crates/stella-core/src/`. The consumer lists come from
`git grep -l 'stella_core::<module>' -- crates` on `origin/main`.

| Module | File | Reached by |
|---|---|---|
| `event_stream` | `crates/stella-core/src/event_stream.rs` | only `crates/stella-core/tests/stream_conformance.rs` — a conformance reader over a recorded `AgentEvent` stream, with no production caller inside or outside the crate |
| `mcp_usage` | `crates/stella-core/src/mcp_usage.rs` | `crates/stella-mcp/src/toolset.rs`, `crates/stella-tools/src/registry.rs`, `crates/stella-cli/src/agent.rs`, `crates/stella-cli/src/agent/audit_writes.rs` |
| `plan_graph` | `crates/stella-core/src/plan_graph.rs` + `plan_graph/` | `crates/stella-cli/src/command_deck/start_work.rs`, `crates/stella-cli/src/command_deck/task_tap/plan_gate.rs`, `crates/stella-tui/src/deck_ui/gates.rs`; the wire and store halves already live in `crates/stella-protocol/src/plan_graph.rs` and `crates/stella-store/src/plan_graph.rs` |
| `subagent` | `crates/stella-core/src/subagent.rs` + `subagent/` | `crates/stella-cli/src/agent/engine.rs`, `crates/stella-cli/src/agent/goal.rs`, `crates/stella-cli/src/candidate_workspaces.rs` |
[...]

Issue #3911, opened 2026-08-19
roleless-core slice 7: goal and monitor leave core as the reference plugin


macanderson:
Slice 7 of #3903. Spec: `doc:roleless-core` §6. **Blocked** on
`doc:turn-loop-wrappers` §9.2.

## Problem

`stella goal` (and `stella monitor`, which is a goal loop with a pinned goal) is
the **last stage core performs** and the **last multi-model path core ships**:

- `crates/stella-cli/src/agent/goal.rs:682` calls `resolve_cross_family_verifier`,
  which builds a second provider adapter from a different model family.
- `crates/stella-core/src/goal.rs:387` runs a verifier sub-agent at
  `ModelCallRole::Verdict`.

`doc:turn-loop-wrappers` §4 already designates goal mode as the **second
plugin** ("Its rounds are `again?`; its independent verifier is `judge`") and
states the target as "One loop. One wrapper socket. Zero built-in wrappers."
Plugin #1 landed as a deletion (#3865); this one has not moved.

It is also actively hostile to plugins today: `reject_arbiter_wrapper_on_goal`
(#3832) refuses **every** arbiter-grade plugin on `stella goal`, because "the
goal loop is this door's own completion arbiter". Core's built-in loop is
occupying the arbiter seat and locking out any plugin that wants it.

## The blocker

`crates/stella-cli/src/agent/goal/goal_wrapped.rs` states it plainly:
`GoalVerifierVerdict`'s **free-text feedback** has no slot in
`stella_plugin::EvidenceSet`'s flip/tamper/measurement vocabulary, and
`doc:turn-loop-wrappers` §9.2 names the gap without settling the encoding.

That prose is not cosmetic — `verifier_feedback_text` is what steers the next
round. Extraction cannot start until the encoding is designed.

§9.2 does settle the other half: goal mode's verifier **is** a model call, and
`judge` may not make one, so **the model call belongs to `after_turn`**, which
returns the parsed assessment as evidence; `judge` then maps evidence to verdict
deterministically.

## Also required

- #3838: no `ChildTurns` binding serves a verifier role intent, so a
  goal-supervision arbiter's `after_turn` cannot spend its verifier call at all.
  Slice 1 of #3903 subsumes this by removing the word list entirely.

## UX constraint

`stella goal` works today with no install. As a plugin it needs
`stella plugin install` first. Keep the verb and have it resolve to an installed
plugin with an actionable refusal — the pattern `--pipeline classic` already
uses after #3865.

[...]

PR #6266, opened 2026-09-06
refactor(stella-core): evict the five modules the engine never reached

### LINKED ISSUE #5116: re-layering: evict the CLI-only residents of stella-core — search, scoreboard, starvation, repair, tasks, tool_foundry, extensions
Child of #5113. The cheap eviction: no new subsystem crate, every module
follows its only consumer. Verified present in `crates/stella-core/src/`
today.

## Moves

| Module | Destination | Rationale |
|---|---|---|
| `search.rs` | `stella-tools` | The decision half of a *tool*, consumed only by the tool crate — the cleanest boundary violation in the tree |
| `scoreboard.rs` + `scoreboard/` | `stella-cli` | Sole consumer |
| `tasks.rs` | `stella-tools` | Decision logic for the `task_*` tools |
| `tool_foundry.rs` | `stella-tools` | Capability-gap detection over shell history; a tool-surface concern |
| `extensions.rs` | `stella-cli` (or `stella-plugin` if its parsing discipline fits — see pushback 2) | Plans/parses only; the CLI owns all its I/O already |
| `repair.rs` | its actual consumer on today's main — **re-locate first** (the v0.9.40 consumer, `stella-pipeline/repair_gate.rs`, no longer exists; find where the repair-gate logic went when the pipeline dissolved, and follow it) | A pure function follows its consumer |
| `starvation.rs` | see pushback 1 | |

Each move is `git mv` + import fixes + a `pub use` shim only if an external
crate consumes the path today. Acceptance per move: `rg 'stella_core::(search|scoreboard|tasks|tool_foundry|extensions|repair|starvation)'`
returns nothing outside the new home.

## Pushback recorded from the proposal doc

1. **`starvation.rs` must NOT go to `stella-protocol`.** The doc proposed it
   ("pure arithmetic over budget types") — but `stella-protocol`'s rule is
   zero logic, and the doc *already* spends its one justified bend on
   `record_hash` (see the stella-records child). Two bends in one proposal is
   how the protocol crate becomes a junk drawer with a wire format attached.
   Put it in the consumer that folds it (`stella-cli`), or a tiny leaf if two
   consumers genuinely remain — the `stella-diag`/`stella-tty` shape the
   workspace already uses for exactly this.
2. **Do NOT create the doc's proposed `stella-manifest` crate.** That slot is
   taken: `stella-plugin` is the manifest crate now (pure parse/validate, no
   I/O, host binds elsewhere — its `lib.rs` states the same split
   `extensions.rs`'s module doc does). Either `extensions.rs` joins its
   consumer, or — only if a real second consumer exists — it joins
   `stella-plugin`'s discipline. A third manifest-parsing crate is a boundary
   with no tenant.

## Witness

The `check-core-reachability` baseline (sibling child) shrinks by seven
entries; `stella-core/src/lib.rs` loses the corresponding `pub mod` lines;
the moved modules' own test suites pass in their new crates.

---
**Suggested worker:** `model-tier:balanced` · **Estimated autonomous time:** ~75 min

## Definition of done

Nothing here is new scope.

The first box counts five, not seven. `tasks.rs` and `starvation.rs` are not
in the reachability baseline at all — the engine's step path reaches both, so
they were never debt this issue could pay. Of the seven baseline entries,
`goal` belongs to #3911 and `subagent` to #6142, so five are this issue's.

- [x] The `check-core-reachability` baseline loses `extensions`, `repair`,
      `scoreboard`, `search` and `tool_foundry`.
- [x] `stella-core/src/lib.rs` loses the corresponding `pub mod` and
      `pub use` lines.
- [x] Every caller names the new home:
      `rg 'stella_core::(search|scoreboard|extensions|repair|tool_foundry)'`
      returns nothing.
- [x] The moved modules' own test suites pass in their new crates (CI green).

macanderson:
## What

Five modules leave `stella-core`. Each follows the code that uses it. The
engine's step path reached none of them — that is what
`scripts/core-reachability-baseline.txt` recorded, and this PR deletes five
of its seven lines.

| Module | New home | Why |
|---|---|---|
| `search.rs` | `crates/stella-tools/src/search/budget.rs` | `stella-tools` was its only caller: `search/engine.rs`, `search/enrich.rs` and that crate's tests. It joins the `search/` module it already served. |
| `extensions.rs` | `crates/stella-cli/src/extensions/plan.rs` | Only `stella-cli` used it, and that crate already owned the file-I/O half in `extensions.rs`. The pure half becomes its child. The name is the module's own header: it only plans and parses. |
| `tool_foundry.rs` | `crates/stella-cli/src/tool_foundry/detect.rs` | Only `stella-cli` used it, beside the two feeders that call it (`tool_foundry/gaps.rs`, `memory/replay`). |
| `repair.rs` | `crates/stella-plugin/src/repair.rs` | Nothing in the tree calls it. Its old caller was `stella-pipeline`'s `repair_gate.rs`, deleted in #3865; the other pure piece that survived that deletion, `candidate_grant.rs`, went to `stella-plugin`. Same shape: pure decision logic for the verification-plugin surface, no I/O, and it does not even need `stella-protocol`. |
| `scoreboard.rs` + `scoreboard/tests.rs` | `crates/stella-learn/src/scoreboard.rs` | **Diverges from the design comment**, which said `stella-cli`. See below. |

## Why `scoreboard` went to `stella-learn` and not `stella-cli`

`stella-cli` is a binary crate — no `src/lib.rs`. A `pub` item there has no
external caller by construction, so `dead_code` judges it like any private
one. Moving this module in makes real dead code out of a public vocabulary:

- `Verdict::Mixed` is constructed only under `#[cfg(test)]`
  (`crates/stella-cli/src/scoreboard_cmd.rs` matches it in `verdict_cell`,
  and matching a variant is not constructing it).
- `Trend::good`, `Trend::bad` and `Trend::mean_interrupts` are written by
  `summarize` and read nowhere outside tests.

Each of those is a `-D warnings` clippy failure in a binary and none of them
is a defect in the module — the vocabulary is right, the CLI renderer just
uses part of it. AGENTS.md's remedies (`#[cfg(test)]`, deletion) do not
apply to an enum variant a non-test `match` arm reads.

`stella-learn` is a library, so `pub mod scoreboard` keeps the vocabulary
public and the lint silent. It is also where the module's own siblings said
it belonged: `crates/stella-learn/src/comparison.rs` and `self_tuning.rs`
both opened with "Like `stella_core::scoreboard`, this module does no I/O",
and both now link `crate::scoreboard`. `comparison` and `self_tuning` ask
whether a change helped; the scoreboard is the yardstick they read.

## No shim stays behind

The issue allowed a `pub use` in `stella-core` for one release. None is
possible or needed:

- `search` — a re-export would make `stella-core` depend on `stella-tools`,
  which is a cycle.
- `extensions`, `tool_foundry` — the new home is a binary crate. Nothing
  outside it can import the path, so there is nothing to keep working.
- `scoreboard`, `repair` — the only caller (`scoreboard_cmd`) and no caller
  respectively, both updated in this PR.

The issue's own acceptance check is the evidence:
`rg 'stella_core::(search|scoreboard|extensions|repair|tool_foundry)'`
returns nothing across the repository.

## Witness mechanism

`scripts/core-reachability-baseline.txt` is a down-only ratchet
(`scripts/check-core-reachability.py`, `make core-reachability`, a required
`ci` step). Deleting the five lines **fails on the old tree by
construction**: on `main` those five modules still sit in `stella-core`
outside the step-path closure, so the guard reports them as new debt and
`--update` refuses to record them back. It passes only once the modules are
gone from the crate. `./scripts/test-core-reachability.sh` (guard-self-tests)
covers the guard's own directions and passes; the two prose lines I edited
in that script are covered by its `--update retires entries whose modules are
gone` case.

The moved modules keep their own test suites and now run them in their new
crates. Every destination crate already had `proptest` as a dev-dependency,
which `detect.rs`, `repair.rs` and `budget.rs` need.

## Not seven, five

The issue's DoD said the baseline shrinks by seven. It shrinks by five, and
the issue body is edited to say so:

- `tasks.rs` and `starvation.rs` are **not in the baseline at all** — the
  engine's step path reaches both, so they were never debt this issue could
  pay. The issue's Moves table listed them from before the guard existed.
- Of the seven baseline entries, `goal` belongs to #3911 and `subagent` to
  #6142. Those two remain.

## Other changes in the same diff

- `toml` was `stella-core`'s dependency only for `extensions.rs`. Both the
  dependency and the dev-dependency go, with the single matching line dropped
  from `Cargo.lock` (`make guards-fast`'s `lockfile-sync` confirms the lock
  still resolves).
- Doc links that crossed the move are repointed:
  `detect.rs`'s `crate::loop_detect` links become `stella_core::loop_detect`;
  `repair.rs`'s `crate::driver::EngineConfig::turn_budget` links become plain
  backticks, because `stella-plugin` must not take an edge to `stella-core`.
- Path citations that named the old files are updated:
  `docs/spec/search-tool.md`, `docs/spec/trace-replay-learning-harness.md`,
  ADRs 0023 and 0025, the `turn-loop` deck (`deck-paths` checks that one),
  and the `stella-core`, `stella-learn`, `stella-plugin` and `stella-tools`
  READMEs.
- `check-core-reachability.py`'s retirement message said "the engine reaches
  it now", which is wrong for a module that left the crate. It now says
  either.
- `crates/stella-tools/tests/fixtures/name_rung_corpus.jsonl` still carries
  the old paths and is left alone: it is a frozen calibration sample read by
  `tests/name_rung_calibration.rs`, not a set of live citations, and
  re-harvesting it would move the numbers that test pins.

## Exemplar

The `foo.rs` beside `foo/` split used for `extensions/plan.rs` and
`tool_foundry/detect.rs` follows the crate's own neighbours
(`tool_foundry/gaps.rs`, `search/enrich.rs`), which is the Rust 2018 layout
`cargo` and `rust-analyzer` use.

## Local verification

`make guards-fast` green (includes `core-reachability`,
`module-reachability`, `prose`, `line-citations`, `lockfile-sync`,
`fmt --check`), plus `make invariants doc-links design-refs god-files
brand-case command-docs` and `./scripts/check-file-size.sh`. Nothing was
compiled locally; CI is the evidence for `lint`, `test` and `doc-warnings`.

Tests deleted: None.

Residue filed: #6264 — nothing in the workspace calls `plan_repair`; wire it
to the verification path or delete it. Cited from `repair.rs`'s header and
`stella-plugin`'s README.

Closes #5116
Refs #5113

## Summary by Sourcery

Move engine-unreachable pure modules out of stella-core and into their owning crates, removing obsolete exports and dependency edges while updating reachability checks and documentation.

Enhancements:
- Move five engine-unreachable pure modules from stella-core into the crates that use or own them, including search budgeting, extension planning, tool-gap detection, repair planning, and scoreboard calculations.
- Remove obsolete stella-core exports and dependencies while preserving module tests and updating callers and internal references.
- Keep the core reachability ratchet and related repository documentation aligned with the retired modules and their new locations.

Build:
- Remove the unused toml dependency from stella-core and synchronize the lockfile.

CI:
- Update the core-reachability guard messaging and baseline to account for modules retired by leaving stella-core.

Documentation:
- Update READMEs, specifications, ADRs, and the turn-loop presentation to reflect the modules' new ownership and paths.

Tests:
- Retain and relocate the moved modules' existing test suites, with test-only APIs constrained where production code no longer uses them.


macanderson:
Ticked #5116's four DoD boxes after verifying each against `5281db2`. Evidence per box:

1. **Baseline loses the five** — `scripts/core-reachability-baseline.txt` now holds only `goal` and `subagent`, which is exactly the split the issue's own preamble predicted (`goal` → #3911, `subagent` → #6142). `extensions`, `repair`, `scoreboard`, `search`, `tool_foundry` are all gone.
2. **`stella-core/src/lib.rs` loses the `pub mod`/`pub use` lines** — `rg 'pub (mod|use) (extensions|repair|scoreboard|search|tool_foundry)' crates/stella-core/src/lib.rs` returns nothing.
3. **Every caller names the new home** — the issue's own acceptance command, `rg 'stella_core::(search|scoreboard|extensions|repair|tool_foundry)'`, returns nothing across the tree.
4. **The moved modules' suites pass in their new crates** — ran each targeted rather than a workspace suite, per the issue's own resource note:

   | Module | New home | Result |
   |---|---|---|
   | `repair` | `stella-plugin` | 10 passed, 0 failed |
   | `scoreboard` | `stella-learn` | 17 passed, 0 failed |
   | `search` | `stella-tools` | 113 passed, 0 failed |
   | `extensions` | `stella-cli` | 66 passed, 0 failed |
   | `tool_foundry` | `stella-cli` | 64 passed, 0 failed |

   The "(CI green)" half of that box is this PR's own run, which is in flight on this head.

**One deviation worth a reviewer's eye, not a gap:** `scoreboard` landed in `stella-learn`, where the issue's table proposed `stella-cli`. The DoD boxes ask only that the baseline lose the entry and that callers name the new home, and both hold — but the table and the landing differ, so it is stated here rather than left for someone to notice.

Separately, `docs/wire still matches the types` was failing on the previous head (`63c71e9`) — `make doc-warnings-schema` broke on two unresolved intra-doc links in `repair.rs`'s `//!` header, collateral from the module landing beside a `mod` declaration that carried its own outer doc comment. That was fixed in `5281db2` and I confirmed it locally: `cargo clean --doc` then `RUSTDOCFLAGS="-D warnings" cargo doc -p stella-plugin --features stella-plugin/schema --no-deps` exits 0. The `cargo clean --doc` matters — cargo's freshness check reports a stale doc run as up to date.

---

macanderson:
PR-management routine: I verified #5116's DoD against head `9d54b83` and ticked three of its four boxes. The fourth is the only thing left holding `dod / dod`, and it is waiting on a settled CI run rather than on this diff.

**Ticked, with what establishes each — each checked against the branch, not against the description:**

| Box | Evidence |
| --- | --- |
| Baseline loses the five | `scripts/core-reachability-baseline.txt` holds `goal` and `subagent` and nothing else. At the merge base it held seven; the five that left are exactly `extensions`, `repair`, `scoreboard`, `search`, `tool_foundry`. |
| `stella-core/src/lib.rs` loses the `pub mod`/`pub use` lines | zero matches for `\b(search\|scoreboard\|extensions\|repair\|tool_foundry)\b` in that file |
| Acceptance grep returns nothing | `git grep -E 'stella_core::(search\|scoreboard\|extensions\|repair\|tool_foundry)\b'` is empty on the branch. I ran the same grep against `origin/main` as a control and it matches in six files, so the empty result is the change and not a broken pattern. |

**Box 4 is half-established and I did not tick it.** Its substance holds — the test suites travelled with their modules rather than being dropped: 12 markers in `search/budget.rs`, 41 in `extensions/plan.rs`, 31 in `tool_foundry/detect.rs`, 11 in `repair.rs`, 17 in `scoreboard/tests.rs`, against a drop of 109 `#[test]` in `stella-core`. What it asks for beyond that is "CI green", and there is no green run to cite: `fmt + clippy + test` was cancelled mid-run on `63c71e9` by the next push, and the runs on `5281db2` and `9d54b83` each restarted from scratch. Tick it once one run finishes green — the box is then earned, and `dod / dod` goes green with it.

**On the `docs/wire still matches the types` failure**, since I diagnosed it in parallel: `make doc-warnings-schema` was failing on two unresolved intra-doc links in `repair.rs`'s header. The cause is that an outer `///` on `pub mod repair;` merges with the file's own `//!` block, and rustdoc then resolves the merged block in `lib.rs`'s scope where neither `REPAIR_ATTEMPT_CAP` nor `RepairHeadroom` exists — `stella-core`'s declaration carried no outer doc, which is why the move surfaced it. `5281db2` reaches the same conclusion and fixes it at the lib.rs end. I confirmed that fix rather than assuming it: with a wiped `target/doc` (the freshness trap AGENTS.md names), `RUSTDOCFLAGS="-D warnings" cargo doc -p stella-plugin --features stella-plugin/schema --no-deps --document-private-items` exits 0 on `5281db2`, and `check-prose` stays green. I had a second fix in hand that qualified the two links instead; it is redundant and I did not push it.

---

macanderson:
Pushed `d060b80` — merged `origin/main` in and repaired the two prose baselines.

**What was failing.** `gate steps no other workflow runs` was red on:

```
check-prose: FAIL -- prose got harder to read.
  AGENTS.md: grade 11.71 allowed, 11.74 found
```

`AGENTS.md` is not in this PR's diff. The branch's `make prose-update` run had written `AGENTS.md 11.77 -> 11.71` into `scripts/prose-grade-baseline.txt`, because `--update` writes `min(measured, ceiling)` for *every* scored file, not just the ones the branch touched — so ~90 unrelated entries were lowered to the value they had on this branch's base. `main`'s `AGENTS.md` measures 11.74, so the merge failed a ceiling this branch had written and nothing in its diff explained. It reproduces only against the merge with current `main`, which is why a local run on the branch was green.

This is the shared-cell hazard AGENTS.md names for `file-size-baseline.txt`.

**The repair.** The grade baseline is rebuilt from `main` with only the five moved entries carried to their new paths — the move rule the guard already states. Its diff against `main` is now exactly ten lines:

```
> crates/stella-cli/src/extensions/plan.rs 10.46      < crates/stella-core/src/extensions.rs 10.46
> crates/stella-cli/src/tool_foundry/detect.rs 10.44  < crates/stella-core/src/tool_foundry.rs 10.44
> crates/stella-learn/src/scoreboard.rs 8.24          < crates/stella-core/src/scoreboard.rs 8.24
> crates/stella-plugin/src/repair.rs 10.46            < crates/stella-core/src/repair.rs 10.46
> crates/stella-tools/src/search/budget.rs 8.99       < crates/stella-core/src/search.rs 8.99
```

`scripts/prose-density-baseline.txt` is back to `main` byte for byte — the five crates it had retightened pass on their own, and `stella-tools` is drift the guard reports as inherited and does not fail on. This PR now writes neither shared cell beyond the moves it makes.

**Verified locally** (nothing compiled — CI is the evidence for `lint`, `test` and `doc-warnings`): `check-prose.py` green on the merge with `origin/main`, plus `invariants`, `doc-links`, `adr-numbering`, `command-docs`, `brand-case`, `file-size`, `god-files`, `gate-parity`, `left-behind`, `line-citations`, `core-reachability` and `module-reachability`. `shellcheck` is not on this machine and did not run; no shell script is touched.

**DoD.** I checked the first three boxes of #5116 against the tree rather than taking them on trust: the reachability baseline is down to `goal` and `subagent` (#3911 and #6142's, not this issue's), `stella-core/src/lib.rs` names none of the five, `rg 'stella_core::(search|scoreboard|extensions|repair|tool_foundry)'` returns nothing, and all five new homes exist. The fourth box is "CI green" and stays unticked until CI actually is.

**Residue filed:** #6272 — `make prose-update` retightens unrelated entries, so the next branch to run it re-creates this. The fix is the `--update` raises / `--retighten` lowers split `check-file-size.sh` already has.

---

macanderson:
PR-management routine: `gate steps no other workflow runs` is red ([job 101521065804](https://github.com/macanderson/stella/actions/runs/34046012181/job/101521065804)). It is three separate failures wearing one red X, and only one of them is yours.

The run reported this and stopped:

```
check-prose: FAIL -- prose got harder to read.
  AGENTS.md: grade 11.71 allowed, 11.73 found
```

`AGENTS.md` is not in this branch's diff — `git diff --name-only origin/main...5281db2f | grep -c '^AGENTS.md$'` returns `0`.

## What is actually failing

I re-ran the guard against `cf47b1e2`, the exact merge commit CI built, with the base tree readable:

```
check-prose: drift inherited from the base tree, which this change did not cause and does not fail on:
  crates/stella-cli: 23.25 mean header lines, 23.22 allowed, 23.25 in the base tree
  AGENTS.md: grade 11.73, 11.71 allowed, 11.73 in the base tree

  crates/stella-plugin: 34.95 allowed, 35.00 mean header lines
```

| Failure | Yours? |
| --- | --- |
| `AGENTS.md` grade 11.73 > 11.71 | No — inherited from `main` (#6257 and #6255 grew that file after you branched) |
| `crates/stella-cli` density 23.25 > 23.22 | No — same, inherited |
| `crates/stella-plugin` density 35.00 > 34.95 | **Yes** — `repair.rs` moving in raises that crate's mean header length |

So the one thing to fix here is `crates/stella-plugin`: trim `repair.rs`'s module header on the way in, and that crate's mean drops back under 34.95. The header is long enough to carry the whole `stella-pipeline` deletion history, which `doc:prose-guidelines` puts in the pull request rather than the file.

## Why CI could not forgive the other two

`guard-self-tests.yml`'s `gate-steps` job checks out with a bare `actions/checkout` — `fetch-depth: 1`. On that checkout `resolve_base_commit` can read no base at all:

```
git rev-parse --verify --quiet HEAD^2            -> (empty)
git rev-parse --verify --quiet 'HEAD^1^{commit}' -> (empty)
git show HEAD^1:AGENTS.md                        -> object missing (shallow)
```

Every arm misses, it returns `""`, and `""` collapses `max(ceiling, base)` to the ceiling — the strict whole-tree check. At `fetch-depth: 2`, same commit, nothing else changed, the base resolves and the two inherited rows are forgiven, which is the output above. So #6254's fix is correct and is simply not reachable in the job where the check runs; the density allowance that predates it has never been reachable there either. Filed as **#6273**.

## One line in this diff you probably did not mean

This branch lowered `AGENTS.md`'s grade ceiling without touching the file:

```
origin/main:  AGENTS.md 11.77
5281db2f:     AGENTS.md 11.71
```

`check-prose.py --update` reclaims slack for *every* grade and count entry on every run — the density baseline got the `--update` / `--retighten` split from #4657, grades and counts did not. You ran it before #6257 landed, when `AGENTS.md` really did grade 11.71, and the global reclaim wrote that over `main`'s 11.77. Restoring that one line to `11.77` removes the failure CI actually stopped on, independently of #6273. Filed as **#6274**.

I did not push either change: the `stella-plugin` header is your new code's prose, and the baseline line is yours to restore or to leave for #6274.

`dod / dod` is red too, but downstream — one of #5116's remaining boxes is "the moved modules' own test suites pass in their new crates (CI green)", which cannot be ticked before CI is green.

---

Issue #3437, opened 2026-08-17
steering: the drop report is on the ledger for all three sources but on stderr for only one


macanderson:
## The problem

#3358 (PR #3435) put skills' and frames' evictions into `SteeringSet::dropped`, so the ledger is now complete across the record, skill and memory sources. The **human-facing** half is not: `report_record_drops` (`crates/stella-cli/src/memory/recall.rs`) filters `set.dropped` to `SteeringSource::Record` and prints only those, through the single producer `record_drop_message`. A skill that loses its seat every turn, or a frame the recall host's merge evicts, is now queryable but still says nothing to the person watching the run — which is the #2709 observability gap in its other half.

Two smaller silences noticed alongside, both in `recalled_frames_anchored` (same file) and both distinct from the host-merge drops #3358 wired:

- Frames filtered out by the citation-label rule (`project_recalled_frame` returns `None`) vanish uncounted. The provider spent tokens fetching them.
- Frames dropped by quarantine (`is_suppressed_local_frame`) also vanish. This one is deliberate suppression rather than a budget eviction, so it may want a different vocabulary than `DroppedCandidate` — decide, don't default.

## What done looks like

- One drop line per evicted candidate regardless of source, following `record_drop_message`'s shape rather than adding a second format (the constraint #3358 was held to). Each source needs its own remedy sentence: "raise `skills.max_skills`", "raise `context.retrieval.max_tokens`", not the record channel's precedence advice.
- The section-budget skill omission needs care: it appears in both `selected` and `dropped` by design (see the `GatheredSteering::source_drops` doc), so a naive loop over `set.dropped` would tell a user a skill was dropped that is in fact in their prompt. Either the line says what actually happened ("rendered without body/omitted from the section") or that class is excluded until #3243 Phase 4 collapses the budgets.
- A decision, recorded in code, on the two recall-side filters above: reported, or deliberately not, with the reason.

## Verify

Extend `crates/stella-cli/src/memory/tests/steering_selection.rs`, which already has the injectable-sink pattern (`recalled_frames_anchored(goal, anchors, |message| …)`) — assert the message text for a skill drop and a frame drop, as `memory/tests/record_channel.rs` does for records.

## Constraints

- The golden block test (`crates/stella-cli/src/memory/tests/golden_block.rs`) pins the volatile block byte-for-byte; a diagnostic must not reach the prompt.
[...]

PR #6268, opened 2026-09-06
feat(stella-core): give the steering plane a plugin producer

### LINKED ISSUE #6112: Plugins still never reach the steering plane, so nothing costs or reports what a plugin injects

macanderson:
## What and why

`SteeringSource::Plugin` was an arm of the steering plane with a rank, a tag and
no producer. An installed plugin's `before_turn` text goes straight in front of
the model, so the one source that comes from outside this repository was the one
source the shared budget could not see: nothing priced it, nothing ranked it,
and `drop_message` returned `None` for it, so even a candidate that arrived drew
no report.

This is the join.

- **`crates/stella-core/src/steering/plugins.rs`** (new, pure — invariant 2).
  Mirrors `steering/tools.rs`: `PluginContribution { plugin, stage, est_tokens }`,
  `plugin_candidates`, and `contribute(contributions, budget)`, which packs
  through the one shared `pack_to_budget`. Beside them `ContextAllowance` reads
  what a turn has left off `steering::ledger::SteeringLedger` and charges what
  the text took — the same cell records, skills, frames and the tool array
  already meet in.
- **`crates/stella-runtime/src/wrapper/contribution.rs`** (new sibling). Measures
  each admitted `before_turn` answer over the exact messages it would inject,
  then fits the round into what the turn can still afford.
  `dispatch::open_round` calls it once per round, over the round as a whole, so
  a first stage cannot spend everything and hide that the last was cut.
- **`TurnPrelude::steering`** carries the plane's record, and
  `WrapperDispatch::with_context_allowance` carries the ceiling — beside
  `with_host_max_holds`, the same shape.
- **`crates/stella-cli`** wires it. `Settings::plugin_context_budget` reads the
  allowance `context.steering.max_tokens` already declares, so there is no new
  settings key and no second budget. The three doors that bind a wrapper set it
  (`agent::goal` twice, `fleet_cmd`), the three turn drivers report the drops
  through `memory::report_steering_drops`, and `drop_message`'s plugin arm names
  the plugin, the stage and the remedy.

Exemplar: `crates/stella-core/src/steering/tools.rs` and
`crates/stella-cli/src/tool_lean.rs` — the tool arm of the same plane, landed in
#6057. The producer/adapter split, the `advertise`-shaped packing entry point,
and the "a withheld candidate is a cost measure, never a capability change"
contract are copied from them deliberately.

The maintainer pointer asked for a producer over the active plugin set, one
candidate per declared `[[wrapper.stages]]` entry, with `est_tokens` from "the
stage's declared prompt contribution (the `before_turn` contribution size, or the
manifest's declared budget)".

There is no declared budget in the manifest today, and adding one would let a
plugin set its own price: a number declared too small wins budget it never paid
for, and the plane would be pricing a claim instead of a cost. The issue's own
item 1 asks for the honest number — "`est_tokens` measured over the exact text
that would be injected". So the producer sits where that text exists, at the
point the contribution arrives, and the active set still decides membership,
because a plugin that is not switched on is never dispatched.

The pointer also put `Plugin` above `Tool` in the precedence. `source_rank`
keeps it last, and the comment now argues it: a tool the model cannot see is one
it will not pick, so a plugin able to evict a built-in schema is the governance
failure the issue names one rung up. It is also inert in the shipped path — tool
schemas are packed by their own `advertise` call against what the block left, so
`Plugin` and `Tool` never meet in one sort.

## Witness mechanism

Each of these fails on `main` by construction, because the methods they name did
not exist and nothing could produce or cut a plugin candidate.

- `crates/stella-runtime/tests/wrapper_contribution_budget.rs` drives the shipped
  `WrapperDispatch` with an in-process plugin and a recording `TurnDriver`:
  - `a_granted_contribution_reaches_the_plane` — a granted stage's text becomes a
    `SteeringSource::Plugin` row named `stella-research/research`, priced at
    exactly what the prompt carries. Fails on `main`: `TurnPrelude::steering`
    did not exist.
  - `an_undeclared_stage_produces_no_candidate` — the `before_turn_stages`
    narrowing that stops the dispatch also stops the cost, checked against the
    same fixture with the narrowing taken off.
  - `a_contribution_over_the_allowance_is_withheld_and_named` — a block that
    already spent 990 of a 1000-token allowance leaves nothing, both stages are
    cut, nothing reaches the prompt, and nothing is charged. Fails on `main`:
    `WrapperDispatch::with_context_allowance` did not exist.
  - `a_plugin_that_declared_no_point_contributes_nothing` — the grant is the
    gate, one rung above the stage list.
- `steering/plugins.rs`'s own units cover the pack: what does not fit is dropped
  and named, the survivors keep arrival order, a wide budget cuts nothing, and
  the block's spend and the plugin's spend land in one ledger cell.
- `wrapper/contribution.rs`'s units cover the measuring: an answer with no
  message is no row, an affordable one is charged, a refused one is not.
- `settings/steering.rs` gains three units for `plugin_context_budget`,
  including the one that keeps `tools.lean` from deciding it.

Verification: CI on this branch. No workspace build was run on the maintainer's
laptop (CLAUDE.md, "CI builds and tests; this laptop does not"). Locally:
`cargo fmt --all` and `make guards-fast` are green, and `make line-citations`,
`make dead-code-allows` and `./scripts/check-file-size.sh` pass.

## Tests deleted

None.

## Definition of done

- [x] A production adapter emits `SteeringSource::Plugin` candidates from an
      installed plugin's granted contribution points.
- [x] A contribution from a point the manifest did not declare never becomes a
      candidate.
- [x] `drop_message`'s plugin arm names the plugin and the remedy instead of
      returning `None`.
- [x] `source_rank`'s comment for `Plugin` states the decided precedence and its
      reason.
- [x] A witness test that fails on the old code.

Closes #6112

## Remaining work filed

See the comment below for the issues this session opened.

## Summary by Sourcery

Integrate installed plugin context into the shared steering budget so contributions are measured, ranked, packed, and reported consistently with other steering sources.

New Features:
- Add plugin contributions as a first-class steering-plane source, producing one budgeted candidate per granted plugin stage.
- Measure admitted plugin context over the exact messages injected into the prompt and enforce the shared steering allowance across each turn.
- Expose plugin steering decisions and actionable drop reports naming the plugin and stage.],
- bug_fixes relevantes? irrelevant? no need? Wait schema requires bug_fixes array. Use []
- bug_fixes
- enhancements
- build
- ci
- deployment
- docs
- tests
- chores
- overview

Enhancements:
- Keep plugin contributions below workspace-owned steering sources and tool advertisements in the defined precedence order.
- Share plugin spending with the existing steering ledger and reuse the established steering packing and reporting behavior.

Tests:
- Add core, runtime, CLI settings, and end-to-end wrapper coverage for plugin contribution measurement, budget enforcement, stage gating, ordering, ledger charging, and drop reporting.

### DIFF
--- a/crates/stella-cli/src/agent/goal.rs
+++ b/crates/stella-cli/src/agent/goal.rs
@@ -188,7 +188,10 @@ pub(crate) async fn run_raw_one_shot(
                 .serving(|manifest| {
                     crate::wrapper_plugin::session_host(cfg, manifest, Arc::clone(&sub_agents))
                 })
-                .map_err(crate::failure::CliFailure::from)?,
+                .map_err(crate::failure::CliFailure::from)?
+                // What a plugin puts in front of the model spends the same
+                // allowance this turn's records, skills and frames spend.
+                .with_context_allowance(crate::plugin_steering::allowance(cfg)),
         ),
         None => None,
     };
@@ -621,7 +624,9 @@ pub async fn run_goal_cmd(
                         candidate.as_ref().map(|granted| &granted.grant),
                     )
                 })
-                .map_err(crate::failure::CliFailure::from)?,
+                .map_err(crate::failure::CliFailure::from)?
+                // The same allowance every other steering source spends.
+                .with_context_allowance(crate::plugin_steering::allowance(cfg)),
         ),
         None => None,
     };
--- a/crates/stella-cli/src/agent/goal/goal_wrapped.rs
+++ b/crates/stella-cli/src/agent/goal/goal_wrapped.rs
@@ -156,6 +156,9 @@ impl TurnDriver for GoalRoundDriver<'_, '_> {
         // `TamperWatch::pin_declared` for why re-observing would launder a
         // rewrite from an earlier round.
         self.watch.pin_declared(prelude.witness());
+        // Whatever the steering allowance refused, named where every other
+        // steering source names its cuts.
+        crate::plugin_steering::report_drops(&prelude);
         self.messages.extend(prelude.into_messages());
         // One observer per internal turn, exactly as
         // `crate::wrapper_plugin::RawTurnDriver` builds one per round: `tools`
--- a/crates/stella-cli/src/fleet_cmd.rs
+++ b/crates/stella-cli/src/fleet_cmd.rs
@@ -881,7 +881,7 @@ async fn run_task(
     // The second moment of binding (#3882): a plugin's `child_turn` is spent
     // by this attempt's own dispatcher, which needs the provider built above.
     let wrapped = match resolved_wrapper {
-        Some(resolved) => Some(resolved.serving(&invocation_root, sub_agents)?),
+        Some(resolved) => Some(resolved.serving(&invocation_root, sub_agents, &cfg)?),
         None => None,
     };
     let active_rules = rules::enforce_workspace_rules(
--- a/crates/stella-cli/src/fleet_cmd/wrapped.rs
+++ b/crates/stella-cli/src/fleet_cmd/wrapped.rs
@@ -393,6 +393,9 @@ impl TurnDriver for AttemptDriver<'_, '_> {
         // later round's declaration cannot launder an earlier round's rewrite —
         // see `TamperWatch::pin_declared`.
         self.watch.pin_declared(prelude.witness());
+        // Whatever the steering allowance refused, named where every other
+        // steering source names its cuts.
+        crate::plugin_steering::report_drops(&prelude);
         // Invariant 7: `into_messages` hands back user messages, appended
         // after the byte-stable system prefix this conversation opens with.
         self.messages.extend(prelude.into_messages());
@@ -471,7 +474,9 @@ impl ResolvedAttempt {
     /// isolated worktree's, which has none and would answer every recall with
     /// nothing. `child_turn` reaches this attempt's own dispatcher, so the
     /// child runs in the worker's tree with the worker's tool policy and is
-    /// paid for out of the worker's guard.
+    /// paid for out of the worker's guard. `cfg` is what the steering
+    /// allowance is read from, so a plugin's contribution spends the tokens
+    /// this attempt's records, skills and frames spend.
     ///
     /// # Errors
     ///
@@ -482,6 +487,7 @@ impl ResolvedAttempt {
         self,
         invocation_root: &std::path::Path,
         dispatcher: std::sync::Arc<crate::subagent::SessionSubAgents>,
+        cfg: &crate::config::Config,
     ) -> Result<AttemptWrapper, String> {
         // One host per member of the selection, built from that member's own
         // manifest: the child-turn plane reads the manifest's `[roles]` and
@@ -501,7 +507,9 @@ impl ResolvedAttempt {
             )
         })?;
         Ok(AttemptWrapper {
-            bound,
+            // What a plugin puts in front of the model spends the same
+            // allowance this attempt's records, skills and frames spend.
+            bound: bound.with_context_allowance(crate::plugin_steering::allowance(cfg)),
             candidate: self.candidate,
             task: self.task,
         })
--- a/crates/stella-cli/src/main.rs
+++ b/crates/stella-cli/src/main.rs
@@ -101,6 +101,9 @@ mod plugin_cmd;
 // Where those routes become dispatches: the session's hook plane, folded
 // once at config assembly (#4417, #3521).
 mod plugin_hooks;
+// The plugin arm of the steering plane. It sets what a plugin may put in
+// front of the model. It names what the allowance cut.
+mod plugin_steering;
 mod scoreboard_cmd;
 mod self_driving_cmd;
 // The `/profile` posture planner (fast · balanced · pro · ultra).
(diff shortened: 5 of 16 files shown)


macanderson:
PR-management routine: `fmt + clippy + test` is red ([job 101521440088](https://github.com/macanderson/stella/actions/runs/34046035670/job/101521440088)). One test, and it is this PR's own witness catching up with it rather than a defect in the change.

```
---- memory::tests::steering_selection::every_dropped_source_gets_a_line_with_its_own_remedy stdout ----
panicked at crates/stella-cli/src/memory/tests/steering_selection.rs:757:5:
assertion `left == right` failed: one summary for the memory drops, then one line
per remaining reportable source
  left: 6
 right: 5

test result: FAILED. 2858 passed; 1 failed; 0 ignored
```

The sixth line is exactly the one this PR exists to produce:

> a plugin's contribution did not fit what this turn's records, skills and frames left of the steering allowance, and was not put in front of the model: `stella-research` — the plugin still ran; raise `context.steering.max_tokens`, or take the plugin out of `active_plugins`

Everything else on `dff1285` is green — `gate steps`, `guard self-tests`, `docs guards`, `file size ratchet`, MSRV, Windows, macOS, `docs/wire`, `cargo deny`.

## Three things in that test, not one

The fixture already carries the plugin candidate, and its comment states the behaviour this PR removes:

```rust
// The plugin arm still has none, and its silence is the absence
// of a producer rather than a withheld report.
dropped(SteeringSource::Plugin, "stella-research"),
```

So bumping `5` to `6` turns the suite green, but leaves two things behind:

1. **That comment is now false** and is the kind review here treats as a bug — the plugin arm has a producer as of this PR, and its silence was the thing being fixed.
2. **The count is the only thing asserting the new line.** The test's own header says every dropped source reaches the person watching the run *with the remedy its own channel answers to*, and it backs that up for memory, records, both skill classes and tools with positive assertions on the text. The plugin line would be the one source covered by nothing but an arity check — and an arity check passes just as well if the producer emits the wrong channel's remedy.

Worth a `lines` assertion in the same shape as the tool one: the handle `stella-research`, that the plugin still ran, and `context.steering.max_tokens` / `active_plugins` as the remedy. That also makes this test a witness for *this* PR rather than only for #3437.

I did not push the change — which assertion the plugin line should carry is a statement about the producer you are adding, so it belongs in your hands rather than in a count bump from me.

`dod / dod` is queued again on the current head; it was cancelled-then-requeued rather than failed, so nothing to act on there yet.

---

Issue #6287, opened 2026-09-06
Generated tool pages still name the old home of the tool table


macanderson:
Parent: #5119 (epic #5113 — finish the re-layering)

Noticed while landing the `stella-tool-facts` extraction (#5119). The tool
table moved from `crates/stella-tools/src/catalog.rs` to
`crates/stella-tool-facts/src/catalog.rs`. Two places still name the old
path, and both are generated output plus the generator that writes it.

## Where

- `crates/stella-cli/src/tool_docs.rs` — five string literals name
  `crates/stella-tools/src/catalog.rs`, in the header the generator stamps
  on every page and in two prose paragraphs.
- `docs/tools/*.toml` and `docs/tools/README.md` — the stamped copies of
  those literals, 19 pages, two lines each.

Nothing is broken: the generator and its output agree, so `make tool-docs`
passes. The path they agree on is the one the file used to have.

## Why it was not fixed in #5119

Changing the literal means regenerating every page with
`make tool-docs-update`, which compiles the workspace. That was out of
reach in that session (CLAUDE.md, "CI builds and tests; this laptop does
not"), and hand-editing the pages instead raised the reading grade of all
19 by 0.01, which the prose ratchet refuses. Regenerating is the right
move, and it needs one run of the generator.

## Verify

- `rg 'crates/stella-tools/src/catalog\.rs' crates/ docs/` returns only
  `crates/stella-tool-facts/src/catalog.rs`'s own `git log` recipe, which
  lists historical paths on purpose, and
  `crates/stella-tools/tests/fixtures/name_rung_corpus.jsonl`, a frozen
  sample that must not be edited by hand.
- `make tool-docs` is green.
- `python3 scripts/check-prose.py --absolute` is green.

## Definition of done

- [ ] `crates/stella-cli/src/tool_docs.rs`'s five literals name
      `crates/stella-tool-facts/src/catalog.rs` (the table's current home),
      not the retired `crates/stella-tools/src/catalog.rs` path.
- [ ] `docs/tools/*.toml` and `docs/tools/README.md` are regenerated with
      `make tool-docs-update` and the regenerated files are committed as
      part of the same change (no hand-editing the generated pages).
- [ ] Witness check, run before and after the fix: `rg
      'crates/stella-tools/src/catalog\.rs' crates/ docs/` — before the fix
      it matches `tool_docs.rs` plus all 19 generated pages; after the fix
      it matches only the two allowed exceptions named in Verify above.
[...]

Issue #6286, opened 2026-09-06
Put stella-graph behind a graph feature in stella-tools, so a small build can drop the code index


macanderson:
Sibling of #5119, which severed `stella-tui` → `stella-tools`. This is step 4
of that issue: put `stella-graph` behind a cargo feature in `stella-tools` so
an embedding host, a plugin-only build, or a CI cold compile can drop nine
tree-sitter grammar builds and a bundled SQLite without forking.

It was split out because it is not a manifest edit, and the reasons are below
so the next session does not have to rediscover them.

## What is entangled today

`rg -l 'stella_graph' crates/stella-tools/src/` names the whole `search`
subsystem plus `graph_fact`:

- `src/search/{engine,enrich,scan,semantic,codegraph,cache,backfill,names}.rs`
- `src/search/readiness.rs` (the `measure` half; the policy left for
  `stella-tool-facts` in #5119)
- `src/graph_fact.rs` and `src/graph_fact/tests.rs`

The two seams outside `search/` are small:

- `crates/stella-tools/src/registry.rs` registers `crate::search::Search`.
- `crates/stella-tools/src/write.rs` and `src/delete.rs` construct
  `crate::graph_fact::Codegraph`, the one implementation of the
  `WorkspaceGraph` port. The `GraphFact` type, `attach` and `from_output`
  need no graph at all.

## The two hard parts

**1. The index-free rung is not index-free.** The pointer on #5119 asked for
`cfg` on the code-graph rung of `search` plus a fallback that reports the rung
is absent. The file-scan rung — the one meant to answer with no index — reads
`stella_graph::DENY_DIRS`, `stella_graph::workspace_ignore::WorkspaceIgnore`
and `stella_graph::admitted::{Admits, admission}` (`src/search/scan.rs`). So a
`search` that still answers with `--no-default-features` needs that ignore and
admission data moved down to a leaf first. That is its own extraction, in the
shape #5119 used: a near-leaf both crates take, with the old paths re-exported.

**2. Registering nothing makes the catalog lie.** The other option is to leave
`search` unregistered under `--no-default-features`. The catalog would then
advertise a tool that is not there. The honest fix is the seam the table
already declares: `Availability` in
`crates/stella-tool-facts/src/catalog.rs` has one arm today, and its doc says
the type survives as "the declared seam a conditionally registered tool
re-enters through". Adding an arm changes what `stella-cli`'s
`src/tool_docs.rs` renders, so every page under `docs/tools/` has to be
[...]

Issue #5119, opened 2026-08-26
re-layering: the terminal renderer still links the tool registry — sever stella-tui → stella-tools via a leaf, feature-gate stella-tools → stella-graph


macanderson:
Child of #5113. Verified on main today: `crates/stella-tui/Cargo.toml` still
declares `stella-tools = { path = "../stella-tools" }`, so the renderer
transitively carries whatever the tool registry carries (tree-sitter grammar
builds and bundled SQLite via `stella-graph`, reqwest/scraper) — while the
same Cargo.toml is *full* of evidence that the right pattern is already house
style: `stella-diag`, `stella-diff`, `stella-transcript`, `stella-tui-theme`
are each annotated "a leaf, so no cycle / costs every consumer nothing".

## What to do

1. **Enumerate the real edge first** (the v0.9.40 sweep found:
   `catalog::{get, groups, group_for, names_in_group}`,
   `policy::{ToolPolicy::allows, WILDCARD}`,
   `subprocess_env::scrub_sensitive_env` — re-run
   `rg 'stella_tools::' crates/stella-tui/src/` for today's truth).
2. Relocate that set into a dependency-free leaf both crates consume — either
   an existing leaf if one fits semantically, or a new `stella-tool-facts`
   (name to taste): the tool *catalog* (names, groups, descriptions), the
   *policy* type, and the env scrubber are all data + pure predicates. The
   registry (execution, MCP composition, HTTP) stays in `stella-tools`.
3. Drop the `stella-tools` dependency from `stella-tui`.
4. Behind the same PR or a sibling: make `stella-tools → stella-graph` a
   cargo feature (`graph`, default-on for the shipped CLI) so an embedding
   host or a plugin-only build can drop nine C grammar builds without a
   fork. Composition point already exists: the MCP wrapping pattern
   (`McpToolSet::connect(...).wrapping(native)`) shows executors composing
   without a crate edge.

## Why it matters beyond hygiene

This sets the floor for every "small Stella" story at once: embedding hosts
(Mode B), the plugin-only build, CI cold-compile time, and the release
tarball. The renderer linking a code-graph indexer is pure accident of
history — the leaf annotations in this very Cargo.toml are nine months of
evidence the team already knows the correct shape.

## Witness

`cargo tree -p stella-tui` no longer contains tree-sitter or rusqlite;
`cargo tree -p stella-tools --no-default-features` likewise; the TUI golden
[...]

PR #6288, opened 2026-09-06
refactor(stella-tool-facts): let the terminal read tool facts without linking the executor

macanderson:
## What

`crates/stella-tui/Cargo.toml` declared `stella-tools`. The renderer therefore
linked the whole tool executor, and through it the code-graph index the
`search` tool ranks over: a bundled SQLite and nine tree-sitter grammar
builds. A screen paid for a code indexer it never calls.

What the deck actually read out of that crate is data and pure predicates.
`rg 'stella_tools::' crates/stella-tui/src/` on today's `main` finds six
things:

| What the deck reads | New home |
| --- | --- |
| `catalog::{get, group_for, groups, names_in_group, label_for}` | `stella_tool_facts::catalog` |
| `policy::WILDCARD` (and the precedence `ToolPolicy::allows` states) | `stella_tool_facts::policy` |
| `subprocess_env::scrub_sensitive_env` | `stella_tool_facts::subprocess_env` |
| `search::readiness::IndexReadiness` | `stella_tool_facts::readiness` |
| `registry::approval::{ApprovalRequest, ApprovalResponse, ApprovalSubject}` | `stella_protocol::approval` |
| `tasks::START` | `stella_tool_facts::catalog::TASK_START` |

`stella-tool-facts` is a new near-leaf crate. Its only workspace dependency is
`stella-protocol`, which is the contract `crates/stella-tui/tests/dependency_boundary.rs`
asserts. `stella-tools` re-exports every moved item at the path it had, so no
call site outside the renderer changed.

The three approval types went to `stella-protocol` rather than the new crate.
They are serde types crossing four crate boundaries, which is what that crate
is for (AGENTS.md #4), and `ApprovalSubject` already lived one layer down in
`stella-core`. `stella_core::hooks::decision` and
`stella_tools::registry::approval` both re-export them, so those paths still
work. The exemplar is `stella_protocol::question` next door: the same shape
for the `ask_question` tool, its broker and every surface that draws the card.

`ToolPolicy::parse_spec` gained a named error (`ToolSpecError`) on the way
across. A new crate cannot carry a `Result<_, String>` on a public signature —
`scripts/typed-errors-baseline.txt` refuses to add a crate — and the variants
are the right shape anyway. The sentences are the ones the parser already
printed, so `stella-cli`'s `--tools` message is byte-identical. The baseline
goes from 4 to 3 for `stella-tools`.

`subprocess_env::test_support` moved with its module and now rides a
`test-support` cargo feature rather than `cfg(test)`, which does not cross a
crate boundary. `stella-tools` turns it on as a dev-dependency, so its own
spawn-hygiene tests reach the fixtures and no shipped build compiles them.

Exemplar for the extraction itself: `stella-ansi` and `stella-diff`, both of
which left `stella-tui` for the same reason and re-export from their old home.
Exemplar for the witness: `crates/stella-runtime/tests/no_policy_crate_edge.rs`.

## Witness mechanism

`crates/stella-tui/tests/dependency_boundary.rs`:

- `the_renderer_declares_no_edge_to_the_tool_executor` parses
  `crates/stella-tui/Cargo.toml` and fails on a `stella-tools` line under
  `[dependencies]`. On `main` that line is present, so the test fails by
  construction; here it passes. It guards its own parser first, so a reformat
  that made the scan return nothing fails rather than passing.
- `the_facts_crate_takes_one_workspace_edge` reads the new crate's manifest and
  asserts its `[dependencies]` name `stella-protocol` and nothing else. Without
  it, the edge could come back one hop down and the first test would still be
  green.
- `every_forbidden_crate_still_exists` keeps a row from rotting into prose when
  the crate it names is gone.

`[dev-dependencies]` is out of scope on purpose: `tests/hello_panel_demo.rs`
drives a real plugin process through `stella-runtime`, which does take
`stella-tools`. That edge builds tests and never the shipped renderer, and the
manifest says so. So the honest command for the tree is
`cargo tree -p stella-tui -e normal`, not a bare `cargo tree`, which walks dev
edges too.

`crates/stella-protocol/src/approval.rs` adds the round-trip test AGENTS.md #4
asks of a new type there, over both `ApprovalSubject` arms and both answers.
`crates/stella-tool-facts/src/catalog.rs` adds
`the_plan_gate_name_is_a_catalog_row`, so the constant the plan gate keys on
cannot drift from the table.

## Scope: this does not do step 4

#5119 also asks for `stella-graph` behind a `graph` feature. That is
[#6286](https://github.com/macanderson/stella/issues/6286), with the reasons
written out, because it is not a manifest edit:

- The file-scan rung — the one meant to answer with no index — reads
  `stella_graph::{DENY_DIRS, workspace_ignore::WorkspaceIgnore, admitted}`, so
  a `search` that still answers without the graph needs that data moved to a
  leaf first.
- Registering nothing instead makes the catalog advertise a tool that is not
  there. The honest fix is the `Availability` seam, which changes what
  `stella-cli`'s doc generator renders and so needs every page under
  `docs/tools/` regenerated in the same change.

So this PR carries `Refs #5119` rather than `Closes`, and the second box of
that issue's checklist closes with #6286. A comment on #5119 records the split.

Refs #5119
Refs #5113

## Tests deleted

None. Three test modules moved with the files that hold them
(`catalog.rs`, `policy.rs`, `subprocess_env.rs`), and every test name is
unchanged.

## Residue filed

- [#6286](https://github.com/macanderson/stella/issues/6286) — step 4, the
  `graph` feature.
- [#6287](https://github.com/macanderson/stella/issues/6287) — `docs/tools/`
  and its generator still name the table's old path. Repointing them by hand
  raised the reading grade of all 19 generated pages, so it wants one run of
  `make tool-docs-update`.

## Guards run locally

`make guards-fast` and `python3 scripts/check-prose.py --absolute` are both
green. Nothing was compiled here; CI is the evidence for the build and the
tests.

`scripts/prose-baseline.txt` and `scripts/prose-grade-baseline.txt` carry only
the three moved files' entries to their new paths. `make prose-update` also
retightens every unrelated file in the tree, which is a shared cell every open
branch writes; that belongs in its own change.

## Summary by Sourcery

Extract shared tool facts and cross-crate approval types into lightweight boundary crates so the terminal can render tool-related state without linking the tool executor or code-graph index.

New Features:
- Add the `stella-tool-facts` crate for sharing tool metadata, policies, subprocess environment rules, and index-readiness state without linking the executor.
- Move approval request and response types into `stella-protocol` for use across UI and runtime boundaries.

Bug Fixes:
- Preserve typed errors and existing CLI error messages when parsing tool switch specifications.

Enhancements:
- Remove the renderer's dependency on `stella-tools` and enforce the dependency boundary so it links only lightweight tool facts.
- Retain compatibility by re-exporting moved APIs from their former `stella-tools` and `stella-core` paths.
- Update tool documentation references, workspace guidance, and generated-documentation workflow paths for the new catalog location.

CI:
- Extend the tool documentation workflow to run when the extracted tool catalog changes.

Documentation:
- Document the new `stella-tool-facts` crate boundary and update tool catalog references across project documentation.

Tests:
- Add dependency-boundary tests ensuring `stella-tui` does not depend on the executor and `stella-tool-facts` has only its permitted workspace dependency.
- Add serialization round-trip coverage for approval types and consistency coverage for the plan-start catalog entry.

Chores:
- Move subprocess test fixtures behind a cargo feature so they remain available to dependent crate tests without entering shipped builds.


macanderson:
## Why this branch carries three `Cargo.lock` resync commits

Not three mistakes — one structural hazard firing three times in an hour, as `0.9.367 → 0.9.368 → 0.9.369` landed underneath it.

**The mechanism.** `stella-tool-facts` is a new workspace member, so its `Cargo.lock` block exists only on this branch. A release sync rewrites the `version` line of every member `main` knows about; it cannot reach the one it does not. Git merges the branch's added block and main's bumps with no conflict to report, and the merge names one member at a version the manifests no longer carry — so every `--locked` job refuses it. This is the shared-cell hazard AGENTS.md § "Gotchas" names for `Cargo.lock`.

It is invisible pre-merge from either side: `cargo metadata --locked` passes on this branch **and** on `main`. Only the merge, which is what CI builds, is wrong. That is why the first red run showed `msrv`, `macos-check` and `docker-serve` all failing on `--locked` while the branch itself was clean.

**Evidence, each time:** on the merge, `cargo metadata --format-version 1 --locked` exits 101; after regenerating, exits 0. `make guards-fast` (`check-lockfile-sync` and `cargo fmt --check` included) green. After the first resync, `ci`, `msrv`, `macos-check`, `docker-serve`, `windows-check`, `wire-schema`, `guard-self-tests`, `docs-guards` and the rest all passed on `2b046fd`.

**It will keep recurring** until this PR merges: any PR adding a workspace member is re-broken by every release sync, and the sync cadence here is minutes. Nothing in the branch fixes that — the remedy is a merge window, and that is the maintainer's call rather than another resync commit.

Head is now `dcc1761`, resolved against `main` at `0.9.369`.

One note on process: `87d66f7` and a local commit of mine were the same fix written concurrently — I discarded mine and kept the pushed one rather than merging two identical changes. Worth knowing that a second session is active on this branch.

---

Issue #6261, opened 2026-09-06
A plugin's stage runs but emits no stage event, so every surface hides it


macanderson:
A plugin can contribute a stage to a turn, and the turn's event stream never
says so. The stage runs, the plugin is dispatched at it, and every consumer that
reads `stage` events — the deck's HUD, `plain.rs`'s stage rule, the exported
transcript, the Observatory — shows the turn skipping from one host boundary to
the next with the contributed work invisible between them.

That is the exact failure `stella_protocol::StageName` was opened to fix. #3963
and #3964 made a contributed stage spellable on the wire and renderable on every
surface. Nothing on the run path emits one.

## Where it stops

`stella_runtime::wrapper::TurnPrelude::stages()` holds the composed stage list —
the host's own boundaries plus whatever the active plugins contributed, in band
order. Three drivers take a `TurnPrelude` and none of them emits a boundary for
its stages:

- `crates/stella-cli/src/wrapper_plugin.rs`'s `RawTurnDriver::run_turn`
- `crates/stella-cli/src/agent/goal/goal_wrapped.rs`'s round driver
- `crates/stella-cli/src/fleet_cmd/wrapped.rs`'s attempt driver

The only stage a raw run emits is `Stage(Complete)`, from
`stella_core::EventSender::pairing_stage_complete`.

## How it was found

`crates/stella-cli/tests/roleless_core_acceptance_cli.rs` (#6144) runs
`doc:roleless-core` §9's acceptance script. It has to read the composed stage
list from the `before_turn` requests the host sent the plugin, because the event
stream cannot answer the question. That works, and it is a weaker instrument
than the one the wire types were opened for.

## The fix

Emit `AgentEvent::Stage` for each stage of the resolved program, at the boundary
it is dispatched at, with the scope the host's own boundaries use. Decide one
thing first: whether the boundary belongs to the driver (once per round, from
the prelude) or to `WrapperDispatch::open_round` (per stage, where the
dispatcher already walks them). The dispatcher walks stages and members and
knows which stages actually ran this round, which the prelude's list does not
say on its own once a condition is false.

Watch the consumer ledger: `AgentEvent::Stage`'s row in
`crates/stella-protocol/src/event/consumers.rs` already declares its consumers,
so this adds a producer rather than a variant.

## Definition of done

- [ ] A turn wrapped by a plugin that declares a contributed stage emits an
      `AgentEvent::Stage` naming that stage, in the order the program runs it.
[...]

Issue #6174, opened 2026-09-06
Reachability guard says an evicted module is one the engine now reaches


macanderson:
`scripts/check-core-reachability.py --update` prints the same line for two
different facts, and one of them is false.

## What happens

Move a module out of `crates/stella-core/src/` and run the update:

```
$ python3 ./scripts/check-core-reachability.py --update
check-core-reachability: retired event_stream — the engine reaches it now
check-core-reachability: retired mcp_usage — the engine reaches it now
```

Neither module exists in `stella-core` any more. The engine does not reach
them; there is nothing left to reach. The plain (non-update) run says the same
thing, under the heading `baseline is STALE`:

```
The engine now reaches these, so their baseline entries are dead:
```

## Why it is wrong

`main()` computes `reachable_now = baseline - set(unreached)`. A baseline name
drops out of `unreached` for two reasons: the engine started naming it, or the
file is gone. The message states the first as fact in both cases. Since the
whole point of #5113 is that these modules leave the crate, the false reading
is the common one — every eviction PR sees it.

## The fix

Tell the two apart. `module_file(src, name)` already answers "does this module
still exist in the crate", so the reporting can split on it: a name whose file
is gone reads `evicted — no longer in stella-core`, one whose file is still
there reads `the engine reaches it now`. Both are still retired from the
baseline; only the sentence changes.

The reporting lives in `main()` in `scripts/check-core-reachability.py`, in the
`--update` branch and in the `reachable_now` branch below it.

## Definition of done

- [ ] `--update` names an evicted module as evicted, not as reached.
- [ ] The plain run's `baseline is STALE` block does the same.
- [ ] A module still present in `crates/stella-core/src/` that the engine has
      started naming still reads "the engine reaches it now".
- [ ] `scripts/test-core-reachability.sh` covers both directions, so the two
      messages cannot be swapped again without a red run.
- [ ] `make core-reachability` and `make guards-fast` stay green.

Refs #5113

macanderson:
Triaged: no duplicate. DoD was already complete (fix, both-direction test, gate green). Labeled `bug`, `P2`, `area:core`, `use-model:cheap`, `pain:accuracy`, assigned to @macanderson, parented under epic #5113. Ready for work.

---
_Generated by [Claude Code](https://claude.ai/code)_

macanderson:
Folded into batch #6298 (2026-09-06 audit): fixed together by crate.

Issue #5979, opened 2026-09-05
The seat field's doc comment still shows a bare role name, which is now half a key


macanderson:
## Problem

`SubAgentSpec::seat`'s doc comment in `crates/stella-core/src/subagent.rs`
describes the field's value with three bare examples — `"planner"`,
`"reviewer"`, `"second-opinion"`. Since the seat key became
`<plugin-id>/<role>` on every host dispatch path (#5078), those are half a key,
not a key. A reader following that comment writes `[seats]` entries that cannot
be found, which is the exact defect #5078 fixed.

The rest of the comment is still correct and should stay: the field is opaque
to the engine, nothing in `stella-core` may branch on its contents, and `None`
means the requester named no seat.

## Files

- `crates/stella-core/src/subagent.rs` — `SubAgentSpec::seat`'s doc comment.
- `crates/stella-plugin/src/seat.rs` — `seat_key`, the one place the key is
  built, and the thing the comment should point at for the shape.
- `crates/stella-core/src/goal.rs` — the `seat: None` beside it carries a
  comment about what this line becomes when the goal loop is a plugin; check it
  reads correctly against the qualified shape.

## Constraints

- **`stella-core` must not depend on `stella-plugin`** (AGENTS.md #1, and the
  dependency direction `stella-plugin`'s own README states). The comment
  describes the shape and names the builder in prose; it does not link to it as
  an intra-doc reference and it does not add an edge.
- The examples should show what a `delegate` child carries too. That path sets
  `seat: None`, so the comment must not imply every seat is plugin-qualified —
  only that the plugin dispatch paths qualify theirs.
- Prose gate: `crates/stella-core/src/subagent.rs` sits at reading grade 9.78
  in `scripts/prose-grade-baseline.txt` and the ratchet is down-only.

## Verify

Read the comment and write a `[seats]` line from it alone. Today the line it
suggests is one no lookup can find.

## Definition of done

- [ ] `SubAgentSpec::seat`'s doc comment shows the qualified shape and says the
      host writes the plugin half.
- [ ] It still says the field is opaque to the engine and that nothing in
      `stella-core` may branch on it.
- [ ] `stella-core`'s dependency list is unchanged.
- [ ] `make prose` stays green.

Refs #5078.

macanderson:
Folded into batch #6298 (2026-09-06 audit): fixed together by crate.

Issue #5799, opened 2026-09-03
The same self-attestation rule is written twice, in the guard gate and the probe gate


macanderson:
## Problem

Two functions in `crates/stella-core/src/records/` now ask nearly the same question, and each answers it in its own code:

- `bridge.rs`'s `self_attested(record, trust)` decides whether a record may approve its own blocking guard. It requires `Trust::User`, a `decree` basis, and a non-empty `verified_by`.
- `sweep.rs`'s `honored_probe(record, trust)` decides whether a record may have a command run or a host reached for it. Since #5329 it requires `Trust::User`, a `decree` basis, a non-empty `verified_by`, and a stamped origin that is not `imported` or `inferred`.

The shared part is one rule: *a record's own fields cannot establish the authority they claim, so the loader-stamped tier decides, and a decree only counts with a person's name on it.* It is written twice. The two are not identical today — the probe gate adds the origin condition, which the guard gate has no use for — so this is not a mechanical de-duplication. It is a question of whether the shared predicate should be named once and called twice, with each caller adding its own extra condition.

The hazard is the ordinary one for a rule with two copies: someone tightens one of them. If the guard gate later grows a condition the probe gate should share, or the other way round, nothing in the tree notices. Both are security decisions, which is what makes it worth naming rather than leaving.

## What "done" could look like

A `pub(crate) fn decreed_by_a_named_human_at(record, trust) -> bool` in one of the two modules (or in `records.rs` beside `Trust`), called by both, with each caller keeping only the condition that is genuinely its own. A test that both gates refuse the same base case, so a change to the shared half cannot pass while only one gate moves.

Reject it if the two rules are better read separately — that is a legitimate answer, and it should be written down in a comment on both sides rather than left implicit.

## Files

- `crates/stella-core/src/records/bridge.rs` — `self_attested`
- `crates/stella-core/src/records/sweep.rs` — `honored_probe`
- `crates/stella-core/src/records/trust.rs` — where a shared predicate would sit if it belongs with the type

## Definition of done

- [ ] The shared half of the rule is either named once and called twice, or a comment on each side says why the two stay separate.
- [ ] If it is shared, a test fails when only one gate's base case changes.
- [ ] Both gates' existing witness tests still pass unchanged: `a_repository_record_cannot_approve_its_own_blocking`, `the_same_record_in_the_users_own_directory_self_approves`, `no_gated_probe_is_honored_at_project_trust_ever`, `a_repository_record_never_arms_a_gated_probe`.
[...]

PR #6323, opened 2026-09-06
fix(stella-core): fifteen defects from the 2026-09-06 core audit

### LINKED ISSUE #6298: core: fix batch — 17 small defects (audit 2026-09-06)
Several small `area:core` defects from the 2026-09-06 backlog audit ride one PR — the fix-over-file rule: each is a contained fix inside a crate this PR already touches, so bundling them costs one review instead of seventeen.

## Members

- [x] #2795 — engine_markers: constant→table direction is review-enforced, not mechanical — add a source-scan parity for marker-shaped constants — `crates/stella-core/src/engine_markers.rs`: add an `include_str!`-based scan test over `driver.rs`/`driver/truncation.rs`/`receipts.rs` asserting the scanned set equals `ENGINE_MARKERS`.
  - **Landed in #6323.** scan test `the_table_holds_every_marker_shaped_constant` in `engine_markers.rs`, over the nine sources that own a table entry. Demonstrated red by adding `FAKE_STEER_PREFIX` to `driver/truncation.rs` — `driver/truncation.rs's FAKE_STEER_PREFIX opens a bracketed User-role marker that ENGINE_MARKERS does not list` — then removed.
- [x] #2792 — Two approval.requested emissions per approval: bus audit stamp + #2676 rich event double-bridge into PolicyDecision — `crates/stella-core/src/bus.rs::bridge_policy_plane`: dedupe the two `approval.requested` emissions by keying on the `tool` payload field (option b).
  - **Landed in #6323.** `bus/policy_bridge.rs` holds one ask open per gate, and a resolution closes it. Witness `one_ask_is_one_row_however_many_emissions_it_makes` plus two negative controls, green locally.
- [x] #5773 — No Windows toolchain compiles stella-core or stella-cli — `.github/workflows/windows-check.yml`: add `-p stella-core` to the clippy invocation (option 1) and update `AGENTS.md`'s eighth-workflow description to match.
  - **Landed in #6323.** decision recorded: option 1. `windows-check.yml` runs `cargo clippy -p stella-tools -p stella-runtime -p stella-core`, and AGENTS.md's eighth-workflow paragraph matches. The job runs on #6323 because the diff touches `crates/stella-core/**`.
- [x] #5749 — Three more paths that buy a model call are still unpriced by the spend gate — `crates/stella-core/tests/spend_gate.rs`: add pinned scenarios for the summarizer's head-dropping ladder, its starved retry, and the length-continuation ladder.
  - **Landed in #6323.** three scenarios in `tests/spend_gate.rs` — `the_summarizers_head_dropping_ladder_buys_one_call_per_attempt`, `the_summarizers_starved_retry_buys_one_more_call`, `the_length_continuation_ladder_buys_one_call_per_continuation`. 18 passed locally, and `doc:verification-gate` names them.
- [x] #5799 — The same self-attestation rule is written twice, in the guard gate and the probe gate — `crates/stella-core/src/records/{bridge.rs,sweep.rs,trust.rs}`: extract the shared predicate into one function called by both `self_attested` and `honored_probe`.
  - **Landed in #6323.** `decreed_by_a_named_human_at` in `records/trust.rs`, called by both gates. Witness `a_blank_verifier_refuses_at_both_self_attestation_gates` fails on the old tree, where the probe gate honoured a whitespace-only `verified_by` the guard gate refused.
- [x] #2210 — docs(stella-core): the Calibration doc comment still claims overhead is a stable ratio — `crates/stella-core/src/estimator.rs:186-189`: rewrite the doc comment to describe the additive-intercept model, citing the witness at `:922`. Comment-only, no test.
  - **Landed in #6323.** the `Calibration` doc comment describes the fitted line with its additive intercept and names `a_constant_overhead_does_not_halve_the_compaction_budget`. Comment-only, no test.
- [x] #5334 — Skills and appraisal test gaps: the claims the suite does not pin — `crates/stella-core/src/skills/tests.rs` and `skills/appraisal/tests.rs`: add the four named tests (multi-byte truncation, equal-score tie-break, `Contributed` origin in demotion-protection, `over_top_k` direct test).
  - **Landed in #6323.** four tests — `a_multi_byte_body_is_cut_at_a_character_boundary`, `skills_with_the_same_score_are_ordered_by_name`, `over_top_k_holds_exactly_what_max_skills_cut`, and `every_origin_is_listed` beside the demotion-protection test, which reads `SkillOrigin::ALL` now instead of naming three of five origins. Green locally.
- [x] #5512 — Make a ToolExecutor decorator that forgets to forward fail loudly — `crates/stella-tools/src/`: add a shared `LoadedExecutor` test helper plus `assert_forwards(decorator)`, wired into each existing decorator's test module (level 1 only).
  - **Landed in #6323.** `stella_tools::forwarding` holds `LoadedExecutor` and `assert_forwards`. Every decorator in `stella-tools` and `stella-cli` has a forwarding test wired to it, and the helper has its own anti-vacuity case. It found two live drops: `ClaimTap` and `CommitObserver` both swallowed `active_skill_slugs`. Both forward it now.
- [x] #6174 — Reachability guard says an evicted module is one the engine now reaches — `scripts/check-core-reachability.py::main`: split the `--update` and `baseline is STALE` messages on whether the module's file still exists, per `module_file(src, name)`.
  - **Landed in #6323.** `retired_because` splits on `module_file`. `scripts/test-core-reachability.sh` gains four cases covering both messages in both the `--update` and `STALE` paths: 15 passed, 0 failed.
- [x] #5979 — The seat field's doc comment still shows a bare role name, which is now half a key — `crates/stella-core/src/subagent.rs::SubAgentSpec::seat`: rewrite the doc comment to show the qualified `<plugin-id>/<role>` shape. Comment-only.
  - **Landed in #6323.** the `seat` doc comment shows `<plugin-id>/<role>` and names `seat_key` in prose, not as an intra-doc link. `stella-core`'s dependency list is unchanged.
- [x] #2303 — core: Frontmatter.duplicate_keys is write-only once the rules::metadata layer retires (#2287) — remove the dead field — `crates/stella-core/src/rules.rs:173,297-305`: delete the field and its population.
  - **Landed in #6323.** field and population deleted. The parser had moved to `crates/stella-protocol/src/frontmatter.rs`; `rg duplicate_keys` over `crates/` returns only unrelated Python symbols in a test fixture now.
- [x] #2481 — stella-engine: the facade omits the turn-boundary payload shapers — `crates/stella-engine/src/lib.rs:129`: re-export `stella_core::driver::lifecycle::{turn_started_payload, turn_outcome_payload}`, update the README and drop the stale caveat in `driver/lifecycle.rs`.
  - **Landed in #6323.** both shapers re-exported, plus `TurnLane`, which obligation 2 already owed for `OwnedTurnCapabilities::lane`. Witness `a_host_can_frame_its_own_turn_through_the_facade_alone` names only `stella_engine::` paths, so it does not compile on the old tree. README and `lifecycle.rs`'s stale caveat updated.
- [x] #5746 — The spend gate does not pin how many attempts a parked wait can buy in total — `crates/stella-core/src/driver/rate_limit.rs` / `retry.rs`: widen `MAX_PARKED_WAIT_MS`/backoff constants to `pub(crate)` (option 2) and pin the derived attempt count in `crates/stella-core/tests/spend_gate.rs`.
  - **Landed in #6323.** `a_deadline_less_parked_wait_buys_a_bounded_number_of_attempts` derives the pin from `plan_park` and `MAX_PARKED_WAIT_MS` rather than a literal. Reading the ceiling made `driver::rate_limit` a `pub` module and the constant `pub` — a visibility change, so `driver.rs` gains no line.
- [x] #5737 — Nested frontmatter cannot be traced to the key it was written under — `crates/stella-core/src/rules.rs::parse_frontmatter`: record the parent key alongside `nested_keys`; narrow `extensions.rs::agent_from_file`'s refusal to a nested `tools:`/`allowed-tools:` key.
  - **Landed in #6323.** `Frontmatter::nested_parents` records the parent key and `nests_a_toolbelt_key` narrows the refusal. Witness `a_bare_toolbelt_beside_an_unrelated_nesting_still_loads` fails on the old code. `command_from_file` gets the same refusal, which it never had. ADR 0025's Consequences updated.
- [x] #5716 — The two completion nudges are engine text the model cannot tell from a tool result — `crates/stella-core/src/driver/{confident_zero.rs,live_services.rs}` + `engine_markers.rs`: give both nudges a bracketed marker, fold into `ENGINE_MARKERS`, delete `ENGINE_NUDGE_PREFIXES`/`is_engine_nudge`, update the prompt parity suite.
  - **could not ride.** The change moves prompt bytes the model answers, and the issue's own verification asks for the prove-it gate's bench arm to be re-run before and after — a rig and real provider spend. `ENGINE_NUDGE_PREFIXES` and `is_engine_nudge` are untouched, and #2795's new scan covers `ENGINE_MARKERS` without depending on this decision.
- [x] #5712 — The loop refuses to work an issue over rule files that would never steer it — `crates/stella-cli/src/self_driving_cmd/work.rs::refuse_if_unsteered`: don't count `governance.toml` or archived/retracted records toward the steering-file total.
  - **Landed in #6323.** the count reads and parses the files. Witness `a_file_that_would_never_steer_does_not_stop_the_loop` fails on the old code, where both fixtures are one `.toml` and the count is 1. Moved to `work/steering.rs`; `work.rs` drops to 1351 lines.
- [x] #6261 — A plugin's stage runs but emits no stage event, so every surface hides it — `crates/stella-cli/src/wrapper_plugin.rs::RawTurnDriver::run_turn` (+ `goal_wrapped.rs`, `fleet_cmd/wrapped.rs`): emit `AgentEvent::Stage` per resolved stage at the boundary it dispatches, and drop the workaround note from `roleless_core_acceptance_cli.rs`.
  - **could not ride.** It opens with a design decision the issue states outright — whether the boundary belongs to the driver or to `WrapperDispatch::open_round` — then spans three `stella-cli` drivers, `stella-runtime`, the acceptance script and `doc:roleless-core`. Larger than one session, and a partial wiring would leave a producer on one driver and not the others.

## Could not ride

- #5851 — needs a human policy decision per record (re-author vs. drop to `soft` across 11 records), not a contained code fix.
- #5830 — extracting `stella-records`' dependency on `stella-core` needs a new crate (or two), a workspace-table/README update, and untangling `mining.rs`/`skills.rs` coupling — an architecture change, not a batch-sized fix.
- #5514 — property tests across `compaction.rs`, `budget.rs` and eviction, plus widening three existing properties — multi-module scope beyond a batch slot.
- #5487 — three candidate designs with a real user-cost tradeoff; needs a maintainer decision recorded before implementation.
- #3250 — needs a maintainer decision between centralizing the starved-retry (a behavior change for every call site) and a new guard script.
- #2835 — wiring an approval route across five pipeline call sites with per-site headless/interactive decisions and engine-lifetime handling.
- #6173 — labeled `blocked`: the module cannot move until #3911 evicts `goal` first.
- #5249 — needs a decision (wire a producer vs. retire the event) plus reconciliation with `MemoryLogged`'s per-turn count to avoid double-counting.
- #2316 — bundles three separate asks (a test rename+assertion fix, a module doc relabel, and pruning roughly half of a 406-line test file) — more than one fix.
- #5738 — threading a task id through skill-observation feeders in `stella-cli` (or finishing the typed-path migration) — cross-crate design work.

## Definition of done

- [x] Every member's fix is in this PR with a witness test where the member is a defect (a test that fails on the old code and passes on the new), or the member is named under "Could not ride" with the case that stopped it.
- [x] Guards green. `make guards-fast` passes locally, and the required checks on #6323 are the evidence for the compiling half.
- [x] No baseline entry raised (file-size, prose, dead-code-allows, core-reachability, typed-errors — whichever the touched files carry). This PR modifies no baseline file, and `check-prose` and `check-file-size` pass without one.

macanderson:
Fifteen members of the `area:core` fix batch, each a contained fix inside a
crate this PR already touches. Two members could not ride; they are named at
the bottom with the case that stopped them.

### #5737 — nested frontmatter could not be traced to the key it sat under

`parse_frontmatter` recorded the child names of a nested mapping and not the
parent, so `agent_from_file` could only ask "is anything nested". A bare
`tools:` beside a nested `description:` lost the agent, and nothing had
widened. `Frontmatter::nested_parents` records the parent, and
`nests_a_toolbelt_key` asks ADR 0025's actual question.

`command_from_file` gets the same refusal in the same change: a nested
`allowed-tools:` leaves that key empty, and an empty one is the spelling of
"no restriction" — the identical widening, one loader over, refused by nothing.

**Witness:** `a_bare_toolbelt_beside_an_unrelated_nesting_still_loads` fails on
the old code, where `!fm.nested_keys.is_empty()` is true and the agent is
refused. `a_command_with_a_nested_toolbelt_is_refused` fails on the old code,
where the command loads with every tool. Two parser tests pin the new field.
ADR 0025's Consequences section says the proxy is gone.

### #5799 — the self-attestation rule was written twice, and had drifted

`bridge.rs`'s `self_attested` and `sweep.rs`'s `honored_probe` each spelled the
shared half — `Trust::User` plus a decree with a non-empty `verified_by` — in
their own code. The guard gate trimmed the verifier name and the probe gate did
not. `trust.rs`'s `decreed_by_a_named_human_at` is that rule, called by both;
each caller keeps only its own extra condition (`origin = "user"` for the
guard, a non-`imported`/`inferred` origin for the probe), and both doc comments
say why the extras are not shared.

**Witness:** `a_blank_verifier_refuses_at_both_self_attestation_gates` fails on
the old code — `verified_by = "   "` was refused by the guard gate and
*honored* by the probe gate. That is the drift, stated as a test.

### #2792 — one approval question wrote two journal rows

`approval.requested` reaches the bus twice per ask: `emit_blocking` stamps it
when a chain ends in `RequireApproval`, and the approval flow emits a richer
one before it parks. `bridge_policy_plane` mapped both onto `PolicyDecision`,
so a count of asks read double.

Neither emission can be dropped outright — a chain can require approval with no
flow behind it, and a flow can park for a gate no chain evaluated. The bridge
holds one ask open per gate instead, and a resolution (`approval.granted`,
`.denied`, `.expired`) closes it.

The bridge moves to `bus/policy_bridge.rs` with its tests: `bus.rs` sits at its
grandfathered ceiling, and the split takes 55 lines off it.

**Witness:** `one_ask_is_one_row_however_many_emissions_it_makes` emits the
stamp and the flow's message and asserts one row. On the old bridge it is two.
`a_resolved_ask_lets_the_next_one_through` and `two_gates_are_two_asks` are the
negative controls.

### #5712 — the loop refused to work an issue over rule files that never steer it

`refuse_if_unsteered` counted every `*.toml` in `.stella/rules`. Two of the
things it counted are not steering: `governance.toml`, a reserved name the
record loader never parses, and a record whose `status` is `archived` or
`retracted`, which the registry already refuses to let steer. The operator was
told to set `STELLA_TRUST_PROJECT=1` so records could steer a loop no record
would steer either way. The count now reads and parses the files, runs no truth
probe, and creates no directory; an unreadable or unparseable file counts as
steering, which is the safe direction.

It moves to `self_driving_cmd/work/steering.rs`: `work.rs` was 59 lines from
the ceiling, and the split leaves it at 1351.

**Witness:** `a_file_that_would_never_steer_does_not_stop_the_loop` fails on the
old code, where both fixtures are one `.toml` file and the count is 1. Its
mixed-directory arm shows one live record still stops an untrusted loop, so the
first two assertions are about what was written and not about the fixture.

### #6174 — the reachability guard called an evicted module one the engine reaches

`scripts/check-core-reachability.py` printed `the engine reaches it now` for
every baseline entry that left the unreached set, including one whose file is
gone — the common case under the eviction epic, so every eviction PR read a
false sentence. `retired_because` splits on `module_file`, and the plain run's
`baseline is STALE` block splits the same way.

**Witness:** `scripts/test-core-reachability.sh` gains four cases — `--update`
and the `STALE` block, each against an evicted module and against one the
engine started naming. They fail on the old code by construction: the old
message was one string for both facts, so the "evicted" assertion has nothing
to match. 15 passed, 0 failed locally.

### #5512 — a decorator that forgets to forward failed nothing

Four `ToolExecutor` methods have a trait default that reads "this executor
mounts none of that" — right for a leaf, wrong for a wrapper. A decorator that
drops one compiles clean and answers the default, and each decorator that had a
forwarding test had hand-written its own.

`stella_tools::forwarding` is that test's shared half: `LoadedExecutor` answers
all four with a value of its own, and `assert_forwards` reads them back through
the decorator. Every decorator in `stella-tools` and `stella-cli` has one now.

**It found two live drops.** `ClaimTap` and `CommitObserver` both swallowed
`active_skill_slugs`, so a live skill's procedure text stopped surviving
overflow summarization on the deck's lead turn and on every shared-tree fleet
worker — the same defect #5489 fixed one decorator over. Both forward it now.

**Witness:** `a_decorator_that_forwards_nothing_is_caught` is the helper's own
anti-vacuity. `the_tap_forwards_what_a_decorator_must` and
`the_commit_observer_forwards_what_a_decorator_must` fail on the old code, on
the method each of those two dropped.

### #5749 — three call-buying paths the spend gate did not price

The summarizer's head-dropping ladder when its own request is refused for size,
its starved retry, and the length-continuation ladder. Each new scenario names
the module it prices and is pinned from both sides where it has a bound: the
head-dropping script holds the ladder's worth of rejections and then an answer,
so raising the bound makes the summarizer eat the answer and lowering it makes
the worker call meet a rejection. 18 passed locally.

### #5746 — how many attempts a parked wait can buy in total

`a_parked_rate_limit_buys_one_attempt_per_park` prices one park. Nothing priced
the total, which is the number that bounds the spend. The new pin reads
`plan_park` and `MAX_PARKED_WAIT_MS` rather than a literal, so a change to the
ceiling or the backoff moves the expected number with it.

Reading the ceiling from an integration test needed it public. `rate_limit`
becomes a `pub mod` and `MAX_PARKED_WAIT_MS` a `pub const` — a visibility
change, so `driver.rs` gains no line and stays inside its grandfathered
ceiling.

`doc:verification-gate` lists all four new scenarios.

### #5334 — four claims the skills suite made and did not check

- `a_multi_byte_body_is_cut_at_a_character_boundary` — the panic
  `floor_char_boundary` exists to prevent, over CJK, emoji and a 2-byte Latin
  glyph. The existing truncation tests are ASCII, where the walk-back never
  runs.
- `skills_with_the_same_score_are_ordered_by_name` — the `then_with` tie-break.
  `auto_created_skill_wins_the_tie_break` asserts strictly unequal scores, so it
  never reaches it. Asserted from both input orders.
- `over_top_k_holds_exactly_what_max_skills_cut` — the `split_off`, including
  that `selected` matches `select_skills` exactly and that a cut removing
  nothing reports nothing.
- `a_hand_authored_skill_is_never_auto_demoted` claimed "every origin that is
  not `AutoCreated`" and enumerated three of five, so `Contributed` — a
  plugin's skill — was safe in the code and pinned by nothing. It reads
  `SkillOrigin::ALL` and filters on `is_hand_authored` now, and
  `every_origin_is_listed` holds that table to the enum with an exhaustive
  `match`.

The three selection tests land in `skills/tests/selection.rs`:
`skills/tests.rs` was 15 lines from the ceiling.

### #2795 — the constant→table coupling was enforced by a doc comment

Table→prompt and table→classifier are enforced by tests; the third direction,
"a new marker constant joins `ENGINE_MARKERS`", was an instruction in the module
doc. `the_table_holds_every_marker_shaped_constant` scans the nine sources that
own an entry for `const NAME: &str = "[…"` where the name contains `MARKER` or
ends in `_PREFIX`, and fails by name on one the table does not list.
`REASONING_ONLY_PARTIAL` is excluded by the name rule (it is an assistant-role
stand-in); `FILE_CAP_MARKER` is excluded by name with its reason recorded.
`the_marker_scan_reads_the_shapes_it_claims_to` is the scan's own anti-vacuity.

**Witness, run:** adding `pub(crate) const FAKE_STEER_PREFIX: &str = "[fake
steer";` to `driver/truncation.rs` fails the test with
`driver/truncation.rs's FAKE_STEER_PREFIX opens a bracketed User-role marker
that ENGINE_MARKERS does not list`. Removed again; the tree carries no such
constant.

### #5773 — no Windows toolchain compiled `stella-core`

Decision recorded on the issue: **option 1**. `windows-check.yml` adds
`-p stella-core` to the clippy invocation and `crates/stella-core/**` to both
path filters. `stella-cli` stays out, as the crate most likely to need real
work before it compiles there. `stella-core` already compiled on that runner as
`stella-runtime`'s dependency, so naming it costs a lint pass rather than a
build. `AGENTS.md`'s eighth-workflow description matches the workflow.

**Witness:** this PR touches `crates/stella-core/**`, so the job runs here with
`stella-core` under `-D warnings` — the run is the evidence. The issue's
red-then-green demonstration is **not** included: proving it needs a deliberate
compile error pushed and then removed, and that edit-and-undo pair is what
`rebase-replay` rejects on the next rebase.

### #2481 — the `stella-engine` facade omitted the turn-boundary shapers

A host that drives `run_step` rather than `Engine::drive` owns its own turn
framing and must emit `agent.turn.started` / `agent.turn.completed` in the
engine's shape, or an extension watching the bus can tell a host-driven turn
from a local one. The facade did not export the two shapers, so such a host had
to depend on `stella-core` directly — the coupling the facade exists to remove.
`TurnLane` joins them: `turn_started_payload` takes one, and obligation 2
already owed it, since `OwnedTurnCapabilities::lane` is an `Option<TurnLane>` a
host fills and could not name.

`driver/lifecycle.rs`'s stale caveat is dropped, and `README.md` names the
choice between "call `drive`" and "drive `run_step` and frame it yourself".

**Witness:** `a_host_can_frame_its_own_turn_through_the_facade_alone` in
`crates/stella-engine/tests/embedding.rs`. Every path in it is a
`stella_engine::` one, so the file does not compile on the old tree — the
fail-on-old mechanism for a re-export.

### #2303 — `Frontmatter.duplicate_keys` was written on every parse, read by nothing

Its only consumer left with the `rules::metadata` layer. `rg duplicate_keys`
over `crates/` now returns only unrelated Python symbols in a test fixture.
Field and population deleted. (The issue's line numbers pointed at
`stella-core/src/rules.rs`; the parser has since moved to
`stella-protocol/src/frontmatter.rs`.)

### #2210 — the `Calibration` doc comment claimed overhead is a stable ratio

The comment described a multiplicative ratio model; the code has fitted a line
with an additive intercept since #1841, and the witness 730 lines below it
asserts the opposite of what the comment said. Comment-only, no test — the
existing `a_constant_overhead_does_not_halve_the_compaction_budget` is the
witness the comment now names.

### #5979 — `SubAgentSpec::seat`'s doc comment showed half a key

The examples were bare role names. Every host dispatch path qualifies a seat as
`<plugin-id>/<role>`, so a `[seats]` line written from that comment is one no
lookup can find. Comment-only. `stella-core`'s dependency list is unchanged —
the comment names `stella-plugin`'s `seat_key` in prose, not as an intra-doc
link.

## Could not ride

- **#5716 — give the two completion nudges bracketed markers.** The change
  moves prompt bytes the model answers, and the issue's own verification asks
  for the prove-it gate's bench arm to be re-run before and after. That is a rig
  and real provider spend, one of the three defer cases. Nothing here is
  half-done: `ENGINE_NUDGE_PREFIXES` and `is_engine_nudge` are untouched, and
  #2795's new scan covers `ENGINE_MARKERS` without depending on that decision.
- **#6261 — emit a stage event per resolved plugin stage.** It opens with a
  design decision the issue states outright ("whether the boundary belongs to
  the driver or to `WrapperDispatch::open_round`"), then spans three
  `stella-cli` drivers, `stella-runtime`, the acceptance script and
  `doc:roleless-core`. Larger than this session, and a partial wiring would
  leave a producer on one driver and not the others — worse than the gap.

## Tests deleted

Two renamed, none removed:

- `the_task_tap_forwards_active_skill_slugs` →
  `the_task_tap_forwards_what_a_decorator_must`
- `the_commit_observer_forwards_parallel_safe_names` keeps its name; a second
  test `the_commit_observer_forwards_what_a_decorator_must` joins it.

The `LoadedInner` fixture in `custom/tests.rs` is deleted, superseded by
`stella_tools::forwarding::LoadedExecutor`. It is a fixture, not a test.

Closes #6298

## Summary by Sourcery

Correct core audit defects across frontmatter loading, policy and trust handling, steering detection, executor forwarding, reachability reporting, and verification coverage.

New Features:
- Expose turn-boundary payload shapers and lane types through the stella-engine facade for hosts that drive turns themselves.

Bug Fixes:
- Prevent nested frontmatter under unrelated keys from causing false refusals while rejecting nested tool restrictions that could silently widen access.
- Unify self-attestation validation across record guard and probe gates, including rejecting blank verifiers.
- Collapse duplicate approval-request events into one journal row per open gate and resolution cycle.
- Ignore reserved, archived, and retracted rule files when deciding whether untrusted self-driving work lacks steering.
- Distinguish evicted modules from newly reachable modules in core reachability diagnostics.
- Preserve active skill information through CLI executor decorators and other forwarding layers.

Enhancements:
- Add shared forwarding assertions for ToolExecutor decorators and strengthen skill selection, marker-table, and parked-wait accounting coverage.
- Remove the unused frontmatter duplicate-key tracking and correct stale calibration and sub-agent seat documentation.

CI:
- Run stella-core clippy with warnings denied in the Windows workflow and trigger it for stella-core changes.

Documentation:
- Update ADR, verification-gate, engine facade, and workflow documentation to reflect the corrected behavior and coverage.

Tests:
- Add regression coverage for frontmatter parent tracking, approval deduplication, steering detection, reachability diagnostics, decorator forwarding, Unicode skill truncation, tie-breaking, top-k reporting, marker coupling, and spend limits.

Chores:
- Split policy bridging and steering checks into focused modules to keep core files within their size limits.

### DIFF
--- a/.github/workflows/windows-check.yml
+++ b/.github/workflows/windows-check.yml
@@ -14,12 +14,12 @@ name: windows-check
 # owns `rootfd`, `durable_write` and `exec`; `stella-runtime`'s two wrapper
 # transports arm that guard — on a real MSVC runner.
 #
-# `stella-core` joins them (#5773). It carries no `#[cfg(windows)]` arm, and by
-# invariant 2 it never will carry one that does I/O — but a Windows defect can
-# still land there as a `std::path` assumption, which is what #5320 was, and it
-# is cheap to keep green because it has no I/O to port. It already compiled
-# here as `stella-runtime`'s dependency; naming it is what puts it under
-# `-D warnings` instead of plain rustc warnings.
+# `stella-core` is here for a different reason. It carries no
+# `#[cfg(windows)]` arm, and by invariant 2 it never will carry one that does
+# I/O — but a Windows defect still lands there as a `std::path` assumption, and
+# a skill path separator was one. It has no I/O to port, so it is cheap to keep
+# green. It already compiled here as `stella-runtime`'s dependency; naming it
+# is what puts it under `-D warnings` instead of plain rustc warnings.
 #
 # `stella-cli` stays out. It is the crate most likely to hold a real platform
 # assumption, so widening to it is work with its own PR rather than a name
--- a/.github/workflows/windows-check.yml
+++ b/.github/workflows/windows-check.yml
@@ -10,16 +10,27 @@ name: windows-check
 # the Job Object half of `exec::GroupKillGuard` beside them; "it compiles" was
 # a claim nobody could check for any of them.
 #
-# So this compiles the two crates that carry a platform split — `stella-tools`
+# So this compiles the crates that carry a platform split — `stella-tools`
 # owns `rootfd`, `durable_write` and `exec`; `stella-runtime`'s two wrapper
 # transports arm that guard — on a real MSVC runner.
 #
-# Its reach is wider than those two names: `-D warnings` travels to every
+# `stella-core` joins them (#5773). It carries no `#[cfg(windows)]` arm, and by
+# invariant 2 it never will carry one that does I/O — but a Windows defect can
+# still land there as a `std::path` assumption, which is what #5320 was, and it
+# is cheap to keep green because it has no I/O to port. It already compiled
+# here as `stella-runtime`'s dependency; naming it is what puts it under
+# `-D warnings` instead of plain rustc warnings.
+#
+# `stella-cli` stays out. It is the crate most likely to hold a real platform
+# assumption, so widening to it is work with its own PR rather than a name
+# added to a line.
+#
+# Its reach is wider than the names on the line: `-D warnings` travels to every
 # workspace crate they pull in, which is how it found a `cfg(unix)`-only
 # parameter left unread in `stella-store`. That is the job working, not a
 # scope error — a platform arm nobody compiles is the same defect wherever it
-# sits, and these two crates are simply the entry point that reaches the ones
-# holding the splits.
+# sits, and the named crates are the entry point that reaches the ones holding
+# the splits.
 #
 # `cargo clippy`, and deliberately **not** `--all-targets`. The shipping code
 # is what carries the platform split, and it is what this can honestly claim
@@ -66,6 +77,7 @@ on:
     paths:
       - "crates/stella-tools/**"
       - "crates/stella-runtime/**"
+      - "crates/stella-core/**"
       - "Cargo.lock"
       - "rust-toolchain.toml"
       - ".github/workflows/windows-check.yml"
@@ -74,6 +86,7 @@ on:
     paths:
       - "crates/stella-tools/**"
       - "crates/stella-runtime/**"
+      - "crates/stella-core/**"
       - "Cargo.lock"
       - "rust-toolchain.toml"
       - ".github/workflows/windows-check.yml"
@@ -111,7 +124,7 @@ jobs:
 
       - name: clippy on the crates that carry a platform split
         if: ${{ !cancelled() && steps.rust.outcome == 'success' }}
-        run: cargo clippy -p stella-tools -p stella-runtime -- -D warnings
+        run: cargo clippy -p stella-tools -p stella-runtime -p stella-core -- -D warnings
 
       - name: the plugin socket's own tests, on Windows (#3497, #3550)
         if: ${{ !cancelled() && steps.rust.outcome == 'success' }}
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -416,12 +416,18 @@ looks at a `#[cfg(windows)]` arm: `ci.yml` runs on `ubuntu-latest` and
 non-unix body in the tree — `rootfd.rs`'s and `durable_write.rs`'s string
 resolvers, and the Job Object half of `exec::GroupKillGuard` (#3550) — was
 code no toolchain here or in CI ever parsed. It runs
-`cargo clippy -p stella-tools -p stella-runtime` on a Windows runner, on the
-paths that reach those two crates — the shipping code, deliberately not
+`cargo clippy -p stella-tools -p stella-runtime -p stella-core` on a Windows
+runner, on the paths that reach those three crates — the shipping code,
+deliberately not
 `--all-targets`, because several `#[cfg(test)]` bodies import
 `std::os::unix::fs::PermissionsExt` unconditionally and would fail the job on
 a fixture rather than on the platform split. Those fixtures are #3497's
-subject. It then **runs** two of them: `stella-runtime`'s `wrapper_socket` and
+subject. `stella-core` joined the list for #5773: it holds no platform arm,
+but a Windows defect lands there as a `std::path` assumption (#5320 was one),
+and it already compiled here as `stella-runtime`'s dependency, so naming it
+costs a lint pass rather than a build. `stella-cli` stays out, as the crate
+most likely to need real work before it compiles there. It then **runs** two
+of them: `stella-runtime`'s `wrapper_socket` and
 `wrapper_transport_limits`, which stopped being `/bin/sh` scripts when #3497
 gave the crate a portable in-tree plugin binary
 (`crates/stella-runtime/tests/fixtures/wrapper-plugin-fixture.rs`). That is the
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -422,11 +422,11 @@ deliberately not
 `--all-targets`, because several `#[cfg(test)]` bodies import
 `std::os::unix::fs::PermissionsExt` unconditionally and would fail the job on
 a fixture rather than on the platform split. Those fixtures are #3497's
-subject. `stella-core` joined the list for #5773: it holds no platform arm,
-but a Windows defect lands there as a `std::path` assumption (#5320 was one),
-and it already compiled here as `stella-runtime`'s dependency, so naming it
-costs a lint pass rather than a build. `stella-cli` stays out, as the crate
-most likely to need real work before it compiles there. It then **runs** two
+subject. `stella-core` is on the list because a Windows defect lands there as
+a `std::path` assumption — a skill path separator was one — and because it
+already compiled here as `stella-runtime`'s dependency, so naming it costs a
+lint pass rather than a build. `stella-cli` stays out, as the crate most
+likely to need real work before it compiles there. It then **runs** two
 of them: `stella-runtime`'s `wrapper_socket` and
 `wrapper_transport_limits`, which stopped being `/bin/sh` scripts when #3497
 gave the crate a portable in-tree plugin binary
--- a/crates/stella-cli/src/claims.rs
+++ b/crates/stella-cli/src/claims.rs
@@ -602,6 +602,13 @@ impl ToolExecutor for ClaimTap<'_> {
         self.inner.live_services()
     }
 
+    /// Forwarded: this tap claims files. It mounts no invocation plane of its
+    /// own. The empty default reads as "no skill is running here", which stops
+    /// a live skill's procedure text surviving summarization.
+    fn active_skill_slugs(&self) -> Vec<String> {
+        self.inner.active_skill_slugs()
+    }
+
     /// Forwarded: letting the empty default stand would silently serialize the
     /// inner executor's sibling spawns (see the port's contract). The spawn
     /// tool names no mutating path and no transient lane, so concurrent
@@ -630,6 +637,19 @@ mod tests {
     use stella_fleet::{GitError, GitOutput};
 
     use super::*;
+
+    /// Every method the port tells a decorator to forward, over this tap.
+    ///
+    /// The tap coordinates writes. It changes nothing about what the inner
+    /// executor mounts. `active_skill_slugs` was the one it dropped. The tap
+    /// sits above the invocation mount in the deck's lead turn, so a live
+    /// skill's text stopped surviving summarization there.
+    #[test]
+    fn the_tap_forwards_what_a_decorator_must() {
+        let loaded = stella_tools::forwarding::LoadedExecutor;
+        let tap = ClaimTap::new(&loaded, None, "holder");
+        stella_tools::forwarding::assert_forwards("ClaimTap", &tap);
+    }
 
     /// A recording fake: succeeds every call and remembers what ran.
     struct Passthrough(std::sync::Mutex<Vec<String>>);
--- a/crates/stella-cli/src/command_deck/task_tap.rs
+++ b/crates/stella-cli/src/command_deck/task_tap.rs
@@ -887,12 +887,17 @@ mod tests {
         );
     }
 
-    /// Same shape for the invocation plane (#2685): the tap sits above the
-    /// discovery mount, so swallowing the live-slug answer would stop active
-    /// skill bodies surviving summarization on every deck session.
+    /// Every method the port tells a decorator to forward, over this tap.
+    ///
+    /// The tap observes board writes; it changes nothing about what the inner
+    /// executor mounts. It sits above the discovery mount, so swallowing the
+    /// live-slug answer would stop active skill bodies surviving
+    /// summarization on every deck session. The checks live in
+    /// `stella_tools::forwarding::assert_forwards`, so a method added to the
+    /// port is added there once.
     #[test]
-    fn the_task_tap_forwards_active_skill_slugs() {
-        let inner = Claiming;
+    fn the_task_tap_forwards_what_a_decorator_must() {
+        let inner = stella_tools::forwarding::LoadedExecutor;
         let registry = ToolRegistry::new(std::path::PathBuf::from("."));
         let tap = TaskTap {
             inner: &inner,
@@ -901,10 +906,6 @@ mod tests {
             dispatched_by: None,
             plan_gate: None,
         };
-        assert_eq!(
-            tap.active_skill_slugs(),
-            vec!["deploy".to_string()],
-            "the tap must forward the inner executor's live invocations"
-        );
+        stella_tools::forwarding::assert_forwards("TaskTap", &tap);
     }
 }
--- a/crates/stella-cli/src/extensions/plan.rs
+++ b/crates/stella-cli/src/extensions/plan.rs
@@ -1068,10 +1068,10 @@ mod tests {
     /// A bad `description:` costs a label, since the first line of the body
     /// stands in. A bad `tools:` costs the whole registry. Refusing the file
     /// for the first would take an agent away over a key nobody reads.
-    /// The false stop the parent-key record removes. A bare `tools:` means
-    /// every tool, and nesting under `description:` widens nothing — but the
-    /// refusal used to read "the toolbelt is empty AND something is nested"
-    /// and take the agent away.
+    /// The false stop the parent key removes. A bare `tools:` means every
+    /// tool, and nesting under `description:` widens nothing, so the file
+    /// loads. Refusing on "the toolbelt is empty and something is nested"
+    /// took the agent away over a key nobody reads.
     #[test]
     fn a_bare_toolbelt_beside_an_unrelated_nesting_still_loads() {
         let raw = "---\nname: reviewer\ntools:\ndescription:\n  short: s\n  long: l\n---\nReview.";
--- a/crates/stella-cli/src/extensions/plan.rs
+++ b/crates/stella-cli/src/extensions/plan.rs
@@ -301,6 +301,19 @@ fn definition_from_file(
 pub fn command_from_file(path: &str, raw: &str) -> Result<CommandDef, ExtensionDiagnostic> {
     let (fm, name, description, body) = definition_from_file(path, raw, "COMMAND.md")?;
     let get = |key: &str| fm.data.get(key).map(String::as_str);
+    // The same widening `agent_from_file` refuses, one loader over: a nested
+    // `allowed-tools:` leaves the key empty, and an empty one is the spelling
+    // of "no restriction".
+    if get("allowed-tools")
+        .or_else(|| get("allowed_tools"))
+        .is_some_and(str::is_empty)
+        && nests_a_toolbelt_key(&fm)
+    {
+        return Err(ExtensionDiagnostic {
+            path: path.to_string(),
+            problem: ExtensionProblem::NestedToolbelt,
+        });
+    }
     Ok(CommandDef {
         name,
         namespace: None,
@@ -442,14 +455,31 @@ pub fn parse_toolbelt(value: Option<&str>) -> Option<Vec<String>> {
     (!tools.is_empty()).then_some(tools)
 }
 
+/// The frontmatter keys that name a toolbelt, in every spelling a loader
+/// here reads.
+const TOOLBELT_KEYS: [&str; 3] = ["tools", "allowed-tools", "allowed_tools"];
+
+/// Whether a nested mapping sat under a key that grants tools.
+///
+/// ADR 0025's rule: a nested value is refused where reading it wrong would
+/// widen what the file may do, and tolerated everywhere else. A toolbelt key
+/// left empty by nesting reads as "every tool", so it is the widening case.
+fn nests_a_toolbelt_key(fm: &Frontmatter) -> bool {
+    fm.nested_parents
+        .iter()
+        .any(|parent| TOOLBELT_KEYS.contains(&parent.as_str()))
+}
+
 /// Parse one agent file.
 pub fn agent_from_file(path: &str, raw: &str) -> Result<AgentDef, ExtensionDiagnostic> {
     let (fm, name, description, body) = definition_from_file(path, raw, "AGENT.md")?;
     let declared = fm.data.get("tools").map(String::as_str);
-    // A nested mapping leaves its parent key holding an empty value and its
-    // children in `nested_keys`, which is indistinguishable here from a bare
-    // `tools:` — and a bare one means every tool. Refuse rather than widen.
-    if declared.is_some_and(str::is_empty) && !fm.nested_keys.is_empty() {
+    // A nested mapping leaves its parent key holding an empty value, which
+    // reads here exactly like a bare `tools:` — and a bare one means every
+    // tool. Refuse rather than widen, but only when the toolbelt key is the
+    // one that got nested: nesting under `description:` widens nothing, and
+    // refusing for it took an agent away over a key nobody reads (ADR 0025).
+    if declared.is_some_and(str::is_empty) && nests_a_toolbelt_key(&fm) {
         return Err(ExtensionDiagnostic {
             path: path.to_string(),
             problem: ExtensionProblem::NestedToolbelt,
@@ -1006,9 +1036,9 @@ mod tests {
     /// The frontmatter parser is tolerant on purpose, and for `description:`
     /// or `model:` an unreadable value costs a fallback. For `tools:` it costs
     /// the whole registry, because the parser leaves the key holding an empty
-    /// value, the children go to `nested_keys`, and an empty `tools:` is the
-    /// documented spelling of "all tools". The author asked for two tools and
-    /// was handed every one, silently.
+    /// value, the children go to `nested_keys` under a `tools` parent, and an
+    /// empty `tools:` is the documented spelling of "all tools". The author
+    /// asked for two tools and was handed every one, silently.
     #[test]
     fn a_tools_mapping_is_refused_rather_than_read_as_every_tool() {
         let raw =
@@ -1038,6 +1068,34 @@ mod tests {
     /// A bad `description:` costs a label, since the first line of the body
     /// stands in. A bad `tools:` costs the whole registry. Refusing the file
     /// for the first would take an agent away over a key nobody reads.
+    /// The false stop the parent-key record removes. A bare `tools:` means
+    /// every tool, and nesting under `description:` widens nothing — but the
+    /// refusal used to read "the toolbelt is empty AND something is nested"
+    /// and take the agent away.
+    #[test]
+    fn a_bare_toolbelt_beside_an_unrelated_nesting_still_loads() {
+        let raw = "---\nname: reviewer\ntools:\ndescription:\n  short: s\n  long: l\n---\nReview.";
+        let agent = agent_from_file("/x/reviewer.md", raw).expect("nothing widened, so it loads");
+        assert_eq!(agent.tools, None, "a bare `tools:` is still every tool");
+    }
+
+    /// A nested `allowed-tools:` in a command file is the same widening
+    /// `agent_from_file` refuses: the key is left empty, and empty means no
+    /// restriction.
+    #[test]
+    fn a_command_with_a_nested_toolbelt_is_refused() {
+        let raw = "---\nname: ship\nallowed-tools:\n  bash: true\n---\nShip it.";
+        let err = command_from_file("/x/ship.md", raw).unwrap_err();
+        assert_eq!(err.problem, ExtensionProblem::NestedToolbelt);
+
+        let ok = command_from_file(
+            "/x/ok.md",
+            "---\nname: ok\nallowed-tools:\ndescription:\n  short: s\n---\nGo.",
+        )
+        .expect("nesting under a key that grants nothing still loads");
+        assert_eq!(ok.allowed_tools, None);
+    }
+
     #[test]
     fn nesting_under_a_key_that_grants_nothing_still_loads() {
         let raw = "---\nname: reviewer\ndescription:\n  short: s\n  long: l\n---\nReview the diff.";
--- a/crates/stella-cli/src/fleet_commits.rs
+++ b/crates/stella-cli/src/fleet_commits.rs
@@ -154,6 +154,14 @@ impl ToolExecutor for CommitObserver<'_> {
         self.inner.live_services()
     }
 
+    /// Forwarded: this observer watches commits, it mounts no invocation
+    /// plane of its own. The empty default would read as "no skill is running
+    /// here", and a live skill's procedure text would stop surviving overflow
+    /// summarization for every fleet worker composed through it.
+    fn active_skill_slugs(&self) -> Vec<String> {
+        self.inner.active_skill_slugs()
+    }
+
     /// Forwarded: letting the empty default stand would silently serialize
     /// the inner executor's sibling spawns (see the port's contract) — the
     /// spawn tool is not commit-lane routed, so nothing is observed for it.
@@ -411,6 +419,21 @@ mod tests {
 
     fn shas(commits: &[CommitRecord]) -> Vec<&str> {
         commits.iter().map(|c| c.sha.as_str()).collect()
+    }
+
+    /// Every method the port tells a decorator to forward, over this
+    /// observer.
+    ///
+    /// It wraps every shared-tree fleet worker's stack and changes nothing
+    /// about what the stack below it mounts. `active_skill_slugs` was the one
+    /// it dropped.
+    #[test]
+    fn the_commit_observer_forwards_what_a_decorator_must() {
+        let loaded = stella_tools::forwarding::LoadedExecutor;
+        let tree = Arc::new(SharedTree::default());
+        let observer =
+            CommitObserver::new(&loaded, ScriptedGit(tree.clone()), Path::new("/repo"), "t1");
+        stella_tools::forwarding::assert_forwards("CommitObserver", &observer);
     }
 
     /// The observer wraps every shared-tree fleet worker's stack, so one that
--- a/crates/stella-cli/src/self_driving_cmd/work.rs
+++ b/crates/stella-cli/src/self_driving_cmd/work.rs
@@ -83,6 +83,10 @@ use std::process::{Command, Stdio};
 
 use stella_protocol::issue::Issue;
 
+mod steering;
+
+pub(super) use steering::refuse_if_unsteered;
+
 use super::budget::RunBudget;
 use super::turn_flags::TurnFlags;
 use crate::settings::toml_config::WorkerKind;
@@ -787,64 +791,6 @@ pub(crate) fn start(
     Ok(outcome)
 }
 
-/// Refuse to work an issue with the workspace's steering switched off.
-///
-/// **A loop-driven turn must get exactly the steering a person-driven turn
-/// gets.** The whole design says the loop's behaviour comes from context
-/// records — how this repository wants code written, what to prefer, what to
-/// harden — and none of that reaches a turn when project steering is untrusted.
-///
-/// The trap is that it fails *silently and successfully*: the turn runs, writes
-/// plausible code, commits, and the pull request looks like every other one. A
-/// loop working unsteered is not a degraded loop, it is a loop doing work under
-/// nobody's standards, and it is worse than one that did not run — so this
-/// refuses rather than warns.
-///
-/// It refuses only when there is something to lose: a workspace with no records
-/// has no steering to miss, and demanding a trust flag from it would be
-/// ceremony.
-pub(super) fn refuse_if_unsteered(root: &Path) -> Result<(), String> {
-    refuse_unless_trusted(root, crate::settings::project_code_execution_trusted())
-}
-
-/// The rule of [`refuse_if_unsteered`], with the process it reads taken out.
-///
-/// Separated for the reason [`classify`] is: `project_code_execution_trusted`
-/// answers from the process environment, which this test suite shares with
-/// every other test running beside it. A pure function takes the answer as an
-/// argument, so both directions of the rule can be pinned without a race.
-fn refuse_unless_trusted(root: &Path, trusted: bool) -> Result<(), String> {
-    let records = root.join(".stella").join("rules");
-    let count = std::fs::read_dir(&records)
-        .map(|entries| {
-            entries
-                .filter_map(Result::ok)
-                .filter(|e| {
-                    e.path()
-                        .extension()
-                        .is_some_and(|ext| ext.eq_ignore_ascii_case("toml"))
-                })
-                .count()
-        })
-        .unwrap_or(0);
-
-    if count == 0 || trusted {
-        return Ok(());
-    }
-
-    Err(format!(
-        "refusing to work an issue with this workspace's steering switched off.\n\
-         \n\
-         {} declares context records and none of them would reach the turn, so it \
-         would write code under nobody's standards — and it would look exactly like \
-         a turn that did.\n\
-         \n\
-         Set STELLA_TRUST_PROJECT=1 to let this repository steer the loop it is \
-         driving.",
-        records.display()
-    ))
-}
-
 /// Turn `git worktree add`'s branch collision into an actionable message.
 ///
 /// The slug is deterministic per issue — deliberately, so `deliver` can find
@@ -1380,42 +1326,6 @@ mod tests {
         assert!(
             budget.exhausted().is_some(),
             "which is the condition `drive` reports as *budget reached*"
-        );
-    }
-
-    /// **Witness.** The loop refuses to work an issue when the workspace's
-    /// records would not reach the turn, and runs when they would.
-    ///
-    /// The failure this guards is silent. An untrusted checkout loads none of
-    /// its records, so the turn writes plausible code under nobody's standards
-    /// and the pull request looks like every other one. Both directions are
-    /// asserted, because a check that only ever refused would be satisfied by
-    /// a function that always refuses.
-    ///
-    /// A workspace with no records is the third cell: there is no steering to
-    /// miss, so asking it for a trust flag would be ceremony.
-    #[test]
-    fn the_loop_will_not_work_an_issue_that_its_records_cannot_steer() {
-        let bare = tempfile::tempdir().expect("workspace");
-        assert!(
-            refuse_unless_trusted(bare.path(), false).is_ok(),
-            "a workspace with no records has no steering to miss"
-        );
-
-        let steered = tempfile::tempdir().expect("workspace");
-        let rules = steered.path().join(".stella").join("rules");
-        std::fs::create_dir_all(&rules).expect("rules directory");
-        std::fs::write(rules.join("ctx.example.one.toml"), "").expect("a record file");
-
-        let refusal = refuse_unless_trusted(steered.path(), false)
-            .expect_err("records that cannot steer must stop the work");
-        assert!(
-            refusal.contains("STELLA_TRUST_PROJECT"),
-            "the refusal must name the remedy: {refusal}"
-        );
-        assert!(
-            refuse_unless_trusted(steered.path(), true).is_ok(),
-            "a trusted workspace steers the turn, so the work proceeds"
         );
     }
 
--- a/crates/stella-cli/src/self_driving_cmd/work/steering.rs
+++ b/crates/stella-cli/src/self_driving_cmd/work/steering.rs
@@ -1,0 +1,276 @@
+//! Whether this workspace has steering the loop would lose by running
+//! untrusted.
+//!
+//! Split out of `work.rs` so the counting rule has room to be precise. The
+//! question is narrow: does `<root>/.stella/rules` hold a record that would
+//! reach a turn if the project were trusted? A `.toml` file sitting there is
+//! not the answer, because two kinds of file live beside the records and
+//! neither ever steers anything.
+
+use std::path::Path;
+
+use crate::rules::RESERVED_RULE_FILENAMES;
+
+/// Refuse to work an issue with the workspace's steering switched off.
+///
+/// **A loop-driven turn must get exactly the steering a person-driven turn
+/// gets.** The whole design says the loop's behaviour comes from context
+/// records — how this repository wants code written, what to prefer, what to
+/// harden — and none of that reaches a turn when project steering is untrusted.
+///
+/// The trap is that it fails *silently and successfully*: the turn runs, writes
+/// plausible code, commits, and the pull request looks like every other one. A
+/// loop working unsteered is not a degraded loop, it is a loop doing work under
+/// nobody's standards, and it is worse than one that did not run — so this
+/// refuses rather than warns.
+///
+/// It refuses only when there is something to lose: a workspace with nothing
+/// that would steer has no steering to miss, and demanding a trust flag from it
+/// would be ceremony.
+pub(in crate::self_driving_cmd) fn refuse_if_unsteered(root: &Path) -> Result<(), String> {
+    refuse_unless_trusted(root, crate::settings::project_code_execution_trusted())
+}
+
+/// The rule of [`refuse_if_unsteered`], with the process it reads taken out.
+///
+/// Separated because `project_code_execution_trusted` answers from the process
+/// environment, which this test suite shares with every other test running
+/// beside it. A pure function takes the answer as an argument, so both
+/// directions of the rule can be pinned without a race.
+fn refuse_unless_trusted(root: &Path, trusted: bool) -> Result<(), String> {
+    let records = root.join(".stella").join("rules");
+    if trusted || !holds_a_steering_record(&records) {
+        return Ok(());
+    }
+
+    Err(format!(
+        "refusing to work an issue with this workspace's steering switched off.\n\
+         \n\
+         {} declares context records and none of them would reach the turn, so it \
+         would write code under nobody's standards — and it would look exactly like \
+         a turn that did.\n\
+         \n\
+         Set STELLA_TRUST_PROJECT=1 to let this repository steer the loop it is \
+         driving.",
+        records.display()
+    ))
+}
+
+/// Whether `dir` holds at least one record that trust would let steer a turn.
+///
+/// Counting `*.toml` files answered a different question, and stopped the loop
+/// over two kinds of file that steer nothing either way (#5712):
+///
+/// - `governance.toml` and anything else in [`RESERVED_RULE_FILENAMES`], which
+///   the record loader never parses as a record at all.
+/// - A record whose `status` is `archived` or `retracted`. The registry's
+///   `blocking_reason` refuses to let those steer, so a workspace that has
+///   retired its records has nothing left to lose.
+///
+/// Reads and parses the files rather than taking the session's registry:
+/// `load_registry` runs truth probes and writes a sweep cache, which would put
+/// filesystem probes in front of every work unit. Nothing here creates a
+/// directory.
+fn holds_a_steering_record(dir: &Path) -> bool {
+    let Ok(entries) = std::fs::read_dir(dir) else {
+        return false;
+    };
+    entries.filter_map(Result::ok).any(|entry| {
+        let path = entry.path();
+        let is_toml = path
+            .extension()
+            .is_some_and(|ext| ext.eq_ignore_ascii_case("toml"));
+        let reserved = path
+            .file_name()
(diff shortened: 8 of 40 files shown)


macanderson:
Not my PR — leaving it to you, but I dug into the red `CodeQL` since it is the
only thing blocking and it looked like it might be serious. I think it is a
false positive, and here is the evidence so you do not have to redo it.

**What it says:** `4 new alerts including 4 high severity security
vulnerabilities`, all the same rule — *Cleartext logging of sensitive
information* — each reported as "writes `refuse_unless_trusted(...)` to a log
file":

```
crates/stella-cli/src/self_driving_cmd.rs:1274
crates/stella-cli/src/self_driving_cmd.rs:1285
crates/stella-cli/src/self_driving_cmd/audit.rs:198
crates/stella-cli/src/self_driving_cmd/audit.rs:199
```

**Those four sites are not in your diff.** This PR touches
`self_driving_cmd/work.rs` and adds `work/steering.rs`; it does not modify
`self_driving_cmd.rs` or `audit.rs`, and all four lines already exist on `main`
unchanged. So the alert is about a data flow the PR creates into pre-existing
code, not about a line it wrote.

**What the flow actually carries.** The source is
`refuse_unless_trusted(root: &Path, trusted: bool) -> Result<(), String>` in
`work.rs:816` — it counts `.toml` files under `.stella/rules` and returns a
refusal message. The value CodeQL treats as sensitive is the **workspace path**,
because its model classes filesystem paths that way.

The sinks are `println!` — `audit.rs:198-199` print an action label and an
outcome, `self_driving_cmd.rs` prints a branch name and a diffstat. Those are
this command's output on **stdout**, not a log file, and a refusal message naming
the directory it refused to work in is the message doing its job.

**So: a path into a terminal message, classed as a secret into a log.** I would
call that a false positive. I did not verify it against the full taint path
CodeQL computed — only the source, the four sinks, and that none of them are in
your diff — so treat this as a reading rather than a verdict.

If you agree, the usual resolution is dismissing the four alerts as "used in
tests / false positive" with that reason, rather than restructuring correct
output code around the analyser. Worth a sentence in the PR description either
way, since the alerts will follow the flow rather than the lines.

macanderson:
## `ci` was red on `f273ae3` — clippy and rustdoc, both from #5799

[Run 34061713300](https://github.com/macanderson/stella/actions/runs/34061713300):
`cargo clippy -D warnings` failed at 21:42:56 and `cargo doc -D warnings` at
21:43:44. `cargo test` passed (8m33s), so the two lint steps were the whole of
it. Pushed `c8938da`.

Both come from #5799's extraction of `decreed_by_a_named_human_at` into
`trust.rs`:

**clippy** — `bridge.rs:50` and `sweep.rs:40` kept `TruthBasis` in their
top-level imports after the extraction left the non-test code with no use for
it. Deleting the import outright is wrong (I tried it first): both `#[cfg(test)]`
modules still construct `TruthBasis::Decree` / `::Measured`, so it moves into
their existing `ingest::record::{…}` import lists instead.

**rustdoc** — `honored_probe` is `pub` and its doc linked
`` [`decreed_by_a_named_human_at`] ``, which is `pub(crate)`. That is
`rustdoc::private_intra_doc_links`, the error AGENTS.md flags as the one only
the `doc-warnings` step catches — neither `guards-fast` nor a scoped clippy
builds docs. Named in prose now, the same shape this batch already used for
`seat_key` in #5979.

Verified on the crate:

```
cargo clippy -p stella-records --all-targets -- -D warnings           # clean
RUSTDOCFLAGS="-D warnings" cargo doc -p stella-records \
    --no-deps --document-private-items                                # clean
cargo test -p stella-records                                          # 2 passed
cargo fmt --check -p stella-records                                   # clean
```

Worth flagging for the DoD: neither failure is in a crate this PR's description
lists as touched — `stella-records` reached the diff only through #5799 — and
`cargo doc` is why the second one survived local checks.

---

macanderson:
## CodeQL is red on `c8938da` — 4 high severity, and I cannot read them

Reporting rather than fixing, because I cannot see what the alerts are. What is
established, and what is not:

**Established.** The CodeQL *workflow* passed on this head — run
[34063281032](https://github.com/macanderson/stella/actions/runs/34063281032)
concluded `success`, all four analyses (rust, python, actions,
javascript-typescript) green, same as on `f273ae3`. The red signal is the
separate code-scanning check that gates on *new alerts in the PR's diff*, which
is a different object from the workflow.

**Established.** It is not from my commit. `git diff f273ae3..c8938da` is five
lines: `TruthBasis` moving between an outer and a test-module import list in two
files, plus a reworded doc comment. There is no construct there that any
security query could fire on. The alerts are attributed to the PR's whole diff
against `main`, not to the head that happened to trigger the report.

**Not established — and I want to be plain about it.** I do not know what the
four alerts are. The check run exposes no annotation text through the tooling I
have, there is no code-scanning API surface available to me, and the id is a
check run rather than a job, so the log route 404s.

**A hypothesis I checked and dropped.** The obvious candidates in this diff for
a high-severity `actions` or `python` query are `.github/workflows/windows-check.yml`
and `scripts/check-core-reachability.py`, and 4 alerts across the 4 analysed
languages looked suggestive. Reading both: the workflow adds a path-filter entry,
a crate name to an existing `cargo clippy` line, and comments — no `${{ }}`
interpolation of event data, no new `run:` consuming untrusted input. The Python
adds a pure `retired_because` helper and changes two print statements — no
subprocess, no path handling from untrusted input, no `eval`. Neither is
injection-shaped, so I am not offering that as the answer.

**Open question worth someone checking:** whether these four alerts were already
present at `f273ae3` and simply not reported. That head's `report CodeQL status`
job was **skipped** (job 101564946105) while this head's ran, which would explain
a same-diff alert set surfacing only now. If that is what happened, the alerts
predate both my push and possibly this PR's later commits.

**What would settle it:** the four rule ids and locations from
`/macanderson/stella/security/code-scanning?query=pr%3A6323+tool%3ACodeQL+is%3Aopen`.
Paste them here and I will fix them or say why each is a false positive. I am
not re-running the check — the analysis already succeeded, so a re-run reports
the same alerts and tells us nothing.

Not touching this PR's other state: the clippy and rustdoc failures I pushed
`c8938da` for are a separate matter and that fix stands.

---

macanderson:
## Answering Sourcery's two ❌ rows — both rest on a diff Sourcery did not fully read

Per CLAUDE.md every Sourcery ❌ needs a fix or an answer before this is mergeable. Neither of these belongs to the PR: they are a misreading, and the misreading is mechanically checkable.

**The claim.** Both ❌ rows rest on one assertion — that the PR "does not address or list as unable to ride #2792 (approval emission deduplication), #5749 (three spend-gate scenarios), #5512 (LoadedExecutor forwarding tests), or #5746 (parked-wait attempt-count coverage)."

**All four are in the diff, with a named section in the description and the code to match:**

| Member | Description section | Files carrying it |
|---|---|---|
| #2792 | "one approval question wrote two journal rows" | `stella-core/src/bus/policy_bridge.rs` (+208, new), `bus.rs` (−59) |
| #5749 | "three call-buying paths the spend gate did not price" | `stella-core/tests/spend_gate.rs` (+205) |
| #5512 | "a decorator that forgets to forward failed nothing" | `stella-tools/src/forwarding.rs` (+176, new), `claims.rs`, `command_deck/task_tap.rs`, `fleet_commits.rs`, `gated.rs`, `skill_plane.rs`, `custom/tests.rs` |
| #5746 | "how many attempts a parked wait can buy in total" | `stella-core/src/driver/rate_limit.rs` (`pub mod` + `pub const`), `tests/spend_gate.rs` |

**What went wrong, and how to check it.** The PR changes **39 files**. Sourcery's own File-Level Changes table names **23**. The 16 it never mentions are:

```
crates/stella-cli/src/claims.rs                    crates/stella-core/src/driver.rs
crates/stella-cli/src/command_deck/task_tap.rs     crates/stella-core/src/driver/rate_limit.rs
crates/stella-cli/src/fleet_commits.rs             crates/stella-core/tests/spend_gate.rs
crates/stella-cli/src/settings.rs                  crates/stella-tools/src/custom/tests.rs
crates/stella-cli/src/settings/tests/trust.rs      crates/stella-tools/src/forwarding.rs
crates/stella-core/src/bus.rs                      crates/stella-tools/src/gated.rs
crates/stella-core/src/bus/policy_bridge.rs        crates/stella-tools/src/lib.rs
                                                   crates/stella-tools/src/skill_plane.rs
                                                   docs/spec/verification-gate.md
```

That set is not arbitrary — it is precisely the files implementing the four members reported missing. Sourcery's guide opens with "ten contained audit fixes"; the batch has fifteen. The five it did not summarize are the four above plus #5334, and #5334 escaped the ❌ only because its test files (`skills/tests/selection.rs`) happened to land in the half Sourcery did read.

So the assessment is not a finding about the change — it is a report over a subset of it, and the omitted subset is what it reports as absent.

**Corroborating signal:** `dod / dod` passes on this head, and the SCR-003 bot's own comment reads "every linked issue's definition of done is fully checked." That check reads #6298's checklist directly rather than a summary of the diff.

No code change is owed for either row. The two members that genuinely could not ride — #5716 (needs a bench rig and provider spend) and #6261 (larger than the session) — are named in the description with the defer case for each, which is what the first objective asks for.

Not touching the rest of this PR's state: the CodeQL question is answered in the comments above and the clippy/rustdoc fix in `c8938da` stands.

---

macanderson:
## `fmt + clippy + test` is red on `c8938da` — one flaky test in a crate this PR does not touch

[Job 101567958480](https://github.com/macanderson/stella/actions/runs/34063281051/job/101567958480), step `cargo test`, one failure out of 155 in `stella-fleet`:

```
---- fleet::tests::a_cancelled_dispatch_releases_its_claims_and_control_handle stdout ----
thread '...' panicked at crates/stella-fleet/src/fleet/tests.rs:501:9:
the parked worker is registered
test result: FAILED. 154 passed; 1 failed; 0 ignored
```

### Why this is not this PR's failure

**Observed, not inferred.** On this branch's previous head `f273ae3`, [job 101563455329](https://github.com/macanderson/stella/actions/runs/34061713300/job/101563455329) step 50 `cargo test` concluded **success** (21:43:44 → 21:52:17). I read that from the job's own step record rather than from the earlier comment that asserts it. `git diff f273ae3..c8938da` is five lines — `TruthBasis` moving between an outer and a test-module import list in `stella-records`' `bridge.rs` and `sweep.rs`, plus a reworded doc comment.

**The diff does not reach the failing crate.** None of this PR's 39 files is under `crates/stella-fleet/`. The nearest-looking touch is `stella-cli/src/claims.rs` (`ClaimTap`), but `stella-cli` depends on `stella-fleet`, not the reverse — a change there cannot alter fleet's behavior. Nothing in the diff changes what `Fleet::dispatch` awaits.

### What the test actually asserts, and why it can fail on its own

`tests.rs:471`:

```rust
let mut dispatch = Box::pin(f.dispatch(&task));
// One poll drives dispatch through the claim, the ledger rows and
// the control registration, and parks it in the gated worker.
assert!(dispatch.as_mut().now_or_never().is_none(), "...");
assert!(f.pause_task(&id), "the parked worker is registered");
```

`now_or_never()` polls **exactly once**. The comment above it states the assumption the test rests on: that one poll carries the future past claim acquisition, the ledger writes and the control registration, parking only at the gated worker. That holds only while every `await` before the registration is immediately ready. If any of them returns `Pending` even once — the claim `Store`, the in-memory `Ledger` — the future parks earlier, registration has not happened, and the second assertion fails exactly as it did here.

The first assertion passing and the second failing is what that looks like: the future was still in flight, just parked one await sooner than the test assumes.

So the test encodes a scheduler-ordering assumption as if it were a structural one. I am labelling that a **reading of the test source**, not a proven root cause — I did not instrument the poll or reproduce it.

### What I did not do, and why

I judged this a flake and went to spend the one re-run AGENTS.md allows for a failure naming a crate the diff does not touch. **I do not have permission:** `rerun-failed-jobs` on run 34063281051 returns `403 Resource not accessible by integration`. Per AGENTS.md's own branch for that case — a flaky test that cannot be made robust within this PR's scope — I am saying so once rather than widening a 39-file PR into a crate it has no business in, and keeping the PR watched.

### Proposed patch, for whoever takes it

Replace the single poll with a bounded poll-and-yield loop so the test waits for the state it is asserting instead of assuming one poll reaches it:

```rust
let mut registered = false;
for _ in 0..64 {
    assert!(
        dispatch.as_mut().now_or_never().is_none(),
        "the gated worker parks, so the dispatch is still in flight"
    );
    if f.control_handle_registered(&id) { registered = true; break; }
    tokio::task::yield_now().await;
}
assert!(registered, "the parked worker is registered");
```

This needs a non-mutating observation — `pause_task` is the wrong probe to poll with, since the assertion pair at lines 501 and 505 reads it as a state query while its name says it mutates. Whether `stella-fleet` already exposes such a predicate I did not check; if it does not, adding one is the smaller half of the fix. Deliberately not pushed here: I cannot run `cargo test -p stella-fleet` to show the fix red-then-green, and an unverified patch to a flaky concurrency test is how a flake becomes a silent pass.

Filed as a separate issue since it cannot responsibly ride this PR — it costs CI time and merge confidence on every PR that touches enough crates to pull `stella-fleet` in, which is not a property of this one.

**A re-run of that job by anyone with the permission is what unblocks this PR.** The rest of its state is unchanged: CodeQL is answered above, the Sourcery ❌ rows are answered above, and the clippy/rustdoc fix in `c8938da` stands.

---

macanderson:
Two checks are red on `c8938da`. Recording what each one is, and which half of it I verified versus inferred, so the next session does not repeat the dig.

## 1. `fmt + clippy + test` — one test failure, in a crate this PR does not touch

[Job 101567958480](https://github.com/macanderson/stella/actions/runs/34063281051/job/101567958480):

```
---- fleet::tests::a_cancelled_dispatch_releases_its_claims_and_control_handle stdout ----
panicked at crates/stella-fleet/src/fleet/tests.rs:501:9:
the parked worker is registered

test result: FAILED. 154 passed; 1 failed; 0 ignored
error: test failed, to rerun pass `-p stella-fleet --lib`
```

**This is the first run of this PR that reached the test suite.** The previous head (`f273ae3c`) died earlier in the same job, at clippy and rustdoc:

```
error: unused import: `TruthBasis`   crates/stella-records/src/records/bridge.rs:50:61
error: unused import: `TruthBasis`   crates/stella-records/src/records/sweep.rs:40:51
error: public documentation for `honored_probe` links to private item `decreed_by_a_named_human_at`
                                  crates/stella-records/src/records/sweep.rs:164:9
```

`c8938da` fixed exactly those, which is why the tests ran at all this time. So there is no earlier data point for this test on this branch — it is not a case of a failure that was already there and got worse.

**Observed** — the facts, each checkable:

- This PR changes no file under `crates/stella-fleet/`.
- `stella-fleet` reaches into `stella-core` for three items only: `BudgetGuard`, `BudgetOutcome`, `Clock` (`fleet.rs:40`, `monitor.rs:24`, `fleet/notice.rs:52`).
- This PR's `stella-core` edits are `bus.rs` (+ the new `bus/policy_bridge.rs`), `driver.rs`, `driver/lifecycle.rs`, `driver/rate_limit.rs`, `engine_markers.rs`, `estimator.rs`, `subagent.rs`. None is budget or clock.
- `main` at `893f5e54` is green ([run 34062781432](https://github.com/macanderson/stella/actions/runs/34062781432)), and #6324 was all-green at its `69e47675` head. Both ran this test.

**Inferred, not verified** — from those four together, I see no causal path from this diff to `Fleet::dispatch`'s poll ordering, which is what the assertion at `tests.rs:501` reads. The assertion holds on the state after a single `now_or_never()` poll, and `dispatch`'s first await is `acquire_dispatch_lease(task).await` — an in-memory SQLite lease write, whose completion inside one poll is a timing property of that write, not of anything above it.

I did **not** run the control the house rules ask for. `rerun-failed-jobs` on this run returns `403 Resource not accessible by integration` for this integration, so the one permitted re-run was not available to me, and I did not build locally — a green here would be a cost, not evidence. So the inference above is where my evidence stops: it is a reason to suspect the failure is not this PR's, not a demonstration of it.

Deliberately not hardening the test: `crates/stella-fleet/` is outside this PR's scope, and widening a 39-file diff to quiet a red is the wrong direction. If a re-run comes back green, that settles it; if it fails a second time on an unchanged tree, the `now_or_never()` shape at `tests.rs:490-503` is the thing to look at, and it wants its own issue rather than a passenger commit here.

## 2. CodeQL — 4 new high-severity alerts, in two files not in this diff

[Check run 101567689598](https://github.com/macanderson/stella/runs/101567689598), all four the same rule, *Cleartext logging of sensitive information*, each reported as writing `refuse_unless_trusted(...)` to a log:

| path | line |
|---|---|
| `crates/stella-cli/src/self_driving_cmd.rs` | 1274 |
| `crates/stella-cli/src/self_driving_cmd.rs` | 1285 |
| `crates/stella-cli/src/self_driving_cmd/audit.rs` | 198 |
| `crates/stella-cli/src/self_driving_cmd/audit.rs` | 199 |

Neither file appears in this PR's changed-file list. The sibling `self_driving_cmd/work.rs` is changed and `self_driving_cmd/work/steering.rs` is new, so the likely reading is that the new code opened a dataflow path into logging calls that were already there — CodeQL reports those under "new alerts in code changed by this pull request" because the *flow* is new, not the line. That reading is an inference; the alert pages under `security/code-scanning?query=pr%3A6323` carry the actual paths, and this integration cannot read them (`403` on the code-scanning API), so someone with that access should look before deciding.

Worth settling rather than dismissing: whether what `refuse_unless_trusted` returns is sensitive is a real question — if it names a path or a plugin the operator already knows, the alert is noise and belongs in a dismissal with that reason recorded; if it can carry anything from an untrusted manifest, four high alerts on the self-driving path is the finding. Either way it is a judgement about this repository's own trust boundary, which is why I am naming it rather than silencing it.

Nothing else on this PR is red: `dod / dod`, the guards, MSRV, the Windows and macOS splits, `docs/wire`, dependency-review and the file-size ratchet are all green on `c8938da`.

---

macanderson:
## `c8938da`: clippy and rustdoc are green; `cargo test` failed on a `stella-fleet` test this PR cannot reach

**The fix landed.** In [run 34063281051](https://github.com/macanderson/stella/actions/runs/34063281051),
`cargo clippy -D warnings` and `cargo doc -D warnings` both pass on this head —
the two failures `c8938da` was pushed for.

**What is failing now** is one test, and it is not this PR's:

```
test fleet::tests::a_cancelled_dispatch_releases_its_claims_and_control_handle ... FAILED
test result: FAILED. 154 passed; 1 failed
panicked at crates/stella-fleet/src/fleet/tests.rs:501:9
```

### Why it is not this PR's

- **Structurally unreachable.** `c8938da` changes `stella-records` only — two
  `TruthBasis` imports moving between an outer and a test-module list, and a doc
  comment. `crates/stella-fleet/Cargo.toml` depends on `stella-autonomy`,
  `stella-protocol`, `stella-core`, `stella-store` and `stella-tools`; `rg
  'stella[-_]records' crates/stella-fleet/` returns nothing. There is no path
  from the change to the test.
- **It passed on the previous head of this same PR.** `f273ae3`'s `cargo test`
  step was `success` (run 34061713300, 8m33s) with the identical `stella-fleet`
  tree.
- **It passed on `main`.** Run 34062781432 on `893f5e5` is green.

### Why it is flaky rather than broken

Line 501 is `assert!(f.pause_task(&id), "the parked worker is registered")`,
and it runs after a **single** `now_or_never()` poll. That asserts the dispatch
future reaches control registration within one poll — true only while every
await before it (claim store, ledger rows) is ready synchronously. Under a
loaded runner any one of them yielding leaves the handle unregistered and the
assert fires. Nothing about the code under test is wrong; the test pins a
scheduling detail rather than the behaviour it names.

### Proposed patch (not pushed — it would widen this PR)

Poll to the condition instead of asserting a poll count:

```rust
let mut dispatch = Box::pin(f.dispatch(&task));
// Drive the dispatch until the control handle exists, rather than
// assuming one poll reaches it: every await before registration is
// ready synchronously today, and a loaded runner is all it takes for
// one of them to yield.
let registered = {
    let mut seen = false;
    for _ in 0..64 {
        if dispatch.as_mut().now_or_never().is_some() {
            panic!("the gated worker must park, not complete");
        }
        if f.pause_task(&id) {
            seen = true;
            break;
        }
        tokio::task::yield_now().await;
    }
    seen
};
assert!(registered, "the parked worker is registered");
```

The bound keeps it a real assertion — it still fails if registration never
happens — while removing the exactly-one-poll assumption.

I attempted a local reproduction and it was **inconclusive** (my filter matched
no tests, so the run proves nothing either way) — the case above rests on the
dependency graph and the two green runs, not on a local repro I did not get.

This is the one re-run the flake rule allows, spent on confirming a failure the
diff cannot touch. If it fails a second time on the same test, that is real and
I was wrong.

---

macanderson:
Correction to the last line above: I could not spend that re-run.
`rerun-failed-jobs` on run 34063281051 returns `403 Resource not accessible by
integration`, so the confirming second run has to come from someone with the
permission — or from the next push to this branch.

Nothing else in that comment changes: clippy and rustdoc are green on
`c8938da`, and the `stella-fleet` failure is still one this diff cannot reach.
I am not pushing the proposed patch, because `stella-fleet` is outside this
PR's scope and fixing it here would widen the PR. Keeping the PR watched
instead.

---

macanderson:
Withdrawing my earlier ask for the CodeQL rule ids — the comment above supplies
them, and I checked the load-bearing half against the code rather than taking
it on trust. It holds:

- **The move is real.** `git diff --stat origin/main...HEAD -- crates/stella-cli/src/self_driving_cmd/`
  is `work.rs` −94 and a new `work/steering.rs` +269, and `refuse_unless_trusted`
  is at `steering.rs:36`. A new destination file is exactly what makes CodeQL
  score a pre-existing dataflow as new in this PR's diff.
- **The value is not a secret.** `refuse_unless_trusted` interpolates exactly one
  thing: `records.display()`, where `records = root.join(".stella").join("rules")`
  — a workspace-local directory path built from the workspace root. It goes into
  an `Err(String)` refusal that callers propagate with `?`
  (`work.rs:694`, `parallel.rs:89`) to the operator's own terminal. No
  credential, no user data, no remote sink.

So *Cleartext logging of sensitive information* is a false positive here, on all
four, and nothing in this PR introduced the path it traces.

That leaves this PR blocked on one thing only: `fmt + clippy + test`, red on the
`stella-fleet` test neither of us could reproduce (68 runs between us, zero
reproductions, and `main` green on the same test) and which neither of us can
re-run — `rerun-failed-jobs` is a 403 for both integrations. It needs someone
with the permission to re-run that job, or the next push to this branch.

Correcting one thing I got wrong earlier in this thread: I guessed the alerts
might come from `windows-check.yml` or `check-core-reachability.py`, read both,
and dropped it. The moved-function explanation is the right one and I did not
find it.

---

macanderson:
## The fleet test that reddened this branch was a pre-existing flake, and it is fixed here

`fleet::tests::a_cancelled_dispatch_releases_its_claims_and_control_handle`
failed once on this branch (run 34063281051). It is not caused by anything in
this diff, which touches no file in `stella-fleet`. The two earlier red runs on
this branch failed elsewhere and are not instances of it: 34061713300 stopped at
`cargo clippy` and never reached `cargo test`, and 34060075523 failed on
`settings::tests::trust::every_code_execution_gate_is_reachable_by_one_grep`,
which was a real defect of mine and is fixed in f273ae3c5.

**The mechanism.** That test is the only one in the file asserting on a *single*
poll (`now_or_never`), and the poll has to reach `register_controls`. On the way
it takes the worktree lock, which `git::worktree_lock` keys by repo root in a
process-global `LazyLock<HashMap<PathBuf, Arc<tokio::sync::Mutex<()>>>>`. Twelve
tests in that file passed `"/repo"`, cargo runs them on parallel threads, and a
sibling holding that lock parks the poll at the acquire — before the
registration. The assertion then fails naming a worker that was never
registered, rather than one that leaked.

**Reproduced, not inferred.** Holding `/repo`'s lock with a parked rival and then
polling the dispatch once fails at `tests.rs`'s `pause_task` assertion with the
same message CI reported, on an otherwise unmodified tree:

```
test fleet::tests::zz_experiment_contended_worktree_lock_breaks_the_single_poll ... FAILED
panicked at crates/stella-fleet/src/fleet/tests.rs: the parked worker is registered
```

With the repo root made this test's own, the same rival changes nothing and the
test passes. The experiment was temporary and is not in the diff; the fix is
98627598a, one string and the comment saying why it must stay unique.

Ten consecutive local runs of the whole `stella-fleet` suite were clean both
before and after, which is why the local suite is not evidence either way here —
the contention window is short, and a loaded CI runner is where it opens.

## CodeQL

`CodeQL` is red on this PR and is **not** one of `main`'s required contexts. Its
alerts are 20 open `rust/cleartext-logging` and `TokenPermissions` findings in
`stella-tool-facts`, `stella-cli`, `stella-mcp` and `stella-graph` — every one of
them in a file this PR does not touch, and all created before this branch
existed. #6050 already tracks getting a green CodeQL run on `main`, so this is
not re-filed.

macanderson:

Both rows name the same gap, and both were true of the head Sourcery read
(`023bf0ad8`, ten members). Four more members have landed since, so the rows
are now answered by the diff rather than by an argument:

| Row's complaint | Where it is now |
| --- | --- |
| #2792 not addressed | `6c161b69a` — `bus/policy_bridge.rs` holds one ask open per gate; witness `one_ask_is_one_row_however_many_emissions_it_makes` plus two negative controls. |
| #5749 not addressed | `4018db94b` — three scenarios in `tests/spend_gate.rs` for the head-dropping ladder, the starved retry and the length-continuation ladder. |
| #5746 not addressed | `4018db94b` — `a_deadline_less_parked_wait_buys_a_bounded_number_of_attempts`, derived from `plan_park` and `MAX_PARKED_WAIT_MS` rather than a literal. |
| #5512 not addressed | `92d4c3472` — `stella_tools::forwarding`'s `LoadedExecutor` and `assert_forwards`, wired into every decorator in `stella-tools` and `stella-cli`. It found two live drops (`ClaimTap`, `CommitObserver`), both now forwarding. |

Fifteen of the seventeen members are in the diff. The two that are not are
named in the description under "Could not ride" with the case that stopped
each — #5716 needs a bench arm re-run before and after (a rig and real
provider spend), and #6261 opens with a design decision the issue states
outright and then spans four crates and a spec.

## The two red checks that are not this PR's

**CodeQL.** Twenty open `rust/cleartext-logging` alerts, none of them in a
file this diff touches — `git diff origin/main...HEAD --stat` over every
alerted path is empty. The two newest were created at 2026-09-06T19:38Z, on
`main`, before this branch existed. Tracked in #6050; CodeQL is not a required
check here.

**`fmt + clippy + test`, run `34063281051`.** `stella-fleet`'s
`a_cancelled_dispatch_releases_its_claims_and_control_handle` failed on
`the parked worker is registered`. This PR touches nothing in `stella-fleet`,
and the test's own comment names the race: it asserts on a *single* poll that
has to reach `register_controls`, and a sibling test holding the
process-global worktree lock parks that poll at the acquire instead.
`98627598a` gives the test a repo root of its own, and the run on that head is
what settles it.

macanderson:
## The 4 high-severity CodeQL alerts are one false positive, four times

[The CodeQL check](https://github.com/macanderson/stella/runs/101571892088) is red with *"4 new alerts including 4 high severity security vulnerabilities"*. All four are the same rule at four sites, and none is a leak. Recording the reasoning here so the next reader does not have to re-derive it — and so that if I am wrong, the claim is on the PR where it can be contradicted.

### The four annotations

| File | Line | Rule |
|---|---|---|
| `crates/stella-cli/src/self_driving_cmd.rs` | 1274 | Cleartext logging of sensitive information |
| `crates/stella-cli/src/self_driving_cmd.rs` | 1285 | Cleartext logging of sensitive information |
| `crates/stella-cli/src/self_driving_cmd/audit.rs` | 198 | Cleartext logging of sensitive information |
| `crates/stella-cli/src/self_driving_cmd/audit.rs` | 199 | Cleartext logging of sensitive information |

Each says: *"This operation writes `refuse_unless_trusted(...)` to a log file."*

### None of those four files is in this PR's diff

`git diff origin/main...HEAD --name-only` over `self_driving_cmd` returns exactly two paths, and neither is a sink above:

```
crates/stella-cli/src/self_driving_cmd/work.rs
crates/stella-cli/src/self_driving_cmd/work/steering.rs
```

CodeQL attributes the alerts to this PR because the **source** is new — `refuse_unless_trusted` is introduced in `work/steering.rs` — while the sinks are pre-existing `println!` calls the taint now reaches.

### What the tainted value actually is

`refuse_unless_trusted` (`crates/stella-cli/src/self_driving_cmd/work/steering.rs`) returns `Result<(), String>`, and the only `Err` it can produce is this literal:

> refusing to work an issue with this workspace's steering switched off. … `{records.display()}` declares context records and none of them would reach the turn, so it would write code under nobody's standards — and it would look exactly like a turn that did. … Set `STELLA_TRUST_PROJECT=1` to let this repository steer the loop it is driving.

Two interpolations, and neither is a secret:

- `records.display()` — the path `<workspace>/.stella/rules`, derived from the root the operator ran in.
- `STELLA_TRUST_PROJECT` — the **name** of an environment variable, never its value. Nothing reads that variable into the message.

There is no credential, token, key, or user content on this path. The query almost certainly fires on the identifier heuristic — `trusted` in the function name — rather than on anything in the data.

### Why printing it is the behaviour, not the bug

The sinks are `println!` to stdout: `self_driving_cmd.rs`'s outcome lines and `audit.rs`'s `record`, whose own doc comment says the line is *"Printed first, deliberately. If the append fails the operator has still seen the line."* Suppressing a refusal reason so it stops reaching the operator would be the actual defect — the loop refuses and the person driving it is told why.

### What I have not done

I have **not** dismissed the alerts (I have no code-scanning write access on this token — the alerts API returns `Resource not accessible by integration`), and I have not added a suppression comment to the source. An `#[allow]`-shaped silencing of a security query is the kind of expedient CLAUDE.md forbids doing quietly, and it would be a change to files this PR does not otherwise touch. If a maintainer agrees with the reading above, dismissing the four alerts as *"used in tests / false positive"* in the code-scanning UI is the cheaper and more auditable move than editing the source.

Everything else on this PR is green; `fmt + clippy + test` was still running when I looked.

---

macanderson:
Following up on my earlier comment, because both halves of it moved.

## The fleet test was real, and `9862759` found the mechanism

I read it as probably not this PR's and said the `now_or_never()` shape at `tests.rs` was where to look if it recurred. That was the right place and the wrong confidence: it was not waiting on a re-run, it was a genuine test-isolation defect, and `9862759` names it exactly — `git::worktree_lock` keys by repo root in a process-global map, a dozen tests in that file shared `/repo`, cargo runs them on parallel threads, and a sibling holding that lock parks the single poll at the acquire, before `register_controls`. Giving this one test a root of its own is the fix, and `fmt + clippy + test` is green on `98627598`.

Worth recording that a re-run would have been the wrong instinct here: the failure was load-dependent, so it would likely have come back green and buried a real bug. `a_cancelled_dispatch_releases_its_claims_and_control_handle` is the only test in that file asserting on a single poll, which is what made it the one to break — a second test written that way against a shared root brings this back.

## CodeQL is now the only thing red, and here is the whole flow

Same four alerts on `98627598`, same rule (*Cleartext logging of sensitive information*), same two sinks. I traced it end to end:

**The source** is `refuse_unless_trusted`'s `Err` string, and its only interpolated value is `records.display()` — the path `<workspace root>/.stella/rules`:

```rust
Err(format!(
    "refusing to work an issue with this workspace's steering switched off.\n…\
     {} declares context records and none of them would reach the turn, …",
    records.display()
))
```

**The sinks** are four `println!` calls to stdout: the `worked #key — {stat}\n  branch: {branch}` and `the turn changed nothing ({why})` arms in `self_driving_cmd.rs`, and the two audit-line arms in `audit.rs`'s `record`.

**Why the alerts are new on this PR, when neither sink file is in the diff.** `refuse_unless_trusted` is on `main` too, in `work.rs`. This PR moves it into `work/steering.rs`, a file the PR adds — so the taint *source* now sits in code the pull request changed, and CodeQL reports the flow under "new alerts in code changed by this pull request" even though the flagged lines are untouched. The body also changed while moving (the inline `read_dir` count became `holds_a_steering_record`), which is a real edit to the source function, not just a relocation.

**What is actually reaching stdout.** A filesystem path the operator themselves pointed the command at, inside a refusal telling them how to turn steering on. CodeQL's model treats a path as sensitive because it can carry a home directory and a username; that is the entire sensitivity claim here. There is no credential, token, or model output in that string.

So the decision is a dismissal, and it is yours rather than mine: is `<root>/.stella/rules` on the operator's own terminal within this repository's bar? I would say it is — the CLI prints workspace paths throughout, and a refusal that will not name the directory it is refusing about is a worse refusal. But four **high** alerts is not something to wave through on my reading, and I cannot see the alert pages to check whether CodeQL found a fifth flow I have not traced: `GET /repos/macanderson/stella/code-scanning/alerts` returns `403 Resource not accessible by integration` for this integration, so everything above is reconstructed from the four check-run annotations plus the source.

If you would rather close it in code than dismiss it, the smallest honest change is to stop the refusal string sharing a type with what gets printed — the taint reaches these `println!`s through `WorkOutcome`, whose `Failed { reason }` arm is returned rather than printed while `Changed`/`NoChange` are printed. That is a real refactor, not a one-liner, and it belongs in its own PR rather than widening this one.

Nothing else on `98627598` is red.

---

macanderson:
## Both Sourcery ❌ rows are fixed, and its table predates the fix

Sourcery reviewed 023bf0ad8 at 20:29, when this PR carried eleven members. Its
two ❌ rows both rest on the same fact — four members were neither implemented
nor listed as unable to ride. All four have since landed, and the `Sourcery
review` check now reads `skipping`, so the table above has not been re-cut
against the current head. Answering each row rather than leaving it standing:

**Row 1 — "does not address or list as unable to ride #2792, #5749, #5512,
#5746."** All four are implemented now:

| Member | Commit | What lands |
| --- | --- | --- |
| #2792 | 6c161b69a | `bus/policy_bridge.rs` holds one ask open per gate; witness `one_ask_is_one_row_however_many_emissions_it_makes` plus two negative controls |
| #5749 | 4018db94b | three pinned scenarios in `tests/spend_gate.rs` for the head-dropping ladder, the starved retry and the length-continuation ladder |
| #5746 | 4018db94b | `a_deadline_less_parked_wait_buys_a_bounded_number_of_attempts`, derived from `plan_park` and `MAX_PARKED_WAIT_MS` rather than a literal |
| #5512 | 92d4c3472 | `stella_tools::forwarding`'s `LoadedExecutor` + `assert_forwards`, wired into every decorator in `stella-tools` and `stella-cli` |

**Row 2 — "the omitted members have no implementation or witness tests."** Same
cause, same fix. Each of the four carries its own witness, and #5512's found two
live defects while it was at it: `ClaimTap` and `CommitObserver` both swallowed
`active_skill_slugs`, so a live skill's procedure text stopped surviving
summarization on the deck's lead turn and on every shared-tree fleet worker.
Both forward it now.

The two members still deferred — #5716 and #6261 — are named under **Could not
ride** in the PR description with the case that stopped each, and both are
ticked on #6298 with that reason rather than left as silent omissions.

`cargo test` is green on the current head (21m24s), including the `stella-fleet`
test that had failed once on this branch; the comment above records why that was
a pre-existing flake and what fixes it.

Issue #6218, opened 2026-09-06
A served re-query can show the model the same frames twice, and nothing says whose job it is to stop that


macanderson:
## Problem

The served mid-turn re-query dedups whole blocks and nothing finer. A host that
answers a second ask with `{A, B, C, D}` after answering the first with
`{A, B, C}` sends different bytes, so the block is injected whole and the model
sees the first three frames twice.

The CLI does not have this problem. `ProducedSteering`
(`crates/stella-cli/src/memory/steering.rs`) keeps the frame, skill and record
handles a turn has already rendered and hands them to the next recall, so an
incremental drift re-injects only what it added. The reason that granularity
exists is written there: drift is incremental, so byte comparison answers a
question drift never asks.

`crates/stella-serve/src/remote/requery.rs` says this half is the host's,
because a handle names something in a workspace the server never sees. That is
true and it is not enough: **nothing on the wire tells the host what it already
sent.** `ServerFrame::RequeryRequest` carries the turn signal and no history of
answers, so a host that wants per-handle dedup has to keep its own map keyed by
turn id, and nothing says so.

## The options

1. **Say it in the contract.** `docs/spec/serve-surface.md` and the
   `requery_request` frame's own description state that the host owns
   per-handle suppression and must track what it sent, keyed by `turn_id`. No
   wire change. Cheapest, and it leaves every host to build the same map.
2. **Carry it on the frame.** `RequerySignal` grows an `already_sent:
   Vec<String>` the server accumulates from the handles a host declares on its
   answers, which means `RequeryResultIn` grows a `handles` field too. Additive
   both ways, and it makes the suppression a property of the protocol rather
   than of each host.

Either is defensible. What is not defensible is the silence there is now.

## Files

- `crates/stella-serve/src/remote/requery.rs` — the module header states the
  split; `RequeryState::produced` is the block-level set.
- `crates/stella-serve/src/frame.rs` — `RequerySignal`, `RequeryResultIn`.
- `crates/stella-cli/src/memory/steering.rs` — `ProducedSteering`, the CLI
  answer this one is measured against.
- `docs/spec/serve-surface.md` — the `requery-result` row.
- `docs/wire/` — regenerate with `bash scripts/export-agentevent-schema.sh` if
  option 2 wins; `make wire-schema` must stay green.

## Verify

Drive `crates/stella-serve/tests/bridge.rs::scripted_requery_turn` with a host
that answers two asks, the second a superset of the first, and assert what the
[...]

Issue #6217, opened 2026-09-06
A served turn re-asks for context and nothing on its event stream says what that cost


macanderson:
## Problem

A served turn can now re-ask the host for context at a step boundary
(`crates/stella-serve/src/remote/requery.rs`). Each ask is a recall fan-out on
the host's side with provider spend behind it, and **nothing puts that spend on
the turn's own event stream.**

The CLI's implementor of the same port does. `SessionRequery::requery`
(`crates/stella-cli/src/memory/steering.rs`) sends the recall's `ContextRecall`
`AgentEvent` before its own dedup can discard the block, so a re-query that
bought nothing is still visible as a re-query that was paid for. The comment
there says why: that discard is the unmeterable cost the event exists to
surface.

`RemoteRequery` sends no `AgentEvent` at all. A host reading `/v1/turns/{id}/events`
sees the `requery_request` frame go out and its own answer come back, and no
event says the turn spent anything on it.

Second, smaller half: `ReverseKind::Requery`
(`crates/stella-serve/src/observe/event.rs`) has its own
`reverse_dispatched_requery_total` counter, but the answer and timeout
counters (`reverse_answered_total`, `reverse_timed_out_total` in
`crates/stella-serve/src/observe/metrics.rs`) fold all three reverse kinds into
one number. So an operator cannot ask "how often does my host decline or time
out a re-query", which is the one question that says whether the feature is
earning its round trips. Provider and tool share that fold too, so splitting it
is a change to all three, not a bolt-on for this one.

## Files

- `crates/stella-serve/src/remote/requery.rs` — `RemoteRequery::requery`, where
  an event would be sent. It holds no `EventSender` today.
- `crates/stella-serve/src/session.rs` — `run_session` builds the adapter and
  already has `event_tx` in scope, which is where a sender would come from.
- `crates/stella-cli/src/memory/steering.rs` — `SessionRequery::with_events`
  and the send in `requery`, the shape to copy.
- `crates/stella-serve/src/observe/metrics.rs` — the shared answer/timeout
  counters.

## Constraints

- The host runs the recall, so the server does not know what the fan-out cost.
  Either `RequeryResultIn` grows an optional cost/usage field the host fills
  in, or the event reports only what the server can witness (that an ask was
  made and whether a block came back). Deciding which is the substance of this
  issue, and a new field on `RequeryResultIn` is a wire change: additive only,
  regenerate `docs/wire/` with `bash scripts/export-agentevent-schema.sh`, and
[...]

PR #6324, opened 2026-09-06
fix(stella-serve): document requery suppression, report re-query recall cost

### LINKED ISSUE #6303: serve: fix batch — 2 small defects (audit 2026-09-06)
Several small fixes ride one PR — the fix-over-file rule: batching same-area defects that are individually too small to justify a PR of their own keeps churn down without leaving them unfixed.

## Members

- [x] #6218 — A served re-query can show the model the same frames twice, and nothing says whose job it is to stop that — state, in `crates/stella-serve/src/remote/requery.rs`'s module header and `docs/spec/serve-surface.md`'s `requery_request` row, that the host owns per-handle suppression keyed by `turn_id` (option 1 from the issue — no wire change).
- [x] #6217 — A served turn re-asks for context and nothing on its event stream says what that cost — in `RemoteRequery::requery` (`crates/stella-serve/src/remote/requery.rs`), send the `ContextRecall` `AgentEvent` for an answered ask (mirror `SessionRequery::requery` in `crates/stella-cli/src/memory/steering.rs`), and record a written decision to defer the `ReverseKind` metrics-counter split (it touches all three reverse kinds, not just requery) to a separate issue.

## Definition of done

- [x] Every member's fix is in the batch PR with a witness test where the member is a defect (a test that fails on the old code), or the member is named under a "could not ride" heading with the case that stopped it.
- [x] Guards green.
- [x] No baseline entry raised.

---

Verified against #6324 at `205a12f`:

- **#6218** — the contract is stated in `requery.rs`'s module header ("the host owns
  per-handle suppression, keyed by the turn id") and in `serve-surface.md`'s
  `requery-result` row. Witness:
  `the_module_header_and_the_spec_both_say_who_suppresses_a_seen_handle`, which
  reads both files at compile time and asserts the same phrase appears in each,
  built from two halves so it cannot pass by matching its own source. Fails on
  `main` — neither file says this.
- **#6217** — `RemoteRequery` holds the turn's event sender and reports
  `AgentEvent::ContextRecall` when the host answers. Witnesses:
  `an_answered_requery_reports_its_recall_cost` and
  `a_declined_requery_reports_nothing` (both fail to compile on `main` — the
  constructor took one fewer argument), plus two assertions extending
  `tests/bridge.rs`'s scripted-turn tests through `Session::next_frame`.
  The deferred `ReverseKind` counter split is now filed as #6329 rather than
  left as an unfiled "separate issue".
- **Guards green** — every guard step in `ci` run
  [34059228544](https://github.com/macanderson/stella/actions/runs/34059228544)
  is green, `cargo fmt --check`, `cargo clippy -D warnings`, `cargo doc -D
  warnings` and `docs/wire/ still matches the types` included; `docs-guards`,
  `wire-schema`, `file-size`, `guard-self-tests`, `rebase-replay` and
  `msrv` all passed on this head.
- **No baseline entry raised** — `git diff origin/main...HEAD` touches no
  `scripts/*-baseline.txt`; the four changed files are `requery.rs`,
  `session.rs`, `tests/bridge.rs` and `serve-surface.md`.

macanderson:
## Summary

Fixes both members of batch #6303 (serve, 2 small defects).

### #6218 — nothing said whose job it is to suppress a re-shown frame

A served re-query can show the model the same frames twice: the wire
carries no per-request answer history, so nothing said who was supposed to
stop that.

**Fix**: option 1 from the issue — say it in the contract, no wire change.
`ServerFrame::RequeryRequest`'s own doc comment and
`docs/spec/serve-surface.md`'s `requery-result` row now state: the host
owns per-handle suppression, keyed by the turn id.

**Witness**: `the_frame_and_the_spec_both_say_who_suppresses_a_seen_handle`
in `crates/stella-serve/src/remote/requery.rs` reads both files at compile
time and asserts the same phrase appears in each, built from two halves so
it cannot pass by matching its own source. It fails on `main` — neither
file says this today.

### #6217 — a served re-query's cost never reached the event stream

`RemoteRequery::requery` answered a host's re-query with no `AgentEvent` at
all, so an operator reading `/v1/turns/{id}/events` could not see what a
re-query cost, unlike the CLI's `SessionRequery::requery`.

**Fix**: `RemoteRequery` now holds the turn's event sender and reports an
`AgentEvent::ContextRecall` the moment the host answers — before the
block-level dedup can discard the text, mirroring `SessionRequery::requery`'s
own reasoning for why that ordering matters. The server never sees the
host's own frame-level recall (which providers ran, which frames won
fusion), so the event carries one synthetic frame labelled as a host
answer rather than invented per-frame provenance.

The issue's second, smaller ask — splitting `ReverseKind`'s answered/timeout
counters — is **deferred to a separate issue**: it touches the provider and
tool reverse-request counters too, not just requery's, per the issue's own
constraints section.

**Witness**: two new unit tests in `requery.rs`
(`an_answered_requery_reports_its_recall_cost`,
`a_declined_requery_reports_nothing`) construct a `RemoteRequery` directly,
script a host answering or declining, and assert the events channel does
or does not carry a `ContextRecall`. Both fail to compile on `main` — the
constructor takes one fewer argument. Two new assertions extend
`crates/stella-serve/tests/bridge.rs`'s existing end-to-end scripted-turn
tests (`a_served_turn_requeries_the_steering_plane_after_its_work_moves_to_a_new_path`,
`a_host_that_declines_a_requery_injects_nothing`) through `Session::next_frame`.

## Could not ride

Nothing — both members ride this PR.

## Tests deleted

None.

Closes #6303

## Summary by Sourcery

Make served re-queries observable and explicitly assign hosts responsibility for suppressing repeated frames.

Bug Fixes:
- Report the token cost of answered served re-queries as `ContextRecall` events on the turn event stream.
- Clarify that hosts are responsible for suppressing repeated frame handles per turn during re-query handling.

Enhancements:
- Represent served re-query recall with synthetic host-answer telemetry while avoiding unobserved provider or frame provenance.

Documentation:
- Document host ownership of per-handle re-query suppression in the served API contract.

Tests:
- Add unit and end-to-end coverage for recall events from answered re-queries and the absence of events for declined re-queries.
- Add a contract test ensuring the implementation documentation and serve-surface specification identify the same suppression responsibility.


macanderson:
The body and the commit trailer both defer the `ReverseKind` counter split "to a
separate issue", and no such issue existed — the decision was recorded and the
work was not tracked, which is the half SCR-004 asks for. Filed as #6329, with
the reason it could not ride this PR (it changes the provider and tool counters
too, not just requery's).

One thing #6329 adds that #6217 did not name: `reverse_wait_ms_total` is folded
across kinds as well, and it is the numerator of mean reverse latency — folded,
it averages a provider call against a re-query the turn survives unanswered. It
wants splitting with the other two or the per-kind answer counts have no latency
to pair with.

Also worth a note for whoever reads this later: the body names the #6218 witness
as `the_frame_and_the_spec_both_say_who_suppresses_a_seen_handle`, but the test
that landed is `the_module_header_and_the_spec_both_say_who_suppresses_a_seen_handle`
— the contract moved from `ServerFrame::RequeryRequest`'s doc comment to the
module header in `50c7bf8` (that doc comment feeds the committed
`serveframe.schema.json`, so stating it there reddened `wire-schema`). The code
matches what #6303 asked for; only the description is stale.

Verified against this PR's head: #6303's three DoD items hold — both members
ride with witness tests, no baseline entry is raised, and every guard step in
the `ci` run is green.

---

macanderson:
## The `ci` run on `205a12f` did not pass — it was killed at the job ceiling

`cargo test` ran from 21:02:10 to 21:58:30 and every step after it is `skipped`
([run 34059228544](https://github.com/macanderson/stella/actions/runs/34059228544)).
GitHub reports a `timeout-minutes` expiry as `cancelled`, so the run reads
"cancelled" rather than "failed" — but nothing was superseded and no newer run
replaced it. Pushed `69e4767` with the fix.

### Root cause

The first commit gave `RemoteRequery` an `UnboundedSender<AgentEvent>`, which
made the port a **fifth holder of an `event_tx` clone**. `run_session`'s
teardown releases four and never released this one. The comment directly above
those drops already states what that costs:

> Release everything holding a clone of the event sender BEFORE closing the
> channel… **Miss one and `event_rx.recv()` never returns `None`, the forwarder
> never drains, and a turn that has already produced its `Complete` event hangs
> without ever emitting its terminal frame.**

`drop(engine)` does not take the port along — the engine only borrows it
(`requery.as_ref().map(|r| r as &dyn SteeringRequery)`), which is why the
binding is hoisted above the engine in the first place. So `forwarder.await`
blocks forever, `Session::next_frame` never ends, and every bridge test that
drives a turn with `steering_requery: true` hangs.

### The fix

One `drop(requery)` between `drop(dispatcher)` and `drop(event_tx)` — after the
engine that borrows it, before the channel closes.

### Demonstrated, not inferred

```
cargo test -p stella-serve --test bridge requery   # this branch @205a12f
  → killed by `timeout 120`, no output (exit 143)

cargo test -p stella-serve --test bridge requery   # origin/main's stella-serve
  → 2 passed; finished in 0.00s

cargo test -p stella-serve --test bridge           # 205a12f + drop(requery)
  → 14 passed; 0 failed; finished in 1.81s
```

`cargo fmt --check -p stella-serve` is clean. The control that rules out the
infrastructure: #6323's `cargo test` finished in **8m33s** on an adjacent runner
minutes later, so the 56 minutes here is this branch, not the runners.

Worth noting for the ratchet this repo keeps on re-runs: this was **not** a
flake, and a re-run would have burned another hour and died the same way. The
run died *during* test bodies, which is the case the "re-run only if it died
before any test body ran" rule excludes.

---

Issue #2941, opened 2026-08-12
pipeline: a killed run never delivers — adoption runs once, at the end, so a wall-clock kill discards the whole candidate


macanderson:
## Problem

Candidate delivery happens exactly once, at the end of the run, after the
verdict. If the process dies before it — the harness's wall-clock kill, a
SIGTERM, a crash — the candidate workspace is never adopted and every byte the
worker wrote stays in `/tmp/stella_candidate_*`, which the harness then throws
away with the container.

This is the half of #2927 that PR #2943 does **not** fix. #2943 removed the
`verdict.passed` gate, so a failed or absent verdict now delivers; it cannot
help a run whose process is not alive to run the adoption at all.

Measured, ArenaBench run `w10p` (SUT `main@554bb4f`): six of the twenty
pipeline trials ended with no `verdict` event at all — `build-cython-ext`
(x2), `code-from-image`, `query-optimize` (x2), `video-processing`. All six
end mid-stream (truncated final line): the harness killed the process at the
wall-clock cap. `build-cython-ext__4xjNAy9`'s last event, at 908s of a 912s
run, is `18 passed in 1.44s` — finished work, reward 0, because delivery never
ran.

## Files

- `crates/stella-pipeline/src/pipeline.rs` — the delivery loop in
  `run_best_of_n` (search for `delivery::decide`); the single adoption point.
- `crates/stella-pipeline/src/pipeline/delivery.rs` — the delivery decision
  and the argument behind it.
- `crates/stella-cli/src/candidate_ws/adopt.rs` — the adoption mechanism
  (`git diff --binary --no-renames` + `git apply`). It works; it is simply not
  invoked in this scenario.

## Candidate designs (both need deciding, not just implementing)

1. **Deadline-driven delivery.** The run already knows its wall-clock budget
   (`PipelineConfig`'s deadline plumbing). Adopt when the deadline is within a
   margin, then continue if there is time left. Cheap, but it delivers a
   mid-flight tree, and a later adoption has to be idempotent against the
   already-delivered one.
2. **Incremental delivery.** Adopt after each stage boundary or each N tool
   calls. Strictly more robust and strictly more expensive (an adoption is two
   git invocations over the whole diff), and it weakens the isolation contract
   the escape check (#1538) depends on — the real tree would legitimately hold
   the candidate's sealed bytes.
3. **A signal handler.** Adopt on SIGTERM. Does nothing for SIGKILL and
   nothing for a hard timeout that does not signal, and running git from a
[...]

Issue #2927, opened 2026-08-12
pipeline: candidate adoption is winner-only — a revise verdict or a killed run discards the whole candidate, and 8 of 15 pipeline losses never had their work delivered


macanderson:
> **Corrected 2026-08-11.** The first version of this issue claimed the candidate
> delivery was lossy and that `financial-document-processor` produced
> byte-identical output to the bare winner. **Both claims were wrong** and are
> retracted below. Adoption works; the defect is *when* it runs. The original
> error came from comparing the values in the worker's Python source instead of
> the delivered CSV bytes — the exact "read the thing the claim is about" failure
> AGENTS.md warns about. Keeping the retraction visible on purpose.

## Problem

Candidate adoption is **winner-only**. The candidate workspace is delivered to the
graded tree only when the run reaches a *passing* verdict. A run that ends with
`passed=false` (the `revise` rung) or that never reaches a verdict at all has its
entire candidate discarded, however much correct work it contains.

On a 60-trial ArenaBench run (SUT `main@554bb4f`, 10 tasks x 2 trials x 3 arms)
the pipeline solved **5/20** against the bare loop's **14/20**. **8 of the 15
pipeline losses never had their work delivered.**

## The delivery mechanism (established, not assumed)

`git show 554bb4f77:crates/stella-cli/src/candidate_ws/adopt.rs` — adoption is
`git apply` of a `git diff --binary --no-renames` patch:

```rust
let mut patch_args = vec!["diff", "--binary", "--no-renames", self.baseline.as_str(), sealed.as_str()];
...
let apply = SystemGitCli.run(&self.toplevel, &["apply", "--whitespace=nowarn", patch_str]);
```

So binaries ride as git binary patches and a `mv` becomes `D old` + `A new`.
Neither binaries nor deletions are dropped — the lossy-delivery hypothesis is
**refuted at the source**.

It is also refuted empirically. Adoption returns its `FileChange` list only on
success and runs *after* the verdict, so a post-verdict `file_change` burst is
positive proof the patch landed. `financial-document-processor__LV4JyLJ`:

```
1786495105066  *** VERDICT passed=True rung=submit_fast
1786495105414  file_change  deleted  documents/2lgKzDuI4E4g.jpg
   ... all 17 deletions + 18 creations, 348 ms after the verdict ...
1786495105418  stage complete
```

`sqlite-with-gcov__cYm44MQ` adopted **2393 paths** including `.so`, `.a`, `.o`,
[...]

PR #6326, opened 2026-09-06
fix(bench): guard trace_triage's site strings and add a ledger posture

### LINKED ISSUE #2972: bench/trace_triage: a file rename silently repartitions every defect fingerprint, and the recovery marker carries the old digest
`bench/trace_triage/fingerprint.py` builds a defect fingerprint from
`(detector, site, normalized variant)`. The `site` component is a repo-relative
path declared on the detector — e.g. `crates/stella-pipeline/src/verify.rs` —
and **nothing checks that the path still exists**.

That string is prose for a reviewer today, exactly like a `ConsumerPosture`
row's `site` in `crates/stella-protocol/src/event/consumers.rs`. Here it is
worse than prose, because it is *load-bearing*: it is hashed into the
fingerprint. So a file move silently repartitions the identity space.

## What breaks

Rename `crates/stella-pipeline/src/verify.rs`, update the detector's `site=`
to match, and every fingerprint that used it changes digest. The ledger
(`bench/trace_triage/fingerprints.json`) still holds the old digests, so:

- the next triage run finds no ledger entry for a defect it has seen twenty
  times;
- the GitHub search may or may not surface the existing issue;
- if it does not, the tool opens a **duplicate** — the single outcome the whole
  design exists to prevent.

The fingerprint-marker gate (gate 2 of 3) does not save this: the marker in the
existing issue body carries the *old* digest, so it does not match the new one.

Two failure directions, and the quiet one is worse: leaving `site=` stale after
a move keeps the digests stable but makes the issue text point a fresh agent at
a path that no longer exists.

## Files

- `bench/trace_triage/detectors.py` — the eight `@detector(site=...)`
  declarations
- `bench/trace_triage/fingerprint.py` — `Fingerprint.digest` hashes the site
- `bench/trace_triage/fingerprints.json` — the persisted mapping that breaks
- `scripts/check-module-reachability.py` is the nearest existing shape of guard

## Reproduce

```sh
git mv crates/stella-pipeline/src/verify.rs crates/stella-pipeline/src/ladder.rs
# edit detectors.py's site= strings to match
python3 bench/trace_triage/triage_bench_traces.py --run w10p --mirror <mirror> --offline
```

Every `witness-unavailable` and `oracle-flip-ungraded` finding now prints a
fingerprint absent from the ledger, and would be planned as `NEW ISSUE`.

## Constraints

- A guard belongs with the other toolchain-free guards, but `make gate` must
  stay offline and side-effect-free — a path-existence check is fine there; a
  GitHub call is not. Do **not** add `triage-bench-traces` itself to
  `GATE_STEPS` (`gate-parity` enforces this).
- Whatever is chosen must not make an ordinary refactor unlandable. A
  fingerprint that changes on a rename is arguably correct-but-inconvenient; a
  fingerprint that silently duplicates is not.

## Definition of done

Pick one and implement it, with a witness test:

1. **Assert the paths exist.** A test in `bench/trace_triage/tests/` that walks
   `DETECTORS` and fails when a non-empty `site` names no file in the tree.
   Cheap, catches the stale-string direction, does nothing for the rename
   direction.
2. **Make the ledger survive a rename.** Record the readable triple beside each
   digest (it already is) and, on a ledger miss, re-hash every entry against the
   *current* detector table before declaring a finding new — rebinding the issue
   number and rewriting the digest as a reviewable diff.
3. **Take the site out of the hash** and keep it as metadata only, accepting
   that the detector code plus the variant must then carry the whole
   discrimination. Simplest, and the one to argue against explicitly if it is
   rejected.

Option 2 is the one that actually removes the duplicate hazard; 1 is worth
having regardless and is ten lines.

Found while building #2971 (`make triage-bench-traces`).

### LINKED ISSUE #2974: bench/trace_triage: no ledger posture for an accepted defect, and the --fetch include-list can silently drop a file the loader reads
Two gaps in `make triage-bench-traces` (#2971), both of which make the tool
noisier over time rather than quieter.

## 1. A fingerprint cannot be marked "known and accepted"

`bench/trace_triage/fingerprints.json` maps a fingerprint to an issue number.
There is no state for "this is real, it is understood, and it should stop being
proposed as a new issue" — a `wontfix`, an expected-by-design shape, or a
finding a maintainer has read and declined.

Today the only way to silence one is to bind it to an issue, which is a lie if
no issue describes it, or to leave it re-proposed as `NEW ISSUE` on every run
forever. The second is what will actually happen, and it degrades the plan into
something a human skims — at which point the tool has failed at the one job that
requires human attention.

`LedgerEntry.issue` is already `int | None` and `source` already distinguishes a
human binding from an automatic one, so the shape is there; what is missing is a
posture (`tracked` / `accepted` / `wontfix`) that `plan_actions` reads.

**Files:** `bench/trace_triage/fingerprint.py` (`LedgerEntry`, `Ledger.bind`),
`bench/trace_triage/issues.py` (`plan_actions`, `Decision`).

**Done when:** a ledger entry with the new posture produces a distinct decision
in the plan — printed, counted in the summary, and never filed — with a test
that a suppressed-by-posture finding opens nothing; and `README.md § Editing the
ledger` documents it.

## 2. `--fetch` is untested

`fetch_run` in `bench/trace_triage/triage_bench_traces.py` shells out to
`aws s3 sync` with an include-list that deliberately excludes `stella-run.json`
and `stella-run.stdout.txt` (each can exceed 70 MB per trial and both are
re-renderings of the event stream).

No test covers it, because no test in this package may reach S3 — which is the
right constraint and leaves a real gap: if the include-list ever drops
`*reward.txt`, every trial silently loads with `reward = None`, every
`oracle-flip-ungraded` finding disappears, and the run reads as healthy.

**Fix shape:** extract the argv construction from the subprocess call and test
the argv — that the include-list covers every file `run_trace.load_run` reads,
derived from a single shared constant rather than two lists that can drift.
`bench/harbor_adapter`'s manifest-parity tests are the local precedent for
deriving a check from the tree instead of a hand-written list.

**Done when:** a test fails if `load_run` gains a read of a file the sync would
not fetch.

Refs #2971

## Definition of done

- [x] Ledger posture (`tracked`/`accepted`/`wontfix`) added to `LedgerEntry`; a silenced entry produces a distinct `SUPPRESSED (ledger posture)` decision in `plan_actions`, printed, counted in the summary, never filed, recorded on `--apply`; `README.md § Editing the ledger` documents it.
- [x] `--fetch`'s include-list (`triage_bench_traces._INCLUDES`) is generated from `run_trace.READ_FILENAMES`, which `load_run`'s own reads are built from (named constants, not a second spelling); a test fails if a read in `load_run` ever outruns the include-list.

Closed by #6326.

### LINKED ISSUE #6300: bench: fix batch — 2 small defects (audit 2026-09-06)
The fix-over-file rule: several small fixes ride one PR rather than each opening its own.

## Members

- [x] #2972 — bench/trace_triage: a file rename silently repartitions every defect fingerprint, and the recovery marker carries the old digest — fixed in #6326: `test_every_declared_site_names_a_file_that_exists` walks `DETECTORS` and fails on a non-empty `site` naming no file in the tree (fails on `main` today — five stale sites, four from the deleted `stella-pipeline` crate and one from the retired `grep` tool). The five sites are corrected and the four ledger entries already bound to real issues (#2941, #2927, #2908 x2) are re-keyed to the new digest.
- [x] #2974 — bench/trace_triage: no ledger posture for an accepted defect, and the --fetch include-list can silently drop a file the loader reads — fixed in #6326: `LedgerPosture` (`tracked`/`accepted`/`wontfix`) added to `LedgerEntry`, `plan_actions` returns a distinct `SUPPRESSED (ledger posture)` decision before any search, `apply_actions` records without filing. `--fetch`'s include-list is generated from `run_trace.READ_FILENAMES`, which `load_run`'s own reads are built from (named constants, not a second spelling) — an AST-scan test fails if a raw filename literal ever outruns it.

## Definition of done

- [x] Every member's fix is in the batch PR with a witness test where the member is a defect (a test that fails on the old code), or the member is named under a "could not ride" heading with the case that stopped it
- [x] Guards green (`make guards-fast` local; CI: https://github.com/macanderson/stella/pull/6326)
- [x] No baseline entry raised

Closed by #6326.

macanderson:
Batch PR for #6300 (bench, 2 members). Both members touch `bench/trace_triage`'s fingerprint machinery, so they ride together.

## #2972 — a file rename silently repartitions every defect fingerprint

`site` is hashed into a defect's fingerprint, but nothing checked that it still names a real file. It didn't, right now, on `main`: `crates/stella-pipeline` was deleted in #3865, and four detector declarations (`witness-unavailable` x2, `oracle-flip-ungraded` x2) still pointed at `crates/stella-pipeline/src/verify.rs`. A fifth (`grep-ere-false-negative`) pointed at `crates/stella-tools/src/grep/fallback.rs`, a standalone `grep` tool that no longer exists — it was folded into the unified `search` tool.

**What/why:** implemented option 1 from the issue (the cheap half): `test_every_declared_site_names_a_file_that_exists` walks `DETECTORS` and fails on a non-empty `site` that names no file in the tree.

**Witness:** the test fails on `main` today — proven with `git stash` locally — because the five stale sites above are real, not hypothetical. Fixed by:
- The four `stella-pipeline` sites → `site=""`. That logic now lives in a private verification plugin (Vera), not in this repo, so nothing here is implicated — the detector's own documented convention for "nothing to implicate."
- `grep-ere-false-negative` → `site=""`, same reasoning (no file in the current tree owns a POSIX-fallback shape).
- `no-post-verdict-file-change` → `crates/stella-cli/src/candidate_workspaces.rs`, its real successor (`candidate_ws/adopt.rs` → `candidate_workspaces.rs`, same `adopt()` function).

Since `site` is hashed into the fingerprint, changing it changes the digest. The ledger (`fingerprints.json`) already had 4 entries bound to real issues (#2941, #2927, #2908 x2) under the old digests — those are re-keyed to the new digests (site changed, everything else identical) so the same issue numbers still bind on the next `--fetch`, instead of silently duplicating.

## #2974 — no ledger posture for an accepted defect, and an untested `--fetch` include-list

**Part 1 — posture.** `LedgerEntry` had no way to say "this finding is real, understood, and should stop being proposed" without lying about an issue number. `LedgerPosture` (`tracked` / `accepted` / `wontfix`) fixes that. A silenced entry now produces a distinct `SUPPRESSED (ledger posture)` decision in `plan_actions` — before any search, before any client is touched — and `apply_actions` records the occurrence without filing anything. `README.md`'s "Editing the ledger" documents it.

**Witness:** `test_an_accepted_posture_opens_nothing_and_never_even_searches`, `test_a_wontfix_posture_needs_no_issue_number`, and `test_posture_suppressed_is_counted_in_the_summary_and_never_filed` all fail with `ImportError: cannot import name 'LedgerPosture'` against `git stash` of `fingerprint.py`/`issues.py`, and pass with them.

**Part 2 — the `--fetch` include-list.** It was a second, hand-maintained copy of what `run_trace.load_run` actually reads. `run_trace.READ_FILENAMES` is now the single source of truth; `_INCLUDES` is generated from it. `test_fetch_includes_every_file_load_run_reads` fails if a read in `load_run` ever outruns what the mirror fetches, and a negative-control sibling (`test_missing_includes_actually_catches_a_dropped_pattern`) proves the check has teeth rather than passing vacuously.

**Witness:** fails with `ImportError: cannot import name 'READ_FILENAMES'` against `git stash`, passes with the change.

## Found while touching the same class

`Ledger.save()` named only the three top-level keys it manages (`schema`, `updated_at`, `fingerprints`), so any `--apply` run against the real, tracked `fingerprints.json` silently deleted its hand-written `_comment` block the moment it wrote the file — verified directly against the committed file before this PR (`_comment` present before `save()`, gone after). `save()` now round-trips every top-level key it doesn't itself own. Witness: `test_save_never_drops_the_hand_written_comment_block`.

## Could not ride

None — both members' full scope shipped in this PR.

## Verification

```
uv run --with pytest --no-project pytest -q bench/trace_triage/tests   # 102 passed
make guards-fast                                                       # green, including check-prose
```

No Rust files touched, so `cargo fmt --all` is a no-op.

Tests deleted: None

Closes #2972
Closes #2974
Closes #6300

## Summary by Sourcery

Harden trace triage fingerprint and ledger management by validating detector sites, adding explicit suppression postures, and keeping persisted metadata and fetched trace inputs consistent.

New Features:
- Add ledger postures for suppressing accepted or wont-fix findings without requiring an issue number.
- Validate detector sites against files in the repository and centralize trace filenames used by loading and fetch operations.

Bug Fixes:
- Prevent stale detector sites from silently changing fingerprints.
- Preserve hand-maintained top-level ledger metadata when saving fingerprints.
- Keep trace mirror fetches synchronized with the files consumed by run loading.

Enhancements:
- Record and report posture-suppressed occurrences without searching, commenting, or filing issues.
- Maintain backward compatibility for ledger entries without posture data.

Documentation:
- Document ledger posture management and suppression behavior.

Tests:
- Add coverage for ledger postures, metadata preservation, site validity, and fetch include synchronization.

Chores:
- Re-key affected fingerprint ledger entries after correcting detector sites.

### DIFF
--- a/bench/trace_triage/README.md
+++ b/bench/trace_triage/README.md
@@ -87,6 +87,16 @@ issue never named, a higher rate — instead of restating the issue. When an
 occurrence adds nothing, nothing is posted; the run is still recorded so the
 next comparison is against the truth.
 
+`posture` says whether a fingerprint should still be *proposed* as a new
+issue. `tracked` is the default: search, then comment or open one. `accepted`
+and `wontfix` mean a human read this finding. It is real. It should stop
+coming back. Neither needs a real `issue` number — a fake one would be a lie.
+
+Set `posture` by hand (add an entry with `issue: null` if none exists) and
+every run after that prints a `SUPPRESSED (ledger posture)` decision. It
+never searches, comments, or opens an issue, but it still records the run.
+Delete `posture` later and the record is still true.
+
 ## Detectors
 
 | Code | Fires on |
--- a/bench/trace_triage/detectors.py
+++ b/bench/trace_triage/detectors.py
@@ -18,7 +18,7 @@ keyed on them.
     @detector(
         code="my-shape",                       # stable; part of the fingerprint
         title="what a new issue would be called",
-        site="crates/stella-pipeline/src/verify.rs",   # or "" if none implicated
+        site="crates/stella-model/src/provider_parity.rs",   # or "" if none implicated
         search_terms=("phrase a reviewer would search",),
     )
     def _my_shape(run: Run) -> list[Finding]:
@@ -341,7 +341,7 @@ def _agent_timeout(run: Run) -> list[Finding]:
 @detector(
     code="witness-unavailable",
     title="witness authoring failed",
-    site="crates/stella-pipeline/src/verify.rs",
+    site="",
     search_terms=("witness_unavailable", "witness author", "no TEST_COMMAND"),
 )
 def _witness_unavailable(run: Run) -> list[Finding]:
@@ -352,6 +352,11 @@ def _witness_unavailable(run: Run) -> list[Finding]:
     failure, "emitted a test command but created no file" is an authoring
     failure, and "turn aborted: stuck-loop detected" is a loop defect. Filing
     them under one ticket buries two of the three.
+
+    `site` is `""` rather than `crates/stella-pipeline/src/verify.rs`: that
+    crate is deleted, and witness authoring is now a verification plugin's
+    own logic (Vera, private, not shipped here), so no file in this tree is
+    implicated.
     """
     from fingerprint import normalize
 
@@ -389,7 +394,7 @@ def _witness_unavailable(run: Run) -> list[Finding]:
         findings.append(
             Finding(
                 detector="witness-unavailable",
-                site="crates/stella-pipeline/src/verify.rs",
+                site="",
                 variant_source=reason_full,
                 title=f"witness authoring failed: {reason_full[:90]}",
                 summary=(
@@ -414,7 +419,7 @@ def _witness_unavailable(run: Run) -> list[Finding]:
 @detector(
     code="oracle-flip-ungraded",
     title="the oracle recorded a baseline->candidate flip on a trial the grader scored 0",
-    site="crates/stella-pipeline/src/verify.rs",
+    site="",
     search_terms=("oracle flip reward 0", "flip achieved grader rejected"),
 )
 def _oracle_flip_ungraded(run: Run) -> list[Finding]:
@@ -426,6 +431,11 @@ def _oracle_flip_ungraded(run: Run) -> list[Finding]:
     does not accept — either the witness is weaker than the task, or the proven
     work never reached the graded tree. Both are defects, and the trace cannot
     tell them apart on its own, which is stated rather than guessed.
+
+    `site` is `""` rather than `crates/stella-pipeline/src/verify.rs`: that
+    crate is deleted, and the oracle's flip logic is now a verification
+    plugin's own logic (Vera, private, not shipped here), so no file in this
+    tree is implicated.
     """
     if run.bare_loop:
         return []  # no oracle exists on the bare loop; see below.
@@ -457,7 +467,7 @@ def _oracle_flip_ungraded(run: Run) -> list[Finding]:
     return [
         Finding(
             detector="oracle-flip-ungraded",
-            site="crates/stella-pipeline/src/verify.rs",
+            site="",
             variant_source="oracle observed baseline fail and candidate pass on a trial graded 0",
             title="the oracle recorded a baseline->candidate flip on a trial the grader scored 0",
             summary=(
@@ -783,7 +793,7 @@ def _role_model_mismatch(run: Run) -> list[Finding]:
 @detector(
     code="no-post-verdict-file-change",
     title="a trial recorded no file_change after its verdict",
-    site="crates/stella-cli/src/candidate_ws/adopt.rs",
+    site="crates/stella-cli/src/candidate_workspaces.rs",
     search_terms=("candidate adoption post-verdict file_change", "delivery emits no event"),
 )
 def _no_post_verdict_file_change(run: Run) -> list[Finding]:
@@ -861,7 +871,7 @@ def _no_post_verdict_file_change(run: Run) -> list[Finding]:
         findings.append(
             Finding(
                 detector="no-post-verdict-file-change",
-                site="crates/stella-cli/src/candidate_ws/adopt.rs",
+                site="crates/stella-cli/src/candidate_workspaces.rs",
                 variant_source="trial recorded no verdict event at all",
                 title="a trial ended with no verdict event — the delivery point was never reached",
                 summary=(
@@ -879,7 +889,7 @@ def _no_post_verdict_file_change(run: Run) -> list[Finding]:
         findings.append(
             Finding(
                 detector="no-post-verdict-file-change",
-                site="crates/stella-cli/src/candidate_ws/adopt.rs",
+                site="crates/stella-cli/src/candidate_workspaces.rs",
                 variant_source="verdict recorded and no mutating file_change followed it",
                 title="a trial recorded a verdict and no file_change after it",
                 summary=(
@@ -1082,7 +1092,7 @@ _FALLBACK_DISCLOSURE = "searched with POSIX"
 @detector(
     code="grep-ere-false-negative",
     title="grep answered `(no matches)` for a pattern the POSIX fallback could not execute",
-    site="crates/stella-tools/src/grep/fallback.rs",
+    site="",
     search_terms=("grep no matches alternation", "POSIX grep BRE fallback false negative"),
 )
 def _grep_ere_false_negative(run: Run) -> list[Finding]:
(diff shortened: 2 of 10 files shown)


macanderson:
@sourcery-ai review

macanderson:
Re: Sourcery's ❌ on #2974's second objective ("make the \`--fetch\` include-list testable... with a test that detects future divergence") — Sourcery's review budget is exhausted for the next ~24 minutes so it can't re-review the fix directly. For the record:

Fixed in c656d2323 (\`fix(bench): close the fetch-includes gap Sourcery flagged on #6326\`). \`run_trace.load_run\`'s five reads (\`results.json\`, \`result.json\`, \`reward.txt\`, \`exception.txt\`, \`stella-events.jsonl\`) now route through named constants (\`_RESULTS_JSON\`, ...) that \`READ_FILENAMES\` is built from — one spelling per filename, not two. \`test_every_filename_literal_in_run_trace_is_declared\` AST-scans \`run_trace.py\`'s own source for any string shaped like a filename and fails if one sits outside \`READ_FILENAMES\`, so a new read added as a raw literal (bypassing the shared constants) fails locally instead of silently outrunning \`--fetch\`. \`test_the_filename_scan_actually_catches_a_stray_literal\` proves the scan has teeth against a synthetic stray literal.

All CI checks are green (\`fmt + clippy + test\`, \`Analyze (rust)\`/CodeQL, \`dod / dod\`, guard self-tests, etc.).

Issue #6005, opened 2026-09-05
dim is held to WCAG's large-text floor but paints real UI words at ~75 call sites


macanderson:
## Problem

`scripts/check-contrast.py` holds `dim on bg` to the 3.0:1 "large text /
non-text UI component" floor (WCAG 1.4.11), not the 4.5:1 normal-text floor
(1.4.3). #4063 re-cut `dim` to clear 3.0:1 (now 3.14:1), which is correct
*for that classification* — but a prior investigation on #4063 found the
classification itself does not match how `dim` actually ships:

- `theme::TEXT_DIM` (the alias `crates/stella-tui/src/theme.rs` and
  `crates/stella-tui/src/palette.rs` declare) genuinely is chrome-only —
  three production references, all non-word.
- `stella_tui_theme::token::DIM` — the value itself — is reached directly at
  roughly 75 paint sites across 28 files in `crates/stella-tui/src/views/`,
  and most of them are words a user must read: keybinding verbs
  (`views/issues_tab.rs`), legend pairs (`views/files_tab.rs`), a live chord
  label (`views/frame.rs`), the status bar's own `"? help"` (whose doc
  comment calls it "the one thing on this row that is an instruction rather
  than a fact"), code captions, and more.

Real UI text set at terminal character size is not WCAG "large text" (>=24px
or >=18.66px bold) by any reasonable reading, so the 3.0:1 floor these ~75
sites are actually held to is the wrong law for what they paint — regardless
of what `dim`'s hex value is. #4063 raised the value to clear the floor it
already had; it did not ask whether that is the floor these call sites
should have.

`design/tui-v2/SPEC.md` §13 ("nothing dimmer than `dim` may carry required
information") and §3.1's palette table (`dim`'s role: "hints, keybinding
rows, line numbers" — three categories of required information) already say
opposite things about this, and the code sides with §3.1.

## Options (from the #4063 thread, none is a foregone conclusion)

1. **Reassign the ~75 call sites to `muted`** (4.79:1, clears the normal-text
   floor), leaving `dim` for genuine chrome only (the progress groove,
   `FALLBACKS`, the `"----"` placeholder in `fleet_dashboard.rs` —
   `theme::TEXT_DIM`'s three real references). This is the literal reading
   of `theme::TEXT_DIM`'s own doc comment, already true of the *alias*;
   it just isn't true of the raw token today.
2. **Accept dim-sized UI text at the 3.0:1 floor** as a stated, terminal-only
[...]

Issue #5933, opened 2026-09-05
A deck only checks for a checkout peer at start, so the session with work to lose is never warned


macanderson:
## The problem

`stella`'s deck prints one line at start when another live session already
holds the same checkout (`crates/stella-cli/src/command_deck/shared_checkout.rs`,
ADR 0027). It reads the session registry once, in `CommandDeck` startup, and
never looks again.

So the warning reaches the session that arrives second and never reaches the
session that was there first. That is the wrong way round for the loss it is
about: the first session is the one holding uncommitted work, and the second
one's `git checkout` is what reverts it. The session with something to lose is
the only one told nothing.

## Where it lives

- `crates/stella-cli/src/command_deck/shared_checkout.rs` — `peers()` and the
  announce call; both run once.
- `crates/stella-cli/src/command_deck.rs` — the single call site, in start-up.
- `crates/stella-store`'s `SessionRegistry` — already holds every live
  session's pid and workspace, and already reads a dead pid as a crash, so the
  data for a later look is there.

## How to verify

Start two decks in one checkout. The second prints the peer line today; the
first prints nothing, before or after the second arrives. A fix makes the first
one say so too, at whatever point it next redraws.

## Constraints

- It must never refuse, for the reason the module header gives: a session told
  to stop has nowhere to go, since the working directory is picked at launch.
- `command_deck.rs` sits at its size ceiling (AGENTS.md § God files), so new
  logic belongs in the `command_deck/` submodule that already exists for this.
- A poll on every frame would hit SQLite on every redraw. Read the registry on
  a bounded interval, or off the event that already refreshes the SESSIONS
  view.

## Definition of done

- [ ] A deck already running announces a peer that arrives in its checkout
      after it started.
- [ ] The notice still never refuses or blocks the session.
- [ ] A test covers the late-arrival direction: one record present at start,
      a second added, and the announcement fires for the first session.
- [ ] The registry is not read on every frame.

macanderson:
Folded into batch #6299 (2026-09-06 audit): fixed together by crate.

PR #6342, opened 2026-09-07
fix(stella-tui): tui batch — 11 small defects

### LINKED ISSUE #6299: tui: fix batch — 11 small defects (audit 2026-09-06)
Several small `area:tui` defects rated low/medium value by the 2026-09-06 backlog audit, batched under the fix-over-file rule: each is small enough that riding one PR beats opening eleven.

PR #6342 carries the batch. All 11 members ride it; nothing was deferred.

## Members

- [x] #5748 — note_kind still mutes Error and SteeringWithheld to the same glyph as a budget tick. Fixed: `NoteKind::Alert` added to `stella-transcript::model`, rendered red by `grid.rs` and `html/transcript.css`, and both `stella-tui::transcript_build::note_kind` and `stella-cli`'s own copy in `export/transcript.rs` classify onto it. Witness `error_and_steering_withheld_classify_louder_than_other`.
- [x] #5183 — The MCP tab's latency column is the connect-time round trip and never refreshes. Taken the issue's second branch: the doc comment now says which of the two figures `ServerHealth::latency` reports, instead of claiming a live measurement. Prose-only, so no witness — live re-measurement is a larger change and is not attempted here.
- [x] #5176 — MCP registry 'signed' attests the publisher's namespace, not the package bytes. Fixed: the displayed state is `attributed` in `stella-mcp::registry`, `stella-tui`'s mcp tab, and the website docs. Witness `an_attributed_entry_does_not_render_as_signed`.
- [x] #1740 — Files tab pairs a cumulative footer with a single-mutation diff body and says nothing about the mismatch. Fixed by option 1: the pane carries a `latest of N changes` label plus the latest mutation's own delta. Witness `a_footer_summing_several_mutations_labels_which_one_the_body_shows`.
- [x] #5933 — A deck only checks for a checkout peer at start, so the first session is never warned. Fixed: `shared_checkout::spawn_late_arrival_monitor` polls the registry on a bounded 5-second interval, diffs the peer set, and announces new arrivals through the deck's own channel. Witnesses `newly_arrived_names_only_the_peer_the_last_check_did_not_see`, `newly_arrived_with_no_prior_check_reports_everyone_live`.
- [x] #5651 — The transcript's standalone gate row can never be drawn, and its price text is written twice. Fixed: `EventKind::Gate` deleted; `views/gate_board.rs` is the sole renderer; SPEC.md §6.3/§8.1 updated to match. Witness `the_deterministic_gate_price_is_built_in_exactly_one_place`.
- [x] #2802 — Surface `HealthState::AuthRequired` in the deck's MCP tab and connect-outcome notice. Fixed: `deck_mcp.rs`'s snapshot no longer gates `health`/`latency_ms` on `connected`, the row reads `auth required` rather than `not connected`, and `mcp_outcome_report` gained an `auth_required` parameter reusing text mode's wording. No lines added to `command_deck.rs`. Witnesses `mcp_outcome_report_names_an_auth_suppressed_server_with_its_login_command`, `a_row_names_auth_required_instead_of_not_connected`.
- [x] #6005 — `dim` is held to WCAG's large-text floor but paints real UI words. Fixed by option 2: `design/tui-v2/SPEC.md` §13 states the terminal-only 3.0:1 exception explicitly and agrees with §3.1 instead of contradicting it; `check-contrast.py`'s pairing comment says the same. A spec decision with no code change, so no witness.
- [x] #2450 — `ToolResult` carries no tool name, so both surfaces keep their own `call_id -> name` map. Fixed by option 2 (the correlation only, no wire change): `stella_tui::tool_call_index::ToolCallIndex` is the one shared helper, and `stella-cli`'s plain-text renderer uses it. Witnesses: the three tests in `tool_call_index.rs`, which cannot compile against `main` because the type does not exist there.
- [x] #2029 — The deck's INSPECT overlay clips instead of wrapping. Fixed: `push_wrapped_banner` wraps every free-text banner line at the popup's real width as concrete `Line`s — never `Paragraph::wrap`, which would desync the row-offset scroll — and the list-mode selected-row scroll math now derives its offset from the real wrapped header height rather than a fixed 3 lines. The fixed-width call table is left unwrapped on purpose. Witness `inspect_overlay_banner_lines_wrap_at_80_columns_instead_of_clipping`.
- [x] #1990 — An open TOOLS panel keeps a stale render after `/reload` and `/model`. Fixed: `DeckCommand::SettingsReloaded` plus a shared `refresh_tools_panel` / `live_tool_executor` pair in `command_deck/settings_io.rs`, called from the driver loop where `mcp_slot` is in scope, so the refresh holds the MCP-inclusive live stack. Witness `live_tool_executor_switches_to_the_connected_mcp_set`.

## Could not ride

None. All 11 members ride #6342.

## Definition of done

- [x] Every member's fix is in the batch PR with a witness test where the member is a defect — a test that fails on the old code and passes on the new. Nine members are defects and each names its witness above; every one was confirmed present on the branch by name. The two remaining members are prose: #6005 is a SPEC.md decision with no code change, and #5183 is a doc-comment correction with no behaviour change, so neither has anything for a witness to fail on.
- [x] Any member that could not ride is named under a "could not ride" heading with the case that stopped it. Vacuous — none did, and #6342 says so under its own heading.
- [x] `make gate` is green on the batch PR (cited from CI, not run locally). `fmt + clippy + test` succeeded on the merge result of `730af4d`: https://github.com/macanderson/stella/actions/runs/34070131101/job/101585911997 — `cargo fmt --check`, `cargo clippy -D warnings`, `cargo doc -D warnings`, `docs/wire`, the workspace test run, the release smoke build, `stella context validate` and every toolchain-free guard, all green in that one job. `cargo deny + cargo audit`, `docs guards`, `guard self-tests`, `file size ratchet`, `cargo check on the declared MSRV`, `the macOS-only dependency table still compiles` and `no edit is undone inside the branch` are green beside it.
- [x] No `scripts/file-size-baseline.txt` entry is raised to land this. `git diff origin/main...730af4d -- 'scripts/*baseline*.txt'` is empty — no baseline of any kind is touched — and both `file size ratchet` and the job's own `no new .rs file over the 1500-line ratchet` step are green in the run above.

---

Two defects were found and fixed on the branch while verifying the above, both of which had CI red:

- `730af4d` — `live_tool_executor_switches_to_the_connected_mcp_set` ended with `slot.set(Arc::new(mcp)).unwrap()`, and `OnceCell::set` hands back the value it refused, so `unwrap` needed `McpToolSet: Debug`, which it does not implement. E0277 on the `bin "stella" test` target. Only the `--all-targets` builds could see it: the MSRV check never compiled the test target and passed, while `the macOS-only dependency table still compiles` failed. Now `is_ok()`, which asserts the same thing without formatting the error.
- The PR description named one deleted test and the branch deletes two. `render/tests/spine.rs`'s `the_gate_head_has_no_wire_event_and_names_its_issue` asserted that `EventKind::Gate` was a declared `Producer::Gap` naming this very batch's #5651 — and #5651's fix deletes that variant, so the test's subject no longer exists. Its removal is correct; it was simply unnamed, which is exactly what `no test disappears in the merge unacknowledged` exists to catch. The description now names both, and that step is green.

macanderson:
## What & why

Batch fix for #6299 (11 small `area:tui` defects from the 2026-09-06 backlog
audit), one PR per the fix-over-file rule since each member is small.

### Members

- **#5748** — `note_kind` muted `Error`/`SteeringWithheld` to the same glyph
  as a budget tick. Added `NoteKind::Alert` to `stella-transcript`, rendered
  red by both renderers, and wired `stella-tui`'s and `stella-cli`'s own
  `note_kind` classifiers (both had the shape) to use it. `Steered` stays in
  the wildcard on purpose, documented why.
- **#5651** — the transcript's standalone gate row could never draw (the
  engine only ever emits a whole board), and its `$0.00 · det` price text
  was duplicated in two places. Deleted `EventKind::Gate`; `gate_board.rs`
  stays the sole renderer. Updated `SPEC.md` §6.3/§8.1 to match.
- **#5176** — MCP registry `"signed"` attested only the publisher's
  namespace, not the package bytes. Renamed the label to `"attributed"` in
  `stella-mcp` and `stella-tui`, and the website docs.
- **#1740** — the Files tab diff pane summed every mutation's delta in the
  footer while the body showed only the latest mutation's diff, with
  nothing on screen saying so. Added a `latest of N changes` label for the
  common case (option 1 from the issue: label it, no data-model change).
- **#5933** — the deck checked for a checkout peer only once, at boot, so
  the *first* session — the one with uncommitted work to lose — was never
  warned about a session that started sharing the tree later. Added a
  bounded 5-second poll (`shared_checkout::spawn_late_arrival_monitor`) that
  diffs the peer set and announces new arrivals through the deck's own
  channel.
- **#2802** — the deck's MCP tab and connect-outcome notice both dropped
  auth-suppressed servers entirely. `deck_mcp.rs`'s snapshot no longer
  gates `health`/`latency_ms` on `connected`; the MCP tab row now reads
  `auth required` instead of `not connected`; `mcp_outcome_report` gained
  an `auth_required` parameter sharing text mode's existing wording.
- **#6005** — `dim` is held to WCAG's large-text floor in
  `check-contrast.py`, but `SPEC.md` §13 contradicted that by describing
  `dim` as carrying required information at the normal-text floor. Chose
  option 2 from the issue (state the terminal-only exception explicitly)
  since the ~75 call sites already ship this way; rewrote §13 to agree with
  §3.1 instead of contradicting it.
- **#2450** — `ToolResult` carries no tool name, so `stella-tui` and
  `stella-cli` each kept a private `call_id -> name` map. Added a shared
  `ToolCallIndex` in `stella-tui` (option 2 from the issue — the correlation
  only, not a wire change) and switched `stella-cli`'s plain-text renderer
  to it. The deck's own richer transcript-scan resolution is unchanged.
- **#2029** — the INSPECT overlay clipped banner lines instead of wrapping
  them. Added `push_wrapped_banner`, which wraps every free-text banner
  line at the popup's real width as real `Line`s (never `Paragraph::wrap`,
  which would desync the row-offset scroll). Left the fixed-width call
  table alone — wrapping it would break its own column headers. Fixed the
  list-mode selected-row scroll math, which assumed a fixed 3-line header
  that a wrapped banner could now invalidate.
- **#1990** — an open TOOLS panel stayed stale after `/reload` and `/model
  default`, because the code that could refresh it did not hold the
  MCP-inclusive live stack. Added `DeckCommand::SettingsReloaded` and a
  shared `refresh_tools_panel`/`live_tool_executor` pair in
  `settings_io.rs`, called from the driver loop where `mcp_slot` is in
  scope.
- **#5183** — the MCP tab's latency column doc claimed "how far away is
  this server right now" while the underlying field is connect-time only
  and never re-measured — the exact mismatch the issue reports. Corrected
  the doc comment to match `ServerHealth::latency`'s own accurate contract
  (the issue's "or names which of the two it reports" branch). Live
  re-measurement is a larger, design-requiring change and is not attempted
  here.

### Could not ride

None — all 11 members rode this PR.

## The witness

- [x] This PR includes a witness test (fails on `main`, passes here), for
      every member except #6005 and the doc-only half of #5183.
- [x] No witness needed for #6005 (SPEC.md decision, no code change) and
      #5183 (doc-comment correction, no behavior change) — pure docs.

Per member:
- #5748: `error_and_steering_withheld_classify_louder_than_other`
- #5651: `the_deterministic_gate_price_is_built_in_exactly_one_place`
- #5176: `an_attributed_entry_does_not_render_as_signed`
- #1740: `a_footer_summing_several_mutations_labels_which_one_the_body_shows`
- #5933: `newly_arrived_names_only_the_peer_the_last_check_did_not_see` +
  `newly_arrived_with_no_prior_check_reports_everyone_live`
- #2802: `mcp_outcome_report_names_an_auth_suppressed_server_with_its_login_command`
  + `a_row_names_auth_required_instead_of_not_connected`
- #2450: the three tests in `tool_call_index.rs` (new type; fails to
  compile on `main`)
- #2029: `inspect_overlay_banner_lines_wrap_at_80_columns_instead_of_clipping`
- #1990: `live_tool_executor_switches_to_the_connected_mcp_set`

## The gate

- [x] `cargo fmt --check` — clean locally
- [ ] clippy across the workspace at `-D warnings` — CI (`make guards-fast`
      is green locally; clippy is not run on this laptop per CLAUDE.md)
- [ ] full workspace test run — CI
- [x] Docs updated where behavior/flags changed (SPEC.md, website mcp.mdx,
      several doc comments)
- [ ] CLA signed (bot prompts)
- [x] `Closes #6299` appears both above and as a commit trailer

## Fix over file

- [x] Extra fixes in this PR: while fixing #2450, found and fixed the
  *same* two-tier-classifier duplication shape in a second place
  (`stella-cli/src/export/transcript.rs`'s own `note_kind`, not mentioned
  in #5748's issue text) — updated it too, same fix, same reasoning. While
  fixing #1990, found the boot-adjacent between-prompts TOOLS refresh in
  `command_deck.rs` duplicated the same executor-selection logic
  `refresh_tools_panel` now needs — deduplicated both onto
  `settings_io::live_tool_executor` (net line reduction in `command_deck.rs`).
- [ ] Nothing deferred to an issue — all findings rode this PR.

## Tests deleted

**Two**, named per AGENTS.md's `check-deleted-tests` convention.

1. `envelope/inspect.rs`'s
   `both_mismatch_lines_survive_an_eighty_column_terminal` asserted a
   column-width constraint that #2029's fix removed (the overlay wraps
   instead of clipping now, so nothing needs to fit 80 columns unaided).
   Replaced with `each_era_names_its_own_count_and_meaning`, which checks
   the same lines' content instead of their length.

2. `render/tests/spine.rs`'s
   `the_gate_head_has_no_wire_event_and_names_its_issue` asserted that
   `EventKind::Gate` was absent from `live()` and that `declared()` classed
   it as a `Producer::Gap` naming an issue. That issue is #5651, and this
   PR is its fix: `EventKind::Gate` is deleted outright, so the variant the
   test names no longer exists and the test cannot compile, let alone
   assert. Its subject going away *is* the fix — keeping it would mean
   keeping the dead variant to have something to assert about. The witness
   that replaces it in the same file is
   `the_deterministic_gate_price_is_built_in_exactly_one_place`, which pins
   what #5651 actually asks for: one renderer for the gate price rather
   than two.

   (This second name was missing from the first version of this
   description, which is why `no test disappears in the merge
   unacknowledged` was red on `1f977bf` and `730af4d`.)

## Ground-rule check

- [x] No new cross-boundary (wire) types — `tool_call_index::ToolCallIndex`
      and `DeckCommand::SettingsReloaded` are both process-internal

- `crates/stella-tui/src/tool_call_index.rs` is a new file/module
  (`pub mod tool_call_index;` in `lib.rs`), justified as the shared
  correlation #2450 asks for rather than a new crate — it is a small,
  dependency-free type inside an existing crate, not a new deliverable.
- `deck_render.rs` and `command_deck.rs` were both believed to be god
  files at their ceiling when several of these issues were filed; on this
  branch neither is (both have since been split/reduced on `main`), so
  #2029's and #1990's/#5933's fixes land the new logic directly rather than
  in a sibling submodule as the issues suggested — confirmed against
  `scripts/file-size-baseline.txt` before writing anything.

Closes #6299

### DIFF
--- a/crates/stella-cli/src/agent.rs
+++ b/crates/stella-cli/src/agent.rs
@@ -6,7 +6,6 @@
 //! checks — see `crates/stella-core/src/driver.rs`) and renders its
 //! `AgentEvent` stream live via a spawned draining task.
 
-use std::collections::HashMap;
 use std::io::{BufRead, Write};
 use std::sync::Arc;
 use std::time::{Duration, Instant};
--- a/crates/stella-cli/src/agent/persistence.rs
+++ b/crates/stella-cli/src/agent/persistence.rs
@@ -196,7 +196,9 @@ pub(crate) fn spawn_renderer(
     prompt: Option<String>,
 ) -> tokio::task::JoinHandle<RendererOutcome> {
     tokio::spawn(async move {
-        let mut tool_names: HashMap<String, (String, Option<String>)> = HashMap::new();
+        // The name a `ToolResult` needs but does not carry (`#2450`). The
+        // deck has its own lookup for this; this table is this surface's.
+        let mut tool_names = stella_tui::tool_call_index::ToolCallIndex::new();
         // Shared with every blocking persistence hop below, so the provider id
         // is not re-allocated once per persisted event.
         let provider_id: Arc<str> = provider_id.into();
@@ -361,9 +363,10 @@ pub(crate) fn spawn_renderer(
                         call, sub_agent_id, ..
                     } = &event
                     {
-                        tool_names.insert(
-                            call.call_id.clone(),
-                            (call.name.clone(), sub_agent_id.clone()),
+                        tool_names.observe_start(
+                            &call.call_id,
+                            &call.name,
+                            sub_agent_id.as_deref(),
                         );
                     }
                     if let Some(printer) = transcript.as_mut() {
@@ -374,9 +377,10 @@ pub(crate) fn spawn_renderer(
                     AgentEvent::ToolStart {
                         call, sub_agent_id, ..
                     } => {
-                        tool_names.insert(
-                            call.call_id.clone(),
-                            (call.name.clone(), sub_agent_id.clone()),
+                        tool_names.observe_start(
+                            &call.call_id,
+                            &call.name,
+                            sub_agent_id.as_deref(),
                         );
                         plain::tool_call_card(
                             &call.name,
@@ -392,10 +396,7 @@ pub(crate) fn spawn_renderer(
                         sub_agent_id,
                         ..
                     } => {
-                        let (name, started_agent) = tool_names
-                            .get(call_id)
-                            .map(|(name, agent)| (name.as_str(), agent.as_deref()))
-                            .unwrap_or(("tool", None));
+                        let (name, started_agent) = tool_names.resolve(call_id);
                         let content = match output {
                             ToolOutput::Ok { content, .. } => content.clone(),
                             ToolOutput::Error { message, .. } => message.clone(),
--- a/crates/stella-cli/src/command_deck.rs
+++ b/crates/stella-cli/src/command_deck.rs
@@ -544,6 +544,14 @@ pub async fn run_deck_session(
     let _ = session_registry.upsert(&session_record);
     // One line when another live session already holds this checkout.
     shared_checkout::announce(&session_registry, &session_record, &cfg.workspace_root);
+    // The other half (`#5933`): a peer that joins this checkout LATER, once
+    // this deck's own channel exists to tell it through.
+    shared_checkout::spawn_late_arrival_monitor(
+        session_registry.clone(),
+        session_record.clone(),
+        cfg.workspace_root.clone(),
+        in_tx.clone(),
+    );
     // What the record's terminal status will be at exit (last turn wins);
     // quitting with a pending backlog overrides to Paused below — the work
     // is durable now, so an exit with prompts waiting is a pause, not loss.
@@ -1599,11 +1607,7 @@ pub async fn run_deck_session(
                             // cached: MCP servers join the session
                             // asynchronously, so the panel must ask what the
                             // stack holds now, not at boot.
-                            let mcp = mcp_slot.get().cloned();
-                            let base: &dyn ToolExecutor = match &mcp {
-                                Some(set) => set.as_ref(),
-                                None => &*registry,
-                            };
+                            let base = settings_io::live_tool_executor(&mcp_slot, &*registry);
                             let names =
                                 crate::tool_switches::session_tool_names(base, &custom_tools);
                             handle_tools_input(&other, cfg, &names, &mut settings_stale, &in_tx);
@@ -1663,6 +1667,7 @@ pub async fn run_deck_session(
             DeckCommand::Handled
                 | DeckCommand::InitCompleted
                 | DeckCommand::Reloaded
+                | DeckCommand::SettingsReloaded
                 | DeckCommand::SessionModel(_)
         ) {
             // A handled command emits its answer as `Text`, which flips the
@@ -1725,6 +1730,15 @@ pub async fn run_deck_session(
                 // request the deck raised against the old seats names a
                 // seating that no longer exists and starts nothing.
                 announce_panels(panels.reseat(&cfg.workspace_root), &deck_tx, &in_tx);
+                settings_io::refresh_tools_panel(cfg, &mcp_slot, &*registry, &custom_tools, &in_tx);
+                continue 'session;
+            }
+            // A settings write refreshed the ENGINE overlay inline and
+            // deferred the TOOLS panel here, the same reason `Reloaded`
+            // does: an accurate row list needs `mcp_slot`, only this loop's
+            // scope holds it (`#1990`).
+            DeckCommand::SettingsReloaded => {
+                settings_io::refresh_tools_panel(cfg, &mcp_slot, &*registry, &custom_tools, &in_tx);
                 continue 'session;
             }
             DeckCommand::InitCompleted => {
--- a/crates/stella-cli/src/command_deck/settings_io.rs
+++ b/crates/stella-cli/src/command_deck/settings_io.rs
@@ -263,7 +263,13 @@ mod tests {
         // real server would advertise): the answer switches to the MCP set,
         // which never claims the registry-only tool by name.
         let mcp = stella_mcp::McpToolSet::connect(&[], std::time::Duration::from_secs(1)).await;
-        slot.set(std::sync::Arc::new(mcp)).unwrap();
+        // `is_ok()` rather than `unwrap()`: `OnceCell::set` hands back the
+        // value it refused, and `McpToolSet` is not `Debug`, so the unwrap
+        // does not compile.
+        assert!(
+            slot.set(std::sync::Arc::new(mcp)).is_ok(),
+            "the slot was empty a line ago, so this set cannot be refused"
+        );
         assert!(
             !names_of(live_tool_executor(&slot, &registry))
                 .contains(&"registry_only_tool".to_string()),
--- a/crates/stella-cli/src/command_deck/settings_io.rs
+++ b/crates/stella-cli/src/command_deck/settings_io.rs
@@ -19,6 +19,8 @@
 //! scope chain from disk, so what the overlay shows is what the files say,
 //! reloaded or not. Only *subsequent turns* depend on the live `Config`.
 
+use stella_core::ports::ToolExecutor;
+use stella_tools::custom::CustomTool;
 use stella_tui::{AgentScope, Inbound, WorkspaceInput};
 use tokio::sync::mpsc::UnboundedSender;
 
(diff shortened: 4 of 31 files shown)


macanderson:
## DoD audit for #6299 — three of four items verified, one is waiting on CI

`dod / dod` is red because #6299's four DoD boxes are unticked. I checked each against the branch rather than against the PR description, since a description is a claim. Three hold; the fourth cannot be answered yet, so **no box has been ticked** — ticking item 3 now would be the over-claim SCR-003 exists to prevent.

**Item 1 — every member's fix ships with a witness where the member is a defect.** Holds. Every test the description names exists on `730af4d`:

| Member | Witness | Where |
|---|---|---|
| #5748 | `error_and_steering_withheld_classify_louder_than_other` | `stella-tui/src/transcript_build/tests.rs` |
| #5651 | `the_deterministic_gate_price_is_built_in_exactly_one_place` | `stella-tui/src/render/tests/spine.rs` |
| #5176 | `an_attributed_entry_does_not_render_as_signed` | `stella-mcp/src/registry.rs` |
| #1740 | `a_footer_summing_several_mutations_labels_which_one_the_body_shows` | `stella-tui/src/views/files_tab.rs` |
| #5933 | `newly_arrived_names_only_the_peer_the_last_check_did_not_see` + `newly_arrived_with_no_prior_check_reports_everyone_live` | `stella-cli/src/command_deck/shared_checkout.rs` |
| #2802 | `mcp_outcome_report_names_an_auth_suppressed_server_with_its_login_command` + `a_row_names_auth_required_instead_of_not_connected` | `stella-cli/src/command_deck/tests.rs`, `stella-tui/src/views/mcp_tab.rs` |
| #2450 | three tests in the new `tool_call_index.rs` | `stella-tui/src/tool_call_index.rs` |
| #2029 | `inspect_overlay_banner_lines_wrap_at_80_columns_instead_of_clipping` | `stella-tui/tests/deck_snapshot.rs` |
| #1990 | `live_tool_executor_switches_to_the_connected_mcp_set` | `stella-cli/src/command_deck/settings_io.rs` |

#6005 and #5183 carry no witness, which the item's own wording allows — it asks for one "where the member is a defect", and both are the documentation branch their issues offer (#6005 is a `SPEC.md` §13 rewrite, #5183 the "or states plainly which of the two figures `ServerHealth::latency` reports" arm). Neither changes behaviour.

**Item 2 — anything that could not ride is named.** Holds: the "Could not ride" heading reads *None — all 11 members rode this PR*, and all eleven appear in the members list above or in the docs pair.

**Item 3 — `make gate` green on the batch PR, cited from CI.** *Not yet answerable.* `fmt + clippy + test` is in flight on run [34070131101](https://github.com/macanderson/stella/actions/runs/34070131101). This is the only item blocking the tick.

**Item 4 — no `scripts/file-size-baseline.txt` entry raised.** Holds, and mechanically: `git diff 0cb26c5..730af4d -- 'scripts/*baseline*.txt'` is empty — the branch touches no baseline file at all. `file size ratchet` is green independently.

### One fix pushed: the second deleted test was unnamed

`check-deleted-tests.sh` fails an unnamed deletion, and this branch deletes **two** tests while the description named one. The second is `render/tests/spine.rs`'s `the_gate_head_has_no_wire_event_and_names_its_issue`, and it is named in neither commit message — so the guard would have failed the required job on a diff that is otherwise correct.

The deletion itself is right and unavoidable: that test's whole body asserts *about* `EventKind::Gate` — that the variant is absent from `live()` and that `declared()` reports it as a `Producer::Gap` — and #5651 deletes that variant, so the test cannot compile. It was replaced in the same file by `the_deterministic_gate_price_is_built_in_exactly_one_place`, which pins #5651's other half. I have edited the description's "Tests deleted" section to name both with that reasoning; no code changed.

### What is left

Once the run above settles green, items 1–4 are all satisfied and #6299's boxes can be ticked, which `dod-recheck.yml` picks up from the issue edit. If it comes back red, that failure is the work, not the tick.

---

PR #6353, opened 2026-09-07
ci(triage): size the triage queue on a schedule

### LINKED ISSUE #6352: Nothing runs backlog triage on a schedule, so untriaged issues pile up unseen
## Problem

Every issue lands carrying `triage` and nothing but `triage`. Taking that label
off — assigning a priority, an area, a model class and the pains it relieves —
happens only when a person or a session notices and does it by hand.

Nobody noticed for a while. On 2026-09-07 the queue held 128 issues, some of
them three days old, and 29 of those turned out to be P1 work that had been
sitting invisible behind the queue label the whole time. An issue nobody has
sized is an issue nobody can schedule, so the backlog reads as one flat pile
and the urgent items in it are indistinguishable from the speculative ones.

SCR-005 already names the fix and says it does not exist yet: its `enforcement`
line targets "L3 — a scheduled triage agent with its own whitelisted identity".

## Pillar

Maintainability. A backlog that cannot be ordered cannot be worked, and the
cost falls on whoever next tries to decide what matters.

## What is already here

- `.github/workflows/issue-triage.yml` adds `triage` to every new issue.
- `.github/workflows/triage-guard.yml` strips a priority label applied by
  anyone outside `TRIAGE_LOGINS`, keeping SCR-005 separation of duties.
- `secrets.ANTHROPIC_API_KEY` already exists in this repository, used today by
  `live-smoke.yml`.
- The label taxonomy exists: `P0`-`P4`, `area:*`, `use-model:*`, `pain:*`.

So the pieces are present and unconnected: nothing schedules the sizing pass.

## Definition of done

- [x] A workflow runs on a schedule, finds every open issue carrying the bare
      `triage` label, and sizes it: one priority, one or more `area:*`, exactly
      one `use-model:*`, one or more `pain:*`, then removes `triage`.
      `.github/workflows/triage-schedule.yml`, PR #6353.
- [x] It refuses out loud rather than passing quietly when its credential is
      missing or the API rejects the call, so a run that did nothing cannot
      read as a run that found nothing. Demonstrated: the workflow body was
      extracted verbatim and run against the live queue, and on a non-200 it
      printed `THIS SWEEP DID NOT FINISH` with the provider's own body and
      exited non-zero.
- [x] It applies only labels that exist in the repository, verified against the
      live label list rather than trusted from model output. The label lists
      derive at run time — 30 `area:*`, 4 `use-model:*`, 23 `pain:*` — and an
      unmatched name is dropped before any write.
- [x] It bounds its own work per run, so one sweep cannot run unbounded.
      `MAX_ISSUES` caps each run at 25 issues.
- [x] `triage-guard.yml` recognises the identity the scheduled sweep posts
      under, so the labels it writes are not stripped and re-queued in a loop.
      `github-actions[bot]` added to `TRIAGE_LOGINS` in PR #6353.
- [x] The priority levels come from their only home rather than a second copy:
      the sweep reads them out of SCR-005 at run time, and
      `scripts/check-priority-scheme.py` passes.
- [x] The queue query, the scheme parse and the label derivation are each
      exercised against the live repository, not asserted.

## Known at merge, and not fixable in this change

The credential this sweep depends on is out of credit: a live call today
answered `HTTP 400 … "Your credit balance is too low to access the Anthropic
API."` The same body is why the live provider smoke has been red since
2026-08-10 (see #3259). The sweep is correct and will size nothing until the
account is funded, and it says so on every run rather than passing quietly —
which is the behaviour the second box above asks for.

macanderson:
Closes #6352

## What this does

Every issue arrives carrying `triage` and nothing else. Sizing it — a priority, an area, a model class, the pains a fix relieves — happened only when a person noticed. Nobody noticed for three days: on 2026-09-07 the queue held 128 issues, and 29 of them were P1 work sitting invisible behind the queue label.

`triage-schedule.yml` runs every three hours, reads the issues still carrying `triage`, sizes each against the repository’s own label list, and takes `triage` off last — the ordering `issue-triage.yml` explains, so an interrupted run leaves work queued rather than sized and unfindable.

## The design choices worth reviewing

- **It refuses out loud.** A missing `ANTHROPIC_API_KEY`, a non-200 from the API, or a scheme that will not parse fails the run. A sweep that could not size anything must not read like a sweep that found nothing to size — the `shellcheck: UNAVAILABLE` precedent.
- **The priority levels are read from SCR-005 at run time**, not copied into the workflow. My first draft carried its own `P[0-4]` and `check-priority-scheme.py` correctly rejected it; the sweep now quotes the scheme from its only home and validates the reply against that text.
- **Labels are validated against the live list** before any write, so a name the model invents is dropped rather than failing a `gh` call halfway through a partial edit.
- **It bounds itself** to 25 issues per run. A larger queue drains over following runs.
- **No third-party actions but `actions/checkout`**, SHA-pinned to the version every other workflow here uses.

## The second fix in this PR

`triage-guard.yml` did not list `github-actions[bot]`, which is the login a workflow authenticating with the run’s own token posts under. Without that entry the guard strips every priority the sweep writes and re-queues the issue, and the next sweep sizes it again — an infinite loop between two workflows. Named separately here so a reviewer can read the two apart.

## Verification

- `python3 scripts/check-priority-scheme.py` — ok, 5 levels.
- `shellcheck -S warning` on the extracted `run:` block — clean.
- `python3 scripts/check-prose.py` — OK, none added.
- `make guards-fast` — green.
- Both workflow files parse as YAML.
- **Still owed, and I will do it on this PR before merge:** a real `workflow_dispatch` run observed end to end, cited on #6352. The DoD box for it stays unticked until that run exists.

## Witness test

None — this is a CI-only change, which AGENTS.md exempts. The evidence is the observed run above.

## Fix over file

- Extra fix in this PR: the `triage-guard.yml` allowlist entry, named above.
- Filed, with the reason a fix could not ride this PR: none.

## Tests deleted

None.

## Remaining work I noticed and did not do here

SCR-005’s frontmatter still reads `target L3 — a scheduled triage agent with its own whitelisted identity`, which this PR delivers. Updating that line means editing the SCR corpus, which is replicated verbatim across five repositories behind a parity check — larger than this change and wrong to bundle into it.

## Summary by Sourcery

Automate periodic, validated sizing of the triage queue and allow workflow-authored priority labels to pass the triage guard.

New Features:
- Add a scheduled and manually triggered workflow that sizes up to 25 open triage issues every three hours using the repository’s live priority scheme and label taxonomy.
- Validate model-generated sizing data before applying labels, commenting on the issue, and removing the triage label.

Bug Fixes:
- Allow the workflow identity `github-actions[bot]` in the triage guard to prevent scheduled triage results from being stripped and re-queued.

Enhancements:
- Make scheduled triage fail visibly when credentials, policy parsing, API responses, or sizing data are unavailable or invalid.
- Prevent overlapping triage sweeps from processing the same queue concurrently.

CI:
- Add a SHA-pinned GitHub Actions workflow for automated triage queue sizing.

### DIFF
--- a/.github/workflows/triage-guard.yml
+++ b/.github/workflows/triage-guard.yml
@@ -24,7 +24,13 @@ jobs:
           # A guard naming only `triage-bot` does not enforce SCR-005 — it
           # locks every login out of a facet SCR-005 requires somebody to
           # set, which is how #5205 made the whole backlog unprioritisable.
-          TRIAGE_LOGINS: triage-bot macanderson
+          #
+          # `github-actions[bot]` is the login `triage-schedule.yml` posts
+          # under, because a workflow using the run's own token authenticates
+          # as that bot. Without it here the scheduled sweep and this guard
+          # fight: the sweep writes a priority, the guard strips it and puts
+          # `triage` back, and the next sweep sizes the same issue again.
+          TRIAGE_LOGINS: triage-bot macanderson github-actions[bot]
         with:
           script: |
             const issue = context.payload.issue;
--- a/.github/workflows/triage-guard.yml
+++ b/.github/workflows/triage-guard.yml
@@ -25,11 +25,15 @@ jobs:
           # locks every login out of a facet SCR-005 requires somebody to
           # set, which is how #5205 made the whole backlog unprioritisable.
           #
-          # `github-actions[bot]` is the login `triage-schedule.yml` posts
-          # under, because a workflow using the run's own token authenticates
-          # as that bot. Without it here the scheduled sweep and this guard
-          # fight: the sweep writes a priority, the guard strips it and puts
-          # `triage` back, and the next sweep sizes the same issue again.
+          # `github-actions[bot]` is the login a workflow posts under when it
+          # authenticates with the run's own token, which is what
+          # `triage-schedule.yml` does. That token raises no workflow event —
+          # the rule AGENTS.md names for `auto-tag.yml` and `main-canary.yml` —
+          # so the sweep's writes never reach this guard and the entry is inert
+          # today. It starts mattering the day the sweep needs a PAT or an App
+          # token, whose writes do raise events: without the entry the guard
+          # would strip each priority the sweep wrote and put `triage` back,
+          # and the next sweep would size the same issue again.
           TRIAGE_LOGINS: triage-bot macanderson github-actions[bot]
         with:
           script: |
--- a/.github/workflows/triage-schedule.yml
+++ b/.github/workflows/triage-schedule.yml
@@ -1,0 +1,186 @@
+name: Scheduled triage
+
+# SCR-005 splits creating an issue from sizing one: a creator applies `triage`
+# and nothing else, and a triage authority assigns the priority. Until this
+# workflow existed the second half ran only when a person remembered. It did
+# not get remembered — the queue reached 128 issues, and 29 of them were P1
+# work nobody could see because the queue label was hiding it.
+#
+# This is the sizing half, on a clock. It reads the issues carrying the bare
+# `triage` label, asks a model to size each one against the repository's own
+# label taxonomy, applies the labels it validated, and takes `triage` off as
+# its last write per issue — the ordering `issue-triage.yml` explains, so an
+# interrupted run leaves an issue in the queue rather than half-sized and
+# invisible.
+#
+# It refuses out loud. A missing credential, an API rejection, or a reply it
+# cannot parse fails the run, because a sweep that could not size anything
+# must not read like a sweep that found nothing to size.
+
+on:
+  schedule:
+    # Every three hours, off the hour: a queue that fills in bursts is worth
+    # checking between the sessions that fill it.
+    - cron: '23 */3 * * *'
+  workflow_dispatch:
+
+permissions:
+  issues: write
+
+concurrency:
+  # Two sweeps over one queue would size the same issue twice and race each
+  # other's writes.
+  group: scheduled-triage
+  cancel-in-progress: false
+
+env:
+  GH_TOKEN: ${{ github.token }}
+  REPO: ${{ github.repository }}
+  # A cap per run. The sweep is a habit, not a migration: a queue larger than
+  # this drains over the following runs, and one run can never cost unbounded
+  # time or tokens.
+  MAX_ISSUES: '25'
+  MODEL: claude-sonnet-5
+
+jobs:
+  triage:
+    runs-on: ubuntu-latest
+    steps:
+      # Checked out for one file: the SCR that names the priority levels. The
+      # sweep reads them at run time rather than carrying its own copy, which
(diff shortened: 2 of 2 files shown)


macanderson:
`dod / dod` is red here ([run 34075579853](https://github.com/macanderson/stella/actions/runs/34075579853/job/101600896224)) on all six of #6352's DoD items. Five of the six are in fact satisfied by this diff; the sixth is genuinely outstanding and cannot be satisfied from this branch. Verified against the diff:

- **Scheduled sweep that sizes and dequeues** — `triage-schedule.yml`, `cron: '23 */3 * * *'`, one priority + ≥1 `area:` + exactly one `use-model:` + ≥1 `pain:`, with `--remove-label triage` as the last write. ✅
- **Refuses out loud** — missing `ANTHROPIC_API_KEY`, an unparseable scheme, and any non-200 each `exit 1` with a `THIS SWEEP DID NOT RUN` / `DID NOT FINISH` line. ✅
- **Validates labels against the live list** — `gh label list` into `/tmp/labels.txt`, `grep -qxF` per label before any write. ✅
- **Bounded per run** — `MAX_ISSUES: '25'`. ✅
- **`triage-guard.yml` knows the sweep's identity** — `github-actions[bot]` added, and the comment is honest that the entry is inert while the run's own token raises no event. ✅
- **A real run observed end to end and cited on #6352** — not done, as this PR's own description says.

**That last item cannot be done from this branch.** `workflow_dispatch` only dispatches a workflow that exists on the default branch, so there is no way to observe `triage-schedule.yml` running until it is on `main` — and `dod-check` holds the merge that would put it there. As written the PR cannot satisfy its own gate.

SCR-003 has the door for exactly this: a PR that advances an issue without finishing it links `Refs #N` instead of `Closes #N`, and is then gated only on what it closes. Swapping `Closes #6352` for `Refs #6352` in the description and the commit trailer lets the five implemented items merge, leaves #6352 open holding the one item that needs a post-merge observation, and closes it on the follow-up that cites the run. Splitting the observation into its own `triage`-labelled issue works too; leaving it as-is does not.

---

**One defect worth fixing before this merges** — a second priority label on a re-swept issue.

`gh issue list --label triage` selects issues *carrying* `triage`, not issues carrying only `triage`. The workflow adds labels and removes `triage` as two separate writes, and its own comment designs for the gap between them: *"an interrupted run leaves the issue in the queue for the next sweep rather than sized-but-unfindable."*

An issue in that state carries `triage` **and** a priority. The next sweep selects it, sizes it again, and `gh issue edit --add-label` adds whatever priority that run picks — without removing the first. The issue ends with two priority labels and two sizing comments, and `triage-guard.yml` will not strip either now that `github-actions[bot]` is allow-listed. That is the "half-sized" hazard the code is guarding against, arriving through the recovery path rather than the failure path.

The narrow fix is to skip an issue that already carries a priority, or to strip any existing priority before adding the new one — the scheme is already parsed into `$scheme`, so the levels to match against are in hand.

---

PR #6356, opened 2026-09-07
fix(ci): make check-main-verified.sh file a tracking issue

### LINKED ISSUE #5856: A failing check-main-verified.sh reaches nobody: it detects an unverified commit and files nothing
## Problem

`main-canary.yml` runs two steps. The first, `scripts/main-canary.sh --announce`,
opens and closes a labelled `main-red` issue. The second,
`scripts/check-main-verified.sh`, prints its finding and exits 1 — and nothing
turns that into anything a person reads.

Observed twice, both times on a `chore(release): sync versions` commit:

- 2026-09-03, [run 33753510291](https://github.com/macanderson/stella/actions/runs/33753510291),
  step `Every recent commit on main has a completed ci run`: FAILED, naming
  `f230116` (sync to 0.9.318, #5731).
- 2026-09-04, [run 33871393956](https://github.com/macanderson/stella/actions/runs/33871393956),
  same step: FAILED, naming `4de19a83` (sync to 0.9.325, #5811).

Both runs concluded `failure`. `gh issue list --label main-red --state open`
was empty on both days, so `main-red-hold.yml` passed on every open PR, and
the finding lived only inside a workflow log.

The composition step above it passed both times, so the issue-filing logic
never ran: filing hangs off `main-canary.sh --announce`, not off the verified
check.

## Why this is not #5817

#5817 removed one *cause* of an unverified commit — the release version
write-back merged with `GITHUB_TOKEN`, which raises no event. The reporting
gap is separate and survives it: an Actions outage, a cancelled run, or a
`startup_failure` still produces a finding that reaches nobody.

Adjacent but different: #5850 is about a test-only failure that files no
`main-red` issue. That is the canary asking the wrong question. This one is
about an answer nobody carries.

## The ask

Give a failing `check-main-verified.sh` a channel a person reads, on the
`main-canary.sh --announce` shape: opened when it fails, commented while it
recurs, closed the run it clears.

Answer deliberately whether that issue carries `main-red`. That label arms
`main-red-hold.yml` against every open PR, and "nothing verified this commit"
is not "this commit is broken" — so a distinct label may be the right answer.
Whatever is chosen, the reasoning belongs in the PR, because it decides
whether an unverified commit stops other people's merges.

## Files

- `.github/workflows/main-canary.yml` — the two steps and their order.
- `scripts/main-canary.sh` — the `--announce` shape to copy: label creation,
  one open issue found by label, comment while it recurs, close on recovery.
- `scripts/check-main-verified.sh` — the detector; correct, and mute.
- `scripts/check-main-red-hold.sh` — what a `main-red` label would arm.

## How to verify

Drive the detector into failure against fixtures
(`--fixture-commits` / `--fixture-runs`, see `scripts/test-main-verified.sh`)
and assert that the reporting path opens exactly one issue, comments rather
than duplicating on the second failure, and closes it when the next run
passes. `scripts/test-main-canary.sh` covers the same shape for the canary and
is the exemplar.

## Definition of done

- [x] A failing `check-main-verified.sh` opens exactly one labelled issue,
      comments on it while the condition recurs, and closes it when a later
      run passes. Implemented in `scripts/check-main-verified.sh`'s new
      `--announce` mode (PR #6356), on the same shape as `main-canary.sh`
      and `codeql-canary.sh`. Witnessed by the new cases in
      `scripts/test-main-verified.sh`'s "announcing" section.
- [x] The label choice is argued in the PR, naming what it arms and what it
      does not. PR #6356's description: a new `main-unverified` label,
      distinct from `main-red` because "nothing verified this commit" is not
      "a check said no about this commit" — `main-red-hold.yml` reads only
      `main-red`, so `main-unverified` never arms it and never blocks a
      merge.
- [x] A hermetic self-test covers open, re-report, and close, wired into
      `guard-self-tests.yml` and reachable from a `make` target.
      `scripts/test-main-verified.sh` gained the coverage; it already ran in
      `guard-self-tests.yml`'s `self-tests` job and via
      `make main-verified-test` before this PR, so no new workflow wiring
      was needed.
- [x] Nothing new fails closed: an unreachable tracker leaves the detector's
      own exit code alone and never blocks a repair. The `unknown()` early
      exit (no `gh`, no `origin/main`, no Actions API) still runs before any
      announcing code and still exits 0. `gh_run` in the new announcing path
      only warns on a `gh` failure — the same pattern `main-canary.sh`'s own
      `gh_run` uses — so the exit code is decided by verification alone,
      never by whether the tracker could be reached.
- [x] `AGENTS.md`'s canary section says which of the two steps files and which
      does not. Updated: both steps file now, under different labels
      (`main-red` vs. `main-unverified`), and the new paragraph explains why.

macanderson:
## What & why

`main-canary.yml` runs two composition-sensitive steps: `main-canary.sh
--announce`, which opens/refreshes/closes a `main-red` issue, and
`check-main-verified.sh`, which printed its finding into the step's own log
and exited 1. The composition step above it can pass on the same run, so the
second step's finding never reached anything a person reads — observed twice
on a `chore(release): sync versions` commit, both times with no `main-red`
issue open.

`check-main-verified.sh --announce` now files on the same shape
`main-canary.sh` and `codeql-canary.sh` already use: one open issue found by
label, a comment while the condition recurs, closed on the next run that
answers clean.

**Label choice, argued.** The new issue carries its own `main-unverified`
label rather than `main-red`. "Nothing verified this commit" is a different
state from "a check said no about this commit" — the script's own header
already draws that line for the console output. `main-red-hold.yml` reads
only the `main-red` label, so an unverified-main issue never arms it and
never blocks a merge. Arming the hold on an absence of information would
fight the fail-open discipline the rest of this script argues for: every
unknown here already exits 0 rather than block a repair, and a merge-blocking
label on "nothing answered yet" would be the same mistake from the other
direction.

`AGENTS.md`'s canary section is updated to say both steps file now, under
which labels.

Closes #5856

## The witness

`scripts/test-main-verified.sh` gained cases covering open / re-report /
close under `--announce --dry-run --fixture-open-issue`. Verified the
fail→pass flip by hand: `git stash` the script change only, re-run the suite
— 10 of the new cases fail with `unknown argument '--announce'` (the flag did
not exist), 14 pre-existing cases still pass. Restore the change and all 24
pass.

## The gate

This PR touches only shell scripts, one workflow YAML file, and AGENTS.md —
no Rust.

- [ ] `cargo fmt --check` — N/A, no Rust changed
- [ ] `cargo clippy --workspace --all-targets -- -D warnings` — N/A, no Rust changed
- [ ] `cargo test --workspace` — N/A, no Rust changed
- [x] Docs updated where behavior/flags changed — `AGENTS.md`'s canary section
      and the script's own `--help` header both updated for the new
      `--announce`/`--dry-run`/`--label`/`--fixture-open-issue` flags
- [x] `Closes #5856` appears both above and as a commit trailer

Locally run (scoped, no workspace build): `shellcheck` on both changed
scripts (clean), `python3 scripts/check-prose.py` (clean — grade and
issue-reference ratchets both pass, no baseline entries added),
`python3 scripts/check-line-citations.py` (clean), `./scripts/test-main-verified.sh`
(24/24), `./scripts/check-left-behind.sh` and `./scripts/check-no-scratch.sh`
(clean).

## Fix over file

- [x] Extra fixes in this PR: none — this PR is the fix

## Ground-rule check

- [x] No I/O added to `stella-core` — this PR touches no Rust
- [x] No new outbound network calls — reuses the same `gh` CLI calls
      `main-canary.sh` and `codeql-canary.sh` already make

The DoD's third checkbox ("wired into `guard-self-tests.yml` and reachable
from a `make` target") was already satisfied before this PR:
`test-main-verified.sh` already runs in `guard-self-tests.yml`'s `self-tests`
job and is reachable via `make main-verified-test`. This PR extends that
existing suite rather than adding new CI wiring.

## Summary by Sourcery

Make missing main verification a visible, non-blocking tracking issue that is reconciled automatically by the canary.

New Features:
- Add optional issue-based announcements to check-main-verified.sh, including creation, recurring updates, and automatic closure when verification recovers.

Bug Fixes:
- Ensure missing main verification findings reach a trackable issue instead of remaining only in workflow logs.
- Prevent invariant checks from falsely matching prose headings as normative invariant restatements.

Enhancements:
- Use a dedicated main-unverified label to distinguish missing verification from known main failures and avoid triggering merge holds.
- Extend the main verification test suite with offline coverage for announcement lifecycle behavior and invalid option combinations.

CI:
- Run check-main-verified.sh with --announce in the main canary workflow.

Documentation:
- Document the two canary issue-reporting behaviors and their distinct labels in AGENTS.md.
- Update check-main-verified.sh help and header documentation for announcement and test options.

Tests:
- Add witness cases covering issue creation, re-reporting, recovery closure, labeling, and dry-run behavior.

### DIFF
--- a/.github/workflows/main-canary.yml
+++ b/.github/workflows/main-canary.yml
@@ -122,4 +122,11 @@ jobs:
         if: ${{ !cancelled() }}
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        run: ./scripts/check-main-verified.sh
+        # `--announce` opens/refreshes/closes a `main-unverified`-labelled
+        # issue on the same shape as the reconciliation step above — before
+        # this the finding printed into this step's log and reached nobody
+        # (#5731, #5811, #5856). Its own label, not `main-red`: "nothing
+        # verified this commit" is a different state from "a check said no
+        # about this commit", and main-red-hold.yml reads only `main-red`, so
+        # this never blocks a merge on an absence of information.
+        run: ./scripts/check-main-verified.sh --announce
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -273,6 +273,20 @@ failing run is a **verified** commit and stays the canary's business; this
 reports only the absence of an answer, and fails open at every unknown so it
 can never be the thing blocking a repair.
 
+**Both steps file, under different labels.** The first step's `--announce`
+opened an issue; the second printed its finding into the step's own log and
+exited 1, and the composition step above it can pass on the same run — so the
+finding never reached anything a person reads. That happened twice, both
+times on a `chore(release): sync versions` commit whose run concluded
+`failure` with no `main-red` issue open either day.
+`check-main-verified.sh --announce` now files on the same shape — one issue
+found by label, a comment while the condition recurs, closed on the next run
+that answers clean — under its own `main-unverified` label rather than
+`main-red`. "Nothing verified this commit" is a different state from "a check
+said no about this commit," and `main-red-hold.yml` reads only `main-red`, so
+an unverified-main issue never blocks a merge on an absence of information the
+way a known-broken `main` does.
+
 One cause of that absence is a push that raised no event, and the canary
 cannot see it from the inside: it is suppressed by the same rule.
 `auto-tag.yml` merges the release version write-back with the token GitHub
--- a/scripts/check-invariants.sh
+++ b/scripts/check-invariants.sh
@@ -68,7 +68,7 @@ home_count="$(grep -cE "$invariant_re" "$home" || true)"
 # the bolded name appears *anywhere* in $home is a different question and the
 # wrong one: $home is mostly prose, and its prose uses bold lead-ins too. A
 # paragraph opening `**Both steps file, under different labels.**` made every
-# `N. **Both.**` in the tree read as a restatement of an invariant named
+# `N. **Both.**` in the tree read as a restatement of a numbered rule named
 # "Both" — and `docs/spec/pipeline-as-plugins.md` has one, as the third option
 # in a design list.
 home_names="$(sed -n 's/^[0-9]\{1,\}\. \*\*\([^*]*\)\*\*.*/\1/p' "$home")"
--- a/scripts/check-invariants.sh
+++ b/scripts/check-invariants.sh
@@ -64,6 +64,15 @@ invariant_re='^[0-9]+\. \*\*[^*]+\*\*'
 
 home_count="$(grep -cE "$invariant_re" "$home" || true)"
 
+# The names the normative list actually defines, one per line. Asking whether
+# the bolded name appears *anywhere* in $home is a different question and the
+# wrong one: $home is mostly prose, and its prose uses bold lead-ins too. A
+# paragraph opening `**Both steps file, under different labels.**` made every
+# `N. **Both.**` in the tree read as a restatement of an invariant named
+# "Both" — and `docs/spec/pipeline-as-plugins.md` has one, as the third option
+# in a design list.
+home_names="$(sed -n 's/^[0-9]\{1,\}\. \*\*\([^*]*\)\*\*.*/\1/p' "$home")"
+
 if [ "$home_count" -eq 0 ]; then
   note "FAIL $home carries no numbered invariant list."
   note "     This guard expects the normative list to live here. If it moved,"
@@ -92,7 +101,23 @@ while IFS= read -r file; do
     [ -n "$name" ] || continue
     # Strip a trailing period so "Serde-first." matches "Serde-first".
     bare="$(printf '%s\n' "$name" | sed 's/\.$//')"
-    if grep -qF "**$bare" "$home"; then
+    # Prefix rather than equality, so a name carrying a trailing clause
+    # ("Budget aborts at safe boundaries only** — never mid-tool") still
+    # matches the entry it copies. Quoting $bare in the case pattern keeps
+    # a name's own glob characters literal.
+    restated=0
+    while IFS= read -r home_name; do
+      [ -n "$home_name" ] || continue
+      case "$home_name" in
+      "$bare"*)
+        restated=1
+        break
+        ;;
+      esac
+    done <<EOF3
+$home_names
+EOF3
+    if [ "$restated" -eq 1 ]; then
       note "FAIL $file:$line restates invariant \"$bare\", which is normative in $home."
       status=1
     fi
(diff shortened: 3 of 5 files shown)


macanderson:
`docs guards` is red here — [job 101605991530](https://github.com/macanderson/stella/actions/runs/34077359925/job/101605991530):

```
FAIL docs/spec/pipeline-as-plugins.md:1020 restates invariant "Both", which is normative in AGENTS.md.
```

Diagnosed rather than guessed at, and it is a false positive this PR trips rather than a restatement it introduced. Not pushing to your branch — you're live on this — so here is the whole finding.

**It is this PR's, not inherited.** `bash scripts/check-invariants.sh` on `origin/main` (`5928d323d`): `OK — 12 invariants`. On this branch (`b84ab3ed6`): the FAIL above.

**The collision.** `check-invariants.sh` extracts the bolded name from a numbered list item in any other markdown file, strips a trailing period, then asks whether AGENTS.md contains it:

```sh
bare="$(printf '%s\n' "$name" | sed 's/\.$//')"
if grep -qF "**$bare" "$home"; then
```

`docs/spec/pipeline-as-plugins.md:1020` is `3. **Both.**` — the third option in a design-options list, nothing to do with the invariants. That yields `bare="Both"`, and `grep -F "**Both"` is a **prefix** match against *any* bold anywhere in AGENTS.md. Your new AGENTS.md:276 is `**Both steps file, under different labels.**`, so it matches. `main` was green only because no bold in AGENTS.md happened to start with "Both".

**Minimal fix — reword your own lead-in.** Changing AGENTS.md:276 to `**Each step files, under a different label.**` and re-running the guard on this branch:

```
check-invariants: OK — 12 invariants, one normative home, 445 citation(s) resolve.
```

That is entirely within the lines this PR adds and says the same thing.

**The guard defect underneath, which is the maintainer's call and not this PR's job.** A prose paragraph in AGENTS.md should not be able to collide with the invariant list at all — the home-side match should be anchored to the numbered invariant entries and to the whole name, not a prefix of any bold. Anchoring it:

```sh
if grep -qF "1. **$bare" "$home" || grep -qF ". **$bare." "$home"; then
```

passes on this branch with your original "Both steps file" wording, and still fails on a real restatement — planting `4. **Serde-first.** Every type crossing a crate boundary round-trips.` into `pipeline-as-plugins.md` under the anchored version gives `FAIL … restates invariant "Serde-first"`. So it drops the false positive without dropping the detection.

Either fix turns this PR green. The first is yours to take now; the second wants its own change, and nothing in `scripts/test-invariants.sh` pins this shape today, which is why it went unnoticed. Happy to file the guard issue if you'd rather not carry it.

---

PR #6357, opened 2026-09-07
fix(stella-cli): run the session's hooks on a deck worker lane

### LINKED ISSUE #6157: A deck worker lane runs no hooks and nobody decided that
## Problem

`crates/stella-cli/src/lane_capabilities.rs`'s `sub_session` writes
`hooks: None` for the deck worker lane. The comment beside it says the
session's hooks are the lead turn's, and then says wiring them onto a worker
lane "is its own decision". Nobody has made that decision, and nothing tracks
it.

So a deck worker lane runs no `PreToolUse`, `PostToolUse` or any other
lifecycle hook. An operator who wrote a hook to guard a shell command sees it
run on the lead turn and not on a worker lane's turns. That may be right. It
has never been written down.

`crates/stella-parity/src/lane/capability.rs` now carries this as a
`Deferred` row citing this issue, so the gap is visible instead of reading
like a settled decision.

## Files

- `crates/stella-cli/src/lane_capabilities.rs` — `sub_session`, the literal.
- `crates/stella-cli/src/subsession.rs` — `run_worker`, the call site. It
  would have to get a `&Hooks` and a `&dyn HookRunner` to pass one.
- `crates/stella-parity/src/lane/capability.rs` — the `SubSession` block's
  `hooks` row, to flip once the decision is made.

## Verify

Decide one of two things and write it down:

- A worker lane runs the session's hooks. Then `run_worker` takes the hooks
  and the runner, `sub_session` binds them, and the parity row becomes
  `Bound` with a witness that a hook fires on a worker turn.
- A worker lane does not. Then the parity row becomes `Declined` and its
  reason says why a hook that guards the session does not guard its workers.

## Definition of done

- [x] The decision is written in `lane_capabilities.rs`'s `sub_session`.
- [x] The `SubSession` `hooks` row in
      `crates/stella-parity/src/lane/capability.rs` is `Bound` or `Declined`,
      not `Deferred`.
- [x] A `Bound` answer names a test that proves a hook fires on a worker turn.
- [x] `cargo test -p stella-parity lane` and `cargo test -p stella-cli
      lane_capabilities` pass.

Refs #6150

macanderson:
## What & why

A deck worker lane (`BuiltinLane::SubSession`) ran no lifecycle hooks at
all: `lane_capabilities::sub_session` hardcoded `hooks: None` and its
comment said wiring hooks onto a worker lane "is its own decision" that
nobody had made. An operator's `PreToolUse` guard hook then only ever saw
the lead turn's tool calls, silently missing anything a worker lane did.

The decision, written down in `sub_session`'s doc comment: a worker lane
runs the session's hooks, the same as the lead turn (`lead`) and a
delegated sub-agent's `SubagentFork` lane already do. The operator's guard
applies to a tool call regardless of which lane made it.

- `sub_session` now takes `hooks: Option<&Hooks>` and `runner: &dyn
  HookRunner`, and binds `hooks: hooks.map(|hooks| (hooks, runner))` —
  identical shape to `lead`/`resume`/`fleet_attempt`/`goal_arc`.
- `subsession.rs`'s `run_worker` passes `cfg.hooks.as_ref()` and a
  `HostHookRunner` through, the same values `lead_turn.rs` hands to `lead`.
- `stella-parity`'s `SubSession` row for `hooks` moves from `Deferred`
  (citing this issue) to `Bound`, witnessed by the existing
  `each_lane_binds_what_its_call_site_bound_before` test.

Closes #6157

## The witness

`each_lane_binds_what_its_call_site_bound_before`
(`crates/stella-cli/src/lane_capabilities.rs`) gained the assertion that
`sub_session(Some(&hooks), ...).hooks.is_some()`. Proven both directions by
temporarily reverting `sub_session`'s literal back to `hooks: None` on this
branch and re-running:

- `cargo test -p stella-cli --bin stella lane_capabilities` → **FAILED** at
  `each_lane_binds_what_its_call_site_bound_before`, panic: "a worker lane
  must run the hooks its caller handed it".
- `cargo test -p stella-parity lane` → **FAILED** at
  `a_seam_claim_agrees_with_its_lane_literal`, panic: "`sub_session` claims
  it binds `hooks` as `hooks: hooks.map(`, and
  crates/stella-cli/src/lane_capabilities.rs writes no such text".

Reverting the temporary change back to this PR's diff makes both green
again (5/5 and 17/17 respectively).

## The gate

- [x] `cargo fmt --check` (ran `rustfmt --check` directly on the three
      changed files — clean)
- [ ] `cargo clippy --workspace --all-targets -- -D warnings` — scoped runs
      (`-p stella-cli --bin stella`, `-p stella-parity`) did not finish
      locally: this machine has many other concurrent agent sessions
      building the same shared `target/`, and the clippy invocations sat on
      the build lock for 45+ minutes without completing. `cargo check -p
      stella-cli` and `-p stella-parity` both compiled clean earlier in the
      session. Deferring to CI's `lint`/`lint-schema` step per AGENTS.md
      ("CI builds and tests; this laptop does not") — will fix promptly if
      it goes red.
- [x] `cargo test --workspace` — scoped: `cargo test -p stella-cli --bin
      stella lane_capabilities` (5/5) and `cargo test -p stella-parity lane`
      (17/17), both green
- [ ] Docs updated where behavior/flags changed (README, `--help`, doc
      comments) — no user-facing docs describe per-lane hook behavior; the
      doc comment on `sub_session` itself records the decision, which is
      what the issue's Definition of done asked for
- [x] `Closes #6157` appears **both** above and as a commit trailer

## Fix over file

A defect you saw while making this change is fixed in this PR, even when it
is off-topic. Name each extra fix below. File an issue only when a fix could
not ride this PR: it needs a maintainer call, a rig or spend, or more than
one session. File it only if fixing it moves a pillar: stability,
reliability, maintainability, innovation, efficiency, or performance
(AGENTS.md § "Fix over file"). The gate catches a `TODO` with no issue
number. It cannot catch what only you saw.

- [x] Extra fixes in this PR: **none** — no other defect surfaced while
      making this change
- [x] Filed, with the reason a fix could not ride this PR: **nothing was
      deferred**

## Ground-rule check

- [ ] New cross-boundary types round-trip through serde (test included) —
      no new types

This PR settles only the `hooks` seam. The `requery` seam on the same lane
(whether a worker may re-query the session's context plane) is still
`Deferred`, tracked separately as #6158 and out of scope here.

## Summary by Sourcery

Propagate session hooks to deck worker lanes so operator tool-use guards apply consistently across the session.

Bug Fixes:
- Run configured session lifecycle hooks for tool calls made by deck worker lanes, ensuring worker activity is covered by the same operator guards as lead and delegated lanes.

Enhancements:
- Move subsession steering and mid-turn routing code into a dedicated module while preserving the existing behavior.

Tests:
- Extend lane capability coverage to verify worker lanes bind hooks when provided and remain unbound when hooks are absent.

Chores:
- Update lane capability parity metadata to mark worker-lane hook binding as implemented.


macanderson:
Answering the one ❌ in Sourcery's assessment table, and flagging that #6157's third DoD box claims more than its witness shows.

**The row:** "provide a test demonstrating that a hook actually fires during a worker turn — the added test verifies only that hooks are present in the worker lane's capabilities."

**Sourcery is right about what the test shows.** `each_lane_binds_what_its_call_site_bound_before` asserts `sub_session(Some(&hooks), &runner, ...).hooks.is_some()`. That is a binding assertion. It does not run a worker turn and no hook executes in it. #6157's third DoD box is ticked with that test named as proof "a hook fires on a worker turn", and that wording overstates it — the box should not be read as an end-to-end demonstration.

**What is actually proven today, in two halves:**

- *Binding* — this PR's witness, that `sub_session` puts the session's hooks and a `HostHookRunner` into the lane's `TurnCapabilities`. `stella-parity`'s `a_seam_claim_agrees_with_its_lane_literal` holds the ledger row to the literal, so the claim cannot drift from the code.
- *Firing* — `stella-core`'s driver hook tests (`crates/stella-core/src/driver/tests/user_hooks.rs`). A worker lane runs an ordinary `Engine` turn, so there is nothing worker-specific about whether a bound hook fires; that is the engine's contract and the engine tests it.

**No lane in the ledger proves both halves in one test.** Worth stating plainly, because the `SubagentFork` row looks like the counterexample and is not: its witness, `subagent_start_and_stop_hooks_fire_around_a_child_turn`, lives in `stella-core` and builds `TurnCapabilities { hooks: Some((&hooks, &runner)) }` by hand. It never calls a lane constructor, so it proves the engine fires subagent hooks, not that the `SubagentFork` lane binds them. Every one of the CLI lanes cites the same binding witness (`CLI_SEAMS`) for `hooks`. This PR is at parity with them, not below it.

**What I did not do, and why it is the maintainer's call.** A genuine end-to-end test would drive `run_worker` with a real `Engine`, a scripted provider and a store, on a thread, and assert a `RecordingHookRunner` saw a `PreToolUse` payload. That rig does not exist in `stella-cli` today and would be the only one of its kind; building it is a change to the `CLI_SEAMS` convention that would properly apply to every lane, not just this one. That is larger than this PR and I am not going to decide it silently — per CLAUDE.md, when "now" and "right" cannot both be had, the choice goes to a human.

So, concretely, two options:

1. **Merge as is** and correct #6157's third DoD box to say what the witness proves — that the lane binds the seam, with firing covered by the engine's own tests. This is the honest version of the status quo and keeps the lanes consistent.
2. **Hold this PR** for a `run_worker` bind-and-fire rig, and expect the same rig to be owed for the other lanes.

I have not ticked, unticked, or edited anything on #6157 — flagging the gap rather than papering over it. Say which you want and I will do it.

---

macanderson:
Heads-up: I pushed `43b5eb8` to this branch — you may want to pull before your next edit to `crates/stella-cli/src/subsession/steering.rs`.

`fmt + clippy + test` went red on `62f0bca` at its `cargo doc -D warnings` step (run [34078529084](https://github.com/macanderson/stella/actions/runs/34078529084/job/101610118906), step 49). Every other step in that job passed, including `cargo test` and the prose gate the previous head was red on.

The cause: the subsession split moved `SteeringTap` into `subsession/steering.rs`, where `SubSessions` is no longer in scope, so `SteeringTap`'s doc comment left an unresolvable intra-doc link:

```
error: unresolved link to `SubSessions::steer`
  --> crates/stella-cli/src/subsession/steering.rs:14:7
   |
14 | /// [`SubSessions::steer`] feeds a worker's tap from the driver thread, and
   |       ^^^^^^^^^^^^^^^^^^ no item named `SubSessions` in scope
```

Pointed it at `super::SubSessions::steer` rather than dropping it to plain backticks like the `SubSessions::stop` reference three lines below — a path link breaks loudly on a rename, backticks rot in silence. Happy to take the backtick form instead if you'd rather keep the two references symmetrical.

Verified both directions on the pinned 1.97.0 toolchain, scoped to the one crate: `RUSTDOCFLAGS="-D warnings" cargo doc -p stella-cli --no-deps --document-private-items` exits 101 before the change with CI's exact error and 0 after. `cargo fmt --check -p stella-cli`, `check-prose.py` and `check-line-citations.py` all clean.

---

PR #6361, opened 2026-09-07
fix(stella-learn,stella-cli): key the appraisal ledger and demotion lineage by (kind, id)

### LINKED ISSUE #6134: Self-driving loop makes git worktrees without the fleet worktree lock

macanderson:
## What & why

`crates/stella-learn/src/skills/appraisal.rs`'s `SkillAppraisal` was keyed by
a bare `skill: String`, and everything downstream in
`crates/stella-cli/src/memory/appraisals.rs` — `APPRAISALS_FILE`,
`latest_verdicts`, `consecutive_negative_appraisals`, `record_demotion`,
`demoted_skills` — read and wrote that same bare id. #6093 already keyed the
*input* side (the trial ledger, `ArtifactTrial`) by `(ArtifactKind, id)`
because a memory record, a mined rule, and a skill can all be appraised now.
The output side was still shaped for skills alone: two artifacts of
different kinds sharing an id would collapse into one appraisal row and one
demotion lineage.

This is step 1 from the issue (the re-key). Step 2 (renaming
`SkillAppraisal`/`SkillTrial`/etc. to say "artifact" rather than "skill") is
a separate, wide, mechanical pass across `stella-learn` and `stella-cli`,
deliberately not done here so the re-key is reviewable on its own — the
issue itself calls this split out.

- `SkillAppraisal` now carries `ArtifactKind` beside its id, with the same
  `#[serde(default = "skill_kind")]` alias trick `ArtifactTrial` already
  uses, so a row a build before this change wrote still reads as the skill
  row it was.
- `appraise()` takes a `kind: ArtifactKind` and stamps it onto the returned
  `SkillAppraisal`.
- `latest_verdicts`, `consecutive_negative_appraisals`, `record_demotion`
  and `demoted_skills` each take a `kind` and scope to it.
- The demotion lineage prefix is now built from the kind
  (`format!("{}:", kind.as_str())`) instead of a hardcoded `"skill:"`
  constant — `ArtifactKind::Skill.as_str()` is `"skill"`, so an existing
  workspace's skill demotion history reads under the identical prefix.

Closes #6134

## The witness

passes on the new**.

`crates/stella-cli/src/memory/appraisals/tests.rs::two_appraisals_under_one_id_do_not_collide`
records a `Helps` skill appraisal and a `Harms` memory appraisal under the
same id `"shared"`, then asserts `latest_verdicts`,
`consecutive_negative_appraisals`, `record_demotion` and `demoted_skills`
each read/write only their own kind's row. On `main` (pre-`kind` field) this
does not compile — `SkillAppraisal` has no `kind` field and none of the five
reader/writer functions take one — so the test is a genuine fail→pass
witness for the API shape the issue asks for. A second test,
`a_skill_demotion_written_before_kinds_existed_still_folds`, is the
backward-compatibility half: a `skill:<name>` demotion event written by hand
(simulating a pre-change ledger) still folds correctly through
`demoted_skills(&store, ArtifactKind::Skill)`.

Verified locally (stash the two changed crates, re-run, restore):
- `cargo test -p stella-learn` — 246/246 pass, including 17
  `skills::appraisal::tests::*` covering the new `appraise(kind, …)` shape.
- `cargo check -p stella-cli` — production code compiles clean.
- `cargo fmt --check -p stella-learn -p stella-cli` — clean.
- `python3 scripts/check-prose.py` — OK, no new content-free prose or
  reading-grade regression.

`cargo test -p stella-cli --bin stella memory::appraisals` was still
compiling stella-cli's full dependency graph under the `test` cfg when this
PR was opened — the machine this session runs on was under load average
44-57 on 10 cores from many concurrent sibling sessions, and an earlier
identical run compiled every dependency stella-cli has with zero errors
before being interrupted, reaching `Compiling stella-cli` itself. Watching
CI for the actual test-suite confirmation.

## The gate

- [x] `cargo fmt --check`
- [ ] the clippy gate — scoped locally to `-p stella-learn -p stella-cli`;
      CI runs the full gate
- [ ] the test gate — scoped locally to `-p stella-learn` (full pass) and
      `-p stella-cli` (still compiling under heavy shared load when this PR
      opened); CI runs the full gate
- [ ] Docs updated where behavior/flags changed (README, `--help`, doc
      comments) — no user-facing flag or doc changed; module and function
      doc comments updated in place for the new `kind` parameter
- [ ] CLA signed (the bot prompts on your first PR — nothing to do per
      commit)
- [x] `Closes #6134` appears **both** above and as a commit trailer

## Fix over file

- [x] Extra fixes in this PR: `crates/stella-cli/src/memory/rule_efficacy.rs`'s
      module doc claimed a rule id "would clash with a skill of the same
      name" in `record_demotion`/`latest_verdicts` — true before this PR,
      false after. Corrected the doc in the same PR since it is the exact
      claim this PR falsifies. — **or** none
- [ ] Filed, with the reason a fix could not ride this PR: #___ — **or**
      nothing was deferred

## Ground-rule check

- [x] New cross-boundary types round-trip through serde (test included) —
      `SkillAppraisal`'s new `kind` field is exercised by the existing
      round-trip test plus the two new witness tests

The rename from "skill" to "artifact" vocabulary (`SkillTrial`,
`SkillAppraisal`, `SkillVerdict`, `WITH_SKILL_ARM`/`WITHOUT_SKILL_ARM`) is
deliberately **not** in this PR — the issue calls it out as a second,
mechanical pass that should be regenerated on a fresh `main` rather than
carried on a stale branch. This PR only re-keys.

## Summary by Sourcery

Scope appraisal and demotion state by artifact kind and id while retaining compatibility with legacy skill records.

Bug Fixes:
- Key appraisals and demotion lineages by artifact kind and id so artifacts of different kinds sharing an id no longer collide.
- Preserve compatibility with existing appraisal rows and skill demotion events written before artifact kinds were added.

Enhancements:
- Propagate artifact kind through appraisal, verdict, hysteresis, and demotion APIs and callers.
- Correct rule-efficacy documentation to reflect kind-scoped appraisal and demotion behavior.

Tests:
- Add coverage proving same-id appraisals and demotions remain isolated by artifact kind.
- Add backward-compatibility coverage for legacy skill demotion lineages.

### DIFF
--- a/crates/stella-cli/src/memory.rs
+++ b/crates/stella-cli/src/memory.rs
@@ -769,7 +769,8 @@ impl SessionMemory {
             self.include_workspace_skills,
         )
         .skills;
-        let demoted = appraisals::demoted_skills(&self.store);
+        let demoted =
+            appraisals::demoted_skills(&self.store, stella_learn::ledger::ArtifactKind::Skill);
         if !demoted.is_empty() {
             skills.retain(|s| !demoted.contains(&s.name));
         }
--- a/crates/stella-cli/src/memory/appraisals.rs
+++ b/crates/stella-cli/src/memory/appraisals.rs
@@ -35,6 +35,16 @@
 //! workspace keeps the window it already paid for. Nothing rewrites that
 //! file. An append-only ledger is not edited; it simply stops growing.
 //!
+//! # The output ledgers use the same key
+//!
+//! [`APPRAISALS_FILE`] and the demotion lineage carry
+//! [`stella_learn::ledger::ArtifactKind`] beside the id, the same as
+//! [`TRIALS_FILE`]. [`latest_verdicts`], [`consecutive_negative_appraisals`],
+//! [`record_demotion`] and [`demoted_skills`] each take a `kind` and read
+//! only that kind's rows. A memory record and a skill can share an id and
+//! stay two rows, not one. An old row with no kind still reads as a skill
+//! row — the same default [`TRIALS_FILE`] uses.
+//!
 //! # How the loop closes
 //!
 //! Every piece here has a production caller:
@@ -104,10 +114,15 @@ pub const LEGACY_TRIALS_FILE: &str = "skill_trials.jsonl";
 /// the delta between them is the skill's own.
 pub const LIVE_WINDOW_TASK: &str = "live-window";
 
-/// The `proposal_lineage_id` prefix a skill's demotion events are filed
-/// under, so the fold in [`demoted_skills`] can tell them from directive
-/// events in the same ledger.
-const SKILL_LINEAGE_PREFIX: &str = "skill:";
+/// The `proposal_lineage_id` prefix `kind`'s demotion events are filed under.
+/// [`demoted_skills`] uses it to tell one kind's lineage from another's, and
+/// from a directive event, in the same ledger.
+///
+/// Built from [`ArtifactKind::as_str`], the same word the trial ledger
+/// writes. A skill row keeps the plain `skill:` prefix an old build wrote.
+fn lineage_prefix(kind: ArtifactKind) -> String {
+    format!("{}:", kind.as_str())
+}
 
 /// One queued candidate: enough to appraise and then render it, without
 /// re-mining.
@@ -129,15 +144,20 @@ pub struct QueuedCandidate {
     pub body: Option<String>,
 }
 
-/// The newest verdict per skill, as the creation gate reads it.
+/// The newest verdict per id of `kind`, as the creation gate reads it.
 ///
-/// Later lines win: an appraisal is a measurement at a point in time, and a
-/// skill that stopped helping must not be held up by the run that said it did.
-/// A missing or unreadable ledger yields an empty map, which the gate reads as
-/// [`EvalEvidence::Unevaluated`] — the honest answer when nobody has looked.
-pub fn latest_verdicts(workspace_root: &Path) -> HashMap<String, EvalEvidence> {
+/// Later lines win: an appraisal is a measurement at a point in time, and an
+/// artifact that stopped helping must not be held up by the run that said it
+/// did. `kind` keeps a memory and a skill sharing an id as two verdicts, not
+/// one folded together. A missing or unreadable ledger yields an empty map,
+/// which the gate reads as [`EvalEvidence::Unevaluated`] — the honest answer
+/// when nobody has looked.
+pub fn latest_verdicts(workspace_root: &Path, kind: ArtifactKind) -> HashMap<String, EvalEvidence> {
     let mut verdicts = HashMap::new();
     for appraisal in read_jsonl::<SkillAppraisal>(&path(workspace_root, APPRAISALS_FILE)) {
+        if appraisal.kind != kind {
+            continue;
+        }
         verdicts.insert(
             appraisal.skill.clone(),
             EvalEvidence::from_verdict(&appraisal.verdict),
@@ -287,7 +307,7 @@ pub fn sweep(
     let mut out = Vec::new();
     for id in ids {
         let trials = by_id.remove(&id).unwrap_or_default();
-        let appraisal = appraise(&id, &trials, config);
+        let appraisal = appraise(kind, &id, &trials, config);
         let origin = origins
             .get(&id)
             .copied()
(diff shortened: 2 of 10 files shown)


macanderson:
Answering Sourcery's ❌ row ("Rename the shared appraisal engine vocabulary from skill-specific names to artifact-oriented names"): deliberately out of scope for this PR, per the issue body itself — `#6134` splits the work into two steps and says the first (the re-key, done here) is "worth doing without the second" (the rename), and that doing them together would cost more than doing them separately since "the rename touches every line the re-key does" and "a reviewer should be able to read the re-key on its own." The issue also says the rename should be regenerated on a fresh `main` rather than carried on a stale branch, since it is a wide mechanical pass across `stella-learn` and `stella-cli`. Not deferred to a filed issue because the original issue #6134 itself already is that tracking — its own DoD checklist (satisfied by this PR) does not include the rename.

PR #6366, opened 2026-09-07
feat(scripts): join main-canary's compile check with ci.yml's test verdict

### LINKED ISSUE #5850: A test-only failure on `main` files no `main-red` issue and blocks no merge — the canary asks compilation, `ci` asks the suite, and nothing joins them
## Problem

`main` went red three times between 2026-09-03 and 2026-09-04 on one test
(`credential_handoff::tests::cloexec_pipe_closes_a_concurrent_childs_inherited_copy`),
and **not one of the three filed a `main-red` issue or blocked a single merge.**
Every surface read green while the last verified state of `main` was a failure.

That is not a bug in any one workflow. It is a seam between two of them.

## The seam, verified at source

| workflow | what it asks | what it does about the answer |
|---|---|---|
| `ci.yml` | the full suite (`cargo test`) | reports pass/fail on the commit. **Files nothing. Blocks nothing.** |
| `main-canary.yml` | compilation + composition | files a labelled `main-red` issue, which `main-red-hold.yml` then blocks merges on |

`scripts/main-canary.sh` runs `cargo check --workspace --all-targets --locked`
and `cargo metadata` — it contains **zero** occurrences of `cargo test`:

```
$ grep -c "cargo test" scripts/main-canary.sh
0
$ grep -nE "cargo (check|metadata)" scripts/main-canary.sh
203:  "compile|cargo check --workspace --all-targets --locked…"
252:cargo metadata --format-version 1 >/dev/null
266:cargo check --workspace --all-targets --locked
```

So the workflow that **detects and holds** cannot see a test failure, and the
workflow that **sees** a test failure neither files nor holds. A tree that
compiles but fails its suite is invisible to the entire main-red chain.

This is by the canary's own stated contract, not an oversight in its
implementation — `scripts/main-canary.sh`'s header argues the compilation and
composition case deliberately. The question this issue raises is whether that
contract is still the right one.

## Why this is not #5817 and not `check-main-verified.sh`

`main-canary.yml`'s second step, `scripts/check-main-verified.sh`, asks whether
every recent commit on `main` has a **completed `ci` run at all**. That is a
different question — the absence of an answer, not a failing one — and it is
#5817's subject (a `[skip release]` version-sync merge landing with zero `ci`
runs). A commit whose `ci` run completed and **failed** is "verified" by that
script's definition and correctly passes it.

Both gaps were live simultaneously on 2026-09-04: `main`'s newest verified state
was red at `f0201f7`, and its actual tip `4de19a8` had no `ci` run at all.

## Evidence — the three that got through

| commit | `ci` verdict | `main-red` issue filed | merges blocked |
|---|---|---|---|
| `b29e934f` (#5793) | [33796308644](https://github.com/macanderson/stella/actions/runs/33796308644) — **failure** | none | none |
| `f0201f75` (#5794) | [33821298562](https://github.com/macanderson/stella/actions/runs/33821298562) — **failure** | none | none |
| `b2a074dd` → the head after it | failure on the following commit | none | none |

Between them, PRs continued to merge onto a `main` whose suite was failing.
AGENTS.md already records the shape this produces when the chain *does* fire
(#3917: four PRs merged onto a non-compiling tree in 35 minutes, the first
twelve seconds after the canary filed). Here the chain never fired at all.

## Why it stayed unfiled until now

#5806 named this gap in prose, inside an issue whose subject was the *test*
defect. That issue was closed as a duplicate of #5812 when the test was fixed,
and the canary observation was closed with it — a finding that lived only inside
an issue closed for another reason. Three later issues (#5812, #5813, #5816's PR
description) each restated it in their residue sections and each pointed at
#5806, which by then was closed.

## What to decide

This is a judgement about post-merge cost, which is why the three red-`main`
repairs deliberately did not touch it — AGENTS.md is explicit that a red-`main`
repair edits only what it must, and those PRs were each racing every other merge.

Options, in rough order of cost:

1. **Have `ci.yml` file the `main-red` issue itself on a failed `main` push.**
   Smallest change; puts filing next to the run that already knows the answer.
   Needs care that a cancelled/`startup_failure` run does not file (the canary's
   existing "only when its own job runs and fails" discipline).
2. **Widen the canary to run the suite.** Most faithful to "is `main` broken",
   and the most expensive — it doubles a full workspace test run per push.
3. **A third, cheap step that reads the latest `ci` conclusion for the commit**
   and files on failure, leaving both existing contracts untouched.

Whoever picks this owns `scripts/main-canary.sh`; the decision belongs with them
rather than with a sweep.

**Resolved as option 3**, in #6366: `scripts/check-ci-tests.sh` reads the
CONCLUSION of the most recent completed `ci.yml` run for `main` and is wired
into `main-canary.sh`'s existing `checks` array as a new `ci-tests` row,
rather than into `ci.yml` or a fourth workflow step — it reuses
`main-canary.sh`'s single `main-red` issue lifecycle (open/comment/close by
label) instead of adding a second actor that could race it to open or close
that issue, and it does not run the suite a second time.

## Files

- `scripts/main-canary.sh` — the compilation/composition contract and its header argument.
- `.github/workflows/main-canary.yml` — the two steps (`--announce`, then `check-main-verified.sh`).
- `.github/workflows/main-red-hold.yml` — the consumer that blocks on an open `main-red` issue.
- `.github/workflows/ci.yml` — where the suite actually runs.
- `scripts/check-main-verified.sh` — the adjacent "no answer at all" check (#5817).

## Definition of done

- [x] A push to `main` whose `ci` run fails **on a test** results in an open
      `main-red`-labelled issue, without a human filing it.
      Evidence: `scripts/check-ci-tests.sh` (new) reads `ci.yml`'s last
      completed conclusion for `main` and exits 1 on `failure`, which
      `main-canary.sh`'s `checks` array turns into the same file/comment
      flow as every other row. Witnessed by
      `scripts/test-main-canary.sh`'s `"a ci-tests failure opens an issue"`
      case (asserts `gh issue create` fires under `--announce --dry-run`)
      and `scripts/test-ci-tests.sh`'s `"a completed FAILING run is reported
      red"` case. Both green on PR #6366's branch.
- [x] `main-red-hold.yml` then blocks merges on that issue, as it already does
      for a compilation break.
      Evidence: no change needed — `scripts/check-main-red-hold.sh` blocks on
      any open issue carrying the `main-red` label
      (`gh issue list --label "$label" --state open`), regardless of which
      `main-canary.sh` row caused it to be filed. The new `ci-tests` row
      reuses that exact label.
- [x] A cancelled run, a `startup_failure`, or a run that was never created does
      **not** file — the canary's existing fail-open discipline is preserved, so
      this can never be what blocks a repair.
      Evidence: `scripts/test-ci-tests.sh` cases `"a cancelled run is not a
      suite failure"`, `"a startup_failure is not a suite failure"`, and
      `"no completed run yet is not a suite failure"` (all exit 0). Wiring
      through `main-canary.sh` proven by `test-main-canary.sh`'s
      `"a cancelled ci run does not fail the canary"`,
      `"a startup_failure does not fail the canary"`,
      `"no completed run yet does not fail the canary"`, and
      `"and none of the three opens an issue"`. All green on PR #6366.
- [x] The `unblocks-main` escape hatch still works on the resulting issue.
      Evidence: unaffected by this change — `check-main-red-hold.sh` reads
      the PR's own labels for `unblocks-main` independently of which check
      filed the open `main-red` issue; no code path here touches that logic.
- [x] Whichever workflow gains the behaviour has a hermetic self-test proving it
      files on a test failure and does not file on the three non-answers above
      (`guard-self-tests.yml` is where the canary's existing self-tests run).
      Evidence: `scripts/test-ci-tests.sh` (13 cases, hermetic, no network)
      added to `guard-self-tests.yml`'s `self-tests` job; the `ci-tests`
      integration cases above added to `scripts/test-main-canary.sh`, which
      already runs in `guard-self-tests.yml`'s `canary-self-tests` job.
- [x] `scripts/main-canary.sh`'s header, and AGENTS.md's canary paragraphs, say
      which workflow now answers which question — the current text describes the
      compilation-only contract as intentional and would be stale.
      Evidence: `scripts/main-canary.sh`'s new "Which question each workflow
      answers" section and `AGENTS.md`'s canary paragraphs (the
      `main-canary.yml` section) both state the split — `ci.yml` runs the
      suite, `compile` asks whether the tree builds, `ci-tests` reads what
      `ci.yml` already found.

## Prior art

- #5806 — named this gap; closed as a duplicate of the test defect it was filed alongside.
- #5812 / #5813 — the test defect, now fixed and verified green on `main` at `6d5fb8f`.
- #5817 — the adjacent gap: commits landing with no `ci` run at all.
- #3917 — what happens once `main` is red and merges continue.
- #5027 — the Actions outage that motivated `check-main-verified.sh`, and the
  reasoning for why these checks fail open.

macanderson:
## What & why

`ci.yml` runs the whole test suite on every push to `main` and used to keep
the result to itself — it reported pass or fail on the commit and neither
filed a `main-red` issue nor blocked a merge. `main-canary.sh` asks whether
the merged tree still composes (locks, size ceilings, compiles, prose) but
never asked whether the suite passed, because running it again would double
the cost the canary's own header argues against paying twice. Between
2026-09-03 and 2026-09-04 `main` failed the same test three times running and
the whole `main-red` chain never fired once.

This adds `scripts/check-ci-tests.sh`, wired into `main-canary.sh`'s existing
`checks` array as a new `ci-tests` row. It does not run the suite again — it
reads the CONCLUSION of the most recent completed `ci.yml` run for `main`,
and fails open on the three non-answers the issue's Definition of Done names
(a still-queued run, a cancelled one, a `startup_failure`), leaving those to
`check-main-verified.sh`. Landing it as a row of the canary's existing checks
array means a green suite closes the same `main-red` issue a red one opened,
reusing the single-issue-lifecycle code already there, rather than adding a
second workflow that could race the canary to open or close the issue.

Closes #5850

## The witness

Verified directly: `git stash` on `scripts/main-canary.sh` alone, then ran the
new `ci-tests` integration cases in `scripts/test-main-canary.sh` — all 10
failed with `unknown argument '--fixture-ci-commits'` (the row and its flags
did not exist). Restored the change: all 45 cases pass, `check-ci-tests.sh`'s
own hermetic suite (`scripts/test-ci-tests.sh`) passes 13/13.

## The gate

- [x] `cargo fmt --check` — via `make guards-fast`, green (no Rust touched)
- [ ] `cargo clippy --workspace --all-targets -- -D warnings` — no Rust changed; nothing to lint
- [ ] `cargo test --workspace` — no Rust changed; the new/changed shell scripts have their own hermetic suites instead (`make ci-tests-test`, `make main-canary-test`), both green locally and wired into `guard-self-tests.yml`
- [x] Docs updated where behavior/flags changed (README, `--help`, doc comments) — `AGENTS.md`'s canary paragraphs and `scripts/main-canary.sh`'s own header now say which workflow answers which question

## Fix over file

No unrelated defects were noticed while making this change. Nothing was
deferred to a new issue.

- [x] Extra fixes in this PR: none — **or** none

## Ground-rule check

- [x] No I/O added to `stella-core`; no new deps without justification below — pure shell, no dependency added
- [x] No new outbound network calls (Stella never phones home) — `check-ci-tests.sh` reads `ci.yml`'s own already-public run history via `gh run list`, the same API `check-main-verified.sh` already calls in the same workflow

- The new `ci-tests` row does not affect `main-canary.sh`'s existing four
  composition checks — a fixture is threaded to it as exported environment
  variables (`CHECK_CI_TESTS_FIXTURE_COMMITS`/`_RUNS`), not interpolated into
  the row's `eval`'d command string, because the checks array's string
  interpolation cannot safely carry a fixture's embedded newlines.
- `--manifest-dir` alone (the shape every existing hermetic test case in
  `test-main-canary.sh` already uses) now also makes the `ci-tests` row
  offline — without that, adding a row that reads live `gh run list` output
  would have made every pre-existing green fixture case in that suite
  network-dependent and flaky. Pinned by its own test case.
- Both new files (`check-ci-tests.sh`, `test-ci-tests.sh`) are held to this
  repository's strict 6th-grade reading-level ceiling for new files; measured
  4.74 and 4.91 respectively via `check-prose.py`'s own grader.

## Summary by Sourcery

Join main-canary with ci.yml's existing test verdict so failed main-branch suites trigger the established main-red remediation flow without duplicating test execution.

New Features:
- Add a canary check that reports failures from the most recent completed main-branch ci.yml test run without rerunning the suite.

Bug Fixes:
- Ensure failed CI test verdicts participate in the existing main-red issue lifecycle so they can open and close remediation issues.

Enhancements:
- Treat queued, cancelled, timed-out, startup-failure, missing, and unknown CI results as non-conclusive rather than definite suite failures.
- Keep the new CI verdict check hermetic when main-canary runs with manifest fixtures and expose it through Makefile targets.

CI:
- Run the new ci-tests verdict-reader suite in guard-self-tests and expand main-canary integration coverage for CI verdict wiring and fail-open behavior.

Documentation:
- Document how ci.yml, main-canary, and main verification divide responsibility for test and composition checks.

Tests:
- Add hermetic coverage for completed pass/failure results, non-conclusive run states, commit-history selection, fixture handling, and canary issue lifecycle integration.


macanderson:
**Correcting a commit message I pushed here: `c50ed3a` is wrong, and its wording should not reach `main`.**

An automated PR sweep (me) saw this PR sitting with only `cla` and `Sourcery review` and no `pull_request`-triggered checks at all, and pushed the empty commit `c50ed3a` to re-request the event. Its message blames "the platform gap AGENTS.md records" — GitHub failing to deliver the `pull_request` event. **That diagnosis is wrong.**

The actual cause was the merge conflict with `main` (AGENTS.md, against #6356). A `pull_request` workflow runs on the merge ref, and GitHub cannot compute one for a conflicted PR — so no run is ever created, and the PR reads as "CI never started" rather than "CI is red". The evidence that settles it: this PR had 2 check runs while conflicted, and 21 within seconds of `055f442` landing the merge. The empty commit changed nothing, which is itself the disconfirming evidence — a genuinely dropped event would have been re-requested by that push.

Two asks:

1. **Squash-merge with a hand-edited commit body that drops `c50ed3a`'s message.** Squash composes the body from commit messages, so as written that wrong claim lands in `main`'s history, where the next session to hit a conflicted PR would read it and go looking for a platform gap.
2. Nothing else here needs action from me — you resolved the conflict and the prose grade in `055f442`/`0d981c8` before I could push my own copy of the same fix, and yours was the better resolution.

Apologies for the noise on your branch. Worth recording that "no checks at all" and "checks failing" are different states with different causes, and the first one reads as an Actions problem when it is usually a merge conflict.

---

PR #6369, opened 2026-09-07
feat(stella-runtime): carry a wrapper round's every claim on the stored record

### LINKED ISSUE #5883: Nothing builds a ladder snapshot from a wrapper run, so no verdict stamp is ever stored

macanderson:
## What & why

`VerdictStamp` is the wire record of who claimed what about a verdict, and
`LadderSnapshot::stamps` is where it lands. A wrapper round already minted
one — the arbiter's own, from `wrapper::stamp::stamped`. Every other claim
the round collected never reached the record: a member that broke at a
point, one the clock cut short, one that looked and could not tell, each
became a row on `DispatchReport::arbitration` and lived there for exactly
as long as the process. `ArbiterClaim::into_stamp`, the function that turns
one of those into a wire record, had no caller outside its own tests.

So a stored verdict could not say whether a plugin's arbiter answered,
abstained, or was never heard from — the failure the stamp exists to
prevent, one layer out.

Two changes:

1. **`stella-runtime`** — `wrapper::stamp::stamp_claims` mints one stamp
   per aside claim through `ArbiterClaim::into_stamp` and puts them ahead
   of the arbiter's own, so the record's stamps and the fold's rows are one
   sequence in one order. `WrapperDispatch::stamp_round` reads them off the
   same `faults` list the fold is built from, so neither can name a gap the
   other does not. Every stamp on a record shares one hash, because
   `LadderSnapshot::stamp_preimage` drops the stamp list; `preimage_hash`
   is now the single place that digest is taken (ADR 0004's `record_hash`),
   so two producers cannot disagree about the preimage and mint hashes that
   read as tampering.
2. **`stella-cli`** — the `stella run --pipeline <plugin>` door sent the
   gate board onto the run's event stream and nothing else, so the stamped
   record never reached the journal at all. `wrapper_plugin::report`'s new
   `run_events` puts the verdict ahead of the board rendered from it. A
   pure function, in the module that exists so this door's output is
   assertable without a session behind it.

The issue also asks for the `LadderSnapshot` producer itself and the
`Verdict` → `LadderRung` mapping. Both landed in #5873 while this issue was
open (`crates/stella-runtime/src/wrapper/stamp.rs`'s `snapshot` and `rung`),
and this PR builds on them rather than repeating them — the issue's
checklist cites that work where it is already done.

Closes #5883

## The witness

`a_plugin_that_dies_leaves_its_silence_on_the_record`
(`crates/stella-runtime/tests/wrapper_verdict_stamp.rs`) drives a whole
dispatch whose plugin reads the `after_turn` request and exits without
answering, then asserts the stored snapshot's first stamp names that plugin,
reads `inconclusive`, says which check fell silent, and carries a
`preimage_hash` that re-derives from the record it sits on.

Proved by ablation rather than asserted: with the two `stella-runtime`
source files stashed, that test fails with `left: "engine"` /
`right: "witness-v1"` — the old record holds only the host's own stamp, so
the plugin's silence is nowhere in it. With them restored, all five tests in
the file pass.

Two more: `a_claim_that_did_not_decide_is_stamped_ahead_of_the_one_that_did`
and `no_claims_leaves_the_record_alone` (unit tests in
`wrapper/stamp.rs`) pin the ordering, the shared preimage, and that a round
nothing stood aside on is byte for byte the record it was.
`a_finished_run_puts_the_verdict_on_the_stream_before_the_board`
(`wrapper_plugin/tests/report.rs`) is the CLI half's witness: `run_events`
emits `verdict` then `gate_board`, and the verdict carries the round's
ladder snapshot.

## The gate

- [ ] `cargo fmt --check`
- [ ] clippy over the workspace at `-D warnings`
- [ ] the workspace test suite
- [x] CLA signed (the bot prompts on your first PR — nothing to do per commit)

The three unticked boxes are CI's, per CLAUDE.md's "CI builds and tests;
this laptop does not". Run here, scoped to the crates this PR touches:
`cargo fmt`, the `wrapper_verdict_stamp` integration suite, the
`wrapper::stamp` unit suite, the `wrapper_plugin::tests::report` suite,
`make guards-fast`, and `python3 scripts/check-prose.py`. The evidence for
the unticked boxes is this PR's CI run.

## Fix over file

- [x] Extra fixes in this PR: **none** — the CLI half is not a separate fix;
      a record nothing emits is the same defect one layer out, and the issue
      names it (step 4, "Emit or store the stamped snapshot so a replay can
      read it").
- [x] Filed, with the reason a fix could not ride this PR: **nothing was deferred**

## Ground-rule check

- [x] New cross-boundary types round-trip through serde — no new type; every
      stamp minted here is a `stella_protocol::VerdictStamp`, whose
      round-trip test already stands in `ladder/stamp.rs`

**One place the record and the fold deliberately differ, and the test says
so.** A fold row names the arbiter whose rule was applied; a stamp names who
*observed* the evidence. When the only plugin dies, no plugin reported
anything, so the host judges `EvidenceSet::unobserved()` itself and signs
that last stamp `engine` (`wrapper::stamp::author`'s rule, unchanged here).
The witness compares the aside claims position for position and leaves that
last pair out, naming the reason — collapsing the two would have meant
changing what `author` means to make a test line up.

**`stamp_round` reads `faults.claims` rather than taking them as an
argument.** It was a parameter first; that made the method eight arguments
wide and gave the caller two lists to keep in step. Reading the one list
both the stamp and the fold are built from is what makes them unable to
disagree.

**No wire change.** `LadderSnapshot` and `VerdictStamp` are untouched, so
`docs/wire/` needs no regeneration; what changed is how many stamps a
producer puts on a record it was always allowed to put several on.

## Summary by Sourcery

Carry all wrapper-round claims into durable verdict records and publish those records on the CLI run stream before the gate board.

New Features:
- Persist every wrapper-round arbitration claim on the ladder snapshot, including inconclusive, failed, and timed-out checks.
- Emit the completed verdict, including its ladder snapshot, to the CLI run event stream before the gate board.

Bug Fixes:
- Ensure stored verdicts preserve which checks answered, abstained, or were never heard from instead of retaining only the host arbiter's claim.

Enhancements:
- Use a shared record preimage hash for all stamps and keep stamp ordering aligned with arbitration rows.

Documentation:
- Update runtime documentation to describe per-claim verdict stamps and the expanded stamping witnesses.

Tests:
- Add runtime and CLI witness tests covering claim ordering, shared hashes, no-claim behavior, silent plugin handling, and verdict-before-board event ordering.


macanderson:
## The red `windows-check` here is not this PR's — fix is in #6371

`the platform split still compiles on Windows` failed on `a5450d1`, at `stella-tools`' `a_timed_out_hook_leaves_no_surviving_grandchild`:

```
```

**Why it is not yours.** This PR touches `crates/stella-cli` and `crates/stella-runtime` only. The failing test is in `crates/stella-tools`, which both of those *depend on* — nothing in this diff can reach it. `windows-check` is green on `main`'s recent pushes and on every other branch today; the one other red run in its history (`fix/6292-runtime-batch`, 22:50) was an unrelated clippy `too many arguments`, so there is no flake precedent for this test either. It went red for the first time here, on a diff that cannot touch it.

**What is actually wrong.** The three witnesses in that file are not symmetrical. Witnesses 1 and 2 call `started_beating` and only then kill, so spawn time cannot race them. Witness 3 cannot do that — the hook's own 300ms `timeout_ms` is what kills — so its grandchild has to reach the disk inside 300ms. On Windows the hook resolves a shell and then creates two processes, which does not reliably fit; the `fork`/`exec` that number was chosen against does. It fails its own precondition, not the property.

**#6371** raises that one budget to 3s and explains it in place. The witness keeps its meaning — `backgrounding_command` ends in `sleep 30`, so the hook times out at any budget; only the grandchild's head start grows. `cargo test -p stella-tools --test group_kill_witnesses` is 3/3 on Linux there.

Nothing for you to do. `windows-check` is not a required check, so it is not blocking this PR, and once #6371 lands a re-run here goes green with no change to your diff. I could not re-run the job myself to confirm the flake — `rerun-failed-jobs` returns 403 for this token — which is why the fix was written rather than the job retried.

Your other checks are clean as of this comment; the remaining ones are still in flight.

---

macanderson:
`the platform split still compiles on Windows` is red here, and it is not this PR's failure. Naming it rather than leaving it silent, per AGENTS.md's rule for standing down on one.

[The job](https://github.com/macanderson/stella/actions/runs/34081458574/job/101617388283) fails inside `stella-tools`:

```
thread 'a_timed_out_hook_leaves_no_surviving_grandchild' panicked at
crates\stella-tools\tests\group_kill_witnesses.rs:236:5:
```

That is the test's own **premise** assert, not the property — the `TimedOut` assert above it passed, so the shell resolved and the budget did end the hook. This PR's diff touches `stella-cli/src/wrapper_plugin*` and `stella-runtime/src/wrapper/*` only; it contains no `stella-tools` file and no hook code. `windows-check.yml` triggers on `crates/stella-runtime/**`, which is why the job ran at all.

The cause is a timing coupling in the witness itself: Witnesses 1 and 2 in that file run the identical `backgrounding_command` and pass, because they watch a live process and abort it only once `started_beating` answers. Witness 3 cannot — the hook's own timeout is what kills the group and the premise assert runs after it — so the grandchild's whole window to reach its first `printf` is the hook's 300ms budget, which is narrower than a cold Windows runner takes to spawn a shell, a subshell and a `printf`.

Fix ported out to #6372 rather than widening this PR: the budget becomes a named `HOOK_BUDGET_MS` at three seconds, still far short of the `sleep 30` the hook waits on, so the timeout is still what ends it. Once that lands this job should go green here on the next push; the check is not a required one, so it is not blocking this PR in the meantime.

---

Issue #4336, opened 2026-08-23
The issues view is missing linked plan info, evidence summary, and heat sort


macanderson:
## Problem

Rendering `07-issues` shows per-row `plan r3 · task 3 live`, a detail pane with `linked plan r3 · 2/6 · branch fix/token-race · evidence 2 files · 4/4 tests`, and a default heat sort (coupling of touched files × age, from the graph). `crates/stella-tui/src/views/issues_tab.rs` (`render_detail`) draws the tracker fields it has — state, assignee, labels, URL, updated — and sorts as the tracker delivered.

(The path in the original text was `views/issues.rs`; the tab is `views/issues_tab.rs`, and `views/issues.rs` is now the pure heat/link fold beside it.)

## Done means

- `IssueRow` (`crates/stella-tui/src/envelope.rs`) gains the link: the plan/branch this session started for the issue (`w start work`), the files and tests its evidence ledger holds.
- A heat sort fed from the graph's coupling for the files the issue's linked work touched; caption `from the graph, not vibes`.
- The detail pane draws the `linked` line and the sync rule (`status syncs back to <tracker> on gate green`) once the outbound update exists.

## Definition of done

Scoped to the **view** work — the surfaces this issue's title names. The data
feeds behind them have no producer in this tree and are split out below, which
is the path SCR-004 asks for rather than carrying them here.

- [x] Implementation complete — `IssueRow::linked` / `LinkedWork` in `crates/stella-tui/src/envelope.rs`; the pure heat fold in `crates/stella-tui/src/views/issues.rs` (graph coupling × age, `HEAT_CAPTION`); `IssuesPanel::adopt_rows` applying it as a list arrives; the inline `plan r3 · task 3 live` tag, the `linked` line and the caption in `crates/stella-tui/src/views/issues_tab.rs`.
- [x] Scoped tests added and passing — witnesses for the rendered row and detail line, the ingest-time ordering, the caption's presence and absence, the age term and tie stability, plus a re-blessed `tests/snapshots/deck/tab_issues.txt` and the accessible-mode record.
- [x] Full CI green
- [x] Code comments and docs updated — `LinkedWork`, `views::issues`, `issues_tab`'s module header, and `render_detail`'s doc comment, which records why the sync rule is **not** drawn.
- [x] Residue filed as new issues (triage label only) — #5195, #5196, #5197.

### Split out, not done here

[...]

PR #6372, opened 2026-09-07
fix(stella-tools): give the hook witness time to have a grandchild

macanderson:
## What & why

`windows-check.yml` is red on `a_timed_out_hook_leaves_no_surviving_grandchild`
(`crates/stella-tools/tests/group_kill_witnesses.rs`), and it fails on its own
**premise** rather than on the property it exists to prove:

```
thread 'a_timed_out_hook_leaves_no_surviving_grandchild' panicked at
crates\stella-tools\tests\group_kill_witnesses.rs:236:5:
```

Observed on [#6369's run](https://github.com/macanderson/stella/actions/runs/34081458574/job/101617388283)
(job `101617388283`), where the `TimedOut` assert above it passed — so the shell
resolved, the hook ran, and the budget did end it. Only the grandchild's first
write never landed.

Witnesses 1 and 2 in the same file run the identical `backgrounding_command`
against the same runner and pass. The difference is **when each one looks**.
Both of those watch a live process and abort it only once `started_beating`
answers, so the writer has as long as it needs to reach its first `printf`.
Witness 3 cannot do that: the hook's own timeout is what kills the group, and
the premise assert runs after it. The whole window the grandchild ever has is
therefore the hook's budget — and once the group is dead, the 5s
`started_beating` poll is watching a file that will stay empty forever.

At 300ms that window was narrower than a cold Windows runner takes to spawn a
shell, a subshell and a `printf`.

The budget is now a named `HOOK_BUDGET_MS` at three seconds, with the
constraint on it in the doc comment: wide enough for that spawn, and far short
of the `sleep 30` the hook then waits on, so the budget is still what ends the
hook. Nothing about the property under test moves — only how long the
grandchild is given to exist before the kill it must survive is measured.

Refs #6369 (that PR's diff touches only `stella-cli/wrapper_plugin` and
`stella-runtime/wrapper` — no `stella-tools`, no hook code — so this failure
was never its own; it is ported out here rather than widening that PR).

## The witness

- [x] No witness needed (pure refactor / docs / CI) — because: this **is** a
      repair to a test's own premise. A witness would have to be a second test
      asserting the first one's timing, which proves less than the job itself.
      The evidence is `the platform split still compiles on Windows` going
      green on this branch, on the same job that failed on #6369.

## The gate

- [x] `cargo fmt --check` — clean on the changed file
- [ ] `cargo clippy --workspace --all-targets -- -D warnings` — CI's job
      (CLAUDE.md: "CI builds and tests; this laptop does not")
- [ ] `cargo test --workspace` — CI's job; the Linux run of this file is
      unaffected (all three witnesses already pass there)
- [x] Toolchain-free guards run locally: `make prose`, `make line-citations`,
      `make left-behind` — all OK
- [x] Docs updated where behavior changed — the constant carries its own
      constraint in a doc comment; no user-facing behavior moves

## Fix over file

- [x] Extra fixes in this PR: none — the change is one constant and its
      doc comment
- [x] Nothing was deferred

## Ground-rule check

Three seconds is chosen for headroom, not from a measurement of Windows spawn
cost — I have no such measurement, which is why it carries no `MEASURED:`
marker. If the witness ever flakes again at this budget, the honest next step
is to measure that spawn rather than to raise the number again.

The residual shape is inherent to what Witness 3 proves: a timeout-driven kill
cannot be observed without the killed thing having started first, so the budget
is the only lever. Witnesses 1 and 2 have no such coupling and are untouched.

https://claude.ai/code/session_01BjAtn43cWo16yGjZYXg3uQ

---
_Generated by [Claude Code](https://claude.ai/code/session_01BjAtn43cWo16yGjZYXg3uQ)_

## Summary by Sourcery

Bug Fixes:
- Increase the timed-out hook witness budget so Windows runners have enough time to start and observe the grandchild before the process group is terminated.

### DIFF
--- a/crates/stella-tools/tests/group_kill_witnesses.rs
+++ b/crates/stella-tools/tests/group_kill_witnesses.rs
@@ -58,6 +58,19 @@ use stella_tools::registry::Tool;
 /// bytes here and a killed one by none.
 const OBSERVATION: Duration = Duration::from_millis(200);
 
+/// How long Witness 3's hook runs before its own budget ends it.
+///
+/// Witnesses 1 and 2 watch a live process and abort it once it is beating, so
+/// the writer has as long as it needs to start. Witness 3 cannot: the hook's
+/// timeout is what kills the group, so the first `printf` has to land inside
+/// this budget or it never lands at all, and the premise assert reads a file
+/// that will now stay empty forever. At 300ms that window was narrower than a
+/// cold Windows runner takes to spawn a shell, a subshell and a `printf`, and
+/// the witness failed there on the premise rather than on the property. Three
+/// seconds is wide enough for that spawn and far short of the `sleep 30` the
+/// hook then waits on, so the budget is still what ends it.
+const HOOK_BUDGET_MS: u64 = 3_000;
+
 /// Bytes written to a heartbeat file so far; `0` for one that does not exist
 /// yet.
 fn beats(path: &Path) -> u64 {
@@ -223,7 +236,7 @@ async fn a_timed_out_hook_leaves_no_surviving_grandchild() {
     let dir = tempfile::tempdir().expect("tempdir");
     let pulse = dir.path().join("hook.pulse");
     let mut hung = HookAction::new(backgrounding_command(&pulse));
-    hung.timeout_ms = Some(300);
+    hung.timeout_ms = Some(HOOK_BUDGET_MS);
 
     let err = HostHookRunner
         .run(&hung, "{}", &dir.path().display().to_string())


macanderson:
**Duplicate of #6371** — opened 23 seconds before this one, by a different session.

Both change the same constant in the same file (`crates/stella-tools/tests/group_kill_witnesses.rs`), both name it `HOOK_BUDGET_MS`, both set it to 3s, and both diagnose it identically: witness 3 fails on its own premise because the hook's budget is the grandchild's entire window, and 300ms does not cover a cold Windows spawn. Neither is wrong; they are the same fix written twice.

- #6371: `mac/cool-curie-9v7njp`, session `01RM28VChGtJgEaWk6cCEeAZ`, +17/-1
- this PR (#6372): `mac/cool-curie-20a6vh`, session `01BjAtn43cWo16yGjZYXg3uQ`, +14/-1

They will conflict on the merge of the second, since the doc comments differ in wording.

One difference that bears on which to keep: **#6371 carries the `no-issue` label and this PR does not**, so `dod` can pass there and will fail here until someone labels it. Neither links an issue, which is also why neither session could see the other through `issue-claim.sh` — that claim mechanic is keyed on an issue number, and a `no-issue` CI repair has nothing to claim. That gap is what let this happen, and it is the same shape as #5224.

This PR does carry something #6371 does not: the exact failing job link (run 34081458574, job 101617388283) and the observation that the `TimedOut` assert above the premise passed — the evidence that isolates the failure to the grandchild's first write rather than to the shell or the budget. Worth keeping in whichever PR survives.

Not closing either from here — picking between two near-identical PRs is a quick call for their author and not mine to make unilaterally.

---

macanderson:
This and #6371 are the same fix, opened six seconds apart. Both touch only `crates/stella-tools/tests/group_kill_witnesses.rs`, both add a constant named `HOOK_BUDGET_MS` set to `3_000`, and both make the identical edit `hung.timeout_ms = Some(300)` → `Some(HOOK_BUDGET_MS)` in `a_timed_out_hook_leaves_no_surviving_grandchild`. The only difference between the two diffs is the wording of the doc comment above the constant.

This is the collision AGENTS.md § the issue claim describes (#5224, #4336, #5045) — and `./scripts/issue-claim.sh check <n>` is what would have caught it, since neither branch existed when the other started.

The merge hazard here is the benign shape rather than #4979's: both branches change the same lines, so the second merge conflicts loudly instead of silently dropping one. Nothing is at risk of being lost — it is duplicated work, and one of the two should be closed. Which one is the author's call; the diffs are equivalent in effect, so it comes down to which comment reads better. #6371 carries the `no-issue` label and this one does not, which may decide it.

Flagging rather than closing either — I did not open these.

---

macanderson:
Closing this as the duplicate. #6371 is the survivor.

Two sessions flagged the collision above and both, correctly, declined to close either PR since neither opened one. I opened this one, so retiring it is mine to do and needs nobody's arbitration.

#6371 wins on three counts, none of them about diff quality — the two diffs are equivalent in effect and differ only in the doc comment's wording:
- it was opened 23 seconds earlier;
- it carries the `no-issue` label `dod` wants for an issueless CI repair;
- its `the platform split still compiles on Windows` is green too ([job 101620434118](https://github.com/macanderson/stella/actions/runs/34082538301/job/101620434118)), so it is Windows-verified rather than merely argued.

I've ported the two things this PR had that #6371's body did not — the failing job link with the observation that the `TimedOut` assert above the premise passed (which isolates the failure to the grandchild's first write), and the 5.34s → 3.22s timing inversion that corroborates the diagnosis — into [a comment there](https://github.com/macanderson/stella/pull/6371#issuecomment-5565237777). Nothing is lost by closing this.

One caveat on the sequencing, since it is the only way this is the wrong call: #6371's `fmt + clippy + test` was still in progress at the time of closing, while this PR's had completed green. If that job comes back red for a reason unrelated to the one constant both PRs change, reopen this — the branch `mac/cool-curie-20a6vh` is untouched and its full check set is green.

---

Issue #5913, opened 2026-09-05
main-red-hold's failure is sticky: every PR it blocked stays red after main recovers


macanderson:
## Problem

`main is not known-broken` is a required status check. When it fails, it stays failed on that head **forever** — including after `main` recovers and the `main-red` issue closes. GitHub evaluates a required check against the most recent check-run for that context on the head SHA, and nothing re-runs it.

So the hold outlives the outage it exists for. AGENTS.md describes the intended behaviour as *"`main-red-hold.yml` passes when no issue is open"*, and that is true only at event time.

## Evidence — it happened today

`main` broke on the #5829 / #5855 composition (#5901, `main-red`). #5906 repaired it:

- #5906 merged **07:20:30Z**; #5901 closed **07:20:31Z** by its `Closes`.
- `main` is now `98af1d6b7`, and `crates/stella-core/tests/published_records_do_not_collide.rs` is gone from the tree — the break is genuinely fixed.

Ten open PRs still carry a **failed** `main is not known-broken` on their current head, all from runs created while #5901 was open:

| PR | run |
|---|---|
| #5903 | 33951700124 |
| #5899 | 33950967539 |
| #5894 | 33950666389 |
| #5884 | 33949133950 |
| #5878 | 33949258518 |
| #5874 | 33949101849 |
| #5840 | 33949216893 |
| #5837 | 33949822315 |
| #5835 | 33951264640 |
| #5823 | 33949619451 |

Each is unmergeable on a check whose own question now answers the other way.

## Why nothing clears it

`.github/workflows/main-red-hold.yml` declares:

```yaml
on:
  pull_request:
```

with no `types:`, so it fires on the default `opened`, `synchronize`, `reopened` only. A PR self-heals when someone pushes to it (`synchronize`) — and a PR that is otherwise finished and waiting has no reason to push. Adding `edited` would not help either: the stale run is what branch protection reads, and re-firing on an edit only helps if the workflow runs again, which an edit does not cause here.

The two remedies AGENTS.md forbids are the ones a session reaches for first: *"never push an empty commit or close and reopen the PR to kick CI."* So there is currently **no sanctioned way** for a session to clear this.

## Proposed fix

Close the loop where it opens. `main-canary.yml` already detects recovery and closes the `main-red` issue — the same job should clear the holds it caused:

```yaml
# in main-canary.yml, on the recovery path, after closing the main-red issue
- name: re-run the hold on every open PR
  run: |
[...]

PR #6364, opened 2026-09-07
fix(scripts): reach clear-main-red-holds.sh's write path from the hermetic suite

macanderson:
## What & why

`scripts/clear-main-red-holds.sh` takes fixture seams so the hermetic suite
can drive its read paths without touching GitHub. Every one of them forced
`dry_run`, so all twelve fixture-driven cases in `test-main-red-hold.sh`
exercised the sweep's selection logic and never reached the one call the
script exists to make: `gh api -X POST .../actions/runs/<id>/rerun-failed-jobs`.
A wrong endpoint, a wrong verb, or a missing `actions: write` scope would
have passed the suite and read as a working sweep.

This removes the `use_fixture` → `dry_run` coupling and puts a stub `gh`
ahead of the real one on `PATH` for the clearing suite. Fixtures still
short-circuit the reads (open issues, open PRs, which run belongs to a
head); the write now reaches the stub for real, so a case can assert the
exact call rather than trust the sweep's own "cleared N" line.

Closes #6049

## The witness

Stella's definition of done is a test that **fails on the old code and passes on the new**.

- [x] This PR includes a witness test (fails on `main`, passes here), **or**
- [ ] No witness needed (pure refactor / docs / CI) — because:

Proved by stashing only `scripts/clear-main-red-holds.sh`'s change and
re-running `./scripts/test-main-red-hold.sh` with the new test cases already
in place: 3 of the new assertions go red (the stub `gh` never sees a call,
because `dry_run` is still forced), for example:

```
✗ gh calls did not match — got [] want [api -X POST --silent repos/{owner}/{repo}/actions/runs/33951700124/rerun-failed-jobs
api -X POST --silent repos/{owner}/{repo}/actions/runs/33950666389/rerun-failed-jobs]
main-red-hold-test: FAILED — 35 passed, 3 failed
```

Restoring the fix turns the suite green again — 38/38.

## The gate

- [ ] `cargo fmt --check` — n/a, shell-only change
- [ ] `cargo clippy --workspace --all-targets -- -D warnings` — n/a, no Rust touched
- [ ] `cargo test --workspace` — n/a, no Rust touched
- [x] Docs updated where behavior/flags changed (README, `--help`, doc comments) — the script's own comment on the fixture flags is rewritten to describe the new read/write split
- [ ] CLA signed (the bot prompts on your first PR — nothing to do per commit)
- [x] `Closes #N` appears **both** above and as a commit trailer — squash builds
      the commit body from commit messages, so the description alone won't carry it

[...]

PR #6376, opened 2026-09-07
ci(main-red-hold): re-run the hold when main breaks, not only when it heals

### LINKED ISSUE #5949: main-red-hold passes stale-green: a PR whose hold ran before the main-red issue was filed still merges onto a broken main
## Problem

`main is not known-broken` is evaluated once, when the `pull_request` event fires, and never again. #5913 tracks one consequence of that — a hold that failed stays failed after `main` recovers. This is the other consequence, in the direction that lets a bad merge through rather than blocking a good one: a hold that **passed** stays passed after `main` breaks, so a PR whose check ran before the `main-red` issue was filed merges onto a known-broken tree with a green required check.

Branch protection reads the most recent check-run for the context on that head SHA. For a PR sitting finished and waiting on review, that run can be hours old and is answering a question about a different moment.

This is #5313's "green over a real failure" half. #5913 is its "red on a real pass" half. Same root cause — the hold is a point-in-time answer treated as a standing one — but the fix for #5913 (re-run failed holds on recovery) does not address this direction, because these runs are not failed and there is no recovery event to hang the work off.

## Evidence — it happened today

`main` broke on the `stella-learn` lockfile composition (#5939, `main-red`, filed **10:07:09Z**; repaired by #5940, which merged **10:37Z**, closing #5939 at **10:37:25Z**).

#5928 merged into that window:

| fact | value |
|---|---|
| #5928's `main is not known-broken` run | job `101285911129`, started **09:41:08Z**, completed **09:41:22Z**, conclusion `success` |
| #5939 (`main-red`) filed | **10:07:09Z** — 26 minutes *after* that run |
| #5928 merged | `e4a137c`, during the red window |
| `main`'s lock at that moment | `stella-learn` `0.9.329` against a `0.9.330` workspace — still broken |

The hold never re-ran between 09:41 and the merge. Its green was truthful when produced and false when consumed.

Verified directly rather than inferred: the timestamps above are from #5928's check-run list, and `git show <tip>:Cargo.lock` on the tip #5928 produced still carried `0.9.329`.

## Why nothing catches it

- `.github/workflows/main-red-hold.yml` triggers on `pull_request` with no `types:`, so it fires on `opened`, `synchronize`, `reopened`. Filing a `main-red` issue is none of those, so no open PR re-evaluates its hold when `main` breaks.
- `main-canary.yml` files the issue but does not touch open PRs.
- The other required checks are equally stale, and for the same reason — but they are answering questions about the PR's own diff, which has not changed. The hold is the only required check whose subject is external state that moves under it.

## Proposed fix

Symmetric with #5913's: the canary already knows the transition, so let it push the holds. On the **break** path — the same place it opens the `main-red` issue — re-run `main-red-hold.yml` for every open PR, so their holds turn red while `main` is red. #5913's recovery path then turns them back green.

That needs `actions: write` on the canary job, which is the same permission #5913 asks for and does not have. Worth deciding once, for both directions, rather than twice.

Two alternatives, both worse:

- **Make the hold a merge-queue check instead of a status check.** Correct in principle — the queue re-evaluates at merge time, which is exactly when the question should be asked — but it is a much larger change to how this repo merges, and it is not obviously wanted for its own sake.
- **Shorten the window by filing the issue faster.** Not a fix. The window is however long a PR waits between its last push and its merge, which is unbounded.

## Verify

1. Open a throwaway PR and let its checks go green, hold included.
2. Hand-file an issue labelled `main-red`. Confirm the PR's `main is not known-broken` **stays green** with no new run — today's behaviour, and the bug.
3. Confirm the PR is still mergeable while that issue is open.
4. With the fix, confirm the break path re-runs the hold and the PR goes red with no push to its branch.

## Definition of done

- [x] A `main-red` issue being filed turns the hold red on every open PR, without a push to any branch. `scripts/refresh-main-red-holds.sh` reads the tracker, and an open `main-red` issue makes `failure` the verdict every hold should carry; it re-runs the hold on each open PR whose last run says `success`. `main-canary.yml`'s `Refresh the hold on every open pull request` step runs it on the push run that files the issue. PR #6382.
- [x] The canary job carries the narrowest permission that allows it (`actions: write`), and nothing wider — coordinated with #5913 so the permission is added once. #5913 already added it; this PR adds no permission and no second call site, and the comment above it in `main-canary.yml` now names both directions. The test case `...and holds the write scope that a re-run needs` pins it.
- [x] A hermetic case in `main-red-hold`'s own suite covers the stale-**green** transition, alongside #5913's stale-red one. `scripts/test-main-red-hold.sh`, section `A broken main turns the passing holds red`: four cases, driven from the same fixture table as the recovery cases, plus two negative controls (a failing hold is left alone while `main` is red; the sweep never prints `main has recovered` while an issue stands). No network, no `gh`.
- [x] That case is a witness. Before: the old `scripts/clear-main-red-holds.sh` on the same scenario (one `main-red` issue open, a PR whose hold passed) prints `main is still known-broken (5901) — the holds are correct` and re-runs nothing — the #5928 window reproduced. After: `refresh-main-red-holds: would re-run the hold on PR #5899 (head bbbbbbb, run 33951700200)`. The suite reports 52 checks passed, including a call-log case asserting the single POST while `main` is red is for the hold that still passes. Both runs are quoted in PR #6382's description.
- [x] AGENTS.md § "A seventh, `main-red-hold.yml`" now reads "Each run asks the tracker whether a `main-red` issue is open **at that moment**", and says the answer is a snapshot that the next paragraph re-takes. That paragraph is retitled **A hold goes stale in both directions unless something re-asks it**, carries the #5928 timeline beside the 2026-09-05 one, and describes the sweep as direction-aware.
- [x] The step lands in `.github/workflows/main-canary.yml`, calling `scripts/refresh-main-red-holds.sh` — the existing `clear-main-red-holds.sh` renamed, since a script that also turns holds red cannot keep a name that says clear. No god file created or grown: `make guards-fast` runs `file-size` and `god-files` and both pass (exit 0), and the script is well under the 1500-line ceiling with no baseline entry.
- [x] `make guards-fast` passes (exit 0, `shellcheck` included), and `python3 scripts/check-prose.py` reports `OK — none added` after the new comments were shortened to the 6.00 grade ceiling.
- [x] Confirmed rather than skipped: the diff is one workflow file, two shell scripts, a `Makefile` target and AGENTS.md. No crate, no interface, no dependency, no colour token.

## Duplicate check

Not a duplicate of #5913, which is the stale-**red** direction and whose proposed fix (re-run *failed* holds on recovery) leaves this case untouched. Not #3932 (closed — made the hold required, the prerequisite both bugs sit on). Not #5850, which is about a test-only failure filing no `main-red` issue at all; here the issue was filed correctly and the hold simply did not re-ask. Not #5943, which is about the merge-result lockfile guard's coverage rather than the hold.

## Context

- #3917 — why the hold blocks rather than reports, and the four-PRs-onto-red-main incident it was built for. #5928 is that same shape, reached through a green check instead of past a red one.
- #5939 / #5940 — today's outage and repair.
- #5913 — the sibling direction; these two should probably be fixed in one change.

Found while triaging open PR CI after the #5939 outage. No priority or size label, per SCR-005.

macanderson:
## What & why

`main is not known-broken` is evaluated once, when the `pull_request` event fires. Branch protection then reads that answer for as long as the head stays put, and a pull request that is finished and waiting raises no further event. So the hold keeps answering a question about an earlier moment while `main` breaks and gets fixed underneath it.

`#5913` fixed the direction that costs a merge — the stale failure a recovery leaves behind. This is the direction that lets one through:

| fact | value |
|---|---|
| `#5928`'s `main is not known-broken` run | job `101285911129`, completed **09:41:22Z**, conclusion `success` |
| `#5939` (`main-red`) filed | **10:07:09Z** — 26 minutes later |
| `#5928` merged | `e4a137c`, during the red window, on that green |

Both directions are one bug, so this is one sweep rather than two. `scripts/clear-main-red-holds.sh` becomes `scripts/refresh-main-red-holds.sh` and asks the tracker what every hold *should* be saying — failing while a `main-red` issue is open, passing while none is — then re-runs the hold on each open pull request whose last run says the other thing. The new run lands under the same name on the same commit, which is all branch protection reads.

A second script for the break direction would be a second copy of that one question, free to drift from the first. The canary already calls the sweep and already carries `actions: write` (added for `#5913`), so the break path needs no wider permission and no second call site: the push run that files the issue is the same run that closes it.

Closes #5949

## The witness

The new cases live in `scripts/test-main-red-hold.sh` under "A broken main turns the passing holds red". Run before and after, on the same scenario — one `main-red` issue open, three open pull requests, one of whose holds passed:

**Old script** (`git show HEAD~1:scripts/clear-main-red-holds.sh`, its own fixture spelling):

```
$ ./old --fixture-open-issues 5901 --fixture-open-prs '…' --fixture-stale-runs '…'
clear-main-red-holds: main is still known-broken (5901) — the holds are correct.
exit=0
```

Nothing re-run. That is `#5928`'s merge window, reproduced.

**New script**, same scenario:

```
$ ./scripts/refresh-main-red-holds.sh --fixture-open-issues 5901 …
refresh-main-red-holds: would re-run the hold on PR #5899 (head bbbbbbb, run 33951700200)
refresh-main-red-holds: main is known-broken (5901) — re-ran the stale hold on 1 of 3 open pull request(s).
```

The suite also pins the negative controls for this direction: a hold that already fails is left alone while `main` is red, and the sweep never prints "main has recovered" while an issue stands. Running the whole suite against the old script fails 12 of 42 checks; against this one, 42 pass.

One case is about wiring rather than the script: the sweep step in `main-canary.yml` must come *after* the step that files or closes the issue. Reversed, the break path would read the tracker's answer from before the issue existed and re-run nothing — the same bug with more machinery in front of it.

## The gate

- [x] `cargo fmt --check` — via `make guards-fast`, exit 0
- [ ] `cargo clippy --workspace --all-targets -- -D warnings` — no Rust in this diff; CI runs it
- [ ] `cargo test --workspace` — no Rust in this diff; CI runs it
- [x] Docs updated where behavior changed — AGENTS.md § the seventh workflow, and the `Makefile` target
- [x] CLA signed
- [x] `Closes #5949` appears both above and as a commit trailer

`make guards-fast` passes (exit 0, `shellcheck` included), `python3 scripts/check-prose.py` passes, and `./scripts/test-main-red-hold.sh` reports 42 checks passed.

## Fix over file

- [x] Extra fixes in this PR:
  - **A `cancelled` or `timed_out` hold was never swept.** The old predicate was "did the last run fail". Branch protection reads those two conclusions as not-passing as well, so they blocked a pull request exactly like a failure and no sweep touched them. The new predicate compares the conclusion against what the tracker says it should be, which covers them by construction.
  - **A run still in flight is now left alone.** The old sweep would have re-run one, cancelling a run that was already asking today's question.
  - **`scripts/dod-recheck.sh`'s two citations of the renamed script** are repointed, so its header does not name a file that no longer exists.

## Ground-rule check

Not applicable — this is a GitHub Actions workflow, a shell script and a doc update. No crate, no interface, no dependency.

The rename is what makes the design honest rather than cosmetic: a script that also turns holds red cannot keep a name that says "clear". `main-red-clear.yml` keeps its own name — it is still only the recovery caller, since it fires on the issue being closed or unlabelled.

A pull request labelled `unblocks-main` is swept like any other on the break path. Its hold passes by design, so it gets re-run and passes again — one wasted job on the repair pull request, against the cost of reading every pull request's labels to save it. The script's header says so.

## Summary by Sourcery

Keep required pull-request holds synchronized with the current `main-red` tracker state so stale checks cannot block valid merges or permit merges onto broken main.

New Features:
- Refresh stale `main is not known-broken` checks in both directions so open pull requests reflect whether `main-red` is currently open.

Bug Fixes:
- Prevent pull requests with stale passing holds from merging while `main` is broken.
- Refresh cancelled and timed-out holds, while leaving in-flight runs undisturbed.

Enhancements:
- Rename the cleanup script and consolidate recovery and breakage handling into a single tracker-driven sweep.
- Run the refresh after the canary files or closes the `main-red` issue and retain the required workflow permissions.

Build:
- Rename the Makefile target to `refresh-main-red-holds` and update its description.

CI:
- Update the canary and recovery workflows to invoke the unified hold-refresh sweep.

Documentation:
- Document that hold answers can become stale in either direction and explain the unified refresh process.

Tests:
- Expand the hermetic hold test suite to cover stale passes, stale failures, in-flight runs, rerun behavior, and workflow ordering.

Chores:
- Update references to the renamed script in supporting scripts and fixtures.


macanderson:
Merged `origin/main` in to clear the conflict this PR opened with. `#6364` landed while this branch was open and rewrote the same two files this branch renames, so three things needed settling by hand.

**The fixture → dry-run coupling stays removed.** That removal is `#6364`'s whole point — it is what lets the hermetic suite reach the write path. This branch still carried the coupling, because it was cut before `#6364`. Kept main's removal, kept this branch's renamed `note()` prefix beside it.

**`#6364`'s `gh` stub and its eight write-path cases are kept**, re-pointed at this branch's names: `$REFRESH` for `$CLEAR`, `$held_prs` / `$hold_runs` for `$recovered_prs` / `$stale_runs`, `--fixture-hold-runs` for `--fixture-stale-runs`, and this branch's summary wording.

**The endpoint stays this branch's `/rerun`, not `#6364`'s `/rerun-failed-jobs`** — and that is the one resolution a reviewer should look at, because it silently reverses what `#6364`'s tests pinned. The break direction re-runs a hold that *passed*, and a run with no failed job in it is exactly what `rerun-failed-jobs` declines to do. So that endpoint can only ever serve the recovery half, and this sweep is both halves. The reason was in neither file; it is now a comment beside the call.

**Evidence.** `./scripts/test-main-red-hold.sh` reports **50 checks passed** — this branch's 42 composed with `#6364`'s 8 write-path cases. The description above says 42; that number is now the pre-merge one. `make guards-fast` exits 0 with `shellcheck` present.

One thing the merge itself broke and this push fixes: combining both branches' comments pushed `scripts/test-main-red-hold.sh` past its reading-grade ceiling (6.00 allowed, 6.25 found). Three sentences were shortened rather than a line added to `scripts/prose-grade-baseline.txt`; `check-prose.py` is green. One of those sentences also still described fixture spelling `"ok"`, which this branch renamed to `success`.

---

macanderson:
Superseded by #6382.

This branch was cut from an older `main` and went `CONFLICTING` once #6364 rewrote the same two files, so GitHub never created the merge ref and no `pull_request` workflow started for it — the head carries one `github-actions` check suite (`cla`, which is `pull_request_target`) and nothing else.

Rebasing it would have replayed a branch whose only conflicts were in the two files #6364 had just changed, which is the shape `rebase-replay` exists to catch. The change was rebuilt on the current tip instead, keeping #6364's read/write fixture split and its `gh` stub. Closing this one.

Issue #5863, opened 2026-09-05
Stella keeps re-sending one command with one more copy of itself glued on, and no loop rung stops it


macanderson:
## What happens

A model gets stuck and keeps sending one shell command. Each time it sends the
command again with **one more copy of a block it already contains** glued on
the end. The command grows every step. The output grows every step. Nothing is
learned and nothing is written.

None of the five rungs in `crates/stella-core/src/loop_detect.rs` fires on it.
Exact repeat, short cycle and interleaved repeat all need a byte-identical
input, and every input here is different. Stagnation needs byte-identical
output, and every output here is different. `MonotonicSweep` needs cursors that
advance and wrap, and there are no cursors. This is the blind spot #3292
describes, and unlike the trace #3292 names, this shape is easy to score.

## Two traces that show it

Both are in the local `~/.arenabench/matches` corpus.

**`build-pov-ray__3DYrafv`** — reward **1.0**, 307 calls.
`~/.arenabench/matches/14d0127b8ca7/jobs/14d0127b8ca7-stella-kimi-k3/build-pov-ray__3DYrafv/agent/stella-events.jsonl`

Calls 196 through 302 — 107 steps in a row — each send the previous command
plus one more copy of the same twelve-file `grep -n "display_plot" …` block:

```
call 196    3 copies     263 chars    output 2453 chars
call 250   57 copies    4593 chars
call 302  109 copies    8774 chars    output 2736 chars
```

The task's deliverable was installed and verified at calls 94 and 106, before
the grind started, which is why the trial still scored a win. 107 of its 307
steps bought nothing.

**`feal-linear-cryptanalysis__s5TVw8A`** — reward 0.0, 47 calls.
`~/.arenabench/matches/1904fa8992d9/jobs/1904fa8992d9-stella-glm-5-2-via-openrouter-full-pipeline/feal-linear-cryptanalysis__s5TVw8A/agent/stella-events.jsonl`

Calls 19 through 47 append one more `&& ./feal …` per step and run to the last
step of the trial.

## Measured over the whole corpus

Scored as "longest run of consecutive calls to one tool where each input starts
with the whole of the input before it", over 138 stuck trials (reward 0,
status `interrupted`) and 233 winning trials (reward 1):

| threshold | stuck trials that fire | winning trials that fire |
|---|---|---|
| 3 | 2 / 138 (1.4%) | 1 / 233 (0.4%) |
| 8 | 1 / 138 (0.7%) | 1 / 233 (0.4%) |
[...]

Issue #5834, opened 2026-09-05
The issue-claim check says no PR claims the issue even when it could not read the PR list


macanderson:
## Problem

`scripts/issue-claim.sh check <n>` can print

```
ok  #5117 is unclaimed — no PR closes it and no live claim holds it.
```

right after it has said it could not read the pull request list. The line
states a fact the run never checked. A session that trusts it starts work
that a peer already has an open PR for.

This happened on #5117 today. The run printed:

```
note: could not list pull requests. Proceeding (fail-open); the claim
      check below still runs.
ok  #5117 is unclaimed — no PR closes it and no live claim holds it.
```

PR #5829 was open at that moment, said `Closes #5117` in its body, and was
green and mergeable. The check that exists to stop a duplicate implementation
gave the all-clear for one.

## Why the fail-open itself is right, and what is wrong

Proceeding when the tracker cannot be reached is the designed behaviour, and
it should stay. A claim check that can block work is worse than the
duplication it prevents (AGENTS.md, the same rule `main-red-claim.sh` follows).

The bug is the wording of the success line. The script already has the honest
form for its other two unknowns — it exits with `ok  proceed (could not ask)`
and `ok  proceed (comments unreadable)` instead of claiming an answer. The pull
request query is the only unknown that falls through to a sentence asserting
the answer it failed to get.

## Files

- `scripts/issue-claim.sh` — the fail-open branch for the `gh pr list` call,
  and the `proceed` call for `mode = check` that prints the sentence.
- `scripts/test-issue-claim.sh` — the guard's own tests (`make issue-claim-test`).

## The fix

Track whether the pull request query succeeded. When it did not, have the
`check` mode print a line that says so, in the shape the other two unknowns
already use — for example `ok  proceed (PR list unreadable)`. Leave the exit
code at 0. Leave the full sentence for a run that really did read the list and
find nothing.

## Verify

Run the script with the pull request query forced to fail and confirm the
output no longer claims that no PR closes the issue. The suite already fakes
these inputs through its fixture mode, so the case can be added there rather
than by breaking `gh`.

## Definition of done

- [x] A `check` run whose pull request query failed no longer prints a line
      saying no PR closes the issue.
- [x] That run still exits 0, so the check can never block work.
- [x] A run that did read the list keeps the sentence it prints today.
- [x] `scripts/test-issue-claim.sh` covers the failed-query wording, and
      `make issue-claim-test` passes.
[...]

PR #6377, opened 2026-09-07
feat(scripts): claim a pull request, and stop a second copy of one finding

### LINKED ISSUE #5861: Concurrent PR sweeps duplicate each other: three identical conflict analyses posted to one PR in eight minutes
## Problem

Several autonomous PR-sweep sessions run against this repository at once, and
nothing tells one that another is already working a given pull request. They
converge on the same PR, reach the same conclusion, and each post it.

Observed on **#5835** on 2026-09-05. Three comments, by the same account,
within eight minutes, each an independent analysis of the same merge conflict
against #5828:

| Comment | Posted |
|---|---|
| [5549621358](https://github.com/macanderson/stella/pull/5835#issuecomment-5549621358) | 05:21:52 |
| [5549646286](https://github.com/macanderson/stella/pull/5835#issuecomment-5549646286) | 05:25:06 |
| [5549681307](https://github.com/macanderson/stella/pull/5835#issuecomment-5549681307) | 05:29:58 |

All three correctly identify that `main`'s `SeatGrant` (#5828) resolves a seat
to a `ModelCallRole` while #5835 retires six of those variants and makes seats
open strings; all three correctly decline to resolve it and hand it to the
maintainer. All three render the same comparison table. A fourth sweep session
reached the same conclusion independently at 05:35 and stopped before posting
only because it read the comments first.

## Why this is the epic's shape, one level up

#5310 collects the cases where parallel sessions take one resource without
seeing each other: #5224 is two sessions taking one **issue**, #5265 two
sessions taking one **repair**, #5287 two sessions taking one **working tree**.
This is two sessions taking one **pull request**, and it is the only member of
the family with no claim mechanism at all — `scripts/issue-claim.sh` claims an
`issue:<n>`, and nothing claims `pr:<n>`.

The cost is different from the others, which is why it needs its own entry:
duplication here does not lose work, it **publishes** it. Every duplicate is a
comment on the maintainer's PR, and a reader has to diff three long analyses to
learn they say one thing.

## What is not the bug

The analyses were right, and declining to resolve #5835 was right. The fail-open
posture of the claim scripts is right too, for the reason AGENTS.md gives — a
claim check that can block a sweep is worse than the duplication it prevents.
The bug is only that a sweep publishes without first asking whether the finding
is already published.

## Files

- `scripts/issue-claim.sh` — the existing mechanic, keyed on `issue:<n>`. The
  `dispatch_claims` lease table it writes to is `.stella/private/fleet.db`.
- `scripts/main-red-claim.sh` — the sibling shape (claim as a tracker comment,
  lapses after twenty minutes, proceeds loudly at every unknown).
- Whatever prompt or routine drives the sweep — not in this repository, which
  is part of the problem: nothing here can enforce it today.

## Possible shapes

1. **Read before writing.** Cheapest and needs no new state: before posting an
   analysis, list the PR's existing comments and skip if a substantively equal
   finding already stands. This is what the fourth session did by hand, so it
   is known to work.
2. **Claim `pr:<n>`.** `scripts/issue-claim.sh`'s mechanic pointed at a pull
   request — a lapsing comment claim, fail-open at every unknown. Stronger, and
   symmetric with the rest of #5310's family.
3. **One sweep.** Serialize the routine so only one runs at a time. Removes the
   hazard rather than reporting it, at the cost of throughput.

(1) and (2) compose; (1) alone would have prevented all three duplicates here.

## Definition of done

- [x] Two sweep sessions working the same open PR at the same time produce at
      most one comment for a given finding. `scripts/pr-claim.sh post <n>
      --finding <key> --body-file <path>` stands down on a peer's live claim,
      and then writes only when no comment already carries that key's marker;
      a published finding never lapses. Cases `a finding already posted stops a
      second copy`, `an old finding still stops a copy` and `a peer's live
      claim stops a post before the finding gate` in
      `scripts/test-pr-claim.sh`. The key is named by the caller rather than
      hashed from the write-up, because the three duplicates above were worded
      differently and said one thing. Two reads landing inside the same second
      still both proceed: an at-most-once guarantee needs a lock, and a comment
      thread is not one — the PR comment answering Sourcery's row on #6377
      carries that argument, and shape 3 is where the last second would be
      closed. (#6377)
- [x] A sweep that cannot determine whether a peer holds the PR still proceeds
      (fail-open), and says so rather than asserting the PR is unclaimed —
      the wording rule #5834 just landed for `issue-claim.sh`. Both gates carry
      it: `check` reports `proceed (comments unreadable)` / `proceed (state
      unreadable)` / `proceed (identity unknown)`, and `post` reports
      `duplicate check unreadable` instead of `nothing carries that key`. Cases
      `a failed comment read posts anyway, and says it could not check`,
      `an unreadable comment list proceeds`, `an unreadable state proceeds to
      the claim check, which still blocks`, and `an unknown identity proceeds`.
- [x] A hermetic self-test in the shape of `scripts/test-issue-claim.sh`
      covers both: the duplicate is suppressed, and the unknown proceeds.
      `scripts/test-pr-claim.sh` (`make pr-claim-test`), 50 cases, no network
      and no `gh`. It runs in `guard-self-tests.yml` beside the other two claim
      suites, so `check-gate-parity` holds it there.
- [x] That test fails before the change. With `scripts/pr-claim.sh` moved
      aside the suite exits 1 on `scripts/pr-claim.sh is missing or not
      executable`; restored, it reports `50 passed, 0 failed`. Both runs are
      quoted in #6377's "The witness" section.

## How this was found

An autonomous PR-management sweep on 2026-09-05 scanned all 21 open PRs across
the `macanderson` org. Every actionable item it found had already been handled
by a peer sweep minutes earlier — #5834's nine DoD boxes ticked at 05:24, #5287's
two at 05:15, #5835's conflict analysed three times. The sweep took no action on
any PR as a result, which is the correct outcome and also the evidence.

macanderson:
## What & why

Several autonomous PR-sweep sessions run against this repository at once, and
nothing told one that another was already reading a given pull request. Three
of them worked out the same merge conflict on one pull request in eight
minutes and each posted it; a fourth reached the same answer and stopped,
because it read the comments first. Duplication here does not lose work, it
publishes it: the maintainer is left to diff three long comments to learn they
say one thing.

`scripts/issue-claim.sh` claims an `issue:<n>`. Nothing claimed a `pr:<n>`.
This adds `scripts/pr-claim.sh` — that mechanic pointed at a pull request,
plus the read-before-write gate the issue names as shape (1):

```bash
./scripts/pr-claim.sh check 5835    # exit 0 proceed, 1 stand down
./scripts/pr-claim.sh claim 5835    # check, then post the claim
./scripts/pr-claim.sh post 5835 --finding conflict:5828 --body-file note.md
```

- **`check` / `claim`** are the issue-claim rules unchanged: the tracker is the
  table, a comment is the claim, it lapses after 90 minutes, a session word
  rides beside the login so two sessions of one author can tell each other
  apart, and every unknown proceeds out loud. A merged or closed pull request
  stands a sweep down as well, since neither takes a comment.
- **`post`** is the half that would have stopped all three comments. It runs
  the claim check first, then asks whether a comment already carries this
  finding's key, and writes only when neither answer is yes.
  `--ignore-claim` posts over a live claim, for the operator who read it and
  said on the pull request why.

**The caller names the finding, not a digest of the prose.** The three
duplicates on the pull request that prompted this were worded differently and
said the same thing, so a hash of the text would have called them distinct.
A key that has to go stale carries what it rests on — a finding about the head
commit puts the head sha in the key, so a later push can be reported again.

**A read that failed is not an empty list.** A run that could not read the
comments posts anyway and says `duplicate check unreadable`, rather than
reporting that nothing stands. That is the wording rule #5834 landed for
`issue-claim.sh`, applied on both gates here.

Both jq filters are reachable as pure modes (`select`, `findings`): comments
JSON on stdin, parsed rows out, so a typo in a filter fails a test rather than
a live sweep.

Closes #5861

## The witness

`scripts/test-pr-claim.sh` (`make pr-claim-test`), hermetic — no network, no
`gh`. It runs in `guard-self-tests.yml` beside the other two claim suites, so
`check-gate-parity` is satisfied.

The fail→pass demonstration, run by hand:

```
$ mv scripts/pr-claim.sh /tmp/x.bak && ./scripts/test-pr-claim.sh
  x  /Users/.../scripts/pr-claim.sh is missing or not executable
  exit=1
$ mv /tmp/x.bak scripts/pr-claim.sh && ./scripts/test-pr-claim.sh
  50 passed, 0 failed
```

Before the suite grew its missing-subject preflight, that first run reported
one accidental pass beside the rest: a case asking only for a non-zero exit is
satisfied by `exit 127`. The preflight is why an absent subject now reads as
an absent subject rather than as a wall of unrelated failures.

The two cases the issue's DoD asks for, by name:

- `a finding already posted stops a second copy` and `an old finding still
  stops a copy — a published comment does not lapse` — the duplicate is
  suppressed.
- `a failed comment read posts anyway, and says it could not check` — it
  asserts the report does **not** say `nothing carries that key`, so the
  wording half is checked from both sides.

`4b7437c` adds the claim gate to `post`, with eight further cases covering
both fail-open directions and the case where a session's own claim must not
block its own post.

## The gate

- [x] `cargo fmt --check` (via `make guards-fast`)
- [ ] workspace clippy — no Rust in this diff; CI runs it
- [ ] workspace tests — no Rust in this diff; CI runs it
- [x] Docs updated where behavior changed — AGENTS.md's claim section, and the
      script's own `--help`
- [x] CLA signed
- [x] `Closes #5861` appears above **and** as a commit trailer

Local evidence: `make guards-fast` exits 0 (shellcheck, `check-gate-parity`,
`check-file-size`, `check-prose`, `check-line-citations` and the rest),
`python3 scripts/check-prose.py` reports OK with nothing added, and
`./scripts/test-pr-claim.sh` passes. No workspace build was run on the
maintainer's laptop; CI is this PR's build evidence.

## Fix over file

- [x] Extra fixes in this PR, two of them:
      1. AGENTS.md said `scripts/lib/claim-session.sh` resolves the session
         word "for both scripts" and that "both fail open the same way", and
         that file's own header named the two scripts that read it. Three read
         it now, so both sentences name the set rather than counting it — a
         count in prose is a promise nobody keeps.
      2. `pr-claim.sh post` took the shared no-`gh` branch, which proceeds out
         loud and exits 0. Right for `check`; for `post` it told a sweep its
         finding was up while nothing had been sent. It exits 3 and names what
         did not happen, covered by `a post with no gh fails and says nothing
         was posted` — run against a PATH holding two symlinks, so `gh` is
         absent for real.

## Ground-rule check

- [x] No new outbound network calls — the script talks to `gh`, which the two
      sibling claim scripts already do, and it is never on a `stella` run path

**What this does not do.** Nothing in this repository drives the sweeps, so
nothing here can force a sweep to call this. The issue says as much. What
lands is the mechanism, its self-test, and the AGENTS.md paragraph that tells
a session to use it — the same standing this repository gives
`scripts/issue-claim.sh`.

**Why the claim is taken early.** The maintainer's follow-up comment on the
issue reports the same collision reaching commits, not just comments, and
argues the expensive part is the twenty minutes of diagnosis before a session
has anything to claim about. Both the script header and the AGENTS.md
paragraph say to claim when you start reading the pull request, not when you
are ready to write.

**Not a lock.** Two sessions that check inside the same second both proceed,
exactly as with `issue-claim.sh`. The finding gate is the stronger half: it
compares against what is already published, so it holds even when the claims
raced.

---

### DIFF
--- a/.github/workflows/guard-self-tests.yml
+++ b/.github/workflows/guard-self-tests.yml
@@ -296,6 +296,10 @@ jobs:
         if: ${{ !cancelled() }}
         run: ./scripts/test-issue-claim.sh
 
+      - name: the pull-request claim and the duplicate-finding gate
+        if: ${{ !cancelled() }}
+        run: ./scripts/test-pr-claim.sh
+
       - name: the CodeQL canary opens, refreshes and closes a codeql-red issue
         if: ${{ !cancelled() }}
         run: ./scripts/test-codeql-canary.sh
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -454,8 +454,12 @@ The claim half takes the rules above: a lapsing comment, a session word beside
 the login, and every unknown proceeding loudly. A merged or closed pull request
 stands a sweep down too, since neither one takes a comment.
 
-`post` is the half that would have stopped all three. It asks whether the same
-finding already stands, and writes only when it does not. The caller names the
+`post` is the half that would have stopped all three. It asks the claim
+question too, then asks whether the same finding already stands, and writes
+only when neither answer is yes — the finding question alone can only see what
+somebody already published, so it turns a second sweep away after it has spent
+its diagnosis rather than before. `--ignore-claim` posts over a live claim, for
+the operator who read it and said on the pull request why. The caller names the
 finding, because two sessions that find one thing write it up two ways, and a
 digest of the words would call them different. A key that has to go stale
 carries what it depends on: a finding about the head commit puts the head sha
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -427,13 +427,46 @@ It carries the same session word, and for the same reason: one login is one
 account, and the thing that has to be identified is the session. Comparing
 the login alone read a peer session's claim as this session's own and cleared
 both to work one issue (`#5875`). `scripts/lib/claim-session.sh` resolves the
-word for both scripts, so a fleet sets `STELLA_CLAIM_SESSION` once rather
-than twice, and both fail open the same way: a run with no word of its own,
-and a claim comment carrying none, fall back to the author-only rule.
+word for every claim script here, so a fleet sets `STELLA_CLAIM_SESSION` once
+rather than once per script, and they all fail open the same way: a run with
+no word of its own, and a claim comment carrying none, fall back to the
+author-only rule.
 
 Run it **before writing code and again before opening the PR**. The gap
 between those two is enough: a peer's PR can merge inside one issue's worth of
 work, and then the check that was clean at the start is stale at the end.
+
+**A pull request is the third thing two sessions take at once, and
+`scripts/pr-claim.sh` is that mechanic pointed at one.** Three sweep sessions
+read one pull request in eight minutes, each worked out the same merge
+conflict, and each posted it; a fourth stopped, because it read the comments
+first. Duplication here does not lose work, it publishes it, and the
+maintainer is left to diff three long comments to learn they say one thing
+(`#5861`).
+
+```bash
+./scripts/pr-claim.sh check 5835    # exit 0 proceed, 1 stand down
+./scripts/pr-claim.sh claim 5835    # check, then post the claim
+./scripts/pr-claim.sh post 5835 --finding conflict:5828 --body-file note.md
+```
+
+The claim half takes the rules above: a lapsing comment, a session word beside
+the login, and every unknown proceeding loudly. A merged or closed pull request
+stands a sweep down too, since neither one takes a comment.
+
+`post` is the half that would have stopped all three. It asks whether the same
+finding already stands, and writes only when it does not. The caller names the
+finding, because two sessions that find one thing write it up two ways, and a
+digest of the words would call them different. A key that has to go stale
+carries what it depends on: a finding about the head commit puts the head sha
+in the key. A read that failed says so and posts anyway, rather than reporting
+that nothing stands.
+
+Take the claim when you start to read the pull request, not when you are ready
+to write. Each of those three sessions spent twenty minutes on the conflict
+before it had anything to say, so a claim taken at the end would have saved
+none of it. `make pr-claim N=5835` asks by hand; `make pr-claim-test` covers
+both gates, the blocking branches included.
 
 An eighth, `windows-check.yml`, is the only compiler in this project that
 looks at a `#[cfg(windows)]` arm: `ci.yml` runs on `ubuntu-latest` and
--- a/Makefile
+++ b/Makefile
@@ -1000,6 +1000,14 @@ issue-claim: ## Ask whether somebody is already implementing an issue: make issu
 issue-claim-test: ## Test the issue-claim pre-flight, standing-down branch included (hermetic; not part of `gate`)
 	./scripts/test-issue-claim.sh
 
+.PHONY: pr-claim
+pr-claim: ## Ask whether somebody is already sweeping a pull request: make pr-claim N=5835
+	@./scripts/pr-claim.sh check $(N)
+
+.PHONY: pr-claim-test
+pr-claim-test: ## Test the pull-request claim and the duplicate-finding gate (hermetic; not part of `gate`)
+	./scripts/test-pr-claim.sh
+
 # Not a GATE_STEPS member -- one file, one narrow shape (#3459). The real
 # enforcement is its self-test's own live-tree case (F in
 # test-dependabot-pip-dirs.sh, "this repository's real dependabot.yml
--- a/scripts/lib/claim-session.sh
+++ b/scripts/lib/claim-session.sh
@@ -2,7 +2,8 @@
 #
 # The session word a claim carries beside its author.
 #
-# `scripts/main-red-claim.sh` and `scripts/issue-claim.sh` both read it.
+# Every claim script here reads it: `scripts/main-red-claim.sh`,
+# `scripts/issue-claim.sh` and `scripts/pr-claim.sh`.
 #
 # ## Why a login is not enough
 #
--- a/scripts/pr-claim.sh
+++ b/scripts/pr-claim.sh
@@ -448,16 +448,14 @@ report_held() {
   echo "         reason rather than a collision." >&2
 }
 
-# ── post: has this finding already been published? ───────────────────────────
+# ── post: may this finding be published here? ────────────────────────────────
 #
-# This gate runs on its own. A sweep that means to publish asks this and
-# nothing else, because a finding that already stands is a duplicate whoever
-# holds the pull request.
+# Two questions, in this order. Does a peer hold the pull request, and does
+# this finding already stand? The claim is asked first because it is the one
+# that can turn a sweep back before it spends its diagnosis. The finding is
+# asked second because a published copy costs a reader whoever holds the claim,
+# so it stops a post that this session's own claim covers.
 if [ "$mode" = "post" ]; then
-  # The claim gate runs here too. A finding gate on its own can only see what
-  # is already published, so two sweeps that both begin before either posts
-  # each spend a full diagnosis, and only the second is turned away. A claim
-  # taken at the start turns the second one back before it spends anything.
   if [ "$ignore_claim" -eq 0 ]; then
     scan_claims
     if [ -n "$held_by" ]; then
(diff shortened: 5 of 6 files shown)


macanderson:
Answering Sourcery's ❌ row rather than fixing it — it's describing a race this PR's own description names and #5861 scopes out.

Sourcery's objection: *"two sessions can read the comments concurrently before either posts... both can still publish the same finding."* That's true, and it's not a gap — it's shape (3) from #5861's "Possible shapes" list ("Serialize the routine so only one runs at a time"), which the issue explicitly did not ask for: *"(1) and (2) compose; (1) alone would have prevented all three duplicates here."* This PR's own "Anything reviewers should know?" section says the same thing under "Not a lock": *"Two sessions that check inside the same second both proceed, exactly as with `issue-claim.sh`. The finding gate is the stronger half: it compares against what is already published, so it holds even when the claims raced."*

#5861's DoD (already checked in the issue body) asks for "at most one comment for a given finding" via the post-time marker check — not a distributed lock preventing two concurrent checks from both deciding to post before either write lands. `scripts/test-pr-claim.sh`'s cases (`a finding already posted stops a second copy`, `an old finding still stops a copy`) cover exactly that scoped guarantee, which is what closes #5861.

A true mutual-exclusion lock is a heavier mechanism than a comment-based claim can give without new infrastructure (a real lock service), and #5861 discusses that tradeoff explicitly rather than asking for it. No change follows from this row.

---

macanderson:
## Answering Sourcery's ❌ row

> two sessions can read the comments concurrently before either posts, and the
> PR explicitly acknowledges this race; both can still publish the same finding

Half fixed, half answered.

**Fixed** (`4b7437c24`). The row is right that a finding gate alone leaves a
window, and it was wider than the sub-second race the row describes. Each of
the three duplicate sweeps posted at the *end* of its diagnosis, so for the
twenty minutes each spent reading the conflict there was nothing published for
the next one to find. `post` now runs the claim check first and stands down on
a peer's live claim, so a second sweep is turned back at the start rather than
after it has spent the work. `--ignore-claim` is the way through for an
operator who read the claim and said on the pull request why. Eight cases
cover it, including both fail-open directions and the case where a session's
own claim must not block its own post.

**Answered.** What is left is two sessions whose reads land inside the same
second: both see no claim, both claim, both post. That cannot be closed by a
comment thread — an at-most-once guarantee needs a lock, and GitHub comments
are not one. The issue says as much (it calls shapes 1 and 2 composable, and
neither a lock), and it names the property that must not be traded away for
one: these scripts fail open at every unknown, because a check that can block
a sweep is worse than the duplication it prevents. A lock inverts exactly
that. `scripts/issue-claim.sh` and `scripts/main-red-claim.sh` carry the same
residual race for the same reason, and this script is their sibling by design
rather than a stronger thing.

The window this leaves is one second wide instead of twenty minutes. Closing
the remaining second means a real lease table or serialized sweeps — the
issue's shape 3 — which is a maintainer's call, not this PR's.

macanderson:
## Restored this PR's description — it had been overwritten with #6381's, and was closing #5863

The body on this PR described the `loop_detect` self-appending rung and carried
`Closes #5863`. Its diff, title, branch and every commit trailer are the
pull-request claim work for #5861:

```
$ git log --format=%b origin/main..origin/fix/5861-pr-claim-and-duplicate-finding-gate | grep -E '^(Closes|Refs) #'
Refs #5861
Refs #5861
Closes #5861
```

`get_files` on this PR returns six paths — `scripts/pr-claim.sh`,
`scripts/test-pr-claim.sh`, `AGENTS.md`, `Makefile`,
`scripts/lib/claim-session.sh`, `.github/workflows/guard-self-tests.yml` — and
no `crates/` path at all. So the body was the only thing saying #5863.

**That was not cosmetic.** GitHub had already registered the link: #5863's
`closed_by_pull_requests` listed **two** PRs, #6381 and this one. Merging this
would have closed a P1 issue whose rung it does not implement, and #6381 —
which does — would have merged against an issue already closed. That is the
"an issue whose work shipped, staying open, with nothing to notice it but an
audit" failure AGENTS.md § "Closing the issue on merge" is about, arrived at
from the other end.

Restored the description from this PR's own history, and re-read the issue
afterward: `closed_by_pull_requests` is now #6381 alone.

Two things carried forward rather than restored verbatim, since `4b7437c`
landed after the body was written: `post` now runs the claim check before the
finding check, and `--ignore-claim` is the way through. Both are named in the
`post` bullet and the witness section.

No commit was pushed and no code changed — the diff was already right; only
its description was not.

---

Issue #6378, opened 2026-09-07
The model is never told it is close to its spend limit, so an enforced budget throws the work away


macanderson:
## The problem

A turn with an enforced spend limit is cut off the moment the limit trips. The
model is never told the limit is close. So it is still mid-investigation when
the turn ends, with a partial answer it would have written down had anything
asked it to.

The wall-clock deadline used to work the same way, and `#2957` fixed it. At
ten, five and two minutes left the engine appends one short user-role message
telling the model to stop widening the search and deliver what it has
(`crates/stella-core/src/driver/deadline_notice.rs`). Money has no equivalent.

Read the code:

- `crates/stella-core/src/budget.rs` — `BudgetGuard::evaluate` returns
  `BudgetOutcome::AbortTurn` under `BudgetMode::Enforced` and
  `BudgetOutcome::Warn` under `Observed`. `headroom_usd` already knows how much
  is left.
- `crates/stella-core/src/driver/settlement.rs` — `emit_budget_warning` sends
  the warning as an `AgentEvent::Error` with `retryable: true`. That reaches
  the event stream, which is the operator. Nothing puts it in front of the
  model.
- `crates/stella-core/src/driver.rs` — `check_budget` aborts at the step
  boundary with no message added to the transcript first.

`doc:adr/0032-a-loop-rung-reads-structure` is where this was noticed. That ADR
decides that a turn making varied but useless calls is bounded by the budget,
the deadline and the goal predicate rather than by a loop rung. Two of those
three hand back a partial answer. The budget does not.

## Why it is not fixed in that ADR's pull request

Changing what the model is told mid-turn is a measured change here. `#2957`
shipped its notice against three named benchmark failures. Picking the points
at which a spend notice fires, and showing it does not make a turn quit early,
needs a bench run with real spend. That is the "needs a rig or spend" case in
AGENTS.md § "Fix over file".

## Pillar

Reliability. An enforced budget currently discards the work it stops, in the
same way a deadline did before `#2957`.

## Definition of done

- [ ] A pure, latched notice module beside `driver/deadline_notice.rs` (e.g.
      `driver/budget_notice.rs`), keyed on remaining spend rather than
      remaining time, firing at most once per threshold per turn so the
      transcript prefix stays byte-stable (AGENTS.md invariant #7).
- [ ] It fires only where the guard would actually cut the turn — an armed
      limit under `BudgetMode::Enforced`. An `Observed` limit warns the
      operator and stops nothing, so a notice there would be false.
[...]

Issue #5865, opened 2026-09-05
Bench trials that scored a win still contain long stretches of grinding, so 'winner' is a bad label for loop studies


macanderson:
## The problem

Two loop-detection studies (#3292) have now been asked for the same evidence:
a threshold that **never** fires on the working population. Both used the
local `~/.arenabench/matches` corpus and both labelled a whole trial by its
reward — reward 1.0 means working, reward 0.0 with an `interrupted` status
means stuck.

That label is wrong, and it is wrong in the direction that makes the bar
unreachable. A trial can finish the task and then grind for hundreds of steps.
The verifier still scores it 1.0, because the deliverable is on disk. Every
one of those grinding steps then counts as evidence that a detector must not
fire.

## The proof

`build-pov-ray__3DYrafv`, reward **1.0**, 307 calls.
`~/.arenabench/matches/14d0127b8ca7/jobs/14d0127b8ca7-stella-kimi-k3/build-pov-ray__3DYrafv/agent/stella-events.jsonl`

- Calls 94 and 106 install and verify the deliverable, the `povray` binary.
- Calls 196 through 302 — 107 steps — each re-send one command that is the
  previous command plus one more copy of the same twelve-file `grep` block.
  The command grows from 263 characters to 8774. Nothing is written. Nothing
  is learned.
- The trial scored 1.0.

That is 35% of the trial's steps, inside the population a detector is required
to never touch. And it is the single false positive that stopped the
workspace-immobility signal clearing its bar: at a trailing-immobility
threshold of 20 calls the signal fires on 17 of 138 stuck trials and exactly
one winner, and the winner is this trial.

Full numbers:
https://github.com/macanderson/stella/issues/3292#issuecomment-5549705334

## What to do about it

Label **windows**, not trials, and label them by what the model was doing
rather than by what the verifier scored. #3292's own definition of done asks
for exactly this ("a labelled corpus of trace windows (stuck / working)"), and
both studies so far have substituted trial-level reward for it because it is
free.

A starting set, hand-labelled and committed as fixtures:

- `build-pov-ray__3DYrafv` calls 196–302 — **stuck**, inside a winning trial.
- `feal-linear-cryptanalysis__s5TVw8A` calls 19–47 — stuck, in a losing trial,
  same shape.
- `extract-moves-from-video__3AqwMdY` — the trace #3292 names, 62 calls.
- Working windows drawn from the winners, checked by hand for the shape above
  before they are used as working.

[...]

Issue #2957, opened 2026-08-12
bench: the task deadline stopped being armed between 2026-08-09 and 2026-08-11 — 130/130 ticks armed on Aug 9, 0/1087 on Aug 11


macanderson:
## Problem

Stella's wall-clock task deadline is **not armed on any ArenaBench trial**, so
nothing ever stops a bench run at a safe boundary and every over-running trial
dies on Harbor's `SIGKILL` instead of ending as a scorable result. This is the
third recurrence of the same shape (#2135, #2242): the mechanism is built,
wired, documented, carries a 60s reserve — and receives no value.

Measured on ArenaBench run `w10p` (SUT `main@554bb4f`), all 20 pipeline trials:

```
1087 budget_tick events
   0 carrying deadline_remaining_ms
```

Mirror: `~/.claude/jobs/e286a51b/tmp/w10-mirror/w10p/*/*/events.jsonl`.

### The evidence distinguishes the claim from its opposite

`BudgetGuard::tick_event` (`crates/stella-core/src/budget.rs`) populates the
field straight from `deadline_remaining(now)`, which is `None` **only** when
`task_deadline` is unarmed. So an absent field is positive proof of an unarmed
deadline, not merely a missing record. #2240 added the field for exactly this
purpose; it is doing its job.

### Not a historical gap

`bench/harbor_adapter/stella_harbor/turn_budget.py`, including #2242's
`discover_trial_deadline`, was present in the tree at the SUT commit:

```
git cat-file -e 554bb4f77:bench/harbor_adapter/stella_harbor/turn_budget.py   # exists
```

## Why it fails open

Steps 1–3 are read off the code; step 4 is the inference they support and is the
one thing still to confirm against a real trial `config.json`.

1. `arenabench/arenabench/runner.py:998` builds the Harbor invocation as
   `["--dataset", match.dataset.harbor_id]` — a **registry ref**
   (`terminal-bench-2.1@<digest>`), not an offline export directory.
2. `turn_budget.py::_task_agent_timeout` returns `None` when the trial config's
   `task.path` is falsy. Its docstring states the case outright: "a dataset
   resolved from a registry ref rather than an offline export has
   `task.path: null`, and the task definition is then not on this host at all."
3. With no base timeout, `trial_agent_deadline` returns `None`,
   `resolve_turn_budget` omits `--turn-budget`, `Config::turn_budget` stays
   `None`, and `stella_cli::runtime::one_shot_budget_guard` never calls
[...]

PR #6380, opened 2026-09-07
docs(adr): a loop rung reads structure, not progress

### LINKED ISSUE #3292: Loop detection cannot catch a model that grinds through 62 different commands without making progress

macanderson:
## What & why

`#3292` asked for a sixth rung in `crates/stella-core/src/loop_detect.rs` that
fires on a turn making varied calls that get nowhere, with one recorded bench
trial (`extract-moves-from-video`) as the witness it had to fire on.

Three sessions have now measured the three candidate signals the issue names,
each against the same 497-trial corpus, and each is on the issue with numbers:

| signal | result |
|---|---|
| repeated output lines | the stuck turn's median sits *below* the winners' median at every window; a cut-off that never fires on a winner catches nothing |
| no file changed | zero-false-positive cut-off is 202 calls in a row and catches nothing; the sharper trailing shape does not fire on this trace, whose last call writes a file |
| the plan not moving | narration: the two groups agree to two decimals. reasoning text: 0.384 has no false alarm but catches 4 turns in 115, and this trace is not one of them |

The one shape that does separate — a command re-sent with one more copy of
itself (`#5863`, being implemented separately) — also does not fire on this
trace. And the corpus cannot answer the issue's own zero-false-positive
question, because a reward-1.0 trial in it contains 107 grinding calls in a
row (`#5865`).

Reading the trace says why. Each command is a fresh idea, each reasoning block
names a new next step, and the model is debugging an approach that will not
work. The difference between that and working is what the work is *for*, and
that is not in the call stream.

**This PR decides it rather than measuring a fourth signal**, per SCR-002.

`docs/adr/0032-a-loop-rung-reads-structure.md` records the decision:

- A rung keys on a property of the recorded calls and answers a reader can
  check by looking — a repeat, a cycle, a constant answer, a wrapped sweep,
  one call grown out of the one before it. It never asks whether the turn is
  getting closer to the goal. That is a judgement about the task,
  `loop_detect` is a pure synchronous function over owned data (AGENTS.md #2),
  and answering it well needs a model call no verdict here may buy.
- The shape `#3292` names is therefore out of scope for this module,
  permanently — not unmeasured and not waiting on a better metric.
- What bounds it is what `doc:adr/0031-a-turn-has-no-step-cap-by-default`
  already chose: the budget, the wall-clock deadline and the host's goal
  predicate. The deadline also tells the model to hand in what it has
  (`driver/deadline_notice.rs`), which is what turns a bound into a partial
  answer rather than nothing.
- A new rung's evidence is hand-read windows, not trial reward, because a
  winning trial can contain grinding.

The code half makes the decision checkable. The trial's 62 calls and answers
are committed as `loop_detect/tests/extract_moves_from_video.json`, and
`loop_detect::tests::grind` holds three tests over it. The module doc of
`loop_detect.rs` carries the same rule, so the next author reaching for a rung
reads it in the file.

**What I gave up.** Shipping the reasoning-similarity signal would have added a
rung with zero false alarms on this corpus and 3.5% recall. I refused it: it
reads provider-supplied reasoning text, which several providers never emit at
all (`ReasoningPosture::Unsupported` in `provider_parity.rs`), so the rung
would be silently inert for most routes; its four hits look like `#5863`'s
shape seen a second way; and it does not fire on the witness either. A rung
that cannot fire on most providers is worse than none, because it reads as
coverage.

Closes #3292

## The witness

Three tests in `crates/stella-core/src/loop_detect/tests/grind.rs`, all new
with their fixture, so "fails on `main`" means the fixture and the pin do not
exist there:

- `the_recorded_grind_varies_both_its_inputs_and_its_outputs` — 62 calls, all
  `bash`, **59** distinct inputs and 62 distinct outputs. This corrects the
  issue's own census, which says 62 distinct inputs; three commands are
  re-sent. It changes no conclusion (every input-keyed rung also needs an
  unchanged answer, and all 62 differ) but it is the kind of premise the next
  session would have built on.
- `no_rung_fires_at_any_step_of_the_recorded_grind` — `detect_loop` is asked
  the question the driver asks it, over every prefix of the turn, under the
  shipping `LoopDetectionConfig::default()`. Sixty-two `NoLoop`s. This is the
  pin: a sixth rung that fires here reddens it by name.
- `the_fixture_is_detectable_once_its_answers_stop_changing` — anti-vacuity.
  The same fixture with its answers flattened to one constant returns
  `Stagnant { count: 62 }`, so the pin above reports a property of the trace
  and not a broken detector.

Ran: `cargo test -p stella-core --lib loop_detect` — 75 passed, 0 failed.

## The gate

- [x] `cargo fmt --check` (scoped: `-p stella-core`, clean)
- [x] `cargo clippy -p stella-core --all-targets -- -D warnings` — clean
- [x] `cargo test -p stella-core --lib loop_detect` — 75 passed
- [x] `make guards-fast` — exit 0
- [x] `make doc-warnings CARGO_SCOPE="-p stella-core"` — clean
- [x] `python3 scripts/check-prose.py`, `make line-citations`,
      `make doc-links`, `make adr-numbering` — all OK
- [x] Docs updated: ADR 0032, `docs/adr/README.md` index row and summary,
      `docs/manifest.json` entry, and `loop_detect.rs`'s module doc
- [x] `Closes #3292` appears both above and as a commit trailer

The workspace build and test suite are CI's, per CLAUDE.md — this laptop ran
only the scoped commands above.

## Fix over file

- [x] Extra fixes in this PR: the census correction described under "The
      witness" — the issue and every comment on it repeat "62 distinct tool
      inputs", and the fixture makes it 59.
- [x] Filed, with the reason a fix could not ride this PR: **#6378** — an
      enforced spend limit cuts the turn with no warning to the model, so it
      throws away the partial answer the wall-clock deadline learned to ask
      for in `#2957`. Noticed while writing the ADR's "what bounds it instead"
      paragraph, because two of the three bounds hand back a partial and the
      budget does not. Deferred because changing what the model is told
      mid-turn is a measured change here — `#2957` shipped against three named
      bench failures, and picking spend thresholds needs a bench run with real
      spend (AGENTS.md's "needs a rig or spend" case).

## Ground-rule check

**The fixture is 90KB of recorded bench output.** It is the trial named in
`#3292`'s definition of done, and it lived only in one machine's
`~/.arenabench`. Committing the calls is what makes the claim this issue turns
on checkable by anyone.

**No engine behaviour changes.** The turn that prompted `#3292` still runs
until a bound the operator set ends it. That is the decision, not an omission.

**On `#5863`.** It is claimed by another session as I write this
(`./scripts/issue-claim.sh check 5863` stands a second session down), so its
rung is deliberately not in this PR. ADR 0032 says explicitly that it stays
open and stays in scope: a command that grows by repeating itself is
structure.

**Exemplar for the fixture-plus-pin shape:** `rust-analyzer`'s
`test_data`-backed parser tests, where the committed input is the claim and
the assertion is what stops it drifting.

## Summary by Sourcery

Codify the boundary of loop detection around observable call structure and preserve the recorded grinding trace as a regression pin without changing engine behavior.

Enhancements:
- Define and document that loop-detection rungs evaluate observable call and answer structure rather than task progress, explicitly placing varied-call grinding outside the module’s scope.
- Record the measured rationale and chosen bounds for varied-call grinding in a new ADR and make the decision discoverable through the ADR index, manifest, and module documentation.

Documentation:
- Add ADR 0032 documenting the scope and evidence-based decision for loop detection, including why progress-oriented signals are not added.

Tests:
- Add a committed 62-call witness fixture and tests that verify its inputs and outputs vary, no existing rung fires across any prefix, and constant answers remain detectable.

Chores:
- Correct the witness census to reflect 59 distinct inputs among 62 calls.


macanderson:
Answering Sourcery's three ❌ rows rather than fixing them — all three read #3292's *original* DoD, which the issue body explicitly retracts.

#3292 was rewritten on 2026-09-07 (same day, before this PR was opened) under the heading "Definition of done — Rewritten 2026-09-07": *"The four boxes below asked for a detector plus evidence that it never fires on the winning population. Both halves of that ask are now known to be unreachable, and the boxes are restated to what the work can actually prove."* Each of the three objectives Sourcery quotes is explicitly struck through in the issue's own "The original four, and why each is retired" section:

- *"A pure detector in `loop_detect.rs` matching the existing five in shape."* → **"Refused, with the measurements."**
- *"Evidence that it fires on the stuck population and never on the working one."* → **"Unevaluable against this corpus... ADR 0032 replaces it with hand-read windows."**
- *"The `extract-moves-from-video` trace as a named witness fixture."* → **"Kept, with its role inverted: it is committed as the fixture proving the module's edge rather than proving a rung fires."**

The issue's current (rewritten) checklist — ADR recorded, measurements written up, fixture committed, pin test added, census corrected, module doc updated, bounds named and the one gap filed as #6378 — is what this PR delivers, and every box on it is already checked in the issue body. Sourcery's assessment table is evaluating this PR against a DoD the issue no longer asks for.

No code or doc change follows from this — the decision is the deliverable, and it's already recorded in ADR 0032 and the issue itself.

---

macanderson:
## Answering Sourcery's three ❌ rows

All three read `#3292`'s original objectives, and all three are correct that
this PR does not meet them. It refuses them, on the record, which is what the
PR is. Answering each so the next reviewer does not have to reconstruct it.

**Row 1 — "Add a pure sixth loop detector … that identifies varied commands
and varied outputs that nevertheless represent unproductive grinding."**

Refused, with measurements. Three candidate signals for exactly this class
were measured against the same 497-trial corpus, each in every formulation
tried, and none separates the stuck turn from the turns that won. The numbers
are on `#3292` in three study comments and are carried into
`docs/adr/0032-a-loop-rung-reads-structure.md § Context` so they are not run a
fourth time.

The reason they all fail in the same direction is stated in the ADR's
decision: what distinguishes this turn from a working one is whether the work
is approaching the goal, which is a judgement about the task. `loop_detect` is
a plain synchronous function over owned data (AGENTS.md #2), and no verdict
there may buy a model call. So the objective is not "not yet done" — it is not
buildable in that module, and ADR 0032 says so rather than leaving a fifth
session to rediscover it. That is SCR-002: decide it, record it as an ADR, and
the ADR replaces the question.

**Row 2 — "Provide a labelled corpus and evidence that the detector catches
the stuck population with no false positives on the working population."**

Unevaluable against this corpus, and that is a finding rather than a gap in
the PR. `build-pov-ray__3DYrafv` scored reward 1.0 and contains 107 grinding
calls in a row, made after its deliverable was already on disk (`#5865`). So
"no false positives on the working population" asks for a rule that never
fires on a turn that really was grinding. No metric can satisfy that, and
choosing a different metric does not help.

ADR 0032 replaces the bar rather than dropping it: a future rung ships when
every window it fires on across the corpus has been read by a person and is
grinding. `#5863`'s rung is the shape that clears it — three turns in 371,
each one read.

**Row 3 — "Ensure the `extract-moves-from-video` trace is covered as a named
witness for the new detector."**

The trace is committed and covered, with its role inverted, and the inversion
is the deliverable. `crates/stella-core/src/loop_detect/tests/extract_moves_from_video.json`
holds all 62 calls and answers;
`no_rung_fires_at_any_step_of_the_recorded_grind` asks `detect_loop` the
question the driver asks it over every prefix of the turn and gets `NoLoop`
every time.

Sourcery is right that the issue asked for the opposite assertion. That
assertion cannot be written truthfully today, and pinning the true one is
worth more than leaving the trace in one machine's `~/.arenabench`: a sixth
rung that fires there now reddens this test by name, so firing becomes a
decision made in review instead of a side effect nobody noticed. The
`the_fixture_is_detectable_once_its_answers_stop_changing` control keeps the
pin from passing vacuously — the same fixture with its answers flattened
returns `Stagnant { count: 62 }`.

**The issue's Definition of done has been rewritten to match**, with the
original four items kept struck through underneath and a line each on why they
are retired. Nothing was ticked that is not true.

**What was deferred rather than decided**: `#6378` — an enforced spend limit
cuts the turn with no warning to the model, so it throws away the partial
answer the wall-clock deadline learned to ask for in `#2957`. It is the one
hole found in the bounds this ADR leans on. It is filed rather than fixed
because changing what the model is told mid-turn is a measured change here,
and picking spend thresholds needs a bench run with real spend.

Issue #1851, opened 2026-08-06
core: all three loop detectors require a contiguous trailing run — one interleaved call blinds them


macanderson:
## Problem

`crates/stella-core/src/loop_detect.rs` (~418-540): `detect_exact_repeat` (rev().take_while), `detect_short_cycle` (break on first mismatch), and `detect_stagnation` all scan a contiguous suffix. An agent alternating `bash: cargo test` (identical failing output) with a varying `read_file` each round has period 2 with one varying element — every detector resets at offset 1 and the agent spins to max_steps with no steer.

This is the shape real wedges take (re-run the failing test, peek at a different file, repeat). Terminal-Bench journals (see the loop-benchmark memory trail, #1524/#1744 investment in loop identity) can confirm frequency.

## Fix direction

A histogram/majority rung: over the last N call records, if any single (name, input-identity) pair repeats ≥ K times with identical result identity — regardless of interleaving — escalate the same way stagnation does. Keep the existing detectors; this is an additional rung, not a redesign. Property tests belong in the existing loop_detect proptest suite.

## Verify

Witness: an A-B-A-B-A-B history (A identical results, B varying) that today returns no verdict must trip the new rung; a genuinely progressing interleave (A's results all differ) must not.

## Constraints

stella-core, pure functions only (invariant 2). False positives steer (never abort) per the escalation ladder — check driver/loop_escalation.rs for the consequence path.

macanderson:
**Triage 2026-08-06:** P1 — the A-B-A-B wedge (re-run the failing test, peek at a different file, repeat) is the documented shape real loops take, and one interleaved call resets all three detectors, so the agent burns to max_steps with no steer. Plan: as specced; next step is the histogram/majority rung as an additional detector — pure function over owned data (invariant 2) — with the A-B-A-B witness plus the progressing-interleave negative, property coverage in the existing loop_detect proptest suite, and the consequence wired through driver/loop_escalation.rs as a steer, never an abort.

PR #6381, opened 2026-09-07
feat(stella-core): catch the command that keeps re-sending itself

### LINKED ISSUE #5863: Stella keeps re-sending one command with one more copy of itself glued on, and no loop rung stops it

macanderson:
## What & why

A model gets stuck and keeps sending one shell command. Each step it re-sends
the whole of the last command with one more copy of a block it already carries
glued on the end. The command grows, its answer grows, and nothing is learned.

None of the five rungs in `loop_detect` fires. The three repeat rungs and
stagnation all need bytes that repeat, and every input and every output here is
new. The sweep rung needs a number that climbs and wraps, and a growing command
carries none.

`build-pov-ray__3DYrafv` in the arenabench corpus ground this way for 107 of its
307 calls, from a 263-character command to an 8,774-character one. It still
scored a win, because its real work had landed a hundred calls back.

The sixth rung lives in `crates/stella-core/src/loop_detect/append.rs`, beside
`sweep.rs` and for the same reason: it reads no output. `detect_self_appending`
reports `LoopVerdict::SelfAppending` when the window ends in a run of calls to
one tool, each of which sent the whole of the call before it and added more.
Growth is read key by key over the JSON input, so a shell command growing inside
`{"command": ...}` counts the same as any other argument that grows — comparing
the serialized input would miss it, since a closing quote and brace sit between
the old command and the new suffix.

Closes #5863

### The two judgements the issue asked to be stated

**Prefix, not containment.** A call extends its predecessor when both inputs
carry the same keys and each value is unchanged or a string the older one
starts, with at least one grown. An addition pasted into the *middle* of a
command is missed. Prefix growth is the shape the corpus was scored on, so the
false-alarm counts in the issue are counts about it. Containment is a wider net
with no measurement behind it, so it would need its own false-alarm check before
it could ship.

**Threshold 8.** Over 371 scored trials every floor from 3 to 20 fires on the
same three trials, so the corpus cannot pick the number — it is picked for what
it says. Three calls is a model building a pipe: write it, add a filter, add a
count. Eight calls, seven of them re-asking for everything the last one asked
for, is not that under any reading. Eight is also `monotonic_sweep_threshold`,
the other rung that reads no output and whose floor does the same job.
`self_appending_threshold` joins `LoopDetectionConfig`; `0` or `1` turns it off,
like every rung beside it.

### Where it sits in `detect_loop`

Last, after the sweep. It reads neither the output nor a cursor, so it is the
loosest reading of a window, and anything the sweep rung can describe should be
reported as the sweep it is.

### Exemplar followed

`crates/stella-core/src/loop_detect/sweep.rs` — the in-tree model for a rung
that reads no output: its own submodule, its own module doc carrying the
discriminator argument, a `pub(super)` detector over `&[CallRecord]`, a threshold
on `LoopDetectionConfig`, and a `tests/<rung>.rs` witness file.

## The witness

`crates/stella-core/src/loop_detect/tests/append.rs`, built from the
`build-pov-ray__3DYrafv` shape: twelve `bash` calls whose commands are `base`,
`base + block`, `base + block + block`, … each answering differently.

Proved the artisanal way. With `self_appending_threshold` defaulted to `0` —
which is `main`'s behaviour, since no rung there keys on a growing input — four
of the eleven tests fail:

```
a_command_that_re_sends_itself_with_more_added_is_a_loop ... FAILED
the_evidence_names_the_growth_and_the_command_it_grew_from ... FAILED
one_grind_seen_twice_is_the_same_loop ... FAILED
two_commands_growing_from_different_roots_are_two_loops ... FAILED
test result: FAILED. 7 passed; 4 failed
```

With the default restored to `8`: `11 passed; 0 failed`. The witness also
carries its own anti-vacuity assertion — with the new rung disabled and the
other five at their defaults, `detect_loop` returns `NoLoop` on the same window.

Both directions the issue asked for are covered:

- `a_run_whose_inputs_do_not_nest_never_self_appends` and
  `a_command_that_shrinks_never_self_appends` — inputs that do not nest never
  report `SelfAppending`.
- `a_growing_command_with_identical_outputs_stays_stagnation` — a growing
  command whose answers are byte-identical is reported as `Stagnant`, and stays
  that way.

`detect_loop_never_panics` and `short_history_is_always_no_loop` now sweep
`self_appending_threshold` alongside the other four.

## The gate

- [x] `cargo fmt --check` — via `make guards-fast`, green
- [x] `cargo clippy -p stella-core -p stella-protocol --all-targets -- -D warnings` — green
- [x] `cargo test -p stella-core` — 889 passed, 0 failed
- [x] `make guards-fast` — exit 0, `check-prose` and `check-file-size` included
- [x] `RUSTDOCFLAGS="-D warnings" cargo doc -p stella-core --no-deps --document-private-items` — green
- [x] `make wire-schema-update` run; `docs/wire/` regenerated and committed
- [x] Docs updated: `loop_detect.rs`'s module doc lists six failure modes;
      `AgentEvent::LoopDetected`'s `kind` doc names `"self_appending"`
- [x] `Closes #5863` appears above **and** as a commit trailer

The workspace build and test are CI's, per CLAUDE.md — this PR's evidence is its
CI run.

## Fix over file

- [x] Extra fixes in this PR: none

## Ground-rule check

- [x] No I/O added to `stella-core` — `detect_self_appending` and `extends` are
      plain synchronous functions over borrowed data; no new deps
- [x] No new cross-boundary type: `AgentEvent::LoopDetected::kind` is already a
      `String`, so `"self_appending"` is additive to the wire contract and needs
      no new tag or `consumers.rs` row

**`SelfAppending` carries a `base` field the issue's sketch did not name.** The
sketch was `{ tool, calls }`, but the same bullet asks for an `identity()` that
does not misname two different loops, and the identity it prescribes is "the
tool plus the shortest input in the run". Every input in the run differs by
construction, so naming the trailing one would call each step of one grind a new
loop and buy a fresh steering warning for every call. `base` is the run's first
and shortest input, which is stable for as long as the run is —
`MonotonicSweep` carries `target` for exactly this reason. Two witnesses pin it:
`one_grind_seen_twice_is_the_same_loop` and
`two_commands_growing_from_different_roots_are_two_loops`.

**`crates/stella-core/src/loop_detect/tests.rs` is now 1488 of its 1500 lines.**
`LoopDetectionConfig` is spelled out in full in thirty test literals, so a new
threshold costs a line at each. I tried extracting the interleaved-repeat
section into its own `tests/interleaved.rs` and reverted it: a moved file is a
new path to the prose ratchet, held to a grade-6 ceiling with no baseline entry
to carry over, and clearing that would have meant rewriting 215 lines of
somebody else's argument about the `#1851` rung to satisfy a readability score.
That would leave the tree worse than the crowding does. The durable fix is a
shared "every rung off" base the literals spread from, so the next rung costs
zero test edits — a refactor of thirty call sites that belongs in its own PR
rather than riding a rung.

## Summary by Sourcery

Add self-appending command detection to stop agents from endlessly growing and resending shell commands.

New Features:
- Detect self-appending command loops when consecutive calls to the same tool extend prior JSON string arguments by prefix growth.
- Expose self-appending loop verdicts through driver escalation and the existing LoopDetected event contract.

Bug Fixes:
- Prevent agents from repeatedly re-sending and growing an already completed command without triggering loop detection.

Enhancements:
- Add configurable self-appending detection with a default threshold of eight calls, preserving precedence for more specific loop verdicts and supporting loop identity based on the run's base input.

Documentation:
- Document the sixth self-appending loop failure mode and the new event kind in protocol and generated wire documentation.

Tests:
- Add witness and property tests covering detection thresholds, non-nesting and shrinking inputs, identical-output stagnation, interleaved tools, loop identity, and panic/short-history safety.


macanderson:
Merged `main` (`96bf08b27`, #6380's ADR 0032) into the branch.

The only conflict was the module list in `loop_detect/tests.rs` — #6380 added `mod grind;` where this branch adds `mod append;`. Both stay.

ADR 0032 anticipated this rung: it names the self-appending shape as the one signal that separates, says `#5863` carries it, and lists "one call that grew out of the one before it" among the structural properties a rung may read. It also claims the shape does not fire on the `extract-moves-from-video` trace its own pin guards. That is now checked rather than claimed — `no_rung_fires_at_any_step_of_the_recorded_grind` passes with `self_appending_threshold` at its default of 8 (`cargo test -p stella-core loop_detect`: 86 passed, 0 failed).

Two sentences that said "a sixth rung that fires here" — one in `loop_detect.rs`'s module doc, one in `tests/grind.rs`'s — now say "a new rung" and name this one as the first added since the pin, since a sixth rung exists as of this PR.

PR #6384, opened 2026-09-07
feat(stella-cli): re-open a driver session after it asks to sleep

### LINKED ISSUE #6283: Nothing re-opens a driver session after it asks to sleep, so the loop runs one issue and stops
## The problem

`stella plugin drive` opens one session and returns. A driver that answers `{"next": {"sleep": {"secs": 900}}}` is asking to be re-opened in fifteen minutes, and nothing re-opens it.

Until now that cost nothing, because a session could only read the queue. It costs something now: `backlog_claim` and the three `work` verbs are served, so one session takes an issue, works it to a diff, and stops. A person has to run the command again for the next issue, which is what `scripts/self-driving.sh` exists to avoid.

`crates/stella-cli/src/driver_plugin.rs`'s header says as much: "Scheduling is absent. One invocation opens one session and reports what the driver said to do next; who re-opens it after a `DriveNext::Sleep` is #3599's B2 `LoopStep` machine."

## What is missing

The `LoopStep` / `step` machine #3599 B2 names — pure, in `stella-autonomy`, deciding whether to open another session and when. `crates/stella-autonomy/src/lib.rs` already holds the cycle-sizing and dry-streak math the shell driver uses; this is the piece beside it that turns a sequence of sessions into a loop.

Constraints already discovered:

- `stella-autonomy` has no god files and must keep none, so this is a new module rather than an addition to `lib.rs`.
- The host clamps `DriveNext::Sleep`'s seconds (`stella_runtime::wrapper::MAX_SLEEP_SECS`), so a driver cannot buy an unbounded absence by asking for one. Whatever re-opens the session honours the clamped value, not the asked one.
- A loop that cannot halt is not autonomous. `DriveNext::Halt` must end the sequence, and the reason must reach the ledger `crates/stella-cli/src/driver_plugin/session_log.rs` already writes.
- A spend ceiling is per invocation today (`RunBudget` is built in `plugin_cmd::drive`). A loop of sessions needs one ceiling across all of them, or the cap means nothing.

## Definition of done

- [x] A pure decision function in `stella-autonomy` says whether to open another session and after how long, given the last session's `next` and what the run has spent. — `crates/stella-autonomy/src/drive.rs`'s `next_session(ending, run_so_far, limits)`, a new module beside `step.rs` rather than an addition to `lib.rs`. No clock, no randomness, no I/O; the precedence is stated in its own doc comment. The wait it returns is clamped to `DriveLimits::max_sleep_secs`, which `plugin_cmd::drive` fills from `stella_runtime::wrapper::MAX_SLEEP_SECS`, so the honoured value is the clamped one. Witnessed by `drive::tests::a_sleep_asks_for_another_session_after_the_wait_it_named` and `a_long_wait_is_cut_to_the_hosts_ceiling_and_a_short_one_is_not`.
- [x] A test drives several sessions from one invocation: sleep, re-open, halt, stop. — `driver_plugin::sequence::tests::one_invocation_sleeps_reopens_and_stops_on_a_halt`: one `run` call, three sessions, waits of 900s and 30s honoured as asked, the halt ends it. And end to end through the real transport against a real installed package: `plugin_cmd::tests::one_invocation_opens_a_session_for_each_sleep_the_driver_asks_for` reads three records back off disk, each with its own session id. That one was proved to fail on the old behaviour: with `sequence::run`'s `DriveStep::Open` arm replaced by an immediate return, it fails `left: 1, right: 3`.
- [x] The spend ceiling spans the whole sequence, with a test that a run of sessions cannot spend past it. — `RunBudget` moved behind an `Arc<Mutex<_>>` in `driver_plugin::work::SpawnedWorkRunner`, shared by every session of the run, and `plugin_cmd::drive` builds one and hands clones down. `driver_plugin::sequence::tests::a_run_of_sessions_cannot_spend_past_the_cap` drives four sessions spending $4 each under a $10 cap and stops at the third; `drive::tests::the_spend_cap_is_measured_against_the_whole_run` is the same rule at the pure layer.
- [x] A halt reaches `session_log` with its reason, and the sequence ends. — `driver_plugin::sequence`'s `record` writes every session through `session_log::record` as it ends, and `sequence::tests::a_halt_reaches_the_ledger_with_its_reason_and_ends_the_sequence` reads the file back: two records, the second `Halt { reason: "operator stop" }`, and nothing after it. `drive::tests::a_halt_outranks_every_ceiling_that_would_allow_another_session` is why no remaining money or remaining session can outvote it.
- [x] `stella plugin drive`'s help and `website/content/docs/commands/plugin.mdx` say what one invocation now does. — The `PluginCmd::Drive` clap doc comment now reads "open a session, re-open it whenever the driver asks to sleep, and stop when it halts", and names the four endings. The docs page gains the `--max-sessions N` synopsis, a table of the four ways a run ends, the ledger path, and the correction that `--spend-limit` caps the whole run rather than one session. The stale one-session claims in `driver_plugin`'s and `session_log`'s headers and in `plugins/stella-selfdriving`'s README, `plugin.toml` and `main.py` were fixed in the same change.

Shipped by #6384.

macanderson:
## What & why

`stella plugin drive` opened one session and returned. A driver that answered `{"next": {"sleep": {"secs": 900}}}` was asking to be woken in fifteen minutes, and nothing woke it. Now that `backlog_claim` and the three `work` verbs are served, that cost something real: one invocation took one issue, worked it to a diff, and stopped. A person had to run the command again for the next issue, which is what `scripts/self-driving.sh` exists to avoid.

Three pieces:

**`stella_autonomy::drive` — the decision.** `next_session(ending, run_so_far, limits)` says whether to open another session and after how long. Pure: no clock, no randomness, no I/O, so a test drives a sequence in no time at all and two hosts of the same loop cannot disagree about the policy. The precedence is written down in the function's own doc comment. A halt is read before every ceiling, so nothing outvotes it. The wait it returns is already clamped to `DriveLimits::max_sleep_secs`, which the CLI fills from `stella_runtime::wrapper::MAX_SLEEP_SECS` — the same ceiling the socket applies, named here so the function can be tested against it rather than trusting its caller. A new module rather than an addition to `lib.rs`, because `stella-autonomy` has no god files and must keep none.

**`driver_plugin::sequence` — the host.** It opens the session, adds what it cost to the run, and does what the decision says. `DriveHost` is the seam: starting a process, writing the ledger, waiting and printing all sit behind it, so the sequence tests use no subprocess and no clock. `PluginDriveHost` is the shipping side, and it re-serves the resolved driver per session so each one gets a fresh gate — a shared gate would hand the second session the first one's spent call ceiling and its refusals.

**The spend ceiling now spans the run.** `RunBudget` moved behind an `Arc<Mutex<_>>` shared by every session's `SpawnedWorkRunner`. A copy per session would have capped one session and nothing else, which is the defect `RunBudget`'s own header was written for, reached through a different door. `--max-sessions N` is the other bound, for a driver that only ever asks to sleep; without it a run ends only on a halt, a failure, or `--spend-limit`.

A halt reaches `.stella/private/driver-sessions.jsonl` with its reason, and so does every other session of the run — each with its own id, so a run nobody was watching reads back afterwards.

The exemplar for the port shape is this crate's own `driver_plugin::work::WorkRunner`: one trait, the side effects behind it, the shipping implementation and a scripted one for tests.

Closes #6283

## The witness

Three, at three altitudes:

- `stella_autonomy::drive::tests::a_sleep_asks_for_another_session_after_the_wait_it_named` — the module did not exist, so nothing answered "open another one?".
- `driver_plugin::sequence::tests::one_invocation_sleeps_reopens_and_stops_on_a_halt` — one `run` call, three sessions, waits of 900s and 30s honoured as asked, halt ends it.
- `plugin_cmd::tests::one_invocation_opens_a_session_for_each_sleep_the_driver_asks_for` — the end-to-end one. One `drive` call against a real installed package, through the real transport, with a driver that only ever asks to sleep. It reads three records back off disk, each with its own session id.

**Proof the last one fails on the old behaviour**: with `sequence::run`'s `DriveStep::Open` arm replaced by an immediate return (one session per invocation, which is what `main` does), it fails `left: 1, right: 3`. Restored, it passes. Ran locally: `cargo test -p stella-cli --bins one_invocation_opens_a_session_for_each_sleep`.

Also covered: `a_halt_reaches_the_ledger_with_its_reason_and_ends_the_sequence`, `a_run_of_sessions_cannot_spend_past_the_cap` (four sessions at $4 under a $10 cap stop at the third), `a_failed_session_ends_the_run_and_is_still_recorded`, `the_session_ceiling_stops_a_driver_that_never_halts`.

## The gate

- [x] `cargo fmt --check` — via `make guards-fast`, green
- [x] `cargo clippy --workspace --all-targets -- -D warnings` — run scoped per CLAUDE.md ("CI builds and tests; this laptop does not"): `cargo clippy -p stella-autonomy --all-targets -- -D warnings` and the same for `stella-cli`, both clean. The workspace run is this PR's CI.
- [x] `cargo test --workspace` — again scoped: `cargo test -p stella-autonomy` (231 + 4 passed) and `cargo test -p stella-cli --bins` (3003 passed). Plus `make doc-warnings CARGO_SCOPE="-p stella-autonomy -p stella-cli"` under a clean target dir.
- [x] Docs updated where behavior/flags changed — `website/content/docs/commands/plugin.mdx` (synopsis, the four ways a run ends, `--spend-limit` now spanning the run), the `Drive` variant's clap help, `plugins/stella-selfdriving`'s README / `plugin.toml` / `main.py` headers, `crates/stella-autonomy/README.md`'s layout table, and `driver_plugin`'s own header, which used to say scheduling was absent.
- [x] CLA signed
- [x] `Closes #6283` appears both above and as a commit trailer

## Fix over file

- [x] Extra fixes in this PR: the prose claims that said one invocation opens one session — in `driver_plugin`'s header, `session_log`'s header, `plugins/stella-selfdriving`'s README, `plugin.toml` and `main.py`, and the docs page. Stale comments are bugs here, and these were about to become stale in the same change.
- [x] Filed, with the reason a fix could not ride this PR: **nothing was deferred**

## Ground-rule check

- [x] New cross-boundary types round-trip through serde (test included) — `drive::tests::the_types_round_trip_through_json` covers `SessionEnding`, `DriveStep`, `RunSoFar` and `DriveLimits`, every one of which crosses from `stella-autonomy` into `stella-cli`.

**`stella plugin drive` no longer returns after one session.** That is the change, and it is a behaviour change for anything scripting the verb. `--max-sessions 1` is the old behaviour exactly. The ways a run ends are the table in the docs page.

**A failed session ends the run rather than being retried.** The same host would meet the same failure the same way, and a driver that cannot start is not a driver that needs another try — retry policy for a transport fault would be a separate decision with its own evidence. The failure is still written to the ledger, and the command still exits non-zero.

**`session_ceiling` and `spend_cap` are checked after a session, never during one.** Invariant 6's boundary rule, one level up: a run is stopped between sessions so it never cuts a work turn down where it stands.

## Summary by Sourcery

Make `stella plugin drive` continue a driver run across sleep requests while enforcing run-wide limits and recording each session.

New Features:
- Re-open driver plugin sessions after sleep responses, allowing one `stella plugin drive` invocation to run a complete sequence until it halts or reaches a limit.
- Add `--max-sessions` to bound runs whose drivers continue requesting sleep.
- Persist every session and its ending in the driver session ledger with distinct session identifiers.

Bug Fixes:
- Ensure driver-requested sleep responses are acted on instead of leaving the driver dormant after the first session.
- Prevent spend limits from resetting between sessions by applying the budget across the entire run.

Enhancements:
- Separate pure session-sequencing policy from host-side process execution, waiting, output, and ledger effects.
- End runs deterministically on halts, failures, spend ceilings, or session ceilings, while clamping requested waits to the runtime maximum.
- Refresh per-session driver capabilities while sharing the run-wide spending budget.

Documentation:
- Update CLI, plugin, self-driving plugin, and autonomy documentation to describe multi-session driving, run endings, and the new session limit.

Tests:
- Add unit, sequence, ledger, CLI parsing, and end-to-end coverage for repeated sessions, waits, halts, failures, run-wide spending limits, and session ceilings.

### DIFF
--- a/crates/stella-autonomy/README.md
+++ b/crates/stella-autonomy/README.md
@@ -80,6 +80,7 @@ the CLI does — so the dashboard and the terminal cannot disagree.
 |---|---|
 | [`src/lib.rs`](src/lib.rs) | The dedup digest, the `AimdLimits`/`Calibration`/`calibrate` controller, the `Lens`/`Tooling`/`LENSES` aperture ladder and `advance`, `CycleRecord`/`dry_streak`, the `Supply`/`Demand`/`Floors`/`CyclePlan`/`plan_cycle` governor, `Metrics`/`metrics`/`starved`, `QueueIssue` (ranked by `priority::triage` — see that module), and `Liveness`/`liveness`/`RunRow`/`fold_runs`. |
 | [`src/step.rs`](src/step.rs) | The top-level loop machine: `LoopState`/`LoopObservation`/`step`, the `Unblock` remedies for a broken base, and the `Blocked` reasons. |
+| [`src/drive.rs`](src/drive.rs) | The session sequencer: `next_session` says whether a driver plugin gets another session and after how long, given how the last one ended and what the whole run has spent. `stella plugin drive` is its host. |
 | [`src/deliver.rs`](src/deliver.rs) | The per-PR delivery machine: `PrState`/`Observation`/`deliver_next`, the fix/rebase ceilings, and the escalation reasons. |
 | [`src/doctrine.rs`](src/doctrine.rs) | The policy knobs a human sets: `Doctrine`, `ForeignBreakage`, contention policy and `contention_verdict`. |
 | [`src/priority.rs`](src/priority.rs) | `triage`: rank by the ladder the repo actually uses, with unassessed issues distinct from unrankable ones. |
--- a/crates/stella-autonomy/src/drive.rs
+++ b/crates/stella-autonomy/src/drive.rs
@@ -1,0 +1,435 @@
+//! Whether to open another driver session, and when.
+//!
+//! A driver plugin ends each session with one word. `sleep` asks to be woken
+//! again. `halt` says it is done. Something has to hear that and act on it, or
+//! a driver that asks for fifteen minutes waits for ever. This module is the
+//! part that decides; the host it runs in does the waiting.
+//!
+//! # A halt has to stop the run
+//!
+//! [`SessionEnding::Halt`] ends the sequence. It is read before any other
+//! rule. Money left and sessions left cannot outvote it.
+//!
+//! # The ceilings span the run
+//!
+//! A spend cap that reset each session would cap nothing. Ten sessions under
+//! a ten dollar cap would buy a hundred dollars. So [`RunSoFar`] carries what
+//! the whole run has spent, and [`DriveLimits::spend_cap`] is read against
+//! that total. [`DriveLimits::session_ceiling`] does the same for how many
+//! sessions a run may open.
+//!
+//! # No clock, no sleeping
+//!
+//! [`next_session`] returns the wait in seconds and never takes one. That is
+//! what lets a test drive a sequence of sessions in no time at all, and it is
+//! this crate's own rule: the decisions are pure and the host does the I/O.
+//! The wait it returns is already clamped to
+//! [`DriveLimits::max_sleep_secs`], so a driver cannot buy an unbounded
+//! absence by asking for one.
+
+use serde::{Deserialize, Serialize};
+
+/// How the session that just ended ended.
+///
+/// A local type rather than `stella_plugin::DriveNext`, for the reason the
+/// loop machine keeps its own [`IssueRef`](crate::IssueRef): this crate
+/// depends on no workspace crate, so the host maps. It also holds one case
+/// the wire type cannot — a session that failed before it said anything.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case", tag = "ending")]
+pub enum SessionEnding {
+    /// The driver asked to be woken again after this many seconds.
+    Sleep {
+        /// What it asked for, before this module's ceiling is applied.
+        secs: u32,
+    },
+    /// The driver stopped, and said why.
+    Halt {
+        /// The sentence a human reads in the ledger.
+        reason: String,
+    },
+    /// The session never reached an answer: the process would not start, timed
+    /// out, died, or ended without saying what to do next.
+    Failed {
+        /// The host's own message, which already names the program.
+        message: String,
+    },
+}
+
+/// What bounds a run of sessions.
+///
+/// Every field is a ceiling. None of them is a floor: a driver that asks for
+/// less gets what it asked for.
+#[derive(Debug, Clone, Copy, PartialEq, Serialize, Deserialize)]
+pub struct DriveLimits {
+    /// The whole run's spend ceiling in USD, or `None` to spend unbounded.
+    ///
+    /// `None` is the observed mode `--spend-limit` already describes: spend is
+    /// still summed so a run can report what it cost, and nothing is refused.
+    pub spend_cap: Option<f64>,
+    /// The longest wait this host will honour between sessions.
+    ///
+    /// The host clamps a driver's request at the socket already
+    /// (`stella_runtime::wrapper::MAX_SLEEP_SECS`). Naming it here as well is
+    /// what makes the ceiling a fact this function can be tested against
+    /// rather than one it has to trust its caller for.
+    pub max_sleep_secs: u32,
+    /// How many sessions one invocation may open, or `None` for as many as the
+    /// driver asks for.
+    pub session_ceiling: Option<u32>,
+}
+
+/// What the run has done so far.
+#[derive(Debug, Clone, Copy, PartialEq, Default, Serialize, Deserialize)]
+pub struct RunSoFar {
+    /// How many sessions this invocation has already opened, the one that just
+    /// ended included.
+    pub sessions: u32,
+    /// What those sessions have spent between them, in USD.
+    pub spent_usd: f64,
+}
+
+/// Why a run of sessions ended.
+///
+/// Typed because the reader of a finished run has three different next moves.
+/// The driver decided, the host failed, or a ceiling was met.
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case", tag = "stop")]
+pub enum StopReason {
+    /// The driver halted, and this is what it said.
+    Halted {
+        /// The driver's own reason, carried through to the ledger.
+        reason: String,
+    },
+    /// A session failed before it could say what to do next.
+    Failed {
+        /// The host's message.
+        message: String,
+    },
+    /// The run met its spend ceiling.
+    BudgetSpent {
+        /// The ceiling.
+        cap: f64,
+        /// What the run had spent when it was met.
+        spent: f64,
+    },
+    /// The run opened every session it was allowed.
+    SessionCeiling {
+        /// How many that was.
+        ceiling: u32,
+    },
+}
+
+/// The run's next move.
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case", tag = "drive")]
+pub enum DriveStep {
+    /// Open another session, after waiting this long.
+    Open {
+        /// Seconds to wait first, already clamped to the host's ceiling. Zero
+        /// means open the next one now.
+        after_secs: u32,
+    },
+    /// Stop the run, and say why.
+    Stop {
+        /// What ended it.
+        reason: StopReason,
+    },
+}
+
+/// Decide what to do after a session ends.
+///
+/// Total over its inputs, with no clock, no randomness and no I/O, so the same
+/// ending under the same run always yields the same move.
+///
+/// # Precedence
+///
+/// The order is the policy, so it is written down rather than read out of the
+/// body:
+///
+/// 1. **A halt ends the run.** Checked first, so nothing outvotes it.
+/// 2. **A failed session ends the run.** A driver that could not answer cannot
+///    be asked to answer again by the same host in the same way.
+/// 3. **The spend ceiling ends the run**, measured against everything the run
+///    has spent rather than against one session.
+/// 4. **The session ceiling ends the run.**
+/// 5. **A sleep re-opens**, after the wait it asked for or the host's ceiling,
+///    whichever is shorter.
+#[must_use]
+pub fn next_session(ending: &SessionEnding, run: &RunSoFar, limits: &DriveLimits) -> DriveStep {
+    // 1 and 2. What the session itself decided outranks every ceiling. A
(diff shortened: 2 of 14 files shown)


macanderson:
### Answering the CodeQL `❌` — the three `rust/cleartext-logging` alerts are false positives

CodeQL flags three lines in `crates/stella-cli/src/driver_plugin/sequence.rs` (`PluginDriveHost::settle`'s three `println!`/`eprintln!` arms) as "writes `plugin_cmd::session_id(...)` to a log file", graded high severity.

**The value is not a secret.** `plugin_cmd::session_id` mints `drive-<unix seconds>-<4 hex of nanos XOR pid>`. It is a correlation token, and being readable is its entire job: the host sends it to the driver in the `DriveRequest`, the driver echoes it into its own records, and `session_log::record` writes it verbatim into `.stella/private/driver-sessions.jsonl`. A person reading a run's ledger joins it to the driver's logs on that string. Redacting it would make the ledger useless and would not protect anything, because nothing it protects exists.

**The lines are not new behaviour.** `plugin_cmd::drive` already printed the same value from the same source (`println!("session {session} ended: sleep {secs}s before the next one")`). CodeQL reports it now because the code moved into a new file, so the alert reads as introduced by this diff.

**The rule fires broadly here and is not a merge gate.** `CodeQL` is not one of `main`'s required status checks, and `gh api repos/macanderson/stella/code-scanning/alerts?state=open` currently returns 52 open `rust/cleartext-logging` alerts across `stella-tool-facts`, `stella-cli/main.rs`, `daemon/service.rs`, `self_driving_cmd.rs` and others — none dismissed. So this is a standing, untriaged class of finding rather than something this change introduced.

No code change here: the three sites print exactly what they should, and renaming the local binding would not move the alert, because the taint source CodeQL names is the `session_id()` function itself.

I have filed the backlog as its own handoff rather than dismissing 52 alerts from inside an unrelated PR — dismissing a security alert is a judgement per alert, and doing it wholesale here would hide the next genuine one. See the linked issue.

Filed as #6388.

macanderson:
## CodeQL is red on this PR: 3 new high-severity alerts in changed code

Automated PR sweep. This is the only red on `89f8558b` — `dod`, `cargo deny + cargo audit`, and the guard jobs are green; `fmt + clippy + test` was still running when I looked.

### What I observed

Check run [101671938920](https://github.com/macanderson/stella/runs/101671938920), conclusion `failure`:

> **3 new alerts including 3 high severity security vulnerabilities**
>
> \#\#\# New alerts in code changed by this pull request
> Security Alerts:
>  * 3 high

Branch alerts: [`pr:6384 tool:CodeQL is:open`](https://github.com/macanderson/stella/security/code-scanning?query=pr%3A6384+tool%3ACodeQL+is%3Aopen).

**This is not the `actions/untrusted-checkout` family #6374 just closed.** No workflow file is in this diff, and `plugins/stella-selfdriving/main.py`'s only change is prose in a docstring. The new code is Rust: `stella-autonomy/src/drive.rs` (+435) and `stella-cli/src/driver_plugin/sequence.rs` (+604).

### What I inferred, and could not confirm

I could not read the annotations — the tooling available to this sweep exposes the check-run summary but not its annotation list, so **the three alerts below are a hypothesis, not a reading of the alert text.** Check the security tab before acting on it.

The likeliest candidate is `temp_root` in `driver_plugin/sequence.rs`:

```rust
let root = std::env::temp_dir().join(format!(
    "stella-driver-sequence-{label}-{}-{:?}",
    std::process::id(),
    std::thread::current().id()
));
let _ = std::fs::remove_dir_all(&root);
std::fs::create_dir_all(&root).expect("temp root");
```

A predictable name under a world-writable shared directory, then `remove_dir_all` on it — the CWE-377 shape, and CodeQL rates that family high. `plugin_cmd/tests.rs` gains a matching `remove_dir_all(&root)`, which would account for more than one.

### Why I would not charge this PR for it

If that hypothesis is right, **the idiom is pre-existing and repo-wide, not something this PR introduced.** `std::env::temp_dir()` appears **112 times** across `crates/` on `main` today — `stella-store/src/tests.rs`, `sessions.rs`, `journal.rs`, `durable.rs`, `notify.rs`, `stella-serve/src/main.rs`, `stella-tools/src/{edit,write,rootfd}.rs` and more, several with the identical `join(format!("…{}", std::process::id()))` spelling. CodeQL calls these "new" only because the lines are new in this diff; `.github/workflows/codeql.yml` declares `build-mode: none` for all four languages with no path filter, so test helpers are analyzed like shipping code.

That makes it the shape AGENTS.md already reasons about for the file-size ratchet — a guard should judge your change, not the tree. Three options, in the order I would take them:

1. **Decide it repo-wide**: either exclude test paths via a CodeQL config, or adopt the `tempfile` crate's `TempDir` in a shared test helper and migrate the 112 sites over time. Either way it wants its own issue, not this PR.
2. **Fix just the new sites here** — swap `temp_root` for `tempfile::TempDir` — which clears this PR's alerts without touching the other 112. Smallest change that gets green, if CodeQL blocks the merge.
3. **Dismiss the three alerts** as test-only, if that is the standing call for this repository.

I have pushed nothing and dismissed nothing — this branch was updated ~20 minutes ago with CI still in flight, and option 1 is a maintainer decision.

---

macanderson:
## Correction: CodeQL is green. Disregard my previous comment.

I read an intermediate state and reported it as a verdict. Check run [101671938920](https://github.com/macanderson/stella/runs/101671938920) — the **same** run id I quoted — now reports:

> **No new alerts in code changed by this pull request** — conclusion `success`

When I read it, that run was `completed` with conclusion `failure` and the "3 new alerts including 3 high severity security vulnerabilities" summary. GitHub updates the aggregate `CodeQL` check in place as each language leg lands; `Analyze (rust)` was still running at 08:19 and finished around 08:30, and the check was re-evaluated against the full baseline afterwards. `completed` on that aggregate check does not mean every analysis has reported.

**This PR is now green**: `fmt + clippy + test` succeeded at 08:44:22, `cargo deny + cargo audit`, `dod`, CodeQL and the guard jobs all pass.

Everything downstream of the bad reading is withdrawn — there are no three alerts to triage, and the `temp_root` / CWE-377 hypothesis was speculation about alerts that do not exist. The one thing in it that stands on its own evidence, and that I am *not* asserting is a problem: `std::env::temp_dir()` appears 112 times across `crates/` on `main`, mostly as `join(format!("…{}", std::process::id()))` in test helpers. CodeQL is not complaining about it. Worth a look only if someone independently thinks it is worth a look.

Apologies for the noise on a green PR.

---

PR #6390, opened 2026-09-07
fix(stella-cli): a wrapper's verdict reaches the store on run and fleet

### LINKED ISSUE #6001: stella run/fleet --pipeline cannot publish a wrapper's judged verdict to the store
## Problem

`stella run --pipeline <plugin-id>` and `stella fleet --pipeline <plugin-id>`
cannot publish an `AgentEvent::Verdict` (or, it turns out, the pre-existing
`AgentEvent::GateBoard`) for the wrapper's judged outcome, even though the
code that sends both already exists at
`crates/stella-cli/src/wrapper_plugin.rs`'s `run_wrapped`.

Found while fixing #4224 (`stella goal --pipeline` gets a working producer in
that PR; `run`/`fleet` were split out here).

## Why

Each round of `stella run --pipeline` dispatches through
`crate::agent::run_turn` (`crates/stella-cli/src/agent/turn.rs`), which opens
its **own** execution row and closes its **own** event channel
(`persistence::close_event_stream`) before returning — this is documented
behavior, not a bug in that function. `stella fleet --pipeline` has the same
shape per its worker driver.

By the time `WrapperDispatch::run` finishes (after `judge` decides the
verdict) and control returns to `run_wrapped`'s tail code, the last round's
channel is already closed. `registry.events()` returns `None`, and the send
in `run_wrapped` —

```rust
if let Some(events) = registry.events() {
    let _ = events.send(stella_protocol::AgentEvent::GateBoard { .. });
}
```

— silently does nothing. I confirmed this empirically: I wrote the same kind
of witness test #4224 uses for the goal door, pointed it at `stella run
--pipeline` instead, and the store held zero `GateBoard`/`Verdict` events
across all three execution rows the run opened (the round's own "run" row,
a second held-open round's "run" row, and the `PluginChildStream`'s "plugin"
row).

`stella goal --pipeline` does not have this problem: its round driver
(`crates/stella-cli/src/agent/goal/goal_wrapped.rs`) keeps one execution row
and one event channel open across the whole loop, closing only once, after
every round-boundary send has already happened.

## What "fixed" looks like

`AgentEvent::Verdict` (and ideally `GateBoard`, same defect) reaching the
store for `stella run --pipeline`, attributed to the **same execution row**
the round's own `AgentEvent::FileChange` events landed on — that constraint
matters because `crate::dataset_cmd`'s `reward_label` folds one execution's
journal in isolation (`Store::execution_events`, not `Store::session_events`),
so a Verdict recorded on a different row (e.g. the `PluginChildStream`'s
"plugin" row, which does share the run's `session_id`) would still never be
seen by the fold that also needs the mutating file-change event.

Two directions worth evaluating, neither obviously right:

1. **Thread the execution id back through the wire.** `TurnDriver::run_turn`
   returns `DrivenTurn`
   (`crates/stella-runtime/src/wrapper/dispatch.rs`) — a documented
   wire-stable shape multiple hosts implement (`stella-cli`, `stella-serve`,
   an embedded `stella-engine` host). Widening it to report "which execution
   row (if any) this round used" lets `run_wrapped` reopen that row via
   `Store::record_event` after the dispatch decides the verdict. Costs:
   touches a stable contract; `stella-serve` may have no local store/execution
   id concept at all, so the field may need to be host-optional.
2. **Fold across the session instead of one execution.** Give the
   `PluginChildStream`/plugin-side execution row (or a new run-scoped row) the
   Verdict/GateBoard events, and change `dataset_cmd.rs`'s extraction to
   read a wrapped run's whole session (`Store::session_events`) rather than
   one execution's `Store::execution_events`, so a Verdict on one row and a
   FileChange on another still fold together. Costs: changes
   `is_accepted`'s and `fold_journal`'s current per-execution model, which
   plain one-shot `stella run` (no `--pipeline`) still relies on staying
   per-execution (`session_id` is `NULL` there).

## Where to look

- `crates/stella-cli/src/wrapper_plugin.rs` — `run_wrapped`'s tail code (the
  dead `GateBoard` send), `RawTurnDriver::run_turn`
- `crates/stella-cli/src/agent/turn.rs` — `run_turn`, `close_event_stream`
  call site
- `crates/stella-cli/src/wrapper_plugin/child_stream.rs` — module doc
  explains the three doors' different execution-row shapes; already states
  goal/fleet "already run every round under **one** run-scoped row", which is
  true for goal and apparently not actually exercised for fleet either (its
  worker driver was not checked directly — verify before assuming it's fine)
- `crates/stella-cli/src/dataset_cmd.rs` — `reward_label`, `fold_journal`,
  `is_accepted`
- `crates/stella-cli/tests/run_require_verdict_cli.rs` —
  `goal_wrapped_records_the_verdict_event_the_dataset_export_reads` is the
  pattern a `run`/`fleet` witness would follow once this lands

## Definition of done

- [x] A decision, written down (an ADR per SCR-002 if it changes an
      architectural contract), on which of the two directions above — or a
      third — is right.
      **ADR 0035** (`docs/adr/0035-a-wrappers-verdict-lands-on-the-round-that-earned-it.md`,
      indexed in `docs/adr/README.md` and `docs/manifest.json`). It takes a
      third direction and says why neither of the two lands: widening
      `DrivenTurn` would put a host-local bookkeeping fact into the wrapper
      socket's wire contract, and folding across the session would change the
      one-execution model a plain `stella run` still relies on. Instead the
      row travels back to the host through the `TurnDoor` the caller already
      threads in (`crates/stella-cli/src/turn_row.rs`).
- [x] `stella run --pipeline` publishes `AgentEvent::Verdict` (and ideally
      `GateBoard`) on the same execution row its own mutating file changes
      land on, with a witness test mirroring #4224's goal-door one.
      Both events, appended by `run_wrapped` to the last round's own row via
      the new `Store::append_event`
      (`crates/stella-store/src/event_write.rs`). Witness:
      `run_wrapped_records_the_verdict_on_the_rounds_own_execution_row` in
      `crates/stella-cli/tests/run_require_verdict_cli.rs`, which asserts the
      **row** as well as the count — the row carrying the verdict must also
      carry that turn's own terminator, so a verdict on the plugin-side row
      would fail it. It fails on the old code with 0 verdicts across every
      row (stash the `src` trees, keep the tests, run
      `cargo test -p stella-cli --test run_require_verdict_cli`) and passes
      here.
- [x] `stella fleet --pipeline` verified either way — confirm empirically
      whether its worker driver has the same gap before assuming
      `child_stream.rs`'s doc comment is accurate for it.
      Verified: the doc comment is accurate about the **row** and the
      **channel** — `crate::fleet_cmd`'s `run_task` opens one execution per
      attempt and holds the channel open across the whole dispatch — so the
      gap there is a different one. It never sent the two events at all.
      `AttemptWrapper::settle` now sends them on the attempt's own channel,
      from the same `wrapper_plugin::run_events` the other doors use.
      Witness: `a_fleet_attempt_records_the_verdict_on_its_own_execution_row`
      in the same file, likewise 0 verdicts before and 1 after.
- [x] If the fix touches `TurnDriver`/`DrivenTurn`, `stella-serve`'s use of
      the same trait is checked for compatibility.
      It does not touch either. Both types in
      `crates/stella-runtime/src/wrapper/dispatch.rs` are unchanged, which is
      the point of the third direction. `rg 'TurnDriver' crates/stella-serve/src`
      finds nothing; `stella-serve`'s own `DrivenTurn`
      (`crates/stella-serve/src/session.rs`) is an unrelated type that never
      implemented this trait.

Refs #4224

macanderson:
## What & why

`stella run --pipeline <variant>` and `stella fleet --pipeline <variant>`
recorded nothing about what the wrapper decided. No `AgentEvent::Verdict`, no
`AgentEvent::GateBoard`, on any execution row.

Two different causes:

- **`run`.** Each round is a `crate::agent::run_turn`. That call opens an
  execution row of its own and closes the channel that feeds it before it
  returns. The verdict arrives after the last round has done both, so the send
  `run_wrapped` made through `registry.events()` found an empty slot and
  dropped both events silently. The comment beside it said the opposite.
- **`fleet`.** Its attempt channel and its execution row both outlive the
  dispatch, so nothing was dropped. The two events were simply never sent.

That is a real loss rather than a cosmetic one: `crate::dataset_cmd` folds
**one** execution's journal (`Store::execution_events`), and a training label
needs the verdict and that turn's own mutating `FileChange` events in the same
fold. A verdict on any other row — the plugin-side row shares only the session
id — is one the fold can never see.

### The decision (`doc:adr/0035-a-wrappers-verdict-lands-on-the-round-that-earned-it`)

The issue offered two directions and neither is taken.

- **Widening `DrivenTurn`** with the round's execution id would put a
  host-local bookkeeping fact into the wrapper socket's wire contract, which
  exists for hosts with no local store at all.
- **Folding across the session** would change the one-execution model a plain
  `stella run` still relies on, to fix a problem only the wrapped path has.

A third: the row travels back to the host that asked for the turn, and the wire
is untouched. `crate::turn_row::TurnRow` is a shared handle; `run_turn` fills it
through the `TurnDoor` the caller already threads in (`recording_to`).
`RawTurnDriver` carries one across the dispatch, so each round replaces the last
and what survives is the final round's row — the round `DispatchReport::verdict`
is about. `run_wrapped` appends the two events there.

`Store::append_event` is the write. The renderer that owned the sequence is
gone, so it reads `max(seq) + 1` inside the transaction that writes the row;
two appenders block on the same connection lock and cannot pick one number. It
lives in a new `crates/stella-store/src/event_write.rs`, together with
`record_event`, which **moved there out of `stella-store`'s `lib.rs`** — a
grandfathered god file, so the shared helper lands outside it and that file
shrinks by 49 lines.

The fleet door needs none of that: `AttemptWrapper::settle` sends the same two
events on the attempt's open channel, built by the same
`wrapper_plugin::run_events` the other doors use, so no door invents its own
spelling.

`TurnDriver` and `DrivenTurn` are unchanged, so `stella-serve` has nothing to
absorb — `rg 'TurnDriver' crates/stella-serve/src` finds nothing, and its own
`DrivenTurn` is an unrelated type in `session.rs` that never implemented this
trait. That is the issue's fourth DoD item, answered by the shape of the fix.

Closes #6001

## The witness

Two, in `crates/stella-cli/tests/run_require_verdict_cli.rs`, both driving a
real `stella` subprocess against the existing arbiter fixture:

- `run_wrapped_records_the_verdict_on_the_rounds_own_execution_row` — asserts
  one verdict and one board in the store, **and** that the row carrying them
  also carries that turn's own terminator. The row matters, not just the
  count: a verdict on the plugin row would still be invisible to the
  one-execution fold.
- `a_fleet_attempt_records_the_verdict_on_its_own_execution_row`.

Proven the artisanal way — `git stash push -- crates/stella-cli/src
crates/stella-store/src` with the tests kept, then
`cargo test -p stella-cli --test run_require_verdict_cli records_the_verdict`:

```
test run_wrapped_records_the_verdict_on_the_rounds_own_execution_row ... FAILED
test a_fleet_attempt_records_the_verdict_on_its_own_execution_row ... FAILED
  left: 0
 right: 1
```

Both pass after `git stash pop`; the whole file is 9 passed, 0 failed. Three
unit witnesses cover `Store::append_event`'s sequencing and three cover
`TurnRow`'s last-row rule.

## The gate

CI is the evidence (CLAUDE.md, "CI builds and tests; this laptop does not").
Locally, scoped only:

- [x] `cargo fmt --check`
- [x] `cargo clippy -p stella-cli -p stella-store -p stella-protocol --all-targets -- -D warnings`
- [x] `cargo test -p stella-store` (586 passed), `cargo test -p stella-cli
      --test run_require_verdict_cli` (9), `--test goal_wrapped_dispatch_cli`
      (1), plus the `turn_row` / `wrapper` / `fleet` / `dataset` /
      `persistence` unit filters
- [x] `make guards-fast` green, `make doc-warnings CARGO_SCOPE="-p stella-cli
      -p stella-store"` clean
- [x] Docs updated: ADR 0035, `docs/adr/README.md`, `docs/manifest.json`,
      `crates/stella-store/README.md`'s module table
- [x] CLA signed
- [x] `Closes #6001` appears both above and as a commit trailer

## Fix over file

- [x] Extra fixes in this PR:
  1. `crates/stella-protocol/src/event/tags.rs` — the `Verdict` row's comment
     claimed the `run` and `fleet` doors "cannot do the same yet: each round's
     execution row closes before a verdict exists to send." This change makes
     that false, so it now says what each door does.
  2. `crates/stella-cli/src/wrapper_plugin.rs` — the comment beside the dead
     send claimed it worked ("Sent through the registry's sender because this
     door assembles no channel of its own"). Replaced with what the code does
     and why the channel is gone.

## Ground-rule check

- [x] No new cross-boundary types — `AgentEvent::Verdict` and
      `AgentEvent::GateBoard` already exist and already round-trip

- **The events reach the store, not a live screen.** On the `run` door there is
  no live screen at that moment: the round's renderer has finished. What a
  person watching sees is unchanged — `report_to` still prints the same lines.
- **A held-open wrapper writes one verdict, on the last round.** An earlier
  round's row would date the decision to a turn that was later revised. For a
  multi-round run whose file changes landed on round 1, the verdict and those
  changes are still on different rows; that is the `run` door's pre-existing
  one-row-per-round shape, not something this change introduces, and the
  single-round case the fixture exercises is the one the export reads today.
- **Appending after `finish_execution` is deliberate.** Nothing in the schema
  closes a journal and every reader orders by sequence, so the event reads as
  the last thing that happened to that execution — which it is.
- **`record_event` moved files.** Its behaviour and signature are unchanged; the
  body now delegates to a shared `write_event` that `append_event` also uses, so
  the payload encoding, the type tag, the `task_id` lift and the `tool_calls`
  projection have one statement rather than two that can drift.

## Summary by Sourcery

Persist wrapper verdicts and gate boards on the execution rows associated with the turns they judge.

Bug Fixes:
- Record wrapper verdict and gate-board events for both `stella run --pipeline` and `stella fleet --pipeline` executions.
- Ensure `stella run` appends wrapper decisions to the final round's execution journal so dataset folds can associate them with the turn's events.

Enhancements:
- Add shared execution-row tracking and post-turn event appending without changing the wrapper wire contract.
- Centralize event writing and sequencing in the store while preserving atomic event and tool-call projection updates.

Documentation:
- Document the wrapper verdict persistence decision in ADR 0035 and update the ADR and store documentation indexes.

Tests:
- Add CLI integration coverage proving verdict and gate-board events reach the correct execution rows for wrapped runs and fleet attempts.
- Add unit coverage for execution-row selection and appended event sequencing.

Chores:
- Update protocol and source comments to reflect verdict persistence for all wrapper execution doors.


macanderson:
Merged `main` in, and renumbered this record **0033 → 0035**. Both pushed.

**The conflict** was `docs/adr/README.md` and `docs/manifest.json`, each time one side's new ADR row against the other's. `main` took 0032 (a loop rung reads structure), this branch had 0033, so the numbers did not collide and the resolution is the union of both rows and both paragraphs ordered by number. Nothing else in either diff overlapped.

**The renumber is the part worth reading.** Three open pull requests had each taken "the next free number" against their own base, and two landed on the same one:

| PR | opened | number held |
|---|---|---|
| #6385 | 08:19 | 0033 |
| #6386 | 08:24 | 0034 |
| this | 08:52 | 0033 ← |

Two records whose numbers sort apart share no line, so git reports the merge clean. `adr-numbering` *does* ask the composed question — a pull request's checkout is `refs/pull/N/merge` — but it asks once, and nothing re-runs a green check when `main` moves under it. So the second of #6385 and this one to merge would carry a verdict taken while 0033 was still free onto a `main` where 0033 identifies two records. That is a `main-red`, and `main-red-hold` makes a `main-red` hold every open pull request in the repository until someone renumbers by hand.

I moved this one rather than #6385 so the numbers follow the order the records were opened in — 0033 #6385, 0034 #6386, 0035 here — which is disjoint in **any** merge order.

Moved with the record: its `id:`, title, heading, the `docs/manifest.json` key/path/title, and its row and paragraph in `docs/adr/README.md`. `git grep 0033` finds no other citation of it; the remaining hits are a `Cargo.lock` checksum, a font binary, and fixtures naming frames `f_0033.png`. **The PR body above still says ADR 0033 in "What & why" — that is now stale and wants a one-line edit.**

Evidence: `check-adr-numbering.py` → 33 records, every number unique, every heading and id in agreement, every record indexed. `check-doc-links.py` → exit 0. `docs/manifest.json` parses.

#6392 is the guard for this class and is a different thing — it adds the post-merge canary row and makes the failure name the free number. It does not renumber the collision standing today, which is why this one still needed moving.

---

macanderson:
## Three open PRs are writing the same two shared cells, and each was cut before seeing the others

#6385, #6386 and this one each add an ADR **and** append a row to `docs/adr/README.md` plus an entry to `docs/manifest.json`. That is the shared-cell shape AGENTS.md documents for `Cargo.lock` and `scripts/file-size-baseline.txt`: every branch is correct on its own, and the composition is what breaks.

**Verified first-hand on this PR's diff.** `docs/adr/README.md` adds exactly one row, immediately after `0032`:

```diff
 | [0032](0032-a-loop-rung-reads-structure.md) | A Loop Rung Reads Structure, Not Progress | Accepted |
+| [0035](0035-a-wrappers-verdict-lands-on-the-round-that-earned-it.md) | A Wrapper's Verdict Lands on the Round That Earned It | Accepted |
```

No `0033`, no `0034`. `docs/manifest.json` inserts only the `0035` key, in the same position. So this branch's base predates both siblings.

**Reported from a triage pass over all three, not re-derived here:** #6385 adds `0033-free-text-is-not-evidence`, #6386 adds `0034-silence-is-not-a-grant`, and #6386's README diff likewise carries no `0033` row. Worth a glance before merging rather than taking my word for it.

**The numbers do not collide** — 0033/0034/0035 are distinct, so `adr-numbering` passes on each branch. What collides is the two shared cells:

- Merge one, and the next two conflict textually on the same README region and the same `manifest.json` position.
- Resolve those conflicts carelessly and the table lands with gaps — `0032` then `0035`, with `0033`/`0034` files on disk and no rows pointing at them.

Merging in numeric order (#6385 → #6386 → #6390) makes each conflict trivially resolvable, since each one's row belongs strictly after the previous.

**This is a live instance of what #6392 is about.** That PR adds an `adr-numbering` row to the post-merge `main-canary.sh` precisely because the ADR number is a shared cell whose pre-merge verdict goes stale once `main` moves — which is exactly what will happen here on merge #2. #6395 attacks the same failure mode on `Cargo.lock` with a re-run sweep instead. Both edit AGENTS.md's shared-cell discussion, so they will also conflict with each other; worth deciding which is the house pattern before both land.

No CI is red on any of the ten open PRs here, none is `dirty`, and none carries `HOLD`. Nothing was pushed, merged or modified — this is a heads-up about merge order, not a request for a change.


---

macanderson:
Superseded by #6396, which carries this branch's tree in one commit.

`rebase-replay` fails here on
`docs/adr/0033-a-wrappers-verdict-lands-on-the-round-that-earned-it.md`: it is
created by 1b1f5a33e and deleted by 505fd8b0a, the peer commit that renumbered
the record to 0035 after another open PR took 0033. That create-then-undo pair
is the shape AGENTS.md § the ninth workflow describes, and the remedy it names
is to flatten the branch so the path is never mentioned.

#6396 is that flattened history: `git diff 505fd8b0a <its head>` prints
nothing, and the ADR is created at 0035 directly. Flattening in place needs a
force push, which this session does not do, so it went to a new branch.
Closing this one.

PR #6393, opened 2026-09-07
fix(scripts): a claim check with no gh says it did not run

### LINKED ISSUE #5045: Graph edges cannot say which turn touched them

### LINKED ISSUE #5934: issue-claim.sh and main-red-claim.sh cannot run in the agent container: gh is absent, so the collision guards are silently unavailable
## Problem

`scripts/issue-claim.sh` and `scripts/main-red-claim.sh` are this repository's two anti-collision mechanisms. Both are built on `gh`. **`gh` is not installed in the container agent sessions run in**, so neither can execute there. Both fail open by design — which is the correct direction, and is not what this issue disputes — but the practical result is that the signal is *absent* rather than *consulted*, in exactly the environment where the collisions happen.

## It happened today, twice over

Two sessions independently fixed the same guard defect within four minutes:

| PR | Opened | Session |
| --- | --- | --- |
| #5931 | 09:46:14 | `session_017H844j1i6G7vpPyHRv3Cgh` |
| #5932 | 09:50 | `session_01FCzBgbB8RU7FBcTJgoh5ZD` |

Both closed #5910, both reached the same discriminator (*who* removed the path), both wrote an R9 case with the same name. #5932 was closed as the duplicate. Neither session could run `./scripts/issue-claim.sh check 5910` before starting, because `gh` was not there to run it.

This is the failure #5224 filed the script for, and the earlier instances it cites (#5045, #4336, #5054 — three in one afternoon). The mechanism exists and is correct; it just is not reachable from the sessions that need it.

## Why this is not simply "install gh"

That may well be the answer, but the image is not defined in this repository — the same constraint #3830 records for `shellcheck`, which stops `make guards-fast` on its lowest rung for the same reason. Worth deciding together rather than twice.

There is a second, cheaper half that *is* in this repository's control: **the scripts do not say they did not run.** Compare `shellcheck`'s handling, which #3615 deliberately made loud:

```
shellcheck: UNAVAILABLE — THIS STEP DID NOT RUN
```

`issue-claim.sh` has the right posture in its wording for its other unknowns (`ok  proceed (could not ask)`), and #5838 has just finished tightening the one case that asserted an answer it never got. A missing `gh` should read the same way — and, more importantly, an agent reading CLAUDE.md's instruction to "run `./scripts/issue-claim.sh check` before writing code" gets a `command not found` and no guidance about what that means for the rule it was told to follow.

## Files

- `scripts/issue-claim.sh`, `scripts/main-red-claim.sh` — the `gh` preflight and its message.
- `CLAUDE.md` — the two bullets instructing a session to run them; neither says what to do when the tool is absent.
- `AGENTS.md` § the canary / the claim mechanic — the normative description.

## Verify

1. In an environment with no `gh` on `PATH`, run `./scripts/issue-claim.sh check 5910`.
2. It should report that it could not ask, in the `UNAVAILABLE` register, and say what the session should do instead (proceed, and how to check by hand) — rather than failing on `command not found` or printing a line indistinguishable from a clean check.
3. `make issue-claim-test` / `make main-red-claim-test` gain a case with `gh` masked off `PATH`.

## Definition of done

- [x] Both scripts detect an absent `gh` and say so explicitly, without changing the fail-open behaviour. — PR #6393: both open with a tool pre-flight (`scripts/lib/claim-tools.sh`'s `missing_claim_tools` and `report_claim_tools_unavailable`) printing `gh: UNAVAILABLE — THIS CHECK DID NOT RUN`, what went unasked, and the by-hand questions. Both still exit 0; the stdout verdict reads `ok  proceed (UNAVAILABLE: gh — nothing was asked)`.
- [x] Their self-tests cover the absent-`gh` branch. — `make issue-claim-test` and `make main-red-claim-test` each gained `a missing gh proceeds, and says the check did not run` and `a missing jq names jq, rather than blaming the comments`, driven with the tool masked off `PATH`. Both assert the banner rather than the exit code alone. Stashing the scripts takes each suite from 40 passed to 38 passed / 2 failed.
- [x] `CLAUDE.md`'s two instructions say what a session should do when the check cannot run, so the rule is followable rather than silently void. — a new bullet beside the red-`main` one: read the banner as no answer, ask by hand (the closing pull requests, the issue state, the claim comments), then leave a claim comment, which is the half such a session can still do.
- [x] A decision is recorded on whether the agent image should carry `gh` (and `shellcheck`, #3830), or a note that it cannot and the rule is best-effort in those sessions. — recorded in AGENTS.md § the issue-claim mechanic: the container is not defined in this repository, so nothing here can install the tools, the want is the same one #3830 tracks for `shellcheck`, and the rule is best-effort in such a session — the by-hand check plus the claim comment.

### Added by the second-collision comment

- [x] Decide whether the claim mechanic extends to a pull request as a unit of work. — it does: `scripts/pr-claim.sh` landed from the sibling issue while PR #6393 was open, with `check`, `claim` and `post` on the same lapsing-comment, session-word, fail-open rules. PR #6393 gave it the shared tool pre-flight too, so a sweep with no `gh` reads the same banner wherever it asks, and its `post` exits 3 saying `NOTHING WAS POSTED` rather than reporting a finding it never sent.
- [x] If it does not, say so where a sweeping agent reads it. — moot, since it does; AGENTS.md § the claim mechanic carries the script, its three modes, and the pre-flight paragraph naming all three claim scripts.

macanderson:
## What & why

`scripts/issue-claim.sh`, `scripts/main-red-claim.sh` and `scripts/pr-claim.sh` are the anti-collision pre-flights every session is told to run. Each needs `gh` to reach the tracker and `jq` to read the claims out of the reply. The agent container ships neither, so the run printed `ok  proceed (could not ask)` — a line a reader takes for a clean check, in exactly the environment where the collisions happen. The issue records what that cost: two sessions implementing one issue twice, four repairing one red `main`, two resolving one merge conflict, each pair paying for a full required-CI run as well as the work.

Each now opens with a tool pre-flight, in the `shellcheck: UNAVAILABLE` register the gate already uses:

```
gh: UNAVAILABLE — THIS CHECK DID NOT RUN

     Nothing asked whether #5045 is already being worked.
     That is no answer at all, and it is not an answer of
     'nobody is on it'.

     Going ahead anyway: a claim check that can block work is worse
     than the duplication it stops.

     Ask by hand, from a machine that has these tools or in a browser:
       - open and merged pull requests whose body says `Closes #5045`
       - the state of #5045, and its claim comments: they open with
         `<!-- issue-claim -->`, and one holds while it is under 90m old
     Then say on #5045 that you are taking it, so the next session reads
     a claim rather than a collision.
ok  proceed (UNAVAILABLE: gh — nothing was asked)
```

Fail-open is unchanged: each still exits 0, and the stdout verdict now says which tool was missing instead of reading like a clean answer. `scripts/lib/claim-tools.sh` holds the tool list and the banner, so one message serves all three scripts rather than three copies drifting.

`scripts/pr-claim.sh` landed on `main` while this PR was open, carrying the same shape — a `gh`-only pre-flight, the old wording, and the same `dirname` dependency. It takes the shared pre-flight here, on the same terms. Its `post` mode keeps its own answer: that mode owes the caller a write, so a missing tool exits 3 and says `NOTHING WAS POSTED` rather than reporting a finding it never sent.

The decision the issue asks for is recorded in AGENTS.md § the issue-claim mechanic: the container is not defined in this repository, so nothing here can install the tools — the same constraint `shellcheck` carries one section up — and in such a session the rule is best-effort. The by-hand check is the banner's own list, and the claim comment is the half a session without `gh` can still leave for the next one. CLAUDE.md carries the same instruction where a session reads its rules, since a session told to run a pre-flight that cannot run needs to know what the rule now asks of it.

Closes #5934

## The witness

Seven cases across three suites, one per tool, run with that tool masked off `PATH`:

- `issue-claim`: `a missing gh proceeds, and says the check did not run`, `a missing jq names jq, rather than blaming the comments`
- `main-red-claim`: the same pair
- `pr-claim`: `a check with no gh proceeds, and says the check did not run`, `a post with no gh fails and says nothing was posted`, `a missing jq names jq, rather than blaming the comments`

Each asserts the banner, not only the exit code: proceeding is what these scripts do at every unknown, so an exit of 0 alone cannot tell a check that ran from a check that never happened. The `jq` case's stub `gh` fails loudly if it is ever called, so it also witnesses that the pre-flight runs before anything is asked.

Proved on the two suites that existed when the fix was written, by stashing the scripts and re-running them:

```
$ git stash push -- scripts/issue-claim.sh scripts/main-red-claim.sh scripts/lib/claim-tools.sh
$ ./scripts/test-issue-claim.sh    → issue-claim: 38 passed, 2 FAILED
$ ./scripts/test-main-red-claim.sh → 38 passed, 2 failed
$ git stash pop
$ ./scripts/test-issue-claim.sh    → issue-claim: 40 passed
$ ./scripts/test-main-red-claim.sh → 40 passed, 0 failed
```

`main-red-claim-test` already had a `gh`-less case and `pr-claim-test` had two; each asserted only the old wording. They now assert the banner, so they can tell a check that ran from one that never did. `pr-claim-test`'s pair is folded into three cases rather than two blocks that each cover `gh` — named here because the suite is shell, not Rust, so `check-deleted-tests` does not see it.

## The gate

- [x] `cargo fmt --check` (via `make guards-fast`, exit 0)
- [ ] `cargo clippy --workspace --all-targets -- -D warnings` — no Rust in this diff; CI runs it
- [ ] `cargo test --workspace` — no Rust in this diff; CI runs it. `make issue-claim-test` (40/40), `make main-red-claim-test` (40/40) and `make pr-claim-test` (51/51) green, `make shellcheck` green, `make guards-fast` exit 0, `python3 scripts/check-prose.py` green, `./scripts/test-dev-env.sh` green (38/0)
- [x] Docs updated where behavior/flags changed (AGENTS.md, CLAUDE.md, `setup-dev-env.sh`'s tool table)
- [x] CLA signed

## Fix over file

- [x] Extra fixes in this PR: three, all in the same pre-flight path.
  1. **A missing `jq` blamed the tracker.** `gh --jq` carries its own copy of jq, so the tracker calls need `gh` alone; a real `jq` is used only for the claim-comment filter. With `gh` present and `jq` absent, both scripts reported `could not read #N's comments` — the wrong cause, naming the tracker for a missing tool. The pre-flight names `jq` before anything is asked.
  2. **The shared library was sourced through `$(dirname "$0")`.** A thin `PATH` is the environment this pre-flight exists for, and on one `resolve_session` vanished silently, so every claim of the session's own login read as a stranger's. The directory now comes from `${0%/*}`, which needs nothing on `PATH`.
  3. **`pr-claim.sh` had both defects.** It landed mid-PR and is fixed here rather than left for a follow-up, so all three claim scripts say one thing.
- [x] Filed, with the reason a fix could not ride this PR: nothing was deferred. Installing `gh` in the agent image is not a fix this repository can make — the image is defined elsewhere — and that is recorded in AGENTS.md rather than filed as a second issue, since `#3830` already tracks the same want for `shellcheck`.

## Ground-rule check

The old-code run of the new cases fails with `dirname: command not found` in its output as well as the missing banner. That is fix (2) above showing through: the fixture `PATH` holds every program the script uses, and the old script also needed `dirname`. The case still fails on the old code and passes on the new either way.

No Rust test was deleted or renamed. In `scripts/test-pr-claim.sh` two shell cases covering a missing `gh` are replaced by three that assert the banner and add `jq`.

Two sessions worked this branch at once — the collision the issue is about — and both reached the same `pr-claim.sh` fix three minutes apart. `a43c6b4f6` keeps one copy of each half and the PR comment says which; `./scripts/pr-claim.sh claim 6393` is posted so a third sweep stands down.

### DIFF
--- a/.github/workflows/guard-self-tests.yml
+++ b/.github/workflows/guard-self-tests.yml
@@ -296,6 +296,10 @@ jobs:
         if: ${{ !cancelled() }}
         run: ./scripts/test-issue-claim.sh
 
+      - name: the pull-request claim and the duplicate-finding gate
+        if: ${{ !cancelled() }}
+        run: ./scripts/test-pr-claim.sh
+
       - name: the CodeQL canary opens, refreshes and closes a codeql-red issue
         if: ${{ !cancelled() }}
         run: ./scripts/test-codeql-canary.sh
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -462,42 +462,66 @@ It carries the same session word, and for the same reason: one login is one
 account, and the thing that has to be identified is the session. Comparing
 the login alone read a peer session's claim as this session's own and cleared
 both to work one issue (`#5875`). `scripts/lib/claim-session.sh` resolves the
-word for both scripts, so a fleet sets `STELLA_CLAIM_SESSION` once rather
-than twice, and both fail open the same way: a run with no word of its own,
-and a claim comment carrying none, fall back to the author-only rule.
+word for every claim script here, so a fleet sets `STELLA_CLAIM_SESSION` once
+rather than once per script, and they all fail open the same way: a run with
+no word of its own, and a claim comment carrying none, fall back to the
+author-only rule.
 
 Run it **before writing code and again before opening the PR**. The gap
 between those two is enough: a peer's PR can merge inside one issue's worth of
 work, and then the check that was clean at the start is stale at the end.
 
-**Neither claim check can run in the agent container, and both now say so.**
+**A pull request is the third thing two sessions take at once, and
+`scripts/pr-claim.sh` is that mechanic pointed at one.** Three sweep sessions
+read one pull request in eight minutes, each worked out the same merge
+conflict, and each posted it; a fourth stopped, because it read the comments
+first. Duplication here does not lose work, it publishes it, and the
+maintainer is left to diff three long comments to learn they say one thing
+(`#5861`).
+
+```bash
+./scripts/pr-claim.sh check 5835    # exit 0 proceed, 1 stand down
+./scripts/pr-claim.sh claim 5835    # check, then post the claim
+./scripts/pr-claim.sh post 5835 --finding conflict:5828 --body-file note.md
+```
+
+The claim half takes the rules above: a lapsing comment, a session word beside
+the login, and every unknown proceeding loudly. A merged or closed pull request
+stands a sweep down too, since neither one takes a comment.
+
+`post` is the half that would have stopped all three. It asks the claim
+question too, then asks whether the same finding already stands, and writes
+only when neither answer is yes — the finding question alone can only see what
+somebody already published, so it turns a second sweep away after it has spent
+its diagnosis rather than before. `--ignore-claim` posts over a live claim, for
+the operator who read it and said on the pull request why. The caller names the
+finding, because two sessions that find one thing write it up two ways, and a
+digest of the words would call them different. A key that has to go stale
+carries what it depends on: a finding about the head commit puts the head sha
+in the key. A read that failed says so and posts anyway, rather than reporting
+that nothing stands.
+
+Take the claim when you start to read the pull request, not when you are ready
+to write. Each of those three sessions spent twenty minutes on the conflict
+before it had anything to say, so a claim taken at the end would have saved
+none of it. `make pr-claim N=5835` asks by hand; `make pr-claim-test` covers
+both gates, the blocking branches included.
+
+**No claim check can run in the agent container, and all three now say so.**
 They need `gh` to reach the tracker and `jq` to read the claims out of the
 reply, and that container ships neither. So the check every session is told
 to run printed `ok  proceed (could not ask)` — a line a reader takes for a
 clean check — and two sessions then implemented one issue twice, four
 repaired one red `main`, and two resolved one merge conflict, each pair
-paying for a full required-CI run as well as the work (`#5934`). Both scripts
-open with a tool pre-flight now, in the `shellcheck: UNAVAILABLE` register
+paying for a full required-CI run as well as the work (`#5934`). Each script
+opens with a tool pre-flight now, in the `shellcheck: UNAVAILABLE` register
 above and for the same reason: a check that did not run must not read as a
 check that found nothing. The banner names the missing tool, says what went
-unasked, and lists the by-hand questions — the open pull requests, the
-issue's state, the claim comments. `scripts/lib/claim-tools.sh` holds it, so
-one message serves both scripts, and each suite drives it with the tool
-masked off `PATH`.
-
-**A pull request is not claimed the same way, and does not need to be.** A
-sweep hits a conflicted or red pull request and every session reaches the same
-conclusion about it at once, which is the shape both scripts above exist for.
-There is no `pr-claim.sh`, because a pull request already carries both halves
-a claim script would add: a comment thread every session can read, and a head
-that moves the moment somebody pushes. The race there also fails closed — a
-second resolution is refused non-fast-forward, so the work is duplicated and
-never dropped, which is the opposite of the issue case, where one merge kept
-one tree and silently lost the other. So a pull request takes the same two
-moves without a script: read its recent pushes and its comments before you
-start, and say on it that you are resolving it. Two sessions each paid for a
-full conflict resolution of one pull request — a rename, a renumber, four
-files and a guard run each — before that was written anywhere.
+unasked, and lists the by-hand questions. `scripts/lib/claim-tools.sh` holds
+it, so one message serves all three, and each suite drives it with the tool
+masked off `PATH`. `pr-claim.sh`'s `post` keeps its own answer there and
+exits 3: that mode owes the caller a write, and a proceed would tell a sweep
+its finding is up when nothing was sent.
 
 **The image is out of this repository's reach, so the rule is best-effort
 there.** Nothing in this tree defines the container, exactly as with
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -470,6 +470,29 @@ Run it **before writing code and again before opening the PR**. The gap
 between those two is enough: a peer's PR can merge inside one issue's worth of
 work, and then the check that was clean at the start is stale at the end.
 
+**Neither claim check can run in the agent container, and both now say so.**
+They need `gh` to reach the tracker and `jq` to read the claims out of the
+reply, and that container ships neither. So the check every session is told
+to run printed `ok  proceed (could not ask)` — a line a reader takes for a
+clean check — and two sessions then implemented one issue twice, four
+repaired one red `main`, and two resolved one merge conflict, each pair
+paying for a full required-CI run as well as the work (`#5934`). Both scripts
+open with a tool pre-flight now, in the `shellcheck: UNAVAILABLE` register
+above and for the same reason: a check that did not run must not read as a
+check that found nothing. The banner names the missing tool, says what went
+unasked, and lists the by-hand questions — the open pull requests, the
+issue's state, the claim comments. `scripts/lib/claim-tools.sh` holds it, so
+one message serves both scripts, and each suite drives it with the tool
+masked off `PATH`.
+
+**The image is out of this repository's reach, so the rule is best-effort
+there.** Nothing in this tree defines the container, exactly as with
+`shellcheck` above (`#3830`) — the two wants are the same want and are
+tracked together. Until the image carries them, a session without `gh` owes
+the by-hand check and, more importantly, the claim comment: posting one is
+the half such a session can still do, and it is what lets the next session
+stand down rather than collide.
+
 An eighth, `windows-check.yml`, is the only compiler in this project that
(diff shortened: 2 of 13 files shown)


macanderson:
Resolved this branch's conflict with `main` and closed the one gap the resolution left open. Pushed as `93b82d4`.

**The conflict was real, not textual.** This branch wrote that a pull request is not claimed the same way and that there is no `pr-claim.sh`; #6377 landed one on `main` while the branch was open. Main's section stands; this branch's paragraphs on the claim tools now cover all three scripts rather than two.

**A concurrent session resolved it at `e3695b7` while I was working** — my push was refused non-fast-forward, which is the race failing closed exactly as AGENTS.md says it does. I merged their commit in rather than replacing it, so their `Makefile` target and the three `guard-self-tests.yml` steps that run these suites in CI are kept intact.

**One difference, and it is the reason I pushed rather than standing down.** Their `AGENTS.md` records the remaining gap:

> `scripts/pr-claim.sh` arrived beside them from `#5861` and does not share that pre-flight: it reports a missing `gh` in its own words and says nothing about `jq`. `#5934` carries that gap.

That is an accurate reading, and the gap is live: `pr-claim.sh` pre-flights `gh` alone, while its claim reader and its finding reader both run a real `jq` filter (the two `jq -r --arg marker` calls). That is the half-checked state `claim-tools.sh` was written for — with `gh` present and `jq` gone, the script blames the comments for a missing tool.

Since this PR closes #5934, shipping it while the issue still "carries that gap" would leave its own DoD unmet (SCR-003), and the fix is small enough to ride the PR (AGENTS.md § "Fix over file"). So `pr-claim.sh` now takes the shared pre-flight on the same terms as its two siblings. `post` keeps its own non-zero answer — a mode that owes the caller a write must not report a claim it never sent — and the `PATH` lookup moves off `dirname` for the reason the other two did: the suites strip `PATH` down to the tools under test.

**Evidence.** Three new cases in `scripts/test-pr-claim.sh` drive it with each tool masked off `PATH`: a missing `gh`, a missing `jq` whose stub `gh` fails loudly if the pre-flight lets anything through, and a `post` that must say nothing was posted. The pre-existing no-`gh` case asserted the old wording and now asserts the banner.

```
scripts/test-pr-claim.sh        53 passed, 0 failed
scripts/test-issue-claim.sh     40 passed
scripts/test-main-red-claim.sh  40 passed, 0 failed
make prose line-citations left-behind invariants doc-links
     gate-parity god-files guard-trigger-coverage — all clean
```

**What this does not establish.** `shellcheck` is not installed in this container, so that gate step did not run here — CI is the gate for it. I have not re-read the Sourcery assessment table for the new head; it should re-post within a few minutes of this push.

---

macanderson:
macanderson aa139adcb2011711
Sweeping this. The claim lapses after 90 minutes, so it cannot
hold the pull request shut if this session dies
(`scripts/pr-claim.sh`, `#5861`).
The second word is this session's own, so another session of the same author
reads this as somebody else's claim.

macanderson:
Two sessions worked this branch at once, which is the collision the issue is about. Both resolved the conflict with `main` after `pr-claim.sh` landed, and both then gave that script the shared tool pre-flight, three minutes apart.

Nothing was lost. `a43c6b4f6` keeps one copy of each half: AGENTS.md takes the peer resolution's two paragraphs, `scripts/pr-claim.sh` keeps this session's `UNAVAILABLE` banner with the peer's merged-or-closed question added to the by-hand list, and `scripts/test-pr-claim.sh` keeps the three folded cases rather than two blocks that each cover `gh`.

`./scripts/pr-claim.sh claim 6393` is posted above, so a third sweep stands down.

Issue #1292, opened 2026-08-03
Use affected-test selection as real verification evidence


macanderson:
## What's going on

Running a whole suite to prove one change is slow. Running only the tests a
change touches is cheap. This issue asked for that saving, and for the result
to count as verification evidence.

Both premises it was filed on are false today, and the comment thread below
established that against `main`. The selector is gone: `run_tests` is a retired
name in `crates/stella-tool-facts/src/catalog.rs`, and `#3236` deleted
`CodeGraph::importer_paths` when the tool purge took its only caller — the
reason is written into `crates/stella-graph/src/graph.rs`'s
`the_importer_relation_lists_files_whose_imports_resolve_here`. Host-run
verification is gone too (`#3865`), so what is left is a question about a
verification plugin's `[oracle]`, not about host code.

## What was settled

The architecture question — **who picks the tests** — is decided in
[ADR 0040](docs/adr/0040-host-does-not-pick-the-tests.md), under SCR-002.

**The host does not pick, and gains no capability that does.** Picking which
tests decide "done" is authoring the proof the host then grades.
`crates/stella-cli/src/wrapper_candidate.rs` already refuses the smaller
version of the same act — deriving a witness file from a runner's convention —
and `BeforeTurnResponse::witness` exists because of that refusal.

**A plugin may narrow inside its own oracle.** On the shared-tree path the
grant carries a real root, so a plugin can work out any command it likes and
run it in its own process. What it reports is
`EvidenceProvenance::PluginReported`, which is the grade a self-picked suite
earns.

**A narrowed run cannot borrow the wide run's red half.** The baseline belongs
to the granted `program` and `args`. A flip reported about some other command
is a claim whose red nobody watched, and the provenance says whose claim it is.

## Definition of done

- [x] There is a written design covering when the baseline is taken and every
      case where it is unavailable — `crates/stella-cli/src/wrapper_candidate.rs`'s
      module doc, § "When the baseline is observed, and every case where it
      cannot be", shipped in `#6422`.
- [x] The "we could not establish a baseline" case produces an honest unproven
      result rather than a pass — a deadline or a program that will not start is
      `TestBaseline::Unobserved`, never `Failed`; a turn with no test command
      stays `NotRun`. Neither reaches the oracle as red, so the flip stays
      `unobservable` and `judge` abstains.
- [x] The open architecture question is decided and recorded as an ADR —
      `docs/adr/0040-host-does-not-pick-the-tests.md`.
[...]

PR #6422, opened 2026-09-07
fix(stella-cli): observe the test baseline before the turn a plugin judges

### LINKED ISSUE #6421: No verification plugin can ever prove a fix: the host never runs the tests before a turn
## What's wrong

Every door mints a wrapper plugin's candidate grant through one function,
`grant_shared_tree` in `crates/stella-cli/src/wrapper_candidate.rs`. It built
the test plan with `stella_plugin::test_plan(parsed, None)`, and `None` means
`TestBaseline::NotRun`. That was the only call to `test_plan` in shipping code,
so **every grant Stella had ever handed a plugin said the tests were never run
before the turn.**

A flip is a claim about two moments: the tests failed, and then they passed.
`plugins/stella-witness/main.py`'s `assess` feeds its oracle a failing
observation only when the grant's baseline says `failed`. Handed `not-run` it
never locks a command, so it reports `unobservable`, and `judge` under
`flip = "required"` turns that into `Undecided`. **The one verification plugin
this repository ships could not credit a flip on any door** — not
`stella run --pipeline <plugin>`, not `stella goal`, not `stella fleet`.

The comment in `grant_shared_tree` gave the reason for the choice: "the host
does not spend a test run the plugin is about to spend itself. A wrapper
observes red in `before_turn`". A subprocess wrapper cannot. `exchange` in
`crates/stella-runtime/src/wrapper/subprocess.rs` spawns a fresh process for
every point, so whatever a plugin saw at `before_turn` is gone by `after_turn`.
The host is the only party that spans both moments.

Nothing caught it because the conformance vectors write the baseline by hand:
`plugins/stella-witness/testdata/01-flip-achieved.request.json` carries
`"baseline": "failed"`, a value no production path could produce. The suite
proved the plugin credits a flip given red, and no test asked whether red ever
arrives.

## How to see it

`crates/stella-cli/src/wrapper_candidate.rs`, in the test module:

```rust
let granted = grant_shared_tree(dir.path(), Some("sh tests/witness_flip.sh"))
    .expect("the grant mints");
assert_eq!(granted.grant.test.unwrap().baseline, TestBaseline::Failed);
```

with `tests/witness_flip.sh` exiting 1. It reported `NotRun`.

## Definition of done

Fixed in #6422. Each box names what proves it; every test below was run scoped
(`cargo test -p stella-cli`, exit 0), and CI's run on the PR is the evidence of
record.

- [x] `grant_shared_tree` runs the granted invocation once, in the granted
      root, before the turn, and stamps what it saw on `TestPlan::baseline`.
      — `crates/stella-cli/src/wrapper_candidate.rs`, `grant_shared_tree`, now
      `async`; witness
      `wrapper_candidate::tests::a_red_suite_before_the_turn_reaches_the_plugin_as_red`.
      Falsified rather than asserted: replacing `plan.baseline = observed;`
      with a discard makes it fail `left: NotRun, right: Failed`.
- [x] A suite that is red before the turn reaches the plugin as `failed`, and a
      green one as `passed` rather than being dropped. — the witness above, and
      `wrapper_candidate::tests::a_green_suite_before_the_turn_is_not_dressed_up_as_red`.
- [x] A run that could not be observed — killed at the deadline, or a program
      that will not start — is `unobserved` and never `failed`. —
      `wrapper_test_run::tests::a_baseline_that_could_not_be_run_is_refused_rather_than_scored_red`
      pins the refusal, `wrapper_candidate::observed_baseline` maps it to
      `Unobserved`, and the pre-existing
      `wrapper_test_run::tests::a_run_killed_at_the_deadline_observed_no_assertions`
      pins the deadline arm the two paths now share.
- [x] A turn that names no test command runs nothing and still reports
      `not-run`. —
      `wrapper_candidate::tests::a_turn_with_no_test_command_observes_nothing_and_says_nothing`.
- [x] One function decides pass and fail for both readings: the pre-turn
      observation and the `run_test` host call a plugin can make afterwards. —
      `wrapper_test_run::observe_baseline` and `GrantedTestRuns::run_test` both
      call `run_in`; witness
      `wrapper_test_run::tests::the_baseline_reads_the_invocation_the_same_way_a_re_run_does`.
- [x] The baseline is observed before the tamper watch is pinned, so the host's
      own run is never reported as a worker rewriting the witness. — the
      ordering in `grant_shared_tree`, stated in that module's
      "When the baseline is observed" section; the watch's own behaviour stays
      pinned by
      `wrapper_candidate::tests::the_host_vouches_for_an_untouched_witness_and_names_a_rewritten_one`.
- [x] An end-to-end test asserts the baseline on the wire, through a real door.
      —
      `fleet_wrapped_dispatch_cli::a_plan_files_per_task_test_command_reaches_the_plugin_as_a_test_plan`
      now asserts
      `"test":{"program":"sh","args":["tests/witness_flip.sh"],"baseline":"failed"}`
      in the request the fixture plugin logged.
- [x] A door that pauses for a whole test suite says what the pause found. —
      `GrantedCandidate::baseline_notice`, printed by `run_raw_one_shot` and
      `resolve_goal_wrapper`; the fleet door drops it and says why. Its exact
      wording is asserted in the red witness above.

## Which pillar

Reliability. The repository's headline claim is verified done rather than
claimed done, and the only path that can deliver it returned `Undecided` on
every run.

## Related

`#3835` gave `stella goal` a grant at all and `#3884` gave the fleet door its
per-task test command; both fixed a different missing half of the same chain.
`#1292` asked for the baseline restructuring as one half of using
affected-test selection as evidence; that half is this issue, and the other
half needs a selector the tool purge deleted.

macanderson:
## What & why

Every door minted a wrapper plugin's candidate grant with
`TestBaseline::NotRun`. `grant_shared_tree` in
`crates/stella-cli/src/wrapper_candidate.rs` called
`stella_plugin::test_plan(parsed, None)`, and that was the only call to
`test_plan` in shipping code — so the three other variants of `TestBaseline`
were unreachable from anything a user runs.

`plugins/stella-witness/main.py`'s `assess` feeds its oracle a failing
observation only when the grant's baseline says `failed`. Handed `not-run` it
never locks a command, reports `unobservable`, and `judge` under
`flip = "required"` turns that into `Undecided`. The one verification plugin
this repository ships could not credit a flip on any door:
`stella run --pipeline <plugin>`, `stella goal` and `stella fleet` alike.

The comment in `grant_shared_tree` gave the reason for the choice — "the host
does not spend a test run the plugin is about to spend itself. A wrapper
observes red in `before_turn`". A subprocess wrapper cannot. `exchange` in
`crates/stella-runtime/src/wrapper/subprocess.rs` spawns a fresh process for
every point, so anything a plugin saw at `before_turn` is gone by `after_turn`.
The host is the only party that spans both moments.

Nothing caught it because the conformance vectors write the baseline by hand:
`plugins/stella-witness/testdata/01-flip-achieved.request.json` carries
`"baseline": "failed"`, a value no production path could produce. The suite
proved the plugin credits a flip given red; no test asked whether red arrives.

**What this changes.** `grant_shared_tree` is now `async`. It runs the granted
invocation once, in the granted root, before the turn, and stamps what it saw
onto `TestPlan::baseline`. Three properties are deliberate:

- **It shares `run_in` with the `run_test` host call** (the new
  `wrapper_test_run::observe_baseline`), so pass and fail are decided in one
  place, off the exit status. A run killed at the deadline is `unobserved` on
  both paths rather than red on one and unobserved on the other.
- **A run that could not happen is never red.** A program that will not start
  comes back as `TestRunDenial::Failed`, which the caller records as
  `unobserved`. Scoring this host's own failure as red would satisfy a flip's
  precondition on infrastructure noise and credit the next clean run as a
  verified fix — the bug `CmdKind` exists to make unrepresentable.
- **The observation happens before the tamper watch is pinned.** A test run can
  rewrite generated files, and a watch pinned first would report this host's own
  run as a worker tampering with the witness.

The exemplar for the shape is the crate's own `wrapper_test_run` module rather
than an outside one: it already owned the spawn, the process group, the
deadline and the `TestObservation` type, and a second runner beside it is the
copy that drifts.

**Cost.** One run of the user's own `--test-command`, before the turn. That is
the price of the flip the flag asks for rather than an extra: with no before
there is nothing a passing suite afterwards could prove. `stella run` without
`--pipeline`, and any turn that names no test command, spend nothing. The
`stella goal` and `stella run` doors print what the pause found; the fleet door
stays silent, on the rule its own module already states for per-attempt notices.

**Where the design is written down.** `wrapper_candidate.rs` gains a
"When the baseline is observed, and every case where it cannot be" section: when
the observation happens, why the host has to be the one to make it, what each of
the four `TestBaseline` answers means, and why narrowing the invocation is a
separate question from whether the baseline is observed at all.

Closes #6421
Refs #1292

## The witness

Four, and the falsifier was run rather than argued. Replacing
`plan.baseline = observed;` with a discard makes
`a_red_suite_before_the_turn_reaches_the_plugin_as_red` fail with
`left: NotRun, right: Failed`; restoring it passes.

| Test | What it pins |
| --- | --- |
| `wrapper_candidate::tests::a_red_suite_before_the_turn_reaches_the_plugin_as_red` | a red suite reaches the plugin as `failed`, with the door's notice |
| `wrapper_candidate::tests::a_green_suite_before_the_turn_is_not_dressed_up_as_red` | a green suite is `passed`, never dressed up as red |
| `wrapper_candidate::tests::a_turn_with_no_test_command_observes_nothing_and_says_nothing` | no test command runs nothing and reports nothing |
| `wrapper_test_run::tests::the_baseline_reads_the_invocation_the_same_way_a_re_run_does` | one decision for both readings |
| `wrapper_test_run::tests::a_baseline_that_could_not_be_run_is_refused_rather_than_scored_red` | a program that will not start is refused, not scored red |
| `fleet_wrapped_dispatch_cli::a_plan_files_per_task_test_command_reaches_the_plugin_as_a_test_plan` | the baseline on the wire, end to end through a real door |

**One assertion was reversed, and it is the point of the change.**
`wrapper_candidate::tests::the_shared_tree_is_granted_with_the_test_the_host_would_run`
asserted `TestBaseline::NotRun` with the comment "the host did not run it, and
says so rather than claiming red". It now asserts `Failed`. No test was deleted.
Nine tests in `wrapper_candidate.rs`, one in `wrapper_test_run.rs` and two in
`round_driver.rs` became `#[tokio::test]` because the function they call is now
`async`; every name is unchanged.

## The gate

- [x] `cargo fmt --check` (scoped: `cargo fmt --check -p stella-cli`)
- [x] `cargo clippy -p stella-cli --all-targets -- -D warnings` — clean
- [x] `cargo test -p stella-cli` — the workspace run is CI's, per CLAUDE.md
- [x] `make guards-fast` — green, `check-prose.py` included
- [x] Docs updated: `wrapper_candidate.rs`'s module doc gains the section above,
      `wrapper_test_run.rs`'s gains the paragraph on the shared runner, and
      `resolve_goal_wrapper`'s doc says why the baseline is minted once for the
      arc rather than per round
- [x] `Closes #6421` appears above and as a commit trailer

## Fix over file

- [x] Extra fixes in this PR: one, in its own commit. Minting the grant now
      spends a whole test run, and on the two doors a person types at it sat in
      front of the last instant refusal a run can hit — a model id the catalog
      does not know. `run_raw_one_shot` and `run_goal_cmd` build the provider
      first now. Both are pure reorderings of adjacent statements with no data
      dependency, and nothing above the new position is paid, so the ordering
      rule the plugin resolve keeps is intact. The fleet door keeps the earlier
      order and says why: its provider is built against a per-worker config
      whose root is the attempt's tree, which is not a value until after the
      call, and `run_fleet`'s pre-flight already resolved the same variant.
- [x] Filed, with the reason a fix could not ride this PR: none deferred

`#1292` stays open and is not this PR's to close. It asks for affected-test
selection to count as verification evidence, and that has two halves: the
baseline restructuring, which is here, and a graph-driven affected-test
selector, which no longer exists in the tree — `run_tests` is a retired name in
`crates/stella-tool-facts/src/catalog.rs` and `CodeGraph::importer_paths` was
deleted with the tool purge (`crates/stella-graph/src/graph.rs`,
`the_importer_relation_lists_files_whose_imports_resolve_here`). Rebuilding it
against the plugin oracle is a separate design slice.

## Ground-rule check

- [x] No new cross-boundary type — `TestBaseline` and `TestPlan` already
      round-trip through `stella-plugin`'s wire corpus, and this PR changes
      which of their values production emits, not their shape

The judgement worth checking is whether the host should spend a test run at all.
The alternative on the table was leaving the baseline to the plugin, which the
old comment assumed; that is unavailable to any subprocess plugin, which is
every plugin this repository ships. An in-process wrapper could carry state
across points, so the host observing the baseline is redundant work for that one
case — and a host that skipped it there would make the grant mean two different
things depending on the transport, which is worse than a duplicated run.

## Summary by Sourcery

Observe granted test commands before wrapped turns so plugins receive accurate baselines and can verify genuine test flips.

New Features:
- Observe each granted test command before a wrapped turn and include its result in the plugin test plan, enabling verification plugins to evaluate flips from real CLI and fleet runs.
- Report pre-turn test observations for interactive and goal doors while preserving silent per-attempt behavior for fleet runs.

Bug Fixes:
- Ensure failing test baselines reach subprocess plugins instead of always being emitted as not-run, allowing required flips to be credited.
- Treat unstartable or incomplete test runs as unobserved rather than incorrectly scoring infrastructure failures as failed tests.
- Run baseline observation before tamper-watch snapshots so test-generated file changes are not mistaken for worker tampering.

Enhancements:
- Reuse the shared test runner for baseline and post-turn execution so both paths apply consistent pass, fail, timeout, and denial semantics.
- Make candidate and wrapper resolution asynchronous to support pre-turn test execution and update affected callers and tests accordingly.

Documentation:
- Document when test baselines are observed, how each baseline state is interpreted, and why observation is performed by the host before tamper watching.
- Document the shared test runner's role in producing consistent baseline and rerun observations.

Tests:
- Add witness coverage for failed, passed, absent, and unobservable baselines, shared runner semantics, and end-to-end fleet propagation.

Chores:
- Add a helper script for waiting on settled pull-request checks.

### DIFF
--- a/.watch-pr.sh
+++ b/.watch-pr.sh
@@ -1,0 +1,16 @@
+#!/bin/sh
+# Wait until no check on PR $1 is pending, then print the settled table.
+# `gh --json` colourises here whatever NO_COLOR says, so strip the escapes.
+i=0
+while [ "$i" -lt 120 ]; do
+  i=$((i + 1))
+  s=$(gh pr checks "$1" -R macanderson/stella --json name,bucket 2>/dev/null |
+    sed 's/\x1b\[[0-9;]*m//g')
+  if [ -n "$s" ] && printf '%s' "$s" | jq -e 'length>0 and (all(.[]; .bucket!="pending"))' >/dev/null 2>&1; then
+    printf '%s' "$s" | jq -r '.[] | "\(.bucket)\t\(.name)"' | sort
+    exit 0
+  fi
+  sleep 30
+done
+echo "TIMED OUT waiting for checks"
+exit 1
--- a/crates/stella-cli/src/agent.rs
+++ b/crates/stella-cli/src/agent.rs
@@ -489,6 +489,7 @@ pub async fn run_interactive(cfg: &Config, budget_limit: Option<f64>) -> Result<
             // grant before its own first round, and a session that never types
             // `/goal` never pays for a resolve it does not use.
             let goal_wrapper = goal::resolve_goal_wrapper(cfg, goal::DEFAULT_GOAL_WRAPPER, None)
+                .await
                 .and_then(|(resolved, candidate)| {
                     let bound = goal::bind_goal_wrapper(cfg, resolved, &sub_agents, &candidate)?;
                     Ok((bound, candidate))
--- a/crates/stella-cli/src/agent/goal.rs
+++ b/crates/stella-cli/src/agent/goal.rs
@@ -156,10 +156,17 @@ pub(crate) async fn run_raw_one_shot(
         ),
         None => None,
     };
+    let provider = build_provider(cfg)?;
     // Pinned *before* the turn runs, which is the whole content of the tamper
     // claim: an identity snapshotted afterwards vouches for nothing. The same
-    // call observes what the tests say before the work, which is the red half a
-    // flip needs (#1292) — and is a whole test run, so it says what it found.
+    // call observes what the tests said before the work, which is the red half
+    // a flip needs, and it says what it found because it is a whole test run.
+    //
+    // Below the provider rather than above it, because that is the last
+    // instant refusal a run can hit — a model id the catalog does not know —
+    // and nobody should sit through a test suite to be told they mistyped one.
+    // Nothing above this line is paid, so the rule the resolve keeps is intact:
+    // every refusal a user can act on lands before the first model call.
     let candidate = match &resolved {
         Some(_) => Some(
             crate::wrapper_candidate::grant_shared_tree(&cfg.workspace_root, test_command)
@@ -174,7 +181,6 @@ pub(crate) async fn run_raw_one_shot(
     {
         eprintln!("  ! {line}");
     }
-    let provider = build_provider(cfg)?;
     // Concrete `Arc<ToolRegistry>` (not `Arc<dyn ToolExecutor>`) so the
     // registry's ledgers are reachable after the turn — the trait object
     // hides them. It still coerces to `&dyn ToolExecutor` for the engine.
@@ -543,11 +549,12 @@ pub(crate) fn goal_plugin_missing_message(variant: &str, reason: &str) -> String
 /// Resolve the wrapper plugin that supplies the goal verb, and pin the
 /// candidate grant its `[oracle]` observes.
 ///
-/// Both happen before the provider is built and before a single paid call — a
-/// `--pipeline` naming nothing installed must fail as a typo, a workspace with
-/// nothing installed must fail as an install, and a `--test-command` the
-/// host's parser refuses must stop the run here rather than after it is paid
-/// for.
+/// Both happen before a single paid call — a `--pipeline` naming nothing
+/// installed must fail as a typo, a workspace with nothing installed must fail
+/// as an install, and a `--test-command` the host's parser refuses must stop
+/// the run here rather than after it is paid for. The caller places this after
+/// it has built a provider, because minting the grant runs the whole test
+/// suite and a model id the catalog does not know is an instant refusal.
 ///
 /// The grant is minted once for the arc, never per round. What the watch
 /// covers is the artifacts the test command names — the witness the flip is
@@ -667,11 +674,12 @@ pub(crate) fn bind_goal_wrapper(
 /// honoured now instead of being refused for want of a wrapper: every goal run
 /// has one.
 ///
-/// The bind happens before the provider is built and before a single paid
-/// call — the ordering [`run_raw_one_shot`] documents, for the same reason. A
-/// `--pipeline` naming nothing installed must fail as a typo, and a goal run
-/// with nothing installed at all must fail as an install, not after the money
-/// is spent.
+/// The bind happens before a single paid call — the ordering
+/// [`run_raw_one_shot`] documents, for the same reason. A `--pipeline` naming
+/// nothing installed must fail as a typo, and a goal run with nothing installed
+/// at all must fail as an install, not after the money is spent. It sits after
+/// the provider is built for the reason that door gives too: minting the grant
+/// runs the whole test suite, and every instant refusal belongs in front of it.
 pub async fn run_goal_cmd(
     cfg: &Config,
     goal: &str,
@@ -688,8 +696,8 @@ pub async fn run_goal_cmd(
         crate::enterprise_telemetry::ExecutionSurface::Goal,
     )?;
     let variant = pipeline.plugin().unwrap_or(DEFAULT_GOAL_WRAPPER);
-    let (resolved, candidate) = resolve_goal_wrapper(cfg, variant, test_command).await?;
     let provider = build_provider(cfg)?;
+    let (resolved, candidate) = resolve_goal_wrapper(cfg, variant, test_command).await?;
     let registry: std::sync::Arc<ToolRegistry> =
         std::sync::Arc::new(crate::write_dirs::registry_for(cfg));
 
--- a/crates/stella-cli/src/agent/goal.rs
+++ b/crates/stella-cli/src/agent/goal.rs
@@ -157,14 +157,23 @@ pub(crate) async fn run_raw_one_shot(
         None => None,
     };
     // Pinned *before* the turn runs, which is the whole content of the tamper
-    // claim: an identity snapshotted afterwards vouches for nothing.
+    // claim: an identity snapshotted afterwards vouches for nothing. The same
+    // call observes what the tests say before the work, which is the red half a
+    // flip needs (#1292) — and is a whole test run, so it says what it found.
     let candidate = match &resolved {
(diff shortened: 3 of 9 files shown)


macanderson:
Replaced by #6427, same tree, clean history.

A session scratch script — a watcher for this PR's own CI run — reached commit
`01bed9a30` through a `git add -A`. Removing it in a follow-up commit left an
edit-and-undo pair inside the branch, which `check-rebase-replay` refuses; its
only remedy is flattening the history, and this session does not force-push.

`fix/6421-baseline-observed-before-the-turn` carries the same three commits
with that file never present. `git diff` between the two heads is empty and
`make rebase-replay` is green on the new one.

Issue #6423, opened 2026-09-07
Put stella-store behind a feature in stella-tools, so an index-free build drops the bundled SQLite too


macanderson:
Sibling of #6286, which put `stella-graph` behind a `graph` feature in
`stella-tools`. That change removed all 22 tree-sitter edges from a
`--no-default-features` build. It did not remove the bundled SQLite, and
#6286's own Definition of done asked for both in one line.

## What is still there

`cargo tree -p stella-tools --no-default-features -e normal -i rusqlite`:

```
rusqlite v0.40.2
└── stella-store
    └── stella-tools
```

`stella-tools` takes `stella-store` unconditionally, and `stella-store` takes
`rusqlite` with a bundled `libsqlite3-sys`, so an index-free build still
compiles and links SQLite. That is the single most expensive C dependency
left in the index-free configuration, and it is the other half of what an
embedding host or a plugin-only build is trying not to pay for.

## Why it did not ride #6286

The `stella-graph` seam was two call sites outside `search/`. The store seam
is wider and reaches types other crates name:

- `crates/stella-tools/src/foundry_gate.rs` re-exports
  `stella_store::AdoptedTool`, so the type is part of this crate's public
  surface at a path callers already spell.
- `crates/stella-tools/src/registry.rs` carries an `mcp_usage` handle typed on
  the store.

Making the dependency optional means either a port trait those two sit
behind, or moving the pure `AdoptedTool` shape down to a near-leaf the way
#5119 moved the tool table into `stella-tool-facts`. Which of the two is a
design call, and it is larger than the session that closed #6286.

## Files

- `crates/stella-tools/Cargo.toml` — the unconditional `stella-store` edge and
  the existing `[features]` block to extend.
- `crates/stella-tools/src/foundry_gate.rs`, `src/registry.rs` — the two seams.
- `scripts/check-no-graph-tree.sh` — the guard #6286 added. It measures
  tree-sitter today and says in its own header that rusqlite is out of scope;
  extending it is where this work gets its evidence.
- `.github/workflows/tools-no-graph.yml` — the CI leg that runs it.

## Verify

- `cargo tree -p stella-tools --no-default-features -e normal -i rusqlite`
  reports no such package.
- `cargo test -p stella-tools --no-default-features` stays green.
- `cargo test -p stella-tools` is unchanged with default features.

## Which pillar

Efficiency and infra cost: a cold CI compile and every embedding host pays
for a bundled SQLite build it never opens.

## Definition of done

[...]

PR #6426, opened 2026-09-07
feat(stella-tools): put the code-graph index behind a graph feature

### LINKED ISSUE #6286: Put stella-graph behind a graph feature in stella-tools, so a small build can drop the code index

macanderson:
## What & why

`crates/stella-tools/Cargo.toml` took `stella-graph` — and `stella-embed` beside it — as unconditional path dependencies. Every host that linked the crate built nine tree-sitter grammars whether or not it wanted code search. Both are optional now, behind a `graph` feature.

`default = ["graph"]`, so `stella-cli` and every shipped build resolve exactly the dependency set they resolved before. This is a subtraction other hosts opt into, never one the CLI takes.

Measured, not asserted:

```
$ cargo tree -p stella-tools -e normal | rg -c tree-sitter
22
$ cargo tree -p stella-tools --no-default-features -e normal | rg -c tree-sitter
0
```

### Which of the two paths, and why

The issue named two ways to keep the surface honest when the feature is off. **This takes the second: `search` is not registered, and the catalog says so.**

`stella_tool_facts::catalog::Availability` grew a `BuildFeature(&'static str)` option — the seam its own doc comment already said it was being kept for. `search` declares `BuildFeature("graph")`, `stella_tools::BUILD_FEATURES` publishes what the binary compiled, `catalog::registered_with(features)` joins the two, and the registry's own pin reads that instead of `always_on()`. So the table and the live registry agree by construction in either build, and a host is never handed a schema for a tool it does not have.

The first path — moving `stella_graph`'s `DENY_DIRS`, `WorkspaceIgnore` and `admitted::{Admits, admission}` down to a leaf so the file-scan rung still answers with no index — is a second extraction in the shape #5119 used, and would have made this a much larger change for a rung that is not what the issue is about.

### What still works with the feature off

The mutation tools keep their graph seam. `graph_fact::Codegraph` is behind the feature; `graph_fact::workspace_graph()` returns it, or `Unindexed`, which answers `None` to both trait questions. `None` is already what `WorkspaceGraph` means by "no index answered", so a `write_file` or `delete_file` in an index-free build states no graph fact rather than a plausible zero.

### Which crates enable the feature, and why

Nobody enables it explicitly. Every crate that depends on `stella-tools` — `stella-cli`, `stella-tui`, `stella-runtime`, `stella-fleet`, `stella-serve` — takes it with default features, which is where `graph` lives. So the shipping binary is unchanged, and the feature is something a host outside this workspace turns *off*.

Closes #6286
Refs #5113

## The witness

Two:

**`crates/stella-tools/tests/graph_feature_witness.rs`** runs under either feature set and checks that the live registry advertises exactly `catalog::registered_with(BUILD_FEATURES)`. Falsified by hand: set `search`'s row back to `Always` and run it with `--no-default-features` —

```
assertion `left == right` failed: the live registry and the catalog disagree
about what this build registers — a host would be handed a schema for a tool
it does not have, or denied one it does
  left:  [... "save_state", "task_assign" ...]
  right: [... "save_state", "search", "task_assign" ...]
```

**`catalog.rs`'s `a_build_feature_row_registers_only_where_the_feature_is_compiled`** pins `satisfied_by` in both directions and shows the feature subtracts one row rather than reshaping the table (`registered_with(&[]) == always_on()`). On `main` this does not compile: the option it asserts on does not exist there.

## The gate

Per CLAUDE.md ("CI builds and tests; this laptop does not") nothing here was run over the workspace. The PR's evidence is its CI run. Scoped locally, both feature sets, all green:

- `cargo clippy -p stella-tools --all-targets -- -D warnings`
- `cargo clippy -p stella-tools --no-default-features --all-targets -- -D warnings`
- `cargo clippy -p stella-tool-facts --all-targets -- -D warnings`
- `cargo clippy -p stella-cli --all-targets -- -D warnings`
- `cargo doc -p stella-tools` with and without default features, `RUSTDOCFLAGS=-D warnings`
- `cargo test -p stella-tools --all-targets`, with and without default features
- `cargo test -p stella-tool-facts`
- `cargo test -p stella-cli --bin stella tool_docs`
- `make guards-fast` (prose, grade, lockfile, fmt, every toolchain-free guard)

The rustdoc runs are not ceremony. Both caught real breaks the clippy runs did not: a doc link to a now-`cfg`-ed `Codegraph` fails only with the feature off, and an outer `///` on the `pub mod search;` line re-parented `search.rs`'s `//!` header so rustdoc resolved its intra-doc links in `lib.rs`'s scope. Both are fixed; the second is written down where the next author will hit it.

- [x] Docs updated: `crates/stella-tools/README.md` gains a "The `graph` feature" section; `docs/tools/` regenerated with `make tool-docs-update`.
- [x] `Closes #6286` appears above **and** as a commit trailer.

### New CI leg

`.github/workflows/tools-no-graph.yml` — its own file, path-filtered, not a required check, on `wire-schema.yml`'s and `macos-check.yml`'s argument. It runs clippy, rustdoc and the tests for the feature-off configuration, then `scripts/check-no-graph-tree.sh` (`make no-graph-tree`), which asserts the tree-sitter edges are gone. That guard checks **both** directions: the feature-on tree must still carry grammars, or a rename that made the match empty would pass it silently forever.

## Fix over file

- [x] Extra fixes in this PR:
  1. **`crates/stella-cli/src/tool_docs.rs` cited a file that has moved.** Five references to `crates/stella-tools/src/catalog.rs`; the table lives in `stella-tool-facts` now. The citation was dead on every generated page under `docs/tools/`, which is why the regenerated diff touches all twenty files.
  2. **`RISK_NOTE` in the same file broke two of this repository's prose rules** — it opened with a filler adverb and closed on an `X, not Y` foil. Rewritten, which also bought back the reading-grade the longer path cost every generated page.
- [x] Filed, with the reason a fix could not ride this PR: #6423

## What this does NOT do — read this before ticking the DoD

**rusqlite is still in the index-free tree**, and #6286's Definition of done asked for it to be gone:

```
$ cargo tree -p stella-tools --no-default-features -e normal -i rusqlite
rusqlite v0.40.2
└── stella-store
    └── stella-tools
```

`stella-tools` takes `stella-store` unconditionally, and `stella-store` takes rusqlite with a bundled `libsqlite3-sys`. Dropping it means a port trait or a second near-leaf extraction for `foundry_gate`'s `stella_store::AdoptedTool` re-export and the registry's `mcp_usage` handle — a design call, larger than this session, and a different subject from the one this issue names. Filed as **#6423** with the repro, the two seams, and its own definition of done. `scripts/check-no-graph-tree.sh` says the same thing in its header rather than measuring something it cannot yet assert.

The DoD box for that line is ticked against the tree-sitter half only, and says so.

## Ground-rule check

- [x] No I/O added to `stella-core`; no new dependencies at all — this PR only makes two existing ones optional

**No tests were deleted.** Seven integration test files under `crates/stella-tools/tests/` gained a crate-level `#![cfg(feature = "graph")]`, so a `--no-default-features` build compiles them to empty test binaries instead of failing on a module that is not there. Every one of them still runs under the default features.

**One existing test learned to skip a row this build does not register.** `loop_comparability`'s sentinel drives every catalog row through a live registry; with `search` unregistered there is no success path to drive. The skip reads `catalog::registered_with(BUILD_FEATURES)` rather than a `cfg`, so it follows the same declaration the registry does, and a row that stops registering for a reason nobody declared still fails.

**`Availability::is_native` answers true for the new option.** The native `ToolRegistry` is still what registers `search` when the feature is on; `is_native` asks who registers, not whether this build did.

## Summary by Sourcery

Put the code-graph index and search capability behind an optional default-enabled feature while keeping feature-off tool registration, graph facts, documentation, and validation consistent.

New Features:
- Add an optional `graph` feature to `stella-tools`, enabled by default, for code-graph search and related dependencies.
- Expose build-aware tool availability so `search` is registered and documented only when the `graph` feature is compiled.

Bug Fixes:
- Prevent index-free builds from advertising or constructing the unavailable `search` tool.
- Make file mutation graph facts report no indexed result when the graph feature is disabled.

Enhancements:
- Keep the live tool registry, catalog, generated documentation, and compiled feature set consistent across feature configurations.

Build:
- Make `stella-graph` and `stella-embed` optional dependencies and add a dependency-tree guard confirming tree-sitter grammars are absent from feature-off builds.

CI:
- Add a path-filtered CI workflow that checks clippy, documentation, tests, and dependency pruning for the no-graph configuration.

Documentation:
- Document the `graph` feature and regenerate tool documentation to describe conditional availability and corrected catalog references.

Tests:
- Add witness coverage for feature-dependent registry and catalog agreement, and gate search-specific tests to graph-enabled builds.

Chores:
- Add a Makefile target for validating the index-free dependency tree while leaving rusqlite removal tracked separately.

### DIFF
--- a/.github/workflows/tools-no-graph.yml
+++ b/.github/workflows/tools-no-graph.yml
@@ -1,0 +1,94 @@
+name: tools-no-graph
+
+# The only build in this project that compiles `stella-tools` without the
+# code-graph index.
+#
+# `stella-graph` and `stella-embed` sit behind the crate's `graph` feature,
+# which is on by default (#6286) — so `ci.yml`, `make gate`, the pre-push hook
+# and every release build resolve the feature and never look at the other
+# configuration. An unbuilt configuration rots: the first `stella_graph::` in
+# an ungated module, or a doc link to a `cfg`-ed item, compiles green
+# everywhere a contributor looks and breaks only for the embedding host or
+# plugin-only build the feature exists for.
+#
+# So this compiles it with the feature off. `--all-targets` because the tests
+# are where the ungated reference lands: the search integration tests carry a
+# crate-level `#![cfg(feature = "graph")]` and would silently start failing to
+# compile if a new one forgot it. Plus `cargo doc`, because rustdoc resolves
+# intra-doc links per feature set and a link to a `#[cfg(feature = "graph")]`
+# item is an error here and nowhere else — which is the exact failure this
+# workflow caught while it was being written.
+#
+# Its own file rather than a job in `ci.yml`, for the reason
+# `wire-schema.yml` and `macos-check.yml` have one: a second full compile of
+# this crate's dependency tree is minutes a diff touching neither crate has no
+# use for. Not a required check.
+on:
+  pull_request:
+    paths:
+      - "crates/stella-tools/**"
+      - "crates/stella-tool-facts/**"
+      - "crates/stella-graph/**"
+      - "crates/stella-embed/**"
+      - "Cargo.lock"
+      - "Cargo.toml"
+      - ".github/workflows/tools-no-graph.yml"
+  push:
+    branches: [main]
+    paths:
+      - "crates/stella-tools/**"
+      - "crates/stella-tool-facts/**"
+      - "crates/stella-graph/**"
+      - "crates/stella-embed/**"
+      - "Cargo.lock"
+      - "Cargo.toml"
+      - ".github/workflows/tools-no-graph.yml"
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+concurrency:
+  group: tools-no-graph-${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  compile:
+    name: stella-tools still builds with no code-graph index
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
+
+      # Despite the action's name this does not build on `stable`:
+      # rust-toolchain.toml pins the repo to a concrete toolchain and rustup
+      # honors it for every cargo invocation inside the checkout. This step
+      # exists to put *a* rustup on PATH.
+      - name: Install Rust (rustup; the build uses the rust-toolchain.toml pin)
+        id: rust
+        uses: dtolnay/rust-toolchain@4cda84d5c5c54efe2404f9d843567869ab1699d4 # stable
+
+      - if: ${{ !cancelled() && steps.rust.outcome == 'success' }}
+        uses: Swatinem/rust-cache@f0d9c3887740aee45f6153b24b3a6b815192ec16 # v2
+
+      - name: clippy, no default features, all targets
+        if: ${{ !cancelled() && steps.rust.outcome == 'success' }}
+        run: cargo clippy -p stella-tools --no-default-features --all-targets --locked -- -D warnings
+
+      - name: rustdoc, no default features
+        if: ${{ !cancelled() && steps.rust.outcome == 'success' }}
+        env:
+          RUSTDOCFLAGS: -D warnings
+        run: cargo doc -p stella-tools --no-deps --document-private-items --no-default-features --locked
+
+      - name: test, no default features
+        if: ${{ !cancelled() && steps.rust.outcome == 'success' }}
+        run: cargo test -p stella-tools --no-default-features --locked
+
+      # The subtraction the feature exists for, asserted rather than assumed.
+      # `cargo check` succeeding says nothing about what got linked, and the
+      # nine tree-sitter grammars are the cost an embedding host is opting out
+      # of.
+      - name: the index-free build links no tree-sitter grammar
+        if: ${{ !cancelled() && steps.rust.outcome == 'success' }}
+        run: ./scripts/check-no-graph-tree.sh
--- a/Makefile
+++ b/Makefile
@@ -963,6 +963,14 @@ recheck-lock-compositions-test: ## Test the lock re-check sweep, its fail-open b
 release-lockfile-test: ## Test that the release version stamp leaves Cargo.lock resolvable, new members included (hermetic; not part of `gate`)
 	./scripts/test-release-lockfile.sh
 
+# Not a gate step: it resolves two dependency trees, which is seconds a diff
+# touching neither crate has no use for, and the configuration it measures is
+# one no shipped build uses. .github/workflows/tools-no-graph.yml is where it
+# runs for real, beside the compile of that configuration.
+.PHONY: no-graph-tree
+no-graph-tree: ## Assert an index-free stella-tools links no tree-sitter grammar (#6286)
+	@./scripts/check-no-graph-tree.sh
+
 # Deliberately not a gate step: it judges the MERGED tree, which is a question
 # no pre-merge run can answer. .github/workflows/main-canary.yml is where it
 # runs for real; this target is here so the logic can be exercised by hand.
--- a/crates/stella-cli/src/tool_docs.rs
+++ b/crates/stella-cli/src/tool_docs.rs
@@ -34,6 +34,7 @@
 //! `make record-golden` establishes the fixture's shape — an env-var-blessed
 //! test that rewrites it, with the plain test run as the drift guard.
 
+use std::borrow::Cow;
 use std::collections::{BTreeMap, BTreeSet};
 use std::path::{Path, PathBuf};
 
@@ -70,14 +71,14 @@ const DOCS_DIR: &str = "docs/tools";
 /// declared.
 const RISK_NOTE: &str = "\
 # risk_level: how bad one honest call is — a reviewed judgement, declared in
-# crates/stella-tools/src/catalog.rs beside the flags above and graded against
+# crates/stella-tool-facts/src/catalog.rs beside the flags above and graded against
 # the rubric on `ToolEntry::risk` (#2716, #3060). A DIFFERENT axis from
 # `read_only`: `delegate` mutates no file and spends real money, while
-# `task_create` mutates a board that dies with the session. Deliberately not
-# derived from the two booleans above — that would be a relabelling, not
-# information. A policy grant is expressed as a ceiling over this grade, and
-# every non-built-in tool (MCP, custom manifest) is graded `high` for being
-# unreviewed.";
+# `task_create` mutates a board that dies with the session. It is not derived
+# from the two booleans above. That would only relabel them.
+#
+# A policy grant is a ceiling over this grade. Every tool that is not a
+# built-in (MCP, a custom manifest) is graded `high` for being unreviewed.";
 
 // ── the committed example fixture ───────────────────────────────────────────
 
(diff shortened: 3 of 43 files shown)


macanderson:
### Answering the one ❌ — "no `rusqlite` in the no-default-features tree"

Sourcery reads the diff correctly. `rusqlite` is still there, and the row stays ❌ rather than being argued away. This is the deferral half of "Fix over file", and here is the case for it.

**The measurement.**

```
$ cargo tree -p stella-tools --no-default-features -e normal -i rusqlite
rusqlite v0.40.2
└── stella-store v0.9.393
    └── stella-tools v0.9.393
```

`rusqlite` does not arrive through `stella-graph`. It arrives through `stella-store`, which `stella-tools` takes unconditionally and which this issue names nowhere except in that one line. #6286's title, its "What is entangled today" section, its "two hard parts", and every file it lists are about `stella-graph`. The SQLite clause was carried along in the opening sentence on the assumption that the two dependencies rode the same edge. They do not.

**Why it could not ride this PR.** Dropping it means making `stella-store` optional, and two seams block that:

- `crates/stella-tools/src/foundry_gate.rs` re-exports `stella_store::AdoptedTool`, so the type is part of this crate's public surface at a path callers already spell.
- `crates/stella-tools/src/registry.rs` carries an `mcp_usage` handle typed on the store.

Neither is a `cfg`. Either a port trait goes in front of them, or the pure `AdoptedTool` shape moves down to a near-leaf the way #5119 moved the tool table into `stella-tool-facts`. Which of the two is a design call, and doing it inside this PR would have made a two-dependency change into a four-crate refactor with a much wider blast radius on the tool registry.

**Where it went.** Filed as **#6423**, written as a handoff: the repro above, both seams by path, the two candidate designs, which pillar it moves, and its own definition of done. It is referenced from this PR's description, and `scripts/check-no-graph-tree.sh` says the same thing in its own header rather than measuring something it cannot yet assert:

```sh
# rusqlite is out of scope here. `stella-tools` always takes `stella-store`,
# and `stella-store` always takes rusqlite, so a small build still bundles
# SQLite. Splitting that seam is its own job, tracked in `#6423`.
```

**What this PR does deliver against that objective**, measured rather than claimed:

```
$ cargo tree -p stella-tools -e normal | rg -c tree-sitter
22
$ cargo tree -p stella-tools --no-default-features -e normal | rg -c tree-sitter
0
```

`#6286`'s Definition of done has been edited to split that line in two — the tree-sitter half ticked against the numbers above, the SQLite half named and pointed at #6423 — so the issue records what shipped and what did not, rather than closing over a box nobody could honestly tick.

Issue #6429, opened 2026-09-07
stella no longer builds against context-graph-protocol HEAD: ContextQueryResult gained two attestation fields


macanderson:
Filed by the autonomous PR sweep. Not tracked anywhere I could find (searched `frame_attestations`, `ContextQueryResult`, `contextgraph`, `downstream canary` in this repo's issues).

## Problem

`context-graph-protocol`'s scheduled downstream canary — *"stella builds against CGP HEAD (advisory on PR, gating on schedule)"* — is **failing**, and it is the only failing check on that repo's `main`. It fails on purpose: the step's own comment says *"Scheduled run: failing so the break is visible rather than a warning on a green run."* So CGP's `main` is not broken; the canary is reporting that **this** repo is the one that will not compile against it.

Run: https://github.com/macanderson/context-graph-protocol/actions/runs/34122768843/job/101744376868 (CGP HEAD `201f2ce6`)

```
error[E0063]: missing fields `frame_attestations` and `result_attestation`
              in initializer of `ContextQueryResult`
  --> crates/stella-context/src/provider.rs:271:12
error[E0063]: missing fields `frame_attestations` and `result_attestation`
              in initializer of `ContextQueryResult`
  --> crates/stella-context/src/retrieval/outcome.rs:345:9
error: could not compile `stella-context` (lib) due to 2 previous errors
```

## Why this is invisible here

The root `Cargo.toml` pins the protocol crates by exact version:

```
contextgraph-types      = "=0.1.2"
contextgraph-host       = "=0.1.2"
contextgraph-trace      = "=0.1.2"
contextgraph-conformance = "=0.1.2"
```

`0.1.2` has no attestation fields, so this repo's own CI is green and stays green. Only the canary, which builds this tree against CGP *source*, sees it. Note the gap is wider than the two fields: **CGP's workspace version is now `2.0.0`** (`Cargo.toml:16`), so the eventual pin move is `0.1.2` → `2.x`, not a patch bump.

## What changed upstream

`contextgraph-types/src/query.rs:77` — `ContextQueryResult` grew the wire home of `SPEC.md` §6.5's evidence (§6.5.5, F11–F13):

```rust
#[serde(default, skip_serializing_if = "Vec::is_empty")]
pub frame_attestations: Vec<FrameAttestation>,
#[serde(default, skip_serializing_if = "Option::is_none")]
pub result_attestation: Option<ProvenanceAttestation>,
```

[...]

Issue #4359, opened 2026-08-23
Adopt context-graph 2.0.0: ContextQueryResult gained two attestation fields and stella-context no longer compiles against CGP HEAD


macanderson:
## Problem

`context-graph-protocol`'s scheduled **Downstream Canary** builds stella against CGP `main` HEAD and runs `cargo test -p stella-cli -p stella-graph -p stella-context`. It is red, and since CGP moved to 2.0.0 it fails at compile time rather than on an assertion:

```
error[E0063]: missing fields `frame_attestations` and `result_attestation`
              in initializer of `ContextQueryResult`
```

CGP 2.0.0 added two fields to `ContextQueryResult` (`contextgraph-types/src/query.rs`): `frame_attestations: Vec<FrameAttestation>` and `result_attestation: Option<ProvenanceAttestation>`. Stella still pins `contextgraph-types` / `-host` / `-trace` / `-conformance` at `"=0.1.2"` in the root `Cargo.toml` `[workspace.dependencies]`, so stella's own `main` is green. This is forward incompatibility with unreleased upstream, which is what the canary exists to catch early.

## Why it cannot be fixed today

crates.io serves only `0.1.2` of every `contextgraph-*` crate (index checked 2026-09-05). Pinning `=2.0.0` fails resolution and `lockfile-sync`, and no partial adaptation compiles against 0.1.2, because the fields do not exist there.

The upstream blocker is **macanderson/context-graph-protocol#102** — the `crates-io` GitHub Environment and token that `release.yml` needs do not exist, so nothing has been published since the by-hand `0.1.2` cut. CGP#16, named in the first version of this issue, is closed and is no longer the blocker.

## What stella attests

Stella's built-in context providers hold no signing key and mint no commitment, so both new fields are empty for every answer stella produces: `Vec::new()` and `None`. Write them out at each site with a comment naming the producer that owes them, rather than `..Default::default()`, which would absorb the next upstream field in silence.

Both fields carry `#[serde(default, skip_serializing_if = ...)]` upstream, so an unattested result serializes exactly as it does today and invariant 4 holds.

## Duplicates folded in

[...]

PR #6436, opened 2026-09-07
fix(stella-protocol): state where the record digest stops matching CGP's

macanderson:
## What & why

`crates/stella-protocol/src/hash.rs`'s module header claimed this workspace's
`record_hash` preimage is "byte-identical to CGP's for export interop". The
root `Cargo.toml` carried the same claim above `serde_json_canonicalizer`.

Reading the rule the claim is about settles it. CGP's
`contextgraph-types/src/record_attest.rs::record_hash_preimage` (workspace
`version = "2.0.0"`) removes the top-level `record_hash` member and
canonicalizes under RFC 8785, and normalizes no nulls:

```rust
let mut preimage = record.clone();
preimage.as_object_mut().ok_or(RecordHashError::NotAnObject)?.remove(RECORD_HASH_MEMBER);
serde_json_canonicalizer::to_vec(&preimage)
```

Step 3 of this module's pipeline strips explicit nulls recursively, so an
explicit `null` and an absent field hash the same here and differently there.
The two preimages agree byte for byte on a record carrying no explicit null,
and disagree on one that does. The claim was true only under a condition
neither copy stated.

Both copies now state the condition, and a test pins both halves, so the
boundary is a checked fact rather than a doc promise. Nothing about the
hashing pipeline changed — this repository's null normalization is ADR 0004's
deliberate choice and every stored digest is unmoved.

The version half of the claim holds and is unchanged, re-verified against CGP
`main`: its workspace pins `serde_json_canonicalizer = "0.3.2"` and reaches it
through `contextgraph-types`' `record-hash` feature
(`record-hash = ["dep:sha2", "dep:serde_json", "dep:serde_json_canonicalizer"]`),
which is the same version this workspace pins.

Refs #4359

### What this does for #4359, and what it does not

#4359's checklist carries one box that does not need the external publish:
*"`contextgraph-conformance`'s `record_hash` coupling (noted at
`Cargo.toml:85`) is re-checked against the version actually landed"*, folded
in from #6429. This is that re-check, done against CGP `main` at 2.0.0, and it
found the coupling comment overstated. `#4359` stays open — every other box
needs `contextgraph-*` 2.0.0 on crates.io, and the sparse index still serves
only `0.1.0` / `0.1.1` (yanked) / `0.1.2` for all four crates, so no pin bump
resolves and the two new `ContextQueryResult` fields cannot be written against
the version this workspace can build. The upstream blocker is
macanderson/context-graph-protocol#102, which needs a person to create the
`crates-io` GitHub Environment and its registry token.

## The witness

- [x] No witness needed (pure refactor / docs / CI) — because: the hashing
      pipeline is byte-for-byte unchanged, which is the point. What lands is a
      corrected claim plus
      `stella_protocol::hash::tests::strip_nulls_is_the_only_divergence_from_cgps_preimage`,
      a characterization test that reimplements CGP's rule and pins both
      directions: byte-identical preimages for a null-free record, and
      differing preimages (`{"a":1}` against `{"a":1,"b":null}`) for one
      carrying an explicit null. It fails if `strip_nulls` is removed, if the
      canonicalizer is swapped, or if either side's rule moves. It is
      reimplemented rather than imported because `stella-protocol` takes no
      `contextgraph-types` dependency and acquiring one to quote a rule this
      short would invert the crate's position in the graph.

## The gate

- [x] `cargo fmt --check`
- [x] `cargo clippy -p stella-protocol --all-targets -- -D warnings` (scoped
      per CLAUDE.md — CI runs the workspace)
- [x] `cargo test -p stella-protocol --lib` — 305 passed, 0 failed
- [x] `make guards-fast` green, and
      `make doc-warnings CARGO_SCOPE="-p stella-protocol"` clean
- [x] CLA signed
- [x] `Refs #4359` appears both above and as a commit trailer

## Fix over file

- [x] Extra fixes in this PR: none — the corrected claim is the whole change.
- [x] Filed, with the reason a fix could not ride this PR: nothing was
      deferred. #4359 stays open on its existing external blocker, which this
      PR leaves unchanged.

## Ground-rule check

The divergence is deliberate and stays. Stella normalizes nulls so an explicit
`null` and an absent field produce one digest, which is what lets a record type
add a `#[serde(skip_serializing_if)]` option without moving every prior hash.
Adopting CGP's rule instead would rewrite every stored `record_hash` in
`.stella/private/context.db`, and this PR proposes nothing of the kind.

Whether the divergence bites in practice depends on whether any record stella
hashes serializes an explicit `null`; the record types carry
`skip_serializing_if = "Option::is_none"`, so today it is a statement about the
algorithms rather than an observed mismatch. That is why the condition belongs
in the comment and in a test: the next record type that omits the attribute
would move the boundary with nothing to notice it.

## Summary by Sourcery

Correct the record-hash interoperability claim and pin the intentional null-normalization boundary against CGP's hashing rule.

Enhancements:
- Clarify that Stella's record-hash preimage matches CGP only for records without explicit null-valued object members, while preserving Stella's deliberate null normalization.
- Document the precise boundary between Stella's and CGP's record-hashing rules.

Tests:
- Add characterization coverage confirming matching preimages for null-free records and the expected divergence for nested, array-contained, and array-element nulls.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -84,9 +84,9 @@ serde_json = "1"
 # RFC 8785 JSON Canonicalization Scheme — the version the Context Graph Protocol
 # workspace pins for `contextgraph-types`' `record-hash` feature, so a shared
 # preimage canonicalizes to the same bytes on both sides. Byte-identity of the
-# whole record digest is narrower than that: stella's preimage also drops
-# explicit nulls and CGP's does not, so the two agree on a null-free record
-# only. `stella_protocol::hash` states the boundary and pins it with a test.
+# whole record digest is narrower: stella's preimage also drops `null`-valued
+# object members, and CGP's keeps them, so the two agree on a record whose
+# objects carry no such member. `stella_protocol::hash` pins that boundary.
 serde_json_canonicalizer = "0.3.2"
 thiserror = "2"
 tokio = { version = "1", features = ["rt-multi-thread", "macros", "time", "process", "fs", "io-util", "sync"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -81,9 +81,12 @@ contextgraph-trace = "=0.1.2"
 contextgraph-conformance = "=0.1.2"
 serde = { version = "1", features = ["derive"] }
 serde_json = "1"
-# RFC 8785 JSON Canonicalization Scheme — pinned to the SAME version the Context
-# Graph Protocol crates use (contextgraph-conformance), so stella's record_hash
-# preimage is byte-identical to CGP's for future export interop.
+# RFC 8785 JSON Canonicalization Scheme — the version the Context Graph Protocol
+# workspace pins for `contextgraph-types`' `record-hash` feature, so a shared
+# preimage canonicalizes to the same bytes on both sides. Byte-identity of the
+# whole record digest is narrower than that: stella's preimage also drops
+# explicit nulls and CGP's does not, so the two agree on a null-free record
+# only. `stella_protocol::hash` states the boundary and pins it with a test.
 serde_json_canonicalizer = "0.3.2"
 thiserror = "2"
 tokio = { version = "1", features = ["rt-multi-thread", "macros", "time", "process", "fs", "io-util", "sync"] }
--- a/crates/stella-protocol/src/hash.rs
+++ b/crates/stella-protocol/src/hash.rs
@@ -10,9 +10,15 @@
 //!    complements the `#[serde(skip_serializing_if = "Option::is_none")]`
 //!    absent-option omission on the record types);
 //! 4. **RFC 8785 JCS** — sort keys, minimal whitespace, canonical numbers, via
-//!    `serde_json_canonicalizer` (the same crate + version CGP uses, so the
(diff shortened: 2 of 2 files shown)


macanderson:
Both Sourcery comments were right and are fixed in `a79c3272`.

**`Cargo.toml:89` (nitpick).** `strip_nulls` retains on object maps and recurses into arrays without removing their elements, so `{"items":[null,1]}` canonicalizes identically on both sides. "explicit null" was wider than what the code does. Both copies of the claim now say `null`-valued object member, and name the array element as the edge of the boundary.

**`hash.rs:201` (testing).** Correct — the test passed with the recursion in `strip_nulls` flattened to the top level. It now covers a nested member (`{"outer":{"a":1,"b":null}}`), a member of an object inside an array (`{"items":[{"a":1,"b":null}]}`), and the null array element that stays on both sides. Verified by mutation rather than asserted: deleting the two recursive arms of `strip_nulls` and re-running gives

```
  left: "{\"outer\":{\"a\":1,\"b\":null}}"
 right: "{\"outer\":{\"a\":1}}"
```

with the recursion restored and `cargo test -p stella-protocol --lib` back to 305 passed, 0 failed.

PR #6441, opened 2026-09-07
feat(loop-bench): fail the nightly when most of its trials never finished

### LINKED ISSUE #6025: Most nightly-bench trials never finish, and nothing gates on that
## Problem

Most `nightly-bench` trials do not finish. Harbor kills them on its own agent
timeout, and that outcome does not gate, so the workflow can report a healthy
loop over a run in which almost nothing ran to completion.

Counted from the uploaded artifacts:

| Run | Trials that raised | Exception |
|---|---|---|
| 33956998843 (2026-09-05) | 3 of 4 | `AgentTimeoutError` after 750s / 900s |
| 33859394391 (2026-09-04) | 3 of 4 | two `AgentTimeoutError`, one deliberate stop |
| 33741033194 (2026-09-03) | 3 of 4 | two `AgentTimeoutError`, one deliberate stop |
| 33387694069 (2026-08-31) | 1 of 4 | `AgentTimeoutError` after 750s |

The table already prints the warning ("the pass rate is over a run that partly
did not happen"), so the harness is telling the truth. Nothing acts on it. With
the loop verdict corrected in #5613, a night where every trial times out reads
`LOOP: 0/4 broken` and passes.

## What to decide

Whether "most of tonight's trials never finished" should be a gate of its own,
beside loop health and the pass-rate floor. The two candidate shapes:

- A crash ceiling: fail when more than N of the requested trials raised. It
  needs a number, and a number wants a few nights of history to set it from —
  the same argument `LOOP_BENCH_MIN_PASS` makes for staying at 0.
- Leave it ungated and report it louder, on the grounds that a harbor timeout
  is the machine's fact and not the engine's.

Answer that first. Whichever way it goes, write the reason down, because the
current state reads as a decision nobody made.

Worth reading beside this: #3528 and #3753, on tasks lost to time rather than
to being wrong.

## Files

- `bench/loop-bench/src/lib.rs` — `Tally::crashed`, `loop_broken`,
  `below_pass_floor`, and the table's crash warning.
- `bench/loop-bench/src/main.rs` — the exit codes the workflow reads.
- `.github/workflows/nightly-bench.yml` — `LOOP_BENCH_MIN_PASS` and the
  harbor timeout.

## How to verify

```
gh api repos/macanderson/stella/actions/runs/<run id>/artifacts
gh api repos/macanderson/stella/actions/artifacts/<artifact id>/zip > a.zip
```

`report.json` carries `tally.crashed` and `tally.total` per run.

## Definition of done

- [x] The question above is answered in this issue: a crash ceiling, or a
      stated decision to leave it ungated. — answered in the comment below and
      in ADR 0038: a ceiling, expressed as a share of the requested trials.
- [x] If a ceiling lands, it has its own exit code and its own test, and the
      number it uses is taken from observed nights rather than guessed. —
      exit `8` (`unfinished_exit`), two tests in `bench/loop-bench/src/tests.rs`,
      and the rule was checked against the 29 nightly `report.json` artifacts
      from 2026-08-10 to 2026-09-07 rather than picked.
- [x] If it stays ungated, the reason is written into `bench/loop-bench`'s
      crate docs where the other verdict decisions live. — the branch not
      taken; the reasoning is written there anyway, as the third section beside
      the loop-health and pass-floor ones.

macanderson:
## What & why

`#6025` asked a question and asked for the answer to be written down: should
"most of tonight's trials never finished" be a gate of its own, or stay
ungated and be reported louder? It also said the current state reads as a
decision nobody made.

**The claim on the issue was that it is blocked on paid benchmark runs. It is
not.** Nothing here spends anything. The 29 nights of history the decision
wanted are already uploaded as workflow artifacts, and every number below was
read out of them with `gh api` — free.

Decided, per SCR-002: **gate it**, as a share rather than a tuned count.

- `loop_bench::most_trials_crashed` is red when more than half the requested
  trials raised. It reads `Tally::crashed`, which is harbor's own record.
- `unfinished_exit` in `bench/loop-bench/src/main.rs` turns that into exit
  `8`, in both the single-config and `--compare` paths. It runs after loop
  health (`1`) and reconciliation (`7`) and before the pass floor (`4`), so a
  run that trips several still names the actionable half, and no pass rate or
  comparison winner is ever published over a sample this thin.
- The table prints the majority case in the words the exit code uses, and
  `nightly-bench.yml`'s summary spells out `8`.

**Why a share and not a number.** The issue argued a ceiling needs a number
and a number needs history. A pass floor does, because it is a bar on a
quality measurement whose normal value nobody knows. This is a bar on how much
of the run happened, and its value follows from what the run asked for: below
half, the survivors are not the pinned task set `nightly-bench.yml` calls its
fixed seed. A share also follows `--tasks`, `--n` and `--trials` when they
move the denominator, where a count pinned to four would come to mean
something else.

**What it costs, measured.** Every `nightly-bench` run since 2026-08-10 that
uploaded a report — 29 of them — asked for four trials. Read from each run's
own `report.json` (`tally.crashed` / `tally.total`):

| crashed / 4 | nights |
|---|---|
| 0 | 1 |
| 1 | 14 |
| 2 | 5 |
| 3 | 9 |
| 4 | 0 |

The rule fires on the nine nights that lost three. Eight were already red for
loop health. The ninth is the night this exists for: run **33306583016**, on
2026-08-30, reported a **green** nightly on one solved trial and three that
never finished. So the gate flips exactly one verdict in 29 nights — it is not
the always-red gate that argued for leaving this ungated.

Closes #6025

## The witness

Two, in `bench/loop-bench/src/tests.rs`:

- `a_night_where_most_trials_died_no_longer_reports_green` builds run
  33306583016's shape from real trial directories: four requested, three
  `AgentTimeoutError` after real work, one solved. It asserts what `main`
  does — `loop_broken == 0`, and `below_pass_floor` off — and then that the
  new gate fires.
- `losing_half_the_trials_still_reports_its_figures` pins the boundary: 0, 2
  and 3 crashes of four, plus a one-trial run.

Verified the artisanal way rather than by relying on a compile error: with
`most_trials_crashed` forced to `false` (the `main` behaviour, no such gate),
both tests FAIL; restored, all 48 lib tests pass.

## The gate

- [x] `cargo fmt --check` — clean
- [x] clippy — `cargo clippy -p loop-bench --all-targets -- -D warnings`, clean.
      Scoped to the crate per CLAUDE.md ("CI builds and tests; this laptop does
      not"); the workspace run is CI's.
- [x] tests — `cargo test -p loop-bench`: 48 + 5 pass
- [x] `make doc-warnings CARGO_SCOPE="-p loop-bench"` — clean
- [x] `make guards-fast` — green, including `prose`, `line-citations`,
      `adr-numbering` and `doc-links`
- [x] Docs updated: the crate docs gain a "third gate" section beside the two
      already there, `bench/loop-bench/README.md` gains the exit-`8` row and
      the reasoning, and `nightly-bench.yml`'s header says what it now gates on
- [x] CLA signed
- [x] `Closes #6025` appears above and as a commit trailer

## Fix over file

- [x] Extra fixes in this PR: two. The module header and README paragraph that
      both said "a crash is deliberately **not** an exit code" were left true
      only of one crash, so they are rewritten rather than left contradicting
      the code beside them. And
      `website/content/docs/guides/ci-engine-gate.mdx` lists `loop-bench`'s
      exit codes for a reader copying the pattern; exit `8` joins that list,
      since a gate reporting over a night that mostly did not run is the
      guide's own thesis.

## Ground-rule check

**The ADR.** SCR-002 says a design decision is recorded as an ADR, so this is
`docs/adr/0038-a-night-that-mostly-died-is-not-a-run.md`. 0037 is claimed by
open PR #6437, checked before taking 0038.

**The alternative I rejected, and its strongest form.** "A harbor timeout is
the machine's fact, not the engine's" is the reasoning that already keeps
`CRASHED` a row verdict rather than an exit code, and it is right for one dead
trial in four. It stops being right at a majority: the workflow already
refuses to start with no provider key on the stated grounds that a green
nightly that measured nothing is the exact failure `bench.yml` complains
about, and three of four trials dying reaches that state by a different road.

**This makes the nightly red more often, on purpose.** Red here is
actionable: it says the task set, the agent timeout, or the runner cannot
finish what the night asks for. The remedy is to fix one of those, never a
greener reading of the same night. The observed rate — one extra red in 29
nights — is what makes that affordable.

## Summary by Sourcery

Gate nightly benchmark results when a majority of requested trials fail to finish, preventing thin samples from producing misleading green verdicts.

New Features:
- Add a nightly benchmark gate that fails when more than half of the requested trials crash or never finish.
- Report unfinished-majority failures through exit code 8 in single-configuration and comparison runs, including clear table and CI summaries.

Bug Fixes:
- Prevent nightly benchmarks from reporting a green verdict or publishing misleading pass-rate and comparison results when most requested trials did not finish.

Enhancements:
- Document the third benchmark gate, its precedence, rationale, and observed impact across nightly history.

CI:
- Update the nightly benchmark workflow to recognize and explain exit code 8.

Documentation:
- Record the gating decision in ADR 0038 and update the ADR index, manifest, benchmark README, and CI engine-gate guide.

Tests:
- Add witness coverage for a real majority-crash nightly shape and the strict-majority boundary.

### DIFF
--- a/.github/workflows/nightly-bench.yml
+++ b/.github/workflows/nightly-bench.yml
@@ -15,6 +15,12 @@ name: nightly-bench
 # it caught cycling. Those signals are visible on a cheap model, which is what
 # makes a nightly affordable at well under a dollar a night.
 #
+# It also gates on whether the night happened at all: a run where more of the
+# requested trials raised than finished exits 8, because a loop verdict over
+# the surviving minority describes a task set nobody asked for. Harbor's agent
+# timeout is what usually kills them, and the remedy is a task set or a timeout
+# this runner can finish, never a greener reading of the same night.
+#
 # NOT on `pull_request`/`push`: every run spends real money and needs Docker,
 # harbor, and a provider key. Schedule + manual dispatch only, exactly like
 # `live-smoke.yml`.
@@ -296,6 +302,7 @@ jobs:
               5) echo "**FAIL (report)** — the run completed but the JSON report could not be written." ;;
               6) echo "**FAIL (no winner)** — \`--require-winner\` was set and no arm cleared both bars." ;;
               7) echo "**FAIL (reconciliation)** — the job directory held trials this run did not ask for, so every figure above covers a task set that is not the requested one." ;;
+              8) echo "**FAIL (unfinished)** — more of the requested trials raised than finished, so every figure above covers a night that mostly did not happen. Look at the exceptions on the table before reading anything else." ;;
               "") echo "**FAIL** — loop-bench never ran; an earlier step failed." ;;
               *) echo "**FAIL** — exit $EXIT_CODE is outside the harness's contract: a build failure or a crash, not a verdict." ;;
             esac
--- a/bench/loop-bench/README.md
+++ b/bench/loop-bench/README.md
@@ -117,15 +117,28 @@ have exited on a step cap, and treating that as a death would invent crashes.
 | `5` | the run finished and the `--json-out` report could not be written (the JSON is dumped on stdout instead) |
 | `6` | `--compare --require-winner` and no arm cleared both bars |
 | `7` | the job dir holds trials this run did not ask for, so every figure covers a task set that is not the requested one |
+| `8` | more of the requested trials raised than finished, so every figure covers a run that mostly did not happen |
 
 `1` outranks `4`: when both fire, the loop failure is the actionable one, and a
-broken loop explains the missing passes anyway. `1` outranks `6` and `7`
+broken loop explains the missing passes anyway. `1` outranks `6`, `7` and `8`
 likewise — a broken loop makes an arm's numbers untrustworthy as a comparison.
+`8` outranks `4` and `6`: a pass rate or a winner read off a run that mostly
+did not happen is not a measurement.
 
-A crash is deliberately **not** an exit code. The gate is loop health, and a
-trial the machine killed is not evidence about the loop — the same reason
-`UNREADABLE` and `BUDGET-CAP` do not gate. What it must never again be is
-silent, or dressed as `ran (unsolved)`.
+One crash is **not** an exit code. The gate is loop health, and a trial the
+machine killed is not evidence about the loop — the same reason `UNREADABLE`
+and `BUDGET-CAP` do not gate. What it must never again be is silent, or
+dressed as `ran (unsolved)`.
+
+`8` is where that reasoning runs out. Once the killed trials outnumber the
+live ones, what survives is not the task set the run asked for, and a verdict
+over it describes a night that mostly did not happen. The rule is a
+strict majority rather than a tuned count: it needs no history to set, and it
+follows `--tasks`, `--n` and `--trials` when they move the denominator. The
+29 nightly runs from 2026-08-10 to 2026-09-07 that uploaded a report each
+asked for four trials — one lost none, fourteen lost one, five lost two, nine
+lost three, none lost all four. The rule fires on those nine, eight of which
+were already red for loop health; the ninth reported green.
 
 ## Requested vs reported (#1299)
 
--- a/bench/loop-bench/src/lib.rs
+++ b/bench/loop-bench/src/lib.rs
@@ -89,6 +89,33 @@
 //! which of the two failed. It is off unless a floor is configured, because
 //! the pass rate this harness produces is a flash-tier model's under a cap:
 //! a floor is only meaningful once a job's own history has established one.
+//!
+//! # The third gate: most of the run never finished (`#6025`)
+//!
+//! A trial harbor killed is not evidence about the loop, which is why
+//! `CRASHED` is a row verdict and not an exit code. That holds while the dead
+//! trials are the minority. It stops holding once they outnumber the live
+//! ones: what survives is not the pinned task set the nightly job calls its
+//! fixed seed, and a verdict over it is a verdict about a run that mostly did
+//! not happen.
+//!
+//! [`most_trials_crashed`] is that check — red when more than half the
+//! requested trials raised. It makes a third claim, distinct from the other
+//! two: `loop_broken` says the loop misbehaved, [`below_pass_floor`] says the
+//! agent solved too little, and this says the night measured too little to
+//! say either.
+//!
+//! A majority rather than a tuned count, so it needs no history to set and it
+//! survives a change to the task set: `--tasks`, `--n` and `--trials` all move
+//! the denominator, and a count pinned to four would then mean something else.
+//!
+//! What it costs is measured, not assumed. The 29 nightly runs from
+//! 2026-08-10 to 2026-09-07 that uploaded a report each asked for four trials.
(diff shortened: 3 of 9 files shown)


macanderson:
## This PR and #6437 write the same three anchors — merge #6437 first

Both branches are green on their own and neither one's CI can see this: it is the shared-cell shape `docs/adr/README.md` § "The number is a shared cell" describes, and both diffs happen to stop one line above that heading.

The ADR **numbers** are fine. This PR took 0038 after checking that #6437 held 0037, and the two added files do not collide. What collides is everything that indexes them. Read off the two diffs:

| File | Hunk | This PR adds | #6437 adds |
|---|---|---|---|
| `docs/adr/README.md` | `@@ -103,6 +103,7 @@` | the `0038` table row, immediately after `\| [0036]…` | the `0037` row, immediately after `\| [0036]…` |
| `docs/adr/README.md` | `@@ -214,6 +215,12 @@` | the 0038 paragraph, immediately after the 0036 paragraph | the 0037 paragraph, in the same position |
| `docs/manifest.json` | `@@ -185,6 +185,11 @@` | the `adr/0038-…` block, immediately after `adr/0036-…` | the `adr/0037-…` block, in the same position |

Identical hunk headers, identical context lines, different insertions at each. The second of the two to merge conflicts.

The quiet failure is worse than the loud one. If a conflict resolution takes both sides at the anchor without reordering, the README table reads 0036, 0038, 0037, and `docs/manifest.json` — which is key-sorted — carries `adr/0038-…` ahead of `adr/0037-…`. Both files still parse, so nothing downstream complains.

### Suggested order

Merge **#6437 (0037) first**, then rebase this branch and place the row, the paragraph and the manifest block *after* 0037 rather than after 0036. That direction is the cheaper one: 0037 sorts first everywhere the two appear, so this branch is the one that has to yield either way.

No change pushed here, and nothing on either PR touched. This is a sequencing dependency between two open PRs rather than a defect in this diff — the branch is correct as it stands, and only stops being correct once #6437 lands.

---

macanderson:
**Keep your 0038 — no action needed here.** Posting so this PR does not also move, which would re-collide.

An automated PR sweep found `0038` claimed by two open branches:

| branch | file |
|---|---|
| this PR | `docs/adr/0038-a-night-that-mostly-died-is-not-a-run.md` |
| #6443 | `docs/adr/0038-a-live-smoke-provider-is-armed-or-declared-unarmed.md` |

Both carry main's 0037 (#6437), so both were numbered against the same tree and neither is wrong on its own — the shared-cell break `docs/adr/README.md` § "The number is a shared cell" describes. `adr-numbering` passes for each branch alone and nothing re-runs it when `main` moves, so the second to merge would have landed two records at 0038.

**#6443 moved to 0039** (`ec36df8`) and this PR keeps **0038**. It was the cheaper half to move: this PR is awaiting review with its number already cited in its own prose, ADR and tests.

The only thing worth checking here is that nothing in this branch cites `0039`, so the two records stay disjoint.

---

Issue #6442, opened 2026-09-07
Buy the six missing provider keys and top up Anthropic and Z.ai so live-smoke covers all nine adapters


macanderson:
## Problem

The live provider smoke suite covers one of its nine adapters. The other eight
are declared unarmed in `crates/stella-model/tests/live_smoke/arming.rs`, each
with the reason it cannot run:

- **No credential in the repository's Actions secrets** (six): `openai`,
  `gemini`, `deepseek`, `xai`, `bedrock`, `vertex`.
- **A working key over an empty account** (two): `anthropic` answers HTTP 400
  `Your credit balance is too low`; `zai` answers HTTP 429 `Insufficient
  balance or no resource package`. Both quoted from run 34126157647
  (2026-09-07).

Only `openrouter` makes a real call. It costs about two thousandths of a cent
per run.

That suite exists because the Anthropic adapter shipped a broken `thinking`
wire shape for months behind mock-only coverage, and a live call is what found
it (#240). While eight adapters never reach their endpoint, that regression
class is uncovered for all eight.

`#5595` fixed the reporting half: the job is green, it says which providers it
did not reach, and it goes red the moment a declared row stops being true. It
could not fix this half, which needs money and credentials.

## What a human has to supply

Six credentials, added as repository secrets (Settings → Secrets and variables
→ Actions):

| Secret | Provider | Notes |
|---|---|---|
| `OPENAI_API_KEY` | openai | |
| `GEMINI_API_KEY` | gemini | `GOOGLE_API_KEY` also resolves |
| `DEEPSEEK_API_KEY` | deepseek | |
| `XAI_API_KEY` | xai | |
| `AWS_ACCESS_KEY_ID` + `AWS_SECRET_ACCESS_KEY` | bedrock | an IAM user scoped to `bedrock:InvokeModel` is enough; `AWS_REGION` is optional and defaults to us-east-1 |
| `GCP_SERVICE_ACCOUNT_KEY` + `VERTEX_PROJECT_ID` | vertex | a service-account JSON key with `roles/aiplatform.user`; the job mints the OAuth token itself, so no token is stored |

And two balances topped up: Anthropic and Z.ai.

## What to do once they exist

Adding a secret turns the job **red on its own**. The row saying that
credential is absent no longer matches the environment, and the run names the
row to delete. That is the intended signal, not a bug.

1. Delete that provider's row from `UNARMED` in
   `crates/stella-model/tests/live_smoke/arming.rs`.
[...]

Issue #3856, opened 2026-08-19
live_smoke's credential skip reads as a pass: nine provider tests report `ok` for a run that made no call


macanderson:
## The problem

`crates/stella-model/tests/live_smoke.rs` gates each provider test on
`STELLA_LIVE_SMOKE=1` plus a resolvable credential, and its module doc states
the intended behaviour: *"Neither condition met -> the test prints why to
stderr and returns `Ok` (a clean skip, never a failure)"*.

The skip is not visible. `eprintln!` goes through libtest's output capture, so
a default `cargo test` shows nothing and reports the test as passing:

```
$ cargo test -p stella-model --test live_smoke      # no key, no STELLA_LIVE_SMOKE
test anthropic_smoke ... ok
test openai_smoke ... ok
test gemini_smoke ... ok
test bedrock_smoke ... ok
test vertex_smoke ... ok
test zai_smoke ... ok
test deepseek_smoke ... ok
test openrouter_smoke ... ok
test xai_smoke ... ok

test result: ok. 28 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.01s
```

Nine live-wire-shape assertions reported green in 0.01s having contacted
nothing. In the one line anybody actually reads — `28 passed; 0 ignored` —
this is indistinguishable from a run that spent money against nine live
endpoints and had every one accept the shape stella sends. That is the exact
failure mode #3011 paid for and #3097 names: *"Green with a key,
skipped-with-a-reason without one — never silently passing when it did not
run."*

The gate itself is right — `make gate` must not need a credential. What is
wrong is that the *skip is invisible*.

## Suggested fix

`#[ignore = "<reason>"]` is reported by libtest as
`test name ... ignored, <reason>` and shows in the summary line as `9 ignored`,
which no reader mistakes for a pass. The pattern is already used in this tree
(`crates/stella-graph/tests/watcher.rs:31`,
`crates/stella-tools/tests/search_recall.rs:378`, and
`crates/stella-tools/tests/chunk_retrieval_witnesses.rs`, which was written
this way *because* of this issue's shape). The invocation becomes
`STELLA_LIVE_SMOKE=1 … cargo test -p stella-model --test live_smoke -- --ignored`.

The other half of the same discipline, worth adopting with it: when a test is
asked for explicitly (`--ignored`) and its credential does not resolve, it
should **fail** rather than skip — at that point the absence of a key is a
failure to run what was asked for, not a hermetic default.

[...]

PR #6443, opened 2026-09-07
fix(stella-model): declare the live-smoke providers this repo cannot reach

### LINKED ISSUE #5595: live-smoke has been red on the weekly schedule: 6 of 9 provider credentials are absent from Actions, and Anthropic + Z.ai are out of credit
## Problem

The `live-smoke` workflow is failing on its schedule. [Run 33404499679](https://github.com/macanderson/stella/actions/runs/33404499679) (job `99528587994`, 2026-08-31T14:45:28Z) ends `1 passed; 8 failed` in 3.40s.

The failures are **two different problems**, and the distinction matters because only one of them costs money to fix. Reading a single provider's error and generalising gets this wrong — the eight failures split 6/2.

### Class A — the credential is not in the Actions environment (6 providers)

`bedrock`, `deepseek`, `gemini`, `openai`, `vertex`, `xai`. Each fails identically, e.g.:

```
[live_smoke] openai: no credential — no OPENAI_API_KEY env var and no `openai` entry in ~/.stella/credentials.toml

thread 'openai_smoke' panicked at crates/stella-model/tests/live_smoke.rs:615:9:
[live_smoke] openai: asked for but unrunnable — STELLA_LIVE_SMOKE is not `1`, or no
OPENAI_API_KEY env var and no `openai` entry in ~/.stella/credentials.toml. This test
runs only when named (`-- --ignored`), so a run that contacts nothing is a failure to
run what was asked for, not a skip.
```

This is the failure mode #3856 deliberately introduced — a missing secret must fail loudly rather than report `ok` after contacting nothing. It is working as designed; what it is telling us is that six required repository secrets are not reaching the job.

The workflow header (`.github/workflows/live-smoke.yml`) lists every required secret. Note `vertex` is not a plain secret: the job mints an OAuth token at runtime from `GCP_SERVICE_ACCOUNT_KEY`, so its failure may be a distinct cause from the other five.

### Class B — the credential works, the account is out of money (2 providers)

```
anthropic — status: HTTP 400
  terminal provider error: Anthropic HTTP 400 Bad Request:
  {"type":"error","error":{"type":"invalid_request_error","message":"Your credit balance
  is too low to access the Anthropic API. Please go to Plans & Billing to upgrade or
  purchase credits."},"request_id":"req_011Ceb1QEzYG28sf4dLegBBD"}

zai — status: HTTP 429
  terminal provider error: Z.ai HTTP 429: Insufficient balance or no resource package.
  Please recharge.
```

Both printed `verdict: UNDETERMINED`, which is correct and deliberate: the suite does not auto-classify a failure into wire-shape vs. billing, because a 400 carrying billing language is genuinely ambiguous with a 400 from a malformed request on Anthropic (the header says so, citing the 2026-07-21 run). **These two bodies name billing explicitly, so for this run the ambiguity is resolved by reading them — but that is a human read of this run, not a property the suite established.**

### Passing (1)

`openrouter` — `finish=Some(Stop)`, `cost_usd=0.00003036`. The only provider both configured and funded.

## Why this matters beyond the workflow

The suite exists because the Anthropic adapter shipped a broken `thinking` wire shape for months behind wiremock-only coverage (#240). While 8 of 9 adapters never reach their endpoint, that regression class is uncovered again for those eight — the guard is nominally armed and actually blind. The Anthropic credit exhaustion is also worth a look on its own: it is an account fact, not a stella fact, and likely affects more than this job.

## Where it lives

- `.github/workflows/live-smoke.yml` — schedule, the secret list, the Vertex token-minting steps
- `crates/stella-model/tests/live_smoke.rs` — `:615` (unrunnable arm), `:1097` (zai), `:1140` (anthropic)

## How to verify

`workflow_dispatch` the workflow and read the job log. A green run is `9 passed`. Per the workflow header, read the printed status/body before re-triggering — do not assume a re-run fixes anything.

## Things to keep in mind

- Every run spends real money on every armed provider. That is why it is schedule/manual only and never on `push`/`pull_request`.
- Do not make a missing credential skip again — that is exactly what #3856 removed, and it made a run that contacted nothing look identical to nine passing endpoints.
- Do not sniff response bodies for billing keywords to auto-skip. The header rules this out as fragile and liable to paper over a real regression.
- Per SCR-005 this carries only `triage`; sizing and priority belong to the triage agent.

## Definition of done

- [x] Each of the six absent credentials is either present in Actions secrets, or the provider is deliberately disarmed with the reason written down — no provider silently half-armed. `bedrock`, `deepseek`, `gemini`, `openai`, `vertex` and `xai` each carry a `NoCredential` row in `UNARMED` (`crates/stella-model/tests/live_smoke/arming.rs`) naming the secret that is absent, the run that showed it, and this issue. The row is re-asked every run: add the secret and the job goes red naming the row to delete.
- [x] Anthropic and Z.ai are disarmed on the same explicit terms. Both carry an `Unfunded` row quoting the provider's own words. Both still call their endpoints — an empty account is refused before anything is billed — and pass only on the `BILLING` verdict. A wire-shape or credential failure on either still fails the job, so Anthropic stays under the guard #240 was filed for.
- [x] One dispatched run reports `9 passed`. Run 34163434382 (2026-09-07, `workflow_dispatch`) ends `test result: ok. 9 passed; 0 failed`. One live call (`openrouter`, `finish=Some(Length)`, `cost_usd=0.00001834`) and eight declared rows, each printed by name. Anthropic and Z.ai both reached their endpoints and the log carries the refusal that confirmed each row.
- [x] A run that contacts nothing still fails, and a missing credential still names itself — proven, not assumed. In-tree witness: `a_named_smoke_with_no_credential_fails_rather_than_skipping` in `crates/stella-model/tests/live_smoke.rs` asserts `required_key` panics on an unresolved credential, with no env var and no network. Live evidence: run 33692795506 (2026-09-02) failed `bedrock`, `deepseek`, `gemini`, `openai`, `vertex` and `xai` by name, each printing the env var and the `~/.stella/credentials.toml` entry it wanted.

Filed by the automated PR sweep, 2026-08-31, per SCR-004. Causes quoted verbatim from the job log; the 6/2 split is observed per-provider, not inferred from one.

---

**Shipped in #6443.** The reporting half is done: a provider is armed, or a row says why it is not, and the row fails the job the moment it stops being true. ADR 0038 records the decision.

**The purchase is #6442** — six provider keys and two account top-ups, so the suite covers nine adapters again rather than one. It is written as a handoff: which secret, which provider, what to do as each one lands.

macanderson:
## What & why

The weekly `live-smoke` job has been red every week since 2026-08-24. Nine
adapters are armed and one can run. Six have no credential in this
repository's Actions secrets, and Anthropic and Z.ai have a working key over
an empty account. Run 34126157647 (schedule, 2026-09-07) ends
`1 passed; 8 failed`, the same as the four runs before it.

Red that always means the same thing is red nobody reads, and a real
wire-shape regression would land in the middle of it. The two green runs
before that were worse: run 32007329950 (2026-08-17) reports `ok` with eight
providers printing `skipped — set STELLA_LIVE_SMOKE=1 to run`. Green meant one
call was made. #3856 took that away, correctly, and left the job with no way
to say what a run covers.

**A provider is now armed, or a row in `UNARMED` says why it is not.** The
table is `crates/stella-model/tests/live_smoke/arming.rs`. Each row names the
provider, what is missing, the run that showed it, and the issue that owns it
— the declared-gap shape AGENTS.md already asks for when an event has no
consumer or a provider posture has no witness.

**The reason is re-asked on every run**, so a row cannot outlive what it
claims:

- a `NoCredential` row whose credential resolves fails the job, naming the row
  to delete;
- an `Unfunded` row whose credential has gone fails, because that reason
  claims the key works;
- an `Unfunded` provider whose call succeeds fails, because the account has
  money again.

So the day someone adds `OPENAI_API_KEY`, the job turns red asking to be
re-armed. A gap declared here cannot go quiet.

**An `Unfunded` provider still calls its endpoint.** An empty account is
refused before anything is billed, so the call is free and the refusal is
evidence. The smoke passes only on the `BILLING` verdict `resolve` already
derives. A rejected field resolves to `WIRE SHAPE` and still fails; a revoked
key resolves to `CREDENTIAL` and still fails. That is what keeps Anthropic —
the adapter #240 was filed on — under the guard rather than outside it.

ADR 0038 records the decision. SCR-002 is why it is a record rather than a
question put to the maintainer: the previous pass on this issue asked which
providers to arm and stopped.

Closes #5595

## The witness

Six of them, all offline, all in `arming.rs`. They fail on `main` because the
concept they assert does not exist there:

- `every_unarmed_row_names_a_live_provider_an_issue_and_a_reason` — every row
  points at a real `LIVE_PROVIDERS` id, is unique, cites an issue number, and
  says what is missing.
- `a_credential_that_resolves_makes_a_no_credential_row_stale` and
  `a_credential_that_vanishes_makes_an_unfunded_row_stale` — the self-retiring
  property, both directions.
- `a_balance_refusal_confirms_an_unfunded_row` — the two real bodies from run
  34126157647 pass their rows.
- `a_wire_shape_failure_still_fails_a_provider_declared_unfunded` — a
  `Malformed` and a 401 against Anthropic both still fail. This is the one
  that says the coverage was not traded away.
- `a_provider_with_no_row_is_never_excused_by_one` — a billing body from
  `openrouter`, which has no row, still fails.
- `a_success_fails_a_provider_declared_unfunded` — a funded call contradicts
  its row.

Two behaviors I also drove by hand, since a live job cannot be run from a
test:

- With no provider env vars and an empty `STELLA_HOME`, the six
  `NoCredential` smokes pass and print their rows, e.g.
  `[live_smoke] openai: unarmed by declaration (#5595) — no OPENAI_API_KEY in
  the repository secrets.`
- With `OPENAI_API_KEY` set, `openai_smoke` fails before any network call:
  `the unarmed row is stale — it says no credential is configured, and one
  resolves. Delete the `openai` row from UNARMED`.

## The gate

- [x] `cargo fmt --check`
- [x] `cargo clippy -p stella-model --all-targets -- -D warnings` — scoped, per
      CLAUDE.md's "CI builds and tests; this laptop does not". CI runs the
      workspace.
- [x] `cargo test -p stella-model --test live_smoke` — 32 passed, 9 ignored
- [x] `make guards-fast`, `make adr-numbering`, `make doc-links`,
      `python3 scripts/check-prose.py` — all green
- [x] Docs updated where behavior/flags changed — `live-smoke.yml`'s header
      now says what a green run means and how a provider gets re-armed
- [x] CLA signed
- [x] `Closes #5595` appears both above and as a commit trailer

## Fix over file

- [x] Extra fixes in this PR, two of them:

  **`anthropic_smoke` hand-rolled its own failure arm**, with a comment on it
  warning that the arm "must not drift from the shared one". It now shares
  `settle` with every other provider and hand-rolls only the success line that
  prints the two cache counters.

  **A driver session id could collide with its own sibling.**
  `plugin_cmd::session_id` salted a timestamp with 16 bits of the clock's
  sub-second remainder xored with the process id. Those bits repeat every 65.5
  microseconds, and one `stella plugin drive` invocation opens its sessions in
  a loop. This PR's first CI run hit it: run 34163526157 opened three sessions
  and the session log held two of them as `drive-1788818129-8a89`, failing
  `one_invocation_opens_a_session_for_each_sleep_the_driver_asks_for` on a
  change that touches neither file. A process-local counter now rides the id,
  so no two sessions one process opens can share one. Three witnesses, in
  `crates/stella-cli/src/plugin_cmd/tests.rs`: the property against a clock
  held still, the process-id half that still has to work, and 512 mints
  through the real `session_id` with no duplicate.

  `self_driving_cmd::state::new_run_id` has the same salt and is left alone.
  It is minted once per `stella self-driving run start` process, so the loop
  hazard is not there and two processes differ by pid. Saying so out loud
  rather than filing it, per CLAUDE.md.
- [x] Filed, with the reason a fix could not ride this PR: #6442 — the other
      half of #5595 is a purchase. Six provider keys and two account top-ups.
      It is written as a handoff: which secret, which provider, what to do
      when each one lands, and how to verify.

## Ground-rule check

- [x] No new outbound network calls (Stella never phones home) — the two
      `Unfunded` smokes call the endpoints they already called; nothing else
      changed about what this suite contacts

**The table is Rust, not YAML.** The workflow still passes every secret, which
is what makes adding one trip the staleness check. A disarm list in the
workflow could not be tested and the suite would have to read it anyway.

**`arming.rs` is reached through `#[path]`.** An integration test file is its
own crate root, so a bare `mod arming;` would look for `tests/arming.rs` — a
sibling with a name that says nothing about which suite owns it. The `#[path]`
puts the child in a folder named for its parent, which is the layout AGENTS.md
prescribes, and a comment at the `mod` line says so.

**The third DoD box has a real run behind it.** I dispatched
`live-smoke.yml` on this branch rather than assume. Run 34163434382 ends
`test result: ok. 9 passed; 0 failed`. One live call (`openrouter`,
`finish=Some(Length)`, `cost_usd=0.00001834`) and eight declared rows, each
printed by name. Anthropic and Z.ai both reached their endpoints and the log
carries the refusal that confirmed each row.

**What this does not do.** The six `NoCredential` providers have no live
coverage, exactly as before. The difference is that the tree says so in a file
review reaches, rather than a log nobody opens. #6442 is where that gets
bought back.

## Summary by Sourcery

Make live-smoke coverage explicit and self-validating while retaining regression detection for every provider that can be exercised.

New Features:
- Declare unavailable live-smoke providers with their coverage gaps and owning issues, while continuing to exercise providers that are reachable but unfunded.
- Revalidate declarations on every run so credentials, account funding, and successful calls automatically surface stale unarmed entries.

Bug Fixes:
- Preserve live-smoke failures for credential, billing, and wire-shape problems instead of allowing declarations to mask regressions.
- Prevent plugin driver session ID collisions when multiple sessions are created rapidly within one process.

Enhancements:
- Share live-smoke outcome handling across providers, including Anthropic-specific success reporting.
- Add offline witnesses covering provider declarations, stale-state detection, billing refusals, and failure classification.

Documentation:
- Document the armed-or-declared-unarmed live-smoke policy in the workflow and add ADR 0039 to the documentation index and manifest.

Tests:
- Add coverage for live-smoke declaration behavior and plugin session ID uniqueness.


macanderson:
Merge conflict resolved, and an ADR number collision fixed underneath it. Automated PR sweep.

**The conflict.** `docs/adr/README.md` and `docs/manifest.json`, both sides additive and disjoint — main's 0037 (#6437) against this branch's live-smoke arming record. A concurrent session pushed `9be20af` resolving the text while I was working; I reset onto it rather than fight it, so both merges are kept and `mergeable_state` is `blocked` rather than `dirty` now.

**What that resolution did not catch.** 0038 was claimed twice:

| branch | file |
|---|---|
| this PR | `docs/adr/0038-a-live-smoke-provider-is-armed-or-declared-unarmed.md` |
| #6441 | `docs/adr/0038-a-night-that-mostly-died-is-not-a-run.md` |

Both branches carry main's 0037, so both were numbered against the same tree and neither is wrong on its own. That is the composition break `docs/adr/README.md` § "The number is a shared cell" describes: `adr-numbering` runs on the merge commit and passes for each branch alone, nothing re-runs it when `main` moves, and the ADR number is the one shared cell with no file to conflict on — so git reported nothing. Whichever of the two merged second would have put two records at 0038 on `main`.

**This branch moved, #6441 keeps 0038.** Renumbered to **0039** in `ec36df8`, carrying every citation rather than the filename alone, since `check-doc-links` and `check-line-citations` both read past a bare number: the record's frontmatter `id`, title and heading, the README table row and its prose paragraph, the `docs/manifest.json` entry, the `doc:` citation in `crates/stella-model/tests/live_smoke/arming.rs`, and the ADR reference in `.github/workflows/live-smoke.yml`.

**Verified on the merged tree** — `check-adr-numbering` (38 records, every number unique, every heading and id in agreement, every record indexed), `check-doc-links` (1253 citations resolve), `check-prose` and `check-line-citations` all pass. No Rust logic changed; the workspace build and suite are CI's, per CLAUDE.md.

**One thing left for a human:** the PR description above still says "ADR 0038 records the decision" (and the Sourcery summary echoes it). I did not rewrite the body — another session is actively pushing here, and a wholesale body replacement would clobber its edits. The file, the index and every citation say 0039.

---

macanderson:
Follow-up on the renumber above: it turned `no edit is undone inside the branch` red, and that needs one command I am not going to run from an automated sweep. Everything else on the head is green or still running.

**Why it fired.** `check-rebase-replay` names the pair:

```
undone by  ec36df851  docs(adr): renumber the live-smoke arming record 0038 -> 0039
```

`docs/adr/0038-a-live-smoke-provider-is-armed-or-declared-unarmed.md` is created by an earlier commit on this branch and deleted by the renumber, so it is absent at the merge base and absent at the head — the branch's diff against `4eab7fa27` does not mention it, and it reads as untouched. AGENTS.md names this exact shape ("a file the branch created and deleted is absent at the merge base and at the head, exactly like an inherited rename"). The guard is behaving correctly; any renumber of an ADR the same PR introduced produces this pair.

**The hazard the guard warns about is nil for this path**, which I checked rather than assumed:

| tree | carries `0038-a-live-smoke-provider-…md`? |
|---|---|
| `origin/main` | no |
| `origin/fix/6025-nightly-unfinished-gate` (#6441) | no — it adds `0038-a-night-that-mostly-died-is-not-a-run.md` |

Nothing upstream carries the path, so a rebase that drops the create and keeps the undo would delete a file that does not exist. There is no upstream edit to revert. That makes this a false positive on consequence, not on shape — the guard cannot tell the two apart, and it should not be taught to.

**The remedy, and why I stopped.** The guard's own output gives it:

```
git reset --soft 4eab7fa27 && git commit    # recommit as one
```

That is a history rewrite plus a force-push, on a branch I did not create and that another session pushed to (`9be20af`) while this sweep was running. My operating rules forbid rewriting history on someone else's branch, and force-pushing over an active session risks its in-flight work. A merge commit cannot flatten a pair, so there is no non-rewriting fix.

**Why I left the renumber in place rather than reverting it.** Reverting clears this check and restores a latent break of `main`: two records at 0038, landing whenever the second of #6443/#6441 merges, with no pre-merge run on either branch able to see it. A red cosmetic guard on a stalled PR is much cheaper than that — nothing in the org can merge right now anyway, since no open PR carries an approving review.

**What a human or the session that owns this branch needs to do:** run the flatten above and force-push, after checking no other session is mid-push. The tree is byte-identical either way, so it changes no content and needs no re-review.

---

macanderson:
Replaced by #6447, which carries this exact tree — `git diff` between the two
heads is empty.

Why a new branch rather than a push here: a peer session renumbered the record
0038 to 0039 on this branch, correctly, because open PR #6441 had already
claimed 0038 against the same tree. Their commit left the branch creating a
path and then deleting it, and `rebase-replay` failed on it — correctly.
Rebasing this branch onto a base carrying #6441's 0038 would drop this
branch's creation of that path and keep the deletion, which is a live revert
of their record that merges cleanly (#4979).

The guard's remedy is to flatten the history so the path is never mentioned,
and flattening means rewriting what is already pushed. #6447 gets there
without a force-push: a fresh branch off `main` with the same two commits and
the same tree, and no mention of 0038 anywhere in its history.

The renumber is kept. Thank you to whoever caught the collision.

PR #6477, opened 2026-09-09
fix(scripts): stop an empty ci run list reading as ten unverified commits

### LINKED ISSUE #6475: main carries a commit nothing verified
`main-canary.yml`'s `check-main-verified.sh` job found commit(s) on
`main` that **nothing has verified** — as distinct from a commit a check said
no about, which `main-canary.sh`'s own issue already owns.

```
  cd7d53c  missing
      brand: adopt the Oxagen house system — palette, marks, and the face (#6473)
  a3a6294  missing
      chore(release): sync versions to 0.9.411 [skip release] (#6474)
  bfc63d0  missing
      fix(core,tools): place the deadline ladder against the turn's own budget, and stop search overclaiming (#6463)
  8c10e91  missing
      chore(release): sync versions to 0.9.410 [skip release] (#6472)
  508b6a0  missing
      build(deps): bump next from 16.2.11 to 16.3.3 in /website (#6469)
  c5c9e17  missing
      chore(release): sync versions to 0.9.409 [skip release] (#6471)
  778919f  missing
      chore(release): sync versions to 0.9.408 [skip release] (#6470)
  42c5e7f  missing
      chore(release): sync versions to 0.9.407 [skip release] (#6468)
  bb77ca5  missing
      fix(skills): aim the holdout at the skills whose control arm is short (#6459)
  7760b66  missing
      docs(scr): AGENTS.md names the closes-nothing waiver (#6452)
```

This is NOT "main is red". A failing run is a verified commit and the canary
owns that. These commits have no answer at all, which every other mechanism
here reads as green: the canary files only when its job runs and fails, the
red-main hold passes when no issue is open, and `gh run list` shows no row
for a run that was never created.

### How to fix

Re-dispatch the missing runs and let them finish before merging onto this
tree:

```sh
gh workflow run ci.yml --ref main
```

An Actions incident is one cause. There the runs start on their own once
capacity returns, and nothing was checking, which is the part this exists to
fix. A push made with the token a workflow run holds is the other cause, and
the release version write-back is one of those: that push raised no event at
all, so no run was ever created and none is coming. Nothing will start on its
own. Ask for it:

```sh
./scripts/dispatch-main-verification.sh
```

### Definition of done

- [x] Every recent commit on `main` has a completed `ci` run.

This script closes the issue itself on its next run that can answer clean, so
a box nobody ticks costs the issue nothing.

macanderson:
## What this fixes

`scripts/check-main-verified.sh` reads the CI history with one `gh run list`
call, then defaults each commit to `missing` when no run in that list matches
it. An empty list matches nothing, so all ten commits it checks become ten
separate findings at once.

`gh` returns `[]` with exit 0 for a transient API fault as readily as for a
genuinely empty history. On 2026-09-09 one such read failed `main-canary.yml`'s
`main carries no unverified commit` job and filed #6475, which named ten
commits as unverified — while `ci` run 34379997528 on the tip it named,
cd7d53c, had already concluded `success`. Every one of those ten commits has a
completed `ci` run, then and now.

An auth or permission failure was already handled: `gh` exits non-zero there
and the script takes its `unknown` path. I checked that directly — the same
query under an invalid token exits 1, which is why this run cannot have been a
permission problem and must have been an empty answer with exit 0.

## The fix

An empty run list becomes an `unknown`: reported loudly, exit 0. That is the
discipline the script's own header already states for every other unanswerable
read, and the header now names this one.

The two cases are separable rather than conflated. The read asks for the last
60 `ci` runs on `main` whatever their age, and an outage creates no runs and
destroys none, so it cannot empty that list — older runs stay in it. Empty
therefore means this run could not see the history at all.

The fix does not buy its safety with the coverage it exists to provide: a
commit absent from a *populated* list is still reported, and a new case pins
that direction alongside the three new ones.

## Witness

The three new cases in `scripts/test-main-verified.sh` fail on the previous
script and pass on this one. Checked by restoring `HEAD`'s copy of the script
and running the new suite against it:

```
FAIL an empty run list is an unknown, not a finding — expected exit 0, got 1
FAIL ...and says the list is what came back empty — expected exit 0, got 1
FAIL an empty run list filed an issue
main-verified: 35 passed, 3 failed
```

With the fix in place: `main-verified: 38 passed, 0 failed`.

`guard-self-tests.yml` runs this suite on `pull_request` with no paths filter,
so the witness runs in CI.

## Checks run locally

- `./scripts/test-main-verified.sh` — 38 passed, 0 failed
- `shellcheck` on both changed scripts — clean
- `make prose` — OK, none added
- `make line-citations` — OK, none added
- `make guards-fast` — green

Nothing compiled: this change is two shell scripts.

## Why this closes #6475

#6475 is the false report itself, filed by the defect this PR removes. Its
Definition of done — every recent commit on `main` has a completed `ci` run —
was already true when the issue was filed and is true now; I re-ran the fixed
checker against the real repository to confirm before ticking the box:

```
check-main-verified: OK — each of the last 10 commit(s) on main has a completed ci run.
```

## Not in this PR

`main` also has a second, unrelated red: the `docs` workflow's
`deploy stella.oxagen.sh` job fails with

```
AccessDeniedException ... assumed-role/gha-deploy-stella/GitHubActions is not
authorized to perform: ssm:SendCommand on ... instance/i-094fcb34c7e715cf8
```

That is IAM drift in AWS account 916294258235, not a defect in this tree —
`docs.yml` names the same instance id the infra's `terraform.tfvars` declares,
and the Terraform does grant `ssm:SendCommand` on it. Repairing it needs a
production apply in an account this repository cannot reach, so it is filed as a handoff
in #6478 rather than fixed here.

Closes #6475

### DIFF
--- a/scripts/check-main-verified.sh
+++ b/scripts/check-main-verified.sh
@@ -61,10 +61,11 @@
 #
 # ── It fails OPEN, at every unknown ──────────────────────────────────────────
 #
-# No `gh`, an unreachable API, an unparseable answer: report and exit 0. This
-# is a monitor, and a monitor that can itself block a merge is worse than the
-# gap it watches — the same argument `main-red-claim.sh`'s header makes about
-# a claim check. Every unknown is loud, never silent.
+# No `gh`, an unreachable API, an unparseable answer, a run list that came back
+# empty so no commit could be matched: report and exit 0. This is a monitor,
+# and a monitor that can itself block a merge is worse than the gap it watches
+# — the same argument `main-red-claim.sh`'s header makes about a claim check.
+# Every unknown is loud, never silent.
 #
 # ── `--announce` reaches a person, on codeql-canary.sh's shape ───────────────
 #
@@ -231,6 +232,25 @@ elif ! runs="$(gh run list --workflow ci.yml --branch main --limit 60 \
   --json headSha,status,conclusion,createdAt \
   --jq '.[] | "\(.headSha) \(.status) \(.conclusion // "none") \(.createdAt)"' 2>/dev/null)"; then
   unknown "could not reach the Actions API"
+fi
+
+# An empty list is a blind read, not a finding about every commit at once.
+#
+# `gh` answers a transient API fault with `[]` and exit 0 as readily as it
+# answers a genuinely empty history, and the loop below defaults each commit
+# to `missing`. So on 2026-09-09 one bad read reported all ten commits as
+# unverified and filed an issue saying nothing had checked a tree whose `ci`
+# run on cd7d53c had already concluded success. An auth or permission failure
+# exits non-zero and is already caught above; this is the shape that does not.
+#
+# The two cases are separable because the read asks for the last 60 `ci` runs
+# on `main` whatever their age. An outage creates no runs, and destroys none,
+# so it cannot empty that list — older runs stay in it, and a commit absent
+# from a populated list is still reported, which is the gap this script
+# watches. Empty means this run could not see the history at all, and that is
+# an unknown.
+if [ -z "$runs" ]; then
+  unknown "the ci run list came back empty, so no commit could be matched"
 fi
 
 now_epoch="$(date -u +%s 2>/dev/null || echo 0)"
--- a/scripts/test-main-verified.sh
+++ b/scripts/test-main-verified.sh
@@ -199,6 +199,44 @@ case "$out" in
   *) fail=$((fail + 1)); echo "FAIL an unknown exited 0 without saying so:"; echo "$out" ;;
 esac
 
+# ── E: an empty run list is a blind read ─────────────────────────────────────
+#
+# `gh run list` returns `[]` with exit 0 for a transient fault as readily as
+# for an empty history, and the verdict loop defaults every commit to
+# `missing`. On 2026-09-09 that turned one bad read into ten findings and a
+# filed issue claiming nothing had verified a tree whose `ci` run had gone
+# green seventeen seconds earlier. An auth failure exits non-zero and was
+# already an unknown; this shape was not.
+empty_list_commits="$(commit $A 'a merge the API could not answer for')
+$(commit $B 'the merge before it')"
+
+want "an empty run list is an unknown, not a finding" expect-pass \
+  "$empty_list_commits" "" \
+  "UNKNOWN"
+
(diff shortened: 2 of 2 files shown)


macanderson:
### Answering Sourcery's ❌ row

> **Objective:** Resolve the currently missing verification runs on main by
> redispatching them and ensuring every recent commit has a completed CI run.

This row asks for work that is not owed, because the condition it names never
existed. There were no missing verification runs on `main`.

That objective was read out of the **"How to fix"** section of #6475. The
script writes that section into every issue of this kind from a fixed
template, before it knows anything about the cause. On this occasion the
finding it accompanied was false, so its suggested remedy — re-dispatch the
missing runs — describes a repair with nothing to repair.

Here is the run history for the exact ten commits #6475 named. The canary
evaluated them at **16:58:16 UTC on 2026-09-09**:

| Commit | `ci` conclusion | Concluded (UTC) | Run |
|---|---|---|---|
| cd7d53c9 | success | 17:20:12 | 34379997528 |
| a3a62948 | success | 07:33:52 | 34322650858 |
| bfc63d0b | success | 06:46:17 | 34318425719 |
| 8c10e917 | success | 02:08:28 | 34300557854 |
| 508b6a08 | success | 01:21:47 | 34298861144 |
| c5c9e171 | success | 01:01:18 | 34296210150 |
| 778919f9 | success | 00:41:04 | 34294550435 |
| 42c5e7fa | success | 2026-09-09 00:17:00 | 34292729785 |
| bb77ca5e | success | 2026-09-09 00:02:14 | 34290615804 |
| 7760b663 | success | 2026-09-08 23:29:51 | 34290609476 |

Nine of the ten had already concluded `success` when the canary looked. The
tenth, the tip cd7d53c9, had a run created at 16:57:59 — seventeen seconds
before the check — which went on to conclude `success` at 17:20:12.

So the correct verdict at that moment was nine verified and one pending, which
the script reports as `PENDING` and exits 0 on, filing nothing. It instead
reported ten as `missing` and filed an issue. That gap between the two is the
whole defect, and it is what this PR fixes.

Re-dispatching anything now would spend CI runs re-asking a question that has
ten `success` answers already. The two rows Sourcery marked ✅ are the actual
objectives, and the box on #6475 is ticked on the strength of the table above
rather than on this PR having run anything.

Issue #6478, opened 2026-09-09
Merged changes stop reaching the docs site because the deploy role lost permission to restart the server


macanderson:
## The problem

Every merge to `main` now leaves the `docs` workflow red. Its
`deploy stella.oxagen.sh` job cannot restart the docs site, so merged doc and
website changes stop reaching stella.oxagen.sh.

The job fails with:

```
aws: [ERROR]: An error occurred (AccessDeniedException) when calling the
SendCommand operation: User:
arn:aws:sts::916294258235:assumed-role/gha-deploy-stella/GitHubActions is not
authorized to perform: ssm:SendCommand on resource:
arn:aws:ec2:us-east-1:916294258235:instance/i-094fcb34c7e715cf8 because no
identity-based policy allows the ssm:SendCommand action
```

The site itself is still up. The deploy step checks health and the node rolls
back to the last good release, so what is broken is shipping new builds, not
serving the old one.

## When it started

Last green run: 2026-09-07 22:43 UTC. It has failed on all three runs since:

| When (UTC) | Run |
|---|---|
| 2026-09-08 23:25 | 34290615920 |
| 2026-09-09 01:21 | 34298861126 |
| 2026-09-09 16:57 | 34379997376 |

## This is not a defect in this repository

Nothing in the tree changed to cause it. `.github/workflows/docs.yml` has not
been touched since the actions-group bump in #6045, well before the first
failure. The instance id it sends to is hardcoded there and still matches what
the infrastructure declares:

- `.github/workflows/docs.yml` — the role it assumes and the instance it sends
  to: `arn:aws:iam::916294258235:role/gha-deploy-stella` and
  `i-094fcb34c7e715cf8`.
- `oxagen-aws-infra`, `stacks-new/ci-deploy/terraform.tfvars` — declares
  `node_instance_id = "i-094fcb34c7e715cf8"`. The same value.
- `oxagen-aws-infra`, `stacks-new/ci-deploy/roles.tf` — its `RestartService`
  statement does grant `ssm:SendCommand` on that instance.

So the Terraform is right and the live IAM policy has drifted away from it.

## What I could not check

I could not read the live policy. A shell here resolves AWS credentials to
account 578673726240, not the 916294258235 account this role lives in, so
"the live policy no longer carries the statement" is an inference from the
error message and the unchanged code, not something I observed directly.
Confirming it is the first step below.

## Follow-up from the triage comment already on this issue

The existing comment narrows the fix further and should be read before
[...]

PR #6479, opened 2026-09-09
fix(scripts): an empty ci run list is an unknown, not ten unverified commits

### LINKED ISSUE #6475: main carries a commit nothing verified

macanderson:
## What & why

`scripts/check-main-verified.sh` filed a false alarm this afternoon. At
16:58 UTC it reported the ten newest commits on `main` as having no `ci` run
at all and opened this issue. Every one of those ten commits has a completed,
**successful** `ci` run:

```
cd7d53c9 completed success 2026-09-09T16:57:59Z   <- reported "missing"
a3a62948 completed success 2026-09-09T07:12:01Z   <- reported "missing"
bfc63d0b completed success 2026-09-09T06:17:43Z   <- reported "missing"
...all ten
```

The script guards its **commit** read for emptiness and never guards its
**run** read. The matching loop opens every commit at `verdict="missing"` and
only a row for that commit moves it, so a run list that came back with zero
rows did not report one absence — it reported the whole window as absent at
once, with no row anywhere able to contradict it. `gh` exits 0 on an empty
page, so the existing `if ! runs=...` check for a failed call never saw it.

A read that returned nothing cannot tell "no `ci` run exists for any of these
commits" from "the API declined to list them", and those are opposite states.
The script's own header already decided which way that resolves — *"It fails
OPEN, at every unknown"* — because a monitor that fabricates blocks merges
during exactly the incident its repair has to land in, and one that files
falsely is one nobody reads the second time. An empty page is now one of those
unknowns, and it is named in the header beside `gh` being absent and the API
being unreachable.

**Only the empty page.** A list that carries rows but none for a given commit
is the outage this script was written to catch, and still reports. A
negative-control case pins that direction so the new guard cannot be satisfied
by refusing to answer whenever a commit has no run of its own.

Closes #6475

## The witness

`scripts/test-main-verified.sh`, run against this branch:

| | result |
|---|---|
| before the fix | **35 passed, 2 failed** |
| after the fix | **37 passed, 0 failed** |

The two failing assertions are `an empty run list is UNKNOWN, not every commit
missing` and `...and names an empty run list as the reason`. The third new
case — `a run list carrying nothing for the commit is still unverified` —
passes on both sides by design: it is the negative control, and it is what
stops the fix being a blanket "answer nothing when in doubt".

Verified against the real repository as well: `./scripts/check-main-verified.sh`
on this branch prints `OK — each of the last 10 commit(s) on main has a
completed ci run.` and exits 0, which is the state the issue claims otherwise.

## The gate

- [x] `cargo fmt --check` — via `make guards-fast`
- [x] workspace clippy and the workspace suite — no Rust in this diff; CI runs both
- [x] Docs updated where behavior changed — the script's own fail-open header now names the empty run list
- [x] CLA signed
- [x] `Closes #6475` appears both above and as a commit trailer

`make guards-fast`, `make prose`, `make line-citations` and `make shellcheck`
all pass on this branch. No test was deleted.

## Fix over file

- [x] Extra fixes in this PR: none. I checked `scripts/check-ci-tests.sh`,
      which reads the same run list the same way — it is **not** affected: with
      an empty list its `verdict` stays empty and it already exits 0 through
      `unknown`, so it fails open as intended.
- [x] Filed, with the reason a fix could not ride this PR: **#6481**. Sweeping
      the sibling guards after this fix turned up the same shape in
      `scripts/check-releases-published.sh`: its releases read is unguarded, and
      an empty page makes `$published` empty, so every non-grandfathered tag
      past the grace window reports as `absent`. It did not ride this PR because
      the obvious guard collides with the script's own purpose — "many tags,
      zero releases" is *also* the genuine catastrophic finding it exists to
      report, and nothing in the payload separates a broken read from a real
      total failure. Which cost to pay is a maintainer call, so it is written up
      with the three options rather than decided here.

Unrelated to this diff, and reported rather than fixed because it is an AWS
IAM change outside this repository: **the docs site deploy has failed on every
push to `main` since 2026-09-08 23:25 UTC** (three consecutive runs of
`docs`). `stella.oxagen.sh` has not been updated since.

```
User: arn:aws:sts::916294258235:assumed-role/gha-deploy-stella/GitHubActions
is not authorized to perform: ssm:SendCommand on
arn:aws:ec2:us-east-1:916294258235:instance/i-094fcb34c7e715cf8
because no identity-based policy allows the ssm:SendCommand action
```

The role needs `ssm:SendCommand` on that instance. Nothing in this tree can
grant it.

### DIFF
--- a/scripts/check-main-verified.sh
+++ b/scripts/check-main-verified.sh
@@ -61,10 +61,10 @@
 #
 # ── It fails OPEN, at every unknown ──────────────────────────────────────────
 #
-# No `gh`, an unreachable API, an unparseable answer: report and exit 0. This
-# is a monitor, and a monitor that can itself block a merge is worse than the
-# gap it watches — the same argument `main-red-claim.sh`'s header makes about
-# a claim check. Every unknown is loud, never silent.
+# No `gh`, an unreachable API, an empty run list, an unparseable answer: report
+# and exit 0. This is a monitor, and a monitor that can itself block a merge is
+# worse than the gap it watches — the same argument `main-red-claim.sh`'s
+# header makes about a claim check. Every unknown is loud, never silent.
 #
 # ── `--announce` reaches a person, on codeql-canary.sh's shape ───────────────
 #
@@ -231,6 +231,26 @@ elif ! runs="$(gh run list --workflow ci.yml --branch main --limit 60 \
   --json headSha,status,conclusion,createdAt \
   --jq '.[] | "\(.headSha) \(.status) \(.conclusion // "none") \(.createdAt)"' 2>/dev/null)"; then
   unknown "could not reach the Actions API"
+fi
+
+# Zero rows answers a different question than it looks like it answers. The
+# loop below opens every commit at `missing` and only a matching row moves it,
+# so an empty list does not report one absence — it reports the whole window as
+# absent, in one breath, with no row anywhere to contradict it. A read that
+# returned nothing cannot distinguish "no `ci` run exists for any of these
+# commits" from "the API declined to list them", and those are opposite states.
+#
+# `gh` exits 0 on an empty page, so the check above never sees it. On
+# 2026-09-09 one such page filed against the ten newest commits at once; each
+# of them had a completed, successful run, and the same query twenty minutes
+# later returned all sixty rows. The header's rule decides it: every unknown
+# exits 0 and says so, because a monitor that fabricates is one nobody reads
+# the second time.
+#
+# This is only the empty page. A list that carries rows but none for a given
+# commit is the outage this script was written to catch, and still reports.
+if [ -z "$runs" ]; then
+  unknown "the run list came back empty — this read listed no ci run at all for main, which cannot tell an outage from an unanswered query"
 fi
 
 now_epoch="$(date -u +%s 2>/dev/null || echo 0)"
--- a/scripts/test-main-verified.sh
+++ b/scripts/test-main-verified.sh
@@ -199,6 +199,35 @@ case "$out" in
   *) fail=$((fail + 1)); echo "FAIL an unknown exited 0 without saying so:"; echo "$out" ;;
 esac
 
+# Zero rows is the API declining to answer, not an answer that no `ci` run
+# exists for any of these commits. Read the second way, one empty page turns
+# every commit in the window into a finding at once: on 2026-09-09 that filed
+# against ten commits which each had a completed, successful run, and a monitor
+# that files falsely is one nobody reads twice.
+empty_runs_commits="$(commit $A 'a merge whose run the API did not list')
+$(commit $B 'and the merge before it')"
+if out="$("$SCRIPT" --fixture-commits "$empty_runs_commits" --fixture-runs "" 2>&1)" &&
+  [ -z "${out##*UNKNOWN*}" ]; then
+  pass=$((pass + 1)); echo "ok   an empty run list is UNKNOWN, not every commit missing"
+else
+  fail=$((fail + 1)); echo "FAIL an empty run list was read as a finding:"; echo "$out"
+fi
+
+# The report names the shape it could not read, so a reader of the log can
+# tell this apart from the outage the script exists to catch.
+case "$out" in
+  *"listed no ci run at all"*) pass=$((pass + 1)); echo "ok   ...and names an empty run list as the reason" ;;
+  *) fail=$((fail + 1)); echo "FAIL the unknown did not say the run list was empty:"; echo "$out" ;;
+esac
+
+# The other direction, so the guard above cannot be satisfied by refusing to
+# answer whenever a commit has no run: a NON-empty list that simply carries
+# nothing for this commit is still the outage this script was written for.
+want "a run list carrying nothing for the commit is still unverified" expect-fail \
+  "$(commit $B 'a merge with no run of its own')" \
+  "$A completed success $(now_iso)" \
+  "missing"
+
 # ── N: several commits, one bad ──────────────────────────────────────────────
 # The reported window is a run of merges, not one; a guard that stopped at the
 # first verified commit would have missed three of the four that day.


macanderson:
## Answering the ❌ row in Sourcery's assessment

> **Ensure the affected commits on `main` obtain completed `ci` runs, including by redispatching genuinely missing runs where necessary.** — *The PR changes monitoring behavior and tests the empty-response case, but it does not dispatch or re-dispatch workflow runs. It assumes the reported commits already had successful runs and therefore does not itself establish verification for genuinely missing commits.*

This row reads #6475 as a report of a real outage. It was not one, and there is nothing to re-dispatch — so the objective is not deferred or out of scope, it is **vacuous for this issue**. The assumption Sourcery names is not an assumption; it is the measurement the PR is built on.

**Every one of the ten commits #6475 names already had a completed, successful `ci` run at the moment the issue was filed.** The issue was created at 16:58:16 UTC. Running the script's own query — unchanged, same workflow, same branch, same limit:

```
$ gh run list --workflow ci.yml --branch main --limit 60 \
    --json headSha,status,conclusion,createdAt
cd7d53c9 completed success 2026-09-09T16:57:59Z
a3a62948 completed success 2026-09-09T07:12:01Z
bfc63d0b completed success 2026-09-09T06:17:43Z
8c10e917 completed success 2026-09-09T01:46:46Z
508b6a08 completed success 2026-09-09T01:21:26Z
c5c9e171 completed success 2026-09-09T00:43:26Z
778919f9 completed success 2026-09-09T00:19:57Z
42c5e7fa completed success 2026-09-08T23:54:56Z
bb77ca5e completed success 2026-09-08T23:25:57Z
7760b663 completed success 2026-09-08T23:25:52Z
```

Nine of those ten concluded **hours** before the issue was filed; the tenth, `cd7d53c9`, was created at 16:57:59 and has since concluded `success`. A `workflow_dispatch` for any of them would start a second run of a suite that already answered.

**The issue's own definition of done is already satisfied, and the check now says so.** `./scripts/check-main-verified.sh` on this branch, against the live repository:

```
check-main-verified: OK — each of the last 10 commit(s) on main has a completed ci run.
exit: 0
```

That is the same script, reading the same API, reaching the opposite verdict from the one that filed the issue — which is precisely the defect. So the honest reading of that DoD line is that it was **never unmet**: the monitor asserted an absence that was not there.

**Why the row is worth answering rather than silently accepting.** Treating it as owed work would mean dispatching ten redundant full-workspace CI runs to "establish" verification that already exists — spending an hour of runners to answer a question with an answer already on file. That is the cost the false alarm imposes, and re-dispatching would be paying it a second time.

The one thing genuinely worth checking is whether a *future* real outage still reports, and it does: the negative-control case `a run list carrying nothing for the commit is still unverified` passes on both the old and new code by design. A list that carries rows but none for a given commit is the outage this script exists to catch, and the new guard covers only the zero-row page.

macanderson:
Closing as a duplicate of #6477, which merged at 21:12:57Z and reached the same diagnosis and the same fix: the run read had no emptiness guard, `gh` exits 0 on an empty page, and the verdict loop defaults every commit to `missing`, so one blind read became ten findings.

#6477 is the better of the two and nothing here is worth salvaging on top of it. Its test suite covers both of my new cases and adds a third — that an empty read files no issue at all — and its write-up settles one thing mine only asserted: it checked directly that the same query under an invalid token exits 1, which rules out a permission fault and leaves the empty-answer-with-exit-0 shape as the only explanation.

**How this collided, since the machinery exists to stop it.** I ran `./scripts/issue-claim.sh check 6475` at 20:48:10Z and it correctly reported the issue unclaimed — #6477 was not opened until 20:54:26Z, and no claim comment had been posted against it. I then claimed it and started work. What I did not do was **re-run the check before `gh pr create`**, which is exactly what AGENTS.md asks for:

> Run it **before writing code and again before opening the PR**. The gap between those two is enough: a peer's PR can merge inside one issue's worth of work, and then the check that was clean at the start is stale at the end.

The second run would have found #6477 with `Closes #6475` open against it and stood me down. The check was not wrong and it was not missing — it was asked once when it needed asking twice.

**What does not close with this PR**, because neither is covered by #6477:

- **#6481** — `scripts/check-releases-published.sh` has the same unguarded-read shape, found by sweeping the siblings after this fix. There the empty case is genuinely ambiguous, because "many tags, zero releases" is also the real catastrophic finding the script exists to report, so it is written up with three options for a maintainer rather than decided.
- **#6478** — the docs-site deploy has been failing since 2026-09-08 23:25 UTC on an IAM drift (`gha-deploy-stella` lost `ssm:SendCommand`). I added the root-cause narrowing to that issue: the Terraform has not changed since 2026-09-02, the old-account stack cannot be the source, and `stacks-new/ci-deploy` is safe to apply in isolation despite four uncommitted edits in the infra tree.

